Issue #2620, opened 2016-04-28
t0170 fails randomly


whyrusleeping:
https://s3.amazonaws.com/archive.travis-ci.org/jobs/126450891/log.txt

@noffle could you take a look at this?

whyrusleeping:
https://s3.amazonaws.com/archive.travis-ci.org/jobs/129801262/log.txt

whyrusleeping:
https://s3.amazonaws.com/archive.travis-ci.org/jobs/129545003/log.txt

whyrusleeping:
https://s3.amazonaws.com/archive.travis-ci.org/jobs/129544938/log.txt

whyrusleeping:
https://s3.amazonaws.com/archive.travis-ci.org/jobs/129527732/log.txt

whyrusleeping:
http://ci.ipfs.team:8111/viewLog.html?tab=buildLog&logTab=tree&filter=debug&expand=all&buildId=1710#_focus=19936

whyrusleeping:
http://ci.ipfs.team:8111/viewLog.html?tab=buildLog&logTab=tree&filter=debug&expand=all&buildId=1735#_focus=19961

whyrusleeping:
https://s3.amazonaws.com/archive.travis-ci.org/jobs/130623729/log.txt

whyrusleeping:
https://s3.amazonaws.com/archive.travis-ci.org/jobs/130623567/log.txt

PR #2720, opened 2016-05-17
marks dht query as known breakage for now

noffle:
tracked by https://github.com/ipfs/go-ipfs/issues/2620


Kubuxu:
LGTM, and was talked through on IRC.

Issue #2239, opened 2016-01-24
Read only API has CORS disabled


Kubuxu:
Read-only API is accessible on public gateway but has CORS disabled.
This makes it impossible to use it from JS in browser which was a point of it.

IMHO CORS handling in general should be as follows:
- gateway: CORS fully allowed (simple and preflighted requests on wild-card domain). Options request added in #2232, we still have to allow CORS..
- read-only API on gateway: as above
- read-write API: CORS allowed for domains white-listed by user (simple and preflighted) until #1532 (API Tokens) come in then CORS fully allowed.

There were opinions that read-only API isn't that read-only as it can make user download some content. It shouldn't be taken as issue as it is already possible with just gateway.

(API might want to introduce change that `GET` uses data only available locally and specifies non-standard request method (i.e. `FETCH`) that downloads data from the network, this way `GET` does not have any side-effects).

Example of missing CORS on read-only API response:

```
± curl -v http://ipfs.io/api/v0/object/get?arg=/ipfs/QmbfChZw78fWH6ckgAb6mWLqm7k8gJJUVKgSYvFw2XiKiq
*   Trying 178.62.61.185...
* Connected to ipfs.io (178.62.61.185) port 80 (#0)
> GET /api/v0/object/get?arg=/ipfs/QmbfChZw78fWH6ckgAb6mWLqm7k8gJJUVKgSYvFw2XiKiq HTTP/1.1
> Host: ipfs.io
> User-Agent: curl/7.46.0
> Accept: */*
>
< HTTP/1.1 200 OK
< Server: nginx/1.9.3
< Date: Sun, 24 Jan 2016 10:30:22 GMT
< Content-Type: application/json
< Transfer-Encoding: chunked
< Connection: keep-alive
< Access-Control-Allow-Headers: X-Stream-Output, X-Chunked-Output
< Access-Control-Expose-Headers: X-Stream-Output, X-Chunked-Output
< Trailer: X-Stream-Error
< Strict-Transport-Security: max-age=15768000
<
{"Links":[],"Data":"\u0008\u0002\u0012\ufffd\u0001#!/usr/bin/env bash\n# adds files to IPFS and creates symlink to FUSE\n# probably will break in some cases and remove your files.\nHASH=$(ipfs add -r -q $1 | tail -1)\nrm -rf $1\nln -s \"/ipfs/$HASH\" \"$1\"\n\u0018\ufffd\u0001"}
* Connection #0 to host ipfs.io left intact
```

[...]

PR #2723, opened 2016-05-17
Add CORS headers to Read Only Gateway Default config

Kubuxu:
Works toward #2239.
*Do not review it yet*


Kubuxu:
I thought it will be simple ... again.

PR #2724, opened 2016-05-17
Remove note about Go1.6 not being supported

Kubuxu:
Recommend Go1.6.2
Note that Go1.5 is still supported

License: MIT

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,8 @@ From there:
 
 #### Install Go
 
-First, you'll need Go. If you don't have it: [Download Go 1.5.2+](https://golang.org/dl/). **Go 1.6 is not yet supported.**
+Build process of ipfs requires at least Go 1.5, but we strongly recommend using 1.6.2. If you don't have it: [Download Go 1.6.2+](https://golang.org/dl/).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -57,7 +57,8 @@ From there:
 
 #### Install Go
 
-First, you'll need Go. If you don't have it: [Download Go 1.5.2+](https://golang.org/dl/). **Go 1.6 is not yet supported.**
+First, you'll need Go. If you don't have it: [Download Go 1.6.2+](https://golang.org/dl/). _Go 1.5 is still being supported._
```
whyrusleeping: I'd phrase it something like:

```
Ipfs requires at least go 1.5 to build, but we strongly recommend using 1.6.2. If you don't have it....
```

Kubuxu: I will phrase it as:
`Build process of ipfs requires at least go 1.5 to build, but we strongly recommend using 1.6.2. If you don't have it....`

Just to not capitalize ipfs (small habit from cjdns).

whyrusleeping:
LGTM

PR #2728, opened 2016-05-18
metrics: add transport label to p2p_peers_total

lgierth:
Gives us per-transport peers counts:

	ipfs_p2p_peers_total{transport="/ip4/tcp"} 25
	ipfs_p2p_peers_total{transport="/ip6/tcp"} 13
	ipfs_p2p_peers_total{transport="/ip6/udp/utp"} 17

License: MIT

### DIFF
--- a/core/corehttp/metrics.go
+++ b/core/corehttp/metrics.go
@@ -29,7 +29,7 @@ func MetricsCollectionOption(handlerName string) ServeOption {
 var (
 	peersTotalMetric = prometheus.NewDesc(
 		prometheus.BuildFQName("ipfs", "p2p", "peers_total"),
-		"Number of connected peers", nil, nil)
+		"Number of connected peers", []string{"transport"}, nil)
 )
 
 type IpfsNodeCollector struct {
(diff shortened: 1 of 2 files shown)


core/corehttp/metrics.go
```diff
@@ -41,13 +41,24 @@ func (_ IpfsNodeCollector) Describe(ch chan<- *prometheus.Desc) {
...
 }
 
-func (c IpfsNodeCollector) PeersTotalValue() float64 {
-	return float64(len(c.Node.PeerHost.Network().Conns()))
+func (c IpfsNodeCollector) PeersTotalValues() map[string]float64 {
+	vals := map[string]float64{}
+	for _, conn := range c.Node.PeerHost.Network().Conns() {
+		tr := ""
+		for _, proto := range conn.RemoteMultiaddr().Protocols() {
+			tr = tr + "/" + proto.Name
```
lgierth: Maybe this would be good to have in go-multiaddr, `Multiaddr.ProtocolsString()`

lgierth:
Hold on, I'll also add a version label, so that we can also graph peer counts by remote-version.

lgierth:
Mh does anything store the remote version? The peerstore maybe? I don't wanna fire any requests, just read existing data.

lgierth:
Ok let's go ahead with this one, without remote-version.

> 17:17+0200 \<@lgierth> daviddias: ideally the agent version -- e.g. that peer is "foo-ipfs/1.2.3"
> 17:18+0200 \<@lgierth> i just thought i'd add that dimension to the peers count metric too while i'm at it -- i got the transport dimension yesterday, e.g. transport="/ip6/udp/utp"
> 17:21+0200 \<@daviddias> lgierth: we don't really store or send that
> 17:21+0200 \<@daviddias> we treat each proto as a standalone thing
> 17:21+0200 \<@daviddias> and the handshake happens per stream
> 17:22+0200 \<@daviddias> for all that is worth, the ipfs version doesn't matter between two peers, as long as they have compatible wire protocols
> 17:22+0200 \<@daviddias> that being said, it would be just one more thing to add to the Identify protocol
> 17:22+0200 \<@daviddias> it can be added
> 17:24+0200 \<@lgierth> ok sounds good
> 17:24+0200 \<@lgierth> i'll leave it be for now and add it to my metrics wishlist

core/corehttp/metrics.go
```diff
@@ -41,13 +41,24 @@ func (_ IpfsNodeCollector) Describe(ch chan<- *prometheus.Desc) {
...
+			val,
+			tr,
+		)
+	}
 }
 
-func (c IpfsNodeCollector) PeersTotalValue() float64 {
-	return float64(len(c.Node.PeerHost.Network().Conns()))
+func (c IpfsNodeCollector) PeersTotalValues() map[string]float64 {
+	vals := map[string]float64{}
```
whyrusleeping: i prefer using `make` for this

whyrusleeping:
I think we store the remote version in the peerstore, although we can add that in later.

whyrusleeping:
aside from one nitpick, this LGTM

core/corehttp/metrics.go
lgierth: Made it `make`

PR #2453, opened 2016-03-07
build: include commit hash & version

hutenosa:
(partially) fixes #2304 

New script `bin/genversion` generates SemVer-compatible string out of closest git tag. The script is called in Makefile to pass IPFS version and commit info as linker arguments.
Dockerfiles got optional build argument `IPFS_VERSION` to avoid including `.git/` into the image. 

License: MIT
Signed-off-by: hutenosa <hutenosa@mm.st>

PR #2734, opened 2016-05-18
docker: fix ipfs version --commit

lgierth:
The dockerfile assumes we're always on a branch, in which case it resolves the head ref in `.git/HEAD` via `.git/refs/`. Solarnet deploys are headless though, i.e. not on a branch but on a specific commit. In this case, Git puts that commit ref directly into `.git/HEAD`, so that the resolution step breaks.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -53,8 +53,8 @@ RUN apk add --update musl go=$GO_VERSION git bash wget ca-certificates \
 	# We get the current commit using this hack,
 	# so that we don't have to copy all of .git/ into the build context.
 	# This saves us quite a bit of image size.
-	&& ref="$(cat .git/HEAD | cut -d' ' -f2)" \
-	&& commit="$(cat .git/$ref | head -c 7)" \
+	&& ref=$(cat .git/HEAD | grep ref | cut -d' ' -f2) \
+	&& commit=$(if [ -z "$ref" ]; then cat .git/HEAD; else cat ".git/$ref"; fi | head -c 7) \
 	&& echo "ldflags=-X github.com/ipfs/go-ipfs/repo/config.CurrentCommit=$commit" \
 	# Build and install IPFS and entrypoint script
 	&& cd $SRC_PATH/cmd/ipfs \
--- a/test/sharness/t0300-docker-image.sh
+++ b/test/sharness/t0300-docker-image.sh
@@ -66,6 +66,14 @@ test_expect_success "simple ipfs add/cat can be run in docker container" '
 	test_cmp expected actual
 '
 
+test_expect_success "version CurrentCommit is set" '
+	docker_exec "$DOC_ID" "wget --retry-connrefused --waitretry=1 --timeout=30 -t 30 \
+		-q -O - http://localhost:8080/version" | grep Commit | cut -d" " -f2 >actual &&
+	docker_exec "$DOC_ID" "ipfs version --commit" | cut -d- -f2 >expected &&
+	[ "$(cat expected | wc -c)" -gt "1" ] && # check there actually is a commit set
+	test_cmp expected actual
+'
+
 test_expect_success "stop docker container" '
 	docker_stop "$DOC_ID"
 '
--- a/test/Dockerfile
+++ b/test/Dockerfile
@@ -44,8 +44,8 @@ RUN apk add --update musl go=$GO_VERSION git bash wget ca-certificates \
 COPY . $SRC_PATH
 
 RUN cd $SRC_PATH \
-	&& ref="$(cat .git/HEAD | cut -d' ' -f2)" \
-	&& commit="$(cat .git/$ref | head -c 7)" \
+	&& ref=$(cat .git/HEAD | grep ref | cut -d' ' -f2) \
+	&& commit=$(if [ -z "$ref" ]; then cat .git/HEAD; else cat ".git/$ref"; fi | head -c 7) \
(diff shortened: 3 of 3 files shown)


GitCop:
There were the following issues with your Pull Request

 * Commit: 403449092c87fda2db3faf5452cec016ecbe54e1
  * Invalid signoff. Commit message must end with  
    **License: MIT  
Signed-off-by: .* <.*>**

 * Commit: d36bc3f1710bfc158afa4fdf65f9d1b1dfe97e74
  * Invalid signoff. Commit message must end with  
    **License: MIT  
Signed-off-by: .* <.*>**

We ask for a few features in the commit message for Open Source licensing hygiene and commit message clarity.
`git commit --amend` can often help you quickly improve the commit message.
[Guidelines](https://github.com/ipfs/community/blob/master/contribution-guidelines.md#commit-messages) and [a script](https://github.com/ipfs/community/blob/master/dev/tools/hooks/setup_commit_msg_hook.sh) are available to help in the long run.
Your feedback on GitCop is welcome on [this issue](https://github.com/ipfs/infrastructure/issues/23).

---
This message was auto-generated by https://gitcop.com

lgierth:
Fixed the signoff

whyrusleeping:
is this related to #2453 at all?

lgierth:
Oh hrm yes very much actually. And failing tests.

Let me take a look (at both).

Kubuxu:
I don't know docker that much but LGTM.

lgierth:
ok now this one is fixed too

Dockerfile
```diff
@@ -37,6 +37,7 @@ ENV SRC_PATH   /go/src/github.com/ipfs/go-ipfs
 # Get the go-ipfs sourcecode
 COPY . $SRC_PATH
 
+RUN apk add --update git
```
lgierth: Oops, pushing one more update to remove this stray line

Dockerfile
```diff
@@ -53,8 +53,8 @@ RUN apk add --update musl go=$GO_VERSION git bash wget ca-certificates \
 	# We get the current commit using this hack,
 	# so that we don't have to copy all of .git/ into the build context.
 	# This saves us quite a bit of image size.
-	&& ref="$(cat .git/HEAD | cut -d' ' -f2)" \
-	&& commit="$(cat .git/$ref | head -c 7)" \
+	&& ref=$(cat .git/HEAD | grep ref | cut -d' ' -f2) \
+	&& commit=$(if [ -z "$ref" ]; then cat .git/HEAD; else cat ".git/$ref"; fi | head -c 7) \
```
chriscool: To get the current commit it's better to use `git rev-parse --short HEAD` and to get the current branch `git symbolic-ref --short HEAD 2>/dev/null || echo "$commit"` (where $commit is the current commit).

test/sharness/t0300-docker-image.sh
```diff
@@ -66,6 +66,14 @@ test_expect_success "simple ipfs add/cat can be run in docker container" '
 	test_cmp expected actual
 '
 
+test_expect_success "version CurrentCommit is set" '
+	docker_exec "$DOC_ID" "wget --retry-connrefused --waitretry=1 --timeout=30 -t 30 \
+		-q -O - http://localhost:8080/version" | grep Commit | cut -d" " -f2 >actual &&
+	docker_exec "$DOC_ID" "ipfs version --commit" | cut -d- -f2 >expected &&
+	[ "$(cat expected | wc -c)" -gt "1" ] && # check there actually is a commit set
```
chriscool: `test -s expected` is better than `[ "$(cat expected | wc -c)" -gt "1" ]`.

Dockerfile
lgierth: Hey thanks for wighing in! -- do you happen to know if `rev-parse` and `symbolic-ref` work without an actual repo, i.e. with only `.git/HEAD` and `.git/refs` or something similar present? The idea with this hack is that we don't have to copy all of `.git`, and thus gain savings in image size.

test/sharness/t0300-docker-image.sh
lgierth: Good suggestion

Dockerfile
chriscool: You need the following for `rev-parse` and `symbolic-ref` to work:

- `.git/HEAD`
- `.git/refs/`
- `.git/objects/` which can be empty

See: https://github.com/git/git/blob/master/setup.c#L266

So it might be a good idea to just `mkdir .git/objects/` so that you can use git commands afterwards.

It could be also a good idea to copy `.git/packed-refs` if it exists in case the refs are packed.

test/sharness/t0300-docker-image.sh
jbenet: i dont think this was ever improved

Dockerfile
jbenet: i think this is working right? did we want the improvements @chriscool proposed?

test/sharness/t0300-docker-image.sh
jbenet: Sorry, nevermind, it was: https://github.com/ipfs/go-ipfs/commit/0eed330  👍

Dockerfile
jbenet: Sorry, nevermind, already improved: https://github.com/ipfs/go-ipfs/commit/0eed330  👍

Issue #2722, opened 2016-05-17
Migrating away from Godeps


Kubuxu:
Many packages in Godeps are our own (ipfs, whyrusleeping) so we should easily be able to push the to gx, others are external but we also can push them to gx.

Question is: do we want to do that, if yes what should be the time frame as it will be quite a bit of work. There are, as for today, 54 packages in Godeps.

Kubuxu:
Dependencies still inside Godpes:
* [ ] bazil.org/fuse
* [ ] bitbucket.org/ww/goautoneg
* [ ] github.com/bren2010/proquint
* [ ] github.com/briantigerchow/pubsub
* [ ] github.com/camlistore/lock
* [ ] github.com/cenkalti/backoff
* [ ] github.com/cheggaaa/pb
* [ ] github.com/codahale/hdrhistogram
* [ ] github.com/codahale/metrics
* [ ] github.com/cryptix/mdns
* [ ] github.com/docker/spdystream
* [ ] github.com/dustin/go-humanize
* [ ] github.com/dustin/randbo
* [ ] github.com/facebookgo/atomicfile
* [ ] github.com/gogo/protobuf/io
* [ ] github.com/gogo/protobuf/proto
* [ ] github.com/hashicorp/golang-lru
* [ ] github.com/ipfs/go-datastore
* [ ] github.com/jbenet/go-base58
* [ ] github.com/jbenet/go-context/frac
* [ ] github.com/jbenet/go-context/io
* [ ] github.com/jbenet/go-detect-race
* [ ] github.com/jbenet/go-fuse-version
* [ ] github.com/jbenet/go-is-domain
* [ ] github.com/jbenet/go-msgio
* [ ] github.com/jbenet/go-multiaddr
* [ ] github.com/jbenet/go-multiaddr-net
* [ ] github.com/jbenet/go-multihash
* [ ] github.com/jbenet/go-os-rename
* [ ] github.com/jbenet/go-peerstream
* [ ] github.com/jbenet/go-random
* [ ] github.com/jbenet/go-random-files
* [ ] github.com/matttproud/golang_protobuf_extensions/pbutil
* [ ] github.com/mitchellh/go-homedir
* [ ] github.com/mtchavez/jenkins
* [ ] github.com/olekukonko/ts
* [ ] github.com/rs/cors
* [ ] github.com/satori/go.uuid
* [ ] github.com/steakknife/hamming
* [ ] github.com/syndtr/goleveldb/leveldb
[...]

PR #2737, opened 2016-05-18
Move backoff to gx from Godeps

Kubuxu:
Part of #2722
License: MIT


Kubuxu:
I will start throwing them in batches of about 5 into one PR because git can't merge JSON files.

PR #2744, opened 2016-05-19
Allow to specify where to put default options in option desciption

### LINKED ISSUE #2702: Default option command messes up for ipfs name publish
```
$ ipfs name publish --help 
...
OPTIONS

  --resolve           bool   - Resolve given path before publishing. Default: true.
  -t,      --lifetime string - Time duration that the record will be valid for.
  
      This accepts durations such as "300s", "1.5h" or "2h45m". Valid time units are
      "ns", "us" (or "µs"), "ms", "s", "m", "h".
  		. Default: 24h.
  --ttl               string - Time duration this record should be cached for (caution: experimental).
```

Look at the `. Default: 24h` string. That should be next to `valid for.`

Kubuxu:
Resolves #2702

License: MIT

### DIFF
--- a/commands/option.go
+++ b/commands/option.go
@@ -3,6 +3,7 @@ package commands
 import (
 	"fmt"
 	"reflect"
+	"strings"
 
 	"gx/ipfs/QmZNVWh8LLjAavuQ2JXuFmuYH3C11xo988vSgp7UQrTRj1/go-ipfs-util"
 )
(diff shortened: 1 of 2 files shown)


RichardLitt:
Good idea!

Kubuxu:
Should I change the tag to `<<default>>` we are using sometimes things like `<date>` in the descriptions. I hope there won't be `<default>` anywhere.

whyrusleeping:
This LGTM, good idea :)

jbenet:
> Should I change the tag to `<<default>>` we are using sometimes things like `<date>` in the descriptions. I hope there won't be `<default>` anywhere.

@Kubuxu valid concern. I'm not sure-- we may have `<default>` somewhere. if we _don't_ want it there, we should endeavor to make it obvious to devs somehow. right not it would just silently replace and i may not notice.

Kubuxu:
I think chaining it to`<<default>>` has lower cost than writing a document and remembering it. I will follow it up.

PR #2746, opened 2016-05-19
Use one verbose flag description

### LINKED ISSUE #2693: Standardising verbose flag description
Currently we are using many different ones:
* Print more verbose output.
* Write extra information.
* Write debug information.
* verbose output

It would be much nicer to use just one and the same description (when it makes sense, there might be some special cases).

If we standardise, what should be the description?

Kubuxu:
Resolves #2693

License: MIT


RichardLitt:
LGTM! RFM.

whyrusleeping:
LGTM

PR #2754, opened 2016-05-23
docker: simplify git ref resolution, thanks @chriscool

lgierth:
`git rev-parse` simply needed `.git/packed-refs` and an empty `.git/objects/` directory.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -50,11 +50,7 @@ RUN apk add --update musl go=$GO_VERSION git bash wget ca-certificates \
 	# Invoke gx
 	&& cd $SRC_PATH \
 	&& gx --verbose install --global \
-	# We get the current commit using this hack,
-	# so that we don't have to copy all of .git/ into the build context.
-	# This saves us quite a bit of image size.
-	&& ref=$(cat .git/HEAD | grep ref | cut -d' ' -f2) \
-	&& commit=$(if [ -z "$ref" ]; then cat .git/HEAD; else cat ".git/$ref"; fi | head -c 7) \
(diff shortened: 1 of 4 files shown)


Kubuxu:
LGTM

Dockerfile
```diff
@@ -50,11 +50,7 @@ RUN apk add --update musl go=$GO_VERSION git bash wget ca-certificates \
 	# Invoke gx
 	&& cd $SRC_PATH \
 	&& gx --verbose install --global \
-	# We get the current commit using this hack,
-	# so that we don't have to copy all of .git/ into the build context.
-	# This saves us quite a bit of image size.
-	&& ref=$(cat .git/HEAD | grep ref | cut -d' ' -f2) \
-	&& commit=$(if [ -z "$ref" ]; then cat .git/HEAD; else cat ".git/$ref"; fi | head -c 7) \
+	&& mkdir .git/objects && commit=$(git rev-parse HEAD | head -c 7) \
```
chriscool: Why don't you use `git rev-parse --short HEAD` instead of `git rev-parse HEAD | head -c 7`?

There is the following about `--short` in the git rev-parse documentation:

```
--short
--short=number

    Instead of outputting the full SHA-1 values of object names try to abbreviate them to a shorter unique name. When no length is specified 7 is used. The minimum length is 4.
```

so it does the right thing by default. It prints a 7 character long name. (See: https://git-scm.com/docs/git-rev-parse/2.8.3)

lgierth: Great point, thanks! Changed to --short.

chriscool:
It looks great to me now!

PR #2764, opened 2016-05-26
build: include commit hash & version

hutenosa:
This is a revived PR corresponding to #2453 
@lgierth do you have an idea how to deprecate properly? Currently both `--commit` and `--full` work, no warnings issued.
- Added Full Version: ipfs version --full
- Removed Commit from :8080/version
- ipfs version --commit and commit field in /api/v0/version are deprecated

License: MIT
Signed-off-by: hutenosa hutenosa@mm.st

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,6 +34,8 @@ ENV GOPATH     /go
 ENV PATH       /go/bin:$PATH
 ENV SRC_PATH   /go/src/github.com/ipfs/go-ipfs
 
+ARG IPFS_FULL_VERSION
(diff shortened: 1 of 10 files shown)


hutenosa:
Don't understand why gitcop complains...

chriscool:
The GitCop status is a new GitCop feature and it may be buggy.
It complained about a previous commit, but as it is not any more in this PR, the status should be ok now.

Kubuxu:
@lgierth what is the status on this?

PR #2005, opened 2015-11-24
Expose corenet on the command line

AtnNn:
This is an incomplete proof of concept. Is there any interest in adding this kind of feature to ipfs?

```
proteus$ ipfs service listen foo -- bc
```

```
nereid$ ipfs service dial -v QmNEpP... foo
Connected to: <peer.ID NEpP...>
1 + 1
2
```

I like using ipfs to store and share files, but I've been wondering if it could also be used to serve streams and dynamic content.

I saw there was some previous work in a similar direction in #1089. There is also [an example](https://github.com/ipfs/examples/tree/master/examples/api/service) for writing custom ipfs nodes in go.

PR #2765, opened 2016-05-27
Expose corenet API

magik6k:
This is continuation of #2005 , still WIP.
What it allows is to communicate in real-time using ipfs identities and libp2p. Currently the implementation uses websockets to communicate with daemon, and it implements channels over them, so the listener can handle more than 1 client at a time. Usage is still much like it was in #2005, but quite a few internals changed. I managed to set-up TAP tunnel over IPFS between 2 hosts using little helper program, so I think I can say that it at least works.
What needs to be done:

- [ ] Get sign-off under @AtnNn commit from #2005
- [x] Clean up commit mess
- [ ] Implement/fix to closing connections
- [ ] Implement/fix graceful shutdown
- [ ] Cleanup code in few places
- [ ] Write tests
- [ ] Write docs

I'm opening this now so that people can see the progress and comment on what is done.

### DIFF
--- a/cmd/ipfs/main.go
+++ b/cmd/ipfs/main.go
@@ -343,7 +343,7 @@ func callCommand(ctx context.Context, req cmds.Request, root *cmds.Command, cmd 
 		}
 
 		// Okay!!!!! NOW we can call the command.
-		res = root.Call(req)
+		res = root.Call(req, os.Stdout, os.Stderr)
 
 	}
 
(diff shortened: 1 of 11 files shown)


lgierth:
I haven't looked at the rest yet, but for API communication, @kubuxu's unix socket proposal is very relevant: ipfs/notes#129

whyrusleeping:
Yeah, this change is going to be (at least) two parts, the first being making the API transport more flexible, and the second is the actual exposing of the corenet stuff. lets start working towards part 1, sound good @Kubuxu ?

whyrusleeping:
@magik6k can this be closed in favor of #3128 ?

magik6k:
Yep, the other one implements the same thing.

PR #2772, opened 2016-05-31
update libp2p to v3.2.3

whyrusleeping:
This contains the ping leak fix and also updates to the official yamux hang fix code

License: MIT


Kubuxu:
SGTM, we really need `gx diff`.

whyrusleeping:
@Kubuxu if you tell me how it should work, i'll write it.

Kubuxu:
It would fetch two packages by hashes and the allow you to recursively follow changes in dependencies.
So:
```
> gx diff Qm...AAA Qm...BBB
Changes in package go-libp2p between Qm...AAA and Qm...BBB:
Version changed: 1.0.0 -> 1.0.1
Dependencies changed are:
  1. go-net - 1.0.2 -> 1.0.3 (Qm...CCC -> Qm...DDD)
Type number to show diff between packages (0 is current package). Press `q` or Ctrl-C to exit.
```
Typing in `1` would really call: `gx diff Qm...CCC -> Qm...DDD`. `0` would show git style diff between versions (it doesn't have to be exactly the same).

Issue #2194, opened 2016-01-13
Add test to t0110 to check that HEAD requests do not return a body


whyrusleeping:
ref: https://github.com/ipfs/go-ipfs/pull/2143#discussion_r49589501

PR #2775, opened 2016-05-31
Add check for HEAD response to be empty

Kubuxu:
Resolves https://github.com/ipfs/go-ipfs/issues/2194

License: MIT

### DIFF
--- a/test/sharness/t0110-gateway.sh
+++ b/test/sharness/t0110-gateway.sh
@@ -116,6 +116,11 @@ test_expect_success "output looks good" '
 	grep "Content-Type: text/html" indexout
 '
 
+test_expect_success "HEAD 'index.html' has no content" '
+	curl -X HEAD --max-time 1 http://127.0.0.1:$port/ipfs/$INDEXHASH/ > output;
+	[ ! -s output ]
+'
+
 # test ipfs readonly api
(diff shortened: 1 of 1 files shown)


test/sharness/t0110-gateway.sh
```diff
@@ -116,6 +116,11 @@ test_expect_success "output looks good" '
 	grep "Content-Type: text/html" indexout
 '
 
+test_expect_success "HEAD 'index.html' has no content" '
+	curl -X HEAD --max-time 1 http://127.0.0.1:$port/ipfs/$INDEXHASH/ > output;
```
whyrusleeping: should have an `&&` between these. otherwise this will pass if the curl command fails

Kubuxu: The curl will fail as for the timeout.

whyrusleeping: why not use `--head` then?

Kubuxu: Because this test is to check if head request has no content. `--head` won't return the content even if it is there.

whyrusleeping: Ah, makes sense. Carry on!

PR #2778, opened 2016-06-01
Add CORS headers to Read Only Gateway Default config

Kubuxu:
This will allow for weider use of local gateway from different sites.

For example: ipfs.pics could use JS to use either their global or user's local gateway.

I might work on piece of JS doing just that but I don't really like JS. :smile: 

License: MIT

### DIFF
--- a/repo/config/init.go
+++ b/repo/config/init.go
@@ -62,6 +62,11 @@ func Init(out io.Writer, nBitsForKeypair int) (*Config, error) {
 			RootRedirect: "",
 			Writable:     false,
 			PathPrefixes: []string{},
+			HTTPHeaders: map[string][]string{
+				"Access-Control-Allow-Origin":  []string{"*"},
+				"Access-Control-Allow-Methods": []string{"GET"},
+				"Access-Control-Allow-Headers": []string{"X-Requested-With"},
+			},
 		},
 	}
 
--- a/test/sharness/t0112-gateway-cors.sh
+++ b/test/sharness/t0112-gateway-cors.sh
@@ -7,10 +7,6 @@
 test_description="Test HTTP Gateway CORS Support"
 
 test_config_ipfs_cors_headers() {
-    ipfs config --json Gateway.HTTPHeaders.Access-Control-Allow-Origin '["*"]'
-    ipfs config --json Gateway.HTTPHeaders.Access-Control-Allow-Methods '["PUT", "GET", "POST"]'
-    ipfs config --json Gateway.HTTPHeaders.Access-Control-Allow-Headers '["X-Requested-With"]'
-
     ipfs config --json API.HTTPHeaders.Access-Control-Allow-Origin '["*"]'
     ipfs config --json API.HTTPHeaders.Access-Control-Allow-Methods '["PUT", "GET", "POST"]'
     ipfs config --json API.HTTPHeaders.Access-Control-Allow-Headers '["X-Requested-With"]'


whyrusleeping:
I like this idea, but people more knowledgable on http should review this, cc @diasdavid @travisperson @dignifiedquire

dignifiedquire:
I would suggest adding a test using curl to ensure it works as expected. See here for some details http://stackoverflow.com/questions/12173990/how-can-you-debug-a-cors-request-with-curl

Kubuxu:
We have tests for CORS headers, this just changes the default.

https://github.com/ipfs/go-ipfs/blob/master/test/sharness/t0112-gateway-cors.sh

dignifiedquire:
then you should add a test without changing the config so we know the default is working as epxected

Kubuxu:
I've modified the test.

lgierth:
This LGTM, sorry for the long wait

lgierth:
Also deployed to gateways

jbenet:
thanks for pushing for that test @dignifiedquire 👍

jbenet:
Was that test actually created? @Kubuxu ?

Kubuxu:
The test case is the fact that we don't have to explicitly set those settings in sharness for tests to succeed. See removed config calls.

jbenet:
This is not a specific test case. if this fails, it will take a long time to debug because of those other things failing not being super clear. remember that people who are not you will be hit by this. there should be test case that tests this as the first one to fail so debugging + fixing is easy.

WardCunningham:
Thank you for this.

lgierth:
@WardCunningham this change is part of v0.4.3-rc4: https://dist.ipfs.io/go-ipfs/v0.4.3-rc4

Mithgol:
I have upgraded to go-ipfs version 0.4.5 and then I still don't see `Access-Control-Allow-Origin: "*"` in response headers of (for example) http://127.0.0.1:8080/ipfs/QmYZUa2E1bsuN37fwzjcawop4wk1kdXxzRyiaGfjKfWM2T though I see such header at https://ipfs.io/ipfs/QmYZUa2E1bsuN37fwzjcawop4wk1kdXxzRyiaGfjKfWM2T without any hindrance.

Does it mean that changes in the default config don't automatically apply to earlier adopters of go-ipfs?

The chat log from fedwiki/wiki-client#173 seems to say they don't:

(the chat log)

Then what exactly should I do myself (and recommend to the other IPFS users) to make the necessary changes in the local configuration and enable CORS?

Kubuxu:
You can run:
```
ipfs config --json Gateway.HTTPHeaders.Access-Control-Allow-Origin '["*"]'
ipfs config --json Gateway.HTTPHeaders.Access-Control-Allow-Methods '["GET"]'
ipfs config --json Gateway.HTTPHeaders.Access-Control-Allow-Headers '["X-Requested-With"]'
````

**Comment by Mithgol** [comment]
Thanks, it has worked.

For future references, on Windows the equivalent commands are

```
ipfs config --json Gateway.HTTPHeaders.Access-Control-Allow-Origin [\"*\"]
ipfs config --json Gateway.HTTPHeaders.Access-Control-Allow-Methods [\"GET\"]
ipfs config --json Gateway.HTTPHeaders.Access-Control-Allow-Headers [\"X-Requested-With\"]
```

lgierth:
I still think we should add these to the default config, as suggested here

Kubuxu:
It is in default config.

Issue #2770, opened 2016-05-30
create issue template 


whyrusleeping:
Lets make an issue template for go-ipfs issues and move towards having issues filed here being solely bugs and directly actionable things. All others should be directed to ipfs/support or ipfs/faq

Kubuxu:
Or ipfs/notes for ideas.

chriscool:
So where should issue go? :-)

Ok, as it is actionable, it could probably stay here. But to be really actionable, we should probably state what we want in the issue template.
 
I suggest we ask for the commit version (from `ipfs version --commit`) and a way to reproduce the problem. Maybe we could ask for the repo version (from `ipfs version --repo`) too?
What else?

whyrusleeping:
I think also asking for operating system and platform would be good info to have as well. There are a lot of issues where it would have saved a lot of time to know that the user was on windows the whole time.

chriscool:
Yeah sure!

whyrusleeping:
relevant info: https://github.com/blog/2111-issue-and-pull-request-templates

whyrusleeping:
we now have an issue template!

PR #2786, opened 2016-06-01
Add ISSUE_TEMPLATE.md

chriscool:
This addresses issue #2770.

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,0 +1,18 @@
+Version/Platform/Processor information (from `ipfs version --all`):
+
+
+
+Type (bug, feature, meta, test failure, question):
+Area (api, commands, daemon, fuse, etc):
+Priority (from P0: functioning, to P4: operations on fire):
+
+Description:
+
+
+
+
+---------------------------------------------------
+This is for you! Please read, and then delete this text before posting it.
+The go-ipfs issues are only for bug reports and directly actionable features.
+Check https://github.com/ipfs/community/blob/master/contributing.md#reporting-issues if that doesn't fit.
+Check https://github.com/ipfs/go-ipfs/blob/master/contribute.md#how-to-report-an-issue if you are not sure how to fill this issue.


Kubuxu:
In my opinion we shouldn't be using headers to note informations. It will get a bit annoying.

Also there was talk about adding note at the bottom that is information about that go-ipfs repo is about bug reports and directly actionable features. More complex feature requests and general questions should go into different repos.

chriscool:
There are no headers any more. About the note at the bottom, I think it could get annoying too. Isn't there another place where we could tell that?

Kubuxu:
No, it is the best we have came up with (as a fix for big number of non-actionable issues in the repo). I would place it at the bottom with something like:
```
---------------------------------------------------
Please read and then delete
go-ipfs issue tracker ....
If that doesn't fit your issue you might want to use one of following repos:
https://github.com/ipfs/notes/isssues - for general notes and ideas
https://github.com/ipfs/faq/issues - for questions
https://github.com/ipfs/specs/issues - for protocol discussions
etc.
```

chriscool:
Ok, here is a version close to what you suggest.

It seems a bit verbose to me though. Couldn't we just provide one link from where the user could get more information about the appropriate places?

Kubuxu:
This might be a good idea. A central hub to redirect peoples' issues (also this `etc.` isn't needed in the text itself).

Kubuxu:
Yeah, this section needs improvements as well.

I would go with what we've got here (thanks) and possibly improve it in future.

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,16 @@
+Operating system version & Platform:
+IPFS commit version (from `ipfs version --commit`):
+IPFS repo version (from `ipfs version --repo`):
```
whyrusleeping: I don't think we need the repo version, if we know the commit, or even just the general semver we can infer it. It might be better to ask for 'ipfs version (and commit, if known)'

whyrusleeping:
cc @em-ly

em-ly:
Thanks @whyrusleeping. We met today and created some prelimary template outline for the Issues template for https://github.com/ipfs/go-ipfs/

**Type:**
    bug
    Meta
    question
    test failure
    panic (severe bug)
    enhancement 

**Platform:**
    Linux
    Windows
    Mac
    Etc.

**Processor:**
    32
    64
    Arm

**Area:**
    api
    bandwidth reduction
    bit swap
    blockstore
    commands
    containers + vms
    core
    daemon + init
    dat
    discovery
    encryption
    files
    fuse
    gateway
    interior
    pins
    libp2p
    merkledag
    nat
    releases
    repo
    routing
    tools
    tracking
    unix vs dag

**Priority:**
    P0 - Operations Functioning
    P1 - Operations Sort of Functioning
    P2 - Operations Not Functioning
    P3 - Operations Foobar (ciritical bug affecting master)
    P4 - Operations on Fire (critical bug affecting release)

Anything else need to be included?

chriscool:
With PR #2790 the versions/platform/processor can be dealt with using `ipfs version --all`.

@em-ly I am not sure how it is possible, without being very verbose, to ask someone reporting an issue to select the Type, Area and Priority as per the above in a issue template.

Or do you mean that we should have something like:

```


Type (bug, feature, meta, test failure, question):
Area (api, commands, daemon, fuse, etc):
Priority (from P0: functioning, to P4: operations on fire):

Description:


---------------------------------------------------
Please read and then delete
This is for is for bug reports and directly actionable features only.
Check [where to report an issue](link) if that doesn't fit.
Check [how to report an issue](link) if you are not sure how to fill this issue. 
```

and the "how to report an issue" linked document would show all the possible values for the Type, Area and Priority fields?

Kubuxu:
@chriscool for me it looks great. Only think is that we don't (and shouldn't) use markdown formatting for links as people will see the template mostly in the raw form.

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,16 @@
...
+
+Description:
+
+
+
+---------------------------------------------------
+Please read and then delete
```
RichardLitt: "This is for you! Please read, and then delete this text before posting it."

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,16 @@
...
+go-ipfs issue tracker is for bug reports and directly actionable features only.
+If that doesn't fit your issue you might want to use one of following repos:
```
RichardLitt: If that doesn't fit your issue, you might want...

RichardLitt:
What about linking to support? What is ipfs/support for, then?

chriscool:
@RichardLitt @em-ly and @Kubuxu here is a new version that tries to take into account what you suggested.

Anyway before this can be merged we need at least the following:

- [] improve https://github.com/ipfs/community/blob/master/contributing.md#reporting-issues so that it shows where different kinds of problem should be reported,
- [] write https://github.com/ipfs/go-ipfs/blob/master/contribute.md#how-to-report-an-issue or something like that with all the options and their meaning @em-ly posted above,
- [] merge PR #2790 to get `ipfs version --all`.

And yeah perhaps we could also add a link to ipfs/support too.

RichardLitt:
Made this: https://github.com/ipfs/community/pull/128

Agreed about making a "How to report an issue" for go-ipfs specifically. Happy to do that.

whyrusleeping:
The information should be best effort by the user, we can correct and fill stuff in post-filing as we see fit

chriscool:
@RichardLitt thanks for ipfs/community#128!
And yeah, it looks like the only thing left before we can merge this is making a "How to report an issue" for go-ipfs specifically using what @em-ly posted above. Thanks!

RichardLitt:
I'm off hiking for a week, but happy to do this when I get back.

RichardLitt:
And back. Will be addressing this this week.

whyrusleeping:
@RichardLitt @em-ly update here?

em-ly:
@RichardLitt is going to work on getting this completed for this week!
Thanks for bumping the issue.

On Tue, Jun 21, 2016 at 12:57 PM, Jeromy Johnson <notifications@github.com>
wrote:

> @RichardLitt <https://github.com/RichardLitt> @em-ly
> <https://github.com/em-ly> update here?
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/ipfs/go-ipfs/pull/2786#issuecomment-227503092>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/ASplmmRBWMdCcvLhZRDSDLqFO6B-l_s3ks5qOBflgaJpZM4Ir26U>
> .
>



-- 
Emily Moore
Project Manager
Protocol Labs, Inc.

RichardLitt:
Created https://github.com/ipfs/go-ipfs/pull/2889. This should be good to go, now?

chriscool:
Yeah, when PR #2889 is merged, I will update this and it should be good to go.

Kubuxu:
I would do something like:

-----------
Version information:
```
please run and copy here the result of: ipfs version -all
```

whyrusleeping:
#2889 is merged!

RichardLitt:
LGTM. Let's try it!

whyrusleeping:
is this RFM?

em-ly:
LGTM! Thanks for all the work on this @chriscool @RichardLitt and @Kubuxu! Let's roll this out and test it throughout the week, edits made as necessary.

whyrusleeping:
Feedback time, I think we need to use something much more minimal, there is too much text in the template and its a bit overwhelming to users who most of the time won't have any idea what half of it means. I propose something closer to:

```
[//]: # (Output From `ipfs version --all`)
- Version Information: 

[//]: # ( Bug, Feature, Enhancement, Etc )
- Type:

[//]: # ( from `P0` "Critical" to `P5` "Relatively Unimportant")
- Priority :

### Description:
```

Using a bit of a hack for markdown comments that the user will see when creating the issue, but that wont render when the issue itself is viewed.

RichardLitt:
[//]: # ( from `P0` "Critical" to `P5` "Relatively Unimportant")
- Priority :
 Just testing.

dignifiedquire:
I just discovered you can use html comments in these, take a look at the template from nodejs/node: https://raw.githubusercontent.com/nodejs/node/master/.github/ISSUE_TEMPLATE.md

These templates also seem to be quite useful and working

- https://github.com/webpack/karma-webpack/tree/master/.github
- https://github.com/karma-runner/karma/blob/master/ISSUE_TEMPLATE.md

whyrusleeping:
@dignifiedquire oooooh, good find! I like that a little better than the empty link hovertext hack i'm using

RichardLitt:
Also worth taking into account: the repos in this repository: https://github.com/devspace/awesome-github-templates

jbenet:
The current issue template is a bit cumbersome. maybe it's just how the formatting works. consider these that @RichardLitt posted: https://github.com/devspace/awesome-github-templates

PR #2788, opened 2016-06-01
Fix/buffer err chan

whyrusleeping:
If one of the other calls failed, it would return early leaving the other goroutine stranded waiting to return its error.

I also localized the context cancellation to more quickly clean up operations in failure situations.


Kubuxu:
A bit late but, LGTM

PR #2790, opened 2016-06-01
Add version --all option

### LINKED ISSUE #2789: Add `ipfs version --all` to report all version information at once
To make it easier for people to report information about ipfs when describing a bug, it could be nice to have them type only one command instead of `ipfs version --commit` and `ipfs version --repo`.

We could also perhaps make it print the current OS and platform while at it and perhaps the go version used to compile the binary too (not sure if it is possible).

This is linked to PR #2786.

Kubuxu:
Resolves #2789

License: MIT

### DIFF
--- a/core/commands/version.go
+++ b/core/commands/version.go
@@ -3,6 +3,7 @@ package commands
 import (
 	"fmt"
 	"io"
+	"runtime"
 	"strings"
 
 	cmds "github.com/ipfs/go-ipfs/commands"
@@ -14,6 +15,8 @@ type VersionOutput struct {
 	Version string
 	Commit  string
 	Repo    string
+	System  string
+	Golang  string
 }
 
 var VersionCmd = &cmds.Command{
@@ -26,12 +29,15 @@ var VersionCmd = &cmds.Command{
 		cmds.BoolOption("number", "n", "Only show the version number.").Default(false),
 		cmds.BoolOption("commit", "Show the commit hash.").Default(false),
 		cmds.BoolOption("repo", "Show repo version.").Default(false),
+		cmds.BoolOption("all", "Show all version information").Default(false),
 	},
 	Run: func(req cmds.Request, res cmds.Response) {
 		res.SetOutput(&VersionOutput{
 			Version: config.CurrentVersionNumber,
 			Commit:  config.CurrentCommit,
 			Repo:    fsrepo.RepoVersion,
+			System:  runtime.GOARCH + "/" + runtime.GOOS, //TODO: Precise version here
+			Golang:  runtime.Version(),
 		})
 	},
 	Marshalers: cmds.MarshalerMap{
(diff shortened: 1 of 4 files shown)


chriscool:
This is great and works for me:

```
> ipfs version --all
go-ipfs version: 0.4.2-0bcfc49
Repo version: 3
System version: amd64/linux
Golang version: go1.5.3
```

Maybe you could add a sharness test?

Thanks!

RichardLitt:
Works nicely on the API. 

```
🐕  curl -i "http://localhost:5001/api/v0/version?all"
HTTP/1.1 200 OK
Access-Control-Allow-Headers: X-Stream-Output, X-Chunked-Output, X-Content-Length
Access-Control-Expose-Headers: X-Stream-Output, X-Chunked-Output, X-Content-Length
Content-Type: application/json
Server: go-ipfs/0.4.2
Trailer: X-Stream-Error
Vary: Origin
Date: Thu, 02 Jun 2016 09:04:06 GMT
Transfer-Encoding: chunked

{"Version":"0.4.2","Commit":"0bcfc49","Repo":"3","System":"amd64/darwin","Golang":"go1.5.3"}
```

Kubuxu:
@RichardLitt all `version` calls will return those values as all options for `version` command are text formatting only.

I will add few tests later.

RichardLitt:
Ah right. :)

Kubuxu:
Added tests also moved the regression tests further apart number wise.

chriscool:
LGTM, thanks!

whyrusleeping:
LGTM

ligi:
just thinking: isn't this changing the API? 
currently curl "http://localhost:5001/api/v0/version?all"                    
returns `{"Version":"0.4.2","Commit":"1654bbf","Repo":"3"}`

afterwards it will return ( with the same api version ) 
`{"Version":"0.4.2","Commit":"0bcfc49","Repo":"3","System":"amd64/darwin","Golang":"go1.5.3"}`

Kubuxu:
This isn't even breaking change (API is just extended).
Also it is v0 version of the API it is and will be modified. I think if we were to version the API at this stage we would go with v01, v02 and so on, but it isn't planned ye t.

refs local changes the API but as it can be seen nobody was previously using it as they would report its incorrect behaviour.

ligi:
the java-api uses refs/local:
https://github.com/ipfs/java-ipfs-api/blob/master/src/main/java/org/ipfs/api/IPFS.java#L104

jbenet:
> ... changes the API but as it can be seen nobody was previously using it as they would report its incorrect behaviour.

Never assume that. Ever.

jbenet:
@ligi thank you very much for flagging this and #2803 as breaking changes, it got me to take a closer look at them (https://github.com/ipfs/go-ipfs/pull/2812#issuecomment-242672559)

Kubuxu:
Is adding fields to JSON structure a breaking change?

PR #2634, opened 2016-05-06
[WIP] Filestore Implementation

kevina:
Closes Issue #875, avoid duplicating files added to ipfs

**NOT READY FOR MERGE**

Rebased #2600 on master.

Quicklinks: [Code](https://github.com/ipfs-filestore/go-ipfs), [README](https://github.com/ipfs-filestore/go-ipfs/blob/master/filestore/README.md), 

TODO to get this merged:
- [ ] Rebase on master turning the chain of commits into a reasonable ChangeSet
- [ ] Agree on major infrastructure changes
  - [ ] Multi-blockstore (#3119)
- [x] Separate out non-filestore bits of infrastructure change into there own pull request
  - Multi-blockstore: #3257
  - DAGService.GetLinks: ~~#3255~~
  - Others: see milestone [Filestore implementation](https://github.com/ipfs/go-ipfs/milestone/26)
- [ ] Code review
- [ ] Merge

Note The filestore is very basic right now, but it is functional.  I will likely continue to improve and submit new pull requests for the enhanced functionally but right now I fell it is important a basic implementation gets in so that it will get used, it can be labeled as an experimental feature and disabled by default, but available for those that want to use it.  I consider the code production ready.

Resolves #875

PR #2792, opened 2016-06-01
Stable bits from Pull Request #2634

kevina:
Stable bits from Pull Request #2634.

### DIFF
--- a/blocks/blocks.go
+++ b/blocks/blocks.go
@@ -20,47 +20,47 @@ type Block interface {
 }
 
 // Block is a singular block of data in ipfs
-type RawBlock struct {
+type BasicBlock struct {
 	multihash mh.Multihash
 	data      []byte
 }
 
 // NewBlock creates a Block object from opaque data. It will hash the data.
-func NewBlock(data []byte) *RawBlock {
-	return &RawBlock{data: data, multihash: u.Hash(data)}
+func NewBlock(data []byte) *BasicBlock {
+	return &BasicBlock{data: data, multihash: u.Hash(data)}
 }
 
 // NewBlockWithHash creates a new block when the hash of the data
 // is already known, this is used to save time in situations where
 // we are able to be confident that the data is correct
-func NewBlockWithHash(data []byte, h mh.Multihash) (*RawBlock, error) {
+func NewBlockWithHash(data []byte, h mh.Multihash) (*BasicBlock, error) {
 	if u.Debug {
 		chk := u.Hash(data)
 		if string(chk) != string(h) {
 			return nil, errors.New("Data did not match given hash!")
 		}
 	}
-	return &RawBlock{data: data, multihash: h}, nil
+	return &BasicBlock{data: data, multihash: h}, nil
 }
 
-func (b *RawBlock) Multihash() mh.Multihash {
+func (b *BasicBlock) Multihash() mh.Multihash {
 	return b.multihash
 }
 
-func (b *RawBlock) Data() []byte {
+func (b *BasicBlock) Data() []byte {
 	return b.data
 }
 
 // Key returns the block's Multihash as a Key value.
-func (b *RawBlock) Key() key.Key {
+func (b *BasicBlock) Key() key.Key {
 	return key.Key(b.multihash)
 }
 
-func (b *RawBlock) String() string {
+func (b *BasicBlock) String() string {
 	return fmt.Sprintf("[Block %s]", b.Key())
 }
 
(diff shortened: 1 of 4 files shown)


kevina:
I think continuous-integration/travis-ci/pr was getting confused as I committed two commits shortly after one another and it was attempting to build with the wrong commit id.  Commit  dbabcf9 does not touch `coreunix/add.go`.

kevina:
@whyrusleeping this is ready for you to have a look, and okay with me if it gets merged.

core/coreunix/add_test.go
```diff
@@ -54,7 +54,8 @@ func TestAddGCLive(t *testing.T) {
 
 	errs := make(chan error)
 	out := make(chan interface{})
-	adder, err := NewAdder(context.Background(), node, out)
+	adder, err := NewAdder(context.Background(), node.Pinning, node.Blockstore, node.DAG)
+	adder.Out = out
```
whyrusleeping: just in case (even though its a test), i'd put the `adder.Out = out` after the error check. If this call does error for some reason, its going to return a nil adder and we will panic on this line and lose information about what the real error was

whyrusleeping:
One nitpick then LGTM

core/coreunix/add_test.go
kevina: Oops.  Your right fixed.

kevina:
@whyrusleeping,okay I fixed the nitpick and all tests are passing.

whyrusleeping:
sweet, LGTM

blocks/blocks.go
```diff
@@ -20,47 +20,47 @@ type Block interface {
 }
 
 // Block is a singular block of data in ipfs
-type RawBlock struct {
+type BasicBlock struct {
```
jbenet: Curious why the name change?

core/coreunix/add.go
```diff
@@ -67,42 +67,46 @@ type AddedObject struct {
 	Bytes int64  `json:",omitempty"`
 }
 
-func NewAdder(ctx context.Context, n *core.IpfsNode, out chan interface{}) (*Adder, error) {
-	mr, err := mfs.NewRoot(ctx, n.DAG, newDirNode(), nil)
+func NewAdder(ctx context.Context, p pin.Pinner, bs bstore.GCBlockstore, ds dag.DAGService) (*Adder, error) {
```
jbenet: curious why the interface change? 

I imagine that this is nicer for users:

```go
node := mkipfsNode(...)
add.NewAdder(ctx, node, out)
```
Was there a case where you didnt have a node, but you had everything else?

jbenet: Also, what happens if users do not set `fileadder.Out` ? does it hang forever sending on a nil channel? that seems error prone to me.

blocks/blocks.go
kevina: Note, that I originally named this RawBlock in commit b84cbec2.  Both commits where factored out of #2634.  The name `RawBlock` was only a temporary name, I meant to squash this in b84cbec2 but messed up the rebase and didn't notice until after it was merged.

core/coreunix/add.go
```diff
@@ -306,7 +310,7 @@ func AddR(n *core.IpfsNode, root string) (key string, err error) {
 // the directory, and and error if any.
 func AddWrapped(n *core.IpfsNode, r io.Reader, filename string) (string, *dag.Node, error) {
 	file := files.NewReaderFile(filename, filename, ioutil.NopCloser(r), nil)
-	fileAdder, err := NewAdder(n.Context(), n, nil)
+	fileAdder, err := NewAdder(n.Context(), n.Pinning, n.Blockstore, n.DAG)
 	if err != nil {
```
kevina: @jbenet, note that nil was originally passed in to NewAdder here.

core/coreunix/add.go
whyrusleeping: @jbenet this was changed because passing in a node gives us very little control over how the add gets done. Using a different blockstore or temporary dagservice for different applications is becoming fairly common.

whyrusleeping: @jbenet also, passing `nil` is just fine.

kevina: Yes there was a case when I didn't have a node but everthing else. This was discussed with @whyrusleeping over IRC. 

There are checks in the code to check if fileadder.Out is nil and if so no attempt is written to the channel.

kevina: Also see https://github.com/ipfs/go-ipfs/pull/2634#issuecomment-222591310.

PR #2690, opened 2016-05-13
fix mfs cp/mv bug

noffle:
Addresses https://github.com/ipfs/go-ipfs/issues/2654

PR #2795, opened 2016-06-02
rework add-mfs to not use caching

whyrusleeping:
License: MIT


lgierth:
What's the symlink-related change in this commit?

noffle:
@lgierth: this PR is a cleaner stab at https://github.com/ipfs/go-ipfs/pull/2690, which has a bit more history on why symlinks are relevant to this. My understanding is that mfs doesn't support symlinks but since `ipfs add` uses mfs we need to support symlink caching to get the correct dag back.

whyrusleeping:
@lgierth its a workaround for the fact that mfs doesnt support symlinks, but we support ipfs add'ing symlinks. You can construct a tree with them in it, but you won't be able to use them via the files api

PR #2796, opened 2016-06-03
docker: make dockerfile differences more obvious

lgierth:
The two produce the same image, except one is optimized for image size, while the other is optimized for build time.


whyrusleeping:
LGTM! thanks :)

PR #2798, opened 2016-06-03
Make bitswap better

whyrusleeping:
Made a few changes to bitswap that significantly improve performance. By focusing on getting the cancel messages out to peers as fast as possible, and delaying sending blocks to peers we receive many cancels from we reduce the number of duplicate block receipts. I ran some tests on a network with 20ms latency between nodes (10ms per link) and have the results here: https://ipfs.io/ipfs/QmNyCqMQe1r3dQsAQEsncNqQGkF2YEyB5w6JRdNLRLDcox/smarter%20bitswap%20test%20info%20%2810ms%29.pdf

I can run more tests, but it takes a little while to do and these results are pretty good.

License: MIT

### DIFF
--- a/exchange/bitswap/decision/engine.go
+++ b/exchange/bitswap/decision/engine.go
@@ -3,6 +3,7 @@ package decision
 
 import (
 	"sync"
+	"time"
 
 	blocks "github.com/ipfs/go-ipfs/blocks"
 	bstore "github.com/ipfs/go-ipfs/blocks/blockstore"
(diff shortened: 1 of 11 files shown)


whyrusleeping:
thats really weird... somehow these changes make the tests in `test/integration` hang *after* the completed successfully. The defer'ed close calls seem to be getting stuck somewhere

whyrusleeping:
test data generated with: https://github.com/whyrusleeping/bitswap-test

Kubuxu:
LGTM, the first commit is failing tests so it is best to squash it in.

jbenet:
Would be great to have those tests runnable to create a graph locally, to test against master.

was it you @Kubuxu making some nice dup graphs? or maybe @ion1?

Kubuxu:
I don't think it was me.

PR #2801, opened 2016-06-03
Moved docs to docs

RichardLitt:
I moved debug-guide and releases to the docs folder. I think we should also move dev.md and roadmap.md there, too -- but I am not sure how often these files are updated, addressed, changed, or, in general, seen. Perhaps we should have a wider discussion about the docs included in this repo?

License: MIT


whyrusleeping:
LGTM

PR #2804, opened 2016-06-03
fix cleanup of empty provider sets

whyrusleeping:
This fixes a minor memory leak. When all providers for a given set would expire, the empty provider set object wouldnt get removed from the map it belongs to.

License: MIT


Kubuxu:
LGTM

PR #2808, opened 2016-06-05
Makefile, bin: Support multiple GOPATH components

karalabe:
Go's `GOPATH` environment variable can contain multiple paths, separated by a colon `:` similar to the `PATH` env var. In such cases Go uses the first path element to install any packages, however the remaining paths are still used to find dependencies. Because of this it's quite a common setup to have multiple paths listed in GOPATH (e.g. one for development and another for appengine, which has its special installer).

The current Makefile however erroneously assumes that `GOPATH` contains a single path element, and tries to locate go-ipfs within by appending a string to `GOPATH`, which is obviously cause an invalid path, and hence fail the `path_check` rule in the Makefile, even though the required condition is met.

~~This PR fixes it by moving path handling back into `bin/check_go_path`, which iterates over all the path components of `GOPATH`, trying each one of them for the go-ipfs match and only failing if none matches the requirements. Further it was kind of weird to pass in `PWD` as a parameter to a script, so I moved the first parameter's evaluation into the script as well.~~

This PR fixes it by splitting up GOPATH along the colon characters, appending the go-ipfs suffix to each component and resolving them individually. The possibly multiple resulting paths are passed to the path verification script to iterate over and check individually.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,7 +41,7 @@ bin/gx-go-v%:
 gx_check: ${gx_bin} ${gx-go_bin}
 
 path_check:
-	@bin/check_go_path $(realpath $(shell pwd)) $(realpath $(GOPATH)/src/github.com/ipfs/go-ipfs)
+	@bin/check_go_path $(realpath $(shell pwd)) $(realpath $(addsuffix /src/github.com/ipfs/go-ipfs,$(subst :, ,$(GOPATH))))
 
 deps: go_check gx_check path_check
 	${gx_bin} --verbose install --global
--- a/bin/check_go_path
+++ b/bin/check_go_path
@@ -1,7 +1,6 @@
 #!/bin/sh
 
 PWD=$1
-EXPECTED=$2
 
 if [ -z "$PWD" ]; then
 	echo "must pass in your current working directory"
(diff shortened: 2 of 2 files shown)


lgierth:
I'm generally cool with this change, but we should be aware it introduces a build dependency on realpath. The previous version used make's built-in realpath.

More thoughts?

karalabe:
Yup, just noticed it myself too. Will try to work around it.

karalabe:
@lgierth Moved `realpath` back into the Makefile and sorted out the path splitting, suffixing and real-resolution within make. The external script now just iterates over the individual paths and checks them without doing any operations. PTAL

Kubuxu:
You are right (just for tracking: https://github.com/ipfs/go-ipfs/issues/2809)

karalabe:
Yup, force pushed to rebuild and all tests are green now. :D

lgierth:
Cool you pulled it off :) I can't speak for your multiple-gopaths scenario, but I tested with my own symlinked-into-gopath scenario and it still works fine.

LGTM

whyrusleeping:
@chriscool and @Kubuxu could you CR this? i want to always double check changes that affect the build system

Kubuxu:
It looks good for me. I hope there isn't another GOPATH edge case waiting around the coroner :smile:.

whyrusleeping:
@Kubuxu my sentiments exactly, lol. Its looks good to me, but theres always something tricky going on. its hard to be sure

chriscool:
It LGTM too. Thanks @karalabe !

PR #2810, opened 2016-06-06
Adds aliases to stats command

### LINKED ISSUE #2680: Move `ipfs repo stat` to `ipfs stats repo`
I think that `ipfs repo stat` would be more appropriate under the `ipfs stats` header.

pfista:
Fixes #2680


whyrusleeping:
repo and bitswap make sense here, but i don't tihnk that object block and files do. The stats command is more for information relevant to the state of this node, where those three are information about data.

whyrusleeping:
LGTM

RichardLitt:
Thanks! See https://github.com/ipfs/http-api-spec/pull/99 for the http-api-spec update.

Issue #2803, opened 2016-06-03
improve http api


ligi:
the http api is weird - already had a short conversation about this in the chat - but I think it is better to put this in an issue.

ping currently returns:
```
⋊> ~ curl http://127.0.0.1:5001/api/v0/ping/QmaCpDMGvV2BGHeYERUEnRQAwe3N8SzbUtfsmvsqQLuvuJ                                     20:11:57
{"Success":false,"Time":0,"Text":"PING QmaCpDMGvV2BGHeYERUEnRQAwe3N8SzbUtfsmvsqQLuvuJ."}
{"Success":true,"Time":745266337,"Text":""}
{"Success":true,"Time":1375473326,"Text":""}
{"Success":true,"Time":230373161,"Text":""}
{"Success":true,"Time":337280582,"Text":""}
{"Success":true,"Time":220376323,"Text":""}
{"Success":true,"Time":928059102,"Text":""}
{"Success":true,"Time":2793408162,"Text":""}
{"Success":true,"Time":1162895192,"Text":""}
{"Success":true,"Time":403357018,"Text":""}
{"Success":true,"Time":1845389336,"Text":""}
{"Success":false,"Time":0,"Text":"Average latency: 1004.19ms"}
```

would be great if it returns just one json-object

swarm/peers currently returns

```
⋊> ~ curl http://127.0.0.1:5001/api/v0/swarm/peers                                                                             20:12:13
[...]

PR #2812, opened 2016-06-06
Fix refs local marshalling

Kubuxu:
Part of #2803

### DIFF
--- a/core/commands/refs.go
+++ b/core/commands/refs.go
@@ -3,7 +3,6 @@ package commands
 import (
 	"bytes"
 	"errors"
-	"fmt"
 	"io"
 	"strings"
 
@@ -169,7 +168,6 @@ var refsMarshallerMap = cmds.MarshalerMap{
 		marshal := func(v interface{}) (io.Reader, error) {
 			obj, ok := v.(*RefWrapper)
 			if !ok {
-				fmt.Println("%#v", v)
 				return nil, u.ErrCast()
 			}
 
--- a/core/commands/refs.go
+++ b/core/commands/refs.go
@@ -117,35 +117,8 @@ NOTE: List all references recursively by using the flag '-r'.
 			}
 		}()
 	},
-	Marshalers: cmds.MarshalerMap{
-		cmds.Text: func(res cmds.Response) (io.Reader, error) {
-			outChan, ok := res.Output().(<-chan interface{})
-			if !ok {
-				return nil, u.ErrCast()
-			}
-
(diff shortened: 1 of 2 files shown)


core/commands/refs.go
```diff
@@ -171,21 +144,47 @@ Displays the hashes of all local objects.
...
+	cmds.Text: func(res cmds.Response) (io.Reader, error) {
+		outChan, ok := res.Output().(<-chan interface{})
+		if !ok {
+			return nil, u.ErrCast()
+		}
+
+		marshal := func(v interface{}) (io.Reader, error) {
+			obj, ok := v.(*RefWrapper)
+			if !ok {
+				fmt.Println("%#v", v)
```
whyrusleeping: should make this a log.Errorf or a log.Warningf

Kubuxu: I think it isn't needed, it might be a left over from someone debugging something.
It got here when I was rearranging the code.

whyrusleeping:
@RichardLitt this changes the API

RichardLitt:
Thanks @whyrusleeping. See https://github.com/ipfs/http-api-spec/pull/98.

jbenet:
This is a breaking change, right? The summary in the changelog DOES NOT imply it:

> - Fix NDJSON output of `ipfs refs local`. (@Kubuxu, ipfs/go-ipfs#2812)

- I understand that this change was necessary to fix a problem, therefore needing to move to ndjson
- The old behavior may be desired by people who have implemented this. It should be possible to get the exact old behavior from the API with a flag in case bindings dont want to change all their code just yet. This can be considered deprecated and removable in the future.
- **NEVER** assume that "because nobody is reporting a problem that nobody is using it". Very wrong, very brittle, very fail. sad doge.

jbenet:
The changelog should say something like:

> - [BREAKING] Fix NDJSON output of `ipfs refs local`. (@Kubuxu, ipfs/go-ipfs#2812)
>   Breaking change necessary to correct buggy behavior. For the old output format, use `ipfs refs local --some-option`.

Kubuxu:
You can get the old behaviour using `text/plain` as a type of http-api.
They change is invisible from the CLI.

I shouldn't have disregarded this.

whyrusleeping:
Lets change the changelog to say:

> - [BREAKING] Fix NDJSON output of `ipfs refs local`. (@Kubuxu, ipfs/go-ipfs#2812)
>   Breaking change necessary to correct buggy behavior. For the old output format, use `ipfs refs local --enc=text`.

PR #2813, opened 2016-06-06
Improve ping API a bit by returning failure in case of only failure

Kubuxu:
Part of https://github.com/ipfs/go-ipfs/issues/2803
License: MIT

### DIFF
--- a/core/commands/ping.go
+++ b/core/commands/ping.go
@@ -118,7 +118,8 @@ func pingPeer(ctx context.Context, n *core.IpfsNode, pid peer.ID, numPings int) 
 		if len(n.Peerstore.Addrs(pid)) == 0 {
 			// Make sure we can find the node in question
 			outChan <- &PingResult{
-				Text: fmt.Sprintf("Looking up peer %s", pid.Pretty()),
+				Text:    fmt.Sprintf("Looking up peer %s", pid.Pretty()),
+				Success: true,
 			}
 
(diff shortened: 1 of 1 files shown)


core/commands/ping.go
```diff
@@ -131,14 +132,18 @@ func pingPeer(ctx context.Context, n *core.IpfsNode, pid peer.ID, numPings int)
...
+		}
 
 		ctx, cancel := context.WithTimeout(ctx, kPingTimeout*time.Duration(numPings))
 		defer cancel()
 		pings, err := n.Ping.Ping(ctx, pid)
 		if err != nil {
 			log.Debugf("Ping error: %s", err)
-			outChan <- &PingResult{Text: fmt.Sprintf("Ping error: %s", err)}
+			outChan <- &PingResult{
+				Text: fmt.Sprintf("Ping error: %s", err)}
```
whyrusleeping: lets explicitly set Success false here (for clarity to the reader), and then put the closing bracket on its own line.

Kubuxu:
Updated.

jbenet:
Not quite sure what this does-- the PR title doesn't tell me, and the code is not clear why-- maybe add some comments why some return success and one fail?

Kubuxu:
We have boolean filed `Success` which as boolean fields default to false, was implying failure in NDJSON messages that didn't fail. Now, we explicitly set success to true on all of them and to false if there is failure.

Issue #437, opened 2014-12-10
feat: `ipfs bitswap` command


mappum:
We're going to need to create the `ipfs bitswap` command.

**Subcommands:**
- `ipfs bitswap wantlist [<peer>]` - Displays a peer's wantlist, or our own if no peer provided
- `ipfs bitswap ledger <peer>` - Displays ledger info for a peer (debt ratio, amount of data exchanged, etc, could have a subcommand to reset a ledger or manually set values)
- `ipfs bitswap strategy` - Displays some info about the strategy in use (could have subcommands to adjust strategy settings)
#415 depends on this

diasdavid:
Currently, we have: 

```bash
$ ipfs bitswap --help

    ipfs bitswap - A set of commands to manipulate the bitswap agent

SUBCOMMANDS:

    ipfs bitswap stat            - show some diagnostic information on the bitswap agent
    ipfs bitswap unwant <key>... - Remove a given block from your wantlist
    ipfs bitswap wantlist        - Show blocks currently on the wantlist

    Use 'ipfs bitswap <subcmd> --help' for more information about each command.
```

Currently `ipfs bitswap <peerId>` is behind the `-p` flag (ipfs bitswap -p <peerId>)

Still missing:
- [ ] ipfs bitswap ledger <peer> - Displays ledger info for a peer (debt ratio, amount of data exchanged, etc, could have a subcommand to reset a ledger or manually set values)
- [ ] ipfs bitswap strategy - Displays some info about the strategy in use (could have subcommands to adjust strategy settings)

whyrusleeping:
I think we can close this, once we actually have bitswap strategies we can worry about a bitswap strategy subcommand

PR #2814, opened 2016-06-06
[WIP] command: bitswap ledger

pfista:
Starting work for #437

### DIFF
--- a/exchange/bitswap/decision/engine.go
+++ b/exchange/bitswap/decision/engine.go
@@ -93,7 +93,7 @@ type Engine struct {
 
 func NewEngine(ctx context.Context, bs bstore.Blockstore) *Engine {
 	e := &Engine{
-		ledgerMap:        make(map[peer.ID]*ledger),
+		ledgerMap:        make(map[peer.ID]*Ledger),
 		bs:               bs,
 		peerRequestQueue: newPRQ(),
 		outbox:           make(chan (<-chan *Envelope), outboxChanBuffer),
@@ -285,12 +285,12 @@ func (e *Engine) PeerDisconnected(p peer.ID) {
 
 func (e *Engine) numBytesSentTo(p peer.ID) uint64 {
 	// NB not threadsafe
-	return e.findOrCreate(p).Accounting.BytesSent
+	return e.FindOrCreate(p).Accounting.BytesSent
 }
 
 func (e *Engine) numBytesReceivedFrom(p peer.ID) uint64 {
 	// NB not threadsafe
-	return e.findOrCreate(p).Accounting.BytesRecv
+	return e.FindOrCreate(p).Accounting.BytesRecv
 }
 
 // ledger lazily instantiates a ledger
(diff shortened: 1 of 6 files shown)


pfista:
@whyrusleeping Would like your input if you think this is the right direction. I don't know if exporting `engine` and `ledger` is the best approach.

whyrusleeping:
@pfista the problem with working on this command now is that the ledger information isnt really used. Although i guess it might be interesting just so we can see how much data we've sent to each peer. 

I think instead of directly exporting the ledger, we should add a way to get a snapshot of a ledger. That way we avoid any risk of race conditions and keep things more well contained.

whyrusleeping:
This looks a lot cleaner. I'll need to take a better look in the morning, but LGTM right now :+1:

pfista:
It's worth mentioning too if this is not really needed anymore we should
just close the issue to avoid adding anything unnecessary.

whyrusleeping:
@pfista i like this command. I just need to get to reviewing it (sorry about the delay, the decentralized web conference and nodeconf took a fair bit of time away from github)

RichardLitt:
Subscribing.

exchange/bitswap/decision/engine.go
```diff
@@ -298,6 +298,22 @@ func (e *Engine) findOrCreate(p peer.ID) *ledger {
...
 
+type LedgerSnapshot struct {
+	DebtRatio     float64
+	BytesSent     uint64
+	BytesReceived uint64
+}
+
+func (e *Engine) LedgerSnapshot(p peer.ID) *LedgerSnapshot {
+	snapshot := &LedgerSnapshot{
+		DebtRatio:     e.findOrCreate(p).Accounting.Value(),
```
whyrusleeping: I think we're going to have to worry about locking a bit here. Lets make a single call to `findOrCreate` which is threadsafe on its own, then call lock on the returned ledger, defer the unlock, and then take the information we need from it

exchange/bitswap/decision/engine.go
```diff
@@ -298,6 +298,27 @@ func (e *Engine) findOrCreate(p peer.ID) *ledger {
...
 
+type LedgerSnapshot struct {
+	DebtRatio     float64
+	BytesSent     uint64
+	BytesReceived uint64
+}
+
+func (e *Engine) LedgerSnapshot(p peer.ID) *LedgerSnapshot {
+	e.lock.Lock()
+	defer e.lock.Unlock()
```
pfista: @whyrusleeping Looking at the other code I noticed you're locking on the engine. Let me know if instead I should actually lock on the ledger like you mentioned in your previous comment.

whyrusleeping: This will deadlock. `findOrCreate` takes the same lock that youre taking here. The correct way to do it will be to take the lock on the ledger you get back.

pfista: Oh duh. I hadn't rebased yet, the old findOrCreate wasn't locking on engine...

whyrusleeping:
closing in favor of #2852

PR #2819, opened 2016-06-07
clean up some dead code in the dht

whyrusleeping:
found some `0xdeadc0de`

License: MIT


Kubuxu:
LGTM, what to say :smile:

lgierth:
stat

freddy

lgierth:
Programming with backspace only! Much better than brainfuck!

Kubuxu:
You get random code and have to only remove code from it to make it do what you want.

We should make code jam like that.

whyrusleeping:
:D

Issue #2748, opened 2016-05-19
ipfs init fails with no stdin


whyrusleeping:
`echo "" | ipfs init` reproduces the issue. This is a problem if i want to run `ipfs init` in a docker container. (although passing `-t` is a workaround)

whyrusleeping:
This appears to be a problem with most commands actually

lgierth:
Likely introduced in #2581, we fixed a similar thing in #1263.

Kubuxu:
I can look into fixing it tomorrow.

Kubuxu:
Resolved in #2749

lgierth:
Well this hasn't exactly been resolved... we just reverted the commit.

dignifiedquire:
This is a real problem for us, ipfsd-ctl the module to control a go daemon from node.js is not able to work with 0.4.2 because of this bug.

lgierth:
hey @sivachandran, would be great to get this fixed for  0.4.3!

PR #2822, opened 2016-06-07
Change logic of stdin input to be last and only if required

Kubuxu:
This PR changes how stdin is used while parsing arguments. Current behaviour is that stdin is used for non-required and variadic arguments which causes confusion.

After the change the stdin is used only in case of missing required argument or if explicitly mentioned by the user with `-`.

Question: Should we print on stderr when ipfs command is listening for stdin to clear up possible confusing hang even more?

Aims to resolve https://github.com/ipfs/go-ipfs/issues/2748

License: MIT

### DIFF
--- a/test/bin/time-out
+++ b/test/bin/time-out
@@ -1,0 +1,91 @@
+#!/bin/bash
+#
+# The Bash shell script executes a command with a time-out.
+# Upon time-out expiration SIGTERM (15) is sent to the process. If the signal
+# is blocked, then the subsequent SIGKILL (9) terminates it.
+#
+# Based on the Bash documentation example.
+
+# Hello Chet,
+# please find attached a "little easier"  :-)  to comprehend
+# time-out example.  If you find it suitable, feel free to include
+# anywhere: the very same logic as in the original examples/scripts, a
+# little more transparent implementation to my taste.
+#
+# Dmitry V Golovashkin <Dmitry.Golovashkin@sas.com>
+
+scriptName="${0##*/}"
+
+declare -i DEFAULT_TIMEOUT=9
+declare -i DEFAULT_INTERVAL=1
+declare -i DEFAULT_DELAY=1
+
+# Timeout.
+declare -i timeout=DEFAULT_TIMEOUT
+# Interval between checks if the process is still alive.
+declare -i interval=DEFAULT_INTERVAL
+# Delay between posting the SIGTERM signal and destroying the process by SIGKILL.
+declare -i delay=DEFAULT_DELAY
+
+function printUsage() {
+    cat <<EOF
+
+Synopsis
+    $scriptName [-t timeout] [-i interval] [-d delay] command
+    Execute a command with a time-out.
+    Upon time-out expiration SIGTERM (15) is sent to the process. If SIGTERM
+    signal is blocked, then the subsequent SIGKILL (9) terminates it.
+
+    -t timeout
+        Number of seconds to wait for command completion.
+        Default value: $DEFAULT_TIMEOUT seconds.
+
+    -i interval
+        Interval between checks if the process is still alive.
+        Positive integer, default value: $DEFAULT_INTERVAL seconds.
+
+    -d delay
+        Delay between posting the SIGTERM signal and destroying the
+        process by SIGKILL. Default value: $DEFAULT_DELAY seconds.
+
+As of today, Bash does not support floating point arithmetic (sleep does),
+therefore all delay/time values must be integers.
+EOF
+}
+
+# Options.
+while getopts ":t:i:d:" option; do
+    case "$option" in
+        t) timeout=$OPTARG ;;
+        i) interval=$OPTARG ;;
+        d) delay=$OPTARG ;;
+        *) printUsage; exit 1 ;;
+    esac
+done
+shift $((OPTIND - 1))
+
+# $# should be at least 1 (the command to execute), however it may be strictly
+# greater than 1 if the command itself has options.
+if (($# == 0 || interval <= 0)); then
+    printUsage
+    exit 1
+fi
+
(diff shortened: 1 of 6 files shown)


whyrusleeping:
@Kubuxu could you provide a couple examples of how this changes command line usage for the end user? Does it have any effect on things like `echo "test" | ipfs add` ?

Kubuxu:
No it doesn't for the most part.

It changes behaviour if parameter is optional user as has to explicitly specify that he wants to use stdin using `-` argument. So it changes semantic of for example `ipfs init` where the file arugment is optional. To initialize repo from stdin one has to do `echo '{... config ... } | ipfs init -`.

As argument for add is required it will wait for input. I think we should add a message on stderr about it as currently `ipfs add` will display help, with this patch it will silently wait for user input.

Something along the lines of:
`Reading standard input, <Ctrl-D> to finish. $CMD_PATH -h for more information.`
if input is a terminal. It would greatly reduce confusion when command doesn't return.

@RichardLitt can you check wording of that?

test/bin/time-out
```diff
@@ -0,0 +1,91 @@
+#!/bin/bash
+#
+# The Bash shell script executes a command with a time-out.
+# Upon time-out expiration SIGTERM (15) is sent to the process. If the signal
+# is blocked, then the subsequent SIGKILL (9) terminates it.
+#
+# Based on the Bash documentation example.
+
+# Hello Chet,
```
RichardLitt: What's this?

test/bin/time-out
```diff
@@ -0,0 +1,91 @@
...
+# please find attached a "little easier"  :-)  to comprehend
+# time-out example.  If you find it suitable, feel free to include
+# anywhere: the very same logic as in the original examples/scripts, a
+# little more transparent implementation to my taste.
+#
+# Dmitry V Golovashkin <Dmitry.Golovashkin@sas.com>
```
RichardLitt: Who is this?

RichardLitt:
Looks fine to me from a wording perspective. Not sure what that shell script is.

test/bin/time-out
Kubuxu: It is from a website that was linked on SO, it was released as a free (public domain).
The script in whole is for running commands with timeout.

RichardLitt: Still, those comments aren't necessary, are they? I would remove them.

whyrusleeping:
tried this out, works good on my end. makes docker a bit happier.

Thanks @Kubuxu !

test/bin/time-out
```diff
@@ -0,0 +1,83 @@
+#!/bin/bash
+#
+# The Bash shell script executes a command with a time-out.
+# Upon time-out expiration SIGTERM (15) is sent to the process. If the signal
+# is blocked, then the subsequent SIGKILL (9) terminates it.
+#
+# Based on the Bash documentation example.
```
jbenet: @Kubuxu did you write this? if not, what is the license?

jbenet:
IIRC, this PR introduced a bunch of complexity and breaking changes. Have these been resolved? if so, where? (this may be ahead in my CR)

jbenet:
Found it https://github.com/ipfs/go-ipfs/pull/2902

test/bin/time-out
Kubuxu: It is Public Domain, I had previously credited the original author but removed it with other comments as for @RichardLitt 's request. I shouldn't have removed the author line, it was a signature for a short letter there.

jbenet:
More here https://github.com/ipfs/go-ipfs/pull/2952/files

jbenet:
would be good to isolate all these commits and see a diff between `v0.4.2..v0.4.3-rc3` with just all of the commits of these PRs. to ensure things got reverted appropriately

lgierth:
> would be good to isolate all these commits and see a diff between v0.4.2..v0.4.3-rc3 with just all of the commits of these PRs. to ensure things got reverted appropriately

I tried to somewhat group the 0.4.3-rc1 changelog entries logically, but the ones for stdin didn't turn out that well :) some entries are pretty good though and we should continue to aim for properly editorialized changelogs (i.e. logically grouped and structured)

lgierth:
> I tried to somewhat group the 0.4.3-rc1 changelog entries logically, but the ones for stdin didn't turn out that well :) some entries are pretty good though and we should continue to aim for properly editorialized changelogs (i.e. logically grouped and structured)

tl;dr merges don't map to changelog entries

test/bin/time-out
jbenet: Author line needs to be back before release.

Kubuxu: This should address the license issue: https://github.com/ipfs/go-ipfs/pull/3152

test/bin/time-out
```diff
@@ -0,0 +1,83 @@
+#!/bin/bash
```
chriscool: I am a bit concerned that this script is written in bash.

test/bin/time-out
```diff
@@ -0,0 +1,83 @@
...
+    -i interval
+        Interval between checks if the process is still alive.
+        Positive integer, default value: $DEFAULT_INTERVAL seconds.
+
+    -d delay
+        Delay between posting the SIGTERM signal and destroying the
+        process by SIGKILL. Default value: $DEFAULT_DELAY seconds.
+
+As of today, Bash does not support floating point arithmetic (sleep does),
+therefore all delay/time values must be integers.
```
chriscool: We have go-sleep that accepts floating point delays.

test/bin/time-out
Kubuxu: @chriscool we removed this script in https://github.com/ipfs/go-ipfs/pull/3152

chriscool: Yeah, I saw it afterwards. Thanks for having written something more portable!

Issue #2820, opened 2016-06-07
ipfs ls gets stuck


Kubuxu:
ipfs object get works but ls gets stuck:
```
ipfs object get QmcmPko2PHCLnLqZTZXjPf4FLKyYjKDpb9VDbH6cVWTsU7
{"Links":[{"Name":".githash","Hash":"QmdG39xEJFK2YmYFxBYip1wCfvhSjVEXqCYnERJY2JY7rC","Size":49},{"Name":"ajax","Hash":"Qmdswc3USf1YoXWbUHUMDAcvwDjkwuPcq5ja4kiFAy5rZx","Size":20796815067}],"Data":"\u0008\u0001"}
```

Goroutine trace: https://gist.github.com/Kubuxu/ab56191ccb9aa725d5230f22bcd86f1a

It is probably due to fact that one file got lost and is no longer available: `fs:/ipfs/QmcmPko2PHCLnLqZTZXjPf4FLKyYjKDpb9VDbH6cVWTsU7/.githash`

Kubuxu:
This can be resolved if we drop `Type` from `ls` result.

whyrusleeping:
yeah... `ipfs ls` has the unixfs info encoded in it. We made `ipfs files ls` to be the 'correct' tool for this purpose. I wonder if we should look at deprecating that feature of `ipfs ls`.

In any case, you can add `--timeout=10s` to the command to make it not take longer than ten seconds

Kubuxu:
The only thing that blocks the node is adding a slash over here: https://github.com/ipfs/go-ipfs/blob/master/core/commands/ls.go#L119

Of course it can be useful from the API but please take a look at #2824

PR #2824, opened 2016-06-08
Add flag to ls to not resolve type of subnodes

Kubuxu:
This makes ls return even if files referenced in the directory are not available.
It also introduces a `resolve-type` flag that forces the resolution of type.

Resolves https://github.com/ipfs/go-ipfs/issues/2820
 
License: MIT

### DIFF
--- a/core/commands/ls.go
+++ b/core/commands/ls.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"text/tabwriter"
 
+	key "github.com/ipfs/go-ipfs/blocks/key"
 	cmds "github.com/ipfs/go-ipfs/commands"
 	core "github.com/ipfs/go-ipfs/core"
 	merkledag "github.com/ipfs/go-ipfs/merkledag"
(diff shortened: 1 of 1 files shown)


whyrusleeping:
I don't like this, it can silently provide misleading information. I think that we could move the type query under a flag (or have a flag to disable it).

Kubuxu:
As it is an int, maybe I will make it `-1` when it could not be resolved, as an error code.

Kubuxu:
How does it look now?

whyrusleeping:
@Kubuxu I just don't like the possiblity of returning partial data. It feels kinda wrong to me...

Kubuxu:
Then IMO we should remove this from API completely (maybe only with special flag), as it has quite high drawback for such small benefit in usual use case (someone doing `ipfs ls`).

whyrusleeping:
Yeah, thats what i meant by moving the type query under a flag.

Kubuxu:
@whyrusleeping what is the correct way to getting a object if and only if it is local?

whyrusleeping:
@Kubuxu use the blockstore (not blockservice) directly. Or construct a new dagservice with an 'offline' exchange

Kubuxu:
Ok, now it doesn't resolve the subnode data unless the data is available locally.
There is also a flag to force resolution.

@RichardLitt this changes API.

whyrusleeping:
@Kubuxu i'd like this to default to true, so we maintain compatibility with our existing behaviour

Kubuxu:
Done

RichardLitt:
@Kubuxu How do I use `resolve-type` in a curl request? Any ideas?

core/commands/ls.go
```diff
@@ -79,21 +87,42 @@ format:
...
-					return
+				t := unixfspb.Data_DataType(-1)
+				linkKey := key.Key(link.Hash)
+				if ok, err := node.Blockstore.Has(linkKey); ok && err == nil {
+					b, err := node.Blockstore.Get(linkKey)
+					if err != nil {
+						res.SetError(err, cmds.ErrNormal)
+						return
+					}
+					linkNode, err = merkledag.DecodeProtobuf(b.Data())
```
jbenet: ok wat-- all this merkledag code shouldnt have to be out here. this should all be handled by util functions in either `unixfs` or some other package. this is leaking abstractions out

PR #2826, opened 2016-06-08
respect contexts while reading messages in dht

whyrusleeping:
If something weird happens during a dht request and the remote peer does not reply to our request, and *also* does not close the underlying connection, this has the possibility to hang.

That said, I think that the issue i noticed was caused by utp hanging up under the hood (the stack trace showed a few utp event loops running)

License: MIT

### DIFF
--- a/routing/dht/dht_net.go
+++ b/routing/dht/dht_net.go
@@ -214,7 +214,7 @@ func (ms *messageSender) SendRequest(ctx context.Context, pmes *pb.Message) (*pb
 	log.Event(ctx, "dhtSentMessage", ms.dht.self, ms.p, pmes)
 
 	mes := new(pb.Message)
-	if err := ms.r.ReadMsg(mes); err != nil {
+	if err := ms.ctxReadMsg(ctx, mes); err != nil {
 		ms.s.Close()
 		ms.s = nil
 		return nil, err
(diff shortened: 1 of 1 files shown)


routing/dht/dht_net.go
```diff
@@ -227,3 +228,23 @@ func (ms *messageSender) SendRequest(ctx context.Context, pmes *pb.Message) (*pb
 
 	return mes, nil
 }
+
+func (ms *messageSender) ctxReadMsg(ctx context.Context, mes *pb.Message) error {
+	t := time.NewTimer(time.Second * 30)
```
Kubuxu: I think we should collect all of those timeouts (and create a one that would be standard), this way if we need to make IPFS work over higher latency for some reason it will be much easier.

whyrusleeping: I honestly think we could get away without the timeout here and just stick to respecting the context and fixing utp. 

What do you think?

Kubuxu: What do you mean by fixing utp?

whyrusleeping: theres a bug in utp that surfaces occasionally that will cause it to hang. Theres nothing we can do to avoid this other than fix utp, or not use utp. The hang that i wrote this PR to 'solve' was caused by this utp bug.

Kubuxu: Ok, let's pull it but we should remember to fix it properly in future.

PR #2832, opened 2016-06-09
In blockstore tests treat Close and Closing the same way

### LINKED ISSUE #2762: TestAllKeysRespectsContext fails randomly
http://ci.ipfs.team:8111/viewLog.html?buildId=2160&buildTypeId=GoIpfs_CiTests&tab=buildLog#_

```
[19:06:48][Step 3/4] TestAllKeysRespectsContext
[19:06:48][TestAllKeysRespectsContext] 	blockstore_test.go:194: should not be closed
[19:06:48][TestAllKeysRespectsContext] TestValueTypeMismatch
[19:06:48][TestAllKeysRespectsContext] TestReturnsErrorWhenSizeNegative
[19:06:48][TestAllKeysRespectsContext] TestRemoveCacheEntryOnDelete
[19:06:48][TestAllKeysRespectsContext] TestElideDuplicateWrite
[19:06:48][TestAllKeysRespectsContext] FAIL
[19:06:48][TestAllKeysRespectsContext] exit status 1
[19:06:48][TestAllKeysRespectsContext] FAIL	github.com/ipfs/go-ipfs/blocks/blockstore	0.010s
[19:06:48][TestAllKeysRespectsContext] Process exited with code 1
[19:06:49][TestAllKeysRespectsContext] Step go tests (Command Line) failed
```

Kubuxu:
This should prevent race condition that caused https://github.com/ipfs/go-ipfs/issues/2762

Resolves #2762 

License: MIT

### DIFF
--- a/blocks/blockstore/blockstore_test.go
+++ b/blocks/blockstore/blockstore_test.go
@@ -190,8 +190,10 @@ func TestAllKeysRespectsContext(t *testing.T) {
 		select {
 		case <-done:
 			t.Fatal("sync is wrong")
+		// As time differentce between Closed and Closing is so small we have to check on
+		//  both those channels at the same time otherwise we get a race which causes
+		//  random test failures sometimes.
 		case <-results.Process().Closed():
-			t.Fatal("should not be closed") // should not be closed yet.
 		case <-results.Process().Closing():
 			// should be closing now!
 			t.Log("closing correctly at this point.")


whyrusleeping:
I'm not sure i follow how this solves the issue, could you elaborate?

Kubuxu:
If the cancel is run in the same goroutine then it will send signal to the already waiting goroutine which will cause the goprocess to close. By running it in separate goroutine we should give ourselves enough time to reach the `select` statement before the `cancel` causes the goprocess to close.

Kubuxu:
This fix doesn't work as I thought it would.
As I am able to replicate the issue I might be able to fix it now.

Kubuxu:
I've changed the approach as that one was not working and I was able to replicate the issue by repeatability running tests.

Kubuxu:
Ok, it looks like it isn't test inconsistency after all. I am getting failures at https://github.com/ipfs/go-ipfs/pull/2832/files#diff-89a41509a8389c8bc3dbf7594aa13416R192 now.

Issue #2828, opened 2016-06-09
Massive DHT traffic on adding big datastes


Kubuxu:
@Magik6k was trying to add cdnjs (something I did few months ago) he have hit https://github.com/ipfs/go-ipfs/issues/2823 but also reported that adding the dataset which was about 21GB created about 3TB of traffic.

@Magik6k could you give more precise data on this?

whyrusleeping:
The solution i'm investigating here is batching together outgoing provide operations, combining findpeer queries and outgoing 'puts'. The issue is that it might require making some backwards incompatible changes to the dht protocol, which isnt really a huge problem since we have multistream

magik6k:
* Network: 800 GiB Out, 2 TiB in
* While adding files it was using ~50/100Mbps
* Around 10-20 Files/s, depending on file size, usually around 10-100KB
* 2633788 files
* 2-3 days IIRC

whyrusleeping:
@magik6k Thanks! I will keep you posted as we make improvements to this.

parkan:
Similar report, while adding ~70,000 objects (~100k ea) we were maxing out our instance's traffic (70MBit in, 160MBit out) for over 14 hours (!!!), at which point someone killed the IPFS daemon.

Back of the envelope, this is at least 1TB outgoing for ~10GB files.

One particular problem: because the daemon was killed before the process completed, it seems that *almost none of the files were pinned*. Running `ipfs repo gc` deleted almost everything:

```
; before gc
NumObjects       2995522
RepoSize         20
RepoPath         /home/ubuntu/.ipfs
Version          fs-repo@4

; after gc
NumObjects       771
RepoSize         20
RepoPath         /home/ubuntu/.ipfs
Version          fs-repo@4
```

(side note, the chunk counts seem way too high too, almost 40x per image instead of expected 4x)

whyrusleeping:
@parkan which version of ipfs are you using? This is definitely a known issue, but it has been improved slightly since 0.4.2

parkan:
@whyrusleeping 0.4.3-rc3

parkan:
@whyrusleeping my intuition is that there's some kind of context that's not being released before a queue is drained, which is supported by *none* of the files being pinned successfully after killing the client. Does this seem plausible?

whyrusleeping:
@parkan are you killing the daemon before the add call is complete?

parkan:
[...]

PR #2841, opened 2016-06-12
a few small changes to make the dht more efficient

whyrusleeping:
Was watching biham during a reproviding round. Its eating up some serious CPU time, this should help alleviate a small amount of that.

License: MIT

### DIFF
--- a/routing/kbucket/table.go
+++ b/routing/kbucket/table.go
@@ -144,9 +144,9 @@ func (rt *RoutingTable) NearestPeer(id ID) peer.ID {
 
 // NearestPeers returns a list of the 'count' closest peers to the given ID
 func (rt *RoutingTable) NearestPeers(id ID, count int) []peer.ID {
-	rt.tabLock.RLock()
-	defer rt.tabLock.RUnlock()
 	cpl := commonPrefixLen(id, rt.local)
+
+	rt.tabLock.RLock()
 
 	// Get bucket at cpl index or last bucket
 	var bucket *Bucket
@@ -170,6 +170,7 @@ func (rt *RoutingTable) NearestPeers(id ID, count int) []peer.ID {
 			peerArr = copyPeersFromList(id, peerArr, plist)
 		}
 	}
+	rt.tabLock.RUnlock()
 
 	// Sort by distance to local peer
 	sort.Sort(peerArr)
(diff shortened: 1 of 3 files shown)


Kubuxu:
LGTM

whyrusleeping:
should improve #2828

routing/kbucket/table.go
```diff
@@ -48,11 +48,11 @@ func NewRoutingTable(bucketsize int, localID ID, latency time.Duration, m pstore
 // Update adds or moves the given peer to the front of its respective bucket
 // If a peer gets removed from a bucket, it is returned
 func (rt *RoutingTable) Update(p peer.ID) {
-	rt.tabLock.Lock()
-	defer rt.tabLock.Unlock()
 	peerID := ConvertPeerID(p)
 	cpl := commonPrefixLen(peerID, rt.local)
 
+	rt.tabLock.Lock()
+	defer rt.tabLock.Unlock()
```
jbenet: this no longer protects `rt.local`, which is fine. BUT it would be nice if the `RoutingTable` struct was more clear about what the mutex protects.

### now

```go
// RoutingTable defines the routing table.
type RoutingTable struct {

	// ID of the local peer
	local ID

	// Blanket lock, refine later for better performance
	tabLock sync.RWMutex

	// latency metrics
	metrics pstore.Metrics

	// Maximum acceptable latency for peers in this cluster
	maxLatency time.Duration

	// kBuckets define all the fingers to other nodes.
	Buckets    []*Bucket
	bucketsize int
}
```

### better

```go
// RoutingTable defines the routing table.
type RoutingTable struct {

	local      ID  // ID of the local peer
	bucketsize int

	// the following fields are all protected by tabLock
	tabLock    sync.RWMutex   // Blanket lock, refine later for better performance
	metrics    pstore.Metrics // latency metrics
	maxLatency time.Duration  // Maximum acceptable latency for peers in this cluster
	Buckets    []*Bucket      // kBuckets define all the fingers to other nodes.
}
```

jbenet: Or at least say:
```go
	// Blanket lock, refine later for better performance
	// protects metrics, maxLatency, and Buckets
	tabLock sync.RWMutex
```

PR #2852, opened 2016-06-14
bitswap: add `ledger` subcommand

thomas-gardner:
Bring on the Strategies.

### DIFF
--- a/core/commands/bitswap.go
+++ b/core/commands/bitswap.go
@@ -5,11 +5,12 @@ import (
 	"fmt"
 	"io"
 
-	"gx/ipfs/QmPSBJL4momYnE7DcUyk2DVhD6rH488ZmHBGLbxNdhU44K/go-humanize"
-
 	key "github.com/ipfs/go-ipfs/blocks/key"
 	cmds "github.com/ipfs/go-ipfs/commands"
 	bitswap "github.com/ipfs/go-ipfs/exchange/bitswap"
+	decision "github.com/ipfs/go-ipfs/exchange/bitswap/decision"
+
+	"gx/ipfs/QmPSBJL4momYnE7DcUyk2DVhD6rH488ZmHBGLbxNdhU44K/go-humanize"
 	peer "gx/ipfs/QmRBqJF7hb8ZSpRcMwUt8hNhydWcxGEhtk81HKq6oUwKvs/go-libp2p-peer"
 	u "gx/ipfs/QmZNVWh8LLjAavuQ2JXuFmuYH3C11xo988vSgp7UQrTRj1/go-ipfs-util"
 )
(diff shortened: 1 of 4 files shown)


whyrusleeping:
please take a look at #2814

core/commands/bitswap.go
```diff
@@ -154,12 +157,16 @@ var bitswapStatCmd = &cmds.Command{
 				return nil, u.ErrCast()
 			}
 			buf := new(bytes.Buffer)
-			fmt.Fprintln(buf, "bitswap status")
-			fmt.Fprintf(buf, "\tprovides buffer: %d / %d\n", out.ProvideBufLen, bitswap.HasBlockBufferSize)
-			fmt.Fprintf(buf, "\tblocks received: %d\n", out.BlocksReceived)
-			fmt.Fprintf(buf, "\tdup blocks received: %d\n", out.DupBlksReceived)
-			fmt.Fprintf(buf, "\tdup data received: %s\n", humanize.Bytes(out.DupDataReceived))
-			fmt.Fprintf(buf, "\twantlist [%d keys]\n", len(out.Wantlist))
+			fmt.Fprintf(buf, "bitswap status\n"+
```
RichardLitt: These should all be initially capitalized, I think.

thomas-gardner:
Hi @whyrusleeping, is `findOrCreate` really necessary? If the ledger doesn't exist, is there any reason to have this command create one?

whyrusleeping:
@thomas-gardner The find or create is necessary for the function of the bitswap code. Its an atomic operation to fetch the ledger for a peer (and create one while holding the lock if it doesnt exist).

Also, this PR is mostly a duplicate of #2814

thomas-gardner:
@whyrusleeping---right, something about creating an unneeded ledger was
confusing to me, but I can see how using `findOrCreate` keeps things
tidy and atomic.

Re diff #2814, this PR doesn't export any of the `Bitswap` fields, & gets
the locking right now ( ...I hope.)

whyrusleeping:
@thomas-gardner alright, if you could add a `ShortDescription` to the command and rebase this on master i'll get it merged in.

jbenet:
Nice!

jbenet:
We should move bitswap out into go-bitswap so it can get dev cycles faster
than in go-ipfs. Can write small tools to test it and leverage it so it can
evolve quickly and merge back into go-ipfs as needsd

thomas-gardner:
Jeromy Johnson:
> @thomas-gardner alright, if you could add a `ShortDescription` to the
> command and rebase this on master i'll get it merged in.

Rebased & added.

whyrusleeping:
alright, LGTM

PR #2854, opened 2016-06-14
Remove errors pointed out by govet

Kubuxu:
License: MIT


whyrusleeping:
Nice, LGTM

Issue #2745, opened 2016-05-19
TestBootstrap fails occasionally 


Kubuxu:
http://ci.ipfs.team:8111/viewLog.html?buildId=2045&buildTypeId=GoIpfs_CiTests&tab=buildLog#_focus=4136&state=4136

More here: https://github.com/ipfs/go-ipfs/issues/188

whyrusleeping:
i hate this one. Its the 'misdial' issue, it affects a few different tests.

whyrusleeping:
https://circleci.com/gh/ipfs/go-ipfs/4055

Kubuxu:
My sanity tests caught it: https://travis-ci.org/ipfs/go-ipfs/jobs/136932615#L3812

Kubuxu:
I am able to replicate failure in less than a minute:
```
cd routing dht
while time go test -run TestBootstrap; do; done
```
Looks like failure about one in five.

whyrusleeping:
https://travis-ci.org/ipfs/go-ipfs/jobs/137017648 heres another one

Kubuxu:
It think I figured out what is happening, the SO_REUSE_PORT flag on a socket makes it possible for socket `127.0.0.1:0` to randomise to the same address. I am running now tests on my server with ` IPFS_REUSEPORT=false` env variable. If I am sure that it is the solution (no failure for few hours) I will work on a solution.

Kubuxu:
I found it pretty interesting that it is a problem but it makes sense, it is birthday paradox with calendar not of 365 days but 2^16. As it turns out for 10% failure rate binding 140 addresses is enough: https://www.wolframalpha.com/input/?i=y%3D+1+-+(((2%5E16)+nPr+n+)%2F(2%5E16)%5En)++from+0+to+200

I still will wait few hours to confirm that issues is gone.

Issue #2809, opened 2016-06-05
TestPeriodicBootstrap fails occasionally


Kubuxu:
same misdial issue as https://github.com/ipfs/go-ipfs/issues/2745

Issue #188, opened 2014-10-19
Timing tests


jbenet:
Need to move away from timing tests, to something sync/mock based like: https://www.youtube.com/watch?v=ndmB0bj7eyw

whyrusleeping:
I think we can probably close this now, our tests are much better in regards to relying on timers

briantigerchow:
https://build.protocol-dev.com/job/gotest/696/console

failure at bd1d8767eb0810de66ee5b226b937d7c81a29e13

ok  	github.com/jbenet/go-ipfs/p2p/net/mock	0.898s
--- FAIL: TestDialBackoff (1.61 seconds)
	dial_test.go:103: new silent peer: <peer.ID QmUDmJ> /ip6/::1/tcp/54504
	dial_test.go:235: s2 should not happen immediately
	dial_test.go:252: s2 took too long
	dial_test.go:260: s3 should not happen yet
	dial_test.go:276: s3 took too long

whyrusleeping:
hrm...  thats a new one. Agreed on the reopen

jbenet:
That's  my fault. it's a timing test :( -- it's hard to do real network tests without using time. i'll try to rework it / make it more generous. maybe should add some signals inside the swarm functions itself.

cryptix:
[Found](https://build.protocol-dev.com/job/race/17421/consoleText) another timing related one for the list: `TestStBackpressureStreamWrite` from `p2p/test/backpressure`.

jbenet:
- [ ] TestBootstrap-- it so annoying.

cryptix:
- [ ] [TestSimultOpenFewStress](https://github.com/jbenet/go-ipfs/blob/88dea5771d32cd3864a904731d4aaa55435be965/p2p/net/swarm/simul_test.go#L59) - maybe have a global or return a chan from `SubtestSwarm` to know when to advance?

jbenet:
Saw this one again:

```
--- FAIL: TestBootstrap (0.56s)
	dht_test.go:303: connecting 30 dhts in a ring
	dht_test.go:79: dial attempt failed: misdial to <peer.ID NsEZww> through /ip4/127.0.0.1/tcp/51254 (got <peer.ID VnbQqa>)
TestBogusPrivateKey.Sign -- this better be a test!
FAIL
FAIL	github.com/ipfs/go-ipfs/routing/dht	2.356s
```

whyrusleeping:
[...]

PR #2855, opened 2016-06-14
Fix DHT tests failing because of repeated addresses

Kubuxu:
Due to SO_REUSEPORT it is possible for a localhost:0
address to repeat. This causes failure in DHT tests
where we spun up a lot of nodes inside test.

As for a birthday paradox it is enough to use 140
ports to get 20% chance for collision which was causing
failure in our case.

The fix is to disable REUSE_PORT routine for the
tests and leave it running for sharness tests where
not that many addresses are used at the same time.

Resolves https://github.com/ipfs/go-ipfs/issues/2745 https://github.com/ipfs/go-ipfs/issues/2809 https://github.com/ipfs/go-ipfs/issues/188

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,9 +4,9 @@ IPFS_MIN_GX_VERSION = 0.6
 IPFS_MIN_GX_GO_VERSION = 1.1
 
 ifeq ($(TEST_NO_FUSE),1)
-  go_test=go test -tags nofuse
+  go_test=IPFS_REUSEPORT=false go test -tags nofuse
 else
-  go_test=go test
+  go_test=IPFS_REUSEPORT=false go test
 endif
 
 
@@ -92,10 +92,10 @@ test_go_race:
 	$(go_test) ./... -race
 
 test_sharness_short:
-	cd test/sharness/ && make
+	make -C test/sharness/
 
 test_sharness_expensive:
-	cd test/sharness/ && TEST_EXPENSIVE=1 make
+	TEST_EXPENSIVE=1 make -C test/sharness/
 
 test_all_commits:
 	@echo "testing all commits between origin/master..HEAD"


whyrusleeping:
Awesome job catching this, yet another `SO_REUSEPORT` issue... yay

jbenet:
- Great find!
- We should continue to have `SO_REUSEPORT` being stress tested because we may find issues with it (maybe it behaves differently or poorly). If not in these tests, maybe in others? Or would there be another way of signaling ports in the tests here?

Makefile
```diff
@@ -92,10 +92,10 @@ test_go_race:
 	$(go_test) ./... -race
 
 test_sharness_short:
-	cd test/sharness/ && make
+	make -C test/sharness/
 
 test_sharness_expensive:
-	cd test/sharness/ && TEST_EXPENSIVE=1 make
+	TEST_EXPENSIVE=1 make -C test/sharness/
```
jbenet: @chriscool was there a reason to do the `make` after a `cd`? pls confirm this change is perfectly fine with you.

jbenet:
Hold up, this disabled `SO_REUSEPORT` in _ALL_ the go tests. this is pretty bad and flying blind. if there's any more trouble with `SO_REUSEPORT` we'll find it with stress testing all the socket use, which the go-ipfs suite does pretty well. Obviously the "accidentally reuse port" problem really sucks... is there another way of going about this? Maybe in a more granular way? (in the failing tests, or in tests which allocate tons of ports). Of course, if we can do a different way of allocating ports even better.

Kubuxu:
Some smarter way of allocating ports would be the best solution, unfortunately it is hard to do without race of conditions (I don't know if there is a way at all to do it 100% correctly).

The go-reuseport is very deep in the stack so disabling it only locally would require some refactors, that is why I decided to disable it globally.

Also any test that would stress go-reuseport would have quite a high change of triggering the collision.

I am all for re-enabling the go-reuseport in tests but we need to work on smart port selection first (0.0.0.0/0 is just not working with SO_REUSEPORT set).

Makefile
chriscool: Yeah, it's fine. One could also use `make -C test/sharness/ TEST_EXPENSIVE=1`. It's a matter of preference whether one wants the shell or `make` to do the `cd` and to set the environment variable.

Issue #2835, opened 2016-06-10
daemon crash on git master


czaks:
This issue happened while simultaneously executing ```ipfs add -pr bigdirtree/``` and downloading some file from the local machine, via ipfs.io to the local machine back. commit: b731c79 (master). The issue happened 3 times, I'm now unsuccessfully trying to reproduce it on a release branch.

```
panic: runtime error: slice bounds out of range

goroutine 103770 [running]:
panic(0xe5d800, 0xc82000e160)
        /usr/lib/golang/src/runtime/panic.go:481 +0x3e6
bufio.(*Reader).Read(0xc8247604e0, 0xc82455d5a0, 0x11, 0x11, 0xca4f40, 0x0, 0x0)
        /usr/lib/golang/src/bufio/bufio.go:214 +0x3c8
io.ReadAtLeast(0x7feb80f060e8, 0xc8247604e0, 0xc82455d5a0, 0x11, 0x11, 0x11, 0x0, 0x0, 0x0)
        /usr/lib/golang/src/io/io.go:297 +0xe6
io.ReadFull(0x7feb80f060e8, 0xc8247604e0, 0xc82455d5a0, 0x11, 0x11, 0x11, 0x0, 0x0)
        /usr/lib/golang/src/io/io.go:315 +0x62
gx/ipfs/QmZ4Qi3GaRbjcx28Sme5eMH7RQjGkt8wHxt2a65oLaeFEV/gogo-protobuf/io.(*varintReader).ReadMsg(0xc825f6c340, 0x7feb80eb0868, 0xc8284237a0, 0x0, 0x0)
        /home/*/src/gx/ipfs/QmZ4Qi3GaRbjcx28Sme5eMH7RQjGkt8wHxt2a65oLaeFEV/gogo-protobuf/io/varint.go:117 +0x1c2
github.com/ipfs/go-ipfs/routing/dht.(*messageSender).ctxReadMsg.func1(0xc824760840, 0xc82236dec0, 0xc8284237a0)
        /home/*/src/github.com/ipfs/go-ipfs/routing/dht/dht_net.go:234 +0x73
created by github.com/ipfs/go-ipfs/routing/dht.(*messageSender).ctxReadMsg
        /home/*/src/github.com/ipfs/go-ipfs/routing/dht/dht_net.go:235 +0x8b

```

whyrusleeping:
Thanks for reporting, I saw this once too but assumed it was some of the code i was messing with.

[...]

PR #2856, opened 2016-06-15
pass reference to reader instead of using the one on the object

whyrusleeping:
this should address #2835. I was able to reproduce the issue last night without this fix, but havent been able to reproduce the issue this morning (with or without the fix)

License: MIT


Kubuxu:
LGTM

PR #2857, opened 2016-06-15
Remove failing blockstore with context test

Kubuxu:
Why is it failing:
process is started, cancel() is called,
between we satart listening to the channels
in select statemnet there is race of three things
that can happent:
 1. Task can complete
 2. Task can start closing <- expected
 3. Task already closed

This race causes failures of the test.

It is basing heavily on race of conditions where
the task not closing, nor the task is completed
before channels are being listened.

It is quite impossible to resolve without
adding bunch of timings in there, which we
want to avoid, as there is no atomic
"send message on channel and select" in Golang

License: MIT


whyrusleeping:
LGTM

Issue #2601, opened 2016-04-25
In datastote.NewKey calling path.Clean is dangerous


kevina:
[I am not sure the best place to post this as this issue is related to both go-ipfs and go-datastore, I can move this to this to go-datastore if necessary.]

I may be completely missing something here but it seams datastore keys are represented via binary `[]byte` strings, possible prefixed with a "/" and are converted to a `datastore.Key` by calling `datastore.NewKey()`.

However, `datastore.NewKey()` also calls `path.Clean()`.  This is a dangerous thing to do with binary strings.  For example the binary `[]byte` representation of the hash might just happen to contain a "//" which `path.Clean` will convert to a "/".  In fact it seams that all of `datastore.Key` is written with the assumption the keys are text and not binary blobs.

Here is the relevant code in [go-datastore/key.go](https://github.com/ipfs/go-datastore/blob/545f590/key.go):
```go
// NewKey constructs a key from string. it will clean the value.
func NewKey(s string) Key {
	k := Key{s}
	k.Clean()
	return k
}

// Clean up a Key, using path.Clean.
func (k *Key) Clean() {
	k.string = path.Clean("/" + k.string)
}
```

The probability of a `[]byte` string for a multihash containing something that `path.Clean()` will remove or change is low, which is why I image this issue has not come up yet.

I am not sure the best way to fix this.  Maybe it might be better to keep the mutihash in its b58 encoding to mesh better with how `datastore.Key` was (supposedly) designed.  Or as a quick fix, maybe it would be better to remove the call to `path.Clean` in the `datastore.Key.Clean()` method.

whyrusleeping:
@kevina yeah, this is a known issue: https://github.com/ipfs/go-ipfs/issues/994

Its the cause of the following failure (and a few others i'm not finding right now).
https://github.com/ipfs/go-ipfs/blob/master/test/sharness/t0080-repo.sh#L112

I'd say we can keep this issue open as it will add more pressure towards us getting this fixed.

whyrusleeping:
I should also note that currently this is much less of an issue than you would think. Since the mangling of keys is consistent, we can always access the keys we like we expect. The only noticeable issue this causes is that when running a gc, any key affected by this bug won't be removed. (also you won't be able to see keys affected by this in an `ipfs refs local`)

kevina:
There is also the small possibility that this bug will create a collision when there normally would not be one, although the probability will be low.  For example lets's assume we have these two (8-byte) keys:
```
[...]

PR #2860, opened 2016-06-16
Write providers to disk to avoid memory leaks

whyrusleeping:
This is a rough first hack on this. It works as intended, but i'm not happy with the actual way of putting the information on disk.

Currently, it puts the list of peers providing a given key to the datastore at `/providers/<KEY>`. This means that everytime a provider is added, we write marshal all the providers and write to the datastore.

License: MIT


whyrusleeping:
I'm thinking it might be best to just write `/providers/<KEY>/<peerid>` for each provider entry instead of storing the whole set at once

whyrusleeping:
@lgierth @Kubuxu @kevina If you guys have thoughts or CR for this, that would be much appreciated

Kubuxu:
LGTM.

whyrusleeping:
This also has the super cool advantage of persisting providers information across node reboots

whyrusleeping:
@Kubuxu @jbenet @kevina I'm converting keys and peerIDs to hex before passing them to the datastore `NewKey` constructor. This is a workaround for the go-datastore bug #2601 

Before merging this, I think we might want to decide if this is the way we want to do this moving forward, or if we should attempt a fix at the go-datastore level (allowing us to skip the hex encoding)

kevina:
@whyrusleeping #2601 is causing some problems for me in the filestore, in particular the filestore maintenance commands will report the managed hash which can confuse users at best and at also cause problems when trying to access the block outside of the filestore, I think we should work to fix this once and for all.  I image any fix will likely introduce a repo. change.  Let's take this discussion over to #2601.

routing/dht/providers/providers.go
```diff
@@ -81,7 +81,7 @@ func NewProviderManager(ctx context.Context, local peer.ID, dstore ds.Datastore)
 const providersKeyPrefix = "/providers/"
 
 func mkProvKey(k key.Key) ds.Key {
-	return ds.NewKey(providersKeyPrefix + hex.EncodeToString([]byte(k)))
+	return ds.NewKey(providersKeyPrefix + base64.StdEncoding.EncodeToString([]byte(k)))
 }
```
kevina: @whyrusleeping I would use RawURLEncoding.  (1) StdEncoding includes '/' in the Alphabet (see https://tools.ietf.org/html/rfc4648) URLEncoding does not (hence it's name) (2) The RAW form does not include the unnecessary '=' padding character.

routing/dht/providers/providers.go
```diff
@@ -81,7 +81,7 @@ func NewProviderManager(ctx context.Context, local peer.ID, dstore ds.Datastore)
 const providersKeyPrefix = "/providers/"
 
 func mkProvKey(k key.Key) ds.Key {
-	return ds.NewKey(providersKeyPrefix + base64.StdEncoding.EncodeToString([]byte(k)))
+	return ds.NewKey(providersKeyPrefix + base64.URLEncoding.EncodeToString([]byte(k)))
 }
```
kevina: @whyrusleeping Recommend we standardize on RawURLEncoding, that is how base64url is generally encoded.

whyrusleeping: ah, without the equals signs? sure.

whyrusleeping:
@kevina @Kubuxu could I get you guys to review this PR? Would be very helpful

Kubuxu:
SGTM but I would like one more pair of eyes on it.

kevina:
@whyrusleeping there is a failed test, should we worry about it?

Kubuxu:
@kevina no it isn't connected

Kubuxu:
`t0060-daemon.sh` compares daemon output in tests. It needs fixing.

whyrusleeping:
@Kubuxu thats fixed in #2891

routing/dht/providers/providers.go
```diff
@@ -0,0 +1,363 @@
...
+	}
+	return nil
+}
+
+func (pm *ProviderManager) getAllProvKeys() ([]key.Key, error) {
+	res, err := pm.dstore.Query(dsq.Query{
+		KeysOnly: true,
+		Prefix:   providersKeyPrefix,
+	})
+
```
kevina: @whyrusleeping, using dstore.Query will work, but it is very slow.  The Query mechanism has a lot of overhead.  The data is being sent to you via an unbuffered channel in a separate go routine. In my own filestore code I was able to get a modest speedup by increasing the channel buffer size used the query, but it was still slow (speed up of 2 to 3x).  By querying the leveldb directly I was able to get a 10-12x speedup when doing a `filestore ls`.  (See ipfs-filestore/go-ipfs#10) I do not know how bad it will hurt performance here but it is something to keep it mind if you notice a slowdown after they code is deployed.

whyrusleeping: @kevina even with KeysOnly set to true? Now that you mention this, i think i'll throw in some timers.

kevina: @whyrusleeping Yes, in fact that is how I was using Query.

Kubuxu: @whyrusleeping would it be possible to improve the perf of that call, I depend on it in bloom PR (with AllKeysChan), better perf on it, shorter bloom build time.

kevina: As a quick hack you can try and increase the channel buffer size here https://github.com/ipfs/go-datastore/blob/master/query/query.go#L185.

A more long term solution might be to rewrite the query interface not use a direct iterator rather then using Goroutines, however Disk IO may nullify this benefit.

kevina: See https://ewencp.org/blog/golang-iterators/ for some performance comparisons.

kevina:
SGTM.  As far as I can tell I don't see any real problems.

routing/dht/providers/providers.go
whyrusleeping: Yeah, using a channel as an iterator sucks. If one of you wants to work on improving the perf of query that would be great. 

We could change the interface to not use a channel, and have it instead just return the next value directly. Then on top of that we could provide a method for turning the direct query result into a channel buffered one for usecases that need it

whyrusleeping:
Choo Choo!

routing/dht/providers/providers.go
jbenet: :+1: for not using channels there

jbenet: Actually, you can avoid using a query entirely if you store the records as ipfs objects properly, and only use leveldb to store `key -> hash_of_providers_object`, similar to how the pinset is stored

Issue #2833, opened 2016-06-10
Windows: makefile doesn't understand path quirk


djdv:
Previously I have used `make` from the msys2 project inside of a native Windows command line shell to build IPFS, pull https://github.com/ipfs/go-ipfs/pull/2808 seems to have broken my ability to build this way. This is more than likely do to the way msys and cygwin convert Windows paths i.e. C:\tmp -> /c/tmp.

![make break](https://cloud.githubusercontent.com/assets/13862850/15955819/4fa977b2-2eb2-11e6-948b-7338b8190b5c.png)

Within an msys2 shell I can build again just fine after setting up Go paths, or installing a second copy of Go inside of the msys environment.

![make break 2](https://cloud.githubusercontent.com/assets/13862850/15955823/53757d1e-2eb2-11e6-876b-64285eea235b.png)

I'm not sure how many people this effects but I don't think it's too uncommon for people to use the native shell and natively installed version of Go if they can.

lgierth:
cc @kalabe got a good idea how to fix building on windows? If it's too tricky I think we should revert #2808 and instead reconsider how we're organizing our build.

djdv:
@lgierth  I think @karalabe is the intended name, kalabe is someone else. Pinging for above.

Kubuxu:
It was reverted, should we close it?

djdv:
I would say yes, but a separate issue should probably be opened for multiple GOPATH paths if that's still desired.

Issue #2851, opened 2016-06-13
Commit  breaks 'make build' on Cygwin


jefft0:
I'm building on Cygwin. I get the following error after the recent change to Makefile in commit https://github.com/ipfs/go-ipfs/commit/ff75bc0318579fa743a35d316c61d2f8c4ab92ce#diff-a723d88d463c5bcca77d3d10a2b337bd 

    go-ipfs must be built from within your $GOPATH directory.
    expected 'C:/work/go/src/github.com/ipfs/go-ipfs' but got '/cygdrive/c/work/go/src/github.com/ipfs/go-ipfs'

When I revert the changes to Makefile and bin/check_go_path, then `make build` works again.

lgierth:
Hey thanks for reporting this and sorry we broke it :) Closing in favor of #2833

PR #2862, opened 2016-06-17
Revert "Makefile, bin: Support multiple GOPATH components"

lgierth:
This reverts commit ff75bc0318579fa743a35d316c61d2f8c4ab92ce,
which broke building on windows. See ipfs/go-ipfs#2833.

cc @karalabe @djdv @jefft0

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,7 +41,7 @@ bin/gx-go-v%:
 gx_check: ${gx_bin} ${gx-go_bin}
 
 path_check:
-	@bin/check_go_path $(realpath $(shell pwd)) $(realpath $(addsuffix /src/github.com/ipfs/go-ipfs,$(subst :, ,$(GOPATH))))
+	@bin/check_go_path $(realpath $(shell pwd)) $(realpath $(GOPATH)/src/github.com/ipfs/go-ipfs)
 
 deps: go_check gx_check path_check
 	${gx_bin} --verbose install --global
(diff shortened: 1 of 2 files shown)


Kubuxu:
Random test failure :D :smiley:

djdv:
Works on my machine. 👌

However, I don't know if this revert is necessary. The recommended ways of building on Windows (https://github.com/ipfs/go-ipfs/blob/master/docs/windows.md) are unaffected by https://github.com/ipfs/go-ipfs/commit/ff75bc0318579fa743a35d316c61d2f8c4ab92ce, the only broken case is when using `make` outside of Cygwin or an msys terminal (i.e. using the native Windows command line).

Reverting would obviously break multipath support, so It may be better to keep that in for now and fix native Windows building later on top of that commit rather than reverting.

Thoughts on this?

jefft0:
It was broken for me inside Cygwin, as I reported https://github.com/ipfs/go-ipfs/issues/2851 .

Kubuxu:
Does it only affect native terminal? If yes them I am also against it. cygwin tools `make` are supposed to be used in cygwin terminal and environment.

jefft0:
It was broken inside a Cygwin terminal.

djdv:
Disregard that, thanks @jefft0, I had not seen the breakage in Cygwin.

whyrusleeping:
@lgierth do we have consensus here? Is this good to merge?

lgierth:
Yes I think there's consensus

PR #2863, opened 2016-06-17
implement some simple dht request read timeouts

whyrusleeping:
Very occasionally the remote side will stop sending on a stream, not close it, and not close the underlying tcp connection. This causes annoying hangs

License: MIT


Kubuxu:
LGTM, as we talked on IRC, it isn't the best solution possible but complexity of others makes good.

whyrusleeping:
running this on my computer, i got one timeout logged. And unfortunately (or fortunately?) there arent any correlated yamux errors

jbenet:
We should come up with very strict guidelines on what it's ok to add this sort of timeout, and how to make it configurable. this is the sort of thing that doesn't scale with the physical size of the network (eg across planets).

PR #2867, opened 2016-06-18
Add way to check if node is online or offline

Kubuxu:
As requested by @lgierth.

@RichardLitt this changes API.

### DIFF
--- a/core/commands/sysdiag.go
+++ b/core/commands/sysdiag.go
@@ -44,8 +44,13 @@ Prints out information about your computer to aid in easier debugging.
 			res.SetError(err, cmds.ErrNormal)
 			return
 		}
+		node, err := req.InvocContext().GetNode()
+		if err != nil {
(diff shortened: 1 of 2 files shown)


whyrusleeping:
LGTM

jbenet:
@Kubuxu changed the API in what way? this type of thing is important to put directly into the commit messages, not just the PR, and to be very explicit about what, to be super clear with everyone. cc @whyrusleeping too about this.

jbenet:
(i think it just _added_ a feature, but just checking)

Kubuxu:
Yes, it is just an added feature, we are pining Richard like that to notify him that this PR includes something that requires update in http-api-spec.

jbenet:
- We should codify API changes directly into commits.
- Makes it easy to include in changelog
- remember many people wont have the full github PR comment history to understand from.

PR #2871, opened 2016-06-18
Added variable latency delay, normal and uniform based

rikonor:
Hey,
I saw the todo for random delays and thought I'd give it a go.

License: MIT
Signed-off-by: Or rikonor@gmail.com

### DIFF
--- a/thirdparty/delay/delay.go
+++ b/thirdparty/delay/delay.go
@@ -1,9 +1,12 @@
 package delay
 
 import (
+	"math/rand"
 	"sync"
 	"time"
 )
+
+var sharedRNG = rand.New(rand.NewSource(time.Now().UnixNano()))
 
 // Delay makes it easy to add (threadsafe) configurable delays to other
 // objects.
@@ -23,8 +26,6 @@ type delay struct {
 	t time.Duration
 }
 
-// TODO func Variable(time.Duration) D returns a delay with probablistic latency
-
 func (d *delay) Set(t time.Duration) time.Duration {
 	d.l.Lock()
 	defer d.l.Unlock()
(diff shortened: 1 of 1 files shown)


thirdparty/delay/delay.go
```diff
@@ -44,3 +43,53 @@ func (d *delay) Get() time.Duration {
...
+type variableUniform struct {
+	delay
+	d   time.Duration // max delta
+	rng *rand.Rand
+}
+
+func (d *variableUniform) Wait() {
+	d.l.RLock()
+	defer d.l.RUnlock()
+	randomDelay := time.Duration(d.rng.Float64() * float64(d.d))
```
Kubuxu: I think unlock should be here so multiple goroutines can Wait at the same time.

thirdparty/delay/delay.go
```diff
@@ -44,3 +43,53 @@ func (d *delay) Get() time.Duration {
...
 	return d.t
 }
+
+// VariableNormal is a delay following a normal distribution
+// Notice that to implement the D interface Set can only change the mean delay
+// the standard deviation is set only at initialization
+func VariableNormal(t, std time.Duration) D {
+	v := &variableNormal{
+		std: std,
+		rng: rand.New(rand.NewSource(int64(time.Now().Nanosecond()))),
```
Kubuxu: Using one random source per one delay instance runs risk of it being used just one time.
Pseudo Random Generators used just one time will not generate uniformly distributed random numbers.

I would recommend using one package wide random generator.

rikonor:
@Kubuxu, thanks for the feedback.
I added a shared rng instance, and left the possibility of providing your own.
Regarding the locks, two goroutines can wait at the same time because they'll both be able to acquire a read lock.

Kubuxu:
Right, I missed that.
Would you mind amending triangular brackets around you events just to make all tests happy.

Apart from that, LGTM.

rikonor:
@Kubuxu - yep, fixed that. Thanks!

Kubuxu:
LGTM

thirdparty/delay/delay.go
```diff
@@ -1,10 +1,13 @@
 package delay
 
 import (
+	"math/rand"
 	"sync"
 	"time"
 )
 
+var sharedRNG = rand.New(rand.NewSource(int64(time.Now().Nanosecond())))
```
whyrusleeping: instead of `.Nanosecond()` we could use `.UnixNano()`, saves a cast

whyrusleeping:
one comment, then LGTM

rikonor:
@whyrusleeping, using `UnixNano` now instead of `Nanosecond` and a cast.

whyrusleeping:
cool, LGTM. Thanks!

lgierth:
This doesn't seem like it's used anywhere?

rikonor:
@lgierth I don't believe that it is. I was just looking for something to do and ended up implementing that. Hope that it can help at some point in the future.

Kubuxu:
We currently do local network dialups in bursts. This is somewhere we can use it.

jbenet:
Maybe this type of thing could be useful as a libp2p transport?

Kubuxu:
We might move it to utility package to use it in libp2p to not send dials in bursts, it isn't huge issues right now.

lgierth:
@jbenet aren't you having a day off? Go out and play :)

Issue #2698, opened 2016-05-15
Cannot restart daemon after files API ops and `repo gc`


kevina:
```
$ ipfs version
ipfs version 0.4.1
$ ipfs daemon &
$ mkdir /tmp/adir
$ echo "hello world" > /tmp/adir/hello.txt
$ ipfs add -r /tmp/adir
added QmT78zSuBmuS4z925WZfrqQ1qHaJ56DQaTfyMUF7F8ff5o adir/hello.txt
added QmfLiVjH2vujCVP2e75zyzBYmpcjktmDeU1YBz6Ct8BBsc adir
$ ipfs files cp /ipfs/QmfLiVjH2vujCVP2e75zyzBYmpcjktmDeU1YBz6Ct8BBsc /adir
$ ipfs pin rm QmfLiVjH2vujCVP2e75zyzBYmpcjktmDeU1YBz6Ct8BBsc
$ ipfs repo gc
$ kill %1
$ ipfs daemon
```
The final command `ipfs daemon` will hang.  When I press Ctrl-c I get:
```
Received interrupt signal, shutting down...
(Hit ctrl-c again to force-shutdown the daemon.)
10:41:21.661 ERROR   cmd/ipfs: error from node construction: error loading filesroot from DAG: Failed to get block for QmWikN9opNVqoFhD8D4ERt6QSBFBaRwvVoc3wdfYJExpLh: context canceled daemon.go:257
Error: error loading filesroot from DAG: Failed to get block for QmWikN9opNVqoFhD8D4ERt6QSBFBaRwvVoc3wdfYJExpLh: context canceled
```
This is very likely related to #2697

kevina:
Note after the steps above, nothing will work even when offline:
```
$ ipfs files ls
Error: error loading filesroot from DAG: merkledag: not found
$ ipfs files flush
Error: error loading filesroot from DAG: merkledag: not found
$ ipfs add -r /tmp/adir/
Error: error loading filesroot from DAG: merkledag: not found
```
To fix this I need to somehow remove the key "/local/filesroot" from ".ipfs/datastore/". Or just start over by killing the ".ipfs/datastore" directory or if I don't care about the cache or my node id just kill ".ipfs" and do an `ipfs init`.

Kubuxu:
Can you also give us output of `ipfs version --commit`.

kevina:
```
$ ipfs version --commit
ipfs version 0.4.1-
```

[...]

PR #2872, opened 2016-06-18
Add Files API root as best-effort pin

### LINKED ISSUE #2697: Files added via files API get garbage collected
It seams that there is nothing protecting files added via the files API from being garbage collected:
```
$ ipfs version
ipfs version 0.4.1
$ echo "hello world"  | ipfs files write --create /hello.txt
$ ipfs files stat  /hello.txt
QmVib14uvPnCP73XaCDpwugRuwfTsVbGyWbatHAmLSdZUS
$ ipfs repo gc
...
removed QmVib14uvPnCP73XaCDpwugRuwfTsVbGyWbatHAmLSdZUS
...
$ ipfs cat QmVib14uvPnCP73XaCDpwugRuwfTsVbGyWbatHAmLSdZUS
Error: merkledag: not found
```
And I can't even remove the file:
```
$ ipfs files rm /hello.txt
Error: merkledag: not found
```
(but for some reason `ipfs files rm -r /hello.txt` works as a workaround).

Am I missing something?

kevina:
This fixes #2697 and #2698.

### DIFF
--- a/merkledag/merkledag.go
+++ b/merkledag/merkledag.go
@@ -357,16 +357,20 @@ func (t *Batch) Commit() error {
 // EnumerateChildren will walk the dag below the given root node and add all
 // unseen children to the passed in set.
 // TODO: parallelize to avoid disk latency perf hits?
-func EnumerateChildren(ctx context.Context, ds DAGService, root *Node, set key.KeySet) error {
+func EnumerateChildren(ctx context.Context, ds DAGService, root *Node, set key.KeySet, bestEffort bool) error {
 	for _, lnk := range root.Links {
 		k := key.Key(lnk.Hash)
 		if !set.Has(k) {
 			set.Add(k)
 			child, err := ds.Get(ctx, k)
 			if err != nil {
-				return err
+				if bestEffort && err == ErrNotFound {
+					continue
+				} else {
+					return err
+				}
 			}
-			err = EnumerateChildren(ctx, ds, child, set)
+			err = EnumerateChildren(ctx, ds, child, set, bestEffort)
 			if err != nil {
 				return err
 			}
--- a/pin/gc/gc.go
+++ b/pin/gc/gc.go
@@ -17,18 +17,19 @@ var log = logging.Logger("gc")
 // GC performs a mark and sweep garbage collection of the blocks in the blockstore
 // first, it creates a 'marked' set and adds to it the following:
 // - all recursively pinned blocks, plus all of their descendants (recursively)
+// - bestEffortRoots, plus all of its descendants (recursively)
 // - all directly pinned blocks
 // - all blocks utilized internally by the pinner
 //
 // The routine then iterates over every block in the blockstore and
 // deletes any block that is not found in the marked set.
-func GC(ctx context.Context, bs bstore.GCBlockstore, pn pin.Pinner) (<-chan key.Key, error) {
+func GC(ctx context.Context, bs bstore.GCBlockstore, pn pin.Pinner, bestEffortRoots []key.Key) (<-chan key.Key, error) {
 	unlocker := bs.GCLock()
 
 	bsrv := bserv.New(bs, offline.Exchange(bs))
 	ds := dag.NewDAGService(bsrv)
 
-	gcs, err := ColoredSet(ctx, pn, ds)
+	gcs, err := ColoredSet(ctx, pn, ds, bestEffortRoots)
 	if err != nil {
 		return nil, err
 	}
@@ -69,7 +70,7 @@ func GC(ctx context.Context, bs bstore.GCBlockstore, pn pin.Pinner) (<-chan key.
 	return output, nil
 }
 
-func Descendants(ctx context.Context, ds dag.DAGService, set key.KeySet, roots []key.Key) error {
+func Descendants(ctx context.Context, ds dag.DAGService, set key.KeySet, roots []key.Key, bestEffort bool) error {
 	for _, k := range roots {
 		set.Add(k)
 		nd, err := ds.Get(ctx, k)
@@ -78,7 +79,7 @@ func Descendants(ctx context.Context, ds dag.DAGService, set key.KeySet, roots [
 		}
 
 		// EnumerateChildren recursively walks the dag and adds the keys to the given set
-		err = dag.EnumerateChildren(ctx, ds, nd, set)
+		err = dag.EnumerateChildren(ctx, ds, nd, set, bestEffort)
 		if err != nil {
 			return err
 		}
@@ -87,11 +88,16 @@ func Descendants(ctx context.Context, ds dag.DAGService, set key.KeySet, roots [
 	return nil
 }
 
-func ColoredSet(ctx context.Context, pn pin.Pinner, ds dag.DAGService) (key.KeySet, error) {
+func ColoredSet(ctx context.Context, pn pin.Pinner, ds dag.DAGService, bestEffortRoots []key.Key) (key.KeySet, error) {
 	// KeySet currently implemented in memory, in the future, may be bloom filter or
 	// disk backed to conserve memory.
 	gcs := key.NewKeySet()
-	err := Descendants(ctx, ds, gcs, pn.RecursiveKeys())
+	err := Descendants(ctx, ds, gcs, pn.RecursiveKeys(), false)
+	if err != nil {
+		return nil, err
+	}
+
+	err = Descendants(ctx, ds, gcs, bestEffortRoots, true)
 	if err != nil {
 		return nil, err
 	}
@@ -100,7 +106,7 @@ func ColoredSet(ctx context.Context, pn pin.Pinner, ds dag.DAGService) (key.KeyS
 		gcs.Add(k)
 	}
 
-	err = Descendants(ctx, ds, gcs, pn.InternalPins())
+	err = Descendants(ctx, ds, gcs, pn.InternalPins(), false)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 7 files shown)


merkledag/merkledag.go
```diff
@@ -375,6 +375,29 @@ func EnumerateChildren(ctx context.Context, ds DAGService, root *Node, set key.K
...
 
+// Like EnumerateChildren but will not fail if Get failes due to key
+// not existing
+func EnumerateChildrenBestEffort(ctx context.Context, ds DAGService, root *Node, set key.KeySet) error {
+	for _, lnk := range root.Links {
+		k := key.Key(lnk.Hash)
+		if !set.Has(k) {
+			set.Add(k)
+			child, err := ds.Get(ctx, k)
+			if err == ErrNotFound {
```
whyrusleeping: lets do something like
```go
switch err {
case ErrNotFound:
//
default:
return err
case nil:
break
}
```

pin/gc/gc.go
```diff
@@ -17,18 +18,19 @@ var log = logging.Logger("gc")
...
 // first, it creates a 'marked' set and adds to it the following:
 // - all recursively pinned blocks, plus all of their descendants (recursively)
+// - the root directory for the files API, plus all of its descendants (recursively)
 // - all directly pinned blocks
 // - all blocks utilized internally by the pinner
 //
 // The routine then iterates over every block in the blockstore and
 // deletes any block that is not found in the marked set.
-func GC(ctx context.Context, bs bstore.GCBlockstore, pn pin.Pinner) (<-chan key.Key, error) {
+func GC(ctx context.Context, bs bstore.GCBlockstore, pn pin.Pinner, filesRoot *mfs.Root) (<-chan key.Key, error) {
```
whyrusleeping: instead of passing in an mfs.Root here, i think the better option is just to pass in a keyset. That way we keep it generic and don't build weird dependencies between packages like this

kevina: @whyrusleeping in commit ed601f6 I passed in the dag Node inside of the files root, this eliminated the extra dependency.  In the second commit 174d2ce I passed in an array of keys to use as the root set for the best effort pins; this is a bit more general but involves a bit more code.  Let me know what version you like best.

kevina:
@whyrusleeping see my line note on your second comment.  There are two versions for you to chose from.

Please do not merge this as is.  I will squash this all down to one commit once we decide what works best.

whyrusleeping:
@kevina i think the latter option of passing the array of best effort root keys is the better of the two for now. My other thought was to enumerate the best effort set before calling gc, and pass that into the gc call as a 'extra keys that arent garbage' parameter. But i think that might be a little more difficult and uglier overall

kevina:
@whyrusleeping, okay, ready to merge once you are happy with it.

test/sharness/t0252-files-gc.sh
```diff
@@ -0,0 +1,36 @@
...
+test_description="test how the unix files api interacts with the gc"
+
+. lib/test-lib.sh
+
+test_init_ipfs
+
+test_expect_success "object not removed after gc" '
+  echo "hello world"  | ipfs files write --create /hello.txt &&
+  ipfs repo gc &&
+  ipfs cat QmVib14uvPnCP73XaCDpwugRuwfTsVbGyWbatHAmLSdZUS
```
whyrusleeping: we should also add another `test_expect_success` to make sure that `/hello.txt` is still accessible through the files api here

kevina: @whyrusleeping Added.

whyrusleeping:
This LGTM. @Kubuxu wanna take a look over too?

Kubuxu:
SGTM

test/sharness/t0252-files-gc.sh
jbenet: So... if this is "best effort", under what conditions should gc remove these, and can we make sure to test that here?

pin/gc/gc.go
```diff
@@ -69,7 +70,7 @@ func GC(ctx context.Context, bs bstore.GCBlockstore, pn pin.Pinner) (<-chan key.
 	return output, nil
 }
 
-func Descendants(ctx context.Context, ds dag.DAGService, set key.KeySet, roots []key.Key) error {
+func Descendants(ctx context.Context, ds dag.DAGService, set key.KeySet, roots []key.Key, bestEffort bool) error {
```
jbenet: i strongly dislike how this is implemented. Adding `bestEffort` here to this function ruins the whole abstraction. You can implement this "best effort" notion _outside_ of this function, just in the GC function.

jbenet:
- The idea of "best effort" pins is good and we already want to incorporate it into the pinning model.
- The idea of making the files api "best effort" makes sense. it solves the problems we had with making it proper full pins.
- That said, the way this PR is implemented is not great. It muddles the abstractions, threading in a "best effort" flag into things, instead of just treating that set of roots as just another set to mark and consider.
- It's also not clear when GC will actually remove these "best effort" pins.
- This is not good quality, and I'm sad this merged as is.
- To be clear, the idea is very good, and we SHOULD have such an abstraction. It should be implemented better though.

test/sharness/t0252-files-gc.sh
kevina: All "best effort" means is not to abort if a child is unreadable.  For the GC to remove the object will need to be detached somehow from the MFS Root.

pin/gc/gc.go
kevina: @jbenet in this function (Descendants) and EnumerateChildren all that bestEffort means is to not abort if a child is unreadable.  In order to collect the set of keys to keep on a "best effort" bases I needed a way to to collected all the children of the initial best effort set that are local on the node.

I am not entirely clear what abstraction I am ruining so I am not 100% sure how to fix this.  I could create a new function to find the Descendants of a keySet that doesn't abort on an error, but to me it seamed better to extend the functionally of the existing function.

@whyrusleeping I could appreciate any insights you can offer.

kevina:
So to be clear all "best effort" means (as agreed to by @whyrusleeping and @Kubuxu) is to simply keep all children of the best effort pin that are reachable from the local node.  (In other words if a child in unavailable simply skip it, rather than aborting or attempting to download it).  Nothing More.  The GC will never remove "best effort" pins.

jbenet:
(Thinking out loud)

Oh wow, that means something totally different, then. "best effort" in computer science tends to mean "up to best abilities, but not unreliable". Specifically, ["best effort" in networks](https://en.wikipedia.org/wiki/Best-effort_delivery). Applied to pins, i took it to mean "pin a subgraph while you can, but GC if you need to", sort of like a "gc these last" type of thing.

What you describe ("pin the graph objects currently present in the node") 

It's a bit concerning, too because it has unpredictable guarantees and consequences, in a very different way than "unreliable (best effort)"-- because how much is pinned forever depends entirely on order of events, which is very hard to reason about.

- What happens if more objects in that graph are downloaded, as a result of other operations? are they suddenly pinned too? forever? and why forever?
- It seems to me this is trying to approximate some use case that is better served by other mechanisms. For example, i might "pin local objects" that are part of a VM i want to boot, leaving the rest of the objects to be fetched later and gc-ed differently (i.e. i want to pin only the critical parts). Or similarly, I might pin the "index sub graph" and first few objects in a video, so i can always start playing it quickly and seek quickly, but let the retrieval of the big data blocks to when they're needed.
- Though in these use cases, i dont actually want the mechanism you describe, i want a different pin which can be configured to pin a chosen subgraph (i.e. what i really want is a pin with an associated IPLD selector).

As it is now, I think your "objects that are currently local" pins can be turned into a couple of tools, and use regular pin sets:
- first, create a tool that list all local refs for a graph. eg.  `ipfs refs local [-r] <root>` or `ipfs refs [-r] --local <root>` which would list out all refs from `<root>`. Can pin those directly.
- Potentially create a recursive pin with a depth parameter (again, covered by IPLD selectors later), so all you need to do is create one object that points to all the local refs you listed, and pin that with the appropriate depth.

Thinking about it another way, let's rename "best effort" into "incremental". Because that's what it really is, right? it's an incremental pin that over time will pin more and more of a subgraph, as it's accessed and retrieved. This has different use cases than i mentioned above (eg "incrementally pin" all my local maps' area in a google maps like thing, or "incrementally pin" a dataset i'm exploring, or a dungeon in a dungeon crawler). Is this more what you mean?

Would be very useful to me to hear what use cases you have in mind for this.

kevina:
@jbenet, please see #2698 for context.  @whyrusleeping actually came up with the name "best effort" and I just used it.

test/sharness/t0252-files-gc.sh
whyrusleeping: @kevina want to take this one on and add a few more tests?

- Test that the read test below outputs the right text
- Exercise a situation that uses the best effort pin: 
  - Add a directory containing a file with `--pin=false`
  - pin the hash directly
  - run a gc so that file is no longer accessible
  - `ipfs files cp <dirhash> /somewhere` (this should work just fine)
  - `ipfs repo gc` should succeed
  - re-add the file that was originally in that directory with `--pin=false`
  - run `ipfs repo gc` again, that file shouldnt be removed.

kevina: OK Will do.

kevina: > @kevina want to take this one on and add a few more tests?

@whyrusleeping See #3150

whyrusleeping:
@jbenet this isnt a separate pin type. All this is is a way to make the files api space not break when you run a GC, but also not require that everything in the files API be local.

PR #2874, opened 2016-06-18
gateway: clean up its surface, and remove BlockList

lgierth:
This patch is in preparation for the gateway's extraction.

It's interesting to trace technical debt back to its
origin, understanding the circumstances in which it
was introduced and built up, and then cutting it back
at exactly the right places.

- Clean up the gateway's surface
  The option builder GatewayOption() now takes only
  arguments which are relevant for HTTP handler muxing,
  i.e. the paths where the gateway should be mounted.
  All other configuration happens through the
  GatewayConfig object.

- Remove BlockList
  I know why this was introduced in the first place,
  but it never ended up fulfilling that purpose.
  Somehow it was only ever used by the API server,
  not the gateway, which really doesn't make sense.
  It was also never wired up with CLI nor fs-repo.
  Eventually @krl started punching holes into it
  to make the Web UI accessible.

- Remove --unrestricted-api
  This was holes being punched into BlockList too,
  for accessing /ipfs and /ipn on the API server.
  With BlockList removed and /ipfs and /ipns freely
  accessible, putting this option out of action
  is safe. With the next major release,
  the option can be removed for good.

License: MIT

### DIFF
--- a/cmd/ipfs/daemon.go
+++ b/cmd/ipfs/daemon.go
@@ -8,7 +8,6 @@ import (
 	_ "net/http/pprof"
 	"os"
 	"sort"
-	"strings"
 	"sync"
 
 	"gx/ipfs/QmPpRcbNUXauP3zWZ1NJMLWpe4QnmEHrd2ba2D3yqWznw7/go-multiaddr-net"
@@ -135,7 +134,7 @@ Headers.
 		cmds.BoolOption(writableKwd, "Enable writing objects (with POST, PUT and DELETE)").Default(false),
 		cmds.StringOption(ipfsMountKwd, "Path to the mountpoint for IPFS (if using --mount). Defaults to config setting."),
 		cmds.StringOption(ipnsMountKwd, "Path to the mountpoint for IPNS (if using --mount). Defaults to config setting."),
-		cmds.BoolOption(unrestrictedApiAccessKwd, "Allow API access to unlisted hashes").Default(false),
+		cmds.BoolOption(unrestrictedApiAccessKwd, "This option has no effect since v0.4.3").Default(false),
 		cmds.BoolOption(unencryptTransportKwd, "Disable transport encryption (for debugging protocols)").Default(false),
 		cmds.BoolOption(enableGCKwd, "Enable automatic periodic repo garbage collection").Default(false),
 		cmds.BoolOption(adjustFDLimitKwd, "Check and raise file descriptor limits if needed").Default(false),
@@ -364,33 +363,11 @@ func serveHTTPApi(req cmds.Request) (error, <-chan error) {
 	apiMaddr = apiLis.Multiaddr()
 	fmt.Printf("API server listening on %s\n", apiMaddr)
 
-	unrestricted, _, err := req.Option(unrestrictedApiAccessKwd).Bool()
-	if err != nil {
-		return fmt.Errorf("serveHTTPApi: Option(%s) failed: %s", unrestrictedApiAccessKwd, err), nil
-	}
-
-	apiGw := corehttp.NewGateway(corehttp.GatewayConfig{
-		Writable: true,
-		BlockList: &corehttp.BlockList{
-			Decider: func(s string) bool {
-				if unrestricted {
-					return true
-				}
-				// for now, only allow paths in the WebUI path
-				for _, webuipath := range corehttp.WebUIPaths {
-					if strings.HasPrefix(s, webuipath) {
-						return true
-					}
-				}
-				return false
-			},
-		},
-	})
 	var opts = []corehttp.ServeOption{
 		corehttp.MetricsCollectionOption("api"),
 		corehttp.CommandsOption(*req.InvocContext()),
 		corehttp.WebUIOption,
-		apiGw.ServeOption(),
+		corehttp.GatewayOption("/ipfs", "/ipns"),
 		corehttp.VersionOption(),
 		defaultMux("/debug/vars"),
 		defaultMux("/debug/pprof/"),
@@ -452,8 +429,8 @@ func serveHTTPGateway(req cmds.Request) (error, <-chan error) {
 	if err != nil {
 		return fmt.Errorf("serveHTTPGateway: req.Option(%s) failed: %s", writableKwd, err), nil
 	}
-	if !writableOptionFound {
-		writable = cfg.Gateway.Writable
+	if writableOptionFound {
+		cfg.Gateway.Writable = writable
 	}
 
 	gwLis, err := manet.Listen(gatewayMaddr)
@@ -474,7 +451,7 @@ func serveHTTPGateway(req cmds.Request) (error, <-chan error) {
 		corehttp.CommandsROOption(*req.InvocContext()),
 		corehttp.VersionOption(),
 		corehttp.IPNSHostnameOption(),
-		corehttp.GatewayOption(writable, cfg.Gateway.PathPrefixes),
+		corehttp.GatewayOption("/ipfs", "/ipns"),
 	}
 
 	if len(cfg.Gateway.RootRedirect) > 0 {
--- a/core/corehttp/gateway_handler.go
+++ b/core/corehttp/gateway_handler.go
@@ -36,12 +36,12 @@ type gatewayHandler struct {
 	config GatewayConfig
 }
 
-func newGatewayHandler(node *core.IpfsNode, conf GatewayConfig) (*gatewayHandler, error) {
+func newGatewayHandler(node *core.IpfsNode, conf GatewayConfig) *gatewayHandler {
 	i := &gatewayHandler{
 		node:   node,
 		config: conf,
 	}
-	return i, nil
+	return i
 }
 
 // TODO(cryptix):  find these helpers somewhere else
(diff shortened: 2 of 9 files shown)


cmd/ipfs/daemon.go
```diff
@@ -452,8 +429,8 @@ func serveHTTPGateway(req cmds.Request) (error, <-chan error) {
 	if err != nil {
 		return fmt.Errorf("serveHTTPGateway: req.Option(%s) failed: %s", writableKwd, err), nil
 	}
-	if !writableOptionFound {
-		writable = cfg.Gateway.Writable
+	if writableOptionFound {
+		cfg.Gateway.Writable = writable
```
Kubuxu: As we are using defaults, the `if` block can be removed and just assign to `cfg.Gateway.Writable`.

cmd/ipfs/daemon.go
```diff
@@ -135,7 +134,7 @@ Headers.
 		cmds.BoolOption(writableKwd, "Enable writing objects (with POST, PUT and DELETE)").Default(false),
 		cmds.StringOption(ipfsMountKwd, "Path to the mountpoint for IPFS (if using --mount). Defaults to config setting."),
 		cmds.StringOption(ipnsMountKwd, "Path to the mountpoint for IPNS (if using --mount). Defaults to config setting."),
-		cmds.BoolOption(unrestrictedApiAccessKwd, "Allow API access to unlisted hashes").Default(false),
+		cmds.BoolOption(unrestrictedApiAccessKwd, "This option has no effect since v0.4.3").Default(false),
```
whyrusleeping: is this true?

whyrusleeping:
overall this LGTM, its cleaner.

cmd/ipfs/daemon.go
lgierth: The option itself is still there, but not being used anymore, lines 365 ff. were the only usage. I left the option there for now so that you don't get `Error: Unrecognized option 'unrestricted-api'` if you were really using it. The new behaviour is as if `--unrestricted-api` was always true.

cmd/ipfs/daemon.go
lgierth: Mh okay, we shouldn't be using `Default(false)` with this one. From what I can tell, `--writable` *if present* is supposed to override `cfg.Gateway.Writable`.

lgierth:
> Mh okay, we shouldn't be using Default(false) with this one. From what I can tell, --writable if present is supposed to override cfg.Gateway.Writable.

Gonna investigate this before merging, hold off

cmd/ipfs/daemon.go
lgierth: So... the second return value (writableOptionFound) is separate from defaults -- it tells specifically whether the option was passed on the CLI.

So this assumption is still true and everything's good as-is:

> From what I can tell, --writable if present is supposed to override cfg.Gateway.Writable.

In other words, the default value for --writable is never used -- only if it has actually been passed.

lgierth:
> Gonna investigate this before merging, hold off

All good - RFM!

cmd/ipfs/daemon.go
jbenet: > The new behaviour is as if --unrestricted-api was always true.

I believe this has since been reverted. But this comment should've raised flags.

test/sharness/t0110-gateway.sh
```diff
@@ -32,10 +32,6 @@ test_expect_success "GET IPFS path output looks good" '
   rm actual
 '
 
-test_expect_success "GET IPFS path on API forbidden" '
-  test_curl_resp_http_code "http://127.0.0.1:$apiport/ipfs/$HASH" "HTTP/1.1 403 Forbidden"
-'
-
```
jbenet: I thought this entire PR was reverted. I guess it was not because this test is not in master. 

cc @lgierth we need this back, i think.

test/sharness/t0061-daemon-opts.sh
```diff
@@ -11,20 +11,11 @@ test_description="Test daemon command"
...
 
-test_expect_success 'api gateway should be unrestricted' '
-  echo "hello mars :$gwyaddr :$apiaddr" >expected &&
-  HASH=$(ipfs add -q expected) &&
-  curl -sfo actual1 "http://$gwyaddr/ipfs/$HASH" &&
-  curl -sfo actual2 "http://$apiaddr/ipfs/$HASH" &&
-  test_cmp expected actual1 &&
-  test_cmp expected actual2
-'
-
```
jbenet: same with this test

misc/completion/ipfs-completion.bash
```diff
@@ -104,7 +104,7 @@ _ipfs_config_show()
 _ipfs_daemon()
 {
     _ipfs_comp "--init --routing= --mount --writable --mount-ipfs= \
-        --mount-ipns= --unrestricted-api --disable-transport-encryption \
```
jbenet: this needs to come back too.

core/corehttp/gateway_handler.go
```diff
@@ -152,12 +152,6 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
 		ipnsHostname = true
 	}
 
-	if i.config.BlockList != nil && i.config.BlockList.ShouldBlock(urlPath) {
-		w.WriteHeader(http.StatusForbidden)
-		w.Write([]byte("403 - Forbidden"))
-		return
-	}
```
jbenet: this blocking is no longer happening in master. I think this is a bug.

jbenet:
Were there other changes to gateway code in this release?

If not, I think i would feel safer if all gateway changes were rebased out and placed at the end together so we can analyze them very clearly. (i guess we can diff the whole release in these files, but making sure this is safe may be worth rewriting history over...)

core/corehttp/gateway_handler.go
lgierth: did you try on :5001? It never happen on :8080

test/sharness/t0110-gateway.sh
lgierth: test has been moved here https://github.com/ipfs/go-ipfs/blob/master/test/sharness/t0400-api-security.sh (because it's not a gateway test)

test/sharness/t0061-daemon-opts.sh
lgierth: same here, test has been moved here https://github.com/ipfs/go-ipfs/blob/master/test/sharness/t0400-api-security.sh (because it's not a gateway test)

misc/completion/ipfs-completion.bash
lgierth: ack right! thanks

lgierth:
It should have all been reverted by 0.4.3-rc1, as part of fixing the KeyThief vuln. Well not reverted, but reintroduced, in the case of restricted api and --unrestricted-api. #2949 and #2956

jbenet:
Right but my question is: how do we know the behaviors are all back to the
same? It may be worth reverting and doing the proper gateway extraction
separately, and all in a larger patch set? Not sure. I think this area
lacks tests to feel super comfortable with it.

core/corehttp/gateway_handler.go
lgierth: > > this blocking is no longer happening in master. I think this is a bug.
> did you try on :5001? It never happened on :8080

Nevermind, I think you mean that this call to `BlockList.ShouldBlock()` doesn't exist anymore. `BlockList` was completely removed, and we now bind the gateway http handler to the http multiplexer using individual explicit paths (from `corehttp.WebUIPaths`, e.g. `/ipfs/QmU3o9bvfenhTKhxUakbYrLDnZU7HezAVxPM6Ehjw9Xjqy`).

The new code is here: https://github.com/ipfs/go-ipfs/blob/6fdfaaf6e4783ae1be7b348e7a6bc0640982c7df/cmd/ipfs/daemon.go#L404-L415

lgierth: There are also proper comments around this very important piece of code now (see above) and for the tests too: https://github.com/ipfs/go-ipfs/blob/master/test/sharness/t0400-api-security.sh

lgierth:
> Right but my question is: how do we know the behaviors are all back to the same? It may be worth reverting and doing the proper gateway extraction separately, and all in a larger patch set? Not sure. I think this area lacks tests to feel super comfortable with it.

But, but... I removed lots of code and the blocking logic is literally 7 lines of code now. [1] It wires `--unrestricted-api` to this logic directly, without any intermediary function calls. BlockList was actually a pretty hefty abstraction and made a very simple knob complicated . The tests are now in a place that makes sense, comments were added everywhere.

I think it's is a net win, making this thing much more obvious. Its obscurity was the reason why I was able to be pretty sure it was dead code, and introduce KeyThief.

[1] [daemon.go](https://github.com/ipfs/go-ipfs/blob/6fdfaaf6e4783ae1be7b348e7a6bc0640982c7df/cmd/ipfs/daemon.go#L404-L415)
[2] [t0400-api-security.sh](https://github.com/ipfs/go-ipfs/blob/6fdfaaf6e4783ae1be7b348e7a6bc0640982c7df/test/sharness/t0400-api-security.sh)

lgierth:
> and introduce KeyThief

well I not actually *introduced* it, but at least opened the gates. Anyway that's a different topic :)

core/corehttp/gateway_handler.go
lgierth: > `BlockList` was completely removed, and we now bind the gateway http handler to the http multiplexer using individual explicit paths

code here: https://github.com/ipfs/go-ipfs/blob/6fdfaaf6e4783ae1be7b348e7a6bc0640982c7df/core/corehttp/gateway.go#L32-L34

PR #2876, opened 2016-06-19
Gateway uses minimal IPFS Core API

lgierth:
This is a follow-up to ipfs/specs#85, which discusses an IPFS Core API. The immediate goal is to slim down the surface area between the gateway, and go-ipfs internals, so that the gateway can easily be extracted into its own package.

- (Disregard the first commit, it'll be rebased out once #2874 is merged.)
- 5e328bf starts a CoreAPI interface offering `Cat()` and `Ls()` alongside data structures and errors, as well as an implementation thereof.
- 8b6cf07 starts using the rump Core API for serving GET requests in the gateway.

A possible direction forward:

- The package `go-ipfs-core-api/interface` provides interfaces, data structures, and errors. Tools like ipget or pinbot can choose between implementations backed by the API server (:5001), or by a proper embedded IpfsNode.
- The package `go-ipfs-core-api` implements `go-ipfs-core-api/interface` backed by a proper IpfsNode. It depends on a lot of go-ipfs internals, like `core`, `merkledag`, `unixfs/io`.
- The existing package `go-ipfs-api` implements `go-ipfs-core-api/interface` backed by the API Server.
- `go-ipfs/core/commands` uses `go-ipfs-core-api` -- most of the logic there should live in the core-api.
- The gateway, FUSE, and whatever else there is also use either `go-ipfs-core-api` or `go-ipfs-api`.

### DIFF
--- a/core/coreapi/coreapi.go
+++ b/core/coreapi/coreapi.go
@@ -1,0 +1,60 @@
+package coreapi
+
+import (
+	"context"
+
+	core "github.com/ipfs/go-ipfs/core"
+	coreiface "github.com/ipfs/go-ipfs/core/coreapi/interface"
+	dag "github.com/ipfs/go-ipfs/merkledag"
+	path "github.com/ipfs/go-ipfs/path"
+	uio "github.com/ipfs/go-ipfs/unixfs/io"
+	cid "gx/ipfs/QmfSc2xehWmWLnwwYR91Y8QF4xdASypTFVknutoKQS3GHp/go-cid"
+)
+
+type UnixfsAPI struct {
+	Context context.Context
+	Node    *core.IpfsNode
+}
+
+func (api *UnixfsAPI) resolve(p string) (*dag.Node, error) {
+	pp, err := path.ParsePath(p)
+	if err != nil {
+		return nil, err
+	}
+
+	dagnode, err := core.Resolve(api.Context, api.Node, pp)
+	if err == core.ErrNoNamesys {
+		return nil, coreiface.ErrOffline
+	} else if err != nil {
+		return nil, err
+	}
+	return dagnode, nil
+}
+
+func (api *UnixfsAPI) Cat(p string) (coreiface.Reader, error) {
+	dagnode, err := api.resolve(p)
+	if err != nil {
+		return nil, err
+	}
+
+	r, err := uio.NewDagReader(api.Context, dagnode, api.Node.DAG)
+	if err == uio.ErrIsDir {
+		return nil, coreiface.ErrIsDir
+	} else if err != nil {
+		return nil, err
+	}
+	return r, nil
+}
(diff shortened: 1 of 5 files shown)


Kubuxu:
So my plan for core-api is to one-to-one map it to RPC API and totally separate it from go-ipfs (and js-ipfs). That is why the API itself should be simple, follow procedure call pattern. 

It would be great if we could work with js-ipfs on it.

Naming suggestion: replacing `Ls` with `ReadDir`, it is clearer, references readdir from C STD lib and more established.

lgierth:
We already have a Core API layed out though: https://github.com/ipfs/specs/issues/85

How would these relate to each other?

Kubuxu:
There isn't much in terms of specs there (and I don't know how much js team has done on it). I think we shouldn't repeat CLI api scheme (we already have working API like that) and design something extremely simplistic.

It is my view on the matter when it comes to creating and implementing new API.

condiosluzverde:
Perhaps the spec differs from ipfs-api in writing style and audience?  The spec might be more prose and expose reasoning and api could be terse details using exact language of arguments and returns.

core/coreapi/coreapi.go
```diff
@@ -0,0 +1,106 @@
...
+	importer "github.com/ipfs/go-ipfs/importer"
+	chunk "github.com/ipfs/go-ipfs/importer/chunk"
+	dag "github.com/ipfs/go-ipfs/merkledag"
+	path "github.com/ipfs/go-ipfs/path"
+	uio "github.com/ipfs/go-ipfs/unixfs/io"
+	mh "gx/ipfs/QmYf7ng2hG5XBtJA3tN34DQ2GUN5HNksEw1rLDkmr6vGku/go-multihash"
+	context "gx/ipfs/QmZy2y8t9zQH2a1b8q2ZSLKp17ATuJoCNxxyMFG5qFExpt/go-net/context"
+)
+
+type Path path.Path
```
whyrusleeping: why do we need to rebind the path type?

diasdavid:
@lgierth @Kubuxu have you seen https://github.com/ipfs/interface-ipfs-core ?

core/coreapi/coreapi.go
lgierth: yep should be an interface instead -- I made a WIP commit while switching computers

lgierth:
> have you seen https://github.com/ipfs/interface-ipfs-core ?

Yes -- what's your question? :) We can make that repo hold both js- and go- interfaces, and go-ipfs-api and go-ipfs-core-api hold the implementations.

diasdavid:
Just making sure it wasn't missed :) It think that if we can have a spec document that works for both, it would be great :)

lgierth:
Closing in favor of #3207

Issue #2870, opened 2016-06-18
Stdin handling with - inconsistent


Kubuxu:
That one is on me
```
 ~/tmp  echo "QmYwAPJzv5CZsnA625s3Xf2nemtYgPpHdWEz79ojWnPbdG" |ipfs ls -
Error: invalid ipfs ref path
 ~/tmp  echo "QmYwAPJzv5CZsnA625s3Xf2nemtYgPpHdWEz79ojWnPbdG" |ipfs ls                        ✘ 
QmZTR5bcpQD7cFgTorqxZDYaew1Wqgfbd2ud9QqGPAkK2V 1688 about
QmYCvbfNbCwFR45HiNP45rwJgvatpiW38D961L5qAhUM5Y 200  contact
QmY5heUM5qgRubMDD1og9fhCPA6QdkMp3QCwd4s7gJsyE7 322  help
QmdncfsVm2h5Kqq9hPmU7oAVX2zTSVP3L869tgTbPYnsha 1728 quick-start
QmPZ9gcCEpqKTo6aq61g2nXGUhM4iCL3ewB6LDXZCtioEB 1102 readme
QmTumTjvcYCAvRRwQ8sDRxh8ezmrcr88YFU7iYNroGGTBZ 1027 security-notes
```

but if we decide on https://github.com/ipfs/go-ipfs/issues/2868 it will resolve itself.

whyrusleeping:
@Kubuxu is this resolved?

Kubuxu:
Yes it is.

PR #2648, opened 2016-05-10
add test to enforce helptext on commands

whyrusleeping:
This is WIP. Not to be merged for a while.

@RichardLitt here you go :)

License: MIT
Signed-off-by: Jeromy why@ipfs.io

PR #2877, opened 2016-06-20
cli: Fix stdin parsing

thomas-gardner:
@Kubuxu requested to nuke `EnableStdin` on `StringArg`, but I reckon it's (slightly) better to just fix the parsing & avoid having to wonder, _is this a valid pipeline or do I need to use xargs?_

EDIT(Kubuxu, for tracking): Resolves https://github.com/ipfs/go-ipfs/pull/2877, Resolves https://github.com/ipfs/go-ipfs/issues/2870

### DIFF
--- a/commands/cli/parse_test.go
+++ b/commands/cli/parse_test.go
@@ -2,7 +2,6 @@ package cli
 
 import (
 	"io"
-	"io/ioutil"
 	"os"
 	"runtime"
 	"strings"
@@ -203,11 +202,7 @@ func TestArgumentParsing(t *testing.T) {
 	}
 
 	test := func(cmd words, f *os.File, res words) {
-		if f != nil {
-			if _, err := f.Seek(0, os.SEEK_SET); err != nil {
-				t.Fatal(err)
-			}
-		}
+		defer f.Close()
 		req, _, _, err := Parse(cmd, f, rootCmd)
 		if err != nil {
 			t.Errorf("Command '%v' should have passed parsing: %v", cmd, err)
@@ -255,15 +250,18 @@ func TestArgumentParsing(t *testing.T) {
 
 	// Use a temp file to simulate stdin
 	fileToSimulateStdin := func(t *testing.T, content string) *os.File {
-		fstdin, err := ioutil.TempFile("", "")
+		fstdin, w, err := os.Pipe()
 		if err != nil {
 			t.Fatal(err)
 		}
-		defer os.Remove(fstdin.Name())
 
-		if _, err := io.WriteString(fstdin, content); err != nil {
-			t.Fatal(err)
-		}
+		go func() {
+			defer w.Close()
+			if _, err := io.WriteString(w, content); err != nil {
+				t.Fatal(err)
+			}
+		}()
+
 		return fstdin
 	}
 
--- a/commands/cli/parse.go
+++ b/commands/cli/parse.go
@@ -289,15 +289,18 @@ func parseArgs(inputs []string, stdin *os.File, argDefs []cmds.Argument, recursi
 		var err error
 		if argDef.Type == cmds.ArgString {
 			if len(inputs) > 0 {
-				// If argument is "-" use stdin
+				// If argument is "-", replace with stdin
 				if inputs[0] == "-" && argDef.SupportsStdin {
 					stringArgs, stdin, err = appendStdinAsString(stringArgs, stdin)
 					if err != nil {
 						return nil, nil, err
 					}
+					inputs = inputs[1:]
 				}
 				// add string values
-				stringArgs, inputs = appendString(stringArgs, inputs)
+				if len(inputs) > 0 {
+					stringArgs, inputs = appendString(stringArgs, inputs)
+				}
 			} else if !argDef.SupportsStdin {
 				if len(inputs) == 0 {
 					// failure case, we have stdin, but our current
@@ -308,6 +311,12 @@ func parseArgs(inputs []string, stdin *os.File, argDefs []cmds.Argument, recursi
 				stringArgs, inputs = appendString(stringArgs, inputs)
 			} else {
 				if stdin != nil && argDef.Required && !fillingVariadic {
+					// ensure that stdin is a FIFO
+					stdInfo, _ := stdin.Stat()
+					if (stdInfo.Mode() & os.ModeNamedPipe) == 0 {
+						return nil, nil, err
+					}
+
 					// if we have a stdin, read it in and use the data as a string value
 					stringArgs, stdin, err = appendStdinAsString(stringArgs, stdin)
 					if err != nil {


commands/cli/parse_test.go
```diff
@@ -255,15 +250,18 @@ func TestArgumentParsing(t *testing.T) {
 
 	// Use a temp file to simulate stdin
```
chriscool: The comment above does not match anymore what is done below.

Kubuxu:
I don't like this solution much as we are again introducing trap of depending on type of input (named pipe in this case) to work. Are you sure that it will work the same with for example unnamed pipes or under docker?

About whether you should use `xargs` or just pipe in: after nuking StringArgs stdin handling it would be easy: If given command wants a file, you can pipe, if it works on argument strings, you want to `xargs` it.

Other thing is I don't know any command that work like `ipfs` now.

whyrusleeping:
@Kubuxu @thomas-gardner how about instead of having the arguments be strings here, we make them some kind of interface. that way, we can push through the parsing here without having to read from stdin, and then when we actually get to the command we can read from stdin as we need it.

Kubuxu:
As I talked with @whyrusleeping it would be very hard to parse parameters on demand as for the fact that parameters are transported over HTTP.

@lgierth could you weigh in?

whyrusleeping:
the biggest issue for me right now is that `ipfs block get --help` hangs.

whyrusleeping:
I don't think we can successfully get away with not checking what type of thing stdin is.

kevina:
> I don't like this solution much as we are again introducing trap of depending on type of input (named pipe in this case) to work. Are you sure that it will work the same with for example unnamed pipes or under docker?

I agree with @Kubuxu on this.  If we allow stdin input, it should have to be specified on the command line.  I do not like the idea at all of a command behaving differently if I pipe something into it, vs if I just run it from the terminal.  Sometimes I just run command that take input from stdin and just manually enter the data, useful in debugging and exploring how a command works.

Kubuxu:
My solution for that is to make `ipfs` behave as every other command and nuke stdin handling for string based arguments in favour of `xargs`. This is why `xargs` was created so every command doesn't have to do it on its own and so they all work in similar manner.

whyrusleeping:
@Kubuxu alright, lets go ahead with that. Explicit selecting stdin in all cases.

thomas-gardner:
I now agree that this is really aberrant behaviour for a UNIX command
(GNU `ls` doesn't accept pipes, for example.). A `git grep "StringA.*E"`
shows that, with two exceptions, all the StringArgs with EnableStdin are
expecting a hash of some sort. So it is actually very easy to know when
to use `xargs`: whenever you're piping a hash.

@Kubuxu , a couple questions: 

0. What should `ipfs` do if a StringArg is being piped in? Ignore it and
error?

1. Using `add` wlog, what should happen on all the following cases (my
expectations are in comments.)

	$ ipfs add
	# error

	$ echo "foo" |ipfs add
	# same as go-ipfs/master

	$ ipfs add -
	# same as go-ipfs/master

	$ echo "foo" |ipfs add -
	# echo "foo" |ipfs && ipfs add -

	$ ipfs add $filename
	# same as go-ipfs/master

	$ ipfs add $filename -
	# ipfs add $filename && ipfs add -

	$ echo "foo" |ipfs add $filename 
	# echo "foo" |ipfs add && ipfs add $filename

	$ echo "foo" |ipfs add $filename -
	# echo "foo" |ipfs add && ipfs add $filename && ipfs add -

and what should happen if a $filename doesn't exist?


===STALE===
Hmm, I should probably have clarified what this patch does.

0: FileArgs are unchanged, so slick lines like

	$ echo "Hello, Pluto?" |ipfs add 

will work as before.

1: Commands will not (should not?) run differently when strings are
piped in; it's just a way to construct the cmd line.

	$ hello="QmYwAPJzv5CZsnA625s3Xf2nemtYgPpHdWEz79ojWnPbdG"
	$ ./ipfs-31e65d7 ls "$hello"
	QmZTR5bcpQD7cFgTorqxZDYaew1Wqgfbd2ud9QqGPAkK2V 1688 about
	...
	$ echo "$hello" |./ipfs-31e65d7 ls >cool_hashes
	QmZTR5bcpQD7cFgTorqxZDYaew1Wqgfbd2ud9QqGPAkK2V 1688 about
	...

2: Manual input can be requested with a hyphen,

	$ ./ipfs-31e65d7 ls -
	# reads string arg until EOF
 
but a pipe will take precedence:

	$ cat cool_hashes |cut -f1 -d' '|./ipfs-31e65d7 cat  - |head -2

                   IPFS -- Inter-Planetary File system

===STALE===

Kubuxu:
Answers:
0. When data is being piped in into command that is not prepared for it, the command doesn't care

1. `ipfs add` command should work similarly to the `cat` command from Unix.
```
	$ ipfs add
	# error
	> read from stdin (display info that stdin is read over stderr)

	$ echo "foo" |ipfs add
	# same as go-ipfs/master
	> add text foo to ipfs

	$ ipfs add -
	# same as go-ipfs/master
	> read from stdin (maybe display info that stdin is read over stderr)

	$ echo "foo" |ipfs add -
	# echo "foo" |ipfs && ipfs add -
	> same as `echo "foo" | ipfs add`

	$ ipfs add $filename
	# same as go-ipfs/master
	> add $filename to ipfs

	$ ipfs add $filename -
	# ipfs add $filename && ipfs add -
	> add filename and read and add from stdin/console, can finish with Ctrl-D

	$ echo "foo" |ipfs add $filename 
	# echo "foo" |ipfs add && ipfs add $filename
        > ignore stdin, add just $filename

	$ echo "foo" |ipfs add $filename -
	# echo "foo" |ipfs add && ipfs add $filename && ipfs add -
        > add both $filename and string "foo"
```
Non existing filename should depend on the command or wherever it is handled now.

My proposition would change:

`cat $hashesfile | ipfs ls` into `cat $hashes | xargs ipfs ls` - same as you would do with normal `ls` command
`cat $hashesfile | ipfs cat` into `cat $hashes | xargs ipfs cat` - same as you would do with normal `cat` command (`xargs cat < $filestocatfile`)
`cat $nodesfile | ipfs bootstrap add` into `cat $nodesfile | xargs ipfs bootstrap add` - same way you would use the command from the command line (you place the input as arugments, no magic conversion in the `ipfs` command itself).

My idea is to model `ipfs` command after something familiar, and no command I know parses stdin and places it as its arguments.

thomas-gardner:
The close is unintentional.

@Kubuxu, how does this look?

Kubuxu:
@thomas-gardner as you removed the branch you will have to create new PR.

It looks generally good. I would also add go test to check that there are no StringArgs with EnableStdin called to root out the noise from code. The test can be written similarly to: https://github.com/ipfs/go-ipfs/pull/2648

We will also need sharness regression tests so this behaviour (hanging on `--help`) doesn't repeat.

Kubuxu:
Also as the message is only cosmetic we can make is so it is only displayed if stdin is character input device.

PR #2883, opened 2016-06-21
Standardized Readme

RichardLitt:
See https://github.com/ipfs/community/issues/124

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2014 Juan Batiz-Benet
+Copyright (c) 2016 Juan Batiz-Benet
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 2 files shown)


whyrusleeping:
LGTM

LICENSE
```diff
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2014 Juan Batiz-Benet
+Copyright (c) 2016 Juan Batiz-Benet
```
jbenet: No, you don't just "update" a copyright notice. the whole point is the earliest time. Also, in general don't remove things from copyright notices. should usually only add. 

So if you want to say 2016, then you'd have to say:

```
Copyright (c) 2014 Juan Batiz-Benet
Copyright (c) 2016 Juan Batiz-Benet
```
or
```
Copyright (c) 2014-2016 Juan Batiz-Benet
```
or 
```
Copyright (c) since 2014 Juan Batiz-Benet
```

RichardLitt: SGTM. Thanks.

PR #2884, opened 2016-06-21
raise file descriptor limit to 1024 by default

whyrusleeping:
This changes the behaviour of the ipfs daemon to raise the file descriptor ulimit to 1024 if it is below 1024 when the daemon starts.

On linux, this will have no effect.
On OSX, this will raise the file descriptor limit every time the daemon is started (as the default there is 256).

License: MIT

### DIFF
--- a/cmd/ipfs/ulimit_unix.go
+++ b/cmd/ipfs/ulimit_unix.go
@@ -9,7 +9,7 @@ import (
 	"syscall"
 )
 
-var ipfsFileDescNum = uint64(2048)
+var ipfsFileDescNum = uint64(1024)
 
 func init() {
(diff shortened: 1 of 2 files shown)


Kubuxu:
LGTM

cmd/ipfs/ulimit_unix.go
```diff
@@ -9,7 +9,7 @@ import (
 	"syscall"
 )
 
-var ipfsFileDescNum = uint64(2048)
+var ipfsFileDescNum = uint64(1024)
```
jbenet: is 1024 truly enough? why not 2048?

Issue #2850, opened 2016-06-13
increase blockstore 'writecache' functionality


whyrusleeping:
we currently have a cache in the blocks/blockstore/writecache.go that caches blocks we've written for later checks to avoid duplicating writes.

The most common by far datastore call is `Has`, it causes a lot of disk contention, and you can view the number of calls per method in the expvars that ipfs exports (`localhost:5001/debug/vars`).

On all `Get`, `Has` and `PutMany` calls we should be setting the cache respectively. We should also investigate both increasing this cache size and take a look at what using a different datastructure (such as a bloom filter) might look like.

kevina:
@whyrusleeping if you are going to use a bloom filter I would think you would want to implement that on the 'flatfs' datastore so that it can be complete and saved to disk.

If the flatfs datastore has a bloom filter than I image the write cache can probably be eliminated.

Kubuxu:
Bloom filter will only work well if most calls to for example `Has` will return false otherwise you have to check the real storage either way.

For `Get` and caching we would have to create weighed bloom filter, something I would be quite interested of doing.

(Note to self, read for a future: [Weighed bloom filter paper](http://faculty.cs.tamu.edu/ajiang/bloom.pdf))

kevina:
@whyrusleeping @Kubuxu do we have any stats on the number of Has() requests on the gateway nodes that return false?  With bitswap negative Has() requests would seam like a common thing.

What I am proposing that might help is to cache Has() itself at the datastore level with a bloom filter can be kept complete by maintaining a copy on disk.  I actually know very little about bloom filters so I have no idea if this is practical so consider this a question.  Is this something work considering?    I also have no idea if the cost of the bloom filter will outweigh the cost of negative stat system calls to check if a file exists in the flatfs datastore.  Thoughts?

Kubuxu:
Yes, we plan making a bloom filter with lru cache behind it. The bloom filter would have to be rebuilt on GC but it isn't a problem as implementation of the bloom filter I am working on now requires <100ns per insertion and <80ns per check.

The bloom filter itself should be quite small.

whyrusleeping:
we can check the number of total has requests, we don't know if they return true or false though.

PR #2885, opened 2016-06-21
Add ARC caching and bloom filter for blockstorage

Kubuxu:
Resolves https://github.com/ipfs/go-ipfs/issues/2850
License: MIT


whyrusleeping:
blocked on #2601

Kubuxu:
@kevina In other issues you said that AllKeysChan my be hard on performance, what is the alternative?

kevina:
@Kubuxu, unfortunately if you use the datastore there are not any.  I decided that for the filestore it would make more sense to just use the LevelDB directly.

If I where you I try hacking go-datastore and increase the channel buffer size here, https://github.com/ipfs/go-datastore/blob/master/query/query.go#L185 and see if that speeds things up (try a buffer size of say 16, than maybe something huge like 1024 or larger).

To see how much overhead is really involved you could hack go-datastore to be able to get the underlying LevelDB involved and query it directly.

Kubuxu:
Rebased on to master, tests are running.

Should I implement block cache in this PR or in a different one?

whyrusleeping:
@Kubuxu do it separately if its not too much trouble.

whyrusleeping:
rebase on master so we dont get the docker failures?

Kubuxu:
I should also squash everything as first commits are failing.

Kubuxu:
Squashed.

core/builder.go
```diff
@@ -131,7 +131,7 @@ func setupNode(ctx context.Context, n *IpfsNode, cfg *BuildCfg) error {
 
 	var err error
 	bs := bstore.NewBlockstore(n.Repo.Datastore())
-	n.Blockstore, err = bstore.WriteCached(bs, kSizeBlockstoreWriteCache)
+	n.Blockstore, err = bstore.BloomCached(bs, 256*1024, kSizeBlockstoreWriteCache)
```
whyrusleeping: lets make this a constant

package.json
```diff
@@ -177,6 +177,12 @@
       "hash": "Qmb1DA2A9LS2wR4FFweB4uEDomFsdmnw1VLawLE1yQzudj",
       "name": "base32",
       "version": "0.0.0"
+    },
+    {
+      "author": "kubuxu",
```
whyrusleeping: why couldnt we use the code in blocks/bloom ?

blocks/blockstore/bloom_cache.go
```diff
@@ -0,0 +1,175 @@
...
+		b.arc.Add(k, false)
+		return ErrNotFound
+	}
+	return err
+}
+
+// if ok == false has is inconclusive
+// if ok == true then has respons to question: is it contained
+func (b *bloomcache) hasCached(k key.Key) (has bool, ok bool) {
+	if k == "" {
```
whyrusleeping: what? we always have the empty key?

blocks/blockstore/bloom_cache.go
```diff
@@ -0,0 +1,175 @@
...
+		return true, true
+	}
+	if b.active {
```
whyrusleeping: this field needs to be locked around

blocks/blockstore/bloom_cache.go
```diff
@@ -0,0 +1,175 @@
...
+		blr := b.bloom.HasTS([]byte(k))
+		if blr == false { // not contained in bloom is only conclusive answer bloom gives
+			return blr, true
```
whyrusleeping: lets hardcode `return false, true`,  just for clarity

blocks/blockstore/bloom_cache.go
```diff
@@ -0,0 +1,175 @@
...
+}
+
+func (b *bloomcache) DeleteBlock(k key.Key) error {
+	if has, ok := b.hasCached(k); ok && !has {
+		return ErrNotFound
+	}
+
+	b.arc.Remove(k) // Invalidate cache before deleting.
+	err := b.blockstore.DeleteBlock(k)
+	if err == nil {
```
whyrusleeping:
```go
switch err {
case nil:
//
case ds.ErrNotFound, ErrNotFound:
//
default:
return err
}
```

package.json
Kubuxu: I spent few hours tweaking this one around.
It has performance order of magnitude higher than other I've found.

Also original author was very smart about how he hashes and uses higher hashbits to expand lower bits for the multiple hash functions bloom filter requires.

blocks/blockstore/bloom_cache.go
Kubuxu: It was there in the write_cache, I just mimicked it.

whyrusleeping: oh weird... i wonder what the reasoning there was

blocks/blockstore/bloom_cache.go
Kubuxu: Does it? I would prefer not to introduce locking here as it is quite slow when compared with speed of bloom.

It would only introduce race if bloom was being deactivated (it isn't in yet, but that doesn't matter) and the pointer to `bloom` was being changed. If we don't change the pointer and we acquire internal bloom's RWLock, we can then work on the internals of the bloom.

I will leave a comment about that if it seems good to you.

core/builder.go
Kubuxu: I am introducing `CacheOpts` opts in the ARC caching of whole blocks. Would it be ok if it went into next PR?

blocks/blockstore/bloom_cache.go
```diff
@@ -0,0 +1,175 @@
...
+	b.rebuildChan = make(chan struct{})
+	b.active = false
+}
+
+func (b *bloomcache) BloomActive() bool {
+	return b.active
+}
+
+func (b *bloomcache) Rebuild() {
+	ctx := context.TODO()
```
whyrusleeping: we should try and get a context passed in, maybe from the constructor?

core/builder.go
whyrusleeping: sure

blocks/blockstore/bloom_cache.go
Kubuxu: Seems like a good idea.

blocks/blockstore/bloom_cache.go
whyrusleeping: rebuild never gets called outside of the constructor currently (which i'm not super keen on), but even the one call in the constructor will cause a race. What if methods are being called right as its being set to true? setting a boolean isnt guaranteed to be atomic

Kubuxu: Yes, but when it is set from false to true then race is that `b.active` is still observed as `false`. This has no negative side-effects apart from call skipping the bloom filter cache check as bloom filter is just being enabled.

Kubuxu:
@whyrusleeping I will add the context passing in the next PR too. I will have to rework all places where ctor is used either way.

whyrusleeping:
@Kubuxu i'm still not sure this is ready. How does it fare on long running nodes? What happens when the filter gets full, or there are lots of deletes, etc

Kubuxu:
Main problem would be if there is a lot of deletes, then the false positive ratio would rise proportionally with ratio of blocks deleted, In worst case scenario (all block deleted) the performance would be where it was before the bloom filter (+1ms of latency).

Other case would be if the repo was much bigger in comparison with the bloom filter preselected size. Our current settings should provide optimal operation (<1% of false positives) up to 30k blocks in storage, at about 100k blocks it will rise to about 15% of false positives (but still it reduces number of Has requests by roughly 73%, 2% of real positives and 15% false positives).

I plan to add auto rebuild (and expand) but I would need concrete metrics for that of real operation. That is why I plan working this week on the metrics interface I mentioned on sprint.

whyrusleeping:
@Kubuxu alright, that sounds good to me. Lets get this in and get some mileage on it

jbenet:
it would be great to see benchmark graphs on a large typical workload with and without this PR. 

We need to get into the habit of benchmarking and graphing with large batteries of tests.

Kubuxu:
In this case the aim of introducing Bloom filtering and ARC caching on `Has` requests wasn't purely about performance but reducing load on the underlying file system, and thus hard drives.

We are not caching blocks (we are trying that in https://github.com/ipfs/go-ipfs/pull/2942, the performance benefits isn't clear), but only if block is available in the datastore or not.

Best metric for this change set will be reduction of disk hits when caching is available, on which I was supposed to work quite a time ago by improving metric collection.

PR #2886, opened 2016-06-21
return a better error if the ref is not an object

### LINKED ISSUE #2807: ipfs get $hashofblock gives confusing error message
When the hash referred is a block, one cannot use `ipfs get` (as it expects a file object) - but instead have to use `ipfs block get`. This is fine, but the error message when attempting is not very helpful in understanding this:
```
$ ipfs get QmSnWHYwDUE1gdcUb6ZYgU2khZ5FkxuWsTvXbw3zGsKc3x
Error: Failed to decode Protocol Buffers: incorrectly formatted merkledag node: Unmarshal failed. proto: illegal wireType 6

$ ipfs block get QmSnWHYwDUE1gdcUb6ZYgU2khZ5FkxuWsTvXbw3zGsKc3x
�ripld-image-versionhtilesize�axayetiles�axayflevel0��fformatcpngdsize�axayddata�
......
````
Expected would be something like "Requested hash $hash refers to a block, not an object. Use ipfs block get instead". Possibly with the detailed error message above following it.

Ping @Kubuxu

whyrusleeping:
closes #2807 

License: MIT

### DIFF
--- a/merkledag/merkledag.go
+++ b/merkledag/merkledag.go
@@ -3,6 +3,7 @@ package merkledag
 
 import (
 	"fmt"
+	"strings"
 	"sync"
 
 	blocks "github.com/ipfs/go-ipfs/blocks"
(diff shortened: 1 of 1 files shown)


Kubuxu:
LGTM

Kubuxu:
Looks even better to me.

RichardLitt:
What API messages should this change? Only `block`?

Kubuxu:
This changes error messages in many cases, but I don't think you have to include it in API changes.

PR #2891, opened 2016-06-22
add a test for auto-raising ulimit

whyrusleeping:
License: MIT

### DIFF
--- a/test/sharness/t0060-daemon.sh
+++ b/test/sharness/t0060-daemon.sh
@@ -121,4 +121,14 @@ test_expect_success "daemon with pipe eventually becomes live" '
   test_fsh cat stdin_daemon_out || test_fsh cat stdin_daemon_err || test_fsh cat stdin_poll_apiout || test_fsh cat stdin_poll_apierr
 '
 
+ulimit -n 512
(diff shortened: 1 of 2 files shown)


Kubuxu:
LGTM

whyrusleeping:
waiting for the travisCI OSX tests to go through... they are currently failing on master and its fixed in this PR

test/sharness/t0060-daemon.sh
```diff
@@ -121,4 +121,14 @@ test_expect_success "daemon with pipe eventually becomes live" '
   test_fsh cat stdin_daemon_out || test_fsh cat stdin_daemon_err || test_fsh cat stdin_poll_apiout || test_fsh cat stdin_poll_apierr
 '
 
+ulimit -n 512
+TEST_ULIMIT_PRESET=1
+test_launch_ipfs_daemon
+
+test_expect_success "daemon raised its fd limit" '
+	grep "ulimit" actual_daemon > /dev/null
```
jbenet: should have a way to check the number. this test does not _test_ the thing in question, it only checks that a word exists in a message printed. Check the number it was raised to. if you can, verify that it was indeed raised with os cmds or something.

PR #2896, opened 2016-06-23
Disable Travis Email notifications

Kubuxu:
There are so many of them that they are just annoyting
and for some reason there isn't feature of telling
Travis not to send notifications to me.

License: MIT

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
 # dist: trusty # KVM Setup
 
+notifications:
+  email: false
+
 os:
   - linux
   - osx


jbenet:
- @Kubuxu why can't you filter these in your inbox? Why do you have to change it for everyone? I was wondering why i stopped getting these.... They are actually quite useful when tests are usually green.
- See https://docs.travis-ci.com/user/notifications/ for more info on notifications in general. 
- We could send to a list instead of individuals, and those of us that want can subscribe to that list.

Kubuxu:
+1 for the list, I was filtering them from the inbox, and I think many people having write access to branches in this repo did the same.

They weren't really useful when you get 30+ of them a day, with mix of feature and master branches of everyone, where many of them were us testing things on CI (at a time we were doing a lot of CI runs to make the tests green).

Issue #2839, opened 2016-06-11
Adding files that are symlinks directly


Kubuxu:
Unix commands when you reference a symlink directly resolve it. IPFS does not.

Example:
```bash
mkdir /tmp/hello
echo "World" > /tmp/hello/world
ln -s /tmp/hello /tmp/sym
cd /tmp/sym
ipfs add -r .

cp -r . ../sym1
ls -l .. | grep sym
```

`cp` resolved the symlink and copied whole directory, IPFS added an meaningless symlink.

whyrusleeping:
@Kubuxu this is resolved now, right?

kevina:
I am reopening this issue because the fix causes problems for me in my filestore code.

`cp -r` (at least on unix) does _not_ resolve symbolic links unless the path end in a '/'.  For example
```
mkdir tmp
cd tmp
mkdir dira
touch dira/afile
ln -s dira dirb
cp -r dirb dirc
cp -r dirb/ dird
ls -l 
drwxr-xr-x 2 kevina kevina 4096 Aug 26 04:15 dira
lrwxrwxrwx 1 kevina kevina    4 Aug 26 04:15 dirb -> dira
lrwxrwxrwx 1 kevina kevina    4 Aug 26 04:15 dirc -> dira
drwxr-xr-x 2 kevina kevina 4096 Aug 26 04:15 dird
```

The original issue reported is due to a bug in how  the special case of `.` is handled and is easy to fix (and I will happy to do so).

Please also see my comments on the pull request #2897 as it is now seams impossible to add symbolic links to the `unixfs` protobuf.

The issue comes down to what do you want this to do:
```
echo "Hello World!" > touch afile
ln -s afile link
ipfs add link
```
Due you want this at add the symbolic link (the original behavior) or the file the link is pointed to (the new behavior)?

How I deal with change in the filestore code will depend on the answer to that question.

@jbenet, you might want to take a look at this.

Kubuxu:
You can add symbolic links inside directories.
We can revert it and think about it more.

Is there any usage pattern where user wants to add a symbolic link, instead of the file referenced by this link if he references the symbolic link directly?


Looks symlinks to files are always resolved.
```
 …/tmp/syms  echo Hello > file
 …/tmp/syms  ln -s file sym
 …/tmp/syms  ls
total 8.0K
-rw-r--r-- 1 kubuxu kubuxu 6 Aug 26 11:01 file
[...]

PR #2897, opened 2016-06-23
Resolve symlink if it is directly referenced in cli

Kubuxu:
Resolves https://github.com/ipfs/go-ipfs/issues/2839
License: MIT

### DIFF
--- a/commands/cli/parse.go
+++ b/commands/cli/parse.go
@@ -388,7 +388,10 @@ func appendFile(fpath string, argDef *cmds.Argument, recursive, hidden bool) (fi
 	}
 
 	fpath = filepath.ToSlash(filepath.Clean(fpath))
-
+	fpath, err := filepath.EvalSymlinks(fpath)
+	if err != nil {
+		return nil, err
+	}
 	stat, err := os.Lstat(fpath)
 	if err != nil {
 		return nil, err
--- a/test/sharness/t0044-add-symlink.sh
+++ b/test/sharness/t0044-add-symlink.sh
@@ -14,7 +14,10 @@ test_expect_success "creating files succeeds" '
 	mkdir -p files/badin
 	echo "some text" > files/foo/baz &&
 	ln -s ../foo/baz files/bar/baz &&
-	ln -s files/does/not/exist files/badin/bad
+	ln -s files/does/not/exist files/badin/bad &&
+	mkdir -p files2/a/b/c &&
+	echo "some other text" > files2/a/b/c/foo &&
+	ln -s b files2/a/d
 '
 
 test_add_symlinks() {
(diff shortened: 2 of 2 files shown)


commands/cli/parse.go
```diff
@@ -418,7 +418,10 @@ func appendFile(fpath string, argDef *cmds.Argument, recursive, hidden bool) (fi
 	}
 
 	fpath = filepath.ToSlash(filepath.Clean(fpath))
-
+	fpath, err := filepath.EvalSymlinks(fpath)
```
whyrusleeping: I think that this doesnt just resolve top level symlinks.

You could check if the final path element refers to a symlink and resolve it *only* in that case

Kubuxu: Top level as directly referenced by user, I should change name.

whyrusleeping:
I'd like to see a test for a path like: `a/b/SYMLINK/c` where 'c' is inside the folder that a symlink points to

test/sharness/t0044-add-symlink.sh
```diff
@@ -23,21 +24,21 @@ test_add_symlinks() {
...
 
 	test_expect_success "output looks good" '
-		echo "QmdocmZeF7qwPT9Z8SiVhMSyKA2KKoA2J7jToW6z6WBmxR" > goodlink_exp &&
+		echo QmcPNXE5zjkWkM24xQ7Bi3VAm8fRxiaNp88jFsij7kSQF1 > goodlink_exp &&
 		test_cmp goodlink_exp goodlink_out
 	'
 
 	test_expect_success "adding a broken symlink works" '
-		ipfs add -q files/bad > badlink_out
+		ipfs add -qr files/badin | head -1 > badlink_out
```
whyrusleeping: does this still work? even with the changes youre making?

Kubuxu: Because in this case we are adding symlink in recursive tree (not referenced directly by the command) this means that this symlink will not be resolved and added in the old way.

Kubuxu:
@whyrusleeping I don't understand.

whyrusleeping:
Something like this:
```bash
mkdir -p a/b/c
ln -s a/b a/d
ipfs add a/d/c
```

Just to put some coverage on resolving a symlink thats in the middle of a path.

Kubuxu:
@whyrusleeping added the test and squashed test fixing with changes in the code.

whyrusleeping:
Cool, LGTM

whyrusleeping:
(waiting on tests to complete)

test/sharness/t0044-add-symlink.sh
```diff
@@ -23,27 +27,34 @@ test_add_symlinks() {
...
 	test_expect_success "output looks good" '
-		echo QmWdiHKoeSW8G1u7ATCgpx4yMoUhYaJBQGkyPLkS9goYZ8 > filehash_exp &&
+		echo QmQRgZT6xVFKJLVVpJDu3WcPkw2iqQ1jqK1F9jmdeq9zAv > filehash_exp &&
 		test_cmp filehash_exp filehash_out
 	'
 
-	test_expect_success "adding a symlink adds the link itself" '
+	test_expect_success "adding a symlink adds the file itself" '
 		ipfs add -q files/bar/baz > goodlink_out
 	'
```
kevina: I believe the intent of this test was really to add a symbolic link as there is limited support for them in the unixfs format.

jbenet: i think so too. we do want to be able to store symlinks eventually

jbenet:
This needs to be reverted. See https://github.com/ipfs/go-ipfs/issues/2839#issuecomment-242910656

jbenet:
Use a flag instead to get the behavior of resolving the link first. 

When in doubt, do what git does.

Issue #2868, opened 2016-06-18
Required String arugments with stdin handling enabled block --help


Kubuxu:
Introduced in: https://github.com/ipfs/go-ipfs/pull/2822

As the string arguments from stdin are resolved before the command is run it halts execution of the command and waits for the input.

The simplest and the most sane solution would be to remove stdin from String arguments.
All Unix based system all include `xargs` command that does exactly that and current behaviour of `ipfs` command rises more confusion than it is useful.

My proposition would be to remove `EnableStdin` from non file type arguments and also log it/crash as an error in command creation routines.

whyrusleeping:
@Kubuxu which commands does this affect?

Kubuxu:
Many of them, there are about 30 StringArgs with Stdin handling enabled.

whyrusleeping:
hrm... I like being able to pipe arguments directly to the commands. Maybe theres a way to defer the processing of the arguments until later?

Kubuxu:
Not without introducing terminal vs pipe inconsistency again: https://github.com/ipfs/go-ipfs/pull/2877

I didn't know that you could directly pipe into commands for quite some time.
Using `xargs` is just natural.

whyrusleeping:
@Kubuxu is this resolved? If so, can you PR in a test that just runs `ipfs block get --help` and makes sure it returns?

Kubuxu:
https://github.com/ipfs/go-ipfs/issues/2913

PR #2902, opened 2016-06-24
commands: remove EnableStdin support for StringArg [v2]

thomas-gardner:
With verbose flag:
- remove EnableStdin() flags on all StringArg,

- remove all unneeded parsing code for StringArg, and print an
informative message if `ipfs` begins reading from a CharDevice,

- remove broken go tests for EnableStdin cli parsing, and add some
trivial test cases for reading FileArg from stdin,

- add a panic to prevent EnableStdin from being set on
StringArg in the future.

Resolves: #2877, #2870, #2868
License: MIT
Signed-off-by: Thomas Gardner <tmg@fastmail.com>

### DIFF
--- a/commands/cli/parse.go
+++ b/commands/cli/parse.go
@@ -1,7 +1,6 @@
 package cli
 
 import (
-	"bytes"
 	"fmt"
 	"os"
 	"path"
@@ -13,7 +12,10 @@ import (
 	cmds "github.com/ipfs/go-ipfs/commands"
 	files "github.com/ipfs/go-ipfs/commands/files"
 	u "gx/ipfs/QmZNVWh8LLjAavuQ2JXuFmuYH3C11xo988vSgp7UQrTRj1/go-ipfs-util"
+	logging "gx/ipfs/QmYtB7Qge8cJpXc4irsEp8zRqfnZMBeB7aTrMEkPk67DRv/go-log"
 )
+
+var log = logging.Logger("commands/cli")
 
 // Parse parses the input commandline string (cmd, flags, and args).
 // returns the corresponding command Request object.
@@ -238,6 +240,8 @@ func parseOpts(args []string, root *cmds.Command) (
 	return
 }
 
+const msgStdinInfo = "ipfs: Reading from %s; send Ctrl-d to stop.\n"
+
 func parseArgs(inputs []string, stdin *os.File, argDefs []cmds.Argument, recursive, hidden bool, root *cmds.Command) ([]string, []files.File, error) {
 	// ignore stdin on Windows
 	if runtime.GOOS == "windows" {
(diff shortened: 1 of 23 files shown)


commands/cli/parse.go
```diff
@@ -325,6 +301,10 @@ func parseArgs(inputs []string, stdin *os.File, argDefs []cmds.Argument, recursi
 				var file files.File
 				var err error
 				if fpath == "-" {
+					stdInfo, _ := stdin.Stat()
```
Kubuxu: I know that this error shouldn't happen in practice but we should check for it either way.
Log it under log.Error with some "This shouldn't happen" message.

commands/cli/parse.go
```diff
@@ -337,6 +317,10 @@ func parseArgs(inputs []string, stdin *os.File, argDefs []cmds.Argument, recursi
 			} else {
 				if stdin != nil && argDef.SupportsStdin &&
 					argDef.Required && !fillingVariadic {
+					stdInfo, _ := stdin.Stat()
```
Kubuxu: Same here, maybe extract it to function "isTerminal" or "printReadInfo" so it doesn't repeat.

Kubuxu:
@thomas-gardner great work on that, I left a comment and thing left to do is changing sharness tests to accommodate this change. Also I am writing down why we decided to make this change and what we are changing.

Kubuxu:
In this PR we are removing support of direct stdin handling for non-file type arguments in go-ipfs. There are two main reasons why we decided for it:
* making non-file type arguments being piped into `ipfs` process behave as expected is very difficult task. It is connected with fact that on CLI side there is very limited logic running and this logic can't decided if some argument will be used or not. This lead to introduction of regression that caused `--help` to hang (https://github.com/ipfs/go-ipfs/issues/2868). Some of you will say: "It was working before", my response is: "No it wasn't". We were cheating by treating console differently from other types of inputs which became a problem with docker and embedded nodes.
* behaviour like that is unprecedented before in any other command. It can be replicated with much more flexibility using the `xargs` command, is explicit and familiar to users and doesn't expose us to implementation edge cases.

Those are two main reasons for removal of direct stdin handling for non-file type arguments in `ipfs` command.

Kubuxu:
As I am really co-creator of this issue/PR @whyrusleeping you would have to review it.

commands/cli/parse_test.go
```diff
@@ -259,53 +247,17 @@ func TestArgumentParsing(t *testing.T) {
...
 		}
-		defer os.Remove(fstdin.Name())
 
 		if _, err := io.WriteString(fstdin, content); err != nil {
 			t.Fatal(err)
 		}
 		return fstdin
 	}
-
-	test([]string{"stdinenabled", "value1", "value2"}, nil, []string{"value1", "value2"})
```
whyrusleeping: Are you sure that all these tests need to be removed?

commands/cli/parse.go
```diff
@@ -12,9 +11,12 @@ import (
 
 	cmds "github.com/ipfs/go-ipfs/commands"
 	files "github.com/ipfs/go-ipfs/commands/files"
+	logging "gx/ipfs/QmYtB7Qge8cJpXc4irsEp8zRqfnZMBeB7aTrMEkPk67DRv/go-log"
```
Kubuxu: This is wrong go-log version. You either used old version or need to rebase on top of master.

commands/cli/parse_test.go
whyrusleeping: We definitely want to have solid coverage over this. I'm not in favor of mass removing tests like this

whyrusleeping: (i noticed you added a bunch in now)

Kubuxu:
LGTM

test/sharness/t0040-add-and-cat.sh
```diff
@@ -223,8 +223,8 @@ test_expect_success "ipfs cat output looks good" '
 	test_cmp expected actual
 '
 
-test_expect_success "ipfs cat accept hash from stdin" '
-	echo "$HASH" | ipfs cat >actual
+test_expect_success "ipfs cat accept hash from built input" '
+	echo "$HASH" | xargs ipfs cat >actual
```
jbenet: I thought this had been reverted, @whyrusleeping @Kubuxu -- the tests on master show it's still here.

jbenet: I want this test file have 0 (or minimal) changes in rc4.

PR #2903, opened 2016-06-24
encode keys to datastore with base32 standard encoding

whyrusleeping:
This addresses #2601 

Following this, we will need to:
- [x] integrate the datastore changes here: https://github.com/ipfs/go-datastore/pull/39
- [x] write a migration (https://github.com/ipfs/fs-repo-migrations/pull/39)
  - [x] tests
- [ ] make ipfs prompt the user to run the migration for them (https://github.com/ipfs/go-ipfs/issues/2907)

License: MIT


whyrusleeping:
I'm in a bit of a conundrum. To merge this, we need to have a migration ready. Before i ship the migration i want to have really solid tests for it. Before i can write tests for the migrations, i need an easy way to install the new version.

whyrusleeping:
@Kubuxu could you help me take a look at why these tests are failing? They pass locally without issue.

repo/fsrepo/defaultds.go
```diff
@@ -31,16 +31,8 @@ func openDefaultDatastore(r *FSRepo) (repo.Datastore, error) {
...
-	// leads to ~66k objects per dir, splitting 256*256-way leads to
-	// only 256.
-	//
-	// The keys seen by the block store have predictable prefixes,
-	// including "/" from datastore.Key and 2 bytes from multihash. To
-	// reach a uniform 256-way split, we need approximately 4 bytes of
-	// prefix.
 	syncfs := !r.config.Datastore.NoSync
-	blocksDS, err := flatfs.New(path.Join(r.path, flatfsDirectory), 4, syncfs)
+	blocksDS, err := flatfs.New(path.Join(r.path, flatfsDirectory), 7, syncfs)
```
kevina: @whyrusleeping I am not 100% sure, but I think this value should be 6 to preserve the original intent.  Here is why, the initial '/' should not be base32 encoded so the original 4 bytes will now be
```
1 + 3*8/5 = 5.8 = 6 bytes
```
instead of
```
4*8/5 = 6.4 = 7 bytes
```

whyrusleeping: @kevina I don't think we were counting the initial slash before. That was (and is) getting cut off before being passed into the flatfs by the mount datastore

whyrusleeping: The initial goal here was to split to 256 different directories, which is 16 bytes. With a prefix length of 7 here, we get 14 bits to shard by.

7 bytes of prefix gets us 6 characters (yay slash), at 5 bits per character that gives us 30 bits. The first 16 bits are taken by the `Qm` prefix, leaving us 14 bits or 16384 way sharding. Dropping down to 6 bytes will leave us with 25 total bits, and 9 bits after the prefix == a 512 way sharding.

Thanks for making me do the math, youre right. (and I like 512 way sharding better than 256 way sharding that we had before)

kevina: @whyrusleeping  it looks like you are correct in that the '/' is not included in the prefix count.  The comment is misleading.  That being said, I think 6 would be a better value in order to avoid too many directories for the average size repo.  (I.e, it would be better to round down than up).

kevina: @whyrusleeping.  Our messages crossed.  It looks like the prefix was _not_ included in the prefix length before ipfs/go-datastore#39 but now it is (see my comments on the pull request).  I will leave it up to you to determine the best value.  (To avoid too many directories I believe a smaller value would be better.)

Kubuxu:
It is just docker failure.

whyrusleeping:
@lgierth i thought we fixed the docker failure

whyrusleeping:
Mergin time.

Anyone who tries building from master after this merge will need to build the migrations and run them, at least until we put builds of the migrations up on dist.

repo/fsrepo/defaultds.go
```diff
@@ -31,16 +31,10 @@ func openDefaultDatastore(r *FSRepo) (repo.Datastore, error) {
...
-	// only 256.
-	//
-	// The keys seen by the block store have predictable prefixes,
-	// including "/" from datastore.Key and 2 bytes from multihash. To
-	// reach a uniform 256-way split, we need approximately 4 bytes of
-	// prefix.
 	syncfs := !r.config.Datastore.NoSync
-	blocksDS, err := flatfs.New(path.Join(r.path, flatfsDirectory), 4, syncfs)
+	// 5 bytes of prefix gives us 25 bits of freedom, 16 of which are taken by
+	// by the Qm prefix. Leaving us with 9 bits, or 512 way sharding
```
jbenet: we'll want to go deeper later on

jbenet:
- +1 for base32 instead of base64.
- curious why not just hex? (so much faster). too long for paths?
- curious why upper case? why not lowercase?

Kubuxu:
> curious why not just hex? (so much faster). too long for paths?

IIRC the main concern was in memory size of keys, but filesystem path size was also a factor.

jbenet:
Sgtm thanks

PR #2256, opened 2016-01-27
vendor libp2p with gx

whyrusleeping:
The biggest part of this PR will be making sure everything is still easy to dev with

Issue #2259, opened 2016-01-28
Config to force hash verification for local retrieval


sivachandran:
As per the IRC discussion https://botbot.me/freenode/ipfs/msg/58755967, content hashes are verified only when the content is retrieved from network. The hashes are not verified when retrieved from local storage. But verifying hashes for local retrieval helps us to detect disk corruption.  So please provide a config to enable hash verification even for local retrieval.

rht:
+1

rht:
e.g. `git-fsck`

whyrusleeping:
If the option proposed here is added, it should add a check here: https://github.com/ipfs/go-ipfs/blob/master/blocks/blockstore/blockstore.go#L93

to rehash the block (call `NewBlock(data)` instead of `NewBlockWithMultihash`) and validate the key matches

PR #2904, opened 2016-06-24
Add option for datastore read rehashing

Kubuxu:
Resolves #2256

Enabled using `Datastore.HashOnRead: true` in IPFS config.

### DIFF
--- a/blocks/blockstore/blockstore.go
+++ b/blocks/blockstore/blockstore.go
@@ -22,7 +22,8 @@ var log = logging.Logger("blockstore")
 // BlockPrefix namespaces blockstore datastores
 var BlockPrefix = ds.NewKey("blocks")
 
-var ValueTypeMismatch = errors.New("The retrieved value is not a Block")
+var ValueTypeMismatch = errors.New("the retrieved value is not a Block")
+var ErrHashMismatch = errors.New("block in storage has different hash than requested")
 
 var ErrNotFound = errors.New("blockstore: block not found")
 
@@ -71,6 +72,12 @@ type blockstore struct {
 	lk      sync.RWMutex
 	gcreq   int32
 	gcreqlk sync.Mutex
+
+	rehash bool
+}
+
+func (bs *blockstore) RuntimeHashing(enabled bool) {
+	bs.rehash = enabled
 }
 
 func (bs *blockstore) Get(k key.Key) (blocks.Block, error) {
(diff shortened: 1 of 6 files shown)


blocks/blockstore/blockstore.go
```diff
@@ -23,6 +23,7 @@ var log = logging.Logger("blockstore")
 var BlockPrefix = ds.NewKey("blocks")
 
 var ValueTypeMismatch = errors.New("The retrieved value is not a Block")
+var ErrHashMismatch = errors.New("Block on disk has different hash than expected.")
```
lgierth: - the block isn't neccessarily on disk, that's up to the datastore, isn't it? ;)
- IIRC no dot at the end. @RichardLitt dug up conventions for go error messages the other day

lgierth:
Would be good to have an explicit default value for this in repo/config/init.go

blocks/blockstore/blockstore.go
Kubuxu: yeah, right

core/builder.go
```diff
@@ -130,7 +130,16 @@ func setupNode(ctx context.Context, n *IpfsNode, cfg *BuildCfg) error {
 	}
 
 	var err error
-	n.Blockstore, err = bstore.WriteCached(bstore.NewBlockstore(n.Repo.Datastore()), kSizeBlockstoreWriteCache)
+	bs := bstore.NewBlockstore(n.Repo.Datastore())
+	n.Blockstore, err = bstore.WriteCached(bs, kSizeBlockstoreWriteCache)
+
+	rcfg, err := n.Repo.Config()
```
lgierth: I can't comment on non-changed lines, so here: the other `n.Repo.Config()` in this function can probably be removed.

The error handling is a bit off here too. You're throwing away the error from `bstore.WriteCached()`, and you can omit the `== nil` check if you just move that thing to *after* the usual `!= nil` check.

Kubuxu: I've screwed up here,  the `err != nil` check should be after 134.

Then latter I am checking for `err == nil` as we are requesting con-fig only in case of online more.

Kubuxu:
@lgierth thanks for looking at it, I made really stupid mistake there with error handling.

whyrusleeping:
lets add a test where we purposely mess with the values in the blocks directory on disk and try to read the data through ipfs.

test/sharness/t0084-repo-read-rehash.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/sh
+#
+# Copyright (c) 2016 Mike Pfister
```
RichardLitt: Who is Mike?

Kubuxu: Someone credited in `t0083`. :smile:

whyrusleeping:
circleCI seems to be taking a break, LGTM

whyrusleeping:
@Kubuxu what was the actual bug resolved here? i think 2256 is incorrect

Kubuxu:
It was #2259

blocks/blockstore/blockstore.go
```diff
@@ -71,6 +72,12 @@ type blockstore struct {
 	lk      sync.RWMutex
 	gcreq   int32
 	gcreqlk sync.Mutex
+
+	rehash bool
+}
+
+func (bs *blockstore) RuntimeHashing(enabled bool) {
```
jbenet: - this is not a clear name, because there _is_ runtime hashing regardless
- use: `HashOnRead`

PR #2908, opened 2016-06-25
test: Run t0080-repo test offline

### LINKED ISSUE #2906: t0080 has bad gc test
Type: test failure
Area: repo sharness tests
Priority: P1

Description:
in t0080-repo.sh, there is a test (currently test_expect_failure) that checks the size of a repo before and after a gc. This is occasionally faulty because the daemon is running during this and will sometimes write random bits of data to the datastore, thus changing the observed size. We should either run this test with the daemon offline, or find a better way to measure

Kubuxu:
Resolves #2906

License: MIT

### DIFF
--- a/test/sharness/t0080-repo.sh
+++ b/test/sharness/t0080-repo.sh
@@ -9,7 +9,7 @@ test_description="Test ipfs repo operations"
 . lib/test-lib.sh
 
 test_init_ipfs
-test_launch_ipfs_daemon
+test_launch_ipfs_daemon --offline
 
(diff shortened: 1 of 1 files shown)


whyrusleeping:
lets wait for #2903 to get in. It was the other half of this failure (prevented gc from collecting all the blocks it needed)

whyrusleeping:
Rebased, things are passing now. Gonna rerun a few times

whyrusleeping:
LGTM

Issue #2293, opened 2016-02-04
Error Messages: What does Code mean? 


RichardLitt:
When using the CLI API, I am frequently getting error messages.

Example:

```
{
  "Message": "This command can't be called directly. Try one of its subcommands.",
  "Code": 1
}
```

What does Code mean? 
Why are some messages only text (ex: `Argument 'root' is required`, while some messages formatted like the above?

whyrusleeping:
i'm not really sure how all this works to be honest, but i think that the code signifies that the exit status of the cli should be 1

RichardLitt:
Who would know?

RichardLitt:
@Kubuxu Do you know? 

I feel this might be a duplicate issue - I remember talking about it somewhere other than here.

whyrusleeping:
Code 1 signifies that the CLI should return an exit code of 1 (and similarly 0 for code 0)

RichardLitt:
I know what this means now.

PR #2915, opened 2016-06-28
Added note about error codes

RichardLitt:
How does this look? See #2293

License: MIT
Signed-off-by: Richard Littauer richard.littauer@gmail.com

### DIFF
--- a/core/commands/root.go
+++ b/core/commands/root.go
@@ -64,6 +64,13 @@ ipfs uses a repository in the local file system. By default, the repo is located
 at ~/.ipfs. To change the repo location, set the $IPFS_PATH environment variable:
 
   export IPFS_PATH=/path/to/ipfsrepo
+
+EXIT STATUS
+
+The CLI will exits with one of the following values:
+
+0     Successful execution.
+1     Failed executions.
 `,
 	},
 	Options: []cmds.Option{


RichardLitt:
@Kubuxu What do you think about this?

Kubuxu:
First thing it fails to build as you used backticks as quotes.
Other thing is that error codes from CLI are general knowledge to most users, it would make sense if our error codes had special meanings (type of crash, system that failed), but they don't. This would better fit http-api docs as the purpose of of the `Code` over there is unclear.

The `Code` in case of error on http corresponds to https://github.com/ipfs/go-ipfs/blob/master/commands/response.go#L17-L22 where first is 0 and last 2.

RichardLitt:
> Other thing is that error codes from CLI are general knowledge to most users...

I disagree with this - or, rather, I disagree with assuming that this is OK. Offloading knowledge building is important for some things, but I think having a place that says this may be useful to newcomers. I think the code in this PR is easy to put in, and helpful. 

I would like there to be more special meanings, too. I know where the codes come from - found that part of the code (thanks for linking it! Makes me feel pretty confident in my CR ability) - and there's a comment there saying this, too. Is there any plan to add more codes?

whyrusleeping:
I think that we should instead do something like manpages do for exit status. See `man grep` or `man man` and look for `EXIT STATUS`

RichardLitt:
How does that look?

whyrusleeping:
LGTM

Kubuxu:
LGTM

PR #2924, opened 2016-07-01
add repo verify command

whyrusleeping:
License: MIT

### DIFF
--- a/core/commands/repo.go
+++ b/core/commands/repo.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"strings"
 
 	bstore "github.com/ipfs/go-ipfs/blocks/blockstore"
 	cmds "github.com/ipfs/go-ipfs/commands"
@@ -263,6 +264,8 @@ var repoVerifyCmd = &cmds.Command{
 			}
 			if fails == 0 {
 				out <- &VerifyProgress{Message: "verify complete, all blocks validated."}
+			} else {
+				out <- &VerifyProgress{Message: "verify complete, some blocks were corrupt."}
 			}
 		}()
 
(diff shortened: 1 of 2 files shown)


kevina:
I would suggest that a non-zero exit code is returned from the command is some blocks failed.

Other than that it LGTM, but there are no test cases and I have not tested it manually.

core/commands/repo.go
```diff
@@ -207,16 +212,100 @@ daemons are running.
...
+		Tagline: "Verify all blocks in repo are not corrupted.",
+	},
+	Run: func(req cmds.Request, res cmds.Response) {
+		nd, err := req.InvocContext().GetNode()
+		if err != nil {
+			res.SetError(err, cmds.ErrNormal)
+			return
+		}
+
+		out := make(chan interface{})
```
jbenet: buffer of 1?

jbenet: i guess dont need it, but usually good to do it for cases like this

core/commands/repo.go
```diff
@@ -207,16 +212,100 @@ daemons are running.
...
+			bs.RuntimeHashing(true)
+
+			keys, err := bs.AllKeysChan(req.Context())
+			if err != nil {
+				log.Error(err)
+				return
+			}
+
+			var fails int
+			var i int
```
jbenet: `total` not `i`

test/sharness/t0084-repo-read-rehash.sh
```diff
@@ -34,4 +34,9 @@ test_expect_success 'getting modified block fails' '
 	grep "block in storage has different hash than requested" err_msg
 '
 
+test_expect_success "block shows up in repo verify" '
+	test_expect_code 1 ipfs repo verify > verify_out &&
+	grep "$H_BLOCK2" verify_out
+'
```
jbenet: this needs much more robust testing!!
- it needs a case with hundreds of objects that all check out fine
- manipulate one object at random, verify should now fail
- fix the object, verify should now pass again

PR #2926, opened 2016-07-01
Resolve symlinks of testfiles

### LINKED ISSUE #2925: commands/cli test failure
https://travis-ci.org/ipfs/go-ipfs/jobs/141500743

cc @thomas-gardner @Kubuxu

Kubuxu:
Resolves #2925

Docker still fails.

License: MIT

### DIFF
--- a/commands/cli/parse_test.go
+++ b/commands/cli/parse_test.go
@@ -4,6 +4,7 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
+	"path/filepath"
 	"runtime"
 	"strings"
 	"testing"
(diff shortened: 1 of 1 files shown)


thomas-gardner:
LGTM

I have a Go question though: why `os.Create` instead of `os.Open`?

Kubuxu:
Open is read-only. Create is read-write-create, there is also OpenFile allowing for precise control.

whyrusleeping:
@Kubuxu what was the problem here? could I get a tldr?

Kubuxu:
It was introduced by me in https://travis-ci.org/ipfs/go-ipfs/builds/140867106 as the docker tests were failing already I didn't notice that tests on Travis OS X are failing too.

On Travis OS X tmp dir is symlinked causing the test to fail as the path expected did not evaluate symlink.

whyrusleeping:
alright, LGTM

jbenet:
This needs to be reverted too. See https://github.com/ipfs/go-ipfs/pull/2897#issuecomment-242910691 and https://github.com/ipfs/go-ipfs/issues/2839#issuecomment-242910656

kevina:
@Kubuxu was this caused by #2897?  Should I revert this is the same pull request?

Kubuxu:
Yes, please.

kevina:
@Kubux @jbenet it looks like @whyrusleeping reverted this in 1aeda7eb6ed21aca3e290f9e9c92c7e5bd942c84 (part of #2952)

PR #2929, opened 2016-07-01
docker: remove go version constraint

lgierth:
There's only ever one version of a given package
per Alpine release, so GO_VERSION is superfluous
and only breaks the build if the go packages is updated.

License: MIT


Kubuxu:
SGTM

whyrusleeping:
SGTM

PR #2930, opened 2016-07-01
core/cmd: Remove file command

Kubuxu:
It is duplicate of files command group

License: MIT


whyrusleeping:
you have angered the tests

Kubuxu:
Closing for now, my plan is to merge it with `ls` command, make the JSON that is send via it more verbose and remove this command which functionally is almost the same just sends more data.

PR #2933, opened 2016-07-02
update go-datastore changes 0.1.2

whyrusleeping:
@kevina these are mostly your changes to the query code

License: MIT


Kubuxu:
LGTM

kevina:
LGTM.  

Note that there may still be room to improve "ipfs refs local" (for example find . -name '*.data' is still faster), but this simple change should not cause any problems.

whyrusleeping:
@kevina some of the cost of refs local is base58 encoding, that shit is expensive.

kevina:
@whyrusleeping Sorry `ipfs refs local` was not made to be taken literally, my benchmark code ipfs/go-datastore#40  does not attempt to decode the key using base58 and my very informal tests found `find . -name '*.data'` faster then than the direct method on the graph.  I did not report it because it was very informal test.

whyrusleeping:
@kevina ah! gotcha, thanks :)

PR #2937, opened 2016-07-02
docker: take -dev version suffix into account

lgierth:
The other commit makes sharness tests work with my symlinked-into-gopath setup.

### DIFF
--- a/test/Makefile
+++ b/test/Makefile
@@ -7,7 +7,7 @@ RANDOM_FILES_SRC = ../Godeps/_workspace/src/github.com/jbenet/go-random-files
 POLLENDPOINT_SRC= ../thirdparty/pollEndpoint
 GOSLEEP_SRC = ./dependencies/go-sleep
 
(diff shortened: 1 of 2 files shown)


lgierth:
This fix is pretty important because it's making master green again!!11

Kubuxu:
Could we just grep the version string with expected hash instead? This way if we choose other suffix (rc for example) it won't break again.

lgierth:
Something like `sed -e s/(-\w)?//`?

Kubuxu:
You could use `ipfs version --enc json` and parse it using sed:
`{"Version":"0.4.3-dev","Commit":"4fa85cc","Repo":"4","System":"amd64/linux","Golang":"go1.6.2"}`

lgierth:
Updated it to parse from `ipfs version --enc json`, the quotes got waay complicated so heredoc we go :)

whyrusleeping:
bashy bashy bashy bash. LGTM

Issue #2907, opened 2016-06-25
Automate downloading and running of migrations


whyrusleeping:
Currently, when ipfs needs to run a migration, it tells the user to download and run it. When we detect a migration is needed, we should ask the user if they want to automatically download and run it, then do so.

This is a feature I want merged in for the next update (since it will require a migration)

milahu:
this breaks on nixos linux

error `no such file or directory`
because [ld-linux.so is not found at /lib64/ld-linux-x86-64.so.2](https://duckduckgo.com/?q=nixos+binary+no+such+file+or+directory+fhs+lib+ld+linux+so&ia=web)


quickfix: use `fs-repo-migrations` from $PATH

<details>


```
ipfs daemon

Initializing daemon...
Kubo version: 0.16.0
Repo version: 12
System version: amd64/linux
Golang version: go1.19.3
Found outdated fs-repo, migrations need to be run.
Run migrations now? [y/N] y
Looking for suitable migration binaries.
Need 1 migrations, downloading.
Downloading migration: fs-repo-11-to-12...
Fetching with HTTP: "https://ipfs.io/ipfs/QmdaCHYBDHEhXCMoynH5UcohEay6m1XayZCcxWZzKAHNVN/fs-repo-11-to-12/versions"
Fetching with HTTP: "https://ipfs.io/ipfs/QmdaCHYBDHEhXCMoynH5UcohEay6m1XayZCcxWZzKAHNVN/fs-repo-11-to-12/v1.0.2/fs-repo-11-to-12_v1.0.2_linux-amd64.tar.gz"
Downloaded and unpacked migration: /run/user/1000/migrations1153952653/fs-repo-11-to-12 (v1.0.2)
Running migration fs-repo-11-to-12 ...
  => Running: /run/user/1000/migrations1153952653/fs-repo-11-to-12 -path=/home/user/.ipfs -verbose=true
The migrations of fs-repo failed:
  migration fs-repo-11-to-12 failed: fork/exec /run/user/1000/migrations1153952653/fs-repo-11-to-12: no such file or directory
If you think this is a bug, please file an issue and include this whole log output.
  https://github.com/ipfs/fs-repo-migrations

Error: migration fs-repo-11-to-12 failed: fork/exec /run/user/1000/migrations1153952653/fs-repo-11-to-12: no such file or directory
```

</details>

PR #2939, opened 2016-07-03
Automatically download and run migrations if needed

whyrusleeping:
This will prompt the user when a migration is needed to let ipfs try to automatically download and run the migrations.

If a version of the migrations tool is installed that is high enough, ipfs will use it instead of downloading its own.

Currently, the download its own function will fail as there is not a build of fs-repo-migrations with ipfs-3-to-4 up on the distributions page yet.

Resolves: https://github.com/ipfs/go-ipfs/issues/2907

License: MIT

### DIFF
--- a/repo/fsrepo/migrations/mfsr.go
+++ b/repo/fsrepo/migrations/mfsr.go
@@ -5,6 +5,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path"
+	"strconv"
 	"strings"
 )
 
@@ -16,27 +17,26 @@ func (rp RepoPath) VersionFile() string {
 	return path.Join(string(rp), VersionFile)
 }
 
-func (rp RepoPath) Version() (string, error) {
+func (rp RepoPath) Version() (int, error) {
 	if rp == "" {
-		return "", fmt.Errorf("invalid repo path \"%s\"", rp)
+		return 0, fmt.Errorf("invalid repo path \"%s\"", rp)
 	}
 
 	fn := rp.VersionFile()
 	if _, err := os.Stat(fn); os.IsNotExist(err) {
-		return "", VersionFileNotFound(rp)
+		return 0, VersionFileNotFound(rp)
 	}
 
 	c, err := ioutil.ReadFile(fn)
 	if err != nil {
-		return "", err
+		return 0, err
 	}
 
-	s := string(c)
-	s = strings.TrimSpace(s)
-	return s, nil
+	s := strings.TrimSpace(string(c))
+	return strconv.Atoi(s)
 }
 
-func (rp RepoPath) CheckVersion(version string) error {
+func (rp RepoPath) CheckVersion(version int) error {
 	v, err := rp.Version()
 	if err != nil {
 		return err
@@ -49,13 +49,36 @@ func (rp RepoPath) CheckVersion(version string) error {
 	return nil
 }
 
-func (rp RepoPath) WriteVersion(version string) error {
+func (rp RepoPath) WriteVersion(version int) error {
 	fn := rp.VersionFile()
-	return ioutil.WriteFile(fn, []byte(version+"\n"), 0644)
+	return ioutil.WriteFile(fn, []byte(fmt.Sprintf("%d\n", version)), 0644)
 }
 
 type VersionFileNotFound string
 
 func (v VersionFileNotFound) Error() string {
 	return "no version file in repo at " + string(v)
+}
+
+func TryMigrating(tovers int) error {
+	if !YesNoPrompt("run migrations automatically? [y/n]") {
+		return fmt.Errorf("please run the migrations manually")
+	}
+
+	return RunMigration(tovers)
+}
+
+func YesNoPrompt(prompt string) bool {
+	var s string
+	for {
+		fmt.Printf("%s ", prompt)
+		fmt.Scanf("%s", &s)
+		switch s {
+		case "y", "Y":
+			return true
+		case "n", "N":
+			return false
+		}
+		fmt.Println("Please press either 'y' or 'n'")
+	}
 }
--- a/repo/fsrepo/migrations/mfsr.go
+++ b/repo/fsrepo/migrations/mfsr.go
@@ -61,7 +61,7 @@ func (v VersionFileNotFound) Error() string {
 }
 
 func TryMigrating(tovers int) error {
-	if !YesNoPrompt("run migrations automatically? [y/n]") {
+	if !YesNoPrompt("Run migrations automatically? [y/N]") {
 		return fmt.Errorf("please run the migrations manually")
 	}
 
@@ -70,7 +70,7 @@ func TryMigrating(tovers int) error {
 
 func YesNoPrompt(prompt string) bool {
 	var s string
-	for {
+	for i := 0; i < 3; i++ {
 		fmt.Printf("%s ", prompt)
 		fmt.Scanf("%s", &s)
 		switch s {
@@ -78,7 +78,11 @@ func YesNoPrompt(prompt string) bool {
 			return true
 		case "n", "N":
 			return false
+		case "":
+			return false
 		}
 		fmt.Println("Please press either 'y' or 'n'")
 	}
+
+	return false
 }
--- a/repo/fsrepo/migrations/mfsr.go
+++ b/repo/fsrepo/migrations/mfsr.go
@@ -59,30 +59,3 @@ type VersionFileNotFound string
 func (v VersionFileNotFound) Error() string {
 	return "no version file in repo at " + string(v)
 }
-
-func TryMigrating(tovers int) error {
-	if !YesNoPrompt("Run migrations automatically? [y/N]") {
-		return fmt.Errorf("please run the migrations manually")
-	}
-
-	return RunMigration(tovers)
-}
-
-func YesNoPrompt(prompt string) bool {
-	var s string
-	for i := 0; i < 3; i++ {
-		fmt.Printf("%s ", prompt)
-		fmt.Scanf("%s", &s)
-		switch s {
-		case "y", "Y":
-			return true
-		case "n", "N":
-			return false
-		case "":
-			return false
-		}
-		fmt.Println("Please press either 'y' or 'n'")
-	}
-
-	return false
-}
--- a/repo/fsrepo/migrations/migrations.go
+++ b/repo/fsrepo/migrations/migrations.go
@@ -16,6 +16,12 @@ import (
 
 var DistPath = "https://ipfs.io/ipns/dist.ipfs.io"
 
+func init() {
+	if dist := os.Getenv("IPFS_DIST_PATH"); dist != "" {
+		DistPath = dist
+	}
+}
+
 const migrations = "fs-repo-migrations"
 
 func RunMigration(newv int) error {
--- a/repo/fsrepo/migrations/migrations.go
+++ b/repo/fsrepo/migrations/migrations.go
@@ -1,0 +1,220 @@
+package mfsr
+
+import (
+	"bufio"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"strconv"
+	"strings"
+)
+
+var DistPath = "https://ipfs.io/ipns/dist.ipfs.io"
+
+const migrations = "fs-repo-migrations"
+
+func RunMigration(newv int) error {
+	migrateBin := "fs-repo-migrations"
+	fmt.Println("  => checking for migrations binary...")
+	_, err := exec.LookPath(migrateBin)
+	if err == nil {
+		// check to make sure migrations binary supports our target version
+		err = verifyMigrationSupportsVersion(migrateBin, newv)
+	}
+
+	if err != nil {
+		fmt.Println("  => usable migrations not found on system, fetching...")
+		loc, err := GetMigrations()
+		if err != nil {
+			return err
+		}
+
+		err = verifyMigrationSupportsVersion(loc, newv)
+		if err != nil {
+			return fmt.Errorf("could not find migrations binary that supports version %d", newv)
+		}
+
+		migrateBin = loc
+	}
+
+	cmd := exec.Command(migrateBin, "-to", fmt.Sprint(newv), "-y")
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+
+	fmt.Printf("  => running migration: '%s -to %d -y'\n\n", migrateBin, newv)
+
+	err = cmd.Run()
+	if err != nil {
+		return fmt.Errorf("migration failed: %s", err)
+	}
+
+	fmt.Println("  => migrations binary completed successfully")
+
+	return nil
+}
+
+func GetMigrations() (string, error) {
+	latest, err := GetLatestVersion(DistPath, migrations)
+	if err != nil {
+		return "", fmt.Errorf("getting latest version of fs-repo-migrations: %s", err)
+	}
+
+	dir, err := ioutil.TempDir("", "go-ipfs-migrate")
+	if err != nil {
+		return "", fmt.Errorf("tempdir: %s", err)
+	}
+
+	out := filepath.Join(dir, migrations)
+
+	err = GetBinaryForVersion(migrations, migrations, DistPath, latest, out)
+	if err != nil {
+		fmt.Printf("  => error getting migrations binary: %s\n", err)
+		fmt.Println("  => could not find or install fs-repo-migrations, please manually install it")
+		return "", fmt.Errorf("failed to find migrations binary")
+	}
+
+	err = os.Chmod(out, 0755)
+	if err != nil {
+		return "", err
+	}
+
+	return out, nil
+}
+
+func verifyMigrationSupportsVersion(fsrbin string, vn int) error {
+	sn, err := migrationsVersion(fsrbin)
+	if err != nil {
+		return err
+	}
+
+	if sn >= vn {
+		return nil
+	}
+
+	return fmt.Errorf("migrations binary doesnt support version %d", vn)
+}
+
+func migrationsVersion(bin string) (int, error) {
+	out, err := exec.Command(bin, "-v").CombinedOutput()
+	if err != nil {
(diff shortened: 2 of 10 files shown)


repo/fsrepo/migrations/migrations.go
```diff
@@ -0,0 +1,220 @@
...
+	"net/http"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"strconv"
+	"strings"
+)
+
+var DistPath = "https://ipfs.io/ipns/dist.ipfs.io"
```
lgierth: let's make this overridable -- will also make it much much easier to test-drive this

Kubuxu: Yup, as we can then deploy to just ipfs.

Stebalien:
Why not just build the migration functionality into IPFS and run it on start if needed? Downloading and running arbitrary code like this makes auditing IPFS impossible and is generally considered a *big* no-no from a security perspective.

jbenet:
> Why not just build the migration functionality into IPFS and run it on start if needed? Downloading and running arbitrary code like this makes auditing IPFS impossible and is generally considered a big no-no from a security perspective.

Certifying the migration code the same way we certify go-ipfs code is enough. Meaning:
- we have the migrations accounted for and all versions hash addressed
- we should do code signing (and we will, havent had the bandwidth)

It's not "arbitrary code", it's "code from the developers", which if you don't trust, then you probably shouldn't be running go-ipfs anyway.

If we really wanted to, we could version lock (with a hash) directly from `go-ipfs`, but i don't think it's necessary. An IPNS pointer and code signing will be enough.

There are many reasons NOT to embed that code in go-ipfs:
- the migrations need to run the code from those old versions with exactly so they just end up vendoring the old code. putting that code in go-ipfs would blow up the binary size
- we can avoid having to lug around code for old migrations
- we can make sure to freeze those migrations and make sure they always run that way (instead of keeping them in a mutable, high velocity codebase)
- these migrations are on the ipfs `fs-repo`, which is go-ipfs independent (other impls will be able to read the same formats and have to evolve, too)

Stebalien:
> It's not "arbitrary code", it's "code from the developers", which if you don't trust, then you probably shouldn't be running go-ipfs anyway.

No, it's whatever happens to be at https://ipfs.io/ipns/dist.ipfs.io when the migrations run. I generally trust the IPFS developers to not be actively malicious but I don't expect them to be perfect and never lose control of a code signing key. Also, as implemented here, all an attacker needs a cert for ipfs.io and the ability to MITM the client (no developer involvement whatsoever).

> If we really wanted to, we could version lock (with a hash) directly from go-ipfs, but i don't think it's necessary.

This is absolutely necessary. While a unfortunate (it kills disconnected upgrades), it's still (mostly) fixes the security issue.

Would these migrations really be that big? Usually, they're pretty tiny (on the order of kilobytes).

> An IPNS pointer and code signing will be enough.

Until someone steels your keys.

---

Basically, this all comes down to (1) ensuring that everyone gets the same code and (2) minimizing the attack surface.

1. If everyone gets the same code, it's much harder to perform targeted attacks. Personally, I use the Tomoyo MAC system to jail most programs on my system so I generally know when a program does something fishy (usually, this turns out to be "just a bug" but it's still nice to know what's going on). If everyone is running the same code, then chances are I (or some other security researcher) will notice the backdoor and report it. However, if you rely on code signing, someone could steal your private keys, hack the distribution server, and distribute malicious binaries to targeted organizations only.

2. I already have to trust the Arch Linux developers to (1) not lose control of their keys and (2) verify updates to some extend before packaging. I don't expect them to examine the code but I do expect them to verify that there's nothing *too* fishy about the release. If you rely on code signing and distribute code out-of-band, I lose this extra verification step and now have to trust even more people not to make mistakes.

Kubuxu:
The `fs-repo-migrations`, as for today's build are 10MiB.

Stebalien:
A lot of that is the go runtime and shared dependencies. For example, if I remove the latest migration (3-4), it's 9.9MiB. However, some of the other migrations appear to be bigger (on the order of a megabyte) so I agree that building this into IPFS may not be the best way to do this (basically, IPFS would never be able to drop any dependencies needed to read old datastore formats).

repo/fsrepo/migrations/migrations.go
whyrusleeping: preference on an env var name?

whyrusleeping:
@Stebalien it prompts you before doing anything. If you want to download and verify everything yourself you still can. This just makes it easier on the 99% of users who are okay with auto updates

repo/fsrepo/migrations/mfsr.go
```diff
@@ -49,13 +49,36 @@ func (rp RepoPath) CheckVersion(version string) error {
...
 	return "no version file in repo at " + string(v)
 }
+
+func TryMigrating(tovers int) error {
+	if !YesNoPrompt("run migrations automatically? [y/n]") {
+		return fmt.Errorf("please run the migrations manually")
+	}
+
+	return RunMigration(tovers)
+}
```
lgierth: Sorry for creeping up on you with this feedback one day before the planned release :( :)

This should be part of the daemon command. With previous migrations, the daemon would abort with a non-zero exit code, now it'll just hang waiting for input. It'll be useful to have a `-migrate=true|false` option which assumes yes/no and omits the prompt. (Actually it's really important for server environments.)

lgierth:
> we could version lock (with a hash) directly from go-ipfs

I agree that version locking with a hash is good. Any hijacking of the ipfs.io CNAME DNS record will make the update vulnerable. This doesn't even require SSL MitM since we're not talking about A/AAAA records.

Stebalien:
> it prompts you before doing anything. If you want to download and verify everything yourself you still can. This just makes it easier on the 99% of users who are okay with auto updates

Good point. That with code signing is probably good enough for most cases. Unfortunately, it doesn't cover the unattended upgrade use case (e.g., running IPFS as a daemon on a NAS/server) but distros could probably include a subset of the migrations in their packages and run them in a post-upgrade script. As a matter of fact, they'll probably do this regardless.

However, I don't really see a reason to not just hash the migration code and embed the hash in IPFS -- checking a hash is *much* easier than code signing. Yes this means you can't fix bugs in the migration code after the fact without also releasing a new version of IPFS. However, this would only be an issue if you expected the migration code to be significantly buggier than IPFS itself.

whyrusleeping:
@Stebalien

> this means you can't fix bugs in the migration code after the fact without also releasing a new version of IPFS

Thats the reason i'm not as comfortable hard coding a hash. If that ends up happening (even though i'm fairly confident it wont) then the only reasonable response (from my end) is to re-release the same version with a new hard-coded migrations hash. Otherwise we end up with a 'bad' version that users would still be able to download and fail with in the future.

lgierth:
> I agree that version locking with a hash is good. Any hijacking of the ipfs.io CNAME DNS record will make the update vulnerable. This doesn't even require SSL MitM since we're not talking about A/AAAA records.

Nevermind, this is actually only an issue if you use a local IPFS daemon.

Stebalien:
> Thats the reason i'm not as comfortable hard coding a hash. If that ends up happening (even though i'm fairly confident it wont) then the only reasonable response (from my end) is to re-release the same version with a new hard-coded migrations hash. Otherwise we end up with a 'bad' version that users would still be able to download and fail with in the future.

At the end of the day, this is solving a different concern. Think of it this way:

By default, X would be included in Y itself as it is logically part of Y. However, X is being extracted into a separate binary to be downloaded on-demand because...

Case 1: ...it would make Y significantly larger and isn't needed for normal operation.
Case 2: ...it can't be distributed with Y (e.g. licensing issues).
Case 3: ...it needs to be updated more frequently than Y.

Only case 3 motivates using code signing over including a hash but *only* case 1 is applicable here. Now, you could argue that, because we're doing this anyways, why not tackle two birds with one stone. However, I feel that a bug in the migration code is actually significantly *less* likely than a bug in the rest of IPFS so the chances of there being a reason to release a new migration tool and there existing *no* reason to release a new version of IPFS is vanishingly small. Basically, doing this would be sacrificing some security and simplicity (again, code signing is non-trivial) for negligible gain.

whyrusleeping:
@Stebalien want to try running the migrations to help test it for me? If i have more people tell me it works well then i'll be more inclined to hard code a hash in

Stebalien:
I have. I'm running ad5730d.

whyrusleeping:
okay cool, everything went smoothly i take it?

Stebalien:
I haven't noticed any problems so far. I just tried extracting an Arch Linux ISO (large file) I put into IPFS before the migration and got the correct result (md5 was correct) so it appears to be working.

repo/fsrepo/migrations/mfsr.go
jbenet: Agreed! 👍

repo/fsrepo/migrations/migrations.go
```diff
@@ -16,6 +16,12 @@ import (
 
 var DistPath = "https://ipfs.io/ipns/dist.ipfs.io"
 
+func init() {
+	if dist := os.Getenv("IPFS_DIST_PATH"); dist != "" {
```
jbenet: i think this should be a variable, not an env var. env vars are really pernicious as is. but added here that this control the distribution updates, rough!

the issue with env vars and _why_ they're really bad for security is that it can turn secure scripts and commands into something unexpected by running before. meaning, in something like this:

```
> ./something
...
> /secure/bin/ipfs daemon --yes-run-migrations
```
`something` may not have been secured to not download anything, doesnt have permissions to change binaries or anything, but it can change an env var. it adds attach surface area.

jbenet: though i think this is irrelevant if the hashes are hard coded.

repo/fsrepo/migrations/mfsr.go
```diff
@@ -70,15 +70,19 @@ func TryMigrating(tovers int) error {
 
 func YesNoPrompt(prompt string) bool {
 	var s string
-	for {
+	for i := 0; i < 3; i++ {
```
jbenet: haha thank you for this.

jbenet:
Thanks everyone for this awesome discussion. Very happy with the careful thought expressed here!

Couple of points, some already mentioned above:
- Relying on `dnslink` is _not_ great in general. [DNS lumps in a ton of authorities and can get co-opted relatively easily](https://github.com/bitcoin-dot-org/bitcoin.org/issues/1325#issuecomment-230176702), and that sucks. Though yes, we already rely on it, which sucks.
- while we dont have code signing yet, and dont use IPNS key links, we should hardcode the hash for safety. i think it's fine to have to release a new version of the code to get the new hash.
- this would also make it possible to use IPFS itself to download the migration! :) :heart: instead of using HTTP and relying on the gateway. (would have to bring up an ephemeral daemon with a temporary repo (because the binary expects a new version of the repo and cant use the current one), or could even use `ipget`)

jbenet:
@whyrusleeping give me a set of commands to try

whyrusleeping:
@jbenet just install this version and make sure your repo is an older v3 version. Then run `ipfs daemon`

lgierth:
I have a commit to add to this: bc6d71f4cdea90f61b0c9f2d1a68c6474e4c65b9

It detects whether the running go-ipfs binary is linked against musl libc, and sets the OS to `linux-musl` in case it does, so that we end up downloading `fs-repo-migrations_v1.1.0_linux-musl-amd64.tar.gz`. This allows our docker image to auto-migrate :) `docker run -it -v /path/to/repo:/data/ipfs ipfs/go-ipfs --migrate=true`

lgierth:
Okay, rebased this branch on top of a recent-ish master, added my OS detection commit, amended to not spawn a shell.

lgierth:
BTW, here's how I've been testing this, with `go-ipfs/ipfs` being from the v0.4.2 tarball: `image=$(docker build -f Dockerfile.fast .) && IPFS_PATH=ipfspath go-ipfs/ipfs init && docker run -it -v `pwd`/ipfspath:/data/ipfs d2dea496599f --migrate=true`

repo/fsrepo/migrations/migrations.go
```diff
@@ -15,7 +15,7 @@ import (
 	"strings"
 )
 
-var DistPath = "https://ipfs.io/ipns/dist.ipfs.io"
+var DistPath = "https://ipfs.io/ipfs/QmUnvqDuRyfe7HJuiMMHv77AMUFnjGyAU28LFPeTYwGmFF"
```
lgierth: :+1:

lgierth:
Feedback from @jbenet about the prompt and userspace: would be good to prompt only when there's a TTY, and otherwise assume --migrate=false. Otherwise SGTHim

Stebalien:
Where does it verify the hash or is that not happening (yet)?

lgierth:
Gonna test this with ipfs-ctl too tomorrow

lgierth:
> Where does it verify the hash or is that not happening (yet)?

- [ ] hardcoded hash -- this will require an fs-repo-migrations release before bumping the repo version on master the next time. I think that's good as it makes it easier for others to test migrations .

> Feedback from @jbenet about the prompt and userspace: would be good to prompt only when there's a TTY, and otherwise assume --migrate=false.

- [x] It looks like `fmt.Scanf()` already does the right thing for us.

> Gonna test this with ipfs-ctl too tomorrow

- [x] I tested this and I think it's fine -- js-ipfsd-ctl seems to always create a fresh repo.

whyrusleeping:
@lgierth did @jbenet get a chance to look at this then?

whyrusleeping:
Chooo Chooo!

repo/fsrepo/migrations/migrations.go
```diff
@@ -0,0 +1,261 @@
...
+	"net/http"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"strconv"
+	"strings"
+)
+
+var DistPath = "https://ipfs.io/ipfs/QmUnvqDuRyfe7HJuiMMHv77AMUFnjGyAU28LFPeTYwGmFF"
```
jbenet: doesn't this need a migration usable by `0.4.3(-*)?` ? i dont think this has the last migration needed

repo/fsrepo/migrations/migrations.go
```diff
@@ -0,0 +1,261 @@
...
+	if err != nil {
+		return "", fmt.Errorf("tempdir: %s", err)
+	}
+
+	out := filepath.Join(dir, migrations)
+
+	err = GetBinaryForVersion(migrations, migrations, DistPath, latest, out)
+	if err != nil {
+		fmt.Printf("  => error getting migrations binary: %s\n", err)
+		fmt.Println("  => could not find or install fs-repo-migrations, please manually install it")
```
jbenet: - give the user more instructions than that. ("manually install it"? from where? wat?)
- direct them to `https://dist.ipfs.io` and maybe `https://github.com/ipfs/fs-repo-migrations`

repo/fsrepo/migrations/migrations.go
lgierth: it has fs-repo-migrations-1.1.0 which includes the 3-to-4 migration: https://github.com/ipfs/fs-repo-migrations/tree/v1.1.0

repo/fsrepo/migrations/migrations.go
```diff
@@ -0,0 +1,261 @@
...
+	sn, err := migrationsVersion(fsrbin)
+	if err != nil {
+		return err
+	}
+
+	if sn >= vn {
+		return nil
+	}
+
+	return fmt.Errorf("migrations binary doesnt support version %d: %s", vn, fsrbin)
```
jbenet: - and then what? what should the user do? this leaves them hanging.
- direct them to https://dist.ipfs.io to get a newer fs-repo-migrations binary, and to file a bug in go-ipfs because no go-ipfs bin should ship with a fs-repo-migrations lower than what it needs.

repo/fsrepo/migrations/migrations.go
```diff
@@ -0,0 +1,261 @@
...
+func migrationsVersion(bin string) (int, error) {
+	out, err := exec.Command(bin, "-v").CombinedOutput()
+	if err != nil {
+		return 0, fmt.Errorf("failed to check migrations version: %s", err)
+	}
+
+	vs := strings.Trim(string(out), " \n\t")
+	vn, err := strconv.Atoi(vs)
+	if err != nil {
+		return 0, fmt.Errorf("migrations binary version check did not return a number")
```
jbenet: what should the user do in these cases?

repo/fsrepo/migrations/migrations.go
```diff
@@ -0,0 +1,261 @@
...
+		archive = "zip"
+	default:
+		archive = "tar.gz"
+	}
+	osv, err := osWithVariant()
+	if err != nil {
+		return err
+	}
+	finame := fmt.Sprintf("%s_%s_%s-%s.%s", distname, vers, osv, runtime.GOARCH, archive)
+	distpath := fmt.Sprintf("%s/%s/%s/%s", root, distname, vers, finame)
```
jbenet: it would be great if all this came from a library somewhere as we'll likely use this to download other things in in https://dist.ipfs.io -- maybe could be the beginning of gx-bin or could be just a small `go-ipfs-dist` bindings lib.

whyrusleeping: Yeah, I wanted to extract it into a separate lib from `ipfs-update` (where this code came from) but ipfs-update still depends on go-ipfs as a dependency. Once we've gotten a little farther down the road of extraction we can definitely make this its own thing (`gx-bin` is a great place for this to live i think)

repo/fsrepo/migrations/migrations.go
```diff
@@ -0,0 +1,261 @@
...
+	}
+	var latest string
+	for i := len(vs) - 1; i >= 0; i-- {
+		if !strings.Contains(vs[i], "-dev") {
+			latest = vs[i]
+			break
+		}
+	}
+	if latest == "" {
+		return "", fmt.Errorf("couldnt find a non dev version in the list")
```
whyrusleeping: This code is only here as a holdover from ipfs-update to filter out the 'ipfs0.4.0-dev' type versions. We shouldnt encounter this during migrations code. In any case, i'll add a better error string here

PR #2941, opened 2016-07-04
Feat/backoff clear connect

### LINKED ISSUE #2456: command to clear swarm dial backoff
while diagnosing some network issues, i often hit the 'dial backoff', where our node decides we've tried to dial that other node enough times, its probably dead, and we wait a little while to try again.

This sucks when i'm manually trying to dial things. We should have a command to clear this, or to ignore the backoff somehow if we are manually running swarm connect commands.

This will likely involve work in both libp2p and go-ipfs

whyrusleeping:
Resolves #2456


Kubuxu:
LGTM

I don't know but last time I looked at it it was hidden much deeper (even including the go-libp2p patch) but it might have been me not knowing libp2p back then at all.

Kubuxu:
Other thing: should we have tests for it?

whyrusleeping:
@Kubuxu Yeah, we should probably have tests for this. But i'm not sure how to reliably do them.
I'll take a hack at it before merging this.

Kubuxu:
LGTM

core/commands/swarm.go
```diff
@@ -241,6 +249,8 @@ ipfs swarm connect /ip4/104.131.131.82/tcp/4001/ipfs/QmaCpDMGvV2BGHeYERUEnRQAwe3
 
 		output := make([]string, len(pis))
 		for i, pi := range pis {
+			swrm.Backoff().Clear(pi.ID)
```
jbenet: this is leaking the underlying abstraction. can this be turned into something more "user/client" oriented? Not sure what name would fit

PR #2942, opened 2016-07-05
blocks/blockstore: Get ARC cache and context passing

Kubuxu:


whyrusleeping:
@Kubuxu this PR results in a small decrease in performance on the 'DirAddOpsPerSec' benchmark in ipfs-whatever:

```
Results  Before  After  % Change
PatchOpsPerSec   71.43   85.76   20.06%
DirAddOpsPerSec  528.10  491.64  -6.90%
Add10MBTime      443.56  439.88  -0.83%
Add10MBStdev     20.63   17.23   -16.48%
```

Kubuxu:
Are you sure that it isn't just measurement error?
On my loaded server I've got: 252/s to 395/s

We should display stdev for all of them.

Kubuxu:
Here are results from my server:

Results | Master | After | % Change
-------|------|--------|---------
PatchOpsPerSec | 14.08  | 13.93 |  -1.09%
DirAddOpsPerSec | 414.69 | 437.25 | 5.44%
DirAddOpsStdev | 55.80 | 27.55 | -50.63%
Add10MBTime | 3.27 | 2.80 | -14.21%
Add10MBStdev | 1.00 | 0.72 | -27.93%

but numbers are all over the place on that server.

Kubuxu:
This is also what I am getting on my notebook:
```
 Results          Before  After   % Change
PatchOpsPerSec   56.28   73.34   30.31%
DirAddOpsPerSec  679.71  643.48  -5.33%
DirAddOpsStdev   23.17   61.40   165.04%
Add10MBTime      0.49    0.41    -17.60%
Add10MBStdev     0.49    0.36    -25.44%
Cat1MBTime       10.38   8.38    -19.33%
Cat1MBStdev      2.13    0.94    -55.90%
```

I can't explain the DirAddOps.

Kubuxu:
Results from C1 Scaleway, no variable load:
```
Results          Before  After   % Change
PatchOpsPerSec   66.91   69.92   4.49%
DirAddOpsPerSec  52.65   52.21   -0.84%
DirAddOpsStdev   1.53    1.71    11.81%
Add10MBTime      2.75    2.84    3.07%
Add10MBStdev     0.05    0.04    -11.82%
Cat1MBTime       140.50  131.18  -6.63%
Cat1MBStdev      19.61   13.90   -29.10%
```

whyrusleeping:
Hrm... that doesnt appear to provide significant improvement.

Kubuxu:
Yeah, I think that in many cases there will be RAM/Disk kernel level cache hit.
I propose we bump it from 0.4.3 milestone, and rethink. Maybe in this case simpler (and with lower flat cost) LRU cache will work better.

whyrusleeping:
@Kubuxu sounds good to me

whyrusleeping:
marking as blocked on better metrics and benchmarking

whyrusleeping:
We should add more tests to ipfs-whatever, and then make it easy to run those benchmarks on various backing datastores.

whyrusleeping:
@Kubuxu I'm gonna close this, if we decide later that adding the arc cache here will improve perf significantly then we can revive the branch from the archive

Kubuxu:
As you removed this branch it won't be in archive. I will repush.

whyrusleeping:
Sorry, I really like the delete branch button. Its fantastic. So much fun to click

PR #2943, opened 2016-07-05
fix handling of dht records and local fixups

whyrusleeping:
This fixes a couple bugs that made ipns not work well in cases where the peer ID had a '/' in it. 

It also fixes a bug where we would never succeed in updating the locally stored entry

License: MIT

### DIFF
--- a/routing/dht/routing.go
+++ b/routing/dht/routing.go
@@ -122,6 +122,13 @@ func (dht *IpfsDHT) GetValue(ctx context.Context, key key.Key) ([]byte, error) {
 		// if someone sent us a different 'less-valid' record, lets correct them
 		if !bytes.Equal(v.Val, best) {
 			go func(v routing.RecvdVal) {
+				if v.From == dht.self {
+					err := dht.putLocal(key, fixupRec)
+					if err != nil {
+						log.Error("Error correcting local dht entry:", err)
+					}
+					return
+				}
 				ctx, cancel := context.WithTimeout(dht.Context(), time.Second*30)
 				defer cancel()
 				err := dht.putValueToPeer(ctx, v.From, key, fixupRec)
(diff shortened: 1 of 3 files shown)


Kubuxu:
LGTM

routing/dht/routing.go
```diff
@@ -122,6 +122,13 @@ func (dht *IpfsDHT) GetValue(ctx context.Context, key key.Key) ([]byte, error) {
 		// if someone sent us a different 'less-valid' record, lets correct them
 		if !bytes.Equal(v.Val, best) {
 			go func(v routing.RecvdVal) {
+				if v.From == dht.self {
+					err := dht.putLocal(key, fixupRec)
+					if err != nil {
+						log.Error("Error correcting local dht entry:", err)
+					}
+					return
+				}
```
lgierth: I guess this is what the self-dials where from?

whyrusleeping: Yeap, that was indeed the source

routing/record/validation.go
```diff
@@ -73,13 +73,19 @@ func (v Validator) IsSigned(k key.Key) (bool, error) {
 // verifies that the passed in record value is the PublicKey
 // that matches the passed in key.
 func ValidatePublicKeyRecord(k key.Key, val []byte) error {
-	keyparts := bytes.Split([]byte(k), []byte("/"))
-	if len(keyparts) < 3 {
-		return errors.New("invalid key")
+	if len(k) != 38 {
+		return errors.New("invalid public key record key")
```
lgierth: I feel like this length check will cause trouble when changing multihashes.

Kubuxu: If we don't want to do that we can do:
```go
bytes.SplitN([]byte(k). []byte("/"), 2)
```
then the second part should contain the rest of the key (even if it contains more slashes).

whyrusleeping: true... I guess i could switch to chopping off the prefix and checking that its a valid multihash

lgierth:
LGTM!

whyrusleeping:
circleCI failure appears to be a caching issue

routing/dht/routing.go
RichardLitt: Should have been lowercase.

PR #2952, opened 2016-07-07
rework stdin handling

whyrusleeping:
no comment

License: MIT

### DIFF
--- a/commands/cli/parse.go
+++ b/commands/cli/parse.go
@@ -304,12 +304,10 @@ func parseArgs(inputs []string, stdin *os.File, argDefs []cmds.Argument, recursi
 		case cmds.ArgString:
 			if len(inputs) > 0 {
 				stringArgs, inputs = append(stringArgs, inputs[0]), inputs[1:]
-			} else {
-				if stdin != nil && argDef.SupportsStdin && !fillingVariadic {
-					if err := printReadInfo(stdin, msgStdinInfo); err == nil {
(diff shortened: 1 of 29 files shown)


whyrusleeping:
cc @jbenet @Kubuxu @lgierth

whyrusleeping:
Alright, this should be ready for review. cc @jbenet @Kubuxu @thomas-gardner

commands/cli/parse.go
```diff
@@ -17,6 +17,11 @@ import (
 
 var log = logging.Logger("commands/cli")
 
+// stdinSpecialName is a name applied to the 'stdin' file so we can differentiate
+// it from potential 'real' files being passed in. The '*' character is invalid in
+// path names and won't appear otherwise.
+const stdinSpecialName = "*stdin*"
```
Kubuxu: `*` is valid in path names.

Kubuxu:
```
 ../test/sharness   patch-1  echo Yes it works > '*stdin*'
 ../test/sharness   patch-1 ?  cat '*stdin*'
Yes it works
```

whyrusleeping: youre right, and this is no longer needed.

RichardLitt:
Does this affect the HTTP API at all?

jbenet:
This PR:
- merged commits that do not pass tests (squash them)
- merged commits with useless commit messages... (squash them, or write better commit messages)

PR #2953, opened 2016-07-08
core: do not run bloom in case of ephemeral node

Kubuxu:
I will do nice and full refactor of the Online vs Ephemeral modes.

This PR includes parts of https://github.com/ipfs/go-ipfs/pull/2942 which cleanup passing of the context and options to the caches.

The important change is https://github.com/ipfs/go-ipfs/commit/697a3457041c42aba1ab61b5fd404320ea9585eb

### DIFF
--- a/blocks/blockstore/caching.go
+++ b/blocks/blockstore/caching.go
@@ -23,9 +23,6 @@ func DefaultCacheOpts() CacheOpts {
 
 func CachedBlockstore(bs GCBlockstore,
 	ctx context.Context, opts CacheOpts) (cbs GCBlockstore, err error) {
-	if ctx == nil {
-		ctx = context.TODO() // For tests
-	}
 	cbs = bs
 
 	if opts.HasBloomFilterSize < 0 || opts.HasBloomFilterHashes < 0 ||
--- a/blocks/blockstore/caching.go
+++ b/blocks/blockstore/caching.go
@@ -1,0 +1,16 @@
+package blockstore
+
+// Next to each option is it aproximate memory usage per unit
+type CacheOpts struct {
+	HasBloomFilterSize   int // 1 bit
+	HasBloomFilterHashes int // No size, 7 is usually best, consult bloom papers
+	HasARCCacheSize      int // 32 bytes
+}
+
+func DefaultCacheOpts() CacheOpts {
+	return CacheOpts{
+		HasBloomFilterSize:   512 * 8 * 1024,
+		HasBloomFilterHashes: 7,
+		HasARCCacheSize:      64 * 1024,
+	}
+}
--- a/blocks/blockstore/caching.go
+++ b/blocks/blockstore/caching.go
@@ -1,5 +1,11 @@
 package blockstore
 
+import (
+	"errors"
+
+	context "gx/ipfs/QmZy2y8t9zQH2a1b8q2ZSLKp17ATuJoCNxxyMFG5qFExpt/go-net/context"
+)
+
 // Next to each option is it aproximate memory usage per unit
 type CacheOpts struct {
 	HasBloomFilterSize   int // 1 bit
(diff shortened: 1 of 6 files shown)


blocks/blockstore/caching.go
```diff
@@ -0,0 +1,47 @@
...
+type CacheOpts struct {
+	HasBloomFilterSize   int // 1 bit
+	HasBloomFilterHashes int // No size, 7 is usually best, consult bloom papers
+	HasARCCacheSize      int // 32 bytes
+	BlockARCCacheSize    int // 512KiB max
+}
+
+func DefaultCacheOpts() CacheOpts {
+	return CacheOpts{
+		512 * 8 * 1024,
```
whyrusleeping: dont leave out the field names please

core/builder.go
```diff
@@ -27,6 +27,10 @@ type BuildCfg struct {
 	// If online is set, the node will have networking enabled
 	Online bool
 
+	// If permement then node should run more expensive processes
```
whyrusleeping: permanent *

whyrusleeping: also, when is this set to true?

Kubuxu: When node is created by `ipfs daemon` command.

Kubuxu:
@whyrusleeping sorry for messing it up at first try. This should be working as intended.

blocks/blockstore/caching.go
```diff
@@ -0,0 +1,45 @@
...
+		HasBloomFilterSize:   512 * 8 * 1024,
+		HasBloomFilterHashes: 7,
+		HasARCCacheSize:      64 * 1024,
+	}
+}
+
+func CachedBlockstore(bs GCBlockstore,
+	ctx context.Context, opts CacheOpts) (cbs GCBlockstore, err error) {
+	if ctx == nil {
+		ctx = context.TODO() // For tests
```
whyrusleeping: I dont like this, youre only saving yourself typing this in two other places.

cmd/ipfs/daemon.go
```diff
@@ -229,7 +229,8 @@ func daemonFunc(req cmds.Request, res cmds.Response) {
 
 	// Start assembling node config
 	ncfg := &core.BuildCfg{
-		Repo: repo,
+		Repo:      repo,
+		Permament: true,
```
whyrusleeping: Since this is temporary, lets add a comment here indicating that.

Kubuxu:
@whyrusleeping updated

whyrusleeping:
This LGTM

core/builder.go
```diff
@@ -27,6 +27,10 @@ type BuildCfg struct {
 	// If online is set, the node will have networking enabled
 	Online bool
 
+	// If permament then node should run more expensive processes
+	// that will improve performance in long run
+	Permament bool
```
jbenet: this is not a good name. the node wont ever be `Permanent`. you're using it here to signify "NotEphemeral", so just use `Ephemeral` and set it to false.

blocks/blockstore/caching.go
```diff
@@ -0,0 +1,42 @@
...
+
+	if opts.HasBloomFilterSize < 0 || opts.HasBloomFilterHashes < 0 ||
+		opts.HasARCCacheSize < 0 {
+		return nil, errors.New("all options for cache need to be greater than zero")
+	}
+
+	if opts.HasBloomFilterSize != 0 && opts.HasBloomFilterHashes == 0 {
+		return nil, errors.New("bloom filter hash count can't be 0 when there is size set")
+	}
+	if opts.HasBloomFilterSize != 0 {
```
jbenet: `> 0` -- reduces the failure cases (negative numbers)

core/builder.go
Kubuxu: Didn't use `Ephemeral` as it would be hard to find all tests and places where node is created.

This part is on my list to refactor for the split to: `Online`, `Offline` and `Ephemeral`.

jbenet: > split to: `Online`, `Offline` and `Ephemeral`

those aren't good distinctions. an ephemeral node may be online and may be offline.

Kubuxu: Currently we don't have Online Ephemeral, but very good point.

PR #2957, opened 2016-07-10
core/commands/config: do not show private key on local network

Kubuxu:
**cc** @lgierth @whyrusleeping @jbenet

### DIFF
--- a/core/commands/config.go
+++ b/core/commands/config.go
@@ -161,7 +161,7 @@ included in the output of this command.
 			return
 		}
 
-		cfg["Identity"].(map[string]interface{})["PrivKey"] = nil
+		delete(cfg["Identity"].(map[string]interface{}), "PrivKey")
 
 		output, err := config.HumanOutput(cfg)
 		if err != nil {
--- a/core/commands/config.go
+++ b/core/commands/config.go
@@ -58,6 +58,14 @@ Set the value of the 'datastore.path' key:
 		args := req.Arguments()
 		key := args[0]
 
+		// This is a temporary fix until we move the private key out of the config file
+		switch key {
+		case "Identity", "Identity.PrivKey":
+			res.SetError(fmt.Errorf("cannot show or change private key through API"), cmds.ErrNormal)
+			return
+		default:
+		}
+
 		r, err := fsrepo.Open(req.InvocContext().ConfigRoot)
 		if err != nil {
 			res.SetError(err, cmds.ErrNormal)
@@ -134,18 +142,34 @@ included in the output of this command.
 	},
 
 	Run: func(req cmds.Request, res cmds.Response) {
-		filename, err := config.Filename(req.InvocContext().ConfigRoot)
+		fname, err := config.Filename(req.InvocContext().ConfigRoot)
 		if err != nil {
 			res.SetError(err, cmds.ErrNormal)
 			return
 		}
 
-		output, err := showConfig(filename)
+		data, err := ioutil.ReadFile(fname)
 		if err != nil {
 			res.SetError(err, cmds.ErrNormal)
 			return
 		}
-		res.SetOutput(output)
+
+		var cfg map[string]interface{}
+		err = json.Unmarshal(data, &cfg)
+		if err != nil {
+			res.SetError(err, cmds.ErrNormal)
+			return
+		}
+
+		cfg["Identity"].(map[string]interface{})["PrivKey"] = nil
+
+		output, err := config.HumanOutput(cfg)
+		if err != nil {
+			res.SetError(err, cmds.ErrNormal)
+			return
+		}
+
+		res.SetOutput(bytes.NewReader(output))
 	},
 }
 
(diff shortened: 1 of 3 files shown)


core/commands/config.go
```diff
@@ -58,6 +58,14 @@ Set the value of the 'datastore.path' key:
 		args := req.Arguments()
 		key := args[0]
 
+		// This is a temporary fix until we move the private key out of the config file
+		switch key {
+		case "Identity", "Identity.PrivKey":
```
jbenet: - may be lowercased.
- can just deny all of identity

```go
if strings.HasPrefix(strings.ToLower(key), "identity")  {
```

test/sharness/t0021-config.sh
```diff
@@ -71,6 +71,46 @@ test_config_cmd() {
...
+  test_expect_success "output looks good" '
+       test_cmp ident_exp ident_out
+  '
+
+  test_expect_success "'ipfs config show' doesn't include privkey" '
+       ipfs config show > show_config &&
+       test_expect_code 1 grep PrivKey show_config
+  '
+
+  test_expect_success "'ipfs config replace' injects privkey back" '
```
jbenet: try once where the keys are lowercased?

core/commands/config.go
whyrusleeping: @jbenet you cant look it up by the lowercase key, but we can add that to the permissions check.

Kubuxu: `ipfs config Identity.PeerID` is a valid usecase.

test/sharness/t0021-config.sh
Kubuxu: added

whyrusleeping:
@lgierth if you could review and give a LGTM here that would be great

core/commands/config.go
```diff
@@ -219,22 +244,20 @@ func getConfig(r repo.Repo, key string) (*ConfigField, error) {
 }
 
 func setConfig(r repo.Repo, key string, value interface{}) (*ConfigField, error) {
-	err := r.SetConfigKey(key, value)
+	keyF, err := getConfig(r, "Identity.PrivKey")
+	if err != nil {
+		return nil, fmt.Errorf("Failed to get PrivKey")
```
RichardLitt: Lower case failed

core/commands/config.go
```diff
@@ -134,18 +143,40 @@ included in the output of this command.
...
+			return
+		}
+
+		idmap, ok := cfg["Identity"].(map[string]interface{})
+		if !ok {
+			res.SetError(fmt.Errorf("config has no identity"), cmds.ErrNormal)
+			return
+		}
+
+		delete(idmap, "PrivKey")
```
jbenet: - this wont work if the map has `privkey`. this should be case insensitive.
- i know this comes from our own structs, so it should be capitalized. 
- but all someone has to do is change the capitalization to `Privkey` there.
- come to think of it, this should be either:
   - a constant there, in that file where the struct is
   - or derived from the struct using reflection, to be very damn sure that doesn't get changed around these checks.

core/commands/config.go
```diff
@@ -219,22 +250,20 @@ func getConfig(r repo.Repo, key string) (*ConfigField, error) {
...
-	data, err := ioutil.ReadFile(filename)
+	privkey := keyF.Value
+	err = r.SetConfigKey(key, value)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to set config value: %s (maybe use --json?)", err)
 	}
-
-	return bytes.NewReader(data), nil
+	err = r.SetConfigKey("Identity.PrivKey", privkey)
```
jbenet: - this code is trying to "set the privatekey again" with the original value, to be sure the user did not change it.
- this is an incorrect way to do this
- and is subject to data races.
- and is subject to "triggered crash" problem (if i can cause a crash _after_ the first write and before the second, i can set it to disk successfully. given i can trigger a crash, i can achieve this by doing it over and over again until i get it right.)
- the right way to do this is to sanitize the input and not let it set `Identity.PrivKey` at all.
- I know it is tricky because the value may be `Identity.PrivKey`, or the whole `Identity` block, or the whole config.
- The "setting it again" thing should not be relied upon for security.
- I would probably "set the private key" into the incoming value.
- Another option is removing `Identity` entirely (not just the private key), after all, the `Identity` is tied to the private key and shouldn't be mutable either...

jbenet: it should probably also error similarly to `replaceConfig` if the incoming value has a protected field.

test/sharness/t0021-config.sh
jbenet: @Kubuxu where is it added? not seeing it

core/commands/config.go
lgierth: I think we should use this opportunity to remove the private key from the config file altogether. `Identity.PeerID` can still remain as a readonly attribute for compat. The preferred way of getting one's own PeerID is `:5001/api/v0/id` and we should state that somewhere.

jbenet: Yeah. maybe we can do another migration (4-5) that moves the key to `$repo/keys/<hash-of-key>` and symlinks `id -> <hash-of-key>`. That would stop this madness.

core/commands/config.go
Kubuxu: I can also loop the map, compare keys to `privkey` case insensitive and remove that. SGTY?

test/sharness/t0021-config.sh
Kubuxu: Oh, I added test case for lower case key in the config file, I will add test case for `ipfs config replace` with lower case.

Kubuxu: Adding in: https://github.com/ipfs/go-ipfs/pull/3141

core/commands/config.go
whyrusleeping: I will address this issue here in case any other callers to `setConfig` come up, but we *do* guard against this entire scenario occuring by filtering the input keys in the config command itself.

PR #2960, opened 2016-07-10
cache encoded data when reading dag nodes from disk

whyrusleeping:
License: MIT

### DIFF
--- a/core/commands/object/patch.go
+++ b/core/commands/object/patch.go
@@ -91,7 +91,7 @@ the limit will not be respected by the network.
 			return
 		}
 
-		rootnd.Data = append(rootnd.Data, data...)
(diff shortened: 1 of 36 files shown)


Kubuxu:
LGTM

core/commands/object/patch.go
```diff
@@ -287,9 +287,7 @@ to a file containing 'bar', and returns the hash of the new object.
 
 		var createfunc func() *dag.Node
 		if create {
-			createfunc = func() *dag.Node {
-				return &dag.Node{Data: ft.FolderPBData()}
-			}
+			createfunc = ft.EmptyDirNode
```
jbenet: - it seems this is making a unixfs dir the default intermediary / create function.
- this is probably not what we want for all patch ops.

PR #2961, opened 2016-07-10
Increase channel buffer size in blockstore.AllKeysChan().

kevina:
In order for AllKeysChan to really benefit from the larger buffer size for key only queries in the datastore it should increase it own channel buffer to the same size as the key only queries channel buffer.


kevina:
@Kubuxu this should help to speed up the initial Bloom filter creation time as you use this method.

Kubuxu:
Thanks, if tests pass, LGTM.

Issue #2914, opened 2016-06-28
`ipfs block rm`


diasdavid:
Currently, the only way of deleting blocks in go-ipfs is through `repo gc`. In js-ipfs land, since the Block Service exposes an interface to delete, we also have exposed that on the block API. It is very convenient for testing and a quick way to delete a specific block.

Can we expose this feature in go-ipfs too?

Kubuxu:
Is should be really easy as we also have `Delete` call in blockstore. Problem with deleting block is that it might be component of different file or directory and as user would probably expect the removal of all blocks connected to file or directory it is even bigger issues.

diasdavid:
@Kubuxu that would be something for the files api. The block API is really well scoped as the put and get only also return or add a single block

Kubuxu:
Ahh, right. So yeah, `ipfs block rm` should be possible and quite easy to do.

kevina:
How should removing pinned blocks be handled?  Disallow?  Allow and ignore pin?  (This could then cause many problems with ipfs commands, for example the garbage collector will abort) Allow and attempt to do something about the pin?  (Not so easy unless the pin a a direct pin.)

whyrusleeping:
Yeah, we need to make sure we don't remove blocks that are pinned.

diasdavid:
Good point. If a block is pinned, it should error saying that it is pinned

kevina:
I agree at least by default.  However, for performance reasons, we might want to add an option to disable this detection if the user knows what they are doing.

I am not 100% sure I understand the pining code but it seams that to detect indirect pins all recursive pins need to be walked, thus making pin detection expensive.  It looks like there might be some caching going on in the datastore under the prefix 'local/pins/indirect/keys/' but it doesn't seam like this is used  by [`pin.IsPinnedWithType()`](https://github.com/ipfs/go-ipfs/blob/master/pin/pin.go#L203).

whyrusleeping:
Yeah, pin detection is expensive. We can optimize with various caching structures though in the future.

kevina:
I know the issues involved so I take responsibility for this one.

RichardLitt:
Participating.

diasdavid:
Any update on this one?  :)

kevina:
@diasdavid please see #2962

whyrusleeping:
we have this now :)

diasdavid:
woot!

PR #2962, opened 2016-07-10
Add "ipfs block rm" command.

kevina:
This is a basic implementation for #2914.

It accepts multiple blocks to remove, however that does not provide any performance benifit just yet as it calls `pins.IsPinned()` for each block, which traverses every recursive pin to check for indirect pins.  This operation costs is O(P*N) where P is the number of pinned blocks and N is the number of blocks to be removed.  With a little effort this can be brought down to O(P) by checking for multiple pins at once.

Since the pin check is expensive it provides an option to ignore pins.  Removing a pinned block will break any command that checks for pins.  Right now if this happens recovering from the situation is very difficult.  At very least a command should be provided to list any broken or missing pins.  An additional optional command can be provided to attempt to repair recursive pins.

TODO:
- [x] Provide command to check for multiple pins at once

No longer relevant:
- [] Provide a `ipfs pin verify` command to list broken or missing pins

Optional and No longer relevant
- [] Provide a command to repair pins (I already wrote similar code in #2634)

### DIFF
--- a/core/commands/block.go
+++ b/core/commands/block.go
@@ -200,15 +200,7 @@ It takes a list of base58 encoded multihashs to remove.
 	Arguments: []cmds.Argument{
 		cmds.StringArg("hash", true, true, "Bash58 encoded multihash of block(s) to remove."),
 	},
-	Options: []cmds.Option{
-		cmds.BoolOption("ignore-pins", "Ignore pins.").Default(false),
-	},
 	Run: func(req cmds.Request, res cmds.Response) {
-		ignorePins, _, err := req.Option("ignore-pins").Bool()
-		if err != nil {
-			res.SetError(err, cmds.ErrNormal)
-			return
-		}
 		n, err := req.InvocContext().GetNode()
 		if err != nil {
 			res.SetError(err, cmds.ErrNormal)
@@ -225,9 +217,6 @@ It takes a list of base58 encoded multihashs to remove.
 		go func() {
 			defer close(outChan)
 			pinning := n.Pinning
-			if ignorePins {
-				pinning = nil
-			}
 			err := rmBlocks(n.Blockstore, pinning, outChan, keys)
 			if err != nil {
 				outChan <- &RemovedBlock{Error: err.Error()}
@@ -271,25 +260,15 @@ type RemovedBlock struct {
 	Error string `json:",omitempty"`
 }
 
-// pins may be nil
 func rmBlocks(blocks bs.GCBlockstore, pins pin.Pinner, out chan<- interface{}, keys []key.Key) error {
-	var unlocker bs.Unlocker
-	defer func() {
-		if unlocker != nil {
-			unlocker.Unlock()
-		}
-	}()
-	stillOkay := keys
-	if pins != nil {
-		// Need to make sure that some operation that is
-		// finishing with a pin is ocurr simultaneously.
-		unlocker = blocks.GCLock()
-		var err error
-		stillOkay, err = checkIfPinned(pins, keys, out)
-		if err != nil {
-			return fmt.Errorf("pin check failed: %s", err)
-		}
+	unlocker := blocks.GCLock()
+	defer unlocker.Unlock()
+
+	stillOkay, err := checkIfPinned(pins, keys, out)
+	if err != nil {
+		return fmt.Errorf("pin check failed: %s", err)
 	}
+
 	for _, k := range stillOkay {
 		err := blocks.DeleteBlock(k)
 		if err != nil {
--- a/core/commands/block.go
+++ b/core/commands/block.go
@@ -13,6 +13,7 @@ import (
 	key "github.com/ipfs/go-ipfs/blocks/key"
 	cmds "github.com/ipfs/go-ipfs/commands"
 	"github.com/ipfs/go-ipfs/pin"
+	ds "gx/ipfs/QmTxLSvdhwg68WJimdS6icLPhZi28aTp6b7uihC2Yb47Xk/go-datastore"
 	mh "gx/ipfs/QmYf7ng2hG5XBtJA3tN34DQ2GUN5HNksEw1rLDkmr6vGku/go-multihash"
 	u "gx/ipfs/QmZNVWh8LLjAavuQ2JXuFmuYH3C11xo988vSgp7UQrTRj1/go-ipfs-util"
 )
@@ -200,6 +201,10 @@ It takes a list of base58 encoded multihashs to remove.
 	Arguments: []cmds.Argument{
 		cmds.StringArg("hash", true, true, "Bash58 encoded multihash of block(s) to remove."),
 	},
+	Options: []cmds.Option{
+		cmds.BoolOption("force", "f", "Ignore nonexistent blocks.").Default(false),
+		cmds.BoolOption("quiet", "q", "Write minimal output.").Default(false),
+	},
 	Run: func(req cmds.Request, res cmds.Response) {
 		n, err := req.InvocContext().GetNode()
 		if err != nil {
@@ -207,6 +212,8 @@ It takes a list of base58 encoded multihashs to remove.
 			return
 		}
 		hashes := req.Arguments()
+		force, _, _ := req.Option("force").Bool()
+		quiet, _, _ := req.Option("quiet").Bool()
 		keys := make([]key.Key, 0, len(hashes))
 		for _, hash := range hashes {
 			k := key.B58KeyDecode(hash)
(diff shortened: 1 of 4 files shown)


RichardLitt:
Nice! Subscribing, for the http-api-spec.

core/commands/block.go
```diff
@@ -185,3 +188,99 @@ func getBlockForKey(req cmds.Request, skey string) (blocks.Block, error) {
...
+`,
+	},
+	Arguments: []cmds.Argument{
+		cmds.StringArg("hash", true, true, "Bash58 encoded multihash of block(s) to remove."),
+	},
+	Options: []cmds.Option{
+		cmds.BoolOption("ignore-pins", "Ignore pins.").Default(false),
+	},
+	Run: func(req cmds.Request, res cmds.Response) {
+		ignorePins, _, err := req.Option("ignore-pins").Bool()
```
whyrusleeping: having this option makes me very nervous. Things break rather spectaculary when we have missing blocks that are pinned

core/commands/block.go
```diff
@@ -185,3 +188,99 @@ func getBlockForKey(req cmds.Request, skey string) (blocks.Block, error) {
...
+				pinning = nil
+			}
+			err := rmBlocks(n.Blockstore, pinning, wtr, keys)
+			if err != nil {
+				wtr.CloseWithError(fmt.Errorf("Some blocks not deleted: %s", err))
+			} else {
+				wtr.Close()
+			}
+		}()
+		res.SetOutput(rdr)
```
whyrusleeping: lets have our output be a channel, and send blocks over it like the refs command and such do.

core/commands/block.go
```diff
@@ -185,3 +188,99 @@ func getBlockForKey(req cmds.Request, skey string) (blocks.Block, error) {
...
+}
+
+type noopLocker struct{}
+
+func (noopLocker) Unlock() {}
+
+// pins may be nil
+func rmBlocks(blocks bs.GCBlockstore, pins pin.Pinner, out io.Writer, keys []key.Key) error {
+	var unlocker bs.Unlocker = noopLocker{}
+	defer unlocker.Unlock()
```
whyrusleeping: this reference will be bound such that it will always be called on the noop lock, even when the value in the unlocker variable is changed. If you want to do things this way, you will need to capture the variable in a closure:
```
defer func() {
    unlocker.Unlock()
}()
```

core/commands/block.go
kevina: I Know.  But I see is as necessary, at least until we can get pinning operations to be O(1).  Please see my initial description on this pr and #2963.

core/commands/block.go
kevina: Just sending hashes of deleted blocks is not flexible enough.  I would like a way to send error conditions.  For example I might in the future want to list all the blocks I fine pinned.  Is is possible to encode this over a channel but it will be more complicated so I opted to just use io.Pipe for now.

core/commands/block.go
kevina: Oops.  Will Fix!

core/commands/block.go
whyrusleeping: This doesnt work the way youre expecting it to. The error won't be written over the pipe, it just gets passed to the next thing that calls read or write on the same pipe object.

kevina: I am not sure I understand what you are getting at. By error condensation I mean just sending generic plain text messages over the pipe and than using `wtr.CloseWithError()` to just indicate that they where some errors.  I have tested `wtr.CloseWithError()` works as expected in the sharness tests.

whyrusleeping: Looking this over again i see, Errors returned from readers in the commands lib are generally treated as more serious 'system level' errors (you will see an error message on the daemon). We don't normally use them for 'user' errors as mixing real errors and 'expected' errors makes handling them all tricky. If you disagree with this, we can rethink how that pipeline should work.

returning errors this way is handled by setting an http trailer on the request which makes handling them in client API libs more complicated.

whyrusleeping:
I'm still not comfortable having the option to ignore pins here, letting a single command have the power to corrupt your repo isnt a very good idea.

core/commands/block.go
kevina: It is more of a lack of understanding than a disagreement at this point.  Could you give me some code pointers so I can better understand what is going on.  I do this sort of thing all over my filestore code so I need to know why this is bad and what I can do to correct it.

whyrusleeping: Think about it this way, if you use `CloseWithError` that manifests as a failed call to `Read` somewhere. What sort of errors would you expect to prevent you from reading a stream of data? Examples where we use this are with `ipfs cat`, if we fail to fetch some of the data for ipfs cat, it will throw an error on the reader, which makes sense because you are failing to read the file youre catting.

In this case, youre failing to read a stream of data because you failed to delete a block? Those feel very unrelated to eachother.

kevina:
@whyrusleeping please have another look.  I send the results as a channel and use PostRun now to process the results.  I also avoid aborting on a non-fatal error and instead just report that the block can't be deleted and continue on to the next block.

I am okay with removing the option to ignore pins, but I want to fix the pin checks so it is `O(p)` instead of `O(p*n)` where `p` is the number of pinned blocks and `n` is the number of blocks to be removed.

whyrusleeping:
Yeah, if we're removing that option then that all looks pretty good.

kevina:
@whyrusleeping okay, for now I removed the option to ignore pins

If you are happy with this fell free to merge.

diasdavid:
@whyrusleeping do you need anything else for this PR? Looking forward to have https://github.com/ipfs/go-ipfs/issues/2914 so that https://github.com/ipfs/js-ipfs-api/pull/305 can be finished with regards to the Block API :)

whyrusleeping:
This looks pretty good to me, it sucks that we have to be so hacky with the commands lib `PostRun` stuff. Maybe we can figure out a better way to restructure that code in the future.

Only thing i'd want to see added here is a test of removing multiple blocks, where a subset of them are pinned.

kevina:
@whyrusleeping okay I added a test that deleted both pinned and un-pinned and non-existent blocks and made sure that the valid, un-pinned blocks where removed.

@Kubuxu I fixed a bug in the arccache.DeleteBlock() method in bf7c5b3.  It seamed a small enough issue that it wasn't worth a separate pull request.

whyrusleeping:
@kevina i'm not so sure that was a bug in the arccache. I guess it depends on your perspective, if i call delete on something that doesnt exist, should it ever be a problem that it doesnt exist?

kevina:
@whyrusleeping the cache changes the behavior of the DeleteBlock() method which does return an error if a key was not found, therefore I consider it a bug.  The cache itself relies in DeleteBlock() returning an error if a block was not found.  Also I would prefer that "block rm" returns an error if a non-existent block is deleted in case I mistyped the key (or more likely I copy and pasted a key but missed a couple of characters.)  If this becomes a blocker I can remove the fix and not test for deleting a non-existent block in my test case.

@Kubuxu, your view on this?

jbenet:
Clarity on what is meant is important:

- "make it so X is not on my local repo" should be idempotent
- "check X exists in my local repo, then remove X from my local repo" not
idempotent

In general, idempotency is a very important property to have. An op like
this should be by default idempotent. You could use a flag to ask it to
signal when a block was not found.

ipfs block rm --error-if-not-present <cid>

kevina:
@jbenet the behavior of `ipfs block rm` is consistent with that of the the `rm` command.  If a block was not found it will complain but still remove any (non-pinned) blocks that are found.

In other words `block rm` is idempotent, the only difference will be the final status code.  All blocks that it can remove will always be removed.

whyrusleeping:
Oh, my bad, I read the diff backwards.

whyrusleeping:
@jbenet agreed that idempotency is important, But we need to clearly define the expectations of the datastore interface. Currently, all the datastore implementations will return `ErrNotFound` when delete is called on a non-existent key. If we want to change that, we need to go through and ensure that *every* datastore implementation behaves the same way.

whyrusleeping:
This LGTM. I'll wait for @Kubuxu to review as well

jbenet:
The datastore is not the `ipfs blocks` command. The datastore Delete wants
to give you more information.

I guess you could do what rm does and add -f

jbenet:
`rm -f`

whyrusleeping:
@jbenet okay, i thought we were having a discussion about the changes made to the `Delete` method.

kevina:
@jbenet I added a "-f" option.

I am of the strong opinion that by default "block rm" should report some sort of error when a block is not found.  Doing a quick Google "idempotent delete" brought up this stack overflow question: http://stackoverflow.com/questions/4088350/is-rest-delete-really-idempotent.  Multiple people state that idempotency is about side effects not the status code, that is "you can send the request more than once without additional changes to the state of the server".

whyrusleeping:
Alright, this all LGTM. Thanks a bunch @kevina !

PR #2966, opened 2016-07-12
Build on OpenBSD

polezaivsani:


whyrusleeping:
I wish we could do wildcards "*BSD", LGTM

Issue #2590, opened 2016-04-19
Possible to only accept connections from a particular IPFS nodes?


kevinmartinmauro:
Is it, or will it ever be, possible to only accept connections from a particular IPFS node? Could this be used as a DDOS/anti-spam feature?

Stebalien:
Yes but that would severely limit the usefulness of IPFS. Eventually, the bitswap protocol is supposed to have some sort of reputation system that should fix this. Basically, you'd prioritize nodes with good reputation over nodes with bad/unknown reputation.

whyrusleeping:
In the next couple months i'm going to be working on 'private networks' where you only accept connections from people who have a certain passcode. Would this suit your usecase?

alikic:
Just found this issue. I am also interested in having a decentralized private storage. I opened https://github.com/ipfs/faq/issues/128 few days ago, but no feedback so far. Private network assumes that peers accept connection requests only from trusted peers, and today it's not possible. Typically trust among parties is established by having a central authority trusted by all parties (e.g. root certificate authority in the TLS case), which then extends trust to other entities (e.g. by issuing certificates in the TLS case). The parties can exchange evidence that they are trusted by the central authority, and then they can trust each other.

From the operational point of view, it is usually very important that trust with the central authority can be revoked / renewed periodically (e.g. by expiring/reissuing/revoking certificates in the TLS case).

I know there are efforts underway around moving communication protocol to TLS, and perhaps some other efforts to improve trust between peers, but they seem to be very low priority. As I want to have something now, I came up with a simple trust solution that should be easy to implement. The central authority would be an application that holds information on all peers in the private network. Peers would have to register with the authority, so they can login there to pull information about all trusted peers. During secure connection handshake, a peer would expect to receive from the other peer a piece of evidence that the other party has indeed logged into the central authority (e.g. the peer's secret).

After looking at the code a bit, it seems to me this can be achieved by introducing two new concepts (interfaces):
1. BootstrapProvider
- default implementation would read the list of trusted peers from the local file, the same as now
- other implementations could read it from some REST service after logging into it
2. TrustManager
- It would be used by runHandshake() in protocol.go (go-lib-secio). Default implementation would keep doing what it is doing now.
- other implementations could check additional information (e.g. check the peer-specific secret that only the central authority knows) before continuing with handshake.

[...]

Issue #1633, opened 2015-09-02
private networks -- shared key or with a PKI


whyrusleeping:

hmeine:
Can you elaborate?  I am currently thinking about exchanging private data via IPFS, and how to ensure that the data can stay within a controlled cluster.  (Think medical images in a hospital, which has strict requirements w.r.t. leak prevention.)

jbenet:
@hmeine we will allow creation of _completely private_ IPFS networks. the plan is to:

1. (trivial impl): use a **single shared** symmetric encryption key, and encrypt the transport with it
2. (harder impl): use a **PKI and only connect to trusted nodes** who either 
  - (a) present a valid certificate, 
  - (b) can be checked online 
  
  (we want to support both PKI use cases because both are good models, with strong proponents and importantly different threat models)

This issue is really about mode (1) (shared key). but we will implement both eventually. (1) should work for you fine, but will leave it up to you to do key rotation to ensure long term security (as nodes join / leave the network, or as human operators join / leave the organization)

hmeine:
Thanks for the explanation.  This would be really useful for our use case, it seems.
It also looks relevant to https://github.com/ipfs/go-ipfs/issues/961, don't you think?

scriptjs:
Hi. You mentioned the end of Sept last year that this was coming soon. Can you provide an update of where we are and how it fits with priorities. Am interested in this and want it to be compatible if possible or contribute. Do you have a blueprint?

ahnick:
I'm also interested to know the current state of private network support.

bronger:
http://ilpubs.stanford.edu:8090/626/1/2003-74.pdf discusses a hierarchical DHT.  This generates an onion-like structure of accessability domains.  I think this is a very flexible and elegant way to achieve privacy in a distributed system.

Originally, I wanted to implement this from scratch for my usecase (research data publication with institute file space < university file space < global filespace), however, since I found out about IPFS, I wonder whether something like that can also be achieved in the IPFS ecosystem.

whyrusleeping:
@troyronda has a proposal about trust management here: https://github.com/securekey/go-ipfs/blob/a2c73ac501f602a2dbbe5260ec7e7fec4d30ee60/docs/trusted-private-network.md

bronger:
Where can one comment on this proposal?

troyronda:
The thread is here: https://github.com/ipfs/notes/issues/146

salsa-dev:
[...]

PR #2968, opened 2016-07-13
Proposal for the Trusted Private Network feature

alikic:
This feature is discussed at:
https://github.com/ipfs/go-ipfs/issues/2590
https://github.com/ipfs/go-ipfs/issues/1633

This is a design proposal, I am looking for feedback. Working code is also available, I'll submit it once the design (with possible agreed modifications) is accepted.


RichardLitt:
@alikic This is great. 

However, this isn't actionable here, at the moment. Can you open an issue and put this in [ipfs/notes](//github.com/ipfs/notes)? We have more to discuss before we implement private networks with go-ipfs, and everything that has to deal with higher level concerns (not actionable features or bugs -- for instance, multikey formats and signing) is being discussed in [notes](//github.com/ipfs/notes), as it has repercussions for multiple implementations.  

I'm going to close this for now. Looking forward to talking about this over there! And again, thanks. :)

PR #2971, opened 2016-07-15
gateway: degrade error in gateway to log to reduce noise

Kubuxu:
It logs all errors including expired IPNS keys and other non important
errors.

License: MIT
Signed-off-by: Jakub Sztandera kubuxu@protonmail.ch

### DIFF
--- a/core/corehttp/gateway_handler.go
+++ b/core/corehttp/gateway_handler.go
@@ -168,7 +168,7 @@ func (i *gatewayHandler) getOrHeadHandler(ctx context.Context, w http.ResponseWr
 		dir = true
 	case namesys.ErrResolveFailed:
 		// Don't log that error as it is just noise
-		w.WriteHeader(http.StatusBadRequest)
+		w.WriteHeader(http.StatusInternalServerError)
 		fmt.Fprintf(w, "Path Resolve error: %s", err.Error())
 		log.Info("Path Resolve error: %s", err.Error())
 		return
--- a/core/corehttp/gateway_handler.go
+++ b/core/corehttp/gateway_handler.go
@@ -160,22 +160,29 @@ func (i *gatewayHandler) getOrHeadHandler(ctx context.Context, w http.ResponseWr
 
 	dr, err := i.api.Cat(ctx, urlPath)
 	dir := false
-	if err == coreiface.ErrIsDir {
+	switch err {
+	case nil:
+		// core.Resolve worked
+		defer dr.Close()
+	case coreiface.ErrIsDir:
 		dir = true
-	} else if err == coreiface.ErrOffline {
-		w.WriteHeader(http.StatusServiceUnavailable)
-		fmt.Fprint(w, "Could not resolve path. Node is in offline mode.")
-		return
-	} else if err == namesys.ErrResolveFailed {
+	case namesys.ErrResolveFailed:
 		// Don't log that error as it is just noise
 		w.WriteHeader(http.StatusBadRequest)
 		fmt.Fprintf(w, "Path Resolve error: %s", err.Error())
+		log.Info("Path Resolve error: %s", err.Error())
 		return
-	} else if err != nil {
+	case coreiface.ErrOffline:
+		if !i.node.OnlineMode() {
+			w.WriteHeader(http.StatusServiceUnavailable)
+			fmt.Fprint(w, "Could not resolve path. Node is in offline mode.")
+			return
+		}
+		fallthrough
+	default:
+		// all other erros
 		webError(w, "Path Resolve error", err, http.StatusBadRequest)
 		return
-	} else {
-		defer dr.Close()
 	}
 
 	etag := gopath.Base(urlPath)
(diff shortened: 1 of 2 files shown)


core/corehttp/gateway_handler.go
```diff
@@ -506,7 +506,7 @@ func webError(w http.ResponseWriter, message string, err error, defaultCode int)
 
 func webErrorWithCode(w http.ResponseWriter, message string, err error, code int) {
 	w.WriteHeader(code)
-	log.Errorf("%s: %s", message, err) // TODO(cryptix): log errors until we have a better way to expose these (counter metrics maybe)
+	log.Infof("%s: %s", message, err) // TODO(cryptix): log until we have a better way to expose these (counter metrics maybe)
```
RichardLitt: Can we split out this TODO into an issue?

lgierth:
I think this is probably insufficient. Some of these are legitimate errors, some are just totally normal from the gateway's perspective. Error handling in the gateway generally needs to be reworked.

lgierth:
But :+1: LGTM, since it reduces log noise.

Kubuxu:
It does and that is what #2972 is for.

whyrusleeping:
I agree with @lgierth here, i dont think this is what we want to do. We should filter out explictly which errors we want to demote, this was here for a reason, we want to know when something has gone wrong and handle it accordingly.

Kubuxu:
Reworked it.

whyrusleeping:
We might also want to handle context deadline exceeded errors (i see these somewhat often too) and add some tests.

jbenet:
+1 to being very explicit about what's getting demoted and making
absolutely sure it's not a problematic condition users should be made aware
of. (What @lgierth said)

Kubuxu:
Updated.

core/corehttp/gateway_handler.go
```diff
@@ -153,12 +154,23 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
...
 	nd, err := core.Resolve(ctx, i.node, path.Path(urlPath))
-	// If node is in offline mode the error code and message should be different
-	if err == core.ErrNoNamesys && !i.node.OnlineMode() {
-		w.WriteHeader(http.StatusServiceUnavailable)
-		fmt.Fprint(w, "Could not resolve path. Node is in offline mode.")
+	switch err {
+	case nil:
+		// core.Resolve worked
+	case core.ErrNoNamesys:
+		if !i.node.OnlineMode() {
```
whyrusleeping: What happens if we get ErrNoNamesys and we're 'online' ? Even if this can never happen we should still check that case here and log an error

Kubuxu:
Updated

Kubuxu:
@lgierth can you CR on this.

whyrusleeping:
LGTM, just waiting on a :+1: from @lgierth

core/corehttp/gateway_handler.go
```diff
@@ -153,12 +154,24 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
...
-	// If node is in offline mode the error code and message should be different
-	if err == core.ErrNoNamesys && !i.node.OnlineMode() {
-		w.WriteHeader(http.StatusServiceUnavailable)
-		fmt.Fprint(w, "Could not resolve path. Node is in offline mode.")
+	switch err {
+	case nil:
+		// core.Resolve worked
+	case namesys.ErrResolveFailed:
+		// Don't log that error as it is just noise
+		w.WriteHeader(http.StatusBadRequest)
```
lgierth: 400 doesn't fit here since the 4xx range is for client errors, where the client can amend the request to possibly get a successful response. 502 Bad Gateway or, (less specific) 500 Internal Server Error would be sufficient.

core/corehttp/gateway_handler.go
```diff
@@ -153,12 +154,24 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
...
+	case core.ErrNoNamesys:
+		if !i.node.OnlineMode() {
+			w.WriteHeader(http.StatusServiceUnavailable)
+			fmt.Fprint(w, "Could not resolve path. Node is in offline mode.")
+			return
+		}
+		fallthrough
+	default:
+		// all other erros
 		webError(w, "Path Resolve error", err, http.StatusBadRequest)
```
lgierth: I'm okay to leave this existing line with 400 for now, but the same applies here.

Kubuxu:
@lgierth updated

Kubuxu:
Wait, it broke the tests...

Kubuxu:
@lgierth can you confirm change in behaviour here:
```
--- FAIL: TestGatewayGet (0.01s)
	gateway_test.go:127: http://127.0.0.1:52637
	gateway_test.go:156: got 500, expected 400 from http://localhost:5001/ipns/nxdomain.example.com
```

lgierth:
This LGTM :+1:, I went and resolved the conflicts with the recent coreapi changes, @Kubuxu I thought that's likely okay :)

Sorry for sitting on it so long!

PR #2973, opened 2016-07-15
config: add config option for bloom filter

Kubuxu:
License: MIT

### DIFF
--- a/repo/config/init.go
+++ b/repo/config/init.go
@@ -85,6 +85,7 @@ func datastoreConfig() (Datastore, error) {
 		StorageGCWatermark: 90, // 90%
 		GCPeriod:           "1h",
 		HashOnRead:         false,
+		BloomFilterSize:    0,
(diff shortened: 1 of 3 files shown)


whyrusleeping:
LGTM, thanks

repo/config/init.go
```diff
@@ -85,6 +85,7 @@ func datastoreConfig() (Datastore, error) {
 		StorageGCWatermark: 90, // 90%
 		GCPeriod:           "1h",
 		HashOnRead:         false,
+		BloomFilterSize:    0,
```
jbenet: so this is disabled by default?

Kubuxu: Yes it is, we decided to do that until we get better metrics, repo size caching (important information to size to bloom filter correctly) and bloom filter expansion.

jbenet: SGTM

Issue #2947, opened 2016-07-06
Build: multiple GOPATH components 


djdv:
Version/Platform/Processor information (from `ipfs version --all`):
N/A
Type (bug, feature, meta, test failure, question):
Bug
Area (api, commands, daemon, fuse, etc):
Build system
Priority (from P0: functioning, to P4: operations on fire):
P1

Description:
A patch was merged https://github.com/ipfs/go-ipfs/pull/2808 that resolved this issue but was reverted https://github.com/ipfs/go-ipfs/pull/2862 there was no original issue ticket filed, I am filling this one so that the revert issue (https://github.com/ipfs/go-ipfs/issues/2833) may be closed without dropping awareness of the issue.

Original text by @karalabe from https://github.com/ipfs/go-ipfs/pull/2808 below:

Go's `GOPATH` environment variable can contain multiple paths, separated by a colon `:` similar to the `PATH` env var. In such cases Go uses the first path element to install any packages, however the remaining paths are still used to find dependencies. Because of this it's quite a common setup to have multiple paths listed in GOPATH (e.g. one for development and another for appengine, which has its special installer).

The current Makefile however erroneously assumes that `GOPATH` contains a single path element, and tries to locate go-ipfs within by appending a string to `GOPATH`, which is obviously cause an invalid path, and hence fail the `path_check` rule in the Makefile, even though the required condition is met.

Kubuxu:
I think the problem is that in Windows the path separator would be `;`.

So we would need if we run `Windows && !Cygwin` then change the path separator to `;`.

PR #2975, opened 2016-07-16
Build: Fix Makefile on Windows

djdv:
This request brings back in a reverted commit with a fix on top for platform delimiters.
Before this gets merged though I would like people to test it on various systems, it works for me in msys2 where it did not originally, but I only use a single GOPATH. If anyone could try building with this and report on the results I'd appreciate it.

@ChristianKniep 
Could you run this and see if you have any build issues? (In relation to https://github.com/ipfs/go-ipfs/issues/2947)

@jefft0
Could you run this inside of Cygwin? I'm not sure if Cygwin will get picked up as Windows or not with this condition.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,7 +41,7 @@ bin/gx-go-v%:
 gx_check: ${gx_bin} ${gx-go_bin}
 
 path_check:
-	@bin/check_go_path $(realpath $(shell pwd)) $(realpath $(GOPATH)/src/github.com/ipfs/go-ipfs)
+	@bin/check_go_path $(realpath $(shell pwd)) $(realpath $(addsuffix /src/github.com/ipfs/go-ipfs,$(subst :, ,$(GOPATH))))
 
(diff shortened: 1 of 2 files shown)


lgierth:
@djdv thanks so much for taking this on! :) I'll give it a try with docker tomorrow!

djdv:
I went ahead and tested on Cygwin with a single GOPATH, it built fine, so no breakage there at least.

@lgierth  Looking forward to seeing the results, I didn't want @karalabe's patch to disappear since it worked on *nix systems originally.

jefft0:
`make install` works for me on Cygwin. Thanks.

ChristianKniep:
@djdv Works for me now, thanks!

karalabe:
Works for me too :+1: 

PS: Sorry for not reacting till now, we're having a bit of a crisis at Ethereum for the last month or so.

whyrusleeping:
LGTM, thanks everyone!

PR #2977, opened 2016-07-19
mfs: fix copying into directory with no given filename

whyrusleeping:
There was a bug where running `ipfs files cp /something /adirectory/` would make a file with a zero length name inside `/adirectory`

License: MIT

### DIFF
--- a/test/sharness/t0250-files-api.sh
+++ b/test/sharness/t0250-files-api.sh
@@ -460,6 +460,24 @@ test_files_api() {
 		ipfs files rm -r /test_dir &&
 		ipfs files rm -r /test_file
 	'
+
+	test_expect_success "make a directory and a file" '
+		ipfs files mkdir /adir &&
+		echo "blah" | ipfs files write --create /foobar
+	'
+
+	test_expect_success "copy a file into a directory" '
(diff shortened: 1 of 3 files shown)


RichardLitt:
LGTM.

whyrusleeping:
cc @Kubuxu for another review

RichardLitt:
@whyrusleeping I reviewed it technically, not just for my normal review. The Gocode looks solid. 

Still, double might be good?

whyrusleeping:
:+1: Thanks richard!

Kubuxu:
I was reviewing that ...

test/sharness/t0250-files-api.sh
```diff
@@ -460,6 +460,24 @@ test_files_api() {
...
 	'
+
+	test_expect_success "make a directory and a file" '
+		ipfs files mkdir /adir &&
+		echo "blah" | ipfs files write --create /foobar
+	'
+
+	test_expect_success "copy a file into a directory" '
+		ipfs files cp /foobar /adir/
+	'
```
jbenet: This test case could've been created from the get go. Let's test more thoroughly from the beginning.

Issue #2976, opened 2016-07-18
Send a message for `--version`


RichardLitt:
Type (bug, feature, meta, test failure, question): Feature
Area (api, commands, daemon, fuse, etc): CLI
Priority (from P0: functioning, to P4: operations on fire): P0

Description:

> I type `ipfs --version` a lot, and it tells me nothing just `Error: Unrecognized option 'version'`. Would be nice to be able to have `--version` also return something, just like `ipfs --help` does.

From my issue in https://github.com/ipfs/ipfs/issues/158.

whyrusleeping:
This has already been done apparently

PR #2990, opened 2016-07-26
Alias 'ipfs --version' to 'ipfs version'

csasarak:
Addresses Issue #2976. This is a pretty simple method since it looks like the main ipfs command pretty much just parses and launches the others. If you think it is likely that we'll have more 'alias' type things that people will want on ipfs (not its subcommands) in the future, I can probably write something (like a dict) mapping aliases to real commands.

### DIFF
--- a/cmd/ipfs/main.go
+++ b/cmd/ipfs/main.go
@@ -103,9 +103,13 @@ func main() {
 	}
 
 	// Handle `ipfs help'
-	if len(os.Args) == 2 && os.Args[1] == "help" {
-		printHelp(false, os.Stdout)
-		os.Exit(0)
+	if len(os.Args) == 2 {
+		if os.Args[1] == "help" {
+			printHelp(false, os.Stdout)
+			os.Exit(0)
+		} else if os.Args[1] == "--version" {
+			os.Args[1] = "version"
+		}
 	}
 
 	// parse the commandline into a command invocation
--- a/test/sharness/t0010-basic-commands.sh
+++ b/test/sharness/t0010-basic-commands.sh
@@ -16,9 +16,22 @@ test_expect_success "ipfs version succeeds" '
 	ipfs version >version.txt
 '
 
+test_expect_success "ipfs --version success" '
+    ipfs --version ||
+    test_fsh ipfs --version
+'
+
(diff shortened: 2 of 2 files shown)


Kubuxu:
Please introduce sharness test case for it, so if we do refactor of it in future it doesn't get lost.

csasarak:
I'll look into that tonight, thanks!

RichardLitt:
Thank you, @csasarak!

csasarak:
I've added two tests that I could think of, details are in the commit message.

whyrusleeping:
LGTM

Kubuxu:
SGTM

test/sharness/t0010-basic-commands.sh
```diff
@@ -16,11 +16,24 @@ test_expect_success "ipfs version succeeds" '
...
 test_expect_success "ipfs version output looks good" '
 	egrep "^ipfs version [0-9]+\.[0-9]+\.[0-9]" version.txt >/dev/null ||
 	test_fsh cat version.txt
 '
 
+test_expect_success "ipfs versions matches ipfs --version" '
+    ipfs version > version.txt &&
+    ipfs --version > version2.txt &&
+    diff version2.txt version.txt ||
+    test_fsh ipfs --version
```
chriscool: Maybe use `test_cmp version.txt version2.txt` instead of those last two lines.

cmd/ipfs/main.go
```diff
@@ -103,9 +103,13 @@ func main() {
...
-	if len(os.Args) == 2 && os.Args[1] == "help" {
-		printHelp(false, os.Stdout)
-		os.Exit(0)
+	if len(os.Args) == 2 {
+		if os.Args[1] == "help" {
+			printHelp(false, os.Stdout)
+			os.Exit(0)
+		} else if os.Args[1] == "--version" {
+			os.Args[1] = "version"
+		}
```
jbenet: - this captures most of the times people will write `--version`
- but it may lead to weird behavior `(ipfs --timeout=10s --version` wont work).
- and it doesn't work on the API (making it only a CLI thing, which could cause confusion).
- I'm fine with it, but noting it's hacky

PR #2993, opened 2016-07-26
use batching datastore for providers storage

whyrusleeping:
License: MIT


Kubuxu:
LGTM

routing/dht/providers/providers.go
```diff
@@ -10,16 +10,26 @@ import (
...
 
 	context "gx/ipfs/QmZy2y8t9zQH2a1b8q2ZSLKp17ATuJoCNxxyMFG5qFExpt/go-net/context"
 )
 
+var batchBufferSize = 256
+
+func init() {
+	if flags.LowMemMode {
+		batchBufferSize = 8
+	}
```
jbenet: this is the sort of thing that could yield very odd debug scenarios. can this be controlled via config instead?

PR #2995, opened 2016-07-27
mount: ensure a private key is loaded before mounting /IPNS/

thomas-gardner:
Spinning up a daemon with `--mount --offline` was leading to a
nil-pointer dereference because offline routing doesn't load a private
key.

License: MIT
Signed-off-by: Thomas Gardner tmg@fastmail.com

### DIFF
--- a/fuse/ipns/mount_unix.go
+++ b/fuse/ipns/mount_unix.go
@@ -17,6 +17,12 @@ func Mount(ipfs *core.IpfsNode, ipnsmp, ipfsmp string) (mount.Mount, error) {
 
 	allow_other := cfg.Mounts.FuseAllowOther
 
+	if ipfs.PrivateKey == nil {
+		if err = ipfs.LoadPrivateKey(); err == nil {
+			return nil, err
+		}
+	}
+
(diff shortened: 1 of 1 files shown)


Kubuxu:
@whyrusleeping I've added this to RC.

@thomas-gardner can you try to write a test case for that?

thomas-gardner:
> @thomas-gardner can you try to write a test case for that?

Sure.

Kubuxu:
For purpose of RC we decided not to introduce these changes, instead we blocked out `--mount --offline` flag combination #3022

We plan allowing this flag combination to work in future version.

whyrusleeping:
Can someone pick this up and add a test case for it? @Kubuxu @lgierth @hsanjuan ?

Kubuxu:
The fork no longer exists which screws up Jenkins. Please reopen with fixed sign-offs.

If author doesn't respond in a week, anyone please feel free to recreate functionality of this PR.

----

Ref for Jenkins bug https://issues.jenkins-ci.org/browse/JENKINS-41246

PR #2999, opened 2016-07-27
Cleanup makefiles

Kubuxu:
This can wait for after release but it shouldn't affect the release in any form either.

We might want to limit number of threads on the sharness or make output sequential of sharness because now they run in parallel.

### DIFF
--- a/test/Makefile
+++ b/test/Makefile
@@ -20,7 +20,8 @@ global-deps:
 deps: global-deps bins
 
 clean:
-	rm $(BINS)
+	$(MAKE) -C sharness clean
+	-rm -rf $(BINS)
 
 bins: $(BINS)
 
--- a/test/Makefile
+++ b/test/Makefile
@@ -71,18 +71,18 @@ bin/iptb: iptb_src $(call find_go_files, $(IPTB_SRC)) IPFS-BUILD-OPTIONS
 test: test_expensive
 
 test_expensive: verify_gofmt
-	cd sharness && make TEST_EXPENSIVE=1
-	cd 3nodetest && make
-	cd dependencies && make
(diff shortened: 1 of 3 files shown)


test/Makefile
```diff
@@ -15,12 +15,13 @@ GOFLAGS =
...
 global-deps:
-	make -C .. deps
+	$(MAKE) -C .. deps
 
 deps: global-deps bins
 
 clean:
-	rm $(BINS)
+	$(MAKE) -C sharness clean
+	-rm -rf $(BINS)
```
lgierth: what's `-rm` doing?

lgierth:
LGTM apart from one question ^

test/Makefile
Kubuxu: `-` prefix makes `make` ignore return code of `rm` command.

lgierth: Okay then I think it's not needed -- `-f` makes it ignore non-existing files/dirs anyhow

thomas-gardner:
You can squash these:

```
build install nofuse: deps
        $(MAKE) -C cmd/ipfs $@

clean uninstall:
        $(MAKE) -C cmd/ipfs $@
```

And if we could drop "deps" as a prerequisite here (it is serious overkill), all 5 options could be squashed into two lines.

Kubuxu:
@thomas-gardner that is awesome trick, Thanks.

test/Makefile
Kubuxu: Yup, I should clear it up everywhere.

whyrusleeping:
@Kubuxu wanna go ahead and rebase this on master?

Kubuxu:
done

PR #3006, opened 2016-07-28
Edit issue reporting guides

RichardLitt:
I think this language makes it a bit more clear how we use issues in this repository.

License: MIT

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,18 +1,23 @@
-Version/Platform/Processor information (from `ipfs version --all`):
+<!-- Output From `ipfs version --all`) -->
+#### Version information:
+
+<!-- Bug, Feature, Enhancement, Etc -->
+#### Type:
+
+<!-- from `P0` "Critical" to `P5` "Relatively Unimportant") -->
+#### Priority:
+
+#### Description:
 
 
 
-Type (bug, feature, meta, test failure, question):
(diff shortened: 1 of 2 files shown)


lgierth:
Good call! Do you think we should also change the issue category examples? Maybe remove `meta` and `question`, and add `performance`, `crash`, `proposal`?

RichardLitt:
I think that would be a good idea! `question` is vague and I don't think it is useful. @em-ly @whyrusleeping thoughts?

RichardLitt:
Yes! I was thinking yesterday circling back to this. 

Do you have any specific pain points you want to bring up?

whyrusleeping:
We should put most of the 'help' text in comments, so it doesnt wind up in the actual issue.  And i think we should also reverse the "priority" numbers. They are the opposite of what is standard. 

Let me see if i can find that other issue where i made a few suggestions...

jbenet:
Yes +1 to reverse prios.

whyrusleeping:
@RichardLitt the discussion was here: https://github.com/ipfs/go-ipfs/pull/2786#issuecomment-237415377

Mostly just want to use comments to hide all the 'help text' from the actual issue.

RichardLitt:
SGTM. Will make a PR.

RichardLitt:
Testing. 


- Version information:


- Type:


- Priority :

### Description:

RichardLitt:
@whyrusleeping Take a look at the edits?

PR #3007, opened 2016-07-28
commands: fix refs 'edges' option work

### LINKED ISSUE #3001: ipfs refs ignores options ( --format, --edges, --unique)
ipfs version 0.4.2

Type: bug
Area: commands
Priority: P0

Description: The `ipfs refs` command appears to ignore the documented options `--format`, and `--edges` (and maybe `--unique`, I'm less sure how to test that).

Kubuxu:
also change it to use format instead of separate variable

Resolves #3001 

License: MIT

### DIFF
--- a/core/commands/refs.go
+++ b/core/commands/refs.go
@@ -74,16 +74,25 @@ NOTE: List all references recursively by using the flag '-r'.
 			return
 		}
 
-		edges, _, err := req.Option("edges").Bool()
+		format, _, err := req.Option("format").String()
 		if err != nil {
 			res.SetError(err, cmds.ErrNormal)
 			return
 		}
 
-		format, _, err := req.Option("format").String()
+		edges, _, err := req.Option("edges").Bool()
 		if err != nil {
 			res.SetError(err, cmds.ErrNormal)
 			return
+		}
+		if edges {
+			if format != "<dst>" {
+				res.SetError(errors.New("using format arguement with edges is not allowed"),
+					cmds.ErrClient)
+				return
+			}
+
+			format = "<src> -> <dst>"
 		}
 
 		objs, err := objectsForPaths(ctx, n, req.Arguments())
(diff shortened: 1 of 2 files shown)


core/commands/refs.go
```diff
@@ -74,17 +74,26 @@ NOTE: List all references recursively by using the flag '-r'.
...
 		}
 
-		format, _, err := req.Option("format").String()
+		edges, _, err := req.Option("edges").Bool()
 		if err != nil {
 			res.SetError(err, cmds.ErrNormal)
 			return
 		}
+		if edges {
+			if format != "<dst>" {
```
JesseWeinstein: Won't `format` be an empty string (rather than `"<dst>"`) if it isn't provided?

JesseWeinstein:
This has no tests for `--format`. While it *looks* like it should work, it would be much better if that was actually tested. I'll see about making a later PR to add them, unless you feel like doing so in this PR.

JesseWeinstein:
Interesting -- in 0.4.2, `--edges` *does* work, if you also explicitly set `--format` to an empty string, like so: `ipfs refs --format '' -e QmYwAPJzv5CZsnA625s3Xf2nemtYgPpHdWEz79ojWnPbdG`

core/commands/refs.go
```diff
@@ -315,8 +322,6 @@ func (rw *RefWriter) WriteEdge(from, to key.Key, linkname string) error {
 		s = strings.Replace(s, "<src>", from.B58String(), -1)
 		s = strings.Replace(s, "<dst>", to.B58String(), -1)
 		s = strings.Replace(s, "<linkname>", linkname, -1)
-	case rw.PrintEdge:
-		s = from.B58String() + " -> " + to.B58String()
 	default:
 		s += to.B58String()
```
JesseWeinstein: Do we still need this default case? It would only trigger if `--format` is explicitly set to the empty string, which seems like a corner case better handled by either an error, or actually printing an empty string for each edge, rather than overriding the explicit intention of the user.

whyrusleeping:
LGTM

core/commands/refs.go
jbenet: no because: `.Default("<dst>")`

PR #3008, opened 2016-07-28
supporting symlinks in web-server

cyisfor:
Just an idea for resolving symlinks in the web-server, instead of just flopping. Only works with links within the inserted directory, of course.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,8 +16,8 @@ else
 endif
 
 dist_root=/ipfs/QmUnvqDuRyfe7HJuiMMHv77AMUFnjGyAU28LFPeTYwGmFF
-gx_bin=bin/gx-v0.8.0
-gx-go_bin=bin/gx-go-v1.2.1
+gx_bin=$(GOPATH)/bin/gx
+gx-go_bin=$(GOPATH)/bin/gx-go
 
 # use things in our bin before any other system binaries
 export PATH := bin:$(PATH)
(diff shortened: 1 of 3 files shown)


lgierth:
Hey, thanks! The diff is a bit hard to review though, because you're mixing the symlink change with structural refactorings, and changes to the build process. Could you reduce it to the symlink change, just so we can get a clearer idea of what your proposed change is?

Kubuxu:
Also downloading prebuilt gx versions was a conscious choice as it allows us to version lock and upgrade when needed.

whyrusleeping:
Please only include a single focused changeset in each pull request.

cyisfor:
@lgierth the structural refactoring was necessary, because what was there assumed that there would be no special case for a symlink. and the build process thing is just using the source code, rather than an opaque binary. Doesn't change the build process. Anyway it's a separate commit, in case you didn't want that.

@Kubuxu Oh, I thought go get allowed you to get certain versions, but apparently it doesn't. What I'd do if you want a version lock is make a git submodule for gx then, and use that, rather than assuming some binary is the one you want. Don't really see why a version lock would be needed, as gx should have a fairly stable interface, but whatever you like.

thomas-gardner:
Author: Jakub Sztandera
Date: 2016-07-28 12:38:46-0700
> Also downloading prebuilt gx versions was a conscious choice as it
> allows us to version lock and upgrade when needed.

Does anyone actually do this? I know I don't use the Makefile anymore
because it is so bloated :(

jbenet:
It's about producing secure, stable software, without trusting
intermediaries, or trusting the future not to screw up. It's kind of what
ipfs itself is all about.

I definitely use the makefile.

whyrusleeping:
I'm closing this, If you want to continue this PR please reopen a new one with a single changeset only touching the code youre modifying.

PR #3009, opened 2016-07-29
fix: on windows, add path use absolute, more add object for father path

liliuhai:
old code run on windows, use absolute path, the ipfs added more object like father path

    License: MIT
    Signed-off-by: liliuhai <liliuhai@peersafe.cn>

### DIFF
--- a/commands/cli/parse.go
+++ b/commands/cli/parse.go
@@ -9,10 +9,11 @@ import (
 	"sort"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


liliuhai:
old code run on windows, use absolute path, the ipfs added more object like father path

License: MIT
Signed-off-by: liliuhai <liliuhai@peersafe.cn>

PR #1933, opened 2015-11-02
Fix path parser in add command

djdv:
Parse paths with filepath instead of just path
https://github.com/ipfs/go-ipfs/issues/1922

PR #3010, opened 2016-07-29
fix: on windows, add path use absolute, more add object for father path

liliuhai:
old code run on winodw,use add -r commands,if use absolute path as param,the result have more added object like father path,but use relativity path,no this problem.

License: MIT
Signed-off-by: liliuhai <liliuhai@peersafe.cn>

### DIFF
--- a/commands/cli/parse.go
+++ b/commands/cli/parse.go
@@ -9,10 +9,11 @@ import (
 	"sort"
 	"strings"
 
-	cmds "github.com/ipfs/go-ipfs/commands"
-	files "github.com/ipfs/go-ipfs/commands/files"
 	logging "gx/ipfs/QmNQynaz7qfriSUJkiEZUrm2Wen1u3Kj9goZzWtrPyu7XR/go-log"
 	u "gx/ipfs/QmZNVWh8LLjAavuQ2JXuFmuYH3C11xo988vSgp7UQrTRj1/go-ipfs-util"
+
+	cmds "github.com/ipfs/go-ipfs/commands"
+	files "github.com/ipfs/go-ipfs/commands/files"
 )
 
 var log = logging.Logger("commands/cli")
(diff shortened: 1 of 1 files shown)


djdv:
I believe it would be more appropriate to change line 401 `fpath, err := filepath.EvalSymlinks(fpath)` to `fpath, err := filepath.ToSlash(filepath.EvalSymlinks(fpath))` instead, fpath is supposed to be standardized to forward slashes to prevent this kind of problem, it seems https://github.com/peersafe/go-ipfs/commit/fe7b01f14e2b8db6e856b567dbd7db4d00f6e366 may have re-introduced this problem by reassigning fpath from standard forward slashes into platform specific paths.

For reference on standard slahs paths https://github.com/ipfs/go-ipfs/pull/1933#issuecomment-155304003

ping @Kubuxu

liliuhai:
I think filepath.EvalSymlinks function cause problem，in winodws，before call EvalSymlinks，fpath is use “\”，but after called fpath changed for “/"，so re-introduced this problem.

djdv:
Sorry, I neglected the multiple return values of EvalSymlinks, this is what I should have suggested: https://github.com/djdv/go-ipfs/commit/31b4cbb51fcd04747ef3331249d0d13a95d48565

liliuhai:
thanks, like this the problem can fixed

djdv:
Sounds good, we'll have to wait for an IPFS member to make judgment since I don't know which method is preferable or if anything else has to be done. I'm glad this regression was spotted, thank you for finding it!

whyrusleeping:
@djdv @liliuhai Would this issue have been caught by our existing sharness tests?

commands/cli/parse.go
```diff
@@ -416,7 +417,7 @@ func appendFile(fpath string, argDef *cmds.Argument, recursive, hidden bool) (fi
 		}
 	}
 
-	return files.NewSerialFile(path.Base(fpath), fpath, hidden, stat)
+	return files.NewSerialFile(path.Base(filepath.ToSlash(fpath)), fpath, hidden, stat)
```
whyrusleeping: instead of doing this here, we should on line 401 do:
```go
fpath = filepath.Clean(fpath)
fpath, err = filepath.EvalSymlinks(fpath)
...
fpath = filepath.ToSlash(fpath)
```

Kubuxu:
I've taken over this PR in #3023.

PR #3018, opened 2016-07-31
Put all Taglines & most Option descriptions in active, present tense

JesseWeinstein:
For consistency. This should probably be documented somewhere, too.

### DIFF
--- a/core/commands/active.go
+++ b/core/commands/active.go
@@ -13,7 +13,7 @@ import (
 
 var ActiveReqsCmd = &cmds.Command{
 	Helptext: cmds.HelpText{
-		Tagline: "List commands run on this ipfs node.",
+		Tagline: "Lists commands run on this ipfs node.",
 		ShortDescription: `
 Lists running and recently run commands.
 `,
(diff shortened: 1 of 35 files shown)


thomas-gardner:
Hi, I think the imperative mood is the default for this subset of
technical writing. Some non-IPFS examples include, 

	`git`, `go`, `gcc --help`.

JesseWeinstein:
Sigh. I was afraid that might be the case. Well, if we settle on imperative, I'll make them all consistent with *THAT*, instead.

RichardLitt:
This should be documented here: https://github.com/ipfs/community/blob/master/docs-styleguide.md. Might be good to open an issue about it in ipfs/community before doing all of the changes again, in case we want to weigh in on it not only for go-ipfs but also for other repos.

whyrusleeping:
blocked on resolution of https://github.com/ipfs/community/issues/150

jbenet:
Please keep using imperative. It's a decades old convention for unix
tooling.

And I think it's much better, too. Think of it like a button. Buttons in an
airplane don't say "[this button] turns the engine on". They say "Turn the
engine on". These descriptions are putting in writing what "the user will
do" by executing that command (pushing the button) instead of doing it
manually.

JesseWeinstein:
OK, I've closed this PR, and will make another (much smaller) one to fix the few taglines & options descriptions that weren't in imperative mood.

Issue #3019, opened 2016-07-31
bug: Automatic migration on Win64 failing


NightRa:
go-ipfs version: 0.4.3-dev- (0.4.3-rc1 binary)
Repo version: 4 (current is 3)
System version: amd64/windows

Type: bug
Area: automatic migration
Priority: P2

Description:
The automatic migration in 0.4.3-rc1 fails with the following:
```
Initializing daemon...
Found old repo version, migrations need to be run.
Run migrations automatically? [y/N] y
  => checking for migrations binary...
  => usable migrations not found on system, fetching...
fetching url: https://ipfs.io/ipfs/QmUnvqDuRyfe7HJuiMMHv77AMUFnjGyAU28LFPeTYwGmFF/fs-repo-migrations/versions
fetching url: https://ipfs.io/ipfs/QmUnvqDuRyfe7HJuiMMHv77AMUFnjGyAU28LFPeTYwGmFF/fs-repo-migrations/v1.1.0/fs-repo-migrations_v1.1.0_windows-amd64.zip
Error: no migration binary found that supports version 4 - failed to check migrations version: exec: "C:\\Users\\Ilan\\AppData\\Local\\Temp\\go-ipfs-migrate167521031\\fs-repo-migrations": file does not exist
```
The mentioned temporary folder contains only the zip file, unextracted.

lgierth:
The auto-migration relies on dist.ipfs.io having fs-repo-migrations binaries for the platform. Would you be interested in adding windows-amd64 builds? I filed an issue for it: ipfs/distributions#72.

In the meantime, you can simply try `go get -u github.com/ipfs/fs-repo-migrations` and run `ipfs daemon` again. If the daemon doesn't pick that up, you can run `fs-repo-migrations` manually.

lgierth:
Oooh nevermind what I'm saying about adding support for windows-amd64 -- it's supposed to already be there...

lgierth:
The auto-migrations code needs to append `.exe` on windows.

NightRa:
@lgierth From the source & the results I see all the things you mention done:
1. It did download "fs-repo-migrations_v1.1.0_windows-amd64.zip" into the temporary folder.
2. unpack.unpackZip does try to extract a file ending in .exe
But apparently it doesn't?

NightRa:
Maybe it doesn't hit this conditional in unpack.go/unpackZip:
```
if fis.Name == dist+"/"+binnom+".exe"
```

[...]

PR #3021, opened 2016-08-01
extract binaries with .exe name on windows

whyrusleeping:
Should address #3019 
We really need to work on some sort of CI for windows.

License: MIT


whyrusleeping:
Verified this works on an AWS box

Kubuxu:
If it runs LGTM.

jbenet:
I think this should have a test case

jbenet:
we want to test all this install/migrations stuff. it doesnt have to be tested on every commit to go-ipfs, so maybe it could all be in another package, like `go-ipfs-migrations` or something

PR #1982, opened 2015-11-20
Add appveyor conf

rht:
- [ ] add test_go_expensive
- [x] add appveyor pr webhook http://help.appveyor.com/discussions/questions/203-auto-run-tests-on-pull-requests @ipfs/go-ipfs-team (can only be done by someone with admin access -- not me)

Issue #1922, opened 2015-10-31
Windows paths with ipfs add


djdv:
On Windows, when recursively adding directories with `ipfs add -r *target*`, ipfs will produce strange trees. The target directory will be walked and added but there is some redundant adding and malformed names being produced, files will have the path they are located at prepended to their filename and be listed in 2 places, the root will contain all files and directories no matter how deep in addition to having files listed (with malformed names) where they should be located, for example a file located at "C:\dir\subdir\file.ext" would live here after adding "C:\dir\subdir\/C:\dir\subdir\file.ext"

These visual examples should better exemplify what I mean:
![listings](https://cloud.githubusercontent.com/assets/13862850/10864379/b37d74ea-7fc2-11e5-946c-8285b4169f31.png)
![listing 1](https://cloud.githubusercontent.com/assets/13862850/10864381/b37f54e0-7fc2-11e5-80f9-74c04c5f13d6.PNG)
![listing 1b](https://cloud.githubusercontent.com/assets/13862850/10864378/b37bc064-7fc2-11e5-9691-cfebb40ec6fc.PNG)
![listing 2](https://cloud.githubusercontent.com/assets/13862850/10864380/b37d6d1a-7fc2-11e5-9766-b242d2af48ce.PNG)

These bad directories can be viewed at these hashes:
QmYtNSri8H4wstyaEwHe9gWRM1cSADCNNJBNcw3mtQuB2T
QmfPYceFpW2JNJcN8NpyncM7vT8tkasnd7pG5A6xQs7RUd

I'm assuming the problem is related to the path delimiters on Windows being '\' and pkg/path or pkg/path/filepath not being used somewhere it needs to be.
There seems to be other instances where this causes issues such as when ipfs add tries to create ~/.ipfs/block if it does not exist, it will return this error:
`Error: mkdir C:\Users\Dominic\.ipfs/blocks/1220bfcc: The system cannot find the path specified.`

-Edit-
For reference:
OS: Windows 10 x64
Go: 1.5.1
IPFS: a93dbf6

wscott:
(from outside observer...) Windows is perfectly happy to use forward
slashes if you hand it a pathname like that. So pathnames should be
normalized to use forward slashes internally for consistency.

jbenet:
this is a golang thing-- we need to use the `filepath` package instead of `path` in more places, and convert the paths on the way out.

[...]

PR #3023, opened 2016-08-01
cli: do ToSlash after EvalSymlinks to remove platform specific slashes

Kubuxu:
Resolves #3010

License: MIT

### DIFF
--- a/commands/cli/parse.go
+++ b/commands/cli/parse.go
@@ -397,11 +397,14 @@ func appendFile(fpath string, argDef *cmds.Argument, recursive, hidden bool) (fi
 		fpath = cwd
 	}
 
-	fpath = filepath.ToSlash(filepath.Clean(fpath))
+	fpath = filepath.Clean(fpath)
 	fpath, err := filepath.EvalSymlinks(fpath)
 	if err != nil {
 		return nil, err
 	}
+	// Repeat ToSlash after EvalSymlinks as it turns path to platform specific
+	fpath = filepath.ToSlash(fpath)
+
 	stat, err := os.Lstat(fpath)
 	if err != nil {
 		return nil, err


commands/cli/parse.go
```diff
@@ -397,11 +397,15 @@ func appendFile(fpath string, argDef *cmds.Argument, recursive, hidden bool) (fi
...
 	}
 
-	fpath = filepath.ToSlash(filepath.Clean(fpath))
+	fpath = filepath.Clean(fpath)
+	fpath = filepath.ToSlash(fpath)
 	fpath, err := filepath.EvalSymlinks(fpath)
 	if err != nil {
 		return nil, err
 	}
+	// Repeat ToSlash after EvalSymlinks as it truns path to platform specific
```
whyrusleeping: typo here `truns`

commands/cli/parse.go
```diff
@@ -397,11 +397,15 @@ func appendFile(fpath string, argDef *cmds.Argument, recursive, hidden bool) (fi
 		fpath = cwd
 	}
 
-	fpath = filepath.ToSlash(filepath.Clean(fpath))
+	fpath = filepath.Clean(fpath)
+	fpath = filepath.ToSlash(fpath)
```
whyrusleeping: no need to run this `ToSlash` here

whyrusleeping:
@Kubuxu what is a repro for the original problem? Since this is a windows issue and we don't have windows CI, i will manually verify this fixes the problem on a windows server box

Kubuxu:
I don't really know.
@liliuhai @djdv

whyrusleeping:
@Kubuxu why does this contain changelog updates?

Kubuxu:
because I screwed up my local master sorry
fixing it

djdv:
This is a subset of the problem outlined in issue https://github.com/ipfs/go-ipfs/issues/1922 and pr https://github.com/ipfs/go-ipfs/pull/1933

Absolute paths are added but mangled fake directories are also added to the root directory of the ipfs object, if I remember correctly it always returns a directory object as well even when -r and -w are not supplied.
The images in the issue may better describe the problem.

Sans the comment this patch mirrors my own (https://github.com/djdv/go-ipfs/commit/31b4cbb51fcd04747ef3331249d0d13a95d48565) which works on my machine to fix this issue, I believe @liliuhai is saying the same in the previous pr as well (https://github.com/ipfs/go-ipfs/pull/3010#issuecomment-236470925).

Adding any Windows file via its absolute path will result in these malformed directory objects. i.e. `ipfs add "C:\somewhere\file.ext" Would result in a structure like this:
```
/*hash*
    C:\somewhere\file.ext 6B
    file.ext *actual size*
```
Assuring that paths are / standardized prevents this from happening.

djdv:
Following up to https://github.com/ipfs/go-ipfs/pull/3010#issuecomment-236563531 here

@whyrusleeping 
I'm not actually familiar with the IPFS tests, if there is a case for adding file(s) via their absolute path then this should have been caught on Windows, but I don't think any of the CI's are testing on Windows anymore, since this is a platform specific problem I don't think it would be caught unless running on a platform that uses path delimiters that are not `/`. The pr that reintroduced it (https://github.com/ipfs/go-ipfs/pull/2897) passed 3/5 checks and I think they were all run on *nix containers.

whyrusleeping:
Tested this branch on my windows box. adding a dir using an absolute path creates the same object as using a relative path now. This LGTM

jbenet:
This should have test cases-- we will have a regression otherwise. 

@djdv what is a good way to setup test cases in windows? 
- we had a CI thing. not sure if it still is alive.
- we could also have some os-specific tests to run with sharness.

djdv:
@jbenet 
tl;dr:
Current sharness tests seem fine but CI's can't use them on Windows. We should probably open an issue for getting those tests working properly on Windows and then make CI's use them. `go test "./..."` works fine on Windows currently though.

Long form:
I think this would have been caught with the current test cases but they would have to have been run on Windows or any OS that uses different standards for file paths. `filepath` returning platform normalized path strings is what caused this, if the tests are run on a platform that already uses paths with the same standards IPFS expects then there will be no issue if we're using `path` or `filepath` on those platforms. 

Testing the resulting hash of an add via an absolute path such as `C:\tmp\file` would have caught this previously, that case would fail where `/tmp/file` would not, however that case would only show up in a Windows environment so I think it must be tested there. A mock filesystem could possibly be used instead for testing paths but I think that sidesteps testing for other platform issues that could be present.

At one point go-ipfs was building and testing with Appveyor on Windows but I don't think that's the case anymore. Relevant: https://github.com/ipfs/go-ipfs/pull/1982
The CI's that can, should be building and testing on Windows, that would help prevent regressions like this.

Testing with go via `go test "./..."` works fine natively but the sharness tests will not work without reworking them a bit to take care of path quirks and glob patterns like "*.go"
screenshot

Maybe an issue should be raised on this and we can make an effort towards making those scripts more Windows aware.

whyrusleeping:
@djdv we disabled the appveyor tests because they were always failing and none of the devs had windows boxes to test any of this with. I have an AWS windows server box right now but thats still far from ideal for developing on. If you're interested, we would love some help getting things running nicely on windows (CI and all). We're also looking into a better CI solution over at https://github.com/ipfs/infrastructure/issues/100

djdv:
@whyrusleeping 
Absolutely, I'll come to the IRC channel when I get some time here. See if I can help out in any way.

PR #3025, opened 2016-08-01
default add progress to false, unless using CLI

whyrusleeping:
Should fix issues seen in https://github.com/ipfs/js-ipfs-api/issues/323

License: MIT


core/commands/add.go
```diff
@@ -81,6 +81,11 @@ You can now refer to the added file in a gateway, like so:
 			return nil
 		}
 
+		_, found, _ := req.Option(progressOptionName).Bool()
```
Kubuxu: That seems quite cheaty.
Maybe we should just introduce something like `request agent`? to the `Request` object.

whyrusleeping: That might be a good idea, except it changes the http api. which i'm explicitly trying to avoid right now.

Kubuxu: I have missed that the code is in PreRun function.

Everything is much less hacky now.

Kubuxu:
LGTM

PR #3026, opened 2016-08-02
blockstore: extract ARC cache from Bloom cache

Kubuxu:
it removes race condition that would happen during various calls

If tests are ok, it is good for review.

License: MIT

### DIFF
--- a/blocks/blockstore/arc_cache.go
+++ b/blocks/blockstore/arc_cache.go
@@ -30,30 +30,28 @@ func (b *arccache) DeleteBlock(k key.Key) error {
 	b.arc.Remove(k) // Invalidate cache before deleting.
 	err := b.blockstore.DeleteBlock(k)
 	switch err {
-	case nil:
+	case nil, ds.ErrNotFound, ErrNotFound:
 		b.arc.Add(k, false)
-	case ds.ErrNotFound, ErrNotFound:
-		b.arc.Add(k, false)
+		return nil
 	default:
 		return err
 	}
-	return nil
 }
 
 // if ok == false has is inconclusive
 // if ok == true then has respons to question: is it contained
 func (b *arccache) hasCached(k key.Key) (has bool, ok bool) {
 	if k == "" {
-		// Return cache invalid so call to blockstore
-		// in case of invalid key is forwarded deeper
+		// Return cache invalid so the call to blockstore happens
+		// in case of invalid key and correct error is created.
 		return false, false
 	}
+
 	h, ok := b.arc.Get(k)
 	if ok {
-		return h.(bool), ok
-	} else {
-		return false, false
+		return h.(bool), true
 	}
+	return false, false
 }
 
 func (b *arccache) Has(k key.Key) (bool, error) {
--- a/blocks/blockstore/arc_cache.go
+++ b/blocks/blockstore/arc_cache.go
@@ -1,0 +1,127 @@
+package blockstore
+
+import (
+	"github.com/ipfs/go-ipfs/blocks"
+	key "github.com/ipfs/go-ipfs/blocks/key"
+	ds "gx/ipfs/QmTxLSvdhwg68WJimdS6icLPhZi28aTp6b7uihC2Yb47Xk/go-datastore"
+	lru "gx/ipfs/QmVYxfoJQiZijTgPNHCHgHELvQpbsJNTg6Crmc3dQkj3yy/golang-lru"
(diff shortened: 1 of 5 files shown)


blocks/blockstore/arc_cache.go
```diff
@@ -0,0 +1,127 @@
...
+
+func (b *arccache) PutMany(bs []blocks.Block) error {
+	var good []blocks.Block
+	for _, block := range bs {
+		if has, ok := b.hasCached(block.Key()); !ok || (ok && !has) {
+			good = append(good, block)
+		}
+	}
+	err := b.blockstore.PutMany(bs)
+	if err == nil {
```
whyrusleeping: i'd reverse the logic here to be more conventional:

```go
if err != nil {
  return err
}

for ....
```

Kubuxu: Sure, it was refactored out so there was more logic in here.

whyrusleeping:
@Kubuxu theres a lot of failing tests here...

blocks/blockstore/arc_cache_test.go
```diff
@@ -0,0 +1,67 @@
...
+	b := blocks.NewBlock([]byte("foo"))
+	cd := &callbackDatastore{f: func() {}, ds: ds.NewMapDatastore()}
+	bs := NewBlockstore(syncds.MutexWrap(cd))
+	cachedbs, err := testArcCached(bs, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+	cachedbs.Put(b)
+
+	cd.Lock()
```
whyrusleeping: whats with locking here? you dont need to lock around a variable instantiation

whyrusleeping:
what was the actual race condition here that youre fixing?

Kubuxu:
The race was happening when tests for ARC was running and the bloom cache finished initializing.

The test failures you see here are because the gateways were down.

blocks/blockstore/arc_cache.go
```diff
@@ -0,0 +1,128 @@
...
+func (b *arccache) hasCached(k key.Key) (has bool, ok bool) {
+	if k == "" {
+		// Return cache invalid so call to blockstore
+		// in case of invalid key is forwarded deeper
+		return false, false
+	}
+	h, ok := b.arc.Get(k)
+	if ok {
+		return h.(bool), ok
+	} else {
```
whyrusleeping: drop the else

blocks/blockstore/arc_cache.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+}
+
+func (b *arccache) Has(k key.Key) (bool, error) {
+	if has, ok := b.hasCached(k); ok {
+		return has, nil
+	}
+
+	res, err := b.blockstore.Has(k)
+	if err == nil {
```
whyrusleeping: switch logic to early return on non-nil error

blocks/blockstore/arc_cache.go
Kubuxu: yeah, it is artefact from the refactor too

whyrusleeping:
Alright, LGTM. waiting on tests to pass now

Kubuxu:
All green.

Kubuxu:
Also the race in question: https://gist.github.com/Kubuxu/575f39058618a3fd5fd3efa518c65c1e

blocks/blockstore/arc_cache.go
```diff
@@ -0,0 +1,126 @@
...
+	lru "gx/ipfs/QmVYxfoJQiZijTgPNHCHgHELvQpbsJNTg6Crmc3dQkj3yy/golang-lru"
+	context "gx/ipfs/QmZy2y8t9zQH2a1b8q2ZSLKp17ATuJoCNxxyMFG5qFExpt/go-net/context"
+)
+
+type arccache struct {
+	arc        *lru.ARCCache
+	blockstore Blockstore
+}
+
+func arcCached(bs Blockstore, lruSize int) (*arccache, error) {
```
jbenet: - `newArcCacheDS`

blocks/blockstore/arc_cache_test.go
jbenet: yeah why is this lock here?

Kubuxu: The race detector was giving me an error otherwise for some reason, I can tinker with it more to find out the exact reason.

Maybe line 47 should be locked instead.

PR #3033, opened 2016-08-03
commands/cli: fix reading from stdin message

### LINKED ISSUE #3002: Many commands are erroneously printing "Reading from stdin" info message
```
commit: d742fb1...

$ ipfs version --all
go-ipfs version: 0.4.3-dev-
Repo version: 4
System version: amd64/linux
Golang version: go1.6.3
```
Type (bug, feature, meta, test failure, question): `bug`
Area (api, commands, daemon, fuse, etc): `commands/files|cli`
Priority (from P0: functioning, to P4: operations on fire): `P0`

Description: As in the title,
```
go-ipfs|master $ ipfs ls --help 2>&1 |head -n1
ipfs: Reading from /dev/stdin; send Ctrl-d to stop.
```
This is not the intended behaviour.

I think it should be possible to move this message into ReaderFile.Read() so that it isn't printed too early---an example of early printing is  `ipfs dht put foo`.

whyrusleeping:
Only print "Reading from /dev/stdin" message when we actually read from
stdin (and not in other cases such as ipfs add --help).

closes #3002 

License: MIT

### DIFF
--- a/test/sharness/t0040-add-and-cat.sh
+++ b/test/sharness/t0040-add-and-cat.sh
@@ -9,6 +9,15 @@ test_description="Test add and cat commands"
 . lib/test-lib.sh
 
 test_add_cat_file() {
+	test_expect_success "ipfs add --help works" '
+		ipfs add --help 2> add_help_err > /dev/null
+	'
+
+	test_expect_success "stdin reading message doesnt show up" '
+		test_expect_code 1 grep "ipfs: Reading from" add_help_err &&
+		test_expect_code 1 grep "send Ctrl-d to stop." add_help_err
+	'
+
     test_expect_success "ipfs add succeeds" '
     	echo "Hello Worlds!" >mountdir/hello.txt &&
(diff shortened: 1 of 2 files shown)


test/sharness/t0040-add-and-cat.sh
```diff
@@ -9,6 +9,15 @@ test_description="Test add and cat commands"
...
 test_add_cat_file() {
+	test_expect_success "ipfs add --help works" '
+		ipfs add --help > add_help_out
+	'
+
+	test_expect_success "stdin reading message doesnt show up" '
+		test_expect_code 1 grep "ipfs: Reading from" add_help_out &&
+		test_expect_code 1 grep "send Ctrl-d to stop." add_help_out
+	'
+
```
Kubuxu: This test would be successful even before the patch, you have to capture both stdout and stderr.

Kubuxu:
LGTM, the mistake in tests got fixed.

PR #3036, opened 2016-08-04
bitswap: fix a minor data race

whyrusleeping:
race detector picked up a minor race condition, Since loop iteration reuses the same local variable, its not safe to take its address and use it concurrently. The fix is to rebind the variable into a controlled scope (creating a new variable) and taking the address of that to pass outwards.

License: MIT

### DIFF
--- a/exchange/bitswap/workers.go
+++ b/exchange/bitswap/workers.go
@@ -173,6 +173,7 @@ func (bs *Bitswap) rebroadcastWorker(parent context.Context) {
 		case <-broadcastSignal.C: // resend unfulfilled wantlist keys
 			log.Event(ctx, "Bitswap.Rebroadcast.active")
 			for _, e := range bs.wm.wl.Entries() {
+				e := e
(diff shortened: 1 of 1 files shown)


Kubuxu:
LGTH, but it would be great to have this explanation either as commit message or a comment.

whyrusleeping:
@Kubuxu added more description to the commit message

Kubuxu:
Great, thank you.  LGTM

exchange/bitswap/workers.go
```diff
@@ -173,6 +173,7 @@ func (bs *Bitswap) rebroadcastWorker(parent context.Context) {
 		case <-broadcastSignal.C: // resend unfulfilled wantlist keys
 			log.Event(ctx, "Bitswap.Rebroadcast.active")
 			for _, e := range bs.wm.wl.Entries() {
+				e := e
```
jbenet: add:
```go
e := e // prevent data race
```

Issue #3032, opened 2016-08-03
Segmentation fault `ipfs daemon --mount`


karlek:
Version/Platform/Processor information (from `ipfs version --all`):
```
go-ipfs version: 0.4.3-dev-1c41cac
Repo version: 4
System version: amd64/linux
Golang version: devel +495e3c6 Tue May 17 04:02:11 2016 +0000
```

Type (bug, feature, meta, test failure, question): `bug`
Area (api, commands, daemon, fuse, etc): `daemon`
Priority (from P0: functioning, to P4: operations on fire): `P2`

Description:
I ran a `ipfs get` over night on a 13~GiB large hash, and when I woke up it hade crashed with a segmentation fault after 20%. 

The `ipfs get` produced this output: `Error: unexpected EOF`

This was the output from the `ipfs daemon`:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x0]

goroutine 10460213 [running]:
panic(0xb5ce40, 0xc4200140d0)
	/home/_/go/src/runtime/panic.go:500 +0x18c
github.com/ipfs/go-ipfs/routing/dht.(*IpfsDHT).findProvidersAsyncRoutine(0xc4201d66e0, 0x104e7e0, 0xc42968e340, 0xc43b082030, 0x22, 0x3, 0xc43b31cf60)
	/home/_/Desktop/privgo/src/github.com/ipfs/go-ipfs/routing/dht/routing.go:385 +0x5bc
created by github.com/ipfs/go-ipfs/routing/dht.(*IpfsDHT).FindProvidersAsync
	/home/_/Desktop/privgo/src/github.com/ipfs/go-ipfs/routing/dht/routing.go:306 +0x16e
```

whyrusleeping:
@karlek thanks for reporting! After this happened, was the daemon still running? or did the process exit?

karlek:
The deamon didn't continue running, it crashed with the panic.

whyrusleeping:
No worries, this one is tricky. Thanks for the info!

whyrusleeping:
@karlek what version of go are you running? It doesnt appear to be a release version.

whyrusleeping:
My first instinct is that the error is a 'hidden nil' where its a typed nil pointer returned through an interface.  This would make it past the `err != nil` check on routing.go:380 and panic on line 385, but the panic trace should look different...

[...]

PR #3037, opened 2016-08-04
dht: add in code to detect and diagnose #3032

whyrusleeping:
Debugging code to detect issue #3032 and prevent it from crashing the daemon.

License: MIT

### DIFF
--- a/routing/dht/routing.go
+++ b/routing/dht/routing.go
@@ -3,6 +3,7 @@ package dht
 import (
 	"bytes"
 	"fmt"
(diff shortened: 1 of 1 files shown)


Kubuxu:
As a workaround, looks great.
Only problem I see with it is that we probably never will receive that data from users.

So LGTM because it can't be done better.

whyrusleeping:
I think we will eventually get someone to report this, and it should only take once to figure it out.

PR #3042, opened 2016-08-05
Implement unixfs sharding

whyrusleeping:
This PR implements a hash array mapped trie using merkledag nodes as a base layer for sharded unixfs directories. 

This changeset is broken up into three distinct parts:
1. Implement the HAMT structure itself. This is a bunch of algorithmic code that i'd like some well thought out sanity checking on.
   - Need confirmation on hash code numbering
   - Thoughts on willf/bitset vs stdlib `math/big.Int`
   - Implementation correctness
   - Consensus on addition of fields to unixfs protobuf
2. Integration of HAMT code into the directory builder logic.
   - provides us a nice abstraction to think about things in terms of a filesystem
   - optimizes smaller directories by not sharding until we hit a threshold
3. Integration of directory builder code into mfs
   - moves mfs from using a simple merkledag `Node` object to using a dirbuilder
   - allows mfs to create sharded directories
   - by way of the previous point, `ipfs add` and friends can now create sharded directories.

Thorough review will be very appreciated!

cc @Kubuxu @jbenet @kevina @diasdavid @dignifiedquire 

also cc @haadcode as you might like to review the datastructure logic

### DIFF
--- a/mfs/dir.go
+++ b/mfs/dir.go
@@ -217,12 +217,12 @@ type NodeListing struct {
 	Hash string
 }
 
-func (d *Directory) ListNames() ([]string, error) {
+func (d *Directory) ListNames(ctx context.Context) ([]string, error) {
 	d.lock.Lock()
 	defer d.lock.Unlock()
 
 	var out []string
-	err := d.dirbuilder.ForEachLink(func(l *node.Link) error {
+	err := d.dirbuilder.ForEachLink(ctx, func(l *node.Link) error {
 		out = append(out, l.Name)
 		return nil
 	})
@@ -235,9 +235,9 @@ func (d *Directory) ListNames() ([]string, error) {
 	return out, nil
 }
 
-func (d *Directory) List() ([]NodeListing, error) {
+func (d *Directory) List(ctx context.Context) ([]NodeListing, error) {
 	var out []NodeListing
-	err := d.ForEachEntry(context.TODO(), func(nl NodeListing) error {
+	err := d.ForEachEntry(ctx, func(nl NodeListing) error {
 		out = append(out, nl)
 		return nil
 	})
@@ -247,7 +247,7 @@ func (d *Directory) List() ([]NodeListing, error) {
 func (d *Directory) ForEachEntry(ctx context.Context, f func(NodeListing) error) error {
 	d.lock.Lock()
 	defer d.lock.Unlock()
-	return d.dirbuilder.ForEachLink(func(l *node.Link) error {
+	return d.dirbuilder.ForEachLink(ctx, func(l *node.Link) error {
 		c, err := d.childUnsync(l.Name)
 		if err != nil {
 			return err
--- a/mfs/dir.go
+++ b/mfs/dir.go
@@ -221,71 +221,59 @@ func (d *Directory) ListNames() ([]string, error) {
 	d.lock.Lock()
 	defer d.lock.Unlock()
 
-	names := make(map[string]struct{})
-	for n, _ := range d.childDirs {
-		names[n] = struct{}{}
-	}
-	for n, _ := range d.files {
-		names[n] = struct{}{}
-	}
-
-	links, err := d.dirbuilder.Links()
+	var out []string
+	err := d.dirbuilder.ForEachLink(func(l *node.Link) error {
+		out = append(out, l.Name)
+		return nil
+	})
 	if err != nil {
 		return nil, err
 	}
 
-	for _, l := range links {
-		names[l.Name] = struct{}{}
-	}
-
-	var out []string
-	for n, _ := range names {
-		out = append(out, n)
-	}
 	sort.Strings(out)
 
 	return out, nil
 }
 
 func (d *Directory) List() ([]NodeListing, error) {
+	var out []NodeListing
+	err := d.ForEachEntry(context.TODO(), func(nl NodeListing) error {
+		out = append(out, nl)
+		return nil
+	})
+	return out, err
+}
+
+func (d *Directory) ForEachEntry(ctx context.Context, f func(NodeListing) error) error {
 	d.lock.Lock()
 	defer d.lock.Unlock()
-
-	var out []NodeListing
-
-	links, err := d.dirbuilder.Links()
-	if err != nil {
-		return nil, err
-	}
-
-	for _, l := range links {
-		child := NodeListing{}
-		child.Name = l.Name
-
+	return d.dirbuilder.ForEachLink(func(l *node.Link) error {
 		c, err := d.childUnsync(l.Name)
 		if err != nil {
-			return nil, err
+			return err
 		}
 
-		child.Type = int(c.Type())
+		nd, err := c.GetNode()
+		if err != nil {
+			return err
+		}
+
+		child := NodeListing{
+			Name: l.Name,
+			Type: int(c.Type()),
+			Hash: nd.Cid().String(),
+		}
+
 		if c, ok := c.(*File); ok {
 			size, err := c.Size()
 			if err != nil {
-				return nil, err
+				return err
 			}
 			child.Size = size
 		}
-		nd, err := c.GetNode()
-		if err != nil {
-			return nil, err
-		}
 
-		child.Hash = nd.Cid().String()
-
-		out = append(out, child)
-	}
-
-	return out, nil
+		return f(child)
+	})
 }
 
 func (d *Directory) Mkdir(name string) (*Directory, error) {
@@ -433,5 +421,5 @@ func (d *Directory) GetNode() (node.Node, error) {
 		return nil, err
 	}
 
-	return nd, err
+	return nd.Copy(), err
 }
--- a/mfs/dir.go
+++ b/mfs/dir.go
@@ -59,7 +59,7 @@ func NewDirectory(ctx context.Context, name string, node node.Node, parent child
 
 // closeChild updates the child by the given name to the dag node 'nd'
 // and changes its own dag node
-func (d *Directory) closeChild(name string, nd *dag.ProtoNode, sync bool) error {
+func (d *Directory) closeChild(name string, nd node.Node, sync bool) error {
 	mynd, err := d.closeChildUpdate(name, nd, sync)
 	if err != nil {
 		return err
@@ -72,7 +72,7 @@ func (d *Directory) closeChild(name string, nd *dag.ProtoNode, sync bool) error 
 }
 
 // closeChildUpdate is the portion of closeChild that needs to be locked around
-func (d *Directory) closeChildUpdate(name string, nd *dag.ProtoNode, sync bool) (*dag.ProtoNode, error) {
+func (d *Directory) closeChildUpdate(name string, nd node.Node, sync bool) (*dag.ProtoNode, error) {
 	d.lock.Lock()
 	defer d.lock.Unlock()
 
@@ -329,13 +329,10 @@ func (d *Directory) Unlink(name string) error {
 }
 
 func (d *Directory) Flush() error {
-	d.lock.Lock()
-	nd, err := d.flushCurrentNode()
+	nd, err := d.GetNode()
 	if err != nil {
-		d.lock.Unlock()
 		return err
 	}
-	d.lock.Unlock()
 
 	return d.parent.closeChild(d.name, nd, true)
 }
--- a/mfs/dir.go
+++ b/mfs/dir.go
@@ -12,6 +12,7 @@ import (
 
 	dag "github.com/ipfs/go-ipfs/merkledag"
 	ft "github.com/ipfs/go-ipfs/unixfs"
+	uio "github.com/ipfs/go-ipfs/unixfs/io"
 	ufspb "github.com/ipfs/go-ipfs/unixfs/pb"
 
 	node "gx/ipfs/QmYDscK7dmdo2GZ9aumS8s5auUUAH5mR1jvj5pYhWusfK7/go-ipld-node"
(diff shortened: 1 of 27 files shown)


whyrusleeping:
Note: tests won't pass yet as I havent vendored some of the new libs. I havent quite decided between [willf/bitset](https://github.com/willf/bitset) and just using the `math/big.Int` to handle the bitfield logic.

whyrusleeping:
going with the stdlib, fewer imports and the stdlib is proven to be pretty stable. Its a hair bit less efficient on some operations, but we can optimize later if it becomes an issue.

kevina:
@whyrusleeping, I am afraid I do not know anything about sharding, so I will have to first learn about that before I am of any use.  Doing a quick search I discovered ipfs/notes#76, is there anything else that would help with content for this pull request?

unixfs/hamt/hamt.go
```diff
@@ -0,0 +1,393 @@
...
+	"gx/ipfs/QmZy2y8t9zQH2a1b8q2ZSLKp17ATuJoCNxxyMFG5qFExpt/go-net/context"
+	"gx/ipfs/QmfJHywXQu98UeZtGJBQrPAR6AtmDjjbe3qjTo9piXHPnx/murmur3"
+
+	dag "github.com/ipfs/go-ipfs/merkledag"
+	format "github.com/ipfs/go-ipfs/unixfs"
+	upb "github.com/ipfs/go-ipfs/unixfs/pb"
+)
+
+const (
+	HashMurmur3 uint64 = iota
```
jbenet: remember that this is configurable on the object

jbenet: it will likely use multihash if this proposal goes through: https://github.com/multiformats/multihash/issues/30

jbenet:
@whyrusleeping this is a very big PR that will be critical to review + merge in stages. if this has any bugs it will be serious trouble and will take weeks of work to find and debug. please:

- rip out unixfs into its own repo first, then we can iterate on the datastructure there first. once we're convinced it really works, we can pull this back into ipfs.
- make it easy to test and verify the hamt data struct stuff is working well. meaning make some tests that shard large directories well. and be able to produce graphviz out of these to inspect and debug formation. 
- this is going to need very serious stress tests. so far i saw very light invariance testing but no serious stresses, with long sequences of operations expected to test how the HAMT grows and shrinks as stuff is added, or testing that the exact same data structures are produced despite sequences of operations being different.

unixfs/hamt/hamt.go
whyrusleeping: Yeah, i'd love to use an 'official' multihash table for this.

whyrusleeping:
@kevina if you want to review the algorithm for the HAMT datastructure and use that to check my implementation, that would be very helpful.

https://en.wikipedia.org/wiki/Hash_array_mapped_trie

whyrusleeping:
@jbenet I moved the HAMT code into its own repo and added a larger test to verify operation order doesnt change the resultant structure: https://github.com/whyrusleeping/unixfs-hamt

the `hamt_stress_test.go` will generate a random set of insertions and deletions and run through them in different permutations and verify the outputs are consistent.

jbenet:
Thanks i will revivew this week

diasdavid:
@whyrusleeping @jbenet mentioned that it would be better rip out all the unixfs into its own repo and this actually might be extremely beneficial to point several people into something isolated they can review fully.

Also, I think the best review I can do is grab this implementation and write the implementation in JavaScript and possibly write the spec for it.

whyrusleeping:
@diasdavid I had it pulled out at one point, but it was really just for the sake of 'putting the code somewhere else to look at it'. Actually extracting this code is difficult because it would require extracting a large number of other go-ipfs packages.

whyrusleeping:
Rebased on latest master (1a365a86fe2048705052ed1feb33cccbf893eb9b).

@diasdavid @dignifiedquire The actual HAMT code is isolated in this single file: https://github.com/ipfs/go-ipfs/blob/07683a7d9a6e83c859c84e75b2eb47d2aee79aa7/unixfs/hamt/hamt.go

All the review i'm looking for is just that file. the rest of the changes are just integration of that up through the stack and i'm not nearly as concerned about that part of the changeset.

unixfs/hamt/hamt.go
```diff
@@ -0,0 +1,466 @@
...
+}
+
+func (ds *HamtShard) indexForBitPos(bp int) int {
+	// TODO: an optimization could reuse the same 'mask' here and change the size
+	//       as needed. This isnt yet done as the bitset package doesnt make it easy
+	//       to do.
+	mask := new(big.Int).Sub(new(big.Int).Exp(big.NewInt(2), big.NewInt(int64(bp)), nil), big.NewInt(1))
+	mask.And(mask, ds.bitfield)
+
+	return popCount(mask)
```
dignifiedquire: I'm not entirely clear on how this function works/what exactly it's purpose is. And in turn I'm having a hard time verifying that the bit operations you do, do what they are supposed to be doing. Would be great to have a comment about what the function does and an explenation why the bit ops do what they do.

unixfs/hamt/hamt.go
```diff
@@ -0,0 +1,466 @@
+package hamt
+
+import (
```
dignifiedquire: To make it easier for people reading the code in the future it would be great to have a comment here at the top describing

- the exact type of HAMT that is implemented
- specific decisions made for this implementation

unixfs/hamt/hamt.go
dignifiedquire: Not the `popCount` that one is clear, but rather the whole `indexForBitPos`.

unixfs/hamt/util.go
```diff
@@ -0,0 +1,61 @@
+package hamt
+
+import (
+	"math/big"
+)
+
+type hashBits struct {
```
dignifiedquire: Would be nice to have some comments explaining what exactly the purpose of `hashBits` as a data structure is and how `.Next` works/what it returns

dignifiedquire:
Reviewed the test cases + the actual implementation in `hamt.go` from a algorithmic perspective, not from a go-code perspective. As far as I can tell this looks good to me 👍 Thanks @whyrusleeping for the additional comments, the make it a lot easier to follow the code along now.

Kubuxu:
One issue I can see is that we never downgrade form HMAT. This means that of someone reaches threshold and then removes files below treshold. We could downgrade but it could end up expensive when someone hovers (add del add del) around the threshold.

whyrusleeping:
@Kubuxu Yeah, you can manually downgrade by simply creating a new directory and copying each child into the new directory (as long as its below the sharding limit). We can add tooling for that later to help with that if it becomes an issue.

Kubuxu:
I was writing it on a phone in a bus and I lost part:
This means that if someone reaches threshold and then removes files below threshold he will get different hash which might be something that he is not expecting.

Also the value of threshold is now arbitrary, should we reason and try to find "best" threshold?

whyrusleeping:
@Kubuxu 

> Also the value of threshold is now arbitrary, should we reason and try to find "best" threshold?

Yeah, It needs to be small enough that moving from unsharded to sharded isnt really expensive, but large enough that its not wasteful to be using a hamt at that layer.

I'm open to suggestions

unixfs/hamt/hamt.go
```diff
@@ -0,0 +1,491 @@
...
+	format "github.com/ipfs/go-ipfs/unixfs"
+	upb "github.com/ipfs/go-ipfs/unixfs/pb"
+
+	node "gx/ipfs/QmUsVJ7AEnGyjX8YWnrwq9vmECVGwBQNAKPpgz5KSg8dcq/go-ipld-node"
+	proto "gx/ipfs/QmZ4Qi3GaRbjcx28Sme5eMH7RQjGkt8wHxt2a65oLaeFEV/gogo-protobuf/proto"
+	"gx/ipfs/QmfJHywXQu98UeZtGJBQrPAR6AtmDjjbe3qjTo9piXHPnx/murmur3"
+)
+
+const (
+	HashMurmur3 uint64 = 0x22
```
Kubuxu: Should we add murmur to multicodec/multihash?

unixfs/hamt/hamt.go
```diff
@@ -0,0 +1,491 @@
...
+	dserv dag.DAGService
+}
+
+// child can either be another shard, or a leaf node value
+type child interface {
+	Node() (node.Node, error)
+	Label() string
+}
+
+func NewHamtShard(dserv dag.DAGService, size int) *HamtShard {
```
Kubuxu: Probably makeHamtShard function should enforce it.

unixfs/hamt/hamt.go
whyrusleeping: We have a multicodec for it IIRC...

Oh, looks like @diasdavid ignored my PR and changed a bunch of stuff... :/

It was here: https://github.com/multiformats/multicodec/pull/23/files

diasdavid: I didn't ignore @whyrusleeping, I had asked a question on it -- https://github.com/multiformats/multicodec/pull/23#discussion_r86863096 --, the only thing that happened was that the main one got merged, so now you can PR to master :) 

Btw, new feature on github, if you click "edit" at the top level of the PR, you can pick a new target:

image

You don't have to close and open again

test/sharness/t0250-files-api.sh
```diff
@@ -481,7 +508,7 @@ test_files_api() {
 }
 
 # test offline and online
-test_files_api
+#test_files_api
```
lgierth: What about this disabled test?

lgierth:
This looks good to me overall

todo:
- [ ] disabled test (see other comment)
- [ ] murmur3 multicodec
- [ ] decide on downgrading -- I personally think it's pretty important

jbenet:
**Important!** If you make a threshold that isn't _perfectly deterministic_, you violate the entire point of using a CHAMP HAMT in the first place. Meaning: **don't do that!**. 

Either:
- [ ] switch to HAMT entirely (dont use the old data structure at all, no threshold madness)
- [ ] make it _perfectly deterministic_.  Trees with given children should ALWAYS be the exact same root hash, no matter what order of operations was involved.

Important requirement before merging:
- [ ] Make a test case that VERIFIES this stability. We cannot bear any non-stable data structure here.

whyrusleeping:
@lgierth @jbenet what should happen if someone tries modifying an old style directory?

kevina:
Could we make the new style directory a distinct Cid type?  Just a possibility.

whyrusleeping:
@kevina eh... not really. Then it wouldnt be unixfs anymore. Plus, we can already tell if its a shard or not, we have the unixfs type field.

kevina:
Okay, just wondering.  It might have solved the problem, if its the old CId type than it is not HAMT if it is the new than it is either the old or HAMT based on the content (likely the size or number of entries.

whyrusleeping:
@kevina thats not the problem we're worried about. The issue is that given any method of creating a directory (any number of insertions or deletions in any order), we should always get the same exact object for the same set of entries.

the issue raised above is that if i add 1001 files to a directory, and then delete two, it won't be the same output object as if i had just added 999 in the first place.

kevina:
@whyrusleeping I was trying to help solve:

> what should happen if someone tries modifying an old style directory?

If not a CId type we could have one a new unixfs type field one for the old (say 0x1), one for the old that could be converted to a HAMT (0x2) is the numbers of files is over the threshold (say 1000) (0x3), and another for the new HAMT type.  If the unixfs type is 0x1 than that will never become 0x3, it it is 0x2 then if the size is over 1000 it will become 0x3.  Having a 0x2 type with a size of over 1000 would be disallowed.  In a similar manor having a size less than 1000 and a type of 0x3 will be disallowed.

The flipflopping between 0x2 and 0x3 will be unavoidable if you want to be deterministic.  If the user knows the size is going to be around this threshold they can explicitly set it to 0x1.

I hope this makes sense, if I am just adding noise to the conversation I apologize.

dignifiedquire:
@whyrusleeping could you rebase this onto `master` please?

whyrusleeping:
rebased!

whyrusleeping:
@pgte thanks for reporting, found the problem. The filestore node wrappers were getting passed places they werent welcome.

whyrusleeping:
Alright. I'm gonna rebase this, then add a flag to disable it so we can ship 0.4.7 with support for this via an experiments flag.

pgte:
@whyrusleeping One last disparity from the JS implementation:

Looks like the UnixFS fileSize is not being populated, and probably should.

/cc @diasdavid

dignifiedquire:
> Alright. I'm gonna rebase this, then add a flag to disable it so we can ship 0.4.7 with support for this via an experiments flag.

@whyrusleeping maybe in 0.4.8? :)

Kubuxu:
Wow, rebased. Is this coming in soon (TM)?

whyrusleeping:
@pgte Anything else we need to finish before this goes forward? I think we've agreed on not setting filesizes on the shards.

pgte:
@whyrusleeping Correct, I don't think we need anything else.

test/sharness/t0250-files-api.sh
Kubuxu: Same

Kubuxu:
Apart from one removed test: https://github.com/ipfs/go-ipfs/pull/3042#discussion_r107789873

test/sharness/t0250-files-api.sh
whyrusleeping: oh, huh. k

mfs/dir.go
```diff
@@ -235,9 +235,9 @@ func (d *Directory) ListNames() ([]string, error) {
 	return out, nil
 }
 
-func (d *Directory) List() ([]NodeListing, error) {
+func (d *Directory) List(ctx context.Context) ([]NodeListing, error) {
 	var out []NodeListing
-	err := d.ForEachEntry(context.TODO(), func(nl NodeListing) error {
```
Kubuxu: I shouldn't have missed that.

whyrusleeping:
Choo Choo! Here comes the merge train! :train::train::train::train::train::train::train::train:

PR #3048, opened 2016-08-06
boot2docker-cli is deprecated, removed from README

zramsay:
- removed instructions to use boot2docker from README, which [has been deprecated](https://github.com/boot2docker/boot2docker-cli#deprecated) in favour of [docker-machine](https://docs.docker.com/machine/migrate-to-machine/).
- in most cases, [Docker for Mac/Windows](https://blog.docker.com/2016/07/docker-for-mac-and-windows-production-ready/) (now out of beta) will do the job => some [Mac users will still need docker-machine](https://docs.docker.com/docker-for-mac/#/what-to-know-before-you-install).

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,6 @@ Please put all issues regarding the Go IPFS _implementation_ in [this repo](http
 - [Getting Started](#getting-started)
   - [Some things to try](#some-things-to-try)
   - [Docker usage](#docker-usage)
-    - [Docker usage with VirtualBox/boot2docker (OSX and Windows)](#docker-usage-with-virtualboxboot2docker-osx-and-windows)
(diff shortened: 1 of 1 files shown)


whyrusleeping:
Cool, thats good to hear!

LGTM

jbenet:
@zramsay should we have a note about docker-machine instead? or will "users just know what to do"?

zramsay:
@jbenet I had considered it, though it's no longer the recommended way to run docker & only necessary for Mac's older than 2011. I suppose the PR message could have been clearer that Docker For Mac/Windows is the recommended way of running Docker on !Linux. IMO, the docker docs cover "what users need to know" to get started with either.

PR #3070, opened 2016-08-10
Feat/changelog 0.4.3 rc3

whyrusleeping:
This should be the final release candidate. Unless something horribly wrong is found, this is it.

We're also waiting on go1.7 to be officially released so that the binaries we release will work on macOS Sierra.


whyrusleeping:
alright, travis CI is being weird... why is that failing? cc @Kubuxu

whyrusleeping:
i'm going to assume these travis issues are an issue with travis itself and merge this

PR #3073, opened 2016-08-11
Add test for 'ipfs config replace'

Kubuxu:


test/sharness/t0021-config.sh
```diff
@@ -81,8 +94,6 @@ test_config_cmd() {
        test_cmp ident_exp ident_out
   '
 
-  # SECURITY
-  # Those tests are here to prevent exposing the PrivKey on the network
```
whyrusleeping: whyd you remove this comment?

whyrusleeping: oh nvm, i see that you moved it up

whyrusleeping:
Can you maybe break that test up a little bit? maybe into a setup step, the command execution, and the output check?

whyrusleeping:
@Kubuxu any update here?

Kubuxu:
How does it look now?

whyrusleeping:
Much better, except you've angered the mighty travis CI.

Kubuxu:
@whyrusleeping OSX/BSD sed is behaving quite strangely with `-i` flag and it looks like it will not support search and execute. Can I add gnu-sed as a requirement on OSX/BSD based systems?

//cc @chriscool

chriscool:
@Kubuxu I think it is better to just avoid using the -i option when using sed.
You can instead redirect into new files for example like:

```
	sed -e /PrivKey/d -e s/10GB/11GB/ newconfig.orig.json >newconfig.step.json &&
	sed -e '"'"'/PeerID/ { s/,$// } '"'"' newconfig.step.json >newconfig.json
```

chriscool:
It looks like there are still sed related problems:

```
expecting success: 

    cp "$IPFS_PATH/config" newconfig.json &&
	</dev/null sed -e /PrivKey/d -e s/10GB/11GB/ -i"~"  newconfig.json &&
	</dev/null sed -e '/PeerID/ { s/,$// } ' -i"~" newconfig.json

sed: 1: "/PeerID/ { s/,$// } 
": bad flag in substitute command: '}'
not ok 49 - setup for config replace test
```

Kubuxu:
It works, Shanress on OSX passed.

whyrusleeping:
LGTM, thanks @Kubuxu and @chriscool :)

Issue #3053, opened 2016-08-07
Hit ~80% code coverage on all packages in this repo


whyrusleeping:
In order to make releases more painless, and the code more stable and dependable in general I would like to try and hit a goal of at least 80% coverage (line based) on all packages.

Here is a nice list of what all that entails:
- [ ] github.com/ipfs/go-ipfs/blocks
- [ ] github.com/ipfs/go-ipfs/blocks/blockstore
- [ ] github.com/ipfs/go-ipfs/blocks/blocksutil
- [ ] github.com/ipfs/go-ipfs/blocks/bloom
- [ ] github.com/ipfs/go-ipfs/blocks/key
- [ ] github.com/ipfs/go-ipfs/blocks/set
- [ ] github.com/ipfs/go-ipfs/blockservice
- [ ] github.com/ipfs/go-ipfs/blockservice/test
- [ ] github.com/ipfs/go-ipfs/commands/cli
- [ ] github.com/ipfs/go-ipfs/commands/files
- [ ] github.com/ipfs/go-ipfs/commands/http
- [ ] github.com/ipfs/go-ipfs/core
- [ ] github.com/ipfs/go-ipfs/core/commands
- [ ] github.com/ipfs/go-ipfs/core/commands/files
- [ ] github.com/ipfs/go-ipfs/core/commands/object
- [ ] github.com/ipfs/go-ipfs/core/commands/unixfs
- [ ] github.com/ipfs/go-ipfs/core/corehttp
- [ ] github.com/ipfs/go-ipfs/core/corenet
- [ ] github.com/ipfs/go-ipfs/core/corerepo
- [ ] github.com/ipfs/go-ipfs/core/corerouting
- [ ] github.com/ipfs/go-ipfs/core/coreunix
- [ ] github.com/ipfs/go-ipfs/core/mock
- [ ] github.com/ipfs/go-ipfs/diagnostics
- [ ] github.com/ipfs/go-ipfs/exchange
- [ ] github.com/ipfs/go-ipfs/exchange/bitswap
- [ ] github.com/ipfs/go-ipfs/exchange/bitswap/decision
- [ ] github.com/ipfs/go-ipfs/exchange/bitswap/message
- [ ] github.com/ipfs/go-ipfs/exchange/bitswap/network
- [ ] github.com/ipfs/go-ipfs/exchange/bitswap/notifications
- [ ] github.com/ipfs/go-ipfs/exchange/bitswap/testnet
[...]

PR #3074, opened 2016-08-11
test: 81% coverage on blockstore

Kubuxu:
Coverage report available at: https://ipfs.io/ipfs/QmTuMtwGCfHrbYyZdQ1RaGNwS2MGsmAkjA8AaB69N7Ya1g/coverage.html#file0

Part of #3053

License: MIT

### DIFF
--- a/blocks/blockstore/arc_cache_test.go
+++ b/blocks/blockstore/arc_cache_test.go
@@ -2,12 +2,15 @@ package blockstore
 
 import (
 	"github.com/ipfs/go-ipfs/blocks"
+	"github.com/ipfs/go-ipfs/blocks/key"
 	"testing"
 
 	ds "gx/ipfs/QmTxLSvdhwg68WJimdS6icLPhZi28aTp6b7uihC2Yb47Xk/go-datastore"
 	syncds "gx/ipfs/QmTxLSvdhwg68WJimdS6icLPhZi28aTp6b7uihC2Yb47Xk/go-datastore/sync"
 	context "gx/ipfs/QmZy2y8t9zQH2a1b8q2ZSLKp17ATuJoCNxxyMFG5qFExpt/go-net/context"
 )
+
+var exampleBlock = blocks.NewBlock([]byte("foo"))
 
 func testArcCached(bs GCBlockstore, ctx context.Context) (*arccache, error) {
 	if ctx == nil {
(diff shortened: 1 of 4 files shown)


Kubuxu:
Docker tests failed but it isn't connected.

blocks/blockstore/blockstore_test.go
```diff
@@ -57,16 +57,21 @@ func TestRuntimeHashing(t *testing.T) {
 	bs := NewBlockstore(ds_sync.MutexWrap(ds.NewMapDatastore()))
 	bl := blocks.NewBlock([]byte("some data"))
 	blBad, err := blocks.NewBlockWithHash([]byte("some other data"), bl.Key().ToMultihash())
+	bl2 := blocks.NewBlock([]byte("some other data"))
```
whyrusleeping: why is this here before the error check from the previous statement?

blocks/blockstore/caching_test.go
```diff
@@ -0,0 +1,35 @@
+package blockstore
+
+import "testing"
+
+func TestCachingOptsLessThanZero(t *testing.T) {
+	opts := DefaultCacheOpts()
+	opts.HasARCCacheSize = -1
+
+	if _, err := CachedBlockstore(nil, nil, opts); err == nil {
+		t.Fatal()
```
whyrusleeping: these fatals should ideally give us some small idea of why it failed

blocks/blockstore/blockstore_test.go
Kubuxu: because it should always success and it is checked in other tests, will add check either way.

whyrusleeping: still, you want to check errors right as they are created, otherwise it just looks sloppy.

Kubuxu:
Updated.

blocks/blockstore/caching_test.go
```diff
@@ -0,0 +1,35 @@
+package blockstore
+
+import "testing"
+
+func TestCachingOptsLessThanZero(t *testing.T) {
+	opts := DefaultCacheOpts()
+	opts.HasARCCacheSize = -1
+
+	if _, err := CachedBlockstore(nil, nil, opts); err == nil {
+		t.Error("wrong ARC setting was not detected")
```
whyrusleeping: Any reason for `Error` over `Fatal`?

Kubuxu: Those test cases are not dependent on each other, so using `Error` all test cases will checked and printed.

`Error = Print + Fail`
`Fatal = Print + FailNow`

whyrusleeping: I guess thats fair.

blocks/blockstore/arc_cache_test.go
```diff
@@ -42,26 +59,119 @@ func TestRemoveCacheEntryOnDelete(t *testing.T) {
...
+	}
+
+	trap("has hit datastore", cd, t)
+
+	if has, err := arc.Has(exampleBlock.Key()); !has || err != nil {
+		t.Fatal("has returned invalid result")
+	}
+}
+
+func TestGetAndDeleteFalseShortCirciot(t *testing.T) {
```
whyrusleeping: typo here, sorry.

whyrusleeping:
just a typo then LGTM

Kubuxu:
Done.

PR #3077, opened 2016-08-12
deps: move go-is-domain to gx

Kubuxu:
It uses [this](https://github.com/jbenet/go-is-domain/pull/5) version of go-is-domain.

License: MIT


whyrusleeping:
LGTM

PR #3084, opened 2016-08-15
test: 100% coverage on blocks/set

Kubuxu:
License: MIT

### DIFF
--- a/blocks/set/set_test.go
+++ b/blocks/set/set_test.go
@@ -1,0 +1,71 @@
+package set
+
+import (
+	"testing"
+
+	bu "github.com/ipfs/go-ipfs/blocks/blocksutil"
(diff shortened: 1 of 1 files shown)


whyrusleeping:
whats with all the bithacking logic? It seems you could make this a tad more readable

Kubuxu:
It makes the slice with elements that have some transformations applied deterministically in this case:

half is added, different half is removed and different half is re-added again.

This way we  cover following cases: just add, add and removed, add removed and add, remove on element not in the set.

I think I can make it a bit more readable

whyrusleeping:
alright, still a bit hacky looking, but LGTM

whyrusleeping:
@Kubuxu want to make sure to check off the boxes in https://github.com/ipfs/go-ipfs/issues/3053

PR #3086, opened 2016-08-15
test: 82% coverage on blocks

Kubuxu:
License: MIT

### DIFF
--- a/blocks/blocks.go
+++ b/blocks/blocks.go
@@ -11,6 +11,8 @@ import (
 	u "gx/ipfs/QmZNVWh8LLjAavuQ2JXuFmuYH3C11xo988vSgp7UQrTRj1/go-ipfs-util"
 )
 
+var errWrongHash = errors.New("Data did not match given hash!")
+
 type Block interface {
 	Multihash() mh.Multihash
 	Data() []byte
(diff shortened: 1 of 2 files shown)


blocks/blocks_test.go
```diff
@@ -14,3 +20,77 @@ func TestBlocksBasic(t *testing.T) {
...
+	}
+
+	if !bytes.Equal(block.Multihash(), hash) {
+		t.Error("wrong multihash")
+	}
+
+	u.Debug = true
+
+	block, err = NewBlockWithHash(data, hash)
+	if err == nil {
```
whyrusleeping: I'd prefer to explicitly check errors in cases like this. I've had a few instances of tests passing while the code itself was breaking because a different error was being returned

whyrusleeping:
One comment, otherwise LGTM

blocks/blocks.go
```diff
@@ -11,6 +11,8 @@ import (
 	u "gx/ipfs/QmZNVWh8LLjAavuQ2JXuFmuYH3C11xo988vSgp7UQrTRj1/go-ipfs-util"
 )
 
+var errWrongHash = errors.New("Data did not match given hash!")
```
whyrusleeping: export this for checking against outside of this package

Kubuxu:
Exposed.

Kubuxu:
@whyrusleeping this is done, I just re-run travis tests, as at the time OSX tests were failing everywhere.

PR #3087, opened 2016-08-15
Update dns help with a correct domain name

kevinsimper:
Closes https://github.com/ipfs/faq/issues/161

### DIFF
--- a/core/commands/dns.go
+++ b/core/commands/dns.go
@@ -26,7 +26,7 @@ This command resolves those links to the referenced object.
 
 For example, with this DNS TXT record:
 
-	> dig +short TXT ipfs.io
+	> dig +short TXT _dnslink.ipfs.io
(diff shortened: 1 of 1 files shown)


core/commands/dns.go
```diff
@@ -26,7 +26,7 @@ This command resolves those links to the referenced object.
 
 For example, with this DNS TXT record:
 
-	> dig +short TXT ipfs.io
+	> dig +short TXT _dnslink.ipfs.io
```
jbenet: this change does not need to happen. it can be either. I'm not sure we _want_ it to happen.

the documentation reads "with this X, then Y" it does not require a working DNS record, and it could use `example.com` instead.

kevinsimper: So you want a broken example? I don't understand?

Kubuxu: Working example is much better than dummy one, `example.com` is nice because it works, you can fetch A record of it, you can open it via HTTP, but it won't work in this case.

whyrusleeping:
I think this is fine. LGTM

PR #3091, opened 2016-08-16
datastore: blockstore should retry when it encounters temp errors

whyrusleeping:
This is still a bit of a WIP, i'm not sure if we should inject the retry datastore in core/builder, or in the blockstore constructor.

License: MIT

### DIFF
--- a/core/builder.go
+++ b/core/builder.go
@@ -4,6 +4,9 @@ import (
 	"crypto/rand"
 	"encoding/base64"
 	"errors"
+	"os"
+	"syscall"
+	"time"
 
 	bstore "github.com/ipfs/go-ipfs/blocks/blockstore"
 	key "github.com/ipfs/go-ipfs/blocks/key"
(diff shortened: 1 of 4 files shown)


Kubuxu:
IMO core/builder is good, blockstore is quite general and it shouldn't know about datastore problems.

core/builder.go
```diff
@@ -127,14 +131,30 @@ func NewNode(ctx context.Context, cfg *BuildCfg) (*IpfsNode, error) {
...
 	if err := n.loadID(); err != nil {
 		return err
 	}
 
+	rds := &retry.Datastore{
+		Batching:    n.Repo.Datastore(),
+		Delay:       time.Millisecond * 200,
+		Retries:     6,
+		TempErrFunc: isTooManyFDError,
+	}
```
jbenet: take that, disk! 💿

whyrusleeping:
@Kubuxu LGTY?

Kubuxu:
LGTM

PR #3098, opened 2016-08-18
fuse: make osxfuse check config check more permissive

### LINKED ISSUE #2650: Upgraded Fuse emits CLI warning OSX, fails to mount with error
Installed stable fuse 2.8.3. CLI warns:
```
Error: unable to check fuse version.

Dear User,

Before mounting, we must check your version of OSXFUSE. We are protecting
you from a nasty kernel panic we found in OSXFUSE versions <2.7.2.[1]. To
make matters worse, it's harder than it should be to check whether you have
the right version installed...[2]. We've automated the process with the
help of a little tool. We tried to install it, but something went wrong[3].
Please install it yourself by running:

	go get github.com/jbenet/go-fuse-version/fuse-version

You can also stop ipfs from running these checks and use whatever OSXFUSE
version you have by running:

	ipfs config DontCheckOSXFUSE true

[1]: https://github.com/ipfs/go-ipfs/issues/177
[2]: https://github.com/ipfs/go-ipfs/pull/533
[3]: exit status 1
package github.com/jbenet/go-fuse-version/fuse-version: cannot download, $GOPATH not set. For more details see: go help gopath
```
I don't believe an installation of go should be a prerequisite for the check since I have installed an ipfs binary.

The key and value appear to have been added regardless (excerpt of `~/.ipfs/config` below):

```
  "Discovery": {
    "MDNS": {
      "Enabled": true,
      "Interval": 10
    }
  },
  "DontCheckOSXFUSE": true,
  "Gateway": {
    "HTTPHeaders": null,
    "PathPrefixes": [],
    "RootRedirect": "",
    "Writable": false
  },
```

Regardless, this results in an invalid key:
```
$ ipfs mount
Error: config key invalid: DontCheckOSXFUSE %!s(bool=true)
You may be able to get this error to go away by setting it again:

	ipfs config DontCheckOSXFUSE true

Either way, please tell us at: http://github.com/ipfs/go-ipfs/issues
```

whyrusleeping:
closes #2650 

License: MIT

### DIFF
--- a/fuse/node/mount_darwin.go
+++ b/fuse/node/mount_darwin.go
@@ -109,7 +109,7 @@ trying to run these checks with:
 [3]: %s
 `
 
-var errStrFixConfig = `config key invalid: %s %s
+var errStrFixConfig = `config key invalid: %s %v
 You may be able to get this error to go away by setting it again:
 
 	ipfs config %s true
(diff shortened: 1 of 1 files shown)


Kubuxu:
LGTM a bit cheaty.

whyrusleeping:
@Kubuxu when dealing with user input, cheaty is the name of the game. https://en.wikipedia.org/wiki/Robustness_principle

PR #3100, opened 2016-08-19
ipfs/cat: disable auto-gc check

whyrusleeping:
This functionality currently isnt very useful or reliable. Disabling it temporarily until we can refine the underlying code.

License: MIT

### DIFF
--- a/core/commands/cat.go
+++ b/core/commands/cat.go
@@ -5,7 +5,6 @@ import (
 
 	cmds "github.com/ipfs/go-ipfs/commands"
 	core "github.com/ipfs/go-ipfs/core"
-	"github.com/ipfs/go-ipfs/core/corerepo"
 	coreunix "github.com/ipfs/go-ipfs/core/coreunix"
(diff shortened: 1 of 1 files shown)


Kubuxu:
LGTM, it would be good to have a comment there but it isn't required.

core/commands/cat.go
```diff
@@ -42,10 +41,13 @@ var CatCmd = &cmds.Command{
...
-			res.SetError(err, cmds.ErrNormal)
-			return
-		}
+		/*
+			if err := corerepo.ConditionalGC(req.Context(), node, length); err != nil {
+				res.SetError(err, cmds.ErrNormal)
+				return
+			}
+		*/
+
```
lgierth: This could have just been removed. We have git history and the changelog, and this code isn't really complicated enough to keep it for future reference :)

whyrusleeping: Yeah... youre probably right.

PR #3101, opened 2016-08-19
reprovider: add config option to set reprovide interval

whyrusleeping:
License: MIT

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -169,7 +169,20 @@ func (n *IpfsNode) startOnlineServices(ctx context.Context, routingOption Routin
 	}
 
 	n.Reprovider = rp.NewReprovider(n.Routing, n.Blockstore)
-	go n.Reprovider.ProvideEvery(ctx, kReprovideFrequency)
+
+	if cfg.Reprovider.Interval != "0" {
+		interval := kReprovideFrequency
+		if cfg.Reprovider.Interval != "" {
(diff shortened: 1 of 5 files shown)


Kubuxu:
LGTM

core/core.go
```diff
@@ -169,7 +169,20 @@ func (n *IpfsNode) startOnlineServices(ctx context.Context, routingOption Routin
...
+		if cfg.Reprovider.Interval != "" {
+			dur, err := time.ParseDuration(cfg.Reprovider.Interval)
+			if err != nil {
+				return err
+			}
+
+			interval = dur
+		}
+
+		go n.Reprovider.ProvideEvery(ctx, interval)
```
jbenet: - i wonder if this will have serious adverse effects by triggering or not triggering a whole large codepath.
- I think this may be more resistant to error and code modification if we always do `go n.Reprovider.ProvideEvery(ctx, interval)` instead of guarding it in an if clause. The func can treats `0` (or a negative number) as "oh, do nothing? ok and it either waits or exits". Your code becomes:

```go
interval := kReprovideFrequency
if cfg.Reprovider.Interval != "" {
  dur, err := time.ParseDuration(cfg.Reprovider.Interval)
  if err != nil {
    return err
  }
  interval = dur
}
go n.Reprovider.ProvideEvery(ctx, interval)
```

And you _always_ guard in `n.Reprovider.ProvideEvery` which should happen already.

PR #3103, opened 2016-08-20
pin: use separate dagservice for storing pinsets

whyrusleeping:
License: MIT

### DIFF
--- a/core/builder.go
+++ b/core/builder.go
@@ -171,13 +171,15 @@ func setupNode(ctx context.Context, n *IpfsNode, cfg *BuildCfg) error {
 
 	n.Blocks = bserv.New(n.Blockstore, n.Exchange)
 	n.DAG = dag.NewDAGService(n.Blocks)
-	n.Pinning, err = pin.LoadPinner(n.Repo.Datastore(), n.DAG)
+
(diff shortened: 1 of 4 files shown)


core/builder.go
```diff
@@ -171,13 +171,15 @@ func setupNode(ctx context.Context, n *IpfsNode, cfg *BuildCfg) error {
 
 	n.Blocks = bserv.New(n.Blockstore, n.Exchange)
 	n.DAG = dag.NewDAGService(n.Blocks)
-	n.Pinning, err = pin.LoadPinner(n.Repo.Datastore(), n.DAG)
+
+	internalDag := dag.NewDAGService(bserv.New(n.Blockstore, offline.Exchange(n.Blockstore)))
```
Kubuxu: We are using offline DAG service in many plances, maybe we should create helper function for that?

whyrusleeping: yeah, thats probably a good idea

Kubuxu:
LGTM

whyrusleeping:
Yeah, lets do that refactor later. Its going to touch a lot of files and i prefer to keep PRs focused

PR #3105, opened 2016-08-20
bitswap: don't re-provide blocks we've provided very recently

whyrusleeping:
Lots of operations will store the same block multiple times in rapid succession, we have optimizations to avoid putting those to disk multiple times, but nothing to avoid providing them multiple times.

This is *a* solution. The other option is to use the same logic as the blockstore (if we have the block already, noop) for providing. The issue with that is that you would no longer be able to re-add a file as a way to re-provide it.

cc @jbenet @Kubuxu for feedback

License: MIT

### DIFF
--- a/blockservice/blockservice.go
+++ b/blockservice/blockservice.go
@@ -43,7 +43,15 @@ func New(bs blockstore.Blockstore, rem exchange.Interface) *BlockService {
 // TODO pass a context into this if the remote.HasBlock is going to remain here.
 func (s *BlockService) AddBlock(b blocks.Block) (key.Key, error) {
 	k := b.Key()
-	err := s.Blockstore.Put(b)
+	has, err := s.Blockstore.Has(k)
+	if err != nil {
+		return k, err
+	}
+	if has {
+		return k, nil
+	}
+
+	err = s.Blockstore.Put(b)
 	if err != nil {
 		return k, err
 	}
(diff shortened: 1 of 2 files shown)


whyrusleeping:
@jbenet the fact that adding a file again makes a reprovide happen is currently a side effect, and isnt exactly documented expected behaviour. I don't believe that changing that, and adding a separate command to reprovide a hash would count as a 'breaking change' do you?

jbenet:
> I don't believe that changing that, and adding a separate command to reprovide a hash would count as a 'breaking change' do you?

It's not officially one, no. it's undocumented behavior. that said, it's THE main way that i get something to work often-- meaning many times adding again will cause provides to actually get to more nodes and help the gateway find me. I would worry that this solution would decrease significantly the UX, i guess making it clear to use the other command may be ok. In general we just need to improve content routing as a whole.

jbenet:
We should quantify how much BW this is expected to improve on real workloads.

whyrusleeping:
In a few tests i ran, it provided less than half the number of blocks over the course of a workload modeled after the one @parkan described in that other issue. I can measure bandwidth soon.

whyrusleeping:
in a few tests i ran using iptb, in a cluster of 10 nodes, where one of them adds the go-ipfs source directory:

- on master (averages over three runs)
  - outgoing bandwidth:  12.2MB
  - incoming bandwidth: 4.5MB
- on this branch:
  - outgoing bandwidth: 7.9MB
  - incoming bandwidth: 2.9MB

whyrusleeping:
Ran more detailed tests, The process was a fresh repo each time, connecting to the main network and adding the ipfs source tree. Bandwidth utilization was recorded by nethogs.

| ipfs version | in | out |
-------------------------------------
| 0.4.2 | 1.3GB | 297MB |
| 0.4.3-rc3 | 951MB | 153MB |
| This PR | 217MB | 50MB |

whyrusleeping:
@Kubuxu CR?

Kubuxu:
LGTM.

whyrusleeping:
travis CI OSX tests appear to be failling for all our repos even though they don't report any actual failures.

blockservice/blockservice.go
```diff
@@ -54,13 +62,27 @@ func (s *BlockService) AddBlock(b blocks.Block) (key.Key, error) {
...
+
+	err := s.Blockstore.PutMany(toput)
 	if err != nil {
 		return nil, err
 	}
 
 	var ks []key.Key
-	for _, b := range bs {
+	for _, b := range toput {
 		if err := s.Exchange.HasBlock(b); err != nil {
```
kevina: @whyrusleeping, this change modifies the return value so that only the keys that don't already exist are returned.  Was this intentional?

PR #3106, opened 2016-08-20
cmds: implement ipfs dht provide command

### LINKED ISSUE #3099: add `ipfs dht provide` command
It would be nice to have a provide command as part of the ipfs dht subcommands. I think its the only DHT rpc not currently exposed through the api

whyrusleeping:
closes #3099 

License: MIT

### DIFF
--- a/core/commands/dht.go
+++ b/core/commands/dht.go
@@ -352,6 +352,7 @@ func provideKeys(ctx context.Context, r routing.IpfsRouting, keys []key.Key) {
 }
 
 func provideKeysRec(ctx context.Context, r routing.IpfsRouting, dserv dag.DAGService, keys []key.Key) {
+	provided := make(map[key.Key]struct{})
 	for _, k := range keys {
 		kset := key.NewKeySet()
 		node, err := dserv.Get(ctx, k)
@@ -371,6 +372,10 @@ func provideKeysRec(ctx context.Context, r routing.IpfsRouting, dserv dag.DAGSer
 		}
 
 		for _, k := range kset.Keys() {
+			if _, ok := provided[k]; ok {
+				continue
+			}
+
 			err = r.Provide(ctx, k)
 			if err != nil {
 				notif.PublishQueryEvent(ctx, &notif.QueryEvent{
@@ -379,6 +384,7 @@ func provideKeysRec(ctx context.Context, r routing.IpfsRouting, dserv dag.DAGSer
 				})
 				return
 			}
+			provided[k] = struct{}{}
 		}
 	}
 
--- a/core/commands/dht.go
+++ b/core/commands/dht.go
@@ -9,12 +9,15 @@ import (
 
 	key "github.com/ipfs/go-ipfs/blocks/key"
 	cmds "github.com/ipfs/go-ipfs/commands"
+	dag "github.com/ipfs/go-ipfs/merkledag"
 	notif "github.com/ipfs/go-ipfs/notifications"
 	path "github.com/ipfs/go-ipfs/path"
+	routing "github.com/ipfs/go-ipfs/routing"
 	ipdht "github.com/ipfs/go-ipfs/routing/dht"
 	pstore "gx/ipfs/QmQdnfvZQuhdT93LNc5bos52wAmdr3G2p6G8teLJMEN32P/go-libp2p-peerstore"
 	peer "gx/ipfs/QmRBqJF7hb8ZSpRcMwUt8hNhydWcxGEhtk81HKq6oUwKvs/go-libp2p-peer"
 	u "gx/ipfs/QmZNVWh8LLjAavuQ2JXuFmuYH3C11xo988vSgp7UQrTRj1/go-ipfs-util"
+	"gx/ipfs/QmZy2y8t9zQH2a1b8q2ZSLKp17ATuJoCNxxyMFG5qFExpt/go-net/context"
 )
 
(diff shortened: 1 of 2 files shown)


core/commands/dht.go
```diff
@@ -9,9 +9,12 @@ import (
 
 	key "github.com/ipfs/go-ipfs/blocks/key"
 	cmds "github.com/ipfs/go-ipfs/commands"
+	dag "github.com/ipfs/go-ipfs/merkledag"
 	notif "github.com/ipfs/go-ipfs/notifications"
 	path "github.com/ipfs/go-ipfs/path"
+	routing "github.com/ipfs/go-ipfs/routing"
 	ipdht "github.com/ipfs/go-ipfs/routing/dht"
+	"golang.org/x/net/context"
```
Kubuxu: This shouldn't be here.
Maybe we should wait to clear up Go 1.7 situation in https://github.com/ipfs/go-ipfs/pull/3110

whyrusleeping:
@Kubuxu could I get some review here? Getting this in ASAP will be very helpful

core/commands/dht.go
```diff
@@ -227,6 +231,159 @@ var findProvidersDhtCmd = &cmds.Command{
 	Type: notif.QueryEvent{},
 }
 
+var provideRefDhtCmd = &cmds.Command{
+	Helptext: cmds.HelpText{
+		Tagline: "Announce to the network that you are providing given values.",
+	},
+
+	Arguments: []cmds.Argument{
+		cmds.StringArg("key", true, true, "The key to find providers for.").EnableStdin(),
```
Kubuxu: "The key to find providers for."

core/commands/dht.go
```diff
@@ -227,6 +231,159 @@ var findProvidersDhtCmd = &cmds.Command{
...
+				return
+			}
+
+			if !has {
+				res.SetError(fmt.Errorf("block %s not found locally, cannot provide", k), cmds.ErrNormal)
+				return
+			}
+
+			keys = append(keys, k)
+		}
```
Kubuxu: This makes the command work in buffer mode instead pipe, do we want that?

whyrusleeping: its not exactly in buffer mode. This part should be quite fast, and can error out. We'd like to be able to error out easily in these cases

core/commands/dht.go
```diff
@@ -227,6 +231,159 @@ var findProvidersDhtCmd = &cmds.Command{
...
+				Extra: err.Error(),
+			})
+		}
+
+		err = dag.EnumerateChildrenAsync(ctx, dserv, node, kset)
+		if err != nil {
+			notif.PublishQueryEvent(ctx, &notif.QueryEvent{
+				Type:  notif.QueryError,
+				Extra: err.Error(),
+			})
```
Kubuxu: Should there be a `return` here?

core/commands/dht.go
```diff
@@ -227,6 +231,159 @@ var findProvidersDhtCmd = &cmds.Command{
...
+
+func provideKeysRec(ctx context.Context, r routing.IpfsRouting, dserv dag.DAGService, keys []key.Key) {
+	for _, k := range keys {
+		kset := key.NewKeySet()
+		node, err := dserv.Get(ctx, k)
+		if err != nil {
+			notif.PublishQueryEvent(ctx, &notif.QueryEvent{
+				Type:  notif.QueryError,
+				Extra: err.Error(),
+			})
```
Kubuxu: Should there be a return here?

core/commands/dht.go
```diff
@@ -227,6 +231,159 @@ var findProvidersDhtCmd = &cmds.Command{
...
+		}
+
+		err = dag.EnumerateChildrenAsync(ctx, dserv, node, kset)
```
whyrusleeping: Yeah, that would be nice. We should make a note to take another look at these interfaces and find better ways to do these things.

core/commands/dht.go
whyrusleeping: yes indeed

core/commands/dht.go
whyrusleeping: yeap, gonna refactor to make that class of problems go away

core/commands/dht.go
Kubuxu: What I mean is it will buffer whole stdin, instead of starting process for each line as we go.

but if we don't do it then we hit problem what to do if command errors out on input later.

Ok, let's leave it as it is, and create discussion issue about that.

core/commands/dht.go
```diff
@@ -227,6 +231,165 @@ var findProvidersDhtCmd = &cmds.Command{
...
+		if err != nil {
+			notif.PublishQueryEvent(ctx, &notif.QueryEvent{
+				Type:  notif.QueryError,
+				Extra: err.Error(),
+			})
```
Kubuxu: Should there be return here?

core/commands/dht.go
whyrusleeping: Ah, i see what youre saying. Yeah... we need to work that out a bit more i think

Kubuxu:
CI fails.

whyrusleeping:
@Kubuxu i think thats just a messed up build agent, i was working with them yesterday and the failure happened on a new one.

Kubuxu:
Ok, it doesn't look like Travis will run any time in close future but Circle says ok.

LGTM

PR #3107, opened 2016-08-20
routing: rework interfaces to make separation easier

whyrusleeping:
I've started to work on separating the content routing layer and the peer routing layer. The third bit that we glom into our current routing interface deals with direct values, i think thats what @diasdavid was starting to call the record store, but i'm not sure its the same scope. Definitely want some feedback here.

cc @jbenet @diasdavid 

License: MIT


diasdavid:
Yeah, seems on the right track to break Peer Routing and Content Routing into separate parts, so that we can have other types of Content Routing (therefore super fast providing)

See https://github.com/libp2p/interface-record-store and https://github.com/libp2p/interface-peer-routing also for more context.

PR #3109, opened 2016-08-21
Updated BASH auto complete

EliasGabrielsson:
- Added support for Files API with relevant completion.
- Local filesystem completion (eg. ipfs add /users/fileA.txt)
- Dynamic completion for ipfs update install
- Merkeldag completion (Partly broken for white spaces, need a little more escaping love on right spots)
- Multiaddr (Need a dirty hack for IP ranges)
- Correct all options and added auto complete.

Blocker list to get all working:

All bugs are marked with TODO:s inline as well.
- ipfs files ls -F´ is needed to get correct auto complete to the files api. See:ls -F`
- The same goes for `ipfs file ls -F´
- Alternatives to ipfs add --chunker, just replace placeholder options.

License: MIT
Signed-off-by: Elias Gabrielsson elias@benefactory.se


whyrusleeping:
the argument to `ipfs diag cmds set-time` is go's time.Duration notation: https://golang.org/pkg/time/#ParseDuration

For the `ls -F` options, thats definitely something you (or another willing contributor) could PR :)

I'll make a note for myself to get a list of placeholder options for the chunker flag.

whyrusleeping:
on a second look, we do a basically `ls -F` already on `ipfs ls`. We append a '/' to all the entries that are directories.

EliasGabrielsson:
@whyrusleeping 
Devils in the detail ;-)
Try: `ipfs file ls` instead of `ipfs ls`.

I will try to figure how a `-F` flag can be implemented.

whyrusleeping:
@EliasGabrielsson sweet, sounds good :)

In the meantime, does anyone know who would be good to review these? cc @Kubuxu @lgierth @chriscool

EliasGabrielsson:
I read today that `ipfs file ls` will be deprecated. In which timeframe are we talking, weeks, months or years?

jbenet:
@EliasGabrielsson were did you read it-- i don't think that's decided yet? it's an option (the only thing i can think of is the proposal to rename `ipfs files` to something else)

whyrusleeping:
@jbenet hes referring to `ipfs file ls` (not `ipfs ls` or `ipfs files ls`). Its a mostly redundant command that currently is pretty confusing. The only difference that it provides is that it reports unixfs file sizes (not including framing) where `ipfs ls` reports the size of the merkledag object.

I think it makes the most sense to deprecate the `ipfs file ls` command in favor of a flag on `ipfs ls`.

jbenet:
Ohhhh right.

Kubuxu:
Can you rebase it, we will merge it then.

EliasGabrielsson:
@Kubuxu Working on a patch to get the script inline with the latest version of ipfs.  Expect an updated pull request in the end of this week.

Kubuxu:
Thank you very much for that.

EliasGabrielsson:
@Kubuxu Rebased and updated to get in line with master. 
There are still some work to do but its fully workable now.

Kubuxu:
Something went wrong with that rebase. 250+ commits in the PR.

EliasGabrielsson:
@Kubuxu 
Apparently I somehow failed. I will close this pull request and make a new clean fork.
It's only one file which is edited anyway.

PR #3110, opened 2016-08-23
update deps for libp2p 3.4.0

whyrusleeping:
This will let us have multiple protocol handlers for the same protocols (and also allow us to upgrade protocols easily)

License: MIT


Kubuxu:
Teamcity doesn't know about Go 1.7, and
@lgierth we need to upgrade docker image to Go 1.7

PR #3118, opened 2016-08-24
remove randbo dep, its no longer needed

whyrusleeping:
I want to continue doing things like this as we find deps are no longer needed. A smaller dependency tree is a happier one 

License: MIT

### DIFF
--- a/mfs/mfs_test.go
+++ b/mfs/mfs_test.go
@@ -13,12 +13,6 @@ import (
 	"testing"
 	"time"
 
-	"github.com/ipfs/go-ipfs/path"
-	ds "gx/ipfs/QmNgqJarToRiq2GBaPJhkmW4B5BxS5B74E1rkGvv2JoaTp/go-datastore"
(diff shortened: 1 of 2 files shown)


whyrusleeping:
Tests failing because CI uses go1.5, the rand reader stuff landed in 1.6

mfs/mfs_test.go
```diff
@@ -552,7 +551,8 @@ func actorMakeFile(d *Directory) error {
 		return err
 	}
 
-	r := io.LimitReader(randbo.New(), int64(77*rand.Intn(123)))
+	rread := rand.New(rand.NewSource(time.Now().UnixNano()))
```
Kubuxu: we have `NewTimeSeededRand` in go-ipfs-util.

whyrusleeping: if we already import from there, then sure, lets use it. Otherwise i'd like to avoid another import of that package. I want it to die.

PR #3128, opened 2016-08-26
Expose corenet interface

magik6k:
This implements ipfs/notes#160 and obsoletes ipfs/go-ipfs#2765

I've tested it manually using two nodes and the connection worked.


core/commands/corenet.go
```diff
@@ -0,0 +1,176 @@
...
+
+		listener, err := corenet.Listen(n, "/app/"+req.Arguments()[1])
+		if err != nil {
+			res.SetError(err, cmds.ErrNormal)
+			return
+		}
+		defer listener.Close()
+
+		for {
+			remote, err := listener.Accept()
```
magik6k: I haven't found a way to get peer address. Should we left determining peer identity up to applications using this interface or is there actually a way to get peer ID?

whyrusleeping: the object returned here should be a stream that you can call `ID()` on (or `.Conn().ID()`)

core/commands/corenet.go
```diff
@@ -0,0 +1,176 @@
...
+		if err != nil {
+			res.SetError(err, cmds.ErrNormal)
+			return
+		}
+
+		if addr != nil {
+			n.Peerstore.AddAddr(peerID, addr, pstore.TempAddrTTL) // temporary
+		}
+
+		if len(n.Peerstore.Addrs(peerID)) == 0 {
```
whyrusleeping: instead of all this, just use `n.Host.Connect`

whyrusleeping: Actually, on second thought. `corenet.Dial` does that for you

core/commands/corenet.go
```diff
@@ -0,0 +1,176 @@
...
+			p, err := n.Routing.FindPeer(ctx, peerID)
+			if err != nil {
+				res.SetError(err, cmds.ErrNormal)
+				return
+			}
+			n.Peerstore.AddAddrs(p.ID, p.Addrs, pstore.TempAddrTTL)
+		}
+
+		remote, err := corenet.Dial(n, peerID, "/app/"+req.Arguments()[2])
+
```
whyrusleeping: dont put spaces between error creation and checking

core/commands/corenet.go
```diff
@@ -0,0 +1,176 @@
...
+				err := remote.Close()
+				if err != nil {
+					res.SetError(err, cmds.ErrNormal)
+					return
+				}
+				res.SetError(err, cmds.ErrNormal)
+				return
+			}
+
+			go func() {
```
whyrusleeping: since we're doing this pattern twice, lets make it a helper func

whyrusleeping:
Alright, i think this would be really fun to have merged in.

Heres what I see us possibly doing: We add a config field value or daemon flag for `ipfs experimental` or `ipfs exp`. When the flag is enabled, this command gets added under `ipfs exp corenet`. We should also make it clear in the `ipfs exp` helptext that commands under it are there solely for experimentation and should not be depended on as they can and will change with a good deal of frequency.

whyrusleeping:
The tests are failing because you have the wrong gx import paths, you can use `gx deps find go-libp2p-peerstore` to check what the right value is for those.

Kubuxu:
@whyrusleeping they are now failing because of that as HEAD changed dep.
@magik6k can you update it.

whyrusleeping:
I'm pretty happy with this. Will need to run the idea by @jbenet, but this is a pretty cool thing to have (even if its just an experiment)

Kubuxu:
Error from teamcity:
```
[19:41:18][Step 4/4] Listening on [0.0.0.0] (family 0, port 10002)
[19:41:18][Step 4/4] expecting success: 
[19:41:18][Step 4/4]   ipfsi 1 exp corenet dial $PEERID_0 /ip4/127.0.0.1/tcp/10002 corenet-test
[19:41:18][Step 4/4] Listening on [0.0.0.0] (family 0, port 10001)
[19:41:18][Step 4/4] 
[19:41:18][Step 4/4] Error: protocol not supported
[19:41:18][Step 4/4] Dial for server to client
[19:41:18][Dial for server to client] Dial for server to client
[19:41:18][Step 4/4] not ok 11 - Dial for server to client
[19:41:18][Step 4/4] #	
[19:41:18][Step 4/4] #	  ipfsi 1 exp corenet dial $PEERID_0 /ip4/127.0.0.1/tcp/10002 corenet-test
[19:41:18][Step 4/4] #	
[19:41:18][Step 4/4] 
[19:41:18][Step 4/4] expecting success: 
[19:41:18][Step 4/4]   kill $LISTENER_PID_PID &&
[19:41:18][Step 4/4]   wait $NC_SERVER_PID $NC_CLIENT_PID
[19:41:18][Step 4/4] 
[19:50:14][Step 4/4] The build go-ipfs::ci_tests_linux #3362 {buildId=5603} has been running for more than 15 minutes. Terminating...
[19:50:14][Step 4/4] Dumping threads before termination:
```

Kubuxu:
Circle is having different errors, something about wrong usage of netcat.

jbenet:
This is a cool idea, lemme review the interface when I'm back on land.

Btw finding a solid way to do this for libp2p would be ideal, and go-ipfs
could inherit the behavior.

Btw, @magik6k check out the plans over at https://github.com/ipfs/xtp --
it's a subproject we want to do. these interfaces MAY be able to relate.

whyrusleeping:
I'll take a look at these test failures when i get a chance. I want to try and get this in (at least as an experimental feature)

tswindell:
I think we should have the ability for a dial out to setup a local listening port, so client software can connect to it like it would a remote service. The current method requires apps to be listening on both sides, no matter whether it's a listen on dial call.

magik6k:
Closing as https://github.com/ipfs/go-ipfs/pull/3943 got merged

PR #3129, opened 2016-08-26
commands: replace <default> keyword with <<default>>

Kubuxu:
This way if we use <default> in description (we use thigns like <hash>
already), it won't conflict and modify the description in unwanted way.

Created because of: https://github.com/ipfs/go-ipfs/pull/2744#issuecomment-242655384

License: MIT


whyrusleeping:
LGTM

PR #3135, opened 2016-08-27
Directly Fix #2839

### LINKED ISSUE #2839: Adding files that are symlinks directly

kevina:
Revert the original fix for #2839 and instead fix the original issue so that adding the current directory works as expected without changing the semantics of how symbolic links are handled.

### DIFF
--- a/test/sharness/t0044-add-symlink.sh
+++ b/test/sharness/t0044-add-symlink.sh
@@ -44,6 +44,16 @@ test_add_symlinks() {
 		echo "QmWYN8SEXCgNT2PSjB6BnxAx6NJQtazWoBkTRH9GRfPFFQ" > badlink_exp &&
 		test_cmp badlink_exp badlink_out
 	'
+
+	test_expect_success "adding with symlink in middle of path is same as\
+adding with no symlink" '
+		mkdir -p files2/a/b/c &&
+		echo "some other text" > files2/a/b/c/foo &&
+		ln -s b files2/a/d
+		ipfs add -rq files2/a/b/c > no_sym &&
+		ipfs add -rq files2/a/d/c > sym &&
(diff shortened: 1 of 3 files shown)


test/sharness/t0044-add-symlink.sh
```diff
@@ -27,34 +23,27 @@ test_add_symlinks() {
...
+		ipfs add -q files/bad > badlink_out
 	'
 
 	test_expect_success "output looks good" '
 		echo "QmWYN8SEXCgNT2PSjB6BnxAx6NJQtazWoBkTRH9GRfPFFQ" > badlink_exp &&
 		test_cmp badlink_exp badlink_out
 	'
-
-	test_expect_success "adding with symlink in middle of path is same as\
-adding with no symlink" '
```
whyrusleeping: This test should still be valid.

kevina: Done.

whyrusleeping:
Alright, This LGTM. @jbenet since this is in response to your review, want to give it a quick look over?

jbenet:
LGTM

whyrusleeping:
Thanks @kevina!

NeoTheFox:
So, just to be clear, would we be able to make `ipfs add -r ` to resolve symlinks if needed now?

PR #3136, opened 2016-08-27
Fixes #3133: Properly handle release candidates in version comparison

### LINKED ISSUE #3133: build: Unable to build with Go release candidates
Version/Platform/Processor information: N/A;
commit 8ffab98d55 on Windows with MSYS2

Type: bug
Area: build system
Priority: P3 (release untested)

Description:
If using a Go release candidate (for example 1.7rc6) the 'check_go_version' utility fails to compare the version number to the minimal 1.5.2.
```
$ make install
bin/check_go_version: linia 69: test: 7rc6: oczekiwano wyrażenia całkowitego
fatal: go version '1.7rc6' should be at least '1.5.2'
=> Please take a look at https://golang.org/doc/install to install or upgrade go.
make: *** [Makefile:32: go_check] Błąd 1
```
the first message means `line 69: test: 7rc6: expected natural expression`

PS: I'm updating to the stable version anyway, but I thought this was worth reporting.

mateon1:
License: MIT
Signed-off-by: Mateusz Naściszewski <matin1111@wp.pl>

### DIFF
--- a/bin/check_go_version
+++ b/bin/check_go_version
@@ -23,55 +23,11 @@ GO_MIN_VERSION="$1"
 
 UPGRADE_MSG="Please take a look at https://golang.org/doc/install to install or upgrade go."
 
-die_upgrade() {
-    printf >&2 "fatal: %s\n" "$@"
-    printf >&2 "=> %s\n" "$UPGRADE_MSG"
-    exit 1
-}
-
-major_number() {
-    vers="$1"
-
-    # Hack around 'expr' exiting with code 1 when it outputs 0
-    case "$vers" in
-        0) echo "0" ;;
-        0.*) echo "0" ;;
-        *) expr "$vers" : "\([^.]*\).*" || return 1
-    esac
-}
-
-check_at_least_version() {
-    MIN_VERS="$1"
-    CUR_VERS="$2"
-    PROG_NAME="$3"
-
-    # Get major, minor and fix numbers for each version
-    MIN_MAJ=$(major_number "$MIN_VERS") || die "No major version number in '$MIN_VERS' for '$PROG_NAME'"
-    CUR_MAJ=$(major_number "$CUR_VERS") || die "No major version number in '$CUR_VERS' for '$PROG_NAME'"
-
-    if MIN_MIN=$(expr "$MIN_VERS" : "[^.]*\.\([^.]*\).*"); then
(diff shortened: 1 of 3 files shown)


whyrusleeping:
cc @chriscool and @lgierth for review

bin/check_go_version
```diff
@@ -23,55 +23,11 @@ GO_MIN_VERSION="$1"
...
-    test "$CUR_MAJ" -gt "$MIN_MAJ" || {
-        test "$CUR_MIN" -gt $(expr "$MIN_MIN" - 1) || die_upgrade "$VERS_LEAST"
-        test "$CUR_MIN" -gt "$MIN_MIN" || {
-            test "$CUR_FIX" -ge "$MIN_FIX" || die_upgrade "$VERS_LEAST"
-        }
-    }
-}
+# Get path to the directory containing this file
+# If $0 has no slashes, uses "./"
+PREFIX=$(expr $0 : "\(.*\/\)") || PREFIX='./'
```
chriscool: It might be safer to use `expr "$0"` instead of `expr $0` in case $0 contains spaces.

bin/check_version
```diff
@@ -0,0 +1,77 @@
...
+die_upgrade() {
+    printf >&2 "fatal: %s\n" "$@"
+    printf >&2 "=> %s\n" "$UPGRADE_MSG"
+    exit 1
+}
+
+major_number() {
+    vers="$1"
+
+    # Hack around 'expr' exiting with code 1 when it outputs 0
```
whyrusleeping: wat. Thats dumb. (I just ran this in my shell)

mateon1: I agree, but that's from the original version checking code by chriscool.
The same goes for an empty string as output, hence the digit captured in both regex alternations a few lines later.

whyrusleeping:
@chriscool Look good to you?

bin/check_version
chriscool: @whyrusleeping and @mateon1 I am not sure what you think is dump.
My opinion is that `expr` exiting with code 1 when it returns 0 is really dump indeed!
It looks like it comes from the fact that it can be used for arithmetics.

chriscool:
@whyrusleeping yeah LGTM now!

whyrusleeping:
Thanks @mateon1!

PR #3141, opened 2016-08-29
cmd: harden the security of privkey field in config show

Kubuxu:
also add test for lower case config replace.

### DIFF
--- a/core/commands/config.go
+++ b/core/commands/config.go
@@ -163,27 +163,11 @@ included in the output of this command.
 			return
 		}
 
-		idmap, ok := cfg["Identity"].(map[string]interface{})
-		if !ok {
-			res.SetError(errors.New("config has no identity"), cmds.ErrNormal)
+		err = scrubValue(cfg, []string{config.IdentityTag, config.PrivKeyTag})
+		if err != nil {
+			res.SetError(err, cmds.ErrNormal)
 			return
 		}
-
-		privKeyKey := "" // make sure we both find the name of privkey and we delete it
-		for key, _ := range idmap {
-			if strings.ToLower(key) == "privkey" {
-				if privKeyKey != "" {
-					res.SetError(errors.New("found multiple PrivKey keys"), cmds.ErrNormal)
-					return
-				}
-				privKeyKey = key
-			}
-		}
-		if privKeyKey == "" {
-			res.SetError(errors.New("haven't found PriveKey key"), cmds.ErrNormal)
-		}
-
-		delete(idmap, privKeyKey)
 
 		output, err := config.HumanOutput(cfg)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


whyrusleeping:
@Kubuxu I would do this more like: https://gist.github.com/whyrusleeping/ec7e22a5f331bf9f1309f38aa366a343

Just to be a bit more generic about scrubbing this, and making sure we don't have issues in other cases (such as "Identify" having its case changed)

core/commands/config.go
```diff
@@ -164,11 +165,25 @@ included in the output of this command.
...
 		idmap, ok := cfg["Identity"].(map[string]interface{})
 		if !ok {
-			res.SetError(fmt.Errorf("config has no identity"), cmds.ErrNormal)
+			res.SetError(errors.New("config has no identity"), cmds.ErrNormal)
 			return
 		}
 
-		delete(idmap, "PrivKey")
+		privKeyKey := "" // make sure we both find the name of privkey and we delete it
+		for key, _ := range idmap {
```
jbenet: - and `idmap` will be nil if `cfg["Identity"]` doesnt work and there is `cfg["identity"]` instead. 
- As we discussed, this is turning out to be a mess. easier to move the key out.

Kubuxu:
I have a different idea, let me show you that.

core/commands/config.go
Kubuxu: If it is nil then command fails, which fails the sharness tests.

whyrusleeping:
@Kubuxu wanna review this when you get a chance?

Kubuxu:
LGTM, we might want to add test time assertions using reflection or selectors on default config that those selectors are up to date.

whyrusleeping:
Alright, this looks good to me. :ship: :bullettrain_side: :bullettrain_front: :tractor: :train2:

core/commands/config.go
```diff
@@ -180,6 +179,47 @@ included in the output of this command.
...
+			}
+		}
+		return "", nil, false
+	}
+
+	cur := m
+	for _, k := range key[:len(key)-1] {
+		foundk, val, ok := find(cur, k)
+		if !ok {
+			return fmt.Errorf("failed to find specified key")
```
jbenet: wont this error out all the way to the user, if the user provides a config _without_ an `Identity.PrivKey` value?

test/sharness/t0021-config.sh
```diff
@@ -111,10 +111,10 @@ test_config_cmd() {
 
   test_expect_success "'ipfs config replace' injects privkey back" '
        ipfs config replace show_config &&
-	   grep "\"PrivKey\":" "$IPFS_PATH/config" | grep -e ": \".\+\"" >/dev/null
+       grep "\"PrivKey\":" "$IPFS_PATH/config" | grep -e ": \".\+\"" >/dev/null
```
jbenet: hard time reconciling this test case with https://github.com/ipfs/go-ipfs/pull/3141/files#r77544560. I thought it would error out on this case (which i would consider a bug-- i.e. i agree with this test case, it should not error to provide a config without key).

PR #3147, opened 2016-08-29
Fix config set private key protection

whyrusleeping:
Should address https://github.com/ipfs/go-ipfs/pull/2957/files#r76539324

### DIFF
--- a/repo/fsrepo/fsrepo.go
+++ b/repo/fsrepo/fsrepo.go
@@ -482,6 +482,14 @@ func (r *FSRepo) SetConfigKey(key string, value interface{}) error {
 		return err
 	}
 
+	// Load private key to guard against it being overwritten.
+	// NOTE: this is a temporary measure to secure this field until we move
+	// keys out of the config file.
+	pkval, err := common.MapGetKV(mapconf, "Identity.PrivKey")
+	if err != nil {
+		return err
+	}
+
 	// Get the type of the value associated with the key
 	oldValue, err := common.MapGetKV(mapconf, key)
 	ok := true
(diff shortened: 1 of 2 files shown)


jbenet:
@whyrusleeping -- @lgierth and i were discussing that it's likely _much_ easier to solve this problem by just moving the key out of the config entirely for good, and at this point we keep sinking energy into making the hack work when there's likely all these complicated edge cases to cover. 

We could move it out in a 4-5 migration, put it in a file in `$repo/keys/id` for now, and deal with future migrations to move to the proper keystore later. Our migrations work pretty reliably (yay!).

repo/fsrepo/fsrepo.go
```diff
@@ -482,6 +482,14 @@ func (r *FSRepo) SetConfigKey(key string, value interface{}) error {
 		return err
 	}
 
+	// Load private key to guard against it being overwritten.
+	// NOTE: this is a temporary measure to secure this field until we move
+	// keys out of the config file.
+	pkval, err := common.MapGetKV(mapconf, "Identity.PrivKey")
```
jbenet: - What if the name of the fields in the structs changes and people forget to change it here? should be a const.

whyrusleeping: I would merge a PR that cleanly moves the private key out of the config if
paired with a migration. I do not have time to work on that.

Kubuxu: @jbenet as we will be moving the key out of config in the nearest future, can we accept current hack (it is hack) and release keymigrations in 0.4.4 or 0.5.

This isn't long term strategy, it is a way to keep things running for till next release.

whyrusleeping:
closing, moved the changes here to #3141

PR #3149, opened 2016-08-30
fdlimit: improve fdlimit autoraising tests

whyrusleeping:
Better tests for this, at least on linux we can check it externally. I don't know of a good way to do this on OSX

### DIFF
--- a/test/sharness/t0060-daemon.sh
+++ b/test/sharness/t0060-daemon.sh
@@ -129,21 +129,13 @@ test_expect_success "daemon raised its fd limit" '
 	grep "raised file descriptor limit to 1024." actual_daemon > /dev/null
 '
 
-get_col_four() {
-	awk '{ print $4 }' $1
-}
+test_expect_success "daemon actually can handle 1024 file descriptors" '
+	hang-fds -hold=2s 1000 '$API_MADDR'
+'
 
-if [ `uname` == "Linux" ]; then
-	test_expect_success "get fd limit through /proc" '
-		cat /proc/$IPFS_PID/limits > limits &&
-		grep "Max open files" limits > fd_limits_line &&
-		limit=$(get_col_four fd_limits_line)
-	'
-
-	test_expect_success "limit from system looks good" '
-		test "$limit" -eq 1024
-	'
-fi
+test_expect_success "daemon didnt throw any errors" '
+	test_expect_code 1 grep "too many open files" daemon_err
+'
 
 test_kill_ipfs_daemon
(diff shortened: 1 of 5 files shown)


Kubuxu:
LGTM

@jbenet do you want to also give it a review?

test/sharness/t0060-daemon.sh
```diff
@@ -121,14 +121,30 @@ test_expect_success "daemon with pipe eventually becomes live" '
...
 test_expect_success "daemon raised its fd limit" '
-	grep "ulimit" actual_daemon > /dev/null
+	grep "raised file descriptor limit to 1024." actual_daemon > /dev/null
 '
 
+get_col_four() {
+	awk '{ print $4 }' $1
+}
+
+if [ `uname` == "Linux" ]; then
```
jbenet: - this is nice
- but we really need to also test in osx. and we shouldn't trust what the limits say, should test for sure that we can indeed open that many fds. 
- http://stackoverflow.com/questions/31115764/setrlimit-doesnt-work-on-os-x-maverick-and-yosemite
- we could do this with a tool like `open-fds <fds> <multiaddr>` which uses `nc` to hang fds open.
- then count the fds with `lsof`. maybe: `lsof -p $1 | grep -v " txt " | wc -l`

Kubuxu: The open-fds process would have to start from go-ipfs process, to inherit the fd limit.
Otherwise we are testing nothing.

Other approach would be to make the fd rising separate package, produce a small binary that tries opening files with and without the fd limit raised.

whyrusleeping: "this is nice"

image

jbenet: Made this for you https://github.com/jbenet/hang-fds

jbenet: I'm using `syscall.Setrlimit` in https://github.com/jbenet/hang-fds/blob/master/hang-fds.go#L32-L63

```
> ulimit -Sn 256
> ulimit -Sn
256
> ./hang-fds 2048 /ip4/127.0.0.1/tcp/1234
hanging 2048 fds at /ip4/127.0.0.1/tcp/1234
error: failed to raise fd limit to 2058 (still 256)
```

whyrusleeping: @jbenet you never actually change the fd limit in that code. you call getrlimit, you check the value, and then you you call setrlimit with the same object.

PR #3151, opened 2016-08-30
Enhance tests for files API root best-effort pin.

kevina:
License: MIT

### DIFF
--- a/test/sharness/t0252-files-gc.sh
+++ b/test/sharness/t0252-files-gc.sh
@@ -11,13 +11,15 @@ test_description="test how the unix files api interacts with the gc"
 test_init_ipfs
 
 test_expect_success "object not removed after gc" '
-  echo "hello world"  | ipfs files write --create /hello.txt &&
+  echo "hello world" > hello.txt &&
+  cat hello.txt | ipfs files write --create /hello.txt &&
   ipfs repo gc &&
   ipfs cat QmVib14uvPnCP73XaCDpwugRuwfTsVbGyWbatHAmLSdZUS
 '
 
 test_expect_success "/hello.txt still accessible after gc" '
-  ipfs files read /hello.txt
+  ipfs files read /hello.txt > hello-actual &&
+  test_cmp hello.txt hello-actual
 '
 
 ADIR_HASH=QmbCgoMYVuZq8m1vK31JQx9DorwQdLMF1M3sJ7kygLLqnW
@@ -27,7 +29,6 @@ test_expect_success "gc okay after adding incomplete node -- prep" '
   ipfs files mkdir /adir &&
   echo "file1" |  ipfs files write --create /adir/file1 &&
   echo "file2" |  ipfs files write --create /adir/file2 &&
-  ipfs cat $FILE1_HASH && 
   ipfs pin add --recursive=false $ADIR_HASH &&
   ipfs files rm -r /adir &&
   ipfs repo gc && # will remove /adir/file1 and /adir/file2 but not /adir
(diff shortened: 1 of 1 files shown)


whyrusleeping:
Sweet, LGTM!

@jbenet wanna take a look here?

test/sharness/t0252-files-gc.sh
```diff
@@ -42,4 +43,34 @@ test_expect_success "gc okay after adding incomplete node" '
...
+'
+
+test_expect_success "run gc and make sure directory contents are removed" '
+  ipfs repo gc &&
+  test_must_fail ipfs cat $HELLO_HASH
+'
+
+test_expect_success "add incomplete directory and make sure gc is okay" '
+  ipfs files cp /ipfs/$MYDIR_HASH /mydir &&
+  ipfs repo gc
```
jbenet: I would do `ipfs cat $HELLO_HASH` here too, just to be sure the thing is really gone and stays gone.

test/sharness/t0252-files-gc.sh
```diff
@@ -42,4 +43,34 @@ test_expect_success "gc okay after adding incomplete node" '
...
 
+MYDIR_HASH=QmRLrxQTdmLZNfDJxFQ9uFrgtj4s6cZU9GzTxCq2YLev24
+HELLO_HASH=QmeV1kwh3333bsnT6YRfdCRrSgUPngKmAhhTa4RrqYPbKT
+
+test_expect_success "add directory with direct pin" '
+  mkdir mydir/ &&
+  echo "hello world!" > mydir/hello.txt &&
+  ipfs add --pin=false -r mydir &&
+  ipfs pin add --recursive=false $MYDIR_HASH &&
+  ipfs cat $HELLO_HASH
```
jbenet: if something changes, the hashes may no longer match watch you got. if it's important that it matches, check it explicitly. if it is not important that it matches, grab the hashes directly from the output of `ipfs add` with something like:

```sh
HELLO_HASH=$(ipfs add --pin=false -q -r mydir/hello.txt | tail -n1)
MYDIR_HASH=$(ipfs add --pin=false -q -r mydir | tail -n1)
```

test/sharness/t0252-files-gc.sh
```diff
@@ -42,4 +43,34 @@ test_expect_success "gc okay after adding incomplete node" '
   ipfs refs $ADIR_HASH
 '
 
+MYDIR_HASH=QmRLrxQTdmLZNfDJxFQ9uFrgtj4s6cZU9GzTxCq2YLev24
+HELLO_HASH=QmeV1kwh3333bsnT6YRfdCRrSgUPngKmAhhTa4RrqYPbKT
```
jbenet: i would name these more usefully:
- `DIR_TO_STAY` or `DIR_PINNED`
- `FILE_TO_GO` or `FILE_UNPINNED`

test/sharness/t0252-files-gc.sh
```diff
@@ -11,13 +11,15 @@ test_description="test how the unix files api interacts with the gc"
...
-  echo "hello world"  | ipfs files write --create /hello.txt &&
+  echo "hello world" > hello.txt
+  cat hello.txt | ipfs files write --create /hello.txt &&
   ipfs repo gc &&
   ipfs cat QmVib14uvPnCP73XaCDpwugRuwfTsVbGyWbatHAmLSdZUS
 '
 
 test_expect_success "/hello.txt still accessible after gc" '
-  ipfs files read /hello.txt
+  ipfs files read /hello.txt > hello-actual
```
jbenet: need `&&` at the end

test/sharness/t0252-files-gc.sh
```diff
@@ -11,13 +11,15 @@ test_description="test how the unix files api interacts with the gc"
 test_init_ipfs
 
 test_expect_success "object not removed after gc" '
-  echo "hello world"  | ipfs files write --create /hello.txt &&
+  echo "hello world" > hello.txt
```
jbenet: need `&&` at the end

jbenet:
Some comments above

test/sharness/t0252-files-gc.sh
kevina: will fix

test/sharness/t0252-files-gc.sh
kevina: will fix

test/sharness/t0252-files-gc.sh
kevina: Will use DIR_PINNED and FILE_UNPINNED.

test/sharness/t0252-files-gc.sh
kevina: Will add.

test/sharness/t0252-files-gc.sh
kevina: Sounds like a good idea will do.

kevina:
@jbenet thanks, I just amended my commit with your corrections.

Sorry about the missing "&&", I actually meant to review this one more time before removing the "in progress" label.

whyrusleeping:
This LGTM. I would change a few stylistic things (like redirect the 'ipfs cat' calls either to dev/null, or check their output explicitly). But thats not a huge deal, and isnt really relevant to what we're testing here.

PR #3152, opened 2016-08-30
test: change time-out script to one from coreutils

Kubuxu:
I had a problem with finding definitive license statement of the previous
script. It was from examples in bash source tree but its licensing was a
bit unclear.

License: MIT


whyrusleeping:
Thanks @Kubuxu, This LGTM. @jbenet?

test/bin/time-out
```diff
@@ -1,83 +1,58 @@
...
-# $# should be at least 1 (the command to execute), however it may be strictly
-# greater than 1 if the command itself has options.
-if (($# == 0 || interval <= 0)); then
-    printUsage
-    exit 1
+#!/bin/sh
+
+# Execute a command with a timeout
+
+# License: LGPLv2
```
jbenet: Cant use LGPL.

jbenet:
> It was from examples in bash source tree but its licensing was a
bit unclear.

maybe point me to it?

test/bin/time-out
Kubuxu: As far as I know We can't link to LGPL, we can use it as a script.

Sharness is on GPLv2 and we are using it.

Kubuxu:
The previous script: http://www.bashcookbook.com/bashinfo/source/bash-4.0/examples/scripts/timeout3

jbenet:
Fun history: http://git.savannah.gnu.org/cgit/bash.git/log/examples/scripts/timeout3?h=bash-4.3-testing

yeah idk, that's sketchy. wish licensing wasnt an issue, but we have to be super careful about this.

jbenet:
This is fun, this code is all over the place: https://github.com/search?p=1&q=Dmitry+V+Golovashkin&ref=opensearch&type=Code

jbenet:
### TL;DR: @golovashkin could you please license your `timeout3` script written for bash in 2011 under the MIT License?

So to summarize, this script: http://git.savannah.gnu.org/cgit/bash.git/tree/examples/scripts/timeout3?h=bash-4.0 -- can be found all over the place (https://github.com/search?p=1&q=Dmitry+V+Golovashkin&ref=opensearch&type=Code), but always the license is unclear. People try to hand-wave around this by "giving full credit to"... and so on, which is not proper licensing, and then they bundle it in their software with other licenses (MIT, and so on). In fact, the FSF had it [removed from the bash source code](http://git.savannah.gnu.org/cgit/bash.git/log/examples/scripts/timeout3?h=bash-4.3-testing), probably because of unclear licensing.

It would be great for the world if Dmitry could just post this exact script and license it MIT, on github, or as a gist. @golovashkin doesn't seem to be a very active account, so i'm going to tweet at him (@golovashkin there too) to see if he could license it MIT for us.

whyrusleeping:
@Kubuxu could we not just do: `sleep 10 | COMMAND` ?

test/sharness/t0500-issues-and-regressions-offline.sh
```diff
@@ -8,17 +8,17 @@ test_description="Tests for various fixed issues and regressions."
 
 test_expect_success "ipfs init with occupied input works - #2748" '
 	export IPFS_PATH="ipfs_path"
-	echo "" | time-out ipfs init &&
+	echo "" | time-out 10 ipfs init &&
 	rm -rf ipfs_path
 '
 test_init_ipfs
 
 test_expect_success "ipfs cat --help succeeds with no input" '
```
jbenet: the text here is not right, it should be `"ipfs cat --help succeeds even if stdin remains open"` 

correct me if i'm wrong, but afaiu (from chatting with @whyrusleeping) this is the case.

test/bin/time-out
jbenet: @Kubuxu we are not embedding sharness in our repo, which we cannot do. We are downloading it to develop with, which is fine with licensing. check how the Makefile does it.

jbenet: > As far as I know we can't link to LGPL, we can use it as a script.

Can we embed it in our source? i don't think we can.

whyrusleeping:
@Kubuxu alright, lets take the easy way out. Write a simple go program to do this.

Kubuxu:
I changed the script to self written go program.
I also addressed the test name comment.

LGTY?

Kubuxu:
Fixed Makefile.

Kubuxu:
Oh, the 0.4.3-rcX is at Go 1.6.

Kubuxu:
I created: https://github.com/ipfs/go-ipfs/pull/3163 which should resolve the problem with Travis and Circle if we want update the RC to Go 1.7. I think that @whyrusleeping was planning that.

Kubuxu:
Rebased.

whyrusleeping:
LGTM, thanks @Kubuxu!

golovashkin:
Folks, you are welcomed to use the script in any way or shape or form, however you like it.
You are also very welcomed to modify it, add the MIT license to it, and resubmit it back to Bash / Chet Ramey.

My SAS email is no longer valid, please use
            golovashkin at gmail.com
(feel free to update the email in the script as well.)

Thank you!!
Best Regards,
DG

jbenet:
Thanks very much @golovashkin -- released it that way at https://github.com/jbenet/bash-timeout3 for posterity.

PR #3160, opened 2016-08-31
Doc/cmds/log notice unused

JesseWeinstein:
This removes the unused `notice` level from the help text, and also clarifies the distinction between the event log output and the other log output (at least, as best as *I* currently understand the distinction).

### DIFF
--- a/core/commands/log.go
+++ b/core/commands/log.go
@@ -35,8 +35,7 @@ var logLevelCmd = &cmds.Command{
 	Helptext: cmds.HelpText{
 		Tagline: "Change the logging level.",
 		ShortDescription: `
-'ipfs log level' is a utility command used to change the logging
-output of a running daemon.
+Change the verbosity of one or all subsystems log output. Does NOT affect the event log.
 `,
 	},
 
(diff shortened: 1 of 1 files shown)


whyrusleeping:
This LGTM, @Kubuxu @RichardLitt ?

Kubuxu:
LGTM

core/commands/log.go
```diff
@@ -35,8 +35,7 @@ var logLevelCmd = &cmds.Command{
 	Helptext: cmds.HelpText{
 		Tagline: "Change the logging level.",
 		ShortDescription: `
-'ipfs log level' is a utility command used to change the logging
-output of a running daemon.
+Change the verbosity of one or all subsystems log output. Does NOT affect the event log.
```
RichardLitt: I would change `NOT` to not. It says it without needing to draw attention to it. Also, `This does not...` is better, I think.

core/commands/log.go
```diff
@@ -91,9 +90,9 @@ subsystems of a running daemon.
 
 var logTailCmd = &cmds.Command{
 	Helptext: cmds.HelpText{
-		Tagline: "Read the logs.",
+		Tagline: "Read the event log.",
 		ShortDescription: `
-'ipfs log tail' is a utility command used to read log output as it is written.
+Outputs event log messages (NOT other log messages) as they are generated.
```
RichardLitt: Again, `NOT` to `not`.

JesseWeinstein:
Suggested changes applied.

RichardLitt:
LGTM.

PR #3162, opened 2016-08-31
 	blockstore: fix PutMany with cache logic

Kubuxu:
Thanks @whyrusleeping for noticing it.

Removed PutMany logic in bloom cache as it can't help with anything.
Fixed ARC cache to use filtered results instead of all blocks.

Also add test case for it.

### DIFF
--- a/blocks/blockstore/bloom_cache.go
+++ b/blocks/blockstore/bloom_cache.go
@@ -1,12 +1,13 @@
 package blockstore
 
 import (
+	"sync/atomic"
+
 	"github.com/ipfs/go-ipfs/blocks"
 	key "github.com/ipfs/go-ipfs/blocks/key"
+
 	bloom "gx/ipfs/QmWQ2SJisXwcCLsUXLwYCKSfyExXjFRW2WbBH5sqCUnwX5/bbloom"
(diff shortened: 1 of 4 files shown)


whyrusleeping:
your tests are broken and stuff

Kubuxu:
Hmm, interesting, they've run fine locally.

blocks/blockstore/bloom_cache.go
```diff
@@ -126,19 +127,11 @@ func (b *bloomcache) Put(bl blocks.Block) error {
...
-		for _, block := range bs {
-			b.bloom.AddTS([]byte(block.Key()))
-		}
-	}
-	return err
+	// bloom cache gives only conclusive resulty if key is not contained
+	// to reduce number of puts we need conclusive infomration if block is contained
+	// this means that PutMany can't be improved with bloom cache so we just
+	// just do a passthrough.
+	return b.blockstore.PutMany(bs)
```
Kubuxu: I have to add all block to bloom....

Will add a test case.

Kubuxu:
@whyrusleeping should be fixed, travis is queued up and previously it was having network problems.

whyrusleeping:
LGTM, thanks for fixing :)

PR #3111, opened 2016-08-23
bump go requirement to 1.7

whyrusleeping:
CI is gonna be fun with this one...

License: MIT
Signed-off-by: Jeromy <why@ipfs.io>

PR #3163, opened 2016-08-31
Update release candidate branch to Go 1.7

Kubuxu:
It replays: https://github.com/ipfs/go-ipfs/pull/3111 on version/0.4.3-rc4

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ os:
 language: go
 
 go:
-    - 1.5.2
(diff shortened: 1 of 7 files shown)


whyrusleeping:
why do we need this?

Kubuxu:
Because I don't go-timeout I wrote is using context, I can make it not use context but AFAIK we wanted to release 0.4.3 with Go 1.7 so it would be better to test it with it.

PR #2667, opened 2016-05-11
Added Default logic to `files` cmd

RichardLitt:
Part of #2484

License: MIT
Signed-off-by: Richard Littauer <richard.littauer@gmail.com>

PR #3167, opened 2016-09-01
Add default True to files flush

RichardLitt:
See https://github.com/ipfs/go-ipfs/pull/2667. 

License: MIT


whyrusleeping:
LGTM, thanks!

PR #3168, opened 2016-09-01
Add default 0 to files int opts

RichardLitt:
See https://github.com/ipfs/go-ipfs/pull/2667. 

License: MIT

### DIFF
--- a/core/commands/files/files.go
+++ b/core/commands/files/files.go
@@ -397,8 +397,8 @@ Examples:
 		cmds.StringArg("path", true, false, "Path to file to be read."),
 	},
 	Options: []cmds.Option{
-		cmds.IntOption("offset", "o", "Byte offset to begin reading from."),
-		cmds.IntOption("count", "n", "Maximum number of bytes to read."),
+		cmds.IntOption("n", "count", "Maximum number of bytes to read.").Default(-1),
+		cmds.IntOption("o", "offset", "Byte offset to begin reading from.").Default(0),
 	},
 	Run: func(req cmds.Request, res cmds.Response) {
 		n, err := req.InvocContext().GetNode()
@@ -461,16 +461,16 @@ Examples:
 		}
 
 		var r io.Reader = &contextReaderWrapper{R: rfd, ctx: req.Context()}
-		count, found, err := req.Option("count").Int()
+		count, _, err := req.Option("count").Int()
 		if err != nil {
 			res.SetError(err, cmds.ErrNormal)
 			return
 		}
-		if found {
-			if count < 0 {
-				res.SetError(fmt.Errorf("Cannot specify negative 'count'."), cmds.ErrNormal)
-				return
-			}
+		if count < 0 {
+			res.SetError(fmt.Errorf("Cannot specify negative 'count'."), cmds.ErrNormal)
+			return
+		}
+		if count > 0 {
 			r = io.LimitReader(r, int64(count))
 		}
 
@@ -565,10 +565,10 @@ stat' on the file or any of its ancestors.
 		cmds.FileArg("data", true, false, "Data to write.").EnableStdin(),
 	},
 	Options: []cmds.Option{
-		cmds.IntOption("offset", "o", "Byte offset to begin writing at."),
 		cmds.BoolOption("create", "e", "Create the file if it does not exist."),
 		cmds.BoolOption("truncate", "t", "Truncate the file to size zero before writing."),
-		cmds.IntOption("count", "n", "Maximum number of bytes to read."),
+		cmds.IntOption("n", "count", "Maximum number of bytes to read.").Default(-1),
+		cmds.IntOption("o", "offset", "Byte offset to begin writing at.").Default(0),
 	},
 	Run: func(req cmds.Request, res cmds.Response) {
 		path, err := checkPath(req.Arguments()[0])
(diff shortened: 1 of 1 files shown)


core/commands/files/files.go
```diff
@@ -403,8 +403,8 @@ Examples:
 		cmds.StringArg("path", true, false, "Path to file to be read."),
 	},
 	Options: []cmds.Option{
-		cmds.IntOption("offset", "o", "Byte offset to begin reading from."),
-		cmds.IntOption("count", "n", "Maximum number of bytes to read."),
+		cmds.IntOption("o", "offset", "Byte offset to begin reading from.").Default(0),
+		cmds.IntOption("n", "count", "Maximum number of bytes to read.").Default(0),
```
Kubuxu: AFAIK we settled on the first option name being the primary one/full.

core/commands/files/files.go
```diff
@@ -571,10 +571,10 @@ stat' on the file or any of its ancestors.
 		cmds.FileArg("data", true, false, "Data to write.").EnableStdin(),
 	},
 	Options: []cmds.Option{
-		cmds.IntOption("offset", "o", "Byte offset to begin writing at."),
 		cmds.BoolOption("create", "e", "Create the file if it does not exist."),
 		cmds.BoolOption("truncate", "t", "Truncate the file to size zero before writing."),
-		cmds.IntOption("count", "n", "Maximum number of bytes to read."),
+		cmds.IntOption("o", "offset", "Byte offset to begin writing at.").Default(0),
+		cmds.IntOption("n", "count", "Maximum number of bytes to read.").Default(0),
```
Kubuxu: same

kevina:
I'm not sure default values for either of these values makes sense.

I would also use -1 as a default for count to make it clear that it is a special value.  The test for the special value for count should be (count <= 0) in order to avoid changing existing behavior.

RichardLitt:
@kevina Perhaps I am going about this wrong. I added default values, partly because it helps show you how you can format your entries for those options. Perhaps I should instead add documentation, and not add a default value?

kevina:
@RichardLitt I think it is fairly obvious how to format these entries.  Are you trying to show that the value can be "0"?  @whyrusleeping @Kubuxu do you have any opinions on this mater?

RichardLitt:
@kevina I'm trying to show what the default is, because that informs the user more than just knowing it is an option. Is the default not `0`?

whyrusleeping:
@RichardLitt think about the options. The second one youre changing is defined as "the *maximum* number of bytes to read."

Setting the default value here to `0` would mean that unless we explicitly specify a count, we would never read anything.

RichardLitt:
Fair. Will edit.

RichardLitt:
Edited.

core/commands/files/files.go
```diff
@@ -618,12 +618,12 @@ stat' on the file or any of its ancestors.
...
 
-		count, countfound, err := req.Option("count").Int()
+		count, _, err := req.Option("count").Int()
 		if err != nil {
 			res.SetError(err, cmds.ErrNormal)
 			return
 		}
-		if countfound && count < 0 {
+		if count < 0 {
 			res.SetError(fmt.Errorf("cannot have negative byte count"), cmds.ErrNormal)
```
kevina: This is wrong.  Count can be -1.

core/commands/files/files.go
```diff
@@ -397,8 +397,8 @@ Examples:
 		cmds.StringArg("path", true, false, "Path to file to be read."),
 	},
 	Options: []cmds.Option{
-		cmds.IntOption("offset", "o", "Byte offset to begin reading from."),
-		cmds.IntOption("count", "n", "Maximum number of bytes to read."),
+		cmds.IntOption("n", "count", "Maximum number of bytes to read.").Default(-1),
+		cmds.IntOption("o", "offset", "Byte offset to begin reading from.").Default(0),
 	},
```
kevina: Why did you change it to make the short run first?  Didn't @Kubuxu say: "AFAIK we settled on the first option name being the primary one/full".

core/commands/files/files.go
```diff
@@ -565,10 +565,10 @@ stat' on the file or any of its ancestors.
 		cmds.FileArg("data", true, false, "Data to write.").EnableStdin(),
 	},
 	Options: []cmds.Option{
-		cmds.IntOption("offset", "o", "Byte offset to begin writing at."),
 		cmds.BoolOption("create", "e", "Create the file if it does not exist."),
 		cmds.BoolOption("truncate", "t", "Truncate the file to size zero before writing."),
-		cmds.IntOption("count", "n", "Maximum number of bytes to read."),
+		cmds.IntOption("n", "count", "Maximum number of bytes to read.").Default(-1),
+		cmds.IntOption("o", "offset", "Byte offset to begin writing at.").Default(0),
 	},
```
kevina: Again, why the change?

core/commands/files/files.go
```diff
@@ -461,16 +461,16 @@ Examples:
...
 			res.SetError(err, cmds.ErrNormal)
 			return
 		}
-		if found {
-			if count < 0 {
-				res.SetError(fmt.Errorf("Cannot specify negative 'count'."), cmds.ErrNormal)
-				return
-			}
+		if count < 0 {
+			res.SetError(fmt.Errorf("Cannot specify negative 'count'."), cmds.ErrNormal)
```
kevina: This is wrong.  Count can be -1.

whyrusleeping: @kevina how can it be -1?

whyrusleeping: Ah I see, we're using that as a sentinel for 'no limit'

whyrusleeping:
I think i prefer the old behaviour where we know whether or not the user specified a value for count (as opposed to using -1 as a special value)

whyrusleeping:
This has taken up a lot of review time and isnt a necessary change. Closing.

kevina:
@whyrusleeping, thanks I agree on closing this issue.

RichardLitt:
Sounds good.

PR #3169, opened 2016-09-01
bitswap: Don't clear 'active' until Connect calls are finished

whyrusleeping:
This has been an issue for a long time, but for some reason the changes to making providers more efficient has made this a much larger problem.

I'm not 100% sure this will solve the excess cpu usage problem we're seeing on the gateways, but its definitely fixing a bug.

License: MIT


Kubuxu:
LGTM.

jbenet:
LGTM

PR #3170, opened 2016-09-01
sharness: fix typo in hash in t0050

whyrusleeping:
was integrating CIDv0 into the codebase and flushed out this bug

License: MIT


Kubuxu:
LGTM, does it mean there is some other issue with mulihash or base58? It should have probably refused to work.

PR #2997, opened 2016-07-27
assets: add security@ipfs.io GPG key

lgierth:
I'll see if I have to update hashes in the tests...

Resolves #3256

License: MIT
Signed-off-by: Lars Gierth larsg@systemli.org

PR #3184, opened 2016-09-03
Fix whitespace discrepency in security-notes

lytedev:
Just remove an extra space that found its way into one of the default text files. It bothered me, but I know it's a very minor thing - if it matters at all, even.


lgierth:
Thanks! This is blocked by #2997

whyrusleeping:
@lgierth assigning you here just to make sure we don't lose track of things

whyrusleeping:
@lgierth could you take care of this?

lgierth:
Hey @lytedev, I have another PR coming which also covers the security-notes, and I took care of the whitespace there. Thank you! :+1:

PR #3185, opened 2016-09-04
Extract dht, kbucket, routing, record, testutils, etc. [work-in-progress]

geoah:
This is a direct continuation of the work done to separate the dht package. It replaces various packages with external ones.

- github.com/libp2p/go-libp2p-kad-dht
- github.com/libp2p/go-libp2p-routing
- github.com/libp2p/go-libp2p-kbucket
- github.com/libp2p/go-libp2p-record
- github.com/libp2p/go-testutil
- github.com/ipfs/go-key
- github.com/ipfs/go-todocounter
- github.com/ipfs/go-libp2p-loggables
- ... probably a couple more

Once libp2p/go-libp2p-kad-dht#6 is merged, there a couple more stuff to do and this should be merge-able as well.

- [x] Merge ipfs/go-ds-measure#1
- [x] Figure out what to do about `ThreadSafeCloserMapDatastore`
- [x] A lot of pinning


thirdparty/datastore2/datastore_closer.go
```diff
@@ -33,3 +34,8 @@ func (w *datastoreCloserWrapper) Batch() (datastore.Batch, error) {
 
 	return bds.Batch()
 }
+
+// TODO(geoah) Move to ./testutils or somewhere else
+func ThreadSafeCloserMapDatastore() ThreadSafeDatastoreCloser {
+	return CloserWrap(syncds.MutexWrap(datastore.NewMapDatastore()))
+}
```
geoah: This is the `ThreadSafeCloserMapDatastore` I'm talking about.
Originally this was under `thirdparty/testutils/datastore.go` but as there was no other reason for `go-testutils` to depend on `go-datastore` it was removed.

I have temporarily put it under `thirdparty/datastore2` but I'm open to suggestions.

whyrusleeping: Thats fine with me. The more we break things out the easier it will become to find homes for these sorts of things.

whyrusleeping:
Looks like we need to delete blocks/key, routing/{dht,kbucket,record} and maybe a few others too, and then we should actually use these new extracted packages :package:

geoah:
@whyrusleeping not sure why the packages were still there.
The external ones were already being used instead of the local ones, I just didn't remove them for some reason.

I have now actually removed the packages. I think I got them all?

geoah:
Closing this in favour or more, smaller PRs.

PR #3187, opened 2016-09-04
Rework go-ipfs to use content IDs version 0

whyrusleeping:
This is a refactor to replace all object identifiers code with content ID version 0. This is a drop in replacement for the existing key.Key and will not change any outward behaviour. This is just a preparatory refactor for the switchover to CIDv1 and ipld

### DIFF
--- a/pin/pin.go
+++ b/pin/pin.go
@@ -4,22 +4,33 @@ package pin
 
 import (
 	"fmt"
+	"os"
 	"sync"
 	"time"
 
 	key "github.com/ipfs/go-ipfs/blocks/key"
-	"github.com/ipfs/go-ipfs/blocks/set"
 	mdag "github.com/ipfs/go-ipfs/merkledag"
+
 	ds "gx/ipfs/QmNgqJarToRiq2GBaPJhkmW4B5BxS5B74E1rkGvv2JoaTp/go-datastore"
 	logging "gx/ipfs/QmSpJByNKFX1sCsHBEp3R73FL4NF6FnQTEGyNAXHm2GS52/go-log"
 	context "gx/ipfs/QmZy2y8t9zQH2a1b8q2ZSLKp17ATuJoCNxxyMFG5qFExpt/go-net/context"
+	cid "gx/ipfs/QmfSc2xehWmWLnwwYR91Y8QF4xdASypTFVknutoKQS3GHp/go-cid"
 )
 
 var log = logging.Logger("pin")
 
 var pinDatastoreKey = ds.NewKey("/local/pins")
 
-var emptyKey = key.B58KeyDecode("QmdfTbBqBPQ7VNxZEYEj14VmRuZBkqFbiwReogJgS1zR1n")
+var emptyKey *cid.Cid
+
+func init() {
+	e, err := cid.Decode("QmdfTbBqBPQ7VNxZEYEj14VmRuZBkqFbiwReogJgS1zR1n")
+	if err != nil {
+		log.Error("failed to decode empty key constant")
+		os.Exit(1)
+	}
+	emptyKey = e
+}
 
 const (
 	linkRecursive = "recursive"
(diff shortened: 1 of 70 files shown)


whyrusleeping:
Note: this is not quite complete yet

blocks/blockstore/bloom_cache.go
```diff
@@ -132,10 +132,11 @@ func (b *bloomcache) PutMany(bs []blocks.Block) error {
 			good = append(good, block)
 		}
 	}
-	err := b.blockstore.PutMany(bs)
+
+	err := b.blockstore.PutMany(good)
 	if err == nil {
-		for _, block := range bs {
-			b.bloom.AddTS([]byte(block.Key()))
+		for _, blk := range good {
```
Kubuxu: This is resolved in https://github.com/ipfs/go-ipfs/pull/3162

blocks/blockstore/arc_cache.go
```diff
@@ -99,12 +99,14 @@ func (b *arccache) PutMany(bs []blocks.Block) error {
 			good = append(good, block)
 		}
 	}
-	err := b.blockstore.PutMany(bs)
+
+	err := b.blockstore.PutMany(good)
```
Kubuxu: This too.

Kubuxu:
I went over it, can't find anything wrong with it at first sight but I would love if @lgierth went over it too.


(also it is in conflict with master now, probably due to changes in the blockstore code I pointed out).

whyrusleeping:
One change i could probably do separately is the `Object` abstraction in `Blockservice`. It doesnt *need* to be done here, but it does simplify things a bit I think. If anyone thinks that it should be done separately I will go ahead and do that.

Other than that, I think that the pin code changes should be looked over closely. @lgierth if you could pay special attention there that would be great :)

lgierth:
> This is a refactor to replace all object identifiers code with content ID version 0.

Is the following accurate? It doesn't replace *all* usage of key.Key -- from the block layer on inwards we assume raw data (where s/Data/RawData). The layer above that will be the one dealing with the different semantics of various CIDs (soon).

I'm looking over `git diff pin/` with this assumption in mind, mh, tricky.

---

```go
// an Object is simply a typed block
type Object interface {
	Cid() *cid.Cid
	blocks.Block
}
```

:+1:

whyrusleeping:
@lgierth yeah, youre right. the `key.Key` is still around at the block layer. We're talking about making the CIDs go all the way down to disk, it has its pros and cons, but it would drastically increase the complexity of this change, and is definitely something we can do in a follow up PR.

pin/pin.go
```diff
@@ -259,90 +269,99 @@ func (p *pinner) isPinnedWithType(k key.Key, mode PinMode) (string, bool, error)
...
-	checkChildren = func(rk key.Key, parentKey key.Key) error {
+	var checkChildren func(*cid.Cid, *cid.Cid) error
+	checkChildren = func(rk, parentKey *cid.Cid) error {
 		parent, err := p.dserv.Get(context.Background(), parentKey)
 		if err != nil {
 			return err
 		}
 		for _, lnk := range parent.Links {
-			k := key.Key(lnk.Hash)
+			c := cid.NewCidV0(lnk.Hash)
```
lgierth: I assume this is explicitly v0 because we don't wanna start using v1 right now already

pin/pin.go
```diff
@@ -259,90 +269,99 @@ func (p *pinner) isPinnedWithType(k key.Key, mode PinMode) (string, bool, error)
...
 	}
 }
 
+func cidSetWithValues(cids []*cid.Cid) *cid.Set {
+	out := cid.NewSet()
+	for _, c := range cids {
+		out.Add(c)
+	}
+	return out
+}
```
lgierth: Is this a candidate for moving to go-cid?

lgierth:
LGTM :+1: 

- `pin/` could see the functions named after Key (e.g. `Pinner.DirectKeys()`) renamed
- some of `pin/` still uses Key, which is mostly the GC code -- I can understand why you didn't wanna touch it *right now*, it can get messy and we can take care of it later.
- these two made it a bit confusing at first, but I think it's good to go

pin/pin.go
whyrusleeping: Yeah, but thinking about it, we could change that to just be a `cid.Cast` and it would work just fine

pin/pin.go
whyrusleeping: Ah, yeah. thats a great point

pin/pin.go
lgierth: Nah it's fine, one silly mistake and we'll have to maintain it :D:D

lgierth:
And I guess this PR has the nice side-effect of reducing the overall dependency on the blocks package :)

pin/pin.go
jbenet: did this get moved? dont see commits after

PR #3189, opened 2016-09-05
metrics: introduce go-metrics-interface

Kubuxu:
It makes introducing metrics much easier.

The interface itself doesn't depend on prometheus, that means it is very light and without it (or other implementation) acts as a general noop.

Importing go-metrics-prometheus injects prometheus implementation.

Usage:
There are three most improtant functions in go-metrics-interface:

* `metrics.New(name, helptext)` - creates new metrics with given name and helptext
* `metrics.NewCtx(ctx, name, helptext)` - creates new metrics with given name and helptext in scope defined in the context
* `metric.CtxSubScope(ctx, name)` - returns context that is a subscope of scope defined in current ctx

As you can see it uses `context.Value` for passing the scope name of the metrics. In my opinion it is one of the better solutions as context is already required in many places, and it introduction in constructors chains isn't bad either.

Current problem I have is that `goprocess` doesn't have something like `context.Value` so passing scopes through `goprocess` is something I haven't figured out yet but it can be done later.

//cc @whyrusleeping @lgierth

### DIFF
--- a/blocks/blockstore/arc_cache.go
+++ b/blocks/blockstore/arc_cache.go
@@ -19,7 +19,7 @@ type arccache struct {
 	total metrics.Counter
 }
 
-func newARCCachedBS(bs Blockstore, ctx context.Context, lruSize int) (*arccache, error) {
+func newARCCachedBS(ctx context.Context, bs Blockstore, lruSize int) (*arccache, error) {
 	arc, err := lru.NewARC(lruSize)
 	if err != nil {
 		return nil, err
--- a/blocks/blockstore/arc_cache.go
+++ b/blocks/blockstore/arc_cache.go
@@ -1,9 +1,11 @@
 package blockstore
 
 import (
-	"github.com/ipfs/go-ipfs/blocks"
 	key "gx/ipfs/Qmce4Y4zg3sYr7xKM5UueS67vhNni6EeWgCRnb7MbLJMew/go-key"
 
+	"github.com/ipfs/go-ipfs/blocks"
+
+	"gx/ipfs/QmRg1gKTHzc3CZXSKzem8aR4E3TubFhbgXwfVuWnSK5CC5/go-metrics-interface"
 	lru "gx/ipfs/QmVYxfoJQiZijTgPNHCHgHELvQpbsJNTg6Crmc3dQkj3yy/golang-lru"
 	context "gx/ipfs/QmZy2y8t9zQH2a1b8q2ZSLKp17ATuJoCNxxyMFG5qFExpt/go-net/context"
 	ds "gx/ipfs/QmbzuUusHqaLLoNTDEVLcSF6vZDHZDLPC7p4bztRvvkXxU/go-datastore"
@@ -12,15 +14,21 @@ import (
 type arccache struct {
 	arc        *lru.ARCCache
 	blockstore Blockstore
+
+	hits  metrics.Counter
+	total metrics.Counter
 }
 
-func arcCached(bs Blockstore, lruSize int) (*arccache, error) {
+func newARCCachedBS(bs Blockstore, ctx context.Context, lruSize int) (*arccache, error) {
 	arc, err := lru.NewARC(lruSize)
 	if err != nil {
 		return nil, err
 	}
+	c := &arccache{arc: arc, blockstore: bs}
+	c.hits = metrics.NewCtx(ctx, "arc.hits_total", "Number of ARC cache hits").Counter()
+	c.total = metrics.NewCtx(ctx, "arc_total", "Total number of ARC cache requests").Counter()
 
-	return &arccache{arc: arc, blockstore: bs}, nil
+	return c, nil
 }
 
 func (b *arccache) DeleteBlock(k key.Key) error {
@@ -42,6 +50,7 @@ func (b *arccache) DeleteBlock(k key.Key) error {
 // if ok == false has is inconclusive
 // if ok == true then has respons to question: is it contained
 func (b *arccache) hasCached(k key.Key) (has bool, ok bool) {
+	b.total.Inc()
 	if k == "" {
 		// Return cache invalid so the call to blockstore happens
 		// in case of invalid key and correct error is created.
@@ -50,6 +59,7 @@ func (b *arccache) hasCached(k key.Key) (has bool, ok bool) {
 
 	h, ok := b.arc.Get(k)
 	if ok {
+		b.hits.Inc()
 		return h.(bool), true
 	}
 	return false, false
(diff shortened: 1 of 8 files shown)


blocks/blockstore/arc_cache.go
```diff
@@ -3,23 +3,31 @@ package blockstore
...
 type arccache struct {
 	arc        *lru.ARCCache
 	blockstore Blockstore
+
+	hits  metrics.Counter
+	total metrics.Counter
 }
 
-func arcCached(bs Blockstore, lruSize int) (*arccache, error) {
+func newARCCachedBS(bs Blockstore, ctx context.Context, lruSize int) (*arccache, error) {
```
whyrusleeping: convention is to place the context first, i think that even though the blockstore is the primary object we should still respect that.

blocks/blockstore/bloom_cache.go
```diff
@@ -17,9 +20,29 @@ func bloomCached(bs Blockstore, ctx context.Context, bloomSize, hashCount int) (
...
+	bc.total = metrics.NewCtx(ctx, "bloom_total",
+		"Total number of requests to bloom cache").Counter()
+
+	fill := metrics.NewCtx(ctx, "bloom_fill_ratio",
+		"Ratio of bloom filter fullnes, (updated once a minute)").Gauge()
+
 	bc.Invalidate()
 	go bc.Rebuild(ctx)
-
+	go func() {
```
whyrusleeping: If there a quick way to check if we even have metrics enabled? It would be nice if metrics were turned off, this goroutine wouldnt exist.

Kubuxu: Yeah, had the same idea, I will add method to go-metrics-interface to check if there was some impl injected.

whyrusleeping: might even be cool to wrap this sort of behaviour in a `metrics.Periodic` type method.

whyrusleeping:
This LGTM. I'm definitely on board with doing things this way.

Some notes for future metrics to add (maybe i should make this an issue?)

- Bitswap wantlist size
- Number of stored provider records
- Number of each dht rpc being made
- Number of diagnostics service messages handled

Kubuxu:
> Number of each dht rpc being made

This option could really use other feature of prometheus: labels, but for that I will have to figure out how to wrap it nicely first.

whyrusleeping:
Alright, so i'm gonna merge this. But since there are a few tweaks we want to make to the interface, lets hold off on pushing it into packages outside this repo (to avoid the pain of massive amounts of updating)

Kubuxu:
Wait a sec.

Kubuxu:
Updated to not run bloom fillrate collector when metrics aren't active.

blocks/blockstore/bloom_cache.go
```diff
@@ -25,24 +25,26 @@ func bloomCached(bs Blockstore, ctx context.Context, bloomSize, hashCount int) (
...
-		<-bc.rebuildChan
-		t := time.NewTicker(1 * time.Minute)
-		for {
-			select {
-			case <-ctx.Done():
-				t.Stop()
-				return
-			case <-t.C:
-				fill.Set(bc.bloom.FillRatio())
+	if metrics.Active() {
```
whyrusleeping: Is this some sort of global? That feels very wrong to me. Why arent we checking if the context has metrics injected into it and using that as a way to tell if we have metrics enabled?

Kubuxu: Context is used to define full name of the metrics.

This checks if there is metrics collector injected into go-metrics-interface.

whyrusleeping: ooooh.... hrm.. I misunderstood how this all worked then. I was under the impression we were avoiding globals entirely by using the contexts to pass metrics down through to where they would be collected.

whyrusleeping: This is still okay, Its just less testable and composable than i was hoping.

Kubuxu: We could do that, but then we won't be able to use metrics from libp2p as the goprocess doesn't support `WithValue`/`Value`.

cmd/ipfs/daemon.go
```diff
@@ -359,8 +360,7 @@ func daemonFunc(req cmds.Request, res cmds.Response) {
 	}
 
 	// initialize metrics collector
-	prometheus.MustRegisterOrGet(&corehttp.IpfsNodeCollector{Node: node})
-	prometheus.EnableCollectChecks(true)
```
whyrusleeping: why was enableCollectChecks removed?

whyrusleeping:
Where do we actually fill the interface? I see the old prometheus stuff... but i think i'm missing where we actually add prometheus as a collector for your metrics package

blocks/blockstore/bloom_cache.go
jbenet: - in general: don't use statics. If you run multiple IPFS daemons, the
metrics may clash. If the lib FORCES you to use a global, then add a
guaranteed-unique prefix. Ideally with a local variable that internally
uses the global, threaded through with the context.
- but I don't care enough about this case to force it, if it's way easier
and safe to have thousands of IPFS daemons in the same process, and nothing
will clash, fine.

cmd/ipfs/daemon.go
Kubuxu: It was removed from API and enabled by default.

Kubuxu:
https://github.com/ipfs/go-ipfs/pull/3189/files#diff-54f2c4a5481de30b8ccb59ad5b6588b3R31 I am using `init()` method of go-metrics-prometheus package.

whyrusleeping:
@Kubuxu i'm not sure i'm comfortable with injecting things that way... It should be an explicit thing we do in an init function of whatever code we import it from.

also need rebasing here

Kubuxu:
The single purpose of go-metrics-prometheus is to be injected into go-metrics-interface, this pacakge doesn't export any other symbols.

I will change it.

Kubuxu:
That was one hello of a rebase.

We really need to do something with this.

whyrusleeping:
I'm not super on board with the way this does metrics... but since it is an improvement on the current situation, we can merge it in.

Some action items to move forward after this:

- fix goprocess to allow context Values to be passed through to derived contexts
- switch away from using a 'global' metrics registration and inject metrics collection into the top level metrics.
- Add flags/config vars to disable/enable metrics collection
- Add more metrics!

Kubuxu:
Rebased it once more

whyrusleeping:
Alright, lets get this in. Thanks @Kubuxu! Better metrics for everyone!

PR #3192, opened 2016-09-06
bitswap: search for wantlist providers a little less often

whyrusleeping:
ref: https://github.com/ipfs/go-ipfs/pull/3182#issuecomment-244826036

This solution is strictly better than that of the current release. The current 'intended' behaviour is simply way too spammy, We can develop more intelligent solutions as we move forward.

License: MIT

### DIFF
--- a/exchange/bitswap/workers.go
+++ b/exchange/bitswap/workers.go
@@ -1,6 +1,7 @@
 package bitswap
 
 import (
+	"math/rand"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


Kubuxu:
In this PR we are both switching to sending only one entry from the wantlist and reducing the frequency of that send. Doesn't it have big chance of sending the entries to rarely this time?

whyrusleeping:
@Kubuxu when compared to what was in the release previously, where we only sent the last key in each wantlist, this is better. I could be convinced to keep the interval the same though.

Kubuxu:
LGTM, this restores behaviour that was happening all along.

Let's think about improving it in future but on 0.4.4.

jbenet:
- To review, the intended behavior is to make sending diffs be CORRECT, and
not rely on full wantlist updates for that.
- We can ease into this by reducing the frequency of wantlist updates, and
logging when we receive a wantlist that differs from our own expected view
of it. I.e. We should make it so that wantlist updates are always
superfluous.
- Given that the transport is reliable, and connections breaking trigger a
new reconnection and new full want list, this will have the indented effect.

PR #2936, opened 2016-07-02
Improve the install.sh script

mkg20001:
This script now looks for all available locations in $PATH and chooses the best

License: MIT
Signed-off-by: mkg20001 mkg20001@gmail.com

PR #2504, opened 2016-03-25
install: Improve linux ipfs install script

mrshu:
- Improve linux ipfs install script by trying to install the ipfs binary
  to every directory in $PATH, starting with `/usr/local/bin` and
  `/usr/bin` which were used in the previous version of this script..

Signed-off-by: mr.Shu mr@shu.io
---

PR #3194, opened 2016-09-06
Refactoring install.sh

kulla:
Hello,

I made little changes for the installation script `install.sh`. The enhancements are:

* I use a for loop for the binpaths (following the [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) principle)
* An error message was added in the end to inform the user when the script couldn't be installed.

### DIFF
--- a/cmd/ipfs/dist/install.sh
+++ b/cmd/ipfs/dist/install.sh
@@ -9,7 +9,7 @@ binpaths="/usr/local/bin /usr/bin"
 
 for binpath in $binpaths; do
   if mv -t "$binpath" "$bin" 2> /dev/null; then
-    echo "installed $binpath/$bin"
+    echo "Moved $bin to $binpath"
     exit 0
   fi
 done
--- a/cmd/ipfs/dist/install.sh
+++ b/cmd/ipfs/dist/install.sh
@@ -11,7 +11,7 @@ binpaths="/usr/local/bin /usr/bin"
 is_write_perm_missing=""
 
 for binpath in $binpaths; do
-  if mv -t "$binpath" "$bin" 2> /dev/null; then
+  if mv "$bin" "$binpath/$bin" 2> /dev/null; then
     echo "Moved $bin to $binpath"
     exit 0
   else
--- a/cmd/ipfs/dist/install.sh
+++ b/cmd/ipfs/dist/install.sh
@@ -1,4 +1,7 @@
 #!/bin/sh
+#
+# Installation script for ipfs. It tries to move $bin in one of the
+# directories stored in $binpaths.
 
 bin=ipfs
 binpaths="/usr/local/bin /usr/bin"
@@ -6,10 +9,6 @@ binpaths="/usr/local/bin /usr/bin"
 # This variable contains a nonzero length string in case the script fails
 # because of missing write permissions.
 is_write_perm_missing=""
-
-# this script is currently brain dead.
-# it merely tries two locations.
-# in the future maybe use value of $PATH.
 
 for binpath in $binpaths; do
   if mv -t "$binpath" "$bin" 2> /dev/null; then
--- a/cmd/ipfs/dist/install.sh
+++ b/cmd/ipfs/dist/install.sh
@@ -1,15 +1,19 @@
 #!/bin/sh
 
 bin=ipfs
+binpaths="/usr/local/bin /usr/bin"
 
 # this script is currently brain dead.
 # it merely tries two locations.
 # in the future maybe use value of $PATH.
 
-for binpath in /usr/local/bin /usr/bin; do
+for binpath in $binpaths; do
   if [ -d "$binpath" ]; then
     mv "$bin" "$binpath/$bin"
     echo "installed $binpath/$bin"
     exit 0
   fi
 done
+
+echo "cannot install $bin in one of the directories $binpaths"
+exit 1
--- a/cmd/ipfs/dist/install.sh
+++ b/cmd/ipfs/dist/install.sh
@@ -3,6 +3,10 @@
 bin=ipfs
 binpaths="/usr/local/bin /usr/bin"
 
+# This variable contains a nonzero length string in case the script fails
+# because of missing write permissions.
+is_write_perm_missing=""
+
 # this script is currently brain dead.
 # it merely tries two locations.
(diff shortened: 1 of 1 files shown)


cmd/ipfs/dist/install.sh
```diff
@@ -1,21 +1,19 @@
...
 
-binpath=/usr/local/bin
-if [ -d "$binpath" ]; then
-  mv "$bin" "$binpath/$bin"
-  echo "installed $binpath/$bin"
-  exit 0
-fi
+for binpath in $binpaths; do
+  if [ -d "$binpath" ]; then
+    mv "$bin" "$binpath/$bin"
```
whyrusleeping: We should check if this succeeds or not and not print 'installed X' if it fails.

cmd/ipfs/dist/install.sh
```diff
@@ -1,21 +1,19 @@
...
 # in the future maybe use value of $PATH.
 
-binpath=/usr/local/bin
-if [ -d "$binpath" ]; then
-  mv "$bin" "$binpath/$bin"
-  echo "installed $binpath/$bin"
-  exit 0
-fi
+for binpath in $binpaths; do
+  if [ -d "$binpath" ]; then
```
whyrusleeping: would be nice to also check `[ -w "$binpath" ]` to see if we're allowed to write in the given directory.

kulla:
@whyrusleeping Thanks for your comments! I moved the `mv' command into the if condition so that the script only exits successfully when one of the `mv' commands succeeds. There is also no need to check whether `$binpath' exists or is writable beforehand because this is done by the `mv' command itself (if for example there is no `$binpath' directory, the `mv' command will exit with an non zero status code and the body of the if statement is not executed). So also rare cases are also handled (for example when `$binpath/$bin' is an existing directory).

I also specified the target directory in the `mv' command explicitly. So `$bin' will not be moved when  `$binpath/$bin' is a directory.

*Sitenote:* Shall `cp' be used instead of `mv'?!

whyrusleeping:
I think youre right and `cp` should probably be used, Its kinda weird that the binary i downloaded would disappear after running 'install'.

I wonder how best to convey to the user that they should try running the install with sudo. This script will fail on pretty much all systems the user might run it on as the directories we try never have user level write permissions (for good reasons).

Kubuxu:
We have two PRs like that already open: https://github.com/ipfs/go-ipfs/pull/2936 and https://github.com/ipfs/go-ipfs/pull/2504.

@whyrusleeping I think it would be best to just join them together, if @kulla wants to do it.
Both of them include valuable notes and comments.

kulla:
Yes, it would be the best to join the PRs together. When I'll have time I will have a look on the other PRs and try to contribute there...

PS: `mv` is changed with `cp`.

kulla:
I made comments in [#2504](https://github.com/ipfs/go-ipfs/pull/2504). Everything else what I have added in this PR is already implemented in [#2504](https://github.com/ipfs/go-ipfs/pull/2504). So it's okay for me to close the discussion and the PR here to continue working in [#2504](https://github.com/ipfs/go-ipfs/pull/2504).

Kubuxu:
It is @whyrusleeping 's call.

Also I think #2504 is abandoned so if you would like to take it over and finish it, it would be great but let's wait for @whyrusleeping.

kulla:
I have a more general question: Is it really necessary to check all directories in `$PATH`?! Concerning [KISS](https://en.wikipedia.org/wiki/KISS_principle) we shall only do this, when this is really necessary. I mean for most users the directory `/usr/local/bin` exists and will be used as the target directory for this script so that there is no difference with the current version of [install.sh](https://github.com/ipfs/go-ipfs/blob/master/cmd/ipfs/dist/install.sh). Are there any unix systems without `/usr/local/bin`?! What are your opinions?

Kubuxu:
Arch for example doesn't have `/usr/local/bin` by default but most people will add it.

kulla:
@Kubuxu I moved my comment to https://github.com/ipfs/go-ipfs/pull/2504#issuecomment-245330691 :smile:

whyrusleeping:
I think this PR is on the right path. The others are very complicated, Simplicity is key here.

jbenet:
Would be nice to detect the write perms fail and ask user to use `sudo`. this will catch a big fail for a lot of users that don't live in the command line.

kulla:
@jbenet Implementation of error message recommending `sudo` (if write permissions are missing) is done. Following https://github.com/ipfs/go-ipfs/pull/2504#discussion_r77872437 I changed `cp` with `mv`. I also included https://github.com/ipfs/go-ipfs/pull/2504#discussion_r77872959 (using of `Moved ./ipfs to /usr/local/bin/ipfs` as success message).

whyrusleeping:
@kulla just to double check, the `-t` flag on `mv` works the same on both gnu and bsd tools? (we've had issues with this in the past)

kulla:
Yeah. you're right. Concerning https://www.freebsd.org/cgi/man.cgi?query=mv&sektion=1 there is no `-t` parameter in bsd's `mv` command... I'll change the move command...

jbenet:
From other thread: use `mv -t "$bin" "$binpath"`

whyrusleeping:
This LGTM

@jbenet cant use `-t`, its not cross platform.

jbenet:
@whyrusleeping yep, saw thanks. LGTM

(from my information reference frame, my comment happened before the others, but it was stuck in an email client for hours before reaching the rest of the world.)

cmd/ipfs/dist/install.sh
```diff
@@ -1,21 +1,34 @@
...
-binpath=/usr/bin
-if [ -d "$binpath" ]; then
-  mv "$bin" "$binpath/$bin"
-  echo "installed $binpath/$bin"
-  exit 0
+if [ -n "$is_write_perm_missing" ]; then
+  echo "It seems that we do not have the necessary write permissions."
+  echo "Perhaps try running this script as a privileged user:"
+  echo
+  echo "    sudo $0"
```
jbenet: yay, good explanation 👍

whyrusleeping:
@kulla thank you so much! This has needed a nice clean update for a while.

kulla:
Thanks to all for your comments! :smile:

PR #3196, opened 2016-09-06
blockstore: change unit of bloom filter to byte from bits

Kubuxu:
License: MIT

### DIFF
--- a/blocks/blockstore/caching.go
+++ b/blocks/blockstore/caching.go
@@ -8,16 +8,16 @@ import (
 
 // Next to each option is it aproximate memory usage per unit
 type CacheOpts struct {
-	HasBloomFilterSize   int // 1 bit
+	HasBloomFilterSize   int // 1 byte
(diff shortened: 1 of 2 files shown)


blocks/blockstore/caching.go
```diff
@@ -8,14 +8,14 @@ import (
...
 type CacheOpts struct {
-	HasBloomFilterSize   int // 1 bit
+	HasBloomFilterSize   int // 1 byte
 	HasBloomFilterHashes int // No size, 7 is usually best, consult bloom papers
 	HasARCCacheSize      int // 32 bytes
 }
 
 func DefaultCacheOpts() CacheOpts {
 	return CacheOpts{
-		HasBloomFilterSize:   512 * 8 * 1024,
```
whyrusleeping: I prefer using shift notation for these: `512 << 10`

whyrusleeping:
One small nitpick, doesnt need to be addressed, LGTM

Kubuxu:
Addressed.

whyrusleeping:
Thanks!

PR #3198, opened 2016-09-09
Extract blocks/key to ipfs/go-key

geoah:
This is the first PR in a series trying to split ipfs/go-ipfs#3185 into more manageable chunks.
In addition to extracting go-key, it updates to latest go-datastore, autobatch, retry-datastore, and failstore.

@whyrusleeping if you still feel this is un-reviewable just let me know and I'll figure something out.


whyrusleeping:
This LGTM, thanks :)

Its harder to review PRs that have both dependency and code changes. This being simply dep changes makes it pretty easy

PR #3201, opened 2016-09-10
README.md: Add sudo to install commands.

kulla:
A little change to `README.md` since most users will need `sudo` to run the installation script sucessfully.


whyrusleeping:
LGTM, thanks!

PR #3206, opened 2016-09-11
gateway: fix --writable flag :|

lgierth:
Broke it in #2874... my unproven assumption apparantly was that `cfg.Gateway.Writable = true` would somehow magically make it into other instances of the config retrieved via `Repo.Config()`...

We're not overloading the repo config with this anymore now. We also have a test for `--writable` now.

### DIFF
--- a/cmd/ipfs/daemon.go
+++ b/cmd/ipfs/daemon.go
@@ -409,9 +409,9 @@ func serveHTTPApi(req cmds.Request) (error, <-chan error) {
 	if err != nil {
 		return fmt.Errorf("serveHTTPApi: Option(%s) failed: %s", unrestrictedApiAccessKwd, err), nil
 	}
-	gatewayOpt := corehttp.GatewayOption(corehttp.WebUIPaths...)
+	gatewayOpt := corehttp.GatewayOption(false, corehttp.WebUIPaths...)
 	if unrestricted {
-		gatewayOpt = corehttp.GatewayOption("/ipfs", "/ipns")
+		gatewayOpt = corehttp.GatewayOption(true, "/ipfs", "/ipns")
 	}
 
 	var opts = []corehttp.ServeOption{
@@ -480,8 +480,8 @@ func serveHTTPGateway(req cmds.Request) (error, <-chan error) {
 	if err != nil {
 		return fmt.Errorf("serveHTTPGateway: req.Option(%s) failed: %s", writableKwd, err), nil
 	}
-	if writableOptionFound {
-		cfg.Gateway.Writable = writable
+	if !writableOptionFound {
+		writable = cfg.Gateway.Writable
 	}
 
 	gwLis, err := manet.Listen(gatewayMaddr)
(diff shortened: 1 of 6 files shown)


lgierth:
Added a test for `--writable=false` too

lgierth:
We can defer this to 0.4.4 since the workaround is to simply set Writable=true in the config. I imagine `--writable` was introduced as a shorthand for `ipfs config --bool Gateway.Writable true`.

test/sharness/t0111-gateway-writeable.sh
```diff
@@ -9,6 +9,22 @@ test_description="Test HTTP Gateway (Writable)"
...
+test_launch_ipfs_daemon --writable
+test_expect_success "ipfs daemon --writable overrides config" '
+  curl -v -X POST http://$GWAY_ADDR/ipfs/ 2> outfile &&
+  grep "HTTP/1.1 201 Created" outfile &&
+  grep "Location: /ipfs/QmbFMke1KXqnYyBBWxB74N4c5SBnJMVAiMNRcGu6x1AwQH" outfile
+'
+test_kill_ipfs_daemon
+
+test_launch_ipfs_daemon --writable=false
+test_expect_success "ipfs daemon --writable=false overrides Writable=true config" '
```
whyrusleeping: Where do we set the config value to true?

lgierth: Good catch, thanks! Moved the call to `test_config_ipfs_gateway_writable` up a bit

lgierth:
Okay we're finally green, I filed #3208 for an unrelated failure.

cmd/ipfs/daemon.go
```diff
@@ -480,8 +480,8 @@ func serveHTTPGateway(req cmds.Request) (error, <-chan error) {
 	if err != nil {
 		return fmt.Errorf("serveHTTPGateway: req.Option(%s) failed: %s", writableKwd, err), nil
 	}
-	if writableOptionFound {
-		cfg.Gateway.Writable = writable
+	if !writableOptionFound {
+		writable = cfg.Gateway.Writable
```
lgierth: These 4 lines are the essence of the bug.

test/sharness/t0111-gateway-writeable.sh
whyrusleeping: I'm still not seeing it... Does it happen in one of the test libs?

test/sharness/t0111-gateway-writeable.sh
```diff
@@ -9,7 +9,22 @@ test_description="Test HTTP Gateway (Writable)"
...
+
+test_launch_ipfs_daemon --writable
+test_expect_success "ipfs daemon --writable overrides config" '
+  curl -v -X POST http://$GWAY_ADDR/ipfs/ 2> outfile &&
+  grep "HTTP/1.1 201 Created" outfile &&
+  grep "Location: /ipfs/QmbFMke1KXqnYyBBWxB74N4c5SBnJMVAiMNRcGu6x1AwQH" outfile
+'
+test_kill_ipfs_daemon
+
 test_config_ipfs_gateway_writable
```
lgierth: it happens right here

test/sharness/t0111-gateway-writeable.sh
lgierth: Hey I think you might have been viewing the outdated diff -- it happens right there: https://github.com/ipfs/go-ipfs/pull/3206/files#r78514885

(and I filed #3208 for the unrelated test failure)

test/sharness/t0111-gateway-writeable.sh
whyrusleeping: oooooh, i see. thanks!

whyrusleeping:
LGTM!

PR #3207, opened 2016-09-11
Initial Core API, and gateway integration

lgierth:
Okay then, this still misses a few tests but is otherwise good for review. This uses the planned extraction of the gateway as a welcome excuse to start working on the Go side of Core API things. We're implementing the unspecified UnixfsAPI here, for now `Cat()` and `Ls()`. Near-future pull requests will implement more of UnixfsAPI (e.g. `Add()`), and of the Object API (i.e. `SetData()`, `AddLink()`, `RmLink()`).

- `core/coreapi/interface` -- will eventually move to the interface-ipfs-core repo, when UnixfsAPI gets specced (out of scope right now)
- `core/coreapi` -- holds the actual implementation. There's kind of a weird construct around context wiring, for the purpose of being able to pass `net/http.Request.Context()`.

cc @kevina @Kubuxu @whyrusleeping for review.

### DIFF
--- a/core/corehttp/gateway_handler.go
+++ b/core/corehttp/gateway_handler.go
@@ -16,9 +16,10 @@ import (
 	chunk "github.com/ipfs/go-ipfs/importer/chunk"
 	dag "github.com/ipfs/go-ipfs/merkledag"
 	dagutils "github.com/ipfs/go-ipfs/merkledag/utils"
+
+	coreiface "github.com/ipfs/go-ipfs/core/coreapi/interface"
 	path "github.com/ipfs/go-ipfs/path"
 	ft "github.com/ipfs/go-ipfs/unixfs"
-	uio "github.com/ipfs/go-ipfs/unixfs/io"
 
 	humanize "gx/ipfs/QmPSBJL4momYnE7DcUyk2DVhD6rH488ZmHBGLbxNdhU44K/go-humanize"
 	routing "gx/ipfs/QmQKEgGgYCDyk8VNY6A65FpuE4YwbspvjXHco1rdb75PVc/go-libp2p-routing"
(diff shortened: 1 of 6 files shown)


core/corehttp/gateway_handler.go
```diff
@@ -108,22 +114,7 @@ func (i *gatewayHandler) optionsHandler(w http.ResponseWriter, r *http.Request)
 	i.addUserHeaders(w) // return all custom headers (including CORS ones, if set)
 }
 
-func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request) {
-	ctx, cancel := context.WithTimeout(i.node.Context(), time.Hour)
-	// the hour is a hard fallback, we don't expect it to happen, but just in case
-	defer cancel()
-
-	if cn, ok := w.(http.CloseNotifier); ok {
```
whyrusleeping: why are we removing the close notify stuff?

whyrusleeping:
The interfaces look pretty good to me. Simple enough to be useful all over. 

@diasdavid could you take a look at it from the perspective of a 'core interface'?

core/corehttp/gateway_handler.go
lgierth: We're wiring this to `net/http.Request.Context()` now (line 70)

whyrusleeping: I dont think that context implicitly has the close notify stuff on it...

lgierth: Mh you're right. sloppy. I'm gonna move it into ServeHTTP() though so it also applies to POST/PUT ok?

lgierth: > I'm gonna move it into ServeHTTP() though so it also applies to POST/PUT ok?

... in the next PR. gonna reinstate it here for now

whyrusleeping: > ... in the next PR

haha, sounds good to me.

lgierth:
Okay I updated this to improve the context wiring. The close-notify stuff is reinstated, and apiOption is gone in favor of passing contexts into the coreapi functions, in order to satisfy this rule:

>  Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter, typically named ctx:

lgierth:
I'm closing this in favor of #3244 which has been based on this PR here.

PR #3210, opened 2016-09-12
Update libp2p to have fixed spdystream dep

whyrusleeping:
Includes the fix to dockers spdystream multiplexer that was causing issues with js-ipfs interop

License: MIT


Kubuxu:
LGTM

lgierth:
> was using a priority queue for organizing packets, but wasn't setting priorities on the incoming stream. lead to issues with dropping packets because of not knowing what stream they belonged to

Copying this from the go-ipfs sprint notes here -- otherwise the person writing the changelog will have not idea what got fixed

PR #3219, opened 2016-09-13
cmd: validate repo/api file and print nicer error message

### LINKED ISSUE #3191: bad error if $repodir/api file is bad/corrupted

0.4.3-rc3

Type (bug, feature, meta, test failure, question): bug
Area (api, commands, daemon, fuse, etc): commands
Priority (from P0: operations on fire, to P4: functioning): P4

Description:

bad error if $repodir/api file is bad/corrupted

``` sh
> rm ~/.ipfs/api # clear anything there
> touch ~/.ipfs/api # bad file
> ipfs daemon
Error:  is not a 'thin waist' address
```

It should be something like

```
> ipfs daemon
Error: failed to parse <path-to-repo/api file. If there is no daemon running, just delete it.
You can do this with:
    ps aux | grep ipfs  # check there is no ipfs daemon
    rm <path-to-repo>/api
```

Kubuxu:
Resolves #3191
License: MIT
Signed-off-by: Jakub Sztandera kubuxu@protonmail.ch

### DIFF
--- a/cmd/ipfs/main.go
+++ b/cmd/ipfs/main.go
@@ -2,6 +2,7 @@
 package main
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"io"
@@ -16,13 +17,6 @@ import (
 	"syscall"
 	"time"
 
-	manet "gx/ipfs/QmT6Cp31887FpAc25z25YHgpFJohZedrYLWPPspRtj1Brp/go-multiaddr-net"
-	ma "gx/ipfs/QmUAQaWbKxGCUTuoQVvvicbQNZ9APF5pDGWyAZSe93AtKH/go-multiaddr"
-
-	context "context"
-	logging "gx/ipfs/QmSpJByNKFX1sCsHBEp3R73FL4NF6FnQTEGyNAXHm2GS52/go-log"
-	u "gx/ipfs/Qmb912gdngC1UWwTkhuW8knyRbcWeu5kqkxBpveLmW8bSr/go-ipfs-util"
-
 	cmds "github.com/ipfs/go-ipfs/commands"
 	cmdsCli "github.com/ipfs/go-ipfs/commands/cli"
 	cmdsHttp "github.com/ipfs/go-ipfs/commands/http"
@@ -31,7 +25,13 @@ import (
 	repo "github.com/ipfs/go-ipfs/repo"
 	config "github.com/ipfs/go-ipfs/repo/config"
 	fsrepo "github.com/ipfs/go-ipfs/repo/fsrepo"
+
+	logging "gx/ipfs/QmSpJByNKFX1sCsHBEp3R73FL4NF6FnQTEGyNAXHm2GS52/go-log"
+	manet "gx/ipfs/QmT6Cp31887FpAc25z25YHgpFJohZedrYLWPPspRtj1Brp/go-multiaddr-net"
 	loggables "gx/ipfs/QmTMy4hVSY28DdwJ9kBz6y7q6MuioFzPcpM3Ma3aPjo1i3/go-libp2p-loggables"
+	ma "gx/ipfs/QmUAQaWbKxGCUTuoQVvvicbQNZ9APF5pDGWyAZSe93AtKH/go-multiaddr"
+	osh "gx/ipfs/QmXuBJ7DR6k3rmUEKtvVMhwjmXDuJgXXPUt4LQXKBMsU93/go-os-helper"
+	u "gx/ipfs/Qmb912gdngC1UWwTkhuW8knyRbcWeu5kqkxBpveLmW8bSr/go-ipfs-util"
 )
 
 // log is the command logger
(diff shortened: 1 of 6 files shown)


cmd/ipfs/main.go
```diff
@@ -595,17 +594,24 @@ func profileIfEnabled() (func(), error) {
 // a running API service. if there is one, it returns a client.
 // otherwise, it returns errApiNotRunning, or another error.
 func getApiClient(repoPath, apiAddrStr string) (cmdsHttp.Client, error) {
-
-	if apiAddrStr == "" {
-		var err error
-		if apiAddrStr, err = fsrepo.APIAddr(repoPath); err != nil {
+	efmt := fmt.Sprintf("failed to parse %[1]s/api file.\n\terror: %%s\n"+
```
whyrusleeping: lets make this a multiline string constant somewhere

cmd/ipfs/main.go
```diff
@@ -595,17 +594,24 @@ func profileIfEnabled() (func(), error) {
...
+		if err != nil {
 			return nil, err
 		}
 	}
-
-	addr, err := ma.NewMultiaddr(apiAddrStr)
-	if err != nil {
-		return nil, err
+	if len(addr.Protocols()) == 0 {
+		return nil, fmt.Errorf(efmt, "multiaddr doesn't provide any protocols")
```
whyrusleeping: While technically correct, i dont think this helps the user understand what to do

cmd/ipfs/main.go
```diff
@@ -595,17 +594,24 @@ func profileIfEnabled() (func(), error) {
...
-		if apiAddrStr, err = fsrepo.APIAddr(repoPath); err != nil {
+	efmt := fmt.Sprintf("failed to parse %[1]s/api file.\n\terror: %%s\n"+
+		"If there is no daemon running, it is safe to delete it.\nYou can do it with:\n"+
+		"\tps aux | grep ipfs # check there is no ipfs daemon\n\trm %[1]s/api\n", repoPath)
+	var addr ma.Multiaddr
+	var err error
+	if len(apiAddrStr) != 0 {
+		addr, err = ma.NewMultiaddr(apiAddrStr)
+		if err != nil {
+			return nil, fmt.Errorf(efmt, err.Error)
```
Kubuxu: This is also wrong, it should be line 609.

cmd/ipfs/main.go
Kubuxu: It doesn't, it is information for us/people that debug others set-ups.

Kubuxu:
Updated

Kubuxu:
Updated once more.

cmd/ipfs/main.go
```diff
@@ -591,21 +590,37 @@ func profileIfEnabled() (func(), error) {
 	return func() {}, nil
 }
 
+var apiFileErrorFmt string = `Failed to parse %[1]s/api file.
+	error: %[2]s
+If there is no daemon running, it is safe to delete it.
+You can do it with:
+	ps aux | grep ipfs # check there is no ipfs daemon
```
whyrusleeping: Not valid advice for windows... not sure what the best way to do this is.

cc @jbenet @lgierth

repo/fsrepo/fsrepo.go
```diff
@@ -293,16 +295,21 @@ func APIAddr(repoPath string) (string, error) {
...
 	n, err := f.Read(buf)
 	if err != nil && err != io.EOF {
-		return "", err
+		return nil, err
 	}
 
 	s := string(buf[:n])
 	s = strings.TrimSpace(s)
-	return s, nil
+	m, err := ma.NewMultiaddr(s)
```
whyrusleeping: should just `return ma.NewMultiaddr(s)` here

cmd/ipfs/main.go
djdv: On Windows that would be `tasklist | findstr ipfs`, alternatively you can defer people to the Task Manager.

lgierth: thanks @djdv, I think something like this works:

```
If you're sure go-ipfs isn't running, you can just delete it.
Otherwise check:
	ps aux | grep ipfs # on Linux and OSX
	tasklist | findstr ipfs # on Windows

Comment (whyrusleeping) [inline cmd/ipfs/main.go]: We can also generate this message in an `init` function where we can do a switch on `runtime.GOOS`

**Comment by whyrusleeping** [comment]
@Kubuxu status update?

**Comment by Kubuxu** [comment]
I remember looking for a lib that we have already imported that included helpers for checking OSes, and then I did something else.

**Comment by Kubuxu** [comment]
Updated

File: cmd/ipfs/main.go [hunk 6]
```diff
@@ -32,7 +25,14 @@ import (
 	repo "github.com/ipfs/go-ipfs/repo"
 	config "github.com/ipfs/go-ipfs/repo/config"
 	fsrepo "github.com/ipfs/go-ipfs/repo/fsrepo"
+
+	logging "gx/ipfs/QmSpJByNKFX1sCsHBEp3R73FL4NF6FnQTEGyNAXHm2GS52/go-log"
+	"gx/ipfs/QmXuBJ7DR6k3rmUEKtvVMhwjmXDuJgXXPUt4LQXKBMsU93/go-os-helper"
```
whyrusleeping: since the package name and the directory name arent the same, lets add `osh` to name this import.

cmd/ipfs/main.go
```diff
@@ -591,21 +591,47 @@ func profileIfEnabled() (func(), error) {
...
-	}
 
-	addr, err := ma.NewMultiaddr(apiAddrStr)
-	if err != nil {
-		return nil, err
+		if err != nil {
+			return nil, fmt.Errorf(apiErrorFmt, repoPath, err.Error())
+		}
+	}
+	if len(addr.Protocols()) == 0 {
```
Kubuxu: This happens when for some reason the API File is empty.

Kubuxu:
Last commit should address it.

cmd/ipfs/main.go
whyrusleeping: Right, But theres no possible way we can end up with that scenario if the user passes in an api addr with `--api` right?

Kubuxu: No idea that is why I added check in https://github.com/ipfs/go-ipfs/pull/3219/commits/7512eb93bdd749ee70048e933e8b015dfc8c2f27

package.json
```diff
@@ -269,6 +269,12 @@
       "hash": "QmaUewj1HPiCX5mjNHmevQiNWr4eeAn7HBfHcGVbRyafdo",
       "name": "floodsub",
       "version": "0.6.2"
+	},
+	{
+      "author": "Kubuxu",
+      "hash": "QmXuBJ7DR6k3rmUEKtvVMhwjmXDuJgXXPUt4LQXKBMsU93",
+      "name": "go-os-helper",
+      "version": "0.0.0"
```
Kubuxu: oh, look at that

whyrusleeping: what am i looking at?

Kubuxu: Oh, ok, not it is good. Just a second ago GH wasn't showing the bracket that is bellow this line.

whyrusleeping:
@Kubuxu wanna rebase this and make sure the tests pass?

Kubuxu:
Rebased.

Kubuxu:
Recreated it from ground 0 (easier than resolving conflicts in import section).

Kubuxu:
@whyrusleeping should be good for merge right now

PR #3222, opened 2016-09-14
Adder: Don't hold the PinLock when not pinning.

kevina:


kevina:
part of #2634

whyrusleeping:
Good catch, LGTM

PR #3225, opened 2016-09-14
"block rm": move core functionally into blockstore_util package

kevina:


kevina:
I refactored the "blockstore rm" code so I can reuse it in #2634.

kevina:
The teamcity failure looks completely unrelated to my change...

kevina:
@whyrusleeping I would like to get this merged ASAP as moving code around is guaranteed to produce conflicts in the future.

whyrusleeping:
What are we accomplishing by moving this out to a separate package?

blocks/blockstore/util/remove.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+type RmBlocksOpts struct {
+	Prefix string
+	Quiet  bool
+	Force  bool
+}
+
+func RmBlocks(blocks bs.GCBlockstore, pins pin.Pinner, out chan<- interface{}, cids []*cid.Cid, opts RmBlocksOpts) error {
+	go func() {
```
kevina: It returns an error in future code that is not ready to be merged.  See https://github.com/ipfs-filestore/go-ipfs/blob/master/blocks/blockstore/util/remove.go#L30.

blocks/blockstore/util/remove.go
```diff
@@ -0,0 +1,100 @@
...
+	if err != nil {
+		return nil, err
+	}
+	for _, r := range res {
+		if !r.Pinned() {
+			stillOkay = append(stillOkay, r.Key)
+		} else {
+			out <- &RemovedBlock{
+				Hash:  r.Key.String(),
+				Error: r.String(),
```
kevina: No.  r.String() returns a message as to why the block is pinned.  If RemovedBlock returns with the Error field it means the block could not be removed.

blocks/blockstore/util/remove.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+type RmError struct {
+	Fatal bool
+	Msg   string
+}
+
+func (err RmError) Error() string { return err.Msg }
+
+func ProcRmOutput(in <-chan interface{}, sout io.Writer, serr io.Writer) *RmError {
```
kevina: The return type is a pointer to RmError which implemets the error interface and can (and does) return nil.  Nevertheless I no longer require the extra information that RmError returns so I removed that structure and just use a normal error now.

kevina:
@whyrusleeping to answer your direct question, this code is it's own package because I am not sure where else to put it.  The code can not go in the blockstore package because that will create a circular dependency with the "pin" package.

The refactoring is necessary so that I can reuse it in #2634  For example in `filestore clean` I use `FilterPinned` to make sure I am not clearing out any pinned blocks.

blocks/blockstore/util/remove.go
whyrusleeping: Ah, :+1:

blocks/blockstore/util/remove.go
whyrusleeping: @kevina I understand, The issue with the (old) code is this: https://play.golang.org/p/0YxTEK55zb

whyrusleeping:
@kevina okay, that sounds good to me.

PR #3220, opened 2016-09-13
routing: use extracted dht and routing code

whyrusleeping:
License: MIT
Signed-off-by: Jeromy <why@ipfs.io>

PR #3227, opened 2016-09-15
[deps/extract-dht] Fix routing/notifications

geoah:
This is a fix for ipfs/go-ipfs#3220

@whyrusleeping seems to pass locally, wanna see what travis says.


geoah:
The cake; erm the test is a lie.

screenshot 2016-09-15 18 55 41

whyrusleeping:
Fantastic!!

PR #3229, opened 2016-09-16
make: be smarter about downloading running gx install

Kubuxu:
this might have unwanted side effect but I can't see any

//cc @chriscool 

License: MIT

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@
 bin/gx
 bin/gx*
 bin/tmp
+.deps.time
--- a/Makefile
+++ b/Makefile
@@ -51,13 +51,17 @@ path_check:
 deps: go_check gx_check path_check
 	${gx_bin} --verbose install --global
 
+.deps.time: package.json
+	$(MAKE) deps
+	touch .deps.time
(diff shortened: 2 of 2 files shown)


lgierth:
I'm not generally opposed to this, but I'm wondering what the "why" is. Is it that gx is too noisy? Then it should get a quiet mode instead, which only prints stuff if actual work is being done.

I can see an unwanted side effect of this PR which is that gx won't have a chance to verify the dependency files on disk -- they might have been legitimately deleted, some bit might have flipped, etc.

I'm not sure it's doing that currently, but it's something worth considering.

Kubuxu:
Noise is one of the reasons, other is that other is that in 99% of cases it is unneded and it inflates rebuild time by 10% in my case.

Checking the dependencies is good point.

chriscool:
I think there is no good reason to skip the go_check and the path_check targets. They are not related to package.json at all and they depend on things like the go version and environment variable that could change and that are important to check every time we build.

Kubuxu:
Closing it for time being, I will rethink it.

PR #3255, opened 2016-09-24
Add DAGService.GetLinks() method and use it in the GC and elsewhere.

kevina:
This method will use the (also new) LinkService if it is available to
retrieving just the links for a MerkleDAG without necessary having to
retrieve the underlying block.

For now the main benefit is that the pinner will not break when a block
becomes invalid due to a change in the backing file.  This is possible
because the metadata for a block (that includes the Links) is stored
separately and thus always available even if the backing file changes.

For context (and test cases) please see the original commits: 5a732268eecddd19034325e631742f3e8d88aebf 8b396fdf2e4cd7be40f982d6367867c117d5bc4e

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -94,15 +94,14 @@ type IpfsNode struct {
 	PrivateKey ic.PrivKey // the local node's private Key
 
 	// Services
-	Peerstore   pstore.Peerstore     // storage for other Peer instances
-	Blockstore  bstore.GCBlockstore  // the block store (lower level)
-	Blocks      *bserv.BlockService  // the block service, get/add blocks.
-	DAG         merkledag.DAGService // the merkle dag service, get/add objects.
-	LinkService merkledag.LinkService
-	Resolver    *path.Resolver // the path resolution system
-	Reporter    metrics.Reporter
-	Discovery   discovery.Service
-	FilesRoot   *mfs.Root
+	Peerstore  pstore.Peerstore     // storage for other Peer instances
+	Blockstore bstore.GCBlockstore  // the block store (lower level)
+	Blocks     *bserv.BlockService  // the block service, get/add blocks.
+	DAG        merkledag.DAGService // the merkle dag service, get/add objects.
+	Resolver   *path.Resolver       // the path resolution system
+	Reporter   metrics.Reporter
+	Discovery  discovery.Service
+	FilesRoot  *mfs.Root
 
 	// Online
 	PeerHost     p2phost.Host        // the network host (server+client)
--- a/core/core.go
+++ b/core/core.go
@@ -94,14 +94,15 @@ type IpfsNode struct {
 	PrivateKey ic.PrivKey // the local node's private Key
 
 	// Services
-	Peerstore  pstore.Peerstore     // storage for other Peer instances
-	Blockstore bstore.GCBlockstore  // the block store (lower level)
-	Blocks     *bserv.BlockService  // the block service, get/add blocks.
-	DAG        merkledag.DAGService // the merkle dag service, get/add objects.
-	Resolver   *path.Resolver       // the path resolution system
-	Reporter   metrics.Reporter
-	Discovery  discovery.Service
-	FilesRoot  *mfs.Root
+	Peerstore   pstore.Peerstore     // storage for other Peer instances
+	Blockstore  bstore.GCBlockstore  // the block store (lower level)
+	Blocks      *bserv.BlockService  // the block service, get/add blocks.
+	DAG         merkledag.DAGService // the merkle dag service, get/add objects.
+	LinkService merkledag.LinkService
+	Resolver    *path.Resolver // the path resolution system
+	Reporter    metrics.Reporter
+	Discovery   discovery.Service
+	FilesRoot   *mfs.Root
 
 	// Online
 	PeerHost     p2phost.Host        // the network host (server+client)
--- a/exchange/bitswap/bitswap.go
+++ b/exchange/bitswap/bitswap.go
@@ -422,3 +422,7 @@ func (bs *Bitswap) GetWantlist() []key.Key {
 	}
 	return out
 }
+
+func (bs *Bitswap) IsOnline() bool {
+	return true
+}
--- a/merkledag/merkledag.go
+++ b/merkledag/merkledag.go
@@ -23,6 +23,10 @@ type DAGService interface {
 	Get(context.Context, *cid.Cid) (*Node, error)
 	Remove(*Node) error
 
+	// Return all links for a node, may be more effect than
+	// calling Get
+	GetLinks(context.Context, *cid.Cid) ([]*Link, error)
+
 	// GetDAG returns, in order, all the single leve child
 	// nodes of the passed in node.
 	GetMany(context.Context, []*cid.Cid) <-chan *NodeOption
@@ -30,8 +34,14 @@ type DAGService interface {
 	Batch() *Batch
 }
 
-func NewDAGService(bs *bserv.BlockService) DAGService {
-	return &dagService{bs}
+// A LinkService returns the links for a node if they are available
+// locally without having to retrieve the block from the datastore.
+type LinkService interface {
+	Get(*cid.Cid) ([]*Link, error)
+}
+
+func NewDAGService(bs *bserv.BlockService) *dagService {
+	return &dagService{Blocks: bs}
 }
 
 // dagService is an IPFS Merkle DAG service.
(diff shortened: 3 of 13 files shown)


kevina:
Part of #2634.

whyrusleeping:
It doesnt appear that the linkservice interface has anything that implements it, and doesnt appear to be used anywhere

kevina:
Right.  It doesn't have a use yet outside of the filestore.  This commit has an implementation 5a732268eecddd19034325e631742f3e8d88aebf.

kevina:
@whyrusleeping, if #2634 is going to get merged within a reasonable amount of time, I am okay with this P.R. not getting merged at the same time as #2634.  I mainly separated it out for easier reviewing.

whyrusleeping:
@kevina I think that we should change the linkservice interface so that the existing dagservice implements it. That way any thing that just needs a link service can still have the dag service passed to it. And any place that needs both should just have the dagservice passed in. Then, for the filestore, we can make a small wrapper around the current dagservice code that overrides the GetLinks method to use the filestore backing instead.

That way, it makes it easy for us to do other things later, like plug in a graph database to help resolve links.

whyrusleeping:
So what i'm getting at here, is that we want the linkservice and the dagservice to ideally be exposed via the same object. Where the linkservice interface is just a limited subset of the dagservices interface. With them separate, a few things start to feel weird. I would never want to call `GetLinks` on the dagservice if i have a linkservice would I? but now i have to check if i have a linkservice, then if not, call out to the dagservice.

So for the purpose of this PR, we should just be adding the linkservice interface with `GetLinks`, and adding that same method to the existing dagservice interface. That way we can pass the dagservice to any function that requires only a linkservice. Later, when we have different linkservice implementations, we can make a wrapper object for the dagservice like:

```go
type FancyDagServiceWithLinkMagic struct {
    DAGService
    linkStuff *linkResolver
}

func (a *FancyDagServiceWithLinkMagic) GetLinks(....) (Link, error) {
    return a.linkStuff.GetLinks(...)
}
```

And upon node construction, use that object as the nodes `DAG`. That way everbody basically gets the extra fancy link resolver boost for free.

kevina:
@whyrusleeping @Kubuxu I just pushed a commit that implemented what we seamed to agree on over IRC.

merkledag/merkledag.go
```diff
@@ -93,6 +103,23 @@ func (n *dagService) Get(ctx context.Context, c *cid.Cid) (*Node, error) {
...
 
+func (n *dagService) GetLinks(ctx context.Context, c *cid.Cid) ([]*Link, error) {
+	node, err := n.Get(ctx, c)
+	if err != nil {
+		return nil, err
+	}
+	return node.Links, nil
+}
+
+func (n *dagService) GetOfflineLinkService() LinkService {
```
whyrusleeping: I think this can be moved to the main `DAGService` interface. There are many situations where we would want to use this to get an offline dagservice.

merkledag/merkledag.go
```diff
@@ -366,19 +393,19 @@ func legacyCidFromLink(lnk *Link) *cid.Cid {
 // EnumerateChildren will walk the dag below the given root node and add all
 // unseen children to the passed in set.
 // TODO: parallelize to avoid disk latency perf hits?
-func EnumerateChildren(ctx context.Context, ds DAGService, root *Node, visit func(*cid.Cid) bool, bestEffort bool) error {
-	for _, lnk := range root.Links {
+func EnumerateChildren(ctx context.Context, ds LinkService, links []*Link, visit func(*cid.Cid) bool, bestEffort bool) error {
```
whyrusleeping: I would prefer this just take a since `*cid.Cid` as its input instead of an array of links. Makes it easier to call in general.

pin/gc/gc.go
```diff
@@ -24,13 +22,12 @@ var log = logging.Logger("gc")
 //
 // The routine then iterates over every block in the blockstore and
 // deletes any block that is not found in the marked set.
-func GC(ctx context.Context, bs bstore.GCBlockstore, pn pin.Pinner, bestEffortRoots []*cid.Cid) (<-chan key.Key, error) {
+func GC(ctx context.Context, bs bstore.GCBlockstore, ls dag.LinkService, pn pin.Pinner, bestEffortRoots []*cid.Cid) (<-chan key.Key, error) {
 	unlocker := bs.GCLock()
 
-	bsrv := bserv.New(bs, offline.Exchange(bs))
-	ds := dag.NewDAGService(bsrv)
+	ls = ls.GetOfflineLinkService()
```
whyrusleeping: Ah, i see why this was on the `LinkService` and not the `DAGService`. Hrm... thats tricky.

pin/gc/gc.go
```diff
@@ -71,16 +68,16 @@ func GC(ctx context.Context, bs bstore.GCBlockstore, pn pin.Pinner, bestEffortRo
...
 		set.Add(key.Key(c.Hash()))
-		nd, err := ds.Get(ctx, c)
+		links, err := ls.GetLinks(ctx, c)
 		if err != nil {
 			return err
 		}
 
-		// EnumerateChildren recursively walks the dag and adds the keys to the given set
-		err = dag.EnumerateChildren(ctx, ds, nd, func(c *cid.Cid) bool {
+		// EnumerateChildren recursively walks the dag and adls the keys to the given set
```
whyrusleeping: Typo?

pin/pin.go
```diff
@@ -521,19 +521,19 @@ func (p *pinner) PinWithMode(c *cid.Cid, mode PinMode) {
 	}
 }
 
-func hasChild(ds mdag.DAGService, root *mdag.Node, child key.Key) (bool, error) {
-	for _, lnk := range root.Links {
+func hasChild(ds mdag.LinkService, links []*mdag.Link, child key.Key) (bool, error) {
```
whyrusleeping: Should accept a `*cid.Cid` instead of `[]*mdag.Link`

merkledag/merkledag.go
kevina: Look at the return type.

merkledag/merkledag.go
kevina: Okay, I will look into it.

pin/gc/gc.go
kevina: Yes, will fix the comment.

pin/pin.go
kevina: Okay, will look into it.

merkledag/merkledag.go
kevina: I did it this way to enforce type safety.  Otherwise we could be passed a LinkService and get a DAGService.  If you really don't like that I can change it.

whyrusleeping: Type safety is good, lets move forward with this

exchange/bitswap/bitswap.go
```diff
@@ -417,3 +417,7 @@ func (bs *Bitswap) GetWantlist() []key.Key {
 	}
 	return out
 }
+
+func (bs *Bitswap) IsOnline() bool {
```
whyrusleeping: It doesnt look like we use this anywhere. Am i missing something or is this just dead code from a different approach?

kevina: It is used.  I added a `IsOnline()` method to exchange to be able to tell if the exchange is online or not.

whyrusleeping: Ah, yeah. I missed it on my previous pass through

whyrusleeping:
Alright, I think this looks good to me. @Kubuxu wanna give it a good look over?

core/core.go
```diff
@@ -92,14 +92,14 @@ type IpfsNode struct {
...
 	// Services
-	Peerstore  pstore.Peerstore     // storage for other Peer instances
-	Blockstore bstore.GCBlockstore  // the block store (lower level)
-	Blocks     *bserv.BlockService  // the block service, get/add blocks.
-	DAG        merkledag.DAGService // the merkle dag service, get/add objects.
-	Resolver   *path.Resolver       // the path resolution system
-	Reporter   metrics.Reporter
-	Discovery  discovery.Service
-	FilesRoot  *mfs.Root
+	Peerstore   pstore.Peerstore     // storage for other Peer instances
```
kevina: Yes.  I was just about to re-base to clean up this diff noise. :smile:

kevina:
Just rebased on master and assuming the tests pass this should be good to go.  @Kubuxu what do you think?

PR #3257, opened 2016-09-25
WIP: Add support for multiple blockstores

### LINKED ISSUE #3119: Discussion: Support for multiple blockstores.
The filestore is a datastore, but it is only designed to handle a subset of the blocks that can be used in IPFS, therefore the main datastore is still needed and some sort of support for multiple block or datastores is needed so both the filestore and datastore can coexist. **This is a required infrastructure change in order to land #2634.**

The following describes how it is currently implemented now.  Please let me know if you agree with and understand the changes.  Once there is a general consensus I can separate out the non-filestore bits to support this infrastructure change so we can work through the implementation details.

Sorry if it is a bit long.

@whyrusleeping please CC anyone else who should be involved.
## Overview

There are several ways to support the "filestore".  What I believe makes the most sense and will be the easiest to implement will be to support a "cache" and then any number of additional "aux"
datastores with the following semantics:
- When looking up a block the "cache" is first tried, if the block is not found then each "aux" datastores is tried in turn.  The order of the "aux" datastores will explicitly set by the user.
- Any operations that modify the datastore only act on the "cache".
- The "aux" datastores are allowed to read-only.  When they are not additional specialized API calls will be required for adding or removing data from the "aux" datastores.
- Each of these datastore is given a name and can be accessed via its name from the repo.
- Duplicate data should be avoided if possible, but not completely disallowed.

These rules imply that the garbage collector should only attempt to remove data from the "cache" and leave the other datastores alone.
## High level implementation details

The multiplexing can either happen at the datastore or the blockstore level. I originally implemented it at the datastore level but changed it to the blockstore level to better interact with caching.  The filestore is still implemented as a datastore (for now).

In the `fsrepo` Normal blocks are mounted under the `/blocks` prefix (this is unchanged) the filestore is mounted under the `/filestore` prefix (this is new).  The `fsrepo` has been enhanced to be able to retrieve the underlying datastore based on its prefix.  (This is required by the filestore.)

The top-level blockstore is now a multi-blockstore that works by checking a pre-configured set of prefixs in turn in order to find a matching key.  Each mount is wrapped in its own blockstore with its
own caching semantics.  The first mount "/blocks" is considered the cache and all Put and Deletes only go to the cache.  The multiblock store interface is as follows:

```
type MultiBlockstore interface {
    GCBlockstore
    FirstMount() Blockstore // returns the first mount
    Mounts() []string       // lists the mounts
    Mount(prefix string) Blockstore  // returns a mount by name
    Locate(key key.Key) []LocateInfo // lists all locations of a block
}
```

The garbage collector uses FirstMount().AllKeysChan(ctx) to get the list of blocks for the list of to try and delete.

Any caching is currently only done on the first mount for now.

As an implementation detail it is worth noting that files are removed or added to the filestore directly using a specialized interface that bypasses the normal Blockstore and Filestore interface.  This was discussed with @whyrusleeping (https://github.com/ipfs/go-ipfs/pull/2634#issuecomment-220493404).

Duplicate blocks (that is blocks found under more than one mount) are not forbidden as doing so would be impractical.  The Locate() method can be used to discover what mount a block is found.  It will list all mounts which can be used to help eliminate the duplicates.
## Other uses

The two mounts `/blocks` and `/filestore` are currently hard coded, with some effort this can be made into a more general purpose mechanize to support multiple blockstores.

One use case I can think of is to have a separate read-only datastore to store permanent content as a alternative to maintaining a large pin-set which currently has performance problems.  The datastore could even be on a readonly filesystem to prevent any possibly of the data accidental being deleted either by user error or a software bug.  Some additional design decisions will need to made for this so I am not proposing it right now, but merely offering it as a possibility.

Another possibility is to support a local cache on a local filesystem and a larger datastore on the cloud.

kevina:
Closes #3119.

Required for #2634.

### DIFF
--- a/blocks/blockstore/blockstore.go
+++ b/blocks/blockstore/blockstore.go
@@ -73,13 +73,10 @@ type gcBlockstore struct {
 }
 
 func NewBlockstore(d ds.Batching) *blockstore {
-	return NewBlockstoreWPrefix(d, "")
+	return NewBlockstoreWPrefix(d, DefaultPrefix)
 }
 
 func NewBlockstoreWPrefix(d ds.Batching, prefix string) *blockstore {
-	if prefix == "" {
-		prefix = DefaultPrefix
-	}
 	var dsb ds.Batching
 	prefixKey := ds.NewKey(prefix)
 	dd := dsns.Wrap(d, prefixKey)
--- a/blocks/blockstore/blockstore.go
+++ b/blocks/blockstore/blockstore.go
@@ -21,7 +21,9 @@ import (
 var log = logging.Logger("blockstore")
 
 // BlockPrefix namespaces blockstore datastores
-var BlockPrefix = ds.NewKey("blocks")
+const DefaultPrefix = "/blocks"
+
+var blockPrefix = ds.NewKey(DefaultPrefix)
 
 var ValueTypeMismatch = errors.New("the retrieved value is not a Block")
 var ErrHashMismatch = errors.New("block in storage has different hash than requested")
@@ -71,20 +73,26 @@ type gcBlockstore struct {
 }
 
 func NewBlockstore(d ds.Batching) *blockstore {
+	return NewBlockstoreWPrefix(d, "")
+}
+
+func NewBlockstoreWPrefix(d ds.Batching, prefix string) *blockstore {
+	if prefix == "" {
+		prefix = DefaultPrefix
+	}
 	var dsb ds.Batching
-	dd := dsns.Wrap(d, BlockPrefix)
+	prefixKey := ds.NewKey(prefix)
+	dd := dsns.Wrap(d, prefixKey)
 	dsb = dd
 	return &blockstore{
 		datastore: dsb,
+		prefix:    prefixKey,
 	}
 }
 
 type blockstore struct {
 	datastore ds.Batching
-
-	lk      sync.RWMutex
-	gcreq   int32
-	gcreqlk sync.Mutex
+	prefix    ds.Key
 
 	rehash bool
 }
@@ -130,11 +138,8 @@ func (bs *blockstore) Get(k *cid.Cid) (blocks.Block, error) {
 func (bs *blockstore) Put(block blocks.Block) error {
 	k := dshelp.CidToDsKey(block.Cid())
 
-	// Has is cheaper than Put, so see if we already have it
-	exists, err := bs.datastore.Has(k)
-	if err == nil && exists {
-		return nil // already stored.
-	}
+	// Note: The Has Check is now done by the MultiBlockstore
+
 	return bs.datastore.Put(k, block.RawData())
 }
 
@@ -145,11 +150,6 @@ func (bs *blockstore) PutMany(blocks []blocks.Block) error {
 	}
 	for _, b := range blocks {
 		k := dshelp.CidToDsKey(b.Cid())
-		exists, err := bs.datastore.Has(k)
-		if err == nil && exists {
-			continue
-		}
-
 		err = t.Put(k, b.RawData())
 		if err != nil {
 			return err
@@ -175,7 +175,7 @@ func (bs *blockstore) AllKeysChan(ctx context.Context) (<-chan *cid.Cid, error) 
 	// KeysOnly, because that would be _a lot_ of data.
 	q := dsq.Query{KeysOnly: true}
 	// datastore/namespace does *NOT* fix up Query.Prefix
-	q.Prefix = BlockPrefix.String()
+	q.Prefix = bs.prefix.String()
 	res, err := bs.datastore.Query(q)
 	if err != nil {
 		return nil, err
--- a/blocks/blockstore/bloom_cache_test.go
+++ b/blocks/blockstore/bloom_cache_test.go
@@ -104,11 +104,11 @@ func TestHasIsBloomCached(t *testing.T) {
 	block := blocks.NewBlock([]byte("newBlock"))
 
 	cachedbs.PutMany([]blocks.Block{block})
-	if cacheFails != 2 {
-		t.Fatalf("expected two datastore hits: %d", cacheFails)
+	if cacheFails != 1 {
+		t.Fatalf("expected datastore hits: %d", cacheFails)
 	}
 	cachedbs.Put(block)
-	if cacheFails != 3 {
+	if cacheFails != 2 {
 		t.Fatalf("expected datastore hit: %d", cacheFails)
 	}
 
--- a/blocks/blockstore/multi.go
+++ b/blocks/blockstore/multi.go
@@ -13,12 +13,18 @@ import (
 	dsq "gx/ipfs/QmbzuUusHqaLLoNTDEVLcSF6vZDHZDLPC7p4bztRvvkXxU/go-datastore/query"
 )
 
+type LocateInfo struct {
+	Prefix string
+	Error  error
+}
+
 type MultiBlockstore interface {
 	Blockstore
 	GCLocker
 	FirstMount() Blockstore
 	Mounts() []string
 	Mount(prefix string) Blockstore
+	Locate(*cid.Cid) []LocateInfo
 }
 
 type Mount struct {
@@ -88,6 +94,15 @@ func (bs *multiblockstore) Get(c *cid.Cid) (blocks.Block, error) {
 		}
 	}
 	return nil, firstErr
+}
+
+func (bs *multiblockstore) Locate(c *cid.Cid) []LocateInfo {
+	res := make([]LocateInfo, 0, len(bs.mounts))
+	for _, m := range bs.mounts {
+		_, err := m.Blocks.Get(c)
+		res = append(res, LocateInfo{m.Prefix, err})
+	}
+	return res
 }
 
 func (bs *multiblockstore) Put(blk blocks.Block) error {
--- a/blocks/blockstore/multi.go
+++ b/blocks/blockstore/multi.go
@@ -1,0 +1,122 @@
+package blockstore
+
+// A very simple multi-blockstore that analogous to a unionfs Put and
+// DeleteBlock only go to the first blockstore all others are
+// considered readonly.
+
+import (
+	//"errors"
+	"context"
+
+	blocks "github.com/ipfs/go-ipfs/blocks"
+	cid "gx/ipfs/QmXfiyr2RWEXpVDdaYnD2HNiBk6UBddsvEP4RPfXb6nGqY/go-cid"
+)
+
+type MultiBlockstore interface {
+	Blockstore
+	GCLocker
+	FirstMount() Blockstore
+	Mounts() []string
+	Mount(prefix string) Blockstore
+}
+
+type Mount struct {
+	Prefix string
+	Blocks Blockstore
+}
+
+func NewMultiBlockstore(mounts ...Mount) *multiblockstore {
+	return &multiblockstore{
+		mounts: mounts,
+	}
+}
+
+type multiblockstore struct {
+	mounts []Mount
+	gclocker
+}
+
+func (bs *multiblockstore) FirstMount() Blockstore {
+	return bs.mounts[0].Blocks
+}
+
+func (bs *multiblockstore) Mounts() []string {
+	mounts := make([]string, 0, len(bs.mounts))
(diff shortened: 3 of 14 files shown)


repo/fsrepo/fsrepo.go
```diff
@@ -557,6 +565,27 @@ func (r *FSRepo) Datastore() repo.Datastore {
 	return d
 }
 
+func (r *FSRepo) DirectMount(prefix string) ds.Datastore {
```
whyrusleeping: Maybe call this `DatastoreAt` or `GetMount`. `DirectMount` just leaves me confused about what this is

kevina:
@whyrusleeping 

> What is the reasoning for doing this at the blockstore level? I thought you were doing this with a multi-datastore type thing?

To better interact with caching.  In particular the bloom filter.

blocks/blockstore/blockstore.go
```diff
@@ -57,21 +57,41 @@ type GCBlockstore interface {
...
+	Blockstore
+	GCLocker
+}
+
 func NewBlockstore(d ds.Batching) *blockstore {
+	return NewBlockstoreWPrefix(d, "")
+}
+
+func NewBlockstoreWPrefix(d ds.Batching, prefix string) *blockstore {
+	if prefix == "" {
```
kevina: Okay.

blocks/blockstore/bloom_cache_test.go
```diff
@@ -104,11 +104,11 @@ func TestHasIsBloomCached(t *testing.T) {
 	block := blocks.NewBlock([]byte("newBlock"))
 
 	cachedbs.PutMany([]blocks.Block{block})
-	if cacheFails != 2 {
-		t.Fatalf("expected two datastore hits: %d", cacheFails)
+	if cacheFails != 1 {
```
kevina: I verified this was correct, the change was due to an implementation detail.  Unfortually, I can't remember why.  If it is important enough I will spend an hour or two looking into it.

whyrusleeping: It always weird me out when i see tests changing. Maybe @Kubuxu is more familiar with this and can check it?

Kubuxu: It was to because `PutMany` on default blockstore was accessing datastore two times, once checking `Has` for a key and then calling `Put`.

This means that now we don't check the datastore with `Has` we just `Put` the value.

blocks/blockstore/blockstore.go
```diff
@@ -112,11 +132,8 @@ func (bs *blockstore) Get(k key.Key) (blocks.Block, error) {
 func (bs *blockstore) Put(block blocks.Block) error {
 	k := block.Key().DsKey()
 
-	// Has is cheaper than Put, so see if we already have it
-	exists, err := bs.datastore.Has(k)
-	if err == nil && exists {
-		return nil // already stored.
-	}
+	// Note: The Has Check is now done by the MultiBlockstore
```
Kubuxu: Are you sure about that? MultiBlockstore just calls Has method of Blockstore but it doesn't mean that Blockstore (on Datastore) will check the Has of datastore.

blocks/blockstore/multi.go
```diff
@@ -0,0 +1,154 @@
...
+			firstErr = err
+		}
+	}
+	return nil, firstErr
+}
+
+func (bs *multiblockstore) Locate(key key.Key) []LocateInfo {
+	res := make([]LocateInfo, 0, len(bs.mounts))
+	for _, m := range bs.mounts {
+		_, err := m.Blocks.Get(key)
```
Kubuxu: Why not use `Has` here?

blocks/blockstore/blockstore.go
Kubuxu: UPDATE: ahh, you call and explicit `Has`.

blocks/blockstore/multi.go
```diff
@@ -0,0 +1,154 @@
...
+		res = append(res, LocateInfo{m.Prefix, err})
+	}
+	return res
+}
+
+func (bs *multiblockstore) Put(blk blocks.Block) error {
+	// Has is cheaper than Put, so see if we already have it
```
Kubuxu: This comment is misleading. You don't check Has before Put because it is cheaper.

If you don't do it, you will end up with data being duplicated in two blockstores.

blocks/blockstore/multi.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+	out := make(chan key.Key, dsq.KeysOnlyBufSize)
+	go func() {
+		defer close(out)
+		for _, in0 := range in {
+			for key := range in0 {
+				out <- key
+			}
+		}
+	}()
```
Kubuxu: If @whyrusleeping agrees with me I would open a goroutine per BS and make them pipe from one AllKeysChan to one external. This way if first BS is slow and second is fast, the first won't slow down whole process.

blocks/blockstore/multi.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+func (bs *multiblockstore) Get(key key.Key) (blocks.Block, error) {
+	var firstErr error
+	for _, m := range bs.mounts {
+		blk, err := m.Blocks.Get(key)
+		if err == nil {
+			return blk, nil
+		}
+		if firstErr == nil || firstErr == ErrNotFound {
```
Kubuxu: I would replace it with `firstErr == nil && err != ErrNotFound`

blocks/blockstore/multi.go
kevina: Not necessary, the problem is with the filestore simply calling Has() will not guarantee the block is available as the backing file might have changed or no longer be available.

blocks/blockstore/multi.go
kevina: Your right, I will fix it to be clearer.

blocks/blockstore/multi.go
Kubuxu: This should be on filestore to figure out, IMO.
Why should we load XMiB into memory from harddrive if other blockstore can figure it out without loading whole content.

blocks/blockstore/multi.go
kevina: If I do that than firstErr could end up being nil even if the block was not found.  The idea is to return ErrNotFound only if that is the case for all the blockstore's.  If not than return the more serious error.

blocks/blockstore/blockstore.go
kevina: So can I take it that this is resolved?

blocks/blockstore/util/remove.go
```diff
@@ -27,14 +27,23 @@ type RmBlocksOpts struct {
 	Force  bool
 }
 
-func RmBlocks(blocks bs.GCBlockstore, pins pin.Pinner, out chan<- interface{}, cids []*cid.Cid, opts RmBlocksOpts) error {
+func RmBlocks(mbs bs.MultiBlockstore, pins pin.Pinner, out chan<- interface{}, cids []*cid.Cid, opts RmBlocksOpts) error {
+	prefix := opts.Prefix
+	if prefix == "" {
+		prefix = mbs.Mounts()[0]
```
whyrusleeping: does this only remove from one blockstore?

core/builder.go
```diff
@@ -179,11 +180,20 @@ func setupNode(ctx context.Context, n *IpfsNode, cfg *BuildCfg) error {
...
 
-	n.Blockstore, err = bstore.CachedBlockstore(bs, ctx, opts)
+	cbs, err := bstore.CachedBlockstore(bs, ctx, opts)
 	if err != nil {
 		return err
 	}
 
+	mounts := []bstore.Mount{{fsrepo.CacheMount, cbs}}
+
+	if n.Repo.DirectMount(fsrepo.FilestoreMount) != nil {
```
whyrusleeping: Filestore mount? does this do anything yet?

core/commands/add.go
```diff
@@ -78,6 +80,7 @@ You can now refer to the added file in a gateway, like so:
 		cmds.BoolOption(hiddenOptionName, "H", "Include files that are hidden. Only takes effect on recursive add.").Default(false),
 		cmds.StringOption(chunkerOptionName, "s", "Chunking algorithm to use."),
 		cmds.BoolOption(pinOptionName, "Pin this object when adding.").Default(true),
+		cmds.BoolOption(allowDupName, "Add even if blocks are in non-cache blockstore.").Default(false),
```
whyrusleeping: This really complicates things... thoughts @Kubuxu ?

core/core.go
```diff
@@ -94,11 +94,11 @@ type IpfsNode struct {
 	PrivateKey ic.PrivKey // the local node's private Key
 
 	// Services
-	Peerstore  pstore.Peerstore     // storage for other Peer instances
-	Blockstore bstore.GCBlockstore  // the block store (lower level)
-	Blocks     bserv.BlockService   // the block service, get/add blocks.
-	DAG        merkledag.DAGService // the merkle dag service, get/add objects.
-	Resolver   *path.Resolver       // the path resolution system
+	Peerstore  pstore.Peerstore       // storage for other Peer instances
+	Blockstore bstore.MultiBlockstore // the block store (lower level)
```
whyrusleeping: should be an interface

repo/fsrepo/defaultds.go
```diff
@@ -20,23 +21,29 @@ const (
 	flatfsDirectory  = "blocks"
 )
 
-func openDefaultDatastore(r *FSRepo) (repo.Datastore, error) {
+const (
+	RootMount      = "/"
+	CacheMount     = "/blocks" // needs to be the same as blockstore.DefaultPrefix
```
whyrusleeping: why is this called the cache mount?

blocks/blockstore/multi.go
```diff
@@ -0,0 +1,158 @@
...
+			firstErr = err
+		}
+	}
+	return nil, firstErr
+}
+
+func (bs *multiblockstore) Locate(c *cid.Cid) []LocateInfo {
+	res := make([]LocateInfo, 0, len(bs.mounts))
+	for _, m := range bs.mounts {
+		_, err := m.Blocks.Get(c)
```
kevina: I am fine with that.  Not 100% happy with it as filestore Has doesn't really guarantee the block is available (and changing it so it does will make the Has() call really expensive), but its not a major problem.

blocks/blockstore/multi.go
```diff
@@ -0,0 +1,158 @@
...
+		res = append(res, LocateInfo{m.Prefix, err})
+	}
+	return res
+}
+
+func (bs *multiblockstore) Put(blk blocks.Block) error {
```
kevina: Yeah, but rather difficult to avoid.

blocks/blockstore/util/remove.go
kevina: Yes.

core/commands/add.go
kevina: i brought this up in IRC.

core/core.go
kevina: what should be an interface?

repo/fsrepo/defaultds.go
kevina: Please see #3119.

core/core.go
whyrusleeping: the blockstore should just be of type `Blockstore`. That way its easy to swap out with any other blockstore implementation.

kevina: @whyrusleeping  I did not design the MultiBlockstore to be able to be swapped out.  (As I see it it, it was designed to allow use of other Blockstores so why would you want to swap it out, but I can understand your viewpoint)   If that is something you want it is likely possible.  I image I can have something done in a few days.

blocks/blockstore/multi.go
kevina: I partly take the Has() comment back.  The filestore Has() needs fixing or the Put below needs fixing.

whyrusleeping:
This is no longer needed

kevina:
@whyrusleeping I beg to differ.  Something like this might still be needed.  For example if you want multiple ipfs-packs in a single node.

whyrusleeping:
Alright, I'll keep the branch around (its also archived on the branch archive repo). But keeping the PR closed until we have an actionable usecase need for it.

PR #3260, opened 2016-09-25
Update goprocess dependency to 1.0.0

kevina:
Update goprocess dependency to 1.0.0 to avoid a version mismatch with what
go-datastore uses.

Part of #2634


kevina:
Note: It seams that all the imports will already pointing to the new version, it was just the `package.json` file that was out of date.

PR #3264, opened 2016-09-26
Explicitly specify the input file to tar with "-f -"

vasild:
On FreeBSD the default one (if no -f is specified) is /dev/sa0,
not stdin.

License: MIT
Signed-off-by: Vasil Dimov <vd@FreeBSD.org>


lgierth:
LGTM, thanks!

PR #3265, opened 2016-09-26
use $(MAKE) instead of 'make' in Makefiles (follow-up to f3dc11c)

vasild:
License: MIT
Signed-off-by: Vasil Dimov <vd@FreeBSD.org>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,7 +80,7 @@ test_short: build test_go_short test_sharness_short
 test_expensive: build test_go_expensive test_sharness_expensive windows_build_check
 
 test_3node:
-	cd test/3nodetest && make
+	$(MAKE) -C test/3nodetest
 
 test_go_short:
 	$(go_test) -test.short ./...


lgierth:
Does `$(make)` resolve to whatever make binary is currently running, while just `make` will use `$PATH` to resolve, and might end up with a different binary? In that case, LGTM.

Kubuxu:
Yes it is the case, also `$(MAKE)` works on BSDs where you have to build go-ipfs with `gmake` as we use GNU extensions in some places.

vasild:
Right, the problem this patch addresses is that the FreeBSD port does set MAKE=gmake in the environment and also runs it as "gmake ...whatever..." but a hardcoded make command in the Makefile ends up picking up /usr/bin/make from PATH. There are also lots of other places in this Makefile which use $(MAKE), I guess this hardcoded "make" was forgotten in f3dc11c which did s/make/$(MAKE)/.

Issue #3262, opened 2016-09-26
Panic while running daemon


whyrusleeping:
#### Version information: 
master

#### Type:
Panic

#### Priority:
P1

#### Description:
Got a panic while running the daemon.

```
Initializing daemon...
Swarm listening on /ip4/10.5.8.247/tcp/4001
Swarm listening on /ip4/127.0.0.1/tcp/4001
Swarm listening on /ip6/::1/tcp/4001
API server listening on /ip4/127.0.0.1/tcp/5001
Gateway (readonly) server listening on /ip4/127.0.0.1/tcp/8080
Daemon is ready
panic: runtime error: index out of range

goroutine 95 [running]:
panic(0xc00500, 0xc420010110)
        /home/whyrusleeping/go/src/runtime/panic.go:500 +0x1a1
github.com/ipfs/go-ipfs/merkledag.(*Node).getPBNode(0xc42502c4b0, 0xc42023ec88)
        /home/whyrusleeping/gopkg/src/github.com/ipfs/go-ipfs/merkledag/coding.go:61 +0x33f
github.com/ipfs/go-ipfs/merkledag.(*Node).Marshal(0xc42502c4b0, 0xc422c3e760, 0xc422c3e760, 0x11387e0, 0xc422c3e760, 0x2)
        /home/whyrusleeping/gopkg/src/github.com/ipfs/go-ipfs/merkledag/coding.go:45 +0x2f
github.com/ipfs/go-ipfs/merkledag.(*Node).EncodeProtobuf(0xc42502c4b0, 0x0, 0x3, 0x22, 0x22, 0x22, 0x30)
        /home/whyrusleeping/gopkg/src/github.com/ipfs/go-ipfs/merkledag/coding.go:80 +0xb3
github.com/ipfs/go-ipfs/merkledag.(*Node).Multihash(0xc42502c4b0, 0xc42023ed98, 0x449915, 0xc42023eda8)
        /home/whyrusleeping/gopkg/src/github.com/ipfs/go-ipfs/merkledag/node.go:271 +0x34
github.com/ipfs/go-ipfs/merkledag.(*Node).Key(0xc42502c4b0, 0xc420063020, 0x0)
        /home/whyrusleeping/gopkg/src/github.com/ipfs/go-ipfs/merkledag/node.go:249 +0x2b
[...]

PR #3268, opened 2016-09-28
resolves issue #3262

kf6nux:
It appears that the slice size of `n.Links` is being modified after the destination capacity of `pbn.Links` is set, thereby causing a panic when attempting to access an index out of range on the new slice.  By using append, we can avoid using mutexes and retain most of the original performance.

### DIFF
--- a/merkledag/coding.go
+++ b/merkledag/coding.go
@@ -52,9 +52,7 @@ func (n *Node) Marshal() ([]byte, error) {
 
 func (n *Node) getPBNode() *pb.PBNode {
 	pbn := &pb.PBNode{}
-	if len(n.Links) > 0 {
-		pbn.Links = make([]*pb.PBLink, len(n.Links))
-	}
+	pbn.Links = make([]*pb.PBLink, 0, len(n.Links))
 
 	sort.Stable(LinkSlice(n.Links)) // keep links sorted
 	for i, l := range n.Links {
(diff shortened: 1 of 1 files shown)


whyrusleeping:
Hrm... I think that the issue in 3262 is likely a race condition. In that event, this change won't fix the problem

kf6nux:
You are correct that it appears to be a race condition, but why do you say this fix won't solve the problem?

whyrusleeping:
Because this fix appends an empty node to the end of an already allocated array. With no race condition involved, it would just append entries to the end of the existing array, resulting in us having a links array of twice the length we expect. And since this doesnt fix the actual race condition, when the race condition actually happens, we're going to have an array of undefined length as the appends will be interspersed between calling goroutines.

kf6nux:
Oh, I forgot to change the make statement.  I'll do that and then it'll be correct.  Let me know if you need a playground proof.

kf6nux:
Fixed.  Let me know if you need a playground proof.

kevina:
@kf6nux you need to fix your commit messages, please see the above message from GitCop.

whyrusleeping:
#3262 was resolved by #3271, Thank you very much for the initiate and work, but i'm going to close this as it is no longer necessary. (let me know if you disagree)

Issue #2917, opened 2016-06-28
IPFS consumes a large amount of network traffic


cminnoy:
Ubuntu 16.01 Intel x64 
ipfs version 0.4.1

Type bug
Area DHT?
Priority high

Description:

Hi,

I'm running two instances of IPFS (on two different machines), each connect to a different ISP.
I noticed for a while a very high network traffic from those nodes. Those nodes run only IPFS,
so no other network traffic from those machines than IPFS network traffic.
Last week, I didn't use IPFS on those machines, but IPFS was still turned on.
During that one week, the first node consumed 58 Gigabyte of network traffic, and the second node
59,9 Gigabyte. This network traffic was completly generated by IPFS while not using IPFS actively.
Sure this must be a serious bug. When I stop the IPFS daemon service, the network traffic consumption stopped immediatly.

Cheers,

Chris

Kubuxu:
0.4.1 is very very old version, please try using build from the current `master` branch as it contains many performance improvements but we are still working on in.

Kubuxu:
0.4.1 is very very old version, please try using build from the current `master` branch as it contains many performance improvements but we are still working on in.

gsf:
Same behavior on my nodes as well, running on master.

Stebalien:
If you run a public facing node, it will serve up all blocks it has downloaded on request. Unfortunately, even if you haven't downloaded anything, you'll still be constantly bombarded with wantlists and download requests.

Kubuxu:
Not really download requests but general chatter (DHT upkeep which we try to balance) and wantlists which we will try to optimize. Reducing passive traffic is on our list and this version includes few improvements, for example https://github.com/ipfs/go-ipfs/pull/2817 but it needs network to adopt it before you will see reduction of traffic.

Stebalien:
It would also be nice if `ipfs stats bw` gave an accurate bandwidth estimate. For me, actual (passive) bandwidth usage is usually over an order of magnitude greater than that reported by ipfs.

cminnoy:
Checked v0.4.3-dev yesterday/today. In 24 hours the DHT consumed 6.2 gigabyte of network traffic (almost half to receive and transfer). Hardly a good figure.

pchiusano:
Great that this is being worked on... this issue is super important for anyone thinking of running IPFS nodes in the cloud where they will have to pay for bandwidth.

[...]

PR #3269, opened 2016-09-28
add experimental dht client mode flag

whyrusleeping:
This allows you to start up an ipfs daemon with `--routing=dhtclient`. In that mode, your dht will not accept incoming requests, but will still be able to query the dht for providers and peer routing information.

License: MIT


whyrusleeping:
This option will be added as a config setting after it gets some mileage as an experimental feature.

whyrusleeping:
Should help with #2917

Kubuxu:
SGTM, should we add it to help text of daemon?

whyrusleeping:
@Kubuxu Hrm... Yeah, I guess we could do that now (as opposed to later after the experimental phase)

PR #3276, opened 2016-09-30
Don't drop error in readStreamedJson.

kevina:
Better fix than #3230 to make sure errors from addAllAndPin make it to
the client.

License: MIT

### DIFF
--- a/commands/http/client.go
+++ b/commands/http/client.go
@@ -180,7 +180,7 @@ func getResponse(httpRes *http.Response, req cmds.Request) (cmds.Response, error
 		// if output is coming from a channel, decode each chunk
 		outChan := make(chan interface{})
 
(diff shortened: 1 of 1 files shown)


whyrusleeping:
@kevina can you link me to the test case for this in your other PR?

kevina:
@whyrusleeping  This commit https://github.com/ipfs/go-ipfs/pull/2634/commits/6857aecd6dee54557f32caa7b8e447e31f71f6d8 contains the bulk of it, the errors are generated in `core/commands/filestore.go`.  The test cases are the ones that have `test_must_fail`.  Without this fix most of tests with `test_must_fail` will fail as the error will be ignored.

whyrusleeping:
Alright, then this LGTM

PR #3284, opened 2016-10-05
update to libp2p 4.0.1 and propogate other changes

whyrusleeping:
cc @Kubuxu @lgierth 

License: MIT


blocks/blockstore/blockstore.go
```diff
@@ -7,14 +7,14 @@ import (
 	"sync"
 	"sync/atomic"
 
+	context "context"
```
lgierth: This can be just `"context"` up there with the other stdlib imports, same for the other occurences. Really no blocking issue though.

core/commands/dht.go
```diff
@@ -68,7 +69,9 @@ var queryDhtCmd = &cmds.Command{
 		events := make(chan *notif.QueryEvent)
 		ctx := notif.RegisterForQueryEvents(req.Context(), events)
 
-		closestPeers, err := dht.GetClosestPeers(ctx, key.Key(req.Arguments()[0]))
+		k := string(b58.Decode(req.Arguments()[0]))
```
lgierth: Any reason for the string cast? It's fine, just wondering.

whyrusleeping: b58.Decode returns bytes, i needed a string

namesys/publisher.go
```diff
@@ -356,9 +356,9 @@ func InitializeKeyspace(ctx context.Context, ds dag.DAGService, pub Publisher, p
 	return nil
 }
 
-func IpnsKeysForID(id peer.ID) (name, ipns key.Key) {
-	namekey := key.Key("/pk/" + id)
-	ipnskey := key.Key("/ipns/" + id)
+func IpnsKeysForID(id peer.ID) (name, ipns string) {
+	namekey := "/pk/" + string(id)
+	ipnskey := "/ipns/" + string(id)
```
lgierth: Where is the changed caller of this?

whyrusleeping: In:

- core/commands/dht_test.go:12
- namesys/publisher.go:68
- namesys/publisher.go:141
- namesys/republisher/repub.go:90

namesys/publisher.go
```diff
@@ -79,8 +79,8 @@ func (p *ipnsPublisher) PublishWithEOL(ctx context.Context, k ci.PrivKey, value
 	return PutRecordToRouting(ctx, k, value, seqnum, eol, p.routing, id)
 }
 
-func (p *ipnsPublisher) getPreviousSeqNo(ctx context.Context, ipnskey key.Key) (uint64, error) {
-	prevrec, err := p.ds.Get(ipnskey.DsKey())
+func (p *ipnsPublisher) getPreviousSeqNo(ctx context.Context, ipnskey string) (uint64, error) {
+	prevrec, err := p.ds.Get(dshelp.NewKeyFromBinary(ipnskey))
```
lgierth: Ah here's the changed caller for `IpnsKeysForID()`, so to speak.

namesys/publisher.go
lgierth: Got it thanks! LGTM :+1:

namesys/publisher.go
lgierth: What I mean is here's the function that gets the return from `IpnsKeysForID()` and which got the signature change

core/commands/dht.go
whyrusleeping: This actually is a bugfix. Previously we we'rent base58 decoding this like we were supposed to.

PR #3294, opened 2016-10-08
Create a write-through block service

kevina:
Create a block service where all writes are guaranteed to go though to the blockstore.

Pull request #3105 (bitswap: don't re-provide blocks we've provided very recently) broke adding files via the filestore (i.e., --no-copy, #2634).  The first attempt to fix this was #3253, this is the second.

This also makes BlockService an interface, although in the end it was not strictly required.

### DIFF
--- a/blockservice/blockservice.go
+++ b/blockservice/blockservice.go
@@ -4,6 +4,7 @@
 package blockservice
 
 import (
+	"context"
 	"errors"
 	"fmt"
 
@@ -11,7 +12,6 @@ import (
 	"github.com/ipfs/go-ipfs/blocks/blockstore"
 	exchange "github.com/ipfs/go-ipfs/exchange"
 
-	context "context"
 	logging "gx/ipfs/QmSpJByNKFX1sCsHBEp3R73FL4NF6FnQTEGyNAXHm2GS52/go-log"
 	cid "gx/ipfs/QmakyCk6Vnn16WEKjbkxieZmM2YLTzkFWizbmGowoYPjro/go-cid"
 )
--- a/blockservice/blockservice.go
+++ b/blockservice/blockservice.go
@@ -37,6 +37,9 @@ type BlockService interface {
 type blockService struct {
 	blockstore blockstore.Blockstore
 	exchange   exchange.Interface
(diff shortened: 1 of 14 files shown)


blockservice/blockservice.go
```diff
@@ -24,10 +24,24 @@ var ErrNotFound = errors.New("blockservice: key not found")
...
+	Close() error
+}
+
+type blockService struct {
 	// TODO don't expose underlying impl details
-	Blockstore blockstore.Blockstore
-	Exchange   exchange.Interface
+	blockstore   blockstore.Blockstore
+	exchange     exchange.Interface
+	// If checkFirst is true than first check that a block doesn't
```
whyrusleeping: typo: s/than/then/

kevina:
> Think you could add a test or two to make sure the 'writeThrough' happens as expected?

Off hand, I am not sure how.  It is not used outside of #2634.  Several test in #2634 will fail if the writes don't make it through to the blockstore.  In particular, In the backing file moves, the way to update the block is to re-add the file under the new name.

One thing I meant to ask you is if you want to use this by default for the normal add path?  This will restore the behavior of readding a file will re-publish it.  If this is done there might be a way to test for that, but I might need some help as I am not sure how to test if a block is (re)published.

blockservice/blockservice.go
kevina: Fixed.

kevina:
@Kubuxu Okay, this is going to need to be rebased.  I'll handle that once I get the all clear.

@whyrusleeping does this Look-Good-To-You now?

kevina:
@whyrusleeping rebased and all test are green.  Does this look good to you.

blockservice/blockservice.go
```diff
@@ -23,77 +23,112 @@ var ErrNotFound = errors.New("blockservice: key not found")
...
+			}
+			if has {
+				continue
+			}
+			toput = append(toput, b)
 		}
-
-		toput = append(toput, b)
+	} else {
+		toput = bs;
```
whyrusleeping: semi colons??? noooo

whyrusleeping: I ran a gofmt and pushed it

PR #3301, opened 2016-10-12
Turn merkledag.Node into an interface

whyrusleeping:


whyrusleeping:
The astute reviewer may note that i'm pretty much cheating here and casting everything to a *ProtoNode. Deal with it for now.

jbenet:
@whyrusleeping you should get LGTMs from @Kubuxu @kevina @lgierth or others. merging yourself without others' review is the fastest way to introduce problems.

PR #3305, opened 2016-10-15
http api: makes sure header is sent even when r is not ready yet. fixes #3304

### LINKED ISSUE #3304: API call to pubsub/sub does not return HTTP header until first message arrives
This way calling `"go-ipfs-api".Request.Send()` blocks until the first message is received. Which is not nice.

keks:
Otherwise API clients block on pubsub subscriptions until the first packet is sent (see, well, #3304).

### DIFF
--- a/commands/http/handler.go
+++ b/commands/http/handler.go
@@ -288,6 +288,7 @@ func sendResponse(w http.ResponseWriter, r *http.Request, res cmds.Response, req
 		log.Error("err: ", err)
 		w.Header().Set(StreamErrHeader, sanitizedErrStr(err))
 	}
+
 }
 
 func flushCopy(w io.Writer, r io.Reader) error {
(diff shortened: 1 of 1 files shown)


commands/http/handler.go
```diff
@@ -298,6 +298,9 @@ func flushCopy(w io.Writer, r io.Reader) error {
 		return err
 	}
 	for {
+		// flush to send header when r is not ready yet
+		f.Flush()
```
whyrusleeping: probably want to check and handle the error here, just in case

keks:
I now check the error. Also I removed the flush at the bottom of the loop so it doesn't flush pointlessly.
Now it also flushes when the reader is drained (`err, n = r.Read();  err == io.EOF && n <= 0`)

whyrusleeping:
Tests are failing now... i'm guessing removing that final flush wasnt quite right

commands/http/handler.go
keks: Flush doens't return an error, that's why the tests failed :D

Kubuxu:
Can you squash it, we don't want to have any commits with build/test failures as it makes git bisecting much harder.

keks:
@Kubuxu uh, sorry. next time!

PR #3307, opened 2016-10-15
Implement 'Raw Node' node type

whyrusleeping:
This PR implements a 'Raw Node' dag node that is essentially just a block. Among other things, this allows us to make unixfs not have leaf nodes with extra framing and fit nicely inside multiples of 1024 and 4096.

A new flag `--raw-leaves` has been added to `ipfs add` that enables this functionality.

Do note that if you create objects with this flag, you will not be able to transfer them to other peers who are running any version before this (meaning those objects will not be able to be processed by ipfs 0.4.3 nodes).

This depends on the previous PR that turns the merkledag.Node into an interface.

### DIFF
--- a/core/corehttp/gateway_handler.go
+++ b/core/corehttp/gateway_handler.go
@@ -1,6 +1,7 @@
 package corehttp
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"io"
@@ -18,10 +19,10 @@ import (
 	path "github.com/ipfs/go-ipfs/path"
 	uio "github.com/ipfs/go-ipfs/unixfs/io"
 
-	"context"
 	routing "gx/ipfs/QmNUgVQTYnXQVrGT2rajZYsuKV8GYdiL91cdZSQDKNPNgE/go-libp2p-routing"
 	humanize "gx/ipfs/QmPSBJL4momYnE7DcUyk2DVhD6rH488ZmHBGLbxNdhU44K/go-humanize"
 	cid "gx/ipfs/QmXUuRadqDq5BuFWzVU6VuKaSjTcNm1gNCtLvvP1TJCW4z/go-cid"
+	node "gx/ipfs/QmZx42H5khbVQhV5odp66TApShV4XCujYazcvYduZ4TroB/go-ipld-node"
 )
 
 const (
@@ -45,7 +46,7 @@ func newGatewayHandler(node *core.IpfsNode, conf GatewayConfig) *gatewayHandler 
 }
 
 // TODO(cryptix):  find these helpers somewhere else
-func (i *gatewayHandler) newDagFromReader(r io.Reader) (*dag.ProtoNode, error) {
+func (i *gatewayHandler) newDagFromReader(r io.Reader) (node.Node, error) {
 	// TODO(cryptix): change and remove this helper once PR1136 is merged
 	// return ufs.AddFromReader(i.node, r.Body)
 	return importer.BuildDagFromReader(
@@ -353,7 +354,7 @@ func (i *gatewayHandler) putHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	var newnode *dag.ProtoNode
+	var newnode node.Node
 	if rsegs[len(rsegs)-1] == "QmUNLLsPACCz1vLxQVkXqqLX5R1X345qqfHbsf67hvA3Nn" {
 		newnode = uio.NewEmptyDirectory()
 	} else {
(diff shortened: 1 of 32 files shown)


kevina:
@whyrusleeping, what is the ETA for this.  (Ideally I would like my filestore code (#2634) to go in before this.)

Other than the usual conflicts my main concern is that this is implemented in a way the filestore can use it.  In order for that to happen the filestore needs to know what format the block is in.  If it is the old format it needs to decode it, if it is raw it needs to do nothing.

whyrusleeping:
@kevina you will be able to check the cid, or do a type switch on the object. It should actually make things easier for you

whyrusleeping:
@Kubuxu @lgierth  could you two review this for me?

kevina:
@whyrusleeping if possible please try not to rebase this, I merged this into kevina/filestore.

kevina:
@whyrusleeping I notice that neither "ipfs cat" or "ipfs get" work with raw nodes.  Can I take it that those parts are just not implemented yet?

Also (although it does not need to go into this pull request) I would suggest that the DagService GetLinks() method has a special case for Raw Nodes to avoid having to unnecessary retrieve a Raw Node when it is not required.

whyrusleeping:
Ohhhh... yeah. I should make sure cat and get work with raw nodes. Thats an issue.

kevina:
@whyrusleeping I can cat large files, but not small ones, for example:
```
kevina@intel:/aux/home/gocode/src/github.com/ipfs/go-ipfs$ ipfs add --raw-leaves Makefile
added zdvgq8kK8fHCAsT89ehxvm78PJqJj6FjHkPEMZTkARQnSQRyy Makefile
kevina@intel:/aux/home/gocode/src/github.com/ipfs/go-ipfs$ ipfs cat zdvgq8kK8fHCAsT89ehxvm78PJqJj6FjHkPEMZTkARQnSQRyy
Error: expected protobuf dag node
```

whyrusleeping:
@kevina fixed the add/cat of small files

test/sharness/t0040-add-and-cat.sh
```diff
@@ -392,17 +407,21 @@ test_add_named_pipe " Post http://$API_ADDR/api/v0/add?encoding=json&progress=tr
...
 test_add_cat_file
 
+test_add_cat_raw
+
 test_expect_success "ipfs add --only-hash succeeds" '
     echo "unknown content for only-hash" | ipfs add --only-hash -q > oh_hash
 '
 
-#TODO: this doesn't work when online hence separated out from test_add_cat_file
+#TODO: this doesnt work when online hence separated out from test_add_cat_file
```
whyrusleeping: I removed it because it messed up my syntax highlighting.

kevina: okay.

kevina:
@whyrusleeping have you tried this when HashOnRead() is true?  It looks like the verification code need updating.

whyrusleeping:
@kevina good catch, fixed! Anything else you can see?

kevina:
> good catch, fixed! Anything else you can see?

I am working on adding raw node support to the filestore, i let you know if I see anything else missing :)

Kubuxu:
I would feel much better about it if we run tests with it enabled by default.
I will do it when I get back home.

importer/helpers/helpers.go
```diff
@@ -124,9 +137,20 @@ func (n *UnixfsNode) SetData(data []byte) {
 	n.ufmt.Data = data
 }
 
+func (n *UnixfsNode) DataSize() uint64 {
```
kevina: Is there a reason you called this DataSize() and not FileSize()?  FileSize() means the size of the file _if this node was the root_.  In #3314 I added a similar helper method but I called it FileSize() and not DataSize().

whyrusleeping: Yeah, `FileSize` probably makes more sense, i'll change it.

whyrusleeping:
@Kubuxu running the tests with it enabled by default will be hard. All the hashes will be different.

core/corehttp/gateway_handler.go
```diff
@@ -417,8 +418,14 @@ func (i *gatewayHandler) putHandler(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 
+		pbnewnode, ok := newnode.(*dag.ProtoNode)
+		if !ok {
+			webError(w, "Cannot read non protobuf nodes through gateway", dag.ErrNotProtobuf, http.StatusBadRequest)
```
Kubuxu: As far as I can see it is put handler?
Does reading rawnodes through gateway work?

whyrusleeping: Yeah, reading raw nodes through the gateway uses the dagreader, which works fine. This nasty hacky area of code is the writeable gateway, which we don't really use...

kevina:
I have added this to the filestore and things seam to work.  I would say to go ahead and let this is even if there are still places that might brake with the raw nodes.

christynelson467:
Do you  understand
IDNK what to do ifni click on this conversation  inline or is itndireceted
to me????????

RichardLitt:
Hey @christynelson467  - thanks for the message. Please try and keep comments relevant to the conversation at hand. Open a new issue if you have any questions about go-ipfs. Thanks.

christynelson467:
fix whatever  is needed on my behalf
pre-approval on all needed for me to stay in guidelines

PR #3308, opened 2016-10-15
cli: refactor to expose argument parsing functionality

kevina:
This is just a small refactor to expose some functionality needed by the filestore (#2634).

Hopefully it is uncontroversial, if you need more an an explanation let me know.


whyrusleeping:
This LGTM

PR #3310, opened 2016-10-16
ds-help: add helper functions to convert from Cid to DsKey and the reverse

kevina:
In the filestore (#2634) i needed to convert from Cid to DsKey and the reverse in several places so I created two helper functions to that effect.

### DIFF
--- a/blocks/blockstore/blockstore.go
+++ b/blocks/blockstore/blockstore.go
@@ -87,7 +87,7 @@ func (bs *blockstore) Get(k *cid.Cid) (blocks.Block, error) {
 		return nil, ErrNotFound
 	}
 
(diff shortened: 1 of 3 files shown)


whyrusleeping:
Maybe these should be in the dshelper package?

kevina:
I take it you mean `thirdparty/ds-help`.  I thought about that, but it will add a dependency to that small package and also these methods are used in the same places the blockstore is used so I thought the blockstore package was a better place to avoid having yet another import.  I can move it though, just give the word.

whyrusleeping:
@kevina the thing is that your helper methods would be useful in nearly every place that that package is used.

kevina:
No problem, I'll move it.

kevina:
@whyrusleeping, moved

kevina:
@whyrusleeping tests passed.  Marking a RFM.

PR #3315, opened 2016-10-18
Fix metrics injection

Kubuxu:
I made last minute change last time and didn't test it properly.
This disabled metrics-interface as injection was being done after all collectors were being registered.

### DIFF
--- a/cmd/ipfs/daemon.go
+++ b/cmd/ipfs/daemon.go
@@ -181,6 +181,13 @@ func defaultMux(path string) corehttp.ServeOption {
 var fileDescriptorCheck = func() error { return nil }
 
 func daemonFunc(req cmds.Request, res cmds.Response) {
+	// Inject metrics before we do anything
+
+	err := mprome.Inject()
+	if err != nil {
+		log.Warningf("Injecting prometheus handler for metrics failed with message: %s\n", err.Error())
+	}
+
 	// let the user know we're going.
 	fmt.Printf("Initializing daemon...\n")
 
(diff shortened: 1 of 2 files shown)


cmd/ipfs/daemon.go
```diff
@@ -181,6 +181,13 @@ func defaultMux(path string) corehttp.ServeOption {
 var fileDescriptorCheck = func() error { return nil }
 
 func daemonFunc(req cmds.Request, res cmds.Response) {
+	// Inject metrics before we do anything
+
+	err := mprome.Inject()
+	if err != nil {
+		log.Warningf("Injecting prometheus handler for metrics failed with message: %s\n", err.Error())
```
whyrusleeping: should this just be a warning? Or should we actually error out here?

Also, can we (in a future PR) have a flag to disable metrics entirely?

Kubuxu: Yes we can have flag, lacking metrics won't hurt working of the daemon so IMO it is better to continue.

If you want I can do Error log here instead.

cmd/ipfs/daemon.go
```diff
@@ -181,6 +181,13 @@ func defaultMux(path string) corehttp.ServeOption {
 var fileDescriptorCheck = func() error { return nil }
 
 func daemonFunc(req cmds.Request, res cmds.Response) {
+	// Inject metrics before we do anything
+
+	err := mprome.Inject()
+	if err != nil {
+		log.Errorf("Injecting prometheus handler for metrics failed with message: %s\n", err.Error())
```
lgierth: How will :5001/debug/metrics/prometheus behave if Inject() fails?

Kubuxu: It will miss the metrics that are being added with metrics-interface and that is it.

Without this patch it runs as if the Inject has failed.

PR #3321, opened 2016-10-19
make path resolver no longer require whole node for construction

whyrusleeping:
This is some preliminiary work before we can get sharding in. Resolving a path now requires a certain bit of context and logic around what the path actually refers to (unixfs vs normal dag traversal)

License: MIT


whyrusleeping:
cc @Kubuxu @lgierth for review

christynelson467:
I turned my miner on is all  okay?

Kubuxu:
LGTM

PR #3324, opened 2016-10-24
allow cid format selection in block put command

whyrusleeping:
License: MIT

### DIFF
--- a/core/commands/block.go
+++ b/core/commands/block.go
@@ -2,7 +2,6 @@ package commands
 
 import (
 	"bytes"
-	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


whyrusleeping:
@Kubuxu what do you mean?

Kubuxu:
You had to change the hashes in the tests, which means that hashes outputted by `ipfs block` API have changed.

whyrusleeping:
Ah, yeah. Youre referring to setting the default format value to be "v0". That makes sense to me.

whyrusleeping:
@Kubuxu there, now the default behaviour doesnt change. Good catch

PR #3325, opened 2016-10-24
Implement cbor ipld nodes and a first pass at the 'dag' command

whyrusleeping:
Alright, this is ready for some review now

License: MIT

### DIFF
--- a/package.json
+++ b/package.json
@@ -278,9 +278,9 @@
     },
     {
       "author": "whyrusleeping",
-      "hash": "QmY7L2aEa1rHjkSSbXJB8oC7825JTpUUvDygmM2JPQeqhr",
+      "hash": "QmRcAVqrbY5wryx7hfNLtiUZbCcstzaJL7YJFBboitcqWF",
       "name": "go-ipld-cbor",
-      "version": "0.2.2"
+      "version": "0.3.0"
     }
   ],
   "gxVersion": "0.4.0",
--- a/package.json
+++ b/package.json
@@ -278,9 +278,9 @@
     },
     {
       "author": "whyrusleeping",
-      "hash": "QmYRzW9YDHVNCDbfFzbS7TEXAG1swE1yjq1basZ5WnJYH4",
+      "hash": "QmY7L2aEa1rHjkSSbXJB8oC7825JTpUUvDygmM2JPQeqhr",
       "name": "go-ipld-cbor",
-      "version": "0.2.1"
+      "version": "0.2.2"
     }
   ],
   "gxVersion": "0.4.0",
--- a/package.json
+++ b/package.json
@@ -277,9 +277,10 @@
       "version": "0.3.2"
     },
     {
-      "hash": "QmQKEgGgYCDyk8VNY6A65FpuE4YwbspvjXHco1rdb75PVc",
-      "name": "go-libp2p-routing",
-      "version": "2.2.2"
+      "author": "whyrusleeping",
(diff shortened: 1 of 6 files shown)


whyrusleeping:
@Kubuxu @lgierth @kevina @hsanjuan Could you guys give me a little review on this one?

Also worth taking a look at the cbor ipld package here: https://github.com/ipfs/go-ipld-cbor

kevina:
@whyrusleeping I'm afraid I have not been following the IPLD very closely.  What exactly is a "cbor ipld node"? In particular what does "cbor" stand for or at least mean.

package.json
```diff
@@ -277,9 +277,10 @@
       "version": "0.3.2"
     },
     {
-      "hash": "QmQKEgGgYCDyk8VNY6A65FpuE4YwbspvjXHco1rdb75PVc",
-      "name": "go-libp2p-routing",
-      "version": "2.2.2"
```
kevina: So "go-libp2p-routing"? is no longer used?

kevina:
@whyrusleeping okay so scratch by filestore notes, it doesn't look like there is enough implemented yet for this node type to make its way to the filestore.  I discovered that CBOR = Concise Binary Object Representation (RFC 7049), but I am even more confused on how this all fits in.  I ping you on IRC tomorrow.  (Sorry for all the noise).

package.json
whyrusleeping: Oh, i guess it still is. But its imported indirectly via the go-libp2p-kad-dht, meaning that me accidentally removing this didnt break anything.

christynelson467:
I see that  every  transaction was removed due to yall  burning a contract
in my file manager  qnd nor upfront advising me of gas charges.  I've
looked and myll options and request  to be removed from all chats in
regards to this matter. others  were  paid off those  same transactions and
I  no longer give  concent to you  access any of my devices.  thank  you

hsanjuan:
@kevina @whyrusleeping actually I appreciate the noise because I have similar questions. Did you discuss on IRC already? It would be good if I can read it too :)

whyrusleeping:
@hsanjuan What questions do you have? I can provide answers here as its a better place for others to find the info

hsanjuan:
@whyrusleeping so CBOR is just another serializing format and this would be used to store IPLD nodes locally, where now Protobuf is used?

Since you asked for review, I'll tell you that you forgot to document most things in go-ipld-cbor (https://godoc.org/github.com/ipfs/go-ipld-cbor) :) 

Shouldn't the JSON->CBOR functions be part of `go-ipld-cbor` too? Since that seems to offer CBOR->JSON already?

Kubuxu:
@hsanjuan protobuf is only for what is used currently which is unixfs, in future we might upgrade unixfs to IPLD but so far it is not planned as far as I know.

Kubuxu:
LGTM, the conversion functions are a bit not nice but it is to be expected.

lgierth:
LGTM, just one question: why two formats for each of cbor or and protobuf, and the possible inconsistency between "dag-pb" and "protobuf"?

whyrusleeping:
@lgierth both dag-pb and protobuf are the same thing. Just different names for it. I've traditionally called it just protobuf, but @diasdavid insists on calling it dag-pb.

If you want we can just pick one, but i'm a fan of 'being liberal in what you accept'

lgierth:
> If you want we can just pick one, but i'm a fan of 'being liberal in what you accept'

cool I was just wondering :+1:

whyrusleeping:
Well, if i merge this now, i will have successfully met the goal we set for ipld integration. I've got a few LGTMs and this command is marked as experimental. Lets do this.

PR #3339, opened 2016-10-31
update go-libp2p-swarm with deadlock fixes

whyrusleeping:
License: MIT
Signed-off-by: Jeromy why@ipfs.io


whyrusleeping:
This should address the goroutine buildup we've been seeing on some of our gateway machines @lgierth

PR #3343, opened 2016-11-01
Fix/mfs/cleanup

polezaivsani:
Just a few minor fixes as i went learning the code

### DIFF
--- a/mfs/system.go
+++ b/mfs/system.go
@@ -1,12 +1,6 @@
 // package mfs implements an in memory model of a mutable ipfs filesystem.
 //
-// It consists of four main structs:
-// 1) The Filesystem
-//        The filesystem serves as a container and entry point for various mfs filesystems
-// 2) Root
-//        Root represents an individual filesystem mounted within the mfs system as a whole
-// 3) Directories
-// 4) Files
+// It's made up out of roots, directories and files.
 package mfs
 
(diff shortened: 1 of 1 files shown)


mfs/system.go
```diff
@@ -1,12 +1,6 @@
 // package mfs implements an in memory model of a mutable ipfs filesystem.
 //
-// It consists of four main structs:
-// 1) The Filesystem
```
whyrusleeping: Granted, this comment wasnt entirely accurate anymore, but i would like to have something at least as informative in its place

polezaivsani: This has just the short description what root is missing, but it's on the type itself. Thought that'll be one less chance of comments and code going out of sync.

Still want this back?

polezaivsani: The the whole thing may not warrant the attention, but still - you've seen the comment on line 3 - doesn't it serve as a good enough overview sans obsolete bits?

whyrusleeping:
The purpose for that comment was to make pages like this (which are autogenerated)  look nice: https://godoc.org/github.com/ipfs/go-ipfs/mfs

It currently doesnt look very nice, but i do want to keep that top section informative at a glance.

Kubuxu:
Closing due to lack of activity.

PR #3345, opened 2016-11-01
namesys: return right after errors

whyrusleeping:
Previously, if this errored, the possibility for goroutines to get stuck sending a second result down the channel existed. This should prevent that from happening.

License: MIT


lgierth:
LGTM - shouldn't it have been crashing on `proto.Unmarshal(nil)`?

whyrusleeping:
@lgierth Huh, maybe. could be that we never actually hit this?

PR #3346, opened 2016-11-01
dht: update to dht code with fixed GetClosestPeers

whyrusleeping:
should make dht resolutions faster.

ref libp2p/go-libp2p-kad-dht#20

License: MIT


lgierth:
LGTM -- let's do it

PR #3348, opened 2016-11-03
Separate out the G.C. Locking from the Blockstore interface.

kevina:
Factored out of #3257 (Add support for multiple blockstores).

License: MIT


whyrusleeping:
This LGTM.

PR #3352, opened 2016-11-04
allow listing of streams to connected peers

whyrusleeping:


whyrusleeping:
@diasdavid @dignifiedquire let me know when you update js-ipfs-api

PR #3353, opened 2016-11-04
respect contexts when returning diagnostics responses

whyrusleeping:
Noticed a few of these building up in long running stack dumps. Easy enough fix.

License: MIT


Kubuxu:
I read it before @whyrusleeping merged it, forgot to LGTM. Sorry.

PR #3357, opened 2016-11-04
no longer hard code gomaxprocs

whyrusleeping:
This hasnt been needed since go started auto-detecting the correct number of processors to use.

License: MIT

### DIFF
--- a/cmd/ipfs/main.go
+++ b/cmd/ipfs/main.go
@@ -10,7 +10,6 @@ import (
 	"net/url"
 	"os"
 	"os/signal"
-	"runtime"
 	"runtime/pprof"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


lgierth:
LGTM -- it would still be nice to override this with an environment variable, but this PR is fine

djdv:
@lgierth 
I believe you can set the environment variable `GOMAXPROCS` to change this at runtime, I can't seem to find a better reference than this though:
>GOMAXPROCS can be set explicitly using the GOMAXPROCS environment variable or by calling runtime.GOMAXPROCS from within a program.

https://docs.google.com/document/d/1At2Ls5_fhJQ59kDK2DFVhFu3g5mATSXqqV5QrxinasI/view

Kubuxu:
LGTM

lgierth:
awesome then

Issue #3336, opened 2016-10-29
HTTP API Add Crashes Daemon


nycoliver:
#### Version information:

```
go-ipfs version: 0.4.5-dev-17c629d
Repo version: 4
System version: amd64/darwin
Golang version: go1.7.3
```

#### Type:

Bug

#### Priority:

P3
#### Description:

HTTP API requests almost always fail with the new go-ipfs. They occasionally work, but most of the time daemon throws the following errors. After enough (~100) requests, the daemon consistently crashes.

```
17:07:01.242 ERROR commands/h: err: multipart: NextPart: EOF handler.go:288
17:07:01.474 ERROR commands/h: err: multipart: NextPart: http: invalid Read on closed Body handler.go:288
17:07:01.626 ERROR commands/h: err: multipart: Part Read: http: invalid Read on closed Body handler.go:288
17:07:01.803 ERROR commands/h: err: multipart: NextPart: http: invalid Read on closed Body handler.go:288
17:07:02.212 ERROR commands/h: err: multipart: NextPart: http: invalid Read on closed Body handler.go:288
panic: runtime error: index out of range

goroutine 2946 [running]:
panic(0x7e1ba0, 0xc4200100f0)
    /usr/local/Cellar/go/1.7.3/libexec/src/runtime/panic.go:500 +0x1a1
github.com/ipfs/go-ipfs/core/coreunix.(*Adder).Finalize(0xc42292d200, 0xc420078020, 0xcfb3c0, 0xc420078020, 0xc420078001)
    /Users/o/Code/gocode/src/github.com/ipfs/go-ipfs/core/coreunix/add.go:198 +0x2cb
github.com/ipfs/go-ipfs/core/commands.glob..func6.1(0xd0a3a0, 0xc4236b7500, 0xc4236b7500, 0x41bf8)
    /Users/o/Code/gocode/src/github.com/ipfs/go-ipfs/core/commands/add.go:211 +0x136
github.com/ipfs/go-ipfs/core/commands.glob..func6.2(0xc423673c20, 0xc4240f5380, 0xd0e340, 0xc423c6c780, 0xd0d700, 0xc42361c930)
    /Users/o/Code/gocode/src/github.com/ipfs/go-ipfs/core/commands/add.go:225 +0x7c
created by github.com/ipfs/go-ipfs/core/commands.glob..func6
    /Users/o/Code/gocode/src/github.com/ipfs/go-ipfs/core/commands/add.go:230 +0x7fb
```

[...]

PR #3366, opened 2016-11-07
Revert "http api: makes sure header is sent even when r is not ready …

### LINKED ISSUE #3304: API call to pubsub/sub does not return HTTP header until first message arrives

keks:
…yet. fixes #3304 (#3305)"

This reverts commit 68d8a298c58ed950bd44118cc5e6d1ffc2def395.

This broke ipfs add from js-ipfs-api. Related discussion on the commit page and at #3336.


whyrusleeping:
thanks @keks. That was a really great bit of deep diving research, we're definitely going to have to find a better way forward for the api

PR #3368, opened 2016-11-08
WIP: Basic Filestore implementation.

kevina:
@whyrusleeping this is not 100% what you want but it is close.  It still uses bits of the multi-blockstore but doesn't use any of the parts you don't like.

Things that still need to be done (for this part):
- [ ] Implement DeleteBlock for the filestore
- [ ] Config option to enable filestore (it is now always enabled due to the complete remove of `ipfs filestore` commands, including `ipfs filestore enable`)
- [ ] Implement a correct `Has` method that checks to make sure the content is still available

Things you may still not like:
- "ipfs add" will not add blocks if any of them happen to be in the filestore

Things I don't like:
 - way to many, most of which needs to be in separate issues.

Others: see #2634 for the full implementation.


filestore/dataobj.go
```diff
@@ -0,0 +1,136 @@
...
+
+type DataObj struct {
+	Flags uint64
+	// The path to the file that holds the data for the object, an
+	// empty string if there is no underlying file
+	FilePath string
+	Offset   uint64
+	Size     uint64
+	ModTime  float64
+	Data     []byte
```
mguentner:
```
Execute bool
Write   bool
```

should be added here.

Explanation:

- **execute bit** whether the execute bit will be set for the OS (fuse)
- **write bit** whether it is possible to alter the content of the file (thus recreating a new version of it) through fuse, it can be set using e.g. `chmod`

whyrusleeping:
@jbenet You mentioned the other day that you wanted to review this. If you find time soon to do so that would be great.

kevina:
@jbenet I am not sure what @whyrusleeping told you but please note that this is only a small fraction of the full filestore code with very little of the functionality of #2634.  I do not consider this very useful, as it doesn't even provide a way to list what is in the filestore.  We also had some rather strong disagreements that I have not been able to write (publicly) about yet.

The test cases are also not as comprehensives as I would like as most of the test cases rely on functionality not in this pull request.

whyrusleeping:
#3629 was merged in favor of this one.Closing now

kevina:
@Kubuxu can we move the branch kevina/filestore-basic to go-ipfs-archived?

Kubuxu:
It will be moved in a month.

PR #3369, opened 2016-11-08
bash_completion: Add a couple missing commands

nathanmusoke:
Add "bitswap" and "repo" to bash completion for `ipfs stats`. They are
listed under `ipfs stats --help`.

Also added their respective subcommands.

License: MIT
Signed-off-by: Nathan Musoke <nathan.musoke@gmail.com>

### DIFF
--- a/misc/completion/ipfs-completion.bash
+++ b/misc/completion/ipfs-completion.bash
@@ -314,12 +314,22 @@ _ipfs_resolve()
 
 _ipfs_stats()
 {
-    _ipfs_comp "bw --help"
+    _ipfs_comp "bitswap bw repo --help"
+}
+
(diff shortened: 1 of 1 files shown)


nathanmusoke:
I noticed a couple commands missing from the bash completion when playing with ipfs. It wasn't obvious to me whether this was intentional, but it didn't seem to be.

If it would be helpful, I'd be happy to go through the rest of the commands and check if any others are missing.

whyrusleeping:
@nathanmusoke thanks for the PR. we don't actively maintain the bash completions so you will very likely find other commands that are not included. For a full list of commands (for future reference) take a look at the output of `ipfs commands`

PR #3377, opened 2016-11-13
Updated Bash complete script

EliasGabrielsson:
Some minor bug still exist but not breaking.
Everything is marked with "TODO" inline the script.

License: MIT
Signed-off-by: Elias Gabrielsson <elias@benefactory.se>

### DIFF
--- a/misc/completion/ipfs-completion.bash
+++ b/misc/completion/ipfs-completion.bash
@@ -1,6 +1,10 @@
 _ipfs_comp()
 {
     COMPREPLY=( $(compgen -W "$1" -- ${word}) )
+    if [[ ${#COMPREPLY[@]} == 1 && ${COMPREPLY[0]} == "--"*"=" ]] ; then
+        # If there's only one option, with =, then discard space
+        compopt -o nospace
+    fi
 }
 
 _ipfs_help_only()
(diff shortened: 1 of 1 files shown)


EliasGabrielsson:
Replacing old pull request. @Kubuxu

Kubuxu:
Hmm, zsh has bash completion compat, which means that I can give it actual use test. I will use it for 1-2 days and get back to you.

Thanks again.

Kubuxu:
Looks like it doesn't work well for complex completions so I will just run it in bash and test out.

Kubuxu:
LGTM, only thing that I observed is that it doesn't support short options `-r` vs `--recursive` and so on but either way it is much much better than current completion.

Thank you very much.

@whyrusleeping good for merge.

EliasGabrielsson:
I have intentionally not implemented short option. It's not a big deal to implement if its needed.

Kubuxu:
Nah, I think in the bash's completion method it might cause more confusion than good.

whyrusleeping:
Alright cool, will be good to finally have this in.

EliasGabrielsson:
Under the development of the complete-script I have wonder how the script could be installed in an easy fashion way. 

`ipfs autocomplete install` - or similar,
a) Search for current shell, eg. `bash` / `zsh`. 
b) Search if there is a script available _somewhere_. How does `ipfs update` work, does it fetch things from github or by ipfs itself? Is version mapped to github versions?
c) Install script on default location based on OS and profilesettings. This can be tricky but I think its do able. An option could be to specify install location.

The `ipfs-install` could also keep track if the script have been installed and autoupdate it when ipfs gets updated by `ipfs-install`

What do you think?

Kubuxu:
Usually it is package mangers that do this and I think it is good enough.

PR #3384, opened 2016-11-14
blockstore: remove expensive debug statement in AllKeysChan

### LINKED ISSUE #3378: Remove Expensive log.Debuf(...) statement in blockstore.AllKeysChan?
[This statement](https://github.com/ipfs/go-ipfs/blob/10915f346af4f63f38749015eb10273f447786b9/blocks/blockstore/blockstore.go#L205) is causing a major slowdown in AllKeysChan,
```
	get := func() (*cid.Cid, bool) {
		...
		log.Debug("blockstore: query got key", c)
		...
	}
```
due to the fact that the Cid is converted to a string which is very expensive due to the conversion to the Base58 representation.  This is done even if the log.Debug does not output anything.  See #3376.

The question is what to do about it?  The easiest thing would be to just remove it or comment it out with a explanation to why.

The other alternative is to only enable it when Debug mode is activated, but this is not the same as the Debug logging level.

I can't seam to find a way to check what the logging level is, if there was I could just use that.

kevina:
Closes #3378 

License: MIT

### DIFF
--- a/blocks/blockstore/blockstore.go
+++ b/blocks/blockstore/blockstore.go
@@ -202,8 +202,6 @@ func (bs *blockstore) AllKeysChan(ctx context.Context) (<-chan *cid.Cid, error) 
 				return nil, true
 			}
 
-			log.Debug("blockstore: query got key", c)
-
 			return c, true
 		}
 	}


whyrusleeping:
just waiting on tests now...

kevina:
@whyrusleeping looks like the tests passes (except js-ipfs-api which I don't see how this change could break).

lgierth:
@kevina it looks unrelated, I filed ipfs/js-ipfs-api#424 for it

whyrusleeping:
LGTM, :ship:

PR #3386, opened 2016-11-15
Use NextSync method in Datastore

kevina:
See ipfs/go-datastore#53 ipfs/go-datastore#54

### DIFF
--- a/blocks/blockstore/blockstore.go
+++ b/blocks/blockstore/blockstore.go
@@ -181,44 +181,27 @@ func (bs *blockstore) AllKeysChan(ctx context.Context) (<-chan *cid.Cid, error) 
 		return nil, err
 	}
 
-	// this function is here to compartmentalize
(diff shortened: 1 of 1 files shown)


blocks/blockstore/blockstore.go
```diff
@@ -186,8 +186,9 @@ func (bs *blockstore) AllKeysChan(ctx context.Context) (<-chan *cid.Cid, error)
 		select {
 		case <-ctx.Done():
```
whyrusleeping: since we're checking the context at the send layer (and this is in a goroutine, and not expected to really block) we can do away with the select and context check entirely here.

kevina: so just use something like:
```
for {
  e, ok := res.NextSync()
  if !ok { return nil, false }
  ...
}
```

whyrusleeping: No need for a for loop there either i think

PR #3389, opened 2016-11-15
configuration for labels-and-milestones sync tool

flyingzumwalt:
I'm not convinced that this is where we want to store these config files for each project. I just set it up to imitate what @haadcode did in https://github.com/haadcode/orbit/tree/master/tools/roadmap-generator

This adds the labels-and-milestones sync tool with configs for the go-ipfs Q4 roadmap.

For more info, see ipfs/pm#257

### DIFF
--- a/tools/labels-and-milestones/.gitignore
+++ b/tools/labels-and-milestones/.gitignore
@@ -1,0 +1,1 @@
+node_modules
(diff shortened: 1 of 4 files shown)


whyrusleeping:
closing as per @flyingzumwalt's instructions. Branch will be kept.

PR #3390, opened 2016-11-15
Roadmap generator and Generated Q4 Roadmap

flyingzumwalt:
* adds the roadmap-generator tool with configs for the go-ipfs Q4 roadmap.
* adds the generated ROADMAP.md

**Please check that the ROADMAP.md is accurate before merging!**

As with #3389, I'm not convinced that this is where we want to store these config files for each project. I just set it up to imitate what @haadcode did in https://github.com/haadcode/orbit/tree/master/tools/roadmap-generator

For more info, see ipfs/pm#257

### DIFF
--- a/roadmap.md
+++ b/roadmap.md
@@ -1,67 +1,72 @@
-# go-ipfs Q2 roadmap
-NOTE: No specific order is implied in this list.
+# go-ipfs - Roadmap
 
-- [ ] fix random test failures
-  - check issues for label 'test_failure'
-- [ ] Improve memory usage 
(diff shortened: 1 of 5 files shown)


whyrusleeping:
Sweet! This is looking good, At least one part of this still says Orbit.

I'm also not convinced that we should be keeping these committed to git, something like `go-ipfs.ipfs.team` might be a cool place for it. Or `roadmaps.ipfs.team/go-ipfs`. That way we can have it autogenerated and pushed up there once a day without having all the noise of comitting it to here.

whyrusleeping:
closing as per @flyingzumwalt's instructions. Branch will be kept.

PR #3391, opened 2016-11-16
update to newer ipld node interface with Copy and better Tree

whyrusleeping:
License: MIT


lgierth:
LGTM -- js-ipfs-api failure is due to ipfs/js-ipfs-api#424

Issue #3313, opened 2016-10-17
Private Networks - First Implementation


Kubuxu:
As private networks are adoption blockers for and few important cases were raised for them, I would like to describe to you our first take on Private Networks.

For the first implementation only Pre-shared Key functionality will be available, given the timeframe and complexity Public Key Infrastructure would be just too complex and needs more talk and preparation. PSK based approach will unblock some of the users and allow them to start using IPFS.

The key will be just one per instance of libp2p Swarm, and thus one per IPFS daemon. This clears up corner cases of "what if the node connects to private and public network at the same time".

We wanted to make it key-per-listen-address but it would be too complex.

---

The crypto was changed but it will be symmetric encryption.

---

The PSK will be stored inside `swarm.key` (if there is better place for it, please comment) file in repo with text based multicoded, occupying `/key/swarm/psk/1.0.0/` as a name. Following it is a second text based multicodec defining base encoding of the following data, thus allowing the keys to be either typed easily manually (base16, base58), transferred over clipboard and shell without problems (base64) or can be also purely binary if you wish so.

Example:

```
 > cat $IPFS_PATH/swarm.key
/key/swarm/psk/1.0.0/
/b16/
FFFF
```

This is of course very weak key but it works as an example.

---

This feature is planned to be done at the end of the November, but it might be earlier. It is already in progress, testing it will take a while.

Part of: https://github.com/ipfs/go-ipfs/issues/1633
Resolves partially: https://github.com/ipfs/notes/issues/146
Relates to: https://github.com/ipfs/go-ipfs/issues/1633

Kubuxu:
//cc @whyrusleeping

whyrusleeping:
cc @jbenet for review. This is the initial implementation that @Kubuxu and I have been discussing.

cpacia:
> but is secio specific, when we move away from it we will have to rethink it.

What is the plan to move away from secio?

Kubuxu:
@cpacia to keep things on topic in this issue I will link an issue from this comment when we have clean writeup about this.

jbenet:
@cpacia the gist is we'll support TLS 1.3 in the future.

jbenet:
@Kubuxu: 

[...]

PR #3396, opened 2016-11-17
Private network implementation

Kubuxu:
Resolves https://github.com/ipfs/go-ipfs/issues/3313

~~Almost done, I still want to create file transfer tests.~~

File transfer tests are down waiting for final CR.


Kubuxu:
@whyrusleeping it should be ready for CR.

repo/fsrepo/fsrepo.go
```diff
@@ -578,6 +581,26 @@ func (r *FSRepo) GetStorageUsage() (uint64, error) {
 	return du, err
 }
 
+func (r *FSRepo) SwarmKey() ([]byte, error) {
```
hsanjuan: Shall the swarm key be part of the ipfs configuration file (as the host identity is?). My first impression is that it would be a more consistent approach.

Kubuxu: No, as we want to move the privkey out of the config for a long time already.

Kubuxu: Config is exposed via the HTTP API and right now we some ugly hacks to not expose the privkey over the API.

Kubuxu: It could be part of keystore but as keystore isn't scheduled yet it will have to work for now. In future we can migrate it into keystore.

hsanjuan: ok

test/sharness/t0180-private-network.sh
```diff
@@ -0,0 +1,135 @@
...
+set_key 4 key2
+
+unset LIBP2P_FORCE_PNET
+
+test_expect_success "start nodes" '
+	iptb start [0-4]
+'
+
+test_expect_success "try connecting node in public network with priv networks" '
+	iptb connect [1-4] 0
```
whyrusleeping: huh, maybe iptb should return failure if one of those doesnt succeed?

Kubuxu: they all fail right now and iptb doesn't return anything.

repo/fsrepo/fsrepo.go
lgierth: Word, it would be great if we could move the private key out of the config file in a similar fashion :+1:

test/sharness/t0180-private-network.sh
```diff
@@ -0,0 +1,135 @@
...
+	[ $(ipfsi 0 swarm peers | wc -l) -eq 0 ]
+'
+
+test_expect_success "try connecting nodes in different private networks" '
+	iptb connect 2 3
+'
+
+test_expect_success "node 3 (pnet 2) swarm is empty" '
+	ipfsi 3 swarm peers &&
+	[ $(ipfsi 3 swarm peers | wc -l) -eq 0 ]
```
lgierth: Should this test wait a bit to make sure it *really* didn't connect?

Kubuxu: The iptb connect command waits for the connect call to finish, by then it is either for sure connected or not.

test/sharness/t0180-private-network.sh
```diff
@@ -0,0 +1,135 @@
...
+# MIT Licensed; see the LICENSE file in this repository.
+#
+
+test_description="Test private network feature"
+
+. lib/test-lib.sh
+
+test_init_ipfs
+
+export LIBP2P_FORCE_PNET=1
```
hsanjuan: I'd  find very useful that there was a similar option in the ipfs config file. `"EnablePrivateNetworks"` kind of thing. Also, an option to indicate the path to the key?

whyrusleeping: Agreed with this, but i'm wanting to avoid using config values for features that are currently experimental. Once private networks are more well defined, we can make it an official feature and give it config values to use

whyrusleeping:
@Kubuxu what is the status on the review of the private net crypto code?

Kubuxu:
@jbenet told me that he will review it tomorrow, it was 3 days ago.

whyrusleeping:
@Kubuxu can we close this now that the other PR merged?

Kubuxu:
Yes

PR #3401, opened 2016-11-21
(WIP) Feat/floodsub-unsubscribe

haadcode:
This PR will add pubsub/unsub (unsubscribe) command to manually cancel subscriptions.

This PR is connected to https://github.com/libp2p/go-floodsub/pull/12.

I'm still unsure what the best approach is here for cancelling the subscriptions as a whole, but this works for me to move forward with js-ipfs-api integration (ie. I can test it).

I don't expect this to be merged necessarily, but leaving this here for reference and future discussions.

### DIFF
--- a/commands/http/handler.go
+++ b/commands/http/handler.go
@@ -147,6 +147,7 @@ func (i internalHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 			select {
 			case <-clientGone:
 			case <-ctx.Done():
+				log.Debug("connection closed: %s", r.URL)
 			}
 			cancel()
 		}()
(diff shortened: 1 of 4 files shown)


haadcode:
cc @whyrusleeping @keks

keks:
Hmm, I'm not sure why this would be necessary. You should have cancelled the subscription by the end of `PubsubSubCmd.Run()`.

haadcode:
From IRC:

< haad> keks: so on the pubsub issue: when this ctx.Done() happens, the http connection to the caller has been closed, and so the caller gets a "connection closed" on their side. however, at that point, Floodsub.Unsub(topic) hasn't been called, so within floodsub (the system state) it might still think that it's subscribed to the topic. so what happens then is that if that same caller tries to subscribe to the same topic again, they'll get "already subscribed" even though the connection has been closed.

< haad> keks: oh, by "this ctx.Done()" I mean this: https://github.com/ipfs/go-ipfs/pull/3401/files#diff-e554e0645e58976ca475d8e0f7728fe1R105

< keks> haad: i see. well the good news is that "already subscribed" is a thing of the past. also i think go-ipfs#3402 cleans up after itself, telling floodsub to cancel the subscription

< haad> keks: it's already "cleaning up", the Floodsub.Unsub(topic) on the line below does that. the problem is that the API caller doesn't know that...

< haad> keks: ...or rather that it happens "async" and after telling the caller API the connection has been closed

haadcode:
@keks your other PR fixes all the problems I had that made me make this PR. We can close this.

PR #3416, opened 2016-11-23
Send empty struct to pubsub cmd output to flush

keks:
So the HTTP headers get sent

### DIFF
--- a/core/commands/pubsub.go
+++ b/core/commands/pubsub.go
@@ -6,6 +6,7 @@ import (
 	"encoding/binary"
 	"fmt"
 	"io"
+	"strings"
 	"sync"
 	"time"
 
@@ -89,10 +90,13 @@ To use, the daemon must be run with '--enable-pubsub-experiment'.
 		go func() {
 			defer sub.Cancel()
 			defer close(out)
+
+			out <- floodsub.Message{}
+
 			for {
 				msg, err := sub.Next(req.Context())
 				if err == io.EOF || err == context.Canceled {
-					break
+					return
 				} else if err != nil {
 					res.SetError(err, cmds.ErrNormal)
 					return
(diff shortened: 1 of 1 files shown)


core/commands/pubsub.go
```diff
@@ -118,16 +122,33 @@ To use, the daemon must be run with '--enable-pubsub-experiment'.
...
+
 			m.Data = append(m.Data, '\n')
 			return bytes.NewReader(m.Data), nil
 		}),
 		"lenpayload": getPsMsgMarshaler(func(m *floodsub.Message) (io.Reader, error) {
 			buf := make([]byte, 8)
-			n := binary.PutUvarint(buf, uint64(len(m.Data)))
-			return io.MultiReader(bytes.NewReader(buf[:n]), bytes.NewReader(m.Data)), nil
+			var (
+				n    int
```
whyrusleeping: I dislike this syntax, lets just instantiate things where they are used.

Also, no need for the `m.Message != nil` check, its functionally equivalent to just saying `data := m.Message`

core/commands/pubsub.go
```diff
@@ -118,16 +122,33 @@ To use, the daemon must be run with '--enable-pubsub-experiment'.
 	},
 	Marshalers: cmds.MarshalerMap{
 		cmds.Text: getPsMsgMarshaler(func(m *floodsub.Message) (io.Reader, error) {
+			if m.Message == nil {
```
whyrusleeping: You can leave this out too, passing a `nil` byte array to a `bytes.NewReader` will result in the same as passing the empty string to a stringsreader

core/commands/pubsub.go
```diff
@@ -118,16 +122,33 @@ To use, the daemon must be run with '--enable-pubsub-experiment'.
...
+				return strings.NewReader(""), nil
+			}
+
 			return bytes.NewReader(m.Data), nil
 		}),
 		"ndpayload": getPsMsgMarshaler(func(m *floodsub.Message) (io.Reader, error) {
+			if m.Message == nil {
```
whyrusleeping: Same here, `append([]byte(nil), '\n')` works just fine

core/commands/pubsub.go
keks: No I can't. We use `m.Data` later, which is `m.Message.Data`, which is a nil dereference.

core/commands/pubsub.go
keks: The idea here is that `data` is nil if `m.Message` is nil.

keks:
Oh and `m.Message` is nil when sending the empty message

core/commands/pubsub.go
whyrusleeping: Yeap, i know

core/commands/pubsub.go
whyrusleeping: If `m.Message` is nil, what is the value of `m.Data` ?

keks: `Message` is embedded and `Data` is part of `Message`.

keks: and if `m.Message` is nil, `m.Message.Data` panics

core/commands/pubsub.go
keks: And where would I store that? `m.Data` does not exist because it is actually `m.Message.Data`, and `m.Message` is `nil`. Panic!

whyrusleeping:
@keks sorry about that, I missed the fact that there was an embedded struct involved.

haadcode:
I endorse this PR and/or service. Would be awesome to get this into 0.4.5-rc1 build.

keks:
@whyrusleeping can you merge? I'll look at sharness tests next week.

Issue #3419, opened 2016-11-24
ipfs pin ls <hash> fetches the object from the network


hsanjuan:
#### Description:
`ipfs pin ls` seems by all means a local operation, but when the daemon is running and a `<ipfs_hash>` argument is passed, it will request blocks from the network when it does not know the hash. i.e:

> ipfs pin ls QmXja3KdmLxxS5F7SxASEB1Kg9AbkQ9PZpF67hyauHEgZa

This has side effects like:
  * Hanging when a hash is not pinned instead of simply showing that it is not pinned locally
  * Fetching objects that the user don't expect to be fetched (and thus [probably] making them available to others). Really bad if you're checking for content you don't want to have. Also, it happens silently.
  * If a user wants to check if a hash is pinned, it forces the user to do `ipfs pin ls` and then filter out the results, as `pin ls` without arguments seems to run locally entirely.

PR #3420, opened 2016-11-24
Issue #3419: Do not recursively resolve DAG just to obtain a CID

hsanjuan:
Resolve() tries to fetch the object to check if it has to recursively do something with children (I think). It's overkill here (I think). See #3419

### DIFF
--- a/core/commands/pin.go
+++ b/core/commands/pin.go
@@ -269,6 +269,11 @@ type RefKeyList struct {
 
 func pinLsKeys(args []string, typeStr string, ctx context.Context, n *core.IpfsNode) (map[string]RefKeyObject, error) {
 
+	mode, ok := pin.StringToPinMode(typeStr)
+	if !ok {
+		return nil, fmt.Errorf("Invalid pin mode '%s'", typeStr)
+	}
+
 	keys := make(map[string]RefKeyObject)
 
 	for _, p := range args {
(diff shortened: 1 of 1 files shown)


core/commands/pin.go
```diff
@@ -269,6 +269,11 @@ type RefKeyList struct {
 
 func pinLsKeys(args []string, typeStr string, ctx context.Context, n *core.IpfsNode) (map[string]RefKeyObject, error) {
 
+	mode, ok := pin.StringToPinMode(typeStr)
+	if !ok {
+		return nil, fmt.Errorf("Invalid pin mode '%s'", typeStr)
```
lgierth: Error message should be lower case

hsanjuan: Thanks for pointing out. Fixed

lgierth:
LGTM

Note this will still hit the network for paths which aren't just plain /ipfs/hash (e.g. /ipns/hash or /ipfs/hash/foo) -- but that's fine and I feel like it's expected.

core/commands/pin.go
```diff
@@ -277,24 +282,18 @@ func pinLsKeys(args []string, typeStr string, ctx context.Context, n *core.IpfsN
 			return nil, err
 		}
 
-		dagNode, err := core.Resolve(ctx, n.Namesys, n.Resolver, pth)
+		c, err := core.ResolveToCid(ctx, n, pth)
 		if err != nil {
 			return nil, err
 		}
 
-		mode, ok := pin.StringToPinMode(typeStr)
```
whyrusleeping: was mode not used?

whyrusleeping: Oh, wait, i see. you just moved this code.

mateon1: This was moved to the beginning of the function, line 272

PR #3421, opened 2016-11-24
Add check if go code is formatted

Kubuxu:


bin/test-go-fmt
```diff
@@ -0,0 +1,13 @@
...
+find . -name '*.go' | xargs gofmt -l > "$T"
+
+if [ -n "$(cat $T)" ]; then
+	echo "Following Go code is not formatted."
+	echo "-----------------------------------"
+	cat "$T"
+	echo "-----------------------------------"
+	echo "Run 'go fmt ./...' in your source directory"
+fi
+rm -f "$T"
```
lgierth: Does it need to `exit 1`?

Kubuxu: It does.

whyrusleeping:
LGTM, been wanting this for a while

PR #3432, opened 2016-11-28
add sharness test for pubsub

keks:
finally!

### DIFF
--- a/test/sharness/lib/iptb-lib.sh
+++ b/test/sharness/lib/iptb-lib.sh
@@ -21,9 +21,15 @@ startup_cluster() {
 	num_nodes="$1"
 	bound=$(expr "$num_nodes" - 1)
 
-	test_expect_success "start up nodes" '
-		iptb start
-	'
+	if [ "$2" = "--enable-pubsub-experiment" ]; then
+		test_expect_success "start up nodes with pubsub enabled" '
+			iptb start --args --enable-pubsub-experiment
+		'
+	else
+		test_expect_success "start up nodes" '
+			iptb start
+		'
+	fi
 
 	test_expect_success "connect nodes to eachother" '
 		iptb connect [1-$bound] 0
--- a/test/sharness/lib/iptb-lib.sh
+++ b/test/sharness/lib/iptb-lib.sh
@@ -19,11 +19,13 @@ check_has_connection() {
 
 startup_cluster() {
 	num_nodes="$1"
+	shift
+	other_args="$@"
 	bound=$(expr "$num_nodes" - 1)
 
-	if [ "$2" = "--enable-pubsub-experiment" ]; then
-		test_expect_success "start up nodes with pubsub enabled" '
-			iptb start --args --enable-pubsub-experiment
+	if test -n "$other_args"; then
+		test_expect_success "start up nodes with additional args" '
+			iptb start --args $other_args
 		'
 	else
 		test_expect_success "start up nodes" '
--- a/test/sharness/t0180-pubsub.sh
+++ b/test/sharness/t0180-pubsub.sh
@@ -24,18 +24,17 @@ test_expect_success 'pubsub' '
 	mkfifo wait ||
 	test_fsh echo init fail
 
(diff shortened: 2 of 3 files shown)


whyrusleeping:
Looks like it fails on sharness (at least on teamcity): http://ci.ipfs.team:8111/viewLog.html?buildId=8663&buildTypeId=GoIpfs_CiTests&tab=buildLog

cc @chriscool for review, this is a bit of strange shell hackery

test/sharness/t0180-pubsub.sh
```diff
@@ -0,0 +1,48 @@
...
+test_expect_success 'pubsub' '
+	echo "testOK" > expected &&
+	mkfifo done &&
+
+	(
+		ipfsi 0 pubsub sub --enc=ndpayload testTopic | 
+			while read line; do
+				echo $line > actual &&
+				echo > done
+				exit
```
chriscool: It's a bit strange to have this "exit" here.
It means that the while loop will in fact never loop, so perhaps it could be an `if read line; then ... fi` instead.

test/sharness/t0180-pubsub.sh
```diff
@@ -0,0 +1,48 @@
...
+			done
+	) &
```
chriscool: It should be explained why we need to fork a process like this, especially as there is a "sleep 1" below.

test/sharness/t0180-pubsub.sh
```diff
@@ -0,0 +1,48 @@
...
+
+	ipfspid=$!
+
+	sleep 1
```
chriscool: Even if the two above commands cannot fail it is a good practice to anyway chain them with the commands below by using "&&" after them.

test/sharness/t0180-pubsub.sh
```diff
@@ -0,0 +1,48 @@
...
+
+	ipfsi 1 pubsub pub testTopic "testOK" &> pubErr1 &&
+
+	cat done &&
+	test_cmp actual expected
```
chriscool: `test_cmp expected actual` is more standard.

test/sharness/t0180-pubsub.sh
keks: Yeah that is a bit hackish I agree. For some reason this loop even needs two runs, I didn't get why and I was glad it worked.

test/sharness/lib/iptb-lib.sh
```diff
@@ -21,9 +21,15 @@ startup_cluster() {
...
-	'
+	if [ "$2" = "--enable-pubsub-experiment" ]; then
+		test_expect_success "start up nodes with pubsub enabled" '
+			iptb start --args --enable-pubsub-experiment
+		'
+	else
+		test_expect_success "start up nodes" '
+			iptb start
+		'
+	fi
```
keks: I tried `iptb start $@` but now I see why that didn't work - its executed in a completely different context. I use your solution now.

chriscool:
Yeah, it LGTM now, thanks!

keks:
@chriscool The tests pass for me locally but fail on Travis and Teamcity. Do you have an idea what the problem could be? Do they pass on your machine?

keks:
I don't think I have a teamcity user account. Can someone click me one? @whyrusleeping @lgierth @jbenet?

whyrusleeping:
@keks the problem is with your previous commit. Under the `ndpayload` encoding, youre outputting an empty line when the 'null' hello message arrives, where instead you should be outputting nothing (since the goal there is to ignore that first message)

whyrusleeping:
This line here: https://github.com/ipfs/go-ipfs/blob/master/core/commands/pubsub.go#L133

Should just return an empty string reader instead of a newline

keks:
Of course! Thank you. "The code is right, it must be the test's fault"...

I will move the checks to `getPsMarshaler` later today because now we have the same code in all three Marshalers.

whyrusleeping:
@keks Thanks a bunch! This is great to have :)

PR #3437, opened 2016-11-29
bitswap: clean up ledgers when disconnecting

whyrusleeping:
License: MIT

### DIFF
--- a/exchange/bitswap/decision/engine.go
+++ b/exchange/bitswap/decision/engine.go
@@ -298,8 +298,32 @@ func (e *Engine) MessageSent(p peer.ID, m bsmsg.BitSwapMessage) error {
 	return nil
 }
 
+func (e *Engine) PeerConnected(p peer.ID) {
+	e.lock.Lock()
+	l, ok := e.ledgerMap[p]
+	if !ok {
+		l = newLedger(p)
+		e.ledgerMap[p] = l
+	}
+	l.lk.Lock()
+	l.ref++
+	l.lk.Unlock()
+	e.lock.Unlock()
+}
+
 func (e *Engine) PeerDisconnected(p peer.ID) {
-	// TODO: release ledger
+	e.lock.Lock()
+	defer e.lock.Unlock()
+	l, ok := e.ledgerMap[p]
+	if !ok {
+		return
+	}
+	l.lk.Lock()
+	l.ref--
+	if l.ref <= 0 {
+		delete(e.ledgerMap, p)
+	}
+	l.lk.Unlock()
 }
 
 func (e *Engine) numBytesSentTo(p peer.ID) uint64 {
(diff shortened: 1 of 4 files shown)


exchange/bitswap/decision/engine.go
```diff
@@ -295,8 +295,24 @@ func (e *Engine) MessageSent(p peer.ID, m bsmsg.BitSwapMessage) error {
...
+	l.lk.Lock()
+	l.ref++
+	l.lk.Unlock()
+}
+
 func (e *Engine) PeerDisconnected(p peer.ID) {
-	// TODO: release ledger
+	e.lock.Lock()
+	defer e.lock.Unlock()
+	l, ok := e.ledgerMap[p]
```
Kubuxu: You are not locking l.lk here, and again we have situation with two locks. It shouts to me `deadlock`.

whyrusleeping: Added the ledger lock. Re locking concerns, these ones are well scoped. the engine lock is either always held first, or not held while taking the ledger lock. And the engine lock is never taken while holding a ledger lock.

Kubuxu:
Other concern: what if PeerConnected gets the instance but can't acquire lock for ledger as it is locked by PeerDisconnected. Then PeerConnected will increase value on ledger that is not longer in ledger map.

whyrusleeping:
@Kubuxu hrm... for that to happen, `PeerConnected` would have to return from `findOrCreate`, and then `PeerDisconnected` would have to take the engine lock, pull the ledger out of the map, and take the ledger lock before the `PeerConnected` process is able to. This *IS* possible. 

One option is to not use `findOrCreate` and instead take the engine lock ourselves throughout the entire call to `PeerConnected`, essentially reimplementing `findOrCreate` in that function.

Kubuxu:
I know that it is something that might happen very rarely or never but those edge cases add up and create hard to track down bugs and instability. If chance of this bug occuring is 0.00001% than chance that it will occur across 100000 runs is more than 60% and if we don't stop possibly introducing bugs like that go-ipfs will be always unstable and unreliable.

whyrusleeping:
@Kubuxu Right, So i think the solution is to make `PeerConnected` hold the engine lock through the entire method.

Kubuxu:
So now it is thread safe, but does function `findOrCreate` makes sense if we have introduced ref counting?

Kubuxu:
Also I am still not a fan of those two locks as some not really connected change can introduce deadlock (locking for engine while holding ledger some ledger) and we might not catch it when we introduce it. We should really look into Actor oriented communication and how bad/good it will be.

Kubuxu:
I rebased it to run coverage on it.

Kubuxu:
It isn't tested anywhere, it might be worth to do that.

whyrusleeping:
@kubuxu this part of the code used to be actor oriented,and was vastly more
complicated, and much more difficult to get working properly. In addition
to requiring a very large amount of coroutines to get running.

Kubuxu:
I am sure that we can make it clean and not so complicated with enough layers of sugarcoating.

I am just almost sure that we will introduce deadlock around this place sooner or later and it won't be diagnosed for a long time as reproduction of this will be almost impossible.

This change LGTM if I get some tests. In case of not directly sharness tested features I would like the `codecov/patch` build check to be green.

Kubuxu:
Ok, it shows as if there was no coverage due to lack of cross package cover testing.

lgierth:
Can I add the RFM label here? Let's continue the locking discussion in #3506.

Issue #3427, opened 2016-11-26
CIDv1 link specifications do not work ( hardcoded mdag expectation somewhere )


mib-kd743naq:
#### Version information:
```
go-ipfs version: 0.4.5-dev-bf9927f
Repo version: 4
System version: amd64/linux
Golang version: go1.7.1
```

#### Type:
`Bug`

#### Priority:
`P4`

#### Description:
Full description at https://ipfs.io/ipfs/QmXUmGtjEKnceju74LDStC24DDmJWxnxdH7aQPLYk9zJe5
```
=== Setup
String "abcdef\ngh" broken into 3 chunks on the (and including) newline

=== Problem
~$ ipfs cat zdj7WmC19DKCa9i2viQtsyqymhUFUgJzU8mZN4ytku3L75XAD | hexdump -C
Error: merkledag: not found
...
```

whyrusleeping:
Do you mind pasting the commands you used to generate the objects in question?

mib-kd743naq:
@whyrusleeping There is no command I can currently share.

[As I said previously](https://github.com/ipfs/notes/issues/189) these are all experiments of mine trying to figure out how I can use ipfs for my own needs, entirely outside of go-ipfs, but remaining compatible with it (i.e. having `http://ipfs.io/ipfs/<my hash>` work).

I am essentially assembling all the bytes by hand, and then feeding them to `http://127.0.0.1:5001/api/v0/block/put?format=raw` or similar. The current "codebase" is a string of piped invocations of small converter utilities, with hand-added sizes etc, and most definitely not suitable for public consumption, sorry :/

whyrusleeping:
okay, can you provide a set of commands like:

```
printf "foobar" | ipfs block put --format=raw
```

That recreate each object?

mib-kd743naq:
Sure thing:

```
~$ for d in \
  Qmb6pVMgw4hMmsLU6Mz28ZV5uHzceGxR6pj7k4wKS2WLNZ \
  QmRnnQXyW5wNz6rvjw4fQQ8XWaLvyfkvJBztC3WGx1gmyg \
  QmaxVysa8dDZtk5g9AfdQouDVZdNZ3D2NFePZrME61X8GX
do \
  ipfs block get $d > $d.data
done
```
```
~$ for l in \
  zdj7WgpD1zodrFvjBSWMXL2y9dR7z7H1TJdBhrTLZcS4LGPH9 \
[...]

Issue #3431, opened 2016-11-28
CIDv1 pin specification doesn't work


mib-kd743naq:
#### Version information:
```
go-ipfs version: 0.4.5-dev-bf9927f
Repo version: 4
System version: amd64/linux
Golang version: go1.7.1
```

#### Type:
Bug

#### Priority:
P2

#### Description:
```
$ ipfs pin add zdj7WgpD1zodrFvjBSWMXL2y9dR7z7H1TJdBhrTLZcS4LGPH9
pinned QmZixeZjnNnKrU8acnNR6mnrVmuqHutVKPSvfmDmEthV7D recursively
```
Did not expect to see a "CIDv0" at the output, likely indicates incorrect index store or somesuch...

whyrusleeping:
Ah, yeah. I think i know whats happening here. The protobuf nodes always default to generating a cidV0. We might need some way to store cid prefix in the nodes to make sure we generate the correct output cid...

whyrusleeping:
@mib-kd743naq can you verify that this was fixed in latest master?

mib-kd743naq:
Works as of 0c413de

PR #3443, opened 2016-11-29
merkledag: retain cid types when roundtripping through a ProtoNode

whyrusleeping:
Should address #3431

License: MIT

### DIFF
--- a/merkledag/node.go
+++ b/merkledag/node.go
@@ -22,6 +22,9 @@ type ProtoNode struct {
 	encoded []byte
 
 	cached *cid.Cid
+
+	// prefix specifies cid version and hashing function
+	prefix cid.Prefix
 }
 
 type LinkSlice []*node.Link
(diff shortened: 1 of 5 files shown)


whyrusleeping:
@mib-kd743naq wanna verify that this solves the problems youre seeing in #3427 #3431 ?

lgierth:
Given that this fixes #3427 and #3431, LGTM :+1:

merkledag/node.go
```diff
@@ -22,6 +22,16 @@ type ProtoNode struct {
...
 
 	cached *cid.Cid
+
+	// Prefix specifies cid version and hashing function
+	Prefix cid.Prefix
+}
+
+var defaultCidPrefix = cid.Prefix{
+	Codec:    cid.DagProtobuf,
+	MhLength: -1,
```
kevina: Should this really be -1?  Does it get set somewhere else.  Just double checking, so if you know this is correct than don't worry about it.

whyrusleeping: -1 implies "use default length".

PR #3445, opened 2016-11-29
bitswap: add a deadline to sendmsg calls

whyrusleeping:
Bitswap occasionally gets stuck because writes hang (yamux....). this should at least keep bitswap from getting stuck here due to yamux:

ref: https://ipfs.io/ipfs/QmQn8YDTHWJzF7GVKUKanBzoxbzd5jQpwsgM2PfFVTKuGR

License: MIT

### DIFF
--- a/exchange/bitswap/network/interface.go
+++ b/exchange/bitswap/network/interface.go
@@ -38,7 +38,7 @@ type BitSwapNetwork interface {
 }
 
 type MessageSender interface {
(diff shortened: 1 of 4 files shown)


whyrusleeping:
logging these as warnings because if we consider them errors, our tests fail since the mocknet conns don't support setdeadlines

lgierth:
LGTM :+1:

PR #3447, opened 2016-11-30
add option to enable go-multiplex experiment

whyrusleeping:
License: MIT

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -14,7 +14,10 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net"
+	"os"
+	"strings"
 	"time"
 
 	bstore "github.com/ipfs/go-ipfs/blocks/blockstore"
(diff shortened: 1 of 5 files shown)


core/core.go
```diff
@@ -207,6 +217,34 @@ func (n *IpfsNode) startOnlineServices(ctx context.Context, routingOption Routin
...
+		MaxStreamWindowSize:    uint32(1024 * 512),
+		LogOutput:              ioutil.Discard,
+	}
+
+	mstpt.AddTransport("/yamux/1.0.0", ymxtpt)
+
+	mstpt.AddTransport("/spdy/3.1.0", spdy.Transport)
+
+	if mplexExp {
+		mstpt.AddTransport("/mplex/1.0.0", mplex.DefaultTransport)
```
diasdavid: Since multiplex doesn't have a formal spec, the best we can do is follow its semver from its main implementation. See:

https://github.com/libp2p/specs/blob/master/7-properties.md#757-protocol-multicodecs

`multiplex	/multiplex/6.7.0`

PR #3448, opened 2016-11-30
bitswap: increase wantlist resend delay to one minute

whyrusleeping:
I honestly thought this was removed a while ago, but here it is (see line 299). I'm going to bump this up to once per minute to make sure nothing weird starts happening. Then later we can completely remove it.

License: MIT

### DIFF
--- a/exchange/bitswap/bitswap.go
+++ b/exchange/bitswap/bitswap.go
@@ -57,7 +57,7 @@ func init() {
 	}
 }
 
-var rebroadcastDelay = delay.Fixed(time.Second * 10)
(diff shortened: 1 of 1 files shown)


Kubuxu:
If it was in all the time, and every 10s, why we are having problems with wantlist consistency?

whyrusleeping:
yamux

whyrusleeping:
For example, check out goroutine 14008165 here: https://ipfs.io/ipfs/QmQn8YDTHWJzF7GVKUKanBzoxbzd5jQpwsgM2PfFVTKuGR

Kubuxu:
😢 

LGTM

PR #3458, opened 2016-12-02
Switch unixfs.Metadata.MimeType to optional

mib-kd743naq:
*** THIS IS A BREAKING CHANGE *** as per the google documentation: ["Required is forever"](https://developers.google.com/protocol-buffers/docs/proto#specifying-field-rules)

Nevertheless this seems like a good idea at this time: there are no known
producers ( nor consumers ) of MetaData nodes, and the current requirement
of MimeType has an extremely narrow application scope.

This change could very well be rejected in lieu of implementing a new type
of node ( e.g. TheRealMetadata ) in the DataType enum.


jbenet:
I'm in favor of this. `required` was a mistake in protobuf. using it was our mistake.

whyrusleeping:
LGTM, thanks @mib-kd743naq

Kubuxu:
Are you sure that this won't break protobuf reading in older builds?

PR #2657, opened 2016-05-10
Add Defaults to `ipfs add`

RichardLitt:
I didn't bother with Chunker, because I think that is a much wider PR. These should all be solid, though. Redid some of the logic to make it smoother.

Part of #2484.

PR #3464, opened 2016-12-04
Revert Add Defaults to `ipfs add` / Fix default logic of `--progress` option

kevina:
This 
* reverts #2657 
* adds (back) the .Default(true) for the `--pin` option
* cleans up the default logic of the `--progress option`

Note: #2657 broke `--progress=false`.  After #2657 when `--progress=false` a useless progress bar is displayed:
```
$ ipfs add --progress=false afile 
added QmbWib2kLXUBpG8GZjRfmB3rEVE75SYCnN44AeLNz5Qe5V afile
 0 B / 96.00 MB [----------------------------------------------------------------------------------------------]   0.00%
```
This pull request fixes that.

The last commit "add cmd: clean up default logic of --progress option" correctly does what I think @RichardLitt was originally trying to do in 03dd669430474ae8396e6fe9eba9e47f01c0ed75 (part of  #2657).  If necessary it can be separated out into its own pull request.

### DIFF
--- a/core/commands/add.go
+++ b/core/commands/add.go
@@ -37,11 +37,9 @@ const (
 
 var AddCmd = &cmds.Command{
 	Helptext: cmds.HelpText{
-		Tagline: "Add a file to ipfs.",
+		Tagline: "Add a file or directory to ipfs.",
 		ShortDescription: `
-Adds contents of <path> to ipfs. Use -r to add directories.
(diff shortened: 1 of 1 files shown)


whyrusleeping:
LGTM, thanks @kevina

RichardLitt:
Thanks, @kevina!

PR #3472, opened 2016-12-06
basic keystore implementation

whyrusleeping:
Thank you to @kyledrake for prompting me to just write this already

License: MIT

### DIFF
--- a/keystore/keystore.go
+++ b/keystore/keystore.go
@@ -69,6 +69,8 @@ func (ks *FSKeystore) Put(name string, k ci.PrivKey) error {
 	_, err = os.Stat(kp)
 	if err == nil {
 		return ErrKeyExists
+	} else if !os.IsNotExist(err) {
+		return err
 	}
 
 	fi, err := os.Create(kp)
--- a/keystore/keystore.go
+++ b/keystore/keystore.go
@@ -1,0 +1,123 @@
+package keystore
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+
+	ci "gx/ipfs/QmfWDLQjGjVe4fr5CoztYW2DYYjRysMJrFe1RCsXLPTf46/go-libp2p-crypto"
+)
+
+type Keystore interface {
+	Put(string, ci.PrivKey) error
+	Get(string) (ci.PrivKey, error)
+	Delete(string) error
+	List() ([]string, error)
+}
+
+var ErrNoSuchKey = fmt.Errorf("no key by the given name was found")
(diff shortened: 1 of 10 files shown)


haadcode:
Oooh! Excellent stuff! Big fan of your work J :)

Thank you @kyledrake for prompting @whyrusleeping to just write this already.

keystore/keystore.go
```diff
@@ -0,0 +1,123 @@
...
+
+	b, err := k.Bytes()
+	if err != nil {
+		return err
+	}
+
+	kp := filepath.Join(ks.dir, name)
+
+	_, err = os.Stat(kp)
+	if err == nil {
```
Kubuxu: To be safe I would use `os.IsExist` and change it to `ErrKeyExists` then, else passthrough the error.

keystore/keystore.go
```diff
@@ -0,0 +1,123 @@
...
+		return ErrKeyExists
+	}
+
+	fi, err := os.Create(kp)
+	if err != nil {
+		return err
+	}
+	defer fi.Close()
+
+	_, err = fi.Write(b)
```
Kubuxu: I thought you wanted to use multicodec so we can migrate much easier?

Kubuxu:
Also there are absolutely no tests for it.

whyrusleeping:
> Also there are absolutely no tests for it.

Yeah yeah, i'm getting there.

keystore/keystore.go
whyrusleeping: How would you use `os.IsExist` here?

Kubuxu: sorry, IsNotExist

whyrusleeping: ah! okay, i see what you want

wigy-opensource-developer:
@whyrusleeping I love the new feature, thanks for implementing it.

What is missing is the renewal of the IPNS records every 4 hours. The namesys/republisher uses a so-called keybook in the peerstore that only contains the `self` key. Probably the republisher needs to use the keystore, too, to enumerate all IPNS records that need a renewal.

As I currently understand, the republisher/reprovider is the only part of the network that fixes data loss caused by peers going offline or leaving the network. If there is no automatic recovery from it in the DHT layer, in the current state of the code, the `ipns name publish --key notself /ipfs/QmBlah` needs to be executed regularly as a workaround to make sure the name can be resolved.

whyrusleeping:
Yeah, that's the next part that needs to be done. Can you open a new issue
for discussing how we should go about doing that? Configuration UX will be
important

Kubuxu:
I have opened it already: https://github.com/ipfs/go-ipfs/issues/3537

PR #3473, opened 2016-12-06
Switch coverage to codecov from coveralls

Kubuxu:
It works, coveralls stopped suddenly. It has also much nicer UI and more config options.

Still WIP, I will comment when it is ready.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,8 @@
 *.orig
 *~
 
+coverage.txt
+
 .ipfs
 bin/gx
 bin/gx*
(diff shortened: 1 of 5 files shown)


Kubuxu:
After this build passes on CircleCI it should be ready.

Comments will be disabled once we merge it to master (config file).

lgierth:
> Comments will be disabled once we merge it to master (config file).

Will it use the github PR status then? I think it's important to show a big red flag if coverage decreases with a PR.

Kubuxu:
Yes, once CircleCI goes through the build status for coverage will be shown.

Disabling comments was required by @whyrusleeping

Kubuxu:
The last few patches also might have to be in master. If not, I will work a bit more to get it to work.

@whyrusleeping should be good to merge.

PR #3474, opened 2016-12-06
Do not require CI to pass for coverage

Kubuxu:
If travis fail (or is cancelled) or random build failure happens, or
js-ipfs-api test is down for a long time we still want the coverage.

WIP

License: MIT


Kubuxu:
I want to check if it works, if not I have other idea still.

Kubuxu:
@whyrusleeping it is done.

PR #3476, opened 2016-12-07
update message.proto

diasdavid:
Add some comments so that I don't forget about these


whyrusleeping:
@diasdavid fix your commit msg pls

diasdavid:
@whyrusleeping there:
image

Kubuxu:
gitcop doesn't like you as it want's email in triangular brackets.

diasdavid:
thank you @Kubuxu, I was missing that

PR #3477, opened 2016-12-07
Add multiple small tests

Kubuxu:


Kubuxu:
It was more for testing codecov than real test pass but there are some tests in there.

PR #3484, opened 2016-12-08
cmd/pubsub: fix peers command topic filtering

### LINKED ISSUE #3467: Pubsub returns wrong peers for a topic
#### Version information:

0.4.5-pre1

#### Type:

Bug

#### Description:

When I call pubsub/peers(topic), go-ipfs returns all peers that are connected to it. Expected behaviour is to return only the ones that are subscribed to the specified topic.

whyrusleeping:
closes #3467 

License: MIT

### DIFF
--- a/core/commands/pubsub.go
+++ b/core/commands/pubsub.go
@@ -274,15 +274,20 @@ To use, the daemon must be run with '--enable-pubsub-experiment'.
 
 var PubsubPeersCmd = &cmds.Command{
 	Helptext: cmds.HelpText{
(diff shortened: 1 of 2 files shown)


core/commands/pubsub.go
```diff
@@ -302,8 +305,13 @@ To use, the daemon must be run with '--enable-pubsub-experiment'.
 			return
 		}
 
+		var topic string
+		if len(req.Arguments()) == 1 {
```
Kubuxu: This will cause confusion if more topics are passed. IMO either error out or use all of them.

PR #3487, opened 2016-12-08
makefile: use functions properly

Kubuxu:
Should solve the logs we see on OSX

But now I have no idea how it worked before

License: MIT


Kubuxu:
No it didn't help.

PR #3493, opened 2016-12-09
Core API refinements and efficiency improvements

### LINKED ISSUE #3490: Core API needs to take flexible inputs
We've layed the groundwork for the Core API in go-ipfs by implementing parts of the Unixfs API. Its functions properly return CIDs, but they only accept path strings as arguments, which requires unneccessary conversions in most situations.

Core API functions should accept:

- CIDs
- Strings
- Multihashes
- Paths
- ipfs.io URIs

The go-cid package in use provides `Parse()` which takes care of it.

lgierth:
cc @hsanjuan @keks @kevina @Kubuxu 

Closes #3490  

commit e1c190a992ea2047a83ad3c9cd8dc303b5564ef7
Author: Lars Gierth <larsg@systemli.org>
Date:   Wed Nov 16 06:21:15 2016 +0100

    coreapi: smarter way of dealing with the different APIs
    
    License: MIT

commit 2e15ba92b4aa4945e122ee94c6ea30ce81379d74
Author: Lars Gierth <larsg@systemli.org>
Date:   Fri Dec 9 16:35:28 2016 +0100

    coreapi: be more flexible about inputs
    
    While coreapi functions previously only accepted path strings,
    they now accept anything from Path structs and strings,
    Multihash structs and strings, and most notably: Cid structs,
    strings and byte arrays. See go-cid.Parse() for more information.
    
    If the argument is a string containing at least one forward-slash,
    it's treated as a path and resolved accordingly, i.e. including
    IPNS resolution and link traversal.
    
    Anything else will be parsed as a Cid and fetched directly using DAG.Get().
    
    This change makes various use cases of the Core API more efficient,
    since there's no more need for expensive string/Path/Cid conversions.
    
    License: MIT

commit 7241c02ef4362591021739ae2c5974211ebf8bec
Author: Lars Gierth <larsg@systemli.org>
Date:   Fri Dec 9 16:49:24 2016 +0100

    gateway: make serving index.html a little more efficient
    
    Prevents unneccessary conversions and lookups :):)
    
    License: MIT

### DIFF
--- a/core/coreapi/interface/interface.go
+++ b/core/coreapi/interface/interface.go
@@ -9,16 +9,15 @@ import (
 	ipld "gx/ipfs/QmYDscK7dmdo2GZ9aumS8s5auUUAH5mR1jvj5pYhWusfK7/go-ipld-node"
 )
 
-// type CoreAPI interface {
-// 	ID() CoreID
-// 	Version() CoreVersion
-// }
-
 type Link ipld.Link
 
 type Reader interface {
 	io.ReadSeeker
 	io.Closer
+}
+
+type CoreAPI interface {
+	Unixfs() UnixfsAPI
 }
 
 type UnixfsAPI interface {
--- a/core/coreapi/interface/interface.go
+++ b/core/coreapi/interface/interface.go
@@ -9,6 +9,16 @@ import (
 	ipld "gx/ipfs/QmYDscK7dmdo2GZ9aumS8s5auUUAH5mR1jvj5pYhWusfK7/go-ipld-node"
 )
 
+type Path interface {
+	String() string
+	Cid() *cid.Cid
+	Root() *cid.Cid
+	Resolved() bool
+}
+
+// TODO: should we really copy these?
+//       if we didn't, godoc would generate nice links straight to go-ipld-node
+type Node ipld.Node
 type Link ipld.Link
 
 type Reader interface {
@@ -18,12 +28,14 @@ type Reader interface {
 
 type CoreAPI interface {
 	Unixfs() UnixfsAPI
+	ResolvePath(context.Context, Path) (Path, error)
+	ResolveNode(context.Context, Path) (Node, error)
 }
 
 type UnixfsAPI interface {
-	Add(context.Context, io.Reader) (*cid.Cid, error)
-	Cat(context.Context, string) (Reader, error)
-	Ls(context.Context, string) ([]*Link, error)
+	Add(context.Context, io.Reader) (Path, error)
+	Cat(context.Context, Path) (Reader, error)
+	Ls(context.Context, Path) ([]*Link, error)
 }
 
 // type ObjectAPI interface {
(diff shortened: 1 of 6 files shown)


whyrusleeping:
blocked on #3490

whyrusleeping:
@lgierth progress here? I think its probably safe to move this to the 0.4.7 milestone for now?

lgierth:
>  progress here? I think its probably safe to move this to the 0.4.7 milestone for now?

Yes good to move, it needs the two changes we discussed but which I haven't posted here yet (will dig them up).

whyrusleeping:
@lgierth any progress here since last update?

lgierth:
will take care of this first thing in the docs sprint, so we can get it into 0.4.8

lgierth:
coreapi: make the interfaces path centric

The new coreiface.Path maps a path to the cid.Cid
resulting from a full path resolution.

The path is internally represented as a go-ipfs/path.Path,
but that doesn't matter to the outside.

Apart from the path-to-CID mapping, it also aims to hold all
resolved segment CIDs of the path. Right now it only exposes
Root(), and only for flat paths a la /ipfs/Qmfoo. In other cases,
the root is nil.

In the future, resolution will internally use
go-ipfs/path.Resolver.ResolvePathComponents and thus always return
the proper resolved segments, via Root(), or a future Segments() func.

- Add coreiface.Path with Cid() and Root().
- Add CoreAPI.ResolvePath() for getting a coreiface.Path.
- All functions now expect and return coreiface.Path.
- Add ParsePath() and ParseCid() for constructing a coreiface.Path.
- Add coreiface.Node and Link which are simply go-ipld-node.Node and Link.
- Add CoreAPI.ResolveNode() for getting a Node from a Path.

lgierth:
Mh, a nicer way to work with ipfspath.Paths here would be good.

lgierth:
And Path should probably be moved into coreiface.

lgierth:
> And Path should probably be moved into coreiface.

Nevermind. I meant the ParsePath/ParseCid/ResolvedPath funcs, but these are just fine.

core/coreapi/interface/interface.go
```diff
@@ -9,22 +9,33 @@ import (
...
+	Unixfs() UnixfsAPI
+	ResolvePath(context.Context, Path) (Path, error)
+	ResolveNode(context.Context, Path) (Node, error)
+}
+
 type UnixfsAPI interface {
-	Add(context.Context, io.Reader) (*cid.Cid, error)
-	Cat(context.Context, string) (Reader, error)
-	Ls(context.Context, string) ([]*Link, error)
+	Add(context.Context, io.Reader) (Path, error)
```
whyrusleeping: Hrm... this one feels weird. I like that `Cat` takes a path and returns a reader, and that `Add` is the inverse operation, but it just feels weird that `Add` returns a path...

I think i'll get over it. The `Path` in question can be thought of as `/ipfs/Cid`

Issue #2129, opened 2015-12-28
Fix auto repo GC tests


whyrusleeping:
t0082 fails very frequently, i'm disabling it until we can figure out how to make the test pass reliably (timing things is difficult).

ref #2128

lgierth:
I have code towards bringing these tests back, but it involves a rewrite of corerepo/gc.go and adding GC status to the repo/stat command: [feat/corerepo-gc](https://github.com/ipfs/go-ipfs/tree/feat/corerepo-gc)

On the upside, the tests won't rely on time anymore then.

PR #3494, opened 2016-12-09
Remove GC timeout, fix GC tests

### LINKED ISSUE #3436: Automatic GC times out
#### Description:

The automatic GC times out. StorageMax is 54G, the repo is 60G big, so the timeout for GC should be something like 6 minutes.

From jupiter:
```
23:41:29.807 WARNI   corerepo: pre-GC: Maximum storage limit exceeded. Maybe unpin some files? gc.go:187
23:41:29.807  INFO   corerepo: Watermark exceeded. Starting repo GC... gc.go:191
23:42:58.626 ERROR   corerepo: context deadline exceeded gc.go:165
```

Note how the context times out after after less than 1.5m already.

lgierth:
Closes #3436 #2129

commit a9b8c8bbd5e243f12d44f67fdb46b80f7f096c5e
Author: Lars Gierth <larsg@systemli.org>
Date:   Fri Dec 9 17:26:53 2016 +0100

    gc: remove wonky timeout
    
    This timeout was:
    - unneccessary
    - based on a metric that doesn't make sense
    
    License: MIT

commit 8f0c7ee3b02ad0a56358d46f32abd7729e05a791
Author: Lars Gierth <larsg@systemli.org>
Date:   Fri Dec 9 17:28:31 2016 +0100

    gc: remove unneccessary full repo scan
    
    GetStorageUsage() is super expensive as it involves a full repo scan,
    and it's already bad enough that we have do it once.
    
    The call that's being removed here is purely for cosmetical purposes:
    printing the number of bytes freed by the GC run. Let's drop it.
    
    License: MIT

commit 49802ebac9526740fb0d3dad2c39fe0b132589ba
Author: Lars Gierth <larsg@systemli.org>
Date:   Fri Dec 9 18:04:55 2016 +0100

    gc: fix and re-enable sharness tests
    
    I'm bumping up the waiting time significantly.
    I prefer a slightly slower test to no test at all.
    
    License: MIT

### DIFF
--- a/core/corerepo/gc.go
+++ b/core/corerepo/gc.go
@@ -194,19 +194,7 @@ func (gc *GC) maybeGC(ctx context.Context, offset uint64) error {
 		if err := GarbageCollect(gc.Node, ctx); err != nil {
 			return err
 		}
-		newStorage, err := gc.Repo.GetStorageUsage()
-		if err != nil {
-			return err
-		}
-		log.Infof("Repo GC done. Released %s\n", humanize.Bytes(uint64(storage-newStorage)))
-		if newStorage > gc.StorageGC {
-			log.Warningf("post-GC: Watermark still exceeded")
-			if newStorage > gc.StorageMax {
-				err := ErrMaxStorageExceeded
-				log.Error(err)
-				return err
-			}
-		}
+		log.Infof("Repo GC done. See `ipfs repo stat` to see how much space got freed.\n")
 	}
 	return nil
 }
--- a/core/corerepo/gc.go
+++ b/core/corerepo/gc.go
@@ -190,11 +190,8 @@ func (gc *GC) maybeGC(ctx context.Context, offset uint64) error {
 		// Do GC here
 		log.Info("Watermark exceeded. Starting repo GC...")
 		defer log.EventBegin(ctx, "repoGC").Done()
-		// 1 minute is sufficient for ~1GB unlink() blocks each of 100kb in SSD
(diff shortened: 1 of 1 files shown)


lgierth:
Oh my, it timed out on me right away again. Is there any way to avoid the sleep? sync(1)?

Kubuxu:
Sync won't matter as it does kernel cache to disk and problem here is go-ipfs process to kernel. I would just suggest bigger files and error margins.

whyrusleeping:
Yeah... i remember this test being really troublesome. Let me know if you need any help @lgierth

lgierth:
If we're trying to wait for IPFS flusing its writes, why not expose the state of that in the `diag` command?

Kubuxu:
I don't think it will be easy.

whyrusleeping:
@lgierth ipfs flushes its writes... what is the problem?

Kubuxu:
The delay it will take to flush the write is unspecified AFAIK.

whyrusleeping:
The commands shouldnt return before things get flushed...  (unless you have no-flush set)

whyrusleeping:
@lgierth progress here?

lgierth:
This is good to go, I've split the stuff about tests out into the feat/corerepo-gc branch, since it turned into a full-blown rewrite of this GC stuff. :)

This one here now only removes the timeout context around gc.GC(), and the second repo scan after the GC run.

lgierth:
OSX failure is #3496

PR #3495, opened 2016-12-09
merkledag: add a concurrency limit to merkledag fetch graph

whyrusleeping:
This should help solve a few 'cant pin giant things' issues.

License: MIT

### DIFF
--- a/merkledag/merkledag.go
+++ b/merkledag/merkledag.go
@@ -449,6 +449,10 @@ func EnumerateChildrenAsync(ctx context.Context, ds DAGService, c *cid.Cid, visi
 	}
 }
 
+// FetchGraphConcurrency is total number of concurrenct fetches that
+// 'fetchNodes' will start at a time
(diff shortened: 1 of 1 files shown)


merkledag/merkledag.go
```diff
@@ -458,8 +458,13 @@ func fetchNodes(ctx context.Context, ds DAGService, in <-chan []*cid.Cid, out ch
 		close(out)
 	}()
 
+	rateLimit := make(chan struct{}, 8)
```
Kubuxu: It would be better to have this magic number extracted and documented.

whyrusleeping: FIIINE.

```
// Eight is 2 to the third power, alternatively it can be represented as 7+1, or the cube root of 512
var Eight = 8
```

lgierth: It's also the what you get when you multiply a random number with two!

PR #3499, opened 2016-12-13
dag get should be a read only command

jbenet:
added CmdDagGet to the read only command set.

tried to add a test, but dag get needs tests in general


whyrusleeping:
commit message is valid, duplicate import fixed, this LGTM

PR #3501, opened 2016-12-13
namesys: add entry to DHT cache after publish

### LINKED ISSUE #2078: name publish doesn't instantly update the cache entry in the local node
When `ipfs name publish` finishes, it doesn't instantly update the cache of the local node.

This is a problem in [my app](https://github.com/fazo96/ipfs-boards): the user updates his profile, but it takes a while (around 20 seconds) until he sees his changes.

Kubuxu:
Resolves #2078

License: MIT

### DIFF
--- a/namesys/namesys.go
+++ b/namesys/namesys.go
@@ -1,13 +1,15 @@
 package namesys
 
 import (
+	"context"
 	"strings"
 	"time"
 
-	context "context"
 	path "github.com/ipfs/go-ipfs/path"
+
 	ds "gx/ipfs/QmRWDav6mzWseLWeYfVd5fvUKiVe9xNH29YfMF438fG364/go-datastore"
 	routing "gx/ipfs/QmbkGVaN9W6RYJK4Ws5FvMKXKDqdRQ5snhtaa92qP6L8eU/go-libp2p-routing"
+	peer "gx/ipfs/QmfMmLGoKzCHDN7cGgk64PJr4iipzidDRME8HABSJqvmhC/go-libp2p-peer"
 	ci "gx/ipfs/QmfWDLQjGjVe4fr5CoztYW2DYYjRysMJrFe1RCsXLPTf46/go-libp2p-crypto"
 )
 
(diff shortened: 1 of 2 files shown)


Kubuxu:
@whyrusleeping it isn't supper nice solution but without some refactor there isn't anything much better might be done.

namesys/namesys.go
```diff
@@ -87,9 +89,44 @@ func (ns *mpns) resolveOnce(ctx context.Context, name string) (path.Path, error)
 
 // Publish implements Publisher
 func (ns *mpns) Publish(ctx context.Context, name ci.PrivKey, value path.Path) error {
-	return ns.publishers["/ipns/"].Publish(ctx, name, value)
+	err := ns.publishers["/ipns/"].Publish(ctx, name, value)
+	if err != nil {
+		return err
+	}
+	ns.addToDHTCache(name, value, time.Now().Add(time.Hour*24))
```
whyrusleeping: lets pull this magic number out and justify it

namesys/namesys.go
```diff
@@ -87,9 +89,44 @@ func (ns *mpns) resolveOnce(ctx context.Context, name string) (path.Path, error)
...
+	return nil
 }
 
-func (ns *mpns) PublishWithEOL(ctx context.Context, name ci.PrivKey, val path.Path, eol time.Time) error {
-	return ns.publishers["/ipns/"].PublishWithEOL(ctx, name, val, eol)
+func (ns *mpns) addToDHTCache(key ci.PrivKey, value path.Path, eol time.Time) {
+	var err error
+	value, err = path.ParsePath(value.String())
+	if err != nil {
+		log.Error("could not parse path")
```
whyrusleeping: maybe add the actual error to this? might aid in debugging random error messages

namesys/namesys.go
```diff
@@ -87,9 +89,44 @@ func (ns *mpns) resolveOnce(ctx context.Context, name string) (path.Path, error)
...
+	}
+	ns.addToDHTCache(name, value, eol)
+	return nil
 }
 
-func (ns *mpns) PublishWithEOL(ctx context.Context, name ci.PrivKey, val path.Path, eol time.Time) error {
-	return ns.publishers["/ipns/"].PublishWithEOL(ctx, name, val, eol)
+func (ns *mpns) addToDHTCache(key ci.PrivKey, value path.Path, eol time.Time) {
+	var err error
+	value, err = path.ParsePath(value.String())
```
whyrusleeping: wait, why are we converting the thing to a string, then back to a path?

namesys/namesys.go
Kubuxu: It is the same as in publisher.go, should I add comment for it?

namesys/namesys.go
Kubuxu: Yes as the publisher can publish hash only but resolver will sanitize this value to be `/ipfs/CID`.

namesys/namesys.go
Kubuxu: This is only cache, I didn't want it to fail the publish (as it was already published and we can't revert it).

namesys/namesys.go
Kubuxu: Or extract it from publisher?

Kubuxu:
Yes, they are separate modules that don't know about echo other's existence.

wigy-opensource-developer:
Is this problem IPNS specific? I love how this PR solves the problem in hand without changing much code, but I see it as a temporary fix that will need to be removed after the real issue has been addressed.

The caching could be moved into the DHT-based IPRS implementation, so the /pk/ namespace could benefit from it as well. Namesys uses both. I am not aware of other IPRS namespaces, but probably it would apply to any record types.

Kubuxu:
IPRS doesn't exist yet, I like to call IPNS a huge PoC that mutable data is possible even though the content is immutable.

Yes, after IPRS is designed, speced out, and implemented it can be moved into one place with infrastructure that will include caching of self published record in its design. When code of namesys was written it wasn't a feature that was taken into account.

namesys/namesys.go
whyrusleeping: I'd extract it from both, and give it a small comment

whyrusleeping: Lets just call it `DefaultRecordTTL` (i had to go look at it again to remember what exactly this was)

Kubuxu: Hah, gave it the same name without reading this. <3

namesys/namesys.go
```diff
@@ -87,9 +89,47 @@ func (ns *mpns) resolveOnce(ctx context.Context, name string) (path.Path, error)
 
 // Publish implements Publisher
 func (ns *mpns) Publish(ctx context.Context, name ci.PrivKey, value path.Path) error {
-	return ns.publishers["/ipns/"].Publish(ctx, name, value)
+	err := ns.publishers["/ipns/"].Publish(ctx, name, value)
+	if err != nil {
+		return err
+	}
+	ns.addToDHTCache(name, value, time.Now().Add(DefaultRecortTTL))
```
whyrusleeping: typo: DefaultRecor<b>d</b>TTL

Cole128:
I am still experiencing slow ipns resolving. For a minute after I publish it will be fast, but then it takes a very long time.

PR #3502, opened 2016-12-13
Update go-ds-measure to 1.1.0

Kubuxu:
This means that now datastore metrics are collected in prometheus.

License: MIT

### DIFF
--- a/package.json
+++ b/package.json
@@ -219,9 +219,9 @@
     },
     {
       "author": "whyrusleeping",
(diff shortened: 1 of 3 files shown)


whyrusleeping:
@Kubuxu did you perhaps forget to pin the hash?

Kubuxu:
@whyrusleeping now it should be ok

whyrusleeping:
LGTM, thanks @Kubuxu !

Issue #3488, opened 2016-12-08
make: rework test/ Makefile


Kubuxu:
It is overly complicated and some of its parts might not work at all. Serious cleanup is required there.

Kubuxu:
It looks like I have started overall rework of our makefiles.

Kubuxu:
See branch https://github.com/ipfs/go-ipfs/tree/feat/makefile/refactor

PR #3504, opened 2016-12-14
Makefile rework and sharness test coverage

Kubuxu:
This PR introduces non-recursive Makefile infrastructure that replaces current Makefile infrastructure.
It also generally cleanups the Makefiles, separates them into nicer sub-modules and centralizes common operations into single definitions.

#### Why non-recursive Makefile?
It allows to depend on any target that is defined in the makefile, this means that for example `gx install` is called once when `make build test_expensive_sharness` is called instead of 4 or 5 times.

It also makes the dependencies much cleaner and allows for reuse of modules. For example sharness coverage collection uses sharness target with amended PATH, previously it might have been possible but not without wiring the coverage collection into sharness Makefile runner code.

#### Isn't non-recursive Makefile much more complex?
Yes, it is a bit more complex but not much more. There are few rules that have to be followed and few complexities added but IMHO it is worth it.

How to NR-make:
1. If make is to generate some file via a target, it MUST be defined in Rules.mk file in the directory of the target.
2. `Rules.mk` file MUST have `include mk/header.mk` statement as the first line and `include mk/footer.mk` statement as the last line (apart from project root `Rules.mk`).
3. It then MUST be included by the closest `Rules.mk` file up the directory tree.
4. Inside a `Rules.mk` special variable accessed as `$(d)` is defined. Its value is current directory, use it so if the `Rules.mk` file is moved in the tree it still works without a problem. Caution: this variable is not available in the recipe part and MUST NOT be used there. Use name of the target or prerequisite to extract it if you need it.
5. Make has only one global scope, this means that name conflicts are a thing. Names SHOULD  follow `VAR_NAME_$(d)` convention. There are exceptions from this rule in form of well defined global variables. Examples: General lists `TGT_BIN`, `CLEAN`; General targets: `TEST`, `COVERAGE`; General variables: `GOFLAGS`, `DEPS_GO`.
3. Any rules, definitions or variables that fit some family SHOULD be defined in `mk/$family.mk` file and included from project root `Rules.mk`

(Based on https://evbergen.home.xs4all.nl/nonrecursive-make.html)

---- 

Resolves: https://github.com/ipfs/go-ipfs/issues/3488


Kubuxu:
It finally works again. Teamcity OSX and js-ipfs-api test failures are unrelated.

Kubuxu:
The sharness test with coverage is about 4 min + 2min for coverage report merge procedure.
The unit test with coverage takes about 8 to 10 min.

Kubuxu:
@lgierth @whyrusleeping this should be ready for CR

Kubuxu:
I have figured out smart way to do cross package coverage. The cross package unit test coverage increased the test time only by 10% over normal coverage unit tests.

kevina:
I have some experience with Makefiles, I am also going to have a look.

cmd/ipfs/Rules.mk
```diff
@@ -0,0 +1,29 @@
...
+PATH := $(realpath $(d)):$(PATH)
+
+# disabled for now
+# depend on *.pb.go files in the repo as Order Only (as they shouldn't be rebuilt if exist)
+# DPES_OO_$(d) := diagnostics/pb/diagnostics.pb.go exchange/bitswap/message/pb/message.pb.go
+# DEPS_OO_$(d) += merkledag/pb/merkledag.pb.go namesys/pb/namesys.pb.go
+# DEPS_OO_$(d) += pin/internal/pb/header.pb.go unixfs/pb/unixfs.pb.go
+
+# uses second expansion to collect all $(DEPS_GO)
+$(IPFS_BIN_$(d)): $(d) $$(DEPS_GO) ALWAYS #| $(DEPS_OO_$(d))
```
Kubuxu: I missed the ldflags for the current hash.

kevina:
Also, this looks like it is going to disrupt my workflow so I want a chance to try it.  Added myself as a reviewer.  Should be able to look at it this week.

kevina:
The makefile is a bit slow.  For example:
```
$ /usr/bin/time make
0.92user 0.16system 0:00.96elapsed 114%CPU (0avgtext+0avgdata 18244maxresident)k
8inputs+64outputs (0major+23775minor)pagefaults 0swaps
```

kevina:
@Kubuxu would it be okay if I add some dummy Makefiles in some key sub-directories that just call the top level Makefile?  For example I often do:
```
make -j4 > log
```
in the shareness directory.  I also sometimes do
```
make deps
```
within the shareness directory and then run a failing test manually.  Now I have to type something a lot longer to get the same effect.

bin/Rules.mk
```diff
@@ -0,0 +1,26 @@
+include mk/header.mk
+
+dist_root_$(d)=/ipfs/QmNZL8wNsvAGdVYr8uGeUE9aGfHjFpHegAWywQFEdSaJbp
+
+$(d)/gx: $(d)/gx-v0.9.0
```
whyrusleeping: what is d?

bin/gen-make-dep-chain
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+(for p ; do
```
whyrusleeping: what is going on here? where is p coming from? i'm so confused. what is tac?

circle.yml
```diff
@@ -35,6 +35,6 @@ dependencies:
 
 test:
   override:
-    - case $CIRCLE_NODE_INDEX in 0) make coverage ;; 1) make test_sharness_expensive ;; esac:
+    - case $CIRCLE_NODE_INDEX in 0) make -j 1 coverage/unit_tests.coverprofile && bash <(curl -s https://codecov.io/bash) -cF unittests -X search -f coverage/unit_tests.coverprofile;; 1) make -j 1 coverage/sharness_tests.coverprofile &&  bash <(curl -s https://codecov.io/bash) -cF sharness -X search -f coverage/sharness_tests.coverprofile;; esac:
```
whyrusleeping: does this really need to be a one-liner? Its very... uh... Kubuxuy :P

cmd/ipfs/Rules.mk
```diff
@@ -0,0 +1,29 @@
...
+	$(go-build)
+
+COVER_BIN_$(d) := $(d)/ipfs-test-cover
+CLEAN += $(COVER_BIN_$(d))
+
+$(COVER_BIN_$(d)): GOTAGS += testrunmain
+$(COVER_BIN_$(d)): $(d) $$(DEPS_GO) ALWAYS
+	$(eval TMP_PKGS := $(shell go list -f '{{range .Deps}}{{.}} {{end}}' $(go-flags-with-tags) ./cmd/ipfs | sed 's/ /\n/g' | grep ipfs/go-ipfs | grep -v ipfs/go-ipfs/Godeps) $(call go-pkg-name,$<))
+	$(eval TMP_LIST := $(call join-with,$(comma),$(TMP_PKGS)))
+	@echo go test $@ -c -covermode atomic -coverpkg ... $(go-flags-with-tags) ./$(@D) # for info
```
whyrusleeping: what is `@D` ?

cmd/ipfs/runmain_test.go
```diff
@@ -0,0 +1,30 @@
...
+
+// this abuses go so much that I felt dirty writing this code
+// but it is the only way to do it without writing custom compiler that would
+// be a clone of go-build with go-test
+func TestRunMain(t *testing.T) {
+	args := flag.Args()
+	os.Args = append([]string{os.Args[0]}, args...)
+	ret := mainRet()
+
+	p := os.Getenv("IPFS_COVER_RET_FILE")
```
whyrusleeping: what is this used for?

exchange/bitswap/message/pb/Rules.mk
```diff
@@ -0,0 +1,8 @@
+include mk/header.mk
```
whyrusleeping: is this include relative to the root dir? That feels really weird...

pin/pin.go
```diff
@@ -26,8 +25,9 @@ var emptyKey *cid.Cid
 func init() {
 	e, err := cid.Decode("QmdfTbBqBPQ7VNxZEYEj14VmRuZBkqFbiwReogJgS1zR1n")
 	if err != nil {
-		log.Error("failed to decode empty key constant")
-		os.Exit(1)
+		msg := "failed to decode empty key constant"
+		log.Error(msg)
+		panic(msg)
```
whyrusleeping: whats the reason for this change? I'd prefer not the have the user see an ugly panic stack trace.

cmd/ipfs/Rules.mk
Kubuxu: It is short form of `$(dir $@)`

pin/pin.go
Kubuxu: I didn't want to exit over there as it would change the coverage, but as it shouldn't happen here I will switch back to `Exit`.

bin/gen-make-dep-chain
Kubuxu: This script is not used right now, committed as I thought I would use it, `tac` is cat that reverses lines, `for p;` is equivalent to `for p in "@";`

circle.yml
Kubuxu: I can make it separate script, AFAIK yaml doesn't support multiline keys.

whyrusleeping: Ah, makes sense. Having it be a separate script (with comments explaining its existence) might help readability a bit

Kubuxu:
@kevina 3d84db9 should resolve the `running make in sharness dir` and 6b38da2 cuts down the flat time by half by not running coverage Rules.mk file unless coverage category or clean targets are used.

whyrusleeping:
I'd also like to get @chriscool to take a look over this, he's far more familiar with best practices on this sort of stuff than i am

test/sharness/Makefile
```diff
@@ -1,57 +1,7 @@
...
-
-.PHONY: all clean $(T) aggregate FORCE
-
-# will fail if curl is not installed.
-# TODO: get rid of this and install curl with git or ipfs.
-install_curl_pls = "curl is required to run tests. please install it"
-curl:
-	@which curl >/dev/null || (echo "$(install_curl_pls)" && false)
+ALWAYS: 
+.PHONY: ALWAYS
```
kevina: When I try to just run make in `test/sharness` I get this
```
Makefile:3: target 'trash' given more than once in the same rule
Makefile:3: target 'trash' given more than once in the same rule
Makefile:3: target 'trash' given more than once in the same rule
Makefile:3: target 'trash' given more than once in the same rule
Makefile:3: target 'trash' given more than once in the same rule
Makefile:3: target 'trash' given more than once in the same rule
Makefile:3: target 'trash' given more than once in the same rule
...
rm -rf test/sharness/test-results
*** test/sharness/t0500-issues-and-regressions-offline.sh ***
...
```
That is a get a bunch of warnings and only a single test runs.

Also it seams now every time I run a single test it seams to try and rebuild all the deps:
```
$ time make  t0210-tar.sh
...
make -C ../.. test/sharness/t0210-tar.sh
make[1]: Entering directory '/aux/home/gocode2/src/github.com/ipfs/go-ipfs'
bin/check_go_version 1.7
gx install --global >/dev/null 2>&1
go build -i  -o "test/bin/random" "github.com/ipfs/go-ipfs/Godeps/_workspace/src/github.com/jbenet/go-random/random"
go build -i  -o "test/bin/multihash" "gx/ipfs/QmYDds3421prZgqKbLpEK7T9Aa2eVdQ7o3YarX1LVLdP2J/go-multihash/multihash"
go build -i  -o "test/bin/pollEndpoint" "github.com/ipfs/go-ipfs/thirdparty/pollEndpoint"
go build -i  -o "test/bin/iptb" "gx/ipfs/QmSXCTgRb3JhrV5ALbWocbNGD8Tc3HVHczFYvYKrSQAau3/iptb"
go build -i  -o "test/bin/go-sleep" "github.com/ipfs/go-ipfs/test/dependencies/go-sleep"
go build -i  -o "test/bin/random-files" "github.com/ipfs/go-ipfs/Godeps/_workspace/src/github.com/jbenet/go-random-files/random-files"
go build -i  -o "test/bin/go-timeout" "github.com/ipfs/go-ipfs/test/dependencies/go-timeout"
go build -i  -o "test/bin/hang-fds" "gx/ipfs/QmPRYAPKG9MhfJLrDT659SGhBsY7V7zHq84cngi8EHudHQ/hang-fds"
go build -i  -o "cmd/ipfs/ipfs" "github.com/ipfs/go-ipfs/cmd/ipfs"
rm -rf test/sharness/test-results
*** test/sharness/t0210-tar.sh ***
...
real    0m10.183s
user    0m9.132s
sys     0m1.208s
```

Kubuxu: Ahh, right default target it different.

Kubuxu:
@kevina standalone sharness fixed in a098655

Kubuxu:
Rebuilding all the go deps was a conscious choice, it is minor overhead but increases the consistency of the build and test significantly. If you want to run single test repetitively, you can do it using the direct sharness test call.

Makefile
```diff
@@ -1,195 +1,10 @@
...
 
-GOFLAGS =
-GOTFLAGS =
+PROTOC = protoc --gogo_out=. --proto_path=.:/usr/local/opt/protobuf/include:$(dir $@) $<
 
-export IPFS_REUSEPORT=false
-export GOFLAGS
-export GOTFLAGS
+# enable second expansion
+.SECONDEXPANSION:
```
Kubuxu: We require GNU make either way even right now, I don't know if it is on purpose or not.

For BSD based systems we recommend using `gmake`.

I could do without this flag, I think, not 100% sure right now.

Jenkinsfile
```diff
@@ -0,0 +1,21 @@
+
+node {
+	def VERSION = sh(returnStdout: true, script: "git rev-parse HEAD").trim()
+	def CACHE_DIR = sh(returnStdout: true, script: "mktemp go-ipfs.cache.XXXXXXXX").trim()
```
VictorBjelkholm: Why is the cache separated from docker? With a correct Dockerfile, packages would already be cached from previous layers created.

Kubuxu: It is build cache that contains binary files that are shortcuts for building. It could be a layer if we forced a build in the container creation, but then build failure would be quite cryptic.

mk/footer.mk
```diff
@@ -0,0 +1,2 @@
+d := $(dirstack_$(sp))
+sp := $(basename $(sp))
```
whyrusleeping: what is the purpose of this?

mk/header.mk
```diff
@@ -0,0 +1,4 @@
+# keep track of dirs
+p := $(sp).x
```
whyrusleeping: maybe add a comment about what/why this is?

mk/footer.mk
Kubuxu: it keeps track of directory stack so when one Rules.mk in sub directory exists, the $(d) is restored to previous one in stack. It is standard non-recursive Makefile boilerplate.

unixfs/pb/Rules.mk
```diff
@@ -0,0 +1,8 @@
+include mk/header.mk
+
+PB_$(d) = $(wildcard $(d)/*.proto)
+TGTS_$(d) = $(PB_$(d):.proto=.pb.go)
+
+#DEPS_GO += $(TGTS_$(d))
```
whyrusleeping: comment left here for a reason?

Kubuxu:
IDK if we should go with fully static seed, it allows for same attacks that all languages protect their hash maps against (precalcing which items would go to which buckets and causing requests that would cause many items in one bucket) but in our case it would be causing the tree to have one deep branch.

unixfs/pb/Rules.mk
Kubuxu: I had them as dependencies of go builds but it caused problems with systems not tracking the time properly which tried to rebuild them always (thus requiring protobuf tools).

whyrusleeping:
@Kubuxu Is this ready? It seems to be failing on circleCI

Kubuxu:
This last commit explains small issue I had in Jenkins with codecoverage, but because Jenkins doesn't provide require variables in Pipelines we will have to use Circle for time being.

Makefile
```diff
@@ -1,195 +1,10 @@
-# Minimum version numbers for software required to build IPFS
-IPFS_MIN_GO_VERSION = 1.7
-IPFS_MIN_GX_VERSION = 0.6
-IPFS_MIN_GX_GO_VERSION = 1.1
+# General tools
 
-GOTAGS =
-GOTAGS += "" # we have to have always at least one tag, empty tag works well
+SHELL=PATH=$(PATH) /bin/sh
```
mateon1: Might this be causing the Windows build issue?

bin/Rules.mk
```diff
@@ -0,0 +1,26 @@
...
+
+dist_root_$(d)=/ipfs/QmTazYLCem8B7pQGLsCj1Q4MxkqGMmD9XugvStLJSQ8uxA
+
+$(d)/gx: $(d)/gx-v0.10.0
+$(d)/gx-go: $(d)/gx-go-v1.4.0
+
+TGTS_$(d) := $(d)/gx $(d)/gx-go
+DISTCLEAN += $(wildcard $(d)/gx-v*) $(wildcard $(d)/gx-go-v*) $(d)/tmp
+
+PATH := $(realpath $(d)):$(PATH)
```
mateon1: I'm not sure if this needs quoting, but it's probably safer to do so.

Makefile
Kubuxu: Possibly thanks.

PR #3508, opened 2016-12-14
RFC: Handle Symbolic links in the Gateway

kevina:
Per @Kubuxu @lgierth request.

Closes https://github.com/ipfs/go-ipfs-gateway/issues/6

This is _one_ way to handle it.

Another would be to just have the `Cat` API call resolve the symbolic link.  Another would be to have the gateway resolve the link instead of using a HTTP Redirect.

I like the HTTP redirect because it is simple and better interacts with the browsers cache.  It also avoids any possible security problems from a malicious symbolic link.

### DIFF
--- a/core/coreapi/interface/errors.go
+++ b/core/coreapi/interface/errors.go
@@ -3,6 +3,7 @@ package iface
 import "errors"
 
 var (
-	ErrIsDir   = errors.New("object is a directory")
-	ErrOffline = errors.New("can't resolve, ipfs node is offline")
+	ErrIsDir     = errors.New("object is a directory")
+	ErrOffline   = errors.New("can't resolve, ipfs node is offline")
+	ErrIsSymLink = errors.New("object is a symbolic link")
 )
--- a/core/coreapi/interface/unixfs.go
+++ b/core/coreapi/interface/unixfs.go
@@ -17,4 +17,7 @@ type UnixfsAPI interface {
 
 	// Ls returns the list of links in a directory
 	Ls(context.Context, Path) ([]*ipld.Link, error)
+
+	// ReadSymLink returns the contents of a symbolic link
+	ReadSymLink(context.Context, Path) (string, error)
 }
--- a/core/coreapi/unixfs.go
+++ b/core/coreapi/unixfs.go
@@ -2,12 +2,16 @@ package coreapi
 
 import (
 	"context"
+	"errors"
 	"io"
 
 	coreiface "github.com/ipfs/go-ipfs/core/coreapi/interface"
 	coreunix "github.com/ipfs/go-ipfs/core/coreunix"
+	mdag "github.com/ipfs/go-ipfs/merkledag"
 	uio "github.com/ipfs/go-ipfs/unixfs/io"
+	ftpb "github.com/ipfs/go-ipfs/unixfs/pb"
 
+	proto "gx/ipfs/QmZ4Qi3GaRbjcx28Sme5eMH7RQjGkt8wHxt2a65oLaeFEV/gogo-protobuf/proto"
 	cid "gx/ipfs/QmcZfnkapfECQGcLZaf9B79NRg7cRa9EnZh4LSbkCzwNvY/go-cid"
 	ipld "gx/ipfs/Qme5bWv7wtjUNGsK2BNGVUFPKiuxWrsqrtvYwCLRw8YFES/go-ipld-format"
 )
@@ -38,12 +42,14 @@ func (api *UnixfsAPI) Cat(ctx context.Context, p coreiface.Path) (coreiface.Read
 	}
 
 	r, err := uio.NewDagReader(ctx, dagnode, dget)
-	if err == uio.ErrIsDir {
+	switch err {
+	case uio.ErrIsDir:
 		return nil, coreiface.ErrIsDir
-	} else if err != nil {
-		return nil, err
+	case uio.ErrCantReadSymlinks:
+		return nil, coreiface.ErrIsSymLink
+	default:
 	}
-	return r, nil
+	return r, err
 }
 
 // Ls returns the contents of an IPFS or IPNS object(s) at path p, with the format:
@@ -74,6 +80,30 @@ func (api *UnixfsAPI) Ls(ctx context.Context, p coreiface.Path) ([]*ipld.Link, e
 		links[i] = &ipld.Link{Name: l.Name, Size: l.Size, Cid: l.Cid}
 	}
 	return links, nil
+}
+
+var NotASymLink = errors.New("not a symbolic link")
+
+func (api *UnixfsAPI) ReadSymLink(ctx context.Context, p coreiface.Path) (string, error) {
+	dagnode, err := api.core().ResolveNode(ctx, p)
+	if err != nil {
+		return "", err
+	}
+	switch n := dagnode.(type) {
+	case *mdag.ProtoNode:
+		pb := new(ftpb.Data)
+		if err := proto.Unmarshal(n.Data(), pb); err != nil {
+			return "", err
+		}
+		switch pb.GetType() {
+		case ftpb.Data_Symlink:
+			return string(pb.GetData()), nil
+		default:
+			return "", NotASymLink
+		}
+	default:
+		return "", NotASymLink
+	}
 }
 
 func (api *UnixfsAPI) core() coreiface.CoreAPI {
--- a/core/corehttp/gateway_handler.go
+++ b/core/corehttp/gateway_handler.go
@@ -185,12 +185,15 @@ func (i *gatewayHandler) getOrHeadHandler(ctx context.Context, w http.ResponseWr
 
 	dr, err := i.api.Unixfs().Cat(ctx, resolvedPath)
 	dir := false
+	symLink := false
 	switch err {
 	case nil:
 		// Cat() worked
 		defer dr.Close()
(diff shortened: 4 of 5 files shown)


whyrusleeping:
While I do like this approach, i don't think its quite right. This will solve the problem of reading paths like `foo/bar/symlink`, but I don't think it correctly handles `foo/symlink/bar`. To do that I believe we will have to make changes to the `path.Resolver` code.

kevina:
@whyrusleeping yeah your right, this was a first stab at the problem.  Do you think `coreapi.Cat` should just resolve all symbolic links?  Or should we be required to expand all symbolic links before calling `coreapi.Cat`?

kevina:
I think the answer to the question is if we fix the resolver to follow the links, `coreapi.Cat` would have to follow them.  That will solve this problem, but then how do we actually read a symbolic links contents?

kevina:
So it looks like to do this in the resolver [ResolveLinks](https://github.com/ipfs/go-ipfs/blob/master/path/resolver.go#L121) (in `path/resolver.go`) will need to made aware of symbolic links so it can follow them.   There leads to several issues:
 1) The resolver will need to know about the UnixFS Protobuf Nodes, when right now it only cares about the links.
 2) We we need some way to prevent cycles, it not just that two symbolic links can get in a cycle, but there can be an cycle with a combination of symbolic links and directories since in IPFS we can have the unix equivalent of hard links for directories.
 3) How should we handle ".."?  Since there is no notion of a parent directory in the IPFS UnixFS dir, what ".." resolves too will depend on the path.
 4) Should we always follow symbolic links in the resolver? If not then we need to figure out how to make this configurable.

I see (1) is unavoidable unless we want to introduce some abstract notion of a symbolic link. For (2) we can either explicitly detect cycles or use a length limit, for (3) for now, the safest thing would be to refuse to resolve symbolic links with ".." in it.  For (4) I think it should be configurable, otherwise we might resolve symbolic links in places where it might not be desirable.

kevina:
Note, if we disallow ".." is the symbolic name, I _think_ it will avoid cycles, but i am not 100% sure.  Especially if the path is "/ipns" or something else where the id is not the hash of the content.  If the id is the hash of the content I don't think there is a way to create a cycle as back links are highly improbable (but I suppressible possible, just next to impossible to create).

whyrusleeping:
> The resolver will need to know about the UnixFS Protobuf Nodes, when right now it only cares about the links.

I was moving towards this with my refactor of the resolver that added the `ResolveOnce` field: https://github.com/ipfs/go-ipfs/blob/master/path/resolver.go#L40
We're going to need to do this anyways for sharding, so i think this is a good time to make that leap.

> We we need some way to prevent cycles, it not just that two symbolic links can get in a cycle, but there can be an cycle with a combination of symbolic links and directories since in IPFS we can have the unix equivalent of hard links for directories.

For this, the resolver should simply count the number of symlinks its resolved so far, and error out after a certain number (I think `MAXSYMLINKS` is generally 32, but i'm not sure)

> How should we handle ".."? Since there is no notion of a parent directory in the IPFS UnixFS dir, what ".." resolves too will depend on the path.

We should allow '..', it should be relative to the paths being resolved, and the 'root' should be `/ipfs/`, `/ipfs/../../../../` == `/ipfs/`, similarly to `/` on unix filesystems.

> Should we always follow symbolic links in the resolver? If not then we need to figure out how to make this configurable.

I think we should always follow symlinks in the resolver. (at least in a unixfs context, via the ResolveOnce function discussed above)

kevina:
@whyrusleeping rather than limit the number of symbolic links, what about limiting the number of path components resolved to say 255.  It might be easier to implement that way.

whyrusleeping:
On `ResolveOnce`:

The intent here was to provide a contextual transformation of the graph. For unixfs sharding, the raw graph looked weird, the raw paths would look something like:

```
foo/AB/CDshardedThing/baz
foo/D9/4Efile.mp3
foo/F2/E1otherThing/stuff.txt
foo/F2/E1otherThing/readme.md
```

But when browsing this in a unixfs context, we wanted it to look like:
```
foo/shardedThing/bar
foo/file.mp3
foo/otherThing/stuff.txt
foo/otherThing/readme.md
```

So to transparently perform this transformation, we are going to add context to the resolver in the form of the `ResolveOnce` function that defines how to resolve the next step at each layer of the path.

The dummy 'unixfs resolve once' function is here: https://github.com/ipfs/go-ipfs/blob/master/unixfs/io/resolve.go

Thinking about this in the context of symlinks makes me realize that each resolution is very likely going to want some degree of state, so that we can do things like keep track of the number of symlinks traversed during this resolution, and total path length, and so on.

To that end, i think we should change it from a `ResolveOnce` function, to a `GetResolveStepper` that returns an object (that can contain state) with a `ResolveOnce` method on it. Then in `ResolveLinks` we can create a new resolveStepper and use it iteratively to resolve the path and follow symlinks.

kevina:
@whyrusleeping if we change `ResolveOnce` to `GetResolveStepper` with state then I _think_ things will work out.

kevina:
### `..` in symbolic links

@whyrusleeping (cc @Kubuxu) I think supporting `..` in symbolic links is a bad idea because an IPFS directory has no notation of a parent directory, so `..` is not deterministic.

Here is how it could lead to unexpected results.  Let say we have the following directories:

```
  QmA12:
    QmA14 content/ 
    QmA15 webpage/

  QmA14:
    QmE12: executable.exe

  QmA15:
    QmW12: index.html
    QmS13: executable.exe -> ../context/executable.exe
```

With the idea that index.html contains a link to the `executable.exe` symbolic link.

Now lets say we have some more directories:
```
  QmA20:
    QmA21 context/
    QmA15 webpage/

  QmA21:
    QmX17: executable.exe
    
```

That is the directory QmA20 points to alternate content.  Now then if we retrieve the web page via:
```
  /ipfs/QmA12/webpage/index.html
```
we will get the intended excitable (QmE12), but if we retrieve the web page via
```
  /ipfs/QmA20/webpage/index.html
```
we will get an alternative (QmX17, maybe with malware) executable.

If we instead use:
```
  /ipfs/QmA20/index.html
```
Then the excitable link just won't work.

### Absolute paths in symbolic links

If we decide an absolute link to be rooted at `/ipfs/QmHASH/` the we run into the same problem that we do with `..`.

If instead we define an absolute link to be rooted at `/` or `/ipfs` then we run into a problem that it will be impossible to refer to the current directory hierarchy as the hash is not known yet.

### Conclusions

I am now necessary saying we should not support `..` but I want to make sure everyone is aware of these potential problems.

whyrusleeping:
If Qm20 is constructed to be malicious, but some user trusts it anyways, then there are easier ways to make them download things (like linking directly to the executable). 

We are supporting ".." in symbolic links.

kevina:
@whyrusleeping I am fine with that.

kevina:
We still need to decide if we want to handle absolute and what the semantics of them will be if we do.

whyrusleeping:
@kevina Yeah, i think @Kubuxu was saying we should error if the symlink linked above the root (i.e. `/ipfs/..` is invalid). And i think that makes sense to me, sound good to you?

kevina:
@whyrusleeping I'm fine with that.  But I was more referring to absolute paths in symbolic links and what it should mean.  For example should a link be `/ipfs/HASH/afile.txt`, `/HASH/afile.txt', `/afile.txt` or just not supported.  See my notes above.

whyrusleeping:
Ah! that part, I think a symlink should be able to contain any valid ipfs path, or relative path. We should probably disallow symlinks containing ipns paths, that could definitely end poorly.

I think `/ipfs/HASH` is valid, but `/HASH` shouldnt be.

cc @Kubuxu and @lgierth for input too

Kubuxu:
It should be `/ipfs/CID`, we don't support just `HASH` in many places (gateway is one of them) so I think it should be the same for CID.

mib-kd743naq:
Hi folks!

What is the status of this PR ( and the corresponding https://github.com/ipfs/go-ipfs-gateway/issues/6 )?

I put together an example of an ipfs-ified tarball: https://ipfs.io/ipfs/QmbWphVroGYPVAkRGNHdzYJCrJHcwbd35ygvy32gBUKNpB, and the following still won't work as of writing this message: https://ipfs.io/ipfs/QmbWphVroGYPVAkRGNHdzYJCrJHcwbd35ygvy32gBUKNpB/linkz.tar.extracted/linkz/pointer/sym

Thanks!

jcaesar:
I just wanted to ask: Would it be generally possible that if a path `/A/B/Sym/C/D` is requested, to only resolve it up to `/A/B/Sym` and then throw out a redirect to `/A/B/Symtarget/C/D`? That would leave the problem of circles to the client. What would be the security implications of that?

kevina:
@jcaesar that could work, but it doesn't really simply much, it might makes things more complicated

I am working on this now.  Getting the resolver to support symbolic links is messy and I had several false starts.  Getting it to return enough information so that the symbolic links can be handed outside of the resolver is equally messy.  Will keep working on it until I get something I am happy with.

kevina:
After talking with @whyrusleeping we agreed this will likely involve some API changes, I will figure out what those all and likely propose it separately.

lidel:
Closing as this was created before we had proper security isolation on the web (subdomain gateways, unique origin per content root).

Continued in https://github.com/ipfs/go-ipfs/issues/8437#issuecomment-920086600

PR #3509, opened 2016-12-15
namesys: fix case where there is no cache

Kubuxu:
License: MIT


whyrusleeping:
@Kubuxu Tests? i'm not sure what scenario this actually resolves

Kubuxu:
See https://github.com/ipfs/go-ipfs/blob/master/namesys/routing.go#L98

whyrusleeping:
For some weird alignment of the stars type reason all the osx tests seem to have failed... for different reasons. Rerunning them to try for better

Kubuxu:
OSX passed on Travis

Issue #340, opened 2014-11-15
Add (optional) color to make output of --help easier to read


cleichner:

diasdavid:
@jbenet are you still up to CR @dborzov's PR (https://github.com/ipfs/go-ipfs/pull/377)? We will need to have it rebased by now though, @dborzov would you have availability to do it?

brodo:
I've started working on this (see screenshot).
Currently, I'm checking if the output is a TTY with the isTty function in `cli/parse.go`. The problem is checking the config. Currently the config is only read on demand. If we want to have the option in the config, it needs to be read it directly in `ipfs/main.go`. My personal preference would be not to put the option in the config file, but introduce another command line flag like `--color`.

If you guys still prefer the config file, I need some help in how to restructure the starting process.

![colorful-help](https://cloud.githubusercontent.com/assets/377911/21111044/fe96561c-c09f-11e6-860a-ea56b5d73b19.png)

hsanjuan:
I like `--color`, but let's see if others can weight in. Maybe there are considerations for having this in the config.

briantigerchow:
> If we want to have the option in the config, it needs to be read it directly in ipfs/main.go.

Perhaps, when no config exists, the program can be treated as though the color option is not specified.

PR #3512, opened 2016-12-16
Colorful help (#340) and some help refactoring

brodo:
Hello, 
this PR is related to issue #340 but also does some other things.

## `Subcommand` map is now an array of `CmdInfos`
- Having an array allows people to change the position of commands in the help text (maps are not ordered).
- Commands now may have a command group. (needed for the main help message but also useful for other commands with many sub-commands)

## HelpText now has an `AdditionalHelp` field
- AdditionalHelp is displayed on the bottom of the help text. It is currently only used in the main help message, but may also be useful for other things.

## There is now a --color command line flag
- Colorful can be enabled with `--color`
- There are no checks if stdout is a tty. If you use --color, you get color! (e.g. when someone wants to pipe it into `less -R`)

This PR touches a lot of files, but most changes are just simple gofmt transformations in order to switch from the `Subcommand` map to the `Subcommand` array.

Cheers,
Julian

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # go-ipfs changelog
 
+### 0.4.4 - 2016-10-11
+
+This release contains an important hotfix for a bug we discovered in how pinning works.
+If you had a large number of pins, new pins would overwrite existing pins.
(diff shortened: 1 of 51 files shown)


brodo:
I'm doing a new PR with all the CI issues fixed.

PR #3513, opened 2016-12-16
coreunix: fix Add() docs regarding pinning

### LINKED ISSUE #3470: coreunix.Add is documented to pin objects but it doesn't
#### Version information:
```
go-ipfs version: 0.4.4-
Repo version: 4
System version: amd64/linux
Golang version: go1.7
```

#### Description:
The documentation of the [coreunix.Add](https://github.com/ipfs/go-ipfs/blob/master/core/coreunix/add.go#L254) method claims to pin the object to the local data store but it doens't seem to do so.

After adding a file using :
```go
// r is a reader from a file
hash, err := coreunix.AddWithContext(ctx, service.node, r)
if err != nil {
	panic("error adding file")
}
fmt.Println(hash)
```

I'm trying to make sure the object is pinned using `ipfs pin ls` but this command returns nothing
I can cat the file with `ipfs cat <hash>` but after doing `ipfs repo gc` 
trying to cat the file again fails with `Error: merkledag: not found` which proof the object was not pinned

lgierth:
Fixes #3470

### DIFF
--- a/core/coreunix/add.go
+++ b/core/coreunix/add.go
@@ -89,7 +89,7 @@ func NewAdder(ctx context.Context, p pin.Pinner, bs bstore.GCBlockstore, ds dag.
 
 }
 
-// Internal structure for holding the switches passed to the `add` call
+// Adder holds the switches passed to the `add` command.
 type Adder struct {
 	ctx        context.Context
 	pinning    pin.Pinner
@@ -114,7 +114,7 @@ func (adder *Adder) SetMfsRoot(r *mfs.Root) {
 	adder.mr = r
 }
 
-// Perform the actual add & pin locally, outputting results to reader
+// Constructs a node from reader's data, and adds it. Doesn't pin.
 func (adder Adder) add(reader io.Reader) (node.Node, error) {
 	chnk, err := chunk.FromString(reader, adder.Chunker)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


core/coreunix/add.go
```diff
@@ -251,12 +253,14 @@ func (adder *Adder) outputDirs(path string, fsn mfs.FSNode) error {
 	}
 }
 
-// Add builds a merkledag from the a reader, pinning all objects to the local
-// datastore. Returns a key representing the root node.
+// Builds a merkledag node from a reader, adds it to the blockstore,
```
hsanjuan: This is not golint compliant

core/coreunix/add.go
```diff
@@ -251,12 +253,14 @@ func (adder *Adder) outputDirs(path string, fsn mfs.FSNode) error {
...
+// and returns the key representing that node.
+// If you want to pin it, use NewAdder() and Adder.PinRoot().
 func Add(n *core.IpfsNode, r io.Reader) (string, error) {
 	return AddWithContext(n.Context(), n, r)
 }
 
+// Same as Add(), but with a custom context.
```
hsanjuan: Also not golint compliant

core/coreunix/add.go
lgierth: Fixed this and the other, and I now have golint installed which looks useful :+1:

Issue #3496, opened 2016-12-09
OSX Teamcity tests leave `nc` process running


whyrusleeping:
The OSX teamcity tests will occasionally all start failling because some test run failed to clean up a netcat listener (on port 5005 I think). We should figure out where/why this is happening and fix it.

whyrusleeping:
This appears to be in `t0235`:

```
14:     go-sleep 0.5s | nc -l 5005 > nc_out &
```

I'm not 100% sure, but i don't believe processes necessarily get killed if their input process dies.

lgierth:
Might need `-o pipefail`

whyrusleeping:
@lgierth i don't think that would do it. The problem is that the `nc` process just isnt dying. setting pipefail would make just make it return non-zero *when* it exits *if* `go-sleep` returns a non-zero exit code.

whyrusleeping:
cc @chriscool and @Kubuxu

Kubuxu:
we can have test after it that always kills it.

but apart from that we should have teamcity run some sort of containers (docker, lxc, lxd, whatever), and scrap the whole container after test is done.

whyrusleeping:
@Kubuxu Yeah, we should be. But i'm not aware of any containerization stuff that works on OSX. I read that docker supports osx 'natively' but i'm not sure whether or not that means the OS inside the container will still be osx.

chriscool:
I would suggest killing the process explicitly like this:

```
test_expect_success "can start nc server" '
	nc -l 5005 > nc_out &
'

test_expect_success "can make http request against nc server" '
	NC_PID=$! &&
	( ipfs cat /ipfs/Qmabcdef --api /ip4/127.0.0.1/tcp/5005 || true ) &&
	go-sleep 0.5s &&
	test_kill_repeat_10_sec $NC_PID
'
```

Kubuxu:
This might be a solution: https://paolozaino.wordpress.com/2015/08/04/how-to-run-your-applications-in-a-mac-os-x-sandbox-to-enhance-security/

Kubuxu:
This might be also general solution for those sorts of problems:
```
trap "exit" INT TERM
trap "kill 0" EXIT
```

PR #3515, opened 2016-12-16
Fixes nc on macOS not closing socket when the stdin sends EOF

Kubuxu:
Uses `-d` option that is supported both by BSD and GNU nc.

Resolves: https://github.com/ipfs/go-ipfs/issues/3496

License: MIT

### DIFF
--- a/test/sharness/t0235-cli-request.sh
+++ b/test/sharness/t0235-cli-request.sh
@@ -11,7 +11,9 @@ test_description="test http requests made by cli"
 test_init_ipfs
 
 test_expect_success "can make http request against nc server" '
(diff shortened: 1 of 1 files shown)


Kubuxu:
The test won't pass until it is killed manually and all PRs use this as a base.

whyrusleeping:
This LGTM, at some point @jbenet will find time to restart the cjdns daemon on the OSX runner and we can move forward with this.

I really need to figure out how to make a startup script for osx

Kubuxu:
@whyrusleeping this could be merged before that as any PR that has base older than this will cause the hang again.

Issue #3209, opened 2016-09-12
Suborigin header must be lowercase


hobofan:
Version/Platform/Processor information (from `ipfs version --all`):

```
go-ipfs version: 0.4.4-dev-0e2b4eb
Repo version: 4
System version: amd64/linux
Golang version: go1.7.1
```

Type (bug, feature, meta, test failure, question): **feature**/**bug**
Area (api, commands, daemon, fuse, etc): **gateway**
Priority (from P0: operations on fire, to P4: functioning): **P2**

Description:

Using Chromium (`Version 53.0.2785.101 (64-bit)`) and the current Chrome for Android version I get the error `Error with Suborigin header: Invalid character 'Q' in suborigin.` when opening any path on the gateway. It appears that this error stops any JS on the site from executing. I am testing it with a single page web app that works fine when the assets are served from a static webserver (e.g. `python -m SimpleHTTPServer`).

The request as made with curl:

```
curl https://ipfs.io/ipfs/QmSA76DiRmaBGDvW2J3CQtS94tC7UdLouKV11b4QWnTG9f/ --verbose
*   Trying 104.236.179.241...
* TCP_NODELAY set
* Connected to ipfs.io (104.236.179.241) port 443 (#0)
* ALPN, offering http/1.1
* Cipher selection: ALL:!EXPORT:!EXPORT40:!EXPORT56:!aNULL:!LOW:!RC4:@STRENGTH
* successfully set certificate verify locations:
*   CAfile: /etc/ssl/certs/ca-certificates.crt
  CApath: none
* TLSv1.2 (OUT), TLS header, Certificate Status (22):
* TLSv1.2 (OUT), TLS handshake, Client hello (1):
* TLSv1.2 (IN), TLS handshake, Server hello (2):
* NPN, negotiated HTTP1.1
* TLSv1.2 (IN), TLS handshake, Certificate (11):
* TLSv1.2 (IN), TLS handshake, Server key exchange (12):
* TLSv1.2 (IN), TLS handshake, Server finished (14):
* TLSv1.2 (OUT), TLS handshake, Client key exchange (16):
* TLSv1.2 (OUT), TLS change cipher, Client hello (1):
* TLSv1.2 (OUT), TLS handshake, Unknown (67):
* TLSv1.2 (OUT), TLS handshake, Finished (20):
* TLSv1.2 (IN), TLS change cipher, Client hello (1):
* TLSv1.2 (IN), TLS handshake, Finished (20):
[...]

PR #3519, opened 2016-12-19
gateway: disable Suborigins as it conflicts the spec

Kubuxu:
The Suborigins spec was changed and we have to adjust, the spec is still
unstable and it might change in future. Currently the only browser
supporting it (Chrome) errors out on it as it doesn't confront spec it
uses.

See https://github.com/ipfs/specs/issues/131

Part of https://github.com/ipfs/go-ipfs/issues/3209

Proper solution requires: deciding on suborigin prefix, propagating go-multibase changes and exposing cdi.EncodeWithBase.

We can do that but I would want this to be fixed in 0.4.5 and we want to release very soon.

License: MIT


lgierth:
LGTM, thanks!

lgierth:
To fix the code coverage it's okay to just comment or remove the code in question. We have revision control and the new code will look different anyhow :)

Kubuxu:
Yeah, I will just remove it.

PR #3523, opened 2016-12-19
bootstrap: add a few /ip6 nodes

lgierth:
As the bootstrappers now have proper IPv6, we can start making use of it. Since configuring it over the weekend, IPv6 usage has already roughly quadrupled.

### DIFF
--- a/repo/config/bootstrap_peers.go
+++ b/repo/config/bootstrap_peers.go
@@ -14,15 +14,23 @@ import (
 // NOTE: This is here -- and not inside cmd/ipfs/init.go -- because of an
 // import dependency issue. TODO: move this into a config/default/ package.
 var DefaultBootstrapAddresses = []string{
-	"/ip4/104.131.131.82/tcp/4001/ipfs/QmaCpDMGvV2BGHeYERUEnRQAwe3N8SzbUtfsmvsqQLuvuJ",  // mars.i.ipfs.io
-	"/ip4/104.236.176.52/tcp/4001/ipfs/QmSoLnSGccFuZQJzRadHn95W2CrSFmZuTdDWP8HXaHca9z",  // neptune.i.ipfs.io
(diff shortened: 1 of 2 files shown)


whyrusleeping:
@lgierth do you want this in 0.4.5?

lgierth:
Ideally yes, I just wasn't sure if you had any reservations :)

lgierth:
Is there any place where we explicitly prefer /ip6 over /ip4? That'd be useful.

jbenet:
yay

whyrusleeping:
@lgierth that happens in libp2p in the addrutil sorter: https://github.com/libp2p/go-libp2p-peerstore/blob/master/addr/sorting.go

mib-kd743naq:
Linking here as very tightly related: https://github.com/ipfs/ipfs/issues/221

The issue isn't theoretical - I actually was in a situation like this when I filed the ticket ( fully open 80/443 outbound, everything else was a hard REJECT )

whyrusleeping:
@lgierth do we want to take this opportunity to add some 'standard' port bootstrappers?

lgierth:
Do you mean 4737?

Kubuxu:
I think he meant 443 and 80.

lgierth:
I dunno -- we'll have /ws on ipfs.io:80 and /wss on ipfs.io:443 very soon, I figure that'd work?

whyrusleeping:
@lgierth okay, I think we can merge this as is for 0.4.6 then, yeah?

mib-kd743naq:
> I dunno -- we'll have /ws on ipfs.io:80 and /wss on ipfs.io:443 very soon, I figure that'd work?

@lgierth will this fully address https://github.com/ipfs/ipfs/issues/221#issue-198335978 ( given one uses a `0.4.<distantfuture>` go-ipfs instance )? Or will the ws-based transport be useful for the js-ipfs exclusively?

lgierth:
> okay, I think we can merge this as is for 0.4.6 then, yeah?

Yes let's do it! :+1: 

> Or will the ws-based transport be useful for the js-ipfs exclusively?

@mib-kd743naq I think it's pretty useful generally -- on the outside it just looks like regular HTTPS.

PR #3538, opened 2016-12-26
keystore: Property based testing

mateon1:
I'm playing around with property based testing, this PR implements a test that randomly performs `Get(key)`, `Put(key, value)`, `Delete(key)` and `List()` operations, and compares results from the keystore and its internal model. (a map of strings to keys)
On my Windows machine the test takes 17s to run, since gopter performs the test 100 times by default.

### DIFF
--- a/keystore/keystore_test.go
+++ b/keystore/keystore_test.go
@@ -9,10 +9,10 @@ import (
 	"strings"
 	"testing"
 
-	ci "gx/ipfs/QmfWDLQjGjVe4fr5CoztYW2DYYjRysMJrFe1RCsXLPTf46/go-libp2p-crypto"
 	"github.com/leanovate/gopter"
 	"github.com/leanovate/gopter/commands"
 	"github.com/leanovate/gopter/gen"
+	ci "gx/ipfs/QmfWDLQjGjVe4fr5CoztYW2DYYjRysMJrFe1RCsXLPTf46/go-libp2p-crypto"
 )
 
 type rr struct{}
@@ -239,6 +239,7 @@ func ValidKey(key string) bool {
 }
 
 type getCommand string
+
 func (cm *getCommand) Run(store commands.SystemUnderTest) commands.Result {
 	val, err := store.(*FSKeystore).Get(string(*cm))
 	if err != nil {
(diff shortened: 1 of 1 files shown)


whyrusleeping:
Hrm... This is pretty nice to have. I wonder if we should have this as part of the main go-ipfs codebase, or a separate 'testing' repo that can have must larger and longer tests.

I also think we need to take a look at the library in use here to make sure its the one we want to move forward with.

cc @lgierth @Kubuxu @hsanjuan

mateon1:
Regarding using Gopter for property tests, there exists a builtin alternative (`testing/quick`), but it doesn't support some common features for this kind of library, like shrinking (AKA finding the smallest failing testcase). The Gopter README explains it better: https://github.com/leanovate/gopter/#synopsis

whyrusleeping:
Hey @mateon1 thanks for this :) However i'm not sure we want to add 17 seconds (definitely more on slow CI boxes) to the tests, plus i'm always leery of adding new dependencies.

That said, gopter seems pretty cool. Any insight on how it compares to https://github.com/dvyukov/go-fuzz ?

mateon1:
I somehow lost @whyrusleeping 's reply in my notifications, sorry.

There is a difference between fuzzing, and property testing.

In this PR, I use a pattern I first saw in Erlang automated testing, where you QuickCheck a series of commands, along with a simple state machine of what you expect to happen if these commands run. If there is a mismatch, that might be a potential bug.
Property testing can also be used to find multiple edge cases for simple functions using a single property, where you would normally write several tests (and possibly miss tricky cases! nil, empty string, zero, small integers, negative integers, integers near overflow, float Infinity/NaN, empty array, ..., duplicate calls for Add/Delete for the same key, ...).
Fuzzing is more appropriate (I feel) to file parsers/network handling code, where the fuzzer can intelligently find edge cases that do bad things in the (bigger) program.
While some of the bugs QC and fuzzing find may be the same, they are ultimately different methods.

I don't mind if this is closed and forgotten, as it was mostly just a go at learning gopter (and learning Go).

By the way, I'm probably going to have some fun with the go-fuzz thing as well :)

Kubuxu:
@mateon1 See my usage of go-fuzz for example in go-cid

Kubuxu:
@mateon1 I am trying to clean up PRs so I will close this.

PR #3541, opened 2016-12-27
Colorful help (#340) and some help refactoring

brodo:
Hello, 
this PR is related to issue #340 but also does some other things.

## `Subcommand` map is now an array of `CmdInfos`
- Having an array allows people to change the position of commands in the help text (maps are not ordered).
- Commands now may have a command group. (needed for the main help message but also useful for other commands with many sub-commands)

## HelpText now has an `AdditionalHelp` field
- AdditionalHelp is displayed on the bottom of the help text. It is currently only used in the main help message, but may also be useful for other things.

## There is now a --color command line flag
- Colorful can be enabled with `--color`
- There are no checks if stdout is a tty. If you use --color, you get color! (e.g. when someone wants to pipe it into `less -R`)

This PR touches a lot of files, but most changes are just simple gofmt transformations in order to switch from the `Subcommand` map to the `Subcommand` array.

Cheers from 33c3,
Julian

### DIFF
--- a/commands/cli/helptext_test.go
+++ b/commands/cli/helptext_test.go
@@ -4,7 +4,9 @@ import (
 	"strings"
 	"testing"
 
+	"bytes"
 	cmds "github.com/ipfs/go-ipfs/commands"
+	"github.com/ipfs/go-ipfs/core/commands"
 )
 
 func TestSynopsisGenerator(t *testing.T) {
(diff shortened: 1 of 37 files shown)


commands/cli/helptext_test.go
```diff
@@ -4,7 +4,9 @@ import (
 	"strings"
 	"testing"
 
+	"bytes"
```
Kubuxu: this package should be joined with those above, this is stdlib package

brodo: The problem with making it part of the Cmd struct is that there is one command, which has two names. (i forgot which it was) If we want to keep this feature, we can not put the name in there there.

brodo:
@whyrusleeping Ok, I'll first divide this PR into three smaller ones. All the other things then can be discussed separately in the three PRs.

Issue #3537, opened 2016-12-25
Keys from keystore are not republished to DHT


Kubuxu:
Needs to be added to `node.IpnsRepub.AddName`.

We need to probably also store the last published entry to keep it through node resets.

wigy-opensource-developer:
Thanks, Kuba. I just added a comment to #3472 about this missing part.

whyrusleeping:
Lets definitely try and get this done in 0.4.8, Don't want the keystore feature to remain semi-gimped for long

whyrusleeping:
more discussion in https://github.com/ipfs/go-ipfs/issues/3808

whyrusleeping:
@Kubuxu this is done now, right?

PR #3547, opened 2016-12-29
Added /name/upload command definition

wigy-opensource-developer:
As part of adding the [light client feature](https://github.com/ipfs/notes/issues/206), a new command `name upload` is added to the admin interface that allows mobile clients to ask an IPFS node to publish an IPNS record for them. Putting this to the admin interface is an intermediate solution. Authorization (do I let someone with that key that signed this record use my node) is now left to some other process that has access to the admin interface.

Renewing (extending EOL) of these records cannot be done by the IPFS node, because the private key is not shared with it. As a consequence, EOL will not be 24hrs from the signing, but about a month from it. TTL is not affected by this limitation and should only depend on the expected frequency of updates to that name.

Republishing the same record periodically on the DHT is missing, like for those that are published with `ipfs name publish --key notself` #3537

As part of this change, namesys/republisher now reuses code from the publisher. Also, I have worked on the sharness tests using FUSE, but some are still failing on Linux.

### DIFF
--- a/core/commands/name.go
+++ b/core/commands/name.go
@@ -28,11 +28,6 @@ Publish an <ipfs-path> to your identity name:
   > ipfs name publish /ipfs/QmatmE9msSfkKxoffpHwNLNKgwZG8eT9Bud6YoPab52vpy
   Published to QmbCMUZw6JFeZ7Wp9jkzbye3Fzp2GGcPgC3nmeUjfVF87n: /ipfs/QmatmE9msSfkKxoffpHwNLNKgwZG8eT9Bud6YoPab52vpy
 
-Publish an <ipfs-path> to another public key:
-
-  > ipfs name publish /ipfs/QmatmE9msSfkKxoffpHwNLNKgwZG8eT9Bud6YoPab52vpy QmbCMUZw6JFeZ7Wp9jkzbye3Fzp2GGcPgC3nmeUjfVF87n
-  Published to QmbCMUZw6JFeZ7Wp9jkzbye3Fzp2GGcPgC3nmeUjfVF87n: /ipfs/QmatmE9msSfkKxoffpHwNLNKgwZG8eT9Bud6YoPab52vpy
-
 Resolve the value of your identity:
 
   > ipfs name resolve
@@ -53,6 +48,7 @@ Resolve the value of a reference:
 
 	Subcommands: map[string]*cmds.Command{
 		"publish": PublishCmd,
+		"upload":  UploadNameCmd,
 		"resolve": IpnsCmd,
 	},
 }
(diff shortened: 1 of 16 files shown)


core/commands/name.go
```diff
@@ -53,6 +48,7 @@ Resolve the value of a reference:
 
 	Subcommands: map[string]*cmds.Command{
 		"publish": PublishCmd,
+		"upload":  UploadNameCmd,
```
whyrusleeping: upload probably isnt the best name for this feature... maybe something like `publish-record` could be used, since its not intented as a common thing?

cc @Kubuxu @diasdavid @lgierth

diasdavid: Given that record stores will be simple stores with get/put interfaces, something like `/record/put` sounds like a good name to me

whyrusleeping:
@wigy-opensource-developer Any update here?

wigy-opensource-developer:
I was waiting for the 0.4.5 release madness to settle a bit before rebasing, but the current pace is even faster. Wow. My goal is to get this upstream, but I waste quite some time with merging imports that were updated. Excuse me for the slowness.

whyrusleeping:
@wigy-opensource-developer sounds good. Lets also go with calling this `ipfs name put` for now.

whyrusleeping:
@wigy-opensource-developer also, could you PR your sharness and travis.yml changes separately? Smaller PRs == easier for us to review == faster merges

wigy-opensource-developer:
Roger. I am planning to spend 2 days on it next week.

whyrusleeping:
@wigy-opensource-developer update here? This would be a really nice feature to have

wigy-opensource-developer:
Sorry for losing focus on my 2 pull requests, but I am getting back to these soon. Recently I am swapping too many hats. And you are pushing out releases like crazy. Much releases, very impressive. Wow. Doge approved.

Kubuxu:
@wigy-opensource-developer do you plan to continue it?

wigy-opensource-developer:
@Kubuxu Hi Jakub! Without any consensus on #3540 it makes no sense for me to rebase this ancient PR onto the recent codebase. I have also did a responsible disclosure on a security flaw that someone without the private key for an IPNS record could make it impossible for the owner to ever successfully upgrade the record. No answer arrived for that.

So however famous IPFS is, the IOP project is implementing our own DHT that can still integrate with IPFS, but does not rely on getting you on the same page in terms of possible use-cases and security questions.

It does not leave a bad taste or anger, but we had to move on somehow. Feel free to contact me on how to improve communication with your team, because we still appreciate all the efforts you put into decentralization.

PR #3550, opened 2016-12-29
Move fetcher feeding to separate goroutine to avoid deadlocking

### LINKED ISSUE #3505: ipfs pin add is very slow / hanging
#### Version information:
go-ipfs version: 0.4.4-
Repo version: 4
System version: amd64/linux
Golang version: go1.7

#### Type:
Bug

#### Priority:
P3
#### Description:

I have a directory with 25G files in IPFS, to which I'm adding another file with 4GB. After adding, I'm trying to pin new directory version. 

```
ipfs pin add QmPfmRdxrBL3nV9gChCMfv9uAbQSGmh1mcZ6Y1BfwdTBFd
# Actively churning disk for some time, lots of ipfs daemon threads competing for IO, then process is consuming ~15% CPU time indefinitely. I left it running for 10 hours, then killed
# However...
ipfs refs -r QmPfmRdxrBL3nV9gChCMfv9uAbQSGmh1mcZ6Y1BfwdTBFd > /dev/null
# Done in ~15 minutes.

Voker57:
`EnumerateChildrenAsync` is sending slices of keys to `fetchNodes`, but receives only one node per iteration, which, when coupled with recent addition of concurrency limit, can lead to deadlock when `toprocess` is full but `nodes` is empty. This patch defers feeding `toprocess` to separate (unbounded) goroutines to allow concurrent IO between the two functions.

Fixes #3505

### DIFF
--- a/merkledag/merkledag.go
+++ b/merkledag/merkledag.go
@@ -392,10 +392,14 @@ func EnumerateChildren(ctx context.Context, ds LinkService, root *cid.Cid, visit
 func EnumerateChildrenAsync(ctx context.Context, ds DAGService, c *cid.Cid, visit func(*cid.Cid) bool) error {
 	toprocess := make(chan []*cid.Cid, 8)
 	nodes := make(chan *NodeOption, 8)
+	var fetcherWg sync.WaitGroup
 
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
-	defer close(toprocess)
+	defer func() {
+		fetcherWg.Wait()
+		close(toprocess)
+	}()
 
 	go fetchNodes(ctx, ds, toprocess, nodes)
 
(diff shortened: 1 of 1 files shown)


Kubuxu:
Could you make it panic safe and note use recover?

Voker57:
@Kubuxu done. CI seems to be broken but my test passed

whyrusleeping:
@Kubuxu lets make sure to review this one thoroughly, this part of the codebase always feels a tad fragile to me

Kubuxu:
It does and is, as the semaphore change shows.

whyrusleeping:
@Voker57 Thanks for this fix, but we've decided that this part of the codebase needed to just get rewritten. I'll close this in favor of #3571 (and please do make sure that PR looks good to you)

Voker57:
closed in favor of #3598

PR #3555, opened 2017-01-02
add fetch flag to pin add and Pin() functions

Voker57:
this allows pinning without fetching data in case it's not needed

### DIFF
--- a/assets/assets.go
+++ b/assets/assets.go
@@ -75,7 +75,7 @@ func addAssetList(nd *core.IpfsNode, l []string) (*cid.Cid, error) {
 		return nil, fmt.Errorf("assets: DAG.Add(dir) failed: %s", err)
 	}
 
-	if err := nd.Pinning.Pin(nd.Context(), dir, true); err != nil {
+	if err := nd.Pinning.Pin(nd.Context(), dir, true, true); err != nil {
 		return nil, fmt.Errorf("assets: Pinning on init-docu failed: %s", err)
 	}
 
--- a/core/commands/pin.go
+++ b/core/commands/pin.go
@@ -44,6 +44,7 @@ var addPinCmd = &cmds.Command{
 	},
 	Options: []cmds.Option{
 		cmds.BoolOption("recursive", "r", "Recursively pin the object linked to by the specified object(s).").Default(true),
+		cmds.BoolOption("fetch", "f", "Fetch the object(s) before pinning").Default(true),
 	},
 	Type: PinOutput{},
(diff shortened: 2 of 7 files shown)


Kubuxu:
The fetch is needed for the GC to function correctly, if whole data isn't present from a pin tree GC will throw an error to prevent loss of data (one missing block could cause removal of TBs of data).

The fetch will check local repo first so it won't be doing unnecessary fetches.

Is that what you were going for?

Voker57:
No, I'm going for speeding up pinning of already local data. I'm doing repeat pins on updated directory and checking for data in local repo is taking vast majority of time (34m for each iteration currently)

Kubuxu:
That is necessary operation for guarantee we want to provide, it also just means that we should improve performance of this process instead.

How big is the data your are pinning and the repo itself?

whyrusleeping:
I think this is probably okay to add, *BUT* we need to add a warning that using this when you don't actually have all blocks local will cause repo corruption.

Kubuxu:
I can +1 on that with very clear warnings and option with name similar to `--i-know-what-I-am-doing-just-let-me-do-it`.

It is potentially more damaging than block rm command.

Voker57:
Added warning. Is scary name really necessary?

whyrusleeping:
I'm still moderately uncomfortable having this named fetch. I do think it should be named something a little scarier. Any thoughts @Kubuxu ?

In reality, the need for this can be alleviated by making `FetchGraph` faster

kevina:
> It is potentially more damaging than block rm command.

Yes.  And in that case @whyrusleeping insistent I check for pins even though it can be very slow.  I has an option to avoid the pin check but I removed it die to @whyrusleeping which I somewhat agree with.

whyrusleeping:
@kevina Yeah... I think i'm really leaning towards closing this PR.

@Voker57 With the recent improvements to the enumerate code, is the perf in an acceptable range for you?

Voker57:
@whyrusleeping absolutely unacceptable, only link caching might help.

whyrusleeping:
@Voker57 Would an `ipfs pin update` command help with your usecase? It would short circuit fetching things that are already pinned by a given pin. the usecase here is for updating pins on large datasets as they change over time

Voker57:
@whyrusleeping yes, it would mostly solve the problem.

whyrusleeping:
@Voker57 Great! We will get that done in the next release

Voker57:
Not much point in this now, especially if #3642 gets merged.

PR #3557, opened 2017-01-03
Fix "ipfs ls" so it works correctly with raw leaves.

### LINKED ISSUE #3548: Recursive add with --raw-leaves produces inaccessible directory
#### Version information:
`go-ipfs version: 0.4.5-dev-99d72c2`

#### Type:
`Bug`

#### Priority:
`P1`

#### Description:

```
~$ mkdir somedir
~$ echo bar > somedir/foo
```
```
~$ ipfs add --raw-leaves -r somedir/
added zb2rhf6GzX4ckKZtjy8yy8iyq1KttCrRyqDedD6xubhY3sw2F somedir/foo
added QmThNTdtKaVoCVrYmM5EBS6U3S5vfKFue2TxbxxAxRcKKE somedir
```
```
~$ ipfs ls QmThNTdtKaVoCVrYmM5EBS6U3S5vfKFue2TxbxxAxRcKKE
Error: incorrectly formatted merkledag node: Unmarshal failed. unexpected EOF
```
```
~$ ipfs add -r somedir/
added QmTz3oc4gdpRMKP2sdGUPZTAGRngqjsi99BPoztyP53JMM somedir/foo
added QmZKetgwm4o3LhNaoLSHv32wBhTwj9FBwAdSchDMKyFQEx somedir
```
```
~$ ipfs ls QmZKetgwm4o3LhNaoLSHv32wBhTwj9FBwAdSchDMKyFQEx
QmTz3oc4gdpRMKP2sdGUPZTAGRngqjsi99BPoztyP53JMM 12 foo
```

kevina:
Closes #3548

### DIFF
--- a/test/sharness/t0045-ls.sh
+++ b/test/sharness/t0045-ls.sh
@@ -111,4 +111,59 @@ test_ls_cmd
 test_ls_cmd_raw_leaves
 test_kill_ipfs_daemon
 
+#
+# test for ls --resolve-type=false
+#
+
+test_expect_success "'ipfs add -r' succeeds" '
+	mkdir adir &&
+	# note: not using a seed as the files need to have truly random content
+	random 1000 > adir/file1 &&
+	random 1000 > adir/file2 &&
+	ipfs add --pin=false -q -r adir > adir-hashes
+'
+
+test_expect_success "get hashes from add output" '
+	FILE=`head -1 adir-hashes` &&
+	DIR=`tail -1 adir-hashes` &&
+	test "$FILE" -a "$DIR"
+'
+
+test_expect_success "remove a file in dir" '
+	ipfs block rm $FILE
+'
+
+test_expect_success "'ipfs ls --resolve-type=false ' ok" '
+	ipfs ls --resolve-type=false $DIR > /dev/null
+'
+
+test_expect_success "'ipfs ls' fails" '
+	test_must_fail ipfs ls $DIR
+'
+
+test_launch_ipfs_daemon --offline
+
+test_expect_success "'ipfs ls --resolve-type=false' ok" '
+	ipfs ls --resolve-type=false $DIR > /dev/null
+'
+
+test_expect_success "'ipfs ls' fails" '
+	test_must_fail ipfs ls $DIR
+'
+
+test_kill_ipfs_daemon
+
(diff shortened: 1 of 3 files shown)


kevina:
Note, the first commit "Add test for adding directory with --raw-leaves option." is not strictly required, but still I good idea to get more coverage when the "--raw-leaves" option is used (I wrote it thinking `ipfs ls` was tested in that file, was mistaken, but decided to keep it anyway).

kevina:
Note: There doesn't seam to be any tests for the `--resolve-type` option in the `ipfs ls` command.  Before I am comfortable with this comment I will need to add some tests for that as I needed to modify that code to fix #3548 and I just caught a bug in the p.r.

kevina:
Okay this should be ready for review now.

Kubuxu:
The 3526c26 is before the fix, does it cause the tests to fail? If it does it should be moved to after the fix?

This way git-bisect will still work.

test/sharness/t0045-ls.sh
```diff
@@ -90,12 +90,84 @@ test_ls_cmd() {
...
+
+test_launch_ipfs_daemon
+
+# now we try `ipfs ls --resolve-type=false` with the daemon online It
+# should not even attempt to retrieve the file from the network.  If
+# it does it should eventually fail as the content is random and
+# should not exist on the network, but we don't want to wait for a
+# timeout so we will kill the request after a few seconds
+test_expect_success "'ipfs ls --resolve-type=false' ok and does not hang" '
+	ipfs ls --resolve-type=false $DIR > /dev/null &
```
Kubuxu: You can use `go-timeout` here.

kevina:
> The 3526c26 is before the fix, does it cause the tests to fail? If it does it should be moved to after the fix?

No.  It was an extra test that turned out to be unnecessary, but I decided to use it anyway.

test/sharness/t0045-ls.sh
```diff
@@ -90,12 +90,84 @@ test_ls_cmd() {
...
+	PID=$!
+	sleep 2
+	kill $PID 2> /dev/null
+	wait $PID
```
kevina: Will do.

kevina: Fixed.

PR #3569, opened 2017-01-06
set the agent version based on version string

whyrusleeping:
License: MIT


core/core.go
```diff
@@ -86,6 +87,10 @@ const (
 	onlineMode
 )
 
+func init() {
+	identify.ClientVersion = "go-ipfs/" + config.CurrentVersionNumber + "/" + config.CurrentCommit
```
Kubuxu: Current commit might might not be there, are you ok with trailing slash there?

whyrusleeping: Yeah, I think thats fine. Makes it more apparent the code was built without a commit

whyrusleeping:
CHANGELOG:
Set libp2p AgentVersion to our current version and commit.

PR #3571, opened 2017-01-07
rewrite enumerate children async to be less fragile

whyrusleeping:
#3550 made us realize how terribly fragile the EnumerateChildrenAsync code was, so this is a cleaner rewrite of it from scratch.

License: MIT

### DIFF
--- a/merkledag/merkledag.go
+++ b/merkledag/merkledag.go
@@ -140,7 +140,7 @@ func (n *dagService) Remove(nd node.Node) error {
 
 // FetchGraph fetches all nodes that are children of the given node
 func FetchGraph(ctx context.Context, c *cid.Cid, serv DAGService) error {
-	return EnumerateChildrenAsync(ctx, serv, c, cid.NewSet().Visit)
+	return EnumerateChildren(ctx, serv, c, cid.NewSet().Visit, false)
 }
 
 // FindLinks searches this nodes links for the given key,
--- a/merkledag/merkledag.go
+++ b/merkledag/merkledag.go
@@ -389,103 +389,92 @@ func EnumerateChildren(ctx context.Context, ds LinkService, root *cid.Cid, visit
 	return nil
 }
 
+// FetchGraphConcurrency is total number of concurrent fetches that
+// 'fetchNodes' will start at a time
+var FetchGraphConcurrency = 8
+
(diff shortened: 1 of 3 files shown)


merkledag/merkledag.go
```diff
@@ -389,103 +389,87 @@ func EnumerateChildren(ctx context.Context, ds LinkService, root *cid.Cid, visit
...
 			}
-		case <-ctx.Done():
-			return ctx.Err()
-		}
+		}()
 	}
-}
+	defer close(feed)
 
-// FetchGraphConcurrency is total number of concurrenct fetches that
```
Voker57: typo: "concurrenct"

whyrusleeping:
closes #3505

whyrusleeping:
CHANGELOG:
rewrite EnumerateChildrenAsync to be less fragile.

Voker57:
I've tried pinning my large directory with this code and following error happened: http://dump.bitcheese.net/files/lokydaf/daemon.log

whyrusleeping:
@Voker57 Whoops, forgot that these keysets arent threadsafe. Did some locking and ran the merkledag tests with the race detector (would be great to have faster CI so we can run the race detector in CI). Should be good now.

Voker57:
This forkbombed or something my PC (extreme load, unresponsive until I killed ipfs). Unbounded concurrency again?

whyrusleeping:
@Voker57 are you able to run `ipfs refs -r <YOUR HASH>` ?

whyrusleeping:
@Voker57 I just removed concurrency from the pin enumeration. This should help, it might be slower, but depending on the situation, it might actually be faster because of disk contention issues with async...

Kubuxu:
There are merge conflicts.

whyrusleeping:
merge conflicts resolved. Awaiting feedback from @Voker57. My 100GB dataset was able to be pinned just fine, but its on a rather beefy machine.

Voker57:
Serial fetching works, but it's very slow: 1.15 hrs vs ~40 minutes using my patch. it's utilizing neither CPU nor IO to full capacity.

Voker57:
I've submitted PR on this PR (is this the best way to do this?) which brings RAM usage to adequate levels. Also, I've rechecked and parallelizing doesn't win anything significant here, except if disk storage is split into several IO channels somehow.

whyrusleeping:
@Voker57 I'm going to merge this one for now so its shipped in 0.4.5 (despite the decreased performance). We will get your patches into the next release.

PR #3573, opened 2017-01-07
docker: More convenient access to ipfs commands

kpcyrd:
I had to run `ipfs repo fsck` on my repo inside a docker container, what I had to do:
```
docker run --entrypoint ipfs myipfs repo fsck
```
This is because `/usr/local/bin/start_ipfs` is set as an entrypoint and it explicitly starts `ipfs daemon`.

This change keeps start_ipfs as an entrypoint (since it's only doing some docker specific checks), but doesn't restrict you in what you can do.

The above command becomes this:
```
docker run myipfs repo fsck
```

While, starting the ipfs daemon still looks the same:
```
docker run myipfs
```

Bad news (sort of): people already passing arguments to the docker image have to edit their setup from this
```
docker run myipfs --some-flags
```
to this:
```
docker run myipfs daemon --some-flags
```

I still think this is the better approach since overwriting the entrypoint is sort of the non-obvious way for docker beginners (making the ipfs docker image harder to use for beginners).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -74,3 +74,4 @@ VOLUME $IPFS_PATH
 # 1. There's an fs-repo, and initializes one if there isn't.
 # 2. The API and Gateway are accessible from outside the container.
 ENTRYPOINT ["/usr/local/bin/start_ipfs"]
+CMD ["daemon"]
--- a/Dockerfile.fast
+++ b/Dockerfile.fast
@@ -53,3 +53,4 @@ RUN cd $SRC_PATH \
 USER ipfs
 VOLUME $IPFS_PATH
 ENTRYPOINT ["/usr/local/bin/start_ipfs"]
+CMD ["daemon"]
--- a/bin/container_daemon
+++ b/bin/container_daemon
@@ -19,4 +19,4 @@ else
   ipfs config Addresses.Gateway /ip4/0.0.0.0/tcp/8080
 fi
 
-exec ipfs daemon "$@"
+exec ipfs "$@"


lgierth:
Mh, I generally agree this is useful, but I'd love to see it backward-compatible.

kpcyrd:
I'm not sure what's the sanest way for backwards compatibility, personally I would probably go with a breaking change for the sake of future users.

If it really needs to be reverse compatible, I would test for `--` on the first argument and if that's the case, prepend `daemon`. This would break `ipfs --api` and `ipfs --help` (as in, impossible to use without `--entrypoint` workarounds).

VictorBjelkholm:
Yeah, I agree with @kpcyrd, it'll probably be more confusing to be so backward-compatible that it becomes harder to use certain flags.

Makes sense to merge this when we do a minor version bump with other breaking-changes, because this change will in general make a lot of things simpler when interacting with the docker image.

whyrusleeping:
Lets just do this in 0.4.6 and have a bit in the release blog thing about it

whyrusleeping:
Can someone write up something small about the changes in the dockerfile here? @lgierth ?

lgierth:
> I had to run ipfs repo fsck on my repo inside a docker container, what I had to do:
>
> ```
> docker run --entrypoint ipfs myipfs repo fsck
> ```

This command *does work* though, so it's merely a matter of convenience. I'm not convinced that's enough to justify breaking backward compatibility.

(Also, I think it's safe to say that the most common operation with the docker image is starting the daemon.)

lgierth:
I agree the way I went with is unfortunate in retrospect (hardcoding `daemon` in the entrypoint script instead of setting CMD), but that's what we have now :(

> Lets just do this in 0.4.6 and have a bit in the release blog thing about it

That doesn't make it less of a breaking change.

whyrusleeping:
@lgierth do you want to close this then?

kpcyrd:
@lgierth the `ipfs repo fsck` might be a bit abstract, but being able to use `ipfs add`, `ipfs cat` and `ipfs ls` are valid usecases, imo.

The only less breaky way I can think of is changing the script to discard the first argument if it's `daemon` and deprecate the old syntax. That way people can migrate to the new future proof syntax and old setups don't break until it's removed (in favor of allowing all commands).

whyrusleeping:
@kpcyrd @lgierth I merged the other dockerfile PR. Is this one still relevant?

kpcyrd:
@whyrusleeping I'm going to keep my branch and reopen this PR a few months after #3685 has been released :)

Thanks everybody.

whyrusleeping:
@kpcyrd sounds good to me, thanks!

PR #3575, opened 2017-01-08
more configurable datastore configs

whyrusleeping:
Closes #3575


whyrusleeping:
@Kubuxu could I get some early review on this?

Kubuxu:
This branch allows for usage for new experimental datastore (sbs). To use it you have to create new fresh repo.

```sh
ipfs init --empty-repo
ipfs config --json Datastore.Spec "{"mounts":[{"child":{"path":"blocks","type":"sbs"},"mountpoint":"/blocks","prefix":"flatfs.datastore","type":"measure"},{"child":{"compression":"none","path":"datastore","type":"levelds"},"mountpoint":"/","prefix":"leveldb.datastore","type":"measure"}],"type":"mount"}"
```

whyrusleeping:
@Kubuxu nice work so far :)

As an aside, we should look at adding json omitempty tags to the config fields we're deprecating, and start planning a migration path. I think we can pretty easily migrate people from the current format forward into the new format with an fs-repo-migration. At the same time we should adjust for 'new defaults' like the CORS thing that people who init'ed with old nodes might not have.

Perhaps with this migration, we even move the private key out of the config??

whyrusleeping:
Idea: 

Maybe try and ship this code in 0.4.8 (perhaps minus the sbs stuff for now) but keep the default config the same. Then in 0.4.9 we can have a migration to update everyone to the new format

whyrusleeping:
*sigh* this keeps slipping. I *really* want to get it in soon...

kevina:
@whyrusleeping is there anything I can do to help here?

whyrusleeping:
@kevina Yeah, if you want to take this over. Drop the 'sbs' commit from @Kubuxu, and add some tests. That would be wonderful

kevina:
@whyrusleeping @Kubuxu Should I should drop the "--wip--" commit also?

repo/config/datastore.go
```diff
@@ -1,40 +1,43 @@
...
 type S3Datastore struct {
 	Region string `json:"region"`
 	Bucket string `json:"bucket"`
 	ACL    string `json:"acl"`
 }
 
+type FlatDS struct {
+	Path      string
+	ShardFunc string
+	Sync      bool
```
kevina: So we changed this from NoSync to Sync, was this intentional?  Is the default now not to enable Sync?

repo/config/init.go
```diff
@@ -76,19 +76,38 @@ func Init(out io.Writer, nBitsForKeypair int) (*Config, error) {
...
+			"type": "mount",
+			"mounts": []interface{}{
+				map[string]interface{}{
+					"mountpoint": "/blocks",
+					"type":       "measure",
+					"prefix":     "flatfs.datastore",
+					"child": map[string]interface{}{
+						"type":      "flatfs",
+						"path":      "blocks",
+						"nosync":    false,
```
kevina: I might be misunderstanding something, but should't this be `"sync": true` now?

kevina:
I dropped the last two commits, they are saved on `feat/datastore-configs` for now.

repo/config/datastore.go
kevina: And and closer inspection it appears that this data structure isn't even used...

kevina:
A few notes before I am ready to finish this:

1) The Datastore specific config structs do not appear to be used.  Should we make use of them or delete them.  Right now the FlatDS stuct is in accuance as the code is still checking for "nosync" even the the FlatDS uses sync.

2) The config code doesn't seam to check for invalid keys (even outside the Datastore specific code) is this something we want to keep as is.

3) I will add "json omitempty tag" to the fields marked depicted as per @whyrusleeping suggestion.

repo/config/init.go
whyrusleeping: yeah, this should probably be sync true now

whyrusleeping:
@kevina yeah, i think we can remove the datastore specific structs. I wrote those before i figured out how this was gonna work. We can just use the interface map stuff.

Checking for 'disallowed keys' might be nice, though not super simple to do. If you want to, go ahead and add this checking, otherwise leave it unchecked.

your '3' SGTM

repo/config/init.go
kevina: except we check for "nosync" still in the code, one sec, I leave a comment by that code.

repo/fsrepo/datastores.go
```diff
@@ -0,0 +1,126 @@
...
+		p = filepath.Join(r.path, p)
+	}
+
+	sshardFun := params["shardFunc"].(string)
+	shardFun, err := flatfs.ParseShardFunc(sshardFun)
+	if err != nil {
+		return nil, err
+	}
+
+	return flatfs.CreateOrOpen(p, shardFun, params["nosync"].(bool))
```
kevina: Okay so here is where we check for "nosync" except that CreateOrOpen parameter is to sync, no "nosync" so this is a bug as written and I am okay with changing it to "sync".

kevina:
@whyrusleeping I agree with you on allowing either the old or the new config, but what exactly was the old config used for.  What exactly would have happened if `type` was not leveldb, it seams like before the flatfs was hardcoded in.

whyrusleeping:
@kevina the type field was never really used before.

kevina:
Okay if the "Spec" field is non-empty I will use the new config, if it is empty I will use the old settings so that both will work for now.  For completeness I will add back the old settings such as "type" and mark then as depicted and `json omitempty`.

kevina:
@whyrusleeping as it is written now if I user creates a new repo using "ipfs init" wouldn't that technically be the new repo version as it will output an incomparable config?  Should the version number reflect this?

Basically to do what you planned I think we need to support to repo version as given in ".ipfs/version".  Does this sound right?

whyrusleeping:
@kevina hrm... youre right. While it wouldnt need a migration forwards, it would need a migration back. Lets go ahead and bump the version number. Lets get this code ready, and worry about a migration once its ready to go

@Kubuxu @lgierth were there other things we wanted to change in the next migration? Something about default config values?

kevina:
From a (private) IRC conversation started by @whyrusleeping:
> *whyrusleeping*: Hey, on the datastore configs stuff
> *kevina*: yep
> *whyrusleeping*: Once you get that stuff working, it would be great to have a tool that is able to convert an ipfs repo into a given config
> *whyrusleeping*: so you can pass this tool the new config
> *whyrusleeping*: it opens up the repo, and transfers everything over
> *whyrusleeping*: might be a little tricky to do, but would be pretty nice
> *kevina*: this is for the migration tool right?
> *whyrusleeping*: yeah
> *whyrusleeping*: for migration, and then for if people want to try out different datastore configs
> *whyrusleeping*: say for example i want to use leveldb for everything
> *kevina*: you also need to be to able to do the reverse, that might be tricky if a user uses the new features
> *whyrusleeping*: yeah
> *whyrusleeping*: the tool can probably cover the 'easy' cases at first
> *whyrusleeping*: and then we can add in more logic later as demand for things changes
> *kevina*: in some cases it may be impossible'
> *kevina*: I would saw to a best effort and fail cleanly if the repo can't be downgraded
> *whyrusleeping*: Yeah, makes sense
> *kevina*: at least an upgraded repo by the tool should be downgraded
> *whyrusleeping*: yeah
> *kevina*: okay, i post these note on the issue for others to see, I will work on it tonight and tommorow
> *whyrusleeping*: great :)
> *kevina*: I am still unsure how exactly to handle testing though
> *kevina*: but one thing at a time :)
> *whyrusleeping*: yeap! one at a time
> *whyrusleeping*: Since we're going to do a migration anyways
> *whyrusleeping*: having the default config be the new style makes sense to me
> *whyrusleeping*: that way you don't really need to add any tests
> *whyrusleeping*: since all the sharness tests would be running with the new config
> *kevina*: that makes sense
> *kevina*: you still want to support both versions for now (old and new style config)
> *kevina*: ?
> *whyrusleeping*: i think we can probably drop the old style
> *whyrusleeping*: its pretty useless
> *whyrusleeping*: since it doesnt actually do much
> *kevina*: ?? but then users will have to run the migration tool for the next ipfs version, I thought we wanted to avoid that
> *whyrusleeping thinks*
> *kevina*: supporting both versions is not that big of a deal
> *kevina*: we can for now then remove support for the old version when it becomes a problem or after a fixed time
> *whyrusleeping*: ah, but the issue someone (i thought you) brought up was that if the user tweaks their config and it works for that version
> *whyrusleeping*: then downgrade to and older ipfs version
> *whyrusleeping*: with the same repo version
> *whyrusleeping*: it will break
> *whyrusleeping*: so as soon as the user tries out any new config things, the repo version needs to be changed
> *kevina*: a single version of ipfs can support two repo versions for now
> *whyrusleeping*: ah, youre saying we should bump the repo version with this change?
> *kevina*: something like that
> *kevina*: the only question is should "ipfs init" create the old or new style config and repo version
> *whyrusleeping*: right
> *whyrusleeping*: i would say the new one
> *whyrusleeping*: though thinking about it more, i think we should probably do a migration
> *whyrusleeping*: we have other things we wanted to fixup, defaults in the config file
> *kevina*: right you brought that up in the issue

kevina:
@whyrusleeping @Kubuxu @lgierth was there any decision about how default values should be implemented anywhere?  I am thinking they should be stored as metadata for the struct.  Something like
```
type Foo struct {
  someField int `default:23`
}
```
Somewhat like how the json metadata tag is used.

whyrusleeping:
@kevina thats an interesting idea, i kinda like it

kevina:
Are there any config values we want to use this one (for example something to default to true)?

whyrusleeping:
Well, theres the default config in repo/config/init.go, but thats really just default values for *new* repos.

whyrusleeping:
I think that idea would be good for adding in new things, though its unclear if the implementation complexity would be high.

Would love to hear thoughts from @lgierth @Kubuxu @hsanjuan

kevina:
All, @whyrusleeping and I had different things in mind.  For this p.r. I will bump all p.r. version.  My previous comments where trying to do something different from what @whyrusleeping had in mind.  Sorry for the confusion.

whyrusleeping:
@kevina no worries, clearly communicating requirements a skill i'm still getting better at :)

kevina:
@whyrusleeping okay like I said I will bump the repo version and write a tool to migrate the config.

How should we handle the case when we can an old config, should we still support it, or error out.  In order to error out we will need some way to detect if the old config is used since right now we ignore unknown keys?

whyrusleeping:
I think we shouldnt handle the old config. The migration tool should rewrite the old syntax to the new syntax. If the old format is given with the new code, we should error out

kevina:
@whyrusleeping ok I pushed an initial implementation which does what I think you want.  The requested utility is at https://github.com/kevina/ipfs-config-5-6 for now.

kevina:
Also please note that Datastore.Path in the old config was unused and was the absolute path of the datastore.  This information is not stored in the new config so a round trip will lose that field.  In my down conversion I just omit the field.

repo/config/init.go
```diff
@@ -40,7 +37,7 @@ func Init(out io.Writer, nBitsForKeypair int) (*Config, error) {
 			Gateway: "/ip4/127.0.0.1/tcp/8080",
 		},
 
-		Datastore: datastore,
+		Datastore: *datastore,
```
whyrusleeping: seems like needless dereferencing. Probably either make this type a pointer, or make the `DefaultDatastoreConfig` method return a concrete object.

repo/fsrepo/datastores.go
```diff
@@ -0,0 +1,126 @@
...
+			return nil, fmt.Errorf("mounts field wasnt an array")
+		}
+
+		return r.openMountDatastore(mounts)
+	case "flatfs":
+		return r.openFlatfsDatastore(params)
+	case "mem":
+		return ds.NewMapDatastore(), nil
+	case "log":
+		child, err := r.constructDatastore(params["child"].(map[string]interface{}))
```
whyrusleeping: probably want to validate the existence of the child key, and make sure its type asserted properly

whyrusleeping: and the same for some of the other accesses around here

repo/fsrepo/datastores.go
```diff
@@ -0,0 +1,126 @@
...
+		return r.openMeasureDB(prefix, child)
+
+	case "levelds":
+		return r.openLeveldbDatastore(params)
+	default:
+		return nil, fmt.Errorf("unknown datastore type: %s", params["type"])
+	}
+}
+
+type mountConfig struct {
```
whyrusleeping: this is unused now

repo/fsrepo/datastores.go
kevina: Would a single combined check be ok:
```
child, ok := params["child"].(map[string]interface{})
if !ok {
  return nil, fmt.Errorf("child key missing or of the wrong type")
}
```
type case won't throw on nil if I remember correctly.

repo/fsrepo/datastores.go
```diff
@@ -0,0 +1,126 @@
...
+	return mount.New(mounts), nil
+}
+
+func (r *FSRepo) openFlatfsDatastore(params map[string]interface{}) (repo.Datastore, error) {
+	p := params["path"].(string)
+	if !filepath.IsAbs(p) {
+		p = filepath.Join(r.path, p)
+	}
+
+	sshardFun := params["shardFunc"].(string)
```
whyrusleeping: more type checks here

whyrusleeping:
@kevina cool, changes looking good. We will want to add some more type checks, and probably some go tests that assert things get constructed successfully (have json inputs and check types on the results).

The next step after that is to take the tool you wrote, and work it into an ipfs-5-to-6 migration.

repo/fsrepo/datastores.go
whyrusleeping: Oh yeah, that should work.

kevina:
@whyrusleeping I addressed your comments, but have not added any go tests.  Some guidance in that area would be helpful.

whyrusleeping:
@kevina i'm thinking of having a few tests that have a json 'spec' field (what the user would have in their config) and then unmarshal that to a map[string]interface{}, and then calling `constructDatastore` on that. Then check that the object returned from constructDatastore is what we expect, (e.g. a measure datastore wrapping flatfs, or a mount datastore with multiple leveldb instances mounted under it at different points)

kevina:
@whyrusleeping Okay I can do that, tested for nested data structures may be tricky but I will see what I can do

whyrusleeping:
@kevina Yeah, it might be tricky for nested things, but i think with some of them you can inspect their internals a bit (like for the mount datastore).

kevina:
@whyrusleeping the mount datastore does not expose the mounts so inspecting the internals will be problematic.  I can use reflect to get some information but because the mounts are accesses via an unexported filed I can't get call Interface() to get to the mount type.  I will have to keep digging into the internals of the measure and other datastores in order to fully inspect the nested datastores.  Here is what I have so far: https://gist.github.com/kevina/50c1255543cbc8a674a30c7f6e674abf

Please advice.

whyrusleeping:
@kevina hrm... I think for now, i'm fine just having checks that individual datastore types get constructed correctly. Moving forward (can be after this PR) we should add methods to datastores concrete types that allow introspection.

Go ahead and skip the in depth reflection checking for now, I didnt realize so many things werent exported.

kevina:
@whyrusleeping I took a stab at the conversion code see https://github.com/ipfs/fs-repo-migrations/pull/62

whyrusleeping:
I'm thinking that we should put some sort of file in the repo outside of the config that denotes the current repo format. That way, to change the datastore format, one could simply change the config, and restart ipfs. On startup, ipfs would check that the repo format matches the config, and if they differ, prompt the user to run a tool that will convert to the new format.

One approach here would be to take the json data of the datastore config, and canonically encode it to minified json, then base64 encode that. That way users won't mistakenly think they can change the datastore by just changing that file.

thoughts @kevina @Kubuxu @magik6k ?

magik6k:
+1, I've ran into this multiple times when testing badger ds.

kevina:
> One approach here would be to take the json data of the datastore config, and canonically encode it to minified json, then base64 encode that. That way users won't mistakenly think they can change the datastore by just changing that file.

I am against any sort of intentional obfuscation like that.  A simple comment at the top stating what the file is and not to modify it is better.  Yeah I know JSON doesn't support comments, but we could easily strip all lines at the start of the file that start with say '#' before passing into the JSON reader.

lbeziaud:
> Yeah I know JSON doesn't support comments, but we could easily strip all lines at the start of the file that start with say '#' before passing into the JSON reader.

Or you could simply include the comment as JSON. Maybe not as readable but the file would remain standard.
```json
{
    "_comment": "my comment",
    "data": "my data"
}
```

kevina:
> Or you could simply include the comment as JSON. Maybe not as readable but the file would remain standard.

Yeah, but that would not be as obvious.

whyrusleeping:
hrm... okay. I can be okay with using json with a comment. I don't know if i like the idea of adding a comment to the json itself (as @lbeziaud suggests) since that makes it more difficult to compare, and also a bit less obvious to any potential tinkerers reading the file.

kevina:
> I'm thinking that we should put some sort of file in the repo outside of the config that denotes the current repo format. That way, to change the datastore format, one could simply change the config, and restart ipfs. On startup, ipfs would check that the repo format matches the config, and if they differ, prompt the user to run a tool that will convert to the new format.

I think it is a good idea.

We would still need a repo version upgrade.  The upgrade will in addition to changing to the new style config also create the database definition file.

The datastore definition file should probably not be identical to the config, there are some parts that can be changed without an upgrade.

whyrusleeping:
Yeah, definitely still need a migration.

I would argue that for simplicitys sake, the file should be exactly equal
to the config, the tool that does the migrations should be able to detect
cases where we can just no-op

kevina:
> I would argue that for simplicitys sake, the file should be exactly equal
to the config, the tool that does the migrations should be able to detect
cases where we can just no-op

My point is that there are some options that don't effect the datastore layout, you example the `sync` in flatfs.  This option can be changed without requiring a migration.

whyrusleeping:
@kevina i get that. And my point is that discerning what would require a migration and what wouldnt isnt trivial, so we should *always* run the tool. The tool itself can detect these scenarios that don't actually require any work to be done and exit out early

whyrusleeping:
Then, if someone really wants to avoid the tool running, and they know what they are doing, they could make the same modifications in both files

kevina:
@whyrusleeping  that seams like an unnecessary annoyance and its probably not the way I would do it, but I don't want to argue the point.

whyrusleeping:
@kevina there shouldnt be anything annoying about it. You make a change to the config and start ipfs, just like happens now. Users won't even notice the migration tool running (unless there are configuration change issues)

kevina:
@whyrusleeping but you also said

> On startup, ipfs would check that the repo format matches the config, and if they differ, prompt the user to run a tool that will convert to the new format.

That gave me the impression that the user will be prompted when ever there is a change in the database config, hence the annoyance part.

Kubuxu:
There is also other option of allowing the datastores specify deterministic hash/IDs depending on the configuration. Other datastores (mount for example) would be combining those hashes to generate new deterministic hash.

kevina:
> There is also other option of allowing the datastores specify deterministic hash/IDs depending on the configuration. 

I like that idea, it doesn't need to be a hash, just some sort of config id/string would work.  For example the mount ds string can be a simple concatenation of the datastores used, possible with an optional separator.   All that matters is that it is deterministic.  i would have a preference to something somewhat human readable to aid in debugging.

whyrusleeping:
> There is also other option of allowing the datastores specify deterministic hash/IDs depending on the configuration. Other datastores (mount for example) would be combining those hashes to generate new deterministic hash.

This is an interesting idea. I also like that option. It would require modifying the datastore interface (maybe adding a `String()` to the interface?)

In any case, i want to move this forward. This work is our top priority right now (this + badger/other + migration)

kevina:
The problem with a String() or related method is we can't return the String until it is created, which means we don't know if a migration is until after the database is created, which is not ideal.

Instead I propose the following interface:
```
// ConfigFromMap creates a new datastore config from a map
ConfigFromMap(map[string]interface{}) (DatastoreConfig, error)

type DatastoreConfig interface {
  // Id is a unique id representing the Datastore config as stored on disk, runtime config values are not
  // part of this Id.  No length limit.
  Id() string
  // Inst, Instantiate a new datastore from this config
  Inst() (DatastoreConfig, error)
}
```

This has the plus of not requiring a Interface change.  For now this code can all be in go-ipfs.  Later on we can move it into the Datastore.  This can table driven by string the ConfigFromMap in a map.  This could also aid in making it plugable.

@whyrusleeping thoughts

whyrusleeping:
@kevina ooh, yes. Thats a good idea, and i think it actually makes a lot of the things we need it for easier.

kevina:
@whyrusleeping okay, I will implement something tonight.

repo/fsrepo/datastores.go
```diff
@@ -0,0 +1,243 @@
...
+// ConfigFromMap creates a new datastore config from a map
+type ConfigFromMap func(map[string]interface{}) (DatastoreConfig, error)
+
+type DatastoreConfig interface {
+	// DiskId is a unique id representing the Datastore config as stored on disk, runtime config values are not
+	// part of this Id.  No length limit.
+	//DiskId() string
+
+	// Inst, Instantiate a new datastore from this config
+	Inst(path string) (repo.Datastore, error)
```
whyrusleeping: maybe `Create` instead of `Inst` ?

repo/fsrepo/datastores.go
```diff
@@ -0,0 +1,243 @@
...
+	mounts []premount
+}
+
+type premount struct {
+	ds     DatastoreConfig
+	prefix ds.Key
+}
+
+func MountDatastoreConfig(params map[string]interface{}) (DatastoreConfig, error) {
+	res := mountDatastoreConfig{}
```
whyrusleeping: personal preference:
```go
var res mountDatastoreConfig
```

repo/fsrepo/datastores.go
```diff
@@ -0,0 +1,243 @@
...
+	mounts, ok := params["mounts"].([]interface{})
+	if !ok {
+		return nil, fmt.Errorf("'mounts' field is missing or not an array")
+	}
+	for _, iface := range mounts {
+		cfg := iface.(map[string]interface{})
```
whyrusleeping: type check here, this is coming from something the user types out.

repo/fsrepo/datastores.go
```diff
@@ -0,0 +1,243 @@
...
+	cfg map[string]interface{}
+}
+
+func FlatfsDatastoreConfig(params map[string]interface{}) (DatastoreConfig, error) {
+	return flatfsDatastoreConfig{params}, nil
+}
+
+func (c flatfsDatastoreConfig) Inst(path string) (repo.Datastore, error) {
+	params := c.cfg
+	p, ok := params["path"].(string)
```
whyrusleeping: maybe all this extraction of info from the generic `map[string]interface{}` should go in the config constructor above? would make the constructor here much cleaner

repo/fsrepo/datastores.go
kevina: Sure.

repo/fsrepo/datastores.go
kevina: okay

repo/fsrepo/datastores.go
kevina: You are probably right.  I was mainly interested in getting something out quick.

kevina:
FYI: @whyrusleeping @Kubuxu in order to get each commit to compile I needed to squash your commits together.

Kubuxu:
sure

whyrusleeping:
works for me

kevina:
Note: I plan to write up some notes on how I see the migration tool working after I get some sleep.

repo/fsrepo/datastores.go
```diff
@@ -136,6 +147,10 @@ func FlatfsDatastoreConfig(params map[string]interface{}) (DatastoreConfig, erro
 	return &c, nil
 }
 
+func (c *flatfsDatastoreConfig) DiskId() string {
+	return fmt.Sprintf("flatfs;%s;%s", c.path, c.shardFun.String())
```
Kubuxu: I am not sure if including path is right thing to do.
Usually when path is changed the user will migrate data himself.

kevina:
Okay,

Yesterday I pused some code that computes the DiskID and writes the file to disk.  If the file does not exist it will simply be created. If it exists the contents will be compared to the computed contents and error out if it differs.  This way if a user really knows what they are doing they can just delete the file and it will be regenerated.  If the configuration is incompatible the relevent datastore will likely return an error.

There are many open question, so in order to proceed further I need a better idea of how we envision the migrataion tool working.  What
follows is a brain dump.  Sorry for the slightly disorganized nature
and the delay in getting this note out.

**A**

There are several different things that can change with the
configuration and it is unclear if migration tool should handle them
all and how it will detect them.  These are

1. Database used changed
2. Database paramaters change that require a migration, for example a prefix length change in flatfs
3. File path changed, this only included changes of the "path" field in the spec, not changing the location of the entire repo
4. Mount prefix changed

I am assuming the migration tool was primary meant for (1) and maybe (2). Detecting any of these changes require some of the paramaters stay the same.  In particular the mount point and file path sort of act as an identifier.

**B**

Should we also write out the old config file.  For same changes it is not needed (such as (1) and (2) assuming (3) is the same) but for others such as (3) or (4) changing it is required.  I would go as far as saying that the migraation tool should be written to be as robust as possible, and even if the old config file is present it should not rely on it. (Note is is also possiblt to get some information from the ID itself, but that is not its indented purpose).

repo/fsrepo/datastores.go
kevina: The path helps to identify the datastore.  Especially if we ever allow multiple datastores on the same mountpoint (like a unionfs).  Also see my comment I just left.

whyrusleeping:
@kevina the migration tool (assuming youre talking about https://github.com/ipfs/ipfs-ds-convert) converts any datastore to any other datastore, it just runs generically. All of the situations you listed should cause the tool to be invoked. Some of these cases have very simple ways that they can be handled (and don't require the fallback of 'read every value from old, and write it to new'), we can add in checks for those in the ipfs-ds-convert tool as optimizations.

kevina:
@whyrusleeping thanks for the pointer, it seams that the tool does require the old spec file so I am now saving it along with the disk id

cc @magik6k

magik6k:
That is right, currently it's loading the config and using it to open 'old' datastore, and a new spec file to create the 'new' datastore. I'll change that to match what/how it's done here.

Can you rebase on current master? I need that for test init profile.

kevina:
@magik6k note that I don't like to rebase continuously (I just did it a few days ago).  There are no conflicts, you can just merge this with master for your own purposes.  cc @whyrusleeping

kevina:
@whyrusleeping @magik6k to be clear:  The current design is that the spec file will be created when the repo is first opened.  This way there is no need for the conversions tools to duplicate code and create this file.  Also, if a user really knows what they are doing and manually did a conversion they can just delete the file and it will be regenerated. If the configuration is incompatible the relevant datastore will likely return an error.

whyrusleeping:
@kevina I think that makes sense, we still get the 'if the user modifies the config, a conversion gets run' safety.

thoughts @Kubuxu?

Kubuxu:
Having the conversion tool write the file gives us one more thing, if the user changes config, runs conversion, changes config again it won't become a problem.

repo/fsrepo/config_test.go
```diff
@@ -0,0 +1,217 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	dsc, err := AnyDatastoreConfig(config.Spec)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	expected := "/blocks:{flatfs;blocks;/repo/flatfs/shard/v1/next-to-last/2};/:{levelds;datastore};"
```
whyrusleeping: I would prefer if this looked like: `mount{.....}`. Just given this string i don't really know what i'm looking at.

kevina: Okay, how about; "mount;/blocks:{flatfs;blocks;/repo/flatfs/shard/v1/next-to-last/2};/:{levelds;datastore}; to be consistent with how the other datastores are coded.

Note, also this is only suppose to capture the on disk format, datastores such as the log and measure datastore don't change the ondisk format so they are not encoded.

kevina:
@Kubuxu 

> Having the conversion tool write the file gives us one more thing, if the user changes config, runs conversion, changes config again it won't become a problem.

That is true.  I just hate to duplicate code, but I suppose we can once this P.R. is merged if the issue is important enough.

I still think that if the file doesn't exist it should just be created.  This will allow a user to specify an alternative configuration on a fresh repo without having to involve the tool.  The idea was  (although I admit I did not double check that it will work) is that a user runs "ipfs init" modifies the config, then starts ipfs, and the datastore and spec file are created with the user modified configuration.

repo/fsrepo/config_test.go
whyrusleeping: yeap, understood about the log and measure bit (would apply to adding caches, and other fun things).

That syntax still feels weird. Since we're going for something the user should be able to read, i would prefer if it was just a little bit more json-ish and used a `[]` for the mounts list.

cc @Kubuxu for thoughts on the syntax

repo/fsrepo/fsrepo.go
```diff
@@ -379,6 +398,61 @@ func (r *FSRepo) openDatastore() error {
 	return nil
 }
 
+var SpecFn = "spec"
```
whyrusleeping: maybe call this `repo_spec`, or `repo_id`. Spec is a very unassuming name that makes me more curious than i should be, calling something `id` generally makes me think I shouldnt touch it

magik6k:
@kevina I don't think that changing datastore config after `ipfs init` would work as `fsrepo.Init` is called by init and at that point there are some keys in the datastore (You can check by calling init with logDatastore+mdns set to false(mdns mutes std logger)).

There are only about 30 lines of code in ds-convert that are responsible for writing spec file, rest is just copy-pasted from here, without any changes (https://github.com/ipfs/ipfs-ds-convert/blob/master/convert/open.go#L44, This bit of code can be easily moved into separate repo, though that's out of scope for this PR and can be done later). I think that we should keep it there for reasons pointed out by @Kubuxu above.

repo/fsrepo/config_test.go
kevina: @whyrusleeping how about
`mount[/blocks:flatfs{blocks:/repo/flatfs/shard/v1/next-to-last/2},/:levelds{datastore},]`

I am just going for something unique and somewhat human readable and easy to generate (hence the extra comma at the end of the list).

kevina:
@magik6k your right about changing the datastore after `ipfs init`.  Although out of the scope of this PR there should probably be a way to initialize a repo with an alternative datastore configuration.

repo/fsrepo/fsrepo.go
kevina: How about `datastore_id`?  The id is not really for the repo but the datastore.  It is a bit more than an id as it also includes the spec but I suppose I can live with that.

magik6k:
You can just pass preconfigured config to ipfs init as an argument

kevina:
@magik6k oh okay, I didn't know that and should of checked

repo/fsrepo/fsrepo.go
whyrusleeping: sure, `datastore_id` works for me. (i think 'id' is the appropriate term as we are using it as a unique identifier for the on-disk datastore configuration)

kevina:
I would like to move forward on this.  There are two outstanding issues:

(1) The format of the `DiskId`.  I don't care what the forward is. However, it needs to properly supported nesting, so there needs to be some sort of grouping character.  The original format string of:
```
/blocks:{flatfs;blocks;/repo/flatfs/shard/v1/next-to-last/2};/:{levelds;datastore};
```
supported nesting by surrendering any nested datastore id's with a '{}'.  If something like this is not used there need to a requirement that it is used in the datastore id if to surrendering any problematic characters (like say `:` `,` `;`) with a set of grouping or quoting characters.  I am not 100% sure but I don't think escaping should be necessary as the id only needs to be unique, not parsable.  The new proposed format does this
```
mount[/blocks:flatfs{blocks:/repo/flatfs/shard/v1/next-to-last/2},/:levelds{datastore},]
```
I like my original format better, but I don't want to waste a lot of time bikesheding on this.

(2) How to handle the creation of the `datastore_id` file.  There are several options

1. Only create it during init and datastore conversions.  It is the users own problem to figure out how to regenerate it they make manual changes to the datastore without using a conversion tool.
2. Create it if it doesn't exist when the datastore is opened and also during datastore conversions.  If a user makes manual changes to the datastrore they can just remove this file and a new one will be recreated.  If the new config doesn't match what is on disk this will likely be caught by the individual datatsores.
3. Create it if it doesn't exist when the datastore is opened.  Conversion code does not create this file.  This is the simplest and avoids any code duplication but does have some problems as earlier indicated by @Kubuxu.
4. A variant of (1) but provide a command to create the `datastore_id` file.  This command can also be used by conversion tools to avoid code duplication.

@whyrusleeping @magik6k @Kubuxu, others thoughts.

whyrusleeping:
we can do option 1, and the conversion tool can just import the same code to generate the `datastore_id`

As for the format, if we're going through the effort of making it human readable, we might as well make it parseable. The argument for making it parseable is so that the conversion tool can do interesting things like diffs (for example, if we only change the /blocks mount, we don't need to deal with moving keys in the `/` namespace).

Stebalien:
> i would have a preference to something somewhat human readable to aid in debugging.

The problem with making things human readable is that humans *will* read them (and mess with things they shouldn't mess with).

What about just base64 (base58?) encoding a canonized JSON (CBOR?) representation of the data? That way *we* can read it, it's readily parsable, and users won't mess with it. We can log it as `disk:XXXXX...` to make it easy to identify.

If we really feel like dogfooding, we could even store it as a CID using the identity hash function (so we don't need a blockstore): `disk:<multibase>-<cidv1>-<cbor>-<id-hash>-<length>-<config>`. However, this is probably more work than it's worth (I'm only suggesting it because dogfooding IPLD is a good way to improve it).

kevina:
I would go with (2) then its just as easy and more flexible.

As far the format, this is just an ID.  To do interesting things with it we can read the Json spec file also in the same file.

On July 11, 2017 6:02:09 PM EDT, Jeromy Johnson <notifications@github.com> wrote:
>we can do option 1, and the conversion tool can just import the same
>code to generate the `datastore_id`
>
>As for the format, if we're going through the effort of making it human
>readable, we might as well make it parseable. The argument for making
>it parseable is so that the conversion tool can do interesting things
>like diffs (for example, if we only change the /blocks mount, we don't
>need to deal with moving keys in the `/` namespace).
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/ipfs/go-ipfs/pull/3575#issuecomment-314585107

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

magik6k:
I'd go with (1) and throw an error if that file doesn't exist. This seems to be the most conservative option and it's probably the way we want to handle data.

example: someone changes spec in config, and removes the spec file thinking it's how it should be done, in worst case whole repo may go corrupt. In case (1) user gets an error and can recover previous config/spec which in most cases like that will be defaults.

Code duplication is not as bad as it seems as the conversion tools needs that same code anyway to open the datastores to do it's work.

kevina:
@magik6k the problem is if the user removes the `datastore_id` file, there needs to be a way for it to be regenerated, hence my suggestion in (4).

kevina:
@Stebalien the `datastore_id` contains both the computed id and the configuration spec.  When opened my code checks that the computed id matches the spec in the configuration file so any tampering will be evident.

Stebalien:
> I would go with (2) then its just as easy and more flexible.

Will 2 always be possible? If so, why bother crating this file in the first place? That is, if we can just recreate this file through introspection, we don't really need it. If there *are* ambiguous cases (cases where we *need* this file), we shouldn't guess.

> As far the format, this is just an ID. To do interesting things with it we can read the Json spec file also in the same file.

I see. For context, `database_id` will include:

```js
{
    "id": "<some identifier>",
     "spec": "<some spec>"
}
```

So, you want ID to be human readable for debuggability and want there to be a unique mapping from ID -> spec? It seems kind of redundant to store both at that point (they should contain equivalent information).

Stebalien:
> the problem is if the user removes the datastore_id file, there needs to be a way for it to be regenerated, hence my suggestion in (4).

That sounds like a good compromise. However, we can cross that bridge when we get there. For now, 1 and 4 are equivalent.

Stebalien:
>  the datastore_id contains both the computed id and the configuration spec. When opened my code checks that the computed id matches the spec in the configuration file so any tampering will be evident.

IMO, tamper resistance is better achieved through either putting a large comment at the top telling the user to not touch the file OR encoding (e.g., base64). Using the ID as a checksum is a bit funky.

kevina:
> So, you want ID to be human readable for debuggability and want there to be a unique mapping from ID -> spec? It seems kind of redundant to store both at that point (they should contain equivalent information).

The "spec" is just a copy of the "spec" config entry it contains runtime components that can be changed without having to convert the datastore.  The "id" is a sting which represents the on disk components of the config.

repo/fsrepo/datastores.go
```diff
@@ -0,0 +1,287 @@
...
+
+	switch params["compression"].(string) {
+	case "none":
+		c.compression = ldbopts.NoCompression
+	case "snappy":
+		c.compression = ldbopts.SnappyCompression
+	case "":
+		fallthrough
+	default:
+		c.compression = ldbopts.DefaultCompression
```
Stebalien: This looks like it ignores garbage values (e.g., "snapy" will silently ignored). if so, this should probably return an error.

repo/fsrepo/fsrepo.go
```diff
@@ -379,6 +398,61 @@ func (r *FSRepo) openDatastore() error {
...
+	fn, err := config.Path(r.path, SpecFn)
+	if err != nil {
+		return err
+	}
+	idspec := map[string]interface{}{
+		"id":   id,
+		"spec": spec,
+	}
+	b, err := json.Marshal(idspec)
+	err = ioutil.WriteFile(fn, b, 0666)
```
Stebalien: While the database directory shouldn't be world readable so this may not make a difference, this file really shouldn't be world writable.

Stebalien:
> The "spec" is just a copy of the "spec" config entry it contains runtime components that can be changed without having to convert the datastore. The "id" is a sting which represents the on disk components of the config.

That seems a bit unnecessary... also, it's guaranteed to get out-of-sync. Actually, what *is* the point of storing the spec here? It seems like we parse it, verify that it matches the ID, and then throw it away.

Sorry for harping on this, it's just that...

(1) Duplicate states often leads to bugs when the states gets out-of-sync.

Right now, it doesn't look like we rely on this spec but, if the data are there, someone will parse it (`readSpec` even makes this easy by returning it!). I just know that some unsuspecting new (or old) contributor will make this mistake, it's only a matter of time.

(2) One should only invent a custom serialization format with good reason.

In this case, the lack of a canonical serializer/deserializer makes it hard to ensure a one-to-one mapping between spec and ID (it's easy to accidentally produce colliding `DiskId`s by, e.g., using special characters in directory names).

Furthermore, if we used a deterministic, standardized encoding format like base64 encoded cbor, datastore configs could generate their `DiskId`s by copying themselves (the datastructure itself) and removing unimportant fields. This would ensure that datastore configs never forget to include important fields in their IDs.

repo/fsrepo/datastores.go
kevina: I agree, but didn't write that bit of code so I left it as is.  @whyrusleeping agree with @Stebalien?

whyrusleeping: Yeah, agreed. That makes sense

kevina:
This is becoming more involved than I would of liked.

My original idea was to write out a minimal config file.  @whyrusleeping didn't like that idea because determining what should be in that config file was non-trivial.  @Kubuxu suggested some sort of id or hash.  This turned out not to be possible with out the `DatastoreConfig` abstraction.  However, now that we have that abstraction I propose we use to return the minimal config, instead of an id.  The minimal config we be of the type `map[string]interface{}` and only include non-runtime fields.

The file written out will then be `datastore_spec` and include the minimized json formatting of the `map[string]interface{}`.  Optionally it can include a comment at the top saying what it is and not to directly modify it.  I am against encoding this using base64 or something similar.

The `datastore_spec` file will be written by the conversion tools and when the repo is first initialized.  If it is not there an error will be thrown.  We should still provide a way to write this out for advanced uses but that can be another P.R.

@whyrusleeping @magik6k @Kubuxu @Stebalien how does this sound?

Stebalien:
Sounds like a great plan! I was concerned that the JSON serializer wouldn't serialize in a canonical order but I appear to have been mistaken.

(and thanks for bearing with me on this)

whyrusleeping:
@kevina that sounds good to me.

kevina:
@Stebalien thanks. @whyrusleeping I try to have something ready by the end of Wed.

kevina:
My internet (or more specifically my lan and wi-fi connection to the fios modem/router) cut out today.  I finally managed to get it working again, for now, but least to say I wasn't able to get much work done.  I will try to have this done Thursday.

kevina:
Okay this is done now.  Sorry for the delay.

repo/fsrepo/datastores.go
Stebalien: @kevina, while we're at it, it would be nice to fix this.

repo/fsrepo/datastores.go
```diff
@@ -0,0 +1,316 @@
...
+	mounts := make([]interface{}, len(c.mounts))
+	for i, m := range c.mounts {
+		c := m.ds.DiskSpec()
+		if c == nil {
+			c = make(map[string]interface{})
+		}
+		c["mountpoint"] = m.prefix
+		mounts[i] = c
+	}
+	cfg["mounts"] = mounts
```
magik6k: This serializes the mount prefix as `"mountpoint":{"string":"/blocks"}`. Should just be the string.

repo/fsrepo/datastores.go
kevina: Oversight, fixed now.

repo/fsrepo/fsrepo.go
```diff
@@ -379,6 +415,20 @@ func (r *FSRepo) openDatastore() error {
 	return nil
 }
 
+var SpecFn = "datastore_spec"
```
magik6k: Should be const at the top, like in repo/config/config.go

repo/fsrepo/datastores.go
kevina: oops, fixed

kevina:
I plan to rebase this to clean up my commits once I get the all clear from @whyrusleeping I will then work on fixing on the migration tool.

whyrusleeping:
@kevina Looked this over, it looks good to me. The next thing you need from me here is a release of this PR with the correct version string, right?

repo/fsrepo/datastores.go
```diff
@@ -0,0 +1,330 @@
...
+			prefix: ds.NewKey(prefix.(string)),
+		})
+	}
+
+	return &res, nil
+}
+
+func (c *mountDatastoreConfig) DiskSpec() DiskSpec {
+	cfg := map[string]interface{}{"type": "mount"}
+	mounts := make([]interface{}, len(c.mounts))
```
magik6k: Noticed one more small thing: mounts in spec should be sorted as the order doesn't matter, see https://github.com/ipfs/go-datastore/blob/master/syncmount/mount.go#L31

kevina: Fixed.

kevina:
@whyrusleeping yep, I need a release with the correct version string so I can fix up fs-repo-migrations tests.  I also need to write out the spec file there.

Note, just rebased.

repo/fsrepo/config_test.go
```diff
@@ -0,0 +1,219 @@
...
+	if dsc.DiskSpec().String() != expected {
+		t.Errorf("expected '%s' got '%s' as DiskId", expected, dsc.DiskSpec().String())
+	}
+
+	ds, err := dsc.Create(dir)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if typ := reflect.TypeOf(ds).String(); typ != "*flatfs.Datastore" {
```
whyrusleeping: I love this test.

repo/fsrepo/datastores.go
whyrusleeping: did this get fixed? I'm not seeing where the sorting happens.

whyrusleeping: oh wait, nvm. It happens just a few lines up.

test/sharness/t0066-migration.sh
```diff
@@ -34,7 +34,7 @@ test_expect_success "ipfs daemon --migrate=true runs migration" '
 	test_expect_code 1 ipfs daemon --migrate=true > true_out
 '
 
-test_expect_success "output looks good" '
+test_expect_failure "output looks good" '
 	grep "Running: " true_out > /dev/null &&
```
kevina: Note, this should be fixed somehow, it seams we have a bit of a chicken-and-egg problem so maybe it would be easier to fix in a separate p.r.

kevina:
@whyrusleeping this should be ready to go...

whyrusleeping:
nitpick while playing around with it, can we do "spec" instead of "Spec"?

kevina:
> nitpick while playing around with it, can we do "spec" instead of "Spec"?

Not easily.  Unless I am missing something, `Spec` comes from the serialization of a field in the `Datastore` struct.

whyrusleeping:
@kevina yeah, we can just change the json field name. It should be a pretty easy change (and we will need to make the same change in the migrations code). It just feels a little weird that everything inside the spec is lowercased, and the rest is capitalized.

kevina:
@whyrusleeping oops, forgot about that will push a fix shortly and update the conversion code

kevina:
@whyrusleeping I am not sure I agree that "Spec" should be lowercase, the config now looks like this:
```
{
    "StorageMax": "10GB",
    "StorageGCWatermark": 90,
    "GCPeriod": "1h",
    "spec": {
      "mounts": [
        {
          "child": {
            "compression": "none",
            "path": "datastore",
            "type": "levelds"
          },
          "mountpoint": "/",
          "prefix": "leveldb.datastore",
          "type": "measure"
        },
        {
          "child": {
            "path": "blocks",
            "shardFunc": "/repo/flatfs/shard/v1/next-to-last/2",
            "sync": true,
            "type": "flatfs"
          },
          "mountpoint": "/blocks",
          "prefix": "flatfs.datastore",
          "type": "measure"
        }
      ],
      "type": "mount"
    },
    "HashOnRead": false,
    "BloomFilterSize": 0
}
```
If you think that is better I will go ahead and push the change and fix the conversion code.

kevina:
I don't think it is better because all the other fields under the Datastore config are title case with just "spec" being lower case.

whyrusleeping:
@kevina hrm.... yeah... youre right. We definitely should (at some point in the future) make the config look nicer.

kevina:
So leave things as they are for now?

whyrusleeping:
@kevina yeah.

This PR is RFM, would be nice to get an approval from @Kubuxu and @Stebalien, but once we have a release ready for the fs-repo-migrations (pending some config changes @lgierth wanted to add) we can ship this finally :)

PR #3581, opened 2017-01-10
"ipfs key list": include the hash of the key id in addition to the name

### LINKED ISSUE #3568: No ( apparent ) way to list the hash of existing keys
#### Version information:
```
go-ipfs version: 0.4.5-dev-d83a911
Repo version: 4
System version: amd64/linux
Golang version: go1.7.4
```

#### Type:
`Bug/Missing Feature`

#### Priority:
`P1`

#### Description:
I can get my `self` IPNS id via the command `ipfs id`. But there seems to be no way to list the hash for some secondary key aside from publishing something to the keyslot with `ipfs name publish -k <key>`

Before this is released as stable `ipfs key list` should gain a way to list the corresponding hashes ( either by default or with an option )

kevina:
First stab, closes #3568

### DIFF
--- a/core/commands/keystore.go
+++ b/core/commands/keystore.go
@@ -125,7 +125,7 @@ var KeyGenCmd = &cmds.Command{
 				return nil, fmt.Errorf("expected a KeyOutput as command result")
 			}
 
-			return strings.NewReader(k.Id), nil
+			return strings.NewReader(k.Id + "\n"), nil
 		},
 	},
 	Type: KeyOutput{},
--- a/core/commands/keystore.go
+++ b/core/commands/keystore.go
@@ -143,7 +143,7 @@ var KeyListCmd = &cmds.Command{
 		Tagline: "List all local keypairs",
 	},
 	Options: []cmds.Option{
-		cmds.BoolOption("show-ids", "l", "also show key ids"),
+		cmds.BoolOption("l", "Show extra information about keys."),
 	},
 	Run: func(req cmds.Request, res cmds.Response) {
 		n, err := req.InvocContext().GetNode()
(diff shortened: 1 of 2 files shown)


kevina:
@whyrusleeping not sure about this at all, feedback welcome, fell free to just take over this pr if it will be faster.

core/commands/keystore.go
```diff
@@ -149,7 +149,28 @@ var KeyListCmd = &cmds.Command{
...
+
+			pubKey := privKey.GetPublic()
+
+			pid, err := peer.IDFromPublicKey(pubKey)
+			if err != nil {
+				res.SetError(err, cmds.ErrNormal)
+				return
+			}
+
+			list = append(list, key+" "+pid.Pretty())
```
whyrusleeping: Using an array of strings here will make the json output of the api hard to parse. Lets make the output be an array of `KeyOutput` structs (which probably means defining a `KeyOutputList` struct because our use of the type system in the commands lib is weird).

Lets add a `-l` or `--show-ids` flag to the command and then in the output marshaler, then check that for whether or not to print the keys alongside the names.
Then in the text output marshaler, use a tabwriter to make sure the columns are aligned evenly when outputting both names and keys.

whyrusleeping:
CHANGELOG:
Add key ID output option to 'ipfs key list' command.

kevina:
@whyrusleeping do you think it would be better for the hash to come first.  It would make it easier to parse by shell utilities if the key name has spaces in it and would also be more consistent with "ipfs ls".  I don't have a strong preference either way.

kevina:
Done.  I listed the hash first when the "-l" option is used, but can easily swap them.

core/commands/keystore.go
```diff
@@ -135,6 +142,9 @@ var KeyListCmd = &cmds.Command{
 	Helptext: cmds.HelpText{
 		Tagline: "List all local keypairs",
 	},
+	Options: []cmds.Option{
+		cmds.BoolOption("show-ids", "l", "also show key ids"),
```
Kubuxu: I would say just `l`, an have it catch all long format output, I think we might have more in future.

kevina:
@whyrusleeping your changes look good to me.  Merge when ready.  Let me know if there is anything else you need from me on this.

PR #3585, opened 2017-01-10
corehttp: cleanup close handling

Kubuxu:
License: MIT


lgierth:
Yes this is better, LGTM :+1:

whyrusleeping:
better, yet seems to break everything...

Kubuxu:
Yup.

PR #3587, opened 2017-01-11
`make source` for installation in unsupported systems

### LINKED ISSUE #3579: Add `make source` for options on building on less common systems
#### Description:

> Note: This process may break if gx or any of its dependencies break as go get will always select the latest code for every dependency, often resulting in mismatched APIs.

To peg the version, it should be sufficient with a redirect gopkg.in/whyrusleeping/gx-go.v1. This is an additional POF, but still better than the current method for "building on less common systems", which is without the version peg.

rht:
Could be further streamlined if the os/dist is auto-detected.
Fixes #3579

License: MIT
Signed-off-by: rht <rhtbot@gmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -149,6 +149,15 @@ windows_build_check:
 	GOOS=windows GOARCH=amd64 go build -o .test.ipfs.exe ./cmd/ipfs
 	rm -f .test.ipfs.exe
 
+source:
+	@echo "note: this command has yet to be tested to build in the system you are using"
+	@echo "installing gx"
+	go get -u github.com/whyrusleeping/gx
+	go get -u github.com/whyrusleeping/gx-go
+	gx --verbose install --global >/dev/null 2>&1
+	@echo "installing go-ipfs"
+	go install ./cmd/ipfs
+
 PHONY += test test_short test_expensive
 
 ##############################################################
--- a/Makefile
+++ b/Makefile
@@ -149,7 +149,7 @@ windows_build_check:
 	GOOS=windows GOARCH=amd64 go build -o .test.ipfs.exe ./cmd/ipfs
 	rm -f .test.ipfs.exe
 
-source:
+install_unsupported:
 	@echo "note: this command has yet to be tested to build in the system you are using"
 	@echo "installing gx"
 	go get -u github.com/whyrusleeping/gx
--- a/README.md
+++ b/README.md
@@ -82,21 +82,9 @@ export PATH=$PATH:$GOPATH/bin
 
 #### Download and Compile IPFS
 
-go-ipfs differs from the vanilla `go get` flow: it uses
-[gx](https://github.com/whyrusleeping/gx)/[gx-go](https://github.com/whyrusleeping/gx-go)
-for dependency management.
-
-First download `go-ipfs` without installing:
-
 ```
 $ go get -u -d github.com/ipfs/go-ipfs
-
 $ cd $GOPATH/src/github.com/ipfs/go-ipfs
-```
-
-Then install `go-ipfs` and its dependencies, including `gx` and `gx-go`:
-
-```
 $ make install
 ```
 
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -149,6 +149,14 @@ windows_build_check:
 	GOOS=windows GOARCH=amd64 go build -o .test.ipfs.exe ./cmd/ipfs
 	rm -f .test.ipfs.exe
 
+source:
+	@echo "installing gx"
+	go get -u gopkg.in/whyrusleeping/gx.v0
```
whyrusleeping: I don't entirely trust gopkg. Lets just use the standard github path here

README.md
```diff
@@ -105,13 +105,9 @@ $ make install
...
 building ipfs anyways (it should work fine in most cases), you can do the
 following:
-
-- Install gx: `go get -u github.com/whyrusleeping/gx`
-- Install gx-go: `go get -u github.com/whyrusleeping/gx-go`
-- Fetch ipfs source: `go get -d github.com/ipfs/go-ipfs 2> /dev/null`
-- Enter source directory: `cd $GOPATH/src/github.com/ipfs/go-ipfs`
-- Install deps: `gx install`
-- Install ipfs: `go install ./cmd/ipfs`
+```
```
whyrusleeping: Make sure to mention that this needs to be done from within the correct location inside `$GOPATH`. The reason i had the explicit step of changing directory into ipfs/go-ipfs was because a lot of people missed that part.

rht: I saw the `$GOPATH` lines, realized they were redundant with the `go get` and `cd` from the previous section.

whyrusleeping: hrm...  I guess thats fine. I was expecting people to not read the previous section, but it makes sense that they would start with the previous one before noticing that it fails.

Kubuxu:
I am not 100% sure if it is the right thing to do, it might not work on BSD systems (gmake vs bmake, and rest of the makefile), and it makes it seem like they are fullly supported. It ingenerate into users binaries (`$GOPATH/bin`) by downloading and installing `gx` and `gx-go` (user might have custom versions installed).

rht:
Supported systems (for the most recent commit, and/or releases), I assume are the systems that have been verified to build in the CI.

> It ingenerate into users binaries ($GOPATH/bin) by downloading and installing gx and gx-go (user might have custom versions installed).

The `-u` flag will override the existing `gx` installation with custom versions.

rht:
> I am not 100% sure if it is the right thing to do, it might not work on BSD systems (gmake vs bmake, and rest of the makefile), and it makes it seem like they are fullly supported.

I wonder if a message `note: this command hasn't been tested to build in the system you are using` would suffice?

README.md
```diff
@@ -82,38 +82,23 @@ export PATH=$PATH:$GOPATH/bin
 
 #### Download and Compile IPFS
 
-go-ipfs differs from the vanilla `go get` flow: it uses
-[gx](https://github.com/whyrusleeping/gx)/[gx-go](https://github.com/whyrusleeping/gx-go)
```
rht: Reduces cognitive overload. The entire install process doesn't depend on whether the user/dev knows `gx` is used under the hood or not.

whyrusleeping: Very fair

Kubuxu:
SGTM generally, I would bo also change the name, `make source` might in future be used to build source package, I would go with `make install_from_source`, or something like that.

@whyrusleeping what do you think.

Kubuxu:
What I mean by that is in future we will probably have command that collects sources to the vendor directory for offline, movable builds and for source distributions.

Currently if I've seen this target I would call it to prep the source code for build, not expecting it to prep the source, build and install go-ipfs.

rht:
That would be `make vendor`.

Installation in an unsupported system requires that everything is compiled and fetched from scratch (even if the system has some of the required libs), often with manual patches.

rht:
> ... to the vendor directory for offline, movable builds and for source distributions.

This is one of the main reason why `go get` still doesn't have the version pinning. The deps are vendored by the maintainer, and it is the responsibility of the maintainer to make sure they tie in seamlessly.

rht:
`s/source/install_unsupported/` -- surely there is no naming collision by now, unless it is going to be reserved for installation for unsupported experimental flags.

Kubuxu:
I will rebuild this target to work with new makefiles. Closing for now.

PR #3589, opened 2017-01-11
update to fixed cbor ipld lib

whyrusleeping:
This fixes cbor parsing to follow the ipld spec in regards to storing links as a special cbor tag instead of mapping the `{"/":"QmFooBar"}` syntax into cbor. We've selected 258 as the ipld link tag value for now since it doesnt appear to actually be defined anywhere in the spec.

cc @diasdavid @jbenet @nicola

License: MIT


Kubuxu:
Tests are failing in t0053-dag.sh.

whyrusleeping:
fixed this in #3616

Issue #3588, opened 2017-01-11
Fix concurrency performance issues of EnumerateChildrenAsync


whyrusleeping:
We changed the pin call to use the serial version of EnumerateChildren as users were reporting resource consumption issues with it. We need to investigate these issues, add a test for them, and then hopefully move pin back to an optimized fetch.

whyrusleeping:
I believe this is now done, thanks @Voker57 :)

PR #3598, opened 2017-01-14
Pass cids instead of nodes around in EnumerateChildrenAsync

Voker57:
Saves a whole lot of RAM on todobuffer

### DIFF
--- a/merkledag/merkledag.go
+++ b/merkledag/merkledag.go
@@ -140,7 +140,7 @@ func (n *dagService) Remove(nd node.Node) error {
 
 // FetchGraph fetches all nodes that are children of the given node
 func FetchGraph(ctx context.Context, c *cid.Cid, serv DAGService) error {
-	return EnumerateChildren(ctx, serv, c, cid.NewSet().Visit, false)
+	return EnumerateChildrenAsync(ctx, serv, c, cid.NewSet().Visit)
 }
 
 // FindLinks searches this nodes links for the given key,
--- a/merkledag/merkledag.go
+++ b/merkledag/merkledag.go
@@ -394,45 +394,40 @@ func EnumerateChildren(ctx context.Context, ds LinkService, root *cid.Cid, visit
 var FetchGraphConcurrency = 8
 
 func EnumerateChildrenAsync(ctx context.Context, ds DAGService, c *cid.Cid, visit func(*cid.Cid) bool) error {
-	if !visit(c) {
-		return nil
-	}
-
-	root, err := ds.Get(ctx, c)
-	if err != nil {
-		return err
-	}
-
-	feed := make(chan node.Node)
-	out := make(chan *NodeOption)
+	feed := make(chan *cid.Cid)
+	out := make(chan node.Node)
 	done := make(chan struct{})
 
 	var setlk sync.Mutex
-
+	
+	errChan := make(chan error)
+	fetchersCtx, cancel := context.WithCancel(ctx)
+	
+	defer cancel()
+	
 	for i := 0; i < FetchGraphConcurrency; i++ {
 		go func() {
-			for n := range feed {
-				links := n.Links()
-				cids := make([]*cid.Cid, 0, len(links))
-				for _, l := range links {
-					setlk.Lock()
-					unseen := visit(l.Cid)
-					setlk.Unlock()
-					if unseen {
-						cids = append(cids, l.Cid)
-					}
+			for ic := range feed {
+				n, err := ds.Get(ctx, ic)
+				if err != nil {
+					errChan <- err
+					return
 				}
-
-				for nopt := range ds.GetMany(ctx, cids) {
+				
+				setlk.Lock()
+				unseen := visit(ic)
+				setlk.Unlock()
+				
+				if unseen {
 					select {
-					case out <- nopt:
-					case <-ctx.Done():
+					case out <- n:
+					case <-fetchersCtx.Done():
 						return
 					}
 				}
 				select {
 				case done <- struct{}{}:
-				case <-ctx.Done():
+				case <-fetchersCtx.Done():
 				}
 			}
 		}()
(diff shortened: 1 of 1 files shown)


Voker57:
Fixes #3588

merkledag/merkledag.go
```diff
@@ -460,16 +443,24 @@ func EnumerateChildrenAsync(ctx context.Context, ds DAGService, c *cid.Cid, visi
...
+		case l := <-out:
+			n, err := ds.Get(ctx, l)
+			if err != nil {
+				return err
 			}
 
 			if next == nil {
-				next = nc.Node
-				send = feed
+				if len(n.Links()) > 0 {
```
whyrusleeping: This can be cleaned up a little bit, Lets add all the new links to the array first, then check for next being nil and handle that case separately.

merkledag/merkledag.go
```diff
@@ -460,16 +443,24 @@ func EnumerateChildrenAsync(ctx context.Context, ds DAGService, c *cid.Cid, visi
...
+					next = n.Links()[0].Cid
+					for _, l := range n.Links()[1:] {
```
Kubuxu: It might be beneficial to: preallocate array for Cids in `.Links()` call, write the Cids to that array and do just one append call. 

This way the todobuffer will be expanded just once instead of N times.

merkledag/merkledag.go
```diff
@@ -460,16 +443,24 @@ func EnumerateChildrenAsync(ctx context.Context, ds DAGService, c *cid.Cid, visi
...
+						todobuffer = append(todobuffer, l.Cid)
+					}
+					send = feed
+				}
 			} else {
-				todobuffer = append(todobuffer, nc.Node)
+				for _, l := range n.Links() {
+					todobuffer = append(todobuffer, l.Cid)
```
Kubuxu: Same (https://github.com/ipfs/go-ipfs/pull/3598/files#r96632953) applies here.

The allocation of the additional array will probably be removed by optimization step in golang.

Voker57:
Addressed both reviews.

Kubuxu:
Can you also change the base of the PR to master, as the branch you are currently pointing to was merged already.

merkledag/merkledag.go
whyrusleeping: In general this isnt an issue in go, slices will optmistically allocate extra space under the hood for you, so sequential appends are generally just as performant as preallocating.

Voker57:
done

merkledag/merkledag.go
```diff
@@ -394,38 +394,21 @@ func EnumerateChildren(ctx context.Context, ds LinkService, root *cid.Cid, visit
...
-
-	feed := make(chan node.Node)
-	out := make(chan *NodeOption)
+	feed := make(chan *cid.Cid)
+	out := make(chan *cid.Cid)
 	done := make(chan struct{})
 
 	var setlk sync.Mutex
 
 	for i := 0; i < FetchGraphConcurrency; i++ {
```
whyrusleeping: Why are we spawning multiple of these goroutines? Looking at it, we should only need one.

merkledag/merkledag.go
```diff
@@ -460,16 +443,25 @@ func EnumerateChildrenAsync(ctx context.Context, ds DAGService, c *cid.Cid, visi
...
 			}
 
-			if next == nil {
-				next = nc.Node
-				send = feed
-			} else {
-				todobuffer = append(todobuffer, nc.Node)
+			links := n.Links()
+			if len(links) > 0 {
+				cids := make([]*cid.Cid, len(links), len(links))
```
whyrusleeping: no need to specify capacity if youre already specifying length to be the same

whyrusleeping: Actually, you can rewrite this whole loop to avoid allocating an extra slice. Just append to the todobuffer

merkledag/merkledag.go
Voker57: you mean FetchGraphConcurrency ones? You're right, looks like I accidentally removed concurrency altogether :) Fixed version incoming

Voker57:
Modified PR to actually be concurrent. Is second changes' request obsolete now?

merkledag/merkledag.go
```diff
@@ -394,56 +394,54 @@ func EnumerateChildren(ctx context.Context, ds LinkService, root *cid.Cid, visit
...
-
-				for nopt := range ds.GetMany(ctx, cids) {
+				
+				setlk.Lock()
+				unseen := visit(ic)
+				setlk.Unlock()
+				
+				if unseen {
+					cids := make([]*cid.Cid, len(n.Links()))
+					for i, l := range n.Links() {
```
whyrusleeping: This seems unecessary, why not just send the node down through out?

whyrusleeping: my worry is that its an extra allocation per node that we could avoid by simply moving this for loop down into the processing below.

merkledag/merkledag.go
```diff
@@ -460,18 +458,19 @@ func EnumerateChildrenAsync(ctx context.Context, ds DAGService, c *cid.Cid, visi
...
-			if nc.Err != nil {
-				return nc.Err
-			}
-
-			if next == nil {
-				next = nc.Node
-				send = feed
-			} else {
-				todobuffer = append(todobuffer, nc.Node)
+		case cids := <-out:
```
whyrusleeping: if out was changed to be a channel of nodes, this could look like:
```go
case nd := <-out:
  for _, lnk := range nd.Links() {
    if next == nil {
      next = lnk.Cid
      send = feed
      continue
    }
    todobuffer = append(todobuffer, lnk.Cid)
  }
```

merkledag/merkledag.go
```diff
@@ -394,56 +394,54 @@ func EnumerateChildren(ctx context.Context, ds LinkService, root *cid.Cid, visit
...
-	out := make(chan *NodeOption)
+	feed := make(chan *cid.Cid)
+	out := make(chan []*cid.Cid)
 	done := make(chan struct{})
 
 	var setlk sync.Mutex
-
+	
+	errChan := make(chan error)
+	fetchersCtx, cancel := context.WithCancel(ctx)
```
whyrusleeping: put a `defer cancel()` after this call and then remove the other call youre making to cancel

Voker57:
Implemented suggested changes.

PR #3600, opened 2017-01-15
Make pin rm not try to find pinned ancestors by default

Voker57:
Also make a switch to pin rm to allow old behavior.
Trying to look up pins which interfere with requested unpin
can be an extremely costly operation and typically is not
required by user.

### DIFF
--- a/pin/pin.go
+++ b/pin/pin.go
@@ -116,7 +116,7 @@ type Pinner interface {
 
 	// Unpin the given cid. If recursive is true, removes either a recursive or
 	// a direct pin. If recursive is false, only removes a direct pin.
-	Unpin(ctx context.Context, cid *cid.Cid, recursive bool) error
+	Unpin(ctx context.Context, cid *cid.Cid, recursive bool, explain bool) error
 
 	// Update updates a recursive pin from one cid to another
 	// this is more efficient than simply pinning the new one and unpinning the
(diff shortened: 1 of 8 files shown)


whyrusleeping:
@Voker57 Is this for the case where youre unpinning something 'recursively' and its pinned directly? I'm unsure the exact usecase here.

Voker57:
@whyrusleeping it's for the case when you're trying to unpin something which is not directly pinned: currently, in that case ipfs checks all the pins to see if they affect the requested nodes, and that can take huge amounts of time in case of big pinned items.

test/sharness/t0080-repo.sh
```diff
@@ -98,7 +104,7 @@ test_expect_success "remove recursive pin, add direct" '
 
 test_expect_success "remove direct pin" '
 	echo "unpinned $HASH" >expected6 &&
-	ipfs pin rm "$HASH" >actual6 &&
+	ipfs pin rm -r=false "$HASH" >actual6 &&
```
whyrusleeping: This behaviour shouldnt change though, `ipfs pin rm` should remove either a recursive or a direct pin

Voker57: But then there is no way to remove only recursive pins. -r will remove direct pins as well. Is that intended way for -r to behave?

Voker57:
@whyrusleeping IMHO default behavior should be "do not explain", since it involves potentially extremely costly operation and chances that user needs resulting information are quite low. Trying to unpin a certain block/file? Then they should know ahead to run pin rm -e to determine pin source. My target case is unpinning already unpinned hash, and I think this is more likely to happen.

test/sharness/t0080-repo.sh
whyrusleeping: Hrm... That was the behavior we initially intended to have. The idea was to mimic the unix `rm` command:

```sh
rm file #works
rm dir #doesnt work
rm -r file #works
rm -r dir #works
```

This does bring up the equivalent of having an `rmdir` analogue though.

What do you think @jbenet @Kubuxu @lgierth  ?

whyrusleeping:
@Voker57 Okay, you've convinced me that having do not explain be the default is the right thing to do. That feels more natural to me now

whyrusleeping:
(bumping to 0.4.7 as getting feedback on changing api behaviour will take a few days)

test/sharness/t0080-repo.sh
Voker57: It's not mimicing rm -r anyway, since it will not delete included pins.

whyrusleeping:
@Kubuxu @lgierth @kevina could I get you guys to review this?

Kubuxu:
Regarding code LGTM, regarding UI, I am not sure but this probably applies to current UI too.

test/sharness/t0080-repo.sh
kevina: I agree that `rm` is a bad model.  The type of a pin does not say anything about the type of a file. In addition I image most pins are recursive and direct pins are only used in special cases.

kevina:
This very closely relates to #3796.  Give me a few days to try this p.r. out and verify the current and changed behavior of of `ipfs pin rm`.

test/sharness/t0080-repo.sh
kevina: I partly take back what I said.  I left another comment.

kevina:
i don't like the change in behavior  `pin --recursive=true` should remove both recursive a direct pins, `pin --recursive=false` should only remove direct pins.

Voker57:
How to remove only recursive pins them? Technically they are completely unrelated, imo interface implying they are is misleading.

kevina:
I'm not sure, but, in my option there should be a way to remove a pin regardless of it's type.  We should not change the default behavior in any case.

Perhaps the recursive option should be removed, and replaced with a `--type` option that defaults to any, this will make it more explicit what is going on.  Or perhaps we can keep the recursive option as is and in addition add a `--type` option.

@whyrusleeping (and others) opinions?

Voker57:
It's not related to PR anyways, i just changed it because I thought current behavior is not intentional and test was not clear on that. I'll remove the change if -r is indeed supposed to remove both pins.

kevina:
Without explain enabled doing a `ipfs pin rm` on a direct pin will now return `Error: not pinned` which is misleading at best.

Voker57:
I've removed changing pin rm -r behavior from PR.

pin/pin.go
```diff
@@ -84,7 +84,7 @@ type Pinner interface {
 	IsPinned(*cid.Cid) (string, bool, error)
 	IsPinnedWithType(*cid.Cid, PinMode) (string, bool, error)
 	Pin(context.Context, node.Node, bool) error
-	Unpin(context.Context, *cid.Cid, bool) error
+	Unpin(context.Context, *cid.Cid, bool, bool) error
```
kevina: Please add parameter to Pin and Unpin so it is clear what the bool parameters are far.

Voker57: named.

Voker57:
Appears to be already changed during some overhaul.

Stebalien:
This was fixed in #6311. I had no idea there was an existing PR.

PR #3605, opened 2017-01-17
add in option to use nil-routing

whyrusleeping:
License: MIT


whyrusleeping:
CHANGELOG:
Add option to disable routing entirely.

PR #3606, opened 2017-01-18
Command name resolve: add -k flag, fix #3567

### LINKED ISSUE #3567: `ipfs name resolve` should support a `-k` option just like `name publish` does
#### Version information:
```
go-ipfs version: 0.4.5-dev-d83a911
Repo version: 4
System version: amd64/linux
Golang version: go1.7.4
```

#### Type:
`Enhancement`

#### Priority:
`P4`

#### Description:
As per subject: I should be able to 
`ipfs name resolve -k foo` after I've done `ipfs name publish -k foo <hash>`

FrankPetrilli:
Adds a -k option to the ipfs name resolve command, allowing resolve of alternate keys you're holding. Fix #3567

License: MIT
Signed-off-by: Frank Petrilli <frank@petril.li>


core/commands/ipns.go
```diff
@@ -89,15 +96,29 @@ Resolve the value of a reference:
 		}
 
 		var name string
-		if len(req.Arguments()) == 0 {
-			if n.Identity == "" {
-				res.SetError(errors.New("Identity not loaded!"), cmds.ErrNormal)
+
+		if kname, found, _ := req.Option("key").String(); found {
```
whyrusleeping: I think we want to disallow having a key specified and an argument specified at the same time. Either you use `ipfs name resolve -k=foo` or `ipfs name resolve /ipns/QmFooBar`, `ipfs name resolve -k=foo /ipns/QmFooBar` doesnt really make much sense.

core/commands/ipns.go
```diff
@@ -42,6 +43,11 @@ Resolve the value of a reference:
   > ipfs name resolve ipfs.io
   /ipfs/QmaBvfZooxWkrv7D3r8LS9moNjzD2o525XMZze69hhoxf5
 
+Resolve the value of another key:
```
whyrusleeping: Perhaps: `Resolve value of key from local keystore:`

Kubuxu:
Looks quite good, it for sure requires sharness tests.

whyrusleeping:
@FrankPetrilli update here? I'm pushing this to 0.4.7, hopefully we can land it then :)

FrankPetrilli:
It looks like the requested changes resolve the issues you brought up; @Kubuxu, I'll add test cases for this in just a few.

core/commands/ipns.go
```diff
@@ -8,6 +8,7 @@ import (
 	cmds "github.com/ipfs/go-ipfs/commands"
 	namesys "github.com/ipfs/go-ipfs/namesys"
 	offline "github.com/ipfs/go-ipfs/routing/offline"
+	peer "gx/ipfs/QmfMmLGoKzCHDN7cGgk64PJr4iipzidDRME8HABSJqvmhC/go-libp2p-peer"
```
Kubuxu: There should be a new line between local and gx imports.

test/sharness/t0100-name.sh
```diff
@@ -60,6 +60,15 @@ test_expect_success "ipfs cat on published content succeeds" '
     test_cmp expected actual
 '
 
+# test with a custom key
+test_expect_success "ipfs resolves published key from keystore" '
```
whyrusleeping: Lets break this test out into a few more steps.

First assertion should just generate a key (and not have the || true bit, if it fails we want to fail).

Second assertion should do the publish, and check its output

Third assertion should do the resolve

core/commands/ipns.go
```diff
@@ -89,15 +97,33 @@ Resolve the value of a reference:
 		}
 
 		var name string
-		if len(req.Arguments()) == 0 {
-			if n.Identity == "" {
-				res.SetError(errors.New("Identity not loaded!"), cmds.ErrNormal)
+
+		if kname, found, _ := req.Option("key").String(); found {
+			if len(req.Arguments()) > 0 {
```
whyrusleeping: Lets rework this a bit to reduce cyclomatic complexity:
```go
kname, found, _ := req.Option("key").String()
if found && len(req.Arguments()) > 0 {
  // fail with that error
}

var name string
if len(req.Arguments()) > 0 {
  name = req.Arguments()[0]
} else {
// 'GetKey' logic from line 106 below to find 'name'
}
```

whyrusleeping:
@FrankPetrilli update here?

whyrusleeping:
@Kubuxu could you take over this PR?

Kubuxu:
Sure, I will have to reopen it as separate PR as this one is done from a fork.

FrankPetrilli:
Apologize for the delay, I've been focused on other areas and the GitHub notifications were lost in the noise.

I'll get a change pushed to take your input into account. Thank you!

whyrusleeping:
@FrankPetrilli Thanks!

PR #3608, opened 2017-01-18
Use newer flatfs sharding scheme

whyrusleeping:
License: MIT


repo/fsrepo/defaultds.go
```diff
@@ -35,7 +35,7 @@ func openDefaultDatastore(r *FSRepo) (repo.Datastore, error) {
 	syncfs := !r.config.Datastore.NoSync
 	// 5 bytes of prefix gives us 25 bits of freedom, 16 of which are taken by
 	// by the Qm prefix. Leaving us with 9 bits, or 512 way sharding
-	blocksDS, err := flatfs.New(path.Join(r.path, flatfsDirectory), 5, syncfs)
+	blocksDS, err := flatfs.CreateOrOpen(path.Join(r.path, flatfsDirectory), flatfs.NextToLast(2), syncfs)
```
mateon1: The comment needs to be updated as well. (again)
If I understand correctly, that this is the last two base32 characters, it's 10 bits of sharding.

whyrusleeping: @mateon1 good catch on the comment. Yeah, its the second to last two base32 characters (since the last character doesnt have 10 full bits of entropy)

whyrusleeping:
Alright, here goes

whyrusleeping:
CHANGELOG:
fsrepo version 5, new flatfs sharding scheme.

PR #3615, opened 2017-01-20
Add multiple performance metrics

### LINKED ISSUE #3607: Testing Metrics
@jbenet and I brainstormed a list of metrics we want integrated into go-ipfs and exported to prometheus for testing.

- [x] Duplicate Blocks Received
- [ ] BW usage
  - [x] up
  - [x] down
- [x] api io throughput - not per end point
- [x] datastore metrics (measure-ds output)
- [ ] dht traffic
- [ ] bitswap traffic
- [ ] total blocks added/received/stored
  - [x] if datastore metrics is enough
-  [ ] pinset sizes (by type)
- [ ] dht bucked fill ratio

@lgierth Since you know the most about these, could you lead getting these put in?

Kubuxu:
Resolves #3607

### DIFF
--- a/cmd/ipfs/daemon.go
+++ b/cmd/ipfs/daemon.go
@@ -28,8 +28,6 @@ import (
 	"gx/ipfs/QmX3QZ5jHEPidwUrymXV1iSCSUhdGxj15sm2gP4jKMef7B/client_golang/prometheus"
 	util "gx/ipfs/QmZuY8aV7zbNXVy6DyN9SmnuH3o9nG852F4aTiSBpts8d1/go-ipfs-util"
 	iconn "gx/ipfs/QmcYnysCkyGezY6k6MQ1yHHdrRiZaU9x3M9Y1tE9qZ5hD2/go-libp2p-interface-conn"
-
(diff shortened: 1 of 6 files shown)


whyrusleeping:
@Kubuxu build failure

Kubuxu:
@whyrusleeping fixed

flyingzumwalt:
Related to today's standup discussion, this PR doesn't add any tests or documentation. 
* It doesn't make it clear _how_ you would get these metrics out of go-ipfs when you run it
* It doesn't test that the metrics are actually being recorded when you run the code

whyrusleeping:
@flyingzumwalt Thanks, mind filing an issue about needing to add tests for metrics?

Since we don't currently have the setup to test metrics in a meaningful way, adding them here is out of scope for this PR as implementing it would be a non-significant endeavor and would take a good hunk of time from our already tight timeline.

Kubuxu:
Good after it passes.

whyrusleeping:
@Kubuxu OSX sharness failures don't appear to be a fluke.

Kubuxu:
It is same as: https://github.com/ipfs/go-ipfs/issues/3677#issuecomment-279249187

whyrusleeping:
@Kubuxu I'm sorry, can you rebase?

Kubuxu:
Rebased

Kubuxu:
❤️

PR #3622, opened 2017-01-22
docker: expose port for /tcp/ws transport

lgierth:
Picked :8081 here because the websockets transport's usage seemed more closely related to the gateway.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,11 +7,16 @@ MAINTAINER Lars Gierth <lgierth@ipfs.io>
 # Please keep these two Dockerfiles in sync.
 
 
-# Ports for Swarm TCP, Swarm uTP, API, Gateway
+# /tcp transport
 EXPOSE 4001
+# /udp/utp transport
 EXPOSE 4002/udp
+# api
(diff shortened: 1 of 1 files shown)


lgierth:
Ideally we'd be able to mount the websockets transports within the gateway.

Kubuxu:
IMHO as it isn't default option we should have separate Docker container and tag for it .

lgierth:
> IMHO as it isn't default option we should have separate Docker container and tag for it .

I feel like that'd just make things more complicated than they are -- it starts with a ws/nows conditional in the infra scripts.

What's the downside to having that port exposed?

Kubuxu:
More ports taken when you start IPFS from docker, if someone starts multiple he has to map yet another port.

lgierth:
Mh yeah I understand. Maybe we should just generally work towards having the /ws listener on :8080?

whyrusleeping:
Have we reached a decision here?

lgierth:
I have WIP code towards muxing the /ws listener into the gateway -- will finish up soon (tm)

Issue #3397, opened 2016-11-18
List of experimental IPFS features


whyrusleeping:
**Update 23-Jul-2017** -- Since #4036 experimental features are documented in [docs/experimental-features.md](https://github.com/ipfs/go-ipfs/blob/master/docs/experimental-features.md).

------

IPFS currently has a few different features that are in the 'experimental' phase. This might be because we arent fully confident in releasing the code as an official feature yet, or because we don't want to make a certain change to the network yet. Whatever the case, we should have a log of all of those features and their states.

Lets use this issue to track the state of these features, and maybe think about putting this info in a doc that we commit to version control.

whyrusleeping:
## `ipfs pubsub`

### State
experimental, default-disabled.

### In Version
master, unreleased.

### How to enable
run your daemon with the `--enable-pubsub-experiment` flag. Then use the `ipfs pubsub` commands.

### Road to being a real feature
- [ ] Needs more people to use and report on how well it works
- [ ] Needs authenticated modes to be implemented
- [ ] needs performance analyses to be done

whyrusleeping:
## Client mode DHT routing
Allows the dht to be run in a mode that doesnt serve requests to the network, saving bandwidth.

### State
experimental.

### In Version
master, unreleased.

### How to enable
run your daemon with the `--routing=dhtclient` flag.

### Road to being a real feature
- [ ] Needs more people to use and report on how well it works.
- [ ] Needs analysis of effect it has on the network as a whole.

whyrusleeping:
## go-multiplex stream muxer
Adds support for using the go-multiplex stream muxer alongside (or instead of) yamux and spdy. This multiplexer is far simpler, and uses less memory and bandwidth than the others, but is lacking on congestion control and backpressure logic. It is available to try out and experiment with.

### State
Experimental

### In Version
master, unreleased

### How to enable
run your daemon with `--enable-mplex-experiment`

To make it the default stream muxer, set the environment variable `LIBP2P_MUX_PREFS` as follows:
```
export LIBP2P_MUX_PREFS="/mplex/6.7.0 /yamux/1.0.0 /spdy/3.1.0"
```

To check which stream muxer is being used between any two given peers, check the json output of the `ipfs swarm peers` command, you'll see something like this:
```
$ ipfs swarm peers -v --enc=json | jq .
{
  "Peers": [
    {
      "Addr": "/ip4/104.131.131.82/tcp/4001",
[...]

PR #3629, opened 2017-01-24
Implement basic filestore 'no-copy' functionality

whyrusleeping:
As per this design doc: https://gist.github.com/whyrusleeping/5565651011b49ddc9ddeec9ffd169050

License: MIT


filestore/filestore.go
```diff
@@ -129,6 +129,15 @@ func (f *Filestore) Has(c *cid.Cid) (bool, error) {
...
+	has, err := f.Has(b.Cid())
+	if err != nil {
+		return err
+	}
+
+	if has {
+		return nil
+	}
+
 	switch b := b.(type) {
```
kevina: @whyrusleeping I don't know what you have in mind for the filestore.  But I very specially designed my filestore so that writes go though.  This way if a file is moved it can simply be readded and any invalid blocks are automatically fixed.

whyrusleeping: That becomes problematic, What then happens if i add dataset A, which contains some subset C. Then later, i add dataset B that also contains some subset C. Then i change my mind and remove dataset B. By your description, the original dataset references would now be lost, with no indication to the user that this has happened.

I think this scenario is worse UX than the reverse (deleting A causing bad refs in B).  In either case, repairs are going to have to be manual, and some things will get sticky.

cc @jbenet for his thoughts on this scenario.

kevina:
> That becomes problematic, What then happens if i add dataset A, which contains some subset C. Then later, i add dataset B that also contains some subset C. Then i change my mind and remove dataset B. By your description, the original dataset references would now be lost, with no indication to the user that this has happened.

Yes this was originally a problem.  In my implementation I fixed this.  See https://github.com/ipfs-filestore/go-ipfs/issues/12 and https://github.com/ipfs-filestore/go-ipfs/pull/23.  However, that solution also requires writes to go through.

> I think this scenario is worse UX than the reverse (deleting A causing bad refs in B). In either case, repairs are going to have to be manual, and some things will get sticky.

Not if my solution in https://github.com/ipfs-filestore/go-ipfs/issues/12 is implemented.

> cc @jbenet for his thoughts on this scenario.

kevina:
> Yes this was originally a problem. In my implementation I fixed this. See ipfs-filestore#12 and ipfs-filestore#23. However, that solution also requires writes to go through.

Another option is to only replace the block if it is invalid.  I considered that, but decided that it was better to support allowing multiple files to point to the same block.

filestore/fsrefstore.go
```diff
@@ -0,0 +1,190 @@
...
+
+	var dobj pb.DataObj
+	if err := proto.Unmarshal(data, &dobj); err != nil {
+		return nil, err
+	}
+
+	out, err := f.readDataObj(&dobj)
+	if err != nil {
+		return nil, err
+	}
```
whyrusleeping: Ah, yeah. `NewBlockWithCid` only does that check if `u.Debug` is true. good catch

kevina:
I see that "--nocopy" is allowed when the daemon is online without any sort of additional checks.  _If the blocks are ever not verified this could become a security problem_.  Even without the security problems this could lead to strange results if files are added on another machine with an identical path.

kevina:
For this to be usable outside of `ipfs-pack` I would strongly recommend that references be allowed "outside the root handler" by simply requiring all paths to be absolute.

filestore/fsrefstore.go
```diff
@@ -0,0 +1,199 @@
...
+	dsns "gx/ipfs/QmRWDav6mzWseLWeYfVd5fvUKiVe9xNH29YfMF438fG364/go-datastore/namespace"
+	dsq "gx/ipfs/QmRWDav6mzWseLWeYfVd5fvUKiVe9xNH29YfMF438fG364/go-datastore/query"
+	proto "gx/ipfs/QmT6n4mspWYEya864BhCUJEgyxiRfmiSY9ruQwTUNpRKaM/protobuf/proto"
+	cid "gx/ipfs/QmcTcsTvfaeEBRFo1TkFgT8sRmgi1n1LTZpecfVP8fzpGD/go-cid"
+)
+
+var FilestorePrefix = ds.NewKey("filestore")
+
+type FileManager struct {
+	ds   ds.Batching
```
kevina: Note.  In my early implementations I wrote the filestore on top of the datastore also.  However, I changed this to directly use the leveldb as I found little value in the abstraction and extra layers.  What actually prompted it was https://github.com/ipfs-filestore/go-ipfs/issues/10.  In particular querying the datastore was very very slow, however I then latter find it very useful to use the features of the LevelDB in order to provide safe verification of the filestore while the deamon was running by using the LevelDB snapshot feature.  With my recent improvements to the datastore query the performance may no longer be a problem, but I still see little value in the extra indirection.

kevina:
@whyrusleeping concerning replacing blocks.  How about for now we make it an option? There needs to be a way to fix broken blocks due to files moving.  One option would be to remove the block and then readd it, but pinning makes that a more complicated operation.

As far as allowing multiple blocks for a hash (ipfs-filestore#12 and ipfs-filestore#23), I agree that we should aim for that but to start with it might be a bit much to review.

filestore/fsrefstore.go
```diff
@@ -0,0 +1,199 @@
...
+func (f *FileManager) AllKeysChan(ctx context.Context) (<-chan *cid.Cid, error) {
+	q := dsq.Query{KeysOnly: true}
+	q.Prefix = FilestorePrefix.String()
+
+	res, err := f.ds.Query(q)
+	if err != nil {
+		return nil, err
+	}
+
+	out := make(chan *cid.Cid)
```
kevina: I would suggest you use a buffered channel.  Use dsq.KeysOnlyBufSize.

filestore/filestore.go
```diff
@@ -0,0 +1,192 @@
...
+
+func (f *Filestore) AllKeysChan(ctx context.Context) (<-chan *cid.Cid, error) {
+	ctx, cancel := context.WithCancel(ctx)
+
+	a, err := f.bs.AllKeysChan(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	out := make(chan *cid.Cid)
```
kevina: I would suggest you use a buffered channel. Use dsq.KeysOnlyBufSize.

jefft0:
Is this branch ready for me to test on my 30 TB of webcam videos or should I wait?

whyrusleeping:
@jefft0 Its almost ready, should be merged in a couple days.

whyrusleeping:
Rebased, waiting on a :+1: from @Kubuxu and CI

kevina:
At @jefft0 I am really not sure how usable this code is going to be for you as it does not allow for files "outside the root handler".  For example if you directory layout is:
```
/home/jeff0/.ipfs
/aux/media
```
and you want to share files in /aux/media you won't be able to and will get an error.

kevina:
@whyrusleeping I would like this comment, to at least be addressed:

> I see that "--nocopy" is allowed when the daemon is online without any sort of additional checks. If the blocks are ever not verified this could become a security problem. Even without the security problems this could lead to strange results if files are added on another machine with an identical path.

whyrusleeping:
@kevina @jefft0 In this case, simply set your `IPFS_PATH` to be in (given the example) `/aux/ipfs` or similar, We're pushing for git style security on file references. They need to be within the scope of the ipfs repository.

@kevina Re the comment, blocks are checked now (thanks for pointing that out), and this feature should not be used with client and daemon on separate machines. We can probably implement a check of some sort for this... any ideas on how to do that off the top of your head?

kevina:
> @kevina @jefft0 In this case, simply set your IPFS_PATH to be in (given the example) /aux/ipfs or similar, We're pushing for git style security on file references. They need to be within the scope of the ipfs repository.

This was never discussed with me and I do not approve of this decision.  But I guess there is no point in starting an extended discussion here.

kevina:
> @kevina Re the comment, blocks are checked now (thanks for pointing that out), and this feature should not be used with client and daemon on separate machines. We can probably implement a check of some sort for this... any ideas on how to do that off the top of your head?

No, that is why I recommend it be disabled by default until we can figure out how to make this secure.

In my code I did a check when, if a file is added when the daemon is online, the contents of the file are sent in addition to the filename.  The file is then opened by the daemon and it verifies that the sent contents match what is on the filesystem.  Not 100% fool proof but minimized the vector of attack.

whyrusleeping:
@kevina actually, the 'repo root' check is what we were using as the safety mechanism for this. Only files within the scope of the ipfs repo can be touched, it does not provide access to any file on the machine.

test/sharness/t0270-filestore.sh
```diff
@@ -0,0 +1,93 @@
...
+
+
+test_expect_success "create a dataset" '
+	random-files -seed=483 -depth=3 -dirs=4 -files=6 -filesize=1000000 somedir
+'
+
+EXPHASH="QmW4JLyeTxEWGwa4mkE9mHzdtAkyhMX2ToGFEKZNjCiJud"
+
+get_repo_size() {
+	du -s -b "$IPFS_PATH" | awk '{ print $1 }'
```
whyrusleeping: Hrm. this doesnt work on OSX. the `-b` option isnt valid on BSD du

kevina:
> actually, the 'repo root' check is what we were using as the safety mechanism for this. Only files within the scope of the ipfs repo can be touched, it does not provide access to any file on the machine.

I don't see that as a very strong safety mechanism.  In particular with the default location of the .ipfs repo $HOME/.ipfs a users home directory could accidentally be exposed.

kevina:
@jefft0 please also note that this implementation requires the use of the --raw-leaves option.  This means that files shared with this code will not be usable in older (pre 0.4.5 I think) versions of IPFS.

whyrusleeping:
>  In particular with the default location of the .ipfs repo $HOME/.ipfs a users home directory could accidentally be exposed.

Right, and that said, the user must manually enable the filestore feature via editing the config. We should be very clear on the messaging for this.

filestore/fsrefstore.go
```diff
@@ -0,0 +1,210 @@
...
+	if !c.Equals(outcid) {
+		return nil, &CorruptReferenceError{fmt.Errorf("data in file did not match. %s offset %d", d.GetFilePath(), d.GetOffset())}
+	}
+
+	return outbuf, nil
+}
+
+func (f *FileManager) Has(c *cid.Cid) (bool, error) {
+	// NOTE: interesting thing to consider. Has doesnt validate the data.
+	// So the data on disk could be invalid, and we could think we have it.
```
Kubuxu: Maybe it could be under the same config as `HashOnRead`

core/commands/add.go
```diff
@@ -154,14 +166,20 @@ You can now refer to the added file in a gateway, like so:
 			n = nilnode
 		}
 
-		dserv := n.DAG
+		addblockstore := n.Blockstore
+		if !fscache && !nocopy {
+			addblockstore = bstore.NewGCBlockstore(n.BaseBlocks, n.GCLocker)
```
whyrusleeping: Its creating a blockstore without the filestore, reusing the same gclock

Kubuxu: Ahh

jefft0:
I'm confused. I have several webcam videos totalling 60 GB. With a fresh installation of go-ipfs (master branch), I did `ipfs init` then `ipfs add --nocopy --raw-leaves` for each file. But now `~/.ipfs` is 60 GB. Is `--nocopy` working yet in master?

Kubuxu:
@whyrusleeping anything on this?

whyrusleeping:
Ah, theres a check i forgot to add, erroring out if `--nocopy` is passed without enabling the filestore.

@jefft0 take a look at "How to enable" here: https://github.com/ipfs/go-ipfs/issues/3397#issuecomment-284337564

jefft0:
@whyrusleeping Much better, thanks! Now `~/,ipfs` is only 68 MB. I'll try adding a lot more files. Where do you want me to report my results?

whyrusleeping:
@jefft0 You can report them here, or open a new issue to discuss. Either way :)

jefft0:
@whyrusleeping I want to keep my main repo in `~/.ipfs`, so instead of changing `IPSF_PATH` I put a symbolic link to my external drive in `~/.ipfs` and do `ipfs add --nocopy` with that path. This works. It would also let me use multiple external drives at different mount points. Do you see any downside to this approach?

whyrusleeping:
@jefft0 hrm... unsure about the implications of this. I'll have to think about the security model of using/allowing symlinks like that.

My first thought though is that it should be fine

whyrusleeping:
@jefft0 you can also put the symbolic link in just `~/`, it doesnt have to be IN the ipfs dir, just in its parent (like the .git folder for git repos)

atommixz:
every time I get
```
$ ps aux | grep ipfs
atommixz  3109 36.2  2.5 879840 207656 ?       Ssl  23:20   1:24 /usr/bin/ipfs daemon --manage-fdlimit=true

$ LANG=C type ipfs_add
ipfs_add is aliased to `ipfs add --nocopy --raw-leaves --recursive --progress'

$ ipfs_add download/torrents
added QmcGUHBF2XT8dTblablablazYsNQch6ECGx transmission/file.ext
 3.24 GB / 61.22 GB [==>---------------------------------------------------]   5.30% 15m33s23:23:30.659 ERROR commands/h: open /home/username/.ipfs/blocks/HE/put-789802368: too many open files client.go:247
Error: open /home/username/.ipfs/blocks/HE/put-789802368: too many open files
```

whyrusleeping:
@atommixz Wanna open a new issue for this?

atommixz:
This helps to me:
```
$ systemctl --user edit --full ipfs 
[Service]
LimitNOFILE=999999
```
This is don't make sence to me `ipfs daemon --manage-fdlimit=true`

atommixz:
I'm trying to download many images from filestore. Everytime it stops at the same place.
```
$ ipfs get Qmblablabla
Saving file(s) to Qmblablabla
 117.00 MB / 2.36 GB [==>------------------------------------------------------]   4.84% 4s
Error: expected protobuf dag node

...
23:49:27.028 ERROR    bitswap: couldnt open sender again after SendMsg(<peer.ID PRB9SN>) failed: dial attempt failed: <peer.ID Rj75eS> --> <peer.ID PRB9SN> dial attempt failed: i/o timeout wantmanager.go:233
23:50:03.062 ERROR    bitswap: couldnt open sender again after SendMsg(<peer.ID aoqJoJ>) failed: dial attempt failed: <peer.ID Rj75eS> --> <peer.ID aoqJoJ> dial attempt failed: i/o timeout wantmanager.go:233
23:51:15.626 ERROR commands/h: err: expected protobuf dag node handler.go:288
23:52:34.252 ERROR commands/h: err: expected protobuf dag node handler.go:288
23:52:59.859 ERROR commands/h: err: expected protobuf dag node handler.go:288

```

Kubuxu:
Can you report it in separate issue. It is much easier to track this way.

atommixz:
whyrusleeping sayd "You can report them here, or open a new issue to discuss. Either way :)"
so I do, because it's about filestore.

Kubuxu:
I think he meant results as it, it worked, performance, small problems.

whyrusleeping:
Yeah, Keeping perf and general 'how it went' things here is fine, but actual problems need to have their own issues so we can address them

jefft0:
I had a problem with too many open files and opened issue #3763. I don't know if it's related to the issue @Kubuxu  has.

core/commands/add.go
```diff
@@ -140,6 +145,13 @@ You can now refer to the added file in a gateway, like so:
 		chunker, _, _ := req.Option(chunkerOptionName).String()
 		dopin, _, _ := req.Option(pinOptionName).Bool()
 		rawblks, _, _ := req.Option(rawLeavesOptionName).Bool()
+		nocopy, _, _ := req.Option(noCopyOptionName).Bool()
+		fscache, _, _ := req.Option(fstoreCacheOptionName).Bool()
+
+		if nocopy && !rawblks {
```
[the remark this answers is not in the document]
Sameyboy88: MONSTERDOG

PR #3631, opened 2017-01-25
migrations: update dist.ipfs.io path

lgierth:


lgierth:
This is so it can actually find the fs-repo-migrations release it needs.

Hold off on merging this, I'm patching in the musl build and publishing an updated dist.ipfs.io.

lgierth:
Okay, should be good to go

Kubuxu:
CHANGELOG: update reference to fs-repo-migrations in the downloader

PR #3635, opened 2017-01-27
Add more info to bitswap stat

whyrusleeping:
Some better info on blocks sent out and data sent and received (at least in terms of data in blocks being sent)

License: MIT


Kubuxu:
This patchset has very high chance of conflicting with my metrics changes.

whyrusleeping:
@Kubuxu can you handle resolving that?

PR #3636, opened 2017-01-27
close file before removing it in check writeable

### LINKED ISSUE #3632: Flatfs initialization bombs on windows
The ultimate cause is something causes a ._check_writeable file to be written to the blocks directory. 

Then when this line of code checks that the directory is empty it returns an error: 

https://github.com/ipfs/go-ds-flatfs/blob/9553971a40e28fd9d1022c66c1dc0b9dc5e03451/flatfs.go#L63

A simple `os.Remove()` fixes it, though I'm not familiar enough with either windows or what IPFS is doing on init to know why that file gets written there.

whyrusleeping:
Should half-close #3632 (along with a flatfs update)

License: MIT

### DIFF
--- a/test/sharness/t0020-init.sh
+++ b/test/sharness/t0020-init.sh
@@ -63,7 +63,8 @@ test_expect_success ".ipfs/ has been created" '
 	test -d ".ipfs" &&
 	test -f ".ipfs/config" &&
 	test -d ".ipfs/datastore" &&
-	test -d ".ipfs/blocks" ||
+	test -d ".ipfs/blocks" &&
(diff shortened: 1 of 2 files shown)


whyrusleeping:
@Kubuxu in windows, the `Remove` call was erroring because we had an open valid file descriptor for the file.

Kubuxu:
Ahh, 100% LGTM then. I've forgot about Windows.

whyrusleeping:
CHANGELOG:
Fix 'Writeable' helper method behavior on windows.

PR #3639, opened 2017-01-27
Feat/0.4.5 changelog

whyrusleeping:
First pass at cleaning up the PR list into a changelog. cc @lgierth @Kubuxu

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,157 @@
 # go-ipfs changelog
 
+### 0.4.5
+
+- Notable changes
+	- IPLD and CIDs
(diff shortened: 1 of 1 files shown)


Kubuxu:
I am not a fan of those empty check mark boxes.

whyrusleeping:
oh, yeah. didnt look at it in non-text mode. Lemme remove those

whyrusleeping:
Removed the checkboxes

Kubuxu:
Your sed eat up some letters from the middle of the words.

lgierth:
The IPLD and deps sections should be reworded before release, but this LGTM :+1:

PR #3640, opened 2017-01-28
Make pinset sharding deterministic

### LINKED ISSUE #3621: 0.4.5: repo does not scale beyond ~ 8000 files
#### Version information:
go-ipfs version: 0.4.5-dev-
Repo version: 5
System version: amd64/linux
Golang version: go1.7.4

#### Type:
Bug

#### Priority:
P0
#### Description:

When adding a lot of files locally (or using the daemon without `--enable-gc`), the repo grows even faster in size after a certain limit. 
It seems to be around 8000, my gut tells me it's 8192 :dizzy:
Unless you have a lot of space with good IO, this makes it very hard to import files without
garbage collecting in between (say every 1k).
This is probably a blocker for 0.4.5

Here is a fancy graph that displays the problem:
Adding 12000 files

After running `ipfs repo gc` the repo size shrinks from 16 GiB to 78 MiB

If you want to reproduce it, go and fetch QmcsrSRuBmxNxcEXjMZ1pmyRgnutCGwfAhhnRfaNn9P94F [@ ipfs.io](https://ipfs.io/ipfs/QmcsrSRuBmxNxcEXjMZ1pmyRgnutCGwfAhhnRfaNn9P94F)

whyrusleeping:
Making this deterministic keeps us from creating an exponential amount
of objects as the number of pins in the set increases.

closes #3621 for the most part
License: MIT

### DIFF
--- a/pin/set.go
+++ b/pin/set.go
@@ -3,7 +3,6 @@ package pin
 import (
 	"bytes"
 	"context"
-	"crypto/rand"
 	"encoding/binary"
 	"errors"
 	"fmt"
@@ -25,14 +24,6 @@ const (
 	// maxItems is the maximum number of items that will fit in a single bucket
 	maxItems = 8192
 )
-
-func randomSeed() (uint32, error) {
-	var buf [4]byte
-	if _, err := rand.Read(buf[:]); err != nil {
-		return 0, err
-	}
-	return binary.LittleEndian.Uint32(buf[:]), nil
-}
 
 func hash(seed uint32, c *cid.Cid) uint32 {
 	var buf [4]byte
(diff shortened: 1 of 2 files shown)


rht:
Confirmed there is no longer exponential explosion, https://github.com/rht/sfpi-benchmark/blob/master/repo_size/graph_10k.png, but pinning still reverses the benefit of block-level dedup.

Kubuxu:
> but pinning still reverses the benefit of block-level dedup.

What you mean by that?

Kubuxu:
From that graph it seems for me that we should start sharding the pinset a bit more early.

rht:
I meant: the storage-saving effect of the block-dedup is shadowed by the pin set. In short, files in an IPfs repo take more space than they were in UNIXfs.

Kubuxu:
IDK if we should go with fully static seed, it allows for same attacks that all languages protect their hash maps against (precalcing which items would go to which buckets and causing requests that would cause many items in one bucket) but in our case it would be causing the tree to have one deep branch.

whyrusleeping:
@Kubuxu I don't think this is an issue, You would have to find 257 items that all share a hash prefix of length n, where the hash function changes at each byte index. And then convince another node to pin each of them individually.

Kubuxu:
Yeah, the split to sub buckets helps in comparison to hash maps.

whyrusleeping:
@Kubuxu :+1: here?

PR #3649, opened 2017-02-03
update go-multihash and bubble up deps

whyrusleeping:
Updated go-multihash with the blake2b and 'double sha256' changes. This was a large process.

Other than the go-multihash changes, two other changes were made:

#### go-multibase was updated in go-cid
one tree bubbling was enough, and its a minor change. 

#### the 'go-net' package had its go-crypto dependency removed
This was done because go-multihash was updated to use a new version of go-crypto, and no user of the 'go-net' package used the code in go-net that required go-crypto.

License: MIT


Kubuxu:
Is it too late to introduce into go-cid utility method that would be equivalent of `String()` but with option of base encoding selection. It could be quite useful in some places.

whyrusleeping:
@Kubuxu we could do it now. Though would be just as easy to do later. Up to you

Kubuxu:
As I understand it includes varint update to multihash, then it will require quite some testing.

whyrusleeping:
rebased on master

whyrusleeping:
I'm thinking that because of this one, we will probably want to ship a pre-release of 0.4.6 next week so that the hex encoded cids will work for browser integration stuff.

cc @flyingzumwalt @Kubuxu @lgierth

Kubuxu:
Ok, lets do it, as the tree update was already done, redoing it has the same cost.

LGTM if tests pass.

PR #3653, opened 2017-02-04
Basic Filestore Utilties

kevina:


kevina:
Only `filestore ls` and `filestore verify` are implemented now.  Still needs tests.

kevina:
@whyrusleeping please let me know if you are okay with these changes so far.

filestore/util.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+func next(qr dsq.Results) (*cid.Cid, *pb.DataObj, error) {
+	v, ok := qr.NextSync()
+	if !ok {
+		return nil, nil, nil
+	}
+
+	k := ds.RawKey(v.Key)
+	c, err := dshelp.DsKeyToCid(k)
```
whyrusleeping: shouldnt this reuse the existing method for reading a dataobj ?

filestore/util.go
```diff
@@ -0,0 +1,184 @@
...
+func mkListRes(c *cid.Cid, d *pb.DataObj, err error) *ListRes {
+	status := StatusOk
+	errorMsg := ""
+	if err != nil {
+		if err == ds.ErrNotFound || err == blockstore.ErrNotFound {
+			status = StatusKeyNotFound
+		} else if err, ok := err.(*CorruptReferenceError); ok {
+			switch err.Code {
+			case FileError:
+				status = StatusFileError
```
whyrusleeping: Why not just reuse the codes from the `CorruptReferenceError` ?

whyrusleeping:
Also make sure to clean out commented out imports

core/commands/filestore.go
```diff
@@ -0,0 +1,139 @@
...
+			return
+		}
+		out := listResAsChan(next, req.Context())
+		res.SetOutput(out)
+	},
+	PostRun: func(req cmds.Request, res cmds.Response) {
+		if res.Error() != nil {
+			return
+		}
+		outChan, ok := res.Output().(<-chan interface{})
```
kevina: This is in PostRun becuase I output to both stdout and stderr something that a marshaler can't do, unless I am mistaken.

filestore/util.go
kevina: what method is that?

kevina:
> Also make sure to clean out commented out imports

Will do towards the end, some of them might still be needed.

core/commands/filestore.go
kevina: Note, I did a very similar thing in `ipfs block rm`: https://github.com/ipfs/go-ipfs/blob/master/core/commands/block.go#L283.

filestore/util.go
kevina: The set of codes returned in ListRes is larger than the codes for CorruptReferenceError.  For a little less type safety I can have CorruptReferenceError use the same codes in ListRes if you prefer it that way,

core/commands/filestore.go
kevina: Also `ipfs add` does not use a marshaler.

filestore/util.go
whyrusleeping: Yeah, lets unify the sets of codes

kevina: Will do.

kevina:
@whyrusleeping this is about done, the final command I plan to add is "clean" which will remove invalid blocks.  I will get to that early wed.

core/commands/filestore.go
whyrusleeping: This should be able to go in the marshaler. You can write to stderr and stdout in the marshaler the same way you can in the PostRun. If we do weird things like this in the post run then it breaks the API expectations in a weird way.

core/commands/filestore.go
```diff
@@ -0,0 +1,315 @@
...
+				if err != nil {
+					res.SetError(err, cmds.ErrNormal)
+					return
+				}
+				if have {
+					fmt.Fprintf(writer, "%v\n", cid)
+				}
+			}
+		}()
+		res.SetOutput(reader)
```
whyrusleeping: Don't use a reader for output like this. Send objects over a channel like we do in the other commands so that the api is easier to deal with.

core/commands/filestore.go
kevina: @whyrusleeping so should the Marshalers just return nil for the Reader?

filestore/util.go
```diff
@@ -0,0 +1,175 @@
...
+	Key      *cid.Cid
+	FilePath string
+	Offset   uint64
+	Size     uint64
+}
+
+func (r *ListRes) FormatLong() string {
+	switch {
+	case r.Key == nil:
+		return "?????????????????????????????????????????????????"
```
whyrusleeping: If this is never expected to happen, we should probably panic here instead of printing an ugly string

filestore/util.go
whyrusleeping: this one? https://github.com/ipfs/go-ipfs/blob/feat/filestore0/filestore/fsrefstore.go#L100

core/commands/filestore.go
kevina: Okay, all I am sending is Cid's, is there a Marshaler I can reuse?

whyrusleeping: The one from `ipfs refs` should work

whyrusleeping:
Also make sure to clean up the unused imports

test/sharness/t0271-filestore-utils.sh
```diff
@@ -0,0 +1,194 @@
...
+	test_expect_success "nocopy add succeeds" '
+		HASH=$(ipfs add --raw-leaves --nocopy -r -q somedir | tail -n1)
+	'
+
+	test_expect_success "nocopy add has right hash" '
+		test "$HASH" = "$EXPHASH"
+	'
+
+	test_expect_success "'ipfs filestore ls' output looks good'" '
+		ipfs filestore ls | LC_ALL=C sort > ls_actual &&
```
whyrusleeping: What is the `LC_ALL=C` for? That seems really weird

test/sharness/t0271-filestore-utils.sh
```diff
@@ -0,0 +1,194 @@
...
+		ipfs filestore verify > verify_actual &&
+		grep no-file verify_actual | grep -q somedir/file1
+'
+
+	test_expect_success "move file back" '
+		mv somedir/file1.bk somedir/file1
+'
+
+	test_expect_success "block okay now" '
+		ipfs cat $FILE1_HASH > /dev/null
```
whyrusleeping: Probably want to shasum the output or something just to be sure. If the filestore read verification was broken then this test could pass.

Also, formatting. The closing quotes should line up with the `test_expect_success`

test/sharness/t0271-filestore-utils.sh
kevina: The LC_ALL is to avoid any influence from the locale environment in the sort order.

test/sharness/t0271-filestore-utils.sh
kevina: > Probably want to shasum the output or something just to be sure. If the filestore read verification was broken then this test could pass.

okay

> Also, formatting.

Sorry.  The formatting is done manually.  Will double check.

filestore/util.go
kevina: It might happen, if the key is corrupt, in which case there will also be an error.  The idea is to not abort and be able to move on to the next entry.

core/commands/filestore.go
kevina: @whyrusleeping okay done, let me know if its okay now.

filestore/util.go
kevina: Reusing that method will involve an extra lookup as we already have the value from the iterator.

filestore/util.go
whyrusleeping: Then maybe print something useful like "corrupted key", or maybe error this entry out earlier when its detected the key is bad.

kevina: This is what a listing with a corrupt key will likely look like to stdout:

```
zb2rhaPkR7ZF9BzSC2BfqbcGivi9QMdauermW9YB6NvS7FZMo   10000 somedir/file2 0
zb2rhav4wcdvNXtaKDTWHYAqtUHMEpygT1cxqMsfK7QrDuHxH  262144 somedir/file3 524288
zb2rhbcZ3aUXYcrbhhDH1JyrpDcpdw1KFJ5Xs5covjnvMpxDR    1000 somedir/file1 0
?????????????????????????????????????????????????
zb2rhebtyTTuHKyTbJPnkDUSruU5Uma4DN8t2EkvYZ6fP36mm  262144 somedir/file3 262144
zb2rhm9VTrX2mfatggYUk8mHLz78XBxVUTTzLvM2N3d6frdAU  213568 somedir/file3 786432
```

And to stderr:
```
error: could not decode cid: some-bad-key-string
```

The key may be corrupt, or any number of other things.  I will be happy to replace '?????????????????????????????????????????????????' with whatever you want (as long as it can not also be a valid key) but I don't want to panic as that will create an unnecessary special case.

whyrusleeping: maybe `<corrupt key string>` ?

filestore/util.go
whyrusleeping: Then refactor that method so that it looks up the value and calls some `dataObjFromValue` method

filestore/util.go
kevina: I prefer `?????????????????????????????????????????????????` but okay, I'll change it tomorrow.

filestore/util.go
kevina: Okay, will look into it tomorrow.

mib-kd743naq:
> I prefer `?????????????????????????????????????????????????` ...

@kevina perhaps I am pointing the obvious, but visual alignment would not possible anyway in more complex scenarios where the hashing type / length is not uniform. E.g. cid links to pb content and the like

kevina:
> perhaps I am pointing the obvious, but visual alignment would not possible anyway in more complex scenarios

@mib-kd743naq yeah I know.  I prefer the `???...` because it is non specific, just that there is a problem.  It's not a big deal at all.

filestore/util.go
kevina: Done.

test/sharness/t0271-filestore-utils.sh
Kubuxu: It isn't needed, sharness sets it.

test/sharness/t0271-filestore-utils.sh
```diff
@@ -0,0 +1,195 @@
...
+
+test_filestore_adds
+
+test_filestore_verify
+
+test_filestore_dups_rm
+
+echo "WORKING DIR"
+echo "IPFS PATH = " $IPFS_PATH
+pwd
```
Kubuxu: Debug code?

test/sharness/t0271-filestore-utils.sh
```diff
@@ -0,0 +1,195 @@
...
+
+	test_expect_success "'ipfs filestore verify' shows file as changed" '
+		ipfs filestore verify > verify_actual &&
+		grep changed verify_actual | grep -q somedir/file3
+	'
+}
+
+cat <<EOF > dups_expect
+$FILE1_HASH
+EOF
```
Kubuxu: Is it outside of the test?

test/sharness/t0271-filestore-utils.sh
```diff
@@ -0,0 +1,195 @@
...
+
+test_filestore_verify
+
+test_filestore_dups_rm
+
+echo "WORKING DIR"
+echo "IPFS PATH = " $IPFS_PATH
+pwd
+
+
```
Kubuxu: General comment here, you seem to interleave workspace prep, function definitions and constant definitions, could we clean it up?

core/commands/filestore.go
```diff
@@ -0,0 +1,306 @@
...
+			res.SetError(err, cmds.ErrNormal)
+			return
+		}
+		ch, err := fs.FileManager().AllKeysChan(req.Context())
+		if err != nil {
+			res.SetError(err, cmds.ErrNormal)
+			return
+		}
+
+		out := make(chan interface{})
```
Kubuxu: Probably best to add some buffer.

core/commands/filestore.go
```diff
@@ -0,0 +1,306 @@
...
+
+var rmFileStore = &cmds.Command{
+	Helptext: cmds.HelpText{
+		Tagline: "Remove IPFS block(s) from just the filestore or blockstore.",
+		ShortDescription: `
+Remove blocks from either the filestore or the main blockstore.
+`,
+	},
+	Arguments: []cmds.Argument{
+		cmds.StringArg("hash", true, true, "Bash58 encoded multihash of block(s) to remove."),
```
Kubuxu: not bash but Base58, but as it is CID then it should accept just CID. CID can have base descriptor.

core/commands/root.go
```diff
@@ -124,6 +124,7 @@ var rootSubcommands = map[string]*cmds.Command{
 	"update":    ExternalBinary(),
 	"version":   VersionCmd,
 	"bitswap":   BitswapCmd,
+	"filestore": FileStoreCmd,
```
Kubuxu: It should be mentioned in helptext of main command.

test/sharness/t0271-filestore-utils.sh
kevina: Ask @whyrusleeping I was using `t0270-filestore-utils.sh` as a template.

Kubuxu:
Can you also rebase it to master.

test/sharness/t0271-filestore-utils.sh
kevina: Yes, I didn't think it mattered.  I can move it into a test if it's important.

core/commands/filestore.go
kevina: Oops, will fix.

kevina:
> Can you also rebase it to master.

No, it depends on `feat/filestore0`.

core/commands/root.go
kevina: Oversight.  Will fix.

Kubuxu:
Can I rebase `feat/filestore0` to master?

test/sharness/t0271-filestore-utils.sh
kevina: Yeah I know. I didn't think it really matters.  I will see what I can do.

Kubuxu: It does as it makes harder to understand the tests, which means that in future, when they will have to be changed, already high chance of mistake will be even higher.

kevina:
@whyrusleeping how do you want to handle this p.r.?  If we don't merge this into the other p.r. before it is merged I am guessing I am going to have to close this one and create a new p.r. against master.

whyrusleeping:
@kevina You can actually now change the branch that the PR is against. Github fixed that issue. Once the other one merges we can change this ones base, rebase, and get it merged

kevina:
@whyrusleeping okay than, I'll wait before I do anything.

kevina:
Just rebased.

whyrusleeping:
@kevina cool, will review

whyrusleeping:
Slating for 0.4.8, will review in the next few days

kevina:
Rebased yet again...

core/commands/filestore.go
```diff
@@ -0,0 +1,306 @@
...
+			}
+			return nil, nil
+		},
+	},
+	Type: filestore.ListRes{},
+}
+
+var dupsFileStore = &cmds.Command{
+	Helptext: cmds.HelpText{
+		Tagline: "Print block both in filestore and non-filestore.",
```
whyrusleeping: Change to `"List blocks that are both in the filestore and standard block storage."`

kevina:
Factored out `filestore rm` as requested on IRC.

whyrusleeping:
test failures are unrelated (though still concerning). I've noted both of them and will be investigating soon.

Otherwise, lets get this show on the road! :ship:

PR #3666, opened 2017-02-08
Fix marshaling of null cbor arrays

whyrusleeping:


whyrusleeping:
@Kubuxu squashed, circle is green. :ship: :ship: :ship:

Kubuxu:
And now OSX sharness fails due to bash escaping incompatibility.

whyrusleeping:
@Kubuxu oh damn. I thought the last commit addressed that difference...

PR #3671, opened 2017-02-10
Report progress during `pin add`

### LINKED ISSUE #3624: pin add needs a progress bar
It currently just prints nothing until its done, this can lead people (myself included) into thinking that its frozen.

kevina:
Closes #3624

License: MIT

### DIFF
--- a/core/commands/pin.go
+++ b/core/commands/pin.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"time"
 
 	cmds "github.com/ipfs/go-ipfs/commands"
 	core "github.com/ipfs/go-ipfs/core"
@@ -33,6 +34,11 @@ type PinOutput struct {
 	Pins []*cid.Cid
 }
 
+type AddPinOutput struct {
+	Pins     []*cid.Cid
+	Progress int `json:",omitempty"`
+}
+
 var addPinCmd = &cmds.Command{
 	Helptext: cmds.HelpText{
 		Tagline:          "Pin objects to local storage.",
@@ -44,8 +50,9 @@ var addPinCmd = &cmds.Command{
 	},
 	Options: []cmds.Option{
 		cmds.BoolOption("recursive", "r", "Recursively pin the object linked to by the specified object(s).").Default(true),
+		cmds.BoolOption("progress", "Show progress"),
 	},
-	Type: PinOutput{},
+	Type: AddPinOutput{},
 	Run: func(req cmds.Request, res cmds.Response) {
 		n, err := req.InvocContext().GetNode()
 		if err != nil {
@@ -61,22 +68,88 @@ var addPinCmd = &cmds.Command{
 			res.SetError(err, cmds.ErrNormal)
 			return
 		}
+		showProgress, _, _ := req.Option("progress").Bool()
 
-		added, err := corerepo.Pin(n, req.Context(), req.Arguments(), recursive)
-		if err != nil {
-			res.SetError(err, cmds.ErrNormal)
+		if !showProgress {
+			added, err := corerepo.Pin(n, req.Context(), req.Arguments(), recursive)
+			if err != nil {
+				res.SetError(err, cmds.ErrNormal)
+				return
+			}
+			res.SetOutput(&AddPinOutput{Pins: added})
 			return
 		}
 
-		res.SetOutput(&PinOutput{added})
+		v := new(dag.ProgressTracker)
+		ctx := v.DeriveContext(req.Context())
+
+		ch := make(chan []*cid.Cid)
+		go func() {
+			defer close(ch)
+			added, err := corerepo.Pin(n, ctx, req.Arguments(), recursive)
+			if err != nil {
+				res.SetError(err, cmds.ErrNormal)
+				return
+			}
+			ch <- added
+		}()
+		out := make(chan interface{})
+		res.SetOutput((<-chan interface{})(out))
+		go func() {
+			ticker := time.NewTicker(500 * time.Millisecond)
+			defer ticker.Stop()
+			defer close(out)
+			for {
+				select {
+				case val, ok := <-ch:
+					if !ok {
+						// error already set just return
+						return
+					}
+					if pv := v.Value(); pv != 0 {
+						out <- &AddPinOutput{Progress: v.Value()}
+					}
+					out <- &AddPinOutput{Pins: val}
+					return
+				case <-ticker.C:
+					out <- &AddPinOutput{Progress: v.Value()}
+				case <-ctx.Done():
+					res.SetError(ctx.Err(), cmds.ErrNormal)
+					return
+				}
+			}
+		}()
 	},
 	Marshalers: cmds.MarshalerMap{
 		cmds.Text: func(res cmds.Response) (io.Reader, error) {
-			added, ok := res.Output().(*PinOutput)
-			if !ok {
+			var added []*cid.Cid
+
+			switch out := res.Output().(type) {
+			case *AddPinOutput:
+				added = out.Pins
+			case <-chan interface{}:
+				progressLine := false
+				for r0 := range out {
+					r := r0.(*AddPinOutput)
+					if r.Pins != nil {
+						added = r.Pins
+					} else {
+						if progressLine {
+							fmt.Fprintf(res.Stderr(), "\r")
+						}
+						fmt.Fprintf(res.Stderr(), "Fetched/Processed %d nodes", r.Progress)
+						progressLine = true
+					}
+				}
+				if progressLine {
+					fmt.Fprintf(res.Stderr(), "\n")
+				}
+				if res.Error() != nil {
+					return nil, res.Error()
+				}
+			default:
 				return nil, u.ErrCast()
 			}
-
 			var pintype string
 			rec, found, _ := res.Request().Option("recursive").Bool()
 			if rec || !found {
@@ -86,7 +159,7 @@ var addPinCmd = &cmds.Command{
 			}
 
 			buf := new(bytes.Buffer)
-			for _, k := range added.Pins {
+			for _, k := range added {
 				fmt.Fprintf(buf, "pinned %s %s\n", k, pintype)
 			}
 			return buf, nil
--- a/merkledag/merkledag.go
+++ b/merkledag/merkledag.go
@@ -139,8 +139,21 @@ func (n *dagService) Remove(nd node.Node) error {
 }
 
 // FetchGraph fetches all nodes that are children of the given node
-func FetchGraph(ctx context.Context, c *cid.Cid, serv DAGService) error {
-	return EnumerateChildrenAsync(ctx, serv, c, cid.NewSet().Visit)
+func FetchGraph(ctx context.Context, root *cid.Cid, serv DAGService) error {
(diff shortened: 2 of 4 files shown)


kevina:
@whyrusleeping let me know if this is along the lines of what you are thinking.

Also, how would we test this output, you said to output something every x time.  But that will be hard to test as the amount of time an operation takes is anything but deterministic.

whyrusleeping:
Looking pretty good so far :+1: A couple important things to keep in mind:

- if the progress flag isnt passed, we should keep the same exact behaviour we have now. This means not returning a channel in the default  (non-progress) case.
- The progress tracking stuff should be very easily testable on its own in the merkledag package, just generate some graphs and call enumerateChildren on those graphs with the progress tracker and verify the output looks correct.
- I'm thinking we can be a little more granular than just incrementing by sets of links. Maybe we keep track of two numbers: total nodes processed, and the total number of nodes we think there are (this second number will obviously grow as we go down the graph, but thats fine)

kevina:
> if the progress flag isnt passed, we should keep the same exact behaviour we have now. This means not returning a channel in the default (non-progress) case.

I would think changing the type of the output based on flag would be a bad idea.

> I'm thinking we can be a little more granular than just incrementing by sets of links. Maybe we keep track of two numbers: total nodes processed, and the total number of nodes we think there are (this second number will obviously grow as we go down the graph, but thats fine)

Okay.  To avoid performance problems was trying to avoid updating the number after each call to visit.  However, if  don't think that will be a problem (I am mainly concerned about the locking overhead) I can update the number one node at a time.

I will add a second number for links so far, but nor sure how useful it will be.

whyrusleeping:
> I would think changing the type of the output based on flag would be a bad idea.

The type remains the same, it just enables a streaming mode for progress output.

> To avoid performance problems...

The cost of fetching and unmarshaling each node is going to vastly exceed the tiny cost of the mutex here (its not even a R/W Lock). If it ends up being an issue we could even switch to atomic integer operations

kevina:
@whyrusleeping 

> The type remains the same, it just enables a streaming mode for progress output.

The type of the object return will change from a  `*AddPinOutput` to a `chan <-interface{}` to a depending on if `--progress` was specified.  Is that okay?

> To avoid performance problems...

Okay.

whyrusleeping:
Yeah, changing between a chan and the concrete type there is fine. The type field in the command object will still stay the same

kevina:
@whyrusleeping okay, I increased the granularity and avoided using a channel when `--progress` is not specified.  The type had to change from PinOutput to AddPinOutput (not sure how to avoid it), but the API should in theory have stayed the same if I understand how it work as the `Progress` field should be admitted if it is "empty" i.e. 0.

One question, should we continue to use the Marshaler to report progress or switch to PostRun?

whyrusleeping:
@kevina Does it work correctly implemented in the marshaler?

kevina:
> @kevina Does it work correctly implemented in the marshaler?

@whyrusleeping I'm sorry I couldn't parse that sentence.

kevina:
@whyrusleeping 

> Add a test for the progress tracker in merkledag_test.go with a few different structures. One running on a single node with no children, one running on a fairly large graph, and maybe one running on a linked list of nodes.

I was planning to, but got stuck on the best way to create some random structures.  Any hints on how to do thie?

> Also change the timer to something more like 500ms

How about one second?  The current time was only for manually testing.

whyrusleeping:
@kevina something like this should work: https://ipfs.io/ipfs/QmRsQHn9yV1VgGU3fEkBtpnwjVgKkrRXFKGS2cRoJMyG4m

kevina:
Okay thanks that helped.  I will try to write the tests, and finish this P.R, later today or tomorrow.

kevina:
Okay, assuming tests pass I consider this done.

core/commands/pin.go
```diff
@@ -61,22 +68,84 @@ var addPinCmd = &cmds.Command{
...
+				res.SetError(err, cmds.ErrNormal)
+				return
+			}
+			res.SetOutput(&AddPinOutput{Pins: added})
 			return
 		}
 
-		res.SetOutput(&PinOutput{added})
+		v := new(dag.ProgressTracker)
+		ctx := context.WithValue(req.Context(), "progress", v)
```
whyrusleeping: We could have a helper function on the progress tracker to do this for us nicely:
```
v := new(ProgressTracker)
ctx := v.DeriveContext(req.Context())
```

core/commands/pin.go
```diff
@@ -61,22 +68,84 @@ var addPinCmd = &cmds.Command{
...
+				if !ok {
+					return nil, u.ErrCast()
+				}
+				progressLine := false
+				for r0 := range ch {
+					r := r0.(*AddPinOutput)
+					if r.Progress == -1 || r.Pins != nil {
+						added = r.Pins
+					} else {
+						if progressLine {
```
whyrusleeping: What is the `if progressLine` stuff for?

merkledag/merkledag.go
```diff
@@ -377,18 +377,39 @@ func EnumerateChildren(ctx context.Context, ds LinkService, root *cid.Cid, visit
 	} else if err != nil {
 		return err
 	}
+	v, _ := ctx.Value("progress").(*ProgressTracker)
```
whyrusleeping: If the context does not have a progress tracker in it, I think this may panic (type asserting a nil value?)

kevina: @whyrusleeping I think it is okay: https://play.golang.org/p/xlvE92J8pj

whyrusleeping: ah, cool. Makes sense now that i think about it

core/commands/pin.go
kevina: I use `\r` to replace write over the current line with a new progress line, but I only want to do that if the last line to stderr was a progress line

core/commands/pin.go
```diff
@@ -61,22 +68,84 @@ var addPinCmd = &cmds.Command{
...
+				return
+			}
+			ch <- added
+		}()
+		out := make(chan interface{})
+		res.SetOutput((<-chan interface{})(out))
+		go func() {
+			defer close(out)
+			for {
+				timer := time.NewTimer(500 * time.Millisecond)
```
hsanjuan: Why not using a time,Ticker? Also, you are recreating timers and only the last one will stop so you are accumulating resources without free-ing them... you could use Timer.Reset() at least.

kevina: The timers should be GCed unless I am missing something.  Timer.Stop() is always called.

whyrusleeping: @hsanjuan Good catch.

We should be making this timer *outside* of the for loop.
The way things are, we are making a new timer every loop, and they all tick every 500ms, and then they don't get closed until the goroutine returns.

hsanjuan: > The timers should be GCed
You're right, it's actually tickers not timers, but it's still generating garbage which needs to be GCed when it could easily not do it.

kevina: I honestly don't think it makes any difference. Nevertheless I just pushed a commit to move the creation of the timer to outside the loop at the expense of slightly more code.

whyrusleeping: It makes a *huge* difference. 

Every call to `NewTimer` creates an entry in the runtimes timer heap: https://golang.org/src/time/sleep.go?s=2220:2252#L84
https://golang.org/src/runtime/time.go#L94

They build up and generate events until you close them. Do not make this mistake, it causes weird side effects

core/commands/pin.go
```diff
@@ -61,22 +68,87 @@ var addPinCmd = &cmds.Command{
...
+				res.SetError(err, cmds.ErrNormal)
+				return
+			}
+			ch <- added
+		}()
+		out := make(chan interface{})
+		res.SetOutput((<-chan interface{})(out))
+		go func() {
+			duration := 500 * time.Millisecond
+			timer := time.NewTimer(duration)
```
whyrusleeping: probably makes sense to use `time.Ticker` here so you don't have to reset it.

kevina: @whyrusleeping  Okay done.  Please double check that the ticker will be correctly garbage collected.

core/commands/pin.go
kevina: Dear @whyrusleeping,

I did make a mistake, but the mistake I think you were pointing out:

> They build up and generate events until you close them. Do not make this mistake, it causes weird side effects

And calling stopTimer which I thought I did should removes them from the heap, https://golang.org/src/runtime/time.go#L71.

This mistake I made was assuming that the defer in a loop would fire before the next iteration.  This is how a similar construct would of worked in C++.  It seams I was wrong and go takes a more simple minded approach.  The defer does not fire until the end of the function as demonstrated in this code (https://play.golang.org/p/zKTz1Kc0K_)
```
func main() {
	for i := 0; i < 10; i++ { 
		defer fmt.Println("ByBy")
		println("Hello")
	}
}
```
I personally think go is wrong here, but that is not something I wish to have an extended discussion on.  I will just know to keep in mind how defer works in go and try not to make similar assumptions based on how C++ works, as go is obviously not C++.

whyrusleeping:
This LGTM, Just waiting on a :+1: from @Kubuxu now

Kubuxu:
I am afraid this will break API implementations, someone might not be expecting that he will get objects with Progress instead of Pins field set.

This is why it is important to spec those API as if we stated that you should filter the NDJSON stream for value you need it would be OK, but right now, anything can happen.
---

I will do proper CR tomorrow.

Kubuxu:
Also I am not happy with testing of this code, I know it can be hard but it should be possible.

kevina:
@Kubuxu 

> Also I am not happy with testing of this code

Is that also showing coverage from the sharness tests?  I added some test which should at least cover most of the new code and some of that might now have been covered before.

Kubuxu:
Ahh, it wasn't rebased to master. Master has sharness test coverage collection.

Could you rebase it?

kevina:
> Could you rebase it?

Done.

kevina:
> I am afraid this will break API implementations, someone might not be expecting that he will get objects with Progress instead of Pins field set.

If I did things correctly.  That will only happen if `--progress` is used.  If not the API should in theory be the same, although I have not tested that because I not show how.

kevina:
@Kubuxu the coverage tests are green now.

kevina:
Actually once it rebased progress indicator will no longer work due to the fact that FetchGraph now calls EnumerateChildrenAsync.  I will fix that by moving the progress reporting into the Visit callback and also make always show the progress indicator to make it easier to test for.

Expect something in the next few hours, end of day at the latest.

kevina:
Okay I rebased and made some changes.  You might want to look at the last two commits.  Note the earlier commits are technically non-functional due to the change of FetchGraph now calling EnumerateChildrenASync instead of EnumerateChildren. I can rebase/squash my commits if you think it will be best.

whyrusleeping:
@kevina Yeah, squashing the broken commits would be good. We try hard to make sure every commit builds and works.

@Kubuxu re: breaking api concerns, The default value here is false, so unless the user (or api client) explicitly passes `--progress=true`, then nothing changes for them.

Kubuxu:
Yeah, didn't notice it at first. So LGTM

kevina:
Okay.  Just squashed and rebased.

PR #3674, opened 2017-02-12
gx publish 0.4.5

whyrusleeping:


Kubuxu:
It contains a lot of unnecessary files (mostly in tests).

It is 58 MB, my clean directory is 3MB.

PR #3675, opened 2017-02-12
gx publish 0.4.5, cleaner

whyrusleeping:
The old hash had a bunch of test stuff that was hidden in it.


whyrusleeping:
oh my god gitcop go home youre drunk

PR #3678, opened 2017-02-12
Revert "Merge pull request #3504 from ipfs/feat/makefile/refactor"

whyrusleeping:
This reverts commit e1619bf3025c26a60c95ed4fce0331bc97492ffc, reversing
changes made to 14d518683cc5931c752abeac0ee098928ed4dab0.

Changes caused test failures and windows build process failures


Kubuxu:
I think that travis fail was random, we should instead improve debug output when it happens.

As to Windows, I would work on it, but nobody reported any problems, I am positive I will be able to resolve the issues quite quickly once I get more info.

djdv:
>nobody reported any problems

I need a way to subtly hint at needing a Windows CI service again while knowing it will cause headaches.

mateon1:
I'm pretty sure the only reason it didn't fail on linux hosts in testing, was that it's rather rare for a directory with spaces to be on $PATH

Kubuxu:
@djdv we have base for it, we will try to do in sometime soon.

@mateon1 I have probably tested this case before I introduced that PATH workaround. We should probably somehow test our Makefiles but it might be easier said than done.

whyrusleeping:
Thanks for fixing @Kubuxu

Issue #3676, opened 2017-02-12
Building with make on Windows


djdv:
#### Version information: 
N/A
#### Type: 
Bug
#### Priority: 
P3
#### Description:
This pull has broken building on Windows for me.
https://github.com/ipfs/go-ipfs/pull/3504
I haven't investigated it yet but it's spitting out this
```
C:\tmp\src\github.com\ipfs\go-ipfs>make install
/bin/sh: Files/Python: No such file or directory
bin/check_go_version 1.7
/bin/sh: Files/Python: No such file or directory
make: *** [mk/golang.mk:45: check_go_version] Error 127
```
I've been using msys2's version of GNU `make` to build the same way for a while via 
```
cd %GOPATH%\src\github.com\ipfs\go-ipfs
gx i --global
make install
```
Reverting to a commit before that one works fine with `make`. I can install manually with `go install` still.

![make issue 2](https://cloud.githubusercontent.com/assets/13862850/22859394/607ef712-f0a7-11e6-972b-1746034f79b8.png)

whyrusleeping:
@djdv Thanks for reporting, very quick turnaround on the issue reporting :)

@Kubuxu We've got some issues here.

Kubuxu:
Can you run with `make --trace build` and paste the output? Also try installing Python as it seems to request it, no idea why, I don't use it.

mateon1:
@Kubuxu It's a quoting issue, involving (at least) $PATH

For me it's
```
/bin/sh: Files/ImageMagick-7.0.4-Q16:/c/Program: No such file or directory
/bin/sh: Files/ImageMagick-7.0.4-Q16:/c/Program: No such file or directory
make: *** [Rules.mk:110: help] Error 127
```
which is $PATH from the first space, to the second.

djdv:
It seems to think `check_go_version` doesn't exist, but it does. Also Python is installed on this system, I was wondering why it's asking for that myself.

```
C:\tmp\src\github.com\ipfs\go-ipfs>make build
/bin/sh: Files/Python: No such file or directory
bin/check_go_version 1.7
/bin/sh: Files/Python: No such file or directory
make: *** [mk/golang.mk:45: check_go_version] Error 127

C:\tmp\src\github.com\ipfs\go-ipfs>make --trace build
/bin/sh: Files/Python: No such file or directory
mk/golang.mk:45: target 'check_go_version' does not exist
bin/check_go_version 1.7
[...]

PR #3680, opened 2017-02-12
make: fix PATH escaping

Kubuxu:
Fixes https://github.com/ipfs/go-ipfs/issues/3676

License: MIT


whyrusleeping:
@mateon1 and @djdv I'm assuming your thumbs up here means you've tested this and it works :)

PR #3683, opened 2017-02-12
misc: increase version number to 0.4.6-dev

Kubuxu:
License: MIT


Kubuxu:
This should be part of automated setup for doing releases.

whyrusleeping:
Thats what i get for not following my checklist through to the end

PR #3685, opened 2017-02-12
Dockerfile: Deprecate implicit daemon argument

kpcyrd:
After the discussion in https://github.com/ipfs/go-ipfs/pull/3573 this patch prints a deprecation warning if:

1) the image has been executed with additional arguments
2) the first argument isn't daemon

This way people are able to migrate to the new syntax without any breaking changes.

License: MIT

### DIFF
--- a/bin/container_daemon
+++ b/bin/container_daemon
@@ -32,7 +32,9 @@ else
   # when overwriting CMD, making this PR safe to merge
   echo "DEPRECATED: arguments have been set but the first argument isn't 'daemon'" >&2
   echo "DEPRECATED: run 'docker run ipfs/go-ipfs daemon $@' instead" >&2
-  echo "DEPRECATED: see https://github.com/ipfs/go-ipfs/pull/3573 for more information" >&2
+  echo "DEPRECATED: see the following PRs for more information:" >&2
+  echo "DEPRECATED: * https://github.com/ipfs/go-ipfs/pull/3573" >&2
+  echo "DEPRECATED: * https://github.com/ipfs/go-ipfs/pull/3685" >&2
 fi
 
 exec ipfs daemon "$@"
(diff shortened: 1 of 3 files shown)


bin/container_daemon
```diff
@@ -19,4 +19,22 @@ else
...
+else
+  # print deprecation warning
+  # go-ipfs used to hardcode "ipfs daemon" in it's entrypoint
+  # this workaround supports the new syntax so people start setting daemon explicitly
+  # when overwriting CMD, making this PR safe to merge
+  echo "DEPRECATED: arguments have been set but the first argument isn't 'daemon'" >&2
+  echo "DEPRECATED: run 'docker run ipfs/go-ipfs daemon $@' instead" >&2
+  echo "DEPRECATED: see the following PRs for more information:" >&2
+  echo "DEPRECATED: * https://github.com/ipfs/go-ipfs/pull/3573" >&2
+  echo "DEPRECATED: * https://github.com/ipfs/go-ipfs/pull/3685" >&2
```
Kubuxu: What would be nice is to have something like `tracking pixel` to know when people made transition but some might deem it to intrusive.

lgierth:
This LGTM, thanks! :+1: 

Apart from the daemon argument, this also makes `--migrate=true` the default, which is reasonable. The default arguments of the container should not lead to a situation where it hangs waiting for stdin (y/n).

kpcyrd:
> Apart from the daemon argument, this also makes --migrate=true the default, which is reasonable. The default arguments of the container should not lead to a situation where it hangs waiting for stdin (y/n).

yup, sorry for the unrelated changes, but that's exactly what happened for me :)

LucasDvp:
what if I want run `ipfs bootstrap rm --all` before `ipfs daemon`

kpcyrd:
@LucasDvp this patch was preparing a setup that would support something like this.

I've filed https://github.com/ipfs/go-ipfs/pull/6188

After that patch has been merged you can do:
```
docker run --rm -v /some/folder:/data ipfs/go-ipfs bootstrap rm --all
docker run --rm -v /some/folder:/data ipfs/go-ipfs
```

PR #3690, opened 2017-02-14
Fuse dockerfile

20zinnm:
Add a Dockerfile with FUSE enabled--good for applications like crawlers, etc, that could just use the file system rather than directly integrating with IPFS.

### DIFF
--- a/Dockerfile.fuse
+++ b/Dockerfile.fuse
@@ -1,5 +1,9 @@
 FROM ipfs/go-ipfs:latest
 
+# This a Docker image based on /Dockerfile with FUSE mounts pre-enabled.
+#
+# This should be run with '--cap-add SYS_ADMIN --device /dev/fuse'.
+
 USER root
 
 RUN apk add fuse shadow --update-cache --repository http://nl.alpinelinux.org/alpine/edge/testing \


whyrusleeping:
This will need to be reviewed by someone who understands dockerfiles well. cc @lgierth @hsanjuan @VictorBjelkholm

20zinnm:
@whyrusleeping You can test it real quick by running `docker build -t go-ipfs-fuse . -f Dockerfile.fuse`, which will build the Docker image, and then running `docker run go-ipfs-fuse --name [name of container]`. That will start a container based on that image and show you the progress. Finally, once it looks done you can run `docker exec -it [name of container] bash` to get access to the container, where you should notice that `cat /ipfs/[hash]` acts like `ipfs get /ipfs/[hash]` and likewise for IPNS.

whyrusleeping:
@20zinnm Can you fix your commit sign offs here?

Kubuxu:
The fork no longer exists which screws up Jenkins. Please reopen with fixed sign-offs.

If author doesn't respond in a week, anyone please feel free to recreate functionality of this PR.

PR #3692, opened 2017-02-15
Make bin/dist_get fall back to other downloaders if one fails

mateon1:
Also refactored the code a bit, and added httpie.
Note: I used redirection for httpie, as `http -d url -o output` complained its output was redirected.

### DIFF
--- a/bin/dist_get
+++ b/bin/dist_get
@@ -13,6 +13,26 @@ check_writeable() {
 	printf "" > "$1" && rm "$1"
 }
 
+try_download() {
+	url="$1"
+	output="$2"
+	command="$3"
+	util_name=$(set -- $3; echo $1)
+
+	if ! have_binary $util_name; then
+		return 1
+	fi
+
+	printf '==> Using %s to download "%s" to "%s"\n' "$util_name" "$url" "$output"
+	if eval "$command"; then
+		echo "==> Download complete!"
+		return
+	else
+		echo "error: couldn't download with $util_name ($?)"
+		return 1
+	fi
+}
+
 download() {
 	dl_url="$1"
 	dl_output="$2"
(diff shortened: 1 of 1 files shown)


bin/dist_get
```diff
@@ -13,6 +13,26 @@ check_writeable() {
 	printf "" > "$1" && rm "$1"
 }
 
+try_download() {
+	url="$1"
+	output="$2"
+	command="$3"
+	util_name=$(set -- $3; echo $1)
```
whyrusleeping: What is this doing?

whyrusleeping:
Very weird circleCI failure... Worth investigating

bin/dist_get
mateon1: This sets arguments within this shell substitution from `$3`, allowing us to get the first word of the command (wget, curl, http) with `$1`.
Arguments outside this shell substitution are not clobbered.

Kubuxu: This might be reason it fails on circle, or other bashizm. CircleCI uses fsh as sh bin. You probably should change the hashbang to bash or remove the bashizm (preferably the latter)

Kubuxu:
List of issues found by shellcheck, some of the might not be because of this PR.

```

In dist_get line 9:
	type "$1" > /dev/null 2> /dev/null
        ^-- SC2039: In POSIX sh, 'type' is undefined.


In dist_get line 20:
	util_name=$(set -- $3; echo $1)
                           ^-- SC2086: Double quote to prevent globbing and word splitting.
                                    ^-- SC2086: Double quote to prevent globbing and word splitting.


In dist_get line 22:
	if ! have_binary $util_name; then
                         ^-- SC2086: Double quote to prevent globbing and word splitting.


In dist_get line 40:
	test "$#" -eq "2" || die "download requires exactly two arguments, was given $@"
                                                                                     ^-- SC2145: Argument mixes string and array. Use * or separate argument.


In dist_get line 68:
				cat "$ua_infile" | tar -O -z -x -f - "$ua_distname/$ua_distname" > "$ua_outfile"
                                    ^-- SC2002: Useless cat. Consider 'cmd < file | ..' or 'cmd file | ..' instead.


In dist_get line 158:
if [ $? -ne 0 ]; then
     ^-- SC2181: Check exit code directly with e.g. 'if mycmd;', not indirectly with $?.


In dist_get line 163:
if [ $? -ne 0 ]; then
     ^-- SC2181: Check exit code directly with e.g. 'if mycmd;', not indirectly with $?.
```

Kubuxu:
You can try running Bash in a mode closer to POSIX with `--posix` flag.

The bashizm is only my guess why things failed, I am not sure.

mateon1:
@Kubuxu That is odd, as the `set --` command seems to work for me in both `sh` and `bash --posix`.
I am going to push some changes that fix shellcheck errors, at least those caused by me.

PR #3691, opened 2017-02-15
Add tests for recursively pinning a dag (i.e. large file).

kevina:
Test for already fixed #3688 

License: MIT
Signed-off-by: Kevin Atkinson <k@kevina.org>

PR #3693, opened 2017-02-15
Add back removed test.

kevina:
License: MIT


kevina:
Okay, this is embarrassing.  I did accidentally remove a test in #3691, sorry.  That is what happens when I am a little two quick to push a quick fix.

PR #3697, opened 2017-02-16
PNet take 2

Kubuxu:


core/commands/swarm.go
```diff
@@ -14,7 +14,7 @@ import (
 	"github.com/ipfs/go-ipfs/repo/fsrepo"
 	iaddr "github.com/ipfs/go-ipfs/thirdparty/ipfsaddr"
 	pstore "gx/ipfs/QmQMQ2RUjnaEEX8ybmrhuFFGhAwPjyL1Eo6ZoJGD7aAccM/go-libp2p-peerstore"
-	swarm "gx/ipfs/QmY8hduizbuACvYmL4aZQbpFeKhEQJ1Nom2jY6kv6rL8Gf/go-libp2p-swarm"
```
whyrusleeping: lets keep all the dep updates in the same commit

Kubuxu:
Done.

core/core.go
```diff
@@ -174,7 +176,25 @@ func (n *IpfsNode) startOnlineServices(ctx context.Context, routingOption Routin
 
 	tpt := makeSmuxTransport(mplex)
 
+<<<<<<< HEAD
```
whyrusleeping: Merge conflict here.

core/core.go
```diff
@@ -190,6 +190,24 @@ func (n *IpfsNode) startOnlineServices(ctx context.Context, routingOption Routin
...
 		n.PNetFingerpint = protec.Fingerprint()
+		go func() {
+			t := time.NewTicker(30 * time.Second)
+			<-t.C // swallow one tick
+			for {
+				select {
+				case <-t.C:
+					if ph := n.PeerHost; ph != nil {
+						if len(ph.Network().Peers()) == 0 {
+							log.Warning("We are in private network and have no peers.")
```
whyrusleeping: Rather unfortunate that these warnings won't show up by default... I wonder if we should look at turning on warning messages by default?

core/core.go
Kubuxu: Sorry that I have missed it.

core/core.go
Kubuxu: I take it back, there is a lot of spam on global warning channel.

Kubuxu:
Fixed the merge conflict and rebased.

Kubuxu:
Coverage is a bit low, as the lack of peer warning isn't tested. There is no clean way to test it and as it is just a log I don't think it matters.

core/core.go
whyrusleeping: Hrm... definitely out of scope for this PR, but we should look at reducing spam there and think about setting the default level to be warning

cmd/ipfs/daemon.go
```diff
@@ -347,6 +348,11 @@ func daemonFunc(req cmds.Request, res cmds.Response) {
 	}
 	node.SetLocal(false)
 
+	if node.PNetFingerpint != nil {
+		fmt.Println("Swarm is limited to private network of peers with the swarm key")
+		fmt.Printf("Swarm key fingerprint: %s\n", hex.EncodeToString(node.PNetFingerpint))
```
whyrusleeping: you can just use `%x`

core/core.go
```diff
@@ -671,15 +708,15 @@ func listenAddresses(cfg *config.Config) ([]ma.Multiaddr, error) {
...
 }
 
-type HostOption func(ctx context.Context, id peer.ID, ps pstore.Peerstore, bwr metrics.Reporter, fs []*net.IPNet, tpt smux.Transport) (p2phost.Host, error)
+type HostOption func(ctx context.Context, id peer.ID, ps pstore.Peerstore, bwr metrics.Reporter, fs []*net.IPNet, tpt smux.Transport, protc ipnet.Protector) (p2phost.Host, error)
 
 var DefaultHostOption HostOption = constructPeerHost
 
 // isolates the complex initialization steps
-func constructPeerHost(ctx context.Context, id peer.ID, ps pstore.Peerstore, bwr metrics.Reporter, fs []*net.IPNet, tpt smux.Transport) (p2phost.Host, error) {
+func constructPeerHost(ctx context.Context, id peer.ID, ps pstore.Peerstore, bwr metrics.Reporter, fs []*net.IPNet, tpt smux.Transport, protec ipnet.Protector) (p2phost.Host, error) {
```
whyrusleeping: At some point (not now) I want to make this more simply composed

whyrusleeping:
LGTM, lets ship this in 0.4.7

core/core.go
Kubuxu: Yeah, we should probably change it to be struct instead of function args.

PR #3698, opened 2017-02-16
WIP: fsrepo: fix musl detection for migrations

lgierth:
Note: the tests here are totally incomplete, I keep getting stuck :/

If someone else wants to take a look today, that'd be super helpful. The migrations code itself should be good.

------

The ldd command used for detection doesn't seem to have a
--version flag on Alpine Linux. It would print the expected
output, but instead of stdout, it would print it on stderr.

The musl detection code would only scan stdout for mentions
of "musl", and would thus *not* download the musl version
of the fs-repo-migrations executable.

This manifested in the well-known "fs-repo-migrations: not found"
error, which you get when executing something that was linked
against a different libc than the one present on the system.

License: MIT

### DIFF
--- a/repo/fsrepo/migrations/migrations.go
+++ b/repo/fsrepo/migrations/migrations.go
@@ -243,27 +243,30 @@ func GetBinaryForVersion(distname, binnom, root, vers, out string) error {
 	return unpackArchive(distname, binnom, arcpath, out, archive)
 }
 
+// osWithVariant returns the OS name with optional variant.
+// Currently returns either runtime.GOOS, or "linux-musl".
 func osWithVariant() (string, error) {
 	if runtime.GOOS != "linux" {
 		return runtime.GOOS, nil
 	}
 
-	bin, err := exec.LookPath(filepath.Base(os.Args[0]))
-	if err != nil {
-		return "", fmt.Errorf("failed to resolve go-ipfs: %s", err)
-	}
-
-	// ldd outputs the system's kind of libc
+	// ldd outputs the system's kind of libc.
 	// - on standard ubuntu: ldd (Ubuntu GLIBC 2.23-0ubuntu5) 2.23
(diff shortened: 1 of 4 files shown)


repo/fsrepo/migrations/migrations.go
```diff
@@ -243,27 +243,30 @@ func GetBinaryForVersion(distname, binnom, root, vers, out string) error {
...
 
-	bin, err := exec.LookPath(filepath.Base(os.Args[0]))
+	// ldd outputs the system's kind of libc.
+	// - on standard ubuntu: ldd (Ubuntu GLIBC 2.23-0ubuntu5) 2.23
+	// - on alpine: musl libc (x86_64)
+	//
+	// we use the combined stdout+stderr,
+	// because ldd --version prints differently on different OSes.
+	// - on standard ubuntu: stdout
+	// - on alpine: stderr (it probably doesn't know the --version flag)
```
Kubuxu: Ubuntu doesn't know --version flag, alpine just prints it on stderr.

test/sharness/t0301-docker-migrate.sh
```diff
@@ -0,0 +1,72 @@
...
+	echo "v1.1.1" >> vers_resp
+'
+
+pretend_server() {
+	cat vers_resp | nc -l -i 1 -p 17233
+}
+
+test_expect_success "startup fake dists server" '
+	pretend_server > dist_serv_out &
+	echo $! > netcat_pid
```
Kubuxu: We should have a setp with `kill $(cat netcat_pid) || true`

whyrusleeping: :+1: Thanks! forgot that

lgierth:
Thanks @whyrusleeping <3 LGTM with kill suggested by kuba

Kubuxu:
Rebsed to master.

PR #3700, opened 2017-02-16
Refactor EnumerateChildren to avoid need for bestEffort paramater.

kevina:
License: MIT

### DIFF
--- a/merkledag/merkledag.go
+++ b/merkledag/merkledag.go
@@ -36,8 +36,10 @@ type DAGService interface {
 }
 
 type LinkService interface {
-	// Return all links for a node, may be more effect than
-	// calling Get in DAGService
+	// GetLinks return all links for a node.  The complete node does not
+	// necessarily have to exist locally, or at all.  For example, raw
+	// leaves cannot possibly have links so there is no need to look
+	// at the node.
 	GetLinks(context.Context, *cid.Cid) ([]*node.Link, error)
 
 	GetOfflineLinkService() LinkService
@@ -114,6 +116,8 @@ func decodeBlock(b blocks.Block) (node.Node, error) {
 	}
 }
 
+// GetLinks return the links for the node, the node doesn't necessarily have
+// to exist locally.
 func (n *dagService) GetLinks(ctx context.Context, c *cid.Cid) ([]*node.Link, error) {
 	if c.Type() == cid.Raw {
 		return nil, nil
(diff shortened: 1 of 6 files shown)


kevina:
WIP.  A test is failing locally for me and I want to see if it okay on the build servers.

kevina:
@whyrusleeping okay the tests are passing.  Let me know if you are like this approach.  If you do I will refactor the Async version in a similar fashion.

kevina:
> Lets make sure not to change the function parameter ordering. the new function parameter should take the place of the linkservice

I only did that because I thought the new order made more sense.  Is there a reason you don't like it?  Just curious.  I have no problem changing them back.

whyrusleeping:
@kevina Eh, I think it makes more sense to have the fetcher first, as its almost as if the cid could be an argument to a method on the 'fetcher function'.

kevina:
> @kevina Eh, I think it makes more sense to have the fetcher first, as its almost as if the cid could be an argument to a method on the 'fetcher function'.

Okay, will keep the order the same.  Will also see what I can do to refactor the Async version, although that will be a little more involved.

merkledag/merkledag.go
```diff
@@ -36,8 +36,10 @@ type DAGService interface {
 }
 
 type LinkService interface {
-	// Return all links for a node, may be more effect than
-	// calling Get in DAGService
+	// Return all links for a node.  The complete node does not
+	// necessarily have to exist locally, or at all.  For example, raw
+	// leaves can not possible have links so there is no need to look
```
whyrusleeping: typo: cannot possibly

merkledag/merkledag.go
```diff
@@ -114,6 +116,8 @@ func decodeBlock(b blocks.Block) (node.Node, error) {
 	}
 }
 
+// GetLinks return the links for the node, the node doesn't necessary
```
whyrusleeping: typo: necessarily

merkledag/merkledag.go
```diff
@@ -138,9 +142,22 @@ func (n *dagService) Remove(nd node.Node) error {
 	return n.Blocks.DeleteBlock(nd)
 }
 
+// Create a functtion to get the links for a node, from the node,
```
whyrusleeping: typo: function, guaranteed

Kubuxu: Also go style of docs says that you should start doc comment with the name of the function.

kevina:
@Kubuxu I fixed my comments to include the start with the name of the function

@whyrusleeping I corrected the typos you pointed out.

whyrusleeping:
Alright cool, this is ready to go in 0.4.7

whyrusleeping:
@kevina got a merge conflict here. Mind rebasing?

kevina:
@whyrusleeping sorry, missed this.  Just rebased and things should be good to go.

whyrusleeping:
Thanks @kevina !

PR #3701, opened 2017-02-17
print out all daemon stderr files when iptb stop fails

whyrusleeping:
License: MIT

### DIFF
--- a/test/sharness/t0130-multinode.sh
+++ b/test/sharness/t0130-multinode.sh
@@ -79,7 +79,8 @@ run_advanced_test() {
 	run_random_dir_test
 
 	test_expect_success "shut down nodes" '
-		iptb stop
+		iptb stop ||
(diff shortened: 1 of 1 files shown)


test/sharness/t0130-multinode.sh
```diff
@@ -79,7 +79,8 @@ run_advanced_test() {
 	run_random_dir_test
 
 	test_expect_success "shut down nodes" '
-		iptb stop
+		iptb stop ||
+			test_fsh cat .iptb/*/daemon.stderr
```
Kubuxu: In my experience: `tail -n +1` is better than cat as it also shows filenames.

I would go with `tail -n +1 .iptb/*/daemon.std*`

whyrusleeping: :+1: cool, thanks :)

PR #3702, opened 2017-02-17
update dists path with 1.2.2 migrations

whyrusleeping:
License: MIT

### DIFF
--- a/repo/fsrepo/migrations/migrations.go
+++ b/repo/fsrepo/migrations/migrations.go
@@ -15,7 +15,7 @@ import (
 	"strings"
 )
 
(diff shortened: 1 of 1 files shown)


Kubuxu:
We might have one more dist fix as it seems that files aren't closed properly in time in the migration code in go-ds-flatfs.

whyrusleeping:
@Kubuxu link to issue?

whyrusleeping:
Got a confirmation from @lgierth that these migrations binaries work for upgrading the docker node gateways.

whyrusleeping:
@Kubuxu :+1:  ?

PR #3703, opened 2017-02-17
Fix formatting in merkledag.go

Kubuxu:
License: MIT


Kubuxu:
It is quite important as it is one of tests on Jenkins.

kevina:
Note.  I have two p.r. that touch `merkledag.go` #3671 and #3700.  They both fix the formatting problems.  If either of them are to go in within the next couple of days It would save me some trouble if this P.R. is closed without merging.

whyrusleeping:
@Kubuxu how did this get through in the first place? I thought gofmt was part of our normal test suite?

Kubuxu:
It was on teamcity, and is on Jenkins. Jenkins is almost working.

PR #3708, opened 2017-02-20
block rm: use Marshalers instead of PostRun to process output

kevina:
License: MIT


kevina:
@whyrusleeping When I did a similar think with the filestore utils you thought it would be better to use a Marshalers: https://github.com/ipfs/go-ipfs/pull/3653#discussion_r99644379.

PR #3709, opened 2017-02-20
block rm: change "force" option to "ignore"

kevina:
"ignore" just seams like a more appropriate name as nothing is being
"forced"

License: MIT

### DIFF
--- a/blocks/blockstore/util/remove.go
+++ b/blocks/blockstore/util/remove.go
@@ -24,7 +24,7 @@ type RemovedBlock struct {
 type RmBlocksOpts struct {
 	Prefix string
 	Quiet  bool
-	Force  bool
+	Ignore bool
 }
 
 func RmBlocks(blocks bs.GCBlockstore, pins pin.Pinner, cids []*cid.Cid, opts RmBlocksOpts) (<-chan interface{}, error) {
@@ -41,7 +41,7 @@ func RmBlocks(blocks bs.GCBlockstore, pins pin.Pinner, cids []*cid.Cid, opts RmB
 
 		for _, c := range stillOkay {
 			err := blocks.DeleteBlock(c)
-			if err != nil && opts.Force && (err == bs.ErrNotFound || err == ds.ErrNotFound) {
+			if err != nil && opts.Ignore && (err == bs.ErrNotFound || err == ds.ErrNotFound) {
 				// ignore non-existent blocks
 			} else if err != nil {
 				out <- &RemovedBlock{Hash: c.String(), Error: err.Error()}
--- a/core/commands/block.go
+++ b/core/commands/block.go
@@ -248,7 +248,7 @@ It takes a list of base58 encoded multihashs to remove.
 		cmds.StringArg("hash", true, true, "Bash58 encoded multihash of block(s) to remove."),
 	},
 	Options: []cmds.Option{
-		cmds.BoolOption("force", "f", "Ignore nonexistent blocks.").Default(false),
+		cmds.BoolOption("ignore", "i", "Ignore nonexistent blocks.").Default(false),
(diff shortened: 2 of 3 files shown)


kevina:
In the original pull request for `block rm` @jbenet suggested `block rm -f` as a way to be silent about deleting non-existent blocks (see https://github.com/ipfs/go-ipfs/pull/2962#issuecomment-240302436).  I take he was modeling that after the `rm` filestore utility.   I want along with it, but never really liked that option name.  This, p.r. changes the name to `-i` or `--ignore`.  I think this is a better name because, unlike the unix `rm` command. the `-f` option does not force anything instead it "ignores" non-exist blocks rather than outputting an error.

@whyrusleeping thoughts?

whyrusleeping:
I think that force is still the right wording for it. cc @jbenet @Kubuxu

whyrusleeping:
'force' is the correct term here i think, we are forcing the removal of a block that we could otherwise not remove

kevina:
@whyrusleeping 

> 'force' is the correct term here i think, we are forcing the removal of a block that we could otherwise not remove

This is incorrect, the block would of been removed anyway.  All `--force` does is silence the error message.

If you still don't agree than I am fine with it, but I want to be clear on that point.

kevina:
[note, deleted last comment since it was inaccurate and it was better to start over]

@whyrusleeping
> 'force' is the correct term here i think, we are forcing the removal of a block that we could otherwise not remove

That is incorrect we are not forcing anything.  All `--force` does is silence an error message when removing non-existent blocks.  The end-result with and without the `--force` option is the same except for the error message and the exit code.

Reopening since i think there is confusion on what the `--force` option actually does.  If we are clear on what it does and you still like `--force` better then I am fine with that and we can close this issue.

whyrusleeping:
@kevina this is identical to unix rm in this case:
```
[whyrusleeping@rumil ~]$ rm foobar
rm: cannot remove 'foobar': No such file or directory
[whyrusleeping@rumil ~]$ echo $?
1
[whyrusleeping@rumil ~]$ rm -f foobar
[whyrusleeping@rumil ~]$ echo $?
0
```

kevina:
Correct.  Force also forces the removal of write protected files, this does not.

With that understanding if you still prefer `--force` go ahead and close this, just seams an odd name to me, that's all.

whyrusleeping:
Thats true... Its not 'deleting a pinned block anyways'.

I'm just not sold on ignore... @Kubuxu @lgierth thoughts?

lgierth:
:-1: Not a good enough reason to break existing interfaces. To us it looks like a simple rename, to a program the -f flag is just gone.

But I'm wondering why this has to fail on non-existent blocks in the first place. It could be properly idempotent. The result is the same (block non-existent), and if you *need* to know the block's state there's `ipfs block stat`. Unix's rm command can fail in various ways that you can --force your way around. block rm doesn't really have any failure modes that you can --force away.

lgierth:
And similarly, `ipfs add` doesn't fail if the stuff is already stored -- it's idempotent.

lgierth:
The only forceable thing I can imagine right now is rm'ing pinned blocks, but I don't think it matters much. We should never ever touch stuff that is pinned. Suddenly pinning would have to deal with blocks that might disappear, and that'd get very messy.

kevina:
@lgierth the idempotent issue was discussed in #2962 and I agreed to use `-f` or `-force` for the idempotent behavior.  I didn't think it was a very good name, but I didn't want to press the issue then.  The p.r./issue exists because I hate to be suck with a bad name.

whyrusleeping:
I think even though ignore may be a marginally better name, we've already named it `--force` and breaking interfaces (even though nobody is probably using them yet) is generally frowned upon.

I'm gonna go ahead and close this

PR #3712, opened 2017-02-20
More Robust GC

kevina:


kevina:
@whyrusleeping this is very much a work in progress, just want some early feedback.

kevina:
This is based on #3700 use this link for a better compassion until that is merged: https://github.com/ipfs/go-ipfs/compare/kevina/enumerate-children-refactor...kevina/more-robust-gc

kevina:
This took longer than I would of like and I am not sure I am happy with the approach but it should work.  Everything is tested, but the test could also use some improvements in order to document why what they are testing is important.

pin/gc/gc.go
```diff
@@ -22,58 +25,73 @@ var log = logging.Logger("gc")
...
+						break loop
 					}
 				}
 			case <-ctx.Done():
-				return
+				break loop
 			}
 		}
+		if errors {
+			errOutput <- CouldNotDeleteSomeBlocksError
```
whyrusleeping: `ErrCouldNotDeleteSomeBlocks`

pin/gc/gc.go
```diff
@@ -22,58 +25,73 @@ var log = logging.Logger("gc")
 //
 // The routine then iterates over every block in the blockstore and
 // deletes any block that is not found in the marked set.
-func GC(ctx context.Context, bs bstore.GCBlockstore, ls dag.LinkService, pn pin.Pinner, bestEffortRoots []*cid.Cid) (<-chan *cid.Cid, error) {
+//
+func GC(ctx context.Context, bs bstore.GCBlockstore, ls dag.LinkService, pn pin.Pinner, bestEffortRoots []*cid.Cid) (<-chan *cid.Cid, <-chan error) {
```
whyrusleeping: I think  a cleaner return value would be to return a channel of 'results'. Where a result would be defined as something like:
```
type Result struct {
    Cid *cid.Cid
    Err error
}
```

Its always simpler to just have to deal with a single channel in cases like these, it can cause tricky race conditions (though the race conditions probably arent an issue here).

pin/gc/gc.go
```diff
@@ -82,28 +100,74 @@ func Descendants(ctx context.Context, ls dag.LinkService, set *cid.Set, roots []
...
+	}
+
+	if errors {
+		return nil, CoundNotFetchAllLinksError
+	} else {
+		return gcs, nil
 	}
+}
+
+var CoundNotFetchAllLinksError = errors.New("garbage collection aborted: could not retrieve some links")
```
whyrusleeping: make sure errors like this have an `Err` prefix

test/sharness/t0087-repo-robust-gc.sh
```diff
@@ -0,0 +1,143 @@
+#!/bin/sh
+#
+# Copyright (c) 2016 Jeromy Johnson
+# MIT Licensed; see the LICENSE file in this repository.
+#
+
+test_description="Test ipfs pinning operations"
+
+. lib/test-lib.sh
+set -e
```
whyrusleeping: Don't use set -e, sharness handles these sorts of things

whyrusleeping: (you can run the tests with -i to stop when a test fails)

test/sharness/t0087-repo-robust-gc.sh
```diff
@@ -0,0 +1,143 @@
+#!/bin/sh
+#
+# Copyright (c) 2016 Jeromy Johnson
+# MIT Licensed; see the LICENSE file in this repository.
+#
+
+test_description="Test ipfs pinning operations"
```
whyrusleeping: probably want to update the description

test/sharness/t0087-repo-robust-gc.sh
```diff
@@ -0,0 +1,143 @@
...
+	LEAF1FILE=.ipfs/blocks/C2/AFKREIBKRCW7HF6NHR6DVWECQXC5RQC7U7PKHKTI53BYYZNQP23DK5FC2Y.data
+
+	LEAF2=zb2rhdUpCuyoQPRtvBW8JhAEfPaNd9dBb3QQCMMWYS1V8hjBR
+	LEAF2FILE=.ipfs/blocks/Q6/BAFKREIDFSUIR43GJPHNDXXQA45GJVNRZBET3CRPUMYJCBLK3RTN7ZAMQ6Q
+
+	LEAF3=zb2rhnwyvreekPqEMAbKqs35KYHxkNi7pXED7L2TfJKveTTva
+	LEAF4=zb2rhnvAVvfDDnndcfQkwqNgUm94ba3zBSXyJKCfVXwU4FXx
+
+	test_expect_success "remove a leaf node from the repo manually" '
+		rm $LEAF1FILE
```
whyrusleeping: quotes around the variable, just in case it changes

test/sharness/t0087-repo-robust-gc.sh
```diff
@@ -0,0 +1,143 @@
...
+		ipfs pin rm $HASH1
+	'
+
+	# make sure the permission problem is fixed on exit, otherwise cleanup
+	# will fail
+	trap "chmod 700 `dirname $LEAF2FILE` 2> /dev/null" 0
+
+	test_expect_success "create a permission problem" '
+		chmod 500 `dirname $LEAF2FILE` &&
+		test_must_fail ipfs block rm $LEAF2 2>&1 | tee err &&
```
whyrusleeping: probably fine just to redirect this to 'err'. May also want to name 'err' something a tiny bit more descriptive like 'block_rm_err'

test/sharness/t0087-repo-robust-gc.sh
kevina: Yeah :)

test/sharness/t0087-repo-robust-gc.sh
kevina: I can remove it if you want, but note it does serve a very useful purpose.  Please see #3721.  I already know about `-i` and that is _not_ why i am using it.

test/sharness/t0087-repo-robust-gc.sh
kevina: So the reason I do it this was is so rather than a simple '2> err' is so that I can see the output of the command when developing the test.  It is far easier than having to do a `cat "trash directory XXX/err"`.  The sharness framework will suppress both stdout and stderr from a test unless '-v' is given so I don't see extra noise as a problem.  If the test does fail for some reason then it is helpful to see the output of the command without having to track down the output file name.  If a test fails on a build server the output file may not even be accessible.

Nevertheless, I understand it is not the convention used in the other tests, so if you still don't like I it I will reluctantly change it after I am done developing the tests and the code.  Let me know.

pin/gc/gc.go
kevina: Yeah that is one of the things I didn't like also.  Working on fixing that today.

pin/gc/gc.go
```diff
@@ -144,30 +144,30 @@ func ColoredSet(ctx context.Context, pn pin.Pinner, ls dag.LinkService, bestEffo
...
 	if errors {
-		return nil, CoundNotFetchAllLinksError
+		return nil, ErrCoundNotFetchAllLinks
 	} else {
 		return gcs, nil
 	}
 }
 
-var CoundNotFetchAllLinksError = errors.New("garbage collection aborted: could not retrieve some links")
+var ErrCoundNotFetchAllLinks = errors.New("garbage collection aborted: could not retrieve some links")
```
whyrusleeping: typo here (not sure if you fixed already, going through commit by commit)

test/sharness/t0087-repo-robust-gc.sh
```diff
@@ -37,6 +39,23 @@ test_gc_robust_part1() {
 		ipfs repo gc
 	'
 
+	test_expect_success "corrupt the root node of 1MB file" '
+		ls -l "$HASH1FILE"
```
whyrusleeping: Why is this ls here?

test/sharness/t0087-repo-robust-gc.sh
```diff
@@ -37,6 +39,23 @@ test_gc_robust_part1() {
...
+		dd if=/dev/zero of="$HASH1FILE" count=1 bs=100 conv=notrunc
+	'
+
+	test_expect_success "'ipfs repo gc' should abort without removing anything" '
+		test_must_fail ipfs repo gc 2>&1 | tee gc_err &&
+		grep -q "could not retrieve links for $HASH1" gc_err &&
+		grep -q "aborted" gc_err
+	'
+
+	test_expect_success "leaf nodes where not removed after gc" '
```
whyrusleeping: typo, s/where/were/

test/sharness/t0087-repo-robust-gc.sh
kevina: I was seeing something.  Will remove.

test/sharness/t0087-repo-robust-gc.sh
```diff
@@ -37,6 +39,23 @@ test_gc_robust_part1() {
...
+		ipfs cat $LEAF3 > /dev/null &&
+		ipfs cat $LEAF4 > /dev/null
```
whyrusleeping: Is this test run both with and without a daemon? If yes, we should add a `--timeout` flag here, if no, we should make sure to run this test with and without a daemon ;)

pin/gc/gc.go
kevina: Fixed.

test/sharness/t0087-repo-robust-gc.sh
kevina: fixed

core/commands/repo.go
```diff
@@ -58,28 +58,20 @@ order to reclaim hard disk space.
 			return
 		}
 
-		gcOutChan, erro := corerepo.GarbageCollectAsync(n, req.Context())
+		gcOutChan := corerepo.GarbageCollectAsync(n, req.Context())
 
-		outChan := make(chan interface{})
+		outChan := make(chan interface{}, len(gcOutChan))
```
whyrusleeping: What does len on a channel tell us? Is it the capacity? or the number of items currently waiting in it?

test/sharness/t0087-repo-robust-gc.sh
kevina: @whyrusleeping  Um, that is what the "--offline" flag is for.

core/commands/repo.go
```diff
@@ -65,24 +66,11 @@ order to reclaim hard disk space.
...
-			switch len(errors) {
-			case 0:
-				return
-			case 1:
-				res.SetError(errors[0], cmds.ErrNormal)
-				return
-			default:
-				res.SetError(corerepo.NewMultiError(errors...), cmds.ErrNormal)
-				return
+			err := corerepo.CollectResult(gcOutChan, func(k *cid.Cid) {
```
whyrusleeping: i like.

whyrusleeping: Though you probably want to select with a context here to make sure there are no accidental hangs

core/commands/repo.go
kevina: Oops, should be `cap`.  Sorry.  Will fix.

core/commands/repo.go
```diff
@@ -51,6 +56,7 @@ order to reclaim hard disk space.
 	},
 	Options: []cmds.Option{
 		cmds.BoolOption("quiet", "q", "Write minimal output.").Default(false),
+		cmds.BoolOption("stream-errors", "Stream errors.").Default(false),
```
whyrusleeping: :+1: I also like this. It is a good option.

pin/gc/gc.go
```diff
@@ -66,7 +66,7 @@ func GC(ctx context.Context, bs bstore.GCBlockstore, ls dag.LinkService, pn pin.
 					err := bs.DeleteBlock(k)
 					if err != nil {
 						errors = true
-						output <- Result{Error: &CouldNotDeleteBlockError{k, err}}
+						output <- Result{Error: &ErrCouldNotDeleteBlock{k, err}}
```
whyrusleeping: I think what you had originally was good. It seems the go convention for error *types* is to suffix them with the word Error.

https://golang.org/pkg/net/#DNSConfigError

core/commands/repo.go
kevina: @whyrusleeping I am never sure when I should be using a context and when it is okay to just read from a channel.  When is checking for ctx.Done() important in the pipeline?

test/sharness/t0087-repo-robust-gc.sh
whyrusleeping: Touche

pin/gc/gc.go
kevina: Okay.  Will revert that commit.

core/commands/repo.go
whyrusleeping: (discussed in IRC, putting here for posterity)

Any time youre sending or receiving on a channel that is outside the control of your immediate goroutine, its a good idea to use a context. In this case, the reader is the commands lib, and the commands lib will stop reading from the channel if the user cancels the operation.

test/sharness/t0087-repo-robust-gc.sh
kevina: resolved via irc, no need for `--timeout` flag.

kevina:
Okay, I address all your comments.  I think this is ready.  Will rebase once #3700 is merged.

whyrusleeping:
@kevina great, thanks!

kevina:
@whyrusleeping rebased

kevina:
@whyrusleeping @Kubuxu rebased

core/commands/repo.go
```diff
@@ -58,23 +65,41 @@ order to reclaim hard disk space.
...
 			defer close(outChan)
-			for k := range gcOutChan {
-				outChan <- k
+			unreportedError := false
+			var lastErr error
+			if streamErrors {
+				for res := range gcOutChan {
+					if unreportedError {
+						outChan <- &GcResult{Error: lastErr.Error()}
+						unreportedError = false
```
whyrusleeping: Instead of trying to return the last error as the responses error, just return all errors through the outChan and if there were errors, have the response error be something like `"encountered errors during gc run"`

kevina: Okay done.

whyrusleeping:
Sorry, been really pressed for time. Will review soon and have this in the 0.4.8 release

Kubuxu:
As a note: This doesn't resolve problems I am having with my repo (and users might have in case of corruption): 1. I can't remove the blocks it prints that are invalid 2. I can't remove invalid pins from the pinset.

kevina:
@Kubuxu Correct.  However, it does make it easier to find what is going on.  It is unsafe to continue if some links cannot be retrieved for a pinned object.

Based on what `isPinnedWithType` is doing I think you should be able to remove the bad CIDs (unless they are pinned indirectly) with `pin rm`.

We could also add a `-f` or `--force` option to force removal of pins without first checking if the object is pinned.

Kubuxu:
I am close to scrapping this repo. I am not able to remove those pins, nor list pins in general.

kevina:
@Kubuxu please don't scrap.  It might make an interesting test case...

Kubuxu:
I have run it over with @whyrusleeping repofix program but I think it is somewhere in btrfs snapshots.

kevina:
@Kubuxu If you can easily restore (and have the space to save) the repo in the state before the repofix program that might be helpful.  If not than don't worry about it.

Kubuxu:
Yes I can, I have snapper running so have snapshots of this repo before repofix. I will archive it on my PC when I get back to my studio (sunday), and then I can expose it for you over 60Mbit/s pipe if you want to download it.

kevina:
Thanks.

core/commands/repo.go
```diff
@@ -87,26 +110,29 @@ order to reclaim hard disk space.
 				return nil, err
 			}
 
-			marshal := func(v interface{}) (io.Reader, error) {
-				obj, ok := v.(*corerepo.KeyRemoved)
+			for v := range outChan {
```
whyrusleeping: This should still use the marshal func construction like it previously did. Output for stdout can go through the same style of returned pipe, and we can print to stderr within that.

The reason for this is to avoid the double error printing that you can see in @Kubuxu's sample output

kevina: I am not 100% sure I am following, but I attempted to do what you want.  It did not seam to help with the duplicate error reporting.

kevina: @whyrusleeping for example here is the output from a test case when run with `-v`
```
expecting success: 
                test_must_fail ipfs repo gc --stream-errors 2>&1 | tee repo_gc_out &&
                grep -q "Error: could not retrieve links for $LEAF1" repo_gc_out &&
                grep -q "Error: could not retrieve links for $LEAF2" repo_gc_out &&
                grep -q "Error: garbage collection aborted" repo_gc_out

Error: could not retrieve links for QmSijovevteoY63Uj1uC5b8pkpDU5Jgyk2dYBqz3sMJUPc: merkledag: not found
Error: could not retrieve links for QmTbPEyrA1JyGUHFvmtx1FNZVzdBreMv8Hc8jV9sBRWhNA: The block referred to by 'QmTbPEyrA1JyGUHFvmtx1FNZVzdBreMv8Hc8jV9sBRWhNA' was not a valid merkledag node
Error: garbage collection aborted: could not retrieve some links
01:35:09.788 ERROR commands/h: encountered errors during gc run client.go:247
Error: encountered errors during gc run
ok 48 - 'ipfs repo gc --stream-errors' should abort and report each error separately

```

core/commands/repo.go
```diff
@@ -39,6 +40,11 @@ var RepoCmd = &cmds.Command{
 	},
 }
 
+type GcResult struct {
+	Key   *cid.Cid
+	Error string `json:",omitempty"`
```
hoenirvili: Why not name it Err ?

core/commands/repo.go
```diff
@@ -88,18 +111,21 @@ order to reclaim hard disk space.
...
 			marshal := func(v interface{}) (io.Reader, error) {
-				obj, ok := v.(*corerepo.KeyRemoved)
+				obj, ok := v.(*GcResult)
 				if !ok {
 					return nil, u.ErrCast()
 				}
 
-				buf := new(bytes.Buffer)
+				if obj.Error != "" {
+					fmt.Fprintf(res.Stderr(), "Error: %s\n", obj.Error)
```
hoenirvili: Why use fmt.Fprintf and not use a standard ipfs logger?

kevina: The intent was for the output to go to the client's stderr.  The logger won't do thus as far as I know.

core/commands/repo.go
```diff
@@ -39,6 +40,11 @@ var RepoCmd = &cmds.Command{
 	},
 }
 
+type GcResult struct {
```
hoenirvili: Every type exported and every attributes that's also exported should have a well described comment above.

core/corerepo/gc.go
```diff
@@ -89,46 +86,70 @@ func GarbageCollect(n *core.IpfsNode, ctx context.Context) error {
...
-	rmed, err := gc.GC(ctx, n.Blockstore, n.DAG, n.Pinning, roots)
-	if err != nil {
-		return err
-	}
+	rmed := gc.GC(ctx, n.Blockstore, n.DAG, n.Pinning, roots)
+
+	return CollectResult(ctx, rmed, nil)
+}
 
+func CollectResult(ctx context.Context, gcOut <-chan gc.Result, cb func(*cid.Cid)) error {
```
hoenirvili: Please add here also a doc string above the function name.

core/commands/repo.go
Kubuxu: Yes, this part of code runs on Client CLI (IIRC) not on the daemon so we are handling logging/errors manually here.

core/corerepo/gc.go
```diff
@@ -89,46 +86,70 @@ func GarbageCollect(n *core.IpfsNode, ctx context.Context) error {
...
 
-func GarbageCollectAsync(n *core.IpfsNode, ctx context.Context) (<-chan *KeyRemoved, error) {
-	roots, err := BestEffortRoots(n.FilesRoot)
-	if err != nil {
-		return nil, err
+func NewMultiError(errs ...error) *MultiError {
+	return &MultiError{errs[:len(errs)-1], errs[len(errs)-1]}
+}
+
+type MultiError struct {
```
hoenirvili: Doc string.

core/corerepo/gc.go
```diff
@@ -89,46 +86,70 @@ func GarbageCollect(n *core.IpfsNode, ctx context.Context) error {
...
+	default:
+		return NewMultiError(errors...)
+	}
 }
 
-func GarbageCollectAsync(n *core.IpfsNode, ctx context.Context) (<-chan *KeyRemoved, error) {
-	roots, err := BestEffortRoots(n.FilesRoot)
-	if err != nil {
-		return nil, err
+func NewMultiError(errs ...error) *MultiError {
```
hoenirvili: Doc string.

core/corerepo/gc.go
hoenirvili: Also you could tell that the type implements the Error interface.

pin/gc/gc.go
```diff
@@ -14,6 +16,11 @@ import (
 
 var log = logging.Logger("gc")
 
+type Result struct {
```
hoenirvili: Doc string.

pin/gc/gc.go
```diff
@@ -83,35 +103,74 @@ func Descendants(ctx context.Context, getLinks dag.GetLinks, set *cid.Set, roots
 	return nil
 }
 
-func ColoredSet(ctx context.Context, pn pin.Pinner, ls dag.LinkService, bestEffortRoots []*cid.Cid) (*cid.Set, error) {
+func ColoredSet(ctx context.Context, pn pin.Pinner, ls dag.LinkService, bestEffortRoots []*cid.Cid, output chan<- Result) (*cid.Set, error) {
```
hoenirvili: Doc string.

pin/gc/gc.go
```diff
@@ -83,35 +103,74 @@ func Descendants(ctx context.Context, getLinks dag.GetLinks, set *cid.Set, roots
...
+	}
+
+	if errors {
+		return nil, ErrCouldNotFetchAllLinks
+	} else {
+		return gcs, nil
 	}
+}
+
+var ErrCouldNotFetchAllLinks = errors.New("garbage collection aborted: could not retrieve some links")
```
hoenirvili: Exported fields should have a docstring. Also, can we have the name ```ErrCannotFetchLinks``` ?

pin/gc/gc.go
```diff
@@ -83,35 +103,74 @@ func Descendants(ctx context.Context, getLinks dag.GetLinks, set *cid.Set, roots
...
+
+var ErrCouldNotDeleteSomeBlocks = errors.New("garbage collection incomplete: could not delete some blocks")
```
hoenirvili: Doc string and can we have ```ErrCannotDeleteBlocks``` ?

pin/gc/gc.go
```diff
@@ -83,35 +103,74 @@ func Descendants(ctx context.Context, getLinks dag.GetLinks, set *cid.Set, roots
...
+
+type CouldNotFetchLinksError struct {
```
hoenirvili: Doc string.

pin/gc/gc.go
```diff
@@ -83,35 +103,74 @@ func Descendants(ctx context.Context, getLinks dag.GetLinks, set *cid.Set, roots
...
+	Key *cid.Cid
+	Err error
+}
+
+func (e *CouldNotFetchLinksError) Error() string {
```
hoenirvili: Doc string.

pin/gc/gc.go
```diff
@@ -83,35 +103,74 @@ func Descendants(ctx context.Context, getLinks dag.GetLinks, set *cid.Set, roots
...
+	return fmt.Sprintf("could not retrieve links for %s: %s", e.Key, e.Err)
+}
+
+type CouldNotDeleteBlockError struct {
```
hoenirvili: Doc string.

pin/gc/gc.go
```diff
@@ -83,35 +103,74 @@ func Descendants(ctx context.Context, getLinks dag.GetLinks, set *cid.Set, roots
...
+	err = Descendants(ctx, getLinks, gcs, pn.InternalPins())
 	if err != nil {
-		return nil, err
+		errors = true
+		output <- Result{Error: err}
+	}
+
+	if errors {
+		return nil, ErrCouldNotFetchAllLinks
+	} else {
```
hoenirvili: You can dropthe ```else``` stmt.

This should look like
```go 
if errors {
  return nil, ErrCouldNotFetchAllLinks
}

return gcs, nil
```

core/corerepo/gc.go
whyrusleeping: Mentioning that it implements the error interface is pretty redundant and not standard go convention

core/commands/repo.go
kevina: @whyrusleeping since using the marshal func construction didn't fix the double error reporting, do you still want to use the marshal func construction, I think it makes the code more complicated to understand.  But I can leave it if you want.

pin/gc/gc.go
whyrusleeping:
```go
// ColoredSet computes the set of nodes in the graph that are pinned by the pins in the given pinner
```

core/commands/repo.go
```diff
@@ -39,6 +40,12 @@ var RepoCmd = &cmds.Command{
 	},
 }
 
+// GcResult is the result returned by "repo gc" command
+type GcResult struct {
```
whyrusleeping: We can actually remove this whole type in favor of the one in the gc package (just change the field names so the json marshaling matches up correctly)

kevina: The problem is we can't pass the error type through the API.  I get:
```
02:40:33.946 ERROR commands/h: json: cannot unmarshal object into Go value of type error client.go:247
Error: json: cannot unmarshal object into Go value of type error
```

core/corerepo/gc.go
whyrusleeping:
```go
// CollectResult collects the output of a garbage collection run and calls the given callback for each object removed. It also collects all errors into a MultiError which is returned after the gc is completed
```

core/corerepo/gc.go
whyrusleeping:
```go
// NewMultiError creates a new MultiError object from a given slice of errors
```

pin/gc/gc.go
whyrusleeping:
```go
// Result represents an incremental output from a garbage collection run. It contains either an error, or the cid of a removed object
```

core/commands/repo.go
whyrusleeping: Fair, lets figure that out later then

whyrusleeping:
@Kubuxu the circleCI failure here is the iptb failed to connect problem again. Didnt you have an idea on why that was happening?

Kubuxu:
@whyrusleeping nope, couldn't figure it out. It happens much more on Travis and Circle than our CI which means it might be some race or instability.

I tried to replicate it locally (1000 runs) didn't find even one failure.

kevina:
> Yes I can, I have snapper running so have snapshots of this repo before repofix. I will archive it on my PC when I get back to my studio (Sunday), and then I can expose it for you over 60Mbit/s uplink pipe if you want to download it.

@Kubuxu did you ever do this.

I am very surprised you can't just remove the broken pins with the latest IPFS version.  See https://github.com/ipfs/go-ipfs/issues/3796#issuecomment-289657878

PR #3715, opened 2017-02-21
Update gx to version 0.11

Kubuxu:
Previous version 0.10 was built with faulty version of Go 1.7 which has
problems with TCP -> DNS -> UDP timeouts.

This results in many failed calls to `gx install`

This wasn't easy to find.

License: MIT

### DIFF
--- a/bin/Rules.mk
+++ b/bin/Rules.mk
@@ -1,8 +1,8 @@
 include mk/header.mk
 
-dist_root_$(d)=/ipfs/QmTazYLCem8B7pQGLsCj1Q4MxkqGMmD9XugvStLJSQ8uxA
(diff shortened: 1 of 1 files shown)


lgierth:
>  codecov/project — 61.81% (+0.13%) compared to 2ef1f69 

How this change increases test coverage remains the secret of codecov

Kubuxu:
We have some tests that are timing sensitive / random. Some branches will be taken one some occasions and others won't.

Like timeout and repeat might fire or not depending on a speed of CI and otherwise the timeout code might not be covered.

Kubuxu:
We should for sure try to eliminate them, but the higher the real coverage the less of those random flips we will see.

Kubuxu:
What is interesting is that pubsub is sometimes tested and sometimes not...

PR #3720, opened 2017-02-22
routing: Implement GetValuesAsync

mildred:
For libp2p/go-libp2p-kad-dht#47

### DIFF
--- a/routing/mock/centralized_client.go
+++ b/routing/mock/centralized_client.go
@@ -75,6 +75,23 @@ func (c *client) GetValues(ctx context.Context, key string, count int) ([]routin
 	return []routing.RecvdVal{{Val: data, From: c.peer.ID()}}, nil
 }
 
+func (c *client) GetValuesAsync(ctx context.Context, key string, count int) <-chan *routing.RecvdVal {
+	log.Debugf("GetValuesAsync: %s", key)
+	res := make(chan *routing.RecvdVal, 1)
+	defer close(res)
+	data, err := c.GetValue(ctx, key)
+	if err != nil {
+		log.Debugf("GetValuesAsync error: %v", err)
+	} else {
+		res <- &routing.RecvdVal{
+			Val:  data,
+			From: c.peer.ID(),
+		}
+	}
+
+	return res
+}
+
 func (c *client) FindProviders(ctx context.Context, key *cid.Cid) ([]pstore.PeerInfo, error) {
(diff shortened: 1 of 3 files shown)


routing/mock/centralized_client.go
```diff
@@ -75,6 +75,23 @@ func (c *client) GetValues(ctx context.Context, key string, count int) ([]routin
 	return []routing.RecvdVal{{Val: data, From: c.peer.ID()}}, nil
 }
 
+func (c *client) GetValuesAsync(ctx context.Context, key string, count int) <-chan *routing.RecvdVal {
+	log.Debugf("GetValuesAsync: %s", key)
+	res := make(chan *routing.RecvdVal, 1)
+	defer close(res)
+	data, err := c.GetValue(ctx, key)
+	if err != nil {
+		log.Debugf("GetValuesAsync error: %v", err)
```
whyrusleeping: this really makes me think that the `RecvdVal` should contain an error field.

whyrusleeping:
@Kubuxu do you think this looks good? Or should we push to make `GetValues` return a channel of things-that-may-contain-errors?

mildred:
I took this approach with errors because another function did. But at first, my first idea was to return a second channel with possible errors. It's very easy to do and I don't mind updating my PR to match that or similar.

whyrusleeping:
@mildred could you return just a single channel, but have the type of the channel contain either a response or an error (like an option type)? Two channels cat get pretty tricky to deal with at times

mildred:
I started the change to make a response with an error built in, but then couldn't find this discussion any more. What's necessary to do that is to change go-libp2p-routing to add an error field to `RecvdVal`. I now see that's exactly what you suggested, so I'll change all three PR to account for that.

mildred:
Updated the PR, see also changes in libp2p/go-libp2p-kad-dht#47 and libp2p/go-libp2p-routing#7

whyrusleeping:
I need to get to reviewing and merging the dependent PRs. Bumping this task to 0.4.8

whyrusleeping:
@Kubuxu I definitely want this as part of the API. I think its really important for users of the DHT

Stebalien:
Replaced by https://github.com/ipfs/go-ipfs/issues/5232.

Stebalien:
Replaced by https://github.com/ipfs/go-ipfs/issues/5232.

PR #3725, opened 2017-02-24
feat: remove broken spdy multiplexer and enable multiplex by default as second option

dignifiedquire:
- spdystream is not fully spdy spec compatible, creating issues when
  doing interop with js, so it is removed
- muliplex is working well now in interop with js, so being enabled by
  default

This means the new default muxers are `yamux,mplex`


whyrusleeping:
@dignifiedquire fix your commit signature

dignifiedquire:
what's wromg with it? I already pushed a fix with the license stuff

dignifiedquire:
@whyrusleeping I think I did what you want

PR #3727, opened 2017-02-26
Update fuse lib, push it to gx

### LINKED ISSUE #3471: Support OSXFUSE 3+
Currently it doesn't work with our fuse lib version.

Kubuxu:
Resolves #3471, part of migration to gx.

@kevina this should resolve some issues with FUSE (and possibly fix the tests hanging).


Kubuxu:
Removed gitcop messages from revert commits.

whyrusleeping:
@Kubuxu why are there revert commits?

whyrusleeping:
I see why the code was needed. I guess it makes sense to revert

Kubuxu:
Those revert commits are for patches that made error message on OSX with OSXFUSE 3.0+ which older version of lib did not support.

whyrusleeping:
@Kubuxu conflict in the package.json, wanna rebase?

PR #3743, opened 2017-03-03
Add support for using CidV1 in 'ipfs add'

### LINKED ISSUE #3699: Add flag to ipfs add to return CIDv1
As discussed on IRC, add a flag for `ipfs add` (and possibly other commands) to use CIDv1. cc @whyrusleeping

kevina:
Closed #3699 

License: MIT

### DIFF
--- a/core/commands/add.go
+++ b/core/commands/add.go
@@ -40,6 +40,8 @@ const (
 	fstoreCacheOptionName = "fscache"
 )
 
+const adderOutChanSize = 8
+
 var AddCmd = &cmds.Command{
 	Helptext: cmds.HelpText{
 		Tagline: "Add a file or directory to ipfs.",
@@ -208,7 +210,7 @@ You can now refer to the added file in a gateway, like so:
 			return
 		}
 
-		outChan := make(chan interface{}, 8)
+		outChan := make(chan interface{}, adderOutChanSize)
 		res.SetOutput((<-chan interface{})(outChan))
 
 		fileAdder.Out = outChan
--- a/core/commands/add.go
+++ b/core/commands/add.go
@@ -236,7 +236,7 @@ You can now refer to the added file in a gateway, like so:
 		fileAdder.Silent = silent
 		fileAdder.RawLeaves = rawblks
 		fileAdder.NoCopy = nocopy
-		fileAdder.Prefix = prefix
+		fileAdder.Prefix = &prefix
 
 		if hash {
 			md := dagtest.Mock()
--- a/core/commands/add.go
+++ b/core/commands/add.go
@@ -202,14 +202,14 @@ You can now refer to the added file in a gateway, like so:
 		bserv := blockservice.New(addblockstore, exch)
 		dserv := dag.NewDAGService(bserv)
 
-		outChan := make(chan interface{}, 8)
-		res.SetOutput((<-chan interface{})(outChan))
-
 		fileAdder, err := coreunix.NewAdder(req.Context(), n.Pinning, n.Blockstore, dserv)
 		if err != nil {
 			res.SetError(err, cmds.ErrNormal)
 			return
 		}
+
+		outChan := make(chan interface{}, 8)
+		res.SetOutput((<-chan interface{})(outChan))
 
 		fileAdder.Out = outChan
 		fileAdder.Chunker = chunker
--- a/core/commands/add.go
+++ b/core/commands/add.go
@@ -38,6 +38,7 @@ const (
 	rawLeavesOptionName   = "raw-leaves"
 	noCopyOptionName      = "nocopy"
 	fstoreCacheOptionName = "fscache"
+	cidVersionOptionName  = "cid-version"
 )
 
 const adderOutChanSize = 8
@@ -88,6 +89,7 @@ You can now refer to the added file in a gateway, like so:
 		cmds.BoolOption(rawLeavesOptionName, "Use raw blocks for leaf nodes. (experimental)"),
 		cmds.BoolOption(noCopyOptionName, "Add the file using filestore. (experimental)"),
 		cmds.BoolOption(fstoreCacheOptionName, "Check the filestore for pre-existing blocks. (experimental)"),
+		cmds.IntOption(cidVersionOptionName, "Cid version. Non-zero value will change default of 'raw-leaves' to true. (experimental)").Default(0),
 	},
 	PreRun: func(req cmds.Request) error {
 		quiet, _, _ := req.Option(quietOptionName).Bool()
@@ -161,6 +163,7 @@ You can now refer to the added file in a gateway, like so:
 		rawblks, rbset, _ := req.Option(rawLeavesOptionName).Bool()
 		nocopy, _, _ := req.Option(noCopyOptionName).Bool()
 		fscache, _, _ := req.Option(fstoreCacheOptionName).Bool()
+		cidVer, _, _ := req.Option(cidVersionOptionName).Int()
 
 		if nocopy && !cfg.Experimental.FilestoreEnabled {
 			res.SetError(errors.New("filestore is not enabled, see https://git.io/vy4XN"),
@@ -174,6 +177,16 @@ You can now refer to the added file in a gateway, like so:
 
 		if nocopy && !rawblks {
 			res.SetError(fmt.Errorf("nocopy option requires '--raw-leaves' to be enabled as well"), cmds.ErrNormal)
+			return
+		}
+
+		if cidVer >= 1 && !rbset {
+			rawblks = true
+		}
+
+		prefix, err := dag.PrefixForCidVersion(cidVer)
+		if err != nil {
+			res.SetError(err, cmds.ErrNormal)
 			return
 		}
 
@@ -223,6 +236,7 @@ You can now refer to the added file in a gateway, like so:
 		fileAdder.Silent = silent
 		fileAdder.RawLeaves = rawblks
 		fileAdder.NoCopy = nocopy
+		fileAdder.Prefix = prefix
 
 		if hash {
 			md := dagtest.Mock()
--- a/core/coreunix/add.go
+++ b/core/coreunix/add.go
@@ -103,7 +103,7 @@ type Adder struct {
 	mroot      *mfs.Root
 	unlocker   bs.Unlocker
 	tempRoot   *cid.Cid
-	Prefix     cid.Prefix
+	Prefix     *cid.Prefix
 }
 
 func (adder *Adder) mfsRoot() (*mfs.Root, error) {
@@ -113,6 +113,7 @@ func (adder *Adder) mfsRoot() (*mfs.Root, error) {
 	rnode := unixfs.EmptyDirNode()
 	rnode.SetPrefix(adder.Prefix)
 	mr, err := mfs.NewRoot(adder.ctx, adder.dagService, rnode, nil)
+	mr.Prefix = adder.Prefix
 	if err != nil {
 		return nil, err
 	}
--- a/core/coreunix/add.go
+++ b/core/coreunix/add.go
@@ -69,13 +69,7 @@ type AddedObject struct {
 }
 
 func NewAdder(ctx context.Context, p pin.Pinner, bs bstore.GCBlockstore, ds dag.DAGService) (*Adder, error) {
-	mr, err := mfs.NewRoot(ctx, ds, unixfs.EmptyDirNode(), nil)
-	if err != nil {
-		return nil, err
-	}
-
 	return &Adder{
-		mr:         mr,
 		ctx:        ctx,
 		pinning:    p,
 		blockstore: bs,
@@ -87,7 +81,6 @@ func NewAdder(ctx context.Context, p pin.Pinner, bs bstore.GCBlockstore, ds dag.
 		Wrap:       false,
 		Chunker:    "",
 	}, nil
-
 }
 
 // Adder holds the switches passed to the `add` command.
@@ -107,13 +100,28 @@ type Adder struct {
 	NoCopy     bool
 	Chunker    string
 	root       node.Node
-	mr         *mfs.Root
+	mroot      *mfs.Root
 	unlocker   bs.Unlocker
 	tempRoot   *cid.Cid
+	Prefix     cid.Prefix
+}
+
+func (adder *Adder) mfsRoot() (*mfs.Root, error) {
+	if adder.mroot != nil {
+		return adder.mroot, nil
+	}
+	rnode := unixfs.EmptyDirNode()
+	rnode.SetPrefix(adder.Prefix)
+	mr, err := mfs.NewRoot(adder.ctx, adder.dagService, rnode, nil)
+	if err != nil {
+		return nil, err
+	}
+	adder.mroot = mr
+	return adder.mroot, nil
 }
 
 func (adder *Adder) SetMfsRoot(r *mfs.Root) {
-	adder.mr = r
+	adder.mroot = r
 }
 
 // Constructs a node from reader's data, and adds it. Doesn't pin.
@@ -122,11 +130,13 @@ func (adder Adder) add(reader io.Reader) (node.Node, error) {
 	if err != nil {
 		return nil, err
 	}
+
 	params := ihelper.DagBuilderParams{
 		Dagserv:   adder.dagService,
 		RawLeaves: adder.RawLeaves,
 		Maxlinks:  ihelper.DefaultLinksPerBlock,
 		NoCopy:    adder.NoCopy,
+		Prefix:    adder.Prefix,
 	}
 
 	if adder.Trickle {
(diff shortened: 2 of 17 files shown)


kevina:
@whyrusleeping I am not sure I am happy with this

What I described in https://github.com/ipfs/go-ipfs/issues/3699#issuecomment-282907030 was only sufficient for adding files with CidV1.  It did not handle symbolic links or directories.

In order to handle directories I needed to touch the `mfs` package.  I added a cidVer parameter to `mfs.Mkdir` and added a `SetCidVersion` method to the Directory class. 

To make sure the root node was CidV1 I decided it was best to pass in the Cid version when creating the adder rather then setting it later.

The changes to the merkeldag package are not strictly necessary, but it seams the best way to set the Cid version of a ProtoNode.

kevina:
@whyrusleeping one question: Do you think `--cid-version=1` should imply `--raw-leaves`?  I think it should and would be one less combination to test for.

whyrusleeping:
@kevina hrm... I think setting cid version to 1 could probably change the default value for raw leaves to true, but i think people should still be able to set raw leaves to false and cid version to 1. Even if we don't allow it through the command line people could manually construct graphs that were that way

kevina:
@whyrusleeping okay, i think this is ready for review

core/commands/add.go
```diff
@@ -78,6 +79,7 @@ You can now refer to the added file in a gateway, like so:
 		cmds.StringOption(chunkerOptionName, "s", "Chunking algorithm to use."),
 		cmds.BoolOption(pinOptionName, "Pin this object when adding.").Default(true),
 		cmds.BoolOption(rawLeavesOptionName, "Use raw blocks for leaf nodes. (experimental)"),
+		cmds.IntOption(cidVersionOptionName, "Cid version. (experimental)").Default(0),
```
Kubuxu: This should mention `enables raw blocks`.

core/commands/add.go
```diff
@@ -162,15 +169,15 @@ You can now refer to the added file in a gateway, like so:
...
-		res.SetOutput((<-chan interface{})(outChan))
-
-		fileAdder, err := coreunix.NewAdder(req.Context(), n.Pinning, n.Blockstore, dserv)
+		fileAdder, err := coreunix.NewAdder(req.Context(), n.Pinning, n.Blockstore, dserv, cidVer)
 		if err != nil {
 			res.SetError(err, cmds.ErrNormal)
 			return
 		}
 
+		outChan := make(chan interface{}, 8)
```
Kubuxu: Extract this constant, maybe use the same as AllKeysChan or some other/new but I would prefer it extracted.

merkledag/cidversion.go
```diff
@@ -0,0 +1,26 @@
+package merkledag
+
+import (
+	"fmt"
+)
+
+// CidVersion is a a cid version number that is guaranteed to be valid
+type CidVersion struct {
```
Kubuxu: I would make it `type CidVersion int`. It will be simpler to use and more idiomatic.

Kubuxu: Also there isn't a reason it couldn't be just a constant?

```
type CidVerision int

const (
CID0 CidVersion = 0
CID1 CidVersion = 1
)
```

You can still have methods on it.

Kubuxu: I think you went the way you did because as a way of preventing creation of different CID versions but IMO it should be done with helper function when CID version is from User Interface.

Kubuxu: Because nothing prevents me from doing `CidVersion{2}` either way.

Kubuxu:
Also the code is not formatter right: https://ci.ipfs.team/blue/organizations/jenkins/go-ipfs/detail/PR-3743/5/pipeline

merkledag/cidversion.go
kevina: > Because nothing prevents me from doing CidVersion{2} either way.

Um, if you do that outside the merkeldag package you will get
```
implicit assignment of unexported field 'version' in merkledag.CidVersion literal
```

kevina: The idea was to provide some additional type safety so once you have a CidVersion you don't need to return an error if the version is invalid as it has already been checked.

core/commands/add.go
kevina: This is not new code I just moved things around a little.  We can do this in a separate p.r.

core/commands/add.go
kevina: I agree, but got stuck on how to state in concisely suggestion welcome.

merkledag/cidversion.go
Kubuxu: Ahh, missed that. Ok, but I would still be for having variables aliasing usable values.

core/commands/add.go
Kubuxu: This would take you 10s. General rule many of as try to strive for is to leave code in better shape than we found it.

core/commands/add.go
Kubuxu: `CID Version. Non-zero value will change default of 'rawleaves' to true.

core/commands/add.go
```diff
@@ -139,7 +141,12 @@ You can now refer to the added file in a gateway, like so:
 		silent, _, _ := req.Option(silentOptionName).Bool()
 		chunker, _, _ := req.Option(chunkerOptionName).String()
 		dopin, _, _ := req.Option(pinOptionName).Bool()
-		rawblks, _, _ := req.Option(rawLeavesOptionName).Bool()
+		rawblks, rawblksFound, _ := req.Option(rawLeavesOptionName).Bool()
+		cidVer, _, _ := req.Option(cidVersionOptionName).Int()
+
+		if cidVer >= 1 && !rawblksFound {
```
kevina: This check is already done in the call to NewAdder.

core/commands/add.go
kevina: I agree, but thought it would be simpler just to add it for the adder for now.

For example, should every command check for this flag if it is applicable, and if it is not supported yet error out?

merkledag/cidversion.go
kevina: Alias added, although not currently used.

kevina:
> I don't like the fact that now every MFS API function will have to take cidversion argument. We will have to either live with it or think about it.

Neither do I.  It is only the `MkDir` call that needs an cidversion argument.  I added it the `MkDir` call because it can create multiple nodes that all will need to be set to the correct Cid version.  There are two additional options.

1) Create a new function `MkDirWCid` and have `MkDir` call it.

2) Walk the tree created and fix the Cid version after the fact. I see this as inefficient both in terms of speed and amount of code, but will avoid having to modify the `MkDir` call. 

@whyrusleeping thoughs?

merkledag/cidversion.go
kevina: Also note, I am not against doing the way you suggested.

I just want to be able to assume that once we have a CidVersion we can guarantee we have something valid and don't want to have to return an error (instead we can panic).  I think it will be reasonable to assume that the user (in this case the programmer) won't do a cast  from an int to to CidVersion and will use the helper function.

@whyrusleeping thoughts.

core/coreunix/add.go
```diff
@@ -67,8 +67,15 @@ type AddedObject struct {
 	Bytes int64  `json:",omitempty"`
 }
 
-func NewAdder(ctx context.Context, p pin.Pinner, bs bstore.GCBlockstore, ds dag.DAGService) (*Adder, error) {
-	mr, err := mfs.NewRoot(ctx, ds, unixfs.EmptyDirNode(), nil)
+func NewAdder(ctx context.Context, p pin.Pinner, bs bstore.GCBlockstore, ds dag.DAGService, v int) (*Adder, error) {
```
whyrusleeping: Lets not change the parameters here. Instead, lets just add a `SetCidVersion` method to the adder.

importer/helpers/dagbuilder.go
```diff
@@ -33,6 +35,9 @@ type DagBuilderParams struct {
 	// instead of using the unixfs TRaw type
 	RawLeaves bool
 
+	// Prefix specifies cid version and hashing function
```
whyrusleeping: CidVersion only specifies the cid version, not the hashing function

importer/helpers/dagbuilder.go
```diff
@@ -139,7 +155,11 @@ func (db *DagBuilderHelper) GetNextDataNode() (*UnixfsNode, error) {
 			raw:     true,
 		}, nil
 	} else {
-		blk := NewUnixfsBlock()
+		blk := &UnixfsNode{
```
whyrusleeping: no `NewUnixfsBlock` anymore?

mfs/ops.go
```diff
@@ -104,7 +105,7 @@ func PutNode(r *Root, path string, nd node.Node) error {
 
 // Mkdir creates a directory at 'path' under the directory 'd', creating
 // intermediary directories as needed if 'mkparents' is set to true
-func Mkdir(r *Root, pth string, mkparents bool, flush bool) error {
+func Mkdir(r *Root, pth string, mkparents bool, flush bool, cidVer dag.CidVersion) error {
```
whyrusleeping: Lets have this be a field on the `Root` object, that way we can avoid changing the mfs APIs

importer/helpers/dagbuilder.go
kevina: Sorry, fixed.  It was a leftover comment from an earlier attempt.

importer/helpers/dagbuilder.go
kevina: Correct.  It was only used in one place so I just inlined the code.  I removed the function to make sure I cached all possible places where it was called to make sure the correct Cid version was always set.

kevina: If you want I can turn it into a method like I did for `NewUnixfsNode`.  But since it was only used in one place I decided it wasn't worth the effort.

mfs/ops.go
kevina: Good idea.  Done.

importer/helpers/dagbuilder.go
whyrusleeping: Yeah... just to maintain continuity with the previous methods. the whole 'raw' thing here is realllly just legacy code to prevent hashes from changing at this point

kevina:
@whyrusleeping okay, this should be ready now

kevina:
@whyrusleeping target this for 0.4.8?  Or do you want to try for a global option instead?

importer/helpers/dagbuilder.go
```diff
@@ -33,6 +35,9 @@ type DagBuilderParams struct {
 	// instead of using the unixfs TRaw type
 	RawLeaves bool
 
+	// CID Version to use
+	CidVersion dag.CidVersion
```
whyrusleeping: I think instead of using a struct for this, we should just use a `cid.Prefix`. That way we make it much easier for us to switch the hash function in the future

kevina:
@whyrusleeping okay, changed to setting the prefix instead of just the CID version

merkledag/node.go
```diff
@@ -28,13 +28,37 @@ type ProtoNode struct {
...
+		return v0CidPrefix, nil
+	case 1:
+		return v1CidPrefix, nil
+	default:
+		return cid.Prefix{}, fmt.Errorf("unknown CID version: %d", version)
+	}
+}
+
+func (n *ProtoNode) SetPrefix(prefix cid.Prefix) {
+	n.Prefix = prefix
```
whyrusleeping: Lets make sure to set `n.Prefix.Codec = cid.DagProtobuf`, just for sanitys sake

kevina: Good idea will do.

kevina: Done

whyrusleeping:
@kevina it looks like go tests are failing

kevina:
@whyrusleeping sorry, it looks like the latest commit, change to using the full prefix broke the tests.  Will look into it Friday and sorry for not catching it myself.

kevina:
@whyrusleeping it was setting the `Codec` in SetPrefix that broke the tests. This should not be done when the prefix is the go-initialized default value.  The only way I could think to test for the go-initialized default value us to check that MhType is non-zero as all the other fields could legitimately be zero.  As an alternative to setting the Prefix we could check for it and panic if it MhType is non-zero and Codec is not DagProtobuf.

I separated out in a separate commit in case we just want to drop that change.

In any case we should be green now.

merkledag/node.go
kevina: Well it turned out not to be such a good idea. :(

whyrusleeping:
@kevina hrm... I think we need to not call `SetPrefix` unless its called up higher. Which means we probably want the prefix in the adder and similar structures to be a pointer. That way if the pointer is nil, we can not worry about calling `SetPrefix` on the created nodes

kevina:
@whyrusleeping except that it's not a pointer...

kevina:
@whyrusleeping please ignore my last comment.  Read to fast.  Deleting.

kevina:
@whyrusleeping what would you think about just having SetPrefix accept a pointer and resting it to the default value if it nil?  Leads to fewer special cases.  The first commit I just pushed implemented this way (merkeldag: change SetPrefix param to a pointer and reset the prefix).

If not I also did it the way I think you had in mind, it leads to more special cases (of checking if the pointer is nil) and it was not 100% clear when to switch from accepting a pointer to a non-pointer. The second commit changes it to do what I think you had in mind.

Let me know which way you prefer.  I will either drop the last commit or squash the two into one depending on your preference.

whyrusleeping:
Making `SetPrefix` accept a pointer SGTM.

merkledag/node.go
```diff
@@ -292,7 +294,7 @@ func (n *ProtoNode) Cid() *cid.Cid {
 	}
 
 	if n.Prefix.Codec == 0 {
```
whyrusleeping: This might panic if its nil

whyrusleeping: oh wait, nvm. You don't store it as a pointer :+1:

kevina:
@whyrusleeping Okay SetPrefix now accepts a pointer.  A nil value will simply reset the pointer to the default value.

whyrusleeping:
changed the title to be more amenable to changelogging

Kubuxu:
Can you also rebase to master (it should solve some build failure issues).

kevina:
@Kubuxu yeah I am working on it.  This is a non-trivial fix.  mfs.Directory.node disappeared and I need to figure what changed and how to fix it.

kevina:
So commit bb09ffd756f17876589baf76a008f3c5fa31beca (implement an HAMT for unixfs directory sharding) complicated things. It should have a fix by the end of Monday.

whyrusleeping:
@kevina alright, i'm gonna bump this to 0.4.9 then. Thanks!

kevina:
@whyrusleeping so setting the MFS root after the fact became problematic.  Have a look at the commit "Fixes" and let me know what you think.  This commit will be squashed but I kept it separate to make it easier to see the new changes.

core/coreunix/add.go
```diff
@@ -108,19 +100,28 @@ type Adder struct {
...
+	if err != nil {
+		return nil, err
+	}
+	adder.mroot = mr
+	return adder.mroot, nil
 }
 
 func (adder *Adder) SetMfsRoot(r *mfs.Root) {
-	adder.mr = r
+	adder.mroot = r
```
whyrusleeping: I'd prefer not changing the naming of things, but not a big deal

unixfs/io/dirbuilder.go
```diff
@@ -79,6 +80,17 @@ func NewDirectoryFromNode(dserv mdag.DAGService, nd node.Node) (*Directory, erro
 	}
 }
 
+// SetPrefix sets the prefix of the root node
+func (d *Directory) SetPrefix(prefix cid.Prefix) {
+	if d.dirnode != nil {
+		d.dirnode.SetPrefix(prefix)
+	}
+	// FIXME: Should we do this? -- kevina
```
whyrusleeping: Yeah, we should probably set the shards prefix info too

core/coreunix/add.go
kevina: I can change it back.  I did this as a trick to make sure I cached where all instance `mr` was used.  I thought about naming the method `mr()` but that seamed a bit terse for a method name, even if it was only for internal use.

unixfs/io/dirbuilder.go
kevina: Okay, will do.  Have something in a few days.

whyrusleeping:
@kevina update here?

kevina:
Have something by the end of tomorrow, if not today.

kevina:
@whyrusleeping sorry for the delay, this should be good to go.  I will squash the "Fixes" commit once you are okay with my changes.

whyrusleeping:
Oh, and could you fix the codeclimate issues?

kevina:
Rebased but tests are failing again, will fix.

kevina:
But `codeclimate` is at least happy.

kevina:
Some how "make deps" in the shareness directory was not rebuilding things correctly for me.  In any case we a green and this is now ready to merge.

PR #3749, opened 2017-03-06
fix: filestore silently being skipped on add if it wasn't enabled

Kubuxu:
License: MIT


Kubuxu:
Nie ma za co.

Issue #3752, opened 2017-03-06
Too many open files during gc


whyrusleeping:
While running a gc i got this error message:
```
13:08:36.065 ERROR blockstore: blockstore.AllKeysChan got err:%!(EXTRA *errors.errorString=walk failed: open /home/whyrusleeping/.ipfs/blocks/FY: too many open files) blockstore.go:201
```

Kubuxu:
The Query method has to be handle the FD problems

gammazero:
This will likely be solved by https://github.com/ipfs/go-ds-flatfs/pull/136

gammazero:
This should be fixed by changes to how batches of data are written.

Handled by:
- https://github.com/ipfs/go-ds-flatfs/pull/142
- https://github.com/ipfs/kubo/pull/11092

PR #3756, opened 2017-03-07
Limit number of objects in mdag batch to prevent too many fds issue

whyrusleeping:
In certain situations the 8MB size limit here would result in a large number of objects in the batch, which combined with peer connections, can easily hit the file descriptor limit.

License: MIT

### DIFF
--- a/merkledag/merkledag.go
+++ b/merkledag/merkledag.go
@@ -68,7 +68,15 @@ func (n *dagService) Add(nd node.Node) (*cid.Cid, error) {
 }
 
 func (n *dagService) Batch() *Batch {
-	return &Batch{ds: n, MaxSize: 8 * 1024 * 1024}
+	return &Batch{
+		ds:      n,
+		MaxSize: 8 << 20,
+
+		// By default, only batch up to 128 nodes at a time.
+		// The current implementation of flatfs opens this many file
+		// descriptors at the same time for the optimized batch write.
+		MaxBlocks: 128,
+	}
 }
 
 // Get retrieves a node from the dagService, fetching the block in the BlockService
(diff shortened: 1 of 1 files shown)


merkledag/merkledag.go
```diff
@@ -68,7 +68,15 @@ func (n *dagService) Add(nd node.Node) (*cid.Cid, error) {
...
 func (n *dagService) Batch() *Batch {
-	return &Batch{ds: n, MaxSize: 8 * 1024 * 1024}
+	return &Batch{
+		ds:      n,
+		MaxSize: 8 << 20,
+
+		// By default, only batch up to 256 nodes at a time.
+		// The current implementation of flatfs opens this many file
+		// descriptors at the same time for the optimized batch write.
+		MaxBlocks: 256,
```
Kubuxu: I would reduce it even further. 256 blocks worst case scenario is 512 open fds (directory and file), and with almost 500 peers it is getting close to 1024 ulimit we set.

whyrusleeping: mmm... fair enough. How about 200 even? I think we should rework how flatfs works/replace it soon.

whyrusleeping:
should resolve #3752

Kubuxu:
It doesn't fix everything, I am still getting this failure:

```
expecting success: 
		ipfs add --raw-leaves -r -q somedir > /dev/null
	
22:51:36.165 ERROR commands/h: open /home/kubuxu/go/src/github.com/ipfs/go-ipfs/test/sharness/trash directory.t0270-filestore.sh/.ipfs/blocks/5H/put-659704546: too many open files client.go:247
Error: open /home/kubuxu/go/src/github.com/ipfs/go-ipfs/test/sharness/trash directory.t0270-filestore.sh/.ipfs/blocks/5H/put-659704546: too many open files
not ok 29 - normal add without fscache duplicates data
#	
#			ipfs add --raw-leaves -r -q somedir > /dev/null
#		
```

It is slightly modified test as previously the error was being hidden by var capture.

Kubuxu:
Ok, now it vanished.

PR #3757, opened 2017-03-07
make raw leaves work with 'ipfs get'

whyrusleeping:
License: MIT

### DIFF
--- a/core/commands/get.go
+++ b/core/commands/get.go
@@ -70,22 +70,24 @@ may also specify the level of compression by specifying '-l=<1-9>'.
 			return
 		}
 
-		pbnd, ok := dn.(*dag.ProtoNode)
-		if !ok {
-			res.SetError(err, cmds.ErrNormal)
+		switch dn := dn.(type) {
+		case *dag.ProtoNode:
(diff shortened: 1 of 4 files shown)


whyrusleeping:
For some reason the filestore tests on OSX are failing: https://s3.amazonaws.com/archive.travis-ci.org/jobs/208472671/log.txt

Looks like an issue with the 'check repo size' stuff

core/commands/get.go
```diff
@@ -70,22 +70,24 @@ may also specify the level of compression by specifying '-l=<1-9>'.
...
+		case *dag.ProtoNode:
+			size, err := dn.Size()
+			if err != nil {
+				res.SetError(err, cmds.ErrNormal)
+				return
+			}
+
+			res.SetLength(size)
+		case *dag.RawNode:
+			res.SetLength(uint64(len(dn.RawData())))
```
Kubuxu: This makes me think that we should have sort of Size method or better unixfs interface.

whyrusleeping: Yeah, unixfs interfaces would and will be nice

PR #3770, opened 2017-03-09
Add ipfs add --quieter

### LINKED ISSUE #3371: Option for recursive add to return only the main hash
#### Version information:

go-ipfs version: 0.4.4-
Repo version: 4
System version: amd64/linux
Golang version: go1.7

#### Type:

Enhancement

#### Priority: 

P4

#### Description:

That would be nice if

`ipfs add -Q -r myDirectory`

could return only the last hash.

Kubuxu:
It will print out only the final hash
Resolves #3371

### DIFF
--- a/core/commands/add.go
+++ b/core/commands/add.go
@@ -26,6 +26,7 @@ var ErrDepthLimitExceeded = fmt.Errorf("depth limit exceeded")
 
 const (
 	quietOptionName       = "quiet"
+	quieterOptionName     = "quieter"
 	silentOptionName      = "silent"
 	progressOptionName    = "progress"
 	trickleOptionName     = "trickle"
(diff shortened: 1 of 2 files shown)


core/commands/add.go
```diff
@@ -278,12 +283,37 @@ You can now refer to the added file in a gateway, like so:
...
-		quiet, _, err := req.Option("quiet").Bool()
+		quiet, _, err := req.Option(quietOptionName).Bool()
+		quieter, _, _ := req.Option(quieterOptionName).Bool()
+		quiet = quiet || quieter
 		if err != nil {
 			res.SetError(u.ErrCast(), cmds.ErrNormal)
 			return
 		}
 
+		if quieter {
```
whyrusleeping: I think this logic can be DRYed up and moved into the loop below

PR #3775, opened 2017-03-10
core: add NatPortMap and default it to true

### LINKED ISSUE #2964: Config option for disabling NAT discovery
In some locations (e.g. datacenters) you don't need NAT discovery. Sometimes it'll take its full couple of seconds though! If we can disable that, restarts of the daemon will be faster.

kevina:
Closes #2964

License: MIT

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -732,12 +732,12 @@ func constructPeerHost(ctx context.Context, id peer.ID, ps pstore.Peerstore, bwr
 		network.Swarm().Filters.AddDialFilter(f)
 	}
 
-	newOpts := []interface{}{bwr}
+	hostOpts := []interface{}{bwr}
 	if opts.NatPortMap {
-		newOpts = append(newOpts, p2pbhost.NATPortMap)
+		hostOpts = append(hostOpts, p2pbhost.NATPortMap)
 	}
 
-	host := p2pbhost.New(network, newOpts...)
+	host := p2pbhost.New(network, hostOpts...)
 
 	return host, nil
 }
--- a/core/core.go
+++ b/core/core.go
@@ -737,7 +737,7 @@ func constructPeerHost(ctx context.Context, id peer.ID, ps pstore.Peerstore, bwr
 		newOpts = append(newOpts, p2pbhost.NATPortMap)
 	}
 
-	host := p2pbhost.New(network, newOpts)
+	host := p2pbhost.New(network, newOpts...)
 
 	return host, nil
 }
--- a/core/core.go
+++ b/core/core.go
@@ -214,7 +214,7 @@ func (n *IpfsNode) startOnlineServices(ctx context.Context, routingOption Routin
 	}
 
 	peerhost, err := hostOption(ctx, n.Identity, n.Peerstore, n.Reporter,
-		addrfilter, tpt, protec)
+		addrfilter, tpt, protec, &ConstructPeerHostOpts{NatPortMap: cfg.Swarm.NatPortMap})
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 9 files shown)


kevina:
@whyrusleeping defaulting `NatPortMap` to be true turned out to be more complicated that you outlined.  In order for it to be done reliably it needs to be set to true when ever a new Config struct is created.  I attempted to do this, but I am sure I didn't get all the case.

Other than that I followed your outline.

This still needs tests that the actual NAT discovery is turned off, but I am not sure how to do that.  Hints welcome.

core/core.go
```diff
@@ -728,7 +732,12 @@ func constructPeerHost(ctx context.Context, id peer.ID, ps pstore.Peerstore, bwr
 		network.Swarm().Filters.AddDialFilter(f)
 	}
 
-	host := p2pbhost.New(network, p2pbhost.NATPortMap, bwr)
+	newOpts := []interface{}{bwr}
```
whyrusleeping: probably call this hostOpts

repo/config/config.go
```diff
@@ -34,6 +34,20 @@ type Config struct {
 	Experimental Experiments
 }
 
+// NewConfig creates any new Config with all the values set the
+// default
+func NewConfig() *Config {
```
whyrusleeping: I would call this `NewDefaultConfig`

repo/fsrepo/serialize/serialize.go
```diff
@@ -63,11 +63,12 @@ func Load(filename string) (*config.Config, error) {
 		return nil, errors.New("ipfs not initialized, please run 'ipfs init'")
 	}
 
-	var cfg config.Config
-	err := ReadConfigFile(filename, &cfg)
+	cfg := config.NewConfig()
+
```
whyrusleeping: I wouldnt add a space here

test/sharness/t0023-config-default.sh
```diff
@@ -0,0 +1,47 @@
...
+test_init_ipfs
+
+test_expect_success "Swarm.NatPortMap set to true in inital config" '
+	echo "true" > expected &&
+	ipfs config Swarm.NatPortMap > actual &&
+	test_cmp expected actual
+'
+
+test_expect_success "Swarm.NatPortMap has line in config file" '
+	grep -q NatPortMap .ipfs/config
```
whyrusleeping: redirect output to devnull

test/sharness/t0023-config-default.sh
```diff
@@ -0,0 +1,47 @@
...
+'
+
+test_expect_success "remove Swarm.NatPortMap from config file" '
+	sed -i "s/NatPortMap/XxxYyyyZzz/" .ipfs/config
```
whyrusleeping: use `$IPFS_PATH`

test/sharness/t0023-config-default.sh
```diff
@@ -0,0 +1,47 @@
...
+'
+
+test_expect_success "load config file by replacing a unrelated key" '
+	ipfs config --json Swarm.DisableBandwidthMetrics false
```
whyrusleeping: Add a note here:
```
Note: this is relying on an implementation detail where setting a config field loads the json blob into the config struct, causing defaults to be properly set. Simply 'reading' the config would read the json directly and not cause defaults to be reset
```

test/sharness/t0023-config-default.sh
kevina: That is what the `-q` is for.

whyrusleeping: ah, gotcha

kevina:
Addressed CR feedback.  I am not going to be comfortable with the PR until we figure out how to test for NAT discovery.  I caught a rather serious bug that should of caused tests to fail in `constructPeerHost`.  This seams to indicate that NAT discovery is not tested at all.

whyrusleeping:
@kevina NAT traversal is not tested, yes. I've looked into testing this (spending a significant amount of time in doing so) and came up empty. It seems that even nat traversal libraries don't have proper tests.

kevina:
Okay than I am going to do some hackish manual testing to make sure that the setting is at least being honored.  I will remove the WIP from the title when that is done.

kevina:
All right, I tested it manually and the settings are being honored.

I not sure I like how the default to true is being handled but if @whyrusleeping and @Kubuxu are okay with it than so am I.

whyrusleeping:
@Kubuxu What do you think?

Kubuxu:
This just highlights defectiveness of config system to me. What do you think about shifting to default + overrides config? If we decide for it, I would love to see some other project that did something similar in past and it worked out well.

kevina:
Or we could change the option to `NoNatPortMap` until we design a better config system.  That is the option I would recommend.

Kubuxu:
@whyrusleeping your call

whyrusleeping:
Alright, we can use `NoNatPortMap`.

Lets also make sure to add this field to the config options document in `docs/config.md`

kevina:
Because removing the default to true was the majority of this pull request I decided to redo the P.R. as #3798.  Closing.  Please don't delete the branch.

kevina:
@whyrusleeping is there a reason you deleted this branch when I asked "Please don't delete the branch." We (or at least me) might want to refer back to it.  I thought the idea was for old branched to be moved to go-ipfs-archived.  For some reason GitHub isn't allowing me to restore this branch so I re-pushed my code from my local repo.

whyrusleeping:
@kevina whoops. I just compulsively click the delete branch button

PR #3782, opened 2017-03-13
fix: remove bloom filter check on Put call in blockstore

Kubuxu:
To prevent put we need to have conclusive information if item is
contained in the repo, bloom filter won't give this information.
It only says if it is for sure not contained.

License: MIT


Kubuxu:
Really hard to tests as it essentially noop. It calls `hasCached` and expect `has` and `ok` to be both true, which can never happen.

whyrusleeping:
@Kubuxu ah, i see. Okay, LGTM

Kubuxu:
Yey, new merge window is open.

PR #3795, opened 2017-03-17
CHANGELOG: Fix config command for FilestoreEnabled .

jefft0:
The command to enable the filestore did not work as-is. It is missing `--json` . Should be:

    ipfs config --json Experimental.FilestoreEnabled true


whyrusleeping:
@jefft0 good catch, thanks. Mind signing off your commit?

jefft0:
I signed off the commit.

whyrusleeping:
Thanks!

PR #3798, opened 2017-03-18
Add DisableNatPortMap option.

kevina:


kevina:
Redo of https://github.com/ipfs/go-ipfs/pull/3775.

kevina:
@whyrusleeping I went with `DisableNatPortMap` instead of `NoNatPortMap` to be consistent with the `DisableBandwidthMetrics` option.

kevina:
Also, manually tested since I am not sire how to write tests for this.

core/core.go
```diff
@@ -709,12 +709,16 @@ func listenAddresses(cfg *config.Config) ([]ma.Multiaddr, error) {
 	return listen, nil
 }
 
-type HostOption func(ctx context.Context, id peer.ID, ps pstore.Peerstore, bwr metrics.Reporter, fs []*net.IPNet, tpt smux.Transport, protc ipnet.Protector) (p2phost.Host, error)
+type ConstructPeerHostOpts struct {
```
Kubuxu: Because HostOption func argument list is getting very long and in other occasion we decided to stop adding params and just add one struct. The existing parameters will be moved to this struct on other occasion.

core/core.go
```diff
@@ -709,12 +709,16 @@ func listenAddresses(cfg *config.Config) ([]ma.Multiaddr, error) {
...
+	DisableNatPortMap bool
+}
+
+type HostOption func(ctx context.Context, id peer.ID, ps pstore.Peerstore, bwr metrics.Reporter, fs []*net.IPNet, tpt smux.Transport, protc ipnet.Protector, opts *ConstructPeerHostOpts) (p2phost.Host, error)
 
 var DefaultHostOption HostOption = constructPeerHost
 
 // isolates the complex initialization steps
-func constructPeerHost(ctx context.Context, id peer.ID, ps pstore.Peerstore, bwr metrics.Reporter, fs []*net.IPNet, tpt smux.Transport, protec ipnet.Protector) (p2phost.Host, error) {
+func constructPeerHost(ctx context.Context, id peer.ID, ps pstore.Peerstore, bwr metrics.Reporter, fs []*net.IPNet, tpt smux.Transport, protec ipnet.Protector, opts *ConstructPeerHostOpts) (p2phost.Host, error) {
```
hoenirvili: If the function contains more than 3, 4 params make a new struct type like ```constructParams``` or ```constructConfig``` and pass it rather than use this long lists of params. 
This way we group things much more tighter and make notify the caller that, he needs to use a single type that holds multiple params in order to call the function. If the func if is not exported make {name}{Params/Cfg-prefix} type not exported also. 
We don't want to **expose** anymore types or functions that we don't use in other packages, because every function/type that are exposed comes with more difficult a cost of maintaining. We should try to avoid any unnecessary burden for the caller and make a stable well defined API.

repo/config/swarm.go
```diff
@@ -3,4 +3,5 @@ package config
 type SwarmConfig struct {
```
hoenirvili: Could you please also add doc strings if we are already here?

hoenirvili:
Thanks again for the PR. This is wonderful that you want to help and add more features to go-ipfs but every PR should contain unit tests. After you will add unit tests we will consider merging this.
Any thoughts on this ? @Kubuxu @whyrusleeping

Kubuxu:
@hoenirvili it was also discussed before, Nat traversal is very hard to test. Currently it is not tested at all and requiring testing for this PR would require creation of test suite for nat traversal in general.

whyrusleeping:
The only somewhat reasonable way i've come up with for testing nat traversal is to have a series of docker images using a special networking setup and a upnp server in another docker image. I had very little luck figuring out how to set this up.

I've even emailed developers of miniupnp asking how they do automated testing, and the general response i get is that testing for this tends to be fairly manual.

@hoenirvili if you can figure out how to actually do automated testing of NAT traversal, that would be fantastic

Issue #3498, opened 2016-12-12
Test utility lib


Kubuxu:
As follow up to: https://github.com/multiformats/go-multibase/pull/6

It might be worth to think a bit about introduction of test utility lib so we don't have to use 3 lines for to pass through an error or to compare two values. Space on a screen is a resource too that shouldn't be wasted.

I would also like for use to settle on one lib so it is uniform across codebases.

https://github.com/stretchr/testify looks quite nice.

What is your opinion?
Should we gx'ify the lib if we start using some?

whyrusleeping:
I'm pretty against external test libs in general, if everyone else disagrees with me, i could be convinced otherwise, but in general i like writing vanilla go tests.

wigy-opensource-developer:
When it comes to me, I like redhead girls. But besides personal taste, why do you dislike external test libs? 😋

Back to serious, you mentioned vanilla tests. Does that mean you do not like to extract methods reusable all around test code? Or you just do not want to depend on a library maintained and vendored independently from Protocol Labs?

hsanjuan:
I agree with @whyrusleeping. I love rspec, and I also landed on `testify` looking for something similar. But Go is Go. We cannot have magic mocks like in Ruby. So it wasn't going to help so much and I went for less dependencies and vanilla tests, which do the job all the same, albeit with some extra lines. Almost everything in Go takes extra space on screen, so I find it a weak reason to introduce idioms and [test-only]-deps.

wigy-opensource-developer:
You are right, @hsanjuan. Assertions for equality, their semantics, their options are so common case that it should be in the Go language itself. That would save developers learning new idioms for these common tasks in each project.

But Go is Go, so what options we have to reduce noise in the test codes?

Grown up on http://xunitpatterns.com/ I really treat test code as first-class citizen that needs the same level of refactoring and maintenance as production code to avoid slow-down in the long-run as new features are added and bugs are fixed.

wigy-opensource-developer:
@Kubuxu Looking at multiformats/go-multibase#10 I got the impression that introducing abstractions for test code seems to be a tough call in this team? :wink:

hoenirvili:
This project needs to stop using the testing pkg from std and move to a more mature and serious one like  https://labix.org/gocheck. 
I'm also a serious contributor on another big open source project [Juju](https://github.com/juju/juju/), which is much bigger than the go-ipfs so it uses this library I mentioned and also this custome one https://github.com/juju/testing.

[...]

PR #3801, opened 2017-03-19
Safe check rlimit syscall and refactoring

hoenirvili:
This PR contains: 
- this contains refactoring
- this adds safe checking around rlimit syscall
- more doc string


Kubuxu:
Ok, few things:
 - please split refactors, added tests and docs into separate comments. 
 - if you are changing the rlimit logic please also change it for freebsd, or preferably extract the code that has to be different on freebsd vs rest (getting and setting the rlimit) and reuse the rest.
 - the dependency of yours is not available - please coordinate with me on IRC to pin it on our nodes
 - more in review commetns

cmd/ipfs/runmain_test.go
```diff
@@ -7,16 +7,15 @@ import (
...
 	"io/ioutil"
 	"os"
-	"testing"
 )
 
 // this abuses go so much that I felt dirty writing this code
 // but it is the only way to do it without writing custom compiler that would
 // be a clone of go-build with go-test
-func TestRunMain(t *testing.T) {
+func (m mainRun) TestRunMain(c *gc.C) {
```
Kubuxu: Not sure if it will fly with coverage collection code. We will see when dependency thing is fixed.

Kubuxu:
The ulimit raising is tested in sharness, so the unit tests are not needed 100% but useful nevertheless. If it causes problems with coverage collection we can move the ulimit raising to separate package and test it alone.

hoenirvili:
I'm open to any good solution. From my standpoint of view I think every file should contains a *_test.go. It's kind of scary that the project is missing so much unit testing.

Kubuxu:
Yes we are constantly trying to push for more unit testing. We have also quite a big sharness testing suite. Checkout `test/sharness/`

cmd/ipfs/main_test.go
```diff
@@ -1,17 +1,20 @@
 package main
 
 import (
-	"testing"
-
 	"github.com/ipfs/go-ipfs/commands"
+
+	gc "gopkg.in/check.v1"
```
Kubuxu: This dependency has to be rewritten, please run `gx-go rewrite`

hoenirvili: Fixed !

whyrusleeping:
This needs to be broken up into logical atomic changesets

cmd/ipfs/ulimit_freebsd.go
```diff
@@ -3,42 +3,53 @@
...
+				)
+			}
+
+			rlimit.Max = max
+			// set the rlimit in unprivileged mode
+			if err = syscall.Setrlimit(syscall.RLIMIT_NOFILE, &rlimit); err != nil {
+				return fmt.Errorf("error setting ulimit without hard limit: %s", err)
+			}
+
+			log.Infof("Successfully raised file descriptor limit to %d.\n", maxFds)
```
Kubuxu: Info is not default on channel so it won't be visible. We want this message to be visible during daemon start, that is the reason it was using `fmt.Printf` before.

cmd/ipfs/runmain_test.go
Kubuxu: This change is breaking coverage collection `func TestRunMain(t *testing.T)` is run when testing starts and it takes over the testing process. In this case it just executes ipfs process not the tests. It is guided by the build flag of this file.

hsanjuan:
Sorry to party-poop too, but there has been previous proposals to bring-in testing helper libraries and so far they have not moved forward. This would effectively introduce testing deps. See https://github.com/ipfs/go-ipfs/issues/3498

hoenirvili:
@hsanjuan check #3498

cmd/ipfs/main_test.go
```diff
@@ -1,17 +1,20 @@
...
 
-func TestIsCientErr(t *testing.T) {
-	t.Log("Catch both pointers and values")
-	if !isClientError(commands.Error{Code: commands.ErrClient}) {
-		t.Errorf("misidentified value")
-	}
-	if !isClientError(&commands.Error{Code: commands.ErrClient}) {
-		t.Errorf("misidentified pointer")
-	}
+func (m mainRun) TestIsCientErr(c *gc.C) {
```
whyrusleeping: This new way of testing this is not shorter, cleaner, or easier to understand. I'm not on board with this.

hoenirvili: I disagree completely.
I think more mantainers of this project should take a look and give more feedback.
The c.Assert besides of line number and fail message it contains  these lines as well "expected: value, obtained: value". which it gives us way more information on "what went wrong". If you want I can use testify to write these tests.
But keep relaying and using the testing pkg from std lib it's not good idea, in the long run. (in my opinion).

whyrusleeping: You are free to have your opinions, but to convince me that this is worthwhile and not just some random contributor trying to push their opinions on a project they just found you'll have to do better convincing me of this. 

You need to choose better examples to prove your point, because this specific one is not helped by shoving a test lib into it. "expected: true, got: false" Is quite clear from reading the existing test and the debuggability is not helped by adding in the extra cognitive overhead of this framework.

Kubuxu: I have to agree with @whyrusleeping on this, in this case `misidentified value` is much more descriptive than `expected: true, got: false`.

Also replacing the `c.Assert(ok, gc.Equals, true)` with
```go
if !ok {
    t.Error("misidentified value")
}
```
Is easier and faster to type, conveys more information and everybody knowns what is going on.
There is no value added or `c.Assert` in this part of code.

hoenirvili:
@whyrusleeping @Kubuxu now it's ok?

cmd/ipfs/init.go
```diff
@@ -9,6 +9,7 @@ import (
 	"path"
 
 	context "context"
+
```
whyrusleeping: while we're modifying this, lets go ahead and drop the named import (holdover from `"golang.org/x/net/context"`) and move it up with the rest of the stdlib

cmd/ipfs/ulimit.go
```diff
@@ -5,15 +5,25 @@ import (
 	"strconv"
 )
 
-var ipfsFileDescNum = uint64(1024)
+// maxFds is the maximum number of file descriptors that the go ipfs
```
whyrusleeping: s/the go ipfs/go-ipfs/

cmd/ipfs/ulimit.go
```diff
@@ -5,15 +5,25 @@ import (
...
+// can use. Default values is 1024. This can be overwritten by the
```
whyrusleeping: s/Default values/The default value/

cmd/ipfs/ulimit.go
```diff
@@ -5,15 +5,25 @@ import (
...
+
+		fds, err := strconv.ParseUint(val, 10, 64)
 		if err != nil {
 			log.Errorf("bad value for IPFS_FD_MAX: %s", err)
-		} else {
-			ipfsFileDescNum = uint64(n)
+			return
 		}
+
+		// assign the
```
whyrusleeping: useless comment

cmd/ipfs/ulimit_freebsd.go
```diff
@@ -3,42 +3,53 @@
...
+	max := rlimit.Max
+	rlimit.Cur, rlimit.Max = nfds, nfds
+
+	// the soft limit is the value that the kernel enforces for the corresponding resource
+	// the hard limit acts as a ceiling for the soft limit
+	// an unprivileged process may only set its soft limit to a value
+	// in the range from 0 up to the hard limit
+	if err = syscall.Setrlimit(syscall.RLIMIT_NOFILE, &rlimit); err != nil {
+		// the process does not have permission so we should only set the soft value
+		if err == syscall.EPERM {
```
whyrusleeping: reverse the logic here: `if err != syscall.EPERM { ...`

cmd/ipfs/ulimit_test.go
```diff
@@ -0,0 +1,106 @@
...
+	if maxFds != uint64(1024) {
+		t.Errorf("Maximum file descriptors default value changed")
+	}
+}
+
+func TestManageInvalidNFds(t *testing.T) {
+	t.Logf("Testing file descriptor invalidity")
+	var err error
+	if err = os.Unsetenv("IPFS_FD_MAX"); err != nil {
+		t.Errorf("Cannot unset the IPFS_FD_MAX env variable")
```
whyrusleeping: Use fatal for these, the rest of the test is invalid if this assertion (or the others) fails.

cmd/ipfs/ulimit_test.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	// call to check and set the maximum file descriptor from the env
+	setMaxFds()
+
+	if err = managefd(); err == nil {
+		t.Errorf("Managefd should return an error")
+	}
+
+	if err != nil {
```
whyrusleeping: else if

cmd/ipfs/ulimit_windows.go
```diff
@@ -0,0 +1,11 @@
+// +build windows
+
+package main
+
+// managefd checks for the maximum number of file descriptors that the system supports
```
whyrusleeping: This comment is incorrect.

cmd/ipfs/ulimit_unix.go
```diff
@@ -3,40 +3,50 @@
...
-			log.Error("adjusting max")
-			rLimit.Max = ipfsFileDescNum
+	max := rlimit.Max
+	rlimit.Cur, rlimit.Max = maxFds, maxFds
+
+	// the soft limit is the value that the kernel enforces for the corresponding resource
+	// the hard limit acts as a ceiling for the soft limit
+	// an unprivileged process may only set its soft limit to a value
+	// in the range from 0 up to the hard limit
+	if err = syscall.Setrlimit(syscall.RLIMIT_NOFILE, &rlimit); err != nil {
```
whyrusleeping: This logic is nearly identical to the other freebsd logic. I'm sure we can DRY it up a bit

cmd/ipfs/ulimit_test.go
```diff
@@ -0,0 +1,106 @@
...
+func TestManagefdWithEnvSet(t *testing.T) {
+	t.Logf("Testing file descriptor manager with IPFS_FD_MAX set")
+	var err error
+	if err = os.Unsetenv("IPFS_FD_MAX"); err != nil {
+		t.Errorf("Cannot unset the IPFS_FD_MAX env variable")
+	}
+
+	rlimit := syscall.Rlimit{}
+	if err = syscall.Getrlimit(syscall.RLIMIT_NOFILE, &rlimit); err != nil {
+		t.Errorf("Cannot get the file descriptor count")
```
Kubuxu: `Fatal` here too as rest is useless if this fails

Kubuxu:
I would suggest pulling the fd managing code to separate package (it has nothing to do with ipfs main process itself) and also switching from directly referencing functions under build tag to using `init` to inject the platform specific code into a private variable. This way we don't have to have build tag for every not supported OS.

Example how it could be done:
Core file (no build tags):
```go
// hasFDManage is used to determine if we should run tests
// (if we have generic, platform antagonistic test (open a lot of empty files)).
var hasFDManage bool = false
var manageFDs func() error = func() { return nil }

func ManageFDs() error {
    return manageFDs()
}
```
Then file that builds on unix:
```go
func init() {
hasFDManage = true
manageFDs = unixManageFDs()
}

func unixManageFDs()
```

Then as freebsd is special, one file for non freebsd unix:
```go
// +build unix,!freebsd

func getFDlimits...
func setFDlimits...
```

and separate one for freebsd:
```
// +build freebsd

func getFDlimits...
func setFDlimits...
```

cmd/ipfs/util/ulimit.go
```diff
@@ -0,0 +1,31 @@
...
+	// not have a valid fds number notify the user
+	if val := os.Getenv("IPFS_FD_MAX"); val != "" {
+
+		fds, err := strconv.ParseUint(val, 10, 64)
+		if err != nil {
+			log.Errorf("bad value for IPFS_FD_MAX: %s", err)
+			return
+		}
+
+		// assign the
```
whyrusleeping: whats this comment here for?

hoenirvili: Sorry, I will delete it.

hoenirvili:
Why not use the runtime.GOOS ?

whyrusleeping:
@hoenirvili I'm assuming youre asking why use build tags over a switch on `runtime.GOOS`. And my answer to that is "you can", but at some point you will have to deal with the fact that the `syscall.RLimit` has different types on different systems. The only decent way to deal with that is to use build tags and normalize it with some small wrapper code.

hoenirvili:
Can you please take a look one more time at the code? What changes should I make?

whyrusleeping:
@hoenirvili Something like this: https://gist.github.com/whyrusleeping/8c3bae34723f209665480f39d2163848

Let me know if i should be more detailed

cmd/ipfs/util/ulimit.go
```diff
@@ -0,0 +1,30 @@
+package util
+
+import (
+	"os"
+	"strconv"
+
+	"github.com/prometheus/common/log"
```
whyrusleeping: This is the wrong logging package

whyrusleeping:
The reasoning for the construction in the linked gist is to ensure that when trying this for new unsupported platforms, one: it doesnt fail to build because we forgot a special file with its particular build tags, and two: its much easier to implement support on new systems, just by implementing the getFDLimit and setFDLimit functions on the new platform.

cmd/ipfs/util/ulimit.go
whyrusleeping: Import `logging "github.com/ipfs/go-logging"` and then do a global `var log = logging.Logger("ulimit")`

hoenirvili:
Sorry for taking so much time... I was very busy these couple of weeks. I've updated the patch, you can take a look and tell what improvements I can add further.

hoenirvili:
What about now ?

whyrusleeping:
@hoenirvili Hey, I'll take another look at this tomorrow. Sorry for taking so long to get back to you here, i've been out for a week

hoenirvili:
@whyrusleeping Thanks !

cmd/ipfs/util/ulimit.go
```diff
@@ -0,0 +1,85 @@
...
+	// getlimit returns the soft and hard limits of file descriptors counts
+	getLimit func() (int64, int64, error)
+	// set limit sets the soft and hard limits of file descriptors counts
+	setLimit func(int64, int64) error
+)
+
+// maxFds is the maximum number of file descriptors that go-ipfs
+// can use. The default value is 1024. This can be overwritten by the
+// IPFS_FD_MAX env variable
+var maxFds = uint64(1024)
```
Kubuxu: Master has increased it to 2048.

cmd/ipfs/util/ulimit.go
```diff
@@ -0,0 +1,85 @@
...
+		if err != nil {
+			log.Errorf("bad value for IPFS_FD_MAX: %s", err)
+			return
+		}
+
+		maxFds = fds
+	}
+}
+
+func ManageFdLimit() (err error) {
```
Kubuxu: remove the naming of the return value here

cmd/ipfs/util/ulimit.go
```diff
@@ -0,0 +1,85 @@
...
+	if !supportsFDManagement {
+		return
```
Kubuxu: and do explicit nil or error return here

cmd/ipfs/util/ulimit.go
```diff
@@ -0,0 +1,85 @@
...
+		return err
+	}
+
+	max := int64(maxFds)
+	// the soft limit is the value that the kernel enforces for the
+	// corresponding resource
+	// the hard limit acts as a ceiling for the soft limit
+	// an unprivileged process may only set it's soft limit to a
+	// alue in the range from 0 up to the hard limit
+	if err = setLimit(max, max); err != nil {
```
Kubuxu: I am not sure this is needed. Just setting it to max, hard should be ok (and before checking if max is < than hard).

cmd/ipfs/util/ulimit.go
```diff
@@ -0,0 +1,85 @@
...
+		if err != syscall.EPERM {
+			return fmt.Errorf("error setting: ulimit: %s", err)
+		}
+
+		// the process does not have permission so we should only
+		// set the soft value
+		if max > hard {
+			return errors.New(
+				"cannot set rlimit, IPFS_FD_MAX is larger than the hard limit",
```
Kubuxu: That might be not true, the build in default rise might be larger than the hard limit

cmd/ipfs/util/ulimit.go
```diff
@@ -0,0 +1,85 @@
+package util
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"strconv"
+	"syscall"
+
+	logging "github.com/ipfs/go-log"
```
Kubuxu: run `gx-go rw` to fix it and allow tests to run

hoenirvili: Fixed

cmd/ipfs/util/ulimit.go
hoenirvili: Fixed

cmd/ipfs/util/ulimit.go
hoenirvili: Fixed

cmd/ipfs/util/ulimit.go
hoenirvili: Fixed

Kubuxu:
Sharness test checking for the message is failing.

hoenirvili:
@Kubuxu Yes, I know but, I don't have experience with the sharness framework to modify the code in order to make the sharness test logic pass. Can you give some advice on this topic?

whyrusleeping:
Hey @hoenirvili! Sorry for taking nearly a month to get back to you here. I pulled your branch down and figured out how to get the tests passing, here was my diff:
```
diff --git a/cmd/ipfs/util/ulimit.go b/cmd/ipfs/util/ulimit.go
index ed7f37fc1..d4669631a 100644
--- a/cmd/ipfs/util/ulimit.go
+++ b/cmd/ipfs/util/ulimit.go
@@ -49,12 +49,17 @@ func ManageFdLimit() error {
        }
 
        setMaxFds()
-       _, hard, err := getLimit()
+       soft, hard, err := getLimit()
        if err != nil {
                return err
        }
 
        max := int64(maxFds)
+
+       if max <= soft {
+               return nil
+       }
+
        // the soft limit is the value that the kernel enforces for the
        // corresponding resource
        // the hard limit acts as a ceiling for the soft limit
@@ -77,9 +82,8 @@ func ManageFdLimit() error {
                        return fmt.Errorf("error setting ulimit wihout hard limit: %s", err)
                }
 
-               fmt.Printf("Successfully raised file descriptor limit to %d.\n", max)
-               return nil
        }
+       fmt.Printf("Successfully raised file descriptor limit to %d.\n", max)
 
        return nil
 }
```

I also removed the file cmd/ipfs/ulimit.go since it is no longer needed

hoenirvili:
@whyrusleeping changed.

kevina:
@whyrusleeping I just rebased

whyrusleeping:
@kevina thanks!

@hoenirvili Thanks for the changes! Sorry about the delay, i'm working on getting through the mountain of PRs now.

PR #3804, opened 2017-03-20
test(commands): fix EOF failure with Multipart.Read

hoenirvili:
The fix contains the changing nil with io.EOF
because acording to the Read interface implementation
An instance, (in our case Part.Read) of this general case
is that a Reader returning a non-zero number of bytes at
the end of the input stream may return either err == EOF or err == nil.
So this means that the Part.Reader chose to return
io.EOF instead of nil.

### DIFF
--- a/commands/files/file_test.go
+++ b/commands/files/file_test.go
@@ -127,7 +127,7 @@ anotherfile
 	if file, err := mpf.NextFile(); file != nil || err != ErrNotDirectory {
 		t.Fatal("Expected a nil file and ErrNotDirectory")
 	}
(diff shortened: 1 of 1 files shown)


Kubuxu:
Could you make commit message less verbose: for example: `test(commands): fix EOF failure with Multipart.Read`.

Also could you add parenthesis around the error logic or split it into separate if condition? It would make this code a bit more readable.

hoenirvili:
@Kubuxu Can you restart just the travis build? I stopped for no reason.

hoenirvili:
@Kubuxu merge?

PR #3805, opened 2017-03-20
daemon: use fsrepo.IsInitialized to test for initialization

### LINKED ISSUE #3773: ipfs daemon --init does not work
#### Version information:

```
go-ipfs version: 0.4.6-
Repo version: 5
System version: amd64/linux
Golang version: go1.8
```

#### Description:

I am passing `--init`, but it does not do what is saying it will do, which is, initialize the repo:

```
$ export IPFS_PATH=/ipfs
$ /usr/local/bin/ipfs daemon --init
Initializing daemon...
Error: no IPFS repo found in /ipfs.
please run: 'ipfs init'
```

kevina:
Use fsrepo.IsInitialized to test for initialization instead of just testing if the directory exists.

Also add test cases for '--init' option, including one the creates an empty directory.

Closes #3773.

### DIFF
--- a/cmd/ipfs/daemon.go
+++ b/cmd/ipfs/daemon.go
@@ -26,7 +26,6 @@ import (
 	iconn "gx/ipfs/QmT6jBTqNKhhb8dbzCEMUNkGhm3RuRActcMhpShAHLpQtp/go-libp2p-interface-conn"
 	"gx/ipfs/QmVCNGTyD4EkvNYaAp253uMQ9Rjsjy2oGMvcdJJUoVRfja/go-multiaddr-net"
 	"gx/ipfs/QmX3QZ5jHEPidwUrymXV1iSCSUhdGxj15sm2gP4jKMef7B/client_golang/prometheus"
-	util "gx/ipfs/QmZuY8aV7zbNXVy6DyN9SmnuH3o9nG852F4aTiSBpts8d1/go-ipfs-util"
 	pstore "gx/ipfs/Qme1g4e3m2SmdiSGGU3vSWmUStwUjc5oECnEriaK9Xa1HU/go-libp2p-peerstore"
 )
 
@@ -227,12 +226,9 @@ func daemonFunc(req cmds.Request, res cmds.Response) {
 
 	if initialize {
 
-		// now, FileExists is our best method of detecting whether ipfs is
-		// configured. Consider moving this into a config helper method
-		// `IsInitialized` where the quality of the signal can be improved over
-		// time, and many call-sites can benefit.
-		if !util.FileExists(req.InvocContext().ConfigRoot) {
-			err := initWithDefaults(os.Stdout, req.InvocContext().ConfigRoot)
+		cfg := req.InvocContext().ConfigRoot
+		if !fsrepo.IsInitialized(cfg) {
+			err := initWithDefaults(os.Stdout, cfg)
 			if err != nil {
 				res.SetError(err, cmds.ErrNormal)
 				return
--- a/test/sharness/t0063-daemon-init.sh
+++ b/test/sharness/t0063-daemon-init.sh
@@ -1,0 +1,52 @@
+#!/bin/sh
+#
+# Copyright (c) 2014 Juan Batiz-Benet
+# MIT Licensed; see the LICENSE file in this repository.
+#
+
+test_description="Test daemon --init command"
+
+. lib/test-lib.sh
+
+# We don't want the normal test_init_ipfs but we need to make sure the
+# IPFS_PATH is set correctly.
+export IPFS_PATH="$(pwd)/.ipfs"
+
+# safety check since we will be removing the directory
+if [ -e "$IPFS_PATH" ]; then
+	echo "$IPFS_PATH exists"
+	exit 1
+fi
+
+test_ipfs_daemon_init() {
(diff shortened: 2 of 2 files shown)


test/sharness/t0063-daemon-init.sh
```diff
@@ -0,0 +1,52 @@
...
+	'
+}
+
+test_expect_success "remove \$IPFS_PATH dir" '
+	rm -rf "$IPFS_PATH"
+'
+test_ipfs_daemon_init
+
+test_expect_success "create empty \$IPFS_PATH dir" '
+	rm -rf "$IPFS_PATH"
```
Kubuxu: best to put `&&` after the rm, I know it "shouldn't" fail but still.

kevina: Sorry.  The "&&" are too easy to forget about.

cmd/ipfs/daemon.go
```diff
@@ -227,11 +226,7 @@ func daemonFunc(req cmds.Request, res cmds.Response) {
 
 	if initialize {
 
-		// now, FileExists is our best method of detecting whether ipfs is
-		// configured. Consider moving this into a config helper method
-		// `IsInitialized` where the quality of the signal can be improved over
-		// time, and many call-sites can benefit.
-		if !util.FileExists(req.InvocContext().ConfigRoot) {
+		if !fsrepo.IsInitialized(req.InvocContext().ConfigRoot) {
```
hoenirvili: Instead of calling twice the ```req.InvocContex().ConfigRoot``` why not make under initialize something like 
```go
if initialize {
   cfg := req.InvocContext()
   if !fsrepo.IsInitialized(cfg) {
       err := initWithDefaults(os.Stdou, cfg)
// code
```

kevina:
@whyrusleeping okay done

test/sharness/t0063-daemon-init.sh
```diff
@@ -0,0 +1,52 @@
...
+
+test_ipfs_daemon_init() {
+	# Doing it manually since we want to launch the daemon with an
+	# empty or non-existent repo; the normal
+	# test_launch_ipfs_daemon does not work since it assumes the
+	# repo was created a particular way with regard to the API
+	# server.
+
+	test_expect_success "'ipfs daemon --init' succeeds" '
+		ipfs daemon --init >actual_daemon 2>daemon_err &
```
Kubuxu: This will be problematic with other daemons people have running (port numbers). Not really sure how to solve it.

whyrusleeping: Yeah... i'm not sure how to fix it either. We should look into adding a way to specify ports from here.

Though not sure if this should block the PR, Its probably a separate concern

kevina: It will cause the test to fail on users machines who also have an IPFS daemon running.  Should we flag it some how?  If so I am not sure how.

Kubuxu: We could check if ports from config are bound (8080, 4001, 5001) using netcat and if any of those are, skip the test. How that sounds?

whyrusleeping: Yeah, lets do that. Its better to print a message about why it fails rather than just randomly failing.

whyrusleeping: (lets not skip the test, I think its okay if it fails and says it needs those ports)

kevina: @Kubuxu do you think you could push a commit, I am not familiar with `netcat`.  If not I figure it out tonight.

kevina: Actually because I use a simple sleep and the init takes some time the test will pass as the daemon won't try to start right away.  I tested the tests with out my fix and the test fails because the daemon fails right away if there is a problem reading the repo.

If I up the timeout the daemon error output will make it fairly clear that there is a port conflict.
```
Error: serveHTTPApi: manet.Listen(/ip4/127.0.0.1/tcp/5001) failed: listen tcp4 127.0.0.1:5001: bind: address already in use
```
but you need to run the test with `-v`.

I am unsure how to use `netcat` to test for port conflict.  `netcat -w0 localhost 8080` does not work on my machine, I get "invalid wait-time 0".  Port testing doesn't seam like a documented use of `netcat` (at least the version (v1.10-41) I have installed on my machine) 

The sleep is 2 seconds now, I can lower it to 1 if you think it would be better.

whyrusleeping: You should be able to use `nc -z -w1 localhost 5001`

kevina: Do you still want to do that?  The test won't fail currently.

whyrusleeping: @kevina Hrm... i think we can probably do without for now. Lets add a todo to add a flag for different ports on the daemon

kevina: @whyrusleeping do you think we I should reduce the timeout to 1 sec to minimize the chance of this test failing?

whyrusleeping: I wouldnt worry about it, lets just make a note to do proper port selection for this later.

kevina:
I added a note as an issue.  This should be good to go...

PR #3806, opened 2017-03-21
commands: improve name and key helptexts

lgierth:
Adds the following ShortDescription to `ipfs key`:

'ipfs key gen' generates a new keypair for usage with IPNS and 'ipfs name publish'.

    > ipfs key gen --type=rsa --size=2048 mykey
    > ipfs name publish --key=mykey QmSomeHash

'ipfs key list' lists the available keys.

    > ipfs key list
    self
    mykey

### DIFF
--- a/core/commands/ipns.go
+++ b/core/commands/ipns.go
@@ -13,21 +13,24 @@ import (
 
 var IpnsCmd = &cmds.Command{
 	Helptext: cmds.HelpText{
-		Tagline: "Get the value currently published at an IPNS name.",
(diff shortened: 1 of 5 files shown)


Kubuxu:
but gitcop

lgierth:
(Changed title)

Fixed the signoff and added more helptext improvements around the IPNS multi-identity use case, for both `ipfs key` and `ipfs name`.

whyrusleeping:
@Kubuxu @VictorBjelkholm jenkins is upset

PR #3809, opened 2017-03-21
Fix #3783: Improve IsPinned() lookups for indirect pins

### LINKED ISSUE #3783: ipfs pin ls <not-pinned-hash> takes very long and uses lots of CPU
#### Description:

While `ipfs pin ls <pinned hash>` works fast. `ipfs pin ls <unpinned hash>` takes forever to inform the user that the hash is not pinned. I also observe around 5 secs 200% CPU usage until the request comes back. Tested with 0.4.5 and happens at least since then.

hsanjuan:
This avoids revisiting already-searched branches and cut down
the IsPinned() check times considerably when recursive pins
share big underlying DAGs. Same mechanism as `merkledag.EnumerateChildren()`.

Here are some benchmarks with a randomly-selected list of 10 of my pins (and bad pins):

This runs `ipfs pin ls $pin` for each of the listed pins and measures how long it takes it.

With 0.4.7:

```
Time for good pin QmSUuF4bnRBRAmaRzJcXHAu7z7tdD8sx1c5hU5kuvdPuDV: 0:00.55elapsed   0
Time for bad pin QmSUuF4bnRBRAmaRzJcXHAu7z7tdD8sx1c5hU5kuvdPuD1: 0:07.84elapsed   1
Time for good pin QmPcd2ypbzvh5ew6DBw9yCR8HBPWDJrLXVEw5DFntKWemk: 0:01.82elapsed   0
Time for bad pin QmPcd2ypbzvh5ew6DBw9yCR8HBPWDJrLXVEw5DFntKWem1: 0:07.19elapsed   1
Time for good pin QmYihqmU45xvKGBW1cGHCSF1Qxe5DTXzYqwFNEuBCvsR4Y: 0:00.23elapsed   0
Time for bad pin QmYihqmU45xvKGBW1cGHCSF1Qxe5DTXzYqwFNEuBCvsR41: 0:07.38elapsed   1
Time for good pin QmSP2Ef89DKFSrJi9T3b8bEeUA4fQSfR9AosKD7QEoXn3q: 0:00.31elapsed   0
Time for bad pin QmSP2Ef89DKFSrJi9T3b8bEeUA4fQSfR9AosKD7QEoXn31: 0:08.11elapsed   1
Time for good pin QmfJ2jZD4ULrbpJyoqxXnetckcwwHrGZMZfwdwKmfdZ4VE: 0:00.55elapsed   0
Time for bad pin QmfJ2jZD4ULrbpJyoqxXnetckcwwHrGZMZfwdwKmfdZ4V1: 0:08.26elapsed   1
```

With this patch:

```
Time for good pin QmSUuF4bnRBRAmaRzJcXHAu7z7tdD8sx1c5hU5kuvdPuDV: 0:00.69elapsed   0
Time for bad pin QmSUuF4bnRBRAmaRzJcXHAu7z7tdD8sx1c5hU5kuvdPuD1: 0:01.59elapsed   1
Time for good pin QmPcd2ypbzvh5ew6DBw9yCR8HBPWDJrLXVEw5DFntKWemk: 0:01.59elapsed   0
Time for bad pin QmPcd2ypbzvh5ew6DBw9yCR8HBPWDJrLXVEw5DFntKWem1: 0:01.80elapsed   1
Time for good pin QmYihqmU45xvKGBW1cGHCSF1Qxe5DTXzYqwFNEuBCvsR4Y: 0:00.31elapsed   0
Time for bad pin QmYihqmU45xvKGBW1cGHCSF1Qxe5DTXzYqwFNEuBCvsR41: 0:01.82elapsed   1
Time for good pin QmSP2Ef89DKFSrJi9T3b8bEeUA4fQSfR9AosKD7QEoXn3q: 0:00.40elapsed   0
Time for bad pin QmSP2Ef89DKFSrJi9T3b8bEeUA4fQSfR9AosKD7QEoXn31: 0:01.80elapsed   1
Time for good pin QmfJ2jZD4ULrbpJyoqxXnetckcwwHrGZMZfwdwKmfdZ4VE: 0:00.61elapsed   0
Time for bad pin QmfJ2jZD4ULrbpJyoqxXnetckcwwHrGZMZfwdwKmfdZ4V1: 0:01.78elapsed   1
```

### DIFF
--- a/pin/pin.go
+++ b/pin/pin.go
@@ -276,8 +276,9 @@ func (p *pinner) isPinnedWithType(c *cid.Cid, mode PinMode) (string, bool, error
 	}
 
 	// Default is Indirect
+	visitedSet := cid.NewSet()
 	for _, rc := range p.recursePin.Keys() {
-		has, err := hasChild(p.dserv, rc, c)
+		has, err := hasChild(p.dserv, rc, c, visitedSet.Visit)
 		if err != nil {
 			return "", false, err
 		}
@@ -519,7 +520,9 @@ func (p *pinner) PinWithMode(c *cid.Cid, mode PinMode) {
 	}
 }
 
-func hasChild(ds mdag.LinkService, root *cid.Cid, child *cid.Cid) (bool, error) {
+// hasChild recursively looks for a Cid among the children of a root Cid.
+// The visit function can be used to shortcut already-visited branches.
+func hasChild(ds mdag.LinkService, root *cid.Cid, child *cid.Cid, visit func(*cid.Cid) bool) (bool, error) {
 	links, err := ds.GetLinks(context.Background(), root)
 	if err != nil {
 		return false, err
@@ -529,14 +532,15 @@ func hasChild(ds mdag.LinkService, root *cid.Cid, child *cid.Cid) (bool, error) 
 		if lnk.Cid.Equals(child) {
 			return true, nil
 		}
+		if visit(c) {
+			has, err := hasChild(ds, c, child, visit)
+			if err != nil {
+				return false, err
+			}
 
-		has, err := hasChild(ds, c, child)
-		if err != nil {
-			return false, err
-		}
-
-		if has {
-			return has, nil
+			if has {
+				return has, nil
+			}
 		}
 	}
 	return false, nil
--- a/pin/pin_test.go
+++ b/pin/pin_test.go
@@ -35,6 +35,17 @@ func assertPinned(t *testing.T, p Pinner, c *cid.Cid, failmsg string) {
 	}
 }
 
+func assertUnpinned(t *testing.T, p Pinner, c *cid.Cid, failmsg string) {
+	_, pinned, err := p.IsPinned(c)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if pinned {
+		t.Fatal(failmsg)
+	}
+}
+
 func TestPinnerBasic(t *testing.T) {
 	ctx := context.Background()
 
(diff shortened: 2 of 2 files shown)


pin/pin.go
```diff
@@ -276,8 +276,9 @@ func (p *pinner) isPinnedWithType(c *cid.Cid, mode PinMode) (string, bool, error
 	}
 
 	// Default is Indirect
+	visitedSet := cid.NewSet()
 	for _, rc := range p.recursePin.Keys() {
-		has, err := hasChild(p.dserv, rc, c)
+		has, err := hasChild(p.dserv, rc, c, visitedSet.Visit)
```
whyrusleeping: I dont think we can reuse the same `visitedSet` across pin checks, unless we pass all the pins we're checking into the call at once. otherwise we might mark a subtree as visited that contains the next pin we're checking

whyrusleeping: Oh wait a second, i misread. 'c' is the thing we're looking for, and it remains the same throughout the entire operation. This LGTM then

pin/pin.go
```diff
@@ -519,7 +520,9 @@ func (p *pinner) PinWithMode(c *cid.Cid, mode PinMode) {
 	}
 }
 
-func hasChild(ds mdag.LinkService, root *cid.Cid, child *cid.Cid) (bool, error) {
+// hasChild recursively looks for a Cid among the children of a root Cid.
+// The visit function can be used to shortcut already-visited branches.
+func hasChild(ds mdag.LinkService, root *cid.Cid, child *cid.Cid, visit func(*cid.Cid) bool) (bool, error) {
```
hoenirvili: Why not make a type with the signature of ```go func(*cid.Cid) bool``` ?

hsanjuan: Because it creates an extra indirection which does not help to figure out what this is doing any more than having the definition directly as parameter. Moreover, this is the only parameter which uses that in the module, it's not hard to read and it is called from a single place which clearly shows that expected usage is to pass cid.Set.Visit function.

pin/pin.go
hsanjuan: @whyrusleeping well, don't you need to revoke the "request changes" review or something? I have no idea if that is possible though.

hsanjuan:
I'll add an extra test with a shared tree

pin/pin.go
hoenirvili: At least change the **hasChild** func format in someting like this.
```go
func hasChild(
ds mdag.LinkService,
root *cid.Cid,
child *cid.Cid,
visit func(*cid.Cid) bool,
) (bool, error) {
//body
}
```

This should be documented somewhere to enforce a number of columns a func can have.

pin/pin.go
whyrusleeping: Hrm... i think it wants me to submit another review with 'changed approved'

pin/pin.go
kevina: I do not believe this is done anywhere else in the IPFS code base.

whyrusleeping: Yeah, in general I don't mind having the function declaration be long. Simply breaking at the most convenient place onto the next line is fine. Breaking it out into separate lines per parameter brings back unpleasant memories of really really old C code.

hsanjuan: I don't like going over 80 chars, but if the codebase style keeps declarations in the same line, I'll leave it as it is for the moment because I like consistency.

hsanjuan:
@whyrusleeping @hoenirvili I have added an extra test.

pin/pin_test.go
```diff
@@ -143,6 +154,122 @@ func TestPinnerBasic(t *testing.T) {
...
+
+	// Pin A5 recursively
+	err := p.Pin(ctx, aNodes[aBranchLen-1], true)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Create node B and add A3 as child
+	b, _ := randNode()
+	err = b.AddNodeLink("mychild", aNodes[3])
```
hoenirvili: Can you make all error checks one line like ``` if err := func(); err != nil {  //code ```

hsanjuan: fixed

pin/pin_test.go
```diff
@@ -145,6 +151,121 @@ func TestPinnerBasic(t *testing.T) {
...
+		t.Fatal(err)
+	}
+
+	assertPinned(t, p, aKeys[0], "A0 should be pinned")
+	assertPinned(t, p, aKeys[1], "A1 should be pinned")
+	assertPinned(t, p, ck, "C should be pinned")
+	assertPinned(t, p, bk, "B should be pinned")
+
+	// Unpin A5 recursively
+	err = p.Unpin(ctx, aKeys[5], true)
```
whyrusleeping: check this error (and the one below)

hsanjuan: fixed!

pin/pin_test.go
whyrusleeping: @hsanjuan I think you changed the error checks in the wrong test, lol

whyrusleeping:
@hsanjuan One more comment about changing the error checking in the wrong test

hsanjuan:
@whyrusleeping sorry! I have fixed it. Honestly hope I haven't broken something else.

PR #3815, opened 2017-03-23
filestore rm

kevina:
Factored out of #3653


whyrusleeping:
Would the recommended way of deduping between filestore and blockstore be `ipfs filestore dups | ipfs filestore rm` ?

kevina:
> Would the recommended way of deduping between filestore and blockstore be ipfs filestore dups | ipfs filestore rm ?

More of less.

filestore/remove.go
```diff
@@ -0,0 +1,89 @@
...
+	out := make(chan interface{}, len(cids))
+
+	var blocks Deleter
+	switch opts.Prefix {
+	case FilestorePrefix.String():
+		blocks = fs.fm
+	case bs.BlockPrefix.String():
+		blocks = fs.bs
+	default:
+		return nil, fmt.Errorf("Unknown prefix: %s\n", opts.Prefix)
```
Kubuxu: Errors should start with lower case.

radfish:
> Would the recommended way of deduping between filestore and blockstore be ipfs filestore dups | ipfs filestore rm ?

Would `ipfs filestore dups | ipfs block rm` also work (to keep in the filestore, but delete  from blockstore)?

kevina:
> Would ipfs filestore dups | ipfs block rm also work (to keep in the filestore, but delete from blockstore)?

@radfish no that will remove it from both the blockstore and filestore.  With this p.r. to remove from the filestore use `ipfs filestore rm` to remove *just* the blockstore use `ipfs filestore rm --non-filestore`.

I am not against adding a `--just-blockstore` or similar option to `ipfs block rm` as that seams more intuitive for me.

PR #3820, opened 2017-03-23
filestore util: Use a Marshaler instead of PostRun...

kevina:
and just output directly to Stderr and Stdout instead of returning
a reader.


kevina:
@whyrusleeping I messed up the rebase of #3653.  For some reason I thought this commit was for `filestore rm`.  I was wrong.

PR #3827, opened 2017-03-24
Make Golint happy in the blocks submodule.

hsanjuan:
This has required changing the order of some parameters and
adding HashOnRead to the Blockstore interface (which I have in turn
added to all the wrapper implementations).

I plan to do more, but let's merge little by little, submodule by submodule.

### DIFF
--- a/blocks/blocks.go
+++ b/blocks/blocks.go
@@ -1,5 +1,6 @@
-// package blocks contains the lowest level of IPFS data structures,
-// the raw block with a checksum.
+// Package blocks contains the lowest level of IPFS data structures.
+// A block is raw data accompanied by a CID. The CID contains the multihash
+// corresponding to the block.
 package blocks
 
 import (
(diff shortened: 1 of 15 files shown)


hsanjuan:
codeclimate is a troll

blocks/blocks.go
```diff
@@ -11,16 +12,20 @@ import (
 	mh "gx/ipfs/QmbZ6Cee2uHjG7hf19qLHppgKDRtaG4CVtMzdmK9VCVqLu/go-multihash"
 )
 
-var ErrWrongHash = errors.New("data did not match given hash!")
+// ErrWrongHash is returned when the Cid of a block is not the expected
+// according to the contents. It is currently used only when debugging.
```
Kubuxu: I think it is also used when HashOnRead is on.

hsanjuan: That uses `ErrHashMismatch = errors.New("block in storage has different hash than requested")` in the blockstore module...

Kubuxu: Ok, let's leave it as it is we might want to merge those errors in future. I must have not noticed it when I were writing the HashOnRead code.

Issue #3837, opened 2017-03-27
ipfs version --commit no longer shows githash


eminence:
#### Version information: 
```
$ ipfs version --all
go-ipfs version: 0.4.8-rc1-
Repo version: 5
System version: amd64/linux
Golang version: go1.7.1
```

#### Description:
Running `ipfs version --commit` or `ipfs version --all` no longer shows the git hash.  In my case, I am building ipfs from source by running "make install" from my `$GOPATH/src/github.com/ipfs/go-ipfs` directory.  See above for an example.  I bisected this, and this problem was introduced by f630222580ffaa9e5c62610773f391a8c6416556

Kubuxu:
Works here.
```
 cmd/ipfs/ipfs version --commit
ipfs version 0.4.8-rc1-a8b56d3ad
```

f630222 introduced it but it was latter resolved in same PR  (iirc). As a note, builds from dist.ipfs.io don't have the hash.

lgierth:
> As a note, builds from dist.ipfs.io don't have the hash.

Why? They used to

Kubuxu:
I don't think they ever had a hash.

eminence:
Here's a more complete transcript on my build session:

```
achin@bigbox ~/devel/gopath/src/github.com/ipfs/go-ipfs $ git clean -dfx
Skipping repository test/sharness/lib/sharness/

achin@bigbox ~/devel/gopath/src/github.com/ipfs/go-ipfs $ git show -s |cat
commit a8b56d3ad7789ca2ec12257a00d8d0408c3c0c9e
Author: Jeromy <jeromyj@gmail.com>
Date:   Fri Mar 24 20:34:49 2017 -0700

    Ipfs 0.4.8-rc1

    License: MIT
    Signed-off-by: Jeromy <jeromyj@gmail.com>



achin@bigbox ~/devel/gopath/src/github.com/ipfs/go-ipfs $ make install
bin/check_go_version 1.7
bin/check_go_path /storage/home/achin/devel/gopath/src/github.com/ipfs/go-ipfs /storage/home/achin/devel/gopath/src/github.com/ipfs/go-ipfs
installing gx v0.11.0
bin/dist_get /ipfs/QmNjXP6N98fYT1i7abgeBHmdR5WoeBeik4DtGiX9iFWK31 gx bin/gx-v0.11.0 v0.11.0
[...]

PR #3838, opened 2017-03-27
make: fix `make install` not using ldflags for git hash

Kubuxu:
Resolves https://github.com/ipfs/go-ipfs/issues/3837

License: MIT

### DIFF
--- a/Rules.mk
+++ b/Rules.mk
@@ -99,8 +99,7 @@ nofuse: GOTAGS += nofuse
 nofuse: build
 .PHONY: nofuse
 
-install: $$(DEPS_GO)
-	go install $(go-flags-with-tags) ./cmd/ipfs
+install: cmd/ipfs-install
 .PHONY: install
 
 install_unsupported:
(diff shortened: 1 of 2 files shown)


whyrusleeping:
@Kubuxu could we perhaps have a test for this?

Kubuxu:
Not sure how to test it, it uses buildsystem and at the same time it installs into GOPATH, it would be hard.

whyrusleeping:
A sharness test should be able to just check `ipfs version --commit` no?

Kubuxu:
This affects `make install` not `make build`. We test for commit in docker tests, not sure if they run.

I will add test for commit, but it will find regression with `build` not `install` and I can already tell that it won't pass with coverage collection (as coverage doesn't insert this variable either).

whyrusleeping:
@Kubuxu how about we default the variable to something that indicates its not an 'official' build?

Kubuxu:
So do we want to confront Semver?

Then I suggest:
Official `0.4.8-rc1`
Unofficial if not git var available `0.4.8-rc1+sometag` IDK what to place there there.
With git: `0.4.8-rc1+git.a8b56d3ad`

whyrusleeping:
those SGTM. The 'sometag' is generally 'dirty'

Kubuxu:
We can also add `+dirty` if there is git hash but there are local (not committed changes)

whyrusleeping:
@Kubuxu is this ready to go?

Kubuxu:
If tests are to be separate, yes.

PR #3846, opened 2017-03-31
implement ipfs pin update

whyrusleeping:
Makes re-pinning a slightly changed dataset much faster

License: MIT

### DIFF
--- a/core/commands/pin.go
+++ b/core/commands/pin.go
@@ -24,9 +24,10 @@ var PinCmd = &cmds.Command{
 	},
 
 	Subcommands: map[string]*cmds.Command{
-		"add": addPinCmd,
-		"rm":  rmPinCmd,
-		"ls":  listPinCmd,
+		"add":    addPinCmd,
+		"rm":     rmPinCmd,
+		"ls":     listPinCmd,
+		"update": updatePinCmd,
 	},
 }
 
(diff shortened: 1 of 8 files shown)


whyrusleeping:
I'll want a solid review of the algorithm here

merkledag/utils/diffenum.go
```diff
@@ -0,0 +1,75 @@
...
+	if err != nil {
+		return fmt.Errorf("get %s: %s", to, err)
+	}
+
+	diff := getLinkDiff(fnd, tnd)
+
+	sset := cid.NewSet()
+	for _, c := range diff {
+		if c.a == nil {
+			err := mdag.EnumerateChildrenAsync(ctx, mdag.GetLinksDirect(dserv), c.b, sset.Visit)
```
Kubuxu: This line is really long

merkledag/utils/diffenum.go
```diff
@@ -0,0 +1,75 @@
...
+				return err
+			}
+		}
+	}
+
+	return nil
+}
+
+type diffpair struct {
+	a, b *cid.Cid
```
Kubuxu: I would rename it to `old, new`. It clears up a lot.

merkledag/utils/diffenum.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+func getLinkDiff(a, b node.Node) []diffpair {
```
Kubuxu: It took me a minute to properly get what this function does but now I understand.

It diffs links as a set, it doesn't matter for it if some link is repeated.

merkledag/utils/diffenum.go
```diff
@@ -0,0 +1,75 @@
...
+
+	sset := cid.NewSet()
+	for _, c := range diff {
+		if c.a == nil {
+			err := mdag.EnumerateChildrenAsync(ctx, mdag.GetLinksDirect(dserv), c.b, sset.Visit)
+			if err != nil {
+				return err
+			}
+		} else {
+			err := DiffEnumerate(ctx, dserv, c.a, c.b)
```
Kubuxu: Are we worried here about too deep recursion? Should it be turned into loop with TODO list instead?

Kubuxu: Also coverage shows that this isn't tested at all.

core/commands/pin.go
```diff
@@ -332,6 +333,83 @@ Example:
...
+new pin and removing the old one.
+`,
+	},
+
+	Arguments: []cmds.Argument{
+		cmds.StringArg("from-path", true, false, "Path to old object."),
+		cmds.StringArg("to-path", true, false, "Path to new object to be pinned."),
+	},
+	Options: []cmds.Option{
+		cmds.BoolOption("unpin", "Remove the old pin.").Default(true),
```
whyrusleeping: Its if you want to use the optimized pinning functionality and still keep the old pin

merkledag/utils/diffenum.go
whyrusleeping: It is really tho?

core/commands/pin.go
Kubuxu: Ahh, right, because it will piggy bank on the already pinned hashes.

kevina:
@whyrusleeping I would like to review this but it may take up to week before I can get to it.

merkledag/utils/diffenum.go
whyrusleeping: cant use 'new' lol, its a keyword

Kubuxu: ouch, annoying

merkledag/utils/diffenum.go
Kubuxu: aka. too wide for my vim setup.

merkledag/utils/diffenum.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	return nil
+}
+
+type diffpair struct {
+	bef, aft *cid.Cid
+}
+
+// getLinkDiff returns a changset (minimum edit distance style) between nodes
```
kevina: I was rather confused by what this is doing.  From the code it is fairly clear but the phrase "minimum edit distance style" confused me.  As someone who wrote a spell checker I was expecting a traditional edit dist algo, and was very confused my the use of the maps.  I would probably have had an easier time understanding what is going on if I wasn't trying to reconcile the code with what a traditional string edit dist. algo would be doing.

pin/pin.go
```diff
@@ -86,6 +87,11 @@ type Pinner interface {
 	Pin(context.Context, node.Node, bool) error
 	Unpin(context.Context, *cid.Cid, bool) error
 
+	// Update updates a recursive pin from one cid to another
+	// this is more efficient than simply pinning the new one and unpinning the
+	// old one
+	Update(context.Context, *cid.Cid, *cid.Cid, bool) error
```
kevina: I would name the parameters here.  The use of the bool in Update differs from the use in pin and unpin.

PR #3848, opened 2017-03-31
make: fix `make install` not installing dependencies

Kubuxu:
License: MIT


Kubuxu:
Travis uses it but we ignored the problem as Travis fails frequently.

Kubuxu:
This was new change made in a bit of a rush.

PR #3850, opened 2017-04-02
commands: restore original error order

keks:
When `ErrNotFound` was added, it was not appended to the iota but inserted in the middle. This means that the numeric value of `ErrImplementation` changed, resulting in possible API breakage. This PR fixes the order.

### DIFF
--- a/commands/response.go
+++ b/commands/response.go
@@ -17,8 +17,8 @@ type ErrorType uint
 const (
 	ErrNormal         ErrorType = iota // general errors
 	ErrClient                          // error was caused by the client, (e.g. invalid CLI usage)
-	ErrNotFound                        // == HTTP 404 Not Found
 	ErrImplementation                  // programmer error in the server
+	ErrNotFound                        // == HTTP 404 Not Found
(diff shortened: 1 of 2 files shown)


keks:
Some more context from #ipfs-dev:

---
```
keks
shouldn't the codes of these errors be static? https://github.com/ipfs/go-ipfs/blob/2cc5ce45b875e45fdbfc779e1a7c4bef99bb9877/commands/response.go#L17
otherwise different versions interprete the result differently
and before ErrImpl was 2, not 3

lgierth (IRC)
keks[m]1: the commands stuff is not currently a public api
ah wait, mh
the line between http api and the commands specifics is a bit blurry at times
i agree that these error codes shouldn't change without a good reason
damn that was me
wanna change it back? i'm about to head to the airport
```

whyrusleeping:
:+1: @keks would you mind writing a test or two for this? Just so we don't accidentally make the same mistake? You can even just make a go test that checks the values of the constants (Or alternatively, explicitly assign the constant values instead of using iota)

keks:
@whyrusleeping done!

PR #3854, opened 2017-04-04
Add FOSSA badge/reports

xizhao:


jbenet:
hey @xizhao ! thanks for pushing this. glad Fossa is out.

- Can you make sure people can view the results without logging in? When i click the image it asks me to log in...
- Can you also make sure to allow people to log in with github and not ask for such stringent permissions? no way im giving R access to all my repos. and certainly no way i'm giving W access to any. There's serious opsec concerns here.

Thanks

jbenet:
Also-- would be great if you keep the original license statement ("MIT"), and we test it first in go-ipfs before deploying org-wide.

xizhao:
Sounds good.

1. We are pushing more public views which will be available without auth, but it will take a few releases before those are generally available.  We're thinking we render the public page as a summary of results, and you'd have to log in to look at it in full as a maintainer.  

2. Currently, the Github API is limited so that you can't get read-only repo scopes for private repos: https://developer.github.com/v3/oauth/#scopes.  Unfortunately, we need code access in order to scan for licenses and deps.  We eventually can move to the beta integrations API where more fine-grained permissions are possible.  

Since IPFS repos are public, we don't need full github access to do daily scans, but if you want per-commit scans and the ability to block PRs, there needs to be a full Github login.

Is this enough to move forward?

whyrusleeping:
Yeah, I'd like to have the no-login pages as @jbenet requested, as well as no requirement to give out my github creds

xizhao:
Sounds good, thanks for the feedback @jbenet & @whyrusleeping! 

We'll push some updates out and I'll follow up soon.

xizhao:
Hi folks, a quick update.  We just release public accessibility for open source projects in FOSSA -- thanks for the feedback on this!

[go-ipfs](https://app.fossa.io/projects/git%2Bhttps%3A%2F%2Fgithub.com%2Fipfs%2Fgo-ipfs) should now be browsable without login on FOSSA!

Kubuxu:
@xizhao would it be possible to have two stages of auth. One for public projects which then can be expanded to privates repos. Giving access to 3rd party to all of my private repos is no-go for me.

xizhao:
@Kubuxu, unfortunately, we can't ask for varying types of auth from different users.  I'd say the best workaround is you can create a throwaway account and invite it to access only your public projects.

Another workaround is you can log in, import your public projects, then revoke your access token in Github.  This will cause future actions that require your auth to fail, but should not give FOSSA access to any private repos you haven't imported.

xizhao:
Checking in for updates, any blockers on this @Kubuxu?

The JSF and projects like Webpack are now running on FOSSA: https://github.com/webpack/webpack

xizhao:
Hi folks, exciting update for you all!

We just migrated to the new Github installations API -- this new API allows us to ask for read-only access scope for repos now as well as let you select which repos FOSSA can access.

@Kubuxu -- If you login with Github it should now ask you to auth without granting permissions and (once you import) ask for only read access on code -- try it out!

Kevin

Stebalien:
This is rather old and appears to have stagnated. We can pick this back up if someone with the right context wants to.

Issue #3786, opened 2017-03-15
Bitswap Sessions


whyrusleeping:
Bitswap is pretty inefficient right now. There is a *lot* we can do to make it better (i would say it would be hard to make it worse, but i know for a fact that statement is false). I've spent a lot of time thinking of ways forward, and have come to a decent (in my opinion) 'next step'. 

This proposal is to add the concept of sessions to bitswap. The idea expands on an existing optimization we do, where we only find providers for the first block in a `GetMany` request, and make more `FindProviders` calls later if nobody is sending us blocks. 

When starting a request for a portion of a graph (`ipfs refs -r`, `ipfs pin`, `ipfs cat`, etc) we would create a new session with the hash in question as the session 'root'. Bitswap would then broadcast that ref to every bitswap partner. Any node that responds with the block (even if its a duplicate) gets added to that sessions 'active set'. Further hashes requested that are children of the session root will be broadcasted just to those peers (we can even split up requests amongst peers). If a requested hash isnt received in a given time frame, we start a `FindProviders` call for that hash, and adds the resultant peers to the sessions 'active set' (or perhaps add them to a 'to try set' and to the 'active set' when a block is received from them).

Doing things this way will allow us to not spam uninterested bitswap peers with our wantlist updates, and also allow us to reduce duplicate blocks by splitting up request amongst peers we know are likely to respond.

Feedback and requests for elaboration will be much appreciated.

lgierth:
If we already assume that the "active set" peers likely have all children of the wanted root, couldn't we tell them to "consider all children of $hash wanted by me"? This would be a super light wantlist update :)

And in case they don't have all children of the wanted root, we can adjust reactively just like you're saying. If we don't receive any more wanted blocks from the session (all children considered wanted), then we remove it from the active set, and try `FindProviders`.

whyrusleeping:
@lgierth That should be possible later with ipld selectors, but for now it would be a breaking change to the protocol (and would make the duplicate blocks problem likely worse)

btc:
Is the duplicate blocks problem the primary issue? Are there other considerations to weigh?

(Writing from an uninformed perspective and curious about the scalability challenges.)

whyrusleeping:
[...]

PR #3867, opened 2017-04-17
Bitswap sessions

whyrusleeping:
This is an implementation of the bitswap sessions enhancement i proposed in https://github.com/ipfs/go-ipfs/issues/3786

Its still a work in progress, and has a few potential issues around tracking of wantlist refcounts in failure modes (if peers don't have any of the things we want and we re-broadcast, it increases the wantlist refcounts unnecessarily), but it generally works quite well and makes bitswap faster and have much less bandwidth overhead. In cases of transferring large files just between one or two peers, i've seen over a 2x speed improvement.

### DIFF
--- a/blockservice/blockservice.go
+++ b/blockservice/blockservice.go
@@ -32,7 +32,6 @@ type BlockService interface {
 	GetBlock(ctx context.Context, c *cid.Cid) (blocks.Block, error)
 	GetBlocks(ctx context.Context, ks []*cid.Cid) <-chan blocks.Block
 	DeleteBlock(o blocks.Block) error
-	NewSession(context.Context) *Session
 	Close() error
 }
 
@@ -79,18 +78,18 @@ func (bs *blockService) Exchange() exchange.Interface {
 	return bs.exchange
 }
 
-func (bs *blockService) NewSession(ctx context.Context) *Session {
-	bswap, ok := bs.Exchange().(*bitswap.Bitswap)
-	if ok {
+func NewSession(ctx context.Context, bs BlockService) *Session {
+	exchange := bs.Exchange()
+	if bswap, ok := exchange.(*bitswap.Bitswap); ok {
 		ses := bswap.NewSession(ctx)
 		return &Session{
 			ses: ses,
-			bs:  bs.blockstore,
+			bs:  bs.Blockstore(),
 		}
 	}
 	return &Session{
-		ses: bs.exchange,
-		bs:  bs.blockstore,
+		ses: exchange,
+		bs:  bs.Blockstore(),
 	}
 }
 
--- a/blockservice/blockservice.go
+++ b/blockservice/blockservice.go
@@ -251,15 +251,18 @@ func (s *blockService) Close() error {
 	return s.exchange.Close()
 }
 
+// Session is a helper type to provide higher level access to bitswap sessions
 type Session struct {
 	bs  blockstore.Blockstore
 	ses exchange.Fetcher
 }
 
+// GetBlock gets a block in the context of a request session
 func (s *Session) GetBlock(ctx context.Context, c *cid.Cid) (blocks.Block, error) {
 	return getBlock(ctx, c, s.bs, s.ses)
 }
 
+// GetBlocks gets blocks in the context of a request session
 func (s *Session) GetBlocks(ctx context.Context, ks []*cid.Cid) <-chan blocks.Block {
 	return getBlocks(ctx, ks, s.bs, s.ses)
 }
--- a/blockservice/blockservice.go
+++ b/blockservice/blockservice.go
@@ -10,9 +10,10 @@ import (
 
 	"github.com/ipfs/go-ipfs/blocks/blockstore"
 	exchange "github.com/ipfs/go-ipfs/exchange"
-	blocks "gx/ipfs/QmXxGS5QsUxpR3iqL5DjmsYPHR1Yz74siRQ4ChJqWFosMh/go-block-format"
+	bitswap "github.com/ipfs/go-ipfs/exchange/bitswap"
 
 	logging "gx/ipfs/QmSpJByNKFX1sCsHBEp3R73FL4NF6FnQTEGyNAXHm2GS52/go-log"
+	blocks "gx/ipfs/QmXxGS5QsUxpR3iqL5DjmsYPHR1Yz74siRQ4ChJqWFosMh/go-block-format"
 	cid "gx/ipfs/Qma4RJSuh7mMeJQYCqMbKzekn6EwBo7HEs5AQYjVRMQATB/go-cid"
 )
 
@@ -31,6 +32,7 @@ type BlockService interface {
 	GetBlock(ctx context.Context, c *cid.Cid) (blocks.Block, error)
 	GetBlocks(ctx context.Context, ks []*cid.Cid) <-chan blocks.Block
 	DeleteBlock(o blocks.Block) error
+	NewSession(context.Context) *Session
 	Close() error
 }
 
@@ -75,6 +77,21 @@ func (bs *blockService) Blockstore() blockstore.Blockstore {
 
 func (bs *blockService) Exchange() exchange.Interface {
 	return bs.exchange
+}
+
+func (bs *blockService) NewSession(ctx context.Context) *Session {
+	bswap, ok := bs.Exchange().(*bitswap.Bitswap)
+	if ok {
+		ses := bswap.NewSession(ctx)
+		return &Session{
+			ses: ses,
+			bs:  bs.blockstore,
+		}
+	}
+	return &Session{
+		ses: bs.exchange,
+		bs:  bs.blockstore,
+	}
 }
 
 // AddBlock adds a particular block to the service, Putting it into the datastore.
@@ -141,16 +158,25 @@ func (s *blockService) AddBlocks(bs []blocks.Block) ([]*cid.Cid, error) {
 func (s *blockService) GetBlock(ctx context.Context, c *cid.Cid) (blocks.Block, error) {
 	log.Debugf("BlockService GetBlock: '%s'", c)
 
-	block, err := s.blockstore.Get(c)
+	var f exchange.Fetcher
+	if s.exchange != nil {
+		f = s.exchange
+	}
+
+	return getBlock(ctx, c, s.blockstore, f)
+}
+
+func getBlock(ctx context.Context, c *cid.Cid, bs blockstore.Blockstore, f exchange.Fetcher) (blocks.Block, error) {
+	block, err := bs.Get(c)
 	if err == nil {
 		return block, nil
 	}
 
-	if err == blockstore.ErrNotFound && s.exchange != nil {
+	if err == blockstore.ErrNotFound && f != nil {
 		// TODO be careful checking ErrNotFound. If the underlying
 		// implementation changes, this will break.
 		log.Debug("Blockservice: Searching bitswap")
-		blk, err := s.exchange.GetBlock(ctx, c)
+		blk, err := f.GetBlock(ctx, c)
 		if err != nil {
 			if err == blockstore.ErrNotFound {
 				return nil, ErrNotFound
@@ -172,12 +198,16 @@ func (s *blockService) GetBlock(ctx context.Context, c *cid.Cid) (blocks.Block, 
 // the returned channel.
 // NB: No guarantees are made about order.
 func (s *blockService) GetBlocks(ctx context.Context, ks []*cid.Cid) <-chan blocks.Block {
+	return getBlocks(ctx, ks, s.blockstore, s.exchange)
+}
+
+func getBlocks(ctx context.Context, ks []*cid.Cid, bs blockstore.Blockstore, f exchange.Fetcher) <-chan blocks.Block {
 	out := make(chan blocks.Block)
 	go func() {
 		defer close(out)
 		var misses []*cid.Cid
 		for _, c := range ks {
-			hit, err := s.blockstore.Get(c)
+			hit, err := bs.Get(c)
 			if err != nil {
 				misses = append(misses, c)
 				continue
@@ -194,7 +224,7 @@ func (s *blockService) GetBlocks(ctx context.Context, ks []*cid.Cid) <-chan bloc
 			return
 		}
 
-		rblocks, err := s.exchange.GetBlocks(ctx, misses)
+		rblocks, err := f.GetBlocks(ctx, misses)
 		if err != nil {
 			log.Debugf("Error with GetBlocks: %s", err)
 			return
@@ -219,4 +249,17 @@ func (s *blockService) DeleteBlock(o blocks.Block) error {
 func (s *blockService) Close() error {
 	log.Debug("blockservice is shutting down...")
 	return s.exchange.Close()
+}
+
+type Session struct {
+	bs  blockstore.Blockstore
+	ses exchange.Fetcher
+}
+
+func (s *Session) GetBlock(ctx context.Context, c *cid.Cid) (blocks.Block, error) {
+	return getBlock(ctx, c, s.bs, s.ses)
+}
+
+func (s *Session) GetBlocks(ctx context.Context, ks []*cid.Cid) <-chan blocks.Block {
+	return getBlocks(ctx, ks, s.bs, s.ses)
 }
--- a/blockservice/blockservice.go
+++ b/blockservice/blockservice.go
@@ -78,6 +78,8 @@ func (bs *blockService) Exchange() exchange.Interface {
 	return bs.exchange
 }
 
+// NewSession creates a bitswap session that allows for controlled exchange of
+// wantlists to decrease the bandwidth overhead.
 func NewSession(ctx context.Context, bs BlockService) *Session {
 	exchange := bs.Exchange()
 	if bswap, ok := exchange.(*bitswap.Bitswap); ok {
--- a/core/commands/bitswap.go
+++ b/core/commands/bitswap.go
@@ -64,7 +64,11 @@ var unwantCmd = &cmds.Command{
 			ks = append(ks, c)
 		}
 
-		bs.CancelWants(ks)
+		// TODO: This should maybe find *all* sessions for this request and cancel them?
+		// (why): in reality, i think this command should be removed. Its
+		// messing with the internal state of bitswap. You should cancel wants
+		// by killing the command that caused the want.
+		bs.CancelWants(ks, 0)
 	},
 }
 
--- a/core/commands/bitswap.go
+++ b/core/commands/bitswap.go
@@ -111,6 +111,11 @@ Print out all blocks currently on the bitswap wantlist for the local peer.`,
 				res.SetError(err, cmds.ErrNormal)
 				return
 			}
+			if pid == nd.Identity {
+				res.SetOutput(&KeyList{bs.GetWantlist()})
+				return
+			}
+
 			res.SetOutput(&KeyList{bs.WantlistForPeer(pid)})
 		} else {
 			res.SetOutput(&KeyList{bs.GetWantlist()})
--- a/exchange/bitswap/bitswap.go
+++ b/exchange/bitswap/bitswap.go
@@ -169,6 +169,9 @@ type Bitswap struct {
 	// Sessions
 	sessions []*Session
 	sessLk   sync.Mutex
+
+	sessID   uint64
+	sessIDLk sync.Mutex
 }
 
 type blockRequest struct {
@@ -219,7 +222,9 @@ func (bs *Bitswap) GetBlocks(ctx context.Context, keys []*cid.Cid) (<-chan block
 		log.Event(ctx, "Bitswap.GetBlockRequest.Start", k)
 	}
 
-	bs.wm.WantBlocks(ctx, keys, nil)
+	mses := bs.getNextSessionID()
+
+	bs.wm.WantBlocks(ctx, keys, nil, mses)
 
 	// NB: Optimization. Assumes that providers of key[0] are likely to
 	// be able to provide for all keys. This currently holds true in most
@@ -241,7 +246,7 @@ func (bs *Bitswap) GetBlocks(ctx context.Context, keys []*cid.Cid) (<-chan block
 		defer close(out)
 		defer func() {
 			// can't just defer this call on its own, arguments are resolved *when* the defer is created
-			bs.CancelWants(remaining.Keys())
+			bs.CancelWants(remaining.Keys(), mses)
 		}()
 		for {
 			select {
@@ -250,6 +255,7 @@ func (bs *Bitswap) GetBlocks(ctx context.Context, keys []*cid.Cid) (<-chan block
 					return
 				}
 
+				bs.CancelWants([]*cid.Cid{blk.Cid()}, mses)
 				remaining.Remove(blk.Cid())
 				select {
 				case out <- blk:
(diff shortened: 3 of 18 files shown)


whyrusleeping:
Review of the code and algorithms involved are wanted, but also desired are test suggestions

kevina:
@whyrusleeping, due to the move it may be a week or so before I can get to this.

whyrusleeping:
@kevina no worries, i'm a bit out of it this week too.

hsanjuan:
Assigned myself so I remember to look at this

whyrusleeping:
next steps:

Have wantlists not just be refcounted, instead, actually track which session IDs are actively requesting them. This way we avoid weirdness when one session says 'want X' multiple times and only cancels once.

whyrusleeping:
This is less of a WIP now, i've got the problem i was worried about resolved. please review!

eminence:
Currently running f39ac4ef14eb94 and encountered some things of note while trying to run `ipfs get` on a large 60GB dataset.

First issue was an exceedingly large number of "too many open files" messages.  See [log here](http://ipfs.io/ipfs/QmYdYHPGVQ1iHB4Wo1qvPiLGgW2MfhV2PDAiQiww5YPhow).

After restarting the daemon, this problem seemed to go away.  I'm see [far fewer](http://ipfs.io/ipfs/QmZMGBKzhYwvcPCiHGKRnWoviAhPCY7uhWyVhd129yaCLm) of these messages.  However, I observed that my bitswap wantlist length kept growing.  

So I killed the `ipfs get`, and an hour later, I still have a large wantlist.  Its length hasn't gone down at all:

```
achin@bigbox ~/devel/gopath/src/github.com/ipfs/go-ipfs $ env IPFS_PATH=/nas/achin/ipfs_tmp ipfs bitswap wantlist |wc -l
3134
```


I have not yet tried these same steps with the latest master to compare results.

eminence:
I just got around to testing this scenario with the latest master.

Using current master branch (f44ff30 at the time of this writing):

`ipfs stats bw` reports a "RateOut" that seems to be between 350kB/s and 450 kB/s.  (As measured by running this command about 30 times and observing the result).  The "RateIn" was roughly between 1.3MB/s and 1.8MB/s.  No wantlist blowups as described in my previous comment.

Using your latest commit (1eca7bd85a):

"RateOut" seems to be between 110 kB/s and 190 kB/s (while "RateIn" was between 2.2 MB/s and 2.4 MB/s).   No wnatlist blowups either.  

So there seems to be clear improvements to amount of outgoing bandwidth.

Other observations:

After downloading about 600MB of a 61GB dataset, I killed the `ipfs get` command.  My wantlist length drops to zero, but I'm still receiving data.  `ipfs stats bw` reports a substantial RateIn, and a bunch of dup data received.  This didn't happen in the master branch.

```
achin@bigbox ~ $ env IPFS_PATH=/nas/achin/ipfs_tmp ipfs stats bw                                                                                                                           Bandwidth
TotalIn: 3.9 GB
TotalOut: 294 MB
RateIn: 1.7 MB/s
RateOut: 232 kB/s
achin@bigbox ~ $ env IPFS_PATH=/nas/achin/ipfs_tmp ipfs stats bitswap
bitswap status
        provides buffer: 0 / 256
        blocks received: 13084                                                                                                                                                                     blocks sent: 0
        data received: 2748817430
        data sent: 0
        dup blocks received: 8401
        dup data received: 2.1 GB
        wantlist [0 keys]
        partners [31]
```

whyrusleeping:
@eminence Hey, wanna try again? I pushed a few fixes. Thanks for trying it out :)

whyrusleeping:
Whoops, posted my comment without refreshing the page. Thats great results, thanks again for the feedback :) I'll keep tweaking things and see if i can't fix that residual dup blocks issue

whyrusleeping:
@eminence also, `ipfs stats bw` has a `--poll` option so you don't need to keep rerunning it

whyrusleeping:
@eminence further questions, how long does the extra traffic continue for after the ipfs get is completed?

eminence:
> how long does the extra traffic continue for after the ipfs get is completed?

I'm not sure.  I didn't let the ipfs-get complete, but rather in interrupted it with a ctrl-c (I'm not sure if this is an important distinction).   I let things run for about 10ish minutes before killing the daemon.

whyrusleeping:
Mind trying again? I resolved another issue that was hampering the sending
of cancels

eminence:
Looks much better.  After killing the `ipfs get`, the download rate fell to about 10 kB/S within 45 seconds.

whyrusleeping:
Alright, last open question here:

When a new peer connects, do we still send them our entire wantlist?

whyrusleeping:
my answer:
Keep a list of all wanted keys that were 'broadcasted' (i.e. sent to all peers) and send *that* wantlist to newly connected peers.

whyrusleeping:
Woooo! Tests are green :D

whyrusleeping:
So a bit of a status update:

Sessions are implemented, but currently only wired up through `ipfs pin add`. More integrations will be added with future PRs
Sessions currently *does not* do improved block request allocation, and as a result you should not expect to see a reduced duplicate block count *with this PR*. Further PRs will be able to fairly easily implement request scheduling logic that will improve that, I didnt want to put too much in this already quite large PR.
With sessions enabled, you should expect to see a much lower outgoing bandwidth costs (upload) as far fewer wantlist updates will be set out to random peers who don't care. There will also be fewer unnecessary DHT findprovs lookups, which will lead to a reduction in both inbound and outbound bandwidth. When a much larger portion of the network migrates to using sessions (including us implementing sessions support in other commands like 'cat' and 'get') the overall idle bandwidth usage of running an ipfs node should be greatly reduced by default, and almost unnoticeably small with `--routing=dhtclient`.

whyrusleeping:
@VictorBjelkholm @Kubuxu jenkins is behaving oddly... The test failure there is for the ipfs shutdown command. The test does not exist in the feat/bitswap-sessions branch, and the test being run is not the same as the code that is in master. It looks like some weird caching issue

exchange/bitswap/session.go
```diff
@@ -0,0 +1,233 @@
...
+	activePeersArr []peer.ID
+
+	bs         *Bitswap
+	incoming   chan blkRecv
+	newReqs    chan []*cid.Cid
+	cancelKeys chan []*cid.Cid
+
+	interest  *lru.Cache
+	liveWants map[string]time.Time
+	liveCnt   int
```
Stebalien: Why can't we just call `len(liveWants)`?

exchange/bitswap/session.go
```diff
@@ -0,0 +1,233 @@
...
+				}(live[0])
+			}
+			s.resetTick()
+		case p := <-newpeers:
+			s.addActivePeer(p)
+		case <-ctx.Done():
+			return
+		}
+	}
+}
```
Stebalien: I'm not sure every session should have it's own event loop. For one, it makes it hard to avoid unnecessary work when fetching subtrees common to multiple sessions. On the other hand, having a per-session event loop does make fairness simpler.

*WARNING: Future-proofing ahead.*

Additionally, Im not sure if sessions should be so... first-class(?). Personally, I see them as metadata that can help us find blocks but hardly the *only* metadata. With a separate event loop for every session, adding block-level predictions that are unrelated to sessions  (e.g., some of the ones I was talking about in the issue) could be a bit awkward.

Basically, in the future, I expect we'll build a prediction engine that:

* Accepts information about:
  - Who probably has a block (input from: gossip, provider searches, command-line arguments).
  - Who probably doesn't have a block (input from: gossip, recently failed requests).
  - Who gave us a block (and on which session we received it).
* Given a CID, provides a stream of peer suggestions in the following order:
  1. Peers it believes have the block in question (learned through the information given to the prediction engine above).
  2. Peers in the current session.
  3. (maybe) Peers in any related sessions (sessions where parent blocks were found)
  4. Everyone (triggers a provider search).

exchange/bitswap/session.go
whyrusleeping: hrm... good point. I think this bit is a relic from before i kept a map of each want

exchange/bitswap/session.go
```diff
@@ -0,0 +1,233 @@
...
+	tofetch        []*cid.Cid
+	activePeers    map[peer.ID]struct{}
+	activePeersArr []peer.ID
+
+	bs         *Bitswap
+	incoming   chan blkRecv
+	newReqs    chan []*cid.Cid
+	cancelKeys chan []*cid.Cid
+
+	interest  *lru.Cache
```
whyrusleeping: no, this is used to track blocks that are relevant to this session, so responses from other sessions can inform us of peers. Even after we've received a block, it remains in the interest cache, because receiving a block from peer X a different session may mean that peer X also has other blocks our session wants.

exchange/bitswap/session.go
```diff
@@ -0,0 +1,233 @@
...
+		case keys := <-s.newReqs:
+			for _, k := range keys {
+				s.interest.Add(k.KeyString(), nil)
+			}
+			if s.liveCnt < activeWantsLimit {
+				toadd := activeWantsLimit - s.liveCnt
+				if toadd > len(keys) {
+					toadd = len(keys)
+				}
+				s.liveCnt += toadd
```
whyrusleeping: removed liveCnt

exchange/bitswap/session.go
```diff
@@ -0,0 +1,233 @@
...
+}
+
+func (s *Session) wantBlocks(ctx context.Context, ks []*cid.Cid) {
+	for _, c := range ks {
+		s.liveWants[c.KeyString()] = time.Now()
+	}
+	s.bs.wm.WantBlocks(ctx, ks, s.activePeersArr, s.id)
+}
+
+func (s *Session) cancel(keys []*cid.Cid) {
```
whyrusleeping: nope, see other comment about the point of the interest cache

core/commands/bitswap.go
```diff
@@ -64,7 +64,11 @@ var unwantCmd = &cmds.Command{
 			ks = append(ks, c)
 		}
 
-		bs.CancelWants(ks)
+		// TODO: This should maybe find *all* sessions for this request and cancel them?
+		// (why): in reality, i think this command should be removed. Its
+		// messing with the internal state of bitswap. You should cancel wants
+		// by killing the command that caused the want.
+		bs.CancelWants(ks, 0)
```
whyrusleeping: Yeah, cancelling the context of any request command cancels the associated wants. So this command really should just get removed, it has no real purpose

exchange/bitswap/session.go
whyrusleeping: > For one, it makes it hard to avoid unnecessary work when fetching subtrees common to multiple sessions.

For now, i'm operating under the assumption that this won't be a super common scenario. And even in cases where it does happen, the 'interested' cache logic helps out with cross-session interplay pretty decently

exchange/bitswap/bitswap.go
```diff
@@ -302,9 +275,19 @@ func (bs *Bitswap) GetBlocks(ctx context.Context, keys []*cid.Cid) (<-chan block
 	}
 }
 
+func (bs *Bitswap) getNextSessionID() uint64 {
+	bs.sessIDLk.Lock()
+	defer bs.sessIDLk.Unlock()
+	bs.sessID++
+	return bs.sessID
+}
```
whyrusleeping: doing it this way because other bits of code need to be able to reference which sessions things belong to. Otherwise we might end up with circular dependencies since bitswap imports wantlists and wantlists reference sessions IDs.

exchange/bitswap/session.go
Stebalien: The "interested" cache will ensure that all interested parties receive the block but won't prevent us from spamming our wantlist more widely than necessary.

Stebalien: Oh, I see what the "interested" cache does now; it ensures that intersecting sessions learn about each others peers.

However, given predictor above, there's still a period where we really don't need to ask the *entire* session for a block because we're pretty sure we know which peers are likely to have it. That is, given the existing system, we could just do what you're doing now with `FindProvidersAsync` and add peers known to have blocks to the session but that still means that we'll need to spam the entire session even if we know which specific peers may have the block.

Basically, the current architecture shrinks our default "ask set" from everyone to the peers in the session however, it makes it difficult to shrink this further when we have more information about where to find a block.

exchange/bitswap/session.go
Stebalien: I see, nice! But there's still the problem of waiting forever if the block gets evicted from the interest cache *before* we receive it (I'll comment on that line to point it out to you).

exchange/bitswap/bitswap.go
```diff
@@ -375,6 +373,10 @@ func (bs *Bitswap) ReceiveMessage(ctx context.Context, p peer.ID, incoming bsmsg
 			k := b.Cid()
 			log.Event(ctx, "Bitswap.GetBlockRequest.End", k)
 
+			for _, ses := range bs.SessionsForBlock(k) {
+				ses.receiveBlockFrom(p, b)
```
Stebalien: Unless I'm mistaken, this is the *only* place that feeds blocks into the session. If a block gets evicted from the interested set *before* we find it, we'll never forward it to the session.

whyrusleeping: oooh, good catch. the 'InterestedIn' check should also check the `liveWants` map.

exchange/bitswap/session.go
whyrusleeping: Yeah, we should also add overlapping `FindProvider` peers to the sessions in a similar way, i like that idea, though doing that wouldnt necessarily allow us to remove this code (since its possible for us to get information from a peer that was not received by `FindProviders`).

Stebalien: I agree, I just think it may be worth waiting a few ms *before* spamming the entire session if we already *know* of some potential providers (i.e., we *don't* have to ask the DHT). 

I still think the current session-centric architecture may make this a bit difficult but it'll probably be easier to just fix it later than to try to hash out all the issues up-front. For all I know, I'm totally wrong and arguing about a non-existent issue.

blockservice/blockservice.go
```diff
@@ -31,6 +32,7 @@ type BlockService interface {
 	GetBlock(ctx context.Context, c *cid.Cid) (blocks.Block, error)
 	GetBlocks(ctx context.Context, ks []*cid.Cid) <-chan blocks.Block
 	DeleteBlock(o blocks.Block) error
+	NewSession(context.Context) *Session
```
Stebalien: This is a public interface and `Session` has private fields.

exchange/bitswap/bitswap.go
Stebalien: Out of curiosity, what happens if the user adds the block to the local blockstore via the commandline while we're trying to fetch it from the network?

blockservice/blockservice.go
whyrusleeping: I really didnt feel like making yet another only-implemented-once interface.

Stebalien: Couldn't this just be a helper function?

```go
func NewSession(ctx context.Context, bs BlockService) *Session {
	exchange := bs.Exchange()
	if bswap, ok := exchange.(*bitswap.Bitswap); ok {
		ses := bswap.NewSession(ctx)
		return &Session{
			ses: ses,
			bs:  bs.Blockstore(),
		}
	}
	return &Session{
		ses: bs.exchange,
		bs:  bs.Blockstore(),
	}
}
```

(we could also make this non-bitswap specific by adding a `NewSession(context.Context) exchange.Interface` method to `exchange.Interface` but we don't need that now).

exchange/bitswap/bitswap.go
whyrusleeping: All blocks added to the datastore get added through either the blockservice or the dagservice (which wraps the blockservice). The blockservice calls "exchange.HasBlock" for each block being added to it, then the blocks get pushed through the notifications channel

Stebalien: Isn't that a bit racy? That is, what happens if the user adds the block after the call to `blockstore.Get` but before the call to `exchange.GetBlock` ([here](https://github.com/ipfs/go-ipfs/blob/8ab93aaeb220a2ea31b0420479eba4b99cc592d4/blockservice/blockservice.go#L148))?

I'm asking because there's also race condition in this PR when a block is neither in the interest cache (we're interested in too many blocks) nor in the liveWants set (we haven't started fetching it yet) but gets fetched by another session (or manually added by the user). In this case, we'll try to fetch the block from the network anyways even if we already have it.

So, I'm wondering if it's worth adding a notification framework to Blockstores where one can wait for blocks to be added. An alternative is to bitswap with ourself (but that seems kind of hacky).

whyrusleeping: That race condition exists in the code today, technically, if you get the timing just right, a call to get could return not found from the blockstore, then the entire put could happen, then the bitswap getblock could be called. 

Adding a notifications framework to the blockstore feels like overkill, and the only other real solutions involve locking, or an extra call to `Has` (which to be fair, can be easily cached).

At one point we did call blockstore.Get on each block again inside bitswap, but i remember it being removed at some point in the name of datastore performance.

blockservice/blockservice.go
```diff
@@ -141,16 +157,25 @@ func (s *blockService) AddBlocks(bs []blocks.Block) ([]*cid.Cid, error) {
 func (s *blockService) GetBlock(ctx context.Context, c *cid.Cid) (blocks.Block, error) {
 	log.Debugf("BlockService GetBlock: '%s'", c)
 
-	block, err := s.blockstore.Get(c)
+	var f exchange.Fetcher
+	if s.exchange != nil {
+		f = s.exchange
+	}
```
Kubuxu: Is this needed, can't we just pass `s.exchange`?

exchange/bitswap/bitswap.go
```diff
@@ -340,7 +323,23 @@ func (bs *Bitswap) HasBlock(blk blocks.Block) error {
...
+	for _, s := range bs.sessions {
+		if s.interestedIn(c) {
+			out = append(out, s)
+		}
+	}
+	return out
+}
+
 func (bs *Bitswap) ReceiveMessage(ctx context.Context, p peer.ID, incoming bsmsg.BitSwapMessage) {
+	atomic.AddUint64(&bs.messagesRecvd, 1)
```
Kubuxu: Atomic will only work if address is 8 bytes, aliased. This might not be the case if there are other variables in the structure, I suggest allocating separate structure for all metrics and keeping all variable that will be used atomically as `(u)int64` and at the front.

exchange/bitswap/session.go
```diff
@@ -0,0 +1,261 @@
...
+	resp chan bool
+}
+
+// TODO: PERF: this is using a channel to guard a map access against race
+// conditions. This is definitely much slower than a mutex, though its unclear
+// if it will actually induce any noticeable slowness. This is implemented this
+// way to avoid adding a more complex set of mutexes around the liveWants map.
+// note that in the average case (where this session *is* interested in the
+// block we received) this function will not be called, as the cid will likely
+// still be in the interest cache.
```
Kubuxu: Go 1.9 has concurrent Maps which will resolve this issue altogether.

exchange/bitswap/session.go
```diff
@@ -0,0 +1,261 @@
...
+					}
+				}(live[0])
+			}
+			s.resetTick()
+		case p := <-newpeers:
+			s.addActivePeer(p)
+		case lwchk := <-s.interestReqs:
+			_, ok := s.liveWants[lwchk.c.KeyString()]
+			lwchk.resp <- ok
+		case <-ctx.Done():
```
Kubuxu: Done should close the ticker, otherwise it will keep the instance forever.

exchange/bitswap/session_test.go
```diff
@@ -0,0 +1,204 @@
...
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	blocksutil "github.com/ipfs/go-ipfs/blocks/blocksutil"
+	blocks "gx/ipfs/QmXxGS5QsUxpR3iqL5DjmsYPHR1Yz74siRQ4ChJqWFosMh/go-block-format"
+
+	cid "gx/ipfs/Qma4RJSuh7mMeJQYCqMbKzekn6EwBo7HEs5AQYjVRMQATB/go-cid"
```
Kubuxu: spacing

blockservice/blockservice.go
whyrusleeping: nope, because then the response wouldnt be `nil` (it would be a typed nil, which in interface form is != nil)

exchange/bitswap/bitswap.go
whyrusleeping: I don't think thats something we actually have to worry about. If that were the case, then i'm quite certain there would be fairly significant warnings around the calls in the docs: https://golang.org/pkg/sync/atomic/#AddUint64

exchange/bitswap/session.go
whyrusleeping: Yeah, that will be useful

exchange/bitswap/bitswap.go
Kubuxu: They do: https://golang.org/pkg/sync/atomic/#pkg-note-BUG

whyrusleeping:
@Kubuxu addressed those concerns

exchange/bitswap/bitswap.go
whyrusleeping: alright, how does that fix look?

exchange/bitswap/bitswap.go
Stebalien: > That race condition exists in the code today

The first one already exists but the second one is a bit different and has a much larger window. To reproduce:

1. Try fetching 3000 nodes that can't be found on the network.
2. Manually add 100th node.
3. Manually add the rest.

The operation will never complete because, when added, the 100th node was neither in the interest set nor in the `liveWants` set. The quick fix is to add an `allWants` set and use that instead of of checking `liveWants` and `interest`.

---

However, *all* these race conditions like this could be fixed by *some* centralized notification system (doesn't have to be built into the blockstore).

whyrusleeping: I'm not sure that a notification system would actually eliminate race conditions like this, it seems very tricky to do without taking a lock around all 'put block' calls that gets released once the notification has finished propogating. Will think about this more after some sleep.

Stebalien: We should be fine as long as we always notify when putting a block into the blockstore. Yes, we may end up *starting* a bitswap search for a block that you *just* got but we can cancel it immediately and don't have to wait for it to finish. My primary worry here is that the bitswap search may *never* finish if the block was retrieved on some side-channel and can't be found on the network.

The way I see this working is as follows:

`BlockService.GetBlock` will:

1. Try the blockstore first. Else...
2. Put a request into a "want" channel.
3. `Subscribe(cid)` on the internal notification system.
4. Call `blockstore.Get` again in case it was delivered between 1 and 3.
5. Wait on the channel returned by `Subscribe`.

`BlockService.AddBlock` will:

1. Put the block.
2. Publish on the notification system.

The main event loop for the BlockService will:

* Wait for requests on the `want` channel:
  1. Throw away duplicate want requests.
  2. Record that we're looking for the block.
  2. Asynchronously:
    1. Ask the exchange to find the block.
    2. Send the response back on an `exchangeResponse` channel.
* Wait on `exchangeResponse` channel.
  1. Record that we're done looking for the block.
  2. Call `BlockService.AddBlock(...)` (triggering the notification system).

Stebalien: (Note: this is very much "future work" in case that wasn't clear.)

whyrusleeping: @Stebalien I pushed a fix that at least doesnt lose track of blocks that may be added locally, and left some TODO notes for future work. I think that covers the major outstanding concerns, there will definitely be followup PRs, but this one should be good to go

Issue #3875, opened 2017-04-24
bug: Bitswap not recording 'Bytes sent'


dgrisham:
#### Version information:

go-ipfs version: 0.4.9-dev-9dc5883a
Repo version: 5
System version: amd64/linux
Golang version: go1.8.1

This bug was also observed in version 0.4.8-dev.

#### Description:

When sending blocks to another user, the number of bytes sent to that user is not recorded in the corresponding Bitswap ledger (the receiving peer *does* record the number of bytes they receive, however).

**Steps to reproduce**:

1. (Optional) Initialize 2 peers to be connected to only each other.
    - On both peers: `ipfs init`, `ipfs bootstrap rm all`, `ipfs bootstrap add <other_peer_multiaddr>`
    - Run `ipfs daemon`, then `ipfs swarm peers` to ensure connections are established.
2. Add file to IPFS on one of the peers, then get the file on the other peer.
    a. Peer 1: `head -c 1000 /dev/urandom | base64 > /tmp/file && ipfs add -q /tmp/file`
    b. Peer 2: `ipfs get <file_hash>`
3. Check Bitswap ledgers with `ipfs bitswap ledger <peer_id>`

After I run those steps, the result on the *sending* peer is:

```
Ledger for <peer.ID XLhkey>
Debt ratio:     0.000000
Exchanges:      0
Bytes sent:     0
Bytes received: 0
```

And on the *receiving* peer:

```
Ledger for <peer.ID PipTao>
Debt ratio:     0.000000
Exchanges:      1
Bytes sent:     0
Bytes received: 1365
```

If I then swap the roles and have the second peer upload a file that the first peer downloads, the first ledger changes to:

```
Ledger for <peer.ID XLhkey>
Debt ratio:     0.000000
Exchanges:      1
Bytes sent:     0
Bytes received: 1365
```

and the second ledger is unchanged from the one above.

I'm going to look into this myself, but I thought I'd post it here in case anyone has an immediate idea as to why this is happening. 

Thanks!

dgrisham:
It looks like the function `MessageSent`, which updates the ledger just like `MessageReceived` when a message is received, wasn't actually called anywhere. I was able to fix it by adding a call to `MessageSent` where it seemed to make the most sense and have verified that the update works as expected, I'm about to submit a pull request with the change. Let me know if there are any other checks or changes I can make to this update, or if I completely missed the reason for this issue and my change was unnecessary.

whyrusleeping:
[...]

PR #3876, opened 2017-04-25
bug fix: Bitswap now records 'BytesSent' in ledger

dgrisham:
This is a fix for the bug detailed in https://github.com/ipfs/go-ipfs/issues/3875. I add the call to `MessageSent()` where it seemed to make the most sense, then manual testing (akin to the **Steps to reproduce** in the linked issue) was done and automated tests were added `bitswap_test.go` to verify that the update worked. Upon running the automated tests, I received a read/write conflict for the ledger that I fixed by adding `Lock()`/`Unlock()` calls for the ledger at the beginning of `MessageSent()`.

### DIFF
--- a/exchange/bitswap/bitswap_test.go
+++ b/exchange/bitswap/bitswap_test.go
@@ -507,7 +507,37 @@ func assertLedgerMatch(ra, rb *decision.Receipt) error {
 	return nil
 }
 
-func TestBitswapBytesSentOneWay(t *testing.T) {
+func assertLedgerEqual(ra, rb *decision.Receipt) error {
+	if ra.Value != rb.Value {
+		return fmt.Errorf("mismatch in ledgers (value/debt ratio): %f vs %f ", ra.Value, rb.Value)
+	}
+
+	if ra.Sent != rb.Sent {
+		return fmt.Errorf("mismatch in ledgers (sent bytes): %d vs %d", ra.Sent, rb.Sent)
+	}
+
+	if ra.Recv != rb.Recv {
+		return fmt.Errorf("mismatch in ledgers (recvd bytes): %d vs %d", ra.Recv, rb.Recv)
+	}
+
+	if ra.Exchanged != rb.Exchanged {
+		return fmt.Errorf("mismatch in ledgers (exchanged blocks): %d vs %d ", ra.Exchanged, rb.Exchanged)
+	}
+
+	return nil
+}
+
+func newReceipt(sent, recv, exchanged uint64) *decision.Receipt {
+	return &decision.Receipt{
+		Peer:      "test",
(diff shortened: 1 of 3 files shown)


dgrisham:
The `TODO` I added is there as a reminder and to reflect the same message in the Bitswap `ReceiveMessage()` function, not something that is meant to be fixed in this PR.

Kubuxu:
GitCop messages are about: https://github.com/ipfs/community/blob/master/contribution-guidelines.md#a-license-and-a-signed-off-by-trailers-are-required

dgrisham:
@Kubuxu wow, I definitely should have gone through the contribution guidelines before this PR. Thanks for the link!

exchange/bitswap/bitswap_test.go
```diff
@@ -489,3 +490,109 @@ func TestWantlistCleanup(t *testing.T) {
...
+	}
+
+	ctx, cancel = context.WithTimeout(context.Background(), time.Second*5)
+	defer cancel()
+	blk, err = instances[0].Exchange.GetBlock(ctx, blocks[1].Cid())
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	ra := instances[0].Exchange.LedgerForPeer(instances[1].Peer)
```
whyrusleeping: I think it might also be useful to assert some actual values here

dgrisham: @whyrusleeping makes sense, I saw that in one of the other tests but forget to include it in mine. I add that in the next push, thanks!

dgrisham:
Is it common for CircleCI tests (for IPFS) to fail on one run then succeed on a subsequent re-run? It looks like it failed once, then I pushed again with only commit message changes, then it passed. I just pushed again and it failed, so maybe a rerun is suitable here?

whyrusleeping:
@dgrisham Yeah.... that particular test failing is a known issue. It only fails on travis CI or circle CI for some reason

whyrusleeping:
@dgrisham mind rebasing this on master to avoid having an extra merge commit?

dgrisham:
@whyrusleeping You got it, just pushed without that merge commit.

Kubuxu:
Nice amount of tests :+1:. Good work.

dgrisham:
@Kubuxu thanks! The tests were straightforward to write given the existing framework and other tests to go off of. The codebase in general is very well written as well, I had no experience with Go before looking at the Bitswap portion of the IPFS codebase but it was easy to pick up the common coding patterns and find what I needed to fix this bug.

Kubuxu:
@dgrisham this is how I started hacking on go-ipfs (I had a bit of experience in Go, mostly C).

PR #3880, opened 2017-04-26
Fix install_unsupported for openbsd, add docs

kpcyrd:
License: MIT

### DIFF
--- a/Rules.mk
+++ b/Rules.mk
@@ -105,12 +105,14 @@ install: cmd/ipfs-install
 install_unsupported:
 	@echo "note: this command has yet to be tested to build in the system you are using"
 	@echo "installing gx"
-	go get -u github.com/whyrusleeping/gx
-	go get -u github.com/whyrusleeping/gx-go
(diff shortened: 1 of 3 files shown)


Rules.mk
```diff
@@ -105,12 +105,14 @@ install: cmd/ipfs-install
 install_unsupported:
 	@echo "note: this command has yet to be tested to build in the system you are using"
 	@echo "installing gx"
-	go get -u github.com/whyrusleeping/gx
-	go get -u github.com/whyrusleeping/gx-go
+	go get -v -u github.com/whyrusleeping/gx
+	go get -v -u github.com/whyrusleeping/gx-go
+	@echo check gx and gx-go
+	gx -v && gx-go -v
```
whyrusleeping: I wouldnt remove this. Its rather important to ensure both are installed properly before continuing

Kubuxu: He adds it.

whyrusleeping: I am great at reading diffs.

Issue #3807, opened 2017-03-21
PeerID in IPNS key workflow


lgierth:
#### Description:

- `ipfs name publish --key` should accept a PeerID in addition to the key name.
- `ipfs key list` should list the PeerID alongside the key name (`-l` arg should be default true).

keremgocen:
Hey @lgierth, could you please give a bit more detail for the expected behavior of these 2 items?

- It seems straightforward to change the default value of `ipfs key list` like [this](https://github.com/keremgocen/go-ipfs/commit/aa12416db52cb90b3a9c3e81195d8705650bf9fa#diff-97187cb5f6f28de4954c862a0f2d1308L158) but do you want to preserve the previous behavior by providing another option?

- `ipfs name publish --key` for this one, as I understand currently it's publishing to/as `self` using the node's own `PeerID`, this would enable the node to publish to other nodes instead of self bypassing the [private key lookup here](https://github.com/ipfs/go-ipfs/blob/9dc5883ac8a633e2aca50d0c479677a2d9631202/core/commands/publish.go#L169)? This part is a bit ambiguous to me, i'd appreciate any details

Kubuxu:
@lgierth @keremgocen I think the current behaviour of `ipfs key list` should stay. It is already established API so we shouldn't change it. We might want to create list of API changes we want to do when we are able to provider two versions of API or we decide to make a non-compatible change to the API.

@keremgocen about the `ipfs name publish --key` currently you are able to use it with key name. Like:
```bash
ipfs key gen --type=rsa --size=2048 mykey
ipfs name publish --key=mykey QmSomeHash
```

@lgierth calls for being able to use `peerID` instead of keyname (`mykey` in this case)`. PeerID can be obtained using `ipfs key list -l` but I am not sure if keystore (code responsible for storing keys) provides good method to lookup PeerID -> keyname or PeerID -> private key.

keremgocen:
Then I shall skip the "-l arg should be default true" part.

@Kubuxu what would be a good way to lookup peerIDs? I see some potential ways of doing so (like this one)[https://github.com/ipfs/go-ipfs/blob/master/core/commands/id.go#L100] but not sure which one would be the best option with my current exposure level to the codebase.

Kubuxu:
[...]

PR #3882, opened 2017-04-27
WIP: add PeerID lookup

keremgocen:
- [x] enhance key lookup logic and enable PeerID lookup in addition to the key name
- [x] write sharness tests

as discussed in #3807

License: MIT
Signed-off-by: Kerem Gocen <keremgocen@gmail.com>

### DIFF
--- a/core/commands/publish.go
+++ b/core/commands/publish.go
@@ -48,6 +48,11 @@ Publish an <ipfs-path> with another name, added by an 'ipfs key' command:
   > ipfs name publish --key=mykey /ipfs/QmatmE9msSfkKxoffpHwNLNKgwZG8eT9Bud6YoPab52vpy
   Published to QmbCMUZw6JFeZ7Wp9jkzbye3Fzp2GGcPgC3nmeUjfVF87n: /ipfs/QmatmE9msSfkKxoffpHwNLNKgwZG8eT9Bud6YoPab52vpy
 
+Alternatively, publish an <ipfs-path> using a valid PeerID(as listed by 'ipfs key list -l'):
+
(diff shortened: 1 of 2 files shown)


core/commands/publish.go
```diff
@@ -176,3 +181,36 @@ func publish(ctx context.Context, n *core.IpfsNode, k crypto.PrivKey, ref path.P
 		Value: ref.String(),
 	}, nil
 }
+
+func keylookup(k string, n *core.IpfsNode) (crypto.PrivKey, error) {
+
+	res, err := n.GetKey(k)
+	if res != nil {
```
Kubuxu: I would check here if error is different than `ErrNoSuchKey` if that is a case It signals different error than should happen here.

core/commands/publish.go
```diff
@@ -176,3 +181,36 @@ func publish(ctx context.Context, n *core.IpfsNode, k crypto.PrivKey, ref path.P
 		Value: ref.String(),
 	}, nil
 }
+
+func keylookup(k string, n *core.IpfsNode) (crypto.PrivKey, error) {
```
Kubuxu: Please make the Node first argument.

keremgocen:
Force pushed on my fork to change commit message for f93e8c58d8b92e1e8d998fb9525e99d187a6dced, should I close this for another PR @Kubuxu ?

Kubuxu:
No, force pushes are OK.

keremgocen:
Do you have any recommendations for how to fix the failing CI [here](https://travis-ci.org/ipfs/go-ipfs/builds/228194943?utm_source=github_status&utm_medium=notification)? Also a review is required to proceed this one forward I see.

Kubuxu:
@keremgocen don't worry about travis too much (it likes to fail on its own, we load their machines a bit too much).

whyrusleeping:
@Kubuxu :+1:  here?

PR #3884, opened 2017-04-28
Implement ipfs shutdown command

whyrusleeping:
License: MIT

### DIFF
--- a/core/commands/shutdown.go
+++ b/core/commands/shutdown.go
@@ -1,0 +1,29 @@
+package commands
+
+import (
+	"fmt"
+
+	cmds "github.com/ipfs/go-ipfs/commands"
+)
+
+var daemonShutdownCmd = &cmds.Command{
+	Helptext: cmds.HelpText{
+		Tagline: "Shut down the ipfs daemon",
+	},
+	Run: func(req cmds.Request, res cmds.Response) {
+		nd, err := req.InvocContext().GetNode()
+		if err != nil {
+			res.SetError(err, cmds.ErrNormal)
+			return
+		}
+
+		if nd.LocalMode() {
+			res.SetError(fmt.Errorf("daemon not running"), cmds.ErrClient)
+			return
+		}
+
+		if err := nd.Process().Close(); err != nil {
+			log.Error("error while shutting down ipfs daemon:", err)
+		}
+	},
+}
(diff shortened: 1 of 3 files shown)


kevina:
@whyrusleeping is there a reason this is a top-level command:
```
  ipfs daemon start
  ipfs daemon stop
```
makes a lot more sense to me,  for backwards comparability `ipfs daemon` can alias to `ipfs daemon start`

core/commands/shutdown.go
```diff
@@ -0,0 +1,29 @@
...
+		Tagline: "Shut down the ipfs daemon",
+	},
+	Run: func(req cmds.Request, res cmds.Response) {
+		nd, err := req.InvocContext().GetNode()
+		if err != nil {
+			res.SetError(err, cmds.ErrNormal)
+			return
+		}
+
+		if !nd.OnlineMode() {
```
kevina: I am not 100% sure but I don't think this works when the daemon is started with the `--offline` flag.

kevina: I am not 100% sure but I don't think this works when the daemon is started with the --offline flag.

test/sharness/t0023-shutdown.sh
```diff
@@ -0,0 +1,23 @@
...
+test_launch_ipfs_daemon
+
+test_expect_success "shutdown succeeds" '
+	ipfs shutdown
+'
+
+test_expect_success "daemon no longer running" '
+	test_expect_code  1 kill -0 $IPFS_PID
+'
+
```
kevina: I would also test the daemon when started with the `--offline` flag.

When I added the test it failed.

whyrusleeping:
Its a top level command because its rather difficult to make a command be a subcommand of the daemon command. I tried that at first and it would have required me to start refactoring things.

whyrusleeping:
@kevina made it work with --offline. Needed to use the `LocalMode` check instead of the `OnlineMode` one

kevina:
BTW @whyrusleeping codeclimate is failing:

> exported var DaemonShutdownCmd should have comment or be unexported

Kubuxu:
Is there any reason to export this command definition.,w:x

whyrusleeping:
Yeah, probably no reason to export it

PR #3888, opened 2017-04-29
Flush cached directories every so often during an add

whyrusleeping:
If youre having trouble adding large amounts of directories to ipfs and run out of memory, try this

License: MIT

### DIFF
--- a/core/coreunix/add.go
+++ b/core/coreunix/add.go
@@ -36,6 +36,8 @@ var log = logging.Logger("coreunix")
 // how many bytes of progress to wait before sending a progress update message
 const progressReaderIncrement = 1024 * 256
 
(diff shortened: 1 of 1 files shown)


core/coreunix/add.go
```diff
@@ -104,6 +106,7 @@ type Adder struct {
 	unlocker   bs.Unlocker
 	tempRoot   *cid.Cid
 	Prefix     *cid.Prefix
+	liveNodes  uint64
```
Kubuxu: Is this ever being increased?

whyrusleeping: damn, i knew i messed something up

whyrusleeping: fixed

PR #3892, opened 2017-05-01
Implement ipfs key {rm, rename}

MichaelMure:
@whyrusleeping So you don't waste time before implementing the good encryption stuff ;)

License: MIT

### DIFF
--- a/core/commands/keystore.go
+++ b/core/commands/keystore.go
@@ -49,6 +49,7 @@ type KeyOutputList struct {
 	Keys []KeyOutput
 }
 
+// KeyRenameOutput define the output type of keyRenameCmd
 type KeyRenameOutput struct {
 	Was       string
 	Now       string
--- a/core/commands/keystore.go
+++ b/core/commands/keystore.go
@@ -33,9 +33,9 @@ var KeyCmd = &cmds.Command{
 		`,
 	},
 	Subcommands: map[string]*cmds.Command{
-		"gen":  KeyGenCmd,
-		"list": KeyListCmd,
-		"rm": KeyRmCmd,
+		"gen":  keyGenCmd,
+		"list": keyListCmd,
+		"rm":   keyRmCmd,
 	},
 }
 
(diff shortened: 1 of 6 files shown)


core/commands/keystore.go
```diff
@@ -202,6 +203,61 @@ var KeyListCmd = &cmds.Command{
 	Type: KeyOutputList{},
 }
 
+var KeyRmCmd = &cmds.Command{
+	Helptext: cmds.HelpText{
+		Tagline: "Remove a keypair",
+	},
+	Arguments: []cmds.Argument{
+		cmds.StringArg("name", true, false, "name of key to remove"),
```
Kubuxu: This command should allow for StdIn handling and multiple argument handling. Then it would be possible to remove multiple keys at the same time easily.

Kubuxu:
See also codeclimate errors. No need to export this command definition. We are switching to not exporting them where not needed.

core/commands/keystore.go
```diff
@@ -202,6 +203,65 @@ var KeyListCmd = &cmds.Command{
 	Type: KeyOutputList{},
 }
 
+var keyRmCmd = &cmds.Command{
+	Helptext: cmds.HelpText{
+		Tagline: "Remove a keypair",
+	},
+	Arguments: []cmds.Argument{
+		cmds.StringArg("name", true, true, "names of keys to remove"),
```
Kubuxu: Call on the return value of `StringArg` `.EnableStdin()` to enable stdin handling for it.

MichaelMure:
@Kubuxu @whyrusleeping I added an implementation of ipfs key rename, because why not.

keystore/keystore.go
```diff
@@ -54,6 +55,14 @@ func NewFSKeystore(dir string) (*FSKeystore, error) {
 	return &FSKeystore{dir}, nil
 }
 
+func (ks *FSKeystore) Has(name string) bool {
+	kp := filepath.Join(ks.dir, name)
+
+	_, err := os.Stat(kp)
```
whyrusleeping: Has should probably also return an error. If we run into some permissions issue, this will report that the key doesnt exist. use `os.IsErrNotExist` to check for file existence, otherwise the error is a real error

whyrusleeping:
I would add a test that makes sure you cant rename self to something else. Then codeclimate has three issues to fix.

After that, good to merge

MichaelMure:
@whyrusleeping It should be good now, can you merge ?

whyrusleeping:
Thanks @MichaelMure :)

PR #3897, opened 2017-05-05
improved gateway directory listing for sharded nodes

whyrusleeping:
License: MIT

### DIFF
--- a/core/corehttp/gateway_handler.go
+++ b/core/corehttp/gateway_handler.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"os"
 	gopath "path"
 	"runtime/debug"
 	"strings"
@@ -20,6 +21,7 @@ import (
 	dagutils "github.com/ipfs/go-ipfs/merkledag/utils"
 	path "github.com/ipfs/go-ipfs/path"
 	ft "github.com/ipfs/go-ipfs/unixfs"
+	uio "github.com/ipfs/go-ipfs/unixfs/io"
 
 	humanize "gx/ipfs/QmPSBJL4momYnE7DcUyk2DVhD6rH488ZmHBGLbxNdhU44K/go-humanize"
 	cid "gx/ipfs/QmYhQaCYEcaPPjxJX7YcPcVKkQfRy6sJ7B3XmGFk82XYdQ/go-cid"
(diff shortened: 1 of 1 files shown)


core/corehttp/gateway_handler.go
```diff
@@ -227,92 +229,102 @@ func (i *gatewayHandler) getOrHeadHandler(ctx context.Context, w http.ResponseWr
 		return
 	}
 
-	links, err := i.api.Unixfs().Ls(ctx, resolvedPath)
+	nd, err := i.api.ResolveNode(ctx, resolvedPath)
```
lgierth: Now that we get the whole node here, we can meld this call and `i.api.Unixfs().Cat(ctx, resolvedPath)` further above into one

whyrusleeping:
I would love to get this reviewed ASAP so we can sneak it into 0.4.9. It makes viewing indexes in large directories on the gateway possible.

core/commands/files/files.go
```diff
@@ -162,38 +162,43 @@ func statNode(ds dag.DAGService, fsn mfs.FSNode) (*Object, error) {
...
+		case mfs.TFile:
+			out.Type = "file"
+		default:
+			return nil, fmt.Errorf("Unrecognized node type: %s", fsn.Type())
+		}
+		out.Blocks = len(nd.Links())
+		out.Size = d.GetFilesize()
+		out.CumulativeSize = cumulsize
+
+	case *dag.RawNode:
```
kevina: Question: There is a lot of code movement in this file.  From what I can tell all that is really going on is there is an optimization for RawNode so that it doesn't have the get the block associated with the RawNode.  Is this correct?

whyrusleeping: Basically, yeah. Though im going to remove the changes to the files code from this PR. Its a separate changeset

whyrusleeping:
Previously, to find if the directory had an index.html, we enumerated all the links, and iterated over them and looked for one named 'index.html'. That worked okay previously, but for large sharded directories, its realy slow. I've switched it to using `Find` on a unixfs directory, which does an efficient lookup.

core/corehttp/gateway_handler.go
```diff
@@ -227,92 +229,102 @@ func (i *gatewayHandler) getOrHeadHandler(ctx context.Context, w http.ResponseWr
...
+			internalWebError(w, err)
+			return
 		}
+		defer dr.Close()
 
+		// write to request
+		http.ServeContent(w, r, "index.html", modtime, dr)
+		return
+	default:
+		internalWebError(w, err)
```
kevina: Not sure what the official go style is here, but I would put the default as the last case, although I have seen it as the first case in go code.

Kubuxu: In this case it isn't about keyword but its meaning. In this case `default` error is still an error, below it there is non-error handler.

PR #3901, opened 2017-05-05
WIP Fix DagModifier for raw-leaves nodes

Kubuxu:


whyrusleeping:
We shouldnt need to make different types of dagreaders, The single dagreader should be able to handle all 'unixfs' nodes (including rawnodes)

whyrusleeping:
@Kubuxu I had some WIP work on this from last night, pushed it here: https://github.com/ipfs/go-ipfs/tree/feat/dagmod-raw

Kubuxu:
I thought it would be just easier as there is a clear upgrade spot when the simple handling of raw-leaves becomes just being like any other node that DagModifier handles.

I don't think there is much reason to complicate the DagModifier code with 1. logic of handling just one solo raw-node 2. logic of upgrade from solo raw-node to multiple raw-nodes linked.

kevina:
@Kubuxu @whyrusleeping do you want me to take this over?

whyrusleeping:
@kevina sure, but i do prefer the approach in the branch i linked above. I think its much simpler than creating a new type of dagmodifier for the case where we have to deal with a different type.

kevina:
@whyrusleeping Okay I will cherry pick your changes in.  I was starting to do something similar.

whyrusleeping:
@kevina sounds good! thanks!

kevina:
@Kubuxu can we close this is favor of https://github.com/ipfs/go-ipfs/pull/4026?

whyrusleeping:
Yeah, i think #4026 closes this. If i'm wrong, please go ahead and reopen it

PR #3906, opened 2017-05-07
Add & update packages

Voker57:
Updated package requirements to versions from master

### DIFF
--- a/core/commands/object/diff.go
+++ b/core/commands/object/diff.go
@@ -7,7 +7,6 @@ import (
 
 	cmds "github.com/ipfs/go-ipfs/commands"
 	core "github.com/ipfs/go-ipfs/core"
-	dag "github.com/ipfs/go-ipfs/merkledag"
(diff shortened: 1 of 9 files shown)


whyrusleeping:
These hashes are different than the ones in: https://github.com/ipfs/go-cid/blob/master/.gx/lastpubver and https://github.com/ipfs/go-ipld-format/blob/master/.gx/lastpubver Could you use those ones instead?

Voker57:
Sure, if that's not outside the PR scope. Is there an easier way to do updates than manually checking lastpubver and running gx update, then checking for all other importers being in sync?

Voker57:
That's problematic currently because latest version of go-libp2p-kad-dht depends on go-libp2p-routing, which depends on go-cid QmYhQaCYEcaPPjxJX7YcPcVKkQfRy6sJ7B3XmGFk82XYdQ, which is not the latest version of go-cid.

PR #3909, opened 2017-05-09
Adding documentation that PubSub Sub can be encoded.

TUSF:
Turns out, the Message will Marshal properly. I tested XML and Protobufs, but doesn't seem the "marshaller" is there. I'm still not familiar enough with the codebase to go poking around, so I'm just adding a bit of documentation that was missed, and is in my opinion pretty useful.


Kubuxu:
We should probably add info about this flag to `ipfs` help text itself. As it is global option for all commands.

PR #3914, opened 2017-05-10
Add Suborigin header to gateway responses

jes:
This existed before but was disabled in 912a972 because the Suborigin spec
changed and it became incompatible.

This commit updates the generated Suborigin header to be conformant with the
latest spec.

Thank you Stskeeps, Kubuxu, and TUSF on IRC for holding my hand with this.

### DIFF
--- a/core/corehttp/gateway_handler.go
+++ b/core/corehttp/gateway_handler.go
@@ -27,6 +27,7 @@ import (
 	cid "gx/ipfs/QmYhQaCYEcaPPjxJX7YcPcVKkQfRy6sJ7B3XmGFk82XYdQ/go-cid"
 	routing "gx/ipfs/QmafuecpeZp3k3sHJ5mUARHd4795revuadECQMkmHB8LfW/go-libp2p-routing"
 	node "gx/ipfs/Qmb3Hm9QDFmfYuET4pu7Kyg8JV78jFa1nvZx5vnCZsK4ck/go-ipld-format"
+	multibase "gx/ipfs/QmcxkxTVuURV2Ptse8TvkqH5BQDwV62X1x19JqqvbBzwUM/go-multibase"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


core/corehttp/gateway_handler.go
```diff
@@ -210,6 +211,33 @@ func (i *gatewayHandler) getOrHeadHandler(ctx context.Context, w http.ResponseWr
...
+	//
+	// Omitted if the path was treated by IPNSHostnameOption(), for example
+	// a request for http://example.net/ would be changed to /ipns/example.net/,
+	// which would turn into an incorrect Suborigin: example.net header.
+	// In this case the correct thing to do is omit the header because it is already
+	// handled correctly without a Suborigin.
+	//
+	// NOTE: This is not yet widely supported by browsers.
+	if !ipnsHostname {
+		// e.g.: 0="ipfs.io", 1="ipfs", 2="QmYuNaKwY...", ...
```
lgierth: Part 0 is simply empty, `urlPath` is something like `/ipfs/QmFoo/a/b`.

core/corehttp/gateway_handler.go
```diff
@@ -210,6 +211,33 @@ func (i *gatewayHandler) getOrHeadHandler(ctx context.Context, w http.ResponseWr
...
+		pathComponents := strings.SplitN(urlPath, "/", 4)
+		suboriginRaw := []byte(strings.ToLower(pathComponents[2]))
```
Kubuxu: This isn't right. You can't 'ToLower' encoded value as it will change the IPFS hash.

core/corehttp/gateway_handler.go
jes: Noted, I'll fix.

core/corehttp/gateway_handler.go
jes: I guess the code isn't clear enough -- this part handles the /ipns/EXAMPLE.COM/ case; when the second part will decode as base58, it is handled accordingly. I'll add a comment to explain this.

Kubuxu: Ahh, I missed that. Yeah, it is better not to be clever in places like that.

whyrusleeping:
@lgierth @Kubuxu hows this one looking?

core/corehttp/gateway_handler.go
```diff
@@ -210,6 +211,39 @@ func (i *gatewayHandler) getOrHeadHandler(ctx context.Context, w http.ResponseWr
...
+		if err != nil {
+			// component 2 doesn't decode with cid, so it must be a hostname
+			suboriginRaw = []byte(strings.ToLower(pathComponents[2]))
+		} else {
+			suboriginRaw = cidDecoded.Bytes()
+		}
+
+		base32Encoded, err := multibase.Encode(multibase.Base32, suboriginRaw)
+		if err != nil {
+			internalWebError(w, err)
```
lgierth: This should be a 4xx error via `webError()`, probably just `400 Bad Request`. It's an error that can only be recovered from by altering the request.

jes: I don't see how base32 encoding can fail -- if it does, surely that's an internal server error, and nothing to do with the client?

lgierth: Mh -- yeah in this case the error can only come from an unsupported encoding parameter: https://github.com/multiformats/go-multibase/blob/master/multibase.go

LGTM then :+1:

jes: :) Thanks mate.

PR #3915, opened 2017-05-11
Add MaxStorage field to output of "repo stat".

### LINKED ISSUE #3836: Let "ipfs repo stat" include StorageMax as set in the configuration (in bytes)
#### Description:

I'm not sure this is a good idea but maybe I'm wrong or there's a better approach.

The underlying problem is to know how many more bytes an ipfs node can store.

`ipfs repo stat` provides RepoSize and `ipfs config show` provides `StorageMax`, so it's just matter of doing the math.If StorageMax came included in `repo stat`, then it would:

  * Save 1 api call
  * Save parsing of the configuration
  * Save converting human-readable configuration values to bytes (extra dependency)

But as I say, I'm not sure the `repo stat` information is the place for a GC-related configuration option.

kevina:
Closes #3836.

### DIFF
--- a/core/commands/repo.go
+++ b/core/commands/repo.go
@@ -189,6 +189,12 @@ Version         string The repo version.
 			} else {
 				fmt.Fprintf(buf, "RepoSize \t %d\n", stat.RepoSize)
 			}
+			maxSizeInMiB := stat.StorageMax / (1024 * 1024)
+			if human && maxSizeInMiB > 0 {
+				fmt.Fprintf(buf, "StorageMax (MiB) \t %d\n", maxSizeInMiB)
+			} else {
+				fmt.Fprintf(buf, "StorageMax \t %d\n", stat.StorageMax)
+			}
 			fmt.Fprintf(buf, "RepoPath \t %s\n", stat.RepoPath)
 			fmt.Fprintf(buf, "Version \t %s\n", stat.Version)
 
--- a/core/commands/repo.go
+++ b/core/commands/repo.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"text/tabwriter"
 
 	bstore "github.com/ipfs/go-ipfs/blocks/blockstore"
 	cmds "github.com/ipfs/go-ipfs/commands"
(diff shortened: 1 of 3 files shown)


kevina:
Note, with the current way formatting is done the output comes out as:
```
$ ipfs repo stat --human
NumObjects       1586
RepoSize (MiB)   67
StorageMax (MiB)         9536
RepoPath         /tmp/gocode2-ipfs
Version          fs-repo@5
```
Due to this code:
```
			buf := new(bytes.Buffer)
			fmt.Fprintf(buf, "NumObjects \t %d\n", stat.NumObjects)
			sizeInMiB := stat.RepoSize / (1024 * 1024)
			if human && sizeInMiB > 0 {
				fmt.Fprintf(buf, "RepoSize (MiB) \t %d\n", sizeInMiB)
			} else {
				fmt.Fprintf(buf, "RepoSize \t %d\n", stat.RepoSize)
			}
			maxSizeInMiB := stat.StorageMax / (1024 * 1024)
			if human && maxSizeInMiB > 0 {
				fmt.Fprintf(buf, "StorageMax (MiB) \t %d\n", maxSizeInMiB)
			} else {
				fmt.Fprintf(buf, "StorageMax \t %d\n", stat.StorageMax)
			}
			fmt.Fprintf(buf, "RepoPath \t %s\n", stat.RepoPath)
			fmt.Fprintf(buf, "Version \t %s\n", stat.Version)
```
should we perhaps not use tab in the output?  Although removing the tab will make the output more difficult to parse.  My vote would be to use:
```
NumObjects:       1586
RepoSize (MiB):   67
StorageMax (MiB): 9536
RepoPath:         /tmp/gocode2-ipfs
Version:          fs-repo@5
```
With a colon after the field names.  But maybe that is for another P.R.

@whyrusleeping thoughts?

hsanjuan:
I prefer nicely formatted outputs. Parsing should be done on json output anyway...

Also, code looks good to me.

whyrusleeping:
@kevina you could use a tabwriter (from the stdlib) That is fine with me

kevina:
@whyrusleeping okay I will do that, and also add a colon after the fields names as that makes reading it easier.  Sound good?

whyrusleeping:
Sounds good to me!

PR #3916, opened 2017-05-11
remove 'ipfs diag net' from codebase

whyrusleeping:
Been noticing quite a large amount of traffic from this lately. It seems people are using it more than intended and causing excess load on the system. That means its time to pull it out. Goodbye old friend

License: MIT


lgierth:
I'm okay removing it too -- tests need fixing

PR #3919, opened 2017-05-12
adder: add support for specifying the hash function

### LINKED ISSUE #3917: Add flag to ipfs add to select hash function to use
I see this as step 1 towards enabling blake2b support. This flag could even default to blake2b-256

kevina:
Closes #3917 

License: MIT

### DIFF
--- a/core/commands/add.go
+++ b/core/commands/add.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"strings"
 
 	bstore "github.com/ipfs/go-ipfs/blocks/blockstore"
 	blockservice "github.com/ipfs/go-ipfs/blockservice"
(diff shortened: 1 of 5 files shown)


kevina:
@whyrusleeping this p.r. will unconfitially set the Cid version to 1 if the `--hash` option is used, even if the value is `sha2-256`.

whyrusleeping:
Great! I would just like to see a test that adds some directory stuff recursively too

kevina:
I'm working on it.  Raw leaves are a bit of a problem.  They need proper support for specifying a hash function.

kevina:
@whyrusleeping okay this should be good to go, the first commit  "Rename NewBlock to NewBlockV0" is not necessary, I did it as a debugging aid and it might be useful to keep.  If not it can be cleanly dropped.

whyrusleeping:
Then theres a code climate issue, documenting the NewNodeWithPrefix would be very helpful

kevina:
@whyrusleeping both issues should be addressed now

PR #3937, opened 2017-05-22
Remove duplicate release checklist

mishmosh:
Removing this one, and keeping https://github.com/ipfs/go-ipfs/blob/master/docs/releases.md (per @larsgeirth in https://github.com/ipfs/blog/issues/97)

Please review/modify thoroughly. I am not familiar enough with go-ipfs release process to comment on contents.


Kubuxu:
This was to be template for release oriented automation (example, creation of issue with checklist) but it should be good enough being in the git history.

PR #3938, opened 2017-05-23
filestore: add "--file-order" option to "filestore ls" and "verify"

### LINKED ISSUE #3922: Make `ipfs filestore verify` process files sequentially
#### Version information:
go-ipfs version: 0.4.8-
Repo version: 5
System version: amd64/darwin
Golang version: go1.8

#### Description:
Currently, `ipfs filestore verify` verifies blocks ordered by hash. This causes an immense amount of unnecessary seeks which, especially on devices with rotating disks, makes verification much slower than it could be and unncessarily stresses system resources as well as causing wear.

If, instead, blocks would be orderd by file (and order therein) this would cause sequential reads which ought to be much faster.

Having no knowledge of the underlying implementation, the question would be: what's necessary to implement this? Are there any architectural features limiting the implementation of this? Is it realistic to do a full sort before the verify procedure and, if not, what could be alternatives (i.e. partial sorting, multiple sorting queues etc)?

kevina:
Closes #3922

### DIFF
--- a/filestore/util.go
+++ b/filestore/util.go
@@ -2,6 +2,7 @@ package filestore
 
 import (
 	"fmt"
+	"sort"
 
 	"github.com/ipfs/go-ipfs/blocks/blockstore"
 	pb "github.com/ipfs/go-ipfs/filestore/pb"
@@ -89,7 +90,10 @@ func List(fs *Filestore, key *cid.Cid) *ListRes {
 // one by one each block in the Filestore's FileManager.
 // ListAll does not verify that the references are valid or whether
 // the raw data is accessible. See VerifyAll().
-func ListAll(fs *Filestore) (func() *ListRes, error) {
+func ListAll(fs *Filestore, fileOrder bool) (func() *ListRes, error) {
+	if fileOrder {
+		return listAllFileOrder(fs, false)
+	}
 	return listAll(fs, false)
 }
 
@@ -105,7 +109,10 @@ func Verify(fs *Filestore, key *cid.Cid) *ListRes {
 // returns one by one each block in the Filestore's FileManager.
 // VerifyAll checks that the reference is valid and that the block data
 // can be read.
-func VerifyAll(fs *Filestore) (func() *ListRes, error) {
+func VerifyAll(fs *Filestore, fileOrder bool) (func() *ListRes, error) {
+	if fileOrder {
+		return listAllFileOrder(fs, true)
+	}
 	return listAll(fs, true)
 }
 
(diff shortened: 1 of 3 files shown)


dokterbob:
Just tested it and it's uncomparably faster! Thanks!

filestore/util.go
```diff
@@ -89,8 +90,14 @@ func List(fs *Filestore, key *cid.Cid) *ListRes {
 // one by one each block in the Filestore's FileManager.
 // ListAll does not verify that the references are valid or whether
 // the raw data is accessible. See VerifyAll().
-func ListAll(fs *Filestore) (func() *ListRes, error) {
-	return listAll(fs, false)
+func ListAll(fs *Filestore, fileOrder bool) (func() *ListRes, error) {
+	if fileOrder {
+		println("ListAll With Sort")
```
whyrusleeping: don't use println

filestore/util.go
```diff
@@ -158,6 +169,93 @@ func next(qr dsq.Results) (*cid.Cid, *pb.DataObj, error) {
...
+	}
+	sort.Sort(entries)
+
+	i := 0
+	return func() *ListRes {
+		if i >= len(entries) {
+			return nil
+		}
+		v := entries[i]
+		i += 1
```
whyrusleeping: `i++`

filestore/util.go
kevina: sorry, that was meant for debugging and I forgot to remove it

kevina:
@whyrusleeping 
> A couple comments

Fixed.

whyrusleeping:
@kevina also a couple codeclimate issues

kevina:
@whyrusleeping I know. It is being two picky.  The check is "Indent Error Flow", but this is not an error flow.  There is nothing is the style guide at https://github.com/golang/go/wiki/CodeReviewComments about this, the style guide is specially about error handling.

kevina:
Also golint had this disclaimer in the README:

>The suggestions made by golint are exactly that: suggestions. Golint is not perfect, and has both false positives and false negatives. Do not treat its output as a gold standard. We will not be adding pragmas or other knobs to suppress specific warnings, so do not expect or require code to be completely "lint-free". In short, this tool is not, and will never be, trustworthy enough for its suggestions to be enforced automatically, for example as part of a build process. Golint makes suggestions for many of the mechanically checkable items listed in Effective Go and the CodeReviewComments wiki page.

whyrusleeping:
@kevina you have redundant else statements. It makes the code cleaner to remove them

kevina:
@whyrusleeping I don't agree, but its not worth an argument, fixed.

PR #3939, opened 2017-05-24
README.md: add Nix to the linux package managers

WilliButz:


README.md
```diff
@@ -71,6 +71,7 @@ From there:
 
 - [Arch Linux](#arch-linux)
 - [Snap](#snap)
+- [Nix](#nix)
```
Kubuxu: Let's to alphabetical order here and bellow as then distro wars can't happen about the order.

README.md
```diff
@@ -83,6 +84,12 @@ Development version of go-ipfs is also on AUR under
 [go-ipfs-git](https://aur.archlinux.org/packages/go-ipfs-git/).
 You can install it using your favourite AUR Helper or manually from AUR.
 
+### Nix
+
+For Linux and MacOSX you can use the purely functional package manager [Nix](https://nixos.org/nix/):
+
+    $ nix-env -i ipfs
```
mateon1: We should probably mention the attribute name (also `ipfs`) as well.

WilliButz:
Okay, I'ts updated now @mateon1, @Kubuxu  :)

README.md
```diff
@@ -83,6 +84,18 @@ Development version of go-ipfs is also on AUR under
...
+### Nix
+
+For Linux and MacOSX you can use the purely functional package manager [Nix](https://nixos.org/nix/):
+
+```
+$ nix-env -i ipfs
+```
+or
+```
+$ nix-env -iA nixos.ipfs
```
mateon1: The main channel isn't necessarily named `nixos`. For the Nix package manager on other systems this would be `nixpkgs`, and could be named anything. I'd add a note like "The attribute name for the package is also `ipfs`" or "If you want to install by attribute name, it's also `ipfs`" instead of this code block.
Sorry for being so picky.

WilliButz: Oh I appreciate the pickiness :)

PR #3940, opened 2017-05-24
make odds of 'process added after close' panic less likely

whyrusleeping:
Theres still a very small chance it could happen, but this makes it much much less likely.

License: MIT


Kubuxu:
Feels like an band-aid (the proper solution would be not to panic there and return error) but LGTM.

whyrusleeping:
@Kubuxu yeah... its a bandaid. I don't feel like attempting to redesign goprocess right now though

It probably does make sense to have `Go` return an error.

PR #3518, opened 2016-12-19
Implemented experimental corenet application protocol support.

tswindell:
Signed-off-by: Tom Swindell <t.swindell@rubyx.co.uk>

PR #3943, opened 2017-05-26
Implemented experimental ptp(corenet) interface

magik6k:
This is built on #3518.

TODO:
- [x] Write some more tests/get good coverage
- [x] Test in some small real-world use
- [x] Cleanup the code a bit more


core/commands/corenet.go
```diff
@@ -0,0 +1,566 @@
...
+	}
+}
+
+//TODO: Ideally I'd like to see these combined into a module in core.
+var apps cnAppRegistry
+var streams cnStreamRegistry
+
+var CorenetCmd = &cmds.Command{
+	Helptext: cmds.HelpText{
+		Tagline: "Application network streams.",
```
whyrusleeping: tagline: "libp2p stream mounting"
shortdescription: "expose a local application to remote peers over libp2p

Note: this command is experimental and subject to change as usecases and APIs are refined"

cc @Kubuxu @lgierth

whyrusleeping:
Lets add a config option `Experimental.Libp2pStreamMounting` and only enable this command if that variable is set to true.

test/sharness/t0180-corenet.sh
```diff
@@ -26,7 +26,16 @@ test_expect_success "test ports are closed" '
   (! (netstat -ln | grep "LISTEN" | grep ":10102 "))
 '
 
-test_expect_success 'start ipfs listener' '
+test_expect_success 'fail without config option being enabled' '
+  ! ipfsi 0 exp corenet ls
```
Kubuxu: Use `test_must_fail` instead of `!` here.

whyrusleeping:
Looking good. The one last thing to consider now is whether or not we should put it under the 'exp' subcommand, or just keep it in the normal namespace. The other experimental commands (filestore and pubsub) exist in the top level, so i'm leaning towards doing that. Also not sure on the naming of 'corenet'.

What do you think @Kubuxu @lgierth @hsanjuan ? (also, @hsanjuan you might be able to rework ipfs cluster to use this feature once its merged :D )

whyrusleeping:
I'm leaning towards having this in the top level namespace (not under an `exp` command). The whole exp idea came before the config setting anyways.

magik6k:
I still want/need to move all the registry logic to a root package, and make it have instance instead of 1 global now. I don't like the naming too, but after trying to come up with something short that would make sense (other than p2pnet, but I'm not sure) I decided to leave `corenet`, though if you can think of anything that would fit I'll happily change it.

Kubuxu:
I am also not really attached to the name.

lgierth:
I'll look at this tomorrow -- awesome work! :)

magik6k:
Should be ready for a review now.

core/commands/corenet.go
```diff
@@ -0,0 +1,520 @@
...
+			w.Flush()
+
+			return buf, nil
+		},
+	},
+}
+
+var corenetListenCmd = &cmds.Command{
+	Helptext: cmds.HelpText{
+		Tagline: "Create application protocol listener and proxy to network multiaddr.",
```
Kubuxu: This should have helptext as it is more complex command.

core/commands/corenet.go
```diff
@@ -0,0 +1,520 @@
...
+			res.SetError(err, cmds.ErrNormal)
+			return
+		}
+
+		if !n.OnlineMode() {
+			res.SetError(errNotOnline, cmds.ErrClient)
+			return
+		}
+
+		addr, peer, err := ParsePeerParam(req.Arguments()[0])
```
Kubuxu: It should be possible to connect only with PeerID, is that possible with this code?

core/commands/root.go
```diff
@@ -47,6 +47,7 @@ ADVANCED COMMANDS
   pin           Pin objects to local storage
   repo          Manipulate the IPFS repository
   stats         Various operational stats
+  exp           Experimental commands
```
Kubuxu: this doesn't exist anymore and corenet should be added here

core/commands/corenet.go
```diff
@@ -0,0 +1,520 @@
...
+					continue
+				}
+				a.Close()
+				if !closeAll {
+					break
+				}
+			}
+		}
+
+		if len(req.Arguments()) != 1 {
```
Kubuxu: not needed

core/commands/corenet.go
```diff
@@ -0,0 +1,520 @@
...
+		}
```
Kubuxu: Do you remove mux handler anywhere here?

core/commands/corenet.go
magik6k: It is

core/commands/corenet.go
magik6k: a.Close above invokes corenet/net/net.go/IpfsListener.Close

Kubuxu: Yeah, that is why it would be nice to move all logic to its package. It wasn't clear.

corenet/apps.go
```diff
@@ -0,0 +1,61 @@
...
+	Apps []*AppInfo
+}
+
+// Register registers appInfo in this registry
+func (c *AppRegistry) Register(appInfo *AppInfo) {
+	c.Apps = append(c.Apps, appInfo)
+}
+
+// Deregister deregisters protocol handler from this registry
+func (c *AppRegistry) Deregister(proto string) {
```
Kubuxu: This should return error if it can't remove it.

ptp/apps.go
```diff
@@ -0,0 +1,65 @@
...
+
+// Close closes the listener. Does not affect child streams
+func (c *AppInfo) Close() error {
+	c.Closer.Close()
+	err := c.Registry.Deregister(c.Protocol)
+	return err
+}
+
+// AppRegistry is a collection of local application protocol listeners.
+type AppRegistry struct {
```
whyrusleeping: probably want to add some locking here to make it threadsafe. I can definitely see people running these commands in parallel

ptp/apps.go
```diff
@@ -0,0 +1,65 @@
...
+	Apps []*AppInfo
+}
+
+// Register registers appInfo in this registry
+func (c *AppRegistry) Register(appInfo *AppInfo) {
+	c.Apps = append(c.Apps, appInfo)
+}
+
+// Deregister deregisters protocol handler from this registry
+func (c *AppRegistry) Deregister(proto string) error {
```
whyrusleeping: will also want to add a listing method that does proper locking

lgierth:
Awesome work :):) :palm_tree: I think the use cases here are a bit underexplained, I don't see any showstoppers with the approach taken, but I'm just not clear enough on the use cases :) Is it just a simple 1-to-1 netcat? Or 1-to-n? What is the UX for accepting more than one connection?

Design questions:
- Why involve listeners and dialers at all? This feels like the layer where you wouldn't deal with IP addresses anymore, only with PeerIDs (or with whatever `ipfs swarm connect` would accept). And, if you bind an `App` to a listener here, wouldn't that also involve all other stream handlers too?
- It's worth considering generalizing this more, to something that just simply manages and exposes stream handlers, connections, and streams. Identify, Bitswap, and the others can be just special cases of that, where go-ipfs itself already has the code to mount them. That'd mean you could adhoc enable or disable the core protocols of IPFS, kill individual streams with individual peers, etc.
- A propos Identify, will Identify announce what's called Apps here?

I haven't thoroughly CRed the code itself yet, but I think we can keep this whole package free of dependencies on go-ipfs code. There are a few instances where you depend on `core.IpfsNode` but actually just use its `go-libp2p-host.Host` and `go-libp2p-peer.ID`.

magik6k:
For now it's more of a tunneling thing (like ssh port forwarding). For more than one connection I'm thinking of a dial option to make it forward more than one connection (so behave like ssh forwarding), perhaps even making it the default behavior and have option for one-time dial.

* I decided to use dialers as there are no easily muxable API systems yet (websocket/custom proto/etc.). I'm not really sure what you mean by other stream handlers being involved, but the answer is probably no - have you seen `ptp/net/net.go`?
* With this code you shouldn't be able to manage built-in stream handlers - the ones used here are prefixed with `/app/` prefix(which I will likely change to `/ptp/`). For now it's only tunnels over libp2p.
* I don't really know what do you mean by Identify here.

Other than that, some naming conventions are still going to change to make things more clear (`app` was left from the base PR).

whyrusleeping:
@lgierth identify should announce these protocols as it just grabs whatever
is registered on the host.

I think we can keep designing this as we move forward. It's clearly marked
as experimental so making api and design changes post merge is perfectly
acceptable. Once people start using it and requesting different use cases,
we will have a much better idea of what exactly is needed to generalize it
better.

core/commands/ptp.go
```diff
@@ -0,0 +1,392 @@
...
+		err = checkEnabled(n)
+		if err != nil {
+			res.SetError(err, cmds.ErrNormal)
+			return
+		}
+
+		if !n.OnlineMode() {
+			res.SetError(errNotOnline, cmds.ErrClient)
+			return
+		}
```
Kubuxu: IMO I would compress code above into once function (returning node or error in case those checks fail) as it is used everywhere.

core/commands/ptp.go
```diff
@@ -0,0 +1,392 @@
...
+
+		if !closeAll {
+			if len(req.Arguments()) == 0 {
+				res.SetError(errors.New("no handlerID nor listener protocol specified"), cmds.ErrNormal)
+				return
+			}
+
+			handlerID, err = strconv.ParseUint(req.Arguments()[0], 10, 64)
+			if err != nil {
+				proto = "/ptp/" + req.Arguments()[0]
```
Kubuxu: What if someone creates protocol that is identified by decimal numbers?

If this is a problem right now, please fix it and create sharness test for it.

core/commands/ptp.go
```diff
@@ -0,0 +1,392 @@
...
+			for _, listener := range n.PTP.Listeners.Listeners {
+				if !closeAll && listener.Protocol != proto {
+					continue
+				}
+				listener.Close()
+				if !closeAll {
+					break
+				}
+			}
+		}
```
Kubuxu: The logic here is quite complex, anyway to make it much simpler to read?

Probably it would be better to completly split off the close all, with handler, and with protocol into separate functions. Maybe even place them in the ptp package.

test/dependencies/ma-pipe-unidir/main.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	defer conn.Close()
+	switch mode {
+	case "recv":
+		io.Copy(os.Stdout, conn)
+	case "send":
+		io.Copy(conn, os.Stdin)
+	default:
+		//TODO: a bit late
```
Kubuxu: Resolve this todo.

Kubuxu:
You have published the package of go-random in weird state.

Make sure that imports in it are not rewritten (`gx-go uw`, and then publish it).

whyrusleeping:
Here goes! :)

PR #3948, opened 2017-05-31
core: make announced swarm addresses configurable

### LINKED ISSUE #3613: More granular address announcements
Sometimes a go-ipfs listener can not directly bind to its external address. This is e.g. the case with NAT, where it binds to an IP address in the local network, while other nodes dial it using the local router's public IP address. We jump through all kinds of hoops to work around that, and one of these is "observed addresses", where other nodes tell us which addresses they've seen us using.

Another example are HTTP-based transports like /ws, which will in many cases be behind an SSL-terminating reverse proxy. Bound address: /ws, external address: /wss. There is a way to discover that we're behind a reverse proxy (by passing a special HTTP header), but that would only work after the first request has come in. Until then, the node would be unaware of its /wss address.

We could just as well have an ExternalAddresses config option though (or ideally both). This might also be useful for announcing /dns addresses. The default bootstrappers will eventually be bootstrap.libp2p.io or so, and they'd be able to announce it.

lgierth:
Fixes #3613

- Tests failing because it depends on libp2p/go-libp2p#196 which I haven't bubbled up yet
- I need to update a bunch of docs regarding the Addresses config section
- A tiny code comment here and there might be good as well

Overall this is good for CR though.

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -45,6 +45,7 @@ import (
 	pnet "gx/ipfs/QmNMCAuxnQFHLGWcvay3DmVFrKuY6Y2nsc9vzsf4gVouJV/go-libp2p-pnet"
 	pstore "gx/ipfs/QmPgDWmTmuzvP7QE5zwo1TmjbJme9pmZHNujB2453jkCTr/go-libp2p-peerstore"
 	routing "gx/ipfs/QmPjTrrSfE6TzLv6ya6VWhGcCgPrUAdcgrDcQyRDX2VyW1/go-libp2p-routing"
+	mafilter "gx/ipfs/QmQBB2dQLmQHJgs2gqZ3iqL2XiuCtUCvXzWt5kMXDf5Zcr/go-maddr-filter"
 	ipnet "gx/ipfs/QmQq9YzmdFdWNTDdArueGyD7L5yyiRQigrRHJnTGkxcEjT/go-libp2p-interface-pnet"
 	dht "gx/ipfs/QmRKEzkaiwud2LnwJ9CgBrKw122ddKPTMtLizV3DNimVRD/go-libp2p-kad-dht"
 	p2phost "gx/ipfs/QmRNyPNJGNCaZyYonJj7owciWTsMd9gRfEKmZY3o6xwN3h/go-libp2p-host"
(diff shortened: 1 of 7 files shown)


core/core.go
```diff
@@ -211,8 +212,17 @@ func (n *IpfsNode) startOnlineServices(ctx context.Context, routingOption Routin
...
+	addrsFactory, err := makeAddrsFactory(cfg.Addresses)
+	if err != nil {
+		return err
+	}
+
+	hostopts := &ConstructPeerHostOpts{
+		AddrsFactory:      addrsFactory,
+		DisableNatPortMap: cfg.Swarm.DisableNatPortMap,
+	}
 	peerhost, err := hostOption(ctx, n.Identity, n.Peerstore, n.Reporter,
```
whyrusleeping: maybe we could also have a `ConstructSwarmOpts` that holds all these other arguments.

core/core.go
```diff
@@ -260,6 +270,50 @@ func (n *IpfsNode) startOnlineServices(ctx context.Context, routingOption Routin
...
+		if err != nil {
+			return nil, err
+		}
+		noAnnAddrs[maddr.String()] = true
+	}
+
+	return func(allAddrs []ma.Multiaddr) []ma.Multiaddr {
+		var addrs []ma.Multiaddr
+		if len(annAddrs) > 0 {
+			addrs = annAddrs[:]
```
whyrusleeping: Why the reslice?

core/core.go
lgierth: Once libp2p/go-libp2p#197 is merged I'd like to remove ConstructPeerHostOpts and use basichost.HostOpts, and would like to do the same with regard to swarm.

core/core.go
lgierth: yeah good question :) removing that

lgierth: Oh yeah right. We build `annAddrs` once and cache it outside the closure, and the reslice is to prevent multiple calls to the closure from getting back the same slice. It's gonna be equal, but not the same.

whyrusleeping:
@lgierth what is the next step here?

lgierth:
This should be good to go -- I rebased, let's see what CI says.

lgierth:
Oh right this depends on a libp2p update. Will take care of that later today.

whyrusleeping:
@lgierth you should be unblocked here now :)

Kubuxu:
Please rebase.

lgierth:
I rebased, and looked at that reslice once more, it's gone now.

Jenkins green

lgierth:
@Kubuxu what was it you committed here, just another rebase?

PR #3949, opened 2017-05-31
apply the megacheck code vetting tool for idiomatic go

zramsay:
- `megacheck` applies each of the `unused`, `gosimple`, and `staticcheck` tools to the code base in order to enforce idiomatic golang. [See here](https://github.com/dominikh/go-tools) for more information.
- Only the least controversial recommendations were applied, e.g., some unused functions were left as they may have an intended use. Unused variables were commented out to reduce the number of recommendations provided by the tool. Other recommended changes I skipped because the concepts are new to me.

### Main Changes 
- error handling: `if err != nil { return err }; return nil` was simplified to `return err`
- bool checks: `if pin == false` is the same as `if !pin`
- replaced the deprecated `os.SEEK_*` with `io.Seek*`
- specifying the capacity in a `make()` was dropped if set to the same as its length.
- replaced `time.Now().Sub` with `time.Since()`
- for loops that could be replaced with `append(a, b...)`, were

### DIFF
--- a/Rules.mk
+++ b/Rules.mk
@@ -66,7 +66,13 @@ include $(dir)/Rules.mk
 # -------------------- #
 #     core targets     #
 # -------------------- #
+PACKAGES_NOVENDOR := $(shell go list github.com/ipfs/go-ipfs/... | grep -v /Godeps/)
 
+megacheck:
+	@go get honnef.co/go/tools/cmd/megacheck
+	@for pkg in ${PACKAGES_NOVENDOR}; do megacheck "$$pkg"; done
+
+.PHONY: megacheck
 
 build: $(TGT_BIN)
 .PHONY: build
--- a/Rules.mk
+++ b/Rules.mk
@@ -66,13 +66,6 @@ include $(dir)/Rules.mk
 # -------------------- #
 #     core targets     #
 # -------------------- #
-PACKAGES_NOVENDOR := $(shell go list github.com/ipfs/go-ipfs/... | grep -v /Godeps/)
-
-megacheck:
-	@go get honnef.co/go/tools/cmd/megacheck
-	@for pkg in ${PACKAGES_NOVENDOR}; do megacheck "$$pkg"; done
-
-.PHONY: megacheck
 
 build: $(TGT_BIN)
 .PHONY: build
@@ -149,6 +142,7 @@ help:
 	@echo '  test_go_short'
 	@echo '  test_go_expensive'
 	@echo '  test_go_race'
+	@echo '  test_go_megacheck'	 - Run the `megacheck` vetting tool
 	@echo '  test_sharness_short'
 	@echo '  test_sharness_expensive'
 	@echo '  test_sharness_race'
--- a/cmd/ipfs/main.go
+++ b/cmd/ipfs/main.go
@@ -38,16 +38,16 @@ import (
 var log = logging.Logger("cmd/ipfs")
 
 var (
-	errUnexpectedApiOutput = errors.New("api returned unexpected output")
-	errApiVersionMismatch  = errors.New("api version mismatch")
-	errRequestCanceled     = errors.New("request canceled")
+	// errUnexpectedApiOutput = errors.New("api returned unexpected output")
+	// errApiVersionMismatch  = errors.New("api version mismatch")
+	errRequestCanceled = errors.New("request canceled")
 )
 
 const (
 	EnvEnableProfiling = "IPFS_PROF"
 	cpuProfile         = "ipfs.cpuprof"
 	heapProfile        = "ipfs.memprof"
-	errorFormat        = "ERROR: %v\n\n"
+	// errorFormat        = "ERROR: %v\n\n"
 )
 
 type cmdInvocation struct {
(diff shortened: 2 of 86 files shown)


zramsay:
the failure on codeclimate is a false positive (calling `context.TODO()` is valid and not equal to `// TODO`. I've submitted a ticket to them.

cmd/ipfs/main.go
```diff
@@ -38,16 +38,16 @@ import (
 var log = logging.Logger("cmd/ipfs")
 
 var (
-	errUnexpectedApiOutput = errors.New("api returned unexpected output")
-	errApiVersionMismatch  = errors.New("api version mismatch")
-	errRequestCanceled     = errors.New("request canceled")
+	// errUnexpectedApiOutput = errors.New("api returned unexpected output")
+	// errApiVersionMismatch  = errors.New("api version mismatch")
```
whyrusleeping: if its actually not being used, should probably just remove it

cmd/ipfswatch/main.go
```diff
@@ -99,7 +100,7 @@ func run(ipfsPath, watchPath string) error {
 	}
 
 	interrupts := make(chan os.Signal)
-	signal.Notify(interrupts, os.Interrupt, os.Kill)
+	signal.Notify(interrupts, os.Interrupt, syscall.SIGTERM)
```
whyrusleeping: is `os.Kill` deprecated?

routing/mock/centralized_test.go
```diff
@@ -45,7 +45,7 @@ func TestClientFindProviders(t *testing.T) {
 	providersFromClient := client.FindProvidersAsync(context.Background(), k, max)
 	isInClient := false
 	for pi := range providersFromClient {
-		if pi.ID == pi.ID {
+		if pi.ID == pi.ID { // <-- typo?
```
whyrusleeping: this code is so old and unused... We can probably just remove routing/mock and the supernode routing stuff in general (as a separate PR)

whyrusleeping:
Also, if you don't mind, could you address the codeclimate issues? (if they feel like 'not my problem' then go ahead and skip them, its a bit aggressive on what it wants you to do)

cmd/ipfswatch/main.go
zramsay: the error message is: `os.Kill cannot be trapped (did you mean syscall.SIGTERM?) (SA1016)` where SA1016 -> Trapping a signal that cannot be trapped

cmd/ipfs/main.go
zramsay: roger.

zramsay:
sounds good, will address. What value is the codeclimate integration providing over and above implementing [this TODO](https://github.com/ipfs/go-ipfs/blob/master/test/bin/verify-go-fmt.sh#L3) ? IMO the integration should be dropped in favour of implementing that as part of `make test` (see last line on original PR comment). Happy to pick that up, and include megacheck enforcement.

routing/mock/centralized_test.go
zramsay: #3950

Rules.mk
```diff
@@ -66,7 +66,13 @@ include $(dir)/Rules.mk
 # -------------------- #
 #     core targets     #
 # -------------------- #
+PACKAGES_NOVENDOR := $(shell go list github.com/ipfs/go-ipfs/... | grep -v /Godeps/)
 
+megacheck:
+	@go get honnef.co/go/tools/cmd/megacheck
+	@for pkg in ${PACKAGES_NOVENDOR}; do megacheck "$$pkg"; done
+
+.PHONY: megacheck
```
Kubuxu: Please add this target to `mk/golang.mk` instead and add comment about it to help print section.

unixfs/hamt/hamt_stress_test.go
```diff
@@ -189,104 +186,3 @@ func genOpSet(seed int64, keep, temp []string) []testOp {
 		}
 	}
 }
-
-// executes the given op set with a repl to allow easier debugging
-func debugExecuteOpSet(ds dag.DAGService, width int, ops []testOp) (*HamtShard, error) {
```
Kubuxu: This probably shouldn't be removed. It could be useful hot-plug debugging tool.

cc @whyrusleeping

zramsay: reinstated it but commented it out

Rules.mk
zramsay: done

zramsay:
circle will probably pass after re-triggered. I've notice some non-determinism with the tests (e.g., jenkins failed randomly on a commit where the only change was to `.codeclimate.yml`

Kubuxu:
Test are deterministic just build machines of free service providers sometimes hang on our network tests.

core/commands/files/files.go
```diff
@@ -664,11 +664,6 @@ stat' on the file or any of its ancestors.
 			return
 		}
 
-		var r io.Reader = input
-		if countfound {
-			r = io.LimitReader(r, int64(count))
```
whyrusleeping: Hrm... Maybe you discovered a bug here. I think the copy below was supposed to be from 'r' instead of 'input'

zramsay: makes sense. The tests are happy with the fix.

whyrusleeping:
@zramsay Thanks a bunch! This is really useful :) If there are more, feel free to file more PRs

zramsay:
no problem! There a couple dozen ~more complex fixes that I'll tackle as part of #3953

PR #3951, opened 2017-06-01
Make sure non-self keys also get republished

### LINKED ISSUE #3808: Non-self IPNS names need republishing
#### Description:

It doesn't look like names added by `ipfs key gen` ever get republished. I didn't actually *test* it, but I'm fairly certain there's no code for it.

It should:

- at startup, call `Republisher.AddName()` for each key in the keystore.
- same for every run of `ipfs key gen`, add the name to the republisher.

cc @edsilv I think this will affect you. You'll have to run `ipfs name publish --key=<name> <hash>` every 12 hours, for now.

kevina:
Closes #3808

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -366,8 +366,7 @@ func (n *IpfsNode) setupIpnsRepublisher() error {
 		return err
 	}
 
-	n.IpnsRepub = ipnsrp.NewRepublisher(n.Routing, n.Repo.Datastore(), n.Peerstore)
-	n.IpnsRepub.AddName(n.Identity)
+	n.IpnsRepub = ipnsrp.NewRepublisher(n.Routing, n.Repo.Datastore(), n.PrivateKey, n.Repo.Keystore())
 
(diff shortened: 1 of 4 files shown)


whyrusleeping:
:+1: i really like the approach. Will make it easy when we migrate to having the primary identity key in the keystore.

kevina:
@whyrusleeping thanks, do you want me to go ahead and rebase and squash this?

whyrusleeping:
@kevina yeah, i'd go ahead and move forward here

kevina:
@whyrusleeping rebased and squashed.

kevina:
@whyrusleeping I made one last tweak.  I kept it is a separate commit so you can see what is changed.  I can squash it if you want.

whyrusleeping:
@kevina as far as i can tell, entrylock is no longer used either. Good catch

whyrusleeping:
Thanks @kevina :)

PR #3952, opened 2017-06-01
bitswap: limit concurrency of bitswap

Kubuxu:
This commit solves goroutine explsion in bitswap.
It creates following:
  - 8 goroutines for handling of the DHT requests
  - 8*8 goroutines for handling connecting with peers

This means that at most ~200 (192) connections will be created at one
time.

License: MIT

### DIFF
--- a/exchange/bitswap/workers.go
+++ b/exchange/bitswap/workers.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	bsmsg "github.com/ipfs/go-ipfs/exchange/bitswap/message"
+	bsnet "github.com/ipfs/go-ipfs/exchange/bitswap/network"
 
 	process "gx/ipfs/QmSF8fPo3jgVBAy8fpdjjYqgG87dkJgUprRBHRd2tmfgpP/goprocess"
 	procctx "gx/ipfs/QmSF8fPo3jgVBAy8fpdjjYqgG87dkJgUprRBHRd2tmfgpP/goprocess/context"
@@ -15,7 +16,7 @@ import (
 	peer "gx/ipfs/QmdS9KpbDyPrieswibZhkod1oXqRwZJrUPzxCofAMWpFGq/go-libp2p-peer"
 )
 
-var TaskWorkerCount = 8
+const TaskWorkerCount = 8
 
 func (bs *Bitswap) startWorkers(px process.Process, ctx context.Context) {
 	// Start up a worker to handle block requests this node is making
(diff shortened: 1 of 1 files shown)


whyrusleeping:
@Kubuxu the provider manager was supposed to be responsible for limiting those dht queries. Can we just fix the limiting on it instead of adding another set of worker logic?

Kubuxu:
It isn't problem with limiting DHT. It is problem with starting too many `ConnectTo` calls.

exchange/bitswap/workers.go
```diff
@@ -215,35 +274,20 @@ func (bs *Bitswap) providerQueryManager(ctx context.Context) {
...
-				activeLk.Lock()
-				kset.Remove(e.Cid)
-				activeLk.Unlock()
-			}(e)
+			ksetLk.Unlock()
 
+			select {
+			case searchSemaphore <- struct{}{}:
+				go bs.dhtWorker(e.Ctx, kset, ksetLk, e)
+				<-searchSemaphore
```
whyrusleeping: this returns the semaphore immediately, you probably want to do this at the end of the dhtWorker routine.

exchange/bitswap/workers.go
```diff
@@ -215,35 +274,20 @@ func (bs *Bitswap) providerQueryManager(ctx context.Context) {
...
-					}(p)
-				}
-				wg.Wait()
-				activeLk.Lock()
-				kset.Remove(e.Cid)
-				activeLk.Unlock()
-			}(e)
+			ksetLk.Unlock()
 
+			select {
```
whyrusleeping: this will still block callers if the semaphore fills up. Take a look at the providesCollector code, I have an expandable buffer to help avoid those issues

lgierth:
This has been running on earth and pluto for the past three weeks -- they didn't show any exceptional behaviour, but it's generally hard to tell at the moment... I'll update them to v0.4.10-rc1 tonight.

Kubuxu:
The version that is currently in this PR is a bit broken. I will fix it after I am done with exam session.

Kubuxu:
I am looking into some lib that would handle it.

exchange/bitswap/workers.go
```diff
@@ -215,35 +276,22 @@ func (bs *Bitswap) providerQueryManager(ctx context.Context) {
 			default:
 			}
 
-			activeLk.Lock()
+			ksetLk.Lock()
 			if kset.Has(e.Cid) {
```
whyrusleeping: might want to use `Visit` here: https://github.com/ipfs/go-cid/blob/master/set.go#L47

whyrusleeping:
@Kubuxu are you planning on working on this more?

Kubuxu:
If I can hand it over, I would prefer to do so. It is a can of worms that keeps exploding every time I touch it.

Stebalien:
Bitswap has been refactored and extracted since this was submitted.

PR #3954, opened 2017-06-02
docker: expose port 8081 for /ws listener

lgierth:
Opening this again because it's been a reality in ipfs/infrastructure for a while.

Yes, the `/ws` listener is soon gonna be part of the gateway and a dedicated port no longer neccessary, but until then the lack of this exposed port is a pain.


whyrusleeping:
LGTM, youre the dockerfile overlord.

PR #3959, opened 2017-06-04
Added info on compiling on FreeBSD.

Sag0Sag0:


whyrusleeping:
Thanks! Sorry about gitcop... its rather picky

Issue #3839, opened 2017-03-28
Let 'ipfs dht findprovs' return more than 20 peers


VictorBjelkholm:
#### Version information:
```
go-ipfs version: 0.4.8-rc1-
Repo version: 5
System version: amd64/darwin
Golang version: go1.8
```

#### Type:
Enhancement

#### Priority:
P5

#### Description:
Currently when running `ipfs dht findprovs` it'll continue until a timeout or finding 20 peers. It would be helpful if we could configure the command not return just because it found 20 peers. In addition, would be helpful to be able to configure the amount of time it'll search as well.

Relevant line: https://github.com/ipfs/go-ipfs/blob/cbf8a0cdb4af60c5106fc41d03902272ac72be6e/core/commands/dht.go#L162

Kubuxu:
There is global ipfs option, `timeout` that causes the command to timeout after some time.

Example: `ipfs --timeout=1s dht findprovs /ipfs/QmVZxZgdnZ8h2L7V5bhLsE4CvAmAaGyLjBhFJHYigPE9X1`

lgierth:
Needs to be documented -- it's not in `ipfs --help`

faddat:
is this what one would use to identify the peers serving a given file?

Kubuxu:
Yes `findprovs` can be used for that.

Kubuxu:
It should be very easy to do, just making https://github.com/ipfs/go-ipfs/blob/master/core/commands/dht.go#L162 a CLI parameter and adding some tests for it.

PR #3966, opened 2017-06-10
Add 'ipfs dht findprovs --num-providers' to allow choosing number of providers to find

torresashjian:
Closes: #3839

I was looking for ways to add tests to findProviders command when new option --num-providers is given. But could not find a good example on existing commands to test options.

One possibility is to use a mocking library to mock the Node, but I didn't want to introduce new library dependencies for such a trivial change.

Please advice if new tests are needed.

License: MIT
Signed-off-by: Miguel Torres <migueltorreslopez@gmail.com>

### DIFF
--- a/core/commands/dht.go
+++ b/core/commands/dht.go
@@ -145,6 +145,7 @@ var findProvidersDhtCmd = &cmds.Command{
 	},
 	Options: []cmds.Option{
 		cmds.BoolOption("verbose", "v", "Print extra information.").Default(false),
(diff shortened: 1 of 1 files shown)


core/commands/dht.go
```diff
@@ -159,7 +160,15 @@ var findProvidersDhtCmd = &cmds.Command{
 			return
 		}
 
-		numProviders := 20
+		numProviders, _, err := res.Request().Option("n").Int()
```
whyrusleeping: nitpick: would prefer using the long option name for the lookup here

PR #3967, opened 2017-06-10
rename and fix sharness issue in ptp test

whyrusleeping:
License: MIT

### DIFF
--- a/test/sharness/t0180-p2p.sh
+++ b/test/sharness/t0180-p2p.sh
@@ -26,8 +26,8 @@ test_expect_success "test ports are closed" '
   (! (netstat -ln | grep "LISTEN" | grep ":10102 "))
 '
 
-test_must_fail 'fail without config option being enabled' '
-  ipfsi 0 p2p stream ls
+test_expect_success 'fail without config option being enabled' '
+  test_must_fail ipfsi 0 p2p stream ls
 '
 
 test_expect_success "enable filestore config setting" '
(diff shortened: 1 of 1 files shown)


whyrusleeping:
@magik6k the ptp tests appear to fail on OSX. Any ideas on a quick fix?

whyrusleeping:
Ah, i think as someone else mentioned, its like the use of 'wait' having different semantics between platforms

magik6k:
What /might/ happen is that the $server_pid process finishes before the `wait` even starts, it (`wait`) then gets the $server_pid and locks up the test. It sounds really, really unlikely, though it's the only logical thing that would affect `wait`.

Simplest thing I can think of now is to use pid files for ma-pipe-.., like in https://github.com/ipfs/go-ipfs/blob/master/test/sharness/t0180-ptp.sh#L119-L126, this way the tests wouldn't lockup at least(assuming it's not a wait issue). Should I do that?

whyrusleeping:
@magik6k i'm pretty sure wait doesnt behave the same on OSX as it does on linux. I think i remember something about this happening before. cc @Kubuxu @chriscool

Kubuxu:
I am not sure, I recall issues with netcat not exiting but not that.

whyrusleeping:
@magik6k i think we can use a loop that calls `kill -0 $PID` until it fails as a cheap 'wait'

magik6k:
I rebased on current master and fixed the wait issue.

PR #3969, opened 2017-06-10
fix shutdown check in t0023

whyrusleeping:
License: MIT

### DIFF
--- a/test/sharness/t0023-shutdown.sh
+++ b/test/sharness/t0023-shutdown.sh
@@ -17,7 +17,11 @@ test_expect_success "shutdown succeeds" '
 '
 
 test_expect_success "daemon no longer running" '
-	test_expect_code  1 kill -0 $IPFS_PID
+	for i in $(test_seq 1 100)
+	do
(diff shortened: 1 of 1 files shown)


test/sharness/t0023-shutdown.sh
```diff
@@ -17,7 +17,11 @@ test_expect_success "shutdown succeeds" '
 '
 
 test_expect_success "daemon no longer running" '
-	test_expect_code  1 kill -0 $IPFS_PID
+	for i in $(test_seq 1 100)
+	do
+		go-sleep 100ms
+		! kill -0 $IPFS_PID 2>/dev/null && return
+	done
```
Kubuxu: I think it will never fail.

whyrusleeping: Its exactly the same code as the other kill check, and that one works

whyrusleeping: If `kill -0 $IPFS_PID` fails, then it returns. which seems right to me

Issue #3974, opened 2017-06-13
ipns pin add doesn't resolve sharded paths


Stebalien:
#### Version information:

go-ipfs version: 0.4.9-8ec0814a8
Repo version: 5
System version: amd64/linux
Golang version: go1.8.3

#### Type:

Bug

#### Severity:

High/Medium

#### Description:

`ipns pin add` doesn't work with sharded paths. To reproduce, cat the wikipedia index and then try pinning it:

```sh
ipfs cat /ipfs/QmXoypizjW3WknFiJnKLwHCnL72vedxjQkDDP1mXWo6uco/wiki/index.html
ipfs pin add /ipfs/QmXoypizjW3WknFiJnKLwHCnL72vedxjQkDDP1mXWo6uco/wiki/index.html
```

The second command fails with:

```
Error: pin: no link named "wiki" under QmXoypizjW3WknFiJnKLwHCnL72vedxjQkDDP1mXWo6uco
```

/cc @whyrusleeping

whyrusleeping:
This is because `ipfs pin add` calls `corerepo.Pin`, and that method uses the non-unixfs context resolver: https://github.com/ipfs/go-ipfs/blob/master/core/corerepo/pinning.go#L35

That should probably be amended to use a unixfs resolver as in: https://github.com/ipfs/go-ipfs/blob/master/core/coreunix/cat.go#L12

Issue #3910, opened 2017-05-09
ipfs resolve doesn't work with sharded dirs


Kubuxu:
#### Description:

Run `ipfs resolve /ipfs/Qm...AAA/file` where `Qm...AAA` is sharded.

whyrusleeping:
Solution is likely to add a unixfs contextual resolver to the resolve command (though maybe we want that as an option?)
see:  https://github.com/ipfs/go-ipfs/blob/master/core/commands/ls.go#L93

whyrusleeping:
@Stebalien has this been resolved?

Stebalien:
@whyrusleeping no. I only fixed pinning. Test case:

```go
ipfs resolve /ipfs/QmXoypizjW3WknFiJnKLwHCnL72vedxjQkDDP1mXWo6uco/wiki/index.html
```

However, this is a deeper issue. We need to switch to using the unixfs resolver for *all* `/ipfs` paths and then add a `/ipld` prefix for resolving IPLD paths (https://github.com/ipfs/notes/issues/248).

Stebalien:
Fixed in the CoreAPI refactor.

PR #3975, opened 2017-06-13
pinning + pathresolver: fix pinning/unpinning of sharded directories

Stebalien:
* Change ResolveToCid to take a Resolver and a NameSystem instead of an ipfs
  Node.
* Make the pin/unpin methods use a Unixfs path resolver.

Closes: #3974

License: MIT

### DIFF
--- a/core/commands/pin.go
+++ b/core/commands/pin.go
@@ -12,6 +12,7 @@ import (
 	dag "github.com/ipfs/go-ipfs/merkledag"
 	path "github.com/ipfs/go-ipfs/path"
 	pin "github.com/ipfs/go-ipfs/pin"
+	uio "github.com/ipfs/go-ipfs/unixfs/io"
 
 	context "context"
 	u "gx/ipfs/QmWbjfz3u6HkAdPh34dgPchGbQjob6LXLhAeCGii2TX69n/go-ipfs-util"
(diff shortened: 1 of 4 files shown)


Stebalien:
We may want to hold this pending the discussion in https://github.com/ipfs/go-ipfs/issues/3910#issuecomment-300380515 (as it's possible to pin IPLD objects).

whyrusleeping:
Hrm.... this is true. but the resolve is just being applied to the user given path. It shouldnt have any effect on pinning ipld objects.

cc @Kubuxu @lgierth

Stebalien:
Can't you path through IPLD objects? Also, whatever happened to `/ipld/...`. Having the pathing scheme *not* declared somewhere in the path is kind of terrible.

whyrusleeping:
You can path through ipld objects, that works fine. And shard nodes are very specific protobufs, so theres no ambiguity between a path in an ipld node and a path through a sharded node. The only issue happens when you want to address an intermediate shard element directly

core/commands/pin.go
```diff
@@ -377,13 +378,18 @@ new pin and removing the old one.
 			return
 		}
 
-		fromc, err := core.ResolveToCid(req.Context(), n, from)
+		r := &path.Resolver{
```
Kubuxu: This construction is more and more frequent around, it might be good to store it in ipfs node struct or have helper for it.

Stebalien: At the end of the day, I think a better solution would be to avoid this resolver dance altogether. The root of the issue comes from the fact that the path doesn't dictate how it should be interpreted: as an IPLD path through IPLD objects (where the segments are keys) or an IPFS path through potentially sharded IPFS directories (where the segments are directory entries). To work around this:

1. It's up to commands to chose the initial resolver.
2. Resolvers do some additional magic under the covers to detect if a node should be treated as an IPLD or IPFS object by looking at it's Cid (they *really* shouldn't be doing this).

The solution I've proposed is to use separate prefixes: `/ipld/...` and `/ipfs/...` (https://github.com/ipfs/notes/issues/248). I believe this is how this was *intended* to work (when designed) anyways.

If we did this,  we could pass around the protocol along with the Cid/relative path so that `core.Resolve` and `core.ResolveToCid` could pick the right resolver for the given protocol.

PR #3982, opened 2017-06-14
WIP: Round Robin PRQ

dgrisham:
This is a preliminary implementation of a round robin peer-request-queue (PRQ) for Bitswap.

## Motivation

The current PRQ orders peers based on a specified comparison function, which was originally intended to one day implement Bitswap strategies. However, this comparison function is limited in that it doesn't map very well onto the idea of Bitswap strategies. For example, it is difficult to say something like "Peer A has a 2x preference over Peer B, so satisfy Peer A's requests twice as much as Peer B's". However, a weighted round robin queue can easily satisfy such a request. For this reason, I'm working to replace the current PRQ implementation with one that uses a weighted round robin queue to send data to users based on their relative weights, and these weights are determined by a pluggable Bitswap strategy.

### RRQueue

The round robin queue is implemented as the `RRQueue` type:

```go
type RRQueue struct {
    peerBurst   int
    roundBurst  int
    strategy    Strategy
    weights     map[peer.ID]float64
    totalWeight float64
    allocations []RRPeer
}
```

A single round satisifies `roundBurst` requests. At the start of each round, the peers' `weights` (determined by the `Strategy`, discussed below) determine the number of requests each peer receives (as a proportion of `roundBurst`), which is stored in `allocations`. For example, if we have two peers as such:

```go
weights['peer1'] = 3
weights['peer2'] = 1
```

then `peer1` will be allocated 3x as many requests as `peer2`. So, if `roundBurst = 1000`, then `peer1` would be allocated 750 requests for this round, `peer2` would be allocated 250 requests. The `totalWeight` is simply used to reduce the number of calculations to find the sum of all `weights` and is used to calculate peer allocations (in this example, `totalWeight` would be equal to `3 + 1 = 4`. `allocations` is calculated for each round robin round, while `weights` persists between rounds and a peer's weight is updated on a `prq.Push`.

`peerBurst` determines the number of consecutive requests a peer will be served *within* the round. For example, if `peerBurst = 2` in the above example, then `peer1` would be served 2 of its 3 requests (then placed at the end of the queue), then `peer2` would be served its only request (and removed from the queue), then `peer1` would be served its final request. `roundBurst` and `peerBurst` are meant to be tunable parameters and are currently set to simple test values.

### Strategies

Bitswap strategies are implemented by the following types:

```go
type StrategyFunc func(r *Receipt) float64

type Strategy struct {
    weightFunction StrategyFunc
    freezeWeight   float64
}
```

A `StrategyFunc` takes in a Bitswap ledger (as an immutable `Receipt`) and calculates a weight to be used in the `RRQueue`. `freezeWeight` is used in freezing a peer, which is a concept that was a part of the previous PRQ implementation. To accommodate freezing, the `RRQueue` reduces a peer's round robin weight based on `freezeWeight` (e.g. `weights['peer1'] /= freezeWeight` when freezing `peer1`).

## Interface

The current implementation maintains the same outward-facing `prq` interface as the previous, with a single caveat: a `Strategy` must be provided to the `newPRQ` function so that the PRQ can determine how to weight peers. For testing purposes, the current strategy is simply one that weights peers based on their ledger `Value`s.

## Testing

Tests are still being written to verify this implementation. To see how it works, you can run the preliminary tests (which only print out info, they do not actually verify anything yet) with `go test -v` in the `exchange/bitswap/decision` directory. As of now, at least one of the `engine_test.go` tests hangs, so `engine_test.go` has been moved to `failing_tests` for testing convenience until the issue is resolved.

## Issues

The way requests are allocated here does not fully reflect the specification in the IPFS whitepaper (draft 3). In particular, since requests are allocated to peers based on their *relative weight to the sum of all peers*, a node may serve poorly-performing peers in the same way as peers who perform well.

For example, say node `i` has two peers, `j` and `k`. If peers `j` and `k` start out with the same weight, they may both stop serving peer `i`'s requests completely (known as 'freeriding') and still receive the same service from `i` (since their relative weights stay the same, and `i` merely provides `j` and `k` with service based on the sum of their weights). This is not an issue if we can assume that a peer `i` will always have a peer `l` who is a long-standing good peer with a large weight (and thus outweighs bad peers `j` and `k`), but without this assumption peer `i` is susceptible to attack by its peers.

One way to address this would be to include a `baselineWeight` in a `Strategy`, which is a value that could be interpreted as the line between a 'good' and a 'bad' peer. Then, the `RRQueue` could somehow use this value to scale the round robin allocations. For example, if `baselineWeight = 2`, `weights['j'] = 1` and `weights['k'] = 1`, we might include peers `j` and `k` in only half of the round robin rounds. However, this would require 'stalling' the round robin round somehow to ensure that peers `j` and `k` have to wait before their requests are fulfilled. We could also simply not serve peers `j` and `k` if their weight goes below some threshold, but this would simply provide a lower bound on the value that determines how much peers `j` and `k` can degrade performance without being penalized.

## Discuss!

All input and questions are welcome, let me know your thoughts on this!


whyrusleeping:
@dgrisham could you rebase this on latest master so that only your changes are here?

dgrisham:
@whyrusleeping yep, just did. `go test -v` in `exchange/bitswap/decision` seems to fail due to an import error in `exchange/bitswap/message/message.go` now, though

whyrusleeping:
@dgrisham make sure the go-libp2p-peer package hash that you imported there is the latest one

dgrisham:
@whyrusleeping the import is in master as well, think it came after I rebased. won't be able to test this until morning (in Europe), but may need to be fixed in master

dgrisham:
I ran `go import <go-libp2p-net_hash>`, which fixed the previous issue, but now I'm getting an error with the `blocks` tests:

```
$ go test -v
# github.com/ipfs/go-ipfs/blocks
./blocks.go:37: cannot use util.Hash(data) (type "github.com/multiformats/go-multihash".Multihash) as type "gx/ipfs/QmVGtdTZdTFaLsaj2RwdVG8jcjNNcp1DE914DKZ2kHmXHw/go-multihash".Multihash in argument to cid.NewCidV0
./blocks.go:59: cannot use b.cid.Hash() (type "gx/ipfs/QmVGtdTZdTFaLsaj2RwdVG8jcjNNcp1DE914DKZ2kHmXHw/go-multihash".Multihash) as type "github.com/ipfs/go-ipfs/vendor/gx/ipfs/QmVGtdTZdTFaLsaj2RwdVG8jcjNNcp1DE914DKZ2kHmXHw/go-multihash".Multihash in return argument
./blocks_test.go:70: cannot use hash (type "github.com/ipfs/go-ipfs/vendor/gx/ipfs/QmVGtdTZdTFaLsaj2RwdVG8jcjNNcp1DE914DKZ2kHmXHw/go-multihash".Multihash) as type "gx/ipfs/QmVGtdTZdTFaLsaj2RwdVG8jcjNNcp1DE914DKZ2kHmXHw/go-multihash".Multihash in argument to cid.NewCidV0
FAIL	github.com/ipfs/go-ipfs/blocks [build failed]
```

Am I doing something weird/not updating dependencies as I should be or something?

whyrusleeping:
@dgrisham the easiest way to fix things is to just run: `gx-go rewrite --fix && gx-go rewrite`

dgrisham:
@whyrusleeping hm, that didn't work unfortunately. ended up re-cloning the go-ipfs and moving my changes into it, seems to be fine now

dgrisham:
## Update -- Alternative Implementation Idea

As mentioned in the OP, there may be an issue with not punishing poorly-performing peers who degrade performance when there are no 'good' peers for them to be compared to. Since we're using the Bitswap ledgers to measure peers, it seems to make more sense to instead think of it as "consider peer `i`'s contribution to me relative to my contribution to peer `i`". So it seems to make sense to *not* compare peers to one another, as the current implementation is doing.

An alternative implementation would be to do something like:

1. Given a `Strategy` weight function, calculate this peer's weight (in the range `[0,1]`).
2. At the beginning of a Round Robin round, for each peer `i` do:
    a. Calculate `rand`,  a random value on a flat distribution within `[0,1]`.
    b. If `rand < weight[i]`, admit peer `i` into the current RR round. Otherwise, don't.

Then we have the question of how to allocate requests to peers within a round. Two options would be:

1. For all peers in the current round, allocate a predefined constant number of requests to those peers.
2. For all `num_peers` peers in the current round, allocate `round_burst / num_peers` requests to each peer.

Both of these options allocate the same number of requests to all peers within a single round. This makes it so that, on a large enough time scale, peers are effectively served fairly (if we assume constant weights, at least...I'm pretty sure that's still true if the weights vary, which of course they will). Option 2, though, allocates a different number of requests on a round-by-round basis (based on `num_peers`, the number of peers in the current round), which I don't think is as 'fair' as Option 1. However, Option 2 has the advantage of setting a hard upper bound on the number of requests that we serve in a single round, which may be desirable for some kind of system guarantee/consistency (the upper bound for Option 1 varies as `requests_per_round_per_peer * num_peers`). I'm more inclined toward Option 1, but I thought I'd mention both for discussion purposes.

dgrisham:
@whyrusleeping Hey! Finally had time to get this rounded off (barring a flag to activate it as an experimental feature), looking for any additional test recommendations. Also pinging @Kubuxu

exchange/bitswap/decision/round_robin_queue.go
```diff
@@ -0,0 +1,114 @@
...
+	if len(rrq.allocations) == 0 {
+		return nil
+	}
+	return rrq.allocations[0]
+}
+
+func (rrq *RRQueue) Shift() {
+	var peer *RRPeer
+	peer, rrq.allocations = rrq.allocations[0], rrq.allocations[1:]
+	rrq.allocations = append(rrq.allocations, peer)
```
magik6k: I think you should just swap references here, reslicing this way may lead to some weird memory/gc behaviour.

exchange/bitswap/decision/strategy_prq.go
```diff
@@ -0,0 +1,225 @@
...
+}
+
+// get first non-trash task
+func (tl *strategy_prq) partnerNextTask(partner *activePartner) *peerRequestTask {
+	for partner.taskQueue.Len() > 0 {
+		task := partner.taskQueue.Pop().(*peerRequestTask)
+		// delete task if it's trash
+		if task.trash {
+			task = nil
+			continue
```
magik6k: Setting the task to nil is not really needed. Inverting the logic would make it a bit simpler:
```go
if !task.trash {
  return task
}
```

exchange/bitswap/decision/strategy_prq.go
```diff
@@ -0,0 +1,225 @@
...
+	taskMap  map[string]*peerRequestTask
+	partners map[peer.ID]*activePartner
+	rrq      *RRQueue
+}
+
+func newStrategyPRQ(strategy Strategy) *strategy_prq {
+	return &strategy_prq{
+		taskMap:  make(map[string]*peerRequestTask),
+		partners: make(map[peer.ID]*activePartner),
+		pQueue:   pq.New(partnerCompare),
```
magik6k: `partnerQueue`?

exchange/bitswap/decision/round_robin_queue.go
dgrisham: Good call :+1:

exchange/bitswap/decision/strategy_prq.go
dgrisham: Yeah, I named it `pQueue` to maintain symmetry with the old PRQ. I can do another PR after and change the name in both.

exchange/bitswap/decision/strategy_prq.go
dgrisham: Hm, good point. I copied this part from the way it was done in the old PRQ, not sure why it was that way.

dgrisham:
@magik6k Hm, maybe. Someone should check me on this, but my understanding is that any issues with a peer having low bandwidth would manifest in packet loss rather than throttling another peer. E.g. if you send more data than peer B can receive within a certain time frame and the line starts to 'clog' around the bottleneck (which is probably peer B's immediate router in most cases) then the router will start to drop packets. But once the data going to peer B is on the wire, you move on and start sending to peer A.

magik6k:
I'm not sure how it's handled internally in libp2p, but TCP (which we are using), will just slow down data upload if the client can't receive it at full speed. On the app side there are 2 things that can happen:
* send()s will be slower - this is not optimal
* data will be buffered in memory - this is really bad (I'm pretty sure that libp2p implements some sort of backoff algorithm, or just uses non-async send(),  so this doesn't happen)

So sending data to peer B would slow down, peer A would exhaust it's allocations way before B, effectively making peer A transfer as slow as peer Bs. Is this correct or am I missing something?

dgrisham:
@magik6k Right, okay, I see what you mean -- haven't had to think about transport protocols in awhile, so I appreciate you talking through this. I'm wondering if the decoupling of the PRQ from the message queues/workers has any importance here. The decision engine will `Pop()` blocks off of the PRQ, the corresponding peer's allocation for the current round-robin round will be reduced accordingly, and the block will be sent to the peer's message queue. Then the (currently 8) message workers handle the messages in the queues.

I think one of the crucial steps in all of that is the peer's allocation is reduced once the block is removed by `Pop()`, not when the message is sent. The round-robin round ends/resets as soon as all of the peers' allocations have been exhausted, so the next round will start regardless of whether all of the data has been sent (since the PRQ is independent from the message workers).

I assume there can still be an issue with a message worker taking a long time sending peer B's message, and too much of this might start to throttle other users, but that seems like it might be an issue on the message worker end and not the PRQ end.

I could very well be missing the mark here, thoughts on all of that?

magik6k:
I don't think having PRQ being independent from message workers is a good thing as we need some sort of feed-back mechanism from the networked part to properly adjust the weights

Let's define couple of scenarios (with what I think may happen, though that's something that should get proper benchmarks):

* **Scenario 1**: (the one discussed above)
  **Peers A, B fetching from C with:**
  * Weights: A - 50; B - 50;
  * Bandwidth: A - 1; B - 2; C - 4

  With these settings currently one of 3 things can happen:
  * Peer B will be throttled to speed of peer A
  * Data will be buffered for both peers. This is not great as we can end-up using large chunk of peer C ram for data that doesn't really need to be there.
  * half of peer A's messages will be dropped, this wastes peer C's resources as it already has read the data and may lead to higher latencies (peer A will have to resend the wantlist more time to actually get the data)

  **What i think should happen**: both A and B should be sent data at max rate they can receive

* **Scenario 2**:
  Peers A, B fetching from C with:
  * Weights: A - 50; B - 50;
  * Bandwidth: A - 1; B - 2; C - 1

  I'm pretty sure that this will work fine, both peer A and B will be sent data at equal rate

* Scenario 3:
  Peers A, B fetching from C with:
  * Weights: A - 33; B - 66;
  * Bandwidth: A - 1; B - 1; C - 1

  This will probably depend heavily on round size, though if the round size is quite big and we decide to buffer the entire round we might get a weird traffic pattern where at the beginning of the round A and B are served at equal rate and when we run out of data for A, we start sending data at full speed to B.

  **What i think should happen**: peer B should be server at 2x the rate or A.

What we want to optimize for is maximum possible bandwidth utilization and fair bandwidth sharing.
To do this in a way that works reasonable well we'd need to feed some sort of bandwidth information back into here. If we did that in scenario 1, weights would be 33 for A and 66 for peer B and the throttling problem would be non-existent(-ish). The problem with this is that bandwidth is not an easy thing to measure and is very dynamic in many real-world applications. It's certainly possible to do that, but is going to be quite hard at this layer.

I'm not sure if it's possible to do that or if any OS supports this - it might be easier to tweak some TCP parameters of underlying connections to tune the speed at which we serve the data to peers to what the bitswap strategy says.

dgrisham:
> I don't think having PRQ being independent from message workers is a good thing as we need some sort of feed-back mechanism from the networked part to properly adjust the weights

Absolutely. One of the longer-term goals of the PRQ is to have the strategy function (which sets peer weights) accept other inputs than just the Bitswap ledger, and values from the message workers would certainly fit into this category. But maybe that's something worth getting into this PR.

> What we want to optimize for is maximum possible bandwidth utilization and fair bandwidth sharing.
>
> To do this in a way that works reasonable well we'd need to feed some sort of bandwidth information back into here. If we did that in scenario 1, weights would be 33 for A and 66 for peer B and the throttling problem would be non-existent(-ish). The problem with this is that bandwidth is not an easy thing to measure and is very dynamic in many real-world applications. It's certainly possible to do that, but is going to be quite hard at this layer.
> 
> I'm not sure if it's possible to do that or if any OS supports this - it might be easier to tweak some TCP parameters of underlying connections to tune the speed at which we serve the data to peers to what the bitswap strategy says.

Agreed with all of this. Ultimately we want to be allocation bandwidth and *not* data as I've done here, this is just a first pass that isn't allocating bandwidth since, as you mentioned, bandwidth is difficult to measure. Maybe @Stebalien has thoughts on your idea of modifying the underlying TCP connections to serve our purposes here?

I'll continue to research approaches to this in the meantime. Might be worthwhile to see how this is handled in Bittorrent clients.

Stebalien:
First, to tackle the specific issue of peer A limiting the bandwidth of peer B.

TCP (and all<sup>&dagger;</sup> of our transports) have backpressure. If you send more than the other side the network can handle, writes to the stream will slow down. Therefore, the correct way to deal with situations like this is *not* to measure bandwidth but to simply rely on the backpressure mechanism.

Unfortunately, I believe the only way to actually fix the issue being discussed is to get rid of workers.

If we had a low-level event loop driven IO system, the worker system would *definitely* be the way to go. In a low-level application, you'd have an event loop that would wake up when a connection is able to receive some number of bytes, write those bytes, and then go to sleep until another connection is ready to receive bytes. However, we can't do that in go (at least, not with our current stream abstractions). Any system that limits itself to N *blocking* workers will have the problems we're discussing here (unless I'm missing something, @whyrusleeping?).

Instead, I believe the best solution is to use resource budgets and rely on backpressure.

That is:

1. When we decide to send a size M block, deduct M+O where O is some set overhead from a memory budget. If we don't have enough memory, wait until we do.
2. Send blocks to peers asynchronously in go routines and, **importantly**, set a timeout (should probably scale with the size of the block we're sending). If we time out, we should kill the stream and set a short backoff on the peer (it means they're very overloaded).
2. Only have one block in flight to any given peer.

Note: spawning go routines can be expensive (likely the reason we have the worker system). However, we can alleviate that by having a large, dynamically sized worker pool (where we can limit the maximum number of workers using a budget):

```go
select {
	case workers<-job:
    default:
    	go worker(workers, job)
}

func worker(workers <-chan job, j job) { /* ... */ }
```

There are a bunch of other optimizations we can make but we can do that down the road.

We should also be using bandwidth/CPU budgets but I'll get into those below.

----

Now for the fun part: what problem are we actually trying to solve? Really, we're trying to:

Optimize: throughput and fairness
Subject to: cpu, memory, and network resource constraints

# Fairness

We don't want to unduly favor any given peer. *How* to do deal with this has been thoroughly discussed here so I'm not really going to go into depth (I also don't really have many strong opinions on this topic).

However, I'd make sure to consider resources *other* than bandwidth. That is, also consider memory*time and, if possible, CPU usage (so that lots of tiny requests aren't favored over a few large ones).

# Resources

We have three constrained resources: network, memory, and CPU. *My* IPFS node shouldn't eat 500MiB and CPU core just because my neighbors want some files from me.

## Memory

As discussed above, the simplest solution is to attack this problem directly by budgeting memory. Whenever handling a request, deduct `block.Size() + expectedOverhead(worker+stream+buffers)`. If we don't have any budget left, wait until enough requests finish. We *could* try to process smaller requests while blocking but, IMO, that's really not worth it as the maximum block size is 2MiB.

## Network

Ideally, we'd be able to say: never use more then N MiB/s bandwidth. However, this can be a bit tricky to do efficiently (at the stream level, at least). We can, but we probably don't have to do that now (or ever, really).

A simpler tactic would be to simply aim for some average bandwidth by limiting the rate at which we queue up blocks to be sent. This is *far* from perfect but I believe it'll work well enough for our purposes. Importantly, it'll actually be quite accurate when under constant heavy load (only being wildly inaccurate under bursty conditions where, IMO, bandwidth constraints are less important).

## CPU

Eventually, we'll probably want to deal with this by putting backpressure on noisy peers. The simplest way to do this would be to sleep occasionally while reading off the bitswap request (wantlist) stream if the peer gets too chatty.

# Throughput

Finally, we don't want slow peers to keep fast peers from having their requests processed. This is actually a well known DoS attack called slow-lorris and is, unfortunately, impossible to solve without either:

1. Payments (make bad actors pay).
2. Reputation (ban bad actors).
3. Lots of memory or, more commonly, cheep reverse proxies (ignore bad actors). This is one of the reasons Cloud Flare exists.

This is because, unfortunately, it's *very* difficult (if not impossible) to distinguish between a bad peer and a slow peer.

IMO, the best we can do is:

1. Set some "too slow limit" (the timeout I talked about above).
2. Set a maximum amount of resources we're willing to give to the network (memory).

And do the best we can. Eventually, we'd like a reputation system that'd allow us to better solve this issue but we don't have that yet :(.

---

<sup>&dagger;</sup> Except for the go-multiplex muxer which will kill a stream if you don't read from it fast enough. However, that will only happen if the receiver is slower than both the sender and the network (unlikely). Also, this is the only transport we have, AFAIK, that suffers from something like this.

PR #3983, opened 2017-06-14
check strong and weak ETag validator

ivan386:
CDN may change strong ETag validator to weak ETag validator.

License: MIT
Signed-off-by: Ivan <ivan386@users.noreply.github.com>


Stebalien:
The code LGTM (although you should try to make the bots a bit happier if possible (I don't think you'll be able to make travis happy)).

whyrusleeping:
looks like its failing a formatting check. Make sure to run gofmt on your code

Stebalien:
In addition to what gitcop says, please squash these commits into a single commit.

PR #3985, opened 2017-06-16
blocks: move block format to it's own repo

Stebalien:
We need to reference it from outside of this repo.


Stebalien:
Let's wait till the repos actually exist. It could take a while.

PR #3990, opened 2017-06-18
blocks: move block format to it's own repo

Stebalien:
We need to reference it from outside of this repo (namely, from go-ipld-format).


Stebalien:
I believe [/ipfs/QmeK8xFAXx4GBpZG6mizck1XAunVtxP8Tmvir5FMwp5MYb](https://ipfs.io/ipfs/QmeK8xFAXx4GBpZG6mizck1XAunVtxP8Tmvir5FMwp5MYb) needs to be pinned for this to build.

Stebalien:
Depends on https://github.com/ipfs/go-ipld-cbor/pull/14 (due to dependency tree update).

kevina:
I don't like little repos with such a small about of code, but maybe its required.  If @whyrusleeping is okay with it so am I.  A few questions:

(1) Did you change anything in `blocks.go` other then move the code into its own repo?

(2) Is the moved code in a repo under `ipfs`.  It likely should be, as a member you should have permission to create a new repo and I am sure @whyrusleeping will be okay with it,

(3) Where did the name `go-block-format` come from.

whyrusleeping:
@kevina yeah, it needs to move outside of go-ipfs so other packages can use it without having to import the entirety of this repo.

Stebalien:
> Did you change anything in blocks.go other then move the code into its own repo?

Nope. I also kept the commit history intact at @whyrusleeping's requrest (although obviously the hashes have changed).

> Is the moved code in a repo under ipfs. It likely should be, as a member you should have permission to create a new repo and I am sure @whyrusleeping will be okay with it,

Yes. https://github.com/ipfs/go-block-format/

> Where did the name go-block-format come from.

I chose it to match go-ipld-format; it serves the same purpose but for blocks. I considered just adding it to go-ipld-format but one should be able to implement block storage and exchange systems without understanding IPLD.

Stebalien:
@lgierth I just noticed that this was tried once before (https://github.com/ipfs/go-blocks) and you moved it *back* into go-ipfs. What was your reasoning?

kevina:
Okay, assuming the tests are fixed I don't see any problems, this looks very straight forward to me.

Stebalien:
I've chatted with lgierth and he has no recollection of what happened here. Furthermore, I see no evidence that this package was ever actually *removed* from this repo. Therefore, I believe that the [ipfs/go-blocks](https://github.com/ipfs/go-blocks) repo just never really got off the ground.

However, I *don't* believe that re-using that repo is the correct solution as this package isn't supposed to have any *real* functionality, just interfaces, abstractions, and helper-functions. We may eventually revive the go-blocks repo (and make it depend on this one) but there's no point in doing that until we actually *need* it.

PR #3993, opened 2017-06-19
Rename PTP to P2P

magik6k:


core/commands/p2p.go
```diff
@@ -30,18 +30,18 @@ type PTPStreamInfoOutput struct {
...
-type PTPStreamsOutput struct {
-	Streams []PTPStreamInfoOutput
+// P2PStreamsOutput is output type of streams command
+type P2PStreamsOutput struct {
+	Streams []P2PStreamInfoOutput
 }
 
-// PTPCmd is the 'ipfs ptp' command
-var PTPCmd = &cmds.Command{
+// P2PCmd is the 'ipfs ppp' command
```
mateon1: Typo, `ppp` instead of p2p

whyrusleeping:
@mateon1 Yeah, we're not entirely sold on the name yet either, but we discussed it in the all hands this morning and decided to go with p2p to avoid confusion about what ptp mean (p2p == peer to peer, ptp == ... point to point?)

Definitely give any feedback you have on this here: https://github.com/ipfs/go-ipfs/issues/3994

PR #3995, opened 2017-06-20
gx import dependencies of dependencies used by go-ipfs

Stebalien:
These were directly used by go-ipfs but not listed in package.json.


Stebalien:
The travis build keeps on timing out at the same place but nothing *should* have changed.

Stebalien:
Oops.

whyrusleeping:
will merge after 0.4.10

Stebalien:
@whyrusleeping the macos test is hanging but that's not related to this PR (this PR doesn't touch code). I think this is ready to merge.

Issue #3989, opened 2017-06-18
mfs support for raw-nodes is incomplete


whyrusleeping:
many commands in the `ipfs files` namespace don't fully support raw-nodes. Simple reproduction:
```
echo "foo" | ipfs add --raw-leaves
ipfs files cp /ipfs/THATHASH /foo
# that bit breaks, but also somewhat succeeds
ipfs files stat --hash /foo
```

kevina:
This sounds like an easy one for me to take on.

whyrusleeping:
@kevina great!

Kubuxu:
It is partially: https://github.com/ipfs/go-ipfs/issues/3921

whyrusleeping:
@Kubuxu yeah, the dagmodifier is a good chunk of it

kevina:
working on this now

PR #3996, opened 2017-06-20
Add better support for Raw Nodes in MFS and elsewhere

kevina:


kevina:
Closes #3989

kevina:
Note this is a work in progress.  In particular the Size() method in DagModifier needs to implemented to support raw nodes with MFS.  I am trying to figure out if it is possible to get the equivalent of Filesize from the generic Node interface.

kevina:
Okay this is no longer a work in progress.  Fully generalize this to allow any type of node to be stored in the MFS tree should probably me another P.R.

Kubuxu:
Could you add test cases from this PR: https://github.com/ipfs/go-ipfs/pull/3901/files

This makes sure that we test as much with RawNodes as we do test with normal ones.

unixfs/mod/dagmodifier.go
```diff
@@ -301,13 +320,26 @@ func (dm *DagModifier) modifyDag(node *mdag.ProtoNode, offset uint64, data io.Re
...
 // appendData appends the blocks from the given chan to the end of this dag
-func (dm *DagModifier) appendData(node *mdag.ProtoNode, spl chunk.Splitter) (node.Node, error) {
+func (dm *DagModifier) appendData(nd node.Node, spl chunk.Splitter) (node.Node, error) {
+
+	var root *mdag.ProtoNode
+	switch nd := nd.(type) {
+	case *mdag.ProtoNode:
+		root = nd
+	case *mdag.RawNode:
+		// TODO: be able to append to rawnodes. Probably requires making this
```
Stebalien: In this case, you're appending a raw node *to* `dm.curNode` which may or may not be raw.

unixfs/mod/dagmodifier.go
```diff
@@ -301,13 +320,26 @@ func (dm *DagModifier) modifyDag(node *mdag.ProtoNode, offset uint64, data io.Re
 }
 
 // appendData appends the blocks from the given chan to the end of this dag
-func (dm *DagModifier) appendData(node *mdag.ProtoNode, spl chunk.Splitter) (node.Node, error) {
+func (dm *DagModifier) appendData(nd node.Node, spl chunk.Splitter) (node.Node, error) {
+
+	var root *mdag.ProtoNode
```
Stebalien: Calling this `root` is misleading (it's the node to-be-appended).

unixfs/mod/dagmodifier.go
```diff
@@ -301,13 +320,26 @@ func (dm *DagModifier) modifyDag(node *mdag.ProtoNode, offset uint64, data io.Re
...
+		return nil, ErrNotUnixfs
+	}
+
 	dbp := &help.DagBuilderParams{
 		Dagserv:  dm.dagserv,
 		Maxlinks: help.DefaultLinksPerBlock,
 	}
 
-	return trickle.TrickleAppend(dm.ctx, node, dbp.New(spl))
+	return trickle.TrickleAppend(dm.ctx, root, dbp.New(spl))
```
Stebalien: The error in TrickleAppend should probably be changed to `ErrNoRawYet` (or `dm.curNode` should be checked here).

kevina:
> Could you add test cases from this PR: https://github.com/ipfs/go-ipfs/pull/3901/files

I thought the test I added where more comprehensive but I didn't see what you will trying to do.  My added tests conflict with your but will make sure mine are just as comprehensive.

unixfs/mod/dagmodifier.go
kevina: This was not my change. @whyrusleeping care to comment

whyrusleeping: Yeah, root is probably not the best name here

unixfs/mod/dagmodifier.go
whyrusleeping: fair point. the comment and error are wrong then, maybe we can actually support this part fine now?

Stebalien: Wait, no; I'm totally wrong. All callers pass `dm.currentNode`. Now, as for why this function even takes a node...

unixfs/mod/dagmodifier.go
kevina: I refactored that method to avoid the problem.

unixfs/mod/dagmodifier.go
kevina: @whyrusleeping I just removed your comment.  If you wish to readd something please give me the correct text to use.

test/sharness/t0250-files-api.sh
```diff
@@ -15,6 +15,7 @@ test_expect_success "can create some files for testing" '
 	FILE1=$(echo foo | ipfs add -q) &&
 	FILE2=$(echo bar | ipfs add -q) &&
 	FILE3=$(echo baz | ipfs add -q) &&
+        FILE9=$(echo zip | ipfs add -q --raw-leaves) &&
```
Kubuxu: Indentation is different here than in the rest of the file

kevina: Yeah sorry about that.  Tab vs space issue.

kevina:
@Kubuxu I redid the tests so they are closer to what you had

Raw nodes are now working in the files API as far as I can tell.

kevina:
@whyrusleeping 

> Code LGTM, i would like to see another test in dagmodifier_test.go that tests dagmodifier.Size() on a raw node though

It was non-obvious how to do this so I instead tested that the size was reported correctly in a sharness test.

whyrusleeping:
@kevina The current dagmodifier tests use: https://github.com/ipfs/go-ipfs/blob/master/unixfs/test/utils.go#L45 which generates a random file of the given size, using a standard importer. You could make a new method there that creates a random file using the raw leaves importer option instead, and run some of the same tests on that node type in dagmodifier_test.go

kevina:
@whyrusleeping most all of the tests in dagmodifier_test has to do with appending data, of which is not supported by raw leaves.  It was non-obvious how to modify the test to support the creation of a node without appending.  I can attempt (with some effort) to write a new set of tests for just raw leaves if you want.  

As an aside, I am not sure of the benefit of supporting the full range of operations on raw leaves.  Appending to a unixfs dir node is useful to add files.  How is appending to a raw node useful?  Is there even an API for appending to a file (I don't see it).  Also when appending, what should we do when it becomes too large.

Kubuxu:
@kevina yes there is an API (using `ipfs files write`), when it is too large it should probably be transformed into unixfs dag.

kevina:
@Kubuxu, oh okay, append would be more complicated though and I think should be pushed to a separate p.r. With this one raw nodes are usable within the files api as long as you don't try to create one within the files API.

whyrusleeping:
@kevina thats fair enough. I think we can live with that restriction for now.

kevina:
@whyrusleeping are you okay with testing the size via the sharness tests only?  The existing tests are designed for when the node is created within the API and I will have to write a new set of go tests to test that the raw node size is reported correctly, of which I am unsure how to do so I will have to spend some time figuring it out.

whyrusleeping:
@kevina yeah, for now i guess testing through the api is good enough.

kevina:
@whyrusleeping then I think this p.r. is ready

whyrusleeping:
@kevina mind rebasing on top of latest master? Should make the tests green again

Issue #1246, opened 2015-05-18
Avoid dialing known _local_ addresses when not in that network


jbenet:
See #1226 

Change `p2p/net` to avoid dialing known local area network addresses when not in that local network. 

For example, it does not make sense to dial a 192.168.0.0/16 address when not within that subnet. This alone will cut out most -- if not all -- of the sysadmin netscan warnings. Most VPSes are in different networks.

We need to:
- [ ] identify a list of all subnets that we should do this with.
- [ ] implement the changes

whyrusleeping:
my thoughts for a simple, "quick and easy" solution are to provide a config setting to allow users to specify which ranges to avoid dialing.

jbenet:
- [ ] implement https://github.com/ipfs/go-ipfs/issues/1246 
- [ ] implement https://github.com/ipfs/go-ipfs/issues/1247

Stebalien:
Moving to libp2p: https://github.com/libp2p/go-libp2p/issues/436

PR #4001, opened 2017-06-21
Implement --profile option in ipfs init

### LINKED ISSUE #3987: add a `--server` flag to ipfs init
It would be good to have flag that sets a bunch of defaults best suited for running ipfs in a server environment. Including turning off mdns, blocking dialing to certain private ip ranges, and (when we have it) setting the announced addresses to the detected public ip of the machine.

magik6k:
Closes #3987

This implements `ipfs init --profile ...` which allows to apply profiles to default config. The server profile does what was requested in #3987.

Profiles can be stacked, this may be very useful when the configurable datastore PR is merged and there are more datastores.

### DIFF
--- a/cmd/ipfs/init.go
+++ b/cmd/ipfs/init.go
@@ -22,36 +22,6 @@ const (
 	nBitsForKeypairDefault = 2048
 )
 
-// TODO: move this out(where?)
-// ConfigProfiles is a map holding configuration transformers
-var ConfigProfiles = map[string]func(*config.Config) error{
-	"server": func(c *config.Config) error {
-
-		// defaultServerFilters has a list of non-routable IPv4 prefixes
-		// according to http://www.iana.org/assignments/iana-ipv4-special-registry/iana-ipv4-special-registry.xhtml
-		defaultServerFilters := []string{
-			"/ip4/10.0.0.0/ipcidr/8",
-			"/ip4/100.64.0.0/ipcidr/10",
-			"/ip4/169.254.0.0/ipcidr/16",
-			"/ip4/172.16.0.0/ipcidr/12",
-			"/ip4/192.0.0.0/ipcidr/24",
-			"/ip4/192.0.0.0/ipcidr/29",
-			"/ip4/192.0.0.8/ipcidr/32",
-			"/ip4/192.0.0.170/ipcidr/32",
-			"/ip4/192.0.0.171/ipcidr/32",
-			"/ip4/192.0.2.0/ipcidr/24",
-			"/ip4/192.168.0.0/ipcidr/16",
-			"/ip4/198.18.0.0/ipcidr/15",
-			"/ip4/198.51.100.0/ipcidr/24",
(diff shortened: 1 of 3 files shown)


cmd/ipfs/init.go
```diff
@@ -21,12 +22,49 @@ const (
 	nBitsForKeypairDefault = 2048
 )
 
+// TODO: move this out(where?)
```
Stebalien: How about `repo/config/profiles.go`?

cmd/ipfs/init.go
```diff
@@ -40,6 +78,7 @@ environment variable:
 	Options: []cmds.Option{
 		cmds.IntOption("bits", "b", "Number of bits to use in the generated RSA private key.").Default(nBitsForKeypairDefault),
 		cmds.BoolOption("empty-repo", "e", "Don't add and pin help files to the local storage.").Default(false),
+		cmds.StringOption("profile", "p", "Apply profile settings to config"),
```
Stebalien: It's non-obvious that you can apply multiple comma-separated profiles.

cmd/ipfs/init.go
```diff
@@ -21,12 +22,49 @@ const (
 	nBitsForKeypairDefault = 2048
 )
 
+// TODO: move this out(where?)
+// ConfigProfiles is a map holding configuration transformers
+var ConfigProfiles = map[string]func(*config.Config) error{
+	"server": func(c *config.Config) error {
```
whyrusleeping: for server, you'll also want to disable mdns

magik6k:
Applied the suggestions

magik6k:
This might be solving #1246 too

cmd/ipfs/init.go
```diff
@@ -132,6 +152,17 @@ func doInit(out io.Writer, repoRoot string, empty bool, nBitsForKeypair int, con
 		}
 	}
 
+	for _, profile := range confProfiles {
+		transformer := config.ConfigProfiles[profile]
```
whyrusleeping:
```go
transformer, ok := ....
if !ok {
   return fmt.Errorf(...
}
```

whyrusleeping:
@magik6k do you think it might be better to use the strings option from #4012 here instead of a comma separated list?

magik6k:
It would be, I actually checked if that existed when I wrote this. When #4012 gets merged I'll change this to use that.

whyrusleeping:
Hrm... we have a weird dependency graph here. This currently feels like its blocked on #4012 which is from an external contributor, so we don't know when it will be completed. Perhaps we should move forward with this as is for now (with the commas) and add support for the 'strings' option later on once that is merged? should be pretty trivial to maintain backwards compatibility there

magik6k:
I agree

whyrusleeping:
@magik6k wanna rebase this in hopes of greener tests?

magik6k:
Rebased. Looks greener too.

PR #4007, opened 2017-06-24
[WIP] Badger datastore

magik6k:
Depends on:
- [x] #3575 (more configurable datastore configs)
- [x] optionally #4001 (--profile option in ipfs init)

TODO:
- [x] Put badger/badger-ds into gx

### DIFF
--- a/repo/config/profile.go
+++ b/repo/config/profile.go
@@ -43,8 +43,12 @@ var ConfigProfiles = map[string]func(*Config) error{
 	},
 	"badgerds": func(c *Config) error {
 		c.Datastore.Spec = map[string]interface{}{
-			"type": "badgerds",
-			"path": "badgerds",
+			"type":   "measure",
+			"prefix": "badger.datastore",
+			"child": map[string]interface{}{
+				"type": "badgerds",
+				"path": "badgerds",
+			},
 		}
 		return nil
 	},
--- a/repo/config/profile.go
+++ b/repo/config/profile.go
@@ -46,8 +46,9 @@ var ConfigProfiles = map[string]func(*Config) error{
 			"type":   "measure",
 			"prefix": "badger.datastore",
 			"child": map[string]interface{}{
-				"type": "badgerds",
-				"path": "badgerds",
+				"type":       "badgerds",
+				"path":       "badgerds",
+				"syncWrites": true,
 			},
 		}
 		return nil
--- a/repo/config/profile.go
+++ b/repo/config/profile.go
@@ -41,4 +41,11 @@ var ConfigProfiles = map[string]func(*Config) error{
 		c.Discovery.MDNS.Enabled = false
 		return nil
 	},
+	"badgerds": func(c *Config) error {
+		c.Datastore.Spec = map[string]interface{}{
+			"type": "badgerds",
+			"path": "badgerds",
+		}
+		return nil
+	},
 }
--- a/repo/fsrepo/datastores.go
+++ b/repo/fsrepo/datastores.go
@@ -355,8 +355,11 @@ func BadgerdsDatastoreConfig(params map[string]interface{}) (DatastoreConfig, er
 	return &c, nil
 }
 
-func (c *badgerdsDatastoreConfig) DiskId() string {
-	return fmt.Sprintf("badgerds;%s", c.path)
+func (c *badgerdsDatastoreConfig) DiskSpec() DiskSpec {
+	return map[string]interface{}{
+		"type": "badgerds",
+		"path": c.path,
+	}
 }
 
 func (c *badgerdsDatastoreConfig) Create(path string) (repo.Datastore, error) {
--- a/repo/fsrepo/datastores.go
+++ b/repo/fsrepo/datastores.go
@@ -343,6 +343,8 @@ type badgerdsDatastoreConfig struct {
 	path string
 }
 
+// BadgerdsDatastoreConfig returns a configuration stub for a badger datastore
+// from the given parameters
 func BadgerdsDatastoreConfig(params map[string]interface{}) (DatastoreConfig, error) {
 	var c badgerdsDatastoreConfig
 	var ok bool
--- a/repo/fsrepo/datastores.go
+++ b/repo/fsrepo/datastores.go
@@ -18,6 +18,7 @@ import (
 
 	badgerds "gx/ipfs/QmNWbaGdPCA3anCcvh4jm3VAahAbmmAsU58sp8Ti4KTJkL/go-ds-badger"
 	levelds "gx/ipfs/QmPdvXuXWAR6gtxxqZw42RtSADMwz4ijVmYHGS542b6cMz/go-ds-leveldb"
+	badger "gx/ipfs/QmQL7yJ4iWQdeAH9WvgJ4XYHS6m5DqL853Cck5SaUb8MAw/badger"
 	ldbopts "gx/ipfs/QmbBhyDKsY4mbY6xsKt3qu9Y7FPvMJ6qbD8AMjYYvPRw1g/goleveldb/leveldb/opt"
 )
 
@@ -340,7 +341,8 @@ func (c measureDatastoreConfig) Create(path string) (repo.Datastore, error) {
 }
 
 type badgerdsDatastoreConfig struct {
-	path string
+	path       string
+	syncWrites bool
 }
 
 // BadgerdsDatastoreConfig returns a configuration stub for a badger datastore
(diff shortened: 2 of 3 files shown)


repo/fsrepo/datastores.go
```diff
@@ -141,6 +145,20 @@ func (r *FSRepo) openLeveldbDatastore(params map[string]interface{}) (repo.Datas
...
+func (r *FSRepo) openBadgerDatastore(params map[string]interface{}) (repo.Datastore, error) {
+	p, ok := params["path"].(string)
+	if !ok {
+		return nil, fmt.Errorf("'path' field is missing or not string")
+	}
+	if !filepath.IsAbs(p) {
+		p = filepath.Join(r.path, p)
+	}
+
+	os.MkdirAll(p, 0755) //TODO: find better way
```
whyrusleeping: i would check errors here

magik6k: This was unpushed for some reason.

Kubuxu:
Please rebase.

whyrusleeping:
some initial numbers, adding all of ipfs dists (~2GB)

badgerds: 19.88 seconds, 9% cpu
flatfs(sync): 109.46 seconds, 2% cpu
flatfs(nosync): 11.91 seconds, 12% cpu

So badger is significantly faster than flatfs, and comparable to flatfs-nosync, while still actually syncing data to disk. This is really great stuff :)

whyrusleeping:
downside is that querying the datastore (`ipfs repo gc` `ipfs refs local` etc) are much slower.

whyrusleeping:
Also, for context:

sha256 hashing all the files from the above experiment took 5.43 seconds
and copying them all to another directory (same disk) took 0.80 seconds

Kubuxu:
It seems for me like badger isn't syncing.

Kubuxu:
Can you try running the sync test and doing manual `sync` latter.

whyrusleeping:
Some more benchmark numbers, adding ipfs dists again:
```
sha256sum everything: 0:05.62
add with badgerds: 0:17.06
add with flatfs: 0:34.58
filestore add w/badger: 0:13.14
filestore add w/flatfs: 0:18.14
add badger+blake2b: 0:11.92
add flatfs+blake2b: 0:33.25
```

whyrusleeping:
This looks like its ready to go. Just probably wanting some code review, cc @magik6k @Stebalien @kevina @Kubuxu 

Also, thank you to @manishrjain and the badger crew for implementing the error handling and pushing more improvements to this. Its looking all around better than our current filesystem based blockstore.

manishrjain:
Glad that it's working for you guys!

I looked at the Badger code usage, you are not setting `SyncWrites=true`. Did you set it during the benchmarks that you run above? (By default, we set it to false for better performance.)

Also, can you ensure that you set GOMAXPROCS=128 -- a value large enough so that if you're running this on SSD, you'll be able to see the max IOPS allowed by the SSD. This is useful for key-value iteration and random Get throughput.

Also, Badger writes work best if you do batch + goroutines. You can also now do `BatchSetAsync`, without the goroutines which would give you good performance (and give back a callback when the write is done).

We're also working on mmap of value log, which would significantly improve the random Get latency.

whyrusleeping:
@manishrjain ah, no. We arent setting `SyncWrites` to true. We should proably do that by default, and then add that to our configuration blob. Why the need for so many threads? That seems odd to me.

Our calling code probably isnt taking advantage of the parallelism of the batch put as much as we should. I can look at tweaking that some and see how it affects the performance.

repo/config/profile.go
```diff
@@ -41,4 +41,11 @@ var ConfigProfiles = map[string]func(*Config) error{
...
 		return nil
 	},
+	"badgerds": func(c *Config) error {
+		c.Datastore.Spec = map[string]interface{}{
+			"type": "badgerds",
+			"path": "badgerds",
+		}
+		return nil
+	},
 }
```
kevina: I may be missing something, but won't this config bypass the measure datastore?  If so will this create a problem?

manishrjain:
> Why the need for so many threads? That seems odd to me.

It's got to do with the fact that disk reads block OS threads, and how many threads can be scheduling disk reads at the same time. Full discussion here:

https://groups.google.com/forum/#!topic/golang-nuts/jPb_h3TvlKE

We set Dgraph by default to 128 threads. It doesn't add that much overhead, so it's a safe change to do.

Consider either using `BatchSetAsync`, or calling `BatchSet` from multiple goroutines. Doing a single Set would always be way slower, because every write call to Badger has to hit the disk. So, the more calls we can batch, the more disk cost can be amortized, and a much higher throughput achieved.

For values below 100 bytes, we have seen 400K key-value writes per second on i3 instances (with local SSD).

magik6k:
We might want to put this in experimental-features doc with some info on how to use this.

repo/config/profile.go
magik6k: IIRC there is global measure in fsrepo

Stebalien:
We can also increase GOMAXPROCS from within go (`runtime.GOMAXPROCS`).

whyrusleeping:
I think this is good to go now. Thoughts? @magik6k @Stebalien @kevina ?

magik6k:
LGTM, though badger-ds could be updated to 0.2.1 (see https://github.com/ipfs/go-ds-badger/commits/master)

kevina:
This LGTM but I have not been following the Badger datastore discussion so I am not really qualified to review this.

repo/fsrepo/datastores.go
```diff
@@ -4,16 +4,21 @@ import (
...
-	levelds "gx/ipfs/QmPdvXuXWAR6gtxxqZw42RtSADMwz4ijVmYHGS542b6cMz/go-ds-leveldb"
 	measure "gx/ipfs/QmSb95iHExSSb47zpmyn5CyY5PZidVWSjyKyDqgYQrnKor/go-ds-measure"
 	flatfs "gx/ipfs/QmUTshC2PP4ZDqkrFfDU4JGJFMWjYnunxPgkQ6ZCA2hGqh/go-ds-flatfs"
+
 	ds "gx/ipfs/QmVSase1JP7cq9QkPT46oNwdp9pT6kBkG3oqS14y3QcZjG/go-datastore"
 	mount "gx/ipfs/QmVSase1JP7cq9QkPT46oNwdp9pT6kBkG3oqS14y3QcZjG/go-datastore/syncmount"
+
+	badgerds "gx/ipfs/QmNWbaGdPCA3anCcvh4jm3VAahAbmmAsU58sp8Ti4KTJkL/go-ds-badger"
+	levelds "gx/ipfs/QmPdvXuXWAR6gtxxqZw42RtSADMwz4ijVmYHGS542b6cMz/go-ds-leveldb"
+	badger "gx/ipfs/QmQL7yJ4iWQdeAH9WvgJ4XYHS6m5DqL853Cck5SaUb8MAw/badger"
```
Stebalien: This isn't declared in package.json. However, if you update to go-ds-badger 0.2.1, you can use the re-exported `badgerds.DefaultOptions` and `badgerds.Options` and avoid this import altogether.

hoffmabc:
Is there going to be a guide to show how to use this?

whyrusleeping:
@Stebalien Could you handle updating the badger-ds dependency?

@hoffmabc Yes, We will add docs around this. But the simple of case of making a new ipfs node that uses this is just `ipfs init --profile=badgerds`. I'm curious though, What sort of structure would you expect docs for this to be in? Its generally pretty easy for us to write "how to use a thing", but its hard to know what users would expect.

Stebalien:
@whyrusleeping done.

hoffmabc:
I'm not sure @whyrusleeping to be honest. That seems pretty straightforward for using this.

hoffmabc:
Do any tests need to be written for this out of curiosity?

whyrusleeping:
We have tests on the data store in isolation, but you're right. It would be
good to have some level of integration testing.

hoffmabc:
Also we initialize ipfs using a config file where leveldb is the only option. Will this be extended to support this datastore?

magik6k:
For tests: we can/should try running whole sharness over this datastore. To not make it take forever to run on travis/circle it could be done as jenkins pipeline stage which would pass additional profiles to ipfs/iptb init in sharness.

magik6k:
Datastore section in `docs/config.md` needs correcting too.

hoffmabc:
is this getting merged today?

whyrusleeping:
@hoffmabc Yeap! :)

Issue #3643, opened 2017-01-29
ipfs add ignore, .ipfsignore


crackcomm:
I would love to be able to run `ipfs add -w -r .` but ignoring `node_modules` or `target` directory, or any other directory.
Something like `.ipfsignore` may be an overkill but something like this is what I am looking for.

whyrusleeping:
Yeah, i'd like to have this too. I implemented it semi-manually for `gx` https://github.com/whyrusleeping/gx/blob/master/gxutil/publish.go

AlbertoElias:
Is there a way to ignore files yet?

Stebalien:
This feature was, unfortunately, blocked on one PR that was blocked on another (large) PR. Path forward:

* Fix: https://github.com/ipfs/go-ipfs-cmdkit/issues/5
* Revive: https://github.com/ipfs/go-ipfs/pull/4013

Unfortunately, I'm not sure anyone will have time to work on this at the moment unless @crackcomm has some free time and feels like reviving their PRs.

kjzz:
Hey @whyrusleeping , Has this issue been finished?If not,i am willing to di this.

marcusnewton1:
+1 good idea

dholms:
Any movement on this? It'd be very useful for us (specifically an `--ignore` or `--exclude` flag).
Happy to contribute as well, but want to check the status first since it seems like PRs addressing this (https://github.com/ipfs/go-ipfs/pull/4013) never got unblocked

Stebalien:
@dholms no progress but not blocked, as far as I know. The next step is https://github.com/ipfs/go-ipfs-cmdkit/issues/5. There are two PRs (see the bottom of that issue) trying to fix that but they're kind of hacky. If you'd like to rebase/clean them up, I can take a look at them.

Note: we have quite a few high-priority issues in flight at the moment so reviews may take a while.

bmann:
@Stebalien we're not using golang in-house at the moment, we can see if we can bounty it.

hsanjuan:
Heads up this issue has been picked up by @corntoole. Any feedback and review help on issues linked above from interested parties is more than welcome!

Stebalien:
Ignore file support has now been merged. You can pass:

* `--ignore=file` to skip a file/directory.
* `--ignore-rules-path=.someignorefile` to apply all ignore rules specified by the file.

[...]

PR #3856, opened 2017-04-06
Extract and rework commands package

keks:
This PR rips out `commands/`, changes a lot of stuff and puts it in [go-ipfs-cmds]. This PR is WIP and primarily here for reviewing.

see #3524 and [go-ipfs-cmds]

Some remarks:

### Backwards compatibility
We don't want to rewrite all of `core/commands` at once, so I built a shimming layer that allows us to use the current go-ipfs/commands.Command structs (with some limitations). All the shimming code is in `go-ipfs-cmds/legacy.go`. That is a lot of code and in the medium term it would be great to dump it. For that we need to use the new commands library in all of core/commands. That is not a short-term project though.

### Shared code
There is quite a lot of shared code between go-ipfs/commands and go-ipfs-cmds. To reduce the number of type conversions, I moved most of it to go-ipfs-cmds/cmdsutil. I don't think this is a very good name though and it might change in the future. `$pkg/${pkg}util` usually contains code that operates on types in `$pkg` and here it's the other way around: cmdsutil contains a lot of basic tools (e.g. an error type) that is used by both go-ipfs-cmds and go-ipfs/commands. Maybe `go-ipfs-cmds/core` is better?

### Basic model

Most changes I made affected the type `Response`, which I split up into `Response` and `ResponseEmitter`. Basically a producer is given a `ResponseEmitter` on which it can call `Emit(v)` a few times (basically it's a channel write) - or just once if `v` is an io.Reader. These values can be received by a consumer which has the corresponding `Response` by calling `v, err := res.Next()`. These can be chained at will - which actually happens when there is a `PostRun`.

### Marshalers, Encoders, PostRun

When I started this project I complained about PostRun. I still think it's a bit weird, but it fulfills an important use case.
In go-ipfs/commands, the `Marshal` function takes the value passed to `SetOutput` and do it's magic to build a bytestream from that. In go-ipfs-cmds there is no singular value, instead we operate on streams. `Encode()` is very similar, but instead it operates on a single emitted value. It is called once per call to `Emit`. This means that no state is kept between emitted values. If you want to e.g. not abort on errors and print an error digest after everything completed, you need to use `PostRun`.
[...]

PR #4012, opened 2017-06-27
Add strings option

crackcomm:
Command strings array option.

It works like that:

```console
$ ipfs -i node_modules -i .git -i yarn.lock -r .
```

It's required to get #3643 done for above to work – ignoring paths.

### DIFF
--- a/commands/cli/parse.go
+++ b/commands/cli/parse.go
@@ -106,15 +106,16 @@ func parseOpts(args []string, root *cmds.Command) (
 	// parseFlag checks that a flag is valid and saves it into opts
 	// Returns true if the optional second argument is used
 	parseFlag := func(name string, arg *string, mustUse bool) (bool, error) {
-		if _, ok := opts[name]; ok {
-			return false, fmt.Errorf("Duplicate values for option '%s'", name)
-		}
-
 		optDef, found := optDefs[name]
 		if !found {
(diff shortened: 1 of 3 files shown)


whyrusleeping:
This feature sounds good to me, makes sense to have and i can definitely see it being useful. I'd like to see some tests around this.

Also, we will have to coordinate adding this feature with @keks refactor work. Shouldnt be a big deal, but definitely need to account for it.

kevina:
@crackcomm the tests failure in ci/circleci look related to your changes

crackcomm:
@kevina passed now

whyrusleeping:
@crackcomm could you add a test or two to commands/cli/parse_test.go for this new flag type?

keks:
The feature and the code LGTM. Since at this point there are zero tests, wouldn't it be better to make this change on top of https://github.com/ipfs/go-ipfs/pull/3856? If we keep changing commands that one will never get merged.

crackcomm:
@keks feel free to do so.

crackcomm:
btw. there are some tests in #4013

keks:
Sure, I'll do that

hsanjuan:
This should be resolved in go-ipfs-cmds if anywhere. So closing this one.

Stebalien:
Fixed a few days ago, actually...

PR #4013, opened 2017-06-27
Add option to ignore files when adding.

### LINKED ISSUE #3643: ipfs add ignore, .ipfsignore

crackcomm:
Current usage:

```
  -i,          --ignore              array  - List of ignore rules. (comma separated).
  -g,          --git-ignore          bool   - Respect .gitignore rules. (experimental).
```

Fixes #3643
Includes #4012 

Breaking changes: `files.NewSerialFile`

### DIFF
--- a/commands/cli/parse.go
+++ b/commands/cli/parse.go
@@ -86,7 +86,29 @@ func ParseArgs(req cmds.Request, inputs []string, stdin *os.File, argDefs []cmds
 			return nil, nil, u.ErrCast()
 		}
 	}
-	return parseArgs(inputs, stdin, argDefs, recursive, hidden, root)
+
+	var rulesFile string
+	if gitignore, _, err := req.Option("git-ignore").Bool(); gitignore {
+		cwd, err := os.Getwd()
+		if err != nil {
+			return nil, nil, err
+		}
+		rulesFile = filepath.Join(cwd, ".gitignore")
+	} else if err != nil {
+		return nil, nil, u.ErrCast()
+	}
+
+	ignore, _, err := req.Option("ignore").Strings()
+	if err != nil {
+		return nil, nil, u.ErrCast()
+	}
+
+	filter, err := files.NewFilter(rulesFile, ignore, hidden)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return parseArgs(inputs, stdin, argDefs, recursive, filter, root)
 }
 
 // Parse a command line made up of sub-commands, short arguments, long arguments and positional arguments
@@ -263,7 +285,7 @@ func parseOpts(args []string, root *cmds.Command) (
 
 const msgStdinInfo = "ipfs: Reading from %s; send Ctrl-d to stop."
 
-func parseArgs(inputs []string, stdin *os.File, argDefs []cmds.Argument, recursive, hidden bool, root *cmds.Command) ([]string, []files.File, error) {
+func parseArgs(inputs []string, stdin *os.File, argDefs []cmds.Argument, recursive bool, filter *files.Filter, root *cmds.Command) ([]string, []files.File, error) {
 	// ignore stdin on Windows
 	if osh.IsWindows() {
 		stdin = nil
@@ -336,7 +358,7 @@ func parseArgs(inputs []string, stdin *os.File, argDefs []cmds.Argument, recursi
 					fpath = stdin.Name()
 					file = files.NewReaderFile("", fpath, r, nil)
 				} else {
-					nf, err := appendFile(fpath, argDef, recursive, hidden)
+					nf, err := appendFile(fpath, argDef, recursive, filter)
 					if err != nil {
 						return nil, nil, err
 					}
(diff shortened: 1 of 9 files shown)


commands/cli/parse.go
```diff
@@ -86,7 +86,29 @@ func ParseArgs(req cmds.Request, inputs []string, stdin *os.File, argDefs []cmds
...
+
+	var rulesFile string
+	if gitignore, _, err := req.Option("git-ignore").Bool(); gitignore {
+		cwd, err := os.Getwd()
+		if err != nil {
+			return nil, nil, err
+		}
+		rulesFile = filepath.Join(cwd, ".gitignore")
+	} else if err != nil {
+		return nil, nil, u.ErrCast()
```
whyrusleeping: i would just return the `err` here

commands/cli/parse.go
```diff
@@ -86,7 +86,29 @@ func ParseArgs(req cmds.Request, inputs []string, stdin *os.File, argDefs []cmds
...
+	}
+
+	ignore, _, err := req.Option("ignore").Strings()
+	if err != nil {
+		return nil, nil, u.ErrCast()
```
whyrusleeping: just `return nil, nil, err`

crackcomm: thanks

commands/cli/parse.go
crackcomm: again, thanks

commands/files/filter.go
```diff
@@ -0,0 +1,36 @@
...
+		filter.Rules, err = ignore.CompileIgnoreLines(rules...)
+	} else {
+		filter.Rules, err = ignore.CompileIgnoreFileAndLines(ignorefile, rules...)
+	}
+	return
+}
+
+// Filter - Returns true if file should be filtered.
+func (filter *Filter) Filter(fpath string) bool {
+	if filter.Hidden && strings.HasPrefix(fpath, ".") {
```
whyrusleeping: I don't know if this would catch hidden files in recursive directories. We might want to use `filepath.Base` here to get just the filename, *then* check if it has the dot prefix.

crackcomm: Input to `Filter` is always a filename actually.

whyrusleeping: Oh? okay, then thats fine

commands/cli/parse.go
crackcomm: should I squash these fixes?

commands/files/filter.go
crackcomm: I discovered it should be the opposite - `Hidden` true means include hidden files.

commands/cli/parse.go
```diff
@@ -86,7 +86,29 @@ func ParseArgs(req cmds.Request, inputs []string, stdin *os.File, argDefs []cmds
 			return nil, nil, u.ErrCast()
 		}
 	}
-	return parseArgs(inputs, stdin, argDefs, recursive, hidden, root)
+
+	var rulesFile string
+	if gitignore, _, err := req.Option("git-ignore").Bool(); gitignore {
```
crackcomm: Here we need to check for `nil`'s anyway.

commands/request.go
```diff
@@ -334,7 +334,7 @@ func (r *request) ConvertOptions() error {
 		}
 
 		kind := reflect.TypeOf(v).Kind()
-		if kind != opt.Type() {
+		if kind != opt.Type() && opt.Type() != Strings && kind != String {
```
crackcomm: Here it should be just `opt.Type() != Strings` otherwise it doesn't convert numbers.

lgierth:
Haven't reviewed the code itself, but I'm wondering why not stick with the convention of a .fooignore file?

crackcomm:
@lgierth I have no idea what is that.

crackcomm:
@lgierth did you mean to _include_ rules from all `.*ignore` files?

lgierth:
Yeah let me clarify this a bit more, @whyrusleeping asked me to, too. Although the solution proposed looks okay on first sight, we haven't looked at any other options. What I'm missing here a bit is the evaluation and weighing of different ways of solving our need for having `ipfs add` ignore files.

Here's another one, off the top of my head, one with an implicit .ipfsignore file.

```
> touch README.md foo
> echo foo > .ipfsignore
> ipfs add -r .
info: skipping files as defined in ./.ipfsignore
added QmA README.md
added QmB .ipfsignore
added QmDir
> ipfs add -f -r .
info: adding otherwise ignored files as defined in ./.ipfsignore
added QmA README.md
added QmB .ipfsignore
added QmC foo
added QmDir2
```
This could be extended to also process an ipfsignore in `$HOME/.ipfsignore`, or generally ipfsignore files in parent directories.

Other options include an explicit ignore file (`--ignorefile=path/to/ipfsignore`), and there's likely good ideas to pick from rsync, git, and other tools that operate on files.

I'm not saying the solution proposed here is bad, only that it's just the first and that we should look at more.

crackcomm:
I personally do not like the idea of additional `.ipfsignore` but I think `--ignorefile` flag is a good idea.

crackcomm:
Let me know if I can do anything to move this forward also #4013.

jacobheun:
Support has been added https://github.com/ipfs/go-ipfs/issues/3643#issuecomment-601801560

PR #4020, opened 2017-06-28
Do not publish public keys extractable from ID

JustinDrake:
License: MIT
Signed-off-by: Justin Drake <drakefjustin@gmail.com>

### DIFF
--- a/namesys/publisher.go
+++ b/namesys/publisher.go
@@ -150,18 +150,24 @@ func PutRecordToRouting(ctx context.Context, k ci.PrivKey, value path.Path, seqn
 		entry.Ttl = proto.Uint64(uint64(ttl.Nanoseconds()))
 	}
 
-	errs := make(chan error, 2)
+	errs := make(chan error, 2) // At most two errors (IPNS, and public key)
+
+	// Attempt to extract the public key from the ID
+	extractedPublicKey := id.ExtractPublicKey()
 
 	go func() {
 		errs <- PublishEntry(ctx, r, ipnskey, entry)
 	}()
 
-	go func() {
-		errs <- PublishPublicKey(ctx, r, namekey, k.GetPublic())
-	}()
+	// Publish the public key if a public key cannot be extracted from the ID
+	if extractedPublicKey == nil {
+		go func() {
+			errs <- PublishPublicKey(ctx, r, namekey, k.GetPublic())
+		}()
 
-	if err := waitOnErrChan(ctx, errs); err != nil {
-		return err
+		if err := waitOnErrChan(ctx, errs); err != nil {
+			return err
+		}
 	}
 
 	return waitOnErrChan(ctx, errs)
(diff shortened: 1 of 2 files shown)


whyrusleeping:
(you'll want to rebase on latest master to make it all work)

JustinDrake:
@whyrusleeping I've rebased.

How do you suggest that the test be written? There are no tests for `publisher.go` that I can build upon. I'm not 100% sure how to test that `PublishPublicKey` is not called. (Its side effects include `log.Debugf` and the networking done in `r.PutValue`, but I don't know how to hook into them.)

whyrusleeping:
@vyzo could you help @JustinDrake out with some tests here?

vyzo:
sure!

vyzo:
@JustinDrake we can add a publisher_test.go file.

We can start with a simple test which directly tests the behaviour of `PutRecordToRouting` wrt to keys. Longer term we certainly want to add some more tests around the publisher code.

The side-effect of `PublishPublicKey` can be checked by looking at the `ValueStore`. You can just use `r.GetValue` -- that should return `ErrNotFound` if the key didn't get published.

vyzo:
Note that you can mock the network and even the value store.
We basically want to ensure that the code is correct and that an RSA key will get published, while an Ed25519 will not.

JustinDrake:
@vyzo Thank you for the suggestions of using the value store, and of testing both the RSA and Ed25519 cases.

@whyrusleeping I now have a test that shows that an Ed25519 key does not publish to the DHT, and (bonus) a test that shows that an RSA key does publish to the DHT.

namesys/publisher_test.go
```diff
@@ -0,0 +1,136 @@
...
+
+	dstore := dssync.MutexWrap(ds.NewMapDatastore())
+	serv := mockrouting.NewServer()
+	r := serv.ClientWithDatastore(context.Background(), &identity{p}, dstore)
+
+	err = PutRecordToRouting(ctx, privKey, value, seqnum, eol, r, id)
+	if err != nil {
+		t.Fatal(err)
+	}
+
```
JustinDrake: addressed 👍

namesys/publisher.go
```diff
@@ -150,18 +150,24 @@ func PutRecordToRouting(ctx context.Context, k ci.PrivKey, value path.Path, seqn
 		entry.Ttl = proto.Uint64(uint64(ttl.Nanoseconds()))
 	}
 
-	errs := make(chan error, 2)
+	errs := make(chan error, 2) // At most two errors (IPNS, and public key)
+
+	// Attempt to extract the public key from the ID
+	var extractedPublicKey = id.ExtractPublicKey()
```
JustinDrake: done 👍

JustinDrake:
Is this good to merge? :)

whyrusleeping:
Thanks @JustinDrake!

PR #4022, opened 2017-06-29
Add CID Codecs to block put

hermanjunge:
Adding codecs found in https://github.com/ipfs/go-cid/blob/b42583c8bfdafac9965f450294b371d1b734c6f6/cid.go#L67-L79

### DIFF
--- a/core/commands/block.go
+++ b/core/commands/block.go
@@ -9,10 +9,10 @@ import (
 
 	util "github.com/ipfs/go-ipfs/blocks/blockstore/util"
 	cmds "github.com/ipfs/go-ipfs/commands"
-	"gx/ipfs/QmXxGS5QsUxpR3iqL5DjmsYPHR1Yz74siRQ4ChJqWFosMh/go-block-format"
 
 	mh "gx/ipfs/QmVGtdTZdTFaLsaj2RwdVG8jcjNNcp1DE914DKZ2kHmXHw/go-multihash"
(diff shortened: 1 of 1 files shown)


hermanjunge:
Force pushed an updated with a more decent commit message 😬

kevina:
@whyrusleeping should we hard code every cid type into go-ipfs?  It seams to me we should use a lookup table from an external library.  Was this discussed somewhere else?

hermanjunge:
Good point @kevina @whyrusleeping, as `mhtype` does have a lookup table (https://github.com/hermanjunge/go-ipfs/blob/c19f1bca895cbab264a6025f5d792616bb0ecb86/core/commands/block.go#L195)

whyrusleeping:
Yeah, we should add a lookup table. Someone want to PR that into go-cid?

hermanjunge:
@whyrusleeping will do. Let's close this meanwhile.

hermanjunge:
@whyrusleeping https://github.com/ipfs/go-cid/pull/26

Once merged, I rebase this PR.

whyrusleeping:
@hsanjuan you should be unblocked now, the go-cid package has been updated in master

hsanjuan:
@whyrusleeping you meant to mention @hermanjunge right ?

whyrusleeping:
Heh, yes i did. I guess my brain saw "oh, theres a j in the middle of that username that starts with an h, thats the right one"

Issue #3921, opened 2017-05-13
Improve DagModifier to fully support CIDv1


Kubuxu:
Currently DagModifier lacks proper support of CIDv1.
We should investigate what is lacking and improve it.

List:
 - raw leaves doesn't work at all: https://github.com/ipfs/go-ipfs/pull/3901
 - switch to using the same CID as the datastructure

kevina:
@Kubuxu @whyrusleeping this sounds like something I can take on if no one else is working on it...

whyrusleeping:
@kevina I *think* @Kubuxu is working on this, I also had a branch pushed that started to address the issue, see https://github.com/ipfs/go-ipfs/pull/3901

Kubuxu:
Yeah, I started working on it but there isn't much done. I just added test to FilesAPI that exposes the problem.

kevina:
@Kubuxu can you push your tests somewhere, I am working on this now in connection with #3989.

whyrusleeping:
@kevina I think its this: https://github.com/ipfs/go-ipfs/pull/3901

kevina:
There are three aspects to the possible aspects to this issue

1. Fully support raw leaves, #4026 makes progress there
2. Support the use of Cidv1 including support for alternative hash functions, I will work on this next.  The easiest way to do this is to allow passing in a custom prefix like is now supported in add
3. Allow non ProtoBuf/Raw nodes to be part of the files API dag.

I am not sure if (3) should is considered part of this issue, but it would seam like something that will be useful to have.  I guess support for that depends on how well ProtoBuf directory entries interact with links to non-Protobuf node.  In particular the Size entry in Link is support to be for "FileSize" which may not always have a meaningful value.

whyrusleeping:
@kevina sounds right to me. number 3 should probably just not freak out about having non-unixfs nodes in the graph. read and write shouldnt be supported on those nodes, and probably also don't want to support traversing them as directories in that context.

kevina:
Here is my planned strategy for support CidV1 and alternative hash functions.

1) When a file or directory is modified the new hash and any new interior nodes in the dag will use the same prefix (which will preserve the Cid version and hash function).
2) When a new file or directory is created it will use the same prefix as the directory it is added to, by default.  The `--cid-version` and `--hash` option can be used to specify a different prefix instead.
[...]

PR #4026, opened 2017-07-01
Add full support for CidV1 in Files API and Dag Modifier

### LINKED ISSUE #3989: mfs support for raw-nodes is incomplete

kevina:
Closes  #3989 #3921

Status: https://github.com/ipfs/go-ipfs/pull/4026#issuecomment-322058542

### DIFF
--- a/core/commands/files/files.go
+++ b/core/commands/files/files.go
@@ -41,21 +41,10 @@ of consistency guarantees. If the daemon is unexpectedly killed before running
 'ipfs files flush' on the files in question, then data may be lost. This also
 applies to running 'ipfs repo gc' concurrently with '--flush=false'
 operations.
-
-The --cid-version and --hash-fun option only apply to newly created files
-and directories.  If not specified these proprieties are inhertied
-from the parent directory.
 `,
 	},
 	Options: []cmds.Option{
 		cmds.BoolOption("f", "flush", "Flush target and ancestors after write.").Default(true),
-		cmds.BoolOption("raw-leaves", "Use raw blocks for newly created leaf nodes. (experimental)"),
-		cmds.IntOption("cid-version", "cid-ver", "Cid version. Non-zero value will change default of 'raw-leaves' to true. (experimental)"),
-		cmds.StringOption("hash-fun", "Hash function to use. Will set Cid version to 1 if used. (experimental)"),
-		// ^^fixme: can't use just "hash" as the option name as the
-		// conflicts with "--hash" usage by the stat command, this is
-		// unfortunate as it creates an inconsistency with the "add"
-		// that uses "hash"
 	},
 	Subcommands: map[string]*cmds.Command{
 		"read":  FilesReadCmd,
@@ -70,6 +59,9 @@ from the parent directory.
 		"chcid": FilesChcidCmd,
 	},
 }
+
+var cidVersionOption = cmds.IntOption("cid-version", "cid-ver", "Cid version to use. (experimental)")
+var hashOption = cmds.StringOption("hash", "Hash function to use. Will set Cid version to 1 if used. (experimental)")
 
 var formatError = errors.New("Format was set by multiple options. Only one format option is allowed")
 
@@ -576,6 +568,13 @@ a beginning offset to write to. The entire length of the input will be written.
 If the '--create' option is specified, the file will be created if it does not
 exist. Nonexistant intermediate directories will not be created.
 
+Newly created files will have the same CID version and hash function of the
+parent directory unless the --cid-version and --hash options are used.
+
+Newly created leaves will be in the legacy format (Protobuf) if the
+CID version is 0, or raw is the CID version is non-zero.  Use of the
+--raw-leaves option will override this behavior.
+
 If the '--flush' option is set to false, changes will not be propogated to the
 merkledag root. This can make operations much faster when doing a large number
 of writes to a deeper directory structure.
@@ -601,6 +600,9 @@ stat' on the file or any of its ancestors.
 		cmds.BoolOption("create", "e", "Create the file if it does not exist."),
 		cmds.BoolOption("truncate", "t", "Truncate the file to size zero before writing."),
 		cmds.IntOption("count", "n", "Maximum number of bytes to read."),
+		cmds.BoolOption("raw-leaves", "Use raw blocks for newly created leaf nodes. (experimental)"),
+		cidVersionOption,
+		hashOption,
 	},
 	Run: func(req cmds.Request, res cmds.Response) {
 		path, err := checkPath(req.Arguments()[0])
@@ -709,6 +711,9 @@ var FilesMkdirCmd = &cmds.Command{
 		ShortDescription: `
 Create the directory if it does not already exist.
 
+The directory will have the same CID version and hash function of the
+parent directory unless the --cid-version and --hash options are used.
+
 NOTE: All paths must be absolute.
 
 Examples:
@@ -723,6 +728,8 @@ Examples:
 	},
 	Options: []cmds.Option{
 		cmds.BoolOption("parents", "p", "No error if existing, make parent directories as needed."),
+		cidVersionOption,
+		hashOption,
 	},
 	Run: func(req cmds.Request, res cmds.Response) {
 		n, err := req.InvocContext().GetNode()
@@ -800,6 +807,10 @@ Change the cid version or hash function of the root node of a given path.
 	},
 	Arguments: []cmds.Argument{
 		cmds.StringArg("path", false, false, "Path to change. Default: '/'."),
+	},
+	Options: []cmds.Option{
+		cidVersionOption,
+		hashOption,
 	},
 	Run: func(req cmds.Request, res cmds.Response) {
 		nd, err := req.InvocContext().GetNode()
@@ -959,7 +970,7 @@ Remove files or directories.
 
 func getPrefix(req cmds.Request) (*cid.Prefix, error) {
 	cidVer, cidVerSet, _ := req.Option("cid-version").Int()
-	hashFunStr, hashFunSet, _ := req.Option("hash-fun").String()
+	hashFunStr, hashFunSet, _ := req.Option("hash").String()
 
 	if !cidVerSet && !hashFunSet {
 		return nil, nil
--- a/core/commands/files/files.go
+++ b/core/commands/files/files.go
@@ -43,6 +43,7 @@ operations.
 	},
 	Options: []cmds.Option{
 		cmds.BoolOption("f", "flush", "Flush target and ancestors after write.").Default(true),
+		cmds.BoolOption("raw-leaves", "Use raw blocks for newly created leaf nodes. (experimental)"),
 	},
 	Subcommands: map[string]*cmds.Command{
 		"read":  FilesReadCmd,
@@ -598,6 +599,7 @@ stat' on the file or any of its ancestors.
 		create, _, _ := req.Option("create").Bool()
 		trunc, _, _ := req.Option("truncate").Bool()
 		flush, _, _ := req.Option("flush").Bool()
+		rawLeaves, _, _ := req.Option("raw-leaves").Bool()
 
 		nd, err := req.InvocContext().GetNode()
 		if err != nil {
@@ -620,6 +622,7 @@ stat' on the file or any of its ancestors.
 			res.SetError(err, cmds.ErrNormal)
 			return
 		}
+		fi.RawLeaves = rawLeaves
 
 		wfd, err := fi.Open(mfs.OpenWriteOnly, flush)
 		if err != nil {
--- a/core/commands/files/files.go
+++ b/core/commands/files/files.go
@@ -58,15 +58,16 @@ from the parent directory.
 		// that uses "hash"
 	},
 	Subcommands: map[string]*cmds.Command{
-		"read":  FilesReadCmd,
-		"write": FilesWriteCmd,
-		"mv":    FilesMvCmd,
-		"cp":    FilesCpCmd,
-		"ls":    FilesLsCmd,
-		"mkdir": FilesMkdirCmd,
-		"stat":  FilesStatCmd,
-		"rm":    FilesRmCmd,
-		"flush": FilesFlushCmd,
+		"read":   FilesReadCmd,
+		"write":  FilesWriteCmd,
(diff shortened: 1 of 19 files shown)


kevina:
Tests are failing.  Working on fixing.

Also plan to add a "--raw-leaves" option to the files api.

kevina:
Support for raw leaves is done.

I am going to tack on full CifV1 support and also support for setting a alternative hash function since they are very closely related.  If necessary I can separate out CidV1 support into a new commit.

kevina:
Full support for CidV1 (including using an alternative hash function is now done) using the following strategy (slightly modified from https://github.com/ipfs/go-ipfs/issues/3921#issuecomment-321193691)

1) When a file or directory is modified the new hash and any new interior nodes in the dag will use the same prefix (which will preserve the Cid version and hash function).
2) When a new file or directory is created it will use the same prefix as the directory it is added to, by default.  The `--cid-version` and `--hash-fun` option can be used to specify a different prefix instead.
3) If the Cid version is 0 legacy leaf nodes will be used, otherwise raw leaves will be used.  This behavior can be changed by using the `--raw-leaves` option when modifying a file.
4) In order to be able to change the MFS root I added a `update` command to the files api.  When called it will update the prefix (cid version and hash function) of a directory.

If any part of this is controversial or needs more though it can easily be factored out of this P.R. and into a new one.

kevina:
@whyrusleeping this has been ready for review for a while now

The only outstanding issue if the `files update` command (ref https://github.com/ipfs/go-ipfs/issues/4171) which is needed to update the root to something other than CIDv0.  This part doesn't have to go in, but many of the tests currently rely on it and will have to be redone.

I wrote a new utility `cid-fmt` (see https://github.com/ipfs/go-cid/pull/31) which can greatly help with these tests.

core/commands/files/files.go
```diff
@@ -716,7 +740,21 @@ Examples:
...
+		root := n.FilesRoot
+		if prefix != nil {
+			// FIXME: This is ugly and may not be correct either
+			// -- kevina
+			newRoot := *root
+			root = &newRoot
+			root.Prefix = prefix
+		}
+
+		err = mfs.Mkdir(root, dirtomake, dashp, flush)
```
whyrusleeping: Hrm.. this is getting complicated. Its starting to feel like we need some sort of `MkdirOpts` here that hold the dashp, flush, and prefix options for the operation.

core/commands/files/files.go
```diff
@@ -756,6 +794,69 @@ are run with the '--flush=false'.
 	},
 }
 
+var FilesUpdateCmd = &cmds.Command{
```
whyrusleeping: not yet sold on `update` as the command name, but its an okay placeholder for now.

core/commands/files/files.go
```diff
@@ -756,6 +794,69 @@ are run with the '--flush=false'.
...
+	Helptext: cmds.HelpText{
+		Tagline: "Change the cid version of hash function of the root node of a given path.",
```
whyrusleeping: s/of/or/

Also, given this description, maybe we should call the command `re-cid` ?

core/commands/files/files.go
kevina: That would work.  Are you thinking:
```
mfs.Mkdir(root, dirtomake, mfs.MkdirOpts{Dashp: dashp, Flush: flush, Prefix: prefix})
```

core/commands/files/files.go
kevina: How about `chcid`?

core/commands/files/files.go
```diff
@@ -756,6 +794,69 @@ are run with the '--flush=false'.
...
+		ShortDescription: `
+Flush a given path to disk. This is only useful when other commands
```
whyrusleeping: ShortDescription needs fixing

core/commands/files/files.go
```diff
@@ -756,6 +794,69 @@ are run with the '--flush=false'.
...
+are run with the '--flush=false'.
+`,
+	},
+	Arguments: []cmds.Argument{
+		cmds.StringArg("path", false, false, "Path to flush. Default: '/'."),
```
whyrusleeping: update text here

core/commands/files/files.go
```diff
@@ -860,8 +961,36 @@ Remove files or directories.
 	},
 }
 
-func getFileHandle(r *mfs.Root, path string, create bool) (*mfs.File, error) {
+func getPrefix(req cmds.Request) (*cid.Prefix, error) {
+	cidVer, cidVerSet, _ := req.Option("cid-version").Int()
+	hashFunStr, hashFunSet, _ := req.Option("hash-fun").String()
```
whyrusleeping: i'm not sure we should use `hash-fun`. ipfs add just uses `--hash`, i'm thinking we should be consistent.

kevina: @whyrusleeping That won't work.  The hash option is already used in one of the sub-commands and will create a conflict.

core/commands/files/files.go
```diff
@@ -39,21 +41,33 @@ of consistency guarantees. If the daemon is unexpectedly killed before running
...
 	Options: []cmds.Option{
 		cmds.BoolOption("f", "flush", "Flush target and ancestors after write.").Default(true),
+		cmds.BoolOption("raw-leaves", "Use raw blocks for newly created leaf nodes. (experimental)"),
+		cmds.IntOption("cid-version", "cid-ver", "Cid version. Non-zero value will change default of 'raw-leaves' to true. (experimental)"),
+		cmds.StringOption("hash-fun", "Hash function to use. Will set Cid version to 1 if used. (experimental)"),
+		// ^^fixme: can't use just "hash" as the option name as the
+		// conflicts with "--hash" usage by the stat command, this is
+		// unfortunate as it creates an inconsistency with the "add"
+		// that uses "hash"
 	},
```
kevina: @whyrusleeping here is the comment on why I can't use "hash" as the option name

core/commands/files/files.go
kevina: Oops, will fix.

core/commands/files/files.go
kevina: Oops, will fix.

kevina:
@whyrusleeping I pushed some commits to address your feedback, I changed `update` to `chcid`.

core/commands/files/files.go
whyrusleeping: Can't we just have the `--hash` option for stat be different from the one on the other commands?

kevina: We could, if you are willing to change the interface for stat.  That is something I didn't want to do.

kevina: @whyrusleeping another option is to use `--mhash` instead of `--hash-fun` this is more specific (multi-hash) and is a good name for eventually making it a global option.

whyrusleeping: I still don't think this is blocking anything. Stat is a separate command, it can have a flag that means something different, having those options on the `ipfs files` command doesnt make sense for a lot of cases too. `ipfs add`, which will at some point be moved under the `ipfs files` command also uses `--hash`. We need to stay consistent here.

core/coreunix/add.go
```diff
@@ -390,7 +389,12 @@ func (adder *Adder) addNode(node node.Node, path string) error {
 	}
 	dir := gopath.Dir(path)
 	if dir != "." {
-		if err := mfs.Mkdir(mr, dir, true, false); err != nil {
+		opts := mfs.MkdirOpts{
+			Mkparents: true,
+			Flush:     false,
+			Prefix:    adder.Prefix,
+		}
+		if err := mfs.Mkdir(mr, dir, opts); err != nil {
```
whyrusleeping: I think the argument to mkdir should be a pointer so we can pass `nil` and have that assume default values.

importer/trickle/trickle_test.go
```diff
@@ -19,10 +19,23 @@ import (
 	u "gx/ipfs/QmWbjfz3u6HkAdPh34dgPchGbQjob6LXLhAeCGii2TX69n/go-ipfs-util"
 )
 
-func buildTestDag(ds merkledag.DAGService, spl chunk.Splitter) (*merkledag.ProtoNode, error) {
+type UseRawLeaves bool
+
+const (
+	ProtoBufLeaves UseRawLeaves = false
```
whyrusleeping: Declaring a type for this seems like overkill.

test/sharness/t0250-files-api.sh
```diff
@@ -59,144 +63,150 @@ test_sharding() {
 		done
 	'
 
-	test_expect_success "listing works" '
+	test_expect_success "listing works $EXTRA" '
```
whyrusleeping: might be good to wrap parentheses around this, nicer output

test/sharness/t0250-files-api.sh
```diff
@@ -205,362 +215,435 @@ test_files_api() {
...
 
-test_expect_success "can create some files for testing with raw-leaves" '
-	create_files --raw-leaves
-'
-test_files_api QmTpKiKcAj4sbeesN6vrs5w3QeVmd4QmGpxRL81hHut4dZ
+tests_for_files_api() {
+	local EXTRA
+	EXTRA=$1
+
+	test_expect_success "can create some files for testing ($extra)" '
```
whyrusleeping: why is this 'extra' lowercased?

test/sharness/t0250-files-api.sh
whyrusleeping: Ah, i see. you do that before passing it into here

unixfs/mod/dagmodifier.go
```diff
@@ -238,43 +232,87 @@ func (dm *DagModifier) Sync() error {
...
+			f, err := ft.FromBytes(nd0.Data())
+			if err != nil {
+				return nil, false, err
+			}
 
-	f, err := ft.FromBytes(node.Data())
-	if err != nil {
-		return nil, false, err
-	}
+			n, err := data.Read(f.Data[offset:])
```
whyrusleeping: might want to use io.ReadFull here.

unixfs/mod/dagmodifier.go
```diff
@@ -238,43 +232,87 @@ func (dm *DagModifier) Sync() error {
...
+			}
 
-		// Hey look! we're done!
-		var done bool
-		if n < len(f.Data[offset:]) {
-			done = true
+			return k, done, nil
+		case *mdag.RawNode:
+			origData := nd0.RawData()
+			bytes := make([]byte, len(origData))
```
whyrusleeping: I think its probably okay to *not* make a copy of the data here, but i guess we can optimize this later (and as is, is safe)

test/sharness/t0250-files-api.sh
```diff
@@ -205,362 +215,435 @@ test_files_api() {
...
+		create_files
+	'
+	ROOT_HASH=QmcwKfTMCT7AaeiD92hWjnZn9b6eh9NxnhfSzN5x2vnDpt
+        CATS_HASH=Qma88m8ErTGkZHbBWGqy1C7VmEmX8wwNDWNpGyCaNmEgwC
```
Kubuxu: spaces

importer/trickle/trickle_test.go
kevina: It makes things easier to read.

test/sharness/t0250-files-api.sh
kevina: Mistake.

core/coreunix/add.go
kevina: I'm not sure what the defaults should be in that case.

core/commands/files/files.go
kevina: > Stat is a separate command, it can have a flag that means something different

Not with the current API.  Currently the `--hash-fun` function is a global option for all `ipfs files` command.  The `ipfs files stat` command` also has the `--hash` option.  It creates this error when the `stat` command is used:
```
$ ipfs files stat
Error: Option name 'hash' used multiple times
```
To resolve this conflict I need to either move the `--hash` option into each of the subcommands, or rename the `--stat` option in `ipfs files stat` to something else.  If I move the `--hash` option to each of the subcommands I should likely move the `--raw-leaves` and `--cid-ver` option also to be consistent.

importer/trickle/trickle_test.go
whyrusleeping: Okay, yeah. I'm fine with it (after reading through the tests)

core/commands/files/files.go
whyrusleeping: > Currently the --hash-fun function is a global option for all ipfs files command. 

dont do that.

> I need to either move the --hash option into each of the subcommands

Yes, do that.

kevina: Done.

unixfs/mod/dagmodifier.go
kevina: This is not new code so I rather leave it alone.  It is expected that the data from the reader might not completely full the buffer.  If I use ReadFull I will then need to test for both io.EOF and io.ErrUnexpectedEOF, otherwise tests will fail.

whyrusleeping: Ah, i see. fine to not change existing code

test/sharness/t0250-files-api.sh
```diff
@@ -97,16 +97,17 @@ test_sharding() {
 }
 
 test_files_api() {
-	local EXTRA ARGS
+	local EXTRA ARGS WRITE_ARGS
```
whyrusleeping: did you mean to call this `RAW_LEAVES` ?

whyrusleeping:
Alright, this looks good to me. Going to need to rebase, and then it would be nice to have a few more of the codeclimate issues resolved before merging.

kevina:
@whyrusleeping rebased and cleaned up some of the codeclimate issues

whyrusleeping:
Thanks @kevina! we're entering the brave new world of cidV1..

kevina:
Your Welcome.  I image there is still more work to do, but this should hopefully make CidV1 practical to use by default.

PR #4032, opened 2017-07-04
extract dag-put logic to subpackage

whyrusleeping:
This is a bit of a WIP, the goal here is to simplify the process of making `ipfs dag put` smarter when new types are added.

License: MIT


Kubuxu:
As far as I understand this can be replaced by: https://github.com/ipfs/go-ipfs/pull/4033

PR #4033, opened 2017-07-04
 plugin: create plugin API and loader, add ipld-git plugin

whyrusleeping:
This is very much a work in progress, but it works and is pretty fun to play around with.

If you want to try it, pull this branch down and:

```
make install
mkdir $IPFS_PATH/plugins
go build -buildmode=plugin -o="$IPFS_PATH/plugins/git.so" ./plugins/git
```

Then start ipfs, it will now be loaded with git support. You can install https://github.com/magik6k/git-remote-ipld and then try out `git push ipld::` to push a git repo into ipfs and then `git clone ipld::GITSHA` to clone it from ipfs.


whyrusleeping:
cc @kumavis and @hermanjunge, this will be how we load in blockchain support too

VictorBjelkholm:
Really cool to have plugin support finally :) However, the following disclaimer exists in the [documentation](https://golang.org/pkg/plugin/): ` Currently plugins only work on Linux. ` but I'm unsure if this is for building with plugins, building the plugins or including the plugins...

whyrusleeping:
@VictorBjelkholm its for all of the above. The plugin stuff only works on linux. If people want this on platforms other than linux, pester the go team.

cmd/ipfs/plugin_linux.go
```diff
@@ -0,0 +1,80 @@
...
+	"plugin"
+
+	node "gx/ipfs/QmUBtPvHKFAX43XMsyxsYpMi3U5VwZ4jYFTo4kFhvAR33G/go-ipld-format"
+)
+
+func init() {
+	LoadPlugins = LoadPluginsLinux
+}
+
+func LoadPluginsLinux(cfgroot string) error {
```
Kubuxu: this can be private.

plugins/git/git.go
```diff
@@ -0,0 +1,39 @@
+package main
+
+import (
+	"compress/zlib"
+	"io"
+
+	"github.com/ipfs/go-cid"
+	"github.com/ipfs/go-ipfs/core/coredag"
+	git "github.com/ipfs/go-ipld-git"
+	"gx/ipfs/QmUBtPvHKFAX43XMsyxsYpMi3U5VwZ4jYFTo4kFhvAR33G/go-ipld-format"
```
Kubuxu: Spacing and go-ipld-git is not imported.

core/coredag/dagtransl.go
```diff
@@ -0,0 +1,81 @@
...
+
+type DagParser func(r io.Reader) ([]node.Node, error)
+
+type FormatParsers map[string]DagParser
+type InputEncParsers map[string]FormatParsers
+
+var DefaultInputEncParsers = InputEncParsers{
+	"json": DefaultJsonParsers,
+	"raw":  DefaultRawParsers,
+}
```
Kubuxu: I am not sure if making this global is the right decision.
This will make testing it harder.

cmd/ipfs/plugin_linux.go
```diff
@@ -0,0 +1,80 @@
...
+	if !ok {
+		return fmt.Errorf("'PluginType' var was not a string (got %T)", typestri)
+	}
+
+	switch *typestr {
+	case "ipld":
+		regfunci, err := plugin.Lookup("Register")
+		if err != nil {
+			return fmt.Errorf("ipld plugin had no Register function: %s", err)
+		}
```
Kubuxu: Regarding API for plugins themselves, I think it might be better to have a plugin expose a structure of callback functions, implemented interface or overridden structure. This way the API is cleanly defined in Go code instead of having to be written somewhere separately.

Kubuxu:
Once we decide on structure more I will add Makefile support for it.

core/coredag/dagtransl.go
whyrusleeping: If we dont use a global, registration becomes a much more complicated thing to do. Any ideas on a good way to do that?

cmd/ipfs/plugin_linux.go
whyrusleeping: Hrm... so we could maybe do something like:
```go
plug, _ := plugin.Lookup("Plugin")
switch plug := plug.(type) {
case IpldPlugin:
case Libp2pPlugin:
case DatastorePlugin:
default:
   // unsupported plugin type!
}
```

Kubuxu: That would work.

core/coredag/dagtransl.go
Stebalien: Can't we do the same thing we're doing with `BlockDecoder`s (pass it to register)?

Kubuxu: Yeah, I am planning that.

Kubuxu: BlockDecoder is doing the same (it is using DefaultBlockDecoder from go-ipld-format).

Stebalien: Good point. The problem isn't really registration, it's *using* the the decoders (e.g. in `ParseInputs`).

Kubuxu: In the Plugin API for IPLD I've sketched out I've made the registration map an argument so at least the plugins can be tested.

We have to figure out how to use the decoders well with non-global states.

Stebalien:
This may be beyond the scope of this PR but it would be *really* nice to make encoders/decoders composable and separate. That is, when we re-encode, always decode to a common (in-memory) format/interface and then re-encode to the target format. Unfortunately, the only way I think of doing this in go is to decode to a dynamically typed `interface{}` of maps, arrays, integers, strings, etc... So, how about something like https://github.com/ipfs/go-ipld-format/issues/18.

Kubuxu:
This should be ready for first CR, I will try adding some tests and more UX regarding plugins on other systems (display warnings if there is a plugin in directory but system is not supported).

Kubuxu:
Hah, I can't request review from @whyrusleeping's own PR.

whyrusleeping:
would be pretty cool if github had a 'take over PR' button

plugin/loader/load_linux.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+	pls, err := pl.Lookup("Plugins")
+	if err != nil {
+		return nil, err
+	}
+
+	typePls, ok := pls.([]iplugin.Plugin)
+	if !ok {
+		return nil, errors.New("filed 'Plugins' didn't contain correct type")
+	} else {
```
whyrusleeping: no need for the else

whyrusleeping:
would be pretty cool if github had a 'take over PR' button.

You also probably don't want to commit the plugin `git.so` binary.

Kubuxu:
Oops.

Kubuxu:
Whoa, tests are green. Git IPLD plugin is being loaded dynamically and works (passes sharness tests written by @magik6k).

Should be ready for CR: @whyrusleeping @Stebalien 

This PR contains quite a bit of Makefile magic to glue plugins and tests together, provide option for preloading plugins (`plugin/loader/preload_list`) and allow for dynamic loading of plugins and preloading them (to load plugin dynamically it has to be defined in a `main` package, to preload it can't).

Kubuxu:
I just saw that coverage collection failed, so please do not merge I have to fix it first.

plugin/loader/load_linux.go
```diff
@@ -0,0 +1,67 @@
...
+	"fmt"
+	"os"
+	"path/filepath"
+	"plugin"
+
+	iplugin "github.com/ipfs/go-ipfs/plugin"
+)
+
+func init() {
+	loadPluginsFunc = linxuLoadFunc
```
Stebalien: s/linxu/linux/g

plugin/loader/initializer.go
```diff
@@ -0,0 +1,45 @@
+package loader
+
+import (
+	"github.com/ipfs/go-ipfs/core/coredag"
+	"github.com/ipfs/go-ipfs/plugin"
+
+	format "gx/ipfs/QmYNyRZJBUYPNrLszFmrBrPJbsBh2vMsefz5gnDpB5M1P6/go-ipld-format"
+)
+
+func initalize(plugins []plugin.Plugin) error {
```
Stebalien: s/initalize/initialize/g

Kubuxu:
I don't think it is possible for me to build plugins with coverage collection. This means I will either have to preload plugins for coverage collection or disable plugins during coverage collection.

I think for now I will disable plugins and plugin tests during coverage collection.

Kubuxu:
Coverage of the patch is low as the sharness won't work well with the plugins.
In future I might work on running go-ipfs with plugins preloaded for testing.

plugin/loader/initializer.go
```diff
@@ -0,0 +1,45 @@
...
+	for _, pl := range plugins {
+		err := runIPLDPlugin(pl)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func runIPLDPlugin(pl plugin.Plugin) error {
```
whyrusleeping: do a type switch up in the 'run' function, and have this function accept the proper type. Then we can error out if the types don't match correctly instead of silently ignoring the problem (which could easily happen with a mis-compiled plugin)

plugin/loader/initializer.go
```diff
@@ -0,0 +1,45 @@
...
+	ipldpl, ok := pl.(plugin.PluginIPLD)
+	if !ok {
+		return nil
+	}
+
+	var err error
+	err = ipldpl.RegisterBlockDecoders(format.DefaultBlockDecoder)
```
whyrusleeping: why not just `err := ....` ?

plugin/loader/initializer.go
```diff
@@ -0,0 +1,45 @@
...
+	if err != nil {
+		return err
+	}
+
+	err = ipldpl.RegisterInputEncParsers(coredag.DefaultInputEncParsers)
```
whyrusleeping: `return ipldpl.Register....`

plugin/plugins/.gitignore
```diff
@@ -0,0 +1,2 @@
+*.so
+*/main
```
whyrusleeping: what is this main ignore for?

test/sharness/t0280-plugin-git.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/sh
+#
+# Copyright (c) 2016 Jakub Sztandera
```
whyrusleeping: its 2017 btw

whyrusleeping:
This LGTM, great work @Kubuxu  :)

I'll wait for your confirmation before merging, but I think this is ready to go.

plugin/plugins/.gitignore
Kubuxu: This is autogenerated package in directory of every plugin. This is what I have to do so the plugin can be either pre-loaded (statically compiled with go-ipfs, for non-linux builds) or be compiled for dynamic linking as plugin.

Go requires plugins to be defined in package named `main`, and at the same time it doesn't allow for static linking of `main` package. So I have main plugin inside its own package and I generate `main` package that rebinds `Plugins` variable.

plugin/loader/initializer.go
Kubuxu: I was thinking about it, but a Plugin could in theory implement multiple interfaces, then typeswich isn't really helping us as it will choose just one interface (the first in the switch statement).

Miss-compiled plugin will error out or log if: the plugin lib can't be loaded, there is no `Plugins` variable, it has wrong type.

PR #4035, opened 2017-07-04
sharness: fix the error message in the mount-publish test case

Stebalien:


Stebalien:
Travis didn't catch this as it doesn't test FUSE.

whyrusleeping:
LGTM, good catch @Stebalien :)

As a side note, do the fuse bindings (and tests) work for you?

Stebalien:
I don't think I've been able to run them all the way through (mostly because some of the longer ones take *hours*).

Kubuxu:
The fuse tests like to hang for no reason. It got better when we updated the FUSE lib but it still might be the case.

Stebalien:
Many of them *do* finish, they just take a while (some of them seem to be hashing *lots* of data).

PR #4036, opened 2017-07-05
docs: add experimental-features.md

lgierth:
So I was thinking whenever we do something about experimental features, instead of posting in https://github.com/ipfs/go-ipfs/issues/3397 we should update this new text document, and link the PR in that issue.

### DIFF
--- a/docs/experimental-features.md
+++ b/docs/experimental-features.md
@@ -202,11 +202,11 @@ P2P command needs to be enabled in config
 Basic usage:
 
 - Open a listener on one node (node A)
-`p2p listener open p2p-test /ip4/127.0.0.1/tcp/10101`
-- Connect your app to the endpoint opened locally (tcp on localhost port 10101)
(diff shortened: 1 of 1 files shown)


docs/experimental-features.md
```diff
@@ -0,0 +1,214 @@
...
+
+`ipfs config --json Experimental.Libp2pStreamMounting true`
+
+### How to use
+
+Basic usage:
+
+- Open a listener on one node (node A)
+`p2p listener open p2p-test /ip4/127.0.0.1/tcp/10101`
+- Connect your app to the endpoint opened locally (tcp on localhost port 10101)
```
magik6k: This is not exactly correct, I updated the comment in the experimental features issue (https://github.com/ipfs/go-ipfs/issues/3397#issuecomment-309306293)

whyrusleeping: @magik6k wanna just push to this branch?

magik6k: done

PR #4042, opened 2017-07-06
remove all instances of `Default(false)`

asymmetric:
Boolean options always default to false.

As mentioned [here](https://github.com/ipfs/go-ipfs/pull/2667#issuecomment-312904012).


Kubuxu:
This should wait until https://github.com/ipfs/go-ipfs/pull/3856 is merged as it can have conflicts with it.

PR #4045, opened 2017-07-06
Add test init profile

magik6k:
This will be useful for sharness tests of external tools. Goal of this profile is to reduce external interference as much as possible.

### DIFF
--- a/repo/config/profile.go
+++ b/repo/config/profile.go
@@ -28,4 +28,17 @@ var ConfigProfiles = map[string]func(*Config) error{
 		c.Discovery.MDNS.Enabled = false
 		return nil
 	},
+	"test": func(c *Config) error {
+		c.Addresses.API = "/ip4/127.0.0.1/tcp/0"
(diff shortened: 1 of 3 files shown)


repo/config/profile.go
```diff
@@ -28,4 +28,16 @@ var ConfigProfiles = map[string]func(*Config) error{
...
+		c.Addresses.API = "/ip4/127.0.0.1/tcp/0"
+		c.Addresses.Gateway = "/ip4/127.0.0.1/tcp/0"
+
+		c.Addresses.Swarm = []string{
+			"/ip4/127.0.0.1/tcp/0",
+		}
+
+		c.Bootstrap = []string{}
+		c.Discovery.MDNS.Enabled = false
+		return nil
```
whyrusleeping: should also set `Swarm.DisableNatPortMap` to true

magik6k: Done

PR #4047, opened 2017-07-07
namesys/pubsub: pubsub Publisher and Resolver

vyzo:
Implements a namesys publisher and resolver with pubsub.

### DIFF
--- a/namesys/namesys.go
+++ b/namesys/namesys.go
@@ -2,14 +2,20 @@ package namesys
 
 import (
 	"context"
+	"errors"
 	"strings"
+	"sync"
 	"time"
 
 	path "github.com/ipfs/go-ipfs/path"
 
 	routing "gx/ipfs/QmPR2JzfKd9poHx9XBhzoFeBBC31ZM3W5iUPKJZWyaoZZm/go-libp2p-routing"
+	p2phost "gx/ipfs/QmRS46AyqtpJBsf1zmQdeizSDEzo1qkWR7rdEuPFAv8237/go-libp2p-host"
+	mh "gx/ipfs/QmU9a9NV9RdPNwZQDYd5uKsm6N6LJLSvLbywDDYFbaaC6P/go-multihash"
+	floodsub "gx/ipfs/QmVNv1WV6XxzQV4MBuiLX5729wMazaf8TNzm2Sq6ejyHh7/go-libp2p-floodsub"
 	ds "gx/ipfs/QmVSase1JP7cq9QkPT46oNwdp9pT6kBkG3oqS14y3QcZjG/go-datastore"
 	peer "gx/ipfs/QmXYjuNuxVzXKJCfWasQk1RqkhVLDM9jtUKhqc2WPQmFSB/go-libp2p-peer"
+	isd "gx/ipfs/QmZmmuAXgX73UQmX1jRKjTGmjzq24Jinqkq8vzkBtno4uX/go-is-domain"
 	ci "gx/ipfs/QmaPbCnUMBohSGo3KnxEa2bHqyJVVeEEcwtqJAYxerieBo/go-libp2p-crypto"
 )
 
@@ -36,9 +42,26 @@ func NewNameSystem(r routing.ValueStore, ds ds.Datastore, cachesize int) NameSys
 			"dht":      NewRoutingResolver(r, cachesize),
 		},
 		publishers: map[string]Publisher{
-			"/ipns/": NewRoutingPublisher(r, ds),
+			"dht": NewRoutingPublisher(r, ds),
 		},
 	}
+}
+
+// AddPubsubNameSystem adds the pubsub publisher and resolver to the namesystem
+func AddPubsubNameSystem(ctx context.Context, ns NameSystem, host p2phost.Host, r routing.IpfsRouting, ds ds.Datastore, ps *floodsub.PubSub) error {
+	mpns, ok := ns.(*mpns)
+	if !ok {
+		return errors.New("unexpected NameSystem; not an mpns instance")
+	}
+
+	pkf, ok := r.(routing.PubKeyFetcher)
+	if !ok {
+		return errors.New("unexpected IpfsRouting; not a PubKeyFetcher instance")
+	}
+
+	mpns.resolvers["pubsub"] = NewPubsubResolver(ctx, host, r, pkf, ps)
+	mpns.publishers["pubsub"] = NewPubsubPublisher(ctx, host, ds, r, ps)
+	return nil
 }
 
 const DefaultResolverCacheTTL = time.Minute
@@ -72,38 +95,100 @@ func (ns *mpns) resolveOnce(ctx context.Context, name string) (path.Path, error)
 		return "", ErrResolveFailed
 	}
 
-	for protocol, resolver := range ns.resolvers {
-		log.Debugf("Attempting to resolve %s with %s", segments[2], protocol)
-		p, err := resolver.resolveOnce(ctx, segments[2])
-		if err == nil {
-			if len(segments) > 3 {
-				return path.FromSegments("", strings.TrimRight(p.String(), "/"), segments[3])
-			} else {
-				return p, err
-			}
+	makePath := func(p path.Path) (path.Path, error) {
+		if len(segments) > 3 {
+			return path.FromSegments("", strings.TrimRight(p.String(), "/"), segments[3])
+		} else {
+			return p, nil
 		}
 	}
+
+	// Resolver selection:
+	// 1. if it is a multihash resolve through "pubsub" (if available),
+	//    with fallback to "dht"
+	// 2. if it is a domain name, resolve through "dns"
+	// 3. otherwise resolve through the "proquint" resolver
+	key := segments[2]
+
+	_, err := mh.FromB58String(key)
+	if err == nil {
+		res, ok := ns.resolvers["pubsub"]
+		if ok {
+			p, err := res.resolveOnce(ctx, key)
+			if err == nil {
+				return makePath(p)
+			}
+		}
+
+		res, ok = ns.resolvers["dht"]
+		if ok {
+			p, err := res.resolveOnce(ctx, key)
+			if err == nil {
+				return makePath(p)
+			}
+		}
+
+		return "", ErrResolveFailed
+	}
+
+	if isd.IsDomain(key) {
+		res, ok := ns.resolvers["dns"]
+		if ok {
+			p, err := res.resolveOnce(ctx, key)
+			if err == nil {
+				return makePath(p)
+			}
+		}
+
+		return "", ErrResolveFailed
+	}
+
+	res, ok := ns.resolvers["proquint"]
+	if ok {
+		p, err := res.resolveOnce(ctx, key)
+		if err == nil {
+			return makePath(p)
+		}
+
+		return "", ErrResolveFailed
+	}
+
 	log.Warningf("No resolver found for %s", name)
 	return "", ErrResolveFailed
 }
 
 // Publish implements Publisher
 func (ns *mpns) Publish(ctx context.Context, name ci.PrivKey, value path.Path) error {
-	err := ns.publishers["/ipns/"].Publish(ctx, name, value)
-	if err != nil {
-		return err
-	}
-	ns.addToDHTCache(name, value, time.Now().Add(DefaultRecordTTL))
-	return nil
+	return ns.PublishWithEOL(ctx, name, value, time.Now().Add(DefaultRecordTTL))
 }
 
 func (ns *mpns) PublishWithEOL(ctx context.Context, name ci.PrivKey, value path.Path, eol time.Time) error {
-	err := ns.publishers["/ipns/"].PublishWithEOL(ctx, name, value, eol)
-	if err != nil {
-		return err
+	var dhtErr error
+
+	wg := &sync.WaitGroup{}
+	wg.Add(1)
+	go func() {
+		dhtErr = ns.publishers["dht"].PublishWithEOL(ctx, name, value, eol)
+		if dhtErr == nil {
+			ns.addToDHTCache(name, value, eol)
+		}
+		wg.Done()
+	}()
+
+	pub, ok := ns.publishers["pubsub"]
+	if ok {
+		wg.Add(1)
+		go func() {
+			err := pub.PublishWithEOL(ctx, name, value, eol)
+			if err != nil {
+				log.Warningf("error publishing %s with pubsub: %s", name, err.Error())
+			}
+			wg.Done()
+		}()
 	}
-	ns.addToDHTCache(name, value, eol)
-	return nil
+
+	wg.Wait()
+	return dhtErr
 }
 
 func (ns *mpns) addToDHTCache(key ci.PrivKey, value path.Path, eol time.Time) {
@@ -137,4 +222,17 @@ func (ns *mpns) addToDHTCache(key ci.PrivKey, value path.Path, eol time.Time) {
 		val: value,
 		eol: eol,
 	})
+}
+
+// GetResolver implements ResolverLookup
+func (ns *mpns) GetResolver(subs string) (Resolver, bool) {
+	res, ok := ns.resolvers[subs]
+	if ok {
+		ires, ok := res.(Resolver)
+		if ok {
+			return ires, true
+		}
+	}
+
+	return nil, false
 }
--- a/namesys/pubsub.go
+++ b/namesys/pubsub.go
@@ -1,0 +1,430 @@
+package namesys
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"strings"
+	"sync"
+	"time"
+
+	pb "github.com/ipfs/go-ipfs/namesys/pb"
+	path "github.com/ipfs/go-ipfs/path"
+	dshelp "github.com/ipfs/go-ipfs/thirdparty/ds-help"
+
+	cid "gx/ipfs/QmNp85zy9RLrQ5oQD4hPyS39ezrrXpcaa7R4Y9kxdWQLLQ/go-cid"
+	routing "gx/ipfs/QmPR2JzfKd9poHx9XBhzoFeBBC31ZM3W5iUPKJZWyaoZZm/go-libp2p-routing"
+	pstore "gx/ipfs/QmPgDWmTmuzvP7QE5zwo1TmjbJme9pmZHNujB2453jkCTr/go-libp2p-peerstore"
+	p2phost "gx/ipfs/QmRS46AyqtpJBsf1zmQdeizSDEzo1qkWR7rdEuPFAv8237/go-libp2p-host"
+	u "gx/ipfs/QmSU6eubNdhXjFBJBSksTp8kv8YRub8mGAPv8tVJHmL2EU/go-ipfs-util"
+	mh "gx/ipfs/QmU9a9NV9RdPNwZQDYd5uKsm6N6LJLSvLbywDDYFbaaC6P/go-multihash"
+	floodsub "gx/ipfs/QmVNv1WV6XxzQV4MBuiLX5729wMazaf8TNzm2Sq6ejyHh7/go-libp2p-floodsub"
+	ds "gx/ipfs/QmVSase1JP7cq9QkPT46oNwdp9pT6kBkG3oqS14y3QcZjG/go-datastore"
+	dssync "gx/ipfs/QmVSase1JP7cq9QkPT46oNwdp9pT6kBkG3oqS14y3QcZjG/go-datastore/sync"
+	peer "gx/ipfs/QmXYjuNuxVzXKJCfWasQk1RqkhVLDM9jtUKhqc2WPQmFSB/go-libp2p-peer"
+	proto "gx/ipfs/QmZ4Qi3GaRbjcx28Sme5eMH7RQjGkt8wHxt2a65oLaeFEV/gogo-protobuf/proto"
+	ci "gx/ipfs/QmaPbCnUMBohSGo3KnxEa2bHqyJVVeEEcwtqJAYxerieBo/go-libp2p-crypto"
+	record "gx/ipfs/QmbxkgUceEcuSZ4ZdBA3x74VUDSSYjHYmmeEqkjxbtZ6Jg/go-libp2p-record"
+	dhtpb "gx/ipfs/QmbxkgUceEcuSZ4ZdBA3x74VUDSSYjHYmmeEqkjxbtZ6Jg/go-libp2p-record/pb"
+)
+
+// PubsubPublisher is a publisher that distributes IPNS records through pubsub
+type PubsubPublisher struct {
+	ctx  context.Context
+	ds   ds.Datastore
+	host p2phost.Host
+	cr   routing.ContentRouting
+	ps   *floodsub.PubSub
+
+	mx   sync.Mutex
+	subs map[string]struct{}
+}
+
+// PubsubResolver is a resolver that receives IPNS records through pubsub
+type PubsubResolver struct {
+	ctx  context.Context
+	ds   ds.Datastore
+	host p2phost.Host
+	cr   routing.ContentRouting
+	pkf  routing.PubKeyFetcher
+	ps   *floodsub.PubSub
+
+	mx   sync.Mutex
+	subs map[string]*floodsub.Subscription
+}
+
+// NewPubsubPublisher constructs a new Publisher that publishes IPNS records through pubsub.
+// The constructor interface is complicated by the need to bootstrap the pubsub topic.
+// This could be greatly simplified if the pubsub implementation handled bootstrap itself
+func NewPubsubPublisher(ctx context.Context, host p2phost.Host, ds ds.Datastore, cr routing.ContentRouting, ps *floodsub.PubSub) *PubsubPublisher {
+	return &PubsubPublisher{
+		ctx:  ctx,
+		ds:   ds,
+		host: host, // needed for pubsub bootstrap
+		cr:   cr,   // needed for pubsub bootstrap
+		ps:   ps,
+		subs: make(map[string]struct{}),
+	}
+}
+
+// NewPubsubResolver constructs a new Resolver that resolves IPNS records through pubsub.
+// same as above for pubsub bootstrap dependencies
+func NewPubsubResolver(ctx context.Context, host p2phost.Host, cr routing.ContentRouting, pkf routing.PubKeyFetcher, ps *floodsub.PubSub) *PubsubResolver {
+	return &PubsubResolver{
+		ctx:  ctx,
+		ds:   dssync.MutexWrap(ds.NewMapDatastore()),
+		host: host, // needed for pubsub bootstrap
+		cr:   cr,   // needed for pubsub bootstrap
+		pkf:  pkf,
+		ps:   ps,
+		subs: make(map[string]*floodsub.Subscription),
+	}
+}
+
+// Publish publishes an IPNS record through pubsub with default TTL
+func (p *PubsubPublisher) Publish(ctx context.Context, k ci.PrivKey, value path.Path) error {
+	return p.PublishWithEOL(ctx, k, value, time.Now().Add(DefaultRecordTTL))
+}
+
+// PublishWithEOL publishes an IPNS record through pubsub
+func (p *PubsubPublisher) PublishWithEOL(ctx context.Context, k ci.PrivKey, value path.Path, eol time.Time) error {
+	id, err := peer.IDFromPrivateKey(k)
+	if err != nil {
+		return err
+	}
+
+	_, ipnskey := IpnsKeysForID(id)
+
+	seqno, err := p.getPreviousSeqNo(ctx, ipnskey)
+	if err != nil {
+		return err
+	}
+
+	seqno++
+
+	return p.publishRecord(ctx, k, value, seqno, eol, ipnskey, id)
+}
+
+func (p *PubsubPublisher) getPreviousSeqNo(ctx context.Context, ipnskey string) (uint64, error) {
+	// the datastore is shared with the routing publisher to properly increment and persist
+	// ipns record sequence numbers.
+	prevrec, err := p.ds.Get(dshelp.NewKeyFromBinary([]byte(ipnskey)))
+	if err != nil {
+		if err == ds.ErrNotFound {
+			// None found, lets start at zero!
+			return 0, nil
+		}
+		return 0, err
+	}
+
+	prbytes, ok := prevrec.([]byte)
+	if !ok {
+		return 0, fmt.Errorf("unexpected type returned from datastore: %#v", prevrec)
+	}
+
+	var dsrec dhtpb.Record
+	err = proto.Unmarshal(prbytes, &dsrec)
+	if err != nil {
+		return 0, err
+	}
+
+	var entry pb.IpnsEntry
+	err = proto.Unmarshal(dsrec.GetValue(), &entry)
+	if err != nil {
+		return 0, err
+	}
+
+	return entry.GetSequence(), nil
+}
+
+func (p *PubsubPublisher) publishRecord(ctx context.Context, k ci.PrivKey, value path.Path, seqno uint64, eol time.Time, ipnskey string, ID peer.ID) error {
+	entry, err := CreateRoutingEntryData(k, value, seqno, eol)
+	if err != nil {
+		return err
+	}
+
+	data, err := proto.Marshal(entry)
+	if err != nil {
+		return err
+	}
+
+	// the datastore is shared with the routing publisher to properly increment and persist
+	// ipns record sequence numbers; so we need to Record our new entry in the datastore
+	dsrec, err := record.MakePutRecord(k, ipnskey, data, true)
+	if err != nil {
+		return err
+	}
+
+	dsdata, err := proto.Marshal(dsrec)
+	if err != nil {
+		return err
+	}
+
+	err = p.ds.Put(dshelp.NewKeyFromBinary([]byte(ipnskey)), dsdata)
+	if err != nil {
+		return err
+	}
+
+	// now we publish, but we also need to bootstrap pubsub for our messages to propagate
+	topic := "/ipns/" + ID.Pretty()
+
+	p.mx.Lock()
+	_, ok := p.subs[topic]
+
+	if !ok {
+		p.subs[topic] = struct{}{}
+		p.mx.Unlock()
+
+		bootstrapPubsub(p.ctx, p.cr, p.host, topic)
+	} else {
+		p.mx.Unlock()
+	}
+
+	log.Debugf("PubsubPublish: publish IPNS record for %s (%d)", topic, seqno)
+	return p.ps.Publish(topic, data)
+}
+
+// Resolve resolves a name through pubsub and default depth limit
+func (r *PubsubResolver) Resolve(ctx context.Context, name string) (path.Path, error) {
+	return r.ResolveN(ctx, name, DefaultDepthLimit)
+}
+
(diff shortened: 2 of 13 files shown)


namesys/pubsub.go
```diff
@@ -0,0 +1,346 @@
...
+	} else {
+		p.mx.Unlock()
+	}
+
+	log.Debugf("PubsubPublish: publish IPNS record for %s", topic)
+	return p.ps.Publish(topic, data)
+}
+
+// Resolve resolves a name through pubsub and default depth limit
+func (r *PubsubResolver) Resolve(ctx context.Context, name string) (value path.Path, err error) {
```
whyrusleeping: I generally don't like named return values, they almost always make the code harder to read.

namesys/pubsub.go
```diff
@@ -0,0 +1,346 @@
...
+			return
+		}
+
+		r.subs[name] = sub
+		go r.handleSubscription(sub, name, pubk)
+		go bootstrapPubsub(r.ctx, r.cr, r.host, name)
+	}
+
+	// resolve to what we may already have in the datastore
+	dsval, err := r.ds.Get(dshelp.NewKeyFromBinary([]byte(name)))
```
whyrusleeping: datastores in the real world will only store and return bytes, we will need to serialize and deserialize

namesys/pubsub.go
```diff
@@ -0,0 +1,346 @@
...
+		// this signals ds.ErrNotFound when we have no record
+		return
+	}
+
+	entry := dsval.(*pb.IpnsEntry)
+
+	// check EOL; if the entry has expired, delete from datastore and return ds.ErrNotFound
+	eol, ok := checkEOL(entry)
+	if ok && eol.Before(time.Now()) {
+		r.ds.Delete(dshelp.NewKeyFromBinary([]byte(name)))
```
whyrusleeping: should check errors from delete

namesys/pubsub.go
```diff
@@ -0,0 +1,346 @@
...
+		return "", ds.ErrNotFound
+	}
+
+	value, err = path.ParsePath(string(entry.GetValue()))
+	return
+}
+
+func (r *PubsubResolver) handleSubscription(sub *floodsub.Subscription, name string, pubk ci.PubKey) {
```
whyrusleeping: hrm... we don't really have a way to kill this. might be nice to have

namesys/pubsub.go
```diff
@@ -0,0 +1,346 @@
...
+	dsval, err := r.ds.Get(dshelp.NewKeyFromBinary([]byte(name)))
+	if err == nil {
+		oentry := dsval.(*pb.IpnsEntry)
+		if entry.GetSequence() <= oentry.GetSequence() {
+			return errors.New("stale update; sequence number too small")
+		}
+	}
+
+	log.Debugf("PubsubResolve: receive IPNS record for %s", name)
+	r.ds.Put(dshelp.NewKeyFromBinary([]byte(name)), entry)
```
whyrusleeping: check errors

namesys/pubsub.go
```diff
@@ -0,0 +1,346 @@
...
+	eol, ok := checkEOL(entry)
+	if ok && eol.Before(time.Now()) {
+		return errors.New("stale update; EOL exceeded")
+	}
+
+	r.mx.Lock()
+	defer r.mx.Unlock()
+
+	// check the sequence number against what we may already have in our datastore
+	dsval, err := r.ds.Get(dshelp.NewKeyFromBinary([]byte(name)))
```
whyrusleeping: serialization, this will come back as bytes (just doesnt happen to with the in memory datastore)

whyrusleeping:
@vyzo I think the flakiness is probably caused by using the routing system to bootstrap in the tests. I think (at least in this single go pubsub test) we should manually connect the hosts, something like:

```diff
        pub := NewPubsubPublisher(ctx, pubhost, ds.NewMapDatastore(), pubmr, floodsub.NewFloodSub(ctx, pubhost))
        privk := pubhost.Peerstore().PrivKey(pubhost.ID())
+       pubpinfo := pstore.PeerInfo{ID: pubhost.ID(), Addrs: pubhost.Addrs()}
 
        name := "/ipns/" + pubhost.ID().Pretty()
 
@@ -109,18 +111,22 @@ func TestPubsubPublishResolve(t *testing.T) {
        res := make([]Resolver, len(reshosts))
        for i := 0; i < len(res); i++ {
                res[i] = NewPubsubResolver(ctx, reshosts[i], resmrs[i], ks, floodsub.NewFloodSub(ctx, reshosts[i]))
+               if err := reshosts[i].Connect(ctx, pubpinfo); err != nil {
+                       t.Fatal(err)
+               }
        }
```

We can test the bootstrap mechanism explicitly in a different test.

vyzo:
re: bootstrap I agree that the flakiness is probably caused because of bootstrap issues. I think we want some delay right after opening a connection  in bootstrapPubsub, so that pubsub has time to do its handshake.

vyzo:
did a rebase to update and resolve conflicts and force pushed on top, so commits are out of order with review comments.

vyzo:
Turned the single test into two, one with pre-connected overlay and the other with bootstrap. 
We might want to finetune the bootstrap test to do less and just test bootstrap, but let's see how well it does with travis.

vyzo:
Actually, the bootstrap test does very little (travis still hasn't scheduled a build job) -- i think we can drop it for now and figure out a way to properly test bootstrap with a more reliable networking substrate.

namesys/pubsub.go
vyzo: removed the usage of named values in https://github.com/ipfs/go-ipfs/pull/4047/commits/62e427c1b5284c5f75f5ee774338df182b0493c2

namesys/pubsub.go
vyzo: Using binary serialization in https://github.com/ipfs/go-ipfs/pull/4047/commits/7e54b97fb422c406a541978a87fa57d0c5e8273c

namesys/pubsub.go
vyzo: done

namesys/pubsub.go
vyzo: done in https://github.com/ipfs/go-ipfs/pull/4047/commits/ed171a7354e9986f1bff8ec901961bcf7fdea603

namesys/pubsub.go
vyzo: done in https://github.com/ipfs/go-ipfs/pull/4047/commits/7e54b97fb422c406a541978a87fa57d0c5e8273c

namesys/pubsub.go
vyzo: Added a PubsubResolver.Cancel method in https://github.com/ipfs/go-ipfs/pull/4047/commits/176fcf185b8ed450d2430522cd3dbb3620192e51

Stebalien:
> One solution is to change the validity field of IPNS records to state where they can be published.

Actually, another solution is to wrap records in a second (signed) pubsub specific record. That avoids breaking compatibility but is a bit wasteful.

Stebalien:
>> One solution is to change the validity field of IPNS records to state where they can be published.

> Actually, another solution is to wrap records in a second (signed) pubsub specific record. That avoids breaking compatibility but is a bit wasteful.

So, @whyrusleeping actually prefers this method (this is actually how the DHT prevents users from publishing other user's IPNS records to the DHT).

vyzo:
> If a publisher does not have pubsub enabled, an attacker can broadcast old but still valid IPNS records via pubsub
>> Actually, another solution is to wrap records in a second (signed) pubsub specific record. That avoids breaking compatibility but is a bit wasteful.

I think we need to have signed messages in pubsub to properly solve this; it covers the poisoning attack and all future avenues for similar attacks without requiring backwards incompatible changes.

vyzo:
> So, how about introducing a PubsubValueStore? Then, instead of passing the DHT to NewRoutingResolver, we'd pass a MultiValueStore that queries multiple ValueStores (the DHT and the pubsub based one) in parallel.

We were discussing unifying the caches with @whyrusleeping and having both resolvers share a cache in the persistent datastore; we decided to leave it for a subsequent PR.

vyzo:
> Router versus ValueStore
> ...
> Thoughts?

@Stebalien 
I agree that we need more unification between the two resolvers, with a minimum of shared cache and validation code.
However, I think it's perhaps better to do it in a follow-up PR --  I didn't want to expand the scope of this PR too much so as to touch the routing resolver.

Stebalien:
> I think we need to have signed messages in pubsub to properly solve this; it covers the poisoning attack and all future avenues for similar attacks without requiring backwards incompatible changes.

Yeah, that's effectively the wrapper solution but pushed down a level. I still have some reservations about this method but I'll have to think on them a bit.

> We were discussing unifying the caches

Sharing the cache is really a secondary goal for me; my primary concern is reducing code duplication (especially reducing duplication of security related code). I think implementing a ValueStore instead of a publisher/resolver would actually be less code.

>  I didn't want to expand the scope of this PR too much so as to touch the routing resolver.

You wouldn't actually be changing the routing resolver, just passing it a different ValueStore (one that dispatches to the DHT and optionally pubsub).

vyzo:
> You wouldn't actually be changing the routing resolver, just passing it a different ValueStore (one that dispatches to the DHT and optionally pubsub).

There is a complication in terms of usage. The pubsub namesys is only added when the pubsub experiment is enabled, which means that we cannot fully construct this value store at resolver initialization -- we would have to track it somewhere and add pubsub to it post-construction. Nothing insurmountable though, just ugliness :)

Stebalien:
Can't we just construct pubsub earlier?

vyzo:
Not sure; it currently only needs a host, but this may change in the future -- eg if we move the bootstrap inside the pubsub interface as we've been discussing with why.

vyzo:
@Stebalien  So I am not sure at all that hiding the pubsub namesys inside a ValueStore is the correct approach:
- it hides pubsub behind an opaque interface, which makes it harder to see it's in use at all. It also drastically changes the design of the code and mixes responsibility for the value store for something that should be a separate component imo.
- It doesn't play nice with the optional nature of pubsub and the delayed initialization, and I don't quite want to move the pubsub initialization around (at least not just yet)
- I don't think it's going to be less code, as we would have to implement a new interface and duplicate code from the dht
- I don't think there is that much code duplicaiton anyway, not for what is usual in Go at least. The sensitive parts are reuse existing code already and the rest is simple logic.

@whyrusleeping thoughts?

vyzo:
>>  I think we need to have signed messages in pubsub to properly solve this; it covers the poisoning attack and all future avenues for similar attacks without requiring backwards incompatible changes.

> Yeah, that's effectively the wrapper solution but pushed down a level. I still have some reservations about this method but I'll have to think on them a bit.

We can do this in two ways: either we blanket sign all messages in pubsub, or we export a new method `PublishSigned()` and we use that one for ipns records and expect all messages to be signed in the resolver.

Stebalien:
> So I am not sure at all that hiding the pubsub namesys inside a ValueStore is the correct approach

I disagree but, after looking into it, I agree that doing this properly may take a while so we should probably fix it later.

> it hides pubsub behind an opaque interface, which makes it harder to see it's in use at all. It also drastically changes the design of the code and mixes responsibility for the value store for something that should be a separate component imo.

IMO, pubsub is, logically, a valuestore (or a multi-valuestore). That is, you associate (publish) values (messages) with keys (topics).

> It doesn't play nice with the optional nature of pubsub and the delayed initialization, and I don't quite want to move the pubsub initialization around (at least not just yet)

I agree. My plan is to replace `IpfsNode.Routing` with a pluggable parallel multi-router that can query multiple sub-routers in parallel. To initialize pubsub routing, you'd call `node.Routing.Register(myPubsubRouter)`.

I thought this would be simple. However, doing this properly requires a few changes to the `IpfsRouting` interface.

> I don't think it's going to be less code, as we would have to implement a new interface and duplicate code from the dht

It shouldn't have to duplicate much code from the DHT. Furthermore, it would reuse the existing record validation code from the DHT so you'd be able to add support for new record types to both the DHT and pubsub for for the price of one.

Also, any code needed by both the DHT and pubsub should be extracted and used by both, not duplicated (when possible given go's lack of a type system).

> I don't think there is that much code duplicaiton anyway, not for what is usual in Go at least. The sensitive parts reuse existing code already and the rest is simple logic.

All of the validation logic would be deduplicated (assuming we extract the validators).

---

However, we probably have two *very* different visions of how this will play out and the only way to settle this would be to actually write the code. Therefore, I agree that we should just go forward with this as-is and then refactor later (once I push some changes to the routing interfaces through).

---

> We can do this in two ways: either we blanket sign all messages in pubsub, or we export a new method `PublishSigned()` and we use that one for ipns records and expect all messages to be signed in the resolver.

For this to actually work, we'd have to *check* the signatures in pubsub. That means we'd have to specify a set of valid publishers for every topic up-front and I'm not sure we can reasonably do that for all pubsub usecases. I'm going to discuss this with the orbit people a bit.

Stebalien:
So, assuming we're going forward with this as-is, it would be nice to parallelize search over the DHT and pubsub (race them). That is, instead of checking pubsub then the DHT, fire off *blocking* pubsub search in parallel with a blocking DHT search and then cancel the context when one of them returns a value.

I'm planning on adding some way to get the last message broadcast on a topic from peers on subscribe so blocking on pubsub will likely allow us to learn about the IPNS record faster than querying the DHT (assuming we already know of some peers subscribed to the IPNS record in question).

vyzo:
>>  So I am not sure at all that hiding the pubsub namesys inside a ValueStore is the correct approach

> I disagree but, after looking into it, I agree that doing this properly may take a while so we should probably fix it later.

That was mostly my disagreement too, that doing something incisive would be a bigger effort and out of scope for the PR.

vyzo:
> So, assuming we're going forward with this as-is, it would be nice to parallelize search over the DHT and pubsub (race them). That is, instead of checking pubsub then the DHT, fire off blocking pubsub search in parallel with a blocking DHT search and then cancel the context when one of them returns a value.

The pubsub resolver should return immediately, as it retrieves values from an in-memory datastore. There is no blocking for the request, modulo subscribing to the topic which is a non-blocking operation (the bootstrap happens in a goroutine).

vyzo:
> I'm planning on adding some way to get the last message broadcast on a topic from peers on subscribe so blocking on pubsub will likely allow us to learn about the IPNS record faster than querying the DHT (assuming we already know of some peers subscribed to the IPNS record in question).

Sure, and when we have the blocking pubsub that retrieves last message, we can revisit the parallelization in the code.

vyzo:
> However, we probably have two very different visions of how this will play out and the only way to settle this would be to actually write the code. Therefore, I agree that we should just go forward with this as-is and then refactor later (once I push some changes to the routing interfaces through).

Our vision for the end goal is probably not so far away! 
I just tend to prefer implement and iterate, where we start a workable solution and minimally invasive changes and improve it over time.

vyzo:
> For this to actually work, we'd have to check the signatures in pubsub. That means we'd have to specify a set of valid publishers for every topic up-front and I'm not sure we can reasonably do that for all pubsub usecases.

We don't have to do it for all topics -- we already have the machinery in the form of topic descriptors. For the case of ipns pubsub we could specify the publishing peer as the only source, and we can sign/check messages only for topics with specific source.

vyzo:
Another possible issue, regarding pubusub bootstrap: memory of the DHT from Provide.
What happens if the provider records expire some 24hrs after publication? 
I think we need to periodically re-provide the bootstrap rendezvous.

namesys/namesys.go
```diff
@@ -36,11 +41,28 @@ func NewNameSystem(r routing.ValueStore, ds ds.Datastore, cachesize int) NameSys
...
+			"dht": NewRoutingPublisher(r, ds),
 		},
 	}
 }
 
+// AddPubsubNameSystem adds the pubsub publisher and resolver to the namesystem
+func AddPubsubNameSystem(ctx context.Context, ns NameSystem, host p2phost.Host, r routing.IpfsRouting, ds ds.Datastore, ps *floodsub.PubSub) error {
+	mpns, ok := ns.(*mpns)
+	if !ok {
+		return errors.New("Unexpected NameSystem; not an mpns instance")
```
Kubuxu: errors should start with lower case.

namesys/namesys.go
```diff
@@ -72,37 +94,85 @@ func (ns *mpns) resolveOnce(ctx context.Context, name string) (path.Path, error)
...
+	err := ns.publishers["dht"].PublishWithEOL(ctx, name, value, eol)
 	if err != nil {
 		return err
 	}
 	ns.addToDHTCache(name, value, eol)
+
+	pub, ok := ns.publishers["pubsub"]
+	if ok {
+		pub.PublishWithEOL(ctx, name, value, eol)
+	}
```
Kubuxu: Can we somehow make it parallel? DHT publish can take ~10s.

Kubuxu: Just an idea.

namesys/pubsub.go
```diff
@@ -0,0 +1,390 @@
...
+// PubsubPublisher is a publisher that distributes IPNS records through pubsub
+type PubsubPublisher struct {
+	ctx  context.Context
+	ds   ds.Datastore
+	host p2phost.Host
+	cr   routing.ContentRouting
+	ps   *floodsub.PubSub
+	subs map[string]struct{}
+	mx   sync.Mutex
+}
```
Kubuxu: Try to split structures with double line into logical part.

Example why: if mutex is at top of a structure it is clear it protects it all, if it is near one group it is clear that it protects this group.

Stebalien:
Sorry, I thought I had replied to this a while ago...

> What happens if the provider records expire some 24hrs after publication?

Our current system for advertising interest in a topic in the DHT is broken anyways. The provides mechanism is supposed to be used to indicate that one has some block, not that one is interested in a pubsub channel.

> We don't have to do it for all topics

We will eventually to prevent those DoS vectors and I'd like to avoid having an ad-hoc verification scheme split up across multiple levels.

For now, it may be best to merge this and fix the DoS issues later (just be sure to fix them before we make this non-experimental).

namesys/pubsub.go
```diff
@@ -0,0 +1,390 @@
...
+	path "github.com/ipfs/go-ipfs/path"
+	dshelp "github.com/ipfs/go-ipfs/thirdparty/ds-help"
+
+	pstore "gx/ipfs/QmPgDWmTmuzvP7QE5zwo1TmjbJme9pmZHNujB2453jkCTr/go-libp2p-peerstore"
+	routing "gx/ipfs/QmPjTrrSfE6TzLv6ya6VWhGcCgPrUAdcgrDcQyRDX2VyW1/go-libp2p-routing"
+	u "gx/ipfs/QmSU6eubNdhXjFBJBSksTp8kv8YRub8mGAPv8tVJHmL2EU/go-ipfs-util"
+	cid "gx/ipfs/QmTprEaAA2A9bst5XH7exuyi5KzNMK3SEDNN8rBDnKWcUS/go-cid"
+	mh "gx/ipfs/QmU9a9NV9RdPNwZQDYd5uKsm6N6LJLSvLbywDDYFbaaC6P/go-multihash"
+	ds "gx/ipfs/QmVSase1JP7cq9QkPT46oNwdp9pT6kBkG3oqS14y3QcZjG/go-datastore"
+	ds_sync "gx/ipfs/QmVSase1JP7cq9QkPT46oNwdp9pT6kBkG3oqS14y3QcZjG/go-datastore/sync"
```
Kubuxu: everything else uses `dssync`

namesys/pubsub.go
Kubuxu: you can also remove the named values in the function definition

namesys/pubsub.go
```diff
@@ -0,0 +1,390 @@
...
+	return p.ps.Publish(topic, data)
+}
+
+// Resolve resolves a name through pubsub and default depth limit
+func (r *PubsubResolver) Resolve(ctx context.Context, name string) (value path.Path, err error) {
+	return r.ResolveN(ctx, name, DefaultDepthLimit)
+}
+
+// ResolveN resolves a name through pubsub with the specified depth limit
+func (r *PubsubResolver) ResolveN(ctx context.Context, name string, depth int) (value path.Path, err error) {
```
Kubuxu: same here

namesys/pubsub.go
```diff
@@ -0,0 +1,390 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	// check EOL; if the entry has expired, delete from datastore and return ds.ErrNotFound
+	eol, ok := checkEOL(entry)
+	if ok && eol.Before(time.Now()) {
+		err = r.ds.Delete(dshelp.NewKeyFromBinary([]byte(name)))
+		if err != nil {
+			log.Warningf("PubsubResolve: error deleting stale value: %s", err.Error())
```
Kubuxu: might be nice to also have `name` that is being removed here

namesys/pubsub.go
vyzo: sure, will do.

namesys/pubsub.go
vyzo: war on named return values!

namesys/pubsub.go
vyzo: sure, will fix.

namesys/pubsub.go
vyzo: sorry, will make it consistent.

namesys/namesys.go
vyzo: Note sure if it's a good idea. It complicates the code a bit and we'll have to wait for the dht anyway in order to report an error.
We _could_ publish via pubsub first however, as this should be non-blocking.

namesys/pubsub.go
vyzo: good point.

namesys/namesys.go
vyzo: Speaking of which, I should catch and log errors from publishing in pubsub.

vyzo:
>>    What happens if the provider records expire some 24hrs after publication?

> Our current system for advertising interest in a topic in the DHT is broken anyways. The provides mechanism is supposed to be used to indicate that one has some block, not that one is interested in a pubsub channel.

I think long term we need to move the bootstrap mechanism inside the pubsub interface (when it's generalized from floodsub). For now, I think I will add a periodic re-provide of the rendezvous record, otherwise the pubsub rendezvous will not be visible for for nodes that try to join more than 24hrs after bootstrap.

vyzo:
I will be pushing updates before a final rebase, so that we can have a coherent review conversation.

namesys/namesys.go
vyzo: Thinking some more about this, the pubsub Publish can be blocking as it might have to perform bootstrap the first time it publishes something; this can take a few seconds.

So yeah, I will parallelize the publish. It's worth the complexity, since both can be potentially slow.

vyzo: Ok, this is a little complex with the error semantics. 
Should we consider it an error if one of them fails? And if multiple of them fail, what error do we return?

vyzo: I think I'll keep the existing serial semantics for now, I don't like the error handling complexity from parallel publish.

vyzo:
@whyrusleeping ping -- thoughts on this?

whyrusleeping:
@vyzo getting back to this now :)

Side note: @MichaelMure this PR should make ipns resolves much faster.

namesys/namesys.go
```diff
@@ -72,38 +95,100 @@ func (ns *mpns) resolveOnce(ctx context.Context, name string) (path.Path, error)
...
+	// Resolver selection:
+	// 1. if it is a multihash resolve through "pubsub" (if available),
+	//    with fallback to "dht"
+	// 2. if it is a domain name, resolve through "dns"
+	// 3. otherwise resolve through the "proquint" resolver
+	key := segments[2]
+
+	_, err := mh.FromB58String(key)
+	if err == nil {
+		res, ok := ns.resolvers["pubsub"]
```
whyrusleeping: This sort of hardcoding resolution choices is okay for now, but it will be good to figure out a generic way of making it configurable

namesys/namesys.go
```diff
@@ -72,38 +95,100 @@ func (ns *mpns) resolveOnce(ctx context.Context, name string) (path.Path, error)
...
 func (ns *mpns) PublishWithEOL(ctx context.Context, name ci.PrivKey, value path.Path, eol time.Time) error {
-	err := ns.publishers["/ipns/"].PublishWithEOL(ctx, name, value, eol)
-	if err != nil {
-		return err
+	var dhtErr error
+
+	wg := &sync.WaitGroup{}
+	wg.Add(1)
+	go func() {
+		dhtErr = ns.publishers["dht"].PublishWithEOL(ctx, name, value, eol)
```
whyrusleeping: I wonder if we should *only* publish to pubsub when in pubsub mode. cc @Stebalien

namesys/pubsub.go
```diff
@@ -0,0 +1,409 @@
...
+	if err != nil {
+		return err
+	}
+
+	dsdata, err := proto.Marshal(dsrec)
+	if err != nil {
+		return err
+	}
+
+	p.ds.Put(dshelp.NewKeyFromBinary([]byte(ipnskey)), dsdata)
```
whyrusleeping: error check here

namesys/pubsub.go
```diff
@@ -0,0 +1,409 @@
...
+		}
+
+		if entry.GetSequence() <= oentry.GetSequence() {
+			return errors.New("stale update; sequence number too small")
+		}
+	}
+
+	log.Debugf("PubsubResolve: receive IPNS record for %s", name)
+
+	err = r.ds.Put(dshelp.NewKeyFromBinary([]byte(name)), data)
```
whyrusleeping: can just `return r.ds.Put(...`

namesys/namesys.go
vyzo: Another approach is to resolve with both resolvers in parallel and use the IPNS record with the largest sequence number.

namesys/namesys.go
vyzo: Wouldn't this break clients who don't have pubsub enabled? We kind of have to publish to the dht to keep existing functionality of ipns intact.

namesys/pubsub.go
vyzo: done

namesys/pubsub.go
vyzo: yes indeed, done.

namesys/namesys.go
whyrusleeping: hrm... youre not *wrong*. But there may be value in having an option to only have pubsub publishing. In any case, we dont have to figure that out for this PR

whyrusleeping:
@vyzo lets add a way to enable or disable this independently of having the pubsub experiment enabled. Perhaps a flag or some subcommand on `ipfs name`? It would also be good to have a way to see the current state of the pubsub name system, names currently subscribed to, current values, time of last update, etc. And a way to unsubscribe from name updates would be good to have too. This should probably go in an `ipfs name pubsub` subcommand or something.

cc @diasdavid @Stebalien @Kubuxu for input here.

The code itself looks pretty good to me, just need some management tools now to move forward.

vyzo:
@whyrusleeping ok, will update once I get back from ICFP.

vyzo:
Let's call the flag `--enable-namesys-pubsub`, which will by necessity enable pubsub.

whyrusleeping:
> Let's call the flag --enable-namesys-pubsub, which will by necessity enable pubsub.

SGTM

vyzo:
Rebased on master.

vyzo:
Reduced bootstrap provide period to 8h and added the --enable-namesys-pubsub option to enable pubsub for IPNS as discussed.

namesys/pubsub.go
```diff
@@ -0,0 +1,427 @@
...
+	var res []string
+	for sub := range r.subs {
+		res = append(res, sub)
+	}
+
+	return res
+}
+
+// Cancel cancels a topic subscription
+func (r *PubsubResolver) Cancel(name string) {
```
whyrusleeping: maybe this should return a bool for whether or not the named thing was actually removed? I'm concerned about the use case of `Cancel("thing that doesnt exist")`

vyzo: Shouldn't that be a noop?

vyzo:
Extracted the unrelated testutil changes to #4281; will rebase once it's merged.

namesys/pubsub.go
vyzo: updated to return a boolean indicator in https://github.com/ipfs/go-ipfs/pull/4047/commits/2e6af50a12d445b673fec0121e1857dc7713b1c3

vyzo:
rebased on master.

namesys/namesys.go
vyzo: problem with that of course is DHT resolution latency.

whyrusleeping:
@vyzo could you rebase this? and maybe squash it down to a few commits?

vyzo:
sure, i can rebase! we can squash down at merge time, i don't mind it being a single commit.

namesys/pubsub.go
```diff
@@ -0,0 +1,430 @@
...
+
+	data := dsval.([]byte)
+	entry := new(pb.IpnsEntry)
+
+	err = proto.Unmarshal(data, entry)
+	if err != nil {
+		return "", err
+	}
+
+	// check EOL; if the entry has expired, delete from datastore and return ds.ErrNotFound
```
whyrusleeping: 'return ds.ErrNotFound` is wrong, we're returning `ErrResolveFailed`

whyrusleeping:
@vyzo a few questions:

1. This doesnt speed up the first resolution, right? If i join the network and do a resolve, i still have to use the dht, unless the publisher publishes again.
2. If I bring my node online, i get a few values resolved through pubsub, then i go offline, and the node publishes an update while i'm offline, then i come back online and resolve, what happens?
3. Is there a discussion anywhere on asking nodes for the latest value? (cc @Stebalien)
4. Any thought towards having an option for nodes to periodically rebroadcast their records? unclear  if this would help or not.

namesys/pubsub.go
vyzo: ok, will fix.

vyzo: wait a minute, there is  a check for ds.ErrNotFound, which explicitly gets translated to ErrResolveFailed.

vyzo:
rebased, and I decided to squash as well so that we have a cleaner multi-commit history.

vyzo:
@whyrusleeping:
>    This doesnt speed up the first resolution, right? If i join the network and do a resolve, i still have to use the dht, unless the publisher publishes again.

correct; the first resolution will be through the DHT.

>    If I bring my node online, i get a few values resolved through pubsub, then i go offline, and the node publishes an update while i'm offline, then i come back online and resolve, what happens?

You will miss that update.
Also, if your node actually shuts down, when you come back online your pubsub state is clean; you are not resubscribed automatically, as the subscriptions are not persisted to disk (perhaps they should though).

>    Is there a discussion anywhere on asking nodes for the latest value? (cc @Stebalien)

No; we don't have this kind of functionality in place anywhere afaict.

>    Any thought towards having an option for nodes to periodically rebroadcast their records? unclear if this would help or not.

It could help in case of lost messages, but not clear it adds much otherwise.

vyzo:
rebased on master

Stebalien:
> Is there a discussion anywhere on asking nodes for the latest value? (cc @Stebalien)

@keks is planning on looking into this as soon as I take a look at his issue on go-floodsub (https://github.com/libp2p/go-floodsub/issues/42).

I thought we discussed this somewhere but it must not have been in this issue (IRC? maybe not?). IIRC, I came to the conclusion that this was enough code for a first pass and that we can add features like that later (keep the PR small and manageable). The correct way to fix that is in floodsub itself (@keks' proposal) so it wouldn't have much bearing here.

whyrusleeping:
Ah, i remember the last thing here. The go-testutil package being imported is a little bit borked. Its imports arent quite right, so we end up with some duplicates. Could someone handle that?

vyzo:
is there anything needed in this PR?

vyzo:
Forced pushed the last commit to restart jenkins -- errored out with some system-specific gremlin.

vyzo:
Updated for changes in the command interface.

vyzo:
it's not clear why jenkins failed -- all the not ok's are marked as known breakages.
I also can't restart the build through the interface for some reason.

keks:
@vyzo I restarted both the broken travis build and the jenkins build. Both fail, but at different places.

Travis:
in test/sharness/t0183-namesys-pubsub.sh
```
expecting success: 
    echo -n "" > expected &&
    ipfsi 1 name pubsub subs > subs1 &&
    ipfsi 2 name pubsub subs > subs2 &&
    test_cmp expected subs1 &&
    test_cmp expected subs2

> diff -u expected subs1
--- expected	2017-11-19 13:20:38.000000000 +0000
+++ subs1	2017-11-19 13:20:38.000000000 +0000
@@ -1 +0,0 @@
--n 

not ok 18 - check subscriptions
```

It looks like `echo -n` prints `-n` on the Travis Mac builders.

Jenkins:
in test/sharness/t0088-repo-stat-symlink.sh
```
expecting success: 
  export IPFS_PATH="sym_link_target" &&
  reposize_direct=$(ipfs repo stat | grep RepoSize | awk '{ print $2 }') &&
  export IPFS_PATH=".ipfs" &&
  reposize_symlink=$(ipfs repo stat | grep RepoSize | awk '{ print $2 }') &&
  test $reposize_symlink -ge $reposize_direct

not ok 4 - 'ipfs repo stat' RepoSize is correct with sym link
```

vyzo:
> It looks like echo -n prints -n on the Travis Mac builders.

wow, that's very unfortunate. I wonder if there is a version of echo that works correctly, otherwise i'll have to use touch with a unique filename.

> in test/sharness/t0088-repo-stat-symlink.sh

that's completely unrelated to this PR.

vyzo:
Changed the sharness test to use `rm && touch` instead of `echo -n`.

vyzo:
Another jenkins failure, similar to the first one -- all the `not ok` are in known breakage.
What's up with our jenkins setup, why is it being flaky?

vyzo:
note that circle and travis were successful.
@keks can you give jenkins another restart to see if the pattern repeats?

whyrusleeping:
@vyzo we use `printf` instead of `echo -n` generally. Its the most cross platform thing i've found

whyrusleeping:
@Stebalien could you check out the gx deps here? I think that the go-testutil dep is a bit wonky (other things are importing the old go-testutil)

Stebalien:
Dammit. I did *not* mean to update that package. Let's try this again from the bottom.

Stebalien:
On the bright side, #4403 should fix that issue. Of course, it'll probably introduce a bunch of ever so fun merge conflicts.

whyrusleeping:
I squashed and rebased, the dependency issues should be fixed now

whyrusleeping:
Strange ipns related failure on jenkins, copied the failure logs here: https://gist.github.com/whyrusleeping/9f4bef1726577e14a67568b9cd997f5a

Any chance this is related to the pubsub changes @vyzo ?

cc @Stebalien as maybe this is a known random failure.

vyzo:
the tests seem to have a 300ms timeout -- maybe that's the cause?
I haven't seen this failing before, but normal ipns resolution shouldn't be affected

vyzo:
oh you squashed it!

vyzo:
@whyrusleeping can you rerun jenkins to see if the problem is persistent? 
it might be something post squash, or it might be a gremlin (or simply 300ms not being enough). I still don't have rebuild permissions on jenkins.

whyrusleeping:
@VictorBjelkholm can you trigger this one to be rebuilt? And also tell me how to do it myself?

PR #4051, opened 2017-07-08
Increase object input size limit to 8mbytes

bobobo1618:
I have objects larger than 2mbytes. It'd be good to document why exactly there's _any_ limit but I'm certainly not the person to do that.

License: MIT
Signed-off-by: Pending

### DIFF
--- a/core/commands/object/object.go
+++ b/core/commands/object/object.go
@@ -22,10 +22,10 @@ import (
 	cid "gx/ipfs/Qma4RJSuh7mMeJQYCqMbKzekn6EwBo7HEs5AQYjVRMQATB/go-cid"
 )
 
-// ErrObjectTooLarge is returned when too much data was read from stdin. current limit 2m
-var ErrObjectTooLarge = errors.New("input object was too large. limit is 2mbytes")
+// ErrObjectTooLarge is returned when too much data was read from stdin. current limit 8m
+var ErrObjectTooLarge = errors.New("input object was too large. limit is 8mbytes")
 
-const inputLimit = 2 << 20
+const inputLimit = 2 << 22
 
 type Node struct {
 	Links []Link


Stebalien:
Yes, that should be documented...

We do this because bitswap (the block (object) exchange protocol) limits the block size to 2MiB and we don't want to add blocks that we can't exchange with other nodes.

Bitswap limits the block size for security. Unfortunately, there's no way to verify *part* of a block against a sha256 hash so we need to download the entire block before we can verify that we've actually received the *right* block. If we didn't limit the block size, we'd have no way to distinguish between an attacker feeding us bogus data and someone sending us a very large block.

For now, if you have a large object, the best way to deal with this is to shard it up into smaller objects.

In the future, we may be able to use a hashing algorithm that allows us to validate pieces of blocks but this would significantly complicate block exchange protocols (blocks would no longer be atomic).

A real-world case where this is a problem is [git over IPFS](https://discuss.ipfs.io/t/git-on-ipfs-links-and-references/730/4).

---

To avoid a potential X-Y problem, would you mind sharing your use-case?

Stebalien:
Note: We've also discussed adding a layer  *on-top* of IPLD that allows one to create large logical objects that are transparently sharded into multiple IPLD objects.

bobobo1618:
> For now, if you have a large object, the best way to deal with this is to shard it up into smaller objects.

> To avoid a potential X-Y problem, would you mind sharing your use-case?

My large object is a directory which I'm mirroring from my filesystem to IPFS using filestore. It has 17k files and hashes link hashes make up ~800k of the 2MB with filenames making up most of the rest. I'm creating this directory object by building an object as a Python dict, serializing it to JSON and throwing that at the HTTP API.

I'm doing this because I need programmatic access to IPFS in order to run an incremental `nocopy` add, which I need because I'm trying (and other than this succeeding) to keep 4TB of local data available on IPFS.

Is there a way to break a directory object into multiple objects without representing it as multiple directories? I understand that I can break it up into multiple directories but I'd rather IPFS reflect my filesystem and I'm not willing to alter the structure of my dataset on disk to facilitate IPFS limitations.

bobobo1618:
It actually looks like my enormous directory is the result of a bug, which is embarassing. I'd still like to know how to handle large directories though.

Kubuxu:
@bobobo1618 To solve the directory issue you can enable directory sharding: `ipfs config --json 'Experimental.ShardingEnabled' true` and readd your data. Resulting directory will be split into multiple objects which can be transferred. 

Increasing object size here won't change anything, bitswap and libp2p will still limit the size.
Closing, if you encounter any more issues with large files/directories, please report.

PR #4060, opened 2017-07-12
Change IPFS to use the new pluggable Block to IPLD decoding framework.

Stebalien:
Later, we should:

1. Pull the other node formats out of IPFS (at least the raw one).
2. Pull out the decoder registration/management into a `go-ipld` library.


Stebalien:
Replaces #4040.

PR #4063, opened 2017-07-12
[WIP] Extract DAGService and friends

Stebalien:
**DO NOT MERGE!**

Refactor IPFS to use: https://github.com/ipfs/go-ipld-format/pull/8

Context: IPLD isn't currently very usable from libp2p because interfaces like the DAGService are bundled into go-ipfs. This PR extracts these interfaces and some of the related helper functions.

### DIFF
--- a/core/commands/pin.go
+++ b/core/commands/pin.go
@@ -17,6 +17,7 @@ import (
 	context "context"
 	u "gx/ipfs/QmSU6eubNdhXjFBJBSksTp8kv8YRub8mGAPv8tVJHmL2EU/go-ipfs-util"
 	cid "gx/ipfs/QmTprEaAA2A9bst5XH7exuyi5KzNMK3SEDNN8rBDnKWcUS/go-cid"
+	node "gx/ipfs/QmVHxZ8ovAuHiHTbJa68budGYAqmMUzb1bqDW1SVb6y5M9/go-ipld-format"
 )
 
 var PinCmd = &cmds.Command{
@@ -548,7 +549,7 @@ func pinLsAll(typeStr string, ctx context.Context, n *core.IpfsNode) (map[string
 	if typeStr == "indirect" || typeStr == "all" {
 		set := cid.NewSet()
 		for _, k := range n.Pinning.RecursiveKeys() {
-			err := dag.EnumerateChildren(n.Context(), n.DAG.GetLinks, k, set.Visit)
+			err := dag.EnumerateChildren(n.Context(), dag.GetLinksWithDAG(n.DAG), k, set.Visit)
 			if err != nil {
 				return nil, err
 			}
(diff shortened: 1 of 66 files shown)


kevina:
@Stebalien 

> We use a GetLinksDirect function here (and a few other places). Why do we do this? Was there any reason to bypass the LinkService or are we doing this because we didn't have a LinkService? @kevina?

Yes, because in some cases we need to make sure the nodes are available locally, using LinkService defeats this purpose.  For example FetchGraph uses it when calling EnumerateChildrenAsync.  If the link service was used then FetchGraph may not fetch some of the children (i.e. leaves).

kevina:
And I should add that other times we don't care if the Nodes exist locally we just want to visit (i.e. call the visit function) for each Cid in the Dag.  This is why `EnumerateChildren`/`EnumerateChildrenAsync` takes in a function that gets the links for a node, not the node itself.

Now that I think about it a bit, `GetLinksDirect` may not be strictly necessary as the `visit` function could retrieve the node itself (to make sure its available), but that just adds extra work.

In any case I am not sure how the presence of the `GetLinksDirect` effects your refactoring as it is a helper function not directly connected to the `DAGService`.

Stebalien:
> In any case I am not sure how the presence of the GetLinksDirect effects your refactoring as it is a helper function not directly connected to the DAGService.

`DAGService` no longer has a `GetLinks` function (it's a helper function that takes a dag and a cid) so one has to construct a curried `GetLinks`.

kevina:
> `DAGService` no longer has a `GetLinks` function (it's a helper function that takes a dag and a cid) so one has to construct a curried `GetLinks`.

Oh, okay.  Well I guess that is unavoidable as I hope we have established that `EumerateChildren`/`EnumerateChildrenAsync` should take a function that gets the links for a node, not the node itself.

whyrusleeping:
@Stebalien This looks good to me, i'm fine to get this moving forward

core/commands/pin.go
```diff
@@ -548,7 +549,7 @@ func pinLsAll(typeStr string, ctx context.Context, n *core.IpfsNode) (map[string
 	if typeStr == "indirect" || typeStr == "all" {
 		set := cid.NewSet()
 		for _, k := range n.Pinning.RecursiveKeys() {
-			err := dag.EnumerateChildren(n.Context(), n.DAG.GetLinks, k, set.Visit)
+			err := dag.EnumerateChildren(n.Context(), dag.GetLinksWithDAG(n.DAG), k, set.Visit)
```
kevina: @Stebalien I take it we cleared this up and you are okay with the current interface?

Stebalien: Yes, and thanks for answering my questions.

Stebalien:
(to explicitly keep this in the comment thread instead of quietly tracking my questions in my first comment)

@kevina got it, all questions answered. I'm moving forward with my original go-ipld-format PR as those changes appear to work.

kevina:
@Stebalien I have not gone over this is with a fine-tooth comb, but I am also okay with moving forward on this.

Stebalien:
This has served its purpose. I'll revive when I make progress on the linked PR.

PR #4067, opened 2017-07-14
gx: update ipldcbor

Stebalien:
I previously optimized the IPLD cbor decoder to *not* encode and then re-decode
objects when constructing them with `WrapObject`. Unfortunately, we rely on this
to canonicalize the object before computing the tree/links.


Kubuxu:
LGTM but I would prefer for https://github.com/ipfs/go-ipfs/pull/4033 to be merged first, resolving merge conflict will be easier on this side.

Stebalien:
Sure. It doesn't look like IPFS is hitting this bug anyways.

whyrusleeping:
@Stebalien wanna rebase?

PR #4069, opened 2017-07-16
Only open a message sender when we have messages to send

### LINKED ISSUE #3651: Occasional stream hang
Ive been observing that every so often under a certain workflow i get a stuck stream. The workflow that i've been doing when this occurs is as follows:

- mars is running, and runs the gexpin service. 
- My laptop ipfs node is adding content (gx packages) and then i pin those packages on mars through the gexpin site
- The process works most of the time, but occasionally hangs

Things i've noticed:
- Mars' wantlist contains the hash i asked it to pin, but my nodes view of mars wantlist does not contain that (`ipfs bitswap wantlist -p QmaCpd...` shows nothing). This implies that mars is hanging while sending me their wantlist and not that my node is hanging while sending the blocks. Which then further implies that the wantmanager stream that we use to send wantlist updates has something wrong with it.
- Mars OOM crashes and reboots roughly once a day (bootstrap node + only 1GB ram == pain)
- This issue is correlated *i think* with my laptop either losing wifi connection or going to sleep. Though i have slept my laptop, opened it back up, and had things work just fine.
- I first noticed the issue with yamux, and tried out using multiplex to see if it would fix the issue, but i still managed to get a repro.
  - the more interesting thing here is that the failure behaviour was slightly different under yamux vs multiplex

yamux vs multiplex:

under yamux, when this bug occurred, i could still ping mars from my node (ping opens a new stream, so this tells us that just one stream is hung and not the multiplexer routine). Under multiplex, this was not possible, no new stream could be opened, and `ipfs ping MARS` just hung, meaning that we couldnt open a new stream, implying that our multiplex writer routine was blocked somewhere and that somewhere on mars' side, a multiplex stream was full of data with nothing to read it.

My next steps will be to deeply investigate the wantmanager wantlist stream sender code to make sure we arent accidentally dropping a stream without closing it, and to look for any other place where such a thing might happen.

whyrusleeping:
resolves #3651 for the most part. Theres still a deeper fix that needs to be done to eliminate this class of errors. 

License: MIT

### DIFF
--- a/exchange/bitswap/testnet/peernet.go
+++ b/exchange/bitswap/testnet/peernet.go
@@ -1,7 +1,7 @@
 package bitswap
 
 import (
-	context "context"
+	"context"
 	bsnet "github.com/ipfs/go-ipfs/exchange/bitswap/network"
 	mockrouting "github.com/ipfs/go-ipfs/routing/mock"
 	testutil "github.com/ipfs/go-ipfs/thirdparty/testutil"
@@ -37,4 +37,4 @@ func (pn *peernet) HasPeer(p peer.ID) bool {
 	return false
 }
 
-var _ Network = &peernet{}
+var _ Network = (*peernet)(nil)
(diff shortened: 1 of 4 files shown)


whyrusleeping:
except for all the tests failing... great.

PR #4072, opened 2017-07-16
fix determination of 'root' node in dag put

whyrusleeping:
License: MIT


whyrusleeping:
If a parser returns a set of nodes, the first one in the returned slice should be the 'root' (e.g. the bitcoin 'block' node, followed by all the transactions)

Issue #3896, opened 2017-05-02
Changes for Ed25519 public key IPFS identities


JustinDrake:
Below are the changes to support Ed25519 public key identity types. It was taken from https://github.com/ipfs/notes/issues/241 incorporating comments from @whyrusleeping and @Kubuxu.

- multiformats/multicodec (@JustinDrake)
  - [X] ~~table.csv: add Ed25519 codec (PR https://github.com/multiformats/multicodec/pull/46)~~
  - [X] table.csv: add identity multihash (PR https://github.com/multiformats/multicodec/pull/50)
- ~~multiformats/go-multicodec-packed (@JustinDrake)~~
  - [X] ~~multicodec.go: add Ed25519 codec (PR https://github.com/multiformats/go-multicodec-packed/pull/10)~~
- ~~multiformats/go-multihash (@JustinDrake)~~
  - [X] ~~multihash.go: add support for the identity hash (PR https://github.com/multiformats/go-multihash/pull/51)~~
- ~~libp2p/go-libp2p-crypto (@JustinDrake)~~
  - [X] ~~ed25519.go: add `UnmarshalEd25519PublicKey` helper function (PR https://github.com/libp2p/go-libp2p-crypto/pull/24)~~
- libp2p/go-libp2p-peer
  - [X] ~~peer.go: Add `ExtractEd25519PublicKey` and `ExtractPublicKey` helper functions (@JustinDrake) (PR https://github.com/libp2p/go-libp2p-peer/pull/15)~~
  - [ ] peer.go: Generalise `IDFromPublicKey` and `IDFromPrivateKey` functions beyond SHA2_256. (As pointed by @whyrusleeping this may be tricky. Most of the time context should be sufficient to determine public key type. If not we may have to resort to heuristics and defaults?) (@whyrusleeping)
- ipfs/go-ipfs-util (@whyrusleeping)
  - [ ] util.go: Remove dependence of `Hash` function which only supports SHA2_256
- libp2p/go-libp2p-kad-dht (@JustinDrake)
  - [X] ~~records.go: Have `GetPublicKey ` extract public key instead of querying caches or the DHT for new identity type (PR https://github.com/libp2p/go-libp2p-kad-dht/pull/57)~~
- ipfs/go-ipfs (@JustinDrake)
[...]

PR #4076, opened 2017-07-17
Add option to specify key type (RSA or Ed25519)

JustinDrake:
See https://github.com/ipfs/go-ipfs/issues/3896 for context

License: MIT
Signed-off-by: Justin Drake <drakefjustin@gmail.com>

### DIFF
--- a/cmd/ipfs/init.go
+++ b/cmd/ipfs/init.go
@@ -16,10 +16,13 @@ import (
 	namesys "github.com/ipfs/go-ipfs/namesys"
 	config "github.com/ipfs/go-ipfs/repo/config"
 	fsrepo "github.com/ipfs/go-ipfs/repo/fsrepo"
+	ci "gx/ipfs/QmaPbCnUMBohSGo3KnxEa2bHqyJVVeEEcwtqJAYxerieBo/go-libp2p-crypto"
 )
 
+// The default keypair is 2048-bit RSA
 const (
 	nBitsForKeypairDefault = 2048
+	keypairTypeDefault     = ci.RSA
 )
 
 var initCmd = &cmds.Command{
@@ -49,6 +52,7 @@ environment variable:
 	},
 	Options: []cmds.Option{
 		cmds.IntOption("bits", "b", "Number of bits to use in the generated RSA private key.").Default(nBitsForKeypairDefault),
+		cmds.IntOption("key-type", "k", "Key type (RSA or Ed25519-id").Default(ci.RSA),
 		cmds.BoolOption("empty-repo", "e", "Don't add and pin help files to the local storage.").Default(false),
 		cmds.StringOption("profile", "p", "Apply profile settings to config. Multiple profiles can be separated by ','"),
 
@@ -90,6 +94,12 @@ environment variable:
 			return
 		}
 
+		keyType, _, err := req.Option("k").Int()
+		if err != nil {
+			res.SetError(err, cmds.ErrNormal)
+			return
+		}
+
 		var conf *config.Config
 
 		f := req.Files()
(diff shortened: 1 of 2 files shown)


cmd/ipfs/init.go
```diff
@@ -90,6 +94,12 @@ environment variable:
 			return
 		}
 
+		keyType, _, err := req.Option("k").Int()
```
magik6k: Use the longer option name here

JustinDrake: I used the short option name for consistency. See `empty, _, err := req.Option("e").Bool()` and `nBitsForKeypair, _, err := req.Option("b").Int()` right above.

Kubuxu: Most other commands use long option names. Looks this one was not migrated.

cmd/ipfs/init.go
```diff
@@ -49,6 +52,7 @@ environment variable:
 	},
 	Options: []cmds.Option{
 		cmds.IntOption("bits", "b", "Number of bits to use in the generated RSA private key.").Default(nBitsForKeypairDefault),
+		cmds.IntOption("key-type", "k", "Key type (RSA or Ed25519-id").Default(ci.RSA),
```
Kubuxu: Use the default you defined in line 25.

cmd/ipfs/init.go
```diff
@@ -16,10 +16,13 @@ import (
 	namesys "github.com/ipfs/go-ipfs/namesys"
 	config "github.com/ipfs/go-ipfs/repo/config"
 	fsrepo "github.com/ipfs/go-ipfs/repo/fsrepo"
+	ci "gx/ipfs/QmaPbCnUMBohSGo3KnxEa2bHqyJVVeEEcwtqJAYxerieBo/go-libp2p-crypto"
```
Kubuxu: Separate local and gx deps with a new line.

JustinDrake: Done 👍

cmd/ipfs/init.go
JustinDrake: I've used the long name for all the options that were using the short name in this file. 👍

cmd/ipfs/init.go
```diff
@@ -49,6 +53,7 @@ environment variable:
 	},
 	Options: []cmds.Option{
 		cmds.IntOption("bits", "b", "Number of bits to use in the generated RSA private key.").Default(nBitsForKeypairDefault),
+		cmds.IntOption("key-type", "k", "Key type (RSA or Ed25519-id").Default(keypairTypeDefault),
```
whyrusleeping: appears to be a typo here, missing paren

whyrusleeping:
One comment, then would be great to have a test that tests ipfs init with one of these keys, and verifies `ipfs id` returns a correct looking key.

repo/config/init.go
```diff
@@ -93,15 +93,22 @@ func datastoreConfig() (Datastore, error) {
...
 
 // identityConfig initializes a new identity.
-func identityConfig(out io.Writer, nbits int) (Identity, error) {
+func identityConfig(out io.Writer, nbits, keyType int) (Identity, error) {
 	// TODO guard higher up
 	ident := Identity{}
-	if nbits < 1024 {
-		return ident, errors.New("Bitsize less than 1024 is considered unsafe.")
+
+	switch keyType {
```
vyzo: So what happens here if the user passed an incorrect key type?
I don't see the user option validated anywhere, it's just an int.

vyzo: Basically you need a default handler that errors out with something helpful for the user.

JustinDrake: Incorrect key type options will error out on line 113 with `sk, pk, err := ci.GenerateKeyPair(keyType, nbits)`.

vyzo: It's a little inconsistent with the intent of the switch, and, more importantly, there are other valid key types supported by libp2p-crypto that we don't use yet in ipfs (eg secp256k1 keys). 

So I think it's better to tell the user we are not ready to handle this type of key rather than fallthrough and possibly succeed with surprising results.

richardschneider:
This has laying around for 1.5 years.  Is it ever going to implemented?

Issue #2074, opened 2015-12-15
files cp: add an "overwrite" option


fazo96:
Sometimes `files cp` could be used to overwrite something existing with a newer version.

In that use case, the file has to be removed first with `files rm`. Maybe consider adding an `overwrite` or `force` option to `files cp`?

Kubuxu:
It would also be useful for `ipfs files mv`. Both normal `mv` and `cp` have `-f` flag to override existing content.

Kubuxu:
As a note, `-f` flag is taken by `flush` option, we will have to use `--force`.

te0d:
Currently `ipfs files mv` always overwrites a target file.

CharlieRoseMarie:
Is this issue still relevant? I'd like to contribute.

te0d:
@CharlieRoseMarie I submitted a pull request on this issues a while ago but its been stagnant.

schomatis:
Hey @te0d! Sorry for the delay, I'll take a look at it.

schomatis:
@CharlieRoseMarie If you're still interested in contributing I can help you find another issue you could work at.

CharlieRoseMarie:
@schomatis I would love that, thank you. :)

hsn10:
Any progress on this? It will make my scripts easier.

tx0c:
still need this in 2022

PR #4079, opened 2017-07-18
 cp --force: add force option to files cp command

te0d:
cp --force: add force option to files cp command

The force option was added to the `files cp` command. Specifying force
will overwrite an existing file with the source file.

License: MIT
Signed-off-by: Tom O'Donnell <todonnel91@gmail.com>


te0d:
Fixes:  #2074

core/commands/files/files.go
```diff
@@ -276,6 +288,27 @@ func getNodeFromPath(ctx context.Context, node *core.IpfsNode, p string) (node.N
...
+
+	pdir, ok := parent.(*mfs.Directory)
+	if !ok {
+		err = fmt.Errorf("No such file or directory: %s", dir)
+		return err
+	}
+
+	child, err := pdir.Child(name)
+	if err == nil && child.Type() == mfs.TFile {
+		pdir.Unlink(name)
```
whyrusleeping: Should check the error returned from `Unlink`

core/commands/files/files.go
```diff
@@ -276,6 +288,27 @@ func getNodeFromPath(ctx context.Context, node *core.IpfsNode, p string) (node.N
...
+	}
+
+	pdir, ok := parent.(*mfs.Directory)
+	if !ok {
+		err = fmt.Errorf("No such file or directory: %s", dir)
+		return err
+	}
+
+	child, err := pdir.Child(name)
+	if err == nil && child.Type() == mfs.TFile {
```
whyrusleeping: how should this behave in the context of symlinks?

core/commands/files/files.go
```diff
@@ -276,6 +288,27 @@ func getNodeFromPath(ctx context.Context, node *core.IpfsNode, p string) (node.N
...
+func unlinkNodeIfExists(node *core.IpfsNode, path string) error {
+	dir, name := gopath.Split(path)
+	parent, err := mfs.Lookup(node.FilesRoot, dir)
+	if err != nil {
+		return err
+	}
+
+	pdir, ok := parent.(*mfs.Directory)
+	if !ok {
+		err = fmt.Errorf("No such file or directory: %s", dir)
```
whyrusleeping: Not necessarily true, this could be a file, but not a directory. Probably should have the error message be `"PATH was not a directory"` or something

core/commands/files/files.go
te0d: Ok, wasn't aware of symlinks. I'll check it out.

schomatis:
@te0d So, if you're feeling adventurous and would like to learn some more about how IPFS work, particularly the MFS layer, I think we could leverage the [`Mv`](https://github.com/ipfs/go-ipfs/blob/master/mfs/ops.go#L16) function (the one behind `ipfs files mv`) to adapt it to perform both operations , move and copy, as it would seem to me (at first sight at least) that basically the only difference between the two is that the first one unlinks the source and the second one doesn't (in theory, we would need to think this through).

If not, and you would just like to get this merged and get it over with (which is more than understandable since we made you wait more than an year as of now), please rebase this PR and we can give it the final touches to bring it home.

te0d:
@schomatis Yes, I can look into adapting Mv to perform both operations. Haven't looked at ipfs in a while and a bit busy the upcoming week, but I'll submit something the following wee.

schomatis:
@te0d No problem, take your time. There's no need to submit a PR from the get-go, I would advice you to first analyze the implementation differences between `ipfs files cp` and `ipfs files mv` so we can discuss them here before you commit time to writing the PR. You can ping me with any questions that come up.

core/commands/files/files.go
```diff
@@ -276,6 +288,32 @@ func getNodeFromPath(ctx context.Context, node *core.IpfsNode, p string) (node.N
...
+	// we are only concerned with unlinking an existing file.
+	child, err := pdir.Child(name)
+	if err == nil && child.Type() == mfs.TFile {
+		err := pdir.Unlink(name)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
```
gammazero, line 314:
```suggestion
	pdir, ok := parent.(*mfs.Directory)
	if !ok {
		return fmt.Errorf("not a directory: %s", dir)
	}
	// Attempt to unlink if child is a file, ignore error since
	// we are only concerned with unlinking an existing file.                                                                                                                                                                
	child, err := pdir.Child(name)
	if err != nil {
		return nil // no child file, nothing to unlink
	}

	if child.Type() != mfs.TFile {
		return fmt.Errorf("not a file: %s", path) 
	}

	return pdir.Unlink(name)
```

gammazero (changes requested):

gammazero:
Replaced by #10823

Thank you @te0d. Needed to open a new PR since this one was too far out of date and needed to make changes to make it current.

Issue #2478, opened 2016-03-17
ipfs init nukes everything in the directory


noffle:
Unexpected behaviour: `ipfs init --force` will destroy everything in `$IPFS_PATH`, not just the files IPFS created. I learned this the hard way by setting `IPFS_PATH=/tmp` and then having my whole `/tmp` dir nuked when I init'd again.

noffle:
A good analogy of why this is disasterous is imagining if `git init --force` wiped everything in the current directory.

whyrusleeping:
ipfs init -f has been removed

lbeziaud:
The `-f` option is still used by [bench_cli_ipfs_add](https://github.com/ipfs/go-ipfs/blob/master/test/bench/bench_cli_ipfs_add/main.go#L67). @whyrusleeping

Stebalien:
See https://github.com/ipfs/go-ipfs/pull/4081

PR #4081, opened 2017-07-19
remove `-f` flags from calls to init

Stebalien:
It doesn't mean anything anymore.


Stebalien:
Fixes: https://github.com/ipfs/go-ipfs/issues/2478#issuecomment-316432988

Issue #4008, opened 2017-06-25
Return the size of each file on `ipfs add`


diasdavid:
@whyrusleeping would you be ok with adding one more value on the response of an `ipfs add`? The respective file size.

This would enable us to greatly reduce the bundle size of `ipfs-api/add`, which is what most users need.

whyrusleeping:
@diasdavid would you want something returned for directories and symlinks?

diasdavid:
@whyrusleeping I feel it is useful to have the 'size' for those. Essentially I just grab the cumulative size https://github.com/ipld/js-ipld-dag-pb/blob/master/src/dag.proto.js#L13

whyrusleeping:
@diasdavid alright, and you just want a `"size":12345` field added to the response object? Do you want it to be a number or a string?

diasdavid:
Let's do a string because you [know](http://www.w3devcampus.com/wp-content/uploads/logoAndOther/logo_JavaScript.png)

lgierth:
:+1: on including that, both the `ls` and `object/stat` commands return numbers though. Might be worth it to be consistent?

Kubuxu:
It might be worth migrating away from returning numbers in JSON, to prevent stuff like https://github.com/ipfs/go-ipfs/issues/4008#issuecomment-311483203

lgierth:
Which issue do you mean? 4008 is this one here

Kubuxu:
Sorry https://github.com/ipfs/go-ipfs/issues/3354

whyrusleeping:
@lgierth I think we should try to migrate numbers in our api to be strings. It would be a breaking change, but i think its pretty important. This is probably worth opening its own issue over

imkmf:
Hey all! Can this be closed with #4082 being merged?

PR #4082, opened 2017-07-19
ipfs add: added size to response of `ipfs add` command

te0d:
The `ipfs add` command was modified to include the added
node's size as a string. The size is included in the
dagnode info sent over the output channel.

License: MIT
Signed-off-by: Tom O'Donnell todonnel91@gmail.com

### DIFF
--- a/core/coreunix/add.go
+++ b/core/coreunix/add.go
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"os"
 	gopath "path"
+	"strconv"
 
 	bs "github.com/ipfs/go-ipfs/blocks/blockstore"
 	bstore "github.com/ipfs/go-ipfs/blocks/blockstore"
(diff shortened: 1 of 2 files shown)


te0d:
Fixes: #4008

whyrusleeping:
@te0d thanks! This looks great. I'd like to have a test for this. @Kubuxu can you think of a simple way to test this? `ipfs add --enc=json` doesnt give the output we want here (note: this seems kinda broken).

Kubuxu:
`curl -F file=@filename  "http://localhost:5001/api/v0/add"` should work and returns JSON

te0d:
Would <https://gist.github.com/te0d/fcdfc0d1578fc3bcb7d1fed36c2a260b> meet testing needs? Is the test in the appropriate location (../test/sharness/t0410-api-add.sh)?

whyrusleeping:
@te0d I'd at least like to make sure we verify the value thats returned is correct, but yeah, thats pretty much what i'm hoping for

diasdavid:
wooooot! 🚀

Issue #3923, opened 2017-05-13
PUT existing file with Writeable Gateway not returning expected result.


Tangent128:
#### Version information:
go-ipfs version: 0.4.8-
Repo version: 5
System version: amd64/linux
Golang version: go1.8

#### Type:
Bug

#### Severity:
Medium

#### Description:
When using a writeable HTTP gateway:

* If I PUT to a directory tree for a path that *doesn't* exist in the tree, the gateway returns the expected result: the `ipfs-hash` for an edited directory tree that contains the uploaded file at the specified path.
* **However**, if I PUT to a directory tree for a path that *already* exists in the tree, I instead get back the `ipfs-hash` for the *file* I uploaded- as if I had used the POST interface.
* I would expect, in that case, to receive the `ipfs-hash` of the root of an edited directory tree that contains the uploaded file in place of the path's former content.

whyrusleeping:
@Tangent128 Interesting, good catch. This seems like a bug.

Do you think you could add a test to `test/sharness/t0111-gateway-writeable.sh` that reproduces the issue?

lgierth:
Getting only the hash of the updated root is what's unexpected. It should always return the hash of (or path to) the resource that was put.

Example:
```
PUT /ipfs/QmFoo/a/b HTTP/1.1

HTTP/1.1 201 Created
Location: /ipfs/QmBar/a/b
```

The general expectation around PUT with HTTP is: if you GET the value of the Location header from the PUT response with status 201, it should yield exactly the same data. I'm not sure whether the spec parlance is SHOULD or MUST, actually. With status 202, GETting the Location value should provide progress updates and yield the eventual URL of the resource.

lgierth:
(Also just noting we don't need status 202 in the gateway, so don't let me confuse you with that.)

Tangent128:
I was referring to the `ipfs-hash` header for my example, not the `Location` header.

But I did some testing, and the `Location` header illustrates the problem as well.

After a PUT to a nonexistent path, the `Location` header behaves as you say, and as it should-
> PUT ipfs/QmFoo/a/b HTTP/1.1
>
> HTTP/1.1 201 Created
> Location: /ipfs/QmBar/a/b

where QmBar is the hash of the new file tree root. If you `GET /ipfs/QmBar/a/b`, you'll get back what you put.

But if /a/b exists in the QmFoo tree, the result instead is:

> PUT ipfs/QmFoo/a/b HTTP/1.1
>
[...]

PR #4086, opened 2017-07-24
Gateway: Replace Target Node Data when Exists on POST

te0d:
Previously, if the specified path of a POST request
existed, the newly generated DAG would have a root
of the specified path instead of only modifying the
data of the existing DAG at that path. These changes
modify behavior so that the entire tree structure
of the initially specified root CID is maintained,
and the only change is the Data at the specified path.

License: MIT
Signed-off-by: Tom O'Donnell <todonnell@gmail.com>


te0d:
Fixes: #3923

whyrusleeping:
@Kubuxu @diasdavid @lgierth want to review this? I dont have much insight into the gateway code

lgierth:
Hey @te0d, thanks for bearing with us! The proposal makes sense -- we shouldn't have allowed POST to existing nodes in the first place, but it's already there so the links-preserving mode you propose is the best thing to do.

I'll set a few hours aside for reviewing the code over the weekend -- could you rebase it on master in the meantime? Thanks!

te0d:
@lgierth I believe I've rebased it on master.

Issue #4090, opened 2017-07-26
Circuit Relay Integration


vyzo:
#### Description:
[Circuit Relay](/libp2p/specs/tree/master/relay) provides us with the means to indirectly connect peers who cannot directly connect to each other, either because of NAT or because of protocol incompatibilities (eg: js interop). 
The relay specification has been implemented in Go as a transport in [go-libp2p-circuit](/libp2p/go-libp2p-circuit).

#### Integration approach:
Per discussion with @whyrusleeping:
For initial deployment, the relay functionality should be purely manual: peers would have to manually connect to relays and explicitly open connections to each other by directly using `/p2p-circuit` addresses with `ipfs swarm connect`.
The relay transport should be enabled by default, with hop relay functionality an opt-in feature.
No relay addresses should be announced by the host through the DHT, until we better understand the heuristics for automatically using them

whyrusleeping:
@vyzo should we close this? or use it as a place to track information about further relay integrations?

vyzo:
Up to you. The basic issue has been resolved, but we do need a place to track further integration information, so we can keep it open.

Winterhuman:
@whyrusleeping Should this be closed?

vyzo:
once project flare is fully deployed, yes

PR #4091, opened 2017-07-26
Circuit Relay integration

vyzo:
#4090 Integrates circuit relay as a transport, enabled by default.

Notes:
- Integration is direct, modulo https://github.com/libp2p/go-libp2p/pull/210
- ~Rebased on top of #4094. requires up-to-date go-libp2p. [BLOCKED]~

Adds the following options to swarm configuration options:
- DisableRelay (default: false) -- set to disable circuit relay transport
- EnableRelayHop (default: false) -- set to enable the node to act as a Hop relay.


vyzo:
summning @whyrusleeping @Stebalien

vyzo:
iptb testing scenario:
```
$ iptb start --args --routing=none
Started daemon /home/vyzo/N1C/ProtocolLabs/testbed/0, pid = 21741
Started daemon /home/vyzo/N1C/ProtocolLabs/testbed/1, pid = 21755
Started daemon /home/vyzo/N1C/ProtocolLabs/testbed/2, pid = 21768

$ iptb shell 1
$ ipfs config Swarm.EnableRelayHop
true

$ iptb shell 2
$ ipfs id
{
	"ID": "QmbCdQcF3f2oPGE2Ch7wdmjfGjD227TBWCZNUMxjXqTZJ7",
	"PublicKey": "CAASogEwgZ8wDQYJKoZIhvcNAQEBBQADgY0AMIGJAoGBAMs+WIfPPtXG66PPIasxnzNtas9yz3Eo4ahMB9lvOJv9QR197z68RnM4aNRQsiLv0NufxNVdkJi+7dD8Kj2HrX8hklZgrnAQISmHD/nNhamUyYRR+qYFG5nZgshV0yOMreFL9WwQgVau2W0KO/80PmOVe7evsbx/wy4A9XzPUDnVAgMBAAE=",
	"Addresses": [
		"/ip4/127.0.0.1/tcp/50527/ipfs/QmbCdQcF3f2oPGE2Ch7wdmjfGjD227TBWCZNUMxjXqTZJ7",
		"/ip4/10.0.1.30/tcp/50527/ipfs/QmbCdQcF3f2oPGE2Ch7wdmjfGjD227TBWCZNUMxjXqTZJ7"
	],
	"AgentVersion": "go-ipfs/0.4.11-dev/159fa52c",
	"ProtocolVersion": "ipfs/0.1.0"
}

$ iptb connect 0 1
connecting QmcsnBupwKXW8XsTxLMtvgKUVsocyunv6FydY2GJNsg5Qo -> QmUzUq7Bf6vBkSdvatRAokZLjPh1kjNNpQLiYZmduXpxSi
Addresses:  [/ip4/127.0.0.1/tcp/38266/ipfs/QmUzUq7Bf6vBkSdvatRAokZLjPh1kjNNpQLiYZmduXpxSi /ip4/10.0.1.30/tcp/38266/ipfs/QmUzUq7Bf6vBkSdvatRAokZLjPh1kjNNpQLiYZmduXpxSi]
trying ipfs swarm connect /ip4/127.0.0.1/tcp/38266/ipfs/QmUzUq7Bf6vBkSdvatRAokZLjPh1kjNNpQLiYZmduXpxSi
$ iptb connect 1 2
connecting QmUzUq7Bf6vBkSdvatRAokZLjPh1kjNNpQLiYZmduXpxSi -> QmbCdQcF3f2oPGE2Ch7wdmjfGjD227TBWCZNUMxjXqTZJ7
Addresses:  [/ip4/127.0.0.1/tcp/50527/ipfs/QmbCdQcF3f2oPGE2Ch7wdmjfGjD227TBWCZNUMxjXqTZJ7 /ip4/10.0.1.30/tcp/50527/ipfs/QmbCdQcF3f2oPGE2Ch7wdmjfGjD227TBWCZNUMxjXqTZJ7]
trying ipfs swarm connect /ip4/127.0.0.1/tcp/50527/ipfs/QmbCdQcF3f2oPGE2Ch7wdmjfGjD227TBWCZNUMxjXqTZJ7
connection success!

$ iptb shell 0
$ ipfs swarm peers
/ip4/127.0.0.1/tcp/38266/ipfs/QmUzUq7Bf6vBkSdvatRAokZLjPh1kjNNpQLiYZmduXpxSi
$ ipfs swarm connect /p2p-circuit/ipfs/QmbCdQcF3f2oPGE2Ch7wdmjfGjD227TBWCZNUMxjXqTZJ7
connect QmbCdQcF3f2oPGE2Ch7wdmjfGjD227TBWCZNUMxjXqTZJ7 success
$ ipfs swarm peers
/ip4/127.0.0.1/tcp/38266/ipfs/QmUzUq7Bf6vBkSdvatRAokZLjPh1kjNNpQLiYZmduXpxSi
/ipfs/QmUzUq7Bf6vBkSdvatRAokZLjPh1kjNNpQLiYZmduXpxSi/p2p-circuit/ipfs/QmbCdQcF3f2oPGE2Ch7wdmjfGjD227TBWCZNUMxjXqTZJ7

$ iptb shell 1
$ ipfs swarm peers
/ip4/127.0.0.1/tcp/44524/ipfs/QmcsnBupwKXW8XsTxLMtvgKUVsocyunv6FydY2GJNsg5Qo
/ip4/127.0.0.1/tcp/50527/ipfs/QmbCdQcF3f2oPGE2Ch7wdmjfGjD227TBWCZNUMxjXqTZJ7

$ iptb shell 2
$ ipfs swarm peers
/ip4/127.0.0.1/tcp/40106/ipfs/QmUzUq7Bf6vBkSdvatRAokZLjPh1kjNNpQLiYZmduXpxSi
/ipfs/QmUzUq7Bf6vBkSdvatRAokZLjPh1kjNNpQLiYZmduXpxSi/p2p-circuit/ipfs/QmcsnBupwKXW8XsTxLMtvgKUVsocyunv6FydY2GJNsg5Qo
```

vyzo:
Rebased on #4094.

vyzo:
Added a test/sharness script that implements the iptb scenario described above.

vyzo:
More iptb interaction: verify the connection works:
```
$ iptb shell 0
$ echo "hello world" | ipfs add
added QmT78zSuBmuS4z925WZfrqQ1qHaJ56DQaTfyMUF7F8ff5o QmT78zSuBmuS4z925WZfrqQ1qHaJ56DQaTfyMUF7F8ff5o

$ iptb shell 2
$ ipfs cat QmT78zSuBmuS4z925WZfrqQ1qHaJ56DQaTfyMUF7F8ff5o
hello world
```

I will extend the sharness test to verify the connection.

vyzo:
I think it's ready for review.
pinging @Stebalien @whyrusleeping @lgierth

core/commands/swarm.go
```diff
@@ -128,7 +129,11 @@ var swarmPeersCmd = &cmds.Command{
 
 			buf := new(bytes.Buffer)
 			for _, info := range ci.Peers {
-				fmt.Fprintf(buf, "%s/ipfs/%s", info.Addr, info.Peer)
+				if strings.Contains(info.Addr, "/p2p-circuit/") {
+					fmt.Fprintf(buf, "%s", info.Addr)
+				} else {
+					fmt.Fprintf(buf, "%s/ipfs/%s", info.Addr, info.Peer)
```
whyrusleeping: Whats the purpose of this?

vyzo: The address already contains the peer id in this case.
It's of the form `.../p2p-circuit/ipfs/QmId`, so pasting the peer id again would result in the address formatted as `.../p2p-circuit/ipfs/QmId/ipfs/QmId`

vyzo: do you want me to add a comment to that extent?

whyrusleeping: Hrm... that feels a bit weird. @Kubuxu @lgierth @Stebalien thoughts?

Stebalien: I'd expect `RemoteMultiaddr` to return just the `/ipfs/QmProxy/p2p-circuit` part, not the final `/ipfs/QmId`. As-is, this is inconsistent with TCP/IP addresses. Is there a technical reason not to do it this way?

Stebalien: Also, https://github.com/ipfs/go-ipfs/issues/4104 (but we can fix that later).

vyzo: `/ipfs/QmRelay/p2p-circuit` is not a valid p2p-circuit Multiaddr.

vyzo: Nonetheless, we could change go-libp2p-circuit to report a partial address, with the remote peer id removed from the circuit.

vyzo:
Rebased on master.

core/commands/swarm.go
vyzo: Updated go-libp2p-circuit to not include the remote peer in the multiaddr in https://github.com/libp2p/go-libp2p-circuit/pull/6

vyzo: Removed the kludge, as go-libp2p-circuit now reports partial multiaddr.

vyzo:
Blocked on https://github.com/libp2p/go-libp2p-circuit/pull/6

vyzo:
Upon discussion with @Stebalien on irc we concluded that it is important to keep the invariant that multiaddrs returned by `Conn.RemoteMultiaddr` are _dialiable_.
A partial p2p-circuit address will not be, so we need to keep the ipfs part in the address as was the original design.
Will revert and add a comment to the code to explain why we need to filter that part out.

core/commands/swarm.go
vyzo: So we actually need it: the address reported by Conn.RemoteMultiaddr must be dialable, and the partial address won't be.

I will add a comment here explaining why we need this.

Stebalien: Basically `/ipfs/QmId` serves two purposes:

1. Routing information: An alias for an internet address (stored in the DHT): `/ip4/xxx.xxx.xxx.xxx/tcp/1234`
2. Identity information: The key of the target machine.

In the `RemoteMultiaddr` case, the second `ipfs` address is routing information.

We'd actually run into the same problem something returned `/ipfs/QmId` as the entire `RemoteMultiaddr` (which would actually make sense if you had a lower-level VPN that understood IPFS addresses).

---

Actually, that brings up a good point. Maybe we should just be checking to see if the last address is an `/ipfs/` address and, if it is, assert that the `QmIds` match instead of appending something. Thoughts @vyzo?

vyzo: > Maybe we should just be checking to see if the last address is an /ipfs/ address (i.e., the multiaddr ends in /ipfs/QmId) and, if it is, assert that the QmIds match instead of appending something.

That's perhaps the generic case to handle routing addresses similar to how circuit reports address.
So instead of checking if it is a p2p-circuit address, check if it ends with the `/ipfs/peerId` and if so only append if it doesn't match. 

Sure, we can do that.

vyzo: I have changed the check to look for the `/ipfs/QmPeer` suffix in the remote multiaddr. 
If the suffix is present, then it doesn't duplicate it, otherwise it appends `/ipfs/QmPeer`.

vyzo: Not sure we need a comment any more, it's not p2p-circuit specific and the code is self-explaining.

core/core.go
```diff
@@ -731,8 +739,35 @@ func constructPeerHost(ctx context.Context, id peer.ID, ps pstore.Peerstore, bwr
...
+			for _, addr := range addrs {
+				_, err := addr.ValueForProtocol(circuit.P_CIRCUIT)
+				if err == nil {
+					continue
+				}
+				raddrs = append(raddrs, addr)
+			}
+			return raddrs
+		}
+		hostOpts = append(hostOpts, p2pbhost.AddrsFactory(filterRelayAddr))
```
whyrusleeping: What happens if we have multiple address factories in the hostOpts?

thirdparty/ipfsaddr/ipfsaddr.go
```diff
@@ -106,6 +107,13 @@ func ParseMultiaddr(m ma.Multiaddr) (a IPFSAddr, err error) {
 
 func Transport(iaddr IPFSAddr) (maddr ma.Multiaddr) {
 	maddr = iaddr.Multiaddr()
+
+	// /ipfs/QmId is part of the transport address for p2p-circuit
+	_, err := maddr.ValueForProtocol(circuit.P_CIRCUIT)
```
whyrusleeping: Would prefer if we at least put a TODO here to clean this up. We need to rethink (later, not now) how addresses get composed and consumed so that we don't have to have this special case here.

core/core.go
vyzo: We would need to compose them.

thirdparty/ipfsaddr/ipfsaddr.go
vyzo: Sure, will add a TODO.

core/core.go
vyzo: Do you want me to add scaffolding for the composition?

whyrusleeping: Yeah, Since we're going to be using the address factory pretty soon anyways for limiting which addresses we announce: https://github.com/ipfs/go-ipfs/pull/3948

vyzo: I will need to add the `AddrsFactory` option to `ConstructPeerHostOpts` in order to scaffold the composition properly.

Then we have two options for the composition: filter first or announce factory first. It makes some sense to use the filter first, the announce factory might actually want to announce some relay addresses in the future. On the other hand it also makes sense for the announce factory to see the p2p-circuit relay address, so that it knows that relay is activated.

vyzo: On the other hand, the address factory can also tell whether relay is in use from the swarm options already, so the latter argument is not all that compelling. 

So the filter should be applied before the announce address factory.

vyzo: Added the scaffolding for address factory composition in https://github.com/ipfs/go-ipfs/pull/4091/commits/231f8e46b2ac26a2b344c958bbe2765c0ea41961:
- added an AddrsFactory option to ConstructPeerHostOpts
- composed the specified address factory with the relay filter when present

thirdparty/ipfsaddr/ipfsaddr.go
vyzo: Added TODO comment in https://github.com/ipfs/go-ipfs/pull/4091/commits/a11f632244df268fbbbdbbb1cec6c2e2a50e7708

Stebalien: I wonder if we could just stop distinguishing between the transport part and the application part. Multiaddrs don't really have a clear transport/application layer separation (they can have transport -> application -> transport -> transport -> application, etc...).

Instead, we could either:

a. Have dialers recursively dial (dial the parts they understand and then recursively call the "main" dialer to dial the rest).
b. Dial like we resolve through IPLD objects and return an intermediate transport + the rest of the undialed path.

thirdparty/ipfsaddr/ipfsaddr.go
```diff
@@ -106,6 +107,16 @@ func ParseMultiaddr(m ma.Multiaddr) (a IPFSAddr, err error) {
 
 func Transport(iaddr IPFSAddr) (maddr ma.Multiaddr) {
 	maddr = iaddr.Multiaddr()
+
+	// /ipfs/QmId is part of the transport address for p2p-circuit
+	// TODO clean up the special case
+	// we need a consistent way of composing and consumig multiaddrs
+	// so that we don't have to do this
+	_, err := maddr.ValueForProtocol(circuit.P_CIRCUIT)
```
Kubuxu: Please create issue for this TODO. TODOs in code get lost in time and space.

vyzo: Sure, will do once we merge.

Kubuxu:
Let's merge https://github.com/ipfs/go-ipfs/pull/3948 before this.

vyzo:
Fine by me, it will be easier for me to merge the address factory conflict.

vyzo:
Blocked on #3948 -- will have to resolve the address factory conflict.

vyzo:
rebased on master and resolved conflict with #3948

Stebalien:
Test failure in `test/sharness/t0060-daemon.sh` looks legitimate:

```
expecting success:
  STARTFILE="ipfs cat /ipfs/$HASH_WELCOME_DOCS/readme" &&
  echo "Initializing daemon..." >expected_daemon &&
  sed "s/^/Swarm listening on /" local_addrs >>expected_daemon &&
  sed "s/^/Swarm announcing /" local_addrs >>expected_daemon &&
  echo "API server listening on /ip4/127.0.0.1/tcp/39358" >>expected_daemon &&
  echo "Gateway (readonly) server listening on /ip4/127.0.0.1/tcp/37737" >>expected_daemon &&
  echo "Daemon is ready" >>expected_daemon &&
  test_cmp expected_daemon actual_daemon

> diff -u expected_daemon actual_daemon
--- expected_daemon	2017-07-31 19:06:33.423316945 +0000
+++ actual_daemon	2017-07-31 19:06:33.127314418 +0000
@@ -1,5 +1,6 @@
 Initializing daemon...
 Swarm listening on /ip4/127.0.0.1/tcp/40291
+Swarm listening on /p2p-circuit/ipfs/QmWRfU35RsUqNYmiu4osu8qMJJVvdajrREGkov12p2ooe1
 Swarm announcing /ip4/127.0.0.1/tcp/40291
 API server listening on /ip4/127.0.0.1/tcp/39358
 Gateway (readonly) server listening on /ip4/127.0.0.1/tcp/37737

not ok 13 - ipfs daemon output looks good
```

vyzo:
yes, it's an issue with the `InterfaceListeningAddrs` reporting the p2p-circuit address, which is not reported by `ipfs swarm addrs local` due to filtering.

vyzo:
So this brings up an interesting question, on whether the p2p-circuit address should be reported by `ipfs swarm addrs local`.

vyzo:
Added an `ipfs swarm addrs listen` subcommand so that we can list listen addresses from the swarm -- without the filter factory. 
This fixes the t0060-daemon test, as we know have a command to find the listening addresses we want to test against.

vyzo:
blocked on swarm dependency update:
- https://github.com/ipfs/go-ipfs/pull/4111
- https://github.com/libp2p/go-libp2p-circuit/pull/8

vyzo:
Rebased on master and updated go-libp2p-circuit dependency.

core/commands/swarm.go
```diff
@@ -128,7 +129,11 @@ var swarmPeersCmd = &cmds.Command{
 
 			buf := new(bytes.Buffer)
 			for _, info := range ci.Peers {
-				fmt.Fprintf(buf, "%s/ipfs/%s", info.Addr, info.Peer)
+				if strings.HasSuffix(info.Addr, fmt.Sprintf("/ipfs/%s", info.Peer)) {
+					fmt.Fprintf(buf, "%s", info.Addr)
+				} else {
+					fmt.Fprintf(buf, "%s/ipfs/%s", info.Addr, info.Peer)
```
lgierth: Hardcoded `/ipfs` addrs unneccessarily make it harder to migrate to `/p2p` in the not-too-distant future. Better to at least get the scheme name via multiaddr, and ideally also use multiaddr methods to modify addresses.

I think I also spotted this in go-libp2p-circuit.

vyzo: I can certainly change it to use the scheme name from multiaddr. As for multiaddr method manipulation, these are strings here so that's inconvenient.

And yes indeed, I am guilty of using hardcoded ipfs strings in go-libp2p-circuit. I will look into fixing it there as well, although it's not critical for this PR.

lgierth: >  although it's not critical for this PR

Agreed :) As for the scheme names in go-libp2p-circuit, those updates can be bubbled up at the next occasion sometime soon.

vyzo: Modified the offending code to use the name from the mutladdr protocol in https://github.com/ipfs/go-ipfs/pull/4091/commits/0e5e00391e453f622e9df19301579e4b571ba945

vyzo: And a minor simplification in https://github.com/ipfs/go-ipfs/pull/4091/commits/b66c942e11cc212bacf1d653efc875248a7cd6b8

vyzo:
jenkins failure unrelated:
```
--- FAIL: TestFetchGraph (0.01s)
	merkledag_test.go:248: merkledag: not found
```

will force push to trigger a new build.

vyzo:
We will have to update the go-libp2p-circuit dependency to 1.1.3 when https://github.com/ipfs/go-ipfs/pull/4127 gets merged.

core/core.go
```diff
@@ -787,15 +793,56 @@ func constructPeerHost(ctx context.Context, id peer.ID, ps pstore.Peerstore, bwr
 	if !opts.DisableNatPortMap {
 		hostOpts = append(hostOpts, p2pbhost.NATPortMap)
 	}
-	if opts.AddrsFactory != nil {
-		hostOpts = append(hostOpts, opts.AddrsFactory)
+
+	addrsFactory := opts.AddrsFactory
+	if !opts.DisableRelay {
+		filterRelayAddr := func(addrs []ma.Multiaddr) []ma.Multiaddr {
```
whyrusleeping: can we pull this function out? feels cleaner to define it as a proper function

docs/config.md
```diff
@@ -221,5 +221,12 @@ improvement, as well as a reduction in memory usage.
 - `DisableNatPortMap`
 Disable NAT discovery.
 
+- `DisableRelay`
+Disables the p2p-circuit relay transport.
```
Stebalien: Will this disable relays entirely or just disable listening on relayed addresses? There are three cases:

1. Listen using relays.
2. Act as a relay.
3. Connect to nodes through relays.

It would be nice to be able to enable/disable these independently. Can we currently do that?

vyzo: Setting to true will disable the relay transport entirely.

The `DislableRelay` controls option 1 and 3, so by default we listen for relay connections and we can connect to peers through relay.
In order to act as a hop relay, you need to `EnableRelayHop`, in addition to having the relay transport enabled.

I don't think it makes sense to have more fine-grained independent configuration, as we can't really separate listening from dial (and do we really want to do that?) and we already have an option to control hop relaying.

core/core.go
vyzo: sure, will do.

vyzo:
Rebasing on master and fixing the conflict.

docs/config.md
Stebalien: As discussed on IRC, I misunderstood what these toggles did. DisableRelay disables relaying entirely and we don't yet support advertising relayed addresses; when implemented, that will probably have an additional toggle.

vyzo: Yes, that's the plan.

vyzo:
@whyrusleeping this is ready to merge.

whyrusleeping:
Merged! ipfs has relay support!! Thank you @vyzo!

whyrusleeping:
@vyzo do you think you could write up a small doc on how to use the relay feature? That would be immensely helpful to have

vyzo:
Sure!  Where should it live though?

Kubuxu:
`docs/experimental-features.md`

vyzo:
#4148 adds documentation in docs/experimental-features.md

PR #4004, opened 2017-06-22
Add pin option for ipfs dag put

magik6k:
This at least partially addresses #3544

PR #4095, opened 2017-07-27
object put --pin option

### LINKED ISSUE #4092: Add pinning support for object.put
#### Version information:

```
go-ipfs version: 0.4.6-
Repo version: 5
System version: amd64/darwin
Golang version: go1.8
```

#### Description:

Followup from https://github.com/ipfs/js-ipfs-api/issues/532

When calling object.put, it would be great if I could specify an option that the object should also be pinned. Currently it seems I have to do two calls. But I think it is pretty common that you want objects you created to be pinned on your instance of IPFS.

magik6k:
Closes #4092

Based off #4004. The question now is what to do about `object patch *` commands.

### DIFF
--- a/core/commands/object/object.go
+++ b/core/commands/object/object.go
@@ -16,6 +16,7 @@ import (
 	core "github.com/ipfs/go-ipfs/core"
 	dag "github.com/ipfs/go-ipfs/merkledag"
 	path "github.com/ipfs/go-ipfs/path"
+	pin "github.com/ipfs/go-ipfs/pin"
 	ft "github.com/ipfs/go-ipfs/unixfs"
 
 	cid "gx/ipfs/QmTprEaAA2A9bst5XH7exuyi5KzNMK3SEDNN8rBDnKWcUS/go-cid"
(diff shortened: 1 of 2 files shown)


whyrusleeping:
@magik6k while you're at it, could you do the same for `ipfs dag put` ?

magik6k:
`dag put` has --pin, I even linked to that PR here(https://github.com/ipfs/go-ipfs/pull/4004)

whyrusleeping:
*facepalm* 
I'm sorry, carry on. Will review soon.

core/commands/object/object.go
```diff
@@ -381,6 +383,16 @@ And then run:
...
+		if err != nil {
+			res.SetError(err, cmds.ErrNormal)
+			return
+		}
+
+		if dopin {
+			defer n.Blockstore.PinLock().Unlock()
+		}
+
 		output, err := objectPut(n, input, inputenc, datafieldenc)
```
whyrusleeping: could we make `objectPut` just return a cid? It returns an `Object` type for some reason that contains a stringed cid and an array of links for some reason. It doesnt ever use the links, seems like we could make this a little cleaner

whyrusleeping:
@magik6k one comment about a 'nice to have', then would be good to rebase on latest master to try for greener CI

magik6k:
@whyrusleeping done

core/commands/object/object.go
```diff
@@ -391,15 +403,24 @@ And then run:
...
+			}
+		}
+
+		res.SetOutput(objectCid)
 	},
 	Marshalers: cmds.MarshalerMap{
 		cmds.Text: func(res cmds.Response) (io.Reader, error) {
-			object := res.Output().(*Object)
-			return strings.NewReader("added " + object.Hash + "\n"), nil
+			object := res.Output().(*cid.Cid)
```
whyrusleeping: We should keep this the same, and wrap the cid in an `Object` type when calling `SetOutput` above, otherwise we're changing the API. this is the last nitpick, i promise :)

magik6k:
@whyrusleeping done

mitar:
Thanks! Awesome!

PR #4099, opened 2017-07-28
Clarify error message on merkeldag not found

timthelion:
Fixes https://github.com/ipfs/website/issues/85

License: MIT
Signed-off-by: Timothy Hobbs <timothyhobbs@seznam.cz>

### DIFF
--- a/merkledag/merkledag.go
+++ b/merkledag/merkledag.go
@@ -24,7 +24,7 @@ func init() {
 	node.Register(cid.DagCBOR, ipldcbor.DecodeBlock)
 }
 
-var ErrNotFound = fmt.Errorf("merkledag: not found")
+var ErrNotFound = fmt.Errorf("merkledag: not found. Is the ipfs daemon running? The daemon must be running in order to get non-local merkledags. Launch it with \"ipfs daemon\".")
 
 // DAGService is an IPFS Merkle DAG service.
 type DAGService interface {


whyrusleeping:
This is a tough one, because `ErrNotFound` could potentially be returned when an online search fails too.

kevina:
I would be inclined to leave the error message alone and maybe add more helpful information upstream.

One thing I would do is make ErrNotFound a class so it can include the Cid of the object not found.  That would break a lot of code though as we would need to either add a IsNotFound helper function or (less preferably) try and cast to the ErrNotFound class.

whyrusleeping:
@kevina I like that idea, it mirrors closely the standard libs `os.IsNotFound(err)` paradigm.

kevina:
@whyrusleeping that was the idea (mirror the `os.IsNotFound(err)` paradigm)

timthelion:
Hm, I'm not familiar with golang error classes and am not sure that it 
is worth my time to learn them. Should I close this?

kevina:
> Hm, I'm not familiar with golang error classes and am not sure that it
> is worth my time to learn them.

Go lang does not really have error classes.  What is needed is something like this:
```
type ErrNotFound struct {
  Cid *cid.Cid
}
func (e ErrNotFound) Error() string {return fmt.Sprintf("merkledag: not found: %s", e.Cid)}
func IsNotFound (e error) bool {_, ok := e.(ErrNotFound); return ok}
```

> Should I close this?

It is up to you.  I will be happy to implement the above if you don't have the time.

kevina:
Also note that the ErrNotFound to `go-ipld-format` which means fixing this error message no requires a change to that package.

hsanjuan:
Thanks @timthelion. I think we can close this one as the approach was not the best, but I'll follow up other other approaches that were tried and see if the issue can be picked up again.

PR #4103, opened 2017-07-28
sharness: use test profile

### LINKED ISSUE #4098: Disable mdns for most sharness tests
We should disable mdns for sharness tests by default and re-enable it only when necessary. Otherwise, our test nodes may end up talking to non test nodes.

Stebalien:
This:

* Disables mdns and fixes #4098 (iptb nodes already have mdns disabled by default).
* Only listens on localhost (although iptb nodes still listen on all).

### DIFF
--- a/test/sharness/lib/test-lib.sh
+++ b/test/sharness/lib/test-lib.sh
@@ -141,19 +141,13 @@ test_init_ipfs() {
 
 	test_expect_success "ipfs init succeeds" '
 		export IPFS_PATH="$(pwd)/.ipfs" &&
-		ipfs init -b=1024 > /dev/null
+		ipfs init --profile=test -b=1024 > /dev/null
 	'
 
-	test_expect_success "prepare config -- mounting and bootstrap rm" '
+	test_expect_success "prepare config -- mounting" '
 		mkdir mountdir ipfs ipns &&
 		test_config_set Mounts.IPFS "$(pwd)/ipfs" &&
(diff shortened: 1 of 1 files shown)


test/sharness/lib/test-lib.sh
```diff
@@ -141,18 +141,13 @@ test_init_ipfs() {
...
 	test_expect_success "prepare config -- mounting and bootstrap rm" '
 		mkdir mountdir ipfs ipns &&
 		test_config_set Mounts.IPFS "$(pwd)/ipfs" &&
 		test_config_set Mounts.IPNS "$(pwd)/ipns" &&
-		test_config_set Addresses.API "/ip4/127.0.0.1/tcp/0" &&
-		test_config_set Addresses.Gateway "/ip4/0.0.0.0/tcp/0" &&
-		test_config_set --json Addresses.Swarm "[
-  \"/ip4/0.0.0.0/tcp/0\"
-]" &&
 		ipfs bootstrap rm --all ||
```
magik6k: This is not needed too

lgierth:
Should Mounts.IPFS and Mounts.IPNS be part of the test profile too?

Otherwise LGTM :+1:

Stebalien:
> Should Mounts.IPFS and Mounts.IPNS be part of the test profile too?

That's non-trivial as we want the mounts to be under the current directory, not at some absolute location.

lgierth:
Ah yep that makes sense :)

Issue #4062, opened 2017-07-12
TestFetchGraph failing randomly


whyrusleeping:
very occasionally we get a failure in TestFetchGraph: https://ci.ipfs.team/blue/organizations/jenkins/go-ipfs/detail/PR-4047/57/pipeline/

```
--- FAIL: TestFetchGraph (0.01s)
	merkledag_test.go:248: merkledag: not found
FAIL
```

This is likely a bug in FetchGraph or EnumerateChildren, definitely needs looking into.

Kubuxu:
Repeated in https://travis-ci.org/ipfs/go-ipfs/jobs/262757884

Stebalien:
This appears to be happening more often...

Stebalien:
So, this is definitely a bug, not just a racy test. `FetchGraph` succeeds and then enumerating the graph fails. Possible locations:

1. `offline_ds.GetLinks` or `GetLinksDirect`. The test case uses this, `FetchGraph` uses `GetLinksDirect`. This seems unlikely
2. `EnumerateChildrenAsync` or `EnumerateChildren`. I'm betting it's `EnumerateChildrenAsync` (it's probably miscounting and not waiting for some subgraph to be enumerated.

Stebalien:
So much for it being `EnumerateChildrenAsync`'s fault. I've tried replacing `EnumerateChildrenAsync` (in `FetchGraph` with `EnumerateChildren` and it still fails occasionally.

They could both be buggy...

Stebalien:
Not `GetLinks` versus `GetLinksDirect` (not that surprising).

Also, I'm pretty sure the bug is *not* in `EnumerateChildren` as it's a simple recursive function.

Stebalien:
So... bitswap is notifying bitswap sessions about blocks *before* adding them to the blockstore. We *really* need to clean up this logic.

(actually, bitswap notifies the session about the block *twice*, once before adding it to the blockstore and once after because bitswap is special).

Stebalien:
As a matter of fact, we had a race condition where we could end up returning a block to the user while bitswap was closing and never record the block in the blockstore.

Stebalien:
As a matter of fact, we had a race condition where we could end up returning a block to the user while bitswap was closing and never record the block in the blockstore.

PR #4109, opened 2017-07-31
bitswap_test: make racy test less racy

### LINKED ISSUE #4108: Random test failure in bitswap
#### Version information:

> go-ipfs version: 0.4.11-dev-06c567d17
Repo version: 5
System version: amd64/linux
Golang version: go1.8.3

#### Type:

Bug

#### Severity:

Medium

#### Description:

```
ERROR   bitswap: nil cid in GetBlock get.go:18

--- FAIL: TestBasicBitswap (0.03s)

	bitswap_test.go:319: Test a one node trying to get one block from another

	bitswap_test.go:338: should have no items in other peers wantlist

FAIL

FAIL	github.com/ipfs/go-ipfs/exchange/bitswap	10.387s
```

Stebalien:
fixes #4108


Stebalien:
(in an effort to make CI reliable...)

Stebalien:
As usual, the failures are unrelated... (99 failures of tests on CI, 99 failures of tests, take one down, pass it around, 9001 failures of tests on CI).

Stebalien:
> We should also object tests that depend on flat timing in future.

YES! Also, a rule that all CI failures must be investigated/reported, even if they look spurious. Always having to re-run CI a few times to get a pass is time consuming (and dangerous).

Stebalien:
Speaking of which... the travis test failure was #4062.

whyrusleeping:
we should add a more generic `waitFor(ctx, check)` method so we can do things like this more frequently.

Stebalien:
Need to:
1. Copy the modifications made to the vendored testutils to the actual library.
2. Unvendor testutils.
3. Use WaitFor from the unvendored testutils.

whyrusleeping:
@Stebalien update here?

Stebalien:
No change. I'll do this now... (it's just really tedious).

Stebalien:
Blocked on https://github.com/ipfs/go-ipfs/pull/4091 apparently (I need to update go-libp2p which now pulls in go-libp2p-circuit).

vyzo:
Note that #4091 is still in circuit-relay-1.1.2, it will need to update to circuit-relay-1.1.3 when libp2p is updated.

core/core_test.go
```diff
@@ -38,7 +38,7 @@ func TestInitialization(t *testing.T) {
 	for i, c := range good {
 		r := &repo.Mock{
 			C: *c,
-			D: testutil.ThreadSafeCloserMapDatastore(),
+			D: ds2.ThreadSafeCloserMapDatastore(),
```
Stebalien: Note: I moved this function from the vendored testutil library to datastore2.

Kubuxu:
Thank you for this.

Kubuxu:
Thank you a lot for this fix.

PR #4113, opened 2017-08-01
Reprovider strategies

magik6k:
Reprovider strategies allow to specify which keys to provide to the content routing.
Supported strategies are:  `all` - all blocks,  `pinned` - all pinned blocks, `roots` - root/directly pined blocks.
Current reprovider is slow(~0.2keys/s with my setup), so constraining key set should speed the process up, announcing only pin roots should still work quite well in most cases.

TODO: 
* [x] Tests
* [x] Docs (experimental?)
* [x] Reduce 'pinned' strategy i/o overhead (it's really bad with larger repos now)
* [ ] Speed up reprovider (probably out of scope of this PR)

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -277,7 +277,7 @@ func (n *IpfsNode) startLateOnlineServices(ctx context.Context) error {
 	default:
 		return fmt.Errorf("unknown reprovider strtaegy '%s'", cfg.Reprovider.Strategy)
 	}
-	n.Reprovider = rp.NewReprovider(n.Routing, keyProvider)
+	n.Reprovider = rp.NewReprovider(ctx, n.Routing, keyProvider)
 
 	if cfg.Reprovider.Interval != "0" {
 		interval := kReprovideFrequency
@@ -290,7 +290,7 @@ func (n *IpfsNode) startLateOnlineServices(ctx context.Context) error {
 			interval = dur
 		}
 
-		go n.Reprovider.ProvideEvery(ctx, interval)
+		go n.Reprovider.ProvideEvery(interval)
 	}
 
 	return nil
--- a/core/core.go
+++ b/core/core.go
@@ -263,7 +263,7 @@ func (n *IpfsNode) startLateOnlineServices(ctx context.Context) error {
 		return err
 	}
 
-	var keyProvider func(context.Context) (<-chan *cid.Cid, error)
+	var keyProvider rp.KeyChanFunc
 
 	switch cfg.Reprovider.Strategy {
 	case "all":
@@ -275,7 +275,7 @@ func (n *IpfsNode) startLateOnlineServices(ctx context.Context) error {
 	case "pinned":
 		keyProvider = rp.NewPinnedProvider(n.Pinning, n.DAG, false)
 	default:
-		return fmt.Errorf("unknown reprovider strtaegy '%s'", cfg.Reprovider.Strategy)
+		return fmt.Errorf("unknown reprovider strategy '%s'", cfg.Reprovider.Strategy)
 	}
 	n.Reprovider = rp.NewReprovider(ctx, n.Routing, keyProvider)
 
--- a/core/core.go
+++ b/core/core.go
@@ -237,22 +237,6 @@ func (n *IpfsNode) startOnlineServices(ctx context.Context, routingOption Routin
 		return err
 	}
 
-	n.Reprovider = rp.NewReprovider(n.Routing, n.Blockstore)
-
-	if cfg.Reprovider.Interval != "0" {
-		interval := kReprovideFrequency
-		if cfg.Reprovider.Interval != "" {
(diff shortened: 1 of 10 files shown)


magik6k:
Low patch coverage is likely caused by not collecting coverage stats from iptb tests. For example `rp.Trigger` in https://codecov.io/gh/ipfs/go-ipfs/compare/542d63461ef74a44d586ad525b1954b694b18a56...92df0d60ceac2bf2298480bf3632a8786551806f/src/exchange/reprovide/reprovide.go?before=exchange/reprovide/reprovide.go#L93 is not marked as covered, when it should be (called by `ipfs bitswap reprovide`)

cc @Kubuxu

core/core.go
```diff
@@ -273,6 +257,45 @@ func (n *IpfsNode) startOnlineServices(ctx context.Context, routingOption Routin
...
+	case "all":
+		fallthrough
+	case "":
+		keyProvider = rp.NewBlockstoreProvider(n.Blockstore)
+	case "roots":
+		keyProvider = rp.NewPinnedProvider(n.Pinning, n.DAG, true)
+	case "pinned":
+		keyProvider = rp.NewPinnedProvider(n.Pinning, n.DAG, false)
+	default:
+		return fmt.Errorf("unknown reprovider strtaegy '%s'", cfg.Reprovider.Strategy)
```
whyrusleeping: typo 'strategy'

docs/config.md
```diff
@@ -15,7 +15,7 @@ a running daemon do not read the config file at runtime.
 - [`Identity`](#identity)
 - [`Ipns`](#ipns)
 - [`Mounts`](#mounts)
-- [`ReproviderInterval`](#reproviderinterval)
+- [`Reproviderl`](#reprovider)
```
whyrusleeping: typo?

exchange/reprovide/reprovide.go
```diff
@@ -5,56 +5,79 @@ import (
...
+		case done = <-rp.trigger:
 		case <-after:
-			err := rp.Reprovide(ctx)
-			if err != nil {
-				log.Debug(err)
-			}
-			after = time.After(tick)
 		}
+
+		unmute := rp.muteTrigger()
```
whyrusleeping: could I get a comment or two on the muting? I think I get it, but would be great to clearly lay out whats going on

exchange/reprovide/reprovide.go
```diff
@@ -5,56 +5,79 @@ import (
...
-	blocks "github.com/ipfs/go-ipfs/blocks/blockstore"
 	backoff "gx/ipfs/QmPJUtEJsm5YLUWhF6imvyCH8KZXRJa9Wup7FDMwTy5Ufz/backoff"
 	routing "gx/ipfs/QmPjTrrSfE6TzLv6ya6VWhGcCgPrUAdcgrDcQyRDX2VyW1/go-libp2p-routing"
 	logging "gx/ipfs/QmSpJByNKFX1sCsHBEp3R73FL4NF6FnQTEGyNAXHm2GS52/go-log"
+	cid "gx/ipfs/QmTprEaAA2A9bst5XH7exuyi5KzNMK3SEDNN8rBDnKWcUS/go-cid"
 )
 
 var log = logging.Logger("reprovider")
 
+type keyChanFunc func(context.Context) (<-chan *cid.Cid, error)
```
whyrusleeping: lets export this type, that way you dont have to redefine it inline up above.

exchange/reprovide/providers.go
```diff
@@ -0,0 +1,99 @@
...
+			set.add(key)
+		}
+
+		for _, key := range pinning.RecursiveKeys() {
+			set.add(key)
+
+			if !onlyRoots {
+				err := merkledag.EnumerateChildren(ctx, dag.GetLinks, key, set.add)
+				if err != nil {
+					return //TODO: propagate to chan / log?
```
whyrusleeping: lets definitely not ignore this error, at least log.Error it (though propogating it upwards would be optimal)

exchange/reprovide/providers.go
```diff
@@ -0,0 +1,99 @@
...
+		}
+
+		close(set.new)
+	}()
+
+	return set, nil
+}
+
+type streamingSet struct {
+	set map[string]struct{}
```
whyrusleeping: you could reuse the cid.Set for this, it has a visit method that does exactly what you want down in 'add'

magik6k: The reason this is done like this is to reduce i/o hit on `merkledag.EnumerateChildren` that happens when reprovider starts on large repo (with cid.Set I got 100% io use for about 5min with my repo)

whyrusleeping: right, what i'm saying is that instead of having `set` here be a `map[string]struct{}`, it could instead be a `*cid.Set`

magik6k: ah, fair point, done

Issue #2783, opened 2016-06-01
Mans: Keep text within 78-char ruler


RichardLitt:
I think we should keep the text for the descriptions wrapped in under the 78-char ruler. Here is an example where that doesn't happen, on my terminal: 

![screen shot 2016-06-01 at 4 48 34 pm](https://cloud.githubusercontent.com/assets/910753/15716032/b43fe166-2818-11e6-86d6-bea090f92557.png)

Even if this is a ridiculous cultural remnant of [IBM's punchcard days](https://programmers.stackexchange.com/questions/148677/why-is-80-characters-the-standard-limit-for-code-width), it may still be worth doing to conform to the industry standard.

I've gone through and manually PRed the descriptions to be wrapped (See #2782, #2780, #2779), but there are still the cases of the options and some of the examples. Should we manually wrap options? And Synopses? If so, how?

whyrusleeping:
We should have a test that we can manually run to verify all our helptext fits within this limit. I think its a good idea to try and keep the 78 (but really 80) character limit

whyrusleeping:
We've got a test for this now. Next up is trying to make it pass.

To run the test:
```
cd test/sharness
make deps
./t0010-basic-commands.sh
```

whyrusleeping:
I'm assuming that since the test is merged to master, and things are passing, that this has been resolved

Kubuxu:
Tests are not passing, they are marked as failed.

whyrusleeping:
If any contributor is looking for something easy to get started with this, one would be great

devedge:
I'm interested in taking this, are there any specific style guidelines I should adhere to when wrapping the text?

magik6k:
Done in https://github.com/ipfs/go-ipfs/pull/4121

PR #4121, opened 2017-08-05
wrap-desc: Wrap help descriptions under 80 chars

devedge:
Make all help command descriptions wrap under 80 characters, as per issue #2783

License: MIT
Signed-off-by: Richard Pajerski <devedge@outlook.com>

### DIFF
--- a/cmd/ipfs/daemon.go
+++ b/cmd/ipfs/daemon.go
@@ -67,18 +67,18 @@ The daemon will start listening on ports on the network, which are
 documented in (and can be modified through) 'ipfs config Addresses'.
 For example, to change the 'Gateway' port:
 
-    ipfs config Addresses.Gateway /ip4/127.0.0.1/tcp/8082
+  ipfs config Addresses.Gateway /ip4/127.0.0.1/tcp/8082
 
 The API address can be changed the same way:
 
-   ipfs config Addresses.API /ip4/127.0.0.1/tcp/5002
+  ipfs config Addresses.API /ip4/127.0.0.1/tcp/5002
 
 Make sure to restart the daemon after changing addresses.
 
 By default, the gateway is only accessible locally. To expose it to
 other computers in the network, use 0.0.0.0 as the ip address:
 
-   ipfs config Addresses.Gateway /ip4/0.0.0.0/tcp/8080
+  ipfs config Addresses.Gateway /ip4/0.0.0.0/tcp/8080
 
 Be careful if you expose the API. It is a security risk, as anyone could
 control your node remotely. If you need to control the node remotely,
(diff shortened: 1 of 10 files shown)


cmd/ipfs/daemon.go
```diff
@@ -92,7 +92,8 @@ do this by setting headers on the API.HTTPHeaders and Gateway.HTTPHeaders
 keys:
 
 	ipfs config --json API.HTTPHeaders.X-Special-Header '["so special :)"]'
-	ipfs config --json Gateway.HTTPHeaders.X-Special-Header '["so special :)"]'
+	ipfs config --json Gateway.HTTPHeaders.X-Special-Header '["so special
+	:)"]'
```
kevina: It probably not a good idea to wrap here.

I personally would remove the tab and replace it with two (maybe up to four) spaces to help solve the problem to keep it all on a single line.

As an aside I would not use tabs in help text as the width of tab characters is not fixed.

But I may not speak for others.  @whyrusleeping thoughts.

core/commands/swarm.go
```diff
@@ -369,7 +370,8 @@ var swarmDisconnectCmd = &cmds.Command{
 'ipfs swarm disconnect' closes a connection to a peer address. The address
 format is an IPFS multiaddr:
 
-ipfs swarm disconnect /ip4/104.131.131.82/tcp/4001/ipfs/QmaCpDMGvV2BGHeYERUEnRQAwe3N8SzbUtfsmvsqQLuvuJ
+ipfs swarm disconnect /ip4/104.131.131.82/tcp/4001/ipfs/QmaCpDMGvV2BGHeYERUEnR
+QAwe3N8SzbUtfsmvsqQLuvuJ
```
kevina: Do not split a hash like this.  put it on its on line, if it still too long then I would leave it.

cmd/ipfs/daemon.go
devedge: No problem, there were a few places like this throughout the docs I can revert back.

whyrusleeping:
Yeah, agreed on using spaces instead of tabs, ensures that the output looks the same for everyone.

core/commands/publish.go
```diff
@@ -49,7 +50,8 @@ Publish an <ipfs-path> with another name, added by an 'ipfs key' command:
   > ipfs name publish --key=mykey /ipfs/QmatmE9msSfkKxoffpHwNLNKgwZG8eT9Bud6YoPab52vpy
   Published to QmbCMUZw6JFeZ7Wp9jkzbye3Fzp2GGcPgC3nmeUjfVF87n: /ipfs/QmatmE9msSfkKxoffpHwNLNKgwZG8eT9Bud6YoPab52vpy
 
-Alternatively, publish an <ipfs-path> using a valid PeerID(as listed by 'ipfs key list -l'):
+Alternatively, publish an <ipfs-path> using a valid PeerID(as listed by 'ipfs
+key list -l'):
```
kevina: Change this two:
```
Alternatively, publish an <ipfs-path> using a valid PeerID (as listed by
'ipfs key list -l'):
```
To avoid splitting the command string (and also add a space after "PeerID")

whyrusleeping:
I think two spaces is fine. @devedge if you can get the last comment from @kevina addressed then i think this will be ready to merge

devedge:
Alright, I'll have it up by tonight

whyrusleeping:
will merge post 0.4.11 release :)

Thanks @devedge

PR #4131, opened 2017-08-08
Add event system to blockstore

Kubuxu:
This is a small stub of something I hope we shift our blockstore in future.
With additional events like EventGet, EventHas we would be able to:
1. have add-on functionality in blockstore (imagine Ethereum blockstore plugin)
2. be much more flexible than what we do now (blockstore wrapping)
3. it will be easier to expand in future

This stub is to allow me to start working on tri-colour GC for go-ipfs which would replace the stop world GC we currently have.

Later I would also like to swich the current Has caches to base onto the event system.
It is quite possible that in future we will have to implement priority to the events but this can be solved.

### DIFF
--- a/filestore/filestore.go
+++ b/filestore/filestore.go
@@ -10,13 +10,12 @@ package filestore
 import (
 	"context"
 
-	"gx/ipfs/QmVA4mafxbfH5aEvNz8fyoxC6J1xhAtw88B4GerPznSZBg/go-block-format"
-
 	"github.com/ipfs/go-ipfs/blocks/blockstore"
 	posinfo "github.com/ipfs/go-ipfs/thirdparty/posinfo"
 
 	logging "gx/ipfs/QmSpJByNKFX1sCsHBEp3R73FL4NF6FnQTEGyNAXHm2GS52/go-log"
 	cid "gx/ipfs/QmTprEaAA2A9bst5XH7exuyi5KzNMK3SEDNN8rBDnKWcUS/go-cid"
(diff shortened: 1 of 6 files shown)


Kubuxu:
Results of benchmark:
```
BenchmarkEventFireSingle-12      	 1000000	        24.1 ns/op
BenchmarkEventFireMultiple-12    	 1000000	        4.37 ns/op
```
and zero allocations (to wide to fit).

Kubuxu:
The codelimate is dubious, for some reason it is old status but the codecov side is green.

filestore/filestore.go
```diff
@@ -10,9 +10,10 @@ package filestore
 import (
 	"context"
 
+	"gx/ipfs/QmVA4mafxbfH5aEvNz8fyoxC6J1xhAtw88B4GerPznSZBg/go-block-format"
```
magik6k: This should be below

Stebalien:
So, put/delete events make a lot of sense, IMO. However...

> have add-on functionality in blockstore (imagine Ethereum blockstore plugin)

I'm weary of event based systems like this. I used to implement *everything* this way but ended up deeply regretting it years down the road because it makes it very difficult to determine what happens at any given point in time and what's responsible for what.

Personally, I'd prefer some kind of dispatching, pluggable, multi-level blockservice.

MichaelMure:
This would also allow later to build an API to give locality update (block added/removed). This is very useful when you build an app on top of IPFS.

Stebalien:
(no longer applies)

PR #4133, opened 2017-08-09
Improve dag API

### LINKED ISSUE #3771: DAG API
Ref: https://github.com/ipfs/interface-ipfs-core/pull/123

```bash
» ipfs dag put --help
#...
  -f,        --format string - Format that the object will be added as. Default: cbor.
  --input-enc         string - Format that the input object will be. Default: json.
#...
```

As discussed previously:

- `format` should take in the names of the formats (dag-cbor, dag-pb, eth-block, etc)
- There should be an option for `hashAlg`
- The endpoint needs to reply with a header that specifies the encoding of the output. A sane default would be `cbor` all the way and/or have an option for `--output-enc` too.

Related question:
- Can we get listed on the --input-enc, the actual supported encoding options available?

For context: https://github.com/ipfs/interface-ipfs-core/issues/81#issuecomment-277271941

magik6k:
Closes #3771

* Added --hash to put
* --input-enc cbor input 'encoder'
* dag-pb support in put

TODO:
* Output cbor encoding
* Have the js-ipfs-api PR working

Depends on https://github.com/ipfs/go-ipld-cbor/pull/24

cc @diasdavid

### DIFF
--- a/core/commands/dag/dag.go
+++ b/core/commands/dag/dag.go
@@ -3,6 +3,7 @@ package dagcmd
 import (
 	"fmt"
 	"io"
+	"math"
 	"strings"
 
 	cmds "github.com/ipfs/go-ipfs/commands"
@@ -11,6 +12,7 @@ import (
 	pin "github.com/ipfs/go-ipfs/pin"
 
 	cid "gx/ipfs/QmTprEaAA2A9bst5XH7exuyi5KzNMK3SEDNN8rBDnKWcUS/go-cid"
+	mh "gx/ipfs/QmU9a9NV9RdPNwZQDYd5uKsm6N6LJLSvLbywDDYFbaaC6P/go-multihash"
 )
 
 var DagCmd = &cmds.Command{
@@ -48,6 +50,7 @@ into an object of the specified format.
 		cmds.StringOption("format", "f", "Format that the object will be added as.").Default("cbor"),
 		cmds.StringOption("input-enc", "Format that the input object will be.").Default("json"),
 		cmds.BoolOption("pin", "Pin this object when adding.").Default(false),
+		cmds.StringOption("hash", "Hash function to use").Default(""),
 	},
 	Run: func(req cmds.Request, res cmds.Response) {
 		n, err := req.InvocContext().GetNode()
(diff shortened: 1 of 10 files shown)


core/commands/dag/dag.go
```diff
@@ -48,6 +50,7 @@ into an object of the specified format.
 		cmds.StringOption("format", "f", "Format that the object will be added as.").Default("cbor"),
 		cmds.StringOption("input-enc", "Format that the input object will be.").Default("json"),
 		cmds.BoolOption("pin", "Pin this object when adding.").Default(false),
+		cmds.StringOption("hash", "Hash function to use").Default(""),
```
diasdavid: can this be `hash-alg` to match https://github.com/ipfs/interface-ipfs-core/tree/master/API/dag#dag-api ?

diasdavid:
@magik6k woot! Thanks for pushing this forward. Did you get to test it with the WIP PR on js-ipfs-api? Ideally we should land both at the same time so that we are 100% convinced that is what we want.

core/commands/dag/dag.go
magik6k: It's `hash` in `ipfs add`: https://github.com/ipfs/go-ipfs/blob/master/core/commands/add.go#L44, I guess it's better for consistency

diasdavid: Fair point.

magik6k:
Ran the tests now:

* Format/inputEnc 'translation' should be removed (https://github.com/ipfs/js-ipfs-api/pull/534/files?diff=unified#diff-88b6c5823413c7e7b38c6868090fef29R35)
* If the serialization is handled on the API side, inputEnc can just be `raw`
* Should I take over the js-api PR?

diasdavid:
@magik6k you are welcome to! :)

whyrusleeping:
@magik6k LGTM, though the OSX sharness tests appear to be failing. Use printf instead of echo, it seems to work better cross platform.

magik6k:
Yeah, I'll have a look at it. Don't merge yet as I want to get https://github.com/ipfs/js-ipfs-api/pull/534 fully functional with this

diasdavid:
@whyrusleeping I believe this wasn't done yet. See @magik6k [comment](https://github.com/ipfs/go-ipfs/pull/4133#issuecomment-322001549)

> Yeah, I'll have a look at it. Don't merge yet as I want to get ipfs/js-ipfs-api#534 fully functional with this

magik6k:
Yes, it was merged as I wanted to put the larger changes into separate PR.

PR #4139, opened 2017-08-12
Update commands with progress bars

jbott:
- Show speed on `ipfs get`
- Show a progress bar (with speed) on `ipfs pin add`

License: MIT
Signed-off-by: John Ott <john@johnott.us>

### DIFF
--- a/merkledag/merkledag.go
+++ b/merkledag/merkledag.go
@@ -175,7 +175,15 @@ func FetchGraph(ctx context.Context, root *cid.Cid, serv DAGService) error {
 	set := cid.NewSet()
 	visit := func(c *cid.Cid) bool {
 		if set.Visit(c) {
-			v.Increment()
+			cnt := 0
+			node, err := serv.Get(ctx, c)
+			if err == nil {
+				stat, err := node.Stat()
+				if err == nil {
+					cnt = stat.LinksSize + stat.DataSize
+				}
+			}
+			v.Increment(cnt)
 			return true
 		} else {
 			return false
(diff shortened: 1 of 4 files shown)


merkledag/merkledag.go
```diff
@@ -175,7 +175,15 @@ func FetchGraph(ctx context.Context, root *cid.Cid, serv DAGService) error {
 	set := cid.NewSet()
 	visit := func(c *cid.Cid) bool {
 		if set.Visit(c) {
-			v.Increment()
+			cnt := 0
+			node, err := serv.Get(ctx, c)
+			if err == nil {
+				stat, err := node.Stat()
+				if err == nil {
+					cnt = stat.LinksSize + stat.DataSize
```
jbott: I'm not entirely sure if this is correct. I tried using `stat.BlockSize` but that did not seem to include the DataSize measurement.

jbott:
I'm not entirely sure how you guys would like these tests fixed. Looks like a failure due to the changed format of `ipfs add --progress`. It looks like test for `ipfs get` doesn't check the progress output?

kevina:
Hi @jbott 

> I'm not entirely sure how you guys would like these tests fixed.

Could you be more specific here.

jbott:
Oops, sorry I wasn't too explicit in what this patch was supposed to accomplish.

This adds size based progress bars to the `ipfs pin add --progress` command, along with enabling the show speed option on `ipfs get`.


The specific test that is failing is here: https://github.com/ipfs/go-ipfs/blob/master/test/sharness/t0085-pins.sh#L96

It's due to the change in the output format of the `pin add` with the `--progress` option. `ipfs get` has no corresponding test that matches this output format, and just choses to ignore checking the command's output (I believe).

If you guys have a preferred way of fixing this test (or alternatively just removing it), I would be happy to do so.

PR #4148, opened 2017-08-17
document circuit relay in docs/experimental-features

vyzo:


vyzo:
Summoning @whyrusleeping @Kubuxu

docs/experimental-features.md
```diff
@@ -212,3 +213,55 @@ Basic usage:
...
+## Circuit Relay
+
+Allows peers to connect through an intermediate relay node when there
+is no direct connectivity.
+
+### State
+Experimental
+
+### In Version
+master
```
magik6k: It's safe to say it will be in `0.4.11` here.

vyzo: cool, will add.

docs/experimental-features.md
```diff
@@ -212,3 +213,55 @@ Basic usage:
...
+try to dial through known relays:
+`ipfs swarm connect /p2p-circuit/ipfs/QmB`
+
+Peers can see their (unspecific) relay address in the output of
+`ipfs swarm addrs listen`
+
+### Road to being a real feature
+
+- [ ] Needs more people to use it and report on how well it works.
+- [ ] Advertise relay addresses to the DHT for NATed or otherwise
```
Stebalien: I've opened an issue to discuss this: https://github.com/libp2p/specs/issues/33

MichaelMure:
After reading this, I still have a question. Once you have configured a relay node, would that mean that all the traffic would go through this node (as a proxy) or would that be a fallback if no better route is found ?

Also, would protocol labs provide public relay enabled node ? @lgierth ?

whyrusleeping:
Yes, we will provide public relay enabled nodes.

vyzo:
> Once you have configured a relay node, would that mean that all the traffic would go through this node (as a proxy) or would that be a fallback if no better route is found ?

For initial (experimental) deployment, it's manual -- you have to manually open connections to peers using a relay.

Issue #3679, opened 2017-02-12
better gc


whyrusleeping:
I have a large enough ipfs repo (over 300 million blocks) that it takes a very large amount of memory to perform a gc. computing the marked set is expensive.

I'm thinking that using something like a bloom filter could make this process use up much less memory, at the expense of not cleaning out every block. The difficulty here is that false positives while enumerating the set of pinned objects could result in drastically lowered performance (we could accidentally think a block that points to *everything* is pinned and end up cleaning out nothing) so selecting parameters to try and avoid this is important.

Another (potentially more complicated but more accurate) option is to use a disk backed prefix tree to store the enumerated pinset (with heavy caching up to some memory limit to make perf tolerable). This just offloads the memory cost of storing the sets to disk, which is generally acceptable, but can be an issue as it would prevent people from running a GC if their disk was full, this is generally considered to be a bad thing.

I'm also interested in strategies that can do "a little gc". Something that allows us to quickly free a smaller subset of the blocks, without the overhead of performing an entire gc scan. Implementing something that lets us do this may require a rethinking of how pinsets and objects are stored.

Kubuxu:
I think stored, re-scalable bloom filer could be a good way to go. We already use bloom filter for Has request filtering. Both GC and Has cacheing would benefit from it.

kevina:
Another disk based alternative is to use two sorted lists.  First write out a list of the colored set, then sort it (duplicates can be removed during the sorting phase).  Then create another list of all keys that are candidates to be removed and sort it.  Then simply compare the two lists.  The second list is not strictly needed as a binary search can be performed on the colored-set list, but will help with performance as the keys are currently not guaranteed to be returned in any order.

As just a way to reduce memory.  Forgo any set data-structures and just use a list as before, but keep it memory.  A sorted list is bound to be more memory efficient than any any deterministic set data-structure.

By not using a set for the colored list, there is a risk of doing duplicate work (as you can't instantly check if a node has already been visited).  This may or not be a problem.

Just an idea.

whyrusleeping:
>  there is a risk of doing duplicate work (as you can't instantly check if a node has already been visited).

In practice, this is a huge cost. Very frequently we have situations where there are 100 or so nodes that point to the same large graphs (example, the go-ipfs source directory in my repo that i add to ipfs very frequently)

whyrusleeping:
[...]

PR #4149, opened 2017-08-17
Next generation of GC

Kubuxu:
TODO:
- [ ] more tests
- figure out what to do with direct pins

License: MIT


Kubuxu:
@whyrusleeping direct pins are a bit of a problem currently. I can introduce this concept to tricolor set (make it four color) but it complicates things a lot. If we want to do it, I would also have to introduce one more color for internal pins (recursive but stopping on internal pins).

What do you think? Do you think it is ok to treat direct pins as best effort pins?

Kubuxu:
Looks like `core/coreunix.TestAddGCLive` resolved itself. I will add more unittests soon. I would like to merge it soon as this works and we can figure out what to do with direct pins latter.

Short live feature branches. 😄

kevina:
@Kubuxu so are direct pins being handled correctly right now?  That is if there is nothing but a direct pin pinning a node, will all of its children still be garbage collected?

Kubuxu:
Yes, I am currently using the same "cheat" as the older GC was using.
I don't like it, but it  can be fixed in future as this isn't a new issue.

pin/gc/gc.go
```diff
@@ -35,144 +54,159 @@ type Result struct {
...
+				output <- Result{Error: criticalError}
+			}
+		}()
+
+		// Enumerate without the lock
+		for {
+			finished, err := tri.EnumerateStep(ctx, bestEffortGetLinks, getLinks)
+			if err != nil {
+				output <- Result{Error: err}
+				criticalError = ErrCannotFetchAllLinks
```
whyrusleeping: we dont want to halt here? Whats the point in continuing?

Kubuxu: To collect all errors like that.

kevina: I would add that this is how I modified the GC to work in https://github.com/ipfs/go-ipfs/pull/3712.  It make it easier for the user to figure out and fix the problem in one go.

pin/gc/gc.go
```diff
@@ -35,144 +54,159 @@ type Result struct {
...
-	errors := false
-	gcs := cid.NewSet()
-	getLinks := func(ctx context.Context, cid *cid.Cid) ([]*node.Link, error) {
-		links, err := ls.GetLinks(ctx, cid)
-		if err != nil {
-			errors = true
-			output <- Result{Error: &CannotFetchLinksError{cid, err}}
+		// Reenumerate, fast as most will be duplicate
+		for {
+			finished, err := tri.EnumerateStep(ctx, getLinks, bestEffortGetLinks)
```
whyrusleeping: shouldnt we technically loop back up to the top of the function? technically you're supposed to just keep doing all that until you end up with an empty gray set, no?

pin/gc/gc.go
```diff
@@ -35,144 +54,159 @@ type Result struct {
...
+			if e.getColor() != tri.white {
+				blackSetSize++
+				continue
+			}
+			whiteSetSize++
+
+			c, err := cid.Cast([]byte(v))
+			if err != nil {
+				// this should not happen
+				panic("error in cast of cid, inpossibru " + err.Error())
```
whyrusleeping: lol, probably want to un-meme this panic.

pin/gc/gc.go
Kubuxu: I don't understand, Line 156 loops this until finished is true L162.

pin/gc/triset.go
```diff
@@ -0,0 +1,160 @@
...
+
+func (t trielement) getEnum() trielement {
+	return t & enumMask
+}
+
+type triset struct {
+	// colors are per triset allowing fast color swap after sweep,
+	// the update operation in map of structs is about 3x as fast
+	// as insert and requres 0 allocations (keysize allocation in case of insert)
+	white, gray, black trielement
```
whyrusleeping: how is this different from just having constants?

pin/gc/triset.go
```diff
@@ -0,0 +1,160 @@
...
+	tr := &triset{
+		white: color1,
+		gray:  color2,
+		black: color3,
+
+		grays:  make([]*cid.Cid, 0, 1<<10),
+		colmap: make(map[string]trielement),
+	}
+
+	tr.freshColor = tr.white
```
whyrusleeping: something feels too complicated here...

pin/gc/triset.go
```diff
@@ -0,0 +1,160 @@
...
+	return tr
+}
+
+// InsertFresh inserts fresh item into a set
+// it marks it with freshColor if it is currently white
+func (tr *triset) InsertFresh(c *cid.Cid) {
+	e := tr.colmap[c.KeyString()]
```
whyrusleeping: `e, ok := ....`, right? i guess its just a uint8, and 0 is colorNull?

pin/gc/triset.go
```diff
@@ -0,0 +1,160 @@
...
+	newEnum := enumFast
+	if strict {
+		newEnum = enumStrict
+	}
+
+	e := tr.colmap[c.KeyString()]
+	cl := e.getColor()
+	// conditions are:
+	// 1. empty
+	// 2. while
```
whyrusleeping: s/while/white/ ?

pin/gc/triset.go
```diff
@@ -0,0 +1,160 @@
...
+		e = tr.colmap[c.KeyString()]
+		tr.grays = tr.grays[:len(tr.grays)-1]
+	}
+
+	strict := e.getEnum() == enumStrict
+
+	// select getLinks method
+	gL := getLinks
+	if strict {
+		gL = getLinksStrict
```
whyrusleeping: should the strict thing wrap the passed in getLinks? or maybe strict gets passed into getLinks?

pin/pin.go
```diff
@@ -454,6 +454,39 @@ func (p *pinner) RecursiveKeys() []*cid.Cid {
...
+		},
+		{
+			Get:    nilErr(p.DirectKeys),
+			Strict: true,
+			Direct: true,
+		},
+		{
+			Get:     nilErr(p.InternalPins),
+			Strict:  true,
+			Inernal: true,
```
whyrusleeping: typo

pin/pinsource.go
```diff
@@ -0,0 +1,19 @@
...
+// Pinsource is structure describing source of the pin
+type PinSource struct {
+	// Get if a function that will be called to get the pins for GCing
+	Get func() ([]*cid.Cid, error)
+	// Strict makes the GC fail if some objects can't be fetched during
+	// recursive traversal of the graph
+	Strict bool
+	// Direct marks the pinned object as the final object in the traversal
+	Direct bool
+	// Inernal marks the pin source which recursive enumeration should be
```
whyrusleeping: s/Inernal/Internal/

pin/gc/gc.go
whyrusleeping: nvm, i see that we do loop up above.

pin/gc/gc.go
```diff
@@ -35,144 +54,159 @@ type Result struct {
...
+
+	err := pn.Flush()
+	if err != nil {
+		output <- Result{Error: err}
+		close(output)
+		return output
+	}
+	addRoots(tri, pn, bestEffortRoots)
 
 	go func() {
```
whyrusleeping: this goroutine should probably be its own function now

pin/gc/triset.go
Kubuxu: yes, 0 is null value

pin/gc/triset.go
Kubuxu: It is future proofing for concurrent/incremental GC and lower overhead for re-GCing.

pin/gc/triset.go
Kubuxu: https://github.com/ipfs/go-ipfs/pull/4149#discussion_r135936030

pin/gc/gc.go
Kubuxu: I am working on refactoring the GC to not be just a function but to have structure, so it will be changed.

whyrusleeping:
ref #3679

whyrusleeping:
@Kubuxu update here?

Kubuxu:
I will fix the conflict and it should be good for merge IIRC.

Kubuxu:
@whyrusleeping @Stebalien this should be done(ish), please give it a CR.

mfs/system.go
```diff
@@ -123,6 +124,24 @@ func (kr *Root) Flush() error {
 	return nil
 }
 
+// PinSource returns information about pinning requirements.
+func (kr *Root) PinSource() *pin.Source {
+	return &pin.Source{
+		Get: func() ([]*cid.Cid, error) {
+			err := kr.Flush()
```
Stebalien: I assume this won't trigger a recursive GC. We take a lock, right? However, this can cause us to go over the GC limit even more. We should probably document the fact that the "max repo size" isn't *really* the max.

Stebalien: We can also get in a bad state here where we can't flush (no disk space) and can't GC (can't flush). Not sure if we can do something about it now but...

pin/source.go
```diff
@@ -0,0 +1,34 @@
...
+	// Get if a function that will be called to get the pins for GCing
+	Get func() ([]*cid.Cid, error)
+	// Strict makes the GC fail if some objects can't be fetched during
+	// recursive traversal of the graph
+	Strict bool
+	// Direct marks the pinned object as the final object in the traversal
+	Direct bool
+	// Internal marks the pin source which recursive enumeration should be
+	// terminated by a direct pin
+	Internal bool
```
Stebalien: Can we have better names for these? I'm still confused, even after reading the comments.

* I understand `Strict` and can't think of a better name.
* `Direct`, IIRC, means non-recursive. That comment is technically correct but leads one to believe that it's a terminal pin in a DAG (`Internal`). Maybe invert this and use `Recursive`?
* `Internal` -> `Terminal`?  Or, maybe, `Never` (i.e., never pin)? Technically, it's a pin that's "internal" to a recursive pin but the name doesn't indicate anything about it being a "terminal" pin.

Stebalien: Also, "internal" is a rather big hammer. I guess it's a decent starting point (for, e.g., excluding large files globally) but, really, I'd like to be able to say, e.g., "pin A recursively but stop at X and Y" and have this pin *not* mess with other pins (that may want X/Y). Are pins like this expressible with this GC without major surgery? This isn't a reason to hold this PR (future work) up but we're going to want to be able to express constraints like this (and, eventually, arbitrary IPLD selectors for, e.g., ipfs-cluster).

pin/source.go
```diff
@@ -0,0 +1,34 @@
...
+// SortValue is sort order for the GC
+func (p Source) SortValue() int {
+	v := 0
+	if !p.Strict {
+		v |= 1 << 0
+	}
+	if p.Direct {
+		v |= 1 << 1
+	}
+	if p.Internal {
```
Stebalien: Can internal be combined with the other two options?

pin/gc/triset.go
```diff
@@ -0,0 +1,160 @@
...
+	// 3. insufficient strictness
+	if cl == colorNull || cl == tr.white || (e.getEnum() < newEnum) {
+		tr.colmap[c.KeyString()] = trielement(tr.gray | newEnum)
+		if cl != tr.gray {
+			tr.grays = append(tr.grays, c)
+		}
+	}
+}
+
+func (tr *triset) blacken(c *cid.Cid, strict trielement) {
```
Stebalien: Any reason this isn't exported while the other functions are?

mfs/system.go
Kubuxu: There isn't much we can do here to recover I think, I need the actual state of the tree to perform the GC without data loss and I can't get it unless I flush the tree so it propagates the DAG updates.

pin/gc/triset.go
Kubuxu: `blaken` has to be handled with care (as it can lead to tree not being enumerated fully), ideally the GC itself would not be using it but it is a "workaround" for the Direct pins (in future we can make Direct pins separate color but and this will allow fully concurrent or incremental GC).

pin/source.go
Kubuxu: > Maybe invert this and use Recursive?

`Direct` works much nicer with default values (false).

> "pin A recursively but stop at X and Y"

This might be a bit hard and we will have to think carefully about it. If we are able to sort the selectors according to their reach (some may be super-sets of others, so may have parts in common but then select different objects).

Big optimization that comes from color set (it was already implemented before but not explicit) is that we don't need to enumerate parts of the tree that were already enumerated. 

> we're going to want to be able to express constraints like this

The color system allows for a lot of flexibility and having it here explicitly is IMO great step forward. Right now internal means "enumerate it last", which means it stops as elements that are already marked black (as they are supposedly fully enumerated but it is not the case in case of `Direct` pins).

pin/source.go
Kubuxu: It should not be combined with `Direct` as it might lead to not enumerating internal tree fully. I will add runtime panic (on PinSource registration) and comment as they should never really be used together.

It can be used with Strict without any issues.

Stebalien: I mostly care for documentation purposes.

pin/gc/gc.go
```diff
@@ -33,146 +27,232 @@ type Result struct {
...
+	AddPinSource(...pin.Source) error
+	Run(ctx context.Context) <-chan Result
+}
 
-	elock := log.EventBegin(ctx, "GC.lockWait")
-	unlocker := bs.GCLock()
-	elock.Done()
-	elock = log.EventBegin(ctx, "GC.locked")
-	emark := log.EventBegin(ctx, "GC.mark")
+type gctype struct {
```
whyrusleeping: maybe a more descriptive name? like `triColorGc`?

pin/gc/triset.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+
+	tr.freshColor = tr.white
+	return tr
+}
+
+// InsertFresh inserts fresh item into a set
+// it marks it with freshColor if it is currently white
+func (tr *triset) InsertFresh(c *cid.Cid) {
+	e := tr.colmap[c.KeyString()]
```
whyrusleeping: You probably want to bind `c.KeyString()` to a variable and reuse it.

whyrusleeping:
Still reviewing, but doesnt this changeset make it so that we now hold every single cid in the blockstore in memory? (as opposed to currently just holding the pinned objects set)

Kubuxu:
You are right, it does. This is Memory vs Time needed during locked phase compromise. In future we will have to probably keep even the pinned set on disk.

pin/source.go
Stebalien: > `Direct` works much nicer with default values (false).
> Right now internal means "enumerate it last", which means it stops as elements that are already marked black (as they are supposedly fully enumerated but it is not the case in case of Direct pins).

Oh, now I understand. Direct is an overloaded term/variable. Direct pins act as both non-recursive pins *and* stop-pins for internal pins. Is that right? At the very least, this algorithm needs to be documented somewhere. Personally, I'd prefer something like:

```go
type PinKind int

const (
    PinRecursive PinKind = iota
    PinTerminal // Terminates an internal pin.
    PinInternal // Better name...
    PinSingle // one-of pins, add these last.
) // sort order is const order.

struct PinSource {
    Kind PinKind // Can only be one type.
    Strict bool
    Get GetFunc
}
```

> Big optimization that comes from color set (it was already implemented before but not explicit) is that we don't need to enumerate parts of the tree that were already enumerated

So, I know of a way to make this work but it will be a lot of work and will require storing extra data; we can deal with it later. The TL;DR is to maintain a (semi-accurate) topological sort of the DAG and operate over that. We'll have to see how much extra work maintaining this datastructure will be but it *will* make executing IPLD queries much faster (IMO, it makes sense to optimize for this).

However, that's not something we'll need to deal with now.

> Shift as in: Selector on node A is X and on node B is Y, they have both selected node C (the same node) and they both simplify to Z.

With topological sorting, I believe we can do this. Without it, we have the problem where we may have already marked C and it's children with respect to query X rooted at A by the time we get around to processing query Y rooted at B.

pin/gc/triset.go
```diff
@@ -0,0 +1,160 @@
...
+	// enum from enumerator
+	// stricter enumerators should have higer value than those less strict
+	enumFast   trielement = 0
+	enumStrict trielement = 1 << 2
+	enumMask   trielement = enumStrict
+)
+
+// this const is used to enable runtime check for things that should never happen
+// will cause panic if they happen
+const pedantic = true
```
Stebalien: Not used.

pin/gc/gc.go
```diff
@@ -33,146 +27,232 @@ type Result struct {
...
+				tri.blacken(c, enumStrict)
+			}
+		}
 	}
 
-	for _, k := range pn.DirectKeys() {
-		gcs.Add(k)
+	// Reenumerate, fast as most will be duplicate
+	for {
+		finished, err := tri.EnumerateStep(ctx, getLinks, bestEffortGetLinks)
```
Stebalien: Are `getLinks` and `bestEffortGetLinks` flipped?

pin/gc/gc.go
```diff
@@ -33,146 +27,232 @@ type Result struct {
...
+		if err != nil {
+			criticalError = err
+			return
+		}
+		for _, c := range cids {
+			if !r.Direct {
+				tri.InsertGray(c, r.Strict)
+			} else {
+				// this special case prevents incremental and concurrent GC
+				tri.blacken(c, enumStrict)
```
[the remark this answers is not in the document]
Kubuxu: Good catch.

kevina:
> @whyrusleeping said:
>> Still reviewing, but doesnt this changeset make it so that we now hold every single cid in the blockstore in memory? (as opposed to currently just holding the pinned objects set)

@Kubuxu said:
> You are right, it does. This is Memory vs Time needed during locked phase compromise. In future we will have to probably keep even the pinned set on disk. I can change if you think it is needed.

This concerns me.  For very large repositories this could be a problem.  This could also be a problem if the in-memory data structures are larger. Do we have some sort of estimate on the memory usage with the orig and this GC for a large repository that is say 20% pinned, 50% pinned, 80% pinned?

whyrusleeping:
Kinda makes me think we should have a `--light` mode where we use bloom filters instead of full sets.

pin/gc/gc.go
```diff
@@ -33,146 +27,232 @@ type Result struct {
...
+
+func getGCLock(ctx context.Context, bs bstore.GCBlockstore) (bstore.Unlocker, *logging.EventInProgress) {
+	elock := log.EventBegin(ctx, "GC.lockWait")
+	unlocker := bs.GCLock()
+	elock.Done()
+	elock = log.EventBegin(ctx, "GC.locked")
+	return unlocker, elock
+}
+
+func addRoots(tri *triset, pn pin.Pinner, bestEffortRoots []*cid.Cid) {
```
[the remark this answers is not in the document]
Stebalien: I'd just push it now. The comments should stay.

Kubuxu:
@kevina 
> Do we have some sort of estimate on the memory usage with the orig and this GC for a large repository that is say 20% pinned, 50% pinned, 80% pinned?

Worst case scenario it will be `total size of all keys + number of keys * 1b`.

Kubuxu:
So in case of sha-256 and CIDv1 it would be about 28M keys per GiB.

pin/gc/gc.go
```diff
@@ -33,146 +27,232 @@ type Result struct {
...
-			errors = true
-			output <- Result{Error: &CannotFetchLinksError{cid, err}}
+loop:
+	for {
+		select {
+		case c, ok := <-keychan:
+			if !ok {
+				break loop
+			}
+			tri.InsertFresh(c)
```
Stebalien: I think there's a bug here (well, not *here* but this is where I started writing the comment...). Specifically, what happens if I add a node (e.g., from the network) after the above `EnumerateStep` calls? I believe it will be garbage collected even if it shouldn't be.

The solution, as far as I can tell, is to leave unexplored gray nodes in the gray set and try to re-explore them after taking the GC lock.

Ideally, we'd have some nice way to subscribe to events from the blockstore so we could just process new blocks...

pin/gc/gc.go
```diff
@@ -33,146 +27,232 @@ type Result struct {
...
+func (g *gctype) gcAsync(ctx context.Context, emark *logging.EventInProgress,
+	output chan Result, tri *triset) {
 
-		// EnumerateChildren recursively walks the dag and adds the keys to the given set
-		err := dag.EnumerateChildren(ctx, getLinks, c, set.Visit)
-		if err != nil {
-			return err
+	defer close(output)
+
+	bestEffortGetLinks := func(ctx context.Context, cid *cid.Cid) ([]*node.Link, error) {
```
ForrestWeston: Can we combine this with the below `getLinks` function, using a flag to handle the `err != dag.ErrNotFound` case?

pin/gc/triset.go
```diff
@@ -0,0 +1,160 @@
...
+// it returns error is the getLinks function errors
+// if the gray set is empty after this step it returns (true, nil)
+func (tr *triset) EnumerateStep(ctx context.Context, getLinks dag.GetLinks, getLinksStrict dag.GetLinks) (bool, error) {
+	var c *cid.Cid
+	var e trielement
+	for next := true; next; next = e.getColor() != tr.gray {
+		if len(tr.grays) == 0 {
+			return true, nil
+		}
+		// get element from top of queue
```
ForrestWeston: I think we mean to say `get element from stack` here - based on line [54](https://github.com/ipfs/go-ipfs/pull/4149/files#diff-6f5952dd6cc3a002b44ca6b6d6d725e2R54) of this file.

pin/gc/gc.go
Kubuxu: `InsertFresh` in this inserts as white. I will change it to `InsertWhite` for time being. 

> what happens if I add a node (e.g., from the network) after the above EnumerateStep calls

If it is pinned it will be enumerated and marked black by enumeration when the lock is taken.

pin/gc/gc.go
Kubuxu: You are right, two types of GetLinks is cruft from previous (or currently used GC), I should had eliminated it.

Kubuxu:
@whyrusleeping @kevina for future we might just switch to using something like badger for GC, especially if we nail down incremental/concurrent GCing. In case of extremely bit repos the GC process could take days so having it done incrementally and being resumable would be a requirement.

pin/gc/gc.go
Stebalien: You sure? I'm pretty sure those roots won't be re-enumerated because they're already black. I'm worried about the `a (pinned) -> b -> c (added late)` case. That is, `a` is a pin root and we only add `c` to the repo after the initial enumeration. I believe, in this case, we'll never mark `c` as black because `a` is already black.

Kubuxu: b can't be black as it references a white (or null, treated as white) node c.

If a is pinned strictly (by recursive pin) then marking b would fail with an error.
if a is pinned non-strictly (by Files API) then it is still better than it was before.

Stebalien: > b can't be black as it references a white (or null, treated as white) node c.

No, if you have `b`, you'll definitely blacken it. `getLinks` will succeed because you have it (even if you don't have `c`). Yes, if you're pinning strictly you'll fail out.

>  if a is pinned non-strictly (by Files API) then it is still better than it was before.

It's still not correct. Actually, can't you just add the white nodes *first* (before you enumerate reachable nodes)? IIRC, that's how this algorithm usually works (for this reason).

pin/gc/gc.go
```diff
@@ -33,146 +27,232 @@ type Result struct {
...
+			} else {
+				// this special case prevents incremental and concurrent GC
+				tri.blacken(c, enumStrict)
+			}
+		}
 	}
 
-	for _, k := range pn.DirectKeys() {
-		gcs.Add(k)
+	// Reenumerate, fast as most will be duplicate
```
Stebalien: Not true for `Internal` pins as far as I can tell. Do we use them a lot?

Kubuxu: No, there will be very little as they are used only for building sharding of the pinset.

pin/gc/gc.go
```diff
@@ -33,146 +27,232 @@ type Result struct {
...
+// Result represents an incremental output from a garbage collection
+// run.  It contains either an error, or the cid of a removed object.
+type Result struct {
+	KeyRemoved *cid.Cid
+	Error      error
+}
+
+func getGCLock(ctx context.Context, bs bstore.GCBlockstore) (bstore.Unlocker, *logging.EventInProgress) {
+	elock := log.EventBegin(ctx, "GC.lockWait")
+	unlocker := bs.GCLock()
```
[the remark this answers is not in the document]
Kubuxu: Yes, to prevent for example blocks for being added. In future we can have incremental GC where those pauses (this is what taking major lock by GC is called) are short but spread over time.

pin/gc/gc.go
Kubuxu: > No, if you have b, you'll definitely blacken it. getLinks will succeed because you have it (even if you don't have c). Yes, if you're pinning strictly you'll fail out.

Yes, then the `c` will be marked gray and processed in the next EnumerateStep.

> can't you just add the white nodes first (before you enumerate reachable nodes)?

I treat every node that doesn't exist in the tricolor set as white. They are being added here as it reduced time needed what we take the repo's GC lock.

pin/gc/gc.go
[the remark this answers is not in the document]
Kubuxu: Or to highlight the problem more cleanly, to prevent `ipfs add` from adding blocks that are referencing some white marked nodes.

pin/gc/gc.go
Stebalien: Not if it isn't strict. In that case, it will be marked gray, in the first `EnumerateStep` and then removed (and assumed to have no children).

> They are being added here as it reduced time needed what we take the repo's GC lock.

That doesn't mean you can't add them *before* the *first* enumerate step. Then, you'd follow the standard algorithm of moving nodes from the white set to the gray set to the black set until the gray set is empty.

pin/gc/gc.go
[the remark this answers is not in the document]
Stebalien: I see. I was thinking that that case might be equivalent to doing the GC before adding the block but that's not true because the fact that we had both blocks at the same time can be observed. Having a GCed system where something can go from unreachable (by a root) to reachable in the middle of a GC run is weird.

pin/gc/gc.go
Kubuxu: That is true, let's do that.

parkan:
seems like there is some overlap with https://github.com/ipfs/go-ipfs/pull/4149

parkan:
seems like there is some overlap with https://github.com/ipfs/go-ipfs/issues/5765

Stebalien:
Possibly. The primary issue with this PR is that it loads the entire pin-set into memory. That's not going to scale.

parkan:
@Stebalien that makes sense

let's close this with that explanation, unless you or @Kubuxu have objections

there's a lot of interest in improved GC, but it seems that this branch has diverged enough that basing any further work on it will probably be counterproductive

Issue #4009, opened 2017-06-26
[RFC] Add an offline hint to contexts


Stebalien:
Instead of manually constructing services without online components, it would be useful to have a context hint that tells services to avoid network requests. That is:

```go
func Offline(ctx context.Context) context.Context {
	return context.WithValue(ctx, Offline, true)
}

func IsOffline(ctx context.Context) bool {
	value, ok := ctx.Value(Offline).(bool)
	return ok && value
}
```

One could then define offline service adapters as follows:

```go
func OfflineDAGService(ds DAGService) DAGService {
	&offlineDAGService{ds}
}

type offlineDAGService struct {
	online DAGService
}

func (ds *offlineDAGService) Get(ctx context.Context, n Node) (Node, error) {
	return ds.online.Get(Offline(ctx), n)
}

func (ds *offlineDAGService) Add(n Node) (*cid.Cid, error) {
	return ds.online.Add(n)
}

func (ds *offlineDAGService) Remove(n Node) error {
	return ds.online.Remove(n)
}

// etc...
```

whyrusleeping:
I like using contexts this way. :+1: here, though we will have to clearly define expectations; what exactly is meant by a 'network call' and how do we behave in contexts where everything is a networked call? the hard part here is bitswap, how would bitswap behave when given an 'offline' context?

whyrusleeping:
cc @Kubuxu @kevina @hsanjuan @lgierth and anyone else who has opinions here

Stebalien:
> the hard part here is bitswap, how would bitswap behave when given an 'offline' context?

I assume it would refuse to work.

---

For context, I want to be able to easily construct arbitrary offline variants of services but *don't* want to add more functions like `GetOfflineLinkService` to *every* service.

whyrusleeping:
> For context, I want to be able to easily construct arbitrary offline variants of services but don't want to add more functions like GetOfflineLinkService to every service.

Yeah, agreed. We just need to be careful and make sure we don't miss adding the offline checking logic to any service that needs it.

kevina:
I am not sure I like the idea of a "hint".  For somethings (like maybe the GC) I greatly prefer a hard guarantee that the service will not make network calls because there is no code for it in the service used, rather than a hint that could be ignored.

kevina:
I do agree that GetOfflineLinkService is a hack, but I think there should be a more uniform way to get at a service that does not use the network rather than ask an existing service to avoid network requests.

[...]

PR #4150, opened 2017-08-17
'FetchBlocks' gateway option

Voker57:
If FetchBlocks=false, don't fetch blocks which are not already present in storage.

Also add Gateway.FetchBlocks option

This mode can be useful for creating gateways which are supposed to serve only chosen content.


Kubuxu:
There is already `ipfs daemon --offline`. Is it not good enough?

Voker57:
@Kubuxu --offline will also disable fetching via API. I want only gateway to be restricted.

magik6k:
I think it might be better to implement this as a config option

cmd/ipfs/daemon.go
```diff
@@ -148,6 +149,7 @@ Headers.
 		cmds.BoolOption(initOptionKwd, "Initialize ipfs with default settings if not already initialized").Default(false),
 		cmds.StringOption(routingOptionKwd, "Overrides the routing option").Default("dht"),
 		cmds.BoolOption(mountKwd, "Mounts IPFS to the filesystem").Default(false),
+		cmds.BoolOption(offlineGatewayKwd, "Do not fetch blocks because of gateway requests").Default(false),
```
whyrusleeping: s/because of/for/

magik6k:
IMO daemon already has too many flags (`--enable-gc` as one example that could be in config). Gateway already has it's config scope, this could be done in config as `Gateway.Offline`. I'd say that all options that make sense in live deployments should be in config.

Stebalien:
How about something like: https://github.com/ipfs/go-ipfs/issues/4180

lgierth:
Agreed that a config option is desirable over a command flag.

I'm afraid "offline" might not be the best name here (and you're also inconsistent naming it "InternalDAG". We're already giving "offline" way too many meanings in various contexts.

lgierth:
Oh FFS where did my comment go? The form apparantly reset to a very early stage of it...

lgierth:
Phew I recovered it -- deleting my two comments above.

-----

Good idea :+1: 

Agreed that a config option is desirable over a command flag, and we'll also want to refactor the gateway a bit. It should get a GatewayOpts struct, similar to the HostOpts struct I implemented in https://github.com/libp2p/go-libp2p/pull/197

I'll put up to debate whether "offline" is a good name though. It already means a couple of different things in go-ipfs. This option seems spiritiually related to the `ipfs add --local` and `ipfs name resolve --local` flags, in that it only operates locally (apart form IPNS lookups I guess). I'm not ready to say one is better of the other, just marking this as an important bit. The reason for it's importance is probably outside of this PR and more in the already existing meanings of "offline". It's important the proposed feature fits in nicely with the existing modes of connectedness and disconnectedness and doesn't unneccessarily complicate them.

lgierth:
Btw, mentioning the `--local` flag, I noticed that its description in `ipfs --help` is completely inaccurate. See #4182.

Voker57:
@lgierth Isn't GatewayConfig already doing same job as proposed GatewayOpts? https://github.com/ipfs/go-ipfs/blob/master/core/corehttp/gateway.go#L14

lgierth:
Yeah you're right that's already useful

whyrusleeping:
Maybe the better term would be `NoFetch` as a field inside the gateway options?

Voker57:
Renamed option to FetchBlocks and moved to config.

nicola:
I want this so badly :(

Voker57:
@nicola yes.

Kubuxu:
Where is the option handled in the gateway? I see it only added to ServeOptions.

Voker57:
Sorry, I screwed up a rebase. Will update.

Kubuxu:
Also:
```
+ ./bin/gx deps dupes

+ test -z package go-ipfs-config imported as both QmSoYrBMibm2T3LupaLuez7LPGnyrJwdRxvTfPUyCp691u and QmQwsBSJXHs3nHBLPJKd1iBNpW9uJ9acSpqbbjafvNJQc3
```

and as tests passed on circleci and travis, they seem to be too weak. I would recommend using IPTB to setup 2 node cluster.

Voker57:
Fixed the PR and improved the tests.

core/coreapi/coreapi.go
```diff
@@ -16,19 +16,26 @@ package coreapi
...
 
 type CoreAPI struct {
 	node *core.IpfsNode
+	dag  ipld.DAGService
 }
 
 // NewCoreAPI creates new instance of IPFS CoreAPI backed by go-ipfs Node.
-func NewCoreAPI(n *core.IpfsNode) coreiface.CoreAPI {
-	api := &CoreAPI{n}
+func NewCoreAPI(n *core.IpfsNode, offlineMode bool) coreiface.CoreAPI {
```
Stebalien: This is really not the right place to configure this.

core/coreapi/coreapi.go
```diff
@@ -16,19 +16,26 @@ package coreapi
...
-
 	logging "gx/ipfs/QmZChCsSt8DctjceaL56Eibc29CVQq4dGKRXC5JRZ6Ppae/go-log"
+	ipld "gx/ipfs/QmdDXJs4axxefSPgK6Y1QhpJWKuDPnGJiqgq4uncb4rFHL/go-ipld-format"
 )
 
 var log = logging.Logger("core/coreapi")
 
 type CoreAPI struct {
 	node *core.IpfsNode
+	dag  ipld.DAGService
```
Stebalien: This is going to be a footgun. Now the CoreAPI needs to make sure to use *it's* dag instead of the global one.

Maybe https://github.com/ipfs/go-ipfs/issues/4009? This keeps on haunting us and I've yet to see a better proposal.

magik6k:
I'll handle this

core/coreapi/coreapi.go
Voker57: why not? I don't see how @magik6k solution is better, except for modifying less code.

Stebalien: In retrospect, it's probably a reasonable place to do this (although I would have used functional options). Really, I was reacting to the fact that the CoreAPI needed to store a separate DAGService (I don't want it to turn into yet another "IpfsNode"). However, it turns out that we have to do that for sessions (no way around that as far as I can tell given the current sessions design) so this really isn't an issue.

Voker57:
Rebased and updated to use new API. It is now similar to #5649 but more limited in scope: will still resolve IPNS fully, which is my desired behavior. Not sure how to cover this by tests, ones included in this PR would still pass but IPNS will not be queried fully with #5649.

core/coreapi/coreapi.go
```diff
@@ -217,6 +216,10 @@ func (api *CoreAPI) WithOptions(opts ...options.ApiOption) (coreiface.CoreAPI, e
 
 	}
 
+	if !settings.FetchBlocks {
+		subApi.dag = dag.NewDAGService(bserv.New(subApi.blockstore, offlinexch.Exchange(subApi.blockstore)))
```
magik6k: `settings.FetchBlocks` needs to set `exchange`/`blocks` too, otherwise it still will be possible to use gateway to fetch stuff from the network with things like `/api/v0/block/get`

core/corehttp/gateway.go
```diff
@@ -8,13 +8,15 @@ import (
...
 	coreapi "github.com/ipfs/go-ipfs/core/coreapi"
+	coreopts "github.com/ipfs/go-ipfs/core/coreapi/interface/options"
 
 	id "gx/ipfs/QmRBaUEQEeFWywfrZJ64QgsmvcqgLSK3VbvGMR2NM2Edpf/go-libp2p/p2p/protocol/identify"
 )
 
 type GatewayConfig struct {
 	Headers      map[string][]string
 	Writable     bool
+	FetchBlocks  bool
```
magik6k: Is this used anywhere?

Voker57: no, leftover.

magik6k:
IPNS should now work in https://github.com/ipfs/go-ipfs/pull/5649, can you check if it works for you?

Voker57:
> IPNS should now work in #5649, can you check if it works for you?

yes, now this PR works like I need it. I'm closing this one.

PR #4152, opened 2017-08-19
Eliminate GetOfflineLinkService method from LinkService interface

kevina:
This eliminates the GetOfflineLinkService method to make https://github.com/ipfs/go-ipld-format/pull/8 easier and also avoid the need for #4009, at least for now.

It does it by keeping the method but instead exposing the type create by `NewDAGService` so that implementation specific methods can be called, with one of them being `GetOfflineLinkService`.  The GC function now expects an offline LinkService.  To help reinforce this notion a trivial type `OfflineLinkService` is used.

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -29,7 +29,7 @@ import (
 	rp "github.com/ipfs/go-ipfs/exchange/reprovide"
 	filestore "github.com/ipfs/go-ipfs/filestore"
 	mount "github.com/ipfs/go-ipfs/fuse/mount"
-	merkledag "github.com/ipfs/go-ipfs/merkledag"
+	mdag "github.com/ipfs/go-ipfs/merkledag"
 	mfs "github.com/ipfs/go-ipfs/mfs"
 	namesys "github.com/ipfs/go-ipfs/namesys"
 	ipnsrp "github.com/ipfs/go-ipfs/namesys/republisher"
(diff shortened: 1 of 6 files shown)


Stebalien:
I'm not sure about replacing that interface with a concrete type; I believe we want the dag service to remain replacable. Also, I've run into the need for an offline NodeGetter/DAGService elsewhere (pubsub) so we'll probably want some form of "don't go online" switch anyways.

kevina:
The idea behind this p.r. is to get things moving until we can figure out a better interface.  If the time comes when we want to swap out the DAGService with something else in an IPFS node (as that is the only place the concrete type is used) we can worry about it then.

kevina:
Also specific examples of where an offline NodeGetter/DAGService is used in pubsub will be helpful.

Stebalien:
> Also specific examples of where an offline NodeGetter/DAGService is used in pubsub will be helpful.

So, for background:

1. I've been working on adapting PubSub to use IPLD for messages. Instead of sending arbitrary data, one would send a CID of an IPLD object along with (optionally) a dag rooted at this CID.
2. For security (DoS) reasons, we need to be able to validate pubsub messages before we forward them (otherwise, it's trivial to DoS pubsub).

I'd like to be able to validate received DAGs using both blocks attached to the messages and blocks already present on the local node but I *do not* want nodes to start bitswaping for blocks during validation; otherwise, peers could trivially trick other peers into downloading arbitrary data.

Really, on thinking about this, fetching these nodes isn't the primary issue, it's *persisting* and *providing* them. This brings us back to that issue about adding hints to contexts.

kevina:
@Stebalien another option is to do what is done within IPFS and get an offline NodeGetter via out of band means.

Stebalien:
I won't have access to that from within pubsub. When the pubsub is constructed, it will be passed a DAGService (or something like that depending on how that refactor falls out) and it won't really have access to the rest of the system.

This problem actually extends past just pubsub/offline. Any library operating over IPLD will want access to a DAGService and will need to be able to pass hints like:

* Don't persist these, they're temporary.
* Persist this, I need it (i.e., pinning).
* Just give this to me if you have it, don't do any work getting it.
* Give this to me, here are my credentials.
* Store this but only make it accessible with certain credentials.

That's one of the nice things about stuffing this stuff in a context.

---

Actually, one way to do this could be to pass a "dead" (already canceled) context (and fix everything to check if the context has been canceled before performing any network requests). Really, this *should* already work.

Stebalien:
Superseded by #4610.

Issue #4137, opened 2017-08-11
Low Power Mode


jcgruenhage:
#### Description:
IPFS on Android (with IPFSDroid) already works kinda well, but having the app running is currently not an option, because it sucks your battery empty in no time. Would it be possible, to implement an API call, that switches a node into low power (and possibly also into low bandwidth) mode, so that running this on mobile devices is more doable? Killing the node all the time would be another way, but that does kinda defeat the purpose of IPFS.

Kubuxu:
First step would be probably running go-ipfs in DHT client mode:
`ipfs daemon --routing=dhtclient`

We are working on connection closing which should further reduce the number of connections thus bandwidth and power usage.

jcgruenhage:
Thanks :)

whyrusleeping:
Yeah, as @Kubuxu pointed out you probably want to use `--routing=dhtclient`. Its probably also worth turning off the reprovider with: `ipfs config Reprovider.Interval "0"` so your node doesnt rebroadcast all this things it has every 12 hours.

Stebalien:
Note: Bitswap sessions (https://github.com/ipfs/go-ipfs/issues/3786) will help reduce background traffic significantly.

jbshirk:
Cross ref
https://discuss.ipfs.io/t/sweet-ipfs-full-ipfs-node-for-android/3779/16

Hazae41:
Sweet-IPFS 0.2.2 adds a "low power" mode by using this config
```
"Swarm": {
    "ConnMgr": {
      "LowWater": 20,
      "HighWater": 100,
      "GracePeriod": "80s"
    }
  }
```

lidel:
`lowpower` [profile](https://github.com/ipfs/kubo/blob/master/docs/config.md#profiles)  was implemented a while ago. it is shutting down various things to decrease impact on battery:

https://github.com/ipfs/kubo/blob/6f730dc268c3712d65efb5b0ab251a714cb85ec4/config/profile.go#L171-L187

Future work would be to default to a pool of [gateways hosting blocks / CARs](https://docs.ipfs.tech/reference/http/gateway/#trustless-verifiable-retrieval) and starting up P2P/DHT only as a fallback.

PR #4154, opened 2017-08-20
Add low power init profile

magik6k:
For context https://github.com/ipfs/go-ipfs/issues/4137

### DIFF
--- a/cmd/ipfs/init.go
+++ b/cmd/ipfs/init.go
@@ -39,6 +39,8 @@ Available profiles:
         running IPFS on machines with public IPv4 addresses.
     'test' - Reduces external interference of IPFS daemon, this
         is useful when using the daemon in test environments.
+    'lowpower' - Reduces daemon overhead on the system. May affect node
+        functionality.
 
 ipfs uses a repository in the local file system. By default, the repo is
 located at ~/.ipfs. To change the repo location, set the $IPFS_PATH
--- a/cmd/ipfs/init.go
+++ b/cmd/ipfs/init.go
@@ -40,7 +40,8 @@ Available profiles:
     'test' - Reduces external interference of IPFS daemon, this
         is useful when using the daemon in test environments.
     'lowpower' - Reduces daemon overhead on the system. May affect node
-        functionality.
+        functionality - performance of content discovery and data fetching
+        may be degraded.
 
 ipfs uses a repository in the local file system. By default, the repo is
 located at ~/.ipfs. To change the repo location, set the $IPFS_PATH
--- a/docs/config.md
+++ b/docs/config.md
@@ -197,6 +197,13 @@ Default: `true`
   -  `Interval`
 A number of seconds to wait between discovery checks.
 
+- `Routing`
+Content routing mode. Can be overridden with daemon `--routing` flag.
+Valid modes are:
+  - `dht` (default)
+  - `dhtclient`
+  - `none`
+  - `supernode` (deprecated)
 
 ## `Gateway`
 Options for the HTTP gateway.
--- a/docs/config.md
+++ b/docs/config.md
@@ -203,7 +203,6 @@ Valid modes are:
   - `dht` (default)
   - `dhtclient`
   - `none`
-  - `supernode` (deprecated)
 
 ## `Gateway`
 Options for the HTTP gateway.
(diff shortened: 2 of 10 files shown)


whyrusleeping:
I like this, but perhaps we should hold off on merging until we have a few more options set? If someone inits with this profile, and we later add more things to it, they wont get those benefits.

As it stands, i'm not sure if these two options are worth enough to merit their own profile at this point.

magik6k:
Agreed, having an option to apply profile to the config after it was generated might be an option too, though for a separate PR.

whyrusleeping:
We should add a lower connection limits here, also when we get [this](https://github.com/ipfs/go-ipfs/issues/4351) implemented, set the default for this to 'dhtclient'

magik6k:
This PR actually implements the routing option (under `Discovery.Routing`)

docs/config.md
```diff
@@ -121,6 +121,13 @@ Default: `true`
   -  `Interval`
 A number of seconds to wait between discovery checks.
 
+- `Routing`
+Content routing mode. Can be overridden with daemon `--routing` flag.
+Valid modes are:
+  - `dht` (default)
+  - `dhtclient`
+  - `none`
+  - `supernode` (deprecated)
```
Stebalien: This has now been removed.

cmd/ipfs/init.go
```diff
@@ -36,6 +36,8 @@ Available profiles:
         running IPFS on machines with public IPv4 addresses.
     'test' - Reduces external interference of IPFS daemon, this
         is useful when using the daemon in test environments.
+    'lowpower' - Reduces daemon overhead on the system. May affect node
```
Stebalien: Maybe `'laptop'`? Honestly, laptop users should use this (less problems with random WiFis, fewer ephemeral DHT nodes, etc. Alternatively, we could have a separate laptop mode that occasionally reprovides pin roots.

repo/config/profile.go
```diff
@@ -41,4 +41,9 @@ var ConfigProfiles = map[string]func(*Config) error{
 		c.Discovery.MDNS.Enabled = false
 		return nil
 	},
+	"lowpower": func(c *Config) error {
+		c.Discovery.Routing = "dhtclient"
+		c.Reprovider.Interval = "0"
+		return nil
```
Stebalien: Shouldn't this also have a lower connection closing watermark?

repo/config/discovery.go
```diff
@@ -2,6 +2,9 @@ package config
 
 type Discovery struct {
 	MDNS MDNS
+
+	//Routing sets default daemon routing mode.
+	Routing string
```
whyrusleeping: I would make `Routing` its own top level section, and then give it a `Type` field (so we can add other parameters later)

cmd/ipfs/init.go
magik6k: Probably the second option, I meant this setting for mobile devices (Android phones, etc.)

magik6k:
We should chose some connection limiter settings here. Would something like 75/100 work? Or is it a 'more testing is needed' kind of situation?

whyrusleeping:
@magik6k I think that with dhtclient mode, even something as low as 20/40 could work well if we set a relatively high grace period (1m or more).

whyrusleeping:
This LGTM, weird that so much of CI is angry though...

magik6k:
Made CI less angry

whyrusleeping:
The travis OSX backlog is pretty nuts...

cmd/ipfs/init.go
VictorBjelkholm: "May affect node functionality" should be better described to tell what functionality it would affect. Seemingly, it seems to run the dht in client mode and also set really low ConnMgr values, meaning it would probably take longer to find and download content. Would be nice to add that to the description.

magik6k: done

magik6k:
Rebased

Kubuxu:
"Nothing works, everything is broken" - by "Someone had to say it at some point"

magik6k:
Fixed

whyrusleeping:
Thanks @magik6k! Let's start telling people about this one :)

PR #4155, opened 2017-08-20
reprovide: fix ipfs bitswap reprovide when interval set to 0

magik6k:
`ipfs bitswap reprovide` would hang if `Reprovider.Interval` was set to 0


whyrusleeping:
Good catch :)

PR #4161, opened 2017-08-22
retreival -> retrieval

needkane:


whyrusleeping:
@needkane could you sign off on your commit?

Issue #177, opened 2014-10-14
Installing OSX FUSE (for Mac OS X)


jbenet:
### old version kernel panic warning

Older versions of OSX FUSE (for Mac OS X) may kernel panic (it's bitten multiple of us already). 
**If you're on OSX, please make sure you have OSX FUSE 2.7.0+**
## Installing OSX FUSE
- from binary distribution
- from `brew`
### from binary distribution
1. Go to http://osxfuse.github.io
2. Download + Install the package.

![](https://www.evernote.com/shard/s198/sh/b5c4511a-5efc-47b0-b0a5-ff1fa5a9f1bf/312eb5f28526b96f364c7d927ffc7749/deep/0/Home---FUSE-for-OS-X.png)
### from `brew`

Simply:

```
brew update && brew install osxfuse
```
### Note: conflicting installs with brew

If you receive the following error: 

```
osxfuse: osxfuse is already installed from the binary distribution and conflicts with this formula.
```

Uninstall OSX FUSE binary from System Preferences before attempting to `brew install osxfuse` again.

![](https://www.evernote.com/shard/s198/sh/2b3b234c-0650-4a96-895a-65bef5eb52df/55689e06b53055fd0c013ee910cef7cd/deep/0/System-Preferences.png)
![](https://www.evernote.com/shard/s198/sh/ba3c1053-d975-4a6f-be02-66aa1c6ef254/7dcd8988773bc5e935a4d883dcb00e44/deep/0/FUSE-for-OS-X-%28OSXFUSE%29.png)

fisch0920:
On Mac OS 10.11.6, when I run `fuse mount`, I get the following:

```
Error: OSXFUSE version 3.5.3 not supported.

OSXFUSE versions <2.7.2 are known to cause kernel panics!
Please upgrade to the latest OSXFUSE version.
It is recommended you install it from the OSXFUSE website:
	http://osxfuse.github.io/

For more help, see:
	https://github.com/ipfs/go-ipfs/issues/177
```

This error message doesn't make any sense seeing as it acknowledges that I have the most recent version of fuse installed (3.5.3).

Here is the output of `fuse-version`:
```
OSXFUSE.FuseVersion: 29
OSXFUSE.AgentVersion: 3.5.3
OSXFUSE.AgentName: OSXFUSE
```

I'm guessing this is a common issue?

VictorBjelkholm:
[...]

Issue #1048, opened 2015-04-10
mount bugs on osx


jbenet:
On OSX.

Mount fails weirdly. After being on for a while, something (possibly laptop sleep) makes the fuse mount get into an odd state. it's still mounted, but will not respond. Morever, running `ipfs mount` again (which should unmount before remounting) fails. It is only my forcing unmount that things recover.

opn:
Following the instructions here - https://github.com/ipfs/go-ipfs/blob/master/docs/fuse.md - I am having difficulty with "ipfs daemon --mount" on OSX. Latest Fuse is installed:
```
OSXFUSE.FuseVersion: 27
OSXFUSE.AgentVersion: 2.8.0
OSXFUSE.AgentName: OSXFUSE
```
and my go version is go1.4.2 darwin/amd64.

Mounting seemed to work fine - all though the structure of the two Volumes could do with a little explanation.

I get permission denied when I try to add a file with nano and then drag and drop in the Finder - I had previously changed the permissions on the mountpoints to my user), and then tried:

```
ipfs config Mounts.FuseAllowOther true
Error: Failed to set config value: Failure to decode config: json: cannot unmarshal string into Go value of type bool
```
Exiting the ipfs daemon did not seem to shut things down cleanly (see below). So I'm going to restart :)

```
^C
Received interrupt signal, shutting down...
Publishing!
Publishing!
07:42:29.138 ERROR     ipnsfs: leveldb: closed system.go:79
07:42:29.138 ERROR  fuse/ipns: Error Shutting Down Filesystem: leveldb: closed
 ipns_unix.go:50
david:~ BabyBlue$ ipfs config Mounts.FuseAllowOther true
Error: Failed to set config value: Failure to decode config: json: cannot unmarshal string into Go value of type bool
david:~ BabyBlue$ go version
go version go1.4.2 darwin/amd64
david:~ BabyBlue$ fuse-version
OSXFUSE.FuseVersion: 27
OSXFUSE.AgentVersion: 2.8.0
OSXFUSE.AgentName: OSXFUSE
david:~ BabyBlue$ ipfs daemon --mount
Initializing daemon...
Publishing!
Publishing!
07:51:58.078 ERROR     ipnsfs: leveldb: closed system.go:79
Error: exit status 64: mount_osxfusefs: mount point /ipfs is itself on a OSXFUSE volume
david:~ BabyBlue$ sudo umount /ipfs
Password:
[...]

PR #4168, opened 2017-08-23
Fix brew osxfuse install instructions.

Stebalien:
Reported by @ejholmes at:

* https://github.com/ipfs/go-ipfs/issues/177#issuecomment-324290935

License: MIT

### DIFF
--- a/docs/fuse.md
+++ b/docs/fuse.md
@@ -18,7 +18,12 @@ Depending on whether you are using OSX or Linux, follow the proceeding instructi
 
 ## Mac OSX -- OSXFUSE
 
-It has been discovered that versions of `osxfuse` prior to `2.7.0` will cause a kernel panic. For everyone's sake, please upgrade (latest at time of writing is `2.7.4`). The installer can be found at https://osxfuse.github.io/. There is also a homebrew formula (`brew install osxfuse`) but users report best results installing from the official OSXFUSE installer package.
+It has been discovered that versions of `osxfuse` prior to `2.7.0` will cause a kernel panic. For everyone's sake, please upgrade (latest at time of writing is `2.7.4`). The installer can be found at https://osxfuse.github.io/. There is also a homebrew formula using cask but users report best results installing from the official OSXFUSE installer package:
+
+```sh
+brew tap caskroom/cask # if you don't have cask yet
+brew cask install osxfuse
+```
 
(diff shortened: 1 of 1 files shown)


docs/fuse.md
```diff
@@ -18,7 +18,12 @@ Depending on whether you are using OSX or Linux, follow the proceeding instructi
 
 ## Mac OSX -- OSXFUSE
 
-It has been discovered that versions of `osxfuse` prior to `2.7.0` will cause a kernel panic. For everyone's sake, please upgrade (latest at time of writing is `2.7.4`). The installer can be found at https://osxfuse.github.io/. There is also a homebrew formula (`brew install osxfuse`) but users report best results installing from the official OSXFUSE installer package.
+It has been discovered that versions of `osxfuse` prior to `2.7.0` will cause a kernel panic. For everyone's sake, please upgrade (latest at time of writing is `2.7.4`). The installer can be found at https://osxfuse.github.io/. There is also a homebrew formula using cask but users report best results installing from the official OSXFUSE installer package:
```
ejholmes: This is a tad confusing:

> There is also a homebrew formula using cask but users report best results installing from the official OSXFUSE installer package:

which then goes on to show the homebrew install method.

Maybe the sentence should read:

> Users report best results installing from the official OSXFUSE installer package, but there is also a homebrew formula using cask:

Stebalien:
@ejholmes 

Could you test IPFS with osxfuse 3.6.3? I'd like to update the suggested version as the 2.7 series is a bit out of date by this point but don't want to recommend a known broken version.

ejholmes:
Yeah. Ended up not working for me:

Stebalien:
Possibly related: https://github.com/ipfs/go-ipfs/issues/1048

Make sure nothing is currently mounted on either `/ipfs` or `/ipns` (sometimes things don't get properly unmounted when stopping the daemon).

whyrusleeping:
ipfs should probably do some sanity checks before attempting to mount. Should help mitigate these random issues a bit

Kubuxu:
The library we use bazil/fuse is not maintained currently. We might have to migrate to https://github.com/hanwen/go-fuse some time soon. We also really need to extract and clean up the fuseIPFS code into its own module.

Stebalien:
We're going to switch to an entirely new system anyways.

PR #4169, opened 2017-08-24
add blocks to the blockstore before returning them from blockservice sessions.

### LINKED ISSUE #4062: TestFetchGraph failing randomly

Stebalien:
fixes #4062 (yay!)

License: MIT

### DIFF
--- a/exchange/bitswap/bitswap.go
+++ b/exchange/bitswap/bitswap.go
@@ -298,6 +298,14 @@ func (bs *Bitswap) CancelWants(cids []*cid.Cid, ses uint64) {
 // HasBlock announces the existance of a block to this bitswap service. The
 // service will potentially notify its peers.
 func (bs *Bitswap) HasBlock(blk blocks.Block) error {
+	return bs.receiveBlockFrom(blk, "")
+}
+
+// TODO: Some of this stuff really only needs to be done when adding a block
+// from the user, not when receiving it from the network.
+// In case you run `git blame` on this comment, I'll save you some time: ask
+// @whyrusleeping, I don't know the answers you seek.
+func (bs *Bitswap) receiveBlockFrom(blk blocks.Block, from peer.ID) error {
 	select {
 	case <-bs.process.Closing():
 		return errors.New("bitswap is closed")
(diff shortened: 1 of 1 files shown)


exchange/bitswap/bitswap.go
```diff
@@ -317,8 +325,11 @@ func (bs *Bitswap) HasBlock(blk blocks.Block) error {
 	// it now as it requires more thought and isnt causing immediate problems.
 	bs.notifications.Publish(blk)
 
-	for _, s := range bs.SessionsForBlock(blk.Cid()) {
-		s.receiveBlockFrom("", blk)
+	k := blk.Cid()
+	ks := []*cid.Cid{k}
+	for _, s := range bs.SessionsForBlock(k) {
+		s.receiveBlockFrom(from, blk)
```
whyrusleeping: hrm... the change from `""` to `from` here is on purpose? Whats the reasoning? ( i dont remember all the implications without my coffee)

Stebalien: I merged two functions. `from` will be "" when the user calls `HasBlock` but will be the peer from which we received the block when called from within bitswap.

Basically, we had two functions, HasBlock and a go routine inside ReceiveMessage where:

1. The second function would:
2. Notify all interested sessions that we had received the block.
3. Remove the block from sessoins' wantlists.
4. Call the first function which would:
5. Put the block in the blockstore.
6. Publish a notification on `bs.notifications`.
7. Notify the session that we received the block (again) but claiming to have received it from "". This was to prevent a race condition where we'd notify the session that we received the block before putting it in the blockstore and then someone would try to get the block from the blockstore, not find it, and ask bitswap for it. I believe we should also have removed this block from the session but I'm not sure why we didn't (the new code does).
8. Record the block in the bitswap engine.
9. Reprovide the block.

I got rid of the first function, moved step 3 into step 7, and made it notify the session with the correct peer in step 7 if we received the block from the network.

---

The bug was that race condition in step 7. The fix works iff you keep the same bitswap instance but, if you replace the bitswap instance with a new one, the caller will not find the block in the blockstore and ask the new bitswap instance for the block (which doesn't know about it).

whyrusleeping: okay, that sounds good. Any way you could write a test?

Stebalien: `TestFetchGraph`? :smile: It's a bit hard to test a race condition...

Stebalien: (this is fixing a test failure so there is, by definition, a test case).

whyrusleeping: okay fineeee

PR #4172, opened 2017-08-28
test: query test might work sometimes, mark it as 'might fail' 

Kubuxu:
License: MIT


whyrusleeping:
@Kubuxu doesnt appear related, but the GC sharness test failure here is odd

Kubuxu:
This is weird failure.

Kubuxu:
Reported it as: https://github.com/ipfs/go-ipfs/issues/4176

whyrusleeping:
Though maybe we should be looking into figuring out whats wrong?

Kubuxu:
I was, but I couldn't figure it out right away and as I am changing GC it might fix it.

Issue #4025, opened 2017-06-30
[gateway] ServiceWorker could take control over gateway endpoints


surma:
#### Description:

Take a look at [QmaEtDG87K69mVR7sc48DemHpj3LqxKaQfDw6qJr2AGbmZ](https://gateway.ipfs.io/ipfs/QmaEtDG87K69mVR7sc48DemHpj3LqxKaQfDw6qJr2AGbmZ):

```html
<!doctype html>
<script>
  navigator.serviceWorker.register('/ipfs/QmRcqyy8yFsVhgErobYrTcF5PSQHC36TqpA19xnBkQpFcB');
</script>
```

The requested ServiceWorker file is [the following](https://gateway.ipfs.io/ipfs/QmRcqyy8yFsVhgErobYrTcF5PSQHC36TqpA19xnBkQpFcB):

```js
self.oninstall = event => event.waitUntil(self.skipWaiting());
self.onactivate = event => event.waitUntil(self.clients.claim());
self.onfetch = event => event.respondWith(new Response('All my base are belong to us.'));
```

This currently *does not work* for only one reason: The Content-Type on  the ServiceWorker file is
not `application/javascript` but `text/plain`. If it was set to `application/javascript`, I could
do the following:

1. Make someone to visit https://gateway.ipfs.io/ipfs/QmaEtDG87K69mVR7sc48DemHpj3LqxKaQfDw6qJr2AGbmZ
2. My ServiceWorker would be installed
3. All subsequent requests to anything that starts with https://ipfs.io/ipfs/ will return ‘All my base are belong to us.’. Even worse, since the introduction of [Foreign-Fetch ServiceWorker](https://developers.google.com/web/updates/2016/09/foreign-fetch), even websites using the CORS-enabled `gateway.ipfs.io` endpoint would receive the same response.

Currently, the [Go FileServer implementation](https://golang.org/src/net/http/sniff.go), which the gateway [uses](https://github.com/ipfs/go-ipfs/blob/fa5c0d40cdd9a2524340b3ee5555e5f0536e42a8/core/corehttp/gateway_handler.go#L301), does not sniff JavaScript to set a `Content-Type: application/javascript` header. If they ever change that, or IPFS exposes control over `Content-Type`, the gateway would become vulnerable to the attack above.

I think ServiceWorker are powerful and they should be allowed on IPFS. ServiceWorker are scoped to
the path where the ServiceWorker file is located. Most of the time it would be located at paths
[...]

PR #4177, opened 2017-08-28
Set filename in Content-Disposition if filename=x is passed in URI query

Voker57:
Helps with referring to hashes by specific filename, eliminated unixfs node need for that

Example: http://ipfs.io/ipfs/Qmfoobar -- downloads Qmfoobar.zip (if mime was detected by first block)
http://ipfs.io/ipfs/Qmfoobar?filename=stuff.zip -- downloads stuff.zip

### DIFF
--- a/core/corehttp/gateway_handler.go
+++ b/core/corehttp/gateway_handler.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 	"os"
 	gopath "path"
 	"runtime/debug"
@@ -131,7 +132,6 @@ func (i *gatewayHandler) optionsHandler(w http.ResponseWriter, r *http.Request) 
 }
 
 func (i *gatewayHandler) getOrHeadHandler(ctx context.Context, w http.ResponseWriter, r *http.Request) {
-
 	urlPath := r.URL.Path
 	escapedURLPath := r.URL.EscapedPath()
 
@@ -266,7 +266,14 @@ func (i *gatewayHandler) getOrHeadHandler(ctx context.Context, w http.ResponseWr
 	}
 
 	if !dir {
-		name := gopath.Base(urlPath)
+		urlFilename := r.URL.Query().Get("filename")
+		var name string
+		if urlFilename != "" {
+			w.Header().Set("Content-Disposition", fmt.Sprintf("inline; filename*=UTF-8''%s", url.PathEscape(urlFilename)))
+			name = urlFilename
+		} else {
+			name = gopath.Base(urlPath)
+		}
 		i.serveFile(w, r, name, modtime, dr)
 		return
 	}
--- a/test/sharness/t0110-gateway.sh
+++ b/test/sharness/t0110-gateway.sh
@@ -31,6 +31,11 @@ test_expect_success "GET IPFS path succeeds" '
   curl -sfo actual "http://127.0.0.1:$port/ipfs/$HASH"
 '
 
+test_expect_success "GET IPFS path with explicit filename succeeds with proper header" "
+  curl -fo actual -D actual_headers 'http://127.0.0.1:$port/ipfs/$HASH?filename=testтест' &&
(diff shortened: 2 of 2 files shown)


core/corehttp/gateway_handler.go
```diff
@@ -132,7 +133,7 @@ func (i *gatewayHandler) optionsHandler(w http.ResponseWriter, r *http.Request)
 }
 
 func (i *gatewayHandler) getOrHeadHandler(ctx context.Context, w http.ResponseWriter, r *http.Request) {
-
+	log.Errorf("see me?")
```
lgierth: Remove stray debugging line

core/corehttp/gateway_handler.go
```diff
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	url "net/url"
```
lgierth: You can just do `"net/url"` here, the import name will be the same

crackcomm:
It works for `.zip` but `.pdf` or `.png` will need `attachment` to download.

Voker57:
That's out of scope of this PR: I just want files that are saved from link to be named properly.

crackcomm:
@Voker57 sounds good, I wanted to suggest a `download` query parameter but only after posting a comment :)

radfish:
@lgierth Is this waiting on any more changes? Is there a need for a branch based on MIME type? ('inline' for application/pdf, and 'attachment' for everything else) or is 'inline' ok to pass for all MIME types?

Without this feature linking to files is a big problem (I often specifically do not want wrapping directories, but a direct link). Luckily, for `application/pdf`, for whatever reason, Firefox seems to grab the filename if you append any query, even any parameter name at all, even verbatim `.../hash?foo=XYZ` opens in the PDF viewer and the title is set to XYZ and if you press the save button, the name is XYZ. Firefox does this, because the headers are identical between `.../hash?foo=XYZ` and `.../hash`. But, for other types, things are bad, no filename in the save dialog.

Stebalien:
> Is this waiting on any more changes?

Looks like it just dropped through the cracks.

> Is there a need for a branch based on MIME type? ('inline' for application/pdf, and 'attachment' for everything else) or is 'inline' ok to pass for all MIME types?

This PR doesn't appear to do any content sniffing, it just sets the appropriate file name.

Stebalien:
I've checked and this does *not* trigger https://github.com/ipfs/go-ipfs/issues/4025.

Stebalien:
@lgierth. I'd like to merge this. Any objections?

Stebalien:
I've added some documentation here: https://github.com/ipfs/go-ipfs/pull/5393

radfish:
Thank you for picking this up!

On Mon, Aug 20, 2018 at 03:05:29PM -0700, Steven Allen wrote:
> > Is there a need for a branch based on MIME type? ('inline' for application/pdf, and 'attachment' for everything else) or is 'inline' ok to pass for all MIME types?
> 
> This PR doesn't appear to do any content sniffing, it just sets the appropriate file name.

The current PR does not do any discrimination based on MIME type, but it
does choose one other parameter besides the filename: for all MIME types
it sets the 'inline' parameter,  as opposed to the 'attachement'
parameter.

The meaning of this parameter is documented in ContentDisposition docs,
but it's self-explanatory, 'inline' tells the viewer to render the
content within the same window (e.g. the browser pdf-viwer plugin for
application/pdf), vs 'attachment' tells the viewer to show a 'Save
As...' dialog.

Browsers probably ignore 'inline' (i.e. treat it as 'attachment') in
case no viewer is available for that type, (e.g. if you send 'inline,
application/zip' the browser will probably open Save As dialog), so
probably it is ok to send 'inline' for all types. Do you think this could
cause a problem with any browser or other client?

Stebalien:
According to the docs, "inline" is the default value so this shouldn't cause a problem. We should probably *also* add a `download=true` flag but that's a separate issue.

(thanks for bringing this to our attention, by the way)

PR #4184, opened 2017-08-30
Fix "ipfs ls" to once again work on arbitrary dag objects.

kevina:
Closes #4075

License: MIT
Signed-off-by: Kevin Atkinson <k@kevina.org>

PR #4192, opened 2017-08-31
Fix the publisher tests

Stebalien:
These passed when the PR was submitted (and tested) but stopped passing after I
unvendored testutil. Unfortunately, GitHub doesn't automatically re-test against
the latest master.

License: MIT


kevina:
I also included a fix in #4184.

Stebalien:
Ah. I'd just say merge that then (I don't think there's much, if any, discussion left on it).

PR #4194, opened 2017-08-31
various fixes for /ipfs fuse code

whyrusleeping:
Including:
- allow mount to work when running the daemon in offline mode
- allow /ipfs interface to handle raw nodes and sharded directories
- fix an issue in the dagservice where we were returning a blockservice not-found error

### DIFF
--- a/fuse/node/mount_unix.go
+++ b/fuse/node/mount_unix.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"strings"
+	"sync"
 
 	core "github.com/ipfs/go-ipfs/core"
 	ipns "github.com/ipfs/go-ipfs/fuse/ipns"
@@ -63,28 +64,26 @@ func doMount(node *core.IpfsNode, fsdir, nsdir string) error {
 	}
 
 	// this sync stuff is so that both can be mounted simultaneously.
-	var fsmount mount.Mount
-	var nsmount mount.Mount
-	var err1 error
-	var err2 error
+	var fsmount, nsmount mount.Mount
+	var err1, err2 error
 
-	done := make(chan struct{})
+	var wg sync.WaitGroup
 
+	wg.Add(1)
 	go func() {
+		defer wg.Done()
 		fsmount, err1 = rofs.Mount(node, fsdir)
-		done <- struct{}{}
 	}()
 
 	if node.OnlineMode() {
+		wg.Add(1)
 		go func() {
+			defer wg.Done()
 			nsmount, err2 = ipns.Mount(node, nsdir, fsdir)
-			done <- struct{}{}
 		}()
-
-		<-done
 	}
 
-	<-done
+	wg.Wait()
 
 	if err1 != nil {
 		log.Errorf("error mounting: %s", err1)
--- a/fuse/node/mount_unix.go
+++ b/fuse/node/mount_unix.go
@@ -75,12 +75,15 @@ func doMount(node *core.IpfsNode, fsdir, nsdir string) error {
 		done <- struct{}{}
 	}()
 
-	go func() {
-		nsmount, err2 = ipns.Mount(node, nsdir, fsdir)
-		done <- struct{}{}
-	}()
+	if node.OnlineMode() {
+		go func() {
+			nsmount, err2 = ipns.Mount(node, nsdir, fsdir)
+			done <- struct{}{}
+		}()
 
-	<-done
+		<-done
+	}
+
 	<-done
 
 	if err1 != nil {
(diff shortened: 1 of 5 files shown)


whyrusleeping:
I had some hacking time on the plane yesterday. Yay.

fuse/node/mount_unix.go
```diff
@@ -75,12 +75,15 @@ func doMount(node *core.IpfsNode, fsdir, nsdir string) error {
...
-		nsmount, err2 = ipns.Mount(node, nsdir, fsdir)
-		done <- struct{}{}
-	}()
+	if node.OnlineMode() {
+		go func() {
+			nsmount, err2 = ipns.Mount(node, nsdir, fsdir)
+			done <- struct{}{}
+		}()
+
+		<-done
```
Stebalien: Can we use a WaitGroup here while you're at it? This is a bit confusing at first glance.

fuse/readonly/readonly_unix.go
```diff
@@ -84,25 +94,37 @@ func (*Root) ReadDirAll(ctx context.Context) ([]fuse.Dirent, error) {
...
 }
 
 // Attr returns the attributes of a given node.
 func (s *Node) Attr(ctx context.Context, a *fuse.Attr) error {
 	log.Debug("Node attr")
+	if rawnd, ok := s.Nd.(*mdag.RawNode); ok {
+		a.Mode = 0444
+		a.Size = uint64(len(rawnd.RawData()))
+		a.Blocks = 1
+		a.Uid = uint32(os.Getuid()) // TODO: should probably cache these calls. No sense making multiple syscalls for each attr call here
```
Stebalien: Can we not just claim that these files are owned by root (does fuse not let us do that)?

fuse/readonly/readonly_unix.go
```diff
@@ -133,31 +155,54 @@ func (s *Node) Attr(ctx context.Context, a *fuse.Attr) error {
...
 		if len(n) == 0 {
-			n = link.Cid.String()
+			n = lnk.Cid.String()
 		}
-		entries[i] = fuse.Dirent{Name: n, Type: fuse.DT_File}
+		// TODO: calling everything a DT_File here might cause issues. But it
+		// will be expensive to query each child. However most shells call an
+		// additional 'stat' on each item in a directory listing, its probably
+		// okay.
+		entries = append(entries, fuse.Dirent{Name: n, Type: fuse.DT_File})
```
Stebalien: As someone who has written programs that rely on this being accurate... :rage:. We may want to start inlining info like this into directories (but that's not very elegant).

Stebalien: Also, if your argument is that most shells will call an additional stat, we might as well compute this up-front and cache it.

merkledag/merkledag.go
```diff
@@ -153,7 +153,10 @@ type sesGetter struct {
 
 func (sg *sesGetter) Get(ctx context.Context, c *cid.Cid) (node.Node, error) {
 	blk, err := sg.bs.GetBlock(ctx, c)
-	if err != nil {
+	switch err {
+	case bserv.ErrNotFound:
+		return nil, ErrNotFound
+	default:
```
Stebalien: You're missing the case where `err` is nil (this is what's causing the segfaults in CI).

fuse/readonly/readonly_unix.go
whyrusleeping: We can, I guess that does work.

fuse/readonly/readonly_unix.go
whyrusleeping: Yeah, lets go ahead and do that.

fuse/readonly/readonly_unix.go
```diff
@@ -133,31 +143,84 @@ func (s *Node) Attr(ctx context.Context, a *fuse.Attr) error {
...
+	}
+
+	var entries []fuse.Dirent
+	err = dir.ForEachLink(ctx, func(lnk *format.Link) error {
+		n := lnk.Name
 		if len(n) == 0 {
-			n = link.Cid.String()
+			n = lnk.Cid.String()
+		}
+		// TODO: calling everything a DT_File here might cause issues. But it
```
Stebalien: No longer relevant.

fuse/readonly/readonly_unix.go
```diff
@@ -133,31 +143,84 @@ func (s *Node) Attr(ctx context.Context, a *fuse.Attr) error {
...
+				case ftpb.Data_Symlink:
+					t = fuse.DT_Link
+				case ftpb.Data_Metadata:
+					log.Error("metadata object in fuse should contain its wrapped type")
+				default:
+					log.Error("unrecognized protonode data type: ", data.GetType())
+				}
+			}
+		default:
+			t = fuse.DT_Unknown
```
Stebalien: Maybe put this up top (i.e., `t := fuse.DT_Unknown`). We have several cases where we just log errors but don't assign anything (I assume it will default to unknown but that's not immediately obvious).

Stebalien:
:shipit:

PR #4195, opened 2017-09-01
config: command to apply profile after init

magik6k:
TODO:
* [x] Revert command
* [x] Tests
* [x] Docs

### DIFF
--- a/core/commands/config.go
+++ b/core/commands/config.go
@@ -142,6 +142,7 @@ Set the value of the 'Datastore.Path' key:
 		"show":    configShowCmd,
 		"edit":    configEditCmd,
 		"replace": configReplaceCmd,
+		"profile": configProfileCmd,
 	},
 }
 
@@ -288,6 +289,59 @@ can't be undone.
 		err = replaceConfig(r, file)
 		if err != nil {
 			res.SetError(err, cmdkit.ErrNormal)
+			return
+		}
+	},
+}
+
+var configProfileCmd = &cmds.Command{
+	Helptext: cmds.HelpText{
+		Tagline: "Apply profiles to config.",
+	},
+
+	Subcommands: map[string]*cmds.Command{
+		"apply": configProfileApplyCmd,
+	},
+}
+
+var configProfileApplyCmd = &cmds.Command{
+	Helptext: cmds.HelpText{
+		Tagline: "Apply profile to config.",
+	},
+	Arguments: []cmds.Argument{
+		cmds.StringArg("profile", true, false, "The profile to apply to the config."),
+	},
+	Run: func(req cmds.Request, res cmds.Response) {
+		args := req.Arguments()
+
+		profile, ok := config.ConfigProfiles[args[0]]
+		if !ok {
+			res.SetError(fmt.Errorf("%s in not a profile", args[0]), cmds.ErrNormal)
+			return
+		}
+
+		r, err := fsrepo.Open(req.InvocContext().ConfigRoot)
+		if err != nil {
+			res.SetError(err, cmds.ErrNormal)
+			return
+		}
+		defer r.Close()
+
+		cfg, err := r.Config()
+		if err != nil {
+			res.SetError(err, cmds.ErrNormal)
+			return
+		}
+
+		err = profile(cfg)
+		if err != nil {
+			res.SetError(err, cmds.ErrNormal)
+			return
+		}
+
+		err = r.SetConfig(cfg)
+		if err != nil {
+			res.SetError(err, cmds.ErrNormal)
 			return
 		}
 	},
--- a/core/commands/config.go
+++ b/core/commands/config.go
@@ -295,7 +295,7 @@ can't be undone.
 }
 
 var configProfileCmd = &cmds.Command{
-	Helptext: cmds.HelpText{
+	Helptext: cmdkit.HelpText{
 		Tagline: "Apply profiles to config.",
 	},
 
@@ -306,48 +306,48 @@ var configProfileCmd = &cmds.Command{
 }
 
 var configProfileApplyCmd = &cmds.Command{
-	Helptext: cmds.HelpText{
+	Helptext: cmdkit.HelpText{
 		Tagline: "Apply profile to config.",
 	},
-	Arguments: []cmds.Argument{
-		cmds.StringArg("profile", true, false, "The profile to apply to the config."),
+	Arguments: []cmdkit.Argument{
+		cmdkit.StringArg("profile", true, false, "The profile to apply to the config."),
 	},
 	Run: func(req cmds.Request, res cmds.Response) {
 		profile, ok := config.Profiles[req.Arguments()[0]]
 		if !ok {
-			res.SetError(fmt.Errorf("%s is not a profile", req.Arguments()[0]), cmds.ErrNormal)
+			res.SetError(fmt.Errorf("%s is not a profile", req.Arguments()[0]), cmdkit.ErrNormal)
 			return
 		}
 
 		err := transformConfig(req.InvocContext().ConfigRoot, profile.Apply)
 		if err != nil {
-			res.SetError(err, cmds.ErrNormal)
+			res.SetError(err, cmdkit.ErrNormal)
 			return
 		}
 	},
 }
 
 var configProfileRevertCmd = &cmds.Command{
-	Helptext: cmds.HelpText{
+	Helptext: cmdkit.HelpText{
 		Tagline: "Revert profile changes.",
 		ShortDescription: `Reverts profile-related changes to the config.
 
 Reverting some profiles may damage the configuration or not be possible.
 Backing up the config before running this command is advised.`,
 	},
-	Arguments: []cmds.Argument{
-		cmds.StringArg("profile", true, false, "The profile to apply to the config."),
+	Arguments: []cmdkit.Argument{
+		cmdkit.StringArg("profile", true, false, "The profile to apply to the config."),
 	},
 	Run: func(req cmds.Request, res cmds.Response) {
 		profile, ok := config.Profiles[req.Arguments()[0]]
 		if !ok {
-			res.SetError(fmt.Errorf("%s is not a profile", req.Arguments()[0]), cmds.ErrNormal)
+			res.SetError(fmt.Errorf("%s is not a profile", req.Arguments()[0]), cmdkit.ErrNormal)
 			return
 		}
 
 		err := transformConfig(req.InvocContext().ConfigRoot, profile.Revert)
 		if err != nil {
-			res.SetError(err, cmds.ErrNormal)
+			res.SetError(err, cmdkit.ErrNormal)
 			return
 		}
 	},
--- a/core/commands/config.go
+++ b/core/commands/config.go
@@ -324,13 +324,14 @@ var configProfileApplyCmd = &cmds.Command{
 			res.SetError(err, cmdkit.ErrNormal)
 			return
 		}
+		res.SetOutput(nil)
 	},
 }
 
 var configProfileRevertCmd = &cmds.Command{
 	Helptext: cmdkit.HelpText{
 		Tagline: "Revert profile changes.",
-		ShortDescription: `Reverts profile-related changes to the config.
+		ShortDescription: `Reverts profile-related changes to the default values.
 
 Reverting some profiles may damage the configuration or not be possible.
 Backing up the config before running this command is advised.`,
@@ -350,6 +351,7 @@ Backing up the config before running this command is advised.`,
 			res.SetError(err, cmdkit.ErrNormal)
 			return
 		}
+		res.SetOutput(nil)
 	},
 }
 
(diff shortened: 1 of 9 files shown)


whyrusleeping:
It would be nice to print out what changes are being made

core/commands/config.go
```diff
@@ -278,6 +279,89 @@ can't be undone.
...
+Reverting some profiles may damage the configuration or not be possible.
+Backing up the config before running this command is advised.`,
+	},
+	Arguments: []cmds.Argument{
+		cmds.StringArg("profile", true, false, "The profile to apply to the config."),
+	},
+	Run: func(req cmds.Request, res cmds.Response) {
+		args := req.Arguments()
+
+		profile, ok := config.Profiles[args[0]]
```
whyrusleeping: instead of binding it to `args` and just using the `[0]` on it, why not do:
```go
profilename := req.Arguments()[0]
```

core/commands/config.go
```diff
@@ -278,6 +279,88 @@ can't be undone.
...
+Backing up the config before running this command is advised.`,
+	},
+	Arguments: []cmds.Argument{
+		cmds.StringArg("profile", true, false, "The profile to apply to the config."),
+	},
+	Run: func(req cmds.Request, res cmds.Response) {
+
+		profile, ok := config.Profiles[req.Arguments()[0]]
+		if !ok {
+			res.SetError(fmt.Errorf("%s in not a profile", req.Arguments()[0]), cmds.ErrNormal)
```
whyrusleeping: s/in/is/

core/commands/config.go
```diff
@@ -278,6 +279,88 @@ can't be undone.
...
+		err := transformConfig(req.InvocContext().ConfigRoot, profile.Unapply)
+		if err != nil {
+			res.SetError(err, cmds.ErrNormal)
+			return
+		}
+	},
+}
+
+func transformConfig(configRoot string, transformer config.Transformer) error {
+	r, err := fsrepo.Open(configRoot)
```
whyrusleeping: if the daemon is running while the user runs this they might run into issues. You cant open the repo again

whyrusleeping: probably best to pass in the repo from the IpfsNode

repo/config/profile.go
```diff
@@ -1,56 +1,84 @@
...
+				"/ip4/192.0.0.170/ipcidr/32",
+				"/ip4/192.0.0.171/ipcidr/32",
+				"/ip4/192.0.2.0/ipcidr/24",
+				"/ip4/192.168.0.0/ipcidr/16",
+				"/ip4/198.18.0.0/ipcidr/15",
+				"/ip4/198.51.100.0/ipcidr/24",
+				"/ip4/203.0.113.0/ipcidr/24",
+				"/ip4/240.0.0.0/ipcidr/4",
+			}
+
```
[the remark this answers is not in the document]
magik6k: Done

core/commands/config.go
magik6k: That's what other commands here do. Looking at repo.Open it should be safe:
```go
func Open(repoPath string) (repo.Repo, error) {
	fn := func() (repo.Repo, error) {
		return open(repoPath)
	}
	return onlyOne.Open(repoPath, fn)
}
```

whyrusleeping: oh, i see. good point.

docs/config.md
```diff
@@ -267,3 +268,17 @@ Disables the p2p-circuit relay transport.
 - `EnableRelayHop`
 Enables HOP relay for the node. If this is enabled, the node will act as
 an intermediate (Hop Relay) node in relay circuits for connected peers.
+
+## Profiles
```
whyrusleeping: Hrm... i'm not sure this belongs here. Maybe put it at the top of the document? to distinguish it from just looking like another config field

magik6k: Moved it to the header, makes a bit more sense now

magik6k:
@whyrusleeping Is there anything else to do here? Getting this in would make https://github.com/ipfs/go-ipfs/pull/4154 a bit easier for me to test.

repo/config/profile.go
```diff
@@ -1,56 +1,88 @@
...
+			c.Addresses.NoAnnounce = append(c.Addresses.NoAnnounce, defaultServerFilters...)
+			c.Swarm.AddrFilters = append(c.Swarm.AddrFilters, defaultServerFilters...)
+			c.Discovery.MDNS.Enabled = false
+			return nil
+		},
+		Unapply: func(c *Config) error {
```
whyrusleeping: Maybe call it `Revert` instead of `Unapply`?

whyrusleeping:
One small comment, then LGTM

repo/config/profile.go
magik6k: Done

core/commands/config.go
```diff
@@ -278,6 +279,85 @@ can't be undone.
...
+			res.SetError(err, cmds.ErrNormal)
+			return
+		}
+	},
+}
+
+var configProfileRevertCmd = &cmds.Command{
+	Helptext: cmds.HelpText{
+		Tagline: "Revert profile changes.",
+		ShortDescription: `Reverts profile-related changes to the config.
```
kevina: This should be `Reverts profile-related changes to the default config.`.

docs/config.md
```diff
@@ -4,6 +4,20 @@ The go-ipfs config file is a json document. It is read once at node instantiatio
...
+- `server` profile
+Recommended for nodes with public IPv4 address, disables host and content
+discovery in local networks.
+
+- `test` profile
+Reduces external interference, useful for running ipfs in test environments.
+Note that with these settings node won't be able to talk to the rest of the
+network without manual bootstrap.
+
 ## Table of Contents
```
kevina: @whyrusleeping should we document the `badgerds` profile here but label it as experimental?

kevina:
I also like we should create backup files automatically.  We could call them: `.config.preapply-PROFILE` and `.config.prerevert.PROFILE`

@whyrusleeping others, thoughs?

whyrusleeping:
@kevina I do like the idea of making backups automatically. That will require some changes to the repo abstraction I think.

docs/config.md
whyrusleeping: Its somewhat out of scope, but probably a good idea.

kevina: Out of scope?

I meant to add under the new Profiles section in case that wasn't clear.

whyrusleeping: @kevina ah, gotcha. Yeah, lets get that in here.

magik6k:
* Rebased, updated to use new commands
* Did what @kevina suggested in the review
* Backups are todo, other than that that PR is mergable

kevina:
Looks good.  I would like to backups gets in with this P.R, but if @whyrusleeping is okay with doing that separately so am I.

magik6k:
Backups done

repo/mock.go
```diff
@@ -28,6 +28,10 @@ func (m *Mock) SetConfig(updated *config.Config) error {
 	return nil
 }
 
+func (m *Mock) BackupConfig(prefix string) (string, error) {
+	return "config-" + prefix + "-backup", nil
```
whyrusleeping: maybe this can just return `errTODO` like the other methods? If we arent actually going to back up the config, we probably should bother to pretend with the name

magik6k: done

docs/config.md
```diff
@@ -4,6 +4,33 @@ The go-ipfs config file is a json document. It is read once at node instantiatio
...
+  Note that with these settings node won't be able to talk to the rest of the
+  network without manual bootstrap.
+
+- `badgerds`
+
+  Replaces default datastore configuration with experimental badger datastore.
+  If you apply this profile after `ipfs init`, you will need to convert your
+  datastore to the new configuration. You can do this using [ipfs-ds-convert](https://github.com/ipfs/ipfs-ds-convert)
+
+  WARNING: badger datastore is experimantal. Make sure to backup your data
```
ForrestWeston: badger datastore is experim**e**ntal

docs/config.md
```diff
@@ -4,6 +4,33 @@ The go-ipfs config file is a json document. It is read once at node instantiatio
...
+#### Profiles
+Configuration profiles allow to tweak configuration quickly. Profiles can be
+applied with `--profile` flag to `ipfs init` or with `ipfs config profile apply`
+command.
+
+Available profiles:
+- `server`
+
+  Recommended for nodes with public IPv4 address, disables host and content
+  discovery in local networks.
```
Stebalien: Or those running on VPSs.

core/commands/config.go
```diff
@@ -293,6 +294,92 @@ can't be undone.
...
 }
 
+var configProfileCmd = &cmds.Command{
+	Helptext: cmdkit.HelpText{
+		Tagline: "Apply profiles to config.",
+	},
+
+	Subcommands: map[string]*cmds.Command{
+		"apply":  configProfileApplyCmd,
+		"revert": configProfileRevertCmd,
```
Stebalien: Actually reverting is impossible (for us). We're not reverting, we're applying the inverse of the profile. Do we really need to provide this feature?

We could, instead, provide inverse profiles: local-discovery and flatfs-ds.

I'm worried that a user will think that profiles act like drop-ins (that's how a lot of unix daemons work) and will believe  that revert will revert the apply.

repo/config/profile.go
```diff
@@ -1,56 +1,89 @@
...
+				"/ip4/192.0.0.171/ipcidr/32",
+				"/ip4/192.0.2.0/ipcidr/24",
+				"/ip4/192.168.0.0/ipcidr/16",
+				"/ip4/198.18.0.0/ipcidr/15",
+				"/ip4/198.51.100.0/ipcidr/24",
+				"/ip4/203.0.113.0/ipcidr/24",
+				"/ip4/240.0.0.0/ipcidr/4",
+			}
+
+			c.Addresses.NoAnnounce = append(c.Addresses.NoAnnounce, defaultServerFilters...)
```
Stebalien: We should probably try merging these instead of just appending.

repo/config/profile.go
```diff
@@ -1,56 +1,89 @@
...
+			c.Swarm.AddrFilters = append(c.Swarm.AddrFilters, defaultServerFilters...)
+			c.Discovery.MDNS.Enabled = false
+			return nil
+		},
+		Revert: func(c *Config) error {
+			c.Addresses.NoAnnounce = []string{}
```
Stebalien: If we do keep the revert command, we should only remove the ones we've added.

core/commands/config.go
kevina: I agree with @Stebalien this is not really a revert so we perhaps we should name it something else, not sure what, "unapply" comes to mind.

The idea of instead providing inverse profiles is also a good idea.

magik6k: I think I like the idea of inverse profiles better, I'll see how that looks

magik6k:
Refactored revert into inverse profiles, cc @Stebalien

test/sharness/t0021-config.sh
```diff
@@ -48,6 +48,33 @@ CONFIG_SET_JSON_TEST='{
...
+  test_expect_success "profile ${profile} changed something" '
+    ipfs config show >actual &&
+    test_must_fail test_cmp expected actual
+  '
+
+  test_expect_success "'ipfs config profile apply ${inverse_profile}' works" '
+    ipfs config profile apply '${inverse_profile}'
+  '
+
+  test_expect_success "config is back to previous state after ${profile} revert" '
```
kevina: This test is a bit misnamed now.

docs/config.md
```diff
@@ -4,6 +4,45 @@ The go-ipfs config file is a json document. It is read once at node instantiatio
 either for an offline command, or when starting the daemon. Commands that execute
 on a running daemon do not read the config file at runtime.
 
+#### Profiles
+Configuration profiles allow to tweak configuration quickly. Profiles can be
+applied with `--profile` flag to `ipfs init` or with `ipfs config profile apply`
+command.
```
kevina: I would mention that a backup is created and what it is called.

core/commands/config.go
```diff
@@ -293,6 +294,64 @@ can't be undone.
...
+		err := transformConfig(req.InvocContext().ConfigRoot, req.Arguments()[0], profile)
+		if err != nil {
+			res.SetError(err, cmdkit.ErrNormal)
+			return
+		}
+		res.SetOutput(nil)
+	},
+}
+
+func transformConfig(configRoot string, backupName string, transformer config.Transformer) error {
```
kevina: I would call `backupName` `configName`.

core/commands/config.go
```diff
@@ -293,6 +294,64 @@ can't be undone.
...
+	if err != nil {
+		return err
+	}
+
+	err = transformer(cfg)
+	if err != nil {
+		return err
+	}
+
+	_, err = r.BackupConfig(backupName + "-")
```
kevina: Nit. I would make this `"pre-" + configName + "-"` to make it clear that the backup before the config was applied.

whyrusleeping:
Finally getting this one in, Thanks @magik6k !

PR #4204, opened 2017-09-05
Remove some dead code

Stebalien:
Speak now or forever lose your code.

.
.
.

Unfortunately, you'll still be able to resurrect it via git. VC makes life
boring.

License: MIT


routing/mock/dht.go
```diff
@@ -1,37 +0,0 @@
-package mockrouting
```
Stebalien: So... I'm pretty sure we don't use this (pending CI). However, this, out of all this dead code, may actually be useful.

thirdparty/iter/iter.go
```diff
@@ -1,5 +0,0 @@
-package iter
```
Stebalien: This is **THE MOST USELESS** package **EVER** (since leftpad).

Issue #4002, opened 2017-06-21
Decide on indentation in sharness and cleanup existing files


Kubuxu:
Indentation in sharness files is a mess. Most of the files are using tabs (3000 lines), we have about 500 lines of double space in the front and 500 of 4 spaces in the front, there are 7k lines total.

We should decide on indentation style, clean it up and set up tooling to control it.

kevina:
Using tabs in go code is fine as that is the standard, but using tabs in shell files is a royal PITA for me (I use emacs).  I would be much happier if we use 4 spaces.

(will write more later).

kevina:
The standard for go is to use tab for indention and spaces for alignment.  This makes a lot of sense.  For example, it allows me to set the tab width to 4 without any ill effects.   For go code, it also well supported in Emacs as it has a very good go mode and what ever it doesn't get right gofmt will fix for you.

As far as I know this is not the case for shell files.  By default emacs will indent using 4 spaces but then replace 8 spaces with a tab character.  The replacing of 8 spaces with a tab character is all around bad so I (as I am sure lots of other people have) disabled this by setting `indent-tabs-mode` to `nil`.

I do not think it is possible to get emacs to use tabs just for indention for shell files.  I can set `sh-basic-offset` to `8` and then `indent-tabs-mode` to `t` to get close, but what is really going on is emacs is indenting with 8 spaces and then replacing the spaces with tab.  This could mess up alignment if the tab width is set to something other than 8.

Indenting within quotes for example:
```
test_expect_success "ipfs version succeeds" '
	ipfs version >version.txt
'
```

Is even more of a pain as emacs does not see that as shell code but rather simply text within quotes.  So to use tabs I have to use the `Ctrl-q <tab>` sequence.  This is a manual process that I put off while actively writing new tests.  Also manually inserting spaces is far easier than tabs.  For one thing it is easier to type (keep in mind that just <tab> never inserts a literal tab characters in emacs) and also because I can quickly fix things up afterwards using the `indent-rigidly` command.
 
Using 4 spaces (or maybe just 2) will make things easier for me as I am fighting emacs less.  I am not totally against using tabs, it just that it is not well supported, at least for emacs, and for me.

If people know of a better shell mode for emacs that addresses my concerns I will withdraw my objection.

vyzo:
fwiw I also use emacs with `indent-tabs-mode` set to `nil`, so I'd rather not have to deal with tabs in shell scripts. So can we please use spaces?

[...]

PR #4212, opened 2017-09-06
Fix indent in sharness tests

magik6k:
Fixes https://github.com/ipfs/go-ipfs/issues/4002, using tabs as it's easiest to check/change to something else.

No more stuff like this


kevina:
This was never really discussed.  Can we hold off on merging it?  See my comments on #4002.

magik6k:
Using 2 space indent now, should be redy to review.

magik6k:
fixed

chriscool:
Sorry to be late about this but see also my comment on #4002 about Sharness, Git and other projects.

PR #4215, opened 2017-09-06
Feat/ed25519 ident

whyrusleeping:
This supercedes #4076 (I couldnt easily push to it).

I think we have some harder problems to solve for this integration. The main pain point I'm seeing is that its not easy, given a public key, to convert it to an ID. It might make sense to add an optional field to the public key protobuf that specifies the method used to convert Key to peer.ID

cc @JustinDrake

### DIFF
--- a/cmd/ipfs/init.go
+++ b/cmd/ipfs/init.go
@@ -16,10 +16,13 @@ import (
 	namesys "github.com/ipfs/go-ipfs/namesys"
 	config "github.com/ipfs/go-ipfs/repo/config"
 	fsrepo "github.com/ipfs/go-ipfs/repo/fsrepo"
+	ci "gx/ipfs/QmaPbCnUMBohSGo3KnxEa2bHqyJVVeEEcwtqJAYxerieBo/go-libp2p-crypto"
 )
 
+// The default keypair is 2048-bit RSA
 const (
 	nBitsForKeypairDefault = 2048
+	keypairTypeDefault     = ci.RSA
 )
 
 var initCmd = &cmds.Command{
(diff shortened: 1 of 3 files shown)


JustinDrake:
Thanks @whyrusleeping for the PR! (Sorry for the delay, I am currently travelling in Israel with limited time and wifi, studying zkSTARKs with Eli Ben-Sasson. Back in the UK on Friday.)

At a glance the PR looks good to me. Did run the IPFS test suite with the flag to generate Ed25519-id identities?

> I think we have some harder problems to solve for this integration.

Are you talking about this PR in particular, or more generally?

> its not easy, given a public key, to convert it to an ID. It might make sense to add an optional field to the public key protobuf that specifies the method used to convert Key to peer.ID

Yes I think `IDFromPublicKey` (and `IDFromPrivateKey`) should have a second parameter to specify the hash type. If no hash type is provided (i.e. `-1`) then some sort of default behaviour kicks in. My suggestion for the default is to use `mh.ID` for Ed25519, and `mh.SHA2_256` for RSA. We may want to explicitly disallow certain combinations, e.g. `mh.ID` for RSA.

whyrusleeping:
cc @diasdavid @Stebalien @vyzo @Kubuxu 

Thoughts here? Would be really great to get this in asap for the open bazaar guys

diasdavid:
@whyrusleeping just to make sure that this whole endeavor is doing:

- Peers can have ed25519 based Identities
- Peers that have an ed25519 based Identity, won't have RSA
- SECIO needs to adapt
- Identify needs to adapt

Is this all?

vyzo:
Longer term we should strive to make ed25519 the default i think.

RubenKelevra:
@Stebalien I think this one can be closed. :)

#7579 set it even as default.

Stebalien:
Thanks!

PR #4219, opened 2017-09-08
docker: switch from musl to glibc, and simplify stuff

lgierth:
The fs-repo-migrations releases on dist.ipfs.io never worked with the go-ipfs docker image because it's based on musl. That means binaries linked against glibc don't work, and I manually built and patched in an additional musl-based tarball for each fs-repo-migrations release. This madness has to stop.

The Dockerfile now has two stages: build and assembly.
This allows for a full-fledged debian build container,
while still resulting in a super-thin glibc-based busybox image.

License: MIT

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:edge
+FROM golang:1.9-stretch
 MAINTAINER Lars Gierth <lgierth@ipfs.io>
 
 # There is a copy of this Dockerfile called Dockerfile.fast,
@@ -6,6 +6,51 @@ MAINTAINER Lars Gierth <lgierth@ipfs.io>
 #
 # Please keep these two Dockerfiles in sync.
 
+ENV GX_IPFS ""
+ENV SRC_DIR /go/src/github.com/ipfs/go-ipfs
+
+COPY . $SRC_DIR
+
+# Build the thing.
+RUN cd $SRC_DIR \
+  # Required for getting the HEAD commit hash via git rev-parse.
+  && mkdir .git/objects \
+  # Allows using a custom (i.e. local) IPFS API endpoint.
+  && ([ -z "$GX_IPFS" ] || echo $GX_IPFS > /root/.ipfs/api) \
+  # Build the thing.
+  && make build
+
+ENV SUEXEC_VERSION v0.2
+ENV TINI_VERSION v0.16.1
+RUN set -x \
+  # Get su-exec, a very minimal tool for dropping privileges
+  && cd /tmp \
+  && git clone https://github.com/ncopa/su-exec.git \
+  && cd su-exec \
+  && git checkout -q $SUEXEC_VERSION \
+  && make \
+  # Get tini, a very minimal init daemon for containers
+  && cd /tmp \
+  && wget -q -O tini https://github.com/krallin/tini/releases/download/$TINI_VERSION/tini \
+  && chmod +x tini
+
+# Get the TLS CA certificates, they're not provided by busybox.
+RUN apt-get install -y ca-certificates
+
+# Now comes the actual target image, which aims to be as small as possible.
+FROM busybox:1-glibc
+MAINTAINER Lars Gierth <lgierth@ipfs.io>
+
+# Get the ipfs binary, entrypoint script, and TLS CAs from the build container.
+ENV SRC_DIR /go/src/github.com/ipfs/go-ipfs
+COPY --from=0 $SRC_DIR/cmd/ipfs/ipfs /usr/local/bin/ipfs
+COPY --from=0 $SRC_DIR/bin/container_daemon /usr/local/bin/start_ipfs
+COPY --from=0 /tmp/su-exec/su-exec /sbin/su-exec
+COPY --from=0 /tmp/tini /sbin/tini
+COPY --from=0 /etc/ssl/certs /etc/ssl/certs
+
+# This shared lib (part of glibc) doesn't seem to be included with busybox.
+COPY --from=0 /lib/x86_64-linux-gnu/libdl-2.24.so /lib/libdl.so.2
 
 # Ports for Swarm TCP, Swarm uTP, API, Gateway, Swarm Websockets
 EXPOSE 4001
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -1,11 +1,46 @@
...
+  # Allows using a custom (i.e. local) IPFS API endpoint.
+  && ([ -z "$GX_IPFS" ] || echo $GX_IPFS > /root/.ipfs/api) \
+  # Build the thing.
+  && make build
+
+# Get the TLS CA certificates, they're not provided by busybox.
+RUN apt-get install -y ca-certificates
+
+# Now comes the actual target image, which aims to be as small as possible.
+FROM busybox:1-glibc
```
Kubuxu: I heard some info that multiple `FROM` directives might be discontinued but as they are in stable Docker and AFAIK the Dockerfile format is still in change-lock state it should be safe to use.

Kubuxu:
@lgierth docker sharness tests are failing. Can you take a look at this

bin/container_daemon
```diff
@@ -1,16 +1,7 @@
 #!/bin/sh
 set -e
-user=ipfs
 repo="$IPFS_PATH"
 
-if [ `id -u` -eq 0 ]; then
-  # ensure folder is writable
-  su-exec "$user" test -w "$repo" || chown -R -- "$user" "$repo"
```
kpcyrd: This is needed to fix permissions (#3744) if you run `docker run -v /does/not/exist/yet:/data/ipfs go-ipfs`, otherwise go-ipfs can't write to it's home and crashes. This is also the reason why there's no `USER ipfs` in the dockerfile itself, as the permissions need to be preserved and then dropped in the entrypoint. :)

Dockerfile
```diff
@@ -1,11 +1,46 @@
...
+COPY --from=0 $SRC_DIR/cmd/ipfs/ipfs /usr/local/bin/ipfs
+COPY --from=0 $SRC_DIR/bin/container_daemon /usr/local/bin/start_ipfs
+COPY --from=0 /etc/ssl/certs /etc/ssl/certs
+
+# This shared lib (part of glibc) doesn't seem to be included with busybox.
+COPY --from=0 /lib/x86_64-linux-gnu/libdl-2.24.so /lib/libdl.so.2
+
+# This installs a very simple program acting as the init process.
+# Makes sure signals are properly passed to the ipfs daemon process.
+ENV TINI_VERSION v0.16.1
```
kpcyrd: On debian, I'd recommend [dumb-init]. :)

[dumb-init]: https://packages.debian.org/stretch/dumb-init

Dockerfile.fast
```diff
@@ -1,55 +1,69 @@
...
+# Create the fs-repo directory and switch to a non-privileged user.
 ENV IPFS_PATH /data/ipfs
-ENV IPFS_LOGGING ""
-ENV GOPATH     /go
-ENV PATH       /go/bin:$PATH
-ENV SRC_PATH   /go/src/github.com/ipfs/go-ipfs
+RUN mkdir -p $IPFS_PATH \
+  && useradd -s /usr/sbin/nologin -d $IPFS_PATH -u 1000 -g 100 ipfs \
+  && chown 1000:100 $IPFS_PATH
+USER ipfs
```
kpcyrd: `USER ipfs` needs to be removed as well so the entrypoint has the permissions to correct the ownership. :)

Dockerfile
```diff
@@ -14,51 +49,18 @@ EXPOSE 5001
...
+# Create the fs-repo directory and switch to a non-privileged user.
 ENV IPFS_PATH /data/ipfs
-# The default logging level
-ENV IPFS_LOGGING ""
-# Golang stuff
-ENV GOPATH     /go
-ENV PATH       /go/bin:$PATH
-ENV SRC_PATH   /go/src/github.com/ipfs/go-ipfs
+RUN mkdir -p $IPFS_PATH && adduser -D -h $IPFS_PATH -u 1000 -g 100 ipfs
+USER ipfs
```
kpcyrd: here as well :)

Kubuxu:
Hmm, it goes beyond my head right now. I would either have to dig deep into how that Dockerfile works or IDK.

kpcyrd:
@Kubuxu the entrypoint is started as root, then the permissions are adjusted here:
https://github.com/ipfs/go-ipfs/blob/c14a995fab81f894371ed06ec989c32bdd628e3d/bin/container_daemon#L7-L8
After that, root permissions aren't needed anymore so the entrypoint is restarted as ipfs:
https://github.com/ipfs/go-ipfs/blob/c14a995fab81f894371ed06ec989c32bdd628e3d/bin/container_daemon#L9-L10
Afterwards, this part is executed with the ipfs user:
https://github.com/ipfs/go-ipfs/blob/c14a995fab81f894371ed06ec989c32bdd628e3d/bin/container_daemon#L13-L42

The most complicated part is the deprecation of the implicit daemon which was added about half a year ago so people start invoking the daemon with the new format. :)

lgierth:
Okay, removed that new test for now -- it wasn't able to clean up after itself :(

lgierth:
That OSX failure on travis looks unrelated, it's t0180-p2p (cc @magik6k https://api.travis-ci.org/jobs/275047939/log.txt?deansi=true)

magik6k:
Looks like 500ms is not enough for travis to launch a program and write a pidfile:
```
  ma-pipe-unidir --listen --pidFile=listener.pid send /ip4/127.0.0.1/tcp/10101 < test0.bin &

  go-sleep 500ms &&
  kill -0 $(cat listener.pid) &&

  ipfsi 1 p2p stream dial $PEERID_0 p2p-test /ip4/127.0.0.1/tcp/10102 2>&1 > dialer-stdouterr.log &&
  ma-pipe-unidir recv /ip4/127.0.0.1/tcp/10102 > client.out &&
  test ! -f listener.pid

cat: listener.pid: No such file or directory

kill: usage: kill [-s sigspec | -n signum | -sigspec] pid | jobspec ... or kill -l [sigspec]
```

PR #4223, opened 2017-09-08
Add documentation for datastore configs

whyrusleeping:
License: MIT

### DIFF
--- a/docs/datastores.md
+++ b/docs/datastores.md
@@ -1,0 +1,76 @@
+# Datastore Configuration Options
+
+This document describes the different possible values for the `Datastore.Spec`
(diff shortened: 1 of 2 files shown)


docs/datastores.md
```diff
@@ -0,0 +1,76 @@
+# Datastore Configuration Options
+
+This document describes the different possible values for the `Datastore.Spec`
+field in the ipfs configuration file.
+
+## flatfs
+Stores each key value pair as a file on the filesystem.
```
magik6k: Might be worth mentioning that flatfs should only be used for `/blocks` as it's `.Query` isn't complete.

PR #4226, opened 2017-09-13
Update badgerds to 0.3.0

magik6k:
Update badger to current master, fix `ipfs gc` bug.

NOTE: This breaks previous badger ds-es as badger bumped their manifest version. I'm not sure if there is a migration for that, if someone wants to rescue their data, convert using ds-convert to flatfs (you prorably will need to remove `QmdfTbBqBPQ7VNxZEYEj14VmRuZBkqFbiwReogJgS1zR1n` using `block rm`)

### DIFF
--- a/package.json
+++ b/package.json
@@ -460,9 +460,9 @@
     },
     {
       "author": "magik6k",
-      "hash": "QmRh9Udeo8qCe4XKCSA2NUunTZbFvbc5d44E5wTZPFf3Fg",
+      "hash": "QmUPQpMxu9BFE2MhmBkiMUBn4VTSuAAZZkX9LT699tQeiw",
       "name": "go-ds-badger",
-      "version": "0.2.1"
+      "version": "0.3.1"
     }
   ],
   "gxVersion": "0.10.0",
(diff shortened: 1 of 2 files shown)


magik6k:
@manishrjain I've updated badger to current master in https://github.com/ipfs/go-ds-badger/pull/7, but then noticed that it won't open 'manifest v1' databases.

* Is it a good idea to use latest badger master for ipfs (Is there a more stable version)?
* Are there going to be any more breaking changes in terms of on-disk format in coming time (without backwards compat)?

We would really like to use badger, it looks really promising, but at the same time we want some long-term stability. I'm fine with waiting for a more stable release of badger.

manishrjain:
We're aiming for a v1.0 release in a month. That would ensure that we don't make breaking changes. Also, it would provide a backup utility to ensure that one can upgrade from one version of Badger to next.

whyrusleeping:
Its probably best to update this now, hopefully there are no more breaking changes to badger, but in any case, we want to not be using the known old one for now.

manishrjain:
P.S. Relevant issue in Badger: https://github.com/dgraph-io/badger/issues/161

whyrusleeping:
In any case, could you provide mechanisms for migrating any on disk
changes?

manishrjain:
As I mentioned above, we are working on providing a backup and restore utility, which would be version agnostic.

PR #4231, opened 2017-09-14
[WIP] update go-stream-muxer

Stebalien:
Introduces a new Reset method on streams that kills both sides of the
connection. Close now officially just closes the write side (what it did
all along...)

* Also pull through shiny new go-multiplexer fixes.
* Also pull in go-reuseport update.

License: MIT Signed-off-by: Steven Allen <steven@stebalien.com>


p2p/registry.go
```diff
@@ -76,8 +78,8 @@ type StreamInfo struct {
 	RemotePeer peer.ID
 	RemoteAddr ma.Multiaddr
 
-	Local  io.ReadWriteCloser
-	Remote io.ReadWriteCloser
+	Local  manet.Conn
```
whyrusleeping: why are these different interfaces?

Stebalien: One is the local port we're listening on, the other is the stream to the other peer. In case of error, I want to be able to reset the stream so I need access to the reset method on `Stream`.

whyrusleeping: Ah, i see. Carry on

Stebalien:
Is this a known problem? https://ci.ipfs.team/blue/organizations/jenkins/go-ipfs/detail/PR-4231/3/pipeline#step-80-log-1085

The daemon seems to have just died...

PR #4232, opened 2017-09-15
bump version to 0.4.11-rc1, update changelog

whyrusleeping:
License: MIT

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,124 @@
 # go-ipfs changelog
 
+### 0.4.11-rc1 2017-09-14
+
+Ipfs 0.4.11 is a larger release that brings many long-awaited features and
+performance improvements. These include new datastore options, more efficient
+bitswap transfers, circuit relay support, ipld plugins and more! Take a look
+at the full changelog below for a detailed list of every change.
+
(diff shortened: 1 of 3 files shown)


magik6k:
Looks like much of the things in feature list is from 4.10 (shutdown, pin update, key rm, ...)

whyrusleeping:
Hrm... did i mess up the script?

kpcyrd:
Could you include a heads up towards the deprecation notice in the docker entrypoint?
https://github.com/ipfs/go-ipfs/blob/e2a3cd3b66992b0e043b51a0c3dda3fa5cacd2c9/bin/container_daemon#L25-L41
This was added about half a year ago and I'd like to change the behaviour to the new syntax in 0.4.12 :)
Basically, if somebody sees the deprecation notice in their logs, they're overwriting the arguments and have to set the first argument of the container to `daemon`. If there's no deprecation warning, they're compatible with the patch I want to apply. :)

whyrusleeping:
@magik6k looks like this is the problem:
```
LAST_TAG=$(git tag | grep -v -- '-rc' | grep 'v'| tail -n1)
```

0.4.10 comes before 0.4.9 when sorted.

PR #4233, opened 2017-09-16
Gateway renders pretty 404 pages if available

jphastings:
I noticed ipfs/ipfs#167 and thought this might be an elegant solution for allowing the gateway to show helpful information when requested files aren't found.

When a requested file isn't found `ipfs-404.html` is looked for in the same directory, rolling up through any of its parents, and displayed (without immutable cache headers) if present.

Run this patch locally and query the following for a demonstration:

* [Pretty 404 page](http://127.0.0.1:8080/ipfs/QmTLbqrzBgRCB2weXi1dmZiNZQW7wfUW4k3gRu7UTcXaoi/nope) at /nope
* [More specific pretty 404 page](http://127.0.0.1:8080/ipfs/QmTLbqrzBgRCB2weXi1dmZiNZQW7wfUW4k3gRu7UTcXaoi/deeper/nope) at /deeper/nope
* [Fallback pretty 404 page](http://127.0.0.1:8080/ipfs/QmTLbqrzBgRCB2weXi1dmZiNZQW7wfUW4k3gRu7UTcXaoi/nope/nope) at /nope/nope

### DIFF
--- a/core/corehttp/gateway_handler.go
+++ b/core/corehttp/gateway_handler.go
@@ -11,6 +11,7 @@ import (
 	gopath "path"
 	"regexp"
 	"runtime/debug"
+	"strconv"
 	"strings"
 	"time"
 
@@ -203,6 +204,10 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
 		webError(w, "ipfs resolve -r "+escapedURLPath, err, http.StatusServiceUnavailable)
 		return
 	default:
+		if i.servePretty404IfPresent(w, r, parsedPath) {
+			return
+		}
+
 		webError(w, "ipfs resolve -r "+escapedURLPath, err, http.StatusNotFound)
 		return
 	}
@@ -287,6 +292,10 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
 	}
 
 	if r.Method == http.MethodHead {
+		return
+	}
+
+	if i.servePretty404IfPresent(w, r, parsedPath) {
 		return
 	}
 
@@ -404,6 +413,36 @@ func (i *gatewayHandler) serveFile(w http.ResponseWriter, req *http.Request, nam
 
 	w = &statusResponseWriter{w}
 	http.ServeContent(w, req, name, modtime, content)
+}
+
+func (i *gatewayHandler) servePretty404IfPresent(w http.ResponseWriter, r *http.Request, parsedPath ipath.Path) bool {
+	resolved404Path, ctype, err := i.searchUpTreeFor404(r, parsedPath)
+	if err != nil {
+		return false
+	}
+
+	dr, err := i.api.Unixfs().Get(r.Context(), resolved404Path)
+	if err != nil {
+		return false
+	}
+	defer dr.Close()
+
+	f, ok := dr.(files.File)
+	if !ok {
+		return false
+	}
+
+	size, err := f.Size()
+	if err != nil {
+		return false
+	}
+
+	log.Debugf("using pretty 404 file for %s", parsedPath.String())
+	w.Header().Set("Content-Type", ctype)
+	w.Header().Set("Content-Length", strconv.FormatInt(size, 10))
+	w.WriteHeader(http.StatusNotFound)
+	_, err = io.CopyN(w, f, size)
+	return err == nil
 }
 
 func (i *gatewayHandler) postHandler(w http.ResponseWriter, r *http.Request) {
@@ -618,4 +657,46 @@ func getFilename(s string) string {
 		return ""
 	}
 	return gopath.Base(s)
+}
+
+func (i *gatewayHandler) searchUpTreeFor404(r *http.Request, parsedPath ipath.Path) (ipath.Resolved, string, error) {
+	filename404, ctype, err := preferred404Filename(r.Header.Values("Accept"))
+	if err != nil {
+		return nil, "", err
+	}
+
+	pathComponents := strings.Split(parsedPath.String(), "/")
+
+	for idx := len(pathComponents); idx >= 3; idx-- {
+		pretty404 := gopath.Join(append(pathComponents[0:idx], filename404)...)
+		parsed404Path := ipath.New("/" + pretty404)
+		if parsed404Path.IsValid() != nil {
+			break
+		}
+		resolvedPath, err := i.api.ResolvePath(r.Context(), parsed404Path)
+		if err != nil {
+			continue
+		}
+		return resolvedPath, ctype, nil
+	}
+
+	return nil, "", fmt.Errorf("no pretty 404 in any parent folder")
+}
+
+func preferred404Filename(acceptHeaders []string) (string, string, error) {
+	// If we ever want to offer a 404 file for a different content type
+	// then this function will need to parse q weightings, but for now
+	// the presence of anything matching HTML is enough.
+	for _, acceptHeader := range acceptHeaders {
+		accepted := strings.Split(acceptHeader, ",")
+		for _, spec := range accepted {
+			contentType := strings.SplitN(spec, ";", 1)[0]
+			switch contentType {
+			case "*/*", "text/*", "text/html":
+				return "ipfs-404.html", "text/html", nil
+			}
+		}
+	}
+
+	return "", "", fmt.Errorf("there is no 404 file for the requested content types")
 }
--- a/core/corehttp/gateway_test.go
+++ b/core/corehttp/gateway_test.go
@@ -235,6 +235,70 @@ func TestGatewayGet(t *testing.T) {
 	}
 }
 
+func TestPretty404(t *testing.T) {
+	ns := mockNamesys{}
+	ts, api, ctx := newTestServerAndNode(t, ns)
+	defer ts.Close()
+
+	f1 := files.NewMapDirectory(map[string]files.Node{
+		"ipfs-404.html": files.NewBytesFile([]byte("Custom 404")),
+		"deeper": files.NewMapDirectory(map[string]files.Node{
+			"ipfs-404.html": files.NewBytesFile([]byte("Deep custom 404")),
+		}),
+	})
+
+	k, err := api.Unixfs().Add(ctx, f1)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	host := "example.net"
+	ns["/ipns/"+host] = path.FromString(k.String())
+
+	for _, test := range []struct {
+		path   string
+		accept string
+		status int
+		text   string
+	}{
+		{"/ipfs-404.html", "text/html", http.StatusOK, "Custom 404"},
+		{"/nope", "text/html", http.StatusNotFound, "Custom 404"},
+		{"/nope", "text/*", http.StatusNotFound, "Custom 404"},
+		{"/nope", "*/*", http.StatusNotFound, "Custom 404"},
+		{"/nope", "application/json", http.StatusNotFound, "ipfs resolve -r /ipns/example.net/nope: no link named \"nope\" under QmcmnF7XG5G34RdqYErYDwCKNFQ6jb8oKVR21WAJgubiaj\n"},
+		{"/deeper/nope", "text/html", http.StatusNotFound, "Deep custom 404"},
+		{"/deeper/", "text/html", http.StatusNotFound, "Deep custom 404"},
+		{"/deeper", "text/html", http.StatusNotFound, "Deep custom 404"},
+		{"/nope/nope", "text/html", http.StatusNotFound, "Custom 404"},
+	} {
+		var c http.Client
+		req, err := http.NewRequest("GET", ts.URL+test.path, nil)
+		if err != nil {
+			t.Fatal(err)
+		}
+		req.Header.Add("Accept", test.accept)
+		req.Host = host
+		resp, err := c.Do(req)
+
+		if err != nil {
+			t.Fatalf("error requesting %s: %s", test.path, err)
+		}
+
+		defer resp.Body.Close()
+		if resp.StatusCode != test.status {
+			t.Fatalf("got %d, expected %d, from %s", resp.StatusCode, test.status, test.path)
+		}
+		body, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			t.Fatalf("error reading response from %s: %s", test.path, err)
+		}
+
+		if string(body) != test.text {
+			t.Fatalf("unexpected response body from %s: got %q, expected %q", test.path, body, test.text)
+		}
+	}
+}
+
 func TestIPNSHostnameRedirect(t *testing.T) {
 	ns := mockNamesys{}
 	ts, api, ctx := newTestServerAndNode(t, ns)


whyrusleeping:
Cool feature. cc @lgierth @diasdavid

whyrusleeping:
As for tests, you probably want to add something in core/corehttp/gateway_test.go, some of the tests in there construct directories, in particular TestIPNSHostnameRedirect creates a directory with an index.html to test that working properly. I imagine a test for this 404 feature would be quite similar.

jphastings:
Thanks for the advice @whyrusleeping, hopefully this covers the new cases. Shout if there's anything else I can do to help!

core/corehttp/gateway_test.go
```diff
@@ -173,6 +173,71 @@ func TestGatewayGet(t *testing.T) {
...
+		t.Fatal(err)
+	}
+
+	_, dagDeep404, err := coreunix.AddWrapped(n, strings.NewReader("Deep custom 404"), "404.html")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	dagRoot.(*dag.ProtoNode).AddNodeLink("deeper", dagDeep404)
+	if err != nil {
```
whyrusleeping: theres no error being set here

jphastings: Good point! I've fixed this in `TestIPNSHostnameRedirect` too.

whyrusleeping:
Thinking about this, what is the need for not setting a cache? Everything is content addressed, right? There should be no need not to cache.

@dignifiedquire @VictorBjelkholm I think you guys might be interested in this PR

Kubuxu:
> Thinking about this, what is the need for not setting a cache? Everything is content addressed, right? There should be no need not to cache.

Choosing ETag for the cache is not simple in this case as the `404.html `is taken for reverse recursive walk but I think using CID of the resolved `404.html` should work but I am not 100% sure.

jphastings:
@whyrusleeping I'm thinking of the case where I request a non-existant file from the gateway today; I don't want my browser's cache to be prevented from trying again, say, tomorrow when that DNSlink entry/peerID points to a new hash where that file now exists.

Right now this PR ensures the gateway sets the ETag of a requested but missing file to be the Cid of the pretty 404 page that was found (as @Kubuxu suggests). Smart browsers can then include it in the `If-None-Match` header of their next request to the gateway for this file which will [avoid sending unnecessary data](https://github.com/ipfs/go-ipfs/blob/70d662994064c72f52f4617e49e96b2c36446fa3/core/corehttp/gateway_handler.go#L200-L204). I reckon this is all we need to do here to allow smart clients to request transfer of the bare minimum to show the most recent data.

Does this address your concern?

diasdavid:
> Thinking about this, what is the need for not setting a cache? Everything is content addressed, right? There should be no need not to cache.

What if the gateway fails to find the file, timeouts and returns a 404 with cache-control: unlimited? An existing hash->file can be cached forever and the corollary is that a non-existing hash->file will someday become an existing hash->file.

VictorBjelkholm:
If you load `ipfs.io/ipfs/$HASH/$FILE` and `$FILE` isn't found, a redirect to a 404 page under `$HASH` would always be a 404. 

As both Etag and Cache-Control uses the URL, it has to be a redirect and if it redirects to `$HASH/404.html`, that redirect would always happen in the future.

Introducing IPNS would make things work differently though, and probably in that case it's best to dump the Cache-Control header. However, if IPFS is used with IPNS, the 404 should be permanent.

> What if the gateway fails to find the file, timeouts and returns a 404 with cache-control: unlimited?

In the case of `ipfs.io/ipfs/$HASH/$FILE`, don't we already see all the links of `$HASH` and if we cannot see the match for $FILE, we can safely show the 404 page?

ivan386:
> As both Etag and Cache-Control uses the URL, it has to be a redirect and if it redirects to $HASH/404.html, that redirect would always happen in the future.

No redirect for 404 page please. I need to know original url where 404 happen.  Then i can redirect to right location by Javascript. 

I think about [.htaccess](https://en.wikipedia.org/wiki/.htaccess) for ipfs.

jphastings:
There's no redirecting happening at the moment. I believe @VictorBjelkholm was talking about internally, ie. the data is being delivered when a missing file is requested.

> In the case of ipfs.io/ipfs/$HASH/$FILE, don't we already see all the links of $HASH and if we cannot see the match for $FILE, we can safely show the 404 page?

I agree — in this case we know that `$FILE` definitely doesn't exist under the `$HASH` and we can safely deliver the data from an appropriate `404.html` page from that URL provided we also set the cache headers to be non-permanent and the ETag to be of the data delivered (ie. the `404.html` page) so that if `$FILE` becomes present (in the case where we're addressing via an IPNS link) the new data can be retrieved.

Further to this, in the case of `ipfs.io/ipfs/$HASH/$FILE` (ie. _not_ IPNS), we could also deliver the data with an immutable cache header, as the contents of the `$HASH` will never change. If we like the idea of this I can change [this conditional](https://github.com/ipfs/go-ipfs/pull/4233/files#diff-19ceb36d78bde310c54cc44a95bec44cR274) so that it allows setting the immutable header for `/ipfs` paths with 404s. In fact, can't we set the immutable cache header for any `/ipfs` path, including for directories (which is [not currently the case on HEAD](https://github.com/ipfs/go-ipfs/pull/4233/files#diff-19ceb36d78bde310c54cc44a95bec44cL262))? The only argument against this is that we are also declaring this business logic to be immutable (if we ever decided to move from `404.html` to `404_error.html` as the page too look for what a customer sees would be dependent on which version of IPFS they first viewed that URL with).

Because of this I'm up for leaving the 404.html page _without_ the immutable cache header, but leave the ETag in place, as it stands. It's extremely unlikely we'd ever need to change the `404.html` to something else, but better to not force immutability on the business logic at this stage I feel!

mitra42:
I think doing this at the Gateway is the wrong place, because if I've understood this correctly, its indicating to the caller (which might be an application, not a browser) that its retrieved the page. Shouldnt this be part of the logic in whatever is calling the Gateway (browser) to decide what to do when the page isn't found. 

For example ... I can see situations where if the file can't be delivered by the Gateway, the app decides to go to somewhere else (such as an Archive) to try a different way to retrieve the same content.

Stebalien:
@mitra42 The gateway will still return a 404 status so the application can decide what to do with it.

However, I believe this should be checking the client's Accept header. If the client doesn't want an HTML page, we shouldn't be returning an HTML 404 page.

jphastings:
That's an excellent idea @Stebalien; only if the request header matches `text/html` should this functionality be present - I'll add that to this PR now!

@mitra42 while I agree with your point in pure terms, my understanding of the Gateway's purpose is  that it should bring some of the benefits of IPFS to browsers which don't have native IPFS support, which leads me to two counter arguments:

- If a browser had native support for IPFS then this could be implemented in that plugin, but there isn't one of those (yet!) and the gateway is the bit of software that is the first step in that direction.
- If browsers were mostly used for displaying non-dynamically generated content then this is a proposal which could be put into effect there (so that it would benefit non-IPFS sourced sites too), but this is more easily implemented "on the server" in the current world of HTTP (where a centralised server is a concept which doesn't exist in IPFS land!) In fact the closest analogue to "on the server" in the world of an IPFS-HTTP bridge _is_ the gateway.

I hope that makes sense!

mitra42:
Agreed,

jphastings:
It might be worth noting that if a person wishing to use this feature wants their pretty 404 page's assets to render correctly without specifying absolute paths (eg. if they want the 404s to work with accessing the file from an IPFS or IPNS URL like `gateway.server/ipfs/HASH/missing_resource` _as well_ as from a DNS link like `example.com/missing_resource`) then some logic would need to be applied to give the browser enough information to know where to get assets from. I've found this snippet works well, though it will need to be tailored to how assets are set up:

```html
<head>
  <!-- things that don't refernce assets -->
  <script>
    const ipfsPath = /^\/ip[fn]s\/(.+?)\//.exec(window.location.pathname);
    const basePath = (ipfsPath == null) ? '/' : ipfsPath[0];
    document.write('<base href="' + basePath + '"/>');
  </script>
  <!-- things that reference assets, eg. CSS and JS -->
</head>
```

jphastings:
@whyrusleeping is there anything further you'd like me to alter on this PR?

whyrusleeping:
This LGTM on a technical level. Now we need to get a signoff from the others on whether or not this is something we really want.

cc @diasdavid @dignifiedquire

core/corehttp/gateway_handler.go
```diff
@@ -598,3 +610,42 @@ func webErrorWithCode(w http.ResponseWriter, message string, err error, code int
...
+		parsedPath, err := coreapi.ParsePath(pretty404)
+		if err == nil {
+			resolvedPath, err := i.api.ResolvePath(ctx, parsedPath)
+			if err == nil {
+				return resolvedPath, nil
+			}
+		}
+	}
+
+	return nil, errors.New("No pretty 404 in any parent folder")
```
Kubuxu: We have general policy that errors should start with lower case (it looks better when they are being cascaded).

core/corehttp/gateway_handler.go
```diff
@@ -598,3 +610,42 @@ func webErrorWithCode(w http.ResponseWriter, message string, err error, code int
...
+		accepted := strings.Split(acceptHeader, ",")
+		for _, spec := range accepted {
+			contentType := strings.Split(spec, ";")[0]
+			if contentType == "*/*" || contentType == "text/*" || contentType == "text/html" {
+				return "404.html", nil
+			}
+		}
+	}
+
+	return "", errors.New("There is no 404 file for the requested content types")
```
Kubuxu: Same here.

jphastings:
Thanks @Kubuxu, I've made those changes.

ivan386:
Maybe in some file like [.htaccess](https://en.wikipedia.org/wiki/.htaccess) will be set right file name and content type for 404,index and other special pages? And we can use it for other site settings in future.

jphastings:
@ivan386 `.htaccess` files are extremely complex in what they support, using that format for describing this single file would introduce confusion as to why other `.htaccess` features weren't supported.

@magik6k the use of `404.html` seems conventional ([Jekyll](https://jekyllrb.com/tutorials/custom-404-page/), [terraform](https://www.terraform.io/404.html), [most](https://github.com/paularmstrong/still#error-pages) [static](https://surge.sh/help/adding-a-custom-404-not-found-page) [site](https://github.com/cryogen-project/cryogen#deploying-your-site) [generators](https://github.com/fredjean/middleman-s3_sync#usage)) but it does seem like it might be annoying to have to rename pages on something as intra-linked as wikipedia.

I'd be up for `.404.html` as that seems close to the conventions, but is even more unlikely to be used in the normal course of a website's design.

Kubuxu:
Problem I see with naming it just `404.html` is that someone might just have a page named like this. This would be the case with english dump of wikipedia and this: https://en.wikipedia.org/wiki/404 page.

jphastings:
@Kubuxu Yeah, this makes sense to me. I've changed the file to `.404.html` (with the preceding dot) which should make it explicit enough to not clash in all but the most obscure of cases.

ivan386:
Ok. Let's use JSON.
.site.conf
```
{
  "ErrorDocument": {
    "404": ".404.html"
  }
}
```

jphastings:
I'm not sure this PR benefits from the added complexity you're proposing @ivan386. Using a file (an `.htaccess` file or a `.site.conf` file or any other format) to specify which files to download and return when certain conditions are met means that the IPFS gateway needs to retrieve an extra file, as well as parsing it and coping with any errors in both those situations.

I'm a firm believer that the best code is the code you don't need to write; the code I've submitted here is as concise as I think I can make it while still returning a pretty 404 document, meaning fewer concepts for the next engineer to wrap their head around or augment in the future. For example, if we had a `.site.conf` file format, then wanted to introduce new features for it, the gateway code would have to be backward compatible and would grow yet further in size and complexity.

The benefits are limited too, as `index.html` is already returned when a directory is requested and in the case where a 504 would be returned there's a strong likelihood that the files within the specified multihash are less available, which means the likelihood of retrieving either your proposed `.site.conf` file or the linked `.504.html` would also be much lower.

jphastings:
@dignifiedquire or @diasdavid — have you had a chance to review this PR? Summary so far:

The PR's description remains accurate: this PR implements functionality within the gateway such that if a file is requested over HTTP that doesn't exist within an IPFS directory over, `text/html` is `Accept`able, and `.404.html` exists in this or a parent directory, then it is returned with suitable caching and HTTP status code.

This accompanies the existing `/` => `index.html` functionality to allow IPFs directories to present a friendlier interface to humans.

Other comments on this thread include verifying the approach and changing to `.404.html` as the magic filename, as it's unlikely to clash with anything else. Please let me know if there's anything I can do to help.

core/corehttp/gateway_handler.go
```diff
@@ -626,3 +638,42 @@ func webErrorWithCode(w http.ResponseWriter, message string, err error, code int
 func internalWebError(w http.ResponseWriter, err error) {
 	webErrorWithCode(w, "internalWebError", err, http.StatusInternalServerError)
 }
+
+func searchUpTreeFor404(ctx context.Context, acceptHeaders []string, parsedPath coreiface.Path, i *gatewayHandler) (coreiface.Path, error) {
```
lgierth: This should be a function on the gatewayHandler struct:
```
-func searchUpTreeFor404(ctx context.Context, acceptHeaders []string, parsedPath coreiface.Path, i *gatewayHandler) (coreiface.Path, error) {
+func (i *gatewayHandler) searchUpTreeFor404(ctx context.Context, acceptHeaders []string, parsedPath coreiface.Path) (coreiface.Path, error) {
```

core/corehttp/gateway_handler.go
```diff
@@ -626,3 +638,42 @@ func webErrorWithCode(w http.ResponseWriter, message string, err error, code int
...
+	for idx := len(pathComponents); idx >= 3; idx-- {
+		pretty404 := gopath.Join(append(pathComponents[0:idx], filename404)...)
+		parsedPath, err := coreapi.ParsePath(pretty404)
+		if err == nil {
+			resolvedPath, err := i.api.ResolvePath(ctx, parsedPath)
+			if err == nil {
+				return resolvedPath, nil
+			}
+		}
+	}
```
lgierth: Working with paths will generally be simpler soon, there's a few changes to `coreiface.Path` I have in mind.

core/corehttp/gateway_handler.go
```diff
@@ -626,3 +638,42 @@ func webErrorWithCode(w http.ResponseWriter, message string, err error, code int
...
+func searchUpTreeFor404(ctx context.Context, acceptHeaders []string, parsedPath coreiface.Path, i *gatewayHandler) (coreiface.Path, error) {
+	filename404, err := preferred404Filename(acceptHeaders)
+	if err != nil {
+		return nil, err
+	}
+
+	pathComponents := strings.SplitAfter(gopath.Dir(parsedPath.String()), "/")
+
+	for idx := len(pathComponents); idx >= 3; idx-- {
+		pretty404 := gopath.Join(append(pathComponents[0:idx], filename404)...)
```
lgierth: I have a hunch this way of walking up the tree has a couple of edge cases, but most of them are fine for now.

Could you add a test for one specific case though? Suppose we request `/foo/bar/`, then `/foo/bar/.ipfs/404.html` should be the first path that's tried. I *think* that it should already be doing that based on `gopath.Dir("/foo/bar/") == "/foo/bar"`, but better make sure.

lgierth: Sorry I meant `/foo/bar/ipfs-404.html` here ^

lgierth:
About caching of 404 responses: I think we should cache these in the long term, but there are a few edge cases around directories and redirects that we should figure out first. Until then I'd like to be cautious with caching.

core/corehttp/gateway_handler.go
jphastings: Hey @lgierth, I'm not sure what test you're suggesting; if we request a directory (eg. `/foo/bar/`) then the index page will be returned instead of this new 404 page logic being called. I've updated this PR to meet your other comments, but if you can help me understand your concern here I'll have another look! :)

jphastings:
Thanks for the careful review @lgierth; my apologies for the stretch between your reply and my changes. I've made the alterations you suggested (`ipfs-404.html`, making `searchUpTreeFor404` a method on `GatewayHandler`), but I've not figured out what you were suggesting on the third — I'll be happy to make changes if you're up for giving some more information.

As to your point about caching, though the first line of the diff seems like I'm altering caching functionality, that's only present as I've separated the concerns of caching (now done through the `preventCache` instead of the `dir` variable) and directory listing (still operating via `dir`) — let me know if my understanding of what's going on is wrong!

jphastings:
I just noticed that rendering a 404 will output `http: multiple response.WriteHeader calls` to stderr. This is because `(*gatewayHandler).serveFile` calls `http.ServeContent`, which always calls `w.WriteHeader(http.StatusOk)` (which is after the 404 is header is set).

This doesn't impact the status code delivered, but it is annoying. I don't want to change how `(*gatewayHandler).serveFile` works in this PR (it's already been open a fair while!) I'll be happy to have a look at changing how that works in a second PR once this is merged.

core/corehttp/gateway_handler.go
jphastings: I believe the tests now cover the case you're talking about @lgierth; [this line](https://github.com/ipfs/go-ipfs/pull/4233/files#diff-56f9b8538e5c90481066b55b7feaf780R273) covers that case ☺️

jphastings:
This PR is finally green again; I've updated the tests to use the new helpers and tended to the comments which @lgierth left. Is there anything further you'd like me to do here?

jphastings:
Hey folks, I've fixed the conflicts with master — can I get some feedback on why this PR isn't being addressed?

Stebalien:
Thank you for the reminder, this fell off our radar. I've added it to the go-ipfs 0.6.0 milestone so we don't forget it.

core/corehttp/gateway_handler.go
```diff
@@ -199,8 +199,15 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
...
 	default:
-		webError(w, "ipfs resolve -r "+escapedURLPath, err, http.StatusNotFound)
-		return
+		resolved404Path, err404 := i.searchUpTreeFor404(r, parsedPath)
+		if err404 != nil {
+			webError(w, "ipfs resolve -r "+escapedURLPath, err, http.StatusNotFound)
+			return
+		}
+
+		w.WriteHeader(http.StatusNotFound)
```
Stebalien, line 208: * This will cause us to try to write the status multiple times. Go doesn't like that.
* This will drop any further headers and log a bunch of errors.

We should probably set an "is404" boolean, and check it later.

core/corehttp/gateway_handler.go
```diff
@@ -602,3 +623,42 @@ func getFilename(s string) string {
 	}
 	return gopath.Base(s)
 }
+
+func (i *gatewayHandler) searchUpTreeFor404(r *http.Request, parsedPath ipath.Path) (ipath.Resolved, error) {
+	filename404, err := preferred404Filename(r.Header["Accept"])
```
Stebalien, line 628:
```suggestion
	filename404, err := preferred404Filename(r.Header.Get("Accept"))
```

core/corehttp/gateway_handler.go
```diff
@@ -602,3 +623,42 @@ func getFilename(s string) string {
...
+	if err != nil {
+		return nil, err
+	}
+
+	pathComponents := strings.SplitAfter(parsedPath.String(), "/")
```
Stebalien, line 633: SplitAfter is probably not the function you're looking for. When we re-join that, we'll end up with two slashes.

core/corehttp/gateway_handler.go
```diff
@@ -602,3 +623,42 @@ func getFilename(s string) string {
...
+
+	for idx := len(pathComponents); idx >= 3; idx-- {
+		pretty404 := gopath.Join(append(pathComponents[0:idx], filename404)...)
+		parsed404Path := ipath.New(pretty404)
+		if err := parsed404Path.IsValid(); err == nil {
```
Stebalien, line 638: We should just break at this point.

core/corehttp/gateway_handler.go
```diff
@@ -602,3 +623,42 @@ func getFilename(s string) string {
...
+
+func preferred404Filename(acceptHeaders []string) (string, error) {
+	// If we ever want to offer a 404 file for a different content type
+	// then this function will need to parse q weightings, but for now
+	// the presence of anything matching HTML is enough.
+	for _, acceptHeader := range acceptHeaders {
+		accepted := strings.Split(acceptHeader, ",")
+		for _, spec := range accepted {
+			contentType := strings.Split(spec, ";")[0]
+			if contentType == "*/*" || contentType == "text/*" || contentType == "text/html" {
```
Stebalien, line 657:
```suggestion
			switch contentType {
			case "*/*", "text/*", "text/html":
```

core/corehttp/gateway_handler.go
```diff
@@ -602,3 +623,42 @@ func getFilename(s string) string {
...
+}
+
+func preferred404Filename(acceptHeaders []string) (string, error) {
+	// If we ever want to offer a 404 file for a different content type
+	// then this function will need to parse q weightings, but for now
+	// the presence of anything matching HTML is enough.
+	for _, acceptHeader := range acceptHeaders {
+		accepted := strings.Split(acceptHeader, ",")
+		for _, spec := range accepted {
+			contentType := strings.Split(spec, ";")[0]
```
Stebalien, line 656:
```suggestion
			contentType := strings.SplitN(spec, ";", 1)[0]
```

core/corehttp/gateway_handler.go
Stebalien: Also, if 404.html happens to be a directory, strange things will happen.

core/corehttp/gateway_handler.go
```diff
@@ -286,6 +293,20 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
...
 	}
 
+	// Render a 404 instead of a directory listing, if one exists
+	if resolved404Path, err404 := i.searchUpTreeFor404(r, parsedPath); err404 == nil {
+		if dr, err := i.api.Unixfs().Get(r.Context(), resolved404Path); err == nil {
+			defer dr.Close()
+
+			if f, ok := dr.(files.File); ok {
+				w.WriteHeader(http.StatusNotFound)
+				i.serveFile(w, r, "ipfs-404.html", modtime, f)
```
Stebalien, line 303: In general, we should not use `serveFile` for these 404 responses:

* We should hard-code text/html as the mime type.
* We should not apply the `Range` header to 404 responses.

(and again, go doesn't like it when you call `WriteHeader` multiple times).

core/corehttp/gateway_handler.go
Stebalien: Ah, now I see. Let's call `r.Header.Values` (go 1.14). I'll go ahead and update our CI.

jphastings:
Thanks for all the great feedback @Stebalien! I've been writing Go at work since about a month after I learned it to write this PR, I'll get my head back in this PR this weekend and address your points.

Stebalien:
No, thank you and sorry this dropped off our radar. Feel free to bug us (or just ping me directly) if this happens again.

And take your time. We're now on a fixed release schedule so if it doesn't make it into the next release, it'll make it into the following.

core/corehttp/gateway_handler.go
```diff
@@ -406,6 +415,41 @@ func (i *gatewayHandler) serveFile(w http.ResponseWriter, req *http.Request, nam
...
+	content := &lazySeeker{
+		size:   size,
+		reader: f,
+	}
+
+	log.Debugf("using pretty 404 file for %s", parsedPath.String())
+	w.Header().Set("Content-Type", ctype)
+	w.Header().Set("Content-Length", strconv.FormatInt(size, 10))
+	w.WriteHeader(http.StatusNotFound)
+	_, err = io.CopyN(w, content, size)
```
jphastings, line 449: Copying content over manually as `http.ServeContent` unpreventably sets its own StatusCode header.

jphastings:
This should cover the changes you suggested @Stebalien ☺️

core/corehttp/gateway_handler.go
```diff
@@ -290,6 +295,10 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
 		return
```
Stebalien, line 295: Technically, if a 404 page is present, we should return 404 here. But I'm fine leaving that edge-case unfixed.

core/corehttp/gateway_handler.go
```diff
@@ -406,6 +415,41 @@ func (i *gatewayHandler) serveFile(w http.ResponseWriter, req *http.Request, nam
...
+
+	size, err := f.Size()
+	if err != nil {
+		return false
+	}
+
+	content := &lazySeeker{
+		size:   size,
+		reader: f,
+	}
```
Stebalien, line 443: This shouldn't be necessary. `ServeContent` needs this because it seeks to determine the file size.

Stebalien: (fixed)

Stebalien:
:partying_face: 

Thanks for your patience on this, this has been a _long_ time coming.

Lekensteyn:
If I am not mistaken, this change implies a lookup for `ipfs-404.html` on every attempt to obtain a directory listing, and subsequently breaks directory listings if the file is found. Is that intentional?

In core/corehttp/gateway_test.go, I *think* that the following tests are wrong:
```go
{"/deeper/", "text/html", http.StatusNotFound, "Deep custom 404"},
{"/deeper", "text/html", http.StatusNotFound, "Deep custom 404"},
```

These and `/` should probably end up with a directory listing.

jphastings:
That was intentional by me — the approach that web servers like NGinx take is `index.html` goes first, a 404 page goes second, then a directory listing if there's no 404 defined. I worked to echo that here.

Lekensteyn:
That sounds rather unusual. By default directory indexes are disabled unless [autoindex](https://nginx.org/r/autoindex) is changed. The [error_page](https://nginx.org/r/error_page) directive only affects the page shown on errors, it does not affect the routing logic (whether to display an index or not).

Did you have a custom config such as [`try_files $uri $uri/index.html =404;`](https://nginx.org/r/try_files) in mind?

With the current approach, there is no way to obtain a directory index if `ipfs-404.html` exists.

jphastings:
That's a good point; I guess it comes down to whether IPFS wants directory listings to always be possible to retrieve on the HTTP gateway.

Given this is my first PR to IPFS I don't feel like I'm qualified to make such a judgement call (beyond what I'd personally like) — so maybe we can lean on @Stebalien?

Lekensteyn:
If directory listings are not desirable, one could add an explicit index.html file to hide the contents. There is probably no point in using this custom 404 page feature for hiding files since people can just query the directory directly, that is where IPFS differs from nginx.

I am also new here and do not know if special files such as index.html and ipfs-404.html are documented anywhere, but that would be nice to have :-)

Stebalien:
Hm. Yeah, I didn't fully think through the ramifications of this. I actually have at least one case where I _want_ to support custom 404 pages _and_ directory listings (dist.ipfs.io).

DavidBurela:
I love that there is thinking around supporting 404 pages.
I've published my blog to IPFS/ENS and lacking 404 was something I wanted
e.g. `http://blog.davidburela.eth/404`

But because it is a blog with pages in this folder format
```
/
  /2019
    /01
    /02
    /03
    ...
  /2020
    /01
    /02
```

it means I'd need to put a `ipfs-404.html` in every single subfolder.  Maybe specifying it once at root of a merkle tree is enough?
`<cid>/ipfs-404.html`

Stebalien:
Specifying it once at the root should be enough. The gateway will search all parent directories up the path for an ipfs-404.html file.

PR #4244, opened 2017-09-17
plugin: support system-wide plugins

### LINKED ISSUE #4238: System-wide plugins
It would be useful to have system-wide plugin directory, like `/usr/lib/ipfs/plugins`. This would allow distributions to package IPFS plugins.

This could be only enabled when building ipfs with some flag, such as `PLUGIN_PATH=/usr/lib/ipfs/plugins make build`.

magik6k:
Closes #4238.

I'm not exactly sure how to wire it to up sharness, cc @Kubuxu.


Rules.mk
```diff
@@ -128,11 +128,12 @@ help:
...
-	@echo '  build        - Build binary at ./cmd/ipfs/ipfs'
-	@echo '  nofuse       - Build binary with no fuse support'
-	@echo '  install      - Build binary and install into $$GOPATH/bin'
-#	@echo '  dist_install - TODO: c.f. ./cmd/ipfs/dist/README.md'
+	@echo '  all           - print this help message'
+	@echo '  build         - Build binary at ./cmd/ipfs/ipfs'
+	@echo '  nofuse        - Build binary with no fuse support'
+	@echo '  install       - Build binary and install into $$GOPATH/bin'
+#	@echo '  dist_install  - TODO: c.f. ./cmd/ipfs/dist/README.md'
+	@echo '  build_plugins - Build plugin binaries
```
Kubuxu: missing single quote at the end

cmd/ipfs/Rules.mk
```diff
@@ -12,7 +12,10 @@ PATH := $(realpath $(d)):$(PATH)
 # DEPS_OO_$(d) += merkledag/pb/merkledag.pb.go namesys/pb/namesys.pb.go
 # DEPS_OO_$(d) += pin/internal/pb/header.pb.go unixfs/pb/unixfs.pb.go
 
-$(d)_flags =-ldflags="-X "github.com/ipfs/go-ipfs/repo/config".CurrentCommit=$(shell git rev-parse --short HEAD)" 
+CONFIG_COMMIT = github.com/ipfs/go-ipfs/repo/config.CurrentCommit=$(shell git rev-parse --short HEAD)
+CONFIG_PLUGIN_PATH = github.com/ipfs/go-ipfs/repo/config.SystemPluginPath=${PLUGIN_PATH}
```
Kubuxu: Use `?=` so it is possible to override this.

magik6k:
Got sharness to work, cc @Kubuxu

test/sharness/Rules.mk
```diff
@@ -49,13 +49,16 @@ $(SHARNESS_$(d)): $(d) ALWAYS
 $(d)/deps: $(SHARNESS_$(d)) $$(DEPS_$(d)) # use second expansion so coverage can inject dependency
 .PHONY: $(d)/deps
 
-test_sharness_deps: $(d)/deps
+$(d)/sharness-ipfs:
+	$(MAKE) cmd/ipfs/ipfs PLUGIN_PATH=./system_plugins
```
Kubuxu: This is invoking makefile recursively, don't do that. I will take a look how to make it not recursive.

cmd/ipfs/Rules.mk
```diff
@@ -12,7 +12,10 @@ PATH := $(realpath $(d)):$(PATH)
 # DEPS_OO_$(d) += merkledag/pb/merkledag.pb.go namesys/pb/namesys.pb.go
 # DEPS_OO_$(d) += pin/internal/pb/header.pb.go unixfs/pb/unixfs.pb.go
 
-$(d)_flags =-ldflags="-X "github.com/ipfs/go-ipfs/repo/config".CurrentCommit=$(shell git rev-parse --short HEAD)" 
+CONFIG_COMMIT ?= github.com/ipfs/go-ipfs/repo/config.CurrentCommit=$(shell git rev-parse --short HEAD)
+CONFIG_PLUGIN_PATH = github.com/ipfs/go-ipfs/repo/config.SystemPluginPath=${PLUGIN_PATH}
```
Kubuxu: I would do 3 things here:

1 . Scope the `CONFIG_COMMIT` to local as `$(d)_commit`
2. Scope the `CONFIG_PLUGIN_PATH` to local as `$(d)_plugin_path`
3. Create file: `config.mk` in root of the project that would contain user configurable variables (using `?=` for defaults with description) and phony target that would print those variables.  Variables in this file would be prefixed with `CONFIG_`. This file would be included early in either `Makefile` or `Rules.mk` (probably `Makefile`).

This way global namespace is not randomly polluted which we are trying to avoid.

magik6k:
Rebased on latest master

magik6k:
So this is potentially broken as it may try to load the preloaded plugins twice, with https://github.com/ipfs/go-ipfs/pull/4806 this should be easy to fix nicely

Stebalien:
@magik6k I've rebased and merged the linked PR.

PR #4246, opened 2017-09-18
"repo stat": Don't crash when Datastore.StorageMax is not defined

### LINKED ISSUE #4241: `ipfs stats repo` fails if `Datastore.StorageMax` is empty
It tries to parse it as a float and fails, causing the whole command to fail out. Ipfs in general seems to run fine without this option set, so the command should either omit that field in this case or have some default.

kevina:
Closes #4241

License: MIT

### DIFF
--- a/core/commands/repo.go
+++ b/core/commands/repo.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"math"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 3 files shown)


kevina:
@whyrusleeping note I use `math.MaxUint64` as the default value in the stat structure to avoid an API change and then simply don't report it in the `repo stat` command if the value is `math.MaxUint64`

whyrusleeping:
@kevina I think '0' works as a default value too, might be a little easier to reason about

kevina:
@whyrusleeping A very large value makes more logical sense.  0 could theoretically have a meaning as in no-storage.  If the the of 0 is used for something bad things could happen.  If a very large value is used, then it will effectively main the same as no limit.  That is why I think `math.MaxUint64` makes more sence.  If it was a floating point number I would use positive Infinity.

kevina:
I can also make `math.MaxUint64` a constant, say `NoLimit`, to make the code easier to read.

PR #4251, opened 2017-09-19
fix dht memory leak

Stebalien:
update go-libp2p-kad-dht to fix a nasty memory leak

License: MIT


Stebalien:
That could be a while (waiting on macos, backlog of... ~2140).

whyrusleeping:
Ran tests locally on OSX, lets skip the travis queue

Issue #4234, opened 2017-09-16
need batch interface for `ipfs dag put`


whyrusleeping:
Playing around with the git ipld plugin and i'm thinking that it would be a lot faster if we were able to do batch puts of dag objects via the http api. This will also probably help out the go-ipld-eth plugin when ingesting lots of data.

hermanjunge:
Yeah. Today, one of my workarounds today is doing 

```
find . -name "eth-state-trie-rlp-*" -exec sh -c "cat {} | ipfs dag put --input-enc raw --format eth-state-trie; echo" \;
```

Kubuxu:
I know that @magik6k hit this too. Spawning so many processes was degrading perf. @magik6k did using plain HTTP was any better?

I also think this was discussed somewhere previously.

Kubuxu:
The thing that was discussed previously was `block put` not `dag put`.

Kubuxu:
I think with most of the use cases the batch operations will use the same `input-enc` and `format` so length prefixed  (8 bytes length prefix with fixed endianness) bytestream should suffice. It would be also quite easy to implement in client libraries (no need for encoding sub-lib).

What do you think?

magik6k:
Sending batches can be done using `req.Files()`, as is done in `ipfs add`.

PR #4254, opened 2017-09-20
dag: Support multiple files in put

magik6k:
See https://github.com/ipfs/go-ipfs/issues/4234

### DIFF
--- a/core/commands/dag/dag.go
+++ b/core/commands/dag/dag.go
@@ -16,7 +16,6 @@ import (
 	cid "gx/ipfs/QmNp85zy9RLrQ5oQD4hPyS39ezrrXpcaa7R4Y9kxdWQLLQ/go-cid"
 	u "gx/ipfs/QmSU6eubNdhXjFBJBSksTp8kv8YRub8mGAPv8tVJHmL2EU/go-ipfs-util"
 	mh "gx/ipfs/QmU9a9NV9RdPNwZQDYd5uKsm6N6LJLSvLbywDDYFbaaC6P/go-multihash"
-	"reflect"
 )
 
 var DagCmd = &cmds.Command{
@@ -93,10 +92,6 @@ into an object of the specified format.
 			}
 		}
 
-		if dopin {
-			defer n.Blockstore.PinLock().Unlock()
-		}
-
 		outChan := make(chan interface{}, 8)
 		res.SetOutput((<-chan interface{})(outChan))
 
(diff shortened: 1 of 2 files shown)


test/sharness/t0053-dag.sh
```diff
@@ -163,7 +163,20 @@ test_dag_cmd() {
...
 	test_expect_success "dag put with dag-pb works output looks good" '
-		printf $HASH > dag_put_exp &&
+		echo $HASH > dag_put_exp &&
+		test_cmp dag_put_exp dag_put_out
+	'
+
+	test_expect_success "dag put multiple files" '
+		printf {\"foo\":\"bar\"} > a.json &&
+		printf {\"foo\":\"baz\"} > b.json &&
+		ipfs dag put a.json b.json > dag_put_out
```
hermanjunge: Awesome!

Will test this week how many files it can get at once w/o choking. Thanks a lot for adding this!

core/commands/dag/dag.go
```diff
@@ -100,52 +97,86 @@ into an object of the specified format.
...
+		}()
 	},
 	Type: OutputObject{},
 	Marshalers: cmds.MarshalerMap{
 		cmds.Text: func(res cmds.Response) (io.Reader, error) {
-			oobj, ok := res.Output().(*OutputObject)
+			outChan, ok := res.Output().(<-chan interface{})
 			if !ok {
-				return nil, fmt.Errorf("expected a different object in marshaler")
+				fmt.Println(reflect.TypeOf(res.Output()))
```
Stebalien: Leftovers from debugging.

core/commands/dag/dag.go
```diff
@@ -100,52 +97,86 @@ into an object of the specified format.
 			defer n.Blockstore.PinLock().Unlock()
```
Stebalien: Given that you add the blocks/pin in a go routine, this needs to go somewhere else (i.e., in the go routine). Also, if you want to improve this slightly, it's probably best not to hold this lock while waiting on the client (either receiving or sending data).

magik6k:
Fixed

core/commands/dag/dag.go
```diff
@@ -96,56 +92,93 @@ into an object of the specified format.
...
+					return fmt.Errorf("no node returned from ParseInputs")
+				}
+
+				var cid *cid.Cid
+				err = func() error {
+					if dopin {
+						defer n.Blockstore.PinLock().Unlock()
+					}
+
+					b := n.DAG.Batch()
```
whyrusleeping: We should put all dag nodes being added into the same batch here, i would move this batch creation up above the start of the for loop.

core/commands/dag/dag.go
```diff
@@ -96,56 +92,93 @@ into an object of the specified format.
...
+						}
+					}
+
+					if err := b.Commit(); err != nil {
+						return err
+					}
+
+					cid = nds[0].Cid()
+					if dopin {
+						n.Pinning.PinWithMode(cid, pin.Recursive)
```
whyrusleeping: Also, since we're batching, we should just gather up all the things we're going to pin into an array, pin them all at once at the end, then only call Pinning.Flush once

core/commands/dag/dag.go
Kubuxu: Are you sure it is right thing to do? What if I start adding 100GB of DAGs?

magik6k: That was my thinking there, though we might just say that users should care about their batch sizes.

whyrusleeping: The DAG batch is size limited, it will autoflush if you put too many items into it

magik6k: done

Issue #3981, opened 2017-06-14
enhancement: "ipfs get" using the Filestore


ProgVal:
#### Type:

Enhancement

#### Severity:

Medium

#### Description:

`ipfs add` now has an experimental `--no-copy` option, which allows to make a file/directory known to IPFS, and serve it to other peers from its original place on the hard disk.

Could you add a similar feature to `ipfs get` to write files to the filesystem, without adding them in `~/.ipfs` too (assuming they are not already there)?

Thanks!

whyrusleeping:
This should be doable, though it will be difficult. We could potentially do this by adding some sort of signaling method to the filestore that says "when you get a block by this hash, you should add it to the filestore with this filepath and offset".

Thoughts @kevina ?

kevina:
Bypassing the "cache" would be difficult.  Copying something in the cache to disk without rehashing is fairly straightforward.  At least it was on my implementation as I implemented as I already implemented it https://github.com/ipfs-filestore/go-ipfs/blob/master/filestore/util/move.go.  Since the current implementation differs from mine i am not 100% sure how useful my code is.

whyrusleeping:
@kevina the difference is that its not being moved to the disk after the fact, we want to write it to the file as its being fetch, avoiding the blockstore entirely

ProgVal:
@whyrusleeping What I had in mind was to write in the blockstore until the file is complete, and then write to the file without recomputing its hash and clear the blockstore.
Your idea is a great feature, but as you said, it is much harder to implement.

whyrusleeping:
@ProgVal ah, that could be done with first fetching the file with `ipfs get`, then running a gc (to clear the blocks from your blockstore) then re-adding it with a no-copy add. Of course, that could all be automated in a flag on get.

The problem there though is that running that gc has the unintended side effect of cleaning out the entire blockstore, which you don't necessarily want. removing blocks from the blockstore after the download is equally as difficult (Requires computing the entire pinset to make sure the blocks we want to remove arent pinned by other things already). I'm actually thinking that the best (and though maybe not the easiest, but the least problematic) option might actually be to try and stream from bitswap straight into the filestore (bypassing the blockstore as i suggest above).

kevina:
@whyrusleeping there is the `ipfs block rm` command that is already implemented, you can batch remove blocks so the pinset is only computed once.

kevina:
[...]

PR #4261, opened 2017-09-21
filestore util: Add basic implementation of 'filestore get' command.

### LINKED ISSUE #4193: Implement "ipfs filestore mv"
See https://github.com/ipfs/go-ipfs/issues/3981

kevina:
This will get a hash and store the results into the filestore.  Afterwards the file is duplicated in both the filestore and normal blockstore.

The next step would be to bypass the normal blockstore when fetching the hash from the network.

When offline this is equivalent to what  I envisioned 'filestore mv' doing and thus closes #4193.

Related to #3981


core/commands/filestore.go
```diff
@@ -201,6 +203,52 @@ var dupsFileStore = &cmds.Command{
 	Type:       RefWrapper{},
 }
 
+var getFileStoreCmd = &cmds.Command{
+	Helptext: cmds.HelpText{
+		Tagline: "Download IPFS objects into filestore.",
+	},
+	Arguments: []cmds.Argument{
+		cmds.StringArg("cid", true, false, "The cid of the IPFS object to be outputted."),
```
whyrusleeping: maybe s/outputted/fetched/ ?

core/commands/filestore.go
```diff
@@ -201,6 +203,52 @@ var dupsFileStore = &cmds.Command{
...
+		if err != nil {
+			res.SetError(err, cmds.ErrNormal)
+			return
+		}
+		filePath, err := filepath.Abs(args[1])
+		if err != nil {
+			res.SetError(err, cmds.ErrNormal)
+			return
+		}
+		println("here we go")
```
whyrusleeping: remove

filestore/get.go
```diff
@@ -0,0 +1,85 @@
...
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	dag "github.com/ipfs/go-ipfs/merkledag"
+	pi "github.com/ipfs/go-ipfs/thirdparty/posinfo"
+	unixfs "github.com/ipfs/go-ipfs/unixfs"
+	cid "gx/ipfs/QmNp85zy9RLrQ5oQD4hPyS39ezrrXpcaa7R4Y9kxdWQLLQ/go-cid"
```
whyrusleeping: I try to always leave a space between go-ipfs imports and gx imports

whyrusleeping:
@kevina so this would be the equivalent of `ipfs get <cid> && ipfs add --nocopy <cid>` ?

filestore/get.go
kevina: Sorry.  Oversight, will fix.

core/commands/filestore.go
kevina: Well, it may not necessary fetch it if its already in local storage.

kevina:
> @kevina so this would be the equivalent of ipfs get <cid> && ipfs add --nocopy <cid> ?

Basically, although slightly more efficient.  This version doesn't pin though.

With a node getter that bypasses the normal block store it will get a block directly to the filestore without duplicating it (assuming it's not already in the local blockstore).

whyrusleeping:
Should be rebased, but LGTM. We should start the conversation for how to fetch things from the network to the filestore, entirely skipping the normal datastore.

kevina:
Rebased and circleci is green.

whyrusleeping:
@kevina Got a merge conflict here, sorry. Mind fixing that up and rebasing?

cc @keks in case there are any issues

kevina:
@keks except for errors this command has no output yet, please let me know if I am doing things correctly.

keks:
Just took a quick look and this seems correct.

PR #4267, opened 2017-09-29
add --local option to 'ipfs dag get'

whyrusleeping:
WIP: need tests

License: MIT


whyrusleeping:
Though maybe the right thing to do here is to handle the `--local` option at a higher level, swapping out the DAG and Resolver for offline variants.

Stebalien:
Another motivation for: #4009

whyrusleeping:
Closing, probably not the right way to do this one.

PR #4272, opened 2017-10-02
README.md: Suggest safer docker flags

### LINKED ISSUE #4271: Docker instructions in readme cause ports 5001 and 8080 to be publicly exposed
The docker instructions in the readme cause ports 5001 (the webui) and 8080 (the ipfs-http proxy) to be publicly exposed instead of only listening on localhost. I don't think that's intended given that ipfs outside of docker defaults to listening to those only on localhost. If it is intended then I think there ought to be much more warning about it in the text; I almost missed this fact while putting ipfs on a server and left the webui open to the world for a while. There is a mention that ports 8080 and 5001 will be exposed, but I assumed docker-ipfs would have the same default config as normal ipfs and only listen to those ports on localhost, or that I was misremembering what port 5001 was used for. It barely crossed my mind that the suggested docker example might actually publicly expose the webui.

Changing the `-p` flags in the command to `-p 127.0.0.1:8080:8080 -p 4001:4001 -p 127.0.0.1:5001:5001` would fix it and make it consistent with dockerless ipfs.

kpcyrd:
Resolves #4271

### DIFF
--- a/README.md
+++ b/README.md
@@ -253,7 +253,7 @@ IPFS files that will persist when you restart the container.
 
 Start a container running ipfs and expose ports 4001, 5001 and 8080:
 
-    docker run -d --name ipfs_host -v $ipfs_staging:/export -v $ipfs_data:/data/ipfs -p 8080:8080 -p 4001:4001 -p 5001:5001 ipfs/go-ipfs:latest
+    docker run -d --name ipfs_host -v $ipfs_staging:/export -v $ipfs_data:/data/ipfs -p 4001:4001 -p 127.0.0.1:8080:8080 -p 127.0.0.1:5001:5001 ipfs/go-ipfs:latest
 
 Watch the ipfs log:
 


README.md
```diff
@@ -253,7 +253,7 @@ IPFS files that will persist when you restart the container.
 
 Start a container running ipfs and expose ports 4001, 5001 and 8080:
 
-    docker run -d --name ipfs_host -v $ipfs_staging:/export -v $ipfs_data:/data/ipfs -p 8080:8080 -p 4001:4001 -p 5001:5001 ipfs/go-ipfs:latest
+    docker run -d --name ipfs_host -v $ipfs_staging:/export -v $ipfs_data:/data/ipfs -p 4001:4001 -p 127.0.0.1:8080:8080 -p 127.0.0.1:5001:5001 ipfs/go-ipfs:latest
```
Kubuxu: 5001 has to be exposed globaly for docker to work

Kubuxu: Oh, 5001 is API. Yeah, SGTM.

Issue #4268, opened 2017-09-29
Allow to remove IPNS entries from my ID


baccenfutter:
#### Version information:
```
go-ipfs version: 0.4.10-
Repo version: 5
System version: amd64/linux
Golang version: go1.8.3
```

#### Description:
> It is currently possible to publish content in IPNS with `ipfs name publish <hash>`. All published content will be listed when resolving the publisher's ID with `ipfs name resolve <id>`.

It should be possible to remove entries under one's ID with something like `ipfs name unpublish <ipnshash>` in a way that

a) they are no longer listed when resolving one's ID
b) it is no longer possible to correlate the IPNS hash with the publisher's ID *(and vice versa)*

Additionally, a wrapper should allow to remove all entries in a single call, such as `ipfs name clear`.

Calmarius:
As a workaround publish /ipfs/QmUNLLsPACCz1vLxQVkXqqLX5R1X345qqfHbsf67hvA3Nn which is an empty directory.

baccenfutter:
I have to admit, I should have played around with the IPNS feature a bit more before opening this issue. When opening it, I had the impression that it is possible to have multiple IPNS hashes published under my ID. Instead, it seems to be the fact that only one IPNS hash can be published under my ID and republishing content will always only make the single same hash point to the new content. It seems to then be intended to just publish a directory containing sub-directories if one wants to somehow sub-categorize the published content; or just publish an empty directory to "delete" the content.

I guess I can always initialize a new repo and/or generate a new ID if I published content I no longer want to be brought in connection with.

All this renders my request obsolete. As a result I am closing it.

djdv:
@baccenfutter 
>it seems to be the fact that only one IPNS hash can be published under my ID 
The `key` subcommand may be of interest to you here, you can generate a new hash for publishing if you wish to publish multiple items under the same repo.
```
ipfs key gen MyNewKey
ipfs name publish --key=MyNewKey *hash*
```
My current strategy is to have my "self" key act as an index for the rest of my IPNS entries.

firnprotocol:
actually, i think this issue is legitimate. even though new `publish`es will overwrite the old one, we might still want to just change to publishing "nothing".

Jorropo:
@firnprotocol any issue with:
```
ipfs name publish /ipfs/bafkqaaa # publish inlined empty file
```
Or:
```
ipfs name publish /ipfs/bafyaabakaieac # publish inlined empty dir
```
?

[...]

PR #4277, opened 2017-10-05
Add option to clear IPNS association

Koshroy:
Clears the IPNS association by publishing a blank path as the resolution
to the IPNS name.

Even though this issue was closed, this PR was created in response to https://github.com/ipfs/go-ipfs/issues/4268


Stebalien:
Hm. It looks like this has been waiting for a review for quite a while... Given it's age, I'm going to give it a short review and then close it to clean up the PR queue a bit.

Thinking about the UX a bit, I'd expect this to act a bit more like the original issue suggested:

* `ipns name clear` (or reset?).
* `ipns name clear --key=...` or `ipns name clear <key>(s)`.

Using a flag doesn't seem like the right solution as the flag/argument are exclusive.

Stebalien:
(we can reopen this PR if you'd like)

PR #4284, opened 2017-10-06
Discussion Point: Using Multihash not CID in blockstore and bitswap

kevina:
This code is broken, that is known.  I did it as an exercise to see exactly what was involved to implement what @diasdavid had in mind in https://github.com/ipfs/go-cid/issues/34#issuecomment-326243777.  I implement enough of the idea to be confident so I know what is involved. Here is what is involved, most of this is likely known but I thought it would be good to spell it out:

(1) Blocks will need to indexed by MultiHash and not the CID.  With our currently infrastructure this means that that additional information provided by the CID will be lost, yet that information is still needed to construct a block given the current interface.  Thus most of the blockstore interface was converted to use Multihash instead of Cid's except for the `Get` method which needs the full Cid in order to correctly reconstruct the block.

(2) It would be temping to make all blocks raw and without any semantic information provided by the CID.  However, that will likely cause problems with the the filestore as it needs to know what type of block it is dealing with.  It may also cause problems elsewhere.

(3) As there is no Semantic information stored in the blockstore/datastore AllKeysChan will return a Multihash instead of CID.

(4) This means that in some cases all that we can return to the user is a Multihash, two noticeable places are "ipfs refs local" and the results of the keys deleted in the GC.  The question is how do we represent these keys to the user, I can think of several ideas: (a) present just the raw multihash but prefix it with a multibase prefix, since this is technically not allowed for CidV0 this could work well, (b) present the keys as Raw CidV1s, this could be confusing to the user (c) add a new Cid type for "unknown" and use that new code to convert the raw multihash to CidV1, this has the advantage of making it clear to the user that the actual Cid type is not known.  (a) could present a problem when we switch to Base32 and we have a need a way to embed a CidV0 into an existing dag, (b) is ambiguous, so I like (c) the best.

(4) AllKeysChan is used is used by the GC, but this is should not be a problem as the full CID is still stored in the pin.

(5) AllKeyChan is used by the republisher.  __This is where the change in the bitswap protocol is needed.__  In particular the proder list (if that is the correct term) would need to consist of multihashes not CID's as we no longer have the semantic information that the CID provides.

(6) As a consequence of storing just a multihash a repo migration will be needed, however since the datastore only cares about raw keys and not what they represent no change in any of the datastore code is needed.

(7) The on-wire protocol will need to be changed due to (5) and likely also changing the wantlist to use Multihashes.

CC @diasdavid @Stebalien @whyrusleeping @lgierth @jbenet (others)

### DIFF
--- a/blocks/blockstore/arc_cache.go
+++ b/blocks/blockstore/arc_cache.go
@@ -7,6 +7,7 @@ import (
 
 	cid "gx/ipfs/QmNp85zy9RLrQ5oQD4hPyS39ezrrXpcaa7R4Y9kxdWQLLQ/go-cid"
 	"gx/ipfs/QmRg1gKTHzc3CZXSKzem8aR4E3TubFhbgXwfVuWnSK5CC5/go-metrics-interface"
+	mh "gx/ipfs/QmU9a9NV9RdPNwZQDYd5uKsm6N6LJLSvLbywDDYFbaaC6P/go-multihash"
 	ds "gx/ipfs/QmVSase1JP7cq9QkPT46oNwdp9pT6kBkG3oqS14y3QcZjG/go-datastore"
 	lru "gx/ipfs/QmVYxfoJQiZijTgPNHCHgHELvQpbsJNTg6Crmc3dQkj3yy/golang-lru"
 )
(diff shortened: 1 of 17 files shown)


Stebalien:
To consider all options, we may be able to do this at the datastore level instead of the blockstore level to avoid breaking stuff quite so much. Having the `Block` type store CIDs is nice as it lets us pass around typed data without actually having to understand it. However, this makes the caching story a bit awkward.

kevina:
Outdated, will redo once we decide on the new Blockstore interface, see https://github.com/ipfs/go-ipfs-blockstore/issues/8.

Issue #4282, opened 2017-10-05
t0180-p2p is racy


Stebalien:
Specifically:

```
test_expect_success 'Test server to client communications' '
  ma-pipe-unidir --listen --pidFile=listener.pid send /ip4/127.0.0.1/tcp/10101 < test0.bin &

  go-sleep 500ms && # GRR...
  kill -0 $(cat listener.pid) &&
```

Stebalien:
#4411 :sob:

PR #4285, opened 2017-10-06
add raw support to the dag put command.

Stebalien:
`dag get` still doesn't properly support raw output but that's a bigger issue.

License: MIT


Stebalien:
Note: `dag get $raw_node_hash` still doesn't work properly because `RawNode` can't be marshaled to JSON. To do this correctly, we really need a nice way to convert between IPLD formats (assuming that IPLD nodes implement a JSON marshaler isn't really a solution).

whyrusleeping:
CircleCI is failing for some reason

Stebalien:
@whyrusleeping https://github.com/ipfs/go-ipfs/issues/4282

PR #4288, opened 2017-10-06
WIP: Connection Manager 

whyrusleeping:
This is the first WIP branch of the connection manager. It allows you to set up a ConnMgr that tries to keep the number of open connections your node has within certain bounds.

Theres still a lot to do here.


whyrusleeping:
supplementary: https://github.com/libp2p/go-libp2p-kad-dht/pull/95

whyrusleeping:
also: https://github.com/libp2p/go-libp2p-circuit/pull/20

whyrusleeping:
proposal:

Merge this roughly as is (without connection tagging logic) so that we can start using it in 'dumb' mode asap. Then, PR the different connection tagging things in individually.

whyrusleeping:
note: disabled by default, you will need to set up the configuration as defined in the config.md doc

core/core.go
```diff
@@ -814,6 +839,10 @@ func constructPeerHost(ctx context.Context, id peer.ID, ps pstore.Peerstore, bwr
 	if !opts.DisableNatPortMap {
 		hostOpts = append(hostOpts, p2pbhost.NATPortMap)
 	}
+	if opts.ConnectionManager != nil {
+		fmt.Println("adding conn manager to host opts")
```
magik6k: Looks like debugging leftover.

core/core.go
```diff
@@ -261,6 +269,22 @@ func (n *IpfsNode) startOnlineServices(ctx context.Context, routingOption Routin
 	return n.Bootstrap(DefaultBootstrapConfig)
 }
 
+func constructConnMgr(cfg config.ConnMgr) (ifconnmgr.ConnManager, error) {
+	switch cfg.Type {
+	case "", "none":
+		return nil, nil
+	default:
```
magik6k: This should probably be on the bottom of the switch

docs/config.md
```diff
@@ -267,3 +267,18 @@ Disables the p2p-circuit relay transport.
...
+Connection manager configuration.
+
+- `Type`
+Sets the type of connection manager to use, options are: `"none"` and `"basic"`.
+
+- `LowWater`
+LowWater is the minimum number of connections to maintain.
+
+- `HighWater`
+HighWater is the number of connections that, when exceeded, will trigger a connection GC operation.
```
magik6k: I don't really like these names. Wouldn't simple `Low`/`High` be enough?

magik6k:
Note: connections used in `ipfs p2p` may need tagging

docs/config.md
whyrusleeping: Hrm... I'd something a little more descriptive than that. I've seen these terms used in this context before. But i'm open to suggestions

kevina: How about a simple `Min` and `Max` or maybe `MinConn` or `MaxConn`?

whyrusleeping:
Maybe we should actually have this enabled by default, limit people to 900-1000 connections or so. Even if it does cause potentially strange behaviour, its probably better than endless "too many open file descriptor" errors...

whyrusleeping:
For reference, the connection closing logic is here: https://github.com/libp2p/go-libp2p-connmgr

core/core.go
```diff
@@ -380,7 +404,7 @@ func setupDiscoveryOption(d config.Discovery) DiscoveryOption {
 			if d.MDNS.Interval == 0 {
 				d.MDNS.Interval = 5
 			}
-			return discovery.NewMdnsService(ctx, h, time.Duration(d.MDNS.Interval)*time.Second)
+			return discovery.NewMdnsService(ctx, h, time.Duration(d.MDNS.Interval)*time.Second, discovery.ServiceTag)
```
diasdavid: Is this change related to this PR?

core/core.go
```diff
@@ -219,11 +221,17 @@ func (n *IpfsNode) startOnlineServices(ctx context.Context, routingOption Routin
...
+	if err != nil {
+		return err
+	}
+
 	hostopts := &ConstructPeerHostOpts{
 		AddrsFactory:      addrsFactory,
 		DisableNatPortMap: cfg.Swarm.DisableNatPortMap,
 		DisableRelay:      cfg.Swarm.DisableRelay,
 		EnableRelayHop:    cfg.Swarm.EnableRelayHop,
+		ConnectionManager: connmgr,
```
diasdavid: Do we need to instantiate the ConnManager here? As a user of libp2p, I would assume that ConnManager is just a thing that exists and that all I need to do is pass an option to select the strategy, saving users from yet importing another package.

whyrusleeping:
To set the config, run: 
```
ipfs config --json Swarm.ConnMgr '{"GracePeriod": "30s","HighWater": 1000,"LowWater": 900,"Type": "basic"}'
```

docs/config.md
```diff
@@ -267,3 +267,18 @@ Disables the p2p-circuit relay transport.
...
 an intermediate (Hop Relay) node in relay circuits for connected peers.
+
+### `ConnMgr`
+Connection manager configuration.
+
+- `Type`
+Sets the type of connection manager to use, options are: `"none"` and `"basic"`.
+
+- `LowWater`
+LowWater is the minimum number of connections to maintain.
```
diasdavid: How is `lowWater` enforced? Does ConnManager run Discovery Mechanisms if we are poorly connected? (Didn't see this in the go-libp2p-connmgr code)

docs/config.md
diasdavid: `Watermark` is something that is familiar to me in terms of setting upper and lower bounds to something in programming. No special feelings attached though.

docs/config.md
```diff
@@ -267,3 +267,18 @@ Disables the p2p-circuit relay transport.
...
+
+- `HighWater`
+HighWater is the number of connections that, when exceeded, will trigger a connection GC operation.
+- `GracePeriod`
+GracePeriod is the length of time that new connections are immune from being closed by the connection manager.
```
diasdavid: I assume this is in seconds? Can get that written down?

p2p/p2p.go
```diff
@@ -11,7 +11,7 @@ import (
 	ma "gx/ipfs/QmXY77cVe7rVRQXZZQRioukUM7aRW3BTcAgJe12MCtb3Ji/go-multiaddr"
 	peer "gx/ipfs/QmXYjuNuxVzXKJCfWasQk1RqkhVLDM9jtUKhqc2WPQmFSB/go-libp2p-peer"
 	pro "gx/ipfs/QmZNkThpqfVXs9GNbexPrfBbXSLNYeKrE7jwFM2oqHbyqN/go-libp2p-protocol"
-	p2phost "gx/ipfs/QmaSxYRuMq4pkpBBG2CYaRrPx2z7NmMVEs34b9g61biQA6/go-libp2p-host"
+	p2phost "gx/ipfs/Qmc1XhrFEiSeBNn3mpfg6gEuYCt5im2gYmNVmncsvmpeAk/go-libp2p-host"
```
diasdavid: Curiosity, in go, interface packages don't get an `interface` name?

core/core.go
whyrusleeping: No, it just happened to change in a dependency that i needed to update here

diasdavid:
I see that https://github.com/libp2p/go-libp2p-connmgr/blob/master/connmgr.go keeps track of the peers to whom we opened connections. I guess in go memory is very cheap, in our JS implementation I would/will implement this ConnManager as just checking everything that is on PeerBook (your PeerStore) and using the `.isConnected` a method in each PeerInfo instance to check that, just to save allocating more memory to store the same info and avoiding having to update in 2 places.

docs/config.md
whyrusleeping: No, its not a 'minimum number of connections' its just the point at which we stop closing connections. So if LowWater is 900 and HighWater is 1000, we will trigger a Trim once we have over 1000, and then close connections until we have 900 or less

docs/config.md
whyrusleeping: Its a time duration (like all the other time related things). 30s == 30 seconds, 1h == one hour, 4m == 4 minutes, etc

p2p/p2p.go
whyrusleeping: No, they do. But this is where we're instantiating the concrete type.

whyrusleeping:
> I see that https://github.com/libp2p/go-libp2p-connmgr/blob/master/connmgr.go keeps track of the peers to whom we opened connections. 

Primarily, it keeps track of *tags* on each connection.

core/core.go
whyrusleeping: It might be nice to have a `libp2p-bundle` package that imports everything all together, but for now the tying together of all the pieces is done here in the core constructor. Importing things at the top level as much as possible makes dependency tree modifications much cheaper. If this were done in a lower package, then every time i make a change to the connection manager logic, i have to update that package (or packages) and then update those into here.

docs/config.md
Stebalien: Note: I'd avoid min/low as they imply that we'll actively seek peers up to this point.

docs/config.md
```diff
@@ -267,3 +267,18 @@ Disables the p2p-circuit relay transport.
...
+- `Type`
+Sets the type of connection manager to use, options are: `"none"` and `"basic"`.
+
+- `LowWater`
+LowWater is the minimum number of connections to maintain.
+
+- `HighWater`
+HighWater is the number of connections that, when exceeded, will trigger a connection GC operation.
+- `GracePeriod`
+GracePeriod is a time duration that new connections are immune from being closed by the connection manager.
```
Stebalien: Unless I'm mistaken, this can trivially be used to DOS a node.

whyrusleeping: I wouldnt call it trivially, but yes, you can use this to exceed a nodes maximum connection count and make a connection closing sweep not reach the low water mark.

Stebalien: Sorry, for some reason, I assumed we had a hard "don't open new connections past the high water mark" limit (I was a bit sleep deprived due to jet lag...). That would have effectively taken the node offline.

---

However, still worried that this will be a way to monopolize a peer by tying up their connection quota with connections in the grace period. Basically, you can ensure that no connection lasts more than 10 seconds. Maybe don't count connections in the grace period?

docs/config.md
whyrusleeping: So... are we in favor of still using `HighWater` and `LowWater` ? I'm getting mixed signals (especially from @Stebalien who thumbs upped davids comment, but suggested changing things). 

I agree that "min" would imply we actively open connections until we have that many, but I think "low" avoids this connotation.

docs/config.md
whyrusleeping: Yeah, connections in the grace period are basically not counted, see here: https://github.com/libp2p/go-libp2p-connmgr/blob/master/connmgr.go#L61

docs/config.md
magik6k: `(GC)Treshold(Conn/Peers)`, `(GC)Target(Conn/Peers)`? `Water` may sound confusing for some even if it's technically correct.

docs/config.md
Stebalien: I see, so we literally just skip them. Nevermind, LGTM.

Kubuxu:
Observations from some of our nodes, connection closing was run 
  - pluto, which is one of bootstrap nodes: has started maxing out its CPU when connection closing - it was almost maxing out its CPU when it was not closing too. Fetching files from pluto is almost impossible but it was hard before that.
 - biham, which is one of the storage hosts: I have observed significant reduction in CPU, FDs and RAM usage and it is still responsive in regards of fetching files. Connection closing for it is great addition.

package.json
```diff
@@ -466,9 +466,15 @@
     },
     {
       "author": "whyrusleeping",
-      "hash": "QmQgLZP9haZheimMHqqAjJh2LhRmNfEoZDfbtkpeMhi9xK",
+      "hash": "QmWRCn8vruNAzHx8i6SAXinuheRitKEGu8c7m26stKvsYx",
       "name": "go-testutil",
-      "version": "1.1.12"
+      "version": "1.1.11"
```
lgierth: Is go-testutil being downgraded?

whyrusleeping: hrm... thats probably the fault of a rebase gone rogue. I'll take care of it.

Kubuxu:
Only comment I have is that it would be good to have info about this feature returned somewhere from the API and have a test that config is being loaded correctly.

package.json
whyrusleeping: Ah, its because we accidentally updated go-testutil in master without bubbling the changes from the rest of the deps at the same time. I think this change is safe for now, we can bubble up those deps and fix things later.

whyrusleeping:
@Kubuxu 

> Only comment I have is that it would be good to have info about this feature returned somewhere from the API and have a test that config is being loaded correctly.

Any suggestions here?

lgierth:
> Any suggestions here?

The network section in `ipfs diag sys` would do

whyrusleeping:
@Kubuxu i think we should merge this now, and do the api introspection as a separate PR. Theres a fair amount of discussion to be had around that (i realize as i write it).

whyrusleeping:
see https://github.com/ipfs/go-ipfs/pull/4308

Kubuxu:
I am not blocking this PR. I just raised a note that we should have it.

whyrusleeping:
Choo Choo!

Kubuxu:
https://www.youtube.com/watch?v=PJqVaWREzk0

skorokithakis:
Semi-relatedly, and because I want to activate this feature and see what difference it makes, is there a command that will show me on the CLI the (number of) active connections so I can tune the high/low water marks sensibly?

whyrusleeping:
@skorokithakis `ipfs swarm peers` lists all active connections. I'm also working on a subcommand for the connection manager which should allow some more fine grained information

Issue #4258, opened 2017-09-21
`ipfs dht provide` fails silently with no peers


whyrusleeping:
if you have no peers in `ipfs swarm peers`, `ipfs dht provide` will fail silently.

PR #4290, opened 2017-10-07
Check for Peers before `dht provide`

Koshroy:
Check if there are any peers before running logic for `dht provide`. In response to https://github.com/ipfs/go-ipfs/issues/4258

License: MIT
Signed-off-by: Koushik Roy <meff@meff.me>

### DIFF
--- a/core/commands/dht.go
+++ b/core/commands/dht.go
@@ -274,6 +274,11 @@ var provideRefDhtCmd = &cmds.Command{
 			return
 		}
 
+		if len(n.PeerHost.Network().Conns()) == 0 {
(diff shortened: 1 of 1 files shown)


core/commands/dht.go
```diff
@@ -274,6 +274,11 @@ var provideRefDhtCmd = &cmds.Command{
 			return
 		}
 
+		if len(n.PeerHost.Network().Conns()) == 0 {
+			res.SetError(errors.New("no peers in the swarm are connected to, cannot provide"), cmds.ErrNormal)
```
whyrusleeping: maybe: "cannot provide, no connected peers" ?

PR #4295, opened 2017-10-10
Fix link to IPFS paper in contribute.md

isaacd9:
- Change broken static.benet.ai hyperlink to non broken and likely more stable arxiv.org hyperlink

License: MIT
Signed-off-by: Isaac Diamond <isaacd9@gmail.com>

### DIFF
--- a/contribute.md
+++ b/contribute.md
@@ -17,7 +17,7 @@ Please look and conform to our [Go Contribution Guidelines](https://github.com/i
 ## General Guidelines:
 
 - See the [dev pseudo-roadmap](dev.md).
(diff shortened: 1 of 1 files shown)


Stebalien:
I believe this is the link we usually use: https://github.com/ipfs/ipfs/blob/master/papers/ipfs-cap2pfs/ipfs-p2p-file-system.pdf (although we should consider using ipfs...).

Regardless, I'm not sure we should be recommending that people implement IPFS according to how it's described in the paper. As far as I know, there have been some significant changes.

isaacd9:
What's the best way to move forward? Should I just remove this line from the documentation then...?

whyrusleeping:
Lets just drop the reference to the paper, and link to the specs for now.

Issue #898, opened 2015-03-08
ipfs add horrendously slow


anarcat:
forgot to file that one, it seems. one of the first thin i tried in ipfs was to add a 3.2GB file on my laptop. the setup was that I had a HDD external drive connected through a SATA USB2 enclosure, on my laptop running Debian Jessie amd64.  The transfer took over an hour.

Doing the same copy with `rsync` takes a few minutes - i am able to get around 60-70MB/s transfer rates on this drive, it seems. Here's a copy of the IRC log for more information:

```
21:52 <anarcat> what does ipfs actually *do*?
21:52 <anarcat> it uploads the file to the dht?
21:57 <anarcat> i am trying to add a 3.2GB file and it's taking a long time, and a lot of memory
21:58 <anarcat>   PID USER      PR  NI    VIRT    RES    SHR S  %CPU %MEM     TIME+ COMMAND
21:58 <anarcat> 29800 anarcat   20   0 2079848 959980   9224 S  30,9 26,3   8:00.67 ipfs
21:58 <anarcat> 863.50 MB / 3.19 GB [=======>----------------------] 26.45 % 33m14s
22:00 <anarcat> at least the memory usage isn't growing so much so far
22:01 <anarcat> uh, and that's the daemon too
22:06 <dPow> Haha, be happy if it completes. Not too long ago a gig would break it
22:07 <anarcat> wow okay :)
22:07 <anarcat> that's the "alpha" part, i guess
22:07 <anarcat> so many promises, this project is exciting :)
22:07 <anarcat> 1.24 GB / 3.19 GB [============>-------------------] 38.87 % 32m55s
22:07 <dPow> whyrusleeping can probably give a very quick explanation to it
22:07 <anarcat> it seems to asymptotically approach a ETA of 30 minutes :P
22:07 <anarcat> dPow: thanks
22:10 <anarcat> this has got to be one of the most frustrating progress bar i've seen in a while
22:10 <anarcat> it just goes in bursts then stalls
22:10 <anarcat> 1.39 GB / 3.19 GB [=============>------------------] 43.50 % 30m24s
22:14 <@whyrusleeping> hey anarcat!
22:14 <@whyrusleeping> the memory usage is actually being addressed in PR 860
22:14 <@whyrusleeping> we had a bug :/
22:14 <joeyh> oh, was that my bug?
22:15 <@whyrusleeping> joeyh: it was indeed :)
[...]

PR #4296, opened 2017-10-11
parallelize batch flushing

Stebalien:
1. Modern storage devices (i.e., SSDs) tend to be highly parallel.
2. Allows us to read and write at the same time (avoids pausing while flushing).

This makes `ipfs add --local` ~3.5x faster with the flatfs datastore (untested
with badger).

fixes https://github.com/ipfs/go-ipfs/issues/898#issuecomment-331849064

License: MIT

### DIFF
--- a/merkledag/batch.go
+++ b/merkledag/batch.go
@@ -9,7 +9,7 @@ import (
 )
 
 // ParallelBatchCommits is the number of batch commits that can be in-flight before blocking.
-// TODO: Experiment with multiple datastores, storage devices, and CPUs to find
+// TODO(#4299): Experiment with multiple datastores, storage devices, and CPUs to find
 // the right value/formula.
 var ParallelBatchCommits = runtime.NumCPU() * 2
 
--- a/merkledag/batch.go
+++ b/merkledag/batch.go
@@ -43,7 +43,8 @@ func (t *Batch) processResults() {
 }
 
 func (t *Batch) asyncCommit() {
-	if len(t.blocks) == 0 || t.commitError != nil {
+	numBlocks := len(t.blocks)
+	if numBlocks == 0 || t.commitError != nil {
 		return
 	}
 	if t.activeCommits >= ParallelBatchCommits {
@@ -61,7 +62,7 @@ func (t *Batch) asyncCommit() {
 	}(t.blocks)
 
 	t.activeCommits++
-	t.blocks = nil
+	t.blocks = make([]blocks.Block, 0, numBlocks)
 	t.size = 0
 
 	return
--- a/merkledag/batch.go
+++ b/merkledag/batch.go
@@ -1,0 +1,98 @@
+package merkledag
+
+import (
+	"runtime"
+
+	cid "gx/ipfs/QmNp85zy9RLrQ5oQD4hPyS39ezrrXpcaa7R4Y9kxdWQLLQ/go-cid"
+	node "gx/ipfs/QmPN7cwmpcc4DWXb4KTB9dNAJgjuPY69h3npsMfhRrQL9c/go-ipld-format"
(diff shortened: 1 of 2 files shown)


Stebalien:
We may want to reduce the parallelism. `2*NumCpu` is slightly faster than `NumCpu` but will allocate 2x the memory (on a 4 thread CPU, it will allocate 64MiB instead of 32MiB).

However, we should probably test badger first (it may work better with increased parallelism).

Stebalien:
This makes adds with sync enabled almost as fast adds with sync disabled with the badger datastore (and the same `2*NumCpu` constant seems to work).

whyrusleeping:
This is really nice, great find here :)

merkledag/batch.go
```diff
@@ -0,0 +1,98 @@
...
+import (
+	"runtime"
+
+	cid "gx/ipfs/QmNp85zy9RLrQ5oQD4hPyS39ezrrXpcaa7R4Y9kxdWQLLQ/go-cid"
+	node "gx/ipfs/QmPN7cwmpcc4DWXb4KTB9dNAJgjuPY69h3npsMfhRrQL9c/go-ipld-format"
+	blocks "gx/ipfs/QmSn9Td7xgxm9EV7iEjTckpUWmWApggzPxu7eFGWkkpwin/go-block-format"
+)
+
+// ParallelBatchCommits is the number of batch commits that can be in-flight before blocking.
+// TODO: Experiment with multiple datastores, storage devices, and CPUs to find
```
Kubuxu: Can you create issues for this instead of in code TODO (they just get forgotten).

merkledag/batch.go
```diff
@@ -0,0 +1,98 @@
...
+			return
+		}
+	}
+	go func(b []blocks.Block) {
+		_, err := t.ds.Blocks.AddBlocks(b)
+		t.commitResults <- err
+	}(t.blocks)
+
+	t.activeCommits++
+	t.blocks = nil
```
Kubuxu: I would preallocate a buffer here of `MaxNodes` as appending will expand the buffer and cause more allocations and copies.

Stebalien: Technically, the max size of this array is 128 pointers to blocks (2KiB). However, it will likely never be greater than 32 pointers (0.5KiB) assuming that we have 256KiB blocks. Does 32 sound like a reasonable default size?

Personally, I don't think that will make much of a difference. We already do 1 allocation per block so this will only add another `log(n)` allocations.

merkledag/batch.go
Stebalien: Fine! (...grumble... we'll never get to it anyways)

merkledag/batch.go
Stebalien: Actually, I just preallocated a blocks array of the same size as the one we just filled. That should be a reasonable guess.

Kubuxu: As go allocates next power of 2, getting to 128 would be 9 reallocations and copies. IMO it is worth it.

merkledag/batch.go
Kubuxu: that might be true but it is still better than having TODO in code.
As you said in your issue, someone might just take a stab at it out of pure boredom.

Stebalien: You're right, I was just being lazy 🙂.

merkledag/batch.go
Stebalien: You're right, my `log(n)` estimate was incorrect anyways. `log(n)` per batch but still `n` overall (7-15% allocation overhead depending on the block sizes).

Stebalien:
After further testing, the effect isn't nearly so pronounced for medium files (the tests above were on single large files) and is probably non-existent for small files as we create a new batch per file. We should consider using the same batch when adding multiple small files.

Issue #4056, opened 2017-07-11
Incorrect `repoSize` when using symbolic links


bitspill:
#### Version information:
```
go-ipfs version: 0.4.10-
Repo version: 5
System version: amd64/linux
Golang version: go1.8.3
```

#### Type:
`Bug`

#### Severity:
`Low`

#### Description:

If the .ipfs directory is a symbolic link the repoSize as returned by `ipfs repo stat` or `ipfs stats repo` will always return 9 bytes regardless of actual data stored

With symlink
```
~$ ls -al
lrwxrwxrwx   1 ipfs ipfs        9 Jun 22  2016 .ipfs -> ipfs_data
drwxrwxr-x   5 ipfs ipfs     4096 Jul 12 01:25 ipfs_data

~$ ipfs repo stat
NumObjects: 204114
RepoSize:   9
StorageMax: 800000000000
RepoPath:   /home/ipfs/.ipfs
Version:    fs-repo@5
```

Without symlink
```
~$ ipfs repo stat
NumObjects: 204114
RepoSize:   49450734482
StorageMax: 800000000000
RepoPath:   /home/ipfs/.ipfs
Version:    fs-repo@5
```

bitspill:
Did some looking and found that filepath.Walk does not follow symbolic links

[1] https://github.com/ipfs/go-ipfs/blob/master/repo/fsrepo/fsrepo.go#L591
[2] https://golang.org/pkg/path/filepath/#Walk

leerspace:
This was fixed in #4305

PR #4305, opened 2017-10-14
handle sym links in when calculating repo size

leerspace:
repo/fsrepo: follow symbolic links when calculating RepoSize

Now if the IPFS path is a symbolic it will be followed before
calculating the repo size.

License: MIT
Signed-off-by: John Reed <john@re2d.xyz>


leerspace:
Fixes #4056

leerspace:
I'm confused by the GitCop failures since the commit already has the license and signed-off-by trailers.

whyrusleeping:
@leerspace sometimes it gets upset because it doesnt perfectly match its regex. If you have the right content, then just ignore it.

Kubuxu:
If it isn't too much to ask for, a sharness test for it would be great.

leerspace:
I started working on the sharness test, but now I'm trying to figure out why every time I run `ipfs repo stat` the RepoSize gets larger by about ~1kB. This is even with no daemon running.

Kubuxu:
@leerspace this is the behaviour of those DBs, every time they are opened they create new logfile and if there are X of them (or other condition) they compact. I would suggest adding few hundreds of KiB to ipfs in sharness and checking if size is larger than X.

Kubuxu:
Also according to `go fmt` this code is not formatted correctly.

whyrusleeping:
@leerspace starting from a fresh repo (as in a test) should yield the same size every run (we have tests that test this). I would just pick a number, run the test, see what it actually was in the test failure output and then adjust the value in your test assertion.

whyrusleeping:
Though I definitely agree that running `ipfs repo stat` with no daemon running should not affect the size of the repo.

leerspace:
I just committed the gofmt fix and a sharness test for this.

whyrusleeping:
@leerspace looks like you might not have set the executable bit on that test file. the test runners arent very happy trying to run it

leerspace:
@whyrusleeping sorry about that. I think it should be fixed now. I did this on Windows, but fortunately there's [a way to set this on Windows](https://stackoverflow.com/a/21694391/1388019).

PR #4307, opened 2017-10-14
remove the rest of the supernode code

Stebalien:
missed in the initial cleanup


whyrusleeping:
still got some in cmd/ipfs/daemon.go, see: `hipster-fast-search-utility supernode`

Stebalien:
@whyrusleeping I left that in to tell users that the option isn't supported. Would you like me to remove it?

whyrusleeping:
Ah, I guess that makes sense to leave in then.

Kubuxu:
@Stebalien I missed this in the initial PR, could you add a sharness test for error message if supernode is selected?

Stebalien:
@Kubuxu picky picky... 😜. Done.

(and thanks for being a stickler for things like this)

PR #4308, opened 2017-10-16
add a subcommand for the connmgr

whyrusleeping:
License: MIT


whyrusleeping:
(note: this relies on uncommitted changes to the connmgr lib)

core/commands/swarm.go
```diff
@@ -812,3 +814,72 @@ func filtersRemove(r repo.Repo, cfg *config.Config, toRemoveFilters []string) ([
...
+		if n.PeerHost == nil {
+			res.SetError(errNotOnline, cmds.ErrClient)
+			return
+		}
+
+		infos := make(map[string]interface{})
+		cmgr := n.PeerHost.ConnManager()
+		switch cmgr := cmgr.(type) {
+		case nil:
+			infos["type"] = "<nil>"
```
Stebalien: Is distinguishing between these two cases really useful? Users aren't going to know what `<nil>` means.

core/commands/swarm.go
```diff
@@ -812,3 +814,72 @@ func filtersRemove(r repo.Repo, cfg *config.Config, toRemoveFilters []string) ([
...
+		case *ifconnmgr.NullConnMgr:
+			infos["type"] = "disabled"
+		case *connmgr.BasicConnMgr:
+			infos["type"] = "basic"
+
+			inf := cmgr.GetInfo()
+			infos["lowWater"] = inf.LowWater
```
Stebalien: I assume these are supposed to be human readable keys? If so, we should probably just say things like "Low Watermark" instead.

core/commands/swarm.go
whyrusleeping: They shouldnt ever see "nil", Everything currently will either return a NullConnMgr or a BasicConnMgr. So maybe I should remove it?

Stebalien: Probably. I'd fail with an error if this isn't supposed to happen.

core/commands/swarm.go
Kubuxu: Those will be the result of JSON endpoint. We can create mappings in TextMarshaller.

Kubuxu:
@whyrusleeping so how about making human readable format having really human readable keys (`lowWater => Low Watermark` and so on)?

whyrusleeping:
@GitCop what is your problem? I have that.

whyrusleeping:
Can someone take this over for me?

lidel:
ConnMgr work continued in https://github.com/ipfs/kubo/issues/9420

PR #4315, opened 2017-10-17
tag peers associated with a bitswap session

whyrusleeping:
also a wip

License: MIT

### DIFF
--- a/exchange/bitswap/session.go
+++ b/exchange/bitswap/session.go
@@ -2,6 +2,7 @@ package bitswap
 
 import (
 	"context"
+	"fmt"
 	"time"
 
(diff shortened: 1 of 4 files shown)


exchange/bitswap/session.go
```diff
@@ -216,6 +223,11 @@ func (s *Session) run(ctx context.Context) {
 		case <-ctx.Done():
 			s.tick.Stop()
 			s.bs.removeSession(s)
+
+			cmgr := s.bs.network.ConnectionManager()
+			for _, p := range s.activePeersArr {
+				cmgr.UntagPeer(p, s.tag)
```
Stebalien: We should consider adding a method for untagging all peers (although we'll have to consider the performance tradeoffs).

PR #4317, opened 2017-10-17
NewStream now creates a connection if necessary

Stebalien:


Stebalien:
I assume we can remove this.

PR #4320, opened 2017-10-18
gateway: fix seeker can't seek on specific files

### LINKED ISSUE #4286: `Seeker can't seek` when referencing data-containing node
#### Version information:
`All versions affected`

#### Type:
`Bug`

#### Severity:
`High`

#### Description:

I put together a demonstration of the problem at [this location](https://ipfs.io/ipfs/QmZL4wivfYBEUutxksKgbpPfFkj1tGNMESYLFF2MWcST8Y/Last%20entry%20in%20this%20subdir%20errors%20on%20latest%20go-ipfs): if you try to look at [the last entry in that directory](https://ipfs.io/ipfs/QmZL4wivfYBEUutxksKgbpPfFkj1tGNMESYLFF2MWcST8Y/Last%20entry%20in%20this%20subdir%20errors%20on%20latest%20go-ipfs/leaf_file_node_including_data%20x%202), the following is observed:
```
curl -D - -s ipfs.io/ipfs/QmTTDNxNJUqF5PZicJqjDjsUed1bHaPWwFdEMvb1iY93Ur
HTTP/1.1 500 Internal Server Error
Date: Fri, 06 Oct 2017 11:16:53 GMT
Content-Type: text/plain; charset=utf-8
Content-Length: 18
Connection: keep-alive
Access-Control-Allow-Headers: Content-Range, X-Chunked-Output, X-Stream-Output
Access-Control-Allow-Methods: GET
Access-Control-Allow-Origin: *
Access-Control-Expose-Headers: Content-Range, X-Chunked-Output, X-Stream-Output
Cache-Control: public, max-age=29030400, immutable
Etag: "QmTTDNxNJUqF5PZicJqjDjsUed1bHaPWwFdEMvb1iY93Ur"
Last-Modified: Thu, 01 Jan 1970 00:00:01 GMT
Suborigin: ipfs000bciqex6g372ssrd2i6o5tc35pdbbpxxgyw2mm56wpb3sxchw6rlangzy
X-Content-Type-Options: nosniff
X-Ipfs-Path: /ipfs/QmTTDNxNJUqF5PZicJqjDjsUed1bHaPWwFdEMvb1iY93Ur
Strict-Transport-Security: max-age=15768000; includeSubDomains; preload

seeker can't seek
```

A complete diagram of all blocks involved and their contents is included at https://ipfs.io/ipfs/QmZL4wivfYBEUutxksKgbpPfFkj1tGNMESYLFF2MWcST8Y/block_dag.svg

magik6k:
Fixes #4286

This is the somewhat hacky approach, but it is still way better (IMO) than rewriting `http.ServeContent` just to support that one case.


mib-kd743naq:
> way better (IMO) than rewriting http.ServeContent just to support that one case.

@magik6k my understanding of @whyrusleeping's https://github.com/ipfs/go-ipfs/issues/4286#issuecomment-336195370 is that aside from "that one case", the way we are using the HTTP server has non-trivial performance implications ( iiuc responses essentially can't be streamed until the entire DAG is enumerated ).

If the above is correct I much rather wait for the proper fix, than having a workaround go into go-ipfs, precluding time being allocated to fix the real problem :/

magik6k:
That is correct, the simplest solution here would be to just trust `dr.pbdata.GetFilesize` -
 https://github.com/ipfs/go-ipfs/pull/4320/files#diff-e2e926fd88d79858d19deb317a28cb7bR245 in providing accurate file size. Looking at http lib internals it shouldn't open up any memory-exhaustion-DOS scenarios.

I'll commit that version of the fix soon

magik6k:
So this PR now does few things:
* Makes gateway use custom seeker which uses Size method from reader when Seek is called with io.SeekEnd, this way we don't need to iterate large part of the dag to start getting blocks. It also fixes The problem with unset subnode sizes
* Makes `pbDagReader` seek do nothing when it doesn't really need to
* Makes `merkledag.dedupeKeys` return keys in correct order

With these changes large files start downloading immediately after first block is found and are streamed without randomly hanging on blocks that would be downloaded last.

core/corehttp/gateway_handler.go
```diff
@@ -372,6 +372,41 @@ func (i *gatewayHandler) getOrHeadHandler(ctx context.Context, w http.ResponseWr
...
+
+func (s *sizeSeeker) Seek(offset int64, whence int) (int64, error) {
+	if whence == io.SeekEnd && offset == 0 {
+		return int64(s.Size()), nil
+	}
+
+	return s.sizeReadSeeker.Seek(offset, whence)
+}
+
+func (i *gatewayHandler) serverFile(w http.ResponseWriter, req *http.Request, name string, modtime time.Time, content io.ReadSeeker) {
```
whyrusleeping: s/serverFile/serveFile/

core/corehttp/gateway_handler.go
```diff
@@ -372,6 +372,41 @@ func (i *gatewayHandler) getOrHeadHandler(ctx context.Context, w http.ResponseWr
...
+			sizeReadSeeker: sp,
+		}
+	}
+
+	http.ServeContent(w, req, name, modtime, content)
+	//TODO: check for errors in ServeContent.. somehow
+
+	// If http.ServeContent can't figure out content size it won't write it to the
+	// responseWriter, Content-Length not being set is a good indicator of this
+	if req.Method != "HEAD" && w.Header().Get("Content-Length") == "" {
```
whyrusleeping: This feels hacky. I think in this case, ServeContent will have already written an error status and message to the responsewriter.

magik6k: This was here before sizeReadSeeker, isn't useful now (maybe as a fallback.. removing for now)

whyrusleeping:
@mib-kd743naq want to confirm that this works for you?

mib-kd743naq:
@whyrusleeping I won't be able to build a local web gateway for about couple days. But if I do my entire test will consist of firing up a gateway, and making sure I can open each file in [https://ipfs.io/ipfs/QmZL4wivfYBEUutxksKgbpPfFkj1tGNMESYLFF2MWcST8Y/Last entry in this subdir errors on latest go-ipfs
](https://ipfs.io/ipfs/QmZL4wivfYBEUutxksKgbpPfFkj1tGNMESYLFF2MWcST8Y/Last%20entry%20in%20this%20subdir%20errors%20on%20latest%20go-ipfs)

So if one of you can validate this faster and it works: I am happy ;)

whyrusleeping:
@mib-kd743naq alright, i'll give it a try!

whyrusleeping:
Appears to work over here (tested via curl)

PR #4324, opened 2017-10-19
default settings for the connection manager

whyrusleeping:
License: MIT

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -271,7 +271,10 @@ func (n *IpfsNode) startOnlineServices(ctx context.Context, routingOption Routin
 
 func constructConnMgr(cfg config.ConnMgr) (ifconnmgr.ConnManager, error) {
 	switch cfg.Type {
-	case "", "none":
+	case "":
+		// 'default' value is the basic connection manager
+		return connmgr.NewConnManager(config.DefaultConnMgrLowWater, config.DefaultConnMgrHighWater, config.DefaultConnMgrGracePeriod), nil
+	case "none":
 		return nil, nil
 	case "basic":
 		grace, err := time.ParseDuration(cfg.GracePeriod)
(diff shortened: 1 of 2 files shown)


core/core.go
```diff
@@ -271,7 +271,10 @@ func (n *IpfsNode) startOnlineServices(ctx context.Context, routingOption Routin
 
 func constructConnMgr(cfg config.ConnMgr) (ifconnmgr.ConnManager, error) {
 	switch cfg.Type {
-	case "", "none":
+	case "":
+		// 'default' value is the basic connection manager
+		return connmgr.NewConnManager(900, 1000, time.Second*20), nil
```
Stebalien: Can we avoid duplicating these constants? Actually, why are we not using values from the config here?

Stebalien: Also:

* We may want to take a look at the OS FD limits (and max out at half that). Does that make sense?
* We may want to make the low water mark lower. I'm worried that 900 will bounce of the ceiling. Maybe 500? Honestly, most peers will want to set these to 100 and 200.

whyrusleeping: this blank case is primarily for old nodes who don't have these fields in their configs, its an implicit default.

As for the numbers, maybe we could do 500-800:

A normal peer will probably want their DHT routing table peers plus 10%, 20 million peers in the network would result in a routing table of 400 peers, so being super conservative, we could set the low water to 500, and the high water at 800?

I'm hesitant to drop it too low, as we havent studied how things function in these environments yet.

Kubuxu: I would be for higher spread too. 500-800 sounds good, maybe a bit low until we flesh it out. Maybe 600-900?

Stebalien:
Modulo the code climate issues.

whyrusleeping:
@kubuxu go review that other PR ;)

PR #4326, opened 2017-10-19
lower yamux accept buffer size

Stebalien:
Should *massively* reduce the amount of memory used by each peer (by 60KiB).

License: MIT


whyrusleeping:
any idea why jenkins failed? (worrisome because we're wanting to use it by default)

Kubuxu:
It was the `ipfs shutdown`.

PR #4330, opened 2017-10-20
docker: move comments to avoid "empty continuation lines"

lgierth:
These apparantly will become Dockerfile syntax errors at some point in the future.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,24 +12,23 @@ ENV SRC_DIR /go/src/github.com/ipfs/go-ipfs
 COPY . $SRC_DIR
 
 # Build the thing.
+# Also: fix getting HEAD commit hash via git rev-parse.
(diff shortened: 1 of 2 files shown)


Stebalien:
Lovely, a language that discourages comments. Do docker scripts not allow one to define subscripts/functions? (we could also just make our own but multiple files can get annoying.

kpcyrd:
> Lovely, a language that discourages comments. Do docker scripts not allow one to define subscripts/functions?

Since go-ipfs uses multi-stage builds now, there is no real need for the long command chains anymore (I think Dockerfile.fast might be obsolete as well).

> (we could also just make our own but multiple files can get annoying.

I wouldn't recommend generating docker files, I was working on a project once that generated docker files using a template engine that was somewhat awkward to work with.

PR #4333, opened 2017-10-21
[WIP] Extract provider logic from BitSwap

### LINKED ISSUE #4311: respect provider strategies on initial adds
We have provider strategies that define which content gets provided. However, they are only respected on *reprovide*, for the initial adding of the content, everything currently gets provided regardless of whether or not it should be given the strategy.

We should rework the providers system to address this issue, moving provides out of the add path should also improve the performance of add.

magik6k:
&lt;WIP&gt;
Fixes #4311

<s>This is not even remotely close to done</s> Getting there, PR mainly for comments / tests / progress tracking

### DIFF
--- a/core/commands/add.go
+++ b/core/commands/add.go
@@ -313,13 +313,19 @@ You can now check what blocks have been created by:
 			}
 
 			// copy intermediary nodes from editor to our actual dagservice
-			_, err := fileAdder.Finalize()
+			nd, err := fileAdder.Finalize()
 			if err != nil {
 				return err
 			}
 
 			if hash {
 				return nil
+			}
+
+			if !local {
+				if err := n.Providers.ProvideRecursive(req.Context, nd, dserv); err != nil {
+					return err
+				}
 			}
 
 			return fileAdder.PinRoot()
--- a/core/commands/bitswap.go
+++ b/core/commands/bitswap.go
@@ -173,7 +173,6 @@ var bitswapStatCmd = &cmds.Command{
 			}
 
 			fmt.Fprintln(w, "bitswap status")
-			fmt.Fprintf(w, "\tprovides buffer: %d / %d\n", out.ProvideBufLen, bitswap.HasBlockBufferSize)
 			fmt.Fprintf(w, "\tblocks received: %d\n", out.BlocksReceived)
 			fmt.Fprintf(w, "\tblocks sent: %d\n", out.BlocksSent)
 			fmt.Fprintf(w, "\tdata received: %d\n", out.DataReceived)
--- a/core/commands/object/object.go
+++ b/core/commands/object/object.go
@@ -512,6 +512,12 @@ Available templates:
 			res.SetError(err, cmdkit.ErrNormal)
 			return
 		}
+
+		if err = n.Providers.Provide(node.Cid()); err != nil {
+			res.SetError(err, cmdkit.ErrNormal)
+			return
+		}
+
 		res.SetOutput(&Object{Hash: node.Cid().String()})
 	},
 	Marshalers: oldcmds.MarshalerMap{
@@ -606,6 +612,11 @@ func objectPut(ctx context.Context, n *core.IpfsNode, input io.Reader, encoding 
 	}
 
 	err = n.DAG.Add(ctx, dagnode)
+	if err != nil {
+		return nil, err
+	}
+
+	err = n.Providers.Provide(dagnode.Cid())
 	if err != nil {
 		return nil, err
 	}
--- a/core/corehttp/gateway_handler.go
+++ b/core/corehttp/gateway_handler.go
@@ -504,6 +504,11 @@ func (i *gatewayHandler) putHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if err := i.node.Providers.Provide(newcid); err != nil {
+		webError(w, "putHandler: Could not provide newnode: ", err, http.StatusInternalServerError)
+		return
+	}
+
 	i.addUserHeaders(w) // ok, _now_ write user's headers.
 	w.Header().Set("IPFS-Hash", newcid.String())
 	http.Redirect(w, r, gopath.Join(ipfsPathPrefix, newcid.String(), newPath), http.StatusCreated)
@@ -560,6 +565,11 @@ func (i *gatewayHandler) deleteHandler(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 
+		if err := i.node.Providers.Provide(c); err != nil {
+			webError(w, "Could not provide node", err, http.StatusInternalServerError)
+			return
+		}
+
 		pathpb, ok := pathNodes[j].(*dag.ProtoNode)
 		if !ok {
 			webError(w, "Cannot read non protobuf nodes through gateway", dag.ErrNotProtobuf, http.StatusBadRequest)
@@ -580,6 +590,11 @@ func (i *gatewayHandler) deleteHandler(w http.ResponseWriter, r *http.Request) {
 
 	// Redirect to new path
 	ncid := newnode.Cid()
+
+	if err := i.node.Providers.Provide(ncid); err != nil {
+		webError(w, "Could not provide node", err, http.StatusInternalServerError)
+		return
+	}
 
 	i.addUserHeaders(w) // ok, _now_ write user's headers.
 	w.Header().Set("IPFS-Hash", ncid.String())
--- a/exchange/bitswap/bitswap.go
+++ b/exchange/bitswap/bitswap.go
@@ -16,7 +16,6 @@ import (
 	notifications "github.com/ipfs/go-ipfs/exchange/bitswap/notifications"
 
 	delay "gx/ipfs/QmRJVNatYJwTAHgdSM1Xef9QVQ1Ch3XHdmcrykjP5Y4soL/go-ipfs-delay"
-	flags "gx/ipfs/QmRMGdC6HKdLsPDABL9aXPDidrpmEHzJqFWSvshkbn9Hj8/go-ipfs-flags"
 	metrics "gx/ipfs/QmRg1gKTHzc3CZXSKzem8aR4E3TubFhbgXwfVuWnSK5CC5/go-metrics-interface"
 	process "gx/ipfs/QmSF8fPo3jgVBAy8fpdjjYqgG87dkJgUprRBHRd2tmfgpP/goprocess"
 	procctx "gx/ipfs/QmSF8fPo3jgVBAy8fpdjjYqgG87dkJgUprRBHRd2tmfgpP/goprocess/context"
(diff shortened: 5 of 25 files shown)


blockservice/blockservice.go
```diff
@@ -38,34 +39,38 @@ type BlockService interface {
...
+	providers  providers.Interface
+
 	// If checkFirst is true then first check that a block doesn't
 	// already exist to avoid republishing the block on the exchange.
 	checkFirst bool
 }
 
 // NewBlockService creates a BlockService with given datastore instance.
-func New(bs blockstore.Blockstore, rem exchange.Interface) BlockService {
+func New(bs blockstore.Blockstore, rem exchange.Interface, p providers.Interface) BlockService {
```
whyrusleeping: The blockservice should not have to know anything about the providers system.

core/commands/bitswap.go
```diff
@@ -167,7 +176,7 @@ var bitswapStatCmd = &cmds.Command{
 			}
 			buf := new(bytes.Buffer)
 			fmt.Fprintln(buf, "bitswap status")
-			fmt.Fprintf(buf, "\tprovides buffer: %d / %d\n", out.ProvideBufLen, bitswap.HasBlockBufferSize)
+			fmt.Fprintf(buf, "\tprovides buffer: %d / %d\n", out.ProvideBufLen, provider.HasBlockBufferSize)
```
whyrusleeping: future note: we might want to have a separate command for the providers subsystem, and remove this information from here (since its no longer relevant to bitswap itself).

exchange/bitswap/bitswap.go
```diff
@@ -37,29 +36,18 @@ const (
 	// TODO: if a 'non-nice' strategy is implemented, consider increasing this value
 	maxProvidersPerRequest = 3
 	providerRequestTimeout = time.Second * 10
-	provideTimeout         = time.Second * 15
 	sizeBatchRequestChan   = 32
 	// kMaxPriority is the max priority as defined by the bitswap protocol
 	kMaxPriority = math.MaxInt32
+
+	self = peer.ID("")
```
whyrusleeping: eh?

exchange/offline/offline.go
```diff
@@ -72,3 +73,19 @@ func (e *offlineExchange) GetBlocks(ctx context.Context, ks []*cid.Cid) (<-chan
 func (e *offlineExchange) IsOnline() bool {
 	return false
 }
+
+type offlineProviders struct{}
```
whyrusleeping: I think we can get away with not even having this at all

exchange/providers/providers.go
```diff
@@ -0,0 +1,87 @@
+package providers
```
whyrusleeping: I think this should probably not live in the exchange subdirectory, maybe make a new top level directory (ideally, we can move it out of the go-ipfs repo, but that can wait)

magik6k:
This PR get's quite noisy with the `providers` interface propagation, I'm going to defer the strategies to the next PR while this will focus on extracting provider logic from bitswap.

path/resolver.go
```diff
@@ -35,12 +36,13 @@ func (e ErrNoLink) Error() string {
 // TODO: now that this is more modular, try to unify this code with the
 //       the resolvers in namesys
 type Resolver struct {
-	DAG dag.DAGService
+	DAG  dag.DAGService
+	prov providers.Interface
 
-	ResolveOnce func(ctx context.Context, ds dag.DAGService, nd node.Node, names []string) (*node.Link, []string, error)
+	ResolveOnce func(ctx context.Context, ds dag.DAGService, prov providers.Interface, nd node.Node, names []string) (*node.Link, []string, error)
```
magik6k: https://github.com/ipfs/go-ipfs/pull/4333/files#diff-cb03f3339782273ac5936e321903f907R16 needs this, though it's probably safe to not pass the interface here and use `offline.Providers` there.

exchange/bitswap/network/ipfs_impl.go
```diff
@@ -174,7 +178,7 @@ func (bsnet *impl) FindProvidersAsync(ctx context.Context, k *cid.Cid, max int)
 
 // Provide provides the key to the network
 func (bsnet *impl) Provide(ctx context.Context, k *cid.Cid) error {
-	return bsnet.routing.Provide(ctx, k, true)
```
whyrusleeping: Does this call here actually send out a message?

exchange/bitswap/stat.go
```diff
@@ -20,7 +20,7 @@ type Stat struct {
 
 func (bs *Bitswap) Stat() (*Stat, error) {
 	st := new(Stat)
-	st.ProvideBufLen = len(bs.newBlocks)
+	st.ProvideBufLen = -1
```
whyrusleeping: its probably fine to remove this entirely

exchange/bitswap/network/ipfs_impl.go
magik6k: It probably does, I'll try to write a test for that

whyrusleeping: We probably want to move away from that now. The equivalent of setting the bsnet.routing.Provide final parameter to 'false'

magik6k: Agreed, though it may be worth making this a configurable thing.

It would be best to have the strategies apply here, though it would require a lot of refactoring to make it fit into the code base nicely.

core/commands/object/object.go
```diff
@@ -509,6 +509,13 @@ Available templates:
 			res.SetError(err, cmdkit.ErrNormal)
 			return
 		}
+
+		err = n.Providers.Provide(k)
+		if err != nil {
```
ForrestWeston: As you have done in other files, these error checks should be condensed to a single line. This will keep `err` in the scope of the if statement.

importer/helpers/helpers.go
```diff
@@ -110,7 +110,11 @@ func (n *UnixfsNode) AddChild(child *UnixfsNode, db *DagBuilderHelper) error {
 		return err
 	}
 
-	_, err = db.batch.Add(childnode)
+	_, err = db.batch.Add(childnode) //TODO: do we need to provide here?
+	err = db.prov.Provide(childnode.Cid())
```
ForrestWeston: Minor - We should be consistent with how we pass the values returned from `Add` to `Provide`. Here we are ignoring the returned `cid` since it is available on the `childnode`. Elsewhere we use a [separate variable](https://github.com/ipfs/go-ipfs/pull/4333/files#diff-19ceb36d78bde310c54cc44a95bec44cR575) to store the `cid` when calling `Provide`. I am not sure which way is more appropriate, what do you think?

magik6k: We should use `node.Cid()`, Add won't be returning cid after refactor in https://github.com/ipfs/go-ipld-format/pull/8 happens, good catch

core/commands/add.go
```diff
@@ -317,6 +317,12 @@ You can now check what blocks have been created by:
 				return nil
 			}
 
+			if !local {
+				if err := n.Providers.ProvideRecursive(req.Context(), nd, dserv); err != nil {
```
magik6k: @whyrusleeping I think it's the last thing I wanted to discuss after removing the [WIP] tag:

Before strategies are implemented (next PR), this will happen on the end of add causing weird hang for bigger adds. Should this be done in background for now (`go Provide..`)?

magik6k:
Rebased

core/commands/add.go
whyrusleeping: I think its probably fine for now. But in the future I think we will want a separate progress bar for providing.

whyrusleeping:
Lets get this one in for 0.4.15. I think we're going to want to make sure that we change the way progress bars work for ipfs add before shipping this into a release, and I want to get 0.4.14 out soon

whyrusleeping:
@magik6k i'm sorry. Thats an ugly set of merge conflicts... Could you rebase this one?

magik6k:
Squashed most of the commits to make it less painful to resolve conflicts.

I think it's ready for a review. 2 things to note: 
* The point of this PR is to extract providing logic from bitswap, it shouldn't change what gets provided
* `ipfs add` will do providing after adding stuff, this can take a while and could use a separate progress bar, but it's quite likely to be removed when provider strategies get implemented. I think that this ux can be somewhat improved by looking if we run add on a daemon and do the providing background.

providers/providers.go
```diff
@@ -0,0 +1,196 @@
...
+	select {
+	case p.newBlocks <- b:
+	// send block off to be provided to the network
+	case <-p.process.Closing():
+		return p.process.Close()
+	}
+	return nil
+}
+
+func (p *providers) provideRecursive(ctx context.Context, n ipld.Node, serv ipld.NodeGetter, done *cid.Set) error {
```
Stebalien: We'll have to be careful not to pass an online dag service in here.

Stebalien:
What is the end goal here? Is the end goal to have the provider service remember which blocks we should be providing? That is, are we planning on supporting options like `ipfs add --provide=[root|files|...]`?

As implemented, this patch forces the programmer to remember to call the provider system every time they add a block they want to provide to the network. Unless we're going to allow the programmer to make provider decisions at that point, this feels like a step backward. If this is the case, I'd rather add some form of EventedBlockstore interface and have a provider service subscribe to events from this service.

On the other hand, providing a way for users to specify how/when they want to provide blocks *is* useful and could further be used for access control.

whyrusleeping:
@Stebalien 

> That is, are we planning on supporting options like `ipfs add --provide=[root|files|...]`?

Yes, we're planning on adding that.

Here's an excerpt from another doc i'm writing:

> Content Routing, (most importantly the providing side of things) should not be inside bitswap, and bitswap should not ever block on some external service (like the DHT) being slow. This is currently a problem as for every block bitswap receives, it pushes that block to the provider service. The provider service is slow, and therefore must either buffer infinitely, or apply backpressure. That backpressure is significantly slowing down bitswap at the moment. There is already a PR open by @magik6k that separates these pieces out, but does not change the behavior. The next step after that PR is merged is to change the way we reason about providing. Every block added through the blockservice should not cause a separate providing event. Higher layers should inform the provider subsystem that they wish to provide a given graph, in a certain way. Then it is up to the provider subsystem to traverse the graph as needed and send out the appropriate provider messages. This way, the provider subsystem does not get one independent call for every node in the graph (requiring O(n) memory to buffer and track), it gets a single call for the root of the graph. This will help us avoid needing to apply backpressure on bitswap to avoid using too much memory there.

whyrusleeping:
In general, having 'write data to the local store' also mean 'broadcast the fact that we have that piece of data' is pretty bad, both in terms of privacy and performance. 'forcing' the programmer to remember to call `Provide` is the right thing to do IMO.

Stebalien:
> In general, having 'write data to the local store' also mean 'broadcast the fact that we have that piece of data' is pretty bad, both in terms of privacy and performance. 'forcing' the programmer to remember to call Provide is the right thing to do IMO.

As long as we bring the reprovider in line with this. I just don't want to end up in an inconsistent situation where users *don't* explicitly provide but then we end up providing in the background anyways.

whyrusleeping:
Yeah, the reprovider gets folded into the 'provider' as part of this

providers/providers.go
```diff
@@ -0,0 +1,196 @@
...
+	Cid *cid.Cid
+	Ctx context.Context
+}
+
+// Interface defines providers interface to libp2p routing system
+type Interface interface {
+	Provide(k *cid.Cid) error
+	ProvideRecursive(ctx context.Context, n ipld.Node, serv ipld.NodeGetter) error
+
+	FindProviders(ctx context.Context, k *cid.Cid) error
```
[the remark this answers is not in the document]
magik6k: It also does dht querying. I guess what you mean that instead of this method bitswap should use `FindProvidersAsync` and only send wantlists to those peers?

For now, I'd rename the method, and remove it as a part of other PR

core/corehttp/gateway_handler.go
```diff
@@ -560,6 +565,11 @@ func (i *gatewayHandler) deleteHandler(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 
+		if err := i.node.Providers.Provide(c); err != nil {
```
michaelavila: @magik6k I have a question: why `c` here? It seems like this will try to provide the same `Cid` for each iteration of the loop. Is that intentional? If so, I'm curious why. I'm using this PR to inform a similar change I'm making and am caught up on this detail. Thanks.

magik6k: Yeah, this is wrong. This should be either `newnode.Cid()` or more likely not at all here (`Provide` below should handle this, not 100% sure)

Stebalien:
@magik6k given the work in https://github.com/ipfs/go-ipfs/issues/5774, should we close this?

magik6k:
Yep, don't delete the branch yet though

PR #4340, opened 2017-10-24
extract go-ipfs-addr

whyrusleeping:
I was feeling like we were making new repos fast enough. This should help resolve that problem, at least for now.

License: MIT


Stebalien:
Beautiful 😂. Another repo is born 🎉. The holy circle of left-pad would be proud 🛐.

PR #4347, opened 2017-10-27
chunker long description first shot

thisconnect:
Part II of #4276 adds the long description

### DIFF
--- a/core/commands/add.go
+++ b/core/commands/add.go
@@ -71,6 +71,30 @@ its filename. For example:
 You can now refer to the added file in a gateway, like so:
 
   /ipfs/QmaG4FuMqEBnQNn3C8XJ5bpW8kLs7zq2ZXgHptJHbKDDVx/example.jpg
+
+The chunker option, '-s', defines the chunking strategy for how 
+ipfs puts larger files into blocks. Blocks with same content can
+be deduplicated. The default is a fixed block size of
+256 * 1024 bytes, 'size-262144'. A variable block size can be
+specified by using the rabin chunker, rabin-[min]-[avg]-[max].
+Using different chunking strategies will produce different hashes 
+for the same file.
+
+  > ipfs add --chunker=size-1024 ipfs-logo.svg
+  added QmZ9KNxxeeLCPrdceBjvVcmSyLbUWMF4ceKDP8yboLNoHT ipfs-logo.svg
+  > ipfs add --chunker=rabin-128-256-2048 ipfs-logo.svg
+  added QmQN64Mbj1WK8wAej5MoFkHkD3aTtSGzGnAFXomdVHNNhs ipfs-logo.svg
+
+You can now check what blocks have been created by:
+
+  > ipfs object links QmQN64Mbj1WK8wAej5MoFkHkD3aTtSGzGnAFXomdVHNNhs
+  Qmb3Wzsmy9RXkyQhERdPd93mBP2jTRUjgVBwx5yWJR5b6k 223   
+  Qmc7ikMLM1SS88YxxogRtjAyuLLfR85ZDKjfy7ahZYD38h 141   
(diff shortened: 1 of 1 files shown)


thisconnect:
First time trying to write some cli doc help. Please feel free to close, change, request change, delete, remove, update and so on :)

thisconnect:
cc @Stebalien

core/commands/add.go
```diff
@@ -71,6 +71,30 @@ its filename. For example:
 You can now refer to the added file in a gateway, like so:
 
   /ipfs/QmaG4FuMqEBnQNn3C8XJ5bpW8kLs7zq2ZXgHptJHbKDDVx/example.jpg
+
+The chunker option, '-s', defines the chunking strategy for how 
+ipfs puts larger files into blocks. Blocks with same content can
+be deduplicated. The default is a fixed block size of
+256 * 1024 bytes, 'size-262144'. A variable block size can be
```
whyrusleeping: Maybe instead of a variable block size, "Content defined chunking can be used in the form of the rabin chunker..."

I'm not super sure on how to best phrase this though. cc @lgierth @Stebalien

whyrusleeping:
@thisconnect thank you so much for this!

core/commands/add.go
Stebalien: First, for the `size-` chunker, I'd note that alternative sizes can be used (just for completeness).

As for wording, I'd use the second person: "Alternatively, you can use the rabin chunker for content defined chunking by specifying rabin-[min]-[avg]-[max] (where min/avg/max refer to the resulting chunk sizes)."

I don't think we really need to go into the respective merits here.

core/commands/add.go
```diff
@@ -71,6 +71,31 @@ its filename. For example:
...
+You can now check what blocks have been created by:
+
+  > ipfs object links QmQN64Mbj1WK8wAej5MoFkHkD3aTtSGzGnAFXomdVHNNhs
+  Qmb3Wzsmy9RXkyQhERdPd93mBP2jTRUjgVBwx5yWJR5b6k 223
+  Qmc7ikMLM1SS88YxxogRtjAyuLLfR85ZDKjfy7ahZYD38h 141
+  QmeUiJHEXZnEwjpCfUUSkDuXFk2nuT6gALCeUXtZkRvcbj 858
+  QmWwiQMNhpjq9kFUiuFbSQoqrQNTVjwo1wgbbx8wPr9PjB 1040
+  QmNPg3SUphwfoJwveDyMRjDDaoitTCE7g9UUqYod8gPwz5 431
+  QmPzBVWJtMPdeaoJK55rua7LgbEsjHseMHsUSUrPJ3AncS 278
+  QmQB28iwSriSUSMqG2nXDTLtdPHgWb4rebBrU7Q1j4vxPv 338
```
thisconnect: Do you think the last part with `ipfs object links` is useful?

core/commands/add.go
```diff
@@ -71,6 +71,31 @@ its filename. For example:
 You can now refer to the added file in a gateway, like so:
 
   /ipfs/QmaG4FuMqEBnQNn3C8XJ5bpW8kLs7zq2ZXgHptJHbKDDVx/example.jpg
+
```
thisconnect: different wording
The chunker option, '-s', specifies the chunking strategy and size for how to store and deduplicate larger files into blocks.

core/commands/add.go
whyrusleeping: Yeah, i do think this is useful. Maybe do it so that its fewer nodes, and then also show one for the fixed size chunker too.

core/commands/add.go
whyrusleeping: maybe just this? 
'...specifies the chunking strategy that dictates how to break files into blocks'

thisconnect: Yes that's better, should be updated in last commit

thisconnect:
codecov doesn't like it :)

whyrusleeping:
oh codecov can go home, theres no code being changed here!

PR #4348, opened 2017-10-28
optimise pin update command

ianopolous:
This handles merkle links that aren't named. And improves the
Peergos usage from worst case 30s to ~20ms

### DIFF
--- a/merkledag/utils/diffenum.go
+++ b/merkledag/utils/diffenum.go
@@ -72,14 +72,15 @@ func getLinkDiff(a, b node.Node) []diffpair {
 		inb[l.Cid.KeyString()] = l
 	}
 	for _, l := range a.Links() {
-		ina[l.Cid.KeyString()] = l
-		if inb[l.Cid.KeyString()] == nil {
+	        var key = l.Cid.KeyString()
+		ina[key] = l
+		if inb[key] == nil {
 			aonly = append(aonly, l.Cid)
 		}
 	}
 
 	var out []diffpair
-	var aindex = 0
+	var aindex int
 
 	for _, l := range b.Links() {
 		if ina[l.Cid.KeyString()] != nil {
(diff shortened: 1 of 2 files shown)


merkledag/utils/diffenum.go
```diff
@@ -65,27 +65,34 @@ type diffpair struct {
...
 func getLinkDiff(a, b node.Node) []diffpair {
-	have := make(map[string]*node.Link)
-	names := make(map[string]*node.Link)
+	ina := make(map[string]*node.Link)
+	inb := make(map[string]*node.Link)
+	var aonly []*cid.Cid
+	for _, l := range b.Links() {
+		inb[l.Cid.KeyString()] = l
+	}
 	for _, l := range a.Links() {
```
whyrusleeping: It will be faster to just call `KeyString()` once, bind it to a local variable here. (also easier on memory)

merkledag/utils/diffenum.go
```diff
@@ -65,27 +65,34 @@ type diffpair struct {
...
-		have[l.Cid.KeyString()] = l
-		names[l.Name] = l
+		ina[l.Cid.KeyString()] = l
+		if inb[l.Cid.KeyString()] == nil {
+			aonly = append(aonly, l.Cid)
+		}
 	}
 
 	var out []diffpair
+	var aindex = 0
```
whyrusleeping: you can just say `var aindex int`

merkledag/utils/diffenum.go
ianopolous: done

whyrusleeping:
I think we should make sure that the name pairing still works properly with this change, so that links with the same name (or 'path' within the object) still create the same resulting 'change' object

ianopolous:
I believe that the name matching property will be maintained with this pr, as long as the Links() function is only a function of the ipld topology of the object. I will write a test to verify this.

merkledag/utils/diffenum.go
```diff
@@ -65,27 +65,35 @@ type diffpair struct {
 // getLinkDiff returns a changeset between nodes 'a' and 'b'. Currently does
 // not log deletions as our usecase doesnt call for this.
 func getLinkDiff(a, b node.Node) []diffpair {
-	have := make(map[string]*node.Link)
-	names := make(map[string]*node.Link)
```
Stebalien: Looks right. The `Name` property of links is a protonode thing (that needs to die).

ianopolous:
Hooray! My first commits to go-ipfs! :-)

PR #4361, opened 2017-11-01
misc: add note about updating before reporting the issue

Kubuxu:
In many cases issues reported are present only in older versions.
I've set 0.4.10 as a cut of as it is quite recent and contains many
fixes and performance improvements.

License: MIT

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,4 +1,9 @@
-<!-- Output From `ipfs version --all` -->
+<!--
+Output From `ipfs version --all`
+
+If your version is older than *0.4.11* please update
+and re-check if the problem persists.
+-->
 #### Version information:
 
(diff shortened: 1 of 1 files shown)


ISSUE_TEMPLATE.md
```diff
@@ -1,4 +1,9 @@
-<!-- Output From `ipfs version --all` -->
+<!--
+Output From `ipfs version --all`
+
+If your version is older that *0.4.10* please update
```
magik6k: s/that/than

Stebalien:
Quite a few (important) fixes made it into 4.11. Any reason not to use that?

Kubuxu:
because 0.4.11 is not released yet.

whyrusleeping:
But 0.4.11 is released: https://dist.ipfs.io/#go-ipfs

Kubuxu:
Fixing.

Kubuxu:
fixed

PR #4364, opened 2017-11-02
gx: update go-ds-badger to 1.1.0

magik6k:
This once again breaks the on-disk format, uses badger v1.0.0.

@whyrusleeping Should I write an upgrade script for the version in 0.4.12-rc1 too, or is just supporting the one in 0.4.11 fine?

### DIFF
--- a/package.json
+++ b/package.json
@@ -460,9 +460,9 @@
     },
     {
       "author": "magik6k",
-      "hash": "QmNm2bfBCtEzmpxVzaW2FZGzWx2KWHprfd27aWoXptrAGa",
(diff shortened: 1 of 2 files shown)


whyrusleeping:
@magik6k is this actually 1.0.0? and we won't have to worry about further random changes?

magik6k:
Yep, it is after the 1.0 tag which was made recently - https://github.com/dgraph-io/badger/tree/v1.0.0

whyrusleeping:
Alright, then lets ship this and not worry about the intermediate versions.

Issue #3267, opened 2016-09-28
Migrate from stdlib syscall to golang.org/x/sys


Kubuxu:
The stdlib won't be updated anymore and in case of Go 2 (if it ever is created) will be removed.

The docs recommend moving to golang.org/x/sys library.
It is something we should do in future but there is no hurry for that.

Moving to golang.org/x/sys includes moving go-ipfs and all of dependencies controlled by us.

whyrusleeping:
lol, we have a technical debt label now. Great.

djdv:
I'm not sure if there are any other cases like this but the standard itself suggests using /x/sys for at least this structure and it's methods on Windows, due to security concerns.
https://golang.org/src/syscall/dll_windows.go#L182
https://golang.org/src/syscall/dll_windows.go#L39

The calls in IPFS should probably be checked against the latest documentation to see if there are any notes of problems that are resolved in /x/ but not in std.

Kubuxu:
@Stebalien I recall you working on this, have you removed all usages?

Stebalien:
Not yet. Sorry, some of those PRs dropped off of my radar. I've submitted a PR to do take care of this in IPFS but we still need to merge/bubble those PRs (Travis is failing on go-reuseport so I'm not entirely sure what's wrong there...).

Also, FYI, we won't be able to *completely* switch. Not everything we use is exported by `golang.org/x/sys/x/unix` (although most of it is).

Stebalien:
We've now switched over as much as we can.

PR #4372, opened 2017-11-06
switch from syscall to golang.org/x/sys

Stebalien:
(mostly)

syscall has been frozen and mostly deprecated. We can't *entirely* switch but
this brings us closer.

progress on #3267

(we still need to pull through some go-reuseport changes)

### DIFF
--- a/cmd/ipfs/ulimit_freebsd.go
+++ b/cmd/ipfs/ulimit_freebsd.go
@@ -4,7 +4,8 @@ package main
 
 import (
 	"fmt"
-	"syscall"
+
+	unix "gx/ipfs/QmPXvegq26x982cQjSfbTvSzZXn7GiaMwhhVPHkeTEhrPT/sys/unix"
 )
 
 func init() {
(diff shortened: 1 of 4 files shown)


Stebalien:
Note: This now leaves a lot of uses of `syscall` in place. Specifically, errors and signals (`sys/unix` doesn't work on windows, obviously).

However, it still makes platform-specific code use `sys/*` instead of `syscall`.

whyrusleeping:
@Stebalien got a merge conflict here after the commands lib refactor. Mind fixing that up?

cc @keks as this looks like it might need to be changed in the newly extracted code

Stebalien:
@keks, fixed in https://github.com/ipfs/go-ipfs-cmdkit/pull/2

@whyrusleeping rebased

whyrusleeping:
@Stebalien you'll have to do a gx update after that PR merges, yeah?

Stebalien:
Yes but this can be merged first. We don't have to switch over all in one go.

PR #4377, opened 2017-11-09
Ipfs v0.4.12

whyrusleeping:
License: MIT


CHANGELOG.md
```diff
@@ -1,6 +1,46 @@
...
+  (including js-ipfs nodes)
+
+Because of this, we selected 600 as a 'LowWater' number, and 900 as a
+'HighWater' number to avoid having to clear out connections too frequently. 
+You can configure different numbers as you see fit via the `Swarm.ConnMgr`
+field in your ipfs config file. See
+[here](https://github.com/ipfs/go-ipfs/blob/master/docs/config.md#connmgr) for
+more details.
+
+Disk utilization during `ipfs add` has been optimized by doing batch writes in
```
Stebalien: Technically, this currently only happens on a per-file basis. We still flush after adding each file. We should also be *adding* files in parallel (or reusing the batch but that makes it hard to report when a file has actually been added).

Probably don't need to explain this though, I just want to make sure we're all on the same page.

CHANGELOG.md
```diff
@@ -1,6 +1,46 @@
 # go-ipfs changelog
 
-## 0.4.12-rc2 2017-10-26
+## 0.4.12 2017-11-09
```
Stebalien: Do we want to mention badger breaking changes since the last release?

whyrusleeping: Oh yeah, we should note that. @magik6k people should be able to use your tool soon, right?

whyrusleeping: It doesnt have to block the release since its an experimental thing, but we can let people know such a tool is coming soon.

CHANGELOG.md
whyrusleeping: If you want to take a stab at rewording this then please do.

CHANGELOG.md
magik6k: Yep, I'll we just merged https://github.com/ipfs/badgerds-upgrade/pull/1, It works reasonably well for me(i.e. I didn't manage to break it) We can point people there from changelog or make a descriptive error in ds-badger if you don't mind waiting till tomorrow.

lgierth: > We can point people there from changelog or make a descriptive error in ds-badger if you don't mind waiting till tomorrow.

Also update docs/experimental-features.md with that info

CHANGELOG.md
Stebalien: I've updated it. Does the new version work for you?

CHANGELOG.md
magik6k: Done in https://github.com/ipfs/go-ipfs/pull/4379

PR #4387, opened 2017-11-13
WIP: experiment in making protobuf nodes with no linknames resolvable

whyrusleeping:
Was thinking about ipld selectors and unixfs and realized I couldnt even reference unixfs sub-files with a path. This is one possible solution...

cc @Stebalien @kevina @lgierth @Kubuxu @diasdavid 

License: MIT

### DIFF
--- a/merkledag/node.go
+++ b/merkledag/node.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strconv"
 
 	cid "gx/ipfs/QmNp85zy9RLrQ5oQD4hPyS39ezrrXpcaa7R4Y9kxdWQLLQ/go-cid"
 	node "gx/ipfs/QmPN7cwmpcc4DWXb4KTB9dNAJgjuPY69h3npsMfhRrQL9c/go-ipld-format"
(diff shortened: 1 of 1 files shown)


Stebalien:
I'm not sure if I understand. UnixFS `Name` fields are not actually used (or aren't supposed to be used, at least) in IPLD paths (without transformations). The path through a unixfs object should be something like `/Links/0/Cid/Links/2/Cid/Links/24/Cid` etc. (by array offset). With IPLD selectors, we'd probably use something like `/Links/[Name="MyName"]/Cid/`... Although this gets tricky with, e.g., HAMPTs...

whyrusleeping:
@Stebalien this is the case where there are no names, and its just the links array. Currently 'resolve' through the protobuf nodes does use the names of the links, otherwise all sorts of things would stop working.

lgierth:
>  The path through a unixfs object should be something like /Links/0/Cid/Links/2/Cid/Links/24/Cid etc. (by array offset)

Uuugh I really disagree - that'd break so many things (and it's an abomination). Just let dag-pb be a custom resolver.

Stebalien:
Well, that's path resolution, a layer on top of IPLD (well should be, that's what we want to use IPLD transforms eventually). As it stands, one can't, e.g., address the data section via that pathing system (or the link sizes, etc...).

Yes, this is utterly broken.

We could make this situation suck slightly less by re-interpreting the ProtoNode format as:

```js
{
    "Data": [bytes],
    "LinkMap": { "<name>": CID, ... }, // non-empty, unique names
    "Links": [ { "Name": "<name>", "Cid": CID, "Size": SIZE }, ... }, // the full array of links
}
```

However, this is one of the reasons I brought up the possibility of reinterpreting protobuf nodes as an entirely new format (one that also decodes the embedded unixfs protobuf) in the context of `/ipfs`. That is, instead of CIDv0 meaning "protonode" it would mean "contextual node". `/ipfs/QmId` (no codec) would be interpreted as `/ipfs/zSomeCidWithACodec`. This would allow us to, e.g., come up with a mapping between protonode+unixfs and ipld-unixfs (given them the same logical structure). 

However, the reactions I got were generally "that's insane" and ranged from "let's do it" to "go home".

Stebalien:
It's been a while, but things here have settled a bit:

1. We've introduced the concept of "ADLs", or "Advanced Data Layouts". The IPFS (well, UnixFS) ADL defines pathing (map lookup) semantics both for normal and sharded directories.
2. Lookup within DagPB will not be special. Children will be looked up via, e.g., `/ipld/Qm.../Links/0/Hash/`.

PR #4391, opened 2017-11-14
Ipfs v0.4.13-rc1

whyrusleeping:
Trying to get this out ASAP

License: MIT


kevina:
As a technicality I think 0.4.12.1-rc1 would be a better name as this looks like a purely bug fix release.

lgierth:
Yeah we don't currently support anything other than X.Y.Z

PR #4392, opened 2017-11-15
Feat/log/uuid

ForrestWeston:
Add an uuid/requestId to root context - this will add a requestId to event logs resulting from a command
Add event logs around path resolution - said logs make use of requestId

### DIFF
--- a/commands/http/handler.go
+++ b/commands/http/handler.go
@@ -138,8 +138,8 @@ func (i internalHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 
 	ctx, cancel := context.WithCancel(node.Context())
-	ctx = logging.ContextWithLoggable(ctx, loggables.Uuid("requestId"))
 	defer cancel()
+	ctx = logging.ContextWithLoggable(ctx, loggables.Uuid("requestId"))
 	if cn, ok := w.(http.CloseNotifier); ok {
 		clientGone := cn.CloseNotify()
 		go func() {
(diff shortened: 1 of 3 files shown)


path/resolver.go
```diff
@@ -135,6 +135,7 @@ func (s *Resolver) ResolvePathComponents(ctx context.Context, fpath Path) ([]nod
 	if err != nil {
 		return nil, err
 	}
+	defer log.EventBegin(ctx, "resolvePathComponents", logging.LoggableMap{"parts": parts, "cid": h}).Done()
```
ForrestWeston: In the case that there is an error in the execution of this method, do we want to add a loggable error to the event log (see L143)?

Stebalien: IMO, yes. We'd probably also want to log the error from L146 as well. Also, this event should probably go at the top and include any errors from `SplitAbsPath`.

Kubuxu: You can save the result value of `EventBegin` and add additional data to it.

commands/http/handler.go
```diff
@@ -137,6 +138,7 @@ func (i internalHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 
 	ctx, cancel := context.WithCancel(node.Context())
+	ctx = logging.ContextWithLoggable(ctx, loggables.Uuid("requestId"))
```
whyrusleeping: I would move this down below the defer cancel.

whyrusleeping:
@VictorBjelkholm whatsup with jenkins?

PR #4393, opened 2017-11-15
Announce and discover swarm addresses over simple UDP broadcasts

ehmry:
This is related to https://github.com/libp2p/go-libp2p/pull/243, but I don't know how to drive gx to make the two work.

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -122,6 +122,7 @@ type IpfsNode struct {
 	Resolver   *path.Resolver       // the path resolution system
 	Reporter   metrics.Reporter
 	Discovery  discovery.Service
+	Broadcast  discovery.Service
 	FilesRoot  *mfs.Root
 
 	// Online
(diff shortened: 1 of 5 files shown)


core/core.go
```diff
@@ -152,7 +153,7 @@ type Mounts struct {
 	Ipns mount.Mount
 }
 
-func (n *IpfsNode) startOnlineServices(ctx context.Context, routingOption RoutingOption, hostOption HostOption, do DiscoveryOption, pubsub, mplex bool) error {
+func (n *IpfsNode) startOnlineServices(ctx context.Context, routingOption RoutingOption, hostOption HostOption, do DiscoveryOption, bo DiscoveryOption, pubsub, mplex bool) error {
```
Stebalien: Any reason to not just pass in an array of these? That would allow us to avoid changing this code again.

Kubuxu:
Closing as https://github.com/libp2p/go-libp2p/pull/243 was dropped.

PR #4398, opened 2017-11-18
fix hamt delete issue

whyrusleeping:
Found this while writing another hamt (using ipld :) )

License: MIT

### DIFF
--- a/unixfs/hamt/util.go
+++ b/unixfs/hamt/util.go
@@ -2,6 +2,7 @@ package hamt
 
 import (
 	"math/big"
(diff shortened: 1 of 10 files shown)


unixfs/hamt/util.go
```diff
@@ -2,6 +2,7 @@ package hamt
 
 import (
 	"math/big"
+	"math/bits"
```
Kubuxu: If we are adding this we should increase checked Go version to 1.9 in package.json and Makefiles.

Issue #4325, opened 2017-10-19
Annonce option is not documented in docs/config.md


Kubuxu:

whyrusleeping:
cc @lgierth

PR #4401, opened 2017-11-20
updating Addresses section

leerspace:
config.md: updating the Addresses section

Adding documentation on Announce and NoAnnounce.

Fixing typo in API default value.

License: MIT
Signed-off-by: John Reed <john@re2d.xyz>

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -43,7 +43,7 @@ Default:
 ```
 
 - `Announce`
-Array of swarm addresses to announce to the network.
+If non-empty, this array specifies the swarm addresses to announce to the network. If empty, the daemon will announce inferred swarm addresses.
 
 Default: `[]`
 
--- a/docs/config.md
+++ b/docs/config.md
@@ -24,7 +24,7 @@ Contains information about various listener addresses to be used by this node.
 - `API`
 Multiaddr describing the address to serve the local HTTP API on.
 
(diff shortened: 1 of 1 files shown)


leerspace:
Ref: #3613

Closes #4325

docs/config.md
```diff
@@ -42,6 +42,16 @@ Default:
 ]
 ```
 
+- `Announce`
+Array of swarm addresses to announce to the network.
+
+Default: `[]`
```
Stebalien: This is a bit misleading. We announce all public addresses by default.

leerspace: Would it be better to describe this as "Array of swarm addresses to announce to the network in addition to the ones that are inferred by default."?

lgierth: If this is non-empty, we'll only announce what's in there. If it's empty, we'll announce the addresses that were inferred.

leerspace: My bad. I was mixing up the "Swarm listening" lines and the "Swarm announcing" lines of the daemon output into a single bucket.

whyrusleeping:
I wonder if gitcop just doesnt like windows line endings?

leerspace:
Maybe. That or it just doesn't like me :).

FWIW this PR was made entirely using GitHub's built in forking/branching/editor tools, so that might also be a factor.

whyrusleeping:
ah, interesting... oh well.

PR #4403, opened 2017-11-21
Update all the things.

Stebalien:
**secio**

* Do the handshake synchronously. This means we'll never see a connection until it has actually been fully verified. It should also prevent us from blocking when checking the remote peer's ID.

**conn, transports, etc.**

* Accept concurrently. This prevents synchronous secio handshakes from blocking
  other secio handshakes.
* Use contexts for timeouts. One can now accurately set dial contexts via timeouts instead of setting a global dial timeout. We still have a global 60s dial timeout for sanity.
* Correctly set the source address.

**reuseport**

* Move away from syscall to sys. This should fix support on some platforms, e.g., linux-arm64.
* Make reuseport compile on netbsd
* Make checking for reuseport support faster.
* Probably more...

**go-msgio/mpool**

* Remove the locks and allocate pools up front (much faster). They're not *that* large.
* Update multiple dependent packages to share a global msgio buffer pool instead of copying it. I'm not entirely sure how we weren't deadlocking and panicking all over the place; we were copying locks.

**cmds, cmdkit**

* Rename DefaultVal -> Default, Default -> WithDefault

Depends on https://github.com/ipfs/go-ipfs-cmds/pull/17


Stebalien:
Merge #4042 first and then I'll fix this.

whyrusleeping:
@Stebalien done did it

Stebalien:
@whyrusleeping done.

PR #4406, opened 2017-11-21
Fix two race conditions (and possibly go routine leaks) in commands

### LINKED ISSUE #4405: Remove concurrent uses of `go-ipfs-cmds.ResponseEmitter`
It's *not* thread-safe.

* [ ] `core/commands/pin.go`: addPinCmd (concurrent SetError)
* [ ] `core/commands/add.go`: AddCmd (PostRun, concurrent SetError)

All other uses of SetError, SetOutput, and Emit appear to be fine.

Stebalien:
fixes #4405

### DIFF
--- a/core/commands/pin.go
+++ b/core/commands/pin.go
@@ -90,15 +90,14 @@ var addPinCmd = &cmds.Command{
 		v := new(dag.ProgressTracker)
 		ctx := v.DeriveContext(req.Context())
 
-		ch := make(chan []*cid.Cid)
+		type pinResult struct {
+			pins []*cid.Cid
+			err  error
+		}
+		ch := make(chan pinResult, 1)
 		go func() {
-			defer close(ch)
 			added, err := corerepo.Pin(n, ctx, req.Arguments(), recursive)
-			if err != nil {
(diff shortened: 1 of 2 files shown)


core/commands/pin.go
```diff
@@ -90,32 +90,31 @@ var addPinCmd = &cmds.Command{
 		v := new(dag.ProgressTracker)
 		ctx := v.DeriveContext(req.Context())
 
-		ch := make(chan []*cid.Cid)
+		type pinResult struct {
+			pins []*cid.Cid
+			err  error
+		}
```
keks: Huh, I did not know you could define types in functions. Neat!

Stebalien: You want to have some *real* fun?

```go
ch := make(chan struct{pins []*cid.Cid, err error}, 1)
//...
ch <- struct{pins []*cid.Cid, err error}{err: errors.New("😎")
```

Kubuxu: I have some other trick for you:

```go
type System struct {
    cntrSafe struct {
        sync.Mutex
        n int
    }
    mapSafe struct {
        sync.Mutex
        m map[int]int
    }
}
````

PR #4413, opened 2017-11-21
Fix some goroutine leaks in commands

Stebalien:
There's a lot left but I'll just make an issue and advertise it on IRC.


Stebalien:
Question: Should we be calling `res.SetError(req.Context().Err(), cmdkit.ErrNormal)` when the context is canceled or can we just bail (what I'm doing here).

Stebalien:
See #4414

PR #4415, opened 2017-11-22
Delete some now unused commands lib code

whyrusleeping:
We left some dead code hanging around after the cmds lib refactor. But I think some changes might have been made to it post-extraction without us realizing what was going on.

cc @keks @ForrestWeston 

License: MIT


keks:
Stebalien's changes are in, so we only need to copy Forrest's. Doing that now.

PR #4418, opened 2017-11-22
Refactoring towards gateway extraction

lgierth:
This replaces the gateway's usage of the `path` package with the Core API. The usages of `importer`, `merkledag`, and `unixfs` packages will land in another PR in the next couple of days.

Note that gateway_handler.go is a huuuuge mess. For now the priority is to extract it into its own package (ipfs/go-ipfs-gateway) and then refactor and cleanup there.

### DIFF
--- a/core/coreapi/interface/interface.go
+++ b/core/coreapi/interface/interface.go
@@ -30,13 +30,13 @@ type Reader interface {
 type CoreAPI interface {
 	Unixfs() UnixfsAPI
 	ResolvePath(context.Context, Path) (Path, error)
-	ResolveNode(context.Context, Path) (Node, error)
+	ResolveNode(context.Context, Path) (Path, Node, error)
 }
 
 type UnixfsAPI interface {
 	Add(context.Context, io.Reader) (Path, error)
-	Cat(context.Context, Path) (Reader, error)
-	Ls(context.Context, Path) ([]*Link, error)
+	Cat(context.Context, Path) (Path, Reader, error)
+	Ls(context.Context, Path) (Path, []*Link, error)
 }
 
 // type ObjectAPI interface {
--- a/core/coreapi/interface/interface.go
+++ b/core/coreapi/interface/interface.go
@@ -11,8 +11,9 @@ import (
 
 type Path interface {
 	String() string
+	Components() []string
 	Cid() *cid.Cid
-	Root() *cid.Cid
+	RootCid() *cid.Cid
 	Resolved() bool
 }
 
@@ -61,4 +62,5 @@ type UnixfsAPI interface {
 // }
 
 var ErrIsDir = errors.New("object is a directory")
-var ErrOffline = errors.New("can't resolve, ipfs node is offline")
+var ErrOffline = errors.New("can't resolve, ipfs is offline")
+var ErrNotFound = errors.New("can't find requested node")
--- a/core/corehttp/gateway_handler.go
+++ b/core/corehttp/gateway_handler.go
@@ -182,7 +182,7 @@ func (i *gatewayHandler) getOrHeadHandler(ctx context.Context, w http.ResponseWr
 		return
 	}
 
-	dr, err := i.api.Unixfs().Cat(ctx, resolvedPath)
+	_, dr, err := i.api.Unixfs().Cat(ctx, resolvedPath)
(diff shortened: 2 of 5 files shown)


lgierth:
I'll fix the codeclimate errors

whyrusleeping:
@lgierth I think it would be really helpful to the whole coreapi effort to have [some examples](https://github.com/polydawn/refmt/blob/master/example_test.go) that show how its used in a couple different usecases. Using the go examples thing also allows to add tests and documentation at the same time :)

core/corehttp/gateway_handler.go
```diff
@@ -417,20 +416,23 @@ func (i *gatewayHandler) putHandler(w http.ResponseWriter, r *http.Request) {
...
 		webError(w, "putHandler: updating named entries not supported", errors.New("WritableGateway: ipns put not supported"), http.StatusBadRequest)
 		return
 	}
 
 	var newnode node.Node
 	if rsegs[len(rsegs)-1] == "QmUNLLsPACCz1vLxQVkXqqLX5R1X345qqfHbsf67hvA3Nn" {
+		// TODO(lgierth): this seems to mean that PUT /ipfs/QmUNLL/foo/bar
+		//                results in bar being an empty directory,
+		//                no matter what
 		newnode = ft.EmptyDirNode()
```
mildred: It seems that `QmUNLLsPACCz1vLxQVkXqqLX5R1X345qqfHbsf67hvA3Nn` is the hash for an empty directory.

core/corehttp/gateway_handler.go
```diff
@@ -443,13 +445,12 @@ func (i *gatewayHandler) putHandler(w http.ResponseWriter, r *http.Request) {
...
-		newPath = path.Join(rsegs[2:])
+		newPath = coreapi.JoinComponents(rsegs[2:])
 	}
 
 	var newcid *cid.Cid
-	rnode, err := core.Resolve(ctx, i.node.Namesys, i.node.Resolver, rootPath)
-	switch ev := err.(type) {
-	case path.ErrNoLink:
+	_, rnode, err := i.api.ResolveNode(ctx, rootPath)
+	if err == coreiface.ErrNotFound {
```
mildred: When changing \core.Resolve` to `api.ResolveNode`, there is an extra `api.node.DAG.Get` call hiding behind. Its result is discarded. Could it be possible to reuse it for later?

core/corehttp/gateway_handler.go
```diff
@@ -525,21 +525,16 @@ func (i *gatewayHandler) deleteHandler(w http.ResponseWriter, r *http.Request) {
...
-	if err != nil {
-		webError(w, "Could not split path", err, http.StatusInternalServerError)
-		return
-	}
+	components := p.Components()[2:]
 
 	tctx, cancel := context.WithTimeout(ctx, time.Minute)
 	defer cancel()
-	rootnd, err := i.node.Resolver.DAG.Get(tctx, c)
+	rootnd, err := i.node.Resolver.DAG.Get(tctx, p.RootCid())
```
mildred: Could it be possible to use the result of the previous `i.api.ResolveNode(ctx, rootPath)` changed up here which result get discarded?

lgierth:
Thanks @mildred, good points! I'll work on this a bit more.

core/coreapi/interface/interface.go
```diff
@@ -29,13 +30,13 @@ type Reader interface {
...
-	ResolveNode(context.Context, Path) (Node, error)
+	ResolveNode(context.Context, Path) (Path, Node, error)
 }
 
 type UnixfsAPI interface {
 	Add(context.Context, io.Reader) (Path, error)
-	Cat(context.Context, Path) (Reader, error)
-	Ls(context.Context, Path) ([]*Link, error)
+	Cat(context.Context, Path) (Path, Reader, error)
+	Ls(context.Context, Path) (Path, []*Link, error)
```
magik6k: As much as I'm not a fan of mutable state, I'd say that we may want to not return resolved paths and just modify the ones passed to the methods - while it makes sense when we implement the API on top of go-ipfs node, doing this in http-api (or in any other wrapper that might be created) imo doesn't, and it would likely require changing the http api to return those resolved paths.

PR #4423, opened 2017-11-24
Adding tests and documentation for package path.

IsaacDunlap:
Adds new test cases, improves test coverage, adds documentation and improves code maintainability.

License: MIT
Signed-off-by: Isaac Dunlap <isaacdunlap@gmail.com>

### DIFF
--- a/path/path.go
+++ b/path/path.go
@@ -5,26 +5,31 @@ import (
 	"path"
 	"strings"
 
-	cid "gx/ipfs/QmNp85zy9RLrQ5oQD4hPyS39ezrrXpcaa7R4Y9kxdWQLLQ/go-cid"
+	"gx/ipfs/QmNp85zy9RLrQ5oQD4hPyS39ezrrXpcaa7R4Y9kxdWQLLQ/go-cid"
 )
 
 // ErrBadPath is returned when a given path is incorrectly formatted
 var ErrBadPath = errors.New("invalid 'ipfs ref' path")
 
-// TODO: debate making this a private struct wrapped in a public interface
+// TODO: debate making Path a private struct wrapped in a public interface -
 // would allow us to control creation, and cache segments.
+
+// Path represents a path to a node. The path may have a prefix indicating
+// whether it is an IPFS or IPNS key.
 type Path string
 
-// FromString safely converts a string type to a Path type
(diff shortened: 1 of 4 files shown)


path/path.go
```diff
@@ -5,26 +5,31 @@ import (
...
 }
 
-// FromCid safely converts a cid.Cid type to a Path type
+// FromCid safely converts a cid.Cid type to a Path type.
 func FromCid(c *cid.Cid) Path {
 	return Path("/ipfs/" + c.String())
 }
 
+// Segments returns the individual segments of the Path (separated by
+// the '/' character.
```
whyrusleeping: missing a close paren

path/path.go
```diff
@@ -116,15 +127,19 @@ func ParseCidToPath(txt string) (Path, error) {
 	return FromCid(c), nil
 }
 
+// IsValid returns true if the path is valid, and false otherwise.
```
whyrusleeping: doesnt return true or false, it returns an error if its invalid, and nil otherwise

IsaacDunlap:
Hi,

I notice that some of the tests fail on mater but others don't. Which tests should I be worried about?

Also, how do I rerun these tests locally so I can get diags and stuff?

Isaac

Stebalien:
> I notice that some of the tests fail on mater but others don't. Which tests should I be worried about?

All the tests should pass (usually, there are a few tests that'll fail once in a while).


> Also, how do I rerun these tests locally so I can get diags and stuff?

`make test`

Or, if you're interested in a specific test:

1. If it's a go test, go to the package in question and run `go test`.
2. If it's a sharness test, just run the appropriate shell script in `test/sharness`.

Stebalien:
The jenkins failure may be spurious. However, the circleci failure looks real. Try running `test/sharness/t0053-dag.sh` locally.

```sh
> cd test/sharness
> make deps
> ./t0053-dag.sh
```

Kubuxu:
I would recommend running it by: `make test/sharness/t0053-dag.sh` in rood project directory. This way you are sure dependencies are up to date.

Other way would be to run `make t0053-dag.sh` in `test/sharness` which is equivalent to the former.

dignifiedquire:
@kubuxu did you happen to pick this up?

PR #4424, opened 2017-11-24
travis: update go version

dvrkps:


whyrusleeping:
This is being done as part of this PR: https://github.com/ipfs/go-ipfs/pull/4398

Its a little bit more tricky as we need to update the gx version we're using (since golang source parsing rules can change between versions)

dvrkps:
@whyrusleeping this stay open until #4398 merge ? may i close ?

whyrusleeping:
#4398 merged, thanks @dvrkps :)

PR #4428, opened 2017-11-26
Refine gateway and namesys logging

lgierth:
The signal/noise ratio on the ipfs.io gateways is incredibly bad nowadays - this is a step to improving that.

All the logs here are based on user input, and they're neither WARN nor INFO worthy, so I'm degrading them all to DEBUG. Also removed a few useless log lines.

I'm already testing this on scrappy, where the logs are now actually readable.

### DIFF
--- a/core/corehttp/gateway_handler.go
+++ b/core/corehttp/gateway_handler.go
@@ -119,7 +119,6 @@ func (i *gatewayHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		errmsg = errmsg + "bad request for " + r.URL.Path
 	}
 	fmt.Fprint(w, errmsg)
-	log.Error(errmsg) // TODO(cryptix): log errors until we have a better way to expose these (counter metrics maybe)
 }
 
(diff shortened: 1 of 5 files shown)


lgierth:
Filed #4429 for the CircleCI failure, and the first TravisCI failure is related to ongoing gateway issues.

core/corehttp/gateway_handler.go
```diff
@@ -510,7 +506,6 @@ func (i *gatewayHandler) putHandler(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 	default:
-		log.Warningf("putHandler: unhandled resolve error %T", ev)
```
Stebalien: If this is a user error, we shouldn't be returning a 502. If it is, we should log it.

lgierth: Mh yes indeed - I'll cover this in the upcoming gateway+coreapi PR.

whyrusleeping:
needs a rebase

lgierth:
@whyrusleeping rebased

PR #4433, opened 2017-11-28
AddWithContext wraps custom context

ForrestWeston:
AddWithContext should use the custom context when calling
NewAdder

License: MIT
Signed-off-by: ForrestWeston <forrest@protocol.ai>


lgierth:
Wow yeah - good catch :)

PR #4437, opened 2017-11-29
Fix/add abort panic

### LINKED ISSUE #4426: Panic on Ctrl-C to ipfs add
#### Description:

```
> cd $GOPATH/src/github.com/ipfs/go-ipfs
> ipfs add -r .
added QmNxgZ2uqh12m9Y2q7gRHQa8hCgXmwf3ZBu2mDz3L6ocUb go-ipfs/CHANGELOG.md
[snip]
 135.70 KB / ? [------------------------------------=---------------------------------------------------------------------------------] ^C
Error: context canceled
panic: context canceled

goroutine 66 [running]:
gx/ipfs/QmamUWYjFeYYzFDFPTvnmGkozJigsoDWUA4zoifTRFTnwK/go-ipfs-cmds.(*chanResponseEmitter).SetError(0xc4201dbda0, 0x12c3420, 0xc420046a10, 0x4)
	/home/user/workspace/gopath/src/gx/ipfs/QmamUWYjFeYYzFDFPTvnmGkozJigsoDWUA4zoifTRFTnwK/go-ipfs-cmds/chan.go:186 +0x121
main.callCommand.func1(0xc4201d0b90, 0x1cb55e0, 0xc4201dbda0, 0x1cb6a40, 0xc4202f0000)
	/home/user/protocol/ipfs/go-ipfs/cmd/ipfs/main.go:369 +0xa6
created by main.callCommand
	/home/user/protocol/ipfs/go-ipfs/cmd/ipfs/main.go:366 +0x9e8
```

keks:
fixes #4426.

### DIFF
--- a/cmd/ipfs/main.go
+++ b/cmd/ipfs/main.go
@@ -364,9 +364,12 @@ func callCommand(ctx context.Context, req cmds.Request, root *cmds.Command, cmd 
 		}
 
 		go func() {
-			err = cmds.Copy(re, res)
+			err := cmds.Copy(re, res)
 			if err != nil {
-				re.SetError(err, cmdkit.ErrNormal|cmdkit.ErrFatal)
+				err = re.Emit(cmdkit.Error{err.Error(), cmdkit.ErrNormal | cmdkit.ErrFatal})
+				if err != nil {
+					log.Error(err)
+				}
 			}
 		}()
 	} else {


whyrusleeping:
@keks could I get a description of what these changes do? and why they fix things?

keks:
Yes! sorry. I hoped the discussion in the liked issue would help.

When pressing Ctrl+C the ResponseEmitter closes and the `cmds.Copy` copying into the ResponseEmitter returns an error (go-ipfs/cmd/ipfs/main.go:369). However, we then call `SetError`, which tries to Emit the cmdkit.Error. Since the ResponseEmitter is already closed, that Emit fails. SetError was expected to be called in a cmd.Run function which is called by a cmd.Call which in turn catches all panics. That's why we decided to make SetError panic when Emit returns an error.
However, we called SetError in cmd/ipfs/main.go, and since the Emit failed, it panics. So now instead of calling SetError, I cann Emit directly and log the error instead of panicking. 

I don't think this is the perfect solution. But all of this will go away with cmds1.0, which I hope to get done this year or Q1 2018.

lgierth:
I think this LGTM -- do our best to do something useful with the error, otherwise just print it

PR #4440, opened 2017-11-29
Fix various IPNS issues

Stebalien:
Found while looking into: https://discuss.ipfs.io/t/ipfs-name-failing-to-resolve/1524

1. Fix the republish docs. We republish every 4 hours by default.
2. Republish after 1min on start. The 1min delay gives us time to start up everything else but ensures that the IPNS record becomes resolvable quickly if we've been offline for a while.
3. Republish every 5min when republishing fails. I doubt this will cost us *that* much bandwidth/CPU usage and will ensure that IPNS will continue working properly even if the node is intermittently offline.

### DIFF
--- a/namesys/republisher/repub.go
+++ b/namesys/republisher/repub.go
@@ -28,6 +28,7 @@ var log = logging.Logger("ipns-repub")
 
 var DefaultRebroadcastInterval = time.Hour * 4
 var InitialRebroadcastDelay = time.Minute * 1
+var FailureRetryInterval = time.Minute * 5
 
 const DefaultRecordLifetime = time.Hour * 24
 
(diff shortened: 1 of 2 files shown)


namesys/republisher/repub.go
```diff
@@ -55,15 +63,22 @@ func NewRepublisher(r routing.ValueStore, ds ds.Datastore, self ic.PrivKey, ks k
 }
 
 func (rp *Republisher) Run(proc goprocess.Process) {
-	tick := time.NewTicker(rp.Interval)
-	defer tick.Stop()
+	timer := time.NewTimer(InitialRebroadcastDelay)
+	defer timer.Stop()
+	if rp.Interval < InitialRebroadcastDelay {
```
vyzo: Doesn't this defeat the purpose of the initial delay?
I think we should simply tick with the InitialRebroadcastDelay unconditionally here.

Stebalien: The initial delay is usually *much shorter* than the interval. If the user asks us to, e.g., publish every 10s, IMO, we should do that regardless of the initial delay (at the moment, not doing that would break the tests).

Alternatively, we could make the initial delay configurable (but that's one more knob).

vyzo: Yeah that's reasonable -- no need to have another knob for this.

Issue #4431, opened 2017-11-27
Long hierarchical IPFS paths fail to resolve (HAMT, object links)


quole:
1. Say you have an image of a blue whale from Wikipedia:

\> `ipfs cat /ipfs/QmXoypizjW3WknFiJnKLwHCnL72vedxjQkDDP1mXWo6uco/I/m/Anim1754_-_Flickr_-_NOAA_Photo_Library.jpg > whale.jpg` #OK

2. But you want to resolve the address down to a plain hash (without the directory structure):

\> `ipfs resolve -r /ipfs/QmXoypizjW3WknFiJnKLwHCnL72vedxjQkDDP1mXWo6uco/I/m/Anim1754_-_Flickr_-_NOAA_Photo_Library.jpg` #FAILS

3. Then you get the following error:

> Error: no link named "I" under QmXoypizjW3WknFiJnKLwHCnL72vedxjQkDDP1mXWo6uco

----
# ipfs.io web interface 

Same issue, different interface

1. Working image:

https://ipfs.io/ipfs/QmXoypizjW3WknFiJnKLwHCnL72vedxjQkDDP1mXWo6uco/I/m/Anim1754_-_Flickr_-_NOAA_Photo_Library.jpg

2. Attempt to resolve:

https://ipfs.io/api/v0/resolve?arg=/ipfs/QmXoypizjW3WknFiJnKLwHCnL72vedxjQkDDP1mXWo6uco/I/m/Anim1754_-_Flickr_-_NOAA_Photo_Library.jpg&r=true

3. Error message:
```
{"Message":"no link named \"I\" under QmXoypizjW3WknFiJnKLwHCnL72vedxjQkDDP1mXWo6uco","Code":0}
```
----

#### Version information:
```
go-ipfs version: 0.4.12-be03fc4
Repo version: 6
System version: amd64/linux
Golang version: go1.9.2
```

kevina:
The problem is that `ipfs resolve` uses `core.Node.Resolver` which as far as I can tell uses `ResolveSingle`.  `ipfs cat` construct its own `path.Resolver` which uses `unixfs/io.ResolveUnixfsOnce`.  I also notice that `ipfs get` uses the same resolver that `ipfs resolve` uses and hence also fails.

kevina:
[...]

PR #4444, opened 2017-11-30
Use ResolveUnixfsOnce as default resolver.

kevina:
Might fix https://github.com/ipfs/go-ipfs/issues/4431

### DIFF
--- a/core/builder.go
+++ b/core/builder.go
@@ -229,7 +229,11 @@ func setupNode(ctx context.Context, n *IpfsNode, cfg *BuildCfg) error {
 		// this is kinda sketchy and could cause data loss
 		n.Pinning = pin.NewPinner(n.Repo.Datastore(), n.DAG, internalDag)
 	}
-	n.Resolver = path.NewBasicResolver(n.DAG)
+	//n.Resolver = path.NewBasicResolver(n.DAG)
+	n.Resolver = &path.Resolver{
+		DAG:         n.DAG,
+		ResolveOnce: uio.ResolveUnixfsOnce,
+	}
 
 	if cfg.Online {
 		if err := n.startLateOnlineServices(ctx); err != nil {


Stebalien:
This seems like a good opportunity to fix the IPLD/IPFS namespace problem. That is, we should be using `/ipld/QmId` for IPLD addresses and `/ipfs/QmId` for IPFS addresses (and choosing the resolver based on that namespace).

kevina:
@Stebalien actually it is [`core.Resolve`](https://github.com/ipfs/go-ipfs/blob/master/core/pathresolver.go#L24) job to handle the prefix.  That function takes a protocol specific resolver as a parameter.  What this is doing is changing the protocol specific resolver normally used from `ResolveSingle` to `ResolveUnixfsOnce`.

Stebalien:
This shouldn't *have* a default resolver. It should chose the resolver based on the protocol.

However, looking into this, it looks like all the resolver code is a mess so this is probably the best fix for now.

Stebalien:
However, I *am* worried that this may break things.

kevina:
We agree that the resolver code is a mess. :smiley:

whyrusleeping:
My worry here is that this will break things like that actually do really want the raw ipld resolver (like `ipfs dag`). Lets add some tests to make sure that you can use `ipfs dag get` to view the raw structure of a sharded directory.

kevina:
Unless I am missing something this will change the resolver not the output of `ipfs dag get` or `ipfs refs`, both commands should still show the raw dag.  What do you expect might break?

whyrusleeping:
@kevina 
```
ipfs dag get QmFooBar/shardedDirectory/AF/B2/C7filename
```

Thats how i expect paths to be addressed with ipfs dag, since its a command that operates on the base ipld layer (and contains no unixfs context)

kevina:
@whyrusleeping I didn't know that was possible, but will add some tests cases for it

kevina:
@whyrusleeping so are you saying you want command like `ipfs dag` to return different results then commands like `ipfs get` `ipfs cat` `ipfs resolve`, etc.

I am not sure I agree, because then we need to go through each command one by one and sort out which resolver to use for each.  We might also want a `--raw-ipld` for commands where wither resolver type could be used.  This in my option is just asking for pain.

I think we should have different syntax for resolving a raw dag vs a higher level path.  Maybe something like.  `QmFooBar/shardedDirectory/@AF/@B2/`.

@Stebalien, and others, thought.

As it stands right now `ipfs cat` and `ipfs ls` use the higher level resolver and most other commands use the lower level resolver.

Stebalien:
> I think we should have different syntax for resolving a raw dag vs a higher level path

The correct fix, as far as I know, is to differentiate between `/ipfs/...` and `/ipld/...`. There is no good fix without doing that.

However, for now, we should probably have `get`, `cat`, `files *` and *maybe* `object *` work with `ipfs` paths and `dag *` `block *` work with ipld paths. In the future, we'd expand this to be "unless otherwise specified (with a path prefix of `/ipld` or `/ipfs`).


Resolve, unfortunately, is tricky. In this case, we really do need to look at the path prefix. If we have an IPNS link that points to `/ipfs/QmId`, we need to use the IPFS resolver.

kevina:
@Stebalien I think the solution then is to make the default resolver also handle `ipld` prefixes.  I can take a stab at it but will likely need some guidance.  I would say the default resolve should default to `/ipfs` when a prefix is not given with an option to change the default to `/ipld` for low level commands such as `dag` `block` etc.

I will also see what I can do about cleaning up the code in general.

kevina:
@Stebalien I think that is a good plan.  My only concern is the performance cost of the extra allocation when prefixing plain CIDs with "/ipfs" only to have it immediately removed, but if it is not a frequent operation I doubt it will mater.

I will see what I can make work, I am thinking that the resolver should take in a map that maps prefixed to protocol specific resolvers.

Stebalien:
> My only concern is the performance cost of the extra allocation when prefixing plain CIDs with "/ipfs" only to have it immediately removed, but if it is not a frequent operation I doubt it will mater.

Doing this at the edge should help. Also, allocations... go-ipfs... *\*snicker\**.

> I will see what I can make work, I am thinking that the resolver should take in a map that maps prefixed to protocol specific resolvers.

Something like that. Personally, I was thinking of something more like a registry. There's a single `Resolver` attached to the `IpfsNode` where one can call `Register(namespace, resolver)` to register a resolver for that namespace. That keeps it modular and allows, e.g., plugins to register new namespaces.

kevina:
@Stebalien 

> Doing this at the edge should help. 

I'm not sure I am following, what exactly do you mean by "at the edge".

Stebalien:
Superseded by the CoreAPI refactor.

PR #4446, opened 2017-12-01
Implement "repo rm-root" command to unlike the files API root.

### LINKED ISSUE #3934: IPFS Files root not avialable fix in 'ipfs repo fsck'
#### Description:

I somehow managed to do this:
```
 ~  ipfs files rm -r /                                                                        ✘ 
Error: error loading filesroot from DAG: merkledag: not found
 ~  ipfs files stat /                                                                         ✘ 
Error: error loading filesroot from DAG: merkledag: not found
 ~  ipfs files rm /                                                                           ✘ 
Error: error loading filesroot from DAG: merkledag: not found
```

and this

```
 ipfs daemon       
Initializing daemon...
Adjusting current ulimit to 2048...
Successfully raised file descriptor limit to 2048.
^C
Received interrupt signal, shutting down...
(Hit ctrl-c again to force-shutdown the daemon.)
16:24:37.894 ERROR   cmd/ipfs: error from node construction: error loading filesroot from DAG: Failed to get block for QmP6wAYvriaFzQjWMb7CPJP6v8LvEuYRh32mEEn6b45WRK: context canceled daemon.go:337
Error: error loading filesroot from DAG: Failed to get block for QmP6wAYvriaFzQjWMb7CPJP6v8LvEuYRh32mEEn6b45WRK: context canceled
```

Having `ipfs repo fsck` check for this problem and fixing it under user confirmation (as it might lead to data loss as files root is best effort pinning would be nice).

kevina:
Closes #3934.

Note I used the new API because the old API seams to initialize an `IpfsNode` even if `cmds.Context.ConstructNode()` is not called.  (Setting `doesNotUseRepo` and/or `doesNotUseConfigAsInput` did not seam to prevent the node initialization.)

### DIFF
--- a/core/commands/repo.go
+++ b/core/commands/repo.go
@@ -11,6 +11,7 @@ import (
 
 	bstore "github.com/ipfs/go-ipfs/blocks/blockstore"
 	oldcmds "github.com/ipfs/go-ipfs/commands"
+	core "github.com/ipfs/go-ipfs/core"
 	e "github.com/ipfs/go-ipfs/core/commands/e"
 	corerepo "github.com/ipfs/go-ipfs/core/corerepo"
 	config "github.com/ipfs/go-ipfs/repo/config"
@@ -18,7 +19,9 @@ import (
 	lockfile "github.com/ipfs/go-ipfs/repo/fsrepo/lock"
 
 	cid "gx/ipfs/QmNp85zy9RLrQ5oQD4hPyS39ezrrXpcaa7R4Y9kxdWQLLQ/go-cid"
+	b58 "gx/ipfs/QmT8rehPR3F6bmwL6zjUN8XpiDBFFpMP2myPdC6ApsWfJf/go-base58"
 	cmdkit "gx/ipfs/QmUyfy4QSr3NXym4etEiRyxBLqqAeKHJuRdi8AACxg63fZ/go-ipfs-cmdkit"
+	ds "gx/ipfs/QmVSase1JP7cq9QkPT46oNwdp9pT6kBkG3oqS14y3QcZjG/go-datastore"
 	cmds "gx/ipfs/QmamUWYjFeYYzFDFPTvnmGkozJigsoDWUA4zoifTRFTnwK/go-ipfs-cmds"
 )
 
@@ -35,7 +38,8 @@ var RepoCmd = &cmds.Command{
 	},
 
 	Subcommands: map[string]*cmds.Command{
-		"stat": repoStatCmd,
+		"stat":    repoStatCmd,
+		"rm-root": repoRmRootCmd,
 	},
 	OldSubcommands: map[string]*oldcmds.Command{
 		"gc":      repoGcCmd,
@@ -264,7 +268,66 @@ daemons are running.
 	},
 	Type: MessageOutput{},
 	Marshalers: oldcmds.MarshalerMap{
-		oldcmds.Text: MessageTextMarshaler,
+		cmds.Text: MessageTextMarshaler,
+	},
+}
+
+var repoRmRootCmd = &cmds.Command{
+	Helptext: cmdkit.HelpText{
+		Tagline: "Unlink the root used by the files API.",
+		ShortDescription: `
+'ipfs repo rm-root' will unlink the root used by the files API ('ipfs
+files' commands) without trying to read the root itself.  The root and
+its children will then be removed by the garbage collector unless
+pinned.  This command can only run when no ipfs daemons are running.
+`,
+	},
+	Run: func(req cmds.Request, res cmds.ResponseEmitter) {
+		configRoot := req.InvocContext().ConfigRoot
+
+		// Can't use a full node as that interferes with the removal
+		// of the files root, so open the repo directly
+		repo, err := fsrepo.Open(configRoot)
+		if err != nil {
+			res.SetError(err, cmdkit.ErrNormal)
+			return
+		}
+
+		// Get the old root and display it to the user so that they can
+		// can do something to prevent from being garbage collected,
+		// such as pin it
+		dsk := core.FilesRootKey()
+		val, err := repo.Datastore().Get(dsk)
+		switch {
+		case err == ds.ErrNotFound || val == nil:
+			cmds.EmitOnce(res, &MessageOutput{"Files API root not found.\n"})
+		default:
+			c, err := cid.Cast(val.([]byte))
+			cidStr := ""
+			if err != nil {
+				cidStr = c.String()
+			} else {
+				cidStr = b58.Encode(val.([]byte))
+			}
+			err = repo.Datastore().Delete(dsk)
+			if err != nil {
+				res.SetError(fmt.Errorf("unable to remove API root: %s.  Root hash was %s", err.Error(), cidStr), cmdkit.ErrNormal)
+				return
+			}
+			cmds.EmitOnce(res, &MessageOutput{fmt.Sprintf("Unlinked files API root.  Root hash was %s.\n", cidStr)})
+		}
+		repo.Close()
+	},
+	Type: MessageOutput{},
+	Encoders: cmds.EncoderMap{
+		cmds.Text: cmds.MakeEncoder(func(req cmds.Request, w io.Writer, v interface{}) error {
+			msg, ok := v.(*MessageOutput)
+			if !ok {
+				return e.TypeErr(msg, v)
+			}
+			_, err := fmt.Fprintf(w, "%s", msg.Message)
+			return err
+		}),
 	},
 }
 
--- a/core/core.go
+++ b/core/core.go
@@ -716,8 +716,10 @@ func (n *IpfsNode) loadBootstrapPeers() ([]pstore.PeerInfo, error) {
 	return toPeerInfos(parsed), nil
 }
 
+func FilesRootKey() ds.Key { return ds.NewKey("/local/filesroot") }
+
 func (n *IpfsNode) loadFilesRoot() error {
-	dsk := ds.NewKey("/local/filesroot")
+	dsk := FilesRootKey()
 	pf := func(ctx context.Context, c *cid.Cid) error {
 		return n.Repo.Datastore().Put(dsk, c.Bytes())
 	}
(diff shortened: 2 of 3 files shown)


whyrusleeping:
cc @keks on the comment about commands above

core/commands/repo.go
```diff
@@ -264,7 +268,66 @@ daemons are running.
...
+}
+
+var repoRmRootCmd = &cmds.Command{
+	Helptext: cmdkit.HelpText{
+		Tagline: "Unlink the root used by the files API.",
+		ShortDescription: `
+'ipfs repo rm-root' will unlink the root used by the files API ('ipfs
+files' commands) without trying to read the root itself.  The root and
+its children will then be removed by the garbage collector unless
+pinned.  This command can only run when no ipfs daemons are running.
```
whyrusleeping: maybe `when no ipfs daemon is running`

core/commands/repo.go
```diff
@@ -264,7 +268,66 @@ daemons are running.
...
+		// can do something to prevent from being garbage collected,
+		// such as pin it
+		dsk := core.FilesRootKey()
+		val, err := repo.Datastore().Get(dsk)
+		switch {
+		case err == ds.ErrNotFound || val == nil:
+			cmds.EmitOnce(res, &MessageOutput{"Files API root not found.\n"})
+		default:
+			c, err := cid.Cast(val.([]byte))
+			cidStr := ""
```
whyrusleeping: i would use `var cidStr string` and declare it before calling cid.Cast

core/commands/repo.go
```diff
@@ -264,7 +268,66 @@ daemons are running.
...
+			if err != nil {
+				cidStr = c.String()
```
whyrusleeping: if the err isnt empty, we probably shouldnt call `.String()` on the nil cid it returned

whyrusleeping:
How does the daemon function when the files root is deleted? Does it just assume an empty root and start over?

Also, we should be very clear about why/when you would ever use this command in the helptext. Its basically only for extremely unfortunate circumstances.

core/commands/repo.go
kevina: Oops.  That was suppose to be if `err == nil`.

core/commands/repo.go
kevina: I just copied and pasted the text from `ipfs repo fsck`.  But will fix.

kevina:
> How does the daemon function when the files root is deleted? Does it just assume an empty root and start over?

Yes.  It just assumes an empty root and start over.

kevina:
> Also, we should be very clear about why/when you would ever use this command in the helptext. Its basically only for extremely unfortunate circumstances.

I am horrible at the wording for these things.  Suggestions welcome.

kevina:
> > Also, we should be very clear about why/when you would ever use this command in the helptext. Its basically only for extremely unfortunate circumstances.

I added some text to the help text.  Let me know if its what you are looking for.  Suggestions welcome.

Kubuxu:
This is good command to have a manual confirmation (`Y` from stdin stream) with flag override but it might be hard to do with how currently the API works.

whyrusleeping:
@Kubuxu @magik6k wanna review this one? 

It would probably be good to have manual confirmation as suggested, but maybe for the user to run it with a '--confirm' flag? Don't want to have to figure out how to do an interactive prompt as part of the changeset.

core/commands/repo.go
```diff
@@ -264,7 +268,74 @@ daemons are running.
...
+	},
+}
+
+var repoRmRootCmd = &cmds.Command{
+	Helptext: cmdkit.HelpText{
+		Tagline: "Unlink the root used by the files API.",
+		ShortDescription: `
+'ipfs repo rm-root' will unlink the root used by the files API ('ipfs
+files' commands) without trying to read the root itself.  The root and
+its children will then be removed by the garbage collector unless
```
magik6k: > [...] will then be removed by the garbage collector [...]

It might bu unclear that we don't actually run gc here

whyrusleeping:
Yeah, I think some safety features might be good. Perhaps:

```
> ipfs repo rm-root
Error: this is a potentially dangerous operation please pass --confirm to proceed
> ipfs repo rm-root --confirm
Error: the root of your mfs directory exists locally. Are you sure you want to unlink this? pass --remote-existing-root to continue
```

core/commands/repo.go
kevina: Okay reworded.

kevina:
@whyrusleeping I added the options as you suggested.

core/commands/repo.go
```diff
@@ -259,7 +263,103 @@ daemons are running.
 	},
 	Type: MessageOutput{},
 	Marshalers: oldcmds.MarshalerMap{
-		oldcmds.Text: MessageTextMarshaler,
+		cmds.Text: MessageTextMarshaler,
+	},
+}
+
+var repoRmRootCmd = &cmds.Command{
```
Stebalien: Doesn't this belong in the `ipfs files` command tree? Also, I'd just call it `ipfs files chroot` and allow the user to change the root arbitrarily.

However, I may just be missing some context here.

core/commands/repo.go
```diff
@@ -259,7 +263,103 @@ daemons are running.
...
+		ctx, ok := env.(*oldcmds.Context)
+		if !ok {
+			res.SetError(fmt.Errorf("expected env to be of type %T, got %T", ctx, env), cmdkit.ErrNormal)
+			return
+		}
+		configRoot := ctx.ConfigRoot
+
+		// Can't use a full node as that interferes with the removal
+		// of the files root, so open the repo directly
+		repo, err := fsrepo.Open(configRoot)
```
Stebalien: That's... unfortunate. I run my daemon as a separate user. I'm fine with this as a temporary limitation, but we should fix it eventually.

core/commands/repo.go
Stebalien: We could even make it: `ipfs files chroot --save /oldroot /newroot` to allow saving the old root inside the new root. Actually, we might just want to do that by default to make this command less safer.

kevina: > Doesn't this belong in the ipfs files command tree?

@Stebalien this is meant to be a low level tool to deal with the case when the root node some how gets deleted and is not available anywhere.  It is in `repo` because it a low level command. Like `repo fsck` or `repo verify`.

`ipfs files chroot` is an interesting idea, but more involved.

core/commands/repo.go
```diff
@@ -259,7 +263,103 @@ daemons are running.
...
+			have, _ = bs.Has(c)
+		} else {
+			cidStr = b58.Encode(val.([]byte))
+		}
+
+		if have && !removeLocalRoot {
+			res.SetError(fmt.Errorf("root %s exists locally. Are you sure you want to unlink this? Pass --remove-local-root to continue", cidStr), cmdkit.ErrNormal)
+			return
+		} else if !have && removeLocalRoot {
+			res.SetError(fmt.Errorf("root does not %s exists locally. Please remove --remove-local-root to continue", cidStr), cmdkit.ErrNormal)
```
Stebalien: The documentation says "even if" not "only if" so I'd expect this case to work.

core/core.go
```diff
@@ -743,8 +743,11 @@ func (n *IpfsNode) loadBootstrapPeers() ([]pstore.PeerInfo, error) {
 	return toPeerInfos(parsed), nil
 }
 
+// FilesRootKey returns the datastore key for the files root
+func FilesRootKey() ds.Key { return ds.NewKey("/local/filesroot") }
```
Stebalien: I'd just make this a variable. I.e., `var FilesRootKey ds.Key = ds.NewKey(...)`

core/commands/repo.go
kevina: I can fix the documentation.  The idea behind this is to prevent this flag from automatically being passed.  It is intentionally annoying.  if this will Create a problem I can remove this check.

Stebalien:
Rebased.

core/commands/repo.go
Stebalien: Reviving...

IMO, having to pass `--confirm` is annoying enough. If a user wants to automate this, that's their problem (and their choice).

I've switched this to act as documented unless someone objects.

core/commands/repo.go
```diff
@@ -262,6 +266,89 @@ daemons are running.
...
+		}
+		defer repo.Close()
+		bs := bstore.NewBlockstore(repo.Datastore())
+
+		// Get the old root and display it to the user so that they can
+		// can do something to prevent from being garbage collected,
+		// such as pin it
+		val, err := repo.Datastore().Get(core.FilesRootKey)
+		if err == ds.ErrNotFound || val == nil {
+			return cmds.EmitOnce(res, &MessageOutput{"Files API root not found.\n"})
```
magik6k: API sounds a bit weird here

```suggestion
			return cmds.EmitOnce(res, &MessageOutput{"FilesRoot not found.\n"})
```

core/commands/repo.go
```diff
@@ -262,6 +266,89 @@ daemons are running.
...
+			cidStr = b58.Encode(val)
+		}
+
+		if have && !removeExistingRoot {
+			return fmt.Errorf("root %s exists locally. Are you sure you want to unlink this? Pass --remove-existing-root to continue", cidStr)
+		}
+
+		err = repo.Datastore().Delete(core.FilesRootKey)
+		if err != nil {
+			return fmt.Errorf("unable to remove API root: %s.  Root hash was %s", err.Error(), cidStr)
```
magik6k: `API` Sounds a bit confusing here IMO

core/commands/repo.go
magik6k: We may want something like this for pin root too, as it can be useful in similar situations

Stebalien:
@magik6k I've renamed this to `ipfs rm-files-root` and improved the wording a bit.

magik6k:
Sharness seems to have failed on this: https://circleci.com/gh/ipfs/go-ipfs/14343?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

lidel:
Superseded by #8648

PR #4451, opened 2017-12-04
Pinning fixes

### LINKED ISSUE #4122: Pin with ton of hashes won't start
#### Description:

`ipfs pin add` with a ton of hashes as input won't start doing anything. It'll just sit there, even with `--progress` it'll never move beyond `Fetched/Processed 0 nodes.

This applies to both `cat hashes.txt | xargs ipfs pin add`, as well as `cat hashes.txt | ipfs pin add`.

The list in question has 1199 items: [pins.txt](https://github.com/ipfs/go-ipfs/files/1202763/pins.txt).

Stebalien:
* Better allocation patterns
* Fix a very small race condition between GC and pinning.
* Fix #4122 by resolving nodes as needed instead of up-front.
* Avoid creating many unnecessary contexts (also related to #4122).

### DIFF
--- a/assets/assets.go
+++ b/assets/assets.go
@@ -80,11 +80,6 @@ func addAssetList(nd *core.IpfsNode, l []string) (*cid.Cid, error) {
 		return nil, err
 	}
 
-	dcid, err := nd.DAG.Add(dir)
-	if err != nil {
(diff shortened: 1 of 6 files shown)


Kubuxu:
Hmm, I thought I sent out the CR (approve), looks like I did not.

PR #4458, opened 2017-12-05
bitswap nits

Stebalien:


magik6k:
Test fail on Jenkins looks unrelated:

```
ok 11 - 'ipfs daemon' succeeds

expecting success: 
    test_wait_for_file 20 100ms "$IPFS_PATH/api"
  
Error: timed out waiting for file: /go/src/github.com/ipfs/go-ipfs/test/sharness/trash directory.t0045-ls.sh/.ipfs/api
not ok 12 - api file shows up
#	
#	    test_wait_for_file 20 100ms "$IPFS_PATH/api"
#	
```

exchange/bitswap/bitswap.go
```diff
@@ -372,16 +372,6 @@ func (bs *Bitswap) ReceiveMessage(ctx context.Context, p peer.ID, incoming bsmsg
 		return
 	}
 
-	// quickly send out cancels, reduces chances of duplicate block receives
-	var keys []*cid.Cid
```
Stebalien: @whyrusleeping this code was dead. We're currently sending out cancels from within sessions. However, this *does* mean that the above comment is no longer correct (it'll likely take a bit longer to send out cancels).

Stebalien:
This should hopefully make gc a bit happier when bitswapping.

exchange/bitswap/decision/engine.go
```diff
@@ -336,12 +337,12 @@ func (e *Engine) numBytesReceivedFrom(p peer.ID) uint64 {
 // ledger lazily instantiates a ledger
 func (e *Engine) findOrCreate(p peer.ID) *ledger {
 	e.lock.Lock()
+	defer e.lock.Unlock()
```
whyrusleeping: I do want to note that this may cause a perf degradation

Issue #4435, opened 2017-11-29
ipfs name follow


lgierth:
cc @vyzo 

```
USAGE
  ipfs name follow <ipns-path> - Track changes to IPNS names.

SYNOPSIS
  ipfs name follow [--pin=true] [--expire=false] [--] <ipns-path>

ARGUMENTS

  <ipns-path> - IPNS path of the name that should be followed.

OPTIONS

  --pin           bool   - Always pin the name's most recent content. Default: true
  --allow-expired bool   - Allow lookup even if a name's TTL is reached. Default: true

DESCRIPTION

  The 'ipfs name follow' command can be used to track changes to an IPNS name.
  A followed name will be refreshed in the background in intervals, so that
  lookups through an 'ipfs' command or the HTTP API are always instant and up-to-date.

  When called with '--pin' (default), after refreshes the name's new content
  is pinned. This is equivalent to 'ipfs pin replace /ipfs/QmOld /ipfs/QmNew'.

  When called with '--allow-expired' (default), lookups of the name through
  the 'ipfs' command or the HTTP API always return the last known content,
  even if the TTL has been reached and the name has expired. If this is the
  case, a warning is printed along with whatever is the command's respones.

  The combination of '--pin' and '--allow-expired' allows for convenient
  offline use cases. For example, on a long flight I'd like to have IPFS docs
  with me, and I'd like them to be usable at /ipns/docs.ipfs.io as always.

  Open question: pubsub

  Open question: refresh after ttl/2 or whatever
  - /ipns/QmFoo:       ttl = ipns-ttl
  - /ipns/dnslink.net: ttl = min(dns-ttl, ipns-ttl)

  Open question: storage of followings
```

Stebalien:
Is this only active when the command is active? I see two different commands here: `ipfs pin add --follow` and `ipfs resolve --follow`.

# Pinning

```
ipfs pin add --follow [--allow-expired=true] [--lazy=false] /ipns/QmId
```

This command will create a new (permanent) pin that will "follow" this IPNS address while the daemon is active.

Note: `--lazy` exists to mirror `--prefetch` below (basically: make a non-strict pin that doesn't download everything up-front).

**Use-cases:** Mirroring, backup, offline-replication.

# Follow

```
ipfs resolve --follow [--prefetch] [--quiet] /ipns/...
```

[...]

PR #4462, opened 2017-12-07
[WIP] Implement ipfs name follow

vyzo:
Implements ipfs follow in a minimally non intrusive way; #4435 

TBD: testing

### DIFF
--- a/core/commands/follow.go
+++ b/core/commands/follow.go
@@ -1,0 +1,170 @@
+package commands
+
+import (
+	"errors"
+	"io"
+	"strings"
+
+	cmds "github.com/ipfs/go-ipfs/commands"
+	e "github.com/ipfs/go-ipfs/core/commands/e"
+
+	"gx/ipfs/QmUyfy4QSr3NXym4etEiRyxBLqqAeKHJuRdi8AACxg63fZ/go-ipfs-cmdkit"
+)
+
+type ipnsFollowResult struct {
+	OK bool
+}
+
+var IpnsFollowCmd = &cmds.Command{
+	Helptext: cmdkit.HelpText{
+		Tagline: "Follow IPNS names.",
+		ShortDescription: `
+Periodically resolve and optionally pin IPNS names in the background.
+`,
+	},
+	Subcommands: map[string]*cmds.Command{
+		"add":    ipnsFollowAddCmd,
+		"pin":    ipnsFollowPinCmd,
+		"list":   ipnsFollowListCmd,
+		"cancel": ipnsFollowCancelCmd,
+	},
+}
+
+var ipnsFollowAddCmd = &cmds.Command{
+	Helptext: cmdkit.HelpText{
+		Tagline: "Follow a name without pinning",
+		ShortDescription: `
+Follows an IPNS name by periodically resolving in the backround.
+`,
+	},
+	Run: func(req cmds.Request, res cmds.Response) {
+		n, err := req.InvocContext().GetNode()
+		if err != nil {
+			res.SetError(err, cmdkit.ErrNormal)
+			return
+		}
+
+		if n.Namecache == nil {
+			res.SetError(errors.New("IPNS Namecache is not available"), cmdkit.ErrClient)
+			return
+		}
+
+		n.Namecache.Follow(req.Arguments()[0], false)
+
+		res.SetOutput(&ipnsFollowResult{true})
+	},
+	Arguments: []cmdkit.Argument{
+		cmdkit.StringArg("name", true, false, "IPNS Name to follow."),
+	},
+	Type: ipnsFollowResult{},
+	Marshalers: cmds.MarshalerMap{
+		cmds.Text: marshalFollowResult,
+	},
+}
+
+var ipnsFollowPinCmd = &cmds.Command{
+	Helptext: cmdkit.HelpText{
+		Tagline: "Follows and pins a name",
+		ShortDescription: `
+Follows an IPNS name by periodically resolving and recursively
+pinning in the backround.
+`,
+	},
+	Run: func(req cmds.Request, res cmds.Response) {
+		n, err := req.InvocContext().GetNode()
+		if err != nil {
+			res.SetError(err, cmdkit.ErrNormal)
+			return
+		}
+
+		if n.Namecache == nil {
+			res.SetError(errors.New("IPNS Namecache is not available"), cmdkit.ErrClient)
+			return
+		}
+
+		n.Namecache.Follow(req.Arguments()[0], true)
+
+		res.SetOutput(&ipnsFollowResult{true})
+	},
+	Arguments: []cmdkit.Argument{
+		cmdkit.StringArg("name", true, false, "IPNS Name to follow."),
+	},
+	Type: ipnsFollowResult{},
+	Marshalers: cmds.MarshalerMap{
+		cmds.Text: marshalFollowResult,
+	},
+}
+
+var ipnsFollowListCmd = &cmds.Command{
+	Helptext: cmdkit.HelpText{
+		Tagline: "List names followed by the daemon",
+	},
+	Run: func(req cmds.Request, res cmds.Response) {
+		n, err := req.InvocContext().GetNode()
+		if err != nil {
+			res.SetError(err, cmdkit.ErrNormal)
+			return
+		}
+
+		if n.Namecache == nil {
+			res.SetError(errors.New("IPNS Namecache is not available"), cmdkit.ErrClient)
+			return
+		}
+
+		res.SetOutput(&stringList{n.Namecache.ListFollows()})
+	},
+	Type: stringList{},
+	Marshalers: cmds.MarshalerMap{
+		cmds.Text: stringListMarshaler,
+	},
+}
+
+var ipnsFollowCancelCmd = &cmds.Command{
+	Helptext: cmdkit.HelpText{
+		Tagline: "Cancels a follow",
+	},
+	Run: func(req cmds.Request, res cmds.Response) {
+		n, err := req.InvocContext().GetNode()
+		if err != nil {
+			res.SetError(err, cmdkit.ErrNormal)
+			return
+		}
+
+		if n.Namecache == nil {
+			res.SetError(errors.New("IPNS Namecache is not available"), cmdkit.ErrClient)
+			return
+		}
+
+		n.Namecache.Unfollow(req.Arguments()[0])
+
+		res.SetOutput(&ipnsFollowResult{true})
+	},
+	Arguments: []cmdkit.Argument{
+		cmdkit.StringArg("name", true, false, "Name follow to cancel."),
+	},
+	Type: ipnsFollowResult{},
+	Marshalers: cmds.MarshalerMap{
+		cmds.Text: marshalFollowResult,
+	},
+}
+
+func marshalFollowResult(res cmds.Response) (io.Reader, error) {
+	v, err := unwrapOutput(res.Output())
+	if err != nil {
+		return nil, err
+	}
+
+	output, ok := v.(*ipnsFollowResult)
+	if !ok {
+		return nil, e.TypeErr(output, v)
+	}
+
+	var state string
+	if output.OK {
+		state = "ok"
+	} else {
+		state = "wtf"
+	}
+
+	return strings.NewReader(state + "\n"), nil
+}
--- a/namecache/namecache.go
+++ b/namecache/namecache.go
@@ -1,0 +1,151 @@
+// Package namecache implements background following (resolution and pinning) of names
+package namecache
+
+import (
+	"context"
+	"strings"
+	"sync"
+	"time"
+
+	merkledag "github.com/ipfs/go-ipfs/merkledag"
+	namesys "github.com/ipfs/go-ipfs/namesys"
+	path "github.com/ipfs/go-ipfs/path"
+	pin "github.com/ipfs/go-ipfs/pin"
+	uio "github.com/ipfs/go-ipfs/unixfs/io"
+
+	logging "gx/ipfs/QmSpJByNKFX1sCsHBEp3R73FL4NF6FnQTEGyNAXHm2GS52/go-log"
+)
+
+const (
+	followInterval = 60 * time.Minute
+)
+
+var log = logging.Logger("namecache")
+
+// NameCache represents a following cache of names
+type NameCache interface {
+	Follow(name string, pinit bool)
+	Unfollow(name string)
+	ListFollows() []string
+}
+
+type nameCache struct {
+	nsys    namesys.NameSystem
+	pinning pin.Pinner
+	dag     merkledag.DAGService
+
+	ctx     context.Context
+	follows map[string]func()
+	mx      sync.Mutex
+}
+
+func NewNameCache(ctx context.Context, nsys namesys.NameSystem, pinning pin.Pinner, dag merkledag.DAGService) NameCache {
+	return &nameCache{
+		ctx:     ctx,
+		nsys:    nsys,
+		pinning: pinning,
+		dag:     dag,
+		follows: make(map[string]func()),
+	}
+}
+
+// Follow spawns a goroutine that periodically resolves a name
+// and (when pinit is true) pins it in the background
+func (nc *nameCache) Follow(name string, pinit bool) {
+	nc.mx.Lock()
+	defer nc.mx.Unlock()
+
+	if _, ok := nc.follows[name]; ok {
+		return
+	}
+
+	ctx, cancel := context.WithCancel(nc.ctx)
+	go nc.followName(ctx, name, pinit)
+	nc.follows[name] = cancel
+}
+
+// Unfollow cancels a follow
+func (nc *nameCache) Unfollow(name string) {
+	nc.mx.Lock()
+	defer nc.mx.Unlock()
+
+	cancel, ok := nc.follows[name]
+	if ok {
+		cancel()
+		delete(nc.follows, name)
+	}
+}
+
+// ListFollows returns a list of names currently being followed
+func (nc *nameCache) ListFollows() []string {
+	nc.mx.Lock()
+	defer nc.mx.Unlock()
+
+	follows := make([]string, 0)
+	for name, _ := range nc.follows {
+		follows = append(follows, name)
+	}
+
+	return follows
+}
+
+func (nc *nameCache) followName(ctx context.Context, name string, pinit bool) {
+	nc.resolveAndPin(ctx, name, pinit)
+
+	ticker := time.NewTicker(followInterval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ticker.C:
+			nc.resolveAndPin(ctx, name, pinit)
+
+		case <-ctx.Done():
+			return
+		}
+	}
+}
+
+func (nc *nameCache) resolveAndPin(ctx context.Context, name string, pinit bool) {
+	log.Debugf("resolving %s", name)
+
+	if !strings.HasPrefix(name, "/ipns/") {
+		name = "/ipns/" + name
+	}
+
+	p, err := nc.nsys.Resolve(ctx, name)
+	if err != nil {
+		log.Debugf("error resolving %s: %s", name, err.Error())
+		return
+	}
+
+	log.Debugf("resolved %s to %s", name, p)
+
+	if !pinit {
+		return
+	}
+
+	log.Debugf("pinning %s", p)
+
+	r := &path.Resolver{
+		DAG:         nc.dag,
+		ResolveOnce: uio.ResolveUnixfsOnce,
+	}
+
+	n, err := r.ResolvePath(ctx, p)
+	if err != nil {
+		log.Debugf("error resolving path %s to node: %s", p, err.Error())
+		return
+	}
+
+	err = nc.pinning.Pin(ctx, n, true)
+	if err != nil {
+		log.Debugf("error pinning path %s: %s", p, err.Error())
+		return
+	}
+
+	err = nc.pinning.Flush()
+	if err != nil {
+		log.Debugf("error flushing pin: %s", err.Error())
+	}
+}
--- a/core/commands/name.go
+++ b/core/commands/name.go
@@ -64,5 +64,6 @@ Resolve the value of a dnslink:
 		"publish": PublishCmd,
 		"resolve": IpnsCmd,
 		"pubsub":  IpnsPubsubCmd,
+		"follow":  IpnsFollowCmd,
 	},
 }
--- a/core/core.go
+++ b/core/core.go
@@ -31,6 +31,7 @@ import (
 	mount "github.com/ipfs/go-ipfs/fuse/mount"
 	merkledag "github.com/ipfs/go-ipfs/merkledag"
 	mfs "github.com/ipfs/go-ipfs/mfs"
+	namecache "github.com/ipfs/go-ipfs/namecache"
 	namesys "github.com/ipfs/go-ipfs/namesys"
 	ipnsrp "github.com/ipfs/go-ipfs/namesys/republisher"
 	p2p "github.com/ipfs/go-ipfs/p2p"
@@ -130,6 +131,7 @@ type IpfsNode struct {
 	Routing      routing.IpfsRouting // the routing system. recommend ipfs-dht
 	Exchange     exchange.Interface  // the block exchange + strategy (bitswap)
 	Namesys      namesys.NameSystem  // the name system, resolves paths to hashes
+	Namecache    namecache.NameCache // the name system follower cache
 	Ping         *ping.PingService
 	Reprovider   *rp.Reprovider // the value reprovider system
 	IpnsRepub    *ipnsrp.Republisher
@@ -457,6 +459,7 @@ func (n *IpfsNode) startOnlineServicesWithHost(ctx context.Context, host p2phost
 
 	// setup name system
 	n.Namesys = namesys.NewNameSystem(n.Routing, n.Repo.Datastore(), size)
+	n.Namecache = namecache.NewNameCache(ctx, n.Namesys, n.Pinning, n.DAG)
 
 	// setup ipns republishing
 	return n.setupIpnsRepublisher()


namecache/namecache.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+// Follow spawns a goroutine that periodically resolves a name
+// and (when pinit is true) pins it in the background
+func (nc *nameCache) Follow(name string, pinit bool) {
+	nc.mx.Lock()
+	defer nc.mx.Unlock()
+
+	if _, ok := nc.follows[name]; ok {
+		return
```
magik6k: Might want to return an error here

namecache/namecache.go
```diff
@@ -0,0 +1,151 @@
...
+	nc.follows[name] = cancel
+}
+
+// Unfollow cancels a follow
+func (nc *nameCache) Unfollow(name string) {
+	nc.mx.Lock()
+	defer nc.mx.Unlock()
+
+	cancel, ok := nc.follows[name]
+	if ok {
```
magik6k: Same as above, If user somehow makes a type there would be no indication

namecache/namecache.go
```diff
@@ -0,0 +1,151 @@
...
+
+	follows := make([]string, 0)
+	for name, _ := range nc.follows {
+		follows = append(follows, name)
+	}
+
+	return follows
+}
+
+func (nc *nameCache) followName(ctx context.Context, name string, pinit bool) {
```
magik6k: Having this as a single worker resolving one name each `followInterval/len(nc.follows)` will ensure no activity spikes when many names are followed.

namecache/namecache.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+func (nc *nameCache) resolveAndPin(ctx context.Context, name string, pinit bool) {
+	log.Debugf("resolving %s", name)
+
+	if !strings.HasPrefix(name, "/ipns/") {
+		name = "/ipns/" + name
+	}
+
+	p, err := nc.nsys.Resolve(ctx, name)
```
magik6k: Is there built in timeout in `nsys.Resolve`? If not we might want to wrap the ctx with some.

namecache/namecache.go
```diff
@@ -0,0 +1,151 @@
...
+		ResolveOnce: uio.ResolveUnixfsOnce,
+	}
+
+	n, err := r.ResolvePath(ctx, p)
+	if err != nil {
+		log.Debugf("error resolving path %s to node: %s", p, err.Error())
+		return
+	}
+
+	err = nc.pinning.Pin(ctx, n, true)
```
magik6k: Should call `pinning.Update` when the content wasn't pinned before previous resolve here.

magik6k: Oh and take the pinning lock somewhere around here - `defer n.Blockstore.PinLock().Unlock()`

core/commands/follow.go
```diff
@@ -0,0 +1,170 @@
...
+		if n.Namecache == nil {
+			res.SetError(errors.New("IPNS Namecache is not available"), cmdkit.ErrClient)
+			return
+		}
+
+		n.Namecache.Follow(req.Arguments()[0], false)
+
+		res.SetOutput(&ipnsFollowResult{true})
+	},
+	Arguments: []cmdkit.Argument{
```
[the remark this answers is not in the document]
vyzo: ok, will reorder.

core/commands/follow.go
```diff
@@ -0,0 +1,170 @@
...
+	e "github.com/ipfs/go-ipfs/core/commands/e"
+
+	"gx/ipfs/QmUyfy4QSr3NXym4etEiRyxBLqqAeKHJuRdi8AACxg63fZ/go-ipfs-cmdkit"
+)
+
+type ipnsFollowResult struct {
+	OK bool
+}
+
+var IpnsFollowCmd = &cmds.Command{
```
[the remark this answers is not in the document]
vyzo: duly noted, but there are some dependencies on common helpers in the package; plus i would rather leave command refactoring out of this PR.

namecache/namecache.go
```diff
@@ -0,0 +1,151 @@
...
+	namesys "github.com/ipfs/go-ipfs/namesys"
+	path "github.com/ipfs/go-ipfs/path"
+	pin "github.com/ipfs/go-ipfs/pin"
+	uio "github.com/ipfs/go-ipfs/unixfs/io"
+
+	logging "gx/ipfs/QmSpJByNKFX1sCsHBEp3R73FL4NF6FnQTEGyNAXHm2GS52/go-log"
+)
+
+const (
+	followInterval = 60 * time.Minute
```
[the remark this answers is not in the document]
vyzo: sure.

namecache/namecache.go
vyzo: ok

namecache/namecache.go
vyzo: will do.

namecache/namecache.go
vyzo: I can coalesce in a single goroutine, more resource economical.

namecache/namecache.go
vyzo: Probably not, i'll add a timeout here.

namecache/namecache.go
vyzo: Oh yeah, the lock -- good point. 
Should `Update` be called instead of `Pin`?

vyzo: Actually seems like a more complicated logic is needed:
- In first resolution, call `Pin`.
- In subsequent resolutions:
  - Do nothing if the name resolves to the same node
  - Call `Update` if the pin has changed

I also wonder whether we should unpin when we cancel a follow.

magik6k: > I also wonder whether we should unpin when we cancel a follow.

I'd say yes. Also call `Pin` instead of `Update` when the hash was pinned before the follow, so we don't unintentionally unpin it.

namecache/namecache.go
vyzo: I am not sure there is significant benefit by having a single goroutine, plus it's a lot more complicated to write if we want to ensure immediate initial resolution upon follow.
The pins will be serialized by the pin lock.

How strongly do you feel about using a single goroutine?

vyzo:
@magik6k updated per review.
The most significant change is in the pinning logic. I am still keeping one goroutine per follow, depending on how strong you feel about it.

TBD: make follow interval a configuration option.

namecache/namecache.go
vyzo: done.

namecache/namecache.go
magik6k: It was more of a suggestion, though it's still something we may want to do in the feature

core/commands/follow.go
```diff
@@ -0,0 +1,182 @@
...
+	"strings"
+
+	cmds "github.com/ipfs/go-ipfs/commands"
+	e "github.com/ipfs/go-ipfs/core/commands/e"
+
+	"gx/ipfs/QmUyfy4QSr3NXym4etEiRyxBLqqAeKHJuRdi8AACxg63fZ/go-ipfs-cmdkit"
+)
+
+type ipnsFollowResult struct {
+	OK bool
```
whyrusleeping: maybe instead of a boolean, we just have a `State` field thats a string, since you convert it into a string later anyways

namecache/namecache.go
whyrusleeping: I would invert this check. `if !ok { return err }`

namecache/namecache.go
```diff
@@ -0,0 +1,254 @@
...
+
+	return fmt.Errorf("Unknown name %s", name)
+}
+
+// ListFollows returns a list of names currently being followed
+func (nc *nameCache) ListFollows() []string {
+	nc.mx.Lock()
+	defer nc.mx.Unlock()
+
+	follows := make([]string, 0)
```
whyrusleeping: could do:
```go
follows := make([]string, 0, len(nc.follows))
``

**Comment by whyrusleeping** [comment]
@vyzo I think its probably worth modifying the namesys interface to return TTLs from resolve calls so we can use them here.

File: namecache/namecache.go (hunk 2 above)
Comment (vyzo) [reply namecache/namecache.go]: ok, prettier this way

**Comment by vyzo** [comment]
Ok, make sense.
I will simply add an extended resolve method to the namesys api instead of changing the old one (that is too much work).

File: core/commands/follow.go [hunk 11]
```diff
@@ -0,0 +1,182 @@
...
+	output, ok := v.(*ipnsFollowResult)
+	if !ok {
+		return nil, e.TypeErr(output, v)
+	}
+
+	var state string
+	if output.OK {
+		state = "ok"
+	} else {
+		state = "wtf"
```
ForrestWeston: nit - Is that actually what we want to return to the user?

core/commands/follow.go
Stebalien: We generally use `SetError` to indicate that the command has failed anyways. We should probably do something like what we do in `ipfs pin add`: `{"Follows": [CIDs]}`.

core/commands/follow.go
```diff
@@ -0,0 +1,182 @@
...
+var IpnsFollowCmd = &cmds.Command{
+	Helptext: cmdkit.HelpText{
+		Tagline: "Follow IPNS names.",
+		ShortDescription: `
+Periodically resolve and optionally pin IPNS names in the background.
+`,
+	},
+	Subcommands: map[string]*cmds.Command{
+		"add":    ipnsFollowAddCmd,
+		"pin":    ipnsFollowPinCmd,
```
Stebalien: I'd collapse pin and add to `add --pin` or something like that.

core/commands/follow.go
```diff
@@ -0,0 +1,182 @@
...
+			res.SetError(err, cmdkit.ErrNormal)
+			return
+		}
+
+		if n.Namecache == nil {
+			res.SetError(errors.New("IPNS Namecache is not available"), cmdkit.ErrClient)
+			return
+		}
+
+		err = n.Namecache.Follow(req.Arguments()[0], false)
```
Stebalien: Should allow multiple arguments.

namecache/namecache.go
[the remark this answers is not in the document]
Stebalien: We may also want to make this a flag `ipfs follow add --refresh-interval X`.

core/commands/follow.go
vyzo: Not sure it adds much though -- in reality it's just a void command that can error.

namecache/namecache.go
vyzo: ok

vyzo:
@Stebalien 
> This PR ignores all of the design discussion from #4435. Why?

Not true. It has been discussed in slack a bit, as the initial design would require very invasive changes to implement. So as a first pass, I opted for the simplest possible implementation in terms of invasiveness
and get there incrementally.

core/commands/follow.go
vyzo: Sure.

core/commands/follow.go
vyzo: Good point.

namecache/namecache.go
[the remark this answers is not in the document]
vyzo: The configurable flag is a good idea.

core/commands/follow.go
vyzo: Actually, I think it's best to just use an empty struct for void and just output ok.
As @Stebalien points out, we indicate errors with SetError anyway.

vyzo: Hrm, can't have it be just void. I will make it a string as why suggested.

core/commands/follow.go
vyzo: Done.

core/commands/follow.go
vyzo: Done.

namecache/namecache.go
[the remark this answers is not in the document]
vyzo: Added `--refresh-interval` parameter to `add`.

Stebalien:
>  It has been discussed in slack a bit

Policy: If it didn't happen on GitHub, it didn't happen. Those decisions need to go on the tracking issue.

> So as a first pass, I opted for the simplest possible implementation in terms of invasiveness
and get there incrementally.

Then we:

1. Need some form of experimental flag (default to off). Otherwise, users *will* come to rely on these commands. We're adding features to the API that we intend to take away.
2. Need issues to track what needs to be done, probably in that tracking issue (maybe with multiple linked issues).

namecache/namecache.go
```diff
@@ -0,0 +1,254 @@
...
+	delete(nc.follows, name)
+	return nil
+}
+
+// ListFollows returns a list of names currently being followed
+func (nc *nameCache) ListFollows() []string {
+	nc.mx.Lock()
+	defer nc.mx.Unlock()
+
+	follows := make([]string, len(nc.follows))[:0]
```
Stebalien: `make([]string, 0, len(nc.follows))`. Yes, technically this works but it's weird.

namecache/namecache.go
```diff
@@ -0,0 +1,254 @@
...
+		return err
+	}
+
+	return nc.pinning.Flush()
+}
+
+func (nc *nameCache) resolve(ctx context.Context, name string) (path.Path, error) {
+	log.Debugf("resolving %s", name)
+
+	if !strings.HasPrefix(name, "/ipns/") {
```
Stebalien: Can we do this at a higher layer?

Stebalien:
Otherwise..., LGTM. Thanks for implementing this (although I do worry that we're accumulating too many half-finished features).

vyzo:
> Need some form of experimental flag (default to off). Otherwise, users will come to rely on these commands. We're adding features to the API that we intend to take away.

Well, not necessarilly take away but potentially change behaviour. So agreed on making an experimental flag for the namecache.

namecache/namecache.go
vyzo: fixed

namecache/namecache.go
vyzo: Probably, but does it matter?

vyzo:
Added experimental flag `--enable-follow-experiment`.

vyzo:
@Stebalien 
A few things that are part of the design discussion in #4435 but not implemented by this PR:
- Offline resolution and resolution of expired names
- TTL-based refresh interval
- Persistence of follows

namecache/namecache.go
keks: Sanitize data as soon as it enters the program. I also advocate moving this into the command's `Run` function above.

core/commands/follow.go
```diff
@@ -0,0 +1,167 @@
+package commands
+
+import (
+	"errors"
+	"io"
+	"strings"
+	"time"
+
+	cmds "github.com/ipfs/go-ipfs/commands"
```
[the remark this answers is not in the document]
vyzo: it's a subcommand of name, which is an old style command and needs references to its children.

namecache/namecache.go
vyzo: The `Run` function is perhaps inappropriate, it doesn't cover direct uses of the API. 
If we are to enforce the `/ipns/` prefix then it should be done at the `Namecache` interface.

Still, it  will have to be done in 3 places instead of 1.

vyzo: I moved it to the `Namecache` interface methods (`Add` and `Cancel`).

lgierth:
This looks pretty good to me 👍 Thanks a lot :)

- Minor issue: should we call the `cancel` command `rm` instead?
  - It's technically correct in that it cancels an ongoing repeated operation, but the opposite of `add` that's used in other places is `rm`.
- UX issue: I ran `ipfs name follow add /ipns/ipfs.io/docs/examples` to try this out, then noticed that it wasn't fetching anything. I then ran the command again with `--pin`, and got an "already following" error.
  1. It would be fine for additional invocations of the same command to override the previous invocations.
  2. I'm thinking we should fetch the nodes regardless of `--pin`, i.e. do the equivalent of `ipfs refs -r`.

lgierth:
By the way, it's great that this works with fully-qualified paths a la `/ipns/ipfs.io/docs/examples`, and not just with plain PeerIDs :)

vyzo:
> Minor issue: should we call the cancel command rm instead?
> It's technically correct in that it cancels an ongoing repeated operation, but the opposite of add that's used in other places is rm.

If we call it rm in other places, then it makes sense to call it rm here too.

vyzo:
> It would be fine for additional invocations of the same command to override the previous invocations.

Agreed, just a little tricky to implement. Having to manually cancel and re-add is not great UX.

> I'm thinking we should fetch the nodes regardless of --pin, i.e. do the equivalent of ipfs refs -r

So you want to traverse the object for references and resolve those too?

core/commands/follow.go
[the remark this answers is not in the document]
keks: Fair enough. I think once the commands refactor has settled a bit we should update the parent command to use the new cmds lib instead though. For now I think it's okay because that process is still ongoing.

whyrusleeping:
I would like to push this forward. I think it will become more and more important in the near future.

@Stebalien @magik6k can you do another round of review, and  @vyzo can you rebase?

vyzo:
sure, will do tmrw morning. 
I also want to work a bit on the usability aspects @lgierth mentioned, i am in agreement that the UX should be nicer.

core/commands/follow.go
```diff
@@ -0,0 +1,167 @@
...
+		ShortDescription: `
+Follows an IPNS name by periodically resolving in the backround.
+`,
+	},
+	Arguments: []cmdkit.Argument{
+		cmdkit.StringArg("name", true, true, "IPNS Name to follow."),
+	},
+	Options: []cmdkit.Option{
+		cmdkit.BoolOption("pin", "Recursively pin the resolved pointer"),
+		cmdkit.StringOption("refresh-interval", "Follow refresh interval."),
```
magik6k: Would be nice to specify/document the default and say what format it's in (like in publish.go/lifetime)

core/commands/follow.go
```diff
@@ -0,0 +1,167 @@
...
+		pin, _, _ := req.Option("pin").Bool()
+
+		refrS, _, err := req.Option("refresh-interval").String()
+		if err != nil {
+			res.SetError(err, cmdkit.ErrNormal)
+			return
+		}
+
+		var refr time.Duration
+		if refrS == "" {
```
magik6k:
```
refreshInterval := nc.DefaultFollowInterval
if refreshIntervalOpt != "" {
  [...]
```

namecache/namecache.go
```diff
@@ -0,0 +1,258 @@
...
+
+	follows := make([]string, 0, len(nc.follows))
+	for name, _ := range nc.follows {
+		follows = append(follows, name)
+	}
+
+	return follows
+}
+
+func (nc *nameCache) followName(ctx context.Context, name string, pinit bool, followInterval time.Duration) {
```
magik6k: I'd use `dopin` instead of `pinit` as it's already used in few places.

namecache/namecache.go
```diff
@@ -0,0 +1,258 @@
...
+		return nil, err
+	}
+
+	if ncid.Equals(cid) {
+		return cid, nil
+	}
+
+	defer nc.bstore.PinLock().Unlock()
+
+	log.Debugf("Updating pin %s -> %s", cid.String(), ncid.String())
```
magik6k: Should also check `nc.pinning.IsPinned(ncid)` as is done above and only unpin the old cid if the new is present.

namecache/namecache.go
```diff
@@ -0,0 +1,258 @@
...
+}
+
+func (nc *nameCache) resolveAndUpdate(ctx context.Context, name string, cid *cid.Cid) (*cid.Cid, error) {
+
+	ptr, err := nc.resolve(ctx, name)
+	if err != nil {
+		return nil, err
+	}
+
+	ncid, err := pathToCid(ptr)
```
magik6k: s/cid/oldcid/; s/ncid/newcid/

magik6k:
Also, some way to prevent unpinning stuff pinned here using other pin interfaces.

diasdavid:
Wooot, it is coming to live! Exciting work @vyzo 👏🏽👏🏽

I've a small nitpick and a proposal to complete the API. The nitpick is that `ipfs name follow add` is a giant command and I believe it can be collapsed to just `ipfs name follow`. Here is my proposal for upgrading the name API.

- `ipfs name publish <IPFS path> --key=<optional key> --pubsub=<topic, default is the IPNS hash> --ttl=<TTL>`
- `ipfs name resolve <IPNS>`
- `ipfs name follow <IPNS> --interval=<interval in which to do IPNS resolutions> --pubsub=<true to subscribe on topic for fast updates, default to IPNS hash>`
- `ipfs name ls` -> return all the names published and resolved, with their latest publish/resolve times
- `ipfs name ls <IPNS>` -> return all the versions that have been resolved so far about this name

Giving a way for users to track all the names they have published and learned about will be useful.

vyzo:
@diasdavid this is a nice proposal for the general name api, but we still need some control commands for managing the follow state. 
I think we should consolidate the name api in a second pass (issue for design and follow up PR), while also moving all related commands into a directory as @magik6k suggests.

vyzo:
rebased and update per @magik6k review.

core/builder.go
```diff
@@ -210,7 +210,7 @@ func setupNode(ctx context.Context, n *IpfsNode, cfg *BuildCfg) error {
 
 	if cfg.Online {
 		do := setupDiscoveryOption(rcfg.Discovery)
-		if err := n.startOnlineServices(ctx, cfg.Routing, cfg.Host, do, cfg.getOpt("pubsub"), cfg.getOpt("ipnsps"), cfg.getOpt("mplex")); err != nil {
+		if err := n.startOnlineServices(ctx, cfg.Routing, cfg.Host, do, cfg.getOpt("pubsub"), cfg.getOpt("ipnsps"), cfg.getOpt("mplex"), cfg.getOpt("follow")); err != nil {
```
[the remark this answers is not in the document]
vyzo: done.

core/commands/follow.go
vyzo: added the default, not sure i want to repeat the same documentation for time formats here.

core/commands/follow.go
vyzo: done

namecache/namecache.go
vyzo: sure enough!

namecache/namecache.go
vyzo: I am not sure I understand why you want this check here. I do a pin update, no explicit unpin.

namecache/namecache.go
vyzo: done

vyzo:
Added a TODO checklist in the description.

diasdavid:
> I think we should consolidate the name api in a second pass (issue for design and follow up PR),

The backwards compatibility ghosts will haunt you 👻

vyzo:
> The backwards compatibility ghosts will haunt you 👻

it's an experimental feature, we can break it :)

eingenito:
@vyzo - what's the status of this PR?

vyzo:
dormant -- i haven't thought about it for a while...

namecache/namecache.go
```diff
@@ -0,0 +1,207 @@
...
+
+		toFetch, err := nc.dag.Get(ctx, change.After)
+		if err != nil {
+			return cid.Undef, err
+		}
+
+		// just iterate over all nodes
+		walker := ipld.NewWalker(ctx, ipld.NewNavigableIPLDNode(toFetch, nc.dag))
+		if err := walker.Iterate(func(node ipld.NavigableNode) error {
+			return nil
```
magik6k: It would be nice to add cids of nodes we iterate to some set and skip children we have already seen, but I'm not sure what's the best way to do this.

cc @schomatis

magik6k:
Still WIP, Pushed the prefetch variant

namecache/namecache.go
schomatis: > It would be nice to add cids of nodes we iterate to some set and skip children we have already seen, but I'm not sure what's the best way to do this.

1. Solution at the `Visitor` level.

The simplest way would be for the caller to keep that set and use it in the `Visitor`, in there we could check if the CID of the current child the `Walker` is pointing to has already been visited and if so, skip it, something like:

```Go

// For every child of the current `node`:
    childCid := node.Links()[walker.ActiveChildIndex()]
    if _, ok := visited[childCid]; ok {
        walker.NextChild();
        // Skip this child, already visited.
    } else {
        break; // out of the `for`, so when this call to `Visit` ends the `Walker`
        // will iterate into this child we haven't visited.
    }

// If we have skipped all of the children the `Walker` will just continue
// the iteration going up to the current `node`'s parent.
```

The downside of this approach is that `NavigableIPLDNode` still prefetches many children in advance, so if we didn't skip all of the children, if we request at least one, `NavigableIPLDNode` will also start prefetching many more after that one (that we may end up actually skipping).

2. Solution at the `NavigableIPLDNode` level.

In that case `NavigableIPLDNode` should be extended to handle this scenario (which would actually be a very nice feature). (Note here that `Walker` has no idea what an IPLD node is, it only operates on the very abstract notion of a DAG of any kind of node that implements the `NavigableNode` interface; the motivation behind this decoupling was to allow the possibility of using the `Walker` for other types of DAG like the dependency graphs that Gx iterates over, although at the moment IPLD is the only type of node that the `Walker` knows).

The way I think this could be implemented (but this is just an untested idea) would be to modify the [`preload`](https://github.com/ipfs/go-ipld-format/blob/ba07363e1f1f53c7fe828012a80e0ed7ae974d1a/navipld.go#L89-L96) function to, instead of prefetching a contiguous batch of `preloadSize` nodes through `GetNodes`, iterate the `childCIDs` one at a time and skip the ones already visited, calling each prefetch separately with `GetNode` (singular).

How should the map of visited nodes be handled in that case? Every `NavigableIPLDNode` is independent of its parent and children, there's no central authority, the way they communicate with each other is through the `context.Context` that was used in `NewWalker`, that context will be passed to every `FetchChild` issued by the Walker when going down the DAG. (Maybe this is too coupled and the `NavigableIPLDNode` should be extended to have its own context now, that gets passed from parent to child.) So the map could be stored there, and there could be a check that if there's no map of visited nodes, then just prefetch everything as `preload` does now. At the moment the `Walker` doesn't visit the DAG in parallel so this map would be safe (but it should be documented that in future refactorings it may need to be protected from multiple accesses).

/cc @hsanjuan who is interested in something very similar, so maybe you two could join forces here.

schomatis: (@hsanjuan in your case, inside the `Visitor` you could artificially add the CIDs of the nodes you want to skip entirely in this map, which would result in the same effect.)

remon-nashid:
This feature could make IPNS more useful in so many more uses case. I wonder if the community is still interested in this implementation, or maybe there is an alternative way to pin IPNS.

PR #4466, opened 2017-12-07
core: get cmd outPath removes trailing slash from args - fixes #3729

### LINKED ISSUE #3729: ipfs get /ipns/example.com fails with unexpected error
Version: 0.4.6-rc1

Reproduction:
```
ipfs get /ipns/multiformats.io/                                 
Saving file(s) to .
 1024 B / 412.86 KB [>-----------------------------------------------------------------]   0.24% 0s
Error: open .: is a directory
```

adamliesko:
Commit a09d97465270ef441791b65bca98d37c25eb0564 was a poor fix, which only
modified the path printed to the StdOut. The arguments of command are
passed to the request's reader, which is turned into a tar Extractor.
The headers of that the tar Extractor were incorrectly populating the
write path, missing the root directory, causing files nested in root
folder to be written into the root output path.

fix visible at: https://asciinema.org/a/mDili4YFnOacgt3dbjHuXTzcT

We are making the assumption, that get Command will have always only 1 argument and it's safe to remove the trailing fix of all the paths. Happy to hear feedback on this from maintainers.

### DIFF
--- a/core/commands/get.go
+++ b/core/commands/get.go
@@ -49,6 +49,15 @@ may also specify the level of compression by specifying '-l=<1-9>'.
 		cmdkit.IntOption("compression-level", "l", "The level of compression (1-9).").WithDefault(-1),
 	},
 	PreRun: func(req cmds.Request) error {
+		if len(req.Arguments()) == 0 {
+			return errors.New("not enough arguments provided")
+		}
+
+		// clean the path argument by removing trailing slash
+		pathArg := req.StringArguments()[0]
+		trimmedPath := strings.TrimSuffix(pathArg, "/")
+		req.SetArguments([]string{trimmedPath})
+
 		_, err := getCompressOptions(req)
 		return err
 	},
@@ -119,8 +128,6 @@ may also specify the level of compression by specifying '-l=<1-9>'.
 					return
 				}
 
-				outPath := getOutPath(req)
-
 				cmplvl, err := getCompressOptions(req)
 				if err != nil {
 					re.SetError(err, cmdkit.ErrNormal)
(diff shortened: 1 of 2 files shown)


whyrusleeping:
Hey @adamliesko thanks for the fix :)

Mind addressing the issues that codeclimate is complaining about?

adamliesko:
@whyrusleeping I refactored and split the tests, but fully adhering to those warnings, it would mean that table driven tests are dead :) - artificial cuts in the size of functions would result in less readability. The other code duplication/similarity is not a real duplication. Do we have some option to ignore it?

Are these coming from CircleCI ok or is it a regression ```Error: verify complete, some blocks were corrupt``` ?

whyrusleeping:
@adamliesko oh weird, i've never seen those sorts of complaints from codeclimate before. cc @Kubuxu

whyrusleeping:
@adamliesko re: the test failures, that looks like a regression. Things have been pretty green on master lately.

Stebalien:
> We are making the assumption, that get Command will have always only 1 argument and it's safe to remove the trailing fix of all the paths. Happy to hear feedback on this from maintainers.

It *should* accept multiple arguments (according to the help, at least). However, that's a different issue that can be addressed separately.

adamliesko:
@whyrusleeping @Stebalien - I see this is a bit too hacky and it probably shows up with the test failures. Maybe making a change on the level of tar extractor or when constructing the Reader inside the request is a better place. I will try to look into it further next week.

Kubuxu:
@whyrusleeping when codeclimate is complaining like that you can go to their page end either mark issues as "won't fix" or approve whole change set.

lidel:
Original issue seems to be fixed in latest release: https://github.com/ipfs/go-ipfs/issues/3729#issuecomment-589229387

PR #4469, opened 2017-12-08
Always set IPFS_PATH in test-lib.sh

### LINKED ISSUE #4467: t0025-datastores.sh uses default ipfs path
I am getting this (I have `IPFS_PATH` set to `/tmp/ipfs`)

```
expecting success: 
  BITS="1024" &&
  ipfs init --bits="$BITS" --profile=badgerds

Error: file "/tmp/ipfs/repo.lock" already locked
not ok 1 - 'ipfs init --profile=badgerds' succeeds
```

```
go-ipfs version: 0.4.14-dev-a4f9333
```

kevina:
Before it was only set when "test_init_ipfs" is called but that is not always used, for example in t0025-datastores.sh.

Fixes #4467.

### DIFF
--- a/test/sharness/lib/test-lib.sh
+++ b/test/sharness/lib/test-lib.sh
@@ -40,6 +40,10 @@ SHARNESS_LIB="lib/sharness/sharness.sh"
 
 # Please put go-ipfs specific shell functions below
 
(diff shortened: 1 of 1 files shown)


Kubuxu:
SGTM, but I haven't looked why Jenkins is failing.

PR #4471, opened 2017-12-08
RFC: coreapi.Dag

magik6k:
TODO:
- [ ] Review
- [x] Docs
- [x] Tests

### DIFF
--- a/core/coreapi/dag.go
+++ b/core/coreapi/dag.go
@@ -1,0 +1,77 @@
+package coreapi
+
+import (
+	"context"
+	"fmt"
+	"io"
+
+	gopath "path"
+
+	coreiface "github.com/ipfs/go-ipfs/core/coreapi/interface"
+	coredag "github.com/ipfs/go-ipfs/core/coredag"
+
+	cid "gx/ipfs/QmNp85zy9RLrQ5oQD4hPyS39ezrrXpcaa7R4Y9kxdWQLLQ/go-cid"
+	mh "gx/ipfs/QmU9a9NV9RdPNwZQDYd5uKsm6N6LJLSvLbywDDYFbaaC6P/go-multihash"
+)
+
+type DagAPI CoreAPI
+
+func (api *DagAPI) Put(ctx context.Context, src io.Reader, inputEnc string, format *cid.Prefix) ([]coreiface.Node, error) {
+	if format == nil {
+		format = &cid.Prefix{
+			Version:  1,
+			Codec:    cid.DagCBOR,
+			MhType:   mh.SHA2_256,
+			MhLength: mh.DefaultLengths[mh.SHA2_256],
+		}
+	}
+
+	codec, ok := cid.CodecToStr[format.Codec]
+	if !ok {
+		return nil, fmt.Errorf("invalid codec %d", format.Codec)
+	}
+
+	nds, err := coredag.ParseInputs(inputEnc, codec, src, format.MhType, format.MhLength)
+	if err != nil {
+		return nil, err
+	}
+	if len(nds) == 0 {
+		return nil, fmt.Errorf("no node returned from ParseInputs")
+	}
+
+	out := make([]coreiface.Node, len(nds))
+	for n, nd := range nds {
+		_, err := api.node.DAG.Add(nd)
+		if err != nil {
+			return nil, err
+		}
+		out[n] = nd
+	}
+
+	return out, nil
+}
+
+func (api *DagAPI) Get(ctx context.Context, path coreiface.Path) (coreiface.Node, error) {
+	return api.core().ResolveNode(ctx, path)
+}
+
+func (api *DagAPI) Tree(ctx context.Context, p coreiface.Path, depth int) ([]coreiface.Path, error) {
+	n, err := api.Get(ctx, p)
+	if err != nil {
+		return nil, err
+	}
+	paths := n.Tree("", depth)
+	out := make([]coreiface.Path, len(paths))
+	for n, p2 := range paths {
+		out[n], err = ParsePath(gopath.Join(p.String(), p2))
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return out, nil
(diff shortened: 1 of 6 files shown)


core/coreapi/dag.go
```diff
@@ -0,0 +1,61 @@
...
+	coreiface "github.com/ipfs/go-ipfs/core/coreapi/interface"
+	coredag "github.com/ipfs/go-ipfs/core/coredag"
+
+	cid "gx/ipfs/QmNp85zy9RLrQ5oQD4hPyS39ezrrXpcaa7R4Y9kxdWQLLQ/go-cid"
+)
+
+type DagAPI CoreAPI
+
+func (api *DagAPI) Add(ctx context.Context, src io.Reader, inputEnc string, format cid.Prefix) ([]coreiface.Node, error) {
+	nds, err := coredag.ParseInputs(inputEnc, cid.CodecToStr[format.Codec], src, format.MhType, format.MhLength)
```
whyrusleeping: we should probably validate format.Codec, otherwise we would return bad data from the `cid.CodeToStr` map

core/coreapi/interface/interface.go
```diff
@@ -38,6 +40,12 @@ type UnixfsAPI interface {
 	Ls(context.Context, Path) ([]*Link, error)
 }
 
+type DagAPI interface {
+	Add(ctx context.Context, src io.Reader, inputEnc string, format cid.Prefix) ([]Node, error)
```
whyrusleeping: Maybe `Put` instead of `Add`? 

I also think its important for people to be able to call these methods without having to import other packages. So making the `format` argument a pointer, and having some sort of defaults for when the user passes `nil` is something we should think about.

core/coreapi/dag.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	return out, nil
+}
+
+func (api *DagAPI) Get(ctx context.Context, path coreiface.Path) (coreiface.Node, error) {
+	return api.core().ResolveNode(ctx, path)
+}
+
+func (api *DagAPI) Tree(ctx context.Context, p coreiface.Path, depth int) ([]coreiface.Path, error) {
```
whyrusleeping: this makes me realize we should have an `ipfs dag tree` command

core/coreapi/dag_test.go
```diff
@@ -0,0 +1,89 @@
...
+)
+
+func TestPut(t *testing.T) {
+	ctx := context.Background()
+	_, api, err := makeAPI(ctx)
+	if err != nil {
+		t.Error(err)
+	}
+
+	_, err = api.Dag().Put(ctx, strings.NewReader(`"Hello"`), "json", nil)
```
whyrusleeping: I would check the Cid of the created object. Just one more additional protection against us breaking things in the future.

core/coreapi/interface/interface.go
```diff
@@ -28,8 +28,10 @@ type Reader interface {
 
 type CoreAPI interface {
 	Unixfs() UnixfsAPI
+	Dag() DagAPI
+
 	ResolvePath(context.Context, Path) (Path, error)
-	ResolveNode(context.Context, Path) (Node, error)
+	ResolveNode(context.Context, Path) (Node, error) //TODO: should this get dropped in favor of DagAPI.Get?
```
whyrusleeping: I think this is fine to keep as a helper alias

magik6k:
This is the first PR in the coreapi series that needs to get merged, mostly due to test utils being done here.

@whyrusleeping do you see anything more TODO here?

core/coreapi/interface/interface.go
```diff
@@ -55,6 +58,37 @@ type UnixfsAPI interface {
 	Ls(context.Context, Path) ([]*Link, error)
 }
 
+// DagAPI specifies the interface to IPLD
+type DagAPI interface {
+	// Put inserts data using specified format and input encoding.
+	// If format is not specified (nil), default dag-cbor/sha256 is used
```
lgierth: `Unless used with WithCodec or WithHash, the defaults "dag-cbor" and "sha256" are used.`

core/coreapi/interface/interface.go
```diff
@@ -55,6 +58,37 @@ type UnixfsAPI interface {
...
+	Put(ctx context.Context, src io.Reader, opts ...options.DagPutOption) ([]Node, error)
```
lgierth: Why is it desirable to have `Put()` return a slide of `Node`s?

whyrusleeping: I guess it could return a slice of cids instead. But we need to know what was added.

lgierth: I mean why is it desirable that it's a slice -- isn't there a root node for the added dag?

whyrusleeping: hrm... I think all of our current usecases would be fine with just the 'root' cid. Can you confirm that @magik6k ?

magik6k: It will be fine, yes. I decided to return a slice as it is what was returned by coredag.ParseInputs, I think the reasoning behind it was that the reader can provide more than one node, though it's not implemented now.

lgierth: Mh okay but what's the scenario that makes us want a slice returned here, versus just calling `Tree()` with the `Node.Cid()` returned from `Put()`? We'll keep all those nodes around in memory to return them from a function whose purpose isn't to iterate a graph (that's what `Tree()` is for).

I feel like we can rather be [consistent with js-ipfs](https://github.com/ipfs/interface-ipfs-core/blob/master/SPEC/DAG.md#dagput) here.

magik6k:
Made `Dag.Put` return a single node, this makes more sense than a slice. When we add a batching interface here in the future, having this method return a slice would be redundant (plus this wasn't really a batching interface..).

whyrusleeping:
@magik6k any idea why the tests are all failing?

PR #4477, opened 2017-12-10
RFC: coreapi.Name,Key

magik6k:
TODO:
- [x] Review
- [x] Docs
- [x] Tests

Note that most of the implementation code comes form respective commands, if this get's merged, I'll rewrite them to use this API.

### DIFF
--- a/core/coreapi/coreapi.go
+++ b/core/coreapi/coreapi.go
@@ -30,17 +30,11 @@ func (api *CoreAPI) Dag() coreiface.DagAPI {
 }
 
 func (api *CoreAPI) Name() coreiface.NameAPI {
-	return &NameAPI{
-		api,
-		nil,
-	}
+	return &NameAPI{api, nil}
 }
 
 func (api *CoreAPI) Key() coreiface.KeyAPI {
-	return &KeyAPI{
-		api,
-		nil,
-	}
+	return &KeyAPI{api, nil}
 }
 
 func (api *CoreAPI) ResolveNode(ctx context.Context, p coreiface.Path) (coreiface.Node, error) {
--- a/core/coreapi/coreapi.go
+++ b/core/coreapi/coreapi.go
@@ -33,6 +33,10 @@ func (api *CoreAPI) Name() coreiface.NameAPI {
 	return (*NameAPI)(api)
 }
 
+func (api *CoreAPI) Key() coreiface.KeyAPI {
+	return (*KeyAPI)(api)
+}
+
 func (api *CoreAPI) ResolveNode(ctx context.Context, p coreiface.Path) (coreiface.Node, error) {
 	p, err := api.ResolvePath(ctx, p)
 	if err != nil {
--- a/core/coreapi/coreapi.go
+++ b/core/coreapi/coreapi.go
@@ -29,6 +29,10 @@ func (api *CoreAPI) Dag() coreiface.DagAPI {
 	return &DagAPI{api, nil}
 }
 
+func (api *CoreAPI) Name() coreiface.NameAPI {
+	return (*NameAPI)(api)
+}
+
 func (api *CoreAPI) ResolveNode(ctx context.Context, p coreiface.Path) (coreiface.Node, error) {
 	p, err := api.ResolvePath(ctx, p)
 	if err != nil {
--- a/core/coreapi/coreapi.go
+++ b/core/coreapi/coreapi.go
@@ -30,11 +30,17 @@ func (api *CoreAPI) Dag() coreiface.DagAPI {
 }
 
 func (api *CoreAPI) Name() coreiface.NameAPI {
-	return (*NameAPI)(api)
+	return &NameAPI{
+		api,
+		nil,
+	}
 }
 
 func (api *CoreAPI) Key() coreiface.KeyAPI {
-	return (*KeyAPI)(api)
+	return &KeyAPI{
+		api,
+		nil,
+	}
 }
 
 func (api *CoreAPI) ResolveNode(ctx context.Context, p coreiface.Path) (coreiface.Node, error) {
--- a/core/coreapi/interface/interface.go
+++ b/core/coreapi/interface/interface.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"errors"
 	"io"
+	"time"
 
 	options "github.com/ipfs/go-ipfs/core/coreapi/interface/options"
 
@@ -27,6 +28,11 @@ type Path interface {
 type Node ipld.Node
 type Link ipld.Link
 
+type IpnsEntry struct {
+	Name  string
+	Value Path
+}
+
 type Reader interface {
 	io.ReadSeeker
 	io.Closer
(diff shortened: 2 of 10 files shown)


whyrusleeping:
@magik6k since these are going to be user facing APIs, lets go ahead and make this all godoc compliant

core/coreapi/coreapi.go
```diff
@@ -25,6 +25,20 @@ func (api *CoreAPI) Unixfs() coreiface.UnixfsAPI {
 	return (*UnixfsAPI)(api)
 }
 
+func (api *CoreAPI) Name() coreiface.NameAPI {
+	return &NameAPI{
+		api,
+		nil,
+	}
+}
```
lgierth: Is this multiline format what gofmt emitted?

core/coreapi/interface/interface.go
```diff
@@ -25,6 +28,11 @@ type Path interface {
 type Node ipld.Node
 type Link ipld.Link
 
+type IpnsEntry struct {
+	Name  string
+	Value Path
+}
```
lgierth: I wonder if we should make `IpnsEntry` an interface, in the spirit of future extensibility. Opinions on that?

core/coreapi/interface/interface.go
```diff
@@ -55,6 +65,76 @@ type UnixfsAPI interface {
...
+	Publish(ctx context.Context, path Path, opts ...options.NamePublishOption) (*IpnsEntry, error)
+
+	// WithValidTime is an option for Publish which specifies for how long the
+	// entry will remain valid. Default value is 24h
+	WithValidTime(validTime time.Duration) options.NamePublishOption
+
+	// WithKey is an option for Publish which specifies the key to use for
+	// publishing. Default value is "self" which is the node's own PeerID.
+	//
+	// You can use .Key API to list and generate more names and their respective keys.
```
lgierth: `You can use KeyAPI`

core/coreapi/key.go
```diff
@@ -0,0 +1,178 @@
...
+		}
+
+		out[key] = pid.Pretty()
+	}
+	return out, nil
+}
+
+func (api *KeyAPI) Rename(ctx context.Context, oldName string, newName string, opts ...caopts.KeyRenameOption) (string, bool, error) {
+	options, err := caopts.KeyRenameOptions(opts...)
+	if newName == "self" {
```
lgierth: This looks like it should be `err != nil`, typo?

core/coreapi/name.go
```diff
@@ -0,0 +1,161 @@
...
+		}
+	}
+
+	if n.Mounts.Ipns != nil && n.Mounts.Ipns.IsActive() {
+		return nil, errors.New("cannot manually publish while IPNS is mounted")
+	}
+
+	if n.Identity == "" {
+		return nil, errors.New("identity not loaded")
+	}
```
lgierth: We don't use `n.Identity`, I think this check can be removed

core/coreapi/interface/options/name.go
```diff
@@ -0,0 +1,89 @@
...
+		return nil
+	}
+}
+
+func (api *NameOptions) WithKey(key string) NamePublishOption {
+	return func(settings *NamePublishSettings) error {
+		settings.Key = key
+		return nil
+	}
+}
```
lgierth: I noticed `key` can be both a key name and peerID (see `keylookup()`) -- cool! Let's make sure to document that :)

magik6k: Done

core/coreapi/interface/interface.go
vyzo: On the topic of future extensibility, we should also consider adding TTL/expiry to the returned values.

magik6k:
Waiting for https://github.com/ipfs/go-ipfs/pull/4471 to be merged for test utils

whyrusleeping:
@magik6k merged the other PR, let me know when to look at this one again

core/coreapi/name_test.go
```diff
@@ -0,0 +1,144 @@
...
+		return
+	}
+
+	if resPath.String() != p.String() {
+		t.Errorf("expected paths to match, '%s'!='%s'", resPath.String(), p.String())
+	}
+}
+
+func TestBasicPublishResolveTimeout(t *testing.T) {
+	t.Skip("ValidTime doesn't appear to work at this time resolution")
```
magik6k: @whyrusleeping, I noticed that ValidTime doesn't seem to work at this time scale. Is this by design (should I drop this test), or is there a better way to do this (or is it a bug)?

magik6k:
Added tests, fixed some bugs they caught, other than that one question and after test are reviewed this PR is RTM.

core/coreapi/interface/interface.go
```diff
@@ -90,6 +103,78 @@ type DagAPI interface {
...
+	// should be used for the key. Default is options.RSAKey
+	//
+	// Supported key types:
+	// * options.RSAKey
+	// * options.Ed25519Key
+	WithType(algorithm string) options.KeyGenerateOption
+
+	// WithSize is an option for Generate which specifies the size of the key to
+	// generated. Default is 0
+	WithSize(size int) options.KeyGenerateOption
```
whyrusleeping: It feels a bit weird having these options be part of the interface... cc @lgierth

core/coreapi/name.go
```diff
@@ -0,0 +1,170 @@
...
+		}
+
+		pubKey := privKey.GetPublic()
+
+		pid, err := peer.IDFromPublicKey(pubKey)
+		if err != nil {
+			return nil, err
+		}
+
+		if pid.Pretty() == k {
```
whyrusleeping: when could this ever not be the case?

core/coreapi/name_test.go
whyrusleeping: hrm... thats tricky. It should be okay at 100ms... I'm not sure this isnt a bug. How does it fail?

core/coreapi/name.go
lgierth: We're iterating over all keys in the keystore and this is our return condition -- it's not a sanity assertion :)

core/coreapi/interface/interface.go
lgierth: The thinking was that in the future some of these option funcs might depend on state of the KeyAPI object. But actually in these cases we can just pass it in as an argument, e.g. `keyapi.Generate(ctx, "mykey", coreopts.WithType(coreopts.Ed25519Key))`

Drawback: the likelyhood of name collisions increases: there's more "type" options than just the one in key/generate. We can prefix these funcs, e.g. `coreopts.WithGenerateType`, or we can make a conversion similar to the `api.KeyAPI()` conversions: `opts := api.Options(); keyopts := opts.KeyOptions()`

whyrusleeping: Oh, so all the options are in the same package?

core/coreapi/name.go
whyrusleeping: Ah, i see now. Carry on

core/coreapi/interface/interface.go
magik6k: They are in one package, though are scoped per API, see https://github.com/ipfs/go-ipfs/pull/4477/files#diff-764a71d2f522b4136e7909befd573776R53.

For key API, `type` will always mean 'key type'. In the case where we create another function for which it would make sense to pass this option, it's possible to reuse that (with some more go types magic)

PR #4479, opened 2017-12-11
Document global options (-r, --enc and --timeout)

### LINKED ISSUE #4432: Document cmdkit's global options
#### Description:

The global options declared by cmdkit should be documented on the root command, and on all other commands.

See https://github.com/ipfs/go-ipfs-cmdkit/blob/f3631e8ddde711a7aefed041806902d907a1f9ae/option.go#L22-L30

People keep asking about e.g. `--timeout` :)

keks:
Closes #4432.

Is this what you had in mind or do you want more detail?

### DIFF
--- a/core/commands/root.go
+++ b/core/commands/root.go
@@ -25,7 +25,7 @@ const (
 var Root = &cmds.Command{
 	Helptext: cmdkit.HelpText{
 		Tagline:  "Global p2p merkle-dag filesystem.",
-		Synopsis: "ipfs [--config=<config> | -c] [--debug=<debug> | -D] [--help=<help>] [-h=<h>] [--local=<local> | -L] [--api=<api>] <command> ...",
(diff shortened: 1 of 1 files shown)


lgierth:
Ideally they'd also show up in the block below that summary line:

```
OPTIONS

  -c,   --config string - Path to the configuration file to use.
  -D,   --debug  bool   - Operate in debug mode.
  --help         bool   - Show the full command help text.
  -h             bool   - Show a short version of the command help text.
  -L,   --local  bool   - Run the command locally, instead of using the daemon.
  --api          string - Use a specific API instance (defaults to /ip4/127.0.0.1/tcp/5001).
```

These ^ are the global options that have been documented so far.

whyrusleeping:
@magik6k do you think you could handle getting this one ready to go?

magik6k:
Looks like the options are now documented in `ipfs --help`, I changed the synopsis to reflect those

gammazero:
Issue #4432 is closed and documentation is present in help text.

PR #4453, opened 2017-12-04
gx: update go-multihash

Stebalien:
Blocked on:

* [x] https://github.com/ipfs/go-ipfs/pull/4455 (broken)
* [x] A gx update of go-ipfs-cmdkit in go-ipfs-cmds.
* [x] https://github.com/ipfs/go-ipfs/pull/4398 (go 1.9)
* [x] https://github.com/ipfs/go-ipfs-cmds/pull/46
* [x] https://github.com/ipfs/go-ipfs-cmds/pull/44

**Shiny ✨**:

* Fewer allocations when parsing multiaddrs (and when dialing). This should significantly reduce CPU usage when using the DHT.
* Default identity hash length to -1 (auto). Was 32 bits.
* Fixes websocket goroutine leak: https://github.com/libp2p/go-ws-transport/issues/21
* CID prefix constructor (🎉). You can now make CIDs by calling `cid.NewPrefixV1(cid.DagCBOR, mh.SHA_256).Sum(data)`. (TODO: Make go-ipfs actually use this, it's a lot nicer than constructing these manually).
* cmds fixes: #4360, #4436, https://github.com/ipfs/go-ipfs-cmds/pull/46, https://github.com/ipfs/go-ipfs-cmds/pull/44

PR #4482, opened 2017-12-13
fix loops in sharness tests to fail the test if the inner command fails

Stebalien:
This was hiding errors.


core/commands/root.go
```diff
@@ -171,7 +171,6 @@ var rootROOldSubcommands = map[string]*oldcmds.Command{
 			"links": ocmd.ObjectLinksCmd,
 			"get":   ocmd.ObjectGetCmd,
 			"stat":  ocmd.ObjectStatCmd,
-			"patch": ocmd.ObjectPatchCmd,
```
Stebalien: @whyrusleeping ^^

test/sharness/t0110-gateway.sh
```diff
@@ -139,11 +139,32 @@ test_expect_success "refs IPFS directory file through readonly API succeeds" '
   test_curl_gateway_api "refs?arg=$HASH2/test"
 '
 
-test_expect_success "test gateway api is sanitized" '
-  for cmd in "add" "block/put" "bootstrap" "config" "dht" "diag" "dns" "get" "id" "mount" "name/publish" "object/put" "object/new" "object/patch" "pin" "ping" "refs/local" "repo" "resolve" "stats" "swarm"  "file" "update" "version" "bitswap"; do
+for cmd in add  \
```
Stebalien: I removed `get`, `version`, `resolve`, and `dns` as these are now exposed on the gateway.

core/commands/root.go
whyrusleeping: ACK

whyrusleeping:
You have angered the CI gods. You must now pay a penance

whyrusleeping:
@Stebalien don't break tests. Thats considered a bad thing.

whyrusleeping:
You could probably just delete the tests that are failing.

Stebalien:
@whyrusleeping  the tests will be fixed after merging #4453.

Stebalien:
Test failure was #4494

PR #4492, opened 2017-12-14
coreapi: Basic object API implementation

magik6k:
TODO:
- [ ] Review
- [x] Docs
- [x] Tests

### DIFF
--- a/core/coreapi/interface/interface.go
+++ b/core/coreapi/interface/interface.go
@@ -191,31 +191,72 @@ type KeyAPI interface {
 	Remove(ctx context.Context, name string) (Path, error)
 }
 
-//TODO: Should this use paths instead of cids?
+// ObjectAPI specifies the interface to MerkleDAG and contains useful utilities
+// for manipulating MerkleDAG data structures.
 type ObjectAPI interface {
+	// New creates new, empty (by default) dag-node.
 	New(context.Context, ...options.ObjectNewOption) (Node, error)
+
+	// WithType is an option for New which allows to change the type of created
+	// dag node.
+	//
+	// Supported types:
+	// * 'empty' - Empty node
+	// * 'unixfs-dir' - Empty UnixFS directory
 	WithType(string) options.ObjectNewOption
 
+	// Put imports the node into merkledag
 	Put(context.Context, Node) (Path, error)
+
+	// Get returns the node for the path
 	Get(context.Context, Path) (Node, error)
+
+	// Data returns reader for data of the node
 	Data(context.Context, Path) (io.Reader, error)
+
+	// Links returns lint or links the node contains
 	Links(context.Context, Path) ([]*Link, error)
+
+	// Stat returns information about the node
 	Stat(context.Context, Path) (*ObjectStat, error)
 
+	// AddLink adds a link under the specified path. child path can point to a
+	// subdirectory within the patent which must be present (can be overridden
+	// with WithCreate option).
 	AddLink(ctx context.Context, base Path, name string, child Path, opts ...options.ObjectAddLinkOption) (Node, error)
+
+	// WithCreate is an option for AddLink which specifies whether create required
+	// directories for the child
 	WithCreate(create bool) options.ObjectAddLinkOption
 
-	RmLink(context.Context, Path, string) (Node, error)
+	// RmLink removes a link from the node
+	RmLink(ctx context.Context, base Path, link string) (Node, error)
+
+	// AppendData appends data to the node
 	AppendData(context.Context, Path, io.Reader) (Node, error)
+
+	// SetData sets the data contained in the node
 	SetData(context.Context, Path, io.Reader) (Node, error)
 }
 
+// ObjectStat provides information about dag nodes
 type ObjectStat struct {
-	Cid            *cid.Cid
-	NumLinks       int
-	BlockSize      int
-	LinksSize      int
-	DataSize       int
+	// Cid is the CID of the node
+	Cid *cid.Cid
+
+	// NumLinks is number of links the node contains
+	NumLinks int
+
+	// BlockSize is size of the raw serialized node
+	BlockSize int
+
+	// LinksSize is size of the links block section
+	LinksSize int
+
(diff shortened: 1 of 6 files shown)


core/coreapi/interface/interface.go
```diff
@@ -38,27 +38,29 @@ type UnixfsAPI interface {
...
 
-// type ObjectStat struct {
-// 	Cid            cid.Cid
-// 	NumLinks       int
-// 	BlockSize      int
-// 	LinksSize      int
-// 	DataSize       int
-// 	CumulativeSize int
-// }
+	AddLink(ctx context.Context, base Path, name string, child Path, create bool) (Node, error) //TODO: make create optional
```
magik6k: I'll make this optional after the approach to optional params in https://github.com/ipfs/go-ipfs/pull/4477 is reviewed/accepted

core/coreapi/object.go
```diff
@@ -0,0 +1,190 @@
...
+
+	_, err := api.node.DAG.Add(node)
+	if err != nil {
+		return nil, err
+	}
+	return node, nil
+}
+
+func (api *ObjectAPI) Put(context.Context, coreiface.Node) error {
+	return errors.New("todo") // TODO: what should this method take? Should we just redir to dag-put?f
```
magik6k: I'm not even sure if we want to keep this method, it seems to me that all of it's use cases are covered by Dag.Put

core/coreapi/interface/interface.go
lgierth: No need for arg names in an interface definition

core/coreapi/interface/interface.go
```diff
@@ -38,27 +38,29 @@ type UnixfsAPI interface {
...
-// 	RmLink(string, string) (cid.Cid, error)
-// }
+//TODO: Should this use paths instead of cids?
+type ObjectAPI interface {
+	New(ctx context.Context) (Node, error)
+	Put(context.Context, Node) error
+	Get(context.Context, Path) (Node, error)
+	Data(context.Context, Path) (io.Reader, error)
+	Links(context.Context, Path) ([]*Link, error)
+	Stat(context.Context, Path) (*ObjectStat, error)
```
lgierth: I've been pondering in #4418 whether we should return a resolved `Path` in almost any place where we already have it resolved internally anyway. In some cases it saves a ton of additional resolve roundtrips, imagine we use the returned `Path` to pass it directly into another Core API function.

This is especially useful in the case of functions that modify existing objects, e.g. SetData or AddLink. If I call it with `/ipfs/QmFoo/a/b/c`, I'd like it to return `/ipfs/QmNewHash/a/b/c`.

(Regardless of that, `Put()` should definitely return a `Path`.)

core/coreapi/object.go
```diff
@@ -0,0 +1,190 @@
...
+
+	coreiface "github.com/ipfs/go-ipfs/core/coreapi/interface"
+	dag "github.com/ipfs/go-ipfs/merkledag"
+	ft "github.com/ipfs/go-ipfs/unixfs"
+)
+
+type ObjectAPI CoreAPI
+
+func (api *ObjectAPI) New(ctx context.Context) (coreiface.Node, error) {
+	node := new(dag.ProtoNode)
```
lgierth: On the CLI/API, `object new` takes a template argument which can be `unixfs-dir` for a directory. We can do this here similarly to the `KeyAPI.WithType()` option in #4477.

core/coreapi/object.go
lgierth: We should still implement it -- it can call `DagAPI.Put()` underneath

core/coreapi/object.go
```diff
@@ -0,0 +1,190 @@
...
+		LinksSize:      stat.LinksSize,
+		DataSize:       stat.DataSize,
+		CumulativeSize: stat.CumulativeSize,
+	}
+
+	return out, nil
+}
+
+func (api *ObjectAPI) AddLink(ctx context.Context, base coreiface.Path, name string, child coreiface.Path, create bool) (coreiface.Node, error) {
+	rootNd, err := api.core().ResolveNode(ctx, base)
```
lgierth: Let's call these variables `base` instead of `root` -- the root of `/ipfs/QmFoo/a/b` would be QmFoo, but the base of this operation is `b`

core/coreapi/interface/interface.go
```diff
@@ -178,27 +181,90 @@ type KeyAPI interface {
...
+
+	// AppendData appends data to the node
+	AppendData(context.Context, Path, io.Reader) (Path, error)
+
+	// SetData sets the data contained in the node
+	SetData(context.Context, Path, io.Reader) (Path, error)
+}
+
+// ObjectStat provides information about dag nodes
+type ObjectStat struct {
```
magik6k: Should I wrap this into an interface for consistency?

lgierth: I think it's fine - given that the object API is quasi-frozen (it's the old merkledag/dag-pb structure), it's very unlikely this struct will ever have any functions attached to it.

core/coreapi/interface/options/object.go
```diff
@@ -0,0 +1,91 @@
...
+	Create bool
+}
+
+type ObjectNewOption func(*ObjectNewSettings) error
+type ObjectPutOption func(*ObjectPutSettings) error
+type ObjectAddLinkOption func(*ObjectAddLinkSettings) error
+
+func ObjectNewOptions(opts ...ObjectNewOption) (*ObjectNewSettings, error) {
+	options := &ObjectNewSettings{
+		Type: "empty",
```
lgierth: The default for this should be "unixfs-dir" as in `ipfs object new --help`

lgierth: > The default for this should be "unixfs-dir" as in ipfs object new --help

Ah, nevermind, I understand :)

lgierth:
This LGTM 👍 - pretty cool!

Next let's extract coreiface to its own package, so we can get go-ipfs-api going.

Issue #4374, opened 2017-11-06
TestBasicBitswap failure


Stebalien:
```
--- FAIL: TestBasicBitswap (0.02s)
	bitswap_test.go:320: Test a one node trying to get one block from another
	bitswap_test.go:363: mismatch in blocks sent: 0 vs 1
```

Stebalien:
It looks like the fetching node is fetching the block from itself...

Stebalien:
I haven't seen this in ages. Closing.

PR #4497, opened 2017-12-15
commands: remove dead code

Stebalien:
@keks, do you know if this cod was supposed to do something?


keks:
At some point it had a purpose but apparently I forgot to delete it when I removed the code that was using it.

Stebalien:
Test failure was #4374.

PR #4499, opened 2017-12-15
improve bitswap tests

### LINKED ISSUE #2444: bitswap tests fail when run with race detector
It doesnt report a race condition, just that its using too many goroutines

Stebalien:
* fix races in testnet
* improve TestBasicBitswap (and make debugging it easier). 

In attempt to debug #4374 and #4108.

Also fixes #2444.

### DIFF
--- a/exchange/bitswap/bitswap_test.go
+++ b/exchange/bitswap/bitswap_test.go
@@ -108,7 +108,7 @@ func TestLargeSwarm(t *testing.T) {
 	if detectrace.WithRace() {
 		// when running with the race detector, 500 instances launches
 		// well over 8k goroutines. This hits a race detector limit.
-		numInstances = 100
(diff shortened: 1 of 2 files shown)


exchange/bitswap/bitswap_test.go
```diff
@@ -108,7 +108,7 @@ func TestLargeSwarm(t *testing.T) {
 	if detectrace.WithRace() {
 		// when running with the race detector, 500 instances launches
 		// well over 8k goroutines. This hits a race detector limit.
-		numInstances = 100
+		numInstances = 75
```
Stebalien: Max go routines with race detector: 8196.
Bitswap workers per peer: n^2.

75 means 75^2 = 5625

That gives us some healthy head-room for other random goroutines.

Issue #4280, opened 2017-10-05
Better buffering


Stebalien:
We should add buffering to all of our message-based APIs so we avoid sending message fragments. This should alleviate libp2p/go-libp2p#235.

Stebalien:
I've now tried:

* A "double buffer" buffering library (write to a buffer while another thread flushes).
* Manually adding buffering to bitswap to avoid sending messages lengths separate from messages.

No dice. Nothing makes a difference. The thing is, bitswap doesn't send a lot of little messages (to the same peer). We send a lot of little messages to many peers (buffering won't help) and many large messages to a single peer (buffering won't help). Even if we end up sending the message length in a separate packet/secio-message, that extra message is easily amortized by the size of the primary message in most cases.

TL;DR: Dead end.

Stebalien:
I take this back. Adding double buffering (with a small delay) helps reduce the *number* of packets quite  a bit (testing with wireshark). I believe the issues with my prior testing were:

1. I was testing bitswap, not stuff like the DHT (tons of small packets). Buffering should help background bandwidth.
2. Localhost is not the right environment to test on. Reducing the number of packets sent is significantly more important on the open internet.

Stebalien:
Closing in favor of #3155.

PR #4503, opened 2017-12-19
buffer bitswap messages

Stebalien:
Otherwise, we write the length and the message as two separate values.

Yes, we'll end up allocating a bit more than we would otherwise but that's
strictly better than sending two secio messages (with two separate MACs).

part of #4280

### DIFF
--- a/exchange/bitswap/message/message.go
+++ b/exchange/bitswap/message/message.go
@@ -1,6 +1,7 @@
 package message
 
 import (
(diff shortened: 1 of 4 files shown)


Stebalien:
This doesn't help *at all*. Not for large chunks, not for small chunks, not for directories, not at all. Maybe some kind of delay based buffering system?

PR #4506, opened 2017-12-19
Add Opentracing plugin support

frrist:
define an interface for creating and configuring tracers
add the Jaeger tracer as a plugin

License: MIT
Signed-off-by: ForrestWeston <forrest@protocol.ai>

### DIFF
--- a/plugin/loader/initializer.go
+++ b/plugin/loader/initializer.go
@@ -3,6 +3,7 @@ package loader
 import (
 	"github.com/ipfs/go-ipfs/core/coredag"
 	"github.com/ipfs/go-ipfs/plugin"
+	"gx/ipfs/QmWLWmRVSiagqP15jczsGME1qpob6HDbtbHAY2he9W5iUo/opentracing-go"
 
 	ipld "gx/ipfs/Qme5bWv7wtjUNGsK2BNGVUFPKiuxWrsqrtvYwCLRw8YFES/go-ipld-format"
 )
@@ -20,24 +21,37 @@ func initialize(plugins []plugin.Plugin) error {
 
 func run(plugins []plugin.Plugin) error {
 	for _, pl := range plugins {
-		err := runIPLDPlugin(pl)
-		if err != nil {
-			return err
+		switch pl := pl.(type) {
+		case plugin.PluginIPLD:
+			err := runIPLDPlugin(pl)
+			if err != nil {
+				return err
+			}
+		case plugin.PluginTracer:
+			err := runTracerPlugin(pl)
+			if err != nil {
+				return err
+			}
+		default:
+			panic(pl)
 		}
 	}
 	return nil
 }
 
-func runIPLDPlugin(pl plugin.Plugin) error {
-	ipldpl, ok := pl.(plugin.PluginIPLD)
-	if !ok {
-		return nil
-	}
-
-	err := ipldpl.RegisterBlockDecoders(ipld.DefaultBlockDecoder)
+func runIPLDPlugin(pl plugin.PluginIPLD) error {
+	err := pl.RegisterBlockDecoders(ipld.DefaultBlockDecoder)
 	if err != nil {
 		return err
 	}
+	return pl.RegisterInputEncParsers(coredag.DefaultInputEncParsers)
+}
 
-	return ipldpl.RegisterInputEncParsers(coredag.DefaultInputEncParsers)
+func runTracerPlugin(pl plugin.PluginTracer) error {
+	tracer, err := pl.InitTracer()
+	if err != nil {
+		return err
+	}
+	opentracing.SetGlobalTracer(tracer)
+	return nil
 }
--- a/package.json
+++ b/package.json
@@ -581,6 +581,12 @@
       "hash": "QmQine7gvHncNevKtG9QXxf3nXcwSj6aDDmMm52mHofEEp",
       "name": "tar-utils",
       "version": "0.0.3"
+    },
+    {
+      "author": "frist",
+      "hash": "QmWLWmRVSiagqP15jczsGME1qpob6HDbtbHAY2he9W5iUo",
+      "name": "opentracing-go",
+      "version": "0.0.3"
     }
   ],
   "gxVersion": "0.10.0",
--- a/plugin/tracer.go
+++ b/plugin/tracer.go
@@ -1,0 +1,11 @@
+package plugin
+
+import (
+	"gx/ipfs/QmWLWmRVSiagqP15jczsGME1qpob6HDbtbHAY2he9W5iUo/opentracing-go"
+)
+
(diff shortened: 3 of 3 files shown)


ForrestWeston:
Points worth noting:
- Plugins do not work on windows.
- Jaeger _should_ be version controlled (currently its not), not sure how to accomplish this yet.

ForrestWeston:
Not sure what's going wrong with Jenkins here, the plugin isn't enabled in the PR, and it looks like all tests pass (not sure on the exit code though):
```
# passed all 243 test(s)
1..243
script returned exit code 2
```

plugin/loader/initializer.go
```diff
@@ -41,3 +52,15 @@ func runIPLDPlugin(pl plugin.Plugin) error {
 
 	return ipldpl.RegisterInputEncParsers(coredag.DefaultInputEncParsers)
 }
+
+func runTracerPlugin(pl plugin.Plugin) error {
```
whyrusleeping: since we're doing a type cast above, we can make this function just accept a `PluginTracer` and skip the cast here. Should do the same for the ipld one too

plugin/plugins/jaeger/jaeger.go
```diff
@@ -0,0 +1,50 @@
...
+func (*jaegerPlugin) Version() string {
+	return "0.0.1"
+}
+
+func (*jaegerPlugin) Init() error {
+	return nil
+}
+
+//Initalize a Jaeger tracer and set it as the global tracer in opentracing api
+func (*jaegerPlugin) InitGlobalTracer(cfg map[string]interface{}) error {
```
whyrusleeping: we should probably use values from the cfg input in the configuration we do here.

ForrestWeston: Good point. I think instead of using a `cfg` we should define tracer specific options in the plugins themselves, since config values can differ across different tracer implementations.

plugin/loader/initializer.go
```diff
@@ -20,24 +20,37 @@ func initialize(plugins []plugin.Plugin) error {
 
 func run(plugins []plugin.Plugin) error {
 	for _, pl := range plugins {
-		err := runIPLDPlugin(pl)
-		if err != nil {
-			return err
+		switch pl.(type) {
+		case plugin.PluginIPLD:
+			err := runIPLDPlugin(pl.(plugin.PluginIPLD))
```
whyrusleeping: no need to do the cast here, the type switch does some magic for you.

plugin/loader/initializer.go
```diff
@@ -20,24 +20,37 @@ func initialize(plugins []plugin.Plugin) error {
...
+	if err != nil {
+		return err
 	}
 
-	err := ipldpl.RegisterBlockDecoders(format.DefaultBlockDecoder)
+	return pl.RegisterInputEncParsers(coredag.DefaultInputEncParsers)
+}
+
+func runTracerPlugin(pl plugin.PluginTracer) error {
+	err := pl.InitGlobalTracer()
```
whyrusleeping: To make type checks easier, I would do:
```go
tr, err := pl.GetTracer()
if err != nil {
  return err
}
opentrace.SetGlobalTracer(tr)
```

That way if the types in your plugin are wrong, this won't fail silently, it will hit type assertion issues and throw an error.

plugin/loader/initializer.go
frrist: Are we sure about this? building with `err := runIPLDPlugin(pl)` instead of `err := runIPLDPlugin(pl.(plugin.PluginIPLD))` fails

frrist:
```
plugin/loader/initializer.go:26:24: cannot use pl (type "github.com/ipfs/go-ipfs/plugin".Plugin) as type "github.com/ipfs/go-ipfs/plugin".PluginIPLD in argument to runIPLDPlugin:
        "github.com/ipfs/go-ipfs/plugin".Plugin does not implement "github.com/ipfs/go-ipfs/plugin".PluginIPLD (missing RegisterBlockDecoders method)
```

whyrusleeping: Ah, i forgot one bit, it needs to be:
```go
switch pl := pl.(type) {
```

cmd/ipfs/main.go
```diff
@@ -181,15 +180,14 @@ func makeExecutor(req *cmds.Request, env interface{}) (cmds.Executor, error) {
...
-		pluginpath := filepath.Join(cctx.ConfigRoot, "plugins")
 
 		// check if repo is accessible before loading plugins
 		ok, err := checkPermissions(cctx.ConfigRoot)
 		if err != nil {
 			return nil, err
 		}
 		if ok {
-			if _, err := loader.LoadPlugins(pluginpath); err != nil {
+			if _, err := loader.LoadPlugins(cctx.ConfigRoot); err != nil {
```
whyrusleeping: Hrm... making that code open up and parse the config on its own breaks some abstractions. The fact that the config is a file on disk is an implementation detail, We might need to rethink some things here.

cc @Stebalien @magik6k We need to get some configuration, and the peer ID to the tracing plugin. I'm not entirely sure yet how that should get fed through, ideas?

frrist: For some context, here is how it may be used with the tracing plugin
https://github.com/ipfs/go-jaeger-plugin/pull/1/files#diff-e92ca16875e3c00b8a36338613029bf9R35

magik6k: This will also break changes in https://github.com/ipfs/go-ipfs/pull/4244/files#diff-153cb71153ee178fe2869148600ce8d5R197.

The problem is that at this stage we may not even have a peerID/config (i.e. user calling `ipfs init`), we may add some sort of late init for plugins that want it and call that above in `buildEnv`, either before or after calling `core.NewNode`

frrist: For now setting the tracer name could be a manual process - maybe we don't want it to be the nodeID, I was doing this for convenience. 

This would still be done via a configuration file (where the tracer name is a field), but that file would reside in the plugin path - possibly under a directory based on the plugin name (e.g. `pluginPath/pluginName/config.cfg`).

I imagine we will want the ability to pass plugins config files in the future?

plugin/loader/initializer.go
```diff
@@ -3,6 +3,7 @@ package loader
 import (
 	"github.com/ipfs/go-ipfs/core/coredag"
 	"github.com/ipfs/go-ipfs/plugin"
+	"gx/ipfs/QmWLWmRVSiagqP15jczsGME1qpob6HDbtbHAY2he9W5iUo/opentracing-go"
```
magik6k: Import grouping

Kubuxu:
I've rebased it on top of the master to allow #4806 to base off it.

Issue #4505, opened 2017-12-19
potential memory leak in ipfs add


whyrusleeping:
Several users have reported ipfs using way too much memory when adding large datasets.

This could be that the mfs structure is not ever flushing directories from memory, we probably need to periodically call `Flush` on parts of the structure during longer adds.

Stebalien:
> Several users have reported ipfs using way too much memory when adding large datasets.

That may be providing. We queue up a bunch of go routines to provide blocks and have no backpressure mechanism. Ask them to try again with `--local` and see if that makes a difference.

Stebalien:
Actually, it looks like you're right.

![add-profile](https://user-images.githubusercontent.com/310393/34173041-06221c10-e4ec-11e7-8975-619fc3a04356.png)

Stebalien:
Several issues:

1. When adding, we need to stream CIDs back to the user instead of queuing them all up.
2. When providing, we need to *not* queue CIDs in memory.
3. Our CIDs are *large*. The 48 bytes for the CID structure and 48 bytes for the `Cid` structure, 48 bytes for the CID multihash (according to the profiler). We should really jump on https://github.com/ipfs/go-cid/issues/3.
4. We allocate 256KiB buffers when importing but don't make any attempts to shrink them down if we happen to read a smaller file. This means we end up batching `ParallelBatchCommits(2*NumCPU) * MatchBatchBlocks(128)` 256KiB blocks in our DAG batch writer. That's *not* good.

whyrusleeping:
@dignifiedquire hrm... thats troublesome, but the users reporting issues here said that their machines had at least 8GB of ram. That wouldnt account for that would it?

Stebalien:
We should try adding an LRU (or ARC) cache for CIDs.

Stebalien:
@whyrusleeping 

1. How many CPUs do they have.
2. You have to take GC into account. GC will lag a bit even if we're allocating a ton of buffers.

Stebalien:
(old)

PR #4508, opened 2017-12-20
Don't lock up 256KiB buffers when adding small files

Stebalien:
This is part of #4505.

~It's WIP because we still need to use some form of buffer pool. As is, allocating and throwing away 256KiB buffers is killing GC.~

### DIFF
--- a/core/coreunix/add.go
+++ b/core/coreunix/add.go
@@ -131,7 +131,7 @@ func (adder *Adder) SetMfsRoot(r *mfs.Root) {
 }
 
 // Constructs a node from reader's data, and adds it. Doesn't pin.
-func (adder Adder) add(reader io.Reader) (node.Node, error) {
+func (adder *Adder) add(reader io.Reader) (node.Node, error) {
 	chnk, err := chunk.FromString(reader, adder.Chunker)
 	if err != nil {
 		return nil, err
--- a/importer/chunk/parse.go
+++ b/importer/chunk/parse.go
@@ -11,7 +11,7 @@ import (
 func FromString(r io.Reader, chunker string) (Splitter, error) {
 	switch {
 	case chunker == "" || chunker == "default":
-		return NewSizeSplitter(r, DefaultBlockSize), nil
+		return DefaultSplitter(r), nil
 
 	case strings.HasPrefix(chunker, "size-"):
 		sizeStr := strings.Split(chunker, "-")[1]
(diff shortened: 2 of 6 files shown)


whyrusleeping:
It would be great if we could figure out a way to put all these buffers back into the buffer pool in the common case

kevina:
( @Stebalien for future reference, I find it annoying when people do a rebase that makes a major change.  (In this case using the mpool from go-msgio).  At least I would appreciate it if you just add new commits and save rebasing for fixups. )

Stebalien:
> It would be great if we could figure out a way to put all these buffers back into the buffer pool in the common case.

That would be nice but would require some form of refcounting on blocks (at least given the current interfaces).

> ( @Stebalien for future reference, I find it annoying when people do a rebase that makes a major change. (In this case using the mpool from go-msgio). At least I would appreciate it if you just add new commits and save rebasing for fixups. )

The new commit completely changed my approach so I figured I'd just do it over. I could have opened a new PR but I figured nobody had had looked at this one yet (that's when I start trying to preserve history). Sorry about that.

kevina:
@Stebalien How is using a memory pool helping here?  In the original code before the rebase you reused the 256Kib buffer unless the buffer was full.  You are doing the same basic thing here.  I don't think it is worth the extra dependency unless we provide (and use) a method to return the returned byte to the pool.

kevina:
More to the point, it looks like you are just using the mpool as a better allocator, is the default go allocator so bad that it helps to the a custom run?

Or is it the case that NewSizeSplitter is called for each new file, in which case I withdraw my objection.

Stebalien:
> Or is it the case that NewSizeSplitter is called for each new file, in which case I withdraw my objection.

It is. I'm using it for the case where we add a bunch of small files. Without this, memory usage is still unstable because the GC lags a bit.

The alternative is to re-use the splitter somehow. I toyed with a reset function but that would be a more invasive change.

kevina:
> It is. I'm using it for the case where we add a bunch of small files. Without this, memory usage is still unstable because the GC lags a bit.

Ok.

I was going to say what's one more dependency but:

> So, this uses the buffer pool from go-msgio. Should we break that into a separate package? It's used all over the place.

Yes i think we should.  I am not particularly happy with bringing in a whole package just to use one of its utility sub-packages.

Other than that it LGTM.

PR #4509, opened 2017-12-20
Ensure that flush on the mfs root flushes its directory

whyrusleeping:
During `ipfs add`, we periodically call flush on the mfs root to free up unused memory. Apparently `Flush` on the mfs root object wasnt actually calling `Flush` on its directory, which led to memory not actually being freed. I'm trying to think of the best way to test this.

License: MIT

### DIFF
--- a/mfs/system.go
+++ b/mfs/system.go
@@ -123,6 +123,36 @@ func (kr *Root) Flush() error {
 	return nil
 }
 
+// FlushMemFree flushes the root directory and then uncaches all of its links.
+// This has the effect of clearing out potentially stale references and allows
+// them to be garbage collected.
+// CAUTION: Take care not to ever call this while holding a reference to any
+// child directories. Those directories will be bad references and using them
+// may have unintended racy side effects.
+// A better implemented mfs system (one that does smarter internal caching and
+// refcounting) shouldnt need this method.
+func (kr *Root) FlushMemFree(ctx context.Context) error {
+	dir, ok := kr.GetValue().(*Directory)
+	if !ok {
+		return fmt.Errorf("invalid mfs structure, root should be a directory")
+	}
+
+	if err := dir.Flush(); err != nil {
+		return err
+	}
+
+	dir.lock.Lock()
+	defer dir.lock.Unlock()
+	for name := range dir.files {
+		delete(dir.files, name)
+	}
+	for name := range dir.childDirs {
+		delete(dir.childDirs, name)
+	}
+
+	return nil
+}
+
 // closeChild implements the childCloser interface, and signals to the publisher that
 // there are changes ready to be published
 func (kr *Root) closeChild(name string, nd node.Node, sync bool) error {
--- a/core/coreunix/add.go
+++ b/core/coreunix/add.go
@@ -443,9 +443,10 @@ func (adder *Adder) addFile(file files.File) error {
 		if err != nil {
 			return err
 		}
-		if err := mr.Flush(); err != nil {
+		if err := mr.FlushMemFree(adder.ctx); err != nil {
 			return err
 		}
+
 		adder.liveNodes = 0
 	}
 	adder.liveNodes++


Stebalien:
> I'm trying to think of the best way to test this.

We may want to add some general sharness tests for stuff like this. That is, upload a file, directory, large, large, small, etc. in one subprocess and repeatedly check memory usage in the other (and see if it goes above some high watermark). Unfortunately, tests like this tend to be flaky (but we could probably set the watermark high enough to avoid false positives).

Stebalien:
> Why broke it :(

Nevermind. I can't read.

mfs/dir.go
```diff
@@ -407,10 +415,14 @@ func (d *Directory) Path() string {
 }
 
 func (d *Directory) GetNode() (node.Node, error) {
+	return d.getNode(false)
+}
+
+func (d *Directory) getNode(clear bool) (node.Node, error) {
```
Stebalien: There *has* to be a cleaner way to do this.

whyrusleeping: yeah, i'm thinking about it. Its pretty obnoxious right now...

kevina: That cross my mind also, but I didn't find it objectionable enough to comment. :)

whyrusleeping:
So, while this fixes the memory leak problem, it does bring up another interesting issue. We have the potential for someone to hold a stale reference to a directory, after it has been removed.

whyrusleeping:
One thing we could do for now is have a different call that we only use in the add code since we know we won't run into any of these stale reference problems.

whyrusleeping:
There... thats a slightly better solution that doesnt introduce weirdness around the directory references.

In any case though, the mfs directory thing needs work. Otherwise any use of mfs will not clear out its cache.

mfs/system.go
```diff
@@ -123,6 +123,36 @@ func (kr *Root) Flush() error {
 	return nil
 }
 
+// FlushMemFree flushes the root directory and then uncaches all of its links.
+// This has the effect of clearing out potentially stale references and allows
+// them to be garbage collected.
+// CAUTION: Take care not to ever call this while holding a reference to any
+// child directories. Those directories will be bad references and using them
+// may have unintended racy side effects.
```
Stebalien: What exactly do you mean here? How will removing these from the map lead to race conditions?

I think what's throwing me is the term "cache". Is it a write cache? A read cache?

mfs/system.go
```diff
@@ -123,6 +123,36 @@ func (kr *Root) Flush() error {
...
+		return err
+	}
+
+	names, err := dir.ListNames(ctx)
+	if err != nil {
+		return err
+	}
+
+	for _, name := range names {
+		dir.Uncache(name)
```
Stebalien: Instead of listing then uncaching, can we just uncache all cached names? That `ListNames` call could get *expensive*.

mfs/system.go
```diff
@@ -123,6 +123,36 @@ func (kr *Root) Flush() error {
...
+// may have unintended racy side effects.
+// A better implemented mfs system (one that does smarter internal caching and
+// refcounting) shouldnt need this method.
+func (kr *Root) FlushMemFree(ctx context.Context) error {
+	dir, ok := kr.GetValue().(*Directory)
+	if !ok {
+		return fmt.Errorf("invalid mfs structure, root should be a directory")
+	}
+
+	if err := dir.Flush(); err != nil {
```
Stebalien: What happens if we concurrently modify this directory? Can we end up uncaching non-flushed entries? Is that an issue?

Should we have a single `FlushAndUncache()` method?

mfs/system.go
Stebalien: Specifically, with sharded directories.

mfs/system.go
whyrusleeping: Thats basically the purpose of this method. In my comment i make it explicit that you shouldnt use this method concurrently with any other operations since there are deeper issues that could potentially happen.

mfs/system.go
whyrusleeping: Oh yeah, youre right. This will go through all names, even if they arent cached.

mfs/system.go
Stebalien: Fair enough. I'm working on improving this situation, FYI.

whyrusleeping:
@Stebalien fixed the 'iterate over every single child' issue.

PR #4515, opened 2017-12-21
[WIP] MFS improvements

### LINKED ISSUE #4514: MFS: Race when reading node and taking read/write lock

Stebalien:
* Simpler mode/access checks.
* Syncing/flushing when we don't need to.


Stebalien:
TODO: Test the race.

Stebalien:
I think I'm going to take a different approach to modes.

PR #4517, opened 2017-12-21
[WIP] MFS improvements

### LINKED ISSUE #4514: MFS: Race when reading node and taking read/write lock

Stebalien:
* Improve flags and mode. Make file open options an extensible struct. Unix uses an int because it's easier to pass in a syscall and it has a fixed size. We don't care about that.
* Avoid syncing/flushing when we don't need to.

### DIFF
--- a/core/commands/files/files.go
+++ b/core/commands/files/files.go
@@ -477,7 +477,7 @@ Examples:
 			return
 		}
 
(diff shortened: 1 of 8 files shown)


schomatis:
This code has been moved to [`go-mfs`](https://github.com/ipfs/go-mfs), what can be extracted from this PR will be continued in https://github.com/ipfs/go-mfs/issues/32.

PR #4525, opened 2017-12-28
clear out memory after reads from the dagreader

whyrusleeping:
I would like to figure out how to add a test or two for this. I feel like testing the internals of the dagreader object is a bit pointless, and its pretty difficult to test if something has been garbage collected.

License: MIT

### DIFF
--- a/unixfs/io/dagreader_test.go
+++ b/unixfs/io/dagreader_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"io"
 	"io/ioutil"
+	"math/rand"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


lgierth:
Maybe a test could add a partial dag, make requests into ranges of it, and then verify we only hold roughly as much memory as we think is appropriate.

whyrusleeping:
Added a test, another round of review please

unixfs/io/dagreader_test.go
```diff
@@ -72,6 +73,44 @@ func TestSeekAndRead(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if !bytes.Equal(buf, inbuf[10000:10100]) {
+		t.Fatal("seeked read failed")
+	}
+
+	pbdr := reader.(*pbDagReader)
+	if pbdr.promises[0] != nil {
```
frrist: What if we checked the correct number of things to be loaded instead?

Issue #3578, opened 2017-01-09
Add + Support an  `offset` flag for `ipfs cat`


diasdavid:
What the title suggests :)

mib-kd743naq:
Surely `ipfs cat ... | tail -c +2000` is enough for this...?

diasdavid:
We need to have support for it coming from the http-api too.

lgierth:
Would Range headers satisfy this need on the http api? We get these for free with go's net/http. They're already used e.g. for skipping in a video/audio file through the gateway.

whyrusleeping:
@diasdavid how would we handle calling `ipfs cat` on multiple hashes? like `ipfs cat QmdrSmgViWKCjRntzsoiFhzQ1Wsce2ig5nZErS7SvXr9Ng QmWJ7ot9xrhkSHwCqm9XAhnXe944nXYQQ3tCSmXfnxUWNj` ?

diasdavid:
Good question. What about simply treating the concatenation of both files as a single file for the offset?

magik6k:
Done in #4538

Issue #4085, opened 2017-07-23
Partial Read / Seeking without downloading the complete file


NateWickstrom:
#### Description:

I'm working on a media player app (on Android) that uses IPFS to play remote music and video files.  It has the go-ipfs daemon running in the background and interacts with it via the http interface (http://127.0.0.1:5001/api/v0).  However, the **/get** command downloads the entire file before the stream is accessible, which isn't ideal since that can take 10s of seconds even for small files;  The **/cat** command only returns streams that start from the beginning of files, i.e. you can't seek the stream, but it can be opened quicker (in a few seconds) - i assume this is because, unlike **get**, the entire file isn't downloaded before it can be opened; And the **/files/read** command only works on local files, so this is a non-starter.

First of all, my assumptions are correct, right?  With go-ipfs you can't open a file from the middle, until it's completely downloaded.  I spent a few hours digging around the specs, code and open issues and couldn't find any specific mention of the feature being supported (yet?) so figured opening an issue was the next step.  I can post some example code, if needed...

Second, if this feature _is_ missing, what are some thoughts on its feasibility and implementation?

whyrusleeping:
Hey @NateWickstrom! Sorry for the long delay in responding. Accessing files through the gateway (127.0.0.1:8080/ipfs/<hash>) allows you to stream and seek, it supports the range headers required for streaming via http.

Currently, you cant seek with cat, but this is something we're planning on adding fairly soon.

As for `files/read`, files arent necessarily local, just referenced by the files api. You can do `ipfs files cp /ipfs/<hash> /foo` and it will put your file into the files tree without downloading it entirely. (this is super useful for interacting with very large directories of things).

negamaxi:
@whyrusleeping any updates on cat seeking?

whyrusleeping:
@negamaxi we just merged a `--offset` flag for `ipfs cat`. Does that work for what you need?

PR #4538, opened 2018-01-02
Add offset option to cat command

frrist:


frrist:
Discussed in issue #4085 & #3578

core/commands/cat.go
```diff
@@ -103,16 +116,26 @@ var CatCmd = &cmds.Command{
...
 		}
+		if offset > int64(read.Size()) {
+			offset = offset - int64(read.Size())
+			continue
+		}
+		count, err := read.Seek(offset, io.SeekStart)
+		if err != nil {
+			return nil, 0, err
+		}
+		offset = offset - count
```
whyrusleeping: shouldnt this always result in offset being 0?

test/sharness/t0040-add-and-cat.sh
```diff
@@ -46,11 +46,44 @@ test_add_cat_file() {
...
+  test_expect_success "ipfs cat with offset succeeds" '
+    ipfs cat --offset 10 "$HASH" >actual
+  '
+
+  test_expect_success "ipfs cat from offset output looks good" '
+    echo "ds!" >expected &&
+    test_cmp expected actual
+  '
+
+  test_expect_success "ipfs cat multipule hashes with offset succeeds" '
```
whyrusleeping: small typo

whyrusleeping:
a couple things, then LGTM.

whyrusleeping:
right, this closes #3578, but i think for #4085 we need a `--length` flag as well. I'm fine having that be a separate PR

Issue #4028, opened 2017-07-03
API for tracking data locallity and transfer progress


MichaelMure:
For my project (https://github.com/MichaelMure/TotallyNotArbore) I need to be able to solve two problems:
1. is this hash present locally, and if not, what percentage is here ?
1. progress statistics would be nice to have (rateIn, rateOut, peersIn, peersOut)

This could be done different ways:
1. `ipfs repo has hash` or `ipfs refs local hash` --> `yes/no/%`
1.  add some logs for added/removed block and use `ipfs log` to track what is going in and out of the repository. I actually did that at some point (https://github.com/ipfs/go-ipfs/pull/1892) but it has been deleted in a refactoring. It also mean that the app using ipfs has to replicate the graph of block in memory and track what happen. Not a fan of this solution
1. `ipfs repo feed` --> `feed of added/removed blocks`. Again, this means that the app has to track everything.
1. `ipfs stats progress hash...` or `ipfs stats bw --hash hash...` --> `[{ hash, local, progress, rateIn, peersIn, rateOut, peersOut }, ...]` It would output a feed of progress for a set of hash, as well as statistics of bandwidth and number of peers participating.

Before I start working on this, I'd like to know if it's something that could be merged in IPFS and what form should it take. Also, I'm not sure what would be practical to implement with the current codebase.

The minimal and easiest to implement is `1.`. I would prefer to have something like `4.` as well.

whyrusleeping:
option 1 feels like it should be a command on the `ipfs refs` subcommand. maybe `ipfs refs has` and `ipfs refs has -r`. It is not possible to generally know what percentage of objects in a given graph you have (any future block you fetch may have an unknown number of children). Unixfs files signal their total filesize, so this could be used as an estimate in that case. The best we can do generally is `number_have_local / number_known_to_exist_in_graph` but the second number will be continuously increasing until all branch nodes have been retrieved.

Your log events were removed in a refactor? Thats not good... Lets make an issue to replace them.

As for the metrics youre looking for, that gets tricky. But i'm betting we can accomplish this by creating a 'notifier' object that accepts notifications from bitswap on incoming and outgoing blocks. Callers could register interest in a block or graph and it could record pertinent info for them (not storing data for objects outside of the interesting set).

Issue #3866, opened 2017-04-17
Keystore/encryption/ipfs usecase for a file-sharing application


MichaelMure:
As asked by @whyrusleeping, here is my usecase for ipfs and in particular for the keystore/encryption feature.

I'm building a file sharing application presented as a social network. The main target is friend to friend sharing (think holiday pictures) with a control on who has access to the data by using encryption. The underlying goal is to reduce the need to use solution like Facebook, Dropbox, Drive... for data exchange that should be kept private and to create a [better solution](https://xkcd.com/949/).

The project is [here](https://github.com/MichaelMure/TotallyNotArbore) (no readme, sorry, I should do that). The tech stack is `electron`, `react`, `redux`, `ipfs-connector` --> `js-ipfs-api` --> `go-ipfs`. I'm working on it full-time and hope to have a first (crippled) version in 1-2 months (there is no chance I'll regret having said that in the future, right ? ;) )

What is described below is a general plan and not fully figured out.

## Identity
I don't use a central authority like an identity server or a blockchain. As the target is a friend-to-friend network, I chose to have identity simply being a crypto keypair (`ipfs key gen` with a `passphrase`) and to build a trust network contact-list for each user. Login in the application will be selecting a profile and entering the passphrase to unlock the private key. Each user publish a profile in his IPNS namespace (`ipfs name publish --key mykey`), including his public key. I guess it's a good idea to also sign this data (`ipfs crypt sign`).

## Contact management
Each user will have a contact list with the keys stored in the `keystore`. Contact can be added 'manually' by entering the public key hash of the contact. The profile is then resolved on IPNS (`ipfs name resolve contact-hash`), fetched, verified (`ipfs crypt verify`)  and imported in the application. I need a way to import the public key in the keystore (**MISSING**). As both user are not necessarily online at the same time `ipfs key send` will not do as it is a push. I need a pull version from the profile data.

Users share their contact list with their contact (with some privacy settings). That mean that once the first contact is added, the application can build a suggestion list for the user to easily add new contacts and build his social graph. The same way as before, the profile is retrieved from ipfs and stored locally.

## Sharing and encryption

There is 3 kind of sharing. The ~~first~~ second version (I guess the first version will have no encryption at all) will only provide only the targeted sharing.

When a user share something, the data is added to the IPFS repo (using the no-copy filestore :heart:). Metadata are stored as well and a notification is sent using pubsub (`ipfs pubsub pub contact-hash share-hash`).

[...]

PR #4541, opened 2018-01-03
Implement the command "ipfs repo has"

MichaelMure:
Following https://github.com/ipfs/go-ipfs/issues/4028, I implemented a `ipfs repo has` command that tell if something is available locally or not.
```
USAGE
  ipfs repo has <key>... - Show if an object is available locally

SYNOPSIS
  ipfs repo has [--recursive | -r] [--] <key>...

ARGUMENTS

  <key>... - Key(s) to check for locality.

OPTIONS

  -r, --recursive bool - Check recursively the graph of objects to build more precise stats.
```

Results look like this:
```
non-recursive
$ ipfs repo has QmYGVhjTfVvBAAf2SAWMJsTDheo7UuyjQigGahmB8YU3ZH
QmYGVhjTfVvBAAf2SAWMJsTDheo7UuyjQigGahmB8YU3ZH	local
$ ipfs repo has QmPXvegq26x982cQjSfbTvSzZXn7GiaMwhhVPHkeTEhrPT
QmPXvegq26x982cQjSfbTvSzZXn7GiaMwhhVPHkeTEhrPT	non local

recursive
$ ipfs repo has -r QmQFnam7qMXfF8R1D3qrETCgESnVP8gt6iEnAbyG73gyqm
QmQFnam7qMXfF8R1D3qrETCgESnVP8gt6iEnAbyG73gyqm	unknow
$ ipfs repo has -r QmQLXHs7K98JNQdWrBB2cQLJahPhmupbDjRuH1b9ibmwVa
QmQLXHs7K98JNQdWrBB2cQLJahPhmupbDjRuH1b9ibmwVa	incomplete	4.1 MB of 6.9 MB (60.31%)
$ ipfs repo has -r QmeUKqHMr5r6wJzt3AVScnUAyC4TkdZGjYJNaFBCL1BmAv
QmeUKqHMr5r6wJzt3AVScnUAyC4TkdZGjYJNaFBCL1BmAv	complete	3.2 MB of 3.2 MB (100.00%)
```

License: MIT

### DIFF
--- a/core/commands/repo.go
+++ b/core/commands/repo.go
@@ -436,9 +436,9 @@ Example:
 
 	non-recursive
 	> ipfs repo has QmYGVhjTfVvBAAf2SAWMJsTDheo7UuyjQigGahmB8YU3ZH
-	QmYGVhjTfVvBAAf2SAWMJsTDheo7UuyjQigGahmB8YU3ZH	local
+	QmYGVhjTfVvBAAf2SAWMJsTDheo7UuyjQigGahmB8YU3ZH	found
 	> ipfs repo has QmPXvegq26x982cQjSfbTvSzZXn7GiaMwhhVPHkeTEhrPT
-	QmPXvegq26x982cQjSfbTvSzZXn7GiaMwhhVPHkeTEhrPT	non local
+	QmPXvegq26x982cQjSfbTvSzZXn7GiaMwhhVPHkeTEhrPT	missing
 
 	recursive
 	> ipfs repo has -r QmQFnam7qMXfF8R1D3qrETCgESnVP8gt6iEnAbyG73gyqm
@@ -450,7 +450,7 @@ Example:
 `,
 	},
 	Arguments: []cmdkit.Argument{
-		cmdkit.StringArg("key", true, true, "Key(s) to check for locality.").EnableStdin(),
+		cmdkit.StringArg("cid", true, true, "CID(s) to check for locality.").EnableStdin(),
 	},
 	Options: []cmdkit.Option{
 		cmdkit.BoolOption("recursive", "r", "Check recursively the graph of objects to build more precise stats"),
@@ -468,16 +468,17 @@ Example:
 			return
 		}
 
-		// Decode all the keys
-		var keys []*cid.Cid
-		for _, arg := range req.Arguments() {
+		// Decode all the cids
+		args := req.Arguments()
+		cids := make([]*cid.Cid, len(args))
+		for i, arg := range args {
 			c, err := cid.Decode(arg)
 			if err != nil {
 				res.SetError(err, cmdkit.ErrNormal)
 				return
 			}
 
-			keys = append(keys, c)
+			cids[i] = c
 		}
 
 		// an offline DAGService will not fetch from the network
@@ -488,8 +489,8 @@ Example:
 
 		ctx := n.Context()
 
-		for _, k := range keys {
-			root, err := offlineDag.Get(ctx, k)
+		for _, cid := range cids {
+			root, err := offlineDag.Get(ctx, cid)
 			if err != nil && err != dag.ErrNotFound {
 				res.SetError(err, cmdkit.ErrNormal)
 				return
@@ -499,7 +500,7 @@ Example:
 
 			if !hasRoot || !recursive {
 				// we don't have precise size information
-				res.Emit(&localityOutput{Hash: k.String(), Local: hasRoot})
+				res.Emit(&localityOutput{Hash: cid.String(), Local: hasRoot})
 				continue
 			}
 
@@ -510,7 +511,7 @@ Example:
 			}
 
 			res.Emit(&localityOutput{
-				Hash:      k.String(),
+				Hash:      cid.String(),
 				Local:     local,
 				SizeLocal: sizeLocal,
 				SizeTotal: sizeTotal,
@@ -540,9 +541,9 @@ Example:
 				}
 			} else {
 				if lo.Local {
-					state = "local"
+					state = "found"
 				} else {
-					state = "non local"
+					state = "missing"
 				}
 			}
 
--- a/core/commands/repo.go
+++ b/core/commands/repo.go
@@ -414,7 +414,7 @@ var repoVersionCmd = &oldcmds.Command{
 	},
 }
 
-type LocalityOutput struct {
+type localityOutput struct {
 	Hash      string
 	Local     bool
 	SizeLocal int `json:",omitempty"`
@@ -476,7 +476,7 @@ var repoHasCmd = &cmds.Command{
 
 			if !hasRoot || !recursive {
 				// we don't have precise size information
-				res.Emit(&LocalityOutput{Hash: k.String(), Local: hasRoot})
+				res.Emit(&localityOutput{Hash: k.String(), Local: hasRoot})
 				continue
 			}
 
@@ -486,7 +486,7 @@ var repoHasCmd = &cmds.Command{
 				return
 			}
 
-			res.Emit(&LocalityOutput{
+			res.Emit(&localityOutput{
 				Hash:      k.String(),
 				Local:     local,
 				SizeLocal: sizeLocal,
@@ -496,7 +496,7 @@ var repoHasCmd = &cmds.Command{
 	},
 	Encoders: cmds.EncoderMap{
 		cmds.Text: cmds.MakeEncoder(func(req cmds.Request, w io.Writer, v interface{}) error {
-			lo, ok := v.(*LocalityOutput)
+			lo, ok := v.(*localityOutput)
 			if !ok {
 				return e.TypeErr(lo, v)
 			}
@@ -538,7 +538,7 @@ var repoHasCmd = &cmds.Command{
 			return nil
 		}),
 	},
-	Type: LocalityOutput{},
+	Type: localityOutput{},
 }
 
 func walkBlock(ctx context.Context, dagserv dag.DAGService, merkleNode node.Node) (bool, int, int, error) {
--- a/core/commands/repo.go
+++ b/core/commands/repo.go
@@ -2,6 +2,7 @@ package commands
 
 import (
 	"bytes"
+	"context"
 	"fmt"
 	"io"
 	"os"
@@ -10,14 +11,19 @@ import (
 	"text/tabwriter"
 
 	bstore "github.com/ipfs/go-ipfs/blocks/blockstore"
+	bservice "github.com/ipfs/go-ipfs/blockservice"
 	oldcmds "github.com/ipfs/go-ipfs/commands"
 	e "github.com/ipfs/go-ipfs/core/commands/e"
 	corerepo "github.com/ipfs/go-ipfs/core/corerepo"
+	"github.com/ipfs/go-ipfs/exchange/offline"
+	dag "github.com/ipfs/go-ipfs/merkledag"
 	config "github.com/ipfs/go-ipfs/repo/config"
 	fsrepo "github.com/ipfs/go-ipfs/repo/fsrepo"
 	lockfile "github.com/ipfs/go-ipfs/repo/fsrepo/lock"
 
+	node "gx/ipfs/QmNwUEK7QbwSqyKBu3mMtToo8SUc6wQJ7gdZq4gGGJqfnf/go-ipld-format"
 	cmds "gx/ipfs/QmP9vZfc5WSjfGTXmwX2EcicMFzmZ6fXn7HTdKYat6ccmH/go-ipfs-cmds"
+	humanize "gx/ipfs/QmPSBJL4momYnE7DcUyk2DVhD6rH488ZmHBGLbxNdhU44K/go-humanize"
 	cmdkit "gx/ipfs/QmQp2a2Hhb7F6eK2A5hN8f9aJy4mtkEikL9Zj4cgB7d1dD/go-ipfs-cmdkit"
 	cid "gx/ipfs/QmeSrf6pzut73u6zLQkRFQ3ygt3k6XFT2kjdYP8Tnkwwyg/go-cid"
 )
@@ -36,6 +42,7 @@ var RepoCmd = &cmds.Command{
 
 	Subcommands: map[string]*cmds.Command{
 		"stat": repoStatCmd,
+		"has":  repoHasCmd,
 	},
 	OldSubcommands: map[string]*oldcmds.Command{
 		"gc":      repoGcCmd,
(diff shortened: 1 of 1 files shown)


core/commands/repo.go
```diff
@@ -406,3 +413,167 @@ var repoVersionCmd = &oldcmds.Command{
...
+	Hash      string
+	Local     bool
+	SizeLocal int `json:",omitempty"`
+	SizeTotal int `json:",omitempty"`
+}
+
+var repoHasCmd = &cmds.Command{
+	Helptext: cmdkit.HelpText{
+		Tagline:          "Show if an object is available locally",
+		ShortDescription: ``,
```
magik6k: Some description would be nice

core/commands/repo.go
```diff
@@ -406,3 +413,167 @@ var repoVersionCmd = &oldcmds.Command{
...
+			}
+
+			var state string
+			if recursive {
+				state = "complete"
+				if !lo.Local {
+					state = "incomplete"
+				}
+				if lo.SizeTotal <= 0 {
+					state = "unknow"
```
magik6k: Typo: "unknown"

core/commands/repo.go
```diff
@@ -406,3 +413,167 @@ var repoVersionCmd = &oldcmds.Command{
...
+				}
+			} else {
+				if lo.Local {
+					state = "local"
+				} else {
+					state = "non local"
```
magik6k: `local`/`non local` sounds a bit weird to me, not sure what would sound better though.

core/commands/repo.go
```diff
@@ -406,3 +413,167 @@ var repoVersionCmd = &oldcmds.Command{
...
+				fmt.Fprintf(w, "%s\t%s\n", lo.Hash, state)
+			}
+
+			return nil
+		}),
+	},
+	Type: localityOutput{},
+}
+
+func walkBlock(ctx context.Context, dagserv dag.DAGService, merkleNode node.Node) (bool, int, int, error) {
```
magik6k: I'd just call `merkleNode` a `nd` as that's what it's usually called

magik6k: This should also probably get `recursive` or `depth` pram as it looks like that we traverse the tree even if --recursive is not specified

core/commands/repo.go
MichaelMure: done

core/commands/repo.go
```diff
@@ -36,6 +42,7 @@ var RepoCmd = &cmds.Command{
 
 	Subcommands: map[string]*cmds.Command{
 		"stat": repoStatCmd,
+		"has":  repoHasCmd,
```
magik6k: Imo this doesn't fit into `ipfs repo`. `ipfs block` or `ipfs dag` would be a better place (I'm not sure which one is better)

core/commands/repo.go
MichaelMure: No, this function is not called without `recursive`

core/commands/repo.go
MichaelMure: I'm open to suggestion ¯\_(ツ)_/¯

core/commands/repo.go
MichaelMure: I choose `ipfs repo` because we are interested about what is inside the repo and nothing else. The question is 'is this stored on my node ?'.

Both `ipfs block` and `ipfs dag` will do network request and treat the graph of object as something that exist somewhere, regardless of the repo state.

MichaelMure:
Some tests would be nice, but I'm not sure how. Is there some graph of file/directory that I could use ? The webui ?

core/commands/repo.go
frrist: I think `unknown` instead of `non local` would be appropriate as we do not have any guarantees the key exists elsewhere/remotely.

MichaelMure: But that's the point of this command, to tell if something is stored in the node, regardless of what happen elsewhere.
To illustrate, my usecase is a file sharing application. I want to know if this particular data has finished downloading or if the data the user want to share is still properly available locally.

whyrusleeping:
@MichaelMure you should just be able to add stuff in a sharness test and test the command out. You can also manually craft a dag that you don't have parts of

core/commands/repo.go
```diff
@@ -406,3 +413,190 @@ var repoVersionCmd = &oldcmds.Command{
...
+		}
+
+		recursive, _, err := req.Option("recursive").Bool()
+		if err != nil {
+			res.SetError(err, cmdkit.ErrNormal)
+			return
+		}
+
+		// Decode all the keys
+		var keys []*cid.Cid
```
Stebalien: Might as well preallocate this.

core/commands/repo.go
```diff
@@ -406,3 +413,190 @@ var repoVersionCmd = &oldcmds.Command{
...
+	QmQLXHs7K98JNQdWrBB2cQLJahPhmupbDjRuH1b9ibmwVa	incomplete	4.1 MB of 6.9 MB (60.31%)
+	> ipfs repo has -r QmeUKqHMr5r6wJzt3AVScnUAyC4TkdZGjYJNaFBCL1BmAv
+	QmeUKqHMr5r6wJzt3AVScnUAyC4TkdZGjYJNaFBCL1BmAv	complete	3.2 MB of 3.2 MB (100.00%)
+`,
+	},
+	Arguments: []cmdkit.Argument{
+		cmdkit.StringArg("key", true, true, "Key(s) to check for locality.").EnableStdin(),
+	},
+	Options: []cmdkit.Option{
+		cmdkit.BoolOption("recursive", "r", "Check recursively the graph of objects to build more precise stats"),
```
Stebalien: Should probably have a quiet option (exits with a zero status if we have everything).

core/commands/repo.go
```diff
@@ -406,3 +413,190 @@ var repoVersionCmd = &oldcmds.Command{
...
+	> ipfs repo has -r QmQFnam7qMXfF8R1D3qrETCgESnVP8gt6iEnAbyG73gyqm
+	QmQFnam7qMXfF8R1D3qrETCgESnVP8gt6iEnAbyG73gyqm	unknow
+	> ipfs repo has -r QmQLXHs7K98JNQdWrBB2cQLJahPhmupbDjRuH1b9ibmwVa
+	QmQLXHs7K98JNQdWrBB2cQLJahPhmupbDjRuH1b9ibmwVa	incomplete	4.1 MB of 6.9 MB (60.31%)
+	> ipfs repo has -r QmeUKqHMr5r6wJzt3AVScnUAyC4TkdZGjYJNaFBCL1BmAv
+	QmeUKqHMr5r6wJzt3AVScnUAyC4TkdZGjYJNaFBCL1BmAv	complete	3.2 MB of 3.2 MB (100.00%)
+`,
+	},
+	Arguments: []cmdkit.Argument{
+		cmdkit.StringArg("key", true, true, "Key(s) to check for locality.").EnableStdin(),
```
Stebalien: CID.

core/commands/repo.go
```diff
@@ -406,3 +413,190 @@ var repoVersionCmd = &oldcmds.Command{
...
+			}
+
+			return nil
+		}),
+	},
+	Type: localityOutput{},
+}
+
+func walkBlock(ctx context.Context, dagserv dag.DAGService, nd node.Node) (bool, int, int, error) {
+	stat, err := nd.Stat()
```
Stebalien: So... time to air a dirty secret... This only exists for unixfs. Don't use it. You *can't* know how much of a dag is missing (only how much of a *unixfs* filesystem is missing).

Have you considered using the `ipfs files` feature and extending the `ls` command? I think this functionality may fit better there. Also, MFS (`ipfs files`) is a *much* nicer way to work with IPFS.

Example:

```
> ipfs files cp /ipfs/QmId /some/path
> ipfs files ls --local --tree --size /some/path
/some/path
└── (1MiB/2MiB) foo
    ├── (2MiB) bar.go
    └── (2MiB) baz
```

core/commands/repo.go
Stebalien: "found", "missing"?

core/commands/repo.go
whyrusleeping: Yeah, i'm on-board with this functionality being in `ipfs files`. Especially since we should be able to do: `ipfs files ls --local --tree --size /ipfs/QmFooBar` (skipping the copy step)

Stebalien: (really, we should merge `ipfs files {ls,etc.}` and `ipfs {ls,}`.

> Especially since we should be able to do...

I agree but we can't currently. Is there an issue for this?

MichaelMure: > So... time to air a dirty secret... This only exists for unixfs. Don't use it. You can't know how much of a dag is missing (only how much of a unixfs filesystem is missing).

I'm confused. If I'm looking at the code of `Stat()`, I don't see anything specific to unixfs. This does works as well:
```
$ echo 'test' | ipfs add
added QmeomffUNfmQy76CQGy9NdmqEnnHU9soCexBnGU3ezPHVH QmeomffUNfmQy76CQGy9NdmqEnnHU9soCexBnGU3ezPHVH
$ ipfs object stat QmeomffUNfmQy76CQGy9NdmqEnnHU9soCexBnGU3ezPHVH
NumLinks: 0
BlockSize: 13
LinksSize: 2
DataSize: 11
CumulativeSize: 13
```
What am I missing ?
Do you mean that `Stat()` is not implemented properly for every kind of ipld block ? If so, why ?

MichaelMure: More questions:
- `ipfs files` is supposed to be MFS, right ? So if `ipfs repo has` is moved to `ipfs files ls --local --tree --size`, it will only works inside MFS. I'm not using mfs for my project so it's not helpful for me.
- what would the `--size` parameter really mean ? Using it alone would intuitively means 'give me various size information you have'. But adding `--local` would somehow change it's meaning to 'check that everything is local properly and report how much if not' ?

Moving the feature to 'ipfs files ls' feels like shoe-horning to me. I think there is space for a proper 'node administrator' command to inspect what is inside a repository (and in my case help me drive ipfs properly).

Stebalien: > Do you mean that Stat() is not implemented properly for every kind of ipld block ? If so, why ?

Yes. To be precise, `Stat`, etc. only apply to DagPB IPLD nodes (so, technically not *only* unixfs). One thing on my todo list is to fix the go interfaces and get rid of that method.

Why? IPLD came after unixfs and the merkledag (DagPB). We wanted to generalize the concept and create a system that:

1. Allowed *structured* linked data. DagPB does not allow much structure. We wanted something more like JSON.
2. Allowed us to interpret and treat all existing existing merkledags (like Ethereum and git) as one big merkleforest. That way, we could write one set of tools to work with any merkle-linked data.

That means we couldn't require any metadata fields like `Size`. We wouldn't have, e.g., been able to interpret a git repo as an IPLD dag if we did that.

---

> I'm not using mfs for my project so it's not helpful for me.

Aside: IMO, you should. It makes managing files much simpler. Eventually, we'd like to using MFS for *all* local file management (where `/ipfs` and `/ipns` would be "virtual" directories within the MFS tree. Eventually, `ipfs CMD` will probably be aliases for `ipfs files CMD` (possibly with some path restrictions).

However, for now, it *would* probably be simpler to just add the flags to the `ipfs ls` command (we can work on unifying the commands later). That should work for your purpose.

> what would the --size parameter really mean ? Using it alone would intuitively means 'give me various size information you have'. But adding --local would somehow change it's meaning to 'check that everything is local properly and report how much if not' ?

Sorry, that was poorly explained.

* `--size` would mean "report the size of the file and how much has been downloaded so far". We could also just add a separate `--local-size` flag for that.
* `--local` would mean "don't fetch anything when running this command". Ideally, we'd provide that as an argument to *all* commands. Otherwise, `ls --tree` would block until all directories (but not their files) have been fetched.

> I think there is space for a proper 'node administrator' command to inspect what is inside a repository (and in my case help me drive ipfs properly).

It would be useful to have a command to check and see if we have locally something without fetching it. However, we *can't* make a general purpose command for getting size information about arbitrary IPLD nodes.

MichaelMure: Thank you, that's helpful.

> Aside: IMO, you should. It makes managing files much simpler. Eventually, we'd like to using MFS for all local file management (where /ipfs and /ipns would be "virtual" directories within the MFS tree. Eventually, ipfs CMD will probably be aliases for ipfs files CMD (possibly with some path restrictions).

For some context, my project is https://github.com/MichaelMure/Arbore. It's a social file sharing app. For a broader view on how I use IPFS, see https://github.com/ipfs/go-ipfs/issues/3866

Unless I miss something (entirely possible), I see MFS as more a problem than a help for me. I'm using the IPFS repository as a black box / buffer area. In particular, I need to:
- import easily files/directory (`ipfs add -r`). I get a nice progress from the API. I don't understand how to do that with MFS. On top of that I would need to manage path inside of MFS where I really don't need to.
- be able to trigger a download (`ipfs pin add`). MFS doesn't seems to support that (?). Each share in Arbore has it's own root CID, so it's very easy to manage. If I want to remove a share from IPFS, I just `pin rm` the root, trigger a garbage collect and IPFS will deal with everything.
- be able to use the filestore. Not using twice the disk size when you handle big files is a nice touch. I don't think MFS is able to do that.
- be able to encrypt files. Ideally it will be done by IPFS when encryption support is added. Will that be supported by MFS ?
---
> It would be useful to have a command to check and see if we have locally something without fetching it. However, we can't make a general purpose command for getting size information about arbitrary IPLD nodes.

What about a command that return detailed size information only if the nodes are compatible ? If the cumulative size is not available, we are still able to tell if everything is local or not, and how much data is here (using `Node.Size()` recursively). 

---

So, unless I got it wrong, MFS is not an option for me. I see two viable options for me:
- move the feature to `ipfs ls` 
- fix the current PR for `ipfs repo has` to not use `Node.Stat()`

@Stebalien @whyrusleeping Thoughts ?

whyrusleeping: @MichaelMure 

> - import easily files/directory (`ipfs add -r`). I get a nice progress from the API. I don't understand how to do that with MFS. On top of that I would need to manage path inside of MFS where I really don't need to.

Keep using `ipfs add`, Its really supposed to just be an alias for `ipfs files add` but that command doesnt exist yet (see the javascript api).

> - be able to trigger a download (`ipfs pin add`). MFS doesn't seems to support that (?). Each share in Arbore has it's own root CID, so it's very easy to manage. If I want to remove a share from IPFS, I just `pin rm` the root, trigger a garbage collect and IPFS will deal with everything.

For now, keep that flow. We need to add a command to `ipfs files` that lets you fully sync a given bit of content. Like `ipfs files cp —full-sync /ipfs/QmFooBar /path/to/thing` 

> - be able to use the filestore. Not using twice the disk size when you handle big files is a nice touch. I don't think MFS is able to do that.

Keep using the filestore, ipfs files doesnt care where the blocks are.

> - be able to encrypt files. Ideally it will be done by IPFS when encryption support is added. Will that be supported by MFS ?

Yeah, It really won't get in your way at all for any of these concerns.



To be clear, what we're recommending (and as I said in my previous comment) the command would look like:

`ipfs files ls —local —tree —size /ipfs/QmFooBar`

Note that the content in question is *not* explicitly within the mfs namespace.

whyrusleeping:
@Stebalien Heres a proposal:

```
USAGE
  ipfs files tree <path>... - Show information about a given subtree

SYNOPSIS
  ipfs files tree [--summary | --local] [--] <key>...

ARGUMENTS

  <path>... - Path to print information about.

OPTIONS

  --summary bool - Print a summary of tree's information.
  --local bool - Provide information about locality of data.
```

Stebalien:
@whyrusleeping :+1: for not overloading the `ls` command.

What exactly would the summary include? Would the tree *always* include file sizes, types, etc?

eingenito:
I think this PR was superseded by #4638. Closing and feel free to reopen if I got that wrong.

PR #4550, opened 2018-01-05
Feat: ipfs repo size command

hsanjuan:
ipfs repo size just display the repository size without needing to
count the blocks, which takes a significant amount of the repo stat
time.

ipfs-cluster queries the repository size to allocate pins to those
peers with most free space using repo/stat. When done too often on a busy
node, this has resulted in downgraded performance. Counting
blocks is an information that cluster does not use.

License: MIT

### DIFF
--- a/core/commands/repo.go
+++ b/core/commands/repo.go
@@ -36,6 +36,7 @@ var RepoCmd = &cmds.Command{
 
 	Subcommands: map[string]*cmds.Command{
 		"stat": repoStatCmd,
+		"size": repoSizeCmd,
 	},
 	OldSubcommands: map[string]*oldcmds.Command{
 		"gc":      repoGcCmd,
@@ -157,6 +158,7 @@ set of stored objects and print repo statistics. It outputs to stdout:
 NumObjects      int Number of objects in the local repo.
 RepoPath        string The path to the repo being currently used.
 RepoSize        int Size in bytes that the repo is currently taking.
+StorageMax      string Maximum datastore size (from configuration)
 Version         string The repo version.
 `,
 	},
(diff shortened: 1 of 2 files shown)


hsanjuan:
This is still relatively slow so I'm not 100% sure it fixes the problem, responses do show a ~40% improvement.

Any of you guys can think of a better way to keeping track of the repo size? Would it be doable to calculate it once and then update the value on every write/gc ?

whyrusleeping:
Maybe we should just add this as a flag to repo stat?

Stebalien:
> Any of you guys can think of a better way to keeping track of the repo size? Would it be doable to calculate it once and then update the value on every write/gc?

I think we should leave it up to the datastore. That is, ask the datastore its size and let it do whatever it needs. For flatfs, we should keep a running count. If you think the performance sucks now, it's going to absolutely *tank* when we upgrade to a kernel with meltdown protections enabled (syscall performance will tank, we're currently doing a `du` (effectively)). For badger, we can just *ask* badger (it provides a `Size` function).

lgierth:
Would it be an option at all to keep track of repo size separately? i.e. start at zero, then increment for every added block, decrement for every deleted block? That'd make repo size an accounting function and not an IO operation anymore.

I don't know how Badger or LevelDB keep track of size, they're probably smarter than flatfs. In that case accounting would still be an option for flatfs?

hsanjuan:
> I think we should leave it up to the datastore. That is, ask the datastore its size and let it do whatever it needs. For flatfs, we should keep a running count.

This sounds like a good way to do. I can try do this. Do you have any pointers? (There's an interface and datastores are implemented in different repos right ?) What datastores are available now?

> Maybe we should just add this as a flag to repo stat?

Sure, it's cosmetic so I can do it at the end (along with tests)

kevina:
Although this might brake backwards compatibility I would lean towards having the stat command only return what is is already computed (much like how the stat system call works on unix) and have *this* command do the actual computation.

hsanjuan:
See #5010

PR #4552, opened 2018-01-06
Fix typos and inconsistencies in commands documentation

eocarragain:
A few minor corrections I noticed when reading through https://ipfs.io/docs/commands/.

In the case of name and publish, the example is currently confusing because the output shows the ipfs path being published to the same ipns as earlier despite the fact it was published with a new key. 

Also, the generated public page on ipfs.io is a bit out of date (April 2017), so some old commands are still there despite being removed from the latest versions, e.g. "ipfs tour"

### DIFF
--- a/core/commands/files/files.go
+++ b/core/commands/files/files.go
@@ -749,8 +749,8 @@ NOTE: All paths must be absolute.
 
 Examples:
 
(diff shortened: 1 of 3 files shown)


whyrusleeping:
LGTM, thanks!

PR #4553, opened 2018-01-06
implement an ipfs cat --length flag

### LINKED ISSUE #4085: Partial Read / Seeking without downloading the complete file

Stebalien:
fixes #4085

(because we too many open bugs dammit)


test/sharness/t0040-add-and-cat.sh
```diff
@@ -78,6 +78,56 @@ test_add_cat_file() {
...
+    test_cmp expected actual
+  '
+
+  test_expect_success "ipfs cat multiple hashes with offset and length succeeds" '
+    ipfs cat --offset 5 --length 15 "$HASH" "$HASH" "$HASH" >actual
+  '
+
+  test_expect_success "ipfs cat multiple hashes with offset and length looks good" '
+    echo " Worlds!" >expected &&
+    echo -n "Hello " >>expected &&
```
kevina: I would use `printf "Worlds!\nHello '"`.

test/sharness/t0040-add-and-cat.sh
```diff
@@ -78,6 +78,56 @@ test_add_cat_file() {
     test_expect_code 1 ipfs cat --offset -102 "$HASH" > actual
   '
 
+  test_expect_success "ipfs cat with length succeeds" '
+    ipfs cat --length 8 "$HASH" >actual
+  '
+
+  test_expect_success "ipfs cat with length output looks good" '
+    echo -n "Hello Wo" >expected &&
```
kevina: Note the `-n` flag is not universally supported by all shells (for example on the mac).  It is better to use `printf` which is more portable.  In this case it will be
```
  printf "Hello Wo" >expected &&
```

test/sharness/t0040-add-and-cat.sh
kevina: Note that is you try this in bash on the command line this will fail as '!' has special meaning in bash when history expansion in enabled.  This is disabled by default in scripts so it won't cause a problem.

Stebalien:
@kevina the code climate issue is a commands error style issue (and is significantly more pervasive). The error I added mimics the existing error so I'd like to leave it that way for now. We should go though these errors and fix them all at once at some point.

kevina:
@Stebalien that doesn't bother me if @whyrusleeping is fine with it also.

core/commands/cat.go
```diff
@@ -50,7 +51,20 @@ var CatCmd = &cmds.Command{
 			return
 		}
 
-		readers, length, err := cat(req.Context(), node, req.Arguments(), int64(offset))
+		max, found, err := req.Option("length").Int()
```
whyrusleeping: USE int64. THIS COULD OVERFLOW

frrist: I don't see a int64 option in go-ipfs-cmdkit, we could use a uint instead? We could also add a new option if not.

whyrusleeping: I was really just giving @Stebalien crap, he was sitting behind me while i reviewed this and pointed out that this should be an int64

PR #4562, opened 2018-01-08
fix defered unlock of pin lock in AddR

Stebalien:
This wasn't causing any problems because this function is actually dead code (but possibly useful?).

Should we just remove this function?


kevina:
I would say to remove it, but it is exported so it could conceivable be used by another package outside of go-ipfs.  I think we have a lot of dead code like this.

@Kubuxu (or anyone else) is there a way to detect all unused exported functions in go-ipfs?

PR #4563, opened 2018-01-08
commands/block: use CIDv1 with custom mhtype

magik6k:
Closes https://github.com/ipfs/go-ipfs/issues/4741 ~Kubuxu

### DIFF
--- a/core/commands/block.go
+++ b/core/commands/block.go
@@ -132,7 +132,7 @@ than 'sha2-256' or format to anything other than 'v0' will result in CIDv1.
 		cmdkit.FileArg("data", true, false, "The data to be stored as an IPFS block.").EnableStdin(),
 	},
 	Options: []cmdkit.Option{
-		cmdkit.StringOption("format", "f", "cid format for blocks to be created with.").WithDefault(""),
+		cmdkit.StringOption("format", "f", "cid format for blocks to be created with."),
 		cmdkit.StringOption("mhtype", "multihash hash function").WithDefault("sha2-256"),
 		cmdkit.IntOption("mhlen", "multihash hash length").WithDefault(-1),
 	},
@@ -172,8 +172,8 @@ than 'sha2-256' or format to anything other than 'v0' will result in CIDv1.
 		var pref cid.Prefix
 		pref.Version = 1
 
-		format := req.Options["format"].(string)
-		if format == "" {
+		format, formatSet := req.Options["format"].(string)
+		if !formatSet {
 			if mhtval == mh.SHA2_256 {
 				format = "v0"
 			} else {
@@ -186,7 +186,7 @@ than 'sha2-256' or format to anything other than 'v0' will result in CIDv1.
 		}
 		formatval, ok := cid.Codecs[format]
 		if !ok {
-			res.SetError(fmt.Errorf("unrecognized format: %s", format), cmdkit.ErrNormal)
+			res.SetError(fmt.Errorf("unrecognized format: '%s'", format), cmdkit.ErrNormal)
 			return
 		}
 		if mhtval != mh.SHA2_256 && pref.Version == 0 {
--- a/core/commands/block.go
+++ b/core/commands/block.go
@@ -3,6 +3,7 @@ package commands
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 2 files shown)


Stebalien:
Shouldn't this actually just return an error if the user passes `-f v0 --mhtype=other-hash --mhlen=other-len`?

core/commands/block.go
```diff
@@ -157,27 +159,32 @@ It reads from stdin, and <key> is a base58 encoded multihash.
...
+			err := fmt.Errorf("unrecognized multihash function: %s", mhtype)
+			res.SetError(err, cmdkit.ErrNormal)
+			return
+		}
+
 		var pref cid.Prefix
 		pref.Version = 1
 
-		format, _ := req.Options["format"].(string)
+		format, userFormat := req.Options["format"].(string)
```
Stebalien: Unfortunately, `userFormat` will always be true there is a default value. The only way I know to fix this, is to simply remove the default.

test/sharness/t0050-block.sh
```diff
@@ -209,4 +209,16 @@ test_expect_success "no panic in output" '
   test_expect_code 1 grep "panic" stat_out
 '
 
+test_expect_success "can set multihash type and length on block put without format" '
+  HASH=$(echo "foooo" | ipfs block put --mhtype=sha3 --mhlen=16)
```
Kubuxu: 16 will be too short of a hash length after cid changes come through, please change it to 20.

magik6k: done

core/commands/block.go
```diff
@@ -121,14 +122,17 @@ var blockPutCmd = &cmds.Command{
...
+than 'sha2-256' or format to anything other than 'v0' will result in CIDv1.
 `,
 	},
 
 	Arguments: []cmdkit.Argument{
 		cmdkit.FileArg("data", true, false, "The data to be stored as an IPFS block.").EnableStdin(),
 	},
 	Options: []cmdkit.Option{
-		cmdkit.StringOption("format", "f", "cid format for blocks to be created with.").WithDefault("v0"),
+		cmdkit.StringOption("format", "f", "cid format for blocks to be created with.").WithDefault(""),
```
Stebalien: I wouldn't specify a default at all. As it is, passing `-f ""` will be considered valid even though it makes no sense.

If you don't specify a default, you can tell if the parameter was passed with `format, formatSet := req.Options["format"].(string)`.

schomatis:
This PR looks correct to me (and it fixes the addressed issue) but let me raise an off-topic question: shouldn't part of this prefix preprocessing logic be in [`go-cid`](https://github.com/ipfs/go-cid#creating-a-cid-from-scratch)?

It seems to me there are a couple of hard-coded pieces of information of what constitutes a valid CID prefix that the `ipfs block` command doesn't need to know because it shouldn't be its responsibility to check for it, e.g., the (now) default CIDv1 specified here (`pref.Version = 1`) will have to be updated if (or when) a new CID version is released.

The `ipfs block` command doesn't do anything with the version and hash function except forward them to `Prefix`, which is the one that computes the hash/CID that `ipfs block` is really interested in (to create block).

Kubuxu:
I think most of the processing here is due to 1. us wanting to provide good error messages for user 2. be lenient with the user input. One check in question (CIDv0 can only be full length SHA256) is currently not implemented but we are in process of changing that (https://github.com/ipfs/go-cid/pull/40/files#diff-beb384810133d725af6b502e9b6e3de8R139)

I am not sure if go-cid would be right place to cover all of those areas but having a helper function somewhere in go-ipfs would not hurt.

schomatis:
> us wanting to provide good error messages for user

That's a good point I hadn't considered, but looking at the errors:

* `unrecognized multihash function`
* `unrecognized format`
* `cannot generate CIDv0 with non-sha256 hash function`

I think all of them can be outsourced without losing expressiveness for the user (the `missing option "mhlen"` seems to me the only error that truly belongs here).


> I am not sure if go-cid would be right place to cover all of those areas but having a helper function somewhere in go-ipfs would not hurt.

That's fine, my main point wasn't where this logic should be placed but that it should be encapsulated somewhere outside this command function.

Anyway, maybe this should be discussed in another issue after this PR is merged, trying to change all this here in this PR now would be to messy, but I wanted to leave this suggestion documented somewhere.

Stebalien:
I think the issue here is that we want to catch all of these errors early *and* provide reasonable defaults (that may change based on the command in question). However, we probably could have something in `go-cid` that takes in a set of CID "constraints" and spits out either an error or the set of CID parameters.

An alternative solution would be to add flag "sets" to go-ipfs-cmds and create a flag set for CID options somewhere.

whyrusleeping:
First PR merged of the 0.4.15 cycle goes to @magik6k :) 👏

PR #4564, opened 2018-01-08
fix circleci cache directory list

Stebalien:
environment variable expansion doesn't work in the cache_directories section.


Stebalien:
I'm not sure this is actually necessary. CircleCI already caches `/home/ubuntu/.go_workspace`. However, it *was* wrong before. We can probably just remove this line.

PR #4566, opened 2018-01-09
Fix/namesys validation

dirkmc:
Update go-ipfs to use new validation function format in libp2p/go-libp2p-record#9


vyzo:
Just a heads up -- @Stebalien (rightfully) does like the accessor functions in the new record:
https://github.com/libp2p/go-libp2p-record/pull/11

dirkmc:
Ok I will fix :)

vyzo:
He is also keen on splitting off the namespace: https://github.com/libp2p/go-libp2p-record/pull/12

dirkmc:
This PR is getting messy, moving to https://github.com/ipfs/go-ipfs/pull/4597

PR #4568, opened 2018-01-09
Update to cmds lib 1.0

whyrusleeping:


core/commands/add.go
```diff
@@ -180,26 +182,35 @@ You can now check what blocks have been created by:
...
+		hashFunStr, hfset := req.Options[hashOptionName].(string)
 
 		if nocopy && !cfg.Experimental.FilestoreEnabled {
 			res.SetError(errors.New("filestore is not enabled, see https://git.io/vy4XN"),
 				cmdkit.ErrClient)
 			return
 		}
 
+		if hfset && hashFunStr != "sha2-256" && cidVer == 0 {
+			cidVer = 1
```
Stebalien: This should be:

```go
if hfset && hashFunStr != "sha2-256" && cidVer == 0 {
    if cidVerSet {
        res.SetError(...)
        return
    }
    cidVer = 1
}
```

core/commands/add.go
```diff
@@ -180,26 +182,35 @@ You can now check what blocks have been created by:
...
+		}
+
+		if cidVer > 0 && !rbset {
+			rawblks = true
```
Stebalien: Should also have the following check:

```go
if cidVer == 0 && rawblks {
    res.SetError(...)
    return
}
```

core/commands/env.go
```diff
@@ -0,0 +1,27 @@
...
+
+	"github.com/ipfs/go-ipfs/commands"
+	"github.com/ipfs/go-ipfs/core"
+	"github.com/ipfs/go-ipfs/repo/config"
+)
+
+func GetNode(env interface{}) (*core.IpfsNode, error) {
+	ctx, ok := env.(*commands.Context)
+	if !ok {
+		return nil, fmt.Errorf("expected env to be of type %T, got %T", ctx, env)
```
Stebalien: Should just panic. This is a programmer error.

core/commands/env.go
```diff
@@ -0,0 +1,27 @@
...
+	}
+
+	return ctx.GetNode()
+}
+
+func GetConfig(env interface{}) (*config.Config, error) {
+	ctx, ok := env.(*commands.Context)
+	if !ok {
+		return nil, fmt.Errorf("expected env to be of type %T, got %T", ctx, env)
```
Stebalien: Same here. Panic.

core/commands/files/files.go
```diff
@@ -235,6 +236,7 @@ var FilesCpCmd = &cmds.Command{
 		}
 
 		flush, _, _ := req.Option("flush").Bool()
+		fmt.Println("flush:", flush)
```
Stebalien: Debugging statement?

core/commands/filestore.go
```diff
@@ -54,15 +53,15 @@ The output is:
 	Options: []cmdkit.Option{
 		cmdkit.BoolOption("file-order", "sort the results based on the path of the backing file"),
 	},
-	Run: func(req cmds.Request, res cmds.ResponseEmitter) {
-		_, fs, err := getFilestore(req.InvocContext())
+	Run: func(req *cmds.Request, res cmds.ResponseEmitter, env cmds.Environment) {
+		_, fs, err := getFilestore(env.(*oldCmds.Context))
```
Stebalien: Unnecessary cast.

core/commands/get.go
```diff
@@ -119,15 +119,19 @@ may also specify the level of compression by specifying '-l=<1-9>'.
 					return
 				}
 
-				outPath := getOutPath(req)
+				outPath, _ := req.Options["output"].(string)
```
Stebalien: Why change this? (Also, we *do* need to strip the slash, IIRC (causes a bug)).

core/commands/get.go
```diff
@@ -251,15 +255,15 @@ func (gw *getWriter) writeExtracted(r io.Reader, fpath string) error {
...
-	cmprs, _, _ := req.Option("compress").Bool()
-	cmplvl, cmplvlFound, _ := req.Option("compression-level").Int()
+func getCompressOptions(req *cmds.Request) (int, error) {
+	cmprs, _ := req.Options["compress"].(bool)
+	cmplvl, _ := req.Options["compression-level"].(int)
 	switch {
 	case !cmprs:
 		return gzip.NoCompression, nil
-	case cmprs && !cmplvlFound:
+	case cmprs && cmplvl == -1:
```
Stebalien: Why? Why not just use `cmplvlFound`? We should return an error if is `cmplvl < 0`.

test/sharness/t0040-add-and-cat.sh
```diff
@@ -505,7 +505,7 @@ test_add_cat_expensive "--cid-version=1" "zdj7WcatQrtuE4WMkS4XsfsMixuQN2po4irkYh
 # encoded with the blake2b-256 hash funtion
 test_add_cat_expensive '--hash=blake2b-256' "zDMZof1kwndounDzQCANUHjiE3zt1mPEgx7RE3JTHoZrRRa79xcv"
 
-test_add_named_pipe " Post http://$API_ADDR/api/v0/add?encoding=json&progress=true&r=true&stream-channels=true:"
```
Stebalien: This worries me. Will this break anything?

djdv:
Pedantic:
Some type assertions ignore the optional 'ok' return like this `x, _ := ...` and others like this `x := ...`.

Stebalien:
@djdv where? As far as I can tell, we always use checked cast (although we sometimes ignore the error). We *have* to do this because the option can be nil.

cmd/ipfs/daemon.go
```diff
@@ -192,22 +192,23 @@ func daemonFunc(req cmds.Request, re cmds.ResponseEmitter) {
...
-	if managed, _, _ := req.Option(adjustFDLimitKwd).Bool(); managed {
+	managefd, _ := req.Options[adjustFDLimitKwd].(bool)
+	if managefd {
 		if err := utilmain.ManageFdLimit(); err != nil {
 			log.Errorf("setting file descriptor limit: %s", err)
 		}
 	}
 
-	ctx := req.InvocContext()
+	cctx := env.(*oldcmds.Context)
```
djdv: @Stebalien 
This is one if I'm not mistaken.

Stebalien: Ah. I was focusing on the `Options` casts. In this case, we *do* want to panic. It's a programmer error. In the `Options` cases, we don't (it just means that the option wasn't passed (or the option has the wrong type but checking that would be a pain and I'd like to fix that with something like https://github.com/ipfs/go-ipfs-cmds/issues/50).

However, you're right. We do occasionally return an error instead of panicing when casting `env`. We should always panic.

core/commands/add.go
```diff
@@ -120,32 +121,33 @@ You can now check what blocks have been created by:
...
 
-		sizeFile, ok := req.Files().(files.SizeFile)
+		sizeFile, ok := req.Files.(files.SizeFile)
 		if !ok {
 			// we don't need to error, the progress bar just won't know how big the files are
 			log.Warning("cannot determine size of input file")
 			return nil
 		}
 
+		// HACK! Using context to pass the size to PostRun
```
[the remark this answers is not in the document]
Stebalien: Fixed.

Stebalien:
Something appears to be wring with recursive file adding.

core/commands/cat.go
```diff
@@ -40,24 +40,27 @@ var CatCmd = &cmds.Command{
...
 		if offset < 0 {
 			res.SetError(fmt.Errorf("Cannot specify negative offset."), cmdkit.ErrNormal)
 			return
 		}
 
-		readers, length, err := cat(req.Context(), node, req.Arguments(), int64(offset))
+		err = req.ParseBodyArgs()
+		if err != nil && !cmds.IsAllArgsAlreadyCovered(err) {
+			res.SetError(err, cmdkit.ErrNormal)
+			return
```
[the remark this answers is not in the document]
Stebalien: @whyrusleeping do you have any idea what's going on here?

test/sharness/t0040-add-and-cat.sh
Stebalien: @whyrusleeping thoughts?

cmd/ipfs/main.go
```diff
@@ -89,316 +89,132 @@ func mainRet() int {
...
-			// we need a newline space.
-			fmt.Fprintf(os.Stderr, "\n")
-			printHelp(false, os.Stderr)
-		}
-		return 1
-	}
+				// ok everything is good. set it on the invocation (for ownership)
+				// and return it.
+				n, err = core.NewNode(ctx, &core.BuildCfg{
+					// TODO(keks) figure out how Online was set before. I think it was set to
```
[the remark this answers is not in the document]
Stebalien: @whyrusleeping ?

cmd/ipfs/main.go
```diff
@@ -89,316 +89,132 @@ func mainRet() int {
...
-	debug, _, err := i.req.Option("debug").Bool()
-	if err != nil {
-		return err
-	}
+	debug, _ := req.Options["debug"].(bool)
 	if debug || os.Getenv("IPFS_LOGGING") == "debug" {
 		u.Debug = true
 		logging.SetDebugLogging()
 	}
 	if u.GetenvBool("DEBUG") {
```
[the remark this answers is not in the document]
Stebalien: Meh. Not new...

commands/request.go
```diff
@@ -63,6 +65,48 @@ func (c *Context) NodeWithoutConstructing() *core.IpfsNode {
 	return c.node
 }
 
+// RootContext returns the node's context.
+func (c *Context) Context() context.Context {
+	n, err := c.GetNode()
+	if err != nil {
+		log.Debug("error getting node: ", err)
+		return nil
```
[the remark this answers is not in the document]
Stebalien: fixed.

cmd/ipfs/main.go
```diff
@@ -89,316 +89,132 @@ func mainRet() int {
...
-	}
+	intrh, ctx := setupInterruptHandler(ctx)
+	defer intrh.Close()
 
 	// Handle `ipfs help'
 	if len(os.Args) == 2 {
 		if os.Args[1] == "help" {
-			printHelp(false, os.Stdout)
-			return 0
+			os.Args[1] = "-h"
```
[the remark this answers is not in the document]
Stebalien: Not new... fix later.

cmd/ipfs/main.go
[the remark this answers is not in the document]
whyrusleeping: mmmm.... yeah. We should figure this one out.

core/commands/add.go
whyrusleeping: Hrm... thats a difficult thing now. 

We don't have an easy way to distinguish between default values and unset values.

Stebalien: We can just not set a default value. However, we should probably just do this in a follow up PR.

Stebalien: I just went ahead and fixed this anyways.

core/commands/add.go
Stebalien: Fixed.

core/commands/cat.go
```diff
@@ -64,14 +61,20 @@ var CatCmd = &cmds.Command{
 			max = -1
 		}
 
-		readers, length, err := cat(req.Context(), node, req.Arguments(), int64(offset), int64(max))
+		err = req.ParseBodyArgs()
```
Stebalien: This dance still annoys me. `Arguments` used to handle this by default but now it doesn't. @whyrusleeping you have any idea why @keks made this change?

cmd/ipfs/main.go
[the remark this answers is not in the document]
keks: Yes, there is. os.Args is simply passed to the cli parser later on. So instead of modifying os.Args we can copy this to a new slice and pass that one to the parser. I changed os.Args because I was afraid something else might later read os.Args and then we'd have inconsistencies.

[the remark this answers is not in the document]
keks: an I actually think it is new

cmd/ipfs/main.go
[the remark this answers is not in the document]
keks: I don't think Online *was* set before, or rather it was set to an uninitialized (thus zero value) variable. But maybe I just couldn't find the place where it was set.

cmd/ipfs/main.go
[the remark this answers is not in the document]
keks: yeah that one was weird but I didn't want to break anything...

cmd/ipfs/main.go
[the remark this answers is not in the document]
Stebalien: Sorry, by not new I mean that we do this elsewhere as well (might as well fix all instances at once).

test/sharness/t0040-add-and-cat.sh
keks: I checked back with why before changing this. The point is simply that when go-ipfs-cmds/cli fills in the default values in the request, and this request is passed on to the client in go-ipfs-cmds/http, which then makes a request that contains the default value for unset options. That's why they show up in the error message. Making requests without explicitly stating options with defaults still works, but we just don't happen to do it this way. This does not break compat to other clients.

keks: but yeah we might add a test for that here, sorry

core/commands/cat.go
keks: It's been a while so I can't really remember, but I think it was because @Kubuxu said he wants to be able to read the body args in a streaming fashion.

whyrusleeping: After a short conversation with @keks, we remembered its because we want to be able to stream input arguments. In which case this still kinda doesnt let us do what we want. Lets try and figure this out...

keks: FIRST!!!1

keks: I think the streaming was implement neither before nor now, but if we always parse on request arrival time we can't optionally stream. So if we did that we'd have to break the API again when that is added.
Right now, nothing uses streamed parsing (IIRC), so it can be added later.

test/sharness/t0040-add-and-cat.sh
Stebalien: Np. When I saw this change, I thought we might be changing the flags. However, this is totally reasonable.

core/commands/cat.go
Stebalien: Hm. Ideally, I wonder if we could just make arguments a channel. However, given that this will be a large change, IMO, we should merge and keep this as a TODO (its an internal interface and not a bug).

core/corehttp/commands.go
```diff
@@ -51,9 +51,6 @@ func addCORSFromEnv(c *cmdsHttp.ServerConfig) {
 	origin := os.Getenv(originEnvKey)
 	if origin != "" {
 		log.Warning(originEnvKeyDeprecate)
-		if len(c.AllowedOrigins()) == 0 {
-			c.SetAllowedOrigins([]string{origin}...)
-		}
```
Stebalien: @whyrusleeping just double checking... this wasn't intentional, was it?

Stebalien:
The tests are going to pass...

core/commands/cat.go
keks: I'm not sure channels are the right tool here, instead I'd propose something like `func (*Request) ParseNextArg() (Argument, error)` or so. With channels it's really weird to access the parse error.

Stebalien: Hm. Good point (although we could have the channel return errors). My primary goal is to make this a single API. How about:

```go
// Collect and return all arguments as a single slice. Most commands will just use this.
// Calling this again will return an error.
func (*Request) CollectArgs() ([]Argument, error)
// returns the next n (at most) arguments, or an error if parsing any fails.
func (*Request) NextArgs(n int) ([]Argument, error)
// Parse and return the next arg.
func (*Request) NextArg() (Argument, error)
```

Thoughts?

keks:
I think if collectarguments returned nil the first Rome, the result should be stored and returned on future calls. Other than that, lgtm. (sorry, mobile)

Am 24. Januar 2018 19:55:14 MEZ schrieb Steven Allen <notifications@github.com>:

core/corehttp/commands.go
whyrusleeping: No, something about that seems wrong.

whyrusleeping:
LGTM, tests pass on OSX locally.

PR #4569, opened 2018-01-09
Added sharding to experimental features

VictorBjelkholm:
Not sure about "Road to being a real feature" and also would like to have clarified when this is being used. My understanding is that it'll shard automatically both when using the Files API and also when doing `ipfs add -r` but I'm not 100% about this.

### DIFF
--- a/docs/experimental-features.md
+++ b/docs/experimental-features.md
@@ -343,3 +343,25 @@ See [Plugin docs](./plugins.md)
 
 - [ ] Needs more testing
 - [ ] Make sure there are no unknown major problems
+
+## Directory Sharding / HAMT
+
+### In Version
+0.4.8
+
+### State
+Experimental
+
+Allows to create directories with unlimited number of entries - currently
+size of unixfs directories is limited by the maximum block size
+
+### Basic Usage:
+
+```
+ipfs config --json Experimental.ShardingEnabled true
+```
+
+### Road to being a real feature
+
+- [ ] Make sure that objects that don't have to be sharded aren't
+- [ ] Generalize sharding and define a new layer between IPLD and IPFS


docs/experimental-features.md
```diff
@@ -343,3 +343,23 @@ See [Plugin docs](./plugins.md)
...
+
+Currently, when too many items are added into a unixfs directory, the object gets too large and you may experience issues. To pervent this problem, and generally make working really large directories more efficient, we have implemented a HAMT structure for unixfs. To enable this feature, run:
+
+### Basic Usage:
+
+```
+ipfs config --json Experimental.ShardingEnabled true
+```
+
+### Road to being a real feature
```
magik6k: IIRC this was a problem:
- [ ] Make sure that objects that don't have to be sharded aren't

docs/experimental-features.md
```diff
@@ -343,3 +343,23 @@ See [Plugin docs](./plugins.md)
...
+
+## Sharding / HAMT
+
+### In Version
+0.4.8
+
+### State
+Experimental
+
+Currently, when too many items are added into a unixfs directory, the object gets too large and you may experience issues. To pervent this problem, and generally make working really large directories more efficient, we have implemented a HAMT structure for unixfs. To enable this feature, run:
```
magik6k: I'd keep the description short, something like: `Allows to create directories with unlimited number of entries - currently size of unixfs directories is limited by the maximum block size.`

Also, try to keep the lines in docs under 80 characters.

VictorBjelkholm:
Thanks for the feedback @Stebalien and @magik6k, I've updated the document now.

VictorBjelkholm:
Could someone clarify if the sharding always happens or only when using `ipfs add-r`/using MFS?

whyrusleeping:
@VictorBjelkholm when sharding is enabled, everything dealing with unixfs directories will use it.

VictorBjelkholm:
@whyrusleeping thanks, then this is ready to be merged.

PR #4575, opened 2018-01-10
coreapi: Pin API

magik6k:
TODO:
- [x] Implement
- [ ] Review
- [ ] Docs
- [x] Tests

### DIFF
--- a/core/coreapi/coreapi.go
+++ b/core/coreapi/coreapi.go
@@ -37,6 +37,10 @@ func (api *CoreAPI) Key() coreiface.KeyAPI {
 	return &KeyAPI{api, nil}
 }
 
+func (api *CoreAPI) Pin() coreiface.PinAPI {
+	return &PinAPI{api, nil}
+}
+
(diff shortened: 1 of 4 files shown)


core/coreapi/pin_test.go
```diff
@@ -0,0 +1,209 @@
...
+	if err != nil {
+		t.Error(err)
+	}
+
+	p, err := api.Unixfs().Add(ctx, strings.NewReader("foo"))
+	if err != nil {
+		t.Error(err)
+	}
+
+	err = api.Pin().Add(ctx, p)
```
frrist: Should be verify the content was pined? e.g. `list, err = api.Pin().Ls(ctx) && len(list) == 1` as is done elsewhere?

magik6k: This is more of an example rather than a test, since this is tested a few times below I skipped that check here.

magik6k:
Rebased

PR #4578, opened 2018-01-13
Make repo gc call CollectGarbage on datastore

magik6k:
Depends on https://github.com/ipfs/go-ds-measure/pull/13

I'm not sure how to approach adding tests for this. Badger does implement this interface and it does get called, though due to it's GC weirdness I'm unable to come up with scenario in which it will actually free-up some space.

(Badger was updated to current master, doesn't seem to break anything from my testing)

### DIFF
--- a/pin/gc/gc.go
+++ b/pin/gc/gc.go
@@ -11,6 +11,7 @@ import (
 	dag "github.com/ipfs/go-ipfs/merkledag"
 	pin "github.com/ipfs/go-ipfs/pin"
 
+	dstore "gx/ipfs/QmPpegoMqhAEqjncrzArm7KVWAkCm78rqL2DPuNjhPrshg/go-datastore"
 	logging "gx/ipfs/QmRb5jh8z2E8hMGN2tkvs1yHynUanqnZ3UeKwgN1i9P1F8/go-log"
 	cid "gx/ipfs/QmcZfnkapfECQGcLZaf9B79NRg7cRa9EnZh4LSbkCzwNvY/go-cid"
 	ipld "gx/ipfs/Qme5bWv7wtjUNGsK2BNGVUFPKiuxWrsqrtvYwCLRw8YFES/go-ipld-format"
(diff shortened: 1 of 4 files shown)


pin/gc/gc.go
```diff
@@ -104,6 +105,18 @@ func GC(ctx context.Context, bs bstore.GCBlockstore, ls dag.LinkService, pn pin.
...
+
+		defer log.EventBegin(ctx, "GC.datastore").Done()
+		gds, ok := ds.(dstore.GCDatastore)
+		if !ok {
+			return
+		}
+
+		err = gds.CollectGarbage()
+		if err != nil {
+			output <- Result{Error: err}
```
Stebalien: This can deadlock if the context has been canceled and output is full. However, we already *have* this bug (above) so feel free to fix it separately. See #4593.

repo/fsrepo/datastores.go
```diff
@@ -366,6 +369,21 @@ func BadgerdsDatastoreConfig(params map[string]interface{}) (DatastoreConfig, er
 		}
 	}
 
+	vls, ok := params["vlogFileSize"]
+	if !ok {
+		c.vlogFileSize = 1 << 30
```
Stebalien: Why? Could you add a comment explaining this.

magik6k: It's the default size, will add a comment

magik6k:
This is currently blocked by go-datastore being updated in go-ds-measure and not here: https://github.com/ipfs/go-ds-measure/commit/d287c4e89e1fc5697d03a29a5256188636eb833b#diff-b9cfc7f2cdf78a7f4b91a753d10865a2

Stebalien:
Blocked on #4610.

Stebalien:
Unblocked, needs rebase.

magik6k:
Rebased

magik6k:
> What is the new vlogFileSize parameter and how does it relate to calling CollectGarabge on the datastores?

Badger is currently the only DS that implements GC, the minimum reclaimable space is defined as `max(10mb, vlogFileSize*gcTreshold)` per vlog. When I tried to create tests for this, I needed this parameter, so the tests wouldn't need to create multi-gb datastore to trigger gc.

> If this is not related to the purpose of this P.R. could we separate it out into a separate commit?
Done

whyrusleeping:
@magik6k sorry, need a rebase here.

Issue #4414, opened 2017-11-21
Audit all commands for goroutine leaks


Stebalien:
We're leaking goroutines on canceled commands all over the place.

To do this audit, I recommend looking at all channel writes and I mean **all**.

A quick grep indicates that the following files *may* have issues (should, at least, be looked at):

* [x] `core/commands/dht.go` -- #4413
* [x] `core/commands/repo.go` -- #4413
* [ ] `core/commands/add.go`
* [ ] `core/commands/dag/dag.go`
* [ ] `core/commands/files/files.go`
* [ ] `core/commands/commands.go`
* [ ] `core/commands/filestore.go`
* [ ] `core/commands/object/object.go`
* [ ] `core/commands/pin.go`
* [ ] `core/commands/ping.go`
* [ ] `core/commands/unixfs/unixfs.go`
* [ ] `core/commands/refs.go`

Basically, anytime we write to a channel that may block (i.e., doesn't have a buffer at least as large what we're writing), we should select on the channel write and `<-req.Context().Done()`.

keks:
@Stebalien Regarding your question in #4413:
> Should we be calling res.SetError(req.Context().Err(), cmdkit.ErrNormal) when the context is canceled or can we just bail (what I'm doing here).

When we cancel, the error should either be correctly handled or the command should return an error indicating the command was cancelled. Maybe even specify `"go-ipfs-cmds".Canceled` of type `cmdkit.Error`, analogue to `"context".Canceled`. If this doesn't happen, we should add something like
```go
select {
case <-ctx.Done():
  res.SetError(ctx.Err(), cmdkit.ErrNormal)
case default:
}
```
at the end of the respective commands. Does that sound reasonable to you?

Stebalien:
> Does that sound reasonable to you?

It's reasonable, just verbose and painful; in some cases, we realize that the context has been canceled in multiple places concurrently (thread safety issues). If the context has been canceled/expired, I assume the client isn't waiting for a response anyways. If they are, can the commands package not handle this?

Stebalien:
Note: We *should not* be closing the channel on context cancel. Otherwise, the other side might think that we're done sending the response.

forstmeier:
Hi, all! Is this issue still active? I'd like to get involved with some entry-level work on the project to get my hands dirty and this issue seemed like it still needed some work done. :smile:

alecbrick:
[...]

PR #4590, opened 2018-01-18
Avoid closing channels in commands if the context has errored (e.g timed out).

lgarron:
Previously, the following could happen:
- Producer creates channel.
- Consumer reads from channel.
- Producer observes that the context has errored, and closes the channel.
- Consumer sees the closed channel before seeing the context error, and assumes
  that command completed without a context error.

This change ensures the following:
- Producer creates channel.
- Consumer reads from channel.
- Producer observes that the context has errored, and DOES NOT close the
  channel.
- Consumer sees that the context has errored, and handles the channel
  appropriately.

Note that per the definition of context [1], a context is "done" iff it has not
"errored" (i.e. ctx.Err() is non-nil).

[1] https://godoc.org/context#Context

Updates call sites touched by #4413 and addresses #4414.

### DIFF
--- a/core/commands/commands.go
+++ b/core/commands/commands.go
@@ -6,6 +6,7 @@
 package commands
 
 import (
+	"context"
 	"fmt"
 	"io"
 	"sort"
@@ -151,4 +152,15 @@ func unwrapOutput(i interface{}) (interface{}, error) {
 	}
 
 	return <-ch, nil
+}
+
+// closeUnlessErr closes `c`, unless `req` encountered an error (e.g. a timeout
+// or a cancellation). `defer closeUnlessErr(ctx, c)` is useful as an
+// alternative for `defer close(c)` if you want to prevent the receiver from
+// seeing the closed channel when there was a context error (which might
+// accidentally let it assume that there was no context error).
+func closeUnlessErr(ctx context.Context, c chan interface{}) {
+	if ctx.Err() == nil {
+		close(c)
+	}
 }
--- a/core/commands/dht.go
+++ b/core/commands/dht.go
@@ -91,7 +91,7 @@ var queryDhtCmd = &cmds.Command{
 		res.SetOutput((<-chan interface{})(outChan))
 
 		go func() {
-			defer close(outChan)
+			defer closeUnlessErr(req.Context(), outChan)
 			for e := range events {
 				select {
 				case outChan <- e:
@@ -183,7 +183,7 @@ var findProvidersDhtCmd = &cmds.Command{
 
 		pchan := dht.FindProvidersAsync(ctx, c, numProviders)
 		go func() {
-			defer close(outChan)
+			defer closeUnlessErr(req.Context(), outChan)
 			for e := range events {
 				select {
 				case outChan <- e:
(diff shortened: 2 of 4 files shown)


lgarron:
@Stebalien, could you take a look and let me know if I should be doing anything differently?

lgarron:
Latest revision (f57391ea8a2a65d2fdc6cb5c71b693af965d31a9
) adds a test for `closeUnlessErr()`.

core/commands/commands.go
```diff
@@ -192,3 +193,14 @@ func unwrapOutput(i interface{}) (interface{}, error) {
 
 	return <-ch, nil
 }
+
+// closeUnlessErr closes `c`, unless `req` encountered an error (e.g. a timeout
+// or a cancellation). `defer closeUnlessErr(ctx, c)` is useful as an
+// alternative for `defer close(c)` if you want to prevent the receiver from
+// seeing the closed channel when there was a context error (which might
+// accidentally let it assume that there was no context error).
+func closeUnlessErr(ctx context.Context, c chan interface{}) {
```
magik6k: Not sure about the name, it doesn't really make it immediately obvious what the method does.

Stebalien: I agree but I can't think of a better name...

core/commands/dht.go
```diff
@@ -307,14 +307,14 @@ var provideRefDhtCmd = &cmds.Command{
...
 		go func() {
-			defer close(outChan)
 			for e := range events {
 				select {
 				case outChan <- e:
 				case <-req.Context().Done():
 					return
 				}
 			}
+			close(outChan)
```
Stebalien: I'd be consistent and use this function here if we use it elsewhere.

Stebalien:
> though it would be nice to get a test which reproduces the issue fixed

This would be very tricky to reproduce, I'm not entirely sure it *can* happen in practice (at the moment). It's a very unlikely race where the request gets canceled *before* the client goes away (e.g., by some server-side timeout) and then the server observes the close on the output channel *before* it observes the closed context.

core/commands/dht.go
lgarron: Done!

core/commands/dht.go
```diff
@@ -437,13 +437,12 @@ var findPeerDhtCmd = &cmds.Command{
 		ctx := notif.RegisterForQueryEvents(req.Context(), events)
 
 		go func() {
-			defer close(outChan)
+			closeUnlessErr(ctx, outChan)
```
Stebalien: defer

core/commands/dht.go
```diff
@@ -307,7 +307,7 @@ var provideRefDhtCmd = &cmds.Command{
...
```
Stebalien: defer

lgarron: Done, thanks!

I *think* I was also using the wrong context, so I've changed `ctx` to `req.Context()`.

core/commands/dht.go
lgarron: ditto

core/commands/commands.go
frrist: I think `maybeCloseChannel` could work, we use the `maybe` phrase elsewhere in go-ipfs for similar things.

Stebalien: IMO, that's worse. We usually use maybe to mean "do this if we need to". In this case, we're saying "do this unless we shouldn't".

Stebalien:
Will rebase and merge after the cmds update thing goes through.

Stebalien:
So, apparently this is *not* the correct solution due to how the commands lib works. Unfortunately, I'm pretty sure there *is* no correct solution at the moment. One *could* call `SetError` with the error but that could panic because the connection may be closed.

PR #4599, opened 2018-01-21
Feat/doc interfaces

whyrusleeping:
Document interfaces

### DIFF
--- a/blocks/blockstore/blockstore.go
+++ b/blocks/blockstore/blockstore.go
@@ -40,12 +40,19 @@ type Blockstore interface {
 	DeleteBlock(*cid.Cid) error
 	Has(*cid.Cid) (bool, error)
 	Get(*cid.Cid) (blocks.Block, error)
+
+	// Put puts a given block to the underlying datastore
 	Put(blocks.Block) error
+
+	// PutMany puts a slice of blocks at the same time using batching
+	// capabilities of the underlying datastore whenever possible.
 	PutMany([]blocks.Block) error
+
 	// AllKeysChan returns a channel from which
 	// the CIDs in the Blockstore can be read. It should respect
 	// the given context, closing the channel if it becomes Done.
 	AllKeysChan(ctx context.Context) (<-chan *cid.Cid, error)
+
 	// HashOnRead specifies if every read block should be
 	// rehashed to make sure it matches its CID.
 	HashOnRead(enabled bool)
--- a/core/core.go
+++ b/core/core.go
@@ -420,6 +420,8 @@ func setupDiscoveryOption(d config.Discovery) DiscoveryOption {
 	return nil
 }
 
+// HandlePeerFound attempts to connect to peer from `PeerInfo`, if it fails
+// logs a warning log.
 func (n *IpfsNode) HandlePeerFound(p pstore.PeerInfo) {
 	log.Warning("trying peer info: ", p)
 	ctx, cancel := context.WithTimeout(n.Context(), discoveryConnTimeout)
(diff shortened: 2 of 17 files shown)


blocks/blockstore/blockstore.go
```diff
@@ -40,12 +40,19 @@ type Blockstore interface {
 	DeleteBlock(*cid.Cid) error
 	Has(*cid.Cid) (bool, error)
 	Get(*cid.Cid) (blocks.Block, error)
+
+	// AddBlock puts a given block to the underlying datastore
```
karalabe: Definitely wrong method name here :D

whyrusleeping: mmmm, copy paste has not served me well

whyrusleeping:
@karalabe Any other bits that feel neglected?

core/core.go
```diff
@@ -670,6 +676,7 @@ func (n *IpfsNode) loadID() error {
 	return nil
 }
 
+// GetKey will return the IpfsNode key with name `name`.
```
whyrusleeping: maybe private key or keystore key instead of IpfsNode key

core/coreapi/coreapi.go
```diff
@@ -15,28 +15,34 @@ type CoreAPI struct {
...
 }
 
-// NewCoreAPI creates new instance of IPFS CoreAPI backed by go-ipfs Node
+// NewCoreAPI creates new instance of IPFS CoreAPI backed by go-ipfs Node.
 func NewCoreAPI(n *core.IpfsNode) coreiface.CoreAPI {
 	api := &CoreAPI{n}
 	return api
 }
 
+// Unixfs returns the Unixfs API.
```
whyrusleeping: I feel like there should be a thing after API...
```
// Unixfs returns the Unixfs API *doodad*
```

thoughts @magik6k @lgierth ?

core/coreapi/interface/interface.go
```diff
@@ -17,9 +17,13 @@ import (
 // Path is a generic wrapper for paths used in the API. A path can be resolved
 // to a CID using one of Resolve functions in the API.
 type Path interface {
+	// String returns the path as a string.
 	String() string
+	// Cid returns path cid
```
whyrusleeping: returns the cid that the path refers to. (as opposed to the root)

core/coreapi/interface/interface.go
```diff
@@ -17,9 +17,13 @@ import (
...
 	Cid() *cid.Cid
+	// Root returns path root
 	Root() *cid.Cid
+	// Resolved returns whether cid is nil
```
whyrusleeping: While this may be technically true, I think we want to say whether or not this path has been fully resolved.

cc @magik6k @lgierth

core/coreapi/interface/interface.go
```diff
@@ -33,22 +37,31 @@ type Reader interface {
...
 	Name() string
+	// Value returns IpnsEntry value
 	Value() Path
 }
 
+// Key specifies the interface to Keys in KeyAPI Keystore
 type Key interface {
+	// Key returns key name
 	Name() string
+	//Path returns key path
```
whyrusleeping: space before Path

core/coreapi/name.go
```diff
@@ -29,14 +29,17 @@ type ipnsEntry struct {
 	value coreiface.Path
 }
 
+// Name returns the ipnsEntry name.
 func (e *ipnsEntry) Name() string {
 	return e.name
 }
 
+// Value returns the inpnEntry value.
```
whyrusleeping: typo

whyrusleeping:
@magik6k wanna give this another look over?

core/coreapi/coreapi.go
frrist: how about `interface`?

frrist: or maybe
`// Unixfs create new instance of IPFS UnixfsAPI backed by go-ipfs Node.`

Issue #4524, opened 2017-12-27
Building for iOS with gomobile fails with: undefined: syscall.O_DSYNC


wehriam:
#### Version information:

N/A

#### Type:

Bug

#### Severity:

Low

#### Description:

Building for iOS with gomobile fails with an `undefined: syscall.O_DSYNC` error associated with [Badger](https://github.com/dgraph-io/badger), see [lines 23-25](https://github.com/dgraph-io/badger/blob/master/y/file_dsync.go#L23-L25).

Compiling with Go 1.10 beta throws the same error.

```
saucer:ipfs-ios wehriam$ gomobile build -target=ios github.com/ipfs/go-ipfs/cmd/ipfs
gomobile: go build -pkgdir=/Users/wehriam/go/pkg/gomobile/pkg_darwin_arm -tags ios -o=/var/folders/2s/cx9fndsd5md4yxcrx6z50rch0000gn/T/gomobile-work-919090730/arm github.com/ipfs/go-ipfs/cmd/ipfs failed: exit status 2
# gx/ipfs/QmXtSA4dxoDKMzhHkwqeoBoBYASov1ZWnLQo38juKNbe3z/badger/y
../../go/src/gx/ipfs/QmXtSA4dxoDKMzhHkwqeoBoBYASov1ZWnLQo38juKNbe3z/badger/y/file_dsync.go:24:21: undefined: syscall.O_DSYNC
```

```
saucer:ipfs-ios wehriam$ go version
go version go1.9.2 darwin/amd64
```

```
saucer:ipfs-ios wehriam$ gomobile version
gomobile version +50b7067 Sun Dec 17 15:46:28 2017 +0000 (android,ios); androidSDK=/Users/wehriam/Library/Android/sdk/platforms/android-26
```

```
saucer:ipfs-ios wehriam$ go env
GOARCH="amd64"
GOBIN=""
GOEXE=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOOS="darwin"
GOPATH="/Users/wehriam/go"
GORACE=""
GOROOT="/usr/local/opt/go/libexec"
GOTOOLDIR="/usr/local/opt/go/libexec/pkg/tool/darwin_amd64"
GCCGO="gccgo"
CC="clang"
[...]

Issue #4512, opened 2017-12-20
Dial error on windows 7


Stebalien:
#### Version information:

0.4.14-dev
#### Type:
Bug
#### Severity:
Critical
#### Description:

No peers in swarm. When dialing, @djdv gets:

```
Error: connect QmSoLMeWqB7YGVLJN3pNLQpmmEk35v6wYtsMGLzSr5QBU3 failure: dial attempt failed: <peer.ID WABRCq> --> <peer.ID SoLMeW> dial attempt failed: dial tcp4 0.0.0.0:4737->178.62.61.185:4001: bind: Only one usage of each socket address (protocol/network address/port) is normally permitted.
```

Caused by something in #4453.

Stebalien:
@djdv could you try disabling reuseport (set the environment variable `IPFS_REUSEPORT=false`). It's not a fix but it'll help me determine what's going on here.

gdeverlant:
I come up with another error with my tests on Windows 7. When I check the HTTP API via Chrome the requests are returning results.

When I try via my video editor to access the HTTA API via the URLLoader class the IPFS node returns 403 - Forbidden error.

![image](https://user-images.githubusercontent.com/26530975/34258813-f989cdf2-e65f-11e7-981d-2c71acc442a5.png)

gdeverlant:
Tried with the HTTPService API

![image](https://user-images.githubusercontent.com/26530975/34260754-2d01ec76-e667-11e7-9ca2-a7d4d3c8d893.png)

gdeverlant:
Accessing the API is working via Socket TCP Programming only...

![image](https://user-images.githubusercontent.com/26530975/34269103-6e4f2948-e683-11e7-9336-b3f5610c2b4a.png)

Stebalien:
@gdeverlant your errors appear to be unrelated to this issue. Please file a new one (or open a thread on https://discuss.ipfs.io as this doesn't look like a bug).

djdv:
@Stebalien 
For context I'm on Windows 10, this may be affecting all versions of Windows, since I'm seeing other users having empty peer tables as well on different versions, when using IPFS past #4453.

Setting IPFS_REUSEPORT to false doesn't seem to have any effect for me, I set it, start the daemon (built from current master), and try to connect to a peer, but encounter the same error as before. I just ran this with Go 1.10 but I was getting the same results on 1.9.2 a few days ago.

Promt 1
```
>ipfs version --all
[...]

PR #4610, opened 2018-01-25
mega update

Stebalien:
* supersedes #4597 but does not entirely replace it (i.e., not fixed).
* supersedes #4152
* punts #4009

Features:

* fixes build for iOS #4524 (@stebalien, @magik6k)
* new DAG interface from ipfs/go-ipld-format (@stebalien).
* (much) faster base58 (@stebalien).
* faster sha256 on some platforms (@stebalien, @lgierth).
* faster blake2b on all platforms (@stebalien, @lgierth).
* (massively) reduce memory usage of peerstore (@stebalien, @whyrusleeping, @Arceliar).
* #4512: fix dialing without reuseport/on Windows (@stebalien). 
* reliably detect DHT capable peers (@stebalien).
* ~correctly validate DHT records (@dirkmc).~ Use new DHT record validators.
* allocate less in the dht (@stebalien).
* (much) faster metrics (@stebalien).
* fix dropped pubsub messages (@vyzo, @richardschneider).
* fix some potential (rare) connection leaks on dial/accept (@FiloSottile).
* reduce excessive memory usage when mass dialing (@whyrusleeping).
* updated badgerdb (@magik6k)

If I'm missing something here, please comment.

To review this PR, please do so commit by commit for your own sanity.

---

Questions:

* [ ] Are we happy with the fix for #4009 (offline dag)? Currently, I just take a blockservice instead and construct an offline dagservice on the fly (which is what `GetOfflineLinkService` did anyways).


namesys/publisher.go
```diff
@@ -295,12 +304,31 @@ func selectRecord(recs []*pb.IpnsEntry, vals [][]byte) (int, error) {
...
+	// Note: The DHT will actually check the signature so we don't
+	// need to do that here
+
+	// Author in key must match author in record
+	parts := strings.Split(r.Key, "/")
+	pid, err := peer.IDB58Decode(parts[0])
+	if err != nil {
+		return ErrInvalidAuthor
+	}
+	if string(pid) != string(r.Author) {
```
dirkmc: We can remove the string casts here now right

test/sharness/t0183-namesys-pubsub.sh
```diff
@@ -26,9 +26,11 @@ test_expect_success 'check namesys pubsub state' '
     test_cmp expected state2
 '
 
+# These commands are *expected* to fail. We haven't published anything yet.
 test_expect_success 'subscribe nodes to the publisher topic' '
-    ipfsi 1 name resolve /ipns/$PEERID_0 &&
-    ipfsi 2 name resolve /ipns/$PEERID_0
+    ipfsi 1 name resolve /ipns/$PEERID_0;
+    ipfsi 2 name resolve /ipns/$PEERID_0;
+    true
```
Stebalien: @vyzo does this change make sense?

vyzo: hrm, shouldn't we be testing if they succeeded?
that's why i had the `&&`.

vyzo: the assumption is that name resolve returns 0 if it succeeds.

Stebalien: Yeah but shouldn't it *not* succeed? Nothing has been published yet, as far as I know.

vyzo: yeah, good point! these are executed for their side-effect of subscribing to the topic.

vyzo: So yes, change makes sense.

core/commands/dht.go
```diff
@@ -69,9 +69,13 @@ var queryDhtCmd = &cmds.Command{
 		events := make(chan *notif.QueryEvent)
 		ctx := notif.RegisterForQueryEvents(req.Context(), events)
 
-		k := string(b58.Decode(req.Arguments()[0]))
+		id, err := peer.IDB58Decode(req.Arguments()[0])
```
Stebalien: This broke a test or two but it's *strictly* more correct.

unixfs/mod/dagmodifier.go
```diff
@@ -330,7 +330,6 @@ func (dm *DagModifier) modifyDag(n node.Node, offset uint64, data io.Reader) (*c
 				return nil, false, err
 			}
 
-			offset += bs
```
Stebalien: @whyrusleeping go vet complained. I believe this is the correct fix here but I just wanted to check. Not really related to this PR, just a drive-by fix.

blockservice/blockservice.go
```diff
@@ -110,38 +122,34 @@ func NewSession(ctx context.Context, bs BlockService) *Session {
...
-	if err != nil {
-		return nil, err
+	if err := s.blockstore.Put(o); err != nil {
+		return err
 	}
 
 	if err := s.exchange.HasBlock(o); err != nil {
-		return nil, errors.New("blockservice is closed")
+		// TODO(stebalien): really an error?
+		return errors.New("blockservice is closed")
```
Stebalien: @whyrusleeping Should we not just log and continue here?

blockservice/blockservice.go
```diff
@@ -153,18 +161,16 @@ func (s *blockService) AddBlocks(bs []blocks.Block) ([]*cid.Cid, error) {
...
-		return nil, err
+		return err
 	}
 
-	var ks []*cid.Cid
 	for _, o := range toput {
 		if err := s.exchange.HasBlock(o); err != nil {
-			return nil, fmt.Errorf("blockservice is closed (%s)", err)
+			// TODO(stebalien): Should this really *return*?
+			return fmt.Errorf("blockservice is closed (%s)", err)
```
Stebalien: Same?

core/commands/pin.go
```diff
@@ -594,7 +596,10 @@ type pinVerifyOpts struct {
 
 func pinVerify(ctx context.Context, n *core.IpfsNode, opts pinVerifyOpts) <-chan interface{} {
 	visited := make(map[string]PinStatus)
-	getLinks := n.DAG.GetOfflineLinkService().GetLinks
+
+	bs := n.Blocks.Blockstore()
+	DAG := dag.NewDAGService(bserv.New(bs, offline.Exchange(bs)))
+	getLinks := dag.GetLinksWithDAG(DAG)
```
Stebalien: This is the new way to do offline stuff. We could also use the context trick if we get tired of this dance.

core/corerepo/gc.go
```diff
@@ -154,7 +154,7 @@ func GarbageCollectAsync(n *core.IpfsNode, ctx context.Context) <-chan gc.Result
 		return out
 	}
 
-	return gc.GC(ctx, n.Blockstore, n.DAG, n.Pinning, roots)
+	return gc.GC(ctx, n.Blockstore, n.Pinning, roots)
```
Stebalien: Now constructs a dag from the blockstore.

importer/helpers/helpers.go
```diff
@@ -133,7 +133,11 @@ func (n *UnixfsNode) FileSize() uint64 {
 }
 
 func (n *UnixfsNode) SetPosInfo(offset uint64, fullPath string, stat os.FileInfo) {
-	n.posInfo = &pi.PosInfo{offset, fullPath, stat}
+	n.posInfo = &pi.PosInfo{
+		Offset:   offset,
+		FullPath: fullPath,
+		Stat:     stat,
+	}
```
Stebalien: `go vet` yelled at me :fearful:.

merkledag/batch.go
```diff
@@ -1,99 +0,0 @@
...
-	blocks "gx/ipfs/Qmej7nf81hi2x2tvjRBF3mcp74sQyuDH4VMYDGd1YtXjb2/go-block-format"
-)
-
-// ParallelBatchCommits is the number of batch commits that can be in-flight before blocking.
-// TODO(#4299): Experiment with multiple datastores, storage devices, and CPUs to find
-// the right value/formula.
-var ParallelBatchCommits = runtime.NumCPU() * 2
-
-// Batch is a buffer for batching adds to a dag.
-type Batch struct {
```
Stebalien: Now in go-ipld-format.

blockservice/blockservice.go
```diff
@@ -110,38 +126,34 @@ func NewSession(ctx context.Context, bs BlockService) *Session {
...
+		// TODO(stebalien): really an error?
+		return errors.New("blockservice is closed")
 	}
 
-	return c, nil
+	return nil
 }
 
-func (s *blockService) AddBlocks(bs []blocks.Block) ([]*cid.Cid, error) {
+func (s *blockService) AddBlocks(bs []blocks.Block) error {
```
vyzo: are we losing any functionality by dropping the cids from the return type?

core/commands/dht.go
vyzo: sounds like faulty tests!

core/commands/pin.go
vyzo: this is 3 lines instead of 1; ergonomically it's not an improvement...

namesys/publisher.go
```diff
@@ -322,7 +322,7 @@ func ValidateIpnsRecord(r *record.ValidationRecord) error {
 	if err != nil {
 		return ErrInvalidAuthor
 	}
-	if string(pid) != string(r.Author) {
+	if pid != r.Author {
```
vyzo: clown shoes!

blockservice/blockservice.go
Stebalien: Not really. We can get the CIDs from the blocks so there's no need to return them here. It just complicates the blockservice (and dagservice which is why I removed this).

namesys/publisher.go
Stebalien: :confused:

core/commands/dht.go
Stebalien: Yeah. I just wonder if anyone was relying on this in practice. One could theoretically use this to find the peers closest to some base58 encoded thing that's *not* a valid peer ID.

core/commands/pin.go
Stebalien: Discussion: https://github.com/ipfs/go-ipfs/issues/4009

So, first, I made implementing `LinkService` optional. If your `NodeGetter` (the get side of the `DAGService`) doesn't implement the `LinkService` optimizations, it doesn't have to implement the interface.

Next, I tried adding a `OfflineNodeGetter` method to the `NodeGetter` interface. However, that really tricky and messy:

1. We realized that we'd probably want `Offline*` methods on every related service (which would kind of suck).
2. Things like bitswap sessions (which I wanted to turn into valid `NodeGetter`s would *also* have to implement this method). I simply couldn't find a clean way to do this.

So, in #4009, I suggested adding an "offline" flag to the context. I *still* think that's the best way to go because I keep running into similar cases where I want to try to do something without touching the network but @kevina reasonably objected that it was a bit magical and easy to screw up.

Hence this solution.

I could add a helper function but, really, I kind of want this to be ugly. I'm not happy with this solution at all and don't want this pattern repeated all over the codebase. For one, it makes the `LinkService` optimization totally useless because we don't *reuse* the DAG (we construct a new one with the blockstore when we do the GC).

vyzo: fair enough.

whyrusleeping:
@Stebalien sharness on OSX is failing.

whyrusleeping:
```
ok 10 - findpeer

expecting success:
  echo "$TEST_DHT_VALUE" | base64 -d | ipfsi 0 dht put "$TEST_DHT_PATH" | sort >putted &&
  [ -s putted ] ||
  test_fsh cat putted

base64: invalid option -- d
Usage:	base64 [-hvD] [-b num] [-i in_file] [-o out_file]
  -h, --help     display this message
  -D, --decode   decodes input
  -b, --break    break encoded string into num character lines
  -i, --input    input file (default: "-" for stdin)
  -o, --output   output file (default: "-" for stdout)
> cat putted

not ok 11 - put with good keys
#
#	  echo "$TEST_DHT_VALUE" | base64 -d | ipfsi 0 dht put "$TEST_DHT_PATH" | sort >putted &&
#	  [ -s putted ] ||
#	  test_fsh cat putted
#
```

whyrusleeping:
```
why@whys-MacBook-Pro ~/g/s/g/i/g/t/sharness> base64 --help
Usage:	base64 [-hvD] [-b num] [-i in_file] [-o out_file]
  -h, --help     display this message
  -D, --decode   decodes input
  -b, --break    break encoded string into num character lines
  -i, --input    input file (default: "-" for stdin)
  -o, --output   output file (default: "-" for stdout)
```

Stebalien:
And this is why we can't have nice, portable things.

---

Also, that's not the only failure. My little legacy command change also caused an error. `repo gc` calls `SetError` after calling `SetOutput(outChan)`. Do you know the *proper* way to error out after setting the output channel?

whyrusleeping:
@Stebalien SetError should still work, under the hood it should be setting the http trailer. Same as if ipfs add fails out halfway through streaming the file

Stebalien:
> SetError should still work, under the hood it should be setting the http trailer. 

I was worried about concurrency but, as long as we call SetError *before* closing the output channel we should be fine.

However, `SetError` *won't* use a trailer unless we set a *fatal* error. Otherwise, it will just do some weird ducktyping stuff.

test/ipfs-test-lib.sh
```diff
@@ -109,3 +109,13 @@ test_seq() {
...
 }
+
+b64decode() {
+    for flag in "-D" "-d"; do
+        if base64 "$flag" 2>/dev/null; then
+            return
+        fi
+    done
+    echo "no compatible base64 command found" >&2
+    return 1
```
kevina: I am not sure how I feel about this little hack.  What if one of the base64 supports both options and they do two different things?

Maybe a simpler and more portable solution would be to write our own tool in go and use that, or install someone else's tool in `test/sharness/bin`.

Stebalien: Meh, it's a test case. MacOS uses `-D` and everything else uses `-d`. We could install and use a different command but I'd rather not unless you *really* think this hack is going to be a problem.

kevina: Could we at least switch based on the OS?  I will be okay with that.

It could be fine, but it could also create a hard to track down bug, several years out if by the very off chance `-D` does something other then decode in a future version.

Stebalien: @kevina fair enough. I was trying to be smart but you're right, being smart is probably a bad idea in this case.

Stebalien:
This actually gives us a tiny but nice performance win (better hashing library? maybe?). Transfer of a large file between two local nodes is now reliably ~1% faster and adding with filestore is also a bit faster (badger may be even faster as we won't have that bottleneck).

Stebalien:
@whyrusleeping tests pass. Some of the commands stuff is still flakey but that's an unrelated bug, as far as I can tell.

Stebalien:
(well, once someone reviews it)

core/core.go
```diff
@@ -43,38 +43,38 @@ import (
...
-	connmgr "gx/ipfs/QmainUZ19HogMBjMocvHmgWEqofbpzLGNAir4AEY1KcJey/go-libp2p-connmgr"
-	mafilter "gx/ipfs/QmcrNvQBZkpbT7zLcGyjdPCiCHLVtaYeg4UVUH7XLfWJWy/go-maddr-filter"
-	addrutil "gx/ipfs/Qmcxa9y1KVC51TVicSKomsnunJGSA9UJSuBvJc28JCip4H/go-addr-util"
-	ds "gx/ipfs/QmdHG8MAuARdGHxx4rPQASLcvhz24fzjSQq7AJRAQEorq5/go-datastore"
-	cid "gx/ipfs/QmeSrf6pzut73u6zLQkRFQ3ygt3k6XFT2kjdYP8Tnkwwyg/go-cid"
-	dht "gx/ipfs/QmfUvYQhL2GinafMbPDYz7VFoZv4iiuLuR33aRsPurXGag/go-libp2p-kad-dht"
+	metrics "gx/ipfs/Qmb1QrSXKwGFWgiGEcyac4s5wakJG4yPvCPk49xZHxr5ux/go-libp2p-metrics"
+	cid "gx/ipfs/QmcZfnkapfECQGcLZaf9B79NRg7cRa9EnZh4LSbkCzwNvY/go-cid"
+	dht "gx/ipfs/Qmdm5sm2xHCXNaWdxpjhFeStvSNMRhKQkqpBX7aDcqXtfT/go-libp2p-kad-dht"
+	node "gx/ipfs/Qme5bWv7wtjUNGsK2BNGVUFPKiuxWrsqrtvYwCLRw8YFES/go-ipld-format"
```
magik6k: With more things in `go-ipld-format` we may want to change the naming from `node` to something like `ipld` - imo node is a bit confusing for some of the things that are now there.

magik6k:
Looked at the rest of the changes locally, other than the two comments LGTM

core/core.go
magik6k: If we decide to, I'm fine with thin being in another PR, this one is big enough.

Stebalien: Meh. Might as well do this now. Anyways, go makes this pretty easy :).

Stebalien:
Please ignore codeclimate. It loves to punish those who refactor for the sins of others.

blockservice/blockservice.go
whyrusleeping: unclear. For now lets leave it as is.

Stebalien: Fair. Reported at #4623 (and the comment has been updated to point to the issue).

Stebalien:
CI failure: #3970.

core/commands/dht.go
whyrusleeping: should probably make sure that this isnt how we do random-walk bootstrapping

whyrusleeping: oh wait, this is in a command. Nvm

blockservice/blockservice.go
whyrusleeping: I wouldn't call that an enhancement

pin/gc/gc.go
```diff
@@ -33,15 +35,16 @@ type Result struct {
...
 
 	elock := log.EventBegin(ctx, "GC.lockWait")
 	unlocker := bs.GCLock()
 	elock.Done()
 	elock = log.EventBegin(ctx, "GC.locked")
 	emark := log.EventBegin(ctx, "GC.mark")
 
-	ls = ls.GetOfflineLinkService()
+	bsrv := bserv.New(bs, offline.Exchange(bs))
+	ds := dag.NewDAGService(bsrv)
```
kevina: This was the way we originally did things back before 721df367a27dfa416e1d93df23ca440fc2431f63.  The reason I put it in was not for catching but for the filestore.  With the current filestore it is not technically needed, but with a richer more functional filestore it might be again.  So that the record is straight I will add a comment outside of the review a little later so its not so easily lost.

blockservice/blockservice.go
kevina: Fair enough.  But it is still possible for something to go wrong while adding the blocks, such as running out of space.  However, since the underlying BlockStore does not indicate which blocks where added there is no way to get this information.  I will make a separate issue for this so we can figure out a proper solution.

pin/gc/gc.go
Stebalien: I'm really starting to think we should just be using a context hint. Anyways, all this will need to be reworked when GC changed to operate over a DAG instead of a blockservice.

kevina:
This was merged before I had a chance to really review it, sorry I was not more on the ball here but I am _not_ happy with the fix for #4009, I'll add my comment to #4009 where it will have a better chance of being seen.

Stebalien:
That wasn't a fix, it was a workable hack to allow us to move forward and agree on a fix (that's why I called it out and asked if it was ok for now).

kevina:
@Stebalien and sorry again I was not on the ball here.  That change did not imminently register until you closed my p.r.

PR #4624, opened 2018-01-29
expand the issue template and remove the severity field

Stebalien:
* Direct users seeking support to the forums.
* Remove the severity field. It doesn't actually help (especially because we can't filter by it and, anyways, *we* end up deciding the severity).
* Put the section comments *inside* the sections. This has been annoying me from day 1.
* Put the welcome text *at the top* so users actually read it.
* Expand on what we want in a description.

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,33 +1,38 @@
 <!--
+Welcome to the go-ipfs bug tracker. This is for you! Please read, and then delete this text before posting it.
+
+If you haven't yet searched the issue tracker for an existing report concerning your issue, please do so now.
+
+The go-ipfs issues are only for bug reports and directly actionable feature requests. Read https://github.com/ipfs/community/blob/master/contributing.md#reporting-issues if your issue doesn't fit either of those categories.
+
+If you have a *SUPPORT QUESTION*, please direct it to our forum at https://discuss.ipfs.io.
+
+Read https://github.com/ipfs/go-ipfs/blob/master/docs/github-issue-guide.md if you are not sure how to fill in this issue.
+-->
+
+#### Version information:
+<!--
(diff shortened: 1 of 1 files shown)


Stebalien:
> I am not convinced removing the Severity field is a good idea and would like others feedback on that.

I've literally never found it useful (not even for the entertainment value). From reading the issue, I can tell how sever it is. Without reading the issue, I can't trust the severity field.

IMO, severity is better handled through labels anyways (searchable, easy to see, not set by the reporter). I wonder if it's possible to give labeling permission to contributors.

kevina:
> IMO, severity is better handled through labels anyways (searchable, easy to see, not set by the reporter). I wonder if it's possible to give labeling permission to contributors.

I agree.  Not sure about the possibility of giving labeling permission to contributors.

Stebalien:
> I agree. Not sure about the possibility of giving labeling permission to contributors.

Fine grained permissions? GitHub? Silly me. We'd have to use a bot.

whyrusleeping:
jenkins fails for some reason? Want to file that as an issue?

kevina:
Since we decided to remove the Severity field we also need to fix https://github.com/ipfs/go-ipfs/blob/master/docs/github-issue-guide.md.

Stebalien:
@whyrusleeping standard "timed out waiting for api file show up" bug.

whyrusleeping:
@Stebalien that generally means that the daemon failed to start for some reason. Not a good sign

Stebalien:
@whyrusleeping unfortunately, it's a common race condition in our tests. Daemon takes too long to start (over 5 seconds). There may be some underlying cause bit I have no idea what it is; I've never reproduced this locally.

whyrusleeping:
Annoying, but alright.

PR #4627, opened 2018-01-30
fix hidden merge conflicts

Stebalien:
Oh for a CI that runs on merge...


Stebalien:
@magik6k looks like a reasonable codeclimate complaint. Is there an issue for deduping this?

magik6k:
@Stebalien this is going to be fixed when commands are rewritten to use coreapi

PR #4628, opened 2018-01-31
namesys: verify signature in ipns validator

dirkmc:


namesys/routing.go
```diff
@@ -131,58 +130,37 @@ func (r *routingResolver) resolveOnce(ctx context.Context, name string) (path.Pa
...
+		log.Debugf("RoutingResolver: could not convert public key hash %s to peer ID: %s\n", name, err)
+		return "", err
 	}
 
-	// check sig with pk
-	if ok, err := pubkey.Verify(ipnsEntryDataForSig(entry), entry.GetSignature()); err != nil || !ok {
-		return "", fmt.Errorf("ipns entry for %s has invalid signature", h)
+	// use the routing system to get the name.
+	_, ipnsKey := IpnsKeysForID(pid)
+	val, err := r.routing.GetValue(ctx, ipnsKey)
```
Stebalien: Probably a good idea to put a comment here stating that the DHT record validator validates the signature.

namesys/validator.go
```diff
@@ -0,0 +1,89 @@
...
+		pubk := kbook.PubKey(pid)
+		if pubk == nil {
+			log.Debugf("public key with hash %s not found in peer store", pid)
+			return ErrSignature
+		}
+
+		// Check the ipns record signature with the public key
+		if ok, err := pubk.Verify(ipnsEntryDataForSig(entry), entry.GetSignature()); err != nil || !ok {
+			log.Debugf("failed to verify signature for ipns record %s", r.Key)
+			return ErrSignature
```
Stebalien: I'd return a different error for each of these cases. As-is, this may make it difficult to debug issues.

namesys/selector.go
```diff
@@ -0,0 +1,67 @@
...
+			continue
+		}
+
+		if besti == -1 || r.GetSequence() > bestSeq {
+			bestSeq = r.GetSequence()
+			besti = i
+		} else if r.GetSequence() == bestSeq {
+			rt, err := u.ParseRFC3339(string(r.GetValidity()))
+			if err != nil {
+				continue
```
Stebalien: I'd at least log an ERROR in these invalid cases. Ideally, we'd panic but let's leave it as a noisy log for a while before we do something like that.

namesys/validator.go
dirkmc: Are you saying a different error for the `ok` and `err` return values of `pubk.Verify()`

Or are you saying a different error for the return values from
`peer.IDFromString(r.Key)`
`kbook.PubKey(pid)`
`pubk.Verify()`

Stebalien: Just different error messages (instead of always using `ErrSignature`). That way, we can distinguish between, e.g., not having the key and failing to validate the signature.

dirkmc: Makes sense

namesys/publisher.go
```diff
@@ -208,7 +194,7 @@ func PublishEntry(ctx context.Context, r routing.ValueStore, ipnskey string, rec
 	}
 
 	log.Debugf("Storing ipns entry at: %s", ipnskey)
-	// Store ipns entry at "/ipns/"+b58(h(pubkey))
+	// Store ipns entry at "/ipns/"+h(pubkey)
```
vyzo: wait, has this changed?

namesys/selector.go
```diff
@@ -0,0 +1,69 @@
...
+
+	for i, r := range recs {
+		if r == nil || r.GetSequence() < bestSeq {
+			continue
+		}
+
+		if besti == -1 || r.GetSequence() > bestSeq {
+			bestSeq = r.GetSequence()
+			besti = i
+		} else if r.GetSequence() == bestSeq {
```
vyzo: Is the else a bug here? you still want to parse when you see a higher sequence number.

namesys/validator.go
```diff
@@ -0,0 +1,101 @@
...
+var ErrKeyFormat = errors.New("record key could not be parsed into peer ID")
+
+// ErrPublicKeyNotFound should be returned when the public key
+// corresponding to the ipns record path cannot be retrieved
+// from the peer store
+var ErrPublicKeyNotFound = errors.New("public key not found in peer store")
+
+// NewIpnsRecordValidator returns a ValidChecker for IPNS records
+// The validator function will get a public key from the KeyBook
+// to verify the record's signature
```
vyzo: can we add a comment that the public key has already been added?

namesys/publisher.go
dirkmc: I didn't change the relevant code, I just noticed that the comment was inaccurate.
[It seems like](https://github.com/ipfs/go-ipfs/blob/master/namesys/publisher.go#L372) the DHT key is `"/ipns/" + string(<multihash>)`
Note that in the IPNS path exposed to the user the hash is B58 encoded. The resolver [decodes it](https://github.com/ipfs/go-ipfs/blob/master/namesys/routing.go#L127) to get the DHT key.

namesys/selector.go
dirkmc: I didn't write this code, I just moved it from publisher.go into its own file, so I'm not 100% sure what the original intent was here. But I believe it's just trying to find the index of the best record, not to parse the record. Am I understanding your question correctly?

namesys/publisher.go
vyzo: ok, cool. yeah, let's fix the comment if it's inaccurate.

namesys/selector.go
vyzo: Well it's not clear -- it seems to intent to parse the record with the best sequence.
The problem with the `else` is that if a new higher sequence is found, then it will be set as best signature but will not be parsed. and that's inconsistent.
I believe that's a bug for the intent of the code to parse the record with the best sequence number, it will only parse the first one!
If we remove the `else` and move the `if` below it would be consistent.

dirkmc: As I understand it the validator parses the record and checks that the EOL is in the correct format and has not expired. The selector just needs to pick the "best" record. In this code it chooses the best record as the record with
- the highest sequence number
- if sequence numbers are equal then
  - the highest expiration
  - if expiration is equal then
    - the highest by byte compare (so that the order is always consistent)

Having said that I think you're right that if the EOL cannot be parsed from the record validity then that record should not be chosen as the best record, so it might be worth adding that check in there.

Stebalien: So, we could avoid that by parsing the validity in the first condition (when we haven't selected a record yet). However, this case *should* never happen. That's why I suggested printing an error for now and eventually upgrading it to a panic.

whyrusleeping:
@dirkmc mind rebasing this one for us?

dirkmc:
Sure, rebase is at https://github.com/ipfs/go-ipfs/pull/4648

vyzo:
@dirkmc there is no reason to open a new pr, you can force push the rebase on this one!

dirkmc:
Ah ok I wasn't sure if that would mess up the history, let me give it a shot

dirkmc:
Looks like that worked, thanks @vyzo

namesys/selector.go
Stebalien: @dirkmc I'm fine with this as-is but if you'd *like* to add that that extra validity check, feel free and tell me when you're comfortable with having this merged.

dirkmc: I added it in, seems like the right thing to do, thanks @Stebalien

dirkmc: And thanks @vyzo for pointing it out :)

Stebalien:
@whyrusleeping I believe this is ready to be merged. We don't have to bubble https://github.com/libp2p/go-libp2p-record/pull/15 first. This doesn't yet *remove* signatures from DHT records but fixes the validation of IPNS records (stops signing them at the DHT record level and starts validating the *IPNS* signature).

PR #4629, opened 2018-01-31
minor drive-by code cleanup

Stebalien:
We'll dedup the code later, for now:

1. Don't drop errors on the floor.
2. Don't modify anything returned by `Links()` directly...

### DIFF
--- a/core/commands/object/object.go
+++ b/core/commands/object/object.go
@@ -657,24 +657,28 @@ func deserializeNode(nd *Node, dataFieldEncoding string) (*dag.ProtoNode, error)
 	case "text":
 		dagnode.SetData([]byte(nd.Data))
 	case "base64":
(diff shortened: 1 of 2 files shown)


core/commands/object/object.go
```diff
@@ -657,24 +657,28 @@ func deserializeNode(nd *Node, dataFieldEncoding string) (*dag.ProtoNode, error)
...
 			return nil, err
 		}
-		dagnode.Links()[i] = &ipld.Link{
+		links[i] = &ipld.Link{
 			Name: link.Name,
 			Size: link.Size,
 			Cid:  c,
 		}
 	}
+	dagnode.SetLinks(links)
```
whyrusleeping: <3 so much less sketchy

PR #4430, opened 2017-11-27
Create Jenkinsfile

VictorBjelkholm:
Got them cross-platform tests for go-ipfs here. Pipeline is here (will move to IPFS org) https://github.com/VictorBjelkholm/jenkins-libs/blob/master/vars/golang.groovy

Still have some refactoring to do with the pipeline and figure out if to skip fuse tests or install fuse on all workers. Thoughts?

PR #4632, opened 2018-01-31
Fix go test on Windows

magik6k:
This fixes windows issues @VictorBjelkholm had in https://github.com/ipfs/go-ipfs/pull/4430#issuecomment-352016810.

There is still one test failing:
```
--- FAIL: TestReprovide (0.05s)
        reprovide_test.go:51: Should have gotten a provider
```

This will need a fix in https://github.com/libp2p/go-libp2p-peer/blob/master/test/utils.go#L23. The problem here is that `time.Now().UnixNano()` resolution on windows is really poor (multiple ms)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+assets/init-doc/*          binary
+core/coreunix/test_data/** binary
--- a/cmd/ipfs/util/ulimit_test.go
+++ b/cmd/ipfs/util/ulimit_test.go
@@ -1,3 +1,5 @@
+// +build !windows
+
 package util
 
 import (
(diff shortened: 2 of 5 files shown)


VictorBjelkholm:
https://github.com/libp2p/go-libp2p-peer/pull/28 has been merged and this seems to be ready to be merged.

magik6k:
It has been merged, though it needs to be propagated. https://github.com/ipfs/go-ipfs/pull/4637/files updates go-libp2p-peer, though one version before thin fix

whyrusleeping:
@magik6k this will need a rebase. Let me know when its good to go

PR #4636, opened 2018-02-01
merge github-issue-guide with the issue template

Stebalien:
This is *one* way to address @kevina's feedback about updating the github-issue-guide. This way, we don't have to maintain two documents (one of which most users won't bother to read).

Basically, I pulled (what I considered to be) the relevant information from the issue guide to the template.

Thoughts?


docs/github-issue-guide.md
```diff
@@ -1,87 +0,0 @@
...
-
-Your processor architecture.
-
-- "x86"
-- "amd64"
-- "Arm"
-
-### Area
-
-What your issue refers to. Multiple items are OK.
```
Stebalien: This didn't exist on the issue template. Should we put it in? Most users probably won't know enough to fill this out, unfortunately. We usually just use labels.

docs/github-issue-guide.md
```diff
@@ -1,87 +0,0 @@
...
-
-We use GitHub Issues to log all of our todos and tasks. Here is
-[a good guide](https://guides.github.com/features/issues/) for them if you are
-unfamiliar.
-
-When logging an issue with go-ipfs, it would be useful if you specified the
-below information, if possible. This will help us triage the issues faster.
-Please title your issues with the type. For instance:
-
-- "bug: Cannot add file with `ipfs add`"
```
Stebalien: We don't actually do this... Should we start?

ISSUE_TEMPLATE.md
```diff
@@ -1,25 +1,38 @@
...
 #### Type:
-<!-- Bug, Feature, Enhancement, Etc -->
+<!--
+
+- "bug": If what you are filing is a bug.
+- "feature": If you'd like to suggest a feature.
+- "enhancement ": If you'd like to suggest an improvement on to existing feature.
+- "test failure": If the tests are failing
+- "doc": If it's related to missing/incorrect documentation.
+- "meta": If its a meta issue about this project/this repository/issue tracker.
```
kevina: You left out panic, I think that is useful for cases when ipfs crashes or other obvious bug.  A panic in my view indicates a programmer error.

docs/github-issue-guide.md
kevina: I am not against it, but unless we do it consistently all it will do is add noise.

docs/github-issue-guide.md
kevina: It could be useful, but you are right most users won't know enough to fill this out.  Users who do likely already have write access and can assign lables, that is if they remember to do so. :)

Stebalien: That *never* happens. I have no idea what you're talking about.

ISSUE_TEMPLATE.md
Stebalien: Ah. Good point. Delineates "I think this is a bug" and "this is definitely a bug".

ISSUE_TEMPLATE.md
```diff
@@ -1,25 +1,39 @@
...
+
+If you can't run `ipfs version --all` or that command fails, include as much information as you can: IPFS version, computer architecture (e.g., Intel x86 64bit), operating system, etc.
 -->
 
 #### Type:
-<!-- Bug, Feature, Enhancement, Etc -->
+<!--
+
+- "bug": If what you are filing is a bug.
+- "panic": If you'd like to report an application panic, segfault, crash, etc.
```
kevina: I suppose that is a bit more specific than the original description of "If it is a severe bug.", but I think I would use this also when the server dies with an error message when responding to a (perhaps invalid) request from the client.

kevina:
I think the threshold for the **codecov/project** test needs adjusting.  This p.r. doesn't even touch the code.  @Kubuxu how about something like 0.5%?

docs/github-issue-guide.md
Kubuxu: I used to tag issues with subsystem/topic labels. It was endless work and it somehow stopped being important.

eingenito:
Sorry - should not have been closed.

eingenito:
These seem like positive changes - should I rebase and merge? It sounds like maybe this was held up by coverage?

Stebalien:
Molly has heavily edited the issue template since then. But yeah, it would be nice to get some of this information into the template.

eingenito:
Refreshed and mergeable if people find it valuable.

momack2:
Updated this with the instructions Steven and I had workshopped. LGTM

PR #4637, opened 2018-02-01
gx: update go-log

Stebalien:


gx-workspace-update.json
```diff
@@ -0,0 +1,93 @@
+{
```
frrist: I think we will want to remove this before we merge

whyrusleeping: @Stebalien I want you to handwrite "I will not use 'git add .'" 100 times

Stebalien: @whyrusleeping right after I add my updated `.gitignore` with `git add .`.

PR #4638, opened 2018-02-01
add a --with-local option to ipfs files stat

MichaelMure:
Ok, new attempt after https://github.com/ipfs/go-ipfs/pull/4541 and https://github.com/ipfs/go-ipfs/pull/4596.

This PR add a `--with-local` option to `ipfs files stat` that compute and display how much of a dag is local.

I had a small problem with a the functions defined in `env.go` that are not reachable from a subdir of `core/commands` so I copied the file for now. I'm not sure how you want to solve that.

On a side note, I would be very pleased if any version can make it to the upcoming RC and version.


MichaelMure:
@whyrusleeping @Stebalien @kevina please have a look.

core/commands/files/files.go
```diff
@@ -69,6 +73,17 @@ var hashOption = cmdkit.StringOption("hash", "Hash function to use. Will set Cid
 
 var formatError = errors.New("Format was set by multiple options. Only one format option is allowed")
 
+type StatOutput struct {
+	Hash           string
+	Size           uint64
+	CumulativeSize uint64
+	Blocks         int
+	Type           string
+	WithLocality   bool
```
kevina: Please add `` `json:",omitempty"` `` after WithLocality so that the API output does not change if the `--with-local` option is used.

kevina:
I would also like to have @keks look over this as you are switching over the to new commands library.

kevina:
> I had a small problem with a the functions defined in env.go that are not reachable from a subdir of core/commands so I copied the file for now. I'm not sure how you want to solve that.

I am looking into this.

kevina:
> I had a small problem with a the functions defined in env.go that are not reachable from a subdir of core/commands so I copied the file for now. I'm not sure how you want to solve that.

Part of the problem is the `files.go` is in its own directory but not a proper package.  Since this is just a single file, the easiest solution is to just move it.  Commit b573768 (branch [kevina/statlocal-fix2](https://github.com/ipfs/go-ipfs/tree/kevina/statlocal-fix2)) does this.  Feel free to cherry pick that change.

The other solution is to properly make `files.go` into its own package.  I tried that in 554752c (branch [kevina/statlocal-fix2](https://github.com/ipfs/go-ipfs/tree/kevina/statlocal-fix2)) but ran into a import cycle:
```
import cycle not allowed
package github.com/ipfs/go-ipfs/cmd/ipfs
        imports github.com/ipfs/go-ipfs/core/commands
        imports github.com/ipfs/go-ipfs/core/commands/files
        imports github.com/ipfs/go-ipfs/core/commands
cmd/ipfs/Rules.mk:22: recipe for target 'cmd/ipfs/ipfs' failed
```
With a little more effort this could be properly fixed, but I think its best just to move `files.go` out of it own directory.

core/commands/files/files.go
MichaelMure: Someone to weight here ? For me it makes more sense like this. There is no compatibility problem, it's just an extra field. The problem is If `WithLocality` is set to false, the field is not present in the api anwser so you need to test if the field is present *and* if it's set to true. Not the end of the world, but ...

whyrusleeping: @MichaelMure No, the same struct would be used on both sides. theres no need to check if its set, we just don't want the API to change where its not wanted. @kevina is right here.

core/commands/files.go
```diff
@@ -49,31 +55,42 @@ operations.
...
-		"write": FilesWriteCmd,
-		"mv":    FilesMvCmd,
-		"cp":    FilesCpCmd,
-		"ls":    FilesLsCmd,
-		"mkdir": FilesMkdirCmd,
-		"stat":  FilesStatCmd,
-		"rm":    FilesRmCmd,
-		"flush": FilesFlushCmd,
-		"chcid": FilesChcidCmd,
+		"read":  lgc.NewCommand(filesReadCmd),
```
whyrusleeping: Alternatively, You could just use the `OldSubcommands` map instead of calling `legacy.NewCommand` on all of these. I'm not that picky though.

core/commands/files.go
```diff
@@ -191,24 +225,24 @@ func statNode(ds ipld.DAGService, fsn mfs.FSNode) (*Object, error) {
 		}
 
 		var ndtype string
-		switch fsn.Type() {
-		case mfs.TDir:
+		switch d.GetType() {
+		case ft.TDirectory, ft.THAMTShard:
```
whyrusleeping: why make this change?

whyrusleeping:
I would really prefer that we not move files.go back up into the "all the commands" directory. But it does make it pretty obnoxious to keep it down there...

kevina:
@whyrusleeping 
> I would really prefer that we not move files.go back up into the "all the commands" directory. But it does make it pretty obnoxious to keep it down there...

To fix the import cycle we will need to move `env.go` into its own package also.  I updated the [kevina/statlocal-fix2](https://github.com/ipfs/go-ipfs/tree/kevina/statlocal-fix2) branch to do this in  ae7324a.

core/commands/files.go
frrist: `d.GetType()` can be of [these types](https://github.com/ipfs/go-ipfs/blob/master/unixfs/format.go#L14) so if `HAMTSharding` was enabled we could potentially get an `unrecognized node type` error here, ya?

keks:
FWIW I never understood why files, object etc. had their own subfolders. Why shouldn't they be in the main command folder?

whyrusleeping:
Yeah, it was really just because I wanted to try and break things up into smaller bits. I'm okay with having them all in the one folder now.

kevina:
I'm okay either way.  One reason `files.go` `object.go` and `dht.go` are in there own package is because they have there own prefix in the logger.  When I moved `files.go` from its own package I needed to rename `log` to `flog`.

The disadvantage of having them in there own package is that common code will also need to be in its own package as I did in ae7324a, in case you missed it.

core/commands/files.go
MichaelMure: @whyrusleeping because we want (or at least I do) support `/ipfs/Qmwhatever`. `getNodeFromPath` will return an `ipld.Node` instead of a `mfs.FSNode`.

MichaelMure: @frrist indeed. I got that from https://github.com/ipfs/go-ipfs/blob/master/mfs/system.go#L92 where there is the same problem. Would that be a file, directory or else ?

frrist: I think they _must_ either be a file or a directory,
> Files is an API for manipulating IPFS objects as if they were a unix filesystem. - [file.go](https://github.com/ipfs/go-ipfs/blob/master/core/commands/files/files.go#L35)

The current definition looks good to me, what do ya'll think?

core/commands/files.go
```diff
@@ -86,79 +103,101 @@ var FilesStatCmd = &cmds.Command{
...
-	},
-	Marshalers: cmds.MarshalerMap{
-		cmds.Text: func(res cmds.Response) (io.Reader, error) {
-			v, err := unwrapOutput(res.Output())
-			if err != nil {
-				return nil, err
-			}
+		withLocal, _ := req.Options["with-local"].(bool)
+		if !withLocal {
+			cmds.EmitOnce(res, o)
```
whyrusleeping: you should just be able to use `res.Emit(o)`

cc @keks

whyrusleeping:
Alright, this LGTM. Will need a rebase.

Can I get a +1 from @Stebalien and @kevina ?

kevina:
@whyrusleeping @MichaelMure because of the file move you need to be careful when reserving the conflict.  Now that I worked out how to fix files.go so it can be in its own package, that might be best.

Note that I can take over the p.r. if it would help.

MichaelMure:
@whyrusleeping rebase is done.
@kevina please take over if you want to make further changes. I can't push to ipfs/go-ipfs so I added you as a collaborator on my own repo if you want to stay in the same place.

kevina:
I double checked the rebase and it looks good, the conflict was due to a dependency update which I made sure was correct.

@MichaelMure you should not need to give me access as I can push directly to your branch since you made a p.r. from it.

@whyrusleeping as I said before now that I think about it I am not sure moving `files.go` out of its directory is the best way and I have figured out to resolve the dependency cycle in ae7324a9a8ff1013218da9537b70606b1b7bbcf2.  `object.go` and `dag.go` are also in there own directory.  What ever we do we should make things uniform.

[Truth be told I am okay either way, but it seams my comments are being lost in the noise and I am not sure anyone saw by commit that fixed the cycle.  I apologize if I making things more complicated than need be.]

MichaelMure:
@whyrusleeping it does works. Good thing is that I didn't even need to change js-ipfs-api.

core/commands/files.go
Stebalien: Ideally, MFS would handle `/ipfs/` as a "mounted" subdirectory but we can fix that later.

whyrusleeping:
Thanks for putting up with our indecision @MichaelMure :)

MichaelMure:
freddy

PR #4641, opened 2018-02-02
Use a bitswap session for 'Cat'

whyrusleeping:
I'm actually not happy with how this turned out. Everything worked out fine until I had to deal with the Hamt code. Theres no distinction between reading from a hamt and writing to it. The solution i chose here was to separate that out, its a bit messy. The other approach would be to just make a hybrid `DagService` implementation that passes all Get calls to an underlying session, and then `Add` calls get routed normally. I think that method seems a lot cleaner... 

cc @Stebalien @magik6k @kevina let me know what you all think. 

License: MIT

### DIFF
--- a/merkledag/merkledag.go
+++ b/merkledag/merkledag.go
@@ -146,6 +146,10 @@ func (sg *sesGetter) GetMany(ctx context.Context, keys []*cid.Cid) <-chan *ipld.
 	return getNodesFromBG(ctx, sg.bs, keys)
 }
 
+func (ds *dagService) Session(ctx context.Context) ipld.NodeGetter {
+	return &sesGetter{bserv.NewSession(ctx, ds.Blocks)}
+}
+
 // FetchGraph fetches all nodes that are children of the given node
 func FetchGraph(ctx context.Context, root *cid.Cid, serv ipld.DAGService) error {
 	var ng ipld.NodeGetter = serv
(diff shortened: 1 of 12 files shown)


whyrusleeping:
If we get this working though, it should make `ipfs cat` and fetching things through the gateway a bit faster.

merkledag/merkledag.go
```diff
@@ -146,6 +146,15 @@ func (sg *sesGetter) GetMany(ctx context.Context, keys []*cid.Cid) <-chan *ipld.
 	return getNodesFromBG(ctx, sg.bs, keys)
 }
 
+func NewSession(ctx context.Context, dag ipld.DAGService) ipld.NodeGetter {
```
Stebalien: * This should probably take a NodeGetter.
* Docs are nice. What happens when the context is canceled? What happens to in-flight requests (esp if we just return the dag and not a session)?

Stebalien: Alternatively, we could define a new interface:

```go
type Session interface {
	Session(ctx context.Context) ipld.NodeGetter
}
func NewSession(ctx context.Context, dag ipld.NodeGetter) ipld.NodeGetter {
	if ses, ok := dag.(Session); ok {
    	return ses.Session(ctx)
    }
    return dag
}
```

Stebalien:
(except the missing documentation)

PR #4643, opened 2018-02-02
 commands: switch object to CoreAPI

magik6k:
This PR wires CoreAPI into command utils and makes `ipfs object *` use the api.

Only changing `ipfs object` to keep this PR small, will open another PRs for other command groups after this one is merged.

### DIFF
--- a/core/coreapi/coreapi.go
+++ b/core/coreapi/coreapi.go
@@ -5,6 +5,7 @@ import (
 
 	core "github.com/ipfs/go-ipfs/core"
 	coreiface "github.com/ipfs/go-ipfs/core/coreapi/interface"
+	caopts "github.com/ipfs/go-ipfs/core/coreapi/interface/options"
 	ipfspath "github.com/ipfs/go-ipfs/path"
 	uio "github.com/ipfs/go-ipfs/unixfs/io"
 
@@ -13,11 +14,12 @@ import (
 
 type CoreAPI struct {
 	node *core.IpfsNode
+	*caopts.ApiOptions
 }
 
 // NewCoreAPI creates new instance of IPFS CoreAPI backed by go-ipfs Node.
 func NewCoreAPI(n *core.IpfsNode) coreiface.CoreAPI {
-	api := &CoreAPI{n}
+	api := &CoreAPI{n, nil}
 	return api
 }
 
@@ -98,16 +100,26 @@ type path struct {
 }
 
 // ParsePath parses path `p` using ipfspath parser, returns the parsed path.
-func ParsePath(p string) (coreiface.Path, error) {
+func (api *CoreAPI) ParsePath(ctx context.Context, p string, opts ...caopts.ParsePathOption) (coreiface.Path, error) {
+	options, err := caopts.ParsePathOptions(opts...)
+	if err != nil {
+		return nil, err
+	}
+
 	pp, err := ipfspath.ParsePath(p)
 	if err != nil {
 		return nil, err
 	}
-	return &path{path: pp}, nil
+
+	res := &path{path: pp}
+	if options.Resolve {
+		return api.ResolvePath(ctx, res)
+	}
+	return res, nil
 }
 
 // ParseCid parses the path from `c`, retruns the parsed path.
-func ParseCid(c *cid.Cid) coreiface.Path {
+func (api *CoreAPI) ParseCid(c *cid.Cid) coreiface.Path {
 	return &path{path: ipfspath.FromCid(c), cid: c, root: c}
 }
 
--- a/core/coreapi/interface/interface.go
+++ b/core/coreapi/interface/interface.go
@@ -73,6 +73,16 @@ type CoreAPI interface {
 	// ResolveNode resolves the path (if not resolved already) using Unixfs
 	// resolver, gets and returns the resolved Node
 	ResolveNode(context.Context, Path) (Node, error)
+
+	// ParsePath parses string path to a Path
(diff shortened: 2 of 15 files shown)


core/coreapi/interface/interface.go
```diff
@@ -73,6 +105,16 @@ type CoreAPI interface {
 	// ResolveNode resolves the path (if not resolved already) using Unixfs
 	// resolver, gets and returns the resolved Node
 	ResolveNode(context.Context, Path) (Node, error)
+
+	// ParsePath parses string path to a Path
+	ParsePath(context.Context, string, ...options.ParsePathOption) (Path, error)
+
+	// WithResolve is an option for ParsePath which when set to true tells
+	// ParsePath to also resolve the path
+	WithResolve(bool) options.ParsePathOption
```
magik6k: I'm wondering if I should make it default to true. cc @lgierth

core/coreapi/interface/interface.go
```diff
@@ -58,6 +58,38 @@ type BlockStat interface {
...
+
+	// DiffRemove is a Type of ObjectChange where a link was removed from the graph
+	DiffRemove
+
+	// DiffMod is a Type of ObjectChange where a link was changed in the graph
+	DiffMod
+)
+
+// ObjectChange represents a change ia a graph
+// TODO: do we want this to be an interface?
```
Stebalien: TODO: Make a decision before merging (or, at least, file an issue).

core/coreapi/coreapi.go
```diff
@@ -112,16 +114,26 @@ type path struct {
 }
 
 // ParsePath parses path `p` using ipfspath parser, returns the parsed path.
-func ParsePath(p string) (coreiface.Path, error) {
+func (api *CoreAPI) ParsePath(ctx context.Context, p string, opts ...caopts.ParsePathOption) (coreiface.Path, error) {
```
Stebalien: The second this *resolves* the path, it's no longer parsing it. Maybe, instead, expose a `ResolvePathString` method (or just keep the steps separate).

core/coreapi/interface/interface.go
Stebalien: This should definitely not be a method on the CoreAPI interface (just a function, or maybe just a type `type WithResolve bool`). However, I notice we're already doing this all over the place. Any reason for it?

Anyways, do we ever actually *need* `ResolvePath` to resolve? Doesn't everything that takes a path do resolution internally just in case?

core/coreapi/coreapi.go
```diff
@@ -112,16 +114,26 @@ type path struct {
...
+	res := &path{path: pp}
+	if options.Resolve {
+		return api.ResolvePath(ctx, res)
+	}
+	return res, nil
 }
 
 // ParseCid parses the path from `c`, retruns the parsed path.
-func ParseCid(c *cid.Cid) coreiface.Path {
+func (api *CoreAPI) ParseCid(c *cid.Cid) coreiface.Path {
```
Stebalien: I really think this should remain a function.

magik6k: It's done that way to expose this function on the public API - which will be eventually implemented by things like go-ipfs-api, so we need an option to abstract this a bit.

core/coreapi/interface/interface.go
magik6k: It's based on a concept called 'functional options' from [this article](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis).

TL;DR - this pattern let's us expand the options without breaking anything, it might be overkill for some methods, but I prefer consistency more. But yeah, a separate method for resolving string path makes sense too

> Anyways, do we ever actually need ResolvePath to resolve? Doesn't everything that takes a path do resolution internally just in case?

The problem is that paths are (currently) immutable and if user wanted to re-use the resolved path (ipns can be slow), we'd need to return the resolved path, which would result in calls to some methods getting a bit messy.

magik6k: Having an explicit resolve will also make `go-ipfs-api` implementation simpler.

Stebalien: > It's based on a concept called 'functional options' from this article.

I like the pattern, I'm just objecting to attaching these functions to the interface itself instead of leaving them as free functions. It actually goes against the spirit of the pattern as it makes it *harder* to grow the API: adding new options will break existing implementations of the CoreAPI interface.

Stebalien: > The problem is that paths are (currently) immutable and if user wanted to re-use the resolved path (ipns can be slow), we'd need to return the resolved path, which would result in calls to some methods getting a bit messy.

Fair enough. However, in this case (and many others) we *know* that we're only going to use the path once (as far as I can tell).

core/coreapi/coreapi.go
Stebalien: Will there ever be alternative implementations of path parsing? I'd rather not have to implement `ParseCid` when I implement the core interface API.

magik6k:
Blocked on https://github.com/ipfs/go-ipfs/pull/4672

core/coreapi/interface/interface.go
Kubuxu: > I like the pattern, I'm just objecting to attaching these functions to the interface itself instead of leaving them as free functions.

I feel exactly the same.

magik6k:
Fixed

whyrusleeping:
rebased it since i messed it up by extracting things

PR #4649, opened 2018-02-04
blockservice: add BlockedFetched/Added/Removed events

lgierth:
... and remove a few log lines that don't seem terribly useful.

This will be useful for ipfs-repeater. The `Bitswap.BlockGetRequest` start/end events don't really cut it, they don't tell me that the block has *actually* been fetched and stored, and that the blockservice is able to provide them.

I put in TODO() contexts there, doing context wiring for Add/Delete felt out of scope, but I can do that too.

### DIFF
--- a/blockservice/blockservice.go
+++ b/blockservice/blockservice.go
@@ -138,6 +138,8 @@ func (s *blockService) AddBlock(o blocks.Block) error {
 		return err
 	}
 
+	log.Event(context.TODO(), "BlockService.BlockAdded", c)
+
 	if err := s.exchange.HasBlock(o); err != nil {
 		// TODO(#4623): really an error?
 		return errors.New("blockservice is closed")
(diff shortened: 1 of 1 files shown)


blockservice/blockservice.go
```diff
@@ -169,6 +171,7 @@ func (s *blockService) AddBlocks(bs []blocks.Block) error {
 	}
 
 	for _, o := range toput {
+		log.Event(context.TODO(), "BlockService.BlockAdded", o.Cid())
```
frrist: Nit: I like naming the events based on the method name, e.g. `BlockService.AddBlocks`.

Kubuxu: I think in this case it makes sense to have custom named events as they are mostly for functional reasons. It would also mean having 4 different events instead of 2.

whyrusleeping:
merge number two is off! Thanks @lgierth :)

PR #4653, opened 2018-02-04
document make targets

### LINKED ISSUE #4647: Better document targets in make output
From my comments on https://github.com/ipfs/go-ipfs/pull/4645:

> You might want to add test_go_test under TESTING TARGETS (when you type just make) and explain what test_go_expensive does.

Just a suggestion, only creating this issue to make sure my comment was not lost in the noise as that p.r was merged without my comment being acknowledged.

Stebalien:
Closes #4647.


test/sharness/Rules.mk
```diff
@@ -60,9 +60,6 @@ test_sharness_expensive: export TEST_EXPENSIVE=1
 test_sharness_expensive: test_sharness_short
 .PHONY: test_sharness_expensive
 
-test_sharness_race: GOFLAGS += -race
```
whyrusleeping: why is this going away?

Stebalien: It doesn't do anything.

whyrusleeping:
Can we figure out the random test failures and make issues for them? Its starting to get uncomfortable.

Stebalien:
@whyrusleeping the jenkins one looks like a jenkins reboot.

The travis one looks like an issue with error handling in the commands library: #4670.

Stebalien:
> It is unclear what if the test_go_race target runs all the tests or just the short ones. It is also not really needed. I would remove it and document to enable race detection use something like GOFLAGS=-race make test_go_expensive or if this is used with out automatic testing just document it.

Let's do that later.

PR #4658, opened 2018-02-04
shutdown notifications engine when closing a bitswap session

whyrusleeping:
License: MIT

### DIFF
--- a/exchange/bitswap/notifications/notifications.go
+++ b/exchange/bitswap/notifications/notifications.go
@@ -2,6 +2,7 @@ package notifications
 
 import (
 	"context"
+	"sync"
 
 	blocks "gx/ipfs/Qmej7nf81hi2x2tvjRBF3mcp74sQyuDH4VMYDGd1YtXjb2/go-block-format"
 
@@ -18,18 +19,33 @@ type PubSub interface {
 }
 
 func New() PubSub {
-	return &impl{*pubsub.New(bufferSize)}
+	return &impl{
+		wrapped: *pubsub.New(bufferSize),
+		cancel:  make(chan struct{}),
+	}
 }
 
 type impl struct {
 	wrapped pubsub.PubSub
+
+	// These two fields make up a shutdown "lock".
+	// We need them as calling, e.g., `Unsubscribe` after calling `Shutdown`
+	// blocks forever and fixing this in pubsub would be rather invasive.
+	cancel chan struct{}
+	wg     sync.WaitGroup
 }
 
 func (ps *impl) Publish(block blocks.Block) {
 	ps.wrapped.Pub(block, block.Cid().KeyString())
 }
 
+// Not safe to call more than once.
 func (ps *impl) Shutdown() {
+	// Interrupt in-progress subscriptions.
+	close(ps.cancel)
+	// Wait for them to finish.
+	ps.wg.Wait()
+	// shutdown the pubsub.
 	ps.wrapped.Shutdown()
 }
 
@@ -44,12 +60,34 @@ func (ps *impl) Subscribe(ctx context.Context, keys ...*cid.Cid) <-chan blocks.B
 		close(blocksCh)
 		return blocksCh
 	}
+
+	// prevent shutdown
+	ps.wg.Add(1)
+
+	// check if shutdown *after* preventing shutdowns.
+	select {
+	case <-ps.cancel:
+		// abort, allow shutdown to continue.
+		ps.wg.Done()
+		close(blocksCh)
+		return blocksCh
+	default:
+	}
+
 	ps.wrapped.AddSubOnceEach(valuesCh, toStrings(keys)...)
 	go func() {
-		defer close(blocksCh)
-		defer ps.wrapped.Unsub(valuesCh) // with a len(keys) buffer, this is an optimization
+		defer func() {
+			ps.wrapped.Unsub(valuesCh)
+			close(blocksCh)
+
+			// Unblock shutdown.
+			ps.wg.Done()
+		}()
+
 		for {
 			select {
+			case <-ps.cancel:
+				return
 			case <-ctx.Done():
 				return
 			case val, ok := <-valuesCh:
@@ -61,6 +99,8 @@ func (ps *impl) Subscribe(ctx context.Context, keys ...*cid.Cid) <-chan blocks.B
 					return
 				}
 				select {
+				case <-ps.cancel:
+					return
 				case <-ctx.Done():
 					return
 				case blocksCh <- block: // continue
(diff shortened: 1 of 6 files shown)


whyrusleeping:
very strange test failure...

whyrusleeping:
not strange, i just wasnt paying attention.

whyrusleeping:
@Stebalien I think my second commit is the right fix for the wantlist thing, but the test doesnt actually reproduce the issue. I think its because there are a couple different places the wantlist is accounted for, and bitswap.Wantlist is not the same as the wantmanager where the sessions track their wants.

exchange/bitswap/session.go
```diff
@@ -83,6 +83,15 @@ func (bs *Bitswap) NewSession(ctx context.Context) *Session {
...
 
 func (bs *Bitswap) removeSession(s *Session) {
+	s.notif.Shutdown()
+
+	live := make([]*cid.Cid, 0, len(s.liveWants))
+	for c := range s.liveWants {
+		cs, _ := cid.Cast([]byte(c))
+		live = append(live, cs)
+	}
+	bs.CancelWants(live, s.id)
```
magik6k: What happens if another session wants a cid which is cancelled here?

whyrusleeping: Wants are tracked per session (note that we pass in the session ID)

Stebalien:
Are you not running into the issue I have here: https://github.com/ipfs/go-ipfs/pull/4659#issuecomment-363140828? It looks like unsubscribing after the session is shut down causes the unsubscribe function to hang (looks like a bug in the pubsub library we're using).

Actually, we should probably just replace that library (https://github.com/briantigerchow/pubsub). It's *way* too complicated for what it does (and we don't actually need to spawn a goroutine like that).

whyrusleeping:
@Stebalien I didn't notice any issues, all the tests passed when i ran them. And i'm wary of replacing a library that has caused us zero problems in three years.

Stebalien:
Calling Unsub after Shutdown will hang forever trying to write to the cmds channel. This patch has the same bug as mine. Try downloading something you don't have from the gateway, cancel, and then run:

```
> curl 'http://localhost:5001/debug/pprof/goroutine?debug=2' | grep 'pubsub.*Unsub'
```

> And i'm wary of replacing a library that has caused us zero problems in three years.

My primary motivation was getting rid of the goroutine but we need goroutines *anyways* (one per subscription, actually) to make our contexts work... However, we still need to fix the unsubscribe issue (and the easiest way would be to at least modify this library).

Stebalien:
@whyrusleeping off the top of your head, do you recall why we have separate notification engines per session instead of just using the main one?

whyrusleeping:
@Stebalien I actually don't remember off the top of my head. I think it was something around me thinking that multiple different subscriptions wouldnt play nicely. But they should... I'd say try it. It seems like it could work.

Stebalien:
@whyrusleeping so, I fixed the unsubscribe after shutdown deadlock by just not unsubscribing after shutting down. I'm not happy with the fix but cleanly fixing the pubsub library didn't seem possible either due to its API (e.g., when a user calls `AddSub(ch, topics...)` on a pubsub instance after shutting it down, should we close `ch`? That could panic if we had already registered it and closed it).

I figured if we couldn't have a nice fix, we might as well have a small fix internally and avoid maintaining a fork.

exchange/bitswap/notifications/notifications.go
```diff
@@ -18,18 +19,33 @@ type PubSub interface {
...
 func (ps *impl) Publish(block blocks.Block) {
 	ps.wrapped.Pub(block, block.Cid().KeyString())
 }
 
+// Not safe to call more than once.
 func (ps *impl) Shutdown() {
+	// Interrupt in-progress subscriptions.
+	close(ps.cancel)
+	// Wait for them to finish.
+	ps.wg.Wait()
```
whyrusleeping: This will wait for all active wants to be cancelled, which happens if the caller closes the session, right? I would like to see a test around this.

Stebalien: No, it'll just wait for all unsubscribes to finish (which should happen immediately after we close the `ps.cancel` channel).

However, I have added a test to ensure that shutting down the PubSub while a subscription is active works (and doesn't block as it did before). Is that what you're looking for?

Stebalien:
Real test failures:

```
--- FAIL: TestMultipleSessions (20.01s)
	session_test.go:284: bad juju
--- FAIL: TestWantlistClearsOnCancel (0.00s)
	session_test.go:318: expected empty wantlist
```

Stebalien:
So that I don't forget where I left off...

Canceling wants for one session sometimes prevents the other session from receiving its requested block. So far, I've narrowed this down to the `mq.out.Cancel(e.Cid)` line in the `addMessage` function of `msgQueue` in `wantmanager.go`. Commenting that line out fixes it.

whyrusleeping:
@Stebalien status here?

Stebalien:
@whyrusleeping still debugging.

Stebalien:
Well... *that's* an old bug. Unfortunately, that fix *still* doesn't fix everything. `go test -v -count 100 -run TestMultipleSessions` hangs after ~20 runs (better than the 1-5 before...).

whyrusleeping:
very weird behaviour here. Putting a sleep after `cancel1()` causes the issue to no longer happen. But putting the `cancel1()` *after* the next `GetBlocks` call also causes the issue to no longer happen.

whyrusleeping:
Putting the `cancel1()` after the `NewSession` call still allows me to reproduce the issue.

whyrusleeping:
Okay, looking at wantlist messages being sent from peer A to peer B. In success cases we have one of:
- A -> B: Want X
or
- A -> B: Want X
- A -> B: Want X
or
- A -> B: Want X
- A -> B: Cancel X
- A -> B: Want X

But in the failure case, I'm always seeing:
- A -> B: Want X
- A -> B: Want X
- A -> B: Cancel X

Which makes sense why it hangs, we cancelled our request. The other peer is respecting that. Now the question is, what makes us cancel?

whyrusleeping:
So with the above pattern, `addMessage` gets called three times:
- Want X (session 1)
- Cancel X (session 1)
- Want X (session 2)

This should result in three messages getting sent, a want, a cancel, and a want. My only thought now is that the messages are getting reordered in transit somehow.

whyrusleeping:
Oh look: https://github.com/ipfs/go-ipfs/blob/master/exchange/bitswap/testnet/virtual.go#L80

The messages are delivered by throwing them off into a goroutine...

whyrusleeping:
adding a random delay to every message send causes the bug to be reproduced ~30% of the time.

Issue #4657, opened 2018-02-04
Goroutine leak and wantlist build-up on master


lgierth:
#### Description:

Right after applying 52301ce8 (latest master) to the ipfs.io gateways, there's a severe leak of `briantigerchow/pubsub.(*PubSub).start` goroutines, and wantlist entries start piling up. Within 18 minutes on scrappy, there were 78k goroutines and 1.9k wantlist entries. (It was still able to fetch a simple `date | ipfs add` from my own node, so it wasn't completely locked up.)

- Diff: https://github.com/ipfs/go-ipfs/compare/3f2c7746...52301ce8
- pprof: https://ipfs.io/ipfs/QmZkk8XwFJopBkK8AZxMVXbHk3MjxkxGBZPrUQmK6TdCKu
- Graphs: http://metrics.ipfs.team/dashboard/db/ipfs-v04x?refresh=30s&orgId=1&from=1517776089172&to=1517779689172

For now I've reverted back to 3f2c7746.

Stebalien:
Trivial reproduction: cat a file on the local gateway (sessions still aren't used for the `cat` command because that doesn't use the same cat function...).

Stebalien:
Ok. *Now* bitswap hangs trying to unsubscribe...

Stebalien:
Fixed in https://github.com/ipfs/go-ipfs/pull/4658.

Also, we've disabled bitswap sessions for the gateway (for other performance reasons, for now).

PR #4659, opened 2018-02-05
[WIP] fix leaks on ipfs cat

Stebalien:
WIP fix for #4657

Still need to validate that this fixes it. Current fixes:

* Correctly (I think?) shutdown pubsub channel when shutting down a session.
  Should stop the goroutine leak.

Missing: explanation for the wantlist issue.

### DIFF
--- a/exchange/bitswap/session.go
+++ b/exchange/bitswap/session.go
@@ -163,8 +163,11 @@ func (s *Session) resetTick() {
 }
 
 func (s *Session) run(ctx context.Context) {
+	defer s.notif.Shutdown()
+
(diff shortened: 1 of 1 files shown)


Stebalien:
Ok. Now bitswap hangs trying to unsubscribe...

whyrusleeping:
uhm. take a look at 4658... >.>

whyrusleeping:
I'm also missing an explanation for the wantlists.

whyrusleeping:
Oh, its that we're not clearing out the livewants from the wantlist

PR #4661, opened 2018-02-05
Extract: importers/chunk module as go-ipfs-chunker

hsanjuan:
This extracts `chunk` to https://github.com/ipfs/go-ipfs-chunker

This is part of an effort to extract `importer` from go-ipfs so they can be re-used by ipfs-cluster with more liberty (and without having to import go-ipfs itself).

Before I leave everything ready to merge, is the naming `go-ipfs-chunker` ok? It might be worth to go all the way and rename the whole thing and references to `go-ipfs-splitter`, or to rename the `Splitter` interface to `Chunker`.


Stebalien:
@hsanjuan please preserve git history when possible (in go-ipfs-chunker). I know we have directions for how to do this somewhere but I can't find them (@whyrusleeping?).

whyrusleeping:
They are here: https://github.com/libp2p/go-libp2p#modularizing-go-libp2p

hsanjuan:
Thanks, that was useful.

I have:

* Imported history to go-ipfs-chunker
* Added README
* Rebased this PR and resolved conflicts
* Point it to the latest go-ipfs-chunker gx release (which is pinned in mars)
* Re-ordered the imports to be consistent and rename `chunk` to `chunker` for the same reason.

PR #4662, opened 2018-02-05
Fix some tests on FreeBSD

magik6k:


test/sharness/t0086-repo-verify.sh
```diff
@@ -15,6 +15,9 @@ sort_rand() {
     Linux)
       sort -R
       ;;
+    FreeBSD)
```
Stebalien: Better just to use `Linux|FreeBSD`.

PR #4664, opened 2018-02-06
Golint-ify unixfs module

hsanjuan:
This resolves all golint warnings in `unixfs`. `hamt.HamtShard` has been renamed to `hamt.Shard` to remove stuttering.

Note that protobuf generated files are not golint compliant, but haven't touched those.


unixfs/io/dirbuilder.go
```diff
@@ -157,6 +162,9 @@ func (d *Directory) Links(ctx context.Context) ([]*ipld.Link, error) {
 	return d.shard.EnumLinks(ctx)
 }
 
+// Find returns the ipld.Node with the given name, if it is contained in this
+// directory. Find only searches in the most inmediate links, and not
+// recursively in the tree.
 func (d *Directory) Find(ctx context.Context, name string) (ipld.Node, error) {
```
Stebalien: Find returns the root node of the file named `name` within this directory (really, it should be called `Lookup`). It does, in fact, traverse the tree (when searching a hamt).

unixfs/mod/dagmodifier.go
```diff
@@ -19,8 +21,12 @@ import (
 	ipld "gx/ipfs/Qme5bWv7wtjUNGsK2BNGVUFPKiuxWrsqrtvYwCLRw8YFES/go-ipld-format"
 )
 
-var ErrSeekFail = errors.New("failed to seek properly")
-var ErrUnrecognizedWhence = errors.New("unrecognized whence")
+// Common errors
+var (
+	ErrSeekFail           = errors.New("failed to seek properly")
+	ErrUnrecognizedWhence = errors.New("unrecognized whence")
+	ErrNotUnixfs          = fmt.Errorf("dagmodifier only supports unixfs nodes (proto or raw)")
```
Stebalien: Might as well be consistent and use the `errors` package.

unixfs/io/bufdagreader.go
```diff
@@ -10,7 +10,9 @@ type bufDagReader struct {
 	*bytes.Reader
 }
 
-func NewBufDagReader(b []byte) *bufDagReader {
+// newBufDagReader returns a DAG reader for the given byte slice.
+// BufDagReader is used to read RawNodes.
+func newBufDagReader(b []byte) *bufDagReader {
```
Stebalien: I'd just make `bufDatReader` public.

unixfs/mod/dagmodifier.go
hsanjuan: out of scope I'd say

unixfs/io/bufdagreader.go
hsanjuan: So between increasing the API surface and decreasing it, I went with the latter (since this method is not used by anyone). But I'll make it public then...

unixfs/mod/dagmodifier.go
hsanjuan: ah, yeah, got it now

unixfs/io/bufdagreader.go
Stebalien: Yeah, but you exposed the `PBDagReader` so I figured we should be consistent at least.

unixfs/io/dagreader.go
```diff
@@ -14,17 +14,23 @@ import (
...
-var ErrIsDir = errors.New("this dag node is a directory")
-
-var ErrCantReadSymlinks = errors.New("cannot currently read symlinks")
+// Common errors
+var (
+	ErrIsDir            = errors.New("this dag node is a directory")
+	ErrCantReadSymlinks = errors.New("cannot currently read symlinks")
+)
 
+// A DagReader represents a ReadSeekCloser which offers additional methods
```
whyrusleeping: > represents a ReadSeekCloser which offers additional methods

This doesnt really get at what the dagreader *does*. I would say something more like:

```
// A DagReader provides read-only read and seek acess to a unixfs file.
```

unixfs/io/dagreader.go
```diff
@@ -14,17 +14,23 @@ import (
...
+// like Size. Different implementations of readers are used for the different
+// types of unixfs/protobuf-encoded nodes.
 type DagReader interface {
 	ReadSeekCloser
 	Size() uint64
 	CtxReadFull(context.Context, []byte) (int, error)
 	Offset() int64
 }
 
+// A ReadSeekCloser implements interfaces to read, write, seek and close.
```
whyrusleeping: 'write' doesnt belong

unixfs/io/dagreader.go
hsanjuan: Thanks

unixfs/io/dagreader.go
hsanjuan: right, more `copy`

hsanjuan:
all comments addressed

PR #4665, opened 2018-02-06
Golint: fix golint warnings in merkledag submodule

hsanjuan:
This fixes all golint warnings in merkledag (except in protobuf generated code).

Main change of note is that `NewDAGService` returns an `ipld.DAGService` and not a non-exported `dagService`.

### DIFF
--- a/merkledag/errservice.go
+++ b/merkledag/errservice.go
@@ -14,34 +14,34 @@ type ErrorService struct {
 
 var _ ipld.DAGService = (*ErrorService)(nil)
 
-// Add returns an error.
+// Add returns the cs.Err.
 func (cs *ErrorService) Add(ctx context.Context, nd ipld.Node) error {
 	return cs.Err
 }
 
-// AddMany returns an error.
+// AddMany returns the cs.Err.
 func (cs *ErrorService) AddMany(ctx context.Context, nds []ipld.Node) error {
 	return cs.Err
 }
 
-// Get returns an error.
+// Get returns the cs.Err.
 func (cs *ErrorService) Get(ctx context.Context, c *cid.Cid) (ipld.Node, error) {
 	return nil, cs.Err
 }
 
-// GetMany many returns an error.
+// GetMany many returns the cs.Err.
 func (cs *ErrorService) GetMany(ctx context.Context, cids []*cid.Cid) <-chan *ipld.NodeOption {
 	ch := make(chan *ipld.NodeOption)
 	close(ch)
 	return ch
 }
 
-// Remove returns an error.
+// Remove returns the cs.Err.
 func (cs *ErrorService) Remove(ctx context.Context, c *cid.Cid) error {
 	return cs.Err
 }
 
-// RemoveMany returns an error.
+// RemoveMany returns the cs.Err.
 func (cs *ErrorService) RemoveMany(ctx context.Context, cids []*cid.Cid) error {
 	return cs.Err
 }
--- a/merkledag/errservice.go
+++ b/merkledag/errservice.go
@@ -14,28 +14,34 @@ type ErrorService struct {
 
 var _ ipld.DAGService = (*ErrorService)(nil)
 
+// Add returns an error.
 func (cs *ErrorService) Add(ctx context.Context, nd ipld.Node) error {
 	return cs.Err
 }
 
+// AddMany returns an error.
 func (cs *ErrorService) AddMany(ctx context.Context, nds []ipld.Node) error {
 	return cs.Err
 }
 
+// Get returns an error.
 func (cs *ErrorService) Get(ctx context.Context, c *cid.Cid) (ipld.Node, error) {
 	return nil, cs.Err
 }
 
+// GetMany many returns an error.
 func (cs *ErrorService) GetMany(ctx context.Context, cids []*cid.Cid) <-chan *ipld.NodeOption {
 	ch := make(chan *ipld.NodeOption)
 	close(ch)
 	return ch
 }
 
+// Remove returns an error.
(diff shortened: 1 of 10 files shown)


merkledag/merkledag.go
```diff
@@ -23,8 +23,14 @@ func init() {
 	ipld.Register(cid.DagCBOR, ipldcbor.DecodeBlock)
 }
 
+// contextKey is a type to use as value for the ProgressTracker contexts.
+type contextKey string
+
+const progressContextKey contextKey = "progress"
+
 // NewDAGService constructs a new DAGService (using the default implementation).
-func NewDAGService(bs bserv.BlockService) *dagService {
```
whyrusleeping: Does golint really complain about this?

I actually prefer to return concrete types whenever possible. It helps to deal with multiple potentially fulfilled interfaces. For example, if `*dagService` implements `X` which is a superset of `DAGService`, I can pass `NewDAGService` to something that wants an `X` without extra type assertions.

merkledag/merkledag_test.go
```diff
@@ -241,9 +242,10 @@ func TestFetchGraph(t *testing.T) {
 	// create an offline dagstore and ensure all blocks were fetched
 	bs := bserv.New(bsis[1].Blockstore(), offline.Exchange(bsis[1].Blockstore()))
 
-	offline_ds := NewDAGService(bs)
+	// we know the default dagService implements LinkGetter
+	offlineDS := NewDAGService(bs).(ipld.LinkGetter)
```
whyrusleeping: This is exactly the case for having `NewDAGService` return the concrete type.

whyrusleeping:
Thank you!! 

Everything except the DAGService return type LGTM... If golint really complains about that it makes me a tad bit sad.

merkledag/merkledag_test.go
hsanjuan: But it is a case only used in tests (and tests in the same module). From an external user's point of view, it's not possible to know what *dagService can do because it's unexported.

merkledag/merkledag.go
hsanjuan: Want me to export `dagService` instead?

kevina:
> Everything except the DAGService return type LGTM... If golint really complains about that it makes me a tad bit sad.

I agree and I want to emphasis that we don't have to follow all off golints suggestions.  If this will cause codeclimate to complain we should add a suggestion.

Stebalien:
> Everything except the DAGService return type LGTM... If golint really complains about that it makes me a tad bit sad.

It does this for a (possibly debatable) reason. The lint disallows returning unexported types because they can't be named outside the package.

hsanjuan:
Ok, I have reverted to what it was before and get the expected:

`merkledag.go:33:43: exported func NewDAGService returns unexported type *merkledag.dagService, which can be annoying to use`

If it is valuable that this is used as a `LinkGetter` (which is not except in tests), I would lean to exporting it though. The user won't have a clue that this returns a `LinkGetter` without looking at sources otherwise.  It's a nuance anyway, I'm fine with whatever.

merkledag/errservice.go
```diff
@@ -14,28 +14,34 @@ type ErrorService struct {
 
 var _ ipld.DAGService = (*ErrorService)(nil)
 
+// Add returns an error.
```
kevina: I find this description confusing, rather than say "an error" something like "returns ErrorService.Err" or maybe just "returns _the_ error" rather than "_an_ error".

merkledag/errservice.go
```diff
@@ -14,28 +14,34 @@ type ErrorService struct {
...
 func (cs *ErrorService) Add(ctx context.Context, nd ipld.Node) error {
 	return cs.Err
 }
 
+// AddMany returns an error.
```
kevina: Ditto.

merkledag/errservice.go
```diff
@@ -14,28 +14,34 @@ type ErrorService struct {
...
 func (cs *ErrorService) AddMany(ctx context.Context, nds []ipld.Node) error {
 	return cs.Err
 }
 
+// Get returns an error.
```
kevina: Etc.

merkledag/merkledag.go
```diff
@@ -23,7 +23,13 @@ func init() {
 	ipld.Register(cid.DagCBOR, ipldcbor.DecodeBlock)
 }
 
+// contextKey is a type to use as value for the ProgressTracker contexts.
+type contextKey string
```
kevina: I would export this type.

hsanjuan: It has no use outside. If someone ever has a use, it can be exported then.

kevina: Yeah, your right.  Sorry.

Stebalien:
FYI,

> If it is valuable that this is used as a LinkGetter (which is not except in tests), I would lean to exporting it though. The user won't have a clue that this returns a LinkGetter without looking at sources otherwise. It's a nuance anyway, I'm fine with whatever.

In general, users shouldn't care if something is a `LinkGetter`. They should just call `ipld.GetLinks(ctx, dag, cid)` which will, internally, try to cast `dag` to a `LinkGetter` if the passed dag implements that interface.

merkledag/merkledag_test.go
```diff
@@ -13,6 +13,8 @@ import (
 	"testing"
 	"time"
 
+	blocks "gx/ipfs/Qmej7nf81hi2x2tvjRBF3mcp74sQyuDH4VMYDGd1YtXjb2/go-block-format"
```
Stebalien: This should be grouped with `cid`, `u`, etc. (we group by standard, in-repo, then out-of-repo).

hsanjuan:
opps. fixed @Stebalien

PR #4672, opened 2018-02-08
Improve paths in CoreApi

### LINKED ISSUE #4666: Think through the CoreAPI Path API and implementation.
* The resolver seems to assume that paths *always* resolve to CIDs (can't end up in the middle of an object).
* We should provide a function for retrieving the namespace.
* It should be possible to distinguish between mutable and immutable paths.
* `coreapi.ParsePath` doesn't actually extract the root.

magik6k:
Fixes #4666 

TODO:
* [x] Path remainders
* [x] Mutable vs Immutable paths
  * Just `path.Namespace() == "ipns"` or something more complex?
* [x] More things from #4666


Stebalien:
Responded to the TODOs here: https://github.com/ipfs/go-ipfs/issues/4666#issuecomment-364239134. However, all that will take a while...

magik6k:
I imported suggestions from #4666, @Stebalien can you have a look? (start from interface changes)

core/coreapi/interface/coreapi.go
```diff
@@ -32,9 +33,18 @@ type CoreAPI interface {
...
 	// ResolvePath resolves the path using Unixfs resolver
-	ResolvePath(context.Context, Path) (Path, error)
+	ResolvePath(context.Context, Path) (ResolvedPath, error)
 
 	// ResolveNode resolves the path (if not resolved already) using Unixfs
 	// resolver, gets and returns the resolved Node
 	ResolveNode(context.Context, Path) (ipld.Node, error)
+
+	// ParsePath parses string path to a Path
+	ParsePath(string) (Path, error)
```
Stebalien: I'm not really a fan of attaching helper functions to instance objects. What's the motivation?

magik6k: > What's the motivation?

We need `ResolvePath` on the interface as it is quite implementation specific, I thrown the other methods here for consistency.

Kubuxu:
@magik6k it looks like test coverage of core api is low in general. Could you work on this in future?

core/coreapi/path.go
```diff
@@ -0,0 +1,154 @@
...
+
+	return &path{path: pp}, nil
+}
+
+func (p *path) String() string {
+	return p.path.String()
+}
+
+func (p *path) Namespace() string {
+	if len(p.path.Segments()) < 1 {
```
Stebalien: Won't *all* paths have namespaces now? It would be nice to check this (and parse out the namespace) on initial parse.

magik6k: > Won't all paths have namespaces now?

They do, `ipfspath.ParsePath` guarantees that.

core/coreapi/path.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+	return node, nil
+}
+
+func resolvePath(ctx context.Context, ng ipld.NodeGetter, nsys namesys.NameSystem, p coreiface.Path) (coreiface.ResolvedPath, error) {
+	if _, ok := p.(coreiface.ResolvedPath); ok {
+		return p.(coreiface.ResolvedPath), nil
+	}
+
+	ipath := p.(*path).path
```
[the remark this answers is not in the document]
magik6k: This needs to be here to get to internal `ipfspath.Path`, at least for now.

I'll deduplicate logic from here and ipfspath in a separate PR, doing so in this PR would make it way bigger than it needs to be.

core/coreapi/interface/path.go
```diff
@@ -6,13 +6,37 @@ import (
...
+	// Mutable returns false if the data pointed to by this path in guaranteed
+	// to not change.
+	//
+	// Note that resolved mutable path can be immutable.
+	Mutable() bool
+}
+
+// ResolvedPath is a resolved Path
+type ResolvedPath interface {
+	// Cid returns the CID referred to by path
```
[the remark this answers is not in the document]
magik6k: Added some docs

core/coreapi/path.go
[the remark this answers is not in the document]
Stebalien: Got it.

core/coreapi/path.go
Stebalien: Could we panic here?

magik6k: I pushed just that: https://github.com/ipfs/go-ipfs/pull/4672/files#diff-aec5b488acbd54adbfbc67351fc51d7aR140

magik6k:
I've added some more tests and fixed rest of the stuff that needed fixing

magik6k:
@Stebalien @Kubuxu can you have a look at thin PR, would be great to have this move (it blocks most of coreapi-related efforts)

magik6k:
On `PathToCid` - Since it would return an error and probably need to be declared on the interface, I'm not sure what is the value added other than not having to type `.Cid()` in some places (note that cids won't be used that much outside the implementation code)

Stebalien:
> On PathToCid - Since it would return an error and probably need to be declared on the interface, I'm not sure what is the value added other than not having to type .Cid() in some places (note that cids won't be used that much outside the implementation code)

I agree (actually, I can't remember why I suggested it in the first place).

Kubuxu:
Interface SGWM.

magik6k:
@whyrusleeping can we try to get this in after 0.4.16 release stuff is done?

whyrusleeping:
@magik6k Yes! Getting the other coreAPI work unblocked is high priority. Sorry for the lag

whyrusleeping:
@magik6k alright, mind rebasing this? This will be the next thing we merge

magik6k:
done

PR #4678, opened 2018-02-10
build: match Go's GOPATH defaults behaviour

lgierth:
This came up while setting up a new system, and trying to do Go things without explicitly setting GOPATH.

### DIFF
--- a/bin/check_go_path
+++ b/bin/check_go_path
@@ -7,10 +7,7 @@ if [ -z "$PWD" ]; then
 	exit 1
 fi
 
-if [ -z "$GOPATH" ]; then
-	echo "GOPATH not set, you must have go configured properly to install ipfs"
-	exit 1
-fi
+[ -z "$GOPATH" ] && GOPATH="$HOME/go"
 
 while [ ${#} -gt 1 ]; do
 	if [ "$PWD" = "$2" ]; then
--- a/bin/check_go_path
+++ b/bin/check_go_path
@@ -7,8 +7,6 @@ if [ -z "$PWD" ]; then
 	exit 1
 fi
 
-[ -z "$GOPATH" ] && GOPATH="$HOME/go"
-
 while [ ${#} -gt 1 ]; do
 	if [ "$PWD" = "$2" ]; then
 		exit 0
--- a/bin/check_go_path
+++ b/bin/check_go_path
@@ -15,5 +15,5 @@ while [ ${#} -gt 1 ]; do
 done
 
 echo "go-ipfs must be built from within your \$GOPATH directory."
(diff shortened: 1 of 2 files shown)


aerth:
if go assumed to be installed, the output of the `go env GOPATH` command should be safe on all platforms.

mk/golang.mk
```diff
@@ -1,6 +1,9 @@
 # golang utilities
 GO_MIN_VERSION = 1.9
 
+# match Go's default GOPATH behaviour
+GOPATH ?= $(HOME)/go
```
Kubuxu: I would replace it with `export GOPATH := $(shell $(GOCC) env GOPATH)`

I am not 100% sure it works but it should.

bin/check_go_path
```diff
@@ -7,10 +7,7 @@ if [ -z "$PWD" ]; then
 	exit 1
 fi
 
-if [ -z "$GOPATH" ]; then
-	echo "GOPATH not set, you must have go configured properly to install ipfs"
-	exit 1
-fi
+[ -z "$GOPATH" ] && GOPATH="$HOME/go"
```
Kubuxu: Witch changes to Makefile this should run without changes.

bin/check_go_path
```diff
@@ -7,11 +7,6 @@ if [ -z "$PWD" ]; then
 	exit 1
 fi
 
-if [ -z "$GOPATH" ]; then
-	echo "GOPATH not set, you must have go configured properly to install ipfs"
-	exit 1
-fi
-
```
magik6k: I'd still keep this check (assuming it didn't break, which it shouldn't) so if something goes wrong it's caught.

Kubuxu: I disagree, we are already passing `GOPATHs` as arguments to this function (as `$2`, `$3` and so on). So this check is IMO not needed.

lgierth: GOPATH doesn't have to be set anymore. I'm actually not sure why the `export` in golang.mk doesn't carry over.

Kubuxu: It does, I just wanted to make this script be runnable semi stand alone.

mk/golang.mk
```diff
@@ -1,6 +1,9 @@
 # golang utilities
 GO_MIN_VERSION = 1.9
 
+# match Go's default GOPATH behaviour
+export GOPATH ?= $(shell $(GOCC) env GOPATH)
```
whyrusleeping: wait a sec, doesnt this need to go below the definition for `GOCC`?

Kubuxu:
Fixed

lgierth:
Thanks for shipping this for me @Kubuxu :heart:

PR #4680, opened 2018-02-11
Validate size in the DagReaders

### LINKED ISSUE #4540: Validate size in the DagReaders
When reading files, we should always validate the size (it doesn't look like it currently does). If we ever hit a chunk that's too large, we should truncate it. If we ever hit a chunk that's too small, we should return zeros (allowing holes in files is actually quite useful). We should also document this in a spec somewhere.

Expected size not matching the actual size of something tends to lead to security problems...

@diasdavid objections? How does js-ipfs deal with this issue?

### LINKED ISSUE #4667: Error: archive/tar: write too long
#### Version information:
go-ipfs version: 0.4.13-
Repo version: 6
System version: amd64/windows
Golang version: go1.9.2

#### Description:
/ipfs/QmWfLgAopxBBpnPsMeQWQhcVfHHEo7natWgtcX5yzw7My3

When get this file from browser:
`http://127.0.0.1/ipfs/QmWfLgAopxBBpnPsMeQWQhcVfHHEo7natWgtcX5yzw7My3 200 Ok`

Result is valid tar archive: [test tar.tar.gz](https://github.com/ipfs/go-ipfs/files/1700847/test.tar.tar.gz) (it gziped for upload to GitHub)
Blocks of this file: [blocks.tar.gz](https://github.com/ipfs/go-ipfs/files/1700817/blocks.tar.gz)

When get from console:
```
>ipfs get QmWfLgAopxBBpnPsMeQWQhcVfHHEo7natWgtcX5yzw7My3
Saving file(s) to QmWfLgAopxBBpnPsMeQWQhcVfHHEo7natWgtcX5yzw7My3
 2.28 KB / 2.28 KB [==============================================] 100.00% 0s
Error: archive/tar: write too long
```
Why "ipfs get" try to read tar?

#### Additional information:

```
>ipfs dag get QmWfLgAopxBBpnPsMeQWQhcVfHHEo7natWgtcX5yzw7My3
{"data":"CAIYgBAggAgggAQggAQ=","links":[{"Cid":{"/":"QmbTswUf339CVJsqB6rSeNum7hrnUsNqFbgkth8g3u2chX"},"Name":"","Size":1165},{"Cid":{"/":"zb2rhX9DvypVUCrp1SyoB6JpvdNEGcujJm7RZ6btt1zJBvnKq"},"Name":"","Size":512},{"Cid":{"/":"zb2rhX9DvypVUCrp1SyoB6JpvdNEGcujJm7RZ6btt1zJBvnKq"},"Name":"","Size":512}]}
```
decode of data field:
```
"data":{
  "Type": 2,           // It's file
  "filesize": 2048,  // sum of blocksizes
  "blocksizes": [
    1024,               // file header size + file data size + file padding size
    512,                 // padding block size
    512                  // padding block size
  ]
}
```

```
>ipfs dag get QmbTswUf339CVJsqB6rSeNum7hrnUsNqFbgkth8g3u2chX
{"data":"CAISgAR0ZXN0IHRhci50eHQAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAMDEwMDc3NwAwMDAwMDAwADAwMDAwMDAAMDAwMDAwMDAwNDQAMTMyMzY0MDYzMTAAMDExNTM3ACAwAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAHVzdGFyADAwAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAADAwMDAwMDAAMDAwMDAwMAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAABiACCAkINwD","links":[{"Cid":{"/":"zb2rhfY128Mz2bimXoxi7aZuc5NPQMWaLPJYVSTLoYWs7rsdz"},"Name":"","Size":36},{"Cid":{"/":"zb2rhX9DvypVUCrp1SyoB6JpvdNEGcujJm7RZ6btt1zJBvnKq"},"Name":"","Size":512}]}
```
decode of data field:
```
data: {
  "Type": 2,                      // It's file
  "Data": "test tar.txt...",   // 512 bytes tar file header
  "filesize": 1024,             // file header size + file data size + file padding size
  "blocksizes": [
    36,                              // file data size
    476                             // file padding size (it's hack)
  ]
}
```

[zb2rhfY128Mz2bimXoxi7aZuc5NPQMWaLPJYVSTLoYWs7rsdz](https://github.com/ipfs/go-ipfs/files/1701076/zb2rhfY128Mz2bimXoxi7aZuc5NPQMWaLPJYVSTLoYWs7rsdz.gz) (36) - file data raw block
```
>ipfs block get zb2rhfY128Mz2bimXoxi7aZuc5NPQMWaLPJYVSTLoYWs7rsdz
This is "test tar.txt" file content.
```

[zb2rhX9DvypVUCrp1SyoB6JpvdNEGcujJm7RZ6btt1zJBvnKq](https://github.com/ipfs/go-ipfs/files/1701078/zb2rhX9DvypVUCrp1SyoB6JpvdNEGcujJm7RZ6btt1zJBvnKq.gz) (512) - padding raw block
```
> ipfs block get zb2rhX9DvypVUCrp1SyoB6JpvdNEGcujJm7RZ6btt1zJBvnKq
```

kevina:
If a node is too large, it is truncated; if it is too small, it is zero extended.

Closes #4540.  Closes #4667.

### DIFF
--- a/test/sharness/t0110-gateway.sh
+++ b/test/sharness/t0110-gateway.sh
@@ -188,7 +188,7 @@ test_expect_success "Add compact blocks" '
   printf "foofoo" > expected
 '
 
-test_expect_success "GET compact blocks succeeds" '
+test_expect_failure "GET compact blocks succeeds" '
   curl -o actual "http://127.0.0.1:$port/ipfs/$FOO2_HASH" &&
   test_cmp expected actual
 '
--- a/test/sharness/t0110-gateway.sh
+++ b/test/sharness/t0110-gateway.sh
@@ -188,7 +188,7 @@ test_expect_success "Add compact blocks" '
   printf "foofoo" > expected
 '
 
-test_expect_failure "GET compact blocks succeeds" '
+test_expect_success "GET compact blocks succeeds" '
   curl -o actual "http://127.0.0.1:$port/ipfs/$FOO2_HASH" &&
   test_cmp expected actual
 '
--- a/test/sharness/t0110-gateway.sh
+++ b/test/sharness/t0110-gateway.sh
@@ -193,6 +193,12 @@ test_expect_success "GET compact blocks succeeds" '
   test_cmp expected actual
 '
 
+test_expect_failure "GET compact blocks succeeds with seek" '
+  curl -I "http://127.0.0.1:$port/ipfs/$FOO2_HASH" &&
+  curl -v -o actual -H "Range:bytes=3-" "http://127.0.0.1:$port/ipfs/$FOO2_HASH" &&
+  test_cmp expected actual
+'
+
 test_kill_ipfs_daemon
 
 test_done
--- a/unixfs/io/pbdagreader.go
+++ b/unixfs/io/pbdagreader.go
@@ -100,8 +100,8 @@ func (dr *PBDagReader) precalcNextBuf(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
-	blocksize := dr.pbdata.Blocksizes[dr.linkPosition]
 	dr.promises[dr.linkPosition] = nil
+	linkPos := dr.linkPosition
 	dr.linkPosition++
 
 	switch nxt := nxt.(type) {
@@ -134,7 +134,12 @@ func (dr *PBDagReader) precalcNextBuf(ctx context.Context) error {
 			return err
 		}
 	}
-	dr.buf = newSizeAdjReadSeekCloser(dr.buf, blocksize)
+	// fixme: a unixfs node with links but no Blocksizes is ill
+	//   defined and should be an error condation, but for now allow it
+	//   to avoid breaking things
+	if linkPos < len(dr.pbdata.Blocksizes) {
+		dr.buf = newSizeAdjReadSeekCloser(dr.buf, dr.pbdata.Blocksizes[linkPos])
+	}
 	return nil
 }
 
--- a/unixfs/io/pbdagreader.go
+++ b/unixfs/io/pbdagreader.go
@@ -51,19 +51,37 @@ type PBDagReader struct {
 var _ DagReader = (*PBDagReader)(nil)
 
 // NewPBFileReader constructs a new PBFileReader.
-func NewPBFileReader(ctx context.Context, n *mdag.ProtoNode, pb *ftpb.Data, serv ipld.NodeGetter) *PBDagReader {
-	fctx, cancel := context.WithCancel(ctx)
+func NewPBFileReader(ctx context.Context, n *mdag.ProtoNode, pb *ftpb.Data, serv ipld.NodeGetter) (*PBDagReader, error) {
 	curLinks := getLinkCids(n)
+	data := pb.GetData()
+
+	// validate
+	if len(curLinks) != len(pb.Blocksizes) {
+		return nil, errors.New("unixfs ill-formed, number of links does not batch blocksize count")
+	}
+	total := uint64(len(data))
+	for _, blocksize := range pb.Blocksizes {
+		total += blocksize
+	}
+	if pb.Filesize == nil {
+		return nil, errors.New("unixfs ill-formed, filesize field missing")
+	}
+	if total != *pb.Filesize {
+		return nil, fmt.Errorf("unixfs ill-formed, actual size of %d does not match size in filesize field (%d)",
+			total, *pb.Filesize)
+	}
+
+	fctx, cancel := context.WithCancel(ctx)
 	return &PBDagReader{
 		node:     n,
 		serv:     serv,
-		buf:      NewBufDagReader(pb.GetData()),
+		buf:      NewBufDagReader(data),
 		promises: make([]*ipld.NodePromise, len(curLinks)),
 		links:    curLinks,
 		ctx:      fctx,
 		cancel:   cancel,
 		pbdata:   pb,
-	}
+	}, nil
 }
 
 const preloadSize = 10
@@ -117,7 +135,11 @@ func (dr *PBDagReader) precalcNextBuf(ctx context.Context) error {
 			// A directory should not exist within a file
 			return ft.ErrInvalidDirLocation
 		case ftpb.Data_File:
-			dr.buf = NewPBFileReader(dr.ctx, nxt, pb, dr.serv)
+			buf, err := NewPBFileReader(dr.ctx, nxt, pb, dr.serv)
+			if err != nil {
+				return err
+			}
+			dr.buf = buf
 		case ftpb.Data_Raw:
 			dr.buf = NewBufDagReader(pb.GetData())
 		case ftpb.Data_Metadata:
@@ -134,12 +156,7 @@ func (dr *PBDagReader) precalcNextBuf(ctx context.Context) error {
 			return err
 		}
 	}
-	// fixme: a unixfs node with links but no Blocksizes is ill
-	//   defined and should be an error condation, but for now allow it
-	//   to avoid breaking things
-	if linkPos < len(dr.pbdata.Blocksizes) {
-		dr.buf = newSizeAdjReadSeekCloser(dr.buf, dr.pbdata.Blocksizes[linkPos])
-	}
+	dr.buf = newSizeAdjReadSeekCloser(dr.buf, dr.pbdata.Blocksizes[linkPos])
 	return nil
 }
 
--- a/unixfs/io/pbdagreader.go
+++ b/unixfs/io/pbdagreader.go
@@ -146,7 +146,11 @@ func (dr *PBDagReader) precalcNextBuf(ctx context.Context) error {
 		}
 	}
 	if len(dr.pbdata.Blocksizes) > 0 {
-		dr.buf = newSizeAdjReadSeekCloser(dr.buf, dr.pbdata.Blocksizes[linkPos])
+		// if blocksizes are set, ensure that we read exactly that amount of data
+		dr.buf, err = newSizeAdjReadSeekCloser(dr.buf, dr.pbdata.Blocksizes[linkPos])
+		if err != nil {
+			return err
+		}
 	}
 	return nil
 }
--- a/unixfs/io/pbdagreader.go
+++ b/unixfs/io/pbdagreader.go
@@ -52,24 +52,13 @@ var _ DagReader = (*PBDagReader)(nil)
 
 // NewPBFileReader constructs a new PBFileReader.
 func NewPBFileReader(ctx context.Context, n *mdag.ProtoNode, pb *ftpb.Data, serv ipld.NodeGetter) (*PBDagReader, error) {
+	err := ft.ValidatePB(n, pb)
+	if err != nil {
+		return nil, err
+	}
+
 	curLinks := getLinkCids(n)
 	data := pb.GetData()
-
-	// validate
-	if len(curLinks) != len(pb.Blocksizes) {
-		return nil, errors.New("unixfs ill-formed, number of links does not batch blocksize count")
-	}
-	total := uint64(len(data))
-	for _, blocksize := range pb.Blocksizes {
-		total += blocksize
-	}
-	if pb.Filesize == nil {
-		return nil, errors.New("unixfs ill-formed, filesize field missing")
-	}
-	if total != *pb.Filesize {
-		return nil, fmt.Errorf("unixfs ill-formed, actual size of %d does not match size in filesize field (%d)",
-			total, *pb.Filesize)
-	}
 
 	fctx, cancel := context.WithCancel(ctx)
 	return &PBDagReader{
--- a/unixfs/io/pbdagreader.go
+++ b/unixfs/io/pbdagreader.go
@@ -52,7 +52,7 @@ var _ DagReader = (*PBDagReader)(nil)
 
 // NewPBFileReader constructs a new PBFileReader.
 func NewPBFileReader(ctx context.Context, n *mdag.ProtoNode, pb *ftpb.Data, serv ipld.NodeGetter) (*PBDagReader, error) {
-	err := ft.ValidatePB(n, pb)
+	err := ft.ValidatePB(n.Links(), pb)
 	if err != nil {
 		return nil, err
 	}
--- a/unixfs/io/pbdagreader.go
+++ b/unixfs/io/pbdagreader.go
@@ -145,7 +145,9 @@ func (dr *PBDagReader) precalcNextBuf(ctx context.Context) error {
 			return err
 		}
 	}
-	dr.buf = newSizeAdjReadSeekCloser(dr.buf, dr.pbdata.Blocksizes[linkPos])
+	if len(dr.pbdata.Blocksizes) > 0 {
+		dr.buf = newSizeAdjReadSeekCloser(dr.buf, dr.pbdata.Blocksizes[linkPos])
+	}
 	return nil
 }
 
@@ -274,6 +276,11 @@ func (dr *PBDagReader) Seek(offset int64, whence int) (int64, error) {
 			dr.linkPosition = 0
 			dr.offset = offset
 			return offset, nil
+		}
+
+		// error out if we have links but not blocksizes
+		if len(pb.Blocksizes) == 0 && len(dr.node.Links()) > 0 {
+			return -1, errors.New("unseekable node (no blocksizes defined)")
 		}
 
 		// skip past root block data
--- a/unixfs/io/pbdagreader.go
+++ b/unixfs/io/pbdagreader.go
@@ -100,6 +100,7 @@ func (dr *PBDagReader) precalcNextBuf(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
+	blocksize := dr.pbdata.Blocksizes[dr.linkPosition]
 	dr.promises[dr.linkPosition] = nil
 	dr.linkPosition++
 
@@ -117,10 +118,8 @@ func (dr *PBDagReader) precalcNextBuf(ctx context.Context) error {
 			return ft.ErrInvalidDirLocation
 		case ftpb.Data_File:
 			dr.buf = NewPBFileReader(dr.ctx, nxt, pb, dr.serv)
-			return nil
 		case ftpb.Data_Raw:
 			dr.buf = NewBufDagReader(pb.GetData())
-			return nil
 		case ftpb.Data_Metadata:
 			return errors.New("shouldnt have had metadata object inside file")
 		case ftpb.Data_Symlink:
@@ -131,8 +130,12 @@ func (dr *PBDagReader) precalcNextBuf(ctx context.Context) error {
 	default:
 		var err error
 		dr.buf, err = NewDagReader(ctx, nxt, dr.serv)
-		return err
+		if err != nil {
+			return err
+		}
 	}
+	dr.buf = newSizeAdjReadSeekCloser(dr.buf, blocksize)
+	return nil
 }
 
 func getLinkCids(n ipld.Node) []*cid.Cid {
@@ -189,13 +192,13 @@ func (dr *PBDagReader) CtxReadFull(ctx context.Context, b []byte) (int, error) {
 
 // WriteTo writes to the given writer.
 func (dr *PBDagReader) WriteTo(w io.Writer) (int64, error) {
+	// If no cached buffer, load one
 	if dr.buf == nil {
 		if err := dr.precalcNextBuf(dr.ctx); err != nil {
 			return 0, err
 		}
 	}
 
-	// If no cached buffer, load one
 	total := int64(0)
 	for {
 		// Attempt to write bytes from cached buffer
--- a/unixfs/io/sizeadj.go
+++ b/unixfs/io/sizeadj.go
@@ -42,20 +42,26 @@ func (r *sizeAdjReadSeekCloser) Read(p []byte) (int, error) {
 
 // Seek implements the Seek method as defined by io.Seeker
 func (r *sizeAdjReadSeekCloser) Seek(offset int64, whence int) (int64, error) {
+	newOffset := r.offset
 	switch whence {
 	case io.SeekStart:
-		r.offset = offset
+		newOffset = offset
 	case io.SeekCurrent:
-		r.offset += offset
+		newOffset += offset
 	case io.SeekEnd:
-		r.offset = r.size + offset
+		newOffset = r.size + offset
 	}
-	if r.offset < 0 {
+	if newOffset < 0 {
 		return -1, errors.New("Seek will result in negative position")
 	}
 	// Its easier just to always use io.SeekStart rather than
 	// correctly adjust offset for io.SeekCurrent and io.SeekEnd.
-	return r.base.Seek(r.offset, io.SeekStart)
+	_, err := r.base.Seek(newOffset, io.SeekStart)
+	if err != nil {
+		return -1, err
+	}
+	r.offset = newOffset
+	return r.offset, nil
 }
 
 // Close implements the Close method as defined by io.Closer
--- a/unixfs/io/sizeadj.go
+++ b/unixfs/io/sizeadj.go
@@ -1,0 +1,109 @@
+package io
+
+import (
+	"errors"
+	"io"
+)
+
+func newSizeAdjReadSeekCloser(base ReadSeekCloser, size uint64) *sizeAdjReadSeekCloser {
+	r := &sizeAdjReadSeekCloser{base: base, size: int64(size)}
+	if r.size < 0 {
+		panic("size limited 2^63 − 1")
+	}
+	return r
+}
+
+type sizeAdjReadSeekCloser struct {
+	base   ReadSeekCloser
+	size   int64
+	offset int64
+}
+
+// Read implements the Read method as defined by io.Reader
+func (r *sizeAdjReadSeekCloser) Read(p []byte) (int, error) {
+	if r.offset >= r.size { // EOF
+		return 0, io.EOF
+	}
+	if int64(len(p)) > r.size-r.offset { // truncate
+		newsize := r.size - r.offset
+		p = p[:newsize]
+	}
+	n, err := r.base.Read(p)
+	if err == nil {
+		_, err = r.base.Read(nil)
+	}
+	if err != io.EOF { // only pad when we get an EOF
+		r.offset += int64(n)
+		return n, err
+	}
+	for ; n < len(p) && r.offset+int64(n) < r.size; n++ { // pad
+		p[n] = 0
+	}
+	r.offset += int64(n)
+	return n, io.EOF
+}
+
+// Seek implements the Seek method as defined by io.Seeker
+func (r *sizeAdjReadSeekCloser) Seek(offset int64, whence int) (int64, error) {
+	switch whence {
+	case io.SeekStart:
+		r.offset = offset
+	case io.SeekCurrent:
+		r.offset += offset
+	case io.SeekEnd:
+		r.offset = r.size + offset
+	}
+	if r.offset < 0 {
+		return -1, errors.New("Seek will result in negative position")
+	}
+	// Its easier just to always use io.SeekStart rather than
+	// correctly adjust offset for io.SeekCurrent and io.SeekEnd.
+	return r.base.Seek(r.offset, io.SeekStart)
+}
+
+// Close implements the Close method as defined by io.Closer
+func (r *sizeAdjReadSeekCloser) Close() error {
+	return r.base.Close()
+}
+
+// WriteTo implemented WriteTo method as defined by io.WriterTo
+func (r *sizeAdjReadSeekCloser) WriteTo(w io.Writer) (int64, error) {
+	lr := &truncWriter{base: w, size: r.size - r.offset}
+	_, err := r.base.WriteTo(lr)
+	n := lr.offset
+	if err != nil {
+		r.offset += n
+		return n, err
+	}
+	if r.offset+n < r.size {
+		zeros := make([]byte, r.size-(r.offset+n))
+		n0, err0 := w.Write(zeros)
+		n += int64(n0)
+		err = err0
+	}
+	r.offset += n
+	return n, err
+}
+
+// truncWriter accepts all bytes written to it, but discards the tail
+// after size bytes are accepted
+type truncWriter struct {
+	base   io.Writer
+	size   int64
+	offset int64
+}
+
+// Write implemented Write method as defined by io.Writer
+func (w *truncWriter) Write(p []byte) (int, error) {
+	truncC := 0
+	if int64(len(p)) > w.size-w.offset {
+		truncC = int(int64(len(p)) - w.size - w.offset)
+		p = p[:w.size]
+	}
+	if len(p) == 0 {
+		return truncC, nil
+	}
+	n, err := w.base.Write(p)
+	w.offset += int64(n)
+	return n + truncC, err
+}
--- a/unixfs/io/sizeadj.go
+++ b/unixfs/io/sizeadj.go
@@ -29,9 +29,6 @@ func (r *sizeAdjReadSeekCloser) Read(p []byte) (int, error) {
 		p = p[:newsize]
 	}
 	n, err := r.base.Read(p)
-	if err == nil {
-		_, err = r.base.Read(nil)
-	}
 	if err != io.EOF { // only pad when we get an EOF
 		r.offset += int64(n)
 		return n, err
--- a/unixfs/io/sizeadj.go
+++ b/unixfs/io/sizeadj.go
@@ -79,13 +79,30 @@ func (r *sizeAdjReadSeekCloser) WriteTo(w io.Writer) (int64, error) {
 		return n, err
 	}
 	if r.offset+n < r.size {
-		zeros := make([]byte, r.size-(r.offset+n))
-		n0, err0 := w.Write(zeros)
-		n += int64(n0)
+		n0, err0 := writeZeros(w, r.size-(r.offset+n))
+		n += n0
 		err = err0
 	}
 	r.offset += n
 	return n, err
+}
+
+func writeZeros(w io.Writer, numZeros int64) (int64, error) {
+	n := int64(0)
+	for numZeros > 0 {
+		l := int64(4 * 1024)
+		if l > numZeros {
+			l = numZeros
+		}
+		zeros := make([]byte, l)
+		n0, err := w.Write(zeros)
+		n += int64(n0)
+		if err != nil {
+			return n, err
+		}
+		numZeros -= l
+	}
+	return n, nil
 }
 
 // truncWriter accepts all bytes written to it, but discards the tail
--- a/unixfs/io/sizeadj.go
+++ b/unixfs/io/sizeadj.go
@@ -5,12 +5,12 @@ import (
 	"io"
 )
 
-func newSizeAdjReadSeekCloser(base ReadSeekCloser, size uint64) *sizeAdjReadSeekCloser {
+func newSizeAdjReadSeekCloser(base ReadSeekCloser, size uint64) (*sizeAdjReadSeekCloser, error) {
 	r := &sizeAdjReadSeekCloser{base: base, size: int64(size)}
 	if r.size < 0 {
-		panic("size limited 2^63 − 1")
+		return nil, errors.New("sizeAdjReadSeekCloser: size limited 2^63 − 1")
 	}
-	return r
+	return r, nil
 }
 
 type sizeAdjReadSeekCloser struct {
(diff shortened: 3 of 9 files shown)


ivan386:
```
>ipfs get -LD QmUeec6DpWz2CA9uZ7GgPAB5nc6QYDg8yax4Wapxs1LSQL
2018-02-11 14:33:45.265625 DEBUG blockservice blockservice.go:183: BlockService GetBlock: 'QmS7RSorsfu1MJpBjcu7L1W2QfsoFDKSrvqAiHzLY4EUH4'
2018-02-11 14:33:45.281250 DEBUG blockservice blockservice.go:183: BlockService GetBlock: 'QmWWC1GHCa7at4K6oMK7Wm7odYCphhRvXYXKAJBreMhRLu'
2018-02-11 14:33:45.281250 DEBUG blockservice blockservice.go:183: BlockService GetBlock: 'QmSKboVigcD3AY4kLsob117KJcMHvMUu6vNFqk1PQzYUpp'
2018-02-11 14:33:45.281250 DEBUG blockservice blockservice.go:183: BlockService GetBlock: 'QmUNLLsPACCz1vLxQVkXqqLX5R1X345qqfHbsf67hvA3Nn'
2018-02-11 14:33:45.296875 DEBUG path resolver.go:55: Resolve: 'QmUeec6DpWz2CA9uZ7GgPAB5nc6QYDg8yax4Wapxs1LSQL'
2018-02-11 14:33:45.296875 DEBUG path resolver.go:144: resolve dag get
2018-02-11 14:33:45.296875 DEBUG blockservice blockservice.go:183: BlockService GetBlock: 'QmUeec6DpWz2CA9uZ7GgPAB5nc6QYDg8yax4Wapxs1LSQL'
Saving file(s) to QmUeec6DpWz2CA9uZ7GgPAB5nc6QYDg8yax4Wapxs1LSQL
 0 B / 3.05 KB [------------------------------------------------------------------------------------------------------------------]   0.00%2
018-02-11 14:33:45.296875 DEBUG blockservice blockservice.go:239: Blockservice: Got data in datastore
2018-02-11 14:33:45.312500 DEBUG blockservice blockservice.go:239: Blockservice: Got data in datastore
2018-02-11 14:33:45.312500 DEBUG blockservice blockservice.go:239: Blockservice: Got data in datastore
2018-02-11 14:33:45.312500 DEBUG blockservice blockservice.go:239: Blockservice: Got data in datastore
2018-02-11 14:33:45.312500 DEBUG blockservice blockservice.go:239: Blockservice: Got data in datastore
 3.05 KB / 3.05 KB [===========================================================================================================] 100.00% 0s
Error: unexpected EOF
2018-02-11 14:33:45.328125 INFO command request.go:105: Shutting down node...
2018-02-11 14:33:45.328125 DEBUG core core.go:544: core is shutting down...
2018-02-11 14:33:45.328125 DEBUG blockservice blockservice.go:274: blockservice is shutting down...
```
No blocksizes. All links write to file.

Stebalien:
Nice!

As @ivan386 says, we should check if `filesize == sum(blocksize)`. Personally, I'd just fail the entire chunk if that check fails (if you want a hole at the beginning of a file, you can always point to the empty node and give it a non-empty size).

unixfs/io/sizeadj.go
```diff
@@ -0,0 +1,109 @@
...
+	if r.offset >= r.size { // EOF
+		return 0, io.EOF
+	}
+	if int64(len(p)) > r.size-r.offset { // truncate
+		newsize := r.size - r.offset
+		p = p[:newsize]
+	}
+	n, err := r.base.Read(p)
+	if err == nil {
+		_, err = r.base.Read(nil)
```
Stebalien: I'd just allow the short read and force the outer reader to loop. Unfortunately, a lot of readers don't like read attempts with empty buffers (and I'd rather not do two reads per outer read).

unixfs/io/sizeadj.go
```diff
@@ -0,0 +1,109 @@
...
+		r.offset += offset
+	case io.SeekEnd:
+		r.offset = r.size + offset
+	}
+	if r.offset < 0 {
+		return -1, errors.New("Seek will result in negative position")
+	}
+	// Its easier just to always use io.SeekStart rather than
+	// correctly adjust offset for io.SeekCurrent and io.SeekEnd.
+	return r.base.Seek(r.offset, io.SeekStart)
```
Stebalien: Shouldn't this return `r.offset`? `base.Seek` could return a short offset.

unixfs/io/sizeadj.go
```diff
@@ -0,0 +1,109 @@
...
+	switch whence {
+	case io.SeekStart:
+		r.offset = offset
+	case io.SeekCurrent:
+		r.offset += offset
+	case io.SeekEnd:
+		r.offset = r.size + offset
+	}
+	if r.offset < 0 {
+		return -1, errors.New("Seek will result in negative position")
```
Stebalien: Shouldn't this rollback the offset (technically)?

kevina:
@Stebalien, two questions

1) If there are links defined but no corresponding blocksizes, should we consider the block ill-formed and fail with an error, after all seeking is impossible (this came up in one of the test cases on a dag that was likely constructed manually, `QmTTDNxNJUqF5PZicJqjDjsUed1bHaPWwFdEMvb1iY93Ur` in `t0110-gateway.sh` test _Add compact blocks_)

2) If `filesize != sum(blocksize)` which should take precedence, I think it would be better to consider it ill-formed and fail with an error.  Is that what you mean when you said "I'd just fail the entire chunk if that check fails"?

unixfs/io/sizeadj.go
kevina: Yes, will fix.

unixfs/io/sizeadj.go
kevina: Yes.

unixfs/io/sizeadj.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+	n, err := r.base.Read(p)
+	if err == nil {
+		_, err = r.base.Read(nil)
+	}
+	if err != io.EOF { // only pad when we get an EOF
+		r.offset += int64(n)
+		return n, err
+	}
+	for ; n < len(p) && r.offset+int64(n) < r.size; n++ { // pad
```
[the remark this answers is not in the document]
kevina: Ok.

unixfs/io/sizeadj.go
kevina: I did, that initially and it broke some code I will look into it more closely.

Stebalien:
> If there are links defined but no corresponding blocksizes, should we consider the block ill-formed and fail with an error...

IMO, yes.

> If `filesize != sum(blocksize)` which should take precedence, I think it would be better to consider it ill-formed and fail with an error. Is that what you mean when you said "I'd just fail the entire chunk if that check fails"?

Yes. I consider this situation to be the same as (1).

unixfs/io/sizeadj.go
Stebalien: That's *not* good...

ivan386:
> if you want a hole at the beginning of a file, you can always point to the empty node and give it a non-empty size

For test: QmRqh1ckCLG6nXoE1aJQT2CHrByk5DWji8v3s4bHcmN975

```
>ipfs dag get QmRqh1ckCLG6nXoE1aJQT2CHrByk5DWji8v3s4bHcmN975
{"data":"CAIS4AMwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAYwBYg4AMg4AMg4AMg4AMg4AM=",
"links":[
{"Cid":{"/":"zb2rhmy65F3REf8SZp7De11gxtECBGgUKaLdiDj7MCGCHxbDW"},"Name":"1","Size":0},
{"Cid":{"/":"zb2rhb3bbxnrqKS2QopcDvbCz5F9nRW7MFzRkhH7LoqhG8AyD"},"Name":"2","Size":480},
{"Cid":{"/":"zb2rhahcgW9tVL7cbPnbdf8y5DzYdbLqCKh2AeBto1JnyCErj"},"Name":"3","Size":480},
{"Cid":{"/":"zb2rhbo1XsFWFP3LCF14NHDikCBn1o8hXVZ6BtR5hvntU5PRp"},"Name":"4","Size":480},
{"Cid":{"/":"zb2rhhkerjTtMxmCwrT9UZC3z7vVuLK8nPnMJzBFRwrEQtiyL"},"Name":"5","Size":480}
]}
```

Here zero size raw link: zb2rhmy65F3REf8SZp7De11gxtECBGgUKaLdiDj7MCGCHxbDW
Need to detect that zero data links and don't try to download it. We waste time on it.

I think that aligning blocks to the end of the file is the best solution. Less data in block.

`ipfs add` do not allow add zero sized file as raw link. It return: QmbFMke1KXqnYyBBWxB74N4c5SBnJMVAiMNRcGu6x1AwQH

```
ipfs dag get QmbFMke1KXqnYyBBWxB74N4c5SBnJMVAiMNRcGu6x1AwQH
{"data":"CAIYAA==","links":[]}
```

test/sharness/t0110-gateway.sh
```diff
@@ -188,7 +188,7 @@ test_expect_success "Add compact blocks" '
   printf "foofoo" > expected
 '
 
-test_expect_success "GET compact blocks succeeds" '
+test_expect_failure "GET compact blocks succeeds" '
```
whyrusleeping: These sorts of changes make me nervous. Whats the technical reason for the swap?

unixfs/io/pbdagreader.go
```diff
@@ -51,19 +51,37 @@ type PBDagReader struct {
 var _ DagReader = (*PBDagReader)(nil)
 
 // NewPBFileReader constructs a new PBFileReader.
-func NewPBFileReader(ctx context.Context, n *mdag.ProtoNode, pb *ftpb.Data, serv ipld.NodeGetter) *PBDagReader {
-	fctx, cancel := context.WithCancel(ctx)
+func NewPBFileReader(ctx context.Context, n *mdag.ProtoNode, pb *ftpb.Data, serv ipld.NodeGetter) (*PBDagReader, error) {
 	curLinks := getLinkCids(n)
+	data := pb.GetData()
+
+	// validate
```
whyrusleeping: I'd like to have this logic broken out separately and have a few unit tests for it

kevina: Can you clarify what you mean here?

whyrusleeping: This validate logic that has been newly added. It should live in its own function. And then we should add a few units tests for that function.

kevina: Okay.

test/sharness/t0110-gateway.sh
Stebalien: That's not good... I assumed that all of the files *we* produced did had blocksizes listed. @whyrusleeping did we change this at some point?

Stebalien:
@ivan386 

> I think that aligning blocks to the end of the file is the best solution. Less data in block.

It adds complexity. I don't want to have one case where we zero-pad the end and another where we zero-pad the beginning.

> `ipfs add` do not allow add zero sized file as raw link. It return: QmbFMke1KXqnYyBBWxB74N4c5SBnJMVAiMNRcGu6x1AwQH

That's a bug. Would you mind reporting it? Also, we technically have an "identity" hash function so, eventually, we should be able to encode this empty block as 4 bytes: `\x01\x37\x00\x00` (CIDv1, DagRaw, Multihash-ID, Length 0).

ivan386:
@Stebalien 
> That's a bug. Would you mind reporting it?

Reported: #4688

> Note: The empty block is zb2rhWm2M1wXXqtqU6pHfovz3DZQ7D54ZD2xN3ynwankHCBCn (can add it with ipfs block put --format=raw <<<'').

zb2rhWm2M1wXXqtqU6pHfovz3DZQ7D54ZD2xN3ynwankHCBCn 1 byte block  with line feed(0A)

Real zero length raw block is: zb2rhmy65F3REf8SZp7De11gxtECBGgUKaLdiDj7MCGCHxbDW

> Also, we technically have an "identity" hash function so, eventually, we should be able to encode this empty block as 4 bytes: \x01\x37\x00\x00 (CIDv1, DagRaw, Multihash-ID, Length 0).

Covert it to base58: z2oTmZ

```
ipfs dag put -f"protobuf" test_dag_identity.json
Error: multihash too short. must be > 3 bytes
```

test_dag_identity.json:
```
{"data":"CAIS4AMwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAYwBYg4AMg4AMg4AMg4AMg4AM=",
"links":[
{"Name":"1","Size":0,"Cid":{"/":"z2oTmZ"}},
{"Name":"2","Size":480,"Cid":{"/":"zb2rhb3bbxnrqKS2QopcDvbCz5F9nRW7MFzRkhH7LoqhG8AyD"}},
{"Name":"3","Size":480,"Cid":{"/":"zb2rhahcgW9tVL7cbPnbdf8y5DzYdbLqCKh2AeBto1JnyCErj"}},
{"Name":"4","Size":480,"Cid":{"/":"zb2rhbo1XsFWFP3LCF14NHDikCBn1o8hXVZ6BtR5hvntU5PRp"}},
{"Name":"5","Size":480,"Cid":{"/":"zb2rhhkerjTtMxmCwrT9UZC3z7vVuLK8nPnMJzBFRwrEQtiyL"}}
]}
```
Report new bug?

test/sharness/t0110-gateway.sh
whyrusleeping: @Stebalien I think this test is testing a specifically constructed block.

whyrusleeping: I think we should add a decent number of unit tests for stuff like this. Testing this sort of specific detailed logic via sharness feels wrong.

Stebalien:
> Real zero length raw block is...

Good catch.

> Report new bug?

The fact that we choke on a zero-length identity multihash is a bug in the multihash library. The fact that we don't currently support "inline blocks" is more of a missing feature (we only recently added the identity multihash and haven't added support for extracting blocks from it yet).

test/sharness/t0110-gateway.sh
kevina: So is the consistence that we should remove this test and instead test that node in a unit test?  And to be clear, do we want this (in my opinion) ill-formed node to fail to validate?

whyrusleeping:
progress here?

unixfs/io/pbdagreader.go
kevina: Okay this is done now.  Sorry for the delay.

kevina:
Assuming the test pass I think this should be good to go.  We can clean up the failed test case later.

test/sharness/t0110-gateway.sh
Kubuxu: This test was added as custom formed blocks might not have sizes but they still should be plain gettable from HTTP. They still can be read sequentially, just efficient seeking is impossible.

unixfs/io/sizeadj.go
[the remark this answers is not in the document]
kevina: I can still do this but I am reluctant to do to the increase in code size.

test/sharness/t0110-gateway.sh
kevina: @Kubuxu so is this still something we want to allow (unseekable blocks)?  We would need some special cases to allow it.

Kubuxu: Yes

kevina: @whyrusleeping thoughs?

mib-kd743naq: @kevina here is the background on this testcase: https://github.com/ipfs/go-ipfs/issues/4286#issuecomment-336163070

Specifically:

>Additionally - the way my project chunks data, my average blocksize is around 4096 bytes ( with a hard-limit on block sizes at 64k ), so forgoing any optional metadata is critical for me in order to reduce the size of my objects ( as the overhead adds up incredibly fast when one deals with hundreds of millions of objects in a DAG ).

[@whyrusleeping's response](https://github.com/ipfs/go-ipfs/issues/4286#issuecomment-336195370):
> The sizes are optional

@kevina please do not break this relied-upon feature ( neither in PB- nor in the future CBOR-UnixFS )

whyrusleeping:
As @Stebalien has done most of the review here, i'd like to get his 👍 on merging

magik6k:
@whyrusleeping see https://github.com/ipfs/go-ipfs/pull/4680#discussion_r171746304

~Kubuxu

test/sharness/t0110-gateway.sh
Stebalien: So, this is a problem. Unix programs *expect* files to be seekable and will break horribly if they aren't (and I'm not going to break every single unix program ever written). However, I also don't want to break existing uses if I can help it and this problem doesn't seem insurmountable.

We *should* be able to fall back on dumb seeking (although we may want to disable this from the gateway). Basically, the logic would be as follows:

1. If blocksizes are defined, use them.
2. If a block with children *has no blocksizes* defined (*none* at all, not some, not half, not incorrect ones, **none**), dumb seek through that portion of the file.

This actually gives us the best of both worlds. Files with lots of tiny blocks can put blocksizes at the top layers of the DAG and omit them in the lower levels.

----

In terms of *implementation*, we can punt on dumb seeking for now. Simply validate blocksizes if present and error when seeking when they're not present.

@kevina how hard would this be?

unixfs/io/sizeadj.go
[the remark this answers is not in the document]
Stebalien: Entirely up to you. I seriously doubt this will be a bottleneck.

Stebalien:
@whyrusleeping, @kevina see my comment [here](https://github.com/ipfs/go-ipfs/pull/4680#discussion_r172658912) (unless I missed some change and this is no longer an issue).

test/sharness/t0110-gateway.sh
kevina: I will see what I can do.

kevina: Done.

kevina:
@Stebalien I pushed a commit which I think should address your concern.

kevina:
@Stebalien I fixed the test case in case that was what you where waiting for.  @magik6k a review will would be helpful from you also

unixfs/unixfs.go
```diff
@@ -258,3 +259,23 @@ func BytesForMetadata(m *Metadata) ([]byte, error) {
 func EmptyDirNode() *dag.ProtoNode {
 	return dag.NodeWithData(FolderPBData())
 }
+
+// ValidatePB validates a unixfs protonode.
+func ValidatePB(n *dag.ProtoNode, pb *pb.Data) error {
+	if len(pb.Blocksizes) == 0 { // special case
+		return nil
```
magik6k: Can probably still check if data size is less-equal than the declared filesize.

unixfs/io/sizeadj.go
```diff
@@ -0,0 +1,112 @@
...
+
+import (
+	"errors"
+	"io"
+)
+
+func newSizeAdjReadSeekCloser(base ReadSeekCloser, size uint64) *sizeAdjReadSeekCloser {
+	r := &sizeAdjReadSeekCloser{base: base, size: int64(size)}
+	if r.size < 0 {
+		panic("size limited 2^63 − 1")
```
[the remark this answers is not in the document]
kevina: I don't think so, but it is easier enough to return an error.

unixfs/io/sizeadj.go
```diff
@@ -0,0 +1,112 @@
...
+	size   int64
+	offset int64
+}
+
+// Read implements the Read method as defined by io.Reader
+func (r *sizeAdjReadSeekCloser) Read(p []byte) (int, error) {
+	if r.offset >= r.size { // EOF
+		return 0, io.EOF
+	}
+	if int64(len(p)) > r.size-r.offset { // truncate
```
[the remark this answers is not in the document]
kevina: I am not sure I understand.

unixfs/io/sizeadj.go
```diff
@@ -0,0 +1,112 @@
...
+// Seek implements the Seek method as defined by io.Seeker
+func (r *sizeAdjReadSeekCloser) Seek(offset int64, whence int) (int64, error) {
+	newOffset := r.offset
+	switch whence {
+	case io.SeekStart:
+		newOffset = offset
+	case io.SeekCurrent:
+		newOffset += offset
+	case io.SeekEnd:
+		newOffset = r.size + offset
```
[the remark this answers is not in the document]
kevina: Oops.  Fixed.

unixfs/io/sizeadj.go
```diff
@@ -0,0 +1,112 @@
...
+func (r *sizeAdjReadSeekCloser) WriteTo(w io.Writer) (int64, error) {
+	lr := &truncWriter{base: w, size: r.size - r.offset}
+	_, err := r.base.WriteTo(lr)
+	n := lr.offset
+	if err != nil {
+		r.offset += n
+		return n, err
+	}
+	if r.offset+n < r.size {
+		zeros := make([]byte, r.size-(r.offset+n))
```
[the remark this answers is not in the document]
kevina: I am not sure what that limit is, and I don't see a strong need for it, however, I can be convinced otherwise.

unixfs/io/pbdagreader.go
```diff
@@ -131,8 +141,17 @@ func (dr *PBDagReader) precalcNextBuf(ctx context.Context) error {
 	default:
 		var err error
 		dr.buf, err = NewDagReader(ctx, nxt, dr.serv)
-		return err
+		if err != nil {
+			return err
+		}
 	}
+	if len(dr.pbdata.Blocksizes) > 0 {
+		dr.buf, err = newSizeAdjReadSeekCloser(dr.buf, dr.pbdata.Blocksizes[linkPos])
```
whyrusleeping: can we add a comment here:
```
// if blocksizes are set, ensure that we read exactly that amount of data
```

unixfs/io/pbdagreader.go
```diff
@@ -264,6 +283,11 @@ func (dr *PBDagReader) Seek(offset int64, whence int) (int64, error) {
 			return offset, nil
 		}
 
+		// error out if we have links but not blocksizes
+		if len(pb.Blocksizes) == 0 && len(dr.node.Links()) > 0 {
```
whyrusleeping: so this means that seeking through a unixfs dag with no blocksizes set will error out. We *could* dumb-seek through by just reading a bunch, but that might produce unexpected perf degredations.

@mib-kd743naq does this work for you (erroring out on seek of no-blocksizes)?

unixfs/io/sizeadj.go
```diff
@@ -0,0 +1,112 @@
...
+		newOffset += offset
+	case io.SeekEnd:
+		newOffset = r.size - offset
+	}
+	if newOffset < 0 {
+		return -1, errors.New("Seek will result in negative position")
+	}
+	// Its easier just to always use io.SeekStart rather than
+	// correctly adjust offset for io.SeekCurrent and io.SeekEnd.
+	_, err := r.base.Seek(newOffset, io.SeekStart)
```
whyrusleeping: what happens if `r.size == 1000`, but `r.base.Size() == 500` and we try to call `r.Seek(700)`? 

It seems like this would error out, when it should succeed and subsequent reads should return up to 300 bytes of zeros. I guess i'm not sure what errors Seek returns when you try to seek past its bounds.

unixfs/io/sizeadj.go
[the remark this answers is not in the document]
whyrusleeping: what happens here if `r.size == 10000000000` and `r.base.Size() == 1` and we end up here? smells like OOMdeath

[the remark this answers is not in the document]
whyrusleeping: this could be trivially used to attack a gateway (a user could construct such a malicious node, and then just try to fetch it through the gateway. The gateway uses `WriteTo` as an optimization)

unixfs/io/sizeadj.go
```diff
@@ -0,0 +1,112 @@
...
+// truncWriter accepts all bytes written to it, but discards the tail
+// after size bytes are accepted
+type truncWriter struct {
+	base   io.Writer
+	size   int64
+	offset int64
+}
+
+// Write implemented Write method as defined by io.Writer
+func (w *truncWriter) Write(p []byte) (int, error) {
```
whyrusleeping: add tests for this function once the issues here are fixed please

unixfs/io/sizeadj.go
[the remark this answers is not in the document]
kevina: Acknowledge.  I am just not sure what the solution should be.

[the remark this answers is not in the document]
whyrusleeping: just something that writes a bunch of zeros without allocating unboundedly? like maybe a `WriteZeros` method that takes a write and a count, and writes zeros in 1k chunks up to the desired amount

unixfs/io/pbdagreader.go
mib-kd743naq: @whyrusleeping I am not sure how to answer this, as I do not have a good mental model *where* does the seeker get invoked.

Coming from the other side: what I need is to keep the ability of [the test cases for #4286](https://ipfs.io/ipfs/QmZL4wivfYBEUutxksKgbpPfFkj1tGNMESYLFF2MWcST8Y/) to be:
- fully traversed/displayed [via the web gateway](https://ipfs.io/ipfs/QmZL4wivfYBEUutxksKgbpPfFkj1tGNMESYLFF2MWcST8Y)
- can be downloaded through `ipfs get QmZL4wivfYBEUutxksKgbpPfFkj1tGNMESYLFF2MWcST8Y`

If both of these continue to work - I am happy ;)

@kevina if you need further size-less testcase construction - just let me know roughly what tree shape you are interested in

unixfs/io/sizeadj.go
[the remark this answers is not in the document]
kevina: Okay done.

unixfs/io/pbdagreader.go
kevina: @mib-kd743naq The tests cases added in #4286 now pass and I manually tested that the hash QmZL4wivfYBEUutxksKgbpPfFkj1tGNMESYLFF2MWcST8Y works with both the gateway and `get`.

unixfs/io/sizeadj.go
kevina: Actually it won't error out, from the documentation (https://golang.org/pkg/io/#Seeker):
> Seeking to an offset before the start of the file is an error. __Seeking to any positive offset is legal__, but the behavior of subsequent I/O operations on the underlying object is implementation-dependent.

unixfs/io/sizeadj.go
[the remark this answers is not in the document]
kevina: Actually, I believe the expected behavior is if you use SeekEnd and won't to seek to a point before the end you should use a negative offset.  A positive offset will seek past the end, which is legal.

Thus the original code was correct.

unixfs/unixfs.go
```diff
@@ -258,3 +259,23 @@ func BytesForMetadata(m *Metadata) ([]byte, error) {
...
+	}
+	if len(n.Links()) != len(pb.Blocksizes) {
+		return errors.New("unixfs ill-formed, number of links does not match blocksize count")
+	}
+	total := uint64(len(pb.GetData()))
+	for _, blocksize := range pb.Blocksizes {
+		total += blocksize
+	}
+	if total != pb.GetFilesize() {
```
Stebalien: So, one case here is `pb.Filesize == nil`. Is `Filesize` required?

unixfs/unixfs.go
Stebalien: There's also the case where there *are* no blocks but we still have inline data.

kevina: > There's also the case where there are no blocks but we still have inline data.

And in that case I don't think there is anything to validate, but I could be mistaken.

unixfs/unixfs.go
kevina: I double check, and if not I will add a check for that.

unixfs/unixfs.go
kevina: Actually I take that back.  I will have a closer look later today.

unixfs/unixfs.go
kevina: Note: the function pb.GetFilesize() returns a uint64, if pb.Filesize == nil it will return 0

Stebalien: Sorry, my question was more: If Filesize is nil, that means it wasn't specified. Do we *require* Filesize to be specified? Should we calculate it from the blocksizes if left unspecified?

mib-kd743naq: @Stebalien given there **must** be a source of size, and given that `blocksizes` are optional, my hunch would be to just make [`Filesize`](https://github.com/ipfs/go-ipfs/blob/v0.4.15/unixfs/pb/unixfs.proto#L15) hard-required.

Stebalien: Hm. I'd like @whyrusleeping and @diasdavid to :+1: this.

@whyrusleeping, @diasdavid:

Will all IPFS files include a Filesize?

whyrusleeping: I think its probably fine to assume that any correctly formatted ipfs/unixfs file will contain a filesize. Seems pretty hard to do a lot of things without it.

kevina: Okay I added a test to make sure it is not nil.

unixfs/io/pbdagreader.go
kevina: Where this will fail is if you use the `Range:` header to retrieve partial content.  If you do something like `GET /ipfs/QmTTDNxNJUqF5PZicJqjDjsUed1bHaPWwFdEMvb1iY93Ur HTTP/1.1` with `Range:bytes=3-` you get back `HTTP/1.1 416 Requested Range Not Satisfiable`.

This can be fixed by doing a dummy seek, or alternatively not have the gateway not return a `Content-Range` response in the header.

unixfs/io/pbdagreader.go
```diff
@@ -131,8 +141,18 @@ func (dr *PBDagReader) precalcNextBuf(ctx context.Context) error {
...
 		var err error
 		dr.buf, err = NewDagReader(ctx, nxt, dr.serv)
-		return err
+		if err != nil {
+			return err
+		}
 	}
+	if len(dr.pbdata.Blocksizes) > 0 {
+		// if blocksizes are set, ensure that we read exactly that amount of data
+		dr.buf, err = newSizeAdjReadSeekCloser(dr.buf, dr.pbdata.Blocksizes[linkPos])
```
Stebalien: You know, I think we can do something simpler here. Given that we *know* the Filesize of the buffer, we can require that `buf` export a `Size()` method. Then, we can just have a check here:

* `buf.Size() > blocksize` -> make a limit seeker
* `==` -> return as-is
* `<` -> zero extend

That way, we can do all of these checks up-front. Thoughts? Too late?

Stebalien:
What currently happens if there are no blocksizes but the filesize doesn't match? Both with and without inline data? We may need to truncate/zero extend in this case as well.

Current rules:

1. Filesize is required.
2. If blocksizes are specified:
    1. They must sum to the filesize.
    2. There must be one for each block.
    3. If a blocksize is greater than the size of a block, zero extend.
    4. If smaller, truncate.
3. If blocksizes are not specified:
   1. Fail (or truncate?) if the *inline* data < Filesize? **NEW**
   2. Truncate/zero extend the entire file to Filesize **NEW**

One thing is a bit inconsistent. If no blocksizes are specified, the filesize is used to truncate/zero extend. If they are specified, the filesize can't do this. We could also just truncate/extend always (i.e., don't check that `sum(blocksizes) == filesize`)...

---

So, I know this is kind of cold feet at the 11th hour however, after seeing all the weird corner-cases that come up in practice, I'm starting to wonder if we should fail more. That is:

1. Filesize is required.
2. If blocksizes are specified:
    1. They must sum to the filesize.
    2. There must be one for each block.
    3. Each block's blocksize must match the filesize specified in the block (or the implied filesize for raw leaves). Basically, when we load a block, we call GetFilesize on the block and check it against the size.
3. If blocksizes are not specified: Keep a running tally of how much data we've read so far. If we hit an EOF before we hit the filesize, return a read error. Same if we hit the filesize before reading the end of the file. We can even do this at the block level: If we load a block that would cause us to exceed the filesize, fail; if the last block isn't big enough, fail.

The downsides are:

1. We lose this truncation/extension technique.
2. We change much of this PR (although, really, it should mostly just be code deletions).

---

* Q: Ditch the truncate/extend and be strict?
    * [ ] Yes.
    * [ ] No. Q: Truncate/zero extend on `sum(blocksize)`/`filesize` mismatch?
        * [ ] Yes.
        * [ ] No. Q: Truncate or fail if inline data < filesize?
            * [ ] Truncate.
            * [ ] Fail.

magik6k:
I'm not sure what the status on this currently is, but:
* This code now lives in go-unixfs
* DagReader received a major refactor since this was opened

Stebalien:
This needs to be worked out at the spec level first. We jumped into coding too quickly, IMO.

Issue #4510, opened 2017-12-20
make: check_gx & dist_get issues on Windows


djdv:
#### Version information:
N/A

#### Type:
Bug

#### Severity:
Medium (easy workaround - make sure gx and gx-go are installed and run make 3 times)
#### Wall of text:
Context: my environment is a Windows machine, my shell is `cmd.exe` and my `make.exe` comes from the msys2 project.

On Windows when running `make` `check_gx` fails to detect that gx and gx-go are already installed, I assume this is because it is checking for `gx` and not `gx.exe`. This triggers `dist_get` to fire, this will download a zip containing gx.exe and gx-go.exe, it attempts to extract them but fails with 
```
caution: filename not matched:  gx/gx
unzip has failed
```
After this 1st run `bin/` is left with some garbage inside ([ref](https://ipfs.io/ipfs/QmbBN9DbCD9JtX34P1pV3ErkMkSgD4Frfn35Dzw1Y2TLxS/post-build-1)).
Running `make` again after this point (2nd) causes it to leave behind a 0 byte `gx` file in `bin/`([ref](https://ipfs.io/ipfs/QmbBN9DbCD9JtX34P1pV3ErkMkSgD4Frfn35Dzw1Y2TLxS/post-build-2)), with this stub in place `check_gx` seems to succeed for `gx` but not `gx-go`, this causes `dist_get` to fire again to fetch `gx-go`, however `unzip` will fail again
```
caution: filename not matched:  gx-go/gx-go
unzip has failed
```

running `make` again (3rd) will cause it to leave another stub in `bin/` during extraction, this time for `gx-go` ([ref](https://ipfs.io/ipfs/QmbBN9DbCD9JtX34P1pV3ErkMkSgD4Frfn35Dzw1Y2TLxS/post-build-3)). `check_gx` will see both stubs in `bin/` at this point and assume we have them, going forth with the standard build procedure. On Windows this build will succeed so long as we have `gx.exe` and `gx-go.exe` in our path since invoking `gx` is synonymous with invoking `gx.exe` on Windows, however if we did not have gx and gx-go installed, the build would fail at this point since `check_gx` was fooled by the stubs.

There are 2 issues overall here, 1 is that .exe files are not being picked up, and 2 is that the downloaded dependencies are not being properly validated, I assume this is because the validation process invokes by name and does not specify a direct path i.e `gx` instead of `bin/gx`.

[...]

PR #4682, opened 2018-02-11
Fix/win make

djdv:
Requesting review from @Kubuxu I'm not very familiar with IPFS's `make` rules or `make` in general so this may need changes.
In my testing this ~~~resolved~~~ partially resolves https://github.com/ipfs/go-ipfs/issues/4510 when using MSYS2 tools.
I'm going to look into gx detection

While working on this I noticed a problem I couldn't work around.
/bin/Rules.mk utilizes a symlink for the binary dependencies, on Windows, by default, users don't have permission to create symlinks with `mklink`, MSYS2's `ln.exe` simply copies the source to the target (not as a link). This ends up not being a problem in practice but I feel the need to mention it. Here's the patch I initially tried without success.

```
diff --git a/bin/Rules.mk b/bin/Rules.mk
index 1a1dd7962..a2576c303 100644
--- a/bin/Rules.mk
+++ b/bin/Rules.mk
@@ -16,7 +16,11 @@ PATH := $(realpath $(d)):$(PATH)

 $(TGTS_$(d)):
        rm -f $@$(binpostfix)
+ifeq ($(OS),Windows_NT)
+       cmd /C mklink "$@$(binpostfix)" "$(notdir $^)$(binpostfix)"
+else
        ln -s $(notdir $^) $@
+endif

 bin/gx-v%:
        @echo "installing gx $(@:bin/gx-%=%)"

```

### DIFF
--- a/bin/Rules.mk
+++ b/bin/Rules.mk
@@ -11,8 +11,12 @@ DISTCLEAN += $(wildcard $(d)/gx-v*) $(wildcard $(d)/gx-go-v*) $(d)/tmp
 PATH := $(realpath $(d)):$(PATH)
 
 $(TGTS_$(d)):
-	rm -f $@
+	rm -f $@$(?exe)
+ifeq ($(WINDOWS),1)
+	cp $^$(?exe) $@$(?exe)
+else
 	ln -s $(notdir $^) $@
+endif
 
 bin/gx-v%:
 	@echo "installing gx $(@:bin/gx-%=%)"
--- a/bin/dist_get
+++ b/bin/dist_get
@@ -57,6 +57,13 @@ unarchive() {
 	ua_infile="$2"
 	ua_outfile="$3"
 	ua_distname="$4"
+	ua_binpostfix=""
+	ua_os=$(uname -o)
+
+	if [ "$ua_os" = "Msys" ] || [ "$ua_os" = "Cygwin" ] ; then
+	    ua_binpostfix=".exe"
+	fi
+	ua_outfile="$ua_outfile$ua_binpostfix"
 
 	if ! check_writeable "$ua_outfile"; then
 		die "unarchive error: cannot write to $ua_outfile"
@@ -66,7 +73,7 @@ unarchive() {
 		tar.gz)
 			if have_binary tar; then
 				echo "==> using 'tar' to extract binary from archive"
-				< "$ua_infile" tar -Ozxf - "$ua_distname/$ua_distname" > "$ua_outfile" \
+				< "$ua_infile" tar -Ozxf - "$ua_distname/$ua_distname$ua_binpostfix" > "$ua_outfile" \
 					|| die "tar has failed"
 			else
 				die "no binary on system for extracting tar files"
@@ -75,7 +82,7 @@ unarchive() {
 		zip)
 			if have_binary unzip; then
 				echo "==> using 'unzip' to extract binary from archive"
-				unzip -p "$ua_infile" "$ua_distname/$ua_distname" > "$ua_outfile" \
+				unzip -p "$ua_infile" "$ua_distname/$ua_distname$ua_binpostfix" > "$ua_outfile" \
 					|| die "unzip has failed"
 			else
 				die "no installed method for extracting .zip archives"


djdv:
>I'm going to look into gx detection

I thought the makefiles tried to detect automatically if gx was already installed and use it if it was, I didn't notice `IPFS_GX_USE_GLOBAL` had to be set to use the system version of gx.

With these commits I'm able to clone the repo, set environment variables, build, and install `go-ipfs` on Windows under MSYS2 using `make build` and `make install` without issues. Both with bundled `gx` and system `gx` (when `IPFS_GX_USE_GLOBAL` is set to 1).

whyrusleeping:
@VictorBjelkholm What is the latest on windows CI through jenkins?

djdv:
Cygwin's `ln` works differently from Msys's, it will create a binary file that works as a symlink inside the Cygwin shell but seems to be causing issues. 
```
==> Download complete!
==> using 'unzip' to extract binary from archive
rm -f bin/gx-go.exe
ln -s gx-go-v1.6.0.exe bin/gx-go.exe
gx.exe install --global
ERROR: req-check hook failed: fork/exec C:\Users\J\go\src\github.com\ipfs\go-ipf
s\bin\gx-go.exe: Access is denied.
make: *** [mk/gx.mk:8: gx-deps] Error 1
```
Now this only happens when `gx` is invoked from `make`, this isn't happening when I invoke `bin/gx.exe install --global` directly from the shell. I don't know who's bug this is.

For now we can just avoid linking altogether on Windows and just copy the binary every time. It's not ideal but I can't think of a way around it.

VictorBjelkholm:
@whyrusleeping 
> What is the latest on windows CI through jenkins?

PR in progress is here: https://github.com/ipfs/go-ipfs/pull/4430

Windows tests seems to work fine, but it's not calling `make` to run them.

bin/Rules.mk
```diff
@@ -1,6 +1,10 @@
 include mk/header.mk
 
 dist_root_$(d)=/ipfs/QmT3CLJKJzWPuN4NAN4LLy69UpKskMF3AuYhXstKdn8V43
+binpostfix :=
```
Kubuxu: We already have `binpostfix` in this makefile system. You can use `$(?exe)` to access it.

bin/Rules.mk
```diff
@@ -11,8 +15,12 @@ DISTCLEAN += $(wildcard $(d)/gx-v*) $(wildcard $(d)/gx-go-v*) $(d)/tmp
 PATH := $(realpath $(d)):$(PATH)
 
 $(TGTS_$(d)):
-	rm -f $@
-	ln -s $(notdir $^) $@
+	rm -f $@$(binpostfix)
+ifeq ($(OS),Windows_NT)
```
Kubuxu: Same with this, use the variable exported here: https://github.com/ipfs/go-ipfs/blob/master/mk/util.mk#L4

mk/gx.mk
```diff
@@ -1,7 +1,11 @@
-gx-path = gx/ipfs/$(shell gx deps find $(1))/$(1)
+binpostfix :=
+ifeq ($(OS),Windows_NT)
+binpostfix = .exe
+endif
+gx-path = gx/ipfs/$(shell gx$(binpostfix) deps find $(1))/$(1)
```
Kubuxu: `gx` in this file should be avialable from PATH variable so path or the postfix should not be needed.

Kubuxu: Does it work without it?

djdv:
@Kubuxu 
Should be better now.

Kubuxu:
@magik6k could you look why Jenkins failed here?

magik6k:
It's a random test fail, I've seen it a few times, reported in https://github.com/ipfs/go-ipfs/issues/4698 (also re-run jenkins, it's green now)

Kubuxu:
@djdv can you try running `make build` twice in a row. It shouldn't be unpacking the file twice.

From another side, I don't see a reason this couldn't get merged. It improves the situation, it just might not be perfect.

djdv:
@Kubuxu 
On my system it downloads and extracts `gx` and `gx-go` on the first run but re-uses them on the second run without downloading or extracting anything. The only time it gets them again is when I remove them with `git clean -fdx`.

Are you encountering double unpacking?

Also I will force push to restart the test if they're working now. I needed to amend the commit anyway (wrong email address was attached originall). No code changes.

Kubuxu:
@djdv no, but this is something I suspected that could happen reading the Makefiles.

@whyrusleeping good to merge

whyrusleeping:
Thank you @djdv! Its really great to have someone holding the windows flag :)

Stebalien:
I can't wait for the day when we can depend on the WSL and cross compile for windows on windows...

PR #4686, opened 2018-02-12
Extract go-detect-race from Godeps

hsanjuan:
I have forked it, put it under ipfs namespace, published to gx

License: MIT


whyrusleeping:
Thanks @hsanjuan!

PR #4689, opened 2018-02-12
Doc: golint-ify path package.

hsanjuan:
This removes all go-lint warnings in the path package.

### DIFF
--- a/path/path.go
+++ b/path/path.go
@@ -1,3 +1,4 @@
+// Package path contains utilities to work with ipfs paths.
 package path
 
 import (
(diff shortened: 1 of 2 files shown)


hsanjuan:
By the way, I will be extracting this too

lgierth:
> By the way, I will be extracting this too

Have you synced with @Stebalien? It's part of the IPLD extraction/disentanglement

Stebalien:
@lgierth this package actually covers IPFS path resolution as well (the resolver is pluggable, technically).

whyrusleeping:
👏 👏

PR #4691, opened 2018-02-13
Change Windows build documentation

djdv:
I rewrote this. Maybe it's better this way?
I'm not sure if the table alignment looks fine or crazy. Other options for that: left aligned table, separated command blocks with text annotations above, single paragraph explaining what will happen followed by a single command block.

I also have graphics that could be added but they're probably unnecessary. 
Embed underneath "Building on Windows": https://ipfs.io/ipfs/QmccXW7JSZMVXidSc7tHsU6aktuaiV923q4yBGHUsdymYo/build.gif
Add as a reference/link near the Cygwin setup portion: https://ipfs.io/ipfs/QmaYFSQa4iHDafcebiKjm1WwuKhosoXr45HPpfaeMbCRpb/cygwin%20-%20install.png

I added MSYS2 as a method, and basically redid everything else. I feel like it is more concise than it was previously, even with the repetition/explanations.
Criticisms welcomed.

Blocked on: https://github.com/ipfs/go-ipfs/pull/4682

### DIFF
--- a/docs/windows.md
+++ b/docs/windows.md
@@ -1,142 +1,147 @@
 # Building on Windows
-
-## Install Git For Windows
-
-As Git is used by the Go language to download dependencies, you need
-to install Git, for example from http://git-scm.com/.
-
-You also must make sure that the directory that contains the Git For
-Windows binary is in the Path environment variable. Note that Git For
-Windows has a 'git' binary in a 'Git\bin' directory and another one in
-a 'Git\cmd' directory. You should only put the 'Git\cmd' directory in
-the Path environment variable.
+![](https://ipfs.io/ipfs/QmccXW7JSZMVXidSc7tHsU6aktuaiV923q4yBGHUsdymYo/build.gif)
 
 ## Install Go
-
-Please install the Go language as explained on
-https://golang.org/doc/install.
-
-To properly install Go, you will need to set some environment
-variables. We recommend you to set them globally using the Control
-Panel, as explained in the documentation above, so that these
-environment variables are automatically available in all the possible
-environments that you might want to use like Git Bash, Windows's cmd,
-Cygwin's terminal, Windows' PowerShell and so on.
-
-You must make sure that the GOROOT environment variable is set and
-that the %GOROOT%/bin directory is in the Path environment variable.
-
-The GOPATH environment variable should also be set to a directory that
-you have created, and the %GOPATH/bin directory should also be in the
-Path environment variable.
-
-## Download go-ipfs and fix Git authentication
-
-Use the following command to download go-ipfs source code:
-
-```
-go get -u github.com/ipfs/go-ipfs
-```
-
-The above command uses Git to download go-ipfs from its GitHub
-repository. If you get authentication problems with Git, you might
-want to take a look at
-https://help.github.com/articles/caching-your-github-password-in-git/
-and use the suggested solution:
-
-```
-git config --global credential.helper wincred
-```
+`go-ipfs` is built on Golang and thus depends on it for all building methods.  
+https://golang.org/doc/install  
+The `GOPATH` environment variable must be set as well.  
+https://golang.org/doc/code.html#GOPATH
 
 ## Choose the way you want to proceed
+`go-ipfs` utilizes `make` to automate builds and run tests, but can be built without it using only `git` and `go`.  
+No matter which method you choose, if you encounter issues, please see the [Troubleshooting](#troubleshooting) section.  
 
-Now there are two ways to download, install the dependencies and to
-build go-ipfs:
+**Using `make`:**  
+MSYS2 and Cygwin provide the Unix tools we need to build `go-ipfs`. You may use either, but if you don't already have one installed, we recommend MSYS2.  
+[MSYS2→](#msys2)  
+[Cygwin→](#cygwin)  
 
-1) There is the "Manual Way", where you don't need to install anymore
-software except the dependencies, but you have a number of commands to
-type.
+**Using build tools manually:**  
+This section assumes you have a working version of `go` and `git` already setup. You may want to build this way if your environment restricts installing additional software, or if you're integrating IPFS into your own build system.  
+[Minimal→](#minimal)  
 
-2) There is a way by installing 'make' through Cygwin and using it to
-do nearly everything. We call this way the "Cygwin Way". It may take
-much more time, because installing Cygwin can take a lot of time, but
-after that it might be easier as many procedures are just a 'make'
-command away.
+## MSYS2
+1. Install msys2 (http://www.msys2.org)  
+2. Run the following inside a normal `cmd` prompt (Not the MSYS2 prompt, we only need MSYS2's tools).  
+An explanation of this block is below.
+```
+SET PATH=%PATH%;\msys64\usr\bin
+pacman --noconfirm -S  git make unzip
+go get -u github.com/ipfs/go-ipfs
+cd %GOPATH%\src\github.com\ipfs\go-ipfs
+make install
+%GOPATH%\bin\ipfs.exe version --all
+```
 
-So use the next steps below that start with "Manual Way" if that's the
-way you want, otherwise scroll down a bit and use the "Cygwin Way"
-steps below.
+If there were no errors, the final command should output version information similar to "`ipfs version 0.4.14-dev-XXXXXXX`" where "XXXXXXX" should match the current short-hash of the `go-ipfs` repo. You can retrieve said hash via this command: `git rev-parse --short HEAD`.  
+If `ipfs.exe` executes and the version string matches, then building was successful.
 
-## Manual Way: download and install dependencies
+|Command|Explanation|
+| ---: | :--- |
+|`SET PATH=%PATH%;\msys64\usr\bin`         |Add msys2's tools to our [`PATH`](https://ss64.com/nt/path.html); Defaults to: (\msys64\usr\bin)|
+|`pacman --noconfirm -S  git make unzip`   |Install `go-ipfs` build dependencies|
+|`go get -u github.com/ipfs/go-ipfs`       |Fetch / Update `go-ipfs` source|
+|`cd %GOPATH%\src\github.com\ipfs\go-ipfs` |Change to `go-ipfs` source directory|
+|`make install`                            |Build and install to `%GOPATH%\bin\ipfs.exe`|
+|`%GOPATH%\bin\ipfs.exe version --all`     |Test the built binary|
 
-The following commands should download or update go-ipfs dependencies
-and then install them:
+To build again after making changes to the source, run:
+```
+SET PATH=%PATH%;\msys64\usr\bin
+cd %GOPATH%\src\github.com\ipfs\go-ipfs
+make install
+```
+
+**Tip:** To avoid setting `PATH` every time (`SET PATH=%PATH%;\msys64\usr\bin`), you can lock it in permanently using `setx` after it's been set once:
+```
+SETX PATH %PATH%
+```
+
+## Cygwin
+1. Install Cygwin (https://www.cygwin.com)  
+2. During the install, select the following packages. (If you already have Cygwin installed, run the setup file again to install additional packages.) A fresh install should look something like [this reference image](https://ipfs.io/ipfs/QmaYFSQa4iHDafcebiKjm1WwuKhosoXr45HPpfaeMbCRpb/cygwin%20-%20install.png).
+    - devel packages
+        - `git`
+        - `make`
+    - archive packages
+        - `unzip`
+    - net packages
+        - `curl`  
+3. Run the following inside a normal `cmd` prompt (Not the Cygwin prompt, we only need Cygwin's tools)  
+An explanation of this block is below.
+```
+SET PATH=%PATH%;\cygwin64\bin
+mkdir %GOPATH%\src\github.com\ipfs
+cd %GOPATH%\src\github.com\ipfs
+git clone https://github.com/ipfs/go-ipfs.git
+cd %GOPATH%\src\github.com\ipfs\go-ipfs
+make install
+%GOPATH%\bin\ipfs.exe version --all
+```
+
+If there were no errors, the final command should output version information similar  to "`ipfs version 0.4.14-dev-XXXXXXX`" where "XXXXXXX" should match the current short-hash of the `go-ipfs` repo. You can retrieve said hash via this command: `git rev-parse --short HEAD`.  
+If `ipfs.exe` executes and the version string matches, then building was successful.
+
+|Command|Explanation|
+| ---: | :--- |
+|`SET PATH=%PATH%;\cygwin64\bin`           |Add Cygwin's tools to our [`PATH`](https://ss64.com/nt/path.html); Defaults to: (\cygwin64\bin)|
+|`mkdir %GOPATH%\src\github.com\ipfs`<br/>`cd %GOPATH%\src\github.com\ipfs`<br/>`git clone https://github.com/ipfs/go-ipfs.git`       |Fetch / Update `go-ipfs` source|
+|`cd %GOPATH%\src\github.com\ipfs\go-ipfs` |Change to `go-ipfs` source directory|
+|`make install`                            |Build and install to `%GOPATH%\bin\ipfs.exe`|
+|`%GOPATH%\bin\ipfs.exe version --all`     |Test the built binary|
+
+To build again after making changes to the source, run:
+```
+SET PATH=%PATH%;\cygwin64\bin
+cd %GOPATH%\src\github.com\ipfs\go-ipfs
+make install
+```
+
+**Tip:** To avoid setting `PATH` every time (`SET PATH=%PATH%;\cygwin64\bin`), you can lock it in permanently using `setx` after it's been set once:
+```
+SETX PATH %PATH%
+```
+
+## Minimal
+While it's possible to build `go-ipfs` with `go` alone, we'll be using `git` and `gx` for practical source management.  
+You can use whichever version of `git` you wish but we recommend the Windows builds at <https://git-scm.com>. `git` must be in your [`PATH`](https://ss64.com/nt/path.html) for `go get` to recognize and use it.
+
+### `gx`
+You may install prebuilt binaries for [`gx`](https://dist.ipfs.io/#gx) & [`gx-go`](https://dist.ipfs.io/#gx-go) if they're available for your platform.
+Alternatively, you can build them from source and install them to `%GOPATH%\bin` by running the following:
 
 ```
 go get -u github.com/whyrusleeping/gx
 go get -u github.com/whyrusleeping/gx-go
+```
+
+### `go-ipfs`
+
+```
+SET PATH=%PATH%;%GOPATH%\bin
+go get -u -d github.com/ipfs/go-ipfs
 cd %GOPATH%/src/github.com/ipfs/go-ipfs
 gx --verbose install --global
-```
-
-## Manual Way: build go-ipfs
-
-To actually build go-ipfs, first go to the cmd/ipfs directory:
-
-```
 cd cmd\ipfs
 ```
+We need the `git` commit hash to be included in our build so that in the extremely rare event a bug is found, we have a reference point later for tracking it. We'll ask `git` for it and store it in a variable. The syntax for the next command is different depending on whether you're using the interactive command line or writing a batch file. Use the one that applies to you.  
+- interactive: `FOR /F %V IN ('git rev-parse --short HEAD') do set SHA=%V`  
+- interpreter: `FOR /F %%V IN ('git rev-parse --short HEAD') do set SHA=%%V`  
 
-Then get the current Git commit:
-
+Finally, we'll build and test `ipfs` itself.
 ```
-git rev-parse --short HEAD
+go install -ldflags="-X "github.com/ipfs/go-ipfs/repo/config".CurrentCommit=%SHA%"
+%GOPATH%\bin\ipfs.exe version --all
 ```
+You can check that the ipfs output versions match with `go version` and `git rev-parse --short HEAD`.  
+If `ipfs.exe` executes and everything matches, then building was successful.
 
-It will output a small number of hex characters that you must pass to
-the actual build command (replace XXXXXXX with these characters):
+## Troubleshooting
+- **Git auth**
+If you get authentication problems with Git, you might want to take a look at https://help.github.com/articles/caching-your-github-password-in-git/ and use the suggested solution:  
+`git config --global credential.helper wincred`
 
-```
-go install -ldflags="-X "github.com/ipfs/go-ipfs/repo/config".CurrentCommit=XXXXXXX"
-```
+- **Anything else**  
+Please search [https://discuss.ipfs.io](https://discuss.ipfs.io/search?q=windows%20category%3A13) for any additional issues you may encounter. If you can't find any existing resolution, feel free to post a question asking for help.
 
-After that ipfs should have been built and should be available in
-"%GOPATH%\bin".
-
-You can check that the ipfs you built has the right version using:
-
-```
-ipfs version --commit
-```
-
-It should output something like "ipfs version 0.4.0-dev-XXXXXXX" where
-XXXXXXX is the current commit that you passed to the build command.
-
-## Cygwin way: install Cygwin
-
-Install Cygwin as explained in the Cygwin documentation:
-
-http://cygwin.com/install.html
-
-By default Cygwin will not install 'make', so you should click on the
-"Devel" category during the Cygwin installation process and then check
-the 'make' package.
-
-## Cygwin way: build go-ipfs
-
-To build go-ipfs using Cygwin you just need to open a Cygwin Terminal
-and then type the following commands:
-
-```
-cd $GOPATH/src/github.com/ipfs/go-ipfs
-make install
-```
-
-After that ipfs should have been built and should be available in
-"%GOPATH%\bin".
-
-You can check that the ipfs you built has the right version using:
-
-```
-ipfs version --commit
-```
+If you encounter a bug with `go-ipfs` itself (not related to building) please use the [issue tracker](https://github.com/ipfs/go-ipfs/issues) to report it.


djdv:
I removed the "eval loops" `FOR /F "tokens=*" %E IN ('go env') DO %E`, from the `make` method, their purpose was just to make sure the environment variables for `go` (specifically `GOPATH`) were set but it's probably better to just demand that the user set it themselves.

Unfortunately there's no better way to take stdout and place it into an environment variable (that I know of) so it has to remain to store the git hash before building in the "minimal" section.

This commit will be squashed later.

docs/windows.md
```diff
@@ -1,142 +1,94 @@
...
-make install
-```
-
-After that ipfs should have been built and should be available in
-"%GOPATH%\bin".
-
-You can check that the ipfs you built has the right version using:
+The syntax for the next command is different depending on how it's executed. If you're on the interactive command line vs executing from inside the batch/cmd interpriter.  
+interactive: `FOR /F %V IN ('git rev-parse --short HEAD') do set SHA=%V`  
+interpriter: `FOR /F %%V IN ('git rev-parse --short HEAD') do set SHA=%%V`
```
whyrusleeping: typo: interpreter

docs/windows.md
```diff
@@ -1,142 +1,94 @@
...
-"%GOPATH%\bin".
-
-You can check that the ipfs you built has the right version using:
-
-```
-ipfs version --commit
-```
-
-It should output something like "ipfs version 0.4.0-dev-XXXXXXX" where
-XXXXXXX is the current commit that you passed to the build command.
```
Mr0grog: I think you should keep the step of verifying the install here. (You could remove the “where		
 -XXXXXXX is the current commit that you passed to the build command” bit since you’ve abstracted that away with a command + env var.)

It might also be useful to give some basic advice on what to do if they *don’t* get the expected output, even if it’s just “post to https://discuss.ipfs.io .”

docs/windows.md
```diff
@@ -1,142 +1,94 @@
...
-"%GOPATH%\bin".
-
-You can check that the ipfs you built has the right version using:
+The syntax for the next command is different depending on how it's executed. If you're on the interactive command line vs executing from inside the batch/cmd interpriter.  
+interactive: `FOR /F %V IN ('git rev-parse --short HEAD') do set SHA=%V`  
+interpreter: `FOR /F %%V IN ('git rev-parse --short HEAD') do set SHA=%%V`
 
 ```
-ipfs version --commit
+go install -ldflags="-X "github.com/ipfs/go-ipfs/repo/config".CurrentCommit=%SHA%"
```
Mr0grog: I think it would definitely be helpful to note that this is the actual build step. It could be as simple as just saying “finally, actually build and install `go-ipfs`.”

docs/windows.md
```diff
@@ -1,142 +1,94 @@
...
-much more time, because installing Cygwin can take a lot of time, but
-after that it might be easier as many procedures are just a 'make'
-command away.
-
-So use the next steps below that start with "Manual Way" if that's the
-way you want, otherwise scroll down a bit and use the "Cygwin Way"
-steps below.
+## Cygwin
+Install Cygwin (https://www.cygwin.com)  
+During the install, select the devel packages `git` & `make`, archive package `unzip`, and net package `curl`
```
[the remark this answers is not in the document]
Mr0grog: Additionally, how should someone install these packages if they already have Cygwin? (If I recall, the answer is run the Cygwin setup again, but not sure if things have since improved.)

docs/windows.md
```diff
@@ -1,142 +1,94 @@
...
-
-The GOPATH environment variable should also be set to a directory that
-you have created, and the %GOPATH/bin directory should also be in the
-Path environment variable.
-
-## Download go-ipfs and fix Git authentication
+`go-ipfs` is built on Golang and thus depends on it for all building methods.  
+https://golang.org/doc/install  
+The `GOPATH` environment variable must be set as well.  
+https://golang.org/doc/code.html#GOPATH
```
Mr0grog: Do you also still need to set `GOROOT`? I think the previous version of this section was a little bit more explanatory for someone who might not be familiar with building Go programs. I know when I first approached Go, I found the various environment variables it required surprising (the equivalents are typically optional customizations with many other languages).

It might make sense to keep most of the previous version of this section as-is or tweak it slightly. Keeping it short is good, but this version strikes me as a little *too* terse.

docs/windows.md
```diff
@@ -1,142 +1,94 @@
...
+pacman --noconfirm -S  git make unzip
 go get -u github.com/ipfs/go-ipfs
+cd %GOPATH%\src\github.com\ipfs\go-ipfs
+make install
 ```
 
-The above command uses Git to download go-ipfs from its GitHub
-repository. If you get authentication problems with Git, you might
-want to take a look at
-https://help.github.com/articles/caching-your-github-password-in-git/
```
Mr0grog: Is this still a common issue? If so, it seems like the advice here should be kept, maybe in a trouble-shooting section at the end?

docs/windows.md
```diff
@@ -1,142 +1,94 @@
...
+## Choose the way you want to proceed
+`go-ipfs` utilizes `make` to automate builds and testing, but can be built without it using only `git` and `go`.  
+The `make` portions cover setting up the environment, "minimal" assumes you have a working version of `go` and `git` already.  
+[make→](#make)  
+[minimal→](#minimal)
+
+## Make
+There are currently 2 documented ways of building with `make` on Windows.  
+[MSYS2→](#msys2)  
+[Cygwin→](#cygwin)
```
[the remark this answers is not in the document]
djdv: I'm unsure about adding advice here, it feels subjective. While there are technical differences, at the end of the day both provide the posix tools we need to build go-ipfs. I think the major decision comes down to what users already have installed, if it's neither, the differences between them are just going to be the differences between the 2 sections, which is not much. Mostly it just depends on if you like a graphical package manager or a command line one (*for our purposes).

I'll think this over and see if I can come up with something appropriate. Maybe defer to an external resource comparing the 2.

docs/windows.md
```diff
@@ -1,142 +1,94 @@
...
-and use the suggested solution:
+|      |      |
+| ---: | :--- |
+|`SET PATH=%PATH%;\msys64\usr\bin`         |Add msys2's bin to `PATH`; Defaults to: (\msys64\usr\bin)|
+|`pacman --noconfirm -S  git make unzip`   |Install `go-ipfs` build dependencies|
+|`go get -u github.com/ipfs/go-ipfs`       |Fetch / Update `go-ipfs` source|
+|`cd %GOPATH%\src\github.com\ipfs\go-ipfs` |Change to `go-ipfs` directory|
+|`make install`                            |Build and install to `%GOPATH%\bin\ipfs.exe`|
 
+You can permenantly append the MSYS2 bin to your `PATH` so that on subsequent builds you can simply run
```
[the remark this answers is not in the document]
djdv: >From my recollection of doing Windows dev years ago, this was non-trivial and buried in system settings somewhere

The biggest problem is that it's different across versions, even on Windows 10 alone I've noticed they changed the interface 3 times now.

There's the `setx` command but I need to test access and availability of it. I think users can invoke it normally and I'm seeing Windows 7 as the minimum version but it may only be included with a specific SDK until Windows 8. If it's included in stock Windows 7 I'd be comfortable just recommending that as the means. The only problem I can think of is that it would be bad if users made a typo for their `PATH`
i.e
`setx PATH %PATH;\some\other\path`
notice the lack of the closing % which would just ruin everything. I guess that can't really be avoided either way though.

I'll write a revision including this and maybe a notice "be very careful to check your input here..."

docs/windows.md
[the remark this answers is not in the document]
djdv: You're right about needing to (download and) run the setup binary again, even if you already have it installed. I can see about adding some kind of wording like "Even if you already have Cygwin installed, you'll need to run the setup file again to add these packages"

docs/windows.md
djdv: The Go installer on Windows sets `GOROOT` for us. I don't remember the exact release that started doing this but it's certainly before our minimum required version for building. I'm leaning towards keeping mention of `GOROOT` out since it should be set already and there's a reference to the Go documentation which is updated independently to include any necessary steps to setup the environment as they change with the language.

Maybe adding a comment to reinforce/advising users read it like: "Make sure to read the documentation at https://golang.org/doc/install and https://golang.org/doc/code.html#GOPATH to ensure your environment is properly setup for building".

docs/windows.md
djdv: I'm not sure the origin of this. Even years ago I never encountered this on any of the environments I used for testing. It may have been an issue upstream that was resolved.

Either way the troubleshooting section is a good idea for all misc. comments.

djdv:
Thinking out loud.
I'd really love some method of annotation for the code blocks, tables were just the best I could come up here. In an ideal state I'd love them to be hidden/revealed somehow, like if I could have a toggle button that displays the annotations but still maintain a code block people can select and copy from when the annotations are collapsed/hidden.

djdv:
Draft updated. I plan on looking into making a comment for Msys2 vs Cygwin, I may just insert an "it's preferential" type statement.
I need to do a consistency/grammar and visual/style review myself but making remark before I get to it would be welcomed.
The graphical header was added as well as a reference image link for Cygwin. I'd like a 👍  or 👎  on the graphical header, not sure if it's too much.
If I missed anything else, please remark on it.

docs/windows.md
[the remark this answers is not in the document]
Mr0grog: > I think the major decision comes down to what users already have installed

Absolutely. But that doesn’t cover a large group of people.

> I'm unsure about adding advice here, it feels subjective.

It’s totally subjective :)

The problem is that you are *starting* the whole tutorial with a choice, and, in particular, one that has nothing to do with the task at hand (building IPFS) and one users may likely not know enough about in order to *even know how to choose.* That can stop people from following the tutorial right at the beginning. (And even for someone enterprising, it adds a lot of extra work: now they have to go research each of these options in order to figure out which to choose.)

It’s OK for it to be subjective (and you can even say so). The goal is to remove the *barrier* of this choice. This could be as simple as “all these options are good; if you don’t already have one of these tools installed or aren’t familiar with them, go with MSYS2 because it makes the process easy” or “go with ‘minimal’ because you don’t have to install anything extra.”

docs/windows.md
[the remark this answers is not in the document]
Mr0grog: > The biggest problem is that it's different across versions, even on Windows 10 alone I've noticed they changed the interface 3 times now.

Is this a good resource to link to? https://www.computerhope.com/issues/ch000549.htm

djdv:
@Mr0grog I think I've addressed all that was mentioned, I'd like another pass at this when you get a chance. 

After all is cleared I'll squash and sign-off the commits.

djdv:
Sorry for the whitespace noise, my editor seems to have inconsistencies with GitHub's renderer.

docs/windows.md
```diff
@@ -1,142 +1,139 @@
...
-The GOPATH environment variable should also be set to a directory that
-you have created, and the %GOPATH/bin directory should also be in the
-Path environment variable.
+## Choose the way you want to proceed
+`go-ipfs` utilizes `make` to automate builds and testing, but can be built without it using only `git` and `go`.  
+No matter which section you choose, if you encounter issues, please see the [Troubleshooting](#troubleshooting) section.  
 
-## Download go-ipfs and fix Git authentication
+Using `make`:  
+These sections cover setting up the build environment, each option is equally valid and differ only in their setup process. If you have no preference or special requirements, MSYS2 is recommended as the first choice.
```
Mr0grog: Hopefully it’s already clear to people that this is about setting up their build environment :)  The thing that might be more useful to clarify here is what MSYS2 and Cygwin are, e.g:

> MSYS2 and Cygwin make it possible to run Unix-style programs (like `make`) on Windows. Both are great, but if you aren’t already using one, we recommend MSYS2.

docs/windows.md
```diff
@@ -1,142 +1,139 @@
...
+|`SET PATH=%PATH%;\cygwin64\bin`           |Add Cygwin's bin to PATH; Defaults to: (\cygwin64\bin)|
+|`mkdir %GOPATH%\src\github.com\ipfs`<br/>`cd %GOPATH%\src\github.com\ipfs`<br/>`git clone https://github.com/ipfs/go-ipfs.git`       |Fetch / Update `go-ipfs` source|
+|`cd %GOPATH%\src\github.com\ipfs\go-ipfs` |Change to `go-ipfs` directory|
+|`make install`                            |Build and install to `%GOPATH%\bin\ipfs.exe`|
+|`%GOPATH%\bin\ipfs.exe version --all`     |Test the built binary|
 
-## Manual Way: build go-ipfs
-
-To actually build go-ipfs, first go to the cmd/ipfs directory:
+You can check that the ipfs output versions match with `go version` and `git rev-parse --short HEAD`, if `ipfs.exe` executes and everything matches, then building was successful.
```
Mr0grog: Same as above from the MSYS2 section.

docs/windows.md
```diff
@@ -1,142 +1,139 @@
...
 
-It will output a small number of hex characters that you must pass to
-the actual build command (replace XXXXXXX with these characters):
+## Minimal
+While it's possible to build `go-ipfs` with `go` alone, we'll be using `git` and `gx` for practical source management.  
+You can use whichever version of `git` you wish but the Windows builds at <https://git-scm.com> are recommended. `git` must be in your `PATH` for `go get` to recognize and use it.
 
+### `gx`
+`gx` and its go-hook `gx-go` can be built by us.  
+Alternativley you may use prebuilt binaries ([gx](https://dist.ipfs.io/#gx), [gx-go](https://dist.ipfs.io/#gx-go)) by placing them in your `PATH` and skipping this section.
```
Mr0grog: Flipping this sentence with the previous one places the text about building from source next to the instructions for doing so. That can help clarify exactly what the code block below is supposed to be. e.g:

> You can install prebuilt binaries for gx and gx-go and skip this section. Otherwise, you can install and build them from source by running the following:

docs/windows.md
```diff
@@ -1,142 +1,139 @@
...
-
-So use the next steps below that start with "Manual Way" if that's the
-way you want, otherwise scroll down a bit and use the "Cygwin Way"
-steps below.
-
-## Manual Way: download and install dependencies
-
-The following commands should download or update go-ipfs dependencies
-and then install them:
+You can check that the ipfs output versions match with `go version` and `git rev-parse --short HEAD`, if `ipfs.exe` executes and everything matches, then building was successful.
```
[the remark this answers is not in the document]
djdv: I agree with needing clarity in the first half, as for if this section is necessary, I can lean either way, it serves to test and verify the binary is what it's expected to be.  It's similar to part of the original document
>You can check that the ipfs you built has the right version using...
It should output something like "ipfs version 0.4.0-dev-XXXXXXX" where XXXXXXX is the current commit that you passed to the build command.

I'll combine this with 
>The XXXXXXX part should match the output you get if you run git rev-parse --short HEAD

docs/windows.md
djdv: Sometimes I'm a little redundant and robotic...
I like the more human friendly wording here but I have qualms with saying Cygwin is "great" ;^)
How does this revision sound?
>MSYS2 and Cygwin provide the Unix tools we need to build `go-ipfs`. You may use either, but if you don't already have one installed, MSYS2 is recommended.

P.S.
I'm a little robotic and redundant sometimes.

djdv:
Addressed review comments, moved the "test binary" sections up, added table headers.

djdv:
@Mr0grog, @nothingismagick
Can you take a look at this?

docs/windows.md
```diff
@@ -1,142 +1,141 @@
...
-Cygwin's terminal, Windows' PowerShell and so on.
-
-You must make sure that the GOROOT environment variable is set and
-that the %GOROOT%/bin directory is in the Path environment variable.
-
-The GOPATH environment variable should also be set to a directory that
-you have created, and the %GOPATH/bin directory should also be in the
-Path environment variable.
+## Choose the way you want to proceed
+`go-ipfs` utilizes `make` to automate builds and testing, but can be built without it using only `git` and `go`.
```
nothingismagick: "to automate builds and run tests,"

docs/windows.md
```diff
@@ -1,142 +1,141 @@
...
-
-You must make sure that the GOROOT environment variable is set and
-that the %GOROOT%/bin directory is in the Path environment variable.
-
-The GOPATH environment variable should also be set to a directory that
-you have created, and the %GOPATH/bin directory should also be in the
-Path environment variable.
+## Choose the way you want to proceed
+`go-ipfs` utilizes `make` to automate builds and testing, but can be built without it using only `git` and `go`.  
+No matter which section you choose, if you encounter issues, please see the [Troubleshooting](#troubleshooting) section.
```
nothingismagick: "No matter which **method** you choose, ..." 

The troubleshooting section only mentions git auth stuff and otherwise sends people to discuss.ipfs.io - it might be helpful to also list a few possible gotchas (sry - with windows I can't help there)

docs/windows.md
```diff
@@ -1,142 +1,141 @@
...
-After that ipfs should have been built and should be available in
-"%GOPATH%\bin".
+## Troubleshooting
+- Git auth  
+If you get authentication problems with Git, you might want to take a look at https://help.github.com/articles/caching-your-github-password-in-git/ and use the suggested solution:  
+`git config --global credential.helper wincred`
 
-You can check that the ipfs you built has the right version using:
+- Anything else  
+Please search <https://discuss.ipfs.io/> for any additional issues you encounter, if you can't find an existing resolution, feel free to make a post asking for help.
```
nothingismagick: "Please search <https://discuss.ipfs.io/search?q=windows%20category%3A13> for any additional issues you may encounter. If you can't find any existing resolution, feel free to post a question asking for help."

Of course, if they are behind the great firewall, this service will not be available - but then again neither will github...

docs/windows.md
```diff
@@ -1,142 +1,141 @@
...
+make install
+```
+This can be acheived with the `setx` command
+```
+SETX PATH %PATH%;\msys64\usr\bin
+```
+
+## Cygwin
+1. Install Cygwin (https://www.cygwin.com)  
+2. During the install, select the following packages. (If you already have Cygwin installed, run the setup file again to install additional packages) A fresh install should look something like [this reference image](https://ipfs.io/ipfs/QmaYFSQa4iHDafcebiKjm1WwuKhosoXr45HPpfaeMbCRpb/cygwin%20-%20install.png).
```
nothingismagick: "...additional packages.) A..."

docs/windows.md
```diff
@@ -1,142 +1,141 @@
...
+**Tip**: You can permenantly append the MSYS2 bin to your `PATH` so that on subsequent builds you can simply run
 ```
-go get -u github.com/whyrusleeping/gx
-go get -u github.com/whyrusleeping/gx-go
-cd %GOPATH%/src/github.com/ipfs/go-ipfs
-gx --verbose install --global
+cd %GOPATH%\src\github.com\ipfs\go-ipfs
+make install
+```
+This can be acheived with the `setx` command
```
nothingismagick: "This can be achieved with the `setx` command: "

docs/windows.md
```diff
@@ -1,142 +1,141 @@
...
+|Command|Explanation|
+| ---: | :--- |
+|`SET PATH=%PATH%;\msys64\usr\bin`         |Add msys2's bin to `PATH`; Defaults to: (\msys64\usr\bin)|
+|`pacman --noconfirm -S  git make unzip`   |Install `go-ipfs` build dependencies|
+|`go get -u github.com/ipfs/go-ipfs`       |Fetch / Update `go-ipfs` source|
+|`cd %GOPATH%\src\github.com\ipfs\go-ipfs` |Change to `go-ipfs` directory|
+|`make install`                            |Build and install to `%GOPATH%\bin\ipfs.exe`|
+|`%GOPATH%\bin\ipfs.exe version --all`     |Test the built binary|
 
+**Tip**: You can permenantly append the MSYS2 bin to your `PATH` so that on subsequent builds you can simply run
```
nothingismagick: permanently

docs/windows.md
```diff
@@ -1,142 +1,141 @@
...
-The above command uses Git to download go-ipfs from its GitHub
-repository. If you get authentication problems with Git, you might
-want to take a look at
-https://help.github.com/articles/caching-your-github-password-in-git/
-and use the suggested solution:
-
-```
-git config --global credential.helper wincred
-```
+If there were no errors, the final command should output version information simillar to "`ipfs version 0.4.14-dev-XXXXXXX`" where "XXXXXXX" should match the current short-hash of the `go-ipfs` repo. You can retrieve said hash via this command: `git rev-parse --short HEAD`. If `ipfs` runs and the version string matches, building was successful.
```
nothingismagick: "...version information **similar** to..."

docs/windows.md
```diff
@@ -1,142 +1,141 @@
...
 
-## Download go-ipfs and fix Git authentication
+**Using `make`:**  
+MSYS2 and Cygwin provide the Unix tools we need to build `go-ipfs`. You may use either, but if you don't already have one installed, MSYS2 is recommended.  
+[MSYS2→](#msys2)  
+[Cygwin→](#cygwin)  
 
-Use the following command to download go-ipfs source code:
+**Using build tools manually:**  
+This section assumes you have a working version of `go` and `git` already setup. You may want to build this way if your environment restricts installing additional software, or if you're intigrating IPFS into your own build system.
```
nothingismagick: integrating

nothingismagick:
@djdv - can't say anything about the content, because windows is not my can of worms - but I did find a massive number of spelling mistakes...

djdv:
@nothingismagick 
Thanks!

>massive number of spelling mistakes

I've now installed `aspell`.

docs/windows.md
```diff
@@ -1,142 +1,141 @@
...
-The GOPATH environment variable should also be set to a directory that
-you have created, and the %GOPATH/bin directory should also be in the
-Path environment variable.
+## Choose the way you want to proceed
+`go-ipfs` utilizes `make` to automate builds and run tests, but can be built without it using only `git` and `go`.  
+No matter which method you choose, if you encounter issues, please see the [Troubleshooting](#troubleshooting) section.  
 
-## Download go-ipfs and fix Git authentication
+**Using `make`:**  
+MSYS2 and Cygwin provide the Unix tools we need to build `go-ipfs`. You may use either, but if you don't already have one installed, MSYS2 is recommended.
```
Mr0grog: Do you mind changing “MSYS2 is recommended” to “**we** recommend MSYS2?” Active voice can help alleviate a subtle question here. (“Recommended by whom?”)

I know you were concerned about making a subjective statement here, and clarifying who is making the recommendation helps clarify for the reader that there is not some special authority saying MSYS2 is better.

docs/windows.md
```diff
@@ -1,142 +1,141 @@
...
 ```
-git rev-parse --short HEAD
+SETX PATH %PATH%;\cygwin64\bin
 ```
 
-It will output a small number of hex characters that you must pass to
-the actual build command (replace XXXXXXX with these characters):
+## Minimal
+While it's possible to build `go-ipfs` with `go` alone, we'll be using `git` and `gx` for practical source management.  
+You can use whichever version of `git` you wish but the Windows builds at <https://git-scm.com> are recommended. `git` must be in your `PATH` for `go get` to recognize and use it.
```
Mr0grog: Same note about passive active voice here as in the intro: “…but **we recommend** the Windows builds at <https://git-scm.com>.” (And if this recommendation is from someone/somewhere else, say their name instead of “we.”)

docs/windows.md
```diff
@@ -1,142 +1,141 @@
...
+
+### `gx`
+You may install prebuilt binaries for [`gx`](https://dist.ipfs.io/#gx) & [`gx-go`](https://dist.ipfs.io/#gx-go) if they're available for your platform.
+Alternatively you can build them from source and install them to `%GOPATH%\bin` by running the following:
```
Mr0grog: Comma after “alternatively.”

docs/windows.md
```diff
@@ -1,142 +1,141 @@
...
 
 ```
-ipfs version --commit
+SET PATH=%PATH%;%GOPATH%\bin
+go get -u -d github.com/ipfs/go-ipfs
+cd %GOPATH%/src/github.com/ipfs/go-ipfs
+gx --verbose install --global
+cd cmd\ipfs
 ```
+We need the `git` commit hash to be included in our build so that in the extremely rare event a bug is found, we have a reference point later for tracking them. We'll ask `git` for it and store it in a variable. Unfortunately the syntax for the next command is different depending on how it's executed. If you're on the interactive command line vs executing from inside the batch/cmd interpreter. Use the one that applies to you.
```
Mr0grog: The first sentence needs both “**a** bug is found” and “tracking **them**” to be the same singular or plural.

> If you're on the interactive command line vs executing from inside the batch/cmd interpreter.

This isn’t a full sentence. It would probably read better combined with the previous sentence, like:

> The syntax for the next command is different depending whether you're on the interactive command line or executing it from inside the batch/cmd interpreter.

docs/windows.md
```diff
@@ -1,142 +1,141 @@
...
-ipfs version --commit
+SET PATH=%PATH%;%GOPATH%\bin
+go get -u -d github.com/ipfs/go-ipfs
+cd %GOPATH%/src/github.com/ipfs/go-ipfs
+gx --verbose install --global
+cd cmd\ipfs
 ```
+We need the `git` commit hash to be included in our build so that in the extremely rare event a bug is found, we have a reference point later for tracking them. We'll ask `git` for it and store it in a variable. Unfortunately the syntax for the next command is different depending on how it's executed. If you're on the interactive command line vs executing from inside the batch/cmd interpreter. Use the one that applies to you.  
+interactive: `FOR /F %V IN ('git rev-parse --short HEAD') do set SHA=%V`  
+interpreter: `FOR /F %%V IN ('git rev-parse --short HEAD') do set SHA=%%V`
```
Mr0grog: This looks fine in plain text, but looks a little confusing when rendered as HTML. Making these a bulleted list would probably help.

docs/windows.md
```diff
@@ -1,142 +1,141 @@
...
-and then type the following commands:
-
+Finally, we'll build and test `ipfs` itself.
 ```
-cd $GOPATH/src/github.com/ipfs/go-ipfs
-make install
+go install -ldflags="-X "github.com/ipfs/go-ipfs/repo/config".CurrentCommit=%SHA%"
+%GOPATH%\bin\ipfs.exe version --all
 ```
+You can check that the ipfs output versions match with `go version` and `git rev-parse --short HEAD`, if `ipfs.exe` executes and everything matches, then building was successful.
```
Mr0grog: The comma after `git rev-parse --short HEAD` should be either a semicolon or a period.

docs/windows.md
```diff
@@ -1,142 +1,141 @@
...
 
-After that ipfs should have been built and should be available in
-"%GOPATH%\bin".
+## Troubleshooting
+- Git auth
```
Mr0grog: Not a big deal, but I feel like these bullet titles might read better if formatted like:

```
- **Git auth:** If you get authentication problems…

- **Anything else:** Please search…
```

docs/windows.md
```diff
@@ -1,142 +1,141 @@
...
+`git config --global credential.helper wincred`
 
-You can check that the ipfs you built has the right version using:
+- Anything else  
+Please search [https://discuss.ipfs.io](https://discuss.ipfs.io/search?q=windows%20category%3A13) for any additional issues you may encounter. If you can't find any existing resolution, feel free to post a question asking for help.
 
-```
-ipfs version --commit
-```
+If you encounter a bug with `go-ipfs` itself (not related to building) please use the issue tracker to report it.
```
Mr0grog: Might be nice to make “issue tracker” a link to https://github.com/ipfs/go-ipfs/issues

docs/windows.md
```diff
@@ -1,142 +1,141 @@
...
+        - `unzip`
+    - net packages
+        - `curl`  
+4. Run the following inside a normal `cmd` prompt (Not the Cygwin prompt, we only need Cygwin's tools)  
+An explanation of this block is below.
+```
+SET PATH=%PATH%;\cygwin64\bin
+mkdir %GOPATH%\src\github.com\ipfs
+cd %GOPATH%\src\github.com\ipfs
+git clone https://github.com/ipfs/go-ipfs.git
```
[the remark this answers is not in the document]
djdv: tl;dr There's a path conflicts between Cygwin's `git` and Win32 `go`. We could more than likely use `go get` if `go` was built under Cygwin, but as it is right now they don't provide prebuilts so we just sidestep this by using `git` directly.
***

Go has `GOPATH` restrictions, it wants an absolute native path such as `C:\Users\dd`\Go.
Cygwin mangles paths to look like this `/cygdrive/c/Users/dd/Go`.
`/` on Windows is a relative path to the current volume's root, so if we set our `GOPATH` to `/cygdrive/c/Users/dd/Go`, the `go` tool refuses to function.
```
go: GOPATH entry is relative; must be absolute path: "/cygdrive/c/Users/dd/Go".
For more details see: 'go help gopath'
```
If we set `GOPATH` to `C:\Users\dd\Go`, `go get` will invoke `git` which receives a mangled path and fails:
```
# cd .; git clone https://github.com/ipfs/go-ipfs C:\Users\dd\Go\src\github.com\i
pfs\go-ipfs
Cloning into 'C:\Users\dd\Go\src\github.com\ipfs\go-ipfs'...
fatal: Invalid path '/cygdrive/c/Users/dd/C:\Users\dd\Go\src\github.com\ipfs\go-ip
fs': No such file or directory
package github.com/ipfs/go-ipfs: exit status 128
```

docs/windows.md
```diff
@@ -1,142 +1,141 @@
...
-way you want, otherwise scroll down a bit and use the "Cygwin Way"
-steps below.
-
-## Manual Way: download and install dependencies
-
-The following commands should download or update go-ipfs dependencies
-and then install them:
+|Command|Explanation|
+| ---: | :--- |
+|`SET PATH=%PATH%;\msys64\usr\bin`         |Add msys2's bin to `PATH`; Defaults to: (\msys64\usr\bin)|
```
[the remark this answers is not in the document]
djdv: Earlier we say "...we only need MSYS2's tools"
Maybe "Add msys2's bin to `PATH`" could be reworded to "Add msys2's tools to our `PATH`" to help imply this? I'll make this change for now.

docs/windows.md
```diff
@@ -1,142 +1,141 @@
...
+|`pacman --noconfirm -S  git make unzip`   |Install `go-ipfs` build dependencies|
+|`go get -u github.com/ipfs/go-ipfs`       |Fetch / Update `go-ipfs` source|
+|`cd %GOPATH%\src\github.com\ipfs\go-ipfs` |Change to `go-ipfs` directory|
+|`make install`                            |Build and install to `%GOPATH%\bin\ipfs.exe`|
+|`%GOPATH%\bin\ipfs.exe version --all`     |Test the built binary|
 
+**Tip**: You can permanently append the MSYS2 bin to your `PATH` so that on subsequent builds you can simply run
```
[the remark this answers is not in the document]
djdv: Because of the way `SETX` works, I've changed the 'tip' bit a little.
`SETX` doesn't actually set the variable for the current session, so what we want to do is, first `SET` `PATH` and then lock that in with `SETX`. In addition we don't need to specify the msys|cygwin path since it's already contained in `%PATH%`, the way it was written before could lead the user to accidentally append it twice if it was already set, now we expect it to be set ahead of time and optionally locked that way.
***
**Tip**: To avoid setting `PATH` every time (`SET PATH=%PATH%;\msys64\usr\bin`), you can lock it in permanently using `setx` after it's been set once:
```
SETX PATH %PATH%
```

djdv:
@Mr0grog 
No worries, the help is greatly appreciated!

I've updated the draft to reflect some of the changes with minor edits from myself.
15182be997df7216d03ccc6bb6d3ecf7bc565f56

https://github.com/ipfs/go-ipfs/pull/4691#discussion_r178377564
https://github.com/ipfs/go-ipfs/pull/4691#discussion_r178372305
https://github.com/ipfs/go-ipfs/pull/4691#discussion_r178375199

docs/windows.md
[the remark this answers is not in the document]
Mr0grog: I feel like “Add msys2's bin to PATH” and “Add msys2's tools to our PATH” are both equally clear here (although using the same wording in both is good). It was more that, if someone needs explanation for this line, they probably don’t know what it means to add them to “PATH.”

docs/windows.md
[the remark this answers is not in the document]
Mr0grog: Ohhhhhhh. That seems like a huge pain; thanks for the explanation. 😕

Mr0grog:
Looks good to me!

docs/windows.md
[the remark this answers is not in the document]
djdv: Ahh, good point. There should be a reference to https://ss64.com/nt/path.html somewhere here, but I'm not sure where to put it. Either after the phrase or maybe wrapping `PATH` as a link in that phrase.

djdv:
Thanks again for all the help @Mr0grog :^)
Not sure how long this ref will stick around but I made 2 minor changes
a24d1f5f31e8df7864c0af2995664123aff366b8
Bold consistency
**Tip**:  -> **Tip:** (colon is included)

and a references on `PATH` for the first instance in each section.
`PATH` -> [`PATH`](https://ss64.com/nt/path.html)

@whyrusleeping 
Unless there are other remarks by anyone, this may be good to go.

Mr0grog:
@djdv do you have any idea if the issue in #4905 (git line ending config) is common with Cygwin’s Git? Should we give that a mention in the troubleshooting section?

djdv:
@Mr0grog 
When writing this I made sure to go from a fresh Windows install to a built binary using only the steps listed, so by default Cygwins git should work for building, however I didn't test commiting changes with them.

I'll run through the set and see if any of them have CRLF defaults. It wouldn't hurt to have it as supplementary info anyway, I'm just unsure the title. "Go requires Unix-like linefeeds"

whyrusleeping:
LGTM, thanks @djdv

PR #4695, opened 2018-02-13
Fix: ipns_test: missing go-detect-race rewrite

hsanjuan:
Apparently missed this one rewrite yesterday and somehow tests didn't catch it.


whyrusleeping:
Ah, it didnt get caught because we don't run fuse tests on CI

PR #4696, opened 2018-02-13
Doc: golint-ify pin package

hsanjuan:
This removes all golint warnings in the `pin` package.

Due to stuttering, I had to rename `pin.PinMode` to `pin.Mode`. I can revert that specific commit though.

Protobuf stuff not touched.

### DIFF
--- a/pin/pin.go
+++ b/pin/pin.go
@@ -43,14 +43,14 @@ const (
 	linkAll       = "all"
 )
 
-// PinMode allows to specify different types of pin (recursive, direct etc.).
+// Mode allows to specify different types of pin (recursive, direct etc.).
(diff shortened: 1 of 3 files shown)


pin/pin.go
```diff
@@ -141,19 +151,19 @@ type Pinner interface {
 	InternalPins() []*cid.Cid
 }
 
+// Pinned represents CID which has been pinned with a pinning strategy.
+// The Via field allows to identify the pinning parent of this CID, in the
```
whyrusleeping: Phrasing here is a bit weird "allows to identify"

whyrusleeping:
@hsanjuan you're the best

Issue #4681, opened 2018-02-11
Ignore keys with invalid names in keystore


matrushka:
#### Version information:
go-ipfs version: 0.4.13-cc01b7f
Repo version: 6
System version: amd64/darwin
Golang version: go1.9.2

#### Type:
Enhancement

#### Description:
Right now dotfiles under the keystore directory causes an error while running `ipfs key list`. Although having a dotfile (or any file with an invalid name) shouldn't cause the whole thing to fail. It should print out a warning about the problem. When it comes to why I need this feature is:

##### Motivations:
* I want to deploy an IPFS node with a specific key inside it by using Kubernetes
* I want to use secret feature of Kubernetes to deploy the key to nodes I want to spin up (http://kubernetes.io/docs/concepts/configuration/secret/)
* I can deploy the key succesfully but kubernetes uses some dotfiles to deploy secret files which breaks all the key related functions.
* So this issue can be solved via either patching Kubernetes or go-ipfs but I thought the go-ipfs behaviour should be more forgiving and informative, since it's preventing access to valid keys and not informing the user about exactly why.

##### How the feature should work:
* While renaming or generating a key the proper error message is displayed so it's all good ✅
* If you have a key with an invalid name inside the keystore directory and `list` is being called it should return all the valid keys and log or display warnings about the invalid ones.

Stebalien:
Ignoring hidden files sounds reasonable.

Want to take a crack at this? It should be fairly straight forward. You'll have to edit `keystore/keystore.go` and make two changes:

1. `Has` should call the `validateName` function before checking if we have the file (so `Has(".specialFile")` doesn't indicate that we have a key named `.specialFile`.
2. `List` should filter the returned list of file names by `validateName` (and log bad files).

matrushka:
@Stebalien sure I'll try to have a look.

PR #4700, opened 2018-02-14
Ignore invalid key files in keystore directory.

matrushka:
Modified keystore to ignore invalid key files inside the keystore directory to address #4681

* Has calls the validateName function before checking if we have the file
* List filters the returned list of file names by validateName.

### DIFF
--- a/keystore/keystore.go
+++ b/keystore/keystore.go
@@ -161,7 +161,7 @@ func (ks *FSKeystore) List() ([]string, error) {
 		return nil, err
 	}
 
-	list := make([]string, 0)
+	list := make([]string, 0, len(dirs))
 
 	for _, name := range dirs {
 		err := validateName(name)
--- a/keystore/keystore.go
+++ b/keystore/keystore.go
@@ -77,6 +77,10 @@ func (ks *FSKeystore) Has(name string) (bool, error) {
 		return false, err
 	}
 
+	if err := validateName(name); err != nil {
+		return false, err
+	}
+
 	return true, nil
 }
 
@@ -149,5 +153,19 @@ func (ks *FSKeystore) List() ([]string, error) {
 		return nil, err
 	}
 
-	return dir.Readdirnames(0)
+	dirs, err := dir.Readdirnames(0)
+	if err != nil {
+		return nil, err
+	}
+
+	var list []string
+
+	for _, name := range dirs {
+		err := validateName(name)
+		if err == nil {
+			list = append(list, name)
+		}
+	}
+
+	return list, err
 }
--- a/keystore/keystore.go
+++ b/keystore/keystore.go
@@ -7,8 +7,11 @@ import (
 	"path/filepath"
 	"strings"
 
+	logging "gx/ipfs/QmRb5jh8z2E8hMGN2tkvs1yHynUanqnZ3UeKwgN1i9P1F8/go-log"
 	ci "gx/ipfs/QmaPbCnUMBohSGo3KnxEa2bHqyJVVeEEcwtqJAYxerieBo/go-libp2p-crypto"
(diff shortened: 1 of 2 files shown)


keystore/keystore_test.go
```diff
@@ -143,6 +144,62 @@ func TestKeystoreBasics(t *testing.T) {
 	}
 }
 
+func TestInvalidKeyFiles(t *testing.T) {
+	tdir, err := ioutil.TempDir("", "keystore-test")
+
+	if err != nil {
+		t.Fatal(err)
+	}
+
```
magik6k: `defer os.RemoveAll(tdir)`

keystore/keystore_test.go
```diff
@@ -146,6 +147,7 @@ func TestKeystoreBasics(t *testing.T) {
 
 func TestInvalidKeyFiles(t *testing.T) {
 	tdir, err := ioutil.TempDir("", "keystore-test")
+	defer os.RemoveAll(tdir)
```
magik6k: should be after the error check.

matrushka:
@magik6k by the way the formatting error was in a file I haven't touched at all. Also running `go fmt ./...` had no effect on that file as well (it's `fuse/ipns/ipns_test.go`).

magik6k:
Oh, for some reason un-fmt'ed code got merged to master, pushed a fix in https://github.com/ipfs/go-ipfs/pull/4701

keystore/keystore.go
```diff
@@ -149,5 +153,19 @@ func (ks *FSKeystore) List() ([]string, error) {
 		return nil, err
 	}
 
-	return dir.Readdirnames(0)
+	dirs, err := dir.Readdirnames(0)
+	if err != nil {
+		return nil, err
+	}
+
+	var list []string
```
Stebalien: Probably best to preallocate `list := make([]string, 0, len(dirs))`.

keystore/keystore.go
```diff
@@ -149,5 +153,19 @@ func (ks *FSKeystore) List() ([]string, error) {
...
+
+	for _, name := range dirs {
+		err := validateName(name)
+		if err == nil {
+			list = append(list, name)
+		}
+	}
+
+	return list, err
```
Stebalien: Returning the error here is a bit confusing. It should always be nil (the go idiom is to just return `list, nil`).

keystore/keystore.go
```diff
@@ -149,5 +153,19 @@ func (ks *FSKeystore) List() ([]string, error) {
...
+	dirs, err := dir.Readdirnames(0)
+	if err != nil {
+		return nil, err
+	}
+
+	var list []string
+
+	for _, name := range dirs {
+		err := validateName(name)
+		if err == nil {
```
Stebalien: Should probably log the error. You'll need to import the logger and construct it at the top of the file:

```go
import (
  // ...
  logging "gx/ipfs/QmRb5jh8z2E8hMGN2tkvs1yHynUanqnZ3UeKwgN1i9P1F8/go-log"
)

var log = logging.Logger("keystore")

// ...
````

keystore/keystore.go
matrushka: I'd leave out the capacity param since we don't know if all the dirs are gonna make it to the list.

Stebalien: Over allocating slightly is better than under allocating, in general (especially when we only over allocate in an error case). Probably not that much of an issue here as this function isn't frequently called but I know it'll continue to bug me.

matrushka: done 👍

Stebalien:
LGTM. Thanks! Mind rebasing to make the tests pass?

PR #4693, opened 2018-02-13
Create a raw node instead of a file node when there is no content.

kevina:
This fixes things so when raw-leaves are enabled a zero size file creates a zero size raw leaf.  ~~When raw-leaves are not enabled the hash created changes from `QmbFMke1KXqnYyBBWxB74N4c5SBnJMVAiMNRcGu6x1AwQH` to `Qmdsf68UUYTSSx3i4GtDJfxzpAEZt7Mp23m3qa36LYMSiW`, since the type field
changed from `TFile` to `TRaw`.~~

Closes #4688

PR #4701, opened 2018-02-14
Fix import order in ipns_test.go

magik6k:
See https://github.com/ipfs/go-ipfs/pull/4693#issuecomment-365623389


whyrusleeping:
Merge conflict already? Maybe this was fixed already?

whyrusleeping:
Yeah, looks like this was included in #4703 

Thanks anyways @magik6k :)

PR #4704, opened 2018-02-14
Feat: remove circular dependencies in merkledag package tests

hsanjuan:
This avoids using unixfs package and importer packages in
merkledag, which removes circular depedencies making it hard
to extract this module.

This comes pers @whyrusleeping suggestion on IRC:

```
22:20:28 <@whyrusleeping> hsanjuan: it would be great to work out those circular deps by using mocks of some sort
22:20:53 <@whyrusleeping> I think the merkledag_test stuff can be replaced with some code that just generates a dag manually
22:21:03 <@whyrusleeping> which is a little annoying, but needs to be done anyways
```

I'm very new here handling DAGs etc, so excuses in advance if something is terribly stupid.

### DIFF
--- a/merkledag/merkledag_test.go
+++ b/merkledag/merkledag_test.go
@@ -134,7 +134,10 @@ func makeTestDAG(t *testing.T, read io.Reader, ds ipld.DAGService) ipld.Node {
 	p := make([]byte, 512)
 	nodes := []*ProtoNode{}
 	var err error
-	_, err = io.ReadFull(read, p)
+	n, err = io.ReadFull(read, p)
+	if n != len(p) {
+		t.Fatal("should have read 512 bytes from the reader")
+	}
 	for err == nil {
 		protoNode := NodeWithData(p)
 		nodes = append(nodes, protoNode)
--- a/merkledag/merkledag_test.go
+++ b/merkledag/merkledag_test.go
@@ -133,19 +133,23 @@ func TestBatchFetchDupBlock(t *testing.T) {
 func makeTestDAG(t *testing.T, read io.Reader, ds ipld.DAGService) ipld.Node {
 	p := make([]byte, 512)
 	nodes := []*ProtoNode{}
-	var err error
-	n, err = io.ReadFull(read, p)
-	if n != len(p) {
-		t.Fatal("should have read 512 bytes from the reader")
-	}
-	for err == nil {
+
+	for {
+		n, err := io.ReadFull(read, p)
+		if err == io.EOF {
+			break
+		}
+
+		if err != nil {
+			t.Fatal(err)
+		}
+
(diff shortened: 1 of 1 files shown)


merkledag/merkledag_test.go
```diff
@@ -129,26 +126,73 @@ func TestBatchFetchDupBlock(t *testing.T) {
 	runBatchFetchTest(t, read)
 }
 
+// makeTestDAG creates a simple DAG from the data in a reader.
+// First, a node is created from each 512 bytes of data from the reader
+// (like a the Size chunker would do). Then all nodes are added as children
+// to a root node, which is returned.
+func makeTestDAG(t *testing.T, read io.Reader, ds ipld.DAGService) ipld.Node {
```
Stebalien: This should be using `io.ReadFull` (`Read` isn't guaranteed to fill the buffer). It should probably also handle short reads. It may currently "work" but I'm worried that users will misuse it.

merkledag/merkledag_test.go
```diff
@@ -181,11 +225,9 @@ func runBatchFetchTest(t *testing.T, read io.Reader) {
 			if !ok {
 				errs <- ErrNotProtobuf
 			}
-
-			read, err := uio.NewDagReader(ctx, firstpb, dagservs[i])
-			if err != nil {
-				errs <- err
-			}
+			_ = firstpb
+			_ = expected
```
Stebalien: We shouldn't need these lines.

merkledag/merkledag_test.go
```diff
@@ -129,26 +126,76 @@ func TestBatchFetchDupBlock(t *testing.T) {
...
+	nodes := []*ProtoNode{}
+	var err error
+	n, err = io.ReadFull(read, p)
+	if n != len(p) {
+		t.Fatal("should have read 512 bytes from the reader")
+	}
+	for err == nil {
+		protoNode := NodeWithData(p)
+		nodes = append(nodes, protoNode)
+		_, err = read.Read(p)
```
Stebalien: Same. Needs a ReadFull. I'd just deduplicate the code and have a `for {` loop with a break/return.

hsanjuan:
Thanks @Stebalien

Stebalien:
Sufficiently paranoid :smile:. LGTM.

whyrusleeping:
@hsanjuan would you be offended if I don't merge this one until we ship 0.4.14?

hsanjuan:
@whyrusleeping when will that be?

whyrusleeping:
@hsanjuan as soon as #4751 is in

Stebalien:
@whyrusleeping FYI, this is *just* a test change.

whyrusleeping:
Thanks @hsanjuan!

Issue #4705, opened 2018-02-15
merkledag\coding.go change links order if it have names in file nodes.


ivan386:
#### Version information:
go-ipfs version: 0.4.13-
Repo version: 6
System version: amd64/windows
Golang version: go1.9.2

#### Description:

Wrong order:
```
>ipfs dag get QmQigZekUpTrhwzoZ1zyfuq3kLvsBRR73chktinPhkewqG
{"data":"CAIYgGQggFwggAQggAQ=",
"links":[
{"Name":"","Size":512,"Cid":{"/":"zb2rhX9DvypVUCrp1SyoB6JpvdNEGcujJm7RZ6btt1zJBvnKq"}},
{"Name":"","Size":512,"Cid":{"/":"zb2rhX9DvypVUCrp1SyoB6JpvdNEGcujJm7RZ6btt1zJBvnKq"}},
{"Name":"test","Size":13388,"Cid":{"/":"QmYPWmgDtUNwLN1CfWnRWcEnfBk1ANLXAeYeoHeCuNNWFE"}}
]}
```

Right order:
```
>ipfs dag get QmQigZekUpTrhwzoZ1zyfuq3kLvsBRR73chktinPhkewqG
{"data":"CAIYgGQggFwggAQggAQ=",
"links":[
{"Name":"test","Size":13388,"Cid":{"/":"QmYPWmgDtUNwLN1CfWnRWcEnfBk1ANLXAeYeoHeCuNNWFE"}},
{"Name":"","Size":512,"Cid":{"/":"zb2rhX9DvypVUCrp1SyoB6JpvdNEGcujJm7RZ6btt1zJBvnKq"}},
{"Name":"","Size":512,"Cid":{"/":"zb2rhX9DvypVUCrp1SyoB6JpvdNEGcujJm7RZ6btt1zJBvnKq"}}
]}
```

It is file node. Sorting breaks the file.

https://github.com/ipfs/go-ipfs/blob/28acec2adb7c7cfe30c458b9fd7bf70768d9e26c/merkledag/coding.go#L37
https://github.com/ipfs/go-ipfs/blob/28acec2adb7c7cfe30c458b9fd7bf70768d9e26c/merkledag/coding.go#L61
https://github.com/ipfs/go-ipfs/blob/28acec2adb7c7cfe30c458b9fd7bf70768d9e26c/merkledag/coding.go#L80

whyrusleeping:
For file nodes you should not have link names.

[...]

PR #4715, opened 2018-02-17
Disable sorting on unmarshalled blocks and json. 

ivan386:
Allow to disable sorting in new ProtoNode.

Fix for: [merkledag\coding.go change links order if it have names in file nodes.](https://github.com/ipfs/go-ipfs/issues/4705)

License: MIT
Signed-off-by: Ivan ivan386@users.noreply.github.com

### DIFF
--- a/core/coreapi/object_test.go
+++ b/core/coreapi/object_test.go
@@ -237,11 +237,11 @@ func TestObjectAddLink(t *testing.T) {
 		t.Errorf("unexpected number of links: %d", len(links))
 	}
 
-	if links[0].Name != "abc" {
+	if links[0].Name != "bar" {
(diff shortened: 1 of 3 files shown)


whyrusleeping:
We can't allow this. This will cause hash differences between users of different versions.

ivan386:
If change links order in block this change it hash. It not happen because links already sorted. By default sort is on for new ProtoNode. Maybe need option nosort for json (dag put, object put) instead of disabling by default? But for unmarshalled block nosort must be on.

magik6k:
Blocks with unsorted links shouldn't be considered valid imo.

(closing due to inactiviny, feel free to continue the discussion)

PR #4725, opened 2018-02-21
make files stat --with-local return immediately when no data is present

MichaelMure:
Currently when the daemon is running, asking for `ipfs files stat QmDoesntExistLocally` will hang instead of returning immediately. This PR fix that by using the offline DAGService earlier. I guess I got confused in the test I was doing when implementing that the third time :-|

License: MIT

### DIFF
--- a/core/commands/files.go
+++ b/core/commands/files.go
@@ -125,7 +125,20 @@ var filesStatCmd = &cmds.Command{
 			return
 		}
 
-		nd, err := getNodeFromPath(req.Context, node, path)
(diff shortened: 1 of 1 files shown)


MichaelMure:
test failure is unrelated

Stebalien:
Test failure is https://github.com/ipfs/go-ipfs/issues/4726 (confirmed unrelated).

MichaelMure:
Can we merge ?

whyrusleeping:
Thanks @MichaelMure, getting this one into 0.4.14 since the feature was introduced in this release.

PR #4733, opened 2018-02-23
Add options for record count and timeout for resolving DHT paths

dirkmc:
License: MIT
Signed-off-by: Dirk McCormick <dirkmdev@gmail.com>

### DIFF
--- a/core/commands/ipns.go
+++ b/core/commands/ipns.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"io"
 	"strings"
+	"time"
 
 	cmds "github.com/ipfs/go-ipfs/commands"
 	e "github.com/ipfs/go-ipfs/core/commands/e"
@@ -57,9 +58,10 @@ Resolve the value of a dnslink:
 	Options: []cmdkit.Option{
 		cmdkit.BoolOption("recursive", "r", "Resolve until the result is not an IPNS name."),
 		cmdkit.BoolOption("nocache", "n", "Do not use cached entries."),
+		cmdkit.UintOption("dht-record-count", "dhtrc", "Number of records to request for DHT resolution."),
+		cmdkit.UintOption("dht-timeout", "dhtt", "Timeout in seconds for DHT resolution. Pass 0 for no timeout."),
 	},
 	Run: func(req cmds.Request, res cmds.Response) {
-
 		n, err := req.InvocContext().GetNode()
 		if err != nil {
 			res.SetError(err, cmdkit.ErrNormal)
@@ -107,16 +109,24 @@ Resolve the value of a dnslink:
 		}
 
 		recursive, _, _ := req.Option("recursive").Bool()
-		depth := 1
-		if recursive {
-			depth = namesys.DefaultDepthLimit
+		rc, rcok, _ := req.Option("dht-record-count").Int()
+		dhtt, dhttok, _ := req.Option("dht-timeout").Int()
+		opts := namesys.DefaultResolveOpts()
+		if !recursive {
+			opts.Depth = 1
+		}
+		if rcok {
+			opts.DhtRecordCount = uint(rc)
+		}
+		if dhttok {
+			opts.DhtTimeout = time.Duration(dhtt) * time.Second
 		}
 
 		if !strings.HasPrefix(name, "/ipns/") {
 			name = "/ipns/" + name
 		}
 
-		output, err := resolver.ResolveN(req.Context(), name, depth)
+		output, err := resolver.Resolve(req.Context(), name, opts)
 		if err != nil {
 			res.SetError(err, cmdkit.ErrNormal)
 			return
--- a/core/coreapi/name.go
+++ b/core/coreapi/name.go
@@ -117,16 +117,16 @@ func (api *NameAPI) Resolve(ctx context.Context, name string, opts ...caopts.Nam
 		resolver = namesys.NewNameSystem(n.Routing, n.Repo.Datastore(), 0)
 	}
 
-	depth := 1
-	if options.Recursive {
-		depth = namesys.DefaultDepthLimit
+	ropts := namesys.DefaultResolveOpts()
+	if !options.Recursive {
+		ropts.Depth = 1
 	}
 
 	if !strings.HasPrefix(name, "/ipns/") {
 		name = "/ipns/" + name
 	}
 
-	output, err := resolver.ResolveN(ctx, name, depth)
+	output, err := resolver.Resolve(ctx, name, ropts)
 	if err != nil {
 		return nil, err
 	}
--- a/namesys/ipns_validate_test.go
+++ b/namesys/ipns_validate_test.go
@@ -115,7 +115,7 @@ func TestResolverValidation(t *testing.T) {
 	}
 
 	// Resolve entry
-	resp, err := resolver.resolveOnce(ctx, id.Pretty())
+	resp, err := resolver.resolveOnce(ctx, id.Pretty(), DefaultResolveOpts())
 	if err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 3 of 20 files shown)


dirkmc:
This PR adds two new parameters to `ipfs name resolve` and `ipfs resolve`:
`ipfs name resolve <hash> --dht-record-count=5 --dht-timeout=20`

Currently when resolving 16 IPNS records will be retrieved from the DHT (where the DHT k-value is 20). The `dht-record-count` parameter allows clients to retrieve less IPNS records from the DHT when resolving.

Currently the timeout for retrieving a public key for an IPNS record is one minute, and the timeout for retrieving the value itself is one minute. The public key and value are retrieved sequentially so it can take up to 2 minutes to time out. The `dht-timeout` parameter allows clients to set a timeout for retrieving both values (eg maximum of 30s to retrieve both).

dirkmc:
@vyzo @Stebalien @whyrusleeping could I ask you guys to take a look at this?
Thanks,
Dirk

core/commands/ipns.go
```diff
@@ -107,16 +109,24 @@ Resolve the value of a dnslink:
...
-		if recursive {
-			depth = namesys.DefaultDepthLimit
+		rc, rcok, _ := req.Option("dht-record-count").Int()
+		dhtt, dhttok, _ := req.Option("dht-timeout").Int()
+		opts := namesys.DefaultResolveOpts()
+		if !recursive {
+			opts.Depth = 1
+		}
+		if rcok {
+			opts.DhtRecordCount = uint(rc)
```
dirkmc: Can get rid of this cast once https://github.com/ipfs/go-ipfs-cmdkit/pull/14 is approved

dirkmc:
My concern was that we would then have three different methods for doing the same thing. Do you think that's not worth worrying about?

vyzo:
My suggestion was more to add the new method that takes the options, and redirect the old method to the new one with default options. That way you wouldn't change the existing code at all.

dirkmc:
I understand, my concern is about having three different methods in the API where we only really need one. Is that not worth worrying about?

vyzo:
Hrm, dunno -- it's nice to have a simple method that takes default options.

whyrusleeping:
When in doubt, variadic options can be a good interface

dirkmc:
Yes you're (both) right, that is nicer, I'll use variadic options

dirkmc:
Ok changed to use variadic options, let me know if that makes sense

core/commands/ipns.go
```diff
@@ -107,16 +110,24 @@ Resolve the value of a dnslink:
...
+		}
+		if dhttok {
+			ropts = append(ropts, nsopts.DhtTimeout(time.Duration(dhtt)*time.Second))
 		}
 
 		if !strings.HasPrefix(name, "/ipns/") {
 			name = "/ipns/" + name
 		}
 
-		output, err := resolver.ResolveN(req.Context(), name, depth)
```
whyrusleeping: where did the depth arg go?

whyrusleeping: oh nvm, i found it

core/coreapi/name.go
```diff
@@ -117,16 +118,16 @@ func (api *NameAPI) Resolve(ctx context.Context, name string, opts ...caopts.Nam
...
-	if options.Recursive {
-		depth = namesys.DefaultDepthLimit
-	}
-
 	if !strings.HasPrefix(name, "/ipns/") {
 		name = "/ipns/" + name
 	}
 
-	output, err := resolver.ResolveN(ctx, name, depth)
+	ropts := []nsopts.ResolveOpt{}
```
whyrusleeping: i prefer writing this as:
`var ropts []nsopts.ResolveOpt`

dirkmc:
I put it in a separate package as a way of namespacing, so that you call `nsopts.Depth(1)` instead of `namespace.Depth(1)`, so that it's clear those methods are working with options. If you think that's not an issue I'm happy to put it all in the top level namesys package.

vyzo:
Fair enough -- not really an issue :)

core/commands/ipns.go
```diff
@@ -57,9 +59,10 @@ Resolve the value of a dnslink:
 	Options: []cmdkit.Option{
 		cmdkit.BoolOption("recursive", "r", "Resolve until the result is not an IPNS name."),
 		cmdkit.BoolOption("nocache", "n", "Do not use cached entries."),
+		cmdkit.UintOption("dht-record-count", "dhtrc", "Number of records to request for DHT resolution."),
+		cmdkit.UintOption("dht-timeout", "dhtt", "Timeout in seconds for DHT resolution. Pass 0 for no timeout."),
```
Stebalien: So, I might be missing something but IPFS commands *have* a timeout flag (e.g., `ipfs --timeout=30s`). And we can always set a DHT resolution timeout by setting a timeout on the context. What's the usecase for the additional timeout?

dirkmc: The use case is that with --dht-timeout the DHT collects as many values as it can within the timeout, and then returns the best value it has found. The global IPFS command --timeout will simply kill the command:
```
> ./cmd/ipfs/ipfs name resolve --timeout=5s QmSgxPeqLrnM1ZB1pno3tYMQfqLB4cWXuAQNfSWsk77apK
Error: Post http://127.0.0.1:5001/api/v0/name/resolve?arg=QmSgxPeqLrnM1ZB1pno3tYMQfqLB4cWXuAQNfSWsk77apK&encoding=json&stream-channels=true&timeout=5s: context deadline exceeded
> ./cmd/ipfs/ipfs name resolve --dhtt=5 QmSgxPeqLrnM1ZB1pno3tYMQfqLB4cWXuAQNfSWsk77apK
/ipfs/QmUNLLsPACCz1vLxQVkXqqLX5R1X345qqfHbsf67hvA3Nn
```

dirkmc: Maybe it would be better for the `--dht-timeout` parameter to take a string though, eg `--dht-timeout=5s` instead of assuming seconds eg `--dht-timeout=5`

Stebalien: I see. It would be nice to explain that in the help but I can't think of a succinct way of putting it...

Kubuxu: @dirkmc it will be better for it to parse the time.

dirkmc: I added a commit to parse the time, and changed the parameter description to
`Max time to collect values during DHT resolution eg \"30s\". Pass 0 for no timeout`
That's the most succinct way I can think of to say it

namesys/ipns_validate_test.go
```diff
@@ -136,9 +137,9 @@ func TestResolverValidation(t *testing.T) {
 	}
 
 	// Record should fail validation because entry is expired
-	_, err = resolver.resolveOnce(ctx, id.Pretty())
-	if err != ErrExpiredRecord {
-		t.Fatal("ValidateIpnsRecord should have returned ErrExpiredRecord")
+	_, err = resolver.resolveOnce(ctx, id.Pretty(), opts.DefaultResolveOpts())
+	if err == nil {
```
whyrusleeping: Why arent we checking for a specific error anymore?

dirkmc: The `resolveOnce()` [method used to call](https://github.com/ipfs/go-ipfs/pull/4733/files#diff-36d8e88d65c9668b43e0c41d99084a99L154) `routing.GetValue()` but [now we call](https://github.com/ipfs/go-ipfs/pull/4733/files#diff-36d8e88d65c9668b43e0c41d99084a99R192) `routing.GetValues()` directly.
The mock `ValueStore` implementation's `GetValue()` method would get a single record and call `VerifyRecord()` on it, which would return `ErrExpiredRecord`/`ErrSignature` etc. We now call the mock `ValueStore` implementation's [GetValues()](https://github.com/ipfs/go-ipfs/pull/4733/files#diff-a822f7067375edcc1e809bf61ae43c6dL265) method, which gets multiple records, calls `VerifyRecord()` on each one individually, and adds the record to the list of valid records if there's no error. If there are no valid records it returns `routing.ErrNotFound`.

dirkmc: I looked into whether I can check for `routing.ErrNotFound` but the go-ipfs-routing mock returns `datastore.ErrNotFound` instead of `routing.ErrNotFound` when it doesn't find a key:
https://github.com/ipfs/go-ipfs-routing/blob/master/mock/centralized_client.go#L46

whyrusleeping: Alright, that seems fine. Thanks for the response!

PR #3314, opened 2016-10-18
Create a FilestoreNode object to carry PosInfo

kevina:
Part of #2634.

PR #4734, opened 2018-02-23
[WIP] Add Filestore support for the trickle DAG builder

### LINKED ISSUE #4052: [add] --nocopy --trickle; copies data and ignores trickle
#### Version information:
go-ipfs version: 0.4.10-00573cb2c
Repo version: 5
System version: amd64/windows
Golang version: go1.8.3

#### Type:
Bug

#### Severity:
Low-Medium
#### Description:
On my machine if I try to use --nocopy with the trickledag format it will still copy file blocks, as well as output a hash for the standard dag format, but even this behaviour is inconsistent. These are the cases I've tried and their results with a log below, the sample data is just 25M from udevrandom.

Nothing has been done yet, file is added with --nocopy
Result: As expected, file data is not copied to the datastore, standard hash returned
File has been added via --nocopy already, it is added again with --nocopy and -t
Result: File data is not copied, but standard dag hash returned, not trickle
Repo is cleaned, file is added with --nocopy and -t
Result: File data is copied, and standard dag hash returned
Repo is cleaned, file added with -t
Result: As expected,  file data is copied to the datastore, trickle hash returned

In addition to this please look at the log below to see the blocks directory changing size, even after everything is removed it seems to have grown is size a few kb. I'm not sure if this is expected or not.
I haven't tested this elsewhere yet. I originally caught this after adding a lot of files like this `add --nocopy -r -w -Q --chunker=rabin --pin=false --cid-version=1 --hash=Blake2b-256`, -t was only provided conditionally (if more than half the files provided were multimedia files) so the arguments would turn into `add --nocopy -r -w -Q --chunker=rabin --pin=false --cid-version=1 --hash=Blake2b-256 -t`, I don't think any of the other arguments there affect --nocopy like this.

```
>cd %HOMEPATH%\.ipfs

>ipfs init
...

>ipfs config --json Experimental.FilestoreEnabled true

>cksum random
2009307347 26214400 random

>du -sh blocks
68K	blocks

>ipfs add --pin=false --nocopy random
added QmYt8DnHxMa2Jo9mcoYMD4rqjobHqhjwTWS4HUXtxuVtMr random

>du -sh blocks
77K	blocks

>ipfs get --output=validate QmYt8DnHxMa2Jo9mcoYMD4rqjobHqhjwTWS4HUXtxuVtMr
Saving file(s) to validate
>cksum validate
2009307347 26214400 validate

>ipfs add --pin=false --nocopy -t random
added QmYt8DnHxMa2Jo9mcoYMD4rqjobHqhjwTWS4HUXtxuVtMr random

>du -sh blocks
77K	blocks

>ipfs get --output=validate QmYt8DnHxMa2Jo9mcoYMD4rqjobHqhjwTWS4HUXtxuVtMr
Saving file(s) to validate
>cksum validate
2009307347 26214400 validate

>ipfs repo gc
removed QmYt8DnHxMa2Jo9mcoYMD4rqjobHqhjwTWS4HUXtxuVtMr
...
removed zb2rhoTbL7rcs2J8enZ5R54gijwPdLGGHS3JtpFw74vDw8MyZ

>du -sh blocks
55K	blocks

>ipfs add --pin=false --nocopy -t random
added QmYt8DnHxMa2Jo9mcoYMD4rqjobHqhjwTWS4HUXtxuVtMr random

>du -sh blocks
26M	blocks

>ipfs get --output=validate QmYt8DnHxMa2Jo9mcoYMD4rqjobHqhjwTWS4HUXtxuVtMr
Saving file(s) to validate
>cksum validate
2009307347 26214400 validate

>ipfs repo gc
removed zb2rhanQbgX9Q4uzrtDC5MtsukDRLBPLPwR4XvYNaJFHpAvSt
...
removed zb2rhdQWSLuJLhMQ7kwEpRT3pGHeHcMbHjFBJo7DmfNGiWfSF

>du -sh blocks
67K	blocks

>ipfs add --pin=false -t random
added QmYDL6Rnkhhog6NFwykAtAD2uT5R62PVBB4F9t85vUX6NM random

>ipfs get --output=validate QmYDL6Rnkhhog6NFwykAtAD2uT5R62PVBB4F9t85vUX6NM
Saving file(s) to validate
>cksum validate
2009307347 26214400 validate

>ipfs repo gc
removed QmWLPNh6mVuqsxpWTiW7YNaGWbMjmxfUUweJ18xaiBfn1s
...
removed QmdtwfJv8g6FQxAnHFVPpNxiasitZGvGzoVtSuScswQZfG

>du -sh blocks
79K	blocks
```

schomatis:
This needs to be updated after https://github.com/ipfs/go-ipfs/issues/5135 is done. Any help from the community is welcome.

---------------
This is a work in progress, not ready to be merged.

As [commented](https://github.com/ipfs/go-ipfs/issues/4052#issuecomment-368146394) in #4052, the support for Filestore will be added only to the `Layout` function for now, not to `Append`. This means that `ipfs add --trickle` will support `--nocopy` (Filestore), but that added node when copied to MFS won't support `ipfs files write` (which uses `Append`).

It builds on top of the #4730 that should be merged first.

Closes #4052.

### DIFF
--- a/importer/helpers/dagbuilder.go
+++ b/importer/helpers/dagbuilder.go
@@ -134,7 +134,10 @@ func (db *DagBuilderHelper) newUnixfsBlock() *UnixfsNode {
 
 // FillNodeLayer will add datanodes as children to the give node until
 // at most db.indirSize nodes are added.
-func (db *DagBuilderHelper) FillNodeLayer(node *UnixfsNode) error {
+// The parameter offset is used to support Filestore and applies only to
+// nodes stored using this feature, it is adjusted with the sizes of the
+// added nodes.
+func (db *DagBuilderHelper) FillNodeLayer(node *UnixfsNode, offset *uint64) error {
 
 	// while we have room AND we're not done
 	for node.NumChildren() < db.maxlinks && !db.Done() {
@@ -142,10 +145,12 @@ func (db *DagBuilderHelper) FillNodeLayer(node *UnixfsNode) error {
 		if err != nil {
 			return err
 		}
+		db.SetPosInfo(child, *offset)
 
 		if err := node.AddChild(child, db); err != nil {
 			return err
 		}
+		*offset += child.FileSize()
 	}
 
 	return nil
--- a/importer/trickle/trickledag.go
+++ b/importer/trickle/trickledag.go
@@ -38,19 +38,8 @@ const layerRepeat = 4
 // explanation.
 func Layout(db *h.DagBuilderHelper) (ipld.Node, error) {
 	root := db.NewUnixfsNode()
-	if err := db.FillNodeLayer(root); err != nil {
+	if err := fillTrickleRec(db, root, -1); err != nil {
 		return nil, err
-	}
-	for level := 1; !db.Done(); level++ {
-		for i := 0; i < layerRepeat && !db.Done(); i++ {
-			next := db.NewUnixfsNode()
-			if err := fillTrickleRec(db, next, level); err != nil {
-				return nil, err
-			}
-			if err := root.AddChild(next, db); err != nil {
-				return nil, err
-			}
-		}
 	}
 
 	out, err := db.Add(root)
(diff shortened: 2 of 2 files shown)


schomatis:
Due to this partial Filestore support (for `Layout` but not for `Append`) I'm not sure I can just add the `SetPosInfo` functionality (and add the `offset` parameter) to the support functions `fillTrickleRec` and `FillNodeLayer` as it was done for the [balanced builder](https://github.com/ipfs/go-ipfs/pull/3314/files#diff-8dd9373aec43cd2bfb199aac6fd93780), see this [commit](https://github.com/ipfs/go-ipfs/pull/4734/commits/179854a2300a968e9e1541422dc0871f7a178c9b) of this PR as an example (which doesn't build because the calls from `Append` weren't adapted). I'm thinking of a few possibilities:

1. Just add `SetPosInfo` for `Append` as it won't make a difference because it won't be called for Filestore nodes. This makes it more homogeneous (`Layout` and `Append` can use the same support functions) but it will be unnecessarily modifying `Append` (as it's not defined how, and if, `Append` will support Filestore).

2. Overload `fillTrickleRec` (and similarly `FillNodeLayer`) to call them from `Append` with "unset"  parameters (e.g., `offset: -1`) that will cause it not to use `SetPosInfo` in that case. It won't be very nice, it will add many `if` clauses.

3. Split the functionality of `fillTrickleRec` in two separate functions, one with Filestore support (with calls to `SetPosInfo`) and  one without (do the same for `FillNodeLayer`). This will likely cause some code duplication.

@kevina WDYT?

kevina:
@schomatis I don't know enough about the trickle DAG builder to make to know for sure.  I would add some test to `t0270-filestore.sh` that use the `add --chunker=rabin --nocopy ...`.

schomatis:
Another (4th) possibility is to refactor the Filestore functionality inside `DagBuilderHelper`, taking this responsibility away from the functions that rely upon the helper.

The function `SetPosInfo` is already contained by it, and (following the balanced DAG builder logic) `SetPosInfo` is always called after `DagBuilderHelper.NewUnixfsNode()` (see [1](https://github.com/ipfs/go-ipfs/blob/master/importer/balanced/builder.go#L35), [2](https://github.com/ipfs/go-ipfs/blob/master/importer/balanced/builder.go#L94)), so the `offset` could be a variable inside the `DagBuilderHelper` structure and it's value would be updated after each call to [`DagBuilderHelper.GetNextDataNode`](https://github.com/ipfs/go-ipfs/blob/master/importer/balanced/builder.go#L82) (instead of getting it indirectly from the node's [`FileSize`](https://github.com/ipfs/go-ipfs/blob/master/importer/balanced/builder.go#L104)).

This option seems more organic to me because it's the helper the centralized point where data nodes are obtained from, and regardless of how they are organized (balanced, trickle, etc.) their offset will be the same (as long as the nodes are processed in a FIFO order). I would still need some pointers on how to organize this refactoring though.

schomatis:
@whyrusleeping WDYT? Could you give me some recommendations here please?

whyrusleeping:
Hey @schomatis, I'm a bit busy working on the 0.4.14 release. If I don't review this in the next couple days please do ping me again.

Also maybe @Kubuxu can help

Kubuxu:
I will try to take a look at this after the weekend.

schomatis:
This code has been moved to `go-unixfs` and with the trickle importer now abandoning the `UnixfsNode` structure this PR adds little value to justify an extraction there.

PR #4738, opened 2018-02-26
remove a spurious debug message

whyrusleeping:
Not sure how this snuck through, but it needs to go

License: MIT


Kubuxu:
Did it even build with it?

whyrusleeping:
Yeah, its a weird golang builtin thing.

PR #4751, opened 2018-03-01
Disallow usage of unsafe hashes for reading and adding content

Kubuxu:
Should be done


Kubuxu:
Short observations:
node.Exchange is not used directly to get blocks
 - it is passed to blockservice
 - commands/add.go uses it directly but passes it to blockservice. `//hash security 001`
 - thus blocked the blockservice

Kubuxu:
Random failure in bitswap:
```
21:15:20.663 ERROR    bitswap: nil cid in GetBlock get.go:18

--- FAIL: TestBasicBitswap (0.01s)

	bitswap_test.go:318: Test a one node trying to get one block from another

	bitswap_test.go:361: stat node 0

	bitswap_test.go:363: stat node 1

	bitswap_test.go:296: mismatch in blocks sent: 0 vs 1

	bitswap_test.go:300: mismatch in blocks recvd: 1 vs 2

	bitswap_test.go:304: mismatch in data sent: 0 vs 1

	bitswap_test.go:308: mismatch in data recvd: 1 vs 2

	bitswap_test.go:365: stat node 2

	bitswap_test.go:372: [Block QmTTA2daxGqo5denp6SwLzzkLJm3fuisYEi9CoWsuHpzfb]

FAIL

FAIL	github.com/ipfs/go-ipfs/exchange/bitswap	4.173s
```
Passes locally

Kubuxu:
Sharness has passed. Restarting tests to make them green.

whyrusleeping:
Lets not call the package removeme. Lets just put it in thirdparty (Which is already synonymous for removeme at this point).

blockservice/blockservice.go
```diff
@@ -149,6 +155,13 @@ func (s *blockService) AddBlock(o blocks.Block) error {
 }
 
 func (s *blockService) AddBlocks(bs []blocks.Block) error {
+	//hash security
```
whyrusleeping: for consistency, add a space before the word hash

magik6k:
@whyrusleeping 
Kubuxu here (I'm at Magik's atm) thus I'm using his account.
I've added some sharness tests.

I think there might be one issue with this whole code, GC might be failing when there are insecure blocks in the bockstore.
I won't have time to fix it today, I can try fixing it tomorrow morning.

kevina:
Regarding GC, **if a block can not be read it is unsafe to continue** as that block could potentially point to other blocks that should not be removed.

There are two solutions (1) somehow allow reading of blocks with insecure hashes, (2) force a repo migration, have the the tool check for insecure blocks and then refuse to upgrade until those blocks are dealt with by the user, the tool should also check if the insure blocks point to secure blocks and inform the user of the situation.  We could also also provide a separate tool that will automatically convert insecure hashes to a secure hash of there choosing.

blockservice/blockservice.go
```diff
@@ -189,10 +202,15 @@ func (s *blockService) GetBlock(ctx context.Context, c *cid.Cid) (blocks.Block,
 		f = s.exchange
 	}
 
-	return getBlock(ctx, c, s.blockstore, f)
+	return getBlock(ctx, c, s.blockstore, f) // hash security
 }
 
 func getBlock(ctx context.Context, c *cid.Cid, bs blockstore.Blockstore, f exchange.Fetcher) (blocks.Block, error) {
+	err := verifcid.ValidateCid(c)
```
Stebalien: Add `// hash security` comment for future grep.

core/commands/add.go
```diff
@@ -258,7 +258,7 @@ You can now check what blocks have been created by:
 			exch = offline.Exchange(addblockstore)
 		}
 
-		bserv := blockservice.New(addblockstore, exch)
+		bserv := blockservice.New(addblockstore, exch) // hash security 001
```
Stebalien: Does this need the hash security comment?

magik6k: This is the only place `n.Exchnage` is used standalone.

kevina:
As this code still allows the use of CID's with insecure hashes, a potentially easy solution to the GC problem allow the reading of insure hashes, but disallow the publishing or writing of insecure hashes.  This will also provide a gentler upgrade path for the rare case when a user has a repo with insecure hashes.

Another solution to _just_ the GC problem is to allow the reading of insecure hashes in the local blockstore, but not the blockservice.  This _may_ also allow the user to read them locally, but I am not sure as I think the blockservice is always used even in offline mode.

kevina:
@Kubuxu @whyrusleeping as I said in IRC I can take over this P.R. if it will help.

Kubuxu:
GC improvements should be done.
If insecure cid are pinned, GC will fail with following error message:
```
Error: "potentially insecure hash functions not allowed"
Please run 'ipfs pin verify to list insecure hashes. If you want to read them, please downgrade your go-ipfs to 0.4.13
garbage collection aborted: could not retrieve some links
```
When user runs `ipfs pin verify --verbose` she will see:
```
> ipfs pin verify --verbose
zDvnoKZMLRv broken
  zDvnoKZMLRv: potentially insecure hash functions not allowed
QmS4ustL54uo8FzR9455qaxZwuMiUhyvMcX9Ba8nUH4uVv ok
QmUNLLsPACCz1vLxQVkXqqLX5R1X345qqfHbsf67hvA3Nn ok
```

To resolve this, she has run:
```
> ipfs pin verify -q | ipfs pin rm
unpinned zDvnoKZMLRv
```
with daemon running.

core/commands/pin.go
```diff
@@ -462,25 +463,23 @@ var verifyPinCmd = &cmds.Command{
 		cmds.Text: func(res cmds.Response) (io.Reader, error) {
 			quiet, _, _ := res.Request().Option("quiet").Bool()
 
-			outChan, ok := res.Output().(<-chan interface{})
+			out, err := unwrapOutput(res.Output())
```
whyrusleeping: this looks like its being changed to only emit a single output. Whats going on here?

test/sharness/t0275-cid-security.sh
```diff
@@ -0,0 +1,61 @@
...
+}
+
+
+# should work offline
+test_cat_get
+
+# should work online
+test_launch_ipfs_daemon
+test_cat_get
+test_kill_ipfs_daemon
```
whyrusleeping: Could you add a test in here that manually puts a bad hash into the datastore, and then runs a gc?

whyrusleeping:
This tenatively looks good to me. I'd like to see a few more tests before moving forward though.

core/commands/pin.go
kevina: I would like to know what is going on also.

whyrusleeping:
@Stebalien want to give the gc changes another lookover here?

core/commands/pin.go
Kubuxu: The marshaller is called for every message pushed down the channel. I am not sure why the legacy layer is doing this but it prints out all the results.

Kubuxu:
@whyrusleeping I've added test for general GC but I am unable to test pin set cases. This is because there is no way to add insecure block to pinset.

kevina:
@Kubuxu if there are insecure hashes will those end up being published?

Kubuxu:
@kevina you mean by reprovider? Probably yes.

Kubuxu:
@kevina reprovider won't be providing insecure hashes anymore

kevina:
@Kubuxu okay, We should be sure and mark any places where Cid's are verified so that they can cleanly be removed once the strict Cid code lands.

blockservice/blockservice.go
```diff
@@ -10,6 +10,7 @@ import (
 	"io"
 
 	exchange "github.com/ipfs/go-ipfs/exchange"
+	"github.com/ipfs/go-ipfs/thirdparty/verifcid"
```
hsanjuan: We should not be putting more stuff in `thirdparty`. If it needs to be re-used across submodules or has interest for other places put it in a different repo, if not, place it in the submodule directly. It complicates extractions afterwards...

core/builder.go
```diff
@@ -17,6 +17,7 @@ import (
 	pin "github.com/ipfs/go-ipfs/pin"
 	repo "github.com/ipfs/go-ipfs/repo"
 	cfg "github.com/ipfs/go-ipfs/repo/config"
+	"github.com/ipfs/go-ipfs/thirdparty/verifbs"
```
hsanjuan: This too

blockservice/blockservice.go
Kubuxu: This is temporary measure, it will be removed as we are able to propagate go-cid.
The temporary measure was applied because we want to release ASAP and have this behaviour fixed in next release.

hsanjuan: Then place that code as submodules of `blockservice`? That way the temporary measure can go together with the temporary code and be fixed with it when it's extracted. Otherwise I will have to do exactly that.

whyrusleeping: @hsanjuan the reason I want it in thirdparty is explicitly because it needs to be moved to the cid package as soon as possible. But (as you know) we have a bunch of stuff in-flight that prevents us from bubbling up the cid package without other complications, so its going here for now as a "please remove me"

hsanjuan: @whyrusleeping  sure, I understand the reasons, but it still introduces a blocker for me.

I'm trying to communicate that if it's placed under `blockservice` with a "please-remove-me" note it doesn't become a blocker while solving your problem at the same time.

If you say this will be removed from `thirdparty` right away after the release, OK. I'm just scared that things will take longer though.

core/commands/pin.go
Stebalien: (this is one of the many reasons I've given up on fixing issues with the compat layer and have started just converting commands to use the new system).

Stebalien:
Failure:

```
  
cp: cannot stat '../t0275-cid-security-data/AFKSEBCGPUJZE.data': No such file or directory
not ok 23 - injecting insecure block
#	
#	    mkdir -p "$IPFS_PATH/blocks/JZ" &&
#	    cp -f ../t0275-cid-security-data/AFKSEBCGPUJZE.data "$IPFS_PATH/blocks/JZ"
#	  
```

Kubuxu:
missed file from commit, fixing (I didn't notice the CI failures for some reason).

Kubuxu:
Fixed

whyrusleeping:
Alright, here goes.

Issue #4749, opened 2018-03-01
IPNS republisher should not rely on DHT to store (possibly expired) records


dirkmc:
The republisher relies on the DHT to store records, but those records may expire before republish occurs (eg if the republishing node goes offline). We should store them somewhere else in the datastore.
See https://github.com/ipfs/go-ipfs/pull/4742#issuecomment-369604767 for more details

dirkmc:
@whyrusleeping do you have a suggestion for how we should store these records?

Stebalien:
Proposal: make the republisher *wrap* the namesys service in a republishing namesys service. When calling `Publish` on the `RepublishingNamesys`, it would store the record in a datastore and then continue to republish that record until replaced or deleted.

Thoughts?

whyrusleeping:
That sounds like a reasonable proposal to me.

dirkmc:
Should we merge in the PR to use variadic options for the namesys interface first?
 https://github.com/ipfs/go-ipfs/pull/4733

Stebalien:
So, as I noted on that PR, the underlying namesys relies on this record itself to keep track of the sequence number.

I can think of three solutions:

1. Have the underlying namesys record the sequence number separately along with the last published value. IMO, we should do that regardless. Additionally, have the republishing nameservice record which names it should republish (and their last publish times?).
2. Have the namesys *expose* known, published records. That would allow us to keep the republisher simple. However, that also complicates the namesys Publisher interface (which, arguably, shouldn't expose this information).
3. Just combine the two.

hsanjuan:
Quick note: you should be able to update to latest-gx-tagged `go-libp2p-kad-dht`, `go-ipfs-routing`, `go-libp2p-record` for this. I should have left everything working on that front.

dirkmc:
I noticed that the Pubsub publisher [puts an IPNS record](https://github.com/ipfs/go-ipfs/blob/master/namesys/pubsub.go#L151) directly into the datastore, so that the republisher can later retrieve the record's sequence number. But it looks like namesys will [always publish](https://github.com/ipfs/go-ipfs/blob/master/namesys/namesys.go#L170) to both the DHT and to Pubsub anyway, so is it necessary for Pubsub publisher to do this?

dirkmc:
And I guess a related question: Should the republisher care about records published directly to Pubsub (but not to the DHT)?

vyzo:
[...]

PR #4753, opened 2018-03-01
Move namesys republisher records into their own place in the datastore

dirkmc:
See https://github.com/ipfs/go-ipfs/issues/4749

Note: This is branched off dirkmc:feat/namesys-value-count (PR https://github.com/ipfs/go-ipfs/pull/4733)


namesys/repubns.go
```diff
@@ -0,0 +1,62 @@
...
+		return err
+	}
+
+	id, err := peer.IDFromPrivateKey(name)
+	if err != nil {
+		return err
+	}
+
+	_, ipnskey := IpnsKeysForID(id)
+	rec, err := ns.dstore.Get(dshelp.NewKeyFromBinary([]byte(ipnskey)))
```
Stebalien: Making assumptions about how the `NameSystem` stores this information is really unfortunate. Having to store everything twice is also really unfortunate.

My original plan was to simply record "we've published record `/ipns/QmId => /ipfs/QmThing`" at time X then call `Publish(name, value)` every X + RepubTime (recording the new time).

However, this brings up a *different* issue: this isn't limited to republishing. We actually need to remember the last sequence number for namesys to work properly. Therefore, while this PR fixes the immediate issue, `NewNameSystem` still returns a broken namesys.

dirkmc: I agree it is unfortunate to make assumptions about how `NameSystem` stores the information. There are instances all over the namesys code of reaching directly into the datastore to get a record. It makes me wonder if the routing interface should have a `GetLocalValue(key)` method that doesn't go out to the DHT or do any validation of records.

I'm not sure I understood your last comment. Are you saying that namesys republishing is broken, because we need to get the last sequence number for a record, but we only look locally instead of going out to the network?

Stebalien: Yes, but, more than that, the underlying namesys implementation itself is broken (as it will forget the last sequence number).

dirkmc: Sorry I'm still not sure I follow... why would it forget the last sequence number? Is the datastore not persistent?

Stebalien: The problem is that the DHT is now deleting expired entries, regardless of whether or not they're *our* expired entry. That means that when namesys tries to get them directly from the datastore, it may not find them.

Although.... arguably, the sequence number doesn't matter for expired records. That is, if the last known record *has* expired, we can safely start over from sequence number 0. Thoughts @whyrusleeping?

In that case, the republisher would simply store a mapping of `RepublisherKey(ipnskey) => last_publish` and namesys wouldn't interact with datastores *at all*. Instead, namesys would call `GetValue` on the `routing.ValueStore` to determine if there exists an unexpired record and use the sequence number from that. That solves all the problems with services digging into eachother's datastores.

dirkmc: What you're suggesting makes sense, and as I understand it that's basically what Publish() does: it [gets the last sequence number](https://github.com/ipfs/go-ipfs/blob/master/namesys/publisher.go#L61) by [looking in the local datastore](https://github.com/ipfs/go-ipfs/blob/master/namesys/publisher.go#L73), and if it doesn't find the record there it [calls GetValue()](https://github.com/ipfs/go-ipfs/blob/master/namesys/publisher.go#L96) on the DHT. If there's no record in the DHT it uses sequence number zero. Maybe republish should do the same thing?

Are you suggesting we remove the check for the record in the local datastore from `Publish()` to avoid having to look directly into the datastore? I assume it's there for performance reasons.

Note also that when there is a GET RPC on the DHT, it will [send the highest sequence number record](https://github.com/libp2p/go-libp2p-kad-dht/blob/master/routing.go#L119) to peers who have an IPNS record with a lower sequence number, so over time the highest sequence number records should gradually replace lower sequence number records throughout the DHT.

What does `last_publish` represent exactly? I'm just wondering about what happens in the case when a node goes down for a couple of days and then comes back up and wants to republish its (now expired) records. Or are we not concerned about that use case?

dirkmc: @Stebalien one issue I see with maintaining the sequence number in `Namesys` rather than within the DHT is that (if I understand correctly) keys can be shared between peers, so if another peer is publishing ipns records with the same key as my peer then my peer will get out of sync on the sequence numbers. One solution is to always call `GetValue()` but that has performance implications, which is presumably why `Publish()` looks directly in the datastore.

Stebalien: > Are you suggesting we remove the check for the record in the local datastore from Publish() to avoid having to look directly into the datastore? I assume it's there for performance reasons.

Yes.

It's probably there so we don't have to go and ask the network. However, for correctness, we should probably be doing that anyways. There's no guarantee (or even reason to believe) that we're the only one publishing the IPNS record. Honestly, it shouldn't be *that* expensive given that we're going to be doing a DHT put to the *exact* same nodes immediately after calling `GetValue`.

> Maybe republish should do the same thing?

I was thinking that republish would just call the normal `Publish` function and let that handle everything. Note: to avoid bumping the sequence number unnecessarily, publish can avoid incrementing it when the value doesn't change. 

> Note also that when there is a GET RPC on the DHT, it will send the highest sequence number record to peers who have an IPNS record with a lower sequence number, so over time the highest sequence number records should gradually replace lower sequence number records throughout the DHT.

Unless I'm mistaken, this should only apply to non-expired IPNS records. When we're dealing with non-expired records, asking the DHT for the record should work just fine.

> What does last_publish represent exactly? I'm just wondering about what happens in the case when a node goes down for a couple of days and then comes back up and wants to republish its (now expired) records. Or are we not concerned about that use case?

We are. Sorry, my comment there was wrong (part of an alternative proposal that I rejected). The republisher does need to record the last *value* published. That is, `RepublisherKey(ipnskey) => /ipfs/QmLastValue`.

Note: I'd *like* to, additionally, store the last time we published every record (or, alternatively, the last time we did a republish). That's where that `last_publish` came from. At the moment, we *always* republish 5 minutes after starting, regardless of when we last republished as we never know how long we've been offline. However, we can do this separately (and probably just store a global "last republish" value).

dirkmc: I'm not sure if the republisher should store the last value published. For example in this scenario:
- Peer A publishes an IPNS record to its own key
- Peer A shares its key with Peer B
- Peer B publishes an IPNS record to Peer A's key (with a higher sequence number)
The value in Peer A's datastore will be updated by Peer B's DHT PUT, however Peer A's republisher's value for key A is now out of date.

Stebalien: Yeah... However, at that point, I'm starting to wonder if we shouldn't just merge the republisher with namesys and avoid having separate systems entirely. That would probably simplify this anyways.

dirkmc: Yes I think that might make things simpler.

I believe there's actually a bug with the current implementation that we should fix as well. It seems like the republisher will periodically [republish the values for _all_ the keys](https://github.com/ipfs/go-ipfs/blob/master/namesys/republisher/repub.go#L103) in its possession, and it [only checks its local datastore](https://github.com/ipfs/go-ipfs/blob/master/namesys/republisher/repub.go#L127) before republish. That could cause a problem in the following scenario:
- Peer A publishes an IPNS record to Peer B's key with sequence number 5. It is retained in Peer A's local DHT datastore.
- Peer B publishes an IPNS record to its own key with sequence number 6.
- Peer A will not receive Peer B's DHT PUT, because Peer A is not one of the closest peers for key B
- When Peer A republishes it checks its local datastore and finds the record with sequence number 5, which it then republishes to key B
- Peer B _will_ receive Peer A's DHT PUT, overwriting the record with sequence number 6 in its local datastore (and in all the closest peers to key B)

I think the solution here again is for the republisher to always go out to the DHT with `GetValue()` before publishing

Stebalien: > Peer B will receive Peer A's DHT PUT, overwriting the new record (sequence number 6) with the old record (sequence number 5) in its local datastore (and in all the closest peers to key B)

Peer B should ignore this put as the sequence number is lower (unless peer B's value has expired).

But I agree, we should always call GetValue first.

dirkmc: It looks like the handler will [check that the record is valid](https://github.com/libp2p/go-libp2p-kad-dht/blob/master/handlers.go#L175) (ie, signature is ok and record is not expired) but it will not compare the sequence number of the new record with its record before overwriting its local copy

Stebalien: That's... horribly wrong. https://github.com/libp2p/go-libp2p-kad-dht/issues/121

dirkmc:
This really sucks, and should probably be fixed before merging:
https://github.com/ipfs/go-ipfs/pull/4753/files#diff-43f57c7bd628b18d615907b8a8662cd5R22

dirkmc:
@vyzo are you referring to pubsub publish or DHT publish?

vyzo:
where, either one (or the republisher) that needs to get the latest seqno.

dirkmc:
The issue is that in the case where Peer B publishes to Peer A's key, Peer A needs to go out to the DHT to get the latest sequence number

dirkmc:
Note also that I commented out the following failing test:
https://github.com/ipfs/go-ipfs/pull/4753/files#diff-60a3006d82cdbc2ff83462dbd635116eR143

The scenario is:
- Peer A has regular namesys
- Peer B has namesys with pubsub enabled
- Peer A publishes value `foo` for key A to the DHT (but not to pubsub)
- Peer B correctly resolves key A to `foo`, because the pubsub resolver fails and then falls back to DHT resolver
- Peer B publishes value `bar` for key A to pubsub and the DHT
- Peer B correctly resolves key A to `bar`, because it had subscribed to key A
- Peer A publishes value `baz` for key A to the DHT (but not to pubsub)
- Peer B still resolves key A to `bar` (instead of the new value `baz`), because it is listening for key A but `baz` was not published to pubsub (only to the DHT)

I just want to make confirm that it is ok for this scenario to fail?

Stebalien:
So, after talking about this with @whyrusleeping, we've decided on the following:

1. Store the last published record in namesys.
2. Provide a way to get a raw record in namesys.
3. Ignore the multiple publisher case for now.

However, doing this without it sucking will be tricky, especially given that the resolver is technically generalized to handle more than namesys but, in practice, contains a lot of hard-coded logic.

vyzo:
@Stebalien while at it, we should make sure the dht and pubsub publishers use the same record, so that we don't have diverging seqno's between the two channels.

Stebalien:
> while at it, we should make sure the dht and pubsub publishers use the same record, so that we don't have diverging seqno's between the two channels.

I've looked into that and came to the same conclusion I came to before: we're plugging IPNS over pubsub in at the wrong level. Unfortunately, fixing this will be a pain.

vyzo:
> I've looked into that and came to the same conclusion I came to before: we're plugging IPNS over pubsub in at the wrong level. Unfortunately, fixing this will be a pain.

Care to elaborate and offer any constructive suggestions on how to fix it?
If we are doing it wrong, then we should fix it, even if there is pain involved.

vyzo:
> We have two full IPNS publishers (dht + pubsub) and are generating IPNS records in each. We should have a single IPNS publisher that uses both the DHT and Pubsub. 

Agreed -- that's what I was suggesting by using a single record for both channels.

> My original suggestion was to make pubsub a valuestore.

How would that work?

dirkmc:
Closing this as it's out of date

PR #4755, opened 2018-03-02
add: hash-only: set the prefix for MFS root

### LINKED ISSUE #4652: [add] Only-hash gives wrong result with wrap-with-directory
#### Version information:
go-ipfs version: 0.4.14-dev-bd32e73db
Repo version: 6
System version: amd64/windows
Golang version: go1.10rc1

#### Type:
Bug

#### Description:
Similar to https://github.com/ipfs/go-ipfs/issues/4065 except with wrap
`add -n -w` isn't consistent with the real outcome of `add -w` in some cases, here's the ones I noticed.
```
>ipfs add -n -r -w --pin=false --cid-version=1 "test data"
added zdj7Wmd9KWyFAMRcGcpEUdACURqNyYeQ941xecjWjEuMeVAxP test data/File 1
added zdj7WiWydjWvL84aCLtYsSXgrSgKQybkENQk1wXqgJBRReHqG test data/File 2
added zdj7WWuLweFipTVRwrq51wPksGQ6DSvHr32g7HQTMPxbv8m5r test data/File 3
added zdj7WYgq9R8Nb5ytBTgqsJNxY8diJw71icnoDdJLLVEwfdGeo test data
added QmcXqVU8WkYYLNHqNnkXvRcmuuPVMaJFJGjEv5GBD3t1LR

>ipfs add -r -w --pin=false --cid-version=1 "test data"
added zdj7Wmd9KWyFAMRcGcpEUdACURqNyYeQ941xecjWjEuMeVAxP test data/File 1
added zdj7WiWydjWvL84aCLtYsSXgrSgKQybkENQk1wXqgJBRReHqG test data/File 2
added zdj7WWuLweFipTVRwrq51wPksGQ6DSvHr32g7HQTMPxbv8m5r test data/File 3
added zdj7WYgq9R8Nb5ytBTgqsJNxY8diJw71icnoDdJLLVEwfdGeo test data
added zdj7Wjd5ruCNE29D5bm7Xi9noTLY7aw57iPAb8maser2VSuWM

>ipfs add -n -r -w --pin=false --hash=Blake2b-256 "test data"
added zDMZof1m5EigTKbJ4rSfhCbP91N4DFV857ZEZGyKQfG7zx2BFEYK test data/File 1
added zDMZof1ktLby57D5zHpR9ojy1au1qdXFRBWZsFLaMuoobbSfR7yY test data/File 2
added zDMZof1kpbehXSgoEr6A58RXnYCTSrZSHnSPj2uSgrhSYeTx8rLg test data/File 3
added zDMZof1m1wSMZakAeHE4zbgQgJ3r8NkLixv5gcuoqfhPsF87tq3f test data
added QmeN2nhSSDfPB39KmNmwCr6FVgfsoQo3CTv5G25qWMdPyZ

>ipfs add -r -w --pin=false --hash=Blake2b-256 "test data"
added zDMZof1m5EigTKbJ4rSfhCbP91N4DFV857ZEZGyKQfG7zx2BFEYK test data/File 1
added zDMZof1ktLby57D5zHpR9ojy1au1qdXFRBWZsFLaMuoobbSfR7yY test data/File 2
added zDMZof1kpbehXSgoEr6A58RXnYCTSrZSHnSPj2uSgrhSYeTx8rLg test data/File 3
added zDMZof1m1wSMZakAeHE4zbgQgJ3r8NkLixv5gcuoqfhPsF87tq3f test data
added zDMZof1kzedajkUuEZK57yyfDoDAJXBkhprX6bEEMD4AgwhQhvcm
```

schomatis:
In the case that the `only-hash` option (`-n`) is set for `ipfs add`, use the same prefix from the `fileAdder` for the MFS root so it is processed using the same CID version and hash function.

Fixes #4652.

This is a work in progress, some tests should be added. Also, the comment is not very illustrative, it should clarify why is it that the prefix needs two be specified twice (for the file adder and the MFS root).

### DIFF
--- a/core/commands/add.go
+++ b/core/commands/add.go
@@ -283,7 +283,10 @@ You can now check what blocks have been created by:
 
 		if hash {
 			md := dagtest.Mock()
-			mr, err := mfs.NewRoot(req.Context, md, ft.EmptyDirNode(), nil)
+			emptyDirNode := ft.EmptyDirNode()
+			// Use the same prefix for the "empty" MFS root as for the file adder.
+			emptyDirNode.Prefix = *fileAdder.Prefix
+			mr, err := mfs.NewRoot(req.Context, md, emptyDirNode, nil)
 			if err != nil {
 				res.SetError(err, cmdkit.ErrNormal)
 				return
--- a/test/sharness/t0043-add-w.sh
+++ b/test/sharness/t0043-add-w.sh
@@ -149,6 +149,15 @@ test_add_w() {
     echo "$add_w_d1_v1" >expected &&
     test_sort_cmp expected actual
   '
+
+  test_expect_success "ipfs add -w -r -n (dir) --cid-version=1 succeeds" '
+    ipfs add -r -w -n --cid-version=1 m/t_1wp-8a2/_jo7 >actual
+  '
+
+  test_expect_success "ipfs add -w -r -n (dir) --cid-version=1 is correct" '
+    echo "$add_w_d1_v1" > expected &&
+    test_sort_cmp expected actual
+  '
 }
 
 test_init_ipfs


djdv:
This works on my node. 👍 

A simple test case you can add for this one, is to capture the hashes generated by any `add` parameter set and compare them with the same set + `-n`, if they do not match the test should fail, otherwise it should pass.

e.g.
The hashes produced by `add -w --cid-version=1 "testdata"` must equal  the output hashes produced by `add -n -w --cid-version=1 "testdata"`.
The hashes produced by `add -w --hash=Blake2b-256 "testdata"` must equal  the output hashes produced by  `add -n -w --hash=Blake2b-256 "testdata"`.

schomatis:
@kevina I based this [test](https://github.com/ipfs/go-ipfs/pull/4755/files#diff-ca92fc54d8e6f136147f734d968cfb90) on your `ipfs add -w -r (dir)` (adding the `-n` option), as the [issue](https://github.com/ipfs/go-ipfs/issues/4652) was occurring for CIDv1, WDYT?

schomatis:
@Stebalien In this PR the prefix used for the `fileAdder` is also applied to the MFS root, to ensure consistency across the file hashes. Is this correct?

The MFS root is the only node that isn't added by the `fileAdder` (because it's already there before the addition of the user files), does that make sense? I want to add an explanation (in the form of a comment) as to why the prefix has to be referenced twice.

Kubuxu:
@schomatis Is it still WIP?

whyrusleeping:
Another one, thanks @schomatis!

PR #4758, opened 2018-03-02
[WIP] MFS improvements

### LINKED ISSUE #4514: MFS: Race when reading node and taking read/write lock

### LINKED ISSUE #4519: Truncating an MFS file 175 times makes it unwritable.
After exactly 175 truncate/writes, mfs returns an empty string from read (forever, apparently).

schomatis:
*This is a work in progress, not ready to be merged.*

This PR is a (rebased) copy of #4517, as [discussed](https://github.com/ipfs/go-ipfs/issues/4519#issuecomment-369711759), the idea is to add the fix for #4519 and get it ready to merge.

* Improve flags and mode. Make file open options an extensible struct. Unix uses an int because it's easier to pass in a syscall and it has a fixed size. We don't care about that.
* Avoid syncing/flushing when we don't need to.
* Fixes #4519: `dagTruncate` now preserves the `Type` of the node being truncated.

### DIFF
--- a/core/commands/files.go
+++ b/core/commands/files.go
@@ -537,7 +537,7 @@ Examples:
 			return
 		}
 
-		rfd, err := fi.Open(mfs.OpenReadOnly, false)
+		rfd, err := fi.Open(mfs.Flags{Read: true, Sync: false})
 		if err != nil {
 			res.SetError(err, cmdkit.ErrNormal)
 			return
@@ -738,7 +738,7 @@ stat' on the file or any of its ancestors.
 			fi.RawLeaves = rawLeaves
 		}
 
-		wfd, err := fi.Open(mfs.OpenWriteOnly, flush)
+		wfd, err := fi.Open(mfs.Flags{Write: true, Sync: flush})
 		if err != nil {
 			res.SetError(err, cmdkit.ErrNormal)
 			return
--- a/fuse/ipns/ipns_unix.go
+++ b/fuse/ipns/ipns_unix.go
@@ -440,19 +440,20 @@ func (dir *Directory) Mkdir(ctx context.Context, req *fuse.MkdirRequest) (fs.Nod
 }
 
 func (fi *FileNode) Open(ctx context.Context, req *fuse.OpenRequest, resp *fuse.OpenResponse) (fs.Handle, error) {
-	var mfsflag int
+	mfsflags := mfs.Flags{Sync: true}
 	switch {
 	case req.Flags.IsReadOnly():
-		mfsflag = mfs.OpenReadOnly
+		mfsflags.Read = true
 	case req.Flags.IsWriteOnly():
-		mfsflag = mfs.OpenWriteOnly
+		mfsflags.Write = true
 	case req.Flags.IsReadWrite():
-		mfsflag = mfs.OpenReadWrite
+		mfsflags.Write = true
+		mfsflags.Read = true
 	default:
 		return nil, errors.New("unsupported flag type")
 	}
 
-	fd, err := fi.fi.Open(mfsflag, true)
+	fd, err := fi.fi.Open(mfsflags)
 	if err != nil {
 		return nil, err
 	}
@@ -508,19 +509,20 @@ func (dir *Directory) Create(ctx context.Context, req *fuse.CreateRequest, resp 
 
 	nodechild := &FileNode{fi: fi}
 
-	var openflag int
+	openflags := mfs.Flags{Sync: true}
 	switch {
 	case req.Flags.IsReadOnly():
-		openflag = mfs.OpenReadOnly
+		openflags.Read = true
 	case req.Flags.IsWriteOnly():
-		openflag = mfs.OpenWriteOnly
+		openflags.Write = true
 	case req.Flags.IsReadWrite():
-		openflag = mfs.OpenReadWrite
+		openflags.Read = true
+		openflags.Write = true
 	default:
 		return nil, nil, errors.New("unsupported open mode")
 	}
 
-	fd, err := fi.Open(openflag, true)
+	fd, err := fi.Open(openflags)
 	if err != nil {
 		return nil, nil, err
 	}
--- a/mfs/fd.go
+++ b/mfs/fd.go
@@ -1,12 +1,22 @@
 package mfs
 
 import (
+	"context"
 	"fmt"
 	"io"
 
 	mod "github.com/ipfs/go-ipfs/unixfs/mod"
 
-	context "context"
+	ipld "gx/ipfs/Qme5bWv7wtjUNGsK2BNGVUFPKiuxWrsqrtvYwCLRw8YFES/go-ipld-format"
+)
+
+type state uint8
+
+const (
+	stateFlushed state = iota
+	stateSynced
+	stateDirty
+	stateClosed
 )
 
 type FileDescriptor interface {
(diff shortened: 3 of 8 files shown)


schomatis:
@Stebalien I added the fix for #4519 and now the tests are passing, now I have to resolve some merge conficts.

Regarding your commit [`WIP add test case.`](https://github.com/ipfs/go-ipfs/pull/4758/commits/f4dc9a41bc5eb3ecae77554c4a75b50865fc57ec), is that a work in progress (something more should be added to the tests)? Or is it OK as it is (now that the test are passing)? If the second, what should be its final commit message?

Stebalien:
:man_dancing: Nice!

> Regarding your commit WIP add test case., is that a work in progress (something more should be added to the tests)? Or is it OK as it is (now that the test are passing)? If the second, what should be its final commit message?

That was the test case that failed due to #4519. If possible, please move your commit before it and amend that commit message to remove the WIP and the "it fails" part.

schomatis:
Attempting to rebase this PR I found several merge conflicts, and even after resolving those there are still some modifications that need to be made to be able to build the program.

So if I rebase this I still need to add an extra commit (or sneak in several extra modifications in the last resolution of the rebase), and the commits of this branch will be unified, meaning that a roll back to a commit of the middle of this series won't be possible (it won't build). I think it will be cleaner to do a merge here, fixing everything in one extra commit, preserving your original commits and making clear that this series of commits belong to a different time frame of the code base. Another option could be to cherry pick one commit at a time into master and tidy them up so they all work.

@Stebalien Could you guide me through this, I'm not a git expert, WDYT?

Stebalien:
> Another option could be to cherry pick one commit at a time into master and tidy them up so they all work.

This is usually how I do it. I can take a stab at it if you'd like.

schomatis:
@Stebalien Thanks, I'll give it a try myself and let you know if I run into trouble.

schomatis:
The PR was automatically closed when the branch got emptied.

schomatis:
@Stebalien Regarding the first [commit][COMMIT] of the series (*ensure we don't flush unnecessarily*), while running the full tests on it (before I was just checking the problematic `TestTruncateAndWrite` and `TestConcurrentWrites`), I've noticed the `TestFlushing` [hangs][HANG] indefinitely.

This hang happens when flushing a file with the `FlushPath` function, particularly in the function [`WaitPub`][WAITPUB_HANG], because the `Republisher` never receives the `Publish` signal so it doesn't probe [`pubnowch`][PUBNOWCH] (that `WaitPub` is waiting to write on).

The `Update` function that signals the [`Publish`][PUBLISH] channel is not called because (correctly) the modified function [`flushUp`][FLUSHUP] doesn't call `closeChild` (that will eventually trigger the `Update`), the [test file][TEST_FILE] being flushed wasn't modified, so according to the new logic there is nothing to do.

The problem is that `WaitPub` finds an [inconsistency][INCONSISTENCY], a CID that hasn't been published, which is actually the only CID which was set ([`setVal`][SETVAL]) during the test (because up to now there wasn't any modification of content), the [MFS root][MFS_ROOT].

In my (very limited) understanding the issue isn't the modification to `flushUp` (it makes sense to avoid a flush when there isn't anything to flush), nor the hang in `WaitPub` (although it is a bit suspicious that the flush operation has to wait for this publish, I'm not familiar with the publishing system) but the fact that the MFS root is marked to be published but it isn't until another (different) file (node) is flushed, handling it in an indirect fashion.

[COMMIT]: https://github.com/ipfs/go-ipfs/pull/4517/commits/41c554bf88c74b56690c8122630fc0662061f3ed

[WAITPUB_HANG]: https://github.com/ipfs/go-ipfs/blob/41c554bf88c74b56690c8122630fc0662061f3ed/mfs/system.go#L201

[PUBNOWCH]: https://github.com/ipfs/go-ipfs/blob/41c554bf88c74b56690c8122630fc0662061f3ed/mfs/system.go#L226-L241

[PUBLISH]: https://github.com/ipfs/go-ipfs/blob/41c554bf88c74b56690c8122630fc0662061f3ed/mfs/system.go#L217

[FLUSHUP]: https://github.com/ipfs/go-ipfs/blob/41c554bf88c74b56690c8122630fc0662061f3ed/mfs/fd.go#L131

[TEST_FILE]: https://github.com/ipfs/go-ipfs/blob/41c554bf88c74b56690c8122630fc0662061f3ed/mfs/mfs_test.go#L819

[INCONSISTENCY]: https://github.com/ipfs/go-ipfs/blob/41c554bf88c74b56690c8122630fc0662061f3ed/mfs/system.go#L194

[MFS_ROOT]: https://github.com/ipfs/go-ipfs/blob/41c554bf88c74b56690c8122630fc0662061f3ed/mfs/system.go#L75

[HANG]: https://ci.ipfs.team/blue/organizations/jenkins/go-ipfs/detail/PR-4758/8/pipeline#step-65-log-177

[SETVAL]: https://github.com/ipfs/go-ipfs/blob/41c554bf88c74b56690c8122630fc0662061f3ed/mfs/system.go#L186

Stebalien:
> although it is a bit suspicious that the flush operation has to wait for this publish, I'm not familiar with the publishing system

That's probably because flush is called by the user and, after flushing, the user expects updates to be published.

> the fact that the MFS root is marked to be published but it isn't until another (different) file (node) is flushed, handling it in an indirect fashion.

I'm pretty sure your fix is correct. We should either not be calling `setVal` or should be setting `lastpub` as well. Calls to `setVal` should always be paired with setting the Publish flag.

Stebalien:
@schomatis I've fixed it by adding a `Set` method for setting the both the current value and the last published value. IMO, this makes the most sense.

I've also:

* Tried to make the republish timer logic faster and easier to follow (hopefully).
* Made WaitPub not racy. Instead of checking to see if there is anything to publish, it just writes to the pubnowch channel and I deduplicate publish operations inside the publish loop.

@whyrusleeping significant change: We no longer publish one final time on close if nothing has been changed. IMO, this is correct but it's still a change (and forced me to change a test).

Stebalien:
This appears to cause a bug in the MFS sharness tests. It looks like there's a case where we *should* be flushing but aren't.

schomatis:
@Stebalien When [opening](https://github.com/schomatis/go-ipfs/blob/04523cc1414db64b3279830ea60a238614527747/mfs/file.go#L84) a file its `state` is not initialized and hence left at `stateFlushed` (I think this should be an explicit initialization even if the value is left at `iota`). The `t0251-files-flushing.sh` is failing because `ipfs files cp` flushes a copied (created but not modified) file set to `stateFlushed` which does nothing. Maybe the default `state` value of a file should be revised, `stateDirty` seems like the safest choice but I'm not sure that's correct either.

I don't know what's causing the errors in `t0250-files-api.sh`.

Stebalien:
Hm. I'm not sure I understand. `ipfs files cp` doesn't appear to call `Open` on the target file, as far as I can tell.

It looks like `ipfs files cp` is calling `PutNode` which adds the file with `directory.AddChild`. That, in turn, appears to add the child to the directory by calling `dirbuilder.AddChild` (no open, flush, etc). What am I missing here?

schomatis:
@Stebalien The `Open` happens during the [`FlushPath`](https://github.com/ipfs/go-ipfs/blob/master/core/commands/files.go#L343) call (there is an undocumented `flush` option set to `true`) which ends up in a file [`Flush`](https://github.com/schomatis/go-ipfs/blob/04523cc1414db64b3279830ea60a238614527747/mfs/file.go#L116).

```
0  0x000000000120e78b in github.com/ipfs/go-ipfs/mfs.(*File).Open
   at ./mfs/file.go:51
1  0x000000000120f53d in github.com/ipfs/go-ipfs/mfs.(*File).Flush
   at ./mfs/file.go:126
2  0x0000000001211f52 in github.com/ipfs/go-ipfs/mfs.FlushPath
   at ./mfs/ops.go:218
3  0x0000000001542891 in github.com/ipfs/go-ipfs/core/commands.glob..func57
   at ./core/commands/files.go:343
4  0x0000000001372cfd in github.com/ipfs/go-ipfs/commands/legacy.NewCommand.func1
   at ./commands/legacy/command.go:38
```

Stebalien:
@schomatis do you recall what's blocking this? It has a bunch of bug fixes and performance improvements.

schomatis:
The flush error discussed above.

schomatis:
This code has been moved to [`go-mfs`](https://github.com/ipfs/go-mfs), what can be extracted from this PR will be continued in https://github.com/ipfs/go-mfs/issues/32.

PR #4764, opened 2018-03-04
migrate repo verify to new cmds library.

Stebalien:
License: MIT

### DIFF
--- a/core/commands/repo.go
+++ b/core/commands/repo.go
@@ -6,7 +6,6 @@ import (
 	"io"
 	"os"
 	"path/filepath"
-	"strings"
 	"text/tabwriter"
 
 	oldcmds "github.com/ipfs/go-ipfs/commands"
@@ -40,7 +39,7 @@ var RepoCmd = &cmds.Command{
 		"gc":      lgc.NewCommand(repoGcCmd),
 		"fsck":    lgc.NewCommand(RepoFsckCmd),
 		"version": lgc.NewCommand(repoVersionCmd),
-		"verify":  lgc.NewCommand(repoVerifyCmd),
+		"verify":  repoVerifyCmd,
 	},
 }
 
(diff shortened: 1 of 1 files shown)


core/commands/repo.go
```diff
@@ -298,63 +291,35 @@ var repoVerifyCmd = &oldcmds.Command{
...
-			}
-
-			obj, ok := v.(*VerifyProgress)
-			if !ok {
-				return nil, e.TypeErr(obj, v)
-			}
-
-			buf := new(bytes.Buffer)
-			if strings.Contains(obj.Msg, "was corrupt") {
-				fmt.Fprintln(os.Stdout, obj.Msg)
```
Stebalien: I removed this distinction. Was it really necessary? (also, I assume this was *supposed* to print to stderr...)

whyrusleeping:
Do you intend for this to come before #4751?

Stebalien:
I don't see how they're related. Does that PR touch repo verify? I'm trying to bypass an issue I've noticed in the cmds lib backwards compatibility layer by simply migrating away from it as fast as possible (the `expected <-chan interface{}, got <nil>` on `^C` error).

core/commands/repo.go
Kubuxu: It was probably to be able to pipe just the corrupted blocks out into other command.

core/commands/repo.go
```diff
@@ -298,63 +291,35 @@ var repoVerifyCmd = &oldcmds.Command{
...
-				return buf, nil
-			}
-
+	Encoders: cmds.EncoderMap{
+		cmds.Text: cmds.MakeTypedEncoder(func(req *cmds.Request, w io.Writer, obj *VerifyProgress) error {
```
Stebalien: Wait for https://github.com/ipfs/go-ipfs-cmds/issues/65 to be fixed before merging.

core/commands/repo.go
Stebalien: I'd guess this was mean to write to stderr as the normal output is written to stdout.

core/commands/repo.go
dignifiedquire: @stebalien the linked issue has since been fixed, was this the only blocker for moving this forward?

Stebalien:
This was actually replaced by a newer PR (one rebased over a PR that parallelized the verification). Thanks for bumping this.

PR #4767, opened 2018-03-05
dag: diff: check CIDs in base case when comparing nodes

### LINKED ISSUE #4591: ipfs object diff returns a change when diffing empty directory with itself
#### Version information:
go-ipfs version: 0.4.13-
Repo version: 6
System version: amd64/darwin
Golang version: go1.9.2

#### Type: 
Bug

#### Severity:
Low

#### Description:
To recreate:
`ipfs files mkdir /testdir`
`HASH=$(ipfs files stat --hash /testdir)`
`ipfs object diff $HASH $HASH`
output is:

> ~ QmUNLLsPACCz1vLxQVkXqqLX5R1X345qqfHbsf67hvA3Nn QmUNLLsPACCz1vLxQVkXqqLX5R1X345qqfHbsf67hvA3Nn ""

I would expect it to be empty

schomatis:
The base case was assuming that the nodes were different (see [issue comment](https://github.com/ipfs/go-ipfs/issues/4591#issuecomment-368596386)).

Fixes #4591.

### DIFF
--- a/merkledag/utils/diff.go
+++ b/merkledag/utils/diff.go
@@ -95,7 +95,12 @@ func ApplyChange(ctx context.Context, ds ipld.DAGService, nd *dag.ProtoNode, cs 
 
 // Diff returns a set of changes that transform node 'a' into node 'b'
 func Diff(ctx context.Context, ds ipld.DAGService, a, b ipld.Node) ([]*Change, error) {
+	// Base case where both nodes are leaves, just compare
+	// their CIDs.
 	if len(a.Links()) == 0 && len(b.Links()) == 0 {
+		if a.Cid().Equals(b.Cid()) {
+			return []*Change{}, nil
+		}
 		return []*Change{
 			&Change{
 				Type:   Mod,
--- a/test/sharness/t0052-object-diff.sh
+++ b/test/sharness/t0052-object-diff.sh
@@ -20,7 +20,11 @@ test_expect_success "create some objects for testing diffs" '
   echo "nested" > foo/baz/dog &&
   C=$(ipfs add -r -q foo | tail -n1)
   echo "changed" > foo/bar &&
-  D=$(ipfs add -r -q foo | tail -n1)
+  D=$(ipfs add -r -q foo | tail -n1) &&
+  echo "" > single_file &&
+  SINGLE_FILE=$(ipfs add -r -q single_file | tail -n1) &&
+  mkdir empty_dir
+  EMPTY_DIR=$(ipfs add -r -q empty_dir | tail -n1)
 '
 
 test_expect_success "diff against self is empty" '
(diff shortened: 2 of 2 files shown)


schomatis:
@whyrusleeping As you implemented `ipfs object diff` could you review this please?

schomatis:
> I think that we should check if the data has changed - this is still going to be sort-of broken if there are links in the node and the data is changed.

@magik6k Could you elaborate more on this please? I'm not sure I follow, isn't the CID reflecting the data changes?

magik6k:
What I mean is that in case where the data changes in top-level node and we have some links, diff won't show any changes:

```
# Add some data
$ printf '{"Data":"foo"}' | ipfs object put
added QmQeGyS87nyijii7kFt1zbe4n2PsXTFimzsdxyE9qh9TST
$ printf '{"Data":"bar"}' | ipfs object put
added QmNeYRbCibmaMMK6Du6ChfServcLqFvLJF76PzzF76SPrZ

# Diff data without links, this works fine
$ ipfs object diff QmQeGyS87nyijii7kFt1zbe4n2PsXTFimzsdxyE9qh9TST QmNeYRbCibmaMMK6Du6ChfServcLqFvLJF76PzzF76SPrZ
~ QmQeGyS87nyijii7kFt1zbe4n2PsXTFimzsdxyE9qh9TST QmNeYRbCibmaMMK6Du6ChfServcLqFvLJF76PzzF76SPrZ ""

# Add more data, this time with links:
$ printf '{"Data":"abcd", "Links":[{"Hash":"QmQeGyS87nyijii7kFt1zbe4n2PsXTFimzsdxyE9qh9TST"}]}' | ipfs object put
added QmWB7VRDLSweiURLoKMtan2ZCneBXmYAHpsrbmvC8tEZav
$ printf '{"Data":"dcba", "Links":[{"Hash":"QmQeGyS87nyijii7kFt1zbe4n2PsXTFimzsdxyE9qh9TST"}]}' | ipfs object put
added QmdjxDJgR1hvXNSfuwWniB3DsBgv4C1SSYu6RAxwbd38H2

# Diff
$ ipfs object diff QmWB7VRDLSweiURLoKMtan2ZCneBXmYAHpsrbmvC8tEZav QmdjxDJgR1hvXNSfuwWniB3DsBgv4C1SSYu6RAxwbd38H2

# No output, we should get one like in the case without links

```

Note that this is also the current behaviour, not caused by the changes here, so sorry for the confusion.

schomatis:
@magik6k Thanks for the clarification, I was under the impression that the data was only reserved for the leaf nodes, not that a node could have both data and links. Yes, this was out of the scope of the original issue, but I could open a new one to discuss this further.

whyrusleeping:
LGTM, lets merge after 0.4.14 ships. 

Thanks for all the PRs @schomatis :) Keep 'em coming!

whyrusleeping:
Three in a row!

PR #4768, opened 2018-03-05
commands/dht: add missing ctx returns

magik6k:


magik6k:
https://github.com/ipfs/go-ipfs/pull/4769 fixes this and then some more.

PR #4769, opened 2018-03-05
convert the dht commands to the new commands lib

Stebalien:
Also, stop accepting variadic arguments in single argument commands.

Also, -150loc :)

License: MIT

### DIFF
--- a/core/commands/dht.go
+++ b/core/commands/dht.go
@@ -1,15 +1,12 @@
 package commands
 
 import (
-	"bytes"
 	"context"
 	"errors"
 	"fmt"
 	"io"
 	"time"
 
-	cmds "github.com/ipfs/go-ipfs/commands"
-	e "github.com/ipfs/go-ipfs/core/commands/e"
 	dag "github.com/ipfs/go-ipfs/merkledag"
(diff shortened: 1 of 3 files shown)


Stebalien:
@magik6k this should fix the bug you told me about on IRC.

core/commands/dht.go
```diff
@@ -145,14 +113,14 @@ var findProvidersDhtCmd = &cmds.Command{
 	},
 
 	Arguments: []cmdkit.Argument{
-		cmdkit.StringArg("key", true, true, "The key to find providers for."),
+		cmdkit.StringArg("key", true, false, "The key to find providers for."),
```
Kubuxu: Why disable variadic here?

Kubuxu: On the other hand, it makes sense as it never worked. Ok.

Stebalien:
Unfortunately, we should iron out https://github.com/ipfs/go-ipfs-cmds/issues/65 first.

eingenito:
@Stebalien I'm closing this because I think it's been completed in https://github.com/ipfs/go-ipfs/pull/5671. Please reopen if that's not correct.

Issue #4719, opened 2018-02-19
`pubsub pub` doesn't work with stdin anymore (regression)


VictorBjelkholm:
#### Version information:

```
go-ipfs version: 0.4.14-dev-986d569ac
Repo version: 6
System version: amd64/linux
Golang version: go1.9.4
```

#### Type:

Bug
#### Description:

In version `0.4.13` of go-ipfs, you could pipe input to `ipfs pubsub pub $TOPIC` and it'll be sent. Trying the latest build of `go-ipfs` from master, that no longer work. No input seems to reach the `pub` command.

Guessing this is because of the cmd-lib refactor stuff.

VictorBjelkholm:
In general, it seems pubsub commands doesn't handle piping very well, at least in my limited tests. `pubsub sub` seems to never pipe the output, not sure why it's being buffered. Maybe it's on purpose?

VictorBjelkholm:
Thanks for acknowledging this issue in the all hands @Stebalien 

What I was trying was to get asciinema streaming over pubsub to work with the new version of asciinema. So basically if we can get pubsub commands to play nicely with named unix pipes, streaming to/from asciinema should work fine.

PR #4770, opened 2018-03-05
fix pubsub pub reading stdin args

whyrusleeping:
License: MIT

### DIFF
--- a/test/sharness/t0180-pubsub.sh
+++ b/test/sharness/t0180-pubsub.sh
@@ -35,7 +35,7 @@ test_expect_success 'pubsub' '
 '
 
 test_expect_success "wait until ipfs pubsub sub is ready to do work" '
-  sleep 1
+  go-sleep 500ms
 '
 
 test_expect_success "can see peer subscribed to testTopic" '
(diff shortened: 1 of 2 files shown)


whyrusleeping:
Test failures in bitswap... @Stebalien any idea there?

whyrusleeping:
It looks like we just need a sleep here: https://github.com/ipfs/go-ipfs/blob/master/exchange/bitswap/session_test.go#L315

Cancelling a context is always an async operation, expecting it to affect something else immediately afterwards is a risky bet. In reality, we shouldnt use a context to govern this, we should have an explicit close method, but that will require more refactoring.

test/sharness/t0180-pubsub.sh
```diff
@@ -57,6 +57,36 @@ test_expect_success "wait until echo > wait executed" '
...
+  (
+    ipfsi 2 pubsub sub --enc=ndpayload testTopic | if read line; then
+        echo $line > actual &&
+        echo > wait2
+      fi
+  ) &
+'
+
+test_expect_success "wait until ipfs pubsub sub is ready to do work" '
+  go-sleep 500ms
```
Kubuxu: Should we poll here for something (to be sure that pubsub is ready) or not?

Kubuxu: I see that we do it in other places so LGTM either way.

whyrusleeping: Ideally, but thats pretty difficult to do. For now, this is good enough

whyrusleeping:
@VictorBjelkholm this fixes #4719

PR #4772, opened 2018-03-05
misc: create CODEOWNERS file

Kubuxu:
This will cause me being tagged for review of every PR, unless we have
subsystem owners, then they will be tagged for review.

License: MIT


whyrusleeping:
If anyone else wants to be auto-tagged for review on certain things, please make it known

Kubuxu:
@whyrusleeping seems like no one else is interested. RFM?

whyrusleeping:
Yeah, SGTM.

PR #4783, opened 2018-03-07
redundant word

yenkuanlee:
remove redundant word


whyrusleeping:
unfortunatlely needs that commit signoff

PR #4784, opened 2018-03-07
[WIP] Make the Node methods on ProtoNode thread-safe

### LINKED ISSUE #3991: Make the Node methods on ProtoNode thread-safe
It's reasonable to expect that structures conforming to the `Node` interface are thread safe. However, `ProtoNode` currently lazily encodes and computes its CID without synchronization. We should be using `sync.Once`.

See the discussion here https://github.com/ipfs/go-ipld-cbor/pull/12#issuecomment-309147328

schomatis:
This is a work in progress, not ready to be merged.

Test of thread-safety of the `ProtoNode` implementation of the `Node` interface (#3991). The `Stat()` method does an in-place [sort](https://github.com/ipfs/go-ipfs/blob/b002acc6896a40a915f10ff1d79ce3130e34f6a6/merkledag/coding.go#L80) that can result in an unordered link slice (giving an [unexpected](https://travis-ci.org/ipfs/go-ipfs/jobs/350140379#L859) hash/CID).

Closes #3991.

### DIFF
--- a/merkledag/node.go
+++ b/merkledag/node.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"sync"
 
 	mh "gx/ipfs/QmZyZDi491cCNTLfAhwcaDii2Kg4pwKRkhqQzURGDvY6ua/go-multihash"
 	cid "gx/ipfs/QmcZfnkapfECQGcLZaf9B79NRg7cRa9EnZh4LSbkCzwNvY/go-cid"
@@ -27,6 +28,11 @@ type ProtoNode struct {
 		encodedValue []byte
 		// Cached CID value.
 		cid *cid.Cid
+		// Lock to initialize cache (just once).
+		initialize sync.Once
+		// If the initialization failed the error needs to be saved
+		// because the marshaling is done only once.
+		initializationError error
 	}
 
 	// Prefix specifies cid version and hashing function
--- a/merkledag/node.go
+++ b/merkledag/node.go
@@ -207,12 +207,16 @@ func (n *ProtoNode) Copy() ipld.Node {
 
 	nnode.Prefix = n.Prefix
 
+	// Invalidate the cache for the copy, which may be modified.
+	// (The cache values are initialized to `nil` but still make it explicit.)
+	nnode.invalidateCache()
+
 	return nnode
 }
 
 // RawData returns the protobuf-encoded version of the node.
 func (n *ProtoNode) RawData() []byte {
-	out, _ := n.EncodeProtobuf(false)
+	out, _ := n.EncodeProtobuf()
 	return out
 }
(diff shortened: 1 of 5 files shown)


schomatis:
@Stebalien Could you take a quick look at the last [commit](https://github.com/ipfs/go-ipfs/pull/4784/commits/1a1574573690055ef0be1b18656b4cbbd29abad9)? It's still a WIP but just to check if I'm going in the right direction. The idea is to encapsulate the access to the node's cache, that way adding the `Once` lock should be pretty straight forward.

merkledag/node.go
```diff
@@ -24,8 +24,11 @@ type ProtoNode struct {
 
 	// cache encoded/marshaled value
 	encoded []byte
-
-	cached *cid.Cid
+	// TODO: encoded vs cache, they are both caches, just that they cache
+	// different parts of the system. As cid is the hash of the encoded
+	// they are closely tied together and this code should reflect that.
+	// If encoded is modified the cid is invalid (and should be set to nil).
+	cachedCID *cid.Cid
```
Stebalien: FYI, you can do this by creating an embedded struct:

```go
type ProtoNode struct {
    // ...
    cached struct {
        once sync.Once
        cid *cid.Cid
        bytes []byte
    }
}
```

merkledag/node.go
```diff
@@ -131,7 +133,7 @@ func (n *ProtoNode) AddRawLink(name string, l *ipld.Link) error {
 
 // RemoveNodeLink removes a link on this node by the given name.
 func (n *ProtoNode) RemoveNodeLink(name string) error {
-	n.encoded = nil
+	n.invalidateEncoding()
```
Stebalien: Nit: we're not invalidating the "encoding", just the "encoded" or cached object (I'd call it `invalidateCache`.

schomatis:
@Stebalien Thanks for the previous review, it made much sense to encapsulate all the cache in a single structure.

I'm trying to add the `Once` lock inside `EncodeProtobuf` (see last [commit](https://github.com/ipfs/go-ipfs/pull/4784/commits/716e5b3795d74eb1cdcc98722e0a598e5cbaaac5)) to control the access to the cached data (now that `Marshal` has been made private`EncodeProtobuf` should be the single point of entry).

In case there is an error during the marshaling inside `EncodeProtobuf`, how should that case be handled? Right now I'm caching also the error, following the proposed API that once a node is committed it's frozen and can't be modified any more (the node should be copied to modify the new copy, not the old original). This means that if the node was improperly created and its marshaling fails (e.g., a corrupt node) the node should stay in that failed state, and another copy would be necessary to fix it. WDYT?

schomatis:
Closing as most of the code has been moved elsewhere, there may still be value here to inspire a similar PR in `go-merkledag`.

Issue #4781, opened 2018-03-07
`ipfs get` doesn't work for small files


Stebalien:
```sh
> ipfs get -o myfile "$(echo "foo" | ipfs add -q)" && cat myfile
```

:scream:

Stebalien:
So, apparently this only happens with go-ipfs 1.10. Possibly related to the http library? Maybe we need to flush something?

patmaddox:
I have two machines running OS X 10.12.6 that both work with homebrew ipfs. One reports go1.9.2 and the other reports go1.10. Both show that ipfs was "poured from bottle."

The machine that's broken is a 10.10.5 machine and homebrew says it was built from source.

I'll experiment with a different go version, but maybe the above info helps somewhat...

schomatis:
@Stebalien The new [`regFileReader.Read()`](https://github.com/golang/go/commit/3bece2fa0e10c2dec37b3f57925def40c8191e37#diff-6ce8fb1d02e21c5b3b730dd7876959e9R635) in `go1.10` now returns in a successful read that reaches the end of the file both the number of bytes read and the `io.EOF` error (I'm not sure the intention behind this change in the API, there's no comment in the `switch` nor an explanation in the commit message besides saying that *regFileReader and sparseFileReader were completely rewritten*, so it may be an unintended modification).

When the tar extractor [detects the `EOF` error](https://github.com/ipfs/go-ipfs/blob/b002acc6896a40a915f10ff1d79ce3130e34f6a6/thirdparty/tar/extractor.go#L116-L119) it assumes nothing has been read (the end of the file has been reached) so it doesn't copy anything to the output and returns no error.

Kubuxu:
That is our mistake as EOF always could be returned with some data.

schomatis:
You're right, as the [`Reader`](https://golang.org/pkg/io/#Reader) interface specifies:

> Callers should always process the n > 0 bytes returned before considering the error err. Doing so correctly handles I/O errors that happen after reading some bytes and also both of the allowed EOF behaviors.

PR #4785, opened 2018-03-07
disable building with go1.10 for release

whyrusleeping:
Something in go1.10 breaks ipfs in some way. This prevents people from building the release on their own and complaining when things break. We will remove this as soon as the 0.4.15 cycle starts.

License: MIT

### DIFF
--- a/cmd/ipfs/break_go10.go
+++ b/cmd/ipfs/break_go10.go
@@ -1,0 +1,4 @@
+// +build go1.10
+package main
+
+var _ = IpfsDoesNotSupportBeingBuiltWithGo1dot10


Kubuxu:
Our whole CI uses Go1.10...

Kubuxu:
@VictorBjelkholm unfortunately for some reason the problems are not showing up in our tests.

So for time being building with Go 1.10 won't help much.

Kubuxu:
@whyrusleeping looks like that build tag didin't work

Kubuxu:
Yes but Jenkins is using Go 1.9 on master. Otherwise we would have noticed long time ago as there was one more test failing apart from one I've added in #4790

VictorBjelkholm:
> unfortunately for some reason the problems are not showing up in our tests

As I understand, the issue applies to very basic usage of `get`, which should surely be caught in your tests? (at least as I understand https://github.com/ipfs/go-ipfs/issues/4781). If this issue only happens with 1.10 and not with 1.9, tests would be passing on 1.9 and with this fix, not be runnable at all on 1.10 and with the other fixes (https://github.com/ipfs/go-ipfs/issues/4793) should be passing.

Kubuxu:
@VictorBjelkholm in short terms yes.

Kubuxu:
@VictorBjelkholm main fix is https://github.com/ipfs/go-ipfs/pull/4790 the #4793 is fixing the patter in other places across the codebase but as far as we know nothing else is broken.

The bug arose because of bad usage of `Read` on our side and changes Golang made to tar standard library in Go 1.10.

PR #4789, opened 2018-03-07
commands: publish: check if key matches node ID

### LINKED ISSUE #4779: Key lookup somehow broken when using --key=<peerId> option for `ipfs name publish`
#### Version information:

0.4.14-dev-199a52d

#### Type:

Bug

#### Description:

I somehow broke key lookup:

```
$ ipfs key list -l
Qma5Tggt66MccA7p4qe3cTRkHEN1jTvCg8FotF2DedW7vM self 

$ ipfs name publish --key=Qma5Tggt66MccA7p4qe3cTRkHEN1jTvCg8FotF2DedW7vM QmYYSXJyYsgqMQGHKgFyYAvrmu5YjpNAVVCHXaGRbYLh9E
Error: no key by the given name or PeerID was found

$ ipfs name publish QmYYSXJyYsgqMQGHKgFyYAvrmu5YjpNAVVCHXaGRbYLh9E
Published to Qma5Tggt66MccA7p4qe3cTRkHEN1jTvCg8FotF2DedW7vM: /ipfs/QmYYSXJyYsgqMQGHKgFyYAvrmu5YjpNAVVCHXaGRbYLh9E
```

I don't recall seeing any error messages, but I was messing around with publishing things to IPNS all afternoon and had to cancel a few publish requests that I had made through the HTTP API because they hadn't returned after more than 10 minutes. It was working earlier today.

leavehouse:
Fixes #4779.

License: MIT
Signed-off-by: Nick Hamann <heavelouse@gmail.com>

### DIFF
--- a/core/commands/publish.go
+++ b/core/commands/publish.go
@@ -205,6 +205,10 @@ func keylookup(n *core.IpfsNode, k string) (crypto.PrivKey, error) {
 		return nil, err
 	}
 
+	if k == n.Identity.Pretty() {
+		return n.PrivateKey, nil
+	}
+
 	keys, err := n.Repo.Keystore().List()
 	if err != nil {
 		return nil, err
--- a/test/sharness/t0100-name.sh
+++ b/test/sharness/t0100-name.sh
@@ -90,4 +90,17 @@ test_expect_success "publish an explicit node ID as key name looks good" '
   test_cmp expected_node_id_publish actual_node_id_publish
 '
 
+# publish with the default node ID as key name
+
+test_expect_success "'ipfs name publish --key=<self peer-id> <hash>' succeeds" '
+  PEERID=`ipfs id --format="<id>"` &&
+  test_check_peerid "${PEERID}" &&
+  ipfs name publish --key=${PEERID} "/ipfs/$HASH_WELCOME_DOCS" >actual_self_node_id_publish
+'
+
+test_expect_success "publish the explicit self node ID as key name looks good" '
+  echo "Published to ${PEERID}: /ipfs/$HASH_WELCOME_DOCS" >expected_self_node_id_publish &&
+  test_cmp expected_self_node_id_publish actual_self_node_id_publish
+'
+
 test_done


Kubuxu:
If it isn't too much to ask, mind adding test for it to sharness, it would be in `test/sharness/t0100-name.sh` file and you can get peer id using `PEERID=$(ipfs config Identity.PeerID)`.

leavehouse:
Yep, added.

core/commands/publish.go
```diff
@@ -205,6 +205,10 @@ func keylookup(n *core.IpfsNode, k string) (crypto.PrivKey, error) {
 		return nil, err
 	}
 
+	if k == n.Identity.Pretty() {
```
magik6k: I'd put this in https://github.com/ipfs/go-ipfs/blob/master/core/core.go#L689 (`n.GetKey`). This way it's also fixing `name.Publish` in coreapi - https://github.com/ipfs/go-ipfs/blob/master/core/coreapi/name.go#L142

leavehouse: It's a good point that we should avoid duplicating this check in coreapi, but is it appropriate for `GetKey` to concern itself with checking for something that isn't a key name? The reason this fix is needed seems to be that the `self` key isn't returned from `n.Repo.Keystore().List()`. (Notice the code adds it manually in https://github.com/ipfs/go-ipfs/blob/master/core/commands/keystore.go#L192-L194)

Do you think it'd be better to fix it there instead?

magik6k: Right.. It isn't the best place (it's accepting the key _name_). Putting it in the return of `n.Repo.Keystore().List()` would imply that it is stored in keystore which it is not. The commands here are going to get rewritten to use coreapi eventually, so I think it is fine to duplicate this check to https://github.com/ipfs/go-ipfs/blob/master/core/coreapi/name.go#L142.

leavehouse: Yeah, certainly we want to avoid duplicating this fix in coreapi, but should `GetKey` be concerned with something that isn't a key name?

I notice that every place that calls `Keystore.List()` either manually adds the self key (https://github.com/ipfs/go-ipfs/blob/master/core/commands/keystore.go#L184-L192, https://github.com/ipfs/go-ipfs/blob/master/core/coreapi/key.go#L97-L105), or currently has a bug because it doesn't manually add it. Would it be appropriate to change the `Keystore.List()` (and, I assume, `Keystore.Get()`) to return and recognize the self key?

magik6k: I'm not sure if this is the correct way, though it feels the cleanest to me

@Stebalien @Kubuxu Is there any reason the self key isn't handled by keystore methods?

Kubuxu:
Let's rethink it, I will write stuff down in some time.

overbool:
@Kubuxu Do you have a better idea about this? Maybe calling `Keystore.List()` either manually adds the self key  in here is a compromise.

PR #4791, opened 2018-03-07
commands: resolve: resolve key names

### LINKED ISSUE #3900: ipfs name resolve -k

leavehouse:
I added a sharness test that just tests resolving the "self" key. Let me know if I need to a more extensive one that generates an additional key

Closes #3900

License: MIT
Signed-off-by: Nick Hamann <heavelouse@gmail.com>

### DIFF
--- a/test/sharness/t0100-name.sh
+++ b/test/sharness/t0100-name.sh
@@ -32,6 +32,16 @@ test_expect_success "resolve output looks good" '
   test_cmp expected2 output
 '
 
+# also test resolving the "self" key
+
+test_expect_success "'ipfs name resolve self' succeeds" '
+  ipfs name resolve self >self_output
+'
+
+test_expect_success "self resolve output looks good" '
+  test_cmp expected2 self_output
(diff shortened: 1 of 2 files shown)


test/sharness/t0100-name.sh
```diff
@@ -32,6 +32,16 @@ test_expect_success "resolve output looks good" '
...
+# also test resolving the "self" key
+
+test_expect_success "'ipfs name resolve self' succeeds" '
+  ipfs name resolve self >self_output
+'
+
+test_expect_success "self resolve output looks good" '
+  test_cmp expected2 self_output
+'
+
```
Kubuxu: Could you add tests with generated keys as well? Not only the `self` key?

overbool: @Kubuxu,  It's been 6 months, Can I take over his job?

schomatis: @overbool Yes, go for it!

PR #4792, opened 2018-03-08
fix incorrect Read calls

Stebalien:
follow up on #4790

review commit by commit

Note: I ignored all test cases.

### DIFF
--- a/unixfs/io/pbdagreader.go
+++ b/unixfs/io/pbdagreader.go
@@ -181,6 +181,11 @@ func (dr *PBDagReader) CtxReadFull(ctx context.Context, b []byte) (int, error) {
 			return total, nil
 		}
 
+		// We haven't hit the end yet.
+		if err != io.EOF {
+			continue
+		}
+
 		// Otherwise, load up the next block
 		err = dr.precalcNextBuf(ctx)
 		if err != nil {
--- a/unixfs/io/pbdagreader.go
+++ b/unixfs/io/pbdagreader.go
@@ -166,27 +166,20 @@ func (dr *PBDagReader) CtxReadFull(ctx context.Context, b []byte) (int, error) {
 	total := 0
 	for {
 		// Attempt to fill bytes from cached buffer
-		n, err := dr.buf.Read(b[total:])
+		n, err := io.ReadFull(dr.buf, b[total:])
 		total += n
 		dr.offset += int64(n)
-		if err != nil {
-			// EOF is expected
-			if err != io.EOF {
-				return total, err
-			}
-		}
-
-		// If weve read enough bytes, return
-		if total == len(b) {
+		switch err {
+		// io.EOF will happen is dr.buf had noting more to read (n == 0)
+		case io.EOF, io.ErrUnexpectedEOF:
+			// do nothing
+		case nil:
 			return total, nil
+		default:
+			return total, err
 		}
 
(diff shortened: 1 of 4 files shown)


commands/channelmarshaler.go
```diff
@@ -34,11 +34,9 @@ func (cr *ChannelMarshaler) Read(p []byte) (int, error) {
 	}
 
 	n, err := cr.reader.Read(p)
-	if err != nil && err != io.EOF {
-		return n, err
-	}
-	if n == 0 {
+	if err == io.EOF {
 		cr.reader = nil
+		err = nil
```
Kubuxu: Are you sure we want to `nil` out the error in this case and not propagate the `io.EOF`? This is `Read` method after all.

Stebalien: Had to think about that for a moment... yes. We're *concatenating* output so this is the *continue* case. The next time the user calls Read, we'll hit the `cr.reader == nil` case. If there are no more readers, we'll return the EOF.

Kubuxu: Ahh, right.

unixfs/io/pbdagreader.go
```diff
@@ -181,6 +181,11 @@ func (dr *PBDagReader) CtxReadFull(ctx context.Context, b []byte) (int, error) {
 			return total, nil
 		}
 
+		// We haven't hit the end yet.
+		if err != io.EOF {
+			continue
+		}
+
```
Kubuxu: @Stebalien what do you think about using `io.ReadFull` or `io.ReadAtLeast` here. They will return `io.ErrUnexpectedEOF` if EOF is hit before the buffer is full and thus telling us to load the next buffer.

This would decrease significantly complexity of this function.

For context: https://golang.org/src/io/io.go?s=10738:10804#L293

Stebalien: I considered it but then I got lazy. However, you are right. It'll probably make this simpler.

Kubuxu:
@Stebalien I've updated it to use io.ReadFull.

Your changes to the dagmodifier seem to cause tests to fail.

Kubuxu:
Just confirmed and it is the 1296cda0d2a3e603c299465d83ef11da6190510c

unixfs/io/pbdagreader.go
Stebalien: So, this function actually needs to read *multiple* files so using ReadFull ends up making it more complex.

Stebalien:
> Your changes to the dagmodifier seem to cause tests to fail.

Fixed.

Kubuxu:
I will squash it to correct commit not to have broken commit in middle of tree.

Stebalien:
> I will squash it to correct commit not to have broken commit in middle of tree.

Fair enough. I just didn't want to hide the change.

commands/channelmarshaler.go
magik6k: Doesn't this create the case @kevina described in https://github.com/ipfs/go-ipfs/issues/4793#issuecomment-371372119 (where we return nothing and no error)?

Stebalien: Good point. We should probably loop.

whyrusleeping:
@Stebalien can i get a rebase here?

Kubuxu:
Rebased

PR #4799, opened 2018-03-10
Update DagModifier.expandSparse() to assign new ipld.Node to the curNode field

ridewindx:
License: MIT
Signed-off-by: Jason Chang <ridewindx@163.com>

### DIFF
--- a/unixfs/mod/dagmodifier.go
+++ b/unixfs/mod/dagmodifier.go
@@ -132,7 +132,11 @@ func (dm *DagModifier) expandSparse(size int64) error {
 		return err
 	}
 	err = dm.dagserv.Add(dm.ctx, nnode)
-	return err
+	if err != nil {
+		return err
+	}
+	dm.curNode = nnode
+	return nil
 }
 
 // Write continues writing to the dag at the current offset


schomatis:
Hey @ridewindx, thanks for the PR. I'm not sure the `curNode` needs to be updated after a call to `appendData`.

It is in fact updated in [`Sync`](https://github.com/ipfs/go-ipfs/blob/5622cc5bb34ddccd2eee7116700bbaf81940d8d4/unixfs/mod/dagmodifier.go#L217), but the function it calls (`trickle.Append`) just returns the same node it received (`curNode`). The function `appendData` doesn't specify in its comments what node it returns, it may be that future append handlers may return different nodes that the one passed as a parameter.

In any case, if indeed it's necessary to update `curNode` it would be easier to follow `Sync`'s pattern where `curNode` is modified as the `appendData` function returns, removing the creation of `nnode`, and thus eliminating the need for the newly added `if err != nil {` clause.

Also, regarding GitCop's complaint, I think the problem may be the missing angle brackets in the email, please refer to the [contributing guidelines](https://github.com/ipfs/community/blob/master/CONTRIBUTING-2.md#a-license-and-a-signed-off-by-trailers-are-required) (GitCop "Guidelines" link should be updated) for instructions on how to set up the git hook to handle that automatically.

ridewindx:
@schomatis Thanks for so detailed reply.

For `trickle.Append`, I have reviewed some parts of the trickle package, and I find that the method `GetDagNode` may return a newly created `ipld.Node`, not the stored `node` field. In addition, as you said, there's no comment for the explanation.

If the `curNode` is always supposed to be updated by  `Sync` method, why the `expandSparse` method exists? It existence should always produce some useful "side effect", is it right? Although, I don't understand the necessity of "grows the file with zero blocks of 4096".

Uh, I will pay attention to the GitCop's complaint.

schomatis:
> For trickle.Append, I have reviewed some parts of the trickle package, and I find that the method GetDagNode may return a newly created ipld.Node, not the stored node field.

https://github.com/ipfs/go-ipfs/blob/ba3cd5b38d5b2e7b07312a3c61231073cdb81d86/importer/helpers/helpers.go#L155-L171

You're correct, the case (I'm assuming) you're referring to is for Filestore support (`posInfo != nil`), but as discussed in another [PR](https://github.com/ipfs/go-ipfs/pull/4734), `Append` won't (currently) support Filestore, so that `if` case shouldn't be taken in this scenario. But anyway, it seems prudent to  update `curNode` just in case. This is just my (partially uninformed) opinion, I don't fully understand how `curNode` should be handled, e.g., the `Truncate` function that calls `expandSparse` does ultimately update it, 

https://github.com/ipfs/go-ipfs/blob/ba3cd5b38d5b2e7b07312a3c61231073cdb81d86/unixfs/mod/dagmodifier.go#L500-L516

but other functions that call it (`Seek` and `WriteAt`) don't, so there may be a pattern I'm missing here.

If we go ahead with this modification please consider modifying this PR to update `curNode` in the same line where [`appendData`](https://github.com/ipfs/go-ipfs/blob/5622cc5bb34ddccd2eee7116700bbaf81940d8d4/unixfs/mod/dagmodifier.go#L217) is called.

> If the curNode is always supposed to be updated by Sync method, why the expandSparse method exists? It existence should always produce some useful "side effect", is it right?

I'm not sure I understand the premise of this question, I don't see the relationship between `Sync` and `expandSparse` you're referring to. I cited earlier the `Sync` function just as an illustration as to how you could update `curNode` in the same line you call `appendData`, but for (the already modified function in your PR) `expandSparse`.

> Although, I don't understand the necessity of "grows the file with zero blocks of 4096".

It's not evident at first sight, but if you study the code of `modifyDag` you'll see that data is written to the tree by way of overwritting already allocated space (buffers),

https://github.com/ipfs/go-ipfs/blob/ba3cd5b38d5b2e7b07312a3c61231073cdb81d86/unixfs/mod/dagmodifier.go#L242-L250

so when a function like `WriteAt` has to write data at some `offset`, it first check that there is enough `size`, otherwise it calls `expandSparse`.

https://github.com/ipfs/go-ipfs/blob/ba3cd5b38d5b2e7b07312a3c61231073cdb81d86/unixfs/mod/dagmodifier.go#L98-L103

This should be more clearly detailed in the `mod` package documentation.

schomatis:
The code has already been refactored in `go-unixfs`.

PR #4800, opened 2018-03-10
Add boolean value to type assertion, for skipping assertion panic

ridewindx:
License: MIT
Signed-off-by: Jason Chang ridewindx@163.com

### DIFF
--- a/mfs/dir.go
+++ b/mfs/dir.go
@@ -405,7 +405,7 @@ func (d *Directory) Path() string {
 	var out string
 	for cur != nil {
 		out = path.Join(cur.name, out)
-		cur = cur.parent.(*Directory)
+		cur, _ = cur.parent.(*Directory)
 	}
 	return out
 }


Stebalien:
Where did you run across this issue? The parent should *always* be a pointer to a directory, as far as I know.

ridewindx:
@Stebalien the parent field of Directory may be the *Root:
`rval, err := NewDirectory(parent, node.String(), node, root, ds)`
[https://github.com/ipfs/go-ipfs/blob/5622cc5bb34ddccd2eee7116700bbaf81940d8d4/mfs/system.go#L93](url)
But *Root is obviously not a *Directory. Uh, is that right?

Stebalien:
Oh. I see. Um, yeah... that can't currently do anything *but* panic (it's currently dead code). Thanks for catching this!

I'd rather use a type switch and *panic* if the parent is neither a directory nor root:

```go
switch parent := cur.parent.(type) {
case *Directory:
	cur = parent
	// ...
case *Root:
	return path
default:
	panic("directory parent neither a directory nor a root")
}
```

Kubuxu:
Could you also add simple unit test for it?

PR #4802, opened 2018-03-10
coreapi: Split the interface into multiple files

magik6k:
The interface file has become quite crowed now, splitting it helps a lot.


core/coreapi/interface/errors.go
```diff
@@ -0,0 +1,6 @@
+package iface
+
+import "errors"
+
+var ErrIsDir = errors.New("object is a directory")
+var ErrOffline = errors.New("can't resolve, ipfs node is offline")
```
Kubuxu: nitpick: var block might be cleaner here.

magik6k:
The only API change other than splitting the interface into multiple files is the removal of mirrored `Node`/`Link` types - https://github.com/ipfs/go-ipfs/pull/4802/files#diff-a8303f091bce9826ba705d0eb8465cc7L30. It didn't really help with anything and only made using/implementing the api require constantly casting things.

PR #4774, opened 2018-03-06
[WIP] coreapi: pubsub, dht, swarm

magik6k:
TODO:
* [ ] Tests
* [ ] Cleanup a bit
* [ ] Review

I'm not sure what to do about the dht api - the commands expose some very verbose information which seems only useful for debugging, and I want to keep the api as simple as possible.

PR #4803, opened 2018-03-10
coreapi: Swarm API

magik6k:
Based on https://github.com/ipfs/go-ipfs/pull/4802

Replaces https://github.com/ipfs/go-ipfs/pull/4774

TODOs:
* [ ] Merge #4802
* [ ] Tests
* [ ] Review


magik6k:
#4807 and #4804 need to get merged first

core/coreapi/swarm.go
```diff
@@ -0,0 +1,147 @@
...
+		}
+
+		out = append(out, ci)
+	}
+
+	return out, nil
+}
+
+func (ci *connInfo) ID() peer.ID {
+	return ci.ID()
```
camelmasa: Sorry for nitpick. This line will loop infinity 🐛

Stebalien: That's not what I'd call a "nitpick".

core/coreapi/interface/swarm.go
```diff
@@ -0,0 +1,37 @@
...
+	ID() peer.ID
+
+	// Address returns the multiaddress via which we are connected with the peer
+	Address() ma.Multiaddr
+
+	// Latency returns last known round trip time to the peer
+	Latency(context.Context) (time.Duration, error)
+
+	// Streams returns list of streams established with the peer
+	// TODO: should this return multicodecs?
```
Stebalien: `protocol.ID`s?

core/coreapi/interface/swarm.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"context"
+	"time"
+
+	peer "gx/ipfs/QmQsErDt8Qgw1XrsXf2BpEzDgGWtB1YLsTAARBup5b6B9W/go-libp2p-peer"
+	ma "gx/ipfs/QmYmsdtJ3HsodkePE3eU3TsCaP2YvPZJ4LoXnNkDE5Tpt7/go-multiaddr"
+)
+
+// PeerInfo contains information about a peer
+type PeerInfo interface {
```
Stebalien: Is this `PeerInfo` or `ConnectionInfo`? It looks like this is a per-connection datastructure.

core/coreapi/interface/swarm.go
```diff
@@ -0,0 +1,37 @@
...
+
+	// Streams returns list of streams established with the peer
+	// TODO: should this return multicodecs?
+	Streams(context.Context) ([]string, error)
+}
+
+// SwarmAPI specifies the interface to libp2p swarm
+type SwarmAPI interface {
+	// Connect to a given address
+	Connect(context.Context, ma.Multiaddr) error
```
Stebalien: Should this take a `peerstore.PeerInfo`? It should probably take either that or a `peer.ID`

magik6k:
> I'd like to port the swarm command over to this interface before merging to make sure it does everything we need (and to test it). Will that blow up the patch too much?

I'm actually doing this right now, for that reason.

magik6k:
(rebased)

core/coreapi/interface/swarm.go
```diff
@@ -0,0 +1,52 @@
...
+// ConnectionInfo contains information about a peer
+type ConnectionInfo interface {
+	// ID returns PeerID
+	ID() peer.ID
+
+	// Address returns the multiaddress via which we are connected with the peer
+	Address() ma.Multiaddr
+
+	// Direction returns which way the connection was established
+	Direction() net.Direction
```
Stebalien: Any reason to have this over `Stat`? The nice thing about `Stat` is that it's extensible.

core/coreapi/interface/swarm.go
```diff
@@ -0,0 +1,52 @@
...
+
+	// Latency returns last known round trip time to the peer
+	Latency(context.Context) (time.Duration, error)
```
Stebalien: As with the pubsub API, we need to consider what this parameter signals. Personally, when I see something taking a context, I immediately think that it does some kind of network operation.

core/coreapi/interface/swarm.go
```diff
@@ -0,0 +1,52 @@
...
+	// Connect to a given peer
+	Connect(context.Context, pstore.PeerInfo) error
+
+	// Disconnect from a given address
+	Disconnect(context.Context, ma.Multiaddr) error
+
+	// Peers returns the list of peers we are connected to
+	Peers(context.Context) ([]ConnectionInfo, error)
+
+	KnownAddrs(context.Context) (map[peer.ID][]ma.Multiaddr, error)
```
Stebalien: *cough* docs *cough* :smile:.

core/coreapi/swarm.go
```diff
@@ -0,0 +1,180 @@
...
+}
+
+func (api *SwarmAPI) Connect(ctx context.Context, pi pstore.PeerInfo) error {
+	if api.node.PeerHost == nil {
+		return coreiface.ErrOffline
+	}
+
+	swrm, ok := api.node.PeerHost.Network().(*swarm.Swarm)
+	if !ok {
+		return fmt.Errorf("peerhost network was not swarm")
```
Stebalien: If possible, can we just make this optional? That is, clear the backoff iff we *have* a swarm but don't error out.

Stebalien: (I realize this probably isn't what we do now but, well, no time better than now)

core/coreapi/interface/swarm.go
magik6k: Assuming you mean `go-libp2p-net.Stat` - `ipfs swarm peers` doesn't return info from the `Extra` map, so we would have to expose it somehow for a good implementation in go-ipfs-api.

Stebalien: Got it. Yeah, that may be tricky as it could have arbitrary data. We can punt on that.

PR #4804, opened 2018-03-10
coreapi: DHT API

magik6k:
Based on https://github.com/ipfs/go-ipfs/pull/4802

Replaces https://github.com/ipfs/go-ipfs/pull/4774

TODOs:
* [x] Merge #4802
* [x] Tests
* [ ] Review


magik6k:
https://github.com/ipfs/go-ipfs/pull/4807 should get merged first

Stebalien:
@magik6k is this ready for review?

magik6k:
Rebased, should be ready for a review

Kubuxu:
CI is red

magik6k:
I think there is a random test fail, will try to fix it first

core/coreapi/dht.go
```diff
@@ -0,0 +1,241 @@
...
+
+var ErrNotDHT = errors.New("routing service is not a DHT")
+
+type DhtAPI CoreAPI
+
+func (api *DhtAPI) FindPeer(ctx context.Context, p peer.ID) (<-chan ma.Multiaddr, error) {
+	dht, ok := api.node.Routing.(*ipdht.IpfsDHT)
+	if !ok {
+		return nil, ErrNotDHT
+	}
```
Stebalien: Why not just call `FindPeer` directly on `ipfs.node.Routing`? The interface requires that method.

core/coreapi/dht.go
```diff
@@ -0,0 +1,241 @@
...
+		pi, err := dht.FindPeer(ctx, peer.ID(p))
+		if err != nil {
+			notif.PublishQueryEvent(ctx, &notif.QueryEvent{
+				Type:  notif.QueryError,
+				Extra: err.Error(),
+			})
+			return
+		}
+
+		notif.PublishQueryEvent(ctx, &notif.QueryEvent{
```
Stebalien: Why not just call `FindPeer`, get the result, return it to the user, and be done with it? We're throwing away all the other events anyways.

core/coreapi/dht.go
```diff
@@ -0,0 +1,241 @@
...
+	return outChan, nil
+}
+
+func (api *DhtAPI) FindProviders(ctx context.Context, p coreiface.Path, opts ...caopts.DhtFindProvidersOption) (<-chan peer.ID, error) {
+	settings, err := caopts.DhtFindProvidersOptions(opts...)
+	if err != nil {
+		return nil, err
+	}
+
+	dht, ok := api.node.Routing.(*ipdht.IpfsDHT)
```
Stebalien: Same here. We shouldn't be casting this. If we need some new functionality, we should add that to the interface.

core/coreapi/dht.go
```diff
@@ -0,0 +1,241 @@
...
+				}
+			}
+		}
+	}()
+
+	go func() {
+		defer close(events)
+		for p := range pchan {
+			np := p
+			notif.PublishQueryEvent(ctx, &notif.QueryEvent{
```
Stebalien: Again, maybe I'm missing something but this seems like pointless complexity. The caller of this method (`FindProviders`) can't even use these events because we're registering for (and consuming) them.

core/coreapi/dht.go
```diff
@@ -0,0 +1,241 @@
...
+	settings, err := caopts.DhtProvideOptions(opts...)
+	if err != nil {
+		return err
+	}
+
+	if api.node.Routing == nil {
+		return errors.New("cannot provide in offline mode")
+	}
+
+	if len(api.node.PeerHost.Network().Conns()) == 0 {
```
Stebalien: Oof. This is going to hurt. It's fine to do this from the command as we amortize over the request and multiple arguments but listing all the connections just to see if we *have* some connections is going to be very expensive. Ideally, the DHT would detect this and return an error. Does it not do that?

core/coreapi/dht.go
```diff
@@ -0,0 +1,241 @@
...
+			case <-ctx.Done():
+				return
+			default:
+			}
+		}
+	}()*/
+
+	//defer close(events)
+	if settings.Recursive {
+		err = provideKeysRec(ctx, api.node.Routing, api.node.DAG, []*cid.Cid{c})
```
Stebalien: I'm pretty sure this needs to be an offline DAG.

core/coreapi/coreapi.go
```diff
@@ -59,6 +59,11 @@ func (api *CoreAPI) Pin() coreiface.PinAPI {
 	return (*PinAPI)(api)
 }
 
+// Dht returns the DhtAPI interface implementation backed by the go-ipfs node
+func (api *CoreAPI) Dht() coreiface.DhtAPI {
```
Stebalien: High level comment: I'd rather not *have* a DHT API. The DHT is just a backend for other stuff (the routing interface). While I know we currently have a command, I'd prefer not to bake this into the CoreAPI. Is there a design discussion for the CoreAPI anywhere?

core/coreapi/dht.go
```diff
@@ -0,0 +1,241 @@
...
+		err := r.Provide(ctx, c, true)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func provideKeysRec(ctx context.Context, r routing.IpfsRouting, dserv ipld.DAGService, cids []*cid.Cid) error {
+	provided := cid.NewSet()
```
Stebalien: This is just a copy of the existing functionality but... it's going to hurt regardless. Could we get a TODO: Use a bloom filter?

core/coreapi/dht.go
```diff
@@ -0,0 +1,241 @@
...
+	for _, c := range cids {
+		kset := cid.NewSet()
+
+		err := dag.EnumerateChildrenAsync(ctx, dag.GetLinksDirect(dserv), c, kset.Visit)
```
Stebalien: Same, could we get a "TODO: provide and enumerate in parallel with a small buffer".

core/coreapi/interface/dht.go
```diff
@@ -0,0 +1,24 @@
...
+	"context"
+
+	options "github.com/ipfs/go-ipfs/core/coreapi/interface/options"
+
+	ma "gx/ipfs/QmWWQ2Txc2c6tqjsBpzg5Ar652cHPGNsQQp2SejkNmkUMb/go-multiaddr"
+	peer "gx/ipfs/QmZoWKhxUmZ2seW4BzX6fJkNR8hh9PsGModr7q171yq2SS/go-libp2p-peer"
+)
+
+// DhtAPI specifies the interface to the DHT
+type DhtAPI interface {
```
Stebalien: I'd rather use the interface defined in https://github.com/libp2p/go-libp2p-routing/ (although we'd probably want to change it to return channels). Alternatively, we could do this for now with a big "we're going to change this" todo.

core/coreapi/coreapi.go
magik6k: > Is there a design discussion for the CoreAPI anywhere?

I'm trying to follow the js coreapi spec in most places - for DHT https://github.com/ipfs/interface-ipfs-core/blob/master/SPEC/DHT.md.

I did only include the most basic functions which may be useful in some cases.

core/coreapi/dht.go
magik6k: This was it the dht command.

Looking at provide code in kad-dht, it invokes `dht.GetClosestPeers`, which I think will error in that case (correct me if I'm wrong) -> https://github.com/libp2p/go-libp2p-kad-dht/blob/2bab49ef85b2d8547e95ca2a699b5be562cc78b6/lookup.go#L60

magik6k:
Thanks for the rreview and pointers.

> Any chance we could get the DHT commands converted over to use this API before merging

I guess it's probably better to get this done sooner than later. This is going to be at least a slightly breaking change. Currently when you run some dht command with `--enc=json` you get a really verbose output of everything the dht is doing. Imo it's weird to produce output that verbose in API, so I think it's fine to clean that up. Is there anything that depends on this output?

core/coreapi/coreapi.go
Stebalien: :+1:

I've submitted an issue to discuss this (https://github.com/ipfs/interface-ipfs-core/issues/249). We can probably move forward discuss this in parallel. It's not the end of the world, just a bit unfortunate.

Stebalien:
> Is there anything that depends on this output?

Not sure... @whyrusleeping should comment on that. I believe it was introduced intentionally as DHT commands can be a bit slow. Note: we can actually avoid breaking this if we have the commands continue to use the output of the query notification channel instead of the output of this command.

camelmasa:
One question, Will you add DHT's GetClosestPeers interface to DHT Core API ❓ 
It's `ipfs dht query <peerID>` in IPFS command.

whyrusleeping:
Whats the blocker here? The DhtApi as defined here looks okay to me, we can't really make it exactly the same as in routing because we need to be able to pass options, but its close enough no?

magik6k:
@whyrusleeping see the discussion in https://github.com/ipfs/interface-ipfs-core/issues/249

core/coreapi/dht.go
magik6k: Will get that with https://github.com/ipfs/go-ipfs/pull/4333 for free

core/coreapi/dht.go
magik6k: https://github.com/ipfs/go-cid/issues/57

magik6k:
So there is the discussion on naming the api in https://github.com/ipfs/interface-ipfs-core/issues/249. As this pr is a dependency of few other PRs, can we just put a big 'This will most likely change' comment on this for now?

whyrusleeping:
@magik6k I'm not opposed to that.

magik6k:
Rebased+Added the comment

magik6k:
@Stebalien @Kubuxu Mind giving this another round of review?

thirdparty/streaming-cid-set/set.go
```diff
@@ -0,0 +1,38 @@
+package streamingset
```
Stebalien: Dammit! We really need to make gx less painful.

Stebalien: Actually, I'll just handle this. Mind making a PR against go-cid?

core/coreapi/dht.go
```diff
@@ -0,0 +1,135 @@
...
+				return err
+			}
+		case err := <-errCh:
+			return err
+		case <-ctx.Done():
+			return ctx.Err()
+		}
+	}
+
+	return nil
```
Stebalien: Should remove this to get the tests to pass.

magik6k:
(should be ready to merge, though we might want one more review)

core/coreapi/dht.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+func provideKeysRec(ctx context.Context, r routing.IpfsRouting, bs blockstore.Blockstore, cids []*cid.Cid) error {
+	provided := cidutil.NewStreamingSet()
+
+	errCh := make(chan error)
+	go func() {
+		for _, c := range cids {
+			dserv := dag.NewDAGService(blockservice.New(bs, offline.Exchange(bs)))
```
Stebalien: Let's construct this *once*.

core/coreapi/interface/options/dht.go
```diff
@@ -0,0 +1,62 @@
...
+	return options, nil
+}
+
+type dhtOpts struct{}
+
+var Dht dhtOpts
+
+// WithRecursive is an option for Dht.Provide which specifies whether to provide
+// the given path recursively
+func (dhtOpts) WithRecursive(recursive bool) DhtProvideOption {
```
Stebalien: What's the rational behind the `With` prefix?

Stebalien:
Hm. Actually, is there any way we could get some better test coverage or will that come when we switch commands over to using this interface?

core/coreapi/interface/options/dht.go
magik6k: Options before refractor used it, this didn't get updated with rebase. Will fix

magik6k:
coverage++ (73% on coreapi/dht.go)

(Circle appears to be borked/annoyingly slow - https://github.com/ipfs/go-ipfs/pull/5389 may or may not be related)

PR #4806, opened 2018-03-10
Stateful plugin loading

PlayerWithoutName:
Depends on #4506
----

The plugins are currently loaded during Executor creation, which takes some extensibility from them. This PR makes them 'stateful', which will allow for more flexibility.


core/builder.go
```diff
@@ -15,6 +15,7 @@ import (
 	dag "github.com/ipfs/go-ipfs/merkledag"
 	resolver "github.com/ipfs/go-ipfs/path/resolver"
 	pin "github.com/ipfs/go-ipfs/pin"
+	"github.com/ipfs/go-ipfs/plugin/loader"
```
magik6k: Missed prefix

Kubuxu:
The `ConstructNode` isn't the right place to pass this information. `ConstructNode` will not be called when running `ipfs daemon` so the `Plugins` can't be passed to the core node itself or used by anything really.

The problem we are having here is: want want to use plugins pre and post commands lib execution. Best course of action I see here is adding `Plugins` variable to `oldcmds.Context`. 

This way the plugins can be loaded before command execution and then latter for example to use created `core.Node`.

@Stebalien what do you think?

whyrusleeping:
cc @frrist

Kubuxu:
It might help with Tracing API as you will be able to access config, just at latter stage after initial `Run` to the plugin. We could also make `Init` run before node and `Run` after the node is created but you won't be able to pass `core.Node` instance due to circular dependency.

frrist:
The main reason for needing access to the nodes config was to determine the peerId of the node running the plugin, this was how we were going to name the tracer. 
Currently this is being solved via an env variable since as @magik6k pointed out, the peerId may not be available if `ipfs init` hasn't been ran. If this allows us to pass the nodes config to a plugin after the node has been initialized that would be great!

PlayerWithoutName:
Yeah, i think after this, you could create like ConfigConsumer interface, implement it on the tracing plugin, then create a method that provides the config for all the plugins that implement that interface and you should be good.

Kubuxu:
@PlayerWithoutName could you rebase your PR. To solve conflicts with https://github.com/ipfs/go-ipfs/pull/4506 please base it on `feat/opentrace` branch.

commands/request.go
```diff
@@ -18,6 +19,7 @@ type Context struct {
 	Online     bool
 	ConfigRoot string
 	ReqLog     *ReqLog
+	Plugins    *loader.PluginLoader
```
magik6k: I'd separate this with a newline from the group above.

plugin/loader/loader.go
```diff
@@ -38,28 +55,21 @@ func LoadPlugins(pluginDir string) ([]plugin.Plugin, error) {
...
-	_, err := os.Stat(pluginDir)
-	if os.IsNotExist(err) {
-		return nil, nil
-	}
-	if err != nil {
-		return nil, err
-	}
+// Initialize all the loaded plugins
+func (loader *PluginLoader) Initialize() error {
+	return initialize(loader.plugins)
```
magik6k: Why not just put these funcs on the struct directly?

plugin/loader/initializer.go
```diff
@@ -19,8 +20,9 @@ func initialize(plugins []plugin.Plugin) error {
 	return nil
 }
 
-func run(plugins []plugin.Plugin) error {
-	for _, pl := range plugins {
+//Run the plugins
+func (loader *PluginLoader) Run() error {
```
Kubuxu: It is good practice to keep the implementation of given structure in one package.
In this case, it means keeping those functions in one file with the definition of the struct.

Kubuxu:
@Stebalien want to review this?

whyrusleeping:
@Kubuxu this one needs a rebase. Then i'll merge it in

PlayerWithoutName:
Rebased

cmd/ipfs/main.go
```diff
@@ -47,6 +47,44 @@ const (
 	heapProfile        = "ipfs.memprof"
 )
 
+type cmdInvocation struct {
```
whyrusleeping: Looks like some code that got deleted in another PR is getting reintroduced here. cc @magik6k

PlayerWithoutName: Fixed, it was legacy code accidentally put up with second commit.

PR #4807, opened 2018-03-11
coreapi: Refactor options

magik6k:
Based on #4802

Suggested by @Stebalien and @Kubuxu in https://github.com/ipfs/go-ipfs/pull/4643#discussion_r166689061.

### DIFF
--- a/core/coreapi/pin_test.go
+++ b/core/coreapi/pin_test.go
@@ -121,7 +121,7 @@ func TestPinRecursive(t *testing.T) {
 		t.Errorf("unexpected pin list len: %d", len(list))
 	}
 
-	list, err = api.Pin().Ls(ctx, opt.Pin.Type("direct"))
+	list, err = api.Pin().Ls(ctx, opt.Pin.Type.Direct())
 	if err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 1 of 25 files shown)


Kubuxu:
Diff based on #4802 https://github.com/ipfs/go-ipfs/compare/feat/coreapi/split-iface...feat/coreapi/refactor-opts

core/coreapi/pin_test.go
```diff
@@ -145,7 +147,7 @@ func TestPinRecursive(t *testing.T) {
 		t.Error("unexpected path")
 	}
 
-	list, err = api.Pin().Ls(ctx, api.Pin().WithType("indirect"))
+	list, err = api.Pin().Ls(ctx, opt.Pin.Type("indirect"))
```
Kubuxu: Nicer UX would be if we had already defined functions for different pin types. Like `opt.Pin.TypeIndirect` or `opt.Pin.Type.Indirect`

magik6k: Done

Kubuxu:
See the codeclimate (https://codeclimate.com/github/ipfs/go-ipfs/pull/4807) you can omit the _ in function receiver if it is not used.

magik6k:
Done

Stebalien:
So, jenkins passed. It looks like something went wrong with the webhook.

whyrusleeping:
nice! This does feel much nicer.

Issue #4808, opened 2018-03-12
Windows initiative 2018


djdv:
This issue serves as an anchor to monitor the broader topic of `go-ipfs` on Windows.
Here I will be highlighting the current issues being focused on and the PR efforts to resolve them. Feel free to add anything to this topic that relates to `go-ipfs` on Windows as a whole.

**Build problems**
- [x] `make`
Fails to detect `gx` and `gx-go` properly
https://github.com/ipfs/go-ipfs/issues/4510
Resolved in: https://github.com/ipfs/go-ipfs/pull/4682
- [x] `gx`
Fails to detect subtools properly (gx installs deps to wrong directory)
https://github.com/whyrusleeping/gx/issues/153
Resolved in: https://github.com/whyrusleeping/gx/pull/154

- [x] Build documentation
A lot of people seem to have trouble building on Windows with the current document. I've proposed a revision here: https://github.com/ipfs/go-ipfs/pull/4691

**interface problems**

- [x] Pipes
https://github.com/ipfs/go-ipfs/issues/3266
Resolved in: https://github.com/ipfs/go-ipfs-cmds/pull/74

- [x] `get` doesn't support NUL
_currently not tracked_
This will require fixes in Go itself and the tar-utils package
https://github.com/golang/go/issues/24482
https://github.com/golang/go/issues/24556
https://github.com/whyrusleeping/tar-utils/pull/2
~~Resolved in: https://github.com/ipfs/go-ipfs/pull/4956 (not yet merged)~~
Resolved independently in tar-utils, rolled in via https://github.com/ipfs/go-ipfs/pull/4857

- [x] Color/Colour output
https://github.com/ipfs/go-ipfs/issues/3068
https://github.com/ipfs/go-ipfs/issues/2124
Resolved in: https://github.com/ipfs/go-log/pull/21 + https://github.com/ipfs/go-ipfs/pull/5007

- [ ] File system character restrictions
https://github.com/ipfs/go-ipfs/issues/2013
Resolved on Windows with: https://github.com/whyrusleeping/tar-utils/pull/2 (merged) + https://github.com/ipfs/go-ipfs/pull/4956 (not yet merged)

**Windows Tests & CI**
We need to pass tests on Windows so new issues can be caught by the CI before becoming a problem.
When Windows is skipped for testing, we suffer from regressions such as:
 https://github.com/ipfs/go-ipfs/issues/4750
 https://github.com/ipfs/go-ipfs/issues/4512
[...]

PR #4809, opened 2018-03-12
Fix get output path

djdv:
This return value is used to open a platform native handle, as such platform native paths should be used.
Reference: https://github.com/ipfs/go-ipfs/issues/4808

Edit: put angle brackets around email to make gitcop happy

### DIFF
--- a/core/commands/get.go
+++ b/core/commands/get.go
@@ -6,7 +6,7 @@ import (
 	"fmt"
 	"io"
 	"os"
-	gopath "path"
+	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


whyrusleeping:
I'm losing track of how many times this has been 'fixed' :/

Kubuxu:
@whyrusleeping this is something different, we were fixing `-o` not working with directories before. This one is about platform specific directory separators.

djdv:
Closing this in favour of https://github.com/ipfs/go-ipfs/pull/4850

djdv:
As it turns out, tar-utils is being moved to a new repo so I'm re-opening this one.

PR #4817, opened 2018-03-15
transport refactor update

Stebalien:
Also:

* update go-testutil
* update the datastore interface (WIP)
* update go-libp2p-record to remove useless DHT signatures

Blocked on:

* Transport refactor.
* Namesys republisher DHT fixes.

DO NOT MERGE

This PR will never be merged as-is, it's simply for review and testing.

https://github.com/libp2p/go-libp2p/issues/297


core/builder.go
```diff
@@ -42,6 +44,10 @@ type BuildCfg struct {
 	// that will improve performance in long run
 	Permanent bool
 
+	// DisableEncryptedConnections disables connection encryption *entierly*.
```
whyrusleeping: typo. entierly

core/commands/swarm.go
```diff
@@ -90,10 +90,13 @@ var swarmPeersCmd = &cmds.Command{
 				Peer: pid.Pretty(),
 			}
 
-			swcon, ok := c.(*swarm.Conn)
-			if ok {
-				ci.Muxer = fmt.Sprintf("%T", swcon.StreamConn().Conn())
-			}
+			/*
+				// FIXME(steb):
+							swcon, ok := c.(*swarm.Conn)
```
whyrusleeping: definitely no need to reproduce the exact output here. But maybe we can print out the full transport.String() or something?

core/commands/swarm.go
```diff
@@ -384,14 +383,13 @@ ipfs swarm connect /ip4/104.131.131.82/tcp/4001/ipfs/QmaCpDMGvV2BGHeYERUEnRQAwe3
 			return
 		}
 
-		snet, ok := n.PeerHost.Network().(*swarm.Network)
+		// FIXME(steb): Nasty
+		swrm, ok := n.PeerHost.Network().(*swarm.Swarm)
```
whyrusleeping: ah, because we need to expose the backoff logic... It probably makes sense to just embed that information in the context.

whyrusleeping:
```go
ctx := context.WithValue(ctx, "swarmClearBackoff", true)
n.Peerhost.Connect(ctx, ...)
```

whyrusleeping: though, downside there is that it becomes much harder to notice when it stops working.

Stebalien: For now, I'm planning on leaving this as it is. This is more for my future self.

core/commands/swarm.go
Stebalien: Unfortunately, that won't include information about the muxer. Currently, we'd show:

```
<swarm.Conn[TCP] /local/maddr (<peer.ID local>) <-> /remote/maddr (<peer.ID remote>)>
```

(which should actually be cleaned up a bit...)

I've been toying with adding an `Info` function to transport conns (modeled after `Context.Value`):

```go
func Info(key interface{}) interface{}
```

This would allow querying for arbitrary information.

Stebalien:
I've been using this locally but haven't bothered fixing the gx deps.

@Kubuxu last review :). This should be a fairly short one (negative diffstat!).

Stebalien:
@Kubuxu I've pushed the gx dep update. Should hopefully build.

THIS PR IS NOT READY TO BE MERGED. It brings in an update to go-libp2p-kad-dht that needs https://github.com/ipfs/go-ipfs/pull/4753.

Stebalien:
This PR is pending a last-minute time-boxed review by @jbenet. This PR and it's dependencies will start to merge themselves (:man_shrugging:) tonight unless anyone objects (please don't).

:bomb:

(you have been sufficiently warned)

whyrusleeping:
🙌 🎊 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉

whyrusleeping:
leeeeroyyyyyyyyyy jenkinsssssss!

PR #4819, opened 2018-03-15
misc: spelling of retrieval

Kubuxu:
Closes #4161


whyrusleeping:
had to google this one myself... nice catch @Kubuxu :)

Kubuxu:
It was #4161. If not for the in-line highlight I would not be able to spot the change :smile:

Issue #4818, opened 2018-03-15
0.4.14-rcX fails js-ipfs-api test suite


Kubuxu:
Reference https://github.com/ipfs/js-ipfs-api/issues/486#issuecomment-373332236

Test output: https://friendpaste.com/72TbssZSoSxmdOpKxT67je

magik6k:
I've done few more runs:

On 0.4.13:
* https://hastebin.com/pixokuqaxi.vbs

On 0.4.14-rc2
* https://hastebin.com/sesacejiye.vbs
* https://hastebin.com/omuhesuwip.vbs
* https://hastebin.com/molexizoji.vbs

Kubuxu:
@VictorBjelkholm @diasdavid could we get some help interpreting these.

I see that there are some that are false positive lookalike. Like the `[]` vs `[null]`.

magik6k:
Changes I see:
* `bitswap/wantlist` now returns `{"Keys":[]}` instead of `{"Keys":null}` for empty wantlist
* `files/stat` fail is caused by changes in https://github.com/ipfs/js-ipfs-api/pull/695 and slightly wrong testcase. on 0.4.14-rc2 `files/stat` returns:
```
{
  "Hash": "Qm...",
  "Size": 1088,
  "CumulativeSize": 1099,
  "Blocks": 0,
  "Type": "file"
}

```

and with --with-local:
```
{
  "Hash": "Qm...",
  "Size": 1088,
  "CumulativeSize": 1099,
  "Blocks": 0,
  "Type": "file",
  "WithLocality": true,
  "Local": true,
  "SizeLocal": 1099
}
```
* Not sure about DHT fails, it happened on 0.4.13 too
* `pubsub/sub` returned an empty object (`{}`) after request, this has been fixed in 0.4.14-rc2
* The other errors seem to be related to the fact that some tests failed.

diasdavid:
Try with https://github.com/ipfs/js-ipfs-api/releases/tag/v18.1.2 test suite. Currently js-ipfs-api Master is in flux due to many Windows + other fixes that were happening in parallel and complicating things, next week it will be golden again.

magik6k:
Tried with js-ipfs-api 18.1.2 + go-ipfs 0.4.14-rc2, fails mostly in the same ways as the master: https://hastebin.com/gigiforudu.vbs

whyrusleeping:
@VictorBjelkholm can we get some automated CI for running js-ipfs-api tests?

[...]

PR #4822, opened 2018-03-16
fix: use EmitOnce where only single response is to be sent

Kubuxu:
Part of #4818 
Also fixes pubsub ls breakage.

License: MIT


whyrusleeping:
@Kubuxu @VictorBjelkholm do we have js-ipfs-api tests running in CI yet?

whyrusleeping:
In any case I think this is ready to go. We'll have to ship another rc, but before we do that, can we make sure theres nothing else weird thats gonna come up?

Kubuxu:
@whyrusleeping no we don't but go-ipfs artefacts are exported so it should be quite easy to add from js-ipfs-api side.

PR #4826, opened 2018-03-17
upgrade commands that used StringArguments

whyrusleeping:
This should make the fix that @Stebalien is working on for the cmds lib argument stdin parsing a little bit simpler.

License: MIT

### DIFF
--- a/core/commands/files.go
+++ b/core/commands/files.go
@@ -57,7 +57,7 @@ operations.
 	},
 	Subcommands: map[string]*cmds.Command{
 		"read":  lgc.NewCommand(filesReadCmd),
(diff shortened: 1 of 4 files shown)


whyrusleeping:
The goal here is to remove all usage of `StringArguments` so that better argument handling can be implemented.

core/commands/files.go
```diff
@@ -1072,6 +1062,35 @@ Remove files or directories.
 	},
 }
 
+func getPrefixNew(req *cmds.Request) (*cid.Prefix, error) {
```
Stebalien: :cry:

(I guess we can fix this later)

Issue #4213, opened 2017-09-06
automatically try relay 


whyrusleeping:
Now that we have relay support its time to start really solving problems with it. It seems to me that the first big thing to look at here is automatically trying to connect via relay if normal methods fail. This is hard for a variety of reasons. We dont want to use a relay if we can possibly connect directly, but we dont want to sit around waiting forever for dial attempts to fail. We also then have to detect if using relay is even an option. For that to be the case we have to know that our target supports relay STOP, and that we are connected to a relay HOP peer that is connected to them.

Once we have a relayed connection to a peer, we should ask them to attempt making a direct connection to us, and if they are successful, use that instead of the relayed connection. This implies we need a good method of migrating our 'chosen' connection to another peer over to another Conn.

There is a lot here, but I wanted to open up the discussion so we can make a plan of attack on getting this all working.

cc @vyzo @Stebalien @diasdavid @Kubuxu @dryajov

dryajov:
Glad to see this issue! 👍 

- Currently we have the `CAN_HOP` message which allows discovering HOPs but not STOPs, being able to do that would be supper useful. I've seen several proposals for both relay capabilities discovery as well as a more general capabilities discovery - it be nice to consolidate those, I'll search around and try to link those issues here.

> We dont want to use a relay if we can possibly connect directly, but we dont want to sit around waiting forever for dial attempts to fail

- This is how I currently handle it in JS:
   - When a `/p2p-circuit` address is being dialed
       - if there are other transports for that peer that we know of, then those are tried first, falling back to circuit if all attempts to dial have failed
       - if there are no dialable transports at all for that peer, then `circuit` is used
    - When a non `/p2p-circuit` address is dialed, circuit is used only as a fallback when all other transports have failed.

Just wanted to document this here, as a possible approach. There also might be some differences on how the GO and the JS implementations currently handle this cases - it is something to investigate and possibly sync up across both implementation, but I know that the core cases should work for both, which should be good enough for now? I believe this can use a separate issue as well.

- On connecting directly 
   - Do we currently have any way to upgrade connections? Would this even qualify as a  connection upgrade? 
       - One approach could be, to dial the peer again after a circuit connection has been established, after asking identify (or something else?) for addition info on the peer, if that connection succeeds, use it, otherwise keep using the relayed connection. Just some thoughts.

dryajov:
Opened this issue to talk about implementation differences - https://github.com/libp2p/js-libp2p-circuit/issues/16

[...]

PR #4830, opened 2018-03-18
Announce relay addresses by default

magik6k:
Depends on https://github.com/libp2p/go-libp2p-swarm/pull/57

With this change we will announce the non-specific relay addresses. From my testing non-specific addresses work well-enough, at least with the current network structure.

There are also a couple of points in https://github.com/ipfs/go-ipfs/issues/4213 which may need to be addressed first for this to work well.

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -869,17 +869,8 @@ func constructPeerHost(ctx context.Context, id peer.ID, ps pstore.Peerstore, bwr
 		hostOpts = append(hostOpts, opts.ConnectionManager)
 	}
 
-	addrsFactory := opts.AddrsFactory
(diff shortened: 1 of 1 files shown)


vyzo:
i think long term we want to figure out a way to bind to relays and announce specific addresses.

bigs:
Note that this would also close #4992

vyzo:
Can we reopen this?

vyzo:
Related:
- https://github.com/libp2p/go-libp2p-circuit/pull/48
- https://github.com/libp2p/specs/issues/72

vyzo:
Yeah, we probably want to close this.

PR #4832, opened 2018-03-19
Use prefix in bootstrapWritePeers

magik6k:
This is probably what should happen here.


whyrusleeping:
That one seems like an oooold bug. Nice work.

PR #4834, opened 2018-03-19
Remove roadmap.md

magik6k:
* It hasn't been updated since 2 years (and zero times since it was created)
* The repo isn't really a great place to track progress of anything, especially with the 'no merging without review' policy.
  * Issues work nicely for this
* It looks like everything from this list is either done or has an issue


Stebalien:
Yeah, we usually use issues for this.

PR #4836, opened 2018-03-19
ipfs release 0.4.14-rc3

whyrusleeping:
@Kubuxu @magik6k wanna run through the api tests to make sure we've caught everything?

Hopefully this is it :)

License: MIT

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,13 @@ did allow manually creating them and fetching them from other peers. If you
 currently have objects using insecure hashes in your local ipfs repo, please
 remove them before updating.
 
+#### Changes from rc2 to rc3
+- Fix bug in stdin argument parsing ([ipfs/go-ipfs#4827](https://github.com/ipfs/go-ipfs/pull/4827))
+- Revert commands back to sending a single response ([ipfs/go-ipfs#4822](https://github.com/ipfs/go-ipfs/pull/4822))
+
+#### Changes from rc1 to rc2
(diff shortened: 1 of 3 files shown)


VictorBjelkholm:
One test is failing, `dht.provide` fails with `SyntaxError: Unexpected token a in JSON at position 0`. Investigating if it's bc js-ipfs-api or go-ipfs

VictorBjelkholm:
So, the URL: `/api/v0/dht/provide?arg%5Bcodec%5D=dag-pb&arg%5Bversion%5D=0&arg%5Bmultihash%5D=%12%20%db%97%05%a3J%da%c4%00%98%08%b6%15%e2ld%3f%7c%1cp%e7%a7%21%ae%0b%c8%fe%cf%05%b1%7b%ac%00&stream-channels=true` used to work for calling `dht.provide` but is not valid anymore. Can you check if this is a intentional change on your side?

Stebalien:
It looks like JS is sending a decoded multihash as the argument. That command never actually worked, it just used to fail silently for some reason.

Stebalien:
However, https://github.com/ipfs/go-ipfs/issues/4837...

VictorBjelkholm:
> That command never actually worked

Yeah, seems the tests is just testing if there is any errors, doesn't actually check if it was provided.

image

Other than that, tests run fine.

Stebalien:
@VictorBjelkholm could you file a bug against both js-ipfs-api and js-ipfs? That request should fail on js-ipfs.

whyrusleeping:
Alright, with the helptext fix in, and signoffs (in IRC) from @VictorBjelkholm and @magik6k that api tests look green. I think we're ready to go.

PR #4840, opened 2018-03-19
check arguments after handling help

### LINKED ISSUE #4837: --help is broken for commands that require an argument that can be read from stdin
Unfortunately, we try to read the argument from stdin *before* parsing the help flag. To reproduce, run `ipfs dht provide --help`.

Stebalien:
otherwise, we block on required arguments from stdin before processing the help flag.

fixes #4837

### DIFF
--- a/test/sharness/t0100-name.sh
+++ b/test/sharness/t0100-name.sh
@@ -99,6 +99,12 @@ test_expect_success "'ipfs name publish' fails" '
   printf '' | test_expect_code 1 ipfs name publish >publish_out 2>&1
 '
 
+test_expect_success "'ipfs name publish --help' succeeds" '
+  PEERID=`ipfs id --format="<id>"` &&
+  test_check_peerid "${PEERID}" &&
+  ipfs name publish --help
+'
+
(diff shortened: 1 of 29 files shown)


test/sharness/t0100-name.sh
```diff
@@ -99,6 +99,12 @@ test_expect_success "'ipfs name publish' fails" '
   printf '' | test_expect_code 1 ipfs name publish >publish_out 2>&1
 '
 
+test_expect_success "'ipfs name publish --help' succeeds" '
+  PEERID=`ipfs id --format="<id>"` &&
```
whyrusleeping: whats with the ipfs id check?

Stebalien: Copied from the other tests. I have no idea but I assume it's to make sure that publishing would otherwise work (although I guess it shouldn't really be necessary in this case).

Stebalien: (fixed, and put the tests in the correct order)

whyrusleeping:
Alright, @Kubuxu @magik6k @VictorBjelkholm does this one look good still?

PR #4844, opened 2018-03-20
fix(mfs): Directory.Path not working, add test

Kubuxu:
Credit goes to @ridewindx

Closes #4800

License: MIT


whyrusleeping:
Thanks @ridewindx and @Kubuxu !

PR #4846, opened 2018-03-20
Fix missing profile docs

### LINKED ISSUE #4843: some profile are missing in the ipfs init command documentation
#### Version information:
go-ipfs version: 0.4.14-rc2-
Repo version: 6
System version: amd64/linux
Golang version: go1.10rc2

#### Type:
Bug

#### Description:
`ipfs init --help` list only 2 profiles (server and test) while more actually exist (https://github.com/ipfs/go-ipfs/pull/4195/files#diff-d39a8a737bc5c13f36e41aa92a0c7602). Ideally, this documentation would be generated to be always up to date.

Additionally, profile coule be listed in `ipfs config profile apply` as well.

magik6k:
Fixes #4843

### DIFF
--- a/repo/config/profile.go
+++ b/repo/config/profile.go
@@ -5,6 +5,15 @@ import "time"
 // Transformer is a function which takes configuration and applies some filter to it
 type Transformer func(c *Config) error
 
+// Profile contains the profile transformer the description of the profile
+type Profile struct {
+	// Description briefly describes the functionality of the profile
+	Description string
+
+	// Transform takes ipfs configuration and applies the profile to it
+	Transform Transformer
+}
+
 // defaultServerFilters has a list of non-routable IPv4 prefixes
 // according to http://www.iana.org/assignments/iana-ipv4-special-registry/iana-ipv4-special-registry.xhtml
 var defaultServerFilters = []string{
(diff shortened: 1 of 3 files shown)


repo/config/profile.go
```diff
@@ -24,54 +33,96 @@ var defaultServerFilters = []string{
...
+var Profiles = map[string]Profile{
+	"server": {
+		Description: `Disables local host discovery, recommended when
+running IPFS on machines with public IPv4 addresses.`,
+
+		Transform: func(c *Config) error {
+			c.Addresses.NoAnnounce = appendSingle(c.Addresses.NoAnnounce, defaultServerFilters)
+			c.Swarm.AddrFilters = appendSingle(c.Swarm.AddrFilters, defaultServerFilters)
+			c.Discovery.MDNS.Enabled = false
+			return nil
```
Kubuxu: Quick Q: shouldn't server profile also disable UPnP/NAT propagation?

repo/config/profile.go
```diff
@@ -24,54 +33,96 @@ var defaultServerFilters = []string{
...
-			},
-		}
-		return nil
+	"badgerds": {
+		Description: `Replaces default datastore configuration with experimental
+badger datastore.
+
+If you apply this profile after ipfs init, you will need
+to convert your datastore to the new configuration.
+You can do this using ipfs-ds-convert.
```
VictorBjelkholm: Can we put a usage example of ipfs-ds-convert as well? Or maybe a link to the readme of ipfs-ds-convert is enough

repo/config/profile.go
```diff
@@ -24,54 +33,96 @@ var defaultServerFilters = []string{
...
-		return nil
+var Profiles = map[string]Profile{
+	"server": {
+		Description: `Disables local host discovery, recommended when
+running IPFS on machines with public IPv4 addresses.`,
+
+		Transform: func(c *Config) error {
+			c.Addresses.NoAnnounce = appendSingle(c.Addresses.NoAnnounce, defaultServerFilters)
+			c.Swarm.AddrFilters = appendSingle(c.Swarm.AddrFilters, defaultServerFilters)
+			c.Discovery.MDNS.Enabled = false
```
VictorBjelkholm: Why is MDNS disabled? Isn't the problem that we spam nearby ip addresses, not that we announce mdns services?

magik6k: IIRC Some providers don't like use of multicast addresses.

repo/config/profile.go
whyrusleeping: Yeah, it should

repo/config/profile.go
magik6k: Done

whyrusleeping:
Thanks @magik6k  :)

Issue #4841, opened 2018-03-20
[meta] better sharness coverage


Stebalien:
This list is a work in progress. Suggestions appreciated.

## General

For all commands, we should:

* Check that `--help` works.
* Check that bad flags are correctly detected.

## CID formatting

For all commands that have CID formatting options (add, block put, dag put, etc.), we check that:

* `--mhtype=notSha256 --cid-version=0` fails (`--cid-version=0` is sometimes written as `--format=v0`).
* `--mhtype=notSha256` defaults to CIDv1.
* `--mhtype=sha256 --cid-version=0` works.
* `--mhtype=anyHash --cid-version=1` works.

## Get directory

Check the directories:

* `ipfs get /ipfs/QmDir/`
* `ipfs get /ipfs/QmDir`
* `ipfs get /ipfs/QmDir/otherDir/`
* `ipfs get /ipfs/QmDir/otherDir`
* `ipfs get /ipfs/QmDir/otherDir/secondDir`
* `ipfs get /ipfs/QmDir/otherDir/secondDir/`

Containing:

* Nothing
* A single file
* A hidden file.
* A directory.
* A hidden directory.
* Some tree of files.

(and ensure that we actually get the right output).

## Get file

Check that the following fail:

* `ipfs get /ipfs/QmFile/`
* `ipfs get /ipfs/QmDir/QmFile/`

whyrusleeping:
Hey @chriscool, wanna help us out a bit?

chriscool:
@whyrusleeping Yeah sure I will take a look at this later today.

chriscool:
@Stebalien about all commands accepting `--help`, there is such a test already:

https://github.com/ipfs/go-ipfs/blob/master/test/sharness/t0010-basic-commands.sh#L64-L75

Stebalien:
> about all commands accepting --help, there is such a test already:

Ah, nice! However, we need to *not* pipe `/dev/null` to stdin. See #4837.

chriscool:
@Stebalien do you mean that there should be the same tests with `echo stuff | $cmd --help </dev/null >/dev/null`?

Stebalien:
Both might actually be a good idea.

kevina:
I would add:
 - `--cid-version=1` enabled raw-leaves by default when applicable

chriscool:
@Stebalien please take a look at:

https://github.com/ipfs/go-ipfs/pull/4849
[...]

PR #4848, opened 2018-03-20
sharness/t0010: check that all the commands fail when passed a bad flag

chriscool:
When any command is passed a flag it doesn't understand, like `--badflag` for example, it should just fail.

Let's check that in a sharness test.

See https://github.com/ipfs/go-ipfs/issues/4841 for more context.

### DIFF
--- a/test/sharness/t0010-basic-commands.sh
+++ b/test/sharness/t0010-basic-commands.sh
@@ -66,7 +66,7 @@ test_expect_success "All commands accept --help" '
   while read -r cmd
   do
     $cmd --help </dev/null >/dev/null ||
-      { echo $cmd doesnt accept --help; echo 1 > fail; }
+      { echo "$cmd doesnt accept --help"; echo 1 > fail; }
   done <commands.txt
 
   if [ $(cat fail) = 1 ]; then
@@ -80,7 +80,7 @@ test_expect_failure "All ipfs root commands are mentioned in base helptext" '
   while read cmd
   do
     grep "  $cmd" help.txt > /dev/null ||
-      { echo missing $cmd from helptext; echo 1 > fail; }
+      { echo "missing $cmd from helptext"; echo 1 > fail; }
   done
 
   if [ $(cat fail) = 1 ]; then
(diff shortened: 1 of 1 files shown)


test/sharness/t0010-basic-commands.sh
```diff
@@ -102,6 +102,19 @@ test_expect_failure "All ipfs commands docs are 80 columns or less" '
   fi
 '
 
+test_expect_success "All ipfs commands fail when passed a bad flag" '
+  echo 0 > fail
+  while read -r cmd
+  do
+    test_must_fail $cmd --badflag </dev/null >/dev/null ||
```
Stebalien: As with `--help` (https://github.com/ipfs/go-ipfs/issues/4841#issuecomment-374742799), we shouldn't close stdin here.

test/sharness/t0010-basic-commands.sh
```diff
@@ -102,6 +102,19 @@ test_expect_failure "All ipfs commands docs are 80 columns or less" '
...
+      { echo $cmd exit with code 0 when passed --badflag; echo 1 > fail; }
```
Stebalien: Should definitely quote this. Also, we can probably just touch fail and check to see if it exists (simpler).

test/sharness/t0010-basic-commands.sh
chriscool: OkI will change that.

test/sharness/t0010-basic-commands.sh
chriscool: About just touching fail, if we do that we then have to make sure that we remove fail in case of failure so that later tests that use the same mechanism don't fail just because of the leftover file.

chriscool: About quoting I tried to follow the same pattern as other tests. I will take a look at quoting everywhere.

chriscool:
We are not closing stdin anymore.

test/sharness/t0010-basic-commands.sh
chriscool: Done.

test/sharness/t0010-basic-commands.sh
Stebalien: Meh. Let's just leave the failure mechanism as-is.

chriscool:
I don't understand again why codecov complains.

@Kubuxu do you know what happens?

Kubuxu:
@chriscool that is normal-ish.

chriscool:
@Kubuxu ok can you merge this one too?

Kubuxu:
@chriscool I can't, currently only Why can.

whyrusleeping:
Thanks @chriscool :) Welcome back!

PR #4849, opened 2018-03-20
No dev null pipe

chriscool:
Any IPFS command should accept the `--help` flag even if /dev/null is not piped to it. Let's properly check that this works.  

See https://github.com/ipfs/go-ipfs/issues/4841 for more context.

### DIFF
--- a/test/sharness/t0010-basic-commands.sh
+++ b/test/sharness/t0010-basic-commands.sh
@@ -67,6 +67,8 @@ test_expect_success "All commands accept --help" '
   do
     $cmd --help >/dev/null ||
       { echo "$cmd doesnt accept --help"; echo 1 > fail; }
+    echo stuff | $cmd --help >/dev/null ||
+      { echo "$cmd doesnt accept --help when using stdin"; echo 1 > fail; }
   done <commands.txt
 
   if [ $(cat fail) = 1 ]; then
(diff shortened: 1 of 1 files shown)


chriscool:
The error in Jenkins is:
```
expecting success: 
  ipfs shutdown
Error: Post http://127.0.0.1:38163/api/v0/shutdown?encoding=json&stream-channels=true: EOFnot ok 14 - shutdown succeeds
#	
#	  ipfs shutdown
#	
```

chriscool:
The error in circle Ci is:

```
expecting success: 
    test_kill_repeat_10_sec $IPFS_PID
  
not ok 29 - 'ipfs daemon' can be killed
#	
#	    test_kill_repeat_10_sec $IPFS_PID
#	  
```

It doesn't look related either.

chriscool:
I don't understand what codecov complains about.

Stebalien:
Don't know. Maybe we don't run unit tests when code isn't changed?

Kubuxu:
We have some code paths it tests that randomly might not run thus resulting in lower coverage. Pushed https://github.com/ipfs/go-ipfs/pull/4867

chriscool:
@Kubuxu ok can you merge this PR?

whyrusleeping:
Needs a rebase

chriscool:
Rebased.

PR #4861, opened 2018-03-22
fix: ipfs file ls: list node size for directories

### LINKED ISSUE #4580: Reported directory size always 0
go-ipfs version: 0.4.13-
Repo version: 6
System version: amd64/linux
Golang version: go1.9.2

#### Description:
The size of directories, returned from the API server, is always `0` - instead of the sum of the directory's contents.

Example:
```sh
$ ipfs --enc=json file ls /ipfs/QmYcutogfHrrxigzVw2Vi8P1cdiuXNn9VecL9cy95wB9MP
```

Result:
```json
{
  "Arguments": {
    "/ipfs/QmYcutogfHrrxigzVw2Vi8P1cdiuXNn9VecL9cy95wB9MP": "QmYcutogfHrrxigzVw2Vi8P1cdiuXNn9VecL9cy95wB9MP"
  },
  "Objects": {
    "QmYcutogfHrrxigzVw2Vi8P1cdiuXNn9VecL9cy95wB9MP": {
      "Hash": "QmYcutogfHrrxigzVw2Vi8P1cdiuXNn9VecL9cy95wB9MP",
      "Size": 0,
      "Type": "Directory",
      "Links": [
        {
          "Name": "reindex",
          "Hash": "Qmac74sj9SgTVVVHFgWrFL6wyUQ8k8sgK4LYcEPTNoz3jW",
          "Size": 18132,
          "Type": "Directory"
        }
      ]
    }
  }
}
```

Expected size: `18132`, plus actual directory data (i.e. filenames).

schomatis:
List the size of the underlying node for directories to honor the unix API.

If a test is needed: only the JSON encoding outputs the file size and I'm not sure how to process that with the `sharness` framework.

Fixes #4580.


schomatis:
I was looking at the test file `t0250-files-api.sh` when I should have checked `t0200-unixfs-ls.sh`, I'll fix the second test.

lgierth:
It would be good to take directory sharding into account here too -- a directory is a `unixfspb.Data_HAMTShard` in these cases. We'd probably want the combined size of all shards.

Stebalien:
> It would be good to take directory sharding into account here too -- a directory is a unixfspb.Data_HAMTShard in these cases. We'd probably want the combined size of all shards.

Really, directories should *list* their sizes. Recursively reading the size of the hamt is going to be really annoying.

lgierth:
Not the size of the contents, just the raw size of the (sharded) directory object. Or do you mean that's something to precompute and stick into the object?

Stebalien:
> Not the size of the contents, just the raw size of the (sharded) directory object. 

That still involves traversing the sharded directory. Not that bad, just annoying.

> Or do you mean that's something to precompute and stick into the object?

Yeah. Ideally, *everything* would have a filesize.

whyrusleeping:
> Yeah. Ideally, everything would have a filesize.

Add that to the ipld unixfs requirements listing

Stebalien:
> Add that to the ipld unixfs requirements listing

https://github.com/ipfs/ipld-unixfs/issues/7

schomatis:
I got lost after the sharded directory comments, should I modify something of this PR?

Stebalien:
@schomatis so, we've been discussing this:

* https://github.com/ipfs/unixfs-v2/issues/7
* https://github.com/ipfs/unixfs-v2/issues/9

And I'm now less convinced we should actually include the size. Normally, filesize tells you how big a file would be when downloaded. However, the directory size (in bytes) won't really tell you anything useful. Thoughts? I don't really have any strong opinions. My initial opinion was "follow unix" but I'm now less convinced.

kevina:
> follow unix

Is not a good guide on this.  Unix uses the physical directory size (that is the list of inodes), not the size of the contents, or even the number of entries.  In some ways this makes sense and it is also easy to implement, but it isn't very useful and it doesn't really make sense for us.

schomatis:
As suggested in the previous comments, this PR doesn't seems to add much value, closing.

PR #4873, opened 2018-03-23
make the tar writer handle sharded ipfs directories

### LINKED ISSUE #4871: ipfs get doesn't work with sharded directories
It serializes files with the tar writer which doesn't currently support sharded directories.

https://github.com/ipfs/go-ipfs/blob/3cbf97bea12c5a1f6c007b094ca4df51b1747831/unixfs/archive/tar/writer.go#L83-L96

Stebalien:
Makes ipfs get work on sharded directories.

fixes #4871

### DIFF
--- a/core/commands/unixfs/ls.go
+++ b/core/commands/unixfs/ls.go
@@ -137,6 +137,10 @@ possible, please use 'ipfs ls' instead.
 			switch t {
 			case unixfspb.Data_File:
 				break
+			case unixfspb.Data_HAMTShard:
+				// We need a streaming ls API for this.
+				res.SetError(fmt.Errorf("cannot list large directories yet"), cmdkit.ErrNormal)
+				return
 			case unixfspb.Data_Directory:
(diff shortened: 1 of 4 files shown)


core/commands/unixfs/ls.go
```diff
@@ -137,6 +137,10 @@ possible, please use 'ipfs ls' instead.
 			switch t {
 			case unixfspb.Data_File:
 				break
+			case unixfspb.Data_HAMTShard:
+				// We need a streaming ls API for this.
+				res.SetError(fmt.Errorf("cannot list large directories yet"), cmdkit.ErrNormal)
+				return
```
lgierth: I think it would be okay to have this without a streaming `ls`. It's just gonna take its time to run, just like dir-listings on the gateway for sharded directories.

The streaming `ls` is an optimization (an important one nevertheless).

whyrusleeping:
I'll review it when I get back to my computer.

PR #4885, opened 2018-03-27
only visit nodes in EnumerateChildrenAsync when asked

Stebalien:
No idea why this was changed this was introduced in:

08f342e8bada4f4eb0c5462a3623f0c2b828240f (part of #3598)


whyrusleeping:
Probably worth asking @Voker57 if there was any particular reason for this.

Voker57:
Looks like a leftover from former logic, my bad.

PR #4888, opened 2018-03-28
dedup keys in GetMany

Stebalien:
Otherwise, GetMany on the children of a node with duplicate links may fail

### DIFF
--- a/merkledag/merkledag.go
+++ b/merkledag/merkledag.go
@@ -201,7 +201,20 @@ func (n *dagService) GetMany(ctx context.Context, keys []*cid.Cid) <-chan *ipld.
 	return getNodesFromBG(ctx, n.Blocks, keys)
 }
 
+func dedupKeys(keys []*cid.Cid) []*cid.Cid {
+	set := cid.NewSet()
+	for _, c := range keys {
+		set.Add(c)
+	}
+	if set.Len() == len(keys) {
+		return keys
+	}
+	return set.Keys()
+}
+
 func getNodesFromBG(ctx context.Context, bs bserv.BlockGetter, keys []*cid.Cid) <-chan *ipld.NodeOption {
+	keys = dedupKeys(keys)
+
 	out := make(chan *ipld.NodeOption, len(keys))
 	blocks := bs.GetBlocks(ctx, keys)
 	var count int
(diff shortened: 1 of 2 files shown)


kevina:
I can see this adding overhead.  Can you elaborate on what you mean by "GetMany on the children of a node with duplicate links may fail".

Stebalien:
We check to see if we get back the correct number of blocks and, if we don't, write an error to the channel. However, lower layers will (correctly) deduplicate requested keys so we'll end up getting back fewer blocks than we requested and think something has gone wrong.

kevina:
Okay, deduplicating at every layer seams like unnecessary overhead, and my thoughts are that it maybe we should just not check at this level, but I am not familiar enough with the code to know for sure.

Stebalien:
> Okay, deduplicating at every layer seams like unnecessary overhead, and my thoughts are that it maybe we should just not check at this level, but I am not familiar enough with the code to know for sure.

We could just say "make sure to deduplicate your own keys". Alternatively, we could *not* return an error when we can't find all requested objects. Really, I think we should go with the second (that's what the datastore does) but that will take some thought.

Also note, this is *definitely* not the performance bottleneck (unfortunately).

merkledag/merkledag.go
```diff
@@ -201,7 +201,20 @@ func (n *dagService) GetMany(ctx context.Context, keys []*cid.Cid) <-chan *ipld.
 	return getNodesFromBG(ctx, n.Blocks, keys)
 }
 
+func dedupKeys(keys []*cid.Cid) []*cid.Cid {
```
Kubuxu: I think this can be done without copying the whole array to map and reverse by running a search over the rest of the array and remembering which indexes are duplicate.

IDK if it would be faster.

Stebalien: The "remember which indexes are duplicate" part is exactly what we're doing when copying to the map. We could *modify* the array instead of creating a new one but then we'd be modifying an argument. We could also sort and then dedup but this is probably faster.

whyrusleeping:
Good catch, but AFAIK the only user of `GetMany` is `GetNodes` (which is now extracted). I wonder if it would be better to just replace `GetMany` with `GetNodes` completely? where did this come up?

Stebalien:
> where did this come up?

A patch I was writing to try to parallelize fetching a bit.

>  I wonder if it would be better to just replace GetMany with GetNodes completely? where did this come up?

You mean, replace it with an interface that fills in a promise array? This interface is useful when you don't care about the order (first come, first serve). Doing that with the `GetNodes` interface is tricky.

PR #4889, opened 2018-03-29
improve performance of HAMTs

Stebalien:
According to the go benchmarks (not testing the datastore, etc.), these changes
make iterating over HAMTs ~50x faster and creating them ~2x faster.

based on #4884, merge after.

### DIFF
--- a/package.json
+++ b/package.json
@@ -581,6 +581,12 @@
       "hash": "QmPdqSMmiwtQCBC515gFtMW2mP14HsfgnyQ2k5xPQVxMge",
       "name": "go-fs-lock",
       "version": "0.1.2"
+    },
+    {
+      "author": "Stebalien",
+      "hash": "QmTbBs3Y3u5F69XNJzdnnc6SP5GKgcXxCDzx6w8m6piVRT",
+      "name": "go-bitfield",
+      "version": "0.1.1"
     }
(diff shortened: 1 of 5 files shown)


Stebalien:
I've split the bitfield out into a separate package.

whyrusleeping:
TIL: go's big int implementation's `Bytes()` and `SetBytes()` method will not accurately round-trip a negative value.

whyrusleeping:
@Stebalien I wonder if you would get a performance improvement by using wordsize uint's instead of bytes, like the big int lib does.

Stebalien:
> I wonder if you would get a performance improvement by using wordsize uint's instead of bytes, like the big int lib does.

Yeah, I probably should. Originally, I was worried about the cost of converting to/from bytes and planned on doing this with an unsafe cast but that's probably unnecessary.

Stebalien:
> I wonder if you would get a performance improvement by using wordsize uint's instead of bytes, like the big int lib does.

So, it doesn't help with indexing but it does help significantly with counting bits. Unfortunately, it makes the conversion code messy as hell. I seel if I can clean it up a bit.

Kubuxu:
@Stebalien are you switching it to 64bit or are we leaving it as it is?

PR #4890, opened 2018-03-29
infer type from CID when possible in ls command

Stebalien:
We don't need to fetch the linked node when it's a raw node, we already know it's a file.


whyrusleeping:
quite the niche optimization here.

Stebalien:
> quite the niche optimization here.

Important for the ImageNet dataset (lots of small files). Really, I was going for https://github.com/ipfs/go-ipfs/issues/4891 but this is more of a drive-by fix.

PR #4894, opened 2018-03-29
improve the documentation for the p2p feature

Stebalien:


Stebalien:
> One thing to note is that when you p2p stream dial, the listening endpoint is one-shoot (unlike ssh's -L).

Are you sure? It looks like listen dials the listener for every inbound stream (although it looks like there are many obvious bugs...).

magik6k:
I meant the listening endpoint on the dialer side (the one on client side) - https://github.com/ipfs/go-ipfs/blob/master/p2p/p2p.go#L87.

The code is messy/weird there, I'll revisit it eventually to fix some of this stuff (likely when I'll get to creating coreapi for it).

Stebalien:
> I meant the listening endpoint on the dialer side (the one on client side) - https://github.com/ipfs/go-ipfs/blob/master/p2p/p2p.go#L87.

Ah... I see. Really, we should support an HTTP upgrade from a command (although supporting both makes sense).

Stebalien:
We should make it work like `ssh -L` (that's what users likely expect)`. Blocked on #4895.

whyrusleeping:
@Stebalien is this unblocked yet?

cc @magik6k

magik6k:
I've cherry-picked this in https://github.com/ipfs/go-ipfs/pull/4929 and adjusted to changes there, so if the refactor pr is going to be merged soon I'd say this can be closed.

Issue #4893, opened 2018-03-29
Configurable multiplexers


Stebalien:
It would be nice to be able to configure the supported multiplexers in the config. My thinking is to add a and a `Multiplexers` config option to the `Swarm` section. One tricky part is naming. For usability, a simple "mplex" or "yamux" would be nice however the full protocol name (with the version) may be a better option. Thoughts?

When we do this, I'd also like to get rid of the `--enable-mplex-experiment` flag as it's no longer an experiment (yeah, yeah, breaking changes... maybe keep it? Wait for https://github.com/ipfs/go-ipfs-cmds/issues/93?).

My eventual goal is to get rid of our "experimental" flags and move to config options. Ideally, the CLI should be stable (and minimal).

jackloughran:
@Stebalien sorry for spamming PRs, but i took a crack at this in #4899

Stebalien:
Thanks for the PR! Yeah, GitCop tends to harass all of our new contributors.

PR #4899, opened 2018-03-30
[WIP] multiplexer config file

jackloughran:
@Stebalien is this along the lines of what you were thinking for #4893?

the `--enable-mplex-experiment` flag still exists, but it will be ignored if there are multiplexers specified in the config file. i also made the configuration options simply `yamax` and `mplex` because it looks like the full protocol name was hard coded anyway.

i will update the documentation and command help information if this approach is the direction you want to go in.

thanks!

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -393,15 +393,16 @@ func makeSmuxTransport(multiplexers []string, mplexExp bool) smux.Transport {
 	var isYamux, isMplex bool
 
 	for _, multiplexer := range multiplexers {
-		if multiplexer == "yamux" {
+		switch multiplexer {
+		case "yamux":
 			isYamux = true
-		} else if multiplexer == "mplex" {
+		case "mplex":
 			isMplex = true
 			if !mplexExp {
 				log.Error("--enable-mplex-experiment is set to false, but mplex " +
-					"is given as a multiplexer in Swarm.Multiplexers in the config.")
+					"is given as a multiplexer the config (Swarm.Multiplexers).")
 			}
-		} else {
+		default:
 			log.Errorf("Unknown multiplexer specified in config: %s", multiplexer)
 		}
 	}
--- a/core/core.go
+++ b/core/core.go
@@ -385,6 +385,11 @@ func makeAddrsFactory(cfg config.Addresses) (p2pbhost.AddrsFactory, error) {
 func makeSmuxTransport(multiplexers []string, mplexExp bool) smux.Transport {
 	mstpt := mssmux.NewBlankTransport()
 
+	multiplexerProtocolNames := map[string]string{
+		"yamex": "/yamux/1.0.0",
+		"mplex": "/mplex/6.7.0",
+	}
+
 	var isYamex, isMplex bool
 
 	for _, multiplexer := range multiplexers {
@@ -392,6 +397,12 @@ func makeSmuxTransport(multiplexers []string, mplexExp bool) smux.Transport {
 			isYamex = true
 		} else if multiplexer == "mplex" {
 			isMplex = true
+			if mplexExp {
+				log.Warning("--enable-mplex-experiment is set to false, " +
+					"but mplex is given as a multiplexer in the config.")
+			}
+		} else {
+			log.Warningf("Unknown multiplexer specified in config: %s", multiplexer)
 		}
 	}
 
@@ -402,6 +413,22 @@ func makeSmuxTransport(multiplexers []string, mplexExp bool) smux.Transport {
 	// If there are any multiplexers defined in the config, ignore mplexExp.
 	if len(multiplexers) > 0 {
 		mplexExp = false
+
+		// set order preference to match the order specified in the config
+		multiplexerProtocols := make([]string, len(multiplexers))
+		for i, multiplexer := range multiplexers {
+			if multiplexer == "yamex" || multiplexer == "mplex" {
+				multiplexerProtocols[i] = multiplexerProtocolNames[multiplexer]
+			}
+		}
+
+		mstpt.OrderPreference = multiplexerProtocols
+	}
+
+	// Allow muxer preference order overriding.  If set, LIBP2P_MUX_PREFS will
+	// overwrite the order specified in the config.
+	if prefs := os.Getenv("LIBP2P_MUX_PREFS"); prefs != "" {
+		mstpt.OrderPreference = strings.Fields(prefs)
 	}
 
 	if isYamex {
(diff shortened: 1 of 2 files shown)


core/core.go
```diff
@@ -382,25 +382,46 @@ func makeAddrsFactory(cfg config.Addresses) (p2pbhost.AddrsFactory, error) {
...
-		LogOutput:              ioutil.Discard,
+	var isYamex, isMplex bool
+
+	for _, multiplexer := range multiplexers {
+		if multiplexer == "yamex" {
+			isYamex = true
+		} else if multiplexer == "mplex" {
+			isMplex = true
+		}
 	}
```
Stebalien: We should at least log if we find a multiplexer we don't know about.

core/core.go
```diff
@@ -382,25 +382,46 @@ func makeAddrsFactory(cfg config.Addresses) (p2pbhost.AddrsFactory, error) {
...
+
+		if os.Getenv("YAMUX_DEBUG") != "" {
+			ymxtpt.LogOutput = os.Stderr
+		}
+
+		mstpt.AddTransport("/yamux/1.0.0", ymxtpt)
+	}
 
-	if mplexExp {
+	if isMplex || mplexExp {
```
Stebalien: It would be nice to warn if `--enable-mplex-experiment` is `false` but mplex is explicitly enabled in the config.

core/core.go
```diff
@@ -382,33 +382,76 @@ func makeAddrsFactory(cfg config.Addresses) (p2pbhost.AddrsFactory, error) {
...
-	if mplexExp {
-		mstpt.AddTransport("/mplex/6.7.0", mplex.DefaultTransport)
+		mstpt.OrderPreference = multiplexerProtocols
 	}
 
-	// Allow muxer preference order overriding
+	// Allow muxer preference order overriding.  If set, LIBP2P_MUX_PREFS will
+	// overwrite the order specified in the config.
 	if prefs := os.Getenv("LIBP2P_MUX_PREFS"); prefs != "" {
 		mstpt.OrderPreference = strings.Fields(prefs)
```
Stebalien: Nit: can we add a `log.Error(...)` here warning the user that `LIBP2P_MUX_PREFS` is deprecated (telling them to use `Swarm.Multiplexers`).

core/core.go
```diff
@@ -382,33 +382,76 @@ func makeAddrsFactory(cfg config.Addresses) (p2pbhost.AddrsFactory, error) {
...
 	}
 
+	if isYamex {
```
Stebalien: This needs to go before setting the order preference (will cause problems otherwise).

core/core.go
```diff
@@ -382,33 +382,76 @@ func makeAddrsFactory(cfg config.Addresses) (p2pbhost.AddrsFactory, error) {
...
+		if multiplexer == "yamex" {
+			isYamex = true
+		} else if multiplexer == "mplex" {
+			isMplex = true
+			if mplexExp {
+				log.Warning("--enable-mplex-experiment is set to false, " +
+					"but mplex is given as a multiplexer in the config.")
+			}
+		} else {
+			log.Warningf("Unknown multiplexer specified in config: %s", multiplexer)
```
Stebalien: nit: I'd turn these warnings into errors. That way, they'll get printed to stdout on daemon start (we'd like to print warnings as well but they're too noisy at the moment).

jackloughran:
@Stebalien thanks for all the feedback.  i changed the logging to `log.Error` and i fixed the order of the statements.  it looks like i need to add some more testing to cover some of these new statements i added.  i also still have to update the documentation to reflect this new configuration option.  are there any changes you think i should make before i do that stuff?

core/core.go
```diff
@@ -382,31 +382,76 @@ func makeAddrsFactory(cfg config.Addresses) (p2pbhost.AddrsFactory, error) {
...
 
-	ymxtpt := &yamux.Transport{
-		AcceptBacklog:          512,
-		ConnectionWriteTimeout: time.Second * 10,
-		KeepAliveInterval:      time.Second * 30,
-		EnableKeepAlive:        true,
-		MaxStreamWindowSize:    uint32(1024 * 512),
-		LogOutput:              ioutil.Discard,
+	multiplexerProtocolNames := map[string]string{
+		"yamex": "/yamux/1.0.0",
```
Kubuxu: Any reason for it being called `yamex` instead of `yamux`?

jackloughran: oh wow thanks for catching that.  im not sure why i did that (i also spelled it with an a above 😬).  ill fix this right now.

Kubuxu:
Hmm, I am sure I sent out another comment with the one about yamux but it didn't go through, writing it now.

core/core.go
```diff
@@ -382,31 +382,76 @@ func makeAddrsFactory(cfg config.Addresses) (p2pbhost.AddrsFactory, error) {
...
+		"yamux": "/yamux/1.0.0",
+		"mplex": "/mplex/6.7.0",
 	}
 
-	if os.Getenv("YAMUX_DEBUG") != "" {
-		ymxtpt.LogOutput = os.Stderr
+	var isYamux, isMplex bool
+
+	for _, multiplexer := range multiplexers {
+		if multiplexer == "yamux" {
```
Kubuxu: I think it will be better to use case statement here instead of ifs.

core/core.go
```diff
@@ -382,31 +382,76 @@ func makeAddrsFactory(cfg config.Addresses) (p2pbhost.AddrsFactory, error) {
...
+		mstpt.AddTransport(multiplexerProtocolNames["yamux"], ymxtpt)
+	}
+
+	if isMplex || mplexExp {
+		mstpt.AddTransport(multiplexerProtocolNames["mplex"], mplex.DefaultTransport)
+	}
+
+	// If there are any multiplexers defined in the config, ignore mplexExp.
+	if len(multiplexers) > 0 {
+		mplexExp = false
```
Kubuxu: Is there any reason to set it to false, it doesn't seem to be used after.

core/core.go
```diff
@@ -382,31 +382,76 @@ func makeAddrsFactory(cfg config.Addresses) (p2pbhost.AddrsFactory, error) {
...
 	}
 
-	if os.Getenv("YAMUX_DEBUG") != "" {
-		ymxtpt.LogOutput = os.Stderr
+	var isYamux, isMplex bool
+
+	for _, multiplexer := range multiplexers {
+		if multiplexer == "yamux" {
+			isYamux = true
+		} else if multiplexer == "mplex" {
```
Kubuxu: @Stebalien @whyrusleeping do you think using just protocol names is good enough here or should we use full protocol descriptor. This way if we ever want to have mplex6 and mplex7 (if it ever exists) both at the same time, it will be possible.

Stebalien: I don't have much of a preference.

jackloughran: @Stebalien @Kubuxu sorry for letting this get stale. i can definitely update the pr to use full protocol descriptors if you guys think that's better.

Kubuxu: I don't think this is necessary. We can just have `mplex` and in future add others.

Issue #4891, opened 2018-03-29
Ls performance changes


Stebalien:
I'd like to make the following changes to the `ls` command for performance reasons:

1. Add a `-q` flag that only returns names. Base58 encoding a bunch of CIDs is actually quite a time-consuming operation. It would be nice to have a variant that just returns names.
2. Replace `--resolve-types` with `--type=[always|local|never]`. Currently, there's no way to say "don't resolve this type, even if we have the object". Having that option would allow us to avoid hitting the datastore for every file listed.

kevina:
Sounds Good To me.

Stebalien:
> Whoa... is this really the case?

When listing a directory with 100,000 files, 100 files per ms is actually kind of slow.

mib-kd743naq:
> When listing a directory with 100,000 files...

Given my figure of 265k/s the conversion of 100,000 CIDs *twice* the size of a typical one should take 300ms. The actual time is likely a fifth of that, as perl itself is embarrassingly slow, it was just faster to code a single command one can paste into their terminal to validate things.

All I am trying to point out is that you are almost certainly solving the wrong problem: if Base58 conversion is a notable factor in your measurements - the converter is seriously flawed.

PR #4900, opened 2018-03-30
Add --quiet and --type options to `ls` command

fyrchik:
Hello!

This is pull request for issue https://github.com/ipfs/go-ipfs/issues/4891

If --quiet is specified, output only names.
New --type option:
- when it is 'local', resolving is performed offline
- when it is 'never', no links are resolved
Non-resolved link's names are always written without trailing slash.

I am also not sure about the desired behaviour of command when both --resolve-type and --resolve options are provided.

I have not found a way to specify fixed range of string values for Option.
Is it worth to implement such possibility in somewhat generic way?

### DIFF
--- a/core/commands/ls.go
+++ b/core/commands/ls.go
@@ -38,6 +38,14 @@ type LsOutput struct {
 	Objects []LsObject
 }
 
+const (
+	resolveLocal  = "local"
+	resolveAlways = "always"
+	resolveNever  = "never"
+)
+
+var errInvalidResolveType = fmt.Errorf("type must be '%s', '%s' or '%s'", resolveAlways, resolveLocal, resolveNever)
+
 var LsCmd = &cmds.Command{
 	Helptext: cmdkit.HelpText{
 		Tagline: "List directory contents for Unix filesystem objects.",
@@ -48,6 +56,10 @@ the following format:
   <link base58 hash> <link size in bytes> <link name>
 
 The JSON output contains type information.
+
+The '--resolve' option specifies whether to include the type of each linked object.
+If 'local', types are only determined for hashes that are pinned or in your local cache.
+Note: setting it to 'always' can be VERY slow.
 `,
 	},
 
@@ -56,7 +68,9 @@ The JSON output contains type information.
 	},
 	Options: []cmdkit.Option{
 		cmdkit.BoolOption("headers", "v", "Print table headers (Hash, Size, Name)."),
-		cmdkit.BoolOption("resolve-type", "Resolve linked objects to find out their types.").WithDefault(true),
+		cmdkit.BoolOption("resolve-type", "Resolve linked objects to find out their types. Note: this option is deprecated, use --type instead."),
+		cmdkit.BoolOption(quietOptionName, "q", "Write only names."),
+		cmdkit.StringOption("resolve", "Include the type of each linked object (always|local|never).").WithDefault(resolveAlways),
 	},
 	Run: func(req cmds.Request, res cmds.Response) {
 		nd, err := req.InvocContext().GetNode()
@@ -71,14 +85,26 @@ The JSON output contains type information.
 			return
 		}
 
-		resolve, _, err := req.Option("resolve-type").Bool()
+		rtype, _, _ := req.Option("resolve").String()
+		if rtype != resolveAlways && rtype != resolveLocal && rtype != resolveNever {
+			res.SetError(errInvalidResolveType, cmdkit.ErrClient)
+			return
+		}
+		resolve, defined, err := req.Option("resolve-type").Bool()
 		if err != nil {
 			res.SetError(err, cmdkit.ErrNormal)
 			return
 		}
+		if defined {
+			if resolve {
+				rtype = resolveAlways
+			} else {
+				rtype = resolveLocal
+			}
+		}
 
 		dserv := nd.DAG
-		if !resolve {
+		if rtype == resolveLocal {
 			offlineexch := offline.Exchange(nd.Blockstore)
 			bserv := blockservice.New(nd.Blockstore, offlineexch)
 			dserv = merkledag.NewDAGService(bserv)
@@ -109,6 +135,7 @@ The JSON output contains type information.
 
 		output := make([]LsObject, len(req.Arguments()))
 
+		quiet, _, _ := req.Option(quietOptionName).Bool()
 		for i, dagnode := range dagnodes {
 			dir, err := uio.NewDirectoryFromNode(nd.DAG, dagnode)
 			if err != nil && err != uio.ErrNotADir {
@@ -140,8 +167,11 @@ The JSON output contains type information.
 					// No need to check with raw leaves
 					t = unixfspb.Data_File
 				case cid.DagProtobuf:
+					if rtype == resolveNever {
+						break
+					}
 					linkNode, err := link.GetNode(req.Context(), dserv)
-					if err == ipld.ErrNotFound && !resolve {
+					if err == ipld.ErrNotFound && rtype == resolveLocal {
 						// not an error
 						linkNode = nil
 					} else if err != nil {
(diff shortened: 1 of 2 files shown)


Stebalien:
> I have not found a way to specify fixed range of string values for Option. Is it worth to implement such possibility in somewhat generic way?

Probably not at the moment.

> I am also not sure about the desired behaviour of command when both --resolve-type and --type options are provided.

I'd return an error. Eventually, I'd like to deprecate and hide the `--resolve-type` flag (not currently possible, AFAIK).

---

Note: We'll also need to discuss this change before merging as we don't have any consensus yet (that issue was just throwing this out there as a way to fix some performance issues). *I'd* like to make this change, but I'm sure others will also have opinions.

Stebalien:
Note: an alternative would be to simply change the meaning of `--resolve-type` and say that `--resolve-type=false` is equivalent to `--type=never`.

core/commands/ls.go
```diff
@@ -71,14 +81,20 @@ The JSON output contains type information.
...
+		if rf && rtype != resolveAlways && rtype != resolveLocal && rtype != resolveNever {
+			res.SetError(errInvalidResolveType, cmdkit.ErrClient)
+			return
+		}
+
 		resolve, _, err := req.Option("resolve-type").Bool()
 		if err != nil {
 			res.SetError(err, cmdkit.ErrNormal)
 			return
 		}
```
kevina: I would change
```go
resolve, _, err := req.Option("resolve-type").Bool()
if err != nil {
	return
}
```
to
```go
{
	resolve, defined, err := req.Option("resolve-type").Bool()
	if err != nil {
		return
	}
	if rf && defined {
		res.SetError(fmt.Errorf("both 'resove-type' and 'resolve' can not be specified", cmdkit.ErrNormal))
		return
	}
	if resolve {
		rtype = resolveAlways
	} else {
		rtype = resolveNever
	}
}
```
(but not I did not test nor even try to compile this code)

This (1) makes sure both `resolve-type` and `resolve` are not both specified. (2) sets rtype to the correct value if `resolve` is specified and (3) prevents the use of the resolve variable elsewhere by putting the entire thing in its own scope.

core/commands/ls.go
```diff
@@ -57,6 +65,8 @@ The JSON output contains type information.
 	Options: []cmdkit.Option{
 		cmdkit.BoolOption("headers", "v", "Print table headers (Hash, Size, Name)."),
 		cmdkit.BoolOption("resolve-type", "Resolve linked objects to find out their types.").WithDefault(true),
+		cmdkit.BoolOption(quietOptionName, "q", "Write only names."),
+		cmdkit.StringOption("type", "Type of link resolver (always|local|never)."),
```
kevina: This should have a `WithDefault`.  From what I can tell this should be (for now) WithDefault(resolveAlways).

core/commands/ls.go
```diff
@@ -71,14 +81,20 @@ The JSON output contains type information.
 			return
 		}
 
+		rtype, rf, _ := req.Option("type").String()
+		if rf && rtype != resolveAlways && rtype != resolveLocal && rtype != resolveNever {
```
kevina: Once `WithDefault` is defined above here is no need to test for `rf`.

core/commands/ls.go
```diff
@@ -71,14 +81,20 @@ The JSON output contains type information.
...
+
 		resolve, _, err := req.Option("resolve-type").Bool()
 		if err != nil {
 			res.SetError(err, cmdkit.ErrNormal)
 			return
 		}
 
 		dserv := nd.DAG
-		if !resolve {
+		if rtype == resolveLocal || !resolve {
```
kevina: This is confusing.  I would not use resolve and just test on `rtype`.  See above comment.

Mr0grog:
Out of curiosity, are the link types documented anywhere other than in the [`unixfs` code](https://github.com/ipfs/go-ipfs/blob/master/unixfs/pb/unixfs.pb.go#L28-L52)? I feel like the values are somewhat inscrutable at first, and it’s not obvious where to go looking for their meanings. Maybe we could put some minimal information or something in the help text?

core/commands/ls.go
Mr0grog: This help text also feels a lot more confusing than the text for `resolve-type`. What about something like:

> Include the type of each linked object (always|local|never). If `local`, types are only determined for hashes that are pinned or in your local cache.

We could also add a note to the end of that indicating that setting it to always can be slow, so people know *why* they’d want to set this.

core/commands/ls.go
```diff
@@ -57,6 +65,8 @@ The JSON output contains type information.
 	Options: []cmdkit.Option{
 		cmdkit.BoolOption("headers", "v", "Print table headers (Hash, Size, Name)."),
 		cmdkit.BoolOption("resolve-type", "Resolve linked objects to find out their types.").WithDefault(true),
```
Mr0grog: Since this is now deprecated, we should probably say so (and say to use `--type` instead)!

core/commands/ls.go
fyrchik: 1) If we set default value for at least one of this options, it will become difficult to check for conflicts. For example, in your variant, if we set default value for `--resolve`, we will always have an error, once we have set `--resolve-type` (if I've understood everything right).

2) I had no default value for the new option, because i wanted to keep interface backwards compatible. If this is not a problem, may be it is better to make `--resolve-type` a string value as @Stebalien had proposed in earlier comment?

Will submit a new request soon.

kevina: @frrist that is not how default values work, the second value returned is only true if the user sets the value, if it is false than the first value returned is the default value.  All options have a default, it is wither false, the empty string, or 0, using WithDefault just makes the default explicit.

There will only be a conflict if both are specified on the command line.  If this is the case both `defined` and `rf` will be true.

kevina:
@Stebalien 

> Note: an alternative would be to simply change the meaning of --resolve-type and say that --resolve-type=false is equivalent to --type=never.

That could work but it would be an API level change as the type for the option will change from a `bool` to a `string`.

core/commands/ls.go
fyrchik: @kevina oh, then I'm wrong, thought it works the other way, will change then.

fyrchik: @kevina i've checked this again and i think second return value is true every time default value is set (both in PreRun and Run).
We invoke FillDefaults() in Parse() https://github.com/ipfs/go-ipfs-cmds/blob/master/cli/parse.go#L23
Which in turn can modify original request with default values: https://github.com/ipfs/go-ipfs-cmds/blob/master/request.go#L158

kevina: @fyrchik that does not sound right and might be a bug if it is true.  I would test this.  Also have a look at the command for the add command.

But maybe I am misunderstand something.  @whyrusleeping @Stebalien any insight?

Also @Stebalien in 4909c5a56d256e0b999d11b065227b94576609ce you removed the `WithDefault(0)` for `cidVersionOptionName`
```
-		cmdkit.IntOption(cidVersionOptionName, "Cid version. Non-zero value will change default of 'raw-leaves' to true. (experimental)").WithDefault(0),
+		cmdkit.IntOption(cidVersionOptionName, "CID version. Defaults to 0 unless an option that depends on CIDv1 is passed. (experimental)"),
```
why?  Does it relate to what @fyrchik is describing.

kevina: Also see #4911.

Kubuxu:
@kevina 
Not really, `--resolve-type` stays being bool and we have code:
```
if !resolve-type {
  type = "never"
}
```

Stebalien:
@Kubuxu @kevina actually, my point there was that we could keep the option as-is but, instead of it resolving types iff we can without hitting the network, it simply *wouldn't* resolve types *period*. As it stands, we don't even differentiate between "is a file" and "unknown type" in the UI and, in general, I can't think of a situation where I'd want to know *some* of the types but not others (usually, I either care about them or I don't).

That's why I suggested that alternative.

PR #4903, opened 2018-03-30
document ipns pubsub

### LINKED ISSUE #4902: Document IPNS over pubsub experiment
It appears that we're missing documentation for this feature in docs/experimental-features.md.

vyzo:
Closes #4902

### DIFF
--- a/docs/experimental-features.md
+++ b/docs/experimental-features.md
@@ -384,7 +384,7 @@ When it is enabled:
 - IPNS publishers push records to a name-specific pubsub topic,
   in addition to publishing to the DHT.
 - IPNS resolvers subscribe to the name-specific topic on first
-  resoltion and receive subsequently published records through pubsub
+  resolution and receive subsequently published records through pubsub
   in real time. This makes subsequent resolutions instant, as they
   are resolved through the local cache. Note that the initial
   resolution still goes through the DHT, as there is no message
--- a/docs/experimental-features.md
+++ b/docs/experimental-features.md
@@ -383,9 +383,12 @@ Utilizes pubsub for publishing ipns records in real time.
 When it is enabled:
 - IPNS publishers push records to a name-specific pubsub topic,
   in addition to publishing to the DHT.
-- IPNS resolvers subscribe to the name-specific topic and receive newly
-  published records through pubsub in real time. Note that the initial resolution
-  still goes through the DHT, as there is no message history in pubsub.
+- IPNS resolvers subscribe to the name-specific topic on first
+  resoltion and receive subsequently published records through pubsub
+  in real time. This makes subsequent resolutions instant, as they
+  are resolved through the local cache. Note that the initial
+  resolution still goes through the DHT, as there is no message
+  history in pubsub.
 
 Both the publisher and the resolver nodes need to have the feature enabled for it
 to work effectively.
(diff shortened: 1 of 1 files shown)


vyzo:
@magik6k yes, this is true; i'll add a note.

docs/experimental-features.md
```diff
@@ -365,3 +365,30 @@ ipfs config --json Experimental.ShardingEnabled true
...
+
+### In Version
+
+0.4.13
+
+### State
+
+Experimental, default-disabled.
+
+Utilizes pubsub for publishing ipns records in real time.
```
Mr0grog: Can we say slightly more about how it works/what it does? e.g:

> When you publish an IPNS record, a message about the record will also be published via pubsub (note: does it get repeatedly sent [and roughly what frequency?] or just once at publish time?). When you resolve an IPNS address, your daemon will subscribe to the pubsub topic for that name so it is immediately informed of changes.

I’m not even totally sure my understanding there is correct, so please fix if it’s not!

docs/experimental-features.md
```diff
@@ -365,3 +365,30 @@ ipfs config --json Experimental.ShardingEnabled true
...
 - [ ] Make sure that objects that don't have to be sharded aren't
 - [ ] Generalize sharding and define a new layer between IPLD and IPFS
+
+---
+
+## IPNS pubsub
+
+### In Version
+
+0.4.13
```
Stebalien: Was it in this version? I thought this only made it into 0.4.14.

docs/experimental-features.md
Stebalien: Really, the important part is: what should the user expect and why would they want to try this feature out?

Mr0grog: Yes. Exactly that ^

docs/experimental-features.md
vyzo: uhm ok, not sure either.

docs/experimental-features.md
vyzo: ok fine, I'll add _another_ sentence :)

docs/experimental-features.md
```diff
@@ -365,3 +365,38 @@ ipfs config --json Experimental.ShardingEnabled true
...
+### State
+
+Experimental, default-disabled.
+
+Utilizes pubsub for publishing ipns records in real time.
+
+When it is enabled:
+- IPNS publishers push records to a name-specific pubsub topic,
+  in addition to publishing to the DHT.
+- IPNS resolvers subscribe to the name-specific topic and receive newly
```
Stebalien: At the moment, it's a bit unclear what this actually means, I'd go with something like:

> IPNS resolvers subscribe to name-specific topics on first resolutions to speedup future resolutions of the same name.

vyzo: i'll massage it a bit.

docs/experimental-features.md
```diff
@@ -383,9 +383,12 @@ Utilizes pubsub for publishing ipns records in real time.
 When it is enabled:
 - IPNS publishers push records to a name-specific pubsub topic,
   in addition to publishing to the DHT.
-- IPNS resolvers subscribe to the name-specific topic and receive newly
-  published records through pubsub in real time. Note that the initial resolution
-  still goes through the DHT, as there is no message history in pubsub.
+- IPNS resolvers subscribe to the name-specific topic on first
+  resoltion and receive subsequently published records through pubsub
```
Mr0grog: Typo on “resol**u**tion”

vyzo: fixed, thanks.

Issue #4766, opened 2018-03-05
Bump Repo Version and Provide Migration before strict Cid code lands


kevina:
Right now https://github.com/ipfs/go-ipfs/pull/4751 is providing a stop-gap measure to avoid allowing insecure hashes.  Things won't break, but the insecure hashes will become useless.  Once the strict Cid code (https://github.com/ipfs/go-cid/pull/40) lands things will likely break badly if there are any insecure hashes.  I recommend we bump the repo version and provide a migration tool that just checks and aborts in there are any CID's with insecure hashes.

I also suggest, though not require, we provide a tool to automatically convert the insure hashes to a secure hash so that data is not lost.  The same tool can just offer to delete them instead.

whyrusleeping:
This is a good idea, a migration is likely the best way forward.

kevina:
@whyrusleeping I would like to go ahead and write this.  I am thinking if the migration tool finds invalid Cid's and stores them in a file in the repo (and possible list then to stdout if there are not too many).  (I want to store the list somewhere because it is very easy to lose the output of the migration command.) 
It then instructs the user to remove them somehow.  It will suggest command to remove them with the previous IPFS version but warn of the obvious data lose.  If that is not adequate I would like to suggest the user contact us somehow because I want to know about cases where insure hashes are already used.  I can then help the user on a case by case bases.  If necessary I can write a tool that could convert the Cid's to a secure hash.

kevina:
If we go ahead with this repo migration I would like to combine this migration with one that will remove all id-hashes from the datastore to make https://github.com/ipfs/go-ipfs/pull/4910 easier.

kevina:
cc @Stebalien @Kubuxu

kevina:
@Stebalien @whyrusleeping told me you are also working on some migration code, what the status of that?  I would like to go ahead and write the code that checks for invalid hashes and also removed identity hashes but don't want to clash with what you are doing.

Stebalien:
This is blocked on #5007. That should be done and ready to merge (once I get a final sign off). That should finally unblock literally everything.

Stebalien:
> If we go ahead with this repo migration I would like to combine this migration with one that will remove all id-hashes from the datastore to make #4910 easier.

If you do that, we should probably land #4910 at the same time. I'd rather not temporarily break ID hashed content (although..., in practice we'll break absolutely nothing because nobody's using them..., maybe not an issue).

kevina:
[...]

PR #4910, opened 2018-04-03
Add initial support for identity hashes.

### LINKED ISSUE #4697: identity hash
#### Version information:
go-ipfs version: 0.4.13-
Repo version: 6
System version: amd64/windows
Golang version: go1.9.2

#### Description:
With identity hash we can inline data in link. Useful for short or zero length data.

> [Trivial hash function](https://en.wikipedia.org/wiki/Hash_function#Trivial_hash_function)
> 
> If the data to be hashed is small enough, one can use the data itself (reinterpreted as an integer) as the hashed value. The cost of computing this "trivial" (identity) hash function is effectively zero. This hash function is perfect, as it maps each input to a distinct hash value.
> 
> The meaning of "small enough" depends on the size of the type that is used as the hashed value. For example, in Java, the hash code is a 32-bit integer. Thus the 32-bit integer Integer and 32-bit floating-point Float objects can simply use the value directly; whereas the 64-bit integer Long and 64-bit floating-point Double cannot use this method.
> 
> Other types of data can also use this perfect hashing scheme. For example, when mapping character strings between upper and lower case, one can use the binary encoding of each character, interpreted as an integer, to index a table that gives the alternative form of that character ("A" for "a", "8" for "8", etc.). If each character is stored in 8 bits (as in extended ASCII[7] or ISO Latin 1), the table has only 28 = 256 entries; in the case of Unicode characters, the table would have 17×216 = 1114112 entries.
> 
> The same technique can be used to map two-letter country codes like "us" or "za" to country names (262 = 676 table entries), 5-digit zip codes like 13083 to city names (100000 entries), etc. Invalid data values (such as the country code "xx" or the zip code 00000) may be left undefined in the table or mapped to some appropriate "null" value.

#### Additional information:
z3vosKrgsh -> 01 55 00 03 00 00 00 is valid 3 bytes length "identity" hash of raw block with same length and data (00 00 00)

```
>ipfs block get z3vosKrgsh
Error: blockservice: key not found

>ipfs get z3vosKrgsh
Error: merkledag: not found
```

kevina:
Still a work in progress.

Tested by hand.

Needs shareness tests.

Closes #4697

### DIFF
--- a/core/builder.go
+++ b/core/builder.go
@@ -17,6 +17,7 @@ import (
 	pin "github.com/ipfs/go-ipfs/pin"
 	repo "github.com/ipfs/go-ipfs/repo"
 	cfg "github.com/ipfs/go-ipfs/repo/config"
+	"github.com/ipfs/go-ipfs/thirdparty/idstore"
 	"github.com/ipfs/go-ipfs/thirdparty/verifbs"
 	uio "github.com/ipfs/go-ipfs/unixfs/io"
 
@@ -189,14 +190,16 @@ func setupNode(ctx context.Context, n *IpfsNode, cfg *BuildCfg) error {
 		opts.HasBloomFilterSize = 0
 	}
 
-	cbs, err := bstore.CachedBlockstore(ctx, bs, opts)
+	wbs, err := bstore.CachedBlockstore(ctx, bs, opts)
 	if err != nil {
 		return err
 	}
 
-	n.BaseBlocks = cbs
+	wbs = idstore.IdStore(wbs)
+
+	n.BaseBlocks = wbs
 	n.GCLocker = bstore.NewGCLocker()
-	n.Blockstore = bstore.NewGCBlockstore(cbs, n.GCLocker)
+	n.Blockstore = bstore.NewGCBlockstore(wbs, n.GCLocker)
 
 	if conf.Experimental.FilestoreEnabled {
 		// hash security
--- a/core/commands/add.go
+++ b/core/commands/add.go
@@ -15,6 +15,7 @@ import (
 	dagtest "github.com/ipfs/go-ipfs/merkledag/test"
 	mfs "github.com/ipfs/go-ipfs/mfs"
 	ft "github.com/ipfs/go-ipfs/unixfs"
+	cide "github.com/ipfs/go-ipfs/thirdparty/cidextra"
 
 	mh "gx/ipfs/QmZyZDi491cCNTLfAhwcaDii2Kg4pwKRkhqQzURGDvY6ua/go-multihash"
 	bstore "gx/ipfs/QmaG4DZ4JaqEfvPWt5nPPgoTzhc1tr1T3f4Nu9Jpdm8ymY/go-ipfs-blockstore"
@@ -279,13 +280,13 @@ You can now check what blocks have been created by:
 		fileAdder.Silent = silent
 		fileAdder.RawLeaves = rawblks
 		fileAdder.NoCopy = nocopy
-		fileAdder.Prefix = &prefix
+		fileAdder.CidOpts = &cide.Opts{Prefix: prefix}
 
 		if hash {
 			md := dagtest.Mock()
 			emptyDirNode := ft.EmptyDirNode()
 			// Use the same prefix for the "empty" MFS root as for the file adder.
-			emptyDirNode.Prefix = *fileAdder.Prefix
+			emptyDirNode.CidOpts = *fileAdder.CidOpts
 			mr, err := mfs.NewRoot(req.Context, md, emptyDirNode, nil)
 			if err != nil {
 				res.SetError(err, cmdkit.ErrNormal)
--- a/core/commands/add.go
+++ b/core/commands/add.go
@@ -45,6 +45,7 @@ const (
 	cidVersionOptionName  = "cid-version"
 	hashOptionName        = "hash"
 	idHashLimitOptionName = "id-hash-limit"
+	inlineOptionName      = "inline"
 )
 
 const adderOutChanSize = 8
@@ -121,7 +122,8 @@ You can now check what blocks have been created by:
 		cmdkit.BoolOption(fstoreCacheOptionName, "Check the filestore for pre-existing blocks. (experimental)"),
 		cmdkit.IntOption(cidVersionOptionName, "CID version. Defaults to 0 unless an option that depends on CIDv1 is passed. (experimental)"),
 		cmdkit.StringOption(hashOptionName, "Hash function to use. Implies CIDv1 if not sha2-256. (experimental)").WithDefault("sha2-256"),
-		cmdkit.IntOption(idHashLimitOptionName, "Id hash maxium size. -1 disables. (experimental)").WithDefault(-1),
+		cmdkit.BoolOption(inlineOptionName, "Inline small objects using identity hash. (experimental)"),
+		cmdkit.IntOption(idHashLimitOptionName, "Identity hash maxium size. (experimental)").WithDefault(64),
 	},
 	PreRun: func(req *cmds.Request, env cmds.Environment) error {
 		quiet, _ := req.Options[quietOptionName].(bool)
@@ -175,6 +177,7 @@ You can now check what blocks have been created by:
 		fscache, _ := req.Options[fstoreCacheOptionName].(bool)
 		cidVer, cidVerSet := req.Options[cidVersionOptionName].(int)
 		hashFunStr, _ := req.Options[hashOptionName].(string)
+		inline, _ := req.Options[inlineOptionName].(bool)
 		idHashLimit, _ := req.Options[idHashLimitOptionName].(int)
 
 		// The arguments are subject to the following constraints.
@@ -285,10 +288,12 @@ You can now check what blocks have been created by:
 		fileAdder.NoCopy = nocopy
 		fileAdder.CidOpts = &cide.Opts{Prefix: prefix}
 
-		err = fileAdder.CidOpts.SetIdHashLimit(idHashLimit)
-		if err != nil {
-			res.SetError(err, cmdkit.ErrNormal)
-			return
+		if inline {
+			err = fileAdder.CidOpts.SetIdHashLimit(idHashLimit)
+			if err != nil {
+				res.SetError(err, cmdkit.ErrNormal)
+				return
+			}
 		}
 
 		if hash {
--- a/core/commands/add.go
+++ b/core/commands/add.go
@@ -14,8 +14,8 @@ import (
 	dag "github.com/ipfs/go-ipfs/merkledag"
 	dagtest "github.com/ipfs/go-ipfs/merkledag/test"
 	mfs "github.com/ipfs/go-ipfs/mfs"
-	ft "github.com/ipfs/go-ipfs/unixfs"
 	cide "github.com/ipfs/go-ipfs/thirdparty/cidextra"
+	ft "github.com/ipfs/go-ipfs/unixfs"
 
 	mh "gx/ipfs/QmZyZDi491cCNTLfAhwcaDii2Kg4pwKRkhqQzURGDvY6ua/go-multihash"
 	bstore "gx/ipfs/QmaG4DZ4JaqEfvPWt5nPPgoTzhc1tr1T3f4Nu9Jpdm8ymY/go-ipfs-blockstore"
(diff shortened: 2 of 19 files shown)


core/builder.go
```diff
@@ -189,14 +190,16 @@ func setupNode(ctx context.Context, n *IpfsNode, cfg *BuildCfg) error {
 		opts.HasBloomFilterSize = 0
 	}
 
-	cbs, err := bstore.CachedBlockstore(ctx, bs, opts)
+	wbs, err := bstore.CachedBlockstore(ctx, bs, opts)
```
magik6k: What does `w` in `wbs` stand for?

thirdparty/idstore/idstore.go
```diff
@@ -0,0 +1,85 @@
...
+func (b *idstore) Put(bl blocks.Block) error {
+	isId, _ := extractContents(bl.Cid())
+	if isId {
+		return nil
+	}
+	return b.bs.Put(bl)
+}
+
+func (b *idstore) PutMany(bs []blocks.Block) error {
+	var good []blocks.Block
```
magik6k: I'd preallocate this to be the size of `len(bs)` as in most cases most blocks will be in this array. I'd also call it `toPut` or something like that, `good` sounds weird to me in this context

core/builder.go
kevina: "wrapped"

thirdparty/idstore/idstore.go
```diff
@@ -2,59 +2,78 @@ package idstore
...
 
+var MaxIdHashLen = 64
+
 // idstore wraps a BlockStore to add support for identity hashes
 type idstore struct {
 	bs bls.Blockstore
 }
 
-func IdStore(bs bls.Blockstore) bls.Blockstore {
+func IdStore(bs bls.Blockstore) *idstore {
```
Kubuxu: Any reason for exporting the implementation structure instead of the interface? It is generally better to export the interface as then it is clear what contract it follows.

kevina: No.  Changed back.

kevina:
Note: `ipfs add --hash=id` is problematic due the implicit creation of directory entry:
```
$ ipfs add --hash=id 64bytes
zB7EoQcn4hmgrzo7SKs4Dn9JxnqRnzMqopdjakoog5EQPgziUnDKQrKCFEkytjoN8TgZnXGt8ndvMwkmzQEXJxq8bgDkQ
Error: identity hash of 87 bytes longer than maximum size of 64 bytes
```
The error is for the wrapping directory entry, not the hash for the 64byte file.  This works:
```
$ ipfs get zB7EoQcn4hmgrzo7SKs4Dn9JxnqRnzMqopdjakoog5EQPgziUnDKQrKCFEkytjoN8TgZnXGt8ndvMwkmzQEXJxq8bgDkQ > out
```

kevina:
@Stebalien @whyrusleeping since id-hashes are primary useful as an optimization to inline tiny files I want ahead and added an option to add.  This change touched a lot of files since it involved changing passing around the cid `Prefix` with a new `CidOpts` that included an option to switch to use the id-hash if the contents was less than a certain size.  Then instead of using `Prefix.sum()` `CidOpts.Sum()` was used.

The add option now has a `--id-hash-limit` option to control when the id hash is used, if you like it I can add it to `ipfs files` command also.

If you are okay with the `Prefix` to `CidOpts` part I would like to get that in ASAP to avoid conflicts, or worse, before some of that code gets factored out into a separate package.

core/commands/add.go
```diff
@@ -14,6 +14,7 @@ import (
 	dag "github.com/ipfs/go-ipfs/merkledag"
 	dagtest "github.com/ipfs/go-ipfs/merkledag/test"
 	mfs "github.com/ipfs/go-ipfs/mfs"
+	cide "github.com/ipfs/go-ipfs/thirdparty/cidextra"
```
Stebalien: I assume this is temporary.

core/commands/add.go
```diff
@@ -119,6 +121,7 @@ You can now check what blocks have been created by:
 		cmdkit.BoolOption(fstoreCacheOptionName, "Check the filestore for pre-existing blocks. (experimental)"),
 		cmdkit.IntOption(cidVersionOptionName, "CID version. Defaults to 0 unless an option that depends on CIDv1 is passed. (experimental)"),
 		cmdkit.StringOption(hashOptionName, "Hash function to use. Implies CIDv1 if not sha2-256. (experimental)").WithDefault("sha2-256"),
+		cmdkit.IntOption(idHashLimitOptionName, "Id hash maxium size. -1 disables. (experimental)").WithDefault(-1),
```
Stebalien: Personally, I wouldn't make this this configurable. I'd have a fixed option and set the size to 64 bytes. I doubt there are any valid reasons for picking a non-standard value here (other than, maybe choosing between 32 bytes and 64 bytes). I *would not* allow more than 64 bytes as CIDs are supposed to be small.

thirdparty/cidextra/cidextra.go
```diff
@@ -0,0 +1,37 @@
...
+
+	mh "gx/ipfs/QmZyZDi491cCNTLfAhwcaDii2Kg4pwKRkhqQzURGDvY6ua/go-multihash"
+	cid "gx/ipfs/QmcZfnkapfECQGcLZaf9B79NRg7cRa9EnZh4LSbkCzwNvY/go-cid"
+)
+
+var MaxIdHashLen = 64
+
+type Opts struct {
+	cid.Prefix
+	idHashThres int // 0 disables, otherwise 1 + limit
```
Stebalien: Why the +1? Zero initialize?

thirdparty/idstore/idstore.go
```diff
@@ -0,0 +1,108 @@
...
+// idstore wraps a BlockStore to add support for identity hashes
+type idstore struct {
+	bs bls.Blockstore
+}
+
+func IdStore(bs bls.Blockstore) bls.Blockstore {
+	return &idstore{bs}
+}
+
+func extractContents(k *cid.Cid) (bool, []byte, error) {
```
Stebalien: We'll probably want to move this (and maybe a few other things) to, e.g., `go-block`. We'll want this stuff usable outside of go-ipfs.

kevina:
> Does this only apply to raw blocks? Ideally, it would apply to all IPLD objects.

No, it also applies to protobuf objects.

core/commands/add.go
kevina: More or less, I would like to get something working first, merge this p.r. and then worry about finding a better place for the two new packages in thirdparty.

thirdparty/idstore/idstore.go
kevina: See my earlier comment, I want to first worry about getting something worry and then we can move the my two new thirdparty packages to a better location.

core/commands/add.go
kevina: Okay I will add an "--inline" option and keep this option, but set the default to 64.

Note there is an internal check that prevents this value from being set to the maximum id-hash size which is currently 64.

Kubuxu:
We are cutting down `thirdparty` packages. The CID security was a huge exception from that because we needed to have a possible security hole fixed before the release and libp2p dependency tree was causing problems.

I need to object introduction of code to `thirdparty`.

kevina:
@Stebalien @Kubuxu the intent of both `thirdparty` packages where for them to be short lived.  Both of them I intent to merge in another package outside of `go-ipfs`, but I thought it would be easier to both code and review if all the relevent changes are in one p.r. rather than spread out over several.  If you will not accept or even review this p.r. since it introduces thirdparty packages please let me know and I will first get those changes into an appropriate external package.

mib-kd743naq:
@kevina in the meantime should it be possible for me to checkout f0a4f297fc and build it locally to verify whether the web gateway will read my 3rd-party blocks with identity data mixed in?

Basically asking whether I should expect that specific commit to *just work* as far as reading identity-containing blocks is concerned?

kevina:
@mib-kd743naq yes it should be.  Before this commit id-nodes where actually supported, they just where not optimized.

kevina:
@Kubuxu @whyrusleeping @Stebalien since id hashes are possible to use without this p.r. there is the very real possibility that blockes with an id-hash could be stored in the datastore and never removed.  I made the `Delete` always pass though the request so the g.c. should in theory remove them if they are not pinned, but this needs testing.  For simplicity I propose we add a config option `DisableIdHashOpt` that is intended to only be used for testing, I can then use this config option to write a sharness test that inject an id hash and then enable the option and see if the GC can remove it.  The alternative is to write a unit test that somehow bypasses the id-store and then calls the GC to see if it is somehow removed, this will likely be more work and I imagine that the config option would be useful in other contexts.

magik6k:
Why not just add that option on the idstore struct and write a small go test for it?

kevina:
@magik6k I could do that, but I think I a better way is to just remove all id hashes from the datastore as part of a migration, see #4766.

mib-kd743naq:
> @kevina in the meantime should it be possible for me to checkout f0a4f29 and build it locally to verify whether the web gateway will read my 3rd-party blocks with identity data mixed in?

> @mib-kd743naq yes it should be. Before this commit id-hashes where actually supported, they just where not optimized.

Great stuff! You are indeed correct that identity hashes already work. E.g. this is a single block:
https://ipfs.io/ipfs/zdjA8o9D3n7NhdXGP5QZRftEyATq8pJamtFVQhQN4VHriUgR8

There is a bit of work to be done around the CLI abort however. Observe:
```
~/inlines$ ipfs ls zdjA8e3nEZNdas7TMDHEJuQB9r1RDQt9NstfH8AtYmetNzQCd
zdjA8dEswZqPMYnazeqZfj679FAH5dSti2R4zMq54AxHXT4k7 122 IllegalOverlargeInline/
zdjA8o9D3n7NhdXGP5QZRftEyATq8pJamtFVQhQN4VHriUgR8 32  LegitInlines/
```
```
~/inlines$ ipfs get zdjA8e3nEZNdas7TMDHEJuQB9r1RDQt9NstfH8AtYmetNzQCd/LegitInlines; echo "ExitCode: $?"
Saving file(s) to LegitInlines
 122 B / 122 B [=========================================================================================================================================================================================] 100.00% 0s
ExitCode: 0
```
```
~/inlines$ ipfs get zdjA8e3nEZNdas7TMDHEJuQB9r1RDQt9NstfH8AtYmetNzQCd/IllegalOverlargeInline; echo "ExitCode: $?"
Saving file(s) to IllegalOverlargeInline
 279 B / 279 B [=========================================================================================================================================================================================] 100.00% 0s
ExitCode: 0
```
```
~/inlines$ ls -la
total 0
drwxr-xr-x  3 ipfs ipfs  26 Apr 25 12:04 .
drwxr-xr-x 18 ipfs ipfs 435 Apr 25 12:03 ..
drwxr-xr-x  2 ipfs ipfs  50 Apr 25 12:04 LegitInlines
```

The `ipfs get` did not exit with non-zero, yet it ( correctly ) did not create a local copy due to an invalid 122byte inline.

Thoughts?

kevina:
@Stebalien @Kubuxu 

> We'll probably want to move this (and maybe a few other things) to, e.g., `go-block`. We'll want this stuff usable outside of go-ipfs.

Go blocks doesn't exist.  I will likely combine the two `thirdparty` into a single package `go-ipfs-idstore`.

Stebalien:
@kevina sorry, I meant `go-block-format`. However, a separate package is fine.

kevina:
@Stebalien I voted against a separate package and instead added the id store to `go-ipfs-blockstore`.  The parts needed for supported id hashes automatically I think best belong in `go-cid` although I can be convinced otherwise.  I am not sure either part belong in `go-block-format`.

Stebalien:
Fair enough (given that I've reviewed that, I should have remembered this...).

> The parts needed for supported id hashes automatically I think best belong in go-cid although I can be convinced otherwise.

A `ExtractInlineBlock(cid) Block` function would have to go in `go-block-format` but a simple `ExtractInlineData(cid) []byte` could go in `go-cid`. Up to you.

kevina:
I didn't think about creating those functions because they are not really needed (but can be useful).  What I think should go in `go-cid` is a renamed `Opts` (see https://github.com/ipfs/go-ipfs/blob/f0a4f297fc095b14aef58ded32864e7fe89f3974/thirdparty/cidextra/cidextra.go) to automatically create inline id when below the threshold. This relates to by comments in https://github.com/ipfs/go-cid/issues/23#issuecomment-394515565.  I am now think that maybe this should be called `Format`.

Stebalien:
Ah. Yeah, good point. And making this a part of `Format` or whatever we end up calling that removes my one reservation with this patch.

kevina:
Closing in favor of #5117.

PR #4919, opened 2018-04-04
Fix wrong added filename

camelmasa:
I think It looks `files.NewSerialFile`'s first argument is wrong.

Note: The `coreunix.AddR` will be removed after implemented Core API (Ref: https://github.com/ipfs/go-ipfs/pull/4562).


Kubuxu:
I am wondering why no tests have detected it.

camelmasa:
@Kubuxu Thanks for checking.

I just changed test data directory path from `test_data` to `test/data`.

Case of original code,  Returned value was changed after I changed just the test data directory path.
Original test case was couldn't detect error when the directory was nested.

core/coreunix/add.go
```diff
@@ -336,7 +337,10 @@ func AddR(n *core.IpfsNode, root string) (key string, err error) {
 		return "", err
 	}
 
-	f, err := files.NewSerialFile(root, root, false, stat)
+	segs := strings.Split(root, "/")
+	name := segs[len(segs)-1]
```
Kubuxu: use `name := filepath.Base(root)` instead of string splitting.
Filepath is aware of system specific separators.

magik6k: This probably broke `go test` on windows - https://ci.ipfs.team/blue/organizations/jenkins/IPFS%2Fgo-ipfs/detail/PR-4919/4/pipeline

whyrusleeping:
Since the author has been gone for ~1mo, can someone else take over this changeset? cc @djdv maybe?

djdv:
Amended in: https://github.com/ipfs/go-ipfs/pull/5167

Issue #4895, opened 2018-03-29
Fixup the p2p command


Stebalien:
This command looks like it could use some serious cleanup.

1. Listeners have `Running` flags that are (1) never set to false, and (2) not atomic (so setting them to false would be useless anyways).
2. The registry has no synchronization and is pretty inefficient.

And probably more. Any takers?

Stebalien:
More: Make it actually work like `ssh -L`. It should forward *every* connection, not just the first. See https://github.com/ipfs/go-ipfs/pull/4894#issuecomment-377287053.

Blocked on #4894.

Issue #3994, opened 2017-06-19
`ipfs p2p` feedback thread


whyrusleeping:
For everyone trying out the `ipfs p2p` command, please add any and all feedback here, this includes (but is not limited to) bugs, opinions, interface critiques, potential usecases, and feature requests.

magik6k:
What I noticed missing from one of the earlier PRs is the ability to tell to which peer connected to a listener. Earlier PRs had a switch that caused ipfs to send `<PeerID>\n` at the beginning of each incoming stream. This should be pretty simple to implement.

ReisenB:
Is this planned to be accessible through the browser with js-ipfs-api?

whyrusleeping:
Unfortunately no, you could set up the listeners from the browser, but you
couldn't actually connect to it since you can't dial TCP connections there.
Unless... @Magik6k could you specify a /ws multiaddr for the listeners?

magik6k:
I'd need to look into how websockets are currently handled in go-ipfs. For dialing this should be relatively simple to do, though listen mode would probably have to use some sort of stream muxer that has js implementation. This should fit quite well into the existing code.

jdgcs:
No opening port 10101, is that right?

`liu@nas:~/ipfs/go-ipfs % ./ipfs p2p listener open p2p-test /ip4/127.0.0.1/tcp/10101
{"Address":"/ip4/127.0.0.1/tcp/10101","Protocol":"/p2p/p2p-test"}
liu@nas:~/ipfs/go-ipfs % ./ipfs p2p listener ls
/ip4/127.0.0.1/tcp/10101 /p2p/p2p-test
liu@nas:~/ipfs/go-ipfs % netstat -an | grep 10101
liu@nas:~/ipfs/go-ipfs %`

magik6k:
This creates listening libp2p handler and when p2p connection is incoming it(ipfs daemon) then opens a tcp connection to a given address and proxies the data. This seems counterintutive at first, but it allows for transparent connection proxying over libp2p

origama:
It would be nice to be able to set the listeners in the config file, so that the daemon could create them at booting time. It makes sense especially if you run IPFS as an init/systemd/docker unit.

magik6k:
I don't think using config for that is a good idea. What may work is extracting whole `ipfs p2p` into a separate program as it doesn't really fit ipfs (imo), this program could then talk to ipfs daemon and reuse its libp2p instance or be standalone.

[...]

PR #4929, opened 2018-04-10
Refactor `ipfs p2p`

magik6k:
Fixes https://github.com/ipfs/go-ipfs/issues/4895.

TODO:
* [x] Reorganize the code
* [x] Fix/Cleanup registry logic
* [x] Make dial work more like `ssh -L` 
* [x] Update docs
* [ ] Review / discuss ideas


core/commands/p2p.go
```diff
@@ -53,49 +56,115 @@ are refined`,
...
 
-// p2pListenerCmd is the 'ipfs p2p listener' command
-var p2pListenerCmd = &cmds.Command{
+var p2pForwardCmd = &cmds.Command{
 	Helptext: cmdkit.HelpText{
-		Tagline:          "P2P listener management.",
-		ShortDescription: "Create and manage listener p2p endpoints",
+		Tagline: "Forward connections to or from libp2p services",
+		ShortDescription: `
+Forward connections to <listen-address> to <target-address>. Protocol specifies
```
whyrusleeping: is there a 'from' missing in this sentence?

docs/experimental-features.md
```diff
@@ -218,36 +218,86 @@ configured, the daemon will fail to start.
...
+going to use `/kickass/1.0`.
+
+**Setup:**
+
+1. A "server" node with peer ID `$SERVER_ID`
+2. A "client" node.
+
+**On the "server" node:**
+
+First, start your application and have it listen on `$APP_PORT`.
```
whyrusleeping: maybe clarify that this is a tcp socket being listened on, such as an ssh or http server.

docs/experimental-features.md
```diff
@@ -218,36 +218,86 @@ configured, the daemon will fail to start.
...
+
+Then, configure the p2p listener by running:
+
+```sh
+> ipfs p2p forward /kickass/1.0 /ipfs /ip4/127.0.0.1/tcp/$APP_PORT
```
whyrusleeping: whats the /ipfs in the middle for?

p2p/local.go
```diff
@@ -0,0 +1,124 @@
...
+)
+
+// localListener manet streams and proxies them to libp2p services
+type localListener struct {
+	ctx context.Context
+
+	p2p *P2P
+	id  peer.ID
+
+	proto string
```
whyrusleeping: maybe make this the `protocol.ID` type?

p2p/listener.go
```diff
@@ -0,0 +1,65 @@
...
+type listenerKey struct {
+	proto  string
+	listen string
+	target string
+}
+
+// ListenerRegistry is a collection of local application proto listeners.
+type ListenerRegistry struct {
+	Listeners map[listenerKey]Listener
+	lk        *sync.Mutex
```
whyrusleeping: why a pointer to a mutex?

core/commands/p2p.go
```diff
@@ -53,49 +56,115 @@ are refined`,
...
+the libp2p protocol to use.
+
+To create libp2p service listener, specify '/ipfs' as <listen-address>
```
djdv: To create **a** libp2p service listener

core/commands/p2p.go
```diff
@@ -53,49 +56,115 @@ are refined`,
...
+		if err != nil {
+			res.SetError(err, cmdkit.ErrNormal)
+			return
+		}
 
-	Subcommands: map[string]*cmds.Command{
-		"ls":    p2pListenerLsCmd,
-		"open":  p2pListenerListenCmd,
-		"close": p2pListenerCloseCmd,
+		//TODO: Do we really want/need implicit prefix?
```
djdv: Thoughts on this?
Would it cause any restrictions or confusion, with or without them?

docs/experimental-features.md
```diff
@@ -218,36 +218,86 @@ configured, the daemon will fail to start.
...
+
+This will configure IPFS to forward all incoming `/p2p/kickass/1.0` streams to
+`127.0.0.1:$APP_PORT` (opening a new connection to `127.0.0.1:$APP_PORT` per
+incoming stream.
+
+**On the "client" node:**
+
+First, configure the p2p dialer to forward all inbound connections on
+`127.0.0.1:SOME_PORT` to the listener behind `/p2p/kickass/1.0` on the server
+node.
```
djdv: >First, configure the p2p dialer to forward all inbound connections on `127.0.0.1:SOME_PORT`  to the listener behind `/p2p/kickass/1.0` on the server node.

It may be fine as is, but I wonder if this is more clear.

>First, configure the client p2p dialer, so that it forwards all inbound connections on `127.0.0.1:SOME_PORT` to the server node listening on `/p2p/kickass/1.0`.

core/commands/p2p.go
```diff
@@ -53,49 +56,115 @@ are refined`,
...
+    - Forward connections to 'myproto' libp2p service to 127.0.0.1:1234
+
+  ipfs p2p forward myproto /ip4/127.0.0.1/tcp/4567 /ipfs/QmPeer
+    - Forward connections to 127.0.0.1:4567 to 'myproto' service on /ipfs/QmPeer
+
+`,
 	},
+	Arguments: []cmdkit.Argument{
+		cmdkit.StringArg("protocol", true, false, "Protocol identifier."),
+		cmdkit.StringArg("listen-address", true, false, "Listening endpoint"),
```
djdv: Missing full stop
>"Listening endpoint."

core/commands/p2p.go
```diff
@@ -138,207 +208,156 @@ var p2pListenerLsCmd = &cmds.Command{
...
-can transparently connect to a p2p service.
-		`,
+		Tagline: "List active p2p streams.",
 	},
-	Arguments: []cmdkit.Argument{
-		cmdkit.StringArg("Peer", true, false, "Remote peer to connect to"),
-		cmdkit.StringArg("Protocol", true, false, "Protocol identifier."),
-		cmdkit.StringArg("BindAddress", false, false, "Address to listen for connection/s (default: /ip4/127.0.0.1/tcp/0)."),
+	Options: []cmdkit.Option{
+		cmdkit.BoolOption("headers", "v", "Print table headers (HagndlerID, Protocol, Local, Remote)."),
```
djdv: HagndlerID -> HandlerID

core/commands/ping.go
```diff
@@ -182,9 +182,11 @@ func pingPeer(ctx context.Context, n *core.IpfsNode, pid peer.ID, numPings int)
 }
 
 func ParsePeerParam(text string) (ma.Multiaddr, peer.ID, error) {
-	// to be replaced with just multiaddr parsing, once ptp is a multiaddr protocol
-	idx := strings.LastIndex(text, "/")
+	idx := strings.LastIndex(text, "/ipfs/")
```
djdv: There's a few instances of the literal string `"/ipfs/"` in the package, would it be better to declare it as a named const?
e.g. `peerPrefix` or something

p2p/remote.go
```diff
@@ -0,0 +1,87 @@
...
+	net "gx/ipfs/QmXfkENeeBvh3zYA51MaSdGUdBjhQ99cP5WQe8zgr6wchG/go-libp2p-net"
+	protocol "gx/ipfs/QmZNkThpqfVXs9GNbexPrfBbXSLNYeKrE7jwFM2oqHbyqN/go-libp2p-protocol"
+)
+
+// remoteListener accepts libp2p streams and proxies them to a manet host
+type remoteListener struct {
+	p2p *P2P
+
+	// Application proto identifier.
+	proto string
```
djdv: If `protocol.ID` type is used in local, change must be made here too

p2p/local.go
```diff
@@ -0,0 +1,124 @@
...
+
+	err := l.p2p.peerHost.Connect(ctx, pstore.PeerInfo{ID: l.peer})
+	if err != nil {
+		return nil, err
+	}
+
+	return l.p2p.peerHost.NewStream(l.ctx, l.peer, protocol.ID(l.proto))
+}
+
+func (l *localListener) acceptConns() {
```
djdv: Not in this PR, but there's opportunity for logging errors inside of here. Maybe later if a p2p log-subsystem is added.

core/commands/ping.go
magik6k: Ideally we'd parse those as a multiaddr, but for that we'd need to define new multiaddr proto for that.

whyrusleeping:
@magik6k It would be great if you could write up a separate document/tutorial on using `ipfs p2p` to tunnel ssh connections. I think a specific actionable example would really make the possibilities here click for a lot of people

docs/experimental-features.md
magik6k: It's the listen address. I could also make it accept the node peerid. I just wanted one command to handle forwarding in both directions.

It would be cleaner if we could specify this as a multiaddress. We'd need to define a multiaddr proto for string protocol names, probably something like `/mux/protoName`. With that the command wouldn't have to accept magic strings, and the protocol name could get integrated with the address. Creating a libp2p service would look like

```
ipfs p2p /mux/someProto /ip4/127.0.0.1/tcp/$APP_PORT
```

And forwarding local connections to a service:
```
ipfs p2p /ip4/127.0.0.1/tcp/$APP_PORT /ipfs/QmPeer/mux/someProto
```

Does that make more sense?

core/commands/p2p.go
magik6k: It basically restricts one from communicating with libp2p services not in `/p2p` namespace (so e.g. you can't use this to talk to other peers' bitswap which uses `/ipfs/bitswap/1.1.0` as protocol name). The restriction was there before (probably put there by me), but now I'm not sure if we really want/need it.

docs/experimental-features.md
whyrusleeping: Yeah, that makes sense... Its just a bit confusing. It would be nice to be able to make that simpler somehow. @Stebalien @Kubuxu @schomatis thoughts?

core/commands/p2p.go
whyrusleeping: It probably makes sense to make the user always type /p2p/foo even if we restrict them to that namespace. That way, in the future, if we remove that restriction, we don't have to do anything weird.

whyrusleeping:
@magik6k cool, looking pretty good. Last thing I want (aside from getting approvals from @Kubuxu and one other) is a comment I can link to explaining how existing users can update their workflow to work with the changed API. Probably worth posting that comment here: https://github.com/ipfs/go-ipfs/issues/3994 and then updating my main comment to reflect these changes

magik6k:
Rebased + Updated the comment and added a summary in https://github.com/ipfs/go-ipfs/issues/3994#issuecomment-394159058

docs/experimental-features.md
schomatis: I'm getting the error `Error: protocol name must be within '/p2p/' namespace` while trying to replicate the example (to better understand this conversation as I don't really know much about libp2p).

schomatis: Ok, I get it now, changing `/kickass/1.0` to `/p2p/kickass/1.0` fixes that and the netcat example is working.

magik6k: oh, should update that too

schomatis: So, going back to the original question, what seems confusing to me (again, I know nothing about libp2p so feel free to ignore this comment) is to have the p2p port in the front, which seems as if the forwarding is happening only at the p2p layer when in fact the message is routed through different layers (p2p-TCP).

If there were only two arguments (as suggested by @magik6k's solution that for me would be the ideal case) then the from-to relationship would become more evident, but if that isn't possible at the moment I would rather have two different commands that would clearly indicate to the user when the message is received at the p2p layer and routed to the TCP layer (and when the other way around); or at least I would reorganize the command arguments to group the p2p protocol with the p2p address.

whyrusleeping:
Alright, i'm okay with this change (the bare `/ipfs/` on the listener is def a bit weird though). I'd like to get some buyin from at least two others before merging it though. Reviewers! Assemble!

core/commands/p2p.go
```diff
@@ -53,49 +56,121 @@ are refined`,
...
+
+		"forward": p2pForwardCmd,
+		"close":   p2pCloseCmd,
+		"ls":      p2pLsCmd,
 	},
 }
 
-// p2pListenerCmd is the 'ipfs p2p listener' command
-var p2pListenerCmd = &cmds.Command{
+var p2pForwardCmd = &cmds.Command{
```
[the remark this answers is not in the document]
Stebalien: There's also a third case: "connect". That is, if I'm listening on localhost:X, I may want to call `ipfs p2p connect myproto /ip4/localhost/tcp/X /ipfs/QmId`. I'd expect IPFS to:

1. Open a `myproto` stream to `/ipfs/QmId`.
2. Connect to `/ip4/localhost/tcp/X`.
3. Start forwarding the connections.
4. Return the local port (on the CLI/API) IPFS used when connecting to `/ip4/localhost/tcp/X`.
5. Close the stream when the client closes their end of the connection.

Users of existing applications will likely use the forward command but developers of new applications that want to puppet a remote ipfs daemon will likely want to use the `connect` command.

---

This can obviously be added later. I'm just motivating the multiple command option.

core/commands/p2p.go
```diff
@@ -53,49 +56,121 @@ are refined`,
...
 
-	Subcommands: map[string]*cmds.Command{
-		"ls":    p2pListenerLsCmd,
-		"open":  p2pListenerListenCmd,
-		"close": p2pListenerCloseCmd,
+		proto := req.Arguments()[0]
+		listen := req.Arguments()[1]
+		target := req.Arguments()[2]
+
+		if !strings.HasPrefix(proto, "/p2p/") {
```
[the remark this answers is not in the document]
magik6k: Yeah, I was thinking about it, probably makes sense to do it now. I'd like one more 👍 before changing that.

cc @whyrusleeping 

(I'd do that with `--allow-custom-protocol` to allow for backward compatibility)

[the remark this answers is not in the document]
whyrusleeping: Yeah, this seems smart. We already have enough confusing overlapping names :/

magik6k:
I split the forward command into 2 commands per @Stebalien request, it's a bit cleaner. (also, updated all the docs I could find).

p2p/stream.go
```diff
@@ -0,0 +1,84 @@
...
+	s.Local.Close()
+	s.Remote.Reset()
+	s.Registry.Deregister(s.id)
+	return nil
+}
+
+func (s *Stream) startStreaming() {
+	go func() {
+		io.Copy(s.Local, s.Remote)
+		s.Reset()
```
Stebalien: This should only reset on error.

Ideally, we'd support half-closed streams but I don't believe we can feasibly do that.

magik6k:
I did rewrite the locking thing, but it's currently broken for remote listener because `p2p.peerHost.SetStreamHandler` and we may end up trying to override the listener and removing the other handler (https://github.com/libp2p/go-libp2p-host/issues/16).

magik6k:
Thanks for the review!

On `Reset()` - there have been some issues with that, i don't quite remember what was the exact issue there, might have been a bug in libp2p which got fixed while this pr is open. Half-open streams - I don't know if Go exposes that in any way, but IIRC there is (e)poll event sent when a connection half-closes.

Stebalien:
> On Reset() - there have been some issues with that, i don't quite remember what was the exact issue there, might have been a bug in libp2p which got fixed while this pr is open.
> Half-open streams - I don't know if Go exposes that in any way, but IIRC there is (e)poll event sent when a connection half-closes.

So, *our* streams support the half-open state. That is, `Close()` only closes one side. Go TCP connections also support half-open connections but you have to cast to a `TCPConn` and call `CloseWrite` to half-close. Unfortunately, our macon wrapper prevents this.

Stebalien:
> I did rewrite the locking thing, but it's currently broken for remote listener because p2p.peerHost.SetStreamHandler and we may end up trying to override the listener and removing the other handler (libp2p/go-libp2p-host#16).

I see... So, an alternative is to have the `Listener` interface define a `Start() error` method. `Register(listener)` would:

1. Lock
2. Register the listener or return an error
3. Unlock
4. Try to start the listener (which would internally call `startStreaming`).
5. On failure, deregister the listener.

That should keep the locking clean while fixing the bug. We should also fix the Host issue but that may take longer as it may change the interface (depending on how we do it).

core/commands/p2p.go
```diff
@@ -296,20 +313,19 @@ var p2pCloseCmd = &cmds.Command{
...
+				return true
 			}
-			if l {
-				out = out && (listen == listener.ListenAddress())
+			if p && proto != listener.Protocol() {
+				return false
 			}
-			if t {
-				out = out && (target == listener.TargetAddress())
+			if l && listen != listener.ListenAddress() {
```
Stebalien: Ah. So, I wasn't paying attention... this needs to be `listen.Equals(listener.ListenAddress())` (same with target below). They're not strings.

p2p/listener.go
```diff
@@ -27,25 +32,21 @@ type ListenerRegistry struct {
...
-}
 
-func (r *ListenerRegistry) unlock() {
-	r.lk.Unlock()
-}
-
-// Register registers listenerInfo in this registry
-func (r *ListenerRegistry) Register(l Listener) {
-	defer r.lk.Unlock()
+	if err := l.start(); err != nil {
```
Stebalien: You're probably going to want to:

1. Lock
2. Register
3. **unlock**
4. Start
5. unregister (under a lock) on error.

That is, starting under a lock is probably not going to work out well.

p2p/stream.go
```diff
@@ -36,15 +36,19 @@ func (s *Stream) Close() error {
 // Reset closes stream endpoints and deregisters it
 func (s *Stream) Reset() error {
 	s.Local.Close()
-	s.Remote.Reset()
+	s.Remote.Close()
```
Stebalien: This should still be Reset, unless I'm missing something...

magik6k: I was debugging the thing and this got commited by accident

p2p/p2p.go
```diff
@@ -1,23 +1,18 @@
...
+	logging "gx/ipfs/Qmbi1CTJsbnBZjCEgc2otwu8cUFPsGpzWXG7edVCLZ7Gvk/go-log"
 )
 
+var log = logging.Logger("p2p-mount")
+
 // P2P structure holds information on currently running streams/listeners
 type P2P struct {
-	Listeners ListenerRegistry
-	Streams   StreamRegistry
+	Listeners *ListenerRegistry
```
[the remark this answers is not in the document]
magik6k: To avoid copying them by accident (I don't think it happens anywhere), and most of our codebase refers to structs like those with pointers already.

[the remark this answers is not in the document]
Stebalien: I've been trying to reduce that as it's not idiomatic go.

core/commands/p2p.go
```diff
@@ -196,149 +266,164 @@ var p2pStreamLsCmd = &cmds.Command{
...
+			}
+			return true
 		}
 
-		listenerInfo, err := n.P2P.Dial(n.Context(), addr, peer, proto, bindAddr)
-		if err != nil {
-			res.SetError(err, cmdkit.ErrNormal)
-			return
+		var closed int
+		for _, listener := range n.P2P.Listeners.Listeners {
```
Stebalien: I think we're going to need to hold a lock while iterating over these. We may want to do this in two steps: filter under a lock and then close.

p2p/local.go
```diff
@@ -0,0 +1,132 @@
...
+	maListener, err := manet.Listen(l.laddr)
+	if err != nil {
+		return err
+	}
+
+	l.listener = maListener
+	return nil
+}
+
+func (l *localListener) Close() error {
```
Stebalien: We may need a lock here. That depends on whether or not we close under a global lock.

p2p/remote.go
```diff
@@ -0,0 +1,104 @@
...
+		panic(err)
+	}
+	return addr
+}
+
+func (l *remoteListener) TargetAddress() ma.Multiaddr {
+	return l.addr
+}
+
+func (l *remoteListener) Close() error {
```
Stebalien: Again, we may need a lock.

p2p/stream.go
```diff
@@ -0,0 +1,88 @@
...
+	return nil
+}
+
+func (s *Stream) startStreaming() {
+	go func() {
+		_, err := io.Copy(s.Local, s.Remote)
+		if err != nil {
+			s.Reset()
+		} else {
+			s.Close()
```
Stebalien: Ok. *Now* I remember the issue here. We can't always call `Close()` on a stream from one thread while another is writing (some transports will have memory races if we do that).

Regardless, given that we can't really "nicely" shut down the connection, maybe we should just reset the stream? This is really bugging me. TCP supports the following usecase:

1. Client: Write and finish with an EOF.
2. Server: Read until EOF, handle request, write and finish with an EOF.
3. Client: Read response until EOF.

As it stands, these streams won't support this.

---

So, for now, maybe we should go back to simply resetting the stream and documenting the limitation. In the future, we can try exposing this half-closed feature for transports that support it.

(we should also put a big fat comment here explaining this)

core/commands/p2p.go
```diff
@@ -53,124 +61,186 @@ are refined`,
...
-			list := v.(*P2PLsOutput)
-			buf := new(bytes.Buffer)
-			w := tabwriter.NewWriter(buf, 1, 2, 1, ' ', 0)
-			for _, listener := range list.Listeners {
-				if headers {
-					fmt.Fprintln(w, "Address\tProtocol")
-				}
+// forwardRemote forwards libp2p service connections to a manet address
+func forwardRemote(ctx context.Context, p *p2p.P2P, proto protocol.ID, target ma.Multiaddr) error {
+	// TODO: return some info
```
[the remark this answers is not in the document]
magik6k: well, the 'handle' here is really just what the user specified. In the forwardLocal case this makes a bit more sense in case where user specifies tcp port 0, where we'd return what the OS picked.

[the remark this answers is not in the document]
Stebalien: Hm. Right, they can just shut it down via the protocol name.

p2p/stream.go
magik6k: > Regardless, given that we can't really "nicely" shut down the connection, maybe we should just reset the stream?

Just doing resets here breaks the tests in some quite hard to debug ways. I'd leave it this way for now and come back once the other half-open stuff is done

p2p/local.go
magik6k: done

p2p/remote.go
magik6k: done

p2p/stream.go
Stebalien: Sounds good. Basically, as it stands, the "correct" way to use this feature is to negotiate a connection finish at the application layer before closing the connection. With the multiaddr-net changes, that shouldn't be necessary.

core/commands/p2p.go
```diff
@@ -296,31 +313,36 @@ var p2pCloseCmd = &cmds.Command{
...
 				continue
 			}
-			listener.Close()
-			closed++
+			todo = append(todo, l)
 		}
-		res.SetOutput(closed)
+		n.P2P.Listeners.Unlock()
+
+		for _, l := range n.P2P.Listeners.Listeners {
```
Stebalien: Range over todo.

Stebalien: We should probably also have a test case (sharness?) that checks to make sure we don't close the wrong listeners.

p2p/local.go
Stebalien: Where?

magik6k: `l.p2p.Listeners.Deregister` now returns a boolean indicating whether the listener got deregistered (which happens with a lock, so this should only happen once). Would have been cleaner if I didn't do `git commit --amend --no-edit` all the time..

Stebalien: Ah. So, the issue is that someone may concurrently try to read `listener`.

magik6k:
done, the locking stuff in listener `Close` should be fine now.

p2p/local.go
Stebalien: It looks like there may still be a race. If there are two simultaneous close calls and a concurrent open call (for the same listener triplet) the one of the close calls may deregister the listener registered by the open call.

I think we need to pass the listener *into* `Deregister` and have deregister make sure it's the right one (compare pointers).

core/commands/p2p.go
```diff
@@ -196,149 +266,177 @@ var p2pStreamLsCmd = &cmds.Command{
...
-		closeAll, _, _ := req.Option("all").Bool()
-		var proto string
+		output := &P2PStreamsOutput{}
 
-		if !closeAll {
-			if len(req.Arguments()) == 0 {
-				res.SetError(errors.New("no protocol name specified"), cmdkit.ErrNormal)
-				return
-			}
+		for id, s := range n.P2P.Streams.Streams {
```
Stebalien: Need to lock while iterating.

core/commands/p2p.go
```diff
@@ -377,19 +475,19 @@ var p2pStreamCloseCmd = &cmds.Command{
 			}
 		}
 
-		for _, stream := range n.P2P.Streams.Streams {
-			if !closeAll && handlerID != stream.HandlerID {
+		for id, stream := range n.P2P.Streams.Streams {
```
Stebalien: Lock while iterating (but not while closing).

p2p/listener.go
```diff
@@ -0,0 +1,87 @@
...
+	r.Lock()
+	k := getListenerKey(l)
+
+	if _, ok := r.Listeners[k]; ok {
+		r.Unlock()
+		return errors.New("listener already registered")
+	}
+
+	r.Listeners[k] = l
+	r.starting[k] = struct{}{}
```
Stebalien: I think a simple state mutex inside the listener may be simpler. I.e., take it on start before registering, release it after starting, and take it before closing.

p2p/remote.go
```diff
@@ -0,0 +1,99 @@
...
+
+	if err := p2p.Listeners.Register(listener); err != nil {
+		return nil, err
+	}
+
+	return listener, nil
+}
+
+func (l *remoteListener) start() error {
+	// TODO: handle errors when https://github.com/libp2p/go-libp2p-host/issues/16 will be done
```
Stebalien: So, I think this may actually be a problem. As implemented, we can register two "remote" listeners on the same protocol with two different local addresses (they'll have different keys).

However, I doubt it's any worse than it is today.

p2p/listener.go
Stebalien: Alternatively, once we fix the stream handler registration problem, we can probably just:

1. Not use a "listener key". Just put listeners in a list, we never really need to look them up.
2. Fully start listening *before* registering. If there's a clash, listening should fail.

Thoughts?

magik6k: > Fully start listening before registering. If there's a clash, listening should fail.

Well, I guess once https://github.com/libp2p/go-libp2p-host/issues/16 is fixed this would be the right way

hsanjuan:
Bumping this, it's a pity that so much work went into this and then it was left unmerged. Currently `ipfs p2p` is quite unusable (connections closed after first stream is finished).. :(

Stebalien:
@hsanjuan 

> Bumping this, it's a pity that so much work went into this and then it was left unmerged. Currently ipfs p2p is quite unusable (connections closed after first stream is finished).. :(

I agree.

As-implemented, this has a few data-races. The plan was to try to fix  libp2p/go-libp2p-host#16 first but that looks like that may take a while. I took a stab at fixing that last week but it looks like it'll be a quite a bit of work to do it properly. 

Therefore, I think we should just get this in shape *without* fixing that issue.

Solution: Use `host.Mux().Ls()` to list the in-use protocols before registering. We'll have to:

1. Make sure to lock to make the call to `Ls()` and registering the protocol handler atomic.
2. Assume that we're the only service dynamically registering protocol handlers (yuck!).
3. Accept the `O(n)` overhead of this check :(.

We can optimize this when `--allow-custom-protocol` is *not* passed by tracking all custom protocols (starting with `/x/`) locally (and not bothering to check `Mux().Ls()`).

@magik6k thoughts?

p2p/local.go
magik6k: This is a rather racy situation either way:

What can happen:
* close successfully; open; close freshly open
* close successfully; close fail; open;
* close successfully; open; close fail (with `listener didn't start yet` error)
* open fail; close successfully; close fail;
* Any other combination of those

I don't think other locking scheme would make this more 'defined' behavior

magik6k:
Using Ls here - `Ls(w io.Writer) error` - feels kind of hacky to use, as the method seems to have been made for internal use, might be simpler to just do it the right way (by fixing `SetStreamHandler`)

Stebalien:
> Using Ls here - Ls(w io.Writer) error - feels kind of hacky to use, as the method seems to have been made for internal use, might be simpler to just do it the right way (by fixing SetStreamHandler)

Oh dear. I thought that method returned a `[]string`. I'll take a second stab at fixing the stream handler registration. I'll try to limit my changes this time...

whyrusleeping:
Feature request: Tag connections with the connmanager as they are used. Using this for ssh, and the connection randomly drops

magik6k:
`SetStreamHandler` not returning errors when protocol is already registered.

p2p/listener.go
```diff
@@ -0,0 +1,122 @@
...
+		Listeners: map[listenerKey]Listener{},
+		starting:  map[listenerKey]struct{}{},
+	}
+
+	addr, err := ma.NewMultiaddr(maPrefix + id.Pretty())
+	if err != nil {
+		panic(err)
+	}
+
+	host.SetStreamHandlerMatch("/x/", func(p string) bool {
```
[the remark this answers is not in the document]
magik6k: Added a test that actually tests communication on custom proto, it disagrees.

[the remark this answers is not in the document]
Stebalien: Ah. You're right. I forgot that the muxer doesn't care about the protocol name when using this function.

magik6k:
Do we want https://github.com/multiformats/go-multistream/pull/31 in first, or is using the match func workaround fine for now?

Stebalien:
> Do we want multiformats/go-multistream#31 in first, or is using the match func workaround fine for now?

I'm fine with the workaround.

Stebalien:
@magik6k my suggestions were intended to simplify/reduce code. If you're not seeing negative diffstats, we may need to go in a different direction.

magik6k:
Rebased

p2p/remote.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+	return addr
+}
+
+func (l *remoteListener) TargetAddress() ma.Multiaddr {
+	return l.addr
+}
+
+func (l *remoteListener) Close() error {
+	_, err := l.p2p.ListenersP2P.Deregister(string(l.proto))
```
Stebalien: Now that we're just registering by *protocol*, this is going to have that close race condition we discussed. That is, if I try to close a listener forwarding to a *specific* local address twice while someone else is trying to *open* a listener (to a different local address), I can close their listener.

I believe the correct fix is to just do *everything* inside the listener manager. That is, no listener handles, etc. Just add a `Close(...)` method to the listener's object that takes the close "constraints" (protocol, localAddr, remoteAddr).

Does that make sense? This is what I meant by reducing some of the abstractions.

magik6k:
I've added `Listeners.Close`, looks like an improvement. I've also tried to merge `Listeners` and `Listener` further, but it resulted in +100 messy lines of duplicated code, so that got hit with `reset --hard`.

Stebalien:
:tada:

I think this has received enough review (4) and we haven't changed the *interfaces* in most of this review so I'm going to go ahead and merge it. It's not *too* late to make changes but we should try to make them *before* the next release (in case anyone gets cold feet and wants to change this).

Stebalien:
Also, @magik6k: :heart: :heart: :heart: :heart: :heart: :heart: :heart:

PR #4933, opened 2018-04-11
Fix typo in error message

maxkerp:
I've never written a line of go in my life, but I came across this typo while using go-ipfs and thought I might as well just fix it myself. I'm pretty sure you guys are really busy.


whyrusleeping:
Thanks @maxkerp :) Busy is definitely a word that could be used to describe things around here

PR #4940, opened 2018-04-16
dag: deduplicate AddNodeLinkClean into AddNodeLink

### LINKED ISSUE #4729: AddNodeLink and AddNodeLinkClean seem to have the same functionality
While reviewing code for issue #4052 I came across the functions [`AddNodeLink`](https://github.com/ipfs/go-ipfs/blob/master/merkledag/node.go#L92) and [`AddNodeLinkClean`](https://github.com/ipfs/go-ipfs/blob/master/merkledag/node.go#L109) that would seem to have the same code except for an extra line in `AddNodeLink` where [the name is set to the link that will be added](https://github.com/ipfs/go-ipfs/blob/master/merkledag/node.go#L100), but the link name is not used in the [`AddRawLink`](https://github.com/ipfs/go-ipfs/blob/master/merkledag/node.go#L121) function that receives the link (as the name is specified as a separate argument), and the link itself seems to be discarded after the call (its members `Size` and `Cid` are copied but not the link structure).

Both functions are equally used in the code base but I'm failing (from reading the code and the comments) to see their difference, am I missing something?

schomatis:
Closes #4729.

### DIFF
--- a/core/coreunix/metadata.go
+++ b/core/coreunix/metadata.go
@@ -25,7 +25,7 @@ func AddMetadataTo(n *core.IpfsNode, skey string, m *ft.Metadata) (string, error
 	}
 
 	mdnode.SetData(mdata)
-	if err := mdnode.AddNodeLinkClean("file", nd); err != nil {
+	if err := mdnode.AddNodeLink("file", nd); err != nil {
 		return "", err
 	}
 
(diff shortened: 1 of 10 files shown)


schomatis:
@Stebalien Thanks for that clarifying comment, I'll add it to the commit message.

PR #4943, opened 2018-04-18
github: remove go-ipfs version from issue template

lgierth:


ISSUE_TEMPLATE.md
```diff
@@ -14,8 +14,8 @@ Read https://github.com/ipfs/go-ipfs/blob/master/docs/github-issue-guide.md if y
 <!--
 Output From `ipfs version --all`
 
-If your version is older than *0.4.11* please update
-and re-check if the problem persists.
+Please check CHANGELOG.md for the latest version of go-ipfs
```
magik6k: Much of the stuff in the changelog may sound cryptic and/or unrelated for new users, it's probably a better idea to just tell the users to update.

lgierth: Okay, pointing to dist.ipfs.io instead

PR #4946, opened 2018-04-18
cid-sec: fix bitswap strom caused by insecure CIDs

Kubuxu:
When we introduced CID security we didn't take into account that bitswap
might repeatedly try getting the objects from the network if it fails
putting them into the blockstore.

A solution from this is not requesting those objects from bitswap.
The proper solution of failing at CID creation will make it much more
cleaner in future.

License: MIT


blockservice/blockservice.go
```diff
@@ -251,15 +251,22 @@ func (s *blockService) GetBlocks(ctx context.Context, ks []*cid.Cid) <-chan bloc
 
 func getBlocks(ctx context.Context, ks []*cid.Cid, bs blockstore.Blockstore, f exchange.Fetcher) <-chan blocks.Block {
 	out := make(chan blocks.Block)
-	for _, c := range ks {
-		// hash security
-		if err := verifcid.ValidateCid(c); err != nil {
-			log.Errorf("unsafe CID (%s) passed to blockService.GetBlocks: %s", c, err)
```
whyrusleeping: Seems like the right solution to this might be to just add a return of a closed block channel after the error here.

Kubuxu: Not really, a file might have an unsafe block near the end of a file, most of the file should be OK to read. Just the end might be bad.

That is why I sort out the unsafe CIDs and fetch and/or read only the good ones.

Kubuxu:
@Stebalien I will try.

PR #4948, opened 2018-04-18
Single ConstructNode

PlayerWithoutName:
Refactored ConstructNode usage, as current one is a bit confusing and it's hard to wire in CoreAPI providing for plugins.

### DIFF
--- a/cmd/ipfs/daemon.go
+++ b/cmd/ipfs/daemon.go
@@ -268,10 +268,6 @@ func daemonFunc(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment
 			return err
 		}
 
-		repo, err = fsrepo.Open(cctx.ConfigRoot)
-		if err != nil {
-			return err
-		}
 	case nil:
 		break
 	}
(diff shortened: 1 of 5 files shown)


cmd/ipfs/daemon.go
```diff
@@ -267,11 +267,6 @@ func daemonFunc(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment
 			return
 		}
 
-		repo, err = fsrepo.Open(cctx.ConfigRoot)
```
Stebalien: I'm pretty sure this is still necessary.

magik6k: This will happen in https://github.com/ipfs/go-ipfs/pull/4948/files#diff-153cb71153ee178fe2869148600ce8d5R113

Stebalien:
@magik6k could I get your opinion on this? Is this moving in the direction you're heading WRT the CoreAPI?

Stebalien:
(closing as @magik6k is reworking all of this code)

PR #4958, opened 2018-04-21
ci: Add more timeouts to the jenkins pipeline

magik6k:
With this hanging jobs should fail a bit earlier. Currently the timeouts are set to about 2x what it takes for a given step normally.


magik6k:
Removed global timeout to keep jobs timing out when waiting in queue

Issue #4953, opened 2018-04-20
Error when renaming keys


martin-looklet:
#### Version information:

go-ipfs version: 0.4.14-
Repo version: 6
System version: amd64/darwin
Golang version: go1.10

#### Type:

Bug

#### Description:

If I create a key with `ipfs key gen` and try to rename it, I get an error, but the key is renamed:
```
> ipfs key gen --type=rsa --size=2048 key1
> ipfs key rename key1 key2 
Error: expected a KeyRenameOutput as command result
> ipfs key list
self
key2
```

grokcoder:
This issue should be closed

Kubuxu:
@grokcoder thanks for fixing it.

PR #4962, opened 2018-04-22
Fix key rename command output error

grokcoder:
fix: Error when renaming keys #4953

License: MIT
Signed-off-by: Xiaoyi Wang <wangxiaoyi@hyperchain.cn>

### DIFF
--- a/core/commands/keystore.go
+++ b/core/commands/keystore.go
@@ -306,7 +306,11 @@ var keyRenameCmd = &cmds.Command{
 	},
 	Marshalers: cmds.MarshalerMap{
 		cmds.Text: func(res cmds.Response) (io.Reader, error) {
-			k, ok := res.Output().(*KeyRenameOutput)
+			v, err := unwrapOutput(res.Output())
+			if err != nil {
+				return nil, err
+			}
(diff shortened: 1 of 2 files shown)


magik6k:
> this can reproduce before I fix it, but failed if I fixed it

That's what I asked for, a test which will fail before the issue was fixed

I looked at the current keystore tests and it looks like there already is a test for rename - https://github.com/ipfs/go-ipfs/blob/master/test/sharness/t0165-keystore.sh#L53, but only in offline mode

Basically, what you want to do is to replace https://github.com/ipfs/go-ipfs/blob/master/test/sharness/t0165-keystore.sh#L72 with something like

```bash
test_key_cmd

test_launch_ipfs_daemon

test_expect_success "cleanup" '
  ipfs key rm fooed
'

test_key_cmd

test_kill_ipfs_daemon
```

grokcoder:
@magik6k thx for explain, the issue what I fixed is the key rename output msg parse error, not the rename function. The rename function is worked but with error output msg.

whyrusleeping:
Thanks @grokcoder!

Issue #4961, opened 2018-04-22
Optimize sharness tests


magik6k:
Some sharness tests take way more time to complete than they have to.

Results from https://ci.ipfs.team/job/IPFS/job/go-ipfs/job/PR-4959/7/testReport/sharnessLinux/ (sort by duration) / https://github.com/ipfs/go-ipfs/pull/4959

* [ ] - `t0181-private-network / 12 - try connecting node in public network with priv networks` - should timeout way sooner
* [ ] - `t0130-multinode` - adds a tree of 5k files multiple times, might be a bit overkill, at least of something that isn't a benchmark
* [ ] - `t0250-files-api` is huuuuge, would be good to split into some separate tests to make use of make parallelism
* [ ] - `t0260-sharding` operates on quite large directories, not sure if we can do much about it
* [ ] - `t0040-add-and-cat` does some stuff on large files, but majority of the time spent there seems to be spent in the ton of tests there, might be worth to split too

kevina:
I am probably responsible for the huge number of tests in `t0250-files-api` and `t0040-add-and-cat`.  I am skeptical on the benefit of splitting these tests as they are just one of many shareness tests that need to be run, so the probability that they will increase parallelism is low.  Nevertheless, if we do split them up I would like to take lead on that.

Kubuxu:
@kevina it will decrease the duration of the tests significantly as these are the tests we end up waiting for at the end of the run.

If we split them in half (online and offline) it is quite possible that they will run in parallel with other tests instead of being the only two tests running at the end.

PR #4966, opened 2018-04-22
Fix: sharness iptb connect timeout

PlayerWithoutName:
Resolves the first point in #4961 by adding timeout to connect. Based on  #4965.


test/sharness/t0181-private-network.sh
```diff
@@ -61,7 +61,7 @@ test_expect_success "start nodes" '
 '
 
 test_expect_success "try connecting node in public network with priv networks" '
-  iptb connect [1-4] 0
+  iptb connect [1-4] 0 --timeout=2s
```
Kubuxu: Can you put the option before the arguments? It is more natural that way.

PR #4970, opened 2018-04-23
Add docs for --profile=lowpower

camelmasa:
Continue the PR https://github.com/ipfs/go-ipfs/pull/4154.

Added description of `lowpower`.


Kubuxu:
I think it might be better to point users toward: `ipfs config profile --help`.

PR #4976, opened 2018-04-25
docs: fix experimental features toc

lanzafame:


lanzafame:
@Stebalien @Kubuxu for some reason the following sharness test failed:
```
not ok 89 - 'ipfs daemon' can be killed
```
I don't seem to be able to re-run circle-ci builds, but considering this is a docs change, I don't I broke the build :)

Kubuxu:
@lanzafame you did not. It is caused by a flaky feature that works but is hard to test without the test being flaky as well.

PR #4979, opened 2018-04-26
Fetch Hamt Children Nodes in Parallel

### LINKED ISSUE #4908: Sharded directory fetching is unusably slow
#### Version information:
0.4.15-dev

#### Type:
Bug/performance issue

#### Description:
More context is available over in https://github.com/tesserai/iptf/issues/2

I'm trying to get reasonable performance for just *listing* the names of entries in a sharded directory that's not yet cached locally. This operation takes hours right now. With @Stebalien's help I've been able to determine that it's only requesting one hash at a time (as indicated by `ipfs bitswap wantlist`.

Seems like IPFS should be requesting more than one block at a time in this scenario. Creating a separate issue to track this specific performance issue separately from others.

child of #5487

kevina:
Closes #4908

### DIFF
--- a/unixfs/hamt/fetcher.go
+++ b/unixfs/hamt/fetcher.go
@@ -24,7 +24,6 @@ type fetcher struct {
 	ctx   context.Context
 	dserv ipld.DAGService
 
-	newJob chan *Shard
 	reqRes chan *Shard
 	result chan result
 
@@ -63,7 +62,6 @@ func startFetcher(ctx context.Context, dserv ipld.DAGService) *fetcher {
 	f := &fetcher{
 		ctx:    ctx,
 		dserv:  dserv,
-		newJob: make(chan *Shard, 16),
 		reqRes: make(chan *Shard),
 		result: make(chan result),
 		idle:   true,
@@ -80,14 +78,13 @@ type result struct {
 	errs []error
 }
 
-// get recursively gets the missing child shards for the hamt object.
+// get gets the missing child shards for the hamt object.
 // The missing children for the passed in shard is returned.  The
-// children of the children are recursively retrieved in the
-// background.  The result is the result of the batch request and not
-// just the single job.  In particular, if the 'errs' field is empty
-// the 'vals' of the result is guaranteed to contain the all the
-// missing child shards, but the map may also contain child shards of
-// other jobs in the batch
+// children are then also retrieved in the background.  The result is
+// the result of the batch request and not just the single job.  In
+// particular, if the 'errs' field is empty the 'vals' of the result
+// is guaranteed to contain the all the missing child shards, but the
+// map may also contain child shards of other jobs in the batch
 func (f *fetcher) get(hamt *Shard) result {
 	f.reqRes <- hamt
 	res := <-f.result
@@ -114,8 +111,6 @@ func (f *fetcher) mainLoop() {
 	var want *Shard
 	for {
 		select {
-		case j := <-f.newJob:
-			f.mainLoopAddJob(j)
 		case id := <-f.reqRes:
 			if want != nil {
 				// programmer error
@@ -128,12 +123,13 @@ func (f *fetcher) mainLoop() {
 					// no children that need to be retrieved
 					f.result <- result{vals: make(map[string]*Shard)}
 				}
+				if f.idle {
+					f.launch()
+				}
 			}
 			if j.res.vals != nil {
 				f.hits++
-				delete(f.jobs, id)
-				f.doneCnt--
-				f.result <- j.res
+				f.mainLoopSendResult(j)
 			} else {
 				if j.idx != -1 {
 					f.misses++
@@ -149,17 +145,15 @@ func (f *fetcher) mainLoop() {
 		case cnt := <-f.done:
 			f.doneCnt += cnt
 			f.launch()
-			log.Infof("fetcher: batch job done")
-			log.Infof("fetcher stats (done, hits, nearMisses, misses): %d %d %d %d", f.doneCnt, f.hits, f.nearMisses, f.misses)
 			if want != nil {
 				j := f.jobs[want]
 				if j.res.vals != nil {
-					delete(f.jobs, want)
-					f.doneCnt--
-					f.result <- j.res
+					f.mainLoopSendResult(j)
 					want = nil
 				}
 			}
+			log.Infof("fetcher: batch job done")
+			log.Infof("fetcher stats (done, hits, nearMisses, misses): %d %d %d %d", f.doneCnt, f.hits, f.nearMisses, f.misses)
 		case <-f.ctx.Done():
 			log.Infof("fetcher: exiting")
 			log.Infof("fetcher stats (done, hits, nearMisses, misses): %d %d %d %d", f.doneCnt, f.hits, f.nearMisses, f.misses)
@@ -169,8 +163,6 @@ func (f *fetcher) mainLoop() {
 	}
 }
 
-// addJob adds a job to retrive the missing child shards for the
-// provided shard
 func (f *fetcher) mainLoopAddJob(hamt *Shard) *job {
 	children := hamt.missingChildShards()
 	if len(children) == 0 {
@@ -184,10 +176,25 @@ func (f *fetcher) mainLoopAddJob(hamt *Shard) *job {
 	f.cidCnt += len(j.cids)
 	f.todo.push(j)
 	f.jobs[j.id] = j
+	return j
+}
+
+func (f *fetcher) mainLoopSendResult(j *job) {
+	f.result <- j.res
+	delete(f.jobs, j.id)
+	f.doneCnt--
+	if len(j.res.errs) != 0 {
+		return
+	}
+	// we want the first child to be the first to run not the last so
+	// add the jobs in the reverse order
+	for i := len(j.cids) - 1; i >= 0; i-- {
+		hamt := j.res.vals[string(j.cids[i].Bytes())]
+		f.mainLoopAddJob(hamt)
+	}
 	if f.idle {
 		f.launch()
 	}
-	return j
 }
 
 type batchJob struct {
@@ -217,7 +224,9 @@ func (f *fetcher) launch() {
 	}
 
 	if len(bj.cids) == 0 {
-		log.Infof("fetcher: entering idle state: no more jobs")
+		if !f.idle {
+			log.Infof("fetcher: entering idle state: no more jobs")
+		}
 		f.idle = true
 		return
 	}
@@ -238,7 +247,6 @@ func (f *fetcher) launch() {
 				fetched.errs = append(fetched.errs, err)
 				continue
 			}
-			f.newJob <- hamt
 			fetched.vals[string(no.Node.Cid().Bytes())] = hamt
 		}
 		for _, job := range bj.jobs {
--- a/unixfs/hamt/fetcher.go
+++ b/unixfs/hamt/fetcher.go
@@ -24,28 +24,33 @@ type fetcher struct {
 	ctx   context.Context
 	dserv ipld.DAGService
 
-	newJob chan *job
+	newJob chan *Shard
 	reqRes chan *Shard
 	result chan result
 
 	idle bool
 
-	done chan struct{}
+	done chan int
 
 	todoFirst *job            // do this job first since we are waiting for its results
 	todo      jobStack        // stack of jobs that still need to be done
 	jobs      map[*Shard]*job // map of all jobs in which the results have not been collected yet
 
 	// stats relevent for streaming the complete hamt directory
+	doneCnt    int // job's done but results not yet retrieved
 	hits       int // job's result already ready, no delay
 	nearMisses int // job currently being worked on, small delay
 	misses     int // job on todo stack but will be done in the next batch, larger delay
+
+	// other useful stats
+	cidCnt int
 }
 
 // batchSize must be at least as large as the largest number of cids
 // requested in a single job.  For best perforamce it should likely be
 // sligtly larger as jobs are poped from the todo stack in order and a
-// job close to the batchSize could forse a very small batch to run.
+// job close to the batchSize could force a very small batch to run.
+// Recommend minimum size: 256 + 64 = 320
 const batchSize = 320
 
 //
@@ -58,27 +63,15 @@ func startFetcher(ctx context.Context, dserv ipld.DAGService) *fetcher {
 	f := &fetcher{
 		ctx:    ctx,
 		dserv:  dserv,
-		newJob: make(chan *job),
+		newJob: make(chan *Shard, 16),
 		reqRes: make(chan *Shard),
 		result: make(chan result),
 		idle:   true,
-		done:   make(chan struct{}),
+		done:   make(chan int),
 		jobs:   make(map[*Shard]*job),
 	}
 	go f.mainLoop()
 	return f
-}
-
-// addJob adds a job to retrive the missing child shards for the
-// provided shard
-func (f *fetcher) addJob(hamt *Shard) bool {
-	children := hamt.missingChildShards()
-	if children == nil {
-		return false
-	}
-	j := &job{id: hamt, cids: children}
-	f.newJob <- j
-	return true
 }
 
 // result contains the result of a job, see getResult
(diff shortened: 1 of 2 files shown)


whyrusleeping:
If we're looking for more perf, using bitswap sessions for this would help, heres a small patch that enables that: https://github.com/ipfs/go-ipfs/compare/feat/ls-session?expand=1

Kubuxu:
Flagging as WIP please remove the flag and ping me when done.

ajbouh:
@kevina What's the latest here? Eagerly awaiting this change! :)

whyrusleeping:
@ajbouh I think right now we just need some cleanup and review. cc @magik6k and @schomatis for review

unixfs/hamt/fetcher.go
```diff
@@ -0,0 +1,297 @@
...
+					f.todoFirst = j
+				} else {
+					f.nearMisses++
+				}
+				want = id
+			}
+		case bj := <-f.done:
+			f.doneCnt += len(bj.jobs)
+			f.cidCnt += len(bj.cids)
+			f.launch()
```
schomatis: I'm having trouble following the `launch()`/`idle` dynamic but this unconditional `launch()` call seems to be guaranteeing that we don't reach a deadlock state where the  `idle` is `false` and no one can call `launch()` to unblock it, right?

unixfs/hamt/fetcher.go
```diff
@@ -0,0 +1,297 @@
...
+	errs []error
+}
+
+// get gets the missing child shards for the hamt object.
+// The missing children for the passed in shard is returned.  The
+// children are then also retrieved in the background.  The result is
+// the result of the batch request and not just the single job.  In
+// particular, if the 'errs' field is empty the 'vals' of the result
+// is guaranteed to contain the all the missing child shards, but the
+// map may also contain child shards of other jobs in the batch
```
schomatis: Not sure I understand, so I may get more than I asked for? Or rather, many different `get` operation results may be returned in a single `result` of a particular `get`?

schomatis:
@kevina Most of the review comments are minor details, so feel free to ignore them, the only relevant two  are at line 152 and 162.

unixfs/hamt/hamt.go
```diff
@@ -284,33 +284,124 @@ func (ds *Shard) getChild(ctx context.Context, i int) (child, error) {
...
+		return nil, err
+	}
+
+	ds.children[i] = c
+	return c, nil
+}
+
+func (ds *Shard) missingChildShards() []*cid.Cid {
+	if len(ds.children) != len(ds.nd.Links()) {
+		panic("inconsistent lengths between children array and Links array")
```
whyrusleeping: Are there any scenarios that this could happen with a maliciously crafted node?

unixfs/hamt/hamt.go
```diff
@@ -284,33 +284,124 @@ func (ds *Shard) getChild(ctx context.Context, i int) (child, error) {
...
+
+	ds.children[i] = c
+	return c, nil
+}
+
+// preloadChildren populates the 'children' array if some child shards
+// are not already loaded they are fetched in parallel using GetMany
+func (ds *Shard) preloadChildren(ctx context.Context, f *fetcher) error {
+	if len(ds.children) != len(ds.nd.Links()) {
+		return fmt.Errorf("inconsistent lengths between children array and Links array")
```
whyrusleeping: feels inconsistent to panic over this in one place, but throw an error here.

unixfs/hamt/fetcher.go
```diff
@@ -0,0 +1,297 @@
...
+			f.mainLoopLogStats()
+		case <-f.ctx.Done():
+			log.Infof("fetcher: exiting")
+			f.mainLoopLogStats()
+			return
+		}
+	}
+}
+
+func (f *fetcher) mainLoopAddJob(hamt *Shard) *job {
```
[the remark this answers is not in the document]
kevina: Possible.  But I prefer to keep it to make it clear that those methods should only be called within the main loop.

unixfs/hamt/fetcher.go
```diff
@@ -0,0 +1,297 @@
...
+	var want *Shard
+	f.start = time.Now()
+	for {
+		select {
+		case id := <-f.reqRes:
+			if want != nil {
+				// programmer error
+				panic("fetcher: can not request more than one result at a time")
+			}
+			j, ok := f.jobs[id]
```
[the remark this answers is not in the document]
kevina: Yes its okay.  If you ask for the same shard with a different pointer it won't be detected, but that should not happen in practice.

unixfs/hamt/fetcher.go
```diff
@@ -0,0 +1,297 @@
...
+
+func (f *fetcher) mainLoopAddJob(hamt *Shard) *job {
+	children := hamt.missingChildShards()
+	if len(children) == 0 {
+		return nil
+	}
+	j := &job{id: hamt, cids: children}
+	if len(j.cids) > batchSize {
+		// programmer error
+		panic("job size larger than batchSize")
```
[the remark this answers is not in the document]
kevina: Yes it is the responsibility of the consumer to now upfront how many children will be requested per shard.  The HAMT shard does not allow more than 256 children in a single node, so this is reasonable.

unixfs/hamt/fetcher.go
```diff
@@ -0,0 +1,297 @@
...
+			}
+			hamt, err := NewHamtFromDag(f.dserv, no.Node)
+			if err != nil {
+				fetched.errs = append(fetched.errs, err)
+				continue
+			}
+			fetched.vals[string(no.Node.Cid().Bytes())] = hamt
+		}
+		for _, job := range bj.jobs {
+			job.res = fetched
```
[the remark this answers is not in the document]
kevina: Yes.  And I do this for code simplicity and performance reasons.

unixfs/hamt/fetcher.go
```diff
@@ -0,0 +1,297 @@
...
+	f.jobs[j.id] = j
+	return j
+}
+
+func (f *fetcher) mainLoopSendResult(j *job) {
+	f.result <- j.res
+	delete(f.jobs, j.id)
+	f.doneCnt--
+	if len(j.res.errs) != 0 {
+		return
```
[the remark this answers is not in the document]
kevina: I am not sure I understand the question.

unixfs/hamt/hamt.go
kevina: Not sure.  The reason I panic here is because I am unsure what to do with the error.

kevina: Actually the reason for the panic is because it should not happen.  This is only called inside `preloadChildren` which already verifies the link.

Nevertheless I changed this method to return an error and handled the error by panicking in the fetcher code.

kevina: Okay.  This should be fixed now.

unixfs/hamt/hamt.go
kevina: Fixed.

unixfs/hamt/fetcher.go
```diff
@@ -0,0 +1,297 @@
...
+
+type job struct {
+	id  *Shard
+	idx int /* index in the todo stack, an index of -1 means the job
+	   is already done or being worked on now */
+	cids []*cid.Cid
+	res  result
+}
+
+type jobStack struct {
```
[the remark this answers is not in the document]
kevina: I stack is a very simple data structure.  To make reusing another implementation worth it go would need to support generics.  Also this stack is somewhat specialized with the `popEmpty` method.

unixfs/hamt/fetcher.go
```diff
@@ -0,0 +1,297 @@
...
+			if want != nil {
+				j := f.jobs[want]
+				if j.res.vals != nil {
+					f.mainLoopSendResult(j)
+					want = nil
+				}
+			}
+			log.Infof("fetcher: batch job done")
+			f.mainLoopLogStats()
+		case <-f.ctx.Done():
```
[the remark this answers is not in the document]
kevina: This should now be fixed.

unixfs/hamt/fetcher.go
```diff
@@ -0,0 +1,297 @@
...
+	for !f.todo.empty() && len(bj.cids)+len(f.todo.top().cids) <= batchSize {
+		j := f.todo.pop()
+		bj.add(j)
+	}
+
+	if len(bj.cids) == 0 {
+		if !f.idle {
+			log.Infof("fetcher: entering idle state: no more jobs")
+		}
+		f.idle = true
```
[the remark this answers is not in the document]
kevina: Possibly.  Although the current code has the limitation that there is only one batch job running at a time.  This can cause stragglers in a batch job to reduce throughput.  One ideas I explored earlier was to launch a new batch job when an existing job was half done.  I rejected this for now because the logic become to complicated.  (For example if we unconditional launch a new job when a batch is half done we might end up with a a large number of outstating jobs, we could limit the jobs also, but that complicated the logic even more.).  I would still like to explore this option at a later date.  I think I will change `idle` to `numBatchJobs` and make it clear that in the current iteration it is either 0 or 1, but at some point this could change.

unixfs/hamt/hamt.go
```diff
@@ -399,30 +492,33 @@ func (ds *Shard) EnumLinks(ctx context.Context) ([]*ipld.Link, error) {
...
+	ctx, cancel := context.WithCancel(ctx0)
+	fetcher := startFetcher(ctx, ds.dserv)
+	err := ds.walkTrie(ctx, func(sv *shardValue) error {
 		lnk := sv.val
 		lnk.Name = sv.key
 
 		return f(lnk)
-	})
+	}, fetcher)
+	cancel()
```
whyrusleeping: why not just defer cancel?

kevina: Fixed.  (I didn't initially because I didn't see the need, but I guess there is no harm in doing it)

unixfs/hamt/fetcher.go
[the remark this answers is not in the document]
kevina: On second thought I think for this p.r. I will leave things as they are. I have reviewed the `idle` logic and am fairly certain it is correct.

kevina:
Okay.  I think this should be ready to go.  It creates a noticeable improvement so it is a good first step and the changes are fairly self contained.

My prefetcher is rather specialized and that is by design.  A dump prefecher will not work in this case because when retrieving a directly listing you are interested in the shards for the directory not the entire contents.

I considered prefetching the blocks and then just discarding the results with the assumption that they will be in the blockstore later.  That might make things easier and I am not against that route.  The only thing is there is a slight chance the blocks will be gc before they are needed, but I admit this is small enough to not worry about it.

Part of the complexity in my design is that I attempt to retrieve the shards in a order that will eventually allow streaming the directory listing.  I am against any design that does not do this.

unixfs/hamt/fetcher.go
```diff
@@ -0,0 +1,321 @@
...
+	// launch job
+	log.Infof("fetcher: starting batch job, size = %d", len(bj.cids))
+	f.idle = false
+	go func() {
+		ch := f.dserv.GetMany(f.ctx, bj.cids)
+		fetched := result{vals: make(map[string]*Shard)}
+		for no := range ch {
+			if no.Err != nil {
+				fetched.errs = append(fetched.errs, no.Err)
+				continue
```
magik6k: Any reason to not abort early here (close context and return)?

unixfs/hamt/fetcher.go
```diff
@@ -0,0 +1,321 @@
...
+			}
+			hamt, err := NewHamtFromDag(f.dserv, no.Node)
+			if err != nil {
+				fetched.errs = append(fetched.errs, err)
+				continue
```
magik6k: And here (abort early)

unixfs/hamt/fetcher.go
kevina: Because it's unnecessary.  The fetcher can continue even if it encounters an error here.

unixfs/hamt/fetcher.go
```diff
@@ -0,0 +1,321 @@
...
+}
+
+// batchSize must be at least as large as the largest number of CIDs
+// requested in a single job. For best performance it should likely be
+// slightly larger as jobs are popped from the todo stack in order and a
+// job close to the batchSize could force a very small batch to run.
+// The recommend minimum size is thus a size slightly larger than the
+// maximum number children in a HAMT node (which is the largest number
+// of CIDs that could be requested in a single job) or 256 + 64 = 320
+const batchSize = 320
```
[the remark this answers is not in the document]
kevina: I can do that, but I already did to arrive at this value.

unixfs/hamt/fetcher.go
```diff
@@ -0,0 +1,321 @@
...
+
+//
+// fetcher public interface
+//
+
+// startFetcher starts a new fetcher in the background
+func startFetcher(ctx context.Context, dserv ipld.DAGService) *fetcher {
+	log.Infof("fetcher: starting...")
```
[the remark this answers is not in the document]
kevina: I would rather keep this at the same level of the rest of the stats output for consistency.

unixfs/hamt/fetcher.go
magik6k: From what I can see, at least in https://github.com/ipfs/go-ipfs/pull/4979/files#diff-689271378656b0bd9fc790b4a0a2b784R393 we throw the result away if there are errors, so it would make sense to not fetch more stuff if we know it won't be used (I might be wrong here though)

magik6k:
>> Instead of using go logger I'd use go-log

> am I missing something?

What I meant is that it would probably be better / more useful to use logging spans here (i.e. https://github.com/ipfs/go-log/blob/5dc2060baaf8db344f31dafd852340b93811d03f/log.go#L63-L77)

unixfs/hamt/fetcher.go
kevina: Note that code code also aborts the context so the amount of extra work done is minimal.

unixfs/hamt/fetcher.go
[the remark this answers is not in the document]
whyrusleeping: @kevina is that benchmarking reproducible? Do you have scripts we can run to try it out and rederive the number ourselves?

unixfs/hamt/fetcher.go
```diff
@@ -0,0 +1,321 @@
...
+
+type job struct {
+	id  *Shard
+	idx int /* index in the todo stack, an index of -1 means the job
+	   is already done or being worked on now */
+	cids []*cid.Cid
+	res  result
+}
+
+func (f *fetcher) mainLoop() {
```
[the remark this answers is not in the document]
whyrusleeping: As this is new code, and not a refactor, i'd prefer getting it a bit cleaner before merging it in. @kevina If you'd like to hack on other things, maybe @magik6k could help out cleaning things up?

unixfs/hamt/fetcher.go
```diff
@@ -0,0 +1,321 @@
...
+	var want *Shard /* want is the job id (the parent shard) of the
+	   results we want next */
+	f.start = time.Now()
+	for {
+		select {
+		case id := <-f.requestCh:
+			if want != nil {
```
whyrusleeping: seems like we could just pull most of the code in this select case into a separate function, making this flow easier to read and reason about (also removing the need for continues)

unixfs/hamt/fetcher.go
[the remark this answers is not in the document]
kevina: I also don't necessary agree it will be an improvement, but see my other comments.

unixfs/hamt/fetcher.go
kevina: Okay.  I agree this will be helpful.  I will do this later today.

unixfs/hamt/fetcher.go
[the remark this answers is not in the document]
kevina: Actually.  That number is based more on reason (as outlined on comments) then benchmarks.   I do not recommend we go below that size.  Larger values might be beneficial at the increase of resource usage.  I don't have any formal benchmarks, I mostly just observed the behavior when fetching a large directory.  I can create an issue to consider increasing the value if desirable.

unixfs/hamt/fetcher.go
kevina: Okay done.

kevina:
> What I meant is that it would probably be better / more useful to use logging spans here (i.e. https://github.com/ipfs/go-log/blob/5dc2060baaf8db344f31dafd852340b93811d03f/log.go#L63-L77)

I am not at all familiar with that code or the proper usage of it.

kevina:
@Stebalien do you now consider this p.r dead.  The last time this got some activity I was in the middle of base32 work so I didn't want to spend a lot of time on this and the remaining reviews I saw more as nits then anything else.  My apologizes if this caused this P.R. to get dropped.

Stebalien:
It was a combination of everyone being too busy to give a careful review this patch being complicated enough to make such a review doable. It's also so targeted at one use-case (sharded directories) that the it's unclear if the burden of maintaining all of this machinery outweighs the benefit. It may turn out that we can't do any better but, for now, @hannahhoward has picked up this work and is starting with some simpler approaches. Eventually, I'd *like* to have a more generalized pre-fetcher (that can prefetch a DAG up to some depth).

For now, just focus on the base32 stuff. That's absolutely critical for the browser folks.

hannahhoward:
closing per previous comments and https://github.com/ipfs/go-unixfs/pull/19 being merged

PR #4991, opened 2018-05-01
plugin: preload git plugin by default

magik6k:
Suggested by @lgierth an IRC, enabling the git plugin is rather painful experience for anyone wanting to do anything git related:

* Linux users have to setup Go workspace, compile the plugin and then put in a right directory
* All the others (osx/windows/*) have to setup Go workspace, unncomment ipldgit line in preload file and compile/install go-ipfs (bonus pain points when using something like ipfs-desktop)

Adding the plugin to preload_list adds about 100K to the binary size, and given that it is 43M it's rather negligible.


lgierth:
> Adding the plugin to preload_list adds about 100K to the binary size, and given that it is 43M it's rather negligible.

Being able to do Git things is absolutely worth the 100KB :)

whyrusleeping:
The sha1 object thing by default bothers me slightly. Maybe we should just put efforts into getting the plugins up on dists?

lgierth:
Maybe let's couple allowing sha1 to git CIDs only then. We'll have the same issue with js-ipfs and ipfs-companion.

PR #5001, opened 2018-05-06
Fix typo of sudo install command

edisonlee55:
Fix typo of sudo install command.


whyrusleeping:
Thanks @edisonlee55! That was probably pretty confusing for a bunch of new users...

PR #5007, opened 2018-05-09
routing fixes/refactor

### LINKED ISSUE #4749: IPNS republisher should not rely on DHT to store (possibly expired) records

### LINKED ISSUE #4954: Tell DHT nodes if their IPNS records are out of date
Due to https://github.com/ipfs/go-ipfs/pull/4733, we call `GetValues` instead of `GetValue`. Unfortunately, this means that the DHT never learns which value is correct so it never corrects peers with outdated values.

(also, the new `getValue` function on the `routingResolver` duplicates a lot of code in `dht.GetValue`)

Stebalien:
* Cleans up namesys a bit (don't use a map for fields, avoid casting).
* Update for routing system refactor (replace GetValues with a more flexible
  GetValue).
* Don't expect the DHT to keep outdated IPNS records.

fixes #4749
fixes #4954

Note: I'm not entirely happy with this patch but it's good enough for now.
Eventually, I'd prefer to pass the IpnsPublisher to the Republisher as I'm not a
fan of having the Republisher dig through the IpnsPublisher's datastore entries.

However, this unblocks us.

Depends on:

* [x] https://github.com/libp2p/go-libp2p-kad-dht/pull/148
* [x] https://github.com/libp2p/go-libp2p-routing/pull/21
* [x] https://github.com/libp2p/go-libp2p-record/pull/17
* [x] https://github.com/ipfs/distributions/pull/190
* [x] https://github.com/ipfs/fs-repo-migrations/pull/72
* [x] https://github.com/ipfs/go-ipfs-routing/pull/5
* [x] [go-libp2p-pusbub-router](https://github.com/ipfs/go-libp2p-pubsub-router) -- extraction of the pubsub router.
* [x] [go-libp2p-routing-helpers](https://github.com/libp2p/go-libp2p-routing-helpers) -- helpers for composing multiple routers. Somewhat complicated but it has >97% test coverage.


namesys/publisher.go
```diff
@@ -12,103 +14,196 @@ import (
...
+func IpnsDsKey(id peer.ID) ds.Key {
+	return ds.NewKey("/ipns/" + base32.RawStdEncoding.EncodeToString([]byte(id)))
+}
+
+// PublishedNames returns the latest IPNS records published by this node and
+// their expiration times.
+//
+// This method will not search the routing system for records published by other
+// nodes.
+func (p *IpnsPublisher) ListPublished(ctx context.Context) (map[peer.ID]*pb.IpnsEntry, error) {
```
Stebalien: Should probably have a test. This isn't actually used right now but I'd like to expose it via a command (and use it from the republisher).

namesys/publisher.go
```diff
@@ -12,103 +14,196 @@ import (
...
+		}
+	}
+}
+
+// GetPublished returns the record this node has published corresponding to the
+// given peer ID.
+//
+// If `checkRouting` is true and we have no existing record, this method will
+// check the routing system for any existing records.
+func (p *IpnsPublisher) GetPublished(ctx context.Context, id peer.ID, checkRouting bool) (*pb.IpnsEntry, error) {
```
Stebalien: `checkRouting`... (bleh)

test/sharness/t0090-get.sh
```diff
@@ -20,7 +20,7 @@ test_ipfs_get_flag() {
   '
 
   test_expect_success "ipfs get $flag output looks good" '
-    printf "%s\n\n" "Saving archive to $HASH$ext" >expected &&
```
Stebalien: This seems strictly better but is technically a breaking change. Any idea why this has changed?

namesys/publisher.go
```diff
@@ -12,103 +14,196 @@ import (
...
 	}
 
-	// increment it
-	seqnum++
-
-	return PutRecordToRouting(ctx, k, value, seqnum, eol, p.routing, id)
-}
+	seqno := rec.GetSequence() // returns 0 if rec is nil
+	if rec != nil && value != path.Path(rec.GetValue()) {
+		// Don't bother incriminating the sequence number unless the
```
dirkmc: incriminating? :)

Stebalien: aspell has a surprisingly small english dictionary...

Stebalien: @kevina you are! Awesome! Will do.

core/core.go
```diff
@@ -437,17 +427,50 @@ func (n *IpfsNode) HandlePeerFound(p pstore.PeerInfo) {
...
 
+	if ipnsps {
+		n.PSRouter = psrouter.NewPubsubValueStore(
+			ctx,
+			host,
+			n.Routing,
+			n.Floodsub,
+			validator,
+		)
+		n.Routing = rhelpers.Tiered{
```
whyrusleeping: :+1:

kevina:
What is the status of this?  Is it likely to go in mostly as is?

Just FYI: I have not looked at this closely but this looks like it might be making API changes to the routing code.  Some additional changes may be required to support sym-links.

Stebalien:
> What is the status of this? Is it likely to go in mostly as is?

That's my plan. I'm almost done with the migration but that's the only thing left.

> Just FYI: I have not looked at this closely but this looks like it might be making API changes to the routing code. Some additional changes may be required to support sym-links.

Symlinks? This should only be touching namesys/routing code (not unixfs path resolution or anything that should handle symlinks). Am I missing something?

kevina:
> Symlinks? This should only be touching namesys/routing code (not unixfs path resolution or anything that should handle symlinks). Am I missing something?

The current routing API makes adding support for symlinks difficult, espacally the handling of `..`

Stebalien:
Does this patch make something worse?

kevina:
I have not looked closely but I don't think so.  Since this should hopefully 
go in soon I will wait until it does before I tackle the API changes needed 
for symbolic links.

whyrusleeping:
there shouldnt be any shared code between routing and symlinks... You might be thinking of the path resolver.

namesys/publisher.go
```diff
@@ -12,103 +14,196 @@ import (
...
+			return nil, fmt.Errorf("found ipns record that we couldn't convert to a value")
+		}
+	case ds.ErrNotFound:
+		if !checkRouting {
+			return nil, nil
+		}
+		_, ipnskey := IpnsKeysForID(id)
+		value, err = p.routing.GetValue(ctx, ipnskey)
+		if err != nil {
+			// Not found or other network issue. Can't really do
```
whyrusleeping: dropping errors makes me nervous, would appreciate maybe a debug or info log of this

Stebalien: Addressed.

Stebalien:
DO NOT MERGE UNTIL: https://github.com/ipfs/infrastructure/issues/408

(well, actually, it's not *that* big of an issue as go-ipfs refers to that dist path by *hash*).

PR #5008, opened 2018-05-09
ipfs release 0.4.15

whyrusleeping:
License: MIT


djdv:
My fault on that one.
If there's any interest, I can patch `tar-utils` to build on versions before Go 1.10, but have no plans to do so if it's not requested.

PR #5010, opened 2018-05-09
Efficient "repo stat" (DiskUsage) and "--size-only" flag

hsanjuan:
This makes use of the PersistentDatastore DiskUsage method to obtain the Repo's storage usage (GetStorageUsage()).
    
Additionally, the --size-only flag has been added to the "ipfs repo stat" command. This avoids counting the number of objects in the repository and returns faster.

Overseeds #4550 

I had to make some gx magic to fix dependencies and be able to use latest go-datastore. go-libp2p-kad-dht depends on it and afaik the changes there cannot be integrated still etc. I'm happy to leave this hanging until this is sorted out. Thus, tagging this as blocked.

### DIFF
--- a/core/commands/repo.go
+++ b/core/commands/repo.go
@@ -150,13 +150,19 @@ var repoStatCmd = &cmds.Command{
 	Helptext: cmdkit.HelpText{
 		Tagline: "Get stats for the currently used repo.",
 		ShortDescription: `
-'ipfs repo stat' is a plumbing command that will scan the local
-set of stored objects and print repo statistics. It outputs to stdout:
+'ipfs repo stat' provides information about the local set of 
+stored objects. It outputs:
+
+RepoSize        int Size in bytes that the repo is currently taking.
+StorageMax      string Maximum datastore size (from configuration)
 NumObjects      int Number of objects in the local repo.
 RepoPath        string The path to the repo being currently used.
-RepoSize        int Size in bytes that the repo is currently taking.
 Version         string The repo version.
 `,
+	},
+	Options: []cmdkit.Option{
+		cmdkit.BoolOption("size-only", "Only report RepoSize and StorageMax."),
+		cmdkit.BoolOption("human", "Output sizes in MiB."),
 	},
 	Run: func(req *cmds.Request, res cmds.ResponseEmitter, env cmds.Environment) {
 		n, err := GetNode(env)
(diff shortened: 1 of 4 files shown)


hsanjuan:
@whyrusleeping is that the case now?

whyrusleeping:
@hsanjuan yeah, #5007 merged

hsanjuan:
Now rebased on master. Thanks @Stebalien for bubbling up deps on some other commit.

core/commands/repo.go
```diff
@@ -203,9 +215,14 @@ Version         string The repo version.
 					fmt.Fprintf(wtr, "StorageMax:\t%d\n", stat.StorageMax)
 				}
 			}
+
+			if sizeOnly {
+				return nil
+			}
+
+			fmt.Fprintf(wtr, "NumObjects:\t%d\n", stat.NumObjects)
```
Stebalien: This is technically a breaking change but I doubt anyone is relying on this order.

core/corerepo/stat.go
```diff
@@ -45,11 +46,29 @@ func RepoStat(n *core.IpfsNode, ctx context.Context) (*Stat, error) {
...
+		NumObjects: count,
+		RepoSize:   sizeStat.RepoSize,
+		StorageMax: sizeStat.StorageMax,
+		RepoPath:   path,
+		Version:    fmt.Sprintf("fs-repo@%d", fsrepo.RepoVersion),
+	}, nil
+}
+
+// RepoSize returns a *Stat object with the RepoSize and StorageMax fields set.
+func RepoSize(ctx context.Context, n *core.IpfsNode) (*Stat, error) {
```
Stebalien: This feels a bit funky... what about:

```go
type Size struct {
	RepoSize   uint64
	StorageMax uint64
}

type Stat struct {
	StorageSize
	NumObjects  uint64
	RepoPath    string
	Version     string
}
```

And then have `RepoSize` return `Size` (and, really, these functions should return be returning these by value). This would also avoid sending zero values for the missing fields.

Thoughts?

core/commands/repo.go
hsanjuan: I hope not. I can turn it around though and leave NumObjects as first line.

core/corerepo/stat.go
hsanjuan: yes this is better

core/commands/repo.go
hsanjuan: I'll just change it. It's marginally better to not change the "API".

hsanjuan:
@Stebalien I have addressed the comments. However, since the command output only supports a single type, `SizeStat` needs to eventually become `Stat`. The final result is 0-valued json fields when doing a request with --size-only. (this is ok for me). If we wanted to return json without those fields, we would need a second command (`ipfs repo size`).

core/commands/repo.go
```diff
@@ -150,65 +150,85 @@ var repoStatCmd = &cmds.Command{
...
-		cmdkit.BoolOption("human", "Output RepoSize in MiB."),
+		cmds.EmitOnce(res, &stat)
 	},
-	Type: corerepo.Stat{},
+	Type: &corerepo.Stat{},
 	Encoders: cmds.EncoderMap{
 		cmds.Text: cmds.MakeEncoder(func(req *cmds.Request, w io.Writer, v interface{}) error {
 			stat, ok := v.(*corerepo.Stat)
 			if !ok {
+				fmt.Println("adios")
```
Stebalien: Debug print statement.

hsanjuan: damn

hsanjuan: fixed

hsanjuan:
@whyrusleeping can we merge this please?

Stebalien:
@hsanjuan once we do the release which should happen any day now.

PR #5016, opened 2018-05-12
Fix goroutine leak in dag put

alecbrick:
In the unlikely case that the context is cancelled, and we're adding more than 8 files.
Part of #4414.


whyrusleeping:
Thanks!

PR #5031, opened 2018-05-21
Fix build on DragonFlyBSD

daftaupe:
License: MIT
Signed-off-by: Pierre-Alain TORET <pierre-alain.toret@protonmail.com>


djdv:
I'm curious if we can just omit these constraints altogether, so new platforms don't have to keep rediscovering this issue when trying to build on unsupported platforms.
To my knowledge, these files are only built when supplying the `nofuse` build tag anyway and the implementation for `nofuse` should work across all go implementations. (it seems to just return a string/error). Further limiting via `GOOS` seems unnecessary.

Kubuxu:
@djdv please open issue about it, it will get lost here.

PR #5033, opened 2018-05-23
build: remove fuse platform constraints

djdv:
Followup to: https://github.com/ipfs/go-ipfs/pull/5031
@daftaupe was expressing issues on IRC when trying to build `go-ipfs` on DragonflyBSD. After investigating, it seemed like the only problem was in the build constraints. I suggested adding dragonfly to the list of build targets. This worked fine but may be unnecessary (see: https://github.com/ipfs/go-ipfs/pull/5031#issuecomment-390706466)

It may be best to just remove these constraints so people don't have to rediscovering this and make source modifications when using the `install_unsupported` `make` target on platforms that are not explicitly defined by us.

### DIFF
--- a/core/commands/mount_nofuse.go
+++ b/core/commands/mount_nofuse.go
@@ -1,5 +1,4 @@
-// +build linux darwin freebsd netbsd openbsd dragonfly
-// +build nofuse
+// +build !windows,nofuse
 
 package commands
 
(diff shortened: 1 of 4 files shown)


djdv:
Also `goimports` moved a dep into the right place. I can remove that change if desired.

daftaupe:
@whyrusleeping, @djdv patch is building fine, at least on my DragonFlyBSD server :)

whyrusleeping:
cool, i guess let's merge this one then. @djdv mind rebasing?

djdv:
@whyrusleeping 
Should be good to go now

PR #5035, opened 2018-05-24
update for the go-ipfs-cmds refactor

keks:
This branch contains changes to go-ipfs so we can use it with this branch of go-ipfs-cmds: https://github.com/ipfs/go-ipfs-cmds/pull/112

### DIFF
--- a/commands/legacy/response.go
+++ b/commands/legacy/response.go
@@ -7,8 +7,8 @@ import (
 	"reflect"
 	"sync"
 
-	"gx/ipfs/QmPTfgFTo9PFr1PvPKyKoeMgBvYPh6cX3aDP7DHKVbnCbi/go-ipfs-cmds"
 	"gx/ipfs/QmSP88ryZkHSRn1fnngAaV2Vcn63WUJzAavnRM9CVdU1Ky/go-ipfs-cmdkit"
+	"gx/ipfs/QmZVPuwGNz2s9THwLS4psrJGam6NSEQMvDTaaZgNfqQBCE/go-ipfs-cmds"
 
 	oldcmds "github.com/ipfs/go-ipfs/commands"
 )
@@ -34,7 +34,10 @@ func (rw *responseWrapper) Output() interface{} {
 		// get first emitted value
 		x, err := rw.Next()
 		if err != nil {
-			return nil
+			ch := make(chan interface{})
+			log.Error(err)
+			close(ch)
+			return (<-chan interface{})(ch)
 		}
 		if e, ok := x.(*cmdkit.Error); ok {
 			ch := make(chan interface{})
@@ -120,16 +123,13 @@ func (r *fakeResponse) Send(errCh chan<- error) {
 	defer close(errCh)
 
 	out := r.Output()
-	if out == nil {
+
+	// don't emit nil or Single{nil}
+	if out == nil || out == (cmds.Single{Value: nil}) {
 		return
 	}
 
-	if ch, ok := out.(chan interface{}); ok {
-		out = (<-chan interface{})(ch)
-	}
-
-	err := r.re.Emit(out)
-	errCh <- err
+	errCh <- r.re.Emit(out)
 	return
 }
 
@@ -141,7 +141,7 @@ func (r *fakeResponse) Request() oldcmds.Request {
 // SetError forwards the call to the underlying ResponseEmitter
 func (r *fakeResponse) SetError(err error, code cmdkit.ErrorType) {
 	defer r.once.Do(func() { close(r.wait) })
-	r.re.SetError(err, code)
+	r.re.CloseWithError(cmdkit.Errorf(code, err.Error()))
 }
 
 // Error is an empty stub
--- a/commands/legacy/response.go
+++ b/commands/legacy/response.go
@@ -8,7 +8,7 @@ import (
 	"sync"
 
 	"gx/ipfs/QmSP88ryZkHSRn1fnngAaV2Vcn63WUJzAavnRM9CVdU1Ky/go-ipfs-cmdkit"
-	"gx/ipfs/QmZVPuwGNz2s9THwLS4psrJGam6NSEQMvDTaaZgNfqQBCE/go-ipfs-cmds"
+	"gx/ipfs/QmSwR1QndLsdnw2mpW2rrywoCUPeY7o8UxtK1dLpPCNFGD/go-ipfs-cmds"
 
 	oldcmds "github.com/ipfs/go-ipfs/commands"
 )
--- a/commands/legacy/response.go
+++ b/commands/legacy/response.go
@@ -8,7 +8,7 @@ import (
 	"sync"
 
 	"gx/ipfs/QmSP88ryZkHSRn1fnngAaV2Vcn63WUJzAavnRM9CVdU1Ky/go-ipfs-cmdkit"
-	"gx/ipfs/QmSwR1QndLsdnw2mpW2rrywoCUPeY7o8UxtK1dLpPCNFGD/go-ipfs-cmds"
+	"gx/ipfs/QmdsFzGmSLMQQaaPhcgGkpDjPocqBWLFA829u6iMv5huPw/go-ipfs-cmds"
 
 	oldcmds "github.com/ipfs/go-ipfs/commands"
 )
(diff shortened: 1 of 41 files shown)


keks:
@Stebalien sharness test `t0272-urlstore.sh`fails and I'm not sure how to fix it.

The issue is that after adding a file through the urlstore, the subsequent `get` call will successfully fetch everything and then error with `failed to fetch all nodes`, so it seems like some if the DAG nodes are not properly added. However, the resulting file matches the file that was added (i.e. `file3` matches `file3.actual` in the trash directory). I traced the error through the code and it is returned by [`res.Emit(reader)`](https://github.com/ipfs/go-ipfs/blob/feat/cmds2/core/commands/get.go#L92), where `reader` is a `*"unixfs/io".PBDagReader`.

The furthest I have traced the error is that it is [returned by a promise](https://github.com/ipfs/go-unixfs/blob/master/io/pbdagreader.go#L120) in the `PBDagReader`, and now I'm somewhat lost because that can only return either an error or the data, and we have all the data, so where does the error come from? I was hoping it was some race condition and the nodes are not written yet when the call to `get` comes in, but sleeping for a second before doing that doesn't help either.

I'm pretty puzzled and meandering through unfamiliar code, so I thought you might have an idea where that issue might come from.

Stebalien:
> I'm pretty puzzled and meandering through unfamiliar code, so I thought you might have an idea where that issue might come from.

The answer is: not your bug, your patch just makes us correctly report errors. I'll look into it.

keks:
That did the trick, thanks! :rainbow:

keks:
@Stebalien I removed the remaining debug output and rebased to master. This is now ready for review.

core/commands/stat.go
```diff
@@ -137,49 +130,44 @@ Example:
...
 			}
 			if !doPoll {
-				return
+				return nil
 			}
 			select {
 			case <-time.After(interval):
 			case <-req.Context.Done():
-				return
+				break
```
Stebalien: `return req.Context.Err()`?

core/commands/urlstore.go
```diff
@@ -111,14 +104,14 @@ time.
 		}
 		root, err := layout(dbp.New(chk))
 		if err != nil {
-			res.SetError(err, cmdkit.ErrNormal)
-			return
+			return err
 		}
 
-		cmds.EmitOnce(res, BlockStat{
+		err = cmds.EmitOnce(res, &BlockStat{
```
Stebalien: Can return directly.

test/sharness/lib/test-lib.sh
```diff
@@ -26,6 +26,7 @@ fi
 # it's too late to pass in --verbose, and --verbose is harder
 # to pass through in some cases.
 test "$TEST_VERBOSE" = 1 && verbose=t
+test "$TEST_IMMEDIATE" = 1 && immediate=t
```
Stebalien: ?

Stebalien:
@keks: This has been unblocked. Once this is finished, we can merge https://github.com/ipfs/go-ipfs-cmds/pull/112 and ship it.

keks:
@Stebalien The sharness test t0252 consistently fails:
```
expecting success: 
  ipfs refs $ADIR_HASH &&
  ipfs repo gc &&
  ipfs refs $ADIR_HASH

Error: failed to fetch all nodes
not ok 6 - gc okay after adding incomplete node
#	
#	  ipfs refs $ADIR_HASH &&
#	  ipfs repo gc &&
#	  ipfs refs $ADIR_HASH
```

I can't tell whether this is because of something else of my changes. I don't really see a connection. Also, sometimes the robust repo gc (t0087) fails sometimes, but I got used to that one.

Stebalien:
The test was broken. I've fixed it.

commands/legacy/response.go
```diff
@@ -120,16 +123,13 @@ func (r *fakeResponse) Send(errCh chan<- error) {
...
 
 	out := r.Output()
-	if out == nil {
-		return
-	}
 
-	if ch, ok := out.(chan interface{}); ok {
-		out = (<-chan interface{})(ch)
+	// don't emit nil or Single{nil}
+	if out == nil || out == (cmds.Single{Value: nil}) {
```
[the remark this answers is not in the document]
keks: The reason I did this is that I had some command sometimes emit a nil at the beginning. This led to the http response emitter call the preamble with nil, which led to wrong state in the http response. I don't see a reason to actually send nils, so I figured it's easiest to just ignore them. I don't remember which command that was, though.

Do you think this is wrong?

core/commands/filestore.go
```diff
@@ -49,71 +49,61 @@ The output is:
...
-
-					if r.ErrorMsg != "" {
-						errors = true
-						fmt.Fprintf(os.Stderr, "%s\n", r.ErrorMsg)
-					} else {
-						fmt.Fprintf(os.Stdout, "%s\n", r.FormatLong())
-					}
+				r, ok := v.(*filestore.ListRes)
+				if !ok {
+					// TODO or just return that error? why didn't we do that before?
```
keks: @Stebalien What do you think should we do here?

core/commands/stat.go
keks: That would return an error and non-zero exit code to the user, even though everything was in order. Do you think this is a good idea?

test/sharness/lib/test-lib.sh
keks: Oh yeah, if you want we can remove this, but in that case I'd open a new PR for this.

When running single tests, I usually run them with `./txxxx-foo-bar.sh -v -i`, which results in the test aborting immediately after an error. This makes finding the failing test in a large test file a lot easier. For some reason this behaviour could not be set using the env, meaning that it could not be used when invoked with `make`. This line allows enabling this mode using `TEST_IMMEDIATE=1 make`.

core/commands/stat.go
Stebalien: Ah, I see. You're right.

commands/legacy/response.go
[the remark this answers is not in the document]
Stebalien: I'm worried we should be *closing* the output (if we're emitting a `Single`). However, this is probably fine (we'll close anyways).

test/sharness/lib/test-lib.sh
Stebalien: Got it, sounds useful, let's keep it (although we should probably document it in the sharness readme).

core/commands/filestore.go
Stebalien: Yes. I'm not sure why we didn't.

core/commands/repo.go
```diff
@@ -99,7 +99,7 @@ order to reclaim hard disk space.
 					}
 				}
 				if errs {
-					res.SetError(fmt.Errorf("encountered errors during gc run"), cmdkit.ErrNormal)
+					outChan <- &GcResult{Error: "encountered errors during gc run"}
```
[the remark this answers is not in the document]
keks: Yeah, not sure what I was thinking here...

keks:
> Fixing the gc command should fix travis. I'm not sure why jenkins didn't fail...

As far as I can see the issue is that it doesn't report all gc errors (some of the `grep`s fail). That doesn't mean we shouldn't use `SetError` for the final error (we should), but I believe there is another error here, but I doubt it's in the commands library. I also had that test fail in other branches when reviewing PRs, so I'm confident it has nothing to do with this change set. I just ran the test with the race detector on and it didn't report anything :man_shrugging:

keks:
Looks like making `repo gc` use cmds 2 fixed the instability. @Stebalien I think I addressed all your concerns, want to merge this?

Stebalien:
@keks LGTM but I'd like @magik6k to take a look.

@magik6k Any feedback and/or objections to these changes? I've done a pretty thorough code review (don't feel like you need to do a line-by-line audit) but I'd like to get a final sign-off from you before we actually merge this into go-ipfs.

Summary:

1. We now *return* errors instead of using `SetError`. This alone ends up removing a bunch of code.
2. We've removed support for multiple "error" responses. This was never really properly supported in, e.g., javascript, but go *technically* supported it.
3. We've made a bunch of low-level changes to reduce race conditions in the commands lib (esp around error handling).

Stebalien:
:tada: :tada: :tada:

keks:
:tada: :confetti_ball:

Issue #3799, opened 2017-03-19
ERROR    bitswap: Error writing block to datastore: sync /ipfs/blocks: invalid argument bitswap.go:303


starrychloe:
```
20:35:43.128 ERROR    bitswap: Error writing block to datastore: sync /ipfs/blocks: invalid argument bitswap.go:303
20:35:43.922 ERROR  providers: error adding new providers: sync /ipfs/datastore: invalid argument providers.go:258
20:35:45.664 ERROR  providers: error adding new providers: sync /ipfs/datastore: invalid argument providers.go:258
```
and it just repeats...

#### Version information:
```
[vagrant@localhost go-ipfs]$ ./ipfs version --all
go-ipfs version: 0.4.6-
Repo version: 5
System version: amd64/linux
Golang version: go1.8
```

#### Type:
Error during running. Must be a bug somewhere.

#### Priority:
P4

#### Description:
IPFS daemon leaves errors in the console as it runs. I don't know what causes it, but clicking on big, old IPFS links seems to cause it. I can view `/ipfs/datastore/` and `/ipfs/blocks` fine. I can access the cat pic http://127.0.0.1:8080/ipfs/QmW2WQi7j6c7UgJTarActp7tDNikE4B2qXtFCfLPdsgaTQ/cat.jpg fine. These links seem to cause the error to print:
http://127.0.0.1:8080/ipfs/QmcYKHpcBJNL9uaxxTZJGuk9QYBfS3cDuYKoELPhfdmRqn
http://127.0.0.1:8080/ipfs/QmbzpQmRULESivRvpP9SZL2vsu1dGkmczwVKCWzD2oXdMP
http://127.0.0.1:8080/ipfs/QmRcsML93Bwi8V9DiGhX4Zs9ypDgHSCWCpAMxd7cpHcS1f

whyrusleeping:
Invalid argument is a new one. 

After a small amount of research i see that some filesystems (particularly on virtualbox for some reason?) do not support calling fsync on directories. This *might* be a lead here.

What filesystem are you using?

starrychloe:
I'm using vboxsf. 

```
[vagrant@localhost ~]$ mount | grep ipfs
ipfs on /media/sf_ipfs type vboxsf (rw,nodev,relatime)
ipfs on /ipfs type vboxsf (rw,nodev,relatime)
```
The host (Windows 8.1) is using NTFS.

Kubuxu:
[...]

PR #5036, opened 2018-05-26
Warning when NoSync setting is set as it does not turn syncing off…

eginez:
… #3799

### DIFF
--- a/repo/fsrepo/fsrepo.go
+++ b/repo/fsrepo/fsrepo.go
@@ -400,6 +400,9 @@ func (r *FSRepo) openDatastore() error {
 	} else if r.config.Datastore.Spec == nil {
 		return fmt.Errorf("required Datastore.Spec entry missing from config file")
 	}
+	if r.config.Datastore.NoSync {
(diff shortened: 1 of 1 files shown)


repo/fsrepo/fsrepo.go
```diff
@@ -400,6 +400,9 @@ func (r *FSRepo) openDatastore() error {
 	} else if r.config.Datastore.Spec == nil {
 		return fmt.Errorf("required Datastore.Spec entry missing from config file")
 	}
+	if r.config.Datastore.NoSync {
+		log.Warning("NoSync is no longer effective. Set 'sync' to false. See https://github.com/ipfs/go-ipfs/blob/master/docs/datastores.md#flatfs.")
```
whyrusleeping: Hrm... I think the wording might be clearer as:
"NoSync is now deprecated in favor of datastore specific settings. If you want to disable fsync on flatfs 'sync' to false,  See <that link>"

whyrusleeping:
Thanks for getting to the root of the problem :)

eginez:
Hey @whyrusleeping I think this should be ready to be merged

whyrusleeping:
Thanks for the ping! and the PR!

lgierth:
Do I understood the linked issue #3799 right, we ignore Datastore.NoSync?

eginez:
That's correct Datastore.NoSync is ignored

Issue #3999, opened 2017-06-21
Publishing of docker images


victorb:
Currently on Docker Hub, we publish latest builds as `:master` instead of Docker common practice which is `:latest`. Would be good to at least update `latest` as well on new version.

Another good tag to have is `stable` as right now, `master` is pointing to a RC version while `latest` points to an old version.

Structure would be something like this:

- `master`/`latest` -> latest shine bleeding-edge build, RCs and so on
- `stable` -> latest stable version, right now that would be 0.4.9

skorokithakis:
Agreed, I have been having trouble with this. Should there also be per-version tags?

Kubuxu:
There is stable image built from `release` branch and tagged `release`, last time the release branch was not updated to contain 0.4.9 which is a mistake.

skorokithakis:
@Kubuxu Yes, it seems that `release` contains 0.4.8, which is breaking a bunch of my stuff now :(

lgierth:
Yeeah we forgot pushing to the release branch for 0.4.9, doing now

skorokithakis:
Great, thank you!

lgierth:
@VictorBjelkholm we've had the following structure since the beginning:

- latest: current release
- release: current release
- master: current master HEAD

Any strong reason to change that?

skorokithakis:
Not to speak for him, but I think Victor's point was more "let's make sure that the tags are always updated" rather than to mandate a tag scheme, really, and I would +1 that.

For me, any scheme is fine, as long as it's consistently kept up to date.

lgierth:
It's already in the release checklist, it was apparantly just forgotten for 0.4.9 :)

skorokithakis:
Ah, great. For me, the biggest UX failure was that search engines return a different Docker repo first, so that led to much confusion for me. If that one can either be removed or at least changed to a note that says "please use this new one", it would save people lots of pain.

lgierth:
Oh mh, the jbenet/go-ipfs image... I've never quite figured out how to properly mirror the ipfs/go-ipfs builds to jbenet/go-ipfs...

skorokithakis:
I don't think you need to mirror (and probably shouldn't), just change the Readme there to the text "please use this other repo" and a link. That would mostly solve the issue, I think.

lgierth:
The idea was to leave existing scripts and deployments working. With just a changed readme, nobody actually notices. We need to either continue maintaining it (preferred), or delete it.

[...]

PR #5043, opened 2018-05-29
Add Docker step to Pre-Release Checklist

lidel:
Suggested in https://github.com/ipfs/go-ipfs/issues/3999#issuecomment-392748363 to avoid problem described in https://github.com/ipfs/go-ipfs/issues/3999#issuecomment-392747159

cc @whyrusleeping @VictorBjelkholm

### DIFF
--- a/docs/releases.md
+++ b/docs/releases.md
@@ -28,6 +28,7 @@ will be allowed up to release day.
 - [ ] publish dist.ipfs.io
 - [ ] publish next version to https://github.com/ipfs/npm-go-ipfs
 - [ ] publish gx version with `gx release`, as per [gx release guidelines](https://github.com/whyrusleeping/gx#publishing-and-releasing)
+- [ ] publish image to [Docker Hub](https://hub.docker.com/r/ipfs/go-ipfs/tags/) and update `latest` and `release` tags to point at it
 
 ## Post-Release
 - [ ] bump version string in `repo/config/version.go` to `vX.Y.Z-dev`


docs/releases.md
```diff
@@ -28,6 +28,7 @@ will be allowed up to release day.
 - [ ] publish dist.ipfs.io
 - [ ] publish next version to https://github.com/ipfs/npm-go-ipfs
 - [ ] publish gx version with `gx release`, as per [gx release guidelines](https://github.com/whyrusleeping/gx#publishing-and-releasing)
+- [ ] publish image to [Docker Hub](https://hub.docker.com/r/ipfs/go-ipfs/tags/) and update `latest` and `release` tags to point at it
```
VictorBjelkholm: The publish should be automatic, but I think it takes a while. We should build image on one machine but then just update the tags. 

I think we can setup the automatic build in Docker Hub to be automatic and tag images based on branches... That way, we can have a `release` branch that would point to the `release` tag in Docker Hub, and you'll only need to update that and push it to update the Docker Hub tags.

lidel: While we are at this: are we able to update or remove the `latest` tag?
IMO it is too ambiguous, and keeping `release` and `master` may be a better idea.
Alternatively it should be automated to point at the same things as `release`

VictorBjelkholm: `latest` is de-facto (or maybe just is) standard in Docker-land, so we should probably keep it. Having `latest` being in sync with `release` I think is what we've agreed before and would map best to how things usually are dealt with. `master` would be the "nightly" basically with all the latest changes.

whyrusleeping:
what is actually involved in publishing an image to dockerhub? updating the release branch on the repo is already part of the process (i just forget sometimes :/ )

olizilla:
As @whyrusleeping said, Dockerhub has been configured to build and tag the docker images when the master and release branches change on the go-ipfs repo.

The intergration is currerntly broken, pending reconnecting github and dockerhub, but that'll get fixed.
https://github.com/ipfs/go-ipfs/pull/5043#issuecomment-493003631

The releaser doesn't have to perform any extra steps, so I'm gonna close this out.

Issue #5044, opened 2018-05-29
ipfs files ls -l doesn't return hash or size when listing files directly


achingbrain:
Listing a directory shows hashes and sizes:

```sh
$ ipfs files ls -l /
data.txt	QmahmHDpSXGKUNxVGre3ZGj9wG9pGwhb691SZdep4gV4XU	142646400
```

But if you list the file directly, extended information is not returned:

```sh
$ ipfs files ls -l /data.txt
data.txt		0
```

schomatis:
From what I understand from reading the files API, there is important information (such as the hash and the size, and I'm not sure if the name also) that doesn't belong to the node that represents the file but rather to the directory that contains it, i.e., that has a link (to the file) where that information is actually stored. This means that there is some information that can only be obtained by querying the directory (AFAIK).

schomatis:
Yes, so scratch most of my last comment, the `files ls` command is not extracting the node to get any information, it's just repeating back the file part of the path passed as argument.

https://github.com/ipfs/go-ipfs/blob/611d572ef23e92af6640f6bde207c3bdc69e48ab/core/commands/files.go#L463-L467

This can be improved, I'll send a PR.

schomatis:
@achingbrain Could you give https://github.com/ipfs/go-ipfs/pull/5045 a try to see how it feels? Keep in mind that the sizes reported by listing the directory and the size might vary by a small amount due to some metadata (ideally those sizes should be unified).

achingbrain:
I must be doing something wrong because I still see 0 for the file size and no hash.

I did:

```
$ cd ~/.gvm/pkgsets/go1.9/global/src/github.com/ipfs/go-ipfs
$ git remote add schomatis https://github.com/schomatis/go-ipfs.git
$ git fetch
$ git checkout fix/ls/dir-size
$ make clean && make install
```

Sorry, I don't know much about go..

Is it going to print out the node size or the sum of the block sizes?

schomatis:
Oh, sorry, the [PR](https://github.com/ipfs/go-ipfs/pull/5045#event-1651124682) I pointed you to is not yet merged (nor it will be in the near future until I figure out the size discrepancies), the commands you're listing sound correct but I would recommend you try the official [GitHub way](https://help.github.com/articles/checking-out-pull-requests-locally/) which don't require to add my repo with the branch but to fetch it indirectly through the PR number (using the original ipfs remote):

```
[...]

PR #5045, opened 2018-05-29
cmd: fix `files ls` to report hash and size for files

schomatis:
Fixes https://github.com/ipfs/go-ipfs/issues/5044.

### DIFF
--- a/core/commands/files.go
+++ b/core/commands/files.go
@@ -462,7 +462,24 @@ Examples:
 			return
 		case *mfs.File:
 			_, name := gopath.Split(path)
-			out := &filesLsOutput{[]mfs.NodeListing{mfs.NodeListing{Name: name, Type: 1}}}
+			out := &filesLsOutput{[]mfs.NodeListing{mfs.NodeListing{Name: name}}}
+			if long {
+				out.Entries[0].Type = int(fsn.Type())
+
+				size, err := fsn.Size()
+				if err != nil {
+					res.SetError(err, cmdkit.ErrNormal)
+					return
+				}
+				out.Entries[0].Size = size
+
+				nd, err := fsn.GetNode()
+				if err != nil {
+					res.SetError(err, cmdkit.ErrNormal)
+					return
+				}
+				out.Entries[0].Hash = nd.Cid().String()
+			}
 			res.SetOutput(out)
 			return
 		default:
--- a/test/sharness/t0250-files-api.sh
+++ b/test/sharness/t0250-files-api.sh
@@ -186,6 +186,12 @@ test_files_api() {
     test_cmp ls_l_expected ls_l_actual
   '
 
+  test_expect_success "file has correct hash and size listed with -l" '
+    echo "file1	$FILE1	4" > ls_l_expected &&
+    ipfs files ls -l /cats/file1 > ls_l_actual &&
+    test_cmp ls_l_expected ls_l_actual
+  '
+
   test_expect_success "can stat file $EXTRA" '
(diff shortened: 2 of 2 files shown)


schomatis:
I'll add some tests (or probably some already existing tests should be fixed).

schomatis:
The test is failing because the reported size varies for reasons I don't quite understand, I'll raise a different issue for that and leave this as blocked in the meanwhile.

Stebalien:
I wonder if we need fo flush first or something...

schomatis:
Now that you mention it, maybe we need to lock the directory, that's what `ForEachEntry` does when it gets the node (that's where I copied the patch from),

https://github.com/ipfs/go-ipfs/blob/1e9e2f453c435272433a96ebfba5c4319cc08534/mfs/dir.go#L262-L274

(but that may just be due to the `childUnsync` call, I can't tell).

Stebalien:
Sorry, I missed that commit before my comment. It looks like you fixed the size issue, right? The code looks corect, at least.

core/commands/files.go
```diff
@@ -462,7 +462,24 @@ Examples:
...
+
+			nd, err := fsn.GetNode()
+			if err != nil {
+				res.SetError(err, cmdkit.ErrNormal)
+				return
+			}
+
+			out := &filesLsOutput{[]mfs.NodeListing{mfs.NodeListing{
+				Name: name,
+				Type: 1,
```
Stebalien: Should probably use `unixfs.TFile` to avoid magic numbers.

schomatis: Yes, I should directly call the `Type()` function, `1` isn't even the actual value of `TFile`.., thanks!

test/sharness/t0250-files-api.sh
```diff
@@ -186,6 +186,12 @@ test_files_api() {
     test_cmp ls_l_expected ls_l_actual
   '
 
+  test_expect_success "file has correct hash and size listed with -l" '
+    echo "file1	$FILE1	4" > ls_l_expected &&
+    ipfs files ls -l /cats/file1 > ls_l_actual &&
+    test_cmp ls_l_expected ls_l_actual
+  '
+
```
kevina: Please add a test for the ls output without `-l` is unchanged when used on a file.

kevina:
> @kevina The field of the NodeListing structure displayed when -l isn't used is just Name,

Okay the reason I bring it up is that 'ls /afile' should not fail if the hash for /afile is not available.  But a quick test on the exiting version of IPFS verified this is not the case, so this is not a regression but something that could be cleaned up later.

so LGTM.

schomatis:
> 'ls /afile' should not fail if the hash for /afile is not available.

Good point, I should be replicating the logic of the `*mfs.Directory` case were that information is loaded only if `long` is true (`-l` option), let me fix that and get back to you.

schomatis:
Fixed, @kevina could you take another look please?

Just to clarify, in the file case, the node has already been fetched through MFS, so getting its hash (or any other information) won't fail, in contrast, in the directory case, after the directory node is fetched other nodes need to be looked in the DAG (files inside the directory) for the command to build its output, leaving open the door for an potential error.

However, even if this case can't fail, it should do the amount of work it needs to do and not more, so it needs to also (as the directory case) discriminate the `-l` true/false cases.

schomatis:
Oh, fun fact, I've just realized that the `Type` field that was hard-coded isn't even needed, I'm still leaving it in the PR just in case future version do show it (it's cheap to compute).

PR #5057, opened 2018-05-31
Announce relay addresses for peers with relays enabled

### LINKED ISSUE #4992: Add option to advertise relay addresses along with other addresses
Currently, if we want to advertise relay addresses, we have to specify all the addresses we want to advertise in the Addresses.Announce config field (because that field correctly replaces all addresses instead of augmenting them). We should have a way to either say "Add relay addresses to my list of addresses that I announce!" or "add this set of addresses to the list of addresses that I announce". 
I'm leaning towards the relay specific option, but opinions are welcome.

bigs:
Closes #4992 by removing code that filters circuit addresses from the host's bind list.

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -869,13 +869,6 @@ func constructPeerHost(ctx context.Context, id peer.ID, ps pstore.Peerstore, bwr
 	}
 
 	addrsFactory := opts.AddrsFactory
-	if !opts.DisableRelay {
-		if addrsFactory != nil {
-			addrsFactory = composeAddrsFactory(addrsFactory, filterRelayAddrs)
-		} else {
-			addrsFactory = filterRelayAddrs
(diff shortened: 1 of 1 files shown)


magik6k:
https://github.com/ipfs/go-ipfs/pull/4830 does the same thing. Also, see https://github.com/ipfs/go-ipfs/issues/4213 for explanation on why this isn't that simple

bigs:
@magik6k kk! your PR is more succinct anyhow, so i'm happy to close. i'll backlink your PR to the issue i came in through so this doesn't happen again. it's late here so i'm probably missing it, but what is the major blocker here? is it that we don't want to degrade performance by advertising an address that isn't easily connectable to, per [this issue](https://github.com/libp2p/specs/issues/33).

PR #5065, opened 2018-06-01
debian stretch fuse update

robotamer:


Stebalien:
I think this is the case with most linux distros at this point. I'd just reword this to: "On some older linux distributions, you may need to add yourself to the fuse group. If no such group exists, you can probably skip this step". That way, the documentation is as generic as possible.

Stebalien:
(also, you'll need to appease GitCop)

PR #5068, opened 2018-06-03
fix the duplicate (multiple versions) sys import

Stebalien:
License: MIT


Stebalien:
(also updates our fuse library)

PR #5072, opened 2018-06-03
Fix panic. Don't handle errors with fallthrough.

Bren2010:
License: MIT
Signed-off-by: Brendan McMillion <brendan@cloudflare.com>


whyrusleeping:
Thanks @Bren2010!

PR #5075, opened 2018-06-04
update multiplexers

Stebalien:
Updates both the yamux and smux multiplexers (and multistream, for that matter).

License: MIT


diasdavid:
🍾

PR #5082, opened 2018-06-05
Bring urlstore fork in sync with master

parkan:
This is my best attempt to bring master into the urlstore branch. Because gx imports are reordered by goimports based on the high bits of their hash, the import conflicts are quite cumbersome to resolve, so I'm not sure that this is the best way. Merged instead of rebased b/c even with rerere you have to re-resolve at every step.

@kyledrake do you have a better approach?

overall, merging some kind of stub version of urlstore into master seems desirable to avoid this process


kyledrake:
I'm not aware of a better approach for this, but @Stebalien may have one.

Stebalien:
FYI, @Kubuxu has rebase the main PR on master: https://github.com/ipfs/go-ipfs/pull/4896

As for merging, we, at the very minimum, need some decent tests.

parkan:
thanks @Kubuxu! closing this

PR #5096, opened 2018-06-07
Fix go-ipfs-chunker version desync due to bad merge

parkan:
this should unbreak the build


whyrusleeping:
@parkan added you to the go-ipfs team. You should be able to push to branches here now (just don't push to master)

PR #5098, opened 2018-06-08
[WIP] unixfs: integrate `pb.Data` into `FSNode` to avoid duplicating fields

schomatis:
```
[WIP] unixfs: integrate `pb.Data` into `FSNode` to avoid duplicating fields

Work in progress. Free sample. Do not merge.

Removed `Data` and `blocksizes` from `FSNode` and used directly the
`pb.Data` fields. Reflected those changes in `FSNodeFromBytes()` and
`GetBytes()` (`Unmarshal` and `Marshal` operations) to show how it
would simplify code.

This takes up more space, `pb.Data` is always created inside `FSNode`
and it stays there instead of just being created and destroyed during
the (un)marshal operations, but simplifies code because now the two
structures relationship is made more evident `FSNode` has no meaning
without the UnixFS format that backs it up (and ideally we should
only have one structure but protocol buffers automatically creates
the format structures and forces some decouple between the format
and the functionality).
```


schomatis:
My main question is exactly how to integrate the `pb.Data` structure into `FSNode`:

1. Whether like an anonymous structure providing direct access to its fields without the `Format` intermediary name. I prefer this one, although it would lead to ugly names like `n.Data.Data` (because the protobuf message is named Data but also one of its fields).

2. Like a named structure (current PR) and check what name would be more suited, `Format` was the first thing that came to mind.

@Stebalien

Stebalien:
I'd like to just embed it however, that would expose a lot of internal state. I believe we may want to eventually expose these datastructures as a library so that may not be the best idea.

---

Are you planning on reusing the `Type` field as well? As it stands, I'm not convinced this code is any simpler and it has a few drawbacks:

1. The `Format` field isn't really valid until you call `GetBytes` (potential foot gun).
2. `GetBytes` is no longer thread-safe (probably not an issue, but getters that modify always weird me out a bit).

whyrusleeping:
Seems like we could just `UpdateFilesize(0)` in the constructor.

schomatis:
Agreed.

I was doubting about doing that, I added it in `GetBytes()` because it assures me that I'm always going to have the `Filesize` initialized before marshaling, but since I'm making the constructor mandatory anyways (as it's the only way to set the `Type`, otherwise `GetBytes()` would err) I can move all the responsibility of the `FSNode` integrity there.

schomatis:
@Stebalien Could you please take another look?

Issue #5083, opened 2018-06-06
Transfer speed does not improve with multiple nodes.


Juelun:
I built a local network with four nodes which connected by router. Each node is installed with ubuntu 16.04.
![image](https://user-images.githubusercontent.com/29215260/41025590-e02ec7d6-69a4-11e8-82a8-e6bdf9fb58d7.png)
 
I used iftop tool to monitor network traffic on each node.
Running ipfs daemon on A node and  added a file as 644MB to IPFS  at A node firstly.  
Then  I tried to get the file on each node one by one(from B node to D node ) with ipfs get command. 

When I ran ipfs daemon on B node and got the file on B node,  it costed 26 seconds to get it.  
The B node network traffic : TX:11.5MB       RX:655MB

After then, running ipfs daemon on C node and I downloaded the file on C node. It costed 20 seconds to get it.
The C node network traffic : TX:20.5MB       RX:1.28GB

Finally,  running ipfs daemon on D node and  I downloaded the file on D node. It costed 35 seconds to get it.
The D node network traffic : TX:35.5MB       RX:1.94GB

The experiment seems if there is a file existing in multiple nodes, and a empty node tries to get it, the node will download the file from multiple nodes with replicated data. The download speed will not be improved with multiple nodes and even be worse than two point transfer.

In my original opinion, I thought the ipfs node will try to get different parts of the same file from different nodes to improve the speed. But, my experiment showed ipfs node will try to get an entire file from every nodes it connected. Is IPFS just designed like that? Or is there any way to make the download speed scale up with multiple nodes?

Juelun:
Pinning file on the other nodes is just to make the contents will not be deleted as garbage collection. But it can not resolve my concern. I have tried it. It did not improve the download speed as well.
My point is that if there is a file  stored in 3 nodes(A, B, C), and I tried to download it on D node.
Then download speed will faster than two nodes transfer even slower than two nodes transfer. 
I used iftop tool to monitor the network traffic at each node. I found the output network traffic of each node from A,B,C sent  was almost equal to an entire file size. And the input network traffic on D node got was almost equal to three times of the file size.
![untitled picture](https://user-images.githubusercontent.com/29215260/41089738-1526520a-6a32-11e8-8975-66bc84f78bfd.png)

ivan386:
https://github.com/ipfs/go-ipfs/blob/f43811a750cc65626d4c5169da2bc455010bbd8d/exchange/bitswap/message/pb/message.proto

[...]

PR #5104, opened 2018-06-11
Ivan386/bitswap 1.2.0

ivan386:
Protocol upgrade

1. Allow to request part of block. 
 Can be requested different parts of same block from different seeds. Increase speed.
2. Confirm block send to seed.
 Peer can chose what block from what peer he wants to get. Eliminates block duplication.

https://github.com/ipfs/go-ipfs/issues/5083

License: MIT
Signed-off-by: Ivan ivan386@users.noreply.github.com


Stebalien:
This kind of change needs a deep, shared understanding of the problem and a concrete, discussed, roadmap/solution.

PR #5117, opened 2018-06-14
Add initial support for identity hashes (Part 1)

kevina:
This provides support for identity hashes by using the new blockstore in https://github.com/ipfs/go-ipfs-blockstore/pull/4.

This is test and works locally, but to go in https://github.com/ipfs/go-ipfs-blockstore/pull/4 needs to be merged and gx-updated, and then go-ipfs-exchange-offline needs to be updated and gx-published.

### DIFF
--- a/core/builder.go
+++ b/core/builder.go
@@ -196,14 +196,16 @@ func setupNode(ctx context.Context, n *IpfsNode, cfg *BuildCfg) error {
 		opts.HasBloomFilterSize = 0
 	}
 
-	cbs, err := bstore.CachedBlockstore(ctx, bs, opts)
+	wbs, err := bstore.CachedBlockstore(ctx, bs, opts)
 	if err != nil {
 		return err
 	}
 
-	n.BaseBlocks = cbs
+	wbs = bstore.NewIdStore(wbs)
+
+	n.BaseBlocks = wbs
 	n.GCLocker = bstore.NewGCLocker()
-	n.Blockstore = bstore.NewGCBlockstore(cbs, n.GCLocker)
+	n.Blockstore = bstore.NewGCBlockstore(wbs, n.GCLocker)
 
 	if conf.Experimental.FilestoreEnabled {
 		// hash security
--- a/test/sharness/t0046-id-hash.sh
+++ b/test/sharness/t0046-id-hash.sh
@@ -1,0 +1,46 @@
+#!/usr/bin/env bash
+
+test_description="Test basic operations with identity hash"
+
+. lib/test-lib.sh
+
+test_init_ipfs
+
+ID_HASH0=z25RnHTQ7DveGAsV6YDSDR8EkWvD
+ID_HASH0_CONTENTS=jkD98jkD975hkD8
+
+test_expect_success "can fetch random id hash" '
+  ipfs cat $ID_HASH0 > expected &&
+  echo $ID_HASH0_CONTENTS > actual &&
+  test_cmp expected actual
+'
+
+test_expect_success "can pin random id hash" '
+  ipfs pin add $ID_HASH0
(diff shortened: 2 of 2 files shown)


kevina:
After this goes in the next step will be to automatically use identity hashes. But that change will require a likely require a go-cid update.

ivan386:
>  if a node that doesn't have id-hash support requests an id-hash from one that does, it will be returned.
This will be flood from all connected peers?

kevina:
> This will be flood from all connected peers?

I don't know enough bitswap to answer that.  I just tested it using two peers, one with the P.R. and one without.

Stebalien:
> This will be flood from all connected peers?

Yes, but that's an issue with bitswap (and already happens for common blocks).

ivan386:
> Yes, but that's an issue with bitswap (and already happens for common blocks).

Common blocks send only from peers that have it. Identity blocks will be send from all peers that knows identity hash.

Stebalien:
Unfortunately, there's really not much we can do here other than wait a few releases between *this* patch and one that allows users to use inline blocks. That is, users of existing releases of IPFS will always either:

1. Get spammed.
2. Never get the block.

I'd rather have 1.

core/builder.go
```diff
@@ -196,18 +196,20 @@ func setupNode(ctx context.Context, n *IpfsNode, cfg *BuildCfg) error {
 		opts.HasBloomFilterSize = 0
 	}
 
-	cbs, err := bstore.CachedBlockstore(ctx, bs, opts)
+	wbs, err := bstore.CachedBlockstore(ctx, bs, opts)
```
schomatis: Small nit: why is the `CachedBlockstore` renamed from `cbs` to `wbs`? Shouldn't `wbs` (which I'm not sure what it stands for) be defined later in `NewIdStore`?

kevina: `wbs` stands for wrapped-data-store.

I could keep `cbs` and then defined yet another variable.  But that could lead to bugs where `cbs` was used when `wbs` was meant to be used.

kevina:
I will rebase this on Thursday when I get back.

core/builder.go
schomatis: Ok, I was just saying because the following self-definition seem a bit weird to me,

```
wbs = bstore.NewIdStore(wbs)
```

and I would have expected to create it from another blockstore, not from itself, like

```
wbs = bstore.NewIdStore(cbs)
```

schomatis: > wbs stands for wrapped-data-store.

Shouldn't it be named `wds` instead? To avoid confusing it with a blockstore (like `cbs`).

kevina:
Sorry I meant wrapped-block-store.

On July 18, 2018 2:43:46 PM CEST, Lucas Molas <notifications@github.com> wrote:
>schomatis commented on this pull request.
>
>
>
>> @@ -196,18 +196,20 @@ func setupNode(ctx context.Context, n
>*IpfsNode, cfg *BuildCfg) error {
> 		opts.HasBloomFilterSize = 0
> 	}
> 
>-	cbs, err := bstore.CachedBlockstore(ctx, bs, opts)
>+	wbs, err := bstore.CachedBlockstore(ctx, bs, opts)
>
>> wbs stands for wrapped-data-store.
>
>Shouldn't it be named `wds` instead? To avoid confusing it with a
>blockstore (like `cbs`).
>
>-- 
>You are receiving this because you were assigned.
>Reply to this email directly or view it on GitHub:
>https://github.com/ipfs/go-ipfs/pull/5117#discussion_r203362952

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

kevina:
Rebased.

Issue #5056, opened 2018-05-31
Avoid renaming imports


schomatis:
As suggested [here](https://github.com/golang/go/wiki/CodeReviewComments#imports).

The most problematic thing is that the aliases used are not homogeneous. I'll be eliminating some of those aliases during the work of https://github.com/ipfs/go-ipfs/issues/5050, although I don't think it's relevant to eliminate *all* of them now from the code it's something that should be taken in consideration for future contributors (maybe this could be added to the [Go contribution guidelines](https://github.com/ipfs/community/blob/master/go-code-guidelines.md#imports)).

I understand that it is useful to rename packages in order to remove redundant prefixes like `go-`, but if that was the original purpose we should document it as a special case, was there any other reason to rename imports?

kevina:
I do not necessary agree with this.  What do others think?

Renaming a long package name that is used frequently import can lead to shorter and easier to read code.

I also don't see why this is a big deal.  Standard tools work well with renamed imports and it is easy to determine an import by just looking at the import list.

schomatis:
> I do not necessary agree with this. What do others think?

@kevina  Thanks for your feedback, yes I'd love to get more opinions to see if I'm jumping the gun here.

My main motivation is to avoid using different names in different files for the same package which makes the code harder to read (see https://github.com/ipfs/go-ipfs/issues/5055) but I would like to know if there are any other motivations to renaming packages than just making them shorter. I'm not sure if shorter names means easier to read code, at least not any name and not any variation of it, maybe we could start standardizing some of those shorter names.

kevina:
@schomatis the length of the import really depends on if the import name is an important part of the name, if it is not I see it as noise and hence prefer shorter names.

I am not against some cleaning up of the code so we use the same alias for an important, especially within the same (or similar) packages.

I also agree the protobuf code is particularly bad and I won't object to cleaning it up.

whyrusleeping:
I disagree with the go convention here. Since go package names arent required to match their  paths, Its far clearer to explicitly name them.

Which of these is more clear?

```go
package main

import (
    "github.com/foo/bar/baz"
)

func main() {
    cats.Meow()
} 
```

or 

```go
package main

import (
    cats "github.com/foo/bar/baz"
)

func main() {
    cats.Meow()
} 
```

this is a pretty severe example, but i have encountered some pretty annoying mismatches here. Even just the difference between `go-foo` in the path and `foo.X` in the code adds non-zero cognitive load.

[...]

PR #5118, opened 2018-06-14
importer: remove `UnixfsNode` from the balanced builder

### LINKED ISSUE #5106: Revisiting the balanced builder
In issue https://github.com/ipfs/go-ipfs/issues/5059 there is a brief introduction of how is a file broken down in nodes and arranged in a DAG, as mentioned in that issue there are many aspects of the procedure that are not clear to someone without experience in the code and many key operations are flat out hidden. I'm taking now a closer look at the [balanced importer/builder](https://github.com/ipfs/go-ipfs/blob/7853e53860805e08a212d78c4baa5d59bff99ba8/importer/balanced/builder.go#L29) in charge of that, and more specifically at the [helper](https://github.com/ipfs/go-ipfs/blob/7853e53860805e08a212d78c4baa5d59bff99ba8/importer/helpers/helpers.go#L40-L48) package that actually hides most of the implementation, to see how this can be more clearly exposed in the code. Most of what is discussed here can also be applied to the trickle importer, but I'm starting with the balanced one as it has a simpler layout.

schomatis:
The `builder.go` file has been completely rewritten, should be reviewed as an entirely [new file](https://github.com/schomatis/go-ipfs/blob/fe2ce15e8619e4f3f709196453c18a1dc84a20c5/importer/balanced/builder.go).

The `dagbuilder.go` has mostly additions and can be reviewed as a [diff](https://github.com/ipfs/go-ipfs/pull/5118/files#diff-fdae772d2b35bc89d019b608d02c1755).

Closes #5106.

------------------------------

```
importer: remove `UnixfsNode` from the balanced builder

The `UnixfsNode` structure has multiple pointers to many (non-complementary)
mutually exclusive node types, only some of them are active (not-`nil`) at a
given time in the code path which made the code too convoluted. Specifically,
the most important distinction between node types was being hidden: leaf nodes
vs internal (non-leaf) nodes.

Remove entirely the use of `UnixfsNode` from the `balanced` package replacing it
in turn with the newly created `FSNodeOverDag` structure that represents the
UnixFS node encoded inside the DAG node, primarily used for internal node
representations. Leaf nodes are handled exclusively in the `NewLeafDataNode`
encapsulating its multiple representations (that we're previously exposed in
`UnixfsNode` as conflicting pointers).

The `builder.go` file has been completely rewritten, although the basic DAG
creation algorithm has been preserved (extending a full DAG by creating a new
root and linking the old one as its child), the most significant modification
has been in the loop of `Layout` that now only handles internal nodes (i.e.,
nodes with `depth` bigger than zero) to be able to adapt `fillNodeRec` to only
that scenario (avoiding the replace logic of the zero `depth` case with the
defective `Set` function, now removed). The `fillNodeRec` now explicitly returns
the `ipld.Node` and the size of the file data it's storing to propagate it
upwards into the DAG.

The `DagBuilderHelper` was heavily extended to incorporate `ipld.Node` functions
that would replace the `UnixfsNode` ones used by the balanced builder:
`NewLeafNode()`, `NewLeafDataNode()` and `AddNodeAndClose()`. Also, the
`ProcessFileStore` function was incorporated to encapsulate all the logic
related to the Filestore support which was scattered throughout the builder
logic, the `offset` that was being passed through most functions is now a part
of the `DagBuilderHelper`.

This has turned out to be a rather big commit, it should have been split into
more smaller and logically cohesive commits, but the `UnixfsNode` was too
entangled inside the logic and that would have required a progressive
modification of the `UnixfsNode` structure as well, which wasn't possible as it
is still being used by the balanced builder (the same reason why most of the
`UnixfsNode`-related functions cannot yet be removed, leaving the `helpers.go`
file mostly untouched).
```

### DIFF
--- a/importer/balanced/builder.go
+++ b/importer/balanced/builder.go
@@ -1,113 +1,255 @@
-// Package balanced provides methods to build balanced DAGs.
-// In a balanced DAG, nodes are added to a single root
-// until the maximum number of links is reached (with leaves
-// being at depth 0). Then, a new root is created, and points to the
-// old root, and incorporates a new child, which proceeds to be
-// filled up (link) to more leaves. In all cases, the Data (chunks)
-// is stored only at the leaves, with the rest of nodes only
-// storing links to their children.
+// Package balanced provides methods to build balanced DAGs, which are generalistic
+// DAGs in which all leaves (nodes representing chunks of data) are at the same
+// distance from the root. Nodes can have only a maximum number of children; to be
+// able to store more leaf data nodes balanced DAGs are extended by increasing its
+// depth (and having more intermediary nodes).
 //
-// In a balanced DAG, nodes fill their link capacity before
-// creating new ones, thus depth only increases when the
-// current tree is completely full.
+// Internal nodes are always represented by UnixFS nodes (of type `File`) encoded
+// inside DAG nodes (see the `go-ipfs/unixfs` package for details of UnixFS). In
+// contrast, leaf nodes with data have multiple possible representations: UnixFS
+// nodes as above, raw nodes with just the file data (no format) and Filestore
+// nodes (that directly link to the file on disk using a format stored on a raw
+// node, see the `go-ipfs/filestore` package for details of Filestore.)
 //
-// Balanced DAGs are generalistic DAGs in which all leaves
-// are at the same distance from the root.
+// In the case the entire file fits into just one node it will be formatted as a
+// (single) leaf node (without parent) with the possible representations already
+// mentioned. This is the only scenario where the root can be of a type different
+// that the UnixFS node.
+//
+//                                                 +-------------+
+//                                                 |   Root 4    |
+//                                                 +-------------+
+//                                                       |
+//                            +--------------------------+----------------------------+
+//                            |                                                       |
+//                      +-------------+                                         +-------------+
+//                      |   Node 2    |                                         |   Node 5    |
+//                      +-------------+                                         +-------------+
+//                            |                                                       |
+//              +-------------+-------------+                           +-------------+
+//              |                           |                           |
+//       +-------------+             +-------------+             +-------------+
+//       |   Node 1    |             |   Node 3    |             |   Node 6    |
+//       +-------------+             +-------------+             +-------------+
+//              |                           |                           |
+//       +------+------+             +------+------+             +------+
+//       |             |             |             |             |
+//  +=========+   +=========+   +=========+   +=========+   +=========+
+//  | Chunk 1 |   | Chunk 2 |   | Chunk 3 |   | Chunk 4 |   | Chunk 5 |
+//  +=========+   +=========+   +=========+   +=========+   +=========+
+//
 package balanced
 
 import (
 	"errors"
 
 	h "github.com/ipfs/go-ipfs/importer/helpers"
+	ft "github.com/ipfs/go-ipfs/unixfs"
 
 	ipld "gx/ipfs/QmWi2BYBL5gJ3CiAiQchg6rn1A8iBsrWy51EYxvHVjFvLb/go-ipld-format"
 )
 
-// Layout builds a balanced DAG. Data is stored at the leaves
-// and depth only increases when the tree is full, that is, when
-// the root node has reached the maximum number of links.
+// Layout builds a balanced DAG layout. In a balanced DAG of depth 1, leaf nodes
+// with data are added to a single `root` until the maximum number of links is
+// reached. Then, to continue adding more data leaf nodes, a `newRoot` is created
+// pointing to the old `root` (which will now become and intermediary node),
+// increasing the depth of the DAG to 2. This will increase the maximum number of
+// data leaf nodes the DAG can have (`Maxlinks() ^ depth`). The `fillNodeRec`
+// function will add more intermediary child nodes to `newRoot` (which already has
+// `root` as child) that in turn will have leaf nodes with data added to them.
+// After that process is completed (the maximum number of links is reached),
+// `fillNodeRec` will return and the loop will be repeated: the `newRoot` created
+// will become the old `root` and a new root will be created again to increase the
+// depth of the DAG. The process is repeated until there is no more data to add
+// (i.e. the DagBuilderHelper’s Done() function returns true).
+//
+// The nodes are filled recursively, so the DAG is built from the bottom up. Leaf
+// nodes are created first using the chunked file data and its size. The size is
+// then bubbled up to the parent (internal) node, which aggregates all the sizes of
+// its children and bubbles that combined size up to its parent, and so on up to
+// the root. This way, a balanced DAG acts like a B-tree when seeking to a byte
+// offset in the file the graph represents: each internal node uses the file size
+// of its children as an index when seeking.
+//
+//      `Layout` creates a root and hands it off to be filled:
+//
+//             +-------------+
+//             |   Root 1    |
+//             +-------------+
+//                    |
+//       ( fillNodeRec fills in the )
+//       ( chunks on the root.      )
+//                    |
+//             +------+------+
+//             |             |
+//        + - - - - +   + - - - - +
+//        | Chunk 1 |   | Chunk 2 |
+//        + - - - - +   + - - - - +
+//
+//                           ↓
+//      When the root is full but there's more data...
+//                           ↓
+//
+//             +-------------+
+//             |   Root 1    |
+//             +-------------+
+//                    |
+//             +------+------+
+//             |             |
+//        +=========+   +=========+   + - - - - +
+//        | Chunk 1 |   | Chunk 2 |   | Chunk 3 |
+//        +=========+   +=========+   + - - - - +
+//
+//                           ↓
+//      ...Layout's job is to create a new root.
+//                           ↓
+//
+//                            +-------------+
+//                            |   Root 2    |
+//                            +-------------+
+//                                  |
+//                    +-------------+ - - - - - - - - +
+//                    |                               |
+//             +-------------+            ( fillNodeRec creates the )
+//             |   Node 1    |            ( branch that connects    )
+//             +-------------+            ( "Root 2" to "Chunk 3."  )
+//                    |                               |
+//             +------+------+             + - - - - -+
+//             |             |             |
+//        +=========+   +=========+   + - - - - +
+//        | Chunk 1 |   | Chunk 2 |   | Chunk 3 |
+//        +=========+   +=========+   + - - - - +
+//
 func Layout(db *h.DagBuilderHelper) (ipld.Node, error) {
-	var offset uint64
-	var root *h.UnixfsNode
-	for level := 0; !db.Done(); level++ {
+	if db.Done() {
+		// No data, return just an empty node.
+		root, err := db.NewLeafNode(nil)
+		if err != nil {
+			return nil, err
+		}
+		// This works without Filestore support (`ProcessFileStore`).
+		// TODO: Why? Is there a test case missing?
 
-		nroot := db.NewUnixfsNode()
-		db.SetPosInfo(nroot, 0)
+		return db.AddNodeAndClose(root)
+	}
 
-		// add our old root as a child of the new root.
-		if root != nil { // nil if it's the first node.
-			if err := nroot.AddChild(root, db); err != nil {
-				return nil, err
+	// The first `root` will be a single leaf node with data
+	// (corner case), after that subsequent `root` nodes will
+	// always be internal nodes (with a depth > 0) that can
+	// be handled by the loop.
+	root, fileSize, err := db.NewLeafDataNode()
+	if err != nil {
+		return nil, err
+	}
+
+	// Each time a DAG of a certain `depth` is filled (because it
+	// has reached its maximum capacity of `db.Maxlinks()` per node)
+	// extend it by making it a sub-DAG of a bigger DAG with `depth+1`.
+	for depth := 1; !db.Done(); depth++ {
+
+		// Add the old `root` as a child of the `newRoot`.
+		newRoot := db.NewFSNodeOverDag(ft.TFile)
+		newRoot.AddChild(root, fileSize, db)
+
+		// Fill the `newRoot` (that has the old `root` already as child)
+		// and make it the current `root` for the next iteration (when
+		// it will become "old").
+		root, fileSize, err = fillNodeRec(db, newRoot, depth)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return db.AddNodeAndClose(root)
+}
+
+// fillNodeRec will "fill" the given internal (non-leaf) `node` with data by
+// adding child nodes to it, either leaf data nodes (if `depth` is 1) or more
+// internal nodes with higher depth (and calling itself recursively on them
+// until *they* are filled with data). The data to fill the node with is
+// provided by DagBuilderHelper.
+//
+// `node` represents a (sub-)DAG root that is being filled. If called recursively,
+// it is `nil`, a new node is created. If it has been called from `Layout` (see
+// diagram below) it points to the new root (that increases the depth of the DAG),
+// it already has a child (the old root). New children will be added to this new
+// root, and those children will in turn be filled (calling `fillNodeRec`
+// recursively).
+//
+//                      +-------------+
+//                      |   `node`    |
+//                      |  (new root) |
+//                      +-------------+
+//                            |
+//              +-------------+ - - - - - - + - - - - - - - - - - - +
+//              |                           |                       |
+//      +--------------+             + - - - - -  +           + - - - - -  +
+//      |  (old root)  |             |  new child |           |            |
+//      +--------------+             + - - - - -  +           + - - - - -  +
+//              |                          |                        |
+//       +------+------+             + - - + - - - +
+//       |             |             |             |
+//  +=========+   +=========+   + - - - - +    + - - - - +
+//  | Chunk 1 |   | Chunk 2 |   | Chunk 3 |    | Chunk 4 |
+//  +=========+   +=========+   + - - - - +    + - - - - +
+//
+// The `node` to be filled uses the `FSNodeOverDag` abstraction that allows adding
+// child nodes without packing/unpacking the UnixFS layer node (having an internal
+// `ft.FSNode` cache).
+//
+// It returns the `ipld.Node` representation of the passed `node` filled with
+// children and the `nodeFileSize` with the total size of the file chunk (leaf)
+// nodes stored under this node (parent nodes store this to enable efficient
+// seeking through the DAG when reading data later).
+//
+// warning: **children** pinned indirectly, but input node IS NOT pinned.
+func fillNodeRec(db *h.DagBuilderHelper, node *h.FSNodeOverDag, depth int) (filledNode ipld.Node, nodeFileSize uint64, err error) {
+	if depth < 1 {
+		return nil, 0, errors.New("attempt to fillNode at depth < 1")
+	}
+
+	if node == nil {
+		node = db.NewFSNodeOverDag(ft.TFile)
+	}
+
+	// Child node created on every iteration to add to parent `node`.
+	// It can be a leaf node or another internal node.
+	var childNode ipld.Node
+	// File size from the child node needed to update the `FSNode`
+	// in `node` when adding the child.
+	var childFileSize uint64
+
+	// While we have room and there is data available to be added.
+	for node.NumChildren() < db.Maxlinks() && !db.Done() {
+
+		if depth == 1 {
+			// Base case: add leaf node with data.
+			childNode, childFileSize, err = db.NewLeafDataNode()
+			if err != nil {
+				return nil, 0, err
+			}
+		} else {
+			// Recursion case: create an internal node to in turn keep
+			// descending in the DAG and adding child nodes to it.
+			childNode, childFileSize, err = fillNodeRec(db, nil, depth-1)
+			if err != nil {
+				return nil, 0, err
 			}
 		}
 
-		// fill it up.
-		if err := fillNodeRec(db, nroot, level, offset); err != nil {
-			return nil, err
-		}
-
-		offset = nroot.FileSize()
-		root = nroot
-
-	}
-	if root == nil {
-		// this should only happen with an empty node, so return a leaf
-		var err error
-		root, err = db.NewLeaf(nil)
+		err = node.AddChild(childNode, childFileSize, db)
 		if err != nil {
-			return nil, err
+			return nil, 0, err
 		}
 	}
 
-	out, err := db.Add(root)
+	nodeFileSize = node.FileSize()
+
+	// Get the final `dag.ProtoNode` with the `FSNode` data encoded inside.
+	filledNode, err = node.Commit()
 	if err != nil {
-		return nil, err
+		return nil, 0, err
 	}
 
-	err = db.Close()
-	if err != nil {
-		return nil, err
-	}
-
-	return out, nil
-}
-
-// fillNodeRec will fill the given node with data from the dagBuilders input
-// source down to an indirection depth as specified by 'depth'
-// it returns the total dataSize of the node, and a potential error
-//
-// warning: **children** pinned indirectly, but input node IS NOT pinned.
-func fillNodeRec(db *h.DagBuilderHelper, node *h.UnixfsNode, depth int, offset uint64) error {
-	if depth < 0 {
-		return errors.New("attempt to fillNode at depth < 0")
-	}
-
-	// Base case
-	if depth <= 0 { // catch accidental -1's in case error above is removed.
-		child, err := db.GetNextDataNode()
-		if err != nil {
-			return err
-		}
-
-		node.Set(child)
-		return nil
-	}
-
-	// while we have room AND we're not done
-	for node.NumChildren() < db.Maxlinks() && !db.Done() {
-		child := db.NewUnixfsNode()
-		db.SetPosInfo(child, offset)
-
-		err := fillNodeRec(db, child, depth-1, offset)
-		if err != nil {
-			return err
-		}
-
-		if err := node.AddChild(child, db); err != nil {
-			return err
-		}
-		offset += child.FileSize()
-	}
-
-	return nil
+	return filledNode, nodeFileSize, nil
 }
--- a/importer/helpers/dagbuilder.go
+++ b/importer/helpers/dagbuilder.go
@@ -7,7 +7,9 @@ import (
 
 	dag "github.com/ipfs/go-ipfs/merkledag"
 	ft "github.com/ipfs/go-ipfs/unixfs"
+	pb "github.com/ipfs/go-ipfs/unixfs/pb"
 
+	pi "gx/ipfs/QmUWsXLvYYDAaoAt9TPZpFX4ffHHMg46AHrz1ZLTN5ABbe/go-ipfs-posinfo"
 	ipld "gx/ipfs/QmWi2BYBL5gJ3CiAiQchg6rn1A8iBsrWy51EYxvHVjFvLb/go-ipld-format"
 	chunker "gx/ipfs/QmXnzH7wowyLZy8XJxxaQCVTgLMcDXdMBznmsrmQWCyiQV/go-ipfs-chunker"
 	cid "gx/ipfs/QmapdYm1b22Frv3k17fqrBYTFRxwiaVJkB299Mfn33edeB/go-cid"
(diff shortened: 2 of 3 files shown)


whyrusleeping:
> Do not review.

ACK

schomatis:
@whyrusleeping I mean, take a look if you want, but right now I've basically expanded the `UnixfsNode` functions in the balanced builder and removed what was not necessary for an internal node, and similarly for the leaf node (internal vs leaf was the basic distinction that `UnixfsNode` was hiding and hence the code was losing an important dimensionality). The result is a lot of boilerplate code, but now, if this (hopefully) passes the tests it needs to be re-encapsulated in more IPLD-oriented functions.

schomatis:
@Mr0grog Could you take a look at the comments please to check if they make sense? Especially the [commit message](https://github.com/ipfs/go-ipfs/pull/5118/commits/fe2ce15e8619e4f3f709196453c18a1dc84a20c5), the [package doc](https://github.com/schomatis/go-ipfs/blob/fe2ce15e8619e4f3f709196453c18a1dc84a20c5/importer/balanced/builder.go#L1-L31), the [`FSNodeOverDag` doc](https://github.com/ipfs/go-ipfs/pull/5118/files#diff-fdae772d2b35bc89d019b608d02c1755R366), and the [`fillNodeRec` doc](https://github.com/schomatis/go-ipfs/blob/fe2ce15e8619e4f3f709196453c18a1dc84a20c5/importer/balanced/builder.go#L88-L106).

Mr0grog:
@schomatis I’ll try and take a look later today or tomorrow. I think it’s safe to say you know a lot more about all these parts than I do now, though!

schomatis:
@Mr0grog Thanks, there's no rush, you don't need to check that what it says it's true but if actually some coherent meaning can be grasped from those lines.

importer/balanced/builder.go
```diff
@@ -27,87 +43,114 @@ import (
...
-// it returns the total dataSize of the node, and a potential error
+// fillNodeRec will fill the given internal (non-leaf) `node` with either
+// more internal nodes or leaf data nodes (when the final `depth` of 1 is
+// reached) containing the file data passed from the `DagBuilderHelper`.
+//
+// If `node` is `nil` it means we're descending into new territory so
+// just create a new node to be filled (recursion call); if it's not
+// `nil` it has been called from `Layout` where the DAG has been extended
+// to a new root, that already has a child (the old root), to be filled
+// with more children (up to `db.Maxlinks()`).
```
Mr0grog: A diagram might help again here:

```
// `node` represents a new root that is being expanded like "Root 2" in the
// diagram below. If it is `nil`, then this function should create a new node
// like "Root 3" in the diagram below. (It should only be `nil` when calling
// itself recursively, having started with a node like "Root 2".)
// 
//                         +-------------+
//                         |   Root 2    |
//                         +-------------+
//                               |       
//                 +-------------+ - - - - - - +
//                 |                           |
//          +-------------+             + - - - - - - +
//          |   Root 1    |             |   Root 3    |
//          +-------------+             + - - - - - - +
//                 |                           |       
//          +------+------+             + - - -+ - - -+
//          |             |             |             |
//     +=========+   +=========+   + - - - - +   + - - - - +
//     | Chunk 1 |   | Chunk 2 |   | Chunk 3 |   | Chunk 4 |
//     +=========+   +=========+   + - - - - +   + - - - - +
//
// The `Layout` function is responsible for creating new roots like "Root 2"
// when more data doesn't fit in the current root. It call this function, which
// is responsible for creating the new branch like "Root 3" by calling itself
// recursively.
```

importer/balanced/builder.go
```diff
@@ -27,87 +43,114 @@ import (
...
+//
+// The `node` to be filled uses the `FSNodeOverDag` abstraction that
+// allows to add child nodes without packing/unpacking the UnixFS layer
```
Mr0grog: “allows to add” → “allows adding”

importer/balanced/builder.go
```diff
@@ -27,87 +43,114 @@ import (
...
+// node (having an internal `ft.FSNode` cache).
+//
+// It returns the `ipld.Node` representation of the passed `node` filled
+// with children and the `nodeFileSize` with the size of the file stored
+// under this node (to propagate it upwards in the DAG to its parent).
```
Mr0grog: Instead of:

> `nodeFileSize` with the size of the file stored under this node

maybe:

> `nodeFileSize` with the total size of all nodes under this node

or 

> `nodeFileSize` with the total size of the file chunk (leaf) nodes stored under this node

“size of the file” makes it unclear whether it’s the *whole* file or just the relevant chunks.

importer/balanced/builder.go
```diff
@@ -27,87 +43,114 @@ import (
 // and depth only increases when the tree is full, that is, when
 // the root node has reached the maximum number of links.
```
Mr0grog: In keeping with the diagrams I suggested in other sections, this could help here:

```go
//            +-------------+        
//            |   Root 1    |        
//            +-------------+        
//                   |               
//      ( fillNodeRec fills in the )
//      ( chunks on the root.      )
//                   |
//            +------+------+        
//            |             |        
//       + - - - - +   + - - - - +
//       | Chunk 1 |   | Chunk 2 |
//       + - - - - +   + - - - - +
//
//                          ↓
//     When the root is full but there's more data...
//                          ↓
//
//            +-------------+        
//            |   Root 1    |        
//            +-------------+        
//                   |               
//            +------+------+        
//            |             |        
//       +=========+   +=========+   + - - - - +
//       | Chunk 1 |   | Chunk 2 |   | Chunk 3 |
//       +=========+   +=========+   + - - - - +
//
//                          ↓
//     ...Layou's job is to create a new root.
//                          ↓
//
//                           +-------------+
//                           |   Root 2    |
//                           +-------------+
//                                 |
//                   +-------------+ - - - - - - - - +
//                   |                               |
//            +-------------+            ( fillNodeRec creates the )
//            |   Root 1    |            ( branch that connects    )
//            +-------------+            ( "Root 2" to "Chunk 3."  )
//                   |                               |
//            +------+------+             + - - - - -+
//            |             |             |
//       +=========+   +=========+   + - - - - +
//       | Chunk 1 |   | Chunk 2 |   | Chunk 3 |
//       +=========+   +=========+   + - - - - +
```

importer/helpers/dagbuilder.go
```diff
@@ -237,3 +345,104 @@ func (db *DagBuilderHelper) Maxlinks() int {
...
+
+	err = db.Close()
+	if err != nil {
+		return nil, err
+	}
+
+	return node, nil
+}
+
+// FSNodeOverDag encapsulates an `ft.FSNode` that will be stored in a
```
Mr0grog: This might just be my general newness to Go, but I find the `ft` here confusing (I know I can look up the fact that it’s referring to the `unixfs` package at the top, but I keep having to do that when I’m new to files AND none of that shows up in the Godoc). What if we said:

> FSNodeOverDag encapsulates a UnixFS `fs.FSNode`

? Is that any better, or is it just confusing in a different way?

importer/helpers/dagbuilder.go
```diff
@@ -237,3 +345,104 @@ func (db *DagBuilderHelper) Maxlinks() int {
...
+// `dag.ProtoNode`. Instead of just having a single `ipld.Node` that
+// would need to be constantly (un)packed to access and modify its
+// internal `FSNode` in the process of creating a UnixFS DAG, this
+// structure stores an `FSNode` cache to manipulate it (add child nodes)
+// directly , and only when the node has reached its final (immutable)
+// state it is committed to a single (indivisible) `ipld.Node`.
```
Mr0grog: “state it is committed” → “state is it committed”

importer/helpers/dagbuilder.go
```diff
@@ -237,3 +345,104 @@ func (db *DagBuilderHelper) Maxlinks() int {
...
+//
+// It is used mainly for internal (non-leaf) nodes, and for some
+// representations of data leaf nodes (that don't use raw nodes or
+// Filestore).
+//
+// It aims at replacing the `UnixfsNode` structure which encapsulated too
```
Mr0grog: “at replacing” → “to replace”

importer/helpers/dagbuilder.go
Mr0grog: I might add that moving to its final state is done by calling `Commit`:

> …and only when the node has reached its final state (signaled by calling `Commit()`) is it committed…

Otherwise it sounds like that happens automatically as a result of doing some other routine action.

importer/helpers/dagbuilder.go
```diff
@@ -237,3 +345,104 @@ func (db *DagBuilderHelper) Maxlinks() int {
...
+}
+
+// FileSize returns the `Filesize` attribute from the underlying
+// representation of the `ft.FSNode`.
+func (n *FSNodeOverDag) FileSize() uint64 {
+	return n.file.FileSize()
+}
+
+// SetFileData stores the `fileData` in the `ft.FSNode`. It
+// is used only when `FSNodeOverDag` represents a leaf node
```
Mr0grog: “is used” → “should be used” ?

This code doesn’t control how it’s called, so this should be a recommendation to the reader about how to use it, not an assertion about what some other part of the codebase might or might not do.

importer/helpers/dagbuilder.go
```diff
@@ -237,3 +345,104 @@ func (db *DagBuilderHelper) Maxlinks() int {
...
+	n.file.AddBlockSize(fileSize)
+
+	return db.batch.Add(child)
+}
+
+// Commit unifies (resolves) the cache nodes into a single `ipld.Node`
+// that represents them: the `ft.FSNode` is encoded inside the
+// `dag.ProtoNode`.
+//
+// TODO: Evaluate making it read-only after committing.
```
Mr0grog: I like this idea (or having a dirty flag) :)

schomatis:
@Mr0grog :heart_eyes:

schomatis:
> With code covering a conceptually complicated area like this, I think diagrams can help a lot. They take up a lot of space (and a little time to create), but they are more than worth the effort for others trying to understand the code. (I used a similar technique extensively in a couple geometry libraries I’ve worked on in the past and had lots of folks say it was very helpful.)

I agree 100% on this, that's why I asked about tools to programmatically construct graphs like the ones you did (https://github.com/ipfs/docs/issues/78), `graphmd` didn't seem to be suited for this, I'll take a closer look at Graphviz to see what kind of text graphs it can generate.

importer/balanced/builder.go
```diff
@@ -1,24 +1,40 @@
...
-// is stored only at the leaves, with the rest of nodes only
-// storing links to their children.
+// Package balanced provides methods to build balanced DAGs. In a
+// balanced DAG, nodes are added to a single root until the maximum
+// number of links is reached (with leaves being at depth 0). Then, a new
+// root is created to point to the old root (incorporating it as its
+// first child), which proceeds to be filled up (link) to more nodes. In
+// all cases, the data (chunk) is stored only at the leaves, with the
+// rest of (internal) nodes only storing links to their children and
+// sizes of the data stored under them.
```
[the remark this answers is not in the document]
schomatis: > Do we mirror the structure so the whole graph has the same depth

Great question! You *mirror* what you already had, and that term definitely needs to go in the description (yes, one nice graph can clarify all of that), the *balanced* adjective already implies it (the 1st graph is balanced and the 2nd isn't) but may not be clear enough. The problem with the old explanation:

> Then, a new root is created, and points to the old root, and incorporates a new child, which proceeds to be filled up (link) to more leaves.

Is that it is correct for the first iteration, but saying that a root's child (or any child for that matter) with be filled with leaves sound like:

```
                                  +-------------+
                                  |   Root 4    |
                                  +-------------+
                                        |       
                          +-------------+ - - - -- - -+
                          |                           |
                    +-------------+            +=============+
                    |   Root 3    |            |   Chunk 4   |
                    +-------------+            +=============+
                          |       
            +-------------+-------------+
            |                           |
     +-------------+             +=============+
     |   Root 1    |             |   Chunk 3   |
     +-------------+             +=============+
            |            
     +------+------+     
     |             |     
+=========+   +=========+
| Chunk 1 |   | Chunk 2 |
+=========+   +=========+
```
which actually is the graph you did for the new explanation that talks about nodes (generic, not internal or leaves, because that depends on the depth), I'm interested in why you reached that conclusion. But the new explanation may also be too plain, let me iterate a bit on your version and ping you back. I think the *mirror* term is the key (and the graph, no question), that's why I talk about a B-tree in the package doc, but that may not be the most appropriate data structure.

importer/balanced/builder.go
```diff
@@ -1,24 +1,40 @@
...
-// creating new ones, thus depth only increases when the
-// current tree is completely full.
+// Internal nodes are always represented by UnixFS nodes (of type `File`)
+// encoded inside DAG nodes, in contrast, leaf nodes with data have
+// multiple possible representations: UnixFS nodes as above, raw nodes
+// with just the file data (no format) and Filestore nodes (supported by
+// raw nodes). In the case the entire file fits into just one node it
+// will be formatted as a (single) leaf node (without parent) with the
+// possible representations already mentioned, this is the only scenario
+// where the root can be of a type different that the UnixFS node.
```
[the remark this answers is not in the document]
schomatis: > I’m not sure what “Filestore nodes (supported by raw nodes)” means.

I left a vague description on purpose because I don't know exactly how Filestore operates, just that it's stored on raw nodes, *support* in communication protocols terminology is which type of lower layer packet carries your data, but I can just say that it's a format stored on a raw node.



> In “possible representations already mentioned, this is the only scenario”, replace the comma with a period or put the “this is the only scenario…” part in parentheses. That helps keep each sentence to one clear idea and makes them much more readable.
> 
> It also might be helpful to say “(see the go-ipfs/unixfs package for details of UnixFS.)” at the end. Or something like that to point to more details on UnixFS.

:+1:

importer/balanced/builder.go
```diff
@@ -1,24 +1,40 @@
...
 //
-// Balanced DAGs are generalistic DAGs in which all leaves
-// are at the same distance from the root.
+// The nodes are filled by recursion (in a DFS way) so the DAG is built
```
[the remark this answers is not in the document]
schomatis: In a manner similar to a depth-first search, I think that concept can be much more familiar to the developer than the DAG term (even if we repeated a lot, it's not that common compared to a basic tree search). But yes, the "bottom up" term might be more clear here, I'll put the depth-first search at the end of the sentence to not distract the reader from the main point.

importer/balanced/builder.go
```diff
@@ -1,24 +1,40 @@
...
+// from the bottom up. Leaf nodes are created first with the file data
+// and its size that is bubbled up to the parent (internal) node, that in
+// turn will aggregate all the file sizes of its children and bubble them
+// up to its parent (to keep propagating the total file size up to the
+// root). In this way the DAG acts like a B-tree (a more appropriate data
+// structure may be in order for this analogy), with each internal node
+// using the file size of its children as an index to perform the search
+// (seek) operation.
```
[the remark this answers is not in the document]
schomatis: Perfect.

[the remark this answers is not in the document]
schomatis: Then yes, I better lose the DFS term, because that will clash with the explanation of the actual search, which is not of a DFS type, I'll just put an emphasis on the recursive aspect of the construction (that doesn't belong to the search).

importer/balanced/builder.go
schomatis: Yes, the package doc should have a filled balanced DAG to make sure we leave a clear idea of what *balanced* implies, and this function can have an incomplete DAG in the process of being filled.

> The `Layout` function is responsible for creating new roots

Yes, that is a very important conceptual distinction, `Layout` is the only function that creates root nodes (and expands the DAG skeleton, while the `fill` call fleshes it out).

importer/balanced/builder.go
schomatis: Yes, that's exactly the type of in-progress creation I was writing about above, great!

importer/helpers/dagbuilder.go
schomatis: It is *definetely* not you (see https://github.com/ipfs/go-ipfs/issues/5056).

> ? Is that any better, or is it just confusing in a different way?

I'd rather rename the package to its full name and have `unixfs.FSNode` (and yes, the `FS` in `FSNode` is also a confusing redundancy, that's what namespaces are for).

importer/balanced/builder.go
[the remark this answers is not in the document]
Mr0grog: > This graph is actually the one you associated with the new explanation that talks about nodes… I'm interested in why you reached that conclusion.

Ah, sorry if I didn’t make that clear. Here’s the sticky bit:

> Then, a new root is created to point to the old root (incorporating it as its first child), which proceeds to be filled up

So according to that sentence, the new root is the the thing that “gets filled up (link) to more nodes.” That’s technically true, but since it doesn’t say anything about creating intermediary parents/roots/internals/whatever, it sounds like we are filling the rest of *the new root* with leaf nodes.

> I think the mirror term is the key

I do think “mirror” is useful, BUT I also wrote all of this comment before getting to the bottom where you wrote “Balanced DAGs are generalistic DAGs in which all leaves are at the same distance from the root,” which totally resolved the question for me. Just moving that from the bottom to be directly below this introductory paragraph (or below a diagram which is below the intro paragraph) would also be plenty clear, I think.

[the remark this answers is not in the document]
schomatis:
```

                                               +-------------+
                                               |   Root 6    |
                                               +-------------+
                                                     |
                          +--------------------------+----------------------------+
                          |                                                       | 
                    +-------------+                                         +-------------+
                    |   Root 3    |                                         |   Root 5    |
                    +-------------+                                         +-------------+
                          |                                                       |        
            +-------------+-------------+                           +-------------+ - - - -
            |                           |                           |              
     +-------------+             +-------------+             +-------------+       
     |   Root 1    |             |   Root 2    |             |   Root 4    |       
     +-------------+             +-------------+             +-------------+       
            |                           |                           |              
     +------+------+             +------+------+             +------+ - - - 
     |             |             |             |             |              
+=========+   +=========+   +=========+   +=========+   +=========+  
| Chunk 1 |   | Chunk 2 |   | Chunk 3 |   | Chunk 4 |   | Chunk 5 |  
+=========+   +=========+   +=========+   +=========+   +=========+  

```

I'm just realizing that most of the graphs have too many `Root` nodes, if by that we're meaning a node that was at some point a root node created by `Layout` (and has since become and internal node when the DAG grew in depth). `Root 1` is correct but `Root 3` is actually `Root 2` (if we're numbering in chronological order), the one drawn as 2 was never actually a root node, just an internal node generated by `fillNodeRec` function at the time that (what you're drawing as) `Root 3` was the root node and `Root 1` was just an internal node (`fillNodeRec` was mirroring that internal node that used to be a root). Is that clear? Maybe we could add a more descriptive legend, e.g.,

```
                    +-------------+              
                    |   Root 2    |              
                    +-------------+              
                          |                      
            +-------------+---------+            
            |                       |            
     +---------------+        +--------------+   
     |   Internal 1  |        |   Internal 3 |   
     |   (ex Root 1) |        |              |   
     +---------------+        +--------------+   
```

(BTW, have you been doing all these graphs manually? I just did a partial redrew of one of yours and became exhausted, we definitely need a tool for this.)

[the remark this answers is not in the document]
Mr0grog: > most of the graphs have too many Root nodes, if by that we're meaning a node that was at some point a root node created by Layout… the one drawn as 2 was never actually a root node

Ah, I didn’t mean anything that specific by root; only that it is the root of some sub-graph. I was trying not to *change* the name of a node between two diagrams that represent the graphs at different points in time and also trying not to introduce too much complexity. If you found that confusing, though, maybe it was a bad approach.

Maybe just “node?” (Still keep the leaves as “chunk,” though.) Not really sure.

```
     +-------------+
     |   Node 1    |
     +-------------+
            |
     +------+------+
     |             |
+=========+   +=========+   + - - - - + 
| Chunk 1 |   | Chunk 2 |   | Chunk 3 |
+=========+   +=========+   + - - - - +

                        ↯

                    +------------+
                    |   Node 2   |
                    +------------+
                          |       
            +-------------+-------------+
            |                           |
     +-------------+             +-------------+
     |    Node 1   |             |    Node 3   |
     +-------------+             +-------------+
            |                           |       
     +------+------+             +------+ - - -+
     |             |             |             |
+=========+   +=========+   +=========+   + - - - - + 
| Chunk 1 |   | Chunk 2 |   | Chunk 3 |   | Chunk 4 |
+=========+   +=========+   +=========+   + - - - - +
```

> Root 1 is correct but Root 3 is actually Root 2 (if we're numbering in chronological order)

Yeah, after reading the code, I realized this was the case. I was sort of numbering them chronologically, but it wasn’t clear from the description which was the right order (I don’t think the description actually needs to make it clear — it’s not important to a consumer of the package and would add unnecessary complex detail to a package-level explanation). It would be a good idea to renumber the chronologically anyway, though :)

> have you been doing all these graphs manually?

Yes, I have probably drawn too many weird ASCII diagrams at this point in my life :P

For diagrams like this, I’m not sure they’re so complex that finding and making a tool do what you want is really all that worthwhile (any more complex and it is, though). I will say that an editor with rectangular selections (TextMate, VS Code, etc.) is *essential,* though. You have to be able to pick up chunks of the graph and move them around, otherwise manipulating these graphs is insanely painful.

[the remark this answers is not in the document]
schomatis: > Ah, sorry if I didn’t make that clear. Here’s the sticky bit:
> 
> Then, a new root is created to point to the old root (incorporating it as its first child), which proceeds to be filled up
> 
> So according to that sentence, the new root is the the thing that “gets filled up (link) to more nodes.” That’s technically true, but since it doesn’t say anything about creating intermediary parents/roots/internals/whatever, it sounds like we are filling the rest of the new root with leaf nodes.
> 
> (The old explanation was clearer for me on this point because it says “incorporates a new child, which proceeds to be filled up.” It’s explicit about it being the new child that is filled.)

Yes, I see it now, thanks. I think I will also reserve the term *fill* for the parent of the leaf nodes to make that extra clear (although any node can be filled, but just for the sake of the explanation).



> Just moving that from the bottom to be directly below this introductory paragraph (or below a diagram which is below the intro paragraph) would also be plenty clear, I think.

Ok. I think the mirror term can be especially useful in the `Layout` explanation when we're talking about how DAGs are built, and leave the *balanced* term for the presentation of the full DAG as you said.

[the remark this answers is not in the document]
schomatis: > It would be a good idea to renumber the chronologically anyway, though :)

Yes, let's use correct numbering without explaining them (at least in the intro, maybe explain them in the `Layout` function, which might be read more by an interested developer than by the consumer of the package).

> For diagrams like this, I’m not sure they’re so complex that finding and making a tool do what you want is really all that worthwhile (any more complex and it is, though). I will say that an editor with rectangular selections (TextMate, VS Code, etc.) is essential, though. You have to be able to pick up chunks of the graph and move them around, otherwise manipulating these graphs is insanely painful.

Yes, I use the Sublime editor for that. My worry is not complexity but time consumption, as of now I haven't seen *any* diagrams like these in the code base, and if we want to incentive their incorporation we need to make it damn easy for the developer to do it, otherwise it won't be done; and we need to find a tool for that, so the developer can focus exclusively (or as much as possible) in what is being done (the layout of the DAG) instead how it will be done (the editing part, not everyone is an ascii ninja like you :)).

schomatis:
@Mr0grog Could you take a look at the documentation fixes in the [last commit](https://github.com/ipfs/go-ipfs/pull/5118/commits/ff2d69f770d6f9d46628ed0e1aacb0ceedf7ee71)?

I moved most of it to the `Layout` and `fillNodeRec` functions, leaving only a basic description of the balanced DAG in the package doc. Thinking about what you said, the consumer of the package will only be interested in *what* layout its data will be organized in, not *how* that layout is built, that should be left for someone reading the actual functions in charge of that.

schomatis:
I've made several changes to the comments of `fillNodeRec` you gave me, they weren't totally accurate with what the functions does, but my changes didn't end up being very convincing either, so take a closer look there please.

schomatis:
@Stebalien @whyrusleeping While we give the final touches to the documentation, could you take a look at the code please? Especially the added [`FSNodeOverDag`](https://github.com/ipfs/go-ipfs/pull/5118/files#diff-fdae772d2b35bc89d019b608d02c1755R366) and [`ProcessFileStore`](https://github.com/ipfs/go-ipfs/pull/5118/files#diff-fdae772d2b35bc89d019b608d02c1755R287).

importer/helpers/dagbuilder.go
```diff
@@ -202,12 +259,63 @@ func (db *DagBuilderHelper) GetNextDataNode() (*UnixfsNode, error) {
...
+	dataSize = uint64(len(fileData))
+
+	// Create a new leaf node containing the file chunk data.
+	node, err = db.NewLeafNode(fileData)
+	if err != nil {
+		return nil, 0, err
+	}
+
+	// Convert this leaf to a `FilestoreNode` if needed.
+	node = db.ProcessFileStore(node, dataSize)
```
whyrusleeping: the downside of doing things this way is that we hold the node in memory for longer than we really need to (technically just need to hash the data, could be done entirely on stack if go let us)

schomatis: I'm not following, *this way* refers to this PR or in general? From what I understand, we normally hold work-in-progress nodes to avoid constantly (un)packing while working on them, reducing computation time at the cost of more memory, but I'm not sure if you're referring to that.

schomatis: Yes, something that I'm seeing throughout the entire code is this mixture of node structures that are used interchangeably for (i) reading/writing and for (ii) building/cache. Ideally, if you are retrieving content (or storing what you received from another peer) the first type is just fine, but if you are building stuff yourself (and that is the very common operation of just adding content with `ipfs add`) then you need an entirely different object, that yes, will look like the first one (and the first one will even be its end product) but actually will behave very differently: it will cache many things while in the process of modifying them, it will have information in a mutable state, it will allow (what the first type would see as) invalid states.

Ideally again, I would expect to have different structures (with a common base or shared methods) to reflect that. And going back to the original raw node comment, yes, someone reading (or writing) the node needs the buffer, but someone building it may only need partial information like the hash.

Sorry, I'm not sure if you're original remark was going in this direction, but it made me think of this situation that I was encountering repeatedly while working on this PR.

schomatis:
@Stebalien PTAL

schomatis:
@Stebalien PTAL

importer/balanced/builder.go
```diff
@@ -1,113 +1,253 @@
...
+//  | Chunk 1 |   | Chunk 2 |   | Chunk 3 |    | Chunk 4 |
+//  +=========+   +=========+   + - - - - +    + - - - - +
+//
+// The `node` to be filled uses the `FSNodeOverDag` abstraction that allows adding
+// child nodes without packing/unpacking the UnixFS layer node (having an internal
+// `ft.FSNode` cache).
+//
+// It returns the `ipld.Node` representation of the passed `node` filled with
+// children and the `nodeFileSize` with the total size of the file chunk (leaf)
+// nodes stored under this node (to propagate it upwards in the DAG to its parent).
```
Mr0grog: Instead of:

> (to propagate it upwards in the DAG to its parent)

I might say:

> (parent nodes store this to enable efficient seeking through the DAG when reading data later)

I think it’s self-evident that the data is being propagated upward, and this describes a little more clearly why we want that information.

importer/balanced/builder.go
```diff
@@ -1,113 +1,253 @@
...
+	ft "github.com/ipfs/go-ipfs/unixfs"
 
 	ipld "gx/ipfs/QmWi2BYBL5gJ3CiAiQchg6rn1A8iBsrWy51EYxvHVjFvLb/go-ipld-format"
 )
 
-// Layout builds a balanced DAG. Data is stored at the leaves
-// and depth only increases when the tree is full, that is, when
-// the root node has reached the maximum number of links.
+// Layout builds a balanced DAG layout. In a balanced DAG of depth 1, leaf nodes
+// with data are added to a single `root` until the maximum number of links is
```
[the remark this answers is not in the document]
schomatis: Totally true, but I can't find any other place in the function for this information, `Layout` actually has only a few code lines, none of which could be easily associated with the two introductory paragraphs, the "magic" is happening in the recursion call, which is hiding *a lot* behind the scenes (something I'm not really a fan of). Anyway, I will have this advice in mind when writing documentation (that was actually my motivation when moving much of the package description inside the individual functions).

schomatis:
Thanks @Mr0grog for the *very* thorough review and *invaluable* diagrams!! This is already shaping up to be a nice reading reference for a new user who wants to get a deeper understanding of how files are handled in IPFS.

whyrusleeping:
In need of a rebase now that other PRs have landed

schomatis:
Fixed, waiting on the tests..

schomatis:
Broke the build.

whyrusleeping:
@schomatis whoops. You got a fix or should I handle?

schomatis:
@whyrusleeping Fixed in https://github.com/ipfs/go-ipfs/pull/5239, should be merged first.

schomatis:
@whyrusleeping RFM now

whyrusleeping:
@schomatis Woo! Thanks for all of these PRs. Making things easier to read and understand is so great

PR #5071, opened 2018-06-03
ci: Archive sharness trash dirs

magik6k:
With this CI will archive any leftover sharness trash directories, this should help debugging failing sharness tests.

Example: https://ci.ipfs.team/blue/organizations/jenkins/IPFS%2Fgo-ipfs/detail/feat%2Fci%2Farchive-trash-dirs/46/artifacts

PR #5119, opened 2018-06-14
disable the MacOS tests in jenkins

Stebalien:
They've never really been enabled but they:

1. Are causing the tests to fail for various reasons (e.g., out of space).
2. Take time.

### DIFF
--- a/ci/Jenkinsfile
+++ b/ci/Jenkinsfile
@@ -154,41 +154,41 @@ ansiColor('xterm') { withEnv(['TERM=xterm-color']) {
           }
         }
       },
-      macOS: {
-        setupStep('macos') { run ->
-          timeout(time: gotest_timeout, unit: 'MINUTES') {
-            run 'go get -v github.com/jstemmer/go-junit-report'
-            run "make gx-deps"
+      //macOS: {
+      //  setupStep('macos') { run ->
+      //    timeout(time: gotest_timeout, unit: 'MINUTES') {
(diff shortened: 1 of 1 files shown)


whyrusleeping:
cc @VictorBjelkholm ?

ci/Jenkinsfile
```diff
@@ -154,41 +154,41 @@ ansiColor('xterm') { withEnv(['TERM=xterm-color']) {
...
-      macSharness: {
-        setupStep('macos') { run ->
-          timeout(time: sharness_timeout, unit: 'MINUTES') {
-            run 'go get -v github.com/jstemmer/go-junit-report'
-            run "make gx-deps"
-
-            try {
-              run "make -j12 test/sharness/test-results/sharness.xml CONTINUE_ON_S_FAILURE=1 TEST_NO_FUSE=1"
-            } catch (err) {
-              throw err
```
VictorBjelkholm: Seems just commenting this and line 167 would give the wanted effect, rather than commenting the entire thing. But since @magik6k initially developed it, I'll leave it up to him to decide.

Although, removing the builds fully feels wrong, as we want to make sure it at least runs somewhat on macOS

magik6k: It's better to comment out the whole blocks imo

magik6k:
Will conflict a bit with https://github.com/ipfs/go-ipfs/pull/5071 but it's easy fix.

Issue #5133, opened 2018-06-18
Feature: "children" pinning mode


hsanjuan:
#### Description:

I have discussed this with someone but apparently it was not written down. IPFS Cluster needs IPFS to pin partial trees, that is, a root CID and its immediate children (this is part of the sharding feature). Thus, I'll be working to add this feature to go-ipfs. I will reference this issue in upcoming PRs.

kevina:
Moved from #5134:

Rather than complicate our pinner one potential solution is to enhance our notation of `best-effort` pins and fetch the needed subtrees separately.

Stebalien:
So, I *would* like some kind of "download manager" to track download progress in MFS/best-effort pins. The main issue here would be accidentally downloading nodes we don't want to keep but keeping them anyways. @kevina what did you have in mind?

---

Also, talking this over with @whyrusleeping, I'd like to do the following:

1. Switch to the refmt based go-ipld-cbor
2. Finalize go-ipld-hamt (requires 1).
3. Store the pinset in go-ipld-hamt.

This would allow us to easily add extra metadata to pins and git rid of 90% of our current logic.

---

What can be done now:

* Depth-limited traversal
* Depth-limited `ipfs refs -r`

hsanjuan:
I'll take on depth limited traversal and `refs -r`.

0zAND1z:
Hi all,

I have been redirected to this issue, cited as a blocking item, to enable the DAG Sharding support in the `ipfs-cluster`.

Hoping that the information is still relevant, I would like to seek some clarity on the following questions:

### **Question 1:** Is the purpose of depth-limit to enable the user to modulate the number of shards to be generated for a given file?   
Consider a scenario where a file bearing a multihash `X` is to be pinned across the cluster be divided into an _**S**_ number of shards(say `x1`, `x2`, `x3` ... `xS`). Does the depth-limit support the creation of these number of shards? Is it necessary that all the **_S_** number of shards be persisted in minimum one node, or could these unique shards be distributed across the cluster?

### **Question 2:** Are the shards subjected to distribution and overlapping for fault tolerance?  
Consider the same file bearing a multihash `X` is to be pinned across the cluster be divided into an **_S_** number of shards(say `x1`, `x2`, `x3` ... `xS`) in an overlapping manner across the **_M_** number of nodes in the cluster? Is this planned & made possible under the scope of the current issue?

All help in answering these questions are appreciated. Thanks.

[...]

PR #5134, opened 2018-06-18
Feat: FetchGraphToDepth() to fetch a graph to a given depth

hsanjuan:
This comes in the context of #5133. It enables Merkledag to fetch
DAGs down to a given depth.

Note that actual usage of depth is expected to be 1, or 2 (and not
an arbitrarily high value), thus I have opted to not complicate things with
branch-pruning optimizations. They can be introduced if they are ever needed
at another point in time.


hsanjuan:
Superseeded by #5142

PR #5137, opened 2018-06-19
Add system requirements to README

### LINKED ISSUE #5136: Add resource requirements to README
People ask every so often (and it seems pretty reasonable for them to do so!) what the basic requirements are for running IPFS. We should add this to the README. (As recommendations, not hard requirements?)

From @whyrusleeping on IRC:

> ah, for go-ipfs, i generally recommend at least one CPU core and 2GB of ram
> though things work decently well on just 1GB
> one of our main bootstrappers has a little less than 1GB
> it restarts itself every few days

I’m happy to write this up as a PR; I just can’t self-assign in this repo :)

Mr0grog:
This adds a short paragraph about system requirements to the “install” section of the README (that seemed like the most applicable place). While writing this, I wondered if it might make sense to add a paragraph about disk space:

> You should also make sure to have about 1.3x the disk space as the amount of data you’d like to be able to store — IPFS takes up a little bit of extra space keeping metadata about your files and about peers on the network. You can also choose [different ways to store your data](https://github.com/ipfs/go-ipfs/blob/master/docs/datastores.md), which will impact the amount of storage space required.

…but I wasn’t sure if that was actually necessary to say or even very accurate (There’s definitely a lot of fudge factor here, and I’m not even sure the 1.3x number is even a great reasonable approximation), so I left it out. I can add it in if people think it’s useful, though.

Fixes #5136.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ Please put all issues regarding the Go IPFS _implementation_ in [this repo](http
 
 - [Security Issues](#security-issues)
 - [Install](#install)
+  - [System Requirements](#system-requirements)
   - [Install prebuilt packages](#install-prebuilt-packages)
   - [From Linux package managers](#from-linux-package-managers)
   - [Build from Source](#build-from-source)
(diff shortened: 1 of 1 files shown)


Stebalien:
> …but I wasn’t sure if that was actually necessary to say or even very accurate (There’s definitely a lot of fudge factor here, and I’m not even sure the 1.3x number is even a great reasonable approximation), so I left it out. I can add it in if people think it’s useful, though.

So, IPFS actually tends to use *less* space, but that really depends on what you're storing. I'd just say that IPFS may use more or less space, depending on what you're storing.

Issue #4763, opened 2018-03-04
Moving away from current pin system


Voker57:
As suggested by @whyrusleeping in #4757

Current pin system has several problems, as I see them:

- Pins are stored in elaborate way in blockstore, enabling whole world to sniff your pins and introducing great deal of complexity to code like internal pins and global pin locking
- Pins are unnamed, so you can't structure/review them easily

Suggestions & reviews are welcome

My proposed PR: #4757
Another suggestion: #4675

hsanjuan:
The main considerations I can give as a pin-system user, out of the top of my head:

* Global locking. Add this to bitswap bugs and you end up perpetually locked.
* Can't enable/disable GC via the API. Fetching without pinning, or putting blocks without/before pinning, to avoid the pinning hurdles, leaves you exposed to GC wiping your content.
* No names (cluster added names for pins but it wouldn't hurt)
* pin/ls --recursive needs to be fast with huge lists
* Unpin should cancel any ongoing pins for the same cid, if things can happen in parallel. 
* Async pinning would be nice to have:
  * I tell ipfs to pin something
  * I check back later to see if it's still pinning, how far it is (size fetched).


> enabling whole world to sniff your pins

Is there anything different as to how the whole world can sniff other non-pinned stuff?

b5:
I'd resurface @Kubuxu's [comment](https://github.com/ipfs/go-ipfs/pull/4757#issuecomment-370238827) raised on #4757 pointing out this would be a great candidate for the RFC process. I'd _love_ to see some of the changes mentioned here, but knowing where pinning is headed in the long run is far more important IMHO

singpolyma:
Async pinning would be a *huge* win

Kubuxu:
Some thoughts:

I think that having pins under MFS (files api) would be very nice feature. This would get us sharding for free, help with pin exploration and gets us the names.
We can significantly reduce complexity of GC by removing direct pins and allowing either to allow pinned items to be pinned strongly or weakly (all content must be here vs keep whatever we have). This would allow for implementation of concurrent and/or incremental GC solving locking problem.

Also regarding GC, we could have `/timed-tmp` in MFS that would keep files there for X time and after that they would be removed. Files would be added them when for example: accessed from gateway, added and so on. They would be automatically removed from there after X time. (Just some wild idea).

In case of using MFS for pinning, under the hood we would: 1. copy the file to pin MFS; 2. start fetch of MFS directory (this operation is not written but should be simple enough).

[...]

Issue #4675, opened 2018-02-09
Store pins in MFS


Stebalien:
Motivations:

* Free HAMT. We can update the pinset without having to load/store the *entire* pinset. See #4673.
* *One* pinset. This should simplify GC.
* Allows use to deprecate (but not remove) the `ipfs pin` command while not having to maintain a bunch of deprecated code.

Blockers:

* An `ipfs files cp --prefetch` flag, maybe even a `--prefetch=background` flag?
* A way to not pin recursively in MFS. This will be tricky... We could have some form of side table but I'd rather not. Really, I'd like pin/prefecth policies in unixfs. However, this is *hard*.
* Link to IPLD from IPFS.
* Probably https://github.com/ipfs/ipld-unixfs.

kevina:
@Stebalien I assume you already know this, but right now anything under the MFS is pinned via a best-effort policy.  What that means is that it won't be gc, but removing it or any of its children won't be blocked.

Maybe a recursive pin could be implemented as a sort of read-only flag.  That is removing anything with the flag will be blocked until the flag is removed, although there are probably a lot of implementation details that need to be worked out, for example the handling of indirect pins.

Now _direct_ pins don't really fit into this model as the gc is free to remove any direct pins children.

Stebalien:
For context, one of my goals here is to make IPFS usable for DAPPs. For this to happen, we need to be able to have namespaces and I'd like to use MFS for that (a nice, simple, single namespace).

> What that means is that it won't be gc, but removing it or any of its children won't be blocked.

You mean by `ipfs block rm`? It would be nice to split this into "always delete X" and "gc X if uneeded". The current middle-ground seems a bit weird. The two usecases I can see are:

1. Free memory (gc).
2. Remove bad bits (force remove).

Proposal:

* Move `ipfs block rm` to `ipfs repo purge` (intentionally vicious). Putting it on `block` ends up confusing users into thinking that they can actually delete blocks from other machines. Note: we'd have to maintain backwards compat by providing an alias (and hiding the help?).
* Make `ipfs repo purge` force delete the block.
* Introduce the ability to pass a list of things to gc to `ipfs repo gc`. That is, allow `ipfs repo gc things...`. This is *usually* what the user wants, IMO.

> Now direct pins don't really fit into this model as the gc is free to remove any direct pins children.

[...]

PR #5142, opened 2018-06-20
Feat: pin add --max-depth (arbitrary depth recursive pins)

hsanjuan:
This implements #5133 introducing an option to limit how deep we fetch and store
the DAG associated to a recursive pin ("--max-depth"). This feature
comes motivated by the need to fetch and pin partial DAGs in order to do
DAG sharding with IPFS Cluster.

This means that, when pinning something to --max-depth, the DAG will be
fetched only to that depth and not more.

In order to get this, the PR introduces new recursive pin types: "recursive1"
means: the given CID is pinned along with its direct children (maxDepth=1)

"recursive2" means: the given CID is pinned along with its direct children
and its grandchildren.

And so on...

This required introducing "maxDepth" limits to all the functions walking down
DAGs (in merkledag, pin, core/commands, core/coreapi, exchange/reprovide modules).

maxDepth == -1 effectively acts as no-limit, and all these functions behave like
they did before.

In order to facilitate the task, a new CID Set type has been added:
thirdparty/recpinset. This set carries the MaxDepth associated to every Cid.
This allows to shortcut exploring already explored branches just like the original
cid.Set does. It also allows to store the Recursive pinset (and replaces cid.Set).
recpinset should be moved outside to a different repo eventually.

TODO: tests
TODO: refs -r with --max-depth

License: MIT
Signed-off-by: Hector Sanjuan <code@hector.link>

### DIFF
--- a/merkledag/merkledag.go
+++ b/merkledag/merkledag.go
@@ -7,6 +7,7 @@ import (
 	"sync"
 
 	bserv "github.com/ipfs/go-ipfs/blockservice"
+	"github.com/ipfs/go-ipfs/thirdparty/recpinset"
 
 	ipldcbor "gx/ipfs/QmSF1Ksgn5d7JCTBt4e1yp4wzs6tpYyweCZ4PcDYp3tNeK/go-ipld-cbor"
 	blocks "gx/ipfs/QmTRCUvZLiir12Qr6MV3HKfKMHX8Nf1Vddn6t2g5nsQSb9/go-block-format"
@@ -159,25 +160,32 @@ func (n *dagService) Session(ctx context.Context) ipld.NodeGetter {
 
 // FetchGraph fetches all nodes that are children of the given node
 func FetchGraph(ctx context.Context, root *cid.Cid, serv ipld.DAGService) error {
+	return FetchGraphMaxDepth(ctx, root, -1, serv)
+}
+
+// FetchGraphMaxDepth fetches all nodes that are children to the given node
+// down to the given depth. maxDetph=0 means "only fetch root", maxDepth=1 means
+// fetch root and its direct children and so on... maxDepth=-1 means unlimited.
+func FetchGraphMaxDepth(ctx context.Context, root *cid.Cid, maxDepth int, serv ipld.DAGService) error {
 	var ng ipld.NodeGetter = serv
 	ds, ok := serv.(*dagService)
 	if ok {
 		ng = &sesGetter{bserv.NewSession(ctx, ds.Blocks)}
 	}
 
+	set := recpinset.New()
 	v, _ := ctx.Value(progressContextKey).(*ProgressTracker)
 	if v == nil {
-		return EnumerateChildrenAsync(ctx, GetLinksDirect(ng), root, cid.NewSet().Visit)
+		return EnumerateChildrenAsyncMaxDepth(ctx, GetLinksDirect(ng), root, maxDepth, set.Visit)
 	}
-	set := cid.NewSet()
-	visit := func(c *cid.Cid) bool {
-		if set.Visit(c) {
+	visit := func(c *cid.Cid, maxDepth int) bool {
+		if set.Visit(c, maxDepth) {
 			v.Increment()
 			return true
 		}
 		return false
 	}
-	return EnumerateChildrenAsync(ctx, GetLinksDirect(ng), root, visit)
+	return EnumerateChildrenAsyncMaxDepth(ctx, GetLinksDirect(ng), root, maxDepth, visit)
 }
 
 // GetMany gets many nodes from the DAG at once.
@@ -255,14 +263,39 @@ func GetLinksWithDAG(ng ipld.NodeGetter) GetLinks {
 // unseen children to the passed in set.
 // TODO: parallelize to avoid disk latency perf hits?
 func EnumerateChildren(ctx context.Context, getLinks GetLinks, root *cid.Cid, visit func(*cid.Cid) bool) error {
+	visitDepth := func(c *cid.Cid, maxDepth int) bool {
+		return visit(c)
+	}
+	return EnumerateChildrenMaxDepth(ctx, getLinks, root, -1, visitDepth)
+}
+
+// EnumerateChildrenMaxDepth walks the dag below the given root to the given
+// depth. The root is at level 0, the children are at level 1 and so on.
+// Thus, setting depth to two, will walk the root, the children, and the
+// children of the children.
+// Setting depth to a negative number will walk the full tree.
+func EnumerateChildrenMaxDepth(ctx context.Context, getLinks GetLinks, root *cid.Cid, maxDepth int, visit func(*cid.Cid, int) bool) error {
+	if maxDepth == 0 {
+		// Root nodes are not marked as visited in this implementation
+		// (they are in the async version)
+		return nil
+	}
+
+	if maxDepth > 0 {
+		maxDepth--
+	}
+
 	links, err := getLinks(ctx, root)
 	if err != nil {
 		return err
 	}
 	for _, lnk := range links {
 		c := lnk.Cid
-		if visit(c) {
-			err = EnumerateChildren(ctx, getLinks, c, visit)
+		if visit(c, maxDepth) {
+			// Note, visit() returns true when maxDepth
+			// is > than existing value (meaning we have to
+			// go deeper than before.
+			err = EnumerateChildrenMaxDepth(ctx, getLinks, c, maxDepth, visit)
 			if err != nil {
 				return err
 			}
@@ -306,8 +339,24 @@ var FetchGraphConcurrency = 8
 //
 // NOTE: It *does not* make multiple concurrent calls to the passed `visit` function.
 func EnumerateChildrenAsync(ctx context.Context, getLinks GetLinks, c *cid.Cid, visit func(*cid.Cid) bool) error {
-	feed := make(chan *cid.Cid)
-	out := make(chan []*ipld.Link)
+	visitDepth := func(c *cid.Cid, maxDepth int) bool {
+		return visit(c)
+	}
+	return EnumerateChildrenAsyncMaxDepth(ctx, getLinks, c, -1, visitDepth)
+}
+
+// EnumerateChildrenAsyncMaxDepth is equivalent to EnumerateChildrenMaxDepth *except* that
+// it fetches children in parallel (down to a maximum depth in the graph).
+//
+// NOTE: It *does not* make multiple concurrent calls to the passed `visit` function.
+func EnumerateChildrenAsyncMaxDepth(ctx context.Context, getLinks GetLinks, c *cid.Cid, maxDepth int, visit func(*cid.Cid, int) bool) error {
+	type linksDepth struct {
+		links    []*ipld.Link
+		maxDepth int
+	}
+
+	feed := make(chan *recpinset.RecPin)
+	out := make(chan *linksDepth)
 	done := make(chan struct{})
 
 	var setlk sync.Mutex
@@ -319,20 +368,37 @@ func EnumerateChildrenAsync(ctx context.Context, getLinks GetLinks, c *cid.Cid, 
 
 	for i := 0; i < FetchGraphConcurrency; i++ {
 		go func() {
-			for ic := range feed {
+			for recPin := range feed {
+				maxDepth := recPin.MaxDepth
+
 				setlk.Lock()
-				shouldVisit := visit(ic)
+				// Note, visit returns true when depth
+				// is > than existing value (meaning we have to
+				// go deeper than before.
+				shouldVisit := visit(recPin.Cid, maxDepth)
 				setlk.Unlock()
 
-				if shouldVisit {
-					links, err := getLinks(ctx, ic)
+				switch {
+				case maxDepth == 0:
+					// done
+				case shouldVisit:
+					if maxDepth > 0 {
+						maxDepth--
+					}
+
+					links, err := getLinks(ctx, recPin.Cid)
 					if err != nil {
 						errChan <- err
 						return
 					}
 
+					outLinks := &linksDepth{
+						links:    links,
+						maxDepth: maxDepth,
+					}
+
 					select {
-					case out <- links:
+					case out <- outLinks:
 					case <-fetchersCtx.Done():
 						return
 					}
@@ -347,10 +413,13 @@ func EnumerateChildrenAsync(ctx context.Context, getLinks GetLinks, c *cid.Cid, 
 	defer close(feed)
 
 	send := feed
-	var todobuffer []*cid.Cid
+	var todobuffer []*recpinset.RecPin
 	var inProgress int
 
-	next := c
+	next := &recpinset.RecPin{
+		Cid:      c,
+		MaxDepth: maxDepth,
+	}
 	for {
 		select {
 		case send <- next:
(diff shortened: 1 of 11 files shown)


kevina:
Please note that we also have the notion of a "best effort" pin used to pin anything off the files root.  It works by keeping the 'GC' from removing anything under it but the GC won't fail if one of the children can not be found.  I am thinking we should just make that pin type explicit.

@hsanjuan does that type of pin not meet your needs?

kevina:
Also if we go trough we this, I want to avoid the special case when the recursion is 0 by defining a direct pin as a recursive pin with a depth of 0.

kevina:
> I don't think it does, honestly I don't fully understand it. I need go-ipfs to NOT fetch all the subtree.

Creating a best-effort pin won't fetch anything.  It will simply prevent any subtrees already fetched from being garbage collected.

Stebalien:
So, we really have three things here:

1. A service that tries to pin data in the background.
2. Some way to block GC unless we can be sure we won't delete something we need (i.e., checks pins before it allows GC to run).
3. Pins (

hsanjuan:
@kevina I'm thinking that deleting Direct pins as a fully different set might have some performance impact (in order to keep current APIs), as direct pins will need to be extracted from the recursive set by filtering them out, thus listing direct pins may become a slower operation if the recursive set is very big (something that doesn't happen now). Is this something acceptable?

hsanjuan:
> Creating a best-effort pin won't fetch anything. It will simply prevent any subtrees already fetched from being garbage collected.

I need to fetch partial subtrees too.

kevina:
> I need to fetch partial subtrees too.

I am thinking this should be implemented directly without complicating the pinier or the `gc`.  You can then create a `best-effort` pin to keep them from getting GC.

However, I am not really against this idea if it will be useful in other contextes.  @Stebalien @whyrusleeping what do you think.

kevina:
> @kevina I'm thinking that deleting Direct pins as a fully different set might have some performance impact (in order to keep current APIs), as direct pins will need to be extracted from the recursive set by filtering them out, thus listing direct pins may become a slower operation if the recursive set is very big (something that doesn't happen now). Is this something acceptable?

I would think so.  But others may disagree.

hsanjuan:
@Kubuxu @whyrusleeping can I get some attention on this?

hsanjuan:
@Stebalien maybe you can help with this?

Stebalien:
Let's go back to the issue and discuss ways to do this right. While this technically addresses the immediate need, it won't solve the long-term issue of needing more flexible ways to specify pins and it abuses an enum to store a number. Really, we need a more flexible pinset that allows for complex pin policies.

hsanjuan:
@Stebalien I am a bit dissapointed it took 1 month to get a full frontal rejection with little alternative proposal. I understand that the "right" way to do this:

* Is a larger change
* Touches more critical sections
* Is all unclear at this point
* Would throw away all the hacks in the current pinning system (mine, and previously existing) anyways

My request is clearly spec-ed and implementable within the current state of things. So instead of deferring to larger abstract change, I would like to get a list of concrete steps that we can take to move forward, possibly in parallel with a the long total-revamp-of-the-pin-system discussion.

For example, this change, provides you with a bunch of areas to discuss on already that would also clarify parts in the larger discussion:

* There are a bunch of functions depth-first-traversing the DAG. It does not seem crazy to update them with a depth limit as I did. The fact that there are 4 or 5 of these, means more lines of code. The bulk of the change is this. Are we against that? This can be PRed separately.
* There is the commands part, with flags defining an API and expected behaviour that we can agree on or not. What would be response format for depth-limited pins? Extra key in the response object? Or are we ok with type "recursive%d"?
* The package in thirdparty should be moved out
* There is a "enum abuse". Or extending the recursive types to account for arbitrary max-depths. This is a hack. What would be the alternative? It needs to keep the API compatible. I did not come up with a way of doing it in less amounts of code, or touching less critical sections, but yes it's a hack. We can add a new single `MaxDepth` pin type, but we'll have to make sure we carry not just the type but metadata associated to it (the actual depth) all around.
* Finally, the last part of the change affects how the new pins are stored in disk. Again, cannot think how to do it in less amounts of code or touching less critical paths. Since most --max-depth values will be 1 or 2, it made sense to do it like this.

Despite the critic, the change does not get much in the way of the current pinning system (it does not change the logic, the api, the storage format) and the standard pinning path.

There is another way of approaching this too. I only need max-depth 1 and 2 (or maybe even reducing to just 1). I can just support those values, introducing 2 specific pin types and potentially reducing the parts of the "hack" to support arbitrary pin depths.

Is there a way of doing this now that you'd consider workable?

Stebalien:
> I am a bit dissapointed it took 1 month to get a full frontal rejection with little alternative proposal.

It's a large patch, touches a bunch of critical code, hasn't been flagged as a priority, and the design wasn't discussed at all before implementing. I don't even look at patches like this until I have some time to actually think about them.

> little alternative proposal

I don't know enough about the current pin system to give one off the top of my head.

> Despite the critic, the change does not get much in the way of the current pinning system (it does not change the logic, the api, the storage format) and the standard pinning path.

This patch adds a cluster-specific, weird addition to pinning by hacking it into the the existing pin system with no discussion. Whatever feature we end up adding to support depth-limited pins, *we'll* have to **maintain** it.

The primary issue here is the lack of discussion and/or context. You should probably read:

* https://github.com/ipfs/go-ipfs/issues/4675
* https://github.com/ipfs/go-ipfs/pull/4757
* https://github.com/ipfs/go-ipfs/issues/4763

---

> There are a bunch of functions depth-first-traversing the DAG. It does not seem crazy to update them with a depth limit as I did. The fact that there are 4 or 5 of these, means more lines of code. The bulk of the change is this. Are we against that? This can be PRed separately.

It should be possible to achieve the same thing by passing the depth to the visit function (can create a new `EnumerateChildrenWithDepth`). EnumerateChildrenWithDepth shouldn't have to care about the max depth (the visit function should just return false when it hits the max depth).

Whatever we do, those functions shouldn't talk about pinning or use types like `RecPin`. That'll just confuse readers.

> There is the commands part, with flags defining an API and expected behaviour that we can agree on or not. What would be response format for depth-limited pins? Extra key in the response object? Or are we ok with type "recursive%d"?

Personally, I'd give them the type "partial" and then add a "max-depth" field, or something like that.

> The package in thirdparty should be moved out.

Yes.

> There is a "enum abuse". Or extending the recursive types to account for arbitrary max-depths. This is a hack. What would be the alternative? It needs to keep the API compatible. I did not come up with a way of doing it in less amounts of code, or touching less critical sections, but yes it's a hack. We can add a new single MaxDepth pin type, but we'll have to make sure we carry not just the type but metadata associated to it (the actual depth) all around.

We (and cluster) will need more complex pins anyways.

> Finally, the last part of the change affects how the new pins are stored in disk. Again, cannot think how to do it in less amounts of code or touching less critical paths. Since most --max-depth values will be 1 or 2, it made sense to do it like this.

The values will usually be 1 or 2 **for cluster**.

kevina:
@hsanjuan @Stebalien I attempted to start a discussion about alternative ways to solve this problem, but it seams I was ignored.

Stebalien:
> In particular it would likely be better to enhance our best-effort pins and then fetch the needed subtrees separately.

We just need to make sure that fills the need. If we do that, we *won't* end up removing any accidentally downloaded nodes.

Let's move the discussion back to the issue.

thirdparty/recpinset/recpinset.go
```diff
@@ -0,0 +1,130 @@
...
+
+// Visit adds a cid and maxDepth to the set if:
+// - it is not already in the set
+// - if it's in the set but the new maxDepth is greater than
+// the existing.
+// It returns true if the RecPin has been added.
+func (s *Set) Visit(c *cid.Cid, maxDepth int) bool {
+	curMaxDepth, ok := s.set[string(c.Bytes())]
+
+	if !ok || IsDeeper(maxDepth, curMaxDepth) {
```
whyrusleeping:
```
[ A ]
  |     \
[ B ]  [ C ]
 |          |
[ C ]  [ D ]
```

So what happens if we visit C in the first (left) tree? It seems like we would call visit on the second one, and pass `1,2` to `IsDeeper` which would return false, and cause us to not ever visit D.

whyrusleeping:
@hsanjuan let's do the depth limited recursion stuff in a separate PR, and discuss pinning improvements where @kevina and @Stebalien are mentioning. 

Sorry for the wait, but let's sketch this out a bit more before moving forward.

merkledag/merkledag.go
```diff
@@ -255,14 +263,39 @@ func GetLinksWithDAG(ng ipld.NodeGetter) GetLinks {
...
+	}
+	return EnumerateChildrenMaxDepth(ctx, getLinks, root, -1, visitDepth)
+}
+
+// EnumerateChildrenMaxDepth walks the dag below the given root to the given
+// depth. The root is at level 0, the children are at level 1 and so on.
+// Thus, setting depth to two, will walk the root, the children, and the
+// children of the children.
+// Setting depth to a negative number will walk the full tree.
+func EnumerateChildrenMaxDepth(ctx context.Context, getLinks GetLinks, root *cid.Cid, maxDepth int, visit func(*cid.Cid, int) bool) error {
```
Stebalien: A more general way to do this (and the async version) would be to:

1. Make the visit function take a *current* depth. I'd *like* to pass a full path (as a `[]string`) but that might get expensive.
2. Have the visit function determine if we should go deeper.

thirdparty/recpinset/recpinset.go
hsanjuan: Sorry, but that's not correct.

Assuming you are traversing this graph with MaxDepth=2 at the beginning:

* The C on the first branch would be visited with maxDepth=0 `(c,0)`
* The C on the second branch would be visited with MaxDepth=1 (c,1)`

1>0, thus Visit will return `true`, thus the functions will keep traversing the path because the previously visited C had a lower depth limit than the new one. 

In this context `maxDepth` means, `the maximum depth of the tree below this CID`. Thus we always keep exploring if `IsDeeper()`. It does not mean `item's depth` as you seemed to assume.

merkledag/merkledag.go
hsanjuan: > Have the visit function determine if we should go deeper.

But this is what is does. You can either remember how far the item is to the limit (maxDepth, as it does now), or you can remeber how deep the item is and what the absolute depth limit is. The way it's done now requires only remembering one thing.

hsanjuan:
@Stebalien 

> This patch adds a cluster-specific, weird addition to pinning by hacking it into the the existing pin system with no discussion. Whatever feature we end up adding to support depth-limited pins, we'll have to maintain it.

This is the discussion. This PR is here to kickstart a discussion: how to support this feature in the current pin system. It was told to me (offline) that I could attempt to do this in the current pin system until the whole MFS thing is figured out. Now you have a bunch of concrete stuff to criticize and propose improvements on, a list of all the pieces that are potentially touched and the minimal approach to it (hacky, yes, but minimal). Sorry but I thought actually understanding the current pin system and propose how to change it was a better approach than nicely asking to someone else to write my feature.

That said, I'll gladly maintain it too whatever the outcome is. Sorry I don't like the stress on the "*we'll* have to maintain it". It leaves me with very mixed feelings about what you're implying and makes me a little sad.

> It should be possible to achieve the same thing by passing the depth to the visit function (can create a new EnumerateChildrenWithDepth). EnumerateChildrenWithDepth shouldn't have to care about the max depth (the visit function should just return false when it hits the max depth).

I think you cannot shortcut branches from the search by knowing only the current depth. You need to know how deep you explored the last time you visited the CID (I tried that way first). At least I could not figure out how to do it just with depth while at the same time not re-exploring branches.

> Whatever we do, those functions shouldn't talk about pinning or use types like RecPin. That'll just confuse readers.

Happy to rename.

> We (and cluster) will need more complex pins anyways.

> The values will usually be 1 or 2 for cluster.

That's fair to say, but since cluster is requesting this feature and is very limited in scope compared to "complex pins", it should make sense that it works for well for the cluster usecase. It can always be improved in the future as more use cases and needs appear. Perhaps having added a single "only children/maxdetph=1" pin type would have been a better approach to this. (as a side not, it would be interesting to hear use-cases of people pinning large amounts of pins all with different depths limits, but I agree this should eventually work well).

@whyrusleeping 

> @hsanjuan let's do the depth limited recursion stuff in a separate PR

Sure, no problem.

Stebalien:
> This is the discussion.

This PR was couched as a finished product. No WIP, "proposal", open design questions, "should we do this", etc. and the remaining TODOs are "write more tests" and "implement refs -r --max-depth". That's why I reacted the way I did. A fair amount of your time was wasted writing this patch and some (although significantly less) of my time was wasted reviewing it.

> That said, I'll gladly maintain it too whatever the outcome is. Sorry I don't like the stress on the "we'll have to maintain it". It leaves me with very mixed feelings about what you're implying and makes me a little sad.

By "we", I meant all of us. My point was that this is a group effort. And no, you won't (and shouldn't) handle every bug report possibly related to this PR.

---

>> It should be possible to achieve the same thing by passing the depth to the visit function (can create a new EnumerateChildrenWithDepth). EnumerateChildrenWithDepth shouldn't have to care about the max depth (the visit function should just return false when it hits the max depth).

> I think you cannot shortcut branches from the search by knowing only the current depth. You need to know how deep you explored the last time you visited the CID (I tried that way first). At least I could not figure out how to do it just with depth while at the same time not re-exploring branches.

Correct. As this patch currently does, the visit function would have to remember how deep it was when it last explored a CID.

However, that means that you can:

1. Only track the *current* depth in `EnumerateChildrenWithDepth` function.
2. Decide whether or not to go deeper in the visit function (based on the current depth, the max depth (local to the visit function), and the lowest depth at which we've explored the current CID (visit function state).

thirdparty/recpinset/recpinset.go
whyrusleeping: Ah, these variables could definitely use some better naming then. and maybe a comment. maybe `maxDepth` -> `curHeight` ?

Stebalien: It is actually `maxDepth`. That is, it's the "max depth" at which we are planning on exploring this CID. When we *increase* the `maxDepth`, we explore the path again.

hsanjuan:
> Decide whether or not to go deeper in the visit function (based on the current depth, the max depth (local to the visit function), and the lowest depth at which we've explored the current CID (visit function state).

Yes, ok that works too and I see the advantage now. Thanks for taking the time to explain.

hsanjuan:
> Are we confident the future solution will support exactly this behavior?

As long as there's an equivalent way of doing things, I'm happy with breaking changes (but this opens a bad, or weird precedent at least, even if it's good for me). Anyway, as you said,  let's do the things we agree on first and revisit the discussion.

achingbrain:
Is this going to be revisited in the future?

From what I understand the ability to pin subsections of graphs is blocking cluster shipping their feature of splitting an unreasonably large DAG across multiple IPFS nodes - this would very much help package manager maintainers increase availability of their registries if stored on IPFS.

lanzafame:
@achingbrain hopefully as this is still blocking the sharding functionality in IPFS Cluster.

sashahilton00:
Is there any news/progress on this issue over the past few months? Whilst I don't was to be 'that guy' who asks others to implement features, ipfs-cluster is currently pretty useless for clusters over a few TB with large files inside due to the requirement to replicate entire files. 

Whilst likely not possible for a while due to the need for me to familiarise myself with the technicalities of IPFS under the hood and the codebase, if there are no plans to revisit this in the near future, I will try to start learning about IPFS and thinking about some of the problems in previous comments, with a view to hopefully writing this feature in X months time. Though whilst considering that option, there seem to be a few mentions that the entire pin system is due for a refactor - is this happening anytime soon, as it makes thinking about an implementation for this feature harder if this is the case.

Issue #5074, opened 2018-06-04
UnixFS protocol buffer message definition


schomatis:
As said in the UnixFS [specification](https://github.com/ipfs/specs/tree/master/unixfs) *there is no clear specification* document and one needs to  examine reference implementations (such as [the one on this repo](https://github.com/ipfs/go-ipfs/tree/1e9e2f453c435272433a96ebfba5c4319cc08534/unixfs)) to understand the format.

The [protocol buffer message definition](https://github.com/ipfs/go-ipfs/blob/1e9e2f453c435272433a96ebfba5c4319cc08534/unixfs/pb/unixfs.proto) has no comments and ambiguous names like `Data`, which is actually used twice (as the name of the message  and also one of its types) leading to *very* confusing code lines such as this one:

https://github.com/ipfs/go-ipfs/blob/1e9e2f453c435272433a96ebfba5c4319cc08534/unixfs/unixfs.go#L190

I don't think we can change any of the identifier names in the `.proto` file without breaking the API but at least we should document it as much as possible to avoid future confusions.

Also I would like clarify exactly what is the connection to the `FSNode` structure (@Stebalien help),

https://github.com/ipfs/go-ipfs/blob/1e9e2f453c435272433a96ebfba5c4319cc08534/unixfs/unixfs.go#L142-L154

which seems to be the counterpart of the protocol-buffer generated structure,

https://github.com/ipfs/go-ipfs/blob/1e9e2f453c435272433a96ebfba5c4319cc08534/unixfs/pb/unixfs.pb.go#L71-L79

but I'm failing to understand where do the responsibilities of one end and the other one start.

schomatis:
One confusing relationship between those structures is how are sizes handled.

* `Data.Filesize`: Size of the data that the structure is holding *plus* the file sizes of its children stored at `Data.subtotal`,

https://github.com/ipfs/go-ipfs/blob/1e9e2f453c435272433a96ebfba5c4319cc08534/unixfs/unixfs.go#L188

* `FSNode.subtotal`: Sum of child nodes `Data.Filesize` that keeps being updated as more children are added,

https://github.com/ipfs/go-ipfs/blob/1e9e2f453c435272433a96ebfba5c4319cc08534/unixfs/unixfs.go#L173-L176

[...]

PR #5153, opened 2018-06-25
[WIP] unixfs: document protocol buffer definition

schomatis:
Do not merge.

Fixes https://github.com/ipfs/go-ipfs/issues/5074.

### DIFF
--- a/unixfs/pb/unixfs.proto
+++ b/unixfs/pb/unixfs.proto
@@ -1,24 +1,79 @@
 package unixfs.pb;
 
+/* FilePart represents a Unix file. */
+//
+// TODO: Rename. What name to use: `File`, `FilePart`, `FileChunk`?
+// `FilePart` makes it easier to understand the recursive nature of
+// this data structure, but the `Part` name throughout the code
+// base could confuse the reader thinking he's holding only a part
+// of the file when most of the time he actually has the root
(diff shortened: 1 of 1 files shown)


unixfs/pb/unixfs.proto
```diff
@@ -1,24 +1,79 @@
...
+// Files can be regular files, directories (what else?). That logic is being
+// disrupted with types like `Metadata` and `HAMTShard` which make me
+// rethink what is actually a file in the UnixFS specification.
+//
+// TODO: Following a question in `filesize`, the leaf and internal nodes
+// need to be differentiated, a leaf has no `blocksizes` and a internal
+// has no `Data` (is this correct?), there should even be different message
+// types for that (not happening, that will definitely break API).
 message Data {
+    // Type of file.
```
whyrusleeping: I wouldnt say this is the type of the 'file'. Its a type for this data node. A file is a more abstract concept than this type

schomatis: Thanks! This comment was an eye-opener, those types are very misleading and there is an *implicit* "more abstract concept" on top of them that should be clarified, https://github.com/ipfs/go-ipfs/issues/5157.

Stebalien:
(mostly todos)

Issue #4371, opened 2017-11-05
Forbid non-cryptographic hashes within /ipfs namespace


lgierth:
It just occured to me that by supporting non-cryptographic hashes (such as murmur3) in multihash, these are also available for CIDs, and can thus be used within the /ipfs namespace (example: ipfs/ipfs-companion#305).

This is not good! /ipfs paths MUST be cryptographically verifiable, it's their primary property. A lack of cryptographic verifiability seriously breaks so many assumptions on so many levels.

This would probably be appropriate to enforce within the respective IPLD modules.

lgierth:
cc @Stebalien @Kubuxu @whyrusleeping

lgierth:
Giving this a bug label. ipfs.io frontpage says:

> Here's how IPFS works
>
> Each file and all of the blocks within it are given a unique fingerprint called a cryptographic hash.
>
> [...]

lidel:
Also, it looks like there is a bug somewhere, probably in go-multihash.
[Spec defines 0x22 as murmur3-128](https://github.com/multiformats/multihash/blob/85080b63ad621fb4c3d125492fb1ee9b45002537/hashtable.csv#L20) while [go-multihash/multihash.go generates only 4 bytes (32bits)](https://github.com/multiformats/go-multihash/blob/710ff25621cb3e38480282cd2ecc6a1cad27ea4c/multihash.go#L140) 

See [go-multihash/multihash.go#L63](https://github.com/multiformats/go-multihash/blob/710ff25621cb3e38480282cd2ecc6a1cad27ea4c/multihash.go#L63) / [#L140](https://github.com/multiformats/go-multihash/blob/710ff25621cb3e38480282cd2ecc6a1cad27ea4c/multihash.go#L140) vs [multihash/hashtable.csv#L20](https://github.com/multiformats/multihash/blob/85080b63ad621fb4c3d125492fb1ee9b45002537/hashtable.csv#L20)

Stebalien:
Yeah... we also have to check:

1. Hash lengths.
2. For sha1, we'd want to verify that the hash isn't in the set of class of known vulnerable hashes. Note: we can't do this by *just* looking at the hash; we need the data as well.

We may want to introduce a new type/constructor to enforce this. Personally, I'd rename the current Multihash type to `InsecureMultihash` or something scary like that and introduce an `InsecureSum` method etc.

magik6k:
[...]

PR #5159, opened 2018-06-27
Ipfs v0.4.16-rc1

whyrusleeping:
License: MIT

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,60 @@
 # go-ipfs changelog
 
+## 0.4.16-rc1 2018-06-27
+- Features
+  - Embed public keys inside ipns records, use for validation ([ipfs/go-ipfs#5079](https://github.com/ipfs/go-ipfs/pull/5079))
+  - Preload git plugin by default ([ipfs/go-ipfs#4991](https://github.com/ipfs/go-ipfs/pull/4991))
+- Improvements
+  - Only resolve dnslinks once in the gateway ([ipfs/go-ipfs#4977](https://github.com/ipfs/go-ipfs/pull/4977))
+  - Libp2p transport refactor update ([ipfs/go-ipfs#4817](https://github.com/ipfs/go-ipfs/pull/4817))
+  - Improve swarm connect/disconnect commands ([ipfs/go-ipfs#5107](https://github.com/ipfs/go-ipfs/pull/5107))
+- Documentation
+  - Fix typo of sudo install command ([ipfs/go-ipfs#5001](https://github.com/ipfs/go-ipfs/pull/5001))
+  - Fix experimental features Table of Contents ([ipfs/go-ipfs#4976](https://github.com/ipfs/go-ipfs/pull/4976))
+  - Fix link to systemd init scripts in the README ([ipfs/go-ipfs#4968](https://github.com/ipfs/go-ipfs/pull/4968))
+  - Add package overview comments to coreapi ([ipfs/go-ipfs#5108](https://github.com/ipfs/go-ipfs/pull/5108))
+  - Add README to docs folder ([ipfs/go-ipfs#5095](https://github.com/ipfs/go-ipfs/pull/5095))
+  - Add system requirements to README ([ipfs/go-ipfs#5137](https://github.com/ipfs/go-ipfs/pull/5137))
+- Bugfixes
+  - Fix goroutine leak in pin verify ([ipfs/go-ipfs#5011](https://github.com/ipfs/go-ipfs/pull/5011))
+  - Fix commit string in version ([ipfs/go-ipfs#4982](https://github.com/ipfs/go-ipfs/pull/4982))
+  - Fix `key rename` command output error ([ipfs/go-ipfs#4962](https://github.com/ipfs/go-ipfs/pull/4962))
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,60 @@
...
+  - Libp2p transport refactor update ([ipfs/go-ipfs#4817](https://github.com/ipfs/go-ipfs/pull/4817))
+  - Improve swarm connect/disconnect commands ([ipfs/go-ipfs#5107](https://github.com/ipfs/go-ipfs/pull/5107))
+- Documentation
+  - Fix typo of sudo install command ([ipfs/go-ipfs#5001](https://github.com/ipfs/go-ipfs/pull/5001))
+  - Fix experimental features Table of Contents ([ipfs/go-ipfs#4976](https://github.com/ipfs/go-ipfs/pull/4976))
+  - Fix link to systemd init scripts in the README ([ipfs/go-ipfs#4968](https://github.com/ipfs/go-ipfs/pull/4968))
+  - Add package overview comments to coreapi ([ipfs/go-ipfs#5108](https://github.com/ipfs/go-ipfs/pull/5108))
+  - Add README to docs folder ([ipfs/go-ipfs#5095](https://github.com/ipfs/go-ipfs/pull/5095))
+  - Add system requirements to README ([ipfs/go-ipfs#5137](https://github.com/ipfs/go-ipfs/pull/5137))
+- Bugfixes
```
Stebalien: May want to note the DHT client fixes (https://github.com/libp2p/go-libp2p-kad-dht/pull/166):

* Don't route DHT traffic to client-only DHT nodes.
* Do reply with peer info when we're directly connected to a peer, even if they're not in our routing table.

Stebalien: Actually, are we including libp2p fixes?

whyrusleeping:
So this list is pulled from the list of merged PRs. I'm down to put more things in here if you think we should. Otherwise we can just talk about it in the notes.

kevina:
@whyrusleeping I was kind of hopping we could get the migration to strict CIDs (see https://github.com/ipfs/go-ipfs/issues/4371#issuecomment-399829328 for checklist) done for this release so the user doesn't have to migrate the repo once again in the next release.  (Most of the hard work should be done, I mainly need feedback.  The migration code needs testing, but I have been dragging me feet on that as testing is a bit hard with out the target release for the migration.)

whyrusleeping:
@kevina we can just have another release out pretty quick. We need this release to ship as soon as possible, the longer we wait, the worse the DHT gets (due to an issue in the DHT in 0.4.15)

kevina:
@whyrusleeping okay, that's your call.  Unless I am missing something. the user will just have to migrate there repo once for this release and then again for the next, something I was hoping to avoid.

whyrusleeping:
yep. Would have been great to avoid that, but shipping these fixes is more important.

CHANGELOG.md
```diff
@@ -1,5 +1,60 @@
...
+  - Fix goroutine leaks in refs.go ([ipfs/go-ipfs#5018](https://github.com/ipfs/go-ipfs/pull/5018))
+  - Fix panic, Don't handle errors with fallthrough ([ipfs/go-ipfs#5072](https://github.com/ipfs/go-ipfs/pull/5072))
+  - Fix how filestore is hooked up with caching ([ipfs/go-ipfs#5122](https://github.com/ipfs/go-ipfs/pull/5122))
+  - Add record validation to offline routing ([ipfs/go-ipfs#5116](https://github.com/ipfs/go-ipfs/pull/5116))
+- General Changes and Refactorings
+  - Remove leftover bits of dead code ([ipfs/go-ipfs#5022](https://github.com/ipfs/go-ipfs/pull/5022))
+  - Remove fuse platform build constraints ([ipfs/go-ipfs#5033](https://github.com/ipfs/go-ipfs/pull/5033))
+  - Warning when legacy NoSync setting is set ([ipfs/go-ipfs#5036](https://github.com/ipfs/go-ipfs/pull/5036))
+  - Clean up and refactor namesys module ([ipfs/go-ipfs#5007](https://github.com/ipfs/go-ipfs/pull/5007))
+  - Create a raw node instead of a file node when there is no content. ([ipfs/go-ipfs#4693](https://github.com/ipfs/go-ipfs/pull/4693))
```
kevina: This is kind of confusing.  Maybe "When raw-leaves are enabled create an empty raw leaf instead of a protobuf node when there is no content"

djdv:
I was hoping to sneak in https://github.com/ipfs/go-ipfs/pull/4956, but I'm okay with punting that to the next release if DHT is in need.

Stebalien:
@kevina so, the CID fixes will be rather wide-reaching and I have a feeling that'll take a while to review, test, bake, and deem ready. I'd like to minimize migrations but, well, I think we'll have to get used to running them pretty frequently anyways. Really, we should focus on making them easier to run.

Stebalien:
@djdv 

> I was hoping to sneak in #4956, but I'm okay with punting that to the next release if DHT is in need.

I've taken a look at that patch and it looks good to me. However, IIRC, you were thinking about erroring out by default? We should probably come to a decision on that before merging.

Also, I'd like to be smarter about file versus directory symlinks if possible.

whyrusleeping:
> Let's just ship this. We can write full release notes when we release.

A man after my own heart <3

Issue #5157, opened 2018-06-26
What does it mean to be a directory in the UnixFS layer?


schomatis:
The `Directory` type of a UnixFS object (contrary to its name) doesn't actually indicate a directory, or at least it's not the *only* possible type of a directory, the `HAMTShard` type also indicates a directory, leading to confusing comparisons like:

https://github.com/ipfs/go-ipfs/blob/ecf7d157a6d5e525b122079367f4b6c2ba25e951/mfs/dir.go#L159-L161

So `Directory` is only the "plain implementation" type of a directory, it's not *the* directory type (a very similar confusion arises with the `File` and `Raw` types). What are the requisites of a UnixFS directory? It seems that there is no interface that would define it, at this point the closest that can be find is the `Directory` from the `unixfs.io` package,

https://github.com/ipfs/go-ipfs/blob/ecf7d157a6d5e525b122079367f4b6c2ba25e951/unixfs/io/dirbuilder.go#L28-L36

but this structure emerged just to accommodate the HAMT implementation (https://github.com/ipfs/go-ipfs/pull/3042, it evolved from a previous `directoryBuilder` structure), not as a clear and documented definition of a directory.

The biggest consequence of all this (IMO) is that a new reader needs to go through the `*hamt.Shard` structure pointer to understand what's happening when content is added to a directory (with the major danger of diving in the *really* complex code of the `hamt` package).

So the main objective is to provide the user a clear explanation of what it can expect from a directory, possibly presenting a documented interface, while also using it to hide as much as possible the HAMT directory variant, to avoid confusing code lines when the user is following the execution path of how are files added to a directory in the MFS hierarchy,

https://github.com/ipfs/go-ipfs/blob/ecf7d157a6d5e525b122079367f4b6c2ba25e951/unixfs/io/dirbuilder.go#L101-L107

For example, looking for common characteristics of the plain and the HAMT (which I mostly do not understand) implementations I'm observing that its children are still referenced as DAG links (HAMT contains a map for a quick access to them but the links remain at the DAG layer); on the contrary, whereas the plain implementation does not hold information in the `Data` field the HAMT stores the bit field there.

Stebalien:
> So Directory is only the "plain implementation" type of a directory, it's not the directory type (a very similar confusion arises with the File and Raw types)

Yes. Sharded directories came later so they got a different type.

[...]

PR #5160, opened 2018-06-27
unixfs: add a directory interface

### LINKED ISSUE #5084: MFS Directory vs UnixFS/IO Directory
The `mfs` and the `unixfs.io` packages have both a `Directory` structure (the first containing the second) and both implementing an `AddChild` function (used in combination to add files/dirs to the MFS filesystem) which makes the code hard to understand when reading the MFS work flow (see https://github.com/ipfs/go-ipfs/issues/5081). It would seem that the rationale behind this is to have MFS representing an abstract directory and UnixFS providing the actual node to support that (but that is just my speculation).

The MFS `Directory` names its inner UnixFS/IO `Directory` as `dirbuilder` (without any comment) which makes it hard to understand their relationship besides what was previously mentioned,

https://github.com/ipfs/go-ipfs/blob/7853e53860805e08a212d78c4baa5d59bff99ba8/mfs/dir.go#L26-L41

https://github.com/ipfs/go-ipfs/blob/7853e53860805e08a212d78c4baa5d59bff99ba8/unixfs/io/dirbuilder.go#L28-L36

My main question right now is: what is the purpose of `unixfs.io.Directory`? If I understand correctly `unixfs.FSNode` is the UnixFS representaion for the directory, `mfs.Directory` is the same at the MFS layer, why is part of the UnixFS functionality regarding directories in a different structure (in `unixfs.io.Directory`)?

schomatis:
Addresses https://github.com/ipfs/go-ipfs/issues/5157.
Closes #5084.

----

```
unixfs: add a directory interface

Add a UnixFS `Directory` that hides implementation details and helps to
distinguish *what* is a UnixFS directory.

Replace the `unixfs.io.Directory` structure that contained the HAMT and basic
directory implementations (through inner pointers) with an interface containing
the same methods. Implement those methods in two clearly distinct structures for
each implementation (`BasicDirectory` and `HAMTDirectory`) avoiding pointer
logic and clearly differentiating which implementation does what.

The potential basic to HAMT transition was being hidden behind the `AddChild`
call at the UnixFS layer (changing one implementation pointer  with the other
one), it is now being explicitly done at the MFS layer.

Rename the `dirbuilder.go` file to `directory.go` and change the `Directory` MFS
attribute `dirbuilder` to `unixfsDir` to be consistent.
```


mfs/dir.go
```diff
@@ -363,6 +363,18 @@ func (d *Directory) AddChild(name string, nd ipld.Node) error {
 		return err
 	}
 
+	//TODO: Resolve where to manage a possible transition to sharding.
```
schomatis: This is ugly, and it's not equivalent to transitioning inside `unix.io.Directory.AddChild`.

mfs/dir.go
```diff
@@ -372,6 +372,30 @@ func (d *Directory) AddChild(name string, nd ipld.Node) error {
 	return nil
 }
 
+// AddChildUnixFSDir adds a child to the inner UnixFS
+// directory and transitions to a HAMT sharding if needed.
+// TODO: Rename. Actually the real confusion is that
+// this structure has a `AddChild` function that calls
+// another `AddChild` function of its inner dir.
+func (d *Directory) AddChildUnixFSDir(name string, nd ipld.Node) error {
```
schomatis: @whyrusleeping What do you think about moving the HAMT transition to the MFS layer?

whyrusleeping: I guess the answer to that question depends on when we will need to do this, but I think its probably fine to have it in mfs

schomatis:
[unixfs: add a directory interface](https://trello.com/c/pvXrREXR/5-unixfs-add-a-directory-interface)

schomatis:
[unixfs: add a directory interface](https://trello.com/c/pvXrREXR/5-unixfs-add-a-directory-interface)

unixfs/io/directory.go
```diff
@@ -0,0 +1,262 @@
...
+	hamt "github.com/ipfs/go-ipfs/unixfs/hamt"
+
+	ipld "gx/ipfs/QmWi2BYBL5gJ3CiAiQchg6rn1A8iBsrWy51EYxvHVjFvLb/go-ipld-format"
+	cid "gx/ipfs/QmapdYm1b22Frv3k17fqrBYTFRxwiaVJkB299Mfn33edeB/go-cid"
+)
+
+// ShardSplitThreshold specifies how large of an unsharded directory
+// the Directory code will generate. Adding entries over this value will
+// result in the node being restructured into a sharded object.
+var ShardSplitThreshold = 1000
```
whyrusleeping: I think this is actually unused now. We should probably remove it

unixfs/io/directory.go
```diff
@@ -0,0 +1,262 @@
...
+// Directory defines a UnixFS directory. It is used for creating, reading and
+// editing directories. It allows to work with different directory schemes,
+// like the basic or the HAMT implementation.
+//
+// It just allows to perform explicit edits on a single directory, working with
+// directory trees is out of its scope, they are managed by the MFS layer
+// (which is the main consumer of this interface).
+type Directory interface {
+
+	// SetPrefix sets the prefix of the root node.
```
whyrusleeping: i'd say "sets the cid prefix"

unixfs/io/directory.go
```diff
@@ -0,0 +1,262 @@
...
+		return &HAMTDirectory{
+			dserv: dserv,
+			shard: shard,
+		}, nil
+	}
+
+	return nil, ErrNotADir
+}
+
+// SetPrefix implements the `Directory` interface.
```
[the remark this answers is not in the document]
whyrusleeping: I think in go, for obvious interface implementation methods, they often just get left blank.  Saying 'implements X' is fine with me too

PR #5167, opened 2018-06-28
add: use file name instead of path name

djdv:
Followup to: https://github.com/ipfs/go-ipfs/pull/4919


Stebalien:
@whyrusleeping this is a tiny bug fix that we might as well pull in before the release.

whyrusleeping:
cool, i'll leave the window open until the weekend and then cut another RC

Issue #5066, opened 2018-06-01
MFS Root vs MFS File/Directory


schomatis:
I would like to understand (and possibly decouple) the relationship between those two.

I'm looking at the `NewRoot` function,

https://github.com/ipfs/go-ipfs/blob/7853e53860805e08a212d78c4baa5d59bff99ba8/mfs/system.go#L69-L104

The `Root` is created from a `ProtoNode` (instead of an `ipld.Node`, why?) that is usually an [empty directory](https://github.com/ipfs/go-ipfs/blob/v0.4.15/unixfs/unixfs.go#L258) and stores a [`node`](https://github.com/ipfs/go-ipfs/blob/v0.4.15/mfs/system.go#L53-L54) reference to it, which doesn't seem to be used anywhere in the code.

They all share the reference to (I'm assuming the same) `DAGService`, is that necessary?

The `Type` field of the `Root` (which one could mistakenly associate with the `Type()` function of the `FSNode` interface to which `Root` doesn't belong) also doesn't seem to be used anywhere in the code. This also make it harder to conceptually distinguish between the Root and the File/Dir types (coupled with other minor details like the `Flush()` function that is part of the `FSNode` interface that `Root` also implements, besides File and Dir, that one could associate with those types).

The switch in `NewRoot` seems to allow the possibility that the root value (`val`) can either be a directory or a file but other functions like `Mkdir` assume the root is a `Directory`,

https://github.com/ipfs/go-ipfs/blob/7853e53860805e08a212d78c4baa5d59bff99ba8/mfs/ops.go#L132

Another thing that may generate confusion is that the `Context` argument received by `NewRoot` is named `parent` and then passed to `NewFile` and `NewDirectory` functions which use the name `parent` not for the context passed (which uses the more standardized name `ctx`) but rather for the `ipld.Node` (which seems to make more sense). Could the `parent context.Context` be renamed `ctx context.Context`? What is the rationale behind that name?

Stebalien:
> The Root is created from a ProtoNode (instead of an ipld.Node, why?)

Probably for historical reasons. This is now technically incorrect as we now support Raw IPLS objects as well.

> that is usually an empty directory and stores a node reference to it, which doesn't seem to be used anywhere in the code.

We may have used that field at one point but yes, it appears unused. Feel free to remove it for now.

> They all share the reference to (I'm assuming the same) DAGService, is that necessary?

Yes? That is, they all need access to *some* DAG service.

[...]

PR #5170, opened 2018-06-29
mfs: make `Root` value a `Directory`

schomatis:
As discussed in https://github.com/ipfs/go-ipfs/issues/5066#issuecomment-395597562 and https://github.com/ipfs/go-ipfs/issues/5066#issuecomment-395706829 (different consecutive comments).

---------------------------

```
mfs: make `Root` value a `Directory`

Make `Root` value explicitly a `Directory` structure instead of the `FSNode`
interface (which also allowed the `File` type). This helps to make the code
easier to reason about: the root of an MFS layout is always a directory, not a
(single) file.

Rename `GetValue()` to `GetDirectory()` to also make it more explicit, the
renamed function now returns a `Directory` so there is no need for type
assertions that were previously done on the `FSNode` interface to check that it
was actually a `Directory`.

`NewRoot()` now doesn't allow to create `Root` structures from DAG nodes that
contain UnixFS files.
```

---------------------------

```
mfs: remove unused `Root` variables `node` and `Type`
```

---------------------------

```
mfs: remove `DAGService` from `Root`

The `Root` structure now explicitly contains a `Directory` (instead of an
`FSNode` interface), use that `Directory`'s `DAGService` instead of its own
`dserv` variable (which was used only once in `closeChild()`). The `DAGService`
in the `Root` and the `Directory` was the same (passed as an argument in the
`NewRoot` initializer function).

This leaves the `Root` structure with only a `Directory` and a `Republisher` and
allows to better rethink its role and whether if those two structures should be
grouped together (and if that group's name should be `Root`).
```

### DIFF
--- a/fuse/ipns/ipns_unix.go
+++ b/fuse/ipns/ipns_unix.go
@@ -118,14 +118,7 @@ func loadRoot(ctx context.Context, rt *keyRoot, ipfs *core.IpfsNode, name string
 
 	rt.root = root
 
-	switch val := root.GetValue().(type) {
-	case *mfs.Directory:
-		return &Directory{dir: val}, nil
-	case *mfs.File:
-		return &FileNode{fi: val}, nil
-	default:
-		return nil, errors.New("unrecognized type")
-	}
+	return &Directory{dir: root.GetDirectory()}, nil
 }
 
 type keyRoot struct {
--- a/mfs/ops.go
+++ b/mfs/ops.go
@@ -1,7 +1,6 @@
 package mfs
 
 import (
-	"errors"
 	"fmt"
 	"os"
 	gopath "path"
@@ -129,7 +128,7 @@ func Mkdir(r *Root, pth string, opts MkdirOpts) error {
 		return fmt.Errorf("cannot create directory '/': Already exists")
 	}
 
-	cur := r.GetValue().(*Directory)
+	cur := r.GetDirectory()
 	for i, d := range parts[:len(parts)-1] {
 		fsn, err := cur.Child(d)
 		if err == os.ErrNotExist && opts.Mkparents {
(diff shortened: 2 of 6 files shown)


schomatis:
Take advantage of the (now accessible) directory in root and try to remove the `DAGService`.

schomatis:
Done. Ready for review.

mfs/ops.go
```diff
@@ -172,12 +171,11 @@ func Mkdir(r *Root, pth string, opts MkdirOpts) error {
 	return nil
 }
 
+// Lookup extracts the root directory and performs a lookup under it.
+// TODO: Now that the root is always a directory, can this function
+// be collapsed with `DirLookup`? Or at least be made a method of `Root`?
```
kevina: This function is used in many places within the code.  I would be okay with making this a method of Root but would not collapse it with `DirLookup` and `DirLookup` will work on any directory and not just the root.

fuse/ipns/ipns_unix.go
```diff
@@ -118,14 +118,7 @@ func loadRoot(ctx context.Context, rt *keyRoot, ipfs *core.IpfsNode, name string
 
 	rt.root = root
 
-	switch val := root.GetValue().(type) {
```
Stebalien: So, I need to get better about submitting comments instead of editing them till my browser crashes and they disappear...

This is the one real reason to allow the root to be something other than a directory. However, really, the "root" is just a way to tie a file/directory in an MFS tree to a publisher. I'd rather just:

1. Hold a directory directly as you suggested here https://github.com/ipfs/go-ipfs/issues/5066#issuecomment-401616027.
2. Provide some way to "watch" files/directories.

Currently, we have the `childCloser` but I wonder if we want a more general system. Thoughts?

schomatis: > So, I need to get better about submitting comments instead of editing them till my browser crashes and they disappear...

I'm glad I'm not the only one, I'm seriously considering submitting a feature request to GitHub that would allow me to explicitly save my drafts..

> Thoughts?

I was considering adding that logic to the directory implementation of `childCloser`, if the directory has a parent call the parent's `closeChild` (current behavior), if not (you are the root) call the republisher and tell it things have changed. That would require every directory to be aware of the republisher, either with a pointer to it or some other way (which is not nice). Anyway, I need to keep studying how the republisher works and see if I can come up with something better.

Stebalien: I was thinking of a more general purpose notification boadcast system where users could register and unregister any number of notifiees (where the parent directory would register itself as a notifiee).

schomatis: Could you elaborate a bit more on that? I thought the directory was the notifier (I may be misunderstanding the word *notifiee*, that's the one that receives the notification, right?).

Kubuxu: Unrelated:
You can save a code comment with a "Start Review", comments will be visible only to you until you send the review.

PR #5172, opened 2018-07-01
mfs: remove unused `Flush()` from `Root`

schomatis:
Towards clarifying https://github.com/ipfs/go-ipfs/issues/5066.

-------------

```
mfs: remove unused `Flush()` from `Root`

The unused `Flush()` function in the `Root` structure helps create the misguided
impression that `Root` implements the `mfs.FSNode` interface (which it doesn't).

```

### DIFF
--- a/mfs/system.go
+++ b/mfs/system.go
@@ -113,20 +113,6 @@ func (kr *Root) GetValue() FSNode {
 	return kr.val
 }
 
-// Flush signals that an update has occurred since the last publish,
-// and updates the Root republisher.
-func (kr *Root) Flush() error {
-	nd, err := kr.GetValue().GetNode()
-	if err != nil {
-		return err
-	}
-
-	if kr.repub != nil {
-		kr.repub.Update(nd.Cid())
-	}
-	return nil
-}
-
 // FlushMemFree flushes the root directory and then uncaches all of its links.
 // This has the effect of clearing out potentially stale references and allows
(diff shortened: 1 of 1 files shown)


Stebalien:
Really, shouldn't we be calling this function on exit?

schomatis:
It seems that `Root`'s `Flush()` was replaced by `FlushMemFree()` (https://github.com/ipfs/go-ipfs/pull/4509) which uses the `Directory` `Flush()` function (but not `Root`'s).

Stebalien:
We replaced the existing *use* of `Flush` with `FlushFreeMem` because we wanted to free memory. However, that doesn't mean that `Flush` isn't otherwise useful (as `FlushFreeMem` is very unsafe).

schomatis:
> that doesn't mean that Flush isn't otherwise useful

Sure, I'm don't want to imply we shouldn't use `Flush()`, just that at this point in time this function doesn't seem to be called anywhere in the code base and generates confusion when trying to understand the role of the MFS root, that was my main motivation to remove it, but I'm not against discussing in a separate issue to include it back in the code path. Would you prefer to keep it and close this PR?

Stebalien:
> Would you prefer to keep it and close this PR?

Yes. A lot of the interfaces in go-ipfs are built with the intention of extracting them into a separate library at some point. This is definitely true for MFS.

Issue #5092, opened 2018-06-07
MFS Republisher


schomatis:
I'm opening this issue to learn more about the republishing system on the MFS (which I currently understand nothing about).

I'm preemptively tagging @Stebalien to see if he can help me with some basic pointers about its main design goals, ~~I'm assuming this system allows to recreate a different set of nodes that would represent any modifications to a file in the MFS.~~

schomatis:
I'm especially interested in understanding what is going on in this function,

https://github.com/ipfs/go-ipfs/blob/7853e53860805e08a212d78c4baa5d59bff99ba8/mfs/system.go#L257-L291

schomatis:
The core publish function is over here,

https://github.com/ipfs/go-ipfs/blob/7853e53860805e08a212d78c4baa5d59bff99ba8/core/core.go#L735-L738

It's not trivial to find and it seems very important in order to understand what is the publisher doing because there are many mentions of the *publish* verb but I'm not seeing any definition of exactly *what* does publish mean in this context.

Stebalien:
> I'm especially interested in understanding what is going on in this function,

That function is calling "pubfunc" whenever the hash at the MFS root changes. Currently, I believe we only use this to persist to disk. However, IIRC, we plan on using this to publish to IPNS (eventually).

The complexity comes from the fact that publishing an IPNS record is slow so we need to batch updates. Algorithm:

1. When we receive the first update after publishing, we set a "long" timeout.
2. When we receive any update, we reset the short timeout.
3. If either the short timeout or the long timeout elapses, we call `publish(latestHash)`.

These long timeout ensures that we delay publishing by at most `TimeoutLong`. The short timeout allows us to publish sooner if it looks like there are no more updates coming down the pipe.

---

> The core publish function is over here,

Publish is fully pluggable. That republisher records the MFS root to disk. We also have some cases where we don't publish anything (e.g., when using a temporary MFS for adding/modifying directory trees).

schomatis:
Documented in https://github.com/ipfs/go-mfs/pull/36.

PR #5173, opened 2018-07-01
[WIP][PoC] analyze MFS `Close()`

schomatis:
Do not merge.

-------
```
[WIP][PoC] coreunix: remove MFS `Close()` from `Finalize()`

I would tend to think that this is handled through `closeChild()` and
`Directory`'s `Flush()` (if responsibly called).
```
-------
```
[WIP][PoC] mfs: remove `Close()` from `Root()`

And the functions of its only consumer, IPNS.
```
-------

The first commit passes the tests (Jenkins doesn't count), the second one doesn't, which make sense: IPNS doesn't use the `closeChild()` and related functions (I'm imagining), modifications are not signaled to MFS as the `ipfs files` commands do (MFS files are just being modified by the filesystem) so it does need an explicit `Close()` call.

If this reasoning is correct we can eliminate the `Close()` call from the `coreunix` package (first commit) and document the function as being used only by IPNS, this will largely clarify its relationship with `closeChild()`.

To do: address this in https://github.com/ipfs/go-ipfs/issues/5092.

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -586,10 +586,6 @@ func (n *IpfsNode) teardown() error {
 	// needs to use another during its shutdown/cleanup process, it should be
 	// closed before that other object
 
-	if n.FilesRoot != nil {
-		closers = append(closers, n.FilesRoot)
-	}
-
 	if n.Exchange != nil {
 		closers = append(closers, n.Exchange)
 	}
--- a/core/coreunix/add.go
+++ b/core/coreunix/add.go
@@ -241,11 +241,6 @@ func (adder *Adder) Finalize() (ipld.Node, error) {
 		return nil, err
 	}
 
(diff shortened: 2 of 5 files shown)


schomatis:
This is strange, the second commit is actually passing the online tests, but in my machine `TestMultipleDirs` fails. Are the FUSE tests disabled here?

schomatis:
Yes, they [are](https://github.com/ipfs/go-ipfs/blob/7e8f6c96047e1a94c1c892aa50e1c75e01b7fda2/.travis.yml#L16-L17).

eingenito:
@schomatis is it safe to close this PR? It sounds like maybe it was for CI testing?

schomatis:
Sure, thanks for the remainder!

PR #5182, opened 2018-07-03
mfs: add `Inode` interface

### LINKED ISSUE #5094: MFS File vs Directory
At this point I'm pretty sure the MFS `Root` structure can be refactored out (https://github.com/ipfs/go-ipfs/issues/5066) so I would like to concentrate on abstracting the common characteristics of the `Directory` and `File` structures, which are many: `parent`, `name`, `dserv`, `lock`/`nodelk`, maybe `node` (in the directory it is hidden behind the `dirbuilder` which could maybe also be refactored out https://github.com/ipfs/go-ipfs/issues/5084). What those two structures should reflect is only what makes a file different from a directory in an MFS context: a file has content that can be modified and a directory has (what could also be denominated as content) entries (links) of other files and directories.

https://github.com/ipfs/go-ipfs/blob/7853e53860805e08a212d78c4baa5d59bff99ba8/mfs/dir.go#L26-L41

https://github.com/ipfs/go-ipfs/blob/fc05376e94125b854e2b837760c44aacbaa97e2d/mfs/file.go#L16-L28

What I would like to discuss is how to encapsulate that common functionality, could we adopt the Unix terminology of a file that can be a directory or a regular file? (which I never really liked as the *file* term ends up carrying too much conflicting meaning, but can't think of anything else)

schomatis:
As suggested in https://github.com/ipfs/go-ipfs/issues/5094#issuecomment-395710376.

--------------

```
mfs: add `Inode` interface

Abstract common characteristics of the MFS `File` and `Directory` structures
inside a separate interface called `Ìnode`.
```

-----------

Closes #5094.

### DIFF
--- a/mfs/inode.go
+++ b/mfs/inode.go
@@ -1,0 +1,54 @@
+package mfs
+
+import (
+	ipld "gx/ipfs/QmWi2BYBL5gJ3CiAiQchg6rn1A8iBsrWy51EYxvHVjFvLb/go-ipld-format"
+)
+
+// Inode abstracts the common characteristics of the MFS `File`
+// and `Directory`. All of its attributes are initialized at
+// creation and can't be modified.
+type Inode interface {
+
+	// Name of this `Inode` in the MFS path (the same value
+	// is also stored as the name of the DAG link).
+	Name() string
+
+	// Parent directory of this `Inode` (which may be the `Root`).
+	Parent() childCloser
+
+	// DagService used to store modifications made to the contents
+	// of the file or directory the `Inode` belongs to.
+	DagService() ipld.DAGService
+}
+
+// inode implements the `Inode` interface.
+type inode struct {
+	name       string
+	parent     childCloser
+	dagService ipld.DAGService
+}
+
+// NewInode creates a new `inode` structure and returns its pointer
+// as the `Inode` interface.
+func NewInode(name string, parent childCloser, dagService ipld.DAGService) Inode {
+	return &inode{
+		name:       name,
+		parent:     parent,
+		dagService: dagService,
+	}
+}
+
+// Name implements the `Inode` interface.
+func (i *inode) Name() string {
+	return i.name
+}
+
+// Parent implements the `Inode` interface.
+func (i *inode) Parent() childCloser {
+	return i.parent
+}
+
+// DagService implements the `Inode` interface.
+func (i *inode) DagService() ipld.DAGService {
+	return i.dagService
+}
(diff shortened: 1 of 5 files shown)


mfs/inode.go
```diff
@@ -0,0 +1,46 @@
...
+func NewInode(name string, parent childCloser, dagService ipld.DAGService) Inode {
+	return Inode{
+		name:       name,
+		parent:     parent,
+		dagService: dagService,
+	}
+}
+
+// GetName returns the `Inode`'s name.
+func (i *Inode) GetName() string {
```
magik6k: Why not access these directly? Would be cleaner imo (and look less than java)

schomatis: > Why not access these directly?

To make it explicit that it is a read-only attribute and should not be modified.

> (and look less than java)

Would you prefer just using `Name()`?

https://golang.org/doc/effective_go.html#Getters

schomatis:
> To make it explicit that it is a read-only attribute and should not be modified.

This is actually not happening right now in this PR, `parent` is being directly accessed and I didn't even notice it. I need to rethink this, adding `Inode` as a named field to protect its members would live the ugly `d.inode.parent().closeChild()`.

https://stackoverflow.com/a/32535022

schomatis:
As mentioned in the Stack Overflow answer, the customary Go solution seems to be to define an interface (`Inode`) and a separate structure to implement it.

mfs/inode.go
```diff
@@ -0,0 +1,54 @@
...
+	// Parent directory of this `Inode` (which may be the `Root`).
+	Parent() childCloser
+
+	// DagService used to store modifications made to the contents
+	// of the file or directory the `Inode` belongs to.
+	DagService() ipld.DAGService
+}
+
+// inode implements the `Inode` interface.
+type inode struct {
```
schomatis: Is this the appropriate name for the structure that implements the `Inode` interface?

whyrusleeping: Yeah, thats what I would do.

whyrusleeping:
Pretty clean refactor, stevens right though, it probably makes sense to just embed a struct here. (unless you want to be able to mock out the inode stuff for some reason in testing)

schomatis:
> (unless you want to be able to mock out the inode stuff for some reason in testing)

Not really, my main motivation for the interface was to be able to enforce read-only attributes (more as a code clarification feature than to protect inadvertent modifications to the object), but that may not be the most Go idiomatic approach, I'll embed the structure then.

schomatis:
@overbool Want to take over this one?

This should be ported to the `go-mfs` repo now and the interface (as suggested above) should be turned to a structure to simplify the code.

overbool:
@schomatis Yes. I'm also interested in this.

schomatis:
Great! Ping me if you have any trouble porting the diff.

PR #5184, opened 2018-07-03
Allow files to be wrapped with a mutator during add

whyrusleeping:
I'm playing around with different ways of importing files in a side project, and this small change allows me to encrypt or compress files on the fly. 

License: MIT
Signed-off-by: Whyrusleeping <why@ipfs.com>

### DIFF
--- a/core/coreunix/add.go
+++ b/core/coreunix/add.go
@@ -71,26 +71,27 @@ func NewAdder(ctx context.Context, p pin.Pinner, bs bstore.GCBlockstore, ds ipld
 
 // Adder holds the switches passed to the `add` command.
 type Adder struct {
-	ctx        context.Context
-	pinning    pin.Pinner
-	blockstore bstore.GCBlockstore
-	dagService ipld.DAGService
-	Out        chan interface{}
-	Progress   bool
-	Hidden     bool
-	Pin        bool
-	Trickle    bool
-	RawLeaves  bool
-	Silent     bool
-	Wrap       bool
-	NoCopy     bool
-	Chunker    string
-	root       ipld.Node
-	mroot      *mfs.Root
-	unlocker   bstore.Unlocker
-	tempRoot   *cid.Cid
-	Prefix     *cid.Prefix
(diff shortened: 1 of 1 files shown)


core/coreunix/add.go
```diff
@@ -71,26 +71,27 @@ func NewAdder(ctx context.Context, p pin.Pinner, bs bstore.GCBlockstore, ds ipld
...
+	Progress    bool
+	Hidden      bool
+	Pin         bool
+	Trickle     bool
+	RawLeaves   bool
+	Silent      bool
+	Wrap        bool
+	NoCopy      bool
+	Chunker     string
+	FileWrapper func(io.Reader) (io.Reader, error)
```
schomatis: Why are all the field listed as modified? The `FileWrapper` is the only new field right? Also, just to be extra annoying, could we add a comment explaining the function the newly added field? I'm a strong believer in the broken windows theory :)

core/coreunix/add.go
```diff
@@ -114,6 +115,14 @@ func (adder *Adder) SetMfsRoot(r *mfs.Root) {
 
 // Constructs a node from reader's data, and adds it. Doesn't pin.
 func (adder *Adder) add(reader io.Reader) (ipld.Node, error) {
+	if adder.FileWrapper != nil {
+		wr, err := adder.FileWrapper(reader)
```
schomatis: So, this is the single point of entry of a reader to the chunker right? (The other similar calls I'm finding in the code are for testing purposes only.)

core/coreunix/add.go
Stebalien: That's gofmt. It aligns everything.

magik6k:
After the files refactor (https://github.com/ipfs/go-ipfs/pull/5661) lands, it should be fairly easy to just wrap file interfaces to do this without having to modify the adder (and it will work via coreapi without touching internals)

Stebalien:
Closing in favor of a `go-ipfs-files` adapter. We can reopen this if that doesn't work.

PR #5185, opened 2018-07-04
gateway: Mutable Filesystem accessablity

TUSF:
URL paths beginning with /~/ will grant access to the mfs from the gateway.
The prefix can be changed to /mfs/ if that is more desirable.

License: MIT
Signed-off-by: Jeremia Dominguez <ragef33@gmail.com>


Stebalien:
So, we were actually thinking of using `/local` for this. However, exposing this from the *gateway* is something we have to think about and definitely put behind a config option (that defaults to off).

We're going to have a meeting next week to discuss the next steps for MFS so I'll be sure to bring this up.

TUSF:
Alright. I changed the path to `/local`, and added the option to enable mfs through the gateway using `--enable-mfs-gateway` when running the daemon. It occurred to me that you probably meant having it be enabled through `ipfs config` or something, but I'm not familiar enough with the source yet to fiddle with that.

PR #5187, opened 2018-07-04
Add support for datastore plugins

whyrusleeping:
This will make it a little bit easier to write your own datastores for ipfs (at least on platforms that go supports plugins >.>).

The next step is making it easier to initialize ipfs with a given datastore config

### DIFF
--- a/plugin/datastore.go
+++ b/plugin/datastore.go
@@ -1,0 +1,14 @@
+package plugin
+
+import (
+	"github.com/ipfs/go-ipfs/repo/fsrepo"
+)
+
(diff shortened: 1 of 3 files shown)


Stebalien:
@magik6k have you tested this and does it work? If so, we can go ahead and merge it.

magik6k:
Yup, with https://github.com/ipfs/go-ipfs/pull/5516, both offline and with daemon.

magik6k:
I'd just test this with an actual datastore plugin, like we do with the git plugin

PR #5190, opened 2018-07-05
dagreader: remove `Offset()` method

schomatis:
Remove `Offset()` from the `DagReader` interface. It's not part of the Unix API
and it wasn't used anywhere except for the tests (a helper function was added to
replace it).


unixfs/mod/dagmodifier_test.go
```diff
@@ -750,3 +750,11 @@ func BenchmarkDagmodWrite(b *testing.B) {
 		}
 	}
 }
+
+func getOffset(reader uio.DagReader) int64 {
```
schomatis: I'm rewriting `getOffset` from `dagreader_test.go`, exporting it doesn't seem to work, is there a cleaner way (without adding it back to `dagreader.go`)?

PR #4977, opened 2018-04-25
only resolve dnslinks once in the gateway

Stebalien:
If the domain has a DNS-Link, we want to use it even if it points to, e.g., an IPNS address that doesn't resolve.

fixes #4973

PR #5199, opened 2018-07-07
gateway: fix further resolution of dnslink failures

lgierth:
A while ago, we noticed that if IpnsHostname failed to look up
a dnslink for a hostname in a Host header, it wouldn't respond
with an error, but instead just skip the Host header part and
carry on.

In the issue were this was flagged, the example was:
`_dnslink.tr.wikipedia-on-ipfs.org => /ipns/QmSomeKey`
This key had stopped to resolve, and thus the resolution of
/ipns/tr.wikipedia-on-ipfs.org would fail. Now you'd expect
an error response, but instead it'd happily continue processing
the request, disregarding the Host header, and allowing requests
to e.g. http://tr.wikipedia-on-ipfs.org/ipns/libp2p.io to succeed.

A previous patch (ipfs/go-ipfs#4977) tried to work around this
issue by limiting the recursion depth of Host: header dnslink
resolutions to 1.

In this patch we start to correctly handle the initial resolution
error and stop processing the request, and we remove the earlier
workaround.

License: MIT


lgierth:
This isn't correct yet -- we'll want to still carry on with the request if "successfully" get a NXDOMAIN response from DNS.

PR #5202, opened 2018-07-07
correctly handle multi-hop dnslink resolution

Stebalien:
Namesys returns `ErrResolveRecursion` when it stops recursing due to a depth limit. It doesn't return success.

Alternative to #5199.


Stebalien:
@lgierth this is an alternative fix that's basically just my original change bug done correctly.

PR #5205, opened 2018-07-09
Refactor code

spartucus:
Keep the code style consistent. since all package alias are lowercase words, no need Camel Peak Name.


spartucus:
"License" and "Signed-off-by"  are added to commit message, have no idea why commit-message-check still fails.

whyrusleeping:
Thanks @spartucus :)

PR #5206, opened 2018-07-09
allow multiple concurrent pin fetches

jbenet:
This commit
- removes the locking around fetching graphs in recursive pins.
- splits the function for readability and safer locking/unlocking
- documents important pinning logic details

This should improve performance significantly for users
pinning multiple large graphs concurrently.

ipfs-cluster should be able to use this directly without having to
implement it at a higher level.

pair programmed with @whyrusleeping

cc @hsanjuan @stebalien

### DIFF
--- a/pin/pin.go
+++ b/pin/pin.go
@@ -210,9 +210,7 @@ func NewPinner(dstore ds.Datastore, serv, internal ipld.DAGService) Pinner {
 }
 
 // Pin the given node, optionally recursive
-func (p *pinner) Pin(ctx context.Context, node ipld.Node, recurse bool) error {
-	p.lock.Lock()
-	defer p.lock.Unlock()
+func (p *pinner) Pin(ctx context.Context, node ipld.Node, recursive bool) error {
 	err := p.dserv.Add(ctx, node)
 	if err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


jbenet:
Link is wrong on gitcop: https://github.com/ipfs/community/blob/master/contribution-guidelines.md#a-license-and-a-signed-off-by-trailers-are-required

jbenet:
gitcop link should be https://github.com/ipfs/community/blob/master/CONTRIBUTING-2.md#a-license-and-a-signed-off-by-trailers-are-required

jbenet:
apparently i should try compiling or something

jbenet:
Since this is important code, and not very much of it, would be great to get multiple reviewers verifying it is indeed correct.

jbenet:
nevermind this is wrong: an interleaved GC _after_ the fetch but _before_ the `pinner.recursePin.Add(c)` may remove data and leave unsound assumptions about the pinset.

currently a pin implies the data is local. pins may become declarative in the future (signal intent but don't necessarily mean the data is all there yet. this would work closet to how ipfs-cluster works now, and would allow ipfs nodes to finish pinning. 

(further note: the commandline could get a `--wait-for-data[=true]` flag that tells whether or not to wait until all the data has been fetched, or only until the pin is persisted. should default to true because users' code today expect `ipfs pin -r <path>` returning 0 to mean the data is now local).

Issue #5192, opened 2018-07-05
Analyzing the DAG reader


schomatis:
It is comprised of the group of structures in the `unixfs.io` package capable of doing a DFS graph traversal of a file DAG. It reads (and performs similar operations) in the same sequence the DAG was built, so the `importer` is a good reference (once #5118 gets merged) to better understand it. I leave some notes here on what could be improved to make it more accessible to the new reader.

The recursive approach is hidden behind the `PBDagReader` (parent) and its `ReadSeekCloser` (child) attribute. The first doesn't call the second, instead it interacts with the interface which the second implements. The interface, although it provides a nice generalization, it obscures the very simple algorithm to traverse a file DAG that only has internal `ProtoNode`s and leaf `RawNode`s (represented at the UnixFS layer as `File` and `Raw` types). This could be made more explicit to leverage the concepts that the reader could already have of what a file DAG is (e.g., if it's already familiarized with the `importer`). Could an iterative algorithm help clarify how the reader operates? Instead of the implicit call stack with the `PBDagReader`s at different depths of the tree (making up the path to the current leaf being read) we could have an explicit stack with the information of each edge of the path at the different depths (this stack would pretty much be the reader itself with a few other attributes of control).

Node promises. Many nodes are being requested at the same time (preloaded) to streamline the graph traversal, this mechanism is abstracted through the `ipld.NodePromise` structure and it should be made very clear to the reader what a promise is, *why* is it used (see https://github.com/ipfs/go-ipfs/pull/5162#issuecomment-402213069). The name of the `promises` attribute of the `PBDagReader` should be changed to reflect that those are actually the child nodes (that will be visited in the future), to focus on *what* they represent  instead of *how* those nodes are fetched (through promises).

The direct manipulation of the protobuf `unixfs.pb.Data` structure in `PBDagReader.pbdata` should be replaced with the more refined `unixfs.FSNode` structure (which contains `unixfs.pb.Data`).

Remove unused `node` attribute (`*mdag.ProtoNode`) of the `PBDagReader`.

As said before the role of the buffer (`buf` attribute) may be misleading, it should be clear that this is the child node one level below being accessed by another `PBDagReader` (or `BufDagReader` in case we've reached the leaf). Using the `ReadSeekCloser` interface instead of `DagReader` (which contains it) is correct but may be confusing if not clarified well enough. (Revisit this point.)

[...]

PR #5216, opened 2018-07-11
unixfs: fix `dagTruncate` to preserve node type

### LINKED ISSUE #4519: Truncating an MFS file 175 times makes it unwritable.

schomatis:
So, in a weird series of coincidences that I don't fully understand I need to fix this old bug (which I already fixed in the PR #4758 that I never finished) to be able to refactor some parts of the code of the DAG reader (see https://github.com/ipfs/go-ipfs/issues/5192#issuecomment-404216727).

I'm selfishly assigning this PR a high priority to get it merged as soon as possible after the upcoming release, as it will simplify my work on the DAG reader.

Fixes #4519.

-----------

```
unixfs: fix `dagTruncate` to preserve node type

Extract the original `FSNode` passed inside the `ipld.Node` argument and modify
its `Blocksizes` (removing all of them and re-adding the ones that were not
truncated). In contrast, the replaced code was creating a new `FSNode` that was
not preserving some of the features of the original one.
```

-----------

```
mfs: add test case for MFS repeated truncation failure
```

### DIFF
--- a/mfs/mfs_test.go
+++ b/mfs/mfs_test.go
@@ -1136,3 +1136,43 @@ func TestTruncateAtSize(t *testing.T) {
 	}
 	fd.Truncate(4)
 }
+
+func TestTruncateAndWrite(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	ds, rt := setupRoot(ctx, t)
+
+	dir := rt.GetDirectory()
+
+	nd := dag.NodeWithData(ft.FilePBData(nil, 0))
+	fi, err := NewFile("test", nd, dir, ds)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	fd, err := fi.Open(OpenReadWrite, true)
+	defer fd.Close()
+	if err != nil {
+		t.Fatal(err)
+	}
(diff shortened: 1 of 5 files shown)


schomatis:
@whyrusleeping WDYT? (This is my most blocking issue at the moment.)

whyrusleeping:
On the `TRUNC_HASH`, I think its fine to change the hash here as this is clearly an erroneous object. We should probably also test that writing to a file, then truncating it back to the previous length results in the same object (unless there are reasons why this shouldnt happen)

schomatis:
> I think its fine to change the hash here as this is clearly an erroneous object.

Great, do you know how was `TRUNC_HASH` generated?



> We should probably also test that writing to a file, then truncating it back to the previous length results in the same object

Yes, will add that.

schomatis:
Fixed `TRUNC_HASH` values and rebased to the merged PR https://github.com/ipfs/go-ipfs/pull/5253 fix so `TestDagTruncateSameSize` won't panic now.

schomatis:
Tests are passing now (minus Jenkins..) so @whyrusleeping I think this is RFM now.

Issue #5088, opened 2018-06-06
ipfs files cp: report which file doesn't exist


schomatis:
In the case of copying a file in MFS I would like to know which of the two files doesn't exist, as the Unix `cp` command would report it (this also applies to moving files and similar commands).
```
cp /this-path-doesnt-exist /this-one-either
# cp: cannot stat '/this-path-doesnt-exist': No such file or directory

ipfs files cp /this-path-doesnt-exist /this-one-either
# Error: file does not exist
# (Which one?)
```

From the code I would think the most direct way is just to wrap the errors that return from `getNodeFromPath` and `PutNode` (that will fail in the [`Directory.Find()`](https://github.com/ipfs/go-ipfs/blob/7853e53860805e08a212d78c4baa5d59bff99ba8/unixfs/io/dirbuilder.go#L167-L172) call) with the name of either the `src` or `dst` path, but maybe the error should be handled deeper inside the code.

https://github.com/ipfs/go-ipfs/blob/7853e53860805e08a212d78c4baa5d59bff99ba8/core/commands/files.go#L330-L340

(Ideally I would even like to specify which part of the path doesn't exist, but this is probably going to far.)

@Stebalien WDYT

kevina:
Our error reporting in general sucks.

Stebalien:
So, `cp` actually does this at the *top* level. If we were mimicking it, we'd do it right before calling SetError in the command.

The most general solution would be to have `DirLookup` and `PutNode` wrap the error in an `os.PathError`. However, that won't cover the `/ipfs/` special case... (I wish we handled this within MFS instead of within the command)

So yeah, your solution sounds good.

PR #5218, opened 2018-07-12
commands/files: add error messages for `cp` paths

schomatis:
Fixes https://github.com/ipfs/go-ipfs/issues/5088.


schomatis:
Why does the `wafflebot` always add the `in progress` label even if I don't do anything after pressing `Create pull request`?

whyrusleeping:
@schomatis because in waffle, the default board position for 'new PRs' is 'in progress'. Its kindof annoying

PR #5224, opened 2018-07-13
[WIP] ci: get gx from dists

### LINKED ISSUE #5222: [jenkins] CI build should fetch gx from dists, not go get
We currently use go get to fetch and install gx. We should just be using the scripts we have in the repo for installing gx binaries.

magik6k:
Fixes #5222

### DIFF
--- a/ci/Jenkinsfile
+++ b/ci/Jenkinsfile
@@ -136,7 +136,8 @@ ansiColor('xterm') { withEnv(['TERM=xterm-color']) {
       windows: {
         setupStep('windows') { run ->
           timeout(time: gotest_timeout, unit: 'MINUTES') {
-            run 'go get -v github.com/jstemmer/go-junit-report github.com/whyrusleeping/gx github.com/whyrusleeping/gx-go'
+            run 'go get -v github.com/jstemmer/go-junit-report'
+            run 'make gx-deps'
             run "gx install --global"
 
             try {


whyrusleeping:
Hrm... jenkins is unhappy

magik6k:
yep, `make gx-deps` is broken on windows, and needs to be fixed for this to work

Stebalien:
@djdv I've tried to fix this by trial and error but have absolutely no idea what's going on here.

Stebalien:
@djdv if you get a chance, mind taking a look? This is a bit hard for the rest of us to debug (without access to a windows machine).

djdv:
@Stebalien 
My fault, this one slipped by me. I'll take a look.

djdv:
This works for me locally, so my initial guess is that this is an issue with the CI env. I'm not sure which interpreter it's using and when, or what it's environment looks like at that moment.
The CI says it's invoking a batch file which would be interprited by `cmd`, but `make` typically invokes `sh` itself. So you get a process tree that typically looks like `cmd>make>sh>go build ...`

For some reason "bin/dist_get" is being turned into simply "bin", and the interpreter is trying to invoke that.

Stebalien:
It looks like we're using http://gnuwin32.sourceforge.net/packages/make.htm in CI. It looks like that uses cmd, not bash.

We could either:

1. Depend on mingw make, use that in jenkins as well, and assume bash.
2. Fix this.

I'm actually fine with 1. I think most windows users are likely to be using mingw for development anyways.

Thoughts?

djdv:
>Depend on mingw make, use that in jenkins as well, and assume bash.

I'm okay with this, we already recommend it in the [build docs](https://github.com/ipfs/go-ipfs/blob/master/docs/windows.md).

While having a closer to native build system is nice imo, there's really no reason for it if we're mainly invoking posix shell scripts with posix tools anyway.

Fixing this would require a good amount of pain.
The options I can think of are:
- Insert more conditions in the make files (not the best)
- Rewrite the shell utilities to be platform native, like Powershell (big effort for little gain, double the maintenance, requires considerations for tools like `tar`, `curl`, et al. as well)
- Rewrite them in something slightly more agnostic, something in the same spirit as [Batsh](https://github.com/BYVoid/Batsh) (not a recommendation, I've never used it).

It may be worth revisiting the build system eventually, but not yet.

djdv:
I replicated the CI environment and it appears that the version of `make` we're using does try to invoke `sh` but `sh` is not in the environment PATH. This is the cause of the first error
>The system cannot find the path specified.

once `sh` is in the PATH we get a simillar error on `make gx-deps` because `rm` is not in the PATH.

Appending to PATH via `SET PATH=%PATH%;%ProgramFiles%\Git\bin;%ProgramFiles%\Git\usr\bin`
Fixes this problem in the CI environment and successfully executes.

This is not likely a proper fix though, really we need to add the tools we depend on explicitly. In this case some shell interpreter (dash,bash) added as `sh` to the PATH, and standard tools like `rm`.
They're included with some of our dependencies but are not added to the PATH, and probably shouldn't be since package `git` may not necessarily imply pulling in `sh`, etc.

I also noticed that there may be some redundancy in installing MS build tools. It looks like we do it via Chocolatey and NPM.

I'm going to see if I can remove any POSIX tools from Chocolatey and instead rely on MSYS2 to manage them.
So we'd pull in Windows tools via Chocolatey (likely including MSYS2 itself) and POSIX ones via MSYS2.
Alternatively I may opt to keep them separate.

Pinging @VictorBjelkholm this may be interesting to you.

djdv:
>It looks like that uses cmd, not bash.

Supplementary note, I believe this is only the fallback for when `sh` is faulty/nonexistent.

Stebalien:
(gx is taking a nap)

Issue #5070, opened 2018-06-03
Repo lock temporarily unavailable on Jenkins


Stebalien:
I'm getting the following error when starting nodes in CI (Jenkins):

> Error: cannot acquire lock: Lock FcntlFlock of /.../repo.lock failed: resource temporarily unavailable

This is probably some kind of resource exhaustion thing but we may need to add a retry loop.

Stebalien:
Alternatively, it could be that the kernel is slow at cleaning up the old IPFS process so, while we *think* the node has been shut down, the process still hasn't been cleaned up so it's still holding the lock. That's actually more likely.

Stebalien:
@magik6k this is the issue we're seeing ^^. The second issue is that `iptb` doesn't report an error when connecting offline nodes. Also, `iptb start` should be failing...

gammazero:
Not seeing this issue anymore. It is either fixed, or is specific to Jenkins which we are not using.

PR #5237, opened 2018-07-16
 unixfs: refactor `precalcNextBuf`

schomatis:
Addresses part of #5192.

--------------------

```
unixfs: refactor switch in `precalcNextBuf`

Do not use `NewDagReader` just for the `RawNode` case.
Treat invalid UnixFS types in the same case.
```

---------------------

```
unixfs: split `precalcNextBuf`

Create new `loadBufNode` function to handle the `buf` logic which is unrelated
to the main `precalcNextBuf` logic of processing promises to fetch nodes.
```

### DIFF
--- a/unixfs/io/dagreader.go
+++ b/unixfs/io/dagreader.go
@@ -3,7 +3,6 @@ package io
 import (
 	"context"
 	"errors"
-	"fmt"
 	"io"
 
 	mdag "github.com/ipfs/go-ipfs/merkledag"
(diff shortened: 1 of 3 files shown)


schomatis:
Rebased, but there are some weird Travis errors I should look into.

Stebalien:
Those errors look like https://github.com/ipfs/go-ipfs/issues/5070

schomatis:
Yes, I'm used to Jenkins failing (I don't really pay attention to it anymore), but I've never seen Travis fail like that (I mean passing only one of the `sharness` tests and one of the Go tests from different architectures). I'm leaving the link here to document it and I'm making push to trigger another test.

https://travis-ci.org/ipfs/go-ipfs/builds/404523422?utm_source=github_status&utm_medium=notification

schomatis:
And the Linux error is on me :)

Stebalien:
Yeah, I get the travis error all the time. We really need to fix that test.

whyrusleeping:
@schomatis youre on fire, I think i've merged more PRs from you in the past few days than i've merged for myself all year o.o

PR #5239, opened 2018-07-16
unixfs: fix `FSNode` accessor in `NewDirectoryFromNode`

schomatis:
Recently merged PR #5189 conflicted with #5160.


schomatis:
One PR was renaming the `GetType()` accessor that the other one was using, but GitHub didn't complain, because the file was being renamed? I would like to understand what went wrong here to avoid making the same mistake in the future

whyrusleeping:
@schomatis I think its because github doesnt rerun CI on merge, which it kinda probably should. Not a big deal, we caught it quick

PR #5243, opened 2018-07-16
DO NOT MERGE update go-ipfs and deps to go-cid 0.7.22

phritz:


kevina:
Please note that go-cid will need to be updated again once the plan for forbidding non-crypo. hashes in Cids is completed (see https://github.com/ipfs/go-ipfs/issues/4371#issuecomment-399829328 for outline of steps)

Stebalien:
@kevina unfortunately, @phritz needs this now. I've already propagated the update to all of go-ipfs's immediate dependencies so I'll fix up this PR.

Stebalien:
@phritz I have an alternative at #5246 that updates everything.

phritz:
@Stebalien cool. Though note that #5246 does not move go-ipld-cbor onto the changes in the refmt branch, which is required to resolve the issue. (There are other ways to resolve the issue, but they are less palatable, so moving ipfs to use refmt would be best if it is possible -- though I cannot say whether it is or is not.)

PR #5257, opened 2018-07-18
[WIP] unixfs: decouple the DAG traversal logic from the DAG reader

schomatis:
Do not merge.

TODO:
* Support node promises in the `Traversal` structure (providing the same functionality as [`precalcNextBuf`](https://github.com/ipfs/go-ipfs/blob/255e15827a9af7eafb432c0829e6da553f69b2e6/unixfs/io/pbdagreader.go#L77)).
* Finish commit documentation.
* Solve issue #5312 which is causing the test failures.
* Possibly move the `dagutils/traversal.go` file to [go-ipld-format](https://github.com/ipfs/go-ipld-format) (https://github.com/ipfs/go-ipfs/issues/5304).
* Rename the `Traversal` structure to `Walker` (that includes comments, commit message, filename).

Closes https://github.com/ipfs/go-ipfs/issues/5192.

--------------------------

```
unixfs: decouple the DAG traversal logic from the DAG reader

Decouple the DAG traversal logic from the `PBDagReader` encapsulating it in the
(new) `Traversal` structure.

Methods like `Read` and `Seek` now have the remaining non-traversal logic inside
the `VisitHandler` (from the `Traversal` structure) where the file DAG logic
dictates how to interpret each node visited in the `Traversal`'s `Iterate` and
`Search` algorithms.

```


whyrusleeping:
just wanted to say that this is really cool

schomatis:
@Mr0grog Could you take a look at this please? It's not finished, but I would like to get your opinion on how understandable is this design, any kind of feedback will be very useful. You don't need to proof read or make any corrections to the documentation at this point (I will bother you with that later), it's just to get your take on how this can fit the mental model described in the documentation of the balanced builder.

From your previous knowledge of the `balanced` package, take a look at this [commit][COMMIT], especially the new [`Traversal`][TRAVERSAL] structure and how it's leveraged in the [`Read`][READ] and [`Seek`][SEEK] methods of the DAG reader (comparing it with the previous [implementation][IMPL]), ignore any `TODO`s in the code.

[COMMIT]: https://github.com/ipfs/go-ipfs/pull/5257/commits/f65807ece9fff57694cd8884c726ffae68bd7e3e
[TRAVERSAL]: https://github.com/ipfs/go-ipfs/pull/5257/commits/f65807ece9fff57694cd8884c726ffae68bd7e3e#diff-5cd1ffeea88ea65055bdd5a0acc6eb55
[READ]: https://github.com/ipfs/go-ipfs/pull/5257/commits/f65807ece9fff57694cd8884c726ffae68bd7e3e#diff-0046ca7d126fec814fde6c7ffc49ab0dR133
[SEEK]: https://github.com/ipfs/go-ipfs/pull/5257/commits/f65807ece9fff57694cd8884c726ffae68bd7e3e#diff-0046ca7d126fec814fde6c7ffc49ab0dR297
[IMPL]: https://github.com/ipfs/go-ipfs/blob/255e15827a9af7eafb432c0829e6da553f69b2e6/unixfs/io/pbdagreader.go#L248

Mr0grog:
@schomatis I don’t know if I’ll be able to get to this before Wednesday or Thursday. Is that ok?

dagutils/traversal.go
```diff
@@ -0,0 +1,384 @@
...
+// what it the current nodes but which nodes (children) have already
+// been visited (assuming always a left to right, zero to `len(Links())`,
+// visiting order).
+//
+// TODO: Add a logic similar to `precalcNextBuf` to request several child
+// nodes in advanced as node promises.
+//
+// TODO: Revisit the name. `Traverser`? (is "traverser" the correct noun?),
+// `Iterator` (this would put to much emphasis on iterate whereas other
+// traversal operations like search are supported), `Topology`?
```
warpfork: `Walk` and `WalkFunc` are the terms chosen in some comparable standard library types, fwiw: https://golang.org/pkg/path/filepath/#Walk

dagutils/traversal.go
```diff
@@ -0,0 +1,384 @@
+package dagutils
+
+import (
+	"context"
+	"errors"
+
+	ipld "gx/ipfs/QmZtNq8dArGfnpCZfx2pUNY7UcjGhVp5qqwQ4hH6mpTMRQ/go-ipld-format"
+)
+
+// Traversal provides methods to move through a DAG of IPLD nodes
```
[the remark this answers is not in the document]
schomatis: Yes, rereading https://github.com/ipfs/go-ipfs/issues/5304#issue-345503022 I see that why's mentioning to actually move these type of packages there instead of `go-merkledag`.

schomatis:
After initial review this PR will need to be split in two: `traversal.go` to `go-ipld-format` and the rest of the DAG reader to `go-unixfs`.

dagutils/traversal.go
Mr0grog: FWIW, “traverse/traversal/traverser” is a technically correct (and not uncommon) English term for this that I have seen in many libraries (it even has a [Wikipedia page](https://en.wikipedia.org/wiki/Tree_traversal)). That said, I think I’ve seen “walk/walking/walker” used just as often and it’s shorter :)

schomatis:
@Mr0grog Just a reminder ping (but again, no hurry).

Mr0grog:
Ah! Sorry, was unclear on the status after the comment about splitting it up. Looking now.

dagutils/traversal.go
```diff
@@ -0,0 +1,507 @@
...
+// operations like reading and seeking inside of a UnixFS file DAG. In
+// contrast with a recursive solution, the implicit path in the call
+// stack is made explicit through the `path` variable.
+//
+// It exposes simple traversal operations like `Iterate` and  `Search`
+// that are implemented through a series of simple move methods (`down`,
+// `up`, `Right`) that modify its `path`. The `down` move method is the
+// analogous of the recursive call and the one in charge of visiting
+// the node (through the `VisitHandler`) and performing some user-defined
+// logic.
```
Mr0grog: `Right()` being the only one of these that is public feels a bit odd to me, but I get why it’s here. See my summary comment for some alternative thoughts.

dagutils/traversal.go
```diff
@@ -0,0 +1,507 @@
...
+//
+// TODO: Encapsulate the position (`path`, `level`, `childIndex`)? It's
+// a big part of the structure, and it should only be one of it, so it
+// would just create a proxy for every move call, e.g., `dt.pos.down()`.
+//
+// TODO: Consider adding methods that would retrieve an attribute indexed
+// by the current `level` to avoid including much indexing like
+// `dt.path[dt.level]` in the code. Maybe another more strong refactoring
+// that would allow to think of the current node in the `path` without
+// concerning at what `level` it's in.
```
Mr0grog: Yeah, there were definitely some points in the code where all the indexing got a bit muddy.

dagutils/traversal.go
```diff
@@ -0,0 +1,507 @@
...
+
+	// This slice has the index of the child each parent in `path` is pointing
+	// to. Its current valid length is determined by `level`. The index in the
+	// parent can be set past all of its child nodes (having a value equal to
+	// `len(Links())`) to signal it has visited (or skipped) all of them. A
+	// leaf node with no children and its index in zero would also comply with
+	// this format.
+	childIndex []uint
+	// TODO: Revisit name, `childPosition`? Do not use the *link* term, put
+	// emphasis in the child (the *what*, not the *how* we get it).
```
Mr0grog: I think this name made plenty of sense for me. The only thing clearer that comes to mind is `currentChildIndex`.

dagutils/traversal.go
```diff
@@ -0,0 +1,507 @@
...
+	//
+	// A level of -1 signals the start case of a new `Traversal` that hasn't
+	// moved yet (or its position has been reset, see `ResetPosition`).
+	// Although this state is an invalid index to the slices, it allows to
+	// centralize all the visit calls in the `down` move (starting at zero
+	// would require a special visit case inside every traversal operation like
+	// `Iterate()` and `Search`). This value should never be returned to after
+	// the first `down` movement, moving up from the root should always return
+	// `ErrUpOnRoot`.
+	level int
```
Mr0grog: Why not call this `depth`? It seems like that would be much more straightforward, especially since your comments define it in terms of “depth” ;)

dagutils/traversal.go
```diff
@@ -0,0 +1,507 @@
...
+	// not when going back up). It is the main API to implement functionality
+	// (like read and seek) on top of the `Traversal` structure.
+	//
+	// Its argument is the current `node` being visited. Any error it returns
+	// (apart from the internal `errDownStopIteration`) will be forwarded to
+	// the caller of the traversal operation (stopping it).
+	//
+	// Any of the exported methods of this API should be allowed to be called
+	// from within this method.
+	VisitHandler func(node ipld.Node) error
```
Mr0grog: I might call this a `Visitor`. In fact, I think that’s generally the common term I’ve seen for it.

dagutils/traversal.go
```diff
@@ -0,0 +1,507 @@
...
+		// of `extendPath`, it can't be zero).
+	}
+}
+
+// ErrDownNoChild signals there are no more child nodes left to visit,
+// the current child index is past the end of this node's links.
+var ErrDownNoChild = errors.New("can't go down, no child available")
+
+// ErrUpOnRoot signals the end of the DAG after returning to the root.
+var ErrUpOnRoot = errors.New("can't go up, already on root")
```
Mr0grog: This error is a pretty important public signal that calling functions will see, but I feel like `UpOnRoot` doesn’t quite explain what this is really about (that you have traversed through the entire graph). Maybe `EndOfDag` or `EndOfGraph` or `TraversalComplete`?

dagutils/traversal.go
```diff
@@ -0,0 +1,507 @@
...
+
+// ErrRightNoChild signals the end of this parent child nodes.
+var ErrRightNoChild = errors.New("can't move right, no more child nodes in this parent")
+
+// errDownStopIteration signals the stop of the traversal operation.
+var errDownStopIteration = errors.New("stop")
```
Mr0grog: I wonder if `Pause` would be better than `Stop` here — in my mind, stopping means you can’t continue from where you left off. With this code, though, you can (and that’s important, because that’s how seeking works).

unixfs/io/dagreader.go
```diff
@@ -77,3 +85,332 @@ func NewDagReader(ctx context.Context, n ipld.Node, serv ipld.NodeGetter) (DagRe
 		return nil, ErrUnkownNodeType
 	}
 }
+
+// NewDagReaderWithSize constructs a new `dagReader` with the file `size`
+// specified.
+func NewDagReaderWithSize(ctx context.Context, n ipld.Node, serv ipld.NodeGetter, size uint64) (DagReader, error) {
```
Mr0grog: It was a little unclear to me at first that this was the recalculated size of the whole file/DAG — I thought it might be the maximum amount of data to read.

schomatis:
@Mr0grog Another great review as usual, thanks. I'll be bothering you with some follow-up questions in the coming days.

Mr0grog:
I’ll bet this will be really useful for simplifying the traversal logic in the new depth-limited `ipfs refs -r` work in #5337, too.

schomatis:
@Mr0grog Regarding the `Iterate`/`Search` unification, indeed there's a strong similarity between them, the first contains the second: `Search` just goes down while `Iterate` when it reaches a leaf node it goes back up again to go down in another direction.

You can have `Iterate` and just issue a stop when you reached the first leaf node, and even though that wouldn't really be a big burden to pass to the user, my main motivation for using two different functions is that conceptually I see both as clearly different operations which the user can relate to both of them. (This is subjective of course, I would like to hear how you see them.)

Even if internally they are implemented with very similar (almost the same code, actually) my main question would be: what is the advantage, from the user's perspective, to unify them and present them only as `Iterate`? Do you see the difference between both operations as a potential source of confusion to the user? If I unify them and the user issues a stop in the first leaf node it visits during an iteration, what would be the conceptual model he would be think of (different from a *search* that reached the end of the road)?

dagutils/traversal.go
```diff
@@ -0,0 +1,507 @@
...
+// NewTraversal creates a new `Traversal` structure from a `root`
+// IPLD node.
+func NewTraversal(ctx context.Context, root ipld.Node, serv ipld.NodeGetter) *Traversal {
+	return &Traversal{
+		path:       []ipld.Node{root},
+		childIndex: []uint{0},
+		level:      -1,
+		// Starting position, "on top" of the root node, see `level`.
+
+		ctx:  ctx,
```
whyrusleeping: should avoid storing the context in the struct, it's generally considered a bad practice.

dagutils/traversal.go
```diff
@@ -0,0 +1,507 @@
...
+	// TODO: Use a function?
+
+	// Attribute needed to fetch nodes, will become really useful once node
+	// promises are implemented.
+	ctx  context.Context
+	serv ipld.NodeGetter
+
+	// The CID of each child of each node in the `path` (indexed by
+	// `level` and `childIndex`).
+	childCIDs [][]*cid.Cid
```
whyrusleeping: that's gonna be a lot of memory, as is the promise set below

dagutils/traversal.go
schomatis: This functionality was inherited from the `PBDagReader` which stored the context to later be able to 
 call `ipld.GetNodes`. How would you advise this be handled?

dagutils/traversal.go
schomatis: This is also inherited from the `PBDagReader` (I'm glad this refactoring exposed some of the already existing characteristics of the current implementation) which has the

```Go
	// NodePromises for each of 'nodes' child links
	promises []*ipld.NodePromise

	// the cid of each child of the current node
	links []*cid.Cid
```

slices, each of those is instantiated in every level of the DAG when the recursive algorithm creates a new `PBDagReader`.

> lot of memory

Have in mind that `level` relates to the depth of the DAG (I should change its name as suggested above), so the size of this matrix grows at an `O(log n)` rate with the size of the DAG.

How would you advise this be handled?

dagutils/traversal.go
schomatis: Yes, you're right. I avoided the term "depth" at first because I considered it an absolute property, "the depth of the entire DAG", and didn't want to give the sense that this property was being modified here. I could use the name `currentDepth` or `positionDepth`, or yes, just `depth` alone and indicate this is related more to the current position in the DAG than the DAG itself.

dagutils/traversal.go
schomatis: > but I feel like UpOnRoot doesn’t quite explain what this is really about (that you have traversed through the entire graph).

That is actually the case only for the `Iterate()` context, in future traversal operations that error may be interpreted differently, but you're right that `Iterate()` should wrap this error (which seems more like an internal error) and return something more meaningful like `EndOfDag` or `IterateComplete`.

dagutils/traversal.go
schomatis: Yes, agree. Maybe some other public method like `SkipNode` should be exported (that would internally call an un-exported `right`), but I feel that the graphic "turn right" movement is something that the user could grasp rather easily and it could relate more to that than a `Skip` function. Not really sure though, I usually lean towards abstracting but this felt like a natural "take look under the hood, you see? there's not much complex logic down there".

unixfs/io/dagreader.go
schomatis: Yes, this was an ugly patch I had to introduce to preserve the `Size()` API, I'll try to document it more clearly.

dagutils/traversal.go
hsanjuan: If we want to re-use this across the go-ipfs codebase the Visit function should:

* Receive the `level` of the visited node
* Receive the parent(s) of the visited node

We probably want to provide a couple default Visit functions to wrap some commonly used logic:

* One implementing full recursive traversal of the dag, pruning already visited branches
* One implementing depth limited traversal, pruning already limited branches (if they are not going to be explored deeper).

schomatis:
@hsanjuan Thanks for the feedback. I'm going to leave the async feature for a second iteration though (I'll open an issue once I get this merged in `go-ipld-format`) to avoid adding more complexity to the code so I can focus now on being sure this actually works for the DAG reader before using it elsewhere.

schomatis:
> There are way more than two kinds of operations you might want to do here :)

Sorry, I think I didn't express myself correctly before, I'm definitely *not* saying that iterate/search cover every use case (the did cover the necessity of the DAG reader that was my test case used to iterate this idea), my motivation to have them both is to avoid passing that burden to the user, that is, instead of forcing *every* user-defined `Visit` that wants to do a search to include the check "if on leaf node stop", this structure should provide that functionality (according to the DRY principle).

The same logic applies for future operations that may be needed: if many parts of the code need (for example) an iterator that goes up and down twice this structure should provide a different operation, e.g., `iterateTwice`, instead of having many parts of the code base repeating the same logic. Of course you can lay all that functionality on top of `Iterate` but we should be evaluating how much of that logic should this structure absorb (in the form of traversal operations added to the API) and how much should be left to the user (e.g., if the user needs a iterator that goes 3 1/2 times around the DAG, then yes, that's on him, the API shouldn't be modified).

> I do think there’s still plenty of room for more specialized functions that basically just wrap Iterate, though.

Exactly, but if those function do something conceptually different than iterate I would want to analyze if I need to be providing a different traversal operation here (that was the motivation to add `Search` in the first place).

I'm interpreting @hsanjuan comment  in this same sense (but I may be missing something),

> We probably want to provide a couple default Visit functions to wrap some commonly used logic:

to provide different operations (like `Search`) to avoid asking the user to do this himself, like the depth check in your code:
>```
> err := traversal.Iterate(func(node ipld.Node) error {
>     // Do whatever work needs to be done on this node
> 
>     if (traversal.level >= maxDepth) {
>         return dagutils.SkipNext(len(node.Links()))
>     }
>     return nil
> })
>```

> Side note: are you planning to shift towards “walk” terminology?

Yes, I'll switch it once I move this to the other repos.

schomatis:
Thanks @Mr0grog and everyone for all the feedback!! (I like the `Seek` proposal.)

I'll be closing this now and start moving it to the corresponding repos, incrementally incorporating these changes and bothering each feedbacker (not sure if that word exists :) to see if the new code correctly reflects your input.

schomatis:
I don't think Why is going to have time to review his comments ([1](https://github.com/ipfs/go-ipfs/pull/5257#pullrequestreview-147422155), [2](https://github.com/ipfs/go-ipfs/pull/5257#pullrequestreview-147422199)) so I'm bothering you @Stebalien just to check if you'd like to have any modifications there.

Mr0grog:
In the mean time, maybe worthwhile to mark those with `// TODO:` or `// FIXME:` or `// PERF:` comments about investigating was to reduce memory usage? That way, there are clear places for someone who wants to do optimization work to jump in.

schomatis:
That's a good idea but I would like to postpone it at the moment, the only place to do some actual optimizations is in the logic around node promises which at the moment I'm not sure if I have implemented them correctly so I don't want to waste anybody's time optimizing something that may be changed in the near future (note to self: bother @Stebalien to take a quick look at https://github.com/ipfs/go-ipfs/pull/5257/commits/694c9a2262191db150d9afe893b108223e9579d1 to see if it makes sense).

hannahhoward:
I think this is super cool and I agree that replicating dag traversal logic in a million places is awkward and it'd be nice to have a generalized version.

My comments generally match the previous comments:

1. This is not ideal for super high speed/parallism, but that can be left for a future iteration (though it's a little tricky cause then you effectively need to walk in multiple directions at once)

2. I agree that Search is a bit nebulous, because the description of the function doesn't clarify that it effectively leaves it to the Visitor to walk right to the appropriate node (I am inferring that but it took a close read of the code)

All things being equal though, pending @Stebalien 's comments I think this is good to merge for now.

Issue #5255, opened 2018-07-18
DAG reader: seeking to the end of the file leaves the offset at zero


schomatis:
Seeking to an offset equal to the size of the file (or a bigger)  leaves the `linkPosition` (unmodified) at the start of the file (where subsequent read operations will pick up from).

https://github.com/ipfs/go-ipfs/blob/9bad5fe6fef11fc5d923e7ebecbe6f53deb044fe/unixfs/io/pbdagreader.go#L275-L286

I'm working on some modifications to the DAG reader and I'm not sure if this is a bug or it is an intended behavior (I don't know what happens in the Unix API when doing something like this) as some of the current tests (like [`TestLargeWriteChunks`](https://github.com/ipfs/go-ipfs/blob/9bad5fe6fef11fc5d923e7ebecbe6f53deb044fe/unixfs/mod/dagmodifier_test.go#L292)) depend on it.

Stebalien:
This should follow POSIX and seek past the end of the file. Reading should return zero bytes and an EOF.

schomatis:
Ok, this is a bug then, thanks.

schomatis:
Fixed in https://github.com/ipfs/go-unixfs/pull/60.

PR #5262, opened 2018-07-19
mfs: seek to 0 before reading in `TestTruncateAndWrite`

schomatis:
Right now this test isn't failing due to another bug that causes seeking to the end to return to the beginning of the file (https://github.com/ipfs/go-ipfs/issues/5255) and the read call works, but it would be nice to get this test fixed as it's failing in a refactoring of the DAG reader (https://github.com/ipfs/go-ipfs/pull/5257).

### DIFF
--- a/mfs/mfs_test.go
+++ b/mfs/mfs_test.go
@@ -1167,12 +1167,18 @@ func TestTruncateAndWrite(t *testing.T) {
 		if l != len("test") {
 			t.Fatal("incorrect write length")
 		}
+
+		_, err = fd.Seek(0, io.SeekStart)
+		if err != nil {
+			t.Fatal(err)
+		}
+
 		data, err := ioutil.ReadAll(fd)
 		if err != nil {
 			t.Fatal(err)
 		}
 		if string(data) != "test" {
-			t.Errorf("read error at read %d, read: %v", i, data)
+			t.Fatalf("read error at read %d, read: %v", i, data)
 		}
 	}
 }


mfs/mfs_test.go
```diff
@@ -1167,12 +1167,18 @@ func TestTruncateAndWrite(t *testing.T) {
...
+			t.Fatal(err)
+		}
+
 		data, err := ioutil.ReadAll(fd)
 		if err != nil {
 			t.Fatal(err)
 		}
 		if string(data) != "test" {
-			t.Errorf("read error at read %d, read: %v", i, data)
+			t.Fatal(fmt.Errorf("read error at read %d, read: %v", i, data))
```
Stebalien: Can just use `Fatalf`.

whyrusleeping:
@schomatis so ftruncate in standard unix doesnt touch the current offset of the file descriptor, right?

Stebalien:
No but the write does. However, yeah, this should probably require a seek after the truncate as well. That looks like a different issue.

schomatis:
> Can just use Fatalf.
> 
Fixed, thanks for the tip!

schomatis:
> @schomatis so ftruncate in standard unix doesnt touch the current offset of the file descriptor, right?

Yes, actually the `Truncate` function isn't modifying the offset, but there seems to be another (different) bug in how `writeStart` is handled that is compensating for the lack of a seek after truncation.

The `writeStart` offset never syncs  its position with `curWrOff` and keeps growing without limit and `Sync()` (incorrectly) interprets it as "at EOF, need to append", so the write happens (as an append to an empty file DAG, which is the same as a write at offset zero) and the test succeeds. I'll submit a separate issue for this.

schomatis:
Some sort of republish error at CircleCI but the other tests are passing, RFM.

whyrusleeping:
Thanks @schomatis!

PR #5265, opened 2018-07-20
pin update fixes

Stebalien:
* flush pins
* correctly print output
* add a sharness test

### DIFF
--- a/core/commands/pin.go
+++ b/core/commands/pin.go
@@ -411,6 +411,12 @@ new pin and removing the old one.
 			return
 		}
 
+		err = n.Pinning.Flush()
+		if err != nil {
+			res.SetError(err, cmdkit.ErrNormal)
+			return
+		}
+
 		res.SetOutput(&PinOutput{Pins: []string{from.String(), to.String()}})
 	},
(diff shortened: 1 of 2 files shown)


schomatis:
@Stebalien What issue/PR/doc would you recommend to learn about the pinning system architecture to better understand the `Flush` call?

Stebalien:
Don't know. Looking at the code, it looks like pins are cached in-memory until the user calls `Flush`. At that point, they're packed into a single protonode and stored in the datastore. `corerepo.Pin` and `corerepo.Unpin` automatically flush (and have a bunch of other interesting logic) but `pinner.{Pin,Unpin,Update}` don't.

schomatis:
There's some interesting logic down there..

schomatis:
> corerepo.Pin and corerepo.Unpin automatically flush (and have a bunch of other interesting logic) but pinner.{Pin,Unpin,Update} don't.

Sounds reasonable.

I can confirm this fixes some weird `unwrapOutput` error.

LGTM, minus the rebase, probably because I took to long to review this, sorry.

whyrusleeping:
will merge in 0.4.18

PR #5267, opened 2018-07-20
Release 0.4.17

Stebalien:
This is a relatively small release to fix a performance regression I introduced.
Hopefully, it'll also get us into a fast release cadence.

The downside is that freezing due to an RC slows everything to a grind.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,65 @@
 # go-ipfs changelog
 
-## 0.4.16
+## 0.4.17-rc1 2018-07-20
+
+Ipfs 0.4.17 is a quick release to fix two important bugs:
+
+1. A major performance regression in bitswap (mostly affecting go-ipfs ->
+   js-ipfs transfers).
+2. A bug resolving files in sharded directories with the gateway.
+
+However, while motivated by those fixes, it contains a few other goodies that
+will excite some users.
+
+The headline feature in this release is [urlstore][] support. Urlstore is a
+generalization of the filestore backend that can fetch file blocks from remote
+URLs on-demand instead of storing them in the local datastore.
+
+Additionally, we've added support for extracting inline blocks from CIDs (blocks
+inlined into CIDs using the identity hash function). However, go-ipfs won't yet
+*create* such CIDs so you're unlikely to see any in the wild.
+
+[urlstore]: https://github.com/ipfs/go-ipfs/blob/master/docs/experimental-features.md#ipfs-urlstore
+
+Features:
+
+* URLStore ([ipfs/go-ipfs#4896](https://github.com/ipfs/go-ipfs/pull/4896))
+* Add trickle-dag support to the urlstore ([ipfs/go-ipfs#5245](https://github.com/ipfs/go-ipfs/pull/5245)).
+* Allow specifying how the data field in the `object get` is encoded ([ipfs/go-ipfs#5139](https://github.com/ipfs/go-ipfs/pull/5139))
+* Add a `-U` flag to `files ls` to disable sorting ([ipfs/go-ipfs#5219](https://github.com/ipfs/go-ipfs/pull/5219))
+* Add an efficient `--size-only` flag to the `repo stat` ([ipfs/go-ipfs#5010](https://github.com/ipfs/go-ipfs/pull/5010))
+* Inline blocks in CIDs ([ipfs/go-ipfs#5117](https://github.com/ipfs/go-ipfs/pull/5117))
+
+Changes/Fixes:
+
+* Make `ipfs files ls -l` correctly report the hash and size of files ([ipfs/go-ipfs#5045](https://github.com/ipfs/go-ipfs/pull/5045))
+* Fix sorting of `files ls` ([ipfs/go-ipfs#5219](https://github.com/ipfs/go-ipfs/pull/5219))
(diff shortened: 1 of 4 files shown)


Stebalien:
It may be worth it to merge https://github.com/ipfs/go-ipfs/pull/5117 first to get another feature in.

Stebalien:
We should probably change this to an RC but this PR can still be reviewed as-is.

Stebalien:
Also needs the versions updated.

magik6k:
Probably want to fix https://github.com/ipfs/go-ipfs/issues/5270 first

Stebalien:
Definitely. And we'll also want to cut an RC.

CHANGELOG.md
```diff
@@ -1,6 +1,53 @@
 # go-ipfs changelog
 
-## 0.4.16
+## 0.4.17  2018-07-20
+
+Ipfs 0.4.17 is a quick release to fix a major performance regression in bitswap.
```
whyrusleeping: Worth noting that the regression only affects go to javascript transfers

CHANGELOG.md
```diff
@@ -1,6 +1,53 @@
...
+However, while motivated by that fix, it contains a few other goodies that will
+excite some users.
+
+The headline feature in this release is [urlstore][] support. Urlstore is a
+generalization of the filestore backend that can fetch file blocks from remote
+URLs on-demand instead of storing them in the local datastore.
+
+Users may also notice reduced CPU usage on IPFS nodes participating in an IPFS
+cluster.
```
whyrusleeping: " due to .... "

CHANGELOG.md
```diff
@@ -1,6 +1,53 @@
...
+
+* Make `ipfs files ls -l` correctly report the hash and size of files ([ipfs/go-ipfs#5045](https://github.com/ipfs/go-ipfs/pull/5045))
+* Fix sorting of `files ls` ([ipfs/go-ipfs#5219](https://github.com/ipfs/go-ipfs/pull/5219))
+* Make the "repo stat --size-only" command efficient ([ipfs/go-ipfs#5010](https://github.com/ipfs/go-ipfs/pull/5010))
+* Improve prefetching in `ipfs cat` and related commands ([ipfs/go-ipfs#5162](https://github.com/ipfs/go-ipfs/pull/5162))
+* Better error message when `ipfs cp` fails ([ipfs/go-ipfs#5218](https://github.com/ipfs/go-ipfs/pull/5218))
+
+Documentation:
+
+* improve documentation about the bloomfilter config options ([ipfs/go-ipfs#4924](https://github.com/ipfs/go-ipfs/pull/4924))
```
whyrusleeping: capitalize 'improve'

CHANGELOG.md
```diff
@@ -1,6 +1,53 @@
 # go-ipfs changelog
 
-## 0.4.16
+## 0.4.17  2018-07-20
```
whyrusleeping: should be rc1

kevina:
Note that by merging  #5117 in without a migration any existing id hashes in the data store will now be impossible to remove by normal means.  I don't think this is an issue but it is worth nothing.  If it is enough of a concern I can create a migration just to remove empty id hashes, but I was hoping to combine that with the strict CID check.

CHANGELOG.md
Stebalien: Affects all, just mostly JS.

Stebalien:
> Note that by merging #5117 in without a migration any existing id hashes in the data store will now be impossible to remove by normal means. I don't think this is an issue but it is worth nothing. If it is enough of a concern I can create a migration just to remove empty id hashes, but I was hoping to combine that with the strict CID check.

Can't we add a migration to do this later, technically?

kevina:
> Can't we add a migration to do this later, technically?

Yes.  But for a brief period there will be hashes in the data store will now be impossible to remove by normal means until the migration is done.  Again on't think this is an issue but wanted to mention it.

Stebalien:
Fair enough, I'm just saying that the user shouldn't even notice.

kevina:
> Fair enough, I'm just saying that the user shouldn't even notice.

Not unless they already have a large number of id hashes in the datastore, which is highly unlikely.

CHANGELOG.md
hsanjuan: the fast repo/stat? yes it's used. But I think it reduces disk IO, not cpu..

Stebalien: Ah. It reduces both, IIRC. However, in retrospect, that's probably more of a cluster release note.

Stebalien:
@whyrusleeping needs tag.

whyrusleeping:
@Stebalien tagged it!

CHANGELOG.md
```diff
@@ -1,6 +1,65 @@
 # go-ipfs changelog
 
-## 0.4.16
+## 0.4.17-rc1 2018-07-20
+
+Ipfs 0.4.17 is a quick release to fix two important bugs:
+
+1. A major performance regression in bitswap (mostly affecting go-ipfs ->
+   js-ipfs transfers).
+2. A bug resolving files in sharded directories with the gateway.
```
magik6k: TBH that issue wasn't present in 0.4.16 - was introduced in https://github.com/ipfs/go-ipfs/pull/4672 which got merged after that release

Stebalien: Hm. Good point. We should fix that before the actual release.

Issue #4485, opened 2017-12-13
Audit Unsafe


Stebalien:
Given #4483, we need to audit all usages of unsafe. We generally avoid it but some of the libraries we use use it.

# Libraries

## go-os-rename

**SAFE**

However, it appears to not handle long paths, unlike the built-in one. Furthermore, it looks like go now replaces existing files on windows (i.e., the reason we use this library no longer applies). We should stop using it.

## go-sockaddr

**SAFE?**

Mostly exported from go-ipfs. However, I've asked the reporter in #4483 to try disabling it to see if that fixes it.

## go-peerstream

**SAFE**

Used to compare pointers to avoid a deadlock.

## websocket

**SAFE**

Used to do a fast XOR. Looks safe.

## bbloom

**UNSAFE?**

Used all over the place and go vet complains that it's used incorrectly. I don't believe we use bbloom in any critical places so we can probably switch to bloom (made by the same author without unsafe).

## gogo-protobuf

**UNKNOWN**

Unclear but probably well vetted. I'm in the process of upgrading this library anyways (so if there are any missing bug fixes, that should pull them in).

## fuse

**LOL**

## go-net

**SAFE?**

Made by the go authors. Regardless we should update this.

## fsnotify.v1

**SAFE**

Looks fine.

## sys

**SAFE?**

We should try to upgrade to the latest version everywhere anyways.

## Badger

**UNKNOWN**

Their mmap code looks fine. I'm mostly worried about the skip-list implementation. However, I doubt the user in #4483 has badger enabled.

## go-codec

**UNKNOWN**

## murmur3

**UNKNOWN**

However, we don't use this so it's definitely not causing any issues.

On the other hand, it looks like it may be doing unaligned loads (problem on arm?).

## go-colorable

**SAFE**

Used on windows for syscalls. Definitely safe.

## go-logging

**SAFE?**

The memory backend uses unsafe. However, it looks like it's just using it to atomically swap pointers. From what I can tell, this all looks safe.

## go-crypto

**SAFE?**

Probably safe. Might as well update.

## leveldb

**UNKNOWN**

Uh.... Yeah. This one's going to be hard to audit. We should just update it and hope.

## go-text

**UNLIKELY**

Written by the go authors. However, we should update it.

## pb (progress bar)

**SAFE**

Only uses unsafe for a few simple ioctls. Unlikely to be an issue.

## GoEndian

**SAFE**

Checked.

## hang-fds

**WHO CARES**

[...]

Issue #4132, opened 2017-08-09
Concurrency-related crashes from Prometheus client


lgierth:
#### Description:

Since I deployed v0.4.11-pre to the gateways, I see regular crashes (~1/h) out of the Prometheus client lib (client_golang). Note that these crashes might have occured earlier too, just less frequently since the gateways way running on a lot less potent machines until yesterday.

These are supposedly fixed in client_golang's master, but updating it is non-trivial since it has seen structural refactorings and is now multiple packages that all need to be gxed.

```
fatal error: concurrent map iteration and map write

goroutine 140025520 [running]:
runtime.throw(0x561c52fef158, 0x26)
  /usr/lib/go/src/runtime/panic.go:596 +0x97 fp=0xc43e745e20 sp=0xc43e745e00
runtime.mapiternext(0xc43e745f58)
  /usr/lib/go/src/runtime/hashmap.go:737 +0x7f0 fp=0xc43e745ed0 sp=0xc43e745e20
runtime.mapiterinit(0x561c536bb180, 0xc431686840, 0xc43e745f58)
  /usr/lib/go/src/runtime/hashmap.go:727 +0x2b5 fp=0xc43e745f28 sp=0xc43e745ed0
gx/ipfs/QmX3QZ5jHEPidwUrymXV1iSCSUhdGxj15sm2gP4jKMef7B/client_golang/prometheus.computeApproximateRequestSize(0xc44a077a00, 0xc4256773e0, 0x13)
  /go/src/gx/ipfs/QmX3QZ5jHEPidwUrymXV1iSCSUhdGxj15sm2gP4jKMef7B/client_golang/prometheus/http.go:303 +0x8d fp=0xc43e745fc8 sp=0xc43e745f28
runtime.goexit()
  /usr/lib/go/src/runtime/asm_amd64.s:2197 +0x1 fp=0xc43e745fd0 sp=0xc43e745fc8
created by gx/ipfs/QmX3QZ5jHEPidwUrymXV1iSCSUhdGxj15sm2gP4jKMef7B/client_golang/prometheus.InstrumentHandlerFuncWithOpts.func1
  /go/src/gx/ipfs/QmX3QZ5jHEPidwUrymXV1iSCSUhdGxj15sm2gP4jKMef7B/client_golang/prometheus/http.go:275 +0x14f
```

lgierth:
[...]

PR #5268, opened 2018-07-21
gx update deps

Stebalien:
Updates:

* go-net
* go-text
* dns
* prometheus
* protobuf

Dedups:

* protobuf

Unblocks:

* https://github.com/libp2p/go-libp2p/pull/285

Part of:

* https://github.com/ipfs/go-ipfs/issues/4485


Stebalien:
I had to play some games to get prometheus working (especially with our version of protobuf). It would be nice to test this on a gateway before merging.

Stebalien:
@lgierth this patch updates prometheus. Can we test it on a gateway?

Also, is https://github.com/ipfs/go-ipfs/issues/4132 still valid?

Stebalien:
Alternative to #5269. We can also do both but this patch *also* deduplicates the protobuf packages.

whyrusleeping:
yolo

Issue #5270, opened 2018-07-21
Resolving links in sharded directories does not seem to work


VictorBjelkholm:
#### Version information:

Gateways running commit 1dede1b1d7046b4a2246b7ec29820655548ad324

#### Type:

Bug
#### Description:

I've tried three projects that are using sharded directories and one without. The one without resolves links correctly but none of the projects using sharded directories works for resolving links. However, it does work when using 0.4.15

- https://localhost:7002/ipns/arch.victor.earth/community works correctly (using version 0.4.15)
- https://ipfs.io/ipns/arch.victor.earth/community gets error message `ipfs resolve -r /ipns/arch.victor.earth/community: no link by that name`

magik6k:
Likely introduced in https://github.com/ipfs/go-ipfs/pull/4672 which switched from using `path.pesolver.ResolvePath` to `path.pesolver.ResolveToLastNode` which doesn't appear to respect `Resolver.ResolveOnce`. As the gateway uses coreapi it got affected by this

VictorBjelkholm:
@magik6k something is still weird here...

https://ipfs.io/ipfs/QmXeh5S5Dk528ueFX3iFRB5TuKUNrWM9dEiLfC4crVwuHD is a sharded directory.

https://ipfs.io/ipfs/QmXeh5S5Dk528ueFX3iFRB5TuKUNrWM9dEiLfC4crVwuHD/PAL_Express.html gives me `no link by that name`

If I look on the IPLD explorer and take that path: https://explore.ipld.io/#/explore/QmXeh5S5Dk528ueFX3iFRB5TuKUNrWM9dEiLfC4crVwuHD it would be `27PAL_Express.html`

https://ipfs.io/ipfs/QmXeh5S5Dk528ueFX3iFRB5TuKUNrWM9dEiLfC4crVwuHD/27PAL_Express.html gives me `inconsistent ResolveOnce / nd.Resolve`

VictorBjelkholm:
Also:

```
$ ipfs resolve /ipfs/QmXeh5S5Dk528ueFX3iFRB5TuKUNrWM9dEiLfC4crVwuHD/PAL_Express.html
Error: no link named "PAL_Express.html" under QmXeh5S5Dk528ueFX3iFRB5TuKUNrWM9dEiLfC4crVwuHD
```

magik6k:
Looking at it

magik6k:
[...]

PR #5271, opened 2018-07-21
Fix resolving links in sharded directories on gateway

magik6k:
Fixes https://github.com/ipfs/go-ipfs/issues/5270

### DIFF
--- a/path/resolver/resolver.go
+++ b/path/resolver/resolver.go
@@ -70,6 +70,10 @@ func (r *Resolver) ResolveToLastNode(ctx context.Context, fpath path.Path) (ipld
 
 	for len(p) > 0 {
 		lnk, rest, err := r.ResolveOnce(ctx, r.DAG, nd, p)
+
+		// Note: have to drop the error here as `ResolveOnce` doesn't handle 'leaf'
+		// paths (so e.g. for `echo '{"foo":123}' | ipfs dag put` we wouldn't be
+		// able to resolve `zdpu[...]/foo`)
 		if lnk == nil {
 			break
 		}
--- a/path/resolver/resolver.go
+++ b/path/resolver/resolver.go
@@ -70,11 +70,11 @@ func (r *Resolver) ResolveToLastNode(ctx context.Context, fpath path.Path) (ipld
 
 	for len(p) > 0 {
 		lnk, rest, err := r.ResolveOnce(ctx, r.DAG, nd, p)
-		if err != nil {
-			return nil, nil, err
-		}
-
 		if lnk != nil {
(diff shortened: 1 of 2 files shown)


plugin/loader/load_linux.go
```diff
@@ -1,3 +1,5 @@
+// +build !noplugin
```
Stebalien: Looks like a `git add .`.

magik6k:
should be ready

path/resolver/resolver.go
```diff
@@ -69,19 +69,28 @@ func (r *Resolver) ResolveToLastNode(ctx context.Context, fpath path.Path) (ipld
...
+			next, err := lnk.GetNode(ctx, r.DAG)
 			if err != nil {
 				return nil, nil, err
 			}
 			nd = next
 			p = rest
+			continue
+		}
+
+		val, rest, err := nd.Resolve(p)
```
whyrusleeping: is 'rest' not used?

path/resolver/resolver.go
```diff
@@ -69,19 +69,28 @@ func (r *Resolver) ResolveToLastNode(ctx context.Context, fpath path.Path) (ipld
 	}
 
 	for len(p) > 0 {
-		val, rest, err := nd.Resolve(p)
-		if err != nil {
-			return nil, nil, err
-		}
+		lnk, rest, err := r.ResolveOnce(ctx, r.DAG, nd, p)
+		if lnk != nil {
```
whyrusleeping: maybe we could clean this loop up a little bit by pulling everything below this if statement out of the for loop, then reversing the logic to 'break' if `lnk == nil`

path/resolver/resolver.go
```diff
@@ -69,25 +69,42 @@ func (r *Resolver) ResolveToLastNode(ctx context.Context, fpath path.Path) (ipld
 	}
 
 	for len(p) > 0 {
-		val, rest, err := nd.Resolve(p)
+		lnk, rest, err := r.ResolveOnce(ctx, r.DAG, nd, p)
+		if lnk == nil {
```
Stebalien: This'll drop the error, as far as I can tell.

whyrusleeping: Ah, interesting. Can we make sure to document that explicitly? At a glance, it looks wrong, so let's make sure nobody comes along and 'fixes' it

Stebalien: We should probably remove the error check below in that case.

magik6k: It can technically still be set at that point, but that would be a rather weird edge case

PR #5277, opened 2018-07-23
Extract config package

magik6k:
Currently IPTB depends on go-ipfs directly which makes it hard to use it with anything that isn't go-ipfs (like go-ipfs-api) - https://github.com/ipfs/iptb/issues/48.

TODO:
* [x] Is this the right thing to do?
* [x] Push extracted repo to github - https://github.com/ipfs/go-ipfs-config
* [x] Fix/figure out how to manage `repo/config/version.go`
* [x] Update iptb to use `go-ipfs-config` - https://github.com/ipfs/iptb/pull/63


repo/config/version.go
```diff
@@ -1,9 +0,0 @@
-package config
-
-// CurrentCommit is the current git commit, this is set as a ldflag in the Makefile
-var CurrentCommit string
-
-// CurrentVersionNumber is the current application's version literal
-const CurrentVersionNumber = "0.4.17-dev"
```
whyrusleeping: Let's find a way to not have the binary version in a different package. I really don't think this stuff needs to be in the config package anyways

magik6k: Yep, that definitely belongs here.

The problem is that we apparently embed that in config on init:
https://github.com/ipfs/go-ipfs/blob/80fcc4ecf177b3b6a8daa10559eb46bcf8a931db/repo/config/init.go#L27-L32

This was introduced in https://github.com/ipfs/go-ipfs/commit/feef5c3415ec249cf307f43a422545de24aee319#diff-22f880839db0df9b05dcad0718392fc9R28 which is in 0.4.16.

From what I can see we can safely drop that as in the only place it's used it's overridden with https://github.com/ipfs/go-ipfs/blob/master/core/corehttp/commands.go#L80

whyrusleeping: That sounds good to me.

whyrusleeping:
ci seems upset

magik6k:
fixed

repo/version.go
```diff
@@ -1,9 +1,9 @@
-package config
```
Stebalien: This doesn't really feel like the right place to put this. Personally, I'd put it in the root along side `doc.go`.

Stebalien:
One nit but, other than that (and a rebase once we cut a release), LGTM.

whyrusleeping:
rebased and squashed this. Since i was at fault for the merge conflicts >.>

kevina:
> Is this the right thing to do?

There didn't seam to be any discussion on this point.  It seams very odd to mean that we are putting the configuration for IPFS in an external package.  That is one thing that seams very IPFS specific.

kevina:
> There are packages which depend on this (iptb to name one)

It just seams to be that there should be a better way, that's all.

magik6k:
Needs a rebase now, but other than that it's good to go (I'll handle rebase tomorrow)

Stebalien:
> It just seams to be that there should be a better way, that's all.

The config package could use some cleanup but allowing other programs to parse the config without pulling in all of go-ipfs seems generally useful.

Stebalien:
@kevina although yeah, I kind of agree that this really does belong in this repo. However, we'll need to fix the gx situation for that to be a viable option.

Once we:

1. Make gx more pleasant to use.
2. Stop rewriting gx paths by default.

We can reevaluate this.

whyrusleeping:
🔥 🔥 👌 🔥 👌  Go magik go!

PR #5283, opened 2018-07-25
urlstore: Accept "200 OK" in addition to "206 Partial Content".

### LINKED ISSUE #5250: urlstore ipfs cat - expected HTTP 206 got 200
Some servers seem to return `200 OK` when range header covers entire file:
```
$ ipfs urlstore add https://archive.archlinux.org/packages/f/fio/
zb2rhZbRnLwHgcz7J3WtwJY2jLRJwWKsvPGMFC6r6kSh1HNP5
$ ipfs cat zb2rhZbRnLwHgcz7J3WtwJY2jLRJwWKsvPGMFC6r6kSh1HNP5
Error: failed to get block for zb2rhZbRnLwHgcz7J3WtwJY2jLRJwWKsvPGMFC6r6kSh1HNP5: expected HTTP 206 got 200
```

cc @kevina

kevina:
Some servers seem to return 200 OK when range header covers entire file.  If the content is wrong we will detect later so there is no harm in accepting either response.

Closes #5250.

### DIFF
--- a/filestore/fsrefstore.go
+++ b/filestore/fsrefstore.go
@@ -216,9 +216,9 @@ func (f *FileManager) readURLDataObj(c *cid.Cid, d *pb.DataObj) ([]byte, error) 
 	if err != nil {
 		return nil, &CorruptReferenceError{StatusFileError, err}
 	}
-	if res.StatusCode != http.StatusPartialContent {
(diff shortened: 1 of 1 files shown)


whyrusleeping:
SGTM

Issue #5231, opened 2018-07-13
Refactor blockstore to work with bare multihashes instead of CIDs


kyledrake:
This is needed to prevent issues with blockstore data duplication when different CIDs are provided for the same multihash.

The "quick fix" version is to check for both, but a proper conversion and migration would be better for preventing dupes.

https://github.com/ipfs/ipfs/issues/337

@kevina

kevina:
See https://github.com/ipfs/go-ipfs/pull/5285 for the "quick fix".

kevina:
See https://github.com/ipfs/go-ipfs-blockstore/issues/8 for a discussion on what the new interface should be.

lidel:
Continued in https://github.com/ipfs/go-ipfs/issues/6815

PR #5285, opened 2018-07-26
Add ability to retrieve blocks even if given using a different CID version

kevina:
~~This is one way to do it.~~

~~Another way is to create (yet another) wrapper blockstore.  However, this block store needs to be implemented very carefully to avoid breaking things.  In particular only the `Get` method can be changed.~~

~~Although it would seam logical the `Has` method can not  be modified because otherwise that might prevent blocks from getting added that should as we check if a block exists before adding it in the blockservice.  This could create a problem if for example a CIDv0 block is in the blockstore and a CIDv1 of the block is added and pinned.  If the Has method was modified then the CIDv1 will never get added and if the CIDv0 version is not pinned it would get removed when the garbage collector is ran and both the CIDv0 and CIDv1of the block is will become inaccessible.~~

~~I added a test case to test for the above case.  Any alternative solutions would need to pass my tests.~~

This needs to be implemented carefully to avoid breaking pinning.  I added a test case to test to show the problem.

See https://github.com/ipfs/go-ipfs/pull/5285#issuecomment-408649310 for updated method.

Towards https://github.com/ipfs/ipfs/issues/337.

### DIFF
--- a/thirdparty/cidv0v1/blockstore.go
+++ b/thirdparty/cidv0v1/blockstore.go
@@ -1,6 +1,7 @@
 package cidv0v1
 
 import (
+	mh "gx/ipfs/QmPnFwZ2JXKnXgMw8CdBPxn7FWh6LLdjUjxV1fKHuJnkr8/go-multihash"
 	blocks "gx/ipfs/QmVzK524a2VWLqyvtBeiHKsUAWYgeAk4DBeZoY7vpNPNRx/go-block-format"
 	cid "gx/ipfs/QmYVNvtQkeZ6AKSwDrjQTs432QtL6umrrK41EBq3cu7iSP/go-cid"
 	bs "gx/ipfs/QmadMhXJLHMFjpRmh85XjpmVDkEtQpNYEZNRpWRvYVLrvb/go-ipfs-blockstore"
@@ -58,7 +59,7 @@ func (b *blockstore) Get(c *cid.Cid) (blocks.Block, error) {
 
 func tryOtherCidVersion(c *cid.Cid) *cid.Cid {
 	prefix := c.Prefix()
-	if prefix.Codec != cid.DagProtobuf {
+	if prefix.Codec != cid.DagProtobuf || prefix.MhType != mh.SHA2_256 || prefix.MhLength != 32 {
 		return nil
 	}
 	var c1 *cid.Cid
--- a/thirdparty/cidv0v1/blockstore.go
+++ b/thirdparty/cidv0v1/blockstore.go
@@ -1,0 +1,71 @@
+package cidv0v1
+
+import (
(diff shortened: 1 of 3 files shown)


magik6k:
I'd say that this should be done on the blockstore layer so that bitswap can benefit from that too. We'll need to make GC operate on multihashes to keep it working in that case.

kevina:
> I'd say that this should be done on the blockstore layer so that bitswap can benefit from that too.

I forgot about that.  I have not tested with bitswap, but for bitswap to work I assume the Has method needs to be implemented, which will create the problem outlined above.

> We'll need to make GC operate on multihashes to keep it working in that case.

That will be difficult or at least messy.  The problem is if you only have the multihash you won't be able to delete anything until a full migration is done.  If the multihash corresponds to a CIDv1 we will not be able to delete the block as we lost the `Codec` field of the CID, thus making it impossible to reconstruct the CID and find the block in the blockstore to delete.

It may be possible to modify the GC other ways to make this work though...

It may be possible to make this work, but it could be messy.

Stebalien:
After discussing this on IRC, I believe that this would be strictly better than the situation we have today so I'm fine with it.

kevina:
> But we should try to get in the codec-agnostic blockstore ASAP.

I agree.

I will rebase this shortly.

kevina:
Rebased.

Stebalien:
By strictly better I mean that, without this patch, we'd go to the network to fetch the block in question (possibly failing, possibly succeeding). In the success case, we'd store the duplicate block anyways. With this patch, we just store it up-front.

thirdparty/cidv0v1/blockstore.go
```diff
@@ -0,0 +1,71 @@
...
+	err = b.Blockstore.Put(block)
+	if err != nil {
+		return nil, err
+	}
+	return block, nil
+}
+
+func tryOtherCidVersion(c *cid.Cid) *cid.Cid {
+	prefix := c.Prefix()
+	if prefix.Codec != cid.DagProtobuf {
```
magik6k: This should check hash alg too

kevina: Fixed.

rob-deutsch:
Okay, so I have a potentially silly question - why do Bockstores/Blockservices/Bitswap still operate on `cid.Cid` instead of `multihash.Multihash`?

I would've thought the multihash is the only important attribute for storage of the block - and that the other fields of CIDv1 were only useful for linking purposes.

kevina:
@rob-deutsch you are correct.  The blockservice will be converted as part of the move to Cidv1 base32.  See  https://github.com/ipfs/go-ipfs/issues/5231 Bitswap will continue to use Cids for some future plan where the links are important.  I do not know anything about this.

In any case this is not the place to discuss this.

momack2:
@kevina is this still open?

PR #5296, opened 2018-07-28
Extract blockservice and verifcid

whyrusleeping:
The extraction is steaming ahead! 

License: MIT


kevina:
(1) This may complicate https://github.com/ipfs/go-ipfs/pull/5285

(2) Why are we extracting verifcid?  It is temporary and will go away once a more  permanent solution lands which I hope to have within a month or so.

kevina:
@whyrusleeping 

> @kevina we can reapply the diffs as needed. This is blocking other things i'm working on. If you want, I can handle porting any conflicting patches.

Okay.  For now I won't rebase #5285 as I am exploring multiple solutions and some of them might require modifying the blockservice (although I would like to avoid it, the alternative might be some nasty modifications to the gc and pinning).

> Thats okay, we can delete it when we finish that. See the note in the readme for that repo, it explicitly says its temporary.

I am not sure I understand the point then.  But it doesn't cause me any problems so I don't have a problem with it.

whyrusleeping:
> Nevermind: I take it is needed because it is used in the blockservice.

Ah, yeah. Thats the reason. Forgot to directly mention that

PR #5298, opened 2018-07-28
Extract merkledag package, move dagutils to top level

whyrusleeping:
Extraction PR 3 out of 5 (I think). Making building tools on top of ipfs much much easier as it will no longer will require importing all of go-ipfs, and also provides the option for people to not have to deal with gx at all :)


whyrusleeping:
@magik6k yeah, forgot to pin it before going to sleep. Doing that now.

PR #5308, opened 2018-07-29
remove bitswap unwant

### LINKED ISSUE #5295: bitswap unwant
#### Version information:
go-ipfs version: 0.4.16-
Repo version: 7
System version: amd64/windows
Golang version: go1.10.2

#### Description:
The command `ipfs bitswap unwant` and API `/api/v0/bitswap/unwant` are not working.

```
> ( ipfs block getQmPfMX1epKLZNxG3cAoXYehjiP3inGpzFxFFKW1EregzgC ) &
> ipfs bitswap wantlist
QmPfMX1epKLZNxG3cAoXYehjiP3inGpzFxFFKW1EregzgC

> ipfs bitswap unwant QmPfMX1epKLZNxG3cAoXYehjiP3inGpzFxFFKW1EregzgC

> ipfs bitswap wantlist
QmPfMX1epKLZNxG3cAoXYehjiP3inGpzFxFFKW1EregzgC
```

Stebalien:
This command messes with internal state and doesn't even work at the moment. If
you don't want a block, you should cancel the request that's trying to fetch it.

An alternative is to replace this command with a dummy command that prints an
error but we'll break user code either way and I'd rather delete code.

fixes #5295

### DIFF
--- a/core/commands/bitswap.go
+++ b/core/commands/bitswap.go
@@ -16,7 +16,6 @@ import (
 	cmds "gx/ipfs/QmPTfgFTo9PFr1PvPKyKoeMgBvYPh6cX3aDP7DHKVbnCbi/go-ipfs-cmds"
 	peer "gx/ipfs/QmQsErDt8Qgw1XrsXf2BpEzDgGWtB1YLsTAARBup5b6B9W/go-libp2p-peer"
 	cmdkit "gx/ipfs/QmSP88ryZkHSRn1fnngAaV2Vcn63WUJzAavnRM9CVdU1Ky/go-ipfs-cmdkit"
-	cid "gx/ipfs/QmZFbDTY9jfSBms2MchvYM9oYRbAF19K7Pby47yDBfpPrb/go-cid"
 )
 
 var BitswapCmd = &cmds.Command{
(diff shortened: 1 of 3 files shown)


ivan386:
It was way to send message that you have block to connected peer that don't send you want message yet.

Stebalien:
> It was way to send message that you have block to connected peer that don't send you want message yet.

I'm not sure if I understand what you're saying but this command attempts to *cancel* a want. That is, it removes a block from a wantlist, even if some other part of the system is requesting it. I'm proposing that we remove it because it destructively interferes with other commands and isn't the correct way to stop asking for a block.

lgierth:
cc @alanshaw

Stebalien:
@diasdavid this command really isn't the correct way to solve any problem and it interferes with other commands. 100% of the time, the user should cancel the command that is *causing* us to want something.

diasdavid:
No blockers from me on removing it from the Public API. @Stebalien can you please (🙏🏽) give a pass through the specs while you issue this update?

- https://github.com/ipfs/specs/tree/master/bitswap#api-spec
- https://github.com/ipfs/specs/tree/master/public-api

//cc @alanshaw, @achingbrain & @hugomrdias

Stebalien:
@diasdavid done: https://github.com/ipfs/specs/pull/188

alanshaw:
@Stebalien what's your release plan for this? go-ipfs 0.5?

We can deprecate this now in `js-ipfs-api` (print a message to the console when used) and also in `js-ipfs`. There's an upcoming 0.32 `js-ipfs` release so it would be good to get the deprecation message in there and then remove it the next release.

Stebalien:
> what's your release plan for this

Just a normal release? We don't really follow semver (yet).

I really wasn't planning on having a deprecation process given that this doesn't currently work (in go at least) and nobody should be using it anyways. The user will just get a "command not found error".

I could *also* just switch it to a dummy command that prints an error.

PR #5316, opened 2018-07-30
Extract unixfs and importers to go-unixfs

whyrusleeping:


whyrusleeping:
This is the last one for now.

whyrusleeping:
@schomatis @Mr0grog I could use some help with the readme for go-unixfs: https://github.com/ipfs/go-unixfs

I started out with a directory explaining what each of the subpackages in it is for, but it would be great to expand that a bit, and maybe show a couple code examples.

schomatis:
> I started out with a directory explaining what each of the subpackages in it is for, but it would be great to expand that a bit, and maybe show a couple code examples.

Yes, that's great! I'll get on it (I can take the lead on this one @Mr0grog, and ping you for a review later on).

schomatis:
I'm a bit lost, should the remaining [`unixfs`](https://github.com/ipfs/go-ipfs/tree/4bca53e0c9b52b1e2a066aa21d1e05d44bffd1ac/unixfs) directory here be ignored? I mean, should I move #5257 PR to `go-unixfs`, right?

PR #5322, opened 2018-07-31
wait for all connections to close before exiting on shutdown.

### LINKED ISSUE #4055: Error message when shutting down ipfs daemon
#### Version information:
```
go-ipfs version: 0.4.10-4679f80
Repo version: 5
System version: amd64/darwin
Golang version: go1.8.3
```

#### Type:
error message without any noticeable error

#### Severity:
very low

#### Description:
I ran `ipfs shutdown` on the daemon, which had been started with `--writable`, and received the following error message (in the second terminal, not in the daemon's terminal).

```
Error: Post http://127.0.0.1:5001/api/v0/shutdown?encoding=json&stream-channels=true: EOF
```

Daemon shutdown worked nonetheless.

Stebalien:
Using httpServer.Shutdown will:

1. Close the listener (preventing new connections).
2. Close each connection as outstanding requests finish.

Turns out this issue wasn't as tricky as I had thought.

This prevent us from shutting down before outstanding requests get a chance to respond.

fixes #4055

### DIFF
--- a/core/commands/shutdown.go
+++ b/core/commands/shutdown.go
@@ -3,31 +3,28 @@ package commands
 import (
 	"fmt"
 
+	cmds "gx/ipfs/QmNueRyPRQiV7PUEpnP4GgGLuK1rKQLaRW7sfPvUetYig1/go-ipfs-cmds"
 	"gx/ipfs/QmdE4gMduCKCGAcczM2F5ioYDfdeKuPix138wrES1YSr7f/go-ipfs-cmdkit"
-
-	cmds "github.com/ipfs/go-ipfs/commands"
 )
 
 var daemonShutdownCmd = &cmds.Command{
 	Helptext: cmdkit.HelpText{
 		Tagline: "Shut down the ipfs daemon",
 	},
-	Run: func(req cmds.Request, res cmds.Response) {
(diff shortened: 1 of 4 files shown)


core/commands/shutdown.go
```diff
@@ -3,31 +3,28 @@ package commands
...
 
+	cmds "gx/ipfs/QmNueRyPRQiV7PUEpnP4GgGLuK1rKQLaRW7sfPvUetYig1/go-ipfs-cmds"
 	"gx/ipfs/QmdE4gMduCKCGAcczM2F5ioYDfdeKuPix138wrES1YSr7f/go-ipfs-cmdkit"
-
-	cmds "github.com/ipfs/go-ipfs/commands"
 )
 
 var daemonShutdownCmd = &cmds.Command{
 	Helptext: cmdkit.HelpText{
 		Tagline: "Shut down the ipfs daemon",
```
magik6k: Maybe `Schedule shutdown of the ipfs daemon`?

Stebalien: Well, it *does* shut it down. This will:

1. Cancel in-progress requests.
2. Wait for the IPFS node to close.
3. Return.

Note: it won't actually *wait* for all in-progress requests to finish. That would deadlock.

Issue #5311, opened 2018-07-30
`ipfs dht findprovs` can't handle `/ipfs/$hash` 


VictorBjelkholm:
#### Version information:
```
go-ipfs version: 0.4.15-
Repo version: 6
System version: amd64/linux
Golang version: go1.10.2
```

#### Type:
Bug

#### Description:

Trying to find out how many providers there are for a hash and it's link, so wanted to run things via `ipfs resolve` that returns in the format of `/ipfs/$hash`. However, `ipfs dht findprovs` only seems to accept the hashes without the `/ipfs` prefix. When trying with `/ipfs/$hash`, the error message is `Error: selected encoding not supported`

```
$ ARXIV_HASH=QmfXH9XtP7xmoTH8WAp4HNSduqWMwLTH8B8TvbTkdgzNAa && ipfs resolve "/ipfs/$ARXIV_HASH"
/ipfs/QmfXH9XtP7xmoTH8WAp4HNSduqWMwLTH8B8TvbTkdgzNAa

$ ARXIV_HASH=QmfXH9XtP7xmoTH8WAp4HNSduqWMwLTH8B8TvbTkdgzNAa && ipfs dht findprovs $(ipfs resolve "/ipfs/$ARXIV_HASH")
Error: selected encoding not supported

$ ipfs dht findprovs /ipfs/QmfXH9XtP7xmoTH8WAp4HNSduqWMwLTH8B8TvbTkdgzNAa
Error: selected encoding not supported

$ ipfs dht findprovs QmfXH9XtP7xmoTH8WAp4HNSduqWMwLTH8B8TvbTkdgzNAa
QmTiCerJDqrC6vLDgcNzu8qNK4e9rTx6bJLtSgQHpkUUfk
QmViCToRk7k8asw6B7XNiC8bjomGG26XqXPWWb4WB5HKPL
QmcQzakZ46etK56gWJWqT3GqZRMCw6ovPUWEUBeCf5pUpK
^C
Error: context canceled
```

whyrusleeping:
need to do a path resolution before passing it to the dht, I think we're currently just calling cid.Parse

grokcoder:
the source code just decode input argument using cid.Decode 

```
c, err := cid.Decode(req.Arguments()[0])
if err != nil {
    res.SetError(err, cmdkit.ErrNormal)
    return
}
```
so what is the legal input for `findprovs`, cid or
 all of 
[...]

PR #5329, opened 2018-08-02
fix: dht findprovs unable to handle /ipfs/hash #5311

grokcoder:
fix `ipfs dht findprovs` can't handle `/ipfs/$hash`  ref #5311

License: MIT
Signed-off-by: Xiaoyi Wang <wangxiaoyi@hyperchain.cn>


grokcoder:
@VictorBjelkholm

PR #5330, opened 2018-08-02
[WIP] gx: update `go-merkledag` to match version used in `go-unixfs`

schomatis:
Do not merge.

See https://github.com/ipfs/go-unixfs/issues/4.

### DIFF
--- a/core/builder.go
+++ b/core/builder.go
@@ -15,9 +15,9 @@ import (
 	cfg "github.com/ipfs/go-ipfs/repo/config"
 	"github.com/ipfs/go-ipfs/thirdparty/verifbs"
 	bserv "gx/ipfs/QmNqRBAhovtf4jVd5cF7YvHaFSsQHHZBaUFwGQWPM2CV7R/go-blockservice"
-	dag "gx/ipfs/QmRy4Qk9hbgFX9NGJRm8rBThrA8PZhNCitMgeRYyZ67s59/go-merkledag"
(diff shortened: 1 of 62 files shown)


schomatis:
Closed due to forced push.

whyrusleeping:
You'll also need to update go-blockservice I think. Run `gx deps dupes` to see if anything looks wrong.

schomatis:
`gx deps dupes` reported (among many duplicates) the old hash of `go-merkledag` being used, I grep'ed for it in the `gx` dir (is there an easier way for this? a `gx` flag to know which package is importing the old hash?) and it seems that the `go-path` version used in `go-ipfs` is using it, although this doesn't affect the build (probably because those two don't interact), should I update the `go-path` version in this same PR?

schomatis:
So, this is breaking the build because I updated `go-merkledag` to the version used by the `master` branch  of `go-unixfs` (for `gx-go link` to work) but `go-ipfs` is using a gx-version of `go-unixfs` which is actually using the old `go-merkledag` version (`QmRy4Q...s59`) that I'm updating here, so I guess I need to first update `go-unixfs` here to the master version gx-release (that uses the new `go-merkledag` version `QmfGzd...9TS`) and then update `go-merkledag`, or maybe that needs to happen in the same PR otherwise I think there might be a version mismatch in the converse relation of the what's happening here now that's breaking the build (my head is aching at this point so I'm going to take a pause here and rethink this, let's not merge anything until then).

schomatis:
Closing in favor of #5336, thanks @hsanjuan!

PR #5331, opened 2018-08-02
block cmd: Use coreapi

magik6k:


magik6k:
Blocked by a bug in go-path

core/commands/block.go
```diff
@@ -282,38 +235,42 @@ It takes a list of base58 encoded multihashes to remove.
...
 
-			cids = append(cids, c)
-		}
-		ch, err := util.RmBlocks(n.Blockstore, n.Pinning, cids, util.RmBlocksOpts{
-			Quiet: quiet,
-			Force: force,
-		})
+			rp, err := api.ResolvePath(req.Context, p)
+			if err != nil {
+				res.SetError(err, cmdkit.ErrNormal)
```
Kubuxu: Should we emit error here too?
What do you think?

Kubuxu: Reading original code we probably shouldn't unless `ResolvePath` does a lot more than just `Decode`.

core/coreapi/block.go
```diff
@@ -38,13 +38,29 @@ func (api *BlockAPI) Put(ctx context.Context, src io.Reader, opts ...caopts.Bloc
...
 	pref.Version = 1
 
+	if settings.Codec == "" {
+		if settings.MhType != mh.SHA2_256 || (settings.MhLength != -1 && settings.MhLength != 32) {
+			settings.Codec = "protobuf"
+		} else {
+			settings.Codec = "v0"
+		}
+	}
+
```
Kubuxu: Should this be in `BlockPutOptions` instead, it seems like it is options handling.

Kubuxu: This way we separate Options validation from implementation.

core/commands/block.go
magik6k: It does handle ipns and path traversal too

PR #5337, opened 2018-08-03
Feat: depth limited refs -r

hsanjuan:
This adds --max-depth to the "refs" commands and allows limiting
the fetching of refs per depth. Other than that, it works as before.

Note that clever branch pruning is only made when the --unique flag
is passed. Otherwise, we re-explore branches to the given depth.

---

First minimal approach. I wonder if we could utilize the EnumerateChildren functions here instead, @Stebalien ? Doing the printing inside the custom visit function. I'm not sure if there are reasons why the DAG traversal logic was re-implemented separately for this command.

Also, now or later, I would like to do an `--async` version of this, so it would be very easy if we re-use EnumerateChildrenAsync().

### DIFF
--- a/core/commands/refs.go
+++ b/core/commands/refs.go
@@ -10,11 +10,11 @@ import (
 	cmds "github.com/ipfs/go-ipfs/commands"
 	"github.com/ipfs/go-ipfs/core"
 	e "github.com/ipfs/go-ipfs/core/commands/e"
-	path "gx/ipfs/QmdMPBephdLYNESkruDX2hcDTgFYhoCt4LimWhgnomSdV2/go-path"
 
-	"gx/ipfs/QmSP88ryZkHSRn1fnngAaV2Vcn63WUJzAavnRM9CVdU1Ky/go-ipfs-cmdkit"
+	cmdkit "gx/ipfs/QmSP88ryZkHSRn1fnngAaV2Vcn63WUJzAavnRM9CVdU1Ky/go-ipfs-cmdkit"
 	ipld "gx/ipfs/QmX5CsuHyVZeTLxgRSYkgLSDQKb9UjE8xnhQzCEJWWWFsC/go-ipld-format"
 	cid "gx/ipfs/QmZFbDTY9jfSBms2MchvYM9oYRbAF19K7Pby47yDBfpPrb/go-cid"
+	path "gx/ipfs/QmdMPBephdLYNESkruDX2hcDTgFYhoCt4LimWhgnomSdV2/go-path"
 )
 
 // KeyList is a general type for outputting lists of keys
@@ -64,6 +64,7 @@ NOTE: List all references recursively by using the flag '-r'.
 		cmdkit.BoolOption("edges", "e", "Emit edge format: `<from> -> <to>`."),
 		cmdkit.BoolOption("unique", "u", "Omit duplicate refs from output."),
 		cmdkit.BoolOption("recursive", "r", "Recursively list links of child nodes."),
+		cmdkit.IntOption("max-depth", "Only for recursive refs, limits fetch and listing to the given depth").WithDefault(-1),
 	},
 	Run: func(req cmds.Request, res cmds.Response) {
 		ctx := req.Context()
@@ -83,6 +84,16 @@ NOTE: List all references recursively by using the flag '-r'.
 		if err != nil {
 			res.SetError(err, cmdkit.ErrNormal)
 			return
+		}
+
+		maxDepth, _, err := req.Option("max-depth").Int()
+		if err != nil {
+			res.SetError(err, cmdkit.ErrNormal)
+			return
+		}
+
+		if !recursive {
+			maxDepth = 1 // write only direct refs
 		}
 
 		format, _, err := req.Option("format").String()
@@ -119,12 +130,12 @@ NOTE: List all references recursively by using the flag '-r'.
 			defer close(out)
 
 			rw := RefWriter{
-				out:       out,
-				DAG:       n.DAG,
-				Ctx:       ctx,
-				Unique:    unique,
-				PrintFmt:  format,
-				Recursive: recursive,
+				out:      out,
+				DAG:      n.DAG,
+				Ctx:      ctx,
+				Unique:   unique,
+				PrintFmt: format,
+				MaxDepth: maxDepth,
 			}
 
 			for _, o := range objs {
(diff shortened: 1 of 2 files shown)


hsanjuan:
Note: i am missing some sharness tests still. Will do that when we have a final approach.

core/commands/refs.go
```diff
@@ -274,43 +290,40 @@ func (rw *RefWriter) writeRefsRecursive(n ipld.Node) (int, error) {
...
-func (rw *RefWriter) writeRefsSingle(n ipld.Node) (int, error) {
-	c := n.Cid()
-
-	if rw.skip(c) {
-		return 0, nil
+// visit returns false if we should not re-explore the tree under a Cid, or
+// true otherwise. The second return argument indicates whether the Cid was seen
+// before.
+func (rw *RefWriter) visit(c *cid.Cid, depth int) (bool, bool) {
+	if rw.seen == nil {
```
Stebalien: I believe disabling unique by default is a memory optimization. If it's disabled, we should avoid using memory linear in the number of keys. We should probably have check for the `Unique` flag up-top and, in that case, only check the depth (don't store anything).

Stebalien: It may also be simpler to have one boolean indicate that we should continue traversing and the other indicate that we should return the CID to the user. That may simplify some of the other logic. It'll also allow us to say "return this but *don't* traverse it which, unless I'm mistaken, should save us a bit of work (possibly saving us from fetching a node we don't need to traverse).

Stebalien:
> I'm not sure if there are reasons why the DAG traversal logic was re-implemented separately for this command.

I believe this command came before the EnumerateChildren functions. We could probably re-use that logic (although we should probably use the non-async one unless the user passes some `--in-order=false` flag as users may be relying on traversal order...).

hsanjuan:
@Stebalien:

1) I think I have addressed both of your comments. Traversal logic is slightly simpler now.

2) I tried to use `merkledag.EnumerateChildren` but doesn't work maintaining the custom `--format` options (needs parent and linkname, apart from the curren Cid), so I left things as they were. Do you think we should provide an `async`/`--in-oder=false` method where we disregard the custom `--format` flag? I'd rather not duplicate the EnumerateChildrenAsync logic. (this would be different PR).

core/commands/refs.go
```diff
@@ -274,43 +286,53 @@ func (rw *RefWriter) writeRefsRecursive(n ipld.Node) (int, error) {
...
 }
 
-func (rw *RefWriter) writeRefsSingle(n ipld.Node) (int, error) {
-	c := n.Cid()
-
-	if rw.skip(c) {
-		return 0, nil
+// visit returns two values:
+// - first indicates if we should keep traversing the DAG.
+// - second indicates if the given Cid should be printed to the user.
```
Stebalien: "indicates" is ambiguous. In the first case, it means "is true" and, in the second case, "is false".

Personally, I'd rather:

1. Say "is true" (explicitly).
2. Invert the second case (i.e., true means print).

core/commands/refs.go
```diff
@@ -231,41 +242,42 @@ type RefWriter struct {
...
-func (rw *RefWriter) writeRefsRecursive(n ipld.Node) (int, error) {
+func (rw *RefWriter) writeRefsRecursive(n ipld.Node, depth int) (int, error) {
 	nc := n.Cid()
 
 	var count int
 	for i, ng := range ipld.GetDAG(rw.Ctx, rw.DAG, n) {
 		lc := n.Links()[i].Cid
-		if rw.skip(lc) {
-			continue
+		unexplored, written := rw.visit(lc, depth+1) // The children are at depth+1
```
Stebalien: "written" isn't quite correct. Really, it means "don't write" (we may not have already written it, unless I'm mis-reading the code).

Note: if we invert this case, this'll obviously become "shouldWrite" or something like that. I'm just dropping a comment here so we don't miss it.

core/commands/refs.go
```diff
@@ -274,43 +286,53 @@ func (rw *RefWriter) writeRefsRecursive(n ipld.Node) (int, error) {
...
+// rw.MaxDepth = 1 disables any pruning. But setting rw.Unique to true will
+// prune already visited branches.
+func (rw *RefWriter) visit(c *cid.Cid, depth int) (bool, bool) {
+	overMaxDepth := rw.MaxDepth >= 0 && depth > rw.MaxDepth
+
+	// When not unique, only avoid exploring too deep
+	// We do not track a set of visited nodes in this case.
+	// We do not print anything too deep though.
+	if !rw.Unique {
+		return !overMaxDepth, overMaxDepth
```
Stebalien: Don't we want to not explore nodes *at* max depth as well?

core/commands/refs.go
```diff
@@ -274,43 +286,53 @@ func (rw *RefWriter) writeRefsRecursive(n ipld.Node) (int, error) {
...
-	}
+	key := string(c.Bytes())
+	oldDepth, ok := rw.seen[key]
 
-	if rw.seen == nil {
-		rw.seen = cid.NewSet()
+	// Never explore over max-depth. Never print nodes over
+	// max depth.
+	if overMaxDepth {
+		return false, false
```
Stebalien: Ditto about exploring nodes *at* max depth.

core/commands/refs.go
```diff
@@ -85,6 +86,16 @@ NOTE: List all references recursively by using the flag '-r'.
...
 		}
 
+		maxDepth, _, err := req.Option("max-depth").Int()
+		if err != nil {
+			res.SetError(err, cmdkit.ErrNormal)
+			return
+		}
+
+		if !recursive {
+			maxDepth = 1 // write only direct refs
```
Stebalien: Should this be `0`? If I'm not mistaken, this patch will currently explore two levels when recursive is disabled. If that's the case, can we *also* write a sharness test?

core/commands/refs.go
Stebalien: Also, shouldn't this check be higher up? No need to do anything if we're over the max depth.

core/commands/refs.go
```diff
@@ -274,43 +286,53 @@ func (rw *RefWriter) writeRefsRecursive(n ipld.Node) (int, error) {
...
-		return false
-	}
+	key := string(c.Bytes())
+	oldDepth, ok := rw.seen[key]
 
-	if rw.seen == nil {
-		rw.seen = cid.NewSet()
+	// Never explore over max-depth. Never print nodes over
+	// max depth.
+	if overMaxDepth {
```
magik6k: I'd put this just below `if !rw.Unique {..}`

Stebalien:
> 2

We can implement new features later (e.g., when we need them).

core/commands/refs.go
```diff
@@ -274,43 +286,53 @@ func (rw *RefWriter) writeRefsRecursive(n ipld.Node) (int, error) {
...
-	has := rw.seen.Has(c)
-	if !has {
-		rw.seen.Add(c)
+	// Branch pruning cases:
+	// - We saw the Cid before and either:
+	//   - Depth is unlimited (MaxDepth = -1)
+	//   - We saw it higher (smaller depth) in the DAG (means we must have
+	//     explored deep enough before)
+	if ok && (rw.MaxDepth < 0 || oldDepth <= depth) {
+		return false, true
```
magik6k: Shouldn't this be `return false, false`?

hsanjuan:
I have clarified the doc the comments and inverted the second return value (and moved up a check). As said, it does print items at MaxDepth, but not "over max depth". Right now, it won't `dag.Get` items who are at maxdepth, because their children will be over it and thus won't need printing. Items at max depth are printed when visiting the Links from the parent.

If we're are good with the code so far, I'll proceed to create some sharness tests (already did a fair amount of manual testing).

core/commands/refs.go
hsanjuan: Sorry, I think I misread the double negation. Right now I think it works as it should i.e.

--max-depth 1 prints just the direct children of the given CID, without fetching them (as they're links from the parent and we know we can't go deeper). Thus, items at maxdepth are never dag.Get(), but they are visited and potentially printed.

hsanjuan:
> 
>>    2

> We can implement new features later (e.g., when we need them).

Cluster would benefit from async, faster `refs -r`, as we do that to fetch content before pinning.

hsanjuan:
Sorry, it's late. It does dag.Get() items at MaxDepth. But I'm thinking this is what we want, because we want to use refs -r not only to print but to fetch blocks. So if it's printed it should be fetched (?)

Stebalien:
> So if it's printed it should be fetched

I was under the impression that we *didn't* get all nodes (e.g., we don't need to fetch raw leaves as we know they won't have links). However, it turns out this isn't the case.

Given how we tend to use this, I think it's reasonable to actually *fetch* all the nodes. In the future, we can add an option that avoids this.

Stebalien:
> Sorry, it's late. It does dag.Get() items at MaxDepth. But I'm thinking this is what we want, because we want to use refs -r not only to print but to fetch blocks. So if it's printed it should be fetched (?)

Actually, I'm pretty sure it'll fetch the children as well. `FetchGraph` will actually *initiate* a fetch and `nd.Get` will simply wait for the fetch to complete. We should probably just:

1. Avoid going deeper once we've hit max depth (not wait until we go *over* the limit).
2. Move the unexplored check *after* the `nd.Get(...)` (possibly adding some short-circuit to avoid fetching the block if we're nether printing it nor exploring it).

core/commands/refs.go
Stebalien: I see. This looks correct.

hsanjuan:
@Stebalien ok, another round. I realized I was ignoring the optimization that justified using promises: we do not need to do `Get` when pruning on an already seen cid/branch. So now:

1. ipld.GetDAG() returns promises for all the children of `n`, which we loop.
2. We visit(): returns if we `goDeeper` and `shouldWrite`. `goDeeper` is `false` if the Cid is at `MaxDepth` (and not over it as before).
3. New: We can skip doing Get on the promise if we printed the Cid before (meaning we did a Get() before) and must not go deeper. In that case we move on to next child. This justifies using promises in the first place. Otherwise we could just loop Links() and DAG.Get() every node.
4. We do `nd.Get()` at this point (either the node is new/not-printed, or we are going deeper). When !Unique we must get/print all nodes anyway because we don't know if they were Get before.
5. New: We increase `count`. I think this was buggy in original implementation and it stayed at 0 for recursive refs. `count` is not used and I'm not sure what it should count. Total number of references ? Unique number of references ? Number of times we Get() ? Whatever it is, the result may not have much sense depending on how branch pruning happens now that MaxDepth is in place.
* We WriteRef if we must
* If `goDeeper`, we go deeper, otherwise, work next Link.

So if I'm not wrong (again) this should:

* Not Get() nodes which were already Get()
* Not visit any cids which are over MaxDepth, thus never get any of their nodes
* Correctly Get() nodes to the right depth.
* Only print nodes which have been successfully Get(). I don't think we should print a Reference if we were not able to fetch its node even if we know it's cid (this should be same behaviour as before).

**fingers crossed**

core/commands/refs.go
```diff
@@ -231,86 +242,117 @@ type RefWriter struct {
...
-		if rw.skip(lc) {
-			continue
-		}
+		goDeeper, shouldWrite := rw.visit(lc, depth+1) // The children are at depth+1
 
-		if err := rw.WriteEdge(nc, lc, n.Links()[i].Name); err != nil {
-			return count, err
+		// Avoid "Get()" on the node. We did a Get on it before
+		// (we printed it) and must not go deeper. This is an
+		// optimization for pruned branches.
```
Stebalien: This comment is technically incorrect, unless I'm mistaken. We have already printed the node *and/or* we've pruned the branch.

core/commands/refs.go
```diff
@@ -231,86 +242,117 @@ type RefWriter struct {
...
+		if !shouldWrite && !goDeeper {
+			continue
 		}
 
+		// We must write it because it's new, or go deeper. In any case
```
Stebalien: s/write/get

core/commands/refs.go
```diff
@@ -231,86 +242,117 @@ type RefWriter struct {
...
+		// we need to make sure this node is fetched and stored.
 		nd, err := ng.Get(rw.Ctx)
 		if err != nil {
 			return count, err
 		}
+		count++
```
Stebalien: Shouldn't we only count it if we print it?

core/commands/refs.go
```diff
@@ -231,86 +242,117 @@ type RefWriter struct {
...
 	}
-	return has
+
+	// Final case, we must keep exploring the DAG from this CID
+	// (unless we hit the depth limit).
+	// We note down its depth because it was either not seen
+	// or is lower than last time.
+	// We print if it was not seen.
+	rw.seen[key] = depth
+	return !atMaxDepth, !ok
```
Stebalien: The comments in this function are *awesome*.

:heart:

core/commands/refs.go
hsanjuan: Sure, sgtm

Stebalien:
LGTM! (modulo tests).

hsanjuan:
Thanks @Stebalien @magik6k . I have added some sharness tests now (last commit).

Stebalien:
Jenkins passes and @magik6k has already reviewed (modulo tests). :bullettrain_front:

PR #5338, opened 2018-08-03
ci: check for gx deps dupes

magik6k:


magik6k:
Tested manually, seems to work as intended

PR #5339, opened 2018-08-03
 key cmd: Refactor to use coreapi 

magik6k:


whyrusleeping:
(the failing tests obvs should be fixed though)

Issue #4498, opened 2017-12-15
CoreAPI progress tracking PR


magik6k:
This PR is for tracking progress on the [CoreAPI](https://github.com/ipfs/go-ipfs/tree/master/core/coreapi) package

List of APIs to support (apis with **bold** names what is needed to replace go-ipfs-api with coreapi-based implementation without removing features, add more if I forgot something):
- [ ] **block**
- [ ] config
- [ ] **dag** - #4471
- [ ] **dht**
- [ ] files
- [ ] key - #4477
- [ ] **name** - #4477
- [ ] **object** - #4492
- [ ] **pin**
- [ ] **pubsub**
- [ ] swarm
- [ ] **unixfs** - Proper add is todo
- [ ] Move to interface-ipfs-core

(I'm aware of https://github.com/ipfs/interface-ipfs-core/issues/66, I'd say that this interface should get to a 'mostly done' stage here as it's much easier to refine it while it's in one place)

lgierth:
Thanks so much, great work :)

lgierth:
> what is needed to replace go-ipfs-api with coreapi-based implementation without removing features

We can just modify the `Shell` functions to use the CoreAPI under the hood.

diasdavid:
@magik6k ❤️ Thank you!

PR #5340, opened 2018-08-03
 coreapi: dag: Batching interface

magik6k:
I think this is the nicest way to get batching in coreapi, but if you have some better ideas, I'm happy to hear them

Part of https://github.com/ipfs/go-ipfs/issues/5215 / https://github.com/ipfs/go-ipfs/issues/4498

### DIFF
--- a/core/coreapi/interface/dag.go
+++ b/core/coreapi/interface/dag.go
@@ -22,6 +22,7 @@ type DagOps interface {
 type DagBatch interface {
 	DagOps
 
+	// Commit commits nodes to the datastore and announces them to the network
 	Commit(ctx context.Context) error
 }
 
@@ -35,5 +36,6 @@ type DagAPI interface {
 	// Tree returns list of paths within a node specified by the path.
 	Tree(ctx context.Context, path Path, opts ...options.DagTreeOption) ([]Path, error)
 
+	// Batch creates new DagBatch
 	Batch(ctx context.Context) DagBatch
 }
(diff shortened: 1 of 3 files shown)


core/coreapi/interface/dag.go
```diff
@@ -4,21 +4,38 @@ import (
...
 	Put(ctx context.Context, src io.Reader, opts ...options.DagPutOption) (ResolvedPath, error)
+}
+
+// DagBatch is the batching version of DagAPI. All implementations of DagBatch
+// should be threadsafe
+type DagBatch interface {
+	DagOps
+
+	// Commit commits nodes to the datastore and announces them to the network
+	Commit(ctx context.Context) error
```
magik6k: Note: need this context for go-ipfs-api implementation (batch will hold a connection and stream files to `dag/put`)

core/coreapi/interface/dag.go
```diff
@@ -4,21 +4,38 @@ import (
...
-	options "github.com/ipfs/go-ipfs/core/coreapi/interface/options"
+	"github.com/ipfs/go-ipfs/core/coreapi/interface/options"
 
 	ipld "gx/ipfs/QmZtNq8dArGfnpCZfx2pUNY7UcjGhVp5qqwQ4hH6mpTMRQ/go-ipld-format"
 )
 
-// DagAPI specifies the interface to IPLD
-type DagAPI interface {
+// DagOps groups operations that can be batched together
+type DagOps interface {
```
whyrusleeping: We will probably also have `Delete` on this interface, right?

magik6k: Well, there isn't `dag rm` so I didn't add it here. But yes, that would be the place for that (will be in Block API, but that's for a separate PR)

PR #5345, opened 2018-08-06
return a json object from config show

### LINKED ISSUE #5335: config/show should be application/json
Right now, go-ipfs HTTP API is responding to config?arg=key with:

```
Content-Type: application/json
```

But for `config/show`, we have:

```
Content-Type: text/plain
```

Since it's a JSON object, doesn't it make sense to change the Content Type to `application/json` on `config/show`?

Ref.: https://github.com/ipfs/js-ipfs-api/issues/822

Stebalien:
This doesn't change there CLI output, it just means that the server responds with a JSON output by default.

Fixes #5335.

### DIFF
--- a/core/commands/config.go
+++ b/core/commands/config.go
@@ -153,7 +153,7 @@ var configShowCmd = &cmds.Command{
 NOTE: For security reasons, this command will omit your private key. If you would like to make a full backup of your config (private key included), you must copy the config file from your repo.
 `,
 	},
-
+	Type: map[string]interface{}{},
 	Run: func(req cmds.Request, res cmds.Response) {
 		cfgPath := req.InvocContext().ConfigRoot
 		fname, err := config.Filename(cfgPath)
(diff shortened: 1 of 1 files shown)


Stebalien:
@diasdavid is this likely to break something?

Stebalien:
So, this does cause a bit of an issue: Users using *old* clients against *newer* servers won't see nicely formatted json, they'll see dense json. However, I'm not *too* concerned about that. Clients with *this* patch applied still work correctly against older servers.

whyrusleeping:
This seems fine to me, but I would double check with the javascripters before merging.

Stebalien:
@hacdias or @alanshaw, thoughts?

hacdias:
If the Content-Type header is now `application/json´ instead of `plain/text`, the js-ipfs-api will still work as expected.

Stebalien:
It is. These are the old headers:

```
< HTTP/1.1 200 OK
< Access-Control-Allow-Headers: X-Stream-Output, X-Chunked-Output, X-Content-Length
< Access-Control-Expose-Headers: X-Stream-Output, X-Chunked-Output, X-Content-Length
< Content-Type: text/plain
< Server: go-ipfs/0.4.18-dev
< Trailer: X-Stream-Error
< Vary: Origin
< X-Stream-Output: 1
< Date: Sat, 11 Aug 2018 16:44:20 GMT
< Transfer-Encoding: chunked
```

These are the new headers:

```
< HTTP/1.1 200 OK
< Access-Control-Allow-Headers: X-Stream-Output, X-Chunked-Output, X-Content-Length
< Access-Control-Expose-Headers: X-Stream-Output, X-Chunked-Output, X-Content-Length
< Content-Type: application/json
< Server: go-ipfs/0.4.18-dev
< Trailer: X-Stream-Error
< Vary: Origin
< Date: Sat, 11 Aug 2018 16:46:12 GMT
< Transfer-Encoding: chunked
```

Note that the `X-Stream-Output` header is also gone.

hacdias:
Great @Stebalien ! 😄

Stebalien:
@Kubuxu looks like this'll work *just fine* with js-ipfs-api as js-ipfs already behaves this way.

Stebalien:
I'll take the comments from @hacdias and @whyrusleeping as "two signoffs".

PR #5347, opened 2018-08-06
RFC: make GC codec agnostic

Stebalien:
This ensures that GC will work regardless of the codec on the CIDs returned by the blockstore.

### DIFF
--- a/pin/gc/gc.go
+++ b/pin/gc/gc.go
@@ -8,8 +8,8 @@ import (
 	"strings"
 
 	pin "github.com/ipfs/go-ipfs/pin"
-	dag "gx/ipfs/Qma2BR57Wqp8w9vPreK4dEzoXXk8DFFRL3LresMZg4QpzN/go-merkledag"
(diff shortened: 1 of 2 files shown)


pin/gc/gc.go
```diff
@@ -76,14 +76,16 @@ func GC(ctx context.Context, bs bstore.GCBlockstore, dstor dstore.Datastore, pn
 		errors := false
 		var removed uint64
 
+		gchashes := MultihashSetFromCids(gcs)
```
Stebalien: Note: we can't just use a multihash set when generating the colored set as we might visit a node with a raw CID first and then *not* visit its children when we visit it using a non-raw CID. Basically, traversal cares about CIDs.

kevina:
Note this p.r.  depends on a codec agnostic blockstore.  Or part of https://github.com/ipfs/go-ipfs/issues/5231.

Stebalien:
As @kevina pointed out, this *also* doesn't fix `block rm`. Closing as we can resurrect this if we need it later.

PR #5355, opened 2018-08-09
update gogo-protobuf

### LINKED ISSUE #3214: Update gogo protobuf dependency
We havent updated our gogo-protobuf dep in a while. The primary reason for this is that its not as simple to do as a lot of our other deps. The repo itself is huge, and our current gogo-protobuf package is roughly 40MB. This make the `gx install` ux slow and really annoying.

We should look at repackaging a newer version with exactly what we need, or finding a better way to shrink this package to just what we need to make the process of keeping things updated easier.

cc @Kubuxu @parkan @vyzo

Stebalien:
This includes:

1. The latest gogo protobuf version. Note: I've intentionally stripped the tests from the package to shrink it to a more manageable size.
2. Regenerated .pb.go files (*except* for those in go-merkledag).

It keeps the hashes the same by using the *existing* generated marshaler in go-merkledag. We can improve on this in the future by, e.g., regenerating everything *except* the marshaler.

fixes #3214


Stebalien:
Want to make sure the tests pass first before I merge/push the rest of the updates to the dependencies.

Stebalien:
Blocked on: https://github.com/ipfs/go-ipfs/pull/5355

PR #5356, opened 2018-08-09
update protobuf files in go-ipfs

Stebalien:
Also:

* Switch to gogo for filestore for consistency.
* Use the "faster" codegen for fewer allocations.
* fix some gx import issues


Stebalien:
@kevina bump.

kevina:
This seams okay.  But why are so many packages removed from `package.json`.

PR #5359, opened 2018-08-09
Allow mfs files.write command to create parent directories

achingbrain:
Adds support for a `-p/--parents` flag to the `files.write` command similar to the one supported by the `files.mkdir` command. If this is true and the directory for the file is not `"/"`, try to create the containing directory before writing to the file.

### DIFF
--- a/core/commands/files.go
+++ b/core/commands/files.go
@@ -721,6 +721,7 @@ stat' on the file or any of its ancestors.
 	Options: []cmdkit.Option{
 		cmdkit.IntOption("offset", "o", "Byte offset to begin writing at."),
 		cmdkit.BoolOption("create", "e", "Create the file if it does not exist."),
+		cmdkit.BoolOption("parents", "p", "Make parent directories as needed."),
 		cmdkit.BoolOption("truncate", "t", "Truncate the file to size zero before writing."),
 		cmdkit.IntOption("count", "n", "Maximum number of bytes to read."),
 		cmdkit.BoolOption("raw-leaves", "Use raw blocks for newly created leaf nodes. (experimental)"),
@@ -735,6 +736,7 @@ stat' on the file or any of its ancestors.
 		}
 
 		create, _ := req.Options["create"].(bool)
+		mkParents, _ := req.Options["parents"].(bool)
 		trunc, _ := req.Options["truncate"].(bool)
 		flush, _ := req.Options["flush"].(bool)
 		rawLeaves, rawLeavesDef := req.Options["raw-leaves"].(bool)
@@ -755,6 +757,14 @@ stat' on the file or any of its ancestors.
 		if offset < 0 {
 			re.SetError(fmt.Errorf("cannot have negative write offset"), cmdkit.ErrNormal)
 			return
+		}
+
+		if mkParents {
+			err := ensureContainingDirectoryExists(nd.FilesRoot, path, prefix)
+			if err != nil {
+				re.SetError(err, cmdkit.ErrNormal)
+				return
+			}
 		}
 
 		fi, err := getFileHandle(nd.FilesRoot, path, create, prefix)
@@ -1144,6 +1154,19 @@ func getPrefix(req oldcmds.Request) (cid.Builder, error) {
 	}
 
 	return &prefix, nil
+}
+
+func ensureContainingDirectoryExists(r *mfs.Root, path string, builder cid.Builder) error {
+	dirtomake := gopath.Dir(path)
+
+	if dirtomake == "/" {
+		return nil
+	}
+
+	return mfs.Mkdir(r, dirtomake, mfs.MkdirOpts{
+		Mkparents:  true,
+		CidBuilder: builder,
+	})
 }
 
 func getFileHandle(r *mfs.Root, path string, create bool, builder cid.Builder) (*mfs.File, error) {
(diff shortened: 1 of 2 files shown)


core/commands/files.go
```diff
@@ -735,6 +736,7 @@ stat' on the file or any of its ancestors.
 		}
 
 		create, _ := req.Options["create"].(bool)
+		dashp, _ := req.Options["parents"].(bool)
```
Stebalien: Let's call this "mkParents" or something like that. Contrary to what the rest of the go-ipfs code might indicate, we don't pay by the letter in variable names :smile:.

core/commands/files.go
```diff
@@ -757,6 +759,22 @@ stat' on the file or any of its ancestors.
 			return
 		}
 
+		dirtomake := gopath.Dir(path)
+
+		if dirtomake != "/" && dashp == true {
```
Stebalien: No need for the `dashp == true`, `dashp` will suffice.

core/commands/files.go
```diff
@@ -757,6 +759,22 @@ stat' on the file or any of its ancestors.
...
+			root := nd.FilesRoot
+
+			err = mfs.Mkdir(root, dirtomake, mfs.MkdirOpts{
```
Stebalien: Given that `getFileHandle` already has special logic for creating missing files, I'd prefer to add this logic to `getFileHandle`, replacing the the `create` bool with a [constant](https://splice.com/blog/iota-elegant-constants-golang/). However, I don't feel too strongly about this.

core/commands/files.go
```diff
@@ -757,6 +759,22 @@ stat' on the file or any of its ancestors.
...
+				Mkparents: dashp,
```
Stebalien: I'd just use the `true` constant. Easier to follow.

achingbrain:
Thanks for the review - I was concerned that adding the mkdir type behaviour to `getFileHandle` might make it responsible for a little too much so split it out into a separate function.  Let me know if you'd like any further changes made.

core/commands/files.go
```diff
@@ -757,6 +759,14 @@ stat' on the file or any of its ancestors.
 			return
 		}
 
+		if mkParents {
+			err := ensureContainingDirectoryExists(nd.FilesRoot, path, flush, prefix)
```
Stebalien: No point in flushing here. We'll flush when we actually create the file.

achingbrain: Ok, I've changed it to not pass a `Flush` option to `mfs.Mkdir`

Stebalien:
@achingbrain mind rebasing?

Stebalien:
I'd like to get a second signoff from either @schomatis or @kevina and then I'll merge.

test/sharness/t0250-files-api.sh
```diff
@@ -597,9 +597,24 @@ test_files_api() {
...
+
+  test_expect_success "can write file and create intermediate directories $EXTRA" '
+    echo "ipfs rocks" | ipfs files write --create --parents /parents/foo/bar/baz/ipfs.txt &&
+    ipfs files stat "/parents/foo/bar/baz/ipfs.txt" | grep -q "^Type: file"
+  '
+
+  test_expect_success "can write file and create intermediate directories with short flags $EXTRA" '
+    echo "ipfs rocks" | ipfs files write -e -p /parents/foo/bar/baz/qux/quux/garply/ipfs.txt &&
+    ipfs files stat "/parents/foo/bar/baz/qux/quux/garply/ipfs.txt" | grep -q "^Type: file"
+  '
```
kevina: Please add another test something like:
```
  test_expect_success "can write another file in the same directory with -e -p $EXTRA" '
    echo "ipfs rocks" | ipfs files write -e -p /parents/foo/bar/baz/qux/quux/garply/ipfs2.txt &&
    ipfs files stat "/parents/foo/bar/baz/qux/quux/garply/ipfs2.txt" | grep -q "^Type: file"
  '
```
To make sure it won't error if you use `-p` (or `--parents`) and all the parent directory already exists.

achingbrain: I've added the extra test.

kevina: Okay.  Assuming the test passes this LGTM.

achingbrain: It worked locally, hopefully CI will agree.

test/sharness/t0250-files-api.sh
```diff
@@ -597,9 +597,29 @@ test_files_api() {
     ipfs files ls /adir | grep foobar
   '
 
+  test_expect_failure "should fail to write file and create intermediate directories with no --parents flag set $EXTRA" '
+    echo "ipfs rocks" | ipfs files write --create /parents/foo/ipfs.txt
+  '
```
kevina: This is not the correct way to do this.  Use `test_must_fail`.  Some thing like:
```
test_expect_success "should fail to write file and create intermediate directories with no --parents flag set $EXTRA" '
  echo "ipfs rocks" | test_must_fail ipfs files write --create /parents/foo/ipfs.txt
'
```
`test_expect_failure` is for broken test that need eventually be fixed.

achingbrain: Fixed.

Stebalien:
@achingbrain ys.

PR #5361, opened 2018-08-09
fix/mfs mv bug

### LINKED ISSUE #5346: MFS `Mv` is buggy
1. It needs to *only* use the target file/directory to determine if the target is a file or a directory instead of looking at the path. Currently, it does both. This means that, e.g., `ipfs files mv /a /foo` where `/foo` is a directory will create a file named `/foo/foo`.
2. After determining that the target is a directory, it needs to lookup the file under that directory *again*. Otherwise, `ipfs files mv /a /b/` won't work when `/b/a` exists.

grokcoder:
fix #5346

### DIFF
--- a/mfs/ops.go
+++ b/mfs/ops.go
@@ -14,10 +14,11 @@ import (
 )
 
 var (
-	errInvalidNodeType = errors.New("invalid argument")
-	errMvDirToFile     = errors.New("can not move directory to file path")
-	errMvParentDir     = errors.New("can not move parent directory to sub directory")
-	errInvalidDirPath  = errors.New("path end with '/' is not a directory")
+	errInvalidNodeType            = errors.New("invalid argument")
+	errMvDirToFile                = errors.New("can not move directory to file path")
+	errMvParentDir                = errors.New("can not move parent directory to sub directory")
+	errInvalidDirPath             = errors.New("path end with '/' is not a directory")
+	errUnableCreateNonExistedFile = errors.New("unable to create un-existed file")
 )
 
 // Mv moves the file or directory at 'src' to 'dst'
@@ -26,12 +27,12 @@ func Mv(r *Root, src, dst string) error {
 		return errMvParentDir
 	}
 
-	src = strings.TrimRight(src, "/")
-	if strings.HasPrefix(dst, src) {
+	trimmedSrc := strings.TrimRight(src, "/")
+	if strings.HasPrefix(dst, trimmedSrc) {
 		return errMvParentDir
 	}
 
-	srcNode, err := DirLookup(r.GetDirectory(), src)
+	srcNode, err := DirLookup(r.GetDirectory(), trimmedSrc)
 	if err != nil {
 		return err
 	}
@@ -58,6 +59,9 @@ func moveDir(r *Root, src, dst string) error {
 
 	dstNode, err := DirLookup(r.GetDirectory(), dst)
 	if err != nil {
+		if err == os.ErrNotExist {
+			return handleDstFileUnExisted(r, src, dst)
+		}
 		return err
 	}
 
@@ -100,6 +104,9 @@ func moveFile(r *Root, src, dst string) error {
 
 	dstNode, err := DirLookup(r.GetDirectory(), dst)
 	if err != nil {
+		if err == os.ErrNotExist {
+			return handleDstFileUnExisted(r, src, dst)
+		}
 		return err
 	}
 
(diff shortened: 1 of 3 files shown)


mfs/ops.go
```diff
@@ -46,21 +34,42 @@ func Mv(r *Root, src, dst string) error {
...
+		dstDir, err = lookupDir(r, dstDirStr)
+		if err != nil {
+			return err
+		}
+
+		fsn, err := dstDir.Child(dirFileName)
+		if err == nil {
+			switch n := fsn.(type) {
+			case *File:
+				return fmt.Errorf("%s : is not a directory", dst)
```
Stebalien: Again, we should overwrite in this case (if the source is also a file).

mfs/ops.go
```diff
@@ -46,21 +34,42 @@ func Mv(r *Root, src, dst string) error {
...
+			case *Directory:
+				dstDir = n
```
Stebalien: Need to check if there exists a matching file in this directory (replacing it if both the source and dest are files or the source is a directory and the dest is an empty directory).

grokcoder: got it, so the `mv` should works as follows ?
```
if src.Type == dir:
     if dest.Type == dir: 
            if desc.Contains(src): case an error
            else move src to dest/
     else if dest.Type == file: case a error
else if src.Type == file:
     if dest.Type == file: replace(src, dest)
     else if desc.Type == directory: 
              if desc.Contains(src): case an error
              else move src to desc/
```

Stebalien: Ok, so, ignore what I said about "POSIX". Let's just do what GNU mv does because nobody cares about POSIX (I was about to quote the manual at you and then I actually read the manual; I don't think the GNU people and the POSIX people talk much).

Three modifications:

1. Replacing empty directories is fine.
2. If dest ends in `/`, assert that it is a directory.
3. If src ends in `/`, assert that it is a directory.

```
if src.Type == dir:
     if dest.Type == dir: 
            if dest.Contains(src):
                if dest.Empty(): unlink
                else: return an error
            move src to dest/
     else if dest.Type == file: case a error
else if src.Type == file:
     if dest.Type == file: replace(src, dest)
     else if desc.Type == directory: 
              if desc.Contains(src):
                  if desc.Type == dir and not desc.Empty():
                      return an error
                  unlink src from desc
              move src to desc/
```

(I'm missing the asserts in that pseudo code)

---

Basically, this command should do what the user expects from using linux commands.

Stebalien: Eventually, we should probably add the `-T`/`--no-target-directory` flags from Linux for consistency.

The other way to have gone would have been to make everything explicit up-front (mv moves a file to dest exactly, none of this target directory nonsense). However, that'll *really* confuse users and would be a breaking change so... not going to happen.

grokcoder:
@Stebalien would you mind to review it again, if the source path is the parent path of dst will case an error.  As for assert while addr ends with "/", what's that mean, print assertion msg?

Stebalien:
>  As for assert while addr ends with "/", what's that mean, print assertion msg?

I mean that if the one of the paths path ends in `/` but the "thing" (if it exists) at that path isn't a directory, return an error.

mfs/ops.go
```diff
@@ -46,21 +34,42 @@ func Mv(r *Root, src, dst string) error {
...
+	var dstDir *Directory
+	if dst[len(dst)-1] == '/' {
+		dstDirStr = dst
+		dstDir, err = lookupDir(r, dstDirStr)
+		if err != nil {
+			return err
+		}
+		_, err = dstDir.Child(srcFname)
+		if err == nil {
+			return os.ErrExist
```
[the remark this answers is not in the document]
grokcoder: fixed

mfs/ops.go
grokcoder: fixed

grokcoder:
@Stebalien would you mind to review again?

mfs/ops.go
```diff
@@ -12,60 +13,175 @@ import (
...
 
-	var dstDirStr string
-	var filename string
-	if dst[len(dst)-1] == '/' {
-		dstDirStr = dst
-		filename = srcFname
+	if srcNode.Type() == TDir {
+		return moveDir(r, src, dst)
+	} else if srcNode.Type() == TFile {
+		if src[len(src)-1] == '/' {
```
Stebalien: The trailing `/` gets trimmed above. We should probably also have a test for this.

grokcoder: fixed

mfs/ops.go
```diff
@@ -12,60 +13,207 @@ import (
...
+	if src == "/" {
+		return errMvParentDir
+	}
 
-	var dstDirStr string
-	var filename string
-	if dst[len(dst)-1] == '/' {
-		dstDirStr = dst
-		filename = srcFname
+	trimmedSrc := strings.TrimRight(src, "/")
```
Stebalien: Is this to catch `mv /a/b/c/ /a/b/c`? If so, I'd do that separately and return a different error (to avoid any confusion).

mfs/ops.go
```diff
@@ -12,60 +13,207 @@ import (
...
+		}
+
+		return srcParDir.Unlink(srcFileName)
+	} else {
+		return errInvalidNodeType
+	}
+}
+
+func handleDstFileUnExisted(r *Root, src, dst string) error {
+	err := Mkdir(r, dst, MkdirOpts{Mkparents: true, Flush: true})
```
Stebalien: Why?

mfs/ops.go
```diff
@@ -12,60 +13,207 @@ import (
...
+	if err != nil {
+		return errUnableCreateNonExistedFile
+	}
+	dstFileName, _, dstParDir, err := getNodeAndParent(r, dst)
+	if err != nil {
+		return err
+	}
+	if err := dstParDir.Unlink(dstFileName); err != nil {
```
Stebalien: You're creating a directory and then deleting it?

grokcoder: Create empty target and then replace it

mfs/ops.go
Stebalien: Ah. No, we shouldn't create *parent* directories.

schomatis:
Hey @grokcoder, are you still interested in continuing with this PR? The code has since been moved to https://github.com/ipfs/go-mfs/pull/19, if not I'll be closing this PR and continue with that one.

grokcoder:
@schomatis please close it!

schomatis:
Continued in https://github.com/ipfs/go-mfs/pull/19.

PR #5364, opened 2018-08-10
add correct test for 1515

### LINKED ISSUE #1515: ipfs repo gc does not fully reverse ipfs add
```
. lib/test-lib.sh
random 100000 41 >gc &&
du -sb "$IPFS_PATH/blocks" | awk "{print \$1}" >expected &&
hash=`ipfs add -q gc` &&
ipfs pin rm -r $hash &&
ipfs repo gc &&
du -sb "$IPFS_PATH/blocks" | awk "{print \$1}" >actual &&
test_cmp expected actual
```

will fail with 4kB or 16 kB of residue.

Stebalien:
When checking to see if GC fully reverses an `ipfs add`, we should check the size of the actual files, not the directory sizes. A bunch of empty directories won't use *that* much space and really shouldn't count against GC.

closes #1515

I'm not sure why this has been sitting around for so long but I never seem to have created this PR.

### DIFF
--- a/test/sharness/lib/test-lib.sh
+++ b/test/sharness/lib/test-lib.sh
@@ -394,6 +394,27 @@ file_size() {
     $_STAT "$1"
 }
 
+directory_size() {
+    local total=0
+    local fsize=0
+    local res=0
+    find "$1" -type f | ( while read fname; do
+        fsize=$(file_size "$fname")
+        res=$?
+        if ! test $? -eq 0; then
+            echo "failed to get filesize" >&2
+            return $res
+        fi
+        total=$(expr "$total" + "$fsize")
+        res=$?
+        if ! test $? -eq 0; then
+            echo "filesize not a number: $fsize" >&2
+            return $res
+        fi
+    done
+    echo "$total" ) # do not remove this subshell
+}
+
 test_check_peerid() {
   peeridlen=$(echo "$1" | tr -dC "[:alnum:]" | wc -c | tr -d " ") &&
   test "$peeridlen" = "46" || {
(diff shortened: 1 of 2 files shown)


test/sharness/lib/test-lib.sh
```diff
@@ -394,6 +394,27 @@ file_size() {
     $_STAT "$1"
 }
 
+directory_size() {
```
magik6k: Why not just something like `du -sb --apparent-size`?

Stebalien: 1. Didn't know about it.
2. MacOS :sad:.

test/sharness/lib/test-lib.sh
```diff
@@ -409,3 +430,9 @@ convert_tcp_maddr() {
 port_from_maddr() {
   echo $1 | awk -F'/' '{ print $NF }'
 }
+
+
+fail() {
```
magik6k: There is `test_fsh` - https://github.com/ipfs/go-ipfs/blob/1054826ac49f66d329d3088143f1a7f2b9930dcc/test/ipfs-test-lib.sh#L22-L27 and it's used all over sharness tests - like `test_cmp expected unique_hash || test_fsh cat add_output`

Stebalien: Ah nice!

a4d442663580:
> A bunch of empty directories won't use that much space and really shouldn't count against GC.

Then the process is still irreversible. Accumulating empty dirs doesn't sound sustainable either. While the effect is negligible for embedded machines, this is not necessarily the case for clusters.

Stebalien:
It's bounded to 1024 empty directories. Whenever we store an object, we store it at `/xx/YYYY...xx`. However, when we delete the object, we don't delete the `/xx/` directory. Given that `xx` is base32, that gives us at most `32^2 = 1024` directories (steady state). An alternative is to just create them all up-front.

PR #5372, opened 2018-08-11
Run DNS lookup for --api endpoint provided in CLI

### LINKED ISSUE #5249: `--api` should accept dns addresses
#### Version information:
go-ipfs version: 0.4.16-92ac43a
Repo version: 7
System version: amd64/linux
Golang version: go1.10.3

#### Type:
Enhancement

#### Description:
I am running ipfs daemon on a home kubernetes cluster.  The daemon runs fine.  I would like to talk to the daemon's API port from another kubernetes pod, but the `ipfs --api` arg seems to only accept IP address literals (afaict?).  In Kubernetes (and many other containerised) environments, the ipfs daemon's IP address is allocated dynamically, and can change on restarts.  Services are usually found via DNS hostnames.

TL;DR: I want something like this to just work: `ipfs --api /dnsaddr/api.ipfs.svc/tcp/5001`

Right now that gives "Error: /dnsaddr/api.ipfs.svc/tcp/5001 is not a 'thin waist' address"

As far as I know, my only option right now is to wrap `ipfs` in some shell script that does a dns lookup and builds a `/ip4/...` address on the fly :(

raulk:
_This is my first PR to go-ipfs, so please be gentle ;-)_

Imports `go-multiaddr-dns` and calls `madns.Resolve()` with the provided multiaddr. There's no need to call `Matches()` explicitly, as the resolution will do no-op if no DNS protocols are found in the input multiaddr (I'm assuming this is part of the contract of `madns.Resolve()`).

If multiple results are returned, only the first one is used. See https://github.com/ipfs/go-ipfs/issues/5249#issuecomment-412286842.

I've added several unit tests in the `dnsresolve_test.go` file. Not sure if that's the best place. I'm happy to change it.

I initially added the call to `madns.Resolve()` in `apiClientForAddr()` but I couldn't assert on the address of `http.Client` as its private, so I had to put it in a separate function to enable unit testing.

Fixes #5249.

### DIFF
--- a/cmd/ipfs/dnsresolve_test.go
+++ b/cmd/ipfs/dnsresolve_test.go
@@ -1,0 +1,72 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"strings"
+	"testing"
+
+	ma "gx/ipfs/QmYmsdtJ3HsodkePE3eU3TsCaP2YvPZJ4LoXnNkDE5Tpt7/go-multiaddr"
+	madns "gx/ipfs/QmfXU2MhWoegxHoeMd3A2ytL2P6CY4FfqGWc23LTNWBwZt/go-multiaddr-dns"
+)
+
+var (
+	ctx         = context.Background()
+	testAddr, _ = ma.NewMultiaddr("/dns4/example.com/tcp/5001")
+)
+
+func makeResolver(n uint8) *madns.Resolver {
+	results := make([]net.IPAddr, n)
+	for i := uint8(0); i < n; i++ {
+		results[i] = net.IPAddr{IP: net.ParseIP(fmt.Sprintf("192.0.2.%d", i))}
+	}
+
+	backend := &madns.MockBackend{
+		IP: map[string][]net.IPAddr{
+			"example.com": results,
+		}}
+
+	return &madns.Resolver{
+		Backend: backend,
+	}
+}
+
+func TestApiEndpointResolveDNSOneResult(t *testing.T) {
+	dnsResolver = makeResolver(1)
+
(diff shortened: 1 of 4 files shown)


Stebalien:
Nice!  We usually don't bother unit testing small helper functions like that but that's just because we're horrible, lazy coders so I definitely won't say no to it.

Mind adding an integration test? You can find the others in `test/sharness`. Basically, all we need to do is start an ipfs daemon `test_init_ipfs` and `test_launch_ipfs_daemon, test that `ipfs --api=/dns4/localhost/tcp/5001 SOMETHING` works, and then stop the daemon with `test_kill_ipfs_daemon`.

raulk:
@Stebalien – added a sharness test, modelled after t0235. Only testing the appearance of the POST, which is enough to assert that the DNS resolution is happening. Had to use an existing domain that resolves to `127.0.0.1`, [localtest.me](readme.localtest.me), as the kernel resolves `localhost` through `/etc/hosts`, not a DNS lookup.

Stebalien:
> Had to use an existing domain that resolves to 127.0.0.1, localtest.me, as the kernel resolves localhost through /etc/hosts, not a DNS lookup.

Actually, I'd much rather use localhost. That way the tests pass while offline (e.g., when on a plane). Really, we don't care about actually *using* DNS, we just want to hit the multiaddr-dns resolver.

raulk:
Thanks @Stebalien, somehow I had missed this. I had wrongly assumed that `net.Resolver.LookupIPAddr()` behaved like `nslookup`, which does not resolve `localhost` because it goes straight to the nameservers. However, `LookupIPAddr()` does resolve `localhost` properly as it goes via the kernel.

All good, thanks 👍

PR #5373, opened 2018-08-12
gossipsub

vyzo:
Adds the option to use gossipsub as the pubsub protocol.

This is handled by adding an `Pubusub.Router` config option, which allows the user to specify the pubsub router to use. 
The possible values are
- `floodsub` (or empty), for the default floodsub protocol
- `gossipsub` for the gossipsub protocol

Requires:
- [X] https://github.com/libp2p/go-libp2p-pubsub-router/pull/4
- [X] https://github.com/ipfs/go-ipfs-config/pull/3

Blocked:
- [x] #5375


vyzo:
Ugh, it seems that the configuration repo has been extracted from go-ipfs in the last day, which breaks my patch.
The option will have to be added to the extracted package ...

vyzo:
Updated `go-ipfs-config` to include the necessary option and rebased.

package.json
```diff
@@ -524,9 +524,9 @@
     },
     {
       "author": "stebalien",
-      "hash": "QmZXeeaAaSUCCizUfvZHK4BM8po7xJHFGodfYAQuDfHvwt",
+      "hash": "QmQYkj73ieX5qYTvY2H2fLhfh48FtUZhBcQVJRYgd3ZRnh",
       "name": "go-libp2p-pubsub-router",
-      "version": "0.3.9"
+      "version": "0.3.10"
```
Stebalien: FYI, I had to re-publish so this'll have to be re-updated.

test/sharness/t0180-pubsub-gossipsub.sh
```diff
@@ -0,0 +1,100 @@
...
+
+# start iptb + wait for peering
+NUM_NODES=5
+test_expect_success 'init iptb' '
+  iptb init -n $NUM_NODES --bootstrap=none --port=0
+'
+
+test_expect_success "enable gossipsub" '
+  for x in $(seq 0 4); do
+    ipfsi 0 config Pubsub.Router gossipsub
```
Stebalien: Should be `ipfsi $x ...`. FYI, you can also use `iptb for-each ipfs config ...`.

Stebalien: It's a bit odd that the tests are still passing.

vyzo: oops, good catch!
It's not odd that the test still pass, gossipsub is backwards compatible with floodsub.

package.json
vyzo: ok, will do.

vyzo: Unfortunately this pulled a bunch of new deps, so lots of things need to be updated recursively... sigh

vyzo: Ok, this update touches too many things -- we'll have to wait for #5375 I think, as there are also code changes involved.

vyzo: I'll wait to update `go-libp2p-pubsub-router` until that pr has been merged.

vyzo:
We are blocked on #5375 as it interleaves a big update with a competing update to `go-libp2p-pubsub-router`.

We will have to wait for those updates to get merged (and rebase) before we can proceed, as it's out of scope to  include a cascade of irrelevant package updates in this pr.

Stebalien:
@vyzo update merged.

test/sharness/t0180-pubsub-gossipsub.sh
Stebalien: Ah. Good to know.

vyzo:
rebased.

vyzo:
hrm, some merge errors -- fixing.

vyzo:
should be ok now.

Stebalien:
Rebased.

test/sharness/t0180-pubsub-gossipsub.sh
```diff
@@ -0,0 +1,100 @@
...
+'
+
+test_expect_success "enable gossipsub" '
+  for x in $(seq 0 4); do
+    ipfsi $x config Pubsub.Router gossipsub
+  done
+'
+
+# this is just a copy of t0180-pubsub; smell.
+startup_cluster $NUM_NODES --enable-pubsub-experiment
```
magik6k: You should be able to wrap this into a function in the pubsub test and call it with different setup, like https://github.com/ipfs/go-ipfs/blob/master/test/sharness/t0170-dht.sh#L10

docs/experimental-features.md
```diff
@@ -43,6 +43,11 @@ experimental, default-disabled.
 run your daemon with the `--enable-pubsub-experiment` flag. Then use the
 `ipfs pubsub` commands.
 
+### gossipsub
+
+You can enable the new experimental gossipsub protocol via configuration:
```
magik6k: I'd add a short line explaining the difference between the default floodsub and gossipsub.

vyzo: hrm, maybe, although that's a topic in itself. there will be a gossipsub spec soon [TM] and perhaps we could link to that instead.

Stebalien: For now, I'd just go with:

>  Gossipsub is a new, experimental routing protocol for pubsub that should waste less bandwidth than the current default routing protocol (floodsub). It's backwards compatible with floodsub so enabling this feature shouldn't break compatibility with existing IPFS nodes.

That should tell users everything they need to know (for now).

test/sharness/t0180-pubsub-gossipsub.sh
Stebalien: Up to you. I'm usually not a stickler on DRYing up tests, especially integration tests. We should *never* have to change these anyways ("don't break userspace").

vyzo: so if anyone has strong feelings about it, i can merge into a single test -- but i am otherwise inclined to keep it as is.

docs/experimental-features.md
vyzo: done.

PR #5374, opened 2018-08-12
ci: call cleanWs after each step

magik6k:
Might make the builds a bit slower until optimized gx is used here, but the ci should at least be more stable.


Stebalien:
Why has tagged a new gx release so I'll try doing a dist build today.

PR #5382, opened 2018-08-15
gx: update go-cid

Stebalien:
Includes the new StreamingSet helper.


Stebalien:
(just a gx update, might as well merge).

PR #5385, opened 2018-08-15
Provide new "cid" sub-command.

### LINKED ISSUE #5229: Tool for converting hashes
Create a tool / subcommand within go-ipfs for converting between cidv0 and cidv1, and different bases (base58, base32, etc).

https://github.com/ipfs/ipfs/issues/337

@kevina

### LINKED ISSUE #5357: RFC: New sub-command `ipfs cid-util`
I propose we create a new sub-command to aid in working with Cids.  I propose the following sub-commands under it:

 * `ipfs cid-util format` - format a Cid in various useful ways
   * `-f <string>`: format string from `cid-fmt` utility
   * `-v 0|1`: cid version
   * `-b <multibase code>`
 * `ipfs cid-util base32` - shortcut for `ipfs cid-util format -v 1 -b base32 ...` to aid in our translation to making base32 CIDv1 the default
 * `ipfs cid-util encoding` - list available multibase encoding 
   * `--prefixes` - also include the single leter prefixes in addition to the code
   * `--numeric` - also include the numeric code (ascii value of prefix)
 * `ipfs cid-util codecs` - list available CID codecs
   * `--numeric` - also include the numeric code 
 * `ipfs cid-util hashes` - list valid multihash codes (excludes insecure hashes)
   * `--numeric` - also include the numeric code 

This could be a separate command, however I think it is more convenient as a sub-command.  For one thing the `ipfs` binary is guaranteed to be available while extra utilities might not be installed.  In addition the `encoding`, `codecs` and `hashes` will be guaranteed to correspond to what is support by go ipfs.  If they are in a separate command this is not guaranteed.

Thoughts?

Closes #5229.

kevina:
Implements #5357.

~Still needs tests.~

Closes #5229. Closes #5357.

### DIFF
--- a/core/commands/cid.go
+++ b/core/commands/cid.go
@@ -36,14 +36,14 @@ var cidFmtCmd = &cmds.Command{
 		LongDescription: `
 Format and converts <cid>'s in various useful ways.
 
-The optional format string either "prefix" or or a printf style format string:
+The optional format string is a printf style format string:
 ` + cidutil.FormatRef,
 	},
 	Arguments: []cmdkit.Argument{
 		cmdkit.StringArg("cid", true, true, "Cids to format."),
 	},
 	Options: []cmdkit.Option{
-		cmdkit.StringOption("f", "Format string."),
+		cmdkit.StringOption("f", "Printf style format string.").WithDefault("%s"),
 		cmdkit.StringOption("v", "CID version to convert to."),
 		cmdkit.StringOption("b", "Multibase to display CID in."),
 	},
@@ -54,17 +54,10 @@ The optional format string either "prefix" or or a printf style format string:
 
 		opts := cidFormatOpts{}
 
-		switch fmtStr {
-		case "":
-			opts.fmtStr = "%s"
-		case "prefix":
-			opts.fmtStr = "%P"
-		default:
-			if strings.IndexByte(fmtStr, '%') == -1 {
-				return fmt.Errorf("invalid format string: %s", fmtStr)
-			}
-			opts.fmtStr = fmtStr
+		if strings.IndexByte(fmtStr, '%') == -1 {
+			return fmt.Errorf("invalid format string: %s", fmtStr)
 		}
+		opts.fmtStr = fmtStr
 
 		switch verStr {
 		case "":
--- a/core/commands/cid.go
+++ b/core/commands/cid.go
@@ -80,22 +80,25 @@ The optional format string is a printf style format string:
 			opts.newBase = mbase.Encoding(-1)
 		}
 
-		res, err := formatCids(req.Arguments, opts)
-		if err != nil {
-			return err
-		}
-		cmds.EmitOnce(resp, res)
-		return nil
+		return emitCids(req, resp, opts)
 	},
-	Encoders: cmds.EncoderMap{
-		cmds.Text: cmds.MakeEncoder(func(req *cmds.Request, w io.Writer, val interface{}) error {
-			for _, v := range val.([]string) {
-				fmt.Fprintf(w, "%s\n", v)
+	PostRun: cmds.PostRunMap{
+		cmds.CLI: streamRes(func(v interface{}, out io.Writer) nonFatalError {
+			r := v.(*CidFormatRes)
+			if r.ErrorMsg != "" {
+				return nonFatalError(fmt.Sprintf("%s: %s", r.CidStr, r.ErrorMsg))
 			}
-			return nil
+			fmt.Fprintf(out, "%s\n", r.Formatted)
+			return ""
 		}),
 	},
-	Type: []string{},
+	Type: CidFormatRes{},
+}
+
+type CidFormatRes struct {
+	CidStr    string // Original Cid String passed in
+	Formatted string // Formated Result
+	ErrorMsg  string // Error
 }
 
 var base32Cmd = &cmds.Command{
@@ -103,7 +106,7 @@ var base32Cmd = &cmds.Command{
 		Tagline: "Convert CIDs to Base32 CID version 1.",
 	},
 	Arguments: []cmdkit.Argument{
-		cmdkit.StringArg("cid", true, true, "Cids to convert."),
+		cmdkit.StringArg("cid", true, true, "Cids to convert.").EnableStdin(),
 	},
 	Run: func(req *cmds.Request, resp cmds.ResponseEmitter, env cmds.Environment) error {
 		opts := cidFormatOpts{
@@ -111,16 +114,10 @@ var base32Cmd = &cmds.Command{
 			newBase: mbase.Encoding(mbase.Base32),
 			verConv: toCidV1,
 		}
-		res, err := formatCids(req.Arguments, opts)
-		if err != nil {
-			return err
-		}
-
-		cmds.EmitOnce(resp, res)
-		return nil
+		return emitCids(req, resp, opts)
 	},
-	Encoders: cidFmtCmd.Encoders,
-	Type:     cidFmtCmd.Type,
+	PostRun: cidFmtCmd.PostRun,
+	Type:    cidFmtCmd.Type,
 }
 
 type cidFormatOpts struct {
@@ -129,12 +126,19 @@ type cidFormatOpts struct {
 	verConv func(cid cid.Cid) (cid.Cid, error)
 }
 
-func formatCids(args []string, opts cidFormatOpts) ([]string, error) {
-	var res []string
-	for _, cidStr := range args {
+func emitCids(req *cmds.Request, resp cmds.ResponseEmitter, opts cidFormatOpts) error {
+	for _, cidStr := range req.Arguments {
+		emit := func(fmtd string, err error) {
+			res := &CidFormatRes{CidStr: cidStr, Formatted: fmtd}
+			if err != nil {
+				res.ErrorMsg = err.Error()
+			}
+			resp.Emit(res)
+		}
 		c, err := cid.Decode(cidStr)
 		if err != nil {
-			return nil, fmt.Errorf("%s: %v", cidStr, err)
+			emit("", err)
+			continue
 		}
 		base := opts.newBase
 		if base == -1 {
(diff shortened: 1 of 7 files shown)


core/commands/cid.go
```diff
@@ -0,0 +1,307 @@
...
+	Helptext: cmdkit.HelpText{
+		Tagline: "Format and convert a CID in various useful ways.",
+		LongDescription: `
+Format and converts <cid>'s in various useful ways.
+
+The optional format string either "prefix" or or a printf style format string:
+` + cidutil.FormatRef,
+	},
+	Arguments: []cmdkit.Argument{
+		cmdkit.StringArg("cid", true, true, "Cids to format."),
```
Stebalien: Can we use `WithDefault("%s")`?

core/commands/cid.go
```diff
@@ -0,0 +1,307 @@
...
+		verStr, _ := req.Options["v"].(string)
+		baseStr, _ := req.Options["b"].(string)
+
+		opts := cidFormatOpts{}
+
+		switch fmtStr {
+		case "":
+			opts.fmtStr = "%s"
+		case "prefix":
+			opts.fmtStr = "%P"
```
Stebalien: This feels a bit magical. What's the motivation? We could also just add a `ipfs cid prefix` command (alias of `ipfs cid format -f '%P' ...`).

core/commands/cid.go
```diff
@@ -0,0 +1,307 @@
...
+			if err != nil {
+				resp.SetError(err, cmdkit.ErrNormal)
+				return
+			}
+			opts.newBase = encoder.Encoding()
+		} else {
+			opts.newBase = mbase.Encoding(-1)
+		}
+
+		res, err := formatCids(req.Arguments, opts)
```
Stebalien: I'd stream these results back. Users may want to feed quite a few CIDs into this command.

Stebalien:
@alanshaw this adds a new `ipfs cid` command. I'm fine with keeping this command "go specific" (it's really just a utility command) but thought you might have some input.

Basically, it makes it easy to inspect CIDs and convert between various formats (bases, versions, etc.).

core/commands/cid.go
kevina: You meant for the `-f` option?  If it is omitted that is the default (see below), I just didn't use WithDefault.  The idea was the `-f` is optional and if its not specified it will just return the CIDs.

core/commands/cid.go
kevina: It was there in the original `cid-fmt` utility and copied over here.  For the `cid-fmt` utility is was the original use case (to get information about cids).  I am not tied to keeping it in this utility.

core/commands/cid.go
kevina: I thought about that but didn't see the benefit.  This is a client side utility that will never block and is should be very fast, so the only benefit will be memory usage.  In addition the number of cid's passed in is limited by the length of command line so memory usage will never get out of control.

core/commands/cid.go
Stebalien: It's nice from a documentation standpoint.

core/commands/cid.go
Stebalien: I'd remove it for now. We can consider either a `prefix` or a `show` command later but we probably don't *need* them now.

core/commands/cid.go
kevina: No Problem.

core/commands/cid.go
kevina: Will do.

core/commands/cid.go
Stebalien: Ah. Well, actually, arguments can be passed in on stdin (newline separated)... That's why I brought this up. Basically, a user may run something like `ipfs refs -u -r QmThing | ipfs cid format -f '%c' | sort -u`.

kevina: I think you have to do something special to make that work.  Because when I tried it I get `Error: argument "cid" is required`.  I actually think this it is a bit of a misfeature to take the arguments from stdin without having to pass in a special flag, but that something to debate later.

warpfork: Fwiw, I agree it's a misfeature to take arguments from stdin without a flag -- I think it's pretty widespread convention now to use a single dash "`-`" argument to mean "read stdin instead of file".  (And that would make just as much sense for "read stdin instead of literal".)

core/commands/cid.go
```diff
@@ -0,0 +1,314 @@
...
+			str, err := cidutil.Format(opts.fmtStr, base, c)
+			if _, ok := err.(cidutil.FormatStringError); ok {
+				// no point in continuing if there is a problem with the format string
+				resp.SetError(err, cmdkit.ErrNormal)
+				return
+			}
+			emit(str, err)
+		}
+	}()
+	resp.Emit(out)
```
Stebalien: You can just call Emit on each CID individually (no need for the goroutine or the channel).

Stebalien:
EnableStdin isn't broken, it's just that:

1. The documentation is all missing.
2. The API sucks.

Basically, you'll have to:

1. Iterate over all the arguments in `Arguments`.
2. Iterate over the arguments from `BodyArgs()`.

You can also just call `ParseBodyArgs()` to move all the arguments from the body into `Arguments`.

kevina:
> You can also just call ParseBodyArgs() to move all the arguments from the body into Arguments.

That will prevent streaming.

Stebalien:
> That will prevent streaming.

It will. The important part here is *allowing* ourselves to stream at some point in the future without breaking the API.

core/commands/cid.go
```diff
@@ -0,0 +1,336 @@
...
+		for code, name := range mbase.EncodingToStr {
+			res = append(res, CodeAndName{int(code), name})
+		}
+		cmds.EmitOnce(resp, res)
+	},
+	Encoders: cmds.EncoderMap{
+		cmds.Text: cmds.MakeEncoder(func(req *cmds.Request, w io.Writer, val0 interface{}) error {
+			prefixes, _ := req.Options["prefix"].(bool)
+			numeric, _ := req.Options["numeric"].(bool)
+			val := val0.([]CodeAndName)
```
Stebalien: For now, you'll need to check if the type is correct. See: https://github.com/ipfs/go-ipfs-cmds/issues/65. Basically, we end up feeding errors to this function as well.

(search the code for examples on how to do this).

core/commands/commands.go
```diff
@@ -152,3 +153,46 @@ func unwrapOutput(i interface{}) (interface{}, error) {
 
 	return <-ch, nil
 }
+
+type nonFatalError string
+
+// streamRes is a helper function to stream results, that possibly
+// contain with non-fatal, the helper function is allowed to panic on
+// internal errors
```
Stebalien: Needs some proof-reading :).

core/commands/commands.go
```diff
@@ -152,3 +153,46 @@ func unwrapOutput(i interface{}) (interface{}, error) {
...
+func streamRes(procVal func(interface{}, io.Writer) nonFatalError) func(*cmds.Request, cmds.ResponseEmitter) cmds.ResponseEmitter {
```
Stebalien: Can we call this `streamResults`? We tend to a bit *too* curt.

core/commands/commands.go
```diff
@@ -152,3 +153,46 @@ func unwrapOutput(i interface{}) (interface{}, error) {
 
 	return <-ch, nil
 }
+
+type nonFatalError string
```
Stebalien: Why can't we just use an `error`?

core/commands/cid.go
```diff
@@ -0,0 +1,336 @@
...
+}
+
+func emitCids(req *cmds.Request, resp cmds.ResponseEmitter, opts cidFormatOpts) {
+	itr := argumentIterator{req.Arguments, req.BodyArgs()}
+	for {
+		cidStr, ok := itr.next()
+		if !ok {
+			break
+		}
+		emit := func(fmtd string, err error) {
```
Stebalien: Let's lift this out of the loop (I don't trust go).

core/commands/cid.go
```diff
@@ -0,0 +1,336 @@
...
+			res := &CidFormatRes{CidStr: cidStr, Formatted: fmtd}
+			if err != nil {
+				res.ErrorMsg = err.Error()
+			}
+			resp.Emit(res)
```
Stebalien: We should handle this error (the client may have gone away).

kevina:
> Is there any way to simplify this? We're introducing multiple new abstractions.

Most (all) of the abstractions are lifted from the `filestore ls` code when I am doing nearly the same thing.  Most of the completely comes from the fact that I believe fairly strongly that we should not abort on non-fatal errors, especially when streaming.

kevina:
> Is there any way to simplify this? We're introducing multiple new abstractions.

On second thought.  I am not 100% sure what you mean.  I thought you where talking about the handling of the `cid fotmat` command and streaming?  Is there anything else you are referring to?

core/commands/cid.go
kevina: It's in the loop as it is a closure.  I do it that way to avoid having to pass is the cidStr variable.

core/commands/cid.go
kevina: I'm not sure what I should do.  Should I log the error and then just return?

kevina: I made at attempt at handling this, please check that it is correct.

core/commands/cid.go
kevina: I just eliminated the need for the emit function.

alanshaw:
Hey everyone! I've tried to replicate the functionalty here as closely as possible for the `jsipfs cid` command. I'd appreciate it if you're able to spare a few minutes to look over https://github.com/ipfs-shipyard/js-cid-tool (if not the code, the README will do) to check I've captured the intentions correctly.

thank you ❤️

kevina:
@alanshaw I looked over the README and I think the tool captures the intentions about the same.  It was never really my intention for these commands to become standardized except for perhaps the `base32` command.

Note that some minor details of the behavior of these commands may change over time.

kevina:
@Stebalien if this is too complicated, I can separate out the part that implements proper streaming into a separate p.r. as I think that is where most of the complexity your seeing is coming from...

alanshaw:
Thanks @kevina 

> It was never really my intention for these commands to become standardized

Understood, I haven't added this to any `interface-*` repos. I'm happy for this to just be a command line tool. I think it's important that `jsipfs` users also have it which is why I created it. Plus it was pretty quick and easy to put together.

Stebalien:
@kevina I was just waiting on *some* second signoff. Now that we have one, this should be good to go. Unfortunately, it'll need a rebase. I'd squash first as doing this commit-by-commit is painful (I tried, but there were significant changes to the commands lib).

kevina:
@Stebalien I rebase later today.

kevina:
@Stebalien actually I already rebased and fixed my code to use the new commands library as indicated in an earlier comment.  I just rebased again and fixed a dep/import problem.

This should be good to go.  There is one failed test in Jenkins, but I don't think that is related.

kevina:
And it looks like rebasing caused GitHub to dismiss the earlier approving reviews...

kevina:
@Stebalien Jenkins is now green.

Stebalien:
> And it looks like rebasing caused GitHub to dismiss the earlier approving reviews...

Yeah, I just turned that off. Sorry.

PR #5392, opened 2018-08-20
name cmd: move subcommands to subdirectory

magik6k:
Just wanted to do this before `ipfs name resolve --stream`

* Moves `GetApi`/`GetNode` utils to subpkg to avoid duplicating them
* Port name commands to the new lib
* Move `ipns name` commands into a sub-package

### DIFF
--- a/core/commands/name/ipns.go
+++ b/core/commands/name/ipns.go
@@ -2,6 +2,7 @@ package name
 
 import (
 	"errors"
+	"fmt"
 	"io"
 	"strings"
 	"time"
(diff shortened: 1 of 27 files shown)


core/commands/name/ipns.go
```diff
@@ -136,29 +145,25 @@ Resolve the value of a dnslink:
...
+	Encoders: cmds.EncoderMap{
+		cmds.Text: cmds.MakeEncoder(func(req *cmds.Request, w io.Writer, v interface{}) error {
 			output, ok := v.(*ResolvedPath)
 			if !ok {
-				return nil, e.TypeErr(output, v)
+				return e.TypeErr(output, v)
 			}
-			return strings.NewReader(output.Path.String() + "\n"), nil
-		},
+			_, err := w.Write([]byte(output.Path.String() + "\n"))
```
Stebalien: Minor nit but we might as well use `fmt.Writeln(w, output.Path.String())`.

core/commands/name/ipnsps.go
```diff
@@ -71,17 +71,18 @@ var ipnspsStateCmd = &cmds.Command{
 				state = "disabled"
 			}
 
-			return strings.NewReader(state + "\n"), nil
-		},
+			_, err := w.Write([]byte(state + "\n"))
```
Stebalien: Ditto.

core/commands/name/ipnsps.go
```diff
@@ -164,7 +160,26 @@ var ipnspsCancelCmd = &cmds.Command{
 				state = "no subscription"
 			}
 
-			return strings.NewReader(state + "\n"), nil
-		},
+			_, err := w.Write([]byte(state + "\n"))
```
Stebalien: Ditto.

Stebalien:
Meh. This is actually just ready to go.

PR #5397, opened 2018-08-23
fix the urlstore tests

Stebalien:
1. GC the intentionally broken files so we don't reuse their blocks.
2. Make sure we actually check the *new* output file instead of an old one.


Stebalien:
@keks this should fix the failing test for you.

PR #5399, opened 2018-08-23
add --name new flag when ipfs adding from stdin 

### LINKED ISSUE #4986: [FR] `ipfs add -w --name` when adding from stdin
#### Version information:
    go-ipfs version: 0.4.13-
    Repo version: 6
    System version: amd64/linux
    Golang version: go1.9.4

#### Type:
Feature Request

#### Description:
`ipfs add` supports `-w` for filename preservation, e.g. `ipfs add -w foo.zip` will let you access the result as `/ipfs/<dir hash>/foo.zip` instead of just `/ipfs/<file hash>`. However, it doesn't work very well when reading data from stdin:

    $ cat test.txt | ipfs add -w
    added Qme9Z5gJLNsQzVELtHh2oQZfjDugHGvLGuNKVtttGAJYAE 
    Qme9Z5gJLNsQzVELtHh2oQZfjDugHGvLGuNKVtttGAJYAE
    added Qmaq9QpAYfjF5sPAeEytKHk9rC3Vcgt6J8PRbCsNk7ypqY

And now test.txt can only be accessed as `/ipfs/Qmaq9QpAYfjF5sPAeEytKHk9rC3Vcgt6J8PRbCsNk7ypqY/Qme9Z5gJLNsQzVELtHh2oQZfjDugHGvLGuNKVtttGAJYAE`

This can be fixed post hoc by creating a new parent object with the right link name (using `ipfs object` or, more conveniently, `ipfs files`), but this is awkward, and (if you need the resulting object pinned) slow.

It would be much more convenient if you could specify a name for the wrapped object on the command line, and have `ipfs add` create the link with that name.

kjzz:
Closes  #4986.

License: MIT


core/coreunix/add.go
```diff
@@ -470,6 +472,9 @@ func (adder *Adder) addFile(file files.File) error {
 		return err
 	}
 
+	if !strings.EqualFold(adder.WpName, "") && adder.Wrap {
```
schomatis: This is the place a normal file is added, not the directory wrapper.

kjzz: i have rebase repo and please help me review this pr @schomatis

kjzz: thank you!

schomatis:
Hey @kjzz, take a look at https://github.com/ipfs/go-ipfs/issues/4986#issuecomment-417656218. We're mixing the name of the directory wrapper with the name of a file coming from `stdin`, which are two different issues.

core/commands/add.go
```diff
@@ -37,6 +37,7 @@ const (
 	progressOptionName    = "progress"
 	trickleOptionName     = "trickle"
 	wrapOptionName        = "wrap-with-directory"
+	wrapPathName          = "name"
```
schomatis: So, as discussed in the issue, this is not exactly related to the `-w` (wrap) flag, is a more generic issue of assigning names to a unnamed file content (such as the one received from `stdin`). Let's change the name of this option to reflect that, e.g., `wrapPathName` -> `pathName`.

core/commands/add.go
```diff
@@ -116,6 +117,7 @@ You can now check what blocks have been created by:
 		cmdkit.BoolOption(trickleOptionName, "t", "Use trickle-dag format for dag generation."),
 		cmdkit.BoolOption(onlyHashOptionName, "n", "Only chunk and hash - do not write to disk."),
 		cmdkit.BoolOption(wrapOptionName, "w", "Wrap files with a directory object."),
+		cmdkit.StringOption(wrapPathName, "Assign path name when use wrap-with-directory option"),
```
schomatis: Similar to the above comment, let's drop the wrap reference and say something like: "Assign a name to an unnamed file" (feel free to reword this).

core/coreunix/add.go
```diff
@@ -83,6 +84,7 @@ type Adder struct {
 	RawLeaves  bool
 	Silent     bool
 	Wrap       bool
+	WpName     string
```
schomatis: `WpName` -> `Name`

core/coreunix/add.go
schomatis: Let's drop the `adder.Wrap` check.

schomatis: Do we need an `EqualFold` call to compare to an empty string? I don't think the case is a problem here.

core/coreunix/add.go
```diff
@@ -470,6 +472,9 @@ func (adder *Adder) addFile(file files.File) error {
 		return err
 	}
 
+	if !strings.EqualFold(adder.WpName, "") && adder.Wrap {
+		return adder.addNode(dagnode, adder.WpName)
```
schomatis: I'd rather avoid duplicating the function call, we can define a new variable here, e.g., `addFileName := file.FileName()`, and if that name is `""` we set it to `adder.Name`. e.g.,

```Go
addFileName := file.FileName()
if addFileName == "" && adder.Name != "" {
  addFileName = adder.Name

  adder.Name = ""
  // Unset the `Name` option to avoid applying the same name to many unnamed files.
  // TODO: This could be turned into a FIFO queue to handle the scenario with more than one unnamed file.
}

adder.addNode(dagnode, addFileName)
```

We do two checks:
1. Apply the `--name` option only to unnamed files (e.g., coming from `stdin`).
2. Apply the `--name` option only once (since I fear that the HTTP commands library may be able to send many unnamed files), otherwise we would have many files with the same name in one directory.

kjzz:
@schomatis thank you very much for your advice.I have modify my pr.

core/coreunix/add.go
```diff
@@ -470,8 +471,13 @@ func (adder *Adder) addFile(file files.File) error {
 		return err
 	}
 
+	addFileName := file.FileName()
+	if addFileName == "" && adder.Name != "" {
```
schomatis: @keks WDYT about checking for an empty filename? I think this covers the `stdin` case in CLI, I'm not sure what effect this may have in the HTTP case, that's why we're resetting the `adder.Name` option to rename only one file (in the case HTTP may send many unnamed files).

keks: It looks like there are several occasions where we use an empty string as file name. I just looked into this and it looks like when using stdin, te `FullName()` should be `os.Stdin.Name()`, i.e. `/dev/stdin`. IMHO that would be a better check, and should also possible on the http server side.

kjzz: hello @keks ,i want to fix this bug.And i do not understand the fullname.can u explain more for me?

keks: @kjzz Sure! The `file` variable is a `"files".File`, an interface described here: https://godoc.org/github.com/ipfs/go-ipfs-cmdkit/files#File. It has the method `FileName()`, which is currently used in the check. But it also has the method `FullName()`, which should return `/dev/stdin` if the file is stdin.  If you check `file.FullName() == "/dev/stdin"` or `file.FullName() == os.Stdin.Name()`, that should be more reliable, because there are other reasons why `file.FileName()` may be the empty string. I only grepped through the code in go-ipfs-cmds and saw a lot of files with empty string as name. I'm not sure this will be a problem in practice, but I can't rule out that it won't. I think checking `file.FullName()` should be much more reliable.

schomatis: Thanks @keks! That sounds much more robust, @kjzz, so the check would be
```
if file.FullName() == os.Stdin.Name() && adder.Name != "" {
```

core/commands/add.go
```diff
@@ -116,6 +117,7 @@ You can now check what blocks have been created by:
 		cmdkit.BoolOption(trickleOptionName, "t", "Use trickle-dag format for dag generation."),
 		cmdkit.BoolOption(onlyHashOptionName, "n", "Only chunk and hash - do not write to disk."),
 		cmdkit.BoolOption(wrapOptionName, "w", "Wrap files with a directory object."),
+		cmdkit.StringOption(pathName, "Assign path name when use wrap-with-directory option"),
```
schomatis: Let's change this to reflect that the option (with the new check) works only for `stdin`, e.g., "Assign a name if the file source is stdin" (or something like that, feel free to reword it).

kjzz: @schomatis @keks In godoc,the file has a function named `fullpath` not `fullname`.So i use `fullpath` to judge the name, And In my test,if i use command such as
```
$cat ./test1.txt | ../ipfs add -w --name "myfile.txt"
added Qmf3xdREx34pewgYZjyvHQsrZxuk1TLUFdzoF95dsXfv3k myfile.txt
added QmSLquBbWRXs3VjTCJBXYv4BHn7NFNZxh5iU3zGNp5aMcf
```
the file path is also is empty string:
```
09:20:50.138 INFO   coreunix: The file path is : add.go:475
```
not os.Stdin(/dev/stdin).
So please help me review my pr.

kjzz:
@schomatis @keks In godoc,the file has a function named fullpath not fullname.So i use fullpath to judge the name, And In my test,if i use command such as

$cat ./test1.txt | ./ipfs add -w --name "myfile.txt"
added Qmf3xdREx34pewgYZjyvHQsrZxuk1TLUFdzoF95dsXfv3k myfile.txt
added QmSLquBbWRXs3VjTCJBXYv4BHn7NFNZxh5iU3zGNp5aMcf
the file path is also is empty string，I have log the full path in ipfs the log is:

09:20:50.138 INFO   coreunix: The file path is : add.go:475
not os.Stdin(/dev/stdin).
So please help me review my pr.

schomatis:
> the file has a function named fullpath not fullname

You're right, we copied it incorrectly, it's `FullPath()`.



> I have log the full path in ipfs the log is:
> 
> 09:20:50.138 INFO   coreunix: The file path is : add.go:475
> not os.Stdin(/dev/stdin).

Could you double check that please? If I add the line
```
fmt.Printf("Fullpath: %s\n", file.FullPath())
```
I'm seeing
```
Fullpath: /dev/stdin
```
(not an empty string).

Feel free to include the log line (temporarily) in the PR, to check that you're printing the correct attribute.

kjzz:
@schomatis 
image
I have take a picture for you.And my computer is mac os.Maybe the computer system will cause difference?

kjzz:
@schomatis  And I have pushed my log line in pr

kjzz:
@schomatis please help check it,Thx.

schomatis:
> And my computer is mac os.Maybe the computer system will cause difference?

Yes, that must be it, because with your patch I'm still seeing the `/dev/stdin`.

@keks Thoughts?

kjzz:
@schomatis @keks so can i use `if (file.FullPath() == "/dev/stdin"||file.FullPath() == "") && adder.Name != "" {` to make different system compatible.

kjzz:
And i have done some compatible processing.It is OK? @schomatis @keks

core/coreunix/add.go
```diff
@@ -470,8 +471,13 @@ func (adder *Adder) addFile(file files.File) error {
 		return err
 	}
 
+	addFileName := file.FileName()
+	if (file.FullPath() == "/dev/stdin" || file.FullPath() == "") && adder.Name != "" {
```
kjzz: To make macos and linux compatible.

keks:
Ah, `FullPath`, not `FullName`, sorry. I'm puzzled. We're using empty string `FullPath()` in multiple places as well so again I don't think it's a good way to use that as a check. Actually I believe it *should* return `/dev/stdin` on Mac and even Windows, I'm looking into that now.

keks:
Okay I just let this run on linux over HTTP and I also get an empty string. So this is not a Mac-only thing.

Unfortunately the story around which files with which names and paths is not very consistent, so it looks like we first need to make sure we can reliably detect the stdin file. Doing that now.

kjzz:
@keks But i think that No matter the file add by HTTP or Stdin.It can still work.Can you give me a counterexample about how the pr will be wrong?
Beceuse i just want to solve problem about the wrap file.You can take a look at [this](https://github.com/ipfs/go-ipfs/issues/4986#issuecomment-417590228)
Thx a lot.

schomatis:
Hey @kjzz, it's not that it won't work in *this* particular case, but we need to be sure this doesn't affect the API in different scenarios. I'm labeling this as blocked until @keks can figure out what's the best way to detect the `stdin` case. Please don't be discouraged by this setback, you did some nice work here, in the meanwhile you can focus on other issues to work on.

kjzz:
Or do you have any idea about check stdin file? @keks

keks:
I did more research on our possibilies here.

What seems to work with both local and http calls is to check that `file.(files.FileInfo).AbsPath() == "/dev/stdin"`. There is a commit with this change on this branch of my repo: https://github.com/keks/go-ipfs/tree/zkj/feat. For the purpose of demonstration I left a `fmt.Println` in the code that prints the `AbsPath`. I don't want to merge that, though. Can you verify this works for both of you @schomatis and @kjzz?

kjzz:
@keks Thx a lot. The `file.(files.FileInfo).AbsPath() ` works for me.And i have update the code.So can you help me review my pr again? This pr to fix the issue #5399 .Thank you so much.

keks:
@kjzz Also, it would be great if you could add sharness tests for this option. They are shell-script style tests. The test for `ipfs add -w` is here: https://github.com/ipfs/go-ipfs/blob/master/test/sharness/t0043-add-w.sh.

I propose you make a new file `t0047-add-name.sh` in which you test the `--name` option both with and without the `-w` flag. If you need any help, please ask! Sorry I forgot to say this earlier.

schomatis:
This is working for me also.

kjzz:
hi @schomatis @keks , i have add test in pr.Please help me review it .Thx a lot

schomatis:
Hey @kjzz, thanks for taking the time to prepare those tests, given that the `--name` option is a minor one I don't think we need a new test file for it, also there's no need to test the `-w` option explicitly since we have decoupled the `--name` option from that one (we can, though, use the `-w` option to test `ipfs cat`ing a file through its name and not its hash), you can compact your tests into something like this

```
  test_expect_success "ipfs add --name" '
    HASH="QmdFyxZXsFiP4csgfM5uPu99AvFiKH62CSPDw5TP92nr7w" &&
    echo "IPFS" | ipfs add --name file.txt > actual &&
    echo "added $HASH file.txt" > expected &&
    test_cmp expected actual
  '

  test_expect_success "ipfs cat with name" '
    HASH=$(echo "IPFS" | ipfs add -w --name file.txt -Q) &&
    ipfs cat /ipfs/$HASH/file.txt > expected &&
    echo "IPFS" > actual &&
    test_cmp expected actual
  '
```

(feel free to modify/extend this test example in whatever way you see fit). Then you can add them at the end of [`test_add_cat_file()`](https://github.com/ipfs/go-ipfs/blob/2dca8c2f965a4e4feb9df98f059f954de16846b6/test/sharness/t0040-add-and-cat.sh#L187).

kjzz:
@schomatis hey ,i have update my test case.could you help me review it.Thx a lot

keks:
Okay, the patch looks good. For some reason the sharness tests failed, but this looks unrelated. Anyway, I've restarted them to see whether that was a glitch or a reproducible problem. Let's wait until they finish.

kjzz:
@keks thx a lot .i have fmt code again.if test have any problem,please ping me

kjzz:
hey @keks ,there are something wrong with ci ?

kjzz:
hello @keks , thanks for your help , the pr has passed the required test .please help me review it.Thx a lot again.

kjzz:
Hey @schomatis , i have updated test in pr.Please help me review it.Thanks a lot!

kjzz:
hey @schomatis @keks , can you help me review my pr？I have finish my test file.Thx a lot.

keks:
Thank you, these look good. Great work!

keks:
@schomatis I don't have merge access. Are you down with making sure this gets merged?

kjzz:
Hi @schomatis , if you have time to review my code . can you help me restart my pr test.There are something wrong with ci.

core/coreunix/add.go
```diff
@@ -470,8 +471,14 @@ func (adder *Adder) addFile(file files.File) error {
 		return err
 	}
 
+	addFileName := file.FileName()
+	addFileInfo := file.(files.FileInfo)
```
Stebalien: @keks are you *sure* this won't panic in some cases? We should probably use a checked type assertion (or comment on why this *must* be true).

kjzz: Hi @schomatis , i have add a checked type assertion in pr.Maybe it will be more robust.

kjzz:
Hi @Stebalien  , i have add a checked type assertion in pr.Maybe it will be more robust.

Stebalien:
@keks @schomatis are we sure this is the best way to fix this? This feels like something that should be fixed in go-ipfs-cmds itself (i.e., allow a `--stdin-name` flag and use this to pick the name for files added via stdin).

Note: As long as we can get agree on the name of the *flag*, we can merge this as-is. I'm just really not a fan of stringly typed stuff and, as-is, this fix is a quite hacky.

core/commands/add.go
```diff
@@ -116,6 +117,7 @@ You can now check what blocks have been created by:
 		cmdkit.BoolOption(trickleOptionName, "t", "Use trickle-dag format for dag generation."),
 		cmdkit.BoolOption(onlyHashOptionName, "n", "Only chunk and hash - do not write to disk."),
 		cmdkit.BoolOption(wrapOptionName, "w", "Wrap files with a directory object."),
+		cmdkit.StringOption(pathName, "Assign a name if the file source is stdin."),
```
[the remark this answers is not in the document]
kevina: That seams a little two verbose to me.  Also, in the future want to allow this to override an existing name and then we will be stuck with both a `--name` and `--stdin-name` option.

For now I recommend we keep it as is and instead return an error (or maybe just a warning) when the option won't have any effect.

kjzz:
hey @Stebalien , i have update the `stdin-name`,and i know what you worried about.And i think  that this way is a useful way to fix the problem but maybe is the not best one.If you have any advice,you can ping me.Thx a lot for helping me review it.

kjzz:
@Stebalien If you have any advice,you can ping me.Thx a lot for helping me review pr.

schomatis:
@Stebalien Agreed that this is far from optimal, but I think the patch is cleanly encapsulated inside the `Adder` and it doesn't decrease code quality. We can use this as a first step towards the correct solution inside `go-ipfs-cmds`, also @kjzz has invested a significant amount of time and effort following our (sub-optimal :) advice so it would be nice to get this merged.

core/commands/add.go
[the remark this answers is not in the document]
Stebalien: > Also, in the future want to allow this to override an existing name and then we will be stuck with both a --name and --stdin-name option.

Actually, my thought was that this should *never* be a general-purpose flag. I believe that was the original suggestion but, as was discovered when writing this patch, that's ambiguous with multiple inputs. We *also* don't want to change the semantics at some point in the future (IMO).

Stebalien:
@schomatis, @keks you're right. This is a strict improvement, even if there's a better way to do this, we might as well merge this and improve it later.

core/commands/add.go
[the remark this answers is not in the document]
Stebalien: We could *also* have a "rename" feature but we'll have to carefully consider how we want to do it.

Note, by "multiple inputs" I mean:

```bash
> ipfs add -w --name "foo" file1 file2
```

We can only ever have one stdin so setting the name for stdin is unambiguous. We can easily add multiple files at the same time.

Stebalien:
Want to make sure @kevina and I are on the same page and then we can merge this.

kevina:
@Stebalien if the goal is for this to only every work from naming an otherwise unnamed file from stdin I think I am okay with `--stdin-name` a more functional `--name` could also be useful, but if we are abandoning the idea I think I am okay with this.

I will get back to this later today or tomorrow and give my final approval then.

kjzz:
Hey @kevina , the pr is aimed at solving  name an otherwise unnamed file from stdin,and i know it is not the best solution.And i will continue to perfect it in the next step.In addition,if @Stebalien have any new issue or advice about this,please ping me and i am very glad to work on it.Thx a lot.

Stebalien:
(this has 3 signoffs so I think it's ready to go; we can always change our minds later as long as we do it fast enough)

djdv:
I'm late on this, and not sure how important it is. But doesn't this unnecessarily break struct alignment?
https://github.com/kjzz/go-ipfs/blob/b7ea4bfc2a2c3b864d688c108245a3429029e185/core/coreunix/add.go#L79-L88

Stebalien:
Yeah, we *could* fix this but this isn't exactly a bottleneck. Feel free to submit a patch if you want but it's probably not worth the time.

PR #5403, opened 2018-08-27
Add postgres datastore type

JamesChristie:
This is work that aims to integrate changed submitted [here](https://github.com/whyrusleeping/sql-datastore/pull/2) and [here](https://github.com/ipfs/go-ipfs-config/pull/5). No changes have been made to the gx package hashes as the work it depends upon has not yet been accepted.


repo/fsrepo/datastores.go
```diff
@@ -410,3 +412,36 @@ func (c *badgerdsDatastoreConfig) Create(path string) (repo.Datastore, error) {
...
+	c.path, ok = params["path"].(string)
+	if !ok {
+		log.Fatal("no path")
+	}
+
+	return &c, nil
+}
+
+type postgresDatastoreConfig struct {
+	path string
```
magik6k: `path` isn't needed here either.

repo/fsrepo/datastores.go
```diff
@@ -410,3 +412,36 @@ func (c *badgerdsDatastoreConfig) Create(path string) (repo.Datastore, error) {
...
+
+func (c *postgresDatastoreConfig) DiskSpec() DiskSpec {
+	return map[string]interface{}{
+		"type": "postgresds",
+		"path": c.path,
+	}
+}
+
+func (postgresDatastoreConfig) Create(path string) (repo.Datastore, error) {
+	pg := postgresdb.Options{}
```
magik6k: This should get Host/User/Password/Database from config.

When initializing node with this backend user could pass those using env vars. in [config profile](https://github.com/ipfs/go-ipfs-config/pull/5/files#diff-e72534bab7cc380ab4281a57e79e8186R141) we'd just call `os.Getenv`.

We then could provide a simple script like the one below to avoid storing passwords in shell history
```bash
#!/bin/sh

[check args]

export IPFS_PG_HOST=$1
export IPFS_PG_USER=$2
export IPFS_PG_DATABASE=$3
printf "${IPFS_PG_USER}@${IPFS_PG_HOST} password:"
stty -echo
read IPFS_PG_PASSWORD
stty echo
export IPFS_PG_PASSWORD

ipfs init --profile=postgres
```

repo/fsrepo/datastores.go
```diff
@@ -410,3 +412,36 @@ func (c *badgerdsDatastoreConfig) Create(path string) (repo.Datastore, error) {
...
+}
+
+type postgresDatastoreConfig struct {
+	path string
+}
+
+func (c *postgresDatastoreConfig) DiskSpec() DiskSpec {
+	return map[string]interface{}{
+		"type": "postgresds",
+		"path": c.path,
```
magik6k: `path` is not needed here, but I'm not sure what to put here that makes sense. My guess would be user+database, as host/password may change frequently in some setups.

repo/fsrepo/datastores.go
i-norden: Thanks for the feedback! Making these changes now. I think we could also use the PGPASSFILE env variable to avoid directly storing the password as an env variable.

.gx/lastpubver
```diff
@@ -1 +1 @@
-0.4.18: QmUJYo4etAQqFfSS2rarFAE97eNGB8ej64YkRT2SmsYD4r
+0.4.20: QmNVba6esWZegye6kfURL5YnQapDYqfiemCUGhstMSFTpj
```
magik6k: go-ipfs releases are a bit more involved as this is a top-level package, you shouldn't call `go release` here.

.gitignore
```diff
@@ -22,3 +22,5 @@ bin/tmp
 vendor
 .tarball
 go-ipfs-source.tar.gz
+
+.idea
```
magik6k: I'd recommend putting this in `$HOME/.gitignore`, project specific gitignores shouldn't really have to care about all IDEs in existence.

plugin/plugins/postgresds/postgresds.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+func (c *datastoreConfig) Create(path string) (repo.Datastore, error) {
+	if c.passfile != "" {
+		info, err := ioutil.ReadFile(c.passfile)
+		if err != nil {
+			return nil, err
+		}
+		envVars := strings.Split(string(info), ":")
+		c.host = envVars[0]
```
magik6k: should also handle port from `[1]`

plugin/plugins/postgresds/postgresds.go
```diff
@@ -0,0 +1,110 @@
...
+func (c *datastoreConfig) DiskSpec() fsrepo.DiskSpec {
+	return map[string]interface{}{
+		"type":     "postgres",
+		"user":     c.user,
+		"database": c.dbname,
+	}
+}
+
+func (c *datastoreConfig) Create(path string) (repo.Datastore, error) {
+	if c.passfile != "" {
```
magik6k: This should happen in `DatastoreConfigParser` so that `DiskSpec` writes the right thing

plugin/plugins/postgresds/postgresds.go
```diff
@@ -0,0 +1,110 @@
+package postgresds
+
+import (
+	"fmt"
+	"io/ioutil"
+	"strings"
+
+	postgresdb "gx/whyrusleeping/QmXyfyKbv3EkZUtN5oxL2c32HNPaJKYsCQhxPufMPxKwzi/sql-datastore/postgres"
```
magik6k: shouldn't this be `gx/ipfs/QmXyfyKbv3EkZUtN5oxL2c32HNPaJKYsCQhxPufMPxKwzi/sql-datastore/postgres`?

plugin/plugins/postgresds/postgresds.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+}
+
+func (c *datastoreConfig) Create(path string) (repo.Datastore, error) {
+	if c.passfile != "" {
+		info, err := ioutil.ReadFile(c.passfile)
+		if err != nil {
+			return nil, err
+		}
+		envVars := strings.Split(string(info), ":")
```
magik6k: should check for length of `envVars`

plugin/plugins/postgresds/postgresds.go
```diff
@@ -0,0 +1,110 @@
...
+	return map[string]interface{}{
+		"type":     "postgres",
+		"user":     c.user,
+		"database": c.dbname,
+	}
+}
+
+func (c *datastoreConfig) Create(path string) (repo.Datastore, error) {
+	if c.passfile != "" {
+		info, err := ioutil.ReadFile(c.passfile)
```
magik6k: given that passfile usually lives in user home directory this should probably use `homedir.Expand(filepath.Clean(c.passfile))`, like in https://github.com/ipfs/go-ipfs/blob/508f307c356645156b2458b2c2a9ca443a637017/repo/fsrepo/fsrepo.go#L189

scripts/ipfs_postgres.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh
```
magik6k: Utility scripts should go to `misc/`

.gx/lastpubver
i-norden: Thanks, still new to gx- as you can probably tell with how messy things are!

.gitignore
i-norden: Will do!

i-norden:
I believe I've addressed all of the above issues. The required [sql-datastore PR](https://github.com/whyrusleeping/sql-datastore/pull/2) has been merged but the [go-ipfs-config PR](https://github.com/ipfs/go-ipfs-config/pull/5) still needs to be merged, [iptb-plugins](https://github.com/ipfs/iptb-plugins) will then need to be gx-ified with the new go-ipfs-config hash, and then finally go-ipfs can be gx-ified with the new go-ipfs-config and iptb-plugins hashes.

plugin/loader/preload_list
```diff
@@ -8,3 +8,4 @@ ipldgit github.com/ipfs/go-ipfs/plugin/plugins/git 0
 badgerds github.com/ipfs/go-ipfs/plugin/plugins/badgerds 0
 flatfs github.com/ipfs/go-ipfs/plugin/plugins/flatfs 0
 levelds github.com/ipfs/go-ipfs/plugin/plugins/levelds 0
+postgresds github.com/ipfs/go-ipfs/plugin/plugins/postgresds 0
```
magik6k: While it may be useful for quite a few people, I wouldn't preload this plugin by default (leaving commented out is fine)

i-norden: Thank you for all the help, magik6k!

magik6k:
Jenkins seems to be unable to fetch dependencies from you - https://ci.ipfs.team/blue/organizations/jenkins/IPFS%2Fgo-ipfs/detail/PR-5403/44/pipeline. (gitcop warning needs fixing tee)

i-norden:
> Jenkins seems to be unable to fetch dependencies from you - https://ci.ipfs.team/blue/organizations/jenkins/IPFS%2Fgo-ipfs/detail/PR-5403/44/pipeline. (gitcop warning needs fixing tee)

After making some very minor last minute changes to sql-datastore.. I forgot to re-gx-ify the package and so it was merged with the wrong hash :( 

My apologies, sql-datastore will need a gx release patch performed.

i-norden:
sql-datastore has been updated with the correct hash and `make gx-deps` is passing where it previous failed, but we still run into an issue with the sql-datastore dependency when running the shell script https://ci.ipfs.team/blue/organizations/jenkins/IPFS%2Fgo-ipfs/detail/PR-5403/51/pipeline. 

I am also confused because the travis-ci builds are passing on [OSX](https://travis-ci.org/ipfs/go-ipfs/jobs/484023923) and only run into this issue with sql-datastore with the [Linux](https://travis-ci.org/ipfs/go-ipfs/jobs/484023921) builds.

Before this can be merged [go-ipfs-config PR](https://github.com/ipfs/go-ipfs-config/pull/5) needs to be merged and its hash updated here and in iptb-plugins, and then the iptb-plugins hash needs to be updated here as well.

Stebalien:
Lol. It looks like gx is case sensitive when it comes to language names.

Stebalien:
Try again with: QmZmPKg1RKY2Cy5czKEBJTfDYzenbGww2pFNQKxLSKrPRB

(you're not the first to find this, apparently: https://github.com/whyrusleeping/gx/issues/187)

i-norden:
Yay that looks to have fixed the problem, thanks! I believe the error it is throwing now is just due to waiting on the updated go-ipfs-config dependencies.

Stebalien:
That was just a bug that has been fixed in master. I've rebased for you.

Issue #5232, opened 2018-07-14
Stream name resolve results as records get discovered


magik6k:
`ipfs name resolve` could get a `--stream` flag, with which it would output records as they get discovered in dht.

With this things like the git remote helper could start fetching objects after just 1 record has been found in the dht, and just shift the tree head if a newer record is found.

magik6k:
This will require implementing `SearchValue` on the routing interface - https://github.com/libp2p/go-libp2p-routing/blob/master/routing.go#L61.

PR #5404, opened 2018-08-27
ipfs name resolve --stream

magik6k:
Implements https://github.com/ipfs/go-ipfs/issues/5232

TODO:
- [x] Remove sync namesys code
- [x] Fix broken stuff
- [x] Wire up the command
- [ ] Test the command


namesys/routing.go
```diff
@@ -76,59 +88,109 @@ func (r *IpnsResolver) resolveOnce(ctx context.Context, name string, options *op
...
-		log.Errorf("encountered error when parsing EOL: %s", err)
-		return "", 0, err
-	}
+	return out
+}
 
-	return p, ttl, nil
+func (r *IpnsResolver) searchValue(ctx context.Context, key string, opts ...ropts.Option) (<-chan []byte, error) {
+	if ir, ok := r.routing.(*dht.IpfsDHT); ok {
+		return ir.SearchValue(ctx, key, opts...)
```
Stebalien: Mind:

1. Fixing the interfaces.
2. Taking over https://github.com/libp2p/go-libp2p-pubsub-router/pull/3 (or start over), making it follow the new API.

Instead?

magik6k: > Taking over libp2p/go-libp2p-pubsub-router#3 (or start over), making it follow the new API.

I wanted to do that, but in a separate PR

Stebalien: Yeah, as-is, this'll fail if ipns-over-pubsub is enabled. We could possibly work around that but I'd rather not.

magik6k:
Blocked on https://github.com/ipfs/go-ipfs/pull/5517

magik6k:
Test fails will get fixed with https://github.com/libp2p/go-libp2p-routing-helpers/pull/9

magik6k:
So there is a slight problem with `SearchValue` interface(doc)/implementations:

https://github.com/libp2p/go-libp2p-routing/blob/master/routing.go#L55
> Implementations may halt the search after a period of time or **may continue searching indefinitely.**

Now that this PR changes namesys to use `SearchValue` under the hood this makes `ipfs name resolve` hang indefinitely when IPNS-Pubsub is enabled.

I think that we should change this to:
 `Implementations MUST stop the search as soon as a good value is found. A good value is a value which would be returned by GetValue.`
It would make things more consistent, and if we ever want to get the 'run indefinitely' thing, we can add a routing option for that easily.

@Stebalien does that make sense?

Stebalien:
> Implementations MUST stop the search as soon as a good value is found. A good value is a value which would be returned by GetValue.

My primary concern is that "good" isn't really well defined. We *try* to get some kind of consensus in the DHT but that only sort of works. We don't even try in pubsub, we just accept the first valid answer.

My thinking *here* was that users would be able to call `ipfs name search QmId....`, reading results of until they're happy. However, the issue with that is "happy". Callers really have no way to decide what a "good" value is either.

Really, I think I was conflating search and subscribe (which we should also have but that's a different story). You're probably right, the *search* should end at some point.

Questions:

* What should pubsub do? Just do what we do for get?
* What should we do when one search "ends"? Is there any way to signal "we got a good value" so we can stop searching other routers?

magik6k:
> Really, I think I was conflating search and subscribe (which we should also have but that's a different story)

We can add a routing option for SearchValue which would make it into subscription as it's logic would be for the most part the same

> What should pubsub do? Just do what we do for get?

Yep, I guess

> What should we do when one search "ends"? Is there any way to signal "we got a good value" so we can stop searching other routers?

In case of `routing-helpers.Parallel`, If we query n routers and one of them closes its channel after sending results, we should close the output channel and cancel the other sub-requests (this is basically what happens in the GetValue case, we pick the first one).

The funny case here is when we get some value from router A, then some better value from router B, and then router A closes it's channel. I think we should still close the output channel and return the better value, but this feels a bit weird.

Stebalien:
> In case of routing-helpers.Parallel, If we query n routers and one of them closes its channel after sending results, we should close the output channel and cancel the other sub-requests (this is basically what happens in the GetValue case, we pick the first one).

Yeah... I guess that's as-good or better than `GetValue`.

> The funny case here is when we get some value from router A, then some better value from router B, and then router A closes it's channel. I think we should still close the output channel and return the better value, but this feels a bit weird.

This is weird because it indicates that A was *wrong*. That is, it returned a sub-optimal answer. A "smart" system would distrust A and then wait for B to finish. However, that may be too unpredictable. Worse, we could hang forever if the "winner" was pubsub.

I'd just go with your gut. We can experiment later.

> We can add a routing option for SearchValue which would make it into subscription as it's logic would be for the most part the same

Sounds reasonable. We can even use options for termination strategy (we can add them as needed). One (not on by default) can be "continue forever".

magik6k:
[Ready for a review]

Stebalien:
@kevina @magik6k assign me when you think this is ready (trying to hand off reviews...).

schomatis:
Sorry, I know nothing about IPNS to add some meaningful review value here.

Stebalien:
@schomatis even a "code looks good, I have no idea about the logic/design" can be helpful; it also helps you learn about this subsystem.

core/corehttp/gateway_test.go
```diff
@@ -57,6 +57,14 @@ func (m mockNamesys) Resolve(ctx context.Context, name string, opts ...nsopts.Re
 	return value, nil
 }
 
+func (m mockNamesys) ResolveAsync(ctx context.Context, name string, opts ...nsopts.ResolveOpt) <-chan namesys.Result {
+	out := make(chan namesys.Result, 1)
+	v, err := m.Resolve(ctx, name, opts...)
+	out <- namesys.Result{Path: v, Err: err}
+	close(out)
+	return nil
```
Kubuxu: You never return the `out` stream.

namesys/base.go
```diff
@@ -1,56 +1,111 @@
...
 
-		if depth > 1 {
-			depth--
+				var subCtx context.Context
+				if subCh != nil {
+					// Cancel previous recursive resolve since it won't be used anyways
+					cancelSub()
+				}
+				subCtx, cancelSub = context.WithCancel(ctx)
+				defer cancelSub()
```
Stebalien: Defering in a loop is generally a bad idea (will allocate for each). Let's just make sure to call `cancelSub`. Maybe a `defer func() { if cancelSub != nil { cancelSub() }` at the top, or something like that.

namesys/base.go
```diff
@@ -1,56 +1,111 @@
...
-		if !matched {
-			return p, nil
-		}
+				if res.err != nil {
+					outCh <- Result{Err: res.err}
+					return
+				}
+				log.Debugf("resolved %s to %s", name, res.value.String())
+				if strings.HasPrefix(res.value.String(), "/ipfs/") {
+					outCh <- Result{Path: res.value}
```
Stebalien: Select on context.

namesys/base.go
```diff
@@ -1,56 +1,111 @@
...
+					break
+				}
+
+				if depth == 1 {
+					outCh <- Result{Path: res.value, Err: ErrResolveRecursion}
```
Stebalien: Select on context.

namesys/base.go
```diff
@@ -1,56 +1,111 @@
...
 
-		if !matched {
-			return p, nil
-		}
+				if res.err != nil {
+					outCh <- Result{Err: res.err}
+					return
+				}
+				log.Debugf("resolved %s to %s", name, res.value.String())
+				if strings.HasPrefix(res.value.String(), "/ipfs/") {
```
Stebalien: This should probably carry the `!prefix` change from #5602 as it has removed https://github.com/ipfs/go-ipfs/pull/5404/files#diff-f43001ea38f40375075f7bccccfd9ab2L38.

core/coreapi/name.go
```diff
@@ -125,12 +123,42 @@ func (api *NameAPI) Resolve(ctx context.Context, name string, opts ...caopts.Nam
 		name = "/ipns/" + name
 	}
 
-	output, err := resolver.Resolve(ctx, name, options.ResolveOpts...)
+	out := make(chan coreiface.IpnsResult)
+	go func() {
+		defer close(out)
+		for res := range resolver.ResolveAsync(ctx, name, options.ResolveOpts...) {
+			p, _ := coreiface.ParsePath(res.Path.String())
```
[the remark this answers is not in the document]
Stebalien: SGTM.

namesys/namesys.go
```diff
@@ -64,47 +64,91 @@ func (ns *mpns) Resolve(ctx context.Context, name string, options ...opts.Resolv
...
+	} else if isd.IsDomain(key) {
+		res = ns.dnsResolver
+	} else {
+		res = ns.proquintResolver
 	}
-	return p, 0, err
+
+	resCh := res.resolveOnceAsync(ctx, key, options)
+	var best onceResult
+	go func() {
```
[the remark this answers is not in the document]
magik6k: The logic doesn't really de-duplicate that nicely, I did wrap channel sending though

core/coreapi/name.go
```diff
@@ -89,9 +89,7 @@ func (api *NameAPI) Publish(ctx context.Context, p coreiface.Path, opts ...caopt
 	}, nil
 }
 
-// Resolve attempts to resolve the newest version of the specified name and
-// returns its path.
-func (api *NameAPI) Resolve(ctx context.Context, name string, opts ...caopts.NameResolveOption) (coreiface.Path, error) {
+func (api *NameAPI) Search(ctx context.Context, name string, opts ...caopts.NameResolveOption) (<-chan coreiface.IpnsResult, error) {
```
kevina: Please document this function.

test/sharness/t0183-namesys-pubsub.sh
```diff
@@ -28,8 +28,8 @@ test_expect_success 'check namesys pubsub state' '
 
 # These commands are *expected* to fail. We haven't published anything yet.
 test_expect_success 'subscribe nodes to the publisher topic' '
-    ipfsi 1 name resolve /ipns/$PEERID_0;
-    ipfsi 2 name resolve /ipns/$PEERID_0;
+    ipfsi 1 name resolve /ipns/$PEERID_0 --timeout=1s;
```
kevina: Why is the timeout now needed?

kevina:
@Stebalien does this still need a review from me?  It can likely go over it, but it will take me several hours as I am not that familiar with that code path?

core/coreapi/name.go
magik6k: There are docs for this on the interface (like all other coreapi functions). We may want to point people there, but I'd this in a separate PR

namesys/namesys.go
```diff
@@ -64,47 +64,91 @@ func (ns *mpns) Resolve(ctx context.Context, name string, options ...opts.Resolv
...
+		defer close(out)
+		for {
+			select {
+			case res, ok := <-resCh:
+				if !ok {
+					if best != (onceResult{}) {
+						ns.cacheSet(key, best.value, best.ttl)
```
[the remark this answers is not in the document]
Stebalien: Because the routers are returning records older than the cache? That's a bit annoying...

Stebalien:
@schomatis 

>  I'm not sure I agree with that, reviewing a 600-line-diff of a PR with very few comments seems to have a cost-benefit ratio that makes it very inefficient to learn about any particular subsystem. I also fear that a thumbs-up on a PR that I don't really understand can generate more noise than signal.

"needs comments", "what the hell is this doing", etc. are all reasonable first-pass reviews.

Is there a better way to learn a subsystem? At the end of the day, you can never really trust code comments anyways (although I agree we should have more). Simply *reading* a subsystem's code is a good start but reviewing changes to a subsystem forces you to really understand how it works (it forces you to understand how those changes might affect the rest of the subsystem).

core/coreapi/name.go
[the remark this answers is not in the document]
Stebalien: Actually, can you comment on this. I won't be the only one to ask.

core/coreapi/name.go
Stebalien: Let's just do this now. Separate PRs for documentation tend to not happen.

namesys/namesys.go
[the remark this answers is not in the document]
Stebalien: Hm, yeah, we probably need generics.

namesys/base.go
```diff
@@ -1,56 +1,119 @@
...
+
+				p := strings.TrimPrefix(res.value.String(), ipnsPrefix)
+				subCh = resolveAsync(subCtx, r, p, subopts)
+			case res, ok := <-subCh:
+				if !ok {
+					subCh = nil
+					break
+				}
+
+				emitResult(ctx, outCh, res)
```
Stebalien: There's a bit of a (benign) race here. We can call this, see that the context has been canceled, and then keep looping (if, e.g. there's a result waiting and we happen to select that channel first).

I'd rather just revert this change and *return* if the context has been canceled. That way we always abort ASAP.

test/sharness/t0183-namesys-pubsub.sh
Stebalien: This *is* quite unfortunate. In the past, `resolve` used to mean "go until we stop actively searching". Now, it means "go until we stop passively searching".

On the other hand, I doubt this'll make a difference in practice. On a real network, we never get negative results from the DHT, we just get timeouts.

PR #5409, opened 2018-08-29
directly parse peer IDs as peer IDs

Stebalien:
No need to parse it as a hash first.

### DIFF
--- a/namesys/routing.go
+++ b/namesys/routing.go
@@ -58,7 +58,7 @@ func (r *IpnsResolver) resolveOnce(ctx context.Context, name string, options *op
 	pid, err := peer.IDB58Decode(name)
 	if err != nil {
 		// name should be a multihash. if it isn't, error out here.
-		log.Debugf("RoutingResolver: bad input hash: [%s]\n", name)
+		log.Debugf("RoutingResolver: IPNS address not a valid peer ID: [%s]\n", name)
 		return "", 0, err
 	}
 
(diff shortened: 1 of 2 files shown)


Stebalien:
@kevina this is one of the few to only changes we'll need to make to go-ipfs to use non-byte-array CIDs.

namesys/routing.go
```diff
@@ -55,19 +55,13 @@ func (r *IpnsResolver) resolveOnce(ctx context.Context, name string, options *op
 	}
 
 	name = strings.TrimPrefix(name, "/ipns/")
-	hash, err := mh.FromB58String(name)
+	pid, err := peer.IDB58Decode(name)
```
kevina: Unless I am missing something, this is not performing the same set of checks.  In particular peer.IDB58Decode does not check it is a valid Multihash: https://github.com/libp2p/go-libp2p-peer/blob/master/peer.go#L121:L127

Stebalien: It calls `mh.FromB58String` internally.

kevina: Oops sorry.

Stebalien:
@kevina could I get a :heavy_check_mark:?

namesys/routing.go
```diff
@@ -55,19 +55,13 @@ func (r *IpnsResolver) resolveOnce(ctx context.Context, name string, options *op
...
 	if err != nil {
 		// name should be a multihash. if it isn't, error out here.
 		log.Debugf("RoutingResolver: bad input hash: [%s]\n", name)
```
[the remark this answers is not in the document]
Stebalien: Fixed (although I chose a different error message as that one isn't quite right either).

PR #5424, opened 2018-09-04
move VersionOption after GatewayOption to fix #5422

### LINKED ISSUE #5422: gateway overrides /version with information about IPFS
#### Version information:
0.4.17

#### Type:
Bug (probably)

#### Description:

When using a domain with a dnslink, any request to example.com/version will return information about the IPFS node running on the gateway, instead of potential information about the webapp's version.
Examples: https://riot.swedneck.xyz/version, https://https://docs.ipfs.io/version, http://explore.ipld.io/version

This breaks the version checking function of the Riot matrix client, which otherwise works perfectly via IPFS.

swedneck:


swedneck:
I'm not sure what you want me to add, i've never worked with the ipfs code (or go) before..

Stebalien:
I think this'll still miss, e.g., `/api/v0/version`. Is that correct? I think the correct order here is:

```
corehttp.MetricsCollectionOption("gateway"),
corehttp.IPNSHostnameOption(),
corehttp.GatewayOption(writable, "/ipfs", "/ipns"),
corehttp.VersionOption(),
corehttp.CheckVersionOption(),
corehttp.CommandsROOption(*cctx),
```

That is:

1. Collect metrics first.
2. Do any IPNS based redirects.
3. Handle /ipfs, /ipns next.
4. Handle /version.
5. Check client's declared a go-ipfs version (if specified). Put this right before the read only commands option as we don't care about the client version for anything *except* the read-only commands.
6. Forward to the read-only commands.

Stebalien:
Testing will be a bit tricky so we can handle that.

swedneck:
Should i change the order to that and push the commit here?

Stebalien:
@swedneck Yes (when you get a chance).

Issue #5367, opened 2018-08-10
Badger on 32-bit architectures


schomatis:
I understand that IPFS can run on 32-bit architectures, but as of now Badger doesn't seem capable of that, see https://github.com/dgraph-io/badger/issues/514 and  #5305.

Stebalien:
We should be able to operate in normal "io" mode on 32bit platforms.

manishrjain:
Submitted https://github.com/dgraph-io/badger/commit/341a135a4ab7d861599a45ca4ce00412c690bf66.

VictorBjelkholm:
Seems this has been merged (https://github.com/dgraph-io/badger/commits/master) but maybe the dep hasn't been updated yet?

Stebalien:
Badger has now been updated.

Stebalien:
New fixed: Badger was doing some atomic operations on some misaligned atomic variables. This has since been fixed but we still need to update badger.

Stebalien:
Badger has since been updated.

PR #5429, opened 2018-09-06
ci: test on 32bit os

magik6k:
To test for stuff like https://github.com/ipfs/go-ipfs/issues/5367

There is also one test fail in gateway test code:
```
core/corehttp/gateway_test.go:42:9: constant 18446744073709551615 overflows uint
```

### DIFF
--- a/test/bin/Rules.mk
+++ b/test/bin/Rules.mk
@@ -30,18 +30,18 @@ TGTS_GX_$(d) := hang-fds
 TGTS_GX_$(d) := $(addprefix $(d)/,$(TGTS_GX_$(d)))
 
 $(TGTS_GX_$(d)):
-	go build -i $(go-flags-with-tags) -o "$@" "$(call gx-path,$(notdir $@))"
+	go build $(go-flags-with-tags) -o "$@" "$(call gx-path,$(notdir $@))"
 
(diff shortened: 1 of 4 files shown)


magik6k:
Blocked on https://github.com/ipfs/go-ipfs/issues/5367 / propagating new badger

test/bin/Rules.mk
```diff
@@ -30,18 +30,18 @@ TGTS_GX_$(d) := hang-fds
 TGTS_GX_$(d) := $(addprefix $(d)/,$(TGTS_GX_$(d)))
 
 $(TGTS_GX_$(d)):
-	go build -i $(go-flags-with-tags) -o "$@" "$(call gx-path,$(notdir $@))"
```
Stebalien: Do you remember why we needed to remove this flag?

magik6k: IIRC when cross compiling, go would try to write objects to /usr

Stebalien: Ah. Yeah, that makes sense.

Stebalien:
Still blocked: https://github.com/ipfs/go-ipfs/issues/5367#issuecomment-457247271

PR #5433, opened 2018-09-06
improve gateway options test

Stebalien:
Unfortunately, this doesn't really test the *actual* gateway config as it uses
the options specified by the test. However, it's a step in the right direction.


schomatis:
@Stebalien Could you provide me some context here please? Or point me to the issue this PR is addressing?

Stebalien:
#5422

PR #5439, opened 2018-09-07
fix(commands): fix filestore.go goroutine leak

overbool:
ipfs/go-ipfs/issues/4414
fix filestore.go goroutine leak

License: MIT


Stebalien:
Thanks!

PR #5440, opened 2018-09-07
remove Godeps

### LINKED ISSUE #2722: Migrating away from Godeps

Stebalien:
fixes #2722


test/3nodetest/Makefile
```diff
@@ -20,7 +19,8 @@ data/filerand: bin/random
 	./bin/random 50000000 > ./data/filerand
 
 bin/random: $(RANDOMSRC)/**/*.go
-	go build -o ./bin/random $(RANDOMSRC)/random
+	go get github.com/jbenet/go-random
+	go build -o ./bin/random github.com/jbenet/go-random/random
```
Stebalien: This is the only real change. It uses go-get, not gx, but, as far as I can tell, we don't use these tests at the moment anyways.

magik6k: Isn't this also https://github.com/ipfs/go-ipfs/pull/5440/files#diff-c8660427d0699f8fe97cb9943c5bd91eR40? I'd just depend on that

Stebalien:
> Isn't this also https://github.com/ipfs/go-ipfs/pull/5440/files#diff-c8660427d0699f8fe97cb9943c5bd91eR40? I'd just depend on that

SGTM. I was stuck trying to figure out the right gx hash (the makefile is independent of our make system) but I realized I can just *use* the other make system. LGTY?

PR #5452, opened 2018-09-13
feat(commands): resolve key

overbool:
Closes https://github.com/ipfs/go-ipfs/issues/3900

License: MIT


Stebalien:
I'd like to discuss the motivation a bit on https://github.com/ipfs/go-ipfs/issues/3900 first. Mixing a global namespace (/ipns/... addresses) and a local namespace (keys) feels a bit funny to me.

whyrusleeping:
@Stebalien whys that? The namespaces are completely disambiguated, keystore names cannot contain slashes.

Stebalien:
(discussion on the issue so it's all in one place)

Issue #4914, opened 2018-04-03
Add a --dry-run flag to config profile apply


Stebalien:
This would print a diff between the current config and the config that *would* be generated. Possibly using, e.g., https://github.com/elgris/jsondiff

preetsinojiya:
Seems like no one is working on this one now. So will work on this one and update the progress.

preetsinojiya:
Any suggestion for the change that aligns with the ipfs preferred code design will be appreciated.

PR #5455, opened 2018-09-13
feat: add dry-run flag for config profile apply command

chenminjian:
Issue: #4914

License: MIT

### DIFF
--- a/core/commands/config.go
+++ b/core/commands/config.go
@@ -16,9 +16,16 @@ import (
 	repo "github.com/ipfs/go-ipfs/repo"
 	fsrepo "github.com/ipfs/go-ipfs/repo/fsrepo"
 
+	"gx/ipfs/QmP2i47tnU23ijdshrZtuvrSkQPtf9HhsMb9fwGVe8owj2/jsondiff"
 	config "gx/ipfs/QmSoYrBMibm2T3LupaLuez7LPGnyrJwdRxvTfPUyCp691u/go-ipfs-config"
 	"gx/ipfs/Qmde5VP1qUkyQXKCfmEUA7bP64V2HAptbJ7phuPp7jXWwg/go-ipfs-cmdkit"
 )
+
+// ConfigUpdateOutput is config profile apply command's output
+type ConfigUpdateOutput struct {
+	Old config.Config
+	New config.Config
+}
 
 type ConfigField struct {
 	Key   string
@@ -333,6 +340,9 @@ var configProfileApplyCmd = &cmds.Command{
 	Helptext: cmdkit.HelpText{
 		Tagline: "Apply profile to config.",
 	},
+	Options: []cmdkit.Option{
+		cmdkit.BoolOption("dry-run", "print difference between the current config and the config that would be generated"),
+	},
 	Arguments: []cmdkit.Argument{
 		cmdkit.StringArg("profile", true, false, "The profile to apply to the config."),
 	},
@@ -343,13 +353,40 @@ var configProfileApplyCmd = &cmds.Command{
 			return
 		}
 
-		err := transformConfig(req.InvocContext().ConfigRoot, req.Arguments()[0], profile.Transform)
+		dryRun, _, _ := req.Option("dry-run").Bool()
+		oldCfg, newCfg, err := transformConfig(req.InvocContext().ConfigRoot, req.Arguments()[0], profile.Transform, dryRun)
 		if err != nil {
 			res.SetError(err, cmdkit.ErrNormal)
 			return
 		}
-		res.SetOutput(nil)
+		res.SetOutput(&ConfigUpdateOutput{
+			Old: *oldCfg,
+			New: *newCfg,
+		})
 	},
+	Marshalers: cmds.MarshalerMap{
+		cmds.Text: func(res cmds.Response) (io.Reader, error) {
+			if res.Error() != nil {
+				return nil, res.Error()
+			}
+
+			v, err := unwrapOutput(res.Output())
+			if err != nil {
+				return nil, err
+			}
+
+			apply, ok := v.(*ConfigUpdateOutput)
+			if !ok {
+				return nil, e.TypeErr(apply, v)
+			}
+
+			diff := jsondiff.Compare(apply.Old, apply.New)
+			buf := jsondiff.Format(diff)
+
+			return strings.NewReader(string(buf)), nil
+		},
+	},
+	Type: ConfigUpdateOutput{},
 }
 
 func buildProfileHelp() string {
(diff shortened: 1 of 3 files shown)


chenminjian:
@Stebalien I use gx publish jsondiff lib, but test fail.  Can you give me some advice?

chenminjian:
@Stebalien I use gx publish jsondiff lib, but test fail. What should I do to solve this？

Stebalien:
* Our CI needs to be able to fetch gx packages so some reachable IPFS node has to have them pinned locally (done).
* Gx packages need to specify a `dvcsImport` path. Otherwise, path rewriting won't work. I've fixed up the gx packages so this should be fine now.

(basically, gx has a lot of rough edges we're working on ironing out).

core/commands/config.go
```diff
@@ -338,12 +346,39 @@ var configProfileApplyCmd = &cmds.Command{
...
+		diff, err := transformConfig(req.InvocContext().ConfigRoot, req.Arguments()[0], profile.Transform, dryRun)
 		if err != nil {
 			res.SetError(err, cmdkit.ErrNormal)
 			return
 		}
-		res.SetOutput(nil)
+		res.SetOutput(&ConfigProfileApplyOutput{
+			Diff: *diff,
+		})
+	},
```
Stebalien: We'll need to add a `Type: ConfigProfileApplyOutput{}` here.

core/commands/config.go
```diff
@@ -338,12 +346,39 @@ var configProfileApplyCmd = &cmds.Command{
...
+			if err != nil {
+				return nil, err
+			}
+
+			apply, ok := v.(*ConfigProfileApplyOutput)
+			if !ok {
+				return nil, e.TypeErr(apply, v)
+			}
+
+			dryRun, _, err := res.Request().Option("dry-run").Bool()
```
Stebalien: We actually realized that we'd prefer to *always* print out the diff, even if we're *not* doing a dry-run (https://github.com/ipfs/go-ipfs/pull/4923#discussion_r182314868). We can just return `strings.NewReader(jsondiff.Format(apply.Diff))` without checking `dryRun`.

chenminjian:
@Stebalien Thanks for your guidance. I find that /elgris/jsondiff doesn't support unmarshal. Can I write a new jsondiff lib for ipfs?

Stebalien:
Hm. I see. Well, we have two options:

1. Send a `[]jsondiff.Item` array (by calling `diff.Items()`) and then reconstruct the diff using `diff.Add` on the client. I'm not a fan of this approach...
2. Send back the old config and the new config and do the diff on the client. Personally, I'd rather do this.

That is, send back `ConfigUpdateOutput { Old: OldConfig, New: NewConfig }` and call `jsondiff` on the client. Does that sound reasonable? That's probably the most flexible solution.

chenminjian:
@Stebalien Thx. The second method looks great and I use it. Could you help me review my code?

Stebalien:
@schomatis in an effort to "spread the love", I'm assigning you to review :smile:.

core/commands/config.go
```diff
@@ -362,29 +398,44 @@ func buildProfileHelp() string {
 	return out
 }
 
-func transformConfig(configRoot string, configName string, transformer config.Transformer) error {
+func transformConfig(configRoot string, configName string, transformer config.Transformer, dryRun bool) (*config.Config, *config.Config, error) {
```
schomatis: It makes more sense to me to just return the `string` with the diff instead of the old and new configs, since the caller has no use for them except to print the diff.

core/commands/config.go
```diff
@@ -362,29 +398,44 @@ func buildProfileHelp() string {
...
+	if dryRun {
+		newCfg := *cfg
+		err = transformer(&newCfg)
+		if err != nil {
+			return nil, nil, err
+		}
+		return &oldCfg, &newCfg, nil
 	}
 
 	err = transformer(cfg)
```
schomatis: We're repeating the `transformer` call (with the same `*cfg` argument I believe). It would seem to me that the actual difference of using the `dry-run` option is whether or not we're persisting the new config (`BackupConfig`/`SetConfig` calls), maybe that can be moved to the `if` and avoid the early `return`. I think that would clarify the `cfg`/`oldCfg`/`newCfg` uses, we just stick to `oldCfg`/`newCfg` nomenclature. Does that make sense?

chenminjian: I think `oldCfg := *cfg` is required. This makes a copy of global `config.Config`.If we don't make a copy, we would update the global `config.Config` unintentionally.
Because `fsrepo.Open()`  returns a global `repo.Repo` instance and `Repo.Config()` returns a point that points to a global `config.Config`.

chenminjian: > We're repeating the `transformer` call (with the same `*cfg` argument I believe). It would seem to me that the actual difference of using the `dry-run` option is whether or not we're persisting the new config (`BackupConfig`/`SetConfig` calls), maybe that can be moved to the `if` and avoid the early `return`. I think that would clarify the `cfg`/`oldCfg`/`newCfg` uses, we just stick to `oldCfg`/`newCfg` nomenclature. Does that make sense?

I think `oldCfg := *cfg` is required. This makes a copy of global `config.Config`.If we don't make a copy, we would update the global `config.Config` unintentionally.
Because `fsrepo.Open()`  returns a global `repo.Repo` instance and `Repo.Config()` returns a point that points to a global `config.Config`.

chenminjian: It's my pleasure.

chenminjian: > Good point! Could you add a couple of comments to the function clarifying that please?

It's my pleasure.

chenminjian:
hey, @schomatis . I have completed modifications according to your advice. Could you help me review my code again? 😊

core/commands/config.go
Stebalien: Not sure I agree with this. Formatting the diff server-side means the client can't choose how it wants to display it (color, custom diff, etc.). How does this make it cleaner?

chenminjian: Both formatting the diff client-side and formatting the diff server-side can solve the problem. If we want to have better flexibility, maybe formatting the diff client-side is a better method.

Stebalien: > if, in the future, this needs to be extended to a more configurable command the two jsondiff lines can easily be extracted outside and the old/new config objects can be returned, but I'd wait to see if that need first arises.

My worry with this is that we'd either need to either add an option or change the API. If we do this all client-side, we can add the logic client-side without changing the API at all.

For me, this comes down to the fact these commands are both reasonably stable API endpoints *and* CLI commands. We may never change the output but, really, I can see us deciding that we don't like the JSON diffing algorithm, etc sometime in the future.

Actually, let's get some feedback from the JS people. Really, they should give feedback on the command itself.

schomatis: > For me, this comes down to the fact these commands are both reasonably stable API endpoints _and_ CLI commands. We may never change the output but, really, I can see us deciding that we don't like the JSON diffing algorithm, etc sometime in the future.

Agreed. @chenminjian Could you make that modification  (adding a comment line explainig what the function returns and why) please?

chenminjian: Ok. I have push the fixed code.

chenminjian:
@Stebalien Is there any other work that needs to be done? If so, please tell me.

schomatis:
@Kubuxu Could you expand some more on that please? How is this PR impacting security?

Kubuxu:
In the past we hardened config endpoints not to return `Identity.PrivKey` because it is possible to break CORS protection and access API endpoint in some browsers.

This would reintroduce this risk.

chenminjian:
This is a good suggestion. Thx a lot. I have updated the pr to stop exposing `Identity.PrivKey` over HTTP API.

Kubuxu:
Thanks for resolving it quickly.

Stebalien:
Blocked on https://github.com/ipfs/go-ipfs/pull/5455, unfortunately.

Stebalien:
(that is, the dereference isn't sufficient to actually *clone* the config).

chenminjian:
Maybe dereference is sufficient. Because `config.Transformer`, which is `transformConfig`' parameter , just updates map' point, not map' content. This does not affect the global config. Does that make sense?

Stebalien:
I see. Yeah, this *should* work... (well, shouldn't be any worse than it already is).

Issue #5443, opened 2018-09-08
Store connection direction for peers


andrewheadricke:
I would like to see in `ipfs swarm peers` or even `ipfs swarm peers -v` the direction of connection for each peer.  I basically want to know which peers connected to me, and which peers I have initiated a connection to.

I am trying to more closely manage my list of peers and this information would be very useful.

I'm guessing this would need to be implemented at the `libp2p` level?

overbool:
Hey @Stebalien, can I try to solve this?

Stebalien:
Go ahead.

overbool:
Hey @Stebalien, I add the direction of connection behind the multiaddr, when i use `ipfs swarm peers -v`. like the follow:

```
/ip4/158.69.53.41/tcp/4001/ipfs/Qm... outbound
/ip4/159.89.147.117/tcp/4001/ipfs/Qm... n/a outbound
/ip4/162.252.217.94/tcp/4001/ipfs/Qm... n/a outbound
/ip4/163.172.89.219/tcp/4001/ipfs/Qm... 239.373985ms outbound
  /ipfs/kad/1.0.0
/ip4/165.227.15.228/tcp/4001/ipfs/Qm... n/a outbound
  /ipfs/bitswap/1.1.0
  /ipfs/kad/1.0.0
```

What do you think of it？

PR #5457, opened 2018-09-13
feat(command): add connection direction

overbool:
License: MIT

Fixes: https://github.com/ipfs/go-ipfs/issues/5443

### DIFF
--- a/core/commands/swarm.go
+++ b/core/commands/swarm.go
@@ -62,6 +62,7 @@ var swarmPeersCmd = &cmds.Command{
 		cmdkit.BoolOption("verbose", "v", "display all extra information"),
 		cmdkit.BoolOption("streams", "Also list information about open streams for each peer"),
 		cmdkit.BoolOption("latency", "Also list information about latency to each peer"),
+		cmdkit.BoolOption("direction", "Also list information about the direction of connection"),
 	},
 	Run: func(req cmds.Request, res cmds.Response) {
 
@@ -79,6 +80,7 @@ var swarmPeersCmd = &cmds.Command{
 		verbose, _, _ := req.Option("verbose").Bool()
 		latency, _, _ := req.Option("latency").Bool()
 		streams, _, _ := req.Option("streams").Bool()
+		direction, _, _ := req.Option("direction").Bool()
 
 		conns := n.PeerHost.Network().Conns()
 		var out connInfos
@@ -86,9 +88,8 @@ var swarmPeersCmd = &cmds.Command{
 			pid := c.RemotePeer()
 			addr := c.RemoteMultiaddr()
 			ci := connInfo{
-				Addr:      addr.String(),
-				Peer:      pid.Pretty(),
-				Direction: inet.Direction(-1),
+				Addr: addr.String(),
+				Peer: pid.Pretty(),
 			}
 
 			/*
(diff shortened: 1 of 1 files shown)


core/commands/swarm.go
```diff
@@ -101,6 +100,9 @@ var swarmPeersCmd = &cmds.Command{
 			*/
 
 			if verbose || latency {
+				// set direction
+				ci.Direction = c.Stat().Direction
```
Stebalien: Instead of overloading the `latency` flag, let's just add a `--direction` flag. I'm usually all for reducing the number of flags we have but tying latency to connection direction feels a bit weird.

core/commands/swarm.go
```diff
@@ -81,15 +81,14 @@ var swarmPeersCmd = &cmds.Command{
...
 		for _, c := range conns {
 			pid := c.RemotePeer()
 			addr := c.RemoteMultiaddr()
-
 			ci := connInfo{
-				Addr: addr.String(),
-				Peer: pid.Pretty(),
+				Addr:      addr.String(),
+				Peer:      pid.Pretty(),
+				Direction: inet.Direction(-1),
```
Stebalien: Let's just let this default to 0 ("unknown"). By using `-1`, we're changing the API a bit: new clients running against old servers will see `Direction: 0` (instead of `Direction: -1`) even if they haven't requested the direction. When they do, they'll *print* the direction as if the user had requested it. See my comment at https://github.com/ipfs/go-ipfs/pull/5457/files#diff-ab43cdc4b214774869f9dc156ebc0fa0R152

core/commands/swarm.go
```diff
@@ -146,6 +148,11 @@ var swarmPeersCmd = &cmds.Command{
 				if info.Latency != "" {
 					fmt.Fprintf(buf, " %s", info.Latency)
 				}
+
+				if info.Direction >= 0 {
```
Stebalien: I'd just check to see if the user has passed the appropriate flag (either `--direction` or `--verbose`). That way we can keep the default direction as `0`.

*Alternatively*, we can just not print anything if the direction is unknown. Either way is fine by me.

overbool:
@Stebalien Thank you for your advice. could you help me review my pr again?

Stebalien:
Given that this is a change to a public API, I'd like to get one more sign off. I'm puling a few others to see if they have any options.

core/commands/swarm.go
```diff
@@ -146,6 +151,11 @@ var swarmPeersCmd = &cmds.Command{
 				if info.Latency != "" {
 					fmt.Fprintf(buf, " %s", info.Latency)
 				}
+
+				if info.Direction != inet.DirUnknown {
```
schomatis: nit (feel free to ignore): can we check for the direction or verbose flag here? I prefer not to rely on default (unset) values.

Stebalien: See: https://github.com/ipfs/go-ipfs/pull/5457#discussion_r217540453. We've explicitly defined the zero value of `Direction` to be `DirUnknown` so this *should* always work.

Stebalien: SGTY or would you actually like to print "unknown" if `--direction` is passed?

schomatis: > SGTY or would you actually like to print "unknown" if `--direction` is passed?

I'm fine either way ("unknown" or any non-empty string would make the output easier to parse in case we add more information fields in the future), my original comment was more about the fact that we are overloading the `DirUnknown` value to mean either unknown and/or unset, but's just a nit, I'm fine with the patch as is.

PR #5462, opened 2018-09-14
fix behaviour of key rename to same name

### LINKED ISSUE #5450: Key renaming to same name causes key to be deleted
### Version information:
```
go-ipfs version: 0.4.17-dfd19c470
Repo version: 7
System version: amd64/darwin
Golang version: go1.10.3
```

### Type:
Bug

### Description:
Renaming a key to the same name causes an error and the key is deleted.

### Reproduction steps:
If you create a key called 'foo', and rename this key to 'foo', there is an error and it disappears:
```
$ ipfs key list
self
$ ipfs key gen foo --type=rsa --size=2048
QmcaW6t1MpTBh46GvkaTUA57kr9WcyFk8Fe6nkxQpWzGm5
$ ipfs key list
self
foo
$ ipfs key rename --force foo foo
Key QmcaW6t1MpTBh46GvkaTUA57kr9WcyFk8Fe6nkxQpWzGm5 renamed to foo with overwriting
$ ipfs key list
self
$
```

^--- My 'foo' key has disappeared :( :(

### Suggested fix:
This is slightly more involved because I think that fixing this issue should also involve changing some existing existing behaviour.

go-ipfs currently responds in the following two ways when a key is renamed to the same name:
```
$ ipfs key rename foo foo
Error: key by that name already exists, refusing to overwrite
$ ipfs key rename --force foo foo
Key QmcaW6t1MpTBh46GvkaTUA57kr9WcyFk8Fe6nkxQpWzGm5 renamed to foo with overwriting
!!!HORRIBLE ERROR!!!!
$ 
```

I would suggest that the behaviour of a same-rename without --force should be changed, and that the bug be fixed, so that the following happens:
```
$ ipfs key rename foo foo
Key QmcaW6t1MpTBh46GvkaTUA57kr9WcyFk8Fe6nkxQpWzGm5 renamed to foo
$ ipfs key rename --force foo foo
Key QmcaW6t1MpTBh46GvkaTUA57kr9WcyFk8Fe6nkxQpWzGm5 renamed to foo
$ 
```

The reason that I think the non-force behaviour should be changed is because nothing is actually being overridden when a key is renamed to the same name.

### Fixing:
Once you agree on a solution, If its okay with you I would like to fix this myself and submit a PR. I have identified the function question at [core/coreapi/key.go#L134](https://github.com/ipfs/go-ipfs/blob/master/core/coreapi/key.go#L134) and begun to implement a fix.

rob-deutsch:
Corrects and amends what happens when keys are renamed to the same name.

Also mildly changed the order of imports in key.go because the VS Code Go linter wasn't quite happy with it.

Fixed #5450

### DIFF
--- a/core/coreapi/key.go
+++ b/core/coreapi/key.go
@@ -8,10 +8,10 @@ import (
 
 	coreiface "github.com/ipfs/go-ipfs/core/coreapi/interface"
 	caopts "github.com/ipfs/go-ipfs/core/coreapi/interface/options"
-	ipfspath "gx/ipfs/QmX7uSbkNz76yNwBhuwYwRbhihLnJqM73VTCjS3UMJud9A/go-path"
 
 	crypto "gx/ipfs/QmPvyPwuCgJ7pDmrKDxRtsScJgBaM5h4EpRL2qQJsmXf4n/go-libp2p-crypto"
 	peer "gx/ipfs/QmQsErDt8Qgw1XrsXf2BpEzDgGWtB1YLsTAARBup5b6B9W/go-libp2p-peer"
+	ipfspath "gx/ipfs/QmX7uSbkNz76yNwBhuwYwRbhihLnJqM73VTCjS3UMJud9A/go-path"
 )
 
 type KeyAPI CoreAPI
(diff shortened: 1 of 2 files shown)


rob-deutsch:
Drat! I submitted the wrong commit.

Closing this and will resubmit.

Issue #5378, opened 2018-08-13
RFC: Handling of CidV1/V0 at the bitswap level


kevina:
As of p.r. #5285 `go-ipfs` will be able to return a block regardless of it if is given in CidV0 or CidV1.  Soon, we will switch the blockstore to work with multihashes instead of CIDs.

However, for a period of time there will still be nodes that work at the CID level instead of the multihash level.  These older nodes are very likely to have most of there content is CidV0.  This can create a small problem when we start switch to CIDv1 at the UX level.  For example, if there is some content that was originally available at CidV0, but we display it to the user as a Base32 CidV1 and the user then copy and pastes the link and tries to retrieve it, they may not be able to retrieve the content if the only nodes that have it are the older nodes.

My question is, do we want to try to solve this at the bitswap level?  What I am thinking is that bitswap firsts tries to get the content using the original CID version (in the above example CIDv1), then after a brief timeout (maybe 5 seconds) it _also_ requests the CIDv0 version of the hash.  Once it retrieves either version it cancels both requests (if that is possible).

@Stebalien @whyrusleeping thoughts?  I don't know enough about bitswap yet to really give any meaningful solutions.

burdakovd:
One option is to internally query content in CIDv0 always (assuming old nodes keep content in CIDv0, and new nodes can handle both types of queries).

Then some time later - when most of the nodes update their software and can handle both types of queries, start querying directly in CIDv1.

kevina:
@burdakovd Unfortunately old nodes may still have content in CIDv1, so we really need to try both CIDv0 and CIDv1 to be sure content doesn't become unavailable.

kevina:
@Stebalien @diasdavid If I understood the discussion today the solution to this was to determine if the node we are talking supports retrieving the block regardless of the CID version.  If the node does not as a temporary measure we will simply request both the blocking using both CIDv0 and CIDv1.  This is dependent on the IPFS reporting the exact version.  I need to verify that is the case.

lidel:
Triage: iiuc this can be closed -- the problem described here no longer exists.

- PR #6816 switched the blockstore to store blocks by raw multihash, shipped in Kubo 0.12.0 (Feb 2022). With this change:
  - `blockstore.Get(ctx, CIDv0(sha2-256, H))` and `blockstore.Get(ctx, CIDv1(raw, sha2-256, H))` both look up the same datastore key via `MultihashToDsKey(H)`
[...]

PR #5464, opened 2018-09-14
WIP: Add global option to specify the multibase encoding.

kevina:
This is a redo of #5289.

It adds a `--cid-base` option for specifying the multibase.  It also provides a `--upgrade-cidv0` option to upgrade CidV0 to CidV1 for display.  Specifying `--cid-base` enables `--upgrade-cidv0` by default.

Unlike the last attempt, this does most of the conversion client side.  The exception to the rule is that as `Path`s are meant to be strings the server creates new paths with the correct base when a CID is part of the path.

When a use provides a hash on the command line, this p.r. attempts to preserve that string when returning information on the same hash unless `--cid-base` is used.  For example in `ipfs pin`.

The multibase of a CID, is now preserved in `ipfs resolve`.  Other commands will need to be reviewed on a case by case basis.

Not every command honers `--cid-base` yet. The implementation details may still change and a large part of the helper code will likely get moved to `go-cidutil`.


kevina:
@Stebalien this is still a WIP but I have a few question: Is it really okay to set a global variable when translating the Cid client side.  It does make things easier but I want be 100% sure the client code (in `core/commands`) won't be used as a library where this can cause a problem.  The alternative is to thread the state though to the point where the CID is translated.

Also note I have a whole bunch of changes not pushed yet and I changed a lot of the implementation details which included.  The general idea is still the same though.

kevina:
This depends on https://github.com/ipfs/go-cidutil/pull/8 and https://github.com/ipfs/go-mfs/pull/5

This should be ready for initial review.  I added support for as many commands as I could:

```
ipfs add -- client side
ipfs block -- skipping for now, will soon need to use raw multihashs
ipfs daemon -- ??? [1]
ipfs dag -- low level, skipping for now
ipfs dns -- ??? [2]
ipfs file ls -- ???
ipfs files ls -- client side
ipfs files stat -- client side
ipfs filestore dups -- server side due to the use of RefWrapper
ipfs filestore ls -- client side
ipfs filestore verify -- client side
ipfs name resolve -- ??? [2]
ipfs object -- low level, skipping for now
ipfs pin add -- client side [m!]
ipfs pin ls -- FIXME
ipfs pin rm -- client side [m!]
ipfs pin update -- TODO
ipfs pin verify -- client side
ipfs ref -- server side due to the use of RefWrapper [m]
ipfs refs local -- server side due to the use of RefWrapper
ipfs resolve -- server side [p]
ipfs tar add -- client side
ipfs urlstore add -- FIXME: Uses BlockStat

[m] Any cids identical to the ones passed in on the command line will
be returned using the same version and encoding.  Othercids will use
the default settings.  Setting --cid-base will override this behavior
and convert all cids.

[m!] These commands should likely have the same behavior as in [m],
but the arguments of the command are not always available in the
Marshalers, thus making it difficult to due client cide without
additional hacks.

[p] The cid passed in on the command line is used to influence the
encoding of any other cids unless --cid-base is used.

[1] should --cid-flags set defaults if launched with the daemon?
Probably not, better to use config settings.

[2] I am not 100% sure but I don't think CID's are ever explicitly
constructed, instead I think it is just returning the string found in
the lookup
```

alanshaw:
@kevina could you explain what the use case is for `--upgrade-cidv0`?

alanshaw:
@kevina does this PR also allow `?cid-base` to be passed to the equivalent HTTP API routes?

kevina:
> @kevina could you explain what the use case is for --upgrade-cidv0?

I am not sure `--upgrade-cidv0` is the best name (I might change it to `--display-cidv1` or something similar).

The idea is that when a user uses `--cid-base=base32` they expect the hashes to be in base32.  This is currently only possible with cidv1 so the CID is upgraded from version 0 to 1.  Once the blockstore (and hopefully bitswap see #5378) are converted this won't be a problem in most cases, but the binary representation is different, so it should be an option so the user can turn the conversion off if required.

kevina:
@kevina does this PR also allow ?cid-base to be passed to the equivalent HTTP API routes?

It allows it, but the JSON output of the request will remain unchanged except in the cases I marked as `server side` in the list mentioned in a comment above.

Stebalien:
Could you add some comments (especially godoc comments)?

kevina:
> Could you add some comments (especially godoc comments)?

Yeah that would be a good idea. :) I work on that tomorrow.  I am also unhappy with the choice of `--upgrade-cidv0` as we might need that option to actually upgrade Cid's for example when pinning.  I am thinking `--output-cidv1` or something similar as that is more direct and to the point.

kevina:
@Stebalien rebased and documentation added.

core/commands/cmdenv/cidbase.go
```diff
@@ -0,0 +1,119 @@
...
+	return *h.enc
+}
+
+// EncoderFromPath returns a a new Encoder that will format Cid like
+// the one in the path if the `--cid-base` option is not used.  (If
+// the `--cid-base` is used then a copy of the base encoder will be
+// returned.)  In particular: if the path contains a version 1 CID
+// then all CIDs will be outputting using the same multibase.  if the
+// path contains a version 0 CID then version 0 CIDs will be outputted
+// as is and version1 cids will use the multibase from the base
```
djdv: version1
->
version 1

core/commands/dns.go
```diff
@@ -54,7 +54,6 @@ The resolver can recursively resolve:
 		cmdkit.BoolOption("recursive", "r", "Resolve until the result is not a DNS link."),
 	},
 	Run: func(req cmds.Request, res cmds.Response) {
-
```
djdv: There's only 2 whitespace changes to this file.
Is this significant?

core/commands/cmdenv/cidbase.go
```diff
@@ -0,0 +1,119 @@
...
+// only be used on the client side and not the server side.
+func (h *CidBaseHandler) UseGlobal() *CidBaseHandler {
+	h.enc = &cidenc.Default
+	return h
+}
+
+// Proc processes the `--cid-base` and `--output-cidv1` options.  If
+// UseGlobal was enabled it will change the value of the global
+// default.
+func (h *CidBaseHandler) Proc() (*CidBaseHandler, error) {
```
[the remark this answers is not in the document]
kevina: That's not quite right.  But I did make some changes to this method.

kevina:
This needs to be redone, but is also blocked now on the switch to the new commands library of the affected commands.

Stebalien:
@kevina why is the switch to the new cmds lib blocking this? #5611 is a complex PR and I really don't want to have to rebase it unless we have to.

alanshaw:
@kevina should I be expecting to see base32 CIDs in the output here:

https://ipfs.io/api/v0/resolve?arg=/ipfs/QmTVyRqL8rdJzpCjZSjjw9bJ1M3ioWuyJhCLpYfgRMmzD4&cid-base=base32

kevina:
> @kevina should I be expecting to see base32 CIDs in the output here:

> https://ipfs.io/api/v0/resolve?arg=/ipfs/QmTVyRqL8rdJzpCjZSjjw9bJ1M3ioWuyJhCLpYfgRMmzD4&cid-base=base32

Yes.

PR #5466, opened 2018-09-14
make warnings on no bootstrap peers less noisy

Stebalien:
We only need to warn the user if we *can't* connect to the network.


core/bootstrap.go
```diff
@@ -141,6 +138,11 @@ func bootstrapRound(ctx context.Context, host host.Host, cfg BootstrapConfig) er
 	// if connected to all bootstrap peer candidates, exit
 	if len(notConnected) < 1 {
 		log.Debugf("%s no more bootstrap peers to create %d connections", id, numToDial)
+		if len(peers) == 0 {
+			// We're *need* to bootstrap but we have no bootstrap
```
schomatis: nit: "We're *need*" -> "We *need*"

Issue #5383, opened 2018-08-15
Graduate pubsub from being experimental


whyrusleeping:
It's probably about time to graduate pubsub out of the realm of just being an experiment.

We should still keep ipns pubsub as an experimental flag, but i think the default floodsub stuff should be enabled.

To do this, let's first have a separate pubsub doc in the docs folder that contains an expanded guide on using pubsub (beyond whats in the experimental features doc). And then just set the flags default value to true.

If the user isnt using any pubsub stuff, there shouldnt be any extra activity going on as a result.

Stebalien:
So, I have two major issues with the current system:

1. Unauthenticated "from" field in messages.
2. Unauthenticated sequence numbers.

Basically, it's embarrassingly insecure in ways that can't be fixed at the application layer (because users can be DoSed at the protocol layer). The DoS here is that an attacker can easily send messages "from" *other* peers to hide their messages.

I can see two ways forward.

# Content addressed

We can replace sequence numbers/from with a hash of the message (plus maybe timeout?). Users would have to put something new into a message to make it a "new" message.

We could also just take the from, data, and seqno and hash them together to get the "ID". However, the fact that the from isn't authenticated still bugs me. This would be a non-breaking way to fix floodsub but, unfortunately, it'd still break gossipsub.

# Sign them

We could authenticate these messages out of the box. Honestly, that's probably what most users would prefer. However, key distribution becomes a bit painful. When receiving a message, the user may have to go back to the sender and ask for the key.

Downsides:

* This would be a completely breaking change.
* Asking for the key could get complicated/slow (for bootstrapping). We could also include the key (or do some kind of optimistic send) but that's also tricky.
* Some systems may not care about authorship. Content addressing + validators allows this kind of thing to be pushed to a higher layer.

Upside:

* Many systems *will* care about authorship and applications will likely find authenticated messages useful. However, we can always provide a pubsub + auth layer to do this...

---

Personally, I'd prefer to just make pubsub2 (fork the code and protocol) and say "it's experimental".

We could then:

1. Use IPLD for messages.
2. Use CIDs for "message IDs".
3. Do validation at a higher layer.

whyrusleeping:
We should just implement signing as planned. Also remember EC keys sidesteps the key distribution problem. For RSA keys we could just force them to be included in the message. This would motivate users to upgrade to ed25519 asap

Stebalien:
Signing as planned works for me (but it will break existing clients).

[...]

PR #5467, opened 2018-09-14
enable pubsub function default

chunzhouyu:


vyzo:
We are not ready to do that yet -- see #5383.

And when we do, it will likely be with a config option instead of a daemon flag.

chunzhouyu:
okay

Issue #5470, opened 2018-09-14
Add support for raw leaves in `ipfs object diff`


schomatis:
See https://github.com/ipfs/go-ipfs/issues/4687#issuecomment-417511872.

overbool:
@schomatis I am fixing this issue

overbool:
@schomatis could you help me review my pr in https://github.com/ipfs/go-ipfs/pull/5472?

schomatis:
On it.

overbool:
@schomatis The `ipfs object` is not very uniform currently, because `ipfs object diff` support the `RawNode` but `ipfs object get` doesn't.

Is there any planning on refactoring these commands?

schomatis:
> Is there any planning on refactoring these commands?

Not really, they are going to be replaced by the `ipfs dag` commands, you could raise another issue to discuss adding support for raw nodes for the entire `ipfs object` command set, but I'm sure we could use your help with a lot of other more relevant issues. That being said, if you're interested in doing it as a way to learn more about how `go-ipfs` works under the hood go for it, just know that we don't plan on supporting these commands much longer (although time is largely a relative concept here :)

PR #5472, opened 2018-09-15
fix(object): add support for raw leaves in object diff

overbool:
License: MIT

Fixes https://github.com/ipfs/go-ipfs/issues/5470

### DIFF
--- a/dagutils/diff.go
+++ b/dagutils/diff.go
@@ -112,8 +112,8 @@ func Diff(ctx context.Context, ds ipld.DAGService, a, b ipld.Node) ([]*Change, e
 	}
 
 	var out []*Change
-	cleanA := a.Copy().(*dag.ProtoNode)
-	cleanB := b.Copy().(*dag.ProtoNode)
+	cleanA := a.Copy()
+	cleanB := b.Copy()
 
 	// strip out unchanged stuff
 	for _, lnk := range a.Links() {
@@ -132,17 +132,7 @@ func Diff(ctx context.Context, ds ipld.DAGService, a, b ipld.Node) ([]*Change, e
 					return nil, err
 				}
 
-				anodepb, ok := anode.(*dag.ProtoNode)
-				if !ok {
-					return nil, dag.ErrNotProtobuf
-				}
-
-				bnodepb, ok := bnode.(*dag.ProtoNode)
-				if !ok {
-					return nil, dag.ErrNotProtobuf
-				}
-
-				sub, err := Diff(ctx, ds, anodepb, bnodepb)
+				sub, err := Diff(ctx, ds, anode, bnode)
 				if err != nil {
 					return nil, err
 				}
@@ -152,8 +142,12 @@ func Diff(ctx context.Context, ds ipld.DAGService, a, b ipld.Node) ([]*Change, e
 					out = append(out, subc)
 				}
 			}
-			cleanA.RemoveNodeLink(l.Name)
-			cleanB.RemoveNodeLink(l.Name)
+			if cleanA, ok := cleanA.(*dag.ProtoNode); ok {
+				cleanA.RemoveNodeLink(l.Name)
+			}
+			if cleanB, ok := cleanB.(*dag.ProtoNode); ok {
+				cleanB.RemoveNodeLink(l.Name)
+			}
 		}
 	}
 
--- a/dagutils/diff.go
+++ b/dagutils/diff.go
@@ -95,29 +95,23 @@ func ApplyChange(ctx context.Context, ds ipld.DAGService, nd *dag.ProtoNode, cs 
 }
 
 // Diff returns a set of changes that transform node 'a' into node 'b'.
-// It supports two nodes forms: ProtoNode and RawNode. Because we treat
-// the nodes as IPLD nodes as long as possible and only convert them
-// to ProtoNode when necessary: when we need to remove links, and at that point
-// (if they have links to remove) we know they are not raw nodes.
+// It only traverses links in the following cases:
+// 1. two node's links number are greater than 0.
+// 2. both of two nodes are ProtoNode.
+// Otherwise, it compares the cid and emits a Mod change object.
 func Diff(ctx context.Context, ds ipld.DAGService, a, b ipld.Node) ([]*Change, error) {
 	// Base case where both nodes are leaves, just compare
 	// their CIDs.
 	if len(a.Links()) == 0 && len(b.Links()) == 0 {
-		if a.Cid().Equals(b.Cid()) {
-			return []*Change{}, nil
-		}
-		return []*Change{
-			{
-				Type:   Mod,
-				Before: a.Cid(),
-				After:  b.Cid(),
-			},
-		}, nil
+		return getChange(a, b)
 	}
 
 	var out []*Change
-	cleanA := a.Copy()
-	cleanB := b.Copy()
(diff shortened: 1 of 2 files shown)


overbool:
@Stebalien, something seems to be incorrect in ci

overbool:
@schomatis Yes, thx for your reply. I add the comment and test cases. Could you help me review this again?

core/commands/object/diff.go
```diff
@@ -6,11 +6,17 @@ import (
 	"io"
 
 	cmds "github.com/ipfs/go-ipfs/commands"
-	e "github.com/ipfs/go-ipfs/core/commands/e"
+	"github.com/ipfs/go-ipfs/core/commands/e"
 	coreiface "github.com/ipfs/go-ipfs/core/coreapi/interface"
 	"github.com/ipfs/go-ipfs/dagutils"
 
-	cmdkit "gx/ipfs/QmSP88ryZkHSRn1fnngAaV2Vcn63WUJzAavnRM9CVdU1Ky/go-ipfs-cmdkit"
```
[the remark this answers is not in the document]
overbool: OK

magik6k:
> @magik6k Could you take a look at the tests please?

It was a random fail, most likely caused by https://github.com/ipfs/go-ipfs/pull/5481

overbool:
@schomatis Thx for your suggestion. I had fixed them.

overbool:
@schomatis Should I split the test cases to keep the rule (one test_cmp call per case)?

schomatis:
> It was a random fail, most likely caused by #5481

@magik6k Sorry, I wasn't very clear. I meant if you could take a look at the `sharness` tests in this PR: https://github.com/ipfs/go-ipfs/pull/5472/files#diff-c85948db06f72554f9c3153c8cfac4c9, particularly if it's ok to call `test_cmp` in the same test case (which seems reasonable to me but just to be sure).

overbool:
@magik6k I had separated the test cases and modified the spelling mistake，could you help me review it again?

schomatis:
Nice work, thanks @overbool !

overbool:
@schomatis thx for your help in this issue

Stebalien:
How will this behave when passed, e.g., DagCBOR nodes? I think this'll just continue as if nothing is wrong and print out nonsensical results. We *should* fail when we encounter non-protobuf/non-raw nodes.

schomatis:
> I think this'll just continue as if nothing is wrong and print out nonsensical results.

Why?

> We _should_ fail when we encounter non-protobuf/non-raw nodes.

Where?

I don't have any experience with CBOR so I fail to see the relation between that type of node and this particular PR.

The title of the PR may be misleading, it emphasizes why we are doing this but not *what* we are changing here. This just doesn't force the `ProtoNode` type assertion unless it needs to: when modifying the links of placeholder nodes  (`cleanA`/`cleanB`) which shouldn't even exist in the first place (we could just be using another container for the `Add`/`Remove` links). There is no especial support for proto/raw nodes, this function operates at the IPLD level comparing CIDs.

overbool:
> However, this will fail for any format that's not ProtoNode as we won't be able to remove the links.

If other format Nodes have links and implement `RemoveNodeLink` function, maybe `Diff()` can support other format Nodes not just `ProtoNode` and `RawNode`.

Currently, `Diff()` just support `ProtoNode` and `RawNode`.

schomatis:
> This is how we're calculating the diff. We're removing all the links present in both nodes A and B from both A and B, leaving only the links present in only one or the other. However, this will fail for any format that's _not_ ProtoNode as we won't be able to remove the links.

Yes, it seems like a reasonable compromise. @overbool Are you ok with making that change? If not, we can merge this PR as-is, since it's already solving #5470 (and the fact that this command is deeply broken is beyond responsibility of this PR).

overbool:
> Are you ok with making that change?

@schomatis What do you mean about `change`?

> One solution is to just treat all non-ProtoNodes as opaque "things" (i.e., stop diffing). Diff would, at the top, check to see if we have two protonodes. If not, it would compare the CIDs and emit a Mod change object instead of actually diffing.

about this?

schomatis:
Yes, instead of returning an error if the conversion to a `ProtoNode` fails generate a `Mod` entry instead (without diffing inside of those node's links).

overbool:
@schomatis, i got it.
However, using new solution means that previous work and tests were wasted.

dagutils/diff.go
```diff
@@ -112,8 +116,8 @@ func Diff(ctx context.Context, ds ipld.DAGService, a, b ipld.Node) ([]*Change, e
 	}
```
schomatis: Not that much (I think), we could add a check here to verify that, if they have links, i.e., they are not raw nodes because they passed the first check (hence we would still be supporting them), they must be proto-nodes; if not, you could also return a `Mod` as before. Actually you could even merge this proto-node check in the previous `if`, WDYT?

overbool:
@Stebalien @schomatis According to your suggestion, I made some changes in `Diff()`.

It only traverses links in the following cases:

1. two node's links number are greater then 0.
1. both of two nodes are ProtoNode.

Otherwise, it compares the cid and emits a Mod change object.

dagutils/diff.go
```diff
@@ -94,26 +94,24 @@ func ApplyChange(ctx context.Context, ds ipld.DAGService, nd *dag.ProtoNode, cs
 	return e.Finalize(ctx, ds)
 }
 
-// Diff returns a set of changes that transform node 'a' into node 'b'
+// Diff returns a set of changes that transform node 'a' into node 'b'.
+// It only traverses links in the following cases:
+// 1. two node's links number are greater then 0.
```
schomatis: nit: "then" -> "than"

overbool: OMG, forgive my frequent spelling mistakes. 😨

test/sharness/t0052-object-diff.sh
```diff
@@ -36,18 +42,39 @@ test_expect_success "identity diff output looks good" '
...
+
 test_expect_success "diff against self (single file) is empty" '
-  ipfs object diff $SINGLE_FILE $SINGLE_FILE > diff_out
+  ipfs object diff $SINGLE_FILE $SINGLE_FILE > diff_out &&
   printf "" > diff_exp &&
   test_cmp diff_exp diff_out
 '
 
+test_expect_success "diff (raw-leaves) against self (single file) is empty" '
+  ipfs object diff $SINGLE_FILE_RAW $SINGLE_FILE_RAW > diff_raw_out
```
magik6k: missing `&&`

test/sharness/t0052-object-diff.sh
```diff
@@ -36,18 +42,39 @@ test_expect_success "identity diff output looks good" '
...
+'
+
 test_expect_success "diff against self (empty dir) is empty" '
   ipfs object diff $EMPTY_DIR $EMPTY_DIR > diff_out
   printf "" > diff_exp &&
   test_cmp diff_exp diff_out
 '
 
+test_expect_success "diff (raw-leaves) against self (empty dir) is empty" '
+  ipfs object diff $EMPTY_DIR_RAW $EMPTY_DIR_RAW > diff_raw_out
```
magik6k: also missing `&&`

overbool:
@magik6k thx for your reply, I had fixed it.

overbool:
@Stebalien  Was this pr what you had in mind?

PR #5476, opened 2018-09-16
fix fuse unmount test

### LINKED ISSUE #5475: TestExternalUnmount fails unmounting ipns
### Version information:
go-ipfs version: 0.4.18-dev-
Repo version: 7
System version: amd64/darwin
Golang version: go1.10.3

### Type:
Bug

### Description:
Reproduction steps:

````
~/go/src/github.com/ipfs/go-ipfs#  cd fuse/node/
~/go/src/github.com/ipfs/go-ipfs/fuse/node# go test
--- FAIL: TestExternalUnmount (0.09s)
	mount_test.go:83: exit status 1
FAIL
exit status 1
FAIL	github.com/ipfs/go-ipfs/fuse/node	0.150s
~/go/src/github.com/ipfs/go-ipfs/fuse/node# 
````

The reason it fails is the following process:
1. [mount_test.go#L45](https://github.com/ipfs/go-ipfs/blob/master/fuse/node/mount_test.go#L45) creates an offline node: `node, err := core.NewNode(context.Background(), nil)`
2. [mount_unix.go#L78](https://github.com/ipfs/go-ipfs/blob/master/fuse/node/mount_unix.go#L78) will only create the ipns node if its an online node: `if node.OnlineMode() {`
3. [mount_test.go#L90](https://github.com/ipfs/go-ipfs/blob/master/fuse/node/mount_test.go#L90) tries to unmount ipns: `err = node.Mounts.Ipns.Unmount()`
4. Of course, this attempt fails

Also, this test is written for Linux only, it won't work on macOS.

I'm working on a fix now

rob-deutsch:
Fuse unmount test uses ipfs instead of ipns, because offline nodes dont actually mount ipns.

Factored out GOOS-aware function to determine unmount command in fuse/mount.

fixed #5475

License: MIT
Signed-off-by: Rob Deutsch <rdeutschob@gmail.com>

### DIFF
--- a/fuse/mount/mount.go
+++ b/fuse/mount/mount.go
@@ -38,14 +38,9 @@ func ForceUnmount(m Mount) error {
 	point := m.MountPoint()
 	log.Warningf("Force-Unmounting %s...", point)
 
-	var cmd *exec.Cmd
-	switch runtime.GOOS {
-	case "darwin":
-		cmd = exec.Command("diskutil", "umount", "force", point)
-	case "linux":
-		cmd = exec.Command("fusermount", "-u", point)
-	default:
-		return fmt.Errorf("unmount: unimplemented")
+	cmd, err := UnmountCmd(point)
+	if err != nil {
+		return err
 	}
 
 	errc := make(chan error, 1)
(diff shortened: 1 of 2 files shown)


Stebalien:
Thanks!

Issue #5126, opened 2018-06-16
`ipfs add -r directory` doesn't upload files and directories with dot prefix


eykrehbein:
When I try to upload a directory including a `.config` file and a `.git` directory, IPFS uploads everything but those files and folders which are prefixed by a dot. I cannot imagine that this is a wanted feature, so I decided to open this issue.

Stebalien:
You can pass `-H` to include hidden files. I also can't imagine why we exclude hidden files by default.

Stebalien:
Unfortunately, fixing this would break too much. We'll have to reconsider in the next iteration of the API.

Stebalien:
(closing as we can't fix this right now)

iuriguilherme:
I vote against adding `.hidden` files by default as it is historical behavior, I rely on this behavior on automatic scripts which publish sites to IPNS. They shall not make it available via IPFS gateways stuff like **.htaccess** (bad) or **.env** (VERY bad).

I would happily change all my scripts but if such a change happen, there **SHOULD** be warnings everywhere so users don't make it "forever" available unwanted files.

PR #5479, opened 2018-09-17
add hidden file by default when add file recursively

kjzz:
fix bug #5126 

License: MIT

### DIFF
--- a/core/commands/add.go
+++ b/core/commands/add.go
@@ -118,7 +118,7 @@ You can now check what blocks have been created by:
 		cmdkit.BoolOption(onlyHashOptionName, "n", "Only chunk and hash - do not write to disk."),
 		cmdkit.BoolOption(wrapOptionName, "w", "Wrap files with a directory object."),
 		cmdkit.StringOption(stdinPathName, "Assign a name if the file source is stdin."),
-		cmdkit.BoolOption(hiddenOptionName, "H", "Include files that are hidden. Only takes effect on recursive add."),
+		cmdkit.BoolOption(hiddenOptionName, "H", "Include files that are hidden. Only takes effect on recursive add.").WithDefault(true),
 		cmdkit.StringOption(chunkerOptionName, "s", "Chunking algorithm, size-[bytes] or rabin-[min]-[avg]-[max]").WithDefault("size-262144"),
 		cmdkit.BoolOption(pinOptionName, "Pin this object when adding.").WithDefault(true),
 		cmdkit.BoolOption(rawLeavesOptionName, "Use raw blocks for leaf nodes. (experimental)"),
--- a/test/sharness/t0040-add-and-cat.sh
+++ b/test/sharness/t0040-add-and-cat.sh
@@ -620,7 +620,7 @@ test_add_cat_expensive "--cid-version=1" "zdj7WcatQrtuE4WMkS4XsfsMixuQN2po4irkYh
 # encoded with the blake2b-256 hash funtion
 test_add_cat_expensive '--hash=blake2b-256' "zDMZof1kwndounDzQCANUHjiE3zt1mPEgx7RE3JTHoZrRRa79xcv"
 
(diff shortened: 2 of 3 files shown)


kjzz:
hey @Stebalien , it has already paased required ci,please help me review it .Thx a lot.

kjzz:
hey @Stebalien  , it has already paased required ci,please help me review it .Thx a lot.

Stebalien:
We should get some discussion on this. I'd *prefer* to have done it this way from the very beginning *however*, this is a breaking change.

@ipfs/go-team thoughts? Objections?

magik6k:
Most unix commands such as `cp -r dirA dirB` or `tar -cvf` will include hidden files (shell expansion with `*` won't). It makes sense for me.
(we should ping js team too at some point)

bigs:
yeah i’m in favor of adding dotfiles by default

schomatis:
> Most unix commands such as `cp -r dirA dirB` or `tar -cvf` will include hidden files

:+1:

Stebalien:
@alanshaw ping.

kjzz:
And when we use command `git add`,it also will include hidden file.

hsanjuan:
This is a silently breaking change (the user won't we warned that something has changed). Things like the [go-ipfs-api](https://godoc.org/github.com/ipfs/go-ipfs-api#Shell.AddDir) will start behaving differently because they don't add the `hidden` parameter to the request and the hidden-file-filter happens server-side. People might have been adding files and using hidden folders as a way to exclude files that they don't want to add, or simply forgot about hidden files, potentially leading to security compromises. I'd hope this doesn't affect anyone but if it does, it's not very nice.

kjzz:
> People might have been adding files and using hidden folders as a way to exclude files that they don't want to add, or simply forgot about hidden files, potentially leading to security compromises.
Agree!So Can we add a file like `.gitignore` ,in ipfs it maybe names `.ipfsignore`? It will help user solve this solution.

djdv:
On OS X and Windows, they have specific file-attributes ("invisible" and "hidden") outside of just a dot in the filename. It's a bit more explicit and standard, but I don't know if there's something like that for other systems.
I think dotfiles should be added by default, but files with "hidden" attributes should not.
However, doing that without having those attributes on all platforms, would lock out the ability to skip files for users. So I don't know how best to handle determining what is and is not "hidden". This seems like the bigger problem to me. But likely out of scope.

No opinion on the decision otherwise.

kjzz:
Hey @Stebalien ,I know that `.ipfsignore` won't actually resolve @hsanjuan 's objection.
However,IMO,we could consider to add `.ipfsigonre` feature to help user hidde some file when we use command `-r`or we add directory.This may be a good user experience.WDYT?
And if you agree,i am willing to do it.

Stebalien:
This work was started in https://github.com/ipfs/go-ipfs/issues/3643 but I'm not entirely sure where it left off.

djdv:
After reading the conventions people have posted. I find it hard to believe that someone would have files they don't intend to share, in a target directory. Other tools don't seem to exclude them either, in the same case.
It seems that "hidden" rarely implies "excluded", they're only excluded in listings, like with `ls`.

That being said, this seems like just a configuration change, and I think that's a reasonable thing to do later, during a redesign, so long as we have some kind of clear statement of why something was changed, and how it might impact a user (security/privacy implications, etc.).

Kubuxu:
This will also breat gx as far as I understand as it will start adding `.git`.

alanshaw:
I’m +1 on adding hidden by default. I’m also pretty sure js-ipfs already does this :P

keks:
I just asked David and he says js-ipfs does what go-ipfs does. So probably not doing that by default.

vyzo:
I think that the breakage of applications from this change has to be seriously considered.
We can't quite merge it yet.

Stebalien:
Yeah, this looks like an API 2.0 thing.

alanshaw:
@keks yep, I re-checked and @diasdavid is correct - sorry for any confusion

PR #5480, opened 2018-09-17
fix(pin): add select for channel handling

overbool:
License: MIT

### DIFF
--- a/core/commands/pin.go
+++ b/core/commands/pin.go
@@ -100,7 +100,10 @@ var addPinCmd = &cmds.Command{
 		ch := make(chan pinResult, 1)
 		go func() {
 			added, err := corerepo.Pin(n, ctx, req.Arguments(), recursive)
-			ch <- pinResult{pins: added, err: err}
+			select {
+			case ch <- pinResult{pins: added, err: err}:
+			case <-ctx.Done():
+			}
 		}()
 
 		ticker := time.NewTicker(500 * time.Millisecond)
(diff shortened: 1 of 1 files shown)


core/commands/pin.go
```diff
@@ -115,12 +118,31 @@ var addPinCmd = &cmds.Command{
 				}
 
 				if pv := v.Value(); pv != 0 {
-					out <- &AddPinOutput{Progress: v.Value()}
+					select {
+					case out <- &AddPinOutput{Progress: v.Value()}:
+					case <-ctx.Done():
+						log.Error(ctx.Err())
```
magik6k: This will print on the daemon, pretty sure we don't want that

kevina: Agree.  Just return the error.

kevina:
> I'd wait for #5035 to land and just rewrite it to the new command lib / res.Emit instead of messing witch channels here

That seams like an overkill to me, but I am not against it.

core/commands/pin.go
```diff
@@ -100,7 +100,10 @@ var addPinCmd = &cmds.Command{
 		ch := make(chan pinResult, 1)
 		go func() {
 			added, err := corerepo.Pin(n, ctx, req.Arguments(), recursive)
-			ch <- pinResult{pins: added, err: err}
+			select {
+			case ch <- pinResult{pins: added, err: err}:
+			case <-ctx.Done():
```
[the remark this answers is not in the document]
overbool: @magik6k Although `ch` is a buffered `chan`, if the `select case` of receiving value from this `chan` returns or exits(like https://github.com/ipfs/go-ipfs/blob/9bf4e4145ea897ced2f9aa2087296b3beda6e941/core/commands/pin.go#L118-L132), this goroutine will be blocked.

PR #5481, opened 2018-09-17
coreapi name: Increase test swarm size

magik6k:
It was too small causing tests to fail occasionally, like in https://github.com/ipfs/go-ipfs/pull/5472


Stebalien:
Why does the swarm size matter?

magik6k:
DHT seems to have problems if there are < 5 peers, iirc this was an issue in other places too. I've never really dug deeper than that, but it might be worth having a look.

Stebalien:
That sounds like a bug. But yeah, we should just merge this and then fix it in go-ipfs (no point in killing CI).

PR #5484, opened 2018-09-17
resolve: use unixfs ResolveOnce

magik6k:
(This is assuming we want to make `ipfs resolve` operate on unixfs paths instead of ipld)

Fixes https://github.com/ipfs/go-ipfs/issues/5270#issuecomment-422112136

### DIFF
--- a/core/coreapi/interface/options/name.go
+++ b/core/coreapi/interface/options/name.go
@@ -2,6 +2,8 @@ package options
 
 import (
 	"time"
+
+	ropts "github.com/ipfs/go-ipfs/namesys/opts"
 )
 
 const (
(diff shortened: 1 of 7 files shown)


magik6k:
Good point, I'll see how that works

magik6k:
Done

core/coreapi/interface/options/name.go
```diff
@@ -40,9 +43,12 @@ func NamePublishOptions(opts ...NamePublishOption) (*NamePublishSettings, error)
...
 func NameResolveOptions(opts ...NameResolveOption) (*NameResolveSettings, error) {
 	options := &NameResolveSettings{
-		Recursive: false,
-		Local:     false,
-		Cache:     true,
+		Depth: 1,
+		Local: false,
+		Cache: true,
+
+		DhtRecordCount: 16,
```
Stebalien: Not sure if we should hard-code this. Can we use the value from the DHT (unless we go with the "resolve options" solution in which case this point is moot).

Stebalien:
Note: I'm fine keeping this as it is, I'd just like to consider a more generalized solution to adding more and more API options.

Stebalien:
@magik6k This is wonderful. These issues have been plaguing us for *ages*.

Issue #5491, opened 2018-09-19
Switch all path resolutions to the CoreAPI


Stebalien:
Currently, we tend to hard-code resolvers to use the unixfs resolver (or the IPLD resolver). Now that we have the CoreAPI, let's switch to using that. That will give us automatic support for *both* `/ipld/` and `/ipfs/` paths!

Stebalien:
@magik6k want to take this?

PR #5492, opened 2018-09-19
Cleanup instances of manual resolver construction

magik6k:
Part of https://github.com/ipfs/go-ipfs/issues/5491

Given how messy and all-over-the-place this PR is, I'd rather keep it small. This only cleans up places which constructed their own resolver. I'll take care of the other parts, which use `node.Resolver` in next PR


schomatis:
Could you provide me some pointers about this PR that would help me review it?

magik6k:
The idea (mentioned in https://github.com/ipfs/go-ipfs/issues/5491) is that we want to start using the coreapi resolver basically everywhere.

The resolver was discussed in https://github.com/ipfs/go-ipfs/issues/4666 / https://github.com/ipfs/go-ipfs/pull/4672, the TL;DR is that it decides which resolver to use based on the prefix, so that `/ipfs/` paths will always resolve with the unixfs resolver (handling hamt) and `/ipld/` with the simple resolver (which basically uses IPLD paths). It also handles `/ipns/`

core/coreapi/interface/util.go
```diff
@@ -1,10 +1,20 @@
...
 
 type Reader interface {
-	io.ReadSeeker
+	ReadSeekCloser
+	Size() uint64
+	CtxReadFull(context.Context, []byte) (int, error)
+}
+
+// A ReadSeekCloser implements interfaces to read, copy, seek and close.
+type ReadSeekCloser interface {
```
schomatis: This code seems to be repeating the interfaces already defined in `go-unixfs`:

https://github.com/ipfs/go-unixfs/blob/master/io/dagreader.go#L22-L37

(Or maybe it's the other way around and UnixFS is duplicating these ones, either way, is it possible for one to reference the other?)

magik6k: That will be doable once the coreapi interface is extracted, which I hope will happen fairly soon - for progress see https://github.com/ipfs/go-ipfs/issues/4498 (I want to have pubsub/swarm/unixfs done before extracting the interface)

PR #5496, opened 2018-09-20
fixed tests of raised fd limits

### LINKED ISSUE #5475: TestExternalUnmount fails unmounting ipns

### LINKED ISSUE #5495: cmd/ipfs/util max fd tests don't work on MacOS
### Version information:
go-ipfs version: 0.4.18-dev-
Repo version: 7
System version: amd64/darwin
Golang version: go1.10.3

### Type:
Bug

### Description:

````
$ go test
--- FAIL: TestManageInvalidNFds (0.00s)
        ulimit_test.go:25: Testing file descriptor invalidity
        ulimit_test.go:45: ManageFdLimit should return an error
Successfully raised file descriptor limit to 9223372036854767616.
FAIL
exit status 1
FAIL    github.com/ipfs/go-ipfs/cmd/ipfs/util   0.023s
$
````
This test can be found [here](https://github.com/ipfs/go-ipfs/blob/777d306f6e66e31a05f43a337bce272050407386/cmd/ipfs/util/ulimit_test.go#L24).

This test is trying to set the FD limit higher than your OS's maximum allowed. The problem is that on MacOS this maximum allowed is at the maximum int64 value (i.e. [unlimited](https://wilsonmar.github.io/maximum-limits/)), and we're experiencing integer wrap around.

### Solution:

I am working on a fix now. The fix is based on changing the util code to use uint64 instead of int64, and adding some error conditions to the FreeBSD code.

PR incoming.

### PR DESCRIPTION
**Pull request by rob-deutsch**
Raising FD limits was erroring when the OS's max was at the maximum signed integer value. Switched the code to using uint64 instead of int64.

Could someone please review, in particular, the FreeBSD parts? Integer checks are not my forte, and I don't have a FreeBSD machine to test on.

fixed #5495

Bonus: The `ManageFdLimit()` function contained a `fmt.Print` statement. That's really annoying because it was printing during tests. I've changed the `ManageFdLimit()` function so that it returns the new limit, and the calling function can do the fmt.Print statement (if it wishes).

### DIFF
--- a/cmd/ipfs/util/ulimit_freebsd.go
+++ b/cmd/ipfs/util/ulimit_freebsd.go
@@ -3,6 +3,9 @@
 package util
 
 import (
+	"errors"
+	"math"
+
 	unix "gx/ipfs/QmVGjyM9i2msKvLXwh9VosCTgP4mL91kC7hDmqnwTTx6Hu/sys/unix"
 )
 
@@ -12,16 +15,22 @@ func init() {
 	setLimit = freebsdSetLimit
 }
 
-func freebsdGetLimit() (int64, int64, error) {
+func freebsdGetLimit() (uint64, uint64, error) {
 	rlimit := unix.Rlimit{}
 	err := unix.Getrlimit(unix.RLIMIT_NOFILE, &rlimit)
-	return rlimit.Cur, rlimit.Max, err
+	if (rlimit.Cur < 0) || (rlimit.Max < 0) {
+		return 0, 0, errors.New("invalid rlimits")
+	}
+	return uint64(rlimit.Cur), uint64(rlimit.Max), err
 }
 
-func freebsdSetLimit(soft int64, max int64) error {
+func freebsdSetLimit(soft uint64, max uint64) error {
+	if (soft > math.MaxInt64) || (max > math.MaxInt64) {
+		return errors.New("invalid rlimits")
+	}
 	rlimit := unix.Rlimit{
-		Cur: soft,
-		Max: max,
+		Cur: int64(soft),
+		Max: int64(max),
 	}
 	return unix.Setrlimit(unix.RLIMIT_NOFILE, &rlimit)
 }
--- a/mk/util.mk
+++ b/mk/util.mk
@@ -20,6 +20,8 @@ SUPPORTED_PLATFORMS += linux-amd64
 SUPPORTED_PLATFORMS += darwin-386
 SUPPORTED_PLATFORMS += darwin-amd64
 
+SUPPORTED_PLATFORMS += freebsd-amd64
+
 space:=
 space+=
 comma:=,
--- a/cmd/ipfs/daemon.go
+++ b/cmd/ipfs/daemon.go
@@ -201,8 +201,12 @@ func daemonFunc(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment
 
 	managefd, _ := req.Options[adjustFDLimitKwd].(bool)
 	if managefd {
-		if err := utilmain.ManageFdLimit(); err != nil {
+		if changedFds, newFdsLimit, err := utilmain.ManageFdLimit(); err != nil {
 			log.Errorf("setting file descriptor limit: %s", err)
+		} else {
+			if changedFds {
+				fmt.Printf("Successfully raised file descriptor limit to %d.\n", newFdsLimit)
+			}
 		}
 	}
 
--- a/cmd/ipfs/util/ulimit.go
+++ b/cmd/ipfs/util/ulimit.go
@@ -45,19 +45,19 @@ func setMaxFds() {
 
 // ManageFdLimit raise the current max file descriptor count
 // of the process based on the IPFS_FD_MAX value
-func ManageFdLimit() error {
+func ManageFdLimit() (changed bool, newLimit uint64, err error) {
 	if !supportsFDManagement {
-		return nil
+		return false, 0, nil
 	}
 
 	setMaxFds()
 	soft, hard, err := getLimit()
 	if err != nil {
-		return err
+		return false, 0, err
 	}
 
 	if maxFds <= soft {
-		return nil
+		return false, 0, nil
 	}
 
 	// the soft limit is the value that the kernel enforces for the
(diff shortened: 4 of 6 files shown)

### CODE REVIEW

**Comment by magik6k** [comment]
Fails to cross-compile to FreeBSD: https://ci.ipfs.team/blue/organizations/jenkins/IPFS%2Fgo-ipfs/detail/PR-5496/3/pipeline/123

**Comment by rob-deutsch** [comment]
Thanks @magik6k . I pushed a new set of commits with the bug fixed, and an extra commit which adds FreeBSD compilation to the `make test` command. Why should Jenkins have all the fun? :P

Would still like someone to review the FreeBSD code. I've just amended it to make it compile, but would like someone more knowledgeable to check the integer handling.

**Comment by magik6k** [comment]
Sharness still fails, should be easy fix: https://ci.ipfs.team/blue/organizations/jenkins/IPFS%2Fgo-ipfs/detail/PR-5496/5/tests

```
> diff -u expected_daemon actual_daemon
--- expected_daemon 2018-09-20 12:31:42.345528549 +0000
+++ actual_daemon 2018-09-20 12:31:40.581579936 +0000
@@ -1,4 +1,5 @@
Initializing daemon...
+Successfully raised file descriptor limit to 0.
Swarm listening on /ip4/127.0.0.1/tcp/36394
Swarm listening on /p2p-circuit/ipfs/QmZkiWzie2TUr3KMTqXsRSBNxJ8xakWw1c5pqdwcsyLArT
Swarm announcing /ip4/127.0.0.1/tcp/36394
```

**Comment by rob-deutsch** [comment]
I'm a little perplexed. I don't get that line when I ran the daemon on my machine.

Its past midnight here, so I'm hoping things are clearer in the morning...

File: mk/util.mk [hunk 1]
```diff
@@ -20,6 +20,8 @@ SUPPORTED_PLATFORMS += linux-amd64
 SUPPORTED_PLATFORMS += darwin-386
 SUPPORTED_PLATFORMS += darwin-amd64
 
+SUPPORTED_PLATFORMS += freebsd-amd64
```
Stebalien: :+1: I believe we didn't have this before due to some reuseport issues (like usual...).

rob-deutsch: Thanks for commenting. I'm not sure how this relates to reuseport though. This just directs `make test` to see whether it can compile a binary for FreeBSD.

Stebalien: IIRC, compiling on FreeBSD was broken at the time these tests were added.

cmd/ipfs/util/ulimit_freebsd.go
```diff
@@ -3,6 +3,7 @@
 package util
 
 import (
+	"errors"
 	unix "gx/ipfs/QmVGjyM9i2msKvLXwh9VosCTgP4mL91kC7hDmqnwTTx6Hu/sys/unix"
```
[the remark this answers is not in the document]
rob-deutsch: Thanks for picking that up! Resubmitted below.

cmd/ipfs/util/ulimit_freebsd.go
```diff
@@ -12,16 +14,25 @@ func init() {
...
+	if (rlimit.Cur < 0) || (rlimit.Max < 0) {
+		return 0, 0, errors.New("Invalid rlimits")
+	}
+	return uint64(rlimit.Cur), uint64(rlimit.Max), err
 }
 
-func freebsdSetLimit(soft int64, max int64) error {
+func freebsdSetLimit(soft uint64, max uint64) error {
+	// Sourced from https://stackoverflow.com/questions/6878590/the-maximum-value-for-an-int-type-in-go
+	const maxUint64 = ^uint64(0)
```
Stebalien: These are now defined in the "math" package (e.g., `math.MaxUint64`).

cmd/ipfs/util/ulimit_freebsd.go
```diff
@@ -12,16 +14,25 @@ func init() {
 	setLimit = freebsdSetLimit
 }
 
-func freebsdGetLimit() (int64, int64, error) {
+func freebsdGetLimit() (uint64, uint64, error) {
 	rlimit := unix.Rlimit{}
 	err := unix.Getrlimit(unix.RLIMIT_NOFILE, &rlimit)
-	return rlimit.Cur, rlimit.Max, err
+	if (rlimit.Cur < 0) || (rlimit.Max < 0) {
+		return 0, 0, errors.New("Invalid rlimits")
```
Stebalien: By convention, this should be lowercase.

Stebalien: (see https://github.com/golang/go/wiki/CodeReviewComments#error-strings)

rob-deutsch:
Correct me if I'm wrong, but I think the code is clean and its failed due to intermittent Travis CI issues, right?

Stebalien:
(yes).

Thanks!

Issue #5497, opened 2018-09-20
Merge 'file ls' and 'ls' commands


magik6k:
`ipfs file ls` is this command which duplicates half of the normal `ls` functionality and missing the other half (plus seems to be a bit outdated).

IMO we should just remove it and merge it's functionality with the normal `ls`

(first commit which mentions it: https://github.com/ipfs/go-ipfs/commit/434871ba18662db366907e358654728a357f9c0e)

kjzz:
@magik6k if you did't finish this work , i am willing to help you do for this?WDYT?

overbool:
@magik6k currently, some `go-unixfs` functions are refactoring: https://github.com/ipfs/go-ipfs/issues/5217 and https://github.com/ipfs/go-unixfs/issues/3.

so, I think this issue could wait until the issues mentioned above are fixed. However, it is also a good choice to solve this issue firstly.

WDYT?

overbool:
@kjzz, I am fixing https://github.com/ipfs/go-unixfs/issues/3 and https://github.com/ipfs/go-ipfs/issues/5217, and noticed this issue at the beginning. But i forgot to ping u, sorry.

kjzz:
The reason why i did not fix it immediately is this featuren may cause breanking.Also the issue should influence `js-ipfs`.So this issue need to be considered carefully by offical team,

overbool:
@kjzz I don't have any experience with `js-ipfs`. If this issue will influence `js-ipfs`, maybe it should be handled with care.

overbool:
@magik6k it's better to fix https://github.com/ipfs/go-ipfs/issues/5502 firstly.

PR #5500, opened 2018-09-20
feat(command): merge file ls and ls

overbool:
License: MIT

Fixes: https://github.com/ipfs/go-ipfs/issues/5497 and https://github.com/ipfs/go-ipfs/issues/5502

### DIFF
--- a/core/commands/ls.go
+++ b/core/commands/ls.go
@@ -29,6 +29,8 @@ type LsLink struct {
 
 type LsObject struct {
 	Hash  string
+	Size  uint64
+	Type  string
 	Links []LsLink
 }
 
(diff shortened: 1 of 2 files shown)


core/commands/ls.go
```diff
@@ -48,6 +48,15 @@ the following format:
...
 The JSON output contains type information.
+The path can be a prefixless ref; in this case, we assume it to be an
+/ipfs ref and not /ipns.
+
+Example:
+
+    > ipfs file ls QmW2WQi7j6c7UgJTarActp7tDNikE4B2qXtFCfLPdsgaTQ
+    cat.jpg
+    > ipfs file ls /ipfs/QmW2WQi7j6c7UgJTarActp7tDNikE4B2qXtFCfLPdsgaTQ
+    cat.jpg
```
kevina: I think this information is overkill.  Syntax for paths should be documented somewhere else.

core/commands/ls.go
```diff
@@ -26,7 +26,7 @@ import (
 type LsLink struct {
 	Name, Hash string
 	Size       uint64
-	Type       unixfspb.Data_DataType
```
kevina: I am not sure I like the idea of replacing an enum type with a string....

overbool:
> We have to maintain API compatibilities so we can only add fields.

@Kubuxu Which modification do you mean?

Kubuxu:
As an example the switch from enum to string.

overbool:
@Kubuxu I had changed it.

overbool:
@Kubuxu I had changed it as your suggestion and updated the test.

PR #5501, opened 2018-09-20
Refactor UnixFS CoreAPI

magik6k:
This implements most options applying to single files supported by `ipfs add`.

TODO:
* [x] --pin, --only-hash
* [x] docstrings
* [x] option handling cleanup

To keep this PR manageable, I'm leaving multi-file / filestore / etc. for next PRs


kevina:
This seams like a overcomplicated way of going about it.  Was this discussed with any one before?

Stebalien:
@kevina which part?

It's using the functional options pattern (https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis) which we've started using pretty much everywhere. It allows us to easily extend these options.

kevina:
I will withhold judgement until I will have a closer look at the "functional options pattern" link provided.  (It just seams to me that it is adding a lot of not necessary useful indirection.)

core/coreapi/interface/options/unixfs.go
```diff
@@ -0,0 +1,187 @@
...
+		settings.RawLeavesSet = true
+		return nil
+	}
+}
+
+// InlineLimit sets the amount of bytes below which blocks will be encoded
+// directly into CID instead of being stored and addressed by it's hash
+//
+// Note that while there is no hard limit on the number of bytes here, it should
+// be kept at something reasonably low like 32b (default for 'ipfs add')
```
kevina: The wording here makes it sounds like the value should be 32 bytes or smaller.  That is misleading.  Values less than 32 bytes should not be encouraged.  The default of 32 was chosen so that when the Inline option is enabled it won't increase the size of the hashes, but a limit of 32 bytes is not very usefel.  A more reasonable upper limit would be 64 bytes, but if you are not concerned with displaying it than values like 128 or even 256 are also okay and based on my informal testing do not cause problems anywhere in IPFS.

core/coreapi/interface/options/unixfs.go
```diff
@@ -0,0 +1,187 @@
...
+const (
+	BalancedLayout Layout = iota
+	TrickleLayout
+)
+
+type UnixfsAddSettings struct {
+	CidVersion int
+	MhType     uint64
+
+	InlineLimit  int
```
kevina: A size 0 as an inline limit is useful.  You should have a separate option for enabling inlining and one for setting the limit.

core/coreapi/unixfs_test.go
```diff
@@ -38,7 +40,6 @@ var hello = "/ipfs/QmQy2Dw4Wk7rdJKjThjYXzfFJNaRKRHhHP5gHHXroJMYxk"
 var helloStr = "hello, world!"
 
 // `echo -n | ipfs add`
-var emptyFile = "/ipfs/QmbFMke1KXqnYyBBWxB74N4c5SBnJMVAiMNRcGu6x1AwQH"
```
kevina: Is there a reason for removing this?

core/coreapi/interface/options/unixfs.go
kevina: It is also far more user friendly to just be able to enable inlining and have it use the default value,

kevina:
> It's using the functional options pattern

Okay.  That is actually a pretty good idea and I take back my earlier comments.

My personal preference is to complete the transformation and convert the add command to use the new code.  This way we can be sure we don't miss anything, but I can understand if we don't want to do that all at once.

core/coreapi/interface/options/unixfs.go
```diff
@@ -124,13 +126,26 @@ func (unixfsOpts) RawLeaves(enable bool) UnixfsAddOption {
...
+// directly into CID instead of being stored and addressed by it's hash.
+// Specifying this option implicitly enables InlineBlocks
+// Default: 32 bytes
 //
-// Note that while there is no hard limit on the number of bytes here, it should
-// be kept at something reasonably low like 32b (default for 'ipfs add')
+// Note that while there is no hard limit on the number of bytes, it should
+// be kept at a reasonably low value, like 64 bytes if you intend to display
+// these hashes. Larger values like 256 bytes will work fine, but may affect
+// de-duplication of smaller blocks
```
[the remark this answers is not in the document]
magik6k: Maybe the wording isn't quite clear. The idea is that when you have many duplicated small blocks (100s of bytes), and a large inline limit, you can end up with quite a bit bigger DAG (i.e. the cumulative block size of the dag would be bigger).

If this is wrong / not clear enough, can you suggest better wording here?

core/coreapi/interface/options/unixfs.go
```diff
@@ -124,13 +126,26 @@ func (unixfsOpts) RawLeaves(enable bool) UnixfsAddOption {
...
 func (unixfsOpts) InlineLimit(limit int) UnixfsAddOption {
 	return func(settings *UnixfsAddSettings) error {
+		settings.InlineBlocks = true
```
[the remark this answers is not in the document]
magik6k: On the other hand, what are the cases somebody (other than in the command implementation) would set this option and then not want to inline blocks? It feels a bit like saying 'I want to inline blocks up to 123 bytes and I want to inline blocks'

core/coreapi/interface/options/unixfs.go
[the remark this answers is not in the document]
kevina: I didn't think of that.

My main concern with larger blocks is if they get too large they might cause a problem _somewhere_ even if they work find 98% of the time.

core/coreapi/interface/options/unixfs.go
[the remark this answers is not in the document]
Stebalien: A user might want to change the defaults. That is, I might want to create a set of default options that change the *default* inline limit without actually enabling inlining (unless requested by the user).

Note: I don't actually care one way or the other about this.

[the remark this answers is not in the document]
magik6k: Well, yeah. The options case makes enough sense, I'll go with that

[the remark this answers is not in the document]
kevina: Yeah, I was thinking along the same line as @Stebalien.  Also, since I assume the idea is to switch the `ipfs add` command to use the new API it best to keep the behavior constant.

core/coreapi/interface/options/unixfs.go
```diff
@@ -0,0 +1,204 @@
...
+// Specifying this option won't enable block inlining. For that use `Inline`
+// option. Default: 32 bytes
+//
+// Note that while there is no hard limit on the number of bytes, it should
+// be kept at a reasonably low value, like 64 bytes if you intend to display
+// these hashes. Larger values like 256 bytes will work fine, but may affect
+// de-duplication of smaller blocks.
+//
+// Setting this value too high may cause various problems, such as render some
+// blocks unfetchable
```
kevina: I think we are saying too much here and at this point will just be confusing the user. I would just simplify this with
> Note that while there is no hard limit on the number of bytes, it should be kept at a reasonably low value, such as 64 and no more then 1k.  Setting this value too high may cause various problems, such as render some blocks unfetchable.

@Stebalien thoughts?

kevina:
@Stebalien do we want to get this in as is.  I personally would rather we wait until all the options are implemented and the `ipfs add` command is converted.  This was we get the benefit of all the sharness tests to make sure here are no bugs or changes in behavior?

magik6k:
I don't really know, it kind of depends on how well `go-ipfs-cmdkit/files` stuff will fit here, I just didn't fiddle with that yet, so it might be anywhere from a few small commits to a few-thousand-diff-lines monster (I'm quite positive that this will be on the manageable side).

I'm fine with going either way, just want to make sure this doesn't get frozen because it grew too big.

kevina:
@magik6k I would attempt to do the complete conversion.  If it becomes unmanageable then I would split out parts into a separate p.r.

I would rather not have lots of duplicate code in `go-ipfs` if this commit lands and then `ipfs add` doesn't get converted for some reason.

Stebalien:
> I'm fine with going either way, just want to make sure this doesn't get frozen because it grew too big.

We can also do that as a followup PR against this one.

magik6k:
Looking at it now, it shouldn't be that bad in one PR (the followup PR would be replacing a large chunk of this PR anyways, this way there will be less reviewing and less PR maintenance/rebasing for me).

core/coreapi/interface/options/unixfs.go
```diff
@@ -170,13 +170,10 @@ func (unixfsOpts) Inline(enable bool) UnixfsAddOption {
...
 //
-// Note that while there is no hard limit on the number of bytes, it should
-// be kept at a reasonably low value, like 64 bytes if you intend to display
-// these hashes. Larger values like 256 bytes will work fine, but may affect
-// de-duplication of smaller blocks.
-//
-// Setting this value too high may cause various problems, such as render some
-// blocks unfetchable
+// Note that while there is no hard limit on the number of bytes, it should be
+// kept at a reasonably low value, such as 64 and no more than 1k. Setting this
```
kevina: @Stebalien should we say 1k here or use a lower value?

kevina: Okay.  I was actually trying create too strong a suggestion.  But I am okay with.

```
// Note that while there is no hard limit on the number of bytes, it should be
// kept at a reasonably low value, such as 64; implementations may choose to
// reject anything larger.
```
And just leave it at that.

Stebalien:
:sweat_smile: merged before we get any new conflicts.

Issue #5217, opened 2018-07-12
Avoid use of `pb.Data`


schomatis:
Avoid direct manipulation of the UnixFS format (`pb.Data`) and use the `FSNode` abstraction instead.

Replacing all `FromBytes()` calls (and removing the function itself),

https://github.com/ipfs/go-ipfs/blob/a9efa7e201707bd1a0d48fe1146a76ec0097db92/unixfs/unixfs.go#L32-L40

with `FSNodeFromBytes()` should be a good first step,

https://github.com/ipfs/go-ipfs/blob/a9efa7e201707bd1a0d48fe1146a76ec0097db92/unixfs/unixfs.go#L153-L162

overbool:
@schomatis can i help you do for it?

schomatis:
Yes, go ahead!

kevina:
Offhand this sounds like a good idea. But I would need to see the P.R. before I am able to give any opinion.  Refactors like this are hard for me to visualize without just doing it.  Also a p.r. makes sure we don't miss any corner cases.

PR #5505, opened 2018-09-21
fix(unixfs): issue #5217 (Avoid use of `pb.Data`)

overbool:
License: MIT

Fixes: https://github.com/ipfs/go-ipfs/issues/5217

### DIFF
--- a/core/commands/files.go
+++ b/core/commands/files.go
@@ -223,25 +223,25 @@ func statNode(nd ipld.Node) (*statOutput, error) {
 
 	switch n := nd.(type) {
 	case *dag.ProtoNode:
-		d, err := ft.FromBytes(n.Data())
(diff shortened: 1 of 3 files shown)


kevina:
@overbool what about removing the `FromBytes` function as described in the original function.

overbool:
@kevina 
> @overbool what about removing the `FromBytes` function as described in the original function.

I had added description in original function.

kevina:
I'm sorry, I misspoken.  Is it possible to just delete the `FromBytes` function now as described in the original _issue_ (#5217)?

The added commented are not really necessary and in my opinion are just adding noise.  Others may disagree with me though.

overbool:
@kevina @schomatis I had done a similar replacement for  here https://github.com/ipfs/go-unixfs/pull/18 and https://github.com/ipfs/go-mfs/pull/7

PR #5506, opened 2018-09-21
add pin lock in AddallPin function

### LINKED ISSUE #4561: The add command's addAllAndPin function doesn't hold the pin lock
While the file adder *does* hold the pin lock while adding a single file, the outer `addAllAndPin` function doesn't. This means that we can GC between adding and pinning.

kjzz:
fix #4561 

License: MIT

### DIFF
--- a/core/commands/add.go
+++ b/core/commands/add.go
@@ -270,13 +270,12 @@ You can now check what blocks have been created by:
 		fileAdder.Hidden = hidden
 		fileAdder.Trickle = trickle
 		fileAdder.Wrap = wrap
-		fileAdder.Pin = dopin
+		fileAdder.Pin = dopin && !hash
 		fileAdder.Silent = silent
 		fileAdder.RawLeaves = rawblks
 		fileAdder.NoCopy = nocopy
 		fileAdder.Name = pathName
 		fileAdder.CidBuilder = prefix
-		fileAdder.Hash = hash
 
 		if inline {
 			fileAdder.CidBuilder = cidutil.InlineBuilder{
--- a/core/commands/add.go
+++ b/core/commands/add.go
@@ -276,6 +276,7 @@ You can now check what blocks have been created by:
 		fileAdder.NoCopy = nocopy
 		fileAdder.Name = pathName
 		fileAdder.CidBuilder = prefix
+		fileAdder.Hash = hash
 
 		if inline {
 			fileAdder.CidBuilder = cidutil.InlineBuilder{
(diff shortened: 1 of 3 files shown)


core/coreunix/add.go
```diff
@@ -10,19 +10,19 @@ import (
 	"path/filepath"
 	"strconv"
 
-	core "github.com/ipfs/go-ipfs/core"
+	"github.com/ipfs/go-ipfs/core"
```
kevina: Is there a reason for these changes?  They are adding a lot of noise?

kjzz: Yeah,actually,it need not to rename the package name.The default name is also `core`,and renamed maybe redundant.

magik6k: There was a (short) discussion on this somewhere, the conclusion IIRC was that for consistency in some places we may actually prefer to have those labels. (some IDEs such as Goland do this by default, you can disable this in `Settings > Go > Imports > Optimize imports on the fly`)

kjzz:
@kevina @magik6k Thx for advice. I have already update my pr.Please help me review again

core/commands/add.go
```diff
@@ -276,6 +276,7 @@ You can now check what blocks have been created by:
 		fileAdder.NoCopy = nocopy
 		fileAdder.Name = pathName
 		fileAdder.CidBuilder = prefix
+		fileAdder.Hash = hash
```
magik6k: Shouldn't this just set `fileAdder.Pin` to false?

kjzz: why should put `fileAdder.Pin` to false? The reason i add `fileAdder.Hash` is that i want to get this param in `coreunix/add.go`.

magik6k: I know, but we don't really need this param there as all it does in disable pinning. See https://github.com/ipfs/go-ipfs/blob/b789295580d2ede0df6e248d73a1a7fb60b253ba/core/coreunix/add.go#L171-L178

We should just do `fileAdder.Pin = dopin && hash`

kjzz:
@magik6k  Thx for advice, i have update my pr as you suggested.Please help review again.

core/commands/add.go
kjzz: And maybe we should do `fileAdder.Pin = dopin && !hash`,because `hash` is mean not to pin fule.

core/commands/add.go
```diff
@@ -270,7 +270,7 @@ You can now check what blocks have been created by:
 		fileAdder.Hidden = hidden
 		fileAdder.Trickle = trickle
 		fileAdder.Wrap = wrap
-		fileAdder.Pin = dopin
+		fileAdder.Pin = dopin && !hash
```
kjzz: And maybe we should do `fileAdder.Pin = dopin && !hash`,because `hash` is mean not to pin file.

kjzz:
@magik6k there are some problem about ci.Can u help me restart it and reviwe my pr again?Thx a lot

magik6k:
(Note that you should be able to re-trigger the ci with `git commit --amend --no-edit; git push --force`)

kjzz:
> (Note that you should be able to re-trigger the ci with `git commit --amend --no-edit; git push --force`)

Thx a lot.

kjzz:
Hey @magik6k , sorry
> (Note that you should be able to re-trigger the ci with `git commit --amend --no-edit; git push --force`)
Because the ci has some error,so i do the commabd `git commit --amend --no-edit; git push --force`.But it hasdismissed your stale review.Please help me review again.very sorry .

kjzz:
hey @magik6k , if you have anytime. please help me review it again . I have performed a misoperation to dimiss your approved.Very sorry.

magik6k:
> the new commit has dismissed your stale review

It used to work fine some time ago. Ah well.

Stebalien:
> It used to work fine some time ago. Ah well.

I enabled that feature but yeah, it's too annoying (now disabled again). I wish there were a less pedantic version (compares the actual *diff*, not just the commits).

PR #5509, opened 2018-09-21
Add sessions when fetching MerkleDAG in LS

hannahhoward:
# Goals

Improve bandwidth usage for LS by adding Sessions

# Implementation

This is simply the code by @whyrusleeping referenced here -- https://github.com/ipfs/go-ipfs/compare/feat/ls-session?expand=1 - in #4979 . Regardless of the ultimate solution we end up using for HAMT directory speedup (#4908 ), is there any reason this can't be merged now? This seems like a tangible improvement, if not in speed at least in bandwidth usage

### DIFF
--- a/core/commands/ls.go
+++ b/core/commands/ls.go
@@ -108,9 +108,11 @@ The JSON output contains type information.
 		}
 
 		output := make([]LsObject, len(req.Arguments()))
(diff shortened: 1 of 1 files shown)


Stebalien:
Looks like a completely reasonable improvement.

PR #5514, opened 2018-09-24
[RFC] implement systemd socket activation

Stebalien:
It works but it could probably be prettier. Thoughts?

Basically, this patch checks for sockets passed via systemd socket-activation and, if present, ignores all other configured listen addresses and uses those. Currently, it only supports socket activation of the gateway and the API.

Motivation: this allows me to auto-start IPFS when I actually need it (but not until then).

### DIFF
--- a/misc/systemd/ipfs-api.socket
+++ b/misc/systemd/ipfs-api.socket
@@ -1,0 +1,15 @@
+# Enabling this will *completely override* any API listeners configured in your
+# config.
+
+[Unit]
+Description=IPFS API Socket
+
+[Socket]
+Service=ipfs.service
+FileDescriptorName=io.ipfs.api
+BindIPv6Only=true
+ListenStream=127.0.0.1:5001
+ListenStream=[::1]:5001
+
+[Install]
+WantedBy=sockets.target
--- a/misc/systemd/ipfs.service
+++ b/misc/systemd/ipfs.service
@@ -1,0 +1,9 @@
+[Unit]
+Description=IPFS Daemon
+
+[Service]
+ExecStart=/usr/bin/ipfs daemon --init --migrate
+KillSignal=SIGINT
+
+[Install]
+WantedBy=default.target
--- a/cmd/ipfs/daemon.go
+++ b/cmd/ipfs/daemon.go
@@ -13,6 +13,7 @@ import (
 	"sync"
 
 	version "github.com/ipfs/go-ipfs"
+	sockets "github.com/ipfs/go-ipfs/cmd/ipfs/sockets"
 	utilmain "github.com/ipfs/go-ipfs/cmd/ipfs/util"
 	oldcmds "github.com/ipfs/go-ipfs/commands"
 	"github.com/ipfs/go-ipfs/core"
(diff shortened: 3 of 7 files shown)


misc/systemd/ipfs-api.socket
```diff
@@ -0,0 +1,15 @@
...
+# config.
+
+[Unit]
+Description=IPFS API Socket
+
+[Socket]
+Service=ipfs.service
+FileDescriptorName=io.ipfs.api
+BindIPv6Only=true
+ListenStream=127.0.0.1:5001
```
lanzafame: It would be great if the port values were retrieved from the `~/.ipfs/config` or even better `~/.config/ipfs/config` or some suitable [configuration directory](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#RuntimeDirectory=). But whatever the solution, this will be confusing for users, i.e. enable systemd sockets, at some point in the future want to change port values, update ~/.ipfs/config, does nothing, restart sockets, still does nothing, google it/remembers to change in systemd files.

misc/systemd/ipfs.service
```diff
@@ -0,0 +1,9 @@
+[Unit]
```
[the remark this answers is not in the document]
Stebalien: I just put these here as starting points. The one *I* use is:

```ini
[Unit]
Description=IPFS Daemon

[Service]
User=ipfs
Group=ipfs
NoNewPrivileges=true
ProtectHome=true
ProtectSystem=full
PrivateTmp=true
PrivateDevices=true
SystemCallArchitectures=native
MemoryDenyWriteExecute=true
Environment=IPFS_PATH=/var/lib/ipfs
Environment=HOME=/var/lib/ipfs
ExecStart=/usr/bin/ipfs daemon --routing=dhtclient --migrate=true
KillSignal=SIGINT

[Install]
WantedBy=default.target
Also=ipfs-api.socket
Also=ipfs-gateway.socket
```

misc/systemd/ipfs-api.socket
Stebalien: Yeah but there's really nothing we can do about that. systemd doesn't let one read a config file.

Well... technically I can write a generator and *generate* these socket files dynamically. However, that probably isn't worth it.

---

We could, alternatively, listen on the union of addresses.

lanzafame: You could use an [`EnvironmentFile`](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#EnvironmentFile=)?

lgierth:
I like the use case -- but how about a multiaddr? :) Could be systemd-specific, or something more general about file descriptors.

lanzafame:
@lgierth wouldn't that require `systemd` to understand what a multiaddr is?

lgierth:
The systemd unit wouldn't change, but you'd have go-ipfs explicitly configured to listen on e.g. `/fd/42` or `/systemd/fd/io.ipfs.go`.

lanzafame:
> The systemd unit wouldn't change, but you'd have go-ipfs explicitly configured to listen on e.g. `/fd/42` or `/systemd/fd/io.ipfs.go`.

Sorry, I am only two coffees into my day, would `/systemd/fd/io.ipfs.go` be passed to which service file directive?

lgierth:
> would /systemd/fd/io.ipfs.go be passed to which service file directive?

It would be the same as this PR, except instead of just optimistically looking for systemd file descriptors, go-ipfs would have e.g. `ipfs config Addresses.Gateway /systemd/io.ipfs.api` set. We wouldn't add a new way of overriding the go-ipfs config, and we'd have something that'd also work for transports later (e.g. `/systemd/fd/io.ipfs.ws/tcp/443/wss`)

Stebalien:
This was a random "let's try this PR". I'll try to revive it at some future point but I'll close it for now.

PR #5516, opened 2018-09-24
[WIP] Swift Datastore

magik6k:
Based on https://github.com/ipfs/go-ipfs/pull/5187

This is a plugin for the swift datastore (which is like Amazon S3 but with more/different providers) - https://github.com/ipfs/go-ds-swift

### DIFF
--- a/plugin/datastore.go
+++ b/plugin/datastore.go
@@ -1,0 +1,14 @@
+package plugin
+
+import (
+	"github.com/ipfs/go-ipfs/repo/fsrepo"
+)
+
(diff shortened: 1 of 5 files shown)


Stebalien:
@magik6k was the resolution here to put these plugins in a separate repo (and document them)?

lidel:
Closing for the similar reason as https://github.com/ipfs/kubo/pull/5561#issuecomment-3563545683

Plugin lives in https://github.com/ipfs/go-ds-swift and if anyone thinks it should be included by default, please fill an issue to discuss the viability and long term maintenance first.

Issue #5341, opened 2018-08-03
Feature: Transparent HTTP proxy over p2p streams


ianopolous:
After talking in depth with @Stebalien about what I was hoping for from #5256 we concluded that my idea is orthogonal to what @haadcode is requesting, hence this issue.

**The idea is a transparent http proxy in the http api which pipes http requests over a p2p stream to another node.** 

E.g. the http api would include a new endpoint like
/api/v0/http_proxy/$name/$multiaddr/$http_path

$multiaddr would be something like /ipfs/$nodeid
The $http path is the path that is sent to the target server
@Stebalien could you explain the use of $name?

The cool thing about this is it would give ANY existing web app (including dynamic stuff like talking to a database etc.) a way to move their hosting from dns location based addressing to ipfs based trivially and everything would just work. (They would still run their back end as is but with an ipfs node proxying the http connections)

We would like to use this in @Peergos to allows us to tunnel writes (et al) to particular storage servers. 

Note that this automatically includes websockets.

Stebalien:
The endpoint would probably be `/http/$peer_id/$name/$http_path` as this will probably be *separate* from the HTTP API. We'd likely even provide a config option for mounting it on a separate port (but allow it to reside on the same host/port as the HTTP API).

In that path:

* $peer_id tells IPFS to open a stream to the peer at $peer_id.
* $name tells IPFS to use negotiate the `/x/http/$name` protocol when opening the stream. Note: `/x/` is the new protocol prefix we're using for `ipfs p2p` protocols (to avoid conflicting with the `/p2p` multiaddr protocol name.
* `$http_path` is the path to the HTTP resource.

We could (possibly later) allow the client to specify multiaddrs for the target peer in HTTP headers.

---

@ianopolous when we discussed this, we talked about using a *multiaddr* instead of just a peer ID in the URL. Unfortunately, it turns out that that would introduce an ambiguity. For example, in `/http/ipfs/QmRelay/p2p-circuit/ipfs/QmTarget/my_protocol/foo/bar`, the multiaddr is `/ipfs/QmRelay/p2p-circuit/ipfs/QmTarget`.

Additionally, putting a full multiaddr in the GET path doesn't allow specifying more than one multiaddr.

---

To be explicit, when the IPFS node receives a request of the form:

```
$METHOD /http/$peer_id/$name/$http_path
...
```

The IPFS node would:

1. Open a stream to $peer_id.
2. Negotiate the `/x/http/$name` protocol.
[...]

Issue #5521, opened 2018-09-25
Switch to the new iptb


Stebalien:
We have a new iptb version with a changed API. Somebody needs to go through and update all invocations of `iptb` in our test cases to use this new API.

Stebalien:
@ipfs/go-team who wants this? It's not exactly a glorious task but we need to get it done.

bigs:
commenting to follow this issue, as i'll be doing similar stuff for libp2p

Stebalien:
@travisperson I know you're busy, but could you make this happen? We have to cut an off-master point release of iptb every time we release because we're stuck on 1.0.

travisperson:
@Stebalien Ya I can update this. I have a branch that has most of the sharness tests updated already from a while back. Let me rebase it, update anything new and I'll open a PR.

Stebalien:
Awesome! Doesn't that depend on https://github.com/ipfs/iptb/pull/91?

travisperson:
Yes. Down the road I want to move the plugins out of the main iptb project. At the moment thought that is a bit difficult due to packages conflicting due to gx. That should be fixed this quarter though so for now they will stay there.

travisperson:
Fixed in https://github.com/ipfs/go-ipfs/pull/5636

PR #5526, opened 2018-09-26
[http_proxy_over_p2p]

cboddy:
This implements an http-proxy over p2p-streams, for context see https://github.com/ipfs/go-ipfs/issues/5341.

This script is a useful test of the functionality.  In case it causes portability issues I've not included it as a sharness test (since it uses python to serve HTTP content, although happy to add it since python be ~ as available as bash). 



(inline since GH doesn't support ``*.sh`` as an attachment)

```
#!/bin/bash                                                                                                                                                   
                                                                                                                                                              
#                                                                                                                                                             
# clean up all the things started in this script                                                                                                              
#                                                                                                                                                             
function teardown() {                                                                                                                                         
    jobs -p | xargs kill -9 ;                                                                                                                                 
}                                                                                                                                                             
trap teardown INT EXIT                                                                                                                                        
                                                                                                                                                              
#                                                                                                                                                             
# serve the thing over HTTP                                                                                                                                   
#                                                                                                                                                             
SERVE_PATH=$(mktemp -d)                                                                                                                                       
echo "YOU ARE THE CHAMPION MY FRIEND" > $SERVE_PATH/index.txt                                                                                                 
cd $SERVE_PATH                                                                                                                                                
# serve this on port 8000
python -m SimpleHTTPServer 8000 &


cd -

IPFS=cmd/ipfs/ipfs

PATH1=$(mktemp -d)
PATH2=$(mktemp -d)

RECEIVER_LOG=$PATH1/log.log
SENDER_LOG=$PATH2/log.log

export IPFS_PATH=$PATH1

#
# start RECEIVER IPFS daemon
#
$IPFS init >> $RECEIVER_LOG 2>&1
$IPFS config --json Experimental.Libp2pStreamMounting true >> $RECEIVER_LOG 2>&1
$IPFS config --json Addresses.API "\"/ip4/127.0.0.1/tcp/6001\"" >> $RECEIVER_LOG 2>&1
$IPFS config --json Addresses.Gateway "\"/ip4/127.0.0.1/tcp/8081\"" >> $RECEIVER_LOG 2>&1
$IPFS config --json Addresses.Swarm "[\"/ip4/0.0.0.0/tcp/7001\", \"/ip6/::/tcp/7001\"]" >> $RECEIVER_LOG 2>&1
$IPFS daemon >> $RECEIVER_LOG 2>&1 &
# wait for daemon to start.. maybe?
# ipfs id returns empty string if we don't wait here..
sleep 5

#
# start a p2p listener on RECIVER to the HTTP server with our content
#
$IPFS p2p listen /x/test /ip4/127.0.0.1/tcp/8000 >> $RECEIVER_LOG 2>&1
FIRST_ID=$($IPFS id -f "<id>")

export IPFS_PATH=$PATH2
$IPFS init >> $SENDER_LOG 2>&1
$IPFS config --json Experimental.Libp2pStreamMounting true >> $SENDER_LOG 2>&1
$IPFS daemon >> $SENDER_LOG 2>&1 &
# wait for daemon to start.. maybe?
sleep 5



# send a http request to SENDER via proxy to RECIEVER that will proxy to web-server

echo "******************"
echo proxy response
echo "******************"
curl http://localhost:5001/proxy/http/$FIRST_ID/test/index.txt



echo "******************"
echo link http://localhost:5001/proxy/http/$FIRST_ID/test/index.txt
echo "******************"
echo "RECEIVER IPFS LOG " $RECEIVER_LOG
echo "******************"
cat $RECEIVER_LOG

echo "******************"
echo "SENDER IPFS LOG " $SENDER_LOG
echo "******************"
cat $SENDER_LOG

```


cboddy:
FYI this link  appears to be broken 
https://github.com/ipfs/community/blob/master/contributing.md#contribution-guidelines

magik6k:
Thanks! You should be able to add a sharness test which hijacks the request the request using netcat as is done in https://github.com/ipfs/go-ipfs/blob/master/test/sharness/t0235-cli-request.sh#L14.

p2p/proxy.go
```diff
@@ -0,0 +1,92 @@
+package p2p
+
+import (
+	"bufio"
+	"fmt"
+	"net"
+	"net/http"
+	"strings"
+
+	core "github.com/ipfs/go-ipfs/core"
```
magik6k: Insert additional `\n` between local and gx imports

p2p/proxy.go
```diff
@@ -0,0 +1,92 @@
...
+			// open connect to peer
+			stream, err := ipfsNode.P2P.PeerHost.NewStream(ipfsNode.Context(), parsedRequest.target, protocol.ID("/x/"+parsedRequest.name))
+			if err != nil {
+				msg := fmt.Sprintf("Failed to open stream '%v' to target peer '%v'", parsedRequest.name, parsedRequest.target)
+				handleError(w, msg, err, 500)
+				return
+			}
+
+			// send request to peer
+			proxyReq, err := http.NewRequest(request.Method, parsedRequest.httpPath, request.Body)
```
magik6k: Does that also work for websockets / multipart requests out of the box?

p2p/proxy_test.go
```diff
@@ -0,0 +1,21 @@
+package p2p
+
+import (
+	"github.com/ipfs/go-ipfs/thirdparty/assert"
```
magik6k: Import order

p2p/proxy.go
```diff
@@ -0,0 +1,92 @@
+package p2p
```
magik6k: As this doesn't really depend on any of the p2p internals, I'd move this to `p2p/proxy`

cboddy: Ah something went awry here; this should have been  in ``core/corehttp``, now moved.

p2p/proxy.go
cboddy: good question, will add a sharness test

cboddy:
@magik6k  thanks for the swift feedback; have started and will add an sharness test as suggested.

Stebalien:
So, I believe this should actually be using the [ReverseProxy](https://golang.org/pkg/net/http/httputil/#ReverseProxy) helper. That will copy and modify all the right headers and handle all the edge-cases. However, that may just make it more complicated so take this as a suggestion. (You won't be able to use the `NewSingleHostReverseProxy` helper, you'll have to implement a custom `RoundTripper`.)

I believe we also need to read while writing. That is, the current version expects to write the *entire* request before it reads any response. We should *start* the request and then read the response as some worker finishes writing the request.

core/corehttp/proxy.go
```diff
@@ -0,0 +1,94 @@
...
+				handleError(w, "Failed to format proxy request", err, 500)
+				return
+			}
+
+			proxyReq.Write(stream)
+
+			s := bufio.NewReader(stream)
+			proxyResponse, err := http.ReadResponse(s, proxyReq)
+
+			defer func() { proxyResponse.Body.Close() }()
```
Stebalien: Do we need this? I believe `proxyResponse.Write` will close the body when done.

Also, I'm not sure if this is *valid* if `err != nil`. Will it not panic?

core/corehttp/proxy.go
```diff
@@ -0,0 +1,94 @@
...
+// /proxy/http/$peer_id/$name/$http_path
+func parseRequest(request *http.Request) (*proxyRequest, error) {
+	path := request.URL.Path
+
+	split := strings.SplitN(path, "/", 6)
+	if len(split) < 6 {
+		return nil, fmt.Errorf("Invalid request path '%s'", path)
+	}
+
+	if split[2] != "http" {
```
Stebalien: Maybe we should just *mount* under `/proxy/http`? We'll probably want a separate proxy implementation per protocol.

ianopolous: done

lanzafame:
@cboddy @Stebalien this might be a little late in the PR cycle to suggest but, if I understand correctly, @hsanjuan has built [`go-libp2p-http`](https://github.com/hsanjuan/go-libp2p-http) for this use case.

hsanjuan:
Hello, yes, https://github.com/hsanjuan/go-libp2p-http will probably simplify this (along with `httputil.ReverseProxy`). go-libp2p-http provides the libp2p transport (`RoundTripper` implementation) that you are looking for the `ReverseProxy`. If I'm correct you just need to initialize it with `libp2p://peerid` as target URL and manually set the transport. At least, it would not add another http RoundTripper to the libp2p world.

The `RoundTrip` in go-libp2p-http is simplified if you compare it to all the default transport's implementation (https://golang.org/src/net/http/transport.go?s=3628:10127#L385) but it's well tested and I think it should be solid enough (used by cluster for a while). It supports reading responses while sending the requests (for streaming requests).

The only problem would be to attach a custom transport tag to every Roundtripper. We can easily enable this upstream (currently a single tag is used for all), if you are willing to go down this path.

cboddy:
@Stebalien @lanzafame  thanks for the feedback.

Will look into using ``ReverseProxy`` as suggested (most of the go stdlib still a bit new to me) and try to push something today/ in the next few days.

core/corehttp/proxy.go
```diff
@@ -0,0 +1,103 @@
...
+		mux.HandleFunc("/proxy/http/", func(w http.ResponseWriter, request *http.Request) {
+			// parse request
+			parsedRequest, err := parseRequest(request)
+			if err != nil {
+				handleError(w, "Failed to parse request", err, 400)
+				return
+			}
+
+			// open connect to peer
+			stream, err := ipfsNode.P2P.PeerHost.NewStream(ipfsNode.Context(), parsedRequest.target, protocol.ID("/x/"+parsedRequest.name))
```
Stebalien: I believe this should be using the request context.

core/corehttp/proxy.go
```diff
@@ -0,0 +1,103 @@
...
+type roundTripper struct {
+	stream *inet.Stream
+}
+
+func (rt *roundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+
+	sendRequest := func() {
+		err := req.Write(*rt.stream)
+		if err != nil {
+			(*(rt.stream)).Close()
```
Stebalien: This should probably be `rt.stream.Reset()`.

core/corehttp/proxy.go
```diff
@@ -0,0 +1,103 @@
...
+		r.URL.Path = req.httpPath //the scheme etc. doesn't matter
+	}
+
+	return &httputil.ReverseProxy{
+		Director:  director,
+		Transport: &roundTripper{streamToPeer}}
+}
+
+type roundTripper struct {
+	stream *inet.Stream
```
Stebalien: No real need for the pointer. This is already an interface (class-like thing) so we can just say `inet.Stream`.

hsanjuan:
fyi, protocol tags per Roundtripper coming about: https://github.com/hsanjuan/go-libp2p-http/pull/12

Stebalien:
One more thing we'll need to do before merging this: put it behind an experimental feature-gate. Take a look at how the `ipfs p2p` command does this (using `config.Experimental.Libp2pStreamMounting`).

I don't really see this needing a long stabilization period, but we should probably leave it experimental *at last* until we stabilize the `ipfs p2p` feature...

ianopolous:
I've tested out all our api calls for Peergos through this, which includes gets, posts, and multiparts. And everything works!


peergos-over-p2p-stream

ianopolous:
@hsanjuan We can't use the RoundTripper in go-libp2p-http because it assumes a different incoming request format, and we don't want to do an extra socket+http round trip. We used your technique to close the stream though.

core/corehttp/proxy.go
ianopolous: fixed

ianopolous:
@Stebalien Does that mean we need a PR to go-ipfs-config to add a new config option and then gx publish that before this can be merged, or can we use the existing Libp2pStreamMounting option?

hsanjuan:
> @hsanjuan We can't use the RoundTripper in go-libp2p-http because it assumes a different incoming request format, and we don't want to do an extra socket+http round trip. We used your technique to close the stream though.

That's not true. You just need to rewrite the path, which you are already doing anyway.

ianopolous:
@hsanjuan If we used go-libp2p-http would that not imply an extra http round trip? Here we receive the incoming request and directly write it to the p2p stream. But my understanding (correct me if I'm wrong - I'm not a Gopher) is that if we rewrote it to use go-libp2p-http we would then be passing through another socket and http parser before getting to the p2p stream?

core/corehttp/proxy.go
```diff
@@ -0,0 +1,128 @@
...
+// we wrap the response body and close the stream
+// only when it's closed.
+type respBody struct {
+	io.ReadCloser
+	stream inet.Stream
+}
+
+// Closes the response's body and the connection.
+func (rb *respBody) Close() error {
+	rb.stream.Close()
```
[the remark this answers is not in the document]
ianopolous: Thanks. Is that better with the latest commit?

hsanjuan:
> @hsanjuan If we used go-libp2p-http would that not imply an extra http round trip? Here we receive the incoming request and directly write it to the p2p stream. But my understanding (correct me if I'm wrong - I'm not a Gopher) is that if we rewrote it to use go-libp2p-http we would then be passing through another socket and http parser before getting to the p2p stream?

@ianopolous no, what go-libp2p-http does is **exactly** what you are doing here. The only difference is that it works with a standard go http client by expecting urls in a way that the standard http library can handle so that you can use a the normal Go's `http.Client` with both http and http-under-libp2p.

Stebalien:
> @Stebalien Does that mean we need a PR to go-ipfs-config to add a new config option and then gx publish that before this can be merged, or can we use the existing Libp2pStreamMounting option?

Let's add a new experiment, it's a pretty orthogonal feature (only using stream mounting on the client). I'll handle the gx update (yeah, I know it's a bit of a pain).

Stebalien:
@hsanjuan mind making a PR against *this* PR? That'll make it easier to see what the change entails.

hsanjuan:
@Stebalien here: https://github.com/cboddy/go-ipfs/pull/4/files

core/corehttp/proxy.go
```diff
@@ -0,0 +1,132 @@
...
+	name     string
+	httpPath string // path to send to the proxy-host
+}
+
+// from the url path parse the peer-ID, name and http path
+// /proxy/http/$peer_id/$name/$http_path
+func parseRequest(request *http.Request) (*proxyRequest, error) {
+	path := request.URL.Path
+
+	split := strings.SplitN(path, "/", 6)
```
hsanjuan: This is going to be problematic with the way protocol tags usually look like because normally they have one or several slashes `/x/something`. I am not sure if this is a spec requirement regarding protocol IDs or something.

I was thinking this would be able to reach the cluster libp2p-http API but it uses `/libp2p-http` for protocol tag, and it's not possible to express this as part of the url path (easily).

I would suggest to settle on a default protocol tag for proxying http and allowing modifications via a query-parameter (or a custom header).

Stebalien: > I would suggest to settle on a default protocol tag for proxying http and allowing modifications via a query-parameter

Unfortunately, we can't use a query parameter (as we can't distinguish between ones that need to be forwarded). We *could* use a header but that's not very user/browser-friendly.

We currently use `/x/something` because `ipfs p2p` listens on `/x/...` protocols (by default). The idea behind this feature was to allow a client to connect to `http://localhost:5001/proxy/http/QmId/something` and have this entire request forwarded to some remote `ipfs p2p` listener. We could just always use the same protocol but I really want to be able to use IPFS to proxy to arbitrary webservers.

On the other hand, being able to connect to ipfs-cluster would be awesome. Additionally, I can see us eventually adding a custom protocol to browsers.

We *could* just encode them.

---

An alternative is to make the *server* smarter. That is, we can:

1. Listen on one protocol (possibly allowing the protocol to be set via a header).
2. Add a new `ipfs http` command.

The user would then run `ipfs http route add / http://localhost:1224/a/b` to, e.g., route all traffic to that webserver. The upside is:

1. This server can add authenticated `x-libp2p-peer-id: QmClient` headers.
2. Flexible routes.
3. We can even serve plain files/directories (act like a gateway).

However, this is starting to look like a *big* feature. Really, we I think we can probably get away with both (although we may still want some encoding scheme to allow *this* feature to talk to cluster, thoughts)?

That is, we can eventually add a single-protocol forwarder at `http://gateway/http/QmId/`.

Thoughts?

ianopolous: Would it be that restrictive to not allow /'s in protocol names? Logically it isn't and I suspect this issue will otherwise crop up in other places with URLs using protocol names..

Stebalien: IMO, not really. However, it *does* restrict this command to `/x/` protocols. I'm just polling for feedback and trying to make sure we don't paint ourselves into a corner.

Regerdless, I don't think this discussion should block this PR. This *is* an experimental API so, if we have a good enough reason, we can still change it.

ianopolous: If you made that prefix "x-" then even that restriction would go away as well.

hsanjuan: I think just as you can have a custom header you could have a custom query parameter `?x-fw-protocol=/x/test` which you strip (I realize this is not pretty but it's so easy to use).

Altenatively, the best is probably url encoding the slashes. That's what encoding is for anyway, but it might give some headaches to users at some point.

I'm not sure the added complexity making server smarter compensates that right now it's quite simple and solves the proposing user's usecase. But in that case `route add` could take a custom protocol tag for the route too.

hsanjuan:
By the way, let's not forget to document this. A section in the https://github.com/ipfs/go-ipfs/blob/master/docs/experimental-features.md doc. I don't know where it should be documented once it's not experimental though (here probably: https://docs.ipfs.io/reference/api/http/ ?).

core/corehttp/proxy.go
Stebalien: > I think just as you can have a custom header you could have a custom query parameter `?x-fw-protocol=/x/test` which you strip (I realize this is not pretty but it's so easy to use).

This is just a bug waiting to happen. I guess we could do `/proxy/http/Qm...?protocol=xyz&path=/a/b/c`.

> If you made that prefix "x-" then even that restriction would go away as well.

To be consistent with pretty much everything else we do, I'd like to stick to `/x/`. For context, "protocols" were envisioned to be `/ipfs/QmFile/path/to/spec`. My suggestion with `x-...` was that we could encode `/x/...` as `x-...` in the url.

ianopolous: The main issue I see with a custom header, or query parameter is you lose transitivity. i.e. I can't pipe a query over a p2p stream to one peer, which then pipes it to another peer. But maybe that's a weird use case..

I think my vote would be for url encoding the protocol name. 

Encoding the path in the query parameter is a no go because then you lose the fundamental feature of a a website "just working" behind a p2p proxy.

hsanjuan: > My suggestion with x-... was that we could encode /x/... as x-... in the url

So cluster uses `/libp2p-http` for protocol, so if you want it to work you cannot use that encoding. I say, let's use a default protocol and put the custom one in a header?. It's not obvious to a user that he should write  `-a-b-c` in one place and then listen to protocol `/a/b/c` in some other.

ianopolous:
This is waiting for a gx bubble up from https://github.com/ipfs/go-ipfs-config now.

ianopolous:
@hsanjuan Do you know if this build issue is a gx import issue or something else? (It builds fine locally)
https://ci.ipfs.team/blue/organizations/jenkins/IPFS%2Fgo-ipfs/detail/PR-5526/3/pipeline/

core/corehttp/proxy.go:36:43: cannot use ipfsNode.P2P.PeerHost (type "gx/ipfs/QmQKod7iLxQK6X3aFYvhDDnFdXo3QjxKeL2F7UrPtdKQR2/go-libp2p-host".Host) as type "gx/ipfs/QmeA5hsqgLryvkeyqeQdvGDqurLkYi3XEPLZP3pzuBJXh2/go-libp2p-host".Host in argument to p2phttp.NewTransport:

core/corehttp/proxy.go
ianopolous: What do we think about url encoding the protocol where it is currently, in the path? Headers won't work because a website served through this won't add the magic header to sub resource requests. Query parameters have the same issue.

Kubuxu:
Check out: https://ci.ipfs.team/blue/organizations/jenkins/IPFS%2Fgo-ipfs/detail/PR-5526/7/pipeline/16

You have some duplicate gx dependencies imported which would explain problem you are having.

hsanjuan:
@ianopolous you'll need to rebase this to go-ipfs/master and then `gx update go-libp2p-http github.com/hsanjuan/go-libp2p-http` .

cboddy:
rebased this onto master..

hsanjuan:
maybe you needed to do `gx install --global` first? @ianopolous

ianopolous:
@Stebalien How do we fix the dupes: https://ci.ipfs.team/blue/organizations/jenkins/IPFS%2Fgo-ipfs/detail/PR-5526/14/pipeline
They don't have anything to do with our code. This is super painful.

hsanjuan:
@ianopolous I think you need to rebase in master and re-update go-libp2p-http to the latest (which uses libp2p 6.0.19 in line with the rest of go-ipfs/master)

ianopolous:
Ok I've rebased and gx updated for the third time. Hopefully we can merge this soon.

ianopolous:
@Kubuxu or anyone: Any chance you could give me a hand fixing gx et al to get this merge-able? Many thanks.

hsanjuan:
@ianopolous If still doesn't work, if you can give me write access to `cboddy` I can try to fix the branch.

Stebalien:
Note: Don't bother with rebasing when you're done, I'll handle that (I'm about to pull though a massive gx update...).

cboddy:
Thanks @Stebalien @hsanjuan  , yes I noticed last night that the multi-part test seems to fail at a rate of about 1 in 10 attempts. 

So far it appears that when it fails it is because the  ``nc`` server receives a truncated request (everything after the first mutipart-boundary is missing), I'm looking into it now.

hsanjuan:
I need to bubble libp2p-6.0.20 to go-libp2p-http. Will do this tomorrow.

Stebalien:
> So far it appears that when it fails it is because the nc server receives a truncated request (everything after the first mutipart-boundary is missing). Not 100% sure what is going on yet; looking into it now.

Oh. *that* issue. Yeah, we hit that in our other tests as well. I believe we just need to implement a custom tool for testing HTTP things in go.

core/corehttp/proxy.go
```diff
@@ -57,8 +57,13 @@ func parseRequest(request *http.Request) (*proxyRequest, error) {
 	if len(split) < 6 {
 		return nil, fmt.Errorf("Invalid request path '%s'", path)
 	}
+	//url-decode the name
+	decodedName, err := url.PathUnescape(split[4])
```
Stebalien: I believe we need to use `RawPath`. Go auto-unescapes. Can we test this?

Stebalien:
I have a PR here to cleanup the tests a bit: https://github.com/cboddy/go-ipfs/pull/5. This uses IPTB and ensures configures the test nodes to not connect to the rest of the network (i.e., don't enable discovery, don't bootstrap).

Stebalien:
So, we have a bit of an issue. Go's ServeMux automatically "cleans up" URLs so it'll helpfully:

1. Decode the slashes.
2. Strip the duplicates.
3. *REDIRECT* the user.

Even if we choose a different escaping scheme for protocols, go will still end up rewriting the rest of the URL for us and there's no way to avoid that (as far as I can tell).

Options:

1. Inject some logic before the mux. That is, we can have a mux-mux that skips the normal mux for proxied requests.
2. Put the path in the query string. That is, `/http/proxy?peer=xyz&proto=/libp2p-http&path=/a/b/c`. This breaks browsing web pages but, really, we're using this to proxy requests.
3. Use a different port.
4. Accept the path normalization. Probably a bad idea from a security standpoint.

ianopolous:
I'm pretty keen not to break browsing a website. That is our second most important use case after a single request. I also hope to use it personally to host arbitrary websites without requiring dns or indeed a public server. 

@Stebalien Can you give some concrete examples where the encoding is a problem? I'd still fallback to restricting the character set for protocol names, seems like you're set on allowing "/"'s so maybe ban "-"'s (or some other non slash chartacter) and then we can encode them that way?

ianopolous:
Hmm it's worse than I thought, if I type the following into a browser it undecodes, before even sending the request!
So this:
localhost:5001/proxy/http/QmaCZqbmBP5u2435pmBT8x5Ks6ZiWNg66ET4ctZNKmw78K/%2fx%2ftest/
becomes
localhost:5001/proxy/http/QmaCZqbmBP5u2435pmBT8x5Ks6ZiWNg66ET4ctZNKmw78K/x/test/

which clearly doesn't work for protocols with "/"'s in them

ianopolous:
@Stebalien I've reverted the url decoding because it isn't going to work. Can we merge this and move the protocol name discussion to another pr?

Stebalien:
>  Which tag of iptb are you using/is expected?

See: https://github.com/ipfs/go-ipfs/issues/5521. I'm using 1.3.19

> I've reverted the url decoding because it isn't going to work. Can we merge this and move the protocol name discussion to another pr?

It's worse than that, unfortunately, as we'll *also* redirect/rewrite the *http* path. We're in a release freeze at the moment so we can think this over for a few days. I'm find merging something that gets us most of the way there but I prefer not merging API changes that we intend to break in the immediate future.

ianopolous:
@Stebalien How long until the next release? I was super hoping to make this one. :-(

> I'm find merging something that gets us most of the way there but I prefer not merging API changes that we intend to break in the immediate future.

This version only works for protocol names without /'s. Unless we remove the protocol from the path entirely, then this shouldn't change for such protocol names. All we would be doing is expanding support to include more protocol names.

Stebalien:
So, thinking about hosting websites from this API for a minute, I realized something rather important: We *can't* expose this from the API for security reasons (XSS).

That means we can either:

1. Expose it from a new port.
2. Expose it from the gateway.

Personally, I think exposing it on the gateway will be the easiest.

---

WRT encoding, names, etc. I'd like to propose a solution that I think will make everyone happy:

`/p2p/$ID[/x/$PROTOCOL]/(http|ws)[$PATH]`

* When a protocol is specified, the final protocol would be `/x/$PROTOCOL/(http|ws)`.
* When a protocol *isn't* specified, we'd use the default protocol for the transport. That is, `/libp2p-ws` or `/libp2p-http` (@hsanjuan any chance we can change these to `/http` and `/ws`? That is, make cluster also listen on `/http`?).

Examples:

* /p2p/Qm.../http/index.html
* /p2p/Qm.../x/backend/http/index.html


Rational:

1. It's almost a multiaddr (I wish we could embed paths in multiaddrs).
2. It makes the sequence of operations very clear (connect to node Qm..., speak protocol X, talk HTTP, etc.).
3. With a browser addon, we can map these onto `libp2p://Qm...` and `libp2p://$PROTOCOL.Qm....` to preserve origins. Really, slashes are *always* going to be an issue for us if we want to support cases like this.
4. We *don't* really need to support *arbitrary* libp2p protocols.

@cboddy @hacdias @ianopolous opinions?

---

Regardless of how it may seem, I really don't want to draw this out indefinitely. I'm fine merging this as-is (on a custom port or as a part of the gateway, behind a feature flag) as long as we agree to revisit this and work under the understanding that the interface is unstable and will likely change.

hsanjuan:
> WRT encoding, names, etc. I'd like to propose a solution that I think will make everyone happy:
> 
> `/p2p/$ID[/x/$PROTOCOL]/(http|ws)[$PATH]`
> 

This means we need an additional `libp2p-ws` for websocket proxying?

>   @hsanjuan any chance we can change these to `/http` and `/ws`? That is, make cluster also listen on `/http`?).

That should be no problem.

LGTM. Good catch with the XSS, it was rather obvious and we almost missed it :/

Stebalien:
> To be clear though, doesn't this imply you are banning using /http in the (explicit) protocol name, as you are using that to determine the end of the name in the path?

Actually, I'm banning slashes entirely. I wonder if that's a bit confusing... Basically, I'm saying that protocols *must* be of the form `/x/CUSTOM/http` or *just* `/http`. The idea is that custom applications would listen on `/x/my-app/http` (or `/x/my-app/ws`).

Eventually I expect this'll be done through a libp2p daemon and applications will (a) register sub-peer and then (b) listen on `/ws` or `/http`. But we're not there yet (except for custom libp2p apps like cluster).

> This means we need an additional libp2p-ws for websocket proxying?

Someday.

hsanjuan:
> Actually, I'm banning slashes entirely

It's fine. This is just for applications that want to be accessible by default using a default reverse proxy exposed in the gateway enpoint of the go-ipfs daemon. It's a very specific usecase for which an extra protocol handler that satisfies the limitations can be just added easily. You can always manually mount these streams with whatever custom protocol in the worst case (now and in a future libp2p daemon).

ianopolous:
@Stebalien I've made those changes to the path spec, and moved the end point to the gateway.

ianopolous:
(I tested both formats of the endpoint with the full Peergos web ui)

ianopolous:
@Stebalien how can we get the gateway address out of iptb? Any chance you could update the sharness tests to the new spec?

hsanjuan:
Overall looks OK, but I'll look into this with a bit more time in a couple of days, unless it's super urgent, in which case just ping me.

hannahhoward:
@hsanjuan -- I wonder if this is worth moving forward cause it's blocking https://github.com/ipfs/go-ipfs/pull/5762 -- which is needed to fix everyone's problems with `gx-go link`ing subpackages with go-ipfs/master right now. Or, if you feel it's worth moving #5672 ahead of this for that reason (understanding it will be a challenging rebase)

core/corehttp/proxy.go
```diff
@@ -0,0 +1,79 @@
...
+	target   string
+	name     protocol.ID
+	httpPath string // path to send to the proxy-host
+}
+
+// from the url path parse the peer-ID, name and http path
+// /p2p/$peer_id/http/$http_path
+// or
+// /p2p/$peer_id/x/$protocol/http/$http_path
+func parseRequest(request *http.Request) (*proxyRequest, error) {
```
hsanjuan: This function looks a bit messy and will get messier adding `/ws`.  I think now it allows things like `/x//http` (weird).


I think the cleanest is to do regexps:

```go
// /p2p/$peer_id/x/$protocol/http/$http_path
// Change to ^/p2p/(\w+)/(x/\w+/(?:http|ws))/(.*)$ when adding /ws
var longProtocolRex = regexp.MustCompile(`^/p2p/(\w+)/(x/\w+/http)/(.*)$`)

// /p2p/$peer_id/http/$http_path
// Change to `^/p2p/(\w+)/(http|ws)/(.*)$` when adding /ws
var shortProtocolRex = regexp.MustCompile(`^/p2p/(\w+)/(http)/(.*)$`)

// from the url path parse the peer-ID, name and http path
// /p2p/$peer_id/http/$http_path
// or
// /p2p/$peer_id/x/$protocol/http/$http_path
func parseRequest(request *http.Request) (*proxyRequest, error) {
	path := request.URL.Path

	matches := longProtocolRex.FindAllStringSubmatch(path, -1)
	if matches == nil {
		matches = shortProtocolRex.FindAllStringSubmatch(path, -1)
	}

	if matches == nil || len(matches) < 1 || len(matches[0]) < 4 {
		return nil, fmt.Errorf("invalid proxy request path: '%s'", path)
	}

	return &proxyRequest{
		target:   matches[0][1],
		name:     protocol.ID("/" + matches[0][2]),
		httpPath: matches[0][3],
	}, nil
}

```

Technically this would allow not having to change the function when we add `/ws`. It imposes that custom protocols have a-zA-Z0-9 characters though.

core/corehttp/proxy.go
```diff
@@ -0,0 +1,79 @@
...
+		return nil, fmt.Errorf("Invalid request path '%s'", path)
+	}
+
+	return &proxyRequest{split[2], protocol.ID("/x/" + split[4] + "/http"), split[6]}, nil
+}
+
+func handleError(w http.ResponseWriter, msg string, err error, code int) {
+	w.WriteHeader(code)
+	fmt.Fprintf(w, "%s: %s\n", msg, err)
+	log.Warningf("server error: %s: %s", err)
```
hsanjuan: Perhaps better to say we're not a "server":
```suggestion
	log.Warningf("http proxy error: %s: %s", err)
```

core/corehttp/proxy_test.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"net/http"
+	"strings"
+	"testing"
+
+	"github.com/ipfs/go-ipfs/thirdparty/assert"
+)
+
+func TestParseRequest(t *testing.T) {
+	url := "http://localhost:5001/p2p/QmT8JtU54XSmC38xSb1XHFSMm775VuTeajg7LWWWTAwzxT/http/path/to/index.txt"
```
hsanjuan: We need to add at least one more test-case here for the `/x/custom/http` case. I would do it so that it can easily be expanded to the `/ws` cases in the future (loop on a slice of struct containing testcases strings and expected results for each).

core/corehttp/proxy_test.go
```diff
@@ -0,0 +1,34 @@
...
+	if err != nil {
+		t.Error(err)
+	}
+	assert.True(parsed.httpPath == "path/to/index.txt", t, "proxy request path")
+	assert.True(parsed.name == "/http", t, "proxy request name")
+	assert.True(parsed.target == "QmT8JtU54XSmC38xSb1XHFSMm775VuTeajg7LWWWTAwzxT", t, "proxy request peer-id")
+}
+
+func TestParseRequestInvalidPath(t *testing.T) {
+	url := "http://localhost:5001/p2p/http/foobar"
```
hsanjuan: Same here, we should test that cases like `/x/http/url` fail as well.

core/corehttp/proxy_test.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/ipfs/go-ipfs/thirdparty/assert"
+)
+
+func TestParseRequest(t *testing.T) {
+	url := "http://localhost:5001/p2p/QmT8JtU54XSmC38xSb1XHFSMm775VuTeajg7LWWWTAwzxT/http/path/to/index.txt"
+	req, _ := http.NewRequest("GET", url, strings.NewReader(""))
+
+	parsed, err := parseRequest(req)
+	if err != nil {
+		t.Error(err)
```
hsanjuan: this should be `t.Fatal`
```suggestion
		t.Fatal(err)
```

core/corehttp/proxy.go
ianopolous: I personally don't think regexes are cleaner. I prefer simpler, faster code. 

I also have the fast path for /http (by checking it first) which I imagine will be more common, being the default. 

Regarding ws, my understanding (I'm not an expert in them by any measure) is that a web socket starts out as a normal http request, but with a special header, which signals the server to treat it differently. If that is true, I don't understand the need for a separate /ws endpoint. Certainly for my usecase we don't need it. 

If we want to ban empty $custom protocol components then that is trivial to add.

core/corehttp/proxy_test.go
ianopolous: I've added more test cases

core/corehttp/proxy_test.go
ianopolous: /p2p/x/http/url is actually valid here, because it will treat x as the multihash in the short form. It will of course fail later on as soon as the p2p implementation tries to parse it as a multihash.

docs/experimental-features.md
```diff
@@ -382,6 +383,87 @@ with `ssh [user]@127.0.0.1 -p 2222`.
...
+
+***On the "server" node:***
+
+First, start your application and have it listen for TCP connections on
+port `$APP_PORT`.
+
+Then, configure the p2p listener by running:
+
+```sh
+> ipfs p2p listen --allow-custom-protocol /http /ip4/127.0.0.1/tcp/$APP_PORT
```
magik6k: We should probably give the `/x/http` example here

test/sharness/t0184-http-proxy-over-p2p.sh
```diff
@@ -0,0 +1,199 @@
...
+
+function serve_http_once() {
+    #
+    # one shot http server (via nc) with static body
+    #
+    local body=$1
+    local status_code=${2:-"200 OK"}
+    local length=$((1 + ${#body}))
+    REMOTE_SERVER_LOG="server.log"
+    echo -e "HTTP/1.1 $status_code\nContent-length: $length\n\n$body" | nc -l $WEB_SERVE_PORT 2>&1 > $REMOTE_SERVER_LOG &
```
magik6k: This is going to be horribly racy, I'd add:
```suggestion
    rm $REMOTE_SERVER_LOG
    echo -e "HTTP/1.1 $status_code\nContent-length: $length\n\n$body" | nc -l $WEB_SERVE_PORT 2>&1 > $REMOTE_SERVER_LOG &
    test_wait_for_file 30 100ms $REMOTE_SERVER_LOG
```

core/corehttp/proxy.go
Stebalien: > I personally don't think regexes are cleaner. I prefer simpler, faster code.

I tend to agree and I think we can refactor this later if we need to.

> If that is true, I don't understand the need for a separate /ws endpoint. Certainly for my usecase we don't need it.

The `ws` case is a slightly different use-case: the websocket would be handled by the proxy and the inner stream itself would be forewarded. Thinking through that case, let's just ignore the ws stuff for now. My proposal wasn't quite right for that use-case but I think we can make it work (although it might not be completely symmetric to http).

core/corehttp/proxy_test.go
hsanjuan: I'm talking about something like ` /p2p/peer/x/http/url`. That is a fail condition in the code that is not tested here (bad format for the long protocol tag mode). I thin you're not covering all the ` return error`  conditions (there are 2 and you're only testing 1 thing).

Stebalien:
Is this blocked on me in any way?

ianopolous:
@stebalien I don't think there's anything else left to do here?

core/corehttp/proxy_test.go
```diff
@@ -0,0 +1,55 @@
...
+			t.Fatal(err)
+		}
+		assert.True(parsed.httpPath == tc.path, t, "proxy request path")
+		assert.True(parsed.name == protocol.ID(tc.name), t, "proxy request name")
+		assert.True(parsed.target == tc.target, t, "proxy request peer-id")
+	}
+}
+
+var invalidtestCases = []string{
+	"http://localhost:5001/p2p/http/foobar",
```
hsanjuan: "http://localhost:5001/p2p/peerID/x/myproto/nothttp/foobar" is the other condition that causes failure and is not tested

ianopolous: Done

ianopolous:
@magik6k I believe @Stebalien said that those random multipart failures are independent of this PR and you guys have been hitting them elsewhere too?

Stebalien:
I think I've fixed the netcat issue. We'll see if we can get the tests to pass.

Stebalien:
OK, tests pass and @hsanjuan has given a :heavy_check_mark: :steam_locomotive: (this train *may* be a bit slow...).

ianopolous:
Hooray! The last critical feature we need for Peergos!

@stebalien is the plan to enable this on ipfs.io?

cboddy:
@Stebalien @magik6k @hsanjuan thanks for your help with this one!

Stebalien:
> @Stebalien is the plan to enable this on ipfs.io?

Probably not for a while, if ever. We'll have to seriously consider the security and performance implications and, really, the public gateway's more of a crutch (we'd prefer if users would just load js-libp2p from their DAPP).

ianopolous:
> Probably not for a while, if ever. We'll have to seriously consider the security and performance implications and, really, the public gateway's more of a crutch (we'd prefer if users would just load js-libp2p from their DAPP).

:-( That's a shame, because can do some really cool demos if that is enabled. Loading js-libp2p will never be an option for us for security reasons.

Stebalien:
Are you forbidding javascript entirely? Can you not spin up a js-ipfs node in a sandboxed worker?

I do agree this would allow for some pretty cool demos but I'd like to sit on it _at least_ until it's stabalized.

ianopolous:
@Stebalien In general, we veto anything that uses npm which is anathema to basic security practices. We're also trying to keep the JS down to a small easily auditable amount (as much as possible) for the same reasons. As well as being much harder to audit for attacks, having a complex P2P daemon running in the same process as that which has the secret keys is also asking for trouble. Browsers explicitly don't defend against spectre attacks within the same page to my knowledge (they'd have to run every worker in an independent OS process, not just different OS threads). We also have reproducible builds for the web-ui already. 

For now we can just preface all ipfs-only demos with, install ipfs and enable these two flags, then browse to this local url.

Stebalien:
> In general, we veto anything that uses npm which is anathema to basic security practices. 

Well, I can't really blame you there...

> Browsers explicitly don't defend against spectre attacks within the same page to my knowledge

Sandboxing has more to do with origins/contexts than pages. I'm not familiar with browser spectre mitigations but, a page in a sandboxed iframe/worker should have equivalent security to a page running in a separate tab.

ianopolous:
> Sandboxing has more to do with origins/contexts than pages. I'm not familiar with browser spectre mitigations but, a page in a sandboxed iframe/worker should have equivalent security to a page running in a separate tab.

If the p2p daemon was running from a separate domain in an iframe, I believe this *might* be safe in *some* browsers (depending on their site isolation properties), but it's much harder to prove it safe, and it requires you to have two separate origins. We don't have any 3rd party hosted content in our web-ui - it's all self hosted (partially to enable easy offline and localhost access). 

How about we continue this on irc to stop spamming all the other contributors to this pr? :-)

Issue #5401, opened 2018-08-27
Allow "Addresses.Gateway" / "Addresses.API" to be array (Allow gateway/api to listen on multiple addresses)


mkg20001:
#### Version information:

```
go-ipfs version: 0.4.17-
Repo version: 7
System version: amd64/linux
Golang version: go1.10.3
```

#### Description: Allow "Addresses.Gateway" / "Addresses.API" properties to be arrays (Allow gateway/api to listen on multiple addresses)

Currently it's only possible to make the gateway/API listen on one address only.

Instead both of these parameters should accept arrays to make it possible to for ex listen on `/ip6/::1/tcp/8080` as well.

kjzz:
hey @Stebalien,if anyone work for this issue?if not,can i do for this?

overbool:
@kjzz sorry, I'm fixing it.

overbool:
@Stebalien this feature refers to go-ipfs-config, should modify it first?

lgierth:
> this feature refers to go-ipfs-config, should modify it first?

Yes, and there's an fs-repo migration required, to do the string-to-array conversion in existing config files: https://github.com/ipfs/fs-repo-migrations -- best to coordinate with @kevina here

lgierth:
Taken care of in #5532 and #5553

PR #5532, opened 2018-09-27
feat(api): allow multiple API/Gateway addresses

overbool:
Refer to: https://github.com/ipfs/go-ipfs-config/pull/11
Fixes: https://github.com/ipfs/go-ipfs/issues/5401

License: MIT

### DIFF
--- a/cmd/ipfs/daemon.go
+++ b/cmd/ipfs/daemon.go
@@ -412,22 +412,32 @@ func serveHTTPApi(req *cmds.Request, cctx *oldcmds.Context) (<-chan error, error
 		return nil, fmt.Errorf("serveHTTPApi: GetConfig() failed: %s", err)
 	}
 
+	apiAddrs := make([]string, 0, 2)
 	apiAddr, _ := req.Options[commands.ApiOption].(string)
 	if apiAddr == "" {
-		apiAddr = cfg.Addresses.API
-	}
-	apiMaddr, err := ma.NewMultiaddr(apiAddr)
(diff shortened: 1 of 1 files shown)


overbool:
@Stebalien could u help me review it? refer to https://github.com/ipfs/go-ipfs-config/pull/11

cmd/ipfs/daemon.go
```diff
@@ -461,15 +471,25 @@ func serveHTTPApi(req *cmds.Request, cctx *oldcmds.Context) (<-chan error, error
 		return nil, fmt.Errorf("serveHTTPApi: ConstructNode() failed: %s", err)
 	}
 
-	if err := node.Repo.SetAPIAddr(apiMaddr); err != nil {
+	if err := node.Repo.SetAPIAddr(listeners[0].Multiaddr()); err != nil {
```
hannahhoward: do we want to connect the repo to only one of the addresses? Should it always be the first?

Stebalien: So, adding support for multiple addrs here would require changing the `~/.ipfs/api` file format (or, more likely, introducing a *second* API file with support for multiple addresses). Choosing the *first* address is probably reasonable for now. Eventually, we may want to consider preferring a loopback address, or something like that, but we can revisit this later.

Stebalien:
Blocked on go-ipfs-config update (in progress).

lgierth:
This has been merged as part of #5553 -- thank you :]

Issue #5522, opened 2018-09-26
Better support for /dnsaddr/ (i.e. in swarm connect)


hsanjuan:
#### Version information: 

master

#### Type:

Enhancement

#### Description:

It should be possible to do `ipfs swarm connect /dnsaddr/something`. Right now it doesn't attempt to resolve and throws `Error: invalid peer address: invalid IPFS address`

overbool:
It seems the same problem like https://github.com/ipfs/go-ipfs/issues/5249.

Maybe we should resolve the dns address before `swarm connect`.

overbool:
@hsanjuan `Error: invalid peer address: invalid IPFS address` implies that the `addr` is invalid ipfs address, because it misses `tcp` and `ipfs` protocol. I try `ipfs swarm connect /dnsaddr/xxx/tcp/4001/ipfs/Qmxxx` in my bash, it print `connect Qmxxx success`.

hsanjuan:
@overbool you are confusing `/dns4` or `/dns6` with `/dnsaddr` (common mistake), but they are different. `/dns4` resolves domain to `/ip4/<ip>`. `/dnsaddr` resolves multiaddress from `_dnsaddr` TXT field associated to the domain.

overbool:
@hsanjuan You are right. I've confused `/dns4` with `/dnsaddr`.

overbool:
It seems the same issue https://github.com/ipfs/go-ipfs/issues/5249. 
Maybe we should resolve `addr` before connecting.

Stebalien:
@overbool that one is for the `--api` flag (that's how we tell the IPFS client which daemon it should connect to).

@hsanjuan we'll have to think about a better way to do https://github.com/multiformats/go-multiaddr-dns/blob/78f39e8892d4f8c5c9f18679cc06d0b40ecab8cf/resolve.go#L167

Basically, when we resolve a dnsaddr, we match the *final* component to try to pull out the right ones. In theory, this allows us to have *multiple* distinct peers with the same hostname. In practice, this is kind of arbitrary. Any solution that tries to solve this issue will also have to solve that.

Note: there are also security concerns but I'm not *that* concerned. If the user doesn't specify a peer ID, they have no reason to believe we'll end up connecting to the right peer (they haven't told us which one to use).

hsanjuan:
Ok, two problems here.

[...]

PR #5535, opened 2018-09-28
fix(swarm): add dnsaddr support in swarm connect

overbool:
Fixes: https://github.com/ipfs/go-ipfs/issues/5522

License: MIT

### DIFF
--- a/core/commands/swarm.go
+++ b/core/commands/swarm.go
@@ -1,11 +1,15 @@
 package commands
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"io"
 	"path"
 	"sort"
+	"strings"
+	"sync"
+	"time"
 
 	commands "github.com/ipfs/go-ipfs/commands"
 	cmdenv "github.com/ipfs/go-ipfs/core/commands/cmdenv"
@@ -22,6 +26,10 @@ import (
 	swarm "github.com/libp2p/go-libp2p-swarm"
 	ma "github.com/multiformats/go-multiaddr"
 	mafilter "github.com/whyrusleeping/multiaddr-filter"
+)
+
+const (
+	dnsResolveTimeout = 10 * time.Second
 )
 
 type stringList struct {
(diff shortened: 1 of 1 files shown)


core/commands/swarm.go
```diff
@@ -557,10 +560,43 @@ func parseAddresses(addrs []string) (iaddrs []iaddr.IPFSAddr, err error) {
...
-	iaddrs, err := parseAddresses(addrs)
+	// resolve the addrs
+
+	var maddrs []ma.Multiaddr
+	for _, addr := range addrs {
+		maddr, err := ma.NewMultiaddr(addr)
+		if err != nil {
+			return nil, err
+		}
+		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
michaelavila: I've looked through this whole PR and everything makes sense to me :+1:. I do have a question about how these timeouts are specified in go-ipfs (both here and more generally). I notice lines like the following (and there are a handful):

https://github.com/ipfs/go-ipfs/blob/master/core/bootstrap.go#L60

But also lines like yours (e.g.):

https://github.com/ipfs/go-ipfs/blob/master/core/commands/pubsub.go#L178

Is there a rule of thumb for deciding when to specify these timeouts with a name vs inline?

Stebalien: Well, we should probably name this. We just get lazy sometimes.

core/commands/swarm.go
```diff
@@ -557,10 +560,43 @@ func parseAddresses(addrs []string) (iaddrs []iaddr.IPFSAddr, err error) {
...
+		raddrs, err := madns.Resolve(ctx, maddr)
```
Stebalien: Can we parallelize this?

core/commands/swarm.go
```diff
@@ -557,10 +560,43 @@ func parseAddresses(addrs []string) (iaddrs []iaddr.IPFSAddr, err error) {
...
+		cancel()
+		if err != nil {
+			return nil, err
+		}
+		if len(raddrs) == 0 {
+			return nil, fmt.Errorf("non-resolvable multiaddr about %v", maddr)
+		}
+		maddrs = append(maddrs, raddrs...)
```
Stebalien: I believe we may need to filter out addresses that *still* don't end in `/ipfs/Qm...`.

hsanjuan:
> we'll insert the resolved addrs into the peerstore instead of the dnsaddr.

Maybe this has changed, but do unresolved addresses in the peerstore get resolved when building a connection? My impression when I was testing this a while ago is they didn't. I remember a discussion or PR in which @lanzafame was involved (??) suggesting to do this at a lower level in libp2p.

A few months back, addresses were only resolved in `basichost.Connect` (https://github.com/libp2p/go-libp2p/blob/master/p2p/host/basic/basic_host.go#L383) but normally things will use `NewStream` directly and I think (i might be wrong) that that codepath does not resolve if it has to open a new connection.

So might be better off storing resolved addresses for the moment?

Stebalien:
Ah... No, they don't. That's broken. Yeah, you're right. But we should *really* fix this. We don't even do routing on that path.

overbool:
>For each address that doesn't end in /ipfs/Qm..., resolve it.

@Stebalien Did you mean that if address ends in `/ipfs/Qm...`, we don't need to resolve it, because it will resolve address inner the `basichost.Connect`?

overbool:
@Stebalien Thx your advice, i had fixed it.

Issue #5530, opened 2018-09-27
Measuring memory usage


rob-deutsch:
I have a question about measuring and understanding go-ipfs' memory usage - because I'm getting conflicting signs.

I'm doing this because go-ipfs is using a lot of memory on my docker instance (~1GB) and I want to understand why.

I've been following the instructions in [debug-guide.md](https://github.com/ipfs/go-ipfs/blob/master/docs/debug-guide.md) but I'm getting different results to what my OS is telling me.

My OS is telling me that I'm using around 256MB:

````
$ ps aux | head -n1
USER               PID  %CPU %MEM      VSZ    RSS   TT  STAT STARTED      TIME COMMAND
$ ps aux | grep ipfs
rob              16719   0.5  3.1 558590036 256180 s002  Ss+   2:24pm   6:28.99 ipfs daemon
$
````

But pprof tells me that it's only using 76.38MB:

````
$ curl localhost:5001/debug/pprof/heap > ipfs.heap
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 78010    0 78010    0     0  3822k      0 --:--:-- --:--:-- --:--:-- 4009k
$ go tool pprof /Users/rob/golang/bin/ipfs ipfs.heap 
ipfs
Type: inuse_space
Time: Sep 27, 2018 at 5:21pm (AEST)
Entering interactive mode (type "help" for commands, "o" for options)
(pprof) top5
Showing nodes accounting for 38.29MB, 50.14% of 76.38MB total
Showing top 5 nodes out of 207
      flat  flat%   sum%        cum   cum%
   13.47MB 17.63% 17.63%    22.97MB 30.07%  gx/ipfs/QmQjMHF8ptRgx4E57UFMiT4YM6kqaJeYxZ1MCDX23aw4rK/golang-lru/simplelru.(*LRU).Add
    9.50MB 12.44% 30.07%     9.50MB 12.44%  container/list.(*List).PushFront
    6.31MB  8.26% 38.34%     6.31MB  8.26%  gx/ipfs/Qmda4cPRvSRyox3SqgJN6DfSZGU5TtHufPTp9uXjFj71X6/go-libp2p-peerstore/pstoremem.(*memoryAddrBook).AddAddrs
[...]

PR #5545, opened 2018-09-29
debug-guide.md added memory statistics command

rob-deutsch:
This would've been valuable to me while investigating #5530.

I discovered that more debug commands are also available, but I don't know enough about them to add them myself (e.g. `debug/pprof-mutex`, `debug/metrics/prometheus`).


rob-deutsch:
Will resubmit with better details.

PR #5546, opened 2018-09-29
debug-guide.md added memory statistics command

rob-deutsch:
This would've been valuable to me while investigating #5530.

I discovered that more debug commands are also available, but I don't know enough about them to add them myself (e.g. `debug/pprof-mutex`, `debug/metrics/prometheus`).

### DIFF
--- a/docs/debug-guide.md
+++ b/docs/debug-guide.md
@@ -4,8 +4,9 @@ This is a document for helping debug go-ipfs. Please add to it if you can!
 
 ### Table of Contents
 - [Beginning](#beginning)
-- [Analysing the stack dump](#analysing-the-stack-dump)
+- [Analyzing the stack dump](#analyzing-the-stack-dump)
 - [Analyzing the CPU Profile](#analyzing-the-cpu-profile)
+- [Analyzing vars and memory statistics](#analyzing-vars-and-memory-statistics)
 - [Other](#other)
 
 ### Beginning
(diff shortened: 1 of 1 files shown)


rob-deutsch:
Thanks @schomatis, resubmitted with suggested additions.

docs/debug-guide.md
```diff
@@ -84,6 +86,10 @@ about `go tool pprof`. My go-to method of analyzing these is to run the `web`
 command, which generates an SVG dotgraph and opens it in your browser. This is
 the quickest way to easily point out where the hot spots in the code are.
 
+### Analysing vars and memory statistics
```
schomatis: nit: could you also add an entry in the "Table of Contents" above pointing here?

rob-deutsch: Drat, my bad, will do

rob-deutsch:
... I also noticed (and fixed) the inconsistent spelling of analyse/analyze, just to make it a tad neater.

schomatis:
> ... I also noticed (and fixed) the inconsistent spelling of analyse/analyze, just to make it a tad neater.

I think they are both right but "analyze" is the accepted American spelling.

rob-deutsch:
In the case of ambiguity I tend to prefer British English, but some after some Googling I learned that US English is generally accepted for this kind of purpose.

Resubmitted with US English and a small frown of disapproval 😉

Kubuxu:
There is also a much simpler option for non-linux folks to download most of the info: going to `localhost:5001/debug/` and doing `Save as` on the links.

Not blocking this PR in any regard, just mentioning it as I discovered it recently.

rob-deutsch:
I'm getting a 404 error on `localhost:5001/debug/`. That sounds like a handy feature though, can we integrate that to master?

Kubuxu:
Sorry: `localhost:5001/debug/pprof/`.

PR #5549, opened 2018-10-02
take gc-lock when updating pins

Stebalien:
Otherwise, we can GC half-way through.

### DIFF
--- a/core/commands/pin.go
+++ b/core/commands/pin.go
@@ -405,6 +405,8 @@ new pin and removing the old one.
 			return
 		}
 
+		defer n.Blockstore.PinLock().Unlock()
+
 		err = n.Pinning.Update(req.Context(), fromc, toc, unpin)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


Stebalien:
(now that we know that holding the pin lock doesn't actually help...)

Stebalien:
@magik6k should `PinAPI.Update` take this lock internally?

michaelavila:
Related: I think lines like https://github.com/ipfs/go-ipfs/blob/master/core/coreunix/add.go#L194 also need to block GC, right?

Stebalien:
> Related

We actually need to take the lock at a higher layer there. That is, we need to hold the lock between adding the file and pinning.

Apparently, this has already been fixed... https://github.com/ipfs/go-ipfs/pull/5506

magik6k:
> @magik6k should `PinAPI.Update` take this lock internally?

Yup, it should map 1-1 to the command with what it does

Stebalien:
Replaced by #5550.

Issue #5554, opened 2018-10-03
Add a `--force` flag for `ipfs files rm`


schomatis:
At the moment the `-r` is acting also as `--force`: in the case the file (node) we're trying to remove is corrupt it will just unlink without caring what it is unlinking (file, directory, corrupted node, etc); see https://github.com/ipfs/go-ipfs/issues/5538#issuecomment-426312110. The `-r` should only allow to remove valid directories and the new `--force` flag should allow to remove anything else.

This would imply changing the API, some users might already rely on this side effect of `-r` but seems subtle enough to modify it without breaking much. @magik6k WDYT?

overbool:
@schomatis OK

Issue #5538, opened 2018-09-28
MFS: unable to remove malformed nodes


djdv:
I managed to add an invalid node to MFS and was unable to remove it.
I added it via something similar to this
```go
mfs.PutNode(APINode.FilesRoot, "/bad-node", new(dag.ProtoNode))
```
trying to remove it via `ipfs files rm /bad-node` results in
>Error: proto: required field "Type" not set

and the file remains in the MFS listing.

I'm not sure if `PutNode` should check validity before put-ing, or if `rm` should account for these kinds of things.

schomatis:
> I'm not sure if `PutNode` should check validity before put-ing,

It should.



> if `rm` should account for these kinds of things.

I'm not sure why this corrupted node would cause the error during the remove operation (it might be related to the MFS directory caching mechanism), I would expect that the modification (unlink) happened only at the (valid) parent directory without caring for the integrity of the child. Could you post a simple example to reproduce the error please?

Also, stop corrupting MFS @djdv!! :rofl:

djdv:
>Could you post a simple example to reproduce the error please?

This should be all that's needed. If I start with a fresh repo `ipfs init`, and make that API call on an instance of it, then try to `rm` that path, it fails with that error.

Something like this should work (with proper imports and adding to commands root).
But I don't want to test it right now 👀 

```go
var CorruptCmd= &cmds.Command{
	Helptext: cmdkit.HelpText{
		Tagline:          "Corrupts MFS, for when you need to",
	},
	Run: func(req *cmds.Request, res cmds.ResponseEmitter, env cmds.Environment) error {
		node, _ := cmdenv.GetNode(env)
		mfs.PutNode(node.FilesRoot, "/bad-node", new(dag.ProtoNode))
		return errors.New("MFS corrupted")
	}

}
```


>Also, stop corrupting MFS @djdv!!


Y̨o̱ű h̡aͦv́e̽ m᷀ŷ͓ ẉo̝r̮d̈́ t͊h̩e̡r͉e̛ w̯i̗l̅l͡ b̢eͤ n͝o̮ m͜o͕rͯe͊ cͅo̅r͍ȑu͒p̮t͓i̕oͪnͪ

schomatis:
@overbool Do you want to take a look here? We need to find the source of the error and understand which check is failing during the addition.

overbool:
@schomatis ok, I'm interested at it.

overbool:
[...]

PR #5555, opened 2018-10-03
feat(command): add force flag for files rm

overbool:
Refer to: https://github.com/ipfs/go-ipfs/issues/5538#issuecomment-426312110, https://github.com/ipfs/go-ipfs/issues/5554

License: MIT

### DIFF
--- a/core/commands/files.go
+++ b/core/commands/files.go
@@ -1002,6 +1002,7 @@ Remove files or directories.
 	},
 	Options: []cmdkit.Option{
 		cmdkit.BoolOption("recursive", "r", "Recursively remove directories."),
+		cmdkit.BoolOption("force", "Forcibly remove target at path; implies -r for directories"),
 	},
 	Run: func(req oldcmds.Request, res oldcmds.Response) {
 		defer res.SetOutput(nil)
@@ -1041,8 +1042,6 @@ Remove files or directories.
 			return
 		}
 
-		dashr, _, _ := req.Option("r").Bool()
-
 		var success bool
 		defer func() {
 			if success {
@@ -1054,8 +1053,10 @@ Remove files or directories.
 			}
 		}()
 
-		// if '-r' specified, don't check file type (in bad scenarios, the block may not exist)
-		if dashr {
+		// if '--force' specified, it will remove anything else,
+		// including file, directory, corrupted node, etc
+		force, _, _ := req.Option("force").Bool()
+		if force {
 			err := pdir.Unlink(name)
 			if err != nil {
 				res.SetError(err, cmdkit.ErrNormal)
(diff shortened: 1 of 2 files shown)


core/commands/files.go
```diff
@@ -1002,6 +1002,7 @@ Remove files or directories.
 	},
 	Options: []cmdkit.Option{
 		cmdkit.BoolOption("recursive", "r", "Recursively remove directories."),
+		cmdkit.BoolOption("force", "Allow to remove anything else"),
```
eingenito: What do you think of `forcibly remove target at path; implies -r for directories` as a description?

overbool:
@schomatis could u help me review this pr?

schomatis:
The `-f` short flag is already used for the global flush flag,

https://github.com/ipfs/go-ipfs/blob/a86cde50de4bd8da0b5eaf6f1936269bafd29c8e/core/commands/files.go#L56,

but that's ok, I actually prefer to just have the long explicit `--force` flag since this should be used with care and (hopefully) not too often.

core/commands/files.go
```diff
@@ -1066,25 +1067,31 @@ Remove files or directories.
...
 		if err != nil {
 			res.SetError(err, cmdkit.ErrNormal)
 			return
 		}
 
-		switch childi.(type) {
+		dashr, _, _ := req.Option("r").Bool()
+
+		switch child.(type) {
 		case *mfs.Directory:
```
schomatis: nit: we can now turn this single-case switch into an `if`

test/sharness/t0250-files-api.sh
```diff
@@ -629,6 +629,20 @@ test_files_api() {
   test_expect_success "repo gc $EXTRA" '
     ipfs repo gc
   '
+
+  # test rm
```
schomatis: I'd like to add a test with an actual corrupted node, which could just be an empty proto-node, but I'm not sure how could we add that to MFS, `ipfs files cp` does check the type:
```
ipfs object new
# QmdfTbBqBPQ7VNxZEYEj14VmRuZBkqFbiwReogJgS1zR1n
ipfs files cp /ipfs/QmdfTbBqBPQ7VNxZEYEj14VmRuZBkqFbiwReogJgS1zR1n /corrupted
# Error: cp: cannot flush the created file /corrupted: proto: required field "Type" not set
```

schomatis: Oh, turn off flush:

```
ipfs files -f=0 cp /ipfs/QmdfTbBqBPQ7VNxZEYEj14VmRuZBkqFbiwReogJgS1zR1n /corrupted
ipfs files rm /corrupted
Error: proto: required field "Type" not set
```

schomatis: OT: What's the difference of adding a file in MFS without flushing? I still don't know how that works (and I'm supposedly the one in charge of reviewing the MFS API :grimacing:)

schomatis: Thinking about this some more, after adding the checks to `PutNode` this won't work anyways (even with flush turned off), so scratch that, I'm fine with the existing tests.

Issue #5556, opened 2018-10-03
`ipfs pin ls` should list recursive pins only by default


djdv:
Currently, `ipfs pin ls` will list all pins.
This is an expensive operation and it seems like most people want `ipfs pin ls --type=recursive` anyway.
I propose that we change this so that recursive pins are listed by default (`--type=recursive`), and `--type=all` should be explicitly required for listing all pins.

Even if we make listing all pins fast, @lgierth points out that `ipfs add` pins objects recursively by default. So we may want to still do this anyway for consistency's sake.

lidel:
cc https://github.com/ipfs/go-ipfs/issues/4891

Stebalien:
I agree we should do something like this but we'll need to improve the options. Really, we need `--not-type=indirect`. We *do* want to include *direct* pins.

---

However, this is a large, API-breaking change.

@whyrusleeping, @diasdavid, @alanshaw (js-land),

How do you all feel about making `ipfs pin ls` *not* list indirect blocks by default. For context, this is *really* slow for reasonable datasets as it'll list *every* block explicitly added to the datastore (all pinned files).

lgierth:
How about `--type=recursive,direct` (there's precedent in go-ipfs)

lidel:
@aschmahmann just fyi current default is pretty bad because `pin ls` slows down with every pin, so it is invisible during dev when repo is empty or has small number of pins, but may lead to bugs in production like https://github.com/ipfs-shipyard/ipfs-webui/issues/1618, where having 1k of pins caused pin ls to take 10 seconds.

PR #5559, opened 2018-10-03
pin: list recursive by default

overbool:
Fixes: https://github.com/ipfs/go-ipfs/issues/5556

License: MIT


Stebalien:
Closing so we can go with the `recursive,direct` solution.

PR #5561, opened 2018-10-04
[WIP] S3 Datastore

hinshun:
Based on #5516

This is some initial work to get S3 datastore working with IPFS.
Uses fork of `go-ds-s3`: https://github.com/ipfs/go-ds-s3/pull/7


magik6k:
Need https://github.com/ipfs/go-ipfs/pull/5187 first

plugin/plugins/s3ds/s3ds.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+}
+
+func (s3c *S3Config) Create(path string) (repo.Datastore, error) {
+	return &loggingDatastore{
+		Datastore: s3ds.NewS3Datastore(s3c.cfg),
+	}, nil
+}
+
+type loggingDatastore struct {
```
magik6k: There already is a logging datastore: https://github.com/ipfs/go-ipfs/blob/master/repo/fsrepo/datastores.go#L66

hinshun: Thanks, just starting to learn the codebase.

hinshun:
I attempted to `gx install` go-ipfs through a ipfs daemon with s3 datastore and it was very slow and got OOM-killed. For smaller loads it seems to work fine but I think we'll need to properly implement `go-ds-s3` with `aws-sdk-go`.

MichaelMure:
Any idea if this would allow to run multiple IPFS nodes with a shared s3 datastore ?

magik6k:
If you only use this for `/blocks/` and disable GC on all nodes then it may be possible

hinshun:
As an update, I rewrote `ipfs/go-ds-s3` with `aws-sdk-go` [here](https://github.com/ipfs/go-ds-s3/pull/7) and was successful in gx installing all of `go-ipfs`'s dependencies.

postables:
I've been trying something similar to this and have noticed a critical issue, which is that if you pin some content, and run `ipfs pin ls` you can see the content as being pinned. If you shutdown the node, start it up and run `ipfs pin ls` again, nothing shows up.

georgyo:
Should this ticket be closed in favor of making ipfs/go-ds-s3 in the default plugin list?

lidel:
S3 plugin lives at https://github.com/ipfs/go-ds-s3#readme and is looking for maintainer (see note in that repo).

I'm closing this PR due to that and the fact it uses deprecated and removed gx.

If anyone is interested in discussing having built-in support for S3 in Kubo, please fill issue in https://github.com/ipfs/kubo/issues explaining your use case. As a rule of thumb, we only include plugins that bring value to significant part of userbase and are actively maintained, thus it has to be true for S3 as well.

Issue #4978, opened 2018-04-25
Error in offline name.publish after id


alanshaw:
#### Version information:
go-ipfs version: 0.4.14-5db3846
Repo version: 6
System version: amd64/darwin
Golang version: go1.10

#### Type:
Bug

#### Description:

Only in offline mode, calling `name.publish` after `id` causes an error.

```
$ ipfs daemon --offline &
[1] 47216
$ Initializing daemon...
Swarm not listening, running in offline mode.
API server listening on /ip4/127.0.0.1/tcp/5001
Gateway (readonly) server listening on /ip4/127.0.0.1/tcp/8090
Daemon is ready
$ ipfs id
{
	"ID": "QmahispzgXbQ5ewaxSjdrst8KNLzREhVHdhxZwAUHQWqLb",
	"PublicKey": "CAASpgIwggEiMA0GCSqGSIb3DQEBAQUAA4IBDwAwggEKAoIBAQDC4Q6B6burDPTlfIkjlF72Q/G72GT7CDBzlZsGI9XKf3XCC5PQiTBcpvKEnUCoBvquiClWE1yJiM4bu1dH/YR5NDBlx9EQzDDmiVvS909cqokRHxA0NY5t83hs8gthoMt6wn15euG2e6LVohqefq6YiiFWgakxDcRyMclLWKiF2t7U6OhMMSd2YdHp15wcHac13ZmdNKhZ/TBhjeKowl4DjAixxLE0kmcK5MsABeQEA+BHukz425K3929onEriSkG7xkSrd0YwKOx6J2Apub7Df8yhcG1PFoTyfTUD/gxkYNrB2jjqUJaNXrK4QDbaDinBNS3KZUn4SQtmxzj2gi4PAgMBAAE=",
	"Addresses": null,
	"AgentVersion": "go-ipfs/0.4.14/5db3846",
	"ProtocolVersion": "ipfs/0.1.0"
}
$ ipfs name publish zdpuArHMUAYi3VtD3f7iSkXxYK9xo687SoNf5stAQNCMzd77k
Error: private key already loaded
```

kjzz:
use cmd "pfs name resolve" can alse cause this error

kjzz:
use cmd "ipfs name rsolve" cause the same error,and i have find the reason.

kjzz:
please assign this issue to me

kjzz:
@alanshaw

kjzz:
i have fix it in pr #4980

[...]

PR #5562, opened 2018-10-04
add offline id test #4978 and refactor command code

kjzz:
fix issue #4978


core/commands/active.go
```diff
@@ -25,7 +29,7 @@ Lists running and recently run commands.
 		res.SetOutput(req.InvocContext().ReqLog.Report())
 	},
 	Options: []cmdkit.Option{
-		cmdkit.BoolOption("verbose", "v", "Print extra information."),
+		cmdkit.BoolOption("verbose", verboseOptionName, "Print extra information."),
```
eingenito: We might want to switch this to constantize (and later use in req.Option()) the long name for this command so that it is consistent with all the others.

PR #5564, opened 2018-10-04
don't use the domain name as a filename in /ipns/a.com

### LINKED ISSUE #5369: IPNS dnslink tld interpreted as file ending
#### Type:
Bug

#### Description:
I made a dns txt record for my site that looks like this
`"dnslink=/ipfs/QmYgdZ7Rhk2PRawzGu1FBHQSRsMaVMCJUF7NcbY6wVr4WC"`.
When you visit https://ipfs.io/ipns/jacob-karlsson.com you get the content-type `application/x-msdos-program` which is not expected. However, when I visit https://ipfs.io/ipfs/QmYgdZ7Rhk2PRawzGu1FBHQSRsMaVMCJUF7NcbY6wVr4WC I get the expected `image/gif`. This happens in both Firefox and Chrome for me, but someone in irc reported it working for them.

My guess is that this is a combination of https://github.com/ipfs/go-ipfs/issues/4543, and ipfs.io having a mime-type definition file somewhere that includes this https://whatis.techtarget.com/fileformat/COM-Command-file-program (I tried to find an actual example of a file like this but didn't manage). I.e. the gateway thinks my website's url is a filename.

Stebalien:
fixes #5369

This should be an uncontroversial fix while we debate content type detection in
general.


core/corehttp/gateway_test.go
```diff
@@ -153,6 +153,7 @@ func TestGatewayGet(t *testing.T) {
 	ns["/ipns/double.example.com"] = path.FromString("/ipns/working.example.com")
 	ns["/ipns/triple.example.com"] = path.FromString("/ipns/double.example.com")
 	ns["/ipns/broken.example.com"] = path.FromString("/ipns/" + k)
+	ns["/ipns/example.man"] = path.FromString("/ipfs/" + k)
```
kevina: I take it that `.man` has special meaning to the go http library?  I would add a test for this.  To me it is very non-obvious.

Alternatively I would have the content be a small html file and use `example.txt`instead and make sure the content type is an html file not a text file.

Stebalien: > I take it that .man has special meaning to the go http library? I would add a test for this. To me it is very non-obvious.
> Alternatively I would have the content be a small html file and use example.txtinstead and make sure the content type is an html file not a text file.

I picked `.man` because it's a valid TLD. I couldn't get this to work with `.html` or `.txt` because IPFS will refuse to resolve these as domains. I considered using `.com` but I needed something that *could* be a COM file (the original issue used a GIF). I didn't want to deal with that.

Stebalien: I'll add a comment.

core/corehttp/gateway_test.go
```diff
@@ -153,6 +153,12 @@ func TestGatewayGet(t *testing.T) {
 	ns["/ipns/double.example.com"] = path.FromString("/ipns/working.example.com")
 	ns["/ipns/triple.example.com"] = path.FromString("/ipns/double.example.com")
 	ns["/ipns/broken.example.com"] = path.FromString("/ipns/" + k)
+	// We picked .man because:
+	// 1. It's a valid TLD.
+	// 2. Go treats it as the file extension for "man" files (even though
+	//    nobody actually *uses* this extension, AFAIK).
```
kevina: This is system dependent.  http.ServeContext calls mime.TypeByExtension which in the sandbox go environment does not detect it as a `application/x-troff-man`.  See: https://play.golang.org/p/ohNcHHsVArq

core/corehttp/gateway_test.go
```diff
@@ -153,6 +153,12 @@ func TestGatewayGet(t *testing.T) {
...
+	// 3. Go accepts "fnord" (the test value) as a valid man file.
```
kevina: ???

Did you mean go does not accept "fnord" as a valid man file?

core/corehttp/gateway_test.go
Stebalien: Ah. Damn, that's why `.com` wasn't working.

core/corehttp/gateway_test.go
Stebalien: It accepts it as a valid man file *if* the extension is ".man". However, it looks like that part isn't important. I thought go looked at *both* the content *and* the extension but that doesn't appear to be the case (https://github.com/ipfs/go-ipfs/pull/5564#discussion_r223095520).

I don't know how I can make these tests platform-independent. At least they appear to pass on Linux now...

Issue #5557, opened 2018-10-03
should help support --help? / help can't help help


djdv:
This is a strange one, but I'm reporting it anyway.
`ipfs help --help`
>Error: Unknown Command "help"

This seems to be the only command that doesn't support the `--help` flag.
While it probably wouldn't do anything more than the bare `help` subcommand itself, I found the error odd.

kjzz:
So you meani that we should delete `ipfs help`?

Stebalien:
Unfortunately, we actually use it. I said "leftover" because it looks like it may be a leftover from before we had a commands library (not sure, it may not have been).

However, we should probably make it behave better. The solution here is probably to just remove the length check. That is, convert "help" to "--help", even if there are other flags. Thoughts?

djdv:
Would it be better to formalize it as a command? With this kind of alias mapping:
`ipfs help add` -> `ipfs add --help`?

Stebalien:
Maybe? I'm not sure it's worth the effort (that would probably require deep changes to the commands library).

kjzz:
> Would it be better to formalize it as a command? With this kind of alias mapping:
> `ipfs help add` -> `ipfs add --help`?

Agree,I think that ipfs <sub-command> --help maybe better,of course it will casue breanking?


> Unfortunately, we actually use it. I said "leftover" because it looks like it may be a leftover from before we had a commands library (not sure, it may not have been).

Sorry, i have misunderstand your reply.

djdv:
For now, we could probably use the same hack, but append to args.
i.e.
remove arg[1]/"help" as well when detected.
`if helpcmd; args = append(args, "--help")`
in phase: `ipfs help add` -> `ipfs add` -> `ipfs add --help`

Right now, we alias `ipfs help` to `ipfs -h` but `ipfs --help` works the same, so it shouldn't break that case either.
It seems to work with other flags too.
`ipfs add -w --help` is the same as `ipfs add --help` alone.

And keep it in mind as the cmdslib evolves. See if it's necessary to formalize it as a `cmd` command.

Not sure if stacking hacks on hacks is a good idea, but for this specific command it seems harmless and unlikely to ever require change.
I guess the only thing we'd need to do is create some kind of stub `cmd` that is there only to hold the documentation for `help` and propagate it in `ipfs commands`.

Thoughts?

kjzz:
Agree！@djdv , I am working for it right now.Thx a lot for advices.

[...]

PR #5566, opened 2018-10-05
delete useless  command `ipfs help`

kjzz:
fix issue #5557 

License: MIT

### DIFF
--- a/cmd/ipfs/main.go
+++ b/cmd/ipfs/main.go
@@ -82,13 +82,9 @@ func mainRet() int {
 	intrh, ctx := setupInterruptHandler(ctx)
 	defer intrh.Close()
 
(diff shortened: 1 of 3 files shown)


Stebalien:
I'd like to do this but our documentation literally says to use `ipfs help`. I think we're just going to have to improve it a bit.

Stebalien:
https://docs.ipfs.io/introduction/install/

Mr0grog:
So… should we update those docs to use `--help` if that’s the preferred form? Can someone file an issue on the docs repo if so?

Stebalien:
I don't think either is really preferred. It's more that our current CLI library doesn't support `help` very well. However, this has now been *properly* fixed in #5573.

PR #5573, opened 2018-10-09
fix `ipfs help` bug #5557

kjzz:
fix issue #5557 

License: MIT

### DIFF
--- a/cmd/ipfs/main.go
+++ b/cmd/ipfs/main.go
@@ -82,12 +82,21 @@ func mainRet() int {
 	intrh, ctx := setupInterruptHandler(ctx)
 	defer intrh.Close()
 
-	// Handle `ipfs help'
-	if len(os.Args) == 2 {
-		if os.Args[1] == "help" {
-			os.Args[1] = "-h"
-		} else if os.Args[1] == "--version" {
+	// Handle `ipfs version` or `ipfs help`
+	if len(os.Args) > 1 {
+		// Handle `ipfs --version'
+		if os.Args[1] == "--version" {
 			os.Args[1] = "version"
+		}
+
+		//Handle `ipfs help` and `ipfs help <sub-command>`
+		if os.Args[1] == "help" {
+			if len(os.Args) > 2 {
+				os.Args = append(os.Args[:1], os.Args[2:]...)
+				os.Args = append(os.Args, "-h")
+			} else {
+				os.Args[1] = "-h"
+			}
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


cmd/ipfs/main.go
```diff
@@ -82,13 +82,22 @@ func mainRet() int {
...
+		// Handle `ipfs --version'
+		if os.Args[1] == "--version" {
 			os.Args[1] = "version"
 		}
+
+		//Handle `ipfs help` and `ipfs help <sub-command>`
+		if os.Args[1] == "help" {
+			if len(os.Args) > 2 {
+				os.Args = append(os.Args[:1], os.Args[2:]...)
+				os.Args = append(os.Args, "-h")
```
djdv: This should be changed to the long flag `--help` as that applies to `ipfs` itself as well as all subcommands
`-h` only works with `ipfs -h` for instance `ipfs add -h` has a different meaning, but `ipfs add --help` is what we would want.

djdv: Disregard this, I was confusing `-h` and `-H`.
Although I am wondering which we should use for help, the short description or the long.

djdv:
My review is squashed.
Just a note for the next person.
I'm wondering if we should use the short help description or the long one. `-h` vs `--help`.
No strong opinion either way.

kjzz:
Thx for your quick review @djdv . 

> I'm wondering if we should use the short help description or the long one. `-h` vs `--help`.

IMO,maybe `-h` will be better.
Because ``

cmd/ipfs/main.go
Kubuxu: I would switch it to `--help` as `-h` is a short help and `--help` is a long help. (help help help :smile)

Kubuxu: Same in line 98.

kjzz:
@djdv @Kubuxu i have update the pr . And I have changed `-h` to `--help` ,please help me review again.Thx.

kjzz:
@djdv @magik6k If you have any time , can you please help me review my pr ?Thx a lot.

djdv:
Sorry for the delay @kjzz 
I wasn't very clear.
My initial review was only blocking on the question of using `-h` or `--help`.
With the decision to use `--help` I approve the changes.
But was waiting for input from someone with merge capabilities. (like @Kubuxu, who has posted above)

kjzz:
Thx for advice @Kubuxu.I have update my pr.Please help me review it again.

Stebalien:
This is a bit funky but, well, it was already kind of a hack. This should at least be less surprising for users.

PR #5574, opened 2018-10-09
coreapi unixfs: remove Cat, use sessions

magik6k:


core/corehttp/gateway_handler.go
```diff
@@ -270,7 +269,7 @@ func (i *gatewayHandler) getOrHeadHandler(ctx context.Context, w http.ResponseWr
 		} else {
 			name = getFilename(urlPath)
 		}
-		i.serveFile(w, r, name, modtime, dr)
+		i.serveFile(w, r, name, modtime, dr.(io.ReadSeeker))
```
Kubuxu: I am cautious of forced casts.

Kubuxu:
Is there a reason for not returning it with Seeker interface by default?

core/commands/cat.go
```diff
@@ -135,32 +135,48 @@ func cat(ctx context.Context, api iface.CoreAPI, paths []string, offset int64, m
...
-		count, err := read.Seek(offset, io.SeekStart)
+
+		count, err := file.Seek(offset, io.SeekStart)
 		if err != nil {
 			return nil, 0, err
 		}
 		offset = 0
 
-		size := uint64(read.Size() - uint64(count))
+		fsize, err = file.Size()
```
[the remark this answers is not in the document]
magik6k: This is just a getter under the hood - https://github.com/ipfs/go-unixfs/blob/master/io/pbdagreader.go#L163

magik6k:
@Stebalien do you think we should test this on a gateway first or can this just be merged?

core/coreapi/interface/unixfs.go
```diff
@@ -29,11 +35,7 @@ type UnixfsAPI interface {
 	//
 	// Note that some implementations of this API may apply the specified context
 	// to operations performed on the returned file
-	Get(context.Context, Path) (files.File, error)
-
-	// Cat returns a reader for the file
-	// TODO: Remove in favour of Get (if we use Get on a file we still have reader directly, so..)
-	Cat(context.Context, Path) (Reader, error)
```
Stebalien: We can probably get away with this but we'll need to be smart about whether we optimize for bandwidth or latency. That is, when calling `Cat`, we clearly care about reading the file in-order (latency to next byte). When calling `Get`, we're indicating we just want the entire file (maximize throughput overall).

We can still do this even if we remove the `Cat` function. However, we'll probably need to add some kind of option (or some method on the `UnixfsFile` interface).

We can handle this later, it's just something to keep in mind.

core/coreapi/path.go
```diff
@@ -73,3 +74,8 @@ func (api *CoreAPI) ResolvePath(ctx context.Context, p coreiface.Path) (coreifac
 
 	return coreiface.NewResolvedPath(ipath, node, root, gopath.Join(rest...)), nil
 }
+
+func (api *CoreAPI) getSession(ctx context.Context) *CoreAPI {
```
Stebalien: Let's add some documentation here. Specifically, we need to *at least* explain that the returned API has a read-only DAG.

core/corehttp/gateway_handler.go
```diff
@@ -178,14 +178,14 @@ func (i *gatewayHandler) getOrHeadHandler(ctx context.Context, w http.ResponseWr
...
+	dr, err := i.api.Unixfs().Get(ctx, resolvedPath)
 	dir := false
-	switch err {
-	case nil:
-		// Cat() worked
-		defer dr.Close()
-	case coreiface.ErrIsDir:
-		dir = true
+	switch {
+	case err == nil:
```
Stebalien: Why change this? If we're going to go with something like this, a simple `if err == nil { ... } else { ... }` would probably be cleaner.

core/corehttp/gateway_handler.go
```diff
@@ -178,14 +178,14 @@ func (i *gatewayHandler) getOrHeadHandler(ctx context.Context, w http.ResponseWr
...
+		dir = dr.IsDirectory()
+		if !dir {
+			defer dr.Close()
```
Stebalien: Actually... shouldn't we be closing this regardless? (I have no idea, this just looks weird).

core/coreapi/interface/unixfs.go
magik6k: Good point, I'd say this is easy enough to handle later

core/corehttp/gateway_handler.go
magik6k: So technically for `go-ipfs-files.File`, in it's current form, `Close` should be only called for non-directories: https://github.com/ipfs/go-ipfs-files/blob/master/file.go#L18-L20

I'm working on refactoring files, making close valid for all files is one of the goals, but I wouldn't want to have this PR blocked on that refactor

Stebalien: SGTM.

PR #5583, opened 2018-10-10
add systemd service file example

sneak:


sneak:
It's not autoinstalled or anything, it's just an example file for the user to configure, with defaults that work on the most commonly deployed systemd distributions.  It's not part of a distribution package or anything.

sneak:
It took me some time to suss out all of these various systemd-related quirks and I am contributing it simply in the hopes that it will same someone else that amount of time.  It's not exceedingly system-specific in anything other than perhaps the choice of StorageMax.

misc/systemd/ipfs.service
```diff
@@ -0,0 +1,31 @@
+[Unit]
+description=ipfs p2p daemon
+After=network.target
+Requires=network.target
```
lanzafame: A ipfs daemon doesn't actually need a network to start and the user can interact with several functions of ipfs without a network connection. Granted most people will want a network connection but part of the issue with putting examples in the repo is that people will take them as how things 'should' or 'must' work and that isn't necessarily the case.

misc/systemd/ipfs.service
```diff
@@ -0,0 +1,31 @@
...
+RestartSec=1
+Restart=always
+PermissionsStartOnly=true
+Nice=18
+StateDirectory=/var/lib/ipfs
+Environment=IPFS_PATH=/var/lib/ipfs
+Environment=HOME=/var/lib/ipfs
+LimitNOFILE=8192
+Environment=IPFS_FD_MAX=8192
+EnvironmentFile=-/etc/sysconfig/ipfs
```
lanzafame: The config file shouldn't exist at this path, see [here](https://www.freedesktop.org/software/systemd/man/file-hierarchy.html#System%20Packages)

misc/systemd/ipfs.service
```diff
@@ -0,0 +1,31 @@
...
+StandardOutput=journal
+WorkingDirectory=/var/lib/ipfs
+ExecStartPre=-adduser --system --group --home /var/lib/ipfs ipfs
```
lanzafame: How does this work on a multi-user machine? Does it mean any non-admin user is stuck having ipfs running in the background and with the server profile enable with no way of changing it to a more appropriate profile or stop ipfs completely?

misc/systemd/ipfs.service
```diff
@@ -0,0 +1,31 @@
...
+ExecStartPre=-mkdir /var/lib/ipfs
+ExecStartPre=-/bin/chown ipfs:ipfs /var/lib/ipfs
+ExecStartPre=-/bin/chmod u+rwx /var/lib/ipfs
+ExecStartPre=-chpst -u ipfs /usr/local/bin/ipfs init --profile=badgerds
```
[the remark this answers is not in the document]
sneak: This install location is taken from the install.sh script in the ipfs official distribution.

misc/systemd/ipfs.service
```diff
@@ -0,0 +1,31 @@
...
+ExecStartPre=-chpst -u ipfs /usr/local/bin/ipfs config profile apply server
```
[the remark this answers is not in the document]
sneak: This being a systemwide service file assumes that the "user" is root and that there is one ipfs daemon running on the system - just like ssh, or httpd, or any other system service.  They don't need to be run each time but otherwise there are additional steps to installing/starting ipfs and it doesn't hurt anything to run them each startup.

misc/systemd/ipfs.service
sneak: Now you're just being pedantic.  There is zero point to using the ipfs daemon without a network connection; all of the local-only options work fine without a daemon.  You certainly don't want the ipfs daemon starting up before the network as it discovers its network interfaces on startup.

misc/systemd/ipfs.service
sneak: The dash means that this service file works fine even if the file doesn't exist.

misc/systemd/ipfs.service
sneak: This is for a service started by root, which is why it uses /var/lib/ipfs - just like ssh, or httpd, or any other system service.  This is entirely normal for a systemd service file.  The "non-admin user" mode of a unix system is mostly outmoded these days anyway.  This isn't a time-sharing mainframe from the 70s, this is a way to run a daemon on a standard distribution quickly for a sysadmin in 2018 with sane defaults.

misc/systemd/ipfs.service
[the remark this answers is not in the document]
sneak: There is no way to run some of the `ExecStartPre`s as root (e.g. the adduser/chown/chmod) and others as the user in question (`ipfs`).  Either all are run as root or all are run as the `User` specified (depending on `PermissionsStartOnly`).  Because root is required for making sure the user exists and the directory has the correct owner/permissions, chpst is required to do the `ipfs init` as the correct user.  The fact that it is not installed by default is not our concern; ipfs is also not installed by default and `/usr/local/bin/ipfs` will not be available unless the user installs that.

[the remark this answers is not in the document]
sneak: Additionally, if someone tries to use this unit file and it fails, checking the systemd journal will show that they need `chpst` quite plainly.

misc/systemd/ipfs.service
lanzafame: No, what I mean is `/etc/sysconfig/<program-name>` is not the convention for systemd service configuration.

misc/systemd/ipfs.service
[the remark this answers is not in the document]
lanzafame: > They don't need to be run each time but otherwise there are additional steps to installing/starting ipfs and it doesn't hurt anything to run them each startup.

They *will* run each time the service is started or restarted, which also means that the configuration as specified in $IPFS_PATH/config will be overwritten each time. Sure it is great for when you are installing ipfs the first time but is annoying everytime after that.

misc/systemd/ipfs.service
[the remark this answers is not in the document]
lanzafame: But even that script provides two possible install locations depending on whether you have `/usr/local/bin` setup in your system...

misc/systemd/ipfs.service
lanzafame: > The "non-admin user" mode of a unix system is mostly outmoded these days anyway. 

This is anyone who shares a linux machine with other people in a household? Not everyone has the luxury of having their own machine. But I will take the answer as this service file doesn't work in multi-user scenario by design.

misc/systemd/ipfs.service
```diff
@@ -0,0 +1,31 @@
...
+Requires=network.target
+
+[Service]
+Type=simple
+User=ipfs
+RestartSec=1
+Restart=always
+PermissionsStartOnly=true
+Nice=18
+StateDirectory=/var/lib/ipfs
```
lanzafame: I don't have an issue with `/var/lib/ipfs` being the place that the repo gets stored but considering the default `StorageMax` of 4TB, which is too large for a general purpose service file, but also most suggested partitioning schemes for linux distros when installing a new distro, places more storage towards the `/home` directory instead of the `/` directory.

misc/systemd/ipfs.service
```diff
@@ -0,0 +1,31 @@
...
+EnvironmentFile=-/etc/sysconfig/ipfs
+StandardOutput=journal
+WorkingDirectory=/var/lib/ipfs
+ExecStartPre=-adduser --system --group --home /var/lib/ipfs ipfs
+ExecStartPre=-mkdir /var/lib/ipfs
+ExecStartPre=-/bin/chown ipfs:ipfs /var/lib/ipfs
+ExecStartPre=-/bin/chmod u+rwx /var/lib/ipfs
+ExecStartPre=-chpst -u ipfs /usr/local/bin/ipfs init --profile=badgerds
+ExecStartPre=-chpst -u ipfs /usr/local/bin/ipfs config profile apply server
+ExecStartPre=-chpst -u ipfs /usr/local/bin/ipfs config Datastore.StorageMax "4000GB"
```
lanzafame: 4TB does *not* fit into ~sane~ defaults.

misc/systemd/ipfs.service
sneak: The administrator of a service (or container) can mount storage wherever they'd like.  This stores state data in /var/lib like every other service stores state data in /var/lib: /var/lib/httpd, /var/lib/ldap, /var/lib/mysql, et c.  If you have a problem with this you have a problem with LSB.

misc/systemd/ipfs.service
sneak: To what value would you like it changed?  4TB is well within the storage capacity of most small hard drives sold today.  I have over 10TB in my backpack and 1TB in my pocket.  It also does not immediately use 4TB as you well know; other services such as MySQL or OpenLDAP or Postgres will happily use as much space in /var/lib as is required to store whatever they are asked to store.

misc/systemd/ipfs.service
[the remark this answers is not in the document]
sneak: If this install location is not acceptable, what path would you like me to use in this file?

misc/systemd/ipfs.service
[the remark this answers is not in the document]
sneak: I have not found this systemd unit file to be annoying in several weeks of use, and if an administrator that uses the example file does, it is trivial to remove any lines that they find to be cumbersome or unnecessary.

misc/systemd/ipfs.service
sneak: Shall I remove this line?  If not, to which value would you like it changed?

misc/systemd/ipfs.service
sneak: > This is anyone who shares a linux machine with other people in a household

Is this the use case we are targeting?

misc/systemd/ipfs.service
sneak: @lanzafame To what value would you like it changed?

misc/systemd/ipfs.service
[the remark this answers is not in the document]
sneak: @lanzafame Yes, this works on Ubuntu, Mint, Debian, et c.  Do you want this line removed?  It means that in the majority case it requires additional commands (beyond this service file) to make the daemon start correctly.

misc/systemd/ipfs.service
lanzafame: This is really something that should be provided by the user via the config file. It is by is nature not something particularly generic. Thinking about this again, service files shouldn't need to be changed to adjust what is a config value.

misc/systemd/ipfs.service
Stebalien: The commands do work offline but, in this case, they'll only work if you happen to be the `ipfs` user or root.

Side note: Units generally should *not* use `Requires=network.target`, only `After=network.target` (see the documentation in `man 7 systemd.special`).

misc/systemd/ipfs.service
```diff
@@ -0,0 +1,31 @@
+[Unit]
+Description=ipfs p2p daemon
+After=network.target
+Requires=network.target
+
+[Service]
+Type=simple
```
Stebalien: No need to reiterate the default.

misc/systemd/ipfs.service
```diff
@@ -0,0 +1,31 @@
...
+User=ipfs
+RestartSec=1
+Restart=always
+PermissionsStartOnly=true
+Nice=18
```
Stebalien: There's no reason to specify this in the default unit. The user can use a drop-in to set it if desired but it's better to leave defaults at their default values for predictability.

misc/systemd/ipfs.service
Stebalien: sysconfig is a a redhat thing. The recommended way to customize a systemd unit (or set an environment variable) is to put a drop-in in `/etc/systemd/system/ipfs.service.d/`

misc/systemd/ipfs.service
```diff
@@ -0,0 +1,31 @@
...
+StateDirectory=/var/lib/ipfs
+Environment=IPFS_PATH=/var/lib/ipfs
+Environment=HOME=/var/lib/ipfs
+LimitNOFILE=8192
+Environment=IPFS_FD_MAX=8192
+EnvironmentFile=-/etc/sysconfig/ipfs
+StandardOutput=journal
```
Stebalien: This is the default.

misc/systemd/ipfs.service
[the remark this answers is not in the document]
Stebalien: The correct way to do this with systemd is to write a `sysusers.d` file:

```
u ipfs - "ipfs daemon" /var/lib/ipfs
```

Running an add-user command on start is very unusual for a systemd unit.

misc/systemd/ipfs.service
```diff
@@ -0,0 +1,31 @@
...
+WorkingDirectory=/var/lib/ipfs
+ExecStartPre=-adduser --system --group --home /var/lib/ipfs ipfs
+ExecStartPre=-mkdir /var/lib/ipfs
```
Stebalien: As with the user account creation, the systemd way to do this is to write a `tmpfiles.d` file:

```
d /var/lib/ipfs 0750 ipfs ipfs -
```

misc/systemd/ipfs.service
```diff
@@ -0,0 +1,31 @@
...
+ExecStartPre=-/bin/chown ipfs:ipfs /var/lib/ipfs
+ExecStartPre=-/bin/chmod u+rwx /var/lib/ipfs
+ExecStartPre=-chpst -u ipfs /usr/bin/env ipfs init --profile=badgerds
+ExecStartPre=-chpst -u ipfs /usr/bin/env ipfs config profile apply server
+ExecStartPre=-chpst -u ipfs /usr/bin/env ipfs config Datastore.StorageMax "4000GB"
```
Stebalien: Let's keep the example systemd config to the defaults. Users can customize it how they will.

misc/systemd/ipfs.service
```diff
@@ -0,0 +1,31 @@
...
+ExecStart=/usr/bin/env ipfs daemon --enable-namesys-pubsub --enable-pubsub-experiment --migrate=true
```
Stebalien: Pass `--init` to initialize the repo if not already initialized. That way, we can skip the `ExecStartPre` command.

misc/systemd/ipfs.service
Stebalien: So, this isn't quite the correct usage of `StateDirectory`. It's primarily used for the `DynamicUser` feature and should be relative (that is, just `ipfs`).

So, we should either drop this *or* switch this unit to a dynamic user. If we go with the latter, we'd want to:

1. Not have a tmpfiles/sysusers config file and *not* create a user account.
2. Set `DynamicUser=true`
3. Set `StateDirectory=ipfs`

The *issue* with this is that, if there's a UID conflict, systemd will have to `chown` the entire directory tree (which, depending on the datastore, could be very expensive).

On the other hand, it means one config file versus 3.

misc/systemd/ipfs.service
Stebalien: So, that line is actually sourcing the environment variables in `/etc/sysconfig/ipfs` (not related to the IPFS config).

Stebalien:
(closing due to inactivity, please comment if you'd like this reopened)

PR #5602, opened 2018-10-15
add support for IPLD over IPNS

Stebalien:
It's not *perfect* but at least `ipfs name resolve` can resolve IPNS records pointing to IPLD paths. `ipfs dag get /ipns` still doesn't work but we can *make* it work once we switch to CoreAPI.

I thought this would be difficult, apparently not...

### DIFF
--- a/namesys/base.go
+++ b/namesys/base.go
@@ -25,15 +25,6 @@ func resolve(ctx context.Context, r resolver, name string, options *opts.Resolve
 		}
 		log.Debugf("resolved %s to %s", name, p.String())
 
-		if strings.HasPrefix(p.String(), "/ipfs/") {
-			// we've bottomed out with an IPFS path
-			return p, nil
-		}
-
-		if depth == 1 {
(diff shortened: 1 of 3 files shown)


namesys/base.go
```diff
@@ -25,15 +25,6 @@ func resolve(ctx context.Context, r resolver, name string, options *opts.Resolve
 		}
 		log.Debugf("resolved %s to %s", name, p.String())
 
-		if strings.HasPrefix(p.String(), "/ipfs/") {
```
Stebalien: I've changed this check to "is not /ipns/". That way, we continue resolving until we can no longer resolve.

magik6k:
This code is mostly rewritten in https://github.com/ipfs/go-ipfs/pull/5404. We should also consider dropping the `prefixes` thing as it's always hard coded to `/ipns/` anyways.

Stebalien:
Let's block this on that PR then.

Issue #5601, opened 2018-10-15
Add slash for dir in `ipfs files ls`


overbool:
We can get the follow result using `ipfs files ls`:

```
dir-a
dir-b
a
b
c
```

Is this a good advice that add the slash for directory to distinguish directory from file?

just like the follow:
```
dir-a/
dir-b/
a
b
c
```

Stebalien:
We can but, unfortunately, this will force us to fetch the first block of every file to resolve the type. However, I don't think this'll be that much of an issue in practice, especially for MFS. This is *also* what `ipfs ls` does so we should probably just mirror that.

Thoughts @schomatis?

schomatis:
We could add the slash when using the `-l` flag which already fetches that information,

https://github.com/ipfs/go-ipfs/blob/9bf4e4145ea897ced2f9aa2087296b3beda6e941/core/commands/files.go#L456-L460

PR #5605, opened 2018-10-16
fix(files): add slash for dir

overbool:
Fixes: https://github.com/ipfs/go-ipfs/issues/5601

License: MIT

### DIFF
--- a/test/sharness/t0250-files-api.sh
+++ b/test/sharness/t0250-files-api.sh
@@ -260,6 +260,13 @@ test_files_api() {
     verify_dir_contents /cats/this/is/a/dir
   '
 
+  test_expect_success "dir has correct name" '
+    DIR_HASH=$(ipfs files stat /cats/this --hash) &&
+    echo "this/	$DIR_HASH	0" > ls_dir_expected &&
+    ipfs files ls -l /cats | grep this/ > ls_dir_actual &&
+    test_cmp ls_dir_expected ls_dir_actual
+  '
+
   test_expect_success "can copy file into new dir $EXTRA" '
     ipfs files cp /ipfs/$FILE3 /cats/this/is/a/dir/file3
   '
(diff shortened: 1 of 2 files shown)


schomatis:
Yes, we can later evaluate adding it for both cases if you feel like it, but this seems like an inexpensive first approach.

schomatis:
Could you add a test case for it? (Or modify the existing ones, I'm not sure if this change will break them, we sometimes check with `grep` for certain strings like the file name.)

test/sharness/t0250-files-api.sh
```diff
@@ -260,6 +260,12 @@ test_files_api() {
     verify_dir_contents /cats/this/is/a/dir
   '
 
+  test_expect_success "dir has correct name" '
+    echo "this/" > ls_dir_expected &&
+    ipfs files ls -l /cats | grep this/ | awk '\''{print $1}'\'' > ls_dir_actual &&
```
schomatis: I think the standard way is to output "filename hash size" in the `ls_dir_actual` instead of using `awk`+`grep`.

overbool: @schomatis yeah, I also think that "filename hash size" is a better way. However, the `hash` is unfixed, because `ipfs files mkdir` uses different `arg`: 
https://github.com/ipfs/go-ipfs/blob/9bf4e4145ea897ced2f9aa2087296b3beda6e941/test/sharness/t0250-files-api.sh#L109-L116

schomatis: We should then save the dir hash generated on every call, although I'm not sure if that's the best approach, @magik6k WDYT?

magik6k: yep, we should grab the generated hash, save it and compare that

overbool:
@schomatis @magik6k I had modified the test case.

Issue #4186, opened 2017-08-30
`ipfs ls` should behave like `ls` when called on individual files.


Stebalien:
#### Version information:

> go-ipfs version: 0.4.11-dev-48476b292
Repo version: 5
System version: amd64/linux
Golang version: go1.8.3

#### Type:

Bug?

#### Severity:

Very Low

#### Description:

In the past, `ipfs ls /ipfs/Qm.../file.txt` listed the internal blocks in the file. Now, it just prints an error. In the near future, it looks like it will go back to listing internal blocks (restoring the original behavior).

Really, it should behave like the standard unix `ls` command and print the filename (if available in this context) and, in the future when we support them, any unix permissions on the file. Basically, internally, we should be listing the parent directory and then filtering out all files except the target file.

The tricky part is what to do when we *don't* know the parent directory (when the file is named by hash). I'd be inclined to print the file hash as the file name and set the permissions to 0777.

Kubuxu:
The permission might be better being `444`.

Stebalien:
Ah. Good point but I'd say 555 (+x) by default for convenience. Otherwise, one wouldn't be able to, e.g., execute `/ipfs/QmId...`.

Kubuxu:
You are not able to execute it from fuse interface right now either way.

Stebalien:
True. But once start storing an executable bit, I'd mark bare files as executable.

PR #5611, opened 2018-10-18
Add --stream option to `ls` command

### LINKED ISSUE #5600: Add option to stream results of `ipfs ls`
Even with many speed improvements, when LS'ing on a large sharded directory, it takes a while to see any results. Since directories are actually already being fetched, it would be nice to start streaming the results as soon as we get them.

Opening this for discussion on what changes should be made across:
Commands
CoreAPI
HTTP
go-unixfs/hamt

child of #5487

hannahhoward:
# Features

Adds the option to stream results from ls -- so that large directories (particularly HAMT dirs) start returning data immediately instead of waiting for the whole DAG to be walked

Fixes #5600 
Requires https://github.com/ipfs/go-unixfs/pull/39

# Implementation

- Update LS Command to use go-ipld-cmds library
- Update LS Command with --stream option
   - Consume EnumLinksAsync in Ls command instead of synchronous Links method
   - Support breaking up of command output. Instead of LsObject having to be a full directory, it can  represent the beginning of a directory, one more more links in a directory, or the end of the directory - so that outputs can be CAT together to produce the same combined output

# For Discussion

- Not mergable until go-unixfs is updated for referenced PR
- What is our preferred method for unit testing commands?
- Suggested refactors for ls command file? Feels a bit overloaded and complicated

### DIFF
--- a/core/commands/ls.go
+++ b/core/commands/ls.go
@@ -111,10 +111,9 @@ The JSON output contains type information.
 		ng := merkledag.NewSession(req.Context, nd.DAG)
 		ro := merkledag.NewReadOnlyDagService(ng)
 
-		output := make([]LsObject, len(req.Arguments))
-
 		stream, _ := req.Options[lsStreamOptionName].(bool)
 		if !stream {
+			output := make([]LsObject, len(req.Arguments))
 
 			for i, dagnode := range dagnodes {
 				dir, err := uio.NewDirectoryFromNode(ro, dagnode)
@@ -148,14 +147,6 @@ The JSON output contains type information.
 			return cmds.EmitOnce(res, &LsOutput{output})
 		}
 
-		outputLinks := make([]LsLink, 1)
-		for i, path := range paths {
-			output[i] = LsObject{
-				Hash:  path,
-				Links: nil,
-			}
-		}
-
 		for i, dagnode := range dagnodes {
 			dir, err := uio.NewDirectoryFromNode(ro, dagnode)
 			if err != nil && err != uio.ErrNotADir {
@@ -169,9 +160,17 @@ The JSON output contains type information.
 				linkResults = dir.EnumLinksAsync(req.Context)
 			}
 
-			output[i].Links = outputLinks
-
 			for linkResult := range linkResults {
+				output := make([]LsObject, len(req.Arguments))
+
+				for i, path := range paths {
+					output[i] = LsObject{
+						Hash:  path,
+						Links: nil,
+					}
+				}
+				outputLinks := make([]LsLink, 1)
+
 				if linkResult.Err != nil {
 					return linkResult.Err
 				}
@@ -181,11 +180,11 @@ The JSON output contains type information.
 					return err
 				}
 				outputLinks[0] = *lsLink
+				output[i].Links = outputLinks
 				if err = res.Emit(&LsOutput{output}); err != nil {
 					return err
 				}
 			}
-			output[i].Links = nil
 		}
 		return nil
 	},
@@ -193,9 +192,21 @@ The JSON output contains type information.
 		cmds.CLI: func(res cmds.Response, re cmds.ResponseEmitter) error {
 			req := res.Request()
 			headers, _ := req.Options[lsHeadersOptionNameTime].(bool)
+			stream, _ := req.Options[lsStreamOptionName].(bool)
+
+			// in streaming mode we can't automatically align the tabs
+			// so we take a best guess
+			var minTabWidth int
+			if stream {
+				minTabWidth = 10
+			} else {
+				minTabWidth = 1
+			}
+
 			multipleFolders := len(req.Arguments) > 1
 			lastDirectoryWritten := -1
-			tw := tabwriter.NewWriter(os.Stdout, 1, 2, 1, ' ', 0)
+
+			tw := tabwriter.NewWriter(os.Stdout, minTabWidth, 2, 1, ' ', 0)
 			for {
 				v, err := res.Next()
 				if err != nil {
--- a/core/commands/ls.go
+++ b/core/commands/ls.go
@@ -37,7 +37,9 @@ type LsObject struct {
 // LsOutput is a set of printable data for directories,
 // it can be complete or partial
 type LsOutput struct {
-	Objects        []LsObject
+	Objects []LsObject
+	// temporary flag to help us figure out where we are in the process of ls-ing
+	// the directory when we are streaming
 	LastObjectHash string
 }
 
@@ -66,7 +68,7 @@ The JSON output contains type information.
 	Options: []cmdkit.Option{
 		cmdkit.BoolOption(lsHeadersOptionNameTime, "v", "Print table headers (Hash, Size, Name)."),
 		cmdkit.BoolOption(lsResolveTypeOptionName, "Resolve linked objects to find out their types.").WithDefault(true),
-		cmdkit.BoolOption(lsStreamOptionName, "s", "Stream directory entries as they are found."),
+		cmdkit.BoolOption(lsStreamOptionName, "s", "Enable exprimental streaming of directory entries as they are traversed."),
 	},
 	Run: func(req *cmds.Request, res cmds.ResponseEmitter, env cmds.Environment) error {
 		nd, err := cmdenv.GetNode(env)
--- a/core/commands/ls.go
+++ b/core/commands/ls.go
@@ -3,11 +3,9 @@ package commands
 import (
 	"fmt"
 	"io"
-	"os"
 	"text/tabwriter"
 
 	cmdenv "github.com/ipfs/go-ipfs/core/commands/cmdenv"
-	e "github.com/ipfs/go-ipfs/core/commands/e"
 	iface "github.com/ipfs/go-ipfs/core/coreapi/interface"
 
 	cid "gx/ipfs/QmR8BauakNcBa3RbE4nbQu76PDiJgoQgz8AJdhJuiU4TAw/go-cid"
@@ -39,7 +37,8 @@ type LsObject struct {
 // LsOutput is a set of printable data for directories,
 // it can be complete or partial
 type LsOutput struct {
-	Objects []LsObject
+	Objects        []LsObject
+	LastObjectHash string
 }
 
 const (
@@ -112,6 +111,8 @@ The JSON output contains type information.
 		ro := merkledag.NewReadOnlyDagService(ng)
 
 		stream, _ := req.Options[lsStreamOptionName].(bool)
+		lastObjectHash := ""
+
 		if !stream {
 			output := make([]LsObject, len(req.Arguments))
 
@@ -144,7 +145,7 @@ The JSON output contains type information.
 				}
 			}
 
-			return cmds.EmitOnce(res, &LsOutput{output})
+			return cmds.EmitOnce(res, &LsOutput{output, lastObjectHash})
 		}
 
 		for i, dagnode := range dagnodes {
@@ -161,15 +162,6 @@ The JSON output contains type information.
 			}
 
 			for linkResult := range linkResults {
-				output := make([]LsObject, len(req.Arguments))
-
-				for i, path := range paths {
-					output[i] = LsObject{
-						Hash:  path,
-						Links: nil,
-					}
-				}
-				outputLinks := make([]LsLink, 1)
 
 				if linkResult.Err != nil {
 					return linkResult.Err
@@ -179,21 +171,24 @@ The JSON output contains type information.
 				if err != nil {
 					return err
 				}
-				outputLinks[0] = *lsLink
-				output[i].Links = outputLinks
-				if err = res.Emit(&LsOutput{output}); err != nil {
+				output := []LsObject{
+					{
+						Hash:  paths[i],
+						Links: []LsLink{*lsLink},
+					},
+				}
+				if err = res.Emit(&LsOutput{output, lastObjectHash}); err != nil {
 					return err
 				}
+				lastObjectHash = paths[i]
 			}
 		}
 		return nil
 	},
-	PostRun: cmds.PostRunMap{
-		cmds.CLI: func(res cmds.Response, re cmds.ResponseEmitter) error {
-			req := res.Request()
+	Encoders: cmds.EncoderMap{
+		cmds.Text: cmds.MakeTypedEncoder(func(req *cmds.Request, w io.Writer, out *LsOutput) error {
 			headers, _ := req.Options[lsHeadersOptionNameTime].(bool)
 			stream, _ := req.Options[lsStreamOptionName].(bool)
-
 			// in streaming mode we can't automatically align the tabs
 			// so we take a best guess
 			var minTabWidth int
@@ -204,56 +199,36 @@ The JSON output contains type information.
 			}
 
 			multipleFolders := len(req.Arguments) > 1
-			lastDirectoryWritten := -1
+			lastObjectHash := out.LastObjectHash
 
-			tw := tabwriter.NewWriter(os.Stdout, minTabWidth, 2, 1, ' ', 0)
-			for {
-				v, err := res.Next()
-				if err != nil {
-					if err == io.EOF {
-						if multipleFolders {
-							fmt.Fprintln(os.Stdout)
+			tw := tabwriter.NewWriter(w, minTabWidth, 2, 1, ' ', 0)
+
+			for _, object := range out.Objects {
+
+				if object.Hash != lastObjectHash {
+					if multipleFolders {
+						if lastObjectHash != "" {
+							fmt.Fprintln(tw)
 						}
-						return nil
+						fmt.Fprintf(tw, "%s:\n", object.Hash)
 					}
-
-					return err
+					if headers {
+						fmt.Fprintln(tw, "Hash\tSize\tName")
+					}
+					lastObjectHash = object.Hash
 				}
 
-				output, ok := v.(*LsOutput)
-				if !ok {
-					return e.TypeErr(output, v)
-				}
+				for _, link := range object.Links {
+					if link.Type == unixfs.TDirectory {
+						link.Name += "/"
+					}
 
-				for i, object := range output.Objects {
-					if len(object.Links) == 0 {
-						continue
-					}
-					if i > lastDirectoryWritten {
-						if i > 0 {
-							if multipleFolders {
-								fmt.Fprintln(tw)
-							}
-						}
-						if multipleFolders {
-							fmt.Fprintf(tw, "%s:\n", object.Hash)
-						}
-						if headers {
-							fmt.Fprintln(tw, "Hash\tSize\tName")
-						}
-						lastDirectoryWritten = i
-					}
-					for _, link := range object.Links {
-						if link.Type == unixfs.TDirectory {
-							link.Name += "/"
-						}
-
-						fmt.Fprintf(tw, "%s\t%v\t%s\n", link.Hash, link.Size, link.Name)
-					}
+					fmt.Fprintf(tw, "%s\t%v\t%s\n", link.Hash, link.Size, link.Name)
 				}
-				tw.Flush()
 			}
-		},
+			tw.Flush()
+			return nil
+		}),
 	},
 	Type: LsOutput{},
 }
--- a/core/commands/ls.go
+++ b/core/commands/ls.go
@@ -3,6 +3,7 @@ package commands
 import (
 	"fmt"
 	"io"
+	"os"
 	"text/tabwriter"
 
 	cmdenv "github.com/ipfs/go-ipfs/core/commands/cmdenv"
@@ -29,20 +30,16 @@ type LsLink struct {
 }
 
 // LsObject is an element of LsOutput
-// It can represent a whole directory, a directory header, one or more links,
-// Or a the end of a directory
+// It can represent all or part of a directory
 type LsObject struct {
-	Hash      string
-	Links     []LsLink
-	HasHeader bool
-	HasLinks  bool
-	HasFooter bool
+	Hash  string
+	Links []LsLink
 }
 
-// LsOutput is a set of printable data for directories
+// LsOutput is a set of printable data for directories,
+// it can be complete or partial
 type LsOutput struct {
-	MultipleFolders bool
-	Objects         []LsObject
+	Objects []LsObject
 }
 
 const (
@@ -114,10 +111,10 @@ The JSON output contains type information.
 		ng := merkledag.NewSession(req.Context, nd.DAG)
 		ro := merkledag.NewReadOnlyDagService(ng)
 
+		output := make([]LsObject, len(req.Arguments))
+
 		stream, _ := req.Options[lsStreamOptionName].(bool)
-		multipleFolders := len(req.Arguments) > 1
 		if !stream {
-			output := make([]LsObject, len(req.Arguments))
 
 			for i, dagnode := range dagnodes {
 				dir, err := uio.NewDirectoryFromNode(ro, dagnode)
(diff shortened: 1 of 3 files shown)


alanshaw:
ping @achingbrain

hannahhoward:
Rebased on master, pending updated go-unixfs. However, in the meantime, ready for review @eingenito @Stebalien @magik6k

core/commands/ls.go
```diff
@@ -60,158 +70,225 @@ The JSON output contains type information.
...
 		}
-
-		output := make([]LsObject, len(req.Arguments()))
-		ng := merkledag.NewSession(req.Context(), nd.DAG)
+		ng := merkledag.NewSession(req.Context, nd.DAG)
 		ro := merkledag.NewReadOnlyDagService(ng)
 
+		stream, _ := req.Options[lsStreamOptionName].(bool)
+		multipleFolders := len(req.Arguments) > 1
+		if !stream {
```
eingenito: I'm fuzzy on the details here but would it work to just extract the streaming version of this code and then call it and buffer it in the non-streaming case, returning it with EmitOnce? Are we trying to preserve a stable order in the non-streaming case? Could/should we sort after in that case? Or are all these a sort of breaking API change that we want to avoid?

hannahhoward:
@magik6k the issue is while the code for this is merged into go-unixfs, no new version is published. Is there an appropriate process to update and publish a subpackage like go-unixfs?

magik6k:
You just run `gx release [patch/minor/major]`, pin the hash somewhere (like using pinbot), push.

Then if the thing you are trying to update has dependencies, run `gx-workspace bubble-list [package-name]` in go-ipfs directory and for each dependency open a PR with updates (`gx update QmDependency` for each changed dependency), and any changes if necessary.

Alternatively, if there are no breaking/few changes and more than say 2-3 dependencies you should try `gx-workspace update` (see its --help for more).

Stebalien:
This is now unblocked. Can someone review it?

core/commands/ls.go
```diff
@@ -21,24 +21,34 @@ import (
...
+// It can represent a whole directory, a directory header, one or more links,
+// Or a the end of a directory
 type LsObject struct {
-	Hash  string
-	Links []LsLink
+	Hash      string
+	Links     []LsLink
+	HasHeader bool
+	HasLinks  bool
+	HasFooter bool
```
kevina: I would like to avoid the need for the extra flags.  We should instead just send data as we get it and let the client figure out the rest.

If a flag at this level is really needed I wold like to use something like `Incomplete` which will signal to the client that not all the links for the given hash has been delivered.

core/commands/ls.go
```diff
@@ -21,24 +21,34 @@ import (
...
 }
 
+// LsOutput is a set of printable data for directories
 type LsOutput struct {
-	Objects []LsObject
+	MultipleFolders bool
```
kevina: It is unclear what this means or it's purpose.

I instead would like to have a single flag `Incomplete` or something like that to signal to the client that more data is coming, however I am not sure if that is even needed, because there might be other ways for the client to figure this out.

core/commands/ls.go
```diff
@@ -60,158 +70,225 @@ The JSON output contains type information.
...
+}
+func newDirectoryHeaderLsObject(hash string) LsObject {
+	return LsObject{hash, nil, true, false, false}
+}
+func newDirectoryLinksLsObject(links []LsLink) LsObject {
+	return LsObject{"", links, false, true, false}
+}
+func newDirectoryFooterLsObject() LsObject {
+	return LsObject{"", nil, false, false, true}
+}
```
schomatis: Agree with @kevina here, all these flags feel too cumbersome, is there an API in `go-ipfs-cmds` to signal end of data?

core/commands/ls.go
```diff
@@ -60,158 +70,225 @@ The JSON output contains type information.
...
+							link.Name += "/"
+						}
+
+						fmt.Fprintf(tw, "%s\t%v\t%s\n", link.Hash, link.Size, link.Name)
 					}
-					fmt.Fprintf(w, "%s\t%v\t%s\n", link.Hash, link.Size, link.Name)
 				}
-				if len(output.Objects) > 1 {
-					fmt.Fprintln(w)
+				if object.HasFooter {
```
schomatis: For example, `HasFooter`: can the link results from different directories be interleaved? Or this is just a signal to print a `\n` when no more links are available in one directory? (and if so, could that just be added to the `object.HasHeader` `if` clause?)

It seems like we are passing too much information through the `Emit` channels (although that may be the customary way in our commands lib).

kevina:
@Stebalien This is now blocking https://github.com/ipfs/go-ipfs/pull/5464.  Could we get https://github.com/ipfs/go-ipfs/pull/5663 in first or separate out the switch to a new commands library into a separate p.r.

kevina:
If required I would be happy to either separate out the command lib. stuff in a separate commit or rebase this on top of #5663.  It shouldn't be _that_ hard and I am not sure I am happy with how things are implemented right now.

core/commands/ls.go
```diff
@@ -60,158 +70,225 @@ The JSON output contains type information.
...
-					if pn, ok := linkNode.(*merkledag.ProtoNode); ok {
-						d, err := unixfs.FSNodeFromBytes(pn.Data())
-						if err != nil {
-							res.SetError(err, cmdkit.ErrNormal)
-							return
-						}
-						t = d.Type()
-					}
+			output[0] = newDirectoryHeaderLsObject(paths[i])
+			if err = res.Emit(&LsOutput{multipleFolders, output}); err != nil {
```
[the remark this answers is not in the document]
kevina: But we don't use it that much, which is why I think it is better to be consistent.  My objection is minor though.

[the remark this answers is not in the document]
magik6k: This was discussed at least once before, the conclussion was that when a funciton only returns an error it's preferred to use the single line syntax.

hannahhoward:
@Stebalien + @kevina I have made the changes requested.

Apologies @kevina you are totally right about not having the extra flags and I just did not understand the commands lib well.

Hopefully this covers everything. However, in the process of doing this change I discovered that the fixes I put into go-unixfs to prevent panics in a cancel do not work 100% of the time. And I had to create a seperate fix here in go-merkledag: https://github.com/ipfs/go-merkledag/pull/19

@Stebalien if I can get a thumbs up on that one I can merge and update, so we can get this merged.

core/commands/ls.go
```diff
@@ -1,44 +1,51 @@
...
 
-	cmds "github.com/ipfs/go-ipfs/commands"
+	cmdenv "github.com/ipfs/go-ipfs/core/commands/cmdenv"
 	e "github.com/ipfs/go-ipfs/core/commands/e"
 	iface "github.com/ipfs/go-ipfs/core/coreapi/interface"
 
 	offline "gx/ipfs/QmPpnbwgAuvhUkA9jGooR88ZwZtTUHXXvoQNKdjZC6nYku/go-ipfs-exchange-offline"
+	unixfs "gx/ipfs/QmQFXpLVAgoQsgsdJspTkHcVpkLznHSTovwaBcBPvvLNoG/go-unixfs"
+	uio "gx/ipfs/QmQFXpLVAgoQsgsdJspTkHcVpkLznHSTovwaBcBPvvLNoG/go-unixfs/io"
+	unixfs_pb "gx/ipfs/QmQFXpLVAgoQsgsdJspTkHcVpkLznHSTovwaBcBPvvLNoG/go-unixfs/pb"
```
kevina: Is there a reason for the underscore in `unixfs_pb`?  I release it makes it easier to read, but is it standard practice?

CC @Stebalien

hannahhoward:
@Stebalien is there anything else holding up merge on this? I don't believe there should be.... also though having build issues seemingly unrelated to my code in CI.

core/commands/ls.go
```diff
@@ -60,158 +67,243 @@ The JSON output contains type information.
...
+
+			for i, dagnode := range dagnodes {
+				dir, err := uio.NewDirectoryFromNode(ro, dagnode)
+				if err != nil && err != uio.ErrNotADir {
+					return fmt.Errorf("the data in %s (at %q) is not a UnixFS directory: %s", dagnode.Cid(), paths[i], err)
+				}
+
+				var links []*ipld.Link
+				if dir == nil {
+					links = dagnode.Links()
```
Stebalien: Gah... https://github.com/ipfs/go-ipfs/issues/4186. I'm going to just break this one day...

core/commands/ls.go
```diff
@@ -60,158 +67,243 @@ The JSON output contains type information.
...
-
-			return buf, nil
 		},
 	},
 	Type: LsOutput{},
 }
+
+func makeDagNodeLinkResults(req *cmds.Request, dagnode ipld.Node) <-chan unixfs.LinkResult {
+	linkResults := make(chan unixfs.LinkResult)
+	go func() {
```
magik6k: So since we need to keep the links in memory for the time of iteration anyways, I think it might be better to just do

```go
links := dagnode.Links()
linkResults := make(chan unixfs.LinkResult, len(links))
for _, l := range dagnode.Links() {
 linkResults <- unixfs.LinkResult{
  Link: l,
  Err:  nil,
 }
}
close(linkResults)
return linkResults

```

core/commands/ls.go
```diff
@@ -60,158 +67,243 @@ The JSON output contains type information.
...
 				}
 			}
 		}
-
-		res.SetOutput(&LsOutput{output})
+		return nil
 	},
-	Marshalers: cmds.MarshalerMap{
-		cmds.Text: func(res cmds.Response) (io.Reader, error) {
+	PostRun: cmds.PostRunMap{
```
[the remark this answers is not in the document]
hannahhoward: see conversation above -- @kevina preferred -- and I think he was right, the output from the run be the same when streamed as it was when it's not, and that requires logic in the postRun to produce the right text output. however, I can add a cmd.MakeTypedEncoder and use re.Emit in the PostRun instead.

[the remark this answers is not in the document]
hannahhoward: I was following the structure of ipfs add here.

core/commands/ls.go
```diff
@@ -60,158 +67,243 @@ The JSON output contains type information.
...
-				for _, link := range object.Links {
-					if link.Type == unixfs.TDirectory {
-						link.Name += "/"
+
+				for i, object := range output.Objects {
+					if len(object.Links) == 0 {
+						continue
+					}
+					if i > lastDirectoryWritten {
+						if i > 0 {
```
magik6k: This part could be simplified:
```
if multipleFolders {
 if i > 0 {
  fmt.Fprintln(tw)
 }
 fmt.Fprintf(tw, "%s:\n", object.Hash)
}
```

core/commands/ls.go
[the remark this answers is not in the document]
magik6k: I see, `ipfs add` is probably a bad example as it uses PostRun mainly for progress bar. It should be possible to do this using `Encoders`, which is the preferred way.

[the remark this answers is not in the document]
kevina: The problem with using Encoders is that they don't support Stderr output.  This is when I generally switch to PostRun.  Here that is not the case so we should still use an Encoder.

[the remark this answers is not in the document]
hannahhoward: So the issue basically has to do with the output of the header/footer lines surrounding the directory objects, which requires either passing information about where exactly you are in a stream in the objects themselves, or using a post run.

I originally had this setup with additional flags on the LSObject so that it correct output could be generated with an encoder (I was using cmd.MakeEncoder but converting to cmd.MakeTypeEncoder would have been trivial):
https://github.com/ipfs/go-ipfs/pull/5611/commits/135d1304143a2adbdb79abd3185f970addab23c1

However @kevina and @schomatis felt it would be better to not add any additional flags to the LSObject and instead handle the output re-arrangement at the CLI level:
https://github.com/ipfs/go-ipfs/pull/5611#discussion_r230522138
https://github.com/ipfs/go-ipfs/pull/5611#discussion_r230539744

So this was the rewrite based on that feedback.

I don't actually have a particular dog in this fight, and am fine with either way, but I guess I do agree that chaining together the right header/footer output around directories is a CLI level concern, that is analogous the progress bar on add. I am fine doing it either way -- and perhaps there is a hybrid approach I'm not seeing.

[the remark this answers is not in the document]
hannahhoward: ok I will try to code this without using PostRun, but I'm not sure how to get the right output between directories on a stream without maintaining some kind of state between calls to the Encoder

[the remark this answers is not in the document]
kevina: @hannahhoward I agree that you will need to maintain some sort of state, I am not sure if you can do this with Encoders, so in that case you might need to use PostRun.  I think using `PostRun` is better than the extra flags.

Using PostRun might break using the API when using the text encoder, see https://github.com/ipfs/go-ipfs/pull/5674#issuecomment-436777547.  So perhaps a hybrid solution like @Stebalien proposed may be in order, but really I am just guessing now.

@Stebalien thoughts.

Stebalien:
Concrete issues:

1. Sending an empty `LsObject` per directory we're not listing feels a bit funky. Not using the index in `LsOutput.Objects` and instead relying on the name could, I think, improve this slightly (we'll have to count the number of arguments to determine if we need to print headers) but I'll discuss other alternatives below.
2. This use of PostRun breaks: `curl 'http://localhost:5001/api/v0/ls?arg=/ipns/ipfs.io&encoding=text'`. Unfortunately, using encoders will be tricky given the current streaming API because we need to know when to print the directory headers.

So, let's talk alternatives:

First, the obvious answer, we can just introduce a new command. However, I'd rather not expand our CLI like that.

We could also emit:

```go
LsOutput{
    Objects: []LsObject{
        {
            Hash: "FirstDirectory",
        },
    },
}
LsOutput{
    Objects: []LsObject{
        {
            Links: []LsLink{
                {
                    Name: "a",
                    Hash: "...",
                    Type: ...,
                    ...
                },
            },
        },
    },
}
LsOutput{
    Objects: []LsObject{
        {
            Links: []LsLink{
                {
                    Name: "b",
                    Hash: "...",
                    Type: ...,
                    ...
                },
            },
        },
    },
}
LsOutput{
    Objects: []LsObject{
        {
            Hash: "SecondDirectory",
        },
    },
}
LsOutput{
    Objects: []LsObject{
        {
            Links: []LsLink{
                {
                    Name: "foo",
                    Hash: "...",
                    Type: ...,
                    ...
                },
            },
        },
    },
}
...
```

That is, don't repeat the directory names each time. We'd just assume that any time we get a "link" without a directory, it belongs to the previous same directory as the previous link.

---

OK, so that covers the alternatives that stick with the *current* types. However, we don't actually *have* to use them. We could switch to (something like):

```go
type LsOutput {
	// One of "", "directory-entry", "streaming-directory".
	Type string `json:",omitempty"`
	Output json.RawMessage
}

type LsDirEntry struct {
	Name, Hash string
	Size       uint64
	Type       unixfspb.Data_DataType
}

type LsObject struct {
	Hash  string
	Links []DirEntry
}

type LsOutput struct {
	Objects []LsObject
}
type LsDirectory struct {
	Name string
}
```

We'd then select the right type for `Output` based on `Type`. This would allow us to continue using an encoder and should be slightly less hacky.

However, it *is* a much bigger change and I haven't tried it in practice. Thoughts?

Really, there is no clean solution here (other than break everything and switch to a new API, which we'll do eventually anyways).

hannahhoward:
@magik6k @Stebalien @kevina 

Ok, I've rewritten this to work with cmd.MakeTypedEncoder and a single extra piece of data.... I can clean this up if it works for people.

Otherwise I'll look at @Stebalien 's comments on Monday and see if there are other approaches.

hannahhoward:
@Stebalien I went ahead and removed the transmission of empty blocks by relying on hashs intead of indexes. This is about as close as I can get to reasonable complexity (one extra flag) without a major refactor of the transmitted data.

hannahhoward:
sorry build issue, resolved now.

kevina:
Unless there is an urgent need to support the `--stream` option I think we should perhaps hold off on this until we can arrive at a better solution.

Using any sort of flags suck as `LastObjectHash` is encoding an implementation detail into the API, which is something we should probably avoid.  I am more okay with a flag that indicates a record is incomplete as that strikes me as being less tied to the implementation.

In order for _me_ to be happy with this I think we need to either (1) just use PostRun, or (2) fix the encoders so that they can maintain some sort of state between runs for the purposes of presents the results better, such as proving headers and footers, or aligning the output across multiple records.  As we move more interfaces to streaming the results rather then present them all at once this is going to become more important.

hannahhoward:
@kevina

I mostly don't have a dog in this fight. I agree that the ideal would be to come up with a global solution to pass state between calls to encoders. I don't think a PostRun is the right solution unless you can do it without breaking encoding=text over the API.

I don't think deferring this ticket is a good idea as it's at least a part of an Q4 OKR that's P1, and and there are actual IPFS users who want fast response time LS-ing large directory. But I would say that's not a decision for either of us and I would defer to @Stebalien and @eingenito.

eingenito:
So naturally I defer to @stebalien about what's acceptable to merge, and I haven't been following the details. But I concur with @hannahhoward that it seems important to move this forward _somehow_. This is a request from a partner that @momack2 championed enough that we put in our OKRs.

kevina:
My wording "urgent need" was perhaps a bit strong.  Having this as a _P1_ may be enough reason.

I will be okay with this if we label the extra flag as temporary and they way `--stream` is implemented as experimental so we will be free to change it once the framework (stateful encoders) is in place.

I stated my opinion, I will defer to @Stebalien at this point.

hannahhoward:
@kevina I went ahead and marked the feature as experimental and put a note about the LastObjectHash data field being a temporary tool meant to mark our place in streaming. I also went ahead and filed an issue about having a protocol for passing state when we do streaming operations: https://github.com/ipfs/go-ipfs/issues/5780

I can also add the word Temporary to the actual LastObjectHash field name if you like. 

I'm hoping we can come to an agreement about merging with these conditions so all @Stebalien has to do is sign off on this, cause he's super busy and it's less time consuming just support an agreed upon compromise than serve as tie-breaker.

kevina:
@hannahhoward if you want my approval without @Stebalien please find a way to do this without the extra flag.

I would try and I find a solution which uses _both_`PostRun` and the Encoders (see https://github.com/overbool/go-ipfs/pull/1 for how it might be done).  It may be possible to make it just work.  If not then I will make sure nothing breaks when `--stream` is not used.  When it is used it is I will try and support `curl 'http://localhost:5001/api/v0/ls?arg=/ipns/ipfs.io&encoding=text'` as best as I can.  This may mean that printing headers and footers is not supported when `--stream` is combined with `encoding=text` (for example if using the API directly bypasses PostRun).  Of course headers and footers should be supported when `--stream` is used via the command line and not directly through the API.

Stebalien:
I agree this is not an optimal solution and I really don't see an optimal way forward. However, this really is kind of burning.

Really, this just gives us one more reason to design a new API from scratch along with the JS team. Unfortunately, that got punted till next year.

I'm going to play with this a bit...

hannahhoward:
@Stebalien @kevina 

I pushed one more version, of the sort that @kevina requested. I'm not entirely happy with it but I don't believe there is a way to avoid this implementation (I initially tried using the encoder in the post run even on stream but that presents issues with making the tabs line up)

There's clean up to be done, but this is as absolutely close as I believe you can get without changing the output format.

It does satisfy encoding=text and it streams correctly.

kevina:
@hannahhoward I think so.  I can review this more carefully tomorrow if required.

kevina:
Okay I am compiling this locally and reviewing it now.

core/commands/ls.go
```diff
@@ -60,158 +66,246 @@ The JSON output contains type information.
...
+				lsLink, err := makeLsLink(req, dserv, resolve, link)
+				if err != nil {
+					return err
+				}
+				output := []LsObject{
+					{
+						Hash:  paths[i],
+						Links: []LsLink{*lsLink},
+					},
+				}
```
kevina: I would remove some of the extra newlines to make this a more readable:
```go
output := []LsObject{{
	Hash:  paths[i],
	Links: []LsLink{*lsLink},
}}
```

hannahhoward:
@kevina yea I left Links as is cause I think pooling links and sending them in chunks when streaming is an open question -- which we may want to explore in a future implementation

core/commands/ls.go
```diff
@@ -60,158 +66,246 @@ The JSON output contains type information.
...
+		Type: t,
+	}, nil
+}
 
-			headers, _, _ := res.Request().Option(lsHeadersOptionNameTime).Bool()
-			output, ok := v.(*LsOutput)
-			if !ok {
-				return nil, e.TypeErr(output, v)
-			}
+func tabularOutput(req *cmds.Request, w io.Writer, out *LsOutput, lastObjectHash string, ignoreBreaks bool) string {
```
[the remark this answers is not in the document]
hannahhoward: It's not quite that. In the case of an tabularOutput used in the encoder (which gets called in the case of an HTTP call with encoding=text), it is the same as streaming -- if we are streaming text over HTTP, we need to ignore breaks cause it's impossible to know where to put them in this case. But we ought to add breaks for multiple directories when using encoding=text but not streaming.

However, in the case of tabularOutput called in PostRun, we never want to ignore breaks -- because now we're able to track state so we can insert the breaks correctly, even when streaming.

core/commands/ls.go
```diff
@@ -60,158 +66,246 @@ The JSON output contains type information.
...
-				output[i].Links[j] = LsLink{
-					Name: link.Name,
-					Hash: link.Cid.String(),
-					Size: link.Size,
-					Type: t,
+				out := v.(*LsOutput)
+				if stream {
+					lastObjectHash = tabularOutput(req, os.Stdout, out, lastObjectHash, false)
+				} else {
+					re.Emit(out)
```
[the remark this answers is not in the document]
kevina: Even if we don't remove the ignoreBreaks parameters I still don't think the `if ... else` is needed.  We can always just call `tabularOutput(req, os.Stdout, out, lastObjectHash, false)` as that will be the end result anyway.

core/commands/ls.go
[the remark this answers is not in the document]
kevina: Sorry.  I wasn't thinking this through enough.

hannahhoward:
@Stebalien this incorporates @kevina 's last round of suggestions and I think is ready to merge, pending your approval.

core/commands/ls.go
kevina: I agree, but we can clean this up afterwards.

hannahhoward:
@kevina did cleanup suggested by @magik6k. thank you for the reminder

core/commands/ls.go
```diff
@@ -212,20 +212,14 @@ The JSON output contains type information.
...
-			}
+	links := dagnode.Links()
+	linkResults := make(chan unixfs.LinkResult, len(links))
+	for _, l := range links {
+		linkResults <- unixfs.LinkResult{
+			Link: l,
+			Err:  nil,
 		}
-	}()
+	}
```
kevina: I believe you still need to close the channel and check for `req.Context.Done()`.

hannahhoward: sorry bout that bit. sloppy coding from moving fast.

I also don't know req.Context.Done() -- @magik6k do you know?

hannahhoward: went ahead and added the channel close.

I'm not sure we have to be super concerned about context.Done() because based on the way the channel is setup, the write is not a blocking write -- because the buffer for the channel is defined as being as long as the list of links. My understanding is that would be an issue if the request getting cancelled in a separate process could potentially leave a different go routine still writing cause it's doing a blocking write to a channel that is no longer being read from. In this case, we aren't starting a separate go routine and the channel writes are synchronous.

Then again, this code has no integration tests I think... I wonder if that's something that needs to be added.

Anyway, this function is really an outlier use case -- so if there's any confusion I can just revert to the original version.

kevina: @hannahhoward did you push the change with the `close`.  I am not seeing it in the latest diff.

Let's keep this as unresolved until we either revert the change or figure out if we need to handle `context.Done()`.

hannahhoward: oops sorry failed to push even though I thought I had.

kevina:
@Stebalien there is one minor issue (keeping me from approving) but otherwise this LGTM.

Stebalien:
Thank @hannahhoward and everyone that reviewed this. Finding a solution when there is no good solution is painful and difficult but you all managed to do it.

:tada:

hannahhoward:
Thanks @Stebalien and @kevina and @magik6k for all the feedback!

PR #5614, opened 2018-10-18
update the webui

Stebalien:
Because the current one is *so* broken it's not worth waiting for an actual release.


Stebalien:
So, we can always upgrade again. However, I'd like this in the next release and I'd like to *cut* a release ASAP.

Stebalien:
@olizilla says "that's OK".

Issue #5610, opened 2018-10-18
docs: Which ports serve which purpose?


funkyfuture:
#### Type:

Documentation enhancement

#### Description:

There are at least five ports that `go-ipfs` is listening to (compare `Dockerfile`), but I can't find anything specific about their purposes as a complete overview in the docs. For example the [ipfs cluster docs](https://cluster.ipfs.io/documentation/security/) do that very well.

As I'm intending to setup a public gateway I'd like to be informed about these aspects in order to configure the whole setup. A documentation of this can also help to understand the software's architecture.

magik6k:
TL;DR version before anyone writes this down properly:

* 4001 - default libp2p swarm port - should be open to public for all nodes if possible
* 5001 - API port - provides write/admin access to the node, shouldn't be exposed at all
* 8080 - Gateway + read only API subset - quite safe to expose, but operating public gateway may still be a risk in some ways (there are some proposals on how to make this safer - like https://github.com/ipfs/go-ipfs/issues/5513)

There may be some other ports open by subsystems (like 5353 udp for mdns), but they don't really matter in most configs

funkyfuture:
thanks so far. in the `Dockerfile` these are also exposed:

- `4002/udp`
- `8081`

i think regarding `8080` the wording can include something like

> you may make this endpoint publicly available with a proxy or webserver as `https://ipfs.example.org`

Stebalien:
Those look out of date. They should be:

* 4002/udp is historical (was used for UTP which has been deprecated, for now at least).
* 8081 *can* be used for the websocket transfer. However ,this isn't enabled by default.

But yes, this should all be commented. Mind documenting it in the docker file?

PR #5615, opened 2018-10-18
Dockerfile: Specifies comments on exposed ports

funkyfuture:
this pr addresses #5610. i have no clue what websocket transfer is in this context, that part may be very wrong.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -51,11 +51,13 @@ COPY --from=0 /etc/ssl/certs /etc/ssl/certs
 # This shared lib (part of glibc) doesn't seem to be included with busybox.
 COPY --from=0 /lib/x86_64-linux-gnu/libdl-2.24.so /lib/libdl.so.2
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -51,11 +51,13 @@ COPY --from=0 /etc/ssl/certs /etc/ssl/certs
...
 
-# Ports for Swarm TCP, Swarm uTP, API, Gateway, Swarm Websockets
+# Swarm TCP; should be exposed to the public
 EXPOSE 4001
-EXPOSE 4002/udp
+# Daemon API; must not be exposed publicly but to client services under you control
 EXPOSE 5001
+# Web Gateway; can be exposed publicly with a proxy, e.g. as https://ipfs.example.org
 EXPOSE 8080
+# Swarm Websockets; must be exposed publilcy when websocket transfer is enabled
```
Stebalien:
```suggestion
# Swarm Websockets; must be exposed publicly when the node is listening using the websocket transport (/ipX/.../tcp/8081/ws).
```

Issue #5594, opened 2018-10-12
`ipfs stats bw` doesn't support --enc json


MichaelMure:
#### Version information:
go-ipfs 0.4.17 and git master

#### Type:

Bug

#### Description:

All commands should support the `--enc` options. It doesn't work for `ipfs stats bw` I suppose because the result of the command is formatted in a `PostRun` instead of an `Encoders`.

overbool:
@MichaelMure `ipfs stats bw` has a option `--poll` that supports uninterrupted output. Therefore, I think it's not easy to achieve this feature by using `Encoders ` instead of `PostRun`.

Stebalien:
This *looks* fixable, it's just a bit tricky. The encoder will be called once per item.

kevina:
How important is this.  Maybe this is something we should wait on until we fix the need for using PostRun in most cases otherwise we run into the same problems we had with `ipfs ls --stream` in particular it appears a header is outputted is some cases.

Stebalien:
IMO, low priority. Most clients will just use HTTP directly.

lidel:
Confirmed this is still a problem in 0.11.

Extracting specific value via `ipfs stats bw --enc=json | jq <selector>` would be easier and more future-proof than  parsing text with sed/awk. If it makes fix easier, we don't need to support JSON output in `--poll` mode, fixing the default output would be more than enough.

PR #5620, opened 2018-10-21
fix(commands): support --enc=json in stat

overbool:
Fixes: https://github.com/ipfs/go-ipfs/issues/5594

License: MIT

### DIFF
--- a/core/commands/stat.go
+++ b/core/commands/stat.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"sync"
 	"time"
 
 	cmdenv "github.com/ipfs/go-ipfs/core/commands/cmdenv"
@@ -39,6 +40,9 @@ const (
 	statPollOptionName     = "poll"
 	statIntervalOptionName = "interval"
 )
+
+// for `--poll` flag in `ipfs stat bw`
+var once sync.Once
 
 var statBwCmd = &cmds.Command{
 	Helptext: cmdkit.HelpText{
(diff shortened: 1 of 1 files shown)


overbool:
@magik6k could u help me review it?

core/commands/stat.go
```diff
@@ -147,35 +151,31 @@ Example:
...
-		cmds.CLI: func(res cmds.Response, re cmds.ResponseEmitter) error {
-			polling, _ := res.Request().Options[statPollOptionName].(bool)
+	Encoders: cmds.EncoderMap{
+		cmds.Text: cmds.MakeEncoder(func(req *cmds.Request, w io.Writer, v interface{}) error {
+			polling, _ := req.Options[statPollOptionName].(bool)
 
 			if polling {
-				fmt.Fprintln(os.Stdout, "Total Up    Total Down  Rate Up     Rate Down")
+				// print only once in `--poll` flag
+				once.Do(func() {
```
magik6k: This won't work for multiple invocations of the command, not sure what's the best way to do that. cc @keks

core/commands/stat.go
```diff
@@ -147,35 +151,31 @@ Example:
...
+
+			bs := v.(*metrics.Stats)
+
+			if !polling {
+				printStats(os.Stdout, bs)
+				return nil
 			}
-		},
+
+			fmt.Fprintf(os.Stdout, "%8s    ", humanize.Bytes(uint64(bs.TotalOut)))
```
Kubuxu: We should be also probably be printing into the writer instead of directly to Stdout.

core/commands/stat.go
keks: The proper way to do this is using `PostRun`. Encoders are for processing individual elements, while `PostRun` is passed the stream that is emitted by `Run` and can do transport (e.g. cli or http) specific presentation.

keks: To clarify: I think we need a second `PostRunFunc` in the `PostRun` map with index `cmds.JSON`.

Stebalien:
(closing due to inactivity)

Issue #5275, opened 2018-07-22
Provide a way to truncate a corrupted Badger DB


schomatis:
Badger needs explicit consent from the user to truncate a corrupted DB when starting: passing a `--truncate` flag. That flag is not being provided as an option right now and the user gets errors like:

```
# IPFS_PATH=/ipfs/ipfs_master/repo ./ipfs daemon
Initializing daemon...
Error: Unable to replay value log: "/ipfs/ipfs_master/repo/badgerds/000017.vlog": Data corruption detected. Value log truncate required to run DB. This would result in data loss.
Received interrupt signal, shutting down...
(Hit ctrl-c again to force-shutdown the daemon.)
```

(and https://github.com/ipfs/go-ipfs/issues/5213#issuecomment-406890792), which are not really useful.

Documentation should also be added as it's a normal scenario to have a (relatively small) portion of the DB corrupted after a crash.

Stebalien:
Will this small amount include "synced" data? If so, and if there's no way for us to actually sync the data, we should move to a different datastore. If not, we should just truncate by default.

Explicitly, I'd expect the following semantics:

1. If `Datastore.Put` has returned, the data must be persisted. Passing the truncate option shouldn't remove this data.
2. If `Datastore.Put` has not returned, we don't care. Passing the truncate option may remove this data.

schomatis:
> Will this small amount include "synced" data?

It shouldn't if `syncWrites` is set (which by default it is), as everything will be directly persisted to the value log on disk, and from that point it can be recovered after a crash. However, if that is a deal breaker, let me double check that and get back to you.

Stebalien:
We need *some* commit point. Some point where we can know that the data has been persisted.

manishrjain:
Yes, that understanding is correct. If it was applied successfully to the value log, it would be replayed back after crash.

Stebalien:
So, it looks like, given the semantics we expect, we should always enable truncation if SyncWrites is enabled.

PR: https://github.com/ipfs/go-ds-badger/pull/30

Stebalien:
@schomatis do you have time to fix this as we discussed? That is:

1. Add a `truncate` config option.
2. Turn it *on* by default.

schomatis:
Yes, I'll try to get to it next week.

> Turn it on by default.

Could you point me to the part of the code that generates the `config` file please?

Stebalien:
[...]

PR #5625, opened 2018-10-22
badger: add truncate flag

schomatis:
Depends on https://github.com/ipfs/go-ipfs-config/pull/20 (merged).

Closes: #5275.

### DIFF
--- a/docs/datastores.md
+++ b/docs/datastores.md
@@ -38,11 +38,15 @@ Uses a leveldb database to store key value pairs.
 ## badgerds
 Uses [badger](https://github.com/dgraph-io/badger) as a key value store.
 
+* `syncWrites`: Synchronize every write to disk.
+* `truncate`: Truncate the DB if a corrupted sector is found (otherwise Badger won't start). This option is always set to `true` in Windows if `syncWrites` is set.
+
 ```json
 {
 	"type": "badgerds",
 	"path": "<location of badger inside repo",
-	"syncWrites": true|false
+	"syncWrites": true|false,
+	"truncate": true|false,
 }
 ```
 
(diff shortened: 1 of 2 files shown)


repo/fsrepo/datastores.go
```diff
@@ -382,6 +383,17 @@ func BadgerdsDatastoreConfig(params map[string]interface{}) (DatastoreConfig, er
 		}
 	}
 
+	truncate, ok := params["truncate"]
```
magik6k: How about (could do this with `syncWrites` above too):
```
c.truncate, ok = params["truncate"].(bool)
if !ok {
    c.truncate = true
}
```

schomatis: That is much cleaner (I didn't even knew I could perform those two operations in the same line) but it will silently fail for cases when the user puts the value as a string, e.g., `"syncWrites": "false"` will actually be interpreted as `true` without the user even noticing it.

schomatis:
> We should also note (or change) the behaviour this has with syncWirtes on Windows: https://github.com/ipfs/go-ds-badger/blob/master/datastore.go#L64

Good point, added it to the documentation.

Stebalien:
@schomatis, @magik6k this is now (finally) unblocked.

schomatis:
The entire `badgerdsDatastoreConfig` structure seems to have been moved now, not sure where.

magik6k:
There: https://github.com/ipfs/go-ipfs/blob/master/plugin/plugins/badgerds/badgerds.go#L41

Stebalien:
Moved to: https://github.com/ipfs/go-ipfs/blob/d40c69f56e3f2abfa8a72d3213a5ca1e8af1ab98/plugin/plugins/badgerds/badgerds.go#L41-L47

schomatis:
Thank you both! Rebased.

PR #5626, opened 2018-10-22
Explicitally use Batch after removing it from importers

hsanjuan:
See:

* https://github.com/ipfs/go-unixfs/pull/41
* https://github.com/ipfs/go-ipld-format/pull/46

License: MIT


hsanjuan:
@schomatis done.

core/coreunix/add.go
```diff
@@ -47,13 +47,31 @@ type Object struct {
 	Size  string
 }
 
+// A batch-adding dag service for use with importers (and only for that)
+// Always Commit() after adding.
+type batching struct {
```
hsanjuan: Unless we make the `DagBuilderHelper` take separately an `ipld.NodeAdder` and an `ipld.NodeGetter` (or an adder + a separate `DAGService`) rather than just a `DAGService`, we need to wrap Batch in a `DAGService`, as it seems `trickle` uses `Get`.

Stebalien: If it uses get as I think it does, we probably need to commit before calling get. Let's go with plan b and make Batch implement DAGService (flushing on get). (unless you can think of a better solution)

hsanjuan: @Stebalien yeah i'll do that to be on the safe side.

hsanjuan:
@Stebalien @schomatis this ready using `BufferedDAG` with all bubbled.

hsanjuan:
I'm not sure test failures have anything to do with this

core/coreunix/add.go
```diff
@@ -49,11 +49,14 @@ type Object struct {
 
 // NewAdder Returns a new Adder used for a file add operation.
 func NewAdder(ctx context.Context, p pin.Pinner, bs bstore.GCBlockstore, ds ipld.DAGService) (*Adder, error) {
+	bufferedDS := ipld.NewBufferedDAG(ctx, ds)
```
magik6k: I'll probably refactor this to happen on the coreapi layer (next to offline/sessions stuff), but this is good enough for now

PR #5627, opened 2018-10-22
update webui to 2.1.0

Stebalien:
This brings us on-par with js-ipfs.


Stebalien:
(doesn't need a review, really)

Issue #4942, opened 2018-04-18
Config is not thread-safe


Stebalien:
While we try to make it thread-safe by copying the config on read (see `Repo.Config()`) this doesn't really fix the problem because we have, e.g., maps inside of it. Furthermore, while not safety issue, the profile apply commands don't compose atomically.

Instead of exposing `Config()`, `SetConfig()`, `BackupConfig()`, etc., we should probably provide the following two methods:

```go
ReadConfig(read func (cfg *Config) error) error

// Not sure about the signature but I'd *like* backups to be atomic.
UpdateConfig(update func (cfg *Config) error, backup bool) (backup string, err error)
```

chenminjian:
I am very interested in this. Can I help with this?

Stebalien:
I'd like to get someone (@magik6k? @schomatis?) to comment on the design first.

magik6k:
Looks good to me

Stebalien:
@chenminjian then go ahead.

chenminjian:
OK.

schomatis:
SGTM

kevina:
This SGTM but I would need to see how the implementation works out.

schomatis:
@Stebalien Is this blocked by https://github.com/ipfs/go-ipfs/pull/5634 or can @chenminjian still attempt a fix here?

Stebalien:
Not really. That fixes the thread-safety issue but still doesn't allow for atomic config updates. For that, we need this change (so we can apply transformations while holding a lock on the entire config.

Stebalien:
Really, I'm not sure if there *is* a thread-safety issue anymore (if ever?). However, having a way to clone the config is still useful.

schomatis:
> However, having a way to clone the config is still useful.

Ok, pinging @chenminjian then, as he seemed really interested in this and has already landed some pretty useful patches regarding the `go-ipfs` config system.

chenminjian:
❤️ Thanks for pinging me. I watched go-ipfs, so I didn't notice this email immediately.

chenminjian:
> However, having a way to clone the config is still useful.
This has been implemented in the latest `go-ipfs-config`.

PR #5634, opened 2018-10-23
fix config data race

Stebalien:
This fixes the data-race in the config. Depends on: https://github.com/ipfs/go-ipfs-config/pull/16.

This does not fix https://github.com/ipfs/go-ipfs/issues/4942 as there's still a logical race: parallel config updates clobber each other.

However, we'll still need the Clone function for the new `--dry-run` flag introduced in https://github.com/ipfs/go-ipfs/pull/5455.


Stebalien:
The tests won't pass till the config change has been bubbled.

PR #5643, opened 2018-10-25
refactor(mfs): use mfs type helper

overbool:
Fixes: https://github.com/ipfs/go-mfs/issues/22

License: MIT

### DIFF
--- a/core/commands/files.go
+++ b/core/commands/files.go
@@ -452,8 +452,9 @@ Examples:
 
 		long, _, _ := req.Option(longOptionName).Bool()
 
-		switch fsn := fsn.(type) {
-		case *mfs.Directory:
+		switch fsn.Type() {
+		case mfs.TDir:
+			fsn := fsn.(*mfs.Directory)
 			if !long {
 				var output []mfs.NodeListing
 				names, err := fsn.ListNames(req.Context())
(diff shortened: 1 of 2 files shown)


Stebalien:
go-mfs is up-to-date. It's just that we need the `pdir` variable.

schomatis:
Oh, I missed that, I now see that those were commited directly to `master`, (pending the final tests) RFM then.

Stebalien:
> Oh, I missed that, I now see that those were commited directly

Yeah, I generally just push gx-workspace updates to master. I guess I could make PRs and then merge them locally but when I have to change 50+ repository, I stop caring.

Stebalien:
So, I agree that those helpers are useful but I'm not sure *this* change actually improves the code. Instead of having clearly safe type switches, we're now just asserting the type.

schomatis:
> Instead of having clearly safe type switches, we're now just asserting the type.

Agreed. Those assertions are pretty safe because we are checking the type but still, this is not adding much code clarity, we're still type casting the `FSNode` interface to do pretty much anything. We should first extend that interface (e.g., add access to name, parent and other information common to both file and directory) for these helper functions to be useful.

Closing this for now while I work on the MFS side (thanks anyway @overbool!).

PR #5645, opened 2018-10-25
remove p2p-circuit addr hack in ipfs swarm peers

Stebalien:
We used to have to do this as these addresses had `/ipfs/QmId` appended at the *transport* layer. However, we were able to remove this with the transport refactor so we can now remove this check.

fixes https://github.com/libp2p/specs/issues/72


Stebalien:
(already tested by the circuit sharness tests)

PR #5646, opened 2018-10-25
refactor(commands/id): use new command

overbool:
License: MIT

### DIFF
--- a/core/commands/id.go
+++ b/core/commands/id.go
@@ -1,18 +1,19 @@
 package commands
 
 import (
-	"bytes"
 	"encoding/base64"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"io"
 	"strings"
 
-	cmds "github.com/ipfs/go-ipfs/commands"
 	core "github.com/ipfs/go-ipfs/core"
+	cmdenv "github.com/ipfs/go-ipfs/core/commands/cmdenv"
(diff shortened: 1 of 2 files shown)


overbool:
@Stebalien Why do we keep both old and new commands? is there any problem about replacing all old command with new command?

Stebalien:
We should switch, we just haven't completed the transition.

The *why* is largely because every time I start doing that I run into https://github.com/ipfs/go-ipfs-cmds/issues/65 and get side-tracked, However, I've now finally "fixed" it here: https://github.com/ipfs/go-ipfs-cmds/pull/128 so maybe this'll stop being an issue...

overbool:
@Stebalien Is there any essential difference between `MakeTypedEncoder` and `MakeEncoder`?

IMO, If we use `MakeTypedEncoder`, we just don't need to do type conversions(because it take in the real parameter instead of interface). In addition to this, are there any obvious advantages about using `MakeTypedEncoder`? if not, why don't we just use `MakeEncoder`?

Stebalien:
> Is there any essential difference between

It's just easier to use. Basically, you don't have to handle the type checks yourself. Really, we should *always* be using MakeTypedEncoder as, IMO, it's easier to read.

We have both because `MakeTypedEncoder` came later.

PR #5647, opened 2018-10-25
configurable pubsub signing

Stebalien:
I'd like to sneak this into the release so we can turn on strict verification ASAP.

Depends on https://github.com/ipfs/go-ipfs-config/pull/18 and a gx update.

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -30,12 +30,11 @@ import (
 	pin "github.com/ipfs/go-ipfs/pin"
 	repo "github.com/ipfs/go-ipfs/repo"
 
-	config "gx/ipfs/QmNUhkTWN7iynJZTj1RcTsQDSRGGkh87zMo9ELypxhY8Y6/go-ipfs-config"
 	bitswap "gx/ipfs/QmNkxFCmPtr2RQxjZNRCNryLud4L9wMEiBJsLgF14MqTHj/go-bitswap"
 	bsnet "gx/ipfs/QmNkxFCmPtr2RQxjZNRCNryLud4L9wMEiBJsLgF14MqTHj/go-bitswap/network"
+	config "gx/ipfs/QmPEpj17FDRpc7K1aArKZp3RsHtzRMKykeK9GVgn4WQGPR/go-ipfs-config"
 	cid "gx/ipfs/QmPSQnBKM9g7BaUcZCvswUJVscQ1ipjmwxN5PXCjkp9EQ7/go-cid"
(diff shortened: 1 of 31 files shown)


core/core.go
```diff
@@ -475,14 +475,23 @@ func (n *IpfsNode) startOnlineServicesWithHost(ctx context.Context, host p2phost
 
 		var service *pubsub.PubSub
 
+		var pubsubOptions []pubsub.Option
```
schomatis: What happens if they are both set?

test/sharness/t0180-pubsub.sh
```diff
@@ -10,83 +10,143 @@ test_expect_success 'init iptb' '
...
+  test_expect_success 'pubsub' '
+    echo "testOK" > expected &&
+    touch empty &&
+    mkfifo wait ||
+    test_fsh echo init fail
+  
+    # ipfs pubsub sub is long-running so we need to start it in the background and
+    # wait put its output somewhere where we can access it
+    (
+      ipfsi 0 pubsub sub --enc=ndpayload testTopic | if read line; then
```
schomatis: What is `ipfsi`? (I mean, where can I read some information about it?)

Stebalien: It's a function defined in `test/sharness/lib/iptb-lib.sh`.

core/core.go
Stebalien: The user won't be able to send any messages. We should error in that case, shouldn't we...

Stebalien: https://github.com/libp2p/go-libp2p-pubsub/pull/122

schomatis: Re-reading this now with some more context it seems like a valid case (although I'm not sure how useful it is), that is, we are requiring signed messages but we don't sign them ourselves.

Stebalien: It is a valid case, technically. But it's so useless that we should probably get rid of it.

Stebalien:
> it's a string, but does it carry some special meaning?

It's an arbitrary (binary) string. At one point, it was the hash of a topic descriptor but that never really got finished.

Stebalien:
@schomatis mind ticking https://github.com/ipfs/go-ipfs-config/pull/18 as well?

core/core.go
Stebalien: (actually, I may just do that in a followup patch once someone takes a look at that PR)

vyzo: if strict signing is set but signing is disabled, then it will be ignored.

Stebalien: Got it. Now that the option validation patch has been merged, it should return an error on construction. That should be the least surprising result.

PR #5649, opened 2018-10-26
Local Gateway option

magik6k:
This is an alternative / more contained version of https://github.com/ipfs/go-ipfs/pull/4150

((todo) some gx stuff still needs fixing)


core/coreapi/coreapi.go
```diff
@@ -92,3 +95,10 @@ func (api *CoreAPI) getSession(ctx context.Context) *CoreAPI {
 	ng := dag.NewReadOnlyDagService(dag.NewSession(ctx, api.dag))
 	return &CoreAPI{api.node, ng}
 }
+
+// Offline returns new api with offline DAG
+func (api *CoreAPI) Offline() *CoreAPI {
+	bs := api.node.Blocks.Blockstore()
+	ng := dag.NewDAGService(bserv.New(bs, offline.Exchange(bs)))
```
Stebalien: I'm really worried we're going to bypass this dag accidentally by calling something on the node that has the wrong dag. Unfortunately, I'm not sure if there's a better solution (that works with sessions, anyways, offline can be done with a context hint).

core/corehttp/gateway.go
```diff
@@ -25,11 +25,16 @@ func GatewayOption(writable bool, paths ...string) ServeOption {
 			return nil, err
 		}
 
+		api := coreapi.NewCoreAPI(n)
+		if cfg.Gateway.NoFetch {
+			api = api.(*coreapi.CoreAPI).Offline()
```
Stebalien: What if we just made this a part of the API? We'll want a `ipfs --offline cat /...` eventually.

core/coreapi/coreapi.go
Stebalien: But I'm probably just paranoid.

core/corehttp/gateway.go
magik6k: +1, but this will be a slightly bigger change than I'm comfortable sneaking into this PR, opened an issue - https://github.com/ipfs/go-ipfs/issues/5654

core/coreapi/coreapi.go
magik6k: Yeah, I'll probably go with context hints (#5654), but that's a larger change

Voker57:
This does not include a lot of code from my PR that tries to make the API actually offline beyond stuff that readonly gateway uses, but it could work I guess if nobody tries to use it for something else, mine does not cover all cases either.

Also, I don't like "negating" config option, it's a bit confusing.

magik6k:
> Also, I don't like "negating" config option, it's a bit confusing.

When adding options to the config, old nodes will use zero value for added fields, which in this case would break existing setups.

Stebalien:
> When adding options to the config, old nodes will use zero value for added fields, which in this case would break existing setups.

We *really* need a comprehensive config system with layers. One has to exist...

magik6k:
Also, convincing the read only HTTP API to not fetch blocks is going to be harder, we should probably just implement the context hints.

magik6k:
I don't see a nice way to work around this without context hints, so we can probably call this blocked for now. I'll try to put context hints near the top on my todo list though.

Voker57:
@magik6k 
> When adding options to the config, old nodes will use zero value for added fields, which in this case would break existing setups.
Isn't that handled by go-ipfs-config? https://github.com/Voker57/go-ipfs-config/commit/9c41d632310ac36f7e0ec6400817d640ade2c8b1

magik6k:
> Isn't that handled by go-ipfs-config? Voker57/go-ipfs-config@9c41d63

Nope, code in `config/init.go` only gets called on init. If values are missing from existing node config, they are set to zero value (which is `false` in this case). Using `FetchBlocks` would basically disable fetching blocks on gateway for all existing nodes (and other programs that create config.Config manually, expecting that (almost) zero-value config will just work).

core/corehttp/gateway_handler.go
```diff
@@ -297,9 +280,11 @@ func (i *gatewayHandler) getOrHeadHandler(ctx context.Context, w http.ResponseWr
 		http.ServeContent(w, r, "index.html", modtime, f)
 		return
 	default:
+		if _, ok := err.(resolver.ErrNoLink); ok {
```
Stebalien: Nit: Would probably be more readable as:

```go
switch err.(type) {
case nil:
...
case resolver.ErrNoLink:
...
default:
}
```

core/corehttp/gateway_handler.go
```diff
@@ -308,12 +293,22 @@ func (i *gatewayHandler) getOrHeadHandler(ctx context.Context, w http.ResponseWr
 
 	// storage for directory listing
 	var dirListing []directoryItem
-	dirr.ForEachLink(ctx, func(link *ipld.Link) error {
+	dirit := dir.Entries()
+	for dirit.Next() {
 		// See comment above where originalUrlPath is declared.
-		di := directoryItem{humanize.Bytes(link.Size), link.Name, gopath.Join(originalUrlPath, link.Name)}
+		s, err := dirit.Node().Size()
+		if err != nil {
```
Stebalien: We can *probably* just ignore this error.

magik6k:
It's currently `Gateway.NoFetch` in config, which shouldn't be too confusing imo. (the coreapi option name is different, but I guess it can be because it's 'lower level')

core/corehttp/gateway_handler.go
Stebalien: IIRC, filesize is technically optional (but I guess we don't have a spec yet).

magik6k: Link size is optional, I'm not sure about this one

Stebalien:
> It's currently Gateway.NoFetch in config, which shouldn't be too confusing imo. (the coreapi option name is different, but I guess it can be because it's 'lower level')

Ah. Fair enough.

core/corehttp/gateway_handler.go
magik6k: Digging deeper in will only fail if protobuf encoding fails - https://github.com/ipfs/go-merkledag/blob/master/node.go#L240 - at which point we probably want to give user the error

Stebalien: "well documented": https://github.com/ipfs/go-ipfs/issues/4286#issuecomment-336195370

magik6k: This won't error on nodes which don't have size and encode successfully (note that it's not the unixfs filesize (which we don't set for directories, I think) - only the merkledag node size - https://github.com/ipfs/go-ipfs/pull/5649/files#diff-48d085152bf55180b23ed7d5bd1b9b0dR126)

magik6k: Is there anything else blocking this?

magik6k:
Should be ready

(Depends on https://github.com/ipfs/iptb-plugins/pull/11, https://github.com/ipfs/go-ipfs-config/pull/19)

PR #5650, opened 2018-10-26
refactor(cmds): use new cmds lib in version, tar and dns

overbool:
License: MIT

### DIFF
--- a/core/commands/dns.go
+++ b/core/commands/dns.go
@@ -1,16 +1,15 @@
 package commands
 
 import (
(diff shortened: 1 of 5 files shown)


schomatis:
@overbool Could you point me to the issue with the original discussion that led to this patch?

overbool:
> @overbool Could you point me to the issue with the original discussion that led to this patch?

This is the original discussion. https://github.com/ipfs/go-ipfs/pull/5646

PR #5656, opened 2018-10-26
gx: update pubsub (again)

Stebalien:
Small patch to mark gossipsub peers as wanted so we don't prune them.


Stebalien:
(gx-only update, merging)

PR #5660, opened 2018-10-26
namesys: properly attach path in name.Resolve

magik6k:
This got broken in the async namesys refactor, somehow there were no tests to catch this.

### DIFF
--- a/core/coreapi/name_test.go
+++ b/core/coreapi/name_test.go
@@ -2,9 +2,11 @@ package coreapi_test
 
 import (
 	"context"
+	"github.com/ipfs/go-ipfs/core"
 	"io"
 	"io/ioutil"
 	"math/rand"
+	"path"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


core/coreapi/name_test.go
```diff
@@ -21,45 +22,150 @@ func addTestObject(ctx context.Context, api coreiface.CoreAPI) (coreiface.Path,
 	return api.Unixfs().Add(ctx, files.NewReaderFile("", "", ioutil.NopCloser(&io.LimitedReader{R: rnd, N: 4092}), nil))
 }
 
-func TestBasicPublishResolve(t *testing.T) {
-	ctx := context.Background()
-	nds, apis, err := makeAPISwarm(ctx, true, 5)
+func appendPath(path coreiface.Path, sub string) coreiface.Path {
+	p, err := coreiface.ParsePath(path.String() + sub)
```
djdv: Even though it will result in the same behaviour, I wonder if we should make a point to use `path.Join` in these instances.

PR #5663, opened 2018-10-26
commands/ls: use new cmds lib

overbool:
License: MIT

### DIFF
--- a/core/commands/ls.go
+++ b/core/commands/ls.go
@@ -6,8 +6,7 @@ import (
 	"io"
 	"text/tabwriter"
 
-	cmds "github.com/ipfs/go-ipfs/commands"
-	e "github.com/ipfs/go-ipfs/core/commands/e"
+	cmdenv "github.com/ipfs/go-ipfs/core/commands/cmdenv"
 	iface "github.com/ipfs/go-ipfs/core/coreapi/interface"
 
 	cid "gx/ipfs/QmPSQnBKM9g7BaUcZCvswUJVscQ1ipjmwxN5PXCjkp9EQ7/go-cid"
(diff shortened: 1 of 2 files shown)


core/commands/ls.go
```diff
@@ -93,43 +81,39 @@ The JSON output contains type information.
 			dserv = merkledag.NewDAGService(bserv)
 		}
 
-		paths := req.Arguments()
+		paths := req.Arguments
```
Stebalien: There really isn't a need to set this anymore, we can just use it directly in the loop below.

core/commands/ls.go
```diff
@@ -173,45 +155,37 @@ The JSON output contains type information.
...
+	Encoders: cmds.EncoderMap{
+		cmds.Text: cmds.MakeTypedEncoder(func(req *cmds.Request, w io.Writer, out *LsOutput) error {
+			headers, _ := req.Options[lsHeadersOptionNameTime].(bool)
 
 			buf := new(bytes.Buffer)
-			w := tabwriter.NewWriter(buf, 1, 2, 1, ' ', 0)
-			for _, object := range output.Objects {
-				if len(output.Objects) > 1 {
-					fmt.Fprintf(w, "%s:\n", object.Hash)
+			tw := tabwriter.NewWriter(buf, 1, 2, 1, ' ', 0)
```
Stebalien: Let's pass `w` into `tabwriter.NewWriter` directly (and get rid of `buf` entirely). We used to put this all in a buffer because we had to return it from the command but that's no longer necessary.

hannahhoward:
@overbool and #5611 is dependent on https://github.com/ipfs/go-unixfs/pull/39 -- which is almost merged, and then I'll move my focus to #5611

kevina:
@Stebalien I would rather we not delay this as this the switch to the new commands lib is blocking any work on https://github.com/ipfs/go-ipfs/pull/5464 and #5611 may take a while to get in.

Issue #5664, opened 2018-10-26
Commands Library Upgrade


Stebalien:
Given the amount of interest in updating commands-lib commands to the latest commands-lib version (and the recently fixed bugs in MakeTypedEncoder), let's have a meta-issue.

General suggestions:

1. Prefer `MakeTypedEncoder`, it removes some boilerplate.
2. Old encoders tended to return `NewReader(someByteBufferOrString)`. In new code, prefer writing directly to the provided writer instead of first building up a temporary buffer/string.

Please *comment* below with the commands you're planning on fixing (just that).
They'll be checked off when actually fixed.

* [x] bitswap ledger #5676
* [x] bitswap reprovide #5676
* [x] bitswap wantlist #5676
* [x] bootstrap add #5678
* [x] bootstrap add default #5678
* [x] bootstrap list #5678
* [x] bootstrap rm #5678
* [x] bootstrap rm all #5678
* [x] config edit - #5659
* [x] config profile - #5659
* [x] config profile apply - #5659
* [x] config replace - #5659
* [x] config show - #5659
* [x] dag get - #5662
* [x] dag put - #5662
* [x] dag resolve - #5662
* [x] dht findpeer #5671
* [x] dht findprovs #5671
* [x] dht get #5671
* [x] dht provide #5671
* [x] dht put #5671
* [x] dht query #5671
* [x] diag cmds - #5659
* [x] diag cmds clear - #5659
* [x] diag cmds set-time - #5659
* [x] diag sys - #5659
* [x] dns - #5650
* [x] file ls #5756
* [x] files chcid #5665
* [x] files cp #5665
* [x] files flush #5665
* [x] files ls #5665
* [x] files mkdir #5665
* [x] files mv #5665
* [x] files read #5665
* [x] files rm #5665
* [x] filestore dups #5673
* [x] filestore verify #5673
* [x] log level - #5659
* [x] log ls - #5659
* [x] log tail - #5659
* [x] ls - #5611 #5663
* [x] mount - #5659
* [x] object data #5666
* [x] object diff #5666
* [x] object get #5666
* [x] object links #5666
* [x] object new #5666
* [x] object patch add-link #5666
* [x] object patch rm-link #5666
* [x] object patch set-data #5666
* [x] object put #5666
* [x] object stat #5666
* [x] p2p close #5672 
* [x] p2p forward #5672
* [x] p2p listen #5672
* [x] p2p ls #5672
* [x] p2p stream close #5672
[...]

PR #5665, opened 2018-10-26
commands/files: use new cmds

overbool:

License: MIT

### DIFF
--- a/core/commands/files.go
+++ b/core/commands/files.go
@@ -1,7 +1,6 @@
 package commands
 
 import (
-	"bytes"
 	"context"
 	"errors"
 	"fmt"
@@ -11,8 +10,6 @@ import (
 	"sort"
 	"strings"
 
-	oldcmds "github.com/ipfs/go-ipfs/commands"
-	lgc "github.com/ipfs/go-ipfs/commands/legacy"
 	core "github.com/ipfs/go-ipfs/core"
 	cmdenv "github.com/ipfs/go-ipfs/core/commands/cmdenv"
 	e "github.com/ipfs/go-ipfs/core/commands/e"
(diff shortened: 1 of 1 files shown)


core/commands/files.go
```diff
@@ -423,60 +411,53 @@ Examples:
...
 				}
-				res.SetOutput(&filesLsOutput{output})
-			} else {
-				listing, err := fsn.List(req.Context())
-				if err != nil {
-					res.SetError(err, cmdkit.ErrNormal)
-					return
-				}
-				res.SetOutput(&filesLsOutput{listing})
+				return res.Emit(&filesLsOutput{output})
```
magik6k: Should use `cmds.EmitOnce` (because http headers)

core/commands/files.go
```diff
@@ -423,60 +411,53 @@ Examples:
...
+			}
+			listing, err := fsn.List(req.Context)
+			if err != nil {
+				return err
 			}
-			return
+			return res.Emit(&filesLsOutput{listing})
```
magik6k: `cmds.EmitOnce`

core/commands/files.go
```diff
@@ -485,58 +466,44 @@ Examples:
...
 				if err != nil {
-					res.SetError(err, cmdkit.ErrNormal)
-					return
+					return err
 				}
 				out.Entries[0].Hash = nd.Cid().String()
 			}
-			res.SetOutput(out)
-			return
+			return res.Emit(out)
```
magik6k: `cmds.EmitOnce`

core/commands/files.go
overbool: > // EmitOnce is a helper that emits a value wrapped in Single, to signal that this will be the only value sent.

This is the `EmitOnce` comment.

@magik6k  I'm not familiar with `go-ipfs-cmds`, so I want to check with you. Should we replace all `res.Emit()` with `cmds.EmitOnce`?

magik6k: We want to use it only in places which return a single value. For channels and readers we want to keep using `res.Emit`. The header logic which is affected by this is in https://github.com/ipfs/go-ipfs-cmds/blob/master/http/responseemitter.go#L266

overbool: @magik6k Got it, thx.

overbool:
@magik6k Fixed it.
@Stebalien All checks have passed, could you help me review it.

core/commands/files.go
```diff
@@ -1063,66 +971,48 @@ Remove files or directories.
...
-		var success bool
-		defer func() {
-			if success {
-				err := pdir.Flush()
-				if err != nil {
-					res.SetError(err, cmdkit.ErrNormal)
-					return
-				}
-			}
-		}()
```
kevina: :+1: for the removal of the defer, it isn't really needed and the logic is simpler now.

PR #5666, opened 2018-10-26
commands/object: use new cmds

overbool:
License: MIT

### DIFF
--- a/core/commands/object/diff.go
+++ b/core/commands/object/diff.go
@@ -8,7 +8,7 @@ import (
 	coreiface "github.com/ipfs/go-ipfs/core/coreapi/interface"
 	"github.com/ipfs/go-ipfs/dagutils"
 
-	cmds "gx/ipfs/QmdTmGruUz23vgzym3uWpnAEQdGdGifQqBvP8UXSRjG8gZ/go-ipfs-cmds"
+	cmds "gx/ipfs/QmSXUokcP4TJpFfqozT69AVAYRtzXVMUjzQVkYX41R9Svs/go-ipfs-cmds"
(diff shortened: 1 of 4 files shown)


overbool:
@Stebalien CI checks have passed, could u help me review it?

core/commands/object/diff.go
```diff
@@ -1,18 +1,21 @@
...
+	cmdenv "github.com/ipfs/go-ipfs/core/commands/cmdenv"
 	coreiface "github.com/ipfs/go-ipfs/core/coreapi/interface"
 	"github.com/ipfs/go-ipfs/dagutils"
 
+	cmds "gx/ipfs/QmSXUokcP4TJpFfqozT69AVAYRtzXVMUjzQVkYX41R9Svs/go-ipfs-cmds"
 	cmdkit "gx/ipfs/Qmde5VP1qUkyQXKCfmEUA7bP64V2HAptbJ7phuPp7jXWwg/go-ipfs-cmdkit"
 )
 
+const (
+	verboseOptionName = "verbose"
```
kevina: Is there a reason for this?

Nothing wrong with, but it's not very consistent as we don't use contacts for any of the other options in this file.

Personally, I would err on the side of not using the constant, but others may disagree.

overbool: @kevina `version` is used in two places, so I use the constant.

kevina: So is almost every other option, or am I missing something.

In any case it not that big of a deal.

Stebalien: Most of the commands have actually been changed to use constants for these.

PR #5667, opened 2018-10-26
simplify dag put and correctly take pin lock

Stebalien:


Stebalien:
(release isn't cut yet!)

PR #5671, opened 2018-10-27
commands/dht: use new cmds lib

overbool:

License: MIT

### DIFF
--- a/core/commands/dht.go
+++ b/core/commands/dht.go
@@ -8,18 +8,18 @@ import (
 	"time"
 
 	cmdenv "github.com/ipfs/go-ipfs/core/commands/cmdenv"
-	dag "gx/ipfs/QmSei8kFMfqdJq7Q68d2LMnHbTWKKg2daA29ezUYFAUNgc/go-merkledag"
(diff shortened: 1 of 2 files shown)


core/commands/dht.go
```diff
@@ -278,71 +229,51 @@ var provideRefDhtCmd = &cmds.Command{
 		cmdkit.BoolOption("verbose", dhtVerboseOptionName, "Print extra information."),
 		cmdkit.BoolOption(recursiveOptionName, "r", "Recursively provide entire graph."),
 	},
-	Run: func(req cmds.Request, res cmds.Response) {
-		n, err := req.InvocContext().GetNode()
+	Run: func(req *cmds.Request, res cmds.ResponseEmitter, env cmds.Environment) error {
+		nd, err := cmdenv.GetNode(env)
```
magik6k: Why the rename to `nd`?

overbool: @magik6k I just wanna write it in the same way in all `core/commands/*.go`

magik6k: Yeah, makes sense. (I thought that `n` was more common/conststently used, but checking now it's not and `nd` is used about as often, I have no preference, so this LGTM)

PR #5673, opened 2018-10-27
commands/filestore: use new cmds lib

overbool:

License: MIT

### DIFF
--- a/core/commands/filestore.go
+++ b/core/commands/filestore.go
@@ -126,15 +126,15 @@ For ERROR entries the error will also be printed to stderr.
 				return filestore.Verify(fs, c)
 			})
 			return res.Emit(out)
-		} else {
-			fileOrder, _ := req.Options[fileOrderOptionName].(bool)
-			next, err := filestore.VerifyAll(fs, fileOrder)
-			if err != nil {
-				return err
-			}
-			out := listResToChan(req.Context, next)
-			return res.Emit(out)
 		}
+
+		fileOrder, _ := req.Options[fileOrderOptionName].(bool)
+		next, err := filestore.VerifyAll(fs, fileOrder)
+		if err != nil {
+			return err
+		}
+		out := listResToChan(req.Context, next)
+		return res.Emit(out)
 	},
 	Encoders: cmds.EncoderMap{
 		cmds.Text: cmds.MakeTypedEncoder(func(req *cmds.Request, w io.Writer, out *filestore.ListRes) error {
--- a/core/commands/filestore.go
+++ b/core/commands/filestore.go
@@ -3,6 +3,7 @@ package commands
 import (
 	"fmt"
 	"io"
+	"os"
 
 	core "github.com/ipfs/go-ipfs/core"
 	cmdenv "github.com/ipfs/go-ipfs/core/commands/cmdenv"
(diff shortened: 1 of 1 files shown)


kevina:
There seams to be a lot of changes here.  Please give me a few days to look this over.

core/commands/filestore.go
```diff
@@ -118,96 +119,89 @@ For ERROR entries the error will also be printed to stderr.
...
+		}
 
-			if r.Status == filestore.StatusOtherError {
-				fmt.Fprintf(res.Stderr(), "%s\n", r.ErrorMsg)
+		return nil
+	},
+	Encoders: cmds.EncoderMap{
+		cmds.Text: cmds.MakeTypedEncoder(func(req *cmds.Request, w io.Writer, out *filestore.ListRes) error {
+			if out.Status == filestore.StatusOtherError {
+				return fmt.Errorf(out.ErrorMsg)
```
kevina: This is a change in behavior.  `filestore.StatusOtherError` is not an error condensation, but rather indicates something is wrong with that entry.  The error message should be printed and execution should continue as normal.  This is documented in the usage:
```
  The output is:
   ...
  ERROR:    internal error, most likely due to a corrupt database
  
  For ERROR entries the error will also be printed to stderr.
```

core/commands/filestore.go
```diff
@@ -118,96 +119,90 @@ For ERROR entries the error will also be printed to stderr.
...
 
-			if r.Status == filestore.StatusOtherError {
-				fmt.Fprintf(res.Stderr(), "%s\n", r.ErrorMsg)
+		return nil
+	},
+	Encoders: cmds.EncoderMap{
+		cmds.Text: cmds.MakeTypedEncoder(func(req *cmds.Request, w io.Writer, out *filestore.ListRes) error {
+			if out.Status == filestore.StatusOtherError {
+				fmt.Fprintf(w, "%s %s\n", out.Status.Format(), out.ErrorMsg)
+				return nil
```
kevina: This is still not right.

The output on `filestore.StatusOtherError` should be _both_ the error to Stderr and the normal status line as it is done in the original code.

kevina: Yes and then continue to print the status line as normal like it is in the original code.  Although I am not sure if it's okay to directly print to Stderr in an Encoder, this command may require a Postrun.  However, if you find other examples of an encoder writting to os.Stderr then it is okay to also do it here.

core/commands/filestore.go
```diff
@@ -150,7 +151,7 @@ For ERROR entries the error will also be printed to stderr.
 	Encoders: cmds.EncoderMap{
 		cmds.Text: cmds.MakeTypedEncoder(func(req *cmds.Request, w io.Writer, out *filestore.ListRes) error {
 			if out.Status == filestore.StatusOtherError {
-				return fmt.Errorf(out.ErrorMsg)
+				fmt.Fprintf(os.Stderr, "%s\n", out.ErrorMsg)
```
kevina: This _might_ be okay.  Although i think we need to use PostRun here.  I created an issue for it n `go-ipfs-cmds`: https://github.com/ipfs/go-ipfs-cmds/issues/133.

Blocking for now to until others weigh in.

kevina: @Stebalien confirmed in ipfs/go-ipfs-cmds#133 that we should be using PostRun here.

overbool:
> @overbool do you want me to take over this commit and fix it to use PostRun?

Ok, thx.

core/commands/filestore.go
```diff
@@ -118,96 +120,100 @@ For ERROR entries the error will also be printed to stderr.
...
+						return nil
+					}
+					return err
+				}
 
-			r, ok := v.(*filestore.ListRes)
-			if !ok {
-				return nil, e.TypeErr(r, v)
-			}
+				list := v.(*filestore.ListRes)
```
kevina: This should be:
```go
list, ok := v.(*filestore.ListRes)
if !ok {
  return nil, e.TypeErr(r, v)
}
```

overbool: @kevina I had changed it, thx

PR #5674, opened 2018-10-27
commands/pin: use new cmds lib

overbool:

License: MIT

### DIFF
--- a/core/commands/pin.go
+++ b/core/commands/pin.go
@@ -13,11 +13,11 @@ import (
 	corerepo "github.com/ipfs/go-ipfs/core/corerepo"
 	pin "github.com/ipfs/go-ipfs/pin"
 
-	cmds "gx/ipfs/QmSXUokcP4TJpFfqozT69AVAYRtzXVMUjzQVkYX41R9Svs/go-ipfs-cmds"
 	offline "gx/ipfs/QmPpnbwgAuvhUkA9jGooR88ZwZtTUHXXvoQNKdjZC6nYku/go-ipfs-exchange-offline"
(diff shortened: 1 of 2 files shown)


core/commands/pin.go
```diff
@@ -65,118 +64,109 @@ var addPinCmd = &cmds.Command{
...
-				if res.Error() != nil {
-					return nil, res.Error()
-				}
-			default:
-				return nil, e.TypeErr(out, v)
+			if out.Pins != nil {
+				added = out.Pins
+			} else {
+				// this can only happen if the progress option is set
+				return fmt.Errorf("Fetched/Processed %d nodes\r", out.Progress)
```
kevina: This is wrong.  This is not an error.

This command should likely use a PostRun.

overbool:
@kevina I found that req.Arguments only contain the first arg if I pass two arguments from Stdin

kevina:
> @kevina I found that `req.Arguments` only contain the first arg if I pass two arguments from Stdin.

Correct.  To fix this you need to call `req.ParseBodyArgs()` before using `req.Arguments`.

kevina:
@overbool this needs a rebase now.

I would like to get this is ASAP.  Let me know if you need any help with this.

overbool:
@kevina sorry for my delay. I had replaced `Encoder` with `PostRun` and fix some bugs. But, This `pr` seems to have a strange problem that it can't sync my latest commit.

overbool:
@kevina sorry for my delay. I had replaced the `Encoder` with `PostRun` and fixed some bugs. Could u help me review it again?

Stebalien:
So, technically, this breaks `curl http://localhost:5001/api/v0/pin/add/...?enc=text`. I believe the correct way to do this is: https://github.com/overbool/go-ipfs/pull/1. That is, handle status in a post-run but forward everything else to the encoder.

PR #5678, opened 2018-10-27
commands/boostrap: use new cmds

overbool:

License: MIT

### DIFF
--- a/core/commands/bootstrap.go
+++ b/core/commands/bootstrap.go
@@ -5,13 +5,13 @@ import (
 	"io"
 	"sort"
 
-	oldcmds "github.com/ipfs/go-ipfs/commands"
+	cmdenv "github.com/ipfs/go-ipfs/core/commands/cmdenv"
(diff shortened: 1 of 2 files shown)


core/commands/bootstrap.go
```diff
@@ -63,78 +62,54 @@ in the bootstrap list).
...
 		}
 
 		if len(inputPeers) == 0 {
-			res.SetError(errors.New("no bootstrap peers to add"), cmdkit.ErrClient)
-			return
+			return errors.New("no bootstrap peers to add")
 		}
 
-		r, err := fsrepo.Open(req.InvocContext().ConfigRoot)
+		ctx := env.(*oldcmds.Context)
```
kevina: Not a requirement, but we should likely create another helper function in `core/commands/cmdenv` and call it `GetConfigRoot`.  This way it makes it clear that the cast is the right thing to do.  It would of saved me 10-15 minutes of review time.

PR #5679, opened 2018-10-27
commands/refs: use new cmds

overbool:

License: MIT

### DIFF
--- a/core/commands/refs.go
+++ b/core/commands/refs.go
@@ -102,32 +102,24 @@ NOTE: List all references recursively by using the flag '-r'.
 			return err
 		}
 
-		out := make(chan interface{})
+		rw := RefWriter{
+			res:      res,
+			DAG:      n.DAG,
+			Ctx:      ctx,
+			Unique:   unique,
+			PrintFmt: format,
+			MaxDepth: maxDepth,
+		}
 
-		go func() {
-			defer close(out)
-
-			rw := RefWriter{
-				out:      out,
-				DAG:      n.DAG,
-				Ctx:      ctx,
-				Unique:   unique,
-				PrintFmt: format,
-				MaxDepth: maxDepth,
-			}
-
-			for _, o := range objs {
-				if _, err := rw.WriteRefs(o); err != nil {
-					select {
-					case out <- &RefWrapper{Err: err.Error()}:
-					case <-ctx.Done():
-					}
-					return
+		for _, o := range objs {
+			if _, err := rw.WriteRefs(o); err != nil {
+				if err := res.Emit(&RefWrapper{Err: err.Error()}); err != nil {
+					return err
 				}
 			}
-		}()
+		}
 
-		return res.Emit(out)
+		return nil
 	},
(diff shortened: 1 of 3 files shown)


core/commands/refs.go
```diff
@@ -169,42 +155,50 @@ Displays the hashes of all local objects.
...
 
 		out := make(chan interface{})
-		res.SetOutput((<-chan interface{})(out))
 
 		go func() {
 			defer close(out)
 
 			for k := range allKeys {
 				select {
 				case out <- &RefWrapper{Ref: k.String()}:
```
magik6k: We could drop this goroutine / channel and just call `res.Emit` directly for each key

overbool: @magik6k Does this mean that similar cases can be refactored in this way, like(the follow code or other commands code)

https://github.com/ipfs/go-ipfs/blob/0df07e3811cbb5294dba4ccd7110002aacb45c9d/core/commands/refs.go#L110-L135

Stebalien: Yes.

core/commands/refs.go
```diff
@@ -74,65 +77,37 @@ NOTE: List all references recursively by using the flag '-r'.
...
 
-		objs, err := objectsForPaths(ctx, n, req.Arguments())
+		objs, err := objectsForPaths(ctx, n, req.Arguments)
 		if err != nil {
-			res.SetError(err, cmdkit.ErrNormal)
-			return
+			return err
 		}
 
 		out := make(chan interface{})
```
kevina: We should likely drop this channel and go routine and use res.Emit directly for each value?

kevina:
@overbool I would also like to get this p.r. in ASAP, let me know if you need any help with this.

overbool:
@kevina I had fixed them, could u help me review again?

core/commands/refs.go
```diff
@@ -74,91 +67,62 @@ NOTE: List all references recursively by using the flag '-r'.
 		cmdkit.BoolOption(refsRecursiveOptionName, "r", "Recursively list links of child nodes."),
 		cmdkit.IntOption(refsMaxDepthOptionName, "Only for recursive refs, limits fetch and listing to the given depth").WithDefault(-1),
 	},
-	Run: func(req cmds.Request, res cmds.Response) {
-		ctx := req.Context()
-		n, err := req.InvocContext().GetNode()
+	Run: func(req *cmds.Request, res cmds.ResponseEmitter, env cmds.Environment) error {
+		err := req.ParseBodyArgs()
```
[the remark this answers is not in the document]
Stebalien: Meh? Honestly, if we're going to do this up-front anyways, we might as well finish parsing the request as early as possible.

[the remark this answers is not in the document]
Stebalien: We can change this later if you'd like.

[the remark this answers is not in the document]
kevina: I don't really have a strong preference but we should at least try and be consistent about when we do this.

In any case fixing it now is a low priority.

PR #5680, opened 2018-10-27
cleanup the ping command

Stebalien:
followup to the ping -> cmds 2.0 change

1. Use emit directly.
2. Abort immediately when the context is canceled (no need to wait a second for the sleep to elapse).
3. Ping once a second instead of once a second +TTL (to be more like ping)
4. Simplify ParsePeerParam


core/commands/ping.go
```diff
@@ -59,166 +59,114 @@ trip latency information.
...
 
-			ctx, cancel := context.WithTimeout(ctx, kPingTimeout)
-			defer cancel()
+			ctx, cancel := context.WithTimeout(req.Context, kPingTimeout)
 			p, err := n.Routing.FindPeer(ctx, pid)
+			cancel()
 			if err != nil {
-				select {
-				case outChan <- &PingResult{Text: fmt.Sprintf("Peer lookup error: %s", err)}:
-				case <-ctx.Done():
```
Stebalien: This should have used the *command* context, not the find-peer context. This context is pretty much guaranteed to be expired at this point (so will return "success" to the user when we fail to lookup the peer). This is causing the JS interop tests to fail (for good reason).

PR #5687, opened 2018-10-29
fix(cmd/gc): Run func does not return error when Emit func returns error

chenminjian:
License: MIT


Stebalien:
(we're going to have to cut an RC2 anyways so we might as well merge this (it is a regression)).

Issue #5657, opened 2018-10-26
Extensive use of PubSub may cause "out of memory" error


romikser:
#### Version information:
I am using go-ipfs as library for my project. Commit I'm using is b5bba2ce80f8f6cd7b268ade72ad670e1102ab54.

#### Type:
Bug

#### Description:
Our application extensively uses pubsub. Recently we've updated IPFS because of gossipsub and started using gossipsub in our project. After some time we've noticed that application crashes with goroutines count ~800000 and "out of memory" error. 

The crash happened when node tried to resolve IPNS during heavy pubsub messages exchange. We've noticed it and disabled ipnsps, which did help.

After that I started investigating the problem. For test purposes I've set up 3 nodes. Two were listening two different pubsub topics, and the third one extensively published to one of topics (every 10 milliseconds I picked topic randomly). After 10 seconds from node publishing to topics I tried to subscribe to another topic and crashed after some time. Judging by logs, the problem is announceRetry method getting called infinite number of times.

I didn't manage to reproduce this bug using floodsub, so maybe the problem is with gossipsub only. Also I didn't manage to crash when I published to one topic instead of two.

P.S. In logs "WE SHOULD CRASH" message is where I started subscribing to topic on sending node.

[Log.log](https://github.com/ipfs/go-ipfs/files/2519611/Log.log)

vyzo:
Uh oh, do you mind opening an issue in https://github.com/libp2p/go-libp2p-pubsub?

Stebalien:
Can you post the output of `curl 'http://localhost:5001/debug/pprof/goroutine?debug=2'` after memory starts building up (you shouldn't have to wait that long).

vyzo:
Are you using validators by any chance? It's the one thing that spawns goroutines in the pubsub implementation.

romikser:
> Uh oh, do you mind opening an issue in https://github.com/libp2p/go-libp2p-pubsub?

I am using in as part of IPFS. I though it would be to start discussion here, because it may be a problem of pubsub configuration used in IPFS.



> I don't think announce retry is the culprit; this creates a short-lived goroutine, which reschedules a new instance of itself if it fails. So there should be only one goroutine active retrying to announce at any given time.
> The basic issue is that you are sending too fast and overflowing the send buffers; perhaps we should make these buffers configurable.

[...]

PR #5688, opened 2018-10-29
gx: update pubsub

Stebalien:
fixes a critical performance issue in the gossipsub pubsub router


vyzo:
Closes https://github.com/ipfs/go-ipfs/issues/5657

Issue #5694, opened 2018-10-30
Support WASM Architecture


magik6k:
So Go now Supports targeting WASM, but with certain limitations (like not having 'normal' syscalls). It's very low priority and probably a bit of work, but would be a nice 'benchmark' of portability.

First step would be to make the thing compile at all, currently stack of errors I get is down to:
```
# gx/ipfs/QmXTpwq2AkzQsPjKqFQDNY2bMdsAT53hUBETeyj8QRHTZU/sha256-simd
../../../gx/ipfs/QmXTpwq2AkzQsPjKqFQDNY2bMdsAT53hUBETeyj8QRHTZU/sha256-simd/cpuid.go:23:14: undefined: haveArmSha
../../../gx/ipfs/QmXTpwq2AkzQsPjKqFQDNY2bMdsAT53hUBETeyj8QRHTZU/sha256-simd/cpuid.go:27:16: undefined: cpuid
../../../gx/ipfs/QmXTpwq2AkzQsPjKqFQDNY2bMdsAT53hUBETeyj8QRHTZU/sha256-simd/cpuid.go:32:13: undefined: xgetbv
../../../gx/ipfs/QmXTpwq2AkzQsPjKqFQDNY2bMdsAT53hUBETeyj8QRHTZU/sha256-simd/cpuid.go:40:18: undefined: cpuid
../../../gx/ipfs/QmXTpwq2AkzQsPjKqFQDNY2bMdsAT53hUBETeyj8QRHTZU/sha256-simd/cpuid.go:44:19: undefined: cpuidex
../../../gx/ipfs/QmXTpwq2AkzQsPjKqFQDNY2bMdsAT53hUBETeyj8QRHTZU/sha256-simd/cpuid.go:52:18: undefined: cpuid
../../../gx/ipfs/QmXTpwq2AkzQsPjKqFQDNY2bMdsAT53hUBETeyj8QRHTZU/sha256-simd/cpuid.go:56:17: undefined: cpuid
../../../gx/ipfs/QmXTpwq2AkzQsPjKqFQDNY2bMdsAT53hUBETeyj8QRHTZU/sha256-simd/cpuid.go:61:14: undefined: xgetbv
../../../gx/ipfs/QmXTpwq2AkzQsPjKqFQDNY2bMdsAT53hUBETeyj8QRHTZU/sha256-simd/cpuid.go:62:20: undefined: cpuidex
[...]

PR #5695, opened 2018-10-30
Move persistent datastores to plugins

magik6k:
Mostly because https://github.com/ipfs/go-ipfs/issues/5694 (wasm support)

Since this is using preload it shouldn't affect binary size / performance / compatibility, and just allows to easily disable modules.


Stebalien:
Jenkins doesn't seem to like this change.

magik6k:
Fixed

PR #5053, opened 2018-05-31
[wip] ci: Run ipfs/js-ipfs-http-client & ipfs/interop tests

magik6k:
This PR:
* Runs tests from https://github.com/ipfs/interop
* And from https://github.com/ipfs/js-ipfs-api
* And pushes more statuses to github so we don't have to wait for everything to finish

TODO:
- [ ] Make sure the tests actually use provided go-ipfs
- [ ] Try to skip parts that don't touch go-ipfs
- [ ] Make sure tests are stable
- [ ] Deduplicate code with jenkins-libs
- [ ] Collect test results

PR #5697, opened 2018-10-30
commands: fix a bunch of tiny commands-lib issues

Stebalien:
* Always check errors returned by emit. Otherwise, we may not notice when the client goes away.
* Make sure to use EmitOnce instead of Emit when appropriate. Otherwise, we break javascript.

(thanks Magik6k for finding this before we cut the release...)

### DIFF
--- a/core/commands/dag/dag.go
+++ b/core/commands/dag/dag.go
@@ -188,7 +188,7 @@ format.
 			}
 			out = final
 		}
-		return res.Emit(&out)
+		return cmds.EmitOnce(res, &out)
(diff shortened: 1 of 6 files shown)


magik6k:
(there are some fixed in https://github.com/ipfs/go-ipfs/pull/5053/commits/bfffaf0aff0adc5ce27fe208f2d604d1d5756542 that are not here)

Stebalien:
@magik6k mind making a PR for that pin fix (along with a test).

PR #5700, opened 2018-10-30
coreapi: fix errisdir

Stebalien:
JavaScript expects this to be "this dag node is a directory". I'm almost of a mind to say "don't parse errors" but, well, we don't give any better alternatives.

### DIFF
--- a/core/coreapi/interface/errors.go
+++ b/core/coreapi/interface/errors.go
@@ -3,6 +3,6 @@ package iface
 import "errors"
 
 var (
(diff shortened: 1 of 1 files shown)


Stebalien:
> Well, after we extract coreapi, we could bundle some standard errors there and compile that with WASM to provide a "proper" way to parse errors in js :D

If we go the "generate the API" approach, we can define them central files. We should also use error codes (it turns out JS is fine parsing json "errors" in `X-Stream-Error` trailing headers).

PR #5701, opened 2018-10-30
gx: update go-path

Stebalien:
fixes the changed path cat error causing the js-ipfs-api tests to fail


magik6k:
This fails: https://github.com/ipfs/go-ipfs/blob/master/core/corehttp/gateway_test.go#L191

Stebalien:
Hm. So, we'll change an error message whatever we do. Thoughts on just changing the test?

magik6k:
The test was changed with the message IIRC, so it's fine

PR #5704, opened 2018-10-30
fix: return the shortest, completely resolved path in the resolve command

### LINKED ISSUE #5703: ipfs resolve isn't behaving well
From the interface-ipfs-core tests,

```
CID=$(ipfs dag put <<<'{"path": {"to": {"file": "asdfasdfasdf"}}}')
ipfs resolve /ipfs/$CID/path/to/file
```

should fail with "found non-link at given path".

Currently, it returns `/ipfs/$CID`.

Really, it *should* fail with "not a valid unixfs file". However, if we replace `/ipfs` with `/ipld`, it still returns the same result. In *that* case, it should probably just return the entire path (IMO, it shouldn't *error*).

However, neither of these fixes will make the JS tests pass.

Stebalien:
fixes #5703

This is a *breaking change*. We *used* to return an error if the path didn't end at a CID but, IMO, that's incorrect. We *should* allow resolution of paths that stop in the middle of an object.

### DIFF
--- a/test/sharness/t0160-resolve.sh
+++ b/test/sharness/t0160-resolve.sh
@@ -18,7 +18,7 @@ test_expect_success "resolve: prepare files" '
 '
 
 test_expect_success "resolve: prepare dag" '
-  dag_hash=$(ipfs dag put <<<"{\"a\": {\"b\": {\"c\": \"asdfasdfasdf\"}}}")
+  dag_hash=$(ipfs dag put <<<"{\"i\": {\"j\": {\"k\": \"asdfasdfasdf\"}}}")
 '
 
 test_resolve_setup_name() {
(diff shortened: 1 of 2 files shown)


core/coreapi/interface/path.go
```diff
@@ -125,6 +125,22 @@ func IpldPath(c cid.Cid) ResolvedPath {
 	}
 }
 
+// Simplify takes a ResolvedPath and returns a *new* ResovledPath that's as
+// short as possible. In the new resolved path, `p.Cid()` will be the same as
+// `p.Root()`.
+func Simplify(p ResolvedPath) ResolvedPath {
```
magik6k: Could be a method on the path object (`p.Simple()` would be a bit nicer IMO)

magik6k: Also, less stuff in `coreiface` namespace

Stebalien: SGTM.

Stebalien:
Closing in favor of just restoring existing behavior (for now).

test/sharness/t0160-resolve.sh
```diff
@@ -65,6 +65,8 @@ test_resolve_cmd() {
   test_resolve "/ipfs/$a_hash/b" "/ipfs/$b_hash"
   test_resolve "/ipfs/$a_hash/b/c" "/ipfs/$c_hash"
   test_resolve "/ipfs/$b_hash/c" "/ipfs/$c_hash"
+  test_resolve "/ipld/$dag_hash/a/b/c" "/ipld/$dag_hash/a/b/c"
```
magik6k: Shouldn't that actually resolve to `/iplds/$c_hash`?

Stebalien: This is a really confusing test. The DAG object is `{"a": {"b": {"c": "asdfasdf"}}}` (no links). I've changed it to `{"i": {"j": {"k": "asdfasdf"}}}`

Stebalien:
@djdv let'd tackle that in a followup PR (the current code was already doing that).

PR #5705, opened 2018-10-30
error on resolving non-terminal paths

Stebalien:
This isn't the *correct* fix but it (mostly) restores the previous behavior.

Note 1: It *also* fixes resolution of `/ipld` paths by avoiding `path.FromCid`.
Note 2: This does not preserve the error. That was a DagCBOR specific error that has nothing to do with the issue.

### DIFF
--- a/core/commands/resolve.go
+++ b/core/commands/resolve.go
@@ -135,9 +135,11 @@ Resolve the value of an IPFS DAG path:
 			return err
 		}
 
-		c := rp.Cid()
+		if rp.Remainder() != "" {
+			return fmt.Errorf("path does not end on a dag-node boundary")
+		}
 
-		return cmds.EmitOnce(res, &ncmd.ResolvedPath{Path: path.FromCid(c)})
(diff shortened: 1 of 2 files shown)


Stebalien:
Alternative to #5704 that mostly preserves the current behavior. In retrospect, I actually prefer just breaking this now (we're not going to pass the js-ipfs-api tests either way).

core/commands/resolve.go
```diff
@@ -135,9 +135,11 @@ Resolve the value of an IPFS DAG path:
 			return err
 		}
 
-		c := rp.Cid()
+		if rp.Remainder() != "" {
+			return fmt.Errorf("path does not end on a dag-node boundary")
```
magik6k: We may want to keep the original error too because https://github.com/ipfs/interface-ipfs-core/pull/385/files#diff-8ce9a93c2e33e2c37099701b755064a3L85

magik6k: i.e. `found non-link at given path`

Stebalien: That error message is just a side-effect of DagCBOR. But yeah, we can keep it for now.

PR #5707, opened 2018-10-30
[WIP] Simple bitswap performance test

eingenito:
**Goal** 
We'd like a way to create a suite of tests to check/measure relative performance, initially of bitswap changes. This is a simple example of that, and I'm curious what people think. It's got some benefits:
* Really fast to write: iptb is hard to beat for orchestrating ipfs nodes
* Easy to simulate simple interactions between nodes
* Probably pretty easy to set up a variety of DAG tree shapes for testing
* Perfectly good smoke test - these are full IPFS nodes going about their business

It's got a lot of problems:
* Not easy to simulate different network conditions
* It's resource intensive and the timings will approach meaninglessness at some no. of nodes.
* It's hard to build up useful performance data to report (like dupes, bytes/sec, bitswap broadcasts etc.) because ipfs doesn't make them readily available. Output here looks like:
```

node1-receive-elapsed-sec: 0.370
node1-receive-duplicate_blocks: 0
node2-receive-elapsed-sec: 0.444
node2-receive-duplicate_blocks: 35
node3-receive-elapsed-sec: 0.571
node3-receive-duplicate_blocks: 78
node4-receive-elapsed-sec: 0.924
node4-receive-duplicate_blocks: 113
node5-receive-elapsed-sec: 0.725
node5-receive-duplicate_blocks: 157
node6-receive-elapsed-sec: 0.837
node6-receive-duplicate_blocks: 180
node7-receive-elapsed-sec: 1.074
node7-receive-duplicate_blocks: 231
```

**Alternatives**
* There are already some tests that do much more than this in https://github.com/ipfs/go-bitswap using a mocked network. 
* We could drive iptb with Go for better control over iptb and better reports. 
* We could drive tests with JS like our interop tests.

**Todo**
This is just a trial effort; we'd want these in test/bench with their own make target, and we'd want to test some more diverse data. 

**Question**
Do we want benchmarks like this?

### DIFF
--- a/test/sharness/t0126-bitswap-perf.sh
+++ b/test/sharness/t0126-bitswap-perf.sh
@@ -1,0 +1,75 @@
+#!/usr/bin/env bash
+#
+# Copyright (c) 2017 Jeromy Johnson
+# MIT Licensed; see the LICENSE file in this repository.
+#
+
+test_description="Test multiple ipfs nodes transferring files via bitswap"
+timing_file="`basename "$0"`-timings.tmp"
+node_count=8
+file_size_kb=10000
+
+. lib/test-lib.sh
+
+time_expect_success() {
+    { time {
+    test_expect_success "$1" "$2";
+    TIMEFORMAT=%5R;
+    }; } 2>> $timing_file
+}
+
+check_file_fetch() {
+  node=$1
+  fhash=$2
+  fname=$3
+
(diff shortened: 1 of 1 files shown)


eingenito:
I see some more interesting tests that I hadn't noticed like this one: https://github.com/ipfs/go-ipfs/blob/master/test/integration/three_legged_cat_test.go. I think I like that better. I'm going to close this for now.

PR #5708, opened 2018-10-31
re-format log output as ndjson

Stebalien:
Apparently, js-ipfs-api expects this. Really, I'm not sure why js-ipfs-api even cares about this endpoint but that's a different issue.

The real fix is to change back to ndjson in go-log but that would force us to update every package (not good when we're trying to push out a release like this).


Stebalien:
(the tests did pass, jenkins is just being cranky)

Issue #5696, opened 2018-10-30
ipfs repo verify is single threaded


baryluk:
```
go-ipfs version: 0.4.18-rc1-
Repo version: 7
System version: amd64/linux
Golang version: go1.11.1
```

`ipfs repo verify` takes ages (hours) on big repos, and it only is using single core/thread on my machine.

Stebalien:
Fixed by @chenminjian. @baryluk, can you try out the latest master?

PR #5710, opened 2018-10-31
enhance(cmd/verify): add goroutine count to improve verify speed

chenminjian:
refs: #5696 

License: MIT

### DIFF
--- a/core/commands/repo.go
+++ b/core/commands/repo.go
@@ -2,12 +2,15 @@ package commands
 
 import (
 	"bytes"
+	"context"
 	"errors"
 	"fmt"
 	"io"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
+	"sync"
 	"text/tabwriter"
 
 	oldcmds "github.com/ipfs/go-ipfs/commands"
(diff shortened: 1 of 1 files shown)


chenminjian:
hey, @Stebalien . I use `Ctrl + C` to interrupt `ipfs repo verify`, then  IPFS `panic: send on closed channel`. It looks like we can't use `goroutine` and `select` simply. I am going to use `sync.WaitGroup` to avoid it. Does that make sense? Can you give me some advice? Thx a lot.

core/commands/repo.go
```diff
@@ -302,13 +302,35 @@ var repoVerifyCmd = &oldcmds.Command{
...
 		var i int
+		var count int
+		verify := make(chan string)
+		defer close(verify)
+
 		for k := range keys {
-			_, err := bs.Get(k)
-			if err != nil {
+			go func() {
+				_, err := bs.Get(k)
```
overbool: `k` should be passed in by the `go func(k cid.Cid){...}(k)`

core/commands/repo.go
```diff
@@ -302,13 +302,35 @@ var repoVerifyCmd = &oldcmds.Command{
...
 		var fails int
 		var i int
+		var count int
+		verify := make(chan string)
+		defer close(verify)
+
 		for k := range keys {
-			_, err := bs.Get(k)
-			if err != nil {
+			go func() {
```
Stebalien: So, spawning a goroutine per key is going to blow through memory like nothing else. We'll have to:

1. Span N goroutines where N is something like `runtime.NumCPU * 2` (ish?).
2. Have them all read off the keys channel.
3. Have them all write to the same "results" channel.
4. Have them call `Done` on a wait-group on exit.
5. Spin up another goroutine that closes the "results" channel when the wait-group finishes.
6. Outside, have a for loop reading off the results channel.

core/commands/repo.go
chenminjian: Thx a lot.

core/commands/repo.go
chenminjian: This is great advice. Thx a lot. :)

PR #5712, opened 2018-11-01
cmds/env: add a config path helper

overbool:
License: MIT

### DIFF
--- a/core/commands/cmdenv/env.go
+++ b/core/commands/cmdenv/env.go
@@ -40,3 +40,13 @@ func GetConfig(env cmds.Environment) (*config.Config, error) {
 
 	return ctx.GetConfig()
 }
+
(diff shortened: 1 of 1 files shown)


core/commands/cmdenv/env.go
```diff
@@ -40,3 +40,13 @@ func GetConfig(env cmds.Environment) (*config.Config, error) {
 
 	return ctx.GetConfig()
 }
+
+// GetConfigPath extracts the config path from the environment
+func GetConfigPath(env cmds.Environment) (string, error) {
```
kevina: I would name this `GetConfigRoot`, to be consistent with the variable it is accessing.  I agree it not the best name, but I think it better to be consistent here.

overbool: @kevina Changed it.

Issue #5716, opened 2018-11-02
Add a `--force` flag to `pin rm` to ignore nonexistent pins and avoid bailing out


schomatis:

overbool:
@schomatis @kevina 

In the original implementation, `ipfs pin rm <invalid-hash> <valid-hash> <invalid-hash>` will print `Error: not pinned` and stop unpinning the `<valid-hash>`. When we add `--force` option, Did you mean that we don't need to print `Error: not pinned` and continue to unpin `<valid-hash>`? If so, maybe we should modify the implementation about `Unpin()` in `core/corerepo/pinning.go` in the meantime. Am I right?

kevina:
I think unpin should be modified to continue to remove all valid pins.  It should by default print an error on the invalid hash but continue anyway and then at the end print an error that not all pins can be removed at the end.  If force is given it should _ignore_ non-existent blocks but continue anyway and not return an error.  In other words, the only think `--force` should do is not print or return an error.

`ipfs block rm` is a good model for this.

kevina:
@schomatis what I am saying is that `pin rm` should not bail out even if the `--force` flag is _not_ given.

Stebalien:
So, forcing without a `--force` flag is a noticeable breaking change. How about:

1. Collecting the errors as @overbool is doing in https://github.com/ipfs/go-ipfs/pull/5843.
2. If `--force` is not specified, emitting a final "errors occurred" error like we do in `ipfs repo gc`.

That way:

1. We don't break anything.
2. We still try to remove what we can.

---

We also need to involve the JS people. It looks like there was already a discussion here: https://github.com/ipfs/interface-ipfs-core/issues/127.

Stebalien:
Let's move the discussion to https://github.com/ipfs/interface-ipfs-core/issues/127#issuecomment-456561650.

PR #5717, opened 2018-11-02
cmds/pin: add force option for pin rm

overbool:
Fixes: https://github.com/ipfs/go-ipfs/issues/5716

License: MIT

### DIFF
--- a/test/sharness/t0085-pins.sh
+++ b/test/sharness/t0085-pins.sh
@@ -8,6 +8,7 @@ test_description="Test ipfs pinning operations"
 
 . lib/test-lib.sh
 
+RANDOM_HASH=Qme8uX5n9hn15pw9p6WcVKoziyyC9LXv4LEgvsmKMULjnV
 
 test_pins() {
   EXTRA_ARGS=$1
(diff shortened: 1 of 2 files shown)


kevina:
@overbool could you add a sharness test for this?

test/sharness/t0085-pins.sh
```diff
@@ -56,6 +57,10 @@ test_pins() {
     cat hashes | ipfs pin rm
   '
 
+  test_expect_success "unpin non-existent hashes with force option" '
```
kevina: To make sure things are working right I would do
```
ipfs pin rm --force <invalid-hash> <valid-hash> <invalid-hash>
```

overbool:
@kevina @schomatis Currently, we had implemented the `PinAPI.Rm` in `coreapi/pin.go`, but we haven't used it yet. Should we use `PinAPI.Rm`?

Another question is should we deprecate `corerepo/pinning.go`? Because I think we can implement all function about `pin` in `coreapi/pin.go`.

schomatis:
I'm not sure, @magik6k might know more about that, but those issues seem more suited to be handled in another PR to avoid mixing different solutions.

magik6k:
Sorry for the delay. Yep, we can remove `corerepo/pinning.go` and just use CoreAPI.

magik6k:
I'm guessing we want https://github.com/ipfs/go-ipfs/pull/5843 in first

PR #5721, opened 2018-11-02
Fix broken links

rmorey:
A few links to docs in this repo needed "blob/master/" in order to work properly.


Stebalien:
@rmorey oops, thanks! Mind fixing the commit message (probably not that critical for this change but it's easier if we just apply this rule to all commits).

olizilla:
The same issue exists in the release notes https://github.com/ipfs/go-ipfs/releases/tag/v0.4.18

Stebalien:
(we really don't need a signoff for something like this and are planning on getting rid of them anyways...)

PR #5741, opened 2018-11-05
gx: fix broken import

magik6k:
Probably sneaked in with some outdated PR


magik6k:
(current hash in package.json - https://github.com/ipfs/go-ipfs/blob/master/package.json#L521)

Stebalien:
(bypassing CI, thanks!)

Issue #5736, opened 2018-11-05
ipfs add command output error when file size is 0.


chenminjian:
#### Version information:

go-ipfs version: 0.4.18-dev-59d0ec287
Repo version: 7
System version: amd64/darwin
Golang version: go1.11.1

#### Type:
Bug

#### Description:

```
$ touch d
$ ipfs add d
added QmbFMke1KXqnYyBBWxB74N4c5SBnJMVAiMNRcGu6x1AwQH d
 0 B / ? [------------------------------------------------------------=]   0.00%
```
It's probably progressbar library’s bug.

Stebalien:
There are two bugs:

1. The library bug you mention. It can't handle 0 totals.
2. A bug in our usage where we don't *set* the total if we finish too quickly (e.g., `echo "a" | ipfs add`).

The second bug can be fixed by setting the total to the current value after the progressbar update loop. The first bug needs to be fixed upstream.

chenminjian:
Ok, I will fix the second bug first.

Stebalien:
Status: @chenminjian has fixed the second bug.

PR #5743, opened 2018-11-06
fix(cmd/add): progressbar output error when input is read from stdin

chenminjian:
refs: #5736

License: MIT

### DIFF
--- a/core/commands/add.go
+++ b/core/commands/add.go
@@ -327,6 +327,14 @@ You can now check what blocks have been created by:
 						return
 					}
 				}
+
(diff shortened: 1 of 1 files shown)


core/commands/add.go
```diff
@@ -327,6 +327,14 @@ You can now check what blocks have been created by:
 						return
 					}
 				}
+
+				if bar.Total == 0 && bar.Get() != 0 {
```
Stebalien: We should only do this if `progress` is true.

chenminjian: Thx a lot. I have made changes to this.

PR #5751, opened 2018-11-08
block: rm extra output

overbool:
License: MIT

### DIFF
--- a/core/commands/block.go
+++ b/core/commands/block.go
@@ -240,13 +240,13 @@ It takes a list of base58 encoded multihashes to remove.
 
 			err = api.Block().Rm(req.Context, rp, options.Block.Force(force))
 			if err != nil {
-				err := res.Emit(&util.RemovedBlock{
+				if err := res.Emit(&util.RemovedBlock{
 					Hash:  rp.Cid().String(),
 					Error: err.Error(),
-				})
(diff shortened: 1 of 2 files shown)


core/commands/block.go
```diff
@@ -249,13 +249,13 @@ It takes a list of base58 encoded multihashes to remove.
 
 			err = api.Block().Rm(req.Context, rp, options.Block.Force(force))
 			if err != nil {
-				err := res.Emit(&util.RemovedBlock{
+				if err := res.Emit(&util.RemovedBlock{
 					Hash:  rp.Cid().String(),
 					Error: err.Error(),
-				})
-				if err != nil {
+				}); err != nil {
```
kevina: @Stebalien may disagree with me, but I would keep the original syntax.  I had to study this to figure out what is going on since the function call was not on a single line and there are a lot of extra parenthesis involved.

Stebalien: I really don't care one way or another. To me, both are clear.

Stebalien:
@overbool mind making a sharness test for this issue?

Stebalien:
@kevina does this conflict with your work or can I merge it.

kevina:
@Stebalien this is minor enough that it shouldn't cause a problem.

Issue #5724, opened 2018-11-02
Gateways should always display "Index of /ipfs/HASH", instead of the URL.


swedneck:
#### Version information:
go-ipfs version: 0.4.18-
Repo version: 7
System version: amd64/linux
Golang version: go1.11.1

#### Type:
Bug/Feature?

#### Description:
Currently when viewing a directory on a gateway, it will display "Index of <url>", which can be anything from `/` to `/ipns/domain.com` to `/ipfs/HASH`.
![](https://ipfs.io/ipfs/QmcSyAiyX3t86aB8bJ2UrQ6kJef1xHcooWVXidn6Aw979e/2018-11-03_00-19-23.png)

I think it should instead ALWAYS display the ipfs hash that's being shown, as this is both more consistent, more useful in case of screenshots (have fun trying to access a dnslink in 20 years), and makes better use of space since the URL will already be displayed in the browser's address bar.
![](https://ipfs.io/ipfs/QmU8c7fbJwDnkY7JPfrh4MLLDncdbECwbPdPf9hTL4Tjer/2018-11-03_00-21-24.png)

There's also the option of displaying both the IPFS hash, and the URL, which is probably the best option if it's not difficult to implement.

Stebalien:
I thought for *sure* I had commented on this somewhere, but I guess not.

I agree we should show the resolved path, however, I'm not sure about *replacing* the existing path as it's quite useful from a UX standpoint.

How about putting it in a footer or, maybe grayed out flushed right? That is:

```
Index of /ipns/ipfs.io/a/b                                                (QmOther)
```

Anyone else with opinions?

jackloughran:
@Stebalien grayed out and flushed right sounds good. Would it be useful to make the grayed out hash only appear if the existing path does not contain the hash? Showing both hashes at once is going to make that line pretty long.

jackloughran:
Something like this could work too:

![screenshot_2018-11-13 ipns litmem es](https://user-images.githubusercontent.com/30052269/48414581-2376ff80-e719-11e8-8c96-3c3c186cacaa.png)

It would be easier to see than a footer, and it wouldn't break as easily on smaller screens.

Stebalien:
That looks wonderful. I'm sorry for leaving this so long; if you have time, please resurrect your PR (didn't notice the progress on this issue till I saw you close it).

PR #5752, opened 2018-11-08
show hash on gateway listing

jackloughran:
This incorporates a change to the directory listing template to display the hash instead of the path on the directory listing page for #5724 

This uses https://github.com/ipfs/dir-index-html/pull/20


Stebalien:
(blocked on discussion)

Issue #5758, opened 2018-11-09
Sorted bitswap wantlist


Stebalien:
The fact that the wantlist is always returned in a random order tends to confuse users (and me!) when trying to determine if we're making progress on an `ipfs get`.

Proposal: sort the wantlist before returning it to the user.

PR #5759, opened 2018-11-10
cmds/bitswap: sort wantlist

overbool:
Fixes: https://github.com/ipfs/go-ipfs/issues/5758

License: MIT


core/commands/bitswap.go
```diff
@@ -64,18 +65,24 @@ Print out all blocks currently on the bitswap wantlist for the local peer.`,
...
-
-			return cmds.EmitOnce(res, &KeyList{bs.WantlistForPeer(pid)})
 		}
+
 		return cmds.EmitOnce(res, &KeyList{bs.GetWantlist()})
 	},
 	Encoders: cmds.EncoderMap{
 		cmds.Text: cmds.MakeTypedEncoder(func(req *cmds.Request, w io.Writer, out *KeyList) error {
+			// sort the keys first
+			keys := make([]string, 0, len(out.Keys))
```
Stebalien: So, we can actually just:

1. import `github.com/ipfs/go-cidutil`.
2. Call `cidutil.Sort(out.Keys)`

overbool: @Stebalien Thx your advise. I had changed it.

Issue #5140, opened 2018-06-19
Handle the case where GC is disabled but a repo limit is configured


Stebalien:
When a repo size limit is configured bug gc is disabled (i.e., no `--enable-gc` flag has been passed), we should either (or both?):

1. Warn on start.
2. Refuse to write to the datastore once the size has been exceeded.

lgierth:
> Warn on start

:+1: This would have to be something like `if Datastore.MaxStorage != defaultMaxStorage` I guess

> Refuse to write to the datastore once the size has been exceeded.

This would probably lead to situations where your node is running fine but has exceeded MaxStorage, you update go-ipfs or restart for whatever reason, and suddenly it refuses to start. We could combine this with MaxStorage usage mandating --enable-gc though, i.e. it'd run GC right away in that case. (And use the warning on start as a migration path.)

parkan:
👍 

ability to quietly exceed maxStorage is definitely not what I would expect -- warning on start would be a good start

postables:
I like both options. I think you could have multiple levels of warnings:
1) warning when datastore usage is at 90%
2) non fatal error when datastore usage is at 95%
3) fatal error, preventing startup when datastore usage is at 99%

@lgierth  makes a good point about preventing writting to the data store once the size is exceeded,  however I would be *more* concerned about what disastrous potentials attempting to write to a datastore that you have exceeded limits for. What if by attempting to write to a datastore whose limits you passed you overwrite already existing data, or fill up the remaining space too much such that an application which *needs* that space may not run properly? 

In the  (my) ideal world, the node would fail to write to the datastore when it fills up, and entering a sort of "locked" state, where you can serve content but are prevented from writting new data, and only GC'ing data.

kjzz:
Hey @Stebalien , Did you have finished this issue?If not , I want to do for this. Because i want to learn more about gc through this issue.

Stebalien:
Go ahead.

kjzz:
Hey @Stebalien, i have create a pr about this issue , please help me review it?

PR #5768, opened 2018-11-12
handle max storage when gc is disabled

kjzz:
feature #5140 

License: MIT

### DIFF
--- a/cmd/ipfs/daemon.go
+++ b/cmd/ipfs/daemon.go
@@ -629,17 +629,49 @@ func mountFuse(req *cmds.Request, cctx *oldcmds.Context) error {
 }
 
 func maybeRunGC(req *cmds.Request, node *core.IpfsNode) (<-chan error, error) {
-	enableGC, _ := req.Options[enableGCKwd].(bool)
-	if !enableGC {
+	gc, err := initGC(req, node)
+	if err != nil {
+		return nil, err
+	}
+	if !gc.EnableGC {
+		log.Warningf("GC: %s", "Repo GC is disabled.")
 		return nil, nil
 	}
 
 	errc := make(chan error)
-	go func() {
-		errc <- corerepo.PeriodicGC(req.Context, node)
-		close(errc)
-	}()
+	// if duration larger than 0, it means GC is enabled.
+	if int64(gc.GCPeriod) > 0 {
+		go func() {
+			errc <- corerepo.PeriodicGC(req.Context, gc)
+			close(errc)
+		}()
+	}
+
 	return errc, nil
+}
+
+func initGC(req *cmds.Request, node *core.IpfsNode) (*corerepo.GC, error) {
+	enableGC, _ := req.Options[enableGCKwd].(bool)
+	cfg, err := node.Repo.Config()
+	if err != nil {
+		return nil, err
+	}
+
+	if cfg.Datastore.GCPeriod == "" {
+		log.Warning("GCPeriod isn't set, the default value is 1h")
+		cfg.Datastore.GCPeriod = "1h"
+	}
+
+	gc, err := corerepo.NewGC(node)
+	if err != nil {
+		return nil, err
+	}
+	gc.EnableGC = enableGC
+
+	node.StorageMax = gc.StorageMax
+	node.EnableGC = gc.EnableGC
+
+	return gc, nil
 }
 
 // merge does fan-in of multiple read-only error channels
--- a/core/coreapi/unixfs.go
+++ b/core/coreapi/unixfs.go
@@ -2,6 +2,7 @@ package coreapi
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/ipfs/go-ipfs/core"
@@ -22,6 +23,10 @@ import (
 	cidutil "gx/ipfs/QmbfKu17LbMWyGUxHEUns9Wf5Dkm8PT6be4uPhTkk4YvaV/go-cidutil"
 	ipld "gx/ipfs/QmcKKBwfz6FyQdHR2jsXrrF6XeSBXYL86anmWNewpFpoF5/go-ipld-format"
 	mfs "gx/ipfs/QmcUXFi2Fp7oguoFT81f2poJpnb44dFkZanQhDBHMoYyG9/go-mfs"
+)
+
+var (
+	errFullStorage = errors.New("DataStorage is full, please enable gc and restart node")
 )
 
 type UnixfsAPI CoreAPI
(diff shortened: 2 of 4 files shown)


core/coreapi/unixfs.go
```diff
@@ -24,6 +25,10 @@ import (
 	mfs "gx/ipfs/QmcUXFi2Fp7oguoFT81f2poJpnb44dFkZanQhDBHMoYyG9/go-mfs"
 )
 
+var (
+	errFullStorage = errors.New("DataStorage is full,please enable gc and restart node")
```
schomatis: nit: add a space after the "," (same for the log errors below)

core/coreapi/unixfs.go
```diff
@@ -126,6 +131,23 @@ func (api *UnixfsAPI) Add(ctx context.Context, files files.File, opts ...options
 		fileAdder.SetMfsRoot(mr)
 	}
 
+	// Get storageMax when node is online.
```
schomatis: Is `Add` the only entry point that will make the repo grow?

core/coreapi/unixfs.go
```diff
@@ -126,6 +131,23 @@ func (api *UnixfsAPI) Add(ctx context.Context, files files.File, opts ...options
...
+	if fileAdder.Pin && !n.EnableGC && n.OnlineMode() {
```
schomatis: Do we need to check for `Pin` here? the file will be added anyway even if we don't pin it, right? (it can be removed later but it will make the repo grow)

core/coreapi/unixfs.go
```diff
@@ -126,6 +131,23 @@ func (api *UnixfsAPI) Add(ctx context.Context, files files.File, opts ...options
...
+		storageUsage, err := n.Repo.GetStorageUsage()
```
schomatis: Is `GetStorageUsage` a faithful metric of repo size? Is it compatible with `StorageMax`? (this is a naive question but I'm not familiar with the repo API)

cmd/ipfs/daemon.go
```diff
@@ -629,19 +629,50 @@ func mountFuse(req *cmds.Request, cctx *oldcmds.Context) error {
...
 
 func maybeRunGC(req *cmds.Request, node *core.IpfsNode) (<-chan error, error) {
-	enableGC, _ := req.Options[enableGCKwd].(bool)
-	if !enableGC {
+	gc, err := initGC(req, node)
+	if err != nil {
+		return nil, err
+	}
+	if !gc.EnableGC {
+		log.Warningf("GC: %s", "Repo GC is disabled.")
```
schomatis: Why do we need this warning? Is it part of the original intention of the PR?

core/corerepo/gc.go
```diff
@@ -23,10 +23,12 @@ var ErrMaxStorageExceeded = errors.New("maximum storage limit exceeded. Try to u
 type GC struct {
 	Node       *core.IpfsNode
 	Repo       repo.Repo
+	EnableGC   bool
```
schomatis: Why do we need this second `EnableGC`?

cmd/ipfs/daemon.go
```diff
@@ -629,19 +629,50 @@ func mountFuse(req *cmds.Request, cctx *oldcmds.Context) error {
...
+
+	if cfg.Datastore.GCPeriod == "" {
+		cfg.Datastore.GCPeriod = "1h"
+	}
+
+	gc, err := corerepo.NewGC(node)
+	if err != nil {
+		return nil, err
+	}
+	gc.EnableGC = enableGC
```
schomatis: Let's keep `enableGC` set-value and load-value operations as close as possible, move the `enableGC, _ := req.Options[enableGCKwd].(bool)` line down here if it's not used anywhere else.

cmd/ipfs/daemon.go
```diff
@@ -629,19 +629,50 @@ func mountFuse(req *cmds.Request, cctx *oldcmds.Context) error {
...
 
+func initGC(req *cmds.Request, node *core.IpfsNode) (*corerepo.GC, error) {
+	enableGC, _ := req.Options[enableGCKwd].(bool)
+	cfg, err := node.Repo.Config()
+	if err != nil {
+		return nil, err
+	}
+
+	if cfg.Datastore.GCPeriod == "" {
+		cfg.Datastore.GCPeriod = "1h"
```
schomatis: Unrelated to this PR (since this was already present in the code) but could we add a warning here to notify the user the `GCPeriod` wasn't set and we're using a default value of an hour?

core/coreapi/unixfs.go
```diff
@@ -126,6 +131,23 @@ func (api *UnixfsAPI) Add(ctx context.Context, files files.File, opts ...options
...
+		if err != nil {
+			return nil, err
+		}
+		storPercentage := float64(storageUsage) / float64(n.StorageMax)
```
schomatis: From `gc.go` I'm seeing that `StorageMax` can be left unset by the user and we'll silently assume a 10GB limit which can make this test a little brittle, a user who just doesn't care about any of this (didn't set the GC, didn't set a max capacity for the repo) will hit this error (which will even prevent it from using IPFS) without understanding what's really going on (and how to fix it).

schomatis:
Left several comments but the most important to address first is https://github.com/ipfs/go-ipfs/pull/5768#discussion-diff-235005123R140 (which may require a change in the approach of the PR's proposed solution).

schomatis:
ping @kjzz

kjzz:
Sorry @schomatis,recently I'm preparing the opening report of my graduation thesis.I will do for this on next week.Thx.

schomatis:
Hey no rush, take your time, definitively concentrate on your thesis now :)

cmd/ipfs/daemon.go
kjzz: > Is it part of the original intention of the PR?
We should print some warning message when the gc is disable.
You can see the comment in the [issue](https://github.com/ipfs/go-ipfs/issues/5140#issue-333793980)

core/corerepo/gc.go
kjzz: Because I need to get this param in `func maybeRunGC(req *cmds.Request, node *core.IpfsNode) (<-chan error, error) `.

cmd/ipfs/daemon.go
kjzz: Good idea!

core/coreapi/unixfs.go
kjzz: In my testing , maybe `GetStorageUsage ` is a faithful metric of repo size.

cmd/ipfs/daemon.go
kjzz: Because i should get this param in other function.
```
func maybeRunGC(req *cmds.Request, node *core.IpfsNode) (<-chan error, error) {
	gc, err := initGC(req, node)
	if err != nil {
		return nil, err
	}
	if !gc.EnableGC {
		log.Warningf("GC: %s", "Repo GC is disabled.")
		return nil, nil
	}
.........
```

kjzz:
@schomatis  sorry for delay,i have update my pr.please help me review it again.Thx a lot!

core/corerepo/gc.go
```diff
@@ -157,35 +165,13 @@ func GarbageCollectAsync(n *core.IpfsNode, ctx context.Context) <-chan gc.Result
...
-		return nil
-	}
-
-	gc, err := NewGC(node)
-	if err != nil {
-		return err
-	}
-
+// PeriodicGC create a goroutine to gc.
+func PeriodicGC(ctx context.Context, gc *GC) error {
```
Stebalien: At this point, this should just be a function on the `GC` instance.

core/coreapi/unixfs.go
Stebalien: No. Unfortunately, this _really_ needs to go _much_ lower. We probably need some kind of datastore wrapper that rejects writes when the datastore is full (and calls a user-provided "free" function at N% of full).

PR #5769, opened 2018-11-12
Suppress bootstrap error

zonque:
When used in private network mode with `Bootstrap` set to `[]`, the daemon currently spits out the following message ever 30 seconds:

```
ERROR       core: no bootstrap nodes configured: go-ipfs may have difficulty connecting to the network bootstrap.go:144
```

While it is good to mention, I see two issues with this:

* In regular environments, running the daemon with no bootstrap nodes will in fact render it unusable, so there should be an error at daemon start time instead of a repetitive message
* For private network setups where nodes discover themselves via MDNS, this message is unnecessary.

This PR changes both the above, with one patch for each of them.

### DIFF
--- a/core/bootstrap.go
+++ b/core/bootstrap.go
@@ -77,6 +77,12 @@ func Bootstrap(n *IpfsNode, cfg BootstrapConfig) (io.Closer, error) {
 	// make a signal to wait for one bootstrap round to complete.
 	doneWithRound := make(chan struct{})
 
+	if len(cfg.BootstrapPeers()) == 0 {
+		// We *need* to bootstrap but we have no bootstrap peers
+		// configured *at all*, inform the user.
+		log.Error("no bootstrap nodes configured: go-ipfs may have difficulty connecting to the network")
+	}
+
 	// the periodic bootstrap function -- the connection supervisor
 	periodic := func(worker goprocess.Process) {
 		ctx := procctx.OnClosingContext(worker)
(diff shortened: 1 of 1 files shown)


zonque:
Hmm, so some test now fail because they instantiate daemons without bootstrap nodes. Can anyone give me a hint on how to quickly fix that?

Stebalien:
> For private network setups where nodes discover themselves via MDNS, this message is unnecessary.

I'm not sure I agree with this. Really, MDNS and private networks are unrelated (unless you mean private networks via private VPNs). Most private networks (that use swarm keys) will *want* private bootstrap nodes.

However, I do agree that this is *way* too noisy. It may make sense to just print a warning *once* and never again.

Stebalien:
The tests are probably failing because bootstrappers can be added and removed at runtime.

zonque:
> I'm not sure I agree with this. Really, MDNS and private networks are unrelated (unless you mean private networks via private VPNs). Most private networks (that use swarm keys) will _want_ private bootstrap nodes.
> 
> However, I do agree that this is _way_ too noisy. It may make sense to just print a warning _once_ and never again.

Hmm, okay. Agreed.

I've force-pushed a new version that does just that.

Thanks for your input!

Stebalien:
LGTM but I'd like someone else to take a quick look for a sanity check.

schomatis:
> The tests are probably failing because bootstrappers can be added and removed at runtime.

Can this be merged with the test errors?

Stebalien:
> LGTM, I would change the error for a warning (but it's just a personal preference so feel free to ignore it).

We *should*, but I'd like to enable warnings by default first. Otherwise, users will never see it.

Stebalien:
> Can this be merged with the test errors?

That has been fixed. The first iteration *aborted* the construction of the periodic bootstrapper when no bootstrappers were configured. This one just warns and continues.

PR #5771, opened 2018-11-13
p2p: report-peer-id option for listen

magik6k:
Was suggested on IRC, when set it basically passes `QmPeerid\n` of connecting peer into listener stream.

(works but needs tests)

### DIFF
--- a/core/commands/p2p.go
+++ b/core/commands/p2p.go
@@ -53,6 +53,7 @@ type P2PStreamsOutput struct {
 
 const (
 	allowCustomProtocolOptionName = "allow-custom-protocol"
+	reportPeerIDOptionName        = "report-peer-id"
 )
 
 var resolveTimeout = 10 * time.Second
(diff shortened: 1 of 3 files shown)


Stebalien:
We should also consider sending a binary peer ID (maybe length-prefixed?, although peer IDs are already prefix-free codes). This may be easier to parse than having to read to a newline and then potentially base58 decode. It also means we can change encoding without breaking things.

However, I haven't thought though this thuroughly.

magik6k:
Technically they carry the multihash prefix which has length. I can see binary form being useful, but I can't think of any nice way to do this other than 2 separate flags

Stebalien:
Meh. The current version *is* quite user friendly.

Any objections @ipfs/go-team?

whyrusleeping:
Eh, I think that practically, the base58 encoded version will be the most useful. If we make it binary, writing a parser becomes just that much more difficult.

Issue #5772, opened 2018-11-13
curl local ipfs server and return 500


chenminjian:
#### Version information:
go-ipfs version: 0.4.19-dev-
Repo version: 7
System version: amd64/darwin
Golang version: go1.11.1

#### Type:
Bug

#### Description:

![wx20181113-111255](https://user-images.githubusercontent.com/13758621/48393613-40c4c180-e74b-11e8-8646-8d8128872bbc.png)

When the mfs file does not exist, I send `files stat` request to ipfs server and return 500.
Even if the file does not exist, we should return 200. Does that make sense?

eingenito:
Makes sense to me.

Stebalien:
(moving from https://github.com/ipfs/go-ipfs/pull/5775#issuecomment-439507822)

I don't think this is the right solution:

* `Type` is the *file type*.
* This is an RPC API, not a normal HTTP endpoint. For us: 200 means "it worked", 404 means "that API endpoint doesn't exist", 400 means "you passed a bad option", *500* means "some error occurred when processing a well-formed request".

Note: I'm not *happy* with how we're using HTTP status codes, that's just how the API currently works. IMO, we should only be using HTTP errors for *transport* errors (missing endpoint, malformed request, command *paniced*, etc) and should have *status* information encoded in the response. However, that ship has kind of sailed.

chenminjian:
Yes, this is not just a problem with `files stat`. We should have status information encoded in the response.

Stebalien:
Unfortunately, it's too late for that. We use the status code to say whether or not the result is an error (any error) and then put the value in the result.

Basically, forget that we're using HTTP here. We *happen* to be using HTTP as the transport but we've chosen our own semantics. They suck but the only real solution is to create an entirely new HTTP API protocol (which is on our TODO list, just not at the top yet).

I'm going to close this as "can't fix" for now as there's really not much we can do without breaking the API.

Stebalien:
(closing for tracking, we can continue discussing this)

PR #5775, opened 2018-11-14
[WIP]fix(cmd/stat): even if file doesn't exist, server should return 200

chenminjian:
refs: #5772 

License: MIT

### DIFF
--- a/core/commands/files.go
+++ b/core/commands/files.go
@@ -147,9 +147,17 @@ var filesStatCmd = &cmds.Command{
 			dagserv = node.DAG
 		}
 
-		nd, err := getNodeFromPath(req.Context, node, api, path)
-		if err != nil {
-			return err
+		var nd ipld.Node
+		if strings.HasPrefix(path, "/ipfs/") {
+			nd, err = getNodeFromDag(req.Context, api, path)
+			if err != nil {
+				return err
+			}
+		} else {
+			nd, err = getNodeFromMfs(node, path)
+			if err != nil {
+				return err
+			}
 		}
 
 		o, err := statNode(nd)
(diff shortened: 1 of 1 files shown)


core/commands/files.go
```diff
@@ -147,9 +148,18 @@ var filesStatCmd = &cmds.Command{
...
-			return err
+		var nd ipld.Node
+		if strings.HasPrefix(path, "/ipfs/") {
+			nd, err = getNodeFromDag(req.Context, api, path)
+			if err != nil {
+				return err
+			}
+		} else {
+			nd, err = getNodeFromMfs(node, path)
+			if err != nil {
```
schomatis: This will return *any* error form MFS, ideally we should check for errors like `os.ErrNotExist` like in

https://github.com/ipfs/go-unixfs/blob/c54d0e47ad42f03de7736c2826c315de9f63b5a6/io/directory.go#L187-L188

Unfortunately I don't think this is guaranteed in the API at the moment, meaning that `BasicDirectory` does return that error (although I'm not sure if it does in all the appropriate cases) and  `HAMTDirectory` definitively doesn't.

So, let's merge this as is, but please add a big `TODO:` here requesting a stable error from the MFS API.

chenminjian:
This commit hides the key points and is not convenient for review. I have split it into two commits. Thank you for your advice. :)

core/commands/files.go
chenminjian: https://github.com/ipfs/go-unixfs/blob/c54d0e47ad42f03de7736c2826c315de9f63b5a6/hamt/hamt.go#L396

It seems that `HAMTDirectory` also returns `os.ErrNotExist`. Is it better to add `os.ErrNotExist` check?

schomatis: Good catch! So, I'm not sure we're returning that error in all the appropriate circumstances, it depends on your main objective, if you want to err on the side of returning maybe more than real "file doesn't exist" (`os.ErrNotExist`) errors or in avoiding the 500 internal error when we actually  should just have returned the `os.ErrNotExist` error, both options are fine by me.

Stebalien:
I don't think this is the right solution:

* `Type` is the *file type*.
* This is an RPC API, not a normal HTTP endpoint. For us: 200 means "it worked", 404 means "that API endpoint doesn't exist", 400 means "you passed a bad option", *500* means "some error occurred when processing a well-formed request".

Note: I'm not *happy* with how we're using HTTP status codes, that's just how the API currently works.

Stebalien:
(actually, let's move discussion back to the issue)

PR #5777, opened 2018-11-14
WIP: Add global option to specify the multibase encoding (client side)

### LINKED ISSUE #5233: Create ipfs.add flag for specifying base of produced CID

### LINKED ISSUE #5234: ipfs resolve should respect CID base

### LINKED ISSUE #5349: RFC: Handling of Alternative Multibases
As we switch to using CidV1 base32 encoding the question is how we should handle this displaying of CID in alternative multibases.

This issue is only about the output.  For the input will we already handle CIDv1 in any base supported by `go-multibase`.

Here are the alternatives, as I see it, and what I estimate it will take to do it.

### 1. Don't Preserve the multibase, Use Settings or Defaults to determine multibase

Accept CidV1 in any multibase, but ignore the base that is used.  Output the CIDs based on either a global `--cid-base` flag or a config value.

This means that the string a user passes in may not be the same string that is returned.  For example
```
$ ipfs pin add zb2rhfwrxLo7wm3ZT2ishCdimPKWdRCe3KuxHFQYMR8Kyj96x
pinned bafkreiekhi5qqy64ovh7jckh5ewnyt6lcvfzqfubv3l66cazo5dknu6dre recursively
```

This also means that https://github.com/ipfs/go-ipfs/issues/5234 (ipfs resolve should respect CID base) will not be implemented.

The most straightforward way to implement this is replacing every call of `cid.String()` with something like `cid.Encode(base)` (see https://github.com/ipfs/go-cid/pull/60) and make sure the base is available at the point when this method needs to be called.  The `WithValue` functionally on contexts can be helpful here.

As most CID's are converted to strings on the server side, rather than the client, doing a quick hack such as setting a global variable is not going to work.

Another way to implement this, at least when working with the API, is to modify our API to make CID's a special type and then set the multibase as part of the writer that formats the output for this user.  This means extra CPU cycles as the CID is converted from a binary represenation to a string twice.  If we go this route I recommend the API be hardcoded to use `base32` as is a more efficient encoding than our current default of `base58btc`.

### 2. Preserve the multibase.  Children of DAGs inherit multi-base of parent when known.

When a CID is decoded preserve the multibase used with the CID so when it is converted back to a string the same multibase will be used.  In addition associate the CID of any children of the DAG with the same multibase prefix.  When a CID does not have a multibase associated use the settings from the command line or config option.

`ipfs pin add` will then display the same base.  In addition commands like `ipfs ls` will use the same base to display directory entries as the hash used for the parent.

```
$ ipfs ls f017012205be9e545b52def2d18a922146c6cdec6bcbf4007b6ee7bec3341673bbf1d8c06
f01551220f6ba0f2491f2371dee0ea66e8812331b78f162697d2f41b5d55b667e5ae6bc67 3606  diff.go
f01551220e43f16413e6e1aa73eee0194d40363b7082d06faae8d6e94f18bd5f8cc08dfee 18299 object.go
f015512201985aa3462ff163c46de0f5e6634bde41f2bd537780ccbe5401aea1cccc49577 9057  patch.go
```

Implementing this will involve associating a multibase with CID's when it is known (see https://github.com/ipfs/go-cid/pull/66).  It will also involve setting the multibase of children CID's in the DAG code.  New CID's will also need to get the base set somewhere.

After thinking about this I think this will involve _less_ code change than #&#xfeff;1 as the base does not need to be set everywhere, although the changes may not be as straightforward.

I think this is the most user friendly approach.

### 3. Attempt to Preserve the multibase.

This solution does not store the multi-base with the CID but makes an attempt to output the same multibase on a ad-hoc bases.  For the current implementation of `ipfs resolve` it turned out to be easy.  `ipfs pin add` is more difficult.  It was in fact after looking at `ipfs pin add` that I decided that #&#xfeff;2 will be required to do it properly, especially when multiple CIDs are given on the command line.

Implementing this involves doing #&#xfeff;1 and in addition detecting the multibase of any CIDs used and setting and using the correct base for the output.  Things can get tricky when CID's with different mutlibases are used at once.  This will involve more code that either other solution.

https://github.com/ipfs/go-ipfs/pull/5289 started effort towards this solution (see the code for `ipfs resolve` and `ipfs ls`)

I do not like this solution as it will only work some of the time and prefer we go with #&#xfeff;1 or #&#xfeff;2 for more predictable output.

kevina:
Redo and simplification of #5464.

After giving it some careful thought I think doing the conversion on the server side for now makes the most sense (see #5789).  In particular this code relies on using global variables and a new type to do most of the work and avoid having to thread the cid base though to the point where they get displayed.  However, using this hack means that `?enc=text` http requests will also ignore the CID base.

The following command are implemented client side:
```
ipfs add
ipfs files ls
ipfs files stat
ipfs ls
ipfs pin add
ipfs pin ls
ipfs pin rm
ipfs pin verify
ipfs tar add
ipfs urlstore add
ipfs filestore dups
ipfs filestore ls
ipfs filestore verify
```

The following are implemented sever side
```
ipfs resolve
ipfs refs
ipfs refs local
```
`ipfs resolve` because it uses a path and `ipfs refs` do to the fact that a formatted string is sent via the API.

Todo (if we decide to go with this implementation):
  - [ ] Rebase and update to use newer go-cidutil dep.  Will also require updating ipfs/go-mfs#5

Depends on ipfs/go-cidutil#8 and ipfs/go-mfs#5.

Closes #5233. Closes #5234. Closes #5349.


alanshaw:
Are you planning on allowing people to specify CID base via the HTTP API?

It's not absolutely necessary, since the CID _can_ be upgraded/converted client side, but it's super convenient for anyone consuming the HTTP API directly for them to be able to specify it...or to flip it round, it's a bit onerous for every API consumer to have to implement an option for doing this client side for each CID they recieve.

I have implemented it for the listed HTTP endoints here https://github.com/ipfs/js-ipfs/pull/1552. I think it would be worthwile for go-ipfs to also have this option.

daviddias:
Agreed with @alanshaw, why not add this arg to the http-api as well. This it trickles nicely from cli -> http-api -> core.

Stebalien:
I believe the plan was to:

1. Do base-conversion client-side.
2. Switch to base32 by-default server-side for V1 CIDs by default (for performance).

Personally, I prefer keeping display logic client-side but I won't really object to a server-side option (given that we already have a text-based API).

alanshaw:
Understood. I don't think it needs to be complicated, for the JS side it was simply adding the additional parameter and reusing the logic that is already being used in the CLI for each route that we want to support it.

I still think there's a case for doing this. Maybe consumers might want to save bytes in their response by using a more compact encoding? I'm sure there's other reasons to allow it that we're not considering...

kevina:
@Stebalien @alanshaw sorry, I have been dragging my feet on this on because there is not any clean solution.

The client side solution ended up being not so nice because sometime we return formatted string with CID in them (see `ipfs refs` for example).  Maybe when we clean these up and start passing real CIDs via the API and we can switch to client side, but for now server side makes more sense.

My current thinking is to switch to server side.  Most of the time CIDs are encoded in the `commands` package.  When there are not, for example with the CoreAPI the plan is to reencode the CIDs after we get them back before calling Emit to avoid having to thread the parameter though outside of the `commands` package.  Eventually I think things that the CoreAPI should return CIDs and not strings and we should perhaps not reuse the same datastructures for the CoreAPI and what we use to send the result over the HTTP API.

@alanshaw note, that specifying `--cid-base` implies an automatic upgrade to from CIDv0 to CIDv1 except for select low-level commands where this will cause a problem.  I will let you know which commands after I redo this P.R. yet again.

alanshaw:
> Eventually I think things that the CoreAPI should return CIDs and not strings

Oh, agree! I opened this yesterday: https://github.com/ipfs/interface-ipfs-core/issues/394

alanshaw:
> The client side solution ended up being not so nice because sometime we return formatted string with CID in them (see ipfs refs for example)

Ah, JS doesn't have `ipfs refs` yet so we'll cross that bridge when we get there ;)

> My current thinking is to switch to server side

In https://github.com/ipfs/js-ipfs/pull/1552, for the CLI we're able to do everything client side for the commands we currently have, aside from `resolve` (which we've already mentioned). Adding `?cid-base` to the HTTP API obviously needed a server side conversion.

> note, that specifying `--cid-base` implies an automatic upgrade to from CIDv0 to CIDv1

Yes, thanks 😄, that's [taken into account](https://github.com/ipfs/js-ipfs/pull/1552/files#diff-c0fa2e49365c2774727eea3febcea3f5) in the PR.

> select low-level commands where this will cause a problem. I will let you know which commands after I redo this P.R. yet again

Yes, interested to know which and why - I've listed the commands I've implemented it for in https://github.com/ipfs/js-ipfs/pull/1552. I'm not aware of any issues...

kevina:
@alanshaw I am confused, using the `--cid-base` parameter does not pass the option to the server, but you support the `cid-base` end points anyway?  Does the js implementation not make a distinction between the client and the server like `go-ipfs` does?

(I am not familiar with how command line requests get translated to API requests but I am assuming `--cid-base=base32` will translate to `?cid-base=` query option.  My apologies if I am mistaken.)

alanshaw:
> @alanshaw I am confused, using the `--cid-base` parameter does not pass the option to the server

We choose the options from the command line that we want to pass to the server (core).

> but you support the `cid-base` end points anyway?

The CLI interacts with a IPFS core _interface_, which might be an actual instance of js-ipfs core or it might be a js-ipfs-api, depending on whether there's a daemon running or not.

So we _could_ pass this option into the core interface to take advantage of the `?cid-base` query param but that would push more logic for encoding CIDs into core, which is [something I want to get away from](https://github.com/ipfs/interface-ipfs-core/issues/394) - core should be returning CID instances.

So, the `?cid-base` query param is used where we actually have to do the conversion server side (for a path) and just _available_ on other endpoints for any other consumer of the HTTP API.

kevina:
Closing this because the general consensus is we are going with a (mostly) server side approach.

kevina:
(please don't delete a branch and allow it to be archived so that we have the code in case we ever need to refer back to it)

PR #5782, opened 2018-11-19
Switch to using request.Context()

whyrusleeping:
Was looking through the http handler code. When we figured out this CloseNotify stuff initially, I remember being super excited. That was quite some time ago, things have gotten better since then. The request.Context() now will cancel when the client leaves, or when the http server is shutdown, which is exactly what we want here. 

This approach also frees us from having to run another extra goroutine. Yay.

License: MIT


Stebalien:
>  Deprecated: the CloseNotifier interface predates Go's context package. New code should use Request.Context instead.

PR #5785, opened 2018-11-19
autorelay

vyzo:
Updates the node construction to automatically construct a routed host and enable autorelay.

Note that instances of the [autonat service](https://github.com/libp2p/go-libp2p-autonat-svc) are required to be (passively) discovered in the network for the autorealy functionality to test for NAT presence, before advertising relay addresses. 
This needs to run in the boostrappers and a plus if it is available in other public hosts, so an option to enable it is required.

Depends on 
- [X] https://github.com/ipfs/go-ipfs-config/pull/21
- [X] https://github.com/libp2p/go-libp2p/pull/500
- [X] https://github.com/libp2p/go-libp2p/pull/495
- [X] libp2p bubble up
- ~Fix https://github.com/libp2p/go-libp2p/issues/467~

TBD:
- [X] resolve test failures
- [X] enable autorelay gating with config option
- [X] instantiate autonat service with config option.
- [X] use explicit autorelay libp2p option.


vyzo:
it's enabled by libp2p when relay is enabled and you pass the new `Routing` option, which also constructs a routed host.
Should I add an option so that users have to explicitly enable it (or disable it)?

vyzo:
Presumably we do want an option in ipfs to enable or disable it, instead of unconditionally enabling.
What should the default state be though? 
Perhaps we should start with it disabled as this is a new and experimental feature, and will also require a release cycle to be effective (because new addresses are propagated with identify push, which older nodes do not support).

vyzo:
options in https://github.com/ipfs/go-ipfs-config/pull/21

core/core.go
```diff
@@ -498,12 +507,17 @@ func (n *IpfsNode) startOnlineServicesWithHost(ctx context.Context, host p2phost
 		n.PubSub = service
 	}
 
-	// setup routing service
-	r, err := routingOption(ctx, host, n.Repo.Datastore(), n.RecordValidator)
-	if err != nil {
-		return err
```
Stebalien: 1. This comment needs to explain what tests need this and why.
2. I'm quite positive there's a non-hacky way to do this.

Stebalien:
Whatever happened to fixing libp2p *immediately* after merging the autorelay change? I thought the entire point of that was to get some code to refactor.

vyzo:
The libp2p refactoring is still on, but I don't think it will change the user interface -- so we might as well get the feature out first!

vyzo:
added autorelay gating with config option.

vyzo:
instantiated the autonat service as well.

vyzo:
@Stebalien this is complete feature-wise, take another look?

vyzo:
rebased on master.

Stebalien:
* What is the difference between a RelayHost and an AutoRelayHost?
* Why does passing the *routing* option to libp2p enable auto relay? They seem *totally* unrelated (well, the former seems like a dependency of the latter).

This patch should be able to avoid the routing option dance by *always* using the libp2p option. We'd need to modify the mock host constructor (a cleaner solution would be to add a Host (constructor) option to libp2p itself but that may be overkill given that we're planning on changing this all up).

vyzo:
> What is the difference between a RelayHost and an AutoRelayHost?

RelayHost is an actual relay, while AutoRelayHost is a host that connects to relays and advertises relay addrs (when NATed).

> Why does passing the routing option to libp2p enable auto relay? They seem totally unrelated (well, the former seems like a dependency of the latter).

Well, the routing is necessary for auto relay, but there is no other explicit option. Perhaps we should add one.

> We'd need to modify the mock host constructor 

I guess we can do that. We'll need the option for explicit autorelay though, otherwise it will be unconditionally enabled.

vyzo:
Added explicit option in https://github.com/libp2p/go-libp2p/pull/500; we'll have to bubble up, and also merge   https://github.com/libp2p/go-libp2p/pull/495

djdv:
Looks like https://github.com/libp2p/go-libp2p/pull/500 and https://github.com/libp2p/go-libp2p/pull/495 have been merged.
I'll look out for patches to be made here. 
(Either before or after the formal libp2p `gx publish`, just checking to see if anything else is/needs changing.)

djdv:
Looks like https://github.com/libp2p/go-libp2p/pull/500 and https://github.com/libp2p/go-libp2p/pull/495 have been merged.
I'll look out for patches to be made here. 
(Either before or after the formal libp2p `gx publish`, just checking to see if anything else is/needs changing.)

vyzo:
rebased on master and explicitly enabled autorelay.

vyzo:
@Stebalien using the option to explicitly enable autorelay; only sticking point is the mock host construction not recognizing the `Routing` libp2p constructor option, which requires duplicate code.

core/core.go
```diff
@@ -261,12 +263,26 @@ func (n *IpfsNode) startOnlineServices(ctx context.Context, routingOption Routin
...
+		libp2pOpts = append(libp2pOpts, libp2p.EnableAutoRelay())
+	}
+
 	peerhost, err := hostOption(ctx, n.Identity, n.Peerstore, libp2pOpts...)
 
 	if err != nil {
 		return err
 	}
 
+	n.PeerHost = peerhost
```
Stebalien: If we're going to move this up here, we should remove the host parameter from `startOnlineServicesWithHost`.

vyzo: ok, will do.

vyzo:
Added documentation.

core/core.go
vyzo: done.

vyzo:
Yeah, testing this end-to-end is hard in ci -- too many components and delays involved; I can't think of a good way to make a sharness test for this. But I am looking forward to testing in the wild!

Stebalien:
Actually, we can totally test this (I think). We can't simulate NATs but we can, carefully, configure our transports:

* Peer A: Listens on QUIC only (but can still dial with TCP).
* Peer B: Has QUIC disabled, has AutoNATService enabled.
* Peer C: Has QUIC enabled, has RelayHop enabled.
* Peer D: Has QUIC disabled.

1. Start all peers.
2. Connect peer A to peer B and peer C.
3. Try to connect peer D to peer A, keep *retrying* every 3 seconds until it works (may take a while).
4. Check `ipfs id` on peer A to see if it's announcing a relay address.

Think that will work?

vyzo:
there are still some obsene delays involved; there is about a 90s delay until the autonat detection boots and fires, which we don't have a way to configure in sharness.
we could add some parameters just for testing though, but it's ugly.

vyzo:
ok, let me cook up a ridiculously slow sharness test then -- how do i mark as expensive?

Stebalien:
`test_expect_success EXPENSIVE "some description" ' ... test ... '`

(take a look at the add-and-cat tests).

vyzo:
So I've cooked up a test as you suggested, but unfortunately it can't work: autorelay only advertises public addresses and there are no public addresses in my NATed laptop :(

magik6k:
This is <s>probably</s> definetly a really bad idea, but - are there any solutions that would allow us to setup fake userspace network and hook ipfs it into using something like `LD_PRELOAD`?

Stebalien:
> are there any solutions that would allow us to setup fake userspace network and hook ipfs it into using something like LD_PRELOAD?

Maybe but probably not given that go tends to make raw syscalls. We could *also* use a syscall filter.

However... we *could* test this from go with a plugin. That is, we could drop in a special plugin that modifies go-multiaddr-net's Unroutable/Private network variables on load. Thoughts? Is that even possible?

Stebalien:
Status: Blocked on at least demonstrating that this works (even if we can't write a test that runs on CI).

vyzo:
I have confirmed that autorelay works with manual testing.
I started an AutoNAT/RelayHop node in a vps, and tested using a double-NATed laptop node.
I connected the laptop node to the vps node, and after a couple of minutes it started advertising relay addrs.

Transcript:
```
$ ipfs id

{
        "ID": "QmRozfTYbjdktgjkC9XNjQNKHz4kui5sfNRzkaGqHmRouE",
        "PublicKey": "CAASpgIwggEiMA0GCSqGSIb3DQEBAQUAA4IBDwAwggEKAoIBAQDSG2Zlqp
oEyWa498W3NzaiOz642nwkP9jgnggpHWrJ3lL6cv2OzM+n9EREshG4TIHuxjY8unvBby8weDzWPYXO9o
YSR052VKSNzZgsMZ1cjuL6YrNgf9t2l9JeHvHZlEdodYKFwwcWgrAVRjAsVMOpo3zylrz19lYk7b5I2O
QgaX4dSNpjxUuZH1M97IokegHnIKxCqdFdQ03WotlQ08ppzQ/kG0UYiZvLKTxbrFw8wR915oBH9ZQf0t
RVaB+5kbJiVG/UmHyAyCmqpxrhW73UrldU9kIVmFxHFoeqP50stiWVtu/ROmiy040EL/krjv88ftQVvMLtBTUUsnyt2wX5AgMBAAE=",
        "Addresses": [
                "/ip4/127.0.0.1/tcp/4001/ipfs/QmRozfTYbjdktgjkC9XNjQNKHz4kui5sfNRzkaGqHmRouE",
                "/ip4/10.0.1.30/tcp/4001/ipfs/QmRozfTYbjdktgjkC9XNjQNKHz4kui5sfNRzkaGqHmRouE",
                "/ip6/::1/tcp/4001/ipfs/QmRozfTYbjdktgjkC9XNjQNKHz4kui5sfNRzkaGqHmRouE",
                "/ip4/79.167.218.87/tcp/2647/ipfs/QmRozfTYbjdktgjkC9XNjQNKHz4kui5sfNRzkaGqHmRouE"
        ],
        "AgentVersion": "go-ipfs/0.4.19-dev/dad0fa54a",
        "ProtocolVersion": "ipfs/0.1.0"
}

# connect to the relay/autonat node:
$ ipfs swarm connect /ip4/172.104.9.124/tcp/4001/ipfs/QmchgNzyUFyf2wpfDMmpGxMKHA3PkC1f3H2wUgbs21vXoh

# after a couple of minutes relay addrs are advertised as expected:
$ ipfs id

{
        "ID": "QmRozfTYbjdktgjkC9XNjQNKHz4kui5sfNRzkaGqHmRouE",
        "PublicKey": "CAASpgIwggEiMA0GCSqGSIb3DQEBAQUAA4IBDwAwggEKAoIBAQDSG2ZlqpoEyWa498W3NzaiOz642nwkP9jgnggpHWrJ3lL6cv2OzM+n9EREshG4TIHuxjY8unvBby8weDzWPYXO9oYSR052VKSNzZgsMZ1cjuL6YrNgf9t2l9JeHvHZlEdodYKFwwcWgrAVRjAsVMOpo3zylrz19lYk7b5I2OQgaX4dSNpjxUuZH1M97IokegHnIKxCqdFdQ03WotlQ08ppzQ/kG0UYiZvLKTxbrFw8wR915oBH9ZQf0tRVaB+5kbJiVG/UmHyAyCmqpxrhW73UrldU9kIVmFxHFoeqP50stiWVtu/ROmiy040EL/krjv88ftQVvMLtBTUUsnyt2wX5AgMBAAE=",
        "Addresses": [
                "/ip4/127.0.0.1/tcp/4001/ipfs/QmRozfTYbjdktgjkC9XNjQNKHz4kui5sfNRzkaGqHmRouE",
                "/ip4/10.0.1.30/tcp/4001/ipfs/QmRozfTYbjdktgjkC9XNjQNKHz4kui5sfNRzkaGqHmRouE",
                "/ip6/::1/tcp/4001/ipfs/QmRozfTYbjdktgjkC9XNjQNKHz4kui5sfNRzkaGqHmRouE",
                "/ip4/79.167.218.87/ipfs/QmRozfTYbjdktgjkC9XNjQNKHz4kui5sfNRzkaGqHmRouE",
                "/ip6/2600:3c03::f03c:91ff:fe87:a3b4/tcp/4001/ipfs/QmchgNzyUFyf2wpfDMmpGxMKHA3PkC1f3H2wUgbs21vXoh/p2p-circuit/ipfs/QmRozfTYbjdktgjkC9XNjQNKHz4kui5sfNRzkaGqHmRouE",
                "/ip4/172.104.9.124/tcp/4001/ipfs/QmchgNzyUFyf2wpfDMmpGxMKHA3PkC1f3H2wUgbs21vXoh/p2p-circuit/ipfs/QmRozfTYbjdktgjkC9XNjQNKHz4kui5sfNRzkaGqHmRouE"
        ],
        "AgentVersion": "go-ipfs/0.4.19-dev/dad0fa54a",
        "ProtocolVersion": "ipfs/0.1.0"
}
```

vyzo:
Steps to reproduce:
1. Create a public node with AutoNAT/RelayHop Enabled:
```
ipfs config --json Swarm.EnableRelayHop true
ipfs config --json Swarm.EnableAutoNATService true
ipfs config --json Swarm.EnableAutoRelay true
```

2. Start a node behind a bad NAT, and connect to your relay using `ipfs warm connect ...`

3. Wait a couple of minutes, and check the output of `ipfs id`; it should change the public addresses to the naked ip plus the public relay addresses.
```

PR #5787, opened 2018-11-21
cmds: add online helper

overbool:
License: MIT


overbool:
@magik6k Actually, we can check whether the node is online by calling `nd.OnlineMode()`. However, in `core/commands/external.go` we need get the node mode before `nd, err := cmdenv.GetNode(env)`, so maybe we should add this helper function.

Stebalien:
I believe this can be replaced with https://github.com/ipfs/go-ipfs/pull/6050 (which simply deletes all this code).

PR #5789, opened 2018-11-22
Add global option to specify the multibase encoding (server side)

### LINKED ISSUE #5233: Create ipfs.add flag for specifying base of produced CID

### LINKED ISSUE #5234: ipfs resolve should respect CID base

### LINKED ISSUE #5349: RFC: Handling of Alternative Multibases

kevina:
Redo of #5464 but this time doing it (mostly) on the server side.

This may not be the most elegant implementation, but for now I think it is the most practical, espacally if we also want to support `?cid-base` end points for API requests.  We can clean this up later to make things more elegant.

It does all the work in the `core/commands` package and does not leak into other API's except for the `ipfs filestore` command where we need to pass an encoder function to `FormatLong` method.

See #5777 for an alternative client side implementation.

This adds support for `--cid-base` and `--output-cidv1` option to the command line and HTTP API for the following commands:
```
ipfs add
ipfs files ls
ipfs files stat
ipfs ls
ipfs pin add
ipfs pin ls
ipfs pin rm
ipfs pin verify
ipfs refs
ipfs refs local
ipfs resolve
ipfs tar add
ipfs urlstore add
ipfs filestore dups
ipfs object get
ipfs object put
ipfs object links
```

Due to the fact that some commands return the CID itself and not a string, some command support the base conversion in text output only until we upgrade to something like `refmt` which will give us more control over the JSON output:
```
ipfs filestore ls
ipfs filestore verify
ipfs bitswap stat
ipfs bitswap wantlist
ipfs dag put
ipfs dag resolve
```

`ipfs dag get` currently does not support `--cid-base` due to the way it is implemented. This is considered a bug.

For some commands the original path string is preserved without any additional processing and is thus unaffected by `--cid-base`:
```
ipfs ls (only the original path displayed when multiple arguments are present, not the actual output) 
ipfs pin update
```

In addition all commands under `bitswap`, `object`, and `dag` do not, by default, auto-upgrade CIDv0 to CIDv1 as these commands are considered low-level and the conversion could cause problems (for example a dag will not survive a `ipfs object get' /  'ipfs object put` round trip.  This can be changes by using `--output-cidv1=true`.

Todo:
- [x] Remove `cidenc.Interface` for right now it doesn't serve a useful purpose and it may make more sense to define it to only have an `Encode` method (and no `Recode` method) so the existing `cid.Encoder` also implements the interface.
- [x] Abandon the `apicid` concept all together as that will be a holdover from the client side implementation and move `apicid.Cid` to `cidenc.Cid`.
 - [x] Rebase to clean up commits and avoid depending on locally gx published versions of go-cidutil.
 - [x] Finalize `--output-cidv1`|`--force-multibase-cids|--force-cid-base` option name

Depends on https://github.com/ipfs/go-cidutil/pull/10.

Closes #5233. Closes #5234. Closes #5349.

### DIFF
--- a/core/commands/add.go
+++ b/core/commands/add.go
@@ -174,6 +174,11 @@ You can now check what blocks have been created by:
 			return fmt.Errorf("unrecognized hash function: %s", strings.ToLower(hashFunStr))
 		}
 
+		enc, err := cmdenv.GetCidEncoder(req)
+		if err != nil {
+			return err
+		}
+
 		events := make(chan interface{}, adderOutChanSize)
 
 		opts := []options.UnixfsAddOption{
@@ -226,7 +231,7 @@ You can now check what blocks have been created by:
 
 			h := ""
 			if output.Path != nil {
-				h = output.Path.Cid().String()
+				h = enc.Encode(output.Path.Cid())
 			}
 
 			res.Emit(&AddEvent{
--- a/core/commands/bitswap.go
+++ b/core/commands/bitswap.go
@@ -74,12 +74,15 @@ Print out all blocks currently on the bitswap wantlist for the local peer.`,
 	},
 	Encoders: cmds.EncoderMap{
 		cmds.Text: cmds.MakeTypedEncoder(func(req *cmds.Request, w io.Writer, out *KeyList) error {
+			enc, err := cmdenv.GetLowLevelCidEncoder(req)
+			if err != nil {
+				return err
+			}
 			// sort the keys first
 			cidutil.Sort(out.Keys)
 			for _, key := range out.Keys {
-				fmt.Fprintln(w, key)
+				fmt.Fprintln(w, enc.Encode(key))
 			}
-
 			return nil
 		}),
 	},
@@ -115,6 +118,10 @@ var bitswapStatCmd = &cmds.Command{
 	},
 	Encoders: cmds.EncoderMap{
 		cmds.Text: cmds.MakeTypedEncoder(func(req *cmds.Request, w io.Writer, s *bitswap.Stat) error {
+			enc, err := cmdenv.GetLowLevelCidEncoder(req)
+			if err != nil {
+				return err
+			}
 			fmt.Fprintln(w, "bitswap status")
 			fmt.Fprintf(w, "\tprovides buffer: %d / %d\n", s.ProvideBufLen, bitswap.HasBlockBufferSize)
 			fmt.Fprintf(w, "\tblocks received: %d\n", s.BlocksReceived)
@@ -125,7 +132,7 @@ var bitswapStatCmd = &cmds.Command{
 			fmt.Fprintf(w, "\tdup data received: %s\n", humanize.Bytes(s.DupDataReceived))
 			fmt.Fprintf(w, "\twantlist [%d keys]\n", len(s.Wantlist))
 			for _, k := range s.Wantlist {
-				fmt.Fprintf(w, "\t\t%s\n", k.String())
+				fmt.Fprintf(w, "\t\t%s\n", enc.Encode(k))
 			}
 			fmt.Fprintf(w, "\tpartners [%d]\n", len(s.Peers))
 			for _, p := range s.Peers {
--- a/core/commands/bitswap.go
+++ b/core/commands/bitswap.go
@@ -12,7 +12,7 @@ import (
 	peer "gx/ipfs/QmY5Grm8pJdiSSVsYxx4uNRgweY72EmYwuSDbRnbFok3iY/go-libp2p-peer"
 	bitswap "gx/ipfs/QmYoGLuLwTUv1SYBmsw1EVNC9MyLVUxwxzXYtKgAGHyEfw/go-bitswap"
 	decision "gx/ipfs/QmYoGLuLwTUv1SYBmsw1EVNC9MyLVUxwxzXYtKgAGHyEfw/go-bitswap/decision"
-	cidutil "gx/ipfs/QmbfKu17LbMWyGUxHEUns9Wf5Dkm8PT6be4uPhTkk4YvaV/go-cidutil"
+	cidutil "gx/ipfs/QmdPQx9fvN5ExVwMhRmh7YpCQJzJrFhd1AjVBwJmRMFJeX/go-cidutil"
 	cmdkit "gx/ipfs/Qmde5VP1qUkyQXKCfmEUA7bP64V2HAptbJ7phuPp7jXWwg/go-ipfs-cmdkit"
 )
 
--- a/core/commands/cmdenv/cidbase.go
+++ b/core/commands/cmdenv/cidbase.go
@@ -1,9 +1,8 @@
 package cmdenv
 
 import (
-	"errors"
+	"strings"
 
-	path "gx/ipfs/QmNYPETsdAu2uQ1k9q9S1jYEGURaLHV6cbYRSVFVRftpF8/go-path"
 	cmds "gx/ipfs/QmWGm4AbZEbnmdgVTza52MSNpEmBdFVqzmAysRbjrRyGbH/go-ipfs-cmds"
 	cidenc "gx/ipfs/QmdPQx9fvN5ExVwMhRmh7YpCQJzJrFhd1AjVBwJmRMFJeX/go-cidutil/cidenc"
 	cmdkit "gx/ipfs/Qmde5VP1qUkyQXKCfmEUA7bP64V2HAptbJ7phuPp7jXWwg/go-ipfs-cmdkit"
@@ -64,10 +63,7 @@ func CidBaseDefined(req *cmds.Request) bool {
 // the base encoder is returned.  If you don't care about the error
 // condition, it is safe to ignore the error returned.
 func CidEncoderFromPath(enc cidenc.Encoder, p string) (cidenc.Encoder, error) {
-	v, err := extractCidString(p)
-	if err != nil {
-		return enc, err
-	}
+	v := extractCidString(p)
 	if cidVer(v) == 0 {
 		return cidenc.Encoder{Base: enc.Base, Upgrade: false}, nil
 	}
@@ -78,16 +74,12 @@ func CidEncoderFromPath(enc cidenc.Encoder, p string) (cidenc.Encoder, error) {
 	return cidenc.Encoder{Base: e, Upgrade: true}, nil
 }
 
-func extractCidString(str string) (string, error) {
-	p, err := path.ParsePath(str)
-	if err != nil {
-		return "", err
+func extractCidString(str string) string {
+	parts := strings.Split(str, "/")
+	if len(parts) > 2 && (parts[1] == "ipfs" || parts[1] == "ipld") {
+		return parts[2]
 	}
-	segs := p.Segments()
-	if segs[0] == "ipfs" || segs[0] == "ipld" {
-		return segs[1], nil
-	}
-	return "", errors.New("no CID found")
+	return str
 }
 
 func cidVer(v string) int {
--- a/core/commands/cmdenv/cidbase.go
+++ b/core/commands/cmdenv/cidbase.go
@@ -1,8 +1,10 @@
 package cmdenv
 
 import (
+	"fmt"
 	"strings"
 
+	cid "gx/ipfs/QmR8BauakNcBa3RbE4nbQu76PDiJgoQgz8AJdhJuiU4TAw/go-cid"
 	cmds "gx/ipfs/QmWGm4AbZEbnmdgVTza52MSNpEmBdFVqzmAysRbjrRyGbH/go-ipfs-cmds"
 	cidenc "gx/ipfs/QmdPQx9fvN5ExVwMhRmh7YpCQJzJrFhd1AjVBwJmRMFJeX/go-cidutil/cidenc"
 	cmdkit "gx/ipfs/Qmde5VP1qUkyQXKCfmEUA7bP64V2HAptbJ7phuPp7jXWwg/go-ipfs-cmdkit"
@@ -59,32 +61,47 @@ func CidBaseDefined(req *cmds.Request) bool {
 // CidEncoderFromPath creates a new encoder that is influenced from
 // the encoded Cid in a Path.  For CidV0 the multibase from the base
 // encoder is used and automatic upgrades are disabled.  For CidV1 the
-// multibase from the CID is used and upgrades are eneabled.  On error
-// the base encoder is returned.  If you don't care about the error
-// condition, it is safe to ignore the error returned.
-func CidEncoderFromPath(enc cidenc.Encoder, p string) (cidenc.Encoder, error) {
-	v := extractCidString(p)
-	if cidVer(v) == 0 {
-		return cidenc.Encoder{Base: enc.Base, Upgrade: false}, nil
+// multibase from the CID is used and upgrades are enabled.
+//
+// This logic is intentionally fuzzy and will match anything of the form
+// `CidLike`, `CidLike/...`, or `/namespace/CidLike/...`.
+//
+// For example:
+//
+// * Qm...
+// * Qm.../...
+// * /ipfs/Qm...
+// * /ipns/bafybeiahnxfi7fpmr5wtxs2imx4abnyn7fdxeiox7xxjem6zuiioqkh6zi/...
+// * /bzz/bafybeiahnxfi7fpmr5wtxs2imx4abnyn7fdxeiox7xxjem6zuiioqkh6zi/...
+func CidEncoderFromPath(p string) (cidenc.Encoder, error) {
+	components := strings.SplitN(p, "/", 4)
+
+	var maybeCid string
+	if components[0] != "" {
+		// No leading slash, first component is likely CID-like.
+		maybeCid = components[0]
+	} else if len(components) < 3 {
+		// Not enough components to include a CID.
+		return cidenc.Encoder{}, fmt.Errorf("no cid in path: %s", p)
+	} else {
+		maybeCid = components[2]
 	}
-	e, err := mbase.NewEncoder(mbase.Encoding(v[0]))
+	c, err := cid.Decode(maybeCid)
 	if err != nil {
-		return enc, err
+		// Ok, not a CID-like thing. Keep the current encoder.
+		return cidenc.Encoder{}, fmt.Errorf("no cid in path: %s", p)
 	}
-	return cidenc.Encoder{Base: e, Upgrade: true}, nil
-}
+	if c.Version() == 0 {
+		// Version 0, use the base58 non-upgrading encoder.
+		return cidenc.Default(), nil
+	}
 
-func extractCidString(str string) string {
-	parts := strings.Split(str, "/")
-	if len(parts) > 2 && (parts[1] == "ipfs" || parts[1] == "ipld") {
-		return parts[2]
+	// Version 1+, extract multibase encoding.
+	encoding, _, err := mbase.Decode(maybeCid)
+	if err != nil {
+		// This should be impossible, we've already decoded the cid.
+		panic(fmt.Sprintf("BUG: failed to get multibase decoder for CID %s", maybeCid))
 	}
-	return str
-}
 
-func cidVer(v string) int {
-	if len(v) == 46 && v[:2] == "Qm" {
-		return 0
-	}
-	return 1
+	return cidenc.Encoder{Base: mbase.MustNewEncoder(encoding), Upgrade: true}, nil
 }
--- a/core/commands/cmdenv/cidbase.go
+++ b/core/commands/cmdenv/cidbase.go
@@ -10,7 +10,7 @@ import (
 )
 
 var OptionCidBase = cmdkit.StringOption("cid-base", "Multibase encoding used for version 1 CIDs in output.")
-var OptionForceCidBase = cmdkit.BoolOption("force-cid-base", "Force multibase prefix by upgrading CIDv0 to CIDv1.")
+var OptionUpgradeCidV0InOutput = cmdkit.BoolOption("upgrade-cidv0-in-output", "Upgrade version 0 to version 1 CIDs in output.")
 
 // GetCidEncoder processes the `cid-base` and `output-cidv1` options and
 // returns a encoder to use based on those parameters.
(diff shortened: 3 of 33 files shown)


kevina:
@eingenito this p.r. is a change in direction from what me and @Stebalien agree on, the other p.r. #5777 is closer to the solution we discussed.  

See he comments at the end of #5349 for more context.

I would like to get this p.r. pretty much as is and then if possible refine it later.  I specially designed it to be an non-invasive as possible so that we will not be locked into a particular implementation.

We do need to agree on is the choice of flags `--cid-base` and `--output-cidv1` and the fact that we want these command to effect the JSON output, however.

core/commands/cmdenv/cidbase.go
```diff
@@ -0,0 +1,86 @@
...
+func EnableCidBaseGlobal(req *cmds.Request, env cmds.Environment) error {
+	enc, err := ProcCidBase(req)
+	if err != nil {
+		return err
+	}
+	cidenc.Default = enc
+	return nil
+}
+
+// FromPath creates a new encoder that is influenced from the encoded
```
magik6k:
```suggestion
// CidEncoderFromPath creates a new encoder that is influenced from the encoded
```

core/commands/pin.go
```diff
@@ -455,14 +480,14 @@ type RefKeyList struct {
 	Keys map[string]RefKeyObject
 }
 
-func pinLsKeys(ctx context.Context, args []string, typeStr string, n *core.IpfsNode, api iface.CoreAPI) (map[string]RefKeyObject, error) {
+func pinLsKeys(ctx context.Context, args []string, typeStr string, n *core.IpfsNode, api iface.CoreAPI) (map[cid.Cid]RefKeyObject, error) {
```
magik6k: This logic is unfortunately still duplicated in `core/coreapi/pin.go`, do you mind updating it there too? (this will get fixed once pin commands are rewritten to use CoreAPI)

core/commands/cmdenv/cidbase.go
```diff
@@ -0,0 +1,86 @@
...
+import (
+	path "gx/ipfs/QmQtg7N4XjAk2ZYpBjjv8B6gQprsRekabHBCnF6i46JYKJ/go-path"
+	cmds "gx/ipfs/Qma6uuSyjkecGhMFFLfzyJDPyoDtNJSHJNweDccZhaWkgU/go-ipfs-cmds"
+	cidenc "gx/ipfs/QmckgkstbdXagMTQ4e1DW2SzxGcjjudbqEvA5H2Rb7uvAT/go-cidutil/cidenc"
+	cmdkit "gx/ipfs/Qmde5VP1qUkyQXKCfmEUA7bP64V2HAptbJ7phuPp7jXWwg/go-ipfs-cmdkit"
+	mbase "gx/ipfs/QmekxXDhCxCJRNuzmHreuaT3BsuJcsjcXWNrtV9C8DRHtd/go-multibase"
+)
+
+var OptionCidBase = cmdkit.StringOption("cid-base", "Multi-base encoding used for version 1 CIDs in output.")
+var OptionOutputCidV1 = cmdkit.BoolOption("output-cidv1", "Upgrade CID version 0 to version 1 in output.")
```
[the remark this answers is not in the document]
kevina: That would be very misleading.  This option only effects the Cid version that is outputted and nothing else, in particular in `ipfs add` it will not change the Hash.

core/commands/cmdenv/cidbase.go
```diff
@@ -0,0 +1,86 @@
...
+
+import (
+	path "gx/ipfs/QmQtg7N4XjAk2ZYpBjjv8B6gQprsRekabHBCnF6i46JYKJ/go-path"
+	cmds "gx/ipfs/Qma6uuSyjkecGhMFFLfzyJDPyoDtNJSHJNweDccZhaWkgU/go-ipfs-cmds"
+	cidenc "gx/ipfs/QmckgkstbdXagMTQ4e1DW2SzxGcjjudbqEvA5H2Rb7uvAT/go-cidutil/cidenc"
+	cmdkit "gx/ipfs/Qmde5VP1qUkyQXKCfmEUA7bP64V2HAptbJ7phuPp7jXWwg/go-ipfs-cmdkit"
+	mbase "gx/ipfs/QmekxXDhCxCJRNuzmHreuaT3BsuJcsjcXWNrtV9C8DRHtd/go-multibase"
+)
+
+var OptionCidBase = cmdkit.StringOption("cid-base", "Multi-base encoding used for version 1 CIDs in output.")
```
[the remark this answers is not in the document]
kevina: Okay will fix.

core/commands/pin.go
kevina: I can do that in a separate P.R.

core/commands/cmdenv/cidbase.go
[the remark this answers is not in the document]
kevina: The reason for this option is because CIDv0 are not suppose to take a multihash so if use use `--cid-base=base32` CidV0 hashes will still output in base58btc which is not very helpful. So we decided that we (me @Stebalien daviddias whyrusleeping and few others) would automatically upgrade CIDv0 to CIDv1, but in the output only.

If `--cid-base` is given this is implied.  This was agreed upon between me and @Stebalien see https://github.com/ipfs/go-ipfs/pull/5289#issuecomment-415902770.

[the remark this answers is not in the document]
magik6k: My problem with this option is that it only allows to upgrade from 0 to 1, which can be a bit limiting.  
If this was an int (and still only affect output), this would also allow:
* Users to say they want to stick with CIDv0 after we switch to CIDv1 by default
* Us to reuse this option if we need it for another migration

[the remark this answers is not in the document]
kevina: I can make this `--output-cid-version` but that will make this option more tedious to use and I don't see any benefit from it.  I really consider this option somewhat of a hack to handle the special case that `CIDv0` can not take a multibase prefix.

> Users to say they want to stick with CIDv0 after we switch to CIDv1 by default

`--output-cidv1=false` will do this

> Us to reuse this option if we need it for another migration

While theoretically true I am not sure if we will want to reuse this option.  This option is only present to deal with the special case that CIDv0 can not take a multibase prefix.  I doubt we will want to reuse it again.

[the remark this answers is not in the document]
kevina: I admit that is we ever have a need to create a new CID version this option `--output-cidv1` would be misleading as it would imply that CIDv2 are downgraded to CIDv1 which was not the intention.  I originally used `--upgrade-cidv0` but changed it as as we might need that option to actually upgrade Cid's for example when pinning.  I am open to other names, but have a strong preference to something reasonable short and not something like  `--upgrade-cidv0-in-output`.

core/commands/pin.go
kevina: Done see: https://github.com/ipfs/go-ipfs/pull/5809

filestore/util.go
```diff
@@ -269,14 +270,14 @@ func mkListRes(c cid.Cid, d *pb.DataObj, err error) *ListRes {
 		return &ListRes{
 			Status:   status,
 			ErrorMsg: errorMsg,
-			Key:      c,
+			Key:      apicid.Cid{Cid: c},
```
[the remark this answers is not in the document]
kevina: Yes I thought about that.  By defining it as a struct we can inherit (in the go-sense) the Cid methods so a `apicid.Cid` can be used like a normal Cid in most places.

core/commands/cmdenv/cidbase.go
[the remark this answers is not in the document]
Stebalien: What about something like `--force-multibase-cids` (or something like that). In the help text, we can explain that it auto-upgrades cidv0 to cidv1 in output. Unfortunately, this flag *doesn't* say what it actually does so it's not much better.

This is probably a distant non-issue but `--output-cidv1` could be a *bit* confusing if we ever get a cidv2.

(I don't have any strong opinions on this API)

filestore/util.go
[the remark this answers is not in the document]
Stebalien: Ah, I see. Yeah, that sounds reasonable.

kevina:
@Stebalien see https://github.com/ipfs/go-ipfs/issues/5349#issuecomment-445100774.

magik6k:
> I'd rather make the CoreAPI as clean as possible for libraries. That is, a go application using the CoreAPI as a library should also be able to use a local IPLD DAGService etc. without having to use multiple CID types.

If we only apply this to exposed string hashes then yeah, this is fine. I don't think there is anything in CoreAPI that returns those, so :+1:.

kevina:
This is almost ready to go.  I did some additional simplifications and updated the top-level description.

I will rebase to clean up my commits once near the end to let people who already started to review this a change to see what's new.

kevina:
@Stebalien (and others) this is ready for review.

core/commands/add.go
```diff
@@ -212,9 +218,13 @@ You can now check what blocks have been created by:
 			_, err = api.Unixfs().Add(req.Context, req.Files, opts...)
 		}()
 
-		err = res.Emit(events)
-		if err != nil {
-			return err
+		for event := range events {
+			output := event.(*coreiface.AddEvent)
```
magik6k: CoreAPI probably shouldn't expose string hashes, pushed a fix in https://github.com/ipfs/go-ipfs/pull/5854

core/commands/bitswap.go
```diff
@@ -74,12 +74,15 @@ Print out all blocks currently on the bitswap wantlist for the local peer.`,
...
 		cmds.Text: cmds.MakeTypedEncoder(func(req *cmds.Request, w io.Writer, out *KeyList) error {
+			enc, err := cmdenv.GetLowLevelCidEncoder(req)
+			if err != nil {
+				return err
+			}
 			// sort the keys first
 			cidutil.Sort(out.Keys)
 			for _, key := range out.Keys {
-				fmt.Fprintln(w, key)
+				fmt.Fprintln(w, enc.Encode(key))
```
magik6k: Not a strong opinion, but maybe we should do `cmdenv.Encode(req, cid)`?

core/commands/cmdenv/cidbase.go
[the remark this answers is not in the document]
magik6k: :+1: for `--force-multibase-cids`

core/commands/bitswap.go
kevina: I don't see the benefit and that will also be ineffecnt as it will require reparsing the command line arguments each time.

kevina: Not to mention that the function would need to return an error, which will make it less convenient to use then the method.

core/commands/add.go
Stebalien: Blocked on https://github.com/ipfs/go-ipfs/pull/5854.

kevina: @Stebalien I would ordinary preferred this the other way around, but since I said I was going to be OOO right now I'm fine with that, but could will get that in by the near year so this P.R. can move forward.

Stebalien: Fair enough. I didn't realize this was a mostly unrelated issue.

(I merged #5854 anyways but I've rebased this PR at the same time)

Stebalien:
The final remaining question is: https://github.com/ipfs/go-ipfs/pull/5789#discussion_r237573246.

core/commands/cmdenv/cidbase.go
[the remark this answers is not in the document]
kevina: I am still in favor of `--ouuput-cidv1` because it is clear and to the point, although I do admit it could be confusing in a dissent future if we ever use CIDv2.  As an alternative how about `--force-cid-base`, this is a little clearer because it makes it obvious that this option is related to the `cid-base` option and because not everone will know what a `multibase` is.  The only thing I don't like about it is is not 100% accurate as this option will also convert to cidv1 for `base58btc` even though it is not strictly needed.

Other alternatives that I though of and dismissed, `--force-multibase-output` (makes it clear that the option effects the output), `--force-multibase` (shorter).

@Stebalien @magik6k @ipfs/go-team thoughts?

Stebalien:
`--force-cid-base` seems reasonable.

kevina:
Note: for those jumping into issue now: See https://github.com/ipfs/go-ipfs/pull/5789#discussion_r237573246 for the discussion on the `--output-cidv1`/`--force-cid-base` command line option.

magik6k:
`--force-cid-base` SGTM

kevina:
@Stebalien 

> what's your objection to `multibase`?

I guess nothing major although `--force-multibase-cids` could mean that CID's are upgraded for other 
purposed other then output.  How about `--force-multibase-output` instead?  Or maybe just `--force-multibase`?

I agree with your points and don't want to bikeshead on this so I will go with what ever you and @magik6k settle on.

Stebalien:
>  How about --force-multibase-output instead?

My only concern with this is that it's not clear that the option only applies to CIDs (we also have peer IDs and potentially other base-encoded things. But I see your point, and I also don't want to bikeshed this to death...

So, we'll let *others* bikeshed it!

@alanshaw and @lidel, this discussion is relevant to you. What would you call the flag that auto-upgrades CIDs from CIDv0 to CIDv1 in output? We've been toying with flags like `--force-multibase-output` and `--force-multibase-cids` (because `--force-multibase-cids-in-output` is a bit long...). `--output-cidv1`, `--upgrade-cidv0`, etc.

Thoughts? Start here: https://github.com/ipfs/go-ipfs/pull/5789#issuecomment-452030458

lidel:
:muscle: :derelict_house: :bike: :selfie:  
My understanding is that the purpose of the flag is to "normalize" CIDs in output, so I'd go with uncreative: 
- `--normalize-cids-in-output`  (["z"](https://books.google.com/ngrams/graph?content=normalize%2Cnormalise&year_start=1900&year_end=2000&corpus=18&smoothing=3&share=&direct_url=t1%3B%2Cnormalize%3B%2Cc0%3B.t1%3B%2Cnormalise%3B%2Cc0) should be ok)
  - or a bit shorter: `--normalize-text-cids`

hsanjuan:
Quick question: How does this relate to  `ipfs add --cid-version` flag? Should it deprecate it?

It would seem that the `force-cid-base` and the `cid-version 1` flag would do a very similar thing? Which also makes me wonder if there should be a global `cid-version` flag instead which you can use to force CidV0 when V1+base32 becomes the default, and for the moment use it to force `cid-version 1` for those that want to auto-upgrade.

kevina:
>Quick question: How does this relate to ipfs add --cid-version flag? Should it deprecate it?

For context note that a multibase is used for display only and does not change the binary representation of the CID.  Upgrading a CIDv0 to v1 changes the binary representation, however `go-ipfs` has a hack in place to allow it to find CIDv0 blocks even if the CIDv1 hash for the block is provided.  For most cases this means the it's okay to use either CIDv0 or CIDv1, but since CIDv0 and CIDv1 have different binary representations, the difference still matters, and hence the need for this flag to cover the corner cases.

That said, the `--force-cid-base` flag will upgrade a CIDv0 to CIDv1 in the output to the user, but internally the CIDv0 is still used,  `--cid-version=1` changes the CID version stored internally.  When adding a large file (larger than the size of a leaf node) `--cid-version=1` will create a different block while `--force-cid-base` won't it will just display the block as CIDv1 and that's it.

Stebalien:
I'm fine with `--upgrade-cids-in-output` (or @kevina's `--upgrade-cidv0-in-output`). I agree it's a bit long but we can define a short flag `-U` (or something like that). I just want this to be accurate.

hsanjuan:
Thanks for the clarification @kevina .

kevina:
@Stebalien the test failures on Travis CI seam unrelated to my changes, and the shareness tests are passing on Cicle CI and Jenkins so we should be good to go test wise.

kevina:
For the record, one of the reasons I objected to such a long name is in the command line help text, which is now (cut off at 80 cols):

```
USAGE
  ipfs - Global p2p merkle-dag filesystem.

SYNOPSIS
  ipfs [--config=<config> | -c] [--debug=<debug> | -D] [--help=<help>] [-h=<h>] 

OPTIONS

  -c,                      --config   string - Path to the configuration file to
  -D,                      --debug    bool   - Operate in debug mode.
  --help                              bool   - Show the full command help text.
  -h                                  bool   - Show a short version of the comma
  -L,                      --local    bool   - Run the command locally, instead 
  --offline                           bool   - Run the command offline.
  --api                               string - Use a specific API instance (defa
  --cid-base                          string - Multibase encoding used for versi
  --upgrade-cidv0-in-output           bool   - Upgrade version 0 to version 1 CI
  --enc,                   --encoding string - The encoding type the output shou
  --stream-channels                   bool   - Stream channel output.
  --timeout                           string - Set a global timeout on the comma

SUBCOMMANDS
...
```
Not a super big deal and something that we can perhaps fix later.

Also, for the record, my original objection to just `--upgrade-cidv0` is that we might want to automatically upgrade CIDv0 pins to CIDv1, but now that I think of it, this is something that is best left to a config option as some sort of migration should be used to convert existing pins when this option is set.

But again, I don't think is is a big deal, especially since it won't be used much, and am more interested in getting this in, then anything else, so I am happy with `--upgrade-cidv0-in-output`.

magik6k:
`--upgrade-cidv0-in-output` is a bit long, but since the alternatives weren't much shorter, so I'm ok with it.

Travis issue is https://github.com/ipfs/go-ipfs/issues/5845. Jenkins is another random fail. Also, needs a rebase.

alanshaw:
I've not got strong feels on this one `--upgrade-cidv0-in-output`  is long but descriptive and as previously observed is unlikely to be used often.

kevina:
@Stebalien let me know when you want to get this in and I will rebase just before.

Stebalien:
@magik6k can you give this a :heavy_check_mark:?

core/commands/pin.go
```diff
@@ -311,19 +322,28 @@ Example:
...
 		}
-
 		if err != nil {
 			return err
 		}
 
-		return cmds.EmitOnce(res, &RefKeyList{Keys: keys})
+		refKeys := map[string]RefKeyObject{}
+		for k, v := range keys {
+			delete(keys, k)
```
Stebalien: This will probably waste more CPU cycles than it's worth in memory.

core/commands/pin.go
```diff
@@ -311,19 +322,28 @@ Example:
...
 		} else {
 			keys, err = pinLsAll(req.Context, typeStr, n)
 		}
-
 		if err != nil {
 			return err
 		}
 
-		return cmds.EmitOnce(res, &RefKeyList{Keys: keys})
+		refKeys := map[string]RefKeyObject{}
```
Stebalien: nit: `refKeys := make(map[string]RefKeyObject{}, len(keys))`

core/commands/cmdenv/cidbase.go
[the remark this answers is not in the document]
kevina: We decided to go with `--upgrade-cidv0-in-output`.

kevina:
@Stebalien this should be good to go.

Not sure why `t0080-repo.sh` is failing on CircleCI, it works locally and with Jenkins.

Stebalien:
(Everything passed, just not all at once. Jenkins is a not having a good day.)

Stebalien:
:fireworks: :rocket: :fireworks:

Issue #5746, opened 2018-11-06
filestore documentation should mention the daemon needs to be restarted


BenLubar:
#### Version information:
```
go-ipfs version: 0.4.18-6f19dd5
Repo version: 7
System version: amd64/linux
Golang version: go1.11.1

(using the Docker Hub image from ipfs/go-ipfs)

docker-compose version 1.22.0, build f46880fe
docker-py version: 3.4.1
CPython version: 3.6.6
OpenSSL version: OpenSSL 1.1.0f  25 May 2017

Docker version 18.06.1-ce, build e68fc7a
```

#### Type:
Task

#### Description:
Running this command:
```
ipfs config --json Experimental.FilestoreEnabled true
```
is not enough; the error message will still be shown when running `ipfs filestore` commands and `--nocopy` will be silently ignored on the `ipfs add` command until the daemon is restarted.

This should be mentioned in the documentation.

joshuahannan:
Can this issue be closed since it looks like it was addressed in #5927 ?

PR #5797, opened 2018-11-26
Improve the Filestore document

bjzhang:
Fix for the #5746. Switch to Filestore need to restart the ipfs node.

License: MIT
Signed-off-by: Bamvor Zhang <jian.zhang@ipfsbit.com>

### DIFF
--- a/docs/experimental-features.md
+++ b/docs/experimental-features.md
@@ -181,7 +181,10 @@ Modify your ipfs config:
 ipfs config --json Experimental.FilestoreEnabled true
 ```
 
-And then pass the `--nocopy` flag when running `ipfs add`
+Restart ipfs node. Otherwise `ipfs add` will not complain and `ipfs filestore`
+will raise error.
+
+Then pass the `--nocopy` flag when running `ipfs add`
 
 ### Road to being a real feature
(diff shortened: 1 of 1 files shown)


docs/experimental-features.md
```diff
@@ -181,7 +181,10 @@ Modify your ipfs config:
 ipfs config --json Experimental.FilestoreEnabled true
 ```
 
-And then pass the `--nocopy` flag when running `ipfs add`
+Restart ipfs node. Otherwise `ipfs add` will not complain and `ipfs filestore`
+will raise error.
+
+Then pass the `--nocopy` flag when running `ipfs add`
```
Stebalien: As written, this is a bit awkward. How about something like:

> Then restart your IPFS node to reload your config.
>
> Finally, when adding files with `ipfs add`, pass the `--nocopy` flag to use the filestore instead of *copying* the files into your local IPFS repo.

bjzhang: Hi, I push the above modification to the new pr: https://github.com/ipfs/go-ipfs/pull/5927. 
I have to create a new pr because the permission of ipbit is changed. Sorry for inconvenience.

PR #5798, opened 2018-11-27
gx: update go-ipfs-config

Stebalien:
* AutoRelay options for #5785.
* Badger truncate-by-default option for #5275, #5625.


Stebalien:
I don't believe merging this before the relevant PRs will cause any issues but please speak up if that's not the case. Unfortunately, due to the fact that this *is* in a separate repo, we kind of have a chicken-egg issue (and I'd rather not *bundle* all the changes together).

kevina:
I never did like the fact that we decided to move `go-ipfs-config` into a separate repo.

I don't see how this can cause a problem, but it appears Jenkins is falling for unrelated reasons.

schomatis:
So this means that we'll be including a `truncate` field now but we will be ignoring it, doing a quick test this doesn't seem to break anything, although it may give the erroneous impression to the user but that seems like a minor issue (assuming https://github.com/ipfs/go-ipfs/pull/5625 will be merged in the near future).

Issue #5763, opened 2018-11-11
Should we hide the peer ids by default in `ipfs bitswap stat`


overbool:
#### Version information:
go-ipfs version: 0.4.19-dev-a88cd82a7
Repo version: 7
System version: amd64/darwin
Golang version: go1.11.1

#### Type:

Enhancement

#### Description:

`ipfs bitswap stat` will print out the follow result:

```
bitswap status
	provides buffer: 0 / 256
	blocks received: 0
	blocks sent: 0
	data received: 0
	data sent: 0
	dup blocks received: 0
	dup data received: 0 B
	wantlist [0 keys]
	partners [284]
		QmNX7qzcDKnSewMT3GZpzMpRpzAndoktgQnmd5JhTi54AE
		QmNXPtM8PsDfsvzBMH7JBDaA39UrWfw6PvRtfTdnAmB9kf
		QmNbEDD6CWjeUj1nixiv3RtvuGQrNZ84BBBKYkTTDay8ph
		QmNcSvqeUjHirbJTXYvRu8ru6ZxvAc91mYGJkV7ExqbQW1
		QmNg1kbAj9FVs9ZGDqbkZxqFFiK4jdrwotwStQyxuRQUW5
		QmNhR9CJ1KoXQiQyp74FFhCniQznsDmbVaiYmXearZVyKw
                ...
                ...
```

Too much output(partners) takes up the entire screen.

Therefore, should we hide them (peer ids) and just print the `partners[284]` by default and add a option `peers` to print all peer ids?

PR #5820, opened 2018-12-05
feat/cmds: hide peers info default in bitswap stat

overbool:
Fixed: https://github.com/ipfs/go-ipfs/issues/5763

License: MIT

### DIFF
--- a/core/commands/bitswap.go
+++ b/core/commands/bitswap.go
@@ -88,10 +88,17 @@ Print out all blocks currently on the bitswap wantlist for the local peer.`,
 	},
 }
 
+const (
+	bitswapVerboseOptionName = "verbose"
+)
+
 var bitswapStatCmd = &cmds.Command{
 	Helptext: cmdkit.HelpText{
 		Tagline:          "Show some diagnostic information on the bitswap agent.",
 		ShortDescription: ``,
+	},
+	Options: []cmdkit.Option{
+		cmdkit.BoolOption(bitswapVerboseOptionName, "v", "Print extra information"),
 	},
 	Type: bitswap.Stat{},
 	Run: func(req *cmds.Request, res cmds.ResponseEmitter, env cmds.Environment) error {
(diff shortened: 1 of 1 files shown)


Stebalien:
In theory, I completely agree with this. Unfortunately, in practice, it changes a long-standing stable interface.

@whyrusleeping thoughts? At the moment, this command is kind of painful due to this issue.

whyrusleeping:
I pretty much always pipe this command to `head -n10`. It would not hurt my feelings if this got changed.

That said, 'bitswap partners' are really supposed to just be peers we are 'actively' trading data with. 'every peer you are connected to' really shouldnt be in that list..

whyrusleeping:
maybe for now, default verbose to true, that way we can use `ipfs bitswap stat -v=false` to get the trimmed output. 

I'm also not against this change as is, but if we go forward with it, we should do a bit of due dilligence to see if anyone is parsing the output of this command and using it for something

eingenito:
I'd suggest not changing the default. It seems monitoring/troubleshooting related and therefore the kind of thing that people might well call in a script (like to check the health of their cluster or something). And because it's a little bit arcane it seems like marginal benefit for a change that's hard to estimate the impact of.

Stebalien:
> And because it's a little bit arcane it seems like marginal benefit for a change that's hard to estimate the impact of.

So, as long as we don't change the *json* output, we should mitigate the damage (although yes, someone is probably parsing the text...).

Maybe a solution is to just *fix* this by only showing *real* bitswap peers. We can probably record a simple "active" refcount for each peer.

However, *that* will have to wait until the current bitswap refactor is done.

eingenito:
> So, as long as we don't change the json output, we should mitigate the damage (although yes, someone is probably parsing the text...).

👍 Interesting. 

> Maybe a solution is to just fix this by only showing real bitswap peers. We can probably record a simple "active" refcount for each peer.

Flip-flop'ing here but if we're comfortable changing the default behavior in the future, and it has little (no?) worth now, maybe it's actually better to merge this and we'll improve the bitswap peers listing when we get to it. I guess right now this is roughly equivalent to `ipfs swarm peers`. 

I can imagine lots of per-peer bitswap stats that might be neat but that would definitely change the json.

Issue #5654, opened 2018-10-26
CoreApi.Offline


magik6k:
Mentioned in https://github.com/ipfs/go-ipfs/pull/5649#discussion_r228379383, it would be nice to have a way to tell CoreAPI that we want to be in offline mode.

There are a few ways to do this:
- Context hints - https://github.com/ipfs/go-ipfs/issues/4009
  - Should be quite contained
  - Easy to implement global `--offline` flag
- `api.Offline()`
  - Should also be quite contained
  - Easy to implement global `--offline` flag too
  - More visible?
- Functional options on the constructor
  - Also not a bad option, should be decently small change
  - Also easy to implement `--offline`

We could also use this to expose sessions (and maybe other global options later?), so I'd say we should go with the more flexible option, which would be context hints

Opinions?

cc @Stebalien

dokterbob:
Trying to add large (150GB+) resources with a running daemon breaks for us (ipfs-search.com). Probably because the daemon crashes during the process.

I might share these resources using `ipfs-pack`. But due to the large resources required by an IPFS daemon, I would strongly prefer to only run one, and there's already one running on our server (eating about 10GB RAM with about 8k connections).

Hence, it would be ideal to be able to add these resources independently of the running (and regularly crashing, due to it hitting hard memory limits) and then adding them as a one-off operation. Hence, being able to do `ipfs add` *as if* there were no daemon running.

Stebalien:
@dokterbob I believe you're looking for the poorly documented `ipfs --local add` flag. *This* issue is about an offline option such that `ipfs get ...` doesn't use bitswap (although it will likely *also* imply `--local`).

Stebalien:
`ipfs --help` will list the global options (including `--local`).

Issue #4182, opened 2017-08-29
Wrong --local/-L flag documentation


lgierth:
#### Description:

`ipfs --help` defines `--local` as "Run the command locally, instead of using the daemon". This is completely inaccurate. The flag is used in only two places:

- `ipfs add --local` uses an Offline Exchange for adding files, skipping the immediate publication of provider records.
- `ipfs name resolve --local` uses an Offline Routing Resolver, avoiding IPNS queries to the DHT.

The flag should be removed from the root command, and instead added to these two commands.

There's a backward compatiblity angle to consider, where tools could possibly be calling `ipfs --local add <file>` instead of the better `ipfs add --local <file>`.

Stebalien:
This command should also be extended to work elsewhere (e.g., `ipfs block stat`). Making this option "just work" everywhere is a pretty decent motivation for https://github.com/ipfs/go-ipfs/issues/4009.

Stebalien:
Fixed. This is now `--offline` everywhere and behaves as expected.

PR #5825, opened 2018-12-06
CoreAPI: Global offline option

magik6k:
For https://github.com/ipfs/go-ipfs/issues/5654. This implements the 'Functional options on the constructor' approach.

I decided to not use `core.IpfsNode` directly in coreapi as this way we have much better control over how the internals are used. Longer term, after coreapi interface is extracted form go-ipfs, I plan on merging a lot of core node and coreapi functionality, but I want to do that together with go-ipfs constructor cleanup as it will be quite breaking for people using `core.IpfsNode` directly (there still are good reasons to do that)

A few bugs and todos are still left, reviews welcome

### DIFF
--- a/core/commands/cmdenv/env.go
+++ b/core/commands/cmdenv/env.go
@@ -6,6 +6,7 @@ import (
 	"github.com/ipfs/go-ipfs/commands"
 	"github.com/ipfs/go-ipfs/core"
 	coreiface "github.com/ipfs/go-ipfs/core/coreapi/interface"
+	options "github.com/ipfs/go-ipfs/core/coreapi/interface/options"
 
 	config "gx/ipfs/QmYyzmMnhNTtoXx5ttgUaRdHHckYnQWjPL98hgLAR2QLDD/go-ipfs-config"
 	cmds "gx/ipfs/QmaAP56JAwdjwisPTu4yx17whcjTr6y5JCSCF77Y1rahWV/go-ipfs-cmds"
@@ -22,13 +23,22 @@ func GetNode(env interface{}) (*core.IpfsNode, error) {
 }
 
 // GetApi extracts CoreAPI instance from the environment.
-func GetApi(env cmds.Environment) (coreiface.CoreAPI, error) {
+func GetApi(env cmds.Environment, req *cmds.Request) (coreiface.CoreAPI, error) {
 	ctx, ok := env.(*commands.Context)
 	if !ok {
 		return nil, fmt.Errorf("expected env to be of type %T, got %T", ctx, env)
 	}
 
-	return ctx.GetAPI()
+	local, _ := req.Options["local"].(bool)
+	api, err := ctx.GetAPI()
+	if err != nil {
+		return nil, err
+	}
+	if local {
+		return api.WithOptions(options.Api.Offline(local))
+	}
+
+	return api, nil
 }
 
 // GetConfig extracts the config from the environment.
--- a/core/commands/cmdenv/env.go
+++ b/core/commands/cmdenv/env.go
@@ -2,6 +2,7 @@ package cmdenv
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/ipfs/go-ipfs/commands"
 	"github.com/ipfs/go-ipfs/core"
@@ -10,7 +11,10 @@ import (
 
 	config "gx/ipfs/QmYyzmMnhNTtoXx5ttgUaRdHHckYnQWjPL98hgLAR2QLDD/go-ipfs-config"
 	cmds "gx/ipfs/QmaAP56JAwdjwisPTu4yx17whcjTr6y5JCSCF77Y1rahWV/go-ipfs-cmds"
+	logging "gx/ipfs/QmcuXC5cxs79ro2cUuHs4HQ2bkDLJUYokwL8aivcX6HW3C/go-log"
 )
+
+var log = logging.Logger("core/commands/cmdenv")
 
 // GetNode extracts the node from the environment.
 func GetNode(env interface{}) (*core.IpfsNode, error) {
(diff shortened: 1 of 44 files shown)


core/coreapi/coreapi.go
```diff
@@ -87,8 +131,110 @@ func (api *CoreAPI) PubSub() coreiface.PubSubAPI {
...
+	subApi.routing = func(allowOffline bool) (routing.IpfsRouting, error) {
+		if !n.OnlineMode() {
+			if !allowOffline {
+				return nil, coreiface.ErrOffline
+			}
+			if err := n.SetupOfflineRouting(); err != nil {
+				return nil, err
+			}
+			subApi.privateKey = n.PrivateKey
+			subApi.namesys = n.Namesys
```
Stebalien: Thread safety?

core/coreapi/coreapi.go
```diff
@@ -15,26 +15,70 @@ package coreapi
...
+	exchange        exchange.Interface
+
+	namesys namesys.NameSystem
+	routing func(bool) (routing.IpfsRouting, error)
+
+	pubSub *pubsub.PubSub
+
+	// TODO: this can be generalized to all functions when we implement some
+	// api based security mechanism
+	isPublishAllowed func() error
```
Stebalien: This method isn't really about security. Really, we *should* allow publishing IPNS records while /ipns is mounted. We don't currently because we haven't implemented that yet.

core/coreapi/coreapi.go
```diff
@@ -87,8 +131,110 @@ func (api *CoreAPI) PubSub() coreiface.PubSubAPI {
...
+
+		peerstore:       n.Peerstore,
+		peerHost:        n.PeerHost,
+		namesys:         n.Namesys,
+		recordValidator: n.RecordValidator,
+		exchange:        n.Exchange,
+
+		pubSub: n.PubSub,
+
+		nd: n,
```
Stebalien: I'm a *bit* concerned users will see this and think these API's have a parent-child relationship. In practice, options from the parent won't apply to the children. *Should* that happen?

core/commands/cmdenv/env.go
```diff
@@ -22,13 +23,22 @@ func GetNode(env interface{}) (*core.IpfsNode, error) {
...
 // GetApi extracts CoreAPI instance from the environment.
-func GetApi(env cmds.Environment) (coreiface.CoreAPI, error) {
+func GetApi(env cmds.Environment, req *cmds.Request) (coreiface.CoreAPI, error) {
 	ctx, ok := env.(*commands.Context)
 	if !ok {
 		return nil, fmt.Errorf("expected env to be of type %T, got %T", ctx, env)
 	}
 
-	return ctx.GetAPI()
+	local, _ := req.Options["local"].(bool)
```
Stebalien: So, this'll apply this option (correctly) to *all* commands fixing part of https://github.com/ipfs/go-ipfs/issues/4182. I wonder if we should fix the "confusing" part while we're at it by renaming this option to "offline" (keeping a hidden alias for compatibility).

That would also allow us to share the same option with `ipfs daemon --offline` which should help to further reduce confusion.

magik6k: I'm fine with doing that (the hidden alias could throw warnings at people). This will probably need support on the js side too

core/coreapi/coreapi.go
magik6k: Yeah, that's a slight problem. Currently we only have 1 boolean option, so not so much, but once we add more, we should carry options struct, and apply `WithOptions` to that instead of the defaults.

core/coreapi/coreapi.go
magik6k: Can fix (`SetupOfflineRouting` also needs fixing)

core/commands/cmdenv/env.go
Stebalien: I didn't realize js replicated that option.

@alanshaw how would you feel about switching from `--local` to `--offline` (keeping `--local` as a backcompat alias)?

core/coreapi/coreapi.go
magik6k: I know, I just wanted to hint that something like this may be useful (but a full security layer implementing coreapi interface would probably be better)

core/commands/cmdenv/env.go
magik6k: Implemented, waiting for @alanshaw's opinion (assuming js-ipfs implements `--local`)

core/coreapi/coreapi.go
```diff
@@ -87,8 +138,121 @@ func (api *CoreAPI) PubSub() coreiface.PubSubAPI {
...
+		nd:         n,
+		parentOpts: settings,
+	}
+
+	var routingLk sync.Mutex
+	var offrt routing.IpfsRouting
+
+	subApi.routing = func(allowOffline bool) (routing.IpfsRouting, error) {
+		routingLk.Lock()
+		defer routingLk.Unlock()
```
Stebalien: We should be able to get rid of this dance now that routing is always configured.

core/coreapi/coreapi.go
```diff
@@ -87,8 +137,99 @@ func (api *CoreAPI) PubSub() coreiface.PubSubAPI {
...
+		exchange:        n.Exchange,
+		routing:         n.Routing,
+
+		pubSub: n.PubSub,
+
+		nd:         n,
+		parentOpts: settings,
+	}
+
+	subApi.isOnline = func(allowOffline bool) error {
```
Stebalien: Code nit: given that this doesn't return a boolean "is online", this should probably be renamed to `checkOnline`.

core/coreapi/coreapi.go
```diff
@@ -87,8 +137,99 @@ func (api *CoreAPI) PubSub() coreiface.PubSubAPI {
...
+		if !n.OnlineMode() && !allowOffline {
+			return coreiface.ErrOffline
+		}
+		return nil
+	}
+
+	subApi.isPublishAllowed = func() error {
```
Stebalien: Same here.

Stebalien:
Mostly nits at this point but you should take a look at https://github.com/ipfs/go-ipfs/pull/5825#discussion_r242755747.

core/coreapi/unixfs.go
```diff
@@ -62,23 +66,15 @@ func (api *UnixfsAPI) Add(ctx context.Context, files files.File, opts ...options
...
+		addblockstore = nilnode.Blockstore
+		exch = nilnode.Exchange
+		pinning = nilnode.Pinning
 	}
 
 	bserv := blockservice.New(addblockstore, exch) // hash security 001
 	dserv := dag.NewDAGService(bserv)
 
-	fileAdder, err := coreunix.NewAdder(ctx, n.Pinning, n.Blockstore, dserv)
+	fileAdder, err := coreunix.NewAdder(ctx, pinning, addblockstore, dserv)
```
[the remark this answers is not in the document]
magik6k: bugfix - `coreunix.NewAdder` uses it only for the `GCLock`, which we don't want/need to take from the 'real' node when we do add with `OnlyHash` set.

Stebalien:
@ipfs/go-team this is an interface change (but not a breaking one).

It replaces `--local` with `--offline`, leaving `--local` for backwards compatibility. It *also* makes `--offline` work with every command (`--local` currently only works with a few commands).

Suggestions? Objections?

magik6k:
(rebased)

PR #5830, opened 2018-12-10
Show hash on gateway

### LINKED ISSUE #5724: Gateways should always display "Index of /ipfs/HASH", instead of the URL.

jackloughran:
@Stebalien this will show the hash if the request doesn't have ipfs in the url:
image
image

fixes #5724


jackloughran:
@Stebalien yes: https://github.com/ipfs/dir-index-html/pull/21. sorry, ran out of time yesterday

core/corehttp/gateway_handler.go
```diff
@@ -347,11 +347,17 @@ func (i *gatewayHandler) getOrHeadHandler(ctx context.Context, w http.ResponseWr
 		}
 	}
 
+	var hash string
+	if !strings.Contains(originalUrlPath, "ipfs") {
```
Stebalien: We should use HasPrefix with `ipfsPathPrefix`. This won't catch IPNS links with the word `ipfs` somewhere in the path.

Stebalien:
(note: I've updated dir-index-html and rebased this)

Stebalien:
@kevina, @magik6k, and/or @swedneck I'd like someone else to sign off on this as both @jackloughran and I participated a bit too much in the design to be neutral third parties.

swedneck:
I can't judge the code, but the screenshots look great.

jackloughran:
@Stebalien thanks for making that change. i got sidetracked with the holidays. looks like this is good to go, but i'd be happy to do anything else over the weekend.

PR #5841, opened 2018-12-13
startup: always load the private key

Stebalien:
Loading this at the last minute means we need a bunch of special cases in *every* command that needs routing, namesys, or even the public key.

If we ever have a case where we don't want to do this, we can add an option to the (eventual) IPFS constructor. Handling this up-front is going to be significantly less error prone.

Motivation: https://github.com/ipfs/go-ipfs/pull/5825/files#diff-fe35ea64d478c4f3fb767a3f618e5d01R863

### DIFF
--- a/cmd/ipfs/init.go
+++ b/cmd/ipfs/init.go
@@ -237,10 +237,5 @@ func initializeIpnsKeyspace(repoRoot string) error {
 	}
 	defer nd.Close()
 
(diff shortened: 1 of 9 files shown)


Stebalien:
Replaced by #5842.

Stebalien:
Replaced by #5842.

PR #5842, opened 2018-12-13
startup: always enable routing

Stebalien:
We can load routing without the key. This means we can avoid the "if I'm offline, lazy-load offline routing" dance.

Motivation: https://github.com/ipfs/go-ipfs/pull/5825/files#diff-fe35ea64d478c4f3fb767a3f618e5d01R863

### DIFF
--- a/cmd/ipfs/init.go
+++ b/cmd/ipfs/init.go
@@ -237,9 +237,11 @@ func initializeIpnsKeyspace(repoRoot string) error {
 	}
 	defer nd.Close()
 
-	err = nd.SetupOfflineRouting()
-	if err != nil {
-		return err
+	if nd.PrivateKey == nil {
+		err = nd.LoadPrivateKey()
+		if err != nil {
+			return err
+		}
 	}
 
 	return namesys.InitializeKeyspace(ctx, nd.Namesys, nd.Pinning, nd.PrivateKey)
--- a/core/coreapi/unixfs_test.go
+++ b/core/coreapi/unixfs_test.go
@@ -707,23 +707,6 @@ func TestGetNonUnixfs(t *testing.T) {
 	}
 }
 
-func TestCatOffline(t *testing.T) {
-	ctx := context.Background()
-	_, api, err := makeAPI(ctx)
-	if err != nil {
-		t.Error(err)
-	}
-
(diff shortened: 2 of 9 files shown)


core/coreapi/unixfs_test.go
```diff
@@ -707,23 +707,6 @@ func TestGetNonUnixfs(t *testing.T) {
 	}
 }
 
-func TestCatOffline(t *testing.T) {
```
Stebalien: This brings up an interesting issue: `ipfs cat` currently works offline (even with IPNS) but `ipfs get` doesn't. After this change, they both work offline.

Should we *not* make this change?

cmd/ipfs/init.go
```diff
@@ -237,9 +237,11 @@ func initializeIpnsKeyspace(repoRoot string) error {
 	}
 	defer nd.Close()
 
-	err = nd.SetupOfflineRouting()
-	if err != nil {
-		return err
+	if nd.PrivateKey == nil {
+		err = nd.LoadPrivateKey()
```
magik6k: I'd say that this is cheap enough to not have to do this lazily too

core/coreapi/unixfs_test.go
magik6k: I'd say that they both should work offline. The only downside is that it might confuse some users because `/ipns` may resolve to old values in datastore without even trying to reach the network (on the other hand we do that with cache too, so..). Maybe we should log a warning from `offroute.GetValue`?

cmd/ipfs/init.go
Stebalien: I'm more worried about not *having* the key. We have quite a few tests ATM where we don't specify one.

(on the other hand, I'm fine making it a requirement)

Stebalien:
> Maybe we should log a warning from offroute.GetValue?

Maybe? I don't want it to be too noisy. Honestly, we use cached values anyways if we can't reach the network (but are operating in "online" mode) so this isn't exactly a new issue (I think?).

cmd/ipfs/init.go
Stebalien: Doing a bit of local end-to-end testing, that .1s is lost in the noise.

Stebalien:
Ok, switching back to always loading the private key: #5844.

PR #5843, opened 2018-12-13
cmds/pin: use coreapi/pin

overbool:
Related: https://github.com/ipfs/go-ipfs/pull/5717#issuecomment-438232351
License: MIT


core/commands/pin.go
```diff
@@ -364,7 +424,7 @@ new pin and removing the old one.
 	Options: []cmdkit.Option{
 		cmdkit.BoolOption(pinUnpinOptionName, "Remove the old pin.").WithDefault(true),
 	},
-	Type: PinOutput{},
+	Type: PinUpdateOutput{},
```
magik6k: While this objectively makes more sense, it's also api-breaking. There are some loose plans to implement 'next-gen' api - https://github.com/ipfs/http-api-spec/issues/116 which should address issues like this, but until this happens we'd like to keep this api as stable as possible,

core/coreapi/interface/options/pin.go
```diff
@@ -8,12 +8,23 @@ type PinLsSettings struct {
 	Type string
 }
 
+// PinRmSettings represents the settings of pin rm command
+type PinRmSettings struct {
+	Recursive bool
+	Force     bool
```
magik6k: `Force` doesn't appear to be used

core/commands/pin.go
```diff
@@ -215,20 +233,62 @@ collected if needed. (By default, recursively. Use -r=false for direct pins.)
...
+				return err
+			}
+
+			rp, err := api.ResolvePath(req.Context, p)
+			if err != nil {
+				return err
+			}
+
+			if err := api.Pin().Rm(req.Context, rp, options.Pin.RmRecursive(recursive)); err != nil {
+				if err := res.Emit(&PinOutput{
```
magik6k: This is still going to break js-ipfs-api.

I'd keep a list of strings with `rp.Cid().String()` and use `Error map[string]error` for tracking errors, and at the end just do `cmds.EmitOnce(&PinOutput{pins, errors})`.

This is somewhat hacky, but should work. Maybe @Stebalien has some better idea

magik6k:
There are some pin-related tests failing - https://ci.ipfs.team/blue/organizations/jenkins/IPFS%2Fgo-ipfs/detail/PR-5843/19/tests

Stebalien:
Unfortunately, it looks like the commands library was duck-typing structs with an `Error` field as an error. Fixed by renaming it to `Errors`.

In the past, errors/non-errors were indistinguishable without duck-typing (in some cases). We've now fixed that so the client *should* only interpret something as an error if:

1. The command returns a non-200 status.
2. The command returns the error in a trailing header.

However, we're still ducktyping for backwards compatibility for the moment.

overbool:
@magik6k @Stebalien 
I don't know much about the difference between `Encoders` and `PostRun`.
Should we replace `Encoders` with `PostRun` (like `blockRmCmd`) about this:  https://github.com/ipfs/go-ipfs/blob/b36037a3a5ca71108d1d4673f4925f8b5b813d0c/core/commands/pin.go#L225-L230

magik6k:
> I don't know much about the difference between `Encoders` and `PostRun`.

The difference is that `Encoders` can run on the server when `?encoding` parameter is specified. `PostRun` only runs on the client side. `PostRun` also allows to track state more easily and do things after the command has ran. Generally `Encoders` should be used instead of `PostRun` where possible.

overbool:
@Stebalien @magik6k All tests passed. Could u help me review it again?

Stebalien:
Let's wait for https://github.com/ipfs/go-ipfs/pull/5789 to avoid having to rework that. We should merge it within a few days.

overbool:
> This is unblocked but will need a rebase.

It`s ok.

core/commands/pin.go
```diff
@@ -226,18 +238,38 @@ collected if needed. (By default, recursively. Use -r=false for direct pins.)
...
+			}
+
+			rp, err := api.ResolvePath(req.Context, p)
+			if err != nil {
+				return err
+			}
+
+			id := enc.Encode(rp.Cid())
+			pins = append(pins, id)
+			if err := api.Pin().Rm(req.Context, rp, options.Pin.RmRecursive(recursive)); err != nil {
```
Stebalien: We need to discuss this with the JS team as well: https://github.com/ipfs/interface-ipfs-core/issues/127#issuecomment-456561650.

overbool:
@Stebalien I think we can add `--force` option in another pr(https://github.com/ipfs/go-ipfs/pull/5717). This pr just replace `corerepo/pin` with `coreapi/pin` and will keep original feature that will bail out when it encounters an error.

Stebalien:
> I think we can add --force option in another pr(#5717).

SGTM.

Issue #5848, opened 2018-12-17
Use Go modules


anacrolix:
With Go 1.11, Go modules may be an option for versioning dependencies.

mhchia:
@anacrolix  May I ask how did you generate the `go.mod` in the PR https://github.com/ipfs/go-ipfs/pull/5849? (Sorry for asking here. I just don't want to interrupt the code review in the PR)

I'm also trying to change from gx to gomod in my project. In my understanding, this needs to resolve the `gx_hash` or version of each gx dependency to git repo and (commit or tag). After resolving, we can make use of `go get` to update the dependency to `go.mod`. To resolve the hash, I had some naive approaches, but not sure if they are correct. Please correct me if anything wrong:

- To find commits, my naive approach was to download and synchronize the corresponding git repo for each dependency, and try to `git grep` the `gx_hash` from the repo, to see which commit first contains that `gx_hash`. However, some repo does not contain `.gx/lastpubver`, possibly because that the repo doesn't use gx and was just pushed to gx(ipfs), or, the repo just doesn't contain `.gx/lastpubver`.
- To find versions(tags), I just search if there is a same version tag in the git repo as the gx version specified in `package.json`. However, I found there are some inconsistence. E.g. in github.com/libp2p/go-libp2p-kad-dht the gx version is `4.4.17` by the time I write this question, while the latest tag in the git repo is `v4.4.12`.

Because of the problems above, I couldn't resolve all gx versions/hashes of my dependencies. I'm really interested in if there is a way to map the gx version/hash to the git repo version/commit. This way we can make use of `go.mod` for projects which previously used `gx`, before a formal solution has come out discussed in https://github.com/ipfs/go-ipfs/issues/5850 and https://github.com/whyrusleeping/gx/issues/200

mhchia:
@anacrolix  Thanks a lot for the explanation. Your approach sounds pretty reasonable. In my situation, I have roughly 100 dependencies including the transitive ones, and I can only resolve the git version/commit for three-fourths of them from gx hashes. So I will rather use the latest versions in my case, following your approach.

> or unpack the gxDependencies.hash value, which may or may not correspond to something git is aware of

It's a cool thing. I thought it is just something similar to the hash of the package. Will see if there is any related information from there.

magik6k:
All gx packages have `.gx/lastpubver` which you could use to assign versions to commits. I'm not sure if it will work everywhere because you may hit duplicate versions on merged branches in some packages.

[...]

PR #5849, opened 2018-12-17
Go module support

anacrolix:
Per #5848 .


lanzafame:
@anacrolix See #5435 for moving to a `gx` unwritten state, which is essentially the same as moving to go modules just minus a go.mod and go.sum.

Stebalien:
This will need quite a bit of discussion before we consider it.

anacrolix:
Where would such a discussion occur?

Stebalien:
We generally like to discuss large changes in issues first. Both in your issue (to discuss adopting go modules) and https://github.com/ipfs/go-ipfs/issues/5850 (where I've brought up simply switching to them entirely).

anacrolix:
Can this be reopened?

Stebalien:
No, it cannot. See: https://github.com/ipfs/go-ipfs/issues/5850#issuecomment-450026195

anacrolix:
I don't follow, the comment does not make sense.

Stebalien:
The issue is that semver/go-dep exist for a reason. If we switch over to go-mod *now* (without full go-dep support), we won't get *any* dependency version checking. That means we can easily end up in the following situation:

1. Make a breaking change in a transitive dependency (D) of go-ipfs.
2. Fix all packages (I) directly depending on D.
3. Fix a bug in package D.
4. In go-ipfs, update D without updating I (to pull in the bugfix).

gx forces *exact* version matching so isn't a problem. If we update package D, we'd *have* to use the latest versions of the packages in I.

go-dep fixes this problem using semver and version constraints.

anacrolix:
Can you link to go-dep, I'm not sure what you're referring to. I don't see how the situation you describe is a problem. One would not update D in step for just for the sake of it (particularly if D is not a direct dependency of go-ipfs), and I's and go-ipfs's tests would fail if one did and it was a problem in either of them. Furthermore if the change in D was truly breaking, the good author of that package would create a new major version, preventing irresponsible updates downstream.

Stebalien:
> Can you link to go-dep?

Sorry, too many package managers. I meant go-mod (the go module system).

> Furthermore, if the change in D was truly breaking, the good author of that package would create a new major version preventing irresponsible updates downstream.

This ^^. If we use this PR as-is, we don't get the ability to specify version constraints like this.

Really, we don't need a breaking change to run into this issue. All we need to do is:

1. Try to make a change in some package I.
2. Realize there's a bug in package D (or a missing feature).
3. Fix the bug/feature in package D.
4. Update the minimum version of D in I.
5. Make the desired change to I.
7. Update package I in go-ipfs.

With full go modules support in packages D and I, updating the minimum version of package I in go-ipfs will automatically update the minimum version of D. As it stands, we could end up updating D without updating I.

---

Basically, we're already using a package manager with basic version constraint checking. Merging *this* PR without fully switching over all of our other packages to use go-mod would be a pretty big step backwards.

anacrolix:
As it is, people not being able to build without using a standard tool is the biggest blocker. This PR addresses that. Updates to transitive dependencies are still visible in updates downstream, they're listed as `// indirect` in the `go.mod` and should be present in commit diffs.

I find that MVS puts all the power in the hands of the most downstream project, which should be able to make the important decisions about dependencies. Existing popular package constraint systems allow middle packages to meddle and alter builds without agreement from the downstream projects. Maybe `gx` doesn't do that, but critically `go.mod` provides the ability to review and audit all changes project by project and issue immutable corrections if errors are made.

Stebalien:
> Updates to transitive dependencies are still visible in updates downstream, they're listed as // indirect in the go.mod and should be present in commit diffs.

That only applies to dependencies that *have* go.mod files. My point was that, as-is, our packages don't have them (and most aren't even using semver).

Issue #5852, opened 2018-12-18
Leaking goroutines on datastore queries


Stebalien:
We're leaking goroutines on datastore queries when:

1. Using flatfs.
2. Closing them early.

Fix: https://github.com/ipfs/go-ds-flatfs/pull/51

Stebalien:
cc @kyledrake

magik6k:
Can this be closed?

PR #5857, opened 2018-12-19
gx: update pubsub, connmgr, and flatfs

### LINKED ISSUE #5817: pubsub: already have connection to peer
Hi,
 I trying use IPFS in my project, and pubsub for the broadcast messages and works fine but on ipfs daemon console appears this message every few minutes.
"16:58:25.275 ERROR     pubsub: already have connection to peer:  <peer.ID Qm*BewHac> pubsub.go:268"
Why appears this message? I'm not be trying connect because the connection between peers is established from first boot.
I think that's is a bug (or i'm not using pubsub correctly)
any idea?

Stebalien:
Fixes #5817 #5852


Stebalien:
(gx only change)

PR #5860, opened 2018-12-19
make: fix building source tarball on macos

Stebalien:
Unfortunately, `readlink -f` doesn't work on macos. This change *manually* converts a relative path to an absolute path.


magik6k:
(don't have mac to test on, but it should work)

PR #5864, opened 2018-12-20
docs/code-flow : Add code flow documentation for add cmd.

nitishm:
License: MIT
Signed-off-by: Nitish Malhotra <nitish.malhotra@gmail.com>

### DIFF
--- a/docs/add-code-flow.md
+++ b/docs/add-code-flow.md
@@ -1,8 +1,6 @@
 # IPFS : The `Add` command demystified
 
-https://github.com/ipfs/go-ipfs/blob/v0.4.18/core/commands/add.go#L208-L213
-
-The goal of this document is to capture the code flow for adding a file (see the `coreapi` package) using the IPFS CLI, in the process exploring some datastructures and packages like `ipld.Node` (aka `dagnode`, `FSNode`, `MFS`, etc.
+The goal of this document is to capture the code flow for adding a file (see the `coreapi` package) using the IPFS CLI, in the process exploring some datastructures and packages like `ipld.Node` (aka `dagnode`), `FSNode`, `MFS`, etc.
 
 **Try this yourself**
 > 
@@ -42,11 +40,11 @@ Within the function, a new `Adder` is created with the configured `Blockstore` a
 - **[`adder.AddAllAndPin(files)`](https://github.com/ipfs/go-ipfs/blob/v0.4.18/core/coreunix/add.go#L403)** - *Entrypoint to the `Add` logic*
     encapsulates a lot of the underlying functionality that will be investigated in the following sections. 
 
-    Our focus will be on the simplest case, a single file, handled by `Adder.addFileNode(path, file)` handled by `Adder.addFile(pathm files.File)`. 
+    Our focus will be on the simplest case, a single file, handled by `Adder.addFile(file files.File)`. 
 
-  - **[`adder.addFile(path, files.File)`](https://github.com/ipfs/go-ipfs/blob/v0.4.18/core/coreunix/add.go#L450)** - *Create the _DAG_ and add to `MFS`*
+  - **[`adder.addFile(file files.File)`](https://github.com/ipfs/go-ipfs/blob/v0.4.18/core/coreunix/add.go#L450)** - *Create the _DAG_ and add to `MFS`*
 
-      The `addFile(file)` method takes the data and converts it into a __DAG__ tree and adds the root of the tree in to the `MFS`.
+      The `addFile(file)` method takes the data and converts it into a __DAG__ tree and adds the root of the tree into the `MFS`.
 
       https://github.com/ipfs/go-ipfs/blob/v0.4.18/core/coreunix/add.go#L508-L521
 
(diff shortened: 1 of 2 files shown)


schomatis:
@nitishm hey, sorry, I totally forgot about this, I've been too focused on the MFS front when I should be prioritizing documents like this one that can have a much greater impact on our community.

docs/add-code-flow.md
```diff
@@ -0,0 +1,97 @@
...
+https://github.com/ipfs/go-ipfs/blob/master/core/coreunix/add.go#L508-L521
+
+There are two main methods to focus on -
+
+### `Adder.add(io.Reader)`
+*Create and return the **root** __DAG__ node*
+
+https://github.com/ipfs/go-ipfs/blob/master/core/coreunix/add.go#L115-L137
+
+This method converts the input _data_ (`io.Reader`) to a `DAG` tree. This is done by splitting the data into _chunks_ using the `Chunker` **(HELP: elaborate on chunker types ?)** and organizing them in a `DAG` (with a *trickle* or *balanced* layout). The method returns the **root** of the __DAG__, formatted as an `ipld.Node`.
```
schomatis: Let's just mention the balanced builder, we can point to the code documentation which has some illustrative diagrams:

https://github.com/ipfs/go-unixfs/blob/6b769632e7eb8fe8f302e3f96bf5569232e7a3ee/importer/balanced/builder.go

(`go-unixfs` version associated with `go-ipfs` 4.18).

docs/add-code-flow.md
```diff
@@ -0,0 +1,97 @@
...
+Assuming the directory structure already exists in the MFS file system, (if it doesn't exist it will be created using `mfs.Mkdir()` function by passing in the `MFS` **root**), the **root** __DAG__ node is added to the `MFS` File system within the `mfs.PutNode()` function.
+
+#### `[MFS] PutNode(mfs.Root, path, ipld.Node)`
+*Insert node at path into given `MFS`*
+
+https://github.com/ipfs/go-mfs/blob/master/ops.go#L101-L113
+
+In this the path is used to determine the `MFS` `Directory`, which is first lookup up in the `MFS` using `lookupDir()` function. This is followed by adding the **root** __DAG__ node (`ipld.Node`) in to the found `Directory` using `directory.AddChild()` method.
+
+#### - `directory.AddChild(filename, ipld.Node)`
```
schomatis: These 3 "add child to directory" subsections could be collapsed into one, we can definitely not mentioned anything HAMT related (using the `BasicDirectory` if we need to give any concrete example). The most important takeaway for the reader should be that MFS handles the UnixFS layer, it adds the child entry in whichever format UnixFS uses and updates its own bookkeeping records.

docs/add-code-flow.md
```diff
@@ -0,0 +1,97 @@
...
+
+---
+
+## `adder.Finalize()`
+*Fetch and return the __DAG__ **root** from the `MFS` and `UnixFS` directory*
+
+https://github.com/ipfs/go-ipfs/blob/master/core/coreunix/add.go#L199-L244
+
+The whole process ends with `adder.Finalize()` which returns the `ipld.Node` from the `UnixFS` `Directory`.
+**(HELP: Do we need to elaborate ?)**
```
schomatis: I don't think so, we should just mention that at the end of `AddAllAndPin` (after `Finalize`) we call `PinRoot` to recursively pin all the files under the MFS root.

nitishm:
I need to get to this soon. 

Will try to wrap this up by the end of next week ! (sans the `delve` part, which might need me to do some investigation)

schomatis:
Sounds good, let me know if you need any help.

nitishm:
@schomatis I cleaned up the doc and formatted it in a better way.

Cant figure out what version of `go-mfs` to reference (kept it as master for now. Will change after your comment) ?

schomatis:
> Cant figure out what version of `go-mfs` to reference

That would be the version checked by the `go-ipfs` version we're documenting, in this case it would be

https://github.com/ipfs/go-ipfs/blob/aefc746f34e5ffdee5fba1915c6603b65a0ebf81/package.json#L533-L535

This is the Gx version that may not coincide with the actual `go-mfs` GitHub version (at some point publishing in Gx meant also tagging a release in the repo but I'm not sure now).

So, that version is not listed in the releases,

https://github.com/ipfs/go-mfs/releases

so basically what I do (there may be other ways, I hope) is to trace back the `.gx/lastpubver` which records the Gx version every time a `gx publish` is done, I go to the history,

https://github.com/ipfs/go-mfs/commits/master/.gx/lastpubver

and find the commit that corresponds to that Gx version published,

https://github.com/ipfs/go-mfs/commit/8b5e8dd9b7aff8807aaa5391f8f48f1c6d513229#diff-d95150d9db6fb31091770e6d6105b08d

where you can see the hash `QmUwXQs8` that matches the one in `package.json` (the hash is always the ultimate source of truth in IPFS, and hence in Gx),

https://github.com/ipfs/go-mfs/commit/8b5e8dd9b7aff8807aaa5391f8f48f1c6d513229#diff-d95150d9db6fb31091770e6d6105b08dR1

Yes, this workflow definitely needs to be improved, I'm not sure why we stop tagging the Gx publishes in GitHub.

Stebalien:
> Yes, this workflow definitely needs to be improved, I'm not sure why we stop tagging the Gx publishes in GitHub.

https://github.com/whyrusleeping/gx/issues/157

(all existing tags are created manually or with https://github.com/whyrusleeping/gx/blob/master/contrib/gx-retrotag.sh)

schomatis:
Oh, too bad, I'll update the tags then.

nitishm:
@schomatis Is there anything else you would like me to add to the docs ? The `delve` piece is still pending but I am not sure how do get enough information in that way. I think keeping it simple with the current format might be sufficient for now.

LMK what you think ?

docs/add-code-flow.md
```diff
@@ -80,18 +78,16 @@ Within the function, a new `Adder` is created with the configured `Blockstore` a
...
 
-                > NOTE: This is not to be confused with the `directory.AddChild(filename, ipld.Node)`, as this operates on the `UnixFS` `BasicDirectory` object only.
+                > NOTE: This is not to be confused with the `directory.AddChild(filename, ipld.Node)`, as this operates on the `UnixFS` `BasicDirectory` object.
 
             - **[UnixFS] [`(BasicDirectory).AddChild(ctx, name, ipld.Node)`](https://github.com/ipfs/go-unixfs/blob/v1.1.16/io/directory.go#L137)** - *Add child to `BasicDirectory`*
 
-                > IMPORTANT: It should be noted that the `BasicDirectory` struct of the `UnixFS` package, encasulates a node object of type `ProtoNode`, which is a different format from the `ipld.Node` we have been working on throughout this document.
+                > IMPORTANT: It should be noted that the `BasicDirectory` object uses the `ProtoNode` type object which is an implementation of the `ipld.Node` interface, seen and used throughout this document. Ideally the `ipld.Node` should always be used, unless we need access tp specific functions from `ProtoNode` (like `Copy()`) that are not available in the interface.
```
schomatis: nit: typo, `tp` -> `to`.

docs/add-code-flow.md
```diff
@@ -100,4 +96,8 @@ Within the function, a new `Adder` is created with the configured `Blockstore` a
 
   - **[`adder.PinRoot()`](https://github.com/ipfs/go-ipfs/blob/v0.4.18/core/coreunix/add.go#L171)** - *Pin all files under the `MFS` **root***
 
-    The whole process ends with `PinRoot` recursively pinning all the files under the `MFS` **root**
\ No newline at end of file
+    The whole process ends with `PinRoot` recursively pinning all the files under the `MFS` **root**
+
+# Also see
+1. https://github.com/ipfs/go-ipfs/#development
```
schomatis: Sorry, I probably wasn't very clear, I meant for you to add a link in the https://github.com/ipfs/go-ipfs/#development section pointing here, so this document wouldn't get lost in the `docs` directory.

docs/add-code-flow.md
```diff
@@ -100,4 +96,8 @@ Within the function, a new `Adder` is created with the configured `Blockstore` a
...
+2. [Concept document about files](https://github.com/ipfs/docs/issues/133)
```
schomatis: We can probably move this to the introduction (before actually starting to talk about MFS/UnixFS).

docs/add-code-flow.md
nitishm: Oh you meant only in the issue conversation space and not the actual doc?! Oops!

schomatis: In the actual entry-point README doc, but instead of pointing from here to there, point there to here.

nitishm: Oh dang. I completely misunderstood.  I didn't realize the link was the the top README.md doc. Alright I will go ahead and add it to the README.md

schomatis: Oh, you're right, sorry, I didn't realize the link was copied without the `README.md` path :grimacing:

PR #5865, opened 2018-12-20
Move coreapi tests to the interface

magik6k:
Needs to be done before interface extraction

### DIFF
--- a/core/coreapi/interface/tests/api.go
+++ b/core/coreapi/interface/tests/api.go
@@ -71,7 +71,7 @@ func TestApi(p Provider) func(t *testing.T) {
 			select {
 			case <-zeroRunning:
 			case <-time.After(time.Second):
-				t.Errorf("%d node(s) not closed", running)
+				t.Errorf("%d test swarms(s) not closed", running)
 			}
 		})
 	}
--- a/core/coreapi/interface/tests/api.go
+++ b/core/coreapi/interface/tests/api.go
@@ -3,6 +3,7 @@ package tests
 import (
 	"context"
 	"testing"
+	"time"
 
 	coreiface "github.com/ipfs/go-ipfs/core/coreapi/interface"
 )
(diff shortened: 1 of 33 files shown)


eingenito:
Hey @magik6k, I'm taking a look at this and I can follow what you did mechanically, but I don't understand why you did it? I have some questions; I apologize if they don't make sense.

* What's the extraction you're referring to in the initial comment?
* What's the motivation for tracking contexts as you do here? Has this been an issue in CI or something?
* I think this will prevent us from running any of these tests individually? Is that a big deal?

Any other context you could provide quickly would be great. I get that I might just be the wrong reviewer but you mentioned this PR about two weeks ago and I wanted to get you some feedback. Thanks.

core/coreapi/interface/tests/api.go
```diff
@@ -0,0 +1,78 @@
...
+		t.Run("Pin", tp.TestPin)
+		t.Run("PubSub", tp.TestPubSub)
+		t.Run("Unixfs", tp.TestUnixfs)
+
+		apis <- -1
+		t.Run("TestsCancelCtx", func(t *testing.T) {
+			select {
+			case <-zeroRunning:
+			case <-time.After(time.Second):
+				t.Errorf("%d node(s) not closed", running)
```
eingenito: nit: we're not really tracking nodes here right? More like constructed swarms.

magik6k:
> What's the extraction you're referring to in the initial comment?

https://github.com/ipfs/interface-ipfs-core/issues/66, basically we don't want external implementations of coreapi to depend on go-ipfs repo

> What's the motivation for tracking contexts as you do here? Has this been an issue in CI or something?

[go-ipfs-http-api](https://github.com/ipfs/go-ipfs-http-client/pull/1) uses IPTB to spawn nodes and relies on contexts to kill those nodes.

> I think this will prevent us from running any of these tests individually? Is that a big deal?

Running all coreapi tests takes about 3-4s for the `core/coreapi` implementation on my PC. It will probably be much slower with http-api+IPTB, but filtering with `-run` still works just fine (e.g. `go test github.com/ipfs/go-ipfs/core/coreapi/test -run='TestIface/Object/TestObjectPut' -v`)

core/coreapi/interface/tests/block.go
```diff
@@ -103,16 +116,17 @@ func TestBlockGet(t *testing.T) {
...
+	defer cancel()
+	api, err := tp.makeAPI(ctx)
 	if err != nil {
 		t.Error(err)
 	}
 
 	res, err := api.Block().Put(ctx, strings.NewReader(`Hello`))
 	if err != nil {
-		t.Error(err)
+		t.Fatal(err)
```
eingenito: What's the pattern for the replacement of Error with Fatal? Are we testing the first use of the CoreAPI generally?

core/coreapi/interface/tests/key.go
```diff
@@ -125,11 +154,12 @@ func TestGenerateSize(t *testing.T) {
 	}
 }
 
-func TestGenerateType(t *testing.T) {
-	ctx := context.Background()
+func (tp *provider) TestGenerateType(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
```
eingenito: These could go below skip?

core/coreapi/interface/tests/block.go
eingenito: I think I understand - it's an existing inconsistency and you didn't want to change them wholesale.

core/coreapi/interface/tests/key.go
Stebalien: (yes but I didn't see that comment before I merged... oh well, it doesn't really matter much).

Issue #5822, opened 2018-12-05
Providing from bitswap slows down the gateways over time.


Stebalien:
We currently send out provider records whenever we fetch a block from bitswap. Unfortunately, when under *constant* load, our provider record queue can back up quite a bit. This:

1. Slowly (over weeks) eats up *gigabytes* of memory.
2. Blocks up the dialer queue. This ends up making it really hard to actually *fetch* content via the gateway because the gateway is busy working through a massive queue of DHT requests.

---

Proposed solution: In addition to all the provider record work being done in #5774, we should consider adding an option to simply disable provider records (at least for files fetched via the gateway).

Stebalien:
That's my thinking but I haven't thought through it fully. This would *also* be fixed by:

1. Having an option to turn off providing globally.
2. Obeying reprovider strategies on provide *as well* and configuring the gateway with a "pin-only" provider strategy.
3. Reducing the number of blocks we end up providing *and* making providing "lazy" (that is, write down somewhere, e.g. in a database, that we want to provide some block and then work through the provider backlog as a background idle task).

(in order of increasing complexity)

---

Really, I just filed this issue to *track* the issue (so we can keep track of what's *currently* killing our gateways :disappointed:).

michaelavila:
Understood. It's closely related to what I'm doing so I want to at least have it in mind. Thanks.

Stebalien:
This has been fixed. We still have the large queue but it's manageable and we've reduced the number of workers.

Issue #4824, opened 2018-03-16
[badger] GC: Data may remain (up to 1GB) after garbage collecting everything


schomatis:
Regarding a future transition to Badger, its garbage collection process [bails out](https://github.com/dgraph-io/badger/blob/8b89769ed71e56cf7f80abdcfd3576efc97bd412/value.go#L891-L892) in the case there is only one value log file, and by default the size of a value log file is set to [1GB](https://github.com/ipfs/go-ipfs/blob/52301ce84121c1446de71b698219656189f12b2e/repo/fsrepo/datastores.go#L374-L375). This translates to the scenario where a user adds data and then "deletes" it (e.g., unpinning and garbage collecting it) and it will still have up to 1GB of disk taken with these values marked as deleted but not reclaimed by Badger's GC.

This is not an error, and for big players 1GB may be negligible, but it will raise a normal end user's attention, and should be taken in consideration (maybe in the form of clear documentation clarifying that, or catching the [`ErrNoRewrite`](https://github.com/ipfs/go-ds-badger/blob/bfda9737bd7f9463fd70b58b7bf8be418ab3cb4d/datastore.go#L265) error and printing a warning to the user when appropriate).

schomatis:
As a simple PoC,

```
ipfs init --profile=badgerds
du -hc  ~/.ipfs
# 84K	total
ipfs add -r --pin=false --silent ~/go/src/github.com/ipfs/go-ipfs
ipfs repo stat
# NumObjects: 5563
# RepoSize:   140915008
du -hc  ~/.ipfs
# 135M	total
ipfs repo gc -q
ipfs repo stat
# NumObjects: 13
# RepoSize:   143023447
du -hc  ~/.ipfs
# 137M	total
```

(compare with running the same commands without `--profile=badgerds`.)

schomatis:
Opened issue dgraph-io/badger#442 to track the minimum value log file size allowed as a possible workaround for the user who wishes to minimize the used disk space after garbage collection.

leerspace:
I might be mistaken, but I don't think that badger's GC is triggered yet during IPFS' GC: #4300. Or, at least it doesn't seem to be yet on 0.4.14-rc2.

schomatis:
[...]

PR #5870, opened 2018-12-21
RFC#2: Add provider to ipfs and provide when adding/fetching (WIP)

michaelavila:
This is a follow up to https://github.com/ipfs/go-ipfs/pull/5840

My goal is to get some feedback on my progress towards reworking providing strategies and to offer a possible solution to https://github.com/ipfs/go-ipfs/issues/5822.

## Overview
What is in this PR:

### Provider
- Use a strategy to decide which CIDs below a CID to provide
- Queue CIDs to provide on disk, using the datastore (the "provide queue")
    - only if we haven't already provided this CID
- Have workers work through the provide queue announcing each CID
- When a CID is successfully announced, track it so that we know not to provide it again
- Call `IpfsNode.Provider.Provide(cid)` all over the place
    - it's possible I've missed some places, I should triple check

### Reprovider
- Queue the list of all currently provided CIDs
- Have workers work through the reprovide queue

What is **NOT** in this PR:
- A robust provide strategy

## TODO
- [x] reprovider based on tracker data and provider module
- [x] rebase
- [x] move `ipfs bitswap reprovide` to `ipfs provider reprovide`
- [ ] move to core api
- [ ] timestamp provider entries?
- [ ] caller configured provider strategies
- [ ] migrating data when nodes get upgraded to this new system
- [ ] tests

## Known Things
- duplicates can be added to the provide/reprovide queues but no duplicate providing work will be done in terms of content routing

### DIFF
--- a/core/commands/tar.go
+++ b/core/commands/tar.go
@@ -61,9 +61,18 @@ represent it.
 
 		c := node.Cid()
 
+//<<<<<<< HEAD
 		return cmds.EmitOnce(res, &AddEvent{
 			Name: it.Name(),
 			Hash: enc.Encode(c),
+//=======
+//		nd.Provider.Provide(c)
+//
+//		fi.FileName()
+//		return cmds.EmitOnce(res, &coreiface.AddEvent{
+//			Name: fi.FileName(),
+//			Hash: c.String(),
+//>>>>>>> Add provider to ipfs and provide when adding/fetching
 		})
 	},
 	Type: AddEvent{},
--- a/core/commands/tar.go
+++ b/core/commands/tar.go
@@ -43,6 +43,11 @@ represent it.
 			return err
 		}
 
+		api, err := cmdenv.GetApi(env, req)
+		if err != nil {
+			return err
+		}
+
 		enc, err := cmdenv.GetCidEncoder(req)
 		if err != nil {
 			return err
@@ -60,19 +65,11 @@ represent it.
 		}
 
 		c := node.Cid()
+		api.Provider().Provide(c)
 
-//<<<<<<< HEAD
 		return cmds.EmitOnce(res, &AddEvent{
 			Name: it.Name(),
 			Hash: enc.Encode(c),
-//=======
-//		nd.Provider.Provide(c)
-//
-//		fi.FileName()
-//		return cmds.EmitOnce(res, &coreiface.AddEvent{
-//			Name: fi.FileName(),
-//			Hash: c.String(),
-//>>>>>>> Add provider to ipfs and provide when adding/fetching
 		})
 	},
 	Type: AddEvent{},
--- a/core/core.go
+++ b/core/core.go
@@ -357,38 +357,6 @@ func (n *IpfsNode) startLateOnlineServices(ctx context.Context) error {
 	n.Reprovider = provider.NewReprovider(ctx, rq, tracker, reproviderInterval, n.Blockstore, n.Routing)
 	n.Reprovider.Run()
 
-	/*
-	// Reprovider - old
-
-	var keyProvider rp.KeyChanFunc
-
-	switch cfg.Reprovider.Strategy {
-	case "all":
-		fallthrough
-	case "":
-		keyProvider = rp.NewBlockstoreProvider(n.Blockstore)
-	case "roots":
-		keyProvider = rp.NewPinnedProvider(n.Pinning, n.DAG, true)
-	case "pinned":
-		keyProvider = rp.NewPinnedProvider(n.Pinning, n.DAG, false)
-	default:
-		return fmt.Errorf("unknown reprovider strategy '%s'", cfg.Reprovider.Strategy)
-	}
-	n.Reprovider = rp.NewReprovider(ctx, n.Routing, keyProvider)
-
-	reproviderInterval := kReprovideFrequency
-	if cfg.Reprovider.Interval != "" {
-		dur, err := time.ParseDuration(cfg.Reprovider.Interval)
-		if err != nil {
-			return err
-		}
-
-		reproviderInterval = dur
-	}
-
-	go n.Reprovider.Run(reproviderInterval)
-	*/
-
 	return nil
 }
 
--- a/core/core.go
+++ b/core/core.go
@@ -21,13 +21,13 @@ import (
 	"time"
 
 	version "github.com/ipfs/go-ipfs"
-	rp "github.com/ipfs/go-ipfs/exchange/reprovide"
 	filestore "github.com/ipfs/go-ipfs/filestore"
 	mount "github.com/ipfs/go-ipfs/fuse/mount"
 	namesys "github.com/ipfs/go-ipfs/namesys"
 	ipnsrp "github.com/ipfs/go-ipfs/namesys/republisher"
 	p2p "github.com/ipfs/go-ipfs/p2p"
 	pin "github.com/ipfs/go-ipfs/pin"
+	provider "github.com/ipfs/go-ipfs/provider"
 	repo "github.com/ipfs/go-ipfs/repo"
 
 	bitswap "github.com/ipfs/go-bitswap"
@@ -124,7 +124,8 @@ type IpfsNode struct {
 	Routing      routing.IpfsRouting // the routing system. recommend ipfs-dht
 	Exchange     exchange.Interface  // the block exchange + strategy (bitswap)
 	Namesys      namesys.NameSystem  // the name system, resolves paths to hashes
-	Reprovider   *rp.Reprovider      // the value reprovider system
+	Reprovider   *provider.Reprovider      // the value reprovider system
+	Provider     *provider.Provider  // the value provider system
 	IpnsRepub    *ipnsrp.Republisher
 
 	AutoNAT  *autonat.AutoNATService
@@ -324,6 +325,41 @@ func (n *IpfsNode) startLateOnlineServices(ctx context.Context) error {
 		return err
 	}
 
+	// Provider
+
+	// TODO: Specify strategy in cfg
+	strategy := provider.NewProvideAllStrategy(n.DAG)
+	tracker := provider.NewTracker(n.Repo.Datastore())
+	queue, err := provider.NewQueue("provider", ctx, n.Repo.Datastore())
+	if err != nil {
+		return err
+	}
+	n.Provider = provider.NewProvider(ctx, strategy, tracker, queue, n.Blockstore, n.Routing)
+	go n.Provider.Run()
+
+	// Reprovider - new
+
+	rq, err := provider.NewQueue("reprovider", ctx, n.Repo.Datastore())
+	if err != nil {
+		return err
+	}
+
+	reproviderInterval := kReprovideFrequency
+	if cfg.Reprovider.Interval != "" {
+		dur, err := time.ParseDuration(cfg.Reprovider.Interval)
+		if err != nil {
+			return err
+		}
+
+		reproviderInterval = dur
+	}
+
+	n.Reprovider = provider.NewReprovider(ctx, rq, tracker, reproviderInterval, n.Blockstore, n.Routing)
+	n.Reprovider.Run()
+
+	/*
+	// Reprovider - old
+
 	var keyProvider rp.KeyChanFunc
 
 	switch cfg.Reprovider.Strategy {
@@ -351,6 +387,7 @@ func (n *IpfsNode) startLateOnlineServices(ctx context.Context) error {
 	}
 
 	go n.Reprovider.Run(reproviderInterval)
+	*/
 
 	return nil
 }
--- a/core/coreapi/block.go
+++ b/core/coreapi/block.go
@@ -75,7 +75,7 @@ func (api *BlockAPI) Get(ctx context.Context, p coreiface.Path) (io.Reader, erro
 		return nil, err
 	}
 
-	api.node.Provider.Provide(rp.Cid())
+	//api.node.Provider.Provide(rp.Cid())
 
 	return bytes.NewReader(b.RawData()), nil
 }
@@ -129,7 +129,7 @@ func (api *BlockAPI) Stat(ctx context.Context, p coreiface.Path) (coreiface.Bloc
 		return nil, err
 	}
 
-	api.node.Provider.Provide(b.Cid())
+	//api.node.Provider.Provide(b.Cid())
 
 	return &BlockStat{
 		path: coreiface.IpldPath(b.Cid()),
--- a/core/coreapi/block.go
+++ b/core/coreapi/block.go
@@ -53,9 +53,13 @@ func (api *BlockAPI) Put(ctx context.Context, src io.Reader, opts ...caopts.Bloc
 		return nil, err
 	}
 
+//<<<<<<< HEAD
 	if settings.Pin {
 		api.pinning.PinWithMode(b.Cid(), pin.Recursive)
 	}
+//=======
+//	api.node.Provider.Provide(b.Cid())
+//>>>>>>> Add provider to ipfs and provide when adding/fetching
 
 	return &BlockStat{path: coreiface.IpldPath(b.Cid()), size: len(data)}, nil
 }
@@ -70,6 +74,8 @@ func (api *BlockAPI) Get(ctx context.Context, p coreiface.Path) (io.Reader, erro
 	if err != nil {
 		return nil, err
 	}
+
+	api.node.Provider.Provide(rp.Cid())
 
 	return bytes.NewReader(b.RawData()), nil
 }
@@ -122,6 +128,8 @@ func (api *BlockAPI) Stat(ctx context.Context, p coreiface.Path) (coreiface.Bloc
 	if err != nil {
 		return nil, err
 	}
+
+	api.node.Provider.Provide(b.Cid())
 
 	return &BlockStat{
 		path: coreiface.IpldPath(b.Cid()),
--- a/core/coreapi/block.go
+++ b/core/coreapi/block.go
@@ -53,13 +53,11 @@ func (api *BlockAPI) Put(ctx context.Context, src io.Reader, opts ...caopts.Bloc
 		return nil, err
 	}
 
-//<<<<<<< HEAD
 	if settings.Pin {
 		api.pinning.PinWithMode(b.Cid(), pin.Recursive)
 	}
-//=======
-//	api.node.Provider.Provide(b.Cid())
-//>>>>>>> Add provider to ipfs and provide when adding/fetching
+
+	api.provider.Provide(b.Cid())
 
 	return &BlockStat{path: coreiface.IpldPath(b.Cid()), size: len(data)}, nil
 }
@@ -75,7 +73,7 @@ func (api *BlockAPI) Get(ctx context.Context, p coreiface.Path) (io.Reader, erro
 		return nil, err
 	}
 
-	//api.node.Provider.Provide(rp.Cid())
+	api.provider.Provide(rp.Cid())
 
 	return bytes.NewReader(b.RawData()), nil
 }
@@ -129,7 +127,7 @@ func (api *BlockAPI) Stat(ctx context.Context, p coreiface.Path) (coreiface.Bloc
 		return nil, err
 	}
 
-	//api.node.Provider.Provide(b.Cid())
+	api.provider.Provide(b.Cid())
 
 	return &BlockStat{
 		path: coreiface.IpldPath(b.Cid()),
--- a/core/corehttp/gateway_handler.go
+++ b/core/corehttp/gateway_handler.go
@@ -489,6 +489,8 @@ func (i *gatewayHandler) putHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	i.node.Provider.Provide(newcid)
+
 	i.addUserHeaders(w) // ok, _now_ write user's headers.
 	w.Header().Set("IPFS-Hash", newcid.String())
 	http.Redirect(w, r, gopath.Join(ipfsPathPrefix, newcid.String(), newPath), http.StatusCreated)
@@ -565,6 +567,8 @@ func (i *gatewayHandler) deleteHandler(w http.ResponseWriter, r *http.Request) {
 
 	// Redirect to new path
 	ncid := newnode.Cid()
+
+	i.node.Provider.Provide(ncid)
 
 	i.addUserHeaders(w) // ok, _now_ write user's headers.
 	w.Header().Set("IPFS-Hash", ncid.String())
--- a/core/corehttp/gateway_handler.go
+++ b/core/corehttp/gateway_handler.go
@@ -489,7 +489,7 @@ func (i *gatewayHandler) putHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	i.node.Provider.Provide(newcid)
+	i.api.Provider().Provide(newcid)
 
 	i.addUserHeaders(w) // ok, _now_ write user's headers.
 	w.Header().Set("IPFS-Hash", newcid.String())
@@ -568,7 +568,7 @@ func (i *gatewayHandler) deleteHandler(w http.ResponseWriter, r *http.Request) {
 	// Redirect to new path
 	ncid := newnode.Cid()
 
-	i.node.Provider.Provide(ncid)
+	i.api.Provider().Provide(ncid)
 
 	i.addUserHeaders(w) // ok, _now_ write user's headers.
 	w.Header().Set("IPFS-Hash", ncid.String())
--- a/provider/provider.go
+++ b/provider/provider.go
@@ -1,0 +1,155 @@
+// Package provider implements structures and methods to provide blocks,
+// keep track of which blocks are provided, and to allow those blocks to
+// be reprovided.
+package provider
+
+import (
+	"context"
+	"fmt"
+	"gx/ipfs/QmR8BauakNcBa3RbE4nbQu76PDiJgoQgz8AJdhJuiU4TAw/go-cid"
+	bstore "gx/ipfs/QmS2aqUZLJp8kF1ihE5rvDGE5LvmKDPnx32w9Z1BW9xLV5/go-ipfs-blockstore"
+	"gx/ipfs/QmZBH87CAPFHcc7cYmBqeSQ98zQ3SX9KUxiYgzPmLWNVKz/go-libp2p-routing"
+	logging "gx/ipfs/QmcuXC5cxs79ro2cUuHs4HQ2bkDLJUYokwL8aivcX6HW3C/go-log"
+	"time"
+)
+
+var (
+	log = logging.Logger("provider")
+)
+
+const (
+	provideOutgoingWorkerLimit = 8
+	provideOutgoingTimeout     = 15 * time.Second
+)
+
+type Strategy func(context.Context, cid.Cid) <-chan cid.Cid
+
+// Provider announces blocks to the network, tracks which blocks are
+// being provided, and untracks blocks when they're no longer in the blockstore.
+type Provider struct {
+	ctx context.Context
+
+	// strategy for deciding which CIDs, given a CID, should be provided
+	strategy Strategy
+	// keeps track of which CIDs have been provided already
+	tracker *Tracker
+	// the CIDs for which provide announcements should be made
+	queue *Queue
+	// where the blocks live
+	blockstore bstore.Blockstore
+	// used to announce providing to the network
+	contentRouting routing.ContentRouting
+}
+
+func NewProvider(ctx context.Context, strategy Strategy, tracker *Tracker, queue *Queue, blockstore bstore.Blockstore, contentRouting routing.ContentRouting) *Provider {
+	return &Provider{
+		ctx:            ctx,
+		strategy:       strategy,
+		tracker:        tracker,
+		queue:          queue,
+		blockstore:   	blockstore,
+		contentRouting: contentRouting,
+	}
+}
+
+// Start workers to handle provide requests.
+func (p *Provider) Run() {
+	p.handleAnnouncements()
+}
+
+// Provide the given cid using specified strategy.
+func (p *Provider) Provide(root cid.Cid) error {
+	cids := p.strategy(p.ctx, root)
+
+	for cid := range cids {
+		isTracking, err := p.tracker.IsTracking(cid)
+		if err != nil {
+			return err
+		}
+		if isTracking {
+			continue
+		}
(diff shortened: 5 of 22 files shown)


pin/gc/gc.go
```diff
@@ -95,10 +101,13 @@ func GC(ctx context.Context, bs bstore.GCBlockstore, dstor dstore.Datastore, pn
 					if err != nil {
 						errors = true
 						output <- Result{Error: &CannotDeleteBlockError{k, err}}
-						//log.Errorf("Error removing key from blockstore: %s", err)
+						// log.Errorf("Error removing key from blockstore: %s", err)
 						// continue as error is non-fatal
 						continue loop
 					}
+					if err := cleanup.Cleanup(k); err != nil {
```
michaelavila: **NOTE** I'm in the process of removing this from the GC path and making clean up a part of reproviding.

Stebalien:
Preliminary comment: inline documentation would be *really* nice. While the code is nice and clean, I'd rather not guess what each subsystem is doing based on naming.

michaelavila:
@Stebalien understood. Do you need me to add anything immediately to clarify for your review? If not, then I'll add documentation of each new `provider` module when I get the `reprovider` done. Which is what I'm working on now.

Stebalien:
A few one-liners would be nice. It *looks* obvious but I don't want to incorrectly guess intention.

provider/provider.go
```diff
@@ -0,0 +1,165 @@
...
+// Start workers to handle provide requests.
+func (p *Provider) Run() {
+	go p.handleIncoming()
+	go p.handlePopulateOutgoing()
+	go p.handleOutgoing()
+}
+
+// Provider the given cid using specified strategy.
+func (p *Provider) Provide(root cid.Cid) {
+	cids := p.strategy(p.ctx, root)
```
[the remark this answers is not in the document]
michaelavila: This makes sense. I'm going to address this feedback last so we can discuss it in more detail after I get the other feedback incorporated. Briefly, though, my thoughts are:

Would we still need to allow a "global" providing strategy to be configured? How does that work with the ability to call provider strategies explicitly? Would we need to use an abstraction at all of the provide call sites that chooses the actual strategy to use based on the config?

provider/queue.go
```diff
@@ -0,0 +1,124 @@
...
+		if id < q.head {
+			q.head = id
+		}
+
+		if id > q.tail {
+			q.tail = id
+		}
+	}
+
+	if q.head == math.MaxUint64 {
```
hannahhoward: NB: Is prepare ever called outside the constructor? Asking cause it wasn't clear when I first looked at this how q.head became MaxUint64, which is necessary for the above loop to work, and then I realized it was set in the constructor. Seems like for clarity on your algorithm, setting head to MaxUint64 should happen shortly before the for loop.

Also, I might instead of calling this "prepare" I might make it a function call getHeadTailFromDatastore() (uint64, uint64), not have it mutate q.head and q.tail at all, and then just call it in constructor and use the values to initialize head and tail? I dunno.

Oh, also, I'm just curious, is there a case in which the queue on disk could become dis-contiguous -- i.e. head = 2, tail = 5 but there is no key written with key = 3 ? Just curious if that could happen (maybe due to the go routines in Provide) and whether that would result in anything bad.

provider/strategy.go
```diff
@@ -0,0 +1,23 @@
...
+	"gx/ipfs/QmR8BauakNcBa3RbE4nbQu76PDiJgoQgz8AJdhJuiU4TAw/go-cid"
+	ipld "gx/ipfs/QmcKKBwfz6FyQdHR2jsXrrF6XeSBXYL86anmWNewpFpoF5/go-ipld-format"
+	"gx/ipfs/QmdURv6Sbob8TVW2tFFve9vcEWrSUgwPqeqnXyvYhLrkyd/go-merkledag"
+)
+
+func NewProvideAllStrategy(dag ipld.DAGService) Strategy {
+	return func(ctx context.Context, root cid.Cid) <-chan cid.Cid {
+		cids := make(chan cid.Cid)
+		go func() {
+			cids <- root
```
hannahhoward: Perhaps not super important but I would make this channel write into a select statement in case context is cancelled, so this go routine doesn't possibly get left around.

provider/provider.go
```diff
@@ -0,0 +1,165 @@
...
+// Move CIDs from the providing queue to the outgoing channel
+func (p *Provider) handlePopulateOutgoing() {
+	for {
+		select {
+		case <-p.ctx.Done():
+			return
+		default:
+		}
+
+		p.lock.Lock()
```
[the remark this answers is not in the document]
michaelavila: This bit of code is probably going away once I figure out the blocking `Dequeue`, but FWIW the lock here is for the combination of `IsEmpty() + Dequeue()`, which I think is necessary in this particular configuration. Like I said though, I think this is going away.

provider/provider.go
```diff
@@ -0,0 +1,165 @@
...
+		if p.queue.IsEmpty() {
+			p.lock.Unlock()
+			time.Sleep(1 * time.Second)
+			continue
+		}
+		// TODO: We should probably not actually Dequeue() right here, or at
+		// least have a plan to replace the entry in the event that something
+		// goes wrong or the process is killed
+		key, err := p.queue.Dequeue()
```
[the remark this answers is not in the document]
michaelavila: I've posted a simple solution to this. Basically, now `Dequeue() (*Entry, error)` where `Entry` has a `Complete() error` that *MUST* be called by the caller of `Dequeue`. If `Complete` is never called then that _job_ is picked up again the next time the queue is restarted.

provider/provider.go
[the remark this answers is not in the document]
Stebalien: > Would we need to use an abstraction at all of the provide call sites that chooses the actual strategy to use based on the config?

Yes. The issue is that we'll likely want different default strategies for different commands (files, other data, pinning, etc.). For example, a "provide every directory/file root" strategy won't make sense as a default for the `ipfs dag` or `ipfs block` commands because *those* work directly with IPLD blocks (which may not be files).

However, it may make sense to introduce *another* service for this. Maybe? That is, something with a function like `GetStrategyFor(stuff) (Strategy, error)` where `stuff` includes things like:

* Whether or not we're pinning.
* The type of data we're handling (raw blocks, ipfs files, etc.).
* Maybe even the root CID?
* Maybe even the current command?

Really, I'm not sure what the right design is here.

provider/provider.go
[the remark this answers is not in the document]
Stebalien: Sounds reasonable.

provider/provider.go
```diff
@@ -0,0 +1,149 @@
...
+
+// Start workers to handle provide requests.
+func (p *Provider) Run() {
+	go p.handlePopulateOutgoing()
+	go p.handleOutgoing()
+}
+
+// Provider the given cid using specified strategy.
+func (p *Provider) Provide(root cid.Cid) {
+	cids := p.strategy(p.ctx, root)
```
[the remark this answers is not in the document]
Stebalien: See https://github.com/ipfs/go-ipfs/pull/5870/files#r247300973

(also, I believe that strategies are supposed to go async immediately.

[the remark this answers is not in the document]
michaelavila: resolving as applying strategy is being moved outside of provider

provider/provider.go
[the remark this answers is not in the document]
michaelavila: Understood, I'm processing this now.

provider/provider.go
```diff
@@ -0,0 +1,165 @@
...
+		case <-p.ctx.Done():
+			return
+		default:
+		}
+
+		p.lock.Lock()
+		if p.queue.IsEmpty() {
+			p.lock.Unlock()
+			time.Sleep(1 * time.Second)
+			continue
```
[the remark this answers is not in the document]
michaelavila: I achieved blocking here by having `Dequeue()` return a `<-chan *Entry`. The `Queue` has a `run()` loop that does the `Dequeue` work and so the context is passed in when creating the `Queue` in `core`.

core/core.go
```diff
@@ -317,6 +319,17 @@ func (n *IpfsNode) startLateOnlineServices(ctx context.Context) error {
...
 	}
 
+	// Provider
+
+	// TODO: Specify strategy in cfg
+	strategy := provider.NewProvideAllStrategy(n.DAG)
+	tracker := provider.NewTracker(n.Repo.Datastore())
+	queue := provider.NewQueue("provider", n.Repo.Datastore())
+	n.Provider = provider.NewProvider(ctx, strategy, tracker, queue, n.Routing)
+	go n.Provider.Run()
```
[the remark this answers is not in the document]
michaelavila: Thanks! This code has all changed quite a bit. I think the `go` here is good again.

provider/strategy.go
michaelavila: Thanks! I went through and put all the necessary contexts in place.

michaelavila:
@hannahhoward all your feedback is accounted for, thanks!

provider/provider.go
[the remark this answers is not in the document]
michaelavila: This did go away. Now we have just one run loop that is doing all the work when a channel is available to be read from and so the lock was able to be removed.

provider/reprovider.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+// Add all the cids in the tracker to the reprovide queue
+func (rp *Reprovider) Reprovide() error {
+	cids, err := rp.tracker.Tracking(rp.ctx)
+	if err != nil {
+		return err
+	}
+	for c := range cids {
+		if err := rp.queue.Enqueue(c); err != nil {
```
magik6k: Could we **not** remove entries from the queue after use, and only reset some counters/pointers there (to reduce stress on datastore)?

core/commands/tar.go
```diff
@@ -56,6 +56,9 @@ represent it.
 
 		c := node.Cid()
 
+		nd.Provider.Provide(c)
+
+		// TODO: why is this here?
```
magik6k: This is fixed in master

core/core.go
```diff
@@ -21,61 +21,61 @@ import (
 	"time"
 
 	version "github.com/ipfs/go-ipfs"
-	rp "github.com/ipfs/go-ipfs/exchange/reprovide"
-	filestore "github.com/ipfs/go-ipfs/filestore"
-	mount "github.com/ipfs/go-ipfs/fuse/mount"
-	namesys "github.com/ipfs/go-ipfs/namesys"
+	"github.com/ipfs/go-ipfs/filestore"
+	"github.com/ipfs/go-ipfs/fuse/mount"
+	"github.com/ipfs/go-ipfs/namesys"
```
magik6k: Imports need fixing

core/coreapi/block.go
```diff
@@ -48,6 +48,8 @@ func (api *BlockAPI) Put(ctx context.Context, src io.Reader, opts ...caopts.Bloc
 		return nil, err
 	}
 
+	api.node.Provider.Provide(b.Cid())
```
magik6k: You shouldn't use `api.node` outside `api.WithOption` to make sure that global options are applied correctly. `Provider` should be added to the api struct in https://github.com/ipfs/go-ipfs/blob/master/core/coreapi/coreapi.go#L47 and https://github.com/ipfs/go-ipfs/blob/master/core/coreapi/coreapi.go#L156

provider/reprovider.go
```diff
@@ -0,0 +1,137 @@
...
+
+func (rp *Reprovider) handleTriggers() {
+	// dont reprovide immediately.
+	// may have just started the daemon and shutting it down immediately.
+	// probability( up another minute | uptime ) increases with uptime.
+	after := time.After(time.Minute)
+	var done doneFunc
+	for {
+		if rp.tick == 0 {
+			after = make(chan time.Time)
```
hannahhoward: This is just a channel that always blocks right? You can just use `nil` if you want.

provider/reprovider.go
```diff
@@ -0,0 +1,137 @@
...
+		if err := rp.queue.Enqueue(c); err != nil {
+			log.Warningf("unable to enqueue cid: %s, %s", c, err)
+			continue
+		}
+	}
+	return nil
+}
+
+// Trigger causes a reprovide
+func (rp *Reprovider) Trigger(ctx context.Context) error {
```
hannahhoward: 2 questions:

1. What is the use case for Trigger? It doesn't appear to be used anywhere? Do we have a way to manually trigger a reprovide? (sorry just not clear cause I haven't interacted with this system much)

2. What happens if Trigger gets called with a timed-out reprovide in progress? (or another manually triggered reprovide) It appears that the current implementation (I think) would cause a reprovide as soon as the current reprovide ended. Is this the intended behavior?

michaelavila: 1. Yes, there is a command: `ipfs bitswap reprovide`

michaelavila: 2. Right now, that's true. There are more changes coming that address this (potentially).

provider/reprovider.go
michaelavila: Probably! I moved this code over, let me try that out.

michaelavila: Updated.

provider/queue.go
```diff
@@ -0,0 +1,212 @@
...
+	if err != nil {
+		return nil, err
+	}
+	q := &Queue{
+		name: name,
+		ctx: ctx,
+		head: head,
+		tail: tail,
+		lock: sync.Mutex{},
+		datastore: datastore,
```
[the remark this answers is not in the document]
michaelavila: Works nicely, thanks.

provider/queue.go
```diff
@@ -0,0 +1,212 @@
...
+
+	return entry, nil
+}
+
+func (q *Queue) queueKey(id uint64) ds.Key {
+	return ds.NewKey(queuePrefix(q.name) + strconv.FormatUint(id, 10))
+}
+
+func queuePrefix(name string) string {
+	return "/" + name + "/queue/"
```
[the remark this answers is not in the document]
michaelavila: Works nicely, thanks.

provider/reprovider.go
```diff
@@ -0,0 +1,137 @@
...
+
+	select {
+	case <-rp.ctx.Done():
+		return context.Canceled
+	case <-ctx.Done():
+		return context.Canceled
+	case rp.trigger <- done:
+		select {
+		case <-progressCtx.Done():
+			return err
```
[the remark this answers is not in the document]
michaelavila: Right now I think it will just return `nil`. Perhaps it should be something like this:

```go
        ...
	case rp.trigger <- done:
		select {
		case <-rp.ctx.Done():
			return rp.ctx.Err()
		case <-progressCtx.Done():
			if err == nil {
				return err
			} else {
				return ctx.Err()
			}
		}
	}
```

provider/reprovider.go
```diff
@@ -0,0 +1,137 @@
...
+
+import (
+	"context"
+	"gx/ipfs/QmS2aqUZLJp8kF1ihE5rvDGE5LvmKDPnx32w9Z1BW9xLV5/go-ipfs-blockstore"
+	"gx/ipfs/QmZBH87CAPFHcc7cYmBqeSQ98zQ3SX9KUxiYgzPmLWNVKz/go-libp2p-routing"
+	"time"
+)
+
+var (
+	reprovideOutgoingWorkerLimit = 8
```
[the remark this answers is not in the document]
michaelavila: I defer to you guys on the larger change, but it's certainly easy enough to make this change here

provider/queue.go
```diff
@@ -0,0 +1,212 @@
...
+	var err error
+	for {
+		// TODO: if head >= tail, then return nil, empty queue error
+		select {
+		case <-q.ctx.Done():
+			return nil, nil
+		default:
+		}
+		nextKey = q.queueKey(q.head)
+		value, err = q.datastore.Get(nextKey)
```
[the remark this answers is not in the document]
michaelavila: @magik6k I'm processing this and would like to educate myself. Can you point me to documentation on the issue you're worried about?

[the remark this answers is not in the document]
magik6k: There isn't really any central place documenting this. The main issue is that in some cases `datastore.Rm` only marks a record for deletion and doesn't free any disk-space up (this happens when `datastore.CollectGarbage` is called (which is currently only done in `ipfs repo gc`)).

One issue which sort-of explains how it works in badger's case is https://github.com/ipfs/go-ipfs/issues/4824, Badger docs are also a good source of information - https://github.com/dgraph-io/badger#garbage-collection

core/core.go
magik6k: yep, we decided some time ago that it's better to keep them even if redundant for consistency

michaelavila: Understood, thanks.

provider/reprovider.go
eingenito: @magik6k can you explain the behavior that you're envisioning and how you think it might impact badger (or I guess another datastore?). I get that we're going to be inserting a lot of tiny records into the datastore and then quickly deleting them. What's the bad effect of doing this? Is it that it fills up the datastore to the configured limit before triggering GC? Or is it that GC becomes way more expensive because badger is slow at collecting lots of little entries?

We could definitely configure a separate datastore as we discussed in the weekly yesterday and we could even call GC on it periodically as part of the providing loop. Would this mitigate the issue?

core/corehttp/gateway_handler.go
```diff
@@ -499,6 +499,8 @@ func (i *gatewayHandler) putHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	i.node.Provider.Provide(newcid)
+
```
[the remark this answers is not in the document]
michaelavila: @eingenito all those paths are covered by the provider calls in `CoreAPI`. The reason you see some provides in `gateway_handler` is because, at the time of writing this code, there were some aspects of gateway_handler that were not covered by coreapi.

provider/reprovider.go
michaelavila: @magik6k I am also curious about the answer to @eingenito's question. After doing research yesterday, I didn't get the sense that we should be overly concerned about this particular usage of badger. Which is probably just me not having all the information you have.

magik6k: I'll start with saying that it's something that definitely needs more testing, and I'll try to get some numbers from go-ds-bench soon(-ish).

Badger keeps data in value logs, separately from LSM tree. I'm not sure about impact of moving lots of data through LSM tree, but my main concern is about the value logs because badger has a rather specific approach to GC. Basically when short-lived data is mixed with it may cause few things:
* Lots of expensive vlog rewrites (I'm not sure how expensive, needs benchmarks)
* Wasted space if short-lived data is mixed with long-lived data in unfortunate proportions (short lived data may never be freed since vlog reclaimable space won't meet one few thresholds)

I'd recommend following the implementation of `RunValueLogGC` and mainly `vlog.doRunGC` - https://github.com/dgraph-io/badger/blob/b9e379e59eac9b828431a6c8f7a553ae38ab88eb/value.go#L1135

michaelavila: @magik6k I think I understand your specific concern better. I believe there's a rule in badger that a value log file needs at least 15% of the file to be eligible for removal for it to do anything with that file. So with just the right proportion of long-lived data to short-lived data in a single file, badger will not clean it up. On the flip side, if the proportion is high in the other direction then those files will experience a lot of churn and therefore need to be cleaned up regularly. 

I think, like you said, we need to get data and the cost of these value log rewrites. First, I'm going to finish integrating the rest of the feedback I've received.

michaelavila: @Stebalien, @magik6k, @Kubuxu: this work will soon be blocked on getting this issue resolved and so I want to get a few more people involved.

My questions are:
1. Do the potential performance implications, in the grand scheme of things, necessitate me taking a different approach?
2. If we _do not_ have enough data to make this decision right now, does this work get blocked until the data is obtained and the implications assessed?

Just to get everyone on the same page: implementation wise, the `Queue` adds entries to `Datastore` and removes the entry when the work done for that entry is complete. The entries are `cids` and the work is providing them via content routing.

The concern, broadly, is due to the way `Badger` stores its data. There is the potential for often rewriting and processing badger's value log files as data is added/removed. Additionally, in the right proportions, data that is removed may never actually be removed (because it's mixed in the value log file with data that hasn't been removed). But, IMO the primary concern is the former issue, the reprocessing of value log files.

Can we hash this out here?

@magik6k thanks for bringing all of this up and please correct me if this is not the gist of where we are

magik6k: > 2. If we _do not_ have enough data to make this decision right now, does this work get blocked until the data is obtained and the implications assessed?

We could try to work-around this.

I might have missed some bits of prior conversation on this, but what is the reason we have to persist this queue instead of just relying on reprovider? Nodes can shutdown for long periods of time, so old provide records will be expired anyways, forcing us to do full reprovide either way.

Kubuxu: @magik6k we persist the queue because it can get very big (RAM size big).

michaelavila: @Kubuxu thanks for chiming in! (I was out yesterday). @magik6k yes, the reason I was asked to put the queue on disk was due to its being unbounded in RAM (now it's just unbounded on disk).

magik6k: IMO if we need to provide so many CIDs that we can't fit them in reasonable amounts of RAM, we're probably trying to provide way too much.

Assuming 64 bytes per entry, we could fit 16k CIDs per MiB of ram, and while I don't know what were the numbers from simulations @Kubuxu ran in London, I know that 16k provider records is a lot of provider records.

Kubuxu: Yes, but we don't have probabilistic providers/providers strategies. We used to have unlimited provider queue, then we added backpressure because we were running out of memory.

---

My vote is: let's worry about performance/utilisation of badger when it becomes a problem or when we will be switching to it by default.

core/core.go
```diff
@@ -317,6 +318,41 @@ func (n *IpfsNode) startLateOnlineServices(ctx context.Context) error {
...
+		}
+
+		reproviderInterval = dur
+	}
+
+	n.Reprovider = provider.NewReprovider(ctx, rq, tracker, reproviderInterval, n.Blockstore, n.Routing)
+	n.Reprovider.Run()
+
+	/*
+	// Reprovider - old
```
michaelavila: Ignore this. It will get deleted.

core/coreapi/block.go
michaelavila: This works. I had put this all in coreapi to begin with, but we opted to move it out at the time. Since this work now also covers reprovide, I'd like to move the `ipfs bitswap reprovide` command to be something more like `ipfs provider reprovide` which I think aligns with the goals of coreapi.

provider/reprovider.go
Stebalien: So, @magik6k's right that we can't really provide more than a on the order of a million records per day and we can easily fit that in memory. However, I also agree with @Kubuxu for the moment: we don't have any fancy provider strategies yet so this queue will grow large (although we _could_ dump the queue and rely on the reprovider when that happens). Really, I have no strong opinions on this.

> so old provide records will be expired anyways, forcing us to do full reprovide either way.

We've discussed avoiding unnecessary full reprovides by remembering our reprovide position/time across restarts.

michaelavila: I'm wondering if we can put the provider work into on-disk queues for now and later revisit those queues. The approach taken in this PR does, as I understand it, help with some of the issues we've seen regarding gateways and their unbounded in-memory provider queues. Also, the approach taken in this PR opens the door to implementing more robust and varied providing strategies, which will help greatly with our handling of nodes with lots of content (one of our current ipfs goals).

After this work is in, I think we can start working out important changes to the details while in parallel beginning work on the larger goal we've set for ourselves (probabilistic strategies).

My concern with abandoning the current approach is just that we've spent a lot of time working out the details and worry about it extending even more further than we would like. If the shape of what we've come up with will allow us to move forward with the more impactful changes we are trying to get to, while allowing us to still tweak the details of what we've done so far, then I'm a big vote for moving forward with what we're currently doing.

If we've found a critical flaw, then I obviously wouldn't advocate moving forward. I just want to sort out which things are critical for this work to get merged. I really have no issue going back and rethinking the approach, I just want to get my thoughts out there as I'm conscious of the amount of time we've spent working out what we've got here.

Last note, there are still a few things to do on this PR so I'm not proposing we merge it as is, just that I focus my work on those things that will prevent a merge (again maybe this is one of those thing) while filing issues for things we want to revisit or consider in more depth (but maybe it could be one of these).

Stebalien: @michaelavila that sounds like a good approach. If this _does_ cause issues, we can always revert or change our minds later.

provider/provider.go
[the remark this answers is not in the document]
michaelavila: @Stebalien It sounds like the plan is to move the provide calls to the call-sites and to choose an appropriate providing strategy for each location. However, there is currently the ability to configure a providing strategy, which we've expressed the desire to preserve. I'm trying to figure out exactly what those configurations would do.

I think a reasonable approach is to determine all the places where a "provide all" currently takes place (e.g. `ipfs add`) and to allow those locations to be configured by the user's provider strategies. This makes the system behave a little different from the way it behaves right now, which I think is unavoidable, but achieves the larger goal of allowing a limited strategy to be used in place where there's the potential to provide too many blocks.

Any thoughts?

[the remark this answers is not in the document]
michaelavila: Just to clarify, in the locations where a "provide all" does not make sense, we just always use whatever providing strategy we decide makes the most sense in that location.

provider/queue.go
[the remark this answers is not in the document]
michaelavila: Conversation continued here: https://github.com/ipfs/go-ipfs/pull/5870#discussion_r254452479

provider/provider.go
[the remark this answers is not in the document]
Stebalien: > I think a reasonable approach is to determine all the places where a "provide all" currently takes place (e.g. ipfs add) and to allow those locations to be configured by the user's provider strategies. 

Exactly.

> This makes the system behave a little different from the way it behaves right now, which I think is unavoidable

Correct. Really, we can probably deprecate the current reprovider strategy system and write a config migration.

---

For example, we could (**eventually**) end up supporting a provider strategy spec like the following:

```js
{
  // Default strategy.
  "default": {
    // all objects.
    "all": {
      "strategy": "root",
      // put it in the queue, announce once, but don't record or reprovide
      // We probably won't support this out of the box...
      "when": "once",
    }
  }
  // For `ipfs add --pin=false`. That is, data _we've_ chosen to add (but not pin).
  "add": {
    "all": {
      "strategy": "root",
    },
    "unixfs": {
      "strategy": "files",
      "when": "once",
    }
  },
  // Either `ipfs pin /ipfs/...` or `ipfs add ...`
  "pin": {
    // fallback.
    // In this case, it only applies ot `ipfs pin /ipld/Qm...`
    "default": {
      "strategy": "all",
    },
    // Applies to things like `ipfs pin /ipfs/Qm...`
    "unixfs": {
      "strategy": "files",
    },
  },
}
```

(however, the user should be able to override the strategy when adding/pinning data.

michaelavila:
I'm just noticing this is happening, putting here so I don't forget:

```shell

λ cmd/ipfs/ipfs init
initializing IPFS node at /Users/me/.ipfs
generating 2048-bit RSA keypair...done
peer identity: QmWotgPWydYsdr9m5UBj7Zyy9dLhX2ecQkXLqRB8x3ZCwb
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x16b37e6]

goroutine 1 [running]:
github.com/ipfs/go-ipfs/provider.(*Provider).Provide(0x0, 0xc0000e04e0, 0x22, 0x20b4460, 0xc00017cee0)
	/Users/me/go/src/github.com/ipfs/go-ipfs/provider/provider.go:62 +0x26
github.com/ipfs/go-ipfs/core/coreapi.(*ObjectAPI).New(0xc000182280, 0x20a0d20, 0xc0000c68a0, 0xc00016e0c0, 0x1, 0x1, 0x0, 0x0, 0xc000184500, 0xc0000d0be0)
	/Users/me/go/src/github.com/ipfs/go-ipfs/core/coreapi/object.go:58 +0x14d
github.com/ipfs/go-ipfs/assets.addAssetList(0xc00000c1e0, 0x2a872c0, 0x7, 0x7, 0x0, 0x0, 0xc00004a000, 0x17)
	/Users/me/go/src/github.com/ipfs/go-ipfs/assets/assets.go:55 +0x178
github.com/ipfs/go-ipfs/assets.SeedInitDocs(...)
	/Users/me/go/src/github.com/ipfs/go-ipfs/assets/assets.go:36
main.addDefaultAssets(0x20808e0, 0xc000010010, 0xc0000d80c0, 0xf, 0x0, 0x0)
	/Users/me/go/src/github.com/ipfs/go-ipfs/cmd/ipfs/init.go:218 +0x1ab
main.doInit(0x20808e0, 0xc000010010, 0xc0000d80c0, 0xf, 0x2aa5d00, 0x800, 0x0, 0x0, 0x0, 0xc000184280, ...)
	/Users/me/go/src/github.com/ipfs/go-ipfs/cmd/ipfs/init.go:167 +0x2d1
main.glob..func2(0xc0002d8c40, 0x20a1060, 0xc0000b8900, 0x207f500, 0xc0000b88a0, 0xc000123d28, 0x1af0afe)
	/Users/me/go/src/github.com/ipfs/go-ipfs/cmd/ipfs/init.go:113 +0x1de
github.com/ipfs/go-ipfs-cmds.(*executor).Execute(0xc00016e078, 0xc0002d8c40, 0x20a1060, 0xc0000b8900, 0x207f500, 0xc0000b88a0, 0xc0000b8900, 0x0)
	/Users/me/go/pkg/mod/github.com/ipfs/go-ipfs-cmds@v0.0.1/executor.go:81 +0x15d
github.com/ipfs/go-ipfs-cmds/cli.Run(0x20a0c60, 0xc0000da580, 0x2a87bc0, 0xc0000f4000, 0x2, 0x2, 0xc0000e2000, 0xc000010010, 0xc0000e2008, 0x1f37090, ...)
	/Users/me/go/pkg/mod/github.com/ipfs/go-ipfs-cmds@v0.0.1/cli/run.go:138 +0xa69
main.mainRet(0x0)
	/Users/me/go/src/github.com/ipfs/go-ipfs/cmd/ipfs/main.go:179 +0x421
main.main()
	/Users/me/go/src/github.com/ipfs/go-ipfs/cmd/ipfs/main.go:85 +0x22
```

michaelavila:
Superseded by https://github.com/ipfs/go-ipfs/pull/6141

PR #5878, opened 2018-12-28
[WIP] WebAssembly support

magik6k:
Part of https://github.com/ipfs/go-ipfs/issues/5694

Putting this in a PR for visibility

Currently this code is POC quality, mostly to test whether it may be actually viable to run go-ipfs in the javascript land (mostly browsers, I don't think running wasm go-ipfs in node makes much sense)

If we actually want to turn this into something usable, there will be a fair bit of cleaning up to do, but it's at lest a staring point.

How to run this thing:
* `go get github.com/libp2p/go-reuseport`, cd into it, `git checkout feat/wasm`, cd back into go-ipfs
* `gx-go link QmXD921xzL9EDRpD6gRm1cb7Khm8VEpZ3NT3nPK7uTX6Fq` (current go-reuseport hash)
* `make misc/wasm/serve`
* Go to http://127.0.0.1:8000, wait a bit for wasm blob to load, press 'Run', wait a bit more
* Type something into text box and press enter
* If you adjust bootstrap list to have your node in it, you should be able to fetch the resulting hash

### DIFF
--- a/Rules.mk
+++ b/Rules.mk
@@ -53,6 +53,9 @@ include $(dir)/Rules.mk
 dir := filestore/pb
 include $(dir)/Rules.mk
 
+dir := misc/wasm
+include $(dir)/Rules.mk
+
 
 # -------------------- #
 #   universal rules    #
--- a/misc/wasm/.gitignore
+++ b/misc/wasm/.gitignore
@@ -1,0 +1,1 @@
+ipfs.wasm
--- a/misc/wasm/Rules.mk
+++ b/misc/wasm/Rules.mk
@@ -1,0 +1,17 @@
+include mk/header.mk
+
+# TODO: target disabling all incompatible plugins
+
+$(d)/ipfs.wasm:
+	GOOS=js GOARCH=wasm $(GOCC) build -tags="nofuse purego" -ldflags="-X "github.com/ipfs/go-ipfs".CurrentCommit=47e9466ac" -o "$@" "github.com/ipfs/go-ipfs/misc/wasm"
+
+.PHONY: $(d)/ipfs.wasm
+
+$(d)/serve:	$(d)/ipfs.wasm
+	go get github.com/shurcooL/goexec
+	@echo 'Listening on http://127.0.0.1:8000'
+	@(cd $(@D) && goexec 'http.ListenAndServe("127.0.0.1:8000", http.FileServer(http.Dir(".")))')
+
+.PHONY: $(d)/serve
+
+include mk/footer.mk
--- a/misc/wasm/Rules.mk
+++ b/misc/wasm/Rules.mk
@@ -2,7 +2,7 @@ include mk/header.mk
 
 # TODO: target disabling all incompatible plugins
 
-$(d)/ipfs.wasm:
+$(d)/ipfs.wasm: deps
 	GOOS=js GOARCH=wasm $(GOCC) build -tags="nofuse purego" -ldflags="-X "github.com/ipfs/go-ipfs".CurrentCommit=47e9466ac" -o "$@" "github.com/ipfs/go-ipfs/misc/wasm"
 
 .PHONY: $(d)/ipfs.wasm
--- a/misc/wasm/index.html
+++ b/misc/wasm/index.html
@@ -1,0 +1,52 @@
+<!doctype html>
+<!--
+Copyright 2018 The Go Authors. All rights reserved.
+Use of this source code is governed by a BSD-style
+license that can be found in the LICENSE file.
+-->
+<html>
+
+<head>
+	<meta charset="utf-8">
+	<title>Go wasm</title>
+</head>
+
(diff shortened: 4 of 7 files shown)


sriharrsha:
Well, 39MB is super huge. for Binary WASM of Go-IPFS. Is there any alternative to improve upon decreasing the size.

magik6k:
There are some ways to compress the code ~10x, I'm not sure how portable wasm though

rjy7wb:
> Well, 39MB is super huge. for Binary WASM of Go-IPFS. Is there any alternative to improve upon decreasing the size.

doesn't wasm have streaming compile etc? its an issue but not a "huge" one

rjy7wb:
is the output  hosted anywhere, maybe put the page on ipfs

whyrusleeping:
@magik6k this is really awesome. Do you intend to work on it more? or maybe try to get it merged in some way?

magik6k:
> Do you intend to work on it more?

I currently have a rather long list of things to do, but I'd like to get this at least to the stage where we could see how well it works/could work (say, by comparing vs js-ipfs)

whyrusleeping:
I managed to get the bundle down to 27mb by removing some of the larger  dependencies we don't use here, like quic and btcec, and a bunch of other things. There's a lot more that can be removed, like some hash functions, the entire commands lib and ipfs commands, and anything related to a disk backed datastore (badger, leveldb, etc).

This is fun :D

whyrusleeping:
spent some more time on this, its really fun.

This tool is really useful: https://github.com/jondot/goweight

I've gotten the uncompressed wasm blob now down to 15.5MB, 3.1MB gzipped. I did this by using hectors ipfs-lite repo, and trimming a bunch of things, notably eradicating any and all imports of `net/http`from the dependency tree. This was difficult, but i did it by cutting out the go-libp2p-nat package (not needed in the browser) and removing all opentracing/opencensus stuff from go-log. Badger also got removed in favor of the in-memory datastore (should try and write something that uses javascript native indexedDB stuff). Additionally, bitswap was having a dependency on both "testing" and the libp2p mock net, i removed this dependency by just deleting the 'testutils.go' file in bitswap. We should make sure that only gets compiled for test builds.

A few things that stick out now:

- gogo protobuf contributes over 3MB to the uncompressed blob size. This seems absurd, protobuf marshaling and unmarshaling really shouldnt take up that much space. Maybe we can swap this out with a more 'trim' protobuf lib? (perhaps just write one?)
- The go runtime is 2.4MB. Eh, what can we do about it? I guess just bug the go team.
- the 'net' package, which shouldnt actually get used in wasm, contributes one whole MB. I think we just depend on some types from here. For wasm compiles, we should try and avoid any and all net function usage.
- the json package is contributing half a MB. Seems like something we could use javascript native stuff for? also why is this a 'core' ipfs dependency?
- my cbor fork is still being used apparently, we should switch over to refmt entirely, will save half a MB
- archive/tar is getting pulled in somehow. This is around 400kb.


In any case, i don't intend for the hack i'm playing with to get used, but what this points out is that a reasonably sized go-ipfs wasm blob is very practical. I think that with some more work, and improvements from the go team, we can get things to be very very small!

whyrusleeping:
One very short term thing that I think would be really cool:

We can compile the go yamux package to wasm, and use it in js to make communication with go nodes easier! (cc @daviddias :) )

Stebalien:
> gogo protobuf contributes

Golang protobuf is 1.8MiB.

> cbor

https://github.com/libp2p/go-libp2p-pnet/pull/26

bmiller59:
@whyrusleeping @magik6k Is there any update on this? I see last activity was about a year ago. What is blocking this getting merged in and what do you see as the path forward? Thanks!

Stebalien:
The @0xProject team has added wasm support to the websocket transport (https://github.com/libp2p/go-ws-transport/). Other than that, this is just waiting for somebody to pick it up.

unicomp21:
Amazing idea, would be cool for both browser and nodejs, especially if typescript bindings could be generated.

Thaina:
Was this idea retracted or having any roadblock?

lidel:
A lot has changed since this 2018 POC. The `gx/ipfs` paths are gone, the dependency graph looks very different, and the IPFS-in-browser story has consolidated around [Helia](https://github.com/ipfs/helia) (JS-native, an order of magnitude smaller than 100+ MB of Go-via-wasm). For the server-side wasm case, `GOOS=wasip1 GOARCH=wasm` is the realistic kubo target.

Posted an updated 2026 summary at https://github.com/ipfs/kubo/issues/5694#issuecomment-4381173557. It lists the four kubo-side changes that compile `cmd/ipfs` to a working `wasip1/wasm` binary, and explains why `GOOS=js` stays out of scope.

If anyone is interested in any of this, please comment in #5694 (or open new PR based on modern code).

Issue #5138, opened 2018-06-19
Add `user-agent` to default list of `Access-Control-Allow-Headers`


sohkai:
#### Version information:

**Mostly to do with public gateways**, but issue persists on recent versions:

```
go-ipfs version: 0.4.13-
Repo version: 6
System version: amd64/darwin
Golang version: go1.9.2
```

#### Type:
Enhancement

#### Description:

When sending requests across CORS, it looks like newer browsers are asking for `user-agent` in `Access-Control-Allow-Headers`, and failing fetches without it, e.g.:

![image](https://user-images.githubusercontent.com/4166642/41591398-df35590c-73b9-11e8-8725-9b0d1669602a.png)

It would be really nice if `user-agent` was included by default in the CORS settings. Also, not sure if setting it twice is the same as a list, but it seems like it [should be a list](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Headers)?

[See a similar API issue (Github's)](https://github.com/octokit/rest.js/issues/817)

bpierre:
In the meantime, it should be possible to update the headers using the config file: https://github.com/ipfs/go-ipfs/blob/7e8f6c96047e1a94c1c892aa50e1c75e01b7fda2/docs/config.md#gateway

ivan386:
@bpierre Yes it possible but this setting must be set on all gateways by default. 

Firefox block request by XMLHttpRequest to all public gateways without it.

lidel:
Big yes for allowing for `user-agent` customization during `fetch`, we want to be compatible with vendors following the spec (https://github.com/whatwg/fetch/issues/37) :+1:

As for the cleanup of defaults:

**tl;dr** I think we should go with (2) AND while at it make sure we will be able to do some cleanup described below.

----

I've looked at our headers recently and started tracking related issues at https://github.com/ipfs/in-web-browsers/issues/132.

As you can see in "go-ipfs v0.4.18 defaults" section of  https://github.com/ipfs/in-web-browsers/issues/132#issuecomment-442039654 we already have "hardcoded" defaults for headers such as `Access-Control-Expose-Headers`.  

[...]

PR #5888, opened 2019-01-03
content-type to allowedHeadersArr

popmanhe:
When using browser(chrow, firefox) to request a resource from IPFS, browser uses a preflight request with content-type header. So without content-type in the allowed list, the request will fail.

### DIFF
--- a/core/corehttp/gateway_handler.go
+++ b/core/corehttp/gateway_handler.go
@@ -190,6 +190,7 @@ func (i *gatewayHandler) getOrHeadHandler(ctx context.Context, w http.ResponseWr
 		"Content-Range",
 		"X-Chunked-Output",
 		"X-Stream-Output",
+		"content-type"
 	}
 
 	var allowedHeaders = strings.Join(allowedHeadersArr, ", ")


Stebalien:
This shouldn't be necessary according to the second paragraph here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Headers

popmanhe:
> 
> 
> This shouldn't be necessary according to the second paragraph here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Headers

I am developing a web site. and from that web site, there is a request that links to a IPFS gateway. So it is a CORS request. But without content-type added to the allowed header in IPFS config, the request fails. 

The following message is from Chrome. It's the same in firefox.

Access to fetch at '**http://localhost:8080/ipfs/QmYwdHKn3bGePULMqhqmTiM2JjQPEc8sYwEuDYDaa2DP6r**' from origin '**http://localhost:3000**' has been blocked by CORS policy: Request header field content-type is not allowed by Access-Control-Allow-Headers in preflight response.

core/corehttp/gateway_handler.go
```diff
@@ -190,6 +190,7 @@ func (i *gatewayHandler) getOrHeadHandler(ctx context.Context, w http.ResponseWr
 		"Content-Range",
 		"X-Chunked-Output",
 		"X-Stream-Output",
+		"content-type"
```
Stebalien:
```suggestion
		"Content-Type"
```

(to be consistent)

Stebalien: Also needs to end with a comma.

Stebalien:
Ah, I see. That only applies to form data and text. I assume you're trying to write to a writable gateway, right?

@lidel or @hsanjuan you both know more about this than I do, does this look correct? That is, should users be able to POST with arbitrary content types to a writable gateway. As far as I can tell, the answer is yes.

popmanhe:
No. It's just a GET request. Just it's a CORS request.
I am using brower's native fetch method to get the content from IPFS gateway
 const response = await fetch(url, {
            headers,
            mode: 'cors',
            method: "GET"
        });

Stebalien:
Wait, are you *passing* a Content-Type header?

hsanjuan:
Also, might kill this one too https://github.com/ipfs/go-ipfs/issues/5138

Stebalien:
I'm going fix this all at once in #5893.

Stebalien:
@popmanhe thanks for finding this.

PR #5893, opened 2019-01-04
gateway: fix CORs headers

### LINKED ISSUE #5138: Add `user-agent` to default list of `Access-Control-Allow-Headers`

### LINKED ISSUE #5892: Allow/Expose-Headers headers overwrite user-provided ones
#### Description: 

Per https://github.com/ipfs/go-ipfs/blob/master/core/corehttp/gateway_handler.go#L197  we set `Access-Control-Allow-Headers` and `Access-Control-Expose-Headers` AFTER `i.addUserHeaders(w)`. 

Per https://github.com/ipfs/go-ipfs/blob/master/docs/config.md#gateway the users are told that they can customize these headers.

Per https://github.com/ipfs/go-ipfs-config/blob/e6bdf3c437bc6820f8d7e0cb337bf4b900fd1f1b/init.go#L67 the default configuration sets `Access-Control-Allow-Headers`, but whatever the user puts here they are overwritten (?).

Am I missing something or should `i.addUserHeaders(w)` happen afterwards setting the defaults?

cc. @lidel

Stebalien:
* fixes #5138 -- always add user-agent to access-control-allow-headers.
* fixes #5888 -- same with content-type.
* fixes #5892 -- extend user-provided headers instead of overriding them.

TODO:

* [ ] Sharness tests.
* [ ] Remove header defaults from the config.


Stebalien:
cc @lidel, @popmanhe.

core/corehttp/commands.go
```diff
@@ -67,12 +67,18 @@ func addHeadersFromConfig(c *cmdsHttp.ServerConfig, nc *config.Config) {
...
-	c.Headers = make(map[string][]string, len(nc.API.HTTPHeaders))
+	c.Headers = make(map[string][]string, len(nc.API.HTTPHeaders)+1)
 
 	// Copy these because the config is shared and this function is called
 	// in multiple places concurrently. Updating these in-place *is* racy.
 	for h, v := range nc.API.HTTPHeaders {
-		c.Headers[h] = v
+		h = http.CanonicalHeaderKey(h)
+		switch h {
+		case cmdsHttp.ACAOrigin, cmdsHttp.ACAMethods, cmdsHttp.ACACredentials:
```
Stebalien: We already skip these in the commands lib so this probably isn't necessary.

Stebalien:
> While we are fixing CORS, it may be worth addressing a nuance related to #5138, #5892 and ipfs/in-web-browsers#132 (comment):

We should fix this but it may require a bit of refactoring to get it right (beyond the scope of this PR).

daviddias:
@Stebalien is it the right time to ask that the IPFS config.json gets organized by system rather than a mix of everything. By this I mean, have a

```
{
  libp2p: {
    // All things libp2p
  },
  Gateway: {
    // All things gateway
  },
  RemoteAPI: {
    All things RemoteAPI
  }
}
```

This would be a great improvement to create easy to parse tutorials. Right now we see things such as:

```
  "Gateway": {
    "APICommands": [],
    "HTTPHeaders": {
      "Access-Control-Allow-Headers": [
        "X-Requested-With",
        "Range"
      ],
      "Access-Control-Allow-Methods": [
        "GET"
      ],
      "Access-Control-Allow-Origin": [
        "*"
      ]
    },
    "PathPrefixes": [],
    "RootRedirect": "",
    "Writable": false
  },
```

Which raises questions such as "Why does the Gateway need CORS headers, it only receives GET requests?", "What is a Writable Gateway?" and more.

lidel:
> is it the right time to ask that the IPFS config.json gets organized by system rather than a mix of everything.

@daviddias Are you able to be more specific? What exactly feels to you like mixing concerns?

My understanding is that we already have separate namespaces for `API` and  `Gateway` ports, and we are able to define different `HTTPHeaders` for each: 
```
$ ipfs config --json Gateway.HTTPHeaders
$ ipfs config --json API.HTTPHeaders 
``` 

`APICommands` is present in `Gateway` section because it is an optional override of a "safe" API subset exposed on Gateway port (see below).

> Which raises questions such as "Why does the Gateway need CORS headers, it only receives GET requests?"

CORS on Gateway  is useful for API subset exposed there. Websites are able to read data via cross-origin XHR to https://ipfs.io/api/v0/dns/ipfs.io?r=true and AFAIK that is why our public Gateway at ipfs.io returns `Access-Control-Allow-Origin: *` currently.

Stebalien:
@daviddias, @lidel is correct but I do agree that the config could use a bit of a reorg with 20-20 hindsight (same as the commands). However, that would be a massively breaking change. Let's reconsider this when we reorganize the API/commands. When we do that, we can insert a config translation layer such that `ipfs-old config` sees the old config and `ipfs-new config` sees the new one.

Issue #5887, opened 2019-01-02
Escape paths in `ipfs ls`


magik6k:
Running `ipfs object patch add-link QmUNLLsPACCz1vLxQVkXqqLX5R1X345qqfHbsf67hvA3Nn $(echo -e '\e[1;1H\e[44;37m\e[K\e[2;1HHI\e[34m_\e[37mTHERE!\e[3;1H\e[K\e[49;39m') QmUNLLsPACCz1vLxQVkXqqLX5R1X345qqfHbsf67hvA3Nn` gets us a really fun hash - `QmUKpty1aF3ak42WTnebPsGkYyq6A6g9Jt2fYpanwzk6tx`.

`ipfs ls` on that hash will print `HI THERE` on a blue background.

We should detect \e (and other characters like space) and escape it in cli encoder

PR #5895, opened 2019-01-06
feat(cmd/ls): link name removes invisible char

chenminjian:
refs: #5887 

License: MIT

### DIFF
--- a/core/commands/ls.go
+++ b/core/commands/ls.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"regexp"
 	"text/tabwriter"
 
 	cmdenv "github.com/ipfs/go-ipfs/core/commands/cmdenv"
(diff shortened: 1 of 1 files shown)


schomatis:
> we might use `strconv.QuoteToASCII` for strings that need escaping

Agreed.

Do we allow this kind of names in the first place? Is there any character whitelist for UnixFS names? (From what I've seen in the code I think not; is the new UnixFSv2 spec considering it?)

magik6k:
cc @mikeal @warpfork

Stebalien:
Take a look at https://github.com/ipld/specs/issues/84 and https://github.com/ipfs/unixfs-v2/issues/3. As far as I know, there isn't a complete proposal. However, UnixFSv2 will have restrictions.

I really don't know the best way to handle existing files (escaping sounds promising but we have to be careful). `ls` uses a weird bash-like escaping.

chenminjian:
Ok, I will look at those discussions first.

schomatis:
> Ok, I will look at those discussions first.

Those references were mostly for my question, you needn't care about those design decisions that escape the scope of this PR, you can definitely check them out if you're interested but I think we (i.e., `go-ipfs` team) should settle for a temporary solution while UnixFSv2 is in the oven.

Stebalien:
(closing as we'll have to rethink this a bit)

Issue #5903, opened 2019-01-08
Slightly confusing/crowded error message when running daemon without init


whyrusleeping:
On a brand new machine I ran `ipfs daemon` and got:

```
why@why-MS-7B46 ~/g/s/g/i/go-ipfs> ipfs daemon
Initializing daemon...
go-ipfs version: 0.4.19-dev-a142aecf9
Repo version: 7
System version: amd64/linux
Golang version: go1.11.2
Successfully raised file descriptor limit to 2048.
Error: no IPFS repo found in /home/why/.ipfs.
please run: 'ipfs init'
Received interrupt signal, shutting down...
(Hit ctrl-c again to force-shutdown the daemon.)
```

Thats a lot of extra information, and the bit telling me what I need to know is a bit hidden. I would expect showing much less to be less confusing, liike:
```
Initializing daemon...
Error: no IPFS repo found in /home/why/.ipfs.
please run: 'ipfs init'
```

PR #5904, opened 2019-01-08
reduce verbosity of daemon start

Stebalien:
Also, avoid printing C-c message unless we *actually* start the daemon:

Now `ipfs daemon` with no repo prints:

```
Initializing daemon...
go-ipfs version: 0.4.19-dev-7a9b1716f
Repo version: 7
System version: amd64/linux
Golang version: go1.11.4
Error: no IPFS repo found in /tmp/xyz.
please run: 'ipfs init'
```

work towards: #5903


Stebalien:
I've kept the version information because it's pretty useful but we could try adding some space.

Stebalien:
I've pushed a commit that adds a new line when exiting the daemon command with an error. That way, it doesn't matter *where* we run into an error, it'll always be clear.

We could do this for all commands but always prepending a newline could look a bit funky.

PR #5906, opened 2019-01-08
ls: Report real file size

magik6k:
Need this for https://github.com/ipfs/go-ipfs-http-client/pull/1 to implement fast `api.Unixfs().Get()`.

### DIFF
--- a/test/sharness/t0045-ls.sh
+++ b/test/sharness/t0045-ls.sh
@@ -45,8 +45,8 @@ EOF
   test_expect_success "'ipfs ls <three dir hashes>' output looks good" '
     cat <<-\EOF >expected_ls &&
 QmfNy183bXiRVyrhyWtq3TwHn79yHEkiAGFr18P7YNzESj:
-QmSix55yz8CzWXf5ZVM9vgEvijnEeeXiTSarVtsqiiCJss   d1/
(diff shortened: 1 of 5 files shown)


test/sharness/t0045-ls.sh
```diff
@@ -69,20 +92,20 @@ EOF
     cat <<-\EOF >expected_ls_headers &&
 QmfNy183bXiRVyrhyWtq3TwHn79yHEkiAGFr18P7YNzESj:
 Hash                                           Size Name
-QmSix55yz8CzWXf5ZVM9vgEvijnEeeXiTSarVtsqiiCJss 246  d1/
-QmR3jhV4XpxxPjPT3Y8vNnWvWNvakdcT3H6vqpRBsX1MLy 1143 d2/
-QmeomffUNfmQy76CQGy9NdmqEnnHU9soCexBnGU3ezPHVH 13   f1
-QmNtocSs7MoDkJMc1RkyisCSKvLadujPsfJfSdJ3e1eA1M 13   f2
+QmSix55yz8CzWXf5ZVM9vgEvijnEeeXiTSarVtsqiiCJss      d1/
+QmR3jhV4XpxxPjPT3Y8vNnWvWNvakdcT3H6vqpRBsX1MLy      d2/
```
Stebalien: Hm. This could be a bit of an issue. I assume there are quite a few people parsing output using `cut`. We should probably print *something* (even if it's just `-`).

PR #5922, opened 2019-01-12
coreapi: replace coreiface.DagAPI with ipld.DAGService

magik6k:
As `api.Dag()` is the interface to IPLD, this feels like the right thing to do.

This wasn't done this way before because we keep format parsers in a global map in `core/coredag`, exposing ipld interfaces directly would make it tricky for outside code to do things like parse json to dag-cbor, but from my experience most Go code which touches IPLD doesn't really need those parsers.

(Since this is mostly a yes/no decision, I'd like this to move relatively fast as it's blocking a chunk of work in go-ipfs-http-api)


magik6k:
> make sure that ipfs-interface-core matches

Conceptually using go-ipld-format.DAGService is going to be closer to js/ipfs-interface-core, only the names are a bit different - `put`/`Add`, `get`/`Get`, `tree`/`Resolver.Tree`

Stebalien:
SGTM.

Issue #4279, opened 2017-10-05
Make badger-ds default


Stebalien:
It's much faster.

* [ ] Make sure the on-disk format is stable.
* [ ] Test it.
* [ ] Make it the default.

schomatis:
@Stebalien Do you mind if I hijack this issue to keep track of all the other issues related to the Badger transition?

Stebalien:
@schomatis go right ahead!

ajbouh:
Is this still on track to happen sometime soon?

Stebalien:
In addition to the issues listed in the description, we're still working through some [recovery issues](https://github.com/ipfs/go-ipfs/issues/5213)  and memory usage is pretty bad (we may be able to [tune](https://github.com/ipfs/go-ds-badger/issues/17) this a bit but I'm getting some [really weird behavior on Linux](https://github.com/ipfs/go-ds-badger/issues/18)).

Basically, we can't roll this out until:

1. We can always recover after a crash.
2. It doesn't eat ram needlessly.

schomatis:
Thanks for the reference, I should add those.

djdv:
I doubt this effects most users but I'm linking it anyway.
My own instance of IPFS runs with flatfs hosted on an SMB/CIFS share.
badger doesn't currently handle this: https://github.com/dgraph-io/badger/issues/699
although it can.

For full context, I do this because my local disks are small. And I can't run IPFS on the remote machine because components of libp2p don't build on Solaris yet.
(when trying to port it I encountered an oddity where the Go standard library says something is implemented but it isn't)

Stebalien:
@magik6k IMO, we _should_ be able to graduate badger from experimental even if we don't go ahead and make it the default. However, we may want to land https://github.com/ipfs/go-ds-badger/issues/51 first.

ZerxXxes:
Hey, whats the status here? I see that https://github.com/ipfs/go-ds-badger/issues/51 is closed, does that mean that the badger datastore is to be considered pretty mature now?

Kubuxu:
We should probably update to badger v2 before using it as default.

RubenKelevra:
I just want to note that flatfs has some advantages: When using ZFS as the underlying filesystem, (with a blocksize set to 256 K for ZFS and raw-leaves for IPFS) you can dedup the blockstorage against a copy of the data outside the block storage you might need to hold for a different service, like http. This isn't possible with data stored inside a database.

[...]

PR #5923, opened 2019-01-13
Clarify that chunker sizes are in bytes

da2x:
Clarify that the Rabin fingerprint chunker is in bytes, and recommend using larger chunk sizes than shown in the provided examples. People are getting confused over chunker sizes and incorrectly assuming they're in kiB and not bytes because of the sizes used in the examples.


core/commands/add.go
```diff
@@ -78,12 +78,16 @@ You can now refer to the added file in a gateway, like so:
...
 256 * 1024 bytes, 'size-262144'. Alternatively, you can use the
-rabin chunker for content defined chunking by specifying
-rabin-[min]-[avg]-[max] (where min/avg/max refer to the resulting
-chunk sizes). Using other chunking strategies will produce
-different hashes for the same file.
+Rabin fingerprint chunker for content defined chunking by specifying
+rabin-[min]-[avg]-[max] (where min/avg/max refer to the desired
+chunk sizes in bytes), e.g. 'rabin-262144-524288-1048576'.
+
+The following examples use very small byte sizes to demonstrate the
```
Stebalien: 1k isn't *that* small and users definitely shouldn't use `2048*1024` (2MiB) chunks. Let's just explain the example (e.g., "For example, to chunk a file into 1KiB chunks...").

core/commands/add.go
```diff
@@ -78,12 +78,16 @@ You can now refer to the added file in a gateway, like so:
 
 The chunker option, '-s', specifies the chunking strategy that dictates
 how to break files into blocks. Blocks with same content can
-be deduplicated. The default is a fixed block size of
+be deduplicated. Different chunking strategies will produce different
```
Stebalien: Nit: We might want to move the pros/cons of chunking to a new paragraph (this is pretty choppy). However, your change is still an improvement so we can mess with that later if you'd like.

core/commands/add.go
da2x: I’m never suggesting using 2 MiB chunks. rabin-262144-524288-1048576 is in bytes and not kilobytes, so it’s 128 kiB–256 kiB–512 kiB. This was what this patch was intended to help clarify.

1 kiB is ridiculously small in this context. You don’t ever want to go through DHT and peer discovery for a file that is split into a thoiusand 1 kiB chunks. The protocol overhead would be enormous. Keep in mind that you’re also storing each individual chunk in a file that takes up a 256 kiB block on most file systems, so the disk packing wastage would also be enormous.

core/commands/add.go
da2x: This sentence is one of the most important things to know about this option and chunking. I moved it up so people would read it early before they get distracted by all the details.

Stebalien: My complaint is that it's "fact, fact, fact..." with no relationship between the facts. But didn't flow all that well before so we can fix this later.

core/commands/add.go
Stebalien: > You don’t ever want to go through DHT and peer discovery for a file that is split into a thoiusand 1 kiB chunks.

You won't have to do that. You'll go through the DHT for the root *node* (which will likely be smaller than 1KiB as it doesn't contain any actual data).

>  Keep in mind that you’re also storing each individual chunk in a file that takes up a 256 kiB block on most file systems, so the disk packing wastage would also be enormous.

Most filesystem have 4KiB blocks. Also note that we're moving towards a datastore that doesn't store each chunk in a separate file. However, I do agree that a 1KiB is small.

> I’m never suggesting using 2 MiB chunks. rabin-262144-524288-1048576 is in bytes and not kilobytes, so it’s 128 kiB–256 kiB–512 kiB. This was what this patch was intended to help clarify.

Ah, yeah, you're right. We might want to just bump those sizes up instead.

Stebalien: Otherwise, I'd be fine merging this as-is.

hinshun: @Stebalien Where can I find more information about `Also note that we're moving towards a datastore that doesn't store each chunk in a separate file.`?

Stebalien: https://github.com/ipfs/go-ipfs/issues/4279

PR #5934, opened 2019-01-21
coreapi: add some seeker tests

magik6k:
This tests dagreader in `go-unixfs/io`, and it turns out it's currently broken in 2 ways:

* <s>When seeking near end of the file and the reading into buffer larger than the file, `Read` seems to read the data, but incorrectly returns io.EOF</s> (this is valid behaviour, just not consistent across readers)

* When seeking with `io.SeekEnd`, DagReader incorrectly inverts the offset.


magik6k:
Test incorrectly handled `io.EOF`, should now be fixed

PR #5925, opened 2019-01-14
[WIP] Unixfs: enforce refs on files when using nocopy

magik6k:
Fixes https://github.com/ipfs/go-ipfs/issues/4558

Depends on:
* https://github.com/ipfs/go-unixfs/pull/56
* https://github.com/ipfs/go-mfs/pull/55

PR #5936, opened 2019-01-22
mfs: make sure to flush after mv and chcid

Stebalien:
We need to call `FlushPath` as `Flush` doesn't call `WaitPub` internally. The correct fix is to rework MFS such that Flush actually flushes but that's a larger project.

Fixes https://github.com/ipfs/go-ipfs/pull/5925#issuecomment-456252134 although I'm not sure what changed. In a followup PR, once https://github.com/ipfs/go-mfs/pull/58 is merged, we can fix go-ipfs to run the teardown function *before* it cancels the context.

### DIFF
--- a/core/commands/files.go
+++ b/core/commands/files.go
@@ -636,6 +636,8 @@ Example:
 			return err
 		}
 
+		flush, _ := req.Options[filesFlushOptionName].(bool)
+
 		src, err := checkPath(req.Arguments[0])
 		if err != nil {
 			return err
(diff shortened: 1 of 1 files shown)


core/commands/files.go
```diff
@@ -645,7 +647,14 @@ Example:
 			return err
 		}
 
-		return mfs.Mv(nd.FilesRoot, src, dst)
+		err = mfs.Mv(nd.FilesRoot, src, dst)
+		if err != nil {
+			return err
+		}
+		if flush {
+			return mfs.FlushPath(nd.FilesRoot, "/")
```
Stebalien: This does, in fact, work. We don't need to care about the individual files.

PR #5942, opened 2019-01-24
 Fix the PUT-existing bug in writable gateway API

Stebalien:
MOVED FROM: https://github.com/ipfs/go-ipfs/pull/3931

@jes wrote:

> See on https://discuss.ipfs.io/t/writeable-http-gateways/210
> 
> Previously, a PUT to a filename that already exists would return the hash
> of the newly-inserted file instead of the hash of the new tree. This is not
> useful, and it makes much more sense for the API to be consistent.
> 
> Furthermore, anybody who knows about the bug is already DELETE-ing the filename
> before PUT-ing the new content, so the fix doesn't trample over that behaviour
> either.
> 
> With this fix, every conceivable usage of the API is either unaffected or
> improved.
> 
> It could do with a unit test.
> 
> ----
> 
> All I did was remove the special-case for the "file already exists" case, and it works correctly. It's easier to see this if you view ```git diff -w```.

### DIFF
--- a/core/corehttp/gateway_handler.go
+++ b/core/corehttp/gateway_handler.go
@@ -421,73 +421,38 @@ func (i *gatewayHandler) putHandler(w http.ResponseWriter, r *http.Request) {
 		newPath = path.Join(rsegs[2:])
 	}
 
-	var newcid cid.Cid
-	rnode, err := core.Resolve(ctx, i.node.Namesys, i.node.Resolver, rootPath)
-	switch ev := err.(type) {
-	case resolver.ErrNoLink:
-		// ev.Node < node where resolve failed
-		// ev.Name < new link
-		// but we need to patch from the root
-		c, err := cid.Decode(rsegs[1])
-		if err != nil {
-			webError(w, "putHandler: bad input path", err, http.StatusBadRequest)
-			return
-		}
-
-		rnode, err := i.node.DAG.Get(ctx, c)
-		if err != nil {
-			webError(w, "putHandler: Could not create DAG from request", err, http.StatusInternalServerError)
-			return
-		}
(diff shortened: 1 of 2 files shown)


Stebalien:
This was previously reviewed by both @whyrusleeping and @Kubuxu. I believe I've addressed their remaining comments (https://github.com/ipfs/go-ipfs/pull/3931#pullrequestreview-46346057).

Stebalien:
This looks like something we'll need to fix:

> Tests / linuxSharness / 40 - We can HTTP GET file just replaced – sharnessLinux.t0111-gateway-writeable

```
Stacktrace
URL="http://localhost:$port/ipfs/$HASH/test/test.txt" && echo "GET $URL" && curl -svo outfile3 "$URL" 2>curl_getExisting.out && test_cmp infile3 outfile3
Standard Output
expecting success:
URL="http://localhost:$port/ipfs/$HASH/test/test.txt" &&
echo "GET $URL" &&
curl -svo outfile3 "$URL" 2>curl_getExisting.out &&
test_cmp infile3 outfile3
GET http://localhost:37288/ipfs/QmfGKo7mm41im56UKjRJFKYqJiTMhsmYcNazSYHSzGmxnF/test/test.txt
> diff -u infile3 outfile3
--- infile3 2019-01-24 15:34:02.691086012 +0000
+++ outfile3 2019-01-24 15:34:05.886928567 +0000
@@ -1 +1 @@
-8372
+24246
not ok 40 - We can HTTP GET file just replaced
Standard Error
```

Stebalien:
Fixed in https://github.com/libp2p/go-libp2p-connmgr/pull/50.

Issue #5690, opened 2018-10-29
Archive size wrong


Stebalien:
When running `ipfs get --archive`, the expected size in the progress bar is wrong:

```bash
> ipfs get --archive /ipfs/QmPwecoMB9jN9DosznNqcApK63PnrEmMduxp1mn5nqS7Pq
Saving archive to QmPwecoMB9jN9DosznNqcApK63PnrEmMduxp1mn5nqS7Pq.tar
 68.00 KiB / 49.63 KiB [=====================================] 137.02% 0s
```

That is, the output file is *actually* 68KiB, 49.63 is the total size one would get when running `ipfs get` without `--archive`.

Stebalien:
Actually, it looks like the size computation is just wrong every which way, we're using the size of the DAG. Worse, we're using it to set the content length.

overbool:
@Stebalien `ipfs get Qm...` without `--archive` prints the protobuf node size. we should unmarshal the data and then computate its size.

```go 
	switch dn := dn.(type) {
		case *dag.ProtoNode:
			//size, err := dn.Size()
			//if err != nil {
			//	return err
			//}
			fsn, err := unixfs.FSNodeFromBytes(dn.Data())
			if err != nil {
				return err
			}
			res.SetLength(fsn.FileSize())
```

Stebalien:
Really, I have no idea how to fix it. We need to compute the size on the server before we start transferring which we often can't do because we don't actually *know* it ahead of time. We have the size of the DAG but not the size of the TAR we're actually transferring (also, the DAG's size may be wrong).

I'm not sure if we can fix this properly with the current protocol.

chenminjian:
Maybe we can store the TAR in ipfs. Then we know the size of it. Does that make sense?

Stebalien:
That's not going to work for large datasets, unfortunately. We can't afford to materialize the entire tar file before shipping it off to the user.

PR #5943, opened 2019-01-24
Refactor ipfs get to use CoreAPI

magik6k:
Depends on:
* [x] https://github.com/ipfs/go-unixfs/pull/59
* [x] https://github.com/ipfs/go-ipfs-files/pull/4


Stebalien:
> why put tar-writer in go-ipfs-files

Because `coreapi.Unixfs().Get()` now returns a `go-ipfs-files.Node` object. Therefore, the `TarWriter` needs to convert this (instead of a dag) into a tarfile.

However: https://github.com/ipfs/go-unixfs/pull/59#issuecomment-458731609

> Am I correct in understanding that neither of these two changes are strickly essential to delivering the refactor to make get use CoreAPI?

We need *some* way to go from a `go-ipfs-files.Node` to a tar because `coreapi.Unixfs().Get()` returns a `go-ipfs-files.Node` instead of a raw IPLD DAG. We could just keep `DagArchive` as it is as a way to go directly from a DAG to a tar file.

magik6k:
> We could just keep DagArchive as it is as a way to go directly from a DAG to a tar file.

As a note - I decided to drop it as the combo of `NewUnixfsFile`+`files.TarWriter` do the same thing with the only difference being the `files` interface in between (there shouldn't be any meaningful performance impact). Plus after switching `ipfs get` to the new thing, it would basically become dead code.

core/commands/get.go
```diff
@@ -66,33 +66,30 @@ may also specify the level of compression by specifying '-l=<1-9>'.
...
-			res.SetLength(size)
-		case *dag.RawNode:
-			res.SetLength(uint64(len(dn.RawData())))
-		default:
+		size, err := file.Size()
+		if err != nil {
 			return err
 		}
 
+		res.SetLength(uint64(size))
```
Stebalien: This is actually incorrect but it's an existing bug... https://github.com/ipfs/go-ipfs/issues/5690

magik6k: The alternative is not setting this at all, but it will break progress bar... (We could pass this in a separate header)

Stebalien:
Error is: "Error: cannot write to readonly DAGService"

Stebalien:
The issue is: https://github.com/ipfs/go-unixfs/blob/0eb1ef8f33371b1b9f74f74655b584144919ce91/file/unixfile.go#L121. This is causing the hamt to serialize and save itself.

magik6k:
Ah, right, it breaks the progress bar ._.

magik6k:
See https://github.com/ipfs/go-unixfs/pull/61

PR #5955, opened 2019-01-29
plugin: add a daemon plugin with access to the CoreAPI

### LINKED ISSUE #5954: "Generic" Plugin type(feature)?
#### Version information:

go-ipfs version: 0.4.18-aefc746
Repo version: 7
System version: amd64/linux
Golang version: go1.11

#### Type:

Feature Request

#### Description:

I would like to be able to create a type of plugin which extends IPFS's functionality by connecting it to the API's of other applications automatically, in particular my use case is making the gateway available via tunnels and hidden services. The current plugin system requires that plugins implement one of the existing plugin interfaces, and implementing the hidden service listeners on top of the existing plugin interfaces is a pretty messy process. In our branch at https://github.com/rtradeltd/go-ipfs I have created a prototype of a "Generic" plugin interface which only requires implementing "plugin.Plugin" with an Exists() function that is intended to the return nil if the plugin was initialized correctly, and an error if something went wrong. If this is acceptable I'd like to submit it.

Stebalien:
fixes #5954

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -30,6 +30,21 @@ related commands.
 
 Datastore plugins add support for additional datastore backends.
 
+### Tracer
+
+(experimental)
+
+Tracer plugins allow injecting an opentracing backend into go-ipfs.
+
+### Daemon
+
+Daemon plugins are started when the go-ipfs daemon is started and are given an
+instance of the CoreAPI. This should make it possible to build an ipfs-based
+application without IPC and without forking go-ipfs.
+
+Note: We eventually plan to make go-ipfs usable as a library. However, this
+plugin type is likely the best interim solution.
+
 ## Available Plugins
 
 | Name                                                                            | Type      | Preloaded | Description                                    |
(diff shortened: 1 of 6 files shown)


plugin/loader/loader.go
```diff
@@ -81,41 +84,83 @@ func (loader *PluginLoader) Initialize() error {
...
 
-//Run the plugins
-func (loader *PluginLoader) Run() error {
+// Inject hooks all the plugins into the appropriate subsystems.
+func (loader *PluginLoader) Inject() error {
 	for _, pl := range loader.plugins {
-		switch pl := pl.(type) {
-		case plugin.PluginIPLD:
-			err := runIPLDPlugin(pl)
+		if pl, ok := pl.(plugin.PluginIPLD); ok {
```
magik6k: I assume this is because we want plugins to be able to implement more that one interface, right?

Stebalien: Yeah, there's no reason for these interfaces to be exclusive.

docs/plugins.md
```diff
@@ -30,6 +30,21 @@ related commands.
 
 Datastore plugins add support for additional datastore backends.
 
+### Tracer
+
+(experimental)
+
+Tracer plugins allow injecting an opentracing backend into go-ipfs.
```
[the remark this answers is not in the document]
Stebalien: Linked.

plugin/loader/loader.go
```diff
@@ -81,41 +84,83 @@ func (loader *PluginLoader) Initialize() error {
...
+	return nil
+}
+
+// Start starts all long-running plugins.
+func (loader *PluginLoader) Start(iface coreiface.CoreAPI) error {
+	for i, pl := range loader.plugins {
+		if pl, ok := pl.(plugin.PluginDaemon); ok {
+			err := pl.Start(iface)
 			if err != nil {
+				closePlugins(loader.plugins[i:])
```
fd: I believe this is a bug. It should read `loader.plugins[:i]` as all plugins _before_ `i` have been started not those _after_ `i`.

magik6k: This seems to be the case, do you mind sending a PR?

fd: I'll send a PR.

PR #5960, opened 2019-02-01
docs: duplicate default helptext in `name publish`

djdv:
> -k,            --key      string - Name of the key to be used or a valid PeerID, as listed by 'ipfs key list -l'. Default: Default: self..


Kubuxu:
Apart from tests randomly failing. A command in line 77 also uses `<<default>>`, it looks like <<default>> is not supported so it should be resolved too.

djdv:
This is kind of strange.
The text for `-t` (line 77) seems to be working as intended and not duplicating "Default:", but does use the same pattern that causes this problem for `-k`. I'm not sure what's going on here.

On my machine with this patch it looks like this
image

Issue #5929, opened 2019-01-18
Make sure Gateway.NoFetch applies fully


magik6k:
https://github.com/ipfs/go-ipfs/pull/5649 implemented `Gateway.NoFetch` option and applies it to `/api` by using global offline coreapi option. This means that read-only commands which don't yet use coreapi aren't covered and need to be converted:

From https://github.com/ipfs/go-ipfs/pull/5649#issuecomment-451337849:
* [x] `/get` - <s>will be hard to port, there is a lot of code surrounding it which will need to be tamed.</s> - #5943
* [ ] `/dns` - needs to get into coreapi (probably under `.Name()`) (doesn't really need to be covered)
* [x] `/ls` - Can be rewritten to use coreapi, but that's going to be quite hard (it can be worked around manually for now)
* [x] `/dag/get` - needs to be ported to coreapi
* [x] `/dag/resolve` - same as above

PR #5962, opened 2019-02-01
Port `ipfs ls` to CoreAPI

magik6k:
(Last) Part of https://github.com/ipfs/go-ipfs/issues/5929

### DIFF
--- a/core/coreapi/interface/options/unixfs.go
+++ b/core/coreapi/interface/options/unixfs.go
@@ -43,8 +43,6 @@ type UnixfsAddSettings struct {
 }
 
 type UnixfsLsSettings struct {
-	Async bool
-
 	ResolveType bool
 	ResolveSize bool
 }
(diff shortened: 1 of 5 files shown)


core/coreapi/interface/options/unixfs.go
```diff
@@ -42,7 +42,15 @@ type UnixfsAddSettings struct {
 	Progress bool
 }
 
+type UnixfsLsSettings struct {
+	Async bool
```
Stebalien: So, what if we just *always* streamed at the CoreAPI layer and collected/sorted the results at the commands layer?

magik6k: Good point, 40 lines less to care about

core/coreapi/interface/options/unixfs.go
```diff
@@ -42,7 +42,13 @@ type UnixfsAddSettings struct {
 	Progress bool
 }
 
+type UnixfsLsSettings struct {
+	ResolveType bool
+	ResolveSize bool
```
Kubuxu: As a side effect of both those operations (resolution of child nodes), they don't need to be separate options. If any of them is true the work is done, so there is no reason to have them both.

IMO have one `ResolveChildren` option.

core/coreapi/interface/unixfs.go
```diff
@@ -16,6 +16,14 @@ type AddEvent struct {
 	Size  string       `json:",omitempty"`
 }
 
+type LsLink struct {
+	Link *ipld.Link
+	Size uint64
+	Type unixfs_pb.Data_DataType
```
Kubuxu: Can we somehow make it not be unixfs-pb specific?

Issue #5963, opened 2019-02-02
files mv cmd has wrong dst filename


chenminjian:
#### Version information:
go-ipfs version: 0.4.19-dev-
Repo version: 7
System version: amd64/darwin
Golang version: go1.11.1

#### Type:
bug

#### Description:
```
$ ipfs files cp /ipfs/QmbAq3aBS8SKPWqiDPcXBSLpxEhp8h4SfiUjtvdjiP6Q5f /a

$ ipfs files mkdir /dir66

$ ipfs files mv /a /dir66

$ ipfs files ls /dir66
dir66
```

`a` is the correct answer instead of `dir66 `.

Stebalien:
(fixed)

PR #5964, opened 2019-02-02
fix(cmd/mv): dst filename error

chenminjian:
#5963 

License: MIT

### DIFF
--- a/test/sharness/t0250-files-api.sh
+++ b/test/sharness/t0250-files-api.sh
@@ -524,6 +524,14 @@ test_files_api() {
     ipfs files mv /cats/this/is /cats/
   '
 
+  test_expect_success "can mv dir and dest dir is / $EXTRA" '
+    ipfs files mv /cats/is /
+  '
+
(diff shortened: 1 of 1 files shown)


magik6k:
(go-mfs PR - https://github.com/ipfs/go-mfs/pull/62)

Stebalien:
Dropped the gx update as mfs has already been updated. Kept the test.

test/sharness/t0250-files-api.sh
```diff
@@ -521,7 +521,11 @@ test_files_api() {
 
   # test mv
   test_expect_success "can mv dir $EXTRA" '
-    ipfs files mv /cats/this/is /cats/
+    ipfs files mv /cats/this/is /
```
Stebalien: We should test moving to _both_ `/cats/` and `/cats`. Maybe move to `/cats/`, then to `/`, then to `/cats`?

chenminjian: OK, I have done it.

PR #5969, opened 2019-02-04
block put --pin

magik6k:
Needed in ipfs/go-ipfs-http-api#1

### DIFF
--- a/core/commands/block.go
+++ b/core/commands/block.go
@@ -146,6 +146,7 @@ than 'sha2-256' or format to anything other than 'v0' will result in CIDv1.
 		cmdkit.StringOption(blockFormatOptionName, "f", "cid format for blocks to be created with."),
 		cmdkit.StringOption(mhtypeOptionName, "multihash hash function").WithDefault("sha2-256"),
 		cmdkit.IntOption(mhlenOptionName, "multihash hash length").WithDefault(-1),
+		cmdkit.BoolOption(pinOptionName, "pin added blocks recursively").WithDefault(false),
 	},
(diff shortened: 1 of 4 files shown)


alanshaw:
@magik6k could I get some background info on why this change?

Is this default false for backwards compatibility? Should we consider adding a pin option and default it to true for _all_ operations that add data to IPFS? Including for example `dag.put`?

magik6k:
go-ipfs-http-client implements DAGService using `block` calls (js-ipfs-http-client does the same thing), and also needs to implement a pinning node adder:
https://github.com/ipfs/go-ipfs/blob/803e9a7e6608468616e0379cafd94f0abe96da51/core/coreapi/interface/dag.go#L7-L13

Calls to this adder automatically pin related nodes, and because go-ipfs-http-client adds nodes through block calls, this is also where pinning needs to happen (because we need to hold a pinning lock)

> Is this default false for backwards compatibility?

Defaulting to false is mostly for backwards compatibility (users may not be aware that adding objects this way pins them, and currently our pinning system isn't happy about many small pins)

> Should we consider adding a pin option [...] for all operations that add data to IPFS?

That would be nice for consistency, but I feel like there should be a better way

> default it to true for all operations

Current pinning system really struggles with many individual pins, I'd say not before we get a more scalable pinning system.

> for example `dag.put`

In go-ipfs `dag.put` already has a pin option - https://github.com/ipfs/go-ipfs/blob/master/core/commands/dag/dag.go#L65, which defaults to false too.

Stebalien:
Waiting on a final signoff by @alanshaw.

PR #5971, opened 2019-02-05
coreapi: fix seek test on http impl

magik6k:


core/coreapi/interface/tests/unixfs.go
```diff
@@ -967,7 +975,7 @@ func (tp *provider) TestGetSeek(t *testing.T) {
 	}
 
 	orig := make([]byte, dataSize)
-	if _, err := f.Read(orig); err != nil {
+	if _, err := io.ReadFull(f, orig); err != nil {
```
magik6k: Need this because http reader reads data in 4k chunks

PR #5975, opened 2019-02-08
gx: update go-libp2p-peer

Stebalien:
Switch _back_ to the 0.4.18 style of peer IDs while we figure things out. See https://github.com/libp2p/specs/issues/138.


Stebalien:
This *also* pulls the new provider query subsystem for bitswap.

Stebalien:
And the DHT query queue fixes, and probably a bunch of others.

Kubuxu:
Wait, how is Circle passing and everything else failing?

Kubuxu:
Circle has separate status for gotest.

PR #5976, opened 2019-02-08
coreapi: cleanup non-gx references

magik6k:
This is the last thing on the 'todo before extracting interface' list.


Kubuxu:
gotest is failing.

PR #5978, opened 2019-02-09
coreapi: extract interface

magik6k:
Now lives in https://github.com/ipfs/interface-go-ipfs-core


Kubuxu:
That will be annoyingly long import name, we didn't think about it while deciding on naming scheme.

Stebalien:
My only comment is that the repo name doesn't match the go naming convention (would otherwise be go-ipfs-interface-core). But don't really mind either way.

magik6k:
The naming is to match https://github.com/ipfs/interface-js-ipfs-core/issues/423#issuecomment-457305646

Stebalien:
I'm not going to bikeshed this too much but I've pinged david to try to get his rational. I *do* like being able to run shell commands over `go-*`.

Stebalien:
We can quibble over naming later.

PR #5979, opened 2019-02-09
Improve "name publish" ttl option documentation

AuHau:
Also refactoring DefaultRecordTTL to DefaultRecordEOL as it represents default value for `lifetime` and not `ttl`.

### DIFF
--- a/core/commands/name/publish.go
+++ b/core/commands/name/publish.go
@@ -78,7 +78,7 @@ Alternatively, publish an <ipfs-path> using a valid PeerID (as listed by
     This accepts durations such as "300s", "1.5h" or "2h45m". Valid time units are
     "ns", "us" (or "µs"), "ms", "s", "m", "h".`).WithDefault("24h"),
 		cmdkit.BoolOption(allowOfflineOptionName, "When offline, save the IPNS record to the the local datastore without broadcasting to the network instead of simply failing."),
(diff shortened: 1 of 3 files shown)


core/commands/name/publish.go
```diff
@@ -78,7 +78,7 @@ Alternatively, publish an <ipfs-path> using a valid PeerID (as listed by
     This accepts durations such as "300s", "1.5h" or "2h45m". Valid time units are
     "ns", "us" (or "µs"), "ms", "s", "m", "h".`).WithDefault("24h"),
 		cmdkit.BoolOption(allowOfflineOptionName, "When offline, save the IPNS record to the the local datastore without broadcasting to the network instead of simply failing."),
-		cmdkit.StringOption(ttlOptionName, "Time duration this record should be cached for (caution: experimental)."),
+		cmdkit.StringOption(ttlOptionName, "Time duration this record should be cached for. Uses same syntax like lifetime option. (caution: experimental)"),
```
Stebalien:
```suggestion
		cmdkit.StringOption(ttlOptionName, "Time duration this record should be cached for. Uses the same syntax as the lifetime option. (caution: experimental)"),
```

Or just `(e.g. "10m")`? IMO, that should get the point across in fewer words.

AuHau: Thanks for the wording check. I am not sure if the `(e.g. "10m")` is not too concise. What about `See 'lifetime' option.`, but I guess that is not much better.

PR #5985, opened 2019-02-11
gx: update go-ipfs-files

Stebalien:
fix compatibility issue with js-ipfs


Stebalien:
The interop tests now pass but there's something wrong with sharding (calculating the wrong hash...).

Stebalien:
Fix in: https://github.com/ipfs/go-ipfs-files/pull/7

Stebalien:
(merging as this is a gx-only patch)

PR #6004, opened 2019-02-19
fix a few interop test issues

Stebalien:
The _interop_ tests passed but the interface tests didn't.

* Return "not a directory" again when trying to cat a directory.
* Use EmitOnce in `config`.
* Return the correct error in `bootstrap rm`.

Other issues:

* We now disagree on size: https://github.com/ipfs/interface-js-ipfs-core/issues/427
* We now normalize paths on add: https://github.com/ipfs/interface-js-ipfs-core/pull/440


core/commands/cat.go
```diff
@@ -128,8 +128,14 @@ func cat(ctx context.Context, api iface.CoreAPI, paths []string, offset int64, m
...
-		file, ok := f.(files.File)
-		if !ok {
+		var file files.File
+		switch f := f.(type) {
+		case files.File:
+			file = f
+		case files.Directory:
+			return nil, 0, iface.ErrIsDir
+		case *files.Symlink:
+			// TODO: Consider returning "error not supported"?
```
Stebalien: We can handle this later, IMO.

core/commands/cat.go
```diff
@@ -128,8 +128,14 @@ func cat(ctx context.Context, api iface.CoreAPI, paths []string, offset int64, m
...
-		file, ok := f.(files.File)
-		if !ok {
+		var file files.File
+		switch f := f.(type) {
+		case files.File:
+			file = f
+		case files.Directory:
+			return nil, 0, iface.ErrIsDir
+		default:
+			// TODO: Consider returning "error not supported" for symlinks.
```
magik6k: Doesn't go-unixfs do that? https://github.com/ipfs/go-unixfs/blob/master/io/dagreader.go#L80

Stebalien: Yeah, I just didn't feel like adding that error to the core API yet (I'm being lazy and trying to cut a release...).

Stebalien: @magik6k https://github.com/ipfs/interface-go-ipfs-core/pull/5

Stebalien:
Travis is failing due to https://github.com/ipfs/go-ipfs-cmds/issues/143 (I think).

PR #6009, opened 2019-02-21
gx: update go-bitswap and go-libp2p-kad-dht

Stebalien:
* go-bitswap: fix some race conditions.
* go-libp2p-kad-dht: fix a goroutine leak.


Stebalien:
(merging as this is just a gx update)

PR #6012, opened 2019-02-22
Introducing EncodedFSKeystore with base32 encoding

AuHau:
Encoding the key's filename with base32 introduces coherent behavior
across different platforms and their case-sensitive/case-insensitive
file-systems. Moreover, it allows a wider character set to be used for the
name of the keys as the original restriction for special FS's characters
(e.g. '/', '.') will not apply.

Tightly coupled with migration: https://github.com/ipfs/fs-repo-migrations/pull/84

### DIFF
--- a/keystore/keystore.go
+++ b/keystore/keystore.go
@@ -9,6 +9,7 @@ import (
 
 	logging "github.com/ipfs/go-log"
 	ci "github.com/libp2p/go-libp2p-crypto"
+	base32 "github.com/whyrusleeping/base32"
 )
 
 var log = logging.Logger("keystore")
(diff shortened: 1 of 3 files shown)


AuHau:
The failing test does not seem to be related to this PR.

Btw. I  started to really like this feature, because it will enable to namespace the key's names like: `websites/adam-uhlir.me`

Stebalien:
Thanks! @Kubuxu can you take a look at this? If it looks good, we can merge it after the release freeze (although we may want to coordinate with #5870 as that will also require a migration).

Kubuxu:
Agree with after release, also I would wait with merging/migration until we can collect everything we might what to migrate.

AuHau:
Fixed conflict. 

Also one question, I have used `mbase` for the encoding/decoding instead of directly using base32. What do you think about it, is it alright? It sort of made sense to me...

Stebalien:
For consistency, it might be better to just use base32. Using multibase doesn't _buy_ us anything in this case.

Kubuxu:
The only blocker is migrations at this point.

AuHau:
Migration is almost ready at: https://github.com/ipfs/fs-repo-migrations/pull/84/
Only tests are missing, I will focus on them today and hopefully, all should be ready then.

Kubuxu:
We will have other things in the migration: https://github.com/ipfs/fs-repo-migrations/issues/85

hsanjuan:
Continuing work at https://github.com/ipfs/go-ipfs/pull/6955

Thanks @AuHau for this!

PR #6019, opened 2019-02-25
coreapi unixfs: change Wrap logic to make more sense 

magik6k:
Currently `coreunix.Adder` wrap logic is weird, which causes `coreapi.Unixfs().Add` to also be weird:

| data          | wrap | hidden | hash               |
|---------------|------|--------|--------------------|
| (file)        | :x:    | :x:      | (QmFile)           |
| (file)        | :heavy_check_mark:   | :x:      | (QmWrapper)/QmFile |
| (dir)/(file)  | :x:    | :x:      | (QmFile)           |
|  (dir)/(file) | :heavy_check_mark:    | :x:      | (QmDir)/filename   |
| (dir)/(.file) | :heavy_check_mark:    | :x:      | (QmDir)/.filename  |
| (dir)/(.file) | :heavy_check_mark:    | :heavy_check_mark:      | (QmDir)/.filename  |

With this PR it does this:

| data          | wrap | hidden | hash                          |
|---------------|------|--------|-------------------------------|
| (file)        | :x:    | :x:      | (QmFile)                      |
| (file)        | :heavy_check_mark:    | :x:      | (QmWrapper)/QmFile            |
| (dir)/(file)  | :x:    | :x:      | (QmDir)/filename              |
|  (dir)/(file) | :heavy_check_mark:    | :x:      | (QmWrapper)/QmDir/filename    |
| (dir)/(.file) | :heavy_check_mark:    | :x:      | (QmWrapper)/QmEmptyDir/       |
| (dir)/(.file) | :heavy_check_mark:    | :heavy_check_mark:      | (QmWrapper)/(QmDir)/.filename |

New behaviour IMO makes much more sense from API consumer perspective.

TODO:
* [ ] Cleanup Adder code
* [x] Fix regressions
* [x] Merge https://github.com/ipfs/interface-go-ipfs-core/pull/10
* [x] Merge https://github.com/ipfs/go-ipfs-cmds/pull/157

Reviews welcome


Kubuxu:
As far as I know, `--warp` main use case is/was preserving the name of a file when a file is added.

I can see how it is a bit weird that it doesn't happen in case of directories but I have to highlight that it is quite a big breaking change and I would expect would break existing workflows and scripts.

magik6k:
`ipfs add --wrap` won't be affected by this change, I just didn't fix it yet.

core/coreapi/unixfs.go
```diff
@@ -94,7 +94,9 @@ func (api *UnixfsAPI) Add(ctx context.Context, files files.Node, opts ...options
 	fileAdder.RawLeaves = settings.RawLeaves
 	fileAdder.NoCopy = settings.NoCopy
 	fileAdder.Name = settings.StdinName
+	fileAdder.BaseName = settings.BaseName
 	fileAdder.CidBuilder = prefix
+	fileAdder.TopHidden = settings.TopHidden
```
Stebalien: See https://github.com/ipfs/interface-go-ipfs-core/issues/17. I'm not sure if we really need this.

(yes, it'll be a breaking change for people using the HTTP API manually but I'm pretty sure they'd be _very_ surprised at the current behavior).

core/coreunix/add.go
```diff
@@ -68,12 +68,14 @@ type Adder struct {
...
 	Progress   bool
 	Hidden     bool
+	TopHidden  bool
 	Pin        bool
 	Trickle    bool
 	RawLeaves  bool
 	Silent     bool
 	Wrap       bool
 	Name       string
+	BaseName   string
```
Stebalien: Has this been pushed to the interface?

Stebalien: I'm also not sure if we really need to wire this all the way through. If we're _just_ wiring this through to name the event, we can probably apply the name in the add command (that is, it's a concern of the current HTTP API, not really of the core API).

magik6k: The end goal is to not have this. I did this only to see if the main change would fundamentally break the `ipfs add` logic.

Stebalien: Got it.

core/commands/add.go
```diff
@@ -185,6 +187,32 @@ You can now check what blocks have been created by:
 
 		events := make(chan interface{}, adderOutChanSize)
 
+		var toadd files.Node = req.Files
+		name := ""
+		if !wrap || pathName != "" {
```
Stebalien: IIRC, `pathName` is _just_ for stdin. That is, `echo foobar | ipfs add --stdin-name=foo`. I wonder if this should be handled at the commands-lib layer as well...

Stebalien: This actually relates to:

> Another problem I faced but that is not covered by a RFC: when adding content that is not a file (say ipfs add with a data buffer), the body part header Content-Disposition doesn't have a proper filename parameter. While this is legal according to the RFCs, the consequence is that those part are parsed as form values (think text field entered by a user), and not files.

(https://github.com/ipfs/ipfs/issues/395)

We should handle this client-side (technically a breaking change but minor enough that I'm fine with it).

core/coreunix/add.go
```diff
@@ -321,42 +266,92 @@ func (adder *Adder) AddAllAndPin(file files.Node) (ipld.Node, error) {
...
+	root = rootdir
+
+	err = root.Flush()
+	if err != nil {
+		return nil, err
+	}
+
+	// if adding a file without wrapping, swap the root to it (when adding a
+	// directory, mfs root is the directory)
+	_, dir := file.(files.Directory)
```
Stebalien: I'm having trouble understanding this comment. Why do we care if the file we're adding is a directory?

magik6k: It made sense before I rewrote this code.

Stebalien:
Here's my attempt to simplify this: https://github.com/ipfs/go-ipfs/pull/6121. It may be totally wrong but it'll at least help me better understand how this all works.

magik6k:
> Why do I care if a file is a directory or not when adding it?

When adding a directory, we add its entries to mfs (so mfs root is the directory). When adding a file, we also add it to mfs with its hash as the name so that wrap works.

If we could drop wrap from adder/coreapi it would make this much simpler (we wouldn't use mfs at all when adding single files)

Stebalien:
> If we could drop wrap from adder/coreapi it would make this much simpler (we wouldn't use mfs at all when adding single files)

Unless you can think of a reason not to, let's just do that. It's really trivial to wrap a file anyways.

go.mod
```diff
@@ -33,12 +33,12 @@ require (
 	github.com/ipfs/go-ipfs-blocksutil v0.0.1
 	github.com/ipfs/go-ipfs-chunker v0.0.1
 	github.com/ipfs/go-ipfs-cmdkit v0.0.1
-	github.com/ipfs/go-ipfs-cmds v0.0.2
+	github.com/ipfs/go-ipfs-cmds v0.0.3
```
jmank88: Needs `v0.0.4` to cover #6118

PR #6024, opened 2019-02-26
fix: show interactive output from install.sh

olizilla:
When trying out the latest rc, i used the `./install.sh`
It hung, with no output. I removed the 2> /dev/null and
it turned out mv was waiting for user input to confirm
the change of file permssions from 555 to 755.

```console
override r-xr-xr-x  oli/staff for /usr/local/bin/./ipfs? (y/n [n])
```

This PR removes piping the output from mv to /dev/null
as it seems like the safest fix.

An alternative would be to just add -f but i've err'd
on the side of caution.

If also tweaked the second condition based on the
recommendations of shellcheck
see: https://github.com/koalaman/shellcheck/wiki/SC2166

License: MIT

### DIFF
--- a/cmd/ipfs/dist/install.sh
+++ b/cmd/ipfs/dist/install.sh
@@ -13,11 +13,11 @@ binpaths="/usr/local/bin /usr/bin"
 is_write_perm_missing=""
 
 for binpath in $binpaths; do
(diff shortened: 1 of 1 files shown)


Stebalien:
Thanks!

PR #6018, opened 2019-02-25
chore: update to Web UI 2.4.0

fsdiogo:
Check the [release notes](https://github.com/ipfs-shipyard/ipfs-webui/releases/tag/v2.4.0) for more info.

It would be awesome if this could be baked into the new release 😇

PR #6025, opened 2019-02-26
Revert "chore: update to Web UI 2.4.0"

Stebalien:
Reverts ipfs/go-ipfs#6018

Opt-out analytics of any form is not OK.

### DIFF
--- a/core/corehttp/webui.go
+++ b/core/corehttp/webui.go
@@ -1,12 +1,11 @@
 package corehttp
 
 // TODO: move to IPNS
-const WebUIPath = "/ipfs/Qmb6o8HKrouK2WXVFS8753BArqnx5xz68tk6779ncPxpd8"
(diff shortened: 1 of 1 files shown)


Stebalien:
@olizilla, @fsdiogo 

Opt-out analytics is not OK in the WebUI. The "do not track" awareness is irrelevant because that's still an opt-out and most users don't even know what it is.

Next time, please mention a change like this _very clearly_. This could _seriously_ damage our relationship with our users.

olizilla:
@Stebalien understood. Would you accept a new release of web ui that flips the default to opt in? 

It would still be massively helpful to get some analytics from users that choose to opt-in. We already have the UI set up to ask the users permission, we just need to flip the check from "is doNotTrack enabled" to "disabled by default". We can get that done today.

Stebalien:
Not yet. This needs to be thought through and discussed: https://github.com/ipfs-shipyard/ipfs-webui/issues/980.

Issue #5867, opened 2018-12-20
0.4.19 release


Stebalien:
I'd like to ship 0.4.19 before the new year, or at least immediately after. Any PRs/changes to take into account?

PRs that look like they're pretty much ready:

* Bitswap Performance (:yum:): https://github.com/ipfs/go-bitswap/pull/27
* Resolve change that didn't make it into the last release: https://github.com/ipfs/go-ipfs/pull/5704
* CID base: https://github.com/ipfs/go-ipfs/pull/5789, depends on https://github.com/ipfs/go-ipfs/pull/5854

I'd also like to fix: https://github.com/libp2p/go-libp2p/issues/510 first, if possible.

Major user-facing changes as I see them:

* Experimental NAT detection and autorelay.
* Bug fixes.

hannahhoward:
What's the plan on this one? Just curious. I'd love to ship the BS improvements :)

hannahhoward:
> BS improvements :)

I mean Bitswap... but might be an apt description.

Stebalien:
We just ran into an issue with Textile around https://github.com/libp2p/team-mgmt/issues/15#issue-380927291. I'm planning on talking with them today.

Unfortunately, my week has been captured by the libp2p team week.

ianopolous:
Thanks for the heads up @Stebalien

Stebalien:
Status:

* Release notes in #6001.
* Blocked on https://github.com/ipfs/go-bitswap/issues/73.

Stebalien:
Blocked by https://github.com/ipfs/go-ipfs/issues/6007.

parkan:
are we still blocked by something here? the dependency issues mentioned itt look to be all closed

Kubuxu:
@parkan we have released first release candidate and we will wait a few days to collect feedback.

Stebalien:
Release pending in https://github.com/ipfs/go-ipfs/pull/6026.

Stebalien:
Ok, so, a few more blockers:

* https://github.com/libp2p/go-libp2p-kad-dht/issues/276
* https://github.com/ipfs/go-bitswap/pull/87

Stebalien:
And another! https://github.com/ipfs/go-ipfs/issues/6028

Stebalien:
[...]

PR #6026, opened 2019-02-26
Release 0.4.19

Stebalien:
I don't anticipate any other changes. I'll merge this and cut the release
tomorrow if nothing happens between now and then.


magik6k:
https://github.com/ipfs/go-ipfs/pull/6019 will introduce API breaking change, but I don't think it's worth blocking release on that. (And there is still this comment https://github.com/ipfs/interface-go-ipfs-core/blob/master/unixfs.go#L39)

Stebalien:
Meh, I consider the CoreAPI to be unstable/internal for now.

Kubuxu:
What about other blockers https://github.com/ipfs/go-ipfs/issues/5867#issuecomment-467676213

Stebalien:
@Kubuxu I'll rebase this once we've merged fixes for those (and possibly cut another RC).

PR #6027, opened 2019-02-27
Gx Bubble. libp2p-6.0.38

hsanjuan:
Includes latest go-libp2p-peer and go-multiaddr.

Per IRC, do not merge until after go-ipfs 0.4.19 stable is released.


hsanjuan:
glad to have a small contribution in the new release :)

PR #6030, opened 2019-02-27
gx: update go-ipfs-cmds, go-bitswap, go-libp2p-kad-dht, and go-mplex

Stebalien:
Fixes the latest batch of bugs found in RC testing.


Stebalien:
(gx update: doesn't need a review)

PR #6034, opened 2019-03-01
commands(dht): return final error

### LINKED ISSUE #6032: missing error output when `ipfs dht put` with invalid record key type
#### Type: bug

When I run command `ipfs dht put temp temp` ， I find  there is a  `invalid record keytype ` error  in code, but why it disappear in my console.

Stebalien:
This error has always been exposed as a value (visible with the `-v` flag) but we should also be returning it as a final error.

fixes #6032


Stebalien:
Tests are failing due to the interaction between the two following bugs:

* https://github.com/libp2p/go-libp2p-routing-helpers/pull/19
* https://github.com/libp2p/go-libp2p-pubsub-router/issues/23

Stebalien:
If someone could review https://github.com/libp2p/go-libp2p-routing-helpers/pull/19, we can unblock and merge this.

PR #6040, opened 2019-03-03
Revert "Really run as non-root user in docker container"

hsanjuan:
Reverts ipfs/go-ipfs#5048 (see discussion there).


Stebalien:
Thanks for catching this!

PR #6051, opened 2019-03-05
fix ls command to use the new coreinterface types

Stebalien:
See: https://github.com/ipfs/interface-go-ipfs-core/pull/14


core/coreapi/unixfs.go
```diff
@@ -170,26 +170,27 @@ func (api *UnixfsAPI) Ls(ctx context.Context, p coreiface.Path, opts ...options.
...
 	}
 
-	switch lnk.Link.Cid.Type() {
+	switch lnk.Cid.Type() {
 	case cid.Raw:
 		// No need to check with raw leaves
 		lnk.Type = coreiface.TFile
-		lnk.Size = lnk.Link.Size
+		lnk.Size = linkres.Link.Size
 	case cid.DagProtobuf:
```
magik6k: This should also fill `lnk.Target` for symlinks.

Kubuxu:
Sharness is not happy.

PR #6052, opened 2019-03-05
feat: make ls only work for directories

Stebalien:
Long ago, `ls` was the _object_ (merkledag) level listing tool. That's no longer
the case.

This is a [breaking change] but I don't think anyone will object.

This is a start on all the "merge `ls` and `file ls` issues.


magik6k:
`sharnessLinux.t0045-ls` needs fixing.

gammazero:
At this point it does not make sense to revert to the old behavior as the current behavior is what is expected. The associated issue has been closed. If there is still a need for this, then please open an issue in the boxo repo with an explanation why this is needed.

Issue #6056, opened 2019-03-06
make install fails - checksum mismatch


reardenlife:
Tried to upgrade from v0.4.19-dev to v0.4.19.

```sh

[root@v48807 go-ipfs]# pwd
/root/go/src/github.com/ipfs/go-ipfs
[root@v48807 go-ipfs]# git pull
Already up-to-date.
[root@v48807 go-ipfs]# make install
bin/check_go_version 1.11
go install -ldflags="-X "github.com/ipfs/go-ipfs".CurrentCommit=918dcbe" ./cmd/ipfs
go: downloading github.com/gogo/protobuf v1.2.1
go: downloading github.com/fd/go-nat v1.0.0
go: downloading github.com/whyrusleeping/go-notifier v0.0.0-20170827234753-097c5d47330f
go: downloading github.com/btcsuite/btcd v0.0.0-20190213025234-306aecffea32
go: downloading github.com/polydawn/refmt v0.0.0-20190221155625-df39d6c2d992
go: downloading github.com/cenkalti/backoff v2.1.1+incompatible
go: downloading github.com/miekg/dns v1.1.4
go: downloading github.com/jackpal/gateway v1.0.4
go: downloading github.com/huin/goupnp v0.0.0-20180415215157-1395d1447324
go: downloading github.com/whyrusleeping/go-smux-multistream v2.0.2+incompatible
go: downloading github.com/libp2p/go-libp2p-netutil v0.0.1
go: downloading github.com/libp2p/go-libp2p-transport-upgrader v0.0.1
go: downloading github.com/libp2p/go-libp2p-discovery v0.0.1
go: downloading github.com/dgraph-io/badger v1.5.5-0.20190226225317-8115aed38f8f
go: downloading github.com/libp2p/go-conn-security v0.0.1
go: downloading github.com/spaolacci/murmur3 v0.0.0-20180118202830-f09979ecbc72
go: verifying github.com/polydawn/refmt@v0.0.0-20190221155625-df39d6c2d992: checksum mismatch
        downloaded: h1:G16v6hoz+26YUMfvEvUkBKeJeIvsljFH2RRFY3MeXUA=
        go.sum:     h1:bzMe+2coZJYHnhGgVlcQKuRy4FSny4ds8dLQjw5P1XE=
make: *** [cmd/ipfs-install] Error 1

```

Kubuxu:
Can you try running `go clean -modcache`?

chenminjian:
[...]

PR #6057, opened 2019-03-06
go: require Golang 1.11.4

Kubuxu:
mod.sum generation had a fix in this version which probably is the
reason for #6056

License: MIT


Kubuxu:
I've confirmed that the issue exists with Go 1.11.3

PR #6062, opened 2019-03-07
Add gxdeps target

bmwiedemann:
Add gxdeps target
Also do not build gx from source, if it is already available on the system.

In a test this also pulled in deps some other way - please help improve this PR


Kubuxu:
We are transitioning off `gx` to go modules. The install unsporrted target can go away as we no longer use gx.

Kubuxu:
Implemented in https://github.com/ipfs/go-ipfs/pull/6063

Issue #6067, opened 2019-03-08
Immediately provide root node when running `ipfs add <cid>` and `ipfs pin add <cid>`


michaelavila:
All providing currently takes place in `go-bitswap` and provides no guarantees to provide roots of a dag first. This means, a sufficiently large file, will take a while before being available on the ipfs gateways.

Can we provide roots in `ipfs add <cid>` and `ipfs pin add <cid>` right away while `go-bitswap` does its providing work?

michaelavila:
Closing this issue as it was resolved by #6068

PR #6068, opened 2019-03-08
Provide root node immediately on add and pin add

michaelavila:
Provide root node immediately when `ipfs add <cid>` and `ipfs pin add <cid>`.

This PR introduces part of the new provider subsystem from the rest of the ongoing provider work so it's more extensive than it might otherwise have been. This seemed like a good way to merge some of that work and begin using it. And he work to get this PR mergeable will get folded back into #5870 

Addresses: https://github.com/ipfs/go-ipfs/issues/6067

### DIFF
--- a/provider/provider.go
+++ b/provider/provider.go
@@ -8,7 +8,6 @@ import (
 	"github.com/ipfs/go-cid"
 	logging "github.com/ipfs/go-log"
 	"github.com/libp2p/go-libp2p-routing"
-	"time"
 )
 
 var (
@@ -17,7 +16,6 @@ var (
 
 const (
 	provideOutgoingWorkerLimit = 8
-	provideOutgoingTimeout     = 15 * time.Second
 )
 
 // Provider announces blocks to the network, tracks which blocks are
@@ -75,14 +73,11 @@ func (p *Provider) handleAnnouncements() {
 func doProvide(ctx context.Context, contentRouting routing.ContentRouting, key cid.Cid) error {
 	// announce
 	log.Info("announce - start - ", key)
-	ctx, cancel := context.WithTimeout(ctx, provideOutgoingTimeout)
 	if err := contentRouting.Provide(ctx, key, true); err != nil {
 		log.Warningf("Failed to provide cid: %s", err)
 		// TODO: Maybe put these failures onto a failures queue?
-		cancel()
 		return err
 	}
-	cancel()
 	log.Info("announce - end - ", key)
 	return nil
 }
--- a/provider/provider.go
+++ b/provider/provider.go
@@ -11,13 +11,9 @@ import (
 	routing "github.com/libp2p/go-libp2p-routing"
 )
 
-var (
-	log = logging.Logger("provider")
-)
+var	log = logging.Logger("provider")
 
-const (
-	provideOutgoingWorkerLimit = 8
-)
+const provideOutgoingWorkerLimit = 8
 
 // Provider announces blocks to the network
 type Provider interface {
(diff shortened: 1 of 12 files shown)


provider/provider.go
```diff
@@ -0,0 +1,70 @@
...
+func (p *provider) Provide(root cid.Cid) error {
+	p.queue.Enqueue(root)
+	return nil
+}
+
+// Handle all outgoing cids by providing (announcing) them
+func (p *provider) handleAnnouncements() {
+	for workers := 0; workers < provideOutgoingWorkerLimit; workers++ {
+		go func() {
+			for {
```
Stebalien: Nit: This should probably check `p.ctx.Err() != nil`. Unfortunately, `select` selects from it's channels _randomly_ so this could keep dequeueing, even on shutdown.

(probably not going to happen but still something we should get better at doing, this is a "do what I say, not what I do" kind of comment).

provider/queue.go
```diff
@@ -0,0 +1,181 @@
...
+	}()
+}
+
+func (q *Queue) queueKey(id uint64) datastore.Key {
+	return datastore.NewKey(strconv.FormatUint(id, 10))
+}
+
+// crawl over the queue entries to find the head and tail
+func getQueueHeadTail(ctx context.Context, name string, datastore datastore.Datastore) (uint64, uint64, error) {
+	q := query.Query{}
```
Stebalien: So, there's a _much_ faster way to do this. We can:

1. Set `Query.Prefix` to our key's prefix. That'll ensure that we only crawl over that prefix.
2. We can set `Query.Order = []query.Order{query.OrderByKey}` and then pull off the first item to get the head.
3. We can set a limit of 1 (I think?).

Note: We'll have to encode our numbers with a fixed-width (e.g., 16 hex characters) to make this work as we use lexical ordering.

Stebalien: (we can perform a second query to find the tail)

provider/provider.go
```diff
@@ -0,0 +1,70 @@
...
+)
+
+var log = logging.Logger("provider")
+
+const provideOutgoingWorkerLimit = 8
+
+// Provider announces blocks to the network
+type Provider interface {
+	Run()
+	Provide(cid.Cid) error
```
Stebalien: (docs please)

provider/queue.go
michaelavila: @Stebalien I want to get a quick priority check on this given the fact that we need to add features to the query mechanism for mount in datastore. (see https://github.com/ipfs/go-datastore/pull/121). I don't mind continuing that work, but I wonder, can we get this work merged and out while working on the datastore (or whatever is the most important thing to do next) and patch a more performant solution when the prerequisite work is complete?

Stebalien: If we set the query _prefix_ so we only have to walk over the records in the provider queue, that should be enough for now.

michaelavila: @Stebalien we actually have the whole datastore namespaced and that's why the query doesn't have a prefix. My understanding is that this will have the affect you're looking for. Is that true?

Stebalien: Got it. Then the only thing left on this PR is docs around Provide/Run (sorry, being picky about this is one of _my_ KRs).

Stebalien:
@michaelavila let's add some docs and then we can merge this and fix the dht query stuff later.

provider/queue.go
michaelavila: Don't be sorry, I appreciate it and I'm working on those now. I just focused on the bigger issue to get it resolved ASAP.

provider/provider.go
michaelavila: Is this what you're talking about? doing the check at the beginning of the loop and returning if we have an error?

michaelavila:
@Stebalien I believe all feedback was correctly accounted for. Please advise if not.

provider/provider.go
Stebalien: Ah, sorry. I literally just meant:

```
for p.ctx.Err() != nil { /* ... */ }
```

The context being canceled isn't an error or anything, it just means we need to stop (so we shouldn't be printing a warning). I only suggested this because the select statement may repeatedly select the dequeue case if both that and the `ctx.Done()` case are ready at the same time.

michaelavila:
@Stebalien Ok, thanks for looking. Updated as suggested. One note: my conditional is the opposite of yours, but I think yours was just a typo.

Stebalien:
:tada:

PR #6069, opened 2019-03-09
pluggable keystores

Stebalien:
This patch allows keystores to be configured via the config and plugins.

It's completely untested for now, but I started this so I figured I'd get it to a point where we could consider finishing it.

Depends on: ...

CC @postables


postables:
Oh snap this is **awesome**! :fireworks: Let me try and integrate the keystore I wrote today and see if I come across any unexpected issues.

repo/fsrepo/fsrepo.go
```diff
@@ -384,8 +384,23 @@ func (r *FSRepo) openConfig() error {
...
 
 func (r *FSRepo) openKeystore() error {
-	ksp := filepath.Join(r.path, "keystore")
-	ks, err := keystore.NewFSKeystore(ksp)
+	spec := r.config.Keystore
+	if spec == nil {
+		spec = map[string]interface{}{"type": "files", "path": "keystore"}
+	}
+
+	ksType, ok := spec["type"]
```
postables: Experimenting with this locally gives me:
```
../../../../pkg/mod/github.com/ipfs/go-ipfs@v0.0.0-20190309025046-b72a340c987a/repo/fsrepo/fsrepo.go:397:25: cannot use ksType (type interface {}) as type string in map index: need type assertion
```

Might need to do ?
```suggestion
	ksType, ok := spec["type"].(string)
```

aschmahmann:
ping @Stebalien what's the status of this and is this something we still want?

PR #6071, opened 2019-03-10
add string type assert to openKeystore

postables:
Adds a string type assertion to `fsrepo.openKeystore` when determining the keystore type

edit: for https://github.com/ipfs/go-ipfs/pull/6069


Stebalien:
Thanks!

PR #6073, opened 2019-03-11
update go-ipfs-config for feat/pluggable-keystore

bonedaddy:
This should fix the failing ci test on https://github.com/ipfs/go-ipfs/pull/6069

I noticed https://github.com/ipfs/go-ipfs-config/pull/29 was failing as well, I'd be happy to take a look at that if wanted/needed.

edit: also noticed the last two PRs i opened caused git-cop to fail, let me know if I should sign to fix this


aschmahmann:
Thanks, this is out of date and #6069 will probably need some work to get this closed.

PR #6087, opened 2019-03-16
resolve: fix recursion

### LINKED ISSUE #4181: default to recursive in ipfs name resolve
ref #4167

### LINKED ISSUE #4293: api call to lookup dnslink name doesn't work
#### Version information:

````
~ ❯❯❯ ipfs version --all                                                    Mon 2017-10-09 14:27:51 -0700
go-ipfs version: 0.4.11-dev-
Repo version: 5
System version: amd64/openbsd
Golang version: go1.8
````

#### Description:

Using the API to resolve ggr.com doesn't work, i.e.:

````
https://ipfs.io/api/v0/dns/ggr.com
````

returns

````
{"Message":"Could not resolve name (recursion limit exceeded).","Code":0}
````
Yet, the name resolves at the CLI fine:
````
~/Downloads ❯❯❯ ipfs --debug name resolve -r ggr.com                        Mon 2017-10-09 12:09:19 -0700
14:09:29.683  INFO   cmd/ipfs: IPFS_PATH /home/scott/.ipfs main.go:301
14:09:29.687 DEBUG   cmd/ipfs: calling pre-command hooks... main.go:295
14:09:29.687 DEBUG   cmd/ipfs: executing command via API main.go:332
/ipfs/QmWqNxbUnGZzmLeyA8jMi1VeyuHpBcYDYH9WcYZY31zijk
````
See https://github.com/ipfs/ipfs-companion/issues/288 for a previous discussion and more detail on this issue.

### LINKED ISSUE #5585: http api/v0/dns says "not a valid domain name", anything else works fine
#### Version information:
go-ipfs version: 0.4.17-
Repo version: 7
System version: amd64/linux
Golang version: go1.10.3

#### Type:
Bug

#### Description:
I set up my website, registered domain `etam-software.eu` and published dnslink TXT record. Accessing through http://127.0.0.1:8080/ipns/etam-software.eu/ works fine, but ipfs-companion firefox addon is not redirecting to my gateway. I found that http api/v0/dns, which is used by addon, returns error, even though everything else works fine:

```
$ dig +noall +answer TXT etam-software.eu
etam-software.eu.       221     IN      TXT     "dnslink=/ipns/QmanzJhzCXAgry6rT7uyhwdwk9HTDSzVZ4R4mgcBy1nd1j"
$ ipfs name resolve -r etam-software.eu
/ipfs/QmZY4bqaNB8yQUFeVS87QEuNKyTLkfT5eZ1QWwoeozu8xG
$ curl http://127.0.0.1:5001/api/v0/dns/etam-software.eu?r=true
{"Message":"not a valid domain name","Code":0,"Type":"error"}
```

### LINKED ISSUE #5635: Resolution commands rework
1. `ipfs name resolve` and `ipfs dns` should be recursive by default, the current behavior is a surprising foot-gun. We should consider just changing this as I'm sure it'll fix *significantly* more bugs than it causes.
2. `ipfs dns` should resolve up to the first non-dnslink path by default. It shouldn't return an *error* when it hits a non-dnslink path, it should just *return* that path.

I'd also like to return a partial result when we hit the recursion limit (along with the error) but that's less pressing.

---

Can anyone think of anything this would break?

@alanshaw, how will this affect JS? Is this doable?

@whyrusleeping any idea why we aren't recursive by default?

### LINKED ISSUE #6086: ipfs dns throws Error: could not resolve name (recursion limit exceeded)
#### Version information:
go-ipfs version: 0.4.18-
Repo version: 7
System version: amd64/windows
Golang version: go1.11.1

#### Description:

run `ipfs dns ipfs.io` on the terminal 

OR go to: https://ipfs.io/api/v0/dns?arg=ipfs.io

Both are showing the exact same error so I assume they're coming from the same place. 

Expected behavior is that it should give an ipns path something like `/ipns/website.ipfs.io`

Stebalien:
1. Recurse by default. This is what users expect and it should only un-break
applications. We have a default limit of 32.
2. When recursion is turned off, return the partial result. Otherwise, these
commands are pretty much useless

fixes #5635
fixes #5585
fixes #4181
fixes #4293
fixes #6086

### DIFF
--- a/core/commands/dns.go
+++ b/core/commands/dns.go
@@ -71,10 +71,7 @@ The resolver can recursively resolve:
 		}
 
 		output, err := resolver.Resolve(req.Context, name, ropts...)
-		if err == namesys.ErrResolveFailed {
-			return err
-		}
-		if err != nil {
+		if err != nil && (recursive || err != namesys.ErrResolveRecursion) {
 			return err
 		}
 		return cmds.EmitOnce(res, &ncmd.ResolvedPath{Path: output})
--- a/core/commands/dns.go
+++ b/core/commands/dns.go
@@ -58,7 +58,7 @@ The resolver can recursively resolve:
 		cmdkit.StringArg("domain-name", true, false, "The domain-name name to resolve.").EnableStdin(),
 	},
 	Options: []cmdkit.Option{
-		cmdkit.BoolOption(dnsRecursiveOptionName, "r", "Resolve until the result is not a DNS link."),
+		cmdkit.BoolOption(dnsRecursiveOptionName, "r", "Resolve until the result is not a DNS link.").WithDefault(true),
 	},
 	Run: func(req *cmds.Request, res cmds.ResponseEmitter, env cmds.Environment) error {
 		recursive, _ := req.Options[dnsRecursiveOptionName].(bool)
--- a/core/commands/name/ipns.go
+++ b/core/commands/name/ipns.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	cmdenv "github.com/ipfs/go-ipfs/core/commands/cmdenv"
+	namesys "github.com/ipfs/go-ipfs/namesys"
 
 	cmdkit "github.com/ipfs/go-ipfs-cmdkit"
 	cmds "github.com/ipfs/go-ipfs-cmds"
@@ -130,7 +131,7 @@ Resolve the value of a dnslink:
 
 		if !stream {
 			output, err := api.Name().Resolve(req.Context, name, opts...)
-			if err != nil {
+			if err != nil && (recursive || err != namesys.ErrResolveRecursion) {
 				return err
 			}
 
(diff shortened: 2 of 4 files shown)


Stebalien:
CC @Kubuxu and @lidel.

Kubuxu:
@Stebalien what do you think about adding to `ResolvedPath` information about the fact whether the resolution was final or not?

Stebalien:
@Kubuxu I'm not seeing how this relates to `ResolvedPath`. Currently, `ResolvePath` paths are always _fully_ resolved.

Kubuxu:
What I mean: `core/commands/name.ResolvedPath`, IIRC before the commands rework it was possible to get partial path from the API (not via cli but via json).
But I aggree, unless we are able to return error and the partial path it would be problematic.

Stebalien:
>  But I aggree, unless we are able to return error and the partial path it would be problematic.

We could still return the partial path and then an error. _However_, unfortunately, javascript hard-codes whether or not the response to any given command is "streamed" or sent as one blob.

So, partial resolution would require a new flag.

Kubuxu:
Let's move with what we have.

Stebalien:
@Kubuxu So... can you approve this?

Stebalien:
(well, I'll take that as approval)

Kubuxu:
Sorry, I was reviewing codecov report and then forgot to click the approve.

PR #6096, opened 2019-03-18
feat: gateway subdomains + http proxy mode

### LINKED ISSUE #5982: Act as HTTP PROXY for http://<cidv1b32>.ipfs.localhost
#### Version information:

0.4.18

#### Type:

feature

### Problem 

[CID-in-subdomain](https://github.com/ipfs/in-web-browsers/issues/89)  creates Origin-based isolation if website is loaded from public gateway:
 - http://bafkreigh2akiscaildcqabsyg3dfr6chu3fgpregiymsck7e7aqa4s52zy.ipfs.dweb.link

..however we continue to have a single Origin for all websites loaded from local gateway:

- http://127.0.0.1:8080/ipfs/bafkreigh2akiscaildcqabsyg3dfr6chu3fgpregiymsck7e7aqa4s52zy/

This means use of local gateway _decreases_ some of security guarantees web developers are used to, and removes some incentives to run local go-ipfs (and redirect to it via ipfs-companion).

### Solution

What if go-ipfs could act as HTTP PROXY for requests made to `http://*.ipfs.*/` and `http://*.ipns.*/` ?

Having that, IPFS Companion could automatically set Gateway port as PROXY for requests to  `http://*.ipfs.localhost`, which would create very nice URLs for local gateway, solving the Origin problem for local go-ipfs :ok_hand: 

#### Initial look at technical feasibility 

My hope is that we could keep changes to minimum and reuse Gateway port to also act as HTTP PROXY for requests to `*.ipfs.localhost`.

In other words, in addition to regular HTTP GET:

```console
$  curl -v 'http://127.0.0.1:8080/ipfs/bafkreigh2akiscaildcqabsyg3dfr6chu3fgpregiymsck7e7aqa4s52zy/'
> GET /ipfs/bafybeie5gq4jxvzmsym6hjlwxej4rwdoxt7wadqvmmwbqi7r27fclha2va HTTP/1.1
> Host: 127.0.0.1:8080
< HTTP/1.1 200 OK
```

Gateway port should also respond to PROXY requests:

```console
$ curl -v --proxy 'http://127.0.0.1:8080' 'http://bafkreigh2akiscaildcqabsyg3dfr6chu3fgpregiymsck7e7aqa4s52zy.ipfs.localhost'
> GET http://bafkreigh2akiscaildcqabsyg3dfr6chu3fgpregiymsck7e7aqa4s52zy.ipfs.localhost/ HTTP/1.1
> Host: bafkreigh2akiscaildcqabsyg3dfr6chu3fgpregiymsck7e7aqa4s52zy.ipfs.localhost
```

AFAIK all HTTP PROXY needs to do is to support [HTTP CONNECT Method](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/CONNECT) (but I think some non-https proxies work even without it):

```
CONNECT bafkreigh2akiscaildcqabsyg3dfr6chu3fgpregiymsck7e7aqa4s52zy.ipfs.localhost:80 HTTP/1.1 
Host: bafkreigh2akiscaildcqabsyg3dfr6chu3fgpregiymsck7e7aqa4s52zy.ipfs.localhost:80 
```

After looking at golang's `net/http`, my initial intuition says it should be possible to detect such requests early via something like:

```go
http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
    if r.Method == http.MethodConnect {
        handleProxyRequest(w, r)
    } else {
        handleRegularRequest(w, r)
    }
})
```

**Would love to get some feedback on this idea**

I feel this is a huge win for both security and UX, and should be done before we [switch to CIDv1 in base32 by default](https://github.com/ipfs/ipfs/issues/337). 

As a sidenote: this technique opens us a venue for future support of alternative naming systems such as ENS (`*.eth.localhost`) or Namecoin (`*.bit.localhost`). 

### Refs

- Migration to CIDv1 (default base32) https://github.com/ipfs/ipfs/issues/337
  - go-ipfs specific https://github.com/ipfs/go-ipfs/issues/5358
- Why CID in Subdomain? https://github.com/ipfs/in-web-browsers/issues/89
- Origin problem for local gateway at go-ipfs 
  - Redirect of CID-in-subdomain and DNSLink websites: https://github.com/ipfs-shipyard/ipfs-companion/issues/667
  - Vanity hostnames for localhost HTTP gateway: https://github.com/ipfs/in-web-browsers/issues/109

### LINKED ISSUE #6498: Subdomain Gateway
> This is a prerequisite for https://github.com/ipfs/go-ipfs/issues/5982,
> but also a very needed standalone feature.

## Context

[Base32-encoded CIDv1](https://github.com/ipfs/ipfs/issues/337) can be used in DNS as a  subdomain label:

```
https://<cidv1b32>.ipfs.example.com
https://<cidv1b32>.ipns.example.com
```

It provides unique Origin per CID and enables websites to get a valid origin-based security perimeter for free. 

Examples of "Subdomain Gateways" in production:
  - https://bafybeiemxf5abjwjbikoz4mc3a3dla6ual3jsgpdr4cjr3oz3evfyavhwq.ipfs.dweb.link/wiki/
  - https://bafybeiemxf5abjwjbikoz4mc3a3dla6ual3jsgpdr4cjr3oz3evfyavhwq.ipfs.cf-ipfs.com/wiki/ (Cloudflare)

## How it works today? What is missing?

Right now HTTP Gateway exposed by go-ipfs does not support HTTP requests for hostnames that match `<cidv1b32>.ipfs.*` – all existing subdomain gateways implement CID conversion and path redirect in Nginx ([config example](https://github.com/ipfs/infra/issues/81#issuecomment-461045160)) or other reverse proxy running in front of go-ipfs.

## Support requests with `Host: <cidv1b32>.ipfs.*`

go-ipfs already supports [DNSLink](http://dnslink.dev/). We need it to also support HTTP requests for hostnames that match `<cidv1b32>.ipfs.*`. It would enable people to run subdomain gateways without the need for running reverse proxy.

Assuming go-ipfs is running a gateway on `localhost:8080`, below should work:

```
$ curl -v 'http://bafkreigh2akiscaildcqabsyg3dfr6chu3fgpregiymsck7e7aqa4s52zy.ipfs.localhost:8080/wiki/'
> GET /wiki/ HTTP/1.1
> Host: bafkreigh2akiscaildcqabsyg3dfr6chu3fgpregiymsck7e7aqa4s52zy.ipfs.localhost:8080
```

## Related
- Migration to CIDv1 (default base32): https://github.com/ipfs/ipfs/issues/337
- CID as a Subdomain: https://github.com/ipfs/in-web-browsers/issues/89
- HTTP PROXY for `http://<cidv1b32>.ipfs.localhost`: https://github.com/ipfs/go-ipfs/issues/5982
- Gateway index pages link to wrong places when using CID-as-subdomain: https://github.com/ipfs/go-ipfs/issues/5309
- Setup hash subdomain proxying: https://github.com/ipfs/infra/issues/81
- Hostname for localhost gateway: https://github.com/ipfs/in-web-browsers/issues/109

Stebalien:
Depends on ipfs/go-ipfs-config#30

* [ ] TODO: Needs base32 peer IDs. The IPNS and p2p proxy stuff doesn't currently work due to this.
* [ ] Think through the "known gateway" concept.

What does this do?

1. Any HTTP CONNECT (proxy) requests will be looped backed to the gateway. This means a browser can, e.g., proxy `http://ipfs.io/ipfs/...` to the local gateway without redirecting (or touching the URL bar).
2. IPFS can handle CIDs in subdomains. That is, `http://CID.ipfs.localhost:8080`.

fixes #5982 
fixes #6498 
cc https://github.com/ipfs/in-web-browsers/issues/89, https://github.com/ipfs-shipyard/ipfs-companion/issues/667


lidel:
I'm trying to look at this by the lens of what potential user wants.
AFAIK simplified Gateway use cases are:

- (A) run go-ipfs on local machine with default settings
- (B) run _Path Gateway_ on some hostname
- (C) run _DNSLink Gateway_ on some IP
- (D) set up _Subdomain Gateway_ on some hostname

I think (A) should be "batteries included" and not require any configuration, (D) requires more advanced setup in DNS anyway, so it is ok for it to be an opt-in, and there is some vagueness about resolution rules when (B) and (C) happen to use the same hostname, which suggests there should be a way to make both configurable or simply do not allow that scenario on a single hostname.

Having that as a starting point, some notes below (apologies, I failed to make this shorter).

## Simplify configuration of Gateway features

I like the idea of having explicit list of hostnames that impacts the way requests are handled, but we should not hardcode any known `PublicGateways`: [we want people to run their own](https://github.com/ipfs/in-web-browsers/issues/89). Let's keep the list empty and dogfood entire configuration process on our gateways.

Would it be more versatile and easier to reason about if we tweak https://github.com/ipfs/go-ipfs-config/pull/30 and add three clear options to control gateway behavior? Something like:

- `Gateway.DNSLink` – optional boolean to control DNSLink support (default: `true`)
- `Gateway.SubdomainHosts` – optional array of hostnames whitelisted for use in subdomains and proxy mode (+ implicit `localhost`)
- `Gateway.PathHosts` - optional array of hostnames that are known to be path-based Gateways where content paths should take priority over DNSLink (+ implicit `127.0.0.1` and `::1`)

Caveats: 
- These config options replace `UseSubdomains` and `PublicGateways`
- Every hostname on `Gateway.SubdomainHosts` needs to redirect non-subdomain paths (`dweb.link/ipfs/<cid>/foo`) to subdomain ones (`<cid>.ipfs.dweb.link/foo`)
- There needs to be some validation that ensures `SubdomainHosts` and `PathHosts` have no overlap or ignore hostnames that are on both
- If we don't allow DNSLink+Path Gateway on the same hostname (and handle `ipfs.io` via Nginx)  the _Gateway Heuristic_ below get MUCH easier, and I think we no longer need `PathHosts`

## Gateway Heuristic

### Localhost

IIUC there is a golden path for localhost:

- Support `<cid>.<ns>.localhost` out of the box 
  (`localhost` as a permanent entry on `Gateway.SubdomainHosts`)
- requests to `localhost` hostname (without subdomain) need to be redirected to `127.0.0.1` (same path, but on raw IP). This will ensure the Origin(s) of Path and Subdomain gateways on a local machine are separate (and users can have **both**!)

### Public Gateways

I feel I am missing some context on [gateway symlinks mentioned here](https://github.com/ipfs/go-ipfs/blob/5abd144b1ff73a6289fb66cca8a9b9fd20268289/core/corehttp/hostname.go#L70-L75) – what are use cases for that? 

So far I was under impression that mixing namespaces such as DNSLink website with Path Gateway is something we want to move away from. It makes security model unnecessarily hard to reason about and I did not see any use in production apart from `ipfs.io` (which we want to move to `dweb.link` anyway).

That being said:

#### If we allow Paths & DNSLink on the same hostname

We use `Gateway.PathHosts` to control which hostnames prioritize _Path Gateway_ over DNSLink. 
Default use case is hosting websites with DNSLink. DNSLink takes a priority _unless_ gateway hostname is listed in `Gateway.PathHosts`:

- IF `Host` is matching `<cid>.<ns>.example.com` and `example.com` is listed in `Gateway.SubdomainHosts`
  THEN  process request as _Subdomain Gateway_
- ELSE
  - IF path is `/ipfs/`, `/ipns/` or `/api/`
    - IF `example.com` is listed in `Gateway.PathHosts`
      THEN process request as _Path Gateway_
    - ELSE
      - IF `Host` is not IP and DNSLink is present
        THEN process request as DNSLink
      - ELSE process request as _Path Gateway_
  - ELSE process request as DNSLink or return 404

#### If DNSLink always takes a priority

We don't need `Gateway.PathHosts` and it gets much easier:

- IF `Host` is matching `<cid>.<ns>.example.com` and `example.com` is listed in `Gateway.SubdomainHosts`
  THEN  process request as _Subdomain Gateway_
- ELSE
  - IF `Host` is not IP and DNSLink is present
    THEN process request as DNSLink
  - ELSE process request as _Path Gateway_

Personally I'd like to keep it simple and avoid mixing DNSLink with Path Gateway if possible, 
but let me know if I missed any nuance.

## Configuration Examples

### DNSLink

```
Gateway.DNSLink = true
Gateway.SubdomainHosts = []
Gateway.PathHosts = []
```

-  `example.com/foo` will load via DNSLink (if present)
-  `example.com/ipfs/foo` will load via DNSLink (if present)
-  `<cid>.ipfs.example.com/foo` will return 404 Not Found

### Path Gateway

```
Gateway.DNSLink = true
Gateway.SubdomainHosts = []
Gateway.PathHosts = ['example.com']
```

-  `example.com/foo` will load via DNSLink (if  we allow mixing with DNSLink)
-  `example.com/ipfs/<cid>` will load via _Path Gateway_
-  `<cid>.ipfs.example.com/foo` will return a 301 redirect to `example.com/ipfs/<cid>/foo`

### Subdomain Gateway

```
Gateway.DNSLink = true
Gateway.SubdomainHosts = ['example.com']
Gateway.PathHosts = []
```

-  `example.com/foo` will load via DNSLink (if we allow mixing with DNSLink)
-  `example.com/ipfs/<cid>/foo` will return a 301 redirect to `<cid>.ipfs.example.com/foo`
-  `<cid>.ipfs.example.com/` will load via _Subdomain Gateway_ 

Thoughts?

lidel:
@Stebalien would be ok for me to help with this more actively in Q1? 
I really want to unblock this. Below is my proposal regarding config changes.

Our gateways are good (extreme) test case for evaluating configuration proposed in my previous comment, because we support all three types of gateways:

```ini
Gateway.DNSLink = true # accept all, return data for any Host with a valid DNSLink
Gateway.SubdomainHosts = ['dweb.link']
Gateway.PathHosts = ['ipfs.io']
```

Its all good, but, there should be super easy way to **run DNSLink gateway only for specific website**.
There should be an easy way to run `good.org` without running into a risk of people pointing DNS of `thepiratebay.org` at the same servers.

What if we replace the boolean `DNSLink` with a third, optional list (catch-all if list is empty or missing)?

Config of our public gateway would look like this:

```ini
Gateway.DNSLinkHosts = ['*'] # catch-all
Gateway.SubdomainHosts = ['dweb.link']
Gateway.PathHosts = ['ipfs.io']
```

And a person who wants to host only DNSLink of a specific website(s) would use:

```ini
Gateway.DNSLinkHosts = ['*.wikipedia-on-ipfs.org'] # excplicit list of supported Hosts
Gateway.SubdomainHosts = [] # implicit localhost
Gateway.PathHosts = [] # implicit 127.0.0.1
```

Thoughts?

Stebalien:
Unless I'm mistaken, I believe our proposals are mostly isomorphic. `PublicGateways` isn't the list of all public gateways that exist, it's the list of public gateways that the go-ipfs proxy will handle.

We might want a better name.

Given:

```
{
  SubdomainHosts: ["x"],
  PathHosts: ["y"],
}
```

This would be expressed in the current patch as:

```
{
  PublicGateways: {
    "x": {
      "UseSubdomains": true,,
      "PathPrefixes": ["/ipfs", "/ipns"],
    },
    "y": {
      "UseSubdomains": false,
      "PathPrefixes": ["/ipfs", "/ipns"],
    },
  }
}
```

The benefits of doing it this way are:

* No chance of conflicting rules.
* We can add flags as we add optional features to gateways. (not a strong motivation)

> There needs to be some validation that ensures SubdomainHosts and PathHosts have no overlap or ignore hostnames that are on both

That's why I had a single list.

> If we don't allow DNSLink+Path Gateway on the same hostname (and handle ipfs.io via Nginx) the Gateway Heuristic below get MUCH easier, and I think we no longer need PathHosts

There's still an issue of hosting a DNSLink website and a subdomain gateway on the same hostname. In the original proposal, we'd _redirect_ paths listed in `PathPrefixes` to the subdomain gateway.

> I feel I am missing some context on gateway symlinks mentioned here – what are use cases for that?

With symlink support, any DNSLink website could become a "gateway" by creating a symlink to /ipfs. The go-ipfs gateway would.

1. Lookup the dnslink record for the website.
2. Resolve the website's IPFS directory.
3. Lookup /ipns/my.domain/ipfs/...
4. See that /ipns/my.domain/ipfs is actually a symlink to /ipfs.
5. Serve /ipfs/...

> So far I was under impression that mixing namespaces such as DNSLink website with Path Gateway is something we want to move away from. It makes security model unnecessarily hard to reason about and I did not see any use in production apart from ipfs.io (which we want to move to dweb.link anyway).

In general, we do. However, users may still want to have path-based gateways (possibly with javascript disabled via CSP) with a nice homepage.

Stebalien:
> Gateway.DNSLink – optional boolean to control DNSLink support (default: true)

This should definitely be an option.

Stebalien:
> What if we replace the boolean DNSLink with a third, optional list (catch-all if list is empty or missing)?

Good point.

Stebalien:
Thinking about it a bit, I like the way you've split the gateway lists (path/subdomain). I'm just not entirely sure how to handle subdomain gateways + dnslink + redirects.

lidel:
### `dweb.link`: all-in-one gateway edge case

>  I'm just not entirely sure how to handle subdomain gateways + dnslink + redirects.

Our `dweb.link` makes a good test case for this, so let's look at it as an example. 

I believe the setup we want is:

- subdomain gateway at `{cid}.ipfs.dweb.link`
- path gateway at `dweb.link/ipfs/{cid}` that redirects to `{cid}.ipfs.dweb.link`
- DNSLink website at `dweb.link` with info about what IPFS gateway is etc
- convention over configuration, where possible

If we agree on the above, then we see there is only one way to handle the "fallback order":

- enabling subdomain gateway implicitly enables redirects from `dweb.link/ipfs/{cid}` to `{cid}.ipfs.dweb.link`
  - this gives us backward-compatibility with existing tools that accept a URL of "IPFS gateway" without the need for differentiating between subdomain and path based ones  
    (examples: IPFS Companion, MetaMask, _{a browser vendor}_)
  - this means "subdomain > path > dnslink"
    - subdomain takes priority over path
    -  **path-based gateway at  `dweb.link/ipfs/{cid}` takes priority over DNSLink for `dweb.link`**



### Different take on the config 

I like how your version enables us to add more configuration options per hostname.
It is not a strong motivation today, but makes things much more flexible and could save us a lot of headache in the future. Worth considering.

@Stebalien  How does the below convention look like?

```json
"Gateway": {
  "PublicGateways": {
    "dweb.link": {
      "ResolveSubdomains": true
    },
    "ipfs.io": {
      "ResolvePaths": true,
    },
    "en.wikipedia-on-ipfs.org": {
      "ResolveDNSLink": true
    }
  }
}
```

The same as above, but with explicit default values:

```json
"Gateway": {
  "ResolveDNSLink": true, /*implicit*/
  "Hosts": {
    "127.0.0.1": { /*implicit*/
      "ResolveSubdomains": false,
      "ResolvePaths": true,
      "ResolveDNSLink": false
    },
    "::1": { /*implicit*/
      "ResolveSubdomains": false,
      "ResolvePaths": true,
      "ResolveDNSLink": false
    },
    "localhost": { /*implicit*/
      "ResolveSubdomains": true,
      "ResolvePaths": false,
      "ResolveDNSLink": false
    },
    "dweb.link": {
      "ResolveSubdomains": true,
      "ResolvePaths": false, /*implicit*/
      "ResolveDNSLink": true
    },
    "ipfs.io": {
      "ResolveSubdomains": false, /*implicit*/
      "ResolvePaths": true,
      "ResolveDNSLink": true
    },
    "en.wikipedia-on-ipfs.org": {
      "ResolveSubdomains": false, /*implicit*/
      "ResolvePaths": false, /*implicit*/
      "ResolveDNSLink": true
    }
  }
}
```


**Note**: 
- localhost gateways are implicit and enabled by default
	- IP ones are path-based
    - `localhost` domain is used for subdomains
    - we need both types to work: IPFS Companion will detect OS support for localhost subdomains and enable it as the default
- `ResolveSubdomains` takes priority over  `ResolvePaths` if both are `true`. Path redirect to Subdomain is always present (even if `ResolvePaths: false`)

Stebalien:
@lidel LGTM! Go for it.

alanshaw:
My only issue with this format is that for most people the default config will be correct and so the host config for each host will just be an empty object.

Maybe we could drop the word "Resolve" from "ResolvePaths", "ResolveDNSLink" etc.?

Otherwise this LGTM!

lidel:
gateway: localhost subdomains with cid conversion
    
    - "known gateways" can be now defined with or without port.
      if port is missing, a version defined without it will be used
      as a fallback
    - subdomain gateway at *.ipfs.localhost is supported out of the box
    - subdomain gateway redirects path-gateway requests to subdomain version
      with automatic conversion to CIDv1 in Base32
      (try http://localhost:8080/ipfs/QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR)


Initial success with `*.ipfs.localhost` working out of the box (regular gateway, not proxies):

IPFS CIDv1:


DNSLink websites:
vincent-2020-01-23--15-59-37


Even ENS looks cool:
eth-2020-01-23--15-57-32

Next steps: testing proxy mode, then tweaking config keys and adding docs.

docs/config.md
```diff
@@ -318,6 +319,110 @@ location /blog/ {
...
+
+`PublicGateways` is a dictionary for overriding default options on specific fully qualified domain names (FQDN).  
+Options for changing behavior per hostname:
+
+- `NoDNSLink`  
+  Overrides global setting of the same name.  
+  Default: `false`
+- `PathPrefixes`  
+  Array of paths that should be mounted at the root of domain name.  
+  **TODO: this does not override global 'PathPrefixes' used for validation of X-Ipfs-Gateway-Prefix header. Should this have different name? eg. PathMounts?**
```
Stebalien: Or just "Paths"?

docs/config.md
```diff
@@ -318,6 +319,110 @@ location /blog/ {
...
+  Example: `["/ipfs", "/ipns", "/api"]`  
+  Default: `[]`
```
Stebalien: Are we sure we want this to default to empty?

docs/config.md
```diff
@@ -318,6 +319,110 @@ location /blog/ {
...
+  - **Note:** when both `PathPrefixes` and DNSLink are mounted, `PathPrefixes` take priority
+- `UseSubdomains`
+  A boolean to configure whether the gateway provides [Origin isolation](https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy)
+  between content roots.  
+  Default: `false`
+  - `true` – mount [subdomain gateway](#https://docs-beta.ipfs.io/how-to/address-ipfs-on-web/#subdomain-gateway) at `http://{cid}.ipfs.{hostname}`
+    - requests for `http://{hostname}/ipfs/{cid}` will return redirect to `http://{cid}.ipfs.{hostname}`
+  - `false` – mount [path gateway](https://docs-beta.ipfs.io/how-to/address-ipfs-on-web/#path-gateway) at `http://{hostname}/ipfs/{cid}`
+    - **(not implemented yet)** due to the lack of Origin isolation, cookies and storage will be disabled by [Clear-Site-Data](https://github.com/ipfs/in-web-browsers/issues/157) header
```
Stebalien: Nice!

lidel: I think I'll drop it from the scope fo this PR and add `Clear-Site-Data` in a separate one as it may require adding new flag (as noted in https://github.com/ipfs/in-web-browsers/issues/157#issuecomment-580341836)

docs/config.md
lidel: I renamed `PathMounts` to `Paths` and removed it from global scope.
Our "know gateways" such as `localhost` are initialized with  `Paths: ["/ipfs", "/ipns"]`:

https://github.com/ipfs/go-ipfs/blob/feat/gateway-subdomains/core/corehttp/hostname.go#L25-L31

For hostnames added by the user, I think  `[]` is a safer default: it requires user to explicitly expose `/ipfs/`, reducing the number of misconfigured DNSLink websites that  act as a public gateways by mistake.

lidel:
Updated config examples:
https://github.com/ipfs/go-ipfs/blob/feat/gateway-subdomains/docs/config.md#gatewaypublicgateways

docs/config.md
```diff
@@ -318,6 +319,100 @@ location /blog/ {
...
+
+### Gateway.PublicGateways
+
+`PublicGateways` is a dictionary for defining gateway behavior on specific fully qualified domain names (FQDN).  
+Options for changing behavior per hostname:
+
+- `Paths`  
+  Array of paths that should be mounted at the root of domain name.  
+  Example: `["/ipfs", "/ipns", "/api"]`  
+  - **Note:** when both `Paths` and DNSLink are mounted, `Paths` takes priority
```
Stebalien: DNSLink mounted?

docs/config.md
```diff
@@ -318,6 +319,100 @@ location /blog/ {
...
+
+- `UseSubdomains`
+  A boolean to configure whether the gateway provides [Origin isolation](https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy)
+  between content roots.  
+  Default: `false`
+  - `true` – mount [subdomain gateway](#https://docs-beta.ipfs.io/how-to/address-ipfs-on-web/#subdomain-gateway) at `http://{cid}.ipfs.{hostname}`
```
Stebalien: nit: shouldn't be specific to `ipfs` (ipns, api, etc?).

docs/config.md
```diff
@@ -318,6 +319,100 @@ location /blog/ {
...
+    - requires respective `Paths` to be set
```
Stebalien: It does? I thought we only needed the path to be set for a redirect to happen. Also, are we using `http://api.hostname`? Or just `hostname/api`? I assume the latter.

docs/config.md
```diff
@@ -318,6 +319,100 @@ location /blog/ {
...
+    - requests for `http://{hostname}/ipfs/{cid}` will return redirect to `http://{cid}.ipfs.{hostname}`
+  - `false` – mount [path gateway](https://docs-beta.ipfs.io/how-to/address-ipfs-on-web/#path-gateway) at `http://{hostname}/ipfs/{cid}`
+    <!-- **(not implemented yet)** due to the lack of Origin isolation, cookies and storage on `Paths` will be disabled by [Clear-Site-Data](https://github.com/ipfs/in-web-browsers/issues/157) header -->
```
Stebalien: Can we do this now? This sounds like it's pretty easy to do.

docs/config.md
```diff
@@ -318,6 +319,100 @@ location /blog/ {
...
+  A boolean to configure whether DNSLink lookup for value in `Host`
+  HTTP header should be performed.  If DNSLink is present, content path stored
+  in the DNS TXT record is mounted at `/` and respective payload is returned to
+  the client.
+
+Default: `false`
+
+### Gateway.PublicGateways
+
+`PublicGateways` is a dictionary for defining gateway behavior on specific fully qualified domain names (FQDN).
```
Stebalien: Are we doing per-gateway dnslink settings?

docs/config.md
```diff
@@ -318,6 +319,100 @@ location /blog/ {
...
+  Disabling public gateway by default (`NoFetch: true`),
+  and enabling selected gateway features only on specific hostname for which data
+  is already present on the node:
+  ```json
+  "Gateway": {
+      "NoFetch": true,
+      "NoDNSLink": true,
+      "PublicGateways": {
+        "en.wikipedia-on-ipfs.org": {
+          "NoDNSLink": false,
```
Stebalien: Does this default to true or false?

lidel: It defaults to false, so DNSLink works out of the box. 

We could simplify example and remove explicit definition, but then it would have an empty object, so I've left it to make example more clear.

docs/config.md
lidel: Yes. 
`NoDNSLink` is both global and per host.
It enables user to opt-out from resolving `Host: ` header globally, then enable it for a specific hostname.

docs/config.md
lidel: Sure. I will add a flag to disable this type of hardening if someone really needs to do it:

> Blocking cookies is a nuclear option that may break some deployments behind reverse proxies that pass all headers. We would need `NoWebSecurity` flag (`false` by default) in case someone really wants to disable things like `Clear-Site-Data` in contexts that provide no origin isolation. 


I'll go with `NoWebSecurity` (`false` by default) and make it work the same as `NoDNSLink` (unless you suggest something else).

docs/config.md
lidel: Yeah not sure what is the best wording here, would appreciate suggestions:

```json
     "PublicGateways": {
        "example.com": {
          "Paths": ["/ipfs"],
          "NoDNSLink": false,
```

Above resolves DNSLink for `example.com` and mounts it at `/`, then mounts `/ipfs/` on top of it, so content-addressed gateway is used instead of a DNSLink subpath. 

Would below be better?

```suggestion
  - **Note:** If DNSLink is enabled `Paths` are mounted on top of `/` and take priority over DNSLink.
```

docs/config.md
lidel: Yes, now that we are not reusing `PathPrefixes`, I want to make it more useful: 
- `Paths` define which content paths are handled by the subdomain gateway.
  - Example: `Paths: ["/ipfs"]` would enable `http://{immutable}.ipfs.{hostname}` but refuse `http://{mutable}.ipns.{hostname}`

docs/config.md
Stebalien: It's just a bit funny because the only place we specify this is where it's _false_. Given that, I'd assume it defaults to true.

docs/config.md
Stebalien: SGTM.

lidel: Note to self: just for sake of merging this sooner than later, we agreed to descope `Clear-Site-Data` from this PR  and land it separately :)

docs/config.md
Stebalien: That sounds reasonable. It's just that "DNSLink is mounted" is a bit confusing.

lidel:
(rebased and updated `docs/config` to match new style  from https://github.com/ipfs/go-ipfs/pull/6855)

Next: finish tests

core/corehttp/gateway_proxy.go
```diff
@@ -53,6 +53,10 @@ func ProxyOption() ServeOption {
 	return func(n *core.IpfsNode, l net.Listener, mux *http.ServeMux) (*http.ServeMux, error) {
 		childMux := http.NewServeMux()
 		mux.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+			// TODO remove below dead code?
+			// Rationale Go does not support requests with CONNECT method :-(
+			// https://golang.org/src/net/http/request.go#L111
+			// Below code block never gets executed
 			if r.Method == http.MethodConnect {
```
lidel: Unless I got it totally wrong, [request handling in `http` does not really support `CONNECT` method](https://golang.org/src/net/http/request.go#L111). 

Based on my debug the inside of this `if` in `ProxyOption` is never used, even when `CONNECT` is forced via `curl --proxytunnel -x`.

It is not a problem because regular HTTP proxy (`-x` without `--proxytunnel` in `curl`) works as expected, and we know that `CONNECT`-based proxy tunneling  won't be used for local subdomains in web browsers.
One liner for proxy testing in browser:
```console
$ chromium --incognito --proxy-server="http://127.0.0.1:8030" "http://bafybeiemxf5abjwjbikoz4mc3a3dla6ual3jsgpdr4cjr3oz3evfyavhwq.ipfs.localhost:8030"
```

To keep the surface of this PR minimal I propose we remove `ProxyOption` and limit its scope to regular HTTP proxy mode (which already works and got sharness tests). 

@Stebalien Thoughts?

Stebalien: Go's http _client_ does not support _sending_ requests with the CONNECT method. If you want to use a proxy, you have to specify a `Proxy` function on the transport (https://golang.org/pkg/net/http/#Transport).

Stebalien: I wonder if something else is hijacking the request? Or handling it first?

lidel: Indeed. All `CONNECT` requests retuned HTTP 404. They were handled by `http.Server`, but no `ServeMux`-based handler was capable of acting on it (they have similar limitation as http client). 

Not sure if this is the most elegant way, but I've created a thin wrapper that checks HTTP method before passing request to `ServeMux` handlers, and if it is `CONNECT` it just returns HTTP 200 to signal to the client that regular proxy GET can be issued:

https://github.com/ipfs/go-ipfs/commit/bcb3a9552b89cd113287697e5caab4ddd5c39b06#diff-6eb71e82338561cf3806f670b1156365R46

For what its worth, I enabled tests for `CONNECT` mode and it all works as expected.

Stebalien: Interesting. That's much simpler than the old approach. LGTM.

test/sharness/lib/test-lib.sh
```diff
@@ -438,3 +438,4 @@ findprovs_expect() {
     test_cmp findprovsOut expected
   '
 }
+
```
Stebalien: extra change?

namesys/routing.go
```diff
@@ -59,6 +59,18 @@ func (r *IpnsResolver) resolveOnceAsync(ctx context.Context, name string, option
...
 	name = strings.TrimPrefix(name, "/ipns/")
+
+	// Fix PeerID represented as CIDv1 with invalid multicodec
+	if name[:4] == "bafy" { // TODO: is this a safe way of detecting base32-dag-pb?
+		ipnsCid, err := cid.Decode(name)
+		if err == nil && ipnsCid.Type() != cid.Libp2pKey {
+			// CIDs in IPNS are expected to have libp2p-key multicodec.
+			// We ease the transition by fixing multicodec on the fly:
+			// https://github.com/ipfs/go-ipfs/issues/5287#issuecomment-492163929
+			name = cid.NewCidV1(cid.Libp2pKey, ipnsCid.Hash()).String()
```
Stebalien: Shouldn't we return a redirect?

namesys/namesys.go
```diff
@@ -112,12 +112,12 @@ func (ns *mpns) resolveOnceAsync(ctx context.Context, name string, options opts.
...
 
 	// Resolver selection:
-	// 1. if it is a multihash resolve through "ipns".
+	// 1. if it is a CID/multihash resolve through "ipns".
 	// 2. if it is a domain name, resolve through "dns"
 	// 3. otherwise resolve through the "proquint" resolver
 
 	var res resolver
-	if _, err := mh.FromB58String(key); err == nil {
+	if _, err := cid.Decode(key); err == nil {
```
Stebalien: We need to decode this as a peer ID. This will break ed25519 IPNS keys (which we should probably have a test for).

https://pkg.go.dev/github.com/libp2p/go-libp2p-core@v0.3.1/peer?tab=doc#Decode

lidel: Good catch.
I will switch to `peer` and add tests for missing PeerID "types" (ed25519, libp2p-key in cid, etc).

namesys/routing.go
lidel: We do return redirect on the gateway.

This is a fix for CLI where redirect is not possible.
User pasting `/ipns/{cidv1-dag-pb}` could produce error because `peer` returns error if multicodec in CIDv1 is other than `libp2p`.

Stebalien: Ah. I'm wondering if we should just return an error that includes the correct path? I'm concerned that if it "just works" with no visible redirect, it'll effectively become a feature.

lidel:
Remaining:
- [ ] add dnslink tests
- [ ] fix sharness failing on CI due to DNS resolver without support for *.localhost

namesys/routing.go
lidel: That is a very good point. I agree returning error with already fixed CID is a fair compromise outside of gateway.

How does this look?

```console
$ ipfs resolve /ipns/bafybeidsocs4bhczzvd7ilegoppt23cfdydexjkomn3dm245zzeqziz5ge
Error: peer ID represented as CIDv1 require libp2p-key multicodec: retry with bafzbeidsocs4bhczzvd7ilegoppt23cfdydexjkomn3dm245zzeqziz5ge
```

Stebalien: Maybe use `/ipns/bafzbe`...?

lidel: Good idea, done in ba682045a110b9bea2274af88689c594e378a090

docs/environment-variables.md
```diff
@@ -82,6 +82,20 @@ the `--migrate` flag).
...
+Useful for testing things like DNSLink without real DNS lookup.
+
+Example:
+
+```console
+$ IPFS_NS_MAP="dnslink-test1.example.com:/ipfs/bafkreicysg23kiwv34eg2d7qweipxwosdo2py4ldv42nbauguluen5v6am,dnslink-test2.example.com:/ipns/dnslink-test1.example.com" ipfs daemon
+...
+$ ipfs resolve -r /ipns/dnslink-test2.example.com
+/ipfs/bafkreicysg23kiwv34eg2d7qweipxwosdo2py4ldv42nbauguluen5v6am
+```
```
lidel: FYSA this is what I ended up adding to get deterministic DNSLink resolution for HTTP Gateway tests.
I believe this is good enough for HTTP Gateway tests, and pretty useful as it will work for IPNS as well.

Rationale: I tried a few approaches but everything felt too complex and not portable (required additional software such as dnsmasq and root access) or too brittle and time-consuming (writing a tool for hijacking syscalls responsible for DNS resolv would work as long `CGO_ENABLED=1`, but it feels like even bigger hack than this one).

ps. If this should not be available outside of test environments, we can make it active only if additional `--enable-ns-map` flag  is passed.

lidel:
Apologies it took so long, but I really wanted to have full test coverage for subdomains and HTTP proxy modes, and as I've added more tests, more bugs had to be squashed. 

@Stebalien I believe this is good enough for a review.

test/sharness/t0114-gateway-subdomains.sh
```diff
@@ -0,0 +1,546 @@
...
+test_kill_ipfs_daemon
+DNSLINK_FQDN="dnslink-subdomain-gw-test.example.org"
+export IPFS_NS_MAP="$DNSLINK_FQDN:/ipfs/$CIDv1"
+test_launch_ipfs_daemon
+
+test_hostname_gateway_response_should_contain \
+  "request for {dnslink}.ipns.example.com returns expected payload" \
+  "$DNSLINK_FQDN.ipns.example.com" \
+  "http://127.0.0.1:$GWAY_PORT" \
+  "$CID_VAL"
```
lidel: Example of DNSLink test in action

core/corehttp/hostname.go
```diff
@@ -0,0 +1,301 @@
...
+	//"::1":             pathGatewaySpec,
+	"localhost":       subdomainGatewaySpec,
+	"ipfs.io":         pathGatewaySpec,
+	"gateway.ipfs.io": pathGatewaySpec,
+	"dweb.link":       subdomainGatewaySpec,
+}
+
+// Find content identifier, protocol, and remaining hostname (host+optional port)
+// of a subdomain gateway (eg. *.ipfs.foo.bar.co.uk)
+var subdomainGatewayRegex = regexp.MustCompile("^(.+).(ipfs|ipns|ipld|p2p|api).([^/?#&]+)$")
```
Stebalien:
```suggestion
var subdomainGatewayRegex = regexp.MustCompile("^(.+)\\.(ipfs|ipns|ipld|p2p|api)\\.([^/?#&]+)$")
```

1. Let's write some tests for this regex.
2. Do we really need a regex? Or can we just split on the first two dots: `CID.PREFIX.EVERYTHINGELSE`?

core/corehttp/hostname.go
```diff
@@ -0,0 +1,301 @@
...
+		return "", false
+	}
+
+	// add prefix if query is present
+	if query != "" {
+		query = "?" + query
+	}
+
+	if ns == "api" || ns == "p2p" {
+		// API and P2P proxy use the same paths on subdomains:
```
Stebalien: I'd actually expect P2P to use `https://http.peerid.p2p.domain/$FORWARDED_PATH`.

core/corehttp/hostname.go
```diff
@@ -0,0 +1,301 @@
...
+		// api.hostname/api/.. and p2p.hostname/p2p/..
+		return fmt.Sprintf(
```
Stebalien: Are we going to get any double slashes here? We might want to use URL from `net/url` to format the URL.

core/corehttp/hostname.go
```diff
@@ -0,0 +1,301 @@
...
+			"http://%s.%s/%s/%s/%s%s",
```
Stebalien: Let's explicitly use the schema from the request. We don't yet support https, but this could turn into a bug in the future.

core/corehttp/hostname.go
```diff
@@ -0,0 +1,301 @@
...
+		// We ease the transition by fixing multicodec on the fly:
+		// https://github.com/ipfs/go-ipfs/issues/5287#issuecomment-492163929
+		if ns == "ipns" && multicodec != cid.Libp2pKey {
+			multicodec = cid.Libp2pKey
+		}
+
+		// if object turns out to be a valid CID,
+		// ensure text representation used in subdomain is CIDv1 in Base32
+		// https://github.com/ipfs/in-web-browsers/issues/89
+		rootID = cid.NewCidV1(multicodec, rootCid.Hash()).String()
```
Stebalien: We should explicitly pick the pace here instead of assuming the default is base32.

core/corehttp/hostname.go
```diff
@@ -0,0 +1,301 @@
...
+			ns,
+			hostname,
+			ns,
+			rootID,
+			rest,
+			query,
+		), true
+	}
+
+	if rootCid, err := cid.Decode(rootID); err == nil {
```
Stebalien: Let's break this into different cases for different namespaces. That way, we can explicitly fail if, e.g., the user passes `/ipfs/invalid`.

Stebalien: Unless we just want to redirect that anyways? Seems like a waste.

core/corehttp/hostname.go
Stebalien: Or just `https://peerid.p2p.domain/http/$FORWARDED_PATH`. Honestly, that's probably simpler to parse. The downside is that base for `$FORWARDED_PATH` will change (like it currently does) but that's not the end of the world.

docs/environment-variables.md
```diff
@@ -82,6 +82,20 @@ the `--migrate` flag).
 
 Default: https://ipfs.io/ipfs/$something (depends on the IPFS version)
 
+## `IPFS_NS_MAP`
+
+Prewarms namesys cache with static records for deteministic tests and debugging.
```
Stebalien: Needs update.

core/corehttp/hostname.go
lidel: Request does not have schema unless we have TLS listener in go-ipfs itself.
We assume TLS termination at reverse proxy such as Nginx, so in go-ipfs `r.URL.Scheme` is always empty. 

What I will do instead is to follow HTTP best practices for handling scheme in reverse proxy setups: 
- add support for [`X-Forwarded-Proto`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto) header
   -  use `https:` Scheme if `X-Forwarded-Proto: https` header is found in request. 
   - use `http:` Scheme otherwise

lidel: Done in https://github.com/ipfs/go-ipfs/pull/6096/commits/61d21c280799426dd60c891ad30d560a5ccae5cd

core/corehttp/hostname.go
lidel: Done in 53413cefe

core/corehttp/hostname.go
lidel: 1. Added tests in 077425eb7ea1e0da45fe9fa53b2b16e279512eef
2. Splitting is not enough because we want to support loading DNSLink websites from  localhost, and we don't know how many sub-levels it will produce
   Example: http://en.wikipedia-on-ipfs.org.ipns.localhost:8080

core/corehttp/hostname.go
lidel: There are some practical contraints we need to consider:

Wildcard TLS certificate will work only for a single subdomain level of "unknowns" (`*.p2p.domain`) so if we want people to be able to set up TLS for P2P proxy at a public gateway, then `https://peerid.p2p.domain/` is the max depth we can go.

Will look into this and comment here later.

core/corehttp/hostname.go
Stebalien: Hm. Won't this fail on `http://dist.ipfs.io.ipns.dweb.link`? We may need to some form of suffix lookup:

1. Lookup "link" in the known gateways.
2. Lookup "dweb.link" in the known gateways.

etc.

core/corehttp/hostname.go
Stebalien: WFM. We could also do `*.http.p2p.domain`.

core/corehttp/hostname.go
```diff
@@ -0,0 +1,301 @@
...
+	core "github.com/ipfs/go-ipfs/core"
+	namesys "github.com/ipfs/go-ipfs/namesys"
+
+	config "github.com/ipfs/go-ipfs-config"
+	nsopts "github.com/ipfs/interface-go-ipfs-core/options/namesys"
+	isd "github.com/jbenet/go-is-domain"
+)
+
+var pathGatewaySpec = config.GatewaySpec{
+	Paths:         []string{"/ipfs/", "/ipns/", "/api/", "/p2p/", "/version"},
```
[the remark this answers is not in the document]
lidel: Good idea, added this + short tests in a41b4b919fc4dddda91968fea185ff7b655137c7

core/corehttp/hostname.go
```diff
@@ -0,0 +1,301 @@
...
+	UseSubdomains: false,
+}
+
+var subdomainGatewaySpec = config.GatewaySpec{
+	Paths:         []string{"/ipfs/", "/ipns/", "/api/", "/p2p/"},
```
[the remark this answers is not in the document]
lidel: :thinking: . .. .  added in 73ce97e2c23493d76d486d608045a9eeb0c2a0a0

core/corehttp/hostname.go
lidel: No, iiuc regex in go is greedy by default so it will match all until `.ipns.`
Added test just to be safe tho: c7bd79c104553070cf2de2e39efc5ea76dabb501
(gonna get some sleep, will finish this PR tomorrow)

Stebalien: I guess that works. But it does mean we can't have `ipfs.ipfs.io`, or anything like that. I'd prefer to do the manual walk through the known gateways as a more reliable test.

core/corehttp/hostname.go
```diff
@@ -0,0 +1,301 @@
...
+	UseSubdomains: true,
+}
+
+var defaultKnownGateways = map[string]config.GatewaySpec{
+	// Note: local IPs act as catch-all, no need to define them here
```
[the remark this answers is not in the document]
lidel: Ah, forgot to add TODO and missed it. 
I removed it and cleaned up thigs in 3b8d91d71c1ca09b2ac96fdfe8f680dffddec1bc

core/corehttp/gateway_handler.go
```diff
@@ -143,17 +142,16 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
...
 	// For example, http://example.net would become /ipns/example.net, and
 	// the redirects and links would end up as http://example.net/ipns/example.net
-	originalUrlPath := prefix + urlPath
-	ipnsHostname := false
-	if hdr := r.Header.Get("X-Ipns-Original-Path"); len(hdr) > 0 {
-		originalUrlPath = prefix + hdr
-		ipnsHostname = true
+	requestURI, err := url.ParseRequestURI(r.RequestURI)
+	if err != nil {
+		webError(w, "failed to parse request path", err, http.StatusInternalServerError)
```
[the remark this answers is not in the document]
lidel: fixed in 6362ffe628f5301b6b026bc1a7883a1ea029eedc

core/corehttp/hostname.go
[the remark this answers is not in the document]
lidel: Ack, this is a very good point. We should make it easier for people, not harder.

If someone whitelisted `/api` by adding it to `Paths` for `example.com`, they expect it to "just work", and that means no redirect, and supporting both `example.com/api/` (interop with path gateways and existing tools) and   `$cid.ipfs.example.com/api/` to avoid CORS issues when accessing API from JS loaded from subdomain content root Origin at `$cid.ipfs.example.com`.

I made that change in aecfa66f0561d1c915c014429689d1df7aa9751a (just `/api`,  `/p2p` was fixed in 835a8f06f977c7f4feb21dfa75c2668772617e05)

core/corehttp/hostname.go
lidel: For the sake of simplicity I made `/p2p/` work the same as `/ipfs/`:

- `/p2p/$peerid/http/index.txt` is represented as `$peerid.p2p.example.com/http/index.txt`
- `/p2p/$peerid/x/custom/http/index.txt` is represented as `$peerid.p2p.example.com/x/custom/http/index.txt`
- `example.com/p2p/$peerid/http/index.txt` redirects to  `$peerid.p2p.example.com/http/index.txt` (and gets same libp2p-key fix as `/ipns/`)

See tests in 835a8f06f977c7f4feb21dfa75c2668772617e05

lidel: Something that bothers me is that `/http/` is pretty meaningless in this context.
What would make `/p2p` super useful is to **make `/http/` implicit in subdomains**.

That way content path would match URIs `/` and relative paths would just work (`/index.txt`). This would make this feature great for showing local web app to someone while behind NAT. 

To illustrate, would it be ok to make:

-  `/p2p/$peerid/http/index.txt` into `$peerid.p2p.example.com/index.txt` 
-  `/p2p/$peerid/x/custom/http/index.txt` into `$peerid.p2p.example.com/x/custom/index.txt` 

?

core/corehttp/hostname.go
lidel: I'd keep it simple, don't feel like we should care about content resolution at this stage.

All `toSubdomainURL` should do is to prepare rootID for use in redirect URL.
Additional validation should happen after redirect, when subdomain request is processed.

Stebalien: This is going to behave strangely. If we fail to parse the CID, we're going to redirect the user to `qmfoobar.ipfs.dweb.link` (lowercase).

Also, don't we need to handle IPNS keys explicitly? Specifically, inline peer IDs?

core/corehttp/hostname.go
Stebalien: > Something that bothers me is that /http/ is pretty meaningless/redundant in HTTP subdomain gateway context

I believe the main concerns were:

* A websocket socket proxy. This is basically "libp2p daemon in a gateway".
  1. Open a websocket to `http://localhost/p2p/Qm.../ws/some/protocol/name`
  2. Socket gets forwarded to a service running at `/some/protocol/name`.
* Custom protocols.

Wouldn't your proposal be ambiguous?

---

Note: this is a very low-priority decision. I'm fine with https://github.com/ipfs/go-ipfs/pull/6096#discussion_r392249872 then breaking it later.

core/corehttp/hostname.go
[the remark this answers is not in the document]
Stebalien: > supporting both example.com/api/ (interop with path gateways and existing tools) and $cid.ipfs.example.com/api/ to avoid CORS issues when accessing API from JS loaded from subdomain content root Origin at $cid.ipfs.example.com.

I would _not_ support `$cid.ipfs.example.com/api` as that will conflict with paths within the `$cid.ipfs.example.com` directory. Is CORS really an issue here?

lidel: I removed `subdomainGatewayRegex` and added test for `.ipfs.*.ipfs.` edge case (ceb111771)

We now do the lookup the way you suggested.

core/corehttp/hostname.go
lidel: Yes, let's keep this PR's surface small and leave `/p2p/` handling as described in https://github.com/ipfs/go-ipfs/pull/6096#discussion_r392249872 for now.

In the future we could improve it without ambiguouity by using different `ns` subdomain for `/p2p/$peerid/http` proxy (eg. `$peerid.p2p-http.example.com/index.txt`)

core/corehttp/hostname.go
lidel: Ugh, I always forget about ed25519 PeerIDs :man_facepalming:  

Added PeerID→CID normalization + tests in cf8a678ad7758e8a9c5c9c0e78b00caebef3385c

Stebalien: I'm still not comfortable returning `http://qmbadcid.ipfs.dweb.link/some/path`. At that point, we'll have erased the original CID so we can't even display a reasonable error to the user.

core/corehttp/hostname.go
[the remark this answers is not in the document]
lidel: Fair point. I'd rather have people setup CORS if they need to, than block them from using specific words in URLs on their websites. Removing path conflicts is more important.

I removed support for `$cid.ipfs.example.com/api` in dd8c804cd24cf844c68a7423570a34566ac384f7 and added bunch of tests to ensure `api` and `ipfs`  can be used as file or directory names in content roots.

core/corehttp/hostname.go
lidel: Hm.. erasure of the original CID is no longer a problem: https://github.com/ipfs/go-ipfs/pull/6982 / df3c593201e17e93cb8da253c28e4c4639f02922 got merged into this PR.

We now resolve every content path before returning redirect because we need to include data in response body for curl. We return HTTP 301 redirect only if original response was HTTP 200, and 200 requires successful IPFS resolve. 

To illustrate, `http://dweb.link/ipfs/QmInvalidCID` will produce HTTP Error 400 with the original (case-sensitive) content path:

```
invalid ipfs path: invalid path "/ipfs/QmInvalidCID": invalid CID: selected encoding not supported
```

I've added sharness test in 4637ab8fa81c1679f8e92e2521ee0e0014345b20 to ensure the content path is always validated before redirect.

lidel:
@Stebalien I believe I've addressed the two remaining issues (see comments above). 
I also switched DNSLink lookup to use CoreAPI in f03d973dfed0fc9f5ff95be6899a7e9080bfe93b – let me know if anything is missing.

Q: What is the best practice regarding conflicting go.mod? 
Is it ok to rebase this PR or should I merge changes from master?

Stebalien:
Merge: my plan is to squash, rebase, then merge.

core/corehttp/hostname.go
```diff
@@ -0,0 +1,371 @@
...
+	}
+	return false
+}
+
+func stripPort(hostname string) string {
+	host, _, err := net.SplitHostPort(hostname)
+	if err == nil {
+		return host
+	}
+	// noop: this should never happen
```
Stebalien: How so? There may not be a port.

Stebalien:
I've squashed this into reasonable commits.

Stebalien:
And I've updated go-ipfs-config to the release version.

core/corehttp/hostname.go
```diff
@@ -0,0 +1,376 @@
...
```
lidel: Pretty sure it is a left-over from one of refactors, can be removed

```suggestion
```

namesys/routing.go
```diff
@@ -59,14 +59,8 @@ func (r *IpnsResolver) resolveOnceAsync(ctx context.Context, name string, option
...
 	name = strings.TrimPrefix(name, "/ipns/")
+
 	pid, err := peer.Decode(name)
-	if err != nil {
-		log.Debugf("RoutingResolver: could not convert public key hash %s to peer ID: %s\n", name, err)
-		out <- onceResult{err: err}
-		close(out)
-		cancel()
-		return out
-	}
```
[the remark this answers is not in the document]
lidel: I don't remember removing this on purpose, we should restore this.

Digression: I'd do it myself but I need to fix my go dev env, go-ipfs requires go 1.14 now, and that requires kernel > 5.2, so need to upgrade a few things :upside_down_face:
commenting so we don't forget

ilyaigpetrov:
I can't wrap my mind around this currently, so I just kindly ask you not to forget the idea of @Stebalien about permanent links if it's appropriate here:

> In that light, we may want something like https://<ipns-hash>.ipns-archive.cf-ipfs.com/<signed-ipns-record>/.... Or maybe the query string? https://<ipns-hash>.ipns.cf-ipfs.com/thing/other-thing?backup=<signed-ipns-record>?

Source: https://github.com/ipld/specs/issues/19#issuecomment-511645001

If it's not appropriate, just ignore me.

Stebalien:
This change won't conflict with that feature.

ilyaigpetrov:
> This change won't conflict with that feature.

In the screenshots above I see `*.ipns.localhost:8030` so if we request `foo.ipns.localhost:8030/<signed-ipns-record>/baz/file` then there would be a conflict with files served from FS and `<signed-ipns-record>`.

Could you, please, elaborate why won't there be a conflict: will be this service launched on special port (e.g. 8030) different from permanent mutable links?

Stebalien:
We'd use one of the proposals in the comment you linked: either (a) use a different subdomain `<ipns-hash>.ipns-archive.localhost:8080` or (b) `<ipns-hash>.ipns.localhost:8080/some/path?backup=<signed-ipns-record>`. Both of those solutions work with with this new feature.

bmwiedemann:
Seems to break on Linux => #7290

PR #6098, opened 2019-03-18
build: add ipget in dist_get

djdv:
Because dogfood (and it's faster to fetch through my LAN with existing nodes than through the gateway usually).

### DIFF
--- a/bin/dist_get
+++ b/bin/dist_get
@@ -43,6 +43,7 @@ download() {
 		die "download error: cannot write to $dl_output"
 	fi
 
+	try_download "$dl_url" "$dl_output" "ipget '$dl_url' -o '$dl_output'" && return
 	try_download "$dl_url" "$dl_output" "wget '$dl_url' -O '$dl_output'" && return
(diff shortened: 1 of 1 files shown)


Stebalien:
So, I'd rather prefer the _local_ gateway, if it exists. I believe the current code does that, right?

djdv:
It appears so
The default `node` value for `ipget` is "fallback"
https://github.com/ipfs/ipget/blob/8f760b0f83fe762d593e06512d7a5450e0215889/main.go#L57
which tries to use the api and failing that will spawn one
https://github.com/noffle/fallback-ipfs-shell/blob/301ebdc20ddff6000cdc27881d411d9bc40496aa/getshell.go#L17

Stebalien:
Ah, nice!

Stebalien:
Actually... We no longer _need_ `dist_get` because we're not building with gx.

djdv:
That's true 😅 
Probably only useful in other repos that use the same set of scripts.
Since this is likely going to go with a script cleanup, closing for now.

PR #6100, opened 2019-03-19
gomod: Update Dockerfile, remove Dockerfile.fast

### LINKED ISSUE #6099: Fix docker files post gx
These (at least the fast one) still assume that we're using gx. We should revisit these assumptions with go modules in mind.

@Kubuxu mind taking this on?

Kubuxu:
The main Dockerfile supports dependencies caching now, rendering
Dockerfile.fast useless.

Resolves #6099


Kubuxu:
Fixed

PR #6111, opened 2019-03-20
circle: test docker and update golang

Stebalien:


Stebalien:
(going to take some work)

PR #6115, opened 2019-03-21
feat: add version deps command [modversion]

Kubuxu:
- [x] Tests

### DIFF
--- a/core/commands/version.go
+++ b/core/commands/version.go
@@ -118,9 +118,13 @@ Print out all dependencies and their versions.`,
 			}
 			return
 		}
-		res.Emit(toDependency(&info.Main))
+		if err := res.Emit(toDependency(&info.Main)); err != nil {
+			return err
+		}
 		for _, dep := range info.Deps {
-			res.Emit(toDependency(dep))
+			if err := res.Emit(toDependency(dep)); err != nil {
+				return err
+			}
 		}
 		return nil
 	},
@@ -131,7 +135,7 @@ Print out all dependencies and their versions.`,
 				fmt.Fprintf(w, " => %s", dep.ReplacedBy)
 			}
 			fmt.Fprintf(w, "\n")
-			return errors.New("test")
+			return nil
 		}),
 	},
 }
--- a/core/commands/version.go
+++ b/core/commands/version.go
@@ -1,14 +1,16 @@
 package commands
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"runtime"
+	"runtime/debug"
 
 	version "github.com/ipfs/go-ipfs"
 	fsrepo "github.com/ipfs/go-ipfs/repo/fsrepo"
 
-	"github.com/ipfs/go-ipfs-cmdkit"
+	cmdkit "github.com/ipfs/go-ipfs-cmdkit"
 	cmds "github.com/ipfs/go-ipfs-cmds"
 )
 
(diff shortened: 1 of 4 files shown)


core/commands/version.go
```diff
@@ -83,3 +88,50 @@ var VersionCmd = &cmds.Command{
...
+			return
+		}
+		res.Emit(toDependency(&info.Main))
+		for _, dep := range info.Deps {
+			res.Emit(toDependency(dep))
+		}
+		return nil
+	},
+	Encoders: cmds.EncoderMap{
+		cmds.Text: cmds.MakeTypedEncoder(func(req *cmds.Request, w io.Writer, dep Dependency) error {
```
Kubuxu: For some reason, if I change `Dependency` to `*Dependency` here I get no output, no error and the encoder is not being run (tested with `panic`).

Kubuxu:
@Stebalien this requires us to bump to Go1.12, is that ok?

core/commands/version.go
Stebalien: https://github.com/ipfs/go-ipfs-cmds/pull/155

core/commands/version.go
```diff
@@ -83,3 +88,50 @@ var VersionCmd = &cmds.Command{
...
+			fmt.Fprintf(w, "%s@%s", dep.Path, dep.Version)
+			if dep.ReplacedBy != "" {
+				fmt.Fprintf(w, " => %s", dep.ReplacedBy)
+			}
+			fmt.Fprintf(w, "\n")
+			return errors.New("test")
```
Stebalien: foobar

Stebalien:
@Kubuxu personally, I'm fine with a go 1.12 bump. Non-buggy go modules require that as well.

---

Other comment: needs test.

core/commands/version.go
Kubuxu: Seems like this error gets ignored. Interesting.

Stebalien:
```
diff --git a/core/commands/version.go b/core/commands/version.go
index 9c5c666d0..c9632f799 100644
--- a/core/commands/version.go
+++ b/core/commands/version.go
@@ -118,9 +118,13 @@ Print out all dependencies and their versions.`,
 			}
 			return
 		}
-		res.Emit(toDependency(&info.Main))
+		if err := res.Emit(toDependency(&info.Main)); err != nil {
+			return err
+		}
 		for _, dep := range info.Deps {
-			res.Emit(toDependency(dep))
+			if err := res.Emit(toDependency(dep)); err != nil {
+				return err
+			}
 		}
 		return nil
 	},
```

Stebalien:
@Kubuxu we'll also need to update the go version in CI.

Kubuxu:
I know, just didn't have time to handle it today.

Kubuxu:
This is ready.

test/sharness/t0010-basic-commands.sh
```diff
@@ -41,6 +41,15 @@ test_expect_success "ipfs version --all has all required fields" '
   grep "Golang version" version_all.txt
 '
 
+test_expect_success "ipfs version deps succeeds" '
+  ipfs version deps >deps.txt
+'
+
+test_expect_success "ipfs version deps output looks good" '
+  head -1 deps.txt | grep "go-ipfs (devel)" &&
+  [[ $(tail -n +2 | egrep -v -c "^[^ @]+@v[^ @]+( => [^ @]+@v[^ @]+)?$") -eq 0 ]] ||
```
Stebalien: Was there supposed to be another command here?

Stebalien: Ah.

Kubuxu: Fixing.

Kubuxu: Fixed.

Kubuxu:
Wait with it, I've done something weird in sharness.

Kubuxu:
I think it is done.

Kubuxu:
It is ready.

PR #6118, opened 2019-03-22
support add --nocopy <url>

### LINKED ISSUE #6065: Support for ipfs add <url>
#### Version information:

```yaml
go-ipfs version: 0.4.19-
Repo version: 7
System version: arm64/android
Golang version: go1.11.5
```

#### Type: 

enhancement

#### Description:

I noticed that IPFS Cluster supported `ipfs-cluster-ctl add <url>`, but that feature isn't in IPFS. I think it's a pretty amazing command and I'd rather use that than `urlstore`.

Relevant : https://github.com/ipfs/ipfs-cluster/issues/701#issuecomment-470753292

jmank88:
This PR enables `add --nocopy` support for URLs, by ~setting `DagBuilderParams.URL` when `files.FileInfo.AbsPath()` is a compatible URL.~ updating dependencies.

~Draft until `go-ipfs-cmds v0.0.4` is released~
Fixes #6065

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -32,12 +32,12 @@ require (
 	github.com/ipfs/go-ipfs-blockstore v0.0.1
 	github.com/ipfs/go-ipfs-chunker v0.0.1
 	github.com/ipfs/go-ipfs-cmdkit v0.0.1
-	github.com/ipfs/go-ipfs-cmds v0.0.2
+	github.com/ipfs/go-ipfs-cmds v0.0.4
 	github.com/ipfs/go-ipfs-config v0.0.1
 	github.com/ipfs/go-ipfs-ds-help v0.0.1
 	github.com/ipfs/go-ipfs-exchange-interface v0.0.1
 	github.com/ipfs/go-ipfs-exchange-offline v0.0.1
-	github.com/ipfs/go-ipfs-files v0.0.1
+	github.com/ipfs/go-ipfs-files v0.0.2
 	github.com/ipfs/go-ipfs-posinfo v0.0.1
 	github.com/ipfs/go-ipfs-routing v0.0.1
 	github.com/ipfs/go-ipfs-util v0.0.1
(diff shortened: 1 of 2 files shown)


core/coreunix/add.go
```diff
@@ -120,6 +121,11 @@ func (adder *Adder) add(reader io.Reader) (ipld.Node, error) {
 		NoCopy:     adder.NoCopy,
 		CidBuilder: adder.CidBuilder,
 	}
+	if fi, ok := reader.(files.FileInfo); ok {
+		if absPath := fi.AbsPath(); filestore.IsURL(absPath) {
+			params.URL = absPath
+		}
+	}
```
Stebalien: I'm not sure we need this given that we're setting AbsPath. We can probably get rid of this parameter all together and construct a `WebFile` in `core/commands/urlstore.go` instead.

jmank88: Would that cause the data to be downloaded again?

jmank88:
This approach has some side effects which urlstore does not (or at least doesn't log about) - all of the paths up to the root are added as directories:
```sh
> ipfs add --nocopy https://google.cloud.storage/v0/b/bucket/o/Qm...?alt=media&token=XXXXX
added Qm... https:/google.cloud.storage/v0/b/bucket/o/Qm...?alt=media
added Qm... https:/google.cloud.storage/v0/b/bucket/o
added Qm... https:/google.cloud.storage/v0/b/bucket
added Qm... https:/google.cloud.storage/v0/b
added Qm... https:/google.cloud.storage/v0
added Qm... https:/google.cloud.storage
added Qm... https:
> ipfs cat Qm...
hello world!
```
Also note the dropping of the `&token...` and the change from `https://` to `https:/` - I'm not sure what to make of that.

Stebalien:
The slash thing is the auto path canonicalization. Actually, I think we have to go back and fix this in go-ipfs-cmds. When we create the web file, we need to set `fpath` to something sane. Ideally the "filename" from the URL, if we can extract a reasonable one.

core/coreunix/add.go
Stebalien: In the `urlstore` command, we download the data on the server (I'm talking about two different commands, we can probably fix _this_ later).

jmank88: Ah I see - this code is completely redundant, and if we use a `WebFile` from `urlstore`, then the URL param is unused. Would `urlstore` itself be redundant also though? Is it still worth keeping around?

jmank88:
I'll put up a fix for `fpath`. Then this PR is reduced to 'update the dependencies'.

core/coreunix/add.go
Stebalien: urlstore will be redundant but we should keep it around for compatibility (for a few releases as it _is_ experimental).

Stebalien:
This PR will probably be folded into #6019 once we fold #6120 into that as well (batch of interdependent changes)

Stebalien:
(folded)

Issue #4153, opened 2017-08-19
No support of punycode domains.


ivan386:
#### Version information:
go-ipfs version: 0.4.10-
Repo version: 5
System version: amd64/windows
Golang version: go1.8.3

#### Description: No support of [punycode](https://en.wikipedia.org/wiki/Punycode) domains.

Address: http://127.0.0.1:8080/ipns/магнит.tk
Result: ipfs resolve -r /ipns/магнит.tk: Could not resolve name.

Address: http://127.0.0.1:8080/ipns/xn--80agpnh5a.tk
Result: ipfs resolve -r /ipns/xn--80agpnh5a.tk: Could not resolve name.

```
nslookup -type=txt xn--80agpnh5a.tk

Non-authoritative answer:
xn--80agpnh5a.tk        text =

        "dnslink=/ipns/ivan386.tk/magnet-converter/"
```
I try to partial [fix](https://github.com/jbenet/go-is-domain/pull/6) this [Issue](https://github.com/jbenet/go-is-domain/issues).

[package idna](https://godoc.org/golang.org/x/net/idna) may help

ivan386:
So maybe its time to accept bug fix?

PR #6119, opened 2019-03-22
deps: switch back to jbenet go-is-domain

Stebalien:
Also fixes https://github.com/ipfs/go-ipfs/issues/4153.


Stebalien:
LGTY steb? LGTM. (just a dep update)

PR #6121, opened 2019-03-22
Simplify file wrapping logic

Stebalien:
I'm not sure if this is entirely correct but basically:

1. The adder now _always_ adds files under a "root" file/directory.
2. All wrap logic is handled outside the file adder.

Thoughts? This would also require removing all wrap logic from the core interface.

Depends on https://github.com/ipfs/interface-go-ipfs-core/pull/21.


core/coreunix/add.go
```diff
@@ -480,6 +403,9 @@ func outputDagnode(out chan<- interface{}, name string, dn ipld.Node) error {
 		return err
 	}
 
+	name = strings.TrimPrefix(name, rootFileName)
+	name = strings.TrimLeft(name, "/")
```
Stebalien: Moderately icky...

core/coreunix/add.go
```diff
@@ -421,28 +357,15 @@ func (adder *Adder) addFile(path string, file files.File) error {
...
-		}
-	}
-
 	it := dir.Entries()
 	for it.Next() {
 		fpath := gopath.Join(path, it.Name())
-		err := adder.addFileNode(fpath, it.Node(), false)
+		node := it.Node()
+		err := adder.addFileNode(fpath, node)
+		node.Close()
```
Stebalien: I assume we need to close this here?

core/commands/add.go
```diff
@@ -249,12 +248,16 @@ You can now check what blocks have been created by:
...
 			}
 
-			if !dir && name != "" {
-				output.Name = name
-			} else {
+			if name != "" {
 				output.Name = path.Join(name, output.Name)
 			}
 
+			if !wrap && output.Name == "" {
```
Stebalien: I switched this to check the _wrap_ flag instead of "is this a directory". Is that correct?

core/coreunix/add.go
magik6k: Yup (though I'd check errors here)

core/coreunix/add.go
Stebalien: Actually, this whole "rootFileName" thing is kind of broken (for now at least). Your point about not using MFS for a single file is probably the right way to go.

Otherwise, we'll need to correctly strip the root filename everywhere we send out an event.

core/commands/add.go
magik6k: I think we allowed "" only for wrapping directories, so this seems correct

core/coreapi/unixfs.go
```diff
@@ -53,31 +51,27 @@ func (api *UnixfsAPI) Add(ctx context.Context, files files.Node, opts ...options
...
-	}
 
-	bserv := blockservice.New(addblockstore, exch) // hash security 001
-	dserv := dag.NewDAGService(bserv)
+		bserv := blockservice.New(addblockstore, api.exchange) // hash security 001
+		dserv = dag.NewDAGService(bserv)
+	}
 
-	fileAdder, err := coreunix.NewAdder(ctx, pinning, addblockstore, dserv)
+	fileAdder, err := coreunix.NewAdder(pinning, addblockstore, dserv)
```
Stebalien: TL;DR:

1. We can now pass a nil dagservice to "only hash". I needed to move this functionality down into the adder because I removed the SetMfsRoot function.
2. Instead of setting `adder.Pin`, we now just pass a nil pinner to disable pinning. Really, I only _needed_ 1 (see below) but I figured I'd make these consistent.

Alternatively, I could have added an `OnlyHash` option to the file adder.

Background:

1. I wanted to move away from MFS for adding single files to get away from all the directory wrapping issues. However, the old adder kind of assumed MFS everywhere. It was technically possible to just not use it but I was worried about leaving the adder in an inconsistent state.
2. The old adder wasn't safely reusable (as far as I can tell).

So I refactored it into:

1. `Adder`: A reusable, immutable "adder" to carry all the add options, dag/pin/gc services, etc.
2. `adderJob`: job for adding files/symlinks (embeds the `Adder`).
3. `dirAdderJob`: job for adding directories (with mfs support) (embeds `adderJob`).

How does that relate to the changes above? Well, `Adder` no longer stores the MFS root (because adder is immutable/reusable). That meant `SetMfsRoot` had to go.

---

Really, we probably shouldn't be using MFS for importing files _at all_. I'm not sure why we are but we can fix that later.

core/coreunix/add.go
```diff
@@ -38,69 +39,92 @@ type Link struct {
...
+func (noopDagService) Get(_ context.Context, _ cid.Cid) (ipld.Node, error) {
+	return nil, ipld.ErrNotFound
+}
+func (noopDagService) GetMany(_ context.Context, _ []cid.Cid) <-chan *ipld.NodeOption {
+	ch := make(chan *ipld.NodeOption)
+	close(ch)
+	return ch
+}
+func (noopDagService) Remove(_ context.Context, _ cid.Cid) error       { return nil }
+func (noopDagService) RemoveMany(_ context.Context, _ []cid.Cid) error { return nil }
```
Stebalien: Should probably go elsewhere but I don't feel like bubbling `go-ipld-format`.

core/coreunix/add.go
```diff
@@ -38,69 +39,92 @@ type Link struct {
...
+func newDirAdderJob(job *adderJob) *dirAdderJob {
 	rnode := unixfs.EmptyDirNode()
-	rnode.SetCidBuilder(adder.CidBuilder)
-	mr, err := mfs.NewRoot(adder.ctx, adder.dagService, rnode, nil)
-	if err != nil {
-		return nil, err
+	rnode.SetCidBuilder(job.CidBuilder)
+	ds := job.dagService
+	if ds == nil {
+		ds = mdutils.Mock()
```
Stebalien: This mock dag should also go in go-ipld-format.

Stebalien:
There's _still_ something wrong with this...

Stebalien:
(not worth debugging for now)

PR #6122, opened 2019-03-25
coreapi: Drop error from ParsePath

magik6k:
See https://github.com/ipfs/interface-go-ipfs-core/pull/22 / https://github.com/ipfs/interface-go-ipfs-core/pull/20


Stebalien:
(rebased)

PR #6125, opened 2019-03-25
Query for provider head/tail

michaelavila:
Query for provider head/tail per Steven's suggestion here: https://github.com/ipfs/go-ipfs/pull/6068#discussion_r266291056

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -20,10 +20,10 @@ require (
 	github.com/ipfs/go-blockservice v0.0.3
 	github.com/ipfs/go-cid v0.0.1
 	github.com/ipfs/go-cidutil v0.0.1
-	github.com/ipfs/go-datastore v0.0.1
+	github.com/ipfs/go-datastore v0.0.2
 	github.com/ipfs/go-detect-race v0.0.1
 	github.com/ipfs/go-ds-badger v0.0.2
-	github.com/ipfs/go-ds-flatfs v0.0.1
+	github.com/ipfs/go-ds-flatfs v0.0.2
 	github.com/ipfs/go-ds-leveldb v0.0.1
 	github.com/ipfs/go-ds-measure v0.0.1
 	github.com/ipfs/go-fs-lock v0.0.1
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -114,3 +114,8 @@ require (
 	gopkg.in/gemnasium/logrus-airbrake-hook.v2 v2.1.2 // indirect
 	gotest.tools/gotestsum v0.3.3
 )
+
+replace (
+	github.com/ipfs/go-datastore => github.com/ipfs/go-datastore v0.0.2-0.20190323043649-bce485ce18d1
+	github.com/ipfs/go-ds-flatfs => github.com/ipfs/go-ds-flatfs v0.0.2-0.20190316035933-d5e3c1fa14d2
```
michaelavila: We obviously don't want to merge with these, they are only here for CI purposes until we put out a new version of datastore.

provider/queue.go
```diff
@@ -142,40 +142,52 @@ func (q *Queue) work() {
...
 
-	var tail uint64
-	var head uint64 = math.MaxUint64
-	for entry := range results.Next() {
-		trimmed := strings.TrimPrefix(entry.Key, "/")
-		id, err := strconv.ParseUint(trimmed, 10, 64)
-		if err != nil {
-			return 0, 0, err
-		}
+func getQueueHead(ds datastore.Datastore) (uint64, error) {
```
michaelavila: I should document these new methods. I am doing that now.

michaelavila: Actually, these methods are internal and relatively small and so I won't add any comments.

go.mod
Stebalien: Released.

Stebalien:
FYI, a "release" is now just a `git tag -as vX.Y.Z`.

Kubuxu:
I would recommend running `go mod tidy` before releases.

Stebalien:
(we should add that to CI)

PR #6135, opened 2019-03-27
Fuse fixes

anacrolix:
- Skip errors due to fuse missing on OSX
- Downgrade bootstrap node error: It's very noisy in the test logs, and probably indicates user error (which may be a warning), rather than an actual, unexpected error.
- Fix automatic installation of go-fuse-version with go modules
- Misc uses of testing.T.Helper


anacrolix:
I thought the GitCop stuff was removed?

Stebalien:
> I thought the GitCop stuff was removed?

We were waiting on final instructions from legal.

anacrolix:
What's the plan now? Should I rebase with the sign-offs, or can you just merge as is?

PR #6138, opened 2019-03-27
Remove duplicate blockstore from the package list

MattSkala:
The go-ipfs-blockstore package is mentioned twice in the Blocks section and once in the Commands section in the packages table. I think it makes sense to only include it once.


Stebalien:
This was probably some kind of merge error. Could you also remove the duplicate lines from package-list.json?

MattSkala:
Sure, done

PR #6162, opened 2019-04-01
DI-based core.NewNode

magik6k:
TODO:
- [x] Fix Regressions
  - [ ] Fuse (probably works)
- [x] Remove old code
- [x] Organize DI Providers / Services
- [ ] (Possibly separate PR) Expose DI /


magik6k:
Should be ready for a review. Few notes:

* I recommend starting with reading the DI library docs, mainly this part - https://godoc.org/go.uber.org/fx#ex-package.
* This PR aims to not introduce too many breaking changes, and make the code less intimidating.
* The main constructor flow starts in `core/builder.NewNode`.
* `core/node` is a new package which contains DI-related node construction logic.
* `core/` package now only provides `IpfsNode` and wraps DI constructor (will change in constructor interface refactor, not sure how exactly yet).
* `core/node/groups.go` groups providers into larger components for a few reasons:
  * It's easier to see what is related
  * Groups are exported, so if someone want's to assemble their own node with different components they don't have to pull everything manually
* Diff might seem big, but most changes are just moving existing constructor code and wrapping it into smaller units.

core/node/core.go
```diff
@@ -0,0 +1,117 @@
...
+)
+
+func BlockServiceCtor(lc fx.Lifecycle, bs blockstore.Blockstore, rem exchange.Interface) blockservice.BlockService {
+	bsvc := blockservice.New(bs, rem)
+
+	lc.Append(fx.Hook{
+		OnStop: func(ctx context.Context) error {
+			return bsvc.Close()
+		},
+	})
```
Stebalien: So, this is going to be a bit error prone. If some start hook fails, we'll never shut this service down. I think the only _clean_ way to do this is to add a `Start` function.

We may want a helper for this:

```go
func CloseOnStop(lc fx.Lifecycle, c io.Closer) {
  lc.Apend(fx.Hook{
    OnStop: func(ctx context.Context) error {
      return c.Close()
    }
  })
}
```

core/node/groups.go
```diff
@@ -0,0 +1,131 @@
...
+	fx.Provide(RecordValidator),
+)
+
+var Providers = fx.Options(
+	fx.Provide(ProviderQueue),
+	fx.Provide(ProviderCtor),
+	fx.Provide(ReproviderCtor),
+
+	fx.Invoke(Reprovider),
+	fx.Invoke(provider.Provider.Run),
```
Stebalien: Shouldn't this be an `OnStart` function registered by the `ProviderCtor`?

core/node/helpers.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+type lcProcess struct {
+	fx.In
+
+	LC   fx.Lifecycle
+	Proc goprocess.Process
+}
+
+func (lp *lcProcess) Run(f goprocess.ProcessFunc) {
```
Stebalien: nit: Maybe just call this `Append`? Not much better but `Run` sounds like something is happening now.

core/node/libp2p.go
```diff
@@ -0,0 +1,591 @@
...
+	swarmkey, err := repo.SwarmKey()
+	if err != nil || swarmkey == nil {
+		return err
+	}
+
+	done := make(chan struct{})
+	lc.Append(fx.Hook{
+		OnStart: func(_ context.Context) error {
+			go func() {
+				t := time.NewTicker(30 * time.Second)
```
Stebalien: We should have a defered `t.Stop()`

core/node/libp2p.go
```diff
@@ -0,0 +1,591 @@
...
+		opts = append(opts, o...)
+	}
+
+	ctx, cancel := context.WithCancel(mctx)
+	lc.Append(fx.Hook{
+		OnStop: func(_ context.Context) error {
+			cancel()
+			return nil
+		},
+	})
```
Stebalien: `lifecycleCtx`?

core/node/libp2p.go
```diff
@@ -0,0 +1,591 @@
...
+	// Ideally, we'd do something like:
+	//
+	// 1. Add some fancy method to introspect into tiered routers to extract
+	//    things like the pubsub router or the DHT (complicated, messy,
+	//    probably not worth it).
+	// 2. Pass the IpfsNode into the RoutingOption (would also remove the
+	//    PSRouter case below.
+	// 3. Introduce some kind of service manager? (my personal favorite but
+	//    that requires a fair amount of work).
+	if dht, ok := out.Routing.(*dht.IpfsDHT); ok {
```
Stebalien: Can we have different DI units provide different routers and then have a single unit collect them all? (then we can just drop the routing option system)

core/node/libp2p.go
```diff
@@ -0,0 +1,591 @@
...
+		out.IpfsRouting = in.BaseRouting
+
+		if ipnsps {
+			out.PSRouter = namesys.NewPubsubValueStore(
+				lifecycleCtx(mctx, lc),
+				in.Host,
+				in.BaseRouting,
+				in.PubSub,
+				in.Validator,
+			)
```
Stebalien: So yeah, we could put this in it's own unit then put the code below in it's own unit. That unit would optionally take every kind of router we can construct and, based on the ones we have, construct the composed router.

core/node/core.go
magik6k: Looking at https://github.com/uber-go/fx/blob/master/app.go#L590, it seems like `OnStop` hooks are called if `OnStart` hooks fail (only the ones that have started - https://github.com/uber-go/fx/blob/master/internal/lifecycle/lifecycle.go#L60-L92)

core/node/libp2p.go
magik6k: This seems to run into a circular dependency - If we want to construct routing after the host, we'd also need to somehow defer enabling autorelay.

Autorelay construction depends on routing [[1]](https://github.com/libp2p/go-libp2p/blob/master/config/config.go#L187-L205), which in that place can only be provided as `libp2p.Option` to `libp2p.New`. We can't get the full routing there as `PubsubRouter` requires `PubSub`, which requires `Host`..

We could try not passing the `AutoRelay` option to Libp2p constructor, and instead construct it manually here, but it feels hacky. Better ideas?

magik6k:
> add some layer of abstraction so we aren't tied to `fx`

There isn't really that much `fx` specific code here (DI systems likely won't differ that much), so if we'd really need to switch to something else, it shouldn't be too hard.

core/node/core.go
Stebalien: My worry is that we're actually _starting_ this service before `OnStart`. That is, we don't have separate construct/start steps.

Stebalien:
> There isn't really that much fx specific code here (DI systems likely won't differ that much), so if we'd really need to switch to something else, it shouldn't be too hard.

Really, my main concern is pluggability but we can probably punt on that.

core/node/libp2p.go
```diff
@@ -0,0 +1,587 @@
...
+	var raddrs []ma.Multiaddr
+	for _, addr := range addrs {
+		_, err := addr.ValueForProtocol(circuit.P_CIRCUIT)
+		if err == nil {
+			continue
+		}
+		raddrs = append(raddrs, addr)
+	}
+	return raddrs
+}
```
Stebalien: This should no longer be necessary as of the latest master. Instead of introducing merge conflicts, can we just remove this here?

core/node/libp2p.go
Stebalien: That sucks... we'll fix this when we add fx support to libp2p but I'd like to merge this first.

What if we had _this_ option depend on `IpfsDHT` specifically and then constructed the real router after we construct the host?

core/node/core.go
magik6k: As far as I can tell, there isn't really anything actively running when we 'start' the block service. `blockservice.Close` only calls `Close` on exchange, which we also do in `OnlineExchangeCtor` few lines below.

We could wrap blockservice into an additional layer which returns an error on all calls until some `start()` method is called, which would expose any places using blockservice before it should be used.

Stebalien: Meh. It's probably fine. I'm just wondering if we should move to a model where we don't actually _start_ systems till we call some `Start()` method. But we can handle that later.

core/node/libp2p.go
magik6k: > What if we had this option depend on IpfsDHT specifically and then constructed the real router after we construct the host?

What about NilRouter? (used when `--routing=none` is set)

I'd say we'll clean this up in the second stage of the refactor (and remove RoutingOption)

core/node/libp2p.go
magik6k: done

core/node/libp2p.go
```diff
@@ -0,0 +1,579 @@
...
+		opts.Opts = append(opts.Opts, libp2p.BandwidthReporter(reporter))
+	}
+	return opts, reporter
+}
+
+type Libp2pOpts struct {
+	fx.Out
+
+	Opts []libp2p.Option `group:"libp2p"`
+}
```
Stebalien: We should be able to get rid of this helper entirely (https://github.com/libp2p/go-libp2p/pull/603). I'll file a followup PR after merging this.

core/bootstrap/bootstrap.go
```diff
@@ -9,27 +9,28 @@ import (
...
+	logging "github.com/ipfs/go-log"
+	"github.com/jbenet/goprocess"
+	"github.com/jbenet/goprocess/context"
+	"github.com/jbenet/goprocess/periodic"
+	"github.com/libp2p/go-libp2p-host"
+	"github.com/libp2p/go-libp2p-loggables"
+	"github.com/libp2p/go-libp2p-net"
+	"github.com/libp2p/go-libp2p-peer"
+	"github.com/libp2p/go-libp2p-peerstore"
+	"github.com/libp2p/go-libp2p-routing"
```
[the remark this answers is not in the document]
Stebalien: (meh, actually, I don't care much either way)

core/node/libp2p.go
```diff
@@ -0,0 +1,579 @@
...
+		ps.AddPrivKey(id, sk)
+		ps.AddPubKey(id, sk.GetPublic())
+	}
+
+	return ps
+}
+
+func P2PAddrFilters(cfg *config.Config) (opts Libp2pOpts, err error) {
+	for _, s := range cfg.Swarm.AddrFilters {
+		f, err := mask.NewMask(s)
```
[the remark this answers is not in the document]
Stebalien: (fixed)

Stebalien:
:rocket: :full_moon:

Stebalien:
@magik6k,

Next up: _all_ of these options need to be documented.

PR #6171, opened 2019-04-03
feat(coreapi): tag all explicit connect requests in the connection manager

### LINKED ISSUE #6164: Tagging of connections with ipfs swarm connect
#### Type:
- "enhancement ": Enhance the command "ipfs swarm connect" with additional parameter, so that a successful connection is tagged for the Connection Manager (and prevent closing
connections added by the user). 

#### Description:
Background: The Connection Manager might unintentional close a connection which was
added (ipfs swarm connect) by the user. [Note: I am not sure, if this is not done already, but I had
the feeling connections are lost, in particular in the moment when downloading content]

Stebalien:
This doesn't _stop_ us from cutting these connections but indicates that there's some hidden usefulness to them.

fixes #6164

### DIFF
--- a/core/coreapi/swarm.go
+++ b/core/coreapi/swarm.go
@@ -27,6 +27,10 @@ type connInfo struct {
 	peer peer.ID
 }
 
+// tag used in the connection manager when explicitly connecting to a peer.
+const connectionManagerTag = "user-connect"
+const connectionManagerWeight = 100
+
 func (api *SwarmAPI) Connect(ctx context.Context, pi pstore.PeerInfo) error {
 	if api.peerHost == nil {
 		return coreiface.ErrOffline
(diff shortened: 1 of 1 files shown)


core/coreapi/swarm.go
```diff
@@ -36,7 +40,12 @@ func (api *SwarmAPI) Connect(ctx context.Context, pi pstore.PeerInfo) error {
 		swrm.Backoff().Clear(pi.ID)
 	}
 
-	return api.peerHost.Connect(ctx, pi)
+	if err := api.peerHost.Connect(ctx, pi); err == nil {
```
inetic: Shouldn't this have been `err != nil`?

Stebalien: :man_facepalming:

PR #6174, opened 2019-04-03
fix error check in swarm connect

Stebalien:
🤦‍♂️

fixes https://github.com/ipfs/go-ipfs/pull/6171/files/d21bb1eed85401412ad0e8749f60de75ee27051a#discussion_r271949856


Stebalien:
The real question is why the tests didn't fail before. The test failure in https://github.com/ipfs/go-ipfs/pull/6175 is clearly related to this...

Stebalien:
(merging as a trivial but critical fix)

PR #6184, opened 2019-04-04
circle: use resource_class: large

Kubuxu:


Kubuxu:
It seems we can't use it. I will try to get it enabled for us.

Issue #6101, opened 2019-03-19
Docker tests in CI


Stebalien:
Since the transition to CircleCI, we've lost our docker tests. We should enable them in CircleCI (not sure if we can just flip a switch or something...).

Stebalien:
@lanzafame (or @mburns?) would one of you have time to get docker image testing working in CircleCI? We can either use our docker sharness tests or do something custom.

Stebalien:
We'll need to take a look at https://circleci.com/docs/2.0/building-docker-images/ and adapt the tests for this.

1. We can't mount in CircleCI.
2. Only docker containers can talk to each other.

Stebalien:
We'll need to modify:

* test/sharness/t0301-docker-migrate.sh
  * We can't mount. We'll need to create a dummy container as described in https://circleci.com/docs/2.0/building-docker-images/.
  * The docker container can't talk to network services running inside the test. Specifically, it can't talk to socat. We'll want to run a special socat container (alpine/socat?) to serve the fake HTTP response.
* .circleci/config.yml
  * We'll need to modify the sharness job's steps to invoke `setup_remote_docker`.

PR #6188, opened 2019-04-05
Dockerfile: Remove 2 year old deprecation warning

kpcyrd:
This finally removes support for the implicit daemon argument and unlocks full access to all subcommands.

### DIFF
--- a/bin/container_daemon
+++ b/bin/container_daemon
@@ -26,22 +26,4 @@ else
   ipfs config Addresses.Gateway /ip4/0.0.0.0/tcp/8080
 fi
 
-# if the first argument is daemon
-if [ "$1" = "daemon" ]; then
-  # filter the first argument until
-  # https://github.com/ipfs/go-ipfs/pull/3573
-  # has been resolved
-  shift
-else
-  # print deprecation warning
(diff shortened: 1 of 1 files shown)


Stebalien:
Could you make sure the docker sharness tests pass. Unfortunately, they're disabled on CI at the moment (#6101).

I _think_ we may need to change calls of `docker_run "$IMAGE_ID"` to `docker_run "$IMAGE_ID" daemon`.

kpcyrd:
@Stebalien it still works since the default arguments are `["daemon", "--migrate=true"]`, running the container without arguments executes `ipfs daemon --migrate=true` before and after the patch.

What changes is:

| Command | Before | After |
| --- | --- | --- |
| `docker run ipfs/go-ipfs` | `ipfs daemon --migrate=true` | `ipfs daemon --migrate=true` |
| `docker run ipfs/go-ipfs daemon` | `ipfs daemon` | `ipfs daemon` |
| `docker run ipfs/go-ipfs --migrate=true` | `ipfs daemon --migrate=true` | `ipfs --migrate=true` |
| `docker run ipfs/go-ipfs add -w /foo` | `ipfs daemon add -w /foo` | `ipfs add -w /foo` |

PR #6199, opened 2019-04-09
dep: update libp2p, circuit, leveldb, kbucket

Stebalien:
* Update go-libp2p for some identify optimizations.
* Update go-libp2p-circuit for some allocation optimizations.
* Update go-ds-leveldb for some query performance improvements.
* Update go-libp2p-kbucket to fix a really stupid bug.


Stebalien:
(just a dep update)

PR #6209, opened 2019-04-11
gitattributes: avoid normalizing known binary files

Stebalien:


Stebalien:
(this was corrupting one of our PNG files)

PR #6211, opened 2019-04-12
release: 0.4.20-rc2

Stebalien:
This time, the build scripts are ready so we can actually fire off a build.


Stebalien:
Build scripts patch: https://github.com/ipfs/distributions/pull/230

PR #6217, opened 2019-04-15
Add Stackdriver export

anacrolix:
I'm using this to export traces to Stackdriver. I'd like some feedback. Perhaps the exporter Location and default sample rate can be taken from the environment, and the export only attempted if particular environment variables are set.


magik6k:
Are there any discussions / issues linked to this?

anacrolix:
@magik6k What sort are you referring to? There's efforts in filecoin to do tracing, and libp2p to tracing and metrics via OpenCensus.

raulk:
@anacrolix – I believe @magik6k is referring to the attempt to introduce Stackdriver specifically. It's not clear to me the kind of feedback you are seeking.

anacrolix:
I'd like to know if OC exporters could be supported, via environment flags, similar to as is done in this PR for Stackdriver.

anacrolix:
I'll update this PR to be more general, to reflect how I've done this in other projects.

magik6k:
Using env vars feels better.

What is the difference in binary size with this patch applied? Can this be made into a plugin?

anacrolix:
@magik6k 
```
master              37757316
stackdriver-tracing 43943460
```

anacrolix:
As a plugin, I wonder if having upstream projects in libp2p directly reference opencensus tracing will defeat the purpose (the dependencies will be pulled in anyway). In particular, while there is no OC tracing in libp2p yet, I think there will be soon, and the cost will already be paid to add that there. It may be possible to avoid that cost, by tapping into OC's `runtime/trace` support, but then we lose a lot of extra features there, and there'll need to be a lot of fiddly OC injection outside libp2p to regain the extra OC tracing features by consumers. I.e. everything that wants OC tracing enabled, will need to call out to the plugin before calling into anything related to libp2p.

magik6k:
6MB is a lot, especially given that it isn't going to be used by 99.99% of users. Can we hide this behind a build tag, say `tracing`?

Stebalien:
It looks like a almost all of the weight comes from the stackdriver exporter, not opencensus. Putting the stackdriver code in a plugin should help quite a bit.

anacrolix:
Thanks @Stebalien that seems like the best solution.

Stebalien:
(we'll reopen this when we pick this work back up)

PR #6223, opened 2019-04-16
release 0.4.20

Stebalien:
This has been running on the gateways for several days. There are still a few issues unresolved but it looks good enough to ship.

closes #6223

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,8 +18,8 @@ As of this release, it's now much easier to run arbitrary IPFS commands within
 the docker container:
 
 ```bash
-> docker run ipfs/go-ipfs config profile apply server # apply the server profile
-> docker run ipfs/go-ipfs # start the daemon
+> docker run --name my-ipfs ipfs/go-ipfs:v0.4.20 config profile apply server # apply the server profile
+> docker start my-ipfs # start the daemon
 ```
 
 This release also [reverts](https://github.com/ipfs/go-ipfs/pull/6040) a change that
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -161,7 +161,7 @@ running a public gateway, upgrading to 0.4.20 will likely fix them.
 #### Relays
 
 After AutoRelay was introduced in go-ipfs 0.4.19, the number of peers connecting
-through relays ski-rocketed to over 120K concurrent peers. This highlighted some
+through relays skyrocketed to over 120K concurrent peers. This highlighted some
 performance issues that we've now fixed in this release. Specifically:
 
 * We've significantly reduced the amount of memory allocated per-peer.
@@ -172,10 +172,10 @@ performance issues that we've now fixed in this release. Specifically:
   rapidly forming new connections. This will actually help _all_ nodes but will
   especially help relays.
   
-If you've enabled relays in go-ipfs 0.4.19 and it hasn't burned down your
-machine yet, this release should improve things significantly. However, relays
-are still under heavy load so running an open relay will continue to be resource
-intensive.
+If you've enabled relay _hop_ (`Swarm.EnableRelayHop`) in go-ipfs 0.4.19 and it
+hasn't burned down your machine yet, this release should improve things
+significantly. However, relays are still under heavy load so running an open
+relay will continue to be resource intensive.
 
 We're continuing to investigate this issue and have a few more patches on the
 way that, unfortunately, won't make it into this release.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,8 +64,8 @@ it doesn't think it's being used.
 
 It can now resolve `/dnsaddr` addresses that _don't_ end in a peer ID. For
 example, you can now run `ipfs swarm connect /dnsaddr/bootstrap.libp2p.io` to
-connect to one of the bootstrap peers at random. NOTE: This is could connect you
-to an _arbitrary_ peer as DNS is not secure (by default). Please do not rely on
+connect to one of the bootstrap peers at random. NOTE: This could connect you to
+an _arbitrary_ peer as DNS is not secure (by default). Please do not rely on
 this except for testing or unless you know what you're doing.
 
 Finally, `ipfs swarm connect` now returns _all_ errors on failure. This should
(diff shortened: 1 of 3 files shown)


Stebalien:
cc @ipfs/go-team

CHANGELOG.md
```diff
@@ -1,8 +1,487 @@
...
+
+The `ipfs swarm connect` command has a few new features:
+
+It now marks the newly created connection as "important". This should ensure
+that the connection manager won't come along later and close the connection if
+it doesn't think it's being used.
+
+It can now resolve `/dnsaddr` addresses that _don't_ end in a peer ID. For
+example, you can now run `ipfs swarm connect /dnsaddr/bootstrap.libp2p.io` to
+connect to one of the bootstrap peers at random. NOTE: This is could connect you
```
alanshaw:
```suggestion
connect to one of the bootstrap peers at random. NOTE: This could connect you
```

CHANGELOG.md
```diff
@@ -1,8 +1,487 @@
...
+to an _arbitrary_ peer as DNS is not secure (by default). Please do not rely on
+this except for testing or unless you know what you're doing.
+
+Finally, `ipfs swarm connect` now returns _all_ errors on failure. This should
```
alanshaw: Is there a change to the HTTP API that I'll need to account for?

CHANGELOG.md
```diff
@@ -1,8 +1,487 @@
...
+
+1. Resolve _recursively_ (up to 32 steps) by default to better match user
+   expectations (these commands used to be non-recursive by default). To turn
+   recursion off, pass `-r false`.
+2. When resolving non-recursively, these commands no longer fail when partially
+   resolving a name. Instead, they simply return the intermediate result.
+
+#### Changed: `ipfs files flush`
+
+The `ipfs files flush` command now returns the CID of the flushed file.
```
alanshaw: cc @achingbrain

CHANGELOG.md
```diff
@@ -1,8 +1,487 @@
...
+2. Replace all uses of the `ipfs urlstore` command with a call to `ipfs add
+   --nocopy`. The `ipfs urlstore` command will be deprecated in a future
+   release.
+
+
+#### Changed: `ipfs swarm connect`
+
+The `ipfs swarm connect` command has a few new features:
+
+It now marks the newly created connection as "important". This should ensure
```
alanshaw: cc @jacobheun @vasco-santos

jacobheun: I'll be looking at adding tagging support in js this quarter for connection management, so this is something we should be able to look at adding in q3.

CHANGELOG.md
```diff
@@ -1,8 +1,487 @@
...
+level. You can read [#5850](https://github.com/ipfs/go-ipfs/issues/5850) for
+some discussion on this matter.
+
+### Docker
+
+As of this release, it's now much easier to run arbitrary IPFS commands within
+the docker container:
+
+```bash
+> docker run ipfs/go-ipfs config profile apply server # apply the server profile
```
lidel: I think some users not familiar with Docker may just  copy&paste commands and get surprised that container started by the second command does not include changes applied by the first one.

A safer example would be to use the same container in both, e.g:

```console
$ docker run --name my-container ipfs/go-ipfs:v0.4.20 config profile apply server # apply the server profile
$ docker start my-container # start the daemon
```

CHANGELOG.md
```diff
@@ -1,8 +1,487 @@
...
+#### Bitswap
+
+We fixed some critical CPU utilization regressions in bitswap for this release.
+If you've been noticing CPU _regressions_ in go-ipfs 0.4.19, especially when
+running a public gateway, upgrading to 0.4.20 will likely fix them.
+
+#### Relays
+
+After AutoRelay was introduced in go-ipfs 0.4.19, the number of peers connecting
+through relays ski-rocketed to over 120K concurrent peers. This highlighted some
```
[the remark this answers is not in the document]
lidel: emoji opportunity: :ski:-rocketed

CHANGELOG.md
Stebalien: No, this is just a behavior tweak. Ideally, we'll improve the connection manager later so it won't be necessary.

CHANGELOG.md
Stebalien: As I noted above, this _shouldn't_ really be necessary. However, we're still killing useful connections so this change makes the issue slightly less annoying.

CHANGELOG.md
Stebalien: I _believe_ we did this to improve the CoreAPI. IIRC, my end goal was to avoid exposing any CIDs to the user until they explicitly call `Flush` so we can make the actual IPFS import part lazy.

CHANGELOG.md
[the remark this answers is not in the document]
Stebalien: :man_facepalming:... too much skiing.

CHANGELOG.md
Stebalien: As someone not familiar with docker, thanks...

Stebalien:
Good enough for me. We can always edit the release notes after the fact and I want this out the door.

CHANGELOG.md
```diff
@@ -1,8 +1,487 @@
...
+`QmXpXzUhcS9edmFBuVafV5wFXKjfXkCQcjAUZsTs7qFf3G` (the `foo` directory) as soon
+as we finish adding the directory to go-ipfs _without_ waiting to finish
+announcing `foo/{0..1000}`. This is especially important in this release
+because we've drastically reduced the number of provide workers.
+
+The second benefit is that this queue is persistent. That means go-ipfs won't
+forget to send out this record, even if it was offline when the content was
+initially added. NOTE: go-ipfs _does_ continuously _re_-send provider records in
+the background twice a day, it just might be a while before it gets around to
+sending one out any specific one.
```
michaelavila: typo (very minor): I think this should be something like "...before it gets around to sending any specific one"

michaelavila:
LGTM wrt Content Routing

CHANGELOG.md
Stebalien: (could you file a PR to fix this)

Issue #4178, opened 2017-08-29
Don't decode URI encode slashes


Stebalien:
#### Version information:

> go-ipfs version: 0.4.11-dev-48476b292
Repo version: 5
System version: amd64/linux
Golang version: go1.8.3

#### Type:

Bug

#### Severity:

Low

#### Description:

We're currently decoding `%2F` to `/` in our paths *before* splitting path segments in our gateway. This means that we're accepting, e.g.:

```
http://127.0.0.1:8080/ipfs%2FQmS4ustL54uo8FzR9455qaxZwuMiUhyvMcX9Ba8nUH4uVv%2Freadme
```

This can cause surprising behavior and allow buggy code that shouldn't work to work. Also, funky URI escape decoding has been known to cause security problems in the past (although I think we're safe in this particular case).

lgierth:
Good catch

reinerRubin:
Do the handlers even need processed paths? I have looked through 
>"core/corehttp/gateway_handler.go"
and have not found the cases when "s/URL.Path/URL.EscapedPath()/" can break anything.

Stebalien:
Thinking about this a bit, I'm not sure if we _want_ to fix this. Yeah, it's a bit funky but we allow url-escaping everything _else_ in these paths. Thoughts?

reinerRubin:
I have also thought about this. But everything else is not interpreted as a path segment. On the other hand the fix seems to be awkward and if there are not any real benefits or security risks it is not worth it.

Stebalien:
I'd say it's not worth it. I'll leave this open for a few days to see if anyone wants to chime in.

the-superpirate:
Folks, I'm stuck with it. Is there any way to workaroud it?
I have a filename in directory named "blabla%2Fblabla" and it cannot be opened through gateway because %2F is decoded to slash beforehand.

Jorropo:
@the-superpirate `blabla%252Fblabla` ?

PR #6228, opened 2019-04-17
core/corehttp/gateway_handler: don't decode URI encoded slashes

reinerRubin:
I tried to solve the 4178 issue (Don't decode URI encode slashes). I am not 100% sure that that split/decode/join by "%2F" is a good idea. But it seems that "%2F" is not contained in any sequence, so it should be safe. But, idk,it seems quite dangerous. Can a path segment between "%2F" be invalid? Maybe do you know any better approaches?

https://github.com/ipfs/go-ipfs/issues/4178


Stebalien:
Under discussion in the parent bug. I'm having second thoughts on my own issue...

PR #6230, opened 2019-04-18
coreapi: remove libp2p DI helper struct

Stebalien:


Stebalien:
(hm, grouping doesn't work like that)

Issue #4607, opened 2018-01-25
The gateway handlers use the wrong context.


Stebalien:
In `core/corehttp/gateway_handler.go`.

They should be using the context from the request, they're using the context from the node.

magik6k:
Fixed in https://github.com/ipfs/go-ipfs/pull/6244

PR #6244, opened 2019-04-22
core/corehttp/gateway_handler: pass a request ctx instead of the node

reinerRubin:
Hello,
I have tried to solve this bug [1]. 

idk, about "tctx" in the "deleteHandler". After tctx is defined some calls after still use original ctx. I have assumed that it is bug and changed to "tctx"     

1 - https://github.com/ipfs/go-ipfs/issues/4607

### DIFF
--- a/core/corehttp/gateway_handler.go
+++ b/core/corehttp/gateway_handler.go
@@ -19,7 +19,7 @@ import (
 	"github.com/dustin/go-humanize"
 	"github.com/ipfs/go-cid"
 	chunker "github.com/ipfs/go-ipfs-chunker"
-	"github.com/ipfs/go-ipfs-files"
+	files "github.com/ipfs/go-ipfs-files"
 	ipld "github.com/ipfs/go-ipld-format"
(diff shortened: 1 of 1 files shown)


core/corehttp/gateway_handler.go
```diff
@@ -542,7 +537,7 @@ func (i *gatewayHandler) deleteHandler(w http.ResponseWriter, r *http.Request) {
 
 	var newnode *dag.ProtoNode = pbnd
 	for j := len(pathNodes) - 2; j >= 0; j-- {
-		if err := i.node.DAG.Add(ctx, newnode); err != nil {
+		if err := i.node.DAG.Add(tctx, newnode); err != nil {
```
Stebalien: Actually, this wasn't a bug. The minute timeout is for _fetching_ the node as that may involve a network operation. Here, we don't really want to time out at all (although the global hour timeout is fine).

reinerRubin: Ok, got it. I'll make usage of "tctx" more explicit. Just put it into a new method or a "block".

PR #6245, opened 2019-04-23
Export Prometheus via OpenCensus and register the DHT metrics

anacrolix:
This uses OC to export metrics via Prometheus. This also means that other OC metrics, such as the ones in go-libp2p-kad-dht can be exposed (via Prometheus, and whatever other exporters are used). I did some checks that no existing metrics move as a result.

See also https://github.com/libp2p/go-libp2p-kad-dht/pull/327.


anacrolix:
@magik6k @raulk

Issue #6254, opened 2019-04-25
ipfs add with two files only adds one of them in 0.4.20


olizilla:
#### Version information:
```console
$ ipfs version --all
go-ipfs version: 0.4.20-
Repo version: 7
System version: amd64/darwin
Golang version: go1.12.4
```
#### Description:

Adding 2 files via the CLI  returns the cid for only 1 of the files:

```console
$ iim use go 0.4.20
$ ipfs version
ipfs version 0.4.20
$ echo "A" > a.txt
$ echo "B" > b.txt
$ ipfs add a.txt b.txt
added QmNRpQVA5n7osjtyjYaWEQpwYnbV1QoVkrSe7oyccMJh1m a.txt
 2 B / 4 B [================================>--------------------------------]  50.00%
```
For comparison, on 0.4.19

```console
$ iim use go 0.4.19
$ ipfs version
ipfs version 0.4.19
$ ipfs add a.txt b.txt
added QmNRpQVA5n7osjtyjYaWEQpwYnbV1QoVkrSe7oyccMJh1m a.txt
added QmbN4uJU4zirdG1g9qcYYAAHNCU6MRHZBSwTx61WGVKich b.txt
 4 B / 4 B [=================================================================] 100.00%
```

Same issue occurs via the http api, found will debugging a build failure on js-ipfs-http-client after updating to 0.4.20, see: https://github.com/ipfs/js-ipfs-http-client/pull/973

olizilla:
on the cli, it doesn't seem to be the argument order that determines which file will be added, but the first on when alphabetically ordered. If you add a.txt and b.txt it doesn't matter which order you pass them in, it's always the CID for a.txt that gets reported

```console
$ ipfs add b.txt a.txt
added QmNRpQVA5n7osjtyjYaWEQpwYnbV1QoVkrSe7oyccMJh1m a.txt
 2 B / 4 B [================================>--------------------------------]  50.00%

$ ipfs add a.txt b.txt
added QmNRpQVA5n7osjtyjYaWEQpwYnbV1QoVkrSe7oyccMJh1m a.txt
 2 B / 4 B [================================>--------------------------------]  50.00%
```

alanshaw:
> only returns the cid

It's a little worse than that - it actually only adds one file:

```console
⨎ ipfs add --only-hash extreme-closeup.jpg 
[...]

PR #6255, opened 2019-04-25
 add: Fix adding multiple files

magik6k:
See https://github.com/ipfs/go-ipfs/issues/6254

### DIFF
--- a/core/commands/add.go
+++ b/core/commands/add.go
@@ -181,24 +181,26 @@ You can now check what blocks have been created by:
 			return err
 		}
 
-		events := make(chan interface{}, adderOutChanSize)
-
-		var toadd files.Node = req.Files
-		name := ""
+		toadd := []files.Node{req.Files}
+		name := []string{""}
 		if !wrap {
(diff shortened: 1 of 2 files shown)


core/commands/add.go
```diff
@@ -230,42 +217,61 @@ You can now check what blocks have been created by:
...
-			if output.Path != nil {
-				h = enc.Encode(output.Path.Cid())
-			}
+		var added int
+		addit := toadd.Entries()
+		for addit.Next() {
+			_, dir := addit.Node().(files.Directory)
+			errCh := make(chan error, 1)
+			events := make(chan interface{}, adderOutChanSize)
+			opts[len(opts)-1] = options.Unixfs.Events(events)
```
Stebalien: Maybe we shouldn't be closing this channel?

Stebalien: Otherwise, we could drop the need for the nil option by writing:

```
opts := append(opts, options.Unixfs.Events(events)) // note the :=
```

PR #6256, opened 2019-04-25
add: reuse the event channel

Stebalien:
@magik6k thoughts? I'm not sure if this is any simpler...

### DIFF
--- a/core/commands/add.go
+++ b/core/commands/add.go
@@ -188,6 +188,8 @@ You can now check what blocks have been created by:
 			})
 		}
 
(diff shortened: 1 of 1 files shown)


core/commands/add.go
```diff
@@ -188,6 +188,8 @@ You can now check what blocks have been created by:
 			})
 		}
 
+		events := make(chan interface{}, adderOutChanSize)
```
magik6k: Sharness complains, probably because we still have buffered events when we get status on `errCh`

Stebalien:
Meh, the previous version is probably simpler.

PR #6258, opened 2019-04-25
humanize for ipfs bitswap stat

myself659:
License: MIT
Signed-off-by: Eric Wu <myself659@163.com>

### DIFF
--- a/core/commands/bitswap.go
+++ b/core/commands/bitswap.go
@@ -90,6 +90,7 @@ Print out all blocks currently on the bitswap wantlist for the local peer.`,
 
 const (
 	bitswapVerboseOptionName = "verbose"
+	bitswapHumanOptionName   = "human"
 )
 
 var bitswapStatCmd = &cmds.Command{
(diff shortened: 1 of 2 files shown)


Stebalien:
To match our other commands, we should:

1. Add a --humanize flag.
2. _Replace_ the sizes with the humanized sizes when humanizing.

myself659:
my little suggest: 
bitswap-stat
as shown in the above picture， I think this way is better with adding --humanize flag.   and it has a better user experience.

myself659:
and I think more users will like this type of cmd.

You can discuss it internally. Thx!

Stebalien:
So, I'm just trying to be consistent with the rest of go-ipfs. Other commands (`ls`, `du`, etc.) provide humanize flags so `ifconfig` is really the outlier here. Actually, the replacement for `ifconfig`, `ip`, also uses a `-human` flag (`ip -s -human addr`).

Stebalien:
Thanks!

Stebalien:
cc @ipfs/wg-js-core

Issue #6002, opened 2019-02-18
Implement offline flag for remaining commands


Stebalien:
Missing commands:

* [ ]refs
* [ ] tar
* [ ] urlstore

To do this, I believe we need the CoreAPI to expose a potentially "offline" but non-pinning DAG.

PR #6259, opened 2019-04-26
commands(feat): use the coreapi in the urlstore command

Stebalien:
This is the first step in deprecating this command and works towards fixing #6002.

### DIFF
--- a/core/commands/urlstore.go
+++ b/core/commands/urlstore.go
@@ -27,7 +27,7 @@ var urlAdd = &cmds.Command{
 	Helptext: cmdkit.HelpText{
 		Tagline: "Add URL via urlstore.",
 		LongDescription: `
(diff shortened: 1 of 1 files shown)


core/commands/urlstore.go
```diff
@@ -32,17 +27,15 @@ var urlAdd = &cmds.Command{
 	Helptext: cmdkit.HelpText{
 		Tagline: "Add URL via urlstore.",
 		LongDescription: `
+DEPRECATED: Use 'ipfs add --nocopy URL'.
```
magik6k: We could also log errors so it's impossible to not notice

core/commands/urlstore.go
```diff
@@ -74,68 +67,40 @@ time.
 			return filestore.ErrUrlstoreNotEnabled
```
magik6k: We probably can drop this check

PR #6263, opened 2019-04-26
attempt to fix concurrent websocket write

Stebalien:
Unfortunately, I can't reproduce the bug locally and I'm not sure if my fix is correct. But it shouldn't hurt.


Stebalien:
dep-only update.

Issue #6237, opened 2019-04-19
Too many open files (regression?)


koalalorenzo:
#### Version information:
```
go-ipfs version: 0.4.20-
Repo version: 7
System version: amd64/linux
Golang version: go1.12.4
```
#### Type: Bug(?)
maybe bug, maybe misconfiguration after upgrading

#### Description:
I have been meeting this issue on a set of different machines of different types with different networks and different configurations. All the nodes are providing a public writable gateway. ( https://siderus.io/ )  All of them reported this:

```
Apr 19 02:33:06 usa-valley2 ipfs[3329]: 02:33:06.470 ERROR     flatfs: too many open files, retrying in 100ms flatfs.go:385
Apr 19 07:22:40 apac-tokyo1 ipfs[533]: 2019/04/19 07:22:40 http: Accept error: accept tcp4 127.0.0.1:8080: accept4: too many open files; retrying in 5ms
Apr 19 09:25:33 eu-stockholm1 ipfs[15101]: 2019/04/19 09:25:33 http: Accept error: accept tcp4 127.0.0.1:8080: accept4: too many open files; retrying in 320ms
Apr 19 09:25:33 eu-stockholm1 ipfs[15101]: 09:25:33.959 ERROR     flatfs: too many open files, retrying in 100ms flatfs.go:385
```

I found this issue after upgrading. The upgrade was executed automatically (no human touch) and all of the machines are similar with the exception of network, cpu and memory that is different based on the average traffic. Since this version we stopped using docker and started using our own Debian packages but previous tests were successful (12 hours of usage of a/b testing). The issue appeared recently and it seems that it might be caused by a set of requests from the users combined with a misconfiguration (`ulimit`?)

Maybe related #4102 #3792 (?)

Is this a bug or should I change the limits? How can I ensure that IPFS respects that limit?

Stebalien:
Could you cat `/proc/$(pidof ipfs)/limits` and the output of `ipfs swarm peers | wc -l`? Go-ipfs automatically raises it's limits put probably not enough to be run as a public gateway _directly_, assuming that's what you're doing.

Note: You may also want to consider sticking an nginx reverse proxy out in front. That should reduce the number of inbound connections.

koalalorenzo:
> Note: You may also want to consider sticking an nginx reverse proxy out in front. That should reduce the number of inbound connections.

[...]

PR #6266, opened 2019-04-26
raise default fd limit to 8192

### LINKED ISSUE #6247: Raise file descriptor limit to system max
We currently raise it to 2048 but there's no good reason not to set this to the absolute max.

Stebalien:
fixes #6247

Really, we need a global _resource_ manager service that can sum requests from
the datastore, libp2p, etc. for more file descriptors. However, we don't have
that.

License: MIT

### DIFF
--- a/cmd/ipfs/util/ulimit.go
+++ b/cmd/ipfs/util/ulimit.go
@@ -1,7 +1,6 @@
 package util
 
 import (
-	"errors"
 	"fmt"
 	"os"
 	"strconv"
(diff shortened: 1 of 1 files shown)


koalalorenzo:
Is this also an attempt to fix #6237  ?

Stebalien:
@koalalorenzo this is an attempt to alleviate it but we'd still like to know why we suddenly started using more file descriptors in 0.4.20 as that usually indicates a bug. However, in this case, the usual culprit, a file descriptor leak, looks unlikely.

PR #6268, opened 2019-04-26
docs: document environment variables

### LINKED ISSUE #5828: Document IPFS_PATH
I think `IPFS_PATH` should be documented. That it's an environment variable, and what it implies. It's referenced in various locations in the docs, but only explicitly offhandedly mentioned that it's an an environment variable in the README. If there are other environment variables that affect the ipfs cmd, it might be grouping the descriptions of these together in a new document.

Stebalien:
fixes #5828

### DIFF
--- a/docs/environment-variables.md
+++ b/docs/environment-variables.md
@@ -1,0 +1,89 @@
+# go-ipfs environment variables
+
+## `LIBP2P_TCP_REUSEPORT` (`IPFS_REUSEPORT`)
+
+go-ipfs tries to reuse the same source port for all connections to improve NAT
+traversal. If this is an issue, you can disable it by setting
+`LIBP2P_TCP_REUSEPORT` to false.
+
+This variable was previously `IPFS_REUSEPORT`.
+
+Default: true
+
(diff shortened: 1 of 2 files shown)


Stebalien:
> Re environment vars: I did notice while looking around that we have a reference to the environment variable "IPFS_LOW_MEM" (./go-ipfs-flags/flags.go:11: if os.Getenv("IPFS_LOW_MEM") != "" {), but I don't actually see it being used anywhere. I mention it just in case. This list otherwise looks like the relevant stuff I'm finding.

It looks like it _is_ used by bitswap.

> Re mux: I can't comment on the stability, but I can take your word for it. For my own benefit, can you quickly elaborate on this experiment?

We have two stream multiplexers: yamux and mplex. We use yamux by default but neither js-libp2p nor rust-libp2p support yamux so we fallback on mplex.

Stebalien:
> IPFS_LOW_MEM

Removed in https://github.com/ipfs/go-bitswap/pull/115.

1. This currently _increases_ memory usage.
2. We should just make this a configuration option.

PR #6274, opened 2019-04-29
WIP Fallback coreapi client

magik6k:
This is part of my plan to revive ipget.

This code is supposed to provide what https://github.com/noffle/fallback-ipfs-shell / https://github.com/whyrusleeping/ipfs-embedded-shell provide, but on CoreAPI level.

(not sure about naming / location / should this be a separate package)

### DIFF
--- a/core/coreapi/client/client.go
+++ b/core/coreapi/client/client.go
@@ -1,0 +1,146 @@
+package client
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+
+	config "github.com/ipfs/go-ipfs-config"
+	"github.com/ipfs/go-ipfs-http-client"
+	"github.com/ipfs/interface-go-ipfs-core"
+	"github.com/mitchellh/go-homedir"
+	"github.com/pkg/errors"
+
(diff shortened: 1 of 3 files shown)


core/coreapi/client/client.go
```diff
@@ -0,0 +1,146 @@
...
+	io.Closer
+}
+
+type ctxApi struct {
+	iface.CoreAPI
+	context.CancelFunc
+}
+
+func (a *ctxApi) Close() error {
+	a.CancelFunc()
```
Stebalien: We should probably _wait_ for the node to exit here.

core/coreapi/client/client.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	return &ctxApi{
+		CoreAPI: api,
+		CancelFunc: func() {},
+	}, nil
+}
+
+var tempEmbedded ApiProvider = func() (ManagedApi, error) {
+	dir, err := ioutil.TempDir("", "ipfs-shell")
```
Stebalien: Ideally, we'd remove this directory (wrap the API so it gets deleted on close?).

Issue #6232, opened 2019-04-18
RFC: Invert DI Constructor Config Handling


Stebalien:
The DI-based constructor currently has a bunch of options that take the config and then spit out some service.

I'd like to consider replacing these with a single super-option constructor that takes the config and then constructs the options necessary to start the node described in that config.

Rational:

* Ideally, applications using libp2p will configure libp2p through options instead of through a config file. This isn't currently possible
* I believe this will make it easier to reason about the config (centralizes everything).

---

Thoughts @magik6k?

Stebalien:
For example, we'd replace options like:

```go
func Pubsub(mctx MetricsCtx, lc fx.Lifecycle, host host.Host, cfg *config.Config) (service *pubsub.PubSub, err error) {
	var pubsubOptions []pubsub.Option
	if cfg.Pubsub.DisableSigning {
		pubsubOptions = append(pubsubOptions, pubsub.WithMessageSigning(false))
	}

	if cfg.Pubsub.StrictSignatureVerification {
		pubsubOptions = append(pubsubOptions, pubsub.WithStrictSignatureVerification(true))
	}

	switch cfg.Pubsub.Router {
	case "":
		fallthrough
	case "floodsub":
		service, err = pubsub.NewFloodSub(lifecycleCtx(mctx, lc), host, pubsubOptions...)

	case "gossipsub":
		service, err = pubsub.NewGossipSub(lifecycleCtx(mctx, lc), host, pubsubOptions...)

	default:
		err = fmt.Errorf("Unknown pubsub router %s", cfg.Pubsub.Router)
	}

	return service, err
}
```

With:

```go
func FloodSub(pubsubOptions ...pubsub.Option) interface{} {
	return func (mctx MetricsCtx, lc fx.Lifecycle, host host.Host) (service *pubsub.PubSub, err error) {
		return pubsub.NewFloodSub(lifecycleCtx(mctx, lc), host, pubsubOptions...)
	}
}

func GossipSub(pubsubOptions ...pubsub.Option) interface{} {
	return func (mctx MetricsCtx, lc fx.Lifecycle, host host.Host) (service *pubsub.PubSub, err error) {
		return pubsub.NewGossipSub(lifecycleCtx(mctx, lc), host, pubsubOptions...)
	}
}

func FromConfig(cfg *config.Config) (fx.Option, error) {
	var opts []fx.Option

    // ...

	// Could live in a different function that _returns_ an `fx.Option` but it doesn't have to.
[...]

PR #6276, opened 2019-04-29
Invert constructor config handling 

magik6k:
See https://github.com/ipfs/go-ipfs/issues/6232

The diff is slightly larger that I initially thought it would be, but it looks reasonably good (there is still room for improvement, but I wanted to keep the PR size down)


core/node/groups.go
```diff
@@ -2,105 +2,284 @@ package node
...
 		BaseLibP2P,
 
-		fx.Provide(libp2p.Security(!cfg.DisableEncryptedConnections)),
-		maybeProvide(libp2p.Pubsub, cfg.getOpt("pubsub") || cfg.getOpt("ipnsps")),
+		fx.Provide(libp2p.AddrFilters(cfg.Swarm.AddrFilters)),
+		fx.Invoke(libp2p.SetupDiscovery(cfg.Discovery.MDNS.Enabled, cfg.Discovery.MDNS.Interval)),
+		fx.Provide(libp2p.AddrsFactory(cfg.Addresses.Announce, cfg.Addresses.NoAnnounce)),
+		fx.Provide(libp2p.SmuxTransport(bcfg.getOpt("mplex"))),
+		fx.Provide(libp2p.Relay(cfg.Swarm.DisableRelay, cfg.Swarm.EnableRelayHop)),
+		fx.Invoke(libp2p.StartListening(cfg.Addresses.Swarm)),
```
Stebalien: We can handle this later but I wonder if this should go elsewhere; somewhere higher up the stack.

core/node/libp2p/nat.go
```diff
@@ -0,0 +1,32 @@
...
+		// collect private net option in case swarm.key is presented
+		opts, _, err := PNet(repo)
+		if err != nil {
+			// swarm key exists but was failed to decode
+			return err
+		}
+
+		if quic {
+			opts.Opts = append(opts.Opts, libp2p.DefaultTransports, libp2p.Transport(libp2pquic.NewTransport))
+		}
```
Stebalien: Small note: Once libp2p starts using fx, this will go away. At that point, we should be able to just pull in the transports group.

Issue #3124, opened 2016-08-25
corrupted ipns entry in local cache


whyrusleeping:
I was able to break my local ipns publishing:

```
➜  ~ ipfs name publish <some hash>
Error: proto: required field "{Unknown}" not set
```

I did this by first publishing an entry normally, then:
```
ipfs dht get /ipns/<my peer ID> > ipns-data
cat ipns-data | ipfs dht put /ipns/<my peer ID>
```

Something must not be quite right here in the round tripping of the record.

This also appears to make it impossible for me to publish new ipns values for the time being.

jbenet:
That would be a great test case of you can get a repro

Stebalien:
Well, the good news is that this now returns an error on put. The bad news is that `get` is still broken.

PR #6277, opened 2019-04-30
work towards fixing dht commands

Stebalien:
Specifically, #3124

Unfortunately, `dht put` is still broken. However, I can confirm that this fixes
`dht get`.


Stebalien:
That issue isn't _completely_ fixed but this is still an improvement.

Stebalien:
Well, I guess I forgot to mention: this is a breaking change! However, there's nothing much we can do about that as DHT records are binary and we _need_ to encode them somehow. Given that this API was fundamentally broken, :shipit:.

djdv:
>We're now base64 encoding values returned by dht get when encoding them into JSON. 

Probably useful for the release notes.
(Explicit back reference for later)
https://github.com/ipfs/go-ipfs/issues/6305

PR #6292, opened 2019-05-02
Introduce first strategic provider: do nothing

michaelavila:
These commits introduce an experimental provider system that, currently, does not provide. Functionality will be layered on to the experimental provider until it has all of the behavior needed to replace the default provider system.

The experimental system can be enabled by issuing the following command:

```sh
ipfs config Experimental.StrategicProviding true
```

Then you **must restart your daemon**.


go.mod
```diff
@@ -34,7 +34,7 @@ require (
 	github.com/ipfs/go-ipfs-chunker v0.0.1
 	github.com/ipfs/go-ipfs-cmdkit v0.0.1
 	github.com/ipfs/go-ipfs-cmds v0.0.5
-	github.com/ipfs/go-ipfs-config v0.0.3
+	github.com/ipfs/go-ipfs-config v0.0.4-0.20190502164316-b06d585017ae
```
michaelavila: Will tag when https://github.com/ipfs/go-ipfs-config/pull/33 is merged.

core/node/groups.go
```diff
@@ -272,17 +234,19 @@ func Online(bcfg *BuildCfg, cfg *config.Config) fx.Option {
...
 }
 
 // Offline groups offline alternatives to Online units
-var Offline = fx.Options(
-	fx.Provide(offline.Exchange),
-	fx.Provide(Namesys(0)),
-	fx.Provide(offroute.NewOfflineRouter),
-	fx.Provide(provider.NewOfflineProvider),
-)
+func Offline(cfg *config.Config) fx.Option {
```
michaelavila: @magik6k I needed to change `Offline` to be a function like `Online`. Let me know if you have concerns about this.

core/node/provider.go
```diff
@@ -2,50 +2,129 @@ package node
...
+
+// StrategicOfflineProviderSysCtor creates a new offline provider system
+func StrategicOfflineProviderSysCtor() provider.System {
+	return provider.NewOfflineProvider()
+}
+
+// ONLINE/OFFLINE
+
+// OnlineProviders groups units managing provider routing records online
+func OnlineProviders(cfg *config.Config) fx.Option {
```
michaelavila: @magik6k I also want to point out that I moved the `Providers` function (used for DI) to the provider node. I did this because there's a bit of logic I wanted to keep near the provider code (like taking into account the experimental flag).

provider/system.go
```diff
@@ -0,0 +1,47 @@
...
+
+// Run the provider system by running the provider and reprovider
+func (s *system) Run() {
+	go s.provider.Run()
+	go s.reprovider.Run()
+}
+
+// Close the provider and reprovider
+func (s *system) Close() error {
+	// TODO: Close reprovider here
```
michaelavila: This doesn't need to be implemented yet because close is a noop on the reprovider at the moment. In the next PR I'm introducing a new reprovider that has close and so will need to be accounted for here.

michaelavila:
@scout I think you'll find this PR useful. You mentioned to me not long ago that you'd like the ability to disable providing in IPFS without disabling content routing. This PR does exactly that. A series of upcoming PRs will flesh out the provider system further, but the ability to disable providing without disabling content routing itself will be preserved.

Is this something you can put to use right away? In its current form you'd need to set an experimental flag and restart the daemon. Will that work for what you have in mind?

eefahy:
Thanks very much @michaelavila ! This will help a ton with debugging and observing bona-fide gateway traffic and save us some bandwidth costs to boot!

core/node/provider.go
magik6k: Anything that touches config should generally live in `groups.go`. This might seem a bit weird right now, but hopefully will make much more sense after the constructor refactors are finished

provider/provider.go
```diff
@@ -1,18 +1,18 @@
...
 	"github.com/ipfs/go-cid"
 	logging "github.com/ipfs/go-log"
-	"github.com/libp2p/go-libp2p-routing"
 )
 
-var log = logging.Logger("provider")
+var (
+	// StrategicProvidingEnabled toggles between the original providing mechanism
+	// and the new strategic providing system
+	StrategicProvidingEnabled = false
```
magik6k: I'd avoid using global vars to configure subsystems, it tends to make some things painful (parallel go tests to name one example)

provider/simple/provider.go
```diff
@@ -0,0 +1,72 @@
...
+
+	cid "github.com/ipfs/go-cid"
+	q "github.com/ipfs/go-ipfs/provider/queue"
+	logging "github.com/ipfs/go-log"
+	routing "github.com/libp2p/go-libp2p-routing"
+)
+
+var logP = logging.Logger("provider.simple")
+
+const provideOutgoingWorkerLimit = 8
```
magik6k: It would be great to be able to set this from config (can be done in a separate PR)

core/node/provider.go
```diff
@@ -2,50 +2,129 @@ package node
...
 		OnStop: func(ctx context.Context) error {
-			return p.Close()
+			return sys.Close()
 		},
 	})
+	return sys
+}
+
+// SimpleOfflineProviderSysCtor creates a new offline provider system
+func SimpleOfflineProviderSysCtor(p provider.Provider, r provider.Reprovider) provider.System {
```
magik6k: We don't really need to wrap this as can just call `fx.Provide(provider.NewSystem)`

core/node/provider.go
```diff
@@ -2,50 +2,129 @@ package node
...
+
+// STRATEGIC
+
+// StrategicProviderSysCtor creates new provider system
+func StrategicProviderSysCtor() provider.System {
+	return provider.NewOfflineProvider()
+}
+
+// StrategicOfflineProviderSysCtor creates a new offline provider system
+func StrategicOfflineProviderSysCtor() provider.System {
```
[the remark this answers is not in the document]
magik6k: Also, this is the same as `StrategicProviderSysCtor` (and we can just do `fx.Provide(provider.NewOfflineProvider)`, without the wrapping methods)

provider/offline.go
```diff
@@ -1,20 +1,28 @@
...
+func (op *offlineProvider) Run() {
+}
 
-func (op *offlineProvider) Provide(cid cid.Cid) error {
+func (op *offlineProvider) Close() error {
 	return nil
 }
 
-func (op *offlineProvider) Close() error {
+func (op *offlineProvider) Provide(_ cid.Cid) error {
```
magik6k:
```suggestion
func (op *offlineProvider) Provide(cid.Cid) error {
```

provider/offline.go
```diff
@@ -1,20 +1,28 @@
...
+	return nil
+}
+
+func (op *offlineProvider) Reprovide(_ context.Context) error {
```
magik6k:
```suggestion
func (op *offlineProvider) Reprovide(context.Context) error {
```

provider/simple/reprovide.go
```diff
@@ -0,0 +1,225 @@
...
+
+		rsys:        rsys,
+		keyProvider: keyProvider,
+		tick:        reprovideIniterval,
+	}
+}
+
+// Close the reprovider
+func (rp *Reprovider) Close() error {
+	return nil
```
magik6k: This probably should close `rp.ctx`

provider/system.go
magik6k: Why not just call it anyways so we won't forget about it?

test/sharness/lib/iptb-lib.sh
```diff
@@ -62,3 +62,24 @@ iptb_wait_stop() {
         go-sleep 10ms
     done
 }
+
+findprovs_empty() {
```
magik6k: This should be in `test-lib.sh`

magik6k:
(also, more smaller commits would make reviewing easier)

provider/simple/reprovide.go
michaelavila: I can add this if we want, but my goal was to change the existing system as little as possible. Not closing `rp.ctx` means it continues behaving the way it has, which I'm OK with since my goal is to replace the whole system and don't want to unintentionally cause a problem that diverts my energy. What do you think?

core/node/provider.go
michaelavila: I'd like to leave these right now as the names communicate my intent and in the next PR I'm going to make better use of this break up. I will definitely pare down what's not needed from there. Let me know if you object to this.

core/node/provider.go
```diff
@@ -2,50 +2,129 @@ package node
...
+
+// SimpleReproviderCtor creates new reprovider
+func SimpleReproviderCtor(reproviderInterval time.Duration) func(helpers.MetricsCtx, fx.Lifecycle, routing.IpfsRouting, simple.KeyChanFunc) (provider.Reprovider, error) {
+	return func(mctx helpers.MetricsCtx, lc fx.Lifecycle, rt routing.IpfsRouting, keyProvider simple.KeyChanFunc) (provider.Reprovider, error) {
+		return simple.NewReprovider(helpers.LifecycleCtx(mctx, lc), reproviderInterval, rt, keyProvider), nil
+	}
+}
 
+// SimpleProviderSysCtor creates new provider system
+func SimpleProviderSysCtor(lc fx.Lifecycle, p provider.Provider, r provider.Reprovider) provider.System {
```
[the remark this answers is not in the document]
michaelavila: I'd like to leave these right now as the names communicate my intent and in the next PR I'm going to make better use of this break up. I will definitely pare down what's not needed from there. Let me know if you object to this.

michaelavila:
@magik6k thanks for the thoughtful review. I've incorporated the majority of your feedback and left a few responses for you to follow up on.

core/node/core.go
```diff
@@ -59,8 +60,9 @@ func Dag(bs blockservice.BlockService) format.DAGService {
 }
 
 // OnlineExchange creates new LibP2P backed block exchange (BitSwap)
-func OnlineExchange(mctx helpers.MetricsCtx, lc fx.Lifecycle, host host.Host, rt routing.IpfsRouting, bs blockstore.GCBlockstore) exchange.Interface {
+func OnlineExchange(mctx helpers.MetricsCtx, lc fx.Lifecycle, cfg *config.Config, host host.Host, rt routing.IpfsRouting, bs blockstore.GCBlockstore) exchange.Interface {
```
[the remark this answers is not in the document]
michaelavila: Understood. It makes sense and I've made this change throughout.

magik6k:
> I had more, small commits but ran into some rebase confusion and decided to squash to make the rebase easier

I'd say it's fine to just merge master into your branch if rebasing gets painful. This way you only resolve conflicts once

core/node/provider.go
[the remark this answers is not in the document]
magik6k: I generally prefer leaving things clean, even if I know that I'll change them very soon, as sometimes (especially with big PR) that tends to never happen

provider/simple/reprovide.go
magik6k: It's probably fine

michaelavila:
@magik6k all feedback has been addressed.

provider/provider.go
michaelavila: I just want to clarify here that this was an artifact from an early version and I just removed it. The experimental subsystem configuration is done by constructing different types of objects.

core/node/groups.go
```diff
@@ -264,25 +226,27 @@ func Online(bcfg *BuildCfg, cfg *config.Config) fx.Option {
 	}
 
 	return fx.Options(
-		fx.Provide(OnlineExchange),
+		fx.Provide(OnlineExchange(!cfg.Experimental.StrategicProviding)),
```
Stebalien: Can we comment like `/* don't provide form bitswap when the strategic provider service is active */`?

michaelavila: I used a variable with a more descriptive name and added the comment.

core/node/core.go
```diff
@@ -59,15 +59,19 @@ func Dag(bs blockservice.BlockService) format.DAGService {
...
-	lc.Append(fx.Hook{
-		OnStop: func(ctx context.Context) error {
-			return exch.Close()
-		},
-	})
-	return exch
+func OnlineExchange(provide bool) interface{} {
+	return func(mctx helpers.MetricsCtx, lc fx.Lifecycle, host host.Host, rt routing.IpfsRouting, bs blockstore.GCBlockstore) exchange.Interface {
+		bitswapNetwork := network.NewFromIpfsHost(host, rt)
+		bitswap.ProvideEnabled = provide
```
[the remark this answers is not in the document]
michaelavila: Understood. I used the `UseHAMTSharding` experimental flag as prior art, but I will make this change. This will require a change in `go-bitswap` as that has already been merged.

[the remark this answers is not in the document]
Stebalien: :disappointed:... Yeah, I think that's because _that_ one:

1. Is hard to feed through. It isn't used in a specific subsystem but by anything that creates a directory.
2. Was probably supposed to graduate from experimental pretty quickly.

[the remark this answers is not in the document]
michaelavila: Cool. Thanks for pointing it out, no a big change. I'll push shortly.

michaelavila:
NOTE TO SELF: Add documentation for the experimental feature in the experimental features document before merging.

provider/simple/reprovide.go
```diff
@@ -0,0 +1,225 @@
...
+		}
+
+		unmute()
+
+		after = time.After(rp.tick)
+	}
+}
+
+// Reprovide registers all keys given by rp.keyProvider to libp2p content routing
+func (rp *Reprovider) Reprovide() error {
```
Stebalien: Should this be exported? That test should probably be calling `Trigger` instead.

michaelavila: @Stebalien I just moved this code and the corresponding tests without changing them. It looks like, in order to call `Trigger` instead, the `Reprovider` needs to be `Run()` in the tests, which isn't currently happening. Are you OK just leaving this as it has been? I'm trying to get it deprecated anyway.

Stebalien: SGTM.

michaelavila:
@magik6k can you please re-review and remove your requested changes? All your feedback has been addressed.

michaelavila:
@Stebalien barring any other feedback, this PR is ready to go.

Stebalien:
SGTM. We'll try to get the pending release out ASAP and then merge this. I'm uploading an RC1 now.

michaelavila:
@magik6k no worries, thanks for following up!

Stebalien:
:rocket:

PR #6301, opened 2019-05-06
License update to dual MIT and Apache 2

momack2:
This series of commits aims to update go-ipfs to follow a dual-licensing best practice based on research into open-source licensing by @ianjdarrow. He recommends a dual MIT and Apache 2.0 license -

> This has two major benefits:

> - There are concerns in the open source community about whether the MIT license leaves users vulnerable to patent infringement claims. We think the pure legal risk is small, but the way the open source community interacts with our project is really important. It makes sense to pick the license that makes the largest number of people comfortable.
- There's now no reason to adopt a separate DCO, since the Apache-2 license grant addresses the same issue.

> Why use a dual license, instead of just Apache-2? The Apache-2 license is incompatible with the GPLv2 license, which includes things like the Linux kernel. With a dual license, GPLv2 projects can just use the MIT license instead. Our goal is to make our software available to as many projects as possible, so we'd rather adopt a licensing scheme that doesn't exclude anyone.

In addition to these commits, we also need to get an explicit OK from current and past contributors to give their consent to relicensing - which will happen in an issue thread.

### DIFF
--- a/COPYRIGHT
+++ b/COPYRIGHT
@@ -1,0 +1,1 @@
+This library is dual-licensed under Apache 2.0 and MIT terms.
--- a/COPYRIGHT
+++ b/COPYRIGHT
@@ -1,1 +1,5 @@
-This library is dual-licensed under Apache 2.0 and MIT terms.
+This project is transitioning from an MIT-only license to a dual MIT/Apache-2.0 license. 
+Unless otherwise noted, all code contributed prior to 2019-05-06 and not contributed by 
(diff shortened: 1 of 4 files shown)


momack2:
Notes from Legal for further reference:
- not having the `© so-and-so` line here is fine, since in MIT and Apache2 licenses the copyrights are retained by the contributors themselves
- if any humans refuse to make the transition from mit to Mit+apache2 after a reasonable percentage of people have acked, we can just call out those particular contributions in the license itself (or rewrite the code to not be dependent on the original contribution)

Issue #251, opened 2014-11-02
IPFS_BOOTSTRAP / env-override of config?


jbenet:
As noted in #250, it would be nice to be able to pass in bootstrap addresses as part of an envvar or a cli param. This would make it easier to programmatically create nodes.

What might make sense is to design a seamless way for env vars to override configuration variables, that way we dont have to create multiple ways of changing them.

Maybe a `IPFS_CONFIG(_Key)+` ->`config(.Key)+` approach. (e.g.`IPFS_Config_Bootstrap`)... hmm i dont like this.

diasdavid:
We can keep it simple by using env vars and giving the option to extend the default config with `--config=` option

hannahhoward:
@gmasgras can you comment on whether this is still relevant based on the Docker-izing work you are doing? Should it remain open?

gmasgras:
Here is a PR that addresses this issue: https://github.com/ipfs/go-ipfs/pull/6325

lidel:
As clarified in https://github.com/ipfs/go-ipfs/pull/8326#issuecomment-912726852:
>  If we ever add config override via env variable, it should be supported in go-ipfs natively ([`go-ipfs/docs/environment-variables.md`](https://github.com/ipfs/go-ipfs/blob/master/docs/environment-variables.md)), and work as "ad-hoc" ephemeral override that does not mutate the config file.

PR #6325, opened 2019-05-13
Feat/config envs docker

gmasgras:
Implements https://github.com/ipfs/go-ipfs/issues/251

Allows passing ENVs to the Docker container to update go-ipfs configuration in `/data/ipfs/config`:
The ENVs should being with `IPFS_CONFIG_` and are case sensitive. Nested properties are expressed via the `_` separator:
 ```
"Swarm": {
    "AddrFilters": null,
    "ConnMgr": {
      "LowWater": 200,`
    }
  }
```
can be set via `IPFS_CONFIG_Swarm_ConnMgr_LowWater=200`

Multiple ENVs:
```
docker run -it \
-e IPFS_CONFIG_Swarm_ConnMgr_LowWater=200 \
-e IPFS_CONFIG_Swarm_ConnMgr_GracePeriod=49s \
-e IPFS_CONFIG_Gateway_Writable=false \
-e IPFS_CONFIG_Swarm_ConnMgr_Type=basic \
-e IPFS_CONFIG_Bootstrap="["/dnsaddr/bootstrap.libp2p.io/ipfs/QmNnooDu7bfjPFoTZYxMNLWUQJyrVwtbZg5gBMjTezGAJN","/dnsaddr/bootstrap.libp2p.io/ipfs/QmQCU2EcMqAqQPR2i9bChDtGNJchTbq5TbXJJ16u]"" \
go-ipfs:latest
```

### DIFF
--- a/bin/container_daemon
+++ b/bin/container_daemon
@@ -1,5 +1,38 @@
 #!/bin/sh
 set -e
+
+envToConfig() {
+  ipfsPrefix="IPFS_CONFIG"
+  configFile="$IPFS_PATH/config"
+  configFileTmp="$(mktemp)"
+
+  env |
+    while IFS='=' read -r name value
+    do
+      case "$name" in
+        "$ipfsPrefix"*)
+          parsedName=${name#$ipfsPrefix*}
+          jsonName=${parsedName//_/.}
+          case "$value" in
+            [*)
+               # encode as json array
+               arr=$(echo "$value" |tr -d '[]' | tr ', ' '\n')
+               jsonValue=$(jq -nr --arg value "$arr" '$value|split("\n")')
+               ;;
+             *)
+               # ensure value is json-quoted if needed
+               jsonValue=$(jq -nr --argjson value "${value}" '$value|tojson'  2>/dev/null || jq -nr --arg value "${value}" '$value|tojson' 2>/dev/null)
+               ;;
+          esac
+
+          jqStr="$(printf '%s = %s' "$jsonName" "$jsonValue")"
+          jq --arg jqStr "$jqStr" ".| $jqStr" < "$configFile" > "$configFileTmp" && mv "$configFileTmp" "$configFile"
+          ;;
+      esac
+    done
+}
+
+
 user=ipfs
 repo="$IPFS_PATH"
 
@@ -22,6 +55,7 @@ else
     *) INIT_ARGS="--profile=$IPFS_PROFILE" ;;
   esac
   ipfs init $INIT_ARGS
+  envToConfig
   ipfs config Addresses.API /ip4/0.0.0.0/tcp/5001
   ipfs config Addresses.Gateway /ip4/0.0.0.0/tcp/8080
 fi
--- a/Dockerfile
+++ b/Dockerfile
@@ -55,6 +55,9 @@ RUN chmod 4755 /usr/local/bin/fusermount
 # This shared lib (part of glibc) doesn't seem to be included with busybox.
 COPY --from=0 /lib/x86_64-linux-gnu/libdl-2.24.so /lib/libdl.so.2
 
+RUN wget -q -O /usr/local/bin/jq https://github.com/stedolan/jq/releases/download/jq-1.5/jq-linux64 \
+  && chmod +x /usr/local/bin/jq
+
 # Swarm TCP; should be exposed to the public
 EXPOSE 4001
 # Daemon API; must not be exposed publicly but to client services under you control

### CODE REVIEW

**Comment by hannahhoward** [comment]
fix #251

File: bin/container_daemon [hunk 1]
```diff
@@ -1,5 +1,38 @@
...
+               # ensure value is json-quoted if needed
+               jsonValue=$(jq -nr --argjson value "${value}" '$value|tojson'  2>/dev/null || jq -nr --arg value "${value}" '$value|tojson' 2>/dev/null)
+               ;;
+          esac
+
+          jqStr="$(printf '%s = %s' "$jsonName" "$jsonValue")"
+          jq --arg jqStr "$jqStr" ".| $jqStr" < "$configFile" > "$configFileTmp" && mv "$configFileTmp" "$configFile"
+          ;;
+      esac
+    done
```
Stebalien: This seems really brittle. Do we need to parse json manually like this or can we just ask the user to pass in valid JSON?

Alternatively, could we just skip manually updating the config file and use `ipfs config --json key value`?

gmasgras: There isn't a good/easy way to pass valid (correctly quoted) JSON through ENVs. We would have to manually parse the value of each ENV and add quotes to it when it's a JSON string (not a boolean or numeric). `jq` takes care of that. 

I considered first using `ipfs config` but that would add extra parsing/checks when passing in booleans (needs--bool flag) or arrays (needs--json flag)

Stebalien:
```
MYVAR='{"foobar": "baz"}'
```

?

---

My concern here isn't really jq, it's the fact that this code will break if it runs into any commas, square brackets, or newlines in a json array. And I'm pretty sure I'm missing a bunch of edge cases/bugs there.

gmasgras: That's not how we'd set `foobar` with this script. It will look more like `IPFS_CONFIG_foobar=baz`.
Array values are handled differently in lines 18-21

Stebalien: >  Array values are handled differently in lines 18-21

That's the buggy code I'm talking about. That's fine for a one-off script but not nearly robust enough for code we have to maintain.

> That's not how we'd set foobar with this script. It will look more like `IPFS_CONFIG_foobar=baz`.

So we're just guessing the type? That'll fail for fields like `Reprovider.Interval` (which is a string/duration but can be a number in the config).

gmasgras: > That's the buggy code I'm talking about. That's fine for a one-off script but not nearly robust enough for code we have to maintain.

I'm not sure why you believe this is buggy, it assumes that anything starting with `[` is an array. I think that's a reasonable assumption.


> So we're just guessing the type? That'll fail for fields like `Reprovider.Interval` (which is a string/duration but can be a number in the config).

No, that doesn't actually fail:
```
/ # export value=120
/ # jq -nr --argjson value "${value}" '$value|tojson'  2>/dev/null || jq -nr --arg value "${value}" '$value|tojson' 2>/dev/null
120
/ # export value=12h
/ # jq -nr --argjson value "${value}" '$value|tojson'  2>/dev/null || jq -nr --arg value "${value}" '$value|tojson' 2>/dev/null
"12h"
```
And we're not "guessing" it, we're using jq to properly format it. Guessing it would be doing a bunch of if checks and basically re-implementing `jq`. 

This is absolutely not meant to be a generic ENV to JSON converter. (JSON is not even meant to be a config file format for the very reason that it shouldn't be necessary to need to know the type of a configuration value)

I think this is reasonable for our use-case of enabling infra (and others) to easily configure go-ipfs when run as a Docker container.

Stebalien: > I'm not sure why you believe this is buggy, it assumes that anything starting with [ is an array. I think that's a reasonable assumption.

It:

* Deletes all brackets commas, regardless of where they show up.
* Will turn something like `[/foo, /bar]` into `["/foo", "", "/bar"]`.
* Doesn't support arrays of anything other than strings.

> No, that doesn't actually fail:

That's exactly what I mean by fail. `Reprovider.Interval` _must_ be a string but there's no way to specify a string with a value like "true", "0", "{}", etc. with this system.

---

So, instead of half-parsing json with some bash, let's just use json and keep this simple. We could also go with a typed approach, something like `IPFS_CONFIG_STRING_foo_bar=...` but it's probably not worth it.

Stebalien: > This is absolutely not meant to be a generic ENV to JSON converter. (JSON is not even meant to be a config file format for the very reason that it shouldn't be necessary to need to know the type of a configuration value)

This is something we have to maintain. If we implement this, users will use it and report bugs in it. I don't want to maintain a bash script of hacks on hacks as we try to paper over all the edge cases here.

Stebalien: To be clear, I think we should totally support this. I just want it to be robust. That's why I'm suggesting we just go with something like:

```
docker run -it \
-e IPFS_CONFIG_Swarm_ConnMgr_LowWater='200' \
-e IPFS_CONFIG_Swarm_ConnMgr_GracePeriod='"49s"' \
-e IPFS_CONFIG_Gateway_Writable='false' \
-e IPFS_CONFIG_Swarm_ConnMgr_Type='"basic"' \
-e IPFS_CONFIG_Bootstrap='"["/dnsaddr/bootstrap.libp2p.io/ipfs/QmNnooDu7bfjPFoTZYxMNLWUQJyrVwtbZg5gBMjTezGAJN","/dnsaddr/bootstrap.libp2p.io/ipfs/QmQCU2EcMqAqQPR2i9bChDtGNJchTbq5TbXJJ16u"]"' \
go-ipfs:latest
```

gmasgras: > there's no way to specify a string with a value like "true", "0", "{}", etc. with this system.

Where in the config is that needed or even possible (since it has to be valid JSON)? You are calling this code "hacks on hacks" because it doesn't cover some unrealistic use-case.. ?

The motivation for using `jq` to properly format the JSON values was to follow the robustness principle of "be conservative in what you do, be liberal in what you accept from others". 

I can (reluctantly, since it makes the user's life harder) remove that line and force the users to deal with passing in properly formatted JSON if it moves this PR forward.

Stebalien: > Where in the config is that needed or even possible (since it has to be valid JSON)?

This isn't what we need now but what we may need later. I don't want to box ourselves into a corner. That's why I said this is unmaintainable.

Also note, `Reprovider.Interval` is a string so we already need `"0"`.

> You are calling this code "hacks on hacks" because it doesn't cover some unrealistic use-case.. ?

I'm saying the code we'll end up with will be hacks on hacks as we try to cover all the edge cases as they come up. This code isn't yet hacks on hacks, it's just buggy.

---

We may also be able to rely on the type checking in go-ipfs. That is, we can run `ipfs config --json "$thing" "$value" || ipfs config "$thing" "$value"`. That will let go-ipfs try treating the value as both a json value and a string. The config is type-checked internally so that shouldn't cause any issues.

Issue #1331, opened 2015-06-04
ipfs object put should accept empty objects


whyrusleeping:
```
whyrusleeping@idril ~> echo \{\} | ipfs object put
Error: no data or links in this node
```

I expected it to just create me an empty object.

jbenet:
:+1:

dylanPowers:
Related to https://github.com/ipfs/go-ipfs/issues/1332

Stebalien:
Triage: Still a bug.

PR #6330, opened 2019-05-14
object put: Allow empty objects

magik6k:
Fixes https://github.com/ipfs/go-ipfs/issues/1331

### DIFF
--- a/core/coreapi/object.go
+++ b/core/coreapi/object.go
@@ -73,7 +73,9 @@ func (api *ObjectAPI) Put(ctx context.Context, src io.Reader, opts ...caopts.Obj
 	switch options.InputEnc {
 	case "json":
 		node := new(Node)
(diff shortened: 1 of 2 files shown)


Stebalien:
See: https://github.com/ipfs/go-ipfs/commit/443073142635d9703a8ff254c09aea543a0f14c7. I ran into trouble detecting mis-encoded XML objects.

magik6k:
If we really want to be correct on how we parse xml, we'd probably want to use `xml.Decoder` directly, but I'm not sure if it's worth the time in this case

Stebalien:
> If we really want to be correct on how we parse xml, we'd probably want to use xml.Decoder directly, but I'm not sure if it's worth the time in this case

Probably not. Want to just drop that test? However, I would backport my JSON error checking into this patch.

magik6k:
(done)

Issue #4721, opened 2018-02-19
dag put and block put return different headers


vmx:
#### Version information:
go-ipfs version: 0.4.14-dev-25eeb1e25
Repo version: 6
System version: amd64/linux
Golang version: go1.9.2

#### Description:

When using the HTTP API, a `put` to `dag` returns `X-Chunked-Output: 1` in the header, while a `put` to `block` does not. Example:

```
$ curl --verbose -F 'data=@/home/vmx/src/pl/somedata.dag-pb' 'http://localhost:5001/api/v0/dag/put?format=dag-pb&input-enc=raw'
*   Trying ::1...
* TCP_NODELAY set
* connect to ::1 port 5001 failed: Connection refused
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 5001 (#0)
> POST /api/v0/dag/put?format=dag-pb&input-enc=raw HTTP/1.1
> Host: localhost:5001
> User-Agent: curl/7.57.0
> Accept: */*
> Content-Length: 218
> Content-Type: multipart/form-data; boundary=------------------------5182c92cb3445de0
> 
< HTTP/1.1 200 OK
< Access-Control-Allow-Headers: X-Stream-Output, X-Chunked-Output, X-Content-Length
< Access-Control-Expose-Headers: X-Stream-Output, X-Chunked-Output, X-Content-Length
< Content-Type: application/json
< Server: go-ipfs/0.4.14-dev
< Trailer: X-Stream-Error
< Vary: Origin
< X-Chunked-Output: 1
< Date: Mon, 19 Feb 2018 23:15:00 GMT
< Transfer-Encoding: chunked
< 
{"Cid":{"/":"Qmd7xRhW5f29QuBFtqu3oSD27iVy35NRB91XFjmKFhtgMr"}}
* Connection #0 to host localhost left intact
```

```
$ curl --verbose -F 'data=@/home/vmx/src/pl/somedata.dag-pb' 'http://localhost:5001/api/v0/block/put?format=v0&mhtype=sha2-256&mhlen=-1'
*   Trying ::1...
* TCP_NODELAY set
* connect to ::1 port 5001 failed: Connection refused
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 5001 (#0)
> POST /api/v0/block/put?format=v0&mhtype=sha2-256&mhlen=-1 HTTP/1.1
[...]

PR #6331, opened 2019-05-14
block cmd: allow adding multiple blocks at once

magik6k:
This makes `block put` consistent with `dag put` (Addressing https://github.com/ipfs/go-ipfs/issues/4721)

This will be a minor breaking change for js, cc @vmx

### DIFF
--- a/core/commands/block.go
+++ b/core/commands/block.go
@@ -6,6 +6,8 @@ import (
 	"io"
 	"os"
 
+	files "github.com/ipfs/go-ipfs-files"
+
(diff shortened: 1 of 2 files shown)


Stebalien:
> This will be a minor breaking change for js, cc @vmx

cc @alanshaw.

We could _also_ have two cases: 1 return value and multiple return values. However, I'd rather do it this way.

Stebalien:
This doesn't appear to break interop. Merging.

PR #6333, opened 2019-05-14
github: turn issue template into a multiple-choice question

Stebalien:
This way:

1. Contributors have to click _something_ so they're more likely to read this.
2. Each issue gets correctly labeled.
3. We can customize the template for different types of issues so contributors don't have to read as much.


Stebalien:
This will give us a dialog like: https://help.github.com/en/articles/about-issue-and-pull-request-templates#issue-templates.

PR #6345, opened 2019-05-16
doc: remove out of date documentation

Stebalien:


Stebalien:
We need to do a code documentation pass eventually but this file is just confusing.

PR #6362, opened 2019-05-21
ci: actually enable golangci-lint

### LINKED ISSUE #6320: CI isn't failing for go tests that don't compile
See: https://github.com/ipfs/go-ipfs/pull/6319

Any ideas?

Stebalien:
also fixes #6320 (golangci-lint _will_ fail if something doesn't compile)

### DIFF
--- a/core/commands/swarm.go
+++ b/core/commands/swarm.go
@@ -22,9 +22,10 @@ import (
 	peer "github.com/libp2p/go-libp2p-peer"
 	pstore "github.com/libp2p/go-libp2p-peerstore"
 	swarm "github.com/libp2p/go-libp2p-swarm"
+	mafilter "github.com/libp2p/go-maddr-filter"
 	ma "github.com/multiformats/go-multiaddr"
 	madns "github.com/multiformats/go-multiaddr-dns"
-	mafilter "github.com/whyrusleeping/multiaddr-filter"
+	mamask "github.com/whyrusleeping/multiaddr-filter"
 )
 
 const (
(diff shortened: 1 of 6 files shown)


core/commands/swarm.go
```diff
@@ -599,8 +600,8 @@ Filters default to those specified under the "Swarm.AddrFilters" config key.
 		}
 
 		var output []string
-		for _, f := range swrm.Filters.Filters() {
-			s, err := mafilter.ConvertIPNet(f)
+		for _, f := range swrm.Filters.FiltersForAction(mafilter.ActionDeny) {
```
Stebalien: Moving away from deprecated APIs.

plugin/plugins/gen_main.sh
```diff
@@ -14,7 +14,7 @@ import (
 	$shortpkg "$pkg"
 )
 
-var Plugins = $shortpkg.Plugins
+var Plugins = $shortpkg.Plugins //nolint
```
Stebalien: Don't lint because this is "unused".

namesys/resolve/pathresolver_test.go
```diff
@@ -16,17 +16,17 @@ func TestResolveNoComponents(t *testing.T) {
 	}
 
 	_, err = resolve.Resolve(n.Context(), n.Namesys, n.Resolver, path.Path("/ipns/"))
-	if err != path.ErrNoComponents {
-		t.Fatal("Should error with no components (/ipns/).", err)
+	if err == nil {
```
magik6k: We probably want to check more than 'there was an error'

michaelavila: Probably not a big deal, but I'm curious, can we expect a specific error here and ultimately fail if we don't find the specific errors we're looking for?

michaelavila: It appears @magik6k and I reviewed at the same time. We are leaving the same feedback here and so I'm going to resolve this particular conversation so it's not in two threads.

Stebalien: But I'm lazy... fine.

PR #6373, opened 2019-05-23
fix race between adding tags and registering services with the connection manager

### LINKED ISSUE #6271: New 0.4.20 swarm connect "important" functionality doesn't work.
#### Version information:
go-ipfs version: 0.4.20-
Repo version: 7
System version: amd64/linux
Golang version: go1.11.5

#### Description:
In the newest release 0.4.20, swarm connection has a new feature stated here:

> The ipfs swarm connect command has a few new features:
> It now marks the newly created connection as "important". This should ensure
that the connection manager won't come along later and close the connection if
it doesn't think it's being used.

I ran ipfs swarm connect to manually connect two nodes of mine that are both running 0.4.20, and they disconnected minutes later. I've been able to test this multiple times.

Stebalien:
Hopefully helps fix #6271 and a bunch of other related issues.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -67,7 +67,7 @@ require (
 	github.com/libp2p/go-libp2p v0.0.28
 	github.com/libp2p/go-libp2p-autonat-svc v0.0.5
 	github.com/libp2p/go-libp2p-circuit v0.0.8
-	github.com/libp2p/go-libp2p-connmgr v0.0.5
+	github.com/libp2p/go-libp2p-connmgr v0.0.6
 	github.com/libp2p/go-libp2p-crypto v0.0.2
 	github.com/libp2p/go-libp2p-host v0.0.3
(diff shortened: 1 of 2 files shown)


makeworld-the-better-one:
Hi @Stebalien. Over [here](https://github.com/ipfs/go-ipfs/issues/6271#issuecomment-498857658) I was wondering about something and told I should come ask in this repo.

> How large is the window between both nodes running it? If one connects, and then another runs the command an hour later, will it still work? 5 minutes later? A day?

Stebalien:
The default grace period is 30 seconds. Basically, if no subsystem decides that the connection is useful within 30 seconds, we may kill it the next time the connection manager culls connections.

makeworld-the-better-one:
Is this the same thing is the GracePeriod value in config? Or is it unrelated.

Stebalien:
It's the same one.

PR #6378, opened 2019-05-25
fix shutdown data race and tie the process to the app

Stebalien:
This:

1. Ensures the process/app shutdown each-other.
2. Uses the process to handle context cancellation to deal with any
   thread-safety issues.

### DIFF
--- a/core/builder.go
+++ b/core/builder.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/ipfs/go-ipfs/core/bootstrap"
 	"github.com/ipfs/go-ipfs/core/node"
+	"github.com/jbenet/goprocess/context"
 )
 
 type BuildCfg = node.BuildCfg // Alias for compatibility until we properly refactor the constructor interface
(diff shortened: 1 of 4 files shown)


core/builder.go
```diff
@@ -27,24 +28,20 @@ func NewNode(ctx context.Context, cfg *BuildCfg) (*IpfsNode, error) {
...
 	n.IsOnline = cfg.Online
 	n.app = app
 
 	if app.Err() != nil {
 		return nil, app.Err()
 	}
 
+	// bind the process to the context so we stop when the context is
+	// canceled
+	goprocessctx.CloseAfterContext(n.Process, ctx)
```
Stebalien: We can also go back to just calling `app.Stop()` when the context is canceled if we don't want to introduce any goprocess stuff here. However, I'm not sure if `app.Stop()` is thread-safe (although I assume it is).

core/core.go
```diff
@@ -124,7 +120,7 @@ type Mounts struct {
 
 // Close calls Close() on the App object
 func (n *IpfsNode) Close() error {
-	return n.app.Stop(n.ctx)
+	return n.Process.Close()
```
Stebalien: Note: If we ever remove the "process" concept, we can just stick a `fx.Shutdowner` here.

Stebalien:
Ah, nevermind, this won't _wait_ for everything to shutdown.

Issue #6356, opened 2019-05-20
"failure on stop" error


Stebalien:
#### Version information:

```
go-ipfs version: 0.4.21-dev-b892db894
Repo version: 7
System version: amd64/linux
Golang version: go1.12.4
```

#### Description:

I got:

```
May 20 14:30:31 raft ipfs[6153]: Initializing daemon...
May 20 14:30:31 raft ipfs[6153]: go-ipfs version: 0.4.21-dev-b892db894
May 20 14:30:31 raft ipfs[6153]: Repo version: 7
May 20 14:30:31 raft ipfs[6153]: System version: amd64/linux
May 20 14:30:31 raft ipfs[6153]: Golang version: go1.12.4
May 20 14:30:37 raft ipfs[6153]: Swarm listening on /ip4/127.0.0.1/tcp/4101
May 20 14:30:37 raft ipfs[6153]: Swarm listening on /ip4/127.0.0.1/udp/4101/quic
May 20 14:30:37 raft ipfs[6153]: Swarm listening on /ip4/192.168.0.129/tcp/4101
May 20 14:30:37 raft ipfs[6153]: Swarm listening on /ip4/192.168.0.129/udp/4101/quic
May 20 14:30:37 raft ipfs[6153]: Swarm listening on /ip6/::1/tcp/4101
May 20 14:30:37 raft ipfs[6153]: Swarm listening on /ip6/::1/udp/4101/quic
May 20 14:30:37 raft ipfs[6153]: Swarm listening on /p2p-circuit
May 20 14:30:37 raft ipfs[6153]: Swarm announcing /ip4/127.0.0.1/tcp/4101
May 20 14:30:37 raft ipfs[6153]: Swarm announcing /ip4/127.0.0.1/udp/4101/quic
May 20 14:30:37 raft ipfs[6153]: Swarm announcing /ip4/192.168.0.129/tcp/4101
May 20 14:30:37 raft ipfs[6153]: Swarm announcing /ip4/192.168.0.129/udp/4101/quic
May 20 14:30:37 raft ipfs[6153]: Swarm announcing /ip6/::1/tcp/4101
May 20 14:30:37 raft ipfs[6153]: Swarm announcing /ip6/::1/udp/4101/quic
May 20 14:30:37 raft ipfs[6153]: API server listening on /ip4/127.0.0.1/tcp/5001
May 20 14:30:37 raft ipfs[6153]: WebUI: http://127.0.0.1:5001/webui
[...]

PR #6380, opened 2019-05-25
core: call app.Stop once

magik6k:
This might help with https://github.com/ipfs/go-ipfs/issues/6356 as calling `app.Stop` twice in parallel could cause a lot of weird races (It may not though).

(If it will help with anything, it will be the `mock repo` part of the issue, `context canceled` is caused by 'wrong' use of contexts in MFS)

### DIFF
--- a/core/builder.go
+++ b/core/builder.go
@@ -2,6 +2,7 @@ package core
 
 import (
 	"context"
+	"sync"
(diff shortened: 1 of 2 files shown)


Stebalien:
Thanks!

PR #6395, opened 2019-06-03
Remove Repo from routing fx provider parameter

Wondertan:
Routing provider receives Repo and does not use it. Probably this is kind of typo.


Stebalien:
Thanks!

Issue #4863, opened 2018-03-22
ipfs dht provide is not parallelized


kyledrake:
Enhancement:

The `ipfs dht provide CID ...` command is not currently parallelized. This means that if a larger list of records to provide is given, it runs pretty slow because it has to run the provides sequentially.

A better approach here would be to batch X number of provides at a time in parallel when a list of many cids is provided, so the end user doesn't have to parallelize on their end.

This is a lower priority improvement. Workaround is to do the parallelizing upstream, or use something like gnu parallel.

magik6k:
This will be really easy to do once https://github.com/ipfs/go-ipfs/pull/4333 is done

magik6k:
cc @michaelavila

michaelavila:
I'll take a look at this with the rest of the provider work. Thanks for pointing it out.

michaelavila:
Just an update:

After a brief discussion [here](https://github.com/ipfs/go-ipfs/pull/6396) it was decided to leave the `ipfs dht provide <hash>` command alone as its a useful debugging mechanism. We are also in the process of a building a more robust provider system that has facilities for responsibly providing many cids at once. The command you will likely want to use in the future is `ipfs provider provide <hash>`. Although this particular subcommand has not been built yet, it's not far off.

I'm going to leave this issue open until the suggested solution is available to use. Future issues filed about parallelizing `ipfs dht provide` should reference here.

guillaumemichel:
Implemented in https://github.com/ipfs/boxo/pull/907

PR #6396, opened 2019-06-03
[WIP] Parallelize dht provide commmand

michaelavila:
https://github.com/ipfs/go-ipfs/issues/4863

### DIFF
--- a/core/commands/dht.go
+++ b/core/commands/dht.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"sync"
 	"time"
 
 	cmdenv "github.com/ipfs/go-ipfs/core/commands/cmdenv"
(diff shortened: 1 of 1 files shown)


michaelavila:
@Stebalien yeah, makes sense. Sorry, I just wanted to get a PR up to see if this caused problems with the slew of tests as is on CI (because I'm having some trouble with the tests locally). I'll be sure to add some guardrails before this work is merge-ready.

Stebalien:
SGTM. Thanks for tackling this!

michaelavila:
Upon thinking about this further, I wonder if we should deprecate the `ipfs dht provide` command? I think we can offer `ipfs provider provide` for providing larger sets of cids. Doing that kind of coordination is really the job of the provider at that layer anyway.

One nice thing about the `ipfs dht provide` command is that it is synchronous, reporting error or success at the end. Maybe we can preserve this behavior through an option of some kind, e.g. `ipfs provider provide -s` which will work in some reasonable way that's yet to be defined.

cc @Stebalien

Stebalien:
I'd say:

1. Leave `dht provide` for now. It's a low-level plumbing command that's useful for debugging. We could still add a recursive option, but this command shouldn't touch the provider subsystem itself.
2. Add a `provider provide` command that adds a provider record to the provider subsystem. Or maybe even `provider add`, `provider rm`, etc? Or `provider provide` provider unprovide`. Naming is annoying...

michaelavila:
@Stebalien right, makes sense. I'm going to leave this `ipfs dht provide` command as is, for debugging purposes, and point people to `ipfs provider provide` (or whatever we come up with) for regular use. Closing this PR.

Issue #6042, opened 2019-03-03
Can pinning get a verbose argument?


markg85:
Hi,

I'm currently executing `ipfs pin add -r <QmHash>`. It's been running for hours and i have absolutely no clue if anything is happening at all!

I'm missing a "verbose" flag. I'm sure IPFS is doing all kinds of stuff right now that would be visible if i raised the logging level, but that might be too much information. I'd like to see logs targeted at the current pin action to have some clue of what's happening. Obviously only if i provide the hypothetical `-v` or `--verbose` flag.

Cheers,
Mark

djdv:
There is the `--progress` flag, which shows how many blocks have been processed. This will at least show you if things are progressing or not.
I would also like to see extended information though. Perhaps display the current block being processed, and maybe warn if there's 0 providers for it after some time.

michaelavila:
@markg85 @djdv What does the first iteration of this look like? I think if we can get a simple first version described, this would be a good introductory task for someone to take on. Can we start by simply listing out each block we are attempting to pin and then a confirmation that we've received and pinned that block? Something like:

```shell
$ ipfs pin add -v -r <QmHash>
...
pinning QmHash
pinned QmHash
```

Is that too simplistic of a first version?

djdv:
>this would be a good introductory task for someone to take on

:+1:

>Is this too simplistic of a first version?

I don't think so, any output seems better than none in this context.

In addition, I encounter people asking "if it's working" often enough, that I wonder if we shouldn't also change the default output. Making it noisy by default, and adding `-Q` and `-q` in the same way they appear in `ipfs add --help`.
At the very least it would improve command consistency.
But this doesn't have to necessarily be part of the first version.

markg85:
It turned out, in my case, that IPFS did absolutely nothing. Or rather, it did actually pin in 0.4.18 (the current when reporting this issue) and did nothing in 0.4.20 (today's current). Yeah, go figure that one out.

I wholeheartedly with @djdv. Whatever output is figured out, it's better then we have now.
This might be a good start imho:

> pinning QmHash
> pinned QmHash <in 1 minute and 20 seconds>

The time would just show how long it took to pin the file.
It could be verbose like this or simple like "01.20".

michaelavila:
@markg85 can you elaborate on what you mean by:

[...]

PR #6403, opened 2019-06-04
core/commands/pin: add verbose pin

reinerRubin:
Hello,
I have tried to add a verbose argument to the pinning command. It add output like this:
```
Pinning QmQZ1WVkddYPpRPm6vPpemhNacjwhupzSMzFEcYMisVpuY                                               
Pinning QmQeYwtLpXuEmvG9vWfRhoeyNgXYw5KExozDTHyp3vBQzc                                              
```
Currently this PR is more like a demo for the PR to go-merkledag [1].  

@michaelavila 

1 - https://github.com/ipfs/go-merkledag/pull/38
https://github.com/ipfs/go-ipfs/issues/6042

### DIFF
--- a/core/commands/pin.go
+++ b/core/commands/pin.go
@@ -43,13 +43,15 @@ type PinOutput struct {
 }
 
 type AddPinOutput struct {
-	Pins     []string
-	Progress int `json:",omitempty"`
+	Pins        []string
+	OngoingPins []string `json:",omitempty"`
+	Progress    int      `json:",omitempty"`
 }
 
 const (
-	pinRecursiveOptionName = "recursive"
-	pinProgressOptionName  = "progress"
+	pinRecursiveOptionName       = "recursive"
+	pinProgressOptionName        = "progress"
+	pinProgressVerboseOptionName = "verbose"
 )
 
 var addPinCmd = &cmds.Command{
@@ -64,6 +66,7 @@ var addPinCmd = &cmds.Command{
 	Options: []cmds.Option{
 		cmds.BoolOption(pinRecursiveOptionName, "r", "Recursively pin the object linked to by the specified object(s).").WithDefault(true),
 		cmds.BoolOption(pinProgressOptionName, "Show progress"),
+		cmds.BoolOption(pinProgressVerboseOptionName, "Show progress more detailed"),
 	},
 	Type: AddPinOutput{},
 	Run: func(req *cmds.Request, res cmds.ResponseEmitter, env cmds.Environment) error {
@@ -74,7 +77,9 @@ var addPinCmd = &cmds.Command{
 
 		// set recursive flag
 		recursive, _ := req.Options[pinRecursiveOptionName].(bool)
-		showProgress, _ := req.Options[pinProgressOptionName].(bool)
+		showProgressBasic, _ := req.Options[pinProgressOptionName].(bool)
+		showProgressVerbose, _ := req.Options[pinProgressVerboseOptionName].(bool)
+		showProgress := showProgressBasic || showProgressVerbose
 
 		if err := req.ParseBodyArgs(); err != nil {
 			return err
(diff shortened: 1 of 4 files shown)


core/commands/pin.go
```diff
@@ -43,13 +43,15 @@ type PinOutput struct {
 }
 
 type AddPinOutput struct {
-	Pins     []string
-	Progress int `json:",omitempty"`
+	Pins        []string
+	OngoingPins []string `json:",omitempty"`
```
reinerRubin: I was not sure about this json tags

michaelavila: I'm actually not sure either. I'll look around.

core/commands/pin.go
```diff
@@ -166,9 +189,18 @@ var addPinCmd = &cmds.Command{
...
-					fmt.Fprintf(os.Stderr, "Fetched/Processed %d nodes\r", out.Progress)
+
+				// this can only happen if the progress/verbose options are set
+				if inProgress := out.Pins == nil; inProgress {
+					if showVerboseProgress {
+						for _, pin := range out.OngoingPins {
+							fmt.Fprintf(os.Stderr,
+								"Pinning %s\n", pin)
+						}
+					} else if showDigitProgress {
```
michaelavila: Should we also include the numeric progress while in verbose mode?

I'm kind of wondering whether there should really be separate `verbose` and `progress` modes. Maybe this work you're doing should expand the meaning of the `progress` flag instead of introducing a new one? cc @magik6k/@Stebalien do either of you have an opinion on whether or not we expand the progress flag to show more infomation? vs introducing a new flag for showing more verbose progress?

michaelavila:
@reinerRubin mind getting the latest master merged into this work so `golint` will run? (I think that's the issue).

reinerRubin:
@michaelavila, I have done a rebase, but circleci shows that golint is ok:
```
ci/circleci: golint — Your tests passed on CircleCI!
 ```

**Comment by michaelavila** [comment]
@reinerRubin great! that's what I suspected, but because you didn't have the golint configs it wasn't running on CI

File: test/sharness/t0085-pins.sh [hunk 3]
```diff
@@ -160,6 +160,20 @@ test_pin_progress() {
...
+test_pin_verbose_progress() {
+  test_pin_dag_init
+
+  test_expect_success "'ipfs pin add --progress --verbose' file" '
+    ipfs pin add --progress --verbose $HASH 2> test_pin_verbose_progress_err
+  '
+
+  test_expect_success "pin progress reported correctly" '
+    cat test_pin_verbose_progress_err
+    grep -q "Pinning" test_pin_verbose_progress_err
```
michaelavila: Can we test for the whole line here? I think we have the hash available in `$HASH`, is that true?

```suggestion
    grep -q "Pinning $HASH" test_pin_verbose_progress_err
```

Another thing that would be nice to have is a test that shows progress for recursively pinning a hash that has a root block and other blocks underneath it. I think that test can be approached the same as the test you have here. 

What do you think?

michaelavila:
@reinerRubin something else to consider: in the existing progress reporting implementation we just print to `stderr`, but I wonder if some (or all) of this should be captured in the logs. Our goal is already to introduce this robust progress reporting to help with debugging, it seems like capturing it in the logs would make it useful for even more debugging.

I'm curious, what are your thoughts on that?

test/sharness/t0085-pins.sh
reinerRubin: Fixed with $HASH.

This test is kind of checks this but without explicit checks. Because the pin output is:
```
```

reinerRubin:
@michaelavila, could you explain? I am not familiar with the ipfs logging system. Should pinned cids be written to the logs and `stderr`? 

If so, I think, it should be logged at the go-merkledag level with or without "ProgressTracker", because usually logs level is a global setting.

PR #6406, opened 2019-06-04
add unixfs get metric

whyrusleeping:
This is a metric I really want to see, I don't know what labels we should add here, but I want to get this running on our gateway nodes ASAP.


whyrusleeping:
we should probably add tags to this, for things like "this is an ipns request" or "this is a path with N segments"

Stebalien:
> LGTM. In future we might consider creating a metrics-collecting layer of CoreApi (I don't know how expensive collecting metrics is, so keeping it to the gateway for now seems to make sense)

I started moving this logic there but we do some path resolution in the gateway before we call `Get`. But yeah, we should have metrics in both places.

PR #6412, opened 2019-06-05
build: strip GOPATH from build paths

Stebalien:
This should help fix some of our problems using plugins with go-ipfs.

### DIFF
--- a/mk/golang.mk
+++ b/mk/golang.mk
@@ -10,6 +10,9 @@ unexport GOFLAGS
 GOFLAGS ?=
 GOTFLAGS ?=
 
+# Try to make building as reproducible as possible by stripping the go path.
+GOFLAGS += -asmflags=all=-trimpath="$(GOPATH)" -gcflags=all=-trimpath="$(GOPATH)"
+
 ifeq ($(tarball-is),1)
 	GOFLAGS += -mod=vendor
 endif


Kubuxu:
How will that work with out of GOPATH builds?

Stebalien:
> How will that work with out of GOPATH builds?

It will still strip the paths from the go modules.

I'm trying to make it easier to build plugins for go-ipfs. Unfortunately, it looks like go barfs if the _paths_ differ (:anger:). My plan is to:

* Strip the paths when building plugins.
* Strip the paths when building go-ipfs with `make`. This _won't_ completely normalize the go-ipfs paths, but will normalize all the dependencies.
* Build the dists with go mod (https://github.com/ipfs/distributions/issues/240)

All together, this will mean:

1. If we're building against a local copy of go-ipfs by using a "replace" directive in the plugin's go.mod, all the paths will match (all stripped except maybe the go-ipfs one).
2. If we're building against a published version of go-ipfs, all the paths will match (all stripped).

We should also consider extracting the plugin interfaces so plugins don't have to depend on go-ipfs itself. That should make this even simpler.

Kubuxu:
We can make plugin interface a separate go module (with separate versioning) while keeping it in go-ipfs repo.

PR #6414, opened 2019-06-06
assets: move away from gx

Stebalien:
1. Use a submodule for dir-index-html. This isn't a go dependency and treating it like one is more trouble than it's worth. We don't actually need to checkout the submodule unless we need to regenerate the assets.
2. Avoid a runtime dependency on dir-index-html (may shave a few bytes off the final binary?).
3. Remove unused code.
4. Avoid bundling unused files.
5. Switch to a maintained version of go-bindata.
6. Use go mod to manage go-bindata.

Any objections to adding a git submodule?


Stebalien:
@whyrusleeping this adds a submodule to go-ipfs (only needed when re-generating this). Any objections?

assets/bindata_dep.go
```diff
@@ -0,0 +1,9 @@
+//+build never,!never
```
Kubuxu: This doesn't work (bindata is downloaded on every build).
I've used buildtag `tools` for that during the initial port.

Kubuxu:
No objections on using submodule on my side.

assets/bindata_dep.go
Stebalien: This just prevents us from importing it at _runtime_. IMO, we should fetch it when pulling in the dependencies (even if we don't need it right now).

Kubuxu: I don't see any benefit for fetching it on every build but I don't mind it.

Stebalien: It _should_ cache it. There's no real benefit to fetching it every build but, this way, we lock the version in. We're also doing this for our test dependencies and golangci-lint.

Stebalien:
@Kubuxu could I get a :heavy_check_mark:?

PR #6421, opened 2019-06-07
Extract provider module to `go-ipfs-provider`

michaelavila:
Moved `provider` module to https://github.com/ipfs/go-ipfs-provider


Kubuxu:
The module itself depends on go-ipfs which isn't great. It will build but from a functional standpoint, it won't work well.

michaelavila:
@Kubuxu yes, I'm still working on this, thanks for the heads up though. There's a dependence on `Pinner` that I think is the only real issue.

core/node/provider.go
```diff
@@ -115,3 +118,9 @@ func SimpleProviders(reprovideStrategy string, reprovideInterval string) fx.Opti
 		fx.Provide(SimpleReprovider(reproviderInterval)),
 	)
 }
+
+func pinnedProviderStrategy(onlyRoots bool) interface{} {
+	return func(pinner pin.Pinner, dag ipld.DAGService) simple.KeyChanFunc {
```
michaelavila: @magik6k @Kubuxu Any thoughts on a preferred alternative to this?

The reason for this is that the "old" reprovider depends on `pin.Pinner` which is a module within `go-ipfs`. To break that dependency I introduced a `Pinner` type in the `go-ipfs-provider` repository, but that is not the type of `Pinner` that's in the dependency injection container. To get around this I do the constructing of the dependency here, within `go-ipfs`, so that I can depend on `pin.Pinner`. The benefit, as I see it, is that this change is relatively small.

We could move the `Pinner` out as well into its own repository, but I really don't want to take on that large of a change right now.

hsanjuan: Would it be possible to leave the `NewPInnedProvider` thing as a provider implementation within go-ipfs and extract the more generalistic ones (namely `BlockstoreProvider`). This looks very weird right now, and I don't need the PinnedProvider anyways.

michaelavila: Offline @magik6k said he's ok with this. Ultimately, he has some work in progress in https://github.com/ipfs/go-ipfs/pull/6387 that will change how all of this works for the better.

michaelavila: @hsanjuan I think that makes sense. Let me try it. And now that I think about it, `ipfs-lite` maybe doesn't have the concept of a `Pinner`, is that true?

michaelavila:
@hsanjuan I have been looking at `ipfs-lite` trying to anticipate how you're going to integrate this code with it, and I have some questions:

* the provider system needs to be `Run()` and `Close()`, how will that work in `ipfs-lite`?
* Is `ipfs-lite` embedded in some other software?

core/node/provider.go
hsanjuan: Yeah that's true, no Pinner. Pinner is a go-ipfs specific concept (for now)

michaelavila:
Currently depends on https://github.com/ipfs/go-ipfs-provider/pull/11

Issue #6428, opened 2019-06-11
Pinning bugs & questions


dirkmc:
1. It seems like `PinAPI` takes a `blockstore.PinLock()` for `Add()` and `Update()` [but not Rm()](https://github.com/ipfs/go-ipfs/blob/master/core/coreapi/pin.go#L59).

2. In [PinAPI.Add()](https://github.com/ipfs/go-ipfs/blob/master/core/coreapi/pin.go#L18) we call `Pin()`, then `Provide()`, then `Flush()`. But if `Provide()` fails then we’re left in a state where a pin has been added in memory but not flushed. @michaelavila points out that `Provide()` adds a CID to a queue and it is actually provided asynchronously, so the current code probably makes sense but it might be worth adding a comment explaining this is the case.

3. In [pinner.Pin()](https://github.com/ipfs/go-ipfs/blob/master/pin/pin.go#L213), we
  - Take a lock before calling `dserv.Add()` (seems like it would be more efficient to take the lock *after* calling `dserv.Add()`)
  - call `dserv.Add()` and then call `dserv.Get()` (for a direct pin) or `mdag.FetchGraph()` (for a recursive pin). Why do we call `Get()` when we've just called `Add()`?

4. We don't flush after pinning in [block.Put()](https://github.com/ipfs/go-ipfs/blob/master/core/coreapi/block.go#L57)

Stebalien:
1. Yes. The pin lock just ensures we don't GC while adding a pin. It doesn't matter when removing a pin.
2. You're right. I wonder if we should _unpin_ in that case? It might be difficult to do that safely.
3. Good question. Could you submit a PR to fix these?
4. That's a bug. Mind submitting a PR?

dirkmc:
@stebalien with respect to point 1 above:
Removing a pin (and flushing the pin sets) will change the set of blocks in the blockstore (because the pin sets are stored in the blockstore). So are you sure we don’t need to take a lock?

dirkmc:
With respect to point 3 I found [this commit](https://github.com/ipfs/go-ipfs/commit/953b610470044bd332522b346e4881b8957c2050#diff-10c79301e027d970f7cba49f4a853d86) with message
> add node to dagserv under lock
  Otherwise, we could run a GC between adding and pinning.

I believe that comment is no longer accurate, because calls to Pin() should be protected by `blockstore.PinLock()`, does that sound right to you?

Stebalien:
[...]

PR #6441, opened 2019-06-12
Fix: flush pin on block put

dirkmc:
See https://github.com/ipfs/go-ipfs/issues/6428

### DIFF
--- a/core/commands/block.go
+++ b/core/commands/block.go
@@ -123,10 +123,10 @@ var blockPutCmd = &cmds.Command{
 		Tagline: "Store input as an IPFS block.",
 		ShortDescription: `
 'ipfs block put' is a plumbing command for storing raw IPFS blocks.
-It reads from stdin, and <key> is a base58 encoded multihash.
+It reads from stdin, and outputs a multihash to stdout.
 
-By default CIDv0 is going to be generated. Setting 'mhtype' to anything other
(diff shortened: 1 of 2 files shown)


core/commands/block.go
```diff
@@ -123,10 +123,10 @@ var blockPutCmd = &cmds.Command{
 		Tagline: "Store input as an IPFS block.",
 		ShortDescription: `
 'ipfs block put' is a plumbing command for storing raw IPFS blocks.
-It reads from stdin, and <key> is a base58 encoded multihash.
+It reads from stdin, and outputs a multihash to stdout.
```
Stebalien:
```suggestion
It reads from stdin, and outputs the block's CID to stdout.
```

core/commands/block.go
```diff
@@ -123,10 +123,10 @@ var blockPutCmd = &cmds.Command{
...
 
-By default CIDv0 is going to be generated. Setting 'mhtype' to anything other
-than 'sha2-256' or format to anything other than 'v0' will result in CIDv1.
+By default the multihash is output in CIDv0 format. Setting 'mhtype' to anything
```
Stebalien:
```suggestion
Unless specified, this command returns dag-pb CIDv0 CIDs. Setting 'mhtype' to anything
```

Stebalien:
Thanks!

PR #6443, opened 2019-06-12
Fix: dont double-fetch when adding a pin

dirkmc:
Also error-out early if trying to direct-pin a key that is already recursively pinned

See #6428

### DIFF
--- a/pin/pin.go
+++ b/pin/pin.go
@@ -213,11 +213,6 @@ func NewPinner(dstore ds.Datastore, serv, internal ipld.DAGService) Pinner {
 func (p *pinner) Pin(ctx context.Context, node ipld.Node, recurse bool) error {
 	p.lock.Lock()
 	defer p.lock.Unlock()
-	err := p.dserv.Add(ctx, node)
-	if err != nil {
-		return err
-	}
-
 	c := node.Cid()
 
 	if recurse {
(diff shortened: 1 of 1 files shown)


pin/pin.go
```diff
@@ -213,11 +213,6 @@ func NewPinner(dstore ds.Datastore, serv, internal ipld.DAGService) Pinner {
 func (p *pinner) Pin(ctx context.Context, node ipld.Node, recurse bool) error {
 	p.lock.Lock()
 	defer p.lock.Unlock()
-	err := p.dserv.Add(ctx, node)
```
Stebalien: I believe we still need to _add_ this node, we can probably just do it outside the lock.

dirkmc: I basically reverted [this commit](https://github.com/ipfs/go-ipfs/commit/953b610470044bd332522b346e4881b8957c2050#diff-10c79301e027d970f7cba49f4a853d86), ie I'm assuming that we have already added before `Pin()` is called.
Is that incorrect?

Stebalien: I really don't know. The safe solution would be to change the API. Let's pause this for a few days while I try to pull in a CoreAPI patch, then we can look into making `Pin` take a CID.

gammazero: This is done outside the lock now:
https://github.com/ipfs/go-ipfs-pinner/blob/master/dspinner/pin.go#L177

pin/pin.go
```diff
@@ -246,6 +241,10 @@ func (p *pinner) Pin(ctx context.Context, node ipld.Node, recurse bool) error {
 
 		p.recursePin.Add(c)
 	} else {
+		if p.recursePin.Has(c) {
+			return fmt.Errorf("%s already pinned recursively", c.String())
+		}
+
```
gammazero, line 247: This is also done in the current pinner:
https://github.com/ipfs/go-ipfs-pinner/blob/master/dspinner/pin.go#L246-L247

gammazero:
Closing: These changes are already in the current pinner.

PR #6445, opened 2019-06-12
dep: update go-bitswap

### LINKED ISSUE #6442: Slow transfer on localhost
# go-ipfs version
ipfs version 0.4.21

I have deployed a small private swarm of IPFS nodes in Kubernetes on a local network (also tried local host with minikube), and ```ipfs get``` is really slow. My CPU should be able to handle this small load, and transfer using scp between "nodes" is lightning fast.

# Transfer results
- 162MB file: 46s (average of multiple tries)
- 325MB file: 2m41s (also averaged)

- Initial ```ipfs add``` results in 3~4s of hashing.
- ```ipfs get```-ing on the node that added the file results in 1s "download"

# Troubleshooting and ideas
I have tried using ```--routing=none``` and ```--routing=dhtclient``` to no avail.

The gap in transfer time between the two file sizes may indicate it comes from hashing I suppose (the merkle tree must be deeper), but the CPU is not at full capacity.

Any idea what may cause this?

Stebalien:
fixes #6442

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,7 +17,7 @@ require (
 	github.com/golangci/golangci-lint v1.16.1-0.20190425135923-692dacb773b7
 	github.com/hashicorp/go-multierror v1.0.0
 	github.com/hashicorp/golang-lru v0.5.1
-	github.com/ipfs/go-bitswap v0.1.1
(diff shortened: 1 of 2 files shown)


Stebalien:
(dep update, underlying PRs approved)

PR #6446, opened 2019-06-13
ci: benchmark

Stebalien:
1. Benchmark the release branch, the master branch, and the current branch.
2. Compare these.
3. Fail if we have a >5% slowdown.


Stebalien:
There's a memory leak on shutdown, somewhere. We aren't properly closing all streams.

PR #6448, opened 2019-06-13
Allow resolution of .eth names via .eth.link

mcdee:
As per https://github.com/protocol/collab-ens/issues/5 this PR handles the situation where IPNS names ending in `.eth` are unresolvable due to `.eth` not being a recognised TLD.

It does so by changing such names to end in `.eth.link`, which will be resolvable by the EthDNS nameserver to obtain data from the relevant ENS records on-chain in Ethereum.

### DIFF
--- a/namesys/namesys.go
+++ b/namesys/namesys.go
@@ -80,12 +80,20 @@ func (ns *mpns) ResolveAsync(ctx context.Context, name string, options ...opts.R
 	return resolveAsync(ctx, ns, name, opts.ProcessOpts(options))
 }
 
+const ethTLD = ".eth"
+const linkTLD = ".link"
+
 // resolveOnce implements resolver.
 func (ns *mpns) resolveOnceAsync(ctx context.Context, name string, options opts.ResolveOpts) <-chan onceResult {
 	out := make(chan onceResult, 1)
 
 	if !strings.HasPrefix(name, ipnsPrefix) {
 		name = ipnsPrefix + name
+	}
+	if strings.HasSuffix(name, ethTLD) {
+		// This is an ENS name.  As we're resolving via an arbitrary DNS server
+		// that may not know about .eth we need to add our link domain suffix.
+		name = name + linkTLD
 	}
 	segments := strings.SplitN(name, "/", 4)
 	if len(segments) < 3 || segments[0] != "" {
--- a/namesys/namesys_test.go
+++ b/namesys/namesys_test.go
@@ -51,6 +51,7 @@ func mockResolverOne() *mockResolver {
 			"QmatmE9msSfkKxoffpHwNLNKgwZG8eT9Bud6YoPab52vpy": "/ipfs/Qmcqtw8FfrVSBaRmbWwHxt3AuySBhJLcvmFYi3Lbc4xnwj",
 			"QmbCMUZw6JFeZ7Wp9jkzbye3Fzp2GGcPgC3nmeUjfVF87n": "/ipns/QmatmE9msSfkKxoffpHwNLNKgwZG8eT9Bud6YoPab52vpy",
 			"QmY3hE8xgFCjGcz6PHgnvJz5HZi1BaKRfPkn1ghZUcYMjD": "/ipns/ipfs.io",
+			"QmQ4QZh8nrsczdUEwTyfBope4THUhqxqc1fx6qYhhzZQei": "/ipfs/QmP3ouCnU8NNLsW6261pAx2pNLV2E4dQoisB1sgda12Act",
 		},
 	}
 }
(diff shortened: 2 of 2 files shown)


mcdee:
> * What if we use `/ens` instead of `/ipns`

Personally I think the idea of adding the resolution protocol to the path is overly proscriptive.  It would also, in my mind, slow the transition to using ENS.

If paths are out there that already use `/ipns/my.domain.com`, for example, and `my.domain.com` moves to ENS, resolution will continue over DNS (successfully) even though ENS would be the primary source of the data and a preferable path to resolution.

There is an argument this could cause problems if DNS and ENS diverge, but this can be resolved by explicitly defining primacy (*e.g.* records in ENS will always override those in DNS).

lidel:
@mcdee I agree that discussion should not block us from enabling `/ipns/example.eth` as soon as possible to unlock use of `.eth` on gateways.

I also like the idea of overriding DNS servers via config of go-ipfs  (eg. array of hostnames at `ipfs config Dns.NameServers` or [DoH](https://wiki.mozilla.org/Trusted_Recursive_Resolver) endpoint at `Dns.ResolverUrl`). That would enable people to switch their Gateways to ENS for all TLDs if they choose to do so, while maintaining compatibility within existing convention. 

Sidenote:
>  Note that DNSLink is already confusing: `/ipns/{fqdn}`  has no relation to actual IPNS (which operates on `/ipns/{libp2p-key}`), it is actually something like `/dns/{fqdn}` (https://github.com/ipfs/notes/issues/348). It is a technical/ux debt we did not deal with yet. We are revisiting "what should be a separate namespace and why" in  https://github.com/ipfs/specs/issues/198


@Stebalien  would it be acceptable to merge this PR as a temporary measure?

Rationale: `.eth` does not exist in public DNS, so it is safe to do this for now. If `.eth` becomes a real TLD in the future, then we will add configuration option to disable this rewrite, or move to `/ens` by then.

Stebalien:
Longer comment:

> It may cause problem in the future if .eth becomes a real TLD owned by Ethiopia.

[`.et`](https://en.wikipedia.org/wiki/.et) is the TLD for Ethiopia. Has anyone submitted a RFC with the IETF for .ETH?

> Prior discussions we had ended with consensus that resolvers other than "real" DNS should get own namespaces (cc @Stebalien mentioning ENS in ipfs/notes#302 (comment))

I agree that the DNS namespace shouldn't necessarily be tied to the DNS protocol.

> There is an argument this could cause problems if DNS and ENS diverge, but this can be resolved by explicitly defining primacy (e.g. records in ENS will always override those in DNS).

We can make that decision when we get there.

aschmahmann:
Wanted to surface here the comment by @Stebalien at https://github.com/multiformats/multicodec/pull/136#pullrequestreview-257769901 regarding how the IPNS multicodec should not be used for things other than IPNS.

In general, people should also not expect that the syntax `/ipns/something-other-than-an-IPNS-key` will be supported indefinitely, it's super confusing.

Stebalien:
> In general, people should also not expect that the syntax /ipns/something-other-than-an-IPNS-key will be supported indefinitely, it's super confusing.

Note: We're going to support that indefinitely as we don't break links. However, there's no reason to introduce the same confusion when encoding them into a compact binary format.

mcdee:
@Stebalien added suitable tests.

Regarding the `.eth` domain: ICANN have put all of the 3-character ccTLDs in limbo for the moment, with the idea that countries may want them one day (ignoring the fact that they all have 2-character ccTLDs and some of them are already used, but hey) so `.eth` is unavailable to us.

Stebalien:
> Regarding the .eth domain: ICANN have put all of the 3-character ccTLDs in limbo for the moment, with the idea that countries may want them one day (ignoring the fact that they all have 2-character ccTLDs and some of them are already used, but hey) so .eth is unavailable to us.

Have you talked to the IETF? That's how `.onion` did it.

Stebalien:
@aschmahmann would you like to continue our discussion here?

Stebalien:
@lidel thanks for writing up the decision.

lidel:
@Stebalien just for the record: will this be released in v0.5.0?

v0.4.22:

```console
$ curl http://127.0.0.1:8080/ipns/ipfs.enstest.eth/
ipfs resolve -r /ipns/ipfs.enstest.eth/: not a valid proquint string

$ ipfs name resolve /ipns/ipfs.enstest.eth
Error: not a valid proquint string
```

master:

```console
$ ipfs name resolve /ipns/ipfs.enstest.eth.link
/ipfs/QmRAQB6YaCyidP37UdDnjFY5vQuiBrcqdyoW1CuDgwxkD4

$ ipfs name resolve /ipns/ipfs.enstest.eth
/ipfs/QmRAQB6YaCyidP37UdDnjFY5vQuiBrcqdyoW1CuDgwxkD4
```

Stebalien:
Yes.

parkan:
hmm am I right in thinking that the sync Resolve code path here doesn’t hit this change? 

https://github.com/ipfs/go-ipfs/blob/b15edf287df6f7d8d99f7b7befb548469579a20c/namesys/namesys.go#L54

PR #6450, opened 2019-06-13
gitmodules: use https so we don't need an ssh key

Stebalien:


Stebalien:
(merging as this fixes a CI issue)

PR #6459, opened 2019-06-21
Document the AddrFilters option

hsanjuan:
Shouldn't have to scan through an old issue to figure out what is happening
and why this is super important.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -342,10 +342,18 @@ Tells reprovider what should be announced. Valid strategies are:
 
 Options for configuring the swarm.
 
-- `AddrFilters`
-An array of address filters (multiaddr netmasks) to filter dials to.
-See [this issue](https://github.com/ipfs/go-ipfs/issues/1226#issuecomment-120494604) for more
-information.
+- `AddrFilters` An array of address filters (multiaddr netmasks) to filter
+automatic dials to. The ipfs daemon will otherwise attempt to randomly dial
+the default listen port on every IP of the the local network IP range. This
+improves autodiscovery of other daemons in the same network, but may also
+trigger netscan alerts on some hosting providers or cause strain in some
+setups.
+
+The `server` configuration profile fills up this list with sensible defaults,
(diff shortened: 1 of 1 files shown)


docs/config.md
```diff
@@ -342,10 +342,18 @@ Tells reprovider what should be announced. Valid strategies are:
...
-- `AddrFilters`
-An array of address filters (multiaddr netmasks) to filter dials to.
-See [this issue](https://github.com/ipfs/go-ipfs/issues/1226#issuecomment-120494604) for more
-information.
+- `AddrFilters` An array of address filters (multiaddr netmasks) to filter
+automatic dials to. The ipfs daemon will otherwise attempt to randomly dial
+the default listen port on every IP of the the local network IP range. This
+improves autodiscovery of other daemons in the same network, but may also
+trigger netscan alerts on some hosting providers or cause strain in some
+setups.
```
Stebalien: This should explain _why_ we do that and how. That is, we're not intentionally dialing every address one-by-one.

(Otherwise, we'll get even more questions)

docs/config.md
```diff
@@ -342,10 +342,18 @@ Tells reprovider what should be announced. Valid strategies are:
...
+
+The `server` configuration profile fills up this list with sensible defaults,
+preventing scans to most common local network (`10.0.0.0/8` or
+`192.168.0.0/16`), but you should always check settings against your own
+network and/or hosting provider.
```
Stebalien: It filters _all_ non-routable addresses per the IANA special registry.

docs/config.md
hsanjuan: Isn't the *why* to improve autodiscovery? I don't have much idea about the *how*.. is it random?

Stebalien: No. We end up dialing local addresses because peers publish these addresses in the DHT. We aren't intentionally scanning the local network.

hsanjuan: Ah, will update accordingly. Thanks!

hsanjuan:
hey, thanks

PR #6461, opened 2019-06-21
Update go-ds-badger

Kubuxu:
License: MIT


Kubuxu:
@Stebalien badger team removed rc/branch we were depending upon.

Kubuxu:
I had to update go-libp2p-peerstore to use also new version because version numbers of badger we completely messed up but this pulled in go-libp2p-core@v0.0.4 which is WIP. That is why we have a replace directive.

Kubuxu:
Resolves: https://github.com/ipfs/go-ipfs/pull/6461

PR #6465, opened 2019-06-24
Include the git blob id of the dir-index bundle in the ETag

mib-kd743naq:
While the content of raw files retrieved via the gateway should never
change, the look and feel of the directory index can and will change
between versions of go-ipfs.

Incorporate the hash of assets/bindata.go into the ETag when appropriate

License: MIT
Signed-off-by: Mib Kd743naq <mib.kd743naq@gmail.com>

### DIFF
--- a/Rules.mk
+++ b/Rules.mk
@@ -54,6 +54,8 @@ include $(dir)/Rules.mk
 dir := filestore/pb
 include $(dir)/Rules.mk
 
+dir := assets
+include $(dir)/Rules.mk
 
 # -------------------- #
 #   universal rules    #
--- a/assets/Rules.mk
+++ b/assets/Rules.mk
@@ -1,0 +1,1 @@
+assets-bindata-blob:=$(shell git hash-object $(dir)/bindata.go)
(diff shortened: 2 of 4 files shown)


mib-kd743naq:
Forgot to add a visual of what this does.

Before:
```
~$ curl -I 127.0.0.1:8080/ipfs/QmS4ustL54uo8FzR9455qaxZwuMiUhyvMcX9Ba8nUH4uVv
HTTP/1.1 200 OK
...
Etag: "QmS4ustL54uo8FzR9455qaxZwuMiUhyvMcX9Ba8nUH4uVv"
```

After:
```
~$ curl -I 127.0.0.1:8080/ipfs/QmS4ustL54uo8FzR9455qaxZwuMiUhyvMcX9Ba8nUH4uVv
HTTP/1.1 200 OK
...
Etag: "DirIndex-ad00a51de6a0a134329b35f56612d3adfb3972d9_CID-QmS4ustL54uo8FzR9455qaxZwuMiUhyvMcX9Ba8nUH4uVv"
```

mib-kd743naq:
> even our binary distribution system manually builds with go build.

Whoa! I didn't know this :) The `README` implies `make` is the one and only way. I like option 2 as well, will close the PR for now, will reopen when I have a chance to red.

Stebalien:
Well, `make` is the _correct_ way but I think this is a case of "we know better so we're going to ignore our own best practice".

PR #6472, opened 2019-06-27
Provide roots only strategy (rough)

michaelavila:
Requires: https://github.com/ipfs/go-ipfs-config/pull/38

This is a quick and straightforward change to **(1)** disable bitswap providing, and **(2)** enable providing of roots in the new providing system. This does not make the codebase better and _will_ be factored differently in time.

The following configs are needed:

```sh
ipfs config Experimental.StrategicProviding true
ipfs config Provider.Strategy "roots"
# restart ipfs
```

Here "roots" is defined by the root hashes created during `ipfs add` and `ipfs pin add` commands.

### DIFF
--- a/core/node/groups.go
+++ b/core/node/groups.go
@@ -235,7 +235,7 @@ func Online(bcfg *BuildCfg, cfg *config.Config) fx.Option {
 		fx.Provide(p2p.New),
 
 		LibP2P(bcfg, cfg),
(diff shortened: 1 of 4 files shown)


michaelavila:
I'm closing this as it was just an option for the recent dht push. Ultimately, provide roots will not use the old reprovider system as is the case in this PR. Thanks for taking a look.

PR #6474, opened 2019-06-28
feat: make it easier to load custom plugins

Stebalien:
1. Allow loading from arbitrary and multiple directories.
2. Make it possible to directly load built-in plugins that _aren't_ in the preload list.
3. Check the plugin loading state transitions.

### DIFF
--- a/cmd/ipfs/main.go
+++ b/cmd/ipfs/main.go
@@ -49,18 +49,20 @@ const (
 func loadPlugins(repoPath string) (*loader.PluginLoader, error) {
 	pluginpath := filepath.Join(repoPath, "plugins")
 
+	plugins, err := loader.NewPluginLoader()
+	if err != nil {
+		return nil, fmt.Errorf("error loading preloaded plugins: %s", err)
+	}
+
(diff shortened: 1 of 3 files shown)


Stebalien:
Inspired by https://github.com/ipfs/go-ipfs-config/pull/5#issuecomment-503213142. Loading custom plugins is currently a bit annoying.

go.mod
```diff
@@ -112,6 +112,7 @@ require (
 	go4.org v0.0.0-20190313082347-94abd6928b1d // indirect
 	golang.org/x/sync v0.0.0-20190423024810-112230192c58 // indirect
 	golang.org/x/sys v0.0.0-20190626221950-04f50cda93cb
+	google.golang.org/appengine v1.4.0 // indirect
```
Kubuxu: I have no idea why this dep started showing up but it is a heisen-dep, it shows up on build and disappears with tidy.

Stebalien: Yeah, it's a bug in go and I don't feel like reporting it.

Kubuxu: Same.

PR #6476, opened 2019-06-28
Update go.sum

michaelavila:
When I run `make install` on master, I have this one change to `go.sum`.


Kubuxu:
It also disappears when you run `go mod tidy`. I know it is annoying as hell.

Stebalien:
Yeah, we need to file an upstream bug report.

PR #6478, opened 2019-06-29
ignore stale API files and deprecate ipfs repo fsck

Stebalien:
1. Users now never need to run `ipfs repo fsck` (and, if they do, it'll do nothing).
2. Stale "api" files are ignored if the daemon isn't running and the repo is initialized.

This means the following finally works reliably:

```
ipfs id
ipfs daemon &
sleep 2
killall -9 ipfs
sleep 2
ipfs id
```


cmd/ipfs/main.go
```diff
@@ -195,21 +194,90 @@ func checkDebug(req *cmds.Request) {
...
 	}
 
-	var exctr cmds.Executor
-	if client != nil && !req.Command.External {
-		exctr = client.(cmds.Executor)
-	} else {
-		exctr = cmds.NewExecutor(req.Root)
+	// Force the daemon when the API flag is passed (unless we're trying to
+	// _run_ the daemon).
+	daemonForced := apiAddr != nil && req.Command != daemonCmd
```
djdv: pedantic nits:
>Force the daemon when the API flag is passed 

Should probably be more specific here, forcing the daemon to what?
Initialize temporarily?

 `daemonForced` -> `daemonRequested`
Despite a user supplying an API addr, commands don't aways force the daemon to do anything.
It's requested, and can be denied if the request is incompatible.

Stebalien: Thanks, updated. I went back and forth on that several times but decided that e.g., `ipfs version` should just be run on the client regardless.

Stebalien:
> In the future, (if we can) it may be good to create the API file in the same way we do for the repo lockfile. (have the OS treat it as a temporary file that is removed when the process is no longer running).
But this PR is still necessary for platforms that don't support such a feature, and can contain stale API files.

At this point, we assume that all platforms support it (in practice, all do). The issue is that the daemon may be running on a separate machine.

Issue #6467, opened 2019-06-25
Could not build arguments for function "reflect".makeFuncStub


NatoBoram:
#### Version information:

```yaml
OS: Windows 10 Pro Insider Preview x86_64
Resolution: 1920x1080
DE: Aero
WM: Explorer
CPU: Intel i5-3570K (4) @ 3.468GHz
GPU: NVIDIA GeForce GTX 660 Ti
Memory: 8135MiB
```

```yaml
go-ipfs version: 0.4.22-dev-
Repo version: 7
System version: amd64/windows
Golang version: go1.12.6
```

#### Description:

When running `ipfs daemon`, I get the following error : 

> Initializing daemon...
> go-ipfs version: 0.4.22-dev-
> Repo version: 7
> System version: amd64/windows
> Golang version: go1.12.6
> 23:18:09.434 ERROR   cmd/ipfs: error from node construction:  could not build arguments for function "reflect".makeFuncStub (C:/ProgramData/scoop/apps/go/current/src/reflect/asm_amd64.s:12): failed to build provider.System: could not build arguments for function "github.com/ipfs/go-ipfs/core/node".SimpleProviderSys.func1 (C:/Users/natob/Go/src/github.com/ipfs/go-ipfs/core/node/provider.go:40): failed to build provider.Provider: could not build arguments for function "github.com/ipfs/go-ipfs/core/node".SimpleProvider (C:/Users/natob/Go/src/github.com/ipfs/go-ipfs/core/node/provider.go:27): failed to build *queue.Queue: function "github.com/ipfs/go-ipfs/core/node".ProviderQueue (C:/Users/natob/Go/src/github.com/ipfs/go-ipfs/core/node/provider.go:22) returned a non-nil error: strconv.ParseUint: parsing "": invalid syntax daemon.go:337
> 
[...]

Issue #6260, opened 2019-04-26
Lean on datastore queries in provider queue


Stebalien:
At the moment, we read in the head/tail of the provider queue on start. However, it's probably easier to just repeatedly query the datastore.

That is,

1. Store provider queue entries as `/provider/queue/$nanosecond-timestamp/$cid` -> (or even just `/provider/queue/$start-timestamp.$counter`).
2. Each time we're signaled (size-one channel?) that something was enqueued, query `Query{Prefix: "/provider/queue/", Order: [OrderByKey]}`.

Motivation: track less state. This is really low-priority, it just may simplify this queue a bit and remove room for bugs.

michaelavila:
Noted :+1:. I'll look into this as I'm working on the remainder of the provider PRs. Thanks for bringing it up.

michaelavila:
Closing as this has been implemented. Thanks for the suggestion!

PR #6480, opened 2019-07-01
Fix node construction queue error

michaelavila:
Fixes https://github.com/ipfs/go-ipfs/issues/6467, by implementing a very simple version of https://github.com/ipfs/go-ipfs/issues/6260.

### DIFF
--- a/provider/queue/queue.go
+++ b/provider/queue/queue.go
@@ -3,8 +3,7 @@ package queue
 import (
 	"context"
 	"fmt"
-	"strconv"
-	"strings"
+	"time"
 
 	cid "github.com/ipfs/go-cid"
 	datastore "github.com/ipfs/go-datastore"
@@ -25,8 +24,6 @@ type Queue struct {
 	// e.g. provider vs reprovider
 	name    string
 	ctx     context.Context
-	tail    uint64
-	head    uint64
 	ds      datastore.Datastore // Must be threadsafe
 	dequeue chan cid.Cid
 	enqueue chan cid.Cid
@@ -37,16 +34,10 @@ type Queue struct {
 // NewQueue creates a queue for cids
 func NewQueue(ctx context.Context, name string, ds datastore.Datastore) (*Queue, error) {
 	namespaced := namespace.Wrap(ds, datastore.NewKey("/"+name+"/queue/"))
-	head, tail, err := getQueueHeadTail(ctx, namespaced)
-	if err != nil {
-		return nil, err
-	}
 	cancelCtx, cancel := context.WithCancel(ctx)
 	q := &Queue{
 		name:    name,
 		ctx:     cancelCtx,
-		head:    head,
-		tail:    tail,
 		ds:      namespaced,
 		dequeue: make(chan cid.Cid),
 		enqueue: make(chan cid.Cid),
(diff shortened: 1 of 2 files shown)


provider/queue/queue.go
```diff
@@ -124,7 +80,26 @@ func (q *Queue) work() {
 
 		for {
 			if c == cid.Undef {
-				k, c = q.nextEntry()
+				head, e := q.getQueueHead()
+
+				if e != nil {
+					log.Errorf("error querying for head of queue: %s, stopping provider", e)
+					return
```
michaelavila: If we get an error querying for head, then the provider will not work so I stop it. Bad idea?

provider/queue/queue.go
```diff
@@ -124,7 +80,26 @@ func (q *Queue) work() {
...
+				} else if head != nil {
+					k = datastore.NewKey(head.Key)
+					c, e = cid.Parse(head.Value)
+					if e != nil {
+						log.Warningf("error parsing queue entry cid with key (%s), removing it from queue: %s", head.Key, e)
+						err := q.ds.Delete(k)
+						if err != nil {
+							log.Errorf("error deleting queue entry with key (%s), due to error (%s), stopping provider", head.Key, err)
+							return
```
michaelavila: If we get an error deleting the bad head entry, then the queue will not work so I stop it. Bad idea?

michaelavila:
There might be a better factoring for this code. I decided to defer those decisions as I'm currently in the process of extracting the provider from go-ipfs and will need to make this same change in the extracted provider anyway. More importantly, this lets me get a fix for a high priority bug in sooner rather than later.

provider/queue/queue.go
michaelavila: The other option is to keep attempting the query while respecting some backoff mechanism. Is that worth the extra effort?

Stebalien: Let's leave it like this for now. We _shouldn't_ get an error.

provider/queue/queue.go
```diff
@@ -151,61 +125,24 @@ func (q *Queue) work() {
...
-}
-
-func getQueueHead(ds datastore.Datastore) (uint64, error) {
-	return getFirstIDByOrder(ds, query.OrderByKey{})
-}
-
-func getQueueTail(ds datastore.Datastore) (uint64, error) {
-	tail, err := getFirstIDByOrder(ds, query.OrderByKeyDescending{})
+func (q *Queue) getQueueHead() (*query.Result, error) {
+	qry := query.Query{Orders: []query.Order{query.OrderByKey{}}}
```
Stebalien: Limit: 1

provider/queue/queue.go
```diff
@@ -135,14 +110,13 @@ func (q *Queue) work() {
 
 			select {
 			case toQueue := <-q.enqueue:
-				nextKey := q.queueKey(q.tail)
+				keyPath := fmt.Sprintf("%d/%s", time.Now().UnixNano(), c.String())
```
Stebalien: Nit: let's use base32 (`c.StringOfBase` or `c.Encode`). We try to use case-insensitive keys in the datastore.

provider/queue/queue.go
```diff
@@ -124,7 +80,26 @@ func (q *Queue) work() {
 
 		for {
 			if c == cid.Undef {
-				k, c = q.nextEntry()
+				head, e := q.getQueueHead()
```
Stebalien: Ideally, we'd:

1. Open a query.
2. Iterate over the query to the end.
3. Open a new query when we get the next CID.

(but we can punt on that)

provider/queue/queue.go
michaelavila: Good catch, that was dumb of me.

provider/queue/queue.go
michaelavila: Understood, I did not realize. I'll make this change now.

provider/queue/queue_test.go
```diff
@@ -2,22 +2,28 @@ package queue
...
 	for i := 0; i < n; i++ {
-		c := blockGenerator.Next().Cid()
-		cids = append(cids, c)
+		data := make([]byte, rand.Intn(256))
+		rand.Read(data)
+		b, err := blocks.NewBlockWithCid(data, cid.NewCidV1(cid.Raw, util.Hash(data)))
+		if err != nil {
+			fmt.Println("This should not have happened.")
+		}
+		cids = append(cids, b.Cid())
```
michaelavila: @Stebalien this change was required because the block generater creates blocks with v1 cids and base32 encoding a v1 cid fails (https://github.com/ipfs/go-cid/blob/master/cid.go#L388). This begs the question, is it ok to do this encoding if it will fail for v0 cids? It's unclear to me whether or not this code might run into a v0 cid.

provider/queue/queue.go
michaelavila: I got you. I'll keep that in mind.

provider/queue/queue_test.go
michaelavila: Maybe we `Base58BCT` if v0 and `Base32` if v1?

Stebalien: Ah. Damn, I forgot about that. Case-sensitive CIDs _should_ be fine (and we can always change this later without any issues as we're just using this for uniqueness).

michaelavila:
@Stebalien this should be good to go.

PR #6486, opened 2019-07-08
document the debug script

Stebalien:


Kubuxu:
The `go tool pprof` calls there use `-symbolize=remote` so the binary is not needed for pprof analisys but it still might be useful.

Issue #6262, opened 2019-04-26
Separate identity values from configuration file


lanzafame:
#### Version information:
go-ipfs version: 0.4.20-
Repo version: 7
System version: amd64/linux
Golang version: go1.12.4

#### Type:
Enhancement

#### Description:
The `Identity` section of the current configuration file should be moved to a separate file.

```json
{
    "Identity": {
        "PeerID": <peer_id>,
        "PrivKey": <priv_key>
    },
    ...
}
```

The identity section prevents the simple configuration of ipfs nodes in cloud-native environments, i.e Kubernetes. Fundamentally, I want to be able to have an ipfs config.json in Consul/k8s ConfigMap and have all ipfs nodes spun up to use that, injected via an env var and using `ipfs daemon --init` flag to generate a key pair if one doesn't exist within `IPFS_PATH`. Currently, I have to run a k8s init container, which starts up before every ipfs container, to run `ipfs init` and then make the config changes, which are hardcoded in the script because there are no env vars for config values either. See [here](https://github.com/lanzafame/ipfs-cluster-k8s/blob/master/ipfs-cluster-base/cluster-setup-confmap.yaml#L46-L52) and [here](https://github.com/lanzafame/ipfs-cluster-k8s/blob/master/ipfs-cluster-base/cluster-statefulset.yaml#L16-L19) for code of what I just explained. 

Related: #744, #783.

[This](https://github.com/ipfs/go-ipfs/issues/744#issuecomment-168354625) obviously didn't happen or it did and the separation never occurred.

@whyrusleeping how should this move forward? To clearly state my agenda, I am not interested in 'where' the identity information goes per se as long as it is no longer in the configuration file. 

Happy to make this change myself, I would just need some assistance in finding where to make this change, I am guessing `go-ipfs-config` but I am not sure about the discussion in #744 about keys being stored in the blockstore or #783 mentioning them going in the repo.

/cc @Stebalien

Stebalien:
We absolutely should split identity from config but that part of the config has been stable since forever so this probably isn't the simplest fix (it'll require a config migration, announcement, light audit of downstream projects, etc.).

However, js-ipfs ran into a similar issue and came up with a different solution:  https://github.com/ipfs/js-ipfsd-ctl/pull/303#discussion_r233553928

We already allow passing a config on init but, unfortunately, we don't actually try to _merge_ the config (which is why that won't work in both this case and js-ipfs's case. Would something like that work for you?

[...]

PR #6489, opened 2019-07-09
feat: add ability to use existing config during init

djdv:
Related to:
https://github.com/ipfs/go-ipfs/issues/6262

This patch adds
`ipfs daemon --init --init-config-file=${PATH}`
and
`ipfs init --config-file=${PATH}`
allowing users to specify existing configuration files instead of (re)generating them.

Does not deal with merging of existing configs and/or profiles (yet?).

### DIFF
--- a/cmd/ipfs/daemon.go
+++ b/cmd/ipfs/daemon.go
@@ -13,6 +13,8 @@ import (
 	"sync"
 
 	version "github.com/ipfs/go-ipfs"
+	config "github.com/ipfs/go-ipfs-config"
+	cserial "github.com/ipfs/go-ipfs-config/serialize"
 	utilmain "github.com/ipfs/go-ipfs/cmd/ipfs/util"
 	oldcmds "github.com/ipfs/go-ipfs/commands"
 	"github.com/ipfs/go-ipfs/core"
@@ -26,11 +28,11 @@ import (
 	migrate "github.com/ipfs/go-ipfs/repo/fsrepo/migrations"
 
 	"github.com/hashicorp/go-multierror"
-	"github.com/ipfs/go-ipfs-cmds"
+	cmds "github.com/ipfs/go-ipfs-cmds"
 	mprome "github.com/ipfs/go-metrics-prometheus"
 	goprocess "github.com/jbenet/goprocess"
 	ma "github.com/multiformats/go-multiaddr"
-	"github.com/multiformats/go-multiaddr-net"
+	manet "github.com/multiformats/go-multiaddr-net"
 	"github.com/prometheus/client_golang/prometheus"
 )
 
@@ -38,6 +40,7 @@ const (
 	adjustFDLimitKwd          = "manage-fdlimit"
 	enableGCKwd               = "enable-gc"
 	initOptionKwd             = "init"
+	initConfigOptionKwd       = "init-config"
 	initProfileOptionKwd      = "init-profile"
 	ipfsMountKwd              = "mount-ipfs"
 	ipnsMountKwd              = "mount-ipns"
@@ -154,6 +157,7 @@ Headers.
 
 	Options: []cmds.Option{
 		cmds.BoolOption(initOptionKwd, "Initialize ipfs with default settings if not already initialized"),
+		cmds.StringOption(initConfigOptionKwd, "Path to existing configuration file to be loaded during --init"),
 		cmds.StringOption(initProfileOptionKwd, "Configuration profiles to apply for --init. See ipfs init --help for more"),
 		cmds.StringOption(routingOptionKwd, "Overrides the routing option").WithDefault(routingOptionDefaultKwd),
 		cmds.BoolOption(mountKwd, "Mounts IPFS to the filesystem"),
@@ -229,16 +233,19 @@ func daemonFunc(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment
 	// first, whether user has provided the initialization flag. we may be
 	// running in an uninitialized state.
 	initialize, _ := req.Options[initOptionKwd].(bool)
-	if initialize {
+	if initialize && !fsrepo.IsInitialized(cctx.ConfigRoot) {
+		cfgLocation, _ := req.Options[initConfigOptionKwd].(string)
+		profiles, _ := req.Options[initProfileOptionKwd].(string)
+		var conf *config.Config
 
-		cfg := cctx.ConfigRoot
-		if !fsrepo.IsInitialized(cfg) {
-			profiles, _ := req.Options[initProfileOptionKwd].(string)
-
-			err := initWithDefaults(os.Stdout, cfg, profiles)
-			if err != nil {
+		if cfgLocation != "" {
+			if conf, err = cserial.Load(cfgLocation); err != nil {
 				return err
 			}
+		}
+
+		if err = doInit(os.Stdout, cctx.ConfigRoot, false, nBitsForKeypairDefault, profiles, conf); err != nil {
+			return err
 		}
 	}
 
--- a/cmd/ipfs/init.go
+++ b/cmd/ipfs/init.go
@@ -7,7 +7,7 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"path"
+	"path/filepath"
 	"strings"
 
 	assets "github.com/ipfs/go-ipfs/assets"
@@ -16,9 +16,9 @@ import (
 	namesys "github.com/ipfs/go-ipfs/namesys"
 	fsrepo "github.com/ipfs/go-ipfs/repo/fsrepo"
 
-	"github.com/ipfs/go-ipfs-cmds"
-	"github.com/ipfs/go-ipfs-config"
-	"github.com/ipfs/go-ipfs-files"
+	cmds "github.com/ipfs/go-ipfs-cmds"
+	config "github.com/ipfs/go-ipfs-config"
+	files "github.com/ipfs/go-ipfs-files"
 )
 
 const (
@@ -27,6 +27,10 @@ const (
 	emptyRepoOptionName    = "empty-repo"
 	profileOptionName      = "profile"
 )
+
+var errRepoExists = errors.New(`ipfs configuration file already exists!
+Reinitializing would overwrite your keys.
+`)
 
 var initCmd = &cmds.Command{
 	Helptext: cmds.HelpText{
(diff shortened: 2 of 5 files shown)


cmd/ipfs/daemon.go
```diff
@@ -230,15 +233,31 @@ func daemonFunc(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment
 	// running in an uninitialized state.
 	initialize, _ := req.Options[initOptionKwd].(bool)
 	if initialize {
+		cfgLocation, _ := req.Options[initConfigOptionKwd].(string)
+		profiles, _ := req.Options[initProfileOptionKwd].(string)
 
-		cfg := cctx.ConfigRoot
-		if !fsrepo.IsInitialized(cfg) {
-			profiles, _ := req.Options[initProfileOptionKwd].(string)
+		if cfgLocation != "" {
```
Stebalien: We shouldn't _re-initialize_ if the config is specified. We need to bypass all this logic if `fsrepo.IsInitialized` returns true.

cmd/ipfs/init.go
```diff
@@ -102,8 +112,22 @@ environment variable:
 			}
 		}
 
+		cfgLocation, _ := req.Options[configOptionName].(string)
 		profile, _ := req.Options[profileOptionName].(string)
 
+		if cfgLocation != "" {
+			if profile != "" {
+				return errInitConfigArgs
```
Stebalien: Any reason to reject this?

cmd/ipfs/daemon.go
djdv: My understanding of the intent of `init`/`--init`, is that we'll be initalizing a new repo.
In this section, if the location of an existing config is specified, we parse it from that location, and write it out to the new repo/config location (`$IPFS_PATH`).
Basically using it as user specified template.

I may be wrong here, but I believe we don't reinitialize (we error when that's attempted).
We check the init target for existence, not the existing config location that was passed to us (through `--init-config-file`).
We populate the repo target config file, from the value passed in to us. Effectively copying it, then proceeding with daemon startup.

`fsrepo.IsInitialized()` is checked:
When the config location is not specified, we check in the else clause here:
https://github.com/ipfs/go-ipfs/blob/f1f726cd70ea7cf1c62d0c20a2a8613fb889b89e/cmd/ipfs/daemon.go#L255
When it is specified, we hit the check in doInit() here:
https://github.com/ipfs/go-ipfs/blob/f1f726cd70ea7cf1c62d0c20a2a8613fb889b89e/cmd/ipfs/init.go#L158

afterwards we continue the daemon startup process.

I believe we're meeting this expected behaviour.

cmd/ipfs/init.go
djdv: Probably not.
I was thinking we needed to merge profiles rather than use the config transformers. I'll update this.

djdv: Okay so this works now.
```CMD
set IPFS_PATH=Z:\ipfs-dev-tmp1
ipfs init
set IPFS_PATH=Z:\ipfs-dev-tmp2
ipfs daemon --init --init-config-file=Z:\ipfs-dev-tmp1\config -profile=badgerds
diff Z:\ipfs-dev-tmp1 Z:\ipfs-dev-tmp2
```
```diff

Only in Z:\ipfs-dev-tmp2: badgerds
Only in Z:\ipfs-dev-tmp1: blocks
diff "Z:\\ipfs-dev-tmp1/config" "Z:\\ipfs-dev-tmp2/config"
11,34c11,18
<       "mounts": [
<         {
<           "child": {
<             "path": "blocks",
<             "shardFunc": "/repo/flatfs/shard/v1/next-to-last/2",
<             "sync": true,
<             "type": "flatfs"
<           },
<           "mountpoint": "/blocks",
<           "prefix": "flatfs.datastore",
<           "type": "measure"
<         },
<         {
<           "child": {
<             "compression": "none",
<             "path": "datastore",
<             "type": "levelds"
<           },
<           "mountpoint": "/",
<           "prefix": "leveldb.datastore",
<           "type": "measure"
<         }
<       ],
<       "type": "mount"
---
>       "child": {
>         "path": "badgerds",
>         "syncWrites": true,
>         "truncate": true,
>         "type": "badgerds"
>       },
>       "prefix": "badger.datastore",
>       "type": "measure"
Only in Z:\ipfs-dev-tmp1: datastore
diff "Z:\\ipfs-dev-tmp1/datastore_spec" "Z:\\ipfs-dev-tmp2/datastore_spec"
1c1
< {"mounts":[{"mountpoint":"/blocks","path":"blocks","shardFunc":"/repo/flatfs/shard/v1/next-to-last/2","type":"flatfs"},{"mountpoint":"/","path":"datastore","type":"levelds"}],"type":"mount"}
\ No newline at end of file
---
> {"path":"badgerds","type":"badgerds"}
\ No newline at end of file
Common subdirectories: Z:\ipfs-dev-tmp1/keystore and Z:\ipfs-dev-tmp2/keystore
```

djdv:
Rewrote the first commit to be more accurate
```diff
-               cmds.StringOption(configOptionName, "Use supplied config instead of generating one"),
+               cmds.StringOption(configOptionName, "Use supplied config as a base instead of the default"),
```

cmd/ipfs/daemon.go
Stebalien: The intent of `--init` is to initialize if not already initialized.

> Initialize ipfs with default settings if not already initialized.

cmd/ipfs/init.go
Stebalien: I'd say allow both (but we can stick to transformers for now). That is:

1. Take the initial config, if supplied.
2. Merge it with the default config (later).
3. Apply transformers (if specified).

cmd/ipfs/daemon.go
djdv: I believe the current patch matches that. The only exception is that we use the settings from the specified config instead of the default config settings.

cmd/ipfs/init.go
```diff
@@ -102,31 +109,41 @@ environment variable:
...
 		}
 
-		profile, _ := req.Options[profileOptionName].(string)
+		cfgLocation, _ := req.Options[configOptionName].(string)
+		profiles, _ := req.Options[profileOptionName].(string)
 
-		var profiles []string
-		if profile != "" {
-			profiles = strings.Split(profile, ",")
+		if cfgLocation != "" {
```
Stebalien: Hm, it looks like we _currently_ accept one argument for the "initial config". We should:

1. Make sure we aren't passed both.
2. Consider deprecating that?

Or maybe we don't need this flag on init (we could _only_ support passing the config as an argument on init?). @hugomrdias, what's the state of javascript here?

cmd/ipfs/init.go
```diff
@@ -102,31 +109,41 @@ environment variable:
...
+			conf, err := cserial.Load(cfgLocation)
+			if err != nil {
+				return err
+			}
+
+			return doInit(os.Stdout, cctx.ConfigRoot, false, nBitsForKeypairDefault, profiles, conf)
```
Stebalien: Why not use the `doInit` call below? This one drops the `empty` flag and passes the wrong `nBitsForKeypair` option (although I guess _that_ doesn't do anything right now).

cmd/ipfs/daemon.go
Stebalien: The difference is that we return an _error_ when we do it in `doInit` (aborting the daemon).

djdv: I've made the patch a little bit more sane.
If I'm following correctly, the intent is to abort the daemon if the config exists.
If we're told to initialize during `daemon` when a repo is already initalized, doInit() will return an already-initialized error to us, which halts daemon startup.
`init` does the same.

image

Stebalien: >  If I'm following correctly, the intent is to abort daemon startup if the repo already exists.

The intent is to not re-initialize if the repo already exists. The "--init" flag is designed to auto-initialize go-ipfs on first start _when necessary_. The `--init-config` flag shouldn't change this.

djdv: Oh I see what's going on now. I was confusing/conflating `--init` and `init` behaviour.
In the latest patch, `--init` won't abort the daemon if the repo is already initialized. 
But `init` will error if the repo is already initalized.
There's also a commit that adds a warning for this but I don't know how useful that is. (droppable)

Stebalien: >  There's also a commit that adds a warning for this but I don't know how useful that is. (droppable)

There should be no warning/error. Running `ipfs daemon --init` with an initialized repo is normal behavior. I.e., I'd expect a decent system service to start IPFS with:

```
ipfs daemon --migrate --init
```

That way, the user can just install IPFS and start it, no setup required.

djdv: Gotcha. 
Branch has been reset to the commit without the warning.

Stebalien:
Status: Blocked on `ipfs init /path/to/config` versus `ipfs init --config /path/to/config`. We currently do the former, this patch does both. I'd prefer to not have two ways to do the same thing if possible.

cc @hugomrdias?

lanzafame:
What does `ipfs init /path/to/config` do?

Stebalien:
@lanzafame at the moment, it initializes with an existing config.

djdv:
@Stebalien 
>ipfs init --config /path/to/config

Whether this was intended or not, I have to be pedantic here since it's relevant.

`ipfs init --config` (global flag) is not the same as `ipfs init --config-file` or `ipfs daemon --init --init-config-file` (subcommand specific flags).

It's also a bit of a hasty blunder on my part to have forgotten the existing `--config` flag as well as the `<default-config>` handling. There's a lot of redundancy happening in this patch.

If we can, I'm not opposed to using `ipfs init --config` to act the same as `ipfs init --config-file`.
But this may be confusing if we handle that global flag in a unique way inside of `init` than we do in other other commands.

Explicitly pointing out that `init` handles config via stdin or a file path argument where `--config-file` only handles file paths.

Stebalien:
Oh dear. `--config` appear to be equivalent to `IPFS_PATH=... ipfs`. That's _really_ confusing and we should make that sane (i.e., make `--config` point to an out-of-repo config). But that's a separate issue.

Yeah, we need to make sure js/go agree here or it's going to be really confusing.

hugomrdias:
in js:
- we don't have the global --config flag
- ipfs init accepts a json string input
  - `ipfs init "{}"`
- ipfs daemon doesn't support init or config 
- we do have IPFS_PATH for the repo path

making ipfs [cmd] --config point to an out-of-repo config will make `ipfs init /path/config` equal to `ipfs init --config /path/config` right ?
If the above is the way to go we don't need anything else, but that is probably a bigger change that we don't want right now.

so maybe,
- ipfs init /path/config (no changes at all to master)
- ipfs daemon --init --init-config /path/config

makes sense ?

On js side i'll code whatever we decide here no blockers there.

djdv:
Went ahead with the change.
Removed `ipfs init --config-file` flag (in favour of existing file+stdin handling via `<default-config>`).
`ipfs daemon --init --init-config-file` became `ipfs daemon --init --init-config`
Cleanup in `init.go` remains in patch.

hugomrdias:
@Stebalien can we unblock this ?

hugomrdias:
@djdv anything i can do to help with this ?

PR #6493, opened 2019-07-10
pin cmd: stream recursive pins

MichaelMure:
Add a --stream flag to stream the results instead of
accumulating the final result in memory.

This is a rework of https://github.com/ipfs/go-ipfs/pull/5005

### DIFF
--- a/core/commands/pin.go
+++ b/core/commands/pin.go
@@ -494,6 +494,12 @@ func pinLsAll(req *cmds.Request, typeStr string, n *core.IpfsNode, emit func(val
 			return err
 		}
 	}
+	if typeStr == "recursive" || typeStr == "all" {
+		err := AddToResultKeys(n.Pinning.RecursiveKeys(), "recursive")
+		if err != nil {
+			return err
+		}
+	}
 	if typeStr == "indirect" || typeStr == "all" {
 		for _, k := range n.Pinning.RecursiveKeys() {
 			var visitErr error
@@ -519,12 +525,6 @@ func pinLsAll(req *cmds.Request, typeStr string, n *core.IpfsNode, emit func(val
 			if err != nil {
 				return err
 			}
-		}
-	}
-	if typeStr == "recursive" || typeStr == "all" {
-		err := AddToResultKeys(n.Pinning.RecursiveKeys(), "recursive")
-		if err != nil {
-			return err
 		}
 	}
 
--- a/core/commands/pin.go
+++ b/core/commands/pin.go
@@ -508,7 +508,7 @@ func pinLsAll(req *cmds.Request, typeStr string, n *core.IpfsNode, emit func(val
 				if r {
 					err := emit(&PinLsOutputWrapper{
 						PinLsObject: PinLsObject{
-							Type: typeStr,
+							Type: "indirect",
 							Cid:  enc.Encode(c),
 						},
 					})
--- a/core/commands/pin.go
+++ b/core/commands/pin.go
@@ -393,24 +393,28 @@ Example:
 	},
 }
 
-type PinLsObject struct {
-	Cid  string `json:",omitempty"`
-	Type string `json:",omitempty"`
-}
-
-type PinLsType struct {
-	Type string
-}
-
-type PinLsList struct {
-	Keys map[string]PinLsType `json:",omitempty"`
-}
-
+// PinLsOutputWrapper is the output type of the pin ls command.
 // Pin ls needs to output two different type depending on if it's streamed or not.
 // We use this to bypass the cmds lib refusing to have interface{}
 type PinLsOutputWrapper struct {
 	PinLsList
 	PinLsObject
+}
+
+// PinLsList is a set of pins with their type
+type PinLsList struct {
+	Keys map[string]PinLsType `json:",omitempty"`
+}
+
+// PinLsType contains the type of a pin
+type PinLsType struct {
+	Type string
+}
+
+// PinLsObject contains the description of a pin
+type PinLsObject struct {
+	Cid  string `json:",omitempty"`
+	Type string `json:",omitempty"`
 }
 
 func pinLsKeys(req *cmds.Request, typeStr string, n *core.IpfsNode, api coreiface.CoreAPI, emit func(value interface{}) error) error {
--- a/core/commands/pin.go
+++ b/core/commands/pin.go
@@ -259,8 +259,9 @@ collected if needed. (By default, recursively. Use -r=false for direct pins.)
 }
 
 const (
-	pinTypeOptionName  = "type"
-	pinQuietOptionName = "quiet"
+	pinTypeOptionName   = "type"
+	pinQuietOptionName  = "quiet"
+	pinStreamOptionName = "stream"
 )
 
 var listPinCmd = &cmds.Command{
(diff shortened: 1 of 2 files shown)


MichaelMure:
Notes:
- pins are streamed and start being emitted immediately
- direct and recursive are streamed super fast first, around 76k pins/s on my test machine
- indirect pins are still streamed very slowly, around 5 pins/s, most likely due to no parallelization in go-merkledag: https://github.com/ipfs/go-merkledag/blob/master/merkledag.go#L287

core/commands/pin.go
```diff
@@ -492,80 +508,127 @@ var verifyPinCmd = &cmds.Command{
...
 type RefKeyList struct {
-	Keys map[string]RefKeyObject
+	Keys map[string]RefObject
 }
 
-func pinLsKeys(ctx context.Context, args []string, typeStr string, n *core.IpfsNode, api coreiface.CoreAPI) (map[cid.Cid]RefKeyObject, error) {
+// Pin ls needs to output two different type depending on if it's streamed or not.
+// We use this to bypass the cmds lib refusing to have interface{}
+type PinLsOutputWrapper struct {
+	RefKeyList
```
Stebalien: We're going to have to use "omitempty" for full backwards compatibility.

MichaelMure:
> This will need to correctly handle CID encoding (introduced since #5005).

Any pointer/example about this ? I don't really get what you are talking about.

core/commands/pin.go
```diff
@@ -337,34 +337,50 @@ Example:
...
-			keys, err = pinLsAll(req.Context, typeStr, n)
+			err = pinLsAll(req.Context, typeStr, n, emit)
 		}
 		if err != nil {
 			return err
 		}
 
-		refKeys := make(map[string]RefKeyObject, len(keys))
-		for k, v := range keys {
-			refKeys[enc.Encode(k)] = v
```
Stebalien: Encoding code.

core/commands/pin.go
```diff
@@ -337,34 +337,50 @@ Example:
 			return err
 		}
 
-		enc, err := cmdenv.GetCidEncoder(req)
```
Stebalien: Other encoding code.

MichaelMure:
A few improvements:
- CID are encoded properly
- better struct naming (they were named after the `refs` command but are only used by `pin/ls` now)
- struct field are properly JSON omitempty-ed for backward compat
- output struct are documented to make codeclimate happy

MichaelMure:
@lanzafame not sure how I would do a go benchmark for this. This PR doesn't fundamentally change anything perf wise, it just start to output the result as it comes. The total run time will be similar.

MichaelMure:
More improvements:
- fixed a bug where indirect pins were reported as "all"
- output the recursive pins first as they don't require walking the graph
- use `EnumerateChildrenAsync` instead of `EnumerateChildren` for faster/parallel walking of the graph

```
root@ip-10-0-25-166:/tmp# timeout -k 1 59s ./ipfsseq pin ls --stream --type=indirect > indirectseq
12:50:51.021 ERROR       core: failure on stop:  context canceled builder.go:47
Killed
root@ip-10-0-25-166:/tmp# wc -l indirectseq
786 indirectseq
root@ip-10-0-25-166:/tmp# timeout -k 1 59s ./ipfsasync pin ls --stream --type=indirect > indirectasync
12:52:52.668 ERROR       core: failure on stop:  context canceled builder.go:47
Error: canceled
root@ip-10-0-25-166:/tmp# wc -l indirectasync
2645 indirectasync
```

For indirect pins:
Sequential walking: 13 pins/s
Parallel walking: 44 pins/s

--> ~3.4x faster !

It also looks like this change to use parallel graph walking might be used elsewhere. A quick search show that at least the GC might benefit from that.

MichaelMure:
Test are now failing due to using `EnumerateChildrenAsync`. If `EnumerateChildren` is used instead, tests are ok (at least as far as I can tell, I have some unrelated failing on my machine).

Stebalien:
>  is used instead, tests are ok (at least as far as I can tell, I have some unrelated failing on my machine).

We probably need to sort in the tests.

MichaelMure:
FYI, I removed the commit that just change `EnumerateChildren` to `EnumerateChildrenAsync`. As you can see, the CI is now okay-ish. If you have insights on the remaining failure and why using `EnumerateChildrenAsync` could be problematic, I'm interested.

Stebalien:
I'm _guessing_ that the traversal order is random, causing the comparison against the expected pin list to fail. We need to sort both before comparing.

MichaelMure:
I checked that, but the tests look ok to me:
https://github.com/ipfs/interop/blob/master/test/pin.js#L89-L90
https://github.com/ipfs/interop/blob/master/test/pin.js#L155-L157

Stebalien:
So... this is a bug. `EnumerateChildren` enumerates children. `EnumerateChildrenAsync` _also_ enumerates the root CID. I believe we should make it operate like `EnumerateChildren` but we're going to have to carefully make sure we don't break something in the process.

MichaelMure:
Hoooo, that explains a lot. One problem I had after adding the recursive pins before the indirect pins to the set and using `EnumerateChildrenAsync` was that indirect pins would not be listed anymore. It was because the root cid was returned first by `EnumerateChildrenAsync`, considered as already enumerated and the whole subgraph was pruned.

Stebalien:
Fix in https://github.com/ipfs/go-ipfs/pull/6499. Well, not really a _fix_, I just renamed these functions so that the names/documentation match up with what they do...

lanzafame:
> @lanzafame not sure how I would do a go benchmark for this. This PR doesn't fundamentally change anything perf wise, it just start to output the result as it comes. The total run time will be similar.

With a benchmark, you can run `go test -run none -bencmark . -benchmem` and you will get number of allocations etc.

Stebalien:
@alanshaw this PR adds a `--stream` (short `-s`) option to `ipfs pin ls` that returns of pins as a stream of :

```json
{"Cid": "Qm1", "Type": "direct"}
{"Cid": "Qm2", "Type": "recursive"}
...
```

Instead of a single:

```json
{
  "Keys": {
    "Qm1": "direct",
    "Qm2": "recursive",
  }
}
```

This is completely backwards compatible but is likely something js-ipfs will want to implement. Thoughts? Objections?

alanshaw:
Yes please. I want this. No objections.

Stebalien:
:rocket:!

Thanks @MichaelMure, @lanzafame, and @alanshaw!

MichaelMure:
Ha, I was waiting for https://github.com/ipfs/go-ipfs/pull/6499#issuecomment-510783458 to get resolved so I could have the neat 3.4x perf improvement with the parallel walking, but ok !

Stebalien:
Test case in https://github.com/ipfs/go-ipfs/pull/6504.

Note: IMO, a benchmark isn't really useful here as this _just_ streams the results, it doesn't reduce overall memory (still need to deduplicate results by storing them in a map).

Stebalien:
> Ha, I was waiting for #6499 (comment) to get resolved so I could have the neat 3.4x perf improvement with the parallel walking, but ok !

Let's do that in a followup. No reason to block this on a different change.

PR #6495, opened 2019-07-10
Docker initscript: allow changing API and Gateway config addresses

pataquets:
Allows easier tinkering with networking configuration, but using previous values as defaults to not break anything.
Minor syntax best practices for safety.
Also, would be worth considering changing `set -e` by `set -euo pipefail` as a best practice. `u` being break on unset vars, and `-o pipefail` to do not hide piped command failures (aka fail loud).

### DIFF
--- a/bin/container_daemon
+++ b/bin/container_daemon
@@ -1,31 +1,31 @@
 #!/bin/sh
 set -e
 user=ipfs
-repo="$IPFS_PATH"
+repo="${IPFS_PATH}"
 : ${addr_api:=/ip4/0.0.0.0/tcp/5001}
 : ${addr_gw:=/ip4/0.0.0.0/tcp/8080}
 
 if [ `id -u` -eq 0 ]; then
-  echo "Changing user to $user"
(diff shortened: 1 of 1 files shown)


bin/container_daemon
```diff
@@ -1,29 +1,31 @@
...
     "") INIT_ARGS="" ;;
-    *) INIT_ARGS="--profile=$IPFS_PROFILE" ;;
+    *) INIT_ARGS="--profile=${IPFS_PROFILE}" ;;
   esac
-  ipfs init $INIT_ARGS
-  ipfs config Addresses.API /ip4/0.0.0.0/tcp/5001
-  ipfs config Addresses.Gateway /ip4/0.0.0.0/tcp/8080
+  ipfs init ${INIT_ARGS}
+  ipfs config Addresses.API ${addr_api}
+  ipfs config Addresses.Gateway ${addr_gw}
```
Stebalien: addr_gw shouldn't have a space but let's quote it anyways.

PR #6501, opened 2019-07-12
Improve pin/update description

hsanjuan:
I was unable to actually understand what pin update does from the existing
description.


Stebalien:
Thanks!

PR #6509, opened 2019-07-13
[WIP] add experimental revive option for  reproducible CIDs

drbh:
## Draft solution for *reviving* a CID. 

Based on the [deep dive specs](https://github.com/ipfs/camp/blob/master/DEEP_DIVES/35-deterministic-cids-reproducible-file-imports-verifiable-http-gateways.md#data-preservation) from IPFS Camp 2019;

This is the case when you have a file - and you know what you expect the CID to be, but you do not know the original params that generated that DAG root.

We implement a `--revive` flag (also called `--expected-cid` in the spec) on the `ipfs add` core command.

**In short**
We want to automatically loop through known default settings check the CID and return to the user if we find the correct params.

## Example CLI use
```bash
ipfs add --revive=QmTfwRggLp9mo7fTF5L5cCxk8djbE6reipLpMMNfYgA6ip /tmp/cat.jpg 
```

## Proof of concept work
🐥Add `revive` flag
🐥Check if flag is called
🐣Loop over default settings
🐣Return useful information on match

## Discussion topics
🥚What is returned to the user?
🥚If its the `fmtstr` what does that look like?

## Answerable questions
❓What are the known default settings (when CID versions change)?
❓How to loop over `*files.multipartDirectory` more than once?

### DIFF
--- a/core/commands/add.go
+++ b/core/commands/add.go
@@ -45,6 +45,7 @@ const (
 	hashOptionName        = "hash"
 	inlineOptionName      = "inline"
 	inlineLimitOptionName = "inline-limit"
+	reviveOptionName      = "revive"
 )
 
 const adderOutChanSize = 8
@@ -129,6 +130,7 @@ You can now check what blocks have been created by:
 		cmds.StringOption(hashOptionName, "Hash function to use. Implies CIDv1 if not sha2-256. (experimental)").WithDefault("sha2-256"),
 		cmds.BoolOption(inlineOptionName, "Inline small blocks into CIDs. (experimental)"),
 		cmds.IntOption(inlineLimitOptionName, "Maximum block size to inline. (experimental)").WithDefault(32),
+		cmds.StringOption(reviveOptionName, "Iterate through known params to revive a CID from a file (experimental)"),
 	},
 	PreRun: func(req *cmds.Request, env cmds.Environment) error {
 		quiet, _ := req.Options[quietOptionName].(bool)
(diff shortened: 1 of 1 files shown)


core/commands/add.go
```diff
@@ -218,6 +220,101 @@ You can now check what blocks have been created by:
...
+				}
+
+				my_opts = append(my_opts, nil) // events option placeholder
+
+				fmt.Println(my_opts)
+
+				// var added int
+				addit := special.Entries()
+
+				for addit.Next() {
```
drbh: `for addit.Next()` doesn't initiate a loop on the second default settings. Seems like the entries are removed on each loop. Is there a way we can persist the file as we try different params?

lanzafame:
Can we iterate over the flag name? I am not a fan of `expected-cid` but `revive` seems really misleading, maybe `reconstruct`?

drbh:
agreed `revive` and `expected-cid` are not the most descriptive. 

I like `reconstruct` alot ✅

drbh:
@lanzafame 

Just pushed some minor updates (renamed vars for golint) 

Also renamed `revive` to `reconstruct`! 

### Use
```
ipfs add --reconstruct=QmTfwRggLp9mo7fTF5L5cCxk8djbE6reipLpMMNfYgA6ip /Users/drbh2/Desktop/cat.jpg
```

olizilla:
If we're offering up a CID to `ipfs add` along with a file, and saying _"figure out what parameters to use to get this CID from that file"_ then I feel like a more idiomatic name for the flag would be `--target=<cid>`. 

```console
ipfs add --target=QmTfwRggLp9mo7fTF5L5cCxk8djbE6reipLpMMNfYgA6ip /path/to/file
```

Not a blocker, I can totally live with `--reconstruct` but i felt compelled to comment.

core/commands/add.go
Stebalien: Technically, you could buffer them all into memory. But that's _really_ not something you want to do.

gammazero:
This is such an infrequent activity that it should be done manually, or by an external script. It should not be a feature of the command.

lidel:
If anyone finds this when searching for "reproducible CIDs" relevant work happened in 
- https://github.com/ipfs/specs/pull/499 
  - #11071

PR #6530, opened 2019-07-18
feat: allow /ipns/webui.ipfs.io on api port

olizilla:
We want to allow user to try out the latest versions of the webui before they are officially released with go-ipfs. 

With this change in place, a user will be able to visit http://localhost:5001/ipns/webui.ipfs.io
and have it fetch the latest version of the webui. go-ipfs can and should continue to ship with a hardcoded CID for a stable release of webui that it has been tested with, but including the ipns address in the list allows us to test out newer versions.

Right now ipfs-companion includes a workaround to strip the origin header from blessed, new versions of webui, to allow this, but it causes other issues, so we'd like to replace that hack, with this feature.

see: https://github.com/ipfs-shipyard/ipfs-companion/issues/736

### DIFF
--- a/core/corehttp/webui.go
+++ b/core/corehttp/webui.go
@@ -23,6 +23,7 @@ var WebUIPaths = []string{
 	"/ipfs/QmRyWyKWmphamkMRnJVjUTzSFSAAZowYP4rnbgnfMXC9Mr",
 	"/ipfs/QmU3o9bvfenhTKhxUakbYrLDnZU7HezAVxPM6Ehjw9Xjqy",
 	"/ipfs/QmPhnvn747LqwPYMJmQVorMaGbMSgA7mRRoyyZYz3DoZRQ",
+	"/ipns/webui.ipfs.io",
 }
 
 var WebUIOption = RedirectOption("webui", WebUIPath)


Stebalien:
I'm a bit concerned about a MITM attack given that DNSLink doesn't check DNSSEC/DNS over TLS. An attacker on, e.g., a coffee shop network could replace the webui and mess with the API directly.

lidel:
What if we require presence of valid DNSSEC records when `/ipns/webui.ipfs.io` is loaded via API port, and return error when unable to verify?

ps. at the moment ipfs.io does not have DNSSEC: [result](https://user-images.githubusercontent.com/157609/61551543-c248be00-aa55-11e9-9e0c-57016e09ac1e.png) from https://dnssec-analyzer.verisignlabs.com/

Stebalien:
Ideally, we'd do that _and_ use DNS over TLS. Alternatively, we could try to use normal certificates. I'm not sure how much DNSSEC should be trusted.

lidel:
Yes, DNS over TLS/HTTPS is the key mitigation against coffee shop MITM. 

Either way, this means go-ipfs should not use DNS provided by OS, as we don't know if the response was obtained over TLS. We need to pick a default resolver with support for DNS over TLS (or HTTPS) + make it possible for users to set a different one via config ([basically what Firefox does for DNS over HTTPS](https://wiki.mozilla.org/Trusted_Recursive_Resolver)).

Filled https://github.com/ipfs/go-ipfs/issues/6532 to discuss the details.

lidel:
I don't believe we need this anymore.

With subdomain gateway on localhost (#6096), users who want to use the latest Web UI can add `http://webui.ipfs.io.ipns.localhost:8080` to `API.HTTPHeaders.Access-Control-Allow-Origin` whitelist in go-ipfs config.

In the future Companion will provide a simple toggle for opt-in into latest webui: https://github.com/ipfs-shipyard/ipfs-companion/issues/860

Ok to close?

PR #6539, opened 2019-07-26
fix and improve the writable gateway

Stebalien:
1. Fix handling of PUT. The simple implementation was the correct implementation, I have no idea what was going on here.
2. Use MFS everywhere to reduce code duplication and add support for sharded directories.
3. _Correctly_ block IPNS.
4. Remove the dependency on `core.IpfsNode`.


lanzafame:
@Stebalien Thanks for this PR!

PR #6544, opened 2019-07-30
Fix/register first block metric

lanzafame:
Register 'time to first block' metric for the gateway only.


lanzafame:
forgot to rebase to stablize branch before pushing.

PR #6545, opened 2019-07-30
register first block metric for gateway only

lanzafame:
Register 'time to first block' metric for the gateway only.

### DIFF
--- a/core/corehttp/metrics.go
+++ b/core/corehttp/metrics.go
@@ -10,7 +10,7 @@ import (
 	promhttp "github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
-// This adds the scraping endpoint which Prometheus uses to fetch metrics.
+// MetricsScrapingOption adds the scraping endpoint which Prometheus uses to fetch metrics.
 func MetricsScrapingOption(path string) ServeOption {
 	return func(n *core.IpfsNode, _ net.Listener, mux *http.ServeMux) (*http.ServeMux, error) {
 		mux.Handle(path, promhttp.HandlerFor(prometheus.DefaultGatherer, promhttp.HandlerOpts{}))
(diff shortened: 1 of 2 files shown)


core/corehttp/metrics.go
```diff
@@ -29,6 +29,16 @@ func MetricsCollectionOption(handlerName string) ServeOption {
 			Objectives:  map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.99: 0.001},
 		}
 
+		if handlerName == "gateway" {
+			if err := prometheus.Register(unixfsGetMetric); err != nil {
+				if are, ok := err.(prometheus.AlreadyRegisteredError); ok {
+					unixfsGetMetric = are.ExistingCollector.(*prometheus.SummaryVec)
```
Stebalien: The race detector is _not_ going to like that. Given how this is currently defined, we should probably just register this on init. We could, alternatively, store the metric in the gateway handler itself and register it on construction. But, as far as I can tell, init seems like the more correct approach (I have no idea, really).

Note: checking "gateway" doesn't really help given that the metric is a global anyways.

lanzafame: The race detector doesn't _seem_ to care but it is out of place, I just didn't want a second place of metric registration. I am happy to move it to the gateway handler itself.

Stebalien: Either way. We just shouldn't be setting a global in a function like this.

lanzafame: Actually, would there ever be the case when that metric is already registered?

Stebalien: Maybe but it shouldn't matter either way (I guess we could return an _error_ if it's already registered but we've registered the _wrong_ metric).

PR #6560, opened 2019-08-07
Added custom DNS resolver with support for DNS-over-TLS and DNS-over-HTTPS

### LINKED ISSUE #6532: Configurable DNS Servers
## Context

Right now go-ipfs uses DNS provided by the operating system.

It brings serious downsides:
- DNS-based surveillance and censorship at ISP level ([context](https://www.zdnet.com/article/mozilla-no-plans-to-enable-dns-over-https-by-default-in-the-uk/))
- Browsing DNSLink websites in coffee shops is risky: We are unable to tell if the DNS query was resolved over unencrypted connection. Most of the DNS servers provided by public WiFis are plaintext. This means someone could do MITM on and inject malicious response, changing DNSLink etc. (#6530)
- Leaking plaintext queries in privacy-sensitive contexts (https://github.com/ipfs/go-ipfs/issues/5945, https://github.com/ipfs-shipyard/ipfs-companion/issues/900)

## Hardening DNS in go-ipfs

There are multiple steps to improve the situation, the basic ones would be:

- Stop using DNS provided by OS. We are unable to guarantee it uses TLS, it should never be used by go-ipfs.
- Pick a safe DNS over TLS (or HTTPS) provider and make it implicit default.
  (see https://wiki.mozilla.org/Trusted_Recursive_Resolver for prior art)
- Make it possible for users to override default DNS resolver via `DNS.*` config
  - Ideally, we should support plaintext DNS (only on localhost), DNS over TLS and DNS over HTTPS.
     - If we want to minimize the scope: focus on DNS over HTTPS (DoH). It works in browser, and we want to use it in js-ipfs  (https://github.com/ipfs/js-ipfs/issues/2212) as replacement for calling `https://ipfs.io/api/v0/dns` all the time. 
     - There should be a way to configure custom resolver per TLD or domain

## Ongoing work

- [x] https://github.com/protocol/web3-dev-team/pull/42: Proposal: configurable DNS resolvers for human-readable names
  - :point_right: **this document is an updated problem statement and definition of the end user UX we aim for** 
- [ ] go-ipfs https://github.com/ipfs/go-ipfs/pull/8068: support custom DoH resolvers, depends on:
  - [x] https://github.com/multiformats/go-multiaddr-dns/pull/26: refactor: Resolver to support custom per-TLD resolvers
  - [x] https://github.com/libp2p/go-doh-resolver/pull/1/: create DoH resolver
  - [x] https://github.com/ipfs/go-namesys/pull/10: make DNS resolver pluggable, use new madns.BasicResolver interface 
  - [x] https://github.com/libp2p/go-libp2p/pull/1073:  option for custom dns resolver 
  - [x] <del>https://github.com/ipfs/interface-go-ipfs-core/pull/69</del>
  - [x] https://github.com/ipfs/go-ipfs-config/pull/126: add custom DNS Resolver configuration
  - [ ] https://github.com/libp2p/go-libp2p/issues/1079: go-libp2p release

ItalyPaleAle:
This PR should fix #6532

In the current implementation, go-ipfs uses the machine's DNS resolver for querying TXT records used by DNSLink. As highlighted in #6532, this can come with some issues related to censorship, privacy, and support for TOR.

I've implemented a custom DNS resolver that supports:

1. Traditional DNS queries (over plain-text UDP), but using a customizable DNS server
1. DNS-over-TLS, with a custom DNS server
1. DNS-over-HTTPS, with a custom DNS server

Note that the PR is still not complete. At the moment, I am missing:

- [X] Being able to pass the configuration for the DNS server from the config file
- [ ] Adding tests

### DIFF
--- a/namesys/dns.go
+++ b/namesys/dns.go
@@ -3,7 +3,6 @@ package namesys
 import (
 	"context"
 	"errors"
-	"net"
 	"strings"
 
 	path "github.com/ipfs/go-path"
@@ -11,18 +10,24 @@ import (
 	isd "github.com/jbenet/go-is-domain"
 )
 
+// LookupTXTFunc is the interface for the lookupTXT property of DNSResolver
 type LookupTXTFunc func(name string) (txt []string, err error)
 
 // DNSResolver implements a Resolver on DNS domains
 type DNSResolver struct {
 	lookupTXT LookupTXTFunc
-	// TODO: maybe some sort of caching?
-	// cache would need a timeout
 }
 
 // NewDNSResolver constructs a name resolver using DNS TXT records.
 func NewDNSResolver() *DNSResolver {
-	return &DNSResolver{lookupTXT: net.LookupTXT}
+	dns := &customDNS{
+		Address:          "1.1.1.1",
+		Protocol:         "dns-over-https",
+		DNSoverHTTPSHost: "cloudflare-dns.com",
+	}
+
+	return &DNSResolver{lookupTXT: dns.LookupTXT}
+	//return &DNSResolver{lookupTXT: net.LookupTXT}
 }
 
 // Resolve implements Resolver.
--- a/namesys/dns.go
+++ b/namesys/dns.go
@@ -3,13 +3,16 @@ package namesys
 import (
 	"context"
 	"errors"
+	"fmt"
 	"net"
+	"strconv"
 	"strings"
 
 	config "github.com/ipfs/go-ipfs-config"
 	path "github.com/ipfs/go-path"
 	opts "github.com/ipfs/interface-go-ipfs-core/options/namesys"
 	isd "github.com/jbenet/go-is-domain"
+	ma "github.com/multiformats/go-multiaddr"
 )
 
 // LookupTXTFunc is the interface for the lookupTXT property of DNSResolver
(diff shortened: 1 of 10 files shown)


ItalyPaleAle:
Ok I managed to understand how to read from config. Still pending tests

ItalyPaleAle:
Note: tests will fail until https://github.com/ipfs/go-ipfs-config/pull/40 is merged

Stebalien:
Thanks! This is awesome!

What if we took the configuration in as a multiaddr? That is:

* `/ip4/1.1.1.1/udp/53` -> DNS
* `/ip4/1.1.1.1/tcp/443/https` -> DNS over HTTPS
* `/ip4/1.1.1.1/tcp/443/https/HOST` -> when we need a hostname. 
* `/ip4/1.1.1.1/tcp/853/tls` -> DNS over TLS (need to define the multiaddr protocol code for TLS but we need one anyways).

Stebalien:
That way, we can even specify a set of resolvers (to be tried in series).

ItalyPaleAle:
I actually like that idea! Would we just have a single configuration entry, then, something like "DNS" (as a string)? If that's empty, it would fallback to the system's resolver.

Does go-ipfs already have a multi-address parser built-in? If you could please give me a pointer to that, I'll look at making the update tonight (unless you want to make it yourself, you do have write access to my forks :) )

Stebalien:
> something like "DNS" (as a string)? If that's empty, it would fallback to the system's resolver.

Just to be safe, I'd have:

```
"DNS": {
  "Resolver": /my/resolver...
  ...
}
```

We may want to make security requirements, caching, etc. configurable in the future.

Stebalien:
> Does go-ipfs already have a multi-address parser built-in? If you could please give me a pointer to that, I'll look at making the update tonight (unless you want to make it yourself, you do have write access to my forks :) )

https://github.com/multiformats/go-multiaddr/

However, we need to quickly add a TLS protocol (we already need one anyways).

Stebalien:
PR to define TLS multicodec: https://github.com/multiformats/multicodec/pull/145

ItalyPaleAle:
Doesn’t look like multiaddress would support something like this however?

`/ip4/1.1.1.1/tcp/443/https/HOST`

I could add the TLS protocol but I’m more worried about the host part?

To be frank, the Host part isn’t even strictly necessary:

1. You could omit that, however then the client will not be able to validate the hostname in the TLS certificate
1. You could use DNS-over-HTTPS by using the domain name rather than the IP. That will work, however there would need to be a DNS query first to resolve the ... resolver’s IP. And that query would go through the machine’s default resolver.

Stebalien:
> I could add the TLS protocol but I’m more worried about the host part?

See https://github.com/multiformats/multiaddr/issues/63#issuecomment-478621623. The current proposal is to define HTTP (which isn't really used in practice yet) as `/http/HOST/path` (given that HTTP 1.1 and on _require_ this). When omitted, the HOST would be implied or omitted.

Looking around, it seems that the _real_ issue is that js-libp2p has gone off and used the `http` protocol without a path.

...

I'll bring this up at the [next core implementations meeting](https://calendar.google.com/event?action=TEMPLATE&tmeid=Y2dvbTRkaG03NWhtYWI5bmNjcmplYjlrNnBqMzRiYjJjZ3BtYWI5aTZoZ2o2ZHBpNzVpNjRkcG42Y18yMDE5MDgxMlQxNjMwMDBaIGlwZnMuaW9fZWFsMzZ1Z3U1ZTc1czIwN2dmamN1MGFlODRAZw&tmsrc=ipfs.io_eal36ugu5e75s207gfjcu0ae84%40group.calendar.google.com&scp=ALL). For now, we can support `/dns/cloudflare-dns.com/tcp/443/https` while we get the multiaddr situation resolved once and for all.

ItalyPaleAle:
So, let me see if I understand correctly:

1. For TLS:
    1. First, need to wait for https://github.com/multiformats/multicodec/pull/145 to be merged
    1. Next step is to add the TLS protocol to https://github.com/multiformats/go-multiaddr/
2. For HTTPS: I'm confused to what the status is. Do we need to wait on https://github.com/multiformats/multiaddr/issues/63 ?

I won't be able to join the call next week, but I assume you'll want to discuss this PR :) If there are comments/feedback, we can take them offline here?

Stebalien:
> I won't be able to join the call next week, but I assume you'll want to discuss this PR :) If there are comments/feedback, we can take them offline here?

We have a rule: it didn't happen if it isn't written somewhere on github. So yes, all resolutions and reasons will end up here.

Note: we'll likely schedule a _second_ call during the first call to actually resolve the issue. I'll keep you in the loop in case you're interested (but it's only tangentially related to _this_ issue).

ItalyPaleAle:
Ok I pushed the code changes you requested:

1. DNS resolver can now be set with a multi-address string "DNS.Resolver", using the format you proposed above. If it's empty, uses the system's resolver
1. Two things are missing:
    1. Support for specifying a host after HTTPS, pending upstream
    1. Support for using the /tls protocol, pending upstream

As before, this depends on https://github.com/ipfs/go-ipfs-config/pull/40 (which I just updated) and tests will fail until that's merged.

Does this design look good to you now? If so, I will start working on the unit tests tonight or tomorrow night

ItalyPaleAle:
As I'm working on this, I noticed that some other parts of the codebase use a DNS resolver, calling the one inside https://github.com/multiformats/go-multiaddr-dns

I will need to update that too, likely

ItalyPaleAle:
I hit a road block while working on this during the weekend.

Turns out that https://github.com/multiformats/go-multiaddr-dns queries DNS too, and I believe that's used to resolve domain names inside multi-addresses in the config file (e.g. when the peers are defined with a DNS multi-address).

I could replace the DNS inside go-multiaddr-dns (either by changing the library or I believe at runtime too). The problem is that the DNS server to use is defined inside the .ipfs/config file, and I'm afraid of creating a bunch of spaghetti code because the config file is read after go-multiaddr-dns is loaded...

Thoughts?

ItalyPaleAle:
@Stebalien any update from today's call?

Stebalien:
Yes. Sorry for the delay. The conclusion was to _ignore_ everything I've told you and to just manually configure all of this. We realized that, technically, we need to be able to specify a _path_ for DNS-over-HTTPS and didn't feel like we'd be able to solve that issue in a single meeting.

I'm really sorry for the confusion, it just looks like multiaddr isn't _quite_ ready for this use-case.

Stebalien:
So yeah. Let's have a `Resolvers` list with a list of objects describing each resolver. Later, when we finally get multiaddr support for this, we can replace this list with a list of multiaddrs (and do some json type-parsing magic to distinguish between the two).

Stebalien:
The real issue is that the HTTP multiaddr really needs to be a URL _template_ (https://tools.ietf.org/html/rfc8484#section-4.1). Something like `https://dnsserver.example.net/dns-query{?dns}`.

Library: https://godoc.org/google.golang.org/api/googleapi/internal/uritemplates (?)

ItalyPaleAle:
Sorry for the radio silence on this, I have been very busy and I will be for at least another couple of weeks.

My biggest problem remains on how to reconcile the fact that go-multiaddr-dns needs a DNS resolver too.

go-multiaddr-dns  is used while loading the config file of IPFS (in fact, it's what resolves the DNS names within the config file). So, when go-multiaddr-dns is loaded, there's no way to pass it the IPFS configuration to set a DNS server

The cleanest solution would be:

1. Create a DNS resolver in a separate package.
1. Defining the DNS configuration in a **separate** config file.
1. Modify both go-multiaddr-dns and go-ipfs to use the DNS resolver from the package, and the separate config file.

Without that, I'm afraid we have a chicken-and-egg problem.

matyapiro31:
How about adding test on dns_test.go?

namesys/dns.go
```diff
@@ -3,26 +3,107 @@ package namesys
...
-	// TODO: maybe some sort of caching?
-	// cache would need a timeout
 }
 
 // NewDNSResolver constructs a name resolver using DNS TXT records.
-func NewDNSResolver() *DNSResolver {
-	return &DNSResolver{lookupTXT: net.LookupTXT}
+func NewDNSResolver(cfg *config.Config) (*DNSResolver, error) {
+	// Check if we're using a custom DNS server
+	if len(cfg.DNS.Resolver) > 0 {
```
matyapiro31, line 29: Please add system environmental variables.

lidel:
Superseded by https://github.com/ipfs/go-ipfs/pull/8068  which takes bit different approach that works for both multiaddrs and dnslink (also, namesys got moved to https://github.com/ipfs/go-namesys).

See "Ongoing work" in https://github.com/ipfs/go-ipfs/issues/6532 for full context.

PR #6586, opened 2019-08-19
Allow to configure/disable the datastore's ARC cache.

MichaelMure:
Require https://github.com/ipfs/go-ipfs-config/pull/41

This PR allow to change the size or disable the ARC cache in the datastore.


MichaelMure:
Test failure seem unrelated.

MichaelMure:
I'm missing a line to actually pass the config value, will add soon.

MichaelMure:
Fixed, but still require https://github.com/ipfs/go-ipfs-config/pull/41 to build.

MichaelMure:
@Stebalien what about this one ?

Stebalien:
Unfortunately, I'm not sure how to add this in a backwards compatible manner (yet). If we add this as-is, the ARC cache will be _disabled_ on all existing nodes as `ARCCacheSize` will default to 0.

We can ship a migration but I'd hate to do that. I believe the best way to do this is https://github.com/ipfs/go-ipfs-config/issues/53.

Stebalien:
Revisiting this, we _can_ make progress on this without rewriting the config system. We now have types like [config.Flag](https://github.com/ipfs/go-ipfs-config/blob/6b35ac321440a0f64cfe6b6015e6b26fc58b6ee3/types.go#L50) where we can encode both some type (e.g., a size as a string) _or_ the "default" (null).

Basically, we'd need a new special "size" type with a default option.

gammazero:
Replaced by #10614

PR #6597, opened 2019-08-21
dep: update datastore and swarm

Stebalien:
* Swarm: fixes data race.
* Datastore: make delete idempotent.


blocks/blockstoreutil/remove.go
```diff
@@ -47,10 +46,20 @@ func RmBlocks(blocks bs.GCBlockstore, pins pin.Pinner, cids []cid.Cid, opts RmBl
 		stillOkay := FilterPinned(pins, out, cids)
 
 		for _, c := range stillOkay {
-			err := blocks.DeleteBlock(c)
-			if err != nil && opts.Force && (err == bs.ErrNotFound || err == ds.ErrNotFound) {
-				// ignore non-existent blocks
-			} else if err != nil {
+			// Kept for backwards compatibility. We may want to
```
djdv: >We may want to remove this sometime in the future.

Maybe worth tagging with an explicit `// TODO` maybe not.

PR #6602, opened 2019-08-22
mount: switch over to the CoreAPI

Stebalien:
1. Fix resolution of sharded directories.
2. Slowly kill off users of the IpfsNode object.


Stebalien:
The second two commits just delete some code.

aschmahmann:
This interop test is now failing: `ipns locally using the same repo across implementations should publish an ipns record to a go daemon and resolve it using a js daemon through the reuse of the same repo - should publish an ipns record to a go daemon and resolve it using a js daemon through the reuse of the same repo`

Stebalien:
> This interop test is now failing

Not anymore? Not sure what this was, but we've fixed some interop issues since then.

Stebalien:
Context:

* Better (unified) path resolution.
* Deleted code.
* Works towards extracting this code into a separate package.

PR #6613, opened 2019-08-29
plugins: add support for plugin configs

Stebalien:
For now, configs specified in `daemon --init-config` and `init CONFIG` are not available. We should fix this eventually but isn't necessary for now (and supporting this will be annoying).

### DIFF
--- a/plugin/loader/loader.go
+++ b/plugin/loader/loader.go
@@ -3,8 +3,11 @@ package loader
 import (
 	"fmt"
 	"os"
+	"path/filepath"
 	"strings"
 
+	config "github.com/ipfs/go-ipfs-config"
+	cserialize "github.com/ipfs/go-ipfs-config/serialize"
 	coredag "github.com/ipfs/go-ipfs/core/coredag"
 	plugin "github.com/ipfs/go-ipfs/plugin"
 	fsrepo "github.com/ipfs/go-ipfs/repo/fsrepo"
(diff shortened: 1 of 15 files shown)


Stebalien:
@djdv CI is passing. This should be ready for review.

plugin/loader/loader.go
```diff
@@ -83,16 +86,32 @@ type PluginLoader struct {
...
 	plugins map[string]plugin.Plugin
 	started []plugin.Plugin
+	config  config.Plugins
+	repo    string
 }
 
 // NewPluginLoader creates new plugin loader
-func NewPluginLoader() (*PluginLoader, error) {
-	loader := &PluginLoader{plugins: make(map[string]plugin.Plugin, len(preloadPlugins))}
+func NewPluginLoader(repo string) (*PluginLoader, error) {
```
djdv: The repo a Plugin receives from the Environment should probably be normalized into absolute paths.
Not sure though, there may be a case for accepting relatives.

Abs patch 1/2
```suggestion
func NewPluginLoader(repo string) (*PluginLoader, error) {
	var err error
	if repo, err = filepath.Abs(repo); err != nil {
		return nil, err
	}
```

plugin/loader/loader.go
```diff
@@ -83,16 +86,32 @@ type PluginLoader struct {
...
+			return nil, err
+		}
+	}
 	for _, v := range preloadPlugins {
 		if err := loader.Load(v); err != nil {
 			return nil, err
 		}
 	}
+
+	if err := loader.LoadDirectory(filepath.Join(repo, "plugins")); err != nil {
```
djdv: Abs patch 2/2
```suggestion
	if err = loader.LoadDirectory(filepath.Join(repo, "plugins")); err != nil {
```

Stebalien: In general, I don't like using `=` in assignments inside if statements. The point of using an if statement like this is to scope the `err` variable to the if statement. Using a normal assignment lets it escape the scope and can be really confusing.

Stebalien: Basically, this should either be left as-is or should be:

```
err = loader.LoadDirectory(...)
if err != nil {
  // ...
}
```

plugin/loader/loader.go
Stebalien: Given that we don't do this anywhere else, I'd like to leave this as it is for consistency (that way the repo path is the same everywhere). I believe it'll _usually_ be an absolute path anyways.

If we do run into issues, we should be consistent and make this path absolute at a higher layer.

PR #6615, opened 2019-08-29
Update README.md

jessicaschilling:
Removed OpenBSD line per discussion started with https://github.com/ipfs/docs/issues/254#issuecomment-526372366


jessicaschilling:
cc @Stebalien 😄

PR #6620, opened 2019-09-04
docs: add ship date and next release issue opening time

alanshaw:
Adds estimated shipping date (can be fuzzy) to the release issue template and clarifies the time before which the next release issue should be opened. This communicates expectations for when the new release is due and will help with planning of included features. The community could even suggest inclusions.

@Stebalien you had the idea of adding estimates for each of the stages - do you want me to add them? I think I might leave these out of the js-ipfs release process for now unless requested as I think the actual release date is most important and we do already clarify that it should take around 3 weeks in the releases documentation.


Stebalien:
Let's leave this as-is.

PR #6624, opened 2019-09-04
docker: libdl dependency

gmasgras:
Docker build is currently broken due to an updated version of libdl.
```
Step 23/34 : COPY --from=0 /lib/x86_64-linux-gnu/libdl-2.24.so /lib/libdl.so.2
COPY failed: stat /var/lib/docker/overlay2/86ed53025341b07961f14ed4d63a25317be51264a0555b6f8a768cc9e749e8ca/merged/lib/x86_64-linux-gnu/libdl-2.24.so: no such file or directory
```
This copies the symlinked libdl.so.2 instead of specific version.


Stebalien:
Thanks!

PR #6643, opened 2019-09-15
ipfs namespace is now being provided to prometheus

lanzafame:
This fixes what was a rather confusing issue in prometheus where the `http_response_size_bytes` metrics was confused to be coming from nginx but was actually from ipfs.

### DIFF
--- a/core/corehttp/metrics.go
+++ b/core/corehttp/metrics.go
@@ -24,6 +24,7 @@ func MetricsCollectionOption(handlerName string) ServeOption {
 		// Adapted from github.com/prometheus/client_golang/prometheus/http.go
 		// Work around https://github.com/prometheus/client_golang/pull/311
 		opts := prometheus.SummaryOpts{
+			Namespace:   "ipfs",
 			Subsystem:   "http",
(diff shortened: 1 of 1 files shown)


core/corehttp/metrics.go
```diff
@@ -24,14 +24,14 @@ func MetricsCollectionOption(handlerName string) ServeOption {
...
 		opts := prometheus.SummaryOpts{
+			Namespace:   "ipfs",
 			Subsystem:   "http",
 			ConstLabels: prometheus.Labels{"handler": handlerName},
 			Objectives:  map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.99: 0.001},
 		}
 
 		reqCnt := prometheus.NewCounterVec(
 			prometheus.CounterOpts{
-				Namespace:   opts.Namespace,
```
Stebalien: Shouldn't we keep this?

lanzafame: Ahh yes we should.

olizilla: nudge @lanzafame

lanzafame:
@olizilla @Stebalien fixed. thanks for the nudge

Issue #6633, opened 2019-09-07
When double-clicked on Windows, ipfs.exe should start the daemon


Stebalien:
Currently, when a user double clicks on ipfs.exe on Windows, a terminal launches, some help text is printed, then the terminal exits. This usually blinks by faster than the user can see.

Instead, we should detect this case (if possible, see https://support.microsoft.com/en-us/help/99115/info-preventing-the-console-window-from-disappearing) and just launch the daemon in the background.

kpp:
Why? Add a daemon.bat to start the daemon

djdv:
Mainly this is because it's confusing to new users who are used to launching programs from their typical environment. Double clicking `ipfs.exe` provides no help or feedback to these users, it simply launches and disappears before they can even read the text.

However, there are also technical considerations here (that are minor, but let's talk about them anyway)
(stop reading here unless you like esoteric Windows knowledge)

`ipfs` doesn't rely on a full terminal or command interpreter to run. Windows provides a console host for CLI programs, which provides everything needed (without the need for `cmd` or a hosting terminal  outside of `conhost.exe`). It allows CLI programs like this to simply exist alone in the graphical environment, separate from a terminal.

Creating a batch file looks similar, but in actuality launches 2 programs prior to the target (the terminal (this persists), the cmd interpreter (this exits), and then a shared conhost (this persists)).
In the case of launching `ipfs` in a GUI environment on Windows, you would want to create a shortcut file (.lnk) with arguments, to launch `ipfs` directly, without any implied dependencies.

However, since the case of running `ipfs` without any arguments is already deemed not useful in this context, we can change the behaviour to just launch ourselves when we know we're in this setting.
(this was implemented here: https://github.com/ipfs/go-ipfs/pull/6646) without the need for users to know about our arguments ahead of time, and without needing them to create a launcher themselves.

![I shouldn't know this](https://user-images.githubusercontent.com/13862850/65426903-2c2b6e00-dddf-11e9-8e06-ad2f52f2528f.png)

kpp:
Still #6646 is hacky.

> However, since the case of running ipfs without any arguments is already deemed not useful in this context, we can change the behaviour to just launch ourselves when we know we're in this setting.

This will add a different behavior based on how you launch ipfs.exe. You may receive several issues like "why ipfs does not start daemon when running from cmd.exe"?

[...]

Issue #1897, opened 2015-10-25
Daemon doesn't handle windows WM_CLOSE


MichaelMure:
I want to embed a daemon instance in a Qt GUI and I've some trouble to properly terminate the daemon process.

As stated in the QProcess terminate() function:

> On Windows, terminate() posts a WM_CLOSE message to all toplevel windows of the process and then to the main thread of the process itself. On Unix and OS X the SIGTERM signal is sent.

What happen is that the daemon doesn't handle this signal and keep running, to be lately abrubtly killed and leaving the repo with a repo.lock file.

I'm not sure about the status of signals on windows and what should be done here. Any idea ?

whyrusleeping:
good catch, i think we should be using `os.Interrupt` instead of `os.SIGINT` in our interupt handler

MichaelMure:
I tried the pull request branch and it doesn't work for me. The daemon keep running as before.

I also tried to use the windows API from the C++ side without more success:
```
GenerateConsoleCtrlEvent(CTRL_C_EVENT, daemon_process_->processId());
GenerateConsoleCtrlEvent(CTRL_BREAK_EVENT, daemon_process_->processId());
```

(insert rant about unreadable api docs here)

djdv:
This can be closed when the builds switch to Go 1.14.
See: https://github.com/ipfs/go-ipfs/pull/6808 (includes a patch for <= 1.13 if needed)

Stebalien:
I'm planning on building the next version with go 1.14 anyways. Thanks for tracking this!

MichaelMure:
> I'm planning on building the next version with go 1.14 anyways. Thanks for tracking this!

Just a reminder: https://github.com/libp2p/go-libp2p-kad-dht/issues/453#issuecomment-591958668

Stebalien:
The query logic has been rewritten from scratch (in a testing branch for now) so we'll have to re-test that once that stabilizes.

PR #6646, opened 2019-09-15
cmd: ipfs handle GUI environment on Windows

djdv:
Resolves: https://github.com/ipfs/go-ipfs/issues/6633

Hacky but maybe acceptable.

If we're not launched from a terminal, we launch the daemon by default.
If we're inside of a terminal, we print help.

https://youtu.be/diam0dQaBdw
(for some reason my cursor didn't get recorded)

### DIFF
--- a/cmd/ipfs/main.go
+++ b/cmd/ipfs/main.go
@@ -22,10 +22,10 @@ import (
 	repo "github.com/ipfs/go-ipfs/repo"
 	fsrepo "github.com/ipfs/go-ipfs/repo/fsrepo"
 
-	"github.com/ipfs/go-ipfs-cmds"
+	cmds "github.com/ipfs/go-ipfs-cmds"
 	"github.com/ipfs/go-ipfs-cmds/cli"
 	cmdhttp "github.com/ipfs/go-ipfs-cmds/http"
-	"github.com/ipfs/go-ipfs-config"
+	config "github.com/ipfs/go-ipfs-config"
 	u "github.com/ipfs/go-ipfs-util"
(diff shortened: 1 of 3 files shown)


djdv:
I'm thinking this should also pass `--init` along with daemon though.
Otherwise it's still going to launch a ghost window on a fresh install.

Opinions?

Stebalien:
Nice! I'm slightly concerned that we'll log something on init and break this but I can't think of a good way to fix that.

> I'm thinking this should also pass --init along with daemon though. Otherwise it's still going to launch a ghost window on a fresh install. Opinions?

:+1:. This makes sense. We've been discussing making this the default anyways.

djdv:
Force pushed over this.
Same as before, but now with `--init` appended as well.

djdv:
We should probably explicitly block on https://github.com/ipfs/go-ipfs/issues/1897
It seems like Golangs signal library doesn't send us anything on a `WM_CLOSE` and instead just kills us immediately if the user sends alt-f4 or clicks the window X. 
:^(

I'll have to see if there's a way to get notified of these events from the OS. Otherwise we can't gracefully shutdown.

djdv:
>It seems like Golang's signal library doesn't send us anything on a WM_CLOSE and instead just kills us immediately if the user sends alt-f4 or clicks the window X.

This is no longer the case in Go 1.14.
It's worth noting that we receive the shutdown signal but this is not a guarantee we'll shutdown gracefully. The OS will still decide to stop the process if shutdown takes longer than its grace period.
(Typically 5 seconds).

Stebalien:
:heart:

Issue #6656, opened 2019-09-21
Offline ipns record lives for 1m only 


kpp:
#### Version information:
go-ipfs version: 0.4.22-
Repo version: 7
System version: amd64/linux
Golang version: go1.12.7

#### Description:

Start `ipfs daemon` with a default config, call `ipfs cat /ipns...` to resolve & cache the content, turn off the internet connection.
The /ipns/ link will be available on the local machine for **1 minute only**. After that minute you will not be able to neither cat nor resolve the link by /ipns path (Error: failed to find any peer in table), however the data is still available through a resolved and cached /ipfs/hash.

postables:
That's due to https://github.com/ipfs/go-ipfs/blob/master/namesys/namesys.go#L51

kpp:
Nice!

kpp:
Should I use `ipfs name publish` with `--lifetime` or with `--ttl`? Will ipfs name publish overwrite the existent key on other nodes if they got connection to the publisher?

Stebalien:
This is working as intended. If the node is offline, we have no way to know if the IPNS record is up-to-date so we return an error after the TTL expires. Note: The TTL is _not_ the same as the EOL. The EOL is when the record is no longer valid, the TTL is how long we keep the record before we go to the network to look for a new one.

Now, that doesn't mean this current model is necessarily _correct_. We really need to better support offline/local IPNS. However, we need to communicate how "fresh" the response is to the user.

postables:
That seems incredibly inefficient. If the record is going to be around for 24 hours by default why is the daemon being forced to re-resolve the record after only 1 minute? 

If the user published it and it's valid for 24 hours by default they shouldn't be searching the network after only 1 minute.

Stebalien:
1. The EOL is how long the record is "valid". It doesn't mean that it's fresh. The EOL is just there as a cutoff to indicate when the record is so stale it should no longer be considered valid.
2. The TTL is how long we cache the record before trying to find a fresh one.

Someone can publish a new record _long_ before the record expires. If we cache records until the EOL, we won't find these new records.

postables:
Fair points, but isn't the TTL also defaulted to 24 hours? IPNS resolution takes ages and makes it basically impractical to use for anything. If we were to cache records like this for the duration of their TTL and not for a minute it would go a long ways to improving the usability of IPNS which for all intents and purposes is virtually unusable.

[...]

PR #6657, opened 2019-09-21
Fix Resolver Cache TTL Check (Issue: 6556)

postables:
Fixes https://github.com/ipfs/go-ipfs/issues/6656


kpp:
One reason the test failed in CI could be https://github.com/ipfs/go-ipfs/issues/6663

postables:
Why is it a 1min cache? If the record is valid for longer than 1 minute it should be in the cache for the duration of it's validity.

Only storing it for 1 minute leads to a ton of excess work being done searching for a record

kpp:
Still `DefaultResolverCacheTTL` should be `MaxResolverCacheTTL`

Stebalien:
@postables for anyone else visiting this PR, the discussion continues in https://github.com/ipfs/go-ipfs/issues/6656.

@kpp. It's the default, not the max. https://github.com/ipfs/go-ipfs/issues/6656#issuecomment-534252128 is definitely a bug.

PR #6672, opened 2019-09-24
plugins: support Close() for Tracer plugins as well

MichaelMure:
Most of the tracers available need to properly close to send the remaining traces before the process exit.

### DIFF
--- a/plugin/loader/loader.go
+++ b/plugin/loader/loader.go
@@ -2,6 +2,7 @@ package loader
 
 import (
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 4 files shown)


MichaelMure:
CI failure is unrelated.

plugin/loader/loader.go
```diff
@@ -266,7 +266,7 @@ func (loader *PluginLoader) Close() error {
 	started := loader.started
 	loader.started = nil
 	for _, pl := range started {
-		if pl, ok := pl.(plugin.PluginDaemon); ok {
+		if pl, ok := pl.(plugin.Closer); ok {
```
Stebalien: `plugin.Closer` feels a bit overkill. What about just type-asserting `pl.(io.Closer)`?

MichaelMure: I'm ok with that if you are, I just though the semantic was a bit off (it's not really an I/O).

MichaelMure:
Fixed as your were suggesting.

plugin/loader/loader.go
Stebalien: `io.Closer` is really just the standard "closer" at this point.

PR #6612, opened 2019-08-29
add experimental daemon plugin for 9P2000.L support

djdv:
Part of: https://github.com/ipfs/package-managers/issues/74

Draft TODO:

- [ ] Needs tests
    - test the interfaces directly (`go test`)
    - maybe blackbox tests (sharness, `fsx`, etc.)
- [x] `go-ipfs/plugin` should support plugin scoped configs (we can punt on this) 
    - We want to remove the `go-ipfs-config` dependency in our plugin, and instead, receive some context from `go-ipfs/plugin`
    - e.g. `var cfg Config.FileSystem = findCfgPathAndLoad(myNonStandardPath, &myWhateverFormat, ...` could be avoided by changing `Plugin.Init()` to something like `Plugin.Init(context, opts...)`
    - Resolved via https://github.com/ipfs/go-ipfs/pull/6613
- [ ] ~Trying to enable the experiment through `ipfs config --json Experimental.FileSystemEnabled true` should work~
    - We can simply remove the plugin from the preload list, and document how to add it in `experimental-features.md` 🠗
- [ ] Documentation
    - package / code
    - experimental-features.md

Other known issues:
- ~`go.mod` has replace directives for this patchset, these need to point to the final dependency commits before merging~
- ~Directory reads in particular need to be fixed before merging. (currently reads entire directory every request and returns requested slice instead of using a forward-only stream)~

PR #6675, opened 2019-09-25
add systemd support

Stebalien:
This change:

1. adds socket activation support for the API and the gateway
2. notifies systemd when go-ipfs is ready
3. notifies systemd when go-ipfs is shutting down

Unlike my previous attempt (https://github.com/ipfs/go-ipfs/pull/5514), this change _merges_ the sockets passed by systemd with those configured in `ipfs config`. This is likely what the user wants.

Motivation:

1. Change 1 makes it possible to start go-ipfs on demand.
2. Change 2 makes it possible to _wait_ for go-ipfs to finish starting. This makes it possible to use systemd's automount feature along with https://github.com/ipfs/go-ipfs/pull/6612 to auto-start IPFS and auto-mount /ipfs on demand.


hsanjuan:
Would you consider putting the sockets part as a separate, re-usable module out of the ipfs codebase? Seems quite useful.

Stebalien:
@hsanjuan done.

olizilla:
@Stebalien could you merge these systemd improvements into the stabalize-dht branch so we can try them on a gateway node pls?

Stebalien:
@olizilla I've rebased the stabilize branch. However, that branch now requires 2048 bit keys so we'll want to wait until we've updated the bootstrappers (or add a patch to that branch to revert it).

PR #6685, opened 2019-09-27
Pinning interop: Pin ls returns appropriate zero value

aschmahmann:


aschmahmann:
We don't have any streaming interop tests so I'm not sure if we should remove the `json:",omitempty"` from `PinLsObject` as well.

Stebalien:
Ugh. We _want_ to omit `PinLsList.Keys` if we're streaming.

What if we set `Keys` to a non-nil empty map? That will cause us to send the correct `{}`.

Stebalien:
When _not_ streaming, that is.

Stebalien:
Wait, that's what we're doing.

Stebalien:
Ah. I didn't test this the right way. omitempty doesn't care about nil versus non-nil empty.

PR #6694, opened 2019-10-01
docs: link remaining docs available, guide people to the right locations

daviddias:
Not a big change, mostly me doing some work figuring out how these fit all together and with the rest of docs available

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,11 +1,25 @@
-# Documentation and Guides
+# Developer Documentation and Guides
+
+If you are looking for User Documentation & Guides, please visit [docs.ipfs.io](https://docs.ipfs.io/).
 
 If you’re experiencing an issue with IPFS, **please follow [our issue guide](github-issue-guide.md) when filing an issue!**
 
 Otherwise, check out the following guides to using and developing IPFS:
 
+## Developing `go-ipfs`
 
-## General Usage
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -14,21 +28,7 @@ Otherwise, check out the following guides to using and developing IPFS:
...
-
-
-## Developing `go-ipfs`
-
-- Building on…
-    - [Windows](windows.md)
-- [Performance Debugging Guidelines](debug-guide.md)
-- [Release Checklist](releases.md)
-
+- [Setting up an IPFS Gateway](https://github.com/ipfs/go-ipfs/blob/master/docs/gateway.md)
```
daviddias: @ipfs/docs found these docs inside the go-ipfs repo. Seem that they should make their way into docs.ipfs.io as well, at least at some point.

Stebalien:
With respect to consolidation, I agree this should go in ipfs/docs. Really, almost (all?) docs should go there.

However, we do need to be careful to signal platform specific features (e.g., go-ipfs supports `curl http://gateway/ipfs/QmFoo?filename=some_filename`).

docs/README.md
jessicaschilling: @daviddias See https://github.com/ipfs/docs/pull/303 😊

PR #6696, opened 2019-10-02
[DOCS] Add note about IPNS and DNSLink

parkan:
Using a DNS alias to point to a gateway is not well documented outside the dnslink.io site, this adds a reference to those docs to the main go-ipfs documentation.

### DIFF
--- a/docs/gateway.md
+++ b/docs/gateway.md
@@ -43,6 +43,18 @@ your query string to explicitly specify the filename. For example:
 
 > https://ipfs.io/ipfs/QmfM2r8seH2GiRaC4esTjeraXEachRt8ZsSeGaWTPLyMoG?filename=hello_world.txt
 
+## IPNS & DNSLink
+
+The gateway supports resolving IPNS and DNSLink names:
+
+```
+http://127.0.0.1:8080/ipns/Qm...
+http://127.0.0.1:8080/ipns/libp2p.io
+```
+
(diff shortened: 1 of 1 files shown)


docs/gateway.md
```diff
@@ -43,6 +43,18 @@ your query string to explicitly specify the filename. For example:
 
 > https://ipfs.io/ipfs/QmfM2r8seH2GiRaC4esTjeraXEachRt8ZsSeGaWTPLyMoG?filename=hello_world.txt
 
+## IPNS & DNSLink
```
Stebalien: This doesn't seem like the right place for this documentation. This document is about gateway features and these are just general IPFS features.

Maybe we should add  sentence to the top saying that the gateway provides access to files/directories served on IPFS and then briefly describe `/ipns/...` and `/ipfs/...`?

parkan:
that's fair, the main thing that I want to point out is the fact that the gateway can accept the ipns/dnslink name via Host header instead of url parameter (or subdomain), which doesn't seem well documented other than on the dnslink site

happy to reword this with more focus on that or move elsewhere

Stebalien:
> that's fair, the main thing that I want to point out is the  fact that the gateway can accept the ipns/dnslink name via Host header  instead of url parameter (or subdomain), which doesn't seem well  documented other than on the dnslink site

Ah, that makes a lot of sense.

Stebalien:
https://github.com/ipfs/go-ipfs/pull/6767

PR #6692, opened 2019-10-01
docs: fix broken url

daviddias:
image

Not sure if pointing to https://docs.ipfs.io/guides/guides/install/ is the "getting started guide" we want (or that we had). @ipfs/docs what is the best/new getting started guide now that the previous url is broken?

PR #6697, opened 2019-10-02
Update README.md

jessicaschilling:
Change first link under "Getting Started" to more specific docs per https://github.com/ipfs/go-ipfs/pull/6692#issuecomment-537110811
Ref https://github.com/ipfs/go-ipfs/pull/6692


jessicaschilling:
@Stebalien -- do you mind having a glance? Thank you!

PR #6700, opened 2019-10-07
fix(resolve): correctly handle .eth domains

### LINKED ISSUE #6699: gateway doesn’t load .eth domains from Host: header
#### Version information:
master

#### Description: 
the changes in https://github.com/ipfs/go-ipfs/pull/6448 should have allowed us to resolve .eth domains even though it’s not a “real” TLD, but this still fails when the name is passed in `Host:` header

#### Works

```sh
$ curl 35.241.8.114/ipns/almonit.eth/
```

### Does Not Work

```sh
$ curl 35.241.8.114 --header “Host:almonit.eth”
```

as discussed w @Stebalien, would be A+ to fix this before DevCon starts properly

Stebalien:
This should have been handled down inside the DNSLink resolver. Otherwise, we'll
break any name happens to end in `.eth`.

also fixes #6699

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -52,7 +52,7 @@ require (
 	github.com/ipfs/go-unixfs v0.2.1
 	github.com/ipfs/go-verifcid v0.0.1
 	github.com/ipfs/interface-go-ipfs-core v0.2.3
-	github.com/jbenet/go-is-domain v1.0.2
+	github.com/jbenet/go-is-domain v1.0.3
 	github.com/jbenet/go-random v0.0.0-20190219211222-123a90aedc0c
 	github.com/jbenet/go-temp-err-catcher v0.0.0-20150120210811-aac704a3f4f2
 	github.com/jbenet/goprocess v0.1.3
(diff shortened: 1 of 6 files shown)


aschmahmann:
@Stebalien can you clarify if my understanding is correct.

The fix to #6699 is to use the updated `go-is-domain` library that accounts for `ETH` as a tld extension which deals with a bug at https://github.com/ipfs/go-ipfs/blob/d9f2bafa45693db2388544097089b3c7a5182110/core/corehttp/ipns_hostname.go#L27

Moving the `.eth -> .eth.link` code into the DNSLink resolver from the namesys resolver is for sanity/correctness, but is unrelated to #6699.

If the above is correct why are the tests being removed, the bug was in the HTTP API not in Namesys?

> we'll break any name happens to end in .eth

Sure, although we don't currently have any resolver other than DNSLink that can end in .eth, right?

Stebalien:
> Moving the .eth -> .eth.link code into the DNSLink resolver from the namesys resolver is for sanity/correctness, but is unrelated to #6699.

Yes.

> If the above is correct why are the tests being removed, the bug was in the HTTP API not in Namesys?

Those tests used a fake DNSLink resolver (bypassing the real dnslink resolver). However, thanks for calling me on that. I've added the tests to the _real_ DNSLink resolver's test suite.

Stebalien:
> Sure, although we don't currently have any resolver other than DNSLink that can end in .eth, right?

/ipfs/QmFoo/foo.eth? I think?

Stebalien:
Ok, the API tests are stalling on master: https://github.com/ipfs/go-ipfs/issues/6703

PR #6705, opened 2019-10-08
Pin ls traverses all indirect pins

### LINKED ISSUE #6527: Pin ls stops traversing recursive pins after hitting a direct pin
We should avoid re-emitting these directly pinned blocks but we still need to continue traversing the tree. This needs a _test_ to be considered fixed.

Regression from #6493.

aschmahmann:
1) Fixes #6527 
2) Deduplicates most of the pin ls code between the CoreAPI version and the commands version (the method is not optimal, but hopefully the CoreAPI will get a streaming option for `pin ls` soon and then this will be easy)
3) Introduces a precedence change for pin ls. A direct/recursive pin is now labeled as such even if also indirectly pinned. The previous precedence ordering was recursive > indirect > direct, it is now recursive/direct > indirect.

Changes 1 and 3 need to be tested, and I will submit a PR for a CoreAPI test (if we want a sharness test we can do that as well, but it seems like overkill). Since the CoreAPI tests all live in interface-go-ipfs-core this test should probably be there as well.

@Stebalien


aschmahmann:
Ideally we should merge the tests (ipfs/interface-go-ipfs-core#47) before merging this so that we can reference the tests from go-ipfs. However, I'm not picky on ordering here as long as the tests actually end up running on the CoreAPI code.

core/commands/pin.go
```diff
@@ -475,69 +477,46 @@ func pinLsKeys(req *cmds.Request, typeStr string, n *core.IpfsNode, api coreifac
...
 
-	keys := cid.NewSet()
-
-	AddToResultKeys := func(keyList []cid.Cid, typeStr string) error {
-		for _, c := range keyList {
-			if keys.Visit(c) {
-				err := emit(&PinLsOutputWrapper{
+	procErr := make(chan error, 1)
+	ctx := req.Context
+	go func() {
```
Stebalien: This goroutine is unnecessary.

core/coreapi/pin.go
```diff
@@ -191,45 +192,104 @@ func (p *pinInfo) Type() string {
...
-	keys := make(map[cid.Cid]*pinInfo)
-
-	AddToResultKeys := func(keyList []cid.Cid, typeStr string) {
-		for _, c := range keyList {
-			keys[c] = &pinInfo{
-				pinType: typeStr,
-				path:    path.IpldPath(c),
+	var pins []coreiface.Pin
+	ctxError := make(chan error, 1)
+	go func() {
```
Stebalien: This goroutine is unnecessary.

core/commands/pin.go
```diff
@@ -475,72 +477,40 @@ func pinLsKeys(req *cmds.Request, typeStr string, n *core.IpfsNode, api coreifac
...
-				if err != nil {
-					return err
-				}
+	procErr := make(chan error, 1)
+	ctx := req.Context
+	for {
+		select {
+		case p, ok := <-pinCh:
+			if !ok {
+				procErr <- nil
```
Stebalien: Unnecessary channel?

core/coreapi/pin.go
```diff
@@ -191,45 +192,96 @@ func (p *pinInfo) Type() string {
...
-		for _, k := range api.pinning.RecursiveKeys() {
-			err := merkledag.Walk(
-				ctx, merkledag.GetLinksWithDAG(api.dag), k,
-				set.Visit,
-				merkledag.SkipRoot(), merkledag.Concurrent(),
-			)
-			if err != nil {
-				return nil, err
+	go func() {
+		defer close(ch)
```
Stebalien: Given the way this function is currently called, this can hide errors: we can observe this channel closing before we read the error off the error channel.

It's generally safer to just use a "result" channel. That makes it easier to call this function as the `for x := range myCh` syntax can be used.

Stebalien: Alternatively, we can close both the error and normal channels and, instead of listening on the error channel, run `err := <-errCh` after finishing the _normal_ loop.

core/coreapi/pin.go
```diff
@@ -191,45 +192,96 @@ func (p *pinInfo) Type() string {
...
+				for _, k := range pin.RecursiveKeys() {
+					keys.Add(k)
+				}
+			}
+			indirectKeys := cid.NewSet()
+			for _, k := range pin.RecursiveKeys() {
+				err := merkledag.Walk(ctx, merkledag.GetLinksWithDAG(dag), k, func(c cid.Cid) bool {
+					r := indirectKeys.Visit(c)
+					if r && keys.Visit(c) {
+						ch <- &pinInfo{
```
Stebalien: This can block.

Stebalien:
@aschmahmann bump (when you get a chance).

core/coreapi/pin.go
aschmahmann: I think I did what you asked here, lmk.

core/coreapi/pin.go
aschmahmann: now checks on context

core/commands/pin.go
aschmahmann: yep, I think a mistake from a some git mistakes. thx for the catch.

aschmahmann:
@Stebalien I think we should just squash these commits when we're ready for merge. Lmk and we can finish any last issues with this PR.

core/coreapi/pin.go
```diff
@@ -194,57 +195,119 @@ func (p *pinInfo) Type() string {
...
+			if err != nil {
+				errCh <- err
+				return
+			}
 
-	out := make([]coreiface.Pin, 0, len(keys))
-	for _, v := range keys {
-		out = append(out, v)
-	}
+			if typeStr != "all" {
```
Stebalien: This needs a comment explaining _why_ we're doing this.

Stebalien: I also inverted the test to `if typeStr == "indirect"` to make it clear that we're skipping this because we only care about indirect pins.

PR #6708, opened 2019-10-08
fix(pin): wait till after fetching to remove direct pin

### LINKED ISSUE #4650: Recursive pinning can lose existing direct pin
#### Description:

This bit looks like it'll lose the existing direct pin if the attempt to add the recursive pin fails. The direct pin will have been removed, and the expected recursive pin won't have succeeded: https://github.com/ipfs/go-ipfs/blob/864c960aa641bff2820b59b5eeb9d93be5d0998d/pin/pin.go#L213-L228

`directPin.Remove()` should probably happen after `recursePin.Add()`

Stebalien:
Otherwise, we could abort while fetching the graph and stay in a state where the direct pin is removed.

fixes #4650

### DIFF
--- a/pin/pin.go
+++ b/pin/pin.go
@@ -225,9 +225,6 @@ func (p *pinner) Pin(ctx context.Context, node ipld.Node, recurse bool) error {
 			return nil
 		}
 
-		if p.directPin.Has(c) {
-			p.directPin.Remove(c)
-		}
 		p.lock.Unlock()
 		// fetch entire graph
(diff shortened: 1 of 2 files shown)


hsanjuan:
I'm assigned as reviewer mostly for info-purposes right?

hsanjuan:
I'm assigned as reviewer mostly for info-purposes right?

Stebalien:
No, sorry, that's just the bot. However, I wouldn't mind an ex post facto review.

PR #6715, opened 2019-10-17
pin: add context and error return to most of the Pinner functions

MichaelMure:
At Infura we have the need for a custom out-of-process (remote actually) `Pinner`. Problem is, at the moment this interface assume that the list of pins is loaded in memory and access are instant and can't fail.

This PR add a context for cancellation and a return error to most of the functions of `Pinner`. Almost everything else is a consequence of those changes (see my comment later for the excessive locking).

This PR use a forked `go-ipfs-provider` to follow those changes, PR at https://github.com/ipfs/go-ipfs-provider/pull/16

### DIFF
--- a/pin/pin.go
+++ b/pin/pin.go
@@ -281,7 +281,9 @@ func (p *pinner) isInternalPin(c cid.Cid) bool {
 // IsPinned returns whether or not the given key is pinned
 // and an explanation of why its pinned
 func (p *pinner) IsPinned(ctx context.Context, c cid.Cid) (string, bool, error) {
-	return p.isPinnedWithType(c, Any)
+	p.lock.RLock()
+	defer p.lock.RUnlock()
+	return p.isPinnedWithType(ctx, c, Any)
 }
 
 // IsPinnedWithType returns whether or not the given cid is pinned with the
@@ -289,12 +291,12 @@ func (p *pinner) IsPinned(ctx context.Context, c cid.Cid) (string, bool, error) 
 func (p *pinner) IsPinnedWithType(ctx context.Context, c cid.Cid, mode Mode) (string, bool, error) {
 	p.lock.RLock()
 	defer p.lock.RUnlock()
-	return p.isPinnedWithType(c, mode)
+	return p.isPinnedWithType(ctx, c, mode)
 }
 
 // isPinnedWithType is the implementation of IsPinnedWithType that does not lock.
 // intended for use by other pinned methods that already take locks
-func (p *pinner) isPinnedWithType(c cid.Cid, mode Mode) (string, bool, error) {
+func (p *pinner) isPinnedWithType(ctx context.Context, c cid.Cid, mode Mode) (string, bool, error) {
 	switch mode {
 	case Any, Direct, Indirect, Recursive, Internal:
 	default:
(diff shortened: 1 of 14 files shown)


pin/pin.go
```diff
@@ -211,22 +211,23 @@ func NewPinner(dstore ds.Datastore, serv, internal ipld.DAGService) Pinner {
...
-	p.lock.Lock()
-	defer p.lock.Unlock()
 	err := p.dserv.Add(ctx, node)
 	if err != nil {
 		return err
 	}
 
 	c := node.Cid()
 
+	p.lock.Lock()
```
MichaelMure: This seemed like an excessive locking to me. Unless I'm missing something, no need to lock the pin lists while writing content to the datastore.

MichaelMure:
CI failure doesn't seems related.

MichaelMure:
@Stebalien could you have a look ?

pin/pin.go
Stebalien: I agree. There was a lot of confusion about which locks had to be held at any given time for a while (we were using the pin lock to guard things that should have been guarded by the GC lock).

pin/pin.go
```diff
@@ -243,13 +244,6 @@ func (p *pinner) Pin(ctx context.Context, node ipld.Node, recurse bool) error {
 
 		p.recursePin.Add(c)
 	} else {
-		p.lock.Unlock()
-		_, err := p.dserv.Get(ctx, c)
-		p.lock.Lock()
-		if err != nil {
-			return err
-		}
-
```
Stebalien: :+1:

pin/pin.go
```diff
@@ -286,15 +280,13 @@ func (p *pinner) isInternalPin(c cid.Cid) bool {
 
 // IsPinned returns whether or not the given key is pinned
 // and an explanation of why its pinned
-func (p *pinner) IsPinned(c cid.Cid) (string, bool, error) {
-	p.lock.RLock()
-	defer p.lock.RUnlock()
```
Stebalien: ?

pin/pin.go
```diff
@@ -347,7 +339,7 @@ func (p *pinner) isPinnedWithType(c cid.Cid, mode Mode) (string, bool, error) {
 
 // CheckIfPinned Checks if a set of keys are pinned, more efficient than
 // calling IsPinned for each key, returns the pinned status of cid(s)
-func (p *pinner) CheckIfPinned(cids ...cid.Cid) ([]Pinned, error) {
```
Stebalien: So... can we feed this shiny new context through below (see the `context.TODO()`).

MichaelMure: Ha yes, will do.

pin/pin.go
MichaelMure: ho shoo, I thought it was cascading with `IsPinnedWithType(ctx context.Context, c cid.Cid, mode Mode)` which already has a lock.

MichaelMure:
Alright, PR comments addressed and rebased.

I connected an extra context that was just waiting for it.

The last thing would be to edit go.mod once https://github.com/ipfs/go-ipfs-provider/pull/16 is merged.

Stebalien:
LGTM! I've merged and bubbled the provider dep change. I'll merge once CI passes.

Stebalien:
:rocket:

Thanks!

PR #6719, opened 2019-10-21
introduce IpfsNode Plugin

MichaelMure:
This PR introduce a new kind of plugin able to directly use `IpfsNode`. As `IpfsNode` is considered internal, no guarantee is made about API stability. This is useful when a plugin needs a direct access to the internal and using `CoreApi` is not enough.

As introducing `PluginNode` make the naming a little akward, this PR also introduce a copy of `PluginDaemon` named `PluginApi` and deprecate `PluginDaemon`. As it is a separate commit, we can remove that easily if you don't like it.


MichaelMure:
`golint` fail because `PluginDaemon` is now deprecated. Other CI failures seems unrelated.

MichaelMure:
Updated as you wanted.

Stebalien:
Thanks!

PR #6727, opened 2019-10-25
Add high-level go-ipfs architecture diagram

momack2:


README.md
```diff
@@ -431,6 +431,10 @@ Some places to get you started on the codebase:
   - PubSub: https://github.com/libp2p/go-libp2p-pubsub
 - [IPFS : The `Add` command demystified](https://github.com/ipfs/go-ipfs/tree/master/docs/add-code-flow.md)
 
+### Map of IPFS Subsystems
+**WIP**: This is a high-level architecture diagram of the various sub-systems of go-ipfs. To be updated with how they interact. Anyone who has suggestions is welcome to comment [here](https://docs.google.com/drawings/d/1OVpBT2q-NtSJqlPX3buvjYhOnWfdzb85YEsM_njesME/edit) on how we can improve this!
```
daviddias: This is very similar to https://github.com/ipfs/js-ipfs/tree/master/examples#understanding-the-ipfs-stack, any reason to move to Google Docs rather than iterating on the previous frame?

momack2:
Thanks for the feedback, @ericronne! Updated the color palette to match the brand book (excepting the libp2p part, which is intentionally a libp2p color), and also make the stylistic updates re outlines and clarity of what each arrow color means. LMK if that seems better!

I'd be happy to do a walkthrough - but also am transcribing a lot of knowledge from Steven/Jeromy into this doc that I'm not the best person to explain. I can give a sense as to purpose/audience though - which is folks who are onboarding on IPFS as deep users and hopefully future contributors who want to grok how the pieces fit together and make sense of all the different repos/libraries that make up the sub-areas of IPFS (so they can hopefully help improve them!)

README.md
daviddias: The tool used is https://monodraw.helftone.com and the source file is https://github.com/ipfs/js-ipfs/tree/master/img. However, ultimately it is just an ASCII drawing which makes it super cool for READMEs and print.

Tool aside, I think what is most important is to have an accurate description of the subsystems that is shared and recognized across the project, rather than multiple that show an incomplete picture.

For reference, here are the ones that exist:

- The IPFS Architecture https://github.com/ipfs/js-ipfs#ipfs-architecture
- How the CLI, Client Library, HTTP API and IPFS Core interface with each other https://github.com/ipfs/js-ipfs#code-architecture-and-folder-structure
- Bitswap https://github.com/ipfs/js-ipfs-bitswap#structure

ericronne:
That feels much easier to follow @momack2 👍 
I forked it just now and made some text semibold, which might help key elements stand out (also tweaked a couple alignment issues). [Copy if you like](https://docs.google.com/drawings/d/19-ajkrI-4te_zLhM16ZYeJ5dLmihvJu8BANHOo1NB54/edit?usp=sharing). Since the ability to edit online is important, your approach feels like a good compromise. It would be nice to create a template from scratch at some point.

README.md
momack2: I disagree - I think having multiple different diagrams focusing on different parts or trying to convey different messages makes a lot of sense. One focuses on folder structure, one focuses on general context for how repos fit together, another deep dives on a specific module, and one is about tracing a call stack. All of these are valuable knowledge sharing tools and they would lose value if merged into one master diagram. If anything, I want to pull out some areas that don't clearly belong in this diagram to narrow it's purpose - so definitely see it coexisting with others, esp on the js side.

momack2: do LMK if there is a specific area where this drawing is inaccurate in it's current state - happy to remedy that!

momack2:
Thanks Eric! Merged those in (and also deleted modules unused by our add / get / request pathways. I think the main communication point here is "how do the systems fit together for basic requests" - so wanted to narrow in on that aim.

README.md
daviddias: To clarify, I believe that we are in agreement in which:
- 👍🏽 to having multiple diagram that focus on different specific parts of the stack (i.e. subsystems or interactions)
- 👎🏽 on having multiple diagrams called IPFS Architecture and having them communicate information that is in conflict
- 👎🏽 on having a diagram that focuses on a subsystem that introduces something that is not coherent with the overall picture

Did I get it write?

momack2:
Agree that we shouldn’t have any diagrams that communicate incorrect or
misleading information. I’d point out that the js-ipfs diagram is for a
different stack / system - and therefore having two different architecture
diagrams that diverge for each notably different implementation can make
sense. As long as all diagrams are correct and not misleading then I think
having multiple is fine (especially across different repos / systems)! :)

README.md
Stebalien: We can just call this the Go-IPFS architecture if you'd like. This diagram _is_ consistent with the go-ipfs architecture.

README.md
```diff
@@ -431,6 +431,10 @@ Some places to get you started on the codebase:
   - PubSub: https://github.com/libp2p/go-libp2p-pubsub
 - [IPFS : The `Add` command demystified](https://github.com/ipfs/go-ipfs/tree/master/docs/add-code-flow.md)
 
+### Map of IPFS Subsystems
```
momack2:
```suggestion
### Map of go-ipfs Subsystems
```

README.md
momack2: cool - changed the title to specify go-ipfs

momack2:
Can has approving review? I _can_ merge without it but want to model good contributor practice. 😜

momack2:
This is when you're getting a request from another node for data you have (via one of the various libp2p methods like pubsub/DHT). If there's a better name I'm happy to add it!

PR #6732, opened 2019-10-29
remove deprecated log.Event calls

lanzafame:
Beginning PR as part of a piece of work to improve the logging in go-ipfs. This removes the deprecated `log.Event` calls. These will be replaced with distributed tracing in a later PR.


Stebalien:
We removed these events when updating deps.

PR #6743, opened 2019-11-02
fix #2203: omit the charset attribute when Content-Type is text/html

### LINKED ISSUE #2203: Including charset in HTTP breaks non UTF-8 sites
Charset returned in HTTP header overrides charset in meta tag but is essential for text files and other content.

Encoding is broken on some sites like: https://ipfs.io/ipfs/QmYJxHZ5MeqKF5vbp8Wei71fNkBPdBSB3CRYSW75fA8yFE/ that use non UTF-8 encoding.

Solution would be to run simple charset meta tag detection. It could be heuristic for sake of speed.

Also good caching would be great.

dreamski21:
fixes #2203

### DIFF
--- a/core/corehttp/gateway_handler.go
+++ b/core/corehttp/gateway_handler.go
@@ -383,7 +383,10 @@ func (i *gatewayHandler) serveFile(w http.ResponseWriter, req *http.Request, nam
 		}
 	}
 
-	mime := http.DetectContentType(content)
+	buf := make([]byte, 512)
+	_, _ = content.Read(buf)
+	mime := http.DetectContentType(buf)
+	
 	if strings.HasPrefix(mime, "text/html;") {
 		mime = "text/html"
 	}
--- a/core/corehttp/gateway_handler.go
+++ b/core/corehttp/gateway_handler.go
@@ -11,7 +11,8 @@ import (
 	"runtime/debug"
 	"strings"
 	"time"
-
+	"mime"
+	
 	"github.com/dustin/go-humanize"
 	"github.com/ipfs/go-cid"
 	files "github.com/ipfs/go-ipfs-files"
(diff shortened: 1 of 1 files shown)


dreamski21:
Grrr... I PR'ed the wrong code. I have actually converted content from `io.ReadSeeker` to `[]byte` as in `http.ServeContent`.

It was ugly and feels redundant. Should I recommit it or is there a better way to convert to `[]byte` or detect the content type?

core/corehttp/gateway_handler.go
```diff
@@ -383,6 +383,15 @@ func (i *gatewayHandler) serveFile(w http.ResponseWriter, req *http.Request, nam
 		}
 	}
 
+	buf := make([]byte, 512)
+	_, _ = content.Read(buf)
```
Stebalien: This is the only way to do this but we should be calling `io.ReadFull`. Otherwise, we could read a single byte.

core/corehttp/gateway_handler.go
```diff
@@ -383,6 +383,15 @@ func (i *gatewayHandler) serveFile(w http.ResponseWriter, req *http.Request, nam
...
+	mime := http.DetectContentType(buf)
+	
+	if strings.HasPrefix(mime, "text/html;") {
+		mime = "text/html"
+	}
+	w.Header().Set("Content-Type", mime)
+
 	http.ServeContent(w, req, name, modtime, content)
```
Stebalien: We need to seek back to 0 before calling this.

core/corehttp/gateway_handler.go
```diff
@@ -383,6 +383,15 @@ func (i *gatewayHandler) serveFile(w http.ResponseWriter, req *http.Request, nam
...
 	}
 
+	buf := make([]byte, 512)
+	_, _ = content.Read(buf)
+	mime := http.DetectContentType(buf)
+	
+	if strings.HasPrefix(mime, "text/html;") {
+		mime = "text/html"
+	}
+	w.Header().Set("Content-Type", mime)
```
Stebalien: The file type detector inside `ServeContent` also uses the extension. We should probably do that as well.

(although ignoring the extension when there's a `?` (query) present may make sense).

core/corehttp/gateway_handler.go
dreamski21: Okay. as you say.

Maybe I'm just confused because, according to the docs, the main difference between [Read](https://golang.org/pkg/io/#Reader) and [ReadFull](https://golang.org/pkg/io/#ReadFull) seems to be that `Read` **may** read up to `len(buf)` while `ReadFull` **must** read exactly `len(buf)`.

Also, I have tested the following code\* with both `.mp4` and `.html` files and it seemed to work just fine. _Like, it was reading more than a single byte._

(\* I couldn't build/test the actual codebase for technical reasons. Sorry for the annoyance.)

```go
func main() {
	w, err := os.Open("clip.mp4")
	_ = err

	buf := make([]byte, 512)
	_, _ = w.Read(buf)
	contentType := http.DetectContentType(buf)
	fmt.Println("Type is", contentType)
	if strings.HasPrefix(contentType, "text/html;") {
		contentType = "text/html"
	}

	fmt.Println("Type is", contentType)
}
```

dreamski21:
Again, I'm horribly sorry for the annoyance. I shouldn't have opened this pull request in the first place as I have zero knowledge of Go. This won't happen again (not here, not on another project).

**You can either close this PR or just maybe fix, either way is fine. I'm just tired of it.**

 _\*Tipped IPFS 20 BATs\*_

core/corehttp/gateway_handler.go
Stebalien: > Maybe I'm just confused because, according to the docs, the main difference between Read and ReadFull seems to be that Read may read up to len(buf) while ReadFull must read exactly len(buf).

Exactly. The issue is that `Read` may _also_ not read the full buffer (e.g., it can return after reading a single byte).

> Like, it was reading more than a single byte.

It probably was, that's just not something we can rely on.

Stebalien:
This is _not_ annoying at all. Fixing small bugs like this is a great way to learn a language and start contributing to a project.

PR #6745, opened 2019-11-04
add prometheus/common/version pkg to export more version info about go-ipfs

lanzafame:
This extends the current setup and offloads it to a very [nice little package](https://github.com/prometheus/common/tree/master/version). I have kept the exported information around so as to not break any prometheus queries or dashboards that anyone has.

### DIFF
--- a/cmd/ipfs/Rules.mk
+++ b/cmd/ipfs/Rules.mk
@@ -13,7 +13,20 @@ PATH := $(realpath $(d)):$(PATH)
 # DEPS_OO_$(d) += merkledag/pb/merkledag.pb.go namesys/pb/namesys.pb.go
 # DEPS_OO_$(d) += pin/internal/pb/header.pb.go unixfs/pb/unixfs.pb.go
 
-$(d)_flags =-ldflags="-X "github.com/ipfs/go-ipfs".CurrentCommit=$(git-hash)"
+DATE := $(shell date +%FT%T%z)
+USER := $(shell whoami)
+GIT_VERSION := $(shell git --no-pager describe --tags --always --dirty)
+GIT_REVISION := $(shell git rev-parse HEAD)
+BRANCH := $(shell git branch | grep "*" | cut -d ' ' -f2)
+
+$(d)_flags =-ldflags "
+$(d)_flags := $($(d)_flags)-X github.com/ipfs/go-ipfs.CurrentCommit=$(git-hash) 
+$(d)_flags := $($(d)_flags)-X github.com/prometheus/common/version.Version=$(GIT_VERSION) 
+$(d)_flags := $($(d)_flags)-X github.com/prometheus/common/version.BuildDate=$(DATE) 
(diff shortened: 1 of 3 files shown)


olizilla:
Can you check that this works when built in docker? We currently explicitly copy specific files from the git repo into the docker build context to allow the commit info to still work when built in docker.

cmd/ipfs/Rules.mk
```diff
@@ -13,7 +13,20 @@ PATH := $(realpath $(d)):$(PATH)
...
+GIT_REVISION := $(shell git rev-parse HEAD)
+BRANCH := $(shell git branch | grep "*" | cut -d ' ' -f2)
+
+$(d)_flags =-ldflags "
+$(d)_flags := $($(d)_flags)-X github.com/ipfs/go-ipfs.CurrentCommit=$(git-hash) 
+$(d)_flags := $($(d)_flags)-X github.com/prometheus/common/version.Version=$(GIT_VERSION) 
+$(d)_flags := $($(d)_flags)-X github.com/prometheus/common/version.BuildDate=$(DATE) 
+$(d)_flags := $($(d)_flags)-X github.com/prometheus/common/version.Branch=$(BRANCH) 
+$(d)_flags := $($(d)_flags)-X github.com/prometheus/common/version.Revision=$(GIT_REVISION) 
+$(d)_flags := $($(d)_flags)-X github.com/prometheus/common/version.BuildUser=$(USER)
```
Stebalien: We shouldn't set this.

cmd/ipfs/Rules.mk
```diff
@@ -13,7 +13,20 @@ PATH := $(realpath $(d)):$(PATH)
 # DEPS_OO_$(d) += merkledag/pb/merkledag.pb.go namesys/pb/namesys.pb.go
 # DEPS_OO_$(d) += pin/internal/pb/header.pb.go unixfs/pb/unixfs.pb.go
 
-$(d)_flags =-ldflags="-X "github.com/ipfs/go-ipfs".CurrentCommit=$(git-hash)"
+DATE := $(shell date +%FT%T%z)
+USER := $(shell whoami)
+GIT_VERSION := $(shell git --no-pager describe --tags --always --dirty)
```
Stebalien: Maybe we should just set this in an init function in `version.go`. That way, we can use the official version.

cmd/ipfs/daemon.go
```diff
@@ -421,6 +422,7 @@ func daemonFunc(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment
 
 	// initialize metrics collector
 	prometheus.MustRegister(&corehttp.IpfsNodeCollector{Node: node})
+	prometheus.MustRegister(promversion.NewCollector("go_ipfs"))
```
Stebalien: Double checking: does this mean we _have_ to pass those build flags? dist.ipfs.io doesn't build using the makefile.

PR #6747, opened 2019-11-04
fix(coreapi/add): close the fake repo used when adding with hash-only

### LINKED ISSUE #6744: ipfs add with onlyhash, memery high usage
#### Version information:
go-ipfs version: 0.4.22-
Repo version: 7
System version: amd64/linux
Golang version: go1.12.7

#### Description:
The VM has 8GB of RAM.
My application will generate a lots of calls `ipfs add -n` with a speed 40 per seonds.
Firstly, the ipfs daemon just cost little memery. however, one hours later, out of memery.
If stop my application, the ipfs will return to normal after a few hours.

i want to know, is there a way to reduce recovery time? thanks!

[here is the data collected by bin/collect-profiles.sh](https://github.com/ipfs/go-ipfs/files/3802961/ipfs-profile-xzj-tmp-per03-2019-11-04T115439%2B0800.tar.gz)

Stebalien:
fixes #6744

### DIFF
--- a/core/coreapi/unixfs.go
+++ b/core/coreapi/unixfs.go
@@ -69,6 +69,7 @@ func (api *UnixfsAPI) Add(ctx context.Context, files files.Node, opts ...options
 		if err != nil {
 			return nil, err
 		}
+		defer nilnode.Close()
 		addblockstore = nilnode.Blockstore
 		exch = nilnode.Exchange
 		pinning = nilnode.Pinning


Stebalien:
Note: there are actually two issues here:

1. #6629 (fixed).
2. Closing the node. Technically, this should already _mostly_ happen when the context is canceled. However, it doesn't hurt to explicitly close the node as well.

PR #6754, opened 2019-11-08
fix bug 6748

hcg1314:
ipfs add with only hash, don't need to announce cid to other peer，because unable to get this file from IPFS network. And waste memery to store the useless cid.

when i modified this code, and close the mock repo(issue 6744)，no memery grow with large calls.


Stebalien:
Thanks!

PR #6761, opened 2019-11-12
chore(dep): update QUIC

Stebalien:
This is a large protocol breaking change. That's why QUIC is experimental.


Stebalien:
Blocked on https://github.com/libp2p/go-libp2p-quic-transport/issues/82

PR #6764, opened 2019-11-15
fix issue 6760, adding with hash-only, high CPU usage.

hcg1314:
IPFS daemon will create a IDENTITY for repo through function GenerateKeyPairWithReader. This action cost so much CPU resource. And IPFS daemon creates a fake repo for each `ipfs add -n xxx`.
So, i create a global fake node,  all the calls `ipfs add -n xxx` can use this fake node to avoid GenerateKeyPairWithReader.

### DIFF
--- a/core/coreapi/unixfs.go
+++ b/core/coreapi/unixfs.go
@@ -28,42 +28,27 @@ import (
 )
 
 type UnixfsAPI CoreAPI
+
 var nilNode *core.IpfsNode
-var lock sync.Mutex
+var once sync.Once
 
 func getOrCreateNilNode() (*core.IpfsNode,error) {
-	lock.Lock()
-	defer lock.Unlock()
-
-	if nilNode != nil {
-		return nilNode,nil
-	}
-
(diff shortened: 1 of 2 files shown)


core/coreapi/unixfs.go
```diff
@@ -27,6 +28,43 @@ import (
...
+
+	if err != nil {
+		return nil, err
+	}
+
+	nilNode = node
+	return nilNode,nil
+}
+
+func CloseFakeRepo() error {
```
Stebalien: I wouldn't bother with this. We don't actually need to close this node.

core/coreapi/unixfs.go
```diff
@@ -27,6 +28,43 @@ import (
 )
 
 type UnixfsAPI CoreAPI
+var nilNode *core.IpfsNode
+var lock sync.Mutex
```
Stebalien: Let's use a [`sync.Once`](https://golang.org/pkg/sync/#Once).

hcg1314: ok

Stebalien:
Thanks!

PR #6771, opened 2019-11-20
extract the pinner to go-ipfs-pinner and dagutils into go-merkledag

MichaelMure:
This PR extract the default `Pinner` and allow to plug a different implementation. This change is composed of:

- migrating the `dagutils` package from this repo into  ̀go-merkledag` to avoid a cyclic dependency: https://github.com/ipfs/go-merkledag/pull/50
- extracting the `Pinner` into https://github.com/MichaelMure/go-ipfs-pinner. Ideally Protocol Labs adopt this repo into its own org.
- removing the duplicate code and making the few changes needed here


MichaelMure:
cc @hsanjuan @Stebalien

PR #6773, opened 2019-11-25
fix: ignore nonexistant when force rm

olizilla:
- Make `ipfs files rm --force /nonexistant` succeed when the path does not exist.
- Add shaness test for removing nonexistant paths
- Refactor duplicated code to find a parent dir into a function

I've been writing scripts against the files api, and having to stat things before removing them is a pain. So this PR aims to make --force do what I'd expect it to.

License: MIT

### DIFF
--- a/core/commands/files.go
+++ b/core/commands/files.go
@@ -17,14 +17,14 @@ import (
 	bservice "github.com/ipfs/go-blockservice"
 	cid "github.com/ipfs/go-cid"
 	cidenc "github.com/ipfs/go-cidutil/cidenc"
-	"github.com/ipfs/go-ipfs-cmds"
-	"github.com/ipfs/go-ipfs-exchange-offline"
+	cmds "github.com/ipfs/go-ipfs-cmds"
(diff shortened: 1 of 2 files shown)


core/commands/files.go
```diff
@@ -997,26 +997,33 @@ Remove files or directories.
...
+		// if '--force' specified, it will remove anything else,
+		// including file, directory, corrupted node, etc
+		force, _ := req.Options[forceOptionName].(bool)
+
 		dir, name := gopath.Split(path)
-		parent, err := mfs.Lookup(nd.FilesRoot, dir)
+
+		pdir, err := getParentDir(nd.FilesRoot, dir)
 		if err != nil {
+			if force {
```
dirkmc:
```suggestion
			if force && err == os.ErrNotExist {
```

dirkmc: Just to be clear: here I'm suggesting using an if statement _instead_ of the switch

olizilla: yep, good point. Done!

Stebalien:
Thanks!

PR #6774, opened 2019-11-29
pin: implement pin/ls with only CoreApi

MichaelMure:
This PR use improvements from https://github.com/ipfs/interface-go-ipfs-core/pull/49 and https://github.com/ipfs/interface-go-ipfs-core/pull/50 to implement `pin ls` only using the CoreApi.

During this refactor I discovered a bug [I introduced in https://github.com/ipfs/go-ipfs/pull/6493, sorry] where the priority between pin type was not respected: when doing a `ipfs pin ls --type=indirect`, direct and recursive pins were not visited first so one of those could have been listed as indirect. This PR also fix this.


MichaelMure:
CC @Stebalien @magik6k

Stebalien:
#6705 fixes the issue you found.

MichaelMure:
@Stebalien as I understand, there is two solutions here:
1. Change `PinAPI.Ls` to return `(chan coreiface.Pin, chan error)` to match what #6705 did 
2. Keep the interface defined in https://github.com/ipfs/interface-go-ipfs-core/pull/49 and https://github.com/ipfs/interface-go-ipfs-core/pull/50 and return `(<-chan Pin, error)`. In that case, rebasing this PR would essentially be copying my implementation over the one currently in master.

Can you confirm which one you'd like ?

Stebalien:
Let's go with the second. I think we need to evaluate using multiple channels for errors everywhere in the future but that's a larger change.

MichaelMure:
Should be good to go now

MichaelMure:
Someone to merge this one?

Stebalien:
I will look at this when I get a chance (probably not till after the holidays) but I'm going to need to think about the interface changes.

MichaelMure:
Rebased on master. It would be nice to merge that alongside https://github.com/ipfs/interface-go-ipfs-core/pull/50.

hsanjuan:
I'll get back to review this after https://github.com/ipfs/interface-go-ipfs-core/pull/50 is through.

MichaelMure:
@Stebalien do you think this could be included in the 0.5.0 ? That would be neat but it's fine if not.

hsanjuan:
hey @MichaelMure , this is not making it to 0.5.0 (currently very much in feature-freeze). That said these are on our radar for merging afterwards.

Stebalien:
Rebased on master.

core/coreapi/pin.go
```diff
@@ -194,123 +214,106 @@ func (p *pinInfo) Type() string {
...
 
 	keys := cid.NewSet()
-	AddToResultKeys := func(keyList []cid.Cid, typeStr string) error {
+
+	AddToResultKeys := func(keyList []cid.Cid, typeStr string) {
 		for _, c := range keyList {
 			if keys.Visit(c) {
-				select {
-				case ch <- &pinInfo{
+				out <- &pinInfo{
```
Stebalien, line 229: Unless I'm missing something, we still need to select on the context, right?

MichaelMure: Looks like it. Also, `ctx` should be the first parameter of `pinLsAll`.

MichaelMure: Should be good now.

Issue #6775, opened 2019-12-03
Improve Add Performance W/ Async Datastores


Stebalien:
fixes https://github.com/ipfs/go-ipfs/issues/6523

* [x] Add async write support to datastores - https://github.com/ipfs/go-datastore/issues/137
* [x] Add `Sync` calls where appropriate (e.g., when pinning).
* [x] Turn syncwrites off on badger in go-ipfs by default.
* [ ] Switch to badger by default - https://github.com/ipfs/go-ipfs/issues/4279.

aschmahmann:
## Proposal for where to add Sync calls in go-ipfs (and related libraries)

The overall plan is to determine the places where a user would expect data to be persisted to disk and make sure we've called sync then such that a crash.

If the data that needs to be persisted is IPLD blocks then we should sync the entire blockstore, and if the data could be in the filestore then we should sync the entire filestore as well.

- Pinning: Any modification of the pin set
  - Call Sync during a Pin Flush on (in order):
      - The blockstore because of potentially new DAG nodes added during a sync
      - The blockstore because of potential changes to the DAG storing the pinset
    - The datastore key that points to the root of the pin set
- Adding data to IPFS: After an add operation is completed the data should be synced
   - Data can potentially be added to either the filestore or blockstore
   - Call Sync on the blockstore + filestore after an add completes
- MFS: Anytime the root is modified
   - Data can potentially be added to either the filestore or blockstore (confirm?)
   - Call Sync during an MFS Flush on (in order):
      - The blockstore + filestore because of potentially new DAG nodes added during a sync
     - The datastore key that points to the MFS root
- IPNS: When an update is submitted (before any network calls are made)
     - Sync an IPNS key whenever it is created and added to the datastore

Stebalien:
So, the goal was to improve add performance on hard disks. However, testing this on my SSD, I'm still seeing a 2x speedup. That is, 145MB/s versus 73MB/s.

:clap::clap::clap:

momack2:
Do we still plan to switch to badger by default, or is this item considered closed (at least, as far as the 0.5 release is concerned in https://github.com/ipfs/go-ipfs/issues/6776?

Stebalien:
@momack2 I've updated the meta issue. You're right, this is "closed" as far as 0.5.0 is concerned.

Stebalien:
This issue isn't really helping anything so I'm closing it. We still want to switch to a different datastore, but, well, details in the badger issue.

PR #6785, opened 2019-12-05
Support Asynchronous Datastores

aschmahmann:
As part of #6775, we need to add Sync calls into the appropriate places in go-ipfs (and its dependencies).

The places that were deemed important to have Sync calls include:

- Pinning: Any time the pin root is flushed all the puts it relies on should have completed
- Adding: After an add operation is completed the data should be synced
- MFS: Anytime the root is modified
- IPNS: When an update is submitted (before any network calls are made)

### DIFF
--- a/core/coreapi/unixfs.go
+++ b/core/coreapi/unixfs.go
@@ -12,6 +12,7 @@ import (
 	blockservice "github.com/ipfs/go-blockservice"
 	cid "github.com/ipfs/go-cid"
 	cidutil "github.com/ipfs/go-cidutil"
+	filestore "github.com/ipfs/go-filestore"
 	bstore "github.com/ipfs/go-ipfs-blockstore"
 	files "github.com/ipfs/go-ipfs-files"
 	ipld "github.com/ipfs/go-ipld-format"
@@ -96,7 +97,29 @@ func (api *UnixfsAPI) Add(ctx context.Context, files files.Node, opts ...options
 	bserv := blockservice.New(addblockstore, exch) // hash security 001
 	dserv := dag.NewDAGService(bserv)
 
-	fileAdder, err := coreunix.NewAdder(ctx, pinning, addblockstore, dserv)
+	// add a sync call to the DagService
+	// this ensures that data written to the DagService is persisted to the underlying datastore
+	// TODO: propagate the Sync function from the datastore through the blockstore, blockservice and dagservice
+	var syncDserv *syncDagService
+	if settings.OnlyHash {
+		syncDserv = &syncDagService{
+			DAGService: dserv,
+			syncFn:     func() error { return nil },
+		}
+	} else {
+		syncDserv = &syncDagService{
+			DAGService: dserv,
+			syncFn: func() error {
+				ds := api.repo.Datastore()
+				if err := ds.Sync(bstore.BlockPrefix); err != nil {
+					return err
+				}
+				return ds.Sync(filestore.FilestorePrefix)
+			},
+		}
+	}
+
+	fileAdder, err := coreunix.NewAdder(ctx, pinning, addblockstore, syncDserv)
 	if err != nil {
 		return nil, err
 	}
@@ -271,4 +294,14 @@ func (api *UnixfsAPI) lsFromLinks(ctx context.Context, ndlinks []*ipld.Link, set
 
 func (api *UnixfsAPI) core() *CoreAPI {
 	return (*CoreAPI)(api)
+}
+
+// syncDagService is used by the Adder to ensure blocks get persisted to the underlying datastore
+type syncDagService struct {
+	ipld.DAGService
+	syncFn func() error
+}
+
+func (s *syncDagService) Sync() error {
+	return s.syncFn()
 }
--- a/core/coreunix/add.go
+++ b/core/coreunix/add.go
@@ -38,6 +38,10 @@ type Link struct {
 	Size       uint64
 }
 
+type syncer interface {
+	Sync() error
+}
+
 // NewAdder Returns a new Adder used for a file add operation.
 func NewAdder(ctx context.Context, p pin.Pinner, bs bstore.GCLocker, ds ipld.DAGService) (*Adder, error) {
 	bufferedDS := ipld.NewBufferedDAG(ctx, ds)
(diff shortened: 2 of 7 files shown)


core/node/core.go
```diff
@@ -77,7 +93,15 @@ func OnlineExchange(provide bool) interface{} {
...
-		return repo.Datastore().Put(dsk, c.Bytes())
+		rootDS := repo.Datastore()
+		if err := rootDS.Sync(blockstore.BlockPrefix.ChildString(c.String())); err != nil {
+			return err
+		}
+		if err := rootDS.Sync(filestore.FilestorePrefix); err != nil {
+			return err
+		}
+
+		return rootDS.Put(dsk, c.Bytes())
```
Stebalien: Should call sync.

core/coreapi/unixfs.go
```diff
@@ -96,7 +97,29 @@ func (api *UnixfsAPI) Add(ctx context.Context, files files.Node, opts ...options
...
+		syncDserv = &syncDagService{
+			DAGService: dserv,
+			syncFn:     func(c cid.Cid) error { return nil },
+		}
+	} else {
+		syncDserv = &syncDagService{
+			DAGService: dserv,
+			syncFn: func(c cid.Cid) error {
+				ds := api.repo.Datastore()
+				if err := ds.Sync(bstore.BlockPrefix.ChildString(c.String())); err != nil {
```
Stebalien: That's not what block keys look like. Take a look at `go-ipfs-ds-help.CidToDsKey`.

core/coreunix/add.go
```diff
@@ -316,6 +320,13 @@ func (adder *Adder) AddAllAndPin(file files.Node) (ipld.Node, error) {
 		return nil, err
 	}
 
+	if asyncDagService, ok := adder.dagService.(syncer); ok {
+		err = asyncDagService.Sync(nd.Cid())
```
Stebalien: Syncing a _single_ block isn't going to do anything useful.

core/node/core.go
```diff
@@ -41,18 +42,33 @@ func BlockService(lc fx.Lifecycle, bs blockstore.Blockstore, rem exchange.Interf
...
 		// not being found in the datastore.
 		// this is kinda sketchy and could cause data loss
-		pinning = pin.NewPinner(repo.Datastore(), ds, internalDag)
+		pinning = pin.NewPinner(rootDS, syncDs, syncInternalDag)
 	}
 
 	return pinning, nil
 }
 
+type syncDagService struct {
```
Stebalien: Do we use this? This doesn't conform to any of the interfaces we've defined.

core/node/core.go
```diff
@@ -77,7 +93,18 @@ func OnlineExchange(provide bool) interface{} {
 func Files(mctx helpers.MetricsCtx, lc fx.Lifecycle, repo repo.Repo, dag format.DAGService) (*mfs.Root, error) {
 	dsk := datastore.NewKey("/local/filesroot")
 	pf := func(ctx context.Context, c cid.Cid) error {
-		return repo.Datastore().Put(dsk, c.Bytes())
+		rootDS := repo.Datastore()
+		if err := rootDS.Sync(blockstore.BlockPrefix.ChildString(c.String())); err != nil {
```
Stebalien: Again, syncing a single block isn't going to help. We either need to sync the tree, probably, just sync all blocks.

core/node/core.go
Stebalien: Ah, now I see why. Let's document that the pinner uses this interface.

Stebalien:
Ok, now that I remember how the pinner is calling `Sync`, this makes more sense. However, the calls to `Sync(cid)` are still wrong. Let's still write a proposed design with what/where/why we need to call sync.

core/coreapi/unixfs.go
aschmahmann: Yep, thanks

aschmahmann:
> However, the calls to Sync(cid) are still wrong

Sure, we can just call Sync on `bstore.BlockPrefix` it's the correct thing to do (note that all the current implementations work fine even with this being done incorrectly).

core/node/core.go
aschmahmann: Will just call Sync on bstore.BlockPrefix

core/coreunix/add.go
aschmahmann: Will change the interface here to not take CIDs, we'll end up syncing the entire blockstore

core/coreapi/unixfs.go
aschmahmann: i.e. any reason to not just sync the whole Blockstore?

core/node/core.go
```diff
@@ -41,18 +42,35 @@ func BlockService(lc fx.Lifecycle, bs blockstore.Blockstore, rem exchange.Interf
 // Pinning creates new pinner which tells GC which blocks should be kept
 func Pinning(bstore blockstore.Blockstore, ds format.DAGService, repo repo.Repo) (pin.Pinner, error) {
 	internalDag := merkledag.NewDAGService(blockservice.New(bstore, offline.Exchange(bstore)))
-	pinning, err := pin.LoadPinner(repo.Datastore(), ds, internalDag)
+	rootDS := repo.Datastore()
+
+	syncFn := func() error { return rootDS.Sync(blockstore.BlockPrefix) }
```
aschmahmann: No reason the pinner needs to sync the Filestore is there?

core/coreunix/add.go
```diff
@@ -316,6 +320,13 @@ func (adder *Adder) AddAllAndPin(file files.Node) (ipld.Node, error) {
 		return nil, err
 	}
 
+	if asyncDagService, ok := adder.dagService.(syncer); ok {
+		err = asyncDagService.Sync()
+		if err != nil {
+			return nil, err
+		}
+	}
+
```
aschmahmann: A little unfortunate that in the common case we'll end up syncing the same key twice since right after this we call `PinRoot` which calls `pin.Flush()`

core/node/core.go
Stebalien: Unfortunately, yes. When pinning, we may _reference_ blocks in the filestore.

core/coreunix/add.go
Stebalien: Yeah... but hopefully, this will basically be a no-op.

PR #6786, opened 2019-12-06
feat: files (mfs) api implementation

hacdias:
This implements the Files (MFS) API

- [ ] Interface: https://github.com/ipfs/interface-go-ipfs-core/pull/54
- [x] Moves almost all `ipfs files <cmd>` related code to an helper package called `fileshelpers`
- [x] Implements the API for all interfaces
- [x] Keeps current functionality, but changes the meaning of `count=0` in `files write` and `files read`. To simplify, if count equals 0, then we want to write or read everything. It doesn't even make sense to not read or not write anything.

License: MIT
Signed-off-by: Henrique Dias <hacdias@gmail.com>

### DIFF
--- a/core/commands/files.go
+++ b/core/commands/files.go
@@ -5,25 +5,20 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"os"
-	gopath "path"
 	"sort"
 	"strings"
 
 	"github.com/ipfs/go-ipfs/core"
 	"github.com/ipfs/go-ipfs/core/commands/cmdenv"
+	"github.com/ipfs/go-ipfs/core/fileshelpers"
 
 	"github.com/dustin/go-humanize"
-	bservice "github.com/ipfs/go-blockservice"
 	cid "github.com/ipfs/go-cid"
-	cidenc "github.com/ipfs/go-cidutil/cidenc"
 	cmds "github.com/ipfs/go-ipfs-cmds"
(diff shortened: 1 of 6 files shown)


hacdias:
@Stebalien I started working on this to support Files API through Core API. I wanted to ask you if you feel this is the right direction...

I created a `fileshelpers` package so I didn't repeat as much code between commands and Core API.

Stebalien:
I think this is the right direction. We can unify the unixfs and files APIs later (and deprecate this interface when we do that).

hacdias:
@Stebalien I think this is done and only needs to merge the interface and remove the replace here.

I don't understand why sharness tests are failing. I checked on my laptop and only mounting ones seem to be failing.

Stebalien:
I'm not sure why they're passing for you. Try running:

```
> cd test/sharness
> make deps
> ./t0250-files-api.sh -v -i # -v means verbose, -i means "stop immediately on failure"
```

`ipfs files read --count 0` is returning too many bytes. It looks like lengths and offsets need to be handled.

Stebalien:
I'm not sure why they're passing for you. Try running:

```
> cd test/sharness
> make deps
> ./t0250-files-api.sh -v -i # -v means verbose, -i means "stop immediately on failure"
```

`ipfs files read --count 0` is returning too many bytes. It looks like lengths and offsets need to be handled.

hacdias:
@Stebalien is this something still desirable or has it been done/can be closed?

PR #6789, opened 2019-12-09
chore(dep): update go-datastore

Stebalien:
Let's get this in so the datastore changes don't block other PRs.


aschmahmann:
@Stebalien run a `go mod tidy` and then LGTM

Stebalien:
Trying again in #6791 to trigger CI.

PR #6799, opened 2019-12-14
Sharness macos cleanups

ribasushi:
While getting familiar with the codebase ran into a large number of issues with test execution on macos.

Things still do not work cleanly with fuse, but at least everything passes on first-take with `NO_TEST_FUSE=1`

### DIFF
--- a/test/sharness/lib/iptb-lib.sh
+++ b/test/sharness/lib/iptb-lib.sh
@@ -26,6 +26,17 @@ iptb() {
         esac
         return 1
     fi
+
+    # If we are looking for clean-start, issue a 'stop' twice for good measure
+    # It sadly doesn't always work :/
+    if test "$GOIPFSTEST_ENSURE_CLEAN_START" = 1 \
+    && test "$#" = 1 \
+    && test "$1" = stop \
+    ; then
+      sleep 1
+      command iptb stop > /dev/null 2>&1 &
+      sleep 1
+    fi
 }
 
 startup_cluster() {
--- a/test/sharness/lib/test-lib.sh
+++ b/test/sharness/lib/test-lib.sh
@@ -55,6 +55,10 @@ test "$TEST_EXPENSIVE" = 1 && test_set_prereq EXPENSIVE
 test "$TEST_NO_DOCKER" != 1 && type docker >/dev/null 2>&1 && test_set_prereq DOCKER
 test "$TEST_NO_PLUGIN" != 1 && test "$TEST_OS" = "LINUX" && test_set_prereq PLUGIN
 
+# this may not be available, skip a few dependent tests
+type socat >/dev/null 2>&1 && test_set_prereq SOCAT
+
+
 # Set a prereq as error messages are often different on Windows/Cygwin
 expr "$TEST_OS" : "CYGWIN_NT" >/dev/null || test_set_prereq STD_ERR_MSG
 
--- a/test/sharness/lib/test-lib.sh
+++ b/test/sharness/lib/test-lib.sh
@@ -52,7 +52,7 @@ TEST_OS="$(uname -s | tr '[a-z]' '[A-Z]')"
 # grab + output options
 test "$TEST_NO_FUSE" != 1 && test_set_prereq FUSE
 test "$TEST_EXPENSIVE" = 1 && test_set_prereq EXPENSIVE
-test "$TEST_NO_DOCKER" != 1 && type docker >/dev/null 2>&1 && test_set_prereq DOCKER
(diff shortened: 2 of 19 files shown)


test/sharness/lib/iptb-lib.sh
```diff
@@ -26,6 +26,17 @@ iptb() {
...
+    # If we are looking for clean-start, issue a 'stop' twice for good measure
+    # It sadly doesn't always work :/
+    if test "$GOIPFSTEST_ENSURE_CLEAN_START" = 1 \
+    && test "$#" = 1 \
+    && test "$1" = stop \
+    ; then
+      sleep 1
+      command iptb stop > /dev/null 2>&1 &
+      sleep 1
+    fi
```
Stebalien: `iptb stop` _should_ wait, and calling stop a second time shouldn't help. Are you sure we don't just have leftovers from a previous failed run?

IPTB currently:

1. Sends sigterm and waits a second.
2. Sends sigterm again and waits 2 seconds.
3. Sends sigquit and waits 5 seconds.
4. Sends sigkill and waits 5 seconds.

test/sharness/lib/test-lib.sh
```diff
@@ -41,6 +41,43 @@ SHARNESS_LIB="lib/sharness/sharness.sh"
 
 # Please put go-ipfs specific shell functions below
 
+if test "$GOIPFSTEST_ENSURE_CLEAN_START" = 1 ; then
```
Stebalien: `make test_sharness_expensive` actually invokes sharness tests in parallel, IIRC. I don't think this logic will work with that.

_However_, when running a sharness test, we `cd` into a "trash" directory. We _could_ kill off all IPFS instances running in that trash directory.

test/sharness/t0131-multinode-client-routing.sh
```diff
@@ -46,9 +46,11 @@ test_expect_success "set up testbed" '
   iptb testbed create -type localipfs -count $NNODES -force -init
 '
 
+# for some reason starting 8/9 together gets stuck on a wait...
 test_expect_success "start up nodes" '
   iptb start -wait [0-7] &&
-  iptb start -wait [8-9] -- --routing=dhtclient
+  iptb start -wait 8 -- --routing=dhtclient &&
+  iptb start -wait 9 -- --routing=dhtclient
```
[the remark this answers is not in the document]
ribasushi: Just filing a bug won't help much, as the thing is bizarrely intermittent, and might even be my-machine-specific. I will look into pinning this down a bit later today instead.

test/sharness/lib/test-lib.sh
ribasushi: Good idea! I definitely wouldn't want to kill them, but rather detect them and fail. The main point of this eercise was to stop my workstation being filled up with random not-shutdown-daemons

test/sharness/lib/iptb-lib.sh
ribasushi: 99% this is not the case, but I will try to instrument this better and get back to this.

Let's hold off merging this until I had a moment to investigate the iptb-specific stuff more in depth.

Stebalien: The code is https://github.com/ipfs/iptb-plugins/blob/d63cc1c88acb3195428f633bba332a5f5f53a8ea/local/localipfs.go#L203 (might be buggy).

test/sharness/t0131-multinode-client-routing.sh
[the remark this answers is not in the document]
Stebalien: Thanks! I just don't like adding workarounds for bugs without tracking the issue somewhere.

test/sharness/lib/iptb-lib.sh
Stebalien: @ribasushi could we pull out everything unrelated to the clean start logic so we can merge that separately?

(stuff sitting in PRs tends to die)

ribasushi: > stuff sitting in PRs tends to die

@Stebalien not my stuff as a general rule, but yes, will separate into "less contentious" piece today-ish ( I am traveling to congress, will be in-and-out )

ribasushi: @Stebalien https://github.com/ipfs/go-ipfs/pull/6805 is now smoking, I am still keeping this PR open to fix the remaining parts, ETA some time this year

Stebalien:
I'm going to close this PR for as it's stalled but leave the branch in place.

ribasushi:
NP. I'm pulling stuff locally into my 'to redo' pile, deleting branch.

PR #6808, opened 2019-12-27
add support for detecting window close events (on Windows)

djdv:
WIP fix for https://github.com/ipfs/go-ipfs/issues/1897

We add a shim `nativeNotifier` who's responsibility is to notify a channel of events/signals not covered by the `signal` pkg.
In this case, we implement this to add support for Windows' window events.

When the window receives a `CTRL+C`, `CTRL+BREAK`, or `WM_CLOSE` event, we will notify the interrupt channel and tell the OS we proccessed the event successfully.
This in turn will cancel the node's context for the first event and exit outright upon subsequent interrupts.

**Current problem**:
When processing `CTRL_C_EVENT` and `CTRL_BREAK_EVENT` events, nothing special has to be considered. But `WM_CLOSE` breaks down to a `CTRL_CLOSE_EVENT` which has special implications in this context.
https://docs.microsoft.com/en-us/windows/console/handlerroutine

When we receive this message, the system will terminate us in either 5 seconds, or when we return from our event handler (whichever is first).
Meaning we can initiate graceful shutdown from any signal, but need a way to wait for shutdown to be completed before returning from the handler, specifically when handling the `CTRL_CLOSE_EVENT` event.

i.e. the node has to...
```Go
case windows.CTRL_CLOSE_EVENT:
	// start shutting down here
	...
	// and block here until shutdown is finished
	return processed // this is the moment the process dies 👀
```

Commands like `shutdown` have direct access to the node and can just call its `Close` method, blocking until the node is finished shutting down.
We're a background thread that gets called at some arbitrary point by the user/OS, and don't have access to the node object so we can't just do the same.

This needs to be taken care of before this can progress out of a draft.

Edit:
[video showing the signal getting caught](https://www.youtube.com/watch?v=atWnbhnqWDI)

### DIFF
--- a/cmd/ipfs/util/signal.go
+++ b/cmd/ipfs/util/signal.go
@@ -12,6 +12,12 @@ import (
 	"syscall"
 )
 
+type nativeNotifierFunc func(ih *IntrHandler, consumerLine chan os.Signal, requestedSignals ...os.Signal)
+
+// if you need to register platform specific signals not handled by Golang's stdlib
+// define this function and set it during `init`
+var nativeNotifier nativeNotifierFunc
+
 // IntrHandler helps set up an interrupt handler that can
 // be cleanly shut down through the io.Closer interface.
 type IntrHandler struct {
@@ -36,7 +42,16 @@ func (ih *IntrHandler) Close() error {
 // handler's wait group to ensure clean shutdown when Close() is called.
 func (ih *IntrHandler) Handle(handler func(count int, ih *IntrHandler), sigs ...os.Signal) {
 	notify := make(chan os.Signal, 1)
+
+	// register with Golang's stdlib
 	signal.Notify(notify, sigs...)
+
+	// if we have platform specific handler logic
+	// call it so it can inject signals from the OS
+	if nativeNotifier != nil {
+		nativeNotifier(ih, notify, sigs...)
+	}
+
 	ih.wg.Add(1)
 	go func() {
 		defer ih.wg.Done()
--- a/cmd/ipfs/util/signal_windows.go
+++ b/cmd/ipfs/util/signal_windows.go
@@ -1,0 +1,62 @@
+package util
+
+import (
+	"os"
+
+	"golang.org/x/sys/windows"
+)
+
+type (
+	DWORD    = uint32
+	MSVCBOOL = uintptr
+)
+
+const (
+	processed = iota
+	notProcessed
+)
+
+func init() {
+	nativeNotifier = windowEventHandler
+}
+
+// windowEventHandler creates and registers a WINAPI `HandlerRoutine` callback
+// which receives window events and relays them as interrupt signals to the provided channel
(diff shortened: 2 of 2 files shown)


cmd/ipfs/util/signal_windows.go
```diff
@@ -0,0 +1,62 @@
...
+			goIsHandlingIntterupts = true
+		}
+	}
+
+	fp := func(ctrlType DWORD) MSVCBOOL {
+		switch ctrlType {
+		case windows.CTRL_CLOSE_EVENT:
+			//NOTE: the OS will terminate our process after receiving this event
+			// either after `SPI_GETHUNGAPPTIMEOUT` or immediately after we return non-0
+			// (whichever is first)
```
djdv: This is 5AM influenza English, look at again later.

cmd/ipfs/util/signal.go
```diff
@@ -36,7 +42,16 @@ func (ih *IntrHandler) Close() error {
...
 	notify := make(chan os.Signal, 1)
+
+	// register with Golang's stdlib
 	signal.Notify(notify, sigs...)
+
+	// if we have platform specific handler logic
+	// call it so it can inject signals from the OS
+	if nativeNotifier != nil {
+		nativeNotifier(ih, notify, sigs...)
+	}
```
Stebalien: This should probably go _below_ the `ih.wg.Add` call (or maybe just move that goroutine to the top of the function?). Otherwise, we could:

1. Receive a window close event before we call `ih.wg.Add`
2. Call `ih.wg.Wait()`
3. Succeed.
4. Either the process gets killed _or_ we then call `ih.wg.Add(1)` and panic (because we've already called wait on the waitgroup.

cmd/ipfs/util/signal_windows.go
```diff
@@ -0,0 +1,62 @@
...
+			sigChan <- os.Interrupt
+			ih.wg.Wait()
+			return processed
+		case windows.CTRL_C_EVENT, windows.CTRL_BREAK_EVENT:
+			if !goIsHandlingIntterupts {
+				sigChan <- os.Interrupt
+			}
+			return processed
```
Stebalien: If go is handling interrupts, will this cause problems? I wonder if we should just _always_ use this function on windows (skipping the `signal.Notify` call).

djdv:
Note:
Conveniently timed, Go 1.14 will handle this directly.
https://github.com/golang/go/commit/5d1a95175e693f5be0bc31ae9e6a7873318925eb

Will need to see how this behaves in 1.14 and hide this patch behind a build constant for <=1.13.
This way we handle the signal ourselves on old versions of the compiler, but don't do anything special on 1.14+.
(It's likely still preferable to catch and handle the signal ourselves even on 1.14 though, will look into)

Stebalien:
1.14 is out

djdv:
Since Go 1.14 handles this natively we can just drop this hack.

Issue #6797, opened 2019-12-14
Fixing Bootstrapping


Stebalien:
#### Version information:

> v0.5.0-dev

#### Description:

Libp2p recently increased the minimum RSA key size to 2048 bits. While almost all nodes in the network are using 2048 bit keys already, most of the built-in bootstrap nodes _are not_.

Luckily, we did have some newer nodes using 2048 bit keys. Specifically, the /dnsaddr/bootstrap.libp2p.io nodes. Unfortunately:

1. These bootstrappers were added (relatively) recently. Nodes with older configs won't use them.
2. The DNS record for these bootstrappers was _too large_: 4 bootstrappers, 4 IP addresses per bootstrapper) to fit into a single UDP packet.

To solve 2, we switched to indirect dnsaddr records: now we have 4 top-level records that each resolve two two dnsaddr records, one for each IP address for each bootstrapper. _These_ records all, individually, fit into single UDP packets.

Unfortunately, go-libp2p doesn't support recursive dnsaddr resolution: https://github.com/libp2p/go-libp2p/issues/744.

---

So, we need to:

* [x] Add support for recursive dnsaddr resolution to go-libp2p (https://github.com/libp2p/go-libp2p/issues/744).
* [x] Add a repo migration that:
    * Checks if we already have the new bootstrappers in the config. If so, skip.
    * Checks to see if we have bootstrappers with known sub 2048 bit keys. If not, skip.
    * Otherwise, remove them and replace these old bootstrappers with the new bootstrappers.
    * Replace the `/ipfs` part of these multiaddrs with `/p2p` (no reason not to).
* [x] Test the migration.
* [x] Update the repo version in go-ipfs.

dirkmc:
Which of [the peers](https://github.com/ipfs/go-ipfs-config/blob/master/bootstrap_peers.go#L21) have sub 2048 bit keys?

olizilla:
The ones addressed by `/dnsaddr`... https://github.com/ipfs/go-ipfs-config/blob/98ec0ae9a3161d355da0615e54ca314232212615/bootstrap_peers.go#L22-L25

Stebalien:
@olizilla & @dirkmc 

[...]

PR #6814, opened 2020-01-07
feat(repo): bump repo version

Stebalien:
We're bumping to version 8 to migrate bootstrappers:

1. Switch from /ipfs to /p2p. This isn't necessary but it's still "nice".
2. Add new bootstrappers to old nodes that don't have them.

part of #6797


Stebalien:
We need to release the migration and update the internal "dists" URL before we can merge this. However, to test the migration, we need to cut a "fake" release (not published). That's what this PR is for right now.

Stebalien:
Continuing in #6955

Issue #6815, opened 2020-01-07
Switch to raw multihashes in datastores


Stebalien:
Part of https://github.com/ipfs/go-ipfs/issues/4143.

Currently, we store blocks by CID in the datastore. However, a single block can have _multiple_ CIDs:

* CIDv0/CIDv1 -- blocks created as cidv0 can be referenced with v1 CIDs
* CIDv1-*/CIDv1-raw -- we can treat any block as a "raw" block
* CIDv1-cbor/CIDv1-dag-cbor -- theoretically, we have non-"dag" versions of our codecs.

Plan:

TODO: put this in the right spot: https://github.com/ipfs/fs-repo-migrations/pull/95

* [x] Create a repo migration to switch from CIDs to multihashes. Luckily, all CIDv0 blocks (most blocks) are valid multihashes so we don't have to move them.
  * [x] https://github.com/ipfs/fs-repo-migrations/pull/95
  * [ ] This migration should iterate over all blocks in the datastore:
    * If the block is stored using CIDv1, we should add it again using the raw multihash (if it's not already there).
    * Delete the CIDv1 block.
    * If the datastore offers transactions, we should ideally perform each add/delete in a transaction.
    * We should probably have multiple threads doing this for better performance.
    * We should do this with synchronous writes _enabled_. Alternatively, we could do this in two passes (add all then delete all) but that's going to double the amount of disk space used.
  * [ ] Optional: Iterate over all provider records and migrate them as well. This is not critical as these will expire anyways.
* [x] Convert CIDs to/from multihashes in the _blockstore_ layer. At this point, the external blockstore interfaces will still use CIDs.
  * [x] https://github.com/ipfs/go-ipfs-ds-help/pull/18
  * [x] https://github.com/ipfs/go-ipfs-blockstore/pull/38
* [x] Fix GC (in go-ipfs) to work when the datastore no longer stores codec information.
  * [x] https://github.com/ipfs/go-ipfs/pull/6816
* [x] Remove the previous CIDv0/CIDv1 blockstore hacks.
  * [x] https://github.com/ipfs/go-ipfs/pull/6816
* [ ] Eventually, switch the blockstore API to use multihashes directly instead of CIDs.
 * 2021-12-14: this will get handled as part of https://github.com/ipfs/go-ipfs-blockstore/issues/8

---
2021-12-02 context on why were' doing this:
* Helps us to move towards CIDv1 by default
[...]

PR #6816, opened 2020-01-07
feat: switch to raw multihashes for blocks

Stebalien:
Part of: https://github.com/ipfs/go-ipfs/issues/6815


hsanjuan:
@Stebalien other than the right import versions here, this should be good.

gc/gc.go
```diff
@@ -154,7 +157,9 @@ func Descendants(ctx context.Context, getLinks dag.GetLinks, set *cid.Set, roots
 
 	for _, c := range roots {
 		// Walk recursively walks the dag and adds the keys to the given set
-		err := dag.Walk(ctx, verifyGetLinks, c, set.Visit, dag.Concurrent())
+		err := dag.Walk(ctx, verifyGetLinks, c, func(k cid.Cid) bool {
+			return set.Visit(k.Hash())
```
Stebalien: So... I should have left a comment here, walking by CIDv1 was intentional.

If we walk by multihash, someone could refer to a structured node (e.g., a DagPB node) as a "Raw" node. If they did that, we'd visit the raw node, see that it was raw, and mark it as visited. As a raw node, it doesn't _have_ any children.

However, if we then encounter the same block as a DagPB node, we'd see that we've visited it and move on. Unfortunately, we'd then GC all it's children.

hsanjuan: Hmmm. So we revert to cid.Set + convertion to multihashes, or do you see a different way to deal with it?

Stebalien: I don't. I think the visit set will always have to be true CIDs, then we'll need to convert to multihashes for the actual cleanup phase.

Note: we'll _also_ need those CIDs for the revert.

hsanjuan: I have reverted the change.

hsanjuan:
@Stebalien some sharness tests are failing because some outputs changed from CidV0 to CidV1 (i.e. `repo gc`) and text output is defaulting to base32 then.

Shall I patch at the core/commands level (i.e. we can convert to CidV0 in the places we do `.Emit`), or update sharness tests? This affects at least `refs k

I think things get messy with sharness. If I'm getting it right, `blockstore.AllKeysChan` is now going to return a list of CidV1.Raw (this affects `refs local`) regardless of what was the original CID type. But `ipfs add` will return a CidV1-DagPB.

So a sharness test adding a hash with CidV1 and then checking that `refs local` shows that hash will fail. Same for some things checking GC output. May also start triggering false positives. Might fix a bunch of those by adding an argument to `refs local` that allows to check for a single CID (rather than listing and grepping).

hsanjuan:
I'm going to work on fixing sharness tests directly

hsanjuan:
Yai all green :) @Stebalien can you check my last commits?

(btw it seems it fast-forwarded a merge from master or something and did not create a merge commit, making the history a bit dirty).

hsanjuan:
This one seems ok in terms of tests and everything, it just needed some rebase.

hsanjuan:
I will look at it more closely though, particularly the GC part.

aschmahmann:
@hsanjuan the only thing I clobbered/ignored in the rebase was 163794eadac2f5265030a64629cf34b88559be27 (t0081-repo-pinning.sh re-enable the offline daemon), since I figured it might've been intentional that we were running some of the tests with the daemon off and some with it on, but in offline mode.

If so we can just remove the extraneous comments. This issue seems unrelated to the main thrust of the PR, but ended up here as you were cleaning things up (thanks for that 🙏).

test/sharness/t0087-repo-robust-gc.sh
```diff
@@ -155,12 +157,12 @@ test_gc_robust_part2() {
...
 
 test_gc_robust_part1
-test_gc_robust_part2
+#test_gc_robust_part2
 
 test_launch_ipfs_daemon_without_network
 
 test_gc_robust_part1
-test_gc_robust_part2
+#test_gc_robust_part2
```
aschmahmann, line 165: Why are these disabled?

gc/gc.go
```diff
@@ -60,6 +70,17 @@ func GC(ctx context.Context, bs bstore.GCBlockstore, dstor dstore.Datastore, pn
 			}
 			return
 		}
+
+		// The blockstore reports raw blocks. We need to remove the codecs from the CIDs.
+		gcs, err = toRawCids(gcs)
```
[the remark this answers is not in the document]
hsanjuan: @aschmahmann I guess you saw the comment that is hidden above by @Stebalien :

```
If we walk by multihash, someone could refer to a structured node (e.g., a DagPB node) as a "Raw" node. If they did that, we'd visit the raw node, see that it was raw, and mark it as visited. As a raw node, it doesn't have any children.

However, if we then encounter the same block as a DagPB node, we'd see that we've visited it and move on. Unfortunately, we'd then GC all it's children.
```

Perhaps we could get away with a custom type of set that wraps additional info in structs and takes that.

test/sharness/t0080-repo.sh
```diff
@@ -109,17 +109,25 @@ test_expect_success "remove direct pin" '
...
-  ipfs refs local >actual8 &&
-  grep "QmYCvbfNbCwFR45HiNP45rwJgvatpiW38D961L5qAhUM5Y" actual8 &&
-  grep "$EMPTY_DIR" actual8 &&
-  grep "$HASH_WELCOME_DOCS" actual8 &&
-  test_must_fail grep "$HASH" actual8
+  HASH_MH=`cid-fmt -b base32 "%M" "$HASH"` &&
+  HARDCODED_HASH_MH=`cid-fmt -b base32 "%M" "QmYCvbfNbCwFR45HiNP45rwJgvatpiW38D961L5qAhUM5Y"` &&
+  EMPTY_DIR_MH=`cid-fmt -b base32 "%M" "$EMPTY_DIR"` &&
+  HASH_WELCOME_DOCS_MH=`cid-fmt -b base32 "%M" "$HASH_WELCOME_DOCS"` &&
+  ipfs refs local | cid-fmt -b base32 --filter "%M" >actual8 &&
```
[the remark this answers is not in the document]
hsanjuan: I don't know, this was done before `ipfs cid format` was there. Worth fixing in a different PR accross all tests?

test/sharness/t0087-repo-robust-gc.sh
hsanjuan: I have no memory of having to disable any tests. You should re-enable.

gc/gc.go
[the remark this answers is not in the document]
aschmahmann: ya, ok it's probably fine. I was thinking we were keep two sets in memory at once but by overriding `gcs` I  suspect you're freeing the data up for memory GC.

It's a bit of a memory waste for a while if you don't have duplicate blocks but I suspect that's not the most important of our GC issues 😅

hsanjuan (approved):

test/sharness/t0087-repo-robust-gc.sh
```diff
@@ -74,8 +78,8 @@ test_gc_robust_part1() {
   test_expect_success "'ipfs repo gc' should still run and remove as much as possible" '
     test_must_fail ipfs repo gc 2>&1 | tee repo_gc_out &&
     grep -q "could not remove $LEAF2" repo_gc_out &&
-    grep -q "removed $LEAF3" repo_gc_out &&
-    grep -q "removed $LEAF4" repo_gc_out &&
+    grep -q "removed $(to_raw_cid $LEAF3)" repo_gc_out &&
+    grep -q "removed $(to_raw_cid $LEAF4)" repo_gc_out &&
```
aschmahmann, line 82: These don't seem necessary since we're already doing raw leaves, but it can't hurt

go.mod
```diff
@@ -25,10 +25,10 @@ require (
 	github.com/ipfs/go-ds-leveldb v0.5.0
 	github.com/ipfs/go-ds-measure v0.2.0
 	github.com/ipfs/go-fetcher v1.6.1
-	github.com/ipfs/go-filestore v0.1.0
+	github.com/ipfs/go-filestore v1.1.0
 	github.com/ipfs/go-fs-lock v0.0.7
 	github.com/ipfs/go-graphsync v0.11.0
-	github.com/ipfs/go-ipfs-blockstore v0.2.1
+	github.com/ipfs/go-ipfs-blockstore v1.1.2
```
aschmahmann, line 31: TODO: Update dist CID to point at latest migration

repo/fsrepo/migrations/fetcher.go
```diff
@@ -10,7 +10,7 @@ import (
 
 const (
 	// Current distribution to fetch migrations from
-	CurrentIpfsDist = "/ipfs/QmP7tLxzhLU1KauTRX3jkVkF93pCv4skcceyUYMhf4AKJR" // fs-repo-migrations v2.0.2
+	CurrentIpfsDist = "/ipfs/QmPweMoUxWFt1MSpYwLWsHB1GBcyYYDKPnANdERMY4U6hK" // fs-repo-11-to-12 v1.0.1
```
aschmahmann, line 13: Note: Bumped to latest version of dist.ipfs.io

PR #6828, opened 2020-01-16
update dockerfile and use openssl

Stebalien:
This PR:

* Updates docker deps (go -> 1.13.5, tini -> 0.18.0)
* Builds go-ipfs using openssl. We don't distribute go-ipfs with OpenSSL support by default to avoid depending on external libraries. However, we completely control the docker environment so there's no reason not to build with OpenSSL support. This significantly reduces CPU usage.
* Explicitly set the permissions on start_ipfs so we don't inherit the defaults set by the host OS.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.13.4-buster
+FROM golang:1.13.6-buster
 LABEL maintainer="Steven Allen <steven@stebalien.com>"
 
 ENV SRC_DIR /go-ipfs
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,12 @@
 FROM golang:1.13.6-buster
 LABEL maintainer="Steven Allen <steven@stebalien.com>"
 
+# Install deps
+RUN apt-get update && apt-get install -y \
+  libssl-dev \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,6 +1,12 @@
-FROM golang:1.13.4-buster
+FROM golang:1.13.6-buster
 LABEL maintainer="Steven Allen <steven@stebalien.com>"
 
+# Install deps
+RUN apt-get update && apt-get install -y \
```
olizilla: Note: I think moving this to the first line means we'll have to explicitly run with `--no-cache=true` when we want to re-update those dependencies. The layer cache uses the string value of the RUN command to see if already has a layer for that.

> aside from the ADD and COPY commands, cache checking does not look at the files in the container to determine a cache match. For example, when processing a `RUN apt-get -y update` command the files updated in the container are not examined to determine if a cache hit exists. In that case just the command string itself is used to find a match.

https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#leverage-build-cache

Stebalien: That's probably for the best, actually. When rebuilding, we usually just want to update ipfs itself.

PR #6832, opened 2020-01-16
chore(dep): update deps

Stebalien:
Update all deps.


Stebalien:
@olizilla I assume we don't scrape `ipfs log tail` for metrics, right?

Stebalien:
I'm asking you for a review because I'm removing some event tracing.

PR #6834, opened 2020-01-17
fix: use pre-defined relays for autorelay

Stebalien:
Unfortunately, we don't currently have any way to pick out good relays from bad.
That means we keep searching, trying bad relays, searching some more, trying
_the same relays_, etc. until we randomly find 3 good stable relays. In
practice, this means we just keep searching forever and keep thrashing the DHT.

see https://github.com/libp2p/go-libp2p/issues/694


Stebalien:
Tested locally.

PR #6836, opened 2020-01-17
Release v0.4.23

Stebalien:
Begin v0.4.23 patch release, based on v0.4.22.

NOT FOR MERGING

Issue: https://github.com/ipfs/go-ipfs/issues/6837


go.mod
```diff
@@ -2,26 +2,25 @@ module github.com/ipfs/go-ipfs
 
 require (
```
b5: With this being a stopgap release, I'm wondering if we can't get a few module dependency upgrades to bring version numbers more up to date _without_ wondering into upstream dependencies for 0.5.0. This would help a number of us who import `go-ipfs` as a dep keep more up-to-date.

Here's my version-minimum wishlist for dependencies within the `github.com/ipfs` namespace. libp2p is a whole other story, and I'm sure I can find others, but first thought I'd check to see if you're amenable to these type of version bumps:

```
require (
  // ...
  
  // currently sitting at v0.0.5. latest release is 0.3.1
  // looks like the scary changes happen at 0.2.0
  github.com/ipfs/go-ds-badger v0.0.7

  // currently at v0.0.3
  github.com/ipfs/go-ipfs-config v0.0.11

  // currently at v0.0.3
  github.com/ipfs/go-ipfs-files v0.0.7

  // currently at 
  github.com/ipfs/interface-go-ipfs-core v0.2.3

)

```

Stebalien: I tried to keep the list of cherry-picked patches to a minimum.

For example, go-ds-badger 0.0.5->0.0.7 rewrites a significant portion of the query logic and adds periodic GC of deleted values. They're both bug fixes but they're definitely not risk-free bug fixes. We want our patch fixes releases to include critical bug fixes without including any bug fixes that may cause more issues than they fix.

Stebalien:
Closing as this was a patch release.

PR #6840, opened 2020-01-20
fix: Fix typo

HowJMay:
Fix typos


Stebalien:
(But thank you for taking the time to submit a PR! Small fixes like this are always welcome.)

Issue #6622, opened 2019-09-04
Docker built go-ipfs fails to report its commit version


olizilla:
Versions of go-ipfs built via our Dockerfile are unable to report their commit version, as initially reported by @gmasgras

To see this in action you can try out the go-ipfs:latest image and ask it for `ipfs version --commit`

```console
$ docker run -d --name ipfs_host -v $ipfs_staging:/export -v $ipfs_data:/data/ipfs -p 4001:4001 -p 127.0.0.1:8080:8080 -p 127.0.0.1:5001:5001 ipfs/go-ipfs:latest

$ docker exec ipfs_host ipfs version --commit
ipfs version 0.4.22-
```

The output of the ipfs verison command should be the same when docker is used to build it. 

We use docker built go-ipfs versions on the gateway and not being able to ask them for their commit will make reporting issues harder.

olizilla:
Some old prior art on this issue https://github.com/ipfs/go-ipfs/pull/2734

olizilla:
Looks like CI tests for this got disabled some time ago in this commit https://github.com/ipfs/go-ipfs/pull/2256/commits/c103efc1a9d3420b1f014a92a4ea7862ac13c57b as part of the move to gx https://github.com/ipfs/go-ipfs/pull/2256

gmasgras:
more context from https://github.com/protocol/bifrost-infra/issues/170#issuecomment-525930434

This is likely because we don't copy .git/objects) in the Docker build context.
see:
https://github.com/ipfs/go-ipfs/blob/master/cmd/ipfs/Rules.mk#L16
https://github.com/ipfs/go-ipfs/blob/master/mk/git.mk#L1
https://github.com/ipfs/go-ipfs/blob/master/.dockerignore#L3)

Stebalien:
Ah. This was broken in c8c0c48e2db2bf5ca683f1305c0d2c69c858d1ea.

Stebalien:
Should be fixed in https://github.com/ipfs/go-ipfs/pull/6626.

Stebalien:
Hm. Apparently, "append" is completely ignored when overridden in the command. I'm switching it to GOTAGS.

PR #6843, opened 2020-01-21
fix(build): instruct users to use GOTAGS, not GOFLAGS

Stebalien:
GOFLAGS will _override_ all the flags. If we just want to set a tag e.g., enable OpenSSL, we should set GOTAGS=openssl.

fixes https://github.com/ipfs/go-ipfs/issues/6622#issuecomment-576713181

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -425,7 +425,7 @@ receiving multiple inbound connections per second.
 To enable openssl support, rebuild go-ipfs with:
 
 ```bash
-> make build GOFLAGS=-tags=openssl
+> make build GOTAGS=openssl
 ```
 
 ### CoreAPI
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,7 +38,7 @@ RUN set -x \
 
 # Now comes the actual target image, which aims to be as small as possible.
 FROM busybox:1.31.0-glibc
-LABEL maintainer="Steven Allen <stven@stebalien.com>"
+LABEL maintainer="Steven Allen <steven@stebalien.com>"
 
 # Get the ipfs binary, entrypoint script, and TLS CAs from the build container.
 ENV SRC_DIR /go-ipfs
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,7 +20,7 @@ COPY . $SRC_DIR
 # Also: fix getting HEAD commit hash via git rev-parse.
 RUN cd $SRC_DIR \
   && mkdir .git/objects \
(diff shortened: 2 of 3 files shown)


olizilla:
I wasn't clear on the source of the issue, so I did some digging... and it seems some `make` lore has been invoked...

> If a variable has been set with a command argument (see Overriding Variables), then ordinary assignments in the makefile are ignored. 
https://www.gnu.org/software/make/manual/html_node/Override-Directive.html#Override-Directive

so when we 

```
$ make -n build GOFLAGS=-tags=openssl
```

...The append has no affect, as GOFLAGS was already defined.

https://github.com/ipfs/go-ipfs/blob/37884530221654fe57870db31ec7a178963b633e/mk/golang.mk#L9-L14

so the use of `+=` here is misleading, as you [pointed out](https://github.com/ipfs/go-ipfs/issues/6622#issuecomment-576757929).

To save us from future misreadings, either GOFLAGS should append, in which case we want to specify `override`, or we should assign the default value for `GOFLAGS` as part of the `?=` statement...

### Appending with `override`

```make
# Try to make building as reproducible as possible by stripping the go path.
override GOFLAGS += "-asmflags=all='-trimpath=$(GOPATH)'" "-gcflags=all='-trimpath=$(GOPATH)'"
$(warning $(GOFLAGS))
```
usage:

```console
$ make -n build GOFLAGS=-tags=openssl
mk/golang.mk:15: -tags=openssl "-asmflags=all='-trimpath=/Users/oli/Code/go'" "-gcflags=all='-trimpath=/Users/oli/Code/go'"

$ make -n build
mk/golang.mk:15:  "-asmflags=all='-trimpath=/Users/oli/Code/go'" "-gcflags=all='-trimpath=/Users/oli/Code/go'"
```

### No appending, just a default value

```make
# Try to make building as reproducible as possible by stripping the go path.
GOFLAGS ?= "-asmflags=all='-trimpath=$(GOPATH)'" "-gcflags=all='-trimpath=$(GOPATH)'"
unexport GOFLAGS
$(warning $(GOFLAGS))
```
usage

```console
$ make -n build GOFLAGS=-tags=openssl
mk/golang.mk:13: -tags=openssl

$ make -n build
mk/golang.mk:13: "-asmflags=all='-trimpath=/Users/oli/Code/go'" "-gcflags=all='-trimpath=/Users/oli/Code/go'"
```

Stebalien:
Yeah, I'll go through the variables in a followup PR.

PR #6854, opened 2020-01-30
Allow building non-amd64 Docker images

cap10morgan:
These changes allowed me to build arm Docker images to run on my Raspberry Pi. I used the new `docker buildx` feature to cross-build them. Let me know how else you'd like that documented / integrated into this repo.

The images I built are hosted here: https://hub.docker.com/repository/docker/cryptorado/go-ipfs

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -56,11 +56,11 @@ RUN chmod 4755 /usr/local/bin/fusermount
 RUN chmod 0755 /usr/local/bin/start_ipfs
 
 # This shared lib (part of glibc) doesn't seem to be included with busybox.
-COPY --from=0 /lib/x86_64-linux-gnu/libdl.so.2 /lib/libdl.so.2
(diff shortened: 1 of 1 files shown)


cap10morgan:
@Stebalien Thanks for merging! Since this is about building multi-arch images (rather than using the pre-built images which is what that section of the readme is about), is there some way I can help get the official images building arm64 and armv7 variants?

Stebalien:
We just let docker hub auto-build from github. I'm not sure how to convince docker hub to build cross-platform.

mburns:
https://hub.docker.com/r/ckulka/multi-arch-example seems like a reasonable guide.

I don't have permission to view/make changes on the Hub side, fwiw, but this seems straightforward if there is interest in making it happen.

Stebalien:
Nice!

Github version: https://github.com/ckulka/docker-multi-arch-example

I wonder if there's any way to avoid duplicating the docker file.

---

@mburns I've given you the correct permissions, I think. Go for it (when you have time).

PR #6855, opened 2020-01-31
doc(config): cleanup

Stebalien:
Trying to make this a bit more consistent and structured.

* Fix wrapping.
* Use headings for all config values.


Stebalien:
@dirkmc 

>  I'd suggest also adding an example where the default isn't shown in the config (eg Bootstrap)

?

PR #6867, opened 2020-02-04
Help text update for 'ipfs key gen'

RubenKelevra:
Add info about the default key type


Stebalien:
@achingbrain I'm seeing the interop tests repeatedly fail but nothing should have changed this on the go-ipfs side. Could you take a look?

PR #6882, opened 2020-02-07
Fix #6878: Improve MFS Cli documentation 

### LINKED ISSUE #6878: Garbage Collector doesn't clean up MFS items
#### Version information:
go-ipfs version: 0.4.23-6ce9a355f
Repo version: 7
System version: amd64/linux
Golang version: go1.13.7

#### Description:
I couldn't find any hin in the CLI documentation, that this should happen so it's for me quite a surprising behavior:

When I create a folder or write a file the item isn't pinned (and there's no option to change this behavior). So I expected, that I need to pin the files, to preserve them from the garbage collector.

But while testing I created a folder and added a file to the folder, run GC manually and the folder and the file still exist.

I expected that the file get removed, and since the folder is then empty, it is removed as well.

I understand, that these would lead to changes in the file structure and thus to a different CID of the folder.

So I expected that either the folder gets removed when the last file has been unpinned (with all other files) or the files are just no longer accessible but are still listed with their CID.

This behavior gets especially strange when you add a file with ipfs add, then copy (`ipfs files cp /ipfs/$CID /$folder/$filename`) it to a folder, and then unpin the file since you don't need it anymore - it will still remain in the repo, but just because it exists in the file structure.

And the other way around too, you pin a file, delete it from the MFS but the GC cannot release the storage since it's somewhere deep in the `ipfs pin ls` list without a file name and you probably never find it again when you have pinned a lot of data.

hsanjuan:


core/commands/files.go
```diff
@@ -39,6 +39,18 @@ var FilesCmd = &cmds.Command{
...
+The files facility interacts with MFS (Mutable File System). MFS acts as a
+single, dynamic filesystem mount. MFS has a root CID which is transparently
+updated when a change happens (and can be checked with "ipfs files stat /").
+
+All files and folders within MFS are respected and will not be cleaned up
+during garbage collections. MFS is independent from the list of pinned items
+("ipfs pin ls"). Calls to "ipfs pin add" and "ipfs pin rm" do not affect
+content on MFS. Similarly, content added with "ipfs add" (which by default
+pins), is not added to MFS. Any content can be put into MFS with the command
+"ipfs files cp /ipfs/<cid> /mfs/path/".
```
mohe2015:
```suggestion
"ipfs files cp /ipfs/<cid> /some/path/".
```
As far as I know the path doesn't need to start with `/mfs` or does it? If it doesn't this may make this more clear.

RubenKelevra: Good catch!

RubenKelevra:
Nice!

But it's missing the information that content can be pinned and part of the MFS, so deleting pinned files and folders won't release the pin and therefore not release the storage for GC. :)

hsanjuan:
> Nice!
> 
> But it's missing the information that content can be pinned and part of the MFS, so deleting pinned files and folders won't release the pin and therefore not release the storage for GC. :)

`Calls to "ipfs pin add" and "ipfs pin rm" do not affect content on MFS.` isn't this exactly that?

core/commands/files.go
hsanjuan: thx

RubenKelevra:
Maybe:

`Calls to "ipfs pin add" and "ipfs pin rm" will add and remove an additionally pin on the content on MFS. If pinned MFS content is removed by calling "ipfs files rm", it will remain pinned.`

core/commands/files.go
```diff
@@ -39,6 +39,22 @@ var FilesCmd = &cmds.Command{
...
 filesystem.
 
+The files facility interacts with MFS (Mutable File System). MFS acts as a
+single, dynamic filesystem mount. MFS has a root CID which is transparently
+updated when a change happens (and can be checked with "ipfs files stat /").
+
+All files and folders within MFS are respected and will not be cleaned up
+during garbage collections. MFS is independent from the list of pinned items
+("ipfs pin ls"). Calls to "ipfs pin add" and "ipfs pin rm" will add and remove
+an separate pin independent of the content on MFS. If MFS content that was
```
Stebalien:
```suggestion
pins independently of MFS. If MFS content that was
```

PR #6887, opened 2020-02-10
feat: add peerlog plugin

olizilla:
add a plugin to log just PeerIDs for nodes we connect to, to allow us to see how many unique peers a node connects to over time.

It's set up a preloaded plugin, so it would be enabled by default. We may want to extract it to a seperate repo so we can make it an optional add on.

**Usage**

```console
$ GOLOG_FILE=~/peer.log IPFS_LOGGING_FMT=json ipfs daemon
```

**Output**

```js
{"level":"info","ts":"2020-02-10T13:54:26.639Z","logger":"plugin/peerlog","caller":"peerlog/peerlog.go:51","msg":"connected","peer":"QmS2H72gdrekXJggGdE9SunXPntBqdkJdkXQJjuxcH8Cbt"}
{"level":"info","ts":"2020-02-10T13:54:59.095Z","logger":"plugin/peerlog","caller":"peerlog/peerlog.go:56","msg":"disconnected","peer":"QmS2H72gdrekXJggGdE9SunXPntBqdkJdkXQJjuxcH8Cbt"}
```

License: MIT


plugin/loader/preload_list
```diff
@@ -8,3 +8,4 @@ ipldgit github.com/ipfs/go-ipfs/plugin/plugins/git *
 badgerds github.com/ipfs/go-ipfs/plugin/plugins/badgerds *
 flatfs github.com/ipfs/go-ipfs/plugin/plugins/flatfs *
 levelds github.com/ipfs/go-ipfs/plugin/plugins/levelds *
+peerlog github.com/ipfs/go-ipfs/plugin/plugins/peerlog *
```
Stebalien: This shouldn't be added to the default preload list. We need to add a way to preload plugins on build.

plugin/plugins/peerlog/peerlog.go
```diff
@@ -0,0 +1,67 @@
...
+
+// Init initializes plugin
+func (*peerLogPlugin) Init(*plugin.Environment) error {
+	fmt.Println("peerLogPlugin enabled - PeerIDs will be logged")
+	return nil
+}
+
+func (*peerLogPlugin) Start(node *core.IpfsNode) error {
+	// Ensure logs from this plugin get printed regardless of global IPFS_LOGGING value
+	logging.SetLogLevel("plugin/peerlog", "info")
```
Stebalien: Apparently, this can return an error. We should do something with it. (panic?)

Stebalien:
Ok, I've fixed those issues. Let's see if this builds.

Stebalien:
Note: to enable the plugin, you now need to build with `make build IPFS_PLUGINS="peerlog"`.

olizilla:
Neat! Merge at will.

momack2:
🙌🚀❤️

PR #6906, opened 2020-02-18
chore: enable peerlog plugin on stabilize-dht branch

olizilla:
Hardcode the peerlog plugin to be enabled in the docker build for the stabilize-dht branch. This will enable the plugin for the bifrost-latest docker autobuild. Dockerhub doesnt allow us to set a build-arg for a specific build, hence this.

The base of this PR is feat/stabilize-dht, not master. The changeset includes a merge commit from master, to pull in the tweak to the peerlog plugin.


Stebalien:
Ok, that failing go test is not good. I think it might be related to something we did in the datastore.

Stebalien:
Oh, of course that's going to fail. It's using unroutable IP addresses.

PR #6914, opened 2020-02-20
Update to go-libp2p 0.6.0

marten-seemann:
This PR moves a bunch of code remove from go-libp2p-pnet to go-ipfs. This is necessary due to the change in the private network interface (see https://github.com/libp2p/go-libp2p/issues/795).


hsanjuan:
> This PR moves a bunch of code remove from go-libp2p-pnet to go-ipfs. This is necessary due to the change in the private network interface (see [libp2p/go-libp2p#795](https://github.com/libp2p/go-libp2p/issues/795)).

This is just the part about parsing the 3-line private network key file right? Are we sure this was not used outside of IPFS? It is going to be hard for people to find it (and use it) if they were relying on it. Might consider keeping and marking as deprecated in libp2p land?

Stebalien:
@hsanjuan has a good point. We might as well keep the _decoding_ code, at least, in go-libp2p-pnet so others can use it.

bonedaddy:
Please don't move deserialization here. This will give me another repository I need to fork and manage.

This is libp2p functionality, please have it stay as such.

core/node/libp2p/pnet.go
```diff
@@ -68,3 +72,85 @@ func PNetChecker(repo repo.Repo, ph host.Host, lc fx.Lifecycle) error {
 	})
 	return nil
 }
+
+func DecodeV1PSK(input io.Reader) (*[32]byte, error) {
```
Stebalien: TODO: Will need to switch back to depending on go-libp2p-core.

marten-seemann: @Stebalien Not sure I understand. Thinking about it though, should this return a `PSK`?

Stebalien: I mean we've moved this function to go-libp2p-core, we shouldn't define it twice.

Stebalien: Right?

marten-seemann: I removed it from https://github.com/libp2p/go-libp2p-pnet/pull/32 already. Let me update the function signature here now.

Stebalien: Np. I was just leaving this as a TODO so we wouldn't forget.

PR #6931, opened 2020-02-25
remove expired GPG key from README

willscott:
Encourage initial bug reports to security@ to be unencrypted for faster triage


Stebalien:
That is, given that few users have access to security@, we can reasonably share a single key. However, we can ask that submitters keep truly sensitive information out of the initial email.

Note: removing the key doesn't improve triage as we don't have many points of contact on that mailing list (and don't want many points of contact).

PR #6943, opened 2020-03-02
docs: fix example for files.write

alanshaw:
* Adds the `--parents` option to the first command so that it is runnable
* Updates the description to specify the option to use if you _do_ want nonexistant intermediate directories created

### DIFF
--- a/core/commands/files.go
+++ b/core/commands/files.go
@@ -686,14 +686,15 @@ a beginning offset to write to. The entire length of the input will be
 written.
 
 If the '--create' option is specified, the file will be created if it does not
-exist. Nonexistant intermediate directories will not be created.
+exist. Nonexistant intermediate directories will not be created unless the
+'--parents' option is specified.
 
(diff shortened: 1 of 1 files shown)


core/commands/files.go
```diff
@@ -686,22 +686,23 @@ a beginning offset to write to. The entire length of the input will be
...
-exist. Nonexistant intermediate directories will not be created.
+exist. Nonexistant intermediate directories will not be created unless the
+'--parents' option is specified.
 
 Newly created files will have the same CID version and hash function of the
-parent directory unless the --cid-version and --hash options are used.
+parent directory unless the '--cid-version' and '--hash' options are used.
 
 Newly created leaves will be in the legacy format (Protobuf) if the
 CID version is 0, or raw is the CID version is non-zero.  Use of the
```
johnnymatthews:
```suggestion
CID version is 0, or raw if the CID version is non-zero.  Use of the
```

Typo?

Issue #5947, opened 2019-01-27
IPNS's key names case sensitivity


AuHau:
#### Version information:
go-ipfs version: 0.4.18-
Repo version: 7
System version: amd64/darwin
Golang version: go1.11.1

#### Type:
bug / doc

#### Description:

Is IPNS key name's matching case insensitive? (By accident) I tried to create two keys with different case sensitivity and got an error that the key name already exists.

I am aware of the base-32 migration, but would not expect this to be a case for the key names. If it is not a bug, but a feature, then I would suggest document it somewhere (unless I have missed it).

Stebalien:
You mean the key name from `ipfs key gen <name>`? That's probably due to your filesystem. Unlike everything else (which we store in the "repo"), we store keys directly in the filesystem.

Although, really, we should consider encoding the names before storing them.

Stebalien:
Encoding (base32) would be great, IMO. However, that'll require a migration: https://github.com/ipfs/fs-repo-migrations/. Luckily, it should be a pretty simple migration relative to most of the migrations in that repo.

Kubuxu:
I question that we want to encode it. It is just adding complexity.
I would propose that we enforce all newly created keys being lower case. This way moving between OS is not a problem and user confusion is also reduced.

@Stebalien what do you think?

AuHau:
Agree that it would be a simpler solution, but what about the existing keys? If you are talking about moving between OSes, then for old keys that could be still a problem. If there would be a migration that would make all existent keys lowercase, that could break things. Also, this approach will result in a change of behavior, that I think is more prone to user's confusion, then encoding itself.

Encoding does not change any behavior, only unites it across all platforms.

That said I don't have any hard preferences, maybe bit leaning to the encoding approach.

Stebalien:
> I have a clarification question though, we are talking about encoding the name of the keys right?

Yes.

>  Do you have knowledge of how many people interacts directly with the key's files and not your API?

If anyone is messing with the keys in the repo directly, they'll have to change their code to match the new behavior.

---

@Kubuxu my worry is that different platforms have different constraints. We could just whitelist `[a-z][0-9]` but users should be able to use any character set they want.

Stebalien:
@AuHau If you'd like to work on this, please go ahead with the encoding. We should store keys as `ENCODE(name).key`, or something like that.

AuHau:
Alright! Sounds good!
I should have something ready in a week time.

[...]

PR #6955, opened 2020-03-05
Introducing EncodedFSKeystore with base32 encoding (#5947)

hsanjuan:
Ported from https://github.com/ipfs/go-ipfs/pull/6012

---

Encoding the key's filename with base32 introduces coherent behaviour
across different platforms and their case-sensitive/case-insensitive
file-systems. Moreover it allows wider character set to be used for the
name of the keys as the original restriction for special FS's characters
(e.g. '/', '.') will not apply.

License: MIT
Signed-off-by: Adam Uhlir <uhlir.a@gmail.com>

### DIFF
--- a/keystore/keystore.go
+++ b/keystore/keystore.go
@@ -7,12 +7,15 @@ import (
 	"path/filepath"
 	"strings"
 
+	base32 "encoding/base32"
+
 	logging "github.com/ipfs/go-log"
 	ci "github.com/libp2p/go-libp2p-core/crypto"
-	base32 "github.com/whyrusleeping/base32"
 )
 
 var log = logging.Logger("keystore")
+
+var codec = base32.StdEncoding.WithPadding(base32.NoPadding)
 
 // Keystore provides a key management interface
 type Keystore interface {
@@ -29,46 +32,20 @@ type Keystore interface {
 	List() ([]string, error)
 }
 
+// ErrNoSuchKey is an error message returned when no key of a given name was found.
 var ErrNoSuchKey = fmt.Errorf("no key by the given name was found")
+
+// ErrKeyExists is an error message returned when a key already exists
 var ErrKeyExists = fmt.Errorf("key by that name already exists, refusing to overwrite")
+
+const keyFilenamePrefix = "key_"
 
 // FSKeystore is a keystore backed by files in a given directory stored on disk.
 type FSKeystore struct {
 	dir string
 }
 
-func validateName(name string) error {
-	if name == "" {
-		return fmt.Errorf("key names must be at least one character")
-	}
-
-	if strings.Contains(name, "/") {
-		return fmt.Errorf("key names may not contain slashes")
-	}
-
-	if strings.HasPrefix(name, ".") {
-		return fmt.Errorf("key names may not begin with a period")
-	}
-
-	return nil
-}
-
-// NewKeystore is a factory for getting instance of Keystore interface implementation
-func NewKeystore(dir string) (Keystore, error) {
-	return NewEncodedFSKeystore(dir)
-}
-
-// NewEncodedFSKeystore is a factory for getting instance of EncodedFSKeystore
-func NewEncodedFSKeystore(dir string) (*EncodedFSKeystore, error) {
-	keystore, err := NewFSKeystore(dir)
-
-	if err != nil {
-		return nil, err
-	}
-
-	return &EncodedFSKeystore{keystore}, nil
-}
-
+// NewFSKeystore returns a new filesystem-backed keystore.
 func NewFSKeystore(dir string) (*FSKeystore, error) {
 	_, err := os.Stat(dir)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


hsanjuan:
Ready on my side, I made minor changes over what was already approved. I tagged a test version and built a custom dist release so that it can be tested with https://github.com/ipfs/fs-repo-migrations/pull/96/files

keystore/keystore.go
```diff
@@ -52,6 +57,24 @@ func validateName(name string) error {
...
 }
 
+// NewKeystore returns a Keystore using the default implementation.
+func NewKeystore(dir string) (Keystore, error) {
+	return NewEncodedFSKeystore(dir)
+}
+
+// NewEncodedFSKeystore is returns a filesystem-backed keystore which encodes
+// key names using base32.
+func NewEncodedFSKeystore(dir string) (*EncodedFSKeystore, error) {
```
Stebalien: I'd just have one keystore. Supporting both seems like extra and unnecessary work.

keystore/keystore.go
```diff
@@ -174,3 +197,108 @@ func (ks *FSKeystore) List() ([]string, error) {
...
+
+	nameWithoutPrefix := strings.ToUpper(name[len(keyFilenamePrefix):])
+	decoder := base32.StdEncoding.WithPadding(base32.NoPadding)
+	data, err := decoder.DecodeString(nameWithoutPrefix)
+
+	if err != nil {
+		return "", err
+	}
+
+	decodedName := string(data[:])
```
Stebalien: Why slice?

keystore/keystore.go
```diff
@@ -174,3 +197,108 @@ func (ks *FSKeystore) List() ([]string, error) {
...
+	return keyFilenamePrefix + strings.ToLower(encodedName), nil
+}
+
+func decode(name string) (string, error) {
+	if !strings.HasPrefix(name, keyFilenamePrefix) {
+		return "", fmt.Errorf("key's filename has unexpected format")
+	}
+
+	nameWithoutPrefix := strings.ToUpper(name[len(keyFilenamePrefix):])
+	decoder := base32.StdEncoding.WithPadding(base32.NoPadding)
```
Stebalien: Global (private) variable?

keystore/keystore.go
```diff
@@ -174,3 +197,108 @@ func (ks *FSKeystore) List() ([]string, error) {
...
 }
+
+const keyFilenamePrefix = "key_"
+
+func encode(name string) (string, error) {
+	if name == "" {
+		return "", fmt.Errorf("key name must be at least one character")
+	}
+
+	encoder := base32.StdEncoding.WithPadding(base32.NoPadding)
```
Stebalien: global private variable?

keystore/keystore.go
```diff
@@ -174,3 +197,108 @@ func (ks *FSKeystore) List() ([]string, error) {
...
+	}
+
+	list := make([]string, 0, len(dirs))
+
+	for _, name := range dirs {
+		decodedName, err := decode(name)
+		if err == nil {
+			list = append(list, decodedName)
+		} else {
+			log.Warnf("Ignoring keyfile with invalid encoded filename: %s", name)
```
Stebalien: Let's log an error. This is bad.

keystore/keystore.go
hsanjuan: this was an old PR that was already approved so I just did minimal maintenance, but ok.

keystore/keystore.go
```diff
@@ -36,42 +36,16 @@ var ErrNoSuchKey = fmt.Errorf("no key by the given name was found")
...
-	if strings.Contains(name, "/") {
-		return fmt.Errorf("key names may not contain slashes")
-	}
-
-	if strings.HasPrefix(name, ".") {
-		return fmt.Errorf("key names may not begin with a period")
-	}
-
-	return nil
-}
```
hsanjuan: I am guessing this is no longer necessary.

Stebalien: Yep.

keystore/keystore.go
```diff
@@ -28,30 +32,25 @@ type Keystore interface {
...
-		return fmt.Errorf("key names may not contain slashes")
-	}
-
-	if strings.HasPrefix(name, ".") {
-		return fmt.Errorf("key names may not begin with a period")
-	}
-
-	return nil
+// NewKeystore returns a Keystore using the default implementation.
+func NewKeystore(dir string) (Keystore, error) {
```
Stebalien: We don't need this function either.

hsanjuan: hmm we do

hsanjuan: removed properly now

hsanjuan:
@Stebalien I have squashed the commit history.

However we need to merge https://github.com/ipfs/fs-repo-migrations/pull/96  and release fs-repo-migrations before proceeding here.

Stebalien:
I've cut a release: https://github.com/ipfs/distributions/pull/285.

Stebalien:
(well, I haven't _published_ it yet but I figured we should test it first)

PR #6972, opened 2020-03-10
feat(peerlog): log protocols/versions

Stebalien:


plugin/plugins/peerlog/peerlog.go
```diff
@@ -60,6 +63,38 @@ func (*peerLogPlugin) Start(node *core.IpfsNode) error {
 		)
 	}
 	node.PeerHost.Network().Notify(&notifee)
+
+	sub, err := node.PeerHost.EventBus().Subscribe(
+		new(event.EvtPeerIdentificationCompleted),
+		eventbus.BufSize(1024),
```
olizilla: for learnings, is there any reason for 1024 here, or just a nice number?

Stebalien: It's a nice round number. Basically, if we slow down handling these events and fill this buffer, identifying new peers may block. This gives us some wiggle room in case logging slows down.

Stebalien:
> for consistency with what's out there, perhaps we should revert to not prefixing it with a slash. it otherwise looks like a multiaddr which is confusing.

Fixed in #6993.

Issue #6975, opened 2020-03-11
Update subdomain redirection logic to only engage on browsers


momack2:
From conversation with @stebalien and @jbenet we should only redirect to a subdomain gateway if the user is detected to be a browser node so as to avoid breaking curl and wget usage (curl doesn't auto-redirect without passing -L and wget won't span hosts on redirect). Note this is needed for 0.5 to avoid breaking existing usage.

Required for https://github.com/ipfs/go-ipfs/issues/6776

momack2:
Ex what ipfs-search does: 
`curl http://localhost:8081/ipfs/QmS4ustL54uo8FzR9455qaxZwuMiUhyvMcX9Ba8nUH4uVv/readme`

Stebalien:
Potential solution: Just return the page with the redirect response but set the "Clear-Site-Data" header. CURL will download the page and won't follow the redirect (and will even politely exit with a zero status by default). Browsers should follow the redirect.

lidel:
@Stebalien I implemented that idea (**Option A** for future reference) in https://github.com/ipfs/go-ipfs/pull/6982. Let me know what you think.

Personally I don't feel too confident with it, produces valid responses, but feels like a hack that could bite us in the future.

What if we go with a much simpler hack (**Option B**):
- see if `User-Agent` match `^(curl|wget)/i`
  - add `Clear-Site-Data` header
  - disable subdomain, no redirect, return regular HTTP 200 response

Stebalien:
My concern is that there are probably other tools that don't follow redirects either (although this looks rare). Option A is the most general option available, as far as I know, and it doesn't rely on user agent sniffing.

If it does bite us, I believe we can switch to option B in the future, right?

 > I feel it is a bit safer and more explicit behavior.

How?

----

I'm fine just white listing curl if option A causes issues, it's just that user-agent sniffing makes me uncomfortable.

lidel:
> I feel [user-agent sniff] is a bit safer and more explicit behavior.
> How?

In my mind User-Agent sniff (**B**) would be bit safer because if it breaks something, user can change/remove `User-Agent` sent to go-ipfs and control gateway behavior that way:

```
$ curl -H "User-Agent:" 
```

If there is an issue with redirect+payload (**A**) user can't easily fix it on the client side alone, they need to either wait for new go-ipfs or put a reverse proxy in front of go-ipfs and fix responses on the fly. 

[...]

PR #6982, opened 2020-03-11
fix(subdomain-gw): curl on localhost (Option A: HTTP301+payload)

lidel:
> This is a PR against [`feat/gateway-subdomains` branch](https://github.com/ipfs/go-ipfs/pull/6096) to resolve concerns from https://github.com/ipfs/go-ipfs/issues/6975

### TL;DR

Return payload with HTTP 301 response:
- curl returns exit code 0 on 301, so no script will fail
- web browsers close connection after reading HTTP 301 and `Location` header, so there should not be any bandwidth wasted
  - just to be sure browsers don't store anything we return `Clear-Site-Data` header with such response

### Description 
When request is sent to `http://localhost:8080/ipfs/$cid` HTTP 301 status code with  "Location" header redirect client to subdomain destination at `$cid.ipfs.localhost:8080`

Redirects are not followed by `curl` in default mode, but will be respected by user agents that follow redirects by default, namely web browsers.

To fix curl, we return correct payload in body of HTTP 301 response, but set `Clear-Site-Data` header to ensure Origin sandbox can't be abused.

### Caveat

<del>This comes with side effect of prometeus client making redundant `WriteHeader` call, which produces warning in logs:</del>

> <del>`2020/03/11 14:57:06 http: superfluous response.WriteHeader call from github.com/prometheus/client_golang/prometheus/promhttp.(*responseWriterDelegator).WriteHeader (delegator.go:58)`</del>

<del>
I attempted to fix prometeus and avoid redundant call in https://github.com/prometheus/client_golang/pull/724 but am not sure if that approach will be approved upstream.</del>

**Update:** `superfluous response.WriteHeader call` fixed in https://github.com/ipfs/go-ipfs/commit/e705f023bbfb00530f73cc60e763d4705c2d4613

### DIFF
--- a/core/corehttp/gateway_handler.go
+++ b/core/corehttp/gateway_handler.go
@@ -38,6 +38,25 @@ type gatewayHandler struct {
 	api    coreiface.CoreAPI
 }
 
+// StatusResponseWriter enables us to override HTTP Status Code passed to
+// WriteHeader function inside of http.ServeContent.  Decision is based on
+// presence of HTTP Headers such as Location.
+type statusResponseWriter struct {
+	http.ResponseWriter
+}
+
+func (sw *statusResponseWriter) WriteHeader(code int) {
+	// Check if we need to adjust Status Code to account for scheduled redirect
+	// This enables us to return payload along with HTTP 301
+	// for subdomain redirect in web browsers while also returning body for cli
+	// tools which do not follow redirects by default (curl, wget).
+	redirect := sw.ResponseWriter.Header().Get("Location")
+	if redirect != "" && code == http.StatusOK {
+		code = http.StatusMovedPermanently
+	}
+	sw.ResponseWriter.WriteHeader(code)
+}
+
 func newGatewayHandler(c GatewayConfig, api coreiface.CoreAPI) *gatewayHandler {
 	i := &gatewayHandler{
 		config: c,
@@ -366,6 +385,7 @@ func (i *gatewayHandler) serveFile(w http.ResponseWriter, req *http.Request, nam
 	}
 	w.Header().Set("Content-Type", ctype)
 
+	w = &statusResponseWriter{w}
 	http.ServeContent(w, req, name, modtime, content)
 }
 
--- a/core/corehttp/hostname.go
+++ b/core/corehttp/hostname.go
@@ -101,8 +101,24 @@ func HostnameOption() ServeOption {
 						// Yes, redirect if applicable
 						// Example: dweb.link/ipfs/{cid} → {cid}.ipfs.dweb.link
 						if newURL, ok := toSubdomainURL(r.Host, r.URL.Path, r); ok {
-							http.Redirect(w, r, newURL, http.StatusMovedPermanently)
-							return
+							// Just to be sure single Origin can't be abused in
+							// web browsers that ignored the redirect for some
+							// reason, Clear-Site-Data header clears browsing
+							// data (cookies, storage etc) associated with
+							// hostname's root Origin
+							// Note: we can't use "*" due to bug in Chromium:
+							// https://bugs.chromium.org/p/chromium/issues/detail?id=898503
+							w.Header().Set("Clear-Site-Data", "\"cookies\", \"storage\"")
+
+							// Set "Location" header with redirect destination.
+							// It is ignored by curl in default mode, but will
+							// be respected by user agents that follow
+							// redirects by default, namely web browsers
+							w.Header().Set("Location", newURL)
+
+							// Note: we continue regular gateway processing:
+							// HTTP Status Code http.StatusMovedPermanently
+							// will be set later, in statusResponseWriter
 						}
 					}
 
--- a/test/sharness/t0114-gateway-subdomains.sh
+++ b/test/sharness/t0114-gateway-subdomains.sh
@@ -141,9 +141,30 @@ test_localhost_gateway_response_should_contain \
 #  payload directly, but redirect to URL with proper origin isolation
 
 test_localhost_gateway_response_should_contain \
-  "request for localhost/ipfs/{CIDv1} redirects to subdomain" \
+  "request for localhost/ipfs/{CIDv1} returns status code HTTP 301" \
+  "http://localhost:$GWAY_PORT/ipfs/$CIDv1" \
+  "301 Moved Permanently"
+
+test_localhost_gateway_response_should_contain \
+  "request for localhost/ipfs/{CIDv1} returns Location HTTP header for subdomain redirect in browsers" \
   "http://localhost:$GWAY_PORT/ipfs/$CIDv1" \
   "Location: http://$CIDv1.ipfs.localhost:$GWAY_PORT/"
+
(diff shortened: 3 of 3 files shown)


core/corehttp/hostname.go
```diff
@@ -101,8 +101,32 @@ func HostnameOption() ServeOption {
...
+							if !hasCT {
+								w.Header().Set("Content-Type", "application/octet-stream")
+							}
+
+							// Send HTTP 301 status code and set "Location"
+							// header with redirect destination: it is ignored
+							// by curl in default mode, but will be respected
+							// by user agents that follow redirects by default,
+							// namely web browsers
 							http.Redirect(w, r, newURL, http.StatusMovedPermanently)
```
Stebalien: We shouldn't use `Redirect` as that will write a response. We need to manually set the response code and the location header.

But doing that looks tricky...

lidel: Yeah, in `http` apis for setting status code seemed to force write to the client.
I'll see if I can make it work somehow without refactoring entire gateway code.

lidel: @Stebalien take a look at my refactor in https://github.com/ipfs/go-ipfs/commit/e705f023bbfb00530f73cc60e763d4705c2d4613, it removes superfluous response.WriteHeader call. 

Gist: if `Location` header is present in ResponseWriter's `Header` map,
we ensure `http.ServeContent()` (called at the very end of processing pipeline) produces HTTP 301 response.

core/corehttp/gateway_handler.go
```diff
@@ -38,6 +39,29 @@ type gatewayHandler struct {
...
+	// Check if we need to adjust Status Code to account for scheduled redirect
+	// This enables us to return HTTP 301
+	// for subdomain redirect in web browsers while also returning body for cli
+	// tools which does not follow redirects by default (curl, wget).
+	redirect := sw.w.Header().Get("Location")
+	if redirect != "" {
+		code = http.StatusMovedPermanently
+	}
+	sw.w.WriteHeader(code)
+}
```
lidel: I came up with this surgical wrapper because [status code is hardcoded to `200`](https://golang.org/src/net/http/fs.go?s=4932:5036#L189) inside `http.ServeContent()`.

Is this idiomatic/acceptable in go? Not sure if we can make this change more elegant while keeping is small without duplicating `http.ServeContent`.

Stebalien: Wrapping is normal. Swapping the code like this is weird but fine.

core/corehttp/gateway_handler.go
```diff
@@ -38,6 +39,29 @@ type gatewayHandler struct {
 	api    coreiface.CoreAPI
 }
 
+// StatusResponseWriter enables us to override HTTP Status Code passed to
+// WriteHeader function inside of http.ServeContent.  Decision is based on
+// presence of HTTP Headers such as Location.
+type statusResponseWriter struct {
+	w http.ResponseWriter
```
Stebalien: If you use:

```go
 type statusResponseWriter struct {
  http.ResponseWriter
}
```

You don't have to re-implement functions you're overriding: https://golang.org/doc/effective_go.html#embedding

lidel: Initially I tried that, but was not able to make embedding work. 

When request for http://localhost:8080/ipfs/bafkreicysg23kiwv34eg2d7qweipxwosdo2py4ldv42nbauguluen5v6am/  is being processed I get a crash with a stack overflow, possibly due to prometheus client doing own wrapping which produces a circular call(?):

```
fatal error: stack overflow

runtime stack:
runtime.throw(0x1ad5596, 0xe)
        /usr/lib/go/src/runtime/panic.go:774 +0x72
runtime.newstack()
        /usr/lib/go/src/runtime/stack.go:1046 +0x6e9
runtime.morestack()
        /usr/lib/go/src/runtime/asm_amd64.s:449 +0x8f

goroutine 1169 [running]:
net/http.(*response).Header(0xc0005a8380, 0xc020f423e8)
        /usr/lib/go/src/net/http/server.go:1056 +0x8a fp=0xc020f42360 sp=0xc020f42358 pc=0xd6de7a
go.(*struct { *github.com/prometheus/client_golang/prometheus/promhttp.responseWriterDelegator; io.ReaderFrom; net/http.Hijacker; net/http.Flusher; net/http.CloseNotifier }).Header(0xc0003c4960, 0xc0007a2000)
        <autogenerated>:1 +0x3f fp=0xc020f42380 sp=0xc020f42360 pc=0x1704adf
go.(*struct { *github.com/prometheus/client_golang/prometheus/promhttp.responseWriterDelegator; io.ReaderFrom; net/http.Hijacker; net/http.Flusher; net/http.CloseNotifier }).Header(0xc0003c49b0, 0xc0003308d0)
        <autogenerated>:1 +0x3f fp=0xc020f423a0 sp=0xc020f42380 pc=0x1704adf
github.com/ipfs/go-ipfs/core/corehttp.(*statusResponseWriter).WriteHeader(0xc000303d90, 0x12d)
        src/github.com/ipfs/go-ipfs/core/corehttp/gateway_handler.go:52 +0x33 fp=0xc020f423d8 sp=0xc020f423a0 pc=0x1714763
github.com/ipfs/go-ipfs/core/corehttp.(*statusResponseWriter).WriteHeader(0xc000303d90, 0x12d)
        src/github.com/ipfs/go-ipfs/core/corehttp/gateway_handler.go:56 +0x7d fp=0xc020f42410 sp=0xc020f423d8 pc=0x17147ad
github.com/ipfs/go-ipfs/core/corehttp.(*statusResponseWriter).WriteHeader(0xc000303d90, 0x12d)
        src/github.com/ipfs/go-ipfs/core/corehttp/gateway_handler.go:56 +0x7d fp=0xc020f42448 sp=0xc020f42410 pc=0x17147ad
github.com/ipfs/go-ipfs/core/corehttp.(*statusResponseWriter).WriteHeader(0xc000303d90, 0x12d)
        src/github.com/ipfs/go-ipfs/core/corehttp/gateway_handler.go:56 +0x7d fp=0xc020f42480 sp=0xc020f42448 pc=0x17147ad
github.com/ipfs/go-ipfs/core/corehttp.(*statusResponseWriter).WriteHeader(0xc000303d90, 0x12d)
        src/github.com/ipfs/go-ipfs/core/corehttp/gateway_handler.go:56 +0x7d fp=0xc020f424b8 sp=0xc020f42480 pc=0x17147ad
github.com/ipfs/go-ipfs/core/corehttp.(*statusResponseWriter).WriteHeader(0xc000303d90, 0x12d)
[...]
```

Stebalien: Ah, you were recursively calling `sw.WriteHeader`. You'd need to call `sw.ResponseWriter.WriteHeader`.

lidel: Thanks! Rebased, this should be fixed in https://github.com/ipfs/go-ipfs/pull/6982/commits/9fe44b7fc29e3845a74283b6ec4a644b75fcd281

PR #6983, opened 2020-03-11
fix(subdomain-gw): curl on localhost (Option B: User-Agent)

lidel:
> This is a PR against [`feat/gateway-subdomains` branch](https://github.com/ipfs/go-ipfs/pull/6096) to resolve concerns from https://github.com/ipfs/go-ipfs/issues/6975
> See [Option A](https://github.com/ipfs/go-ipfs/pull/6982) for alternative fix.

### TL;DR

When request to `localhost/ipfs/*` is processed:
- add `Clear-Site-Data` header
- see if `User-Agent` is matching known cli tool `^(curl|wget)/i`
  - on positive match, return regular HTTP 200 response instead of subdomain redirect


lidel:
Oh boy, wrong target, ignore this PR.

PR #6990, opened 2020-03-12
feat: don't put the block back with the other CID version.

Stebalien:
We used to do this so we'd provide the block with the correct CID version.
However, we've switched to providing raw multihashes in the DHT so this is no
longer useful.

### DIFF
--- a/gc/gc.go
+++ b/gc/gc.go
@@ -28,6 +28,33 @@ type Result struct {
 	Error      error
 }
 
+func upgradeCid(c cid.Cid) cid.Cid {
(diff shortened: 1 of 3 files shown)


Stebalien:
I'm punting this till 0.6

Stebalien:
Status:

1. We need to fix the tests.
2. We'd like this logic to be generic over "raw mutlihash versus CIDv1" instead of "CIDv0 versus CIDv1". That way, downgrading from a "raw multihash only" implementation should be easier. This will require changing the GC logic.

PR #6993, opened 2020-03-13
fix: revert changes to the user agent

Stebalien:
Go-ipfs user agents usually look like go-ipfs/VERSION/COMMIT_or_EMPTY. However,
we changed this to /go-ipfs/VERSION(/COMMIT)? on master for a while. This change
reverts this to make parsing the user agent simpler.

### DIFF
--- a/test/sharness/t0026-id.sh
+++ b/test/sharness/t0026-id.sh
@@ -1,0 +1,38 @@
+#!/usr/bin/env bash
+
+test_description="Test to make sure our identity information looks sane"
+
+. lib/test-lib.sh
+
+test_init_ipfs
+
+test_id_compute_agent() {
+    AGENT_VERSION="$(ipfs version --number)" || return 1
+    AGENT_COMMIT="$(ipfs version --number --commit)" || return 1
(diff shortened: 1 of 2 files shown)


test/sharness/t0026-id.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
+
+test_description="Test to make sure out identity information looks sane"
```
olizilla:
```suggestion
test_description="Test to make sure our identity information looks sane"
```

olizilla:
hmm odd that the docker build fails with

```
github.com/blang/semver@v3.6.1+incompatible: invalid version: +incompatible suffix not allowed: module contains a go.mod file, so semantic import versioning is required
The command '/bin/sh -c cd $SRC_DIR   && go mod download' returned a non-zero code: 1
```

but `build` succeeds.

Stebalien:
Because that's the only CI that tests macos?

Stebalien:
Ah, no, it's because upstream messed up.

PR #6998, opened 2020-03-16
Support pipes when named on the cli explicitly

ribasushi:
Still throws an error when a pipe is encountered during directory recursion.

~**WILL NOT** pass CI without https://github.com/ipfs/go-ipfs-cmds/pull/184~


Stebalien:
Rebased and waiting for tests to pass.

PR #6999, opened 2020-03-16
fix(keystore): avoid racy filesystem access

Stebalien:
Instead of checking then performing a file operation, perform the file operation and check the error.

Merge to master after the base PR. I'm just filing this now so I don't forget.


hsanjuan:
hmm good catch

Issue #6870, opened 2020-02-04
DAG Import/Export


Stebalien:
We should have DAG Import/Export commands for importing from and exporting to CAR files.

Usage:

```
> ipfs dag import < my.car
QmRoot1
QmRoot2
> ipfs export SomeSelector > my.car
```

Stebalien:
Note: for now, `SomeSelector` should just be a CID.

rvagg:
@ribasushi 

Some test data:

 * https://github.com/rvagg/js-datastore-car/tree/master/test contains my best tests for CAR files, not too sophisticated but covers bases. `go.car` in that repo is the portable fixture I reuse across my various tests and expect to be consistent. See fixture-data.js for how this is built and managed through the testing process. The tests do various manipulations with permutations of writing and reading.
* ^ is based on the tests in https://github.com/ipfs/go-car/blob/master/car_test.go but adds CBOR blocks as well
* https://github.com/rvagg/go-car-util is just a bundle of experiments and doesn't have tests but may be interesting for inspecting data.
* https://github.com/rvagg/go-datastore-zipcar/blob/master/zipcar_test.go uses the same fixture data (although this is for `.zcar` files) but the Go code might be useful as it parallels the JS CAR code.

ribasushi:
@Stebalien remaining questions:
- I am lost on how to properly/elegantly "lock against GC" - too many options
- is `api.WithOptions(options.Api.Offline(true))` the proper way to do 
- why are we not supporting pipes: objections to fixing https://github.com/ipfs/go-ipfs/commit/1b26ca724f9dda116ab959db14e5bafc64bbf51b#diff-460d26c181d655b56384f413c5430c87R269-R274 ?

Stebalien:
> do we need to somehow integrate https://github.com/ipfs/go-verifcid in go-car ? currently only go-ipfs is "protected" and even there I am not entirely sure ( though tests check out )

The underlying blockstore verifies on read & write, so this shouldn't be necessary.

> is api.WithOptions(options.Api.Offline(true)) the proper way to do

Yes.

> why are we not supporting pipes: objections to fixing

UX issue around `ipfs add -r` adding fifos and blocking. But as we discussed on the call, allowing this for direct `ipfs add fifo` is probably fine.

pooja:
Hey all -- is this the correct PR to track for progress on this issue? https://github.com/ipfs/go-ipfs/pull/7038

Wondering what the overall status is!

[...]

PR #7011, opened 2020-03-18
DAG Import/Export

ribasushi:
Implements #6870.


core/commands/dag/dag.go
```diff
@@ -241,3 +261,259 @@ var DagResolveCmd = &cmds.Command{
...
+				}
+				// not ideal, but the pinning api takes only paths :(
+				rp := path.NewResolvedPath(
+					ipfspath.FromCid(c),
+					c,
+					c,
+					"",
+				)
+
+				if err := api.Pin().Add(req.Context, rp, options.Pin.Recursive(true)); err != nil {
```
Stebalien: this is probably the deadlock. `api.Pin` takes the pin lock internally so we don't GC while validating that we actually have the DAG.

We have to call the _non_ core API pinner (ipfsnode.Pinner), unfortunately.

Note: to debug things like this, it can be useful to take a stack dump with pprof.

ribasushi:
@Stebalien @rvagg @warpfork : this is the final implementation of the import/export functionality. The only thing missing are the sharness tests, which are coming some time Saturday: I am still not entirely happy with my test cases.

Sorry it took forever, there were a lot of... dusty corners to connect properly.

I specifically want someone to check my concurrency work ( around the progress meters ). I have not worked with such code a whole lot, and while there was a lot to crib from existing progress meters in ther commands, nothing was quite close to what I needed.

I will convert this into a proper PR once I push the sharness bits.

core/commands/dag/dag.go
```diff
@@ -241,3 +278,421 @@ var DagResolveCmd = &cmds.Command{
...
+				if ret.err != nil {
+					return ret.err
+				}
+				roots = ret.roots
+				if progressTicker != nil {
+					progressTicker.Stop()
+					if err := res.Emit(&CarImportOutput{ObjectCounts: counts}); err != nil {
+						return err
+					}
+					os.Stderr.WriteString("\n")
```
willscott: it's pretty weird that the return doesn't happen as part of the ResponseEmitter.

core/commands/dag/dag.go
```diff
@@ -241,3 +278,421 @@ var DagResolveCmd = &cmds.Command{
...
+
+				if block, err := node.Blockstore.Get(c); err != nil {
+					ret.PinError = err
+				} else if nd, err := ipld.Decode(block); err != nil {
+					ret.PinError = err
+				} else if err := node.Pinning.Pin(req.Context, nd, true); err != nil {
+					ret.PinError = err
+				} else if err := node.Pinning.Flush(req.Context); err != nil {
+					ret.PinError = err
+				}
```
willscott: failing to pin will set PinError, which is then emitted, but doesn't cause the overall command to return unsuccessfully. It may not be a fatal error - e.g. it seems okay to keep going through the other potential pinnable roots, but does seem like it may be worth reflecting in the command exit status.

core/commands/dag/dag.go
```diff
@@ -241,3 +278,421 @@ var DagResolveCmd = &cmds.Command{
...
+	// similar to pinner.Pin/pinner.Flush
+	batch := ipld.NewBatch(req.Context, (*api).Dag())
+
+	// It is not guaranteed that a root in a header is actually present in the same ( or any )
+	// .car file. This is the case in version 1, and ideally in further versions too
+	// Accumulate any root CID seen in a header, and supplement its actual node if/when encountered
+	// We will attempt a pin *only* at the end in case all car files were well formed
+	//
+	// The boolean value indicates whether we have encountered the root within the car file's
+	expectedRootsSeen := make(expectedRootsSeen)
```
willscott: prefer part of this comment to accompany the typedef higher up in the file, and also to not shadow the type with the instance.

core/commands/dag/dag.go
```diff
@@ -241,3 +278,421 @@ var DagResolveCmd = &cmds.Command{
...
+
+	it := req.Files.Entries()
+	for it.Next() {
+
+		file := files.FileFromEntry(it)
+		if file == nil {
+			ret <- importResult{err: errors.New("expected a file handle")}
+			return
+		}
+		defer func() { file.Close() }()
```
willscott: this will cause all the files to stay open until the end of the function. seems like worth having the file be closed within the scope - either factoring out the body into a function, or explicitly closing each loop iteration.

core/commands/dag/dag.go
```diff
@@ -241,3 +278,421 @@ var DagResolveCmd = &cmds.Command{
...
+				Root: c,
+				Selector: sb.ExploreRecursive(
+					gipselector.RecursionLimitNone(),
+					sb.ExploreAll(sb.ExploreRecursiveEdge()),
+				).Node(),
+			}},
+		)
+
+		if showProgress, _ := req.Options[progressOptionName].(bool); showProgress {
+			stopProgress := make(chan bool)
```
willscott: rather than the channel, might be more conventional at this point to make a `context.WithCancel`

core/commands/dag/dag.go
```diff
@@ -241,3 +278,421 @@ var DagResolveCmd = &cmds.Command{
...
+	ImportLoop:
+		for {
+			select {
+			case ret := <-retCh:
+				if ret.err != nil {
+					return ret.err
+				}
+				roots = ret.roots
+				if progressTicker != nil {
+					progressTicker.Stop()
```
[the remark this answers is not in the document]
ribasushi: This is where I emit ( coupe lines down ) a trailing "\n". The thinking was to avoid a garbled CLI display by:
- making sure no further updates take place
- do one last udate
- newline so the "\r" is neutralised

Am I missing something?

core/commands/dag/dag.go
ribasushi: I will push some unrelated fixes for this, so it may show outdated shortly. But the reason the error is carried through is to be able to provide the following ( I am breaking one of the pins artificially here ) :
```
admin:~/devel/go-ipfs$ go run ./cmd/ipfs dag import --enc=json --progress=false lotus_testnet_256.car devnet.car 
{"Root":{"Cid":{"/":"bafy2bzaceaxm23epjsmh75yvzcecsrbavlmkcxnva66bkdebdcnyw3bjrc74u"},"PresentInCar":true,"PinErrorMsg":""}}
{"Root":{"Cid":{"/":"bafy2bzacebedrc4n2ac6cqdkhs7lmj5e4xiif3gu7nmoborihajxn3fav3vdq"},"PresentInCar":true,"PinErrorMsg":""}}
{"Root":{"Cid":{"/":"bafy2bzacede2hsme6hparlbr4g2x6pylj43olp4uihwjq3plqdjyrdhrv7cp4"},"PresentInCar":true,"PinErrorMsg":"not today"}}
{"Root":{"Cid":{"/":"bafy2bzaceb55n7uxyfaelplulk3ev2xz7gnq6crncf3ahnvu46hqqmpucizcw"},"PresentInCar":true,"PinErrorMsg":""}}
```
Your point of *NOT* exitting with 0 when this happens is valid though: @Stebalien how does one do that? Also what does this mean for http 4xx-what?

core/commands/dag/dag.go
ribasushi: On CLI the responseemitter is converted to "stuff\r" here: https://github.com/ipfs/go-ipfs/pull/7011/files#diff-15172926c4147422f472139313467179R439 . In order to keep the stats on the screen I have to do a "\n" at one and one spot only. Do you have suggestions how to make this nicer?

core/commands/dag/dag.go
ribasushi: As a "custom" we don't close anything in the CLI in the first place, this is kinda new. By design `it` opens everything there is to open ahead of time ( I left a comment about that). But yeah good point, defer removed!

core/commands/dag/dag.go
ribasushi: Ugh... this is ( still ) above my head. I suspect it will be faster for me to grok how to do this in the future if you just push a commit directly addressing this...?

core/commands/dag/dag.go
```diff
@@ -303,7 +303,11 @@ pin each individual root specified in the car headers, before GC runs again.
 		cmds.BoolOption(pinRootsOptionName, "Pin optional roots listed in the .car headers after importing.").WithDefault(true),
 	},
 	PreRun: func(req *cmds.Request, env cmds.Environment) error {
-		if silent, _ := req.Options[silentOptionName].(bool); silent {
+		silent, _ := req.Options[silentOptionName].(bool)
+		encType, _ := req.Options[cmds.EncLong].(string)
+
+		// force-disable progress unless on non-silent CLI
+		if silent || encType != "text" {
```
ribasushi: @Stebalien is this reasonable to do here, or do we have a centralized way to switching on this? I suppose also having TTY detection would be nice somewhere centrally...

core/commands/dag/dag.go
```diff
@@ -241,3 +278,421 @@ var DagResolveCmd = &cmds.Command{
...
+		// isn't possible: tough luck
+		api, _ = api.WithOptions(options.Api.Offline(true))
+
+		// grab a gc lock so that regardless of the size of the streamed-in cars
+		// nothing will disappear on us before we had a chance to pin the roots
+		// at the very end
+		// This is especially important for use cases like dagger:
+		//    ipfs dag import $( ... | ipfs-dagger --stdout=carfifos )
+		//
+		// !!!NOTE!!! - We take .GCLock, *not* .PinLock. The way the current stack is
```
Stebalien: Take the pin lock. The GCLock is for the GC process.

core/commands/dag/dag.go
```diff
@@ -241,3 +278,421 @@ var DagResolveCmd = &cmds.Command{
...
+			}
+		}
+
+		// opportunistic pinning: try whatever sticks
+		if doPinRoots {
+
+			for c, seen := range *roots {
+
+				// We need to re-retrieve a block, convert it to ipld, and feed it
+				// to the Pinning interface, sigh...
```
Stebalien: We can change the pinning interfaces in a followup patch.

core/commands/dag/dag.go
Stebalien: To exit with a non-zero status code, return an error at the end. Something like "failed to recursively pin some roots".

In terms of the HTTP response, we indicate errors either by:

1. Returning 500.
2. Sending a error trailing HTTP header (if we discover the error after returning the first result.

core/commands/dag/dag.go
willscott: Can you have a message or flag sent on the ResponseEmitter that indicates "we're done", and have the encoder print a `\n` or otherwise clean up output when that happens? that way the printing all happens in one place.

ribasushi: @willscott you are correct, for a different reason. The emitter + post-run is what happens in the CLI tool - the actual run happens in a running damoen. The os.stderr are different... sigh.

I'll redo this shortly

core/commands/dag/dag.go
Stebalien: If we're going to check this, we should do so inside `PostRun`.

ribasushi: @Stebalien there are 2 things here, to clarify:
 
- Checking/resetting stuff coming from req.Options is ok in PreRun
- Checking for whether stderr is a tty cn only happen in PostRun

Correct?

Stebalien: > Checking/resetting stuff coming from req.Options is ok in PreRun

In general, yes. You can do this if you need to change some flag before it's sent to the server.

> Checking for whether stderr is a tty cn only happen in PostRun

Yes. PreRun and PostRun are always run on the local terminal. Run may run on the local terminal or may run on a remote terminal.

core/commands/dag/dag.go
```diff
@@ -241,3 +279,435 @@ var DagResolveCmd = &cmds.Command{
 	},
 	Type: ResolveOutput{},
 }
+
+// It is not guaranteed that a root in a header is actually present in the same ( or any )
+// .car file. This is the case in version 1, and ideally in further versions too
+// Accumulate any root CID seen in a header, and supplement its actual node if/when encountered
+// We will attempt a pin *only* at the end in case all car files were well formed
```
rvagg: > attempt a pin *only* at the end in case all car files were well formed

I can't quite see this in the code. It looks like it's done on a per-root basis, not per-car.

core/commands/dag/dag.go
```diff
@@ -241,3 +279,435 @@ var DagResolveCmd = &cmds.Command{
...
+				// rp := path.NewResolvedPath(
+				// 	ipfspath.FromCid(c),
+				// 	c,
+				// 	c,
+				// 	"",
+				// )
+				//
+				// if err := api.Pin().Add(req.Context, rp, options.Pin.Recursive(true)); err != nil {
+
+				ret := RootMeta{Cid: &c, PresentInCar: seen}
```
rvagg: What if `seen` is `false` but the CID exists in the blockstore already. I _think_ this would mean that it gets pinned regardless of the "malformed" CAR. Should the next block of if/else's be contained in an `if seen`?

core/commands/dag/dag.go
```diff
@@ -241,3 +279,435 @@ var DagResolveCmd = &cmds.Command{
...
+
+		// The second part of the above - make a super-thin wrapper around
+		// a blockservice session, translating Session.GetBlock() to Blockstore.Get()
+		// The function also doubles as our progress meter
+		var wrapper getBlockFromSessionWrapper = func(c cid.Cid) (blk.Block, error) {
+			nodeCount++
+			return sess.GetBlock(req.Context, c)
+		}
+		sb := gipselectorbuilder.NewSelectorSpecBuilder(gipfree.NodeBuilder())
+		car := gocar.NewSelectiveCar(
```
rvagg: Do we care about the case where you don't have a complete graph available? I believe that the `*` selector is going to result in an error if it reaches an edge that `sess.GetBlock()` can't return for a CID. We just can't currently disambiguate the error type from ipld-prime to know whether that may be acceptable (e.g. if a `--allow-incomplete` type flag was introduced), but we could probably change it to get one if that's needed. (https://github.com/ipld/go-ipld-prime/blob/eb71617f4aeb7e73f36276f14e76e47fd90241ef/traversal/walk.go#L119)

core/commands/dag/dag.go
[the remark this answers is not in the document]
ribasushi: I was missing... a lot of things, see https://github.com/ipfs/go-ipfs/pull/7011#discussion_r396024829 and https://github.com/ipfs/go-ipfs/pull/7011/commits/67642d8d5e7a5b56bd8e1b5f3c3e4f64f930aac9

core/commands/dag/dag.go
ribasushi: All fixed in https://github.com/ipfs/go-ipfs/pull/7011/commits/67642d8d5e7a5b56bd8e1b5f3c3e4f64f930aac9 and the orresponding export portion ( see commit msg ) : https://github.com/ipfs/go-ipfs/pull/7011/commits/3d9ed1a083efacb63aa7b2ec5effef8fdbe792bb

core/commands/dag/dag.go
ribasushi: Done in https://github.com/ipfs/go-ipfs/pull/7011/commits/aa85f2efada25768a7b38f6162efe11be4aba25a

core/commands/dag/dag.go
ribasushi: @Stebalien generalized as https://github.com/ipfs/go-ipfs/pull/7011/commits/de52029e1894de636f9c6953503111ffac6493ef

core/commands/dag/dag.go
ribasushi: Nod, likely not worth it before more ipld-prime stuff lands anyway. I am just leaving comments for the future self.

core/commands/dag/dag.go
ribasushi: Done: https://github.com/ipfs/go-ipfs/pull/7011/commits/0dae3a2248e6d149193362d997d75a15642fc094

I will open a separate issue for the test in question.

ribasushi: ( the sharness test gives me sufficient confidence the lock is enough, and we won't lose blocks )

core/commands/dag/dag.go
ribasushi: @rvagg on export we want to either output the complete graph ( potentially grabbing it from the network ), or to fail. I.e. at present we do not at all have `--allow-incomplete` in spirit, so that the .car files *produced* by go-ipfs are exclusively "well formed and deterministic" ( deterministic as dictated by go-ipld-prime ).

The import of course is a complete free-for-all ( also by design ) as can be seen here:
https://github.com/ipfs/go-ipfs/blob/03fe9bde87462aa11775bf6e8361a56fd967f89d/test/sharness/t0054-dag-car-import-export.sh#L41-L55
based on the datasets: https://github.com/ipfs/go-ipfs/blob/03fe9bde87462aa11775bf6e8361a56fd967f89d/test/sharness/t0054-dag-car-import-export-data/README.md

Let me know if this is potentially problematic...

core/commands/dag/dag.go
ribasushi: Correct - each individual root is opportunistically pinned, to maximize the benefit of what we just ingested being pinned.

core/commands/dag/dag.go
ribasushi: This is correct/by design: https://github.com/ipfs/go-ipfs/blob/03fe9bde87462aa11775bf6e8361a56fd967f89d/test/sharness/t0054-dag-car-import-export.sh#L83-L86

go.mod
```diff
@@ -54,6 +54,7 @@ require (
 	github.com/ipfs/go-unixfs v0.2.4
 	github.com/ipfs/go-verifcid v0.0.1
 	github.com/ipfs/interface-go-ipfs-core v0.2.6
+	github.com/ipld/go-car v0.0.5-0.20200316204026-3e2cf7af0fab
```
Stebalien: Go ahead and cut a CAR release when ready.

core/commands/dag/dag.go
```diff
@@ -241,3 +277,487 @@ var DagResolveCmd = &cmds.Command{
...
+			for {
+				select {
+				case ret := <-retCh:
+					if ret.err != nil {
+						return ret.err
+					}
+					roots = ret.roots
+
+					progressTicker.Stop()
+					if err := res.Emit(&CarImportOutput{ObjectCounts: &counts}); err != nil {
```
Stebalien: This isn't safe. `Emit` can do anything it wants with the object, including send it to another goroutine.

core/commands/dag/dag.go
```diff
@@ -241,3 +277,487 @@ var DagResolveCmd = &cmds.Command{
...
+		//
+		unlocker := node.Blockstore.PinLock()
+		defer func() { unlocker.Unlock() }()
+
+		doPinRoots, _ := req.Options[pinRootsOptionName].(bool)
+		showProgress, _ := req.Options[progressOptionName].(bool)
+
+		retCh := make(chan importResult, 1)
+		var counts ObjectCounts
+		go importWorker(req, &res, &api, &counts, retCh)
```
Stebalien: This is _definitely_ not safe. We can't update counts from a different thread without a lock.

core/commands/dag/dag.go
```diff
@@ -241,3 +277,487 @@ var DagResolveCmd = &cmds.Command{
...
+				"Pinned root\t%s\t%s\n",
+				enc.Encode(*event.Root.Cid),
+				event.Root.PinErrorMsg,
+			)
+			return err
+		}),
+	},
+}
+
+func importWorker(req *cmds.Request, re *cmds.ResponseEmitter, api *iface.CoreAPI, counts *ObjectCounts, ret chan importResult) {
```
Stebalien: Taking interfaces by pointer is an anti-pattern. Just pass the interfaces by value.

core/commands/dag/dag.go
```diff
@@ -241,3 +277,487 @@ var DagResolveCmd = &cmds.Command{
...
 	Type: ResolveOutput{},
 }
+
+// It is not guaranteed that a root in a header is actually present in the same ( or any )
+// .car file. This is the case in version 1, and ideally in further versions too
+// Accumulate any root CID seen in a header, and supplement its actual node if/when encountered
+// We will attempt a pin *only* at the end in case all car files were well formed
+//
+// The boolean value indicates whether we have encountered the root within the car file's
+type expectedRootsSeen map[cid.Cid]bool
```
Stebalien: This type just obfuscates the underlying type and doesn't really gain us anything.

core/commands/dag/dag.go
```diff
@@ -241,3 +277,487 @@ var DagResolveCmd = &cmds.Command{
...
+type importResult struct {
+	roots *expectedRootsSeen
```
Stebalien: Maps are already allocated on the heap. We don't need the pointer.

core/commands/dag/dag.go
```diff
@@ -241,3 +277,487 @@ var DagResolveCmd = &cmds.Command{
...
+								os.Stderr,
+								"Exported .car size:\t%d\r",
+								exportSize,
+							)
+						}
+
+						if readErr == io.EOF {
+							return re.Close()
+						} else if readErr != nil {
+							return re.CloseWithError(err)
```
Stebalien: You can call `re.CloseWithError` in both cases. Calling with an "eof" is equivalent to calling `Close()`.

core/commands/dag/dag.go
```diff
@@ -241,3 +277,487 @@ var DagResolveCmd = &cmds.Command{
...
+					return re.CloseWithError(err)
+				}
+
+				if r, ok := v.(io.Reader); ok {
+					// we got a reader passed as a response
+					// proxy it through with an increasing counter
+					for {
+						len, readErr := r.Read(buf)
+						if len > 0 {
+							if err := re.Emit(bytes.NewBuffer(buf[:len])); err != nil {
```
Stebalien: This may happen to work (should add a check to prevent it) when not running the daemon, but simply won't work when actually running the daemon. When running the daemon, we can only return:

1. Repeated results of the same type (encoded as json). To do this, we have to specify the type.
2. Streaming bytes (no type specified).

We can't return both.

Stebalien:
Returning progress on `export` seems insanely complicated and not particularly useful unless we already know the number of blocks to expect.

Additionally, the current method simply won't _work_ with the current commands library. We can't mix data blocks and objects.

Stebalien:
Let's just drop progress on get. Progress on _add_ is useful and should be pretty simple to implement.

core/commands/dag/dag.go
ribasushi: The reason `&counts` is a pointer ( here and your next comment ) is that this is the only way to activate `omitempty` higher up: hhttps://github.com/ipfs/go-ipfs/pull/7011/files#diff-15172926c4147422f472139313467179R73

Want to make sure you agree with the rationale before I add a bunch of mutexes.

core/commands/dag/dag.go
ribasushi: It is a ternary:
- no idea
- seen mentioned as root
- seen mentioned as root and saw the corresponding payload of the block

Can you elaborate?

core/commands/dag/dag.go
ribasushi: I wasn't sure abut both not being possible, so opted for the loop. Will simplify.

ribasushi:
> Returning progress on export seems insanely complicated and not particularly useful unless we already know the number of blocks to expect.

All I am doing is copying bytes from one pipe to another ( with a complication-conditional that I now realise is not needed ). `export` can mean going to the network. Without a progress of any sort it is not clear whether the dag is just humongous, or we are stuck on a particular spot of the DAG forever. Having this progress was also what allowed quick diagnosis of go-ipld-prime being broken. The entire pattern was copied over from: https://github.com/ipfs/go-ipfs/blob/master/core/commands/cat.go#L82-L103

> Additionally, the current method simply won't work with the current commands library. We can't mix data blocks and objects.

This code works 100%, I just elected not to add tests for it. I could do that if you are unconvinced :)

> Let's just drop progress on get.

I can do that, but having long-running network operations without any way to tell "is this thing alive?" is pretty bad UX. Please let me know this is really what you want to do.

> Progress on add is useful and should be pretty simple to implement.

Progress on add ( import ) is already in this PR... what am I missing?

core/commands/dag/dag.go
```diff
@@ -241,3 +277,487 @@ var DagResolveCmd = &cmds.Command{
...
+						}
+
+						if readErr == io.EOF {
+							return re.Close()
+						} else if readErr != nil {
+							return re.CloseWithError(err)
+						}
+					}
+				} else {
+					// some sort of encoded response, just get on with it
```
Stebalien: This case should be impossible. Let's log an error instead.

core/commands/dag/dag.go
Stebalien: Sorry, I misread this the first time. We _can_ repeatedly stream new byte readers.

However, we really shouldn't read 4MiB, write 4MiB, etc. Ideally, we'd print out a rate and amount downloaded.

core/commands/dag/dag.go
```diff
@@ -241,3 +277,487 @@ var DagResolveCmd = &cmds.Command{
...
+		cmds.BoolOption(silentOptionName, "No output."),
+		cmds.BoolOption(pinRootsOptionName, "Pin optional roots listed in the .car headers after importing.").WithDefault(true),
+	},
+	PreRun: func(req *cmds.Request, env cmds.Environment) error {
+		silent, _ := req.Options[silentOptionName].(bool)
+		encType, _ := req.Options[cmds.EncLong].(string)
+
+		if silent || encType != "text" {
+			// force-disable progress unless on non-silent CLI
+			req.Options[progressOptionName] = false
```
Stebalien: If I ask for progress but set the encoding type to `json`, I'd expect json progress updates. I wouldn't mess with this in pre-run.

core/commands/dag/dag.go
rvagg: better change the comment then, it seems misleading talking about the "car files" being "well formed" since you may find one root and pin that and not find another and not pin that.

core/commands/dag/dag.go
rvagg: this goes back to the question of "malformed" then, are you sure you want to have a lose edge like this? might be better to be a little more strict about the contents of a car - "you don't have the root you say you do, bad luck", instead you are opening up edges that could become important features for some in the future when we decide to be more strict with a future CAR format, or could lead to unexpected behaviour for others.

core/commands/dag/dag.go
```diff
@@ -241,3 +277,487 @@ var DagResolveCmd = &cmds.Command{
...
+		} else if _, specified := req.Options[progressOptionName]; !specified {
+			// enable progress implicitly if a TTY
+			errStat, _ := os.Stderr.Stat()
+			if 0 != (errStat.Mode() & os.ModeCharDevice) {
+				req.Options[progressOptionName] = true
```
Stebalien: I'd just set the progress option _default_ to true in the options. That way, the client can override it.

core/commands/dag/dag.go
rvagg: Comment above, it's not necessarily problematic, it just opens up API feature space that may be difficult to claw back in the future if we decide to be more strict. My personal inclination with such things these days is to be explicit and strict and open up feature space only in response to requests and demonstrated need (it also gives you a much more explicit space to construct tests around). Not my call here, so just a word of caution.

core/commands/dag/dag.go
Stebalien: I mean defining a custom `expectedRootsSeen` type when it's just a simple CID set just means I have to lookup the definition of `expectedRootsSeen` to understand what it is. The underlying type (`map[cid.Cid]bool`) makes sense.

core/commands/dag/dag.go
Stebalien: That's fine although I'm not sure why omitempty is important in this case. My point is that we should be copying this object safely instead of sharing it.

core/commands/dag/dag.go
```diff
@@ -241,3 +277,487 @@ var DagResolveCmd = &cmds.Command{
...
+	},
+	Arguments: []cmds.Argument{
+		cmds.StringArg("selector", true, false, "Selector representing the dag to export").EnableStdin(),
+	},
+	Options: []cmds.Option{
+		cmds.BoolOption(progressOptionName, "p", "Display progress on CLI. Defaults to true when STDERR is a TTY."),
+	},
+	PreRun: func(req *cmds.Request, env cmds.Environment) error {
+		// enable progress implicitly if a TTY
+		if _, specified := req.Options[progressOptionName]; !specified {
```
Stebalien: We can handle this entirely in PostRun.

Stebalien:
Ok, I've been mixing up parts of import/export. I'll need to re-read this but it looks more reasonable on a second pass.

However, I'd still like to punt some of this logic till a future version. This is very tricky code with subtle issues and addressing them all in one PR will be frustrating for all of us.

> This code works 100%, I just elected not to add tests for it. I could do that if you are unconvinced.

You're right, I was mixing up import/export and https://github.com/ipfs/go-ipfs/pull/7011/files#r398257738 (I thought this case was reachable).

core/commands/dag/dag.go
```diff
@@ -241,3 +277,487 @@ var DagResolveCmd = &cmds.Command{
...
+		}
+
+		return nil
+	},
+	Run: func(req *cmds.Request, res cmds.ResponseEmitter, env cmds.Environment) error {
+
+		c, err := cid.Decode(req.Arguments[0])
+		if err != nil {
+			return fmt.Errorf(
+				"unable to parse selector (currently only bare CIDs are supported): %s",
```
rvagg: What's the format of this argument when proper selectors are allowed? I assume it'll allow parsing of @creationix's new string-form selectors at some point, but aren't we still going to need to specify a root _separate_ to the selector? Is that going to need some extra syntax to squish CID+selector together or will this need to be 2 arguments, one for CID and one for selector? And in that case, should this argument just be renamed from "selector" to "root" for now with the expected addition of a selector argument to come later?

rvagg: See also https://github.com/ipld/go-car/blob/3e2cf7af0fab7d59ac0eb2ebc3dcad1c06fa68d0/selectivecar.go#L21-L24 where these two things are distinct.

warpfork: :+1: I advocate for a tuple here.

Selectors are a declarative document which, when interpreted, navigates from one starting node, to a series of reached nodes (being all those visited) and matched nodes (being those which the selector... selects).

Starting a selector at a particular CID is a common operation -- great; it composes cleanly with the above.

rvagg:
Been pondering the import semantics all day and want to continue on from my inline comment earlier: I think the first version of this should be fairly exclusive and not attempt to handle too much novelty. Rejecting zero-root CARs is probably a good idea for now (since it's essentially baked in go CARv1) and being more picky about how to handle the root-not-in-body case is probably a good idea too - it's badly formed. Don't even bother trying to pin that root.

There's two main reasons for this line of thinking:

1. Don't make it easy for the user to give you bad data or bad tooling will proliferate and persist.
2. Don't open your API too wide because it's much much harder to restrict an expansive API than to expand a restrictive one. Expand the functionality options in response to real, justified use-cases. Being too expansive is also going to make a more restrictive CARv2 more difficult, it'll end up being a more heavily breaking API when introduced here.

The two areas I can see justified for not making import a strict mirror of export in terms of its exclusiveness are: well ordered, structured and complete DAG ("deterministic", too hard, but we will want to make that reasonable to check in CARv2) and >1 roots (although export could easily be expanded here to accept more than one root already).

core/commands/dag/dag.go
ribasushi: Adressed here https://github.com/ipfs/go-ipfs/pull/7037/files#diff-15172926c4147422f472139313467179R418-R421

core/commands/dag/dag.go
ribasushi: We do not read/write 4MiB. This is the max we can read from the pipe at once, but other than that we just block and pull through whatever slice is placed on the pipe.

https://github.com/ipfs/go-ipfs/pull/7037/files#diff-15172926c4147422f472139313467179R418-R421

core/commands/dag/dag.go
ribasushi: All done in the new PR https://github.com/ipfs/go-ipfs/pull/7037

core/commands/dag/dag.go
ribasushi: They can override it ( slightly rewritten ): https://github.com/ipfs/go-ipfs/pull/7037/files#diff-15172926c4147422f472139313467179R363-R373

core/commands/dag/dag.go
ribasushi: Done

core/commands/dag/dag.go
ribasushi: Addressed as follows: https://github.com/ipfs/go-ipfs/pull/7036/files#diff-15172926c4147422f472139313467179R254-R263

The error text: https://github.com/ipfs/go-ipfs/pull/7036/files#diff-15172926c4147422f472139313467179R269

This should alleviate the duality ( @warpfork , @rvagg  please confirm )

ribasushi:
I have addressed all *EX* port pieces in a separate PR to make thins simpler to review.
The Import part coming up in another 2 PRs...

ribasushi:
New PR's:
export functionality only: https://github.com/ipfs/go-ipfs/pull/7036
export with cli progress: https://github.com/ipfs/go-ipfs/pull/7037
import functionality only: https://github.com/ipfs/go-ipfs/pull/7038

Regarding @rvagg comment earlier: I still feel that `dag import` should be nothing more than a more convenient `dag put`. This does mean that the streams it accepts would be .car files that are not valid anywhere else but... so what?

If the group feels this is not the direction we should pursue: I will add further strictness checks as requested, and will in parallel write something else to provide the "more convenient `dag put`" I need for other purposes :)

core/commands/dag/dag.go
ribasushi: Fixed in  https://github.com/ipfs/go-ipfs/pull/7038

core/commands/dag/dag.go
ribasushi: Fixed in https://github.com/ipfs/go-ipfs/pull/7038

core/commands/dag/dag.go
ribasushi: Fixed in https://github.com/ipfs/go-ipfs/pull/7038

ribasushi:
@Stebalien the sharness test is a fluke ( pubsub stuff failed ). In any case this PR is now too different from the "broken up" ones, I am closing it, and killing the branch, as all suggestions here got incorporated ( aside from the discussion on "is import ok as-is", which we will do in the import PR https://github.com/ipfs/go-ipfs/pull/7038

PR #7023, opened 2020-03-23
feat: improve key lookup

Stebalien:
* Support peer IDs encoded as CIDs.
* Support looking up the identity key by peer ID.

### DIFF
--- a/core/coreapi/name.go
+++ b/core/coreapi/name.go
@@ -139,10 +139,16 @@ func (api *NameAPI) Resolve(ctx context.Context, name string, opts ...caopts.Nam
 }
 
 func keylookup(self ci.PrivKey, kstore keystore.Keystore, k string) (ci.PrivKey, error) {
+	////////////////////
+	// Lookup by name //
+	////////////////////
+
+	// First, lookup self.
 	if k == "self" {
 		return self, nil
 	}
 
+	// Then, look in the keystore.
 	res, err := kstore.Get(k)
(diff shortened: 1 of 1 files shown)


core/coreapi/name.go
```diff
@@ -170,7 +194,7 @@ func keylookup(self ci.PrivKey, kstore keystore.Keystore, k string) (ci.PrivKey,
 			return nil, err
```
hsanjuan: Why is it not doing IDFromPrivateKey(privKey) directly?

core/coreapi/name.go
```diff
@@ -157,6 +163,24 @@ func keylookup(self ci.PrivKey, kstore keystore.Keystore, k string) (ci.PrivKey,
...
+	//////////////////
+	targetPid, err := peer.Decode(k)
+	if err != nil {
+		return nil, keystore.ErrNoSuchKey
+	}
+
+	// First, check self.
+	pid, err := peer.IDFromPrivateKey(self)
+	if err != nil {
+		return nil, fmt.Errorf("failed to determine peer ID for private key")
```
hsanjuan: Not including the original error message?

core/coreapi/name.go
hsanjuan: Just a couple of lines above, could not add comment directly there.

Stebalien: No idea. It may not have existed? I'll change that.

PR #7030, opened 2020-03-25
Forgotten go-fmt

ribasushi:
Not sure why CI didn't catch this


lidel:
Yeah, I don't believe CI checks this. The question is: is it worth adding?

IIRC in the past they changed the way gofmt formats things: if that happens again it could break the build when CI uses different go version than developer.

PR #7036, opened 2020-03-26
Dag export command, complete

ribasushi:
This is the most non-controversial part of all. The commented-out go-ipld-prime hickup is being debugged by @warpfork and me over at https://github.com/ribasushi/gip-muddle-up


core/commands/dag/dag.go
```diff
@@ -33,6 +39,7 @@ to deprecate and replace the existing 'ipfs object' command moving forward.
 		"put":     DagPutCmd,
 		"get":     DagGetCmd,
 		"resolve": DagResolveCmd,
+		"export":  DagExportCmd,
```
willscott: this file is quickly reaching the level of complexity where it may make sense to split the different sub-commands into their own files.

ribasushi: @Stebalien thoughts?

core/commands/dag/dag.go
```diff
@@ -241,3 +248,103 @@ var DagResolveCmd = &cmds.Command{
...
+		// anything else. We explicitly *DO NOT* take a lock during this
+		// operation: even if we lose some of the blocks we just received
+		// due to a conflicting GC: we will just re-retrieve anything we
+		// potentially lost when the car is being streamed out
+		node, err := cmdenv.GetNode(env)
+		if err != nil {
+			return err
+		}
+
+		if err := mdag.FetchGraph(req.Context, c, node.DAG); err != nil {
```
ribasushi: @Stebalien in one of our earlier conversations you asked me to do this pre-fetch before using the NewSession thing below ( it's commented out at present, but ignore that )

During the pre-fetch the CLI remains silent for minutes, pretty crappy UX. Could we simply not run this prefetch, and let the car writer pull stuff out, provided https://github.com/ipld/go-car/issues/27 gets implemented instead...?

core/commands/dag/dag.go
Stebalien: Yeah, that's a fair point. But I'd leave it this way then break it up in a followup patch to reduce the diff.

core/commands/dag/dag.go
Stebalien: I agree, and this will actually slow us down in the common case where we already have the data. The real solution is to teach the selector engine to prefetch (or at least hint that something needs to be loaded), but we can always do that later.

core/commands/dag/dag.go
```diff
@@ -241,3 +248,103 @@ var DagResolveCmd = &cmds.Command{
...
+	Helptext: cmds.HelpText{
+		Tagline: "Streams the selected DAG as a .car stream on stdout.",
+		ShortDescription: `
+'ipfs dag export' fetches a dag and streams it out as a well-formed .car file.
+Note that at present only single root selections / .car files are supported.
+The output of blocks happens in strict DAG-traversal, first-seen, order.
+`,
+	},
+	Arguments: []cmds.Argument{
+		cmds.StringArg("root", true, false, "Expression evaluting to a single root of a dag to export").EnableStdin(),
```
Stebalien: For now, let's just say "CID". Otherwise, the user isn't going to have any idea what to pass in. We can change the help text later when we add support for additional selectors.

core/commands/dag/dag.go
```diff
@@ -241,3 +248,103 @@ var DagResolveCmd = &cmds.Command{
...
+				node.DAG,
+				[]cid.Cid{c},
+				pipeW,
+			); err != nil {
+				errCh <- err
+			}
+		}()
+
+		if err := res.Emit(pipeR); err != nil {
+			pipeW.Close() // ignore errors if any
```
Stebalien: We shouldn't need to call this.

core/commands/dag/dag.go
```diff
@@ -241,3 +248,103 @@ var DagResolveCmd = &cmds.Command{
...
+		// 		Selector: sb.ExploreRecursive(
+		// 			gipselector.RecursionLimitNone(),
+		// 			sb.ExploreAll(sb.ExploreRecursiveEdge()),
+		// 		).Node(),
+		// 	}},
+		// )
+
+		pipeR, pipeW := io.Pipe()
+
+		errCh := make(chan error, 2) // we only report the 1st error
```
Stebalien: This is a really nice approach! I've always just used a mess of if statements.

core/commands/dag/dag.go
```diff
@@ -241,3 +248,103 @@ var DagResolveCmd = &cmds.Command{
...
+				if err := pipeW.Close(); err != nil {
+					errCh <- fmt.Errorf("stream flush failed: %s", err)
+				}
+				close(errCh)
+			}()
+
+			//if err := car.Write(pipeW); err != nil {
+			if err := gocar.WriteCar(
+				req.Context,
+				node.DAG,
```
Stebalien: Ideally, we'd call `merkledag.NewSession(ctx, node.DAG)`. But we'll have to change go-car to accept an `ipld-format.NodeGetter` instead of requiring a full `DAGService`.

ribasushi: This is not applicable - we will switch to the selector-based interface the moment go-ipld-prime is fixed ( the interface in `carobj.Write` is different 😿 )

core/commands/dag/dag.go
ribasushi: Fixed

core/commands/dag/dag.go
ribasushi: gone

core/commands/dag/dag.go
ribasushi: done

ribasushi:
Ready for final re-review ( every comment addressed, sharness fixed )

core/commands/dag/dag.go
Stebalien: But that's not ready yet and may not be ready for a while. This fix is easy: https://github.com/ipld/go-car/pull/28

core/commands/dag/dag.go
```diff
@@ -241,3 +247,100 @@ var DagResolveCmd = &cmds.Command{
...
+		errCh := make(chan error, 2) // we only report the 1st error
+		go func() {
+			defer func() {
+				if err := pipeW.Close(); err != nil {
+					errCh <- fmt.Errorf("stream flush failed: %s", err)
+				}
+				close(errCh)
+			}()
+
+			//if err := car.Write(pipeW); err != nil {
```
Stebalien: remove?

core/commands/dag/dag.go
```diff
@@ -241,3 +247,100 @@ var DagResolveCmd = &cmds.Command{
...
+		if err := res.Emit(pipeR); err != nil {
+			return err
+		}
+
+		err = <-errCh
+
+		// minimal user friendliness
+		if err != nil &&
+			!node.IsOnline &&
+			err.Error() == "merkledag: not found" {
```
Stebalien: `err == format.ErrNotFound`

@hsanjuan is also working on improving this error message.

core/commands/dag/dag.go
Stebalien: Ah, wait, actually... we need to call `pipeR.Close()` to unblock the writer, I think (it can't hurt to call it anyways).

Stebalien:
I missed a few things and I'd like to avoid making any assumptions about when the fixes to go-ipld-prime will ship.

ribasushi:
no worries, I am going down the list already, fixes to be pushed in ~15

core/commands/dag/dag.go
ribasushi: Fair enough, I am still not in a spot where I can tell when it is ok to change an interface other teams use...
Updated to use your branch!

core/commands/dag/dag.go
ribasushi: I modeled original code on the example at https://golang.org/pkg/io/#Pipe is showing a writer close only, and no reader closures. That's also consistent with "posix-ish thinking". Are you sure you want to move to `pipeR.Close()` instead?

ribasushi: I just did it anyway, even though it doesn't seem right... Closing a reader will just block a writer, though these are virtual writers so shouldn't matter.

Stebalien: Closing the reader will _unblock_ the writer: https://golang.org/pkg/io/#PipeReader.Close. I'm concerned about the following:

1. Copying from the pipe to the http writer fails because the http connection dies.
2. Emit returns an error.
3. We walk away but the CAR writer is still trying to write to the pipe.

Closing the writer will unblock the CAR writer and cause it to abort with an error.

test/sharness/t0054-dag-car-import-export.sh
```diff
@@ -0,0 +1,25 @@
...
+
+test_expect_success "correct error" '
+  test_cmp offline_fetch_error_expected offline_fetch_error_actual
+'
+
+test_expect_success "basic offline export of 'getting started' dag" '
+  ipfs dag export QmS4ustL54uo8FzR9455qaxZwuMiUhyvMcX9Ba8nUH4uVv >/dev/null
+'
+
+test_done
```
Stebalien: The code looks good. But this will need some tests.

ribasushi: There are are way more tests in the import part of the PR, because they mesh together ( import shuffled / export / check that the original converge )

Plus there are tests in the original that *didn't* get split correctly.

TLDR: let's merge this + the progress part **if the code is good**, then I will re-bae the importer and recheck the tests / fix conflicts etc.

core/commands/dag/dag.go
ribasushi: Oh... ok... that makes sense... BUT
Then the code can be made much safer like:
```
diff --git a/core/commands/dag/dag.go b/core/commands/dag/dag.go
index 6835d0945..ec4e88c94 100644
--- a/core/commands/dag/dag.go
+++ b/core/commands/dag/dag.go
@@ -295,6 +295,7 @@ The output of blocks happens in strict DAG-traversal, first-seen, order.
                // if err := car.Write(pipeW); err != nil {}
 
                pipeR, pipeW := io.Pipe()
+               defer pipeR.Close() // ignore the error if any, ublock the writer goroutine to exit
 
                errCh := make(chan error, 2) // we only report the 1st error
                go func() {
@@ -319,7 +320,6 @@ The output of blocks happens in strict DAG-traversal, first-seen, order.
                }()
 
                if err := res.Emit(pipeR); err != nil {
-                       pipeR.Close() // ignore the error if any
                        return err
                }
 
```

Not pushing it now not to muddle things up, but if you agree with add it to the import rebase

Stebalien: You're right. That's much better.

test/sharness/t0054-dag-car-import-export.sh
Stebalien: Well, I want to see passing tests somewhere first (the tests on the original PR look like they're currently failing). But we can base the import PR on this one and add the export tests there.

ribasushi: @Stebalien ok, making it so. There are also *more* tests in the progress bar PR, perhaps check this quick ( it's small ) and I'll work on rebasing import on all that + master

So you can look imto import afresh as a standalone diff

ribasushi: Also when you say tests: sharness tests are enough right? Or you have something more in mind?

Stebalien: Just sharness.

core/commands/dag/dag.go
ribasushi: Everything is terrible. **THIS** is the correct way to close the pipe: https://github.com/ipfs/go-ipfs/pull/7036/commits/70834450926f53902

This is the only extra commit I added since your LGTM, the rest is just merging master. Not planning to push more commits here unless the above closer needs fixing.

core/commands/dag/dag.go
```diff
@@ -307,6 +307,24 @@ The output of blocks happens in strict DAG-traversal, first-seen, order.
 		// if err := car.Write(pipeW); err != nil {}
 
 		pipeR, pipeW := io.Pipe()
+		// this is so convoluted because we:
+		// - want to close the readpipe on error so that the goroutine dies
+		// - can not do so on success, because during a no-deamon process
+		//   scenario the PostRun will not be able to grab the EOF from the
+		//   now-closed pipe
```
Stebalien: Interesting. That means that, when no daemon is run, we'll end up returning from `Run` before we're done writing the CAR.

This feels like a problem with the commands library.

core/commands/dag/dag.go
```diff
@@ -307,6 +307,24 @@ The output of blocks happens in strict DAG-traversal, first-seen, order.
...
+		// 		cmd/ipfs/ipfs dag export QmdmQXB2mzChmMeKY47C43LxUdg1NDJ5MWcKMKxDu7RgQm >/dev/null
+		//  0s  0 B / ? [------------------------------------------------------------------------=]Error: io: read/write on closed pipe
+		//  1s  107.08 MiB / ? [-----------------------------------------=----------] 106.59 MiB/s
+		//
+		var emittingError error
+		defer func() {
+			if emittingError != nil {
+				pipeR.Close()
+			}
+		}()
```
Stebalien: Given all this complexity, I'd just move this back to where it was.

ribasushi: There are 3 spots where we return errors. I would need to add a closer to all of them... Do that?

core/commands/dag/dag.go
ribasushi: Note - I ran abut a hundred tests with sha256sum on the receiving end: I never got a "truncated run", which makes me think this is strictly an EOF detection thing

core/commands/dag/dag.go
Stebalien: We only need it in https://github.com/ipfs/go-ipfs/pull/7036/files/63a56cf7bee571c5c1408a198e68be9f197bc6a7..70834450926f53902039ad3fbcc62c07b6f37758#diff-15172926c4147422f472139313467179R352. The next line waits on the error channel (which won't return till we're done writing so there's no need to unblock the writer).

core/commands/dag/dag.go
Stebalien: Hm. That's odd. Actually, I wouldn't even expect this error. We shouldn't get past the `<-errCh` line unless we finish writing and close the pipe, or we hit an error.

ribasushi: I just force-pushed removing this cap commit ( which puts everything the way it was )
I am fixing up import / merging / fixing sharness
Will dig into the error above a bit later

ribasushi:
@Stebalien this currently is exactly at where you said "LGTM" + the concurrency-free progressbar. If this passes sharness - it should be good to merge as-is.

I am working on the import part based off https://github.com/ipfs/go-ipfs/pull/7036/commits/ccbd08b16f96

ribasushi:
And of course it fails tests, seemingly unrelated:
```
=== FAIL: core/coreapi/test TestIface/Dht (1.37s)
2020-03-28T05:50:04.000Z	ERROR	core	core/builder.go:47	failure on stop: context canceled
github.com/ipfs/go-ipfs/core.NewNode.func2
	/home/circleci/ipfs/go-ipfs/core/builder.go:47
    --- FAIL: TestIface/Dht (1.37s)
```

Stebalien:
That's a known flaky test, unfortunately.

core/commands/root.go
```diff
@@ -190,6 +190,7 @@ var rootROSubcommands = map[string]*cmds.Command{
 	"dag": {
 		Subcommands: map[string]*cmds.Command{
 			"get":     dag.DagGetCmd,
+			"export":  dag.DagExportCmd,
```
ribasushi, line 193: @Stebalien now that I know what this list is for - do we want to keep this? I.e. should someone be able to request a .car from the gateway ( it seems **really** cool, but maybe load implications... )

Stebalien: Hm. Good point. Let's remove it for now and add a TODO to add it back later once we've done more load testing. It _would_ be cool, but this feature isn't going to get much hammering before the release.

ribasushi:
@Stebalien latest push on this branch merges master one last time, addresses https://github.com/ipfs/go-ipfs/pull/7036#discussion_r405716667, and removes urfave/cli.

Stebalien:
Rebased to remove merge commits.

PR #7037, opened 2020-03-26
Add simple byte-counting export progress

ribasushi:
This is the simple, non-controversial, export-progress-meter modeled on how `ipfs cat` works. R made agains the base mplementing branch for ease of review.

### DIFF
--- a/core/commands/dag/dag.go
+++ b/core/commands/dag/dag.go
@@ -1,10 +1,13 @@
 package dagcmd
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"math"
+	"os"
 	"strings"
+	"time"
 
 	"github.com/ipfs/go-ipfs/core/commands/cmdenv"
 	"github.com/ipfs/go-ipfs/core/coredag"
@@ -23,6 +26,12 @@ import (
 	//gipfree "github.com/ipld/go-ipld-prime/impl/free"
 	//gipselector "github.com/ipld/go-ipld-prime/traversal/selector"
 	//gipselectorbuilder "github.com/ipld/go-ipld-prime/traversal/selector/builder"
+
+	"gopkg.in/cheggaaa/pb.v1"
+)
+
+const (
+	progressOptionName = "progress"
 )
 
 var DagCmd = &cmds.Command{
@@ -261,6 +270,9 @@ The output of blocks happens in strict DAG-traversal, first-seen, order.
 	Arguments: []cmds.Argument{
 		cmds.StringArg("root", true, false, "CID of a root to recursively export").EnableStdin(),
 	},
+	Options: []cmds.Option{
+		cmds.BoolOption(progressOptionName, "p", "Display progress on CLI. Defaults to true when STDERR is a TTY."),
+	},
 	Run: func(req *cmds.Request, res cmds.ResponseEmitter, env cmds.Environment) error {
 
 		c, err := cid.Decode(req.Arguments[0])
@@ -333,5 +345,62 @@ The output of blocks happens in strict DAG-traversal, first-seen, order.
 		}
 
 		return err
+	},
+	PostRun: cmds.PostRunMap{
+		cmds.CLI: func(res cmds.Response, re cmds.ResponseEmitter) error {
+
+			var showProgress bool
+			val, specified := res.Request().Options[progressOptionName]
+			if !specified {
+				// default based on TTY availability
+				errStat, _ := os.Stderr.Stat()
+				if 0 != (errStat.Mode() & os.ModeCharDevice) {
+					showProgress = true
+				}
+			} else if val.(bool) {
+				showProgress = true
+			}
+
+			// simple passthrough, no progress
+			if !showProgress {
+				return cmds.Copy(re, res)
+			}
+
+			bar := pb.New64(0).SetUnits(pb.U_BYTES)
+			bar.Output = os.Stderr
+			bar.ShowSpeed = true
+			bar.ShowElapsedTime = true
+			bar.RefreshRate = 500 * time.Millisecond
+			bar.Start()
+
+			var processedOneResponse bool
+			for {
+				v, err := res.Next()
+				if err == io.EOF {
+
+					// We only write the final bar update on success
+					// On error it looks too weird
+					bar.Finish()
+
+					return re.Close()
+				} else if err != nil {
+					return re.CloseWithError(err)
+				} else if processedOneResponse {
+					return re.CloseWithError(errors.New("unexpected multipart response during emit, please file a bugreport"))
+				}
+
+				r, ok := v.(io.Reader)
+				if !ok {
+					// some sort of encoded response, this should not be happening
+					return errors.New("unexpected non-stream passed to PostRun: please file a bugreport")
+				}
+
+				processedOneResponse = true
+
+				if err := re.Emit(bar.NewProxyReader(r)); err != nil {
+					return err
+				}
+			}
+		},
 	},
 }


core/commands/dag/dag.go
```diff
@@ -347,4 +357,69 @@ The output of blocks happens in strict DAG-traversal, first-seen, order.
...
+					return re.CloseWithError(err)
+				}
+
+				if r, ok := v.(io.Reader); ok {
+					// we got a reader passed as a response
+					// proxy it through with an increasing counter
+					for {
+						len, readErr := r.Read(buf)
+						if len > 0 {
+							if err := re.Emit(bytes.NewBuffer(buf[:len])); err != nil {
```
Stebalien: Side note:

* In this case, you'd actually want https://golang.org/pkg/bytes/#NewReader. `bytes.NewBuffer` works, but is overkill.
* Alternatively, you could use (and keep reusing) a https://golang.org/pkg/io/#LimitedReader, resetting `N` each time.

However, can we just use `progressBarForReader` from the `get`/`cat` commands? IIRC, we're allowed to set the size to 0.

core/commands/dag/dag.go
```diff
@@ -347,4 +357,69 @@ The output of blocks happens in strict DAG-traversal, first-seen, order.
...
+				}
+			}()
+
+			for {
+				v, err := res.Next()
+				if err != nil {
+					return re.CloseWithError(err)
+				}
+
+				if r, ok := v.(io.Reader); ok {
```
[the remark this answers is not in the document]
ribasushi: :)

core/commands/dag/dag.go
```diff
@@ -347,4 +357,69 @@ The output of blocks happens in strict DAG-traversal, first-seen, order.
...
+
+			// simple passthrough
+			if !showProgress {
+				return cmds.Copy(re, res)
+			}
+
+			// We will be supplied a reader, simply print out how many bytes were passed as a crude progress
+
+			var exportSize int64
+			buf := make([]byte, 4*1024*1024) // this does not mean "read 4MB at a time", rather make a buffer larger than any reasonable pipe size
```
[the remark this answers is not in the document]
ribasushi: That's... troublesome. Read() is the basic libc call of "read anything you can from 1B up to buffer size". If the stack can't handle this... we are in trouble :)

In any case moot as I'll switch this to the progressbar in a moment ( though it's own UX is horrible, but that's in another PR way way down the road )

core/commands/dag/dag.go
ribasushi: > can we just use progressBarForReader 

Making it so

ribasushi: @Stebalien, @willscott  ok, I force pushed the PR. It is indeed much neater this way. I did not copy any of the "clear STDERR" crazy from the `commands` package, I am not sure why we are jumping through so many hoops over there

In action:
( while retrieval runs )
```
admin:~/devel/go-ipfs$ cmd/ipfs/ipfs dag export QmdmQXB2mzChmMeKY47C43LxUdg1NDJ5MWcKMKxDu7RgQm | sha256sum
 13s  12.89 MiB / ? [-------=-----------------------------------------------------------------------] 972.67 KiB/s 
```

when done:
```
cmd/ipfs/ipfs dag export QmdmQXB2mzChmMeKY47C43LxUdg1NDJ5MWcKMKxDu7RgQm | sha256sum
 1s  107.08 MiB / ? [-------------------------------------------=---------------------------------] 99.08 MiB/s 1s 
5a49df60c4842e7fe71849410025a8a5a98e2d0e6e71f90b4913635f7df7250d  -

```

core/commands/dag/dag.go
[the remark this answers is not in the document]
ribasushi: Taking it back - the ux of the base package is excellent. We screw it up ourselves with crap like https://github.com/ipfs/go-ipfs/blob/master/core/commands/get.go#L136-L143

core/commands/dag/dag.go
Stebalien: >  I did not copy any of the "clear STDERR" crazy from the commands package, I am not sure why we are jumping through so many hoops over there

Run an `add` command and look at the output. Without the clearing, we'd see something like:

```
added QmFoobar----------------------------] 99.08MiB/s 1s
added QmOther----------------------------] 99.08MiB/s 1s
...
```

ribasushi: Ah... that's only on recursion though. Each piece that got submitted on the CLI *should* get its own bar / sizing that doesn't disappear on me after.

Adding it as a note when I get a chance to come back and clean up some of that, post-dagger.

core/commands/dag/dag.go
```diff
@@ -343,4 +355,53 @@ The output of blocks happens in strict DAG-traversal, first-seen, order.
...
+				}
+
+				r, ok := v.(io.Reader)
+				if !ok {
+					// some sort of encoded response, this should not be happening
+					return errors.New("unexpected non-stream passed to PostRun: can't happen")
+				}
+
+				if err := func() error {
+					bar := pb.New64(0).SetUnits(pb.U_BYTES)
```
Stebalien: How does this look when adding multiple files? It would be nice to see:

```
Exported Qm...
Exported Qm...
Exported Qm...
[ progress ]
```

Instead of having dead progress bars in between. But it's not a _huge_ issue.

ribasushi: > How does this look when adding multiple files? 

We can only export one root/DAG ( single CID spec, no selectors ). The exporter only sees "bytes written" into a single stream on STDOUT. No files are created.

Stebalien: Oh, yeah. I knew that...

ribasushi: I think I see what you mean... let me rewire this / rebase on top of the approved export.

core/commands/dag/dag.go
Stebalien: > Ah... that's only on recursion though. Each piece that got submitted on the CLI should get its own bar / sizing that doesn't disappear on me after.

Note: `ipfs add` uses a single bar for _everything_, calculating the total size up-front.

core/commands/dag/dag.go
ribasushi: No you are right - there may be a reason for the server to emit multiple read-ables for some reason. So having the bar outside of the loop make sense. I will redo it.

Stebalien: That shouldn't happen... and I'm not sure what we'd be able to do about it. Would we just concatenate them?

ribasushi: Yep, concat.

Alternatively we can throw saying "unexpected multipart export..." or somesuch.

*OR* we can leave the multi-bar spawning as-is ( no changes to PR ), then there will eb a visual cue "something went wrong"

But if you say "this shouldn't happen" - erroring is what I prefer.

Stebalien: I'd error.

ribasushi: Addressed as https://github.com/ipfs/go-ipfs/pull/7037/files#diff-15172926c4147422f472139313467179R388-R396

There is more code on the closer, as that became a can of worms, but it will go in the other PR once this is folded in ( I promise it will make sense in a bit )

Waiting for CI to run here, and will merge this into the other branch.

core/commands/dag/dag.go
ribasushi: I see... :sadpanda:

Stebalien:
@ribasushi please avoid merging with non-trivial unreviewed changes. The plan is to merge `feat/carfile-export-only` all at once with _no_ in-depth review because all the sub-patches should have already been reviewed.

ribasushi:
@Stebalien you did sign off on both the target and the destination...

Stebalien:
That signoff is from 20:29 PDT. The discussion ended at 21:19 PDT (https://github.com/ipfs/go-ipfs/pull/7037#discussion_r399617116). I never saw that change.

ribasushi:
I went off my mail client, thought you saw the progress bar and then hit approve. My bad.

Is there something wrong with https://github.com/ipfs/go-ipfs/pull/7037/files#diff-15172926c4147422f472139313467179R388-R396 ? I'll fix whatever's needed.

Stebalien:
Nope, the change looks good. I'm just trying to prevent bugs from slipping through.

PR #7038, opened 2020-03-26
Dag import functionality only ( silent / no CLI progress )

ribasushi:
This still works over "loosely defined" .car files
Please refer to the sharness tests for extra info

We can tighten this up if the sentiment is "Postel was wrong"


core/commands/dag/dag.go
```diff
@@ -241,3 +261,248 @@ var DagResolveCmd = &cmds.Command{
...
+		if err != nil {
+			return err
+		}
+
+		// on import ensure we do not reach out to the network for any reason
+		// if a pin based on what is imported + what is in the lockstore
+		// isn't possible: tough luck
+		api, _ = api.WithOptions(options.Api.Offline(true))
+
+		// grab a gc lock so that regardless of the size of the streamed-in cars
```
willscott: comment doesn't reflect the pinlock below

core/commands/dag/dag.go
```diff
@@ -241,3 +261,248 @@ var DagResolveCmd = &cmds.Command{
...
+		// This is especially important for use cases like dagger:
+		//    ipfs dag import $( ... | ipfs-dagger --stdout=carfifos )
+		//
+		unlocker := node.Blockstore.PinLock()
+		defer func() { unlocker.Unlock() }()
+
+		doPinRoots, _ := req.Options[pinRootsOptionName].(bool)
+
+		retCh := make(chan importResult, 1)
+		go importWorker(req, res, api, retCh)
```
willscott: you don't get anything from the go routine complexity in this construction. is there another use or reason for this?

core/commands/dag/dag.go
```diff
@@ -241,3 +261,248 @@ var DagResolveCmd = &cmds.Command{
...
+			return done.err
+		}
+
+		// It is not guaranteed that a root in a header is actually present in the same ( or any )
+		// .car file. This is the case in version 1, and ideally in further versions too
+		// Accumulate any root CID seen in a header, and supplement its actual node if/when encountered
+		// We will attempt a pin *only* at the end in case all car files were well formed
+		//
+		// The boolean value indicates whether we have encountered the root within the car file's
+		//var roots map[cid.Cid]bool
```
willscott: is this used as a hint, or intentional?
if its a hint to help the reader understand the type, it might be better to have it in code so it stays in sync
```
var roots map[cid.Cid]bool
roots = done.roots
```

core/commands/dag/dag.go
```diff
@@ -241,3 +261,248 @@ var DagResolveCmd = &cmds.Command{
...
+				"Pinned root\t%s\t%s\n",
+				enc.Encode(event.Root.Cid),
+				event.Root.PinErrorMsg,
+			)
+			return err
+		}),
+	},
+}
+
+func importWorker(req *cmds.Request, re cmds.ResponseEmitter, api iface.CoreAPI, ret chan importResult) {
```
willscott: consider refactoring to separate file

core/commands/dag/dag.go
```diff
@@ -241,3 +261,248 @@ var DagResolveCmd = &cmds.Command{
...
+			// encountered something known to be a root, for the first time
+			if seen, exists := roots[nd.Cid()]; exists && !seen {
+				roots[nd.Cid()] = true
+			}
+		}
+
+		// every single file in it() is already open before we start
+		// just close here sooner rather than later for neatness
+		// this won't/can't help with not running out of handles
+		file.Close()
```
willscott: should these be closed in error cases as well? is there a deferred close scheduled as well?

core/commands/dag/dag.go
ribasushi: Another PR will include progress feedback utilizing this.

core/commands/dag/dag.go
ribasushi: @Stebalien thoughts?

core/commands/dag/dag.go
ribasushi: Fixed

core/commands/dag/dag.go
ribasushi: Fixed

core/commands/dag/dag.go
```diff
@@ -241,3 +261,249 @@ var DagResolveCmd = &cmds.Command{
...
+		ShortDescription: `
+'ipfs dag import' parses .car files and adds all objects to the blockstore.
+Currently only car files with version 1 are supported, without particular
+hard limits on the contents (so called car-determinism is not enforced).
+
+By default, after all supplied (one or more) car files have been processed,
+an attempt is made to pin recursively each individual root specified in any car
+headers. This is done at the end of the import, allowing pinning to take place
+over the contents of all cars (a dag scattered over multiple files) plus
+anything present in the blockstore.
```
Stebalien: I'd rewrite this to focus on _what_ happens, not how, and to remove terms like "car-determinism" (not something anyone outside of the core IPLD team will understand).

Something along the lines of:

> `ipfs dag import` imports all blocks present in the CAR (Content Addressed aRchive) file, pinning the roots (as specified in the CAR file headers) unless `--pin-roots` is set to false.
>
> Note:
> 1. This command will add all blocks in the CAR file, not just those reachable from the roots. However, these other blocks will not be pinned and may be garbage collected later.
> 2. other caveats...
> 
> Maximum supported CAR version: 1

core/commands/dag/dag.go
```diff
@@ -241,3 +261,249 @@ var DagResolveCmd = &cmds.Command{
...
+			return err
+		}
+
+		api, err := cmdenv.GetApi(env, req)
+		if err != nil {
+			return err
+		}
+
+		// on import ensure we do not reach out to the network for any reason
+		// if a pin based on what is imported + what is in the lockstore
```
Stebalien: s/lockstore/blockstore

core/commands/dag/dag.go
```diff
@@ -241,3 +261,249 @@ var DagResolveCmd = &cmds.Command{
...
+		// isn't possible: tough luck
+		api, _ = api.WithOptions(options.Api.Offline(true))
```
Stebalien: Error checking?

core/commands/dag/dag.go
Stebalien: It may not matter in practice, but we probably _should_ anyways.

core/commands/dag/dag.go
```diff
@@ -241,3 +261,249 @@ var DagResolveCmd = &cmds.Command{
...
+				// rp := path.NewResolvedPath(
+				// 	ipfspath.FromCid(c),
+				// 	c,
+				// 	c,
+				// 	"",
+				// )
+				//
+				// if err := api.Pin().Add(req.Context, rp, options.Pin.Recursive(true)); err != nil {
+
+				ret := RootMeta{Cid: c, PresentInCar: seen}
```
Stebalien: Is "seen" useful? It doesn't guarantee that the entire DAG was in the CAR.

core/commands/dag/dag.go
Stebalien: Simply emitting a response every time we add a block is probably going to be simpler than having a separate goroutine.

ribasushi: Wouldn't it be **way** too expensive to JSON encode a counter message on every block for multi-thousand block cars. This is why the original implementation used a timer-ticker based emitter. If you feel strongly about it: I will switch to emitting on every block, but ewwww.

core/commands/dag/dag.go
ribasushi: Not useful for functionality at all. But useful as feedback on non-silent:
https://github.com/ipfs/go-ipfs/pull/7038/files#diff-aaf3aece3f30477f8db9258627b55791R134-R137

At least I find it very useful on my end. I could remove all of this if it is deemed confusing.

core/commands/dag/dag.go
Stebalien: Knowing all the CIDs in a CAR would be generally useful. But I see your point. We'd only want to enable that with a flag in special cases.

core/commands/dag/dag.go
ribasushi: Rewired. It's not pretty in diff-form ( hideous actually ), but the actual resulting code is ok.

core/commands/dag/dag.go
ribasushi: Fixed

core/commands/dag/dag.go
ribasushi: Done
Result:
```
SYNOPSIS
  ipfs dag import [--silent] [--pin-roots=false] [--] <path>...

ARGUMENTS

  <path>... - The path of a .car file.

OPTIONS

  --silent     bool - No output.
  --pin-roots  bool - Pin optional roots listed in the .car headers after importing. Default: true.

DESCRIPTION

  'ipfs dag import' imports all blocks present in supplied .car
  ( Content Address aRchive ) files, recursively pinning any roots
  specified in the CAR file headers, unless --pin-roots is set to false.
  
  Note:
  	This command will import all blocks in the CAR file, not just those
  	reachable from the specified roots. However, these other blocks will
  	not be pinned and may be garbage collected later.
  
  	The pinning of the roots happens after all car files are processed,
  	permitting import of DAGs spanning multiple files.
  
  	Pinning takes place in offline-mode exclusively, with no option to change
  	this: if the collection of blocks from the imported CAR files and what
  	is currently present in the blockstore is insufficient pinning of the
  	affected root will fail
  
  Maximum supported CAR version: 1
```

ribasushi:
This PR is now retargetted against the export branch to make diff reading easier. Sharness tests are complete, with a few more "pathological cases" pending for a subsequent PR

ribasushi:
:scratchhead:
```
The following errors were encountered parsing test results:

sharness/sharness.xml
    ParseError at [row,col]:[25710,1] Message: Invalid byte 1 of 1-byte UTF-8 sequence.
```

Locally the full sharness passes...

ribasushi:
One needs to run `make -O -j 10 coverage/sharness_tests.coverprofile test/sharness/test-results/sharness.xml TEST_GENERATE_JUNIT=1 CONTINUE_ON_S_FAILURE=1` locally ( lifted from `.circleci` ) in order to observe the malformed XML. Regular `make test` hides all the the STDOUTs.

ribasushi:
Sharness is now properly fixed, I ended up running into another known mis-design along the way: 
https://github.com/ipfs/go-ipfs/pull/7038/commits/3e40ce1b8e3272f1aea3d0341057b412133b5f5a#diff-15172926c4147422f472139313467179R509-R517

@Stebalien ready for final review, I will re-add the progress stuff once you are happy with the tests / code / etc.

I would also like to hear from @mikeal if he has strong(er) objections to the current general design of the importer: https://github.com/ipfs/go-ipfs/pull/7038#discussion_r399629168

core/commands/dag/dag.go
warpfork: A ticking progress reporter goroutine also has the nice attribute of being in a position to notice (and report) if some operations stall.

core/commands/dag/dag.go
```diff
@@ -258,6 +273,277 @@ var DagResolveCmd = &cmds.Command{
...
+				}
+
+				if err := batch.Add(req.Context, nd); err != nil {
+					return err
+				}
+				blockCount++
+				if 0 == blockCount%200 {
+					// work around https://github.com/ipfs/go-ds-flatfs/issues/36 for the time being
+					// batch up-to 200 at a time ( fly under MacOS's default of 256 )
+					if err := batch.Commit(); err != nil {
```
Stebalien: Did you _actually_ run into this problem? `ipfs add` has the same behavior.

Looking at the code, this is probably because we usually test with the daemon (which raises the file descriptor limit to 8196). Let's not hack a fix in here.

Stebalien: Also note: we _can_ reduce the batch size when constructing the batch.

core/commands/dag/dag.go
Stebalien: But it's not even correct. We'll return "true" if it's present in one of the CARs.

core/commands/dag/dag.go
ribasushi: Yes I have ran into it. Part of the sharness tests runs **without** a daemon ( as would be seen in the wild in many cases ). If you remove this block - that test will fail.

I didn't realize the batch size can be controlled, will push a fix momentarily.

core/commands/dag/dag.go
ribasushi: Yes - "present in one of the supplied cars" is exactly what I was after. Would naming it `PresentInImport` fix the ambiguity?

core/commands/dag/dag.go
ribasushi: Welp the batch **size** can be controlled, and is currently [defaulting to `128`](https://github.com/ipfs/go-ipld-format/blob/v0.1.0/batch.go#L203-L206). But the batch **application** is carried out async with [`ParallelBatchCommits`](https://godoc.org/github.com/ipfs/go-ipld-format#pkg-variables) concurrent jobs. Combined together they blow through the `ulimit` since blocks come in at cid-decoding-speed.

Since this is a global variable, the only reasonable way to flip it would be to muck with `atomic.` and that doesn't sound right.

How should we fix this? Add an option-defaulting-to-0 to ignore the global value of `ParallelBatchCommits`? Not using a batch at all?

Stebalien: My real concern here is that these are the default limits we're using for `ipfs add` as well. Given that, we should consider reducing the default batch size (or just fixing flatfs).

ribasushi: I looked briefly at `ipfs add` and did not see other locks/governors. I suppose we are not hitting a problem with `ipfs add` sufficiently often because. 
1) it usually with a daemon running ( higher ulimit )
2) the FS contention gets in the way allowing us to fly just under the limit, by luck

Stebalien: I know why...

1. We flush when we finish adding each individual file.
2. The default chunk size is 256KiB.
3. The maximum amount of data we can buffer is 8MiB * NCPU * 2.

That means we can end up with `(8196*NCPU*2)/256` outstanding blocks (usually ~256).

Regardless, we should fix the batch: https://github.com/ipfs/go-ipld-format/pull/56.

Then we should benchmark and see how this affects add performance on flatfs.

ribasushi: Do we have prior art for benchmarking? A typical workstation OS with an SSD will not display a difference for obvious reasons.

Stebalien: Unfortunately, no. However, I seem to remember a pretty dramatic difference when messing with these parameters last time, even on an SSD.

ribasushi: That's ok, I'll improvise. Finishing up something else at the moment, will likely not look into this until way later today.

core/commands/dag/dag.go
ribasushi: This is now addressed as https://github.com/ipfs/go-ipfs/pull/7038/commits/c1b707b17d11968967145f52ada7c45f1bc607ed#diff-aaf3aece3f30477f8db9258627b55791

core/commands/dag/dag.go
ribasushi: Ok, sharness passes now with no ulimit fixes and no workarounds
Had to pull in https://github.com/ipfs/go-merkledag/pull/53 and https://github.com/ipfs/go-ipld-format/pull/55 in order to benefit from https://github.com/ipfs/go-ipld-format/pull/56 

Initial testing looks good, will have the final number tomorrow morning, day got too long

Stebalien:
I've rebased feat/carfile-export-only on master. I was planning on then rebasing this branch on feat/carfile-export-only so I could review the changes, but they appear to have divergent histories.

Stebalien:
I've released a new go-ipld-format that reverts the ErrNotFound changes: https://github.com/ipfs/go-ipld-format/releases/tag/v0.2.0.

ribasushi:
>  they appear to have divergent histories

Export was merged into it, hence git getting confused. I generally avoid force-pushes during review exactly for this reason, but no biggie.

Something is wrong with sharness locally is why I haven't pushed yet, any moment now...

test/sharness/t0054-dag-car-import-export-data/README.md
```diff
@@ -0,0 +1,24 @@
+# Dataset description/sources
```
ribasushi, line 1: @Stebalien forgot to ask you:
```
du -sh test/sharness/t0054-dag-car-import-export-data/
2.9M	test/sharness/t0054-dag-car-import-export-data/
```

Is this acceptable to be included in the repo as-is? If not - I can rework the test to use xz ( and skip if not found )

Stebalien: I'd compress them. It's not terrible, but 2.9MiB is a bit much.

ribasushi: Fix incoming with the comments ( no code changes )

test/sharness/t0054-dag-car-import-export.sh
```diff
@@ -4,22 +4,169 @@
...
+
+  touch spin.gc
+  test_expect_success "fifo import" '
+    do_import 0 \
+      pipe_testnet \
+      pipe_devnet \
+      ../t0054-dag-car-import-export-data/combined_naked_roots_genesis_and_128.car \
+    | sort > basic_fifo_import_actual
+  '
+  # FIXME - the fact we reliably fail this is indicative of some sort of race...
```
Stebalien, line 90: Let's make sure to update this comment. This isn't due to a race, it's because we're creating blocks and aren't sure where they're coming from.

Actually, looking at this, I think I know where they're coming from. We're flushing multiple times, one for each root. If we have multiple roots, that would do it.

core/commands/dag/dag.go
```diff
@@ -258,6 +273,267 @@ var DagResolveCmd = &cmds.Command{
...
+
+				ret := RootMeta{Cid: c, PresentInImport: seen}
+
+				if block, err := node.Blockstore.Get(c); err != nil {
+					ret.PinErrorMsg = err.Error()
+				} else if nd, err := ipld.Decode(block); err != nil {
+					ret.PinErrorMsg = err.Error()
+				} else if err := node.Pinning.Pin(req.Context, nd, true); err != nil {
+					ret.PinErrorMsg = err.Error()
+				} else if err := node.Pinning.Flush(req.Context); err != nil {
```
Stebalien, line 390: I'd prefer to flush at most once at the end, but let's leave it this way for now. We'll have to think about the safety implications.

core/commands/dag/dag.go
```diff
@@ -258,6 +273,267 @@ var DagResolveCmd = &cmds.Command{
...
+				// rp := path.NewResolvedPath(
+				// 	ipfspath.FromCid(c),
+				// 	c,
+				// 	c,
+				// 	"",
+				// )
+				//
+				// if err := api.Pin().Add(req.Context, rp, options.Pin.Recursive(true)); err != nil {
+
+				ret := RootMeta{Cid: c, PresentInImport: seen}
```
Stebalien, line 382: Continuing conversation from https://github.com/ipfs/go-ipfs/pull/7038#discussion_r399571054 after rebase.

Stebalien: > Yes - "present in one of the supplied cars" is exactly what I was after. Would naming it PresentInImport fix the ambiguity?

No, it's still a lie. We're pinning the _DAG_, not just the roots. To me, `PresentInImport` indicates that the _DAG_ is present, not that the _root_ is present. The error message reinforces this by saying that the "data" was/wasn't present.

As a user, if I ever saw that error message and/or this type definition, I'd assume that the CAR contained all the data I tried to import. Given that we're not actually validating the CAR, this is not good.

At best, this feature is a heuristic to indicate that the CAR might have contained the DAG.

If we're going to have a feature like this, we need to implement it correctly and validate that the import contained _all_ blocks referenced by the root. However, that could force us to keep quite a few CIDs in memory so I'd definitely punt that to a future patch.

Stebalien: TL;DR: I don't want to give users a false sense of security by including a sanity check that only sort of works. I realize this was helpful in debugging this command, but it only catches _one_ way in which a CAR file can be broken.

Stebalien:
LGTM except for the "root seen" flag. It's much easier to remove it now and add it (or something like it) back later. IMO, we'd be better off with `ipfs dag import [--dry-run] [--verify=[properties...]] ...` (later).

core/commands/dag/dag.go
ribasushi: Actually you told me I need to pin+flush like that, originally I had it at the end... :)
But yeah, agreed we should punt, knowing more dark corners of the pin infra now. I'll open an issue in a bit referencing this.

ribasushi:
Force-pushed to lose the 3mb of testdata from the commit chain. Addressing the other notes in separate commits shortly.

core/commands/dag/dag.go
Stebalien: I did? I wonder why I said that...

test/sharness/t0054-dag-car-import-export.sh
ribasushi: Done with reference of issue. I will cross-post the test to be fixed in the issue once we merge and the githash is stable.

core/commands/dag/dag.go
ribasushi: Punted to https://github.com/ipfs/go-ipfs/issues/7120

ribasushi:
@Stebalien the only new thing to review is https://github.com/ipfs/go-ipfs/pull/7038/commits/eff4223dbdc1fcd6f29bd9da60ceb096735197d2

Everything else pushed is sharness cruft

PR #7043, opened 2020-03-27
Keystore: minor maintenance

hsanjuan:
```
keystore: Switch from Bytes() to MarshalPrivateKey()

Bytes is deprecated.
```

```
keystore: create new keys with 0400 permissions (as spec'ed) 

Spec is pretty much out of date but specifies this.

https://github.com/ipfs/specs/blob/master/KEYSTORE.md
```


hsanjuan:
Am I too lazy for sending two unrelated changes in the same PR or is this ok?

PR #7045, opened 2020-03-27
chore: update golang

Stebalien:


willscott:
note: circleci has a new image we should eventually be transitioning to: "cimg/go" - the main difference is that it defaults to a user acct rather than a root acct, so installation of dependencies may not transition seamlessly

PR #7055, opened 2020-03-30
feat: tls by default

Stebalien:
Switches to TLS as the default security transports.


Stebalien:
> only question would be if it's worth having a config option for preferring secio, but since fallback to secio should still work, that seems like something that can be deferred until someone expresses a need for it.

If it's an issue in practice and TLS is a _problem_, we'll cut a patch release.

Stebalien:
I'd _like_ to just keep the config option but we have no way of changing defaults (see https://github.com/ipfs/go-ipfs-config/issues/53).

Stebalien:
(updated docs)

PR #7058, opened 2020-03-30
fix: get rid of shutdown errors

Stebalien:
Instead of feeding through the top-level context, feed through a cancel-free context (that still carries the same context values). Then, when the top-level context is canceled, call `stop` to shut everything down in-order. Finally, cancel the inner context to make sure everything has been cleaned up.

Ideally, we just wouldn't use contexts for this. But this is strictly better than what we have.

### DIFF
--- a/core/builder.go
+++ b/core/builder.go
@@ -3,6 +3,7 @@ package core
 import (
 	"context"
 	"sync"
+	"time"
 
 	"github.com/ipfs/go-ipfs/core/bootstrap"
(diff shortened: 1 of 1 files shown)


Stebalien:
Merging without review because I'm trying to cut an early RC before Monday and printing a scary ERROR message on shutdown is not something I want to deal with in the RC.

Stebalien:
@willscott please review post-merge when you get a chance.

PR #7064, opened 2020-03-31
fix: remove internal relay discovery

Stebalien:
This logic collects a list of known relays by testing every new connection.

It exists so we can dial /p2p-circuit/p2p/QmFoobar addresses (circuit addresses that don't specify the relay). However, this kind of address is useless outside of basic demos as a random relay is practically guaranteed to not be connected to the target peer. Picking a random relay to connect to some peer is almost _never_ the desired behavior.

### DIFF
--- a/core/node/libp2p/relay.go
+++ b/core/node/libp2p/relay.go
@@ -11,7 +11,7 @@ func Relay(disable, enableHop bool) func() (opts Libp2pOpts, err error) {
 			// Enabled by default.
 			opts.Opts = append(opts.Opts, libp2p.DisableRelay())
 		} else {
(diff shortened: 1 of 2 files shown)


willscott:
```
expecting success: 
  ipfsi 0 swarm connect /p2p-circuit/p2p/$PEERID_2 > peers_out

Error: connect QmbYxz2oBPzdY4pnBs26WuC4BEa6nnwns3fWaJwWM9mrQK failure: failed to dial : all dials failed
  * [/p2p-circuit] Failed to dial through 0 known relay hosts
```
looks like there's a sharness test to update with this

Stebalien:
Yep.

PR #7068, opened 2020-04-01
feat: basic ipfs network diagnostics

Stebalien:
This implements a basic command and subsystem for diagnosing various networking issues.

### DIFF
--- a/core/commands/netdiag.go
+++ b/core/commands/netdiag.go
@@ -1,0 +1,133 @@
+package commands
+
+import (
+	"fmt"
+	"io"
+	"text/template"
+
+	"github.com/ipfs/go-ipfs/core/commands/cmdenv"
+	"github.com/ipfs/go-ipfs/doctor"
+
+	cmds "github.com/ipfs/go-ipfs-cmds"
+
+	libp2pNetwork "github.com/libp2p/go-libp2p-core/network"
+)
+
+type NetworkDiagnostics struct {
+	doctor.Status
+	IsOnline bool
+}
+
+const netDiagPublicText = `
+You are online and reachable from the public network.
+`
+
+const netDiagUnknownText = `
+IPFS has not yet determined if your node is reachable from the public network.
+`
+
+const netDiagPrivateText = `
+{{ if .UsingRelays -}}
+Your IPFS node appears to be behind a Firewall or NAT and can only be reached
+through a relay.
+{{- else -}}
+{{if .AutoRelayEnabled -}}
+Your IPFS node appears to be behind a Firewall or NAT and will try to
+automatically find a relay. However, it is not currently reachable.
+{{- else -}}
+Your IPFS node appears to be behind a Firewall or NAT and cannot be reached by
+nodes in the public network.
+
+To join the network via a relay, please enable the "auto-relay" feature:
+
+    ipfs config --bool Swarm.EnableAutoRelay true 
+
+And then restart IPFS.
+{{- end -}}
+{{end}}
+
+Please note, relays are public and highly constrained resources.
+
+For the best performance, please enable Universal Plug n Play (UPnP) on your
+router so IPFS can automatically configure it to open an inbound port.
+
+{{ if or .TCPPorts .UDPPorts -}}
+Alternatively, you can manually configure your router. Please forward the following ports:
+
+{{ if .TCPPorts -}}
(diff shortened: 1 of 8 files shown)


Stebalien:
(needs tests)

core/commands/netdiag.go
```diff
@@ -0,0 +1,133 @@
...
+{{ if .TCPPorts -}}
+TCP:{{ range $element := .TCPPorts}} {{$element}}{{end}}
+{{- end}}{{if .UDPPorts -}}
+UDP:{{ range $element := .UDPPorts}} {{$element}}{{end}}
+{{- end}}
+
+To (your current IP): {{.LocalIP}}
+{{- end}}
+{{if .Gateway -}}
+Your router's administrative console can likely be found at:
```
willscott, line 66: this is less likely to be actionable in corporate / educational situations. Maybe clearer specification of "if you are on a personal home network" or similar?

doctor/doctor.go
```diff
@@ -0,0 +1,240 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
```
willscott, line 182: a redirect would be okay, but shouldn't be followed?

Stebalien: Go follows redirects. This should (I think?) give me the _final_ status. Then I call `Location()` to get the final location header (if present).

autonome:
@jacobheun Is there an epic/meta issue for debugging/tools/troubleshooting?

jacobheun:
> Is there an epic/meta issue for debugging/tools/troubleshooting?

@autonome not at the moment, but this would be helpful. I think there is some overlap here with metrics/telemetry but they're discrete concerns. I'll open up a meta issue for tracking/discussion.

aschmahmann:
Just a note, that while I've updated this PR and it looks like it's passing tests we don't really have any tests to make sure this works (testing this is hard bc it's network level/configuration things). I've done a bit of local testing and found the results a little flaky (it wouldn't always realize if I was unreachable).

We'll need to put a bit of time into review + testing before we ship this. Would love to get this in though, it should be super useful.

aschmahmann:
cc @aarshkshah1992 you may want to modify this PR or use it as inspiration given your work on https://github.com/libp2p/go-libp2p/pull/1042

doctor/doctor.go
```diff
@@ -0,0 +1,240 @@
...
+	switch n.reachability {
+	case libp2pNetwork.ReachabilityPrivate:
+		status = "private"
+	case libp2pNetwork.ReachabilityPublic:
+		status = "public"
+	default:
+	}
+
+	n.printf("Firewall Status: %s", status)
+	if !n.hasNagged {
```
aarshkshah1992, line 156: @stebalien Would you remember why you put in this flag ? What's the use case ?

doctor/doctor.go
```diff
@@ -0,0 +1,240 @@
...
+		n.hasNagged = true
+		n.printf("NOTE: Your node appears to be behind a firewall. Please run `ipfs diag net` to diagnose.")
```
aarshkshah1992, line 158: How can we be sure here that we are behind a firewall ?

doctor/doctor.go
willscott: it appears to be so that it only says it once, and then remembers that it has said it so it doesn't say it again, right?

aarshkshah1992:
@willscott @aschmahmann I've made some small changes and have updated the deps and rebased. This PR looks in pretty good shape. Please can. I get a review ?

Will augment it with the NAT Type event on a separate PR as that work will time time to get merged.

willscott:
@aarshkshah1992 - looking at this conversation, the main thing this PR was waiting for was some sort of testing to gain confidence that we tell people the right things given their circumstances, and that future changes don't break this. Thoughts on any clear paths to gain confidence in this code?

willscott:
concretely: lets put in a sanity test: run a daemon. run this command. make sure daemon and command exit cleanly.

aschmahmann:
This is needs a rebase and may be dependent on some updates in go-libp2p to support detecting the ability to do NAT traversal.

PR #7097, opened 2020-04-04
HTTP API: Disallow GET requests

hsanjuan:
Better security by limiting the HTTP API to POST requests. (see commit descriptions).


hsanjuan:
This might break tests, not sure, I will fix those on another day.

core/corehttp/commands.go
```diff
@@ -140,15 +141,15 @@ func commandsOption(cctx oldcmds.Context, command *cmds.Command) ServeOption {
 }
 
 // CommandsOption constructs a ServerOption for hooking the commands into the
-// HTTP server.
+// HTTP server. These can be called only with POST, "OPTIONS" and "HEAD".
 func CommandsOption(cctx oldcmds.Context) ServeOption {
-	return commandsOption(cctx, corecommands.Root)
+	return commandsOption(cctx, corecommands.Root, []string{"POST", "OPTIONS", "HEAD"})
```
ribasushi, line 146: What is the rationale for allowing `OPTIONS` and `HEAD` on write opts?

Also do we not recognize `PUT`/`DELETE` anywhere?

hsanjuan: Ok, I was tired. I have removed HEAD (as this is akin to GET).

OPTIONS is necessary for CORS reasons, as it is how browsers figure out what's allowed and what not.

PUT/DELETE are not handled (from now on). In this API all request methods do the same thing.

hsanjuan:
I have a test failure that seems unrelated: 

```
=== Failed
=== FAIL: core/coreapi/test TestIface/Dht/TestDhtFindPeer (1.29s)
panic: runtime error: index out of range [0] with length 0 [recovered]
	panic: runtime error: index out of range [0] with length 0

```

hsanjuan:
@lidel can you help with the webui test failure (I have the feeling it will take you way shorter than me to figure out the place to fix it).

lidel:
I will take a look as soon I get back home (eta 1h)

hsanjuan:
Ok:

* Sharness is green

Pending:

* Reviews
* Need to merge and bubble https://github.com/ipfs/iptb-plugins/pull/37
* Need to fix webui test
* Final squash and fix commit messages.

core/corehttp/commands.go
```diff
@@ -117,11 +117,17 @@ func patchCORSVars(c *cmdsHttp.ServerConfig, addr net.Addr) {
...
 }
 
-func commandsOption(cctx oldcmds.Context, command *cmds.Command) ServeOption {
+func commandsOption(cctx oldcmds.Context, command *cmds.Command, allowGet bool) ServeOption {
 	return func(n *core.IpfsNode, l net.Listener, mux *http.ServeMux) (*http.ServeMux, error) {
 
 		cfg := cmdsHttp.NewServerConfig()
-		cfg.SetAllowedMethods(http.MethodGet, http.MethodPost, http.MethodPut)
+		cfg.AllowGet = allowGet
+		corsAllowedMethods := []string{http.MethodPost, http.MethodPut}
```
Stebalien, line 125: I'm not currently allowing PUT in the commands lib. Will that be an issue?

lidel: AFAIK the only place where we use `PUT` is in "writable gateway",
which is on the Gateway port, but outside of `/api/v0` namespace.
Is there other use?

Stebalien: I don't think so. I'll remove this.

lidel:
PSA: this will require IPFS Companion  v2.11.x or later (shipped a Beta for testing: [v2.11.0.904](https://github.com/ipfs-shipyard/ipfs-companion/releases/tag/v2.11.0.904))

alexander255:
Of course I'll adapt to this in py-ipfs-http-client, but may I ask what the rationale is/was for blocking GET on idempotent & read-only API endpoints? Why can't things like `/api/v0/get?arg=<cid>` continue to be allowed GET access additionally to POST?

Stebalien:
Basically, the HTTP API is not a REST API, it's a simple RPC API. In go-ipfs, at least, it's defined as a set of transport agnostic commands, then exposed through an HTTP API.

Note: You can continue to use GET requests against the read-only API on the gateway.

alexander255:
Too bad :slightly_frowning_face:… I have some code deployed that uses the HTTP API as read-only API since its port is stable (5001) while the gateway port is unusable (8080 is way to generic). The decision to also enforce this for the read-only parts of the API parts that are also exposed via the gateway is final, I suppose? Does the read-only API exposed via the gateway support using POST or does only support GET instead?

hsanjuan:
> Does the read-only API exposed via the gateway support using POST or does it only support GET instead?

I think both.

Stebalien:
We have no intention of changing this as it simplifies reasoning about the system. Is it not possible for you to use POST?

lidel:
@alexander255 @hsanjuan 👉 **`POST` restriction is only on the API port**

A subset of `/api/v0` exposed on the Gateway accepts `GET` just fine (0.5.0-rc2):

```console
$ curl -X GET http://127.0.0.1:5001/api/v0/dns/ipfs.io
405 - Method Not Allowed

$ curl -X GET http://127.0.0.1:8080/api/v0/dns/ipfs.io
{"Path":"/ipfs/bafybeibdwwmb725qy4yph56jg4bs6uimcst5hnvuraarowgdaytzmgsw24"}
```

alexander255:
With the read-only API on gateway supporting POST, I can still pretend they are the same so no problem for me then. I just feels weird to use POST there, but if it makes stuff easier for you so be it. :+1:

PR #7103, opened 2020-04-06
feat(dht): update to cypress DHT in backwards compatibility mode

Stebalien:
This upgrades to the next DHT version, but in backwards compatibility mode.

### DIFF
--- a/core/node/libp2p/routingopt.go
+++ b/core/node/libp2p/routingopt.go
@@ -26,7 +26,9 @@ func constructDHTRouting(mode dht.ModeOpt) func(ctx context.Context, host host.H
 }
 
 var (
-	DHTOption       RoutingOption = constructDHTRouting(dht.ModeAuto)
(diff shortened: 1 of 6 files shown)


core/node/libp2p/routingopt.go
```diff
@@ -8,26 +8,40 @@ import (
...
+
 type RoutingOption func(context.Context, host.Host, datastore.Batching, record.Validator) (routing.Routing, error)
 
 func constructDHTRouting(ctx context.Context, host host.Host, dstore datastore.Batching, validator record.Validator) (routing.Routing, error) {
 	return dht.New(
 		ctx, host,
-		dhtopts.Datastore(dstore),
-		dhtopts.Validator(validator),
+		append([]dht.Option{
+			dht.Concurrency(10),
```
willscott, line 28: duplicated?

Stebalien: Ah, oops.

PR #7127, opened 2020-04-09
feat: introduce the dual WAN/LAN DHT

Stebalien:
And only join the WAN DHT if we're publicly reachable.

TODO:

* [ ] Enable AutoNATService by default on all DHT nodes?
* [ ] Add more tests.

### DIFF
--- a/core/commands/dht.go
+++ b/core/commands/dht.go
@@ -81,19 +81,23 @@ var queryDhtCmd = &cmds.Command{
 			dht = nd.DHT.LAN
 		}
 
-		closestPeers, err := dht.GetClosestPeers(ctx, string(id))
-		if err != nil {
-			cancel()
-			return err
-		}
-
+		errCh := make(chan error, 1)
 		go func() {
+			defer close(errCh)
 			defer cancel()
-			for p := range closestPeers {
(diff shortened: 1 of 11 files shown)


Stebalien:
Testing Plan:

1. Make a network of 50 nodes with loopback addresses, and a network of 50 nodes with WAN addresses.
2. Add one node to both network.
3. Verify that the node on both networks can find a peer they're not already connected two on both networks.

The test should use the "mock" network and should live in the "integration" test folder. We can model it after https://github.com/ipfs/go-ipfs/blob/5ecf44a88217f46472aa535a54153edc2e2bede6/test/integration/three_legged_cat_test.go#L67-L150.

Gotchas:

* The mocknet doesn't make any effort to pick the right source/target IP when creating a fake connection. That should be fine in this case as long as only the _test_ node lives on both networks. All other nodes only live on one or the other.

@willscott think you could pick this up?

willscott:
integration test to merge into this branch in https://github.com/ipfs/go-ipfs/pull/7151

Stebalien:
I'm going to merge this for now given that we now have a test PR and we're not going to ship a release without this.

Issue #7123, opened 2020-04-09
Strage sharness failure in CAR tests


Stebalien:
> Error: Unknown command `""$("`

https://app.circleci.com/pipelines/github/ipfs/go-ipfs/2385/workflows/312a4b14-636d-4e3a-a327-4cf9ba8db077/jobs/29097

ribasushi:
Spent 2h chasing the wrong problem, the error you see is an incorrect argument pass 🤦 

The timeout however is indicative of another problem, I only managed to get it once locally in a slightly different config. Will finish figuring it out + PR tomorrow

Stebalien:
Another case: https://app.circleci.com/pipelines/github/ipfs/go-ipfs/2441/workflows/4c474b78-9880-48d4-9f2d-cc50329a419b/jobs/29650/artifacts

This is after disabling the GC logic. I'm pretty sure it's the FIFO import.

PR #7140, opened 2020-04-10
Beef up and harden import/export tests

ribasushi:
Fixes ( or at least makes cleaner ) https://github.com/ipfs/go-ipfs/issues/7123

At this point I am convinced that this comment I made way back is in fact real: https://github.com/ipfs/go-ipfs/pull/7011/files/aa5019413741f2d34c9723d376591cf8c4b6b465#diff-15172926c4147422f472139313467179R330-R341
We did switch away from gclock a while back, and likely other changes I made cause it to trigger less frequently. Reproducing this locally is nigh impossible however :(

So instead the timeout+QUIT contraptions should tell us more next time it happens on the slower circleci boxes.

### DIFF
--- a/test/sharness/t0054-dag-car-import-export.sh
+++ b/test/sharness/t0054-dag-car-import-export.sh
@@ -29,14 +29,13 @@ reset_blockstore() {
 do_import() {
   node=$1; shift
 
-  bash -c "while [[ -e spin.gc ]]; do ipfsi $node repo gc >>gc_out 2>&1; done" & gc1_pid=$!
-  bash -c "while [[ -e spin.gc ]]; do ipfsi $node repo gc >>gc_out 2>&1; done" & gc2_pid=$!
+  touch spin.gc
+  timeout -s QUIT 15 bash -c "while [[ -e spin.gc ]]; do ipfsi $node repo gc &>>gc_out; done" & gc1_pid=$!
+  timeout -s QUIT 15 bash -c "while [[ -e spin.gc ]]; do ipfsi $node repo gc &>>gc_out; done" & gc2_pid=$!
 
-  ipfsi $node dag import "$@"
+  timeout -s QUIT 10 bash -c "ipfsi $node dag import $* 2>&1"
 
   rm -f spin.gc || true
-  sleep 3
-  kill $gc1_pid $gc2_pid || true
 }
 
 run_online_imp_exp_tests() {
@@ -56,7 +55,6 @@ EOE
 EOE
 
 
-  touch spin.gc
   test_expect_success "basic import" '
     do_import 0 \
       ../t0054-dag-car-import-export-data/combined_naked_roots_genesis_and_128.car \
@@ -118,7 +116,6 @@ EOE
     bash -c "sleep 60; kill $cat1_pid $cat2_pid 2>/dev/null" &
   ' &
 
-  touch spin.gc
   test_expect_success "fifo import" '
     do_import 0 \
       pipe_testnet \
--- a/test/sharness/t0054-dag-car-import-export.sh
+++ b/test/sharness/t0054-dag-car-import-export.sh
@@ -9,6 +9,7 @@ export -f ipfsi
 set -o pipefail
 
 tar -C ../t0054-dag-car-import-export-data/ --strip-components=1 -Jxf ../t0054-dag-car-import-export-data/test_dataset_car_v0.tar.xz
+tab=$'\t'
 
 reset_blockstore() {
   node=$1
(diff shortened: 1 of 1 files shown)


Stebalien:
Are we talking about two separate issues here? There's:

1. The sharness issue.
2. The GC issue.

Your tests prove that this issue can't be related to GC locks by successfully verifying that none of the pinned blocks in the CARs are garbage collected. If this were a GC race, we'd be deleting pinned data.

We _know_ what at least part of the GC issue is. The issue is that we're garbage collecting parts of the pin set. We could be generating and garbage collecting additional data, but we'd have to look into that by looking at the blocks GC is deleting.

ribasushi:
@Stebalien not a GC race - a GC **deadlock**. I am 99% sure this is what you observed in 7123. What I am modifying is that if we trigger a generous timeout: we will dump a full stacktrace where I "expect nothing" otherwise.

The other thing ( ephemeral pinsets showing up ) is tracked in a different ticket.

Stebalien:
Ah, got it.

The bug we were running into _there_ was that we were invoking the CoreAPI while already holding the PinLock. However, we could definitely have a different deadlock somewhere.

Stebalien:
Doing some spelunking: I don't think it's a starvation issue. That is, repeated GC should eventually allow a `dag import` to proceed.

test/sharness/lib/0001-Detect-type-of-aliasing-unsupported-by-some-shells.patch
```diff
@@ -0,0 +1,46 @@
+From 6a43df1919da862f56c9af867a28bf2d7858e5ca Mon Sep 17 00:00:00 2001
```
Stebalien, line 1: I agree that the behavior of sharness is surprising, but I'm concerned that this will cause more problems than it fixes. It already has a bunch of special cases to avoid rejecting valid code.

I'd rather document this in the sharness README and move on unless we can find a more reliable way to do this.

test/sharness/t0054-dag-car-import-export.sh
```diff
@@ -9,44 +9,52 @@ export -f ipfsi
 set -o pipefail
 
 tar -C ../t0054-dag-car-import-export-data/ --strip-components=1 -Jxf ../t0054-dag-car-import-export-data/test_dataset_car_v0.tar.xz
+tab=$'\t'
 
 reset_blockstore() {
-  node=$1
-  ipfsi $1 pin ls --quiet --type=recursive | ipfsi $1 pin rm &>/dev/null
-  ipfsi $1 repo gc &>/dev/null
+  node=$1; shift
```
Stebalien, line 15: Do we need the shift?

test/sharness/lib/0001-Detect-type-of-aliasing-unsupported-by-some-shells.patch
ribasushi: Nobody will ever discover the sharness README :)
No worries though - I will pull this commit out, and use it locally instead.
Unless you are open to have it gated on an environment variable...?

test/sharness/t0054-dag-car-import-export.sh
ribasushi: Just symmetry with the next function. I will remove it.

test/sharness/lib/0001-Detect-type-of-aliasing-unsupported-by-some-shells.patch
ribasushi: Force-pushed a removal ( it was the first commit )

Stebalien: Ideally, we'd trigger it when we get a failure. However, in this case, that wouldn't have helped regardless because we weren't checking for errors.

You're right that nobody will read the README... Honestly, it's _not_ terrible given that we can catch and deal with any false positives. What if we include it but put in a comment explaining exactly what it does, where it comes from (it's non-obvious that we're patching it in), and that it's a heuristic and that we should remove it if it causes problems?

ribasushi: I'll open a separate PR once I had a chance to write up better prose. The the error will link to that.

ribasushi: For now let's merge the rest to catch the lock if it happens again ( all the `"$(` stuff, while annoying was a red herring )

Issue #6811, opened 2020-01-02
Adding some go.mod information along the example


constraintAutomaton:
#### Location

`docs/examples/go-ipfs-as-a-library`

#### Description

I found the example clear and helpful but when it comes to applying it to a real project, it can be hard without information about go.mod , just adding some information about how to start a real project could really help the begginer.

Stebalien:
We can actually just add a go.mod file to that subdirectory.

chelneru:
Can you provide some information on how to start a real project? I am struggling to get the example running outside of the project repository.  I tried to copy the go.mod and go.sum files but still it's not working.

hsanjuan:
> Can you provide some information on how to start a real project? I am struggling to get the example running outside of the project repository. I tried to copy the go.mod and go.sum files but still it's not working.

Can't help without details of what exactly is not working...

chelneru:
Thanks for reply @hsanjuan ,

I want to run the example of using go-ipfs as a library outside of ipfs's folder from repository. 

I have a separate folder  `%GOPATH%\src\github.com\example` with 

- main.go
- go.sum
- go.mod

Main.go contains the example from using ipfs as a library.
Go.mod, I copied the [contents](https://github.com/ipfs/go-ipfs/blob/master/go.mod)from the repository while changing the go version at the bottom to 1.14, changed the module name at the top and added an extra row for requirements:
`github.com/ipfs/go-ipfs v0.5.0-dev` 

Go.sum is identical to the one on the ipfs repository.

I run `go get` inside the folder to retrieve all the dependencies and when I run the file I get the following errors :
github.com/libp2p/go-libp2p-loggables
..\libp2p\go-libp2p-loggables\loggables.go:19:26: undefined: log.Loggable
..\libp2p\go-libp2p-loggables\loggables.go:20:9: undefined: log.Metadata
..\libp2p\go-libp2p-loggables\loggables.go:27:21: undefined: log.Loggable
..\libp2p\go-libp2p-loggables\loggables.go:28:9: undefined: log.Metadata
..\libp2p\go-libp2p-loggables\loggables.go:33:23: undefined: log.Metadata
..\libp2p\go-libp2p-loggables\loggables.go:38:9: undefined: log.Metadata
 github.com/ipfs/go-ipfs-blockstore
..\ipfs\go-ipfs-blockstore\blockstore.go:224:8: log.Warningf undefined (type *log.ZapEventLogger has no field or method Warningf)
[...]

PR #7146, opened 2020-04-12
Add module files for go-ipfs-as-a-library example

lthibault:
Fixes https://github.com/ipfs/go-ipfs/issues/6811.


hsanjuan:
Should we just document how to setup the example on a separate location?

My main issue with this is that we might forget to keep this go.mod up to date, therefore the example might break when people want to use it with newer go-ipfs versions.

Stebalien:
Eh, I'd add it. It's a good starting point.

However, we should probably wait till the release so we can make it depend on v0.5.0 instead of a specific git commit.

hsanjuan:
I put it in the milestone so that we don't forget...

Stebalien:
@lthibault could you rebase this and update go-ipfs to 0.5.0?

lthibault:
@Stebalien Done.

PR #7147, opened 2020-04-13
feat(cmd): add silent option for command repo gc

### LINKED ISSUE #7129: ipfs repo gc --quiet  is  not quiet
#### Version information:

0.5.0-rc1 with badgerds and enabled sharding

#### Description:

Calling `ipfs repo gc --quiet` does not seem to do anything, 
`--quiet` still prints every (removed?) CID. 

My expectation would  be to supress all output, or only print summary at the end.

will7200:
closes #7129

### DIFF
--- a/core/commands/repo.go
+++ b/core/commands/repo.go
@@ -51,6 +51,7 @@ type GcResult struct {
 const (
 	repoStreamErrorsOptionName = "stream-errors"
 	repoQuietOptionName        = "quiet"
+	repoSilentOptionName       = "silent"
 )
 
 var repoGcCmd = &cmds.Command{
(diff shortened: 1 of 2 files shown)


ribasushi:
@will7200 this is great 🤩 
Would you be able to add some tests that silence indeed takes place? Something like https://github.com/ipfs/go-ipfs/blob/5ecf44a88/test/sharness/t0080-repo.sh#L18-L34 but comparing the `..._actual` with `/dev/null` instead.

To run a test locally you do:
```
make test_sharness_deps
cd test/sharness
bash the-test-you-want-to-run.sh
```

Thanks!

P.S. ignore the current test failures you received, we are looking into them...

will7200:
@ribasushi
I added the test case. Let me know if this is sufficient or should be expanded on.

lidel (approved):
Better late than never! :upside_down_face:

PR #7151, opened 2020-04-14
integration test for the dual dht

willscott:
This adds an integration test.

Currently, an `*IpfsDHT` is exposed by the core, which will not be populated due to the use of the `dual` DHT. this test will not be able to refresh the routing table as a result, and will fail at present on the `feat/dual-dht` branch.

### DIFF
--- a/test/integration/wan_lan_dht_test.go
+++ b/test/integration/wan_lan_dht_test.go
@@ -102,6 +102,7 @@ func RunDHTConnectivity(conf testutil.LatencyConfig, numPeers int) error {
 		wanAddr := makeAddr(uint32(i), true)
 		wanPeer.Peerstore.AddAddr(wanPeer.Identity, wanAddr, peerstore.PermanentAddrTTL)
 		for _, p := range wanPeers {
+			mn.LinkPeers(p.Identity, wanPeer.Identity)
 			mn.ConnectPeers(p.Identity, wanPeer.Identity)
 		}
 		wanPeers = append(wanPeers, wanPeer)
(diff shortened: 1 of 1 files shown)


willscott:
the test now passes locally when run with a go-libp2p containing the fix from https://github.com/libp2p/go-libp2p/pull/887

test/integration/wan_lan_dht_test.go
```diff
@@ -0,0 +1,277 @@
...
+
+	// create network
+	mn := mocknet.New(ctx)
+	mn.SetLinkDefaults(mocknet.LinkOptions{
+		Latency:   conf.NetworkLatency,
+		Bandwidth: math.MaxInt32,
+	})
+
+	testPeer, err := core.NewNode(ctx, &core.BuildCfg{
+		Online: true,
```
Stebalien, line 81: We should start the test peer as a client. Otherwise, the test record could get put on the test peer.

test/integration/wan_lan_dht_test.go
```diff
@@ -0,0 +1,277 @@
...
+
+	// choose a lan peer and validate lan DHT is functioning.
+	i := rand.Intn(len(lanPeers))
+	if testPeer.PeerHost.Network().Connectedness(lanPeers[i].Identity) == corenet.Connected {
+		i = (i + 1) % len(lanPeers)
+		if testPeer.PeerHost.Network().Connectedness(lanPeers[i].Identity) == corenet.Connected {
+			_ = testPeer.PeerHost.Network().ClosePeer(lanPeers[i].Identity)
+			testPeer.PeerHost.Peerstore().ClearAddrs(lanPeers[i].Identity)
+		}
+	}
```
Stebalien, line 177: Now that we're trying to find a record, not a peer, this shouldn't be necessary.

test/integration/wan_lan_dht_test.go
```diff
@@ -0,0 +1,277 @@
...
+
+	// choose a wan peer and validate wan DHT is functioning.
+	i = rand.Intn(len(wanPeers))
+	if testPeer.PeerHost.Network().Connectedness(wanPeers[i].Identity) == corenet.Connected {
+		i = (i + 1) % len(wanPeers)
+		if testPeer.PeerHost.Network().Connectedness(wanPeers[i].Identity) == corenet.Connected {
+			_ = testPeer.PeerHost.Network().ClosePeer(wanPeers[i].Identity)
+			testPeer.PeerHost.Peerstore().ClearAddrs(wanPeers[i].Identity)
+		}
+	}
```
Stebalien, line 236: ditto. Now that we're calling FindProviders instead of FendPeer, disconnecting doesn't buy us anything.

PR #7154, opened 2020-04-14
test(sharness): improve CAR tests to remove some potential races

Stebalien:
* Instead of ensuring that we GC nothing, verify that the repo validates.
* Avoid a race between removing the fifos and trying to read from them.
* Avoid all 'bash -c' invocations. Instead, make liberal use of subshells and wait.

Hopefully fixes #7123.

PR #7155, opened 2020-04-14
Chore/reintroduce fifo test

ribasushi:
Bring back the test removed in https://github.com/ipfs/go-ipfs/pull/7154, with better description and workflow.

This should no longer hang, while still tests the actual use case for dagger

I would still like to add the timeouts here and there, but there's no way to do that over a shell function (`ipfsi`) without `bash -c`

### DIFF
--- a/test/sharness/t0054-dag-car-import-export.sh
+++ b/test/sharness/t0054-dag-car-import-export.sh
@@ -11,6 +11,13 @@ set -o pipefail
 tar -C ../t0054-dag-car-import-export-data/ --strip-components=1 -Jxf ../t0054-dag-car-import-export-data/test_dataset_car_v0.tar.xz
 tab=$'\t'
 
+test_cmp_sorted() {
+  # use test_cmp to dump out the unsorted file contents as a diff
+  [[ "$( sort "$1" | sha256sum )" == "$( sort "$2" | sha256sum )" ]] \
+    || test_cmp "$1" "$2"
+}
+export -f test_cmp_sorted
+
 reset_blockstore() {
   node=$1
 
@@ -65,11 +72,11 @@ EOE
       ../t0054-dag-car-import-export-data/combined_naked_roots_genesis_and_128.car \
       ../t0054-dag-car-import-export-data/lotus_testnet_export_128_shuffled_nulroot.car \
       ../t0054-dag-car-import-export-data/lotus_devnet_genesis_shuffled_nulroot.car \
-    | sort > basic_import_actual
+    > basic_import_actual
   '
 
   test_expect_success "basic import output as expected" '
-    test_cmp basic_import_expected basic_import_actual
+    test_cmp_sorted basic_import_expected basic_import_actual
   '
 
   test_expect_success "basic fetch+export 1" '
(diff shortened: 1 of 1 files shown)


Stebalien:
> Bring back the test removed in #7154, with better description and workflow.

Which tests? That PR didn't remove any tests.

test/sharness/t0054-dag-car-import-export.sh
```diff
@@ -105,29 +113,60 @@ EOE
...
+        # ipfs import starts and blocks on part 1
+        ipfsi 0 dag import \
+          pipe_testnet \
+          ../t0054-dag-car-import-export-data/combined_naked_roots_genesis_and_128.car \
+          pipe_devnet \
+        2>&1 & ipfs_pid=$!
+
+        # Flush part 1, unblock everything, delete pipes as soon as possible
+        timeout 30 cat ../t0054-dag-car-import-export-data/lotus_testnet_export_128_shuffled_nulroot.car > pipe_testnet
+        rm -f pipe_devnet pipe_testnet
```
Stebalien, line 157: This races with opening `pipe_devnet` in ipfs. There's no guarantee that we'll open it till the command has finish.

ribasushi:
> Avoid a race between removing the fifos and trying to read from them.

that *was* a test ;)

test/sharness/t0054-dag-car-import-export.sh
```diff
@@ -105,29 +113,60 @@ EOE
...
+
+        wait "$ipfs_pid"
```
Stebalien, line 159: No need to wait on a specific PID. This should wait for _all_ children (including the `cat`).

test/sharness/t0054-dag-car-import-export.sh
```diff
@@ -105,29 +113,60 @@ EOE
...
+  reset_blockstore 0
+  reset_blockstore 1
+
+  test_expect_success "sequenced fifo import" '
+    (
+        # Part 2 is smaller, goes in first and blocks waiting on ipfs
+        # What is tested here is that by the time import 1 + roots finishes and the
+        # pipes are deleted, the preexisting open file handles will carry through,
+        # even though pipe_devnet is no longer on the filesystem
+        timeout 30 cat ../t0054-dag-car-import-export-data/lotus_devnet_genesis_shuffled_nulroot.car > pipe_devnet &
```
Stebalien, line 146: I would remove all these timeouts or set them to several minutes. CI can be _very_ slow at times.

Stebalien:
> that was a test ;)

Ok, I think I understand. However, IPFS makes no guarantees about opening all files before it starts adding them. We should not be trying to test something IPFS doesn't guarantee.

test/sharness/t0054-dag-car-import-export.sh
ribasushi: > There's no guarantee that we'll open it till the command has finish.

The current code does open **every** piece on the cli before processing the first one. This is very handy for fifo processing. I was merely ensuring this stays as is since I was planning to rely on that.

This is what the test was for originally, I just wrote it badly. If this is not something you want to guarantee going forward: no biggie, I'll close.

PR #7162, opened 2020-04-15
add autonat config options

Stebalien:
1. Enable AutoNATService on _all_ nodes by default. If it's an issue, we can disable it in RC3 but this will give us the best testing results.
2. Expose options to configure AutoNAT rate limiting.


Stebalien:
> is it worth noting that Enabled actually means that the service will become active if it determines the local node is publicly reachable by others?

Didn't we change that default to "if it doesn't know that it's not reachable"?

Stebalien:
But yeah, I'll quickly document that.

PR #7165, opened 2020-04-15
fix: subdomain redirect for dir CIDs

### LINKED ISSUE #7164: Subdomain redirect doesn't apply to directory listings
#### Version information:

> v0.5.0-rc2

#### Description:

When listing a _directory_ with no index file, the subdomain gateway redirect isn't applied.

> http://localhost:8080/ipns/dist.ipfs.io/go-ipfs

lidel:
Closes #7164

Note: due to the way `http` in golang works, status code can be set only once. 
There is a known issue with `prometheus/client_golang`: it sometimes produces `superfluous response.WriteHeader call` and if it is executed first, the status code set in our code is ignored. 

In this PR I set the redirect status code before executing the template to avoid potential race condition, feels like the least invasive way of making this work reliably.

### DIFF
--- a/core/corehttp/gateway_handler.go
+++ b/core/corehttp/gateway_handler.go
@@ -342,6 +342,16 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
 		BackLink: backLink,
 		Hash:     hash,
 	}
+
+	// See statusResponseWriter.WriteHeader
+	// and https://github.com/ipfs/go-ipfs/issues/7164
+	// Note: this needs to occur before listingTemplate.Execute otherwise we get
+	// superfluous response.WriteHeader call from prometheus/client_golang
+	if w.Header().Get("Location") != "" {
+		w.WriteHeader(http.StatusMovedPermanently)
+		return
+	}
+
 	err = listingTemplate.Execute(w, tplData)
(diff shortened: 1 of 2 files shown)


core/corehttp/gateway_handler.go
```diff
@@ -342,6 +342,15 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
 		BackLink: backLink,
 		Hash:     hash,
 	}
+
+	// See statusResponseWriter.WriteHeader
+	// and https://github.com/ipfs/go-ipfs/issues/7164
+	// Note: this needs to occur before listingTemplate.Execute otherwise we get
+	// superfluous response.WriteHeader call from prometheus/client_golang
+	if w.Header().Get("Location") != "" {
+		w.WriteHeader(http.StatusMovedPermanently)
```
lidel, line 351: @Stebalien Do we need to return directory listing payload with the redirect?
I think it would be safer to `return` here, right after we write the status code header.

Stebalien: I agree. Directory listings are for browsers.

lidel: @Stebalien updated this PR to include mentioned `return`

PR #7199, opened 2020-04-23
update log helptext to match actual levels

djdv:
It looks like "warning" and "critical" levels were removed, but the help text wasn't updated to reflect this.
```
> ipfs.exe version --all
go-ipfs version: 0.5.0-rc3
Repo version: 9
System version: amd64/windows
Golang version: go1.13.10
> ipfs.exe log level all warning
Error: unrecognized level: "warning"
> ipfs.exe log level all critical
Error: unrecognized level: "critical"
```

### DIFF
--- a/core/commands/log.go
+++ b/core/commands/log.go
@@ -24,7 +24,7 @@ output of a running daemon.
 
 There are also two environmental variables that direct the logging 
 system (not just for the daemon logs, but all commands):
-    IPFS_LOGGING - sets the level of verbosity of the logging. One of: debug, info, warning, error, critical
+    IPFS_LOGGING - sets the level of verbosity of the logging. One of: debug, info, warn, error, dpanic, panic, fatal
     IPFS_LOGGING_FMT - sets formatting of the log output. One of: color, nocolor
 `,
 	},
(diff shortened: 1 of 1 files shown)


core/commands/log.go
```diff
@@ -24,7 +24,7 @@ output of a running daemon.
 
 There are also two environmental variables that direct the logging 
 system (not just for the daemon logs, but all commands):
-    IPFS_LOGGING - sets the level of verbosity of the logging. One of: debug, info, warning, error, critical
+    IPFS_LOGGING - sets the level of verbosity of the logging. One of: debug, info, error
```
Stebalien, line 27:
```suggestion
    IPFS_LOGGING - sets the level of verbosity of the logging. One of: debug, info, warn, error
```

core/commands/log.go
```diff
@@ -49,8 +49,8 @@ the event log.
 		// TODO use a different keyword for 'all' because all can theoretically
 		// clash with a subsystem name
 		cmds.StringArg("subsystem", true, false, fmt.Sprintf("The subsystem logging identifier. Use '%s' for all subsystems.", logAllKeyword)),
-		cmds.StringArg("level", true, false, `The log level, with 'debug' the most verbose and 'critical' the least verbose.
-			One of: debug, info, warning, error, critical.
+		cmds.StringArg("level", true, false, `The log level, with 'debug' the most verbose and 'error' the least verbose.
+			One of: debug, info, error.
```
Stebalien, line 53:
```suggestion
			One of: debug, info, warn, error.
```

djdv:
I noticed the other values like `DPANIC` within the log package are accepted via this command too. Should they be added?
>// Supported strings are: DEBUG, INFO, WARN, ERROR, DPANIC, PANIC, FATAL, and
// their lower-case forms.

Stebalien:
Ideally, yes.

djdv:
Should be all good now.
(I did notice some of the helptext keeps to the 80 character bound, but not all of it. So I'm guessing this is fine without an explicit line-break)

Stebalien:
I'd try to keep to 80 characters, we don't wrap elsewhere.

Could you rebase on master (and squash)? That should fix CI.

djdv:
I put the possible values on a newline for both `IPFS_LOGGING` and `IPFS_LOGGING_FMT` on a newline that's indented under them. Both fall under 80 and look consistent. If they should look another way just let me know.

This is how it's displaying for me.
image

Issue #7198, opened 2020-04-23
Context-Scoped Sessions


Stebalien:
Currently, bitswap sessions are hard to work with because we need to:

1. Create a new session.
2. Pass this session to all services that need access to the exchange.

Usually 2 involves _re-creating_ these services. This can be very painful and we have many places in go-ipfs where we simply aren't doing this.

Instead, we should associate a session with a _context_. That way, whenever we use the context, we use the session. When we cancel the context, we cancel the session.

## Design

Store a session "key" in the session to identify the session.

```go
// SessionKey is an opaque type uniquely identifying a session. Each SessionKey
// should be allocated with new(SessionKey).
type SessionKey struct{}

// private types for storing session information 

type sessionContextKey struct{}
type sessionContextValue struct {
	sesKey *SessionKey
	sesCtx context.Context
}

// NewSession registers a new session with the context. The session will be
// closed when the passed-in context is canceled.
//
// If there's already a session associated with the context, the existing
// session will be used.
//
// This function does not initialize any state, it just creates a new
// *SessionKey and associates it with the context.
func NewSession(ctx context.Context) context.Context {
	if _, ok := ctx.Value(sessionContextKey{}).(*sessionContextValue); ok {
		// Leave the session alone.
		return ctx
	}
	return context.WithValue(ctx, sessionContextKey{}, &sessionContextValue{
		sesKey: new(SessionKey),
		sesCtx: ctx, // save this so we can construct the session with the _session_ context.
	})
}

// GetSession loads the fetcher from the context. The returned context is the
// context associated with the session itself.
//
// If there is no session associated with this context, this function returns
// a new session key and the passed-in context.
func GetSession(ctx context.Context) (*SessionKey, context.Context) {
	if s, ok := ctx.Value(sessionContextKey{}).(*sessionContextValue); ok {
		return s.sesKey, s.sesCtx
	}
	return new(SessionKey), ctx
}
```

## Alternatives

I've considered two alternatives:

1. Store the exchange/fetcher in the context.
2. Store an abstract session in the context.

### Embedded Exchange

We could:

1. Create a session.
2. Store the session/"fetcher" in the context.
3. Use this session inside the blockservice instead of the actual exchange.

On one hand, this is nice because we can embed alternative exchanges for a specific request. For example, we could embed an _offline_ exchange to perform an offline lookup.

On the other hand:

1. This is spooky action at a distance.
2. This _only_ solves this particular problem and doesn't help services exchange information related to sessions.

[...]

PR #7201, opened 2020-04-23
feat: use the context to thread sessions through all commands

Stebalien:
* Ensure we have a single session per gateway request.
* Use a session in the ipfs refs command.
* Start a session before resolving the pin root (also fixes sessions while pinning).
* Update go-merkledag to reliably create sessions.

part of https://github.com/ipfs/go-ipfs/issues/7198


core/commands/dag/dag.go
```diff
@@ -552,6 +553,10 @@ The output of blocks happens in strict DAG-traversal, first-seen, order.
 			return err
 		}
 
+		ctx, cancel := context.WithCancel(req.Context)
+		ctx = exchange.NewSession(req.Context)
```
dirkmc, line 557: Not sure if I'm following this right, should this be
```suggestion
		ctx = exchange.NewSession(ctx)
```

dirkmc:
We should probably merge this PR first: [Send CANCELs when session context is cancelled](https://github.com/ipfs/go-bitswap/pull/375)

Mitumaru (approved):

PR #7202, opened 2020-04-23
fix: fix duplicate block issue in bitswap

Stebalien:
1. Increase broadcast timeout.
2. Fix use of sessions when pinning.

This also includes some bitswap changes to reduce some memory allocations and optimize some hot spots.


core/node/core.go
```diff
@@ -75,6 +80,10 @@ func (s *syncDagService) Sync() error {
 	return s.syncFn()
 }
 
+func (s *syncDagService) Session(ctx context.Context) format.NodeGetter {
```
Stebalien, line 83: This lets mdag.FetchGraph correctly setup a bitswap session. A better fix is in #7201 but that's not going to make it into the release.

PR #7204, opened 2020-04-23
chore: update dht

Stebalien:
fixes a small bug in query notifications that was causing a test to fail


Stebalien:
@aschmahmann is updating go-libp2p-discovery sufficient to fix the race?

Issue #7205, opened 2020-04-23
Missing content path when dir listing on subdomain or DNSLink gw


lidel:
#### Version information:

0.5.0-rc3

#### Description:

Directory listing in Subdomain mode is missing the CID info at the top:

> ![path-2020-04-23--14-02-07](https://user-images.githubusercontent.com/157609/80097239-43942a80-856b-11ea-8428-3ffe9a5f01d5.png)

> ![subdomain-2020-04-23--14-01-33](https://user-images.githubusercontent.com/157609/80097251-4858de80-856b-11ea-81b9-ef1d9730ea92.png)

I am not sure if it is a feature (could be something we did for DNSLink websites in past), but feels like a bug in subdomain context. 

I'd like to change this so we ALWAYS show the full content path in `Index of ..`
(including subdomains and DNSLink websites)

Thoughts? @jessicaschilling @Stebalien

jessicaschilling:
Sounds good to me. If it's not too off-topic, this would also be a good chance to fix https://github.com/ipfs/dir-index-html/issues/2 -- in which the content of "Index of ..." is separately linked for each directory (so you can use it to traverse directory structure by clicking). I'm afraid it's not something I can fix, though. :(

Stebalien:
I agree. We should always show the full path, even for DNSLink websites.

PR #7219, opened 2020-04-26
feat(gateway): show the absolute path and CID every time

Stebalien:
For gateway directory listings, always the absolute path and the CID of the current directory. Even for dnslink websites.

fixes https://github.com/ipfs/go-ipfs/issues/7205


Stebalien:
2020-04-26-161619_1279x834_scrot

PR #7221, opened 2020-04-27
feat: add a dht stat command

Stebalien:
Currently, it just prints out the routing tables.

TODO

* [ ] ~Tests~ It's a debugging command.
* [x] Maybe better output formatting?

### DIFF
--- a/core/commands/stat_dht.go
+++ b/core/commands/stat_dht.go
@@ -1,0 +1,194 @@
+package commands
+
+import (
+	"fmt"
+	"io"
+	"text/tabwriter"
+	"time"
+
+	cmdenv "github.com/ipfs/go-ipfs/core/commands/cmdenv"
+
+	cmds "github.com/ipfs/go-ipfs-cmds"
+	"github.com/libp2p/go-libp2p-core/network"
+	pstore "github.com/libp2p/go-libp2p-core/peerstore"
+	dht "github.com/libp2p/go-libp2p-kad-dht"
+	kbucket "github.com/libp2p/go-libp2p-kbucket"
+)
+
+type dhtPeerInfo struct {
+	ID            string
+	Connected     bool
+	AgentVersion  string
+	LastUsefulAt  string
+	LastQueriedAt string
+}
+
+type dhtStat struct {
+	Name    string
+	Buckets []dhtBucket
+}
+
+type dhtBucket struct {
+	LastRefresh string
+	Peers       []dhtPeerInfo
+}
+
+var statDhtCmd = &cmds.Command{
+	Helptext: cmds.HelpText{
+		Tagline: "Returns statistics about the node's DHT(s)",
+		ShortDescription: `
+Returns statistics about the DHT(s) the node is participating in.
+
+This interface is not stable and may change from release to release.
+`,
+	},
+	Arguments: []cmds.Argument{
+		cmds.StringArg("dht", false, true, "The DHT whose table should be listed (wan or lan). Defaults to both."),
+	},
+	Options: []cmds.Option{},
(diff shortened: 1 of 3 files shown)


core/commands/dht.go
```diff
@@ -39,13 +45,191 @@ var DhtCmd = &cmds.Command{
...
+	Name    string
+	Buckets []dhtBucket
+}
+
+type dhtBucket struct {
+	LastRefresh string
+	Peers       []dhtPeerInfo
+}
+
+var statDhtCmd = &cmds.Command{
```
willscott, line 74: given the level of information here, is it more sensible to have this live as `ipfs diag dht`?

core/commands/dht.go
```diff
@@ -39,13 +45,191 @@ var DhtCmd = &cmds.Command{
...
+	Helptext: cmds.HelpText{
+		Tagline: "Returns stats about the DHTs",
```
willscott, line 76:
```suggestion
		Tagline: "Returns statistics about the DHT",
```

core/commands/dht.go
```diff
@@ -39,13 +45,191 @@ var DhtCmd = &cmds.Command{
...
+
+		return nil
+	},
+	Encoders: cmds.EncoderMap{
+		cmds.Text: cmds.MakeTypedEncoder(func(req *cmds.Request, w io.Writer, out dhtStat) error {
+			tw := tabwriter.NewWriter(w, 4, 4, 2, ' ', 0)
+			defer tw.Flush()
+
+			// Regex to remove fraction from duration.
+			re := regexp.MustCompile(`\.\d+`)
```
willscott, line 169: declare as variable outside of function?

core/commands/dht.go
Stebalien: Possibly? I'm trying to mirror `ipfs bitswap stat` but I agree that lumping this all under the `diag` command probably makes more sense (and really, that's where the `ipfs bitswap` commands should live as well.

Stebalien: Actually, I'm not sure. `diag` is mostly for true diagnostics (running commands, system information, possibly pprof dumps in the future, etc.).

willscott:
It seems fine to have this as a DHT sub-command. I think the only question is if it's worth having a warning or otherwise making it clear that the format and underlying implementation-specific info that's showing up may change and this isn't a structure we're signing up to support long term.

core/commands/dht.go
Stebalien: Done.

core/commands/dht.go
Stebalien: Done.

Stebalien:
1. I've added a comment explaining that this command is unstable.
2. I've moved it to `ipfs stats dht`.

core/commands/stat_dht.go
```diff
@@ -0,0 +1,201 @@
...
+	Encoders: cmds.EncoderMap{
+		cmds.Text: cmds.MakeTypedEncoder(func(req *cmds.Request, w io.Writer, out dhtStat) error {
+			tw := tabwriter.NewWriter(w, 4, 4, 2, ' ', 0)
+			defer tw.Flush()
+
+			// Formats a time into XX ago and remove any decimal
+			// parts. That is, change "2m3.00010101s" to "2m3s ago".
+			now := time.Now()
+			since := func(t time.Time) string {
+				return string(decimalRegex.ReplaceAll(
```
willscott, line 145: locally, if i do
`now.Sub(t).Round(time.Second).String()` i don't see a decimal. When is this regex needed?

Stebalien: You're right... that's a lot simpler.

PR #7232, opened 2020-04-28
fix: non-blocking peerlog logging

Stebalien:
Avoid ever blocking new connections in the peer logger. Instead:

1. Send all new peers to a highly buffered channel.
2. Emit "dropped event" errors whenever we detect that we're dropping events and falling behind.
3. Don't log protocols, they're too large.
4. Don't log disconnects, we don't need them.

### DIFF
--- a/plugin/plugins/peerlog/peerlog.go
+++ b/plugin/plugins/peerlog/peerlog.go
@@ -2,16 +2,31 @@ package peerlog
 
 import (
 	"fmt"
+	"sync/atomic"
 
 	core "github.com/ipfs/go-ipfs/core"
 	plugin "github.com/ipfs/go-ipfs/plugin"
 	logging "github.com/ipfs/go-log"
-	eventbus "github.com/libp2p/go-eventbus"
 	event "github.com/libp2p/go-libp2p-core/event"
 	network "github.com/libp2p/go-libp2p-core/network"
+	"github.com/libp2p/go-libp2p-core/peer"
+	"github.com/libp2p/go-libp2p-core/peerstore"
+	"go.uber.org/zap"
 )
 
 var log = logging.Logger("plugin/peerlog")
+
+type eventType int
+
+const (
+	eventConnect eventType = iota
+	eventIdentify
+)
+
+type plEvent struct {
+	kind eventType
+	peer peer.ID
+}
 
 // Log all the PeerIDs we see
 //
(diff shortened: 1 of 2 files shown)


Stebalien:
Context: this is the plugin we use to count peers in the network. Unfortunately, something sometimes goes wrong with logging and it starts blocking. When that happens, we end up halting everything because we can't log.

This logs from a separate goroutine when we can so we don't completely block everything else.

plugin/plugins/peerlog/peerlog.go
```diff
@@ -41,60 +67,123 @@ func (*peerLogPlugin) Version() string {
...
+	}
+
 	var notifee network.NotifyBundle
 	notifee.ConnectedF = func(net network.Network, conn network.Conn) {
-		// TODO: Log transport, country, etc?
-		log.Infow("connected",
-			"peer", conn.RemotePeer().Pretty(),
-		)
-	}
-	notifee.DisconnectedF = func(net network.Network, conn network.Conn) {
```
Stebalien, line 60: We're intentionally removing this event.

plugin/plugins/peerlog/peerlog.go
```diff
@@ -41,60 +67,123 @@ func (*peerLogPlugin) Version() string {
...
-		log.Infow("disconnected",
-			"peer", conn.RemotePeer().Pretty(),
-		)
+		pl.emit(eventConnect, conn.RemotePeer())
 	}
 	node.PeerHost.Network().Notify(&notifee)
 
-	sub, err := node.PeerHost.EventBus().Subscribe(
-		new(event.EvtPeerIdentificationCompleted),
-		eventbus.BufSize(1024),
```
Stebalien, line 69: There's a default buffer size of 16 and that should be enough for us now.

plugin/plugins/peerlog/peerlog.go
```diff
@@ -41,60 +67,123 @@ func (*peerLogPlugin) Version() string {
...
-				agent, err := node.Peerstore.Get(e.Peer, "AgentVersion")
-				if err != nil {
-					log.Errorw("failed to get agent version", "error", err)
-					continue
-				}
-				log.Infow(
-					"identified",
-					"peer", e.Peer.Pretty(),
-					"agent", agent,
-					"protocols", protocols,
```
Stebalien, line 93: We're removing the protocols (too much data).

plugin/plugins/peerlog/peerlog.go
```diff
@@ -41,60 +67,123 @@ func (*peerLogPlugin) Version() string {
...
 // Init initializes plugin
-func (*peerLogPlugin) Init(*plugin.Environment) error {
+func (pl *peerLogPlugin) Init(*plugin.Environment) error {
+	pl.events = make(chan plEvent, eventQueueSize)
 	return nil
 }
 
-func (*peerLogPlugin) Start(node *core.IpfsNode) error {
+func (pl *peerLogPlugin) collectEvents(node *core.IpfsNode) {
+	go func() {
```
willscott, line 76: why not call `go collectEvents()` and skip the inner anonymous function?

plugin/plugins/peerlog/peerlog.go
```diff
@@ -41,60 +67,123 @@ func (*peerLogPlugin) Version() string {
...
-					"peer", e.Peer.Pretty(),
-					"agent", agent,
-					"protocols", protocols,
-				)
+				pl.emit(eventIdentify, e.Peer)
 			}
 		}
 	}()
+
+	go pl.collectEvents(node)
```
willscott, line 185: a go routine is spun up on this function here already.

willscott:
odd that this is the place where logging gets backed up. it doesn't seem like _that_ much more than what other parts of the system are logging.

plugin/plugins/peerlog/peerlog.go
Stebalien: Oops. Thanks!

Stebalien:
I believe the real issue is writing to stdout. We _usually_ only enable error logging which is very low-throughput.

PR #7251, opened 2020-04-29
Choose Key type at initialization

willscott:
* [ ] should be rebased over a released go-ipfs-config version
* [ ] Add a command for 'rolling over' identity

### DIFF
--- a/cmd/ipfs/init.go
+++ b/cmd/ipfs/init.go
@@ -19,13 +19,15 @@ import (
 	cmds "github.com/ipfs/go-ipfs-cmds"
 	config "github.com/ipfs/go-ipfs-config"
 	files "github.com/ipfs/go-ipfs-files"
+	options "github.com/ipfs/interface-go-ipfs-core/options"
 )
 
 const (
-	nBitsForKeypairDefault = 2048
-	bitsOptionName         = "bits"
-	emptyRepoOptionName    = "empty-repo"
-	profileOptionName      = "profile"
+	algorithmDefault    = options.RSAKey
+	algorithmOptionName = "algorithm"
+	bitsOptionName      = "bits"
+	emptyRepoOptionName = "empty-repo"
+	profileOptionName   = "profile"
 )
 
 var errRepoExists = errors.New(`ipfs configuration file already exists!
@@ -54,7 +56,8 @@ environment variable:
 		cmds.FileArg("default-config", false, false, "Initialize with the given configuration.").EnableStdin(),
 	},
 	Options: []cmds.Option{
-		cmds.IntOption(bitsOptionName, "b", "Number of bits to use in the generated RSA private key.").WithDefault(nBitsForKeypairDefault),
+		cmds.StringOption(algorithmOptionName, "a", "Cryptographic algorithm to use for key generation.").WithDefault(algorithmDefault),
+		cmds.IntOption(bitsOptionName, "b", "Number of bits to use in the generated RSA private key."),
 		cmds.BoolOption(emptyRepoOptionName, "e", "Don't add and pin help files to the local storage."),
 		cmds.StringOption(profileOptionName, "p", "Apply profile settings to config. Multiple profiles can be separated by ','"),
 
@@ -82,7 +85,8 @@ environment variable:
 	Run: func(req *cmds.Request, res cmds.ResponseEmitter, env cmds.Environment) error {
 		cctx := env.(*oldcmds.Context)
 		empty, _ := req.Options[emptyRepoOptionName].(bool)
-		nBitsForKeypair, _ := req.Options[bitsOptionName].(int)
+		algorithm, _ := req.Options[algorithmOptionName].(string)
+		nBitsForKeypair, nBitsGiven := req.Options[bitsOptionName].(int)
 
 		var conf *config.Config
 
@@ -106,8 +110,24 @@ environment variable:
 			}
 		}
 
+		var err error
+		var identity config.Identity
+		if nBitsGiven {
+			identity, err = config.CreateIdentity(os.Stdout, []options.KeyGenerateOption{
+				options.Key.Size(nBitsForKeypair),
+				options.Key.Type(algorithm),
+			})
+		} else {
+			identity, err = config.CreateIdentity(os.Stdout, []options.KeyGenerateOption{
+				options.Key.Type(algorithm),
+			})
+		}
+		if err != nil {
+			return err
+		}
+
 		profiles, _ := req.Options[profileOptionName].(string)
-		return doInit(os.Stdout, cctx.ConfigRoot, empty, nBitsForKeypair, profiles, conf)
+		return doInit(os.Stdout, cctx.ConfigRoot, empty, &identity, profiles, conf)
 	},
 }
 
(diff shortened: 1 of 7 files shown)


test/dependencies/go.mod
```diff
@@ -17,13 +17,13 @@ require (
...
 	github.com/ipfs/iptb v1.4.0
-	github.com/ipfs/iptb-plugins v0.2.2
+	github.com/ipfs/iptb-plugins v0.2.3-0.20200429202754-744595369b71
 	github.com/ipld/go-ipld-prime v0.0.2-0.20191108012745-28a82f04c785
 	github.com/jbenet/go-random v0.0.0-20190219211222-123a90aedc0c
 	github.com/jbenet/go-random-files v0.0.0-20190219210431-31b3f20ebded
 	github.com/libp2p/go-libp2p v0.5.2
-	github.com/libp2p/go-libp2p-core v0.3.1
-	github.com/multiformats/go-multiaddr v0.2.0
+	github.com/libp2p/go-libp2p-core v0.5.2
```
Stebalien, line 25: I'd upgrade all deps here:

```go
go get $(go list -json -m all | jq -r 'select(.Indirect != true and .Main != true) | "\(.Path)@latest"')
go mod tidy
```

willscott: leads to
```
github.com/ipfs/go-ipfs/test/dependencies/graphsync-get imports
	github.com/ipld/go-ipld-prime/impl/free: module github.com/ipld/go-ipld-prime@latest found (v0.0.3), but does not contain package github.com/ipld/go-ipld-prime/impl/free
```

should i cut a release of iptb-plugins? otherwise latest won't pick up the master commit i think

Stebalien: Ah, you'll probably need to downgrade go-ipld-prime. But yeah, feel free to cut a release of iptb-plugins.

aschmahmann:
This PR seems to be blocked for now by js-ipfs support for Ed25519 IPNS Keys (e.g. https://github.com/libp2p/js-libp2p-crypto/issues/145) as we'd prefer not to roll the defaults to something that'll be problematic for js-ipfs. This shouldn't be blocked for too long, but if it is we can revisit.

cmd/ipfs/init.go
```diff
@@ -19,10 +19,13 @@ import (
 	cmds "github.com/ipfs/go-ipfs-cmds"
 	config "github.com/ipfs/go-ipfs-config"
 	files "github.com/ipfs/go-ipfs-files"
+	options "github.com/ipfs/interface-go-ipfs-core/options"
 )
 
 const (
 	nBitsForKeypairDefault = 2048
+	algorithmDefault       = options.Ed25519Key
```
aschmahmann, line 27: Since we're going to separate out the default changes from the initialization the default should be RSA here.

cmd/ipfs/init.go
```diff
@@ -106,8 +111,16 @@ environment variable:
 			}
 		}
 
+		identity, err := config.CreateIdentity(os.Stdout, []options.KeyGenerateOption{
+			options.Key.Size(nBitsForKeypair),
+			options.Key.Type(algorithm),
+		})
```
aschmahmann, line 117: As we discussed we should get an error if you try to pass bits and an algorithm that doesn't accept/need bits (e.g. Ed25519). We should probably do this in `config.CreateIdentity` instead of here though.

Note: If we're going to change this then do you think we should allow bits=256 for Ed25519 or error? If setting bits == error then the sharness tests below may need to change, but we can do that in the PR that bubbles up the config changes.

go.mod
```diff
@@ -56,6 +56,7 @@ require (
 	github.com/ipfs/go-verifcid v0.0.1
 	github.com/ipfs/interface-go-ipfs-core v0.3.0
 	github.com/ipld/go-car v0.1.0
+	github.com/jbenet/go-fuse-version v0.0.0-20160322195114-6d4c97bcf253 // indirect
```
aschmahmann, line 59: this (and the go.sum additions) look extraneous

test/sharness/t0020-init.sh
```diff
@@ -81,9 +81,9 @@ test_expect_success "ipfs peer id looks good" '
 
 test_expect_success "ipfs init output looks good" '
   STARTFILE="ipfs cat /ipfs/$HASH_WELCOME_DOCS/readme" &&
-  echo "initializing IPFS node at $IPFS_PATH" >expected &&
-  echo "generating $BITS-bit RSA keypair...done" >>expected &&
+  echo "generating ED25519 keypair...done" >expected &&
   echo "peer identity: $PEERID" >>expected &&
+  echo "initializing IPFS node at $IPFS_PATH" >>expected &&
```
aschmahmann, line 86: The init tests don't take any flags which is why these tests are going to fail now (we switched back to RSA defaults without changing the tests).

We'll probably need to do a copy-paste here so we can handle Ed25519, RSA, and none. What do you think?

cmd/ipfs/init.go
petar: Maybe not allowing to set the bits is better. Otherwise, we would have to require setting them (to be symmetric with RSA usage), and this would break existing users of elliptic keys.

petar: Are you preferring the changes to `config.CreateIdentity` to be in a separate PR?

aschmahmann: It has to be it's in a different repo (go-ipfs-config). However, if you want to do the change there and bubble it up in this PR that's fine too, otherwise we can have a separate bubble PR

test/sharness/t0020-init.sh
petar: See if the newly pushed code looks flexible enough.

cmd/ipfs/init.go
```diff
@@ -24,7 +24,7 @@ import (
 
 const (
 	nBitsForKeypairDefault = 2048
```
aschmahmann, line 26: This const is now unused right?

test/sharness/t0020-init.sh
```diff
@@ -50,52 +50,87 @@ test_expect_success "ipfs cat no repo message looks good" '
...
-'
+# $1 must be one of 'rsa', 'ed25519' or '' (for default key algorithm).
+test_ipfs_init_flags() {
+        TEST_ALG=$1
+
+        # test that init succeeds
+        test_expect_success "ipfs init succeeds" '
+        export IPFS_PATH="$(pwd)/.ipfs" &&
+        echo "IPFS_PATH: \"$IPFS_PATH\"" &&
+        BITS="2048" &&
```
aschmahmann, line 61: I'd rename to RSA_BITS while we're here

test/sharness/t0020-init.sh
```diff
@@ -108,10 +143,16 @@ test_expect_success "ipfs peer id looks good" '
 '
 
 test_expect_success "'ipfs init --empty-repo' output looks good" '
```
aschmahmann, line 145: This is fine, but seems like we might as well put it into the function too.

cmd/ipfs/init.go
petar: It's used in on more place: https://github.com/ipfs/go-ipfs/pull/7251/files#diff-54f2c4a5481de30b8ccb59ad5b6588b3R251
Thoughts?

aschmahmann: I don't think it's really worth keeping.

We can already do `ipfs init` without any bits flag and things still work since the size will default to the one used in config.CreateIdentity() for the RSA Type. Also, it'll end up actually getting used in zero places once we switch to Ed25519 keys by default since `ipfs daemon --init` will use the defaults, which won't include the RSA bits.

PR #7256, opened 2020-04-29
fix: do not use hard coded IPNS Publish maximum timeout duration

### LINKED ISSUE #7244: IPNS: context deadline exceeded
I was excited to read that 0.5 speeds up IPNS. So I wanted to give it a spin. Unfortunately I get an error now:

```
ligi@komputing:~$ ipfs add -Q -r /opt/source-verify/production/source-verify/repository/
QmfXZmzSaEZvEWrJtTjJkmXDV5Dpoxz2diHFPHvnsrexa3
ligi@komputing:~$ ipfs name publish QmfXZmzSaEZvEWrJtTjJkmXDV5Dpoxz2diHFPHvnsrexa3
Error: context deadline exceeded
```
running this version:

```
go-ipfs version: 0.5.0
Repo version: 9
System version: amd64/linux
Golang version: go1.13.10

```

aschmahmann:
fixes #7244

Sometimes IPNS publishes can take longer than a minute. Since DHT lookups now terminate without needing to query a huge portion of the network (even if the lookups take a while until more of the network upgrades to v0.5.0+), we should allow the user's IPNS queries to go on until they're ready to cancel them.

Note: There's still a 30 second timeout on doing a record Get before Publishing, however this doesn't bother me too much since it both doesn't cause an error to be returned and the whole code path is a bit strange in general since it seems to only get called if either:
1) The IPNS key is first being used (in which case we expect the Get to return no info, so we might as well have a timeout)
2) The datastore (not the blocks section, but the rest of the datastore) has been cleaned out/restored from a backup (not really supported)
3) The IPNS key is being added to a new machine (also not supported)

### DIFF
--- a/namesys/publisher.go
+++ b/namesys/publisher.go
@@ -22,7 +22,6 @@ import (
 
 const ipnsPrefix = "/ipns/"
 
-const PublishPutValTimeout = time.Minute
 const DefaultRecordEOL = 24 * time.Hour
 
 // IpnsPublisher is capable of publishing and resolving names to the IPFS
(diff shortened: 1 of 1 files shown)


xmaysonnave:
Today while working on HTTPError and TimeoutError while publishing and IPNS name I realized that even I'm behind a nat/firewall the errors vanished when I deactivated my DNSCrypt (Debian Buster).

PR #7259, opened 2020-04-30
systemd service enhancements + run as system user

RubenKelevra:
enable restart on crash and run as system user 'ipfs'

_add some optional parameters with short help texts_

### DIFF
--- a/misc/systemd/ipfs.service
+++ b/misc/systemd/ipfs.service
@@ -1,9 +1,22 @@
 [Unit]
-Description=IPFS Daemon
+Description=InterPlanetary File System (IPFS) daemon
+After=network.target
 
 [Service]
+
+# enable for 1-1024 port listening
+#AmbientCapabilities=CAP_NET_BIND_SERVICE 
+# enable to specify a custom path see docs/environment-variables.md for further documentations
+#Environment=IPFS_PATH=/custom/ipfs/path
+# enable to specify a higher limit for open files/connections
+#LimitNOFILE=1000000
+
 Type=notify
+User=ipfs
+Group=ipfs
+WorkingDirectory=/var/lib/ipfs
 ExecStart=/usr/bin/ipfs daemon --init --migrate
+Restart=on-failure
 KillSignal=SIGINT
 
 [Install]
--- a/misc/systemd/ipfs-sysusers.conf
+++ b/misc/systemd/ipfs-sysusers.conf
@@ -1,0 +1,3 @@
+u ipfs - "IPFS daemon" /var/lib/ipfs
+g ipfs -
+m ipfs ipfs


misc/systemd/ipfs.service
```diff
@@ -1,9 +1,29 @@
 [Unit]
-Description=IPFS Daemon
+Description=InterPlanetary File System (IPFS) daemon
+After=network.target
```
Stebalien, line 3: Networking isn't strictly speaking necessary. Other services should be able to load content already available on the local node offline.

misc/systemd/ipfs.service
```diff
@@ -1,9 +1,29 @@
...
+# enable to specify a higher swarm FD-Limit
+#Environment=LIBP2P_SWARM_FD_LIMIT=800
+# enable to enforce a memory limit
+#MemoryLimit=5000M
+# enable to specify the preferenced connection muxer
+#Environment=LIBP2P_MUX_PREFS="/mplex/6.7.0 /yamux/1.0.0 /spdy/3.1.0"
+# enable to specify a limit for I/O load while starting
+#StartupIOWeight=10
+# enable to specify a limit for I/O load while running
+#IOWeight=50
```
Stebalien, line 22: I would drop everything but the `AmbientCapabilities` and `LimitNOFILE` lines. The environment variables are better documented in docs/environment-variables.md and the rest of these are just general systemd parameters.

misc/systemd/ipfs.service
```diff
@@ -1,9 +1,29 @@
...
+
 Type=notify
-ExecStart=/usr/bin/ipfs daemon --init --migrate
+ExecStart=/usr/bin/ipfs daemon --init --migrate --enable-gc --enable-pubsub-experiment --enable-namesys-pubsub
```
Stebalien, line 25: Definitely don't include these. If users want to enable experiments and/or GC, they can do so themselves.

misc/systemd/ipfs@.service
```diff
@@ -0,0 +1,31 @@
+[Unit]
```
Stebalien, line 1: Users will likely need to just write their own custom unit in this case. Providing a templated unit doesn't really buy much.

misc/systemd/ipfs.service
```diff
@@ -1,9 +1,29 @@
...
+Restart=on-failure
```
Stebalien, line 26: :+1:

misc/systemd/ipfs.service
RubenKelevra: But the default config does bind sockets, which will fail without a network target already present, right?

misc/systemd/ipfs.service
RubenKelevra: I thought 0.6 was enabling those both (--enable-pubsub-experiment --enable-namesys-pubsub) anyway by default?

And enabling gc by default would follow the principle of least astonishment since there are options in the config file for GC by default.

One would expect that they actually DO anything. But they are completely ignored without adding this parameter.

misc/systemd/ipfs@.service
RubenKelevra: Well, the [archlinux package](https://git.archlinux.org/svntogit/community.git/tree/trunk/ipfs@.service?h=packages/go-ipfs) provides such a ipfs@.service file, because it's much simpler for the users to just add an override config with `systemctl edit ipfs@user` than to create a custom service file.

This way the user could add just the necessary changes, while everything else gets updated by the package.

The idea was to remove the necessity to provide a custom service file by bringing it upstream. I don't think it hurts to provide this file, does it?

misc/systemd/ipfs.service
RubenKelevra: How about changing it to point to the docs/environment-variables.md for the environment variables, but keep one as an example (the path - since this is probably the most common).

The general systemd parameters are just an example of useful resource limitations for such a daemon - sure, I can remove them if you think that's not helpful.

Stebalien: SGTM.

> The general systemd parameters are just an example of useful resource limitations for such a daemon - sure, I can remove them if you think that's not helpful.

I'd leave them out. systemd has a ton of tunables and these aren't all that specific to go-ipfs (except LimitNOFILE and AmbientCapabilities).

misc/systemd/ipfs.service
Stebalien: Yes, but that doesn't mean we should be enabling experiments in the default service file.

> And enabling gc by default would follow the principle of least astonishment since there are options in the config file for GC by default.

If we start enabling GC by default, it will be after a thorough audit of GC and likely a rewrite. The systemd unit should not enable things that are not otherwise enabled by default (except `--init` and `--migrate` because those are important for unattended operation).

misc/systemd/ipfs@.service
Stebalien: Ideally, packages would use a dynamic user or a special `ipfs` user.

If users just want to run an IPFS daemon as a specific user, they should:

1. Enable `user@$uid.service` on start.
2. Enable the `ipfs` unit for that user.

misc/systemd/ipfs.service
Stebalien: Hm, maybe? We usually bind to 0.0.0.0 (no networking required) and 127.0.0.1. I assumed systemd brought up the loopback interface before starting services.

However, I guess it doesn't hurt to enable this for now. My concern is that I'd like early units to be able to pull files out of IPFS if necessary.

RubenKelevra: network.target just means that the network stack is started, not necessarily that there is any network connection possible, this would be network-online.target.

I think setting network.target is a sane default without blocking any requests on boot.

misc/systemd/ipfs.service
RubenKelevra: Alright, I revert that :)

misc/systemd/ipfs@.service
RubenKelevra: I agree that users should use a user-daemon for most situations - this is already documented [in the ArchLinux wiki](https://wiki.archlinux.org/index.php/IPFS#Using_a_service_to_start_the_daemon).

So what's now the sane default? Change the regular `ipfs.service` to run as user `ipfs` and advice the package maintainer that they have to create system-user `ipfs`?

Because running ipfs as root isn't a great default.

Stebalien: > So what's now the sane default? Change the regular ipfs.service to run as user ipfs and advice the package maintainer that they have to create system-user ipfs?

SGTM. But if we do that, we should change the default home to /var/lib/ipfs

RubenKelevra: In the service file as an environment parameter?

RubenKelevra:
@Stebalien is the failed test related to my changes? I don't how it depends on the systemd file - but maybe I overlooked something.

misc/systemd/ipfs@.service
RubenKelevra: I've added a sysusers.conf with the path and set the working directory accordingly.

RubenKelevra:
> This unit is a minimal and conservative starting point, let's avoid adding too much.

Ready for a new review :)

PR #7261, opened 2020-04-30
Updates bash completions to be compatible with zsh

cachance7:
The `compopt` bash builtin is not supported by zsh's bash compatibility
features. By checking for zsh presence and using `complete` instead,
this script becomes suitable for use with both bash and zsh.

### DIFF
--- a/misc/completion/ipfs-completion.bash
+++ b/misc/completion/ipfs-completion.bash
@@ -1,9 +1,9 @@
 _do_comp()
 {
-  if [ -n "$ZSH_VERSION" ]; then
-    complete $@
-  else
+  if [[ $(type compopt) == *"builtin" ]]; then
     compopt $@
+  else
+    complete $@
   fi
 }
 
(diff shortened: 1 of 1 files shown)


misc/completion/ipfs-completion.bash
```diff
@@ -1,9 +1,18 @@
+_do_comp()
+{
+  if [ -n "$ZSH_VERSION" ]; then
+    complete $@
+  else
+    compopt $@
+  fi
```
willscott, line 7:
```suggestion
  if [[ $(type compopt) == *"builtin" ]]; then
    compopt $@
  else
    complete $@
  fi
```

willscott: perhaps do this by seeing which one is available, rather than an env var

cachance7: I tried looking for something similar but couldn't find anything that worked in both bash and zsh. This suggestion does not work in zsh.

cachance7:
@willscott Updated based on your feedback. Thanks!

PR #7262, opened 2020-05-01
Fixes #7252 - Uses gabriel-vasile/mimetype to support additional content types

### LINKED ISSUE #7252: Gateway content type detection : ODS documents served as ZIP (MIME)
#### Description:

It seems `http.DetectContentType` does not work well for these types.

We could switch to https://github.com/gabriel-vasile/mimetype ? (also see https://stackoverflow.com/a/52266327 for more options). Unsure if there is a reason have have sticked with the current detector.

Related #2164 

Reported at https://discuss.ipfs.io/t/bugs-in-0-5-0-regarding-file-names-types/7845

gowthamgts:
This fixes https://github.com/ipfs/go-ipfs/issues/7252 by replacing  `http.DetectContentType()` with `mimetype.Detect()` from https://github.com/gabriel-vasile/mimetype library.

I've tested locally with the ODS files, mp3 files and html files. Please let me know if there's anything further that needs to be done!

### DIFF
--- a/core/corehttp/gateway_handler.go
+++ b/core/corehttp/gateway_handler.go
@@ -3,7 +3,6 @@ package corehttp
 import (
 	"context"
 	"fmt"
-	"github.com/gabriel-vasile/mimetype"
 	"io"
 	"mime"
 	"net/http"
@@ -16,6 +15,7 @@ import (
 	"time"
 
 	humanize "github.com/dustin/go-humanize"
+	"github.com/gabriel-vasile/mimetype"
 	"github.com/ipfs/go-cid"
 	files "github.com/ipfs/go-ipfs-files"
 	dag "github.com/ipfs/go-merkledag"
--- a/core/corehttp/gateway_handler.go
+++ b/core/corehttp/gateway_handler.go
@@ -377,12 +377,16 @@ func (i *gatewayHandler) serveFile(w http.ResponseWriter, req *http.Request, nam
 	} else {
 		ctype = mime.TypeByExtension(gopath.Ext(name))
 		if ctype == "" {
-			buf := make([]byte, 512)
-			n, _ := io.ReadFull(content, buf[:])
 			// uses https://github.com/gabriel-vasile/mimetype library to determine the content type.
 			// Fixes https://github.com/ipfs/go-ipfs/issues/7252
-			ctype = mimetype.Detect(buf[:n]).String()
-			_, err := content.Seek(0, io.SeekStart)
+			mimeType, err := mimetype.DetectReader(content)
+			if err != nil {
+				http.Error(w, "cannot detect content-type", http.StatusInternalServerError)
+				return
+			}
+
+			ctype = mimeType.String()
+			_, err = content.Seek(0, io.SeekStart)
 			if err != nil {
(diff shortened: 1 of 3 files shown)


core/corehttp/gateway_handler.go
```diff
@@ -378,7 +379,9 @@ func (i *gatewayHandler) serveFile(w http.ResponseWriter, req *http.Request, nam
 		if ctype == "" {
 			buf := make([]byte, 512)
 			n, _ := io.ReadFull(content, buf[:])
-			ctype = http.DetectContentType(buf[:n])
+			// uses https://github.com/gabriel-vasile/mimetype library to determine the content type.
+			// Fixes https://github.com/ipfs/go-ipfs/issues/7252
+			ctype = mimetype.Detect(buf[:n]).String()
```
hsanjuan, line 384: I think we could actually use `DetectReader(content)` directly. This will read at most 3072 bytes. It is more than the current 512 but worth it if that gives you an accurate type (and also means it will read less for things that can be detected with less, which are hopefully most).

gowthamgts: okay. I assume we don't have to seek to the start if we're using `DetectReader(content)`?

gowthamgts: @hsanjuan - fixed via 9201b1d.

core/corehttp/gateway_handler.go
```diff
@@ -376,10 +377,16 @@ func (i *gatewayHandler) serveFile(w http.ResponseWriter, req *http.Request, nam
...
 		if ctype == "" {
-			buf := make([]byte, 512)
-			n, _ := io.ReadFull(content, buf[:])
-			ctype = http.DetectContentType(buf[:n])
-			_, err := content.Seek(0, io.SeekStart)
+			// uses https://github.com/gabriel-vasile/mimetype library to determine the content type.
+			// Fixes https://github.com/ipfs/go-ipfs/issues/7252
+			mimeType, err := mimetype.DetectReader(content)
+			if err != nil {
+				http.Error(w, "cannot detect content-type", http.StatusInternalServerError)
```
hsanjuan, line 384: Can you append the err.Error() to the error message?

gowthamgts: @hsanjuan - fixed via 214d29e

core/corehttp/gateway_handler.go
```diff
@@ -3,6 +3,7 @@ package corehttp
 import (
 	"context"
 	"fmt"
+	"github.com/gabriel-vasile/mimetype"
```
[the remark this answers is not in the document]
gowthamgts: fixed via e93f20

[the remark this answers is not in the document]
gowthamgts: reverse merged from master as well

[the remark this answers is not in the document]
Stebalien: Please rebase instead of reverse-merging.

Stebalien:
Hm. Apparently I didn't push my rebase to the right origin...

Stebalien:
@gowthamgts thanks for the fix!

PR #7265, opened 2020-05-01
systemd: add helptext

RubenKelevra:
General help text added to instruct new user to make permanent changes to the service files

<strike>Add a longer timeout for startup and shutdown</strike> _Removed after discussions_

### DIFF
--- a/misc/systemd/ipfs.service
+++ b/misc/systemd/ipfs.service
@@ -1,3 +1,17 @@
+# This file will be overwritten on package upgrades, avoid customizations here.
+#
+# To make persistant changes, create file in 
+# "/etc/systemd/system/ipfs.service.d/overwrite.conf" with 
+# `systemctl edit ipfs.service`. This file will be parsed after this 
+# file has been parsed.
+#
+# To overwrite a variable, like ExecStart you have to specify it once
+# blank and a second time with a new value, like:
+# ExecStart=
+# ExecStart=/usr/bin/ipfs daemon --flag1 --flag2
+#
+# For more info about custom unit files see systemd.unit(5).
+
 [Unit]
 Description=InterPlanetary File System (IPFS) daemon
 After=network.target


misc/systemd/ipfs.service
```diff
@@ -14,8 +28,11 @@ After=network.target
 Type=notify
 User=ipfs
 Group=ipfs
+TimeoutStartSec=15min
+TimeoutStopSec=15min
+TimeoutAbortSec=15min
```
hsanjuan, line 33: This might be too much. If something has not started in 2 minutes it's already a clear issue. Hiding it by increasing timeouts does not really help. I'm not sure that you would not have had the same issue as in #7265 , just 15 mins later...

misc/systemd/ipfs.service
```diff
@@ -14,8 +28,11 @@ After=network.target
...
 WorkingDirectory=/var/lib/ipfs
-ExecStart=/usr/bin/ipfs daemon --init --migrate
```
hsanjuan, line 18: Perhaps we need a separate service to run migrations? I agree that running migrations by default might get us to ugly places...

misc/systemd/ipfs.service
hsanjuan: What do other people think?

RubenKelevra: Well, I have one VM with a badger storage on a cheap provider which offers "storage" VMs with very little CPU power.

The startup time after a crash on badger is on this machine between 5 and 10 minutes.

Previously there was no limit at all for starting up ipfs, so this issue never appeared, but since we changed from simple to notify systemd now expects an "I'm ready" signal back within 1 minute and 30 seconds.

If you run IPFS on an SSD storage with a fast CPU and a small repo one could lower this value with an override.conf, but this is a safe and sane default IMHO.

misc/systemd/ipfs.service
RubenKelevra: I'm looking into options, but usually there's a script which is run by the package manager which either converts the database or returns immediately if there's no need to upgrade.

Here's [such a script](https://git.archlinux.org/svntogit/packages.git/tree/trunk/mariadb.install?h=packages/mariadb) as an example for an Arch package (in this case the service can start with the old database format):

RubenKelevra: My issue with a separate service is, that it would be potentially called on the next boot, after an update.

@Stebalien [expressed](https://github.com/ipfs/go-ipfs/pull/7259#discussion_r418184621) that he likes the idea that IPFS can serve files on boot for booting services. So if a service would convert the database while the machine is booting, there's the potential that the boot of the machine will halt until the database conversion is completed.

I think the right way is to either hint the user to do a conversion, with a message from the package manager or have the package manager call a script to convert the database while applying the update - depending on the philosophy of the distribution.

Stebalien: The right solution here is to:

1. Pass the `--migrate` flag.
2. Send `EXTEND_TIMEOUT_USEC=...` to the systemd notify socket repeatedly on start when migrating.

Stebalien: Issue: https://github.com/ipfs/go-ipfs/issues/7269

RubenKelevra: This sounds like a much better solution than mine. Thanks for looking into this!

I would still like to remove the --migration flag until #7269 is implemented (which can then add --migrate) back to the service. This way the service file can be used without risking that the conversation might be interrupted by systemd.

misc/systemd/ipfs.service
RubenKelevra: > IPFS should start pretty much instantly. We should not set our timeouts based on unusual setups.

Well, this VM is has one virtual core of a Intel Xeon CPU E5-2609 v3 @ 1.90GHz, and 2 GB of memory and no other background tasks. OS is CentOS7.

I've compiled the current master of IPFS and used the current service file and now I'm unable to start IPFS on this machine.

Not entirely sure what you mean with unusual setup, though. It's just not a high-performance server with SSDs and multiple cores, but I think IPFS should be able to start on smaller machines without having to modify the service file by hand.

```
Mai 04 11:32:36 vidar.pacman.store ipfs[3130]: Initializing daemon...
Mai 04 11:32:36 vidar.pacman.store ipfs[3130]: go-ipfs version: 0.6.0-dev
Mai 04 11:32:36 vidar.pacman.store ipfs[3130]: Repo version: 9
Mai 04 11:32:36 vidar.pacman.store ipfs[3130]: System version: amd64/linux
Mai 04 11:32:36 vidar.pacman.store ipfs[3130]: Golang version: go1.14.2
Mai 04 11:34:06 vidar.pacman.store systemd[1]: ipfs@ipfs.service start operation timed out. Terminating.
Mai 04 11:35:36 vidar.pacman.store systemd[1]: ipfs@ipfs.service stop-sigterm timed out. Killing.
Mai 04 11:35:36 vidar.pacman.store systemd[1]: ipfs@ipfs.service: main process exited, code=killed, status=9/KILL
Mai 04 11:35:36 vidar.pacman.store systemd[1]: Failed to start InterPlanetary File System (IPFS) daemon.
Mai 04 11:35:36 vidar.pacman.store systemd[1]: Unit ipfs@ipfs.service entered failed state.
Mai 04 11:35:36 vidar.pacman.store systemd[1]: ipfs@ipfs.service failed.
Mai 04 11:35:36 vidar.pacman.store systemd[1]: ipfs@ipfs.service holdoff time over, scheduling restart.
Mai 04 11:35:36 vidar.pacman.store systemd[1]: Stopped InterPlanetary File System (IPFS) daemon.
Mai 04 11:35:36 vidar.pacman.store systemd[1]: Starting InterPlanetary File System (IPFS) daemon...
Mai 04 11:35:36 vidar.pacman.store ipfs[3196]: Initializing daemon...
Mai 04 11:35:36 vidar.pacman.store ipfs[3196]: go-ipfs version: 0.6.0-dev
Mai 04 11:35:36 vidar.pacman.store ipfs[3196]: Repo version: 9
Mai 04 11:35:36 vidar.pacman.store ipfs[3196]: System version: amd64/linux
Mai 04 11:35:36 vidar.pacman.store ipfs[3196]: Golang version: go1.14.2
Mai 04 11:37:06 vidar.pacman.store systemd[1]: ipfs@ipfs.service start operation timed out. Terminating.
```

RubenKelevra: After several failed attempts the database opening seems to accelerated by caching (maybe on the VM host) - so it's just an IO limitation:

```
Mai 04 12:01:45 vidar.pacman.store systemd[1]: Starting InterPlanetary File System (IPFS) daemon..
Mai 04 12:01:45 vidar.pacman.store ipfs[3294]: Initializing daemon...
Mai 04 12:01:45 vidar.pacman.store ipfs[3294]: go-ipfs version: 0.6.0-dev
Mai 04 12:01:45 vidar.pacman.store ipfs[3294]: Repo version: 9
Mai 04 12:01:45 vidar.pacman.store ipfs[3294]: System version: amd64/linux
Mai 04 12:01:45 vidar.pacman.store ipfs[3294]: Golang version: go1.14.2
Mai 04 12:01:54 vidar.pacman.store ipfs[3294]: Swarm listening on /ip4/10.176.233.122/tcp/443
Mai 04 12:01:54 vidar.pacman.store ipfs[3294]: Swarm listening on /ip4/127.0.0.1/tcp/443
Mai 04 12:01:54 vidar.pacman.store ipfs[3294]: Swarm listening on /ip4/127.0.0.1/tcp/443
Mai 04 12:01:54 vidar.pacman.store ipfs[3294]: Swarm listening on /ip4/94.176.233.122/tcp/443
Mai 04 12:01:54 vidar.pacman.store ipfs[3294]: Swarm listening on /ip6/2a02:7b40:5eb0:e97a::1/tcp/443
Mai 04 12:01:54 vidar.pacman.store systemd[1]: Started InterPlanetary File System (IPFS) daemon.
Mai 04 12:01:54 vidar.pacman.store ipfs[3294]: Swarm listening on /ip6/::1/tcp/443
Mai 04 12:01:54 vidar.pacman.store ipfs[3294]: Swarm announcing /ip4/94.176.233.122/tcp/443
Mai 04 12:01:54 vidar.pacman.store ipfs[3294]: Swarm announcing /ip6/2a02:7b40:5eb0:e97a::1/tcp/443
Mai 04 12:01:54 vidar.pacman.store ipfs[3294]: API server listening on /ip4/127.0.0.1/tcp/5001
Mai 04 12:01:54 vidar.pacman.store ipfs[3294]: WebUI: http://127.0.0.1:5001/webui
Mai 04 12:01:54 vidar.pacman.store ipfs[3294]: Gateway (readonly) server listening on /ip4/127.0.0.1/tcp/8080
Mai 04 12:01:54 vidar.pacman.store ipfs[3294]: Daemon is ready
```

misc/systemd/ipfs.service
Stebalien: Either way will break things:

* With the --migrate flag and with no extended timeout, we'll fail half-way through the migration if it happens to be a long one (the last few have been short).
* Without the --migrate flag, it'll _always_ fail to start.

We can make sure to fix #7269 before releasing a long-running migration.

Given this, I strongly disagree. Let's leave the --migrate flag.

misc/systemd/ipfs.service
Stebalien: * Where is the underlying VM data stored (what filesystem? remote?)?
* If this is a _common_ problem, we need to fix it and make go-ipfs start faster.
* How large is your repo?

My guess is that your provider provides _terrible_ read latency but likely high read throughput. It should never take go-ipfs more than several seconds to startup.

RubenKelevra: > * Where is the underlying VM data stored (what filesystem? remote?)?

It isn't advertised as a remote filesystem, my best guess is that it's a fibre channel box connected to the host.

The only other thing the provider tells me, is that it's limited to 200 IOPS.

> My guess is that your provider provides _terrible_ read latency but likely high read throughput. It should never take go-ipfs more than several seconds to startup.

I tried to measure the latency, and it isn't really looking bad:

```
[ipfs@vidar badgerds]$ time dd if=000000.vlog of=/dev/zero bs=4K count=1
1+0 records in
1+0 records out
4096 bytes (4,1 kB) copied, 0,0225068 s, 182 kB/s

real    0m0.085s
user    0m0.003s
sys     0m0.000s
[ipfs@vidar badgerds]$ time dd if=000001.vlog of=/dev/zero bs=4K count=1
1+0 records in
1+0 records out
4096 bytes (4,1 kB) copied, 0,0204767 s, 200 kB/s

real    0m0.023s
user    0m0.000s
sys     0m0.002s
[ipfs@vidar badgerds]$ time dd if=000002.vlog of=/dev/zero bs=4K count=1
1+0 records in
1+0 records out
4096 bytes (4,1 kB) copied, 0,0491106 s, 83,4 kB/s

real    0m0.051s
user    0m0.001s
sys     0m0.001s
[ipfs@vidar badgerds]$ time dd if=000003.vlog of=/dev/zero bs=4K count=1
1+0 records in
1+0 records out
4096 bytes (4,1 kB) copied, 0,00929099 s, 441 kB/s

real    0m0.012s
user    0m0.001s
sys     0m0.001s
```

I just think the database requires many small read requests, which just hit the IOPS limiter.

> * If this is a _common_ problem, we need to fix it and make go-ipfs start faster.

I think many people haven't hit this issue yet since the service file wasn't on "notify" and the badger datastore wasn't stable. I'm just early to run into this issue.

I think ipfs should just give the badger datastore all the time it needs to open, and don't enforce any time limit. If we think of Rasberry Pie nodes, which need to be able to read/write from a slow SD-Card on one end of the spectrum, it's probably best to just inform systemd that we're still alive and still opening the database - instead of getting a kill signal.

> * How large is your repo?

```
[ipfs@vidar ~]$ ipfs repo stat --human
NumObjects: 677400
RepoSize:   154 GB
StorageMax: 1.0 TB
RepoPath:   /home/ipfs/.ipfs
Version:    fs-repo@9
```

misc/systemd/ipfs.service
RubenKelevra: fixed

misc/systemd/ipfs.service
RubenKelevra: Let's track this down in a ticket:

https://github.com/ipfs/go-ipfs/issues/7273

RubenKelevra:
@Stebalien @hsanjuan 

ready for a new review.

RubenKelevra:
@Stebalien @hsanjuan ready for a new review :)

Stebalien:
Let's try to figure out _why_ IPFS is taking so long to stop for you, not just extend the stop timeout.

RubenKelevra:
> Let's try to figure out _why_ IPFS is taking so long to stop for you, not just extend the stop timeout.

Sure.

I remove those changes from this PR as well since I like to push some other changed on this file in the next days and I want to avoid merge conflicts. :)

Issue #7160, opened 2020-04-14
GC should skip over raw blocks


Stebalien:
Currently, when GC traverses the pin set, it loads all blocks, even raw blocks. This means:

1. We need to perform unnecessary work (fetching raw blocks from disk).
2. When using the filestore, if the backing data disappears, the associated leaf blocks will return errors on read. At the moment, this will completely _break_ GC.

We should simply avoid loading raw blocks when garbage collecting (or at _least_ just check for existence instead of reading the block). This will:

1. Improve GC performance significantly when working with large amounts of data with raw leaves.
2. Make GC continue to work even if the filestore gets corrupted.

RubenKelevra:
How about adding three flags to the GC operation to determine how the GC should handle filestore and urlstore? 

1) **--filestore[/urlstore]=ignore**

    Just skip all filestore/URL store entries completely and don't check if they are accessible.

2) **--filestore[/urlstore]=keep**

    Keep all entries which are still accessible, regardless if they are pinned or not.

    Warn about non-accessible files/urls.

3) **--filestore[/urlstore]=drop** - _default_

    Drop entries from IPFS if they are not pinned/in the MFS

hsanjuan:
Is this work meant to be scheduled soon or would we rather wait to doing a larger GC revamp (possibly after raw-multihashes and pinning system changes)? I have tentatively tagged as P3.

Stebalien:
P3 seems right given that it won't help with _performance_ in many cases. My main interest is actually allowing GC to progress even when filestore blocks are missing.

balajijinnah:
The GC should avoid fetching raw blocks if the given CID is codec is of raw. Is this right? 

If it's right. I would like to raise a PR for this.

Stebalien:
Exactly. PRs welcome!

Stebalien:
This idiot (@Stebalien <-- this one) made bad assumptions and filed an issue without actually verifying the issue. Apparently, we don't load raw blocks when garbage collecting.

Stebalien:
@balajijinnah thank you for implementing this, and again, I apologize for not verifying the issue first.

Issue #7269, opened 2020-05-04
Long running migrations can cause systemd to terminate daemon on start


Stebalien:
#### Version information:

> 0.5.0

#### Description:

Now that go-ipfs supports systemd's "notification" system, we need to tell systemd to repeatedly extend time startup timeout while performing repo migrations. Otherwise, systemd may kill the daemon thinking it timed out on startup.

We can do this by repeatedly sending `EXTEND_TIMEOUT_USEC=...` to systemd's notification socket using github.com/coreos/go-systemd/v22/daemon. See `cmd/ipfs/daemon_linux.go` for how we interact with systemd's notification service.

RubenKelevra:
Can we notify systemd the same way while we open the database on startup (which is basically the reason for the slow startup documented [here](https://github.com/ipfs/go-ipfs/pull/7265#discussion_r418645508) and on shutdown when we for example run data compaction jobs, flush data to disk, or clean up sockets etc.?

Stebalien:
Yep. We can wrap all of https://github.com/ipfs/go-ipfs/blob/538bff085abb18a16fbf14386358c00c7078c222/cmd/ipfs/daemon.go#L257-L290.

Stebalien:
However, we _don't_ do database compactions on stop (explicitly disabled for this reason) so we shouldn't delay there. If IPFS gets stuck on shutdown, we should just die and cleanup when we next restart.

RubenKelevra:
I've seen ipfs hit the one minute 30-second limit dozens of times on different machines, without badger db. My guess is, that it's a TCP cleanup issue.

I think it's still nicer to clean up everything cleanly and as long as we make progress we notify systemd.

Maybe add a hard limit, after which we stop notifying on shutdown, to make sure we don't hang indefinitely.

Stebalien:
Odd. We shouldn't be spending any time cleaning up TCP connections or anything like that, unless we have a bug somewhere. Can you reproduce this?

RubenKelevra:
> Odd. We shouldn't be spending any time cleaning up TCP connections or anything like that, unless we have a bug somewhere. Can you reproduce this?

Well, I converted all my nodes to badgerds with the 0.5.0 release.

I need to convert one back to see how it can be reproduced.

RubenKelevra:
@Stebalien 

> I've seen ipfs hit the one minute 30-second limit dozens of times on different machines, without badger db. My guess is, that it's a TCP cleanup issue.

[...]

Issue #7227, opened 2020-04-27
Atomically create API file


Stebalien:
We should:

1. Atomically create the `IPFS_HOME/api` file to ensure we never leave an empty/corrupt one behind. The relevant code is in the SetAPIAddr function in repo/fsrepo/fsrepo.go. We should create a temporary file, write to the temporary file, then atomically rename to "api".
2. If we read the "api" file and it's empty, we should ignore it.

Reported on the forums: https://discuss.ipfs.io/t/error-multiaddr-must-start-with-a/7824

PR #7272, opened 2020-05-04
skip traversing raw blocks while garbage collecting

balajijinnah:
Fix for #7160

Signed-off-by: Tiger <rbalajis25@gmail.com>

### DIFF
--- a/gc/gc.go
+++ b/gc/gc.go
@@ -173,6 +173,10 @@ func ColoredSet(ctx context.Context, pn pin.Pinner, ng ipld.NodeGetter, bestEffo
 	errors := false
 	gcs := cid.NewSet()
 	getLinks := func(ctx context.Context, CID cid.Cid) ([]*ipld.Link, error) {
+		if CID.Type() == cid.Raw {
+			return nil, nil
+		}
+
 		links, err := ipld.GetLinks(ctx, ng, CID)
 		if err != nil {
 			errors = true
@@ -182,16 +186,7 @@ func ColoredSet(ctx context.Context, pn pin.Pinner, ng ipld.NodeGetter, bestEffo
 				return nil, ctx.Err()
 			}
 		}
-
-		filteredLinks := []*ipld.Link{}
-		for _, link := range links {
-			// Skip raw links, since it doesn't contain any leaf links to traverse.
-			if link.Cid.Type() == cid.Raw {
-				continue
-			}
-			filteredLinks = append(filteredLinks, link)
-		}
-		return filteredLinks, nil
+		return links, nil
 	}
 	rkeys, err := pn.RecursiveKeys(ctx)
 	if err != nil {
@@ -208,6 +203,10 @@ func ColoredSet(ctx context.Context, pn pin.Pinner, ng ipld.NodeGetter, bestEffo
 	}
 
 	bestEffortGetLinks := func(ctx context.Context, CID cid.Cid) ([]*ipld.Link, error) {
+		if CID.Type() == cid.Raw {
+			return nil, nil
+		}
+
 		links, err := ipld.GetLinks(ctx, ng, CID)
 		if err != nil && err != ipld.ErrNotFound {
 			errors = true
@@ -217,16 +216,7 @@ func ColoredSet(ctx context.Context, pn pin.Pinner, ng ipld.NodeGetter, bestEffo
 				return nil, ctx.Err()
 			}
 		}
-
-		filteredLinks := []*ipld.Link{}
-		for _, link := range links {
-			// Skip raw links, since it doesn't contain any leaf links to traverse.
-			if link.Cid.Type() == cid.Raw {
-				continue
-			}
-			filteredLinks = append(filteredLinks, link)
-		}
-		return filteredLinks, nil
+		return links, nil
 	}
 	err = Descendants(ctx, bestEffortGetLinks, gcs, bestEffortRoots)
 	if err != nil {
--- a/gc/gc.go
+++ b/gc/gc.go
@@ -172,17 +172,26 @@ func ColoredSet(ctx context.Context, pn pin.Pinner, ng ipld.NodeGetter, bestEffo
 	// disk backed to conserve memory.
 	errors := false
 	gcs := cid.NewSet()
-	getLinks := func(ctx context.Context, cid cid.Cid) ([]*ipld.Link, error) {
-		links, err := ipld.GetLinks(ctx, ng, cid)
+	getLinks := func(ctx context.Context, CID cid.Cid) ([]*ipld.Link, error) {
+		links, err := ipld.GetLinks(ctx, ng, CID)
 		if err != nil {
 			errors = true
 			select {
-			case output <- Result{Error: &CannotFetchLinksError{cid, err}}:
+			case output <- Result{Error: &CannotFetchLinksError{CID, err}}:
 			case <-ctx.Done():
 				return nil, ctx.Err()
 			}
 		}
-		return links, nil
+
+		filteredLinks := []*ipld.Link{}
+		for _, link := range links {
+			// Skip raw links, since it doesn't contain any leaf links to traverse.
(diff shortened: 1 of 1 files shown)


gc/gc.go
```diff
@@ -172,17 +172,26 @@ func ColoredSet(ctx context.Context, pn pin.Pinner, ng ipld.NodeGetter, bestEffo
 	// disk backed to conserve memory.
 	errors := false
 	gcs := cid.NewSet()
-	getLinks := func(ctx context.Context, cid cid.Cid) ([]*ipld.Link, error) {
-		links, err := ipld.GetLinks(ctx, ng, cid)
+	getLinks := func(ctx context.Context, CID cid.Cid) ([]*ipld.Link, error) {
```
Stebalien, line 175: Please revert the variable name change.

gc/gc.go
```diff
@@ -172,17 +172,26 @@ func ColoredSet(ctx context.Context, pn pin.Pinner, ng ipld.NodeGetter, bestEffo
...
+
+		filteredLinks := []*ipld.Link{}
+		for _, link := range links {
+			// Skip raw links, since it doesn't contain any leaf links to traverse.
+			if link.Cid.Type() == cid.Raw {
+				continue
+			}
+			filteredLinks = append(filteredLinks, link)
+		}
+		return filteredLinks, nil
```
Stebalien, line 194: A simpler (and more reliable) approach would be to filter CIDs of raw blocks at the _top_ of this function. That is, a raw `cid` is passed in, return `nil, nil` (no links, no error).

gc/gc.go
```diff
@@ -198,17 +207,26 @@ func ColoredSet(ctx context.Context, pn pin.Pinner, ng ipld.NodeGetter, bestEffo
...
```
Stebalien, line 229: ditto about moving this check to the top.

gc/gc.go
balajijinnah: If I use cid as variable, variable is shadowing the cid package. 
I'm not able to use cid.Raw

Stebalien: Ah, then I'd use `id, `c` or `key`.  Upper-case is usually reserved for exported variables.

gc/gc.go
```diff
@@ -172,12 +172,16 @@ func ColoredSet(ctx context.Context, pn pin.Pinner, ng ipld.NodeGetter, bestEffo
 	// disk backed to conserve memory.
 	errors := false
 	gcs := cid.NewSet()
-	getLinks := func(ctx context.Context, cid cid.Cid) ([]*ipld.Link, error) {
-		links, err := ipld.GetLinks(ctx, ng, cid)
+	getLinks := func(ctx context.Context, CID cid.Cid) ([]*ipld.Link, error) {
```
Stebalien, line 175:
```suggestion
	getLinks := func(ctx context.Context, link cid.Cid) ([]*ipld.Link, error) {
```

gc/gc.go
```diff
@@ -172,12 +172,16 @@ func ColoredSet(ctx context.Context, pn pin.Pinner, ng ipld.NodeGetter, bestEffo
...
+		if CID.Type() == cid.Raw {
```
Stebalien, line 176:
```suggestion
		if link.Type() == cid.Raw {
```

gc/gc.go
```diff
@@ -172,12 +172,16 @@ func ColoredSet(ctx context.Context, pn pin.Pinner, ng ipld.NodeGetter, bestEffo
...
+			return nil, nil
+		}
+
+		links, err := ipld.GetLinks(ctx, ng, CID)
```
Stebalien, line 180:
```suggestion
		links, err := ipld.GetLinks(ctx, ng, link)
```

gc/gc.go
```diff
@@ -172,12 +172,16 @@ func ColoredSet(ctx context.Context, pn pin.Pinner, ng ipld.NodeGetter, bestEffo
...
 		if err != nil {
 			errors = true
 			select {
-			case output <- Result{Error: &CannotFetchLinksError{cid, err}}:
+			case output <- Result{Error: &CannotFetchLinksError{CID, err}}:
```
Stebalien, line 184:
```suggestion
			case output <- Result{Error: &CannotFetchLinksError{link, err}}:
```

gc/gc.go
```diff
@@ -198,12 +202,16 @@ func ColoredSet(ctx context.Context, pn pin.Pinner, ng ipld.NodeGetter, bestEffo
 		}
 	}
 
-	bestEffortGetLinks := func(ctx context.Context, cid cid.Cid) ([]*ipld.Link, error) {
-		links, err := ipld.GetLinks(ctx, ng, cid)
+	bestEffortGetLinks := func(ctx context.Context, CID cid.Cid) ([]*ipld.Link, error) {
```
Stebalien, line 205:
```suggestion
	bestEffortGetLinks := func(ctx context.Context, link cid.Cid) ([]*ipld.Link, error) {
```

gc/gc.go
```diff
@@ -198,12 +202,16 @@ func ColoredSet(ctx context.Context, pn pin.Pinner, ng ipld.NodeGetter, bestEffo
...
+		if CID.Type() == cid.Raw {
```
Stebalien, line 206:
```suggestion
		if link.Type() == cid.Raw {
```

gc/gc.go
```diff
@@ -198,12 +202,16 @@ func ColoredSet(ctx context.Context, pn pin.Pinner, ng ipld.NodeGetter, bestEffo
...
+			return nil, nil
+		}
+
+		links, err := ipld.GetLinks(ctx, ng, CID)
```
Stebalien, line 210:
```suggestion
		links, err := ipld.GetLinks(ctx, ng, link)
```

gc/gc.go
```diff
@@ -198,12 +202,16 @@ func ColoredSet(ctx context.Context, pn pin.Pinner, ng ipld.NodeGetter, bestEffo
...
 		if err != nil && err != ipld.ErrNotFound {
 			errors = true
 			select {
-			case output <- Result{Error: &CannotFetchLinksError{cid, err}}:
+			case output <- Result{Error: &CannotFetchLinksError{CID, err}}:
```
Stebalien, line 214:
```suggestion
			case output <- Result{Error: &CannotFetchLinksError{link, err}}:
```

gc/gc.go
```diff
@@ -172,12 +172,16 @@ func ColoredSet(ctx context.Context, pn pin.Pinner, ng ipld.NodeGetter, bestEffo
...
 	errors := false
 	gcs := cid.NewSet()
-	getLinks := func(ctx context.Context, cid cid.Cid) ([]*ipld.Link, error) {
-		links, err := ipld.GetLinks(ctx, ng, cid)
+	getLinks := func(ctx context.Context, link cid.Cid) ([]*ipld.Link, error) {
+		if link.Type() == cid.Raw {
+			return nil, nil
+		}
+
+		links, err := ipld.GetLinks(ctx, ng, link)
```
sinkuu, line 180: `GetLinks` seems to already have fast path for raw blocks.

```go
func GetLinks(ctx context.Context, ng NodeGetter, c cid.Cid) ([]*Link, error) {
	if c.Type() == cid.Raw {
		return nil, nil
	}
	...
```

Stebalien: Well... Now really I feel like an idiot.

@balajijinnah I'm so sorry. Apparently, we already _have_ this optimization, I just never thought to actually check.

balajijinnah: Lol, Even I didn't even check it. 

It's a good experience to contribute. Looking to contribute more.

Stebalien:
Well, I'm not going to say no to that! Depending on how familiar you are
with go, you might want to try
https://github.com/ipfs/go-ipfs/issues/7227 or https://github.com/ipfs/go-ipfs/issues/7269.

PR #7282, opened 2020-05-06
write api file automically

balajijinnah:
Fix: #7227 
Signed-off-by: Tiger <rbalajis25@gmail.com>

### DIFF
--- a/repo/fsrepo/fsrepo.go
+++ b/repo/fsrepo/fsrepo.go
@@ -361,7 +361,7 @@ func (r *FSRepo) Path() string {
 func (r *FSRepo) SetAPIAddr(addr ma.Multiaddr) error {
 	// Create a temp file to write the address, so that we don't leave empty file when the
 	// program crashes after creating the file.
-	f, err := os.Create(filepath.Join(r.path, apiFile+"-tmp"))
+	f, err := os.Create(filepath.Join(r.path, "."+apiFile+".tmp"))
 	if err != nil {
 		return err
 	}
@@ -374,7 +374,7 @@ func (r *FSRepo) SetAPIAddr(addr ma.Multiaddr) error {
 	}
 
 	// Atomically rename the temp file to the correct file name.
-	return os.Rename(filepath.Join(r.path, apiFile+"-tmp"), filepath.Join(r.path, apiFile))
+	return os.Rename(filepath.Join(r.path, "."+apiFile+".tmp"), filepath.Join(r.path, apiFile))
 }
 
 // openConfig returns an error if the config file is not present.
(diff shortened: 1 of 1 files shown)


repo/fsrepo/fsrepo.go
```diff
@@ -359,14 +359,22 @@ func (r *FSRepo) Path() string {
 
 // SetAPIAddr writes the API Addr to the /api file.
 func (r *FSRepo) SetAPIAddr(addr ma.Multiaddr) error {
-	f, err := os.Create(filepath.Join(r.path, apiFile))
+	// Create a temp file to write the address, so that we don't leave empty file when the
+	// program crashes after creating the file.
+	f, err := os.Create(filepath.Join(r.path, apiFile+"-tmp"))
```
Stebalien, line 364: I'd name this something like:

```suggestion
	f, err := os.Create(filepath.Join(r.path, "." + apiFile+".tmp"))
```

repo/fsrepo/fsrepo.go
```diff
@@ -359,14 +359,22 @@ func (r *FSRepo) Path() string {
...
-	return err
+	if _, err = f.WriteString(addr.String()); err != nil {
+		return err
+	}
+	if err = f.Close(); err != nil {
+		return err
+	}
+
+	// Atomically rename the temp file to the correct file name.
+	return os.Rename(filepath.Join(r.path, apiFile+"-tmp"), filepath.Join(r.path, apiFile))
```
Stebalien, line 377:
```suggestion
	return os.Rename(filepath.Join(r.path, "." + apiFile+".tmp"), filepath.Join(r.path, apiFile))
```

repo/fsrepo/fsrepo.go
```diff
@@ -359,14 +359,22 @@ func (r *FSRepo) Path() string {
 
 // SetAPIAddr writes the API Addr to the /api file.
 func (r *FSRepo) SetAPIAddr(addr ma.Multiaddr) error {
```
Stebalien, line 361: If we fail to rename the temporary file, let's remove it.

balajijinnah: I would like to keep as it is. It feels like too much defensive programming. 

When users run the program next time, newly created tmp will override the existing tmp file. 

If you like to remove, I'll add code to remove the tmp file if rename fails

Stebalien: Well, it is defensive programming. While this is unlikely to fail on Linux, renaming _can_ fail on Windows if some other user is accessing the API file at the same time. Yes, we'll end up deleting the temporary file later, but that isn't a reason to not at least _try_ to clean it up early.

balajijinnah: Made changes. Thanks for the explaining the reason.

PR #7286, opened 2020-05-07
systemd: enable systemd hardening features

RubenKelevra:
## Motivation

IPFS is a network-facing service, which just needs very limited access to the system itself outside of it's state-directory.

Systemd allows doing a lot of hardening to separate a network process from the system and reduce the attack surface.

This PR aims to enable all possible options for the latest systemd-version.


misc/systemd/ipfs.service
```diff
@@ -3,9 +3,39 @@ Description=InterPlanetary File System (IPFS) daemon
...
+ProtectSystem=strict
+ProtectKernelTunables=true
+ProtectKernelModules=true
+ProtectKernelLogs=true
+PrivateDevices=true
+DevicePolicy=closed
+ProtectControlGroups=true
+RestrictAddressFamilies=AF_UNIX AF_INET AF_INET6 AF_NETLINK
+ProtectHostname=true
+PrivateTmp=true
```
Stebalien, line 17: This breaks the mount feature, unfortunately.

RubenKelevra: Yeah, I haven't yet tested out all features. That's why it's still a draft.

Basic HTTP-Gateway functionality with QUIC and URLStore/Filestore works fine - since Filestore is limited to /var/lib/ipfs anyway.

RubenKelevra:
I've added a reduced set of security settings, for the users who need the fuse-mount-feature.

misc/systemd/ipfs.service
```diff
@@ -3,9 +3,49 @@ Description=InterPlanetary File System (IPFS) daemon
...
+PrivateUsers=true
 # enable for 1-1024 port listening
 #AmbientCapabilities=CAP_NET_BIND_SERVICE 
+
+# if you need the fuse-mount-feature, remove the standard security
+# settings above and use the reduced set below
+#ReadWritePaths="/var/lib/ipfs/" "/ipfs" "/ipns"
+#ProtectSystem=strict
+#ProtectControlGroups=true
+#PrivateTmp=true
```
Stebalien, line 41: This will break the mount feature by running the daemon in a private mount namespace.

misc/systemd/ipfs.service
```diff
@@ -3,9 +3,49 @@ Description=InterPlanetary File System (IPFS) daemon
 After=network.target
 
 [Service]
```
Stebalien, line 5: Let's make the default security features the minimal subset that don't break go-ipfs features. Then we can have additional commented out sections: "if you don't need feature X, you can enable ...".

misc/systemd/ipfs.service
RubenKelevra: Oh, right... :/

misc/systemd/ipfs.service
RubenKelevra: I'm split on this. I think hardening is crucial on network services while the mount feature isn't used that often.

I wonder if it's not the better solution to slice the mount feature out of go-ipfs and have a separate daemon just querying the local API and mount the filesystem.

Since fuse is quite a deep integration in the kernel, there's quite a security risk to run a daemon which also talks to the web IMHO.

Until than, how about adding a flag which enables the fuse functionality, or disables it, like `--manage-fdlimit` can disable the changes to the fd-limit.

---

That said, the `ipfs` system user isn't in the fuse group, so manual intervention is necessary anyway to use the fuse feature, regardless of the security settings.

Stebalien: > I wonder if it's not the better solution to slice the mount feature out of go-ipfs and have a separate daemon just querying the local API and mount the filesystem.

It is, but that's not how it works now.

I agree that hardening is important, but I'm not going to ship a default unit that doesn't work with some go-ipfs features. We can always add a commented out section explaining how to enable the additional hardening features.

> That said, the ipfs system user isn't in the fuse group, so manual intervention is necessary anyway to use the fuse feature, regardless of the security settings.

Hm. That we should fix by adding `fuse` to the supplementary groups. Manual intervention is still required to allow accessing the mount points by other users, but that's easily discoverable.

RubenKelevra: > That we should fix by adding `fuse` to the supplementary groups

Right, but that's out of the scope of this PR. I'll open a new one.

RubenKelevra:
I've added a separate hardened service file. So the maintainer can choose between hardening and the fuse mount functionality.

misc/systemd/ipfs-hardened.service
```diff
@@ -0,0 +1,70 @@
...
+[Service]
+# hardening
+ReadWritePaths="/var/lib/ipfs/"
+NoNewPrivileges=true
+ProtectSystem=strict
+ProtectKernelTunables=true
+ProtectKernelModules=true
+ProtectKernelLogs=true
+PrivateDevices=true
+DevicePolicy=closed
```
Stebalien, line 33: This is implied by PrivateDevices, right?

misc/systemd/ipfs-hardened.service
```diff
@@ -0,0 +1,70 @@
...
+RemoveIPC=true
+RestrictSUIDSGID=true
+CapabilityBoundingSet=CAP_NET_BIND_SERVICE
+
+# enable for 1-1024 port listening
+#AmbientCapabilities=CAP_NET_BIND_SERVICE 
+# enable to specify a custom path see docs/environment-variables.md for further documentations
+#Environment=IPFS_PATH=/custom/ipfs/path
+# enable to specify a higher limit for open files/connections
+#LimitNOFILE=1000000
```
Stebalien, line 56: Given all these restrictions, will IPFS be able to set it's own file descriptor limits? If not, we should probably set this.

RubenKelevra: Yes, it can.

misc/systemd/ipfs-hardened.service
RubenKelevra: Well, the other way around, `PrivateDevices` will set `DevicePolicy=closed`.

The reasoning for setting both is, some older systemd-versions might support just the simpler `DevicePolicy=closed` but doesn't yet support `PrivateDevices`.

In this case `PrivateDevices` would be ignored and `DevicePolicy=closed` not set.

PR #7296, opened 2020-05-09
fix(mkreleasenotes): include commits directly to master

Stebalien:
This will include:

* Commits to master.
* Squashed merges.

In the release notes.


bin/mkreleaselog
```diff
@@ -6,8 +6,30 @@ export GOPATH="$(go env GOPATH)"
```
ribasushi, line 6: Not everybody has go-ipfs pulled down as an actual dep

```suggestion
! [[ -d "$GOPATH/src/github.com/ipfs/go-ipfs" ]] \
&& echo 'You must leave the current repository dir and `go get -v github.com/ipfs/go-ipfs` before using this script' \
&& exit 1
```

Stebalien: Well, this script assumes you're using a GOPATH and won't work without it, unfortunately.

PR #7302, opened 2020-05-10
systemd: make sure to wait for syslog before starting

RubenKelevra:


Stebalien:
That shouldn't be necessary, and is not a built-in target.

RubenKelevra:
Ah, damnit - I read an outdated documentation. The non-generic `systemd-journald.socket` will be added automatically for a while.

PR #7310, opened 2020-05-13
fix(mkreleaselog): make robust against running in different working directories

Stebalien:
1. Cd into the repo root before running `go list`.
2. Make sure the repo lives in the GOPATH.

fixes https://github.com/ipfs/go-ipfs/pull/7296#discussion_r422461135

### DIFF
--- a/bin/mkreleaselog
+++ b/bin/mkreleaselog
@@ -161,7 +161,13 @@ recursive_release_log() {
     local start="${1:-$(git tag -l | sort -V | grep -v -- '-rc' | grep 'v'| tail -n1)}"
     local end="${2:-$(git rev-parse HEAD)}"
     local repo_root="$(git rev-parse --show-toplevel)"
-    local package="$(go list)"
+    local package="$(cd "$repo_root" && go list)"
+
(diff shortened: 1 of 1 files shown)


Stebalien:
https://github.com/ipfs/go-ipfs/blob/8e7cead9af8fc0d8886e7b601275e6061a526dd6/bin/mkreleaselog#L5

(`go env GOPATH` will print out the default if the environment variable is unset)

ribasushi:
@Stebalien I misphrased. What I meant is that the **user** needs to know what GOPATH is, and most newcomers to go post-modules do not really understand all that (I still don't... fully).

Thus this is a bit cryptic:
https://github.com/ipfs/go-ipfs/blob/048c1c37b3811c1568f7a6f12ae07d2ba24935ed/bin/mkreleaselog#L167

Ok, it requires this, how do I fix it? :) In any case: this is maint tooling, not that important, just wanted to clarify what I meant.

Stebalien:
Ah, got it. Yeah, I'm tempted to say "if you're messing with this script, you probably know what GOPATH is".

PR #7330, opened 2020-05-18
fix(gateway): ensure directory listings have Content-Type text/html

Lekensteyn:
Files already have an explicit Content-Type set. Be sure to do this for
directory listings as well to avoid a fallback to autodetection in
net/http. That fallback fails when a ResponseWriter is installed that
performs compression.

### DIFF
--- a/core/corehttp/gateway_handler.go
+++ b/core/corehttp/gateway_handler.go
@@ -291,6 +291,15 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
 		return
 	}
 
+	// See statusResponseWriter.WriteHeader
+	// and https://github.com/ipfs/go-ipfs/issues/7164
+	// Note: this needs to occur before listingTemplate.Execute otherwise we get
+	// superfluous response.WriteHeader call from prometheus/client_golang
+	if w.Header().Get("Location") != "" {
+		w.WriteHeader(http.StatusMovedPermanently)
+		return
+	}
+
 	if r.Method == http.MethodHead {
 		return
 	}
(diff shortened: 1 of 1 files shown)


Stebalien:
How exactly did you run into this problem?

Lekensteyn:
One of the configured `corehttp.ServeOption`s in our gateway added gzip compression if the client supports it via `Accept-Encoding`. That resulted in empty `Content-Type` headers when compression is enabled, and for some reason it resulted in a reverse proxy displaying the HTML index as plain text.

Stebalien:
Ah, I see.

core/corehttp/gateway_handler.go
```diff
@@ -286,6 +286,7 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
 		return
 	}
 
+	w.Header().Set("Content-Type", "text/html")
```
Stebalien, line 289: Let's move this down to where we actually render the template (https://github.com/ipfs/go-ipfs/pull/7330/files#diff-19ceb36d78bde310c54cc44a95bec44cR353).

Lekensteyn: It was done on purpose, ensuring that a successful response to HEAD and GET requests both return the same value. There seem to be some `internalWebError` error cases though where text/html might not be desirable though. I'll have a look.

Stebalien: Yeah, I was concerned about the error cases. Maybe just apply it before HEAD?

Stebalien:
(i.e., inside the branch and again before rendering the template)

core/corehttp/gateway_handler.go
Lekensteyn: What about explicitly setting `Content-Type: text/plain` if we know for a fact that we are returning a plaintext error message (if this is not already done)? That should treat all cases where this issue could manifest, namely that content-sniffing is prone to emitting the wrong result.

Stebalien: Good point. We can set that inside `internalWebError` which should cover all cases. Could you add that change?

Lekensteyn: It looks like error cases are already correctly treated, `http.Error` will overwrite the `Content-Type`.

Lekensteyn:
There is only one error case now, namely when the directory contents cannot be read. In that special case, the content-type will be text/plain while the HEAD request will return text/html. I'd say that is acceptable.

Since the previous version, #4233 got merged which introduces a new edge case: if `ipfs-404.html` is present, the directory index is not shown. However even for that case, text/html is appropriate.

Stebalien:
Thanks!

Issue #7246, opened 2020-04-29
bad links via IPNS mountpoint


neededanametoreportbug:
ipfs version --all
go-ipfs version: 0.5.0
Repo version: 9
System version: amd64/linux
Golang version: go1.13.10

Trying to access content via the IPNS filesystem mount point results in broken links to the IPFS filesystem mount point, except for the local peer.

eg:
ipfs mount
>IPFS mounted at: /ipfs
>IPNS mounted at: /ipns

ipfs name resolve ipfs.io
>/ipfs/bafybeih2idmvr2uum3k7btt6mjxfzeb7ank5g6zkmrfbxkvu6ownzbrtky

cd /ipns/ipfs.io
>bash: cd: /ipns/ipfs.io: No such file or directory

ls /ipns/ipfs.io
>lr-xr-xr-x. 1 root root 0 Apr 29 19:26 /ipfs.io -> /ipfs//ipfs/

Looks like the link has an unnecessary "/ipfs/" added to the destination

However local peer is visible via the ipns mount point though, eg :/ipns/local and /ipns/[peer id]

hsanjuan:
I can reproduce this and it's a regression from v0.4.23. However I am not sure when it happened.

PR #7331, opened 2020-05-20
fix: correctly trim resolved IPNS addresses

Stebalien:
fixes https://github.com/ipfs/go-ipfs/issues/7246

### DIFF
--- a/test/sharness/t0030-mount.sh
+++ b/test/sharness/t0030-mount.sh
@@ -15,13 +15,15 @@ if ! test_have_prereq FUSE; then
   test_done
 fi
 
+
+export IPFS_NS_MAP="welcome.example.com:/ipfs/$HASH_WELCOME_DOCS"
+
 # start iptb + wait for peering
 NUM_NODES=5
 test_expect_success 'init iptb' '
   iptb testbed create -type localipfs -count $NUM_NODES -init
 '
 startup_cluster $NUM_NODES
-
 
 # test mount failure before mounting properly.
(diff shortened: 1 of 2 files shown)


Stebalien:
This test is tests resolving keys through the `/ipns` mountpoint. That code just calls `Resolve` using the CoreAPI so the test should be generic.

We do need to start running this in CI, but we'll need a special job for fuse-only test.

> Does it work successfully for you?

It does... do the other tests work?

test/sharness/t0030-mount.sh
```diff
@@ -53,6 +55,12 @@ test_expect_success FUSE "'ipfs mount' output looks good" '
   test_cmp expected actual
 '
 
+test_expect_success FUSE "can resolve ipns names" '
```
aschmahmann, line 58: Continuing from https://github.com/ipfs/go-ipfs/pull/7331#issuecomment-631180756 so we can do this in a thread-like fashion.

Running `./t0030-mount.sh -v` passes on master, but with this branch fails with:
`cat: ipns/welcome.example.com/ping: No such file or directory`

Stebalien: Did you make sure to rebuild? Otherwise, I have no idea. Does `ipns/local` exist when running the test?

aschmahmann: I think I had a rebuilding issue... so dumb. Sorry for the confusion

PR #7333, opened 2020-05-20
misc config/experimental-features doc fixes

Stebalien:
* sort config sections (ignore that commit when reviewing)
* add a pubsub config section
* update pubsub experimental feature.


Stebalien:
@aschmahmann feel free to make those changes (and any other doc cleanups, we have many...)

PR #7335, opened 2020-05-20
chore: update libp2p

Stebalien:


Stebalien:
(eh, we can do a post-merge check with this one as well)

PR #7339, opened 2020-05-21
fix: support directory listings even if a 404 page is present

Stebalien:
fixes https://github.com/ipfs/go-ipfs/pull/4233#issuecomment-631454543

Basically, there's a trade-off here:

1. We can support directory listings while supporting 404 pages (this PR).
2. If a 404 page is present, directory listings don't work.

Given that option 1 is more flexible and users shouldn't be _too_ confused if they land on a directory with no index.html page, I've gone with that option.


Stebalien:
cc @Lekensteyn @jphastings

PR #7340, opened 2020-05-21
feat: bootstrap in dht when the routing table is empty

Stebalien:
Otherwise, we could end up with only DHT clients and never re-bootstrap.

I've left the default go-ipfs bootstrapping code in for now as it's technically possible to disable the DHT entirely.

### DIFF
--- a/core/node/libp2p/host.go
+++ b/core/node/libp2p/host.go
@@ -10,10 +10,11 @@ import (
 	routing "github.com/libp2p/go-libp2p-core/routing"
 	record "github.com/libp2p/go-libp2p-record"
 	routedhost "github.com/libp2p/go-libp2p/p2p/host/routed"
-	"go.uber.org/fx"
 
 	"github.com/ipfs/go-ipfs/core/node/helpers"
 	"github.com/ipfs/go-ipfs/repo"
+
+	"go.uber.org/fx"
 )
 
 type P2PHostIn struct {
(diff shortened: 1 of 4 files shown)


core/node/libp2p/host.go
```diff
@@ -43,9 +46,21 @@ func Host(mctx helpers.MetricsCtx, lc fx.Lifecycle, params P2PHostIn) (out P2PHo
...
+		return out, err
+	}
+	var bootstrappers []ma.Multiaddr
+	for _, addrStr := range cfg.Bootstrap {
+		addr, err := ma.NewMultiaddr(addrStr)
+		if err != nil {
+			return out, fmt.Errorf("failed to parse bootstrap peer address %q: %s", addrStr, err)
+		}
+		bootstrappers = append(bootstrappers, addr)
+	}
```
Stebalien, line 60: Changing `dht.BootstrapPeers` to take `peer.AddrInfo` would make this code simpler...

core/node/libp2p/host.go
```diff
@@ -43,9 +46,21 @@ func Host(mctx helpers.MetricsCtx, lc fx.Lifecycle, params P2PHostIn) (out P2PHo
 	}
 
 	ctx := helpers.LifecycleCtx(mctx, lc)
+	cfg, err := params.Repo.Config()
+	if err != nil {
+		return out, err
+	}
+	var bootstrappers []ma.Multiaddr
+	for _, addrStr := range cfg.Bootstrap {
```
aarshkshah1992, line 54: @Stebalien If the default for `cfg.Bootstrap` the `DefaultBootstrapPeers` in `go-libp2p-kad-dht` ?

core/node/libp2p/host.go
aarshkshah1992: @Stebalien Have made the change here:
https://github.com/libp2p/go-libp2p-kad-dht/pull/651

Also, we should now use `go-libp2p-kad-dht.GetDefaultBootstrapPeerAddrInfos` to set the default for `cfg. Bootstrap`.

core/node/libp2p/host.go
aarshkshah1992, line 54: @Stebalien Is the default for cfg.Bootstrap the DefaultBootstrapPeers in go-libp2p-kad-dht ?

core/node/libp2p/host.go
aarshkshah1992: @Stebalien Have made the change here:
libp2p/go-libp2p-kad-dht#651

Also, we should now use go-libp2p-kad-dht.GetDefaultBootstrapPeerAddrInfos to set the default for cfg. Bootstrap.

aschmahmann: I'm not sure I really understand what makes this easier, but also no complaints from me :)

Stebalien: It's easier in this case because `cfg` has a special function to get AddrInfos.

core/node/libp2p/host.go
Stebalien: Basically, yes. Although I think it's duplicated.

PR #7349, opened 2020-05-22
quic: remove experimental status and add it to the default config

RubenKelevra:


core/node/groups.go
```diff
@@ -130,7 +130,6 @@ func LibP2P(bcfg *BuildCfg, cfg *config.Config) fx.Option {
 		maybeProvide(libp2p.BandwidthCounter, !cfg.Swarm.DisableBandwidthMetrics),
 		maybeProvide(libp2p.NatPortMap, !cfg.Swarm.DisableNatPortMap),
 		maybeProvide(libp2p.AutoRelay, cfg.Swarm.EnableAutoRelay),
-		maybeProvide(libp2p.QUIC, cfg.Experimental.QUIC),
```
Stebalien, line 133:
```go
 		fx.Provide(libp2p.QUIC),
```

Stebalien: (nvm)

Stebalien:
You'll also need to change DefaultTransports in core/node/libp2p/transport.go to `simpleOpt(libp2p.ChainOptions(libp2p.DefaultTransports, libpp2quic.NewTransport))`.

RubenKelevra:
@Stebalien The private network test is expected to fail since QUIC doesn't support private networks. I guess we have to modify this test, to remove QUIC from it.

Stebalien:
@RubenKelevra hm, I forgot about that. I think we need to avoid constructing the QUIC transport when a swarm key is present.

RubenKelevra:
Or we just turn QUIc only on, if we're starting to listening on a QUIC port.

This way the user can determine if he/she wants QUIC with the listening definition.

Since the test only listens to TCP it wouldn't fail in this case.


Feel free to push the commit here, if you decided on that. Both is a bit over my head right now :)

Issue #7318, opened 2020-05-14
Subdomain support for CIDs longer than 63


lidel:
> I hoped to punt this until we need to switch away from sha256 in CIDs, but we may need to solve this problem sooner than expected due to ED25519 keys being new default soon (https://github.com/ipfs/go-ipfs/issues/6916)

## Problem: DNS label limit of 63

> [RFC 1034](https://tools.ietf.org/html/rfc1034#page-7): "each node has a label, which is zero to 63 octets in length"

The default CIDv1 Base32  with multihash of sha256 and RSA libp2p-key fits:
- Default CID: http://bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi.ipfs.dweb.link
- RSA: https://bafzbeih7uocwo6vmbjusf4wzw6h5cruhw3gf4jxhevmxssnz5vkgryk2za.ipns.dweb.link

but if we use ED25519  libp2p-key then we are 2 characters over the limit:
- ED25519 libp2p-key: https://bafzaajaiaejca4syrpdu6gdx4wsdnokxkprgzxf4wrstuc34gxw5k5jrag2so5gk.ipns.dweb.link
- CID created with `--hash sha2-512` will be even longer: https://bafkrgqe3ohjcjplc6n4f3fwunlj6upltggn7xqujbsvnvyw764srszz4u4rshq6ztos4chl4plgg4ffyyxnayrtdi5oc4xb2332g645433aeg.ipfs.dweb.link

Label longer than 63 characters  means the hostname can't resolve:

```console
$ ping bafzaajaiaejca4syrpdu6gdx4wsdnokxkprgzxf4wrstuc34gxw5k5jrag2so5gk.ipns.dweb.link
ping: bafzaajaiaejca4syrpdu6gdx4wsdnokxkprgzxf4wrstuc34gxw5k5jrag2so5gk.ipns.dweb.link: Name or service not known
```

And links are not picked up by tools like Slack:

> ![oops-2020-05-14--17-59-27](https://user-images.githubusercontent.com/157609/81957284-ad54a100-960c-11ea-808f-1de28688f440.png)

**Note:** I used ED25519 as an example, but not limited to that single type of CID. Even if we find a way to fit ED25519 in a single label, the problem remains for CIDs with a multihash created with longer hash functions.

For e

## Solutions

I am happy to open PR with a fix, but unsure if I have the best fix in mind, would love to gather feedback first.

[...]

PR #7358, opened 2020-05-25
feat: support long CIDs in subdomains by splitting at 63rd char

lidel:
> - This is a "generic" fix for #7318 that solves DNS problem for all CIDs
> - IPNS will get additional one in form of ED25519 libp2p-keys represented in Base36, but that will be a separate PR when https://github.com/ipfs/go-cid/pull/107 lands. 
> - **Note:** significant downside of this approach is no TLS (impossible to get a wildcard cart for more than one level of of subdomains)

This PR adds subdomain gateway support for CIDs longer than 63 characters:

- CID is split after reaching 63 character limit counting from right to left. 
- Requests made with random splits are redirected to canonical split version to ensure every CID gets exactly one Origin.
- go  and sharness tests document behavior for longer CIDs

"Splitting logic" is in `toDNSSafePrefix` – would love some eyes on it. 
I am unsure if I it is the most efficient way to split into chunks of 63 starting from right to left, suggestions welcome. 

Ref.
- https://tools.ietf.org/html/rfc1034#page-7
- https://github.com/ipfs/go-ipfs/issues/7318

### DIFF
--- a/core/corehttp/hostname.go
+++ b/core/corehttp/hostname.go
@@ -294,30 +294,31 @@ func isPeerIDNamespace(ns string) bool {
 func toDNSSafePrefix(id string) (prefix string) {
 	s := strings.Replace(id, ".", "", -1) // remove separators if present
 
-	// Return ID as-is if already in canonical form and safe for use in DNS
-	if s == id && len(s) <= dnsLabelMaxLength {
-		return id
+	// Return if things fit after dot removal
+	if len(s) <= dnsLabelMaxLength {
+		return s
 	}
 
-	var b strings.Builder
-	split := dnsLabelMaxLength - 1
-	end := len(s) - 1
+	parts := make(
+		[]string,
+		// same as ceil( len(s) / dnsLabelMaxLength )
+		(len(s)+dnsLabelMaxLength-1)/dnsLabelMaxLength,
+	)
 
-	// Iterate from the right to left to maximize length of first label
-	for i := end; i >= 0; i-- {
-		b.WriteByte(s[i])
-		if end-i%dnsLabelMaxLength == split && i != end {
-			b.WriteRune('.')
-		}
+	firstPartLen := len(s) % dnsLabelMaxLength
+
+	// if it divides by 63 perfectly - full part
+	if firstPartLen == 0 {
+		firstPartLen = dnsLabelMaxLength
 	}
 
-	// Reverse produced string
-	chars := []rune(b.String())
-	for i, j := 0, len(chars)-1; i < j; i, j = i+1, j-1 {
-		chars[i], chars[j] = chars[j], chars[i]
+	// Iterate from right to left to maximize length of right-most labels
+	for i := len(parts) - 1; i > 0; i-- {
+		parts[i] = s[(i-1)*dnsLabelMaxLength+firstPartLen : i*dnsLabelMaxLength+firstPartLen]
 	}
+	parts[0] = s[:firstPartLen]
 
-	return string(chars)
+	return strings.Join(parts, ".")
 }
 
 // Converts a hostname/path to a subdomain-based URL, if applicable.
--- a/core/corehttp/hostname.go
+++ b/core/corehttp/hostname.go
@@ -41,6 +41,9 @@ var defaultKnownGateways = map[string]config.GatewaySpec{
 	"dweb.link":       subdomainGatewaySpec,
 }
 
+// Label's max length in DNS (https://tools.ietf.org/html/rfc1034#page-7)
+const dnsLabelMaxLength int = 63
+
 // HostnameOption rewrites an incoming request based on the Host header.
 func HostnameOption() ServeOption {
 	return func(n *core.IpfsNode, _ net.Listener, mux *http.ServeMux) (*http.ServeMux, error) {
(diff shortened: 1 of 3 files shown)


core/corehttp/hostname.go
```diff
@@ -266,6 +290,36 @@ func isPeerIDNamespace(ns string) bool {
 	}
 }
 
+// Converts an identifier to DNS-safe representation
+func toDNSSafePrefix(id string) (prefix string) {
```
ribasushi, line 294: I have a better version of this but something is wrong with my setup - tests won't run.
I'll push a PR against this branch when it's ready

lidel: Thank you, 45e3271a0dad2a3438aacc56513f771ee8d9679d is now part of this PR.

core/corehttp/hostname.go
```diff
@@ -151,16 +154,29 @@ func HostnameOption() ServeOption {
...
 				}
 
-				// Do we need to fix multicodec in PeerID represented as CIDv1?
-				if isPeerIDNamespace(ns) {
-					keyCid, err := cid.Decode(rootID)
-					if err == nil && keyCid.Type() != cid.Libp2pKey {
+				// Check if rootID is a valid CID
+				if rootCID, err := cid.Decode(rootID); err == nil {
+					// Do we need to redirect CID to a canonical DNS representation?
+					canonicalPrefix := toDNSSafePrefix(rootID)
```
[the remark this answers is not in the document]
lidel: Simplified names in https://github.com/ipfs/go-ipfs/pull/7358/commits/a74dada3c2302c0e12ebec20d3154964957bad37

lidel:
@Stebalien ready for you to take a look

core/corehttp/hostname.go
```diff
@@ -266,6 +290,37 @@ func isPeerIDNamespace(ns string) bool {
...
+	// if it divides by 63 perfectly - full part
+	if firstPartLen == 0 {
+		firstPartLen = dnsLabelMaxLength
+	}
+
+	// Iterate from right to left to maximize length of right-most labels
+	for i := len(parts) - 1; i > 0; i-- {
+		parts[i] = s[(i-1)*dnsLabelMaxLength+firstPartLen : i*dnsLabelMaxLength+firstPartLen]
+	}
+	parts[0] = s[:firstPartLen]
```
Stebalien, line 319: Wouldn't you want to maximize the left-most labels? That is, make the right-most labels as small as possible?

We're only over by two base-32 numerals so we can _probably_ get `32**2 = 1024` wildcard certificates. It looks like we can register 100 names per cert, and 50 certs which gives us 5000 certs.

lidel: We wanted to maximize the first part after [public suffix](https://publicsuffix.org), but you have a point, it may be possible to get TLS certs if we hardcode the first split to be after two characters from the right. 

My main concern: while technically possible, I am not sure if it is practical / better than using Base36 for ED25519.

Assuming we have all the certs in place, this fix would work only for ED25519 represented in base32, everything else (eg. sha2-512)  will still be over the wildcard length limit:

- :+1: ED25519 libp2p-key is only 2 characters over the limit, so it fits in two labels: https://bafzaajaiaejca4syrpdu6gdx4wsdnokxkprgzxf4wrstuc34gxw5k5jrag2so5.gk.ipns.dweb.link
  - `bafzaajaiaejca4syrpdu6gdx4wsdnokxkprgzxf4wrstuc34gxw5k5jrag2so5` . `gk`
- :anger:  CID created with `--hash sha2-512` will be too long, needs to be split into more than two chunks anyway: https://bafkrgqe3ohjcjplc6n4f3fwunlj6upltggn7xqujbsv.nvyw764srszz4u4rshq6ztos4chl4plgg4ffyyxnayrtdi5oc4xb2332g645433a.eg.ipfs.dweb.link:
  - `bafkrgqe3ohjcjplc6n4f3fwunlj6upltggn7xqujbsv`  . `nvyw764srszz4u4rshq6ztos4chl4plgg4ffyyxnayrtdi5oc4xb2332g645433a` . `eg`


 
Let's Encrypt limits are ([src](https://community.letsencrypt.org/t/wildcard-certificate-limit/86536/2)):
> - You can only fit 100 domains onto one certificate
> - Each Registered Domain may only appear on 50 certificates per week.

Interesting part is that according to the [rate limits](https://letsencrypt.org/docs/rate-limits/) they

> use the Public Suffix List to calculate the registered domain. 

`ipfs.dweb.link` and `ipns.dweb.link` are both on Public Suffix List, which means they are "excluded" from limit calculation, and the first CID chunk becomes "Registered Domain", effectively removing the weekly limit. 

I'd appreciate input from @mburns and @MichaelMure on how feasible it is for us, or operator like Infura to get around 100 certs, each for 100 wildcard names to support TLS on `*.aa.dweb.link`, `*.ab.dweb.link` ...  etc, **and if it's better than using Base36 and not splitting IPNS ED25519 names.**

Thoughts?

MichaelMure: I found out about this public suffix "loophole" earlier today as well. It could make things easier, but it's also somewhat abusive and risky (do they even update their list?). There is also the possibility to apply for a raised rate limit but I don't know of feasible that is. Not even talking about the vendor locking with LetsEncrypt.

One problem coming from this solution is also that it pretty much means doing TLS termination manually. AWS for example limit to 25 certificates attached to a load balancer. It compound quickly in complexity.

Sadly we might need to go that way anyway because we need to have a user ID as a subdomain (we need to link requests to a user somehow, basic auth break some usecases, we can't touch the path part of the URL).

In any case, base36 is way way easier for us.

Stebalien: I agree with using base36. And you're right, _this_ feature is just to support sha512 etc. so it doesn't really matter.

> We want to maximize the first part after public suffix, giving the most bits to parent labels, following the spirit of DNS hierarchy.

I'm not sure I follow. Why would we want to maximize that part? Are we worried about cookies/origins?

mburns: Managing that many certs is doable, but feels a bit icky, technically speaking.

all else being equal, sounds like base36 is preferable.

lidel: > Why would we want to maximize that part? Are we worried about cookies/origins?

AFAIK in practice, it does not matter. 
But if we can pick, I'd maximize "parent" labels on the right, as a _precaution_.  

My rationale: the surface for bugs always exists.
To illustrate: a mild version of "Origin sharing" is a thing. Two sibling subdomains can _mutually_ agree to use parent Origin for cookies (_both_ setting `document.domain = aa.ipfs.example`, assuming `example.com` is not on Public Suffix List). 

Let's say the future brings a bug/vulnerability in Origin-separation code in one of browser vendors. Maximizing parent label makes it harder/infeasible to pull this class of attacks off, as generating parent label if way more difficult if it needs to match 63 instead of 2 chars.

ribasushi: TLDR: seconding what @lidel said

In fact this very thread contains something that walks and talks like a precursor to a phishing attack: the ability to blanket register a star-TLS cert to all possible N-character prefixes when N is short enough at the top.

Before this thread started this was also one of the 1st questions I asked @lidel when I was doing review, because I was dumb and misread the code: "@lidel why do you let the rightmost side be short and thus brute-forceable?"

I do not have a good feeling how to "productize" this into an outright vulnerability, but leaving a "mid-to-top-level" part of DNS trivially forgeable doesn't... smell right at all.

lidel:
Let's park this for now (closing to reduce confusion),
and evaluate alternative that does not break TLS:  #7441

PR #7360, opened 2020-05-25
Include the git blob id of the dir-index bundle in the ETag

ribasushi:
While the content of raw files retrieved via the gateway should never
change, the look and feel of the directory index can and will change
between versions of go-ipfs.

Incorporate the hash of assets/bindata.go into the ETag when appropriate

@Stebalien this is what we spoke about <gulp> last year @ camp. Goes with @jessicaschilling's recent changes.

### DIFF
--- a/assets/bindata_version_hash.go
+++ b/assets/bindata_version_hash.go
@@ -1,0 +1,5 @@
+package assets
+
+const (
+	BindataVersionHash = "c1aa0601ac3eac2c50b296cf618a6747eeba8579"
+)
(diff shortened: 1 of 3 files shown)


ribasushi:
Context: https://github.com/ipfs/go-ipfs/pull/6465#pullrequestreview-255129697

assets/bindata_version_hash.go
```diff
@@ -0,0 +1,5 @@
+package assets
```
Stebalien, line 1: Can we add a comment here to say that this shouldn't be edited?

ribasushi: ✅ done

Stebalien:
Thanks!

Issue #6097, opened 2019-03-18
Add functionality for advanced control over peer set


whyrusleeping:
I'd really like to have the ability to specify peer IDs in my config file that my node will try to always remain connected to. Additionally, it would be nice to specify strategies for these peers, like "always", which attempts to reconnect on any disconnect (respecting backoff rules), "preferred" which never closed a connection to that peer, but doesnt necessarily try and hold open a connection, and maybe one more that just increases the likelihood the conn manager will hold open the connection, but still allows it to be closed as needed.

This will enable me to keep connections open to friends machines, making transfers between us much more reliable.

whyrusleeping:
Additionally, the infra team wants this to hold connections between the gateways and our pinning servers open.

raulk:
This will percolate to the connection manager (current, interim and future).

brianmcmichael:
This feature would mean that I can set up a gateway node that could 'permanently' connect to pin storage node and speed up propagation when the machine holding the content is known.

obo20:
We would also highly benefit from this type of feature. We're currently running an automated "swarm connect {gatewayAddr}" with our nodes every 5 minutes or so to keep these connections open.

Having an official "supported" way of keeping nodes connected would be amazing.

whyrusleeping:
@Stebalien @raulk A pretty quick and non-invasive way of doing this would be to add a list of peers to the connection manager that lets it handle the ‘preferred’ case (not closing the connection). Then, in go-ipfs we can have a little process that is fed a list of ‘always’ peers that it dials, and listens for disconnects from.

brianmcmichael:
Geth has the "static nodes" feature which may be useful as a development/design pattern for go-ipfs.

https://github.com/ethereum/go-ethereum/wiki/Connecting-to-the-network#static-nodes

raulk:
@whyrusleeping let's do that quickly. Connection manager v2 proposal is in the works, but there's no reason we cannot implement a _protected set_. I'll work on a patch.

raulk:
@whyrusleeping 

> like "always", which attempts to reconnect on any disconnect (respecting backoff rules), 

Do you think libp2p should take care of reestablishing the connection? This would require changes in the host and/or the swarm, e.g. when you `host.Connect()` you could specify a _supervision policy_ for that connection.

> "preferred" which never closed a connection to that peer, but doesnt necessarily try and hold open a connection,

[...]

PR #7362, opened 2020-05-26
feat: implement peering service

Stebalien:
MVP for #6097

This feature will repeatedly reconnect (with a randomized exponential backoff) to peers in a set of "peered" peers.

In the future, this should be extended to:

1. Include a CLI for modifying this list at runtime.
2. Include additional options for peers we want to _protect_ but not connect to.
3. Allow configuring timeouts, backoff, etc.
4. Allow groups? Possibly through textile threads.
5. Allow for runtime-only peering rules.
6. Different reconnect policies.

But this MVP should be a significant step forward.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -709,9 +709,39 @@ Default: `false`
 ### `Peering`
 
 Configures the peering subsystem. The peering subsystem configures go-ipfs to
-connect to, remain connected to, and reconnect to a set of peers. Peers should
-use this subsystem to create "sticky" links between frequently used peers for
-improved reliability.
+connect to, remain connected to, and reconnect to a set of nodes. Nodes should
+use this subsystem to create "sticky" links between frequently useful peers to
+improve reliability.
+
+Use-cases:
+
+* An IPFS gateway connected to an IPFS cluster should peer to ensure that the
+  gateway can always fetch content from the cluster.
+* A dapp may peer embedded go-ipfs nodes with a set of pinning services or
+  textile cafes/hubs.
+* A set of friends may peer to ensure that they can always fetch each other's
+  content.
+
+When a node is added to the set of peered nodes, go-ipfs will:
+
+1. Protect connections to this node from the connection manager. That is,
+   go-ipfs will never automatically close the connection to this node and
+   connections to this node will not count towards the connection limit.
+2. Connect to this node on startup.
+3. Repeatedly try to reconnect to this node if the last connection dies or the
+   node goes offline. This repeated re-connect logic is governed by a randomized
+   exponential backoff delay ranging from ~5 seconds to ~10 minutes to avoid
+   repeatedly reconnect to a node that's offline.
+
+Peering can be asymmetric or symmetric:
+
+* When symmetric, the connection will be protected by both nodes and will likely
+  be vary stable.
+* When asymmetric, only one node (the node that configured peering) will protect
+  the connection and attempt to re-connect to the peered node on disconnect. If
+  the peered node is under heavy load and/or has a low connection limit, the
+  connection may flap repeatedly. Be careful when asymmetrically peering to not
+  overload peers.
 
 #### `Peering.Peers`
 
--- a/docs/config.md
+++ b/docs/config.md
@@ -139,6 +139,8 @@ documented in `ipfs config profile --help`.
 - [`Pubsub`](#pubsub)
     - [`Pubsub.Router`](#pubsubrouter)
     - [`Pubsub.DisableSigning`](#pubsubdisablesigning)
+    - [`Peering`](#peering)
+        - [`Peering.Peers`](#peeringpeers)
 - [`Reprovider`](#reprovider)
     - [`Reprovider.Interval`](#reproviderinterval)
     - [`Reprovider.Strategy`](#reproviderstrategy)
@@ -156,6 +158,7 @@ documented in `ipfs config profile --help`.
         - [`Swarm.ConnMgr.LowWater`](#swarmconnmgrlowwater)
         - [`Swarm.ConnMgr.HighWater`](#swarmconnmgrhighwater)
         - [`Swarm.ConnMgr.GracePeriod`](#swarmconnmgrgraceperiod)
+
 
 ## `Addresses`
 
@@ -702,6 +705,26 @@ message because spoofed messages can be used to silence real messages by
 intentionally re-using the real message's message ID.
 
 Default: `false`
+
+### `Peering`
+
+Configures the peering subsystem. The peering subsystem configures go-ipfs to
+connect to, remain connected to, and reconnect to a set of peers. Peers should
+use this subsystem to create "sticky" links between frequently used peers for
+improved reliability.
+
+#### `Peering.Peers`
+
+The set of peers with which to peer. Each entry is of the form:
+
+```js
+{
+  "ID": "QmSomePeerID", # The peers ID.
+  "Addrs": ["/ip4/1.2.3.4/tcp/1234"] # Known addresses for the peer. If none are specified, the DHT will be queried.
+}
+```
+
+Additional fields may be added in the future.
 
 ## `Reprovider`
 
--- a/go.mod
+++ b/go.mod
@@ -32,7 +32,7 @@ require (
 	github.com/ipfs/go-ipfs-blockstore v0.1.4
 	github.com/ipfs/go-ipfs-chunker v0.0.5
 	github.com/ipfs/go-ipfs-cmds v0.2.9
-	github.com/ipfs/go-ipfs-config v0.6.1
+	github.com/ipfs/go-ipfs-config v0.7.0
 	github.com/ipfs/go-ipfs-ds-help v0.1.1
 	github.com/ipfs/go-ipfs-exchange-interface v0.0.1
 	github.com/ipfs/go-ipfs-exchange-offline v0.0.1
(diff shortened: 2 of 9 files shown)


Stebalien:
Depends on https://github.com/ipfs/go-ipfs-config/pull/96

docs/config.md
```diff
@@ -703,6 +706,26 @@ intentionally re-using the real message's message ID.
 
 Default: `false`
 
+### `Peering`
```
willscott, line 709: include expected behavior / semantics for how this behaves when one of the two peers has it, and when both have it mutually.

go.mod
```diff
@@ -32,7 +32,7 @@ require (
 	github.com/ipfs/go-ipfs-blockstore v0.1.4
 	github.com/ipfs/go-ipfs-chunker v0.0.5
 	github.com/ipfs/go-ipfs-cmds v0.2.9
-	github.com/ipfs/go-ipfs-config v0.6.1
```
willscott, line 35: todo: update to released version

peering/peering.go
```diff
@@ -0,0 +1,259 @@
...
+	"github.com/ipfs/go-log"
+	"github.com/libp2p/go-libp2p-core/host"
+	"github.com/libp2p/go-libp2p-core/network"
+	"github.com/libp2p/go-libp2p-core/peer"
+	"github.com/multiformats/go-multiaddr"
+)
+
+// maxBackoff is the maximum time between reconnect attempts.
+const (
+	maxBackoff = 10 * time.Minute
```
willscott, line 19:
```suggestion
const (
	// maxBackoff is the maximum time between reconnect attempts.
	maxBackoff = 10 * time.Minute
```

peering/peering.go
```diff
@@ -0,0 +1,259 @@
...
+// peerHandler keeps track of all state related to a specific "peering" peer.
+type peerHandler struct {
+	peer   peer.ID
+	host   host.Host
+	ctx    context.Context
+	cancel context.CancelFunc
+
+	mu    sync.Mutex
+	addrs []multiaddr.Multiaddr
+	timer *time.Timer
```
willscott, line 45: better name for this?

peering/peering.go
```diff
@@ -0,0 +1,259 @@
...
+
+	nextDelay time.Duration
+}
+
+func (ph *peerHandler) stop() {
```
willscott, line 50: cancel ph.ctx ?

peering/peering_test.go
```diff
@@ -0,0 +1,6 @@
+package peering
+
+import "testing"
+
+func TestPeeringService(t *testing.T) {
```
willscott, line 5: 😢

Stebalien: Oops. I went and wrote a sharness test instead. But you're right, I should have one here as well.

test/sharness/t0171-peering.sh
```diff
@@ -0,0 +1,127 @@
...
+  iptb run -- ipfs log level peering debug
+'
+
+check_peers
+
+disconnect() {
+    ipfsi "$1" swarm disconnect "/p2p/$(peer_id "$2")"
+}
+
+# Bidiractional peering shouldn't cause problems (e.g., simultaneous connect
```
aschmahmann, line 90:
```suggestion
# Bidirectional peering shouldn't cause problems (e.g., simultaneous connect
```

peering/peering.go
```diff
@@ -0,0 +1,259 @@
...
+
+	if ph.timer != nil {
+		ph.timer.Stop()
+		ph.timer = nil
+	}
+}
+
+func (ph *peerHandler) nextBackoff() time.Duration {
+	// calculate the timeout
+	if ph.nextDelay < maxBackoff {
```
aschmahmann, line 62: technically this isn't actually the max backoff. The largest potential backoff is maxBackoff * 2.5. I'd either change the `maxBackoff` comment or just do a check `if ph.nextDelay > maxBackoff; return maxBackoff`

peering/peering.go
```diff
@@ -0,0 +1,259 @@
...
+			ph.timer.Reset(ph.nextBackoff())
+		}
+		// Otherwise, someone else has stopped us so we can assume that
+		// we're either connected or someone else will start us.
+		ph.mu.Unlock()
+	}
+
+	// Always call this. We could have connected since we processed the
+	// error.
+	ph.stopIfConnected()
```
aschmahmann, line 94: nit: this isn't useful if `err != nil and ph.timer != nil` we could just return after unlocking the mutex. I'm fine keeping it if you think it's easier to read though.

peering/peering.go
```diff
@@ -0,0 +1,259 @@
...
+// PeeringService maintains connections to specified peers, reconnecting on
+// disconnect with a back-off.
+type PeeringService struct {
+	host host.Host
+
+	mu    sync.RWMutex
+	peers map[peer.ID]*peerHandler
+
+	ctx    context.Context
+	cancel context.CancelFunc
```
aschmahmann, line 130: I don't understand why this is necessary? We only use the context and cancellation to start a `peerHandler`, but every time we call `ph.cancel` we'll also call `ph.stop`. It seems to me we could remove this and just use `handler.ctx, handler.cancel = context.WithCancel(context.Background)` without any effect.

peering/peering.go
```diff
@@ -0,0 +1,259 @@
...
+func (ps *PeeringService) RemovePeer(id peer.ID) {
+	ps.mu.Lock()
+	defer ps.mu.Unlock()
+
+	if handler, ok := ps.peers[id]; ok {
+		logger.Infow("peer removed", "peer", id)
+		ps.host.ConnManager().Unprotect(id, connmgrTag)
+
+		handler.stop()
+		handler.cancel()
```
aschmahmann, line 225: It seems like `handler.cancel()` be called inside of `handler.stop`, especially with @willscott's suggestion https://github.com/ipfs/go-ipfs/pull/7362/files#r430115812.

Am I missing something?

peering/peering.go
```diff
@@ -0,0 +1,259 @@
...
+//
+// Add peer may also be called multiple times for the same peer. The new
+// addresses will replace the old.
+func (ps *PeeringService) AddPeer(info peer.AddrInfo) {
+	ps.mu.Lock()
+	defer ps.mu.Unlock()
+
+	if handler, ok := ps.peers[info.ID]; ok {
+		logger.Infow("updating addresses", "peer", info.ID, "addrs", info.Addrs)
+		handler.addrs = info.Addrs
```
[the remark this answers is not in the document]
Stebalien: Good catch!

docs/config.md
```diff
@@ -139,6 +139,8 @@ documented in `ipfs config profile --help`.
 - [`Pubsub`](#pubsub)
     - [`Pubsub.Router`](#pubsubrouter)
     - [`Pubsub.DisableSigning`](#pubsubdisablesigning)
+    - [`Peering`](#peering)
+        - [`Peering.Peers`](#peeringpeers)
```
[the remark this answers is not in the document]
Stebalien: I figured it was a minimal enough feature that that wasn't _really_ necessary. I want people to start using this feature. Thoughts?

peering/peering.go
```diff
@@ -0,0 +1,259 @@
...
+	}
+}
+
+func (ph *peerHandler) nextBackoff() time.Duration {
+	// calculate the timeout
+	if ph.nextDelay < maxBackoff {
+		ph.nextDelay += ph.nextDelay/2 + time.Duration(rand.Int63n(int64(ph.nextDelay)))
+	}
+	return ph.nextDelay
+}
```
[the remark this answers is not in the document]
Stebalien: Yeah, well, go...

peering/peering.go
Stebalien: True. But yeah, I think this is easier to read (and doesn't make much of a difference).

peering/peering.go
Stebalien: You're right. it was mostly historical. I'll remove it.

peering/peering.go
Stebalien: (need to push)

peering/peering_test.go
Stebalien: Done.

peering/peering.go
Stebalien: Hm. You're right. I've added a proper cap (with a 10% jitter at the cap).

docs/config.md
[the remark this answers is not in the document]
Stebalien: But we can revisit this decision once the RC has shipped.

Stebalien:
@aschmahmann could you give this a once over post merge? I think it's good enough to go in the RC, but I'd like you to look at my changes.

docs/config.md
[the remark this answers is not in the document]
aschmahmann: Seems fine, I wasn't really sure what qualified as experimental vs not which is why I asked 😄

peering/peering_test.go
```diff
@@ -0,0 +1,158 @@
...
+		return h1.Network().Connectedness(h2.ID()) == network.Connected
+	}, 30*time.Second, 10*time.Millisecond)
+
+	// Now explicitly connect to p3.
+	require.NoError(t, h1.Connect(ctx, peer.AddrInfo{ID: h3.ID(), Addrs: h3.Addrs()}))
+	require.Eventually(t, func() bool {
+		return h1.Network().Connectedness(h2.ID()) == network.Connected
+	}, 30*time.Second, 100*time.Millisecond)
+
+	require.Len(t, h1.Network().Peers(), 3)
```
aschmahmann, line 70: I don't know if it's worth doing anything about this, but noting that there's an unlikely race with the connmgr here where we prune in between connecting and doing the size check.

Stebalien: I set the high water to 100 so that shouldn't happen.

PR #7365, opened 2020-05-26
feat: add noise support

Stebalien:
We can add a proper configuration option in a followup patch after the RC.

### DIFF
--- a/core/node/libp2p/transport.go
+++ b/core/node/libp2p/transport.go
@@ -1,14 +1,20 @@
 package libp2p
 
 import (
+	"fmt"
+
 	"github.com/libp2p/go-libp2p"
 	metrics "github.com/libp2p/go-libp2p-core/metrics"
+	noise "github.com/libp2p/go-libp2p-noise"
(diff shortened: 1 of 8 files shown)


core/node/libp2p/transport.go
```diff
@@ -21,7 +27,7 @@ func Transports(pnet struct {
 	return opts
 }
 
-func Security(enabled bool) interface{} {
+func Security(enabled bool, securityTransportOverride []string) interface{} {
```
aarshkshah1992, line 30: @Stebalien Why have we added the `securityTransportOverride ` option ? Is there any use case for it right now ?

Stebalien: At the moment, go-ipfs prefers TLS, then SECIO, then Noise. That means the noise transport will _never_ be used by default. The override option lets us actually _use_ noise by:

* Putting it before TLS.
* Removing SECIO.
* Etc...

See the tests and the new experimental-features documentation for details. If it doesn't make sense, I'll expand the documentation in a followup PR.

Issue #7260, opened 2020-04-30
Use go-multiaddr-dns DefaultResolver instead of std lib Resolver


cpacia:
Would be nice to swap out the `net.LookupTXT` in the dns resolver for the `madns.DefaultResolver.LookupTXT` function.

This would allow one to just override the `madns.DefaultResolver` and it would work for both libp2p and IPFS. As it stands now there is no way to configure this resolver. 

https://github.com/ipfs/go-ipfs/blob/f006fc164771d6ab2e2b2608992bc6e5d1f17a8f/namesys/dns.go#L28

matyapiro31:
Ref: #6915

hsanjuan:
@Stebalien this looks like an ok proposal unless I'm missing something.

Stebalien:
`madns.DefaultResolver` isn't quite right as madns and dnslink are two separate specs. Ideally we'd construct resolvers with the right dns resolver, but threading that through is difficult. We might need a global `go-dns` package.

lidel:
We are tracking configurable DNS servers (for all record types) in https://github.com/ipfs/go-ipfs/issues/6532

PR #7376, opened 2020-05-27
Add custom dns server address support(partialy)

matyapiro31:
This add an option of dns server address to `namesys.NewDNSResolver()` and `namesys.DNSResolver`.
I haven't add config of dns server now.


matyapiro31:
I don't know what causes this errors.

cpacia:
@matyapiro31 maybe fix it this way. https://github.com/ipfs/go-ipfs/issues/7260

Likely a better pattern as you only need to override in one place.

Issue #7378, opened 2020-05-27
"ipfs id" address format is inconsistent


Stebalien:
#### Version information:

> v0.5.1

#### Description:

`ipfs id QmSelf` prints out:

```
{
	"ID": "QmSelf",
	"PublicKey": "...",
	"Addresses": [
		"/ip4/104.131.131.82/udp/4001/quic/p2p/QmSelf",
		"/ip4/104.131.131.82/tcp/4001/p2p/QmSelf"
	],
	"AgentVersion": "go-ipfs/0.5.1/8431e2e",
	"ProtocolVersion": "ipfs/0.1.0"
}
```

`ipfs id QmOtherPeer` prints out:

```
{
	"ID": "QmOtherPeer",
	"PublicKey": "...",
	"Addresses": [
		"/ip4/104.131.131.82/udp/4001/quic", // note the lack of peer ID.
		"/ip4/104.131.131.82/tcp/4001"
	],
	"AgentVersion": "go-ipfs/0.5.1/8431e2e",
	"ProtocolVersion": "ipfs/0.1.0"
}
```

We should either drop the `/p2p/...` part from `ipfs id QmSelf`, or add it to `ipfs id QmOtherPeer`. Unfortunately, both solutions are breaking. I'd prefer the first solution because the `/p2p/Qm...` part is redundant. However, that option is _also_ the more breaking option.

Stebalien:
cc @achingbrain what does js-ipfs do?

achingbrain:
js-ipfs includes `/p2p` in the QmSelf addresses list but doesn't support args to the `jsipfs id` command right now.

Were it to, it'd just print out whatever js-libp2p gives us which would be similar to the output of `jsipfs swarm peers` so would most likely include the `/p2p`.

PR #7397, opened 2020-05-30
fix(commands): print consistent addresses in ipfs id

Stebalien:
Consistently append `/p2p/QmMyId` to addresses when calling `ipfs id Me` and `ipfs id NotMe`.

Fixes https://github.com/ipfs/go-ipfs/issues/7378.


Stebalien:
Well, the test is to fix the issue where `ipfs id` and `ipfs id remote_peer` were inconsistent. However, you're right. I don't think we have any tests that directly check if these addresses contain `/p2p`. I'll add another test.

PR #7408, opened 2020-06-03
github issue: add system version information section

RubenKelevra:


RubenKelevra:
I think it doesn't hurt to collect a bit more informations, different Kernels will use different network settings. An example would be ECN and TCP-fastopen defaults was changed in the past as well as TCP-timeouts.

I'm using the zen-Kernel for example, which uses the BBR congestion control algorithm from google by default.

Such informations might be helpful in some cases to trace down a network-related bug.

PR #7409, opened 2020-06-03
feat: return supported protocols in id output

Stebalien:
This is really useful for debugging.

TODO: fix tests if @aschmahmann and @achingbrain OK this change.

### DIFF
--- a/core/commands/id.go
+++ b/core/commands/id.go
@@ -97,15 +97,17 @@ EXAMPLE:
 			return errors.New(offlineIdErrorMessage)
 		}
 
-		p, err := n.Routing.FindPeer(req.Context, id)
-		if err == kb.ErrLookupFailure {
+		// We need to actually connect to run identify.
+		err = n.PeerHost.Connect(req.Context, peer.AddrInfo{ID: id})
+		switch err {
+		case nil:
+		case kb.ErrLookupFailure:
 			return errors.New(offlineIdErrorMessage)
-		}
-		if err != nil {
+		default:
 			return err
 		}
 
-		output, err := printPeer(n.Peerstore, p.ID)
+		output, err := printPeer(n.Peerstore, id)
 		if err != nil {
 			return err
 		}
--- a/core/commands/id.go
+++ b/core/commands/id.go
@@ -36,6 +36,7 @@ type IdOutput struct {
 	Addresses       []string
 	AgentVersion    string
 	ProtocolVersion string
+	Protocols       []string
 }
 
 const (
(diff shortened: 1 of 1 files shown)


jacobheun:
I was trying to do this the other day, this would be great to have support for.

core/commands/id.go
```diff
@@ -97,15 +99,17 @@ EXAMPLE:
 			return errors.New(offlineIdErrorMessage)
 		}
 
-		p, err := n.Routing.FindPeer(req.Context, id)
-		if err == kb.ErrLookupFailure {
+		// We need to actually connect to run identify.
+		err = n.PeerHost.Connect(req.Context, peer.AddrInfo{ID: id})
```
aschmahmann, line 103: This is a fine conceptual change. Just a note that this doesn't actually do anything as the DHT (our only PeerRouting system) will connect to the peer, and will not return any addresses if it fails. https://github.com/libp2p/go-libp2p-kad-dht/blob/fc3558cc35274f5d1727dad8201f821df1ee6317/routing.go#L675-L682

core/commands/id.go
```diff
@@ -163,6 +168,13 @@ func printPeer(ps pstore.Peerstore, p peer.ID) (interface{}, error) {
 	for _, a := range addrs {
 		info.Addresses = append(info.Addresses, a.String())
 	}
+	sort.Strings(info.Addresses)
```
aschmahmann, line 171: I added address sorting to this function, any objections?

core/commands/id.go
```diff
@@ -198,6 +210,9 @@ func printSelf(node *core.IpfsNode) (interface{}, error) {
 		for _, a := range addrs {
 			info.Addresses = append(info.Addresses, a.String())
 		}
+		sort.Strings(info.Addresses)
```
aschmahmann, line 213: I added address sorting to this function, any objections?

jacobheun:
> are we ok merging this change without the corresponding js-ipfs one, or do we need to block on that?

No need to block on JS, I created https://github.com/ipfs/js-ipfs/issues/3219 to track adding support there.

core/commands/id.go
Stebalien: Hm. I thought I had some trouble where we weren't actually _waiting_ for the connect event to finish.

But honestly, we should do this anyways. We shouldn't assume that the router will always end up connecting to the target peer.

core/commands/id.go
Stebalien: LGTM!

achingbrain:
This is being added to js in https://github.com/ipfs/js-ipfs/pull/3250

PR #7429, opened 2020-06-05
feat: nice errors when failing to load plugins

### LINKED ISSUE #7305: Prebuilt binary do not load plugin
#### Version information:

go-ipfs version: 0.5.1
Repo version: 9
System version: amd64/darwin
Golang version: go1.13.10

#### Description:

I made an IPLD plugin for my own usage, it is based on the [go-ipfs-example-plugin repo ](https://github.com/ipfs/go-ipfs-example-plugin) and I tune it so that it can run on IPFS v0.5.1

First, I made a local copy of go-ipfs v0.5.1 and made my plugin import this local copy. After build and install my own plugin to $IPFS_PATH/plugins and run by my build from source go-ipfs daemon,  everything works perfectly

However, when I tried to import the official go-ipfs v0.5.1 rather than my local copy and run by prebuilt binary of macOS 64bit binary, the plugin just did not loaded in, and even did not call the `Init` of the plugin

By the way, both build from source go-ipfs and prebuilt binary are using the same $IPFS_PATH, and share the same config file

Stebalien:
partially fixes #7305

The other half is to actually _build_ with CGO support by default.

### DIFF
--- a/plugin/loader/load_nocgo.go
+++ b/plugin/loader/load_nocgo.go
@@ -1,0 +1,18 @@
+// +build !cgo,!noplugin
+// +build linux darwin
+
+package loader
+
+import (
+	"errors"
+
+	iplugin "github.com/ipfs/go-ipfs/plugin"
+)
+
+func init() {
+	loadPluginFunc = nocgoLoadPlugin
+}
+
+func nocgoLoadPlugin(fi string) ([]iplugin.Plugin, error) {
+	return nil, errors.New("not built with cgo support")
+}
--- a/plugin/loader/load_unix.go
+++ b/plugin/loader/load_unix.go
@@ -1,56 +1,20 @@
-// +build !noplugin
-// +build linux,cgo darwin,cgo
+// +build cgo,!noplugin
+// +build linux darwin
 
 package loader
 
 import (
 	"errors"
-	"fmt"
-	"os"
-	"path/filepath"
 	"plugin"
 
 	iplugin "github.com/ipfs/go-ipfs/plugin"
 )
 
 func init() {
-	loadPluginsFunc = linuxLoadFunc
+	loadPluginFunc = unixLoadPlugin
 }
 
-func linuxLoadFunc(pluginDir string) ([]iplugin.Plugin, error) {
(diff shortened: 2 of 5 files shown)


plugin/loader/load_unix.go
```diff
@@ -5,52 +5,16 @@ package loader
...
-			return fmt.Errorf("loading plugin %s: %s", fi, err)
-		}
-		return nil
-	})
-
-	return plugins, err
-}
-
-func loadPlugin(fi string) ([]iplugin.Plugin, error) {
+func unixLoadPlugin(fi string) ([]iplugin.Plugin, error) {
```
Stebalien, line 17: Before, we wouldn't try loading plugins on platforms that didn't support plugins.

Now, we try loading plugins everywhere, but return an error if we find a plugin that we can't load (either because the platform is unsupported, or because something is wrong with the plugin).

Stebalien:
bump @dirkmc

M3T4F34R:
fix/plugin-load-error

plugin/loader/load_nocgo.go
```diff
@@ -0,0 +1,17 @@
+// +build linux,!cgo,!noplugin darwin,!cgo,!noplugin
```
aschmahmann, line 1:
```suggestion
// +build !cgo,!noplugin
// +build linux darwin
```

nit: This might be a little easier to look at. Would also make the corresponding change in `plugin/loader/load_unix.go` since it makes it easier to visually verify that only one of `load_nocgo, load_noplugin, load_unix` actually are used.

Stebalien: Ah, nice. That's much cleaner.

Stebalien:
@aschmahmann it was just waiting on a review.

plugin/loader/load_nocgo.go
```diff
@@ -0,0 +1,18 @@
+// +build !cgo,!noplugin !cgo,!noplugin
```
aschmahmann, line 1:
```suggestion
// +build !cgo,!noplugin
```

Issue #7357, opened 2020-05-25
Switch to base36 by default in all text output (overriding ipfs/go-ipfs/issues/4143 )


ribasushi:
Now that [base36 support](https://github.com/multiformats/multibase/blob/3806057e6f6c09c7cc0a6dc14de22fa50c9e0b51/multibase.csv#L17-L19) is landing into go-ipfs [for unrelated reasons](https://github.com/ipfs/go-ipfs/issues/6916) I would like to resurrect [my pitch from 3 years ago](https://github.com/ipfs/go-ipfs/issues/4143#issuecomment-326410468) to default to `base36` rendering as opposed to `base32`. The internals of go-ipfs operate over binary CIDs so the proposed change is strictly concerned with "UX".

Let me know your thoughts!

✅ Pros
- Consistency across the board: libp2p-key and unixfs-entry pointers all look like what the wider public knows as "a CID"
- Brings "semi-avalanche effect" to identity CIDs. E.g.

| | | |
| --- | --- | --- |
| Data: | `sufficiently long payload` | `sufficient1y long payload` |
| B32Cid: | `bafkqagttovtgm2ldnfsw45dmpeqgy33om4qhaylznrxwczak` | `bafkqagttovtgm2ldnfsw45brpeqgy33om4qhaylznrxwczak` |
| B36Cid: | `kumq0v1xsl33xy9ospzfmf6uuibi0jxtdsrmhovipx6o0a` | `kumq0v1xsl33xy9ospzfmbp3zrpszmf3tbmvd50ou83pje`|

- Increases the usable size of an identity-cid from 34 to 36 bytes of "blocksize" while still usable in subdomains (max 63 characters): `1 + ceil( (4+blocksize) * log(256) / log(base32-or-36) )`

❌ Cons
- Early adopters are already familiar with `bafy...`, another switch would [create further confusion](https://discuss.ipfs.io/t/why-is-ipfs-url-different-now/8053)
- Somewhat slower ( on my macbook I can encode and decode  `( 1 * 1000 * 1000 * 1000 ) / 2672 = 374251` b36 CIDs in 1 second ):
```
go-multibase$ go test -count=1 ./... -cpu=1 -bench . -benchmem
goos: darwin
goarch: amd64
pkg: github.com/multiformats/go-multibase
BenchmarkRoundTrip/base32         	 2190152	       541 ns/op	     256 B/op	       4 allocs/op
[...]

Issue #6916, opened 2020-02-20
Switch to ed25519 keys by default so that we can reasonably store signed content records


jacobheun:
*To be confirmed*:
- Add support for specifying key type in `ipfs init`
- Switch initialization to use ed25519 as the default instead of rsa

## Success Criteria
- Newly created IPFS nodes will use ed25519 keys by default
- Users can rotate their existing keys to ed25519

## What's Left
Mandatory:
- [x] Key rotation https://github.com/ipfs/go-ipfs/pull/7515
- [x] Sharness test bug https://github.com/ipfs/go-ipfs/pull/7542
- [x] Having IPNS keys (in the CLI, everywhere?) output base36 by default https://github.com/ipfs/go-ipfs/pull/7554

Nice to Have:
- [ ] go-multiaddr support for CIDs as PeerIDs https://github.com/libp2p/go-libp2p/issues/976
- [ ] perhaps have an option for switching peerID outputs everywhere to base36 - this can be deferred

RubenKelevra:
Will `self` also be changed to ed25519?

willscott:
> Users can rotate their existing keys to ed25519
What does this mean, exactly?
It's easy enough to generate a new ed25519 key. is there / does there need to be a way to indicate that identity transition to peers? Enumerating what needs to change in a rotation would be useful.

Stebalien:
It means #5880. That is, users need to be able to "re-key" their node without changing anything else.

Stebalien:
@RubenKelevra yes. `self` _is_ the node's ID. Note: we _won't_ automatically change user keys, we're just (a) providing an easy way to do this and (b) changing the defaults for newly initialized nodes.

RubenKelevra:
> @RubenKelevra yes. `self` _is_ the node's ID. Note: we _won't_ automatically change user keys, we're just (a) providing an easy way to do this and (b) changing the defaults for newly initialized nodes.

Will this be backwards compatible? I mean, can for example 0.4.23 handle nodes with ed25519 keys today?

Stebalien:
Yes.

lidel:
Note to self: when we switch to ed25519 by default, we should ensure that keys are always represented as CIDv1 in Base32 with libp2p multicodec (see [RFC/0001-text-peerid-cid.md](https://github.com/libp2p/specs/blob/master/RFC/0001-text-peerid-cid.md)).

[...]

PR #7441, opened 2020-06-08
feat: support ED25519 at subdomain gw with TLS

lidel:
> The goal of this PR is to be alternative to #7358 and solve #7318 in a way that makes it possible to have functional TLS for IPNS on public subdomain gateways.

### Rationale 

If we split original CID into multiple sub labels (#7358) DNS resolution will work, but **no public gateway will be able to get TLS certificate** that supports double wildcards like `*.*.ipfs.example.com`. 

When splitting is used, visitors using web browsers will see a [TLS Error page](https://user-images.githubusercontent.com/157609/86123445-cc829180-bad9-11ea-8471-83683ee48654.png), which effectively kills UX and defeats the purpose of subdomain gateway (one step forward  by increasing security by providing Origin isolation, two steps backward by removing ability to do TLS validation when loading website backed by a long CID).

Given this reality, I believe a better trade off could be to try shortening CID by converting it to Base36, and if that fails return an error before redirect to subdomain occurs.

### This PR

- adds subdomain gateway support for ED25519 CIDs in a way that fits in a single DNS label to enable TLS for every IPNS website.
- cleans up subdomain redirect logic and adds more explicit error handling so user gets human-readable error with proper IPFS context instead of generic DNS failure 

### TL;DR on router logic:

When CID is longer than 63 characters, router at /ipfs/* and /ipns/* converts to Base36, and if that does not help, returns a human readable 400 Bad Request error.

### Examples

Given a subdomain gateway at `example.com`:

ED25519 fits perfectly when converted to Base36, so it gets loaded just fine:
```
http://example.com/ipns/12D3KooWP3ggTJV8LGckDHc4bVyXGhEWuBskoFyE6Rn2BJBqJtpa 
  → HTTP301 – http://k51qzi5uqu5dl2yn0d6xu8q5aqa61jh8zeyixz9tsju80n15ssiyew48912c63.ipns.example.com
```

CID produced by `ipfs add --cid-version 1 --hash sha2-512` is too long, so an error is returned instead:

```
http://example.com/ipfs/bafkrgqhhyivzstcz3hhswshfjgy6ertgmnqeleynhwt4dlfsthi4hn7zgh4uvlsb5xncykzapi3ocd4lzogukir6ksdy6wzrnz6ohnv4aglcs 
  → HTTP 400 "CID incompatible with DNS label length limit of 63"
```

Note that returning this error overrides https://github.com/ipfs/go-ipfs/pull/6982 which means tools like `curl` won't be able to load content from long CIDs via the subdomain gateway. I don't think that is a bug, pretty sure it is how we want it, but mentioning here for completenes. 

### Why we dropped the previous iteration

We looked into automatically replacing root node which has shorter CID, but as noted in https://github.com/ipfs/go-ipfs/pull/7441#issuecomment-640686006 its not worth it.

<details>
  <summary>Click here to expand notes from that older version</summary>

> ## **Note:** This is WIP – don't look at the code yet, opening draft PR for early feedback on the idea

### TL;DR

The idea evaluated here is to redirect `example.com/ipfs/$CID` to `$dnsCID.ipfs.example.com` where `$dnsCID` is a new CID created with a hash function known to fit in a single DNS label (<=63chars).

This way public gateways can **enable HTTPS without TLS errors for content roots backed by longer hash functions**.

### Rationale 

If we split original CID into multiple sub labels (#7358) DNS resolution will work, but **no public gateway will be able to get TLS certificate** that supports double wildcards like `*.*.ipfs.example.com`. This means [errors like this](https://blog.almonit.eth.link#x-ipfs-companion-no-redirect).

This means when splitting is used, visitors using web browsers will see a TLS Error page, which effectively kills UX and defeats the purpose of subdomain gateway (one step forward  by increasing security by providing Origin isolation, two steps backward by removing ability to do TLS validation when loading website backed by a long CID).

Given this reality, I believe a better trade off could be to create a new root CID that is compatible with DNS and make websites work in existing user agents that way.

Thoughts?

### Open questions

- Is this worth it?
- **Should we redirect to a warning page, that explains original CID can't be loaded over DNS, and gives user a choce if they want to rehash and proceed to "new CID" (..or open it from a well-known non-subdomain gateway?)** 
- Which API to use? How to ensure newly minted CID is provided to the network?
- Should we do the same from IPNS? (un-inlining ed25519) or do we want to opportunistically use Base36 there instead?
- Should we keep support for split CIDs, and redirect them to a new CID that fits in a single label?

</details>

### DIFF
--- a/core/corehttp/hostname.go
+++ b/core/corehttp/hostname.go
@@ -41,12 +41,15 @@ var defaultKnownGateways = map[string]config.GatewaySpec{
 	"dweb.link":       subdomainGatewaySpec,
 }
 
+// Label's max length in DNS (https://tools.ietf.org/html/rfc1034#page-7)
+const dnsLabelMaxLength int = 63
+
 // HostnameOption rewrites an incoming request based on the Host header.
 func HostnameOption() ServeOption {
 	return func(n *core.IpfsNode, _ net.Listener, mux *http.ServeMux) (*http.ServeMux, error) {
 		childMux := http.NewServeMux()
 
-		coreApi, err := coreapi.NewCoreAPI(n)
+		coreAPI, err := coreapi.NewCoreAPI(n)
 		if err != nil {
 			return nil, err
 		}
@@ -101,7 +104,12 @@ func HostnameOption() ServeOption {
 					if gw.UseSubdomains {
 						// Yes, redirect if applicable
 						// Example: dweb.link/ipfs/{cid} → {cid}.ipfs.dweb.link
-						if newURL, ok := toSubdomainURL(host, r.URL.Path, r); ok {
+						newURL, err := toSubdomainURL(host, r.URL.Path, r)
+						if err != nil {
+							http.Error(w, err.Error(), http.StatusBadRequest)
+							return
+						}
+						if newURL != "" {
 							// Just to be sure single Origin can't be abused in
 							// web browsers that ignored the redirect for some
 							// reason, Clear-Site-Data header clears browsing
@@ -131,7 +139,7 @@ func HostnameOption() ServeOption {
 				// Not a whitelisted path
 
 				// Try DNSLink, if it was not explicitly disabled for the hostname
-				if !gw.NoDNSLink && isDNSLinkRequest(r.Context(), coreApi, host) {
+				if !gw.NoDNSLink && isDNSLinkRequest(r.Context(), coreAPI, host) {
 					// rewrite path and handle as DNSLink
 					r.URL.Path = "/ipns/" + stripPort(host) + r.URL.Path
 					childMux.ServeHTTP(w, r)
@@ -158,14 +166,42 @@ func HostnameOption() ServeOption {
 					return
 				}
 
-				// Do we need to fix multicodec in PeerID represented as CIDv1?
-				if isPeerIDNamespace(ns) {
-					keyCid, err := cid.Decode(rootID)
-					if err == nil && keyCid.Type() != cid.Libp2pKey {
-						if newURL, ok := toSubdomainURL(hostname, pathPrefix+r.URL.Path, r); ok {
-							// Redirect to CID fixed inside of toSubdomainURL()
+				// Check if rootID is a valid CID
+				if rootCID, err := cid.Decode(rootID); err == nil {
+					// Do we need to redirect root CID to a canonical DNS representation?
+					dnsCID, err := toDNSPrefix(rootID, rootCID)
+					if err != nil {
+						http.Error(w, err.Error(), http.StatusBadRequest)
+						return
+					}
+					if !strings.HasPrefix(r.Host, dnsCID) {
+						dnsPrefix := "/" + ns + "/" + dnsCID
+						newURL, err := toSubdomainURL(hostname, dnsPrefix+r.URL.Path, r)
+						if err != nil {
+							http.Error(w, err.Error(), http.StatusBadRequest)
+							return
+						}
+						if newURL != "" {
+							// Redirect to deterministic CID to ensure CID
+							// always gets the same Origin on the web
 							http.Redirect(w, r, newURL, http.StatusMovedPermanently)
 							return
+						}
+					}
+
+					// Do we need to fix multicodec in PeerID represented as CIDv1?
+					if isPeerIDNamespace(ns) {
+						if rootCID.Type() != cid.Libp2pKey {
+							newURL, err := toSubdomainURL(hostname, pathPrefix+r.URL.Path, r)
+							if err != nil {
+								http.Error(w, err.Error(), http.StatusBadRequest)
+								return
+							}
+							if newURL != "" {
+								// Redirect to CID fixed inside of toSubdomainURL()
+								http.Redirect(w, r, newURL, http.StatusMovedPermanently)
+								return
+							}
 						}
 					}
 				}
@@ -183,7 +219,7 @@ func HostnameOption() ServeOption {
 			// 1. is wildcard DNSLink enabled (Gateway.NoDNSLink=false)?
 			// 2. does Host header include a fully qualified domain name (FQDN)?
 			// 3. does DNSLink record exist in DNS?
-			if !cfg.Gateway.NoDNSLink && isDNSLinkRequest(r.Context(), coreApi, host) {
+			if !cfg.Gateway.NoDNSLink && isDNSLinkRequest(r.Context(), coreAPI, host) {
 				// rewrite path and handle as DNSLink
 				r.URL.Path = "/ipns/" + stripPort(host) + r.URL.Path
 				childMux.ServeHTTP(w, r)
@@ -273,18 +309,38 @@ func isPeerIDNamespace(ns string) bool {
 	}
 }
 
+// Converts an identifier to DNS-safe representation that fits in 63 characters
+func toDNSPrefix(rootID string, rootCID cid.Cid) (prefix string, err error) {
+	// Return as-is if things fit
+	if len(rootID) <= dnsLabelMaxLength {
+		return rootID, nil
+	}
+
+	// Convert to Base36 and see if that helped
+	rootID, err = cid.NewCidV1(rootCID.Type(), rootCID.Hash()).StringOfBase(mbase.Base36)
+	if err != nil {
+		return "", err
+	}
+	if len(rootID) <= dnsLabelMaxLength {
+		return rootID, nil
+	}
+
+	// Can't win with DNS at this point, return error
+	return "", fmt.Errorf("CID incompatible with DNS label length limit of 63: %s", rootID)
+}
+
 // Converts a hostname/path to a subdomain-based URL, if applicable.
-func toSubdomainURL(hostname, path string, r *http.Request) (redirURL string, ok bool) {
+func toSubdomainURL(hostname, path string, r *http.Request) (redirURL string, err error) {
 	var scheme, ns, rootID, rest string
 
 	query := r.URL.RawQuery
 	parts := strings.SplitN(path, "/", 4)
-	safeRedirectURL := func(in string) (out string, ok bool) {
+	safeRedirectURL := func(in string) (out string, err error) {
 		safeURI, err := url.ParseRequestURI(in)
 		if err != nil {
-			return "", false
+			return "", err
 		}
-		return safeURI.String(), true
+		return safeURI.String(), nil
 	}
 
 	// Support X-Forwarded-Proto if added by a reverse proxy
@@ -304,11 +360,11 @@ func toSubdomainURL(hostname, path string, r *http.Request) (redirURL string, ok
 		ns = parts[1]
 		rootID = parts[2]
 	default:
-		return "", false
+		return "", nil
 	}
 
 	if !isSubdomainNamespace(ns) {
-		return "", false
+		return "", nil
 	}
 
 	// add prefix if query is present
@@ -327,25 +383,42 @@ func toSubdomainURL(hostname, path string, r *http.Request) (redirURL string, ok
 	}
 
 	// If rootID is a CID, ensure it uses DNS-friendly text representation
-	if rootCid, err := cid.Decode(rootID); err == nil {
-		multicodec := rootCid.Type()
+	if rootCID, err := cid.Decode(rootID); err == nil {
+		multicodec := rootCID.Type()
+		var base mbase.Encoding = mbase.Base32
 
-		// PeerIDs represented as CIDv1 are expected to have libp2p-key
-		// multicodec (https://github.com/libp2p/specs/pull/209).
-		// We ease the transition by fixing multicodec on the fly:
-		// https://github.com/ipfs/go-ipfs/issues/5287#issuecomment-492163929
-		if isPeerIDNamespace(ns) && multicodec != cid.Libp2pKey {
-			multicodec = cid.Libp2pKey
+		// Normalizations specific to /ipns/{libp2p-key}
+		if isPeerIDNamespace(ns) {
+			// Using Base36 for /ipns/ for consistency
+			// Context: https://github.com/ipfs/go-ipfs/pull/7441#discussion_r452372828
+			base = mbase.Base36
+
+			// PeerIDs represented as CIDv1 are expected to have libp2p-key
+			// multicodec (https://github.com/libp2p/specs/pull/209).
+			// We ease the transition by fixing multicodec on the fly:
+			// https://github.com/ipfs/go-ipfs/issues/5287#issuecomment-492163929
+			if multicodec != cid.Libp2pKey {
+				multicodec = cid.Libp2pKey
+			}
 		}
 
-		// if object turns out to be a valid CID,
-		// ensure text representation used in subdomain is CIDv1 in Base32
-		// https://github.com/ipfs/in-web-browsers/issues/89
-		rootID, err = cid.NewCidV1(multicodec, rootCid.Hash()).StringOfBase(mbase.Base32)
+		// Ensure CID text representation used in subdomain is compatible
+		// with the way DNS and URIs are implemented in user agents.
+		//
+		// 1. Switch to CIDv1 and enable case-insensitive Base encoding
+		//    to avoid issues when user agent force-lowercases the hostname
+		//    before making the request
+		//    (https://github.com/ipfs/in-web-browsers/issues/89)
+		rootCID = cid.NewCidV1(multicodec, rootCID.Hash())
+		rootID, err = rootCID.StringOfBase(base)
 		if err != nil {
-			// should not error, but if it does, its clealy not possible to
-			// produce a subdomain URL
-			return "", false
+			return "", err
+		}
+		// 2. Make sure CID fits in a DNS label, adjust encoding if needed
+		//    (https://github.com/ipfs/go-ipfs/issues/7318)
+		rootID, err = toDNSPrefix(rootID, rootCID)
+		if err != nil {
+			return "", err
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


core/corehttp/hostname.go
```diff
@@ -266,6 +283,19 @@ func isPeerIDNamespace(ns string) bool {
...
 }
 
+// Converts an identifier to DNS-safe representation that fits in 63 characters
+func toDNSPrefix(ctx context.Context, api iface.CoreAPI, id string) (prefix string) {
+	// Return as-is if things fit
+	if len(id) <= dnsLabelMaxLength {
+		return id
+	}
+
+	// TODO: get original root
```
lidel, line 293: Which API would be the best one to get original root node, 
if we fetch it with intention to re-hash using different hash function, while keeping everything else intact (codec, links, metadata etc)

core/corehttp/hostname.go
```diff
@@ -266,6 +283,19 @@ func isPeerIDNamespace(ns string) bool {
...
+	// TODO: create a new CID, representing a dns-safe root
```
lidel, line 294: Which APIs would be the most efficient way to create a new unixfs block?
How to ensure its persisted in the datastore?

ribasushi:
( have not looked at the code )

I think there's a fundamental flaw with the approach, namely:

> redirect example.com/ipfs/$CID to $dnsCID.ipfs.example.com where $dnsCID is a new CID created with a hash function known to fit

You are essentially reducing the security of the chain to the attack-resistance of "hash function known to fit", thus entirely obviating the case for using a hash function with >256 bits of output. (you also introduce a double-publish step which adds potentially enormous complexity to ipfs-lite implementations)

I think we should first explore what are practical use cases of using a hash function with long output, and whether substituting another hash in its place violates the original assumptions.

Note - it is important to make a distinction between the size of a function's state, and the size of the digest used to construct a CID. E.g. sha3-512 is the only widely used hash function with an [internal state capacity larger than its block size](https://en.wikipedia.org/wiki/SHA-3#Instances) which **would not be reduced** by taking the first 256 bits of its output in post-processing.

core/corehttp/hostname.go
Stebalien: `api.ResolveNode`

core/corehttp/hostname.go
Stebalien: Hm. This is a bit tricky given just the CoreAPI. I guess it would be best to use the `api.Blocks().Put(...)` API for now, but that doesn't work cleanly with native types.

ribasushi:
@lidel I love this last iteration! :shipit:
Though this makes the case for https://github.com/ipfs/go-ipfs/issues/7357 even stronger ( reduce the "why is my hash different?!?!" questions )

core/corehttp/hostname.go
```diff
@@ -334,11 +390,16 @@ func toSubdomainURL(hostname, path string, r *http.Request) (redirURL string, ok
 		// if object turns out to be a valid CID,
 		// ensure text representation used in subdomain is CIDv1 in Base32
 		// https://github.com/ipfs/in-web-browsers/issues/89
-		rootID, err = cid.NewCidV1(multicodec, rootCid.Hash()).StringOfBase(mbase.Base32)
+		rootCID = cid.NewCidV1(multicodec, rootCID.Hash())
+		rootID, err = rootCID.StringOfBase(mbase.Base32)
```
lidel, line 394: Note: we still default to Base32 here..

core/corehttp/hostname.go
```diff
@@ -334,11 +390,16 @@ func toSubdomainURL(hostname, path string, r *http.Request) (redirURL string, ok
...
+		if err != nil {
+			return "", err
+		}
+
+		// make sure CID fits in DNS label
+		rootID, err = toDNSPrefix(rootID, rootCID)
```
lidel, line 400: ..but if Base32 was too long, Base36 will be tested inside `toDNSPrefix`

test/sharness/t0114-gateway-subdomains.sh
```diff
@@ -480,6 +480,66 @@ test_hostname_gateway_response_should_contain \
   "http://127.0.0.1:$GWAY_PORT" \
   "$CID_VAL"
 
+## Test subdomain handling of CIDs that do not fit in a single DNS Label (>63chars)
+## https://github.com/ipfs/go-ipfs/issues/7318
+## ============================================================================
+
+# TODO: replace with cidv1
```
lidel, line 487: Right now `ipfs key gen -t ed25519 ed25519-test-key` outputs libp2p-key as Base58btc Multihash (thats why we hardcoded `12D..` below and not `bafy..`).

It is unclear if go-ipfs will switch text representation of libp2p-keys to CIDv1 in Base36 as part of switch to ed25519 (https://github.com/ipfs/go-ipfs/issues/6916), so for now we test Base58btc.

lidel:
Finished tests (go&sharness), and updated PR description to reflect current version.
CI is green – ready for review / comments.

cc @autonome @gozala for visibility as this impacts behavior of subdomain gateways in the browser

core/corehttp/hostname.go
aschmahmann: Is the idea that even if we changed the default libp2p-key output to be base36 that people who put base32 RSA libp2p-keys into the browser domain or path wouldn't end up with the domain (and therefore origin) changed?

Similarly, is this solution too flexible? Do we need a canonical base encoding for the domains so that data stored in browser local storage isn't lost?

core/corehttp/hostname.go
```diff
@@ -151,16 +159,44 @@ func HostnameOption() ServeOption {
...
-				// Do we need to fix multicodec in PeerID represented as CIDv1?
-				if isPeerIDNamespace(ns) {
-					keyCid, err := cid.Decode(rootID)
-					if err == nil && keyCid.Type() != cid.Libp2pKey {
-						if newURL, ok := toSubdomainURL(hostname, pathPrefix+r.URL.Path, r); ok {
-							// Redirect to CID fixed inside of toSubdomainURL()
+				// Check if rootID is a valid CID
+				if rootCID, err := cid.Decode(rootID); err == nil {
+					// Do we need to redirect root CID to a canonical DNS representation?
+					dnsCID, err := toDNSPrefix(rootID, rootCID)
```
aschmahmann, line 165: This can convert some IPFS CIDs that are a bit too big into base36. Was that intended?

core/corehttp/hostname.go
```diff
@@ -151,16 +159,44 @@ func HostnameOption() ServeOption {
...
+					if err != nil {
+						http.Error(w, err.Error(), http.StatusBadRequest)
+						return
+					}
+					if !strings.HasPrefix(r.Host, dnsCID) {
```
aschmahmann, line 170: What case is this meant to deal with?

core/corehttp/hostname.go
lidel: Unsure what you mean, `toDNSPrefix` will:
1. Convert ed25519 to Base36 and return no error (because Base36 version fits in DNS label)
2. Return error if conversion to Base36 did not help and resulting CID is still too long

core/corehttp/hostname.go
```diff
@@ -151,16 +159,44 @@ func HostnameOption() ServeOption {
...
-				if isPeerIDNamespace(ns) {
-					keyCid, err := cid.Decode(rootID)
-					if err == nil && keyCid.Type() != cid.Libp2pKey {
-						if newURL, ok := toSubdomainURL(hostname, pathPrefix+r.URL.Path, r); ok {
-							// Redirect to CID fixed inside of toSubdomainURL()
+				// Check if rootID is a valid CID
+				if rootCID, err := cid.Decode(rootID); err == nil {
+					// Do we need to redirect root CID to a canonical DNS representation?
+					dnsCID, err := toDNSPrefix(rootID, rootCID)
+					if err != nil {
```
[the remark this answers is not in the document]
lidel: This is purely because of Security Model in modern browsers.

**tl;dr** We can't have path-based and subdowmain-based gateways **on the same hostname** because that would defeat the purpose of Origin isolation provided by the subdomain gw. Namely, `example.com/ipfs/{long-cid}/index.html` could manage permissions/cookies/storage of every CID loaded  under `*.ipfs.example.com` 


Refs.
- https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy
- https://web.dev/same-origin-policy/

---

That being said, we could modify behavior proposed in this PR and give a configuration option to **redirect to path-based gateway on other hostname** instead of returning Error 400 Bad Request. But that is something we should do in separate PR, I believe.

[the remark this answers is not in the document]
aschmahmann: Ah ok, seems reasonable. Although I guess if we wanted to we could reserve a subdomain so we have `cidtoolong.ipfs.example.com/{long-cid}/index.html`. You're right that we could do this in a subsequent PR though

core/corehttp/hostname.go
aschmahmann: Perhaps I'm confused. `toDNSPrefix` looks like 

```
	// Return as-is if things fit
	if len(rootID) <= dnsLabelMaxLength {
		return rootID, nil
	}

	// Convert to Base36 and see if that helped
	rootID, err = cid.NewCidV1(rootCID.Type(), rootCID.Hash()).StringOfBase(mbase.Base36)
	if err != nil {
		return "", err
	}
	if len(rootID) <= dnsLabelMaxLength {
		return rootID, nil
	}
```

None of this refers to Ed25519 keys or IPNS keys in general, we're just converting whatever CID we get into base36. Couldn't we have a long IPFS CID that is around the same length as the libp2p-key + identity multihash for Ed25519 keys? For example, `echo "111111111.111111111.111111" | ipfs add --inline --inline-limit=36` outputs a 65 character base32 CID. However, it'll get converted into a 62 character base 36 CID and rendered on the gateway.

My question was if this behavior is intentional

core/corehttp/hostname.go
lidel: This is a failsafe to ensure that if (for some reason) requested CID is not the same as DNS-safe-version we redirect to DNS-safe-version.

This could occur if user opened `{ed25519-in-base32}.ipns.example.com` in a browser that uses Gateway port as HTTP Proxy. In that mode DNS resolution does not fail due to the 63char limit, because hostname is passed to HTTP Proxy (go-ipfs) as an opaque string (no DNS lookup), so entire thing will load just fine.

Why we care about normalization in this edge case? 

- Someone could copy `{ed25519-in-base32}` from address bar and share with a friend that does not use gateway in proxy mode, and that would produce DNS failure. By redirecting to `{ed25519-in-base36}` version we ensure consistency of Origins across IPNS websites loaded form the subdomain gw, including proxy modes.
- We use proxy mode in Firefox because of https://bugzilla.mozilla.org/show_bug.cgi?id=1220810 so this edge case could impact some users of IPFS Companion.

core/corehttp/hostname.go
lidel: Ah, yes, this is on purpose.
I explicitly avoided code that works only for specific codec, `toDNSPrefix` tries to fix all long CIDs.

AFAIK the switch to Base36 helps only for ED25519, but as you noted, something else could be right around the limit and I see no reason why it should not benefit from this as well.

core/corehttp/hostname.go
lidel: The idea here is to PR a surgical change to **use Base36 only where it is really needed and continue with Base32 as default for subdomains** (so already existing subdomain Origins don't change, and we don't need to worry about any data lost).

I don't believe we need to add any additional normalization.

aschmahmann: My question was basically about what we want to happen if I use base36 where I could've used base32.

Leaving as is may lead to me having two different views (with different local data) for the same website.

Making them canonical is both a bit of a pain and uses an arbitrary rule that we likely end up stuck with (like the Cidv0-1 size cutoff).

Wanted to double check we are ok with the pros and cons of our solution not

aschmahmann: My question was basically about what we want to happen if I use base36 where I could've used base32.

Leaving as is may lead to me having two different views (with different local data) for the same website.

Making them canonical is both a bit of a pain and uses an arbitrary rule that we likely end up stuck with (like the Cidv0-1 size cutoff).

Wanted to double check we are ok with the pros and cons of our solution not

lidel: I think RSA in Base32 is ok for now (when RSA is the default and CLI output is in Base58btc), use of Base36 is still non-existent. 

My thinking was that when we switch libp2p-key in CLI output to CIDv1 as part of ed25519 epic (https://github.com/ipfs/go-ipfs/issues/6916), then we will make a decision about Base36 being "canonical" base for `/ipns/` and then may want to adjust this section to use the same Base36 for all CID libp2p-keys (RSA and ED25519)

But as long RSA is the default and Base36 is fairly new, I'd rather not rock the boat prematurely.

ribasushi: > Yes, the idea here is to PR a surgical change to use Base36 only where it is really needed and continue with Base32 as default for subdomains (so already existing subdomain Origins don't change, and we don't need to worry about any data lost).

I really think this is the wrong tradeoff. The amount of "data lost" right now is still absolutely minimal, compared to what is in store if we end up needing to switch to b36.

@aschmahmann is correct: if you do not bite the bullet ans make the switch now, you are stuck with yet another special case forever†

† forever, n: an excruciatingly long period time

aschmahmann: @ribasushi @lidel I suspect for IPNS if we switched to base36 everywhere probably no one would mind since those users are probably keeping updated so that they can get IPNS performance improvements as they come. However, how bad is the damage for IPFS users in subdomains? We pushed that out in 0.5 so it's still pretty new, but there are presumably some users already relying on it.

It seems like we might be able to get away with just b36 encoding all the subdomains to avoid worrying about:
1) Canonical encodings: What happens if a user uses a b36 subdomain for something that could have been b32? Won't it be weird that the state of those two websites could be different?
2) Annoying rules: We can solve 1, by just coming up with some canonicalization like "it's b32 unless it won't fit then it's b36", but that may be annoying to think about longer term compared with just deciding to use b36 everywhere.

lidel: Ad. Canonical encodings: IMO not an issue in practice. The same HTML can be loaded from different hostnames on the web, and each gets distinct Origin. I think its a feature of IPFS to be able to represent the same CID with different base, we should not force people to use b36 if their CID works with b32 and they prefer it for some reason and explicitly used that in subdomain (eg. third party tooling implements RFC version of b32, and no b36 yet).

----

Just to be clear: I am all for simplicity of using b36 for entire `/ipns/` (including RSA).

My only concern was that **we've just added b36 in the last release** and if we flip the switch and use it for all libp2p-keys on subdomains, and people start copying and pasting those CIDs, various things will break for people: mind that there is no support for b36 in non-go non-js IPFS-related libs and clients at all etc, and we have no bandwidth in Q3 to help with changing that.

That is why I wanted to keep status quo for at least one or two more releases, and effectively only use b36 for ed25519 here.

That being said, 
@ribasushi @aschmahmann if you both are ok with using b36 for entire `/ipns/` (including RSA), I can update this PR to do just that – let me know.

ribasushi: Formal :+1: from me on flipping **as much as possible** to `b36`. I fully recognize the pain it **will** cause. That pain is utterly dwarfed by the cost of doing another switch further down the road.

Stebalien: I'd like to be as consistent as possible. Given that IPNS isn't widely used for obvious reasons anyways, I'd suggest at least using base36 for IPNS. I'm fine not canonicalizing CIDs.

> The same HTML can be loaded from different hostnames on the web, and each gets distinct Origin.

That's fine. The problem is if we decide to start canonicalizing CIDs to base36 for consistency in the future. If we do that, we'll be _changing_ the origin.

aschmahmann: :+1: I tend to think that we might as well do the b36 encoding for IPNS. IPNS users tend to be pretty happy when fixes show up and there'll probably be at least a few fixes in 0.7 or 0.8 that make it worth their while.

As an aside I've been thinking about this a bit and I suspect the confusion here stems from us not really telling people what the "correct" way is for users to describe an IPFS/IPNS URL. Do we even have a single answer here, or is it context dependent? For example, we'd like people to ideally use `ipfs://b32encodedID`, but since `ipfs://` isn't yet widely supported we probably like `dweb.link/ipfs/b32encodedID`. Not having an canonical answer means we're stuck guessing about how people "could" be using URLs and so makes reasoning about upgrade paths difficult.

lidel: Great! I'm updated this PR to use b35 for ALL libp2p-keys in subdomains (79d571f272d640f0d4164f6f9d94c5fb44983822)

@aschmahmann I believe that was the last thing blocking this PR – mind taking a final look and merging if it is ok?

ps. regarding canonical address, for the time being it is: `https://dweb.link/ipfs/{cid}` | `dweb.link/ipns/{libp2p-key}`
Works everywhere, can be shared and loads fine, and is future proof. 
We need to work with gui/docs/ecosystem to make that more evident in our apps, docs and comms.

aschmahmann: Thanks will take a look. Is there any base encoding recommended for the dweb.link paths, or is that left unspecified? 💯 on documenting the path based recommended addressing though since it will at least allow us to point people at something if they ever run into issues copy-pasting a subdomain-based URL.

aschmahmann:
@lidel. I merged the other PR you tagged me in first which led to this having a small merge conflict. Do you mind rebasing and pushing?

lidel:
@aschmahmann rebased, merge at will :+1:

PR #7449, opened 2020-06-09
go: bump minimal dependency to 1.14.4

RubenKelevra:
due to major bugfixes in the .4 release


Stebalien:
I don't want to force this for this release. I'll be _building_ with go 1.14.4, but advanced users should be able to choose.

PR #7452, opened 2020-06-10
feat: allow /ipns/webui.ipfs.io on api port

BajlandoKG:
We want to allow user to try out the latest versions of the webui
before they are officially released with go-ipfs. Right now
ipfs-companion includes a workaround to strip the origin header
from blessed, new versions of webui, to allow this, but it causes
other issues, so we'd like to replace that hack, with this feature.

see: https://github.com/ipfs-shipyard/ipfs-companion/issues/736


BajlandoKG:
Duplicate of #

PR #7461, opened 2020-06-11
feat: gateway redirect to another domain

MichaelMure:
This PR add a way to configure the gateway redirection (path to CID style) to another domain.

Benefits include following the recommended practice of hosting the CID style gateway on another domain to get a strong Origin isolation. Additionally it allow us (Infura) to migrate our traffic out of infura.io in one painless step at the same time.

Although I tested it locally (including running on :80), some unit tests would be nice to have. It's tricky to do though with the current code structure as `HostnameOption()` require a full `IpfsNode` just to get the config.

cc @lidel

Note: require https://github.com/ipfs/go-ipfs-config/pull/101

### DIFF
--- a/core/corehttp/hostname.go
+++ b/core/corehttp/hostname.go
@@ -92,9 +92,16 @@ func HostnameOption() ServeOption {
 
 					// Should this gateway use subdomains instead of paths?
 					if gw.UseSubdomains {
+						hostname := r.Host
+
+						// Redirect to another host if configured
+						if gw.RedirectHost != "" {
+							hostname = gw.RedirectHost
+						}
+
 						// Yes, redirect if applicable
 						// Example: dweb.link/ipfs/{cid} → {cid}.ipfs.dweb.link
-						if newURL, ok := toSubdomainURL(r.Host, r.URL.Path, r); ok {
+						if newURL, ok := toSubdomainURL(hostname, r.URL.Path, r); ok {
 							// Just to be sure single Origin can't be abused in
 							// web browsers that ignored the redirect for some
 							// reason, Clear-Site-Data header clears browsing


MichaelMure:
Btw, this would allow you to redirect the traffic from the path gateway at ipfs.io to dweb.link.

MichaelMure:
Actually, maybe `RedirectDomain` should have the port as well if necessary, instead of extracting it from the original request.

MichaelMure:
> Actually, maybe `RedirectDomain` should have the port as well if necessary, instead of extracting it from the original request.

Did that, the PR is simple and should enable more scenario.

Stebalien:
Any reason not to use an nginx reverse proxy?

MichaelMure:
Yes, we need more complex features (user auth control, specialized rate limiting, routing, analytics ...). Some of those subsystem already exist for other product and can be reused.

Stebalien:
I don't understand. Why would needing _more_ complex features mean redirects should be implemented in go-ipfs itself. So far, we've stuck to implementing HTTP features that are either:

1. Needed by desktop users.
2. Can't be (easily) implemented by sticking a reverse proxy out in front.

Otherwise, go-ipfs should always be run behind a reverse proxy _anyways_ (in production) for load balancing, caching, rate limiting, etc.

MichaelMure:
Ha, I understand your question now.

Yes, we do have a reverse proxy in front of go-ipfs, just not nginx. The reason behind this PR is that the gateway handling logic is:
- quite complex: plenty of specific HTTP headers, subtle details that we want to match to not break things
- sensitive: PL did a great job to assess subtle security issue. We want to leverage that work instead of masking it behind a crude proxy.
- hard to replicate: it's not a go package that we can just bolt within our proxy. We would have to somehow replicate that and especially maintain that when it will inevitably change.

It's way easier now and in the long run to just let go-ipfs do it's thing and make the proxy "transparent" (https://github.com/ipfs/go-ipfs/pull/7468) when needed.

core/corehttp/hostname.go
```diff
@@ -92,9 +92,16 @@ func HostnameOption() ServeOption {
 
 					// Should this gateway use subdomains instead of paths?
 					if gw.UseSubdomains {
```
lidel, line 94: `RedirectHost` should work for all gateways, not just subdomains

Stebalien:
I assume the way you'd implement this would be to just redirect _all_ gateway traffic to the new domain. That is, have a blanket `/ipfs`, `/ipns`, etc. redirect. Then the new gateway can handle the subdomain redirect.

Stebalien:
My question is really just why? The real issue here seems to be about migrating a gateway from one domain to another for security reasons. That seems like something that can and should be done via some reverse proxy rules.

Unless there's some reason we need this support on the desktop as well?

lidel:
Migrating gateways is the only use case I see, and I feel https://github.com/ipfs/go-ipfs/pull/7468 may be a better way to tackle that specific case (reverse proxy could  simply set `X-Forwarded-Host` to the new hostname, and go-ipfs would do the rest)

lidel:
Closing this in favour of a simpler solution proposed in https://github.com/ipfs/go-ipfs/pull/7468

PR #7468, opened 2020-06-12
feat: support X-Forwarded-Host header

MichaelMure:
Chained after https://github.com/ipfs/go-ipfs/pull/7461

This PR add support for `X-Forwarded-Host` to use the original host of the request when a proxy is used.

cc @lidel

### DIFF
--- a/test/sharness/t0114-gateway-subdomains.sh
+++ b/test/sharness/t0114-gateway-subdomains.sh
@@ -664,6 +664,36 @@ test_hostname_gateway_response_should_contain \
   "http://127.0.0.1:$GWAY_PORT/" \
   "$CID_VAL"
 
+## ============================================================================
+## Test support for X-Forwarded-Host
+## ============================================================================
+
+# set explicit subdomain gateway config for the hostname
+ipfs config --json Gateway.PublicGateways '{
+  "example.com": {
+    "UseSubdomains": true,
+    "Paths": ["/ipfs", "/ipns", "/api"]
+  }
+}' || exit 1
+# restart daemon to apply config changes
+test_kill_ipfs_daemon
+test_launch_ipfs_daemon --offline
+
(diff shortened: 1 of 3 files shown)


MichaelMure:
@lidel this should address all your concerns. Maybe it would need more tests ? I'll let you judge on that.

test/sharness/t0114-gateway-subdomains.sh
```diff
@@ -664,6 +664,31 @@ test_hostname_gateway_response_should_contain \
...
+test_launch_ipfs_daemon --offline
+
+test_expect_success "request for http://fake.domain.com/ipfs/{CID} doesn't match the example.com gateway" "
+  curl -H \"Host: fake.domain.com\" -sD - \"http://127.0.0.1:$GWAY_PORT/ipfs/$CIDv1\" > response &&
+  test_should_contain \"200 OK\" response
+"
+
+test_expect_success "request for http://fake.domain.com/ipfs/{CID} with X-Forwarded-Host: example.com match the example.com gateway" "
+  curl -H \"Host: fake.domain.com\" -H \"X-Forwarded-Host: example.com\" -sD - \"http://127.0.0.1:$GWAY_PORT/ipfs/$CIDv1\" > response &&
+  test_should_contain \"Location: http://$CIDv1.ipfs.example.com/\" response
```
lidel, line 689: Do you mind also adding  `X-Forwarded-Proto: https` to the request here, and update response check to confirm it redirects to `https://`?

That way we would ensure both `X-Forwarded-Host` and `X-Forwarded-Proto` are guarded against regressions :ok_hand:

MichaelMure: Test added.

MichaelMure:
Sharness fail is unrelated.

PR #7472, opened 2020-06-13
systemd: specify repo path, to avoid unnecessary subdirectory

RubenKelevra:
## Reational

Currently, the service files will create a hidden directory '/var/lib/ipfs/.ipfs/' which is not necessary for a system-user installation since it won't store any additional files - like in a regular user directory.

This change will set the directory to '/var/lib/ipfs/' to avoid this.

### DIFF
--- a/misc/systemd/ipfs-hardened.service
+++ b/misc/systemd/ipfs-hardened.service
@@ -62,6 +62,7 @@ Type=notify
 User=ipfs
 Group=ipfs
 StateDirectory=ipfs
+Environment=IPFS_PATH="${HOME}"
 ExecStart=/usr/bin/ipfs daemon --init --migrate
 Restart=on-failure
 KillSignal=SIGINT
--- a/misc/systemd/ipfs.service
+++ b/misc/systemd/ipfs.service
@@ -33,6 +33,7 @@ Type=notify
 User=ipfs
 Group=ipfs
 StateDirectory=ipfs
+Environment=IPFS_PATH="${HOME}"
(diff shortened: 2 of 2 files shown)


Stebalien:
That's reasonable, but I'm pretty sure `~` isn't valid. You probably need "${HOME}".

RubenKelevra:
> That's reasonable, but I'm pretty sure `~` isn't valid. You probably need "${HOME}".

`~` works fine, IPFS will read just the one character and expand the home itself. :)

RubenKelevra:
Well, `~` works fine for the service as well as in a shell. :thinking: 

I've tested it before I made the PR.

I thought IPFS makes the expansion, but it works without. The environment variable gets expanded by the system:

```console
$ export IPFS_PATH=~
$ echo $IPFS_PATH
/home/ruben
$ /usr/bin/ipfs init
initializing IPFS node at /home/ruben
generating 2048-bit RSA keypair...done
peer identity: QmX35MVTjvJ3BxPABcphzsJJcPwUkKVL96icYP23mHgqWY
to get started, enter:

	ipfs cat /ipfs/QmQPeNsJPyVWPFDVHb77w8G42Fvo15z4bG2X8D2GhfbSXc/readme

```

Stebalien:
Ah, you're right. go-ipfs does expand it (although I have no idea why). However, please use `$HOME` instead of relying on this. This is going to be very confusing for anyone reading the unit file.

RubenKelevra:
@Stebalien please rereview :)

PR #7477, opened 2020-06-15
fix: override GOFLAGS

### LINKED ISSUE #7475: Setting GOFLAGS overrides commit hash
#### Version information:

> v0.5.1

#### Description:

Setting `GOFLAGS` on the make command line overrides all the flags we pass to go. This is _generally_ what the user wants. However, the user usually doesn't want to _also_ override the commit hash.

We should have a separate `INTERNAL_GOFLAGS` variable for flags like this.

Stebalien:
That way we can combine our GOFLAGS with the user specified ones.

This also switches to the new -trimpath build flag for reproducible builds.

fixes #7475

### DIFF
--- a/mk/golang.mk
+++ b/mk/golang.mk
@@ -6,12 +6,13 @@ export GO111MODULE=on
 # pre-definitions
 GOCC ?= go
 GOTAGS ?=
-unexport GOFLAGS
-GOFLAGS ?=
 GOTFLAGS ?=
 
+# Unexport GOFLAGS so we only apply it where we actually want it.
+unexport GOFLAGS
 # Try to make building as reproducible as possible by stripping the go path.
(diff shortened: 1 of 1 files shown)


aschmahmann:
@Stebalien is this just waiting on review, but otherwise good to go?

Stebalien:
Yes.

mk/golang.mk
```diff
@@ -6,12 +6,13 @@ export GO111MODULE=on
...
-unexport GOFLAGS
-GOFLAGS ?=
 GOTFLAGS ?=
 
+# Unexport GOFLAGS so we only apply it where we actually want it.
+unexport GOFLAGS
 # Try to make building as reproducible as possible by stripping the go path.
-GOFLAGS += "-asmflags=all='-trimpath=$(GOPATH)'" "-gcflags=all='-trimpath=$(GOPATH)'"
+# Override so we can combine with the user's go flags.
+override GOFLAGS += -trimpath
```
petar, line 15: I verified go build handles repeated flags.

PR #1440, opened 2015-07-02
Assert and convert config value into predefined struct

rht:
Basically with this PR, 'true' is parsed as boolean, making explicit '--bool'/'--number'/'--string' optional if there is already config struct.

Lower dose version of https://github.com/ipfs/go-ipfs/pull/1355 that doesn't break backward compatibility.

PR #7479, opened 2020-06-16
feat: transport config

Stebalien:
Add a config section for configuring transports. This way, users can disable transports (especially QUIC), and set muxer/security transport priorities.

### DIFF
--- a/core/node/libp2p/relay.go
+++ b/core/node/libp2p/relay.go
@@ -5,17 +5,16 @@ import (
 	relay "github.com/libp2p/go-libp2p-circuit"
 )
 
-func Relay(disable, enableHop bool) func() (opts Libp2pOpts, err error) {
+func Relay(enableRelay, enableHop bool) func() (opts Libp2pOpts, err error) {
 	return func() (opts Libp2pOpts, err error) {
-		if disable {
-			// Enabled by default.
-			opts.Opts = append(opts.Opts, libp2p.DisableRelay())
-		} else {
+		if enableRelay {
 			relayOpts := []relay.RelayOpt{}
 			if enableHop {
 				relayOpts = append(relayOpts, relay.OptHop)
 			}
 			opts.Opts = append(opts.Opts, libp2p.EnableRelay(relayOpts...))
+		} else {
+			opts.Opts = append(opts.Opts, libp2p.DisableRelay())
 		}
 		return
 	}
--- a/core/node/libp2p/sec.go
+++ b/core/node/libp2p/sec.go
@@ -1,0 +1,39 @@
+package libp2p
+
+import (
+	config "github.com/ipfs/go-ipfs-config"
+	"github.com/libp2p/go-libp2p"
+	noise "github.com/libp2p/go-libp2p-noise"
+	secio "github.com/libp2p/go-libp2p-secio"
+	tls "github.com/libp2p/go-libp2p-tls"
+)
+
+func Security(enabled bool, tptConfig config.Transports) interface{} {
+	if !enabled {
+		return func() (opts Libp2pOpts) {
+			// TODO: shouldn't this be Errorf to guarantee visibility?
+			log.Warnf(`Your IPFS node has been configured to run WITHOUT ENCRYPTED CONNECTIONS.
+		You will not be able to connect to any nodes configured to use encrypted connections`)
+			opts.Opts = append(opts.Opts, libp2p.NoSecurity)
+			return opts
+		}
+	}
+
+	// Using the new config options.
+	return func() (opts Libp2pOpts) {
+		opts.Opts = append(opts.Opts, prioritizeOptions([]priorityOption{{
+			priority:        tptConfig.Security.TLS,
+			defaultPriority: 100,
+			opt:             libp2p.Security(tls.ID, tls.New),
+		}, {
+			priority:        tptConfig.Security.SECIO,
+			defaultPriority: 200,
+			opt:             libp2p.Security(secio.ID, secio.New),
+		}, {
+			priority:        tptConfig.Security.Noise,
+			defaultPriority: 300,
+			opt:             libp2p.Security(noise.ID, noise.New),
+		}}))
+		return opts
+	}
+}
--- a/core/node/libp2p/sec.go
+++ b/core/node/libp2p/sec.go
@@ -11,8 +11,7 @@ import (
 func Security(enabled bool, tptConfig config.Transports) interface{} {
 	if !enabled {
 		return func() (opts Libp2pOpts) {
-			// TODO: shouldn't this be Errorf to guarantee visibility?
-			log.Warnf(`Your IPFS node has been configured to run WITHOUT ENCRYPTED CONNECTIONS.
+			log.Errorf(`Your IPFS node has been configured to run WITHOUT ENCRYPTED CONNECTIONS.
 		You will not be able to connect to any nodes configured to use encrypted connections`)
 			opts.Opts = append(opts.Opts, libp2p.NoSecurity)
 			return opts
--- a/core/node/libp2p/smux.go
+++ b/core/node/libp2p/smux.go
@@ -1,54 +1,79 @@
 package libp2p
 
 import (
+	"fmt"
 	"os"
 	"strings"
 
+	config "github.com/ipfs/go-ipfs-config"
 	"github.com/libp2p/go-libp2p"
 	smux "github.com/libp2p/go-libp2p-core/mux"
 	mplex "github.com/libp2p/go-libp2p-mplex"
 	yamux "github.com/libp2p/go-libp2p-yamux"
 )
 
-func makeSmuxTransportOption(mplexExp bool) libp2p.Option {
+func yamuxTransport() smux.Multiplexer {
+	tpt := *yamux.DefaultTransport
+	tpt.AcceptBacklog = 512
+	if os.Getenv("YAMUX_DEBUG") != "" {
+		tpt.LogOutput = os.Stderr
+	}
+
+	return &tpt
+}
+
+func makeSmuxTransportOption(tptConfig config.Transports) (libp2p.Option, error) {
 	const yamuxID = "/yamux/1.0.0"
 	const mplexID = "/mplex/6.7.0"
 
 	ymxtpt := *yamux.DefaultTransport
 	ymxtpt.AcceptBacklog = 512
 
-	if os.Getenv("YAMUX_DEBUG") != "" {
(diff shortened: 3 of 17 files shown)


test/sharness/t0125-twonode.sh
```diff
@@ -89,38 +89,47 @@ test_expect_success "set up tcp testbed" '
...
 '
 run_advanced_test
 
 # test multiplex muxer
 echo "Running advanced tests with mplex"
-export LIBP2P_MUX_PREFS="/mplex/6.7.0"
-run_advanced_test "--enable-mplex-experiment"
-unset LIBP2P_MUX_PREFS
+test_expect_success "disable yamux" '
+  iptb run -- ips config --json Swarm.Transports.Multiplexers.Yamux false
```
aschmahmann, line 109:
```suggestion
  iptb run -- ipfs config --json Swarm.Transports.Multiplexers.Yamux false
```

test/sharness/t0125-twonode.sh
```diff
@@ -89,38 +89,47 @@ test_expect_success "set up tcp testbed" '
...
+'
+run_advanced_test
+
+test_expect_success "re-enable yamux" '
+  iptb run -- ips config --json Swarm.Transports.Multiplexers.Yamux null
```
aschmahmann, line 114:
```suggestion
  iptb run -- ipfs config --json Swarm.Transports.Multiplexers.Yamux null
```

test/sharness/t0125-twonode.sh
```diff
@@ -89,38 +89,47 @@ test_expect_success "set up tcp testbed" '
...
+
 # Test TCP transport
 echo "Testing TCP"
-tcp_addr='"[\"/ip4/127.0.0.1/tcp/0\"]"'
 test_expect_success "use TCP only" '
-  ipfsi 0 config --json Addresses.Swarm '${tcp_addr}' &&
-  ipfsi 1 config --json Addresses.Swarm '${tcp_addr}'
+  iptb run -- ips config --json Swarm.Transports.Network.QUIC false &&
+  iptb run -- ips config --json Swarm.Transports.Network.Relay false &&
+  iptb run -- ips config --json Swarm.Transports.Network.Websocket false &&
```
aschmahmann, line 102:
```suggestion
  iptb run -- ipfs config --json Swarm.Transports.Network.QUIC false &&
  iptb run -- ipfs config --json Swarm.Transports.Network.Relay false &&
  iptb run -- ipfs config --json Swarm.Transports.Network.Websocket false &&
```

test/sharness/t0125-twonode.sh
```diff
@@ -89,38 +89,47 @@ test_expect_success "set up tcp testbed" '
...
 run_advanced_test
 
 # test QUIC
 echo "Running advanced tests over QUIC"
-addr1='"[\"/ip4/127.0.0.1/udp/0/quic\"]"'
 test_expect_success "use QUIC only" '
-  ipfsi 0 config --json Addresses.Swarm '${quic_addr}' &&
-  ipfsi 1 config --json Addresses.Swarm '${quic_addr}'
+  iptb run -- ips config --json Swarm.Transports.Network.QUIC true &&
+  iptb run -- ips config --json Swarm.Transports.Network.TCP false
```
aschmahmann, line 132:
```suggestion
  iptb run -- ipfs config --json Swarm.Transports.Network.QUIC true &&
  iptb run -- ipfs config --json Swarm.Transports.Network.TCP false
```

go.mod
```diff
@@ -32,7 +32,7 @@ require (
 	github.com/ipfs/go-ipfs-blockstore v0.1.4
 	github.com/ipfs/go-ipfs-chunker v0.0.5
 	github.com/ipfs/go-ipfs-cmds v0.2.9
-	github.com/ipfs/go-ipfs-config v0.7.1
+	github.com/ipfs/go-ipfs-config v0.7.3-0.20200616074558-7fbc0dffff0c
```
aschmahmann, line 35: TODO: update once the other PR is merged + released

docs/config.md
```diff
@@ -5,7 +5,7 @@ is read once at node instantiation, either for an offline command, or when
 starting the daemon. Commands that execute on a running daemon do not read the
 config file at runtime.
 
-#### Profiles
+## Profiles
```
aschmahmann, line 8: ❤️ This entire section ❤️

docs/config.md
```diff
@@ -265,25 +319,33 @@ Configures how many AutoNAT requests to service per `AutoNAT.Throttle.Interval`.
 
 Default: 30
 
+Type: `integer`
```
aschmahmann, line 322: Most of the fields with type `integer` do not have any specifications on what the valid ranges of the field are and what, if any, special values exist. It'd be fantastic to have those added in, although given that it's unrelated to the transport configs there's no reason to block on it.

For example, I'm pretty sure <=0 here means the limit is disabled (i.e. is equivalent to infinity) with 0 being used by default in the code to indicate disabled.

It's worth noting that many of these integers have the same (or nearly the same) semantics as `priority`, and it might be worth renaming `priority` to something more broad (I don't have any good name ideas at the moment, so feel free to ignore unless you have a better name).

docs/config.md
```diff
@@ -334,8 +404,9 @@ we'd want to use 1199120 bytes. As of writing, [7 hash
 functions](https://github.com/ipfs/go-ipfs-blockstore/blob/547442836ade055cc114b562a3cc193d4e57c884/caching.go#L22)
 are used, so the constant `k` is 7 in the formula.
 
+Default: `0` (disabled)
```
aschmahmann, line 407: Aside: do we have any interest in "reclaiming" the 0 value to mean `default` now and using another indicator (e.g. -1, false, etc.) to indicate disabled? Doing this in docs now may make it easier for us to update in the future given that we'll know the user's preference.

On the other hand, resolving https://github.com/ipfs/go-ipfs-config/issues/53 and is going to be a bigger endeavor that will likely involve users needing to update their config files, so small things like this might be irrelevant.

core/node/libp2p/relay.go
```diff
@@ -5,17 +5,16 @@ import (
...
-			opts.Opts = append(opts.Opts, libp2p.DisableRelay())
-		} else {
+		if enableRelay {
 			relayOpts := []relay.RelayOpt{}
 			if enableHop {
 				relayOpts = append(relayOpts, relay.OptHop)
 			}
 			opts.Opts = append(opts.Opts, libp2p.EnableRelay(relayOpts...))
+		} else {
+			opts.Opts = append(opts.Opts, libp2p.DisableRelay())
```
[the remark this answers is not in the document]
Stebalien: The relay transport is enabled in libp2p by default. We're explicitly passing "enable" so we can pass options to it.

core/node/libp2p/sec.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/libp2p/go-libp2p"
+	noise "github.com/libp2p/go-libp2p-noise"
+	secio "github.com/libp2p/go-libp2p-secio"
+	tls "github.com/libp2p/go-libp2p-tls"
+)
+
+func Security(enabled bool, tptConfig config.Transports) interface{} {
+	if !enabled {
+		return func() (opts Libp2pOpts) {
+			// TODO: shouldn't this be Errorf to guarantee visibility?
```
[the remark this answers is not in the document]
Stebalien: I agree. I'm not sure why this was a warning in the first place.

core/node/libp2p/smux.go
```diff
@@ -1,54 +1,76 @@
...
-	return libp2p.ChainOptions(opts...)
 }
 
-func SmuxTransport(mplex bool) func() (opts Libp2pOpts, err error) {
+func SmuxTransport(tptConfig config.Transports) func() (opts Libp2pOpts, err error) {
 	return func() (opts Libp2pOpts, err error) {
-		opts.Opts = append(opts.Opts, makeSmuxTransportOption(mplex))
-		return
+		res, err := makeSmuxTransportOption(tptConfig)
+		opts.Opts = append(opts.Opts, res)
```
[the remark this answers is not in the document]
Stebalien: We're mutating the output. But yeah, checking the error is still best practice.

core/node/libp2p/transport.go
```diff
@@ -3,63 +3,44 @@ package libp2p
...
 	"go.uber.org/fx"
 )
 
-// default security transports for libp2p
-var defaultSecurityTransports = []string{"tls", "secio", "noise"}
+func Transports(tptConfig config.Transports) interface{} {
+	return func(pnet struct {
+		fx.In
+		Fprint PNetFingerprint `optional:"true"`
+	}) (opts Libp2pOpts, err error) {
```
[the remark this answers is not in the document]
Stebalien: Again, output, not input. In this case, I'm inclined to just leave it.

test/sharness/t0125-twonode.sh
```diff
@@ -89,38 +89,47 @@ test_expect_success "set up tcp testbed" '
...
 
 # test Noise
 
 echo "Running advanced tests with NOISE"
 noise_transports='"[\"noise\"]"'
 test_expect_success "use noise only" '
-  ipfsi 0 config --json Experimental.OverrideSecurityTransports '${noise_transports}' &&
-  ipfsi 1 config --json Experimental.OverrideSecurityTransports '${noise_transports}'
+  iptb run -- ipfs config --json Swarm.Transports.Security TLS false &&
+  iptb run -- ipfs config --json Swarm.Transports.Security Secio false
```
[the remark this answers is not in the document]
Stebalien: This is a problem, but I think we can revisit it later. We _may_ want a second `DisableDefaults` flag.

docs/config.md
Stebalien: I've updated the documentation.

docs/config.md
Stebalien: I think we need to move away from the whole "zero value" hack in general. In this case, `0` _should_ mean disabled.

Stebalien:
Ok, so I've "fixed" the config commands to mostly do what we need. They're still buggy but that's an issue for a future release.

Stebalien:
Digging into the type checking, it _looks_ like it was a failed attempt to auto-decode the input into the correct type (e.g., detect if something should be a bool, etc.). However:

1. It looks like it never worked correctly.
2. In the original PR, https://github.com/ipfs/go-ipfs/pull/1440, the exact bugs we're running into here were pointed out but, apparently, weren't a problem? I have no idea how that got merged.

PR #7490, opened 2020-06-18
docs: wording, reflow text, typos, unify json indentations, unify console quotes

RubenKelevra:


docs/plugins.md
```diff
@@ -108,25 +96,20 @@ Go-ipfs supports two types of plugins: External and Preloaded.
...
 
 #### In-tree
 
-To build plugins included in
-[plugin/plugins](https://github.com/ipfs/go-ipfs/tree/master/plugin/plugins),
-run:
+To build plugins included in [plugin/plugins](https://github.com/ipfs/go-ipfs/tree/master/plugin/plugins), run:
 
-```bash
+```console
```
Stebalien, line 105: Let's just switch to bash everywhere. It gives us some nice syntax highlighting.

docs/implement-api-bindings.md
```diff
@@ -11,119 +10,88 @@ Sections:
...
-Despite all the generalization spoken about above, the IPFS API is actually very
-simple. You can inspect all the requests made with `nc` and the `--api` option
-(as of [this PR](https://github.com/ipfs/go-ipfs/pull/1598), or `0.3.8`):
+Despite all the generalization spoken about above, the IPFS API is actually very simple. You can inspect all the requests made with `nc` and the `--api` option (as of [this PR](https://github.com/ipfs/go-ipfs/pull/1598), or `0.3.8`):
 
-```
-> nc -l 5002 &
-> ipfs --api /ip4/127.0.0.1/tcp/5002 swarm addrs local --enc=json
+```console
+nc -l 5002 &
```
Stebalien, line 80: Let's keep the `> ` (and use it everywhere). Otherwise, it's impossible to distinguish between input and output.

docs/config.md
```diff
@@ -1,28 +1,20 @@
 # The go-ipfs config file
 
-The go-ipfs config file is a JSON document located at `$IPFS_PATH/config`. It
-is read once at node instantiation, either for an offline command, or when
-starting the daemon. Commands that execute on a running daemon do not read the
-config file at runtime.
+The go-ipfs config file is a JSON document located at `$IPFS_PATH/config`. It is read once at node instantiation, either for an offline command, or when starting the daemon. Commands that execute on a running daemon do not read the config file at runtime.
```
Stebalien, line 3: Why reflow the text? Lots of users view these documents in text editors.

docs/plugins.md
RubenKelevra: Okay, I just used console because that was on the first file I changed :)

docs/implement-api-bindings.md
RubenKelevra: how about $? Since we use bash that's actually more close to the real shell. The ArchLinux uses this for example :)

docs/config.md
RubenKelevra: Because the new files have no line breaks, just the old ones. So I removed all line breaks from all old files to have that consistent.

I don't know any text-editor which doesn't support soft-wrapping :)

Using no hard-wrapping has the advantage of using the available screen space completely.

docs/implement-api-bindings.md
Stebalien: I agree `$` would make more sense, but we use `>` almost everywhere (there was a good reason to pick it, IIRC, but I can't remember where that discussion happened). Changing this isn't worth it.

docs/config.md
Stebalien: Let's just leave this. Mixing refactors like this with actual changes makes it impossible to review.

RubenKelevra: Okay :)

gammazero:
We have already fixed the spelling and wording errors in other PRs, and decided not to modify the line breaks and tabs. If you feel these or other changed are still necessary, please open a new PR.

PR #7505, opened 2020-06-23
fix: use static binaries in docker container

### LINKED ISSUE #7422: Docker arm image fails to run
Hi,

The current IPFS docker images "install just fine" on ARM but give you this error when running:
`exec user process caused "exec format error"`

Now i've been trying to resolve that.
The first step was to build the docker image for arm (just rebuilding it on an arm machine). This on it's own does make me suspect that the current docker images are not build for ARM64. I hope that could be added?

Rebuilding from git works just fine (not when using a zip, see #7420), but running is then again giving an error. Just a different one. This time:
`exec user process caused "no such file or directory"`

Next i took a deep dive in the docker container. Running it with:
`docker run -d -it --name ipfs_host --entrypoint /bin/sh --restart unless-stopped -v $ipfs_staging:/export -v $ipfs_data:/data/ipfs -p 4001:4001 -p 127.0.0.1:8080:8080 -p 127.0.0.1:5001:5001 ipfs`

Note the `--entrypoint /bin/sh` override and `-it` so that i can actually get inside that container and run commands.

Which i can then do with:
`docker exec -it ipfs_host /bin/sh`

Attaching probably works too, i just picked this route.
Then manually running what the actual endpoint did gives me this:
```
/ # /sbin/tini -- /usr/local/bin/start_ipfs
/bin/sh: /sbin/tini: not found
```

Which is weird as /sbin/tini and /usr/local/bin/start_ipfs definitely do exist as files.
Just running /usr/local/bin/start_ipfs won't work either:
```
/ # /usr/local/bin/start_ipfs
Changing user to ipfs
/usr/local/bin/start_ipfs: line 9: su-exec: not found
/usr/local/bin/start_ipfs: exec: line 11: su-exec: not found
```

Now the issue is getting interesting as su-exec does in fact exist.
So manually running tini and su-exec gave me even more interesting results...
```
/ # /sbin/tini
/bin/sh: /sbin/tini: not found
/ # /sbin/su-exec 
/bin/sh: /sbin/su-exec: not found
```
I am suspecting a dependency to be missing here, but i don't know which one. It's surprisingly difficult to get LDD working in that container so i kinda left it there.
I did download `tini-static-arm64` to confirm my suspicion. And sure enough, the static version does run!

However, su-exec still has the same issue. I wouldn't be surprised if the fix for tini also fixes su-exec.

Cheers,
Mark

Stebalien:
This is more robust cross-platform/architecture.

fixes #7422

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,14 +33,14 @@ ENV TINI_VERSION v0.19.0
 RUN set -eux; \
     dpkgArch="$(dpkg --print-architecture)"; \
     case "${dpkgArch##*-}" in \
-        "amd64" | "armhf" | "arm64") tiniArch="tini-$dpkgArch" ;;\
+        "amd64" | "armhf" | "arm64") tiniArch="tini-static-$dpkgArch" ;;\
         *) echo >&2 "unsupported architecture: ${dpkgArch}"; exit 1 ;; \
     esac; \
   cd /tmp \
   && git clone https://github.com/ncopa/su-exec.git \
   && cd su-exec \
   && git checkout -q $SUEXEC_VERSION \
-  && make \
+  && make su-exec-static \
   && cd /tmp \
   && wget -q -O tini https://github.com/krallin/tini/releases/download/$TINI_VERSION/$tiniArch \
   && chmod +x tini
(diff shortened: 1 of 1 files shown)


markg85:
I confirmed that these changes fix running IPFS within a docker container on an odroid-xu4.

PR #7515, opened 2020-06-30
Add ipfs cli 'rotate' command to rotate identity private keys

petar:


cmd/ipfs/rotate.go
```diff
@@ -0,0 +1,91 @@
...
+ipfs uses a repository in the local file system. By default, the repo is
+located at ~/.ipfs. To change the repo location, set the $IPFS_PATH
+environment variable:
+
+    export IPFS_PATH=/path/to/ipfsrepo
+`,
+	},
+	Arguments: []cmds.Argument{},
+	Options: []cmds.Option{
+		cmds.IntOption(bitsOptionName, "b", "Number of bits to use in the generated RSA private key.").WithDefault(nBitsForKeypairDefault),
```
aschmahmann, line 30: Let's use something more like the `init` command from #7251 so we can handle Ed25519 keys.

cmd/ipfs/rotate.go
```diff
@@ -0,0 +1,91 @@
...
+		nBitsForKeypair, _ := req.Options[bitsOptionName].(int)
+		return doRotate(os.Stdout, cctx.ConfigRoot, nBitsForKeypair)
+	},
+}
+
+func doRotate(out io.Writer, repoRoot string, nBitsForKeypair int) error {
+	// Open repo
+	if err := checkWritable(repoRoot); err != nil {
+		return err
+	}
```
aschmahmann, line 59: Is this necessary or shouldn't `fsrepo.Open` cover our bases?

cmd/ipfs/rotate.go
```diff
@@ -0,0 +1,91 @@
...
+	keystore := repo.Keystore()
+	if err := keystore.Put(cfg.Identity.PeerID, privKey); err != nil {
+		return err
+	}
+
+	// Generate new identity
+	identity, err := config.CreateIdentity(out, []options.KeyGenerateOption{options.Key.Size(nBitsForKeypair)})
+	if err != nil {
+		return err
+	}
```
aschmahmann, line 86: Would default to Ed25519 keys here

cmd/ipfs/rotate.go
```diff
@@ -0,0 +1,91 @@
...
+
+	// Save old identity to keystore
+	privKey, err := cfg.Identity.DecodePrivateKey("") //XXX
+	if err != nil {
+		return err
+	}
+	keystore := repo.Keystore()
+	if err := keystore.Put(cfg.Identity.PeerID, privKey); err != nil {
+		return err
+	}
```
aschmahmann, line 80: What do you think about using some format like `prevKey-{Timestamp}` and appending `-{conflictNumber}` if the key name already exists? Might be easier for people to track/handle.

Additionally (or as a replacement) we should have a flag that allows specifying the name of the output key (and returning an error if the name already exists).

petar: What is the conflict number for? There is an alternative way to make the name unique, e.g.
KeyRotation-PeerID-FromPublicKey-ToPublicKey-Timestamp
(The time stamp is just for convenience, not necessary for uniqueness.)

petar: Take a look at latest commit. Specifically:
* I have to places marked as XXX, where I am not sure what "passphrase" to pass DecodePrivateKey.
* See how you like the key name generation

aschmahmann: >I have to places marked as XXX, where I am not sure what "passphrase" to pass DecodePrivateKey.

Don't worry about it. If you look at the function you'll see we don't even use the passphrase, it's just a TODO.

> See how you like the key name generation

`From -> To` seems like overkill. At that point you might as well just do `From` since that contains all the unique information about the key. I'm a little worried about not knowing what base representation people expect to see their keys in. WDYT about circumventing this problem entirely for the time being and just forcing people to pass in the name to use for their old key (and checking to see if it exists so we don't overwrite it). This might let us ignore this problem until we figure out our story on how we're going to represent PeerIDs of various types throughout IPFS (e.g. do we use base36 everywhere?)

test/sharness/t0027-rotate.sh
```diff
@@ -0,0 +1,48 @@
...
+'
+
+test_expect_success "Compare second ID and key to first" '
+ipfs id -f="<id>" > second_id &&
+ipfs id -f="<pubkey>" > second_key &&
+! test_cmp first_id second_id &&
+! test_cmp first_key second_key
+'
+
+test_expect_success "checking ID" '
```
aschmahmann, line 29: looks good. I'd add one more test here to check that the keystore-id for `oldkey` matches `first_id`

go.mod
```diff
@@ -56,6 +56,7 @@ require (
 	github.com/ipfs/go-verifcid v0.0.1
 	github.com/ipfs/interface-go-ipfs-core v0.3.0
 	github.com/ipld/go-car v0.1.0
+	github.com/jbenet/go-fuse-version v0.0.0-20160322195114-6d4c97bcf253 // indirect
```
aschmahmann, line 59: I'd try and get rid of the extraneous go.mod and go.sum additions.

cmd/ipfs/rotate.go
```diff
@@ -0,0 +1,110 @@
...
+located at ~/.ipfs. To change the repo location, set the $IPFS_PATH
+environment variable:
+
+    export IPFS_PATH=/path/to/ipfsrepo
+`,
+	},
+	Arguments: []cmds.Argument{},
+	Options: []cmds.Option{
+		cmds.StringOption(oldKeyOptionName, "o", "Keystore name for the old/rotated-out key."),
+		cmds.StringOption(algorithmOptionName, "a", "Cryptographic algorithm to use for key generation.").WithDefault(algorithmDefault),
```
aschmahmann, line 38: We should default to Ed25519 keys for rotation here, since we're going to include it with #7251 anyway.

test/sharness/t0027-rotate.sh
```diff
@@ -0,0 +1,50 @@
...
+ipfs id -f="<pubkey>" > first_key
+'
+
+test_launch_ipfs_daemon
+
+test_kill_ipfs_daemon
+
+test_expect_success "rotating keys" '
+ipfs rotate --oldkey=oldkey
+'
```
aschmahmann, line 20: Maybe we should put this in a function like you did in the init PR, so we can verify RSA, Ed25519 and the default all work.

petar:
> This PR is good to go. If we switch the default rotation to RSA and maybe use functions in the test to test RSA + Ed25519 + defaults.

Default RSA will work awkwardly with the new strict rule that requires not specifying bits when using ed25519: If you set the defaults to RSA+2048, then whenever you want to use --algorithm=ed25519 you don't have a way of unsetting the number of bits flag.

Maybe best is if rotation has NO default?

aschmahmann:
Having a default is fine. Can't we just have the defaults to mimic what we have in `init` where we default to an algorithm but no bits, and only pass the bits into `config.Identity` when they're specified?

petar:
> Having a default is fine. Can't we just have the defaults to mimic what we have in `init` where we default to an algorithm but no bits, and only pass the bits into `config.Identity` when they're specified?

Sure. This works too.

test/sharness/lib/test-lib.sh
```diff
@@ -193,7 +193,7 @@ test_init_ipfs() {
 
   test_expect_success "ipfs init succeeds" '
     export IPFS_PATH="$(pwd)/.ipfs" &&
-    ipfs init --profile=test -b=2048 > /dev/null
+    ipfs init --profile=test -a=rsa -b=2048 > /dev/null
```
aschmahmann, line 196: We should probably switch this from `-a rsa -b=2048` to just nothing. My guess is that `test_init_ipfs` is supposed to be testing the defaults, this means when we switch the defaults in the next PR we'll be affecting all the tests (probably a good thing)

test/sharness/t0027-rotate.sh
```diff
@@ -8,7 +8,10 @@ test_description="Test rotate command"
 test_rotate() {
         TEST_ALG=$1
 
-        test_init_ipfs
+        test_expect_success "ipfs init" '
+        export IPFS_PATH="$(pwd)/.ipfs" &&
+        ipfs init --profile=test -a=rsa -b=2048 > /dev/null
```
aschmahmann, line 13: Why RSA and not the defaults here? It doesn't seem like there should be any difference rotating from an RSA or Ed25519 key, but wanted to double check if you thought it was worth a test.

petar: Yep, defaults is fine: This is meant to initialize ipfs, not to test "rotating from one type to another".

aschmahmann: Fair enough, we can just add a test that rotates from `Ed25519 -> RSA -> Ed25519`

petar: Fixed.

test/sharness/t0027-rotate.sh
```diff
@@ -4,13 +4,23 @@ test_description="Test rotate command"
 
 . lib/test-lib.sh
 
-# $1 must be one of 'rsa', 'ed25519' or '' (for default key algorithm).
 test_rotate() {
-        TEST_ALG=$1
+        FROM_ALG=$1
+        TO_ALG=$2
```
aschmahmann, line 9: This is fine, although I think we can simplify it so that we:
1. first test basic rotation on default, RSA, Ed25519 using the test_rotate function which indicates that rotation basically works
2. test rotating Ed25519 -> RSA -> Ed25519 which indicates that we can rotate to/from each key type. We can even use `test_expect_success` for these steps to make them easier to debug/notice.

PR #7531, opened 2020-07-09
add flag to "ipfs key gen" to output keys in b36/CIDv1, also make it …

petar:
…default

### DIFF
--- a/core/commands/keystore.go
+++ b/core/commands/keystore.go
@@ -67,7 +67,7 @@ var keyGenCmd = &cmds.Command{
 	Options: []cmds.Option{
 		cmds.StringOption(keyStoreTypeOptionName, "t", "type of the key to create: rsa, ed25519").WithDefault("rsa"),
 		cmds.IntOption(keyStoreSizeOptionName, "s", "size of the key to generate"),
-		cmds.StringOption(keyFormatOptionName, "f", "output format: b58mh or b36cid").WithDefault("b36cid"),
+		cmds.StringOption(keyFormatOptionName, "f", "output format: b58mh or b36cid").WithDefault("b58mh"),
 	},
 	Arguments: []cmds.Argument{
 		cmds.StringArg("name", true, false, "name of key to create"),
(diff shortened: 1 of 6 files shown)


core/commands/keystore.go
```diff
@@ -65,6 +67,7 @@ var keyGenCmd = &cmds.Command{
 	Options: []cmds.Option{
 		cmds.StringOption(keyStoreTypeOptionName, "t", "type of the key to create: rsa, ed25519").WithDefault("rsa"),
 		cmds.IntOption(keyStoreSizeOptionName, "s", "size of the key to generate"),
+		cmds.BoolOption(cidOptionName, "c", "return a base-36 CIDv1 encoding of the key").WithDefault(true),
```
aschmahmann, line 70: Given that we're planning on using b36 as the default the documentation and style of this flag doesn't really make much sense. i.e. what does it mean if "return a base-36 CIDv1 encoding of the key" is false, how is it encoded?

We could do the opposite sort of flag (e.g. "--b58Mh"), but that leaves us slightly exposed in the event we ever want to change the default going forward (hopefully not for a long time 🤞). WDYT?

core/commands/keystore.go
```diff
@@ -112,12 +115,20 @@ var keyGenCmd = &cmds.Command{
 	Type: KeyOutput{},
 }
 
+func formatID(id peer.ID, useCID bool) string {
+	if useCID {
+		return b36.EncodeToStringLc(peer.ToCid(id).Bytes())
```
aschmahmann, line 120: Is there a reason to use this more explicit form rather than `peer.ToCid(id).StringOfBase(mbase.Base36)`?

core/commands/keystore.go
petar: I missed that method. Fixed.

core/commands/keystore.go
petar: I missed that method. Fixed.

core/commands/keystore.go
petar: I think, there should be explicit flags for every supported variant, because it sounds like the default may flip-flop. So how about a flag with mutually-exclusive variants: --format b58mh | b36cid?

aschmahmann: I think that seems reasonable. If someone wants anything else they can pass the b36cid into the `ipfs cid` function. I'm not sure whether the cid option should be called b36cid or just cid, with a comment saying it's currently b36, both seem fine so your call.

petar: I like b58mh and b36cid, because they have the same schema.

PR #7536, opened 2020-07-12
[BOUNTY] Directory page UI improvements

neatonk:
> BOUNTY details: https://github.com/ipfs/dir-index-html/issues/37

These changes are needed to prepare for the Directory page UI improvements
implemented in https://github.com/ipfs/dir-index-html/pull/38.

- update dir-index-html type structs
- emit CID of each directoryItem
- emit size of directory
- emit breadcrumbs

@jessicaschilling 
@aschmahmann


jessicaschilling:
Thanks, @neatonk! @aschmahmann -- can you please have a look as you're able?

neatonk:
@lidel I think you need to actually commit the submodule change or the generate step will not use it.

neatonk:
Thanks @lidel! I'll update the links.

jessicaschilling:
@lidel, looks like this is ready for your review when you're able. Thanks!

core/corehttp/gateway_indexPage.go
```diff
@@ -11,18 +11,51 @@ import (
...
 
+func breadcrumbs(path string) []breadcrumb {
+	var ret []breadcrumb
+	var pathParts = strings.Split(path, "/")
+	for i, pathPart := range pathParts {
+		if pathPart == "" {
+			continue
+		}
+		if pathPart == "ipfs" {
+			ret = append(ret, breadcrumb{Name: pathPart})
```
lidel, line 44: This `for` loop needs to be refactored. 

I believe we want to exclude first namespace (`/ipfs/` and `/ipns/`) from being turned into a link **only if it is the very first segment**. 

Right now it works only for `/ipfs/` :

> 
> 

However, the code is too naive, and disables link for subdirectory:

> 


Test: http://bafybeig2ed6fx4mzkgopqafswqgjclffnscfphwsjmajt2otat5wc2zaim.ipfs.localhost:8080/foo/ipfs/bar

neatonk: Thanks @lidel, you are correct. I have updated the `breadcrumbs` function to correctly handle any valid path and the links you provided should now work as expected.

lidel: Sweet, breadcrumbs look great now! 

> Screenshot_2020-07-23 Files on IPFS(1)


> Screenshot_2020-07-23 Files on IPFS

jessicaschilling:
https://github.com/ipfs/dir-index-html/pull/39 takes care of reconciling `dir-index.html` and `dir-index-uncat.html` (plus typographic tweak suggested by @lidel in https://github.com/ipfs/dir-index-html/pull/38#discussion_r458448753).

Regarding @lidel's comment above,
> Not a part of the bounty, could be a separate PR:
@neatonk, not sure if adding that test is something you'd be interested in?

jessicaschilling:
@neatonk @lidel Release 1.1.0 is available now, added some checkmarks to the [list](https://github.com/ipfs/go-ipfs/pull/7536#pullrequestreview-454430056) above: https://github.com/ipfs/dir-index-html/releases/tag/v1.1.0

jessicaschilling:
Heads up: https://github.com/ipfs/dir-index-html/pull/40 introduces a build script that, if it's the approach we want to take, may be worth cutting a new release for and using here (if only for efficiency).

neatonk:
I'd argue for leaving ipfs/dir-index-html#40 out of this PR for simplicity.

neatonk:
I will not be able to commit to adding `test/sharness/t0115-gateway-dir-index.sh` as I am running low on free time at the moment.

neatonk:
@lidel I've updated this PR to use [dir-index-html@v1.1.0](https://github.com/ipfs/dir-index-html/tree/v1.1.0) and fixed the gateway tests which match on the "Index of" lines. Apologies for not having time to add the sharness tests you mentioned.

jessicaschilling:
New https://github.com/ipfs/dir-index-html/releases/tag/v1.2.0 should be applicable to straight-out sub in to this PR without any modifications, per https://github.com/ipfs/dir-index-html/pull/40#pullrequestreview-456126397 -- @lidel, up to you whether you want to do so. Thanks!

lidel:
Switched to dir-index-html v1.2.0, will add quick sharness tests shortly.

jessicaschilling:
@neatonk - sent you a note re bounty!

lidel:
@aschmahmann done! 

-----

FYSA rebasing this PR is indeed tricky: good call on double-checking.

`go generate .` relies on revision of `assets/dir-index-html` submodule and will discard any uncommited changes made to it. Unfortunately changing that revision was part of this PR, so you have uncommited diff while resolving rebase conflicts, which means a manual rebase will most likely fail/lose data. 

To avoid that problem, I went with `git rebase -Xours origin/master` which automatically picked local version, without ever leaving the tree in "uncommited" state.

assets/README.md
```diff
@@ -45,6 +45,7 @@ commit.
 ```bash
 > go generate .
 > git add bindata.go
+> git add bindata_version_hash.go
```
aschmahmann, line 48: @lidel it looks like you forgot to run this while rebasing as I don't see any changes to the bindata_version_hash.go file.

aschmahmann: @Stebalien merged #7609 and I rebased on top and added the relevant files.

lidel: Tested locally and looks good, super happy about deterministic assets :+1:

aschmahmann: absolutely, it also helped with the fact that different OSes would give different modes.

PR #7542, opened 2020-07-15
repeat gateway subdomain test for all key types

petar:


test/sharness/t0114-gateway-subdomains.sh
```diff
@@ -109,25 +113,25 @@ test_expect_success "Add the test directory" '
...
 test_expect_success "Publish test text file to IPNS using RSA keys" '
-  PEERID=$(ipfs key gen -f=b58mh --type=rsa --size=2048 test_key_rsa | head -n1 | tr -d "\n")
-  IPNS_IDv0=$(echo "$PEERID" | ipfs cid format -v 0)
-  IPNS_IDv1=$(echo "$PEERID" | ipfs cid format -v 1 --codec libp2p-key -b base36)
-  IPNS_IDv1_DAGPB=$(echo "$IPNS_IDv0" | ipfs cid format -v 1 -b base36)
-  test_check_peerid "${PEERID}" &&
+  RSA_KEY=$(ipfs key gen -f=b58mh --type=rsa --size=2048 test_key_rsa | head -n1 | tr -d "\n")
+  RSA_IPNS_IDv0=$(echo "$RSA_KEY" | ipfs cid format -v 0)
+  RSA_IPNS_IDv1=$(echo "$RSA_KEY" | ipfs cid format -v 1 --codec libp2p-key -b base36)
+  RSA_IPNS_IDv1_DAGPB=$(echo "$RSA_IPNS_IDv0" | ipfs cid format -v 1 -b base36)
```
petar, line 119: Should the "ipfs cid" command have a "--codec protobuf" flag like below?

aschmahmann: It's not necessary since when `--codec` isn't specified it leaves the codec alone and the codec for cidv0 (which RSA keys are) is always DagPB.

I'm not sure whether making this more solid by adding `--codec protobuf` would make us more or less likely to catch errors.

PR #7546, opened 2020-07-16
Key import and export cli commands

### LINKED ISSUE #4240: import and export ipfs private keys
#### Version information: 0.4.11-rc1-2eb4c861
go-ipfs version: 0.4.11-rc1-2eb4c861
Repo version: 6
System version: amd64/linux
Golang version: go1.8.3

#### Description: Add functionality to import and export ipfs keys
Please add commands to ipfs in order to be able to export and import (private) keys (à la gpg).

If there is no other suggestion something like `ipfs key export <name>` and `ipfs key import <file> --name=<name>` would do. They could possibly be adapted to import/export only the public key.

#### Reasonable usecase
I use ipns to publish my personal blog, and I'd like to be able to publish to it from more than one machine. Moreover, if my pc break I wouldn't be able to update my blog any more.

#### Attempt at finding a solution
Searching online I was not able to find a way to export my keys (and `ipfs help` shows nothing) except https://github.com/tswindell/ipfs_keys_export that apparently does not work since `ipfs config Identity.PrivKey` gives error `Error: cannot show or change private key through API`.

#### Quick solutions?
Apart from implementing these (which I guess may take some time), is there any hack I can now perform to be able to import/export keys? For example, are they saved in a specific file with a specific well behaving format?

rendaw:
Fixes #4240 

This modifies `ipfs key generate` to add an export flag, and adds `ipfs key export` `ipfs key import` `ipfs key identify`.  The latter prints the public key fingerprint for an exported key.

I ran some tests, but I had issues with running the tests locally so I'm hoping that some CI will run the tests for me if I make a PR.

I mentioned some concerns in the linked issue.  Also not sure how to handle the private key in the sharness tests -- I could move that into a variable, but I didn't see precedent for that so I kept it inline for the first revision.


rendaw:
Are the tests private?  I'm getting redirected to login when I try to look at sharness.

aschmahmann:
Thanks @rendaw, I'll try and take a look at this PR next week. It looks like the CI results are publicly available via the green/red marks but that the runs themselves are not publicly available.

rendaw:
Awesome, thanks!  Also FWIW I included the kitchen sink here, I figured it's easier to add stuff early and remove it later than add stuff later.

aschmahmann:
Thanks @rendaw for the work. A few of the kitchen sink things to remove here are:
- `ipfs key gen` should not allow exporting keys
- `ipfs key identify` doesn't seem super useful right now

I can try and do some pushes to this branch and see if we can push it over the finish line.

rendaw:
Great, thanks!  Removed the export option from gen.

core/commands/keystore.go
```diff
@@ -143,6 +151,159 @@ func formatID(id peer.ID, formatLabel string) string {
...
+			return err
+		}
+		defer r.Close()
+
+		sk, err := r.Keystore().Get(name)
+		if err != nil {
+			return fmt.Errorf("key with name '%s' doesn't exist", name)
+		}
+
+		encoded, err := crypto.MarshalPrivateKey(sk)
```
aschmahmann, line 193: @rendaw I switched us from using base58btc encoded strings to just sending the raw bytes over the wire and working with files.

While in the future it's pretty easy for us to add new base encodings to import/export (e.g. --type=raw/multibase) it's actually pretty painful for us to allow for imports to be strings OR files.

I went with files over strings here as it seems likely to me that people will want to ultimately store these blobs in files. Additionally, the keystore currently utilizes files so people are likely already used to backing up keys this way.

core/commands/keystore.go
```diff
@@ -143,6 +151,159 @@ func formatID(id peer.ID, formatLabel string) string {
 	}
 }
 
+var keyExportCmd = &cmds.Command{
+	Helptext: cmds.HelpText{
+		Tagline: "Export a keypair",
+		ShortDescription: `
+Exports a named libp2p key to disk.
+
+By default, the output will be stored at './<key-name>', but an alternate
```
Stebalien, line 160: Maybe add a `.key` extension?

core/commands/keystore.go
```diff
@@ -143,6 +151,159 @@ func formatID(id peer.ID, formatLabel string) string {
...
+		cmds.StringArg("name", true, false, "name of key to export").EnableStdin(),
+	},
+	Options: []cmds.Option{
+		cmds.StringOption(outputOptionName, "o", "The path where the output should be stored."),
+	},
+	Run: func(req *cmds.Request, res cmds.ResponseEmitter, env cmds.Environment) error {
+		name := req.Arguments[0]
+
+		if name == "self" {
+			return fmt.Errorf("cannot export key with name 'self'")
```
Stebalien, line 174: Hm. Is there any reason not to allow this?

aschmahmann: I guess not, but we have to the keys live in different places and it's a little messier. Let's add it in a followup PR

core/commands/keystore.go
aschmahmann: Sure, that seems reasonable. I'll add it now.

aschmahmann: Done

rendaw:
Awesome, thanks for helping me out here!  The sounds good.  The actual form of this functionality isn't so important to me anyway.

PR #7548, opened 2020-07-16
Move ipfs rotate to ipfs key rotate

aschmahmann:
Moves the `ipfs rotate` command to `ipfs key rotate`

@Stebalien I need some help here verifying that I've correctly done the isolation of the `ipfs key rotate` command to only run offline.

You previously mentioned that `curl -X POST localhost:5001/api/v0/commands` should not print out server-only commands, but other server only commands like `repo fsck` and `config show` do show up that way.


core/commands/keystore.go
```diff
@@ -34,6 +38,7 @@ publish'.
 		"list":   keyListCmd,
 		"rename": keyRenameCmd,
 		"rm":     keyRmCmd,
+		"rotate": keyRotateCmd,
```
Stebalien, line 41: If we add it here, it'll show up in the global command list. To make it not show up, we'd need to "link it in" to the commands tree in cmd/ipfs/ipfs.go (_copying_ the keystore command first). Honestly, as long as we forbid calling this (make sure that works), it's probably not worth it.

core/commands/keystore.go
```diff
@@ -285,3 +290,108 @@ func keyOutputListEncoders() cmds.EncoderFunc {
...
+
+		return nil
+	},
+	Run: func(req *cmds.Request, res cmds.ResponseEmitter, env cmds.Environment) error {
+		cctx := env.(*oldcmds.Context)
+		nBitsForKeypair, nBitsGiven := req.Options[keyStoreBitsOptionName].(int)
+		algorithm, _ := req.Options[keyStoreAlgorithmOptionName].(string)
+		oldKey, ok := req.Options[keyStoreOldKeyOptionName].(string)
+		if !ok {
+			return fmt.Errorf("keystore name for backing up old key must be provided")
```
Stebalien, line 343: Eh, I'd just default to the peer ID.

core/commands/keystore.go
```diff
@@ -285,3 +290,108 @@ func keyOutputListEncoders() cmds.EncoderFunc {
...
+	cfg, err := repo.Config()
+	if err != nil {
+		return fmt.Errorf("reading config from repo (%v)", err)
+	}
+
+	// Generate new identity
+	var identity config.Identity
+	if nBitsGiven {
+		identity, err = config.CreateIdentity(out, []options.KeyGenerateOption{
+			options.Key.Size(nBitsForKeypair),
```
Stebalien, line 367: What happens if I don't specify this?

core/commands/keystore.go
```diff
@@ -285,3 +290,108 @@ func keyOutputListEncoders() cmds.EncoderFunc {
...
+	},
+	Run: func(req *cmds.Request, res cmds.ResponseEmitter, env cmds.Environment) error {
+		cctx := env.(*oldcmds.Context)
+		nBitsForKeypair, nBitsGiven := req.Options[keyStoreBitsOptionName].(int)
+		algorithm, _ := req.Options[keyStoreAlgorithmOptionName].(string)
+		oldKey, ok := req.Options[keyStoreOldKeyOptionName].(string)
+		if !ok {
+			return fmt.Errorf("keystore name for backing up old key must be provided")
+		}
+		return doRotate(os.Stdout, cctx.ConfigRoot, oldKey, algorithm, nBitsForKeypair, nBitsGiven)
```
Stebalien, line 345: Hm. This shouldn't use stdout. We should probably:

1. Create a pipe.
2. Emit the reader.
2. Pass the writer to doRotate.

Stebalien: On the other hand, it would be _nice_ to output structured information (e.g., the new peer ID). Maybe we should just:

1. Create a pipe.
2. Pass the writer to doRotate.
3. For every line written, emit `Result{Message: ...}`.
4. At the end, emit `Result{ID: ...}`?

But this isn't _really_ necessary as the user can always just query `ipfs id`.

test/sharness/t0027-rotate.sh
```diff
@@ -35,13 +35,13 @@ test_rotate() {
...
-                ipfs rotate -a=rsa -b=2048 --oldkey=oldkey
+                ipfs key rotate -a=rsa -b=2048 --oldkey=oldkey
                 ;;
         ed25519)
-                ipfs rotate -a=ed25519 --oldkey=oldkey
+                ipfs key rotate -a=ed25519 --oldkey=oldkey
                 ;;
         *)
-                ipfs rotate --oldkey=oldkey
+                ipfs key rotate --oldkey=oldkey
```
Stebalien, line 44: Let's add a test with cURL, asserting that this cannot be run remotely.

core/commands/keystore.go
aschmahmann: If people find it annoying we can do it later, for now I'd rather people be confused if they run rotate twice as to which key was their most recent one (yes, they could probably do an ls and use the filesystem to figure it out but why not have them use a name)

core/commands/keystore.go
aschmahmann: If `nBitsGiven=false` then we use whatever the defaults are as defined in config.CreateIdentity (2048 bits for RSA, and the  non-configurable builtin 256 for Ed25519)

core/commands/keystore.go
Stebalien: Fair enough.

aschmahmann:
Decided to leave the command where it is for now.

PR #7553, opened 2020-07-20
add ipfs dag stat command

aschmahmann:
Adds a command to get simple stats out from a DAG (e.g. how big is it and how many blocks does it have).

This PR certainly has room for improvement, but adding this command is probably better than not.


core/commands/dag/dag.go
```diff
@@ -668,3 +671,134 @@ The output of blocks happens in strict DAG-traversal, first-seen, order.
...
+		}
+
+		rp, err := api.ResolvePath(req.Context, path.New(req.Arguments[0]))
+		if err != nil {
+			return err
+		}
+
+		if len(rp.Remainder()) > 0 {
+			return fmt.Errorf("cannot return size for anything other than a DAG with a root CID")
+		}
```
aschmahmann, line 715: This is supportable, but didn't seem necessary for an initial release of this command. Open to counter opinions.

core/commands/dag/dag.go
```diff
@@ -668,3 +671,134 @@ The output of blocks happens in strict DAG-traversal, first-seen, order.
...
+				// Size() function.
+				if len(stat.Hash) == 0 {
+					size, err := current.Node.Size()
+					if err != nil {
+						return err
+					}
+					dagstats.Size += size
+				} else {
+					dagstats.Size += uint64(stat.BlockSize)
+				}
```
aschmahmann, line 746: This is dumb. @Stebalien do you know if there's a better/cleaner way to handle this? Should I just be checking if the node is DagPB instead of checking `len(stat.Hash)!=0`?

The latter seemed slightly more generic since it appears that the pattern is either to return a full stat object or an empty stat object with nothing in between.

core/commands/dag/dag.go
```diff
@@ -668,3 +671,134 @@ The output of blocks happens in strict DAG-traversal, first-seen, order.
...
+}
+
+var DagStatCmd = &cmds.Command{
+	Helptext: cmds.HelpText{
+		Tagline: "Gets stats for a DAG",
+		ShortDescription: `
+'ipfs dag size' fetches a dag and returns various statistics about the DAG.
+Statistics include size and number of blocks.
+
+Note: This command skips duplicate blocks in reporting both size and the number of blocks
```
aschmahmann, line 691: We should be able to report both sets of data (duplicates + not) together, but it requires a bit of state tracking to do it correctly. This didn't seem necessary for an initial release of this command, but I'm open to counter opinions.

@obo20 you had a request for this command would you prefer to see the data with the duplicate blocks or without?

core/commands/root.go
```diff
@@ -191,6 +191,7 @@ var rootROSubcommands = map[string]*cmds.Command{
 		Subcommands: map[string]*cmds.Command{
 			"get":     dag.DagGetCmd,
 			"resolve": dag.DagResolveCmd,
+			"stat":    dag.DagStatCmd,
```
aschmahmann, line 194: @Stebalien I notice that `ipfs dag export` isn't listed as a ReadOnly command (https://github.com/ipfs/go-ipfs/pull/7036#discussion_r405716667). This should be no more expensive then a Get so I'm assuming it's fine to expose this. Seem reasonable?

core/commands/dag/dag.go
obo20: @aschmahmann I can see both ways having usefulness, but deduplicated would probably be the most useful since that's one of the strengths of IPFS.

core/commands/root.go
ribasushi: @aschmahmann was requested here: https://github.com/ipfs/go-ipfs/pull/7036#discussion_r405692836

core/commands/dag/dag.go
Stebalien: I'd drop the `Stat` call entirely. It's deprecated from IPLD's standpoint.

Stebalien: I.e., just get the node size.

aschmahmann: I can't do `node.Size` for DagPB since it will give me the cumulative node size https://github.com/ipfs/go-merkledag/blob/749fd8717d46b4f34c9ce08253070079c89bc56d/node.go#L236.

Were you thinking it'd be better to do `len(nd.RawData())`?

Stebalien: I really hate this code... Yes, `len(nd.RawData())` is the only way to go.

achingbrain:
This is really useful and something I use `ipfs files stat` for all the time.

It would be great if this could be a top level command, e.g. `ipfs stat /ipfs/path` instead of tucking it away under the dag subcommand.

aschmahmann:
> It would be great if this could be a top level command, e.g. ipfs stat /ipfs/path instead of tucking it away under the dag subcommand.

We have `ipfs x stat` in a few different places already (`ipfs bitswap stat`, `ipfs files stat`) I think making it top level would be a little more confusing. It also makes it a little more clear what the difference is between `ipfs files stat` and `ipfs dag stat` one of them operates on UnixFS objects so could return things like "the file size", the other operates on DAGs and so returns IPLD block information.

aschmahmann:
@achingbrain for reasons... we also have `ipfs stats X`. It seems like moving data based stats (e.g. files, object, dag) as aliases under the `ipfs stats` header was examined at https://github.com/ipfs/go-ipfs/pull/2810#issuecomment-223855758 and passed on.

I don't really have strong feelings about these aliases (other then I don't know why we have them at all), so if you'd like `ipfs stats dag` = `ipfs dag stat` we could do that in another PR where we move the other data based commands. WDYT?

core/commands/dag/dag.go
```diff
@@ -668,3 +671,115 @@ The output of blocks happens in strict DAG-traversal, first-seen, order.
...
+`,
+	},
+	Arguments: []cmds.Argument{
+		cmds.StringArg("root", true, false, "CID of a DAG root to get statistics for").EnableStdin(),
+	},
+	Options: []cmds.Option{
+		cmds.BoolOption(progressOptionName, "p", "Return progressive data while reading through the DAG").WithDefault("true"),
```
Stebalien, line 698: `WithDefault(true)`

core/commands/dag/dag.go
```diff
@@ -668,3 +671,115 @@ The output of blocks happens in strict DAG-traversal, first-seen, order.
...
+			return re.Emit(dagStats)
+		},
+	},
+	Encoders: cmds.EncoderMap{
+		cmds.Text: cmds.MakeTypedEncoder(func(req *cmds.Request, w io.Writer, event *DagStat) error {
+			_, err := fmt.Fprintf(
+				w,
+				"%v\n",
+				event,
+			)
```
aschmahmann, line 781: This seems like the pattern we use for emitting updates as well as a final result, but would like some confirmation that this is reasonable.

test/sharness/t0053-dag.sh
```diff
@@ -268,6 +268,41 @@ test_dag_cmd() {
     test_cmp resolve_obj_exp resolve_obj &&
     test_cmp resolve_data_exp resolve_data
   '
+
+  test_expect_success "dag stat of simple IPLD object" '
```
aschmahmann, line 272: Added a few tests here, do we need more before shipping this?

core/commands/dag/dag.go
```diff
@@ -668,3 +671,115 @@ The output of blocks happens in strict DAG-traversal, first-seen, order.
...
+		}
+
+		obj, err := api.Dag().Get(req.Context, rp.Cid())
+		if err != nil {
+			return err
+		}
+
+		dagstats := &DagStat{}
+		err = traverse.Traverse(obj, traverse.Options{
+			DAG:   api.Dag(),
```
Stebalien, line 724: We should probably try to create a bitswap session here, but I'm fine punting that if necessary (this API will usually be called on local DAGs, I assume).

core/commands/dag/dag.go
Stebalien: It is... I just wish we had a better way.

test/sharness/t0053-dag.sh
Stebalien: This seems reasonable.

PR #7554, opened 2020-07-20
IPNS format keys in b36cid

petar:


petar:
To run sharness locally, during dev, run this from the test/sharness dir:

`TEST_NO_FUSE=1 GOIPFSTEST_SKIP_LOCAL_DEVTREE_DEPS_CHECK=1 make`

core/commands/keystore.go
```diff
@@ -94,7 +93,7 @@ var keyGenCmd = &cmds.Command{
 		if sizefound {
 			opts = append(opts, options.Key.Size(size))
 		}
-		if err = verifyFormatLabel(req.Options[keyFormatOptionName].(string)); err != nil {
+		if err = coreapi.VerifyIDFormatLabel(req.Options[keyFormatOptionName].(string)); err != nil {
```
aschmahmann, line 96: Let's use a local verifyID format since it's particular to this command

petar:
> We need more sharness tests here. In particular we should
> 
> 1. Like with the previous PR, have a function that will let us test that Ed25519 + RSA keys both work (it's not quite in line with this PR but we need to do it anyway
> 2. Make sure that we can publish using a b58mh as well as a b36cid. The results can all come out as `Publish to ${b36cidKey}` though.

I've added a test for (2). For (1), it seems one good strategy is to wrap all tests in a function parameterized by ed25519/rsa and use this to initialize the self ID, because most tests are based on the self id. Would you prefer a different approach?

aschmahmann:
> (1), it seems one good strategy is to wrap all tests in a function parameterized by ed25519/rsa  and use this to initialize the self ID, because most tests are based on the self id. Would you prefer a different approach?

We need to test publishing with the `--key` flag anyway. So maybe something like this:
- Have a function that tests with ed25519, rsa, and self
- Test that publishing with no --key flag means we published with self
- Test that we can publish with `--key=peerIDRepresentationOfKey`, or `--key=friendlyName`

petar:
> > (1), it seems one good strategy is to wrap all tests in a function parameterized by ed25519/rsa  and use this to initialize the self ID, because most tests are based on the self id. Would you prefer a different approach?
> 
> We need to test publishing with the `--key` flag anyway. So maybe something like this:
> 
> * Have a function that tests with ed25519, rsa, and self
> * Test that publishing with no --key flag means we published with self
> * Test that we can publish with `--key=peerIDRepresentationOfKey`, or `--key=friendlyName`

I think now all these cases are covered.

go.mod
```diff
@@ -1,5 +1,7 @@
 module github.com/ipfs/go-ipfs
 
+// replace github.com/ipfs/interface-go-ipfs-core => /Users/petar/src/github.com/ipfs/interface-go-ipfs-core
+
```
aschmahmann, line 4: TODO: Remove

test/sharness/t0100-name.sh
```diff
@@ -8,209 +8,272 @@ test_description="Test ipfs repo operations"
...
+        '
+
+        test_expect_success "resolve output looks good" '
+        printf "/ipfs/%s\n" "$HASH_WELCOME_DOCS" >expected2 &&
+        test_cmp expected2 output
+        '
+
+        # publish with an explicit node ID as key name
+
+        test_expect_success "'ipfs name publis --allow-offline --key=<peer-id> <hash>' succeeds" '
```
aschmahmann, line 262:
```suggestion
        test_expect_success "'ipfs name publish --allow-offline --key=<peer-id> <hash>' succeeds" '
```

test/sharness/t0100-name.sh
```diff
@@ -8,209 +8,272 @@ test_description="Test ipfs repo operations"
...
+        # test publishing with B36CID and B58MH resolve to the same B36CID
+
+        test_expect_success "generate and verify a new key" '
+        B58MH_ID=`ipfs key list -f=b58mh -l | grep self | cut -d " " -f1` &&
+        B36CID_ID=`ipfs key list -f=b36cid -l | grep self | cut -d " " -f1` &&
+        test_check_peerid "${B58MH_ID}" &&
+        test_check_peerid "${B36CID_ID}"
+        '
+
+        test_expect_success "'ipfs name publis --allow-offline --key=<peer-id> <hash>' succeeds" '
```
aschmahmann, line 240:
```suggestion
        test_expect_success "'ipfs name publish --allow-offline --key=<peer-id> <hash>' succeeds" '
```

test/sharness/t0100-name.sh
```diff
@@ -8,209 +8,272 @@ test_description="Test ipfs repo operations"
...
+        test_expect_code 1 ipfs name publish "/ipfs/$HASH_WELCOME_DOCS"
+        '
+
+        test_kill_ipfs_daemon
+
+        test_expect_success "clean up ipfs dir" '
+        rm -rf "$IPFS_PATH"
+        '
+}
+test_name_with_self 'rsa'
```
aschmahmann, line 206: Let's add one more case for the default too

test/sharness/t0100-name.sh
```diff
@@ -8,209 +8,272 @@ test_description="Test ipfs repo operations"
...
+        B58MH_ID=`ipfs key list -f=b58mh -l | grep self | cut -d " " -f1` &&
+        B36CID_ID=`ipfs key list -f=b36cid -l | grep self | cut -d " " -f1` &&
+        test_check_peerid "${B58MH_ID}" &&
+        test_check_peerid "${B36CID_ID}"
+        '
+
+        test_expect_success "'ipfs name publis --allow-offline --key=<peer-id> <hash>' succeeds" '
+        ipfs name publish --allow-offline  --key=${B58MH_ID} "/ipfs/$HASH_WELCOME_DOCS" >b58mh_published_id &&
+        ipfs name publish --allow-offline  --key=${B36CID_ID} "/ipfs/$HASH_WELCOME_DOCS" >b36cid_published_id
+        '
```
aschmahmann, line 243: This code tests publishing with the `self` key which what is tested in the other function `test_name_with_self `. Did you mean to use the generated key here?

test/sharness/t0100-name.sh
```diff
@@ -8,209 +8,272 @@ test_description="Test ipfs repo operations"
...
+
+        test_expect_success "'prepare keys" '
+        case $GEN_ALG in
+        rsa)
+                export KEY=`ipfs key gen -f=b58mh --type=rsa --size=2048 key` &&
+                export KEY_B36CID=`ipfs key list -f=b36cid -l | grep key | cut -d " " -f1`
+                ;;
+        ed25519)
+                export KEY=`ipfs key gen -f=b36cid --type=ed25519 key`
```
aschmahmann, line 224: It seems like this should be exported as b58, same as for RSA. If we want to increase test coverage by testing that publishing with b58 results in b36 output and publishing with b36 results in b36 output we can just do that.

test/sharness/t0100-name.sh
petar: No: This test is focused on using an explicit key. Moved it to test_name_with_self.

core/commands/fmt.go
```diff
@@ -0,0 +1,33 @@
...
+	switch formatLabel {
+	case "b58mh":
+		return nil
+	case "b36cid":
+		return nil
+	}
+	return fmt.Errorf("invalid output format option")
+}
+
+func formatID(id peer.ID, formatLabel string) string {
```
aschmahmann, line 20: Since these are only used by the `ipfs key` commands, any reason not to move them to that file? Most of the files in ipfs/core/commands correspond to commands.

test/sharness/t0100-name.sh
aschmahmann: this seems fine now

test/sharness/t0100-name.sh
```diff
@@ -8,209 +8,282 @@ test_description="Test ipfs repo operations"
...
+
+        # test IPNS + IPLD
+
+        test_expect_success "'ipfs dag put' succeeds" '
+        HELLO_HASH="$(echo "\"hello world\"" | ipfs dag put)" &&
+        OBJECT_HASH="$(echo "{\"thing\": {\"/\": \"${HELLO_HASH}\" }}" | ipfs dag put)"
+        '
+        test_expect_success "'ipfs name publish --allow-offline /ipld/...' succeeds" '
+        PEERID=`ipfs key list -f=b36cid -l | grep self | cut -d " " -f1` &&
+        test_check_peerid "${PEERID}" &&
```
aschmahmann, line 146: Any reason we're redefining PEERID and checking it again, or just copy-paste?

test/sharness/t0100-name.sh
```diff
@@ -8,209 +8,282 @@ test_description="Test ipfs repo operations"
...
+        '
+
+        test_expect_failure "publish with our explicit node ID looks good" '
+        echo "Published to ${PEERID}: /ipfs/$HASH_WELCOME_DOCS" >expected_node_id_publish &&
+        test_cmp expected_node_id_publish actual_node_id_publish
+        '
+
+        # test publishing with B36CID and B58MH resolve to the same B36CID
+
+        test_expect_success "generate and verify a new key" '
```
aschmahmann, line 111:
```suggestion
        test_expect_success "verify self key output" '
```

test/sharness/t0100-name.sh
petar: copy/paste.

PR #7561, opened 2020-07-27
refactor: use dir-index-html/dist/dir-index.html

lidel:
> ## Do not merge,   https://github.com/ipfs/dir-index-html/pull/40 decided to keep old path instead.

This PR depends on https://github.com/ipfs/dir-index-html/pull/40 where @jessicaschilling cleaned up the build process and introduced clear separation between source and generated assets.


lidel:
Not needed, ipfs/dir-index-html#40 decided to keep old path instead

PR #7579, opened 2020-08-05
use ed25519 keys by default

petar:


test/sharness/t0184-http-proxy-over-p2p.sh
```diff
@@ -167,7 +167,7 @@ test_expect_success 'setup p2p listener on the receiver' '
 '
 
 test_expect_success 'setup environment' '
-    RECEIVER_ID="$(iptb attr get 1 id)"
```
aschmahmann, line 170: @petar does this test fail if we only use base36 representation for the RECEIVER_ID_CIDv1 below? I thought the gateway redirect would work here. If not then we might need a minor modification to the gateway code.

cmd/ipfs/init.go
```diff
@@ -24,7 +24,7 @@ import (
 )
 
 const (
-	algorithmDefault    = options.RSAKey
```
aschmahmann, line 27: Also need to do this in core/commands/keystore.go

aschmahmann:
🚀

PR #7582, opened 2020-08-07
use b36 keys by default for keys and IPNS

petar:


core/commands/keystore.go
```diff
@@ -77,7 +77,7 @@ var keyGenCmd = &cmds.Command{
 	Options: []cmds.Option{
 		cmds.StringOption(keyStoreTypeOptionName, "t", "type of the key to create: rsa, ed25519").WithDefault("rsa"),
 		cmds.IntOption(keyStoreSizeOptionName, "s", "size of the key to generate"),
-		cmds.StringOption(keyFormatOptionName, "f", "output format: b58mh or b36cid").WithDefault("b58mh"),
+		cmds.StringOption(keyFormatOptionName, "", "output format: b58mh or base36").WithDefault("base36"),
```
aschmahmann, line 80: Maybe replace all of the text here with something like:

Encoding used for keys: Can either be a multibase encoded CID or a base58btc encoded multihash. Takes {b58mh|base36|k|base32|b...}.

core/commands/keystore.go
```diff
@@ -218,7 +218,7 @@ var keyImportCmd = &cmds.Command{
 		Tagline: "Import a key and prints imported key id",
 	},
 	Options: []cmds.Option{
-		cmds.StringOption(keyFormatOptionName, "f", "output format: b58mh or b36cid").WithDefault("b58mh"),
+		cmds.StringOption(keyFormatOptionName, "", "output format: b58mh or base36").WithDefault("b58mh"),
```
aschmahmann, line 221:
```suggestion
		cmds.StringOption(keyFormatOptionName, "", "output format: b58mh or base36").WithDefault("base36"),
```

core/commands/keystore.go
```diff
@@ -415,26 +426,28 @@ var keyRmCmd = &cmds.Command{
...
+		return err
 	}
-	return fmt.Errorf("invalid output format option")
 }
 
 func formatID(id peer.ID, formatLabel string) string {
 	switch formatLabel {
-	case "b58mh":
+	case "b58mh", "v0":
 		return id.Pretty()
```
aschmahmann, line 440: Let's use `id.IDB58Encode()` instead of `id.Pretty()` it's more precise in what we're looking for here and is never going to change. I know the function says it's deprecated, but it seems the safest to use.

core/commands/keystore.go
```diff
@@ -415,26 +426,28 @@ var keyRmCmd = &cmds.Command{
...
-	case "b36cid":
-		if s, err := peer.ToCid(id).StringOfBase(mbase.Base36); err != nil {
+	default:
+		enc, err := mbase.EncoderByName(formatLabel)
+		if err != nil {
+			panic("invalid IPNS key encoding")
```
aschmahmann, line 444: Separating the validation function from the formatting function feels like a footgun. As we may accidentally call `formatID` without calling `verifyIDFormatLabel` and ending up with a panic. 

How about creating a tiny struct to combine these two functions, one for creating the encoder and the other for running it.

test/sharness/t0160-resolve.sh
```diff
@@ -22,8 +22,10 @@ test_expect_success "resolve: prepare dag" '
 '
 
 test_expect_success "resolve: prepare keys" '
-    self_hash=$(ipfs id -f="<id>") &&
-    alt_hash=$(ipfs key gen -f=b58mh -t rsa alt)
+    self_hash=$(ipfs key list --ipns-base=base36 -l | grep self | cut -d " " -f1) &&
+    alt_hash=$(ipfs key gen --ipns-base=base36 -t rsa alt)
+    echo self_hash $self_hash
+    echo $(ipfs id -f="<id>")
```
aschmahmann, line 28: These were just for debugging right?

core/commands/resolve.go
```diff
@@ -128,6 +111,23 @@ Resolve the value of an IPFS DAG path:
 			return cmds.EmitOnce(res, &ncmd.ResolvedPath{Path: ipfspath.Path(p.String())})
 		}
 
+		var enc cidenc.Encoder
+		switch {
+		case !cmdenv.CidBaseDefined(req) && !strings.HasPrefix(name, "/ipns/"):
```
aschmahmann, line 116: @petar @Stebalien now that we've introduced `--ipns-base` should we change this behavior to determine the IPNSBase and then do a switch on the ResolvedPath to figure out whether we should use the CID encoder or the IPNS encoder (i.e. if it starts with /ipfs | /ipld or starts with /ipns?)

aschmahmann:
@petar also need to switch to b36 by default here https://github.com/ipfs/go-ipfs/blob/97fee07bd271033094aed21a12eae116db2f93a1/core/commands/name/ipnsps.go#L96.

I somewhat doubt if we need to add the `--ipns-base` flag here too, but I guess we can.

test/sharness/t0183-namesys-pubsub.sh
```diff
@@ -56,15 +62,15 @@ test_expect_success 'wait for the flood' '
 
 test_expect_success 'resolve name in subscriber nodes' '
     echo "/ipfs/$HASH_FILE" > expected &&
-    ipfsi 1 name resolve /ipns/$PEERID_0 > name1 &&
-    ipfsi 2 name resolve /ipns/$PEERID_0 > name2 &&
+    ipfsi 1 name resolve /ipns/$PEERID_0_BASE36 > name1 &&
+    ipfsi 2 name resolve /ipns/$PEERID_0_BASE36 > name2 &&
```
aschmahmann, line 66: Unless I've missed them somewhere else we should add some sanity checks that IPNS over PubSub works (i.e. publishes + resolves) with b36, b58, and when publishing with one and resolving with the other. 

It's probably fine, but given that IPNS over PubSub is experimental it's easier for bugs to creep in and get missed

core/commands/keybase/keybase.go
```diff
@@ -0,0 +1,37 @@
+package keybase
+
+import (
+	peer "github.com/libp2p/go-libp2p-core/peer"
+	mbase "github.com/multiformats/go-multibase"
+)
+
+const KeyFormatOptionName = "ipns-base"
```
aschmahmann, line 8: I wouldn't make a package level `KeyFormatOptionName` that is "ipns-base". Either this package is generically dealing with keys or it's only dealing with IPNS keys.

We can keep the package as something like keybase, but then this option should be more like `IPNSKeyFormatOptionName`

core/commands/keybase/keybase.go
```diff
@@ -0,0 +1,37 @@
+package keybase
```
aschmahmann, line 1: Maybe `keyencode` is better here as it's the name of the name struct.

Also, `keybase` may be confusing, especially with a project not so far removed from our space having that name.

test/sharness/t0183-namesys-pubsub.sh
petar: It turns out that to do this, we need to either: (a) add --ipns-base to 'ipfs add' or (b) expand the supported bases for 'ipfs cid'.

aschmahmann: Why? We could just do `ipfs name resolve b58key` and `ipfs name resolve b36key` and that should just work.

Looking at this file I don't think we're really testing IPNS over PubSub very well anyway since the DHT is likely very fast in this two node test resulting in the tests passing even if PubSub is sort of broken. We don't have to fix this issue in this PR though.

PR #7598, opened 2020-08-17
Upgrade graphsync deps

ribasushi:
There was supposed to be a bit more work landing here, but I won't get to it in time this week.
Let's unblock https://github.com/ipfs/go-graphsync/issues/82 for now, maybe will have the more-involved go-car fix for rc2


docs/examples/go-ipfs-as-a-library/go.mod
```diff
@@ -3,11 +3,13 @@ module github.com/ipfs/go-ipfs/examples/go-ipfs-as-a-library
...
-	github.com/libp2p/go-libp2p-core v0.5.2
-	github.com/libp2p/go-libp2p-peerstore v0.2.3
-	github.com/multiformats/go-multiaddr v0.2.1
+	github.com/ipfs/interface-go-ipfs-core v0.4.0
+	github.com/libp2p/go-libp2p-core v0.6.0
+	github.com/libp2p/go-libp2p-peerstore v0.2.6
+	github.com/multiformats/go-multiaddr v0.2.2
 )
+
+replace github.com/ipfs/go-ipfs => ./../../..
```
aschmahmann, line 15: Interesting. Obviously we need this for now (i.e. until the go-ipfs 0.7.0 release) in order to get things to work, but should we just keep this replace directive here since it's an example?

go.mod
```diff
@@ -55,7 +55,7 @@ require (
 	github.com/ipfs/go-unixfs v0.2.4
 	github.com/ipfs/go-verifcid v0.0.1
 	github.com/ipfs/interface-go-ipfs-core v0.4.0
-	github.com/ipld/go-car v0.1.0
+	github.com/ipld/go-car v0.1.1-0.20200429200904-c222d793c339
```
aschmahmann, line 58: Is there a release of go-car we can use here?

ribasushi: No, this is what I was referring to in my original comment: go-car needs a bit more work for a release.

docs/examples/go-ipfs-as-a-library/go.mod
ribasushi: In e.g. go-car it is permanent: https://github.com/ipld/go-car/blob/v0.1.0/car/go.mod#L11
I don't mind either way, we can even remove it right now... ( there are no tests against it )

go.mod
ribasushi: @aschmahmann this is go-car master fwiw, so you are not depending on some random commit. I need to sneak in a particular fix in ( no API changes ), and that's another day or 2 away.

aschmahmann: So should I be merging this and then we can update go-car later or am I waiting to merge on a go-car release?

ribasushi: Let's merge this now, I won't be ready in a reasonable timeframe.

PR #7599, opened 2020-08-17
Move key rotation command to ipfs key rotate

aschmahmann:


core/commands/keystore.go
```diff
@@ -413,6 +417,105 @@ var keyRmCmd = &cmds.Command{
...
+	},
+	Run: func(req *cmds.Request, res cmds.ResponseEmitter, env cmds.Environment) error {
+		cctx := env.(*oldcmds.Context)
+		nBitsForKeypair, nBitsGiven := req.Options[keyStoreSizeOptionName].(int)
+		algorithm, _ := req.Options[keyStoreTypeOptionName].(string)
+		oldKey, ok := req.Options[oldKeyOptionName].(string)
+		if !ok {
+			return fmt.Errorf("keystore name for backing up old key must be provided")
+		}
+		return doRotate(os.Stdout, cctx.ConfigRoot, oldKey, algorithm, nBitsForKeypair, nBitsGiven)
```
aschmahmann, line 466: I'm using Writer's here and Stdout to mimic what we do in `ipfs init`. We could potentially return objects like we do for other commands here, but the user could just call `ipfs id` to get the new ID, so there's not much lost.

core/commands/keystore.go
```diff
@@ -413,6 +418,105 @@ var keyRmCmd = &cmds.Command{
...
+environment variable:
+
+    export IPFS_PATH=/path/to/ipfsrepo
+`,
+	},
+	Arguments: []cmds.Argument{},
+	Options: []cmds.Option{
+		cmds.StringOption(oldKeyOptionName, "o", "Keystore name for the old/rotated-out key."),
+		cmds.StringOption(keyStoreTypeOptionName, "t", "type of the key to create: rsa, ed25519").WithDefault(keyStoreAlgorithmDefault),
+		cmds.IntOption(keyStoreSizeOptionName, "s", "size of the key to generate"),
```
aschmahmann, line 439: Not sure what I should do here for the option names. Previously, `type, t` and `size, s` were copied from `ipfs init` and used `algorithm, a` and `bits, b`.  It seems to me like keeping these names consistent with `ipfs key gen` is the most reasonable.

We could also just add more aliases (e.g. `type, t, algorithm, a`), but this seems excessive for a new command. Thoughts?

namesys/ipns_resolver_validation_test.go
```diff
@@ -146,7 +146,7 @@ func genKeys(t *testing.T, keyType int) (ci.PrivKey, peer.ID, string, string) {
 	return sk, id, PkKeyForID(id), ipns.RecordKey(id)
 }
 
-func createIPNSRecordWithEmbeddedPublicKey(sk ci.PrivKey, val []byte, seq uint64, eol time.Time) (*ipns_pb.IpnsEntry, error){
+func createIPNSRecordWithEmbeddedPublicKey(sk ci.PrivKey, val []byte, seq uint64, eol time.Time) (*ipns_pb.IpnsEntry, error) {
```
aschmahmann, line 149: go fmt that was ignored in a previous commit... my bad. Need to get linting rules better set up.

core/commands/keystore.go
Stebalien: I agree. I'd just do what `ipfs key gen` does and break things (we haven't cut a release yet anyways).

core/commands/keystore.go
```diff
@@ -413,6 +418,105 @@ var keyRmCmd = &cmds.Command{
...
+	},
+	NoRemote: true,
+	PreRun: func(req *cmds.Request, env cmds.Environment) error {
```
Stebalien, line 442: nit: I'm not sure if this is actually necessary. Opening the repo below should take care of this.

core/commands/keystore.go
Stebalien: Seems reasonable. If it's an issue later, we can always fix this to make `--enc=json` work properly.

PR #7609, opened 2020-08-19
fix: make assets deterministic

Stebalien:
Set the "modtime" to the start of the IPFS project, and the mode to 0644.


lidel (commented):
:heart:

PR #7612, opened 2020-08-20
fix: ipfs pin ls - ignore pins that have errors

aschmahmann:


Stebalien:
Shouldn't we be returning the error?

core/commands/pin.go
```diff
@@ -499,6 +499,9 @@ func pinLsAll(req *cmds.Request, typeStr string, api coreiface.CoreAPI, emit fun
 	}
 
 	for p := range pins {
+		if p.Err() != nil {
+			continue
```
aschmahmann, line 503: @stebalien we could return errors but it doesn't look like we have anything to handle those. No idea if anything upstream it's prepared to actually handle the errors.

This might be easier to sneak into the RC since there won't be any surprises, just fewer issues.

Stebalien: I meant, we could just return an error and abort. IIRC, we'll only get one error anyways.

aschmahmann: That seems like it should work

core/commands/pin.go
```diff
@@ -499,6 +499,9 @@ func pinLsAll(req *cmds.Request, typeStr string, api coreiface.CoreAPI, emit fun
 	}
 
 	for p := range pins {
+		if p.Err() != nil {
+			return err
+		}
```
aschmahmann, line 504: @Stebalien Is this what you were looking for?

Stebalien:
Also relevant https://github.com/ipfs/interface-go-ipfs-core/pull/62.

PR #7614, opened 2020-08-21
Add DHT option for WAN with no diversity

petar:
Needed in local tests, where diversity does not hold.


aschmahmann:
Stale and no longer needed. Can always reopen if this becomes useful

PR #7617, opened 2020-08-22
Added new plugin type: PluginScoreLedger

wolneykien:
With a PluginScoreLedger plugin it is now possible to replace the
default BitSwap decision logic with a different one. The plugin
support depends on the experimental BitSwap option `WithScoreLedger`
— see `80c1a81b63e5` and related commits in the `go-bitswap` project
and also `ebf11a6556cc` commit in the `go-ipfs-config` project.

A map of score ledgers and new `RegisterScoreLedger` function were
added to `core/node/core.go` in order to support score ledger
injection. The selection of the particular ledger is now possible with
enchanced `OnlineExchange()` function accepting the name of the
ledger (`""` is used for the default behavior). It is configured with
`Experimental.WithScoreLedger` configuration option.

In order to check this modifications and plugin loading two new tests
were added to `test/integration` package: `TestScoreLedgerNotLoaded`
and `TestScoreLedgerLoadStartStop`. To minimize code duplication I've
decided to slightly modify the `core/mock/mock.go` adding new
`NewMockNodeWithConfig()` function there. The code that loads and
injects plugin objects is placed in the `plugins_common_test.go` to
help implementing other plugin tests.

### DIFF
--- a/core/mock/mock.go
+++ b/core/mock/mock.go
@@ -26,13 +26,26 @@ import (
 
 // NewMockNode constructs an IpfsNode for use in tests.
 func NewMockNode() (*core.IpfsNode, error) {
+	return NewMockNodeWithConfig(nil)
+}
+
+// Constructs an IpfsNode with the specified configuration for use in
+// tests.
+func NewMockNodeWithConfig(conf *config.Config) (*core.IpfsNode, error) {
 	ctx := context.Background()
 
-	// effectively offline, only peer in its network
-	return core.NewNode(ctx, &core.BuildCfg{
+	bcfg := &core.BuildCfg{
 		Online: true,
 		Host:   MockHostOption(mocknet.New(ctx)),
-	})
+	}
+	if conf != nil {
+		bcfg.Repo = &repo.Mock{
+			D: syncds.MutexWrap(datastore.NewMapDatastore()),
+			C: *conf,
+		}
+	}
+	// effectively offline, only peer in its network
+	return core.NewNode(ctx, bcfg)
 }
 
 func MockHostOption(mn mocknet.Mocknet) libp2p2.HostOption {
--- a/core/node/core.go
+++ b/core/node/core.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	"github.com/ipfs/go-bitswap"
+	"github.com/ipfs/go-bitswap/decision"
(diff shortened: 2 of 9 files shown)


aschmahmann:
> I see that 2 sharness tests have failed due to some expectations in stat0 and stat1 files. But I don't know how to see the actual content of that files.

If you run the individual sharness tests manually on your machine (e.g. run `make test_sharness_deps` in the go-ipfs folder, and then `./t0125-twonode -v [note, you can add -i to terminate the test on first error]) then when the test ends there should be a trash folder where you can examine the files.

wolneykien:
> If you run the individual sharness tests manually on your machine (e.g. run `make test_sharness_deps` in the go-ipfs folder, and then `./t0125-twonode -v [note, you can add -i to terminate the test on first error]) then when the test ends there should be a trash folder where you can examine the files.

Thanks,  @aschmahmann . The `twonode` test seems to be unstable: in the first run it said "failed 8 among 63", but "passed all 63 test(s)" in the second run.

wolneykien:
Just have synced this branch with `go-ipfs/master` and it has passed sharness on my node.

aschmahmann:
@wolneykien the currently supported plugins are:
- Datastore
- Tracer
- IPLD
- Daemon
- DaemonInternal

Adding a specific plugin for one option type in Bitswap seems like overkill to me, and would open the door to adding new plugins for many many more features. The ones we have so far have been useful for many community projects.

You could accomplish what you're looking for by just building go-ipfs as a library and passing in a custom Bitswap constructor now that your Bitswap PR has landed. You may also be able to get away with a DaemonInternal plugin.

wolneykien:
> Adding a specific plugin for one option type in Bitswap seems like overkill to me,

May be you're right. Now, I'm not sure about a _specific_ plugin type. However, I think it would be nice to have a way to pass that (and any other!) option to `Bitswap` from an _ordinary_ plugin, say, from a `Daemon` plugin, isn't it?

> You could accomplish what you're looking for by just building go-ipfs as a library

Thanks, that seems to be a good idea!

> You may also be able to get away with a `DaemonInternal` plugin.

Yes. Anyway, for my current experiments I need to access the P2P `Host` interface that is inaccessible from `CoreAPI`.

wolneykien:
#7653 and #7652 sound great! Thanks.

PR #7620, opened 2020-08-24
sharness: use dnsaddr example in ipfs p2p command tests

aschmahmann:


test/sharness/t0180-p2p.sh
```diff
@@ -84,7 +84,7 @@ test_expect_success 'S->C Setup(dnsaddr/addr/p2p/peerID) client side' '
 '
 
 test_expect_success 'S->C Setup(dnsaddr/addr) client side' '
-  ipfsi 1 p2p forward /x/p2p-test /ip4/127.0.0.1/tcp/10104 /dnsaddr/cluster.ipfs.io 2>&1 > dialer-stdouterr.log
```
aschmahmann, line 87: I only changed this one and not the one above since that one has multiple addresses. It turns out it doesn't really matter what address this record points to (online or not) for this test, so as long as this example record is stable with a single entry we should be fine.

PR #7623, opened 2020-08-25
docs(config): fix Peering header

lidel:
This PR  fixes the header for `Peering`, which was under pubsub for some reason.
That looked like a mistake: Peering is top level namespace.

cc @aschmahmann


jacobheun (approved):
lgtm

PR #7625, opened 2020-08-26
Fix #7624: Do not fetch dag nodes when checking if a pin exists

### LINKED ISSUE #7624: pin ls $cid hits the network for non pinned $cid
Version: master, ipfs 0.6.0

`ipfs pin ls --type recursive $cid` seems to hit the network for a non pinned CID.

This is quite breaking and a regression from 0.5.0. IPFS Cluster relies on this command to check if something is pinned. When there are no providers for an item, everything will hang.

I will bisec this in a minute.

hsanjuan:


core/coreapi/pin.go
```diff
@@ -57,9 +57,9 @@ func (api *PinAPI) Ls(ctx context.Context, opts ...caopts.PinLsOption) (<-chan c
 }
 
 func (api *PinAPI) IsPinned(ctx context.Context, p path.Path, opts ...caopts.PinIsPinnedOption) (string, bool, error) {
-	dagNode, err := api.core().ResolveNode(ctx, p)
+	resolved, err := api.core().ResolvePath(ctx, p)
```
aschmahmann, line 60: @hsanjuan this is better, but I think still has the bad behavior if you do `ipfs pin ls QmAbc/field` where `field` points to `QmXyz` 

https://github.com/ipfs/go-ipfs/blob/7d95359c3e01c85d8fe394b1e24769353d74c6bf/core/coreapi/path.go#L68

calls

https://github.com/ipfs/go-path/blob/22a2352f71a8a53598e50ff04107a8ff0619af1f/resolver/resolver.go#L60

which calls

https://github.com/ipfs/go-path/blob/22a2352f71a8a53598e50ff04107a8ff0619af1f/resolver/resolver.go#L92

We may need a PR to go-path as well.

aschmahmann: fixed by ipfs/go-path#30

PR #7633, opened 2020-08-30
fix: ipfs dht put/get commands with peerIDs encoded as CIDs

aschmahmann:


core/commands/dht.go
```diff
@@ -676,20 +675,15 @@ func printEvent(obj *routing.QueryEvent, out io.Writer, verbose bool, override p
 
 func escapeDhtKey(s string) (string, error) {
 	parts := path.SplitList(s)
-	switch len(parts) {
-	case 1:
-		k, err := b58.Decode(s)
-		if err != nil {
-			return "", err
-		}
-		return string(k), nil
```
aschmahmann, line 685: What was this ever for?

core/commands/dht.go
```diff
@@ -676,20 +675,15 @@ func printEvent(obj *routing.QueryEvent, out io.Writer, verbose bool, override p
...
-	case 3:
-		k, err := b58.Decode(parts[2])
-		if err != nil {
-			return "", err
-		}
-		return path.Join(append(parts[:2], string(k))), nil
-	default:
+	if len(parts) != 3 ||
+		parts[0] != "" ||
+		!(parts[1] == "ipns" || parts[1] == "pk") {
```
aschmahmann, line 680: There doesn't seem like there's a reason to support anything other than these two namespaces. Should we even support querying for `/pk` anymore?

Stebalien: I wouldn't hard-code these. There's no reason not to just let the DHT handle it (in case we add a new validator).

core/commands/dht.go
Stebalien: Maybe for find-provider? But that's not an issue anymore.

core/commands/dht.go
aschmahmann: I think I'd rather this yell at us early. The problem is that we're using `peer.Decode(parts[2])` which seems unlikely to be valid with any other validator. This is a little more informative than a decoding error.

Stebalien: Up to you. I just don't like having checks at multiple points. It's nice to be able to add a new record type without having to modify this check here.

PR #7658, opened 2020-09-07
add 'ipfs repo migrate' command

### LINKED ISSUE #7471: ipfs migration without starting the daemon
Currently, it seems to be impossible to do a migration (if necessary) without starting the daemon. 

I'm a package maintainer and like script the migration after installation of the new package instead of running the migration on the first run of the daemon, which - as discussed earlier in a PR - might hit the timeout of systemd for starting a service.

I like to have an `ipfs migrate --assume-yes` or something like this for unattended migration while updating a package with a package manager.

It should return 0 if everything is fine, regardless if a migration was necessary or not.

zaibon:
this command allows to run the repo migration without starting the daemon.

resolves #7471

### DIFF
--- a/test/sharness/t0066-migration.sh
+++ b/test/sharness/t0066-migration.sh
@@ -90,11 +90,11 @@ test_expect_success "ipfs repo migrate succeed" '
 
 test_expect_success "output looks good" '
   grep "Found outdated fs-repo, starting migration." migrate_out > /dev/null &&
-  grep "Success: fs-repo has been migrated to version 10" migrate_out > /dev/null
+  grep "Success: fs-repo has been migrated to version 11" migrate_out > /dev/null
 '
 
-test_expect_success "manually reset repo version to 10" '
-  echo "10" > "$IPFS_PATH"/version
+test_expect_success "manually reset repo version to 11" '
+  echo "11" > "$IPFS_PATH"/version
 '
 
 test_expect_success "detect repo does not need migration" '
--- a/test/sharness/t0066-migration.sh
+++ b/test/sharness/t0066-migration.sh
@@ -90,11 +90,11 @@ test_expect_success "ipfs repo migrate succeed" '
 
 test_expect_success "output looks good" '
   grep "Found outdated fs-repo, starting migration." migrate_out > /dev/null &&
-  grep "Success: fs-repo has been migrated to version 11" migrate_out > /dev/null
+  grep "Success: fs-repo migrated to version $IPFS_REPO_VER" true_out > /dev/null
 '
 
 test_expect_success "manually reset repo version to 11" '
-  echo "11" > "$IPFS_PATH"/version
+  echo "$IPFS_REPO_VER" > "$IPFS_PATH"/version
 '
 
 test_expect_success "detect repo does not need migration" '
--- a/test/sharness/t0066-migration.sh
+++ b/test/sharness/t0066-migration.sh
@@ -84,4 +84,26 @@ test_expect_success "output looks good" '
   grep "Please get fs-repo-migrations from https://dist.ipfs.io" daemon_out > /dev/null
 '
 
+test_expect_success "ipfs repo migrate succeed" '
+  test_expect_code 0 ipfs repo migrate > migrate_out
+'
+
+test_expect_success "output looks good" '
+  grep "Found outdated fs-repo, starting migration." migrate_out > /dev/null &&
+  grep "Success: fs-repo has been migrated to version 10" migrate_out > /dev/null
+'
+
+test_expect_success "manually reset repo version to 10" '
+  echo "10" > "$IPFS_PATH"/version
+'
+
+test_expect_success "detect repo does not need migration" '
+  test_expect_code 0 ipfs repo migrate > migrate_out
+'
+
+test_expect_success "output looks good" '
+  grep "Repo does not require migration" migrate_out > /dev/null
+'
+cat migrate_out
+
 test_done
--- a/core/commands/commands_test.go
+++ b/core/commands/commands_test.go
@@ -219,6 +219,7 @@ func TestCommands(t *testing.T) {
 		"/repo/stat",
 		"/repo/verify",
 		"/repo/version",
+		"/repo/migrate",
 		"/resolve",
 		"/shutdown",
 		"/stats",
--- a/core/commands/repo.go
+++ b/core/commands/repo.go
@@ -408,7 +408,7 @@ var repoMigrateCmd = &cmds.Command{
 			return err
 		}
 
-		fmt.Println("Repo migrated successfully.")
+		fmt.Printf("Success: fs-repo has been migrated to version %d.\n", fsrepo.RepoVersion)
 		return nil
 	},
 }
--- a/core/commands/repo.go
+++ b/core/commands/repo.go
@@ -11,11 +11,13 @@ import (
 	"sync"
 	"text/tabwriter"
 
-	humanize "github.com/dustin/go-humanize"
+	oldcmds "github.com/ipfs/go-ipfs/commands"
 	cmdenv "github.com/ipfs/go-ipfs/core/commands/cmdenv"
 	corerepo "github.com/ipfs/go-ipfs/core/corerepo"
 	fsrepo "github.com/ipfs/go-ipfs/repo/fsrepo"
+	migrate "github.com/ipfs/go-ipfs/repo/fsrepo/migrations"
 
+	humanize "github.com/dustin/go-humanize"
 	cid "github.com/ipfs/go-cid"
 	bstore "github.com/ipfs/go-ipfs-blockstore"
 	cmds "github.com/ipfs/go-ipfs-cmds"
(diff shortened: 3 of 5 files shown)


aschmahmann:
Thanks @zaibon for starting this. Could you do two more things:

1) As specified in the issue you linked we need an `--allow-downgrade` flag. While this command has one it's not hooked up to anything.
2) There is nothing testing that this command works. You can modify https://github.com/ipfs/go-ipfs/blob/83c57eda1fe4c50660761b401b4e4f016e6ecfd0/test/sharness/t0066-migration.sh to also test that `ipfs repo migrate` and `ipfs repo migrate --allow-downgrade` work as expected

Thanks again and let me know if you need any guidance here.

zaibon:
Thanks for the review @aschmahmann.

> As specified in the issue you linked we need an --allow-downgrade flag. While this command has one it's not hooked up to anything.

Right, forgot that part. What should be the behavior when this flag is set ? At the moment I always call `migrate.RunMigration(fsrepo.RepoVersion)` should I read the desired version from somewhere else and then compare the current version with the desired version and only allow downgrade if the flag is set ?

> There is nothing testing that this command works. You can modify https://github.com/ipfs/go-ipfs/blob/83c57eda1fe4c50660761b401b4e4f016e6ecfd0/test/sharness/t0066-migration.sh to also test that ipfs repo migrate and ipfs repo migrate --allow-downgrade work as expected

Yep, will do, thanks for the guidance here.

test/sharness/t0066-migration.sh
```diff
@@ -13,7 +13,7 @@ test_init_ipfs
 test_expect_success "setup mock migrations" '
   mkdir bin &&
   echo "#!/bin/bash" > bin/fs-repo-migrations &&
-  echo "echo 5" >> bin/fs-repo-migrations &&
+  echo "echo 10" >> bin/fs-repo-migrations &&
```
zaibon, line 16: Had to change this here to make the detection of the `fs-repo-migrations` to work since it seems now the code also checks the version the migration tool is able to handle.

zaibon:
@aschmahmann I hooked up the `allow-downgrade` flag. For this I had to add an extra argument to the `RunMigration` method. Not sure this is ok, I'm usually not a big fan of boolean argument. What's thought about it ?

I also added some test for the new command. I'm not sure how I can test the downgrade though, some input there would be welcome :-)

aschmahmann:
@zaibon sorry it's taking a while to get to this. A little busy with things related to the v0.7.0 release, but will try and get to this next week.

zaibon:
@aschmahmann ping ;-)

aschmahmann:
Thanks @zaibon for the reminder and apologies it slipped off my radar. I'll try and take a look over the weekend.

I was actually thinking about this PR yesterday because some of the pain around migrations has made implementing something like https://github.com/ipfs/fs-repo-migrations/issues/98 more important. This would likely rely on `ipfs repo migrate` instead of a separate `fs-repo-migrations` binary which would make this command even more useful (and make testing it simpler too).

zaibon:
Nice, I'll have a look at https://github.com/ipfs/fs-repo-migrations/issues/98.

Should I already try to rebase this PR since I guess it's a bit out of date with master or wait your review ?

aschmahmann:
> Should I already try to rebase this PR since I guess it's a bit out of date with master or wait your review 

I suspect it's an easy rebase, but if it's a pain for some reason then I'd hold off until there's some progress on ipfs/fs-repo-migrations#98. If we do basically what's recommended in that issue then this command will replace the fs-repo-migrations binary for people doing manual migration manipulation.

gammazero:
@zaibon Wanted to let you know that this PR has not been forgotten about.  The work to [overhaul the migrations and binary distribution](https://github.com/ipfs/fs-repo-migrations/issues/98#issuecomment-763329183) is code-complete, and is waiting until there is time to review it all. Once #7857 is merged, then you can rebase this PR and make any necessary changes.  Hopefully in the next couple of weeks.

zaibon:
@gammazero seems https://github.com/ipfs/fs-repo-migrations/issues/98 is now closed. Can I start rebasing and make the required change here or is there anything else in the pipeline ?

gammazero:
@zaibon You can rebase now.  You will need to make some changes to how you call `RunMigration`.  See here:
https://github.com/ipfs/go-ipfs/blob/master/cmd/ipfs/daemon.go#L291-L293

aschmahmann:
@zaibon not sure if this is ready for re-review yet, but when you're ready just request review from me or tag me

gammazero:
@zaibon I took a look at the changes to the sharness tests, and realized that what was there (before your changes) needs to be fixed.  go-ipfs no longer uses `fs-repo-migrations` to perform migrations, and the tests were not doing what they were supposed to be doing.

Attn: @aschmahmann 
I have created a PR #8053 to fix this.  It includes comments explaining what is going on as well.  Please rebase this PR as soon as it can be merged. (sorry for another roadblock).

zaibon:
@aschmahmann I just rebased on top of #8053, so this PR should be ready now.

test/sharness/t0066-migration.sh
```diff
@@ -90,11 +90,11 @@ test_expect_success "ipfs repo migrate succeed" '
 
 test_expect_success "output looks good" '
   grep "Found outdated fs-repo, starting migration." migrate_out > /dev/null &&
-  grep "Success: fs-repo has been migrated to version 10" migrate_out > /dev/null
+  grep "Success: fs-repo has been migrated to version 11" migrate_out > /dev/null
```
gammazero, line 93: This looks like a duplicate of the test at line 72-74.

You may want to take this test from the current master branch and reapply your changes.

test/sharness/t0066-migration.sh
```diff
@@ -90,11 +90,11 @@ test_expect_success "ipfs repo migrate succeed" '
...
 '
 
-test_expect_success "manually reset repo version to 10" '
-  echo "10" > "$IPFS_PATH"/version
+test_expect_success "manually reset repo version to 11" '
+  echo "11" > "$IPFS_PATH"/version
```
gammazero, line 97: The variable `IPFS_REPO_VER` holds the original version, so you should be able to do:
```sh
echo "$IPFS_REPO_VER" > "$IPFS_PATH"/version
```
This way when there is a new repo version, this test will not need to be updated.

gammazero (changes requested):

gammazero:
@aschmahmann With the recent changes to allow migrations over IPFS, and to import downloaded migrations, we probably want this PR to at least be able to fetch migrations via IPFS.  This command would also benefit from being able to import downloaded migrations as CAR files.

aschmahmann:
> we probably want this PR to at least be able to fetch migrations via IPFS. 

👍

> This command would also benefit from being able to import downloaded migrations as CAR files.

We get this basically for free if we can fetch the migrations using IPFS then if the user really wants to work with CAR files they can use `ipfs dag import <migration car>` before they run `ipfs repo migrate` and that'll do the job.

test/sharness/t0066-migration.sh
zaibon: This test is up to date with master branch, all the new code is only related this me PR.

This line looks similar to 71-74, cause it does test the same thing but it is the result of another command. In 71-74 we test the result of the `ipfs daemon` command, while here we test the result of the `ipfs migrate` command.

zaibon:
@aschmahmann Do you still need some action from me on this ?

gammazero:
@zaibon IPFS repo migrations has been a moving target, and there is additional functionality that your command should include.  This is the ability to download migrations from IPFS and from gateways according to a new Migrations section of the IPFS config.  What this means for this PR is the following:
1. Wait for PR #8153 to merge
2. Rebase this PR on master
3. Replace your code [here](https://github.com/ipfs/go-ipfs/pull/7658/files#diff-daf81b7f18381e7ad219f36448f52d0db26222ec838be8dfc37629694dc682e5R403-R404) with code based on the example [here](https://github.com/ipfs/go-ipfs/blob/chore/expose-migration-api/cmd/ipfs/daemon.go#L296-L315)

BigLep:
@zaibon : can you please rebase so we can merge this?

zaibon:
@BigLep done.

Matchjuan859:
am not sure wwhATA U NED ME TO DUE

gammazero:
I took care of [step 3 above](https://github.com/ipfs/go-ipfs/pull/7658#issuecomment-848194387) to allow this command to use IPFS to download migrations.

BigLep:
Arg, and not it looks like CI is not passing.  I have limited internet bandwidth to track it down, but I believe I sawt his mentioned somewhere else.  @aschumann is there another issue we can track against so can then merge this?

BigLep:
@gammazero : can you please rebase on master and see if it's good to go?

gammazero:
@aschmahmann @BigLep As far as I can tell, this PR is done.  The problem is that it does not have privileges to run the sharness tests using a `2xlarge` instance as configured in the `.circleci/config.yml` [here](https://github.com/ipfs/go-ipfs/blob/master/.circleci/config.yml#L124), and this is causing failure.

gammazero:
Closing: This PR is replaced by #8428 to move it into a branch to avoid CI problems associated with the repo.

PR #7659, opened 2020-09-08
Fix --ipns-base alias

aschmahmann:
This PR first refactors `--ipns-base` by putting all the options in one location and then removes the empty alias that was being used.

### DIFF
--- a/core/commands/keyencode/keyencode.go
+++ b/core/commands/keyencode/keyencode.go
@@ -1,11 +1,14 @@
 package keyencode
 
 import (
+	cmds "github.com/ipfs/go-ipfs-cmds"
 	peer "github.com/libp2p/go-libp2p-core/peer"
 	mbase "github.com/multiformats/go-multibase"
 )
 
-const IPNSKeyFormatOptionName = "ipns-base"
+const ipnsKeyFormatOptionName = "ipns-base"
+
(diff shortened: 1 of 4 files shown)


core/commands/keyencode/keyencode.go
```diff
@@ -1,11 +1,14 @@
 package keyencode
```
aschmahmann, line 1: Any ideas why code coverage is yelling at me here? Is this just because I removed some lines, or is there something sneaky that I've missed? The report looks pretty weird.

core/commands/keyencode/keyencode.go
```diff
@@ -1,11 +1,14 @@
...
 import (
+	cmds "github.com/ipfs/go-ipfs-cmds"
 	peer "github.com/libp2p/go-libp2p-core/peer"
 	mbase "github.com/multiformats/go-multibase"
 )
 
-const IPNSKeyFormatOptionName = "ipns-base"
+const ipnsKeyFormatOptionName = "ipns-base"
+
+var OptionIPNSBase = cmds.StringOption(ipnsKeyFormatOptionName, "Encoding used for keys: Can either be a multibase encoded CID or a base58btc encoded multihash. Takes {b58mh|base36|k|base32|b...}.").WithDefault("base36")
```
aschmahmann, line 11: @Stebalien Any reason why reusing the same option repeatedly is not recommended?

petar (approved):
lgtm

core/commands/keyencode/keyencode.go
Stebalien: No, we do that for some shared options. It should be fine.

Stebalien (approved):

PR #7662, opened 2020-09-09
chore: replace an fmt.Printf with a log.Infof()

moul:
the corehttp library is used by multiple external projects:

https://github.com/search?q=github.com%2Fipfs%2Fgo-ipfs%2Fcore%2Fcorehttp+-user%3Aipfs&type=Code

### DIFF
--- a/core/corehttp/corehttp.go
+++ b/core/corehttp/corehttp.go
@@ -73,9 +73,9 @@ func ListenAndServe(n *core.IpfsNode, listeningMultiAddr string, options ...Serv
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


aschmahmann:
@hsanjuan is there a way to do this more uniformly across repos?

Stebalien:
This is going to be a pretty significant breaking change to downstream projects. I agree that the design is sub-par, but I'm not sure if _this_ change is worth it (given that it's easy to just call `Listen` then `Serve`).

aschmahmann:
Looks like we can close this, lmk if I'm missing something and feel free to reopen.

PR #7664, opened 2020-09-09
chore: bump go-ipfs-cmds

aschmahmann:


lidel (approved):
Tested with `curl -X POST -A "Mozilla Electron" 127.0.0.1:5001/api/v0/id` and https://github.com/ipfs/go-ipfs-cmds/pull/201 works as expected (no longer 403)

PR #7665, opened 2020-09-09
Revert Go trimpath

aschmahmann:


mk/golang.mk
```diff
@@ -10,9 +10,9 @@ GOTFLAGS ?=
 
 # Unexport GOFLAGS so we only apply it where we actually want it.
 unexport GOFLAGS
-# Try to make building as reproducible as possible by stripping the go path.
 # Override so we can combine with the user's go flags.
-override GOFLAGS += -trimpath
+# Try to make building as reproducible as possible by stripping the go path.
+override GOFLAGS += "-asmflags=all='-trimpath=$(GOPATH)'" "-gcflags=all='-trimpath=$(GOPATH)'"
```
aschmahmann, line 15: We could also just remove these 3 lines entirely and not worry about reproducible builds that aren't using our `dist` build script from https://github.com/ipfs/distributions. At some point we could add in some make flag like `make build-reproducible` that uses the dist build script.

However, leaving this in and keeping behavior the same seems reasonable for now.

Stebalien (approved):

PR #7673, opened 2020-09-15
Gateway proxying for {cid}.ipfs and {cid}.ipns hostnames

arthuredelstein:
Hi! I have been looking for a way to reduce the friction for users to get started browsing IPFS websites, in order to encourage more widespread adoption. I was wondering if there is an URI format for IPFS where:

* URIs for IPFS websites are relatively simple, don't include a gateway address or require redirecting to localhost
* URIs are compatible with existing browsers
* Users' privacy is protected when they browse across different IPFS sites
* Users aren't required to install a native "helper app" to browse IPFS sites "natively"

An approach that seems to address these issues is to use a URI format whereby:
* ipfs hashes are written as http://{CIDv1}.ipfs/...
* ipns hashes are written as http://{CIDv1}.ipns/...

For example, the Wikipedia mirror URI is written as http://bafybeiemxf5abjwjbikoz4mc3a3dla6ual3jsgpdr4cjr3oz3evfyavhwq.ipfs/wiki/

I believe this new address format has a few advantages:

* There's no gateway domain, so users can immediately see that an IPFS site does not belong to a particular gateway, but rather stands on its own. The URI format already looks mostly familiar to new users because it begins with http://, and suggests it might be a website in the "ipfs" category.

* The address format is already compatible with browsers, and http:// URIs are automatically turned into clickable links by, Google Docs, Github, etc.

* Web browsers are moving toward isolating state (including cookies and supercookies) by eTLD+1. Once that is implemented, each {CIDv1}.ipfs domain will automatically gets its own cookie jar.

* Because we are using the proxy mechanism, a WebExtensions add-on is enough to get these URIs to connect: no native helper app is needed to make this address format work in browsers that support WebExtensions.

It turns out that, to get this URI format to work in most browsers, we can use common browser functionality. A small patch (this pull request) enhances the gateway so that it acts as an HTTP(S) proxy for such URIs. (The patch doesn't interfere with any of the existing functionality of go-ipfs.) To start browsing, the user needs only to install an add-on that makes a proxy connection to a go-ipfs gateway (remote or local) specifically for requests to IPFS in the proposed URI format. As a proof of principle, I wrote a simple add-on for Firefox, Chrome, and Edge that makes use of this proxy functionality:

https://github.com/arthuredelstein/dweb-proxy/

If you want to try it yourself, here's the installable add-on for Firefox: https://arthuredelstein.net/dweb/. It takes 60 seconds to try, and on installation provides some helpful example URIs for testing. And if you run ./ipfs daemon locally, you can direct the add-on to proxy its connections through http://localhost:8080, to support the same URI format.

Here's a screenshot of the add-on in use: 
image

Helpfully, the gateway already works as a proxy for .eth domains. So you can also enter http://ethereum.eth in the browser's address bar and it just works.

My goal is to make this patch suitable for existing public gateways (such as cloudflare-ipfs.com) that can handle larger numbers of users. Then any add-on supporting this URI format will be able to offer performant support at scale.

I also think this approach helps to demonstrate that native IPFS support in browsers won't need to be a major lift. That is: no significant modifications to browser internals, apart from a p2p networking component, will be required to support this URI format.

I hope this pull request is of interest! I'd be glad to discuss this patch further and make any revisions as needed.

### DIFF
--- a/core/corehttp/hostname.go
+++ b/core/corehttp/hostname.go
@@ -78,6 +78,21 @@ func HostnameOption() ServeOption {
 				host = xHost
 			}
 
+			// Are we being asked to proxy a {cid}.ipfs or {cid}.ipns host?
+			if strings.HasSuffix(host, ".ipfs") || strings.HasSuffix(host, ".ipns") {
+				labels := strings.Split(host, ".")
+				if len(labels) == 2 {
+					rootID := labels[0]
+					ns := labels[1]
+					if _, err := cid.Decode(rootID); err == nil {
+						r.URL.Path = "/" + ns + "/" + rootID + r.URL.Path
+						// Serve path request
+						childMux.ServeHTTP(w, r)
+						return
+					}
+				}
+			}
+
 			// HTTP Host & Path check: is this one of our  "known gateways"?
 			if gw, ok := isKnownHostname(host, knownGateways); ok {
 				// This is a known gateway but request is not using


lidel (changes requested):
Thank you for this writeup @arthuredelstein 

I apologize for long response, had no bandwidth to make it shorter, but I hope its digestible and useful reference for people interested in this problem space. 

If not, just jump to "Suggested direction for this PR" where I suggest a much easier way of enabling proxy mode by using `dweb` as a subdomain hostname :) 


### Crash course on why we don't use TLD-based addressing

We've briefly looked into this before (https://github.com/ipfs/in-web-browsers/issues/109), but this approach has some real deal-breakers:

- > A small patch (this pull request) enhances the gateway so that it acts as an HTTP(S) proxy for such URIs.
  - :broken_heart:  Browsers will mark `http://*.ipfs` as insecure context because `http://` without TLS is used on non-loopback address.
    This means:
      - :broken_heart: bad UX of broken padlock  / "insecure" warnings in address bar
        - sidenote: Firefox incorrectly does this for subdomains atm, see [bug 1220810](https://bugzilla.mozilla.org/show_bug.cgi?id=1220810). Chromium does the right thing.
      - :broken_heart: [no new Web APIs](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts/features_restricted_to_secure_contexts)
      - :broken_heart: cross-origin requests will trigger mixed-content errors.
      - :broken_heart: proxying `https://` won't work because there is no TLS cert

- :broken_heart: `.ipfs` and `.ipns` are not real TLDs, nor :broken_heart:  there is a RFC that makes them a [special TLD like `.onion`](https://tools.ietf.org/html/rfc7686)
  - :broken_heart: no real TLD means ICANN can decide they want to sell them at one point either IPFS project will have to pay them whatever amount they want, or :broken_heart: someone else buys this and we have a problem of namespace conflict between this PR and how TLD resolves over regular DNS.
  - :dizzy:  :thought_balloon:  The idea of  "special TLD" RFC was explored in https://github.com/arewedistributedyet/arewedistributedyet/issues/34. Received little interest because at the end of the day it hinders the adoption by requiring customized web browser (`.onion` works out-of-the-box only in Tor Browser etc) and if that is the case its better to invest time in non-`http://` URIs than trying to fit IPFS into security context build around HTTP and client-server relationship.

- > Because we are using the proxy mechanism, a WebExtensions add-on is enough to get these URIs to connect: no native helper app is needed to make this address format work in browsers that support WebExtensions.
  - :broken_heart: works, but can't be "the canonical experience" because it  breaks local gateway: it no longer works out of the box and requires additional the setup (instead of using localhost URI in any user agent, user needs to manually set up proxy or install the browser extension, which is only available for Chromium and Firefox)




Thus, the TLD approach was discarded as "canonical way of addressing on the web" and we started looking into new URIs:

- :green_heart: `ipfs://` URI/URN  delivers better UX 
- :green_heart:  `ipfs://` and `ipns://` got [approved by IANA](https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml)  this year. 
  - While HTTP can be used behind the scenes, that becomes an implementation detail. 
  - On the addressing level there is no dependency on location and  there is no HTTP in UI:
    - We are no longer forced to rely on TLS/PKI for secure context and those new URIs can have own rules on how Origin-based security perimeter is defined. 
   -  :green_heart:  Opera for Android ships with support for those URIs that stay in the address bar and each CID gets own Origin
   - :green_heart:  Brave is working on native support too, when `ipfs://` is entered they will prompt user if they want to start own node, or open resources using a gateway – https://github.com/brave/brave-browser/issues/10220
   
# Suggested direction for this PR: reuse subdomain feature

I am against adding special code for handling `.ipfs` and `.ipns` – it adds unnecessary complexity, and gateway code is already more complex than we initially wanted. 

However, running a subdomain gateway at `.ipfs.dweb` and `.ipns.dweb` is possible without any patch. And it will work in proxy mode too. 

You can enable it via:

```console
$ ipfs config --json Gateway.PublicGateways '{
    "dweb": {
      "UseSubdomains": true,
      "Paths": ["/ipfs", "/ipns"]
    }
  }'
```

If you restart the ipfs daemon to reload config, you should be able to open  `http://bafybeiemxf5abjwjbikoz4mc3a3dla6ual3jsgpdr4cjr3oz3evfyavhwq.ipfs.dweb` in proxy mode just fine.

What I suggest to do here is to change this PR to be just one line:

```diff
var defaultKnownGateways = map[string]*config.GatewaySpec{
	"localhost":       subdomainGatewaySpec,
	"ipfs.io":         pathGatewaySpec,
	"gateway.ipfs.io": pathGatewaySpec,
	"dweb.link":       subdomainGatewaySpec,
+	"dweb":            subdomainGatewaySpec,
}
```

That way the `ipfs config` command would not be necessary to enable proxy mode for `http://*.ipfs.dweb` URLs, it will be available by default.

# TL;DR 

- NO for `.ipfs` but OK for PR with  one-liner mentioned above to enables subdomain gateway at `*.ipfs.dweb` 
- While `.dweb` subdomains won't be a canonical way of addressing,
  - it is a good demonstration of proxy mode without hitting TLS errors caused by HSTS at `dweb.link` or `ipfs.io`
  - could make it easier for some user agents to shipp "native" IPFS sooner (set up http proxy, quick hack to mark everything loaded from `*.dweb` as secure context, wire up `ipfs://` to use `*.dweb` behind the scenes  – pretty  clean separation of concerns, Origins look better in console when debugging :sparkles: )

aschmahmann:
closing this as it seems like @lidel's 1 line suggestion seems like the way to do this if we wanted to go this route. If this is still useful to you feel free to open a new PR taking that approach.

PR #7677, opened 2020-09-16
feat(gateway): Content-Disposition improvements

### LINKED ISSUE #7648: Add fallback for old browsers when using ?filename parameter
When using the `?filename` parameter, the IPFS gateway adds a `Content-Disposition` header that is RFC 5987 encoded. 

For instance, if I passed `?filename=hello.zip` IPFS gives this as the `Content-Disposition`:

```
content-disposition: inline; filename*=UTF-8''hello.zip
```

However, not all browsers/clients correctly support RFC 5987. This can lead to unexpected behaviour with the filename parameter seemingly being ignored or treated incorrectly. 

The typical fallback is to provide both a `filename` and `filename*` in the `Content-Disposition`:

```
content-disposition: inline; filename="hello.zip"; filename*=UTF-8''hello.zip
```

It would be good if the IPFS gateway did this to increase compatibility. 

I'm not super familiar with Go, but I could have a go at this?

lidel:
This PR improves the way gateway supports `Content-Disposition`, an HTTP header used by browsers as a name hint when saving a file:

- ASCII-only fallback for old browsers that do not implement RFC 5987 (Closes #7648)
- Support for direct download mode with `Content-Disposition: attachment` when `?filename` is accompanied by `&download=true` 

This is a fully backward-compatible change, sharness tests included.

cc (https://github.com/ipfs/go-ipfs/pull/4177  – @voker57 @crackcomm @radfish) https://github.com/ipfs/in-web-browsers/issues/132 @andrew @gozala @autonome

### DIFF
--- a/core/corehttp/gateway_handler.go
+++ b/core/corehttp/gateway_handler.go
@@ -34,6 +34,8 @@ const (
 	ipnsPathPrefix = "/ipns/"
 )
 
+var onlyAscii = regexp.MustCompile("[[:^ascii:]]")
+
 // gatewayHandler is a HTTP handler that serves IPFS objects (accessible by default at /ipfs/<path>)
 // (it serves requests like GET /ipfs/QmVRzPKPzNtSrEzBFm2UZfxmPAgnaLke4DMcerbsGGSaFe/link)
 type gatewayHandler struct {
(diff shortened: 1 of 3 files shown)


achingbrain (approved):

aschmahmann (approved):
Looks good, thanks 🙏

PR #7705, opened 2020-09-30
Badger2 integration

gammazero:
This is a provisional PR and should not be merged until dependencies are merged into their master branches.

- Add badger2 plugin
- Update docs to describe badger2
- Add tests for badger2
- Set dependencies to pending PRs with necessary updates


docs/datastores.md
```diff
@@ -51,6 +51,22 @@ Uses [badger](https://github.com/dgraph-io/badger) as a key value store.
...
+
+* `syncWrites`: Flush every write to disk before continuing. Setting this to false is safe as go-ipfs will automatically flush writes to disk before and after performing critical operations like pinning. However, you can set this to true to be extra-safe (at the cost of a 2-3x slowdown when adding files).
+* `truncate`: Truncate the DB if a partially written sector is found (defaults to true). There is no good reason to set this to false unless you want to manually recover partially written (and unpinned) blocks if go-ipfs crashes half-way through a adding a file.
+
+```json
+{
+	"type": "badger2ds",
+	"path": "<location of badger2 inside repo>",
+	"syncWrites": true|false,
+	"truncate": true|false,
```
aschmahmann, line 66: Should add a compression flag here (and in the various config locations) if it's not an init-only parameter.

test/sharness/t0025-datastores.sh
```diff
@@ -13,4 +13,17 @@ test_expect_success "'ipfs pin ls' works" '
   ipfs pin ls | wc -l | grep 9
 '
 
+test_expect_success "cleanup repo" '
+  rm -rf "$IPFS_PATH"
+'
+
+test_expect_success "'ipfs init --profile=badger2ds' succeeds" '
+  BITS="2048" &&
```
aschmahmann, line 21: BITS can be removed here and above, as they're no longer needed.

plugin/plugins/badger2ds/badger2ds.go
```diff
@@ -0,0 +1,127 @@
+package badger2ds
+
```
aschmahmann, line 2: If Badger isn't going to fix their x86 issues (especially the compilation error), then we should split this package into two. One with a build directive for `386` (or easier just calling the file badger2ds_386.go) that just returns errors and the other with a `// +build !386` directive.

That should mostly get things to work, although to get tests to pass we'll also need to add an exception to not run the badger2 sharness tests on 386.

aschmahmann (changes requested):

plugin/plugins/badger2ds/badger2ds.go
gammazero: Badger has fixed the build failures, which allows out builds to work.  Config changes on our side allow our tests to pass. So, we should be ok, and we are completely avoiding the mmap problem now so we do not need to issue the warning about 32bit.  Badger should fix their 32bit tests, but we are avoiding the cause of those failures (mmap + 32bit).

docs/datastores.md
gammazero: Maybe this should be a separate PR.  If we do this then we should also add in options for encryptions and cache size.  The cache size is needed, because the badger2 documentation recommends using a cache if either compression or encryption is enabled. And if those are enabled, then it may be responsible to add some IPFS tests with those options.

test/sharness/t0025-datastores.sh
gammazero: Done

docs/datastores.md
gammazero: Done

aschmahmann:
@gammazero your go.mod files are conflicting with master. Could you rebase when you get a chance?

repo/fsrepo/config_test.go
```diff
@@ -234,3 +246,81 @@ func TestMeasureConfig(t *testing.T) {
...
+	}
+
+	ds, err := dsc.Create(dir)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// The name of the package in go-ds-badger2 is "badger", so the type
+	// returned by dsc.Create() is "*badger.Datastore"
+	if typ := reflect.TypeOf(ds).String(); typ != "*badger.Datastore" {
```
aschmahmann, line 323: would using reflect.Typeof(ds).PkgPath() help disambiguate here?

docs/datastores.md
```diff
@@ -51,6 +51,26 @@ Uses [badger](https://github.com/dgraph-io/badger) as a key value store.
...
+* `blockCacheSize`: Specifies how much data cache should hold in memory. If compression is disabled, adding a cache leads to unnecessary overhead which may affect read performance.
+
+```json
+{
+	"type": "badger2ds",
+	"path": "<location of badger2 inside repo>",
+	"syncWrites": true|false,
+	"truncate": true|false,
+	"compression": "none"|"snappy"|"zstd1"|"zstd2"|"zstd3",
+	"blockCacheSize": <size of block cache>
```
aschmahmann, line 70: WDYT about having some value for each of these signifying the default (e.g. in https://github.com/ipfs/go-ipfs/blob/master/docs/config.md#types we have some default values). This would allow us (the implementers) to change the default compression and blockCacheSize without users needing to be change their configs. Of course, if they wanted to fix these parameters they could explicitly put them in the config file.

My thinking is to use `null or missing` to mean "apply the default value" as we do with Flag and Priority and then define what our current default values are (no compression, no blockCacheSize).

aschmahmann (changes requested):
Looking good. Added a couple small comments and a suggestion on the compression + blockCacheSize parameters.

docs/datastores.md
gammazero: I would prefer that anytime a string parameter is missing or has a null (or empty `''"`) value, that this means use the default of our choosing (which could change with updates) -- _when appropriate_.  There are some cases where a value cannot change with an upgrade, such as a filesystem location. So, for these we cannot use a default for a missing or null. In those cases, we must supply a default value at initialization time, and report and error if the value is ever unspecified.

So, I do not think we can make a general statement that all missing/null string values get set to some default. This needs to be done on a case-by-case basis.  Some strings can be missing/null and a reasonable default will be used, and for others an initial value is provided but a value must always be present.

So, to address this, I have amended the description specifically for the new badger2 config values, that omitting the values from the configuration results in choosing the default setting.

repo/fsrepo/config_test.go
gammazero: Unfortunately `PkgPath()` always returns `""` here.  I cannot yet find a clear explanation why, but it appears to do with the way interfaces are embedded.  However, if we phase-out badger1 then this will no longer be an issue.

plugin/plugins/badger2ds/badger2ds.go
```diff
@@ -80,6 +85,43 @@ func (*badger2dsPlugin) DatastoreConfigParser() fsrepo.ConfigFromMap {
...
+		case "":
+			c.compression = badger2ds.DefaultOptions.Compression
+			c.zstdCompressionLevel = badger2ds.DefaultOptions.ZSTDCompressionLevel
+		default:
+			return nil, fmt.Errorf("unrecognized value for compression: %s", cm)
+		}
+
+		bcs, ok := params["blockCacheSize"]
+		if !ok {
+			// default to 0 (disabled)
```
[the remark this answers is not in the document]
gammazero: Comment has been clarified to state that missing value results in go-ds-badger2 defaults.  This is to tell developers where to look to see what values will be used.  Mostly, this means using default values from badger/v2, but the point is to give tell developers where the next link in this chain is.

docs/datastores.md
```diff
@@ -51,6 +51,26 @@ Uses [badger](https://github.com/dgraph-io/badger) as a key value store.
...
+* `truncate`: Truncate the DB if a partially written sector is found (defaults to true). There is no good reason to set this to false unless you want to manually recover partially written (and unpinned) blocks if go-ipfs crashes half-way through a adding a file.
+* `compression`: Configure compression. When compression is enabled, every block is compressed using the specified algorithm. This option doesn't affect existing tables. Only the newly created tables are compressed. Compression can be configured to use the snappy algorithm or ZSTD with level 1, 2 or 3.  An unspecified value uses the default configuration.
+* `blockCacheSize`: Specifies how much data cache should hold in memory. If compression is disabled, adding a cache leads to unnecessary overhead which may affect read performance. A value of 0 means no block cache, and an unspecified value uses the default configuration.
+
+```json
+{
```
aschmahmann, line 64: It looks like both badger and badger2 documentation do not specify that you can set `vlogFileSize`.  @Stebalien this is just an oversight right?

Stebalien: Yes.

aschmahmann: @gammazero could you add in the `vlogFileSize` documentation as well?

gammazero: Done.

Stebalien (approved):
:rocket:

aschmahmann (approved):
This looks good to go pending a release of go-ds-badger2 and upgrading go-ipfs-config to v0.10.0

aschmahmann:
This is blocked by the badger folks figuring out if there latest breaking changes will be handled within badger v2 or whether there will be a badger v3. The go-ipfs-config PR adding badger 2 support has been reverted until this is resolved.

dokterbob:
@aschmahmann There's now a Badger v3 https://github.com/dgraph-io/badger/releases/tag/v3.2011.1

However, go-ds-badger2 is actively maintained. Requirements mentioned by you have been met a long time ago and, even as an experimental feature, it could gain great benefit and insight to have something *less broken* to test with. This might also yield useful feedback while a move towards badger 3 is made.

We currently have a badgerdb in IPFS which is broken in [obvious ways](https://github.com/ipfs/go-ds-badger/issues/54#issuecomment-819866098) making it unsuitable for production.

dokterbob:
In short; all that's necessary to have a badger v2, based on actively maintained code, in master, is to resolve these merge conflicts. Getting v3 support is probably a fair amount of work and might delay having something working by several months/up to a year. 

It would be great to have this in user's hands!

godcong:
I tried to use V3, but it will panic in some cases.

dokterbob:
@aschmahmann Poke!

dokterbob:
At ipfs-search.com we currently have a [script](https://github.com/ipfs-search/ipfs-search-deployment/blob/main/roles/ipfs-gc/templates/ipfs_gc.sh) running from a cron job which stops IPFS, flattens the datastore using the badger binary and then restarts IPFS.

I know you've got a lot on your hands, but it would be great to see some progress on this one!

dokterbob:
Poke!

feld:
flatfs is not good when you've got 10s to 100s of gigs in IPFS. One of my use cases can survive with a manual flatten as suggested by @dokterbob but it would be nice if we had badger v2 or v3 to work with

feld:
@dokterbob I have a branch with Badger2 working on IPFS v0.14.0 and garbage collection does work properly

https://github.com/feld/kubo/tree/feld/badger2-rebase-v0.14.0

lidel:
Closing this one – iirc long term we will more likely introduce pebble (https://github.com/ipfs/kubo/issues/10347) and short-to-mid term discourage use of badgerv1 (https://github.com/ipfs/kubo/pull/10517)

PR #7708, opened 2020-09-30
move CircleCI sharness tests into a separate container

petar:
Needed in order to establish a network link between sharness and the mock Remote Pinning API service (which also runs in a container in CI).


gammazero:
No longer using CircleCi

Issue #7541, opened 2020-07-15
Docker Container on Rock64 64Bit ARM. Doesn't run unless built from source.


BunkerMelon:
**Version Info**
Golang version: 1.14
IPFS Version: Latest, Docker container
Architecture: aarch64/Cortex-A53
Operating System: Ubuntu 20.05.4

**Description**
Wanted to test running IPFS docker image on my Pine64 SBC. Went over to dockerhub and found the go-ipfs image
Went ahead and pulled this image; "docker pull ipfs/go-ipfs".

This downloaded and extracted fine, but when doing a test run with "**docker run -d --name ipfs_host -v $ipfs_staging:/export -v $ipfs_data:/data/ipfs -p 4001:4001 -p 127.0.0.1:8080:8080 -p 127.0.0.1:5001:5001 ipfs/go-ipfs:latest**" I get the container hash that the container started, but when checking logs "**docker logs -f ipfs_host**" I recieve **exec user process caused "exec format error"** similair to the problem encountered here #7422 

When running docker ps -a I can see the container was exited as soon as it was started.

I was able to solve this problem by cloning the repo and building the image myself (no changes to the Dockerfile) and re-running the docker run command listed above. I have a spare Rock64 where I can reproduce this for testing.

Stebalien:
This probably has something to do with how we build and deploy to dockerhub (https://github.com/ipfs/go-ipfs/blob/480defab689610550ee3d346e31441a2bb881fcb/.circleci/config.yml#L308-L340). We probably need some extra logic to get this to work cross-architecture.

Unfortunately, this is not something the go-ipfs team is likely to have time to tackle in the near future. However, if you have some spare time and want to try to get cross building working on CircleCI, it looks like https://www.docker.com/blog/multi-arch-build-what-about-circleci/ is a good place to start.

BunkerMelon:
Ah okay, that makes sense. Thank you for the confrimation. If I do get some spare time, I will look deeper into the cross building issue. 

Totally understandable that this isn't a top priority, just thought I would bring it to the teams attention if they weren't aware. If I come up with any solutions for this, I will surely return back with a solution.

Stebalien:
Yeah, thanks for bringing this up.

mahoekst:
Also an issue if you want to run ipfs-go in docker on a Raspberry currently. The build from the docker file works flawlessly though. But it would be nice to be able to refer to a single image from the docker-compose file to support amd64 and arm64

[...]

PR #7710, opened 2020-10-02
[WIP] Docker multi-arch arm7/8

mburns:
connected to #7541
# What

Docker has a (still experimental) [`buildx` plugin to make multi-arch docker images](https://github.com/docker/buildx) simply.

# Why

Using `buildx` lets us leverage a single Dockerfile, rather [having one per-architecture](https://github.com/ckulka/docker-multi-arch-example).

Adapted from https://www.docker.com/blog/multi-arch-build-what-about-circleci/


mburns:
>failed to solve: rpc error: code = Unknown desc = failed to load LLB: runtime execution on platform linux/arm64 not supported

hsanjuan:
@mburns this looked interesting, any context on why it was discarded?

mburns:
It is worth reviving, I just didn't want this particular attempt languishing.

PR #7722, opened 2020-10-11
Add missing plugin support on FreeBSD

tzdybal:
Go plugins are supported on FreeBSD, but build tags in `plugin/loader/load_unix.go` enable IPFS plugins only on Linux and Darwin.


gammazero (approved):
LGTM

PR #7730, opened 2020-10-19
fix: do not create a new (unused) peerID when initializing from config

aschmahmann:


petar (approved):
lgtm

straiforos (commented):
LGTM, nice work seems like a logical change, the daemon seems like an odd place for it when init is required.

PR #7735, opened 2020-10-23
return an error when trying to download fs-repo-migrations for linux + musl

aschmahmann:


Stebalien (approved):

Stebalien (commented):
Maybe say: Please run "go run ... migrate command" to build and run the migrations yourself?

aschmahmann:
The function is generic across binaries. The resulting error should inform that the issue was with fs-repo-migrations and that the user must build the binary themselves. As long as they build the binary and stick it in their path everything should be fine.

PR #7743, opened 2020-10-29
Fix | Updating multiaddress net library to be confident on the security for…

straiforos:
… enterprise use cases

This seems to be a minor fix I noticed before I attempt to add secure https only for ipfs to secure the man in the middle attack vector. I plan to use caddy and read the two code bases to the best of my ability. I am a complete noob at golang as I am working to find my code calling as a nondogmatic developer. I plan to be building an IPFS CDN provider and need to ensure this is as close to enterprise ready to the best of my ability. So that all being said I appreciate the help in advance to get this minor references using the right multiaddress net library to use. I am not aware if there is a reason this was left but I took the liberty to modify it so I can feel confident in extending the manet package to support a TLS secured net.Listener.

PR #7744, opened 2020-10-29
Fix | Updating multiaddress net library to be confident on the security for…

straiforos:
… enterprise use cases

Closed https://github.com/ipfs/go-ipfs/pull/7743 to conform more correctly to the IPFS Go guidelines. I modified the library very manually and did not research sum or mod files but I get the gist. This is my research of how I can secure the http endpoint. I plan to create an adapter to configure the SSL more easily so if there are ideas already brainstormed I would appreciate the help!

Otherwise I plan to extend the MA Net library to support TLS wrapped net.Listeners, and I am very new to this so I expect challenges and probably blind spots but willing to put the leg work in to get there to create enterprise ready solutions on this alpha project! Showing its true value to everyone!

Please refer to my own comments on my work in the previous Pull 7743

### DIFF
--- a/p2p/local.go
+++ b/p2p/local.go
@@ -9,7 +9,7 @@ import (
 	"github.com/libp2p/go-libp2p-core/peer"
 	"github.com/libp2p/go-libp2p-core/protocol"
 	ma "github.com/multiformats/go-multiaddr"
-	"github.com/multiformats/go-multiaddr/net"
+	manet "github.com/multiformats/go-multiaddr/net"
 )
 
 // localListener manet streams and proxies them to libp2p services
--- a/test/dependencies/go.mod
+++ b/test/dependencies/go.mod
@@ -6,7 +6,7 @@ require (
 	github.com/Kubuxu/gocovmerge v0.0.0-20161216165753-7ecaa51963cd
 	github.com/golangci/golangci-lint v1.26.0
 	github.com/ipfs/go-blockservice v0.1.3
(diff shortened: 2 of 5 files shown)


straiforos:
> … enterprise use cases
> 
> Closed #7743 to conform more correctly to the IPFS Go guidelines. I modified the library very manually and did not research sum or mod files but I get the gist. This is my research of how I can secure the http endpoint. I plan to create an adapter to configure the SSL more easily so if there are ideas already brainstormed I would appreciate the help!
> 
> Otherwise I plan to extend the MA Net library to support TLS wrapped net.Listeners, and I am very new to this so I expect challenges and probably blind spots but willing to put the leg work in to get there to create enterprise ready solutions on this alpha project! Showing its true value to everyone!
> 
> Please refer to my own comments on my work in the previous Pull 7743

Figured out where the http client lives, I am a noob, it is the go-ipfs-cmds/http, sorry to clutter a PR with a question. Looking forward to feedback on this PR.

aschmahmann:
Thanks, there aren't any security issues here. We just moved the code from multiaddr-net to multiaddr/net for maintainability.

It's a good dep update though. Will get the deps updated before the next release.

straiforos:
> It's a good dep update though. Will get the deps updated before the next release.

Sounds good I will close this for reference. I will ensure that the library is removed with the next release since that will be important for my product goals. Thanks for taking the time!

PR #7753, opened 2020-11-05
fix: ipfs dag export uses the CoreAPI and respects the offline flag

aschmahmann:
`ipfs dag import/export` were directly using the `IpfsNode` instead of a `CoreAPI` which meant that `--offline` doesn't "automatically" work unless we actually do some manual work inside of the code (i.e. `cmdenv.GetNode(env)` doesn't do anything special with the offline flag).

While we could add some magic there to make this problem less common we could also just move more towards the CoreAPI. This PR attempts that latter.

Note: Draft one has only fixed Export, still need to do Import.

cc @ribasushi

### DIFF
--- a/core/commands/dag/dag.go
+++ b/core/commands/dag/dag.go
@@ -550,7 +550,7 @@ The output of blocks happens in strict DAG-traversal, first-seen, order.
 			)
 		}
 
-		node, err := cmdenv.GetNode(env)
+		api, err := cmdenv.GetApi(env, req)
 		if err != nil {
 			return err
 		}
@@ -588,7 +588,7 @@ The output of blocks happens in strict DAG-traversal, first-seen, order.
 				req.Context,
 				mdag.NewSession(
 					req.Context,
-					node.DAG,
+					api.Dag(),
 				),
 				[]cid.Cid{c},
 				pipeW,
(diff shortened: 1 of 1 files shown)


core/commands/dag/dag.go
```diff
@@ -606,9 +606,16 @@ The output of blocks happens in strict DAG-traversal, first-seen, order.
...
 		// minimal user friendliness
 		if err != nil &&
-			!node.IsOnline &&
 			err == ipld.ErrNotFound {
-			err = fmt.Errorf("%s (currently offline, perhaps retry after attaching to the network)", err)
+			explicitOffline, _ := req.Options["offline"].(bool)
+			if explicitOffline {
+				err = fmt.Errorf("%s (currently offline, perhaps retry without the offline flag)", err)
+			} else {
+				node, envErr := cmdenv.GetNode(env)
```
ribasushi, line 614: I *believe* if you got here ( `err == ipld.ErrNotFound` ), IPFS already tried GetNode, so that'd be a redundant second call... but I might be wrong.

Otherwise looks great!

aschmahmann: I swapped out GetNode for GetAPI above. Am I missing something?

ribasushi: Ignore what I said, pre-coffee 🤦 
This is good to go!

P.S. node == ipfs object in this context, I misfired node == block

core/commands/dag/dag.go
```diff
@@ -606,9 +606,16 @@ The output of blocks happens in strict DAG-traversal, first-seen, order.
 
 		// minimal user friendliness
 		if err != nil &&
-			!node.IsOnline &&
 			err == ipld.ErrNotFound {
```
Stebalien, line 609: This seems redundant.

Stebalien (approved):

core/commands/dag/dag.go
aschmahmann: Given that there could be more errors that have nothing to do with being offline (e.g. datastore fetching errors, context cancellations, etc.) and that this was already here doesn't seem like it hurts.

aschmahmann:
Merging since handling import is a little more in depth and we can cover it in a separate PR.

PR #7797, opened 2020-12-01
More p2p proxy checks

aschmahmann:
Added a few checks to the p2p proxy to return errors to the client rather than just logging them on the daemon.

### DIFF
--- a/core/corehttp/p2p_proxy.go
+++ b/core/corehttp/p2p_proxy.go
@@ -65,7 +65,7 @@ func parseRequest(request *http.Request) (*proxyRequest, error) {
 	}
 
 	split = strings.SplitN(path, "/", 7)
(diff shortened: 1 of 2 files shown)


core/corehttp/p2p_proxy.go
```diff
@@ -60,12 +61,16 @@ func parseRequest(request *http.Request) (*proxyRequest, error) {
 		return nil, fmt.Errorf("Invalid request path '%s'", path)
 	}
 
+	if _, err := peer.Decode(split[2]); err != nil {
+		return nil, fmt.Errorf("Invalid request path '%s'", path)
+	}
```
aschmahmann, line 66: This effectively changes the error code for an invalid peerID from 502 to 400 and shows an error to the user instead of just logging on the server.

Stebalien (approved):

Issue #7559, opened 2020-07-23
Add support for remote Pinning Services


jacobheun:
## Summary
The goal of this is to add a Pinning Services client in go-ipfs that can integration with Pinning Services that conform to the [Pinning Services Spec](https://github.com/ipfs/pinning-services-api-spec). Users will be able to add Pinning Services to their IPFS configuration and then leverage those services while pinning.

## Initial Use Case
The desired use case being fulfilled here is to enable IPFS Desktop / WebUI users to configure their local IPFS node to remotely manage pins with pinning services on their behalf.

## Implementation Details
Pinning Services Spec: https://github.com/ipfs/pinning-services-api-spec
Spec Version: v1.0.0 (to be tagged after go-ipfs changes ship in 0.8.0)

## Stages
The implementation may be implemented in a phased approach. This will allow for an iterative rollout of support.

### 1. Remote Pinning Support
This phase will implement the client and add support for manual pinning. This will enable users to specify the remote services for a single cid.

- [x] Add support for `Pinning.RemoteServices` in the config https://github.com/ipfs/go-ipfs-config/pull/104
  - `Pinning.RemoteServices[].Api.Key` **must** be write only.
- [x] Implement the Pinning Services client for the spec version mentioned above. https://github.com/ipfs/go-ipfs/issues/7559
- [x] Update the `pin` API to accept a list of services to pin remotely https://github.com/ipfs/go-ipfs/issues/7559 
  - Should be able to do this in the CLI and over the http API
  - All remote pins are recursive only
  - The pin should also occur locally
  - When Pins are created/modified go-ipfs should connect to the `provider` multiaddrs specified by the Pinning Service to accelerate transfer times.

### 2. MFS Pinning Policy
This phase adds support for pinning policies and introduces the "all in MFS" policy. Policies will be configurable per Pinning Service to allow for more control.

- [x] Add support for watching changes to MFS and updating the pin for MFS root on the configured remote services https://github.com/ipfs/go-ipfs/pull/7798
  - When the MFS root changes, the client should make a modify pin request to the Pinning Service. _Retention policies are dependent on the service and are not yet supported in the api_
- [x] Add support for `Pinning.RemoteServices[].Policy` in the config.
- [x] Add the implicit `manual`(default) and `all-in-mfs` policies.
  - `manual` has the same behavior mentioned in phase 1.
  - `all-in-mfs` will watch for changes in MFS and update the pins for all Services with the policy

## Related/References
[...]

PR #7798, opened 2020-12-01
add remote pinning policy for mfs

### LINKED ISSUE #7818: Error in `config show`:  case mismatch in config
#### Version information:

go-ipfs version: 0.8.0-rc1-02d15ac41
Repo version: 11
System version: amd64/linux
Golang version: go1.15.2

#### Description:
After upgrading/migrating from  `0.7.0-ea77213e3` to `0.8.0-rc1-02d15ac41` I get this error when running "config show":

```
$ ipfs config show
Error: case mismatch in config, expected "Pinning" but got ""
```

My config file doesn't have any `"Pinning"` fields (since I'm using the same config from when I was running 0.7.0.  I didn't manually add any fields, and it appears that IPFS didn't automatically migrate the config)

petar:
> This builds on top of https://github.com/ipfs/go-ipfs/issues/7559

Needed for https://github.com/ipfs/go-ipfs/issues/7559 &  https://github.com/ipfs/ipfs-gui/issues/91
Needs https://github.com/ipfs/go-ipfs-config/pull/116

Closes #7818


cmd/ipfs/daemon.go
```diff
@@ -457,6 +469,99 @@ func daemonFunc(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment
...
+			case errCh <- err:
+			case <-ctx.Done():
+			}
+			return
+		}
+		opts = append(opts, pinclient.PinOpts.WithOrigins(addrs...))
+	}
+
+	// Execute remote pin request
+	// TODO: fix panic when pinning service is down
```
aschmahmann, line 555: Is this still true? I'm pretty sure that's fixed

core/commands/pin/remotepin.go
```diff
@@ -480,8 +480,10 @@ var addRemotePinServiceCmd = &cmds.Command{
 			cfg.Pinning.RemoteServices = map[string]config.RemotePinningService{}
 		}
 		cfg.Pinning.RemoteServices[name] = config.RemotePinningService{
-			ApiEndpoint: url,
-			ApiKey:      key,
+			Api: config.RemotePinningServiceApi{
+				ApiEndpoint: url,
+				ApiKey:      key,
+			},
```
aschmahmann, line 486: The config layout for service.Api is changed in this PR, it should probably be in the other one (and this one rebased)

cmd/ipfs/daemon.go
```diff
@@ -457,6 +469,99 @@ func daemonFunc(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment
...
+	_, err := c.Add(ctx, cid, opts...)
```
aschmahmann, line 556: Ideally we'd use the `Replace` function here, but not everyone supports it yet. WDYT @lidel?

----

If we stick with just adding more and more pins and relying on the user's arrangement with the pinning service to remove old copies of MFS then perhaps we want to poll the service on startup for the latest pin and set `lastRootCid`.

cmd/ipfs/daemon.go
```diff
@@ -457,6 +469,99 @@ func daemonFunc(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment
 	return errs
 }
 
+const mfsRepinInterval = time.Minute
```
aschmahmann, line 472: @petar @lidel is this the kind of thing users are going to want to change from a config file? They can always pin more manually if they're really concerned, but was wondering aloud here. If this could get more complicated we may want further nesting in the config like `Policies.PinMFS.Enabled`, but I don't want to go overboard here for no reason.

aschmahmann (changes requested):

cmd/ipfs/daemon.go
```diff
@@ -457,6 +469,99 @@ func daemonFunc(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment
...
+			// pin on all remote services in parallel to prevent DoS attacks
+			var wg sync.WaitGroup
+			for svcName, svcConfig := range cfg.Pinning.RemoteServices {
+				if svcConfig.Policies.PinMFS == nil || !*svcConfig.Policies.PinMFS {
+					continue
+				}
+				wg.Add(1)
+				go func() {
+					defer wg.Done()
+					pinMFS(cctx.Context(), node, lastRootCid, svcName, svcConfig, errCh)
```
aschmahmann, line 524: `go vet` error here on loop closure which are always fun 😄.

> loopclosure: loop variable svcName captured by func literal (govet)

You can either set `name := svcName` in the for loop and use that in the goroutine, or pass arguments into the goroutine function.

cmd/ipfs/daemon.go
```diff
@@ -457,6 +469,99 @@ func daemonFunc(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment
...
 
+const mfsRepinInterval = time.Minute
+
+func pinMFSOnChange(cctx *oldcmds.Context, node *core.IpfsNode) (<-chan error, error) {
+	errCh := make(chan error)
+	go func() {
+		var lastRootCid cid.Cid
+		for {
+			// poll periodically
+			tmo := time.NewTimer(mfsRepinInterval)
```
gammazero, line 480: minor: Instead of creating a new timer each iteration,
```go
var tmo *time.Timer
for {
    if tmo == nil {
        tmo = time.NewTimer(mfsRepinInterval)
    } else {
        tmo.Reset(mfsRepinInterval)
    }
    ...
```

cmd/ipfs/daemon.go
```diff
@@ -457,6 +469,99 @@ func daemonFunc(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment
...
+			select {
+			case <-cctx.Context().Done():
+				return
```
gammazero, line 483:
```go
tmo.Stop()
return
```

cmd/ipfs/daemon.go
```diff
@@ -457,6 +469,99 @@ func daemonFunc(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment
 	return errs
 }
 
+const mfsRepinInterval = time.Minute
+
+func pinMFSOnChange(cctx *oldcmds.Context, node *core.IpfsNode) (<-chan error, error) {
+	errCh := make(chan error)
+	go func() {
+		var lastRootCid cid.Cid
```
gammazero, line 477: Do you think it might be good to `defer close(errCh)` here, so that a routine waiting on `errCh` can know that all errors were collected, upon seeing the channel closed?

gammazero (commented):
LGTM - with a couple of minor comments

cmd/ipfs/daemon.go
petar: Absolutely. Thanks.

cmd/ipfs/daemon.go
lidel: I'd like to keep it simple, but I also see how some deployments may want to limit the number of remote API calls and eg. embedded device with a battery or limited data plan will want to limit sync to once an hour or once a day, so ability to customize the check interval will be requested anyway.

Could we reuse the convention from [`Datastore.GCPeriod`](https://github.com/ipfs/go-ipfs/blob/master/docs/config.md#datastoregcperiod)?
It uses the `duration` type (an empty string means the default value).



What if we rename it to `MFSRepinCheckInterval` of type `duration` where empty/missing value simply means `0` (off)? 
That way we keep it simple and give all the flexibility users could want from this feature.

cmd/ipfs/daemon.go
lidel: Keeping old copies pinned and shifting "expiration" burden to pinning service sounds like a feature that some use cases will appreciate, and some will want to avoid. 

It sounds like a flag  that we should add (`KeepOldSnapshots: false` by default)

cmd/ipfs/daemon.go
petar: I went with the idea to mimic Datastore.GCPeriod. Funnily, polling of a different type is still necessary: to catch changes in the config. Take a look and see what you think.

lidel (changes requested):

cmd/ipfs/daemon.go
```diff
@@ -463,6 +475,152 @@ func daemonFunc(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment
...
+	node *core.IpfsNode,
+	cid cid.Cid,
+	svcName string,
+	svcConfig config.RemotePinningService,
+	errCh chan<- error,
+) (lastPin, error) {
+	c := pinclient.NewClient(svcConfig.Api.ApiEndpoint, svcConfig.Api.ApiKey)
+
+	// Prepare Pin.name
+	opts := []pinclient.AddOption{pinclient.PinOpts.WithName("policy-mfs")}
```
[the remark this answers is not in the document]
aschmahmann: 1.  Fixed. I missed it (and also forgot to update the go-ipfs-config version) when rebasing on master after merging the other PR :see_no_evil:

2. Not unreasonable to add the PeerID, although when the user does an `ipfs key rotate` the PeerID will change which might be a little annoying for users and also is something we'd need to consider if we poll the pinning services on startup at all.

3. I think having a "special" or "policy" marker in the name could be useful. I'm not normally a fan of the lexicographical race to the top, but this is probably fine.
   - We may want to consider the ordering of the name elements since we don't support regex type queries over the pins. e.g. are people more likely to want `__policy-mfs*` for all their peers or `__policy-peerid*` for all the policies for a given peer (assuming `peerid` is even relevant for other policies). It doesn't necessarily have to be the same ordering for MFS as for any other policy, but also worth a bit of considering about how this format might generalize to other policies.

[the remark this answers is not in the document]
petar: 2. Another option is to let people set the name of the MFS pin in the policy itself. I tend to like this better than the peer-id (which can change).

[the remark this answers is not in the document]
lidel: 2. I would not worry about key rotation – no data is lost, it will rarely happen, and user can unpin old ones manually. 
   But I like Petar's idea: if we have `Policies.MFS` as struct, then `Policies.MFS.PinName` gives an  option to define deterministic pin name (when empty, default generated from peerid is used). 
3. Good point on "scoping" them per peerid. 
   With `__policy/peerid/mfs`, when sorted by name, pins produced by policies will be grouped per peerid (better UX)

test/sharness/t0700-remotepin.sh
```diff
@@ -64,18 +88,9 @@ test_expect_success "output includes meaningful error" '
...
-  test_expect_code 1 grep RemoteServices show_config
-'
-
-test_expect_success "'ipfs config replace' injects remote services back" '
-  test_expect_code 1 grep -q -E "test_.+_svc" show_config &&
-  ipfs config replace show_config &&
-  test_expect_code 0 grep -q test_pin_svc "$IPFS_PATH/config" &&
-  test_expect_code 0 grep -q test_invalid_key_svc "$IPFS_PATH/config" &&
-  test_expect_code 0 grep -q test_invalid_url_path_svc "$IPFS_PATH/config" &&
-  test_expect_code 0 grep -q test_invalid_url_dns_svc "$IPFS_PATH/config"
```
lidel, line 78: We need to have a test for `ipfs config replace` because it is a separate code path and is used by _Settings_ screen of WebUI at  http://127.0.0.1:5001/webui

Right now I am unable to edit config via WebUI – even if  unrelated values are edited (eg. port in `Addresses.Swarm`) I am unable to update config via `ipfs config replace` as it fails due to:

```
cannot show or change pinning services through this API (try: ipfs pin remote service --help)
```

lidel (changes requested):
Found two problems:

1.  Over time, pins duplicates get created.
     To keep things easier to review I proposed a fix in separate PR against this one: https://github.com/ipfs/go-ipfs/pull/7864

2. Broken `config replace`  means broken WebUI 
   (details inline  below)

cmd/ipfs/pinmfs.go
```diff
@@ -0,0 +1,220 @@
+package main
+
+import (
+	"context"
+	_ "expvar"
+	"fmt"
+	_ "net/http/pprof"
```
aschmahmann, line 7: Why are these unused imports imported?

cmd/ipfs/pinmfs.go
```diff
@@ -0,0 +1,220 @@
...
+				// skip services where MFS is not enabled
+				svcName, svcConfig := svcName_, svcConfig_
+				log.Infof("pinning considering service %s for mfs pinning", svcName)
+				if !svcConfig.Policies.MFS.Enable {
+					log.Infof("pinning service %s is not enabled", svcName)
+					ch <- lastPin{}
+					continue
+				}
+				// read mfs pin interval for this service
+				repinInterval, err := time.ParseDuration(svcConfig.Policies.MFS.RepinInterval) //COV
```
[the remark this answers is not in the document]
petar: Sounds good.

cmd/ipfs/pinmfs.go
petar: I think leftover from debugging. Removed.

test/sharness/t0700-remotepin.sh
petar: I took a pass at upgrading 'config replace'. See if it meets your needs.

lidel:
`API.Key` restore  in `config replace` was nearly ok, but I had to refactor it to comply with output of `ipfs config show`
(webui reads config via `config show` and when user edits it, it is saved with `config replace`, so we can't break the cycle)

I cleaned it up a bit, should be easier to follow now + added tests and the sharness should be green (will check on Monday). 

Remaining TODO:
- [ ] confirm the sharness is green
- [ ] add docs
- [ ] test with Pinata again (and decide if we want to cleanup duplicate pins)

core/commands/config.go
```diff
@@ -222,54 +221,49 @@ NOTE: For security reasons, this command will omit your private key and remote s
...
-	if !ok && !okIfMissing {
-		return fmt.Errorf("%s, not found", strings.Join(key, "."))
+func scrub_map(m map[string]interface{}, key []string, okIfMissing bool) (map[string]interface{}, error) {
+	if len(key) == 0 {
+		return nil, nil // delete value
 	}
-
-	delete(cur, todel)
-	return nil
+	n := map[string]interface{}{}
```
aschmahmann, line 252: nit: make(map[string]interface{}) is more idiomatic

core/commands/config.go
```diff
@@ -222,54 +221,49 @@ NOTE: For security reasons, this command will omit your private key and remote s
...
 	}
+	return nil, nil
+}
 
-	todel, _, ok := find(cur, key[len(key)-1])
-	if !ok && !okIfMissing {
-		return fmt.Errorf("%s, not found", strings.Join(key, "."))
+func scrub_map(m map[string]interface{}, key []string, okIfMissing bool) (map[string]interface{}, error) {
+	if len(key) == 0 {
+		return nil, nil // delete value
```
aschmahmann, line 250: This can cause a null reference exception if the caller expects the map to exist (which we do since we return `&cfg`). We may want to add a Go test here for this

core/commands/config.go
```diff
@@ -222,54 +221,49 @@ NOTE: For security reasons, this command will omit your private key and remote s
...
 	}
-
-	delete(cur, todel)
-	return nil
+	n := map[string]interface{}{}
+	for k, v := range m {
+		if key[0] == "*" || strings.EqualFold(strings.ToLower(key[0]), strings.ToLower(k)) {
```
aschmahmann, line 254: Same EqualFold comment/question as above

core/commands/config.go
```diff
@@ -524,33 +518,57 @@ func replaceConfig(r repo.Repo, file io.Reader) error {
...
+	newCfg.Identity.PrivKey = pkstr
+
+	// Handle Pinning.RemoteServices (API.Key of each service is a secret)
 
-	// Handle Pinning.RemoteServices (ApiKey of each service is secret)
-	// Note: these settings are opt-in and may be missing
+	newServices := newCfg.Pinning.RemoteServices
+	oldServices, err := getRemotePinningServices(r)
+	if err != nil {
+		return tryRemoteServiceApiErr
```
aschmahmann, line 528: why this error? It seems like an error relating to the remote pinning service config being broken.

core/commands/config.go
```diff
@@ -524,33 +518,57 @@ func replaceConfig(r repo.Repo, file io.Reader) error {
...
 	}
 
-	// detect if existing config has any remote services defined..
-	if remoteServicesTag, err := getConfig(r, config.RemoteServicesSelector); err == nil {
+	// re-apply API details and confirm every modified service already existed
+	for name, oldSvc := range oldServices {
+		if newSvc, hadSvc := newServices[name]; hadSvc {
+			// fail if input changes any of API details
+			// (interop with config show: allow Endpoint as long it did not change)
+			if len(newSvc.API.Key) != 0 || (len(newSvc.API.Endpoint) != 0 && newSvc.API.Endpoint != oldSvc.API.Endpoint) {
```
aschmahmann, line 541: WDYT about adding more specific errors and/or docs here so people know what they need to remove from their config files before trying to do an `ipfs config replace`?

test/sharness/t0700-remotepin.sh
```diff
@@ -37,19 +38,56 @@ test_expect_success "test 'ipfs pin remote service ls'" '
...
+'
+
+# expect PIN to be created
+test_expect_success "verify MFS root is being pinned" '
+  ipfs files cp /ipfs/bafkqaaa /mfs-pinning-test-$(date +%s.%N) &&
+  ipfs files flush &&
+  sleep 31 &&
+  ipfs files stat / --enc=json | jq -r .Hash > mfs_cid &&
+  ipfs pin remote ls --service=test_pin_mfs_svc --name=mfs_test_pin --status=queued,pinning,pinned,failed --enc=json | tee ls_out | jq -r .Cid > pin_cid &&
+  cat mfs_cid ls_out &&
```
aschmahmann, line 63: intentional or just for debugging? (same below)

test/sharness/t0700-remotepin.sh
```diff
@@ -64,23 +102,24 @@ test_expect_success "output includes meaningful error" '
...
+test_expect_success "'ipfs config show' does not include Pinning.RemoteServices[*].API.Key" '
+  ipfs config show | tee show_config | jq -r .Pinning.RemoteServices > remote_services &&
+  test_expect_code 1 grep \"Key\" remote_services &&
+  test_expect_code 1 grep fake_api_key show_config &&
+  test_expect_code 1 grep "$TEST_PIN_SVC_KEY" show_config
 '
 
-test_expect_success "'ipfs config replace' injects remote services back" '
-  test_expect_code 1 grep -q -E "test_.+_svc" show_config &&
+test_expect_success "'ipfs config replace' injects Pinning.RemoteServices[*].API.Key back" '
```
aschmahmann, line 112: WDYT about another test to make sure that we can change or add a policy in the config and the keys + endpoints stay the same?

cmd/ipfs/pinmfs.go
```diff
@@ -0,0 +1,251 @@
+package main
+
+import (
+	"context"
+	_ "expvar"
```
aschmahmann, line 5: why do we import this?

cmd/ipfs/daemon.go
```diff
@@ -439,6 +439,13 @@ func daemonFunc(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment
 	// initialize metrics collector
 	prometheus.MustRegister(&corehttp.IpfsNodeCollector{Node: node})
 
+	// start MFS pinning thread
+	pinErr := make(chan error)
+	err = pinMFSOnChange(daemonConfigPollInterval, cctx, &ipfsPinMFSNode{node}, pinErr)
+	if err != nil {
+		return err //COV
```
aschmahmann, line 446: was this a TODO?

cmd/ipfs/pinmfs.go
```diff
@@ -0,0 +1,251 @@
...
+	"fmt"
+	"time"
+
+	cid "github.com/ipfs/go-cid"
+	config "github.com/ipfs/go-ipfs-config"
+	"github.com/ipfs/go-ipfs/core"
+	ipld "github.com/ipfs/go-ipld-format"
+	pinclient "github.com/ipfs/go-pinning-service-http-client"
+	"github.com/libp2p/go-libp2p-core/host"
+	peer "github.com/libp2p/go-libp2p-core/peer"
```
aschmahmann, line 15: nit: grouping. let's do `go-libp2p-core`, then `ipfs` then `go-ipfs + go-ipfs-config`)

aschmahmann (changes requested):
Taking another look through the main part of this PR, but added some comments on config + API key scrubbing

core/commands/config.go
```diff
@@ -148,22 +148,21 @@ Set the value of the 'Datastore.Path' key:
...
-	var scope []string
-	for _, name := range roots {
-		scope := append(scope, name)
-		impactedKey := strings.Join(scope, ".")
-		// blockedScope=foo.bar.BLOCKED should return true
-		// for parent and child impactedKeys: foo.bar and foo.bar.BLOCKED.subkey
-		if strings.HasPrefix(impactedKey, blockedScope) || strings.HasPrefix(blockedScope, impactedKey) {
-			return true
+// Returns true if key matches any prefix of glob.
+func matchesGlobPrefix(key string, glob []string) bool {
```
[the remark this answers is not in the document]
petar: Done.

core/commands/config.go
```diff
@@ -148,22 +148,21 @@ Set the value of the 'Datastore.Path' key:
...
+	k := strings.Split(key, ".")
+	for i, g := range glob {
+		if i >= len(k) {
+			break
+		}
+		if g == "*" {
+			continue
+		}
+		if !strings.EqualFold(strings.ToLower(k[i]), strings.ToLower(g)) {
```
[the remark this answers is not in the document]
petar: good call.

cmd/ipfs/pinmfs.go
```diff
@@ -0,0 +1,251 @@
...
+func (x *ipfsPinMFSNode) Identity() peer.ID {
+	return x.node.Identity
+}
+
+func (x *ipfsPinMFSNode) PeerHost() host.Host {
+	return x.node.PeerHost
+}
+
+func pinMFSOnChange(configPollInterval time.Duration, cctx pinMFSContext, node pinMFSNode, errCh chan<- error) error {
+	go func() {
```
aschmahmann, line 56: This goroutine seems to be the only thing we do. Any reason not to have the function be synchronous and then call `go pinMFSOnChange`?

Also, this function seems pretty long/difficult to follow, perhaps we can separate out the timer + state polling logic from the update logic or separate out parts of the update logic. This may also help reduce the number of channels for us to consider at any given point.

aschmahmann (changes requested):

test/sharness/t0700-remotepin.sh
lidel: Intentional, this output is very useful when ` -v` is passed while running sharness to identify cause of regression.

core/commands/config.go
lidel: Yes, this should be `return err` or a wrapped one.

core/commands/config.go
lidel: Tried to go that route, but there are too many variants of things that can go wrong with config of services defined in  `Pinning.*`  that its easier to return `tryRemoteServiceApiErr` for all of them. 

This indicates that the problem is related to pinning service config, suggesting to use of `pin remote service` instead  without going into too much detail hy or what went wrong – I believe its the most useful error for end user.

test/sharness/t0700-remotepin.sh
```diff
@@ -37,19 +38,56 @@ test_expect_success "test 'ipfs pin remote service ls'" '
...
+  test_cmp repin_interval expected_repin_interval &&
+  test_cmp pin_name expected_pin_name &&
+  test_cmp enable expected_enable
+'
+
+# expect PIN to be created
+test_expect_success "verify MFS root is being pinned" '
+  ipfs files cp /ipfs/bafkqaaa /mfs-pinning-test-$(date +%s.%N) &&
+  ipfs files flush &&
+  sleep 31 &&
```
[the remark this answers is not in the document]
lidel: My understanding is that due to the MFS policy timer  loop being hardcoded and executed every 30 seconds we need to wait at least 31 seconds to be sure repinning occured.

core/commands/config.go
petar: I added some useful error text that is more informative.

cmd/ipfs/pinmfs.go
petar: leftover from debugging

cmd/ipfs/daemon.go
petar: yep, cleaned.

cmd/ipfs/pinmfs.go
```diff
@@ -0,0 +1,251 @@
...
+					if r, err := pinMFS(cctx.Context(), node, rootCid, svcName, svcConfig, errCh); err != nil {
+						ch <- lastPin{}
+					} else {
+						ch <- r
+					}
+				}()
+			}
+			for i := 0; i < len(cfg.Pinning.RemoteServices); i++ {
+				x := <-ch
+				lastPins[x.ServiceName] = x
```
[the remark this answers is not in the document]
petar: It was intended, because time=0 informs code that reads it to do the right thing.
But I'll make it more explicit for clarity.

cmd/ipfs/pinmfs.go
```diff
@@ -0,0 +1,251 @@
...
+	lsPinCh, lsErrCh := c.Ls(ctx, pinclient.PinOpts.FilterName(pinName), pinclient.PinOpts.FilterStatus(pinStatuses...))
+	existingRequestID := "" // is there any pre-existing MFS pin with pinName (for any CID)?
+	alreadyPinned := false  // is CID for current MFS already pinned?
+	pinTime := time.Now().UTC()
+	for ps := range lsPinCh {
+		existingRequestID = ps.GetRequestId()
+		if ps.GetPin().GetCid() == cid && ps.GetStatus() != pinclient.StatusFailed {
+			alreadyPinned = true
+			pinTime = ps.GetCreated().UTC()
+			break
```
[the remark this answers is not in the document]
petar: I added a fix.

cmd/ipfs/pinmfs.go
```diff
@@ -0,0 +1,251 @@
...
+		}
+	}
+	if err := <-lsErrCh; err != nil {
+		err = fmt.Errorf("error while listing remote pins: %v", err)
+		select {
+		case errCh <- err:
+		case <-ctx.Done():
+		}
+		return lastPin{}, err
```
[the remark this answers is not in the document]
petar: Yep. Done.

lidel (changes requested):
I took it for a spin and works as expected with our reference pinning service implementation at [rb-pinning-service-api](https://github.com/ipfs-shipyard/rb-pinning-service-api).  :+1: 

Found some minor issues around edge cases (not blockers, we can fix them in separate PRs/RC, but would be nice to include in this one if possible):

### Swallowed errors from pinning service client

I also tested with Pinata and they had a small bug in message format (fix wip), which breaks our client with `remote pinning service returned http error 400: json: cannot unmarshal object into Go struct field FailureError.error.details of type string` error, which makes it retry pin over and over again, which accumulates errors in memory until I exit the process. 

Those pending errors are then returned in bulk when one exits the process :

```
2021-01-28T14:11:16.251+0100	DEBUG	remotepinning/mfs	pinning loop is awake, 1 remote services
2021-01-28T14:11:16.251+0100	DEBUG	remotepinning/mfs	pinning considering service pinata-test for mfs pinning
2021-01-28T14:11:16.251+0100	DEBUG	remotepinning/mfs	pinning MFS root QmebJqoF9hdcmr2XwfHHv6FjQxkWSvjN266E7PPu9aZboP to pinata-test
2021-01-28T14:11:17.489+0100	DEBUG	remotepinning/mfs	pinning to pinata-test: replacing existing MFS root pin with QmebJqoF9hdcmr2XwfHHv6FjQxkWSvjN266E7PPu9aZboP
^C
Received interrupt signal, shutting down...
(Hit ctrl-c again to force-shutdown the daemon.)

Error: 36 errors occurred:
	* remote pinning service returned http error 400: json: cannot unmarshal object into Go struct field FailureError.error.details of type string
	* remote pinning service returned http error 400: json: cannot unmarshal object into Go struct field FailureError.error.details of type string
	* remote pinning service returned http error 400: json: cannot unmarshal object into Go struct field FailureError.error.details of type string
[...]
```

 
Pinata is working on a fix and this should not be a blocker, but I am not sure if silently swallowing this type of errors is a good idea in general.

:point_right: We should log those error responses via `remotepinning/mfs` or `remotepinning/client` logger to save a lot of time in debugging.

### Bogus error on disabling  MFS policy

This is cosmetic, but when I disable MFS policy I get this error (but it gets disabled just fine):

```
$ ipfs config --json Pinning.RemoteServices.rb-pinning-service.Policies.MFS.Enable true
$ ipfs config --json Pinning.RemoteServices.rb-pinning-service.Policies.MFS.Enable false
Error: failed to get config value: "Pinning.RemoteServices.rb-pinning-service.Policies.MFS key has no attributes"
```

petar:
Fixed the swallowing issue.

cmd/ipfs/pinmfs.go
```diff
@@ -0,0 +1,252 @@
...
+				}
+				ch <- lastPin{}
+				continue
+			}
+		}
+
+		mfslog.Debugf("pinning MFS root %s to %s", rootCid, svcName)
+		go func() {
+			if r, err := pinMFS(ctx, node, rootCid, svcName, svcConfig); err != nil {
+				mfslog.Debugf("pinning MFS root %s to %s error (%v)", rootCid, svcName, err)
```
lidel, line 165: Any reason why we not make it `ERROR` instead? Will make it extra clear when service has problems, and avoid people filling issues against go-ipfs when pinning fails due to  remote service  error:

```suggestion
				mfslog.Errorf("pinning MFS root %s to %s error (%v)", rootCid, svcName, err)
```

lidel (changes requested):
- :green_heart:  `config replace` works fine with  http://127.0.0.1:5001/webui/
- :green_heart:   logger `remotepinning/mfs` shows errors correctly now
  ```
  2021-01-28T20:45:07.412+0100	DEBUG	remotepinning/mfs	pinning MFS root QmdLUV2zPmXpZGyb3ASyxyFwP9fHJZom87KF9Lpr3bCVA9 to pinata-test error (remote pinning service returned http error 400: json: cannot unmarshal object into Go struct field FailureError.error.details of type string)
  ```
- :thinking: If there were pinning issues, I still see the accumulated errors  when i interrupt `ipfs daemon`: 
  ``` 
  ^C
  Received interrupt signal, shutting down...
  (Hit ctrl-c again to force-shutdown the daemon.)
  
  Error: 5 errors occurred:
	* remote pinning service returned http error 400: json: cannot unmarshal object into Go struct field FailureError.error.details of type string
	* remote pinning service returned http error 400: json: cannot unmarshal object into Go struct field FailureError.error.details of type string
	* remote pinning service returned http error 400: json: cannot unmarshal object into Go struct field FailureError.error.details of type string
	* remote pinning service returned http error 400: json: cannot unmarshal object into Go struct field FailureError.error.details of type string
	* remote pinning service returned http error 400: json: cannot unmarshal object into Go struct field FailureError.error.details of type string
  ```
  I believe printing errors in realtime is better (see below)

aschmahmann (approved):

lidel (approved):
:ship:  RC2

Issue #7702, opened 2020-09-28
Removing support for X-Ipfs-Gateway-Prefix


lidel:
###  Probably not(?)

I searched [gh projects](https://github.com/search?q=%22X-Ipfs-Gateway-Prefix%22&type=code) and when this header is mentioned it is set to empty string "just in case" (or a repo is a fork of go-ipfs).

Looks like an old, long time dead feature, and that is for a reason: it was introduced in 2016 (https://github.com/ipfs/go-ipfs/pull/1971) and AFAIK the only use for it was displaying blog under `ipfs.io/blog`, which we stopped doing in 2018 (https://github.com/ipfs/blog/pull/182). 

### Why we stopped using it?

Since 2016 we realized nesting multiple gateway types under single Origin on the web is too complex and brittle. 
IPFS community moved towards dedicated Origins, which are much easier to isolate and reason about (we moved `ipfs.io/blog` to subdomain at `blog.ipfs.io`) and the need for `X-Ipfs-Gateway-Prefix` went away.

## Proposal: remove support for `X-Ipfs-Gateway-Prefix` 

Existence of this header makes gateway code harder to maintain, and suggests risky deployment decisions. Right now the header is already ignored in various places, but nobody seems to care to fill a bug report ;-) 

Instead of fixing it I'd like to remove it from `docs/config.md` and gateway code to decrease complexity, avoid origin-related bugs in the future, and make it more clear that one should mount only one DNSLink per Origin.

If someone is still depending on this header, please comment below describing your use case. 
A better way most likely exists.

## "But I AM using it!"

If you are using it, please comment below and describe your use case. 
Most likely there is a better/safer way to serve your use case these days.

jbenet:
**this may be different, but:**
- I've seen problems on sites hosted with dns in some places where directory listings link to `$domain/ipfs/$hash/$path` instead of `$domain/$path`.
- recently i recall running into this with fleek. 

- example: https://slingshot.filecoin.io/media/logos/
- these should link to `https://slingshot.filecoin.io/media/logos/$filename` but instead link to `https://slingshot.filecoin.io/ipfs/QmZ33q4sWmnFXer6UMyAmx3J85ZhwarjXuRnDTzBp1auvG/media/logos/$filename` which ofc doesn't work.
- not sure if this is a misconfigured gateway or what.

lidel:
@jbenet yes, it is different.

[...]

PR #7802, opened 2020-12-05
feat(gw): URI router for Web API navigator.registerProtocolHandler

### LINKED ISSUE #7686: Support Web API navigator.registerProtocolHandler
### What

I propose we add support for `?uri=` which would decode passed URI and return HTTP 301 redirect to proper `/ipfs/` or `/ipns/` path. 

This is pretty trivial to add, but wanted to agree on `?url=` (or other means) before I submit PR.

cc @autonome @gozala @aschmahmann @achingbrain 

### Why

Now that [Chromium 86](https://blog.chromium.org/2020/09/chrome-86-improved-focus-highlighting.html) safelisted `ipfs://`,`ipns://` and `dweb:` we should support requests produced by [`navigator.registerProtocolHandler`](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/registerProtocolHandler) on gateways.

Namely, if  someone registers `dweb.link` as URI handler for `ipfs://`:

```js
navigator.registerProtocolHandler('ipfs', 'https://dweb.link/?uri=%s', 'ipfs resolver')
```

Then opening `ipfs://QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR` Chromium will produce HTTP GET with URI-encoded URI:

```
https://dweb.link/?uri=ipfs%3A%2F%2FQmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR
```

Using URL parameter `?uri=` seems safer than `/uri/%s` because it does not clash with path-based logic already present in gateway code, but I am open for suggestions, if there is a better way to represent this.

dennis-tra:
@lidel This PR attempts to close #7686 

---

This commit adds support for requests produced by navigator.registerProtocolHandler on gateways. Now one can register `dweb.link` as an URI handler for `ipfs://`:

```
navigator.registerProtocolHandler('ipfs', 'https://dweb.link/ipfs/?uri=%s', 'ipfs resolver')
```

Then opening `ipfs://QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR` will produce an HTTP GET call to:

```
https://dweb.link/ipfs?uri=ipfs%3A%2F%2FQmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR
```

The query parameter `uri` will now be parsed and the given content identifier resolved via:

`https://dweb.link/ipfs/QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR`

--- 

Note: 
1. If the path parameter `ipfs` and scheme of the `uri` param value don't match we'll redirect to the `uri` param value. E.g. A call to `https://dweb.link/ipfs/?uri=ipns://content-identifier` will be redirected to `https://dweb.link/ipns/content-identifier`
2. I didn't test path combinations where `X-Ipfs-Gateway-Prefix` is set due to the discussion in #7702.

### DIFF
--- a/core/corehttp/gateway_handler.go
+++ b/core/corehttp/gateway_handler.go
@@ -181,10 +181,17 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
 	}
 	originalUrlPath := prefix + requestURI.Path
 
-	// Query parameter handling to support requests produced by navigator.registerProtocolHandler.
-	// E.g. This code will redirect calls to /ipfs/?uri=ipfs%3A%2F%2Fcontent-identifier
-	// to /ipfs/content-identifier.
-	if uri := r.URL.Query().Get("uri"); uri != "" {
+	// ?uri query param support for requests produced by web browsers
+	// via navigator.registerProtocolHandler Web API
+	// https://developer.mozilla.org/en-US/docs/Web/API/Navigator/registerProtocolHandler
+	// TLDR: redirect /ipfs/?uri=ipfs%3A%2F%2Fcid%3Fquery%3Dval to /ipfs/cid?query=val
+	if uriParam := r.URL.Query().Get("uri"); uriParam != "" {
+		// Browsers will pass URI in URL-escaped form, we need to unescape it first
+		uri, err := url.QueryUnescape(uriParam)
+		if err != nil {
+			webError(w, "failed to unescape uri query parameter", err, http.StatusBadRequest)
+			return
+		}
 		u, err := url.Parse(uri)
 		if err != nil {
 			webError(w, "failed to parse uri query parameter", err, http.StatusBadRequest)
@@ -194,7 +201,11 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
 			webError(w, "uri query parameter scheme must be ipfs or ipns", err, http.StatusBadRequest)
 			return
 		}
-		http.Redirect(w, r, gopath.Join("/", prefix, u.Scheme, u.Host), http.StatusMovedPermanently)
+		path := u.Path
+		if u.RawQuery != "" { // preserve query if present
+			path = path + "?" + u.RawQuery
+		}
+		http.Redirect(w, r, gopath.Join("/", prefix, u.Scheme, u.Host, path), http.StatusMovedPermanently)
 		return
 	}
 
--- a/core/corehttp/gateway_handler.go
+++ b/core/corehttp/gateway_handler.go
@@ -181,6 +181,23 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
 	}
 	originalUrlPath := prefix + requestURI.Path
 
+	// Query parameter handling to support requests produced by navigator.registerProtocolHandler.
+	// E.g. This code will redirect calls to /ipfs/?uri=ipfs%3A%2F%2Fcontent-identifier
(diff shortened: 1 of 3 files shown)


lidel:
@dennis-tra I'm sorry I did not review this yet, been busy with 0.8.0, will review as soon I can.

lidel (approved):
@dennis-tra LGTM, I've added
-  decoding step (because [`navigator.registerProtocolHandler`](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/registerProtocolHandler) will always URL-encode entire URI before passing it to the endpoint
- support for query params to enable things like `ipfs://QmFoo?filename=cat.jpg` 
- tests with longer path that includes unicode characters to confirm `Location` header  is properly url-encoded

@aschmahmann this is kinda small, any chance to review and include it in 0.8.0 RC2?

core/corehttp/gateway_handler.go
```diff
@@ -181,10 +181,17 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
...
-	// E.g. This code will redirect calls to /ipfs/?uri=ipfs%3A%2F%2Fcontent-identifier
-	// to /ipfs/content-identifier.
-	if uri := r.URL.Query().Get("uri"); uri != "" {
+	// ?uri query param support for requests produced by web browsers
+	// via navigator.registerProtocolHandler Web API
+	// https://developer.mozilla.org/en-US/docs/Web/API/Navigator/registerProtocolHandler
+	// TLDR: redirect /ipfs/?uri=ipfs%3A%2F%2Fcid%3Fquery%3Dval to /ipfs/cid?query=val
+	if uriParam := r.URL.Query().Get("uri"); uriParam != "" {
+		// Browsers will pass URI in URL-escaped form, we need to unescape it first
+		uri, err := url.QueryUnescape(uriParam)
```
dennis-tra, line 190: I think `.Query()` will already decode the query parameters. [From the docs](https://golang.org/pkg/net/url/#URL.Query):

> Query parses RawQuery and returns the corresponding values. It silently discards malformed value pairs. To check errors use ParseQuery.

It's not entirely clear if "parsing" also means "decoding" but the distinction to `RawQuery` and the quick check may be evidence enough 🤷‍♂️:

```go
u, _ := url.Parse("http://example.com/?uri=ipfs%3A%2F%2F")
q := u.Query()
fmt.Println(q.Get("uri")) 
// ipfs://
```

dennis-tra:
@lidel thanks for your review and the improvements :) would be great to already see it in `0.8.0` ✌️

core/corehttp/gateway_handler.go
lidel: @dennis-tra good catch! 
I think I was exploring API in a go playground with snippet that did not use `.Query` and did not check if it's truly needed in our case.  Removed in d4ce4b1

dennis-tra:
Hi @lidel,

I just rebased on `master` to trigger the CI again since there was a failing check. Everything looks good now - any chance to get this merged?

lidel:
@dennis-tra thanks for rebase! I've added sharness tests just to be extra sure it won't beg broken by a mistake in the future.

I talked with @aschmahmann and we should be able to include this in 0.8.0-rc2, just need to block some time for the review this month.

I bumped priority for this, because when this ships, we will ask browser vendors with upcoming built-in URI support such as Brave and Opera to switch to this `?uri=` router, removing the complexity from their end.

PR #7819, opened 2020-12-09
docs: Pinning.RemoteServices: API → Api

lidel:
@aschmahmann @petar 

- v0.8.0-rc1 uses `Api`, and this PR changes docs to that
- `Api` introduces tiny inconsistency across the config, because we already use [`Addresses.API`](https://github.com/ipfs/go-ipfs/blob/master/docs/config.md#addressesapi)
  - if you feel its worth changing to `API` then we still want to merge this PR, and then change both code and docs in RC2


petar:
The Go convention is also to use API, so maybe we should change the go side?

lidel:
Yup, closing without merging: https://github.com/ipfs/go-ipfs-config/pull/116 will change the config to use `API`

Issue #7814, opened 2020-12-09
Confusing error message from "ipfs key export"


teknomunk:
#### Version information:
go-ipfs version: 0.8.0-dev-42d3f541d
Repo version: 11
System version: amd64/linux
Golang version: go1.15.6

Installed from a package I built today from https://aur.archlinux.org/packages/go-ipfs-git/

#### Description:
I reproduced the issue reported on the forums at [ERROR with ipfs key export](https://discuss.ipfs.io/t/error-with-ipfs-key-export/9580) on my local machine.

When you run the command "ipfs key export $KEY_NAME" (where $KEY_NAME is not "self") while the ipfs daemon is running, it responds with the message:

Error: lock /home/teknomunk/.ipfs/repo.lock: someone else has the lock

When you stop the ipfs daemon, the command completes without issue.

If the requirement for key export to be an offline command cannot be removed, I think the error message should let the operator know that the command can only be run while the daemon is not.

gammazero:
@teknomunk Thank you for bringing this to our attention.  It probably makes sense to provide the same error that is given when trying to run `ipfs key rotate`:
```
Error: ipfs daemon is running. please stop it to run this command
```

PR #7821, opened 2020-12-11
Improve error message when running key command that locks repo

gammazero:
Fixes Issue #7814

ipfs key commands that open the repo directly will not work when another process has the repo open and locked.  For this reason these commands cannot run when the ipfs daemon is running, and an informative error message should be given to the user.

This is already done for the `ipfs key rotate` command, and should be done for other commands that directly open and lock the repo.  This PR applies the same PreRun check to these commands as is used for `ipfs key rotate`.

While unlikely, it is possible to see the error reported if another process opens the repo at the same moment.  In that case the "someone else has the lock" error is shown.

### DIFF
--- a/core/commands/keystore.go
+++ b/core/commands/keystore.go
@@ -150,6 +150,7 @@ path can be specified with '--output=<path>' or '-o=<path>'.
 		cmds.StringOption(outputOptionName, "o", "The path where the output should be stored."),
 	},
 	NoRemote: true,
+	PreRun:   daemonNotRunning,
 	Run: func(req *cmds.Request, res cmds.ResponseEmitter, env cmds.Environment) error {
 		name := req.Arguments[0]
 
@@ -229,6 +230,7 @@ var keyImportCmd = &cmds.Command{
 		cmds.StringArg("name", true, false, "name to associate with key in keychain"),
 		cmds.FileArg("key", true, false, "key provided by generate or export"),
 	},
+	PreRun: daemonNotRunning,
 	Run: func(req *cmds.Request, res cmds.ResponseEmitter, env cmds.Environment) error {
 		name := req.Arguments[0]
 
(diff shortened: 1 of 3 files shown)


core/commands/keystore.go
```diff
@@ -229,6 +230,7 @@ var keyImportCmd = &cmds.Command{
 		cmds.StringArg("name", true, false, "name to associate with key in keychain"),
 		cmds.FileArg("key", true, false, "key provided by generate or export"),
 	},
+	PreRun: daemonNotRunning,
```
aschmahmann, line 233: Import can be run while the daemon is running (give it a shot, it works). Although it's concerning that breaking this didn't cause our tests to fail. We should probably add online/daemon sharness tests to the keystore tests (i.e. verify import works, export fails, rotate fails).

core/commands/keystore.go
```diff
@@ -556,3 +543,20 @@ func keyOutputListEncoders() cmds.EncoderFunc {
 		return nil
 	})
 }
+
+func daemonNotRunning(req *cmds.Request, env cmds.Environment) error {
```
aschmahmann, line 547: note: this function is also copy-pasted in `init.go`, we may want to deduplicate.

aschmahmann (changes requested):

core/commands/keystore.go
gammazero: Deduplicated by exporting this.  There may be a better file to define this function in, but left it here for now since this is the only place in this package that uses it.

core/commands/keystore.go
gammazero: Fixed

gammazero: Added tests to check import, export, and rotate while daemon running.

aschmahmann (approved):

Issue #7707, opened 2020-09-30
Release v0.8.0


aschmahmann:
> Release Issue Template

# go-ipfs 0.8.0 Release

We're happy to announce go-ipfs 0.8.0. This is planned to be a fairly small release focused on integrating in the new pinning service/remote pinning [API](https://github.com/ipfs/pinning-services-api-spec) that makes the experience of managing pins across pinning services easier and more uniform.

## 🗺 What's left for release

- Remote Pinning
   - [x] go-IPFS integration: #7559, #7661, etc.
      - [x] Modify config to store pinning service information
      - [x] Add commands for manually working with the pinning service API
      - [x] Add basic policy support (e.g. automatically pin/unpin MFS root)
   - [x] HTTP API integration (js: https://github.com/ipfs/js-ipfs/pull/3293)
   - [ ] ~WebUI integration: https://github.com/ipfs-shipyard/ipfs-webui/pull/1615~ Dropped from v0.8.0, although the latest deployed WebUI is always available at https://webui.ipfs.io/#/
- QUIC
   - [x] update quic-go to support QUIC draft-32

# 🚢 Estimated shipping date

TBD

### 🧷 Remote pinning services

There is now support for asking remote services to pin data for you. This means anyone can implement the [spec](https://ipfs.github.io/pinning-services-api-spec/) (developed in this [repo](https://github.com/ipfs/pinning-services-api-spec)) and allow for pin management.

All of the CLI (and corresponding HTTP API) commands are available under `ipfs pin remote`.

This remote pinning service comes with a redesign of how we're thinking about pinning and includes some commonly requested features such as:
- Pins can have names (and coming soon metadata)
- The same content can be pinned multiple times, but of course stored only once
  - This allows applications using the same pinning service to manage their own pins without worrying about removing content important to another application
- Data can be pinned in either the foreground or background

Examples include:
```
ipfs pin remote service add myservice https://myservice.tld:1234/api/path myaccess key

ipfs pin remote add /ipfs/bafymydata --service=myservice --name=myfile 
ipfs pin remote ls --service=myservice --name=myfile
ipfs pin remote ls --service=myservice --cid=bafymydata
ipfs pin remote rm --serivce=myservice --name=myfile
```
A few notes:

Remote pinning services work with recursive pins. This means commands like `ipfs pin remote ls` will not list indirectly pinned CIDs.

[...]

PR #7823, opened 2020-12-11
docs: improved --help for pin remote commands

### LINKED ISSUE #7825: rename `url` in pin remote to endpoint
Per @lidel we use `endpoint` option everywhere in ipfs-client, but then e.g. if user omits endpoint option go-ipfs will produce error with message "argument "url" is required", which can be very confusing for the user.

We should either change naming in go-ipfs as well from url to endpoint everywhere, or change it back to `url` in js-ipfs.

lidel:
> Part of https://github.com/ipfs/go-ipfs/issues/7559 and https://github.com/ipfs/go-ipfs/issues/7707 closes #7825

This PR adds docs in form of a longer description that is displayed when `--help` is passed on the CLI.
Those are pretty important, as CLI crowd often does not read docs online, and prefer to explore features this way. 

While at it, unified some language to match fields in the config file and other places (thx @Gozala!) + included some tips and best practices which should improve onboarding experience.

@jessicaschilling this falls under our effort to improve onboarding: lmk if anything sounds awkward, or we could rephrase to make it even more intuitive.

cc @ipfs/wg-pinning-services 

# `--help` previews

Below are previews of the initial version of this PR:

<details>
  <summary>Click to expand: ipfs pin remote service add --help</summary>

```
USAGE
  ipfs pin remote service add <service> <endpoint> <key> - Add remote pinning service.

SYNOPSIS
  ipfs pin remote service add [--] <service> <endpoint> <key>

ARGUMENTS

  <service>  - Service name.
  <endpoint> - Service endpoint.
  <key>      - Service key.

DESCRIPTION

  Add a credentials for access to a remote pinning service and store them in the
  config under Pinning.RemoteServices map.
  
  TIP:
  
    To add services and test them by fetching pin count stats:
  
    $ ipfs pin remote service add goodsrv https://pin-api.example.com secret-key
    $ ipfs pin remote service add badsrv  https://bad-api.example.com invalid-key
    $ ipfs pin remote service ls --stat
    goodsrv   https://pin-api.example.com 0/0/0/0
    badsrv    https://bad-api.example.com invalid
```
</details>

<details>
  <summary>Click to expand: ipfs pin remote service ls --help</summary>

```
USAGE
  ipfs pin remote service ls - List remote pinning services.

SYNOPSIS
  ipfs pin remote service ls [--stat]

OPTIONS

  --stat  bool - Try to fetch and display current pin count on remote service (queued/pinning/pinned/failed). Default: false.

DESCRIPTION

  List remote pinning services.
  
  By default only a name and an endpoint are listed, however one can pass '--stat'
  to test each endpoint by fetching pin counts for each state:
  
    $ ipfs pin remote service ls --stat
    goodsrv   https://pin-api.example.com 0/0/0/0
    badsrv    https://bad-api.example.com invalid
  
  TIP: pass '--enc=json' for more useful JSON output.
```
</details>

<details>
  <summary>Click to expand: ipfs pin remote add --help</summary>

```
USAGE
  ipfs pin remote add <ipfs-path-or-cid> - Pin object to remote pinning service.

SYNOPSIS
  ipfs pin remote add [--service=<service>] [--name=<name>] [--background] [--] <ipfs-path-or-cid>

ARGUMENTS

  <ipfs-path-or-cid> - Path to object(s) to be pinned.

OPTIONS

  --service     string - Name of the remote pinning service to use (mandatory).
  --name        string - An optional name for the pin.
  --background  bool   - Add to the queue on the remote service and return immediately (does not wait for pinned status). Default: false.

DESCRIPTION

  Asks remote pinning service to pin an IPFS object from a given path or a CID.
  
  To pin CID 'bafkqaaa' to service named 'mysrv' under a pin named 'mypin':
  
    $ ipfs pin remote add --service=mysrv --name=mypin bafkqaaa
  
  The above command will block until remote service returns 'pinned' status,
  which may take time depending on the size and available providers of the pinned
  data.
  
  If you prefer to not wait for pinning confirmation and return immediatelly
  after remote service confirms 'queued' status, add the '--background' flag:
  
    $ ipfs pin remote add --service=mysrv --name=mypin --background bafkqaaa
  
  Status of background pin requests can be inspected with the 'ls' command:
  
    $ ipfs pin remote ls --service=mysrv --cid=bafkqaaa --status=queued,pinning,pinned,failed

```
</details>

<details>
  <summary>Click to expand: ipfs pin remote ls --help</summary>

```
USAGE
  ipfs pin remote ls - List objects pinned to remote pinning service.

SYNOPSIS
  ipfs pin remote ls [--service=<service>] [--name=<name>] [--cid=<cid>]... [--status=<status>]...

OPTIONS

  --service  string - Name of the remote pinning service to use (mandatory).
  --name     string - Return pins with names that contain provided value (case-sensitive, exact match).
  --cid      array  - Return pins for the specified CIDs (comma separated).
  --status   array  - Return pins with the specified statuses (queued,pinning,pinned,failed). Default: [pinned].

DESCRIPTION

  Returns a list of objects that are pinned to a remote pinning service.
  
  NOTE: by default it will only show matching objects in 'pinned' state.
  Pass '--status=queued,pinning,pinned,failed' to list pins in all states
```
</details>

<details>
  <summary>Click to expand: ipfs pin remote rm --help</summary>

```
USAGE
  ipfs pin remote rm - Remove pins from remote pinning service.

SYNOPSIS
  ipfs pin remote rm [--service=<service>] [--name=<name>] [--cid=<cid>]... [--status=<status>]... [--force]

OPTIONS

  --service  string - Name of the remote pinning service to use (mandatory).
  --name     string - Remove pins with names that contain provided value (case-sensitive, exact match).
  --cid      array  - Remove pins for the specified CIDs.
  --status   array  - Remove pins with the specified statuses (queued,pinning,pinned,failed). Default: [pinned].
  --force    bool   - Allow removal of multiple pins matching the query without additional confirmation. Default: false.

DESCRIPTION

  Removes remote pins allowing them to be garbage collected if needed.
  
  This command accepts the same search query parameters as 'ls' and it is a good
  practice to execute 'ls' before 'rm' to confirm the list of pins to be removed.
  
  To remove a single pin for a specific CID:
  
    $ ipfs pin remote ls --service=mysrv --cid=bafkqaaa
    $ ipfs pin remote rm --service=mysrv --cid=bafkqaaa
  
  When more than one pin is matching the query on the remote service an error is
  returned.  To confirm removal of multiple pins pass '--force':
  
    $ ipfs pin remote ls --service=mysrv --name=popular-name
    $ ipfs pin remote rm --service=mysrv --name=popular-name --force
  
  NOTE: When no '--status' is passed, implicit '--status=pinned' is used.
  To list and then remove all pending pin requests pass an explicit status list:
  
    $ ipfs pin remote ls --service=mysrv --status=queued,pinning,failed
    $ ipfs pin remote rm --service=mysrv --status=queued,pinning,failed --force

```
</details>

### DIFF
--- a/core/commands/pin/remotepin.go
+++ b/core/commands/pin/remotepin.go
@@ -56,7 +56,7 @@ const pinNameOptionName = "name"
 const pinCIDsOptionName = "cid"
 const pinStatusOptionName = "status"
 const pinServiceNameOptionName = "service"
-const pinServiceURLOptionName = "url"
+const pinServiceEndpointOptionName = "endpoint"
 const pinServiceKeyOptionName = "key"
 const pinServiceStatOptionName = "stat"
 const pinBackgroundOptionName = "background"
@@ -89,20 +89,41 @@ func printRemotePinDetails(w io.Writer, out *RemotePinOutput) {
 
 // remote pin commands
 
-var pinServiceNameOption = cmds.StringOption(pinServiceNameOptionName, "Name of the remote pinning service to use.")
+var pinServiceNameOption = cmds.StringOption(pinServiceNameOptionName, "Name of the remote pinning service to use (mandatory).")
 
 var addRemotePinCmd = &cmds.Command{
 	Helptext: cmds.HelpText{
 		Tagline:          "Pin object to remote pinning service.",
-		ShortDescription: "Stores an IPFS object from a given path to a remote pinning service.",
+		ShortDescription: "Asks remote pinning service to pin an IPFS object from a given path.",
+		LongDescription: `
+Asks remote pinning service to pin an IPFS object from a given path or a CID.
+
+To pin CID 'bafkqaaa' to service named 'mysrv' under a pin named 'mypin':
+
+  $ ipfs pin remote add --service=mysrv --name=mypin bafkqaaa
+
+The above command will block until remote service returns 'pinned' status,
+which may take time depending on the size and available providers of the pinned
+data.
+
+If you prefer to not wait for pinning confirmation and return immediatelly
+after remote service confirms 'queued' status, add the '--background' flag:
+
+  $ ipfs pin remote add --service=mysrv --name=mypin --background bafkqaaa
+
+Status of background pin requests can be inspected with the 'ls' command:
+
+  $ ipfs pin remote ls --service=mysrv --cid=bafkqaaa --status=queued,pinning,pinned,failed
+
+`,
 	},
 
 	Arguments: []cmds.Argument{
-		cmds.StringArg("ipfs-path", true, false, "Path to object(s) to be pinned."),
+		cmds.StringArg("ipfs-path-or-cid", true, false, "Path to object(s) to be pinned."),
 	},
 	Options: []cmds.Option{
-		cmds.StringOption(pinNameOptionName, "An optional name for the pin."),
 		pinServiceNameOption,
+		cmds.StringOption(pinNameOptionName, "An optional name for the pin."),
 		cmds.BoolOption(pinBackgroundOptionName, "Add to the queue on the remote service and return immediately (does not wait for pinned status).").WithDefault(false),
 	},
 	Type: RemotePinOutput{},
@@ -218,15 +239,18 @@ Returns a list of objects that are pinned to a remote pinning service.
 `,
 		LongDescription: `
 Returns a list of objects that are pinned to a remote pinning service.
+
+NOTE: by default it will only show matching objects in 'pinned' state.
+Pass '--status=queued,pinning,pinned,failed' to list pins in all states.
 `,
 	},
 
 	Arguments: []cmds.Argument{},
 	Options: []cmds.Option{
-		cmds.StringOption(pinNameOptionName, "Return pins objects with names that contain provided value (case-sensitive, exact match)."),
-		cmds.DelimitedStringsOption(",", pinCIDsOptionName, "Return only pin objects for the specified CID(s); optional, comma separated."),
-		cmds.DelimitedStringsOption(",", pinStatusOptionName, "Return only pin objects with the specified statuses (queued,pinning,pinned,failed)").WithDefault([]string{"pinned"}),
 		pinServiceNameOption,
+		cmds.StringOption(pinNameOptionName, "Return pins with names that contain provided value (case-sensitive, exact match)."),
+		cmds.DelimitedStringsOption(",", pinCIDsOptionName, "Return pins for the specified CIDs (comma separated)."),
+		cmds.DelimitedStringsOption(",", pinStatusOptionName, "Return pins with the specified statuses (queued,pinning,pinned,failed).").WithDefault([]string{"pinned"}),
 	},
 	Run: func(req *cmds.Request, res cmds.ResponseEmitter, env cmds.Environment) error {
 		ctx, cancel := context.WithCancel(req.Context)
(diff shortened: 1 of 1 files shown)


core/commands/pin/remotepin.go
```diff
@@ -308,20 +332,41 @@ func flattenCommaList(list []string) []string {
...
 var rmRemotePinCmd = &cmds.Command{
 	Helptext: cmds.HelpText{
-		Tagline: "Remove pinned objects from remote pinning service.",
-		ShortDescription: `
-Removes the pin from the given object allowing it to be garbage
-collected if needed.
+		Tagline:          "Remove pins from remote pinning service.",
+		ShortDescription: "Removes the remote pin allowing it to be garbage collected if needed.",
+		LongDescription: `
+Removes remote pins allowing them to be garbage collected if needed.
```
jessicaschilling, line 338:
```suggestion
Removes remote pins, allowing them to be garbage-collected if needed.
```

core/commands/pin/remotepin.go
```diff
@@ -308,20 +332,41 @@ func flattenCommaList(list []string) []string {
...
+
+This command accepts the same search query parameters as 'ls' and it is a good
```
jessicaschilling, line 340:
```suggestion
This command accepts the same search query parameters as 'ls', and it is good
```

jessicaschilling (commented):
Some small readability nits but looks great. Thank you!

aschmahmann:
Thanks 😃, this is great. Appreciate the comments from @jessicaschilling. Let me know when you're ready to merge.

aschmahmann (approved):

lidel:
@aschmahmann applied improvements from @jessicaschilling, should be ready for merge :+1:

core/commands/pin/remotepin.go
```diff
@@ -89,20 +89,41 @@ func printRemotePinDetails(w io.Writer, out *RemotePinOutput) {
...
+Status of background pin requests can be inspected with the 'ls' command:
+
+  $ ipfs pin remote ls --service=mysrv --cid=bafkqaaa --status=queued,pinning,pinned,failed
+
+`,
 	},
 
 	Arguments: []cmds.Argument{
-		cmds.StringArg("ipfs-path", true, false, "Path to object(s) to be pinned."),
+		cmds.StringArg("ipfs-path-or-cid", true, false, "Path to object(s) to be pinned."),
```
aschmahmann, line 122: I think leaving this at `ipfs-path` is actually less confusing since this is what we call it in other functions (e.g. `ipfs get`). It also corresponds to the paths in https://github.com/ipfs/go-path.

aschmahmann, line 122: Note: If someone passes in an IPNS path it'll get resolved to IPFS before being pinned. IIRC this is how everything else that uses "ipfs-path" works.

In the future we (+ pinning services) may want to support pinning IPNS names, but we can likely tackle that when we get there.

core/commands/pin/remotepin.go
```diff
@@ -89,20 +89,41 @@ func printRemotePinDetails(w io.Writer, out *RemotePinOutput) {
...
+data.
+
+If you prefer to not wait for pinning confirmation and return immediately
+after remote service confirms 'queued' status, add the '--background' flag:
+
+  $ ipfs pin remote add --service=mysrv --name=mypin --background bafkqaaa
+
+Status of background pin requests can be inspected with the 'ls' command:
+
+  $ ipfs pin remote ls --service=mysrv --cid=bafkqaaa --status=queued,pinning,pinned,failed
```
aschmahmann, line 116: We support `--status=queued --status=pinning ...` via the `StringsOption` option.

We currently have some code that allow each `--status=ABC` that has commas in it to be flattened together, IMO this behavior is a little confusing and if we keep both then we should definitely document it.

@lidel why are we supporting the comma separated version is it for user ergonomics, compatibility with clients that refuse to send repeated HTTP query parameters or both?

If it's just for user ergonomics we may want to not tie ourselves to processing this server side and add this into the CLI processing for go-ipfs-cmds (e.g. https://github.com/ipfs/go-ipfs-cmds/blob/edc78ef36c1dadd4d1a87737e9a1727e59d36c45/cli/parse.go#L101).

core/commands/pin/remotepin.go
```diff
@@ -366,11 +411,26 @@ collected if needed.
...
+  goodsrv   https://pin-api.example.com 0/0/0/0
+  badsrv    https://bad-api.example.com invalid
+
+`,
 	},
 	Arguments: []cmds.Argument{
 		cmds.StringArg(pinServiceNameOptionName, true, false, "Service name."),
-		cmds.StringArg(pinServiceURLOptionName, true, false, "Service URL."),
+		cmds.StringArg(pinServiceEndpointOptionName, true, false, "Service endpoint."),
 		cmds.StringArg(pinServiceKeyOptionName, true, false, "Service key."),
```
aschmahmann, line 434: These are arguments not options, so we should name the variables `ArgumentName` instead of `OptionName`

core/commands/pin/remotepin.go
```diff
@@ -427,7 +487,7 @@ var rmRemotePinServiceCmd = &cmds.Command{
 		ShortDescription: "Remove credentials for access to a remote pinning service.",
 	},
 	Arguments: []cmds.Argument{
-		cmds.StringArg("remote-pin-service", true, false, "Name of remote pinning service to remove."),
+		cmds.StringArg(pinServiceNameOptionName, true, false, "Name of remote pinning service to remove."),
```
aschmahmann, line 490: Argument name

aschmahmann (changes requested):

core/commands/pin/remotepin.go
lidel: Done: 4e2d17fd7fdca179f167812fd344299e27c3491d

core/commands/pin/remotepin.go
lidel: Done: 4e2d17fd7fdca179f167812fd344299e27c3491d

core/commands/pin/remotepin.go
lidel: @aschmahmann 
It is just for user ergonomics.  I was playing with CLI and the alternative is pretty tedious.
I am considering adding `--status=any` (https://github.com/ipfs/go-pinning-service-http-client/issues/7) to improve it even further.

My concern with adding it as a global feature of go-ipfs-cmds is that it will block use of `,` in values, it becomes a value separator. Here, its fine, as `,` won't be part of `status` or `cid`, but we don't know if its equally safe for other places: what if I want to pass multiple unix file paths, and one of them has `,` in file or dir name?

aschmahmann: What about if we allowed passing an option into StringsOption with a delimiter. If the delimiter is defined (e.g.  a comma, colon, etc) the it would be used, but otherwise you'd have to pass multiple options instead of using a delimiter

lidel: Yeah, that would be :ok_hand: – removes risk I described.

aschmahmann: Cool, I'll poke around go-ipfs-cmds and see how annoying this is to implement. Thinking it should be a little painful but not too bad.

Note: we'll need docs describing this behavior and that the delimiter only applies to the CLI since the HTTP API docs are generated from these comments too 😬

aschmahmann: This has been implemented in go-ipfs-cmds and the dep bubbled into go-ipfs, this PR can be rebased on master to take advantage of it. #7863

lidel:
@aschmahmann rebased and looks green, mind  taking a final look?

aschmahmann (approved):
Looks great, thank you @lidel and @jessicaschilling. All our examples here use the comma delimited syntax without specific callouts that:
1) It is valid to do repeated option passing (i.e. `ipfs pin remote ls --service=mysrv --status=queued --status=pinned`
2) It is valid for the repeated options to be delimited (i.e. `ipfs pin remote ls --service=mysrv --status=queued,pinned --status=pinning,failed`)
3) The HTTP API doesn't work with any of this delimited business

We may want to do these in subsequent PRs (and perhaps just leave 3 to https://github.com/ipfs/http-api-docs/pull/40).

For now I'm going to merge this though since it's much better than what we currently have.

PR #7828, opened 2020-12-14
fix: normalize remote service endpoint

### LINKED ISSUE #7826: remote pinning service registered with trailing `/` are not handled propely
It appears that that trailing `/` in the registered service endpoint are not handled properly, as in requests to the endpoint end up double slashed.

Here is the illustration of this problem

```
ipfs pin remote service ls --stat
pinbot     http://127.0.0.1:1139/ invalid
pinbot-fix http://127.0.0.1:1139  0/0/0/0
ridicule http://127.0.0.1:1139/ridicule/ invalid
```
Which causes following requests to appear on the endpoint:

```
//pins?limit=1&status=pinned
//pins?limit=1&status=pinning
//pins?limit=1&status=queued
//pins?limit=1&status=failed
/pins?limit=1&status=pinned
/pins?limit=1&status=pinning
/pins?limit=1&status=queued
/pins?limit=1&status=failed
/ridicule//pins?limit=1&status=pinned
/ridicule//pins?limit=1&status=pinning
/ridicule//pins?limit=1&status=queued
/ridicule//pins?limit=1&status=failed
```

I would expect to receive following instead:

```
/pins?limit=1&status=pinned
/pins?limit=1&status=pinning
/pins?limit=1&status=queued
/pins?limit=1&status=failed
```

I think go implementation should do a better URL resolution by taking into account trailing slashes.

lidel:
> Closes #7826 reported by @gozala, needed for 0.8.0 (#7707)

This PR adds URI normalization for HTTP-based remote pinning service endpoint and applies it:
1. before endpoint is stored in  `$IPFS_PATH/config` (when user adds via `pin remote service add`)
2. after endpoint is read from config (in case someone entered it up manually)

ps. alternative approach is to address this upstream in [go-pinning-service-http-client](https://github.com/ipfs/go-pinning-service-http-client), but by doing it in `pin remote` command code we ensure invalid/malformed URIs are not saved to the config file, which is a better UX.

### DIFF
--- a/core/commands/pin/remotepin.go
+++ b/core/commands/pin/remotepin.go
@@ -740,7 +740,7 @@ func getRemotePinServiceInfo(env cmds.Environment, name string) (endpoint, key s
 
 func normalizeEndpoint(endpoint string) (string, error) {
 	uri, err := neturl.ParseRequestURI(endpoint)
-	if err != nil || !strings.HasPrefix(uri.Scheme, "http") {
+	if err != nil || !(uri.Scheme == "http" || uri.Scheme == "https") {
 		return "", fmt.Errorf("service endpoint must be a valid HTTP URL")
 	}
 
--- a/core/commands/pin/remotepin.go
+++ b/core/commands/pin/remotepin.go
@@ -752,7 +752,7 @@ func normalizeEndpoint(endpoint string) (string, error) {
 	uri.Path = strings.TrimSuffix(uri.Path, "/")
 
 	// remove any query params
-	if uri.RawQuery != "" || uri.RawFragment != "" {
+	if uri.RawQuery != "" {
 		return "", fmt.Errorf("service endpoint should be provided without any query parameters")
 	}
 
--- a/core/commands/pin/remotepin.go
+++ b/core/commands/pin/remotepin.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	neturl "net/url"
+	gopath "path"
 
 	"golang.org/x/sync/errgroup"
 
@@ -744,10 +745,20 @@ func normalizeEndpoint(endpoint string) (string, error) {
 	if err != nil || !strings.HasPrefix(uri.Scheme, "http") {
 		return "", fmt.Errorf("service endpoint must be a valid HTTP URL")
 	}
-	// avoid //pins (https://github.com/ipfs/go-ipfs/issues/7826)
+
+	// cleanup trailing and duplicate slashes (https://github.com/ipfs/go-ipfs/issues/7826)
+	uri.Path = gopath.Clean(uri.Path)
+	uri.Path = strings.TrimSuffix(uri.Path, ".")
 	uri.Path = strings.TrimSuffix(uri.Path, "/")
-	// avoid /pins/pins
-	uri.Path = strings.TrimSuffix(uri.Path, "/pins")
+
+	// remove any query params
+	if uri.RawQuery != "" || uri.RawFragment != "" {
+		return "", fmt.Errorf("service endpoint should be provided without any query parameters")
+	}
+
+	if strings.HasSuffix(uri.Path, "/pins") {
+		return "", fmt.Errorf("service endpoint should be provided without the /pins suffix")
+	}
 
 	return uri.String(), nil
 }
--- a/core/commands/pin/remotepin.go
+++ b/core/commands/pin/remotepin.go
@@ -444,13 +444,11 @@ TIP:
 		}
 
 		name := req.Arguments[0]
-		url := req.Arguments[1]
-		key := req.Arguments[2]
-
-		endpoint, err := normalizeEndpoint(url)
+		endpoint, err := normalizeEndpoint(req.Arguments[1])
 		if err != nil {
 			return err
 		}
+		key := req.Arguments[2]
 
 		cfg, err := repo.Config()
 		if err != nil {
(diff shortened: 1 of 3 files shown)


core/commands/pin/remotepin.go
```diff
@@ -669,3 +673,16 @@ func getRemotePinServiceInfo(env cmds.Environment, name string) (url, key string
...
+
+func normalizeEndpoint(endpoint string) (string, error) {
+	uri, err := neturl.ParseRequestURI(endpoint)
+	if err != nil || !strings.HasPrefix(uri.Scheme, "http") {
+		return "", fmt.Errorf("service endpoint must be a valid HTTP URL")
+	}
+	// avoid //pins (https://github.com/ipfs/go-ipfs/issues/7826)
+	uri.Path = strings.TrimSuffix(uri.Path, "/")
+	// avoid /pins/pins
+	uri.Path = strings.TrimSuffix(uri.Path, "/pins")
```
aschmahmann, line 685: This could be problematic. In theory this could be the URL the pinning service uses, although that seems like it'd be a terrible idea.

Perhaps we could return an error here like "did you mean X instead of Y"?

core/commands/pin/remotepin.go
```diff
@@ -643,7 +643,11 @@ func getRemotePinService(env cmds.Environment, name string) (*pinclient.Client,
...
 	}
-	return pinclient.NewClient(url, key), nil
+	endpoint, err := normalizeEndpoint(url)
+	if err != nil {
+		return nil, err
+	}
+	return pinclient.NewClient(endpoint, key), nil
 }
 
 func getRemotePinServiceInfo(env cmds.Environment, name string) (url, key string, err error) {
```
aschmahmann, line 653: whether in this PR or the docs one we should rename url to endpoint in this file, we missed a couple

core/commands/pin/remotepin.go
```diff
@@ -669,3 +673,16 @@ func getRemotePinServiceInfo(env cmds.Environment, name string) (url, key string
 	}
 	return service.Api.Endpoint, service.Api.Key, nil
 }
+
+func normalizeEndpoint(endpoint string) (string, error) {
+	uri, err := neturl.ParseRequestURI(endpoint)
+	if err != nil || !strings.HasPrefix(uri.Scheme, "http") {
```
aschmahmann, line 679: Should we also error if there are query parameters or URL fragments?

core/commands/pin/remotepin.go
```diff
@@ -669,3 +673,16 @@ func getRemotePinServiceInfo(env cmds.Environment, name string) (url, key string
...
+		return "", fmt.Errorf("service endpoint must be a valid HTTP URL")
+	}
+	// avoid //pins (https://github.com/ipfs/go-ipfs/issues/7826)
+	uri.Path = strings.TrimSuffix(uri.Path, "/")
```
aschmahmann, line 683: https://github.com/ipfs/go-ipfs/issues/7826 seems to be asking for path normalization. IIUC this could include any number of repeated `/`s in the uri and I'm not sure if there are any other normalization considerations.

Perhaps Go's `path` package's Clean function could be of use here, or maybe there's something more specific to URIs.

core/commands/pin/remotepin.go
lidel: Added error in 2ffdce5

core/commands/pin/remotepin.go
lidel: Will do in docs one.

core/commands/pin/remotepin.go
lidel: Added in 2ffdce500d5a788a19fbff1e1d4c8c5220bf3df2

core/commands/pin/remotepin.go
lidel: I went with `path.Clean` and added more tests:  2ffdce500d5a788a19fbff1e1d4c8c5220bf3df2

core/commands/pin/remotepin.go
lidel: Actually, I decided to do the remaining part here, to minimize the number of conflicts: e4bc5510649e1f53ace2e0407957ab7c5c35b04c

core/commands/pin/remotepin.go
```diff
@@ -667,5 +666,32 @@ func getRemotePinServiceInfo(env cmds.Environment, name string) (url, key string
...
-	return service.Api.Endpoint, service.Api.Key, nil
+	endpoint, err = normalizeEndpoint(service.Api.Endpoint)
+	if err != nil {
+		return "", "", err
+	}
+	return endpoint, service.Api.Key, nil
+}
+
+func normalizeEndpoint(endpoint string) (string, error) {
+	uri, err := neturl.ParseRequestURI(endpoint)
```
aschmahmann, line 677: A couple questions:
1. Is it safe/expected for us to resolve `.` and `..` as gopath.Clean does, or is that really the job the HTTP server?
2. Should we use `ParseRequest` and error on fragments? Otherwise we end up escaping the `#` in the URI

lidel: 1. It is safe, could be done on either end. Doing it on our end makes things easier to debug, because normalized URL is stored in plain text config file.
2. Fragments are really not expected, won't happen "by mistake", so does not seem worth additional code: a remote service will simply return Error 404 for a missing file.

core/commands/pin/remotepin.go
```diff
@@ -667,5 +666,32 @@ func getRemotePinServiceInfo(env cmds.Environment, name string) (url, key string
...
+	if err != nil || !strings.HasPrefix(uri.Scheme, "http") {
```
aschmahmann, line 678: Maybe let's be more precise here and use "http" or "https" and not "httpxyz"

lidel (commented):
lgtm, ship it

aschmahmann (approved):

PR #7831, opened 2020-12-18
Escape non-printable characters in user output

gammazero:
Replaces control characters and non-printable characters with escape sequences, in any fields that are printed by the CLI, which could have been user input.

Output from `ipfs cat` is unchanged.

### DIFF
--- a/core/commands/cmdenv/env.go
+++ b/core/commands/cmdenv/env.go
@@ -72,10 +72,11 @@ func GetConfigRoot(env cmds.Environment) (string, error) {
 	return ctx.ConfigRoot, nil
 }
 
-// EscNonPrint converts non-printable characters and backslash into Go
-// escape sequences, if the given string contains any.
+// EscNonPrint converts non-printable characters and backslash into Go escape
+// sequences.  This is done to display all characters in a string, including
+// those that would otherwise not be displayed or have an undesirable effect on
+// the display.
 func EscNonPrint(s string) string {
-	// First see if escaping is needed, to avoid creating garbage.
 	if !needEscape(s) {
 		return s
 	}
--- a/core/commands/cmdenv/env.go
+++ b/core/commands/cmdenv/env.go
@@ -72,13 +72,27 @@ func GetConfigRoot(env cmds.Environment) (string, error) {
 	return ctx.ConfigRoot, nil
 }
 
-// EscNonPrint converts control characters and non-printable characters into Go
+// EscNonPrint converts non-printable characters and backslash into Go
 // escape sequences, if the given string contains any.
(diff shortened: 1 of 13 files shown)


core/commands/cmdenv/env.go
```diff
@@ -70,3 +71,28 @@ func GetConfigRoot(env cmds.Environment) (string, error) {
 
 	return ctx.ConfigRoot, nil
 }
+
+// EscNonPrint converts non-printable characters and backslash into Go
+// escape sequences, if the given string contains any.
```
aschmahmann, line 76: WDYT about adding a comment that we're trying to mimic the escape quoting style from https://www.gnu.org/software/coreutils/manual/html_node/Formatting-the-file-names.html?

core/commands/cmdenv/env.go
```diff
@@ -70,3 +71,28 @@ func GetConfigRoot(env cmds.Environment) (string, error) {
...
+func EscNonPrint(s string) string {
+	// First see if escaping is needed, to avoid creating garbage.
```
aschmahmann, line 78: you mean garbage in the sense of memory that needs garbage collecting, right? Maybe just clarify that we're avoiding creating extraneous strings.

aschmahmann (approved):
LGTM, left a few small comments it'd be nice to fix up before merging.

core/commands/cmdenv/env.go
gammazero: The link says essentially what the comment here says, so I don't think it adds much, other than it hints that we are doing this for the same reason.  Also, I fear that people will misread that and think it means identical content.

Added brief description of purpose.

core/commands/cmdenv/env.go
gammazero: IMO, best to simply remove the comment as code's intent is clear.

core/commands/cmdenv/env.go
```diff
@@ -70,3 +71,28 @@ func GetConfigRoot(env cmds.Environment) (string, error) {
...
+	}
+
+	esc := strconv.Quote(s)
+	// Remove first and last quote, and unescape quotes.
+	return strings.ReplaceAll(esc[1:len(esc)-1], `\"`, `"`)
+}
+
+func needEscape(s string) bool {
+	if strings.ContainsRune(s, '\\') {
+		return true
```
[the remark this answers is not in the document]
gammazero: Added a simple unit test.

core/commands/cmdenv/env.go
aschmahmann: Fair enough, if/when we add other escaping options we can add more descriptions

PR #7864, opened 2021-01-15
fix: replace pre-existing pins for MFS root

lidel:
This is a refactor of #7798 that replaces the existing MFS root pin (if present) instead of creating a new pin.

- find by name all pre-existing pins across all Statuses (not just 'pinned')
- replace pre-existing pin (or create a new one if none found)

Without this we would be adding duplicate pins when pinning service
is still processing previous pin request (queued, pinning):

```
QmV6QnQNuyJFq88fSVVY9u5fxftQEKu6unsD62QpLQSefP  queued  policy/12D3KooWEcsoib3zFds4USH9bASPWBvJHRvfHYbxjR8QCHQ7y1vj/mfs
QmbnBveHJSqcqmVK8UE64riotVzMGirCKYE1SQLXcG25dT  queued  policy/12D3KooWEcsoib3zFds4USH9bASPWBvJHRvfHYbxjR8QCHQ7y1vj/mfs
Qmf1FHNVp74F4TGtNg1myXZgaV4ApwmYgC8iXcUX8cidDy  queued  policy/12D3KooWEcsoib3zFds4USH9bASPWBvJHRvfHYbxjR8QCHQ7y1vj/mfs
QmXnVAeJ9braE7ho7hdPqkmd9cB2ewKtzXNcFFFm3W32EA  queued  policy/12D3KooWEcsoib3zFds4USH9bASPWBvJHRvfHYbxjR8QCHQ7y1vj/mfs
QmeSUtG78SnqoCLDnZw5kYx2w9QnCp8JhkKHxK537y2JET  queued  policy/12D3KooWEcsoib3zFds4USH9bASPWBvJHRvfHYbxjR8QCHQ7y1vj/mfs
Qmacb2NPrz1tzpB3GarzqaYuLb54o71iJAAL61jGbmJEmu  queued  policy/12D3KooWEcsoib3zFds4USH9bASPWBvJHRvfHYbxjR8QCHQ7y1vj/mfs
QmRk9WAfArEpEPw6xc5hTUmKPeRM85fUCJoSqGPWYfMCB4  queued  policy/12D3KooWEcsoib3zFds4USH9bASPWBvJHRvfHYbxjR8QCHQ7y1vj/mfs
QmVYD7stnRgG4HHgrnkpbCUvLTvjLmZbhTJNe5ZNx7bDmh  queued  policy/12D3KooWEcsoib3zFds4USH9bASPWBvJHRvfHYbxjR8QCHQ7y1vj/mfs
QmTiCXsj9nwFCDpUxq3fAUvZW49s3FCHPtxb8KeYoGfmri  queued  policy/12D3KooWEcsoib3zFds4USH9bASPWBvJHRvfHYbxjR8QCHQ7y1vj/mfs

```

## Notes

-  `Replace`  at  Pinata will still produce duplicate pins, but for a different reason:
   -  This is temporary bug on their end and should not block us from switching to `Replace` method. 
    - Our CI should be green, because we don't use Pinata for tests there, and `Replace` from [rb-pinning-service-api](https://github.com/ipfs-shipyard/rb-pinning-service-api) works as expected.

### DIFF
--- a/cmd/ipfs/pinmfs.go
+++ b/cmd/ipfs/pinmfs.go
@@ -179,10 +179,12 @@ func pinMFS(
 	lsPinCh, lsErrCh := c.Ls(ctx, pinclient.PinOpts.FilterName(pinName), pinclient.PinOpts.FilterStatus(pinStatuses...))
 	existingRequestID := "" // is there any pre-existing MFS pin with pinName (for any CID)?
 	alreadyPinned := false  // is CID for current MFS already pinned?
+	pinTime := time.Now()
 	for ps := range lsPinCh {
 		existingRequestID = ps.GetRequestId()
-		if ps.GetPin().GetCid() == cid {
-			alreadyPinned = ps.GetStatus() != pinclient.StatusFailed
+		if ps.GetPin().GetCid() == cid && ps.GetStatus() != pinclient.StatusFailed {
+			alreadyPinned = true
+			pinTime = ps.GetCreated()
 			break
 		}
 	}
@@ -197,8 +199,8 @@ func pinMFS(
 
 	// CID of the current MFS root is already pinned, nothing to do
 	if alreadyPinned {
-		log.Infof("pinning MFS to %s: pin for %s already exists, skipping", svcName, cid)
-		return lastPin{Time: time.Now(), ServiceName: svcName, ServiceConfig: svcConfig, CID: cid}, nil
+		log.Infof("pinning MFS to %s: pin for %s exists since %q, skipping", svcName, cid, pinTime)
+		return lastPin{Time: pinTime, ServiceName: svcName, ServiceConfig: svcConfig, CID: cid}, nil
 	}
 
 	// Prepare Pin.name
@@ -241,5 +243,5 @@ func pinMFS(
 			return lastPin{}, err
 		}
 	}
-	return lastPin{Time: time.Now(), ServiceName: svcName, ServiceConfig: svcConfig, CID: cid}, nil
+	return lastPin{Time: pinTime, ServiceName: svcName, ServiceConfig: svcConfig, CID: cid}, nil
 }
--- a/cmd/ipfs/pinmfs.go
+++ b/cmd/ipfs/pinmfs.go
@@ -93,7 +93,7 @@ func pinMFSOnChange(configPollInterval time.Duration, cctx pinMFSContext, node p
 			// get the most recent MFS root cid
 			rootNode, err := node.RootNode()
 			if err != nil {
-				log.Errorf("pinning reading mfs root (%v)", err)
+				log.Errorf("pinning reading MFS root (%v)", err)
 				select {
 				case errCh <- err:
 				case <-cctx.Context().Done():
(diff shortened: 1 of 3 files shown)


cmd/ipfs/pinmfs.go
```diff
@@ -169,12 +169,16 @@ func pinMFS(
 		pinName = fmt.Sprintf("policy/%s/mfs", node.Identity().String())
 	}
 
-	// check if same pin exists
-	lsPinCh, lsErrCh := c.Ls(ctx, pinclient.PinOpts.FilterCIDs(cid), pinclient.PinOpts.FilterName(pinName))
-	pinFound := false
+	// check if MFS pin exists (across all possible states) and inspect its CID
+	pinStatuses := []pinclient.Status{pinclient.StatusQueued, pinclient.StatusPinning, pinclient.StatusPinned, pinclient.StatusFailed}
```
aschmahmann, line 173: Why failed? If the pinning service failed to pin, wouldn't we want it to try again?

cmd/ipfs/pinmfs.go
```diff
@@ -169,12 +169,16 @@ func pinMFS(
...
+	lsPinCh, lsErrCh := c.Ls(ctx, pinclient.PinOpts.FilterName(pinName), pinclient.PinOpts.FilterStatus(pinStatuses...))
+	existingRequestID := "" // is there any pre-existing MFS pin with pinName (for any CID)?
+	alreadyPinned := false  // is CID for current MFS already pinned?
 	for ps := range lsPinCh {
+		existingRequestID = ps.GetRequestId()
```
aschmahmann, line 178: IIUC this means we're replacing the oldest pin with the correct name. This doesn't seem unreasonable, but it's non-obvious so I wanted to clarify that this is the intended behavior.

lidel: Correct. The assumption is name is unique enough to not have more than a single pre-existing one, but if for some reason there is more, we just replace last one we inspected.

cmd/ipfs/pinmfs.go
lidel: This is a good point!

- [ ] TODO: check status and don't skip if the CID is the same as current MFS root.

cmd/ipfs/pinmfs.go
```diff
@@ -185,7 +189,10 @@ func pinMFS(
 		}
 		return lastPin{}, err
 	}
-	if pinFound {
+
+	// CID of the current MFS root is already pinned, nothing to do
+	if alreadyPinned {
+		log.Infof("pinning MFS to %s: pin for %s already exists, skipping", svcName, cid)
 		return lastPin{}, nil
 	}
```
petar, line 197: (This is a bug in the original PR too.) We should return here something like
lastPin{Time: time.Now(), ServiceName: svcName, ServiceConfig: svcConfig, CID: cid},
because upstream lastPin{} is taken to mean failed pinning.

lidel:
Quick update:
- When pin already exists, we now use creation time returned by remote service instead of `time.Now()`
- Added sharness test that confirms the MFS root pin gets replaced with new CID when MFS is modified.
- Normalized timestamps to UTC for easier debugging across services 
  (Pinning Services API defaults to UTC, but this makes it explicit)
- Tweaked conditions and logged strings a bit to produce human-readable logs
- (keeping this as a draft until I am able to test Replace operation against Pinata)

lidel:
@petar @aschmahmann I propose we merge these fixes/changes into https://github.com/ipfs/go-ipfs/pull/7798 so we avoid split-brain reviews (and we need to switch to `Replace` operation anyway).

aschmahmann:
merging as @petar gave the 👍. We can continue on the other PR.

PR #7869, opened 2021-01-16
Make ipns publish related code more generic

yareky:
This pull request will make IPNS publishing code more generic by adding peer.ID parameter and removing many redundant peer.IDFromPrivateKey(key) calls.


Stebalien:
The ID is always derived from the private key. Unfortunately, this patch will make it really easy to accidentally publish with mismatched keys/ids. It also doubles the amount of code without really adding anything.

PR #7883, opened 2021-01-27
Do not fetch recursive pins from pinner unnecessarily

gammazero:
When fetching all pins, the recursive pins are fetched from the pinner two times.  The second fetch is unnecessary and copies all recursive pins into a slice again.

Additionally, the output channel is now buffered.  This allows the goroutine to exit in the case the pinner returns an error and there is no reader for the output channel.  This might be possible if a canceled context causes the caller to abandon waiting to read the output of Ls().

### DIFF
--- a/core/coreapi/pin.go
+++ b/core/coreapi/pin.go
@@ -219,6 +219,9 @@ func (p *pinInfo) Err() error {
 }
 
 // pinLsAll is an internal function for returning a list of pins
+//
+// The caller must keep reading results until the channel is closed to prevent
+// leaking the goroutine that is fetching pins.
 func (api *PinAPI) pinLsAll(ctx context.Context, typeStr string) <-chan coreiface.Pin {
 	out := make(chan coreiface.Pin, 1)
 
--- a/core/coreapi/pin.go
+++ b/core/coreapi/pin.go
@@ -220,7 +220,7 @@ func (p *pinInfo) Err() error {
 
 // pinLsAll is an internal function for returning a list of pins
 func (api *PinAPI) pinLsAll(ctx context.Context, typeStr string) <-chan coreiface.Pin {
-	out := make(chan coreiface.Pin)
+	out := make(chan coreiface.Pin, 1)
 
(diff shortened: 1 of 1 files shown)


core/coreapi/pin.go
```diff
@@ -220,7 +220,7 @@ func (p *pinInfo) Err() error {
 
 // pinLsAll is an internal function for returning a list of pins
 func (api *PinAPI) pinLsAll(ctx context.Context, typeStr string) <-chan coreiface.Pin {
-	out := make(chan coreiface.Pin)
+	out := make(chan coreiface.Pin, 1)
```
Stebalien, line 223: why do we need this?

Stebalien (approved):

core/coreapi/pin.go
Stebalien: > Additionally, the output channel is now buffered. This allows the goroutine to exit in the case the pinner returns an error and there is no reader for the output channel. This might be possible if a canceled context causes the caller to abandon waiting to read the output of Ls().

Ah, I see. Unfortunately, that doesn't really fix the issue as we're not guaranteed to see that the context has been canceled immediately.

See https://github.com/ipfs/interface-go-ipfs-core/pull/62

gammazero: Yes, you are correct that if the caller abandons waiting, we may still try to deliver other results before delivering the error... in which case the goroutine is back to blocking on writing to the channel.  The idea is that this _may help_ if the caller gives up waiting (context times out, or is canceled) due to waiting too long for results.

Since this does not fix actually prevent leaking the goroutine, I can remove the buffering, and instead add a comment stating that the caller must keep reading results until the channel is closed to prevent leaking the goroutine.

Stebalien: I'm fine keeping the buffering _just_ to be nice. I'd also add a comment.

Let's punt on the interface changes for now.

gammazero: Done

Stebalien (approved):

PR #7897, opened 2021-02-03
Fetch big files by "stream" CID.

petar:
- [x] generate a "chunking manifest" of files added to ipfs, to be used for inserting the files into a "big file store" (or another name)
- [ ] compute the stream cid of the added file

### DIFF
--- a/core/coreunix/add.go
+++ b/core/coreunix/add.go
@@ -11,9 +11,9 @@ import (
 	"github.com/ipfs/go-cid"
 	bstore "github.com/ipfs/go-ipfs-blockstore"
 	chunker "github.com/ipfs/go-ipfs-chunker"
-	"github.com/ipfs/go-ipfs-files"
-	"github.com/ipfs/go-ipfs-pinner"
-	"github.com/ipfs/go-ipfs-posinfo"
+	files "github.com/ipfs/go-ipfs-files"
+	pin "github.com/ipfs/go-ipfs-pinner"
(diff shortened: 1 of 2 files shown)


core/bigfilestore/bigfilestore.go
```diff
@@ -0,0 +1,297 @@
...
+	case nil:
+		return nil
+	case blockstore.ErrNotFound:
+		if err1 == blockstore.ErrNotFound {
+			return blockstore.ErrNotFound
+		}
+		return nil
+	default:
+		return err2
+	}
```
gammazero, line 184: nit: maybe simplify and replace the `switch { ... }` with:
```go
if err2 != nil {
    if err2 == blockstore.ErrNotFound {
        return err1
    }
    return err2
}
return nil
```

core/bigfilestore/bigfilestore.go
```diff
@@ -0,0 +1,297 @@
...
+func (b *BigFileStore) GetSize(c cid.Cid) (int, error) {
+	size, err := b.bs.GetSize(c)
+	switch err {
+	case nil:
+		return size, nil
+	case blockstore.ErrNotFound:
+		chunks, err := b.GetBigBlock(c)
+		if err == ds.ErrNotFound {
+			return 0, blockstore.ErrNotFound
+		}
```
gammazero, line 239: This `if err == ds.ErrNoFound {` is not needed.  Just use the check below.

PR #7900, opened 2021-02-05
all: gofmt -s

mvdan:
This "simplify" flag mainly removes redundant types in expressions.

Not particularly important, but a nice change that also makes gopls not
show warnings.


willscott (approved):
is there a way to have this happen in the CI lint check as well, so it stays this way?

https://github.com/ipfs/go-ipfs/blob/master/.circleci/config.yml#L362
https://github.com/ipfs/go-ipfs/blob/master/mk/golang.mk#L67-L74
https://github.com/ipfs/go-ipfs/blob/master/bin/test-go-fmt#L7

mvdan:
Yep, from what you shared it seems like we could just add `-s` to that last file. I did not do that because the CI and Makefile setup here seems very complex, so I default to not touching it. But I'm happy to do that if it sounds good. /cc @lidel

willscott:
Yeah, i think adding -s to the gofmt line makes sense. i also haven't looked into the golanglint-ci config to see if there's a way to specify that for it's use of gofmt in it's lint checks.

assets/bindata.go
```diff
@@ -370,18 +370,18 @@ type bintree struct {
 }
 
 var _bintree = &bintree{nil, map[string]*bintree{
```
Stebalien, line 372: We'll need to modify https://github.com/ipfs/go-ipfs/blob/884a5aebd748336db963e91daf7b916054486a99/assets/assets.go#L5 to pass `-s`.

Stebalien (approved):
One nit but otherwise :heart:.

Stebalien:
> CI and Makefile setup here seems very complex...

1. You open `info make`.
2. You trigger "timeskip". See you in 10 years.

Stebalien:
(I believe I've applied the required fixes)

mvdan:
Beware that Go tools including gofmt don't understand GNU flags, so you probably want `-s -w` instead of `-sw` :)

Stebalien:
:rage:

PR #7902, opened 2021-02-05
feat: support non-ICANN DNSLink names

lidel:
This PR aims to decouple DNSLink resolution from the TLD safelist, enabling users to experiment with new TLDs with regular go-ipfs binary. 

Future work will include configurable DNS resolver per TLD, but this PR is scoped to simply make go-ipfs use DNS protocol, and don't care about TLDs specific to some explicitly blessed DNS network (ICANN, ENS, OpenNIC, etc).

- removes reliance on TLD safelist from `jbenet/go-is-domain`  as suggested in  https://github.com/jbenet/go-is-domain/issues/16#issuecomment-740967950 and  https://github.com/jbenet/go-is-domain/pull/19#pullrequestreview-584687951

- switch DNSLink checks to `net.isDomainName` which only cares about    name being compliant with RFC 1035, RFC 3696, and is not concerned   if TLD was blessed by ICANN or some other authority 

cc @autonome 

# TODO

- [ ]  find exported alternative to `net.isDomainName` or simply vendor snippet from https://golang.org/src/net/dnsclient.go?s=1626:1658#L53

### DIFF
--- a/core/corehttp/hostname.go
+++ b/core/corehttp/hostname.go
@@ -13,7 +13,6 @@ import (
 	core "github.com/ipfs/go-ipfs/core"
 	coreapi "github.com/ipfs/go-ipfs/core/coreapi"
 	namesys "github.com/ipfs/go-ipfs/namesys"
-	isd "github.com/jbenet/go-is-domain"
 	"github.com/libp2p/go-libp2p-core/peer"
 	mbase "github.com/multiformats/go-multibase"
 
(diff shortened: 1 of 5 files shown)


Stebalien:
@lidel what's the status here?

lidel:
Want to do this, but low priority for now (we need support for DNS over HTTPS for this to be truly useful).
Will get back to it when got spare time, or will make this part of  https://github.com/protocol/web3-dev-team/pull/42

Stebalien:
If this isn't likely to land, can we close it and reopen later?

lidel:
Makes sense, especially now that namesys got moved to https://github.com/ipfs/go-namesys

bertrandfalguiere:
> If this isn't likely to land, can we close it and reopen later?

Shouldn't it just be moved to "ice box" or labelled as such? Since the feature is still not implemented?

lidel:
I created https://github.com/ipfs/go-ipfs/issues/8060 and added it to maintenance backlog (small unit of work, can be completed within one week, we need to do it soon anyway)

PR #7903, opened 2021-02-05
fix(build): preserve the GOFLAGS environment variable

### LINKED ISSUE #7881: Fails to build offline with vendored dependencies (building package on Linux / openSUSE)
#### Version information:
0.7.0 tarball

When trying to build package for openSUSE/Factory (complete build instructions are [available](https://build.opensuse.org/package/view_file/home:mcepl:work/go-ipfs/go-ipfs.spec?expand=1)), which is basically:
```sh
export GOCACHE=$(readlink -f vendor/)
BUILDMOD="-buildmode=pie -trimpath -modcacherw"
export CFLAGS="%{optflags}"
make GOFLAGS="-buildmode=pie -trimpath -modcacherw" nofuse
```
build fails in the openSUSE build system, because it tries to access network (but all build systems of all major Linux distributions don’t have network access, because of security reasons):
```
[    8s] + make 'GOFLAGS=-buildmode=pie -trimpath -modcacherw' nofuse
[    8s] go: bazil.org/fuse@v0.0.0-20200117225306-7b5117fecadc: Get "https://proxy.golang.org/bazil.org/fuse/@v/v0.0.0-20200117225306-7b5117fecadc.mod": dial tcp: lookup proxy.golang.org on [::1]:53: read udp [::1]:56410->[::1]:53: read: connection refused
[    8s] go version go1.14.13 linux/amd64
[    8s] bin/check_go_version 1.14.4
[    8s] plugin/loader/preload.sh > plugin/loader/preload.go
[    8s] go fmt plugin/loader/preload.go >/dev/null
[    8s] go: bazil.org/fuse@v0.0.0-20200117225306-7b5117fecadc: Get "https://proxy.golang.org/bazil.org/fuse/@v/v0.0.0-20200117225306-7b5117fecadc.mod": dial tcp: lookup proxy.golang.org on [::1]:53: read udp [::1]:44033->[::1]:53: read: connection refused
[    8s] make: *** [plugin/loader/Rules.mk:9: plugin/loader/preload.go] Error 1
[    8s] error: Bad exit status from /var/tmp/rpm-tmp.zpp5aA (%build)
[    8s]
```

Vendoring was done by running ``make build && go mod vendor`` in the unpacked tarball.

[Complete build log](https://github.com/ipfs/go-ipfs/files/5876818/_log.txt) with all details of package versions etc.

Stebalien:
Previously, we avoided this variable because we were storing some flags with spaces (potentially). Unfortunately, that means these flags are not set for some commands (e.g., `go fmt`) leading to issues when building from vendored sources.

Now we:

1. Use GOFLAGS where possible.
2. Pass tags and special flags on the commandline where necessary.

This should fix #7881.

### DIFF
--- a/cmd/ipfs/Rules.mk
+++ b/cmd/ipfs/Rules.mk
@@ -13,7 +13,7 @@ PATH := $(realpath $(d)):$(PATH)
 # DEPS_OO_$(d) += merkledag/pb/merkledag.pb.go namesys/pb/namesys.pb.go
 # DEPS_OO_$(d) += pin/internal/pb/header.pb.go unixfs/pb/unixfs.pb.go
 
-$(d)_flags =-ldflags="-X "github.com/ipfs/go-ipfs".CurrentCommit=$(git-hash)"
+$(d)_flags =-ldflags=-X="github.com/ipfs/go-ipfs.CurrentCommit=$(git-hash)"
 
 $(d)-try-build $(IPFS_BIN_$(d)): GOFLAGS += $(cmd/ipfs_flags)
 
--- a/mk/golang.mk
+++ b/mk/golang.mk
@@ -22,16 +22,16 @@ TEST_GO :=
 TEST_GO_BUILD :=
 CHECK_GO :=
 
-go-pkg-name=$(shell $(GOCC) list $(go-tags) github.com/ipfs/go-ipfs/$(1))
-go-main-name=$(notdir $(call go-pkg-name,$(1)))$(?exe)
-go-curr-pkg-tgt=$(d)/$(call go-main-name,$(d))
-go-pkgs=$(shell $(GOCC) list github.com/ipfs/go-ipfs/...)
-
 # Go tags cannot be set in GOFLAGS.
 go-tags=$(if $(GOTAGS), -tags="$(call join-with,$(space),$(GOTAGS))")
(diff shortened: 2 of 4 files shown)


mvdan:
Out of curiosity, why do we need `-mod=vendor` explicitly? If the vendor directory was created by a `go mod vendor` on a recent Go version, Go 1.14 and later should use the vendor folder automatically by default:

https://golang.org/doc/go1.14#go-command

> When the main module contains a top-level vendor directory and its go.mod file specifies go 1.14 or higher, the go command now defaults to -mod=vendor for operations that accept that flag. A new value for that flag, -mod=mod, causes the go command to instead load modules from the module cache (as when no vendor directory is present).

Stebalien:
@mvdan probably upstream bugs.

mvdan:
Right, which is why I'm trying to dig a bit, to report a bug upstream if there is one. Do you have an example of what breaks without `-mod=vendor` or just using `GOFLAGS`?

Stebalien:
Looking at the output in #7881, it looks like `go list` was failing. But I can't reproduce that _manually_ so I wonder if it was somehow due to the build environment. You're right that the presence of a `vendor` directory should make all this "just work".

aschmahmann:
This should not be necessary to fix the linked issue, more investigation into the core issue required.

PR #7904, opened 2021-02-06
peering: add logs before many-second waits

mvdan:
This test takes a full minute to run, and I was honestly thinking my run
of "go test -v" had simply hung, as I saw no output and no apparent
resource usage.

The least we can do is print a few log messages before the potentially
long waits, to hint that we're still making progress. Each of these
"Eventually" and "Never" calls ends up blocking the test for a few
seconds at a time.

### DIFF
--- a/peering/peering_test.go
+++ b/peering/peering_test.go
@@ -57,11 +57,13 @@ func TestPeeringService(t *testing.T) {
 	require.NoError(t, ps1.Start())
 
 	// We should eventually connect.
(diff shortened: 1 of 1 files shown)


mvdan:
The CI failure looks like some sort of unrelated timeout flake.

Stebalien (approved):
We _should_ make this faster, but this is a good stop-gap.

Stebalien:
Hm... we may need to increase timeouts for the linter again...

PR #7913, opened 2021-02-15
docs: Pinning.RemoteServices.Policies.MFS

lidel:
This PR adds docs for `Pinning.RemoteServices.Policies` 

cc https://github.com/ipfs/go-ipfs/issues/7559 https://github.com/ipfs/ipfs-docs/pull/369 https://github.com/ipfs/ipfs-gui/issues/91

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -895,7 +895,7 @@ Type: `bool`
 Optional name to use for remote pin that represents MFS root CID.  
 When left empty, default name will be generated.
 
-Default: `"policy/12.../mfs"`
+Default: `"policy/{PeerID}/mfs"`, e.g. `"policy/12.../mfs"`
 
 Type: `string`
 
--- a/docs/config.md
+++ b/docs/config.md
@@ -879,8 +879,8 @@ Contains additional opt-in policies for the remote pinning service.
 When this policy is enabled, it follows changes to MFS
 and updates the pin for MFS root on the configured remote service.
 
-A pin request to the remote service is sent only when MFS root has changed
-and time has passed, since the previous pin is bigger than `RepinInterval`.
+A pin request to the remote service is sent only when MFS root CID has changed
+and enough time has passed since the previous request (determined by `RepinInterval`).
 
 ###### `Pinning.RemoteServices: Policies.MFS.Enabled`
 
--- a/docs/config.md
+++ b/docs/config.md
@@ -181,6 +181,8 @@ does (e.g, `"1d2h4m40.01s"`).
         - [`Pinning.RemoteServices.API`](#pinningremoteservices-api)
           - [`Pinning.RemoteServices.API.Endpoint`](#pinningremoteservices-apiendpoint)
           - [`Pinning.RemoteServices.API.Key`](#pinningremoteservices-apikey)
+        - [`Pinning.RemoteServices.Policies`](#pinningremoteservices-policies)
+          - [`Pinning.RemoteServices.Policies.MFS`](#pinningremoteservices-policiesmfs)
 - [`Pubsub`](#pubsub)
     - [`Pubsub.Router`](#pubsubrouter)
     - [`Pubsub.DisableSigning`](#pubsubdisablesigning)
@@ -847,7 +849,7 @@ Example:
         "API" : {
           "Endpoint" : "https://pinningservice.tld:1234/my/api/path",
           "Key" : "someOpaqueKey"
-        }
+				}
       }
     }
   }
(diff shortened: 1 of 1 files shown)


docs/config.md
```diff
@@ -868,6 +870,44 @@ The key through which access to the pinning service is granted
...
+Default: `false`
+
+Type: `bool`
+
+###### `Pinning.RemoteServices: Policies.MFS.PinName`
+
+Optional name to use for remote pin that represents MFS root CID.  
+When left empty, default name will be generated.
+
+Default: `"policy/12.../mfs"`
```
aschmahmann, line 898: I think we should make it more explicit that we're using the PeerID.
```suggestion
Default: `"policy/{PeerID}/mfs"`, e.g. `"policy/12.../mfs"`
```

docs/config.md
```diff
@@ -868,6 +870,44 @@ The key through which access to the pinning service is granted
 
 Type: `string`
 
+#### `Pinning.RemoteServices: Policies`
+
+Contains additional opt-in policies for the remote pinning service
+
+##### `Pinning.RemoteServices: Policies.MFS`
+
+When this policy is enabled it will be following changes to MFS
```
jessicaschilling, line 879:
```suggestion
When this policy is enabled, it follows changes to MFS
```

docs/config.md
```diff
@@ -868,6 +870,44 @@ The key through which access to the pinning service is granted
...
+and updating the pin for MFS root on the configured remote service.
```
jessicaschilling, line 880:
```suggestion
and updates the pin for MFS root on the configured remote service.
```

docs/config.md
```diff
@@ -868,6 +870,44 @@ The key through which access to the pinning service is granted
...
+
+Pin request to the remote service is sent only when MFS root changed
```
jessicaschilling, line 882:
```suggestion
A pin request to the remote service is sent only when MFS root has changed
```

docs/config.md
```diff
@@ -868,6 +870,44 @@ The key through which access to the pinning service is granted
...
+and time passed since the previous pin is bigger than `RepinInterval`.
```
jessicaschilling, line 883:
```suggestion
and time has passed, since the previous pin is bigger than `RepinInterval`.
```
This is unclear to me. Does "since" mean "because" in this case?

docs/config.md
```diff
@@ -868,6 +870,44 @@ The key through which access to the pinning service is granted
...
+
+Controls if this policy is active.
+
+Default: `false`
+
+Type: `bool`
+
+###### `Pinning.RemoteServices: Policies.MFS.PinName`
+
+Optional name to use for remote pin that represents MFS root CID.
```
jessicaschilling, line 895:
```suggestion
Optional name to use for a remote pin that represents the MFS root CID.  
```

docs/config.md
```diff
@@ -868,6 +870,44 @@ The key through which access to the pinning service is granted
...
+Controls if this policy is active.
+
+Default: `false`
+
+Type: `bool`
+
+###### `Pinning.RemoteServices: Policies.MFS.PinName`
+
+Optional name to use for remote pin that represents MFS root CID.  
+When left empty, default name will be generated.
```
jessicaschilling, line 896:
```suggestion
When left empty, a default name will be generated.
```

docs/config.md
```diff
@@ -868,6 +870,44 @@ The key through which access to the pinning service is granted
...
+
+Default: `"policy/12.../mfs"`
+
+Type: `string`
+
+###### `Pinning.RemoteServices: Policies.MFS.RepinInterval`
+
+Defines how often (at most) the pin request should be sent to remote service.
```
jessicaschilling, line 904:
```suggestion
Defines how often (at most) the pin request should be sent to the remote service.  
```

docs/config.md
```diff
@@ -868,6 +870,44 @@ The key through which access to the pinning service is granted
...
+When left empty, default name will be generated.
+
+Default: `"policy/12.../mfs"`
+
+Type: `string`
+
+###### `Pinning.RemoteServices: Policies.MFS.RepinInterval`
+
+Defines how often (at most) the pin request should be sent to remote service.  
+When left empty, default interval will be used. Values lower than `1m` will be ignored.
```
jessicaschilling, line 905:
```suggestion
If left empty, the default interval will be used. Values lower than `1m` will be ignored.
```

jessicaschilling (changes requested):
Few text nits and one question.

docs/config.md
```diff
@@ -868,6 +870,44 @@ The key through which access to the pinning service is granted
...
+
+Contains additional opt-in policies for the remote pinning service.
+
+##### `Pinning.RemoteServices: Policies.MFS`
+
+When this policy is enabled, it follows changes to MFS
+and updates the pin for MFS root on the configured remote service.
+
+A pin request to the remote service is sent only when MFS root has changed
+and time has passed, since the previous pin is bigger than `RepinInterval`.
```
lidel, line 883: @jessicaschilling  hm.. i think your question was valid, and I did not fully get what you meant when I merged this.

IPFS node  checks if MFS changed every 30s and if it did change, then it checks when the last successful pin was executed.

If `last_pin_time` + `RepinInterval` > `now`, then it waits until it becomes otherwise.
In other words, we never send pin request more often than the value in `RepinInterval` (which is 5 minutes by default).

Would below be better?

```suggestion
A pin request to the remote service is sent only when MFS root CID has changed
and enough time has passed since the previus request (determined by `RepinInterval`).
```

jessicaschilling: Makes perfect sense, thank you!
one note: `previous` 😊

lidel:
@aschmahmann I believe this is ready for your final :eyes:

jessicaschilling (approved):

petar (approved):
lgtm

aschmahmann (approved):

PR #7914, opened 2021-02-16
Create codeql-analysis.yml

naveensrinivasan:
What type of PR is this?
Enable codeql analysis to capture any security issues in code.

/kind cleanup

What this PR does / why we need it:
Discover vulnerabilities across a codebase with CodeQL

Which issue(s) this PR fixes:
None

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,40 @@
+# https://docs.github.com/en/free-pro-team@latest/github/finding-security-vulnerabilities-and-errors-in-your-code/configuring-code-scanning#changing-the-languages-that-are-analyzed
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,40 @@
...
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'go', 'python' ]
```
willscott, line 21: there's no python in this repo. is there a reason you test it?

naveensrinivasan: The bot built that by analyzing the code. I can remove python. :)

Stebalien (approved):
After removing python, this seems reasonable. It's maintained by GitHub so there shouldn't be any security issues with pulling the action in directly rather than forking.

Stebalien:
Thanks!

PR #7927, opened 2021-02-19
Set supported platforms by go-version

hannahhoward:
Looks like darwin-386 is in the list of supported platforms still as of 0.8.0, which won't compile on go-1.15 and up. And now there's darwin-arm64 (native Apple M1) starting on Go 1.16

This PR set supported platforms correctly by go version (no darwin-386 for >=1.15, no darwin-arm64 unless >=1.16)

It also makes a change to check-version as \+ is interpreted literally for `expr` on several POSIX systems including Mac.
License: MIT

### DIFF
--- a/mk/util.mk
+++ b/mk/util.mk
@@ -17,9 +17,13 @@ SUPPORTED_PLATFORMS += linux-arm64
 SUPPORTED_PLATFORMS += linux-386
 SUPPORTED_PLATFORMS += linux-amd64
 
-SUPPORTED_PLATFORMS += darwin-386
 SUPPORTED_PLATFORMS += darwin-amd64
-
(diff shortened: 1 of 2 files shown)


mk/util.mk
```diff
@@ -17,9 +17,13 @@ SUPPORTED_PLATFORMS += linux-arm64
 SUPPORTED_PLATFORMS += linux-386
 SUPPORTED_PLATFORMS += linux-amd64
 
-SUPPORTED_PLATFORMS += darwin-386
 SUPPORTED_PLATFORMS += darwin-amd64
-
+ifeq ($(shell bin/check_go_version "1.15.0" 2>/dev/null; echo $$?),1)
+SUPPORTED_PLATFORMS += darwin-386
+endif
```
Stebalien, line 23: I'd drop support for this entirely, but we can do that later.

Stebalien (approved):

PR #7930, opened 2021-02-19
feat(gw): /ipfs/ipfs/{cid} → /ipfs/{cid}

lidel:
This PR enables Gateways to recover from invalid paths like  `/ipfs/ipfs/{cid}`  and redirect them to proper one, when possible.

_raison d'être_  is to 
- smooth out unforeseen addressing issues during initial adoption, and fix cosmetic mistakes like the value in `image` tag [here](https://dweb.link/ipfs/bafybeien3m7mdn6imm425vc2s22erzyhbvk5n3ofzgikkhmdkh5cuqbpbq/metadata.json)  :cat2::rainbow_flag:  
-  make broken URIs resolve in Brave and Opera (because thats where most of people will paste those broken `ipfs://` URI to address bar)
- provide hint to developers on what should be fixed

cc @autonome

### DIFF
--- a/core/corehttp/gateway_handler.go
+++ b/core/corehttp/gateway_handler.go
@@ -238,7 +238,7 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
 
 	parsedPath := ipath.New(urlPath)
 	if pathErr := parsedPath.IsValid(); pathErr != nil {
-		if fixupSuperfluousNamespace(w, urlPath, r.URL.RawQuery) {
+		if prefix == "" && fixupSuperfluousNamespace(w, urlPath, r.URL.RawQuery) {
 			// the error was due to redundant namespace, which we were able to fix
 			// by returning error/redirect page, nothing left to do here
 			return
--- a/core/corehttp/gateway_handler.go
+++ b/core/corehttp/gateway_handler.go
@@ -39,7 +39,17 @@ var onlyAscii = regexp.MustCompile("[[:^ascii:]]")
 
 // HTML-based redirect for errors which can be recovered from, but we want
 // to provide hint to people that they should fix things on their end.
-var redirectTemplate = template.Must(template.New("redirect").Parse(`<!DOCTYPE html><html><head><meta http-equiv="Content-Type" content="text/html; charset=utf-8" /><meta http-equiv="refresh" content="10;url={{.RedirectURL}}" /><link rel="canonical" href="{{.RedirectURL}}" /></head><body><pre>{{.ErrorMsg}}</pre><pre>(if a redirect does not happen in 10 seconds, use "{{.SuggestedPath}}" instead)</pre></body></html>`))
+var redirectTemplate = template.Must(template.New("redirect").Parse(`<!DOCTYPE html>
+<html>
+	<head>
+		<meta charset="utf-8">
+		<meta http-equiv="refresh" content="10;url={{.RedirectURL}}" />
+		<link rel="canonical" href="{{.RedirectURL}}" />
+	</head>
+	<body>
+		<pre>{{.ErrorMsg}}</pre><pre>(if a redirect does not happen in 10 seconds, use "{{.SuggestedPath}}" instead)</pre>
+	</body>
+</html>`))
 
 type redirectTemplateData struct {
 	RedirectURL   string
@@ -228,27 +238,9 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
 
 	parsedPath := ipath.New(urlPath)
 	if pathErr := parsedPath.IsValid(); pathErr != nil {
-		// Attempt to fix redundant /ipfs/ namespace as long as resulting
-		// 'intended' path is valid.  This is in case gremlins were tickled
-		// wrong way and user ended up at /ipfs/ipfs/{cid} or /ipfs/ipns/{id}
-		// like in bafybeien3m7mdn6imm425vc2s22erzyhbvk5n3ofzgikkhmdkh5cuqbpbq
-		// :^))
-		intendedPath := ipath.New(strings.TrimPrefix(urlPath, "/ipfs"))
-		if err := intendedPath.IsValid(); err == nil {
-			intendedURL := strings.Replace(r.URL.String(), urlPath, intendedPath.String(), 1)
-			// return HTML that
-			// - points at correct canonical path via <link> header
-			// - displays error
-			// - redirects to intendedURL after a delay
-			err = redirectTemplate.Execute(w, redirectTemplateData{
-				RedirectURL:   intendedURL,
-				SuggestedPath: intendedPath.String(),
-				ErrorMsg:      pathErr.Error(),
-			})
-			if err != nil {
-				internalWebError(w, err)
-				return
-			}
+		if fixErr := fixupSuperfluousNamespace(w, r, pathErr, urlPath); fixErr == nil {
+			// the error was due to redundant namespace, which we were able to fix
+			// by returning error/redirect page, nothing left to do here
 			return
 		}
 		// unable to fix path, returning error
@@ -815,4 +807,28 @@ func preferred404Filename(acceptHeaders []string) (string, string, error) {
 	}
 
 	return "", "", fmt.Errorf("there is no 404 file for the requested content types")
+}
+
+// Attempt to fix redundant /ipfs/ namespace as long as resulting
+// 'intended' path is valid.  This is in case gremlins were tickled
+// wrong way and user ended up at /ipfs/ipfs/{cid} or /ipfs/ipns/{id}
+// like in bafybeien3m7mdn6imm425vc2s22erzyhbvk5n3ofzgikkhmdkh5cuqbpbq :^))
+func fixupSuperfluousNamespace(w http.ResponseWriter, r *http.Request, pathErr error, urlPath string) error {
+	intendedPath := ipath.New(strings.TrimPrefix(urlPath, "/ipfs"))
+	err := intendedPath.IsValid()
+	if err != nil {
+		// not a superfluous namespace
+		return err
+	}
+	intendedURL := strings.Replace(r.URL.String(), urlPath, intendedPath.String(), 1)
+	// return HTTP 400 (Bad Request) with HTML error page that:
+	// - points at correct canonical path via <link> header
+	// - displays human-readable error
+	// - redirects to intendedURL after a short delay
+	w.WriteHeader(http.StatusBadRequest)
+	return redirectTemplate.Execute(w, redirectTemplateData{
+		RedirectURL:   intendedURL,
+		SuggestedPath: intendedPath.String(),
+		ErrorMsg:      fmt.Sprintf("invalid path: %q should be %q", urlPath, intendedPath.String()),
+	})
 }
--- a/core/corehttp/gateway_handler.go
+++ b/core/corehttp/gateway_handler.go
@@ -238,7 +238,7 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
 
 	parsedPath := ipath.New(urlPath)
 	if pathErr := parsedPath.IsValid(); pathErr != nil {
-		if fixupSuperfluousNamespace(w, r, urlPath) {
+		if fixupSuperfluousNamespace(w, urlPath, r.URL.RawQuery) {
 			// the error was due to redundant namespace, which we were able to fix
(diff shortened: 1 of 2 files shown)


core/corehttp/gateway_handler.go
```diff
@@ -217,6 +217,17 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
 
 	parsedPath := ipath.New(urlPath)
 	if err := parsedPath.IsValid(); err != nil {
+		// Attempt to fix redundant /ipfs/ namespace as long resulting
```
autonome, line 220: missing word "as long as"

autonome (approved):

Stebalien:
This seems weird. This will magically make some _invalid_ URLs "just work". Those invalid URLs will then propagate and become "normal" (requiring support).

Instead of patching go-ipfs to fix URLs like this for people, we should provide clear error messages describing what may/may not have gone wrong.

lidel:
@Stebalien tbh we already have pretty clear error message:
https://dweb.link/ipfs/ipfs/Qmcg8f4F9cig2JWXunxJcdBe58Q5myYXPmGfuMn1TVeswD/nft.mp4
produces:
```
invalid ipfs path: invalid path "/ipfs/ipfs/Qmcg8f4F9cig2JWXunxJcdBe58Q5myYXPmGfuMn1TVeswD/nft.mp4": invalid CID: selected encoding not supported
```

I don't think we can make things more clear than "invalid path /ipfs/ipfs". 

You have a valid point that silently fixing this == enabling misconfiguration to go unnoticed. 
What if we return HTML with error message + redirect after 5seconds?  

```html
<!DOCTYPE html><html><head><meta http-equiv="Content-Type" content="text/html; charset=utf-8" /><meta http-equiv="refresh" content="5;url={intendedURL}" /></head><body>invalid ipfs path: invalid path "{urlPath}": invalid CID: selected encoding not supported</body></html>
````

**Comment by Stebalien** [comment]
Yeah, that seems reasonable.

**Comment by lidel** [comment]
@Stebalien implemented, looks like this:

>

**Comment by Stebalien** [comment]
It may be nicer to replace the entire error message explaining what the issue is, but either way works for me. "/ipfs/ipfs/Qm... should be /ipfs/Qm..." would make more sense than "invalid CID".

File: core/corehttp/gateway_handler.go [hunk 2]
```diff
@@ -36,6 +37,16 @@ const (
 
 var onlyAscii = regexp.MustCompile("[[:^ascii:]]")
 
+// HTML-based redirect for errors which can be recovered from, but we want
+// to provide hint to people that they should fix things on their end.
+var redirectTemplate = template.Must(template.New("redirect").Parse(`<!DOCTYPE html><html><head><meta http-equiv="Content-Type" content="text/html; charset=utf-8" /><meta http-equiv="refresh" content="10;url={{.RedirectURL}}" /><link rel="canonical" href="{{.RedirectURL}}" /></head><body><pre>{{.ErrorMsg}}</pre><pre>(if a redirect does not happen in 10 seconds, use "{{.SuggestedPath}}" instead)</pre></body></html>`))
```
Stebalien, line 42: Some newlines would be nice (allowed by backtick strings).

core/corehttp/gateway_handler.go
```diff
@@ -216,8 +227,32 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
 	}
 
 	parsedPath := ipath.New(urlPath)
-	if err := parsedPath.IsValid(); err != nil {
-		webError(w, "invalid ipfs path", err, http.StatusBadRequest)
+	if pathErr := parsedPath.IsValid(); pathErr != nil {
+		// Attempt to fix redundant /ipfs/ namespace as long as resulting
```
Stebalien, line 231: Can we break this into some form of "fix path" helper function?

core/corehttp/gateway_handler.go
```diff
@@ -216,8 +227,32 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
...
+			intendedURL := strings.Replace(r.URL.String(), urlPath, intendedPath.String(), 1)
+			// return HTML that
+			// - points at correct canonical path via <link> header
+			// - displays error
+			// - redirects to intendedURL after a delay
+			err = redirectTemplate.Execute(w, redirectTemplateData{
+				RedirectURL:   intendedURL,
+				SuggestedPath: intendedPath.String(),
+				ErrorMsg:      pathErr.Error(),
+			})
```
Stebalien, line 247: This could be used for XSS. Our error messages usually don't allow arbitrary HTML. We likely need to do some escaping.

core/corehttp/gateway_handler.go
```diff
@@ -216,8 +227,32 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
...
+		// like in bafybeien3m7mdn6imm425vc2s22erzyhbvk5n3ofzgikkhmdkh5cuqbpbq
+		// :^))
+		intendedPath := ipath.New(strings.TrimPrefix(urlPath, "/ipfs"))
+		if err := intendedPath.IsValid(); err == nil {
+			intendedURL := strings.Replace(r.URL.String(), urlPath, intendedPath.String(), 1)
+			// return HTML that
+			// - points at correct canonical path via <link> header
+			// - displays error
+			// - redirects to intendedURL after a delay
+			err = redirectTemplate.Execute(w, redirectTemplateData{
```
Stebalien, line 243: This should return a status code like "invalid request".

core/corehttp/gateway_handler.go
```diff
@@ -216,8 +227,32 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
...
 	parsedPath := ipath.New(urlPath)
-	if err := parsedPath.IsValid(); err != nil {
-		webError(w, "invalid ipfs path", err, http.StatusBadRequest)
+	if pathErr := parsedPath.IsValid(); pathErr != nil {
+		// Attempt to fix redundant /ipfs/ namespace as long as resulting
+		// 'intended' path is valid.  This is in case gremlins were tickled
+		// wrong way and user ended up at /ipfs/ipfs/{cid} or /ipfs/ipns/{id}
+		// like in bafybeien3m7mdn6imm425vc2s22erzyhbvk5n3ofzgikkhmdkh5cuqbpbq
+		// :^))
+		intendedPath := ipath.New(strings.TrimPrefix(urlPath, "/ipfs"))
```
Stebalien, line 236: I'd slightly rather test these cases explicitly (`/ipfs/ipns`, `/ipfs/ipfs`, etc.). It's not critical, but it took me a few minutes to prove to myself that there was no way this could go wrong (e.g., can't redirect `/ipfs` to `/`, etc.).

(I don't feel strongly about this)

Stebalien (changes requested):

core/corehttp/gateway_handler.go
lidel: Fixed in https://github.com/ipfs/go-ipfs/pull/7930/commits/b81b7549d34e0b549c0035286bdd02b2be50030c

core/corehttp/gateway_handler.go
lidel: we use `html/template` which takes care of escaping: 

```
<pre>invalid path: &#34;/ipfs/ipfs/QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR&#34; should be &#34;/ipfs/QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR&#34;</pre>
```

After https://github.com/ipfs/go-ipfs/pull/7930/commits/b81b7549d34e0b549c0035286bdd02b2be50030c  we also  pass string to template via  `%q`

core/corehttp/gateway_handler.go
lidel: Changed to HTTP 400 (StatusBadRequest) in https://github.com/ipfs/go-ipfs/pull/7930/commits/b81b7549d34e0b549c0035286bdd02b2be50030c

core/corehttp/gateway_handler.go
lidel: Added explicit check in https://github.com/ipfs/go-ipfs/pull/7930/commits/450baef0e9c113e4751b559dd0562f7c98302668

core/corehttp/gateway_handler.go
Stebalien: Got it. I assumed `text/template`.

core/corehttp/gateway_handler.go
```diff
@@ -781,3 +808,28 @@ func preferred404Filename(acceptHeaders []string) (string, string, error) {
...
+// like in bafybeien3m7mdn6imm425vc2s22erzyhbvk5n3ofzgikkhmdkh5cuqbpbq :^))
+func fixupSuperfluousNamespace(w http.ResponseWriter, r *http.Request, urlPath string) bool {
+	if !(strings.HasPrefix(urlPath, "/ipfs/ipfs/") || strings.HasPrefix(urlPath, "/ipfs/ipns/")) {
+		return false // not a superfluous namespace
+	}
+	intendedPath := ipath.New(strings.TrimPrefix(urlPath, "/ipfs"))
+	if err := intendedPath.IsValid(); err != nil {
+		return false // not a valid path
+	}
+	intendedURL := strings.Replace(r.URL.String(), urlPath, intendedPath.String(), 1)
```
Stebalien, line 824: This makes me a little nervous. Can't this be mounted at some `/ipfs` prefix? Probably fine, just a bit funky.

Stebalien: But I'm fine with it as-is.

Stebalien (approved):

core/corehttp/gateway_handler.go
lidel: @Stebalien good catch. Been a while, iirc I did this  to  preserve `?query`  params, but it is not safe to write it into HTML.

Simplified this in https://github.com/ipfs/go-ipfs/pull/7930/commits/a35ffee1364b72e50f66fdd1a03f7a3665d99309  and added query roundtrip to ensure nothing weird ends up in HTML returned to the browser.

It now turns `?q=<script>foobar` to `?q=%3Cscript%3Efoobar`

core/corehttp/gateway_handler.go
```diff
@@ -781,3 +808,33 @@ func preferred404Filename(acceptHeaders []string) (string, string, error) {
 
 	return "", "", fmt.Errorf("there is no 404 file for the requested content types")
 }
+
+// Attempt to fix redundant /ipfs/ namespace as long as resulting
+// 'intended' path is valid.  This is in case gremlins were tickled
+// wrong way and user ended up at /ipfs/ipfs/{cid} or /ipfs/ipns/{id}
+// like in bafybeien3m7mdn6imm425vc2s22erzyhbvk5n3ofzgikkhmdkh5cuqbpbq :^))
+func fixupSuperfluousNamespace(w http.ResponseWriter, urlPath string, urlQuery string) bool {
```
Stebalien, line 816: So... I think we _also_ need to take the `prefix` into account. Users can (and sometimes do) mount the gateway under some other path.

Stebalien (changes requested):
Hm. One more thing.

core/corehttp/gateway_handler.go
lidel: I don't believe the prefix thing adds any net value. Security/complexity price is too high imo.
See my proposal to remove support for X-Ipfs-Gateway-Prefix: https://github.com/ipfs/go-ipfs/issues/7702# – I forgot about it, if you give me :+1:  there I can remove it in a separate PR.

Stebalien: I agree but can we have a bit of a deprecation period here?

1. Do our best to support it here.
2. Print big warnings for the next release cycle with a request to open an issue.
3. Remove it in the release cycle after that.

Stebalien: Alternatively, _here_ we can just say "if prefix is non-empty, print an error and don't provide the nice redirect".

lidel: Brave and other vendors don't rely on prefix, so I added a simple failsafe in https://github.com/ipfs/go-ipfs/pull/7930/commits/763a120ed28c6cd2e77077c6bb0281c583a6b580    (disables fixup when prefix is present). 

This is less code and will be easier to remove after prefix deprecation ends (we will track that in https://github.com/ipfs/go-ipfs/issues/7702#issuecomment-803175828).

Stebalien (approved):

PR #7941, opened 2021-02-25
chore: update deps

Stebalien:
Primarily,

* update semver (go mod support)
* update graphsync & go-ipld-prime
* update golang.org/x packages.


aschmahmann:
@Stebalien it looks like we're still trying to build for `GOOS=darwin GOARCH=386` which is what's causing the build failure here.

aschmahmann:
#7944 is required first

warpfork (approved):
I think this is a case of "tests pass -> lgtm"?

PR #7942, opened 2021-02-25
fix: correctly return pin ls errors

aschmahmann:


Stebalien (approved):

Stebalien:
Is there an easy way to test this? Probably not... but just in case.

aschmahmann:
> Is there an easy way to test this? Probably not... but just in case.

Using sharness, not as far as I can tell (IIRC errors could come from context cancellation or a datastore error). We could potentially start adding Go tests here with some mocking, but not sure how gross that would be.

PR #7943, opened 2021-02-25
feat: remove secio support

Stebalien:
We've had a reliable and enabled by default TLS implementation since 0.4.23 (over a year ago) and turned off SECIO in September of last year. We might as well remove support entirely in the next release and encourage users to upgrade their networks.

Noise is faster, anyways.


aschmahmann (approved):
SGTM, 👋 SECIO.

PR #7944, opened 2021-02-25
chore: bump minimum go version to 1.15

aschmahmann:
bumping minimum version to 1.15 to indicate we are no longer explicitly supporting 1.14, and bumping CI testing to use 1.15.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,7 +33,7 @@ default_environment: &default_environment
 executors:
   golang:
     docker:
-      - image: circleci/golang:1.14.4
+      - image: circleci/golang:1.15.8
     working_directory: ~/ipfs/go-ipfs
     environment:
       <<: *default_environment
@@ -60,7 +60,7 @@ executors:
       E2E_IPFSD_TYPE: go
   dockerizer:
     docker:
-      - image: circleci/golang:1.14.4
+      - image: circleci/golang:1.15.8
     environment:
       IMAGE_NAME: ipfs/go-ipfs
       WIP_IMAGE_TAG: wip
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,7 +33,7 @@ default_environment: &default_environment
 executors:
   golang:
     docker:
-      - image: circleci/golang:1.15.8
+      - image: circleci/golang:1.15.2
     working_directory: ~/ipfs/go-ipfs
     environment:
       <<: *default_environment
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -33,7 +33,7 @@ default_environment: &default_environment
 executors:
   golang:
     docker:
-      - image: circleci/golang:1.14.4
+      - image: circleci/golang:1.15.8
```
aschmahmann, line 36: It's a bit awkward that we track this here, below, and in the orb https://github.com/ipfs/ci-helpers/blob/fb84aaad053f46509acd7264420a57629ec95046/orbs/ci-go/%40root.yml#L26. Does this have to be duplicated?

go.mod
```diff
@@ -111,4 +111,4 @@ require (
 	golang.org/x/sys v0.0.0-20210124154548-22da62e12c0c
 )
 
-go 1.14
+go 1.15
```
aschmahmann, line 114: There's no inherent reason we have to do this I guess, but given we're no longer testing with 1.14 this seems reasonable. WDYT @Stebalien?

.circleci/config.yml
Stebalien: I think we _only_ set it here to override the orb's default. We can update the one in the orb and delete this if you'd like.

go.mod
Stebalien: SGTM!

Stebalien:
Also need to update https://github.com/ipfs/go-ipfs/blob/6527675401e64e8f6c6b030aa1f16c7822f251d2/mk/golang.mk#L2.

Stebalien:
And the readme.

Stebalien:
Looks like the dockerfile also needs to be updated.

Stebalien (commented):
Given that sharness is now running in a VM with ubuntu, we may need to explicitly install go with the correct version.

.circleci/config.yml
gammazero: Looks like you may also want to change line 63 to have the same version of go.

At this point, probably want to update to test with `go1.16.2`

aschmahmann: I switched everything to 1.15.2 since that's what CircleCI gives us for the sharness containers and I figured it'd be good to keep all the testing aligned for now. We should figure out running tests for Go 1.15 and 1.16 (i.e. latest and latest-1)

Stebalien (approved):

Issue #7948, opened 2021-02-26
ipfs update repo tool fs-repo-migration can exe in the same directory


godcong:
System variables must be set every time fs-repo is executed, which is relatively inconvenient.
I added some changes, when fs-repo and ipfs are in the same directory, then it can be executed without setting.

aschmahmann:
The fs-repo-migrations in general is under a rework as is its integration into go-ipfs. Currently you can solve this just fine by putting the fs-repo-migrations binary in your PATH. However, even supporting that behavior seems like it could be a bit of a trap for users in the event they update go-ipfs, but forget to update the migrations binary.

Under what situation would you be manually putting the fs-repo-migration binary in the same place as go-ipfs and expecting go-ipfs to run the binary for you instead of just manually calling the migration binary itself?

godcong:
Because I keep failing when I need to use `fs-repo-migrations`

Stebalien:
Unfortunately, blindly running the repo migrations from the current directory is not an option as it makes it much easier for an attacker to trick a user into running untrusted code.

The options for now are:

1. Set PATH to include the directory where you installed fs-repo-migrations.
2. Let go-ipfs fetch/run fs-repo-migrations for you.
3. Manually run the fs-repo-migrations command.

PR #7949, opened 2021-02-26
add fs-repo work directory support

godcong:
as #7948 i said
this pr only add a work directory support


Stebalien (changes requested):
Using the repo migration from the current directory (whatever it happens to be) is a security issue.

If you need to run the migrations from a non-standard location:

1. Stop your daemon.
2. Run `./fs-repo-migrations`
3. Start your daemon.

Issue #7952, opened 2021-02-27
race condition when running ipfs daemon


marten-seemann:
#### Version information:

```
go-ipfs version: 0.8.0
Repo version: 11
System version: amd64/darwin
Golang version: go1.16
```

#### Description:

This is easy to reproduce: `go build -race cmd/ipfs && cmd/ipfs/ipfs`. The race condition occurs less than 20s after booting the node.

```
WARNING: DATA RACE
Read at 0x00c000776348 by goroutine 309:
  github.com/ipfs/go-ipfs/commands.(*Context).GetNode()
      /Users/marten/src/go/src/github.com/ipfs/go-ipfs/commands/context.go:59 +0x4a
  github.com/ipfs/go-ipfs/commands.(*Context).Context()
      /Users/marten/src/go/src/github.com/ipfs/go-ipfs/commands/context.go:101 +0x55
  main.startPinMFS.func1()
      /Users/marten/src/go/src/github.com/ipfs/go-ipfs/cmd/ipfs/pinmfs.go:75 +0xdb

Previous write at 0x00c000776348 by goroutine 210:
  github.com/ipfs/go-ipfs/commands.(*Context).GetNode()
      /Users/marten/src/go/src/github.com/ipfs/go-ipfs/commands/context.go:63 +0x12f
  github.com/ipfs/go-ipfs/commands.(*Context).Context()
      /Users/marten/src/go/src/github.com/ipfs/go-ipfs/commands/context.go:101 +0x55
  main.pinMFSOnChange()
      /Users/marten/src/go/src/github.com/ipfs/go-ipfs/cmd/ipfs/pinmfs.go:101 +0x168

Goroutine 309 (running) created at:
  main.startPinMFS()
      /Users/marten/src/go/src/github.com/ipfs/go-ipfs/cmd/ipfs/pinmfs.go:67 +0xda
  main.daemonFunc()
      /Users/marten/src/go/src/github.com/ipfs/go-ipfs/cmd/ipfs/daemon.go:443 +0x16a4
  fmt.Fscanf()
      /usr/local/Cellar/go/1.16/libexec/src/fmt/scan.go:143 +0xee
  fmt.Sscanf()
      /usr/local/Cellar/go/1.16/libexec/src/fmt/scan.go:114 +0x191
  github.com/syndtr/goleveldb/leveldb/storage.fsParseName()
[...]

PR #7964, opened 2021-03-04
Fix race condition with node creation

gammazero:
This fixes issue #7952

Concurrent calls to` Context.GetNode` result in a race to create the node on first access.

### DIFF
--- a/commands/context.go
+++ b/commands/context.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"strings"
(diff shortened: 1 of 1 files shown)


gammazero:
This was closed because the context is copied, which copies the mutex.  That is probably not what we want, and it causes the  vet warning:
``` 
assignment copies lock value to cmdctx: github.com/ipfs/go-ipfs/commands.Context contains sync.Mutex
```

Issue #36, opened 2014-09-08
Namespace description


jbenet:
@whyrusleeping  here's a rough spec of ipns / the namespace:

- Make an `NSResolver` interface.

```Go
type NSResolver interface {
  Resolve(name string) (string, error)
}
```
- Implement DNS, proquint `NSResolvers` (which @Bren2010 started).

- Implement a `RoutingResolver` (the meat of IPNS) which does: 

```Go
// RoutingName is the de-serialized name structure that is stored (serialized) 
// in the routing system. Basically, a hash + a digital signature. (serialization can be
// protobuf, or a simple binary format)
type RoutingName {
  Hash        u.Key
  Signature []byte
}

// RoutingResolver implements NSResolver for the main IPFS SFS-like naming
type RoutingResolver {
  routing routing.IpfsRouting
  dag      mdag.DAGService
}

func (r *RoutingResolver) Resolve (name string) (string, error) {

  // name should be a multihash. if it isn't, error out here.

  // use the routing system to get the name.
  // /ipns/<name>
  key := u.Key(name) // or whatever
  val, err := r.routing.Get(key)
  if err != nil {
    return "", err
  }

  // name should be a public key retrievable from ipfs
  // /ipfs/<name>
  node, err :=  r.dag.Get(key)
  if err != nil {
    return "", err
  }
  
  // get PublicKey from node.Data
  pk := PublicKey(node.Data)

  // check sig with pk
  if !pk.Verify(val.Hash, val.Signature) {
    return "", errors.New("Invalid value. Not signed by PrivateKey corresponding to %v", pk)
  }

  // ok sig checks out. this is a valid name.
  return string(val.Hash)
}

```

- implement `NameService` or `NameSystem` object:

```Go
// NameSystem is the object implementing IPNS. It adheres to the `NSResolver` 
// interface, and uses child `NSResolvers`, depending on what the name looks like.
// 
// hashes -> routing, domains -> dns, proquints -> proquint
// we can add `.bit` and `.onion` resolution here in the future
// 
// in non-routing cases, the values returned are hashes to lookup in the routing system.
// (most of the time, if not ALL of the time. still deciding on this). meaning that 
// a dns record: `DNS TXT ipns=<some-hash>` needs to then resolve `<some-hash>`
// via the routing system.
type NameSystem struct {
  routing   RoutingResolver
  dns        DNSResovler
  proquint ProquintResolver
}
```

[...]

Issue #3503, opened 2016-12-13
Extra newline at end of "dag get" result


ianopolous:
#### Version information:
go-ipfs version: 0.4.5-dev-4cb236c
Repo version: 4
System version: amd64/linux
Golang version: go1.7.1

#### Type:
Bug

#### Priority:
P1 (for me)
#### Description:
Using either the cli or http api the dag get command seems to add an extra newline 0x0A on the end of the raw data.

To reproduce using cli:
```bash
echo '{"data":1234}' | ipfs dag put -f cbor
```
then 
```bash
ipfs dag get zdpuAs3whHmb9T1NkHSLGF45ykcKrEBxSLiEx6YpLzmKbQLEB | wc -c
```
should give 13, but gives 14, and the extra byte on the end is a newline

The same JSON through the http api results in the same extra byte. The JSON is not special, similar JSON with more/different elements does the same.

Kubuxu:
JSON isn't a stable format, we have no way of knowing was was the initial format of the json you have passed in.

Also you are passing it in with a new line, but doing `printf "%s" '{"data":1234}'` won't change anything.
The formatting of the output JSON format is arbitrary if you store it as pure data.

ianopolous:
_ipfs dag put --help_ says that the default input format is JSON.

djdv:
I think Kubuxu is saying that `echo` is outputting your string but also a newline, so what you meant to pass in and what is actually being passing in are different. `printf` will print the string and not add a newline. 
![untitled](https://cloud.githubusercontent.com/assets/13862850/21167895/d3f7ca56-c17d-11e6-8513-b74e123c1a66.png)

Kubuxu:
> JSON isn't a stable format, we have no way of knowing was was the initial format of the json you have passed in.

Yes, input format is JSON but storage format is CBOR, which is stable format.
There are infinitely many ways you can store same data in JSON (`{"data":1234}`, `{"data":1234 }`, `{"data": 1234}`, and so on) but from the usability standpoint it is more important that the same information has the same address.

If you want to store the white-space information, use plain `ipfs add` which doesn't interpret data your are passing it.

ianopolous:
@djdv I tried the printf suggestion, but, as Jakub predicted, it gives the same result. 

[...]

PR #7976, opened 2021-03-12
IPLD Prime In IPFS: Target Merge Branch

hannahhoward:
This feature branch updates a number of dependencies to directly make use of go-ipld-prime under the hood. The major exception is the unixfs subsystem, which remains the same via the [go-ipld-legacy]() shim.

Before this branch is feature complete, a known / remaining sub-conversation that is ongoing is "How should we directly expose the newer ipld-prime node interface to programs making use of IPFS?". This conversation is occurring in https://github.com/ipfs/go-ipfs/pull/8046 which will merge into this branch when ready.

* [x] https://github.com/ipfs/go-ipfs/pull/8046
* [x] https://github.com/ipfs/go-ipfs/pull/8028
* [x] https://github.com/ipfs/go-path/pull/36
    * [x] Address Remaining PR comments
    * [x] @warpfork approval
    * [x] @aschmahmann approval
* [x] https://github.com/ipfs/go-ipfs/pull/7995
   * [x] https://github.com/ipfs/go-ipfs/pull/8257
      * [x] https://github.com/ipld/go-ipld-prime/pull/204 ?
      * [x] https://github.com/ipld/go-ipld-prime/pull/205 ?
      * [x] fix any remaining sharness tests
* [x] https://github.com/ipfs/go-ipfs-provider/pull/32
   * [x] Rebase on  master
   * [x] @warpfork review
   * [x] @warpfork approval
   * [x] @aschmahmann review
   * [x] @aschmahmann approval
* [x] https://github.com/ipfs/interface-go-ipfs-core/pull/67
   * [x] Merge https://github.com/ipfs/interface-go-ipfs-core/pull/68 -- since its already referenced by this branch
   * [x] Update to tagged go-dagwriter
   * [x] Rebase on master
   * [x] Walk this one back a bit.  Dep bumps only for this round -- no new API support promises.
   * [x] @warpfork review
   * [x] @warpfork approval
   * [x] @aschmahmann approval? (already discussed changes)
* [x] https://github.com/ipfs/go-merkledag/pull/67
   * [x] Rebase on master
   * [x] @warpfork review
   * [x] @warpfork approval
   * [x] @aschmahmann  re-review? (not clear if a full review conducted but there are comments)
   * [x] @aschmahmann approval 
* [x] make tagged versions of sub-repos and propagate
Do this all at once, on a mergeparty day.
For each repo: look at the go mod file; update any dep not pointing to a clean tag; wait for CI green; create tag in that repo; move to next repo.
Order of merge/tag should be:
	* [x] https://github.com/ipld/go-ipld-prime (@master) tagged v0.11.0
	* [x] https://github.com/multiformats/go-multicodec (@master) tagged v0.3.0
	* [x] https://github.com/ipfs/go-ipld-legacy (@master/deps updated) tagged v0.1.0
	* [x] https://github.com/ipld/go-codec-dagpb (@master/deps updated) tagged v1.3.0
	* [x] https://github.com/ipfs/go-merkledag (`#67` merged) tagged v0.4.0
	* [x] https://github.com/ipfs/go-fetcher (@master/deps updated) tagged v1.4.0
	* [x] https://github.com/ipfs/go-unixfsnode (@master/deps updated) tagged v1.1.2 (may have no changes, just explicit dep prop for clarity)
	* [x] https://github.com/ipfs/go-path (`#36` merged) tagged v0.1.0
	* [x] https://github.com/ipfs/interface-go-ipfs-core (`#67` merged) tagged v0.5.0
	* [x] https://github.com/ipfs/go-ipfs-provider (`#32` merged) tagged v0.6.1
	* [x] https://github.com/ipfs/go-ipld-git (`#46` merged) tagged v0.1.0
	* [x] https://github.com/ipfs/go-ipfs-api/pull/248 (no tag needed, master is pulled by CI)
	* ... now back to self, this PR!
* [x] rebase on master
* [ ] @aschmahmann review of this branch for merge
* [ ] @aschmahmann approval of this branch
* [ ] merge


test/sharness/t0280-plugin-git.sh
```diff
@@ -20,16 +20,16 @@ test_dag_git() {
...
+  test_expect_failure "path traversals work" '
     echo \"YmxvYiA3ACcsLnB5Zgo=\" > file1 &&
     ipfs dag get baf4bcfhzi72pcj5cc4ocz7igcduubuu7aa3cddi/object/parents/0/tree/dir2/hash/f3/hash > out1
   '
 
-  test_expect_success "outputs look correct" '
+  test_expect_failure "outputs look correct" '
     test_cmp file1 out1
   '
 }
```
Stebalien, line 35: I assume we're going to revert this?

fuse/readonly/readonly_unix.go
```diff
@@ -65,20 +66,41 @@ func (s *Root) Lookup(ctx context.Context, name string) (fs.Node, error) {
...
+	var fnd ipld.Node
+	switch cidLnk.Cid.Prefix().Codec {
+	case cid.DagProtobuf:
+		fnd, err = mdag.ProtoNodeConverter(blk, nd)
+	case cid.Raw:
+		fnd, err = mdag.RawNodeConverter(blk, nd)
 	default:
 		log.Error("fuse node was not a protobuf node")
-		return nil, fuse.ENOTSUP
+		return nil, fuse.ENOENT
```
warpfork, line 96: Why does this error code change?

test/sharness/t0053-dag.sh
```diff
@@ -133,48 +179,47 @@ test_dag_cmd() {
   '
 
   test_expect_success "output looks correct" '
-    echo "{\"data\":\"CAISBGZvbwoYBA==\",\"links\":[]}" > cat_exp &&
+    echo -n "{\"Data\":{\"/\":{\"bytes\":\"CAISBGZvbwoYBA==\"}},\"Links\":[]}" > cat_exp &&
```
warpfork, line 182: Did we intend to change capitalization here or is this accidental?

This seems like it may need a changelog entry.  I can't say for sure how much people might use whatever API this is, but I suspect it might be nonzero?

aschmahmann (changes requested):
Not sure if this was totally ready for review yet, but figured I'd try and get a bit ahead of the ball here on the review😄

.circleci/config.yml
```diff
@@ -229,7 +229,7 @@ jobs:
       - run:
           name: Cloning
           command: |
-            git clone https://github.com/ipfs/go-ipfs-api.git
+            git clone -b ipld-in-ipfs https://github.com/ipfs/go-ipfs-api.git
```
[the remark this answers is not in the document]
willscott: good point. this is reflecting the changed codec names for the `dag put` subcommand. Opened the change there as https://github.com/ipfs/go-ipfs-api/pull/248

There's no go.mod or versioning there, but we'll i guess merge that PR at the same time, and remove this branch before merging?

plugin/plugins/git/git.go
```diff
@@ -33,43 +31,19 @@ func (*gitPlugin) Version() string {
 }
 
 func (*gitPlugin) Init(_ *plugin.Environment) error {
+	// register a custom identifier in the reserved range for import of "zlib-encoded git objects."
+	// TODO: give this a name.
+	multicodec.RegisterDecoder(uint64(mc.ReservedStart+mc.GitRaw), decodeZlibGit)
```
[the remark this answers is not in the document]
rvagg: I think so, because this plugin is "zlib-encoded git object" rather than just "git object", and only the git multicodec is registered in its `init`, this one gets a special codec in the reserved multicodec range.

test/sharness/t0053-dag.sh
```diff
@@ -25,31 +25,38 @@ test_expect_success "make an ipld object in json" '
...
-    IPLDHASH=$(cat ipld_object | ipfs dag put)
+  test_expect_success "can add an ipld object using protobuf and --cid=base=base32" '
+    IPLDHASHb32=$(cat ipld_object | ipfs dag put --input-enc=dag-json -f dag-cbor --cid-base=base32)
   '
 
-  test_expect_success "output looks correct" '
-    EXPHASH="bafyreidjtjfmavdk7epvztob2m5vlm3pxp3gjmpyewro4qlbw5n4f4iz64"
-    test $EXPHASH = $IPLDHASH
+  test_expect_success "output looks correct (does not upgrade to CIDv1)" '
+    test $EXPHASH = $IPLDHASHb32
```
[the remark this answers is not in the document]
rvagg: ah .. yes, so the default now as per line 51 is base32, and this has been changed without updating the message, same with the previous one which mentions protobuf. :facepalm:. I'll get my head back in here and sort out again what these were supposed to do and make the messages match, probably next week.

test/sharness/t0053-dag.sh
```diff
@@ -81,20 +115,32 @@ test_dag_cmd() {
...
+    echo -n "{\"beep\":[0,\"bop\"],\"dict\":\"ionary\"}" > sub2_exp &&
     test_cmp sub2_exp sub2 &&
-    echo "[0,\"bop\"]" > sub3_exp &&
+    echo -n "[0,\"bop\"]" > sub3_exp &&
     test_cmp sub3_exp sub3 &&
-    echo "0" > sub4_exp &&
+    echo -n "0" > sub4_exp &&
     test_cmp sub4_exp sub4 &&
-    echo "\"bop\"" > sub5_exp &&
+    echo -n "\"bop\"" > sub5_exp &&
```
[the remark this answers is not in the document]
rvagg: yes, we removed the newline for all `dag get` output

test/sharness/t0053-dag.sh
rvagg: Yes, changed to match the data model form for DAG-PB: https://ipld.io/specs/codecs/dag-pb/spec/#logical-format

This is now consistent across Go & JS (long-standing problem) and matches the pb schema too. And also DAG-JSON with its bytes format making a showing here too now—although this test has padding it in and shouldn't since we've fixed that @ https://github.com/ipld/go-ipld-prime/pull/216 so I'll have to do another pass here.

test/sharness/t0053-dag.sh
```diff
@@ -133,48 +179,47 @@ test_dag_cmd() {
...
 
-  test_expect_success "can add an ipld object with sha3 hash" '
-    IPLDHASH=$(cat ipld_object | ipfs dag put --hash sha3)
+  test_expect_success "can add an ipld object with sha3-512 hash" '
+    IPLDHASH=$(cat ipld_object | ipfs dag put --hash sha3-512)
   '
 
   test_expect_success "output looks correct" '
-    EXPHASH="bafyriqgae54zjl3bjebmbat2rjem4ewj6vni6jxohandmvk3bibfgv3sioyeidppsghvulryxats43br3b7afa6jy77x6gqzqaicer6ljicck"
+    EXPHASH="bafyriqforjma7y7akqz7nhuu73r6liggj5zhkbfiqgicywe3fgkna2ijlhod2af3ue7doj56tlzt5hh6iu5esafc4msr3sd53jol5m2o25ucy"
```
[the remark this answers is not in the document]
rvagg: sorry, explained over in the origin PR for this https://github.com/ipfs/go-ipfs/pull/7995/files#r677096234, further explained by the `Size` changes below (a fix in what the cbor encoder is doing)

test/sharness/t0053-dag.sh
```diff
@@ -271,10 +316,10 @@ test_dag_cmd() {
 
   test_expect_success "dag stat of simple IPLD object" '
     ipfs dag stat $NESTED_HASH > actual_stat_inner_ipld_obj &&
-    echo "Size: 15, NumBlocks: 1" > exp_stat_inner_ipld_obj &&
+    echo "Size: 8, NumBlocks: 1" > exp_stat_inner_ipld_obj &&
     test_cmp exp_stat_inner_ipld_obj actual_stat_inner_ipld_obj &&
     ipfs dag stat $HASH > actual_stat_ipld_obj &&
-    echo "Size: 61, NumBlocks: 2" > exp_stat_ipld_obj &&
+    echo "Size: 54, NumBlocks: 2" > exp_stat_ipld_obj &&
```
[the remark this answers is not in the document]
rvagg: https://github.com/ipfs/go-ipfs/pull/7995/files#r677098266
Current impl is doing a float64 for `123`, which I assume comes from the naive json parsing that it used to have. Now we assume DAG-JSON rules with numbers being ints unless they have a decimal point.

test/sharness/t0053-dag.sh
```diff
@@ -25,31 +25,38 @@ test_expect_success "make an ipld object in json" '
   printf "{\"hello\":\"world\",\"cats\":[{\"/\":\"%s\"},{\"water\":{\"/\":\"%s\"}}],\"magic\":{\"/\":\"%s\"},\"sub\":{\"dict\":\"ionary\",\"beep\":[0,\"bop\"]}}" $HASH1 $HASH2 $HASH3 > ipld_object
 '
 
+test_expect_success "make the same ipld object in dag-cbor, dag-json and dag-pb" '
+  echo "omREYXRhT6JkRGF0YWF4ZUxpbmtzgGVMaW5rc4A=" | base64 -d > ipld_object_dagcbor
+  echo "Cg+iZERhdGFheGVMaW5rc4A=" | base64 -d > ipld_object_dagpb
+  echo "{\"Data\":{\"/\":{\"bytes\":\"omREYXRhYXhlTGlua3OA\"}},\"Links\":[]}" > ipld_object_dagjson
```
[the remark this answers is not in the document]
rvagg: Yeah, in this file all the hashes come from me as I messed around the most in here. I generated (most of) them using the JS implementations as a double-check, I didn't take this new version of go-ipfs' word for it and bake it into fixtures.

rvagg:
Why can't I path into a plain DAG-PB block? `ipfs dag get ipld/QmWLnXhYWtku9zjbN7otfm77Fob2og7Q4Ha6FYqakNDVxj/Data`? I can't seem to do this even with the current go-ipfs (using `/data`, `/links`, or `/Data`, `/Links`) but I would have expected to be able to do it now that we're presenting a pretty standard DAG-PB codec and the same pathing works with other codecs. Is it because of the automatic UnixFS lens applied to block loading or something in the pathing / resolver?

willscott:
3 things came out of discussion with @aschmahmann today:

* @rvagg i think the reason you can't do it now is that the nodereifier used for pathing is doing the unixfs semantics on the node when pathing.
@aschmahmann asks if we can look to see if we can add back a `/ipld/` at the beginning of a path as a signal to work with a link system that doesn't have the reifier as a way to support this additional feature.

* @rvagg - your changes to codecs should mean the change to https://github.com/ipfs/go-ipfs-api is no longer needed - that we should be back to dag-cbor as the default codec returned from a dag put command. We should try circle back on master pull of go-ipfs-api and confirm.

* We should write release notes noting breaking changes

willscott:
Draft section of changelog / release notes:

---

IPLD Levels Up

The interpretation of data has been largely migrated from the previous ipld-format library to the newer ipld-prime construction.
This allows for interaction with newer IPLD-prime based tooling, additional codec support, and some minor performance improvements.
This is significant refactor of a core component that touches many parts of IPFS, and does come with some breaking changes:

* IPLD plugins (those implementing the `PluginIPLD` interface) are not registered into IPLD prime. Such plugins should instead register their encoder and decoder functions in their initialization functions against the multicodec table. We demonstrate the change with the bundled git plugin.
* The semantics of `dag put` and `dag get` change.
  * `dag get` takes the `format` flag which accepts a multicodec for the output format to print
  * `dag put` uses `input-enc` to specify the multicodec format data is currently in, and `format` for the multicodec format it should be stored in.
  * The flag names remain the same, but uses of a previously limited set of explicitly named formats have now been standardized, which results in changes to some identifiers. In particular, what was previously specified as `json` should now be specified as `dag-json`, to indicate that IPLD links should be accepted and parsed.

rvagg:
I'd also add under that "`dag put` and `dag get`" section:

* Previous json output using `dag get` is now formal dag-json as specified at https://ipld.io/specs/codecs/dag-json/, in particular this change will be noticed with the encoding of bytes which are now represented as `{"/":{"bytes":"base64bytes"}}`
* When exported with another codec via `dag get` (such as dag-json) dag-pb is now represented as specified at https://ipld.io/specs/codecs/dag-pb/, in particular the change in case may be noticed: `{"Data":..., "Links":[...]}`.

warpfork:
> When exported with another codec via dag get (such as dag-json) dag-pb is now represented as specified at https://ipld.io/specs/codecs/dag-pb/, in particular the change in case may be noticed: {"Data":..., "Links":[...]}.

Is anything else changing, or just the capitalization?

Did we intend to change this capitalization?  (I have to confess, I would've flagged the TitleCase in earlier rounds of review, except I actually presumed it was something were doing _because_ some legacy somewhere had already baked us into capitalizing it.)

test/sharness/t0053-dag.sh
```diff
@@ -81,25 +194,46 @@ test_dag_cmd() {
...
+    ipfs pin add $IPLDPBHASH
+  '
+
+  test_expect_success "can pin dag-cbor object" '
+    ipfs pin add $IPLDCBORHASH
+  '
+
+  test_expect_success "can pin dag-json object" '
+    ipfs pin add $IPLDJSONHASH
   '
```
rvagg, line 232: @willscott @aschmahmann so ... I've done a bunch of cleanup in 0053 to make it make sense and also test some additional stuff (including the `/ipld/` pathing in dag-pb that's now been enabled). But pinning is now no longer working for these last 3, although the first one works and I have no explanation for it. Just above this pinning section you'll see the `dag get /ipld/$IPLDPBHASH/Links/0/Name` working correctly, yet when we `ipfs pin add $IPLDPBHASH` it doesn't work. But it's also not working for the dag-cbor or dag-json forms either but it is working for `$IPLDHASH` which is the `{"hello":"world"...}` object.

`$IPLDCBORHASH`, `$IPLDJSONHASH` and `$IPLDPBHASH` are now references to this data: https://github.com/ipld/codec-fixtures/tree/master/fixtures/dagpb_2link%2Bdata - in 0053 we're taking the same binary form in each of these codecs, feeding it in and asking for it to be stored as each of the other 3 codecs. So 3x3 `dag add`s and checking the resulting CIDs, and it's all good and we can even `dag get` them. But they won't `dag pin`... It's `merkledag: not found`.

Any thoughts on where to look?

rvagg: oh 🤦 I've introduced an incomplete dag, so pin isn't happy about not having the blocks referenced, my bad, now to fix but keep enough of the test functionality intact.

rvagg:
Fixed those tests up, 0043 should be pretty good now and includes a `/ipld` pathing test as well as a the new inter-codec tests although they're less sophisticated than I wanted. But I'd like to add in more complexity using codec-fixtures in the future; it's not strictly required for this branch.

@warpfork and I have been discussing this `data`->`Data` & `links`->`Links` thing and doing a little bit of (PL-archaeology in the process) and how this change could probably do with some more documentation. We're moving from a dag-pb-is-magic land to a place where it's _less_ blessed (the auto-unixfs stuff still blesses it a bit) and we get access to a pure data model form of the block rather than some special cased functionality. We've had this in our dag-pb spec, maybe since we wrote it, it's just been added to over time as we've understood the situation better: https://ipld.io/specs/codecs/dag-pb/spec/#alternative-legacy-pathing - and with this change, the world is more unified, all our codecs and verticals are doing the same thing with dag-pb. _But_ it's going to be a change for users and their expectations if they have any process of interacting with raw dag-pb through the JSON output, or, I believe, the named-link pathing functionality as mentioned in that spec. So we need to work out what used to work, what still does work, and what's changed wrt to data model and traversal of dag-pb blocks—both casing of `Data` and `Links` and also this named-link stuff.

test/sharness/t0053-dag.sh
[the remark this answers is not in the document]
rvagg: resolved now, 0053 should be clean and pretty clear about what it's doing

willscott:
> So we need to work out what used to work, what still does work, and what's changed 

I would say that if we document what the behavior is now, that should be sufficient.
Our understanding is that the not-well-formalized or defined previous edge cases of the dag semantics were not widely exercised, and we agreed it was okay for their semantics to change.

BigLep:
2021-08-10 discussion: next step here is to generate "release notes".  Specifically this means:
We need a table that enumerates the before and after of the relevant pieces.  @aschmahmann will provide more details.  

Goal after 2021-08-11 synchronous time is comment with checkboxes with remaining items.

core/node/core.go
```diff
@@ -82,6 +90,24 @@ func (s *syncDagService) Session(ctx context.Context) format.NodeGetter {
...
+		}
+		return basicnode.Prototype.Any, nil
+	})
+	return fc
+}
+
+// Resolver returns a resolver that's configured to look up unixfs paths
+func Resolver(fetcherFactory fetcher.Factory) *resolver.Resolver {
+	return resolver.NewBasicResolver(fetcherFactory)
+}
```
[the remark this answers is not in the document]
aschmahmann: We're going to deal with this in a follow up issue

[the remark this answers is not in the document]
BigLep: Do we have the followup issue?

rvagg:
Now we have the sharness from https://github.com/ipfs/go-ipfs/issues/3503 (t0055-dag-put-json-new-line.sh) it's failing and needed fixing so I've done that. It now doesn't have the expected failure but I also had to update the command in there which reminds me that there's another item that needs to go onto the changelog:

* The default `input-enc` for `dag put` is now `dag-json` and its default `format` is now `dag-cbor`. These were `json` and `cbor` respectively. However, with this update, `cbor` is no longer an alias for `dag-cbor`, but rather a plain CBOR format (which will not encode CIDs and does not have strict deterministic encoding rules) and `json` is not an alias for `dag-json` but rather a plain JSON format (which also won't encode CIDs and will encode bytes in the Go-specific base64 string format rather than the dag-json method). See https://ipld.io/specs/codecs/ for more information on IPLD codecs.

BigLep:
> Now we have the sharness from #3503 (t0055-dag-put-json-new-line.sh) it's failing and needed fixing so I've done that. It now doesn't have the expected failure but I also had to update the command in there which reminds me that there's another item that needs to go onto the changelog

Do we have a place for the in-progress changelog?  If we need something, https://github.com/ipfs/go-ipfs/issues/8176 could be used.

rvagg:
@biglep [above](https://github.com/ipfs/go-ipfs/pull/7976#discussion_r683663733), @aschmahmann suggested we do a 0.10 TBD in CHANGELOG.md so I'm working on that now and I'll integrate a shortish summary of the implications of going with option 1 as described in https://github.com/ipld/ipld/pull/128 too.

CHANGELOG.md
```diff
@@ -1,5 +1,27 @@
...
+  * Using `dag put` and a `format` option of `dag-pb` now requires that the input conform to this dag-pb specified form. Previously, input using `{"data":"...","links":[...]}` was accepted, now it must be `{"Data":"...","Links":[...]}`.
+  * Previously it was not possible to use paths to navigate to any of these properties of a dag-pb node, the only possible paths were named links, e.g. `dag get QmFoo/NamedLink` where `NamedLink` was one of the links whose name was `NamedLink`. This functionality remains the same, but by prefixing the path with `/ipld/` we enter data model pathing semantics and can `dag get /ipld/QmFoo/Links/0/Hash` to navigate to links or `/ipld/QmFoo/Data` to simply retrieve the data section of the node, for example.
```
aschmahmann, line 21: > e.g. `dag get QmFoo/NamedLink` where `NamedLink` was one of the links whose name was `NamedLink`. This functionality remains the same

This is true in many cases, but IMO is not something we want to guarantee. If we happen to make the UnixFS reifier stricter such that `/ipfs/QmNonUnixFSDagPbObject/LinkName` no longer resolved (and you needed to go the `/Links/0` route) that doesn't seem to scary to me (i.e. we don't need to write lots of path resolution tests for non-UnixFS DagPB objects to verify `/ipfs/QmNonUnixFSDagPbObject/LinkName` still works).

@rvagg WDYT, is it safer to not mention the fate `/ipfs/QmNonUnixFSDagPbObject/LinkName` here?

ribasushi: > `/ipfs/QmNonUnixFSDagPbObject/LinkName`

☝️ I believe this is literally what @gozala's niftysave meta-wrapper does, that is: you have users in he wild.

/cc @mikeal ( I think Gozala is OOO )

go.mod
```diff
@@ -17,16 +17,18 @@ require (
...
 	github.com/ipfs/go-ds-badger v0.2.7
 	github.com/ipfs/go-ds-flatfs v0.4.5
 	github.com/ipfs/go-ds-leveldb v0.4.2
 	github.com/ipfs/go-ds-measure v0.1.0
+	github.com/ipfs/go-fetcher v1.3.1-0.20210811164155-5ca55a2bea5e
 	github.com/ipfs/go-filestore v0.0.3
 	github.com/ipfs/go-fs-lock v0.0.7
 	github.com/ipfs/go-graphsync v0.8.0
-	github.com/ipfs/go-ipfs-blockstore v0.1.6
+	github.com/ipfs/go-ipfs-blockstore v0.1.4
```
aschmahmann, line 31: We shouldn't be downgrading the blockstore

willscott: bumped back to where it was

CHANGELOG.md
aschmahmann: @ribasushi IIRC that uses DAG-CBOR not DagPB so is both not Qm and also not a problem here.

ribasushi: Ah I misread the `NonUnixFSDagPb` part.
You are correct, my comment does not apply

test/sharness/t0055-dag-put-json-new-line.sh
```diff
@@ -8,14 +8,14 @@ test_init_ipfs
 
 test_expect_success 'create test JSON files' '
   WANT_JSON="{\"data\":1234}"
-  WANT_HASH="bafyreidcqah6v3py3ujdc3ris22rjlfntaw7ajrus2f2477kpnizulaoea"
+  WANT_HASH="bafyreidbm2zncsc3j25zn7lofgd4woeh6eygdy73thfosuni2rwr3bhcvu"
```
aschmahmann, line 11: Is this due to int/float business too?

CHANGELOG.md
```diff
@@ -1,5 +1,27 @@
...
 
+## v0.10.0 TBD
+
+**IPLD Levels Up**
+
+The handling of data serialization as well as many aspects of DAG traversal and pathing have been migrated from older libraries, including [go-merkledag](https://github.com/ipfs/go-merkledag) and [go-ipld-format](https://github.com/ipfs/go-ipld-format) to the new **[go-ipld-prime](https://github.com/ipld/go-ipld-prime)** library and its components. This allows us to use many of the newer tools afforded by go-ipld-prime, stricter and more uniform codec implementations, support for additional (pluggable) codecs, and some minor performance improvements.
+
+This is significant refactor of a core component that touches many parts of IPFS, and does come with some **breaking changes**:
+
+* **IPLD plugins** (those implementing the `PluginIPLD` interface) are not registered into go-ipld-prime. Such plugins should instead register their encoder and decoder functions in their initialization functions against the multicodec table. We demonstrate the change with the [bundled git plugin](./plugin/plugins/git/).
```
aschmahmann, line 11: There is also a breaking change to the `PluginIPLD` interface, IPLD plugins need to be updated to use go-ipld-prime.

core/node/provider.go
```diff
@@ -172,7 +172,12 @@ func SimpleProviders(reprovideStrategy string, reprovideInterval string) fx.Opti
...
 func pinnedProviderStrategy(onlyRoots bool) interface{} {
-	return func(pinner pin.Pinner, dag ipld.DAGService) simple.KeyChanFunc {
-		return simple.NewPinnedProvider(onlyRoots, pinner, dag)
+	type input struct {
+		fx.In
+		Pinner      pin.Pinner
+		IPLDFetcher fetcher.Factory `name:"ipldFetcher"`
+	}
+	return func(in input) simple.KeyChanFunc {
+		return simple.NewPinnedProvider(onlyRoots, in.Pinner, in.IPLDFetcher)
```
aschmahmann, line 181: We are getting panics here:

```
panic: invalid key for map dagpb.PBNode: ".": no such field

goroutine 29177 [running]:
github.com/ipld/go-ipld-prime/fluent.(*mapNodeAssembler).AssembleEntry(0xc00476e290, 0x2d2b5c8, 0x1, 0x0, 0x0)
	C:/Users/adin/go/pkg/mod/github.com/ipld/go-ipld-prime@v0.11.0/fluent/fluentBuilder.go:168 +0x1d1
github.com/ipld/go-ipld-prime/traversal/selector/builder.(*selectorSpecBuilder).Matcher.func1(0x2d74cf8, 0xc00476e290)
	C:/Users/adin/go/pkg/mod/github.com/ipld/go-ipld-prime@v0.11.0/traversal/selector/builder/builder.go:156 +0x4f
github.com/ipld/go-ipld-prime/fluent.(*nodeAssembler).CreateMap(0xc00476e280, 0x1, 0x2cae3e0)
	C:/Users/adin/go/pkg/mod/github.com/ipld/go-ipld-prime@v0.11.0/fluent/fluentBuilder.go:95 +0x151
github.com/ipld/go-ipld-prime/fluent.MustBuildMap.func1(0x2d856b0, 0xc00476e280)
	C:/Users/adin/go/pkg/mod/github.com/ipld/go-ipld-prime@v0.11.0/fluent/fluentBuilder.go:31 +0x5b
github.com/ipld/go-ipld-prime/fluent.MustBuild(0x2d50660, 0x395cff0, 0xc00569fa40, 0x0, 0x0)
	C:/Users/adin/go/pkg/mod/github.com/ipld/go-ipld-prime@v0.11.0/fluent/fluentBuilder.go:27 +0xc6
github.com/ipld/go-ipld-prime/fluent.MustBuildMap(0x2d50660, 0x395cff0, 0x1, 0x2cae3e0, 0x0, 0x0)
	C:/Users/adin/go/pkg/mod/github.com/ipld/go-ipld-prime@v0.11.0/fluent/fluentBuilder.go:31 +0x94
github.com/ipld/go-ipld-prime/traversal/selector/builder.(*selectorSpecBuilder).Matcher(0xc00476e270, 0x0, 0x0)
	C:/Users/adin/go/pkg/mod/github.com/ipld/go-ipld-prime@v0.11.0/traversal/selector/builder/builder.go:155 +0x6b
github.com/ipfs/go-fetcher/helpers.BlockAllOfType(0x2d712d0, 0xc00946a300, 0x2d715e0, 0xc0094e3140, 0x2d63b78, 0xc00476e260, 0x2d50660, 0x395cff0, 0xc00663afe0, 0x0, ...)
	C:/Users/adin/go/pkg/mod/github.com/ipfs/go-fetcher@v1.4.0/helpers/traversal.go:46 +0xb1
github.com/ipfs/go-fetcher/helpers.BlockAll(0x2d712d0, 0xc00946a300, 0x2d715e0, 0xc0094e3140, 0x2d63b78, 0xc00476e260, 0xc00663afe0, 0x0, 0x0)
	C:/Users/adin/go/pkg/mod/github.com/ipfs/go-fetcher@v1.4.0/helpers/traversal.go:38 +0x194
github.com/ipfs/go-ipfs-provider/simple.pinSet.func1(0x2d712d0, 0xc00300aa00, 0xc008eb4250, 0x26ff2574258, 0xc00300c100, 0x2d4fe20, 0xc003000460, 0x26ff2679f00)
	C:/Users/adin/go/pkg/mod/github.com/ipfs/go-ipfs-provider@v0.6.0/simple/reprovide.go:239 +0x9a5
created by github.com/ipfs/go-ipfs-provider/simple.pinSet
	C:/Users/adin/go/pkg/mod/github.com/ipfs/go-ipfs-provider@v0.6.0/simple/reprovide.go:215 +0xbd
```

It looks like there's a bug in the fetcher helpers https://github.com/ipfs/go-fetcher/blob/ef6ef50eaf5c3968d412d87980e3ff45a7e3e6c2/helpers/traversal.go#L31-L50.

I'm not really familiar with this area of the codebase, but it looks like we're turning the selector into an IPLD node before executing it and that https://github.com/ipfs/go-fetcher/blob/ef6ef50eaf5c3968d412d87980e3ff45a7e3e6c2/helpers/traversal.go#L34 is just wrong and it needs to be a `basicnode.Prototype.Any`.

`BlockAllOfType` seems like it should just disappear as a function since it seems like the idea was that it could walk a graph if all the nodes were the same type (a bit sketchy as is), but instead the only thing the passed in `NodePrototype` does is tell selector how it should be serialized.

I'll start a new PR and reference it here.

aschmahmann: https://github.com/ipfs/go-fetcher/pull/24

test/sharness/t0055-dag-put-json-new-line.sh
rvagg: yes

old:

```
a1                                                # map(1)
  64                                              #   string(4)
    64617461                                      #     "data"
  fb 4093480000000000                             #   float(1234)
```

new:

```
a1                                                # map(1)
  64                                              #   string(4)
    64617461                                      #     "data"
  19 04d2                                         #   uint(1234)
```

CHANGELOG.md
rvagg: @aschmahmann I think this is your call since you have a much clearer picture of the direction of IPFS. We have a long-standing note about this ["alternative pathing"](https://ipld.io/specs/codecs/dag-pb/spec/#alternative-legacy-pathing) in the dag-pb spec and I've always just considered it a legacy hold-over, like so many of the dag-pb decisions, that we're living with. I don't find `QmFoo/NamedLink` particularly offensive as a legacy convenience holdover but it really fits in the ADL category in this new world. Having `PathedPBNode` sit in between seem like a reasonable compromise and even if `UnixFSPBNode` evolves to be much smarter and stricter, go-unixfsnode could still retain this fallaback to `PathedPBNode` that will _just_ do named pathing. But you're the one with a better imagination for what future evolution might involve here. I've just seen notes throughout go-unixfsnode alluding to that suggestion and I'm not sure what that really means.

This is the bit that's doing the fallback now, perhaps it just stays and even if `DecodeUnixFSData()` starts rejecting more things we still get the fallback to `PathedPBNode` (which is the `defaultReifier()`)? https://github.com/ipfs/go-unixfsnode/blob/f6007e84b8c4d0d910a11c5380c3e06eeef06403/reification.go#L25-L29

Your call.

core/node/provider.go
aschmahmann: resolved by the above PR

CHANGELOG.md
aschmahmann: Thanks @rvagg. It's a tough call, but I think for now we'll leave it as you suggested (i.e. just telling people about the existence of data model pathing and not explicitly dissuading the use of `PathedPBNode` pathing). 

Hopefully there aren't too many people using non-UnixFS DagPB right now and relying on this functionality, but we can push the discussion of if there are any downsides to leaving `PathedPBNode` pathing as a fallback for `/ipfs` for another day. In the meanwhile we're not breaking anyone so there's not much to say here.

core/commands/dag/get.go
```diff
@@ -25,14 +38,34 @@ func dagGet(req *cmds.Request, res cmds.ResponseEmitter, env cmds.Environment) e
...
+	encoder, err := multicodec.LookupEncoder(uint64(fCodec))
+	if err != nil {
+		return fmt.Errorf("invalid encoding: %s - %s", format, err)
+	}
+
+	r, w := io.Pipe()
+	go func() {
+		defer w.Close()
+		if err := encoder(finalNode, w); err != nil {
+			_ = res.CloseWithError(err)
```
aschmahmann, line 66: This hangs if there is an error (e.g. `ipfs dag get bafysomedagcbor --format=git-raw` or if we have `{ "foo" : "bar"}`  `ipfs dag get bafyroot/foo --format=raw`).

aschmahmann: I think using `w.CloseWithError` will fix this.

aschmahmann: Seems fixed.

BigLep:
2021-08-17 update on outstanding items:

@aschmahmann: create issues for the following to fix during the release phase:
1. Missing test for hang when doing dag get: https://github.com/ipfs/go-ipfs/pull/7976/commits/9339e826c6514ab4dd5351d409964a0548efc22d
2. Bugs related to git codec (EOF error are getting ignored by CLI)
3. Bug where can only get raw bytes using the git codec but not round trip it.  It should not do the first or also do the second.

aschmahmann (approved):
Merging as this seems good enough for go-ipfs v0.10.0-rc1. There are a few outstanding issues linked above that we'll need to deal with before the final release.

Thanks everyone who helped push this over the finish line 🙇‍♂️, this was a huge and much needed endeavor to start upgrading our IPLD library usage.

PR #7991, opened 2021-03-18
doc: document security fix policy

Stebalien:
1. All releases may contain unannounced security fixes.
2. By default, security fixes will not be backported.

### DIFF
--- a/docs/releases.md
+++ b/docs/releases.md
@@ -83,6 +83,14 @@ Patch releases will usually follow a compressed release cycle and should take 2-
 
 Some patch releases, especially ones fixing one or more complex bugs, may undergo the full release process.
 
+## Security Fix Policy
+
+Any release may contain security fixes. Unless the fix addresses a bug being exploited in the wild, the fix will _not_ be called out in the release notes. Please make sure to update ASAP.
+
+By policy, the team will usually wait until about 2 weeks after the final release to announce any fixed security issues. However, depending on the impact and ease of discovery of the issue, the team may wait more or less time. It is important to always update to the latest version ASAP and file issues if you're unable to update for some reason.
+
+Finally, unless a security issue is actively being exploited or a significant number of users are unable to update to the latest version (e.g., due to a difficult migration, breaking changes, etc.), security fixes will _not_ be backported to previous releases.
+
 ## Performing a Release
 
 The release is managed by the `Lead Maintainer` for `go-ipfs`. It starts with the opening of an issue containing the content available on the [RELEASE_ISSUE_TEMPLATE](./RELEASE_ISSUE_TEMPLATE.md) not more than **48 hours** after the previous release.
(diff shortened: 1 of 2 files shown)


docs/releases.md
```diff
@@ -83,6 +83,14 @@ Patch releases will usually follow a compressed release cycle and should take 2-
 
 Some patch releases, especially ones fixing one or more complex bugs, may undergo the full release process.
 
+## Security Fix Policy
+
+Any release may contain security fixes. Unless the fix addresses a bug being exploited in the wild, the fix will _not_ be called out in the release notes. Please make sure to update ASAP.
+
+By policy, the team will usually wait until about 2 weeks after the final release to announce any fixed security issues. However, depending on the impact and ease of discovery of the issue, the team may wait more or less time. It is important to always update to the latest version ASAP and file issues if you're unable to update for some reason.
+
+Finally, unless a security issue is actively being exploited or a significant number of users are unable to update to the latest version (e.g., due to a difficult migration, breaking changes, etc.), security fixes will _not_ be backported to previous releases.
```
Stebalien, line 92: We should probably expand on this. This comes down to:

1. Patch releases take a lot of time/effort.
2. We fix _lots_ of bugs that might theoretically be security fixes but it would take a significant amount of investigation to determine the potential impact. We _really_ don't want anyone to continue to use older go-ipfs releases assuming that "everything's good" because security fixes get backported.
3. We _definitely_ can't backport security fixes immediately because it would effectively turn the issue into a zero-day.

docs/releases.md
```diff
@@ -83,6 +83,14 @@ Patch releases will usually follow a compressed release cycle and should take 2-
 
 Some patch releases, especially ones fixing one or more complex bugs, may undergo the full release process.
 
+## Security Fix Policy
+
+Any release may contain security fixes. Unless the fix addresses a bug being exploited in the wild, the fix will _not_ be called out in the release notes. Please make sure to update ASAP.
```
schomatis, line 88: This should at least be encoded as a minor version update, I don't know how realistic it is to offload the responsibility of monitoring and updating to any release.

docs/releases.md
jacobheun: This just ties to our existing maintenance policy though, yes? Although it doesn't look like we have this written down anywhere. We only support the latest minor. We may, in the future support multiple versions, but that is a ways out. Any back porting is an exception to the rule, which I think this change already clarifies.

docs/releases.md
jacobheun: A mailing list may be a better approach here for notification than relying on versioning. We could then link the opt in list in the security readme. I don't want to block this on that decision though.

docs/releases.md
Stebalien: Yeah, writing down our normal policy would be good too. But I do want to call this out so nobody thinks they're fine as long as they run the latest patch release.

docs/releases.md
Stebalien: Patch releases may also include super-critical security releases, although I guess those will usually be announced (or really well hidden but that's difficult).

jacobheun (approved):
There are some additions we can make in the future but this is an improvement on the existing policy declaration so giving this a 👍

PR #7994, opened 2021-03-20
chore: deprecate Gateway.PathPrefixes

lidel:
Context: https://github.com/ipfs/go-ipfs/issues/7702 and @Stebalien comments there 

This PR deprecates `Gateway.PathPrefixes` and `X-Ipfs-Gateway-Prefix` header:
- marks `Gateway.PathPrefixes` as deprecated in `docs/config.md`
-  logs error during startup if `Gateway.PathPrefixes` is used  (not empty)
   - I went with log because it prints red ERROR that makes it hard to miss (but lmk if should be regular stderr instead):
     >

### DIFF
--- a/cmd/ipfs/daemon.go
+++ b/cmd/ipfs/daemon.go
@@ -677,6 +677,10 @@ func serveHTTPGateway(req *cmds.Request, cctx *oldcmds.Context) (<-chan error, e
 		opts = append(opts, corehttp.RedirectOption("", cfg.Gateway.RootRedirect))
 	}
 
+	if len(cfg.Gateway.PathPrefixes) > 0 {
+		log.Error("Support for X-Ipfs-Gateway-Prefix and Gateway.PathPrefixes is deprecated and will be removed in the next release. Please comment on the issue if you're using this feature: https://github.com/ipfs/go-ipfs/issues/7702")
+	}
+
 	node, err := cctx.ConstructNode()
 	if err != nil {
 		return nil, fmt.Errorf("serveHTTPGateway: ConstructNode() failed: %s", err)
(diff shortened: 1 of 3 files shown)


cmd/ipfs/daemon.go
```diff
@@ -319,9 +319,9 @@ func daemonFunc(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment
 
 	// Start assembling node config
 	ncfg := &core.BuildCfg{
-		Repo:                        repo,
-		Permanent:                   true, // It is temporary way to signify that node is permanent
-		Online:                      !offline,
+		Repo:      repo,
+		Permanent: true, // It is temporary way to signify that node is permanent
+		Online:    !offline,
```
Stebalien, line 324: You may need to update your go (or gofmt) version. This shouldn't change.

cmd/ipfs/daemon.go
```diff
@@ -677,6 +677,10 @@ func serveHTTPGateway(req *cmds.Request, cctx *oldcmds.Context) (<-chan error, e
 		opts = append(opts, corehttp.RedirectOption("", cfg.Gateway.RootRedirect))
 	}
 
+	if len(cfg.Gateway.PathPrefixes) > 0 {
+		log.Error("Support for X-Ipfs-Gateway-Prefix and Gateway.PathPrefixes is deprecated and will be removed in the next release: https://github.com/ipfs/go-ipfs/issues/7702")
```
Stebalien, line 681: Maybe a note about "please comment on the issue if you're using this feature"? That way we can ask how and try to come up with an alternative/upgrade path without causing too much trouble. I don't think anyone actually _is_ using this feature, but still.

Stebalien (approved):

PR #8001, opened 2021-03-22
Fix: innacuracies in MFS command documentation.

hsanjuan:
These also end up in API docs.


Stebalien (approved):

woss:
for reference here is the topic that triggered this https://discuss.ipfs.io/t/is-it-possible-to-share-the-mfs-from-one-node-to-another/10513/32

Issue #7938, opened 2021-02-24
Pinning progress


achingbrain:
#### Version information:

```
$ ipfs --version --all
go-ipfs version: 0.8.0
Repo version: 11
System version: amd64/darwin
Golang version: go1.15.8
```

#### Description:

I'm looking at exposing the `progress` arg to `ipfs.pin.addAll` for `ipfs-http-client` and am trying to nail down the behaviour of go-IPFS.

If I add a directory recursively I see progress:

```console
$ ipfs add -r --pin=false .
added QmdAtxgJwAaK41woVuM4tq9qr6NPw8cC212TajNAot7gqV ipfs/CHANGELOG.md
added QmdaGY5JWEVPDR2WoMYETnhqJvbCgCZTvocjF1YqXnJMHt ipfs/CODE_OF_CONDUCT.md
added QmVFQ52asX4zgY6Sbp7tmDe8Vs6AemgiLfj6nkRuRrJkuC ipfs/CONTRIBUTING.md
added QmfNPdCxwKXQxo1fKjhCggaPLA9rfG4uJSZuuPWgrSSnnb ipfs/COPYRIGHT
added QmdcxcVZw8Pojj1rhTCSp4uoNTV2fY8JwpBb3Bj5XNDMpd ipfs/LICENSE-APACHE
added QmRcNnUwgBVjBeJLiBV2uSi8CmFdSvoHj62kWrCZRBA749 ipfs/LICENSE-MIT
... more files here
added QmbgNwtUFsPkrnz9QYwwLWXKP1u1SrmMqUqpats3hvtHkn ipfs/node_modules
added QmYdcvCbpKGq2bT7Y35cnhGDAmFdULQ99bcQRre7C3iFDL ipfs/src
added QmX74v4CMyrGWEjA8tKeQKFJDXowEJ5Zfzjbdo2dnMAXot ipfs/test/utils
added Qma4PYzbxzYkUq36g5eZi8QdNVC7Rf6dP9MxgtgFNRMQcx ipfs/test
added QmdLWTWW9cRQJuiyeAXDw9ZMgcaSdu9SRPEw2M43wMpReU ipfs
 175.16 KiB / 175.16 KiB [=======================================================================] 100.00%
```

If I then pin that directory with the progress CLI arg, I see no progress:

```console
$ ipfs pin add -r --progress=true QmdLWTWW9cRQJuiyeAXDw9ZMgcaSdu9SRPEw2M43wMpReU
[...]

PR #8023, opened 2021-03-25
fix: omit empty pins slice when reporting pin progress

Stebalien:
Otherwise, we get messages of the form `{Pins: null, Progress: 100}`. After this change, we'll get `{Progress: 100}`.

### DIFF
--- a/core/commands/pin/pin.go
+++ b/core/commands/pin/pin.go
@@ -44,8 +44,8 @@ type PinOutput struct {
 }
 
 type AddPinOutput struct {
-	Pins     []string
-	Progress int `json:",omitempty"`
+	Pins     []string `json:",omitempty"`
+	Progress int      `json:",omitempty"`
 }
 
 const (


Stebalien:
@achingbrain does this change have any impact on the HTTP API as far as js-ipfs is concerned?

achingbrain:
No, js-IPFS doesn't report pinning progress yet, I was waiting for https://github.com/ipfs/go-ipfs/issues/7938 to determine what sort of behaviour should be implemented.

core/commands/pin/pin.go
```diff
@@ -44,8 +44,8 @@ type PinOutput struct {
 }
 
 type AddPinOutput struct {
-	Pins     []string
-	Progress int `json:",omitempty"`
+	Pins     []string `json:",omitempty"`
```
aschmahmann, line 47: Is the idea here that while the progress is being reported there's no reason to report the empty pins?

aschmahmann (approved):
Left a question, but since it works for @achingbrain in js-ipfs LGTM

core/commands/pin/pin.go
Stebalien: Yes. Otherwise, it could look like nothing was pinned.

PR #8031, opened 2021-03-27
config.md: fix typos/improve wording

RubenKelevra:


docs/config.md
```diff
@@ -35,7 +35,7 @@ documented in `ipfs config profile --help`.
 - `local-discovery`
 
   Sets default values to fields affected by the server
-  profile, enables discovery in local networks.
+  profile enables discovery in local networks.
```
Stebalien, line 38: Enables local discovery (enabled by default). Useful to re-enable local discovery after it's disabled by another profile (e.g., the `server` profile).

docs/config.md
```diff
@@ -601,7 +600,7 @@ Examples:
 
 #### `Gateway.PublicGateways: Paths`
 
-Array of paths that should be exposed on the hostname.
+An array of paths that are exposed on the hostname.
```
Stebalien, line 603: I would stick with "should be". That is, you're telling IPFS what to do, not what "is".

Stebalien: That or "An array of paths that ipfs should expose".

Stebalien (commented):
Only two small nits.

Otherwise, thanks for cleaning up the docs!

Issue #7552, opened 2020-07-19
support the accept content type header set to "application/json" for ipns subdomain requests


lucidNTR:
Followup to #7431, that was closed but had the valuable feature of accessing json manifest of folder

requests to a "folder", especially the root of a folder published to ipns such as http://bafzbeibxww7y6v5metkgmxeqkvls74uqvptbhommrb5b7xianigcicdvuy.ipns.localhost/
should be served the same response as
https://ipfs.io/api/v0/ls?arg=bafzbeibxww7y6v5metkgmxeqkvls74uqvptbhommrb5b7xianigcicdvuy

IF the accept content type header is set to "application/json" that way there is no interference with existing gateway behaviour and with serving the index.html file at the root

lidel:
Thank you for filling this @lucidNTR – below are my thoughts why we should consider this:

### On mutability of directory listings 

Directory listing is a special type of response returned by the gateway in that it is generated by the gateway itself. Even for immutable `/ipfs/`, the response can change, because we may update the HTML/CSS. 

If we add support for JSON response, then we could truly return an immutable response, instead of less  permanent `Etag`-based cache control hint.

### Needs: mandatory versioning 

The only concern is that doing plain `application/json` would leak the current schema of `/api/v0/ls` into Gateway outside `/api/v0`. What happens when we change the API and introduce `/api/v1/ls` that produces different JSON?

I think we need to include API version in the request somehow:
- `Accept: application/json;version=0`
- `Accept: application/vnd.ipfs.v0+json`

so we are able to return clear error when client asks for API response in a version that is no longer supported. Having API version as a part of request would also let us return proper cache control headers that mark response as immutable under `/ipfs/*`

### TLDR: this would be useful feature

Personally, I would be open to a PR that adds this feature, as long versioning is mandatory.
Thoughts @aschmahmann?

aschmahmann:
@lidel I'm generally fine if we want to basically alias something we already do (e.g. `/api/v0/ls`) using accept headers.

A couple thoughts + questions though:
1. It doesn't seem like it's worth thinking too hard about caching in browsers since if you really wanted caching you could just use a local ipfs daemon.
2. I'm not very familiar with accept headers, if we were to remove this functionality in the future (or a particular gateway wanted to disable it) what would the result look like?

lidel:
1. Avoiding unnecessary localhost requests still saves some resources, but I agree its diminishing returns in practice :)
2. A regular HTML directory listing would be returned.

[...]

PR #8037, opened 2021-03-30
feat: serve CBOR encoded DAG nodes from the gateway

alanshaw:
This PR enables the gateway to serve CBOR encoded DAG nodes. They are returned as JSON objects, the same encoding used on the CLI.

Closes https://github.com/ipfs/go-ipfs/issues/8823 and https://github.com/ipfs/in-web-browsers/issues/182

### DIFF
--- a/core/corehttp/gateway_handler.go
+++ b/core/corehttp/gateway_handler.go
@@ -322,7 +322,7 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
 			modtime = time.Unix(1, 0)
 		}
 
-		setContentDispositionHeader(r.URL.Query(), w.Header())
+		maybeSetContentDispositionHeader(r.URL.Query(), w.Header())
 		name := r.URL.Query().Get("filename")
 		if name == "" {
 			name = getFilename(urlPath)
@@ -530,7 +530,7 @@ func (i *gatewayHandler) serveFile(w http.ResponseWriter, req *http.Request, nam
 func (i *gatewayHandler) serveCBOR(w http.ResponseWriter, r *http.Request, n format.Node) {
 	w.Header().Set("Cache-Control", "public, max-age=29030400, immutable")
 	w.Header().Set("Content-Type", "application/json")
-	setContentDispositionHeader(r.URL.Query(), w.Header())
+	maybeSetContentDispositionHeader(r.URL.Query(), w.Header())
 
 	name := r.URL.Query().Get("filename")
 	if name == "" {
@@ -862,9 +862,9 @@ func fixupSuperfluousNamespace(w http.ResponseWriter, urlPath string, urlQuery s
 	}) == nil
 }
 
-// setContentDispositionHeader sets the Content-Disposition header if a
+// maybeSetContentDispositionHeader sets the Content-Disposition header if a
 // "filename" was included in the URL querystring.
-func setContentDispositionHeader(qs url.Values, h http.Header) {
+func maybeSetContentDispositionHeader(qs url.Values, h http.Header) {
 	urlFilename := qs.Get("filename")
 	if urlFilename == "" {
 		return
--- a/core/corehttp/gateway_handler.go
+++ b/core/corehttp/gateway_handler.go
@@ -1,7 +1,9 @@
 package corehttp
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
 	"html/template"
 	"io"
@@ -21,6 +23,7 @@ import (
 	"github.com/ipfs/go-cid"
 	files "github.com/ipfs/go-ipfs-files"
 	assets "github.com/ipfs/go-ipfs/assets"
+	format "github.com/ipfs/go-ipld-format"
 	dag "github.com/ipfs/go-merkledag"
 	mfs "github.com/ipfs/go-mfs"
 	path "github.com/ipfs/go-path"
@@ -264,6 +267,16 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
 		return
 	}
 
+	if resolvedPath.Cid().Prefix().Codec == cid.DagCBOR {
+		n, err := i.api.Dag().Get(r.Context(), resolvedPath.Cid())
+		if err != nil {
+			webError(w, "ipfs dag get "+escapedURLPath, err, http.StatusNotFound)
+			return
+		}
+		i.serveCBOR(w, r, n)
+		return
+	}
+
 	dr, err := i.api.Unixfs().Get(r.Context(), resolvedPath)
 	if err != nil {
 		webError(w, "ipfs cat "+escapedURLPath, err, http.StatusNotFound)
@@ -309,18 +322,9 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
 			modtime = time.Unix(1, 0)
 		}
 
-		urlFilename := r.URL.Query().Get("filename")
-		var name string
-		if urlFilename != "" {
-			disposition := "inline"
-			if r.URL.Query().Get("download") == "true" {
-				disposition = "attachment"
-			}
-			utf8Name := url.PathEscape(urlFilename)
-			asciiName := url.PathEscape(onlyAscii.ReplaceAllLiteralString(urlFilename, "_"))
-			w.Header().Set("Content-Disposition", fmt.Sprintf("%s; filename=\"%s\"; filename*=UTF-8''%s", disposition, asciiName, utf8Name))
-			name = urlFilename
-		} else {
+		setContentDispositionHeader(r.URL.Query(), w.Header())
+		name := r.URL.Query().Get("filename")
+		if name == "" {
 			name = getFilename(urlPath)
 		}
 		i.serveFile(w, r, name, modtime, f)
@@ -521,6 +525,25 @@ func (i *gatewayHandler) serveFile(w http.ResponseWriter, req *http.Request, nam
 
 	w = &statusResponseWriter{w}
 	http.ServeContent(w, req, name, modtime, content)
+}
+
+func (i *gatewayHandler) serveCBOR(w http.ResponseWriter, r *http.Request, n format.Node) {
+	w.Header().Set("Cache-Control", "public, max-age=29030400, immutable")
+	w.Header().Set("Content-Type", "application/json")
+	setContentDispositionHeader(r.URL.Query(), w.Header())
+
+	name := r.URL.Query().Get("filename")
+	if name == "" {
+		name = getFilename(r.URL.Path)
+	}
+	modtime := time.Unix(1, 0)
+
+	b, err := json.Marshal(n)
+	if err != nil {
+		internalWebError(w, err)
+		return
+	}
+	http.ServeContent(w, r, name, modtime, bytes.NewReader(b))
 }
 
 func (i *gatewayHandler) servePretty404IfPresent(w http.ResponseWriter, r *http.Request, parsedPath ipath.Path) bool {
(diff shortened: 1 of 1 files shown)


ribasushi (approved):
LGTM with one suggestion. Good to merge as-is.

core/corehttp/gateway_handler.go
```diff
@@ -523,6 +527,25 @@ func (i *gatewayHandler) serveFile(w http.ResponseWriter, req *http.Request, nam
 	http.ServeContent(w, req, name, modtime, content)
 }
 
+func (i *gatewayHandler) serveCBOR(w http.ResponseWriter, r *http.Request, n format.Node) {
+	w.Header().Set("Cache-Control", "public, max-age=29030400, immutable")
+	w.Header().Set("Content-Type", "application/json")
```
lidel, line 532: Not sure about this. We need to make a decision on default behavior before this ships:
-  @warpfork suggested we should aim at returning more user-friendly HTML by default,  similar to explore.ipld.io, for all non-unixfs DAGs. 
   - I kinda agree: we already return custom HTML for directory listings (and not JSON)
   - We want gateways to support everything, not just dag-cbor
- What if for dag-cbor we return JSON only when request had an explicit `Content-Type: application/json` header?
  - This way makes semantic sense (I explicitly ask for JSON representation of dag-cbor) and won't break if we add HTML explorer for IPLD universe as the default response
  - For now, if content-type is missing from request, return error saying that "To preview dag-cbor via gateway, request with `Content-type: application/json` header"

core/corehttp/gateway_handler.go
```diff
@@ -523,6 +527,25 @@ func (i *gatewayHandler) serveFile(w http.ResponseWriter, req *http.Request, nam
 	http.ServeContent(w, req, name, modtime, content)
 }
 
+func (i *gatewayHandler) serveCBOR(w http.ResponseWriter, r *http.Request, n format.Node) {
+	w.Header().Set("Cache-Control", "public, max-age=29030400, immutable")
```
lidel, line 531: This is invalid if DAG is fetched from anything other than `/ipfs/` like  `/ipns/` (not immutable). See [gateway_handler.go#L318-L319](https://github.com/ipfs/go-ipfs/blob/ea2bb34a7ad0704077180936f1ec3fe7ec25c6d2/core/corehttp/gateway_handler.go#L318-L319). 
This may require a bigger refactor, as we want to avoid separate code paths for unixfs and cbor.

lidel (changes requested):
I am supportive for this, with some caveats:
- CBOR is not JSON, so the request should explicitly ask for JSON  (details inline)
- Cache-Control needs more nuance (details inline)
- Needs sharness tests similar to `test/sharness/t0115-gateway-dir-listing.sh`

Note that we want to support CAR export too: https://github.com/ipfs/in-web-browsers/issues/170

core/corehttp/gateway_handler.go
Gozala: @mikeal pointed out that

> there’s a lot of caches out there that aren’t smart about this and you’ll get responses from cache that don’t match the accept header

This is really unfortunate, but likely relevant concern.

I was thinking maybe gateway code could consider file extension if there is one and infer Content-Type from there ? E.g. if content targeted is `ipfs://${cid}/metadata.json` in dag-cbor format it's probably reasonable to assume that content type `application/json` should be used. In fact that seems like a reasonable approach to dag-pb as well.

Although I realize this is getting uncomfortable.

alanshaw: I'd be _ok_ with streaming the bytes as `application/cbor` and add a querystring parameter `?json` or `?encoding=json` to have it converted to JSON with a `application/json` header.

IMHO we should be serving data from the gateway that is by default the easiest format to consume from a web application. JSON is that. Also, right now we have NO support, and this is significantly better than nothing and what's here _is_ consistent with the CLI.

So I'd prefer to keep what's happening here and maybe add `?encoding=raw` for getting the raw cbor bytes.

I like @gozala's idea but it feels like too many special cases.

core/corehttp/gateway_handler.go
alanshaw: Good catch.

core/corehttp/gateway_handler.go
lidel: > you’ll get responses from cache that don’t match the accept header

lidel: We really should keep complexity at bay and avoid special-casing different DAG types too much.
I propose we  introduce DAG CBOR support in (IMO) the least controversial way:

- Stream `application/cbor` by default
  - Returning the original, space-efficient format makes more sense than JSON
    - Rationale: if someone wants JSON by default, then they should use `dag-json`
  - IIUC [dag-cbor](https://github.com/ipld/specs/blob/master/block-layer/codecs/dag-cbor.md) is a subset of cbor, so this is a valid content-type (I'd like to avoid introducing proprietary one if we don't have to). 
-  JSON requires conversion step, so we should support  `application/json` as an explicit opt-in:
   - requesting `Content-Type: application/json`
     - Rationale: standards matter and we can't be sloppy if we want to be respected by serious vendors and standards bodies
   - passing `?enc=json` parameter (as a fallback way for use in poorly written software)
     - Rationale:  the same parameter could be re-used for returning JSON directory listings in https://github.com/ipfs/go-ipfs/issues/7552, keeping our conventions neat.

mikeal: Just require the header and the querystring. The problem you’ll run into with it only being a fallback is that the user, client developer, and the meddling proxy/cache are different actors and can’t always coordinate whether or not to explicitly opt into a feature.

mikeal: Good point on the address bar. It would actually have inconsistent behavior based on your browser plugins because everyone with JSON viewer plugin includes the content type in browser requests.

lidel: Let's continue in https://github.com/ipfs/in-web-browsers/issues/182

lidel (commented):
After thinking about this some more, and playing with this PR, I feel @alanshaw  convinced me that defaulting `dag-cbor` response to  `application/json`  contributes to a way better onboarding flow for newcomers and also better interop (more tools speak JSON than CBOR). 

I created https://github.com/ipfs/in-web-browsers/issues/182 to discuss the overall strategy when it comes to IPLD on gateways, the default behavior is one of open questions there – would like to hear more voices before  we commit to this.

Would appreciate comments there, so we have an artifact of the decision outside this PR.

alanshaw:
I've come to realise a possible reason for CBOR (or other encoding) not being implemented yet is pathing within a node - typically you can `/ipfs/{cid}/internal/path` to get the value `138` out of a node like `{ internal: { path: 138 } }`  - this PR does not take this into account and I'm not sure how it would work right now if we allow returning raw cbor (I guess re-encode `138` in CBOR?).

lidel:
@alanshaw  After sleeping on this I think we should not return JSON by default, but instead return an error page informing user to retry request for  JSON or CBOR  either via `content-type` header or `?enc=` param. 

That ensures devs build on deterministic response types and enable us to introduce user-friendly UI as suggested in https://github.com/ipfs/in-web-browsers/issues/182#issuecomment-815039662

warpfork:
>  I think we should not return JSON by default, but instead return an error informing user to retry request with explicit format (JSON or CBOR) either via content-type header or ?enc= param.
> 
> That ensures devs will build on deterministic response types and enable us to introduce user-friendly UI for exploring DAGs in the future

Big agree, :100: .  I think we should leave room for that kind of experience to be built into the Gateway.  It would be consistent with what we do for unixfsv1 directories already and we should leave room to carry that design logic further; and for the encoding types, explicit is good.

anorth:
Let me join the chorus suggesting against converting objects to JSON by default.

Despite it being almost the only case right now, unixfs is a special case of IPLD data that can be interpreted by a gateway. When a client asks the gateway for some CID and that CID is a unixfs root, the gateway doesn't give the client what they actually asked for (an IPLD node), but infers a request for the DAG data to be traversed and rendered to a flat file (which does not have the same hash). Or analogous behaviour for a directory. This is great for the use case of URLs in hyperlinks, rendering in browsers etc. We don't ask the client to pass headers or query parameters for this "special" rendering behaviour because it's too awkward for the browser URL bar case; so the default is to render.

Imagine some node that the gateway currently doesn't know how to render, but in the future might be told how to render as something a browser can handle. We should preserve a similar default approach of assuming that the request is from a browser URL bar. Maybe some third party has an alternative schema for chunking flat files, and we later add support for it too. We don't want to change the response data and break old clients relying on the default JSON.

Most other contexts outside a browser URL bar have easy capacity to pass parameters describing what content they want (JS code in browsers, server-side code, `curl` invocations, ...). It seems a bit backwards that "give me the raw data with this CID" is not the no-parameter default, but the browser case determines that. So IMO all the other response types should be explicit, controlled by parameters from the client. Maybe we even need an `?enc=` (blank value) to request no transformation, in case the caller doesn't know (or want to work out from CID) what the native encoding is. 

When serving IPLD data, another reason not to JSONify everything is that it breaks the verifiability of the content from the CID. It's ok for that to be an explicit request, but unfortunate as a default behaviour. JS code can parse CBOR too. 

> IMHO we should be serving data from the gateway that is by default the easiest format to consume from a web application.

I would tweak this. We should be serving data from the gateway that is by default the easiest format to consume from a web *browser*, because the URL bar is the least good place to require parameters describing what the client wants, and is already set by precedent. But web *applications* are just as capable as other clients of requesting a particular data format – the request and response will both usually be handled by code.

willscott:
This might become much easier once we have https://github.com/ipfs/go-ipfs/pull/7995 which will allow for a `dag get` to use arbitrary ipld-prime codec encodings.  cc @aschmahmann for recommendation on if this implies the cmds route for encoding vs as enumerated encoders on the specific subcommands

core/corehttp/gateway_handler.go
```diff
@@ -538,12 +537,23 @@ func (i *gatewayHandler) serveCBOR(w http.ResponseWriter, r *http.Request, n for
 	}
 	modtime := time.Unix(1, 0)
 
-	b, err := json.Marshal(n)
-	if err != nil {
-		internalWebError(w, err)
-		return
+	var contentType string
+	var data []byte
+	if r.URL.Query().Get("enc") == "json" || r.Header.Get("Content-Type") == "application/json" {
```
jeremy-im, line 542: I think this should be: `r.Header.Get("Accept")`
That header in the request is about the body of the request.

jeremy-im: ... oh, except "Accept" can be a comma-separated list, so the header value needs to be parsed - not just `==` compared.

jeremy-im: this seems to work pretty well:

```
diff --git a/core/corehttp/gateway_handler.go b/core/corehttp/gateway_handler.go
index 11e5db671..5b847ae10 100644
--- a/core/corehttp/gateway_handler.go
+++ b/core/corehttp/gateway_handler.go
@@ -39,6 +39,7 @@ const (
 )
 
 var onlyAscii = regexp.MustCompile("[[:^ascii:]]")
+var anyJSON = regexp.MustCompile(`^application/(|.+\+)json$`)
 
 // HTML-based redirect for errors which can be recovered from, but we want
 // to provide hint to people that they should fix things on their end.
@@ -537,9 +538,24 @@ func (i *gatewayHandler) serveCBOR(w http.ResponseWriter, r *http.Request, n for
 	}
 	modtime := time.Unix(1, 0)
 
+	useJSON := false
+	if r.URL.Query().Get("enc") == "json" {
+		useJSON = true
+	} else {
+		for _, v := range r.Header.Values("Accept") {
+			for _, a := range strings.Split(v, ",") {
+				mediatype, _, err := mime.ParseMediaType(a)
+				if err == nil && anyJSON.MatchString(mediatype) {
+					useJSON = true
+					break
+				}
+			}
+		}
+	}
+
 	var contentType string
 	var data []byte
-	if r.URL.Query().Get("enc") == "json" || r.Header.Get("Content-Type") == "application/json" {
+	if useJSON {
 		contentType = "application/json"
 		b, err := json.Marshal(n)
 		if err != nil {
```

jeremy-im: d'oh! : forgot the label on the outer loop so the break can break all the way out.

lidel:
Quick check-in notes: 
- this needs to be refactored to use `?format=` proposed in
  https://github.com/ipfs/go-ipfs/issues/8234
  - may be easier to tackle `block` `car` `cbor` and `json` use cases in a single PR to ensure DRY code around HTTP headers      (either reuse this one, or open a new one)

lidel:
@BigLep yes, this is unblocked, I've  created https://github.com/ipfs/go-ipfs/issues/8823  with some additional notes, so it is not blocked on my availability.

Aiming at 0.14 sounds good: ideally we would resolve https://github.com/ipfs/go-ipfs/pull/8568 before this ships, as people will start using dag-json and dag-cbor way more thanks to this, and we want good UX around it, but in this specific case it is not a hard blocker.

Implementation wise, it should be pretty easy to add on top of refactor that happened in https://github.com/ipfs/go-ipfs/pull/8758:

https://github.com/ipfs/go-ipfs/blob/5b43672a8733525ad9fab9be03d14d983f3f9914/core/corehttp/gateway_handler.go#L307-L326 

Anyone with spare time could do it, given a spare day. 
Writing tests will take most of the time here, but we could reuse test fixtures from:
- https://ipld.io/specs/codecs/dag-pb/fixtures/cross-codec/
- https://ipld.io/specs/codecs/dag-json/fixtures/cross-codec/
- https://ipld.io/specs/codecs/dag-cbor/fixtures/cross-codec/

lidel:
Continued in:
- https://github.com/ipfs/kubo/pull/9335
  - https://github.com/ipfs/in-web-browsers/issues/201
  - https://github.com/ipfs/in-web-browsers/issues/202

PR #8040, opened 2021-03-31
fix: set systemd startup timeout to infinity

### LINKED ISSUE #6741: large repo + diskUsage calc timesout systemd unit
#### Version information:
go-ipfs version: 0.4.22-
Repo version: 7
System version: amd64/linux
Golang version: go1.12.9
Please check dist.ipfs.io for a newer version of go-ipfs and update if necessary. Report back if the problem persists.

#### Description:
I have a very large repo (10TB+ on XFS). Sometimes (I don't know the interval), ipfs-daemon decides it needs to refresh the blocks/diskUsage.cache file *on startup*, and takes so long computing the new usage, that systemd thinks the service has failed to start and kills the process. How can I disable the recalculation or make it run only after the daemon has started up, in a background thread?

### LINKED ISSUE #7269: Long running migrations can cause systemd to terminate daemon on start

### LINKED ISSUE #8034: Very long startup with badgerds
I noticed that go-ipfs will take quite a while to startup when using badgerds. By default the systemd service files allow for 2 minute startup times. I've configured this to be 15 minutes and the service has a hard time completing this in time.

[I've recommended a while ago](https://github.com/ipfs/go-ipfs/pull/7265#pullrequestreview-404174496) to extend the startup time since badgerds takes a while sometimes to start-up - as well as the `--migration` step might be interrupted when stopping the service after 2 minutes.

I don't think it's a good practice to provide a systemd file that might interrupt the `--migration` after an update, as well as cleanup operations on badgerds, which leaves the service needing manual interventions.

Sending systemd a signal like 'startup completed' on the other hand might break some other services which wait for the service to be ready for network operations, so we can't just send a `ready` signal and continue fiddling around with the database...

The right solution would be to extend the startup timeout as well as the shutdown timeout by sending `EXTEND_TIMEOUT_USEC=…` to systemd when a database operation is still working on the operation. This ensures systemd isn't killing the daemon prematurely.

Maybe we should add a notification while this signal is sent to the log, to make sure the administrator is aware of what is going on.

I've captured such a startup's stack trace with a `kill -SIGABRT` after 222 seconds runtime:

[long_startup.log](https://github.com/ipfs/go-ipfs/files/6222130/long_startup.log)

#### Version information:
go-ipfs version: 0.9.0-dev-3f9c3f455
Repo version: 11
System version: amd64/linux
Golang version: go1.16.2

Stebalien:
fixes #7269
fixes #8034
fixes #6741

### DIFF
--- a/misc/systemd/ipfs-hardened.service
+++ b/misc/systemd/ipfs-hardened.service
@@ -58,6 +58,13 @@ CapabilityBoundingSet=CAP_NET_BIND_SERVICE
 #don't use swap
 MemorySwapMax=0
 
+# Don't timeout on startup. Opening the IPFS repo can take a long time in some cases (e.g., when
+# badger is recovering) and migrations can delay startup.
+#
+# Ideally, we'd be a bit smarter about this but there's no good way to do that without hooking
+# systemd dependencies deeper into go-ipfs.
+TimeoutStartSec=infinity
+
 Type=notify
 User=ipfs
 Group=ipfs
--- a/misc/systemd/ipfs.service
+++ b/misc/systemd/ipfs.service
@@ -29,6 +29,13 @@ After=network.target
 #don't use swap
 MemorySwapMax=0
 
+# Don't timeout on startup. Opening the IPFS repo can take a long time in some cases (e.g., when
+# badger is recovering) and migrations can delay startup.
+#
+# Ideally, we'd be a bit smarter about this but there's no good way to do that without hooking
+# systemd dependencies deeper into go-ipfs.
+TimeoutStartSec=infinity
+
 Type=notify
 User=ipfs
 Group=ipfs


Stebalien:
cc @RubenKelevra

RubenKelevra:
LGTM

PR #8046, opened 2021-04-02
Node API

hannahhoward:
# Goals

Provide a direct interface to IPLD prime based tools for reading and writing DAGs
implement https://github.com/ipfs/interface-go-ipfs-core/pull/68

# Implementation

Implement the node API by just combining go-fetcher.FetcherConfig + go-dagwriter.DagWritingService

# For Discussion

I went ahead and augmented the fetcher with unixfs functionality -- should we do this?


willscott:
@aschmahmann notes that for interfaces added to the `interfaces-go-ipfs-core`,  we'd like for the `go-ipfs-http-client` to also be able to implement that interface. e.g.: https://github.com/ipfs/go-ipfs-http-client/blob/3af36210f80fb86aae50da582b494ceddd64c3de/dag.go

We need to think about whether there's a sane way to plum the exposed interface over the http interface

PR #8053, opened 2021-04-06
Update migration sharness tests for new migrations

gammazero:
With the new migrations, go-ipfs no longer uses `fs-repo-migrations` to do repo migrations, and was downloading real migration binaries from the network and running them.  This caused failure, but was not caught because the test was expecting `ipfs daemon --migrate` to fail for other reasons.

This PR fixes the migration tests by creating the appropriate fake migration binaries in the `PATH` so that those are run instead of downloading the real ones.  This also fixes a test that was previously marked broken.

### DIFF
--- a/test/sharness/t0066-migration.sh
+++ b/test/sharness/t0066-migration.sh
@@ -57,7 +57,6 @@ test_expect_success "output looks good" '
   grep "Success: fs-repo migrated to version 11" true_out > /dev/null
 '
 
-# Daemon exits with 1 after the prompt gets "n" when asking to migrate. 
 test_expect_success "'ipfs daemon' prompts to auto migrate" '
   test_expect_code 1 ipfs daemon > daemon_out 2> daemon_err
 '
--- a/test/sharness/t0066-migration.sh
+++ b/test/sharness/t0066-migration.sh
@@ -59,7 +59,7 @@ test_expect_success "output looks good" '
 
 # Daemon exits with 1 after the prompt gets "n" when asking to migrate. 
 test_expect_success "'ipfs daemon' prompts to auto migrate" '
-  test_expect_code 1 bash -c "echo n | ipfs daemon > daemon_out 2> daemon_err"
+  test_expect_code 1 ipfs daemon > daemon_out 2> daemon_err
 '
 
 test_expect_success "output looks good" '
--- a/test/sharness/t0066-migration.sh
+++ b/test/sharness/t0066-migration.sh
@@ -10,16 +10,27 @@ test_description="Test migrations auto update prompt"
 
 test_init_ipfs
 
+# Create fake migration binaries instead of letting ipfs download from network
+# To test downloading and running actual binaries, comment out this test.
 test_expect_success "setup mock migrations" '
   mkdir bin &&
-  echo "#!/bin/bash" > bin/fs-repo-migrations &&
-  echo "echo 5" >> bin/fs-repo-migrations &&
-  chmod +x bin/fs-repo-migrations &&
+  echo "#!/bin/bash" > bin/fs-repo-7-to-8 &&
+  echo "echo fake applying 7-to-8 repo migration" >> bin/fs-repo-7-to-8 &&
+  chmod +x bin/fs-repo-7-to-8 &&
+  echo "#!/bin/bash" > bin/fs-repo-8-to-9 &&
+  echo "echo fake applying 8-to-9 repo migration" >> bin/fs-repo-8-to-9 &&
+  chmod +x bin/fs-repo-8-to-9 &&
+  echo "#!/bin/bash" > bin/fs-repo-9-to-10 &&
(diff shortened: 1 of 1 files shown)


test/sharness/t0066-migration.sh
```diff
@@ -30,17 +41,25 @@ test_expect_success "output looks good" '
...
+  grep "applying 8-to-9 repo migration" true_out > /dev/null &&
+  grep "applying 9-to-10 repo migration" true_out > /dev/null &&
+  grep "applying 10-to-11 repo migration" true_out > /dev/null &&
+  grep "Success: fs-repo migrated to version 11" true_out > /dev/null
 '
 
+# Daemon exits with 1 after the prompt gets "n" when asking to migrate. 
 test_expect_success "'ipfs daemon' prompts to auto migrate" '
-  test_expect_code 1 ipfs daemon > daemon_out 2> daemon_err
+  test_expect_code 1 bash -c "echo n | ipfs daemon > daemon_out 2> daemon_err"
```
aschmahmann, line 62: Why has this part of the test changed? Isn't the prompt for auto-migration code still the same?

test/sharness/t0066-migration.sh
```diff
@@ -10,16 +10,27 @@ test_description="Test migrations auto update prompt"
 
 test_init_ipfs
 
+# Create fake migration binaries instead of letting ipfs download from network
+# To test downloading and running actual binaries, comment out this test.
```
aschmahmann, line 14: We weren't testing real binaries here before, right? So this is just a heads up for developers/testers.

test/sharness/t0066-migration.sh
gammazero: The prompt is the same, but needed to `echo n` to answer the prompt so that it would not just hang.  Needed the `bash -c ...` to handle the pipe in the command.

test/sharness/t0066-migration.sh
gammazero: The actual migrations are tested in fs-repo-migrations sharness tests.  This part tests that the go-ipfs will actually run the migrations.

The part that is missing from a complete end-to-end test is testing go-ipfs downloading migrations from the network.

test/sharness/t0066-migration.sh
gammazero: After retesting, I do not know why this change seemed necessary.  Changed back to original.

test/sharness/t0066-migration.sh
```diff
@@ -30,15 +41,23 @@ test_expect_success "output looks good" '
...
-  grep "Success: fs-repo has been migrated to version 5." true_out > /dev/null
+test_expect_success "output looks good" '
+  grep "applying 7-to-8 repo migration" true_out > /dev/null &&
+  grep "applying 8-to-9 repo migration" true_out > /dev/null &&
+  grep "applying 9-to-10 repo migration" true_out > /dev/null &&
+  grep "applying 10-to-11 repo migration" true_out > /dev/null &&
+  grep "Success: fs-repo migrated to version 11" true_out > /dev/null
 '
 
+# Daemon exits with 1 after the prompt gets "n" when asking to migrate.
```
aschmahmann, line 60: It's not getting "n" anymore, just waiting right?

test/sharness/t0066-migration.sh
```diff
@@ -30,15 +41,23 @@ test_expect_success "output looks good" '
   grep "Please get fs-repo-migrations from https://dist.ipfs.io" false_out
 '
 
+# The migrations will succeed, but the daemon will still exit with 1 because
+# the fake migrations do not update the repo version number.
+#
+# If run with real migrations, the daemon continues running and must be killed.
 test_expect_success "ipfs daemon --migrate=true runs migration" '
-  test_expect_code 1 ipfs daemon --migrate=true > true_out
+  test_expect_code 1 ipfs daemon --migrate=true > true_out 2>&1
```
aschmahmann, line 49: What is this change doing for us?

gammazero: That was needed because PR #8054 was not merged yet.  Some of the output was now going to stderr that was expected to go to stdout.  That made a subsequent test, that scanned `true_out` fail, but for the wrong reasons.

gammazero: If I rebase on top of #8054, then that change can go away.

test/sharness/t0066-migration.sh
gammazero: yes. attention to detail, thank you.  Test description is sufficient, no comment needed.

test/sharness/t0066-migration.sh
aschmahmann: Sounds good, let's rebase now that the bug is fixed and keep it as is.

gammazero: Done.  All tests passing.

aschmahmann (approved):

PR #8055, opened 2021-04-06
chore: dont docker tag rc as latest

### LINKED ISSUE #3999: Publishing of docker images

olizilla:
- add guard for rc releases; publish them to docker hub with the matching git tag
- make the semver regex stricter and only publish as latest when a full semver tag with no pre-release suffix is the name of the git tag.
- add `release` tag as an alias of `latest` as per https://github.com/ipfs/go-ipfs/issues/3999#issuecomment-742228981

Tested manually as the push-docker-tag.sh script is set up for it:
```shell
./push-docker-tags.sh $(date -u +%F) testingsha release v0.9.0-test dry
Nothing to do. No docker tag defined for branch: release, tag: v0.9.0-test

./push-docker-tags.sh $(date -u +%F) testingsha release v0.9.0-rc1 dry
DRY RUN! I would have tagged and pushed the following...
docker tag ipfs/go-ipfs:wip ipfs/go-ipfs:v0.9.0-rc1
docker push ipfs/go-ipfs:v0.9.0-rc1

./push-docker-tags.sh $(date -u +%F) testingsha release v0.9.0 dry
DRY RUN! I would have tagged and pushed the following...
docker tag ipfs/go-ipfs:wip ipfs/go-ipfs:v0.9.0
docker push ipfs/go-ipfs:v0.9.0
DRY RUN! I would have tagged and pushed the following...
docker tag ipfs/go-ipfs:wip ipfs/go-ipfs:latest
docker push ipfs/go-ipfs:latest
DRY RUN! I would have tagged and pushed the following...
docker tag ipfs/go-ipfs:wip ipfs/go-ipfs:release
docker push ipfs/go-ipfs:release
```

fixes #3999

License: MIT

### DIFF
--- a/bin/push-docker-tags.sh
+++ b/bin/push-docker-tags.sh
@@ -6,7 +6,7 @@
 # A bit like dockerhub autobuild config, but somewhere we can version control it.
 # 
 # The `docker-build` job in .circleci/config.yml builds the current commit
-# in docker and tags it as ipfs/go-ipfs:wip
+# in docker and tags it as ipfs/go-ipfs:wip
 #
 # Then the `docker-publish` job runs this script to decide what tag, if any,
 # to publish to dockerhub.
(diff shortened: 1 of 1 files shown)


bin/push-docker-tags.sh
```diff
@@ -59,9 +59,13 @@ pushTag () {
   fi
 }
 
-if [[ $GIT_TAG =~ ^v[0-9]+ ]]; then
+if [[ $GIT_TAG =~ ^v[0-9]+\.[0-9]+\.[0-9]+-rc ]]; then
+  pushTag "$GIT_TAG"
+
+elif [[ $GIT_TAG =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
   pushTag "$GIT_TAG"
   pushTag "latest"
```
gammazero, line 67: Consider dropping the `latest` tag.  It may be confusing whether it refers to the latest release, to an RC, or something else.

olizilla: I agree in principle, but in practice we get issues when tags that have been around a while stop getting updates, so it's probably least surprising to keep it working. That's also why I have begrudgingly restored the release tag here too.

gammazero (approved):

PR #8061, opened 2021-04-08
Experimental dht client

aschmahmann:
This feature enables an option for an experimental DHT client that makes a few key optimizations over the standard client.
1. It periodically crawls the network to find all the peers and then finds the closest peers to the target in Kademlia space based on the local list of peers
2. It does not wait on all peers from a given query to return, but instead tries to guess when we'll be done hearing from peers and instead be waiting on timeouts.
   - The current metric is to wait for some percentage (30%) of the closest bucket-size (i.e. 20) peers to return and then wait for some dead time (a few hundred ms) where no new responses come in

Depends on:
- [x] https://github.com/ipfs/go-ipfs-config/pull/125
- [x] https://github.com/libp2p/go-libp2p-kad-dht/pull/709

A couple quirks about the current state of this PR and the client:
- I've manually set the experimental option to always on for debugging purposes (will turn off before merge)
  - turned off
- The DHT client crawls for updates once an hour and takes about 10 minutes to complete in my tests on the live network
   - This means the node's DHT is unusable for the first 10 minutes of operation
   - This means many of the DHT tests will fail since the node will either not have an updated view of the network or will simply have an empty routing table
- Given the utility of a faster DHT client I'm hoping that an experimental client that's much much faster and that the community can hammer on will be appreciated while we work on improving the client to handle more of the edge cases

----

This PR also includes an experimental provider system which utilizes the experimental dht client and is bundled with it - the originating PR was #8070

### DIFF
--- a/core/commands/stat_dht.go
+++ b/core/commands/stat_dht.go
@@ -12,6 +12,7 @@ import (
 	"github.com/libp2p/go-libp2p-core/network"
 	pstore "github.com/libp2p/go-libp2p-core/peerstore"
 	dht "github.com/libp2p/go-libp2p-kad-dht"
+	"github.com/libp2p/go-libp2p-kad-dht/fullrt"
 	kbucket "github.com/libp2p/go-libp2p-kbucket"
 )
 
@@ -43,7 +44,8 @@ This interface is not stable and may change from release to release.
 `,
 	},
 	Arguments: []cmds.Argument{
-		cmds.StringArg("dht", false, true, "The DHT whose table should be listed (wan or lan). Defaults to both."),
+		cmds.StringArg("dht", false, true, "The DHT whose table should be listed (wanserver, lanserver, wan, lan). "+
+			"wan and lan refer to client routing tables. When using the experimental DHT client only WAN is supported. Defaults to wan and lan."),
 	},
 	Options: []cmds.Option{},
 	Run: func(req *cmds.Request, res cmds.ResponseEmitter, env cmds.Environment) error {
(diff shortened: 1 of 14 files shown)


core/node/libp2p/routing.go
```diff
@@ -32,23 +35,89 @@ type p2pRouterOut struct {
...
-				return dr.Close()
+func BaseRouting(experimentalDHTClient bool) interface{} {
+	return func(mctx helpers.MetricsCtx, lc fx.Lifecycle, in processInitialRoutingIn) (out processInitialRoutingOut, err error) {
+		var dr *ddht.DHT
+		if dht, ok := in.Router.(*ddht.DHT); ok {
+			dr = dht
+
+			lc.Append(fx.Hook{
+				OnStop: func(ctx context.Context) error {
+					return dr.Close()
```
Stebalien, line 66: This is strange... I'd expect DI function constructing the DHT to register a close hook.

core/commands/stat_dht.go
```diff
@@ -67,12 +69,62 @@ This interface is not stable and may change from release to release.
...
+
 			switch name {
 			case "wan":
+				if separateClient {
+					client, ok := nd.DHTClient.(*fullrt.FullRT)
+					if !ok {
+						return cmds.Errorf(cmds.ErrClient, "could not generate stats for the WAN DHT client type")
+					}
+					peerMap := client.Stat()
+					buckets := make([]dhtBucket, 1)
```
gammazero, line 89: nit: not needed here, just declare in-line when constructing `dhtStat` line 113
```go
dhtStat{
    Name:    name,
    Buckets: []dhtBucket{*b},
}
```

gammazero (approved):

core/node/libp2p/routing.go
aschmahmann: yep, although this was here before so I'm going to leave it for now. I suspect it's related to the initial routing being set specially via a variable in the build config object.

core/commands/stat_dht.go
aschmahmann: thx, just going to leave it for now (don't want to think too hard about if this change is worth while). Either way we should really be showing the peers in multiple buckets just like the regular client, and we can do that it was just a bit of a pain so I skipped it for now.

core/commands/stat_dht.go
```diff
@@ -43,7 +44,8 @@ This interface is not stable and may change from release to release.
 `,
 	},
 	Arguments: []cmds.Argument{
-		cmds.StringArg("dht", false, true, "The DHT whose table should be listed (wan or lan). Defaults to both."),
+		cmds.StringArg("dht", false, true, "The DHT whose table should be listed (wanserver, lanserver, wan, lan). "+
+			"wan and lan refer to client routing tables. When using the experimental DHT client only WAN is supported. Defaults to wan and lan."),
```
alanshaw, line 48: What are wanserver/lanserver and are they new? Can the description explain 🙏

docs/experimental-features.md
```diff
@@ -547,3 +548,56 @@ ipfs config --json Experimental.GraphsyncEnabled true
...
+prepared. This means operations like searching the DHT for particular peers or content will not work
+   - You can see if the DHT has been initially populated by running `ipfs stats dht`
+3. Currently, the accelerated DHT client is not compatible with LAN-based DHTs and will not perform operations against
+them
+
+### How to enable
+
+```
+ipfs config --json Experimental.AcceleratedDHTClient true
+```
```
alanshaw, line 597: Add a note that if the daemon is running it must be restarted?

PR #8064, opened 2021-04-08
Feat/migration ipfs download

gammazero:
go-ipfs can now fetch migrations via IPFS

Project proposal: https://github.com/protocol/web3-dev-team/pull/86

### DIFF
--- a/cmd/ipfs/daemon.go
+++ b/cmd/ipfs/daemon.go
@@ -476,7 +476,10 @@ func daemonFunc(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment
 
 	// Add any files downloaded by migration.
 	if keepMigrations {
-		addMigrations(cctx.Context(), node, fetcher, pinMigrations)
+		err = addMigrations(cctx.Context(), node, fetcher, pinMigrations)
+		if err != nil {
+			fmt.Fprintln(os.Stderr, "Could not add migragion to IPFS:", err)
+		}
 		fetcher.Close()
 		os.RemoveAll(migrations.DownloadDirectory)
 		migrations.DownloadDirectory = ""
--- a/cmd/ipfs/daemon.go
+++ b/cmd/ipfs/daemon.go
@@ -297,27 +297,16 @@ func daemonFunc(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment
 			return err
 		}
 
-		keep := migrationCfg.Keep
-		if keep == "" {
-			keep = config.DefaultMigrationKeep
-		}
-		switch keep {
-		case "discard":
-		case "cache":
-			cacheMigrations = true
-		case "pin":
-			pinMigrations = true
-		default:
-			return errors.New("unrecognized value for %s, must be 'cache', 'pin', or 'discard'")
-		}
-
-		dlSources := migrationCfg.DownloadSources
-		if len(dlSources) == 0 {
-			dlSources = config.DefaultMigrationDownloadSources
-		}
-		fetcher, err = getMigrationFetcher(dlSources, migrationCfg.Peers)
+		fetcher, err = getMigrationFetcher(migrationCfg, &cctx.ConfigRoot)
 		if err != nil {
 			return err
+		}
+		defer fetcher.Close()
+
+		if migrationCfg.Keep == "cache" {
+			cacheMigrations = true
+		} else if migrationCfg.Keep == "pin" {
+			pinMigrations = true
 		}
 
 		if cacheMigrations || pinMigrations {
@@ -339,19 +328,7 @@ func daemonFunc(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment
 			fmt.Printf("  %s\n", err)
 			fmt.Println("If you think this is a bug, please file an issue and include this whole log output.")
 			fmt.Println("  https://github.com/ipfs/fs-repo-migrations")
-			fetcher.Close()
 			return err
-		}
-
-		if cacheMigrations || pinMigrations {
-			defer fetcher.Close()
-		} else {
-			// If there is an error closing the IpfsFetcher, then print error, but
-			// do not fail because of it.
-			err = fetcher.Close()
-			if err != nil {
-				log.Errorf("error closing IPFS fetcher: %s", err)
-			}
 		}
 
 		repo, err = fsrepo.Open(cctx.ConfigRoot)
@@ -481,9 +458,16 @@ func daemonFunc(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment
 		if err != nil {
 			fmt.Fprintln(os.Stderr, "Could not add migragion to IPFS:", err)
 		}
-		fetcher.Close()
 		os.RemoveAll(migrations.DownloadDirectory)
 		migrations.DownloadDirectory = ""
+	}
+	if fetcher != nil {
+		// If there is an error closing the IpfsFetcher, then print error, but
+		// do not fail because of it.
+		err = fetcher.Close()
+		if err != nil {
+			log.Errorf("error closing IPFS fetcher: %s", err)
+		}
 	}
 
 	// construct http gateway
--- a/cmd/ipfs/daemon.go
+++ b/cmd/ipfs/daemon.go
@@ -200,6 +200,34 @@ func defaultMux(path string) corehttp.ServeOption {
 	}
 }
 
+// readMigrationConfig reads the migration config out of the config, avoiding reading anything other
+// than the migration section. That way, we're free to make arbitrary changes to all _other_
+// sections in migrations.
+func readMigrationConfig(repoRoot string) (*config.Migration, error) {
+	var cfg struct {
+		Migration config.Migration
+	}
+
+	cfgPath, err := config.Filename(repoRoot)
+	if err != nil {
+		return nil, err
+	}
+
+	cfgFile, err := os.Open(cfgPath)
+	if err != nil {
+		return nil, err
+	}
+
+	defer cfgFile.Close()
+
+	err = json.NewDecoder(cfgFile).Decode(&cfg)
+	if err != nil {
+		return nil, err
+	}
+
+	return &cfg.Migration, nil
+}
+
 func daemonFunc(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment) (_err error) {
 	// Inject metrics before we do anything
 	err := mprome.Inject()
@@ -293,26 +321,12 @@ func daemonFunc(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment
 			return fmt.Errorf("fs-repo requires migration")
 		}
 
-		var cfg struct {
-			Migration config.Migration
+		migrationCfg, err := readMigrationConfig(cctx.ConfigRoot)
+		if err != nil {
+			return err
 		}
 
-		{
-			cfgPath, err := config.Filename(cctx.ConfigRoot)
-			if err != nil {
-				return err
-			}
-			cfgFile, err := os.Open(cfgPath)
-			if err != nil {
-				return err
-			}
-			err = json.NewDecoder(cfgFile).Decode(&cfg)
-			if err != nil {
-				return err
-			}
-		}
-
-		keep := cfg.Migration.Keep
+		keep := migrationCfg.Keep
 		if keep == "" {
 			keep = config.DefaultMigrationKeep
 		}
@@ -326,11 +340,11 @@ func daemonFunc(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment
 			return errors.New("unrecognized value for %s, must be 'cache', 'pin', or 'discard'")
 		}
 
-		dlSources := cfg.Migration.DownloadSources
+		dlSources := migrationCfg.DownloadSources
 		if len(dlSources) == 0 {
 			dlSources = config.DefaultMigrationDownloadSources
 		}
-		fetcher, err = getMigrationFetcher(dlSources, cfg.Migration.Peers)
+		fetcher, err = getMigrationFetcher(dlSources, migrationCfg.Peers)
 		if err != nil {
 			return err
 		}
--- a/cmd/ipfs/daemon.go
+++ b/cmd/ipfs/daemon.go
@@ -49,9 +49,6 @@ const (
 	ipfsMountKwd              = "mount-ipfs"
 	ipnsMountKwd              = "mount-ipns"
 	migrateKwd                = "migrate"
-	migrateDownloadKwd        = "download-migration"
-	migrateKeepKwd            = "keep-migration"
-	migratePeersKwd           = "migration-peers"
 	mountKwd                  = "mount"
 	offlineKwd                = "offline" // global option
 	routingOptionKwd          = "routing"
@@ -177,9 +174,6 @@ Headers.
 		cmds.BoolOption(enableGCKwd, "Enable automatic periodic repo garbage collection"),
 		cmds.BoolOption(adjustFDLimitKwd, "Check and raise file descriptor limits if needed").WithDefault(true),
 		cmds.BoolOption(migrateKwd, "If true, assume yes at the migrate prompt. If false, assume no."),
-		cmds.StringOption(migrateDownloadKwd, "Comma-separated list of \"http\", \"ipfs\", or custom gateway URL"),
-		cmds.StringOption(migrateKeepKwd, "What to do with downloaded migrations: \"keep\", \"pin\", \"discard\". Default: \"keep\""),
-		cmds.StringOption(migratePeersKwd, "IPFS peers to connect to when downloading migration binaries. Comma-separated list of addresses like those returned by ipfs id"),
 		cmds.BoolOption(enablePubSubKwd, "Instantiate the ipfs daemon with the experimental pubsub feature enabled."),
 		cmds.BoolOption(enableIPNSPubSubKwd, "Enable IPNS record distribution through pubsub; enables pubsub."),
 		cmds.BoolOption(enableMultiplexKwd, "DEPRECATED"),
@@ -275,7 +269,7 @@ func daemonFunc(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment
 		}
 	}
 
-	var keepMigrations, pinMigrations bool
+	var cacheMigrations, pinMigrations bool
 	var fetcher migrations.Fetcher
 
 	// acquire the repo lock _before_ constructing a node. we need to make
@@ -298,28 +292,33 @@ func daemonFunc(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment
 			return fmt.Errorf("fs-repo requires migration")
 		}
 
-		switch keep, _ := req.Options[migrateKeepKwd].(string); keep {
+		// Read from existing config
+		cfg, err := cctx.GetConfig()
+
+		keep := cfg.Migration.Keep
+		if keep == "" {
+			keep = config.DefaultMigrationKeep
+		}
+		switch keep {
+		case "discard":
+		case "cache":
+			cacheMigrations = true
 		case "pin":
 			pinMigrations = true
-			fallthrough
-		case "", "keep":
-			keepMigrations = true
-		case "discard":
 		default:
-			return errors.New("unrecognized value for %s, must be 'keep', 'pin', or 'discard'")
+			return errors.New("unrecognized value for %s, must be 'cache', 'pin', or 'discard'")
 		}
 
-		// Try to read existing config, but do not fail if it cannot be read.
-		peers, _ := req.Options[migratePeersKwd].(string)
-
-		// Get migration fetcher(s) according to download policy
-		policy, _ := req.Options[migrateDownloadKwd].(string)
-		fetcher, err = getMigrationFetcher(policy, peers)
+		dlSources := cfg.Migration.DownloadSources
+		if len(dlSources) == 0 {
+			dlSources = config.DefaultMigrationDownloadSources
+		}
+		fetcher, err = getMigrationFetcher(dlSources, cfg.Migration.Peers)
 		if err != nil {
 			return err
 		}
 
-		if keepMigrations {
+		if cacheMigrations || pinMigrations {
 			// Create temp directory to store downloaded migration archives
 			migrations.DownloadDirectory, err = ioutil.TempDir("", "migrations")
 			if err != nil {
@@ -342,7 +341,7 @@ func daemonFunc(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment
 			return err
 		}
 
-		if keepMigrations {
+		if cacheMigrations || pinMigrations {
 			defer fetcher.Close()
 		} else {
 			// If there is an error closing the IpfsFetcher, then print error, but
@@ -475,7 +474,7 @@ func daemonFunc(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment
 	}
 
 	// Add any files downloaded by migration.
-	if keepMigrations {
+	if cacheMigrations || pinMigrations {
 		err = addMigrations(cctx.Context(), node, fetcher, pinMigrations)
 		if err != nil {
 			fmt.Fprintln(os.Stderr, "Could not add migragion to IPFS:", err)
--- a/cmd/ipfs/daemon.go
+++ b/cmd/ipfs/daemon.go
@@ -314,15 +314,20 @@ func daemonFunc(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment
 		// Fetchers are tried in the order given to NewMultiFetcher
 		multiFetcher := migrations.NewMultiFetcher(f1, f2)
 		err = migrations.RunMigration(cctx.Context(), multiFetcher, fsrepo.RepoVersion, "", false)
+		closeErr := fetchIpfs.Close()
 		if err != nil {
 			fmt.Println("The migrations of fs-repo failed:")
 			fmt.Printf("  %s\n", err)
 			fmt.Println("If you think this is a bug, please file an issue and include this whole log output.")
 			fmt.Println("  https://github.com/ipfs/fs-repo-migrations")
-			fetchIpfs.Close()
 			return err
 		}
-		fetchIpfs.Close()
+
+		// If there was an error closing the IpfsFetcher, then write error, but
+		// do not fail because of it.
+		if closeErr != nil {
+			log.Errorf("error closing IPFS fetcher: %s", closeErr)
+		}
 
 		repo, err = fsrepo.Open(cctx.ConfigRoot)
 		if err != nil {
--- a/cmd/ipfs/daemon.go
+++ b/cmd/ipfs/daemon.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"encoding/json"
 	"errors"
 	_ "expvar"
 	"fmt"
@@ -198,34 +197,6 @@ func defaultMux(path string) corehttp.ServeOption {
 		mux.Handle(path, http.DefaultServeMux)
 		return mux, nil
 	}
-}
-
-// readMigrationConfig reads the migration config out of the config, avoiding reading anything other
-// than the migration section. That way, we're free to make arbitrary changes to all _other_
-// sections in migrations.
-func readMigrationConfig(repoRoot string) (*config.Migration, error) {
-	var cfg struct {
-		Migration config.Migration
-	}
-
-	cfgPath, err := config.Filename(repoRoot)
-	if err != nil {
-		return nil, err
-	}
-
-	cfgFile, err := os.Open(cfgPath)
-	if err != nil {
-		return nil, err
-	}
-
-	defer cfgFile.Close()
-
-	err = json.NewDecoder(cfgFile).Decode(&cfg)
-	if err != nil {
-		return nil, err
-	}
-
-	return &cfg.Migration, nil
 }
 
 func daemonFunc(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment) (_err error) {
--- a/cmd/ipfs/daemon.go
+++ b/cmd/ipfs/daemon.go
@@ -291,15 +291,15 @@ func daemonFunc(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment
 			return fmt.Errorf("fs-repo requires migration")
 		}
 
-		// TODO: Get optional peers to connect to.
-		//   - Read from existing config?
-		//   - Read from cli?
-		//   - Both?
-		var peers []string
+		// Read from existing config
+		cfg, err := cctx.GetConfig()
+		if err != nil {
+			return fmt.Errorf("migrate: GetConfig() failed: %s", err)
+		}
 
 		// Fetch migrations from current distribution, or location from environ
 		fetchHttp := migrations.NewHttpFetcher(migrations.GetDistPathEnv(migrations.CurrentIpfsDist), "", "go-ipfs", 0)
-		fetchIpfs := ipfsfetcher.NewIpfsFetcher(migrations.GetDistPathEnv(migrations.CurrentIpfsDist), 0, peers)
+		fetchIpfs := ipfsfetcher.NewIpfsFetcher(migrations.GetDistPathEnv(migrations.CurrentIpfsDist), 0, cfg.Peering.Peers)
 
 		var (
 			f1 migrations.Fetcher = fetchHttp
--- a/cmd/ipfs/daemon.go
+++ b/cmd/ipfs/daemon.go
@@ -48,6 +48,7 @@ const (
 	ipfsMountKwd              = "mount-ipfs"
 	ipnsMountKwd              = "mount-ipns"
 	migrateKwd                = "migrate"
+	migrateIpfsKwd            = "ipfs"
 	mountKwd                  = "mount"
 	offlineKwd                = "offline" // global option
 	routingOptionKwd          = "routing"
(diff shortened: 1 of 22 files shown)


repo/fsrepo/migrations/ipfsfetcher_test.go
```diff
@@ -0,0 +1,47 @@
+package migrations
+
+import (
+	"bufio"
+	"context"
+	"testing"
+)
+
+func TestIpfsFetcher(t *testing.T) {
+	t.Skip("manually-run dev test only")
```
mvdan, line 10: maybe hide it behind a flag, then :) otherwise the only way to run it is by editing the source.

is there any way to eventually make this test work in a local way without extra setup? e.g. by setting up some IPFS/HTTP servers/nodes.

repo/fsrepo/migrations/ipfsfetcher_test.go
```diff
@@ -0,0 +1,47 @@
...
+	fetcher := NewIpfsFetcher("", 0, nil)
+	defer fetcher.Close()
+
+	rc, err := fetcher.Fetch(ctx, "go-ipfs/versions")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer rc.Close()
+
+	var out []string
```
mvdan, line 24: I think this would be clearer if called `lines` :)

mvdan (commented):
Just did a first pass as it's quite late. I can do a second one tomorrow.

repo/fsrepo/migrations/ipfsfetcher_test.go
gammazero: done

repo/fsrepo/migrations/ipfsfetcher_test.go
gammazero: Done - hidden behind flag

For the HttpFetcher I created a dummy test server.  I am not really sure there is a great way to do this for IPFS.  Mocking interfaces seems a bit excessive.

repo/fsrepo/migrations/ipfsfetcher.go
```diff
@@ -0,0 +1,297 @@
...
+			fmt.Fprintf(os.Stderr, "attempting to connect to peer: %q\n", pi)
+			err := ipfs.Swarm().Connect(ctx, *pi)
+			if err != nil {
+				fmt.Fprintf(os.Stderr, "failed to connect to %s: %s", pi.ID, err)
+			}
+			fmt.Fprintf(os.Stderr, "successfully connected to %s\n", pi.ID)
+		}(pi)
+	}
+	wg.Wait()
+	return nil
```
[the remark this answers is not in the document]
gammazero: Done, all errors returned

repo/fsrepo/migrations/ipfsfetcher.go
```diff
@@ -0,0 +1,297 @@
...
+	}
+
+	u, err := url.Parse(path)
+	if err != nil {
+		return nil, fmt.Errorf("%q could not be parsed: %s", path, err)
+	}
+
+	switch proto := u.Scheme; proto {
+	case "ipfs", "ipld", "ipns":
+		ipfsPath = ipath.New(filepath.Join("/", proto, u.Host, u.Path))
```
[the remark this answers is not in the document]
gammazero: Right, that is a mistake.

Done - use path instead of filepath

repo/fsrepo/migrations/ipfsfetcher.go
```diff
@@ -0,0 +1,297 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	// TODO: Is there a better way to check it plugins are loaded first?
+	err = setupPlugins(defaultPath)
+	// Need to ignore errors here because plugins may already be loaded when
+	// run from ipfs daemon.
+	if err != nil {
+		fmt.Println("Ignored plugin error:", err)
```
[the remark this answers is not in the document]
gammazero: Will print to stderr for now, and make it look less like an error.  If ipfsfetcher is only going to be used from within go-ipfs, then it may be better to not try to load the pluging at all.

repo/fsrepo/migrations/ipfsfetcher.go
```diff
@@ -0,0 +1,297 @@
...
+			f.ipfsCancel()
+
+			// Wait until ipfs is stopped
+			<-f.ipfsCtx.Done()
+		}
+
+		if f.ipfsTmpDir != "" {
+			// Remove the temp ipfs dir
+			if err := os.RemoveAll(f.ipfsTmpDir); err != nil {
+				fmt.Fprintln(os.Stderr, err)
```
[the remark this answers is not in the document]
gammazero: done

repo/fsrepo/migrations/ipfsfetcher.go
```diff
@@ -0,0 +1,297 @@
...
+
+	if distPath != "" {
+		if !strings.HasPrefix(distPath, "/") {
+			distPath = "/" + distPath
+		}
+		f.distPath = distPath
+	}
+
+	if fetchLimit != 0 {
+		if fetchLimit == -1 {
```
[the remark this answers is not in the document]
gammazero: done

repo/fsrepo/migrations/ipfsfetcher.go
```diff
@@ -0,0 +1,297 @@
...
+// Specifying 0 for fetchLimit sets the default, -1 means no limit.
+func NewIpfsFetcher(distPath string, fetchLimit int64, peers []string) *IpfsFetcher {
+	f := &IpfsFetcher{
+		limit:    defaultFetchLimit,
+		distPath: LatestIpfsDist,
+		peers:    peers,
+	}
+
+	if distPath != "" {
+		if !strings.HasPrefix(distPath, "/") {
```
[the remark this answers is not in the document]
gammazero: distPath is a URL path, so I will treat it as such.  The use of filepath on it was a mistake.

repo/fsrepo/migrations/ipfsfetcher.go
```diff
@@ -0,0 +1,297 @@
...
+)
+
+type IpfsFetcher struct {
+	distPath string
+	limit    int64
+	peers    []string
+
+	openOnce  sync.Once
+	closeOnce sync.Once
+	err       error
```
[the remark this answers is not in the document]
gammazero: done

gammazero:
@mvdan I addressed all review comments and made suggested changes.  One other big change I needed to make was to move IpfsFetcher into its own package.  This is so that things outside of ipfs, that import `migrations`, are not forced to link with all the IpfsFetcher baggage.

gammazero:
Remaining items:
- Migration policy config
- Node should not listen for inbound connections
- Node should not expose an API/gateway.
- HTTP fetcher => HTTPS fetcher
- Read IPFS peers from existing config

repo/fsrepo/migrations/ipfsfetcher_test.go
mvdan: I'm just a bit worried that we have a test that will practically never be run. Mocking isn't great, but if running a real IPFS node for the test is too difficult, at least testing with a mock by default would be better than testing nothing by default. The flag could still be used to test against a real IPFS node instead of a mock.

iand: Another pattern is to use an environment variable and put the name of it in the Skip message. Inspired by this blog post https://peter.bourgon.org/blog/2021/04/02/dont-use-build-tags-for-integration-tests.html

repo/fsrepo/migrations/ipfsfetcher/ipfsfetcher.go
```diff
@@ -0,0 +1,288 @@
...
+	// create temporary ipfs directory
+	dir, err := ioutil.TempDir("", "ipfs-temp")
+	if err != nil {
+		return "", fmt.Errorf("failed to get temp dir: %s", err)
+	}
+
+	// configure the temporary node
+	cfg.Routing.Type = "dhtclient"
+
+	// Disable listening for inbound connections
```
aschmahmann, line 160: We could do that, but then if you try and import the data into the new node from the old node for seeding purposes you're going to need to either:
- Use the import/export APIs (e.g. use CAR files to move the data around)
- Have the temporary node connect to the new node so it can pull the data using Bitswap

repo/fsrepo/migrations/ipfsfetcher/ipfsfetcher.go
```diff
@@ -0,0 +1,288 @@
...
+	u, err := url.Parse(fetchPath)
+	if err != nil {
+		return nil, fmt.Errorf("%q could not be parsed: %s", fetchPath, err)
+	}
+
+	switch proto := u.Scheme; proto {
+	case "ipfs", "ipld", "ipns":
+		ipfsPath = ipath.New(path.Join("/", proto, u.Host, u.Path))
+	case "http", "https":
+		ipfsPath = ipath.New(u.Path)
```
aschmahmann, line 232: Why would we get an HTTP(S) URL here? If we can't then let's remove this code, if we can then can let's discuss our options since this seems it will get us into trouble (e.g. what about subdomain gateways, maybe path gateways can start with some prefix like foo.tld/custom/ipfs/Qmxyz).

repo/fsrepo/migrations/ipfsfetcher/ipfsfetcher_test.go
```diff
@@ -0,0 +1,97 @@
...
+	"context"
+	"flag"
+	"testing"
+)
+
+var runIpfsTest = flag.Bool("ipfstest", false, "Run IpfsFetcher tests")
+
+func TestIpfsFetcher(t *testing.T) {
+	if !*runIpfsTest {
+		t.Skip("manually-run dev test, use '-ipfstest' flage to run")
```
aschmahmann, line 14:
```suggestion
		t.Skip("manually-run dev test, use the '-ipfstest' flag to run")
```

aschmahmann, line 14: Is the reason this test is manual because it makes network requests?

aschmahmann (changes requested):

cmd/ipfs/daemon.go
```diff
@@ -48,6 +49,7 @@ const (
 	ipfsMountKwd              = "mount-ipfs"
 	ipnsMountKwd              = "mount-ipns"
 	migrateKwd                = "migrate"
+	migrateIpfsKwd            = "ipfs"
```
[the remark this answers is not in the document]
aschmahmann: The proposal https://github.com/protocol/web3-dev-team/pull/86/files#diff-991744cab1b93840a21929cbf5aa7bc00a37ae4244b5c569df086214e36e47cfR98 indicates we should just do HTTP first and fallback to IPFS for a release or two and then switch to IPFS first.

Are we running into testing issues doing it that way?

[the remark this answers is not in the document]
gammazero: This was done as a way to test/demo before and [configurable policy](https://github.com/protocol/web3-dev-team/pull/86/files#diff-991744cab1b93840a21929cbf5aa7bc00a37ae4244b5c569df086214e36e47cfR66) is available.  I have an update that will provide this policy and will remove the migrate `ipfs` flag.

repo/fsrepo/migrations/ipfsfetcher/ipfsfetcher.go
gammazero: Since the files are already available on local disk (since they need to be run), it is just as easy for me to add the files to the real node after migration. So, I think this is fine since I am not using the temp node for anything other than downloading migrations.

aschmahmann: > it is just as easy for me to add the files to the real node after migration

Right, but if we do this it needs to be as a CAR file not a UnixFS import of some binary since we want the CIDs to match and `ipfs add <file>` is not guaranteed to be the same across versions and boxing ourselves into some parameters now wouldn't be a good idea either. i.e. go-ipfs needs to understand "I am importing this DAG" not "I am importing this binary", whether it's a CAR file via DAG import, or UnixFS via Bitswap is 🤷.

cmd/ipfs/daemon.go
```diff
@@ -288,9 +291,30 @@ func daemonFunc(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment
 			return fmt.Errorf("fs-repo requires migration")
 		}
 
+		// Read from existing config
+		cfg, err := cctx.GetConfig()
```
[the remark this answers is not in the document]
gammazero: It seems like allowing an entire config to be specified for a temp node also invites many problems, as users will not necessarily know what the new configuration looks like until after the migration and may typically just use their existing config (leading to issues mentioned above).  Also, the configuration for a temp, used only to download a migration or two, does not need more than a default config.  Users not knowing this may think it best to use a config similar to what they are using for their real node.

Instead of a config file, WDYT about letting the user specify peer(s) on the command line, if they happen to know one 
 or more that may be useful for migration?

repo/fsrepo/migrations/fetcher.go
```diff
@@ -45,15 +46,21 @@ func NewMultiFetcher(f ...Fetcher) Fetcher {
...
 			// Transferred using this fetcher
-			return
+			return rc, nil
 		}
+		errs = append(errs, err)
 	}
-	return
+	err := fmt.Errorf("fetch errors:")
+	for i := range errs {
+		err = fmt.Errorf("%s\n%d) %s", err.Error(), i, errs[i])
```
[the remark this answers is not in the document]
gammazero: Returns list of errors if all fetchers fail.  Switched to using `multierror`

repo/fsrepo/migrations/ipfsfetcher/ipfsfetcher.go
gammazero: Removed.  If someone were to try to use the IpfsFetcher with an HTTP path, then returning an is acceptable, and they really mean to use the HttpFetcher.

repo/fsrepo/migrations/ipfsfetcher/ipfsfetcher_test.go
gammazero: Fixed

gammazero: Yes, and I did not know a good way to simulate or mock an IPFS node, but I wanted to have some test that a developer could run to check if things are working.  WDYT? Leave it, toss it, create a better test?

cmd/ipfs/daemon.go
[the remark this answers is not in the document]
aschmahmann: > It seems like allowing an entire config to be specified for a temp node also invites many problems...

Ya, it's using a pretty big hammer to deal with a smaller set of problems but needing to modify the config for the purposes of migrations at all seems like a potentially advanced requirement for an end user, perhaps sufficiently so that they might as well get access to the full range of tools.

However, there are also potentially a large number of options that could be specified during a migration so idk if command line flags for each of them is really a reasonable alternative. For example, if someone is on a private network and therefore needs the network symmetric key they're a bit out of luck with just some peer flags. If we make our routing systems more configurable we're similarly out of luck.

cc @Stebalien as this and https://github.com/ipfs/go-ipfs/pull/8064#discussion_r611867631 seem to be related to the comments in the proposal PR (https://github.com/protocol/web3-dev-team/pull/86/files#diff-991744cab1b93840a21929cbf5aa7bc00a37ae4244b5c569df086214e36e47cfR66 and https://github.com/protocol/web3-dev-team/pull/86/files#r610934803)

repo/fsrepo/migrations/ipfsfetcher/ipfsfetcher_test.go
aschmahmann: There are some tests in the integration test package (e.g. https://github.com/ipfs/go-ipfs/blob/master/test/integration/three_legged_cat_test.go) that could be of inspiration here.

If it's a pain to write the test lmk and we can decide on an alternative. Using a flag/env var seems fine (I'm not sure if we have an appropriate one, although perhaps this one https://github.com/ipfs/go-ipfs/blob/ef866a1400b3b2861e5e8b6cc9edc8633b890a0a/test/integration/addcat_test.go#L173 would do), although I don't really trust tests that don't run in CI since they aren't reliably run before merging PRs and are subject to breaking.

@Stebalien lmk if you have any additional thoughts here given you've been thinking about our CI situation a bit, or it's just 👍.

Stebalien: We need to test this in CI but we can use sharness tests. If a test is disabled in CI, it might as well not exist.

repo/fsrepo/migrations/fetcher.go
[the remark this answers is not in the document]
Stebalien: FYI, https://github.com/uber-go/multierr/ nicer in my experience (doesn't matter too much).

cmd/ipfs/daemon.go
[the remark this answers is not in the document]
Stebalien: > An alternative way to do this is to have a --migration-temp-node-config (or better name blush) option that takes a config file to use for the temporary node. This could protect us from a variety of bugs/corner cases as it allows users to configure the temporary node as if it was a regular one.

We _could_, maybe, later, support that for advanced users/distros. That's not going to really help in this case because users will be _very_ confused about why their migration options aren't being applied. Users are going to be less confused if we start a temporary node that ignores their config.

---

I've left a comment on the proposal. IMO, we should just try to read the relevant part of the config and ignore the rest. Downsides:

1. Migrations can't touch the migration part of the config. Honestly, I'm fine with that.
2. If we decide to move the config, things could get... tricky. But trying a set of configs in order of precedence, _just_ to read a small portion probably isn't a huge issue.

We could also do this on the command-line, but it could get icky (as @aschmahmann said, lots of options).

cmd/ipfs/migration.go
```diff
@@ -0,0 +1,160 @@
...
+	ufs := ifaceCore.Unixfs()
+
+	// Add migration files
+	for _, filePath := range binPaths {
+		f, err := os.Open(filePath)
+		if err != nil {
+			return err
+		}
+
+		ipfsPath, err := ufs.Add(ctx, files.NewReaderFile(f), options.Unixfs.Pin(pin))
```
aschmahmann, line 108: This is the thing that doesn't work https://github.com/ipfs/go-ipfs/pull/8064#discussion_r611921297.

For example, if we add the migration binaries using the current defaults for `ipfs add` and then change the defaults in a subsequent release then the swarms won't converge.

aschmahmann: Note: @gammazero if this is a pain to implement for you given time constraints I'm happy to either drop this requirement for the time being or I'll just code up the CAR import/export stuff on my own (likely requires copy-pasting code from the commands package into a function on the IpfsNode and then just calling it).

cmd/ipfs/daemon.go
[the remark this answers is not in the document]
aschmahmann: I think I'm fine with this approach. If we define the migrations config options as described in the proposal and then we try to deserialize the config file (or object) into a struct that only has the migration values that should be fine.

This will mean that unlike the rest of the config which is nominatively typed in that backwards incompatible changes are tied to a repo version, that we are likely to use structural typing for the migrations specific section (although if this proves too hard we can add a field into the migrations struct to track versioning).

This reliance on structural typing means we basically don't have to worry too much about the potential migration config changes for the time being and it's possible we'll end up with a repo/package that contains all the migration configs (like what we used to do with fs-repo-migrations, but really small since it's only for a portion of the config file).

For the time being we can probably punt on dealing with private swarms or people with custom bootstrap nodes since they can:
1. Just use HTTP, like they've been doing so far
2. Download the migrations binary manually, or run the full fs-repo-migrations binary or ipfs-update

If this becomes a problem and we need the sledge hammer of passing in `--migration-temp-node-config` we can do that, but let's delay until we have a user + use case who needs this.

cmd/ipfs/daemon.go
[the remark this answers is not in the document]
gammazero: removed

repo/fsrepo/migrations/ipfsfetcher/ipfsfetcher.go
gammazero: - When migrations are downloaded via IPFS, add the migration archives to IPFS by connecting to the temporary migration node and getting the migrations over IPFS.
- When migrations are downloaded via HTTP, add the migration archive files directly to IPFS.  After talking to @Stebalien he thought it was still worth to add the files directly, even though it might result in a different dag.  If we decide later to not do that, it is a trivial change.

repo/fsrepo/migrations/ipfsfetcher/ipfsfetcher_test.go
gammazero: @aschmahmann I will see what I can do with sharness, but likely this is going to have to become a job for your team if it needs to get done before the upcoming release.

cmd/ipfs/migration.go
gammazero: I went with the approach of connecting to the temporary migration node and then getting the data over IPFS.  Exposing a CAR interface on the node may be a better solution in the long run, and we can revisit that.

repo/fsrepo/migrations/fetcher.go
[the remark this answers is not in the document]
gammazero: daemon was already using https://github.com/hashicorp/go-multierror. 🤷

cmd/ipfs/daemon.go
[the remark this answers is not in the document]
gammazero: I am going to leave the configuration for a later time.  Currently, go-ipfs accepts a couple of optional flags that allow the user to specify how to download the migrations (`--download-migration`) and if/how to keep downloaded migrations (`--keep-migration`).  Even if this information is available from a config file in the future, I think these flags will will still be useful to override that.

gammazero:
Rebased on master to pick up fixed sharness tests.

cmd/ipfs/daemon.go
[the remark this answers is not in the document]
Stebalien: So, I thought this solution would be fine... but I've changed my mind. We have _way_ too many special-purpose daemon flags that should be settings. We're already trying to get rid of the existing daemon flags, I don't want to add more that we'll need to maintain into the future for backwards compatibility.

If we want to be able to override config options, we need a coordinated system to do that but this isn't that.

repo/fsrepo/migrations/ipfsfetcher/ipfsfetcher.go
```diff
@@ -0,0 +1,332 @@
...
+	ipath "github.com/ipfs/interface-go-ipfs-core/path"
+	peer "github.com/libp2p/go-libp2p-core/peer"
+	ma "github.com/multiformats/go-multiaddr"
+)
+
+const (
+	// Default maximum download size
+	defaultFetchLimit = 1024 * 1024 * 512
+
+	tempNodeTcpAddr = "/ip4/127.0.0.1/tcp/40010"
```
aschmahmann, line 34: Any reason this can't be `/ip4/127.0.0.1/tcp/0` we can then get the peer's "real" address by looking at the result of `ipfs.Swarm().LocalAddrs(ctx)`

gammazero: Did not know that was an option.  Nice.

Done.

gammazero:
@aschmahmann If you merge PR #8076, then I can rebase and use CAR files to import downloaded migrations.  I can also do that in a separate PR.  In any case, once #8076 is in, import via CAR will be simple.  Making that change to use CAR files will also enable PR #7658 to do the same.

aschmahmann:
@gammazero yep, that's the idea. That PR seems close to being good, but has a bug causing it to fail sharness at the moment. In any event I wouldn't block on that PR here.

cmd/ipfs/daemon.go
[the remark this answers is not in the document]
gammazero: @Stebalien Done.  Now reading migration setting from config file, and using defaults for any values not present.  Removed CLI flags for specifying migration configuration.

cmd/ipfs/migration.go
```diff
@@ -82,8 +57,13 @@ func getMigrationFetcher(downloadPolicy string, peers string) (migrations.Fetche
 				return nil, errors.New("custom gateway scheme must be http or https")
 			}
 			fetchers = append(fetchers, migrations.NewHttpFetcher(fetchDistPath, u.String(), httpUserAgent, 0))
+		case "":
+			// Ignore empty string
```
aschmahmann, line 61: Why?

cmd/ipfs/migration_test.go
```diff
@@ -56,27 +71,52 @@ func TestGetMigrationFetcher(t *testing.T) {
 	if mf.Len() != 3 {
 		t.Fatal("expected3 fetchers in MultiFetcher")
```
aschmahmann, line 72:
```suggestion
		t.Fatal("expected 3 fetchers in MultiFetcher")
```

repo/fsrepo/migrations/ipfsfetcher/ipfsfetcher.go
```diff
@@ -240,21 +231,22 @@ func (f *IpfsFetcher) startTempNode(peers []peer.AddrInfo) error {
...
-	// Parse peer addresses and asynchronously connect to peers
-	if len(peers) != 0 {
-		connectPeers(ctxIpfsLife, ifaceCore, peers)
+	// Asynchronously connect to peers
+	if len(f.peers) != 0 {
+		go connectPeers(ctxIpfsLife, ipfs, f.peers)
 	}
 
-	a, err := ma.NewMultiaddr(tempNodeTcpAddr)
+	addrs, err := ipfs.Swarm().LocalAddrs(ctx)
```
aschmahmann, line 239: What's the idea behind where the background context is getting used and where the passed in context is getting used?

If this function doesn't complete then the temporary node is useless right (aside from the fact that if MDNS is working then everything is probably fine)?

If we don't get the addresses do we need to error + close the node?

cmd/ipfs/migration.go
gammazero: Because we state that if they leave the DownloadSources empty, that results in the default behavior.  I thought it might be considered reasonable by some to think `[""]` means empty.

repo/fsrepo/migrations/ipfsfetcher/ipfsfetcher.go
gammazero: The background context is used as a mechanism to shutdown the temporary node. Quote from `go-ipfs/core/builder.go`:
>      // Shut down the application if the lifetime context is canceled.
>      // NOTE: we _should_ stop the application by calling `Close()` 
>      // on the process. But we currently manage everything with contexts.                                   

So, the background context is for now shutdown, in the absence of a `node.Close()` method, and the passed-in context is used for cancellation of the current function call.  

True, if the function does not complete, then the node is useless.  However, failure to get the local swarm address only means that the downloaded migrations cannot be fetched from the temp node.... So, right, it does not need to fail out here.

cmd/ipfs/migration_test.go
gammazero: Fixed

aschmahmann (changes requested):

cmd/ipfs/migration.go
```diff
@@ -0,0 +1,292 @@
...
+		}
+	default:
+		return errors.New("Cannot get migrations from unknown fetcher type")
+	}
+
+	return nil
+}
+
+// addMigrationFiles adds the files at paths to IPFS, optionally pinning them
+func addMigrationFiles(ctx context.Context, node *core.IpfsNode, paths []string, pin bool) error {
```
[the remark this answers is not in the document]
aschmahmann: Given that we just merged CAR export on gateways (https://github.com/ipfs/go-ipfs/pull/8111) I feel much less strongly about this and think we can leave it as it is.

What we can do is in a later RC (or next release) switch the HTTP Fetcher to really be a CAR fetcher and then this problem goes away. For the time being we control how the migrations are created so we can just leave the defaults.

cc @Stebalien as a heads up

repo/fsrepo/migrations/fetcher.go
```diff
@@ -45,15 +49,38 @@ func NewMultiFetcher(f ...Fetcher) Fetcher {
...
+	for _, fetcher := range f.fetchers {
+		if err := fetcher.Close(); err != nil {
+			errs = multierror.Append(errs, err)
+		}
+	}
+	return errs
+}
+
+func (f *MultiFetcher) LastUsed() Fetcher {
+	return f.fetchers[0]
```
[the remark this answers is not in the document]
gammazero: Yeah, this is terrible.  Don't know what the plan was here.  This is fixes by never changing the data structure when trying different fetchers.  See next comment.

repo/fsrepo/migrations/fetcher.go
```diff
@@ -45,15 +49,38 @@ func NewMultiFetcher(f ...Fetcher) Fetcher {
...
+func (f *MultiFetcher) Fetch(ctx context.Context, ipfsPath string) (io.ReadCloser, error) {
+	var errs error
+	for i, fetcher := range f.fetchers {
+		rc, err := fetcher.Fetch(ctx, ipfsPath)
 		if err == nil {
-			// Transferred using this fetcher
-			return
+			// Transferred using this fetcher. If not first, swap with first.
+			if i != 0 {
+				f.fetchers[0], f.fetchers[i] = f.fetchers[i], f.fetchers[0]
```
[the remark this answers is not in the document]
gammazero: I do not change the order, note a preference, or mark any fetchers unusable.  Each fetch just tries them in order. When the daemon tries to add migration downloads locally, it also walks through all the fetchers and gets any downlodads from each.

repo/fsrepo/migrations/ipfsfetcher/ipfsfetcher.go
```diff
@@ -0,0 +1,312 @@
...
+	switch proto := u.Scheme; proto {
+	case "ipfs", "ipld", "ipns":
+		ipfsPath = ipath.New(path.Join("/", proto, u.Host, u.Path))
+	default:
+		return nil, fmt.Errorf("%q is not an IPFS path", fetchPath)
+	}
+	return ipfsPath, ipfsPath.IsValid()
+}
+
+func setupPlugins() error {
```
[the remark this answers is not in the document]
gammazero: It does work.  Since the temporary download node is using the same code as the daemon, any plugins that work for one will work for the other. We do not load the plugins twice because the error that results in handled.

Is it necessary?... Probably only for testing.  As long as the IpfsFetcher is used from within a running ipfs daemon, this is not needed as the plugins are already loaded.  If we wanted to use IpfsFetcher outside of the daemon, then we do need to load the plugins, or we get an error like `failed to initialize ephemeral node: unknown datastore type: flatfs`.

Calling `setupPlugins` when plugins are already loaded seems harmless, but should not be necessary if it is only ever called from within the daemon.  Therefore, I moved it out to the test code.

repo/fsrepo/migrations/ipfsfetcher/ipfsfetcher_test.go
```diff
@@ -0,0 +1,103 @@
...
+	"context"
+	"flag"
+	"testing"
+)
+
+var runIpfsTest = flag.Bool("ipfstest", false, "Run IpfsFetcher tests")
+
+func TestIpfsFetcher(t *testing.T) {
+	if !*runIpfsTest {
+		t.Skip("manually-run dev test, use the '-ipfstest' flag to run")
```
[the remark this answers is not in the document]
gammazero: Added `skipUnlessEpic` function that works the same way as in addcat_teat.go

repo/fsrepo/migrations/ipfsfetcher/ipfsfetcher.go
```diff
@@ -0,0 +1,312 @@
...
+	})
+	return f.closeErr
+}
+
+func (f *IpfsFetcher) AddrInfo() peer.AddrInfo {
+	return f.addrInfo
+}
+
+func (f *IpfsFetcher) FetchedPaths() []ipath.Path {
+	return f.fetched
```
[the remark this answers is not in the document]
gammazero: This was meant to be without a mutex, since it must only be read after all fetching is complete.  I have added the mutex for consistency, but if it were needed here then that is really a misuse of this function.

cmd/ipfs/migration.go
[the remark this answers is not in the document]
gammazero: Let me know if you want me to remove the support for adding downloads from HTTP to IPFS.  ...since the CAR download is on the horizon, and less code to worry about.

cmd/ipfs/daemon.go
```diff
@@ -288,8 +292,36 @@ func daemonFunc(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment
...
+			// Create temp directory to store downloaded migration archives
+			migrations.DownloadDirectory, err = ioutil.TempDir("", "migrations")
+			if err != nil {
+				return err
+			}
+			defer func() {
+				if migrations.DownloadDirectory != "" {
+					os.RemoveAll(migrations.DownloadDirectory)
+				}
+			}()
```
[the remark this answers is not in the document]
gammazero: This is done in a deferred function here, so that cleanup will happen if the daemon returns early due to an error.  It is done later after finishing migration work, so that that directory is not sitting around for the lifetime of the daemon and is not left behind if the daemon exits ungracefully.

Added comment to explain.

repo/fsrepo/migrations/ipfsfetcher_test.go
gammazero: This was handled by adding a `skipUnlessEpic` function like as is done here:
 https://github.com/ipfs/go-ipfs/blob/ef866a1400b3b2861e5e8b6cc9edc8633b890a0a/test/integration/addcat_test.go#L173?

repo/fsrepo/migrations/ipfsfetcher/ipfsfetcher.go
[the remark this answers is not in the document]
aschmahmann: Makes sense, we may need to revisit this later

aschmahmann (approved):

PR #8068, opened 2021-04-12
support custom DoH resolvers

vyzo:
Depends on:
- https://github.com/libp2p/go-libp2p/pull/1073
- https://github.com/ipfs/go-namesys/pull/10
- https://github.com/ipfs/interface-go-ipfs-core/pull/69
- https://github.com/ipfs/go-ipfs-config/pull/126


vyzo:
seems like `go-doh-resolver` has an implicit dependency on go 1.16; fixing upstream package.

vyzo:
squashed the gomod dep update commits.

core/node/dns.go
```diff
@@ -0,0 +1,18 @@
...
+	madns "github.com/multiformats/go-multiaddr-dns"
+)
+
+func DNSResolver(cfg *config.Config) (*madns.Resolver, error) {
+	var opts []madns.Option
+	if cfg.DNS.DefaultResolver != "" {
+		opts = append(opts, madns.WithDefaultResolver(doh.NewResolver(cfg.DNS.DefaultResolver)))
+	}
+	for domain, url := range cfg.DNS.CustomResolvers {
+		opts = append(opts, madns.WithDomainResolver(domain, doh.NewResolver(url)))
```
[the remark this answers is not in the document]
vyzo: the `net.Resolver` type doesn't offer a public constructor unfortunately.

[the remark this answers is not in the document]
aschmahmann: Looks like you can create a resolver struct with a custom Dialer https://stackoverflow.com/questions/59889882/specifying-dns-server-for-lookup-in-go.

Either way though we need some syntax for how we describe the addresses as standard vs DoH.

cc @lidel for thoughts. We could just use plain IP address vs HTTPS url. It could be nice to use multiaddrs since they're more uniform, but I seem to recall we ran into problems there last time (perhaps they aren't blockers though?).

core/coreapi/coreapi.go
```diff
@@ -140,6 +142,10 @@ func (api *CoreAPI) PubSub() coreiface.PubSubAPI {
 	return (*PubSubAPI)(api)
 }
 
+func (api *CoreAPI) DNS() coreiface.DNSAPI {
+	return (*DNSAPI)(api)
```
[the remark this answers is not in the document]
vyzo: following sync discussion, I am retracting that.

core/node/dns.go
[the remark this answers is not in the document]
vyzo: Let's go with https urls -- and for now we can just support DoH and error if the url is not an https url.

vyzo:
I'll give it a rebase to resolve the conflicts in `go.{mod,sum}`

vyzo:
gave it  a rebase, resolved the conflict, and squashed a few commits.

core/node/dns.go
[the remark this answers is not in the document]
vyzo: Added check for urls starting with `https://` before creating a DoH resolver.
When we have more resolvers or decide we want to use something else, we can easily add support by checking the url.

core/node/groups.go
```diff
@@ -281,6 +283,7 @@ func Online(bcfg *BuildCfg, cfg *config.Config) fx.Option {
 func Offline(cfg *config.Config) fx.Option {
 	return fx.Options(
 		fx.Provide(offline.Exchange),
+		fx.Provide(DNSResolver),
```
aschmahmann, line 286: This preserves existing behavior. @Stebalien Is this the correct behavior, or would we want some `nilDNSResolver` here?

core/node/dns.go
```diff
@@ -0,0 +1,27 @@
...
+
+	config "github.com/ipfs/go-ipfs-config"
+	doh "github.com/libp2p/go-doh-resolver"
+	madns "github.com/multiformats/go-multiaddr-dns"
+)
+
+func DNSResolver(cfg *config.Config) (*madns.Resolver, error) {
+	var opts []madns.Option
+	if url := cfg.DNS.DefaultResolver; url != "" {
+		if !strings.HasPrefix(url, "https://") {
```
aschmahmann, line 15: @lidel do we currently need to add support for non-DoH resolvers aside from using the built-in Go (i.e. OS-based) resolver as the default resolver?

core/node/groups.go
Stebalien: Yes, we should use the `nilDNSResolver`. I'd consider this a bug fix.

Stebalien: But we can also punt to a second PR.

Stebalien:
This needs some config documentation (`config.md`). Especially, how do we format DoH addresses?

Related: consider an upgrade path to DNS over TLS, and maybe even DNS over libp2p.

vyzo:
sure, I'll add some documentation.

Upgrade path could be very easy by using the appropriate url scheme -- eg
`tls:` or `libp2p:`.

core/node/dns.go
lidel: No, we only need DoH. Other things like [DoT](https://en.wikipedia.org/wiki/DNS_over_TLS) can be added if needed, but not needed atm.

core/node/dns.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	"fmt"
+	"strings"
+
+	config "github.com/ipfs/go-ipfs-config"
+	doh "github.com/libp2p/go-doh-resolver"
+	madns "github.com/multiformats/go-multiaddr-dns"
+)
+
+func DNSResolver(cfg *config.Config) (*madns.Resolver, error) {
```
lidel, line 12: Sidenote regarding backward-compatibility: 
I believe we should include DoH for ENS as an implicit entry here.

- We want to remove hardcoded resolver hack from https://github.com/ipfs/go-ipfs/pull/6448  and https://github.com/ipfs/go-namesys/pull/8
- At the same time, we can't break resolution of `.eth` names in default config at this point.
  - We also don't want to hardcode it in https://github.com/ipfs/go-ipfs-config/ as that is noisy, and makes it hard to migrate default like this in the future. 
  - User should be able to override the default (implicit) ENS resolver  by adding an explicit `eth.` key to the config,

I believe the least painful way is to `append` an implicit DoH resolver ( `eth.` → `https://eth.link/dns-query`) to the list if the config is missing an explicit key for `eth.`

lidel (changes requested):
- We most likely will  accept Multiadrs and URLs for DNS resolver endpoints,
  but for now only `https://` URL with DoH endpoint will be used in practice (Brave, ENS)
- I requested config changes in https://github.com/ipfs/go-ipfs-config/pull/126#discussion_r611954154 / https://github.com/ipfs/go-ipfs-config/pull/126#pullrequestreview-633965792, it also includes example which we could re-use in docs.

core/node/dns.go
aschmahmann: > I believe the least painful way is to append an implicit DoH resolver ( eth. → https://eth.link/dns-query) to the list if the config is missing an explicit key for eth

Do you mean add it into the config file (i.e. via a migration), or add it in the code that creates the resolver?

If via the config file, what if the user just wants to use a single resolver for `.` and so removes `eth.`, will we add it back in?

If we want `ipfs init` to always result in a config + binary that has a DoH binding specifically for `eth.` (as we do now) then we're going to need to some code somewhere with that mapping.

core/node/groups.go
vyzo: It's pretty easy to do, so I can do it here.

core/node/dns.go
vyzo: Do we have to do this in this pr? Sounds like we might want to do in a subsequent one.

vyzo: Note that I am fine with adding an implicit entry for `eth.`.

vyzo: updated.

core/node/groups.go
vyzo: done.

core/node/dns.go
lidel: @aschmahmann I meant code that creates resolver. We don't want to pollute configs with brittle default that can change over  time.  
@vyzo implemented it the way I imagined.

core/node/dns.go
```diff
@@ -0,0 +1,58 @@
...
+		return nil, fmt.Errorf("invalid resolver url: %s", url)
+	}
+
+	return doh.NewResolver(url), nil
+}
+
+func DNSResolver(cfg *config.Config) (*madns.Resolver, error) {
+	var opts []madns.Option
+
+	hasEth := false
```
lidel, line 27: Can we make this more generic and base  implicit defaults on a map that can grow/change over time?

```golang
map[string]string{
	"eth.": "https://eth.resolver.cloudflare-eth.com/dns-query",
	"crypto.": "https://eth.resolver.cloudflare-eth.com/dns-query",
}
```

I've added `.crypto` from [UnstoppableDomains](https://unstoppabledomains.com/)  because `eth.link` uses Cloudflare behind the scenes anyway :man_shrugging:  and Opera/Brave will support UD names out of the box too.

My thinking here is to use an endpoint that is as generic and "trusted" as possible, to minimize DNS leak caused by the initial clear text lookup for implicit DoH endpoint.

docs/config.md
```diff
@@ -713,14 +716,20 @@ Below is a list of the most common public gateway setups.
...
+   - **X-Forwarded-Proto:** if you run go-ipfs behind a reverse proxy that provides TLS, make it add a `X-Forwarded-Proto: https` HTTP header to ensure users are redirected to `https://`, not `http://`. It will also ensure DNSLink names are inlined to fit in a single DNS label, so they work fine with a wildcart TLS cert ([details](https://github.com/ipfs/in-web-browsers/issues/169)). The NGINX directive is `proxy_set_header X-Forwarded-Proto "https";`.:
+  
+     `http://dweb.link/ipfs/{cid}` → `https://{cid}.ipfs.dweb.link`
+     
+     `http://dweb.link/ipns/your-dnslink.site.example.com` → `https://your--dnslink-site-example-com.ipfs.dweb.link`
+     
+   - **X-Forwarded-Host:** we also support `X-Forwarded-Host: example.com` if you want to override subdomain gateway host from the original request:
    
+     `http://dweb.link/ipfs/{cid}` → `http://{cid}.ipfs.example.com`
+
```
lidel, line 732: @vyzo I noticed your editor removed trailing spaces in https://github.com/ipfs/go-ipfs/pull/8068/commits/c042e3a773b35d5dd2df4760b45ba3973a64d7b7, which mangled some paragraphs together. 
Don't worry about it, I fixed it in https://github.com/ipfs/go-ipfs/pull/8068/commits/7bffd06b8f512c701af85a739849880b804ad325, this is  just fyi :)

lidel (changes requested):
This is taking a nice shape :+1: 
I tweaked docs + left some inline comments  regarding implicit defaults.
(I believe we should include implicit defaults in this PR, so we can then simply remove special-casing of  `eth` in go-namesys and be done with it)

docs/config.md
```diff
@@ -1423,3 +1425,32 @@ other IPFS and libp2p implementations. Unlike Yamux:
...
+    "Resolvers": {
+      "eth.": "https://different-ens.example.net/dns-query",
+      "crypto.": "https://unstoppablesomething.example.com/dns-query",
+      "libre.": "https://ns1.iriseden.fr/dns-query",
+      ".": "https://doh-ch.blahdns.com:4443/dns-query"
+    }
+  }
+```
+
+Default: `null`
```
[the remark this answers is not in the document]
vyzo: ah ok, I'll set it to {}.

core/node/dns.go
vyzo: Sure, that's easy to do.

docs/config.md
vyzo: Ugh, sorry, I have emacs set to `delete-trailing-whitespace` by default.

core/node/dns.go
vyzo: you probably meant to use `https://resolver.unstoppable.io/dns-query` for the `crypto.` resolver in the defaults -- I'll use that one.

vyzo: oh it's the same thing, nm.

docs/config.md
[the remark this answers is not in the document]
vyzo: done.

core/node/dns.go
vyzo: done.

lidel (changes requested):
Took this for a spin and works as expected :rocket: 
Only remaining asks is to add DNS record cache and make config bit more flexible (details below and inline)

--- 

### DNS Cache

I think we need to add cache – DoH lookups are too expensive (~1 second each) for this to be useful for loading websites (each subresource would be slowed down by this).  I filled  https://github.com/multiformats/go-multiaddr-dns/issues/28 for this, because we should have a global cache, and that seems like a good home for it.

### Config

- it should be possible to disable implicit default by setting value to "" (we need this, so people can run ENS resolver on localhost)
- `{}` needs more work – details inline

docs/config.md
```diff
@@ -1459,6 +1459,6 @@ Example:
 }
 ```
 
-Default: `null`
+Default: `{}`
```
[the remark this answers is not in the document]
vyzo: ah ok, will do -- i somehow thought they were autogenerated.

vyzo:
DoH resolver cache implementation in https://github.com/libp2p/go-doh-resolver/pull/3

vyzo:
there is a failing sharness test, not sure if it is related:
```
130 - request for example.com/ipns/{fqdn} with X-Forwarded-Proto redirects to TLS-safe label in subdomain - sharnessLinux.t0114-gateway-subdomains

curl -H "Host: example.com" -H "X-Forwarded-Proto: https" -sD - "http://127.0.0.1:42979/ipns/en.wikipedia-on-ipfs.org/wiki" > response && test_should_contain "Location: https://en-wikipedia--on--ipfs-org.ipns.example.com/wiki" response
```

vyzo:
Updated deps to `go-libp2p@master` and `go-ipfs-config@master` and squashed the gomod commits.

vyzo:
Updated for the changes in the namesys constructor api.

aschmahmann:
Looks like there's a failing sharness test here

```
130 - request for example.com/ipns/{fqdn} with X-Forwarded-Proto redirects to TLS-safe label in subdomain - sharnessLinux.t0114-gateway-subdomains

curl -H "Host: example.com" -H "X-Forwarded-Proto: https" -sD - "http://127.0.0.1:46881/ipns/en.wikipedia-on-ipfs.org/wiki" > response && test_should_contain "Location: https://en-wikipedia--on--ipfs-org.ipns.example.com/wiki" response
```

lidel (changes requested):
_Ain't nobody pass my sharness tests_ :trollface: 

jk, @vyzo @aschmahmann  not sure if this helps, but I was able to narrow down the regression:

### DNSLink lookup is broken in offline mode 

`./t0114-gateway-subdomains.sh` runs IPNS and DNSLink tests in offline mode, and that fails because something changed and DNSLink now can't be resolved when node runs in `ipfs daemon --offline`:

```
$  ipfs resolve -r /ipns/en.wikipedia-on-ipfs.org/wiki
ipfs resolve -r /ipns/en.wikipedia-on-ipfs.org/wiki: could not resolve name: "en.wikipedia-on-ipfs.org" is missing a DNSLink record (https://docs.ipfs.io/concepts/dnslink/)
```

Did we disable DNS lookups in offline mode on purpose?
This  feels like a big **breaking change**: effectively one can no longer do quick CID lookup like `ipfs resolve -r /ipns/en.wikipedia-on-ipfs.org/wiki`  without running node in online mode.

The way I see it, if someone wants to disable DNS lookups they should set "." to empty string or "http://fake-resolver". 

### IPNS lookup is also broken in offline mode

Not sure if related to previous one, but when I run client without daemon or with `ipfs daemon --offline` then IPNS is being interpreted(?) as DNS name :upside_down_face: :

```

# ipfs daemon --offline
$ ipfs resolve -r /ipns/k51qzi5uqu5dgutdk6i1ynyzgkqngpha5xpgia3a5qqp4jsh0u4csozksxel2r
Error: could not resolve name: "k51qzi5uqu5dgutdk6i1ynyzgkqngpha5xpgia3a5qqp4jsh0u4csozksxel2r" is missing a DNSLink record (https://docs.ipfs.io/concepts/dnslink/)

# ipfs daemon (online) works ok:
$ ipfs resolve -r /ipns/k51qzi5uqu5dgutdk6i1ynyzgkqngpha5xpgia3a5qqp4jsh0u4csozksxel2r
/ipfs/QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR
```

aschmahmann:
> Not sure if related to previous one, but when I run client without daemon or with ipfs daemon --offline then IPNS is being interpreted(?) as DNS name 🙃 :

I think https://github.com/ipfs/go-namesys/pull/9 was just incorrect in that it interprets `ErrResolveFailed` as a DNSLink error even in non-DNSLink cases.

> DNSLink lookup is broken in offline mode... Did we disable DNS lookups in offline mode on purpose?

Yes. I unresolved the discussion so you can see it and add any comments/questions there https://github.com/ipfs/go-ipfs/pull/8068#discussion_r611898939. I'd really expect an offline mode to not make network queries at all, but it is a behavior change.

vyzo:
Undid the `NilDNSResolvre` contraption and this resolves the regression.

core/node/groups.go
vyzo: undone, as it is a behaviour change that leads to a regression; we can easily consider it in a follow up pr.

vyzo:
Updated go-namesys@master and squashed the gomod commits.

@Stebalien @aschmahmann @lidel this should be ready for merge now!

aschmahmann (approved):
LGTM, I'd prefer to only reference released versions of subpackages in go-ipfs though. We can wait a bit and see if we can get in the releases of the subpackages soon and if not just make a couple back-to-back small releases.

Stebalien:
Updated to a released go-libp2p & go-ipfs-config.

PR #8069, opened 2021-04-12
fix(gw): remove hardcoded hostnames

### LINKED ISSUE #7317: Even when explicitely set, the subdomain gateway config still include ipfs.io, gateway.ipfs.io and dweb.link
go-ipfs v0.5.1

https://github.com/ipfs/go-ipfs/blob/master/core/corehttp/hostname.go#L37-L42

It seems ok to me to implicitly accept `localhost` when no public gateway is defined but including the Protocol Labs gateway as well seems too much. At least when explicitly defined, those should not be included as well.

lidel:
This PR closes #7317 by removing hard-coded PL hostnames from the default config, making the localhost the only implicit gateway hostname defined out-of-the-box.

@mburns to double-check if `dweb.link` uses explicit [`Gateway.PublicGateways`](https://github.com/ipfs/go-ipfs/blob/master/docs/config.md#gatewaypublicgateways-usesubdomains) config so it is not impacted by this change.

### DIFF
--- a/core/corehttp/hostname.go
+++ b/core/corehttp/hostname.go
@@ -23,12 +23,7 @@ import (
 	nsopts "github.com/ipfs/interface-go-ipfs-core/options/namesys"
 )
 
-var defaultPaths = []string{"/ipfs/", "/ipns/", "/api/", "/p2p/", "/version"}
-
-var pathGatewaySpec = &config.GatewaySpec{
-	Paths:         defaultPaths,
-	UseSubdomains: false,
-}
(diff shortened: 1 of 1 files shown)


MichaelMure (approved):

mburns:
👍  to this change

Bifrost Infra explicitly managed the `dweb.link` and `(gateway.)ipfs.io` and their equivalent Staging counterparts.

Stebalien (approved):

PR #8070, opened 2021-04-13
experimental provider system

aschmahmann:
Adds an experimental provider system that makes use of the experimental dht client.

Depends on https://github.com/ipfs/go-ipfs-provider/pull/34, links to #8061 

The point of this system is to be able to quickly provide by being sympathetic to how Kademlia works. For the time being it tries to get records into the DHT very quickly at the expense of having a spike in resource usage during that time.

It also comes with a few extras that have been added for sanity purposes during testing/evaluation and that may be useful in production including:
- A stat command to track how many provider records we are emitting (only for the experimental system)
- Tracking when we last provided our records so that rebooting the node doesn't cause us to automatically start reproviding unless we need to

### DIFF
--- a/core/commands/stat_provide.go
+++ b/core/commands/stat_provide.go
@@ -1,0 +1,51 @@
+package commands
+
+import (
+	"fmt"
+
+	cmds "github.com/ipfs/go-ipfs-cmds"
+	"github.com/ipfs/go-ipfs/core/commands/cmdenv"
+
+	"github.com/ipfs/go-ipfs-provider/batched"
+)
+
+var statProvideCmd = &cmds.Command{
+	Helptext: cmds.HelpText{
+		Tagline: "Returns statistics about the node's (re)provider system.",
+		ShortDescription: `
+Returns statistics about the content the node is advertising.
+
+This interface is not stable and may change from release to release.
+`,
+	},
+	Arguments: []cmds.Argument{},
+	Options:   []cmds.Option{},
+	Run: func(req *cmds.Request, res cmds.ResponseEmitter, env cmds.Environment) error {
+		nd, err := cmdenv.GetNode(env)
+		if err != nil {
+			return err
+		}
+
+		if !nd.IsOnline {
+			return ErrNotOnline
+		}
+
+		sys, ok := nd.Provider.(*batched.BatchProvidingSystem)
+		if !ok {
+			return fmt.Errorf("can only return stats if Experimental.AcceleratedDHTClient is enabled")
+		}
+
+		stats, err := sys.Stat(req.Context)
+		if err != nil {
+			return err
+		}
+
+		if err := res.Emit(stats); err != nil {
+			return err
+		}
+
+		return nil
+	},
+	Encoders: cmds.EncoderMap{},
+	Type:     batched.BatchedProviderStats{},
+}
--- a/core/node/provider.go
+++ b/core/node/provider.go
@@ -7,13 +7,16 @@ import (
 
 	"github.com/ipfs/go-ipfs-pinner"
 	"github.com/ipfs/go-ipfs-provider"
+	"github.com/ipfs/go-ipfs-provider/batched"
 	q "github.com/ipfs/go-ipfs-provider/queue"
 	"github.com/ipfs/go-ipfs-provider/simple"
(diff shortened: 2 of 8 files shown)


core/commands/stat_provide.go
```diff
@@ -0,0 +1,51 @@
...
+			return err
+		}
+
+		if !nd.IsOnline {
+			return ErrNotOnline
+		}
+
+		sys, ok := nd.Provider.(*batched.BatchProvidingSystem)
+		if !ok {
+			return fmt.Errorf("can only return stats if the experimental DHT client is enabled")
```
lidel, line 35: Below will make it easier to google and find `experimental-features.md`  :)
```suggestion
			return fmt.Errorf("can only return stats if Experimental.AcceleratedDHTClient is enabled")
```

lidel (approved):
The way experiment is introduced LGTM, some small style tweaks inline.

core/commands/stat_provide.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/ipfs/go-ipfs-provider/batched"
+)
+
+var statProvideCmd = &cmds.Command{
+	Helptext: cmds.HelpText{
+		Tagline: "Returns statistics about the node's (re)provider system.",
+		ShortDescription: `
+Returns statistics about the content the node is advertising.
+
+This interface is not stable and may change from release to release.
```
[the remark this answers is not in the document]
aschmahmann: Not sure, probably? Maybe start an issue in go-ipfs for tracking and collecting feedback before starting on the PR.

core/node/provider.go
```diff
@@ -59,29 +62,78 @@ func SimpleProviderSys(isOnline bool) interface{} {
...
+
+// BatchedProviderSys creates new provider system
+func BatchedProviderSys(isOnline bool, reprovideInterval string) interface{} {
+	return func(lc fx.Lifecycle, cr libp2p.BaseIpfsRouting, q *q.Queue, keyProvider simple.KeyChanFunc, repo repo.Repo) (provider.System, error) {
+		r, ok := (cr).(provideMany)
+		if !ok {
+			return nil, fmt.Errorf("BatchedProviderSys requires a content router that supports provideMany")
+		}
+
+		reproviderInterval := kReprovideFrequency
```
gammazero, line 78: minor: Please rename `reproviderInterval` to `reprovideDuration`.  Otherwise, it is difficult to distinguish from the `reprovideInterval` parameter.

docs/experimental-features.md
```diff
@@ -547,3 +548,56 @@ ipfs config --json Experimental.GraphsyncEnabled true
...
+     done in rounds
+   - Users who previously had a lot of content but were unable to advertise it on the network will see an increase in
+     egress bandwidth as their nodes start to advertise all of their CIDs into the network. If you have lots of data
+     entering your node that you don't want to advertise consider using [Reprovider Strategies](config.md#reproviderstrategy)
+     to reduce the number of CIDs that you are reproviding. Similarly, if you are running a node that deals mostly with
+     short-lived temporary data (e.g. you use a separate node for ingesting data then for storing and serving it) then
+     you may benefit from using [Strategic Providing](#strategic-providing) to prevent advertising of data that you
+     ultimately will not have.
+2. Currently, the DHT is not usable for queries for the first 5-10 minutes of operation as the routing table is being
+prepared, this means operations like searching the DHT for particular peers or content will not work either
```
gammazero, line 588: >"prepared, this"

"prepared.  This"

gammazero (approved):
LGTM, with a couple very minor comments

docs/experimental-features.md
aschmahmann: Thanks, note that for docs correction like these using GitHub's "insert a suggestion" functionality (the icon with the +- on it) is useful since you can just make a commit fixing the change from GitHub.

core/node/provider.go
aschmahmann: thx for the catch, that's definitely confusing to look at. I changed it to `reprovideIntervalDuration` since `reprovideDuration` sounds like how long a reprovide takes as opposed to the gap between reprovides. Maybe it's a little bit of a long name, but 🤷.

PR #8071, opened 2021-04-13
feat: support DNSLink on non-ICANN DNS names

### LINKED ISSUE #8060: Support non-ICANN DNSLink names
> I wrote long elaborate issue about this and then lost it due to github bug, so all you get is this lousy TLDR :( 

Configurable DNS resolver is coming (https://github.com/protocol/web3-dev-team/pull/42, https://github.com/multiformats/go-multiaddr-dns/pull/26), we need to fully decouple our stack from ICANN's TLDs:

- remove reliance on TLD safelist from `jbenet/go-is-domain`  as suggested in  https://github.com/jbenet/go-is-domain/issues/16#issuecomment-740967950 and  https://github.com/jbenet/go-is-domain/pull/19#pullrequestreview-584687951

- switch DNSLink checks to something equivalent to `net.isDomainName` (sadly private). The check should only  care about    the name being compliant with RFC 1035 and 3696, and is not concerned  if TLD was blessed by ICANN or some other authority 
  - potential candidate: `dns.IsDomainName(s string) (labels int, ok bool) ` from `"github.com/miekg/dns"`

mycorrhizas:
Closes #8060

I switched `IsDomain()` from `jbenet/go-is-domain`  to `IsDomainName` from `github.com/miekg/dns` as suggested.

I had to change one of the tests for everything to pass. This may be an error on my side. However, I could not understand why `http://k51qzi5uqu5di608geewp3nqkg0bpujoasmka7ftkyxgcm3fh1aroup0gsdrna.ipns.localhost/` was expected?

### DIFF
--- a/core/corehttp/hostname.go
+++ b/core/corehttp/hostname.go
@@ -349,15 +349,27 @@ func knownSubdomainDetails(hostname string, knownGateways gatewayHosts) (gw *con
 	return nil, "", "", "", false
 }
 
+// isDomainNameAndNotPeerID returns bool if string looks like a valid DNS name AND is not a PeerID
+func isDomainNameAndNotPeerID(hostname string) bool {
+	if len(hostname) == 0 {
+		return false
+	}
+	if _, err := peer.Decode(hostname); err == nil {
+		return false
+	}
+	_, ok := dns.IsDomainName(hostname)
+	return ok
+}
+
 // isDNSLinkName returns bool if a valid DNS TXT record exist for provided host
 func isDNSLinkName(ctx context.Context, ipfs iface.CoreAPI, host string) bool {
-	fqdn := stripPort(host)
+	dnslinkName := stripPort(host)
 
-	if _, ok := dns.IsDomainName(fqdn); !ok && len(fqdn) == 0 {
+	if !isDomainNameAndNotPeerID(dnslinkName) {
 		return false
 	}
 
-	name := "/ipns/" + fqdn
+	name := "/ipns/" + dnslinkName
 	// check if DNSLink exists
 	depth := options.Name.ResolveOption(nsopts.Depth(1))
 	_, err := ipfs.Name().Resolve(ctx, name, depth)
@@ -476,7 +488,7 @@ func toSubdomainURL(hostname, path string, r *http.Request, ipfs iface.CoreAPI) 
 	}
 
 	// Normalize problematic PeerIDs (eg. ed25519+identity) to CID representation
-	if _, ok := dns.IsDomainName(rootID); !ok && isPeerIDNamespace(ns) {
+	if isPeerIDNamespace(ns) && !isDomainNameAndNotPeerID(rootID) {
 		peerID, err := peer.Decode(rootID)
 		// Note: PeerID CIDv1 with protobuf multicodec will fail, but we fix it
 		// in the next block
--- a/core/corehttp/hostname.go
+++ b/core/corehttp/hostname.go
@@ -13,8 +13,9 @@ import (
 	core "github.com/ipfs/go-ipfs/core"
 	coreapi "github.com/ipfs/go-ipfs/core/coreapi"
 	namesys "github.com/ipfs/go-namesys"
-	isd "github.com/jbenet/go-is-domain"
 	"github.com/libp2p/go-libp2p-core/peer"
+	dns "github.com/miekg/dns"
+
 	mbase "github.com/multiformats/go-multibase"
 
 	config "github.com/ipfs/go-ipfs-config"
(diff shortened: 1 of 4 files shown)


core/corehttp/hostname.go
```diff
@@ -13,8 +13,9 @@ import (
 	core "github.com/ipfs/go-ipfs/core"
 	coreapi "github.com/ipfs/go-ipfs/core/coreapi"
 	namesys "github.com/ipfs/go-namesys"
-	isd "github.com/jbenet/go-is-domain"
```
Angie3120, line 16: isd "github.com/jbenet/go-is-domain

Angie3120 (changes requested):

core/corehttp/hostname.go
Angie3120: Just making a hange

Angie3120 (changes requested):

core/corehttp/hostname_test.go
```diff
@@ -55,7 +55,7 @@ func TestToSubdomainURL(t *testing.T) {
 		{httpRequest, "localhost", "/ipns/QmY3hE8xgFCjGcz6PHgnvJz5HZi1BaKRfPkn1ghZUcYMjD", "http://k2k4r8n0flx3ra0y5dr8fmyvwbzy3eiztmtq6th694k5a3rznayp3e4o.ipns.localhost/", nil},
 		{httpRequest, "localhost", "/ipns/bafybeickencdqw37dpz3ha36ewrh4undfjt2do52chtcky4rxkj447qhdm", "http://k2k4r8l9ja7hkzynavdqup76ou46tnvuaqegbd04a4o1mpbsey0meucb.ipns.localhost/", nil},
 		// PeerID: ed25519+identity multihash → CIDv1Base36
-		{httpRequest, "localhost", "/ipns/12D3KooWFB51PRY9BxcXSH6khFXw1BZeszeLDy7C8GciskqCTZn5", "http://k51qzi5uqu5di608geewp3nqkg0bpujoasmka7ftkyxgcm3fh1aroup0gsdrna.ipns.localhost/", nil},
+		{httpRequest, "localhost", "/ipns/12D3KooWFB51PRY9BxcXSH6khFXw1BZeszeLDy7C8GciskqCTZn5", "http://12D3KooWFB51PRY9BxcXSH6khFXw1BZeszeLDy7C8GciskqCTZn5.ipns.localhost/", nil},
```
lidel, line 58: This is invalid, case-sensitive identifiers are not allowed in DNS labels, we need to make it work without changing the test:

```suggestion
		{httpRequest, "localhost", "/ipns/12D3KooWFB51PRY9BxcXSH6khFXw1BZeszeLDy7C8GciskqCTZn5", "http://k51qzi5uqu5di608geewp3nqkg0bpujoasmka7ftkyxgcm3fh1aroup0gsdrna.ipns.localhost/", nil},
```

core/corehttp/hostname.go
```diff
@@ -481,7 +484,7 @@ func toSubdomainURL(hostname, path string, r *http.Request, ipfs iface.CoreAPI)
 	}
 
 	// Normalize problematic PeerIDs (eg. ed25519+identity) to CID representation
-	if isPeerIDNamespace(ns) && !isd.IsDomain(rootID) {
+	if _, ok := dns.IsDomainName(rootID); !ok && isPeerIDNamespace(ns) {
```
lidel, line 487: This won't work, because IsDomainName(peerIDs) returns false-positive.

lidel (changes requested):
Thank you @mycorrhizas! 

I think we need to make this smarter and replace `IsDomainName` with a helper function that uses `IsDomainName` internally, but first correctly detects valid PeerID and returns "false" before calling `IsDomainName`. 

Otherwise, we get false-positives when a valid  PeerID is used in subdomain (thats why your test did not pass).

Samples you can use for testing:

- `QmY3hE8xgFCjGcz6PHgnvJz5HZi1BaKRfPkn1ghZUcYMjD` (legacy peerid multihash  in base58btc)
- `12D3KooWFB51PRY9BxcXSH6khFXw1BZeszeLDy7C8GciskqCTZn5` (new peerid multihash in base58btc)
- `k2k4r8l9ja7hkzynavdqup76ou46tnvuaqegbd04a4o1mpbsey0meucb` (new peerid as CIDv1 in base36)

lmk if anything is unclear

lidel:
Ok, things got a bit more complicated because we have various PRs  in flight, all related to DNS and DNSLink.

I think we need some things to land first, to unblock this PR:

- go-namesys release 
  - with  https://github.com/ipfs/go-namesys/pull/10 and https://github.com/ipfs/go-namesys/pull/12
  - and without `github.com/jbenet/go-is-domain`  https://github.com/ipfs/go-namesys/pull/13
- support for configurable DNS servers (https://github.com/ipfs/go-ipfs/issues/6532)

mycorrhizas:
@lidel apologies for the late reply 

thanks for explaining and resolving the issue with `isDomainNameAndNotPeerID`?

lidel (approved):
LGTM, but as I did some work here this needs more eyes.

@aschmahmann as noted during today's  triage, this is ready for your final :eyes: 

FYSA https://github.com/ipfs/go-namesys/pull/13 already landed in namesys, so this is the final piece of the puzzle.
We may want to merge this AFTER https://github.com/ipfs/go-ipfs/pull/8068 as it should be easier to rebase this one if needed, than the other way around.

aschmahmann (approved):
LGTM, we'll merge after the DoH PR

Issue #7968, opened 2021-03-10
Webgateway: sets 'last-modified' header to now() 


RubenKelevra:
While the web gateway sets the CID as etag in the http header which allows to cache resources properly, the last-modified tag is always the current time.

This doesn't really match what the tag is supposed to reflect: The last time the ressource changed.

Since IPFS has no way of knowing when this occurred, wouldn't it be more correct to not send a last-modified tag at all?

RFC2616 doesn't require a server to have this tag in the response header:

```
HTTP/1.1 servers SHOULD send Last-Modified whenever feasible.
```

Stebalien:
We set the modification time to "infinitely in the past" when using /ipfs for better caching. However, for /ipns, we set it to the current time to make sure the latest file is always "newer".

But yeah, we could probably just not set the modification time for /ipns.

lidel:
The gateway was refactored into the `boxo/gateway` in https://github.com/ipfs/boxo library. 

For immutable `/ipfs/` content, `Last-Modified` is no longer sent.

For mutable `/ipns/` content, `time.Now()` is used only when no better value is available (e.g., UnixFS 1.5 mtime or IPNS record timestamp).

Closing as I believe this was fixed since at least Kubo v0.24.0. If you still see incorrect Last-Modified headers in Kubo 0.40.0-rc1 or later, please open a new issue.

Issue #1818, opened 2015-10-09
max-age, ETag for /ipns requests


ion1:
The gateway does not seem to provide a Cache-Control/max-age header or an ETag for `/ipns` URLs.

It might be useful to assume *some* reasonable non-zero max-age for content on IPNS. A minute? Fifteen seconds?

It would also be good to serve an ETag corresponding to whatever IPFS hash the IPNS path resolved to.

jbenet:
> It might be useful to assume some reasonable non-zero max-age for content
on IPNS. A minute? Fifteen seconds?

It could be determined by the IPNS record -- some will have a calculable
TTL.

>

lgierth:
Until then, we could set `must-revalidate` and a quasi infinite max-age

ion1:
I’m not sure if I have understood this correctly, but I’m under the impression that `must-revalidate` only applies if the content has already expired. http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.9.4

lgierth:
You're right, thanks -- so that means `must-revalidate` + `max-age=0` = `cache-indefinitely-unless-updated`

lgierth:
Note that for `/ipfs/`, we can still set a quasi-infinite max-age.

ion1:
It seems to me that `must-revalidate` means that given an object that expired according to `max-age`, a cache must return an error instead of a stale copy if it fails to connect to the origin server. I’m not sure `must-revalidate` is of any use here. Unless I’m misinterpreting it, of course.

ion1:
DNS lets you say explicitly for how long a record can be cached. Perhaps we should do that with IPNS names? `ipfs name publish --ttl=3h` which would be taken advantage of by IPFS nodes who have resolved the name as well as by the gateway in the Cache-Control header.

ion1:
It was mentioned on IRC that `resolve` does not provide TTL information to be used for `max-age`.

`resolveOnce` could be changed to return a TTL value. `resolve` could start at infinity-ish and compute the minimum of the previous TTL and what `resolveOnce` returned at each iteration.

Given that the DNS lookup functions in [the net package for Go](https://golang.org/pkg/net/) do not seem to provide TTL information at the moment, the DNS `resolveOnce` could return a reasonable guess like one minute. If a DNS record points to `/ipns/<keyhash>`, that IPNS record can still decrement the TTL according to the iteration above if a low cache time is desired.

ion1:
[...]

PR #8074, opened 2021-04-14
disable-last-modified-behaviour-ipns-routes

mycorrhizas:
[7968](https://github.com/ipfs/go-ipfs/issues/7968) `Webgateway: sets 'last-modified' header to now() `

@RubenKelevra @Stebalien 

As per the [documentation](https://golang.org/pkg/net/http/#ServeContent):

`If modtime is not zero time, ServeContent includes it in the Last-Modified header in the response. If the request includes an If-Modified-Since header, ServeContent uses modtime to determine whether the content needs to be sent at all.`

To disable this behaviour, pass a "zero" date to ServeContent.

This change only effects `ipns` routes not `ipfs` routes.

### DIFF
--- a/core/corehttp/gateway_handler.go
+++ b/core/corehttp/gateway_handler.go
@@ -299,8 +299,9 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
 	// set these headers _after_ the error, for we may just not have it
 	// and don't want the client to cache a 500 response...
 	// and only if it's /ipfs!
-	// TODO: break this out when we split /ipfs /ipns routes.
-	modtime := time.Now()
+
+	// ipns routes do not require LastModified header, disable by passing zero date to ServeContent
+	modtime := time.Time{}
 
 	if f, ok := dr.(files.File); ok {
 		if strings.HasPrefix(urlPath, ipfsPathPrefix) {


RubenKelevra:
Question: Don't we already cache the ipns record and "know" when we last updated it? 🤔

Shouldn't we use this value instead of sending a "0"?

mycorrhizas:
@RubenKelevra 

Do you know where the `ipns` record is cached? 

If so, it would make sense to use this value instead of "0".

However, if it's hard to extract, it might be best to just omit the header with "0".

RubenKelevra:
@mycorrhizas nope. I only remember the setting in the config to specify the size. Would have to look it up where the value is used.

lidel (changes requested):
I've added sharness tests for HTTP headers related to caching but think this PR needs more work before can be safely merged.

Namely,  I am worried that removal of `Last-Modified`  without replacing it with an explicit `Cache-Control` will introduce performance regression around HTTP caching. See my comment about browsers using `Last-Modified` as weak signal for how long to cache page: https://github.com/ipfs/go-ipfs/issues/1818#issuecomment-450866591


##   `Cache-Control` based on `/ipns/` TTL

Ideally, `/ipns/*` should return `Cache-Control` based on TTL from IPNS record or DNSLink (TTL of a TXT record on DNS).   


@aschmahmann   to resolve this,  we need to extend the [namesys interface](https://github.com/ipfs/go-namesys/blob/master/interface.go) to return TTL of both record types in addition to `Path`. 

I think we decided that is out of scope for https://github.com/ipfs/go-ipfs/pull/8068, so we may want to park this PR until 0.9 ships and aim at 0.10+,  but we need to do this at some point: otherwise every IPNS website will be a bit slower due to unnecessary requests than it should.

BigLep:
This is blocked until TTLs are respected in DNSLink and IPNS.

mycorrhizas:
> This is blocked until TTLs are respected in DNSLink and IPNS.

👍

PR #8076, opened 2021-04-15
Move dag import and export into the local node CoreAPI implementation

aschmahmann:
This is an attempt to make importing/exporting CAR files into go-ipfs generally more usable before we extend the CoreAPI to include it.

The immediate use is to just unlock making it possible to do CAR import/export without going through the commands lib and either the CLI or HTTP API.

### DIFF
--- a/core/commands/dag/dag.go
+++ b/core/commands/dag/dag.go
@@ -2,6 +2,7 @@ package dagcmd
 
 import (
 	"fmt"
+	"github.com/ipfs/go-ipfs/core/coreapi"
 	"io"
 
 	"github.com/ipfs/go-ipfs/core/commands/cmdenv"
@@ -55,13 +56,7 @@ type ResolveOutput struct {
 
 // CarImportOutput is the output type of the 'dag import' commands
 type CarImportOutput struct {
-	Root RootMeta
-}
-
-// RootMeta is the metadata for a root pinning response
-type RootMeta struct {
-	Cid         cid.Cid
-	PinErrorMsg string
+	Root coreapi.RootMeta
 }
 
 // DagPutCmd is a command for adding a dag node
@@ -154,11 +149,6 @@ var DagResolveCmd = &cmds.Command{
 		}),
 	},
 	Type: ResolveOutput{},
-}
-
-type importResult struct {
-	roots map[cid.Cid]struct{}
-	err   error
 }
 
 // DagImportCmd is a command for importing a car to ipfs
--- a/core/coreapi/dag.go
+++ b/core/coreapi/dag.go
@@ -3,6 +3,7 @@ package coreapi
 import (
 	"context"
 	"errors"
+	"fmt"
 	"io"
 
 	cid "github.com/ipfs/go-cid"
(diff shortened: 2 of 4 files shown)


core/commands/dag/dag.go
```diff
@@ -2,6 +2,7 @@ package dagcmd
 
 import (
 	"fmt"
+	"github.com/ipfs/go-ipfs/core/coreapi"
```
aschmahmann, line 5: TODO: group with other go-ipfs imports

core/coreapi/dag.go
```diff
@@ -56,6 +61,244 @@ func (api *dagAPI) Session(ctx context.Context) ipld.NodeGetter {
...
+type RootMeta struct {
+	Cid         cid.Cid
+	PinErrorMsg string
+}
+
+// ImportMany imports CAR files into the blockstore and returns the root CIDs. If pinning was requested errors that
+// occurred when importing each root will be returned as well.
+//
+// This function's API is not yet stable
+func (api *dagAPI) ImportMany(ctx context.Context, directory files.Directory, doPinRoots bool) (<-chan RootMeta, <-chan error) {
```
aschmahmann, line 74: Really not a fan of this function signature, but it does emulate what we had before. Up for suggestions, or maybe we just merge this PR since it's better than what we've got and work on improving this later. Marked this as unstable to mark my dissatisfaction.

The use of files.Directory seems pretty suspect here too, should it be some typed iterator/slice/channel of ReaderClosers?

aschmahmann: A simpler `Import` function could also be added here separate from what's used by the commands lib.

core/coreapi/dag.go
```diff
@@ -56,6 +61,244 @@ func (api *dagAPI) Session(ctx context.Context) ipld.NodeGetter {
...
+		ret <- importResult{err: err}
+		return
+	}
+
+	ret <- importResult{roots: roots}
+}
+
+// Export exports the DAG rooted at a given CID into a CAR file. The error channel will return at most one error and
+// the reader is usable before the error channel has been closed.
+func (api *dagAPI) Export(ctx context.Context, c cid.Cid) (io.ReadCloser, <-chan error) {
```
aschmahmann, line 260: I'm much happier with the `Export` API, although returning error channels still doesn't feel great here. Up for suggestions.

core/coreapi/dag.go
```diff
@@ -56,6 +61,244 @@ func (api *dagAPI) Session(ctx context.Context) ipld.NodeGetter {
...
+			api.Session(ctx),
+			[]cid.Cid{c},
+			pipeW,
+		); err != nil {
+			errCh <- err
+		}
+
+		if closeErr := pipeW.Close(); closeErr != nil && err == nil {
+			errCh <- fmt.Errorf("stream flush failed: %s", closeErr)
+		}
```
aschmahmann, line 296: I kept the code mostly the same, except that we now only return a single error instead of returning two errors assuming that the caller will only process one of them which is what we did before. This required moving this pipe closure + error checking out of the defer statement.

We could also just wait on both functions to complete and return a multierror.

core/coreapi/dag.go
gammazero: Suggestion: Instead of returning an error channel, what about calling [`CloseWithError`](https://golang.org/pkg/io/#PipeWriter.CloseWithError) on pipeW.  That way the caller will get the error when trying to read from the `io.ReadCloser` returned here, if an error happens before reading is finished.
Example: https://play.golang.org/p/5caJVZb869-

core/coreapi/dag.go
```diff
@@ -56,6 +61,244 @@ func (api *dagAPI) Session(ctx context.Context) ipld.NodeGetter {
...
+func (api *dagAPI) importWorker(ctx context.Context, dir files.Directory, ret chan importResult) {
+	// this is *not* a transaction
+	// it is simply a way to relieve pressure on the blockstore
+	// similar to pinner.Pin/pinner.Flush
+	batch := ipld.NewBatch(ctx, api)
+
+	roots := make(map[cid.Cid]struct{})
+
+	it := dir.Entries()
+	for it.Next() {
```
gammazero, line 186: Would this benefit from starting up a number of goroutines (one per entry) and doing the imports concurrently?

aschmahmann: Almost certainly (up until some limit of worker nodes).

Also, this code does not properly respect contexts. If for some reason the batching datastore underneath does not return an error on `batch.Add(ctx, node)` we won't abort during this process.

core/coreapi/dag.go
aschmahmann: Thanks, that looks useful! @ribasushi do you recall any reason we didn't do this before (e.g. it didn't play well with the commands lib emitter)?

ribasushi: I simply didn't know about `CloseWithError` at the time: 👍

core/coreapi/dag.go
ribasushi: One could certainly do things here in parallel, but be careful not switching to parallel full-on imports, as this will break a use case hat goes something like:
`ipfs dag import {{named-fifo-pipe-containing-blocks-and-NO-ROOTS-to-pin}} {{named-fifo-pipe-containing-NO-BLOCKS-and-a-list-of-roots-to-pin-based-on-previous-fifo}}`

ribasushi (approved):
The actual code-migration LGTM, perhaps further improvements ( paralellisation ) could land in subsequent PRs

👍 either way

aschmahmann:
I think I introduced a bug somewhere here since there's a test that's hanging. Need to poke around to find out what I messed up, before merging 😄

PR #8081, opened 2021-04-20
Add fetcher config top level dependency

hannahhoward:
# Goals

Move dagpb dependency up to go-ipfs and out of go-fetcher, remove custom construction of FetcherConfig everywhere

# Implementation

- Construct a top level fetcher factory instance.
- Feed it to the resolver
- Remove custom config when using the resolver

# For Discussion

- will submit node api pr once this is approved


willscott (approved):

core/node/core.go
```diff
@@ -84,11 +90,22 @@ func (s *syncDagService) Session(ctx context.Context) format.NodeGetter {
 	return merkledag.NewSession(ctx, s.DAGService)
 }
 
+// FetcherConfig returns a fetcher config that can build new fetcher instances
+func FetcherConfig(bs blockservice.BlockService) fetcher.Factory {
```
aschmahmann, line 94: @hannahhoward @Stebalien do you think these lines of code for adding in UnixFS support are something that people are going to be missing?

Embedding go-ipfs as a library is not the friendliest at the moment so I'm wondering if people are expecting UnixFS support to live elsewhere and if this should live in a helper package somewhere.

Not a huge deal since we can wait until the requests come in to extract this code out, but if there are users who would be confused (or if we're going to feel the need to put this code in the release notes for go-path or go-merkledag) then we could put it somewhere else (e.g. a helper package in unixfsnode or go-merkledag)

hannahhoward: it was in `go-unixfsnode`, people objected...

I would rather it live here.

Issue #7972, opened 2021-03-11
Data Transfer Stall


Stebalien:
**Update** (May 28th): We have deployed https://github.com/ipfs/go-ipfs/pull/8082 to monitor BS with the latest fix https://github.com/ipfs/go-bitswap/pull/477. We are still running into the 'this request was not processed in time' error but this is likely a false positive due to an incorrect timeout setting in the monitoring tool that we should calibrate for a GW node in production. We need to make this setting a normal IPFS config that can be modified through the standard CLI to avoid restarting the node and better calibrate the timeout. (There's a tribute issue about it that _might_ be picked up next week: https://github.com/ipfs/go-ipfs/issues/8157.)

---------------------------------

#### Version information:

go-ipfs v0.8.0

#### Description:

We saw a bitswap stall on some of the preload nodes where they were unable to download any content, even from connected nodes (https://github.com/ipfs/js-ipfs/issues/3577).

This looks related to https://github.com/ipfs/go-ipfs/issues/5183 but _that_ issue was mostly about _specific_ content. That is, a specific request would fail, and we'd get into a stuck state where we couldn't download that specific content.

The preload nodes did not appear to be spinning, stuck, etc. but restarting them fixed the issue.

Possible causes:

* Deadlock somewhere.
* Full wantlist? Maybe we have some cases where wants are getting stuck in the wantlist and never clearing, building up over time?

BigLep:
@aschmahmann is going to provide more context/info.

aschmahmann:
There was a report that a node was not sending data over bitswap despite having the data locally. The user was able to run `ipfs block stat --offline <cid>` and get a result, and yet the data seemed not to transfer over Bitswap.

I investigated further using `vole`  https://github.com/aschmahmann/vole to directly send the node a Bitswap request such as:

`vole.exe bitswap check <cid> <peer-multiaddr>` and got the response `{"Found":false,"Responded":false,"Error":null}`, which means we did not get a response or an error on whether we received the other peer has the data or not. IIRC the peer was still able to respond to requests from other peers about other CIDs, but I don't recall at the moment what other types of probing I tried from my machine 😟.

Stebalien:
Matt noted that disabling QUIC seemed to help. This issue _could_ be related to https://github.com/libp2p/go-libp2p-quic-transport/issues/57 (where we could be failing to timeout somewhere in bitswap, backing everything up).

schomatis:
[...]

PR #8082, opened 2021-04-20
Monitor tool: query CIDs in BitSwap localhost

schomatis:
# Meta

Patch to track in the BitSwap engine a failure to respond to a requested CID, see https://github.com/ipfs/go-ipfs/issues/7972#issuecomment-819780004.

*This is a draft PR without actual intention of merging, ever.* It is only submitted in this format to streamline the review process. As such attention should not be on proficient Go code or alignment with `go-ipfs` intrinsics but only on what can be added to a production GW node to monitor and log activity related to this issue without disrupting normal operation.

Most of the actual logic is in https://github.com/ipfs/go-bitswap/pull/475.

We hook a task on the node constructor to make a request every X seconds of a list of CIDs to our own BitSwap engine and see if they are all accounted for within those same X seconds.

TODO:
- [x] Fix the interval X.
- [x] Fix the number of CIDs we request.
- [ ] Confirm the level of logs we can have without causing any disruption. Do we rotate them?
- [x] Make all these configuration options we can adjust on the fly without restarting the GW node.
- [ ] For a second iteration: see if we want to replicate this test from an external process outside localhost (to better test the transport layer such as QUIC) and refactor accordingly.
- [x] Fix `go-bitswap` dependency replace to a commit (commenting out the branch replace).

-------------------------------------------

# PR description

This PR tests the BitSwap (BS) response mechanism with the objective of checking for inconsistencies such as https://github.com/ipfs/go-ipfs/issues/7972#issuecomment-819780004. It modifies a [`v0.8.0`](https://github.com/ipfs/go-ipfs/releases/tag/v0.8.0) node to start a goroutine along the node startup process that will constantly request a list of CIDs to its own BS component. (Analogous to a localhost connection where we bypass part of the network stack, we connect here directly to the node's own `Host` peer.)

This injected goroutine runs a BS test utility derived from the [`vole`](https://github.com/aschmahmann/vole) tool to request CIDs and monitor that the BS response accounts for all of them (we don't focus at the moment on whether we have the CIDs or not but only on BS responding to *all* requests, see cited issue for more background). Most of the BS logic, including the `vole` request CID tool, lives in a separate PR in the BS repo: https://github.com/ipfs/go-bitswap/pull/475.

The previous architecture translates into the fact that we no longer have to run two separate processes to test BS (as in the original `vole`), one running a node and the other sending requests to it, but only one: the node itself functioning as usual and *also* running this constant check in the background requesting *itself* for a list of artificial CIDs (again, we don't focus on actual meaningful content right now). The rationale for this design is to be able to deploy this modified version *in place* of a production GW node (where the original issue was first spotted) that handles a meaningful load (making this test run in a more realistic scenario and with more chances of finding anything useful).

While developing this the original issue was *likely* fixed by https://github.com/ipfs/go-bitswap/pull/477 (cherry picked here), but we still finish and deploy it in case we can find any other related problems.

Along with the requests in this patch we include some extra logging (`LogTargetPeer`) in the BS code base targeting specifically the "fake" peer (in the goroutine inside the node itself) from where the requests originate to spot if at any point the pipeline of task creation and dispatching ("pushing" and "popping" in BS terminology) that leads to reporting back a requested CID is broken. We can very easily modify this log pattern to go deeper in the code flow of interest if needed. We err for now on the side of less verbosity to not flood the log in this first test. The modified node should be monitored to make sure this is not the case and that the logs are rotated properly.

## Configuration and use

There are 3 configuration variable that are added to this patch (in the form of environment variables for simplicity) that need to be set *before* running it and can also be modified on the fly without restarting the node (to minimize operation disruption in production).

* `BS_CFG_TIMEOUT`: the time we wait for the response of the requested CIDs. **This is probably the most sensitive configuration of all** that can make the test meaningful or meaningless, and will probably need to be experimented with before arriving at an acceptable value. It is probably important to take into account that the connection is in "localhost" mode with virtually zero latency so the timeout will mainly be derived from the node average load. (After the timeout elapses, either with all responses or not, we rerun the test in an infinite loop in the goroutine until the node stops.)

* `BS_CFG_CID_REQ_NUM`: Number of CIDs requested on each test. Not sure how critical this configuration is. Ideally we would want to mimic a typical client request.

* `BS_CFG_LOG`: Log level for the BS and related components. The rest is set to `WARN`. During tests it is recommended to set this to `INFO` (the level at which we added the extra logging). In any case the log levels can be adjusted on the fly through the standard `ipfs log` CLI.

(See more details in the [documentation](https://github.com/ipfs/go-ipfs/pull/8082/files#diff-4140a14426af9479ef5f4cbe610e3251f8a2425b1503811955c7ea81adc049cbR109-R125) of the code itself.)

A typical configuration run during local (non-production) testing of this patch can be seen in [`run-bs-monitor.bash`](https://github.com/ipfs/go-ipfs/pull/8082/files#diff-57368464713abdbde4a15be0881be9ea08b03628f018f64787aa6c0f7a8555d8).

Apart from these extra configurations the node should be run as it is normally configured in production. The added configurations shouldn't interfere with the rest of the node's operations.

## Log output

A typical log output of a test would be:

```
INFO	bitswap	connecting through TCP
INFO	engine	[TRACKING-BS-PEER] pushing task for CID [bafkqaaju]
INFO	engine	[TRACKING-BS-PEER] pushing task for CID [bafkrefclej3xpvg5d7dby34ij5egihicwtisduy]
INFO	engine	[TRACKING-BS-PEER] pushing task for CID [bafkqaajq]
[...]
INFO	engine	[TRACKING-BS-PEER] popped [10] tasks from queue
INFO	engine	[TRACKING-BS-PEER] popped task CID: [bafkrefdlq2zhh7zu7tqz224aj37vup2xi6w2j2q]
INFO	engine	[TRACKING-BS-PEER] popped task CID: [bafkrefclej3xpvg5d7dby34ij5egihicwtisduy]
[...]
INFO	engine	[TRACKING-BS-PEER] popped task CID: [bafkrefcoa5aikyv63ofwbtqfyhpm7y5nc23sema]
INFO	engine	[TRACKING-BS-PEER] popped task CID: [bafkqaajq]
INFO	engine	[TRACKING-BS-PEER] popped task CID: [bafkqaajt]
INFO	engine	[TRACKING-BS-PEER] message sent [map[blocks:[bafkqaajr bafkqaajq bafkqaajt bafkqaaju bafkqaajs] presences:[{bafkrefguonpdujs6c3xoap2zogfzwxidagoapwa DontHave} {bafkrefcoa5aikyv63ofwbtqfyhpm7y5nc23sema DontHave} {bafkrefdlq2zhh7zu7tqz224aj37vup2xi6w2j2q DontHave} {bafkrefclej3xpvg5d7dby34ij5egihicwtisduy DontHave} {bafkrefc75tvwn76in44nsutynrwws3dzyln4eoi DontHave}] wants:[]]]
```

What we need to look out for is an `ERROR` in the BS log like:

```
ERROR	bitswap	CheckBitswapCID: did not get HAVE/DONT-HAVE response on CIDs:
```

### DIFF
--- a/cmd/ipfs/daemon.go
+++ b/cmd/ipfs/daemon.go
@@ -29,6 +29,7 @@ import (
 	migrate "github.com/ipfs/go-ipfs/repo/fsrepo/migrations"
 	sockets "github.com/libp2p/go-socket-activation"
 
+	logging "github.com/ipfs/go-log"
 	cmds "github.com/ipfs/go-ipfs-cmds"
 	mprome "github.com/ipfs/go-metrics-prometheus"
 	options "github.com/ipfs/interface-go-ipfs-core/options"
@@ -207,6 +208,13 @@ func daemonFunc(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment
 
 	// let the user know we're going.
 	fmt.Printf("Initializing daemon...\n")
+
+	if level := os.Getenv("BS_LOG"); level != "" {
+		logging.SetAllLoggers(logging.LevelWarn)
+		logging.SetLogLevel("bitswap", level)
+		logging.SetLogLevel("engine", level)
+		logging.SetLogLevel("bs:peermgr", level)
+	}
 
 	defer func() {
 		if _err != nil {
--- a/cmd/ipfs/daemon.go
+++ b/cmd/ipfs/daemon.go
@@ -209,11 +209,14 @@ func daemonFunc(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment
 	// let the user know we're going.
 	fmt.Printf("Initializing daemon...\n")
 
-	if level := os.Getenv("BS_CFG_LOG"); level != "" {
+	bsLogLevel := os.Getenv("BS_CFG_LOG");
+	if  bsLogLevel == "" {
+		panic("BS_CFG_LOG not set, recommended to set to INFO")
+	} else {
 		logging.SetAllLoggers(logging.LevelWarn)
-		logging.SetLogLevel("bitswap", level)
-		logging.SetLogLevel("engine", level)
-		logging.SetLogLevel("bs:peermgr", level)
+		logging.SetLogLevel("bitswap", bsLogLevel)
+		logging.SetLogLevel("engine", bsLogLevel)
+		logging.SetLogLevel("bs:peermgr", bsLogLevel)
 	}
 
 	defer func() {
--- a/cmd/ipfs/daemon.go
+++ b/cmd/ipfs/daemon.go
@@ -211,13 +211,11 @@ func daemonFunc(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment
 
 	bsLogLevel := os.Getenv("BS_CFG_LOG");
 	if  bsLogLevel == "" {
-		panic("BS_CFG_LOG not set, recommended to set to INFO")
-	} else {
-		logging.SetAllLoggers(logging.LevelWarn)
-		logging.SetLogLevel("bitswap", bsLogLevel)
-		logging.SetLogLevel("engine", bsLogLevel)
-		logging.SetLogLevel("bs:peermgr", bsLogLevel)
+		bsLogLevel = "info"
 	}
+	logging.SetLogLevel("bitswap", bsLogLevel)
+	logging.SetLogLevel("engine", bsLogLevel)
+	logging.SetLogLevel("bs:peermgr", bsLogLevel)
 
 	defer func() {
 		if _err != nil {
--- a/cmd/ipfs/daemon.go
+++ b/cmd/ipfs/daemon.go
@@ -209,7 +209,7 @@ func daemonFunc(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment
 	// let the user know we're going.
 	fmt.Printf("Initializing daemon...\n")
 
-	if level := os.Getenv("BS_LOG"); level != "" {
+	if level := os.Getenv("BS_CFG_LOG"); level != "" {
 		logging.SetAllLoggers(logging.LevelWarn)
 		logging.SetLogLevel("bitswap", level)
 		logging.SetLogLevel("engine", level)
--- a/core/node/core.go
+++ b/core/node/core.go
@@ -3,6 +3,8 @@ package node
 import (
 	"context"
 	"fmt"
+	"os"
+	"strconv"
 	"time"
 
 	"github.com/ipfs/go-bitswap"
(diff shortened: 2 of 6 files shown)


cmd/ipfs/daemon.go
```diff
@@ -208,6 +209,13 @@ func daemonFunc(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment
 	// let the user know we're going.
 	fmt.Printf("Initializing daemon...\n")
 
+	if level := os.Getenv("BS_LOG"); level != "" {
+		logging.SetAllLoggers(logging.LevelWarn)
+		logging.SetLogLevel("bitswap", level)
+		logging.SetLogLevel("engine", level)
```
schomatis, line 215: The engine is too verbose so we might want to skip this and only track the other two.

core/node/core.go
```diff
@@ -97,11 +102,49 @@ func OnlineExchange(provide bool) interface{} {
...
+						// We do the check every X seconds. This time is also
+						// as much as we are willing to wait for the response
+						// on all CIDs requested. This means we only do *one*
+						// test at a time.
+						testContext, _ := context.WithTimeout(ctx, time.Second*5) // fixme: parameter, make this a config so we can adjust on the fly
+
+						checkBitswapResponse(testContext, host.ID())
+
+						select {
+						case <-ctx.Done():
```
Stebalien, line 118: So... this is one of the quirks of fx. This context will (well, may...) get canceled once we're finished starting.

Either call `helpers.LifecycleCtx(lc)`, or, e.g., close a channel from an `OnStop` hook.

go.mod
```diff
@@ -112,3 +112,7 @@ require (
 )
 
 go 1.14
+
+// Only works with: GOPROXY=direct (https://github.com/golang/go/issues/32955#issuecomment-509318558)
+// Still, the branch name will get rewritten with the commit..
+replace github.com/ipfs/go-bitswap => github.com/ipfs/go-bitswap monitor/data-stall
```
Stebalien, line 118: It's usually simpler (especially when giving code to others) to just specify the commit directly without a replace directive.

1. Specify `github.com/ipfs/go-bitswap@monitor/data-stall` in the dependencies.
2. Run `GOPROXY=direct go mod tidy` (will resolve `monitor/data-stall` to a commit.

After that, everything will work even with the default GOPROXY.

Stebalien: (not important, just FYI)

core/node/core.go
```diff
@@ -97,11 +102,49 @@ func OnlineExchange(provide bool) interface{} {
 				return exch.Close()
 			},
 		})
+		lc.Append(fx.Hook{
+			OnStart: func(ctx context.Context) error {
+				go func() {
+					for {
```
Stebalien, line 108: FYI, https://golang.org/pkg/time/#Tick

Stebalien: Ah, nvm. I see why you did it this way.

core/node/core.go
schomatis: Ouch! Thanks for flagging this, I would have never expected this behavior:

https://github.com/uber-go/fx/blob/4adc2bf73f2cdcd423bf22725985177670a91d4d/app.go#L738-L742

go.mod
schomatis: Yes, this is not meant for actual production, I just wanted a way to link both development branches without having to manually update on each change. This was the only way I found to link to a moving target.

core/node/core.go
Stebalien: Yeah, I'm pretty sure we have a few bugs due to this in lotus.

cmd/ipfs/daemon.go
schomatis: Most of the added logs though are in that component for the `INFO` level so we might want to move them to the more general BS or create a new log subsystem.

cmd/ipfs/daemon.go
```diff
@@ -208,6 +209,16 @@ func daemonFunc(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment
 	// let the user know we're going.
 	fmt.Printf("Initializing daemon...\n")
 
+	bsLogLevel := os.Getenv("BS_CFG_LOG");
+	if  bsLogLevel == "" {
+		panic("BS_CFG_LOG not set, recommended to set to INFO")
+	} else {
```
aschmahmann, line 215: Can you set this to some default that works for you? I'd like to make deploying this on a gateway as simple as possible.

Ideally this includes just asking infra to checkout the branch and build it before running the rest of the normal code (e.g. and not having to run `run-bs-monitor.bash`). I'm sure we can do it, but if we could just build the branch that'd likely be much easier.

schomatis: Updated to 'info' as default. Also removed the 'warn' log level for the rest of the subsystems to be as least disruptive as possible of standing GW configurations.

schomatis:
We need to make these two configuration values part of the `ipfs config` CLI interface so we can change them on the fly without restarting the node. Currently they are set as env.vars and once we start the node there is no easy way to modify them (so even if the routine here constantly checks for updates to these they are always constant).

https://github.com/ipfs/go-ipfs/pull/8082/files#diff-4140a14426af9479ef5f4cbe610e3251f8a2425b1503811955c7ea81adc049cbR128-R129

go.mod
```diff
@@ -32,7 +32,7 @@ require (
 	github.com/ipfs/go-ipfs-blockstore v0.1.4
 	github.com/ipfs/go-ipfs-chunker v0.0.5
 	github.com/ipfs/go-ipfs-cmds v0.6.0
-	github.com/ipfs/go-ipfs-config v0.12.0
+	github.com/ipfs/go-ipfs-config v0.16.1-0.20210830194616-e846165ea1ca
```
coryschwartz, line 35: This will need to be undone if https://github.com/ipfs/go-ipfs-config/pull/144 is merged.

core/node/core.go
```diff
@@ -97,11 +103,61 @@ func OnlineExchange(provide bool) interface{} {
 				return exch.Close()
 			},
 		})
+		lc.Append(fx.Hook{
+			OnStart: func(ctx context.Context) error {
+
+				go func() {
+					for cbs != nil {
```
coryschwartz, line 110: @schomatis I changed this to use the config, this works so long as the tester has Internal.Bitswap setup in the config, and does nothing otherwise.

go.mod
schomatis: Will likely never merge this (at least not as an embedded process) so neither that PR.

core/node/core.go
schomatis: @coryschwartz Great, thanks! Just to confirm if I understand this correctly, a change through the `ipfs config` CLI to these configurations will be reflected on the fly in the `*config.InternalBitswap` pointer here without needing to restart the daemon, right?

schomatis:
We haven't used this tool in a while (nor needed it) and since it wasn't meant to be merged I'll be closing it now.

PR #8084, opened 2021-04-21
Add unixfs pathing to graphsync

hannahhoward:
# Goals

Enable graphsync in go-ipfs to perform selections based on unixfs directory paths. 

# Implementation

- update go-graphsync, go-multihash, go-ipld-prime, go-car, and add go-unixfsnode
- add the node reifier function to graphsync construction so that graphsync queries can traverse unixfs nodes with named pathing
- this does not require the rest of the IPLD-In-IPFS work

License: MIT


ribasushi:
This plan looks *really* concerning to me.

If I understand things correctly we are about to land this heuristic https://github.com/ipfs/go-unixfsnode/blob/v1.1.1/reification.go#L25-L29 into all spots below:

> update go-graphsync, go-multihash, go-ipld-prime, go-car, and add go-unixfsnode

We seem to be doing this **without the possibility of turning this heuristic off**. This means that we are further dividing the  space into "blessed" and "second-class citizen" types of DAGs, this time in the selector space. This already happened in graphsync, where [the only hope one has to move anything is if they use DagPB or DagCBOR](https://github.com/filecoin-project/go-fil-markets/issues/365#issuecomment-673013618), because the alternative is having the entire "supply chain" being updated to a new codec you want to use (e.g. BTC or ETH ).

The effect of this in medium to long term would be to severely deprioritize or outright eliminate the notion of context-signalling from protocol evolution. Given how much effort we spend on type-annotation everywhere else in our stack ( `multi*` ) it seems really out of character to push for signalling elimination in this layer. In practice that would result in most systems gravitating towards selectors transferred across nodes would be the "simplest thing possible" almost "ipfs pathing-style" simple, and then various consumers would all be expected to have a (growing) list of selection contexts they would need to apply the selector to and "see what happens".

More specifically to the case at hand - signalling is not really necessary here in the first place: as a producer of the selector I already know it will be applied to UnixFS - I have this external context. What ought to happen is the selector in question should be "compiled down" to an actual DM-level selector and *that* is what would be sent to the provider who would only need to know how to speak DM and nothing more.

aschmahmann:
> What ought to happen is the selector in question should be "compiled down" to an actual DM-level selector and that is what would be sent to the provider who would only need to know how to speak DM and nothing more.

💯

It's annoying enough that both parties need to support the same codecs in order to transfer the DAG, however it's a pretty unavoidable part of IPLD at the moment (short of WASM magic + CIDs as a codec type). Adding application specific types with implicit signaling (which is effectively what ADLs are) to the data transfer layer seems like a big problem.

Even if we wanted to do this ADLs would need to be negotiated ahead of time in the GraphSync protocol, or users would need to fork the protocol so instead of `/graphsync/1.0.0` it would be `/graphsync-with-unixfs-support/1.0.0`. Since it'd be nice to just keep the protocol simpler we should make sure the concept of an ADL (and the associated special selector) doesn't leave the application node.

Additionally, if there are network-wide parameters that are not negotiated inside of the protocol and that the user can configure then the user should be forced to choose a different protocol ID. As an example, in the DHT code we added some [validation](https://github.com/libp2p/go-libp2p-kad-dht/blob/71598922a0596aee42afd06b048450429a14d3fa/dht_options.go#L135) to try and prevent people from instantiating non compliant `/ipfs/kad/1.0.0` protocol implementations by accident.

hannahhoward:
@ribasushi @aschmahmann 

Part of the problem with the selector being "compiled down"  is there is no equivalent "DM level selector" in the current Selector spec. The only way to find these paths is to iterate the Links array of the PB Node and look for the one with the matching name. There is no current selector that expresses doing that at the PB level.

If we want to add such conditional selection, we could augment the Selector spec (and the implementation in go-ipld-prime -- there is an existing unimplemented rough spec for conditional selectors though I'm not sure if it even covers this use case)

This is a strawman PR-- there's no need to merge. The request here came directly from Textile, and I provided a way to do it. Though if it's not merging we should work with @jsign to make sure they can still do what they need to do. It doesn't have to be the final way to implement it. However I think we should seriously consider how we're going to support this -- cause I suspect everyone is going to want to use this kind of UnixFS v1 selector very quickly.

I would push back against the idea that we can just keep schemas and ADLS inside the application boundary. It significantly limits the utility of the protocol. There are many selector queries beyond this that are either impossible to express at the direct node level or it's not at all ergonomic to do so. For example, we use tuple structs all over Filecoin and constructing selectors for tuple structs without schemas is no fun. (though this is a case where at least compiling a selector down is potentially possible if you have the schema). Schemas for specialized requests are also a potential big speed up in serving requests. Application developers are likely to want to build all kinds of schemas and ADLs for themselves and want a way to transfer them with Graphsync. I expect also that over time people will want to implement custom selectors -- extensions to the selector spec itself.

We don't need to rename the protocol though to support ADL negotiation. Graphsync already supports arbitrary extensions. I think we may need to eventually build a capabilities query around extensions, and maybe some kind of global extensions registry, especially as more things come into the wild. But for the time being, if you want to extend the protocol, just add an extension for an ADL. Make sure part of the extension includes sending confirmation in the response that the extension is supported. Either both sides will support it, or they won't, and the requestor can cancel the query if it's not supported.

Anyway, since the request to do this came outside of my regular work stream I am going to assume my work here is done and @ribasushi you and @jsign will work out a way to implement whatever is needed.

warpfork:
> What ought to happen is the selector in question should be "compiled down" to an actual DM-level selector and that is what would be sent to the provider who would only need to know how to speak DM and nothing more.

Okay, goalpost check -- on where the goalposts *can be* -- this is a nice dream.  **It's also impossible.**

... Unless the ADL logic is zero percent data-dependent.  Which happens approximately never.  Such an ADL would generally be pretty bad at its job.  (Imagine a sharding mechanism where the answer is "always fully shard, all the way, regardless of how much data you have".  It's so non-optimal that we can say approximately no one ever does this.)

So: It is not generally possible to "precompile" a selector plus knowledge of an ADL into a steamrolled selector which can work on the data without knowledge of the ADL.

(You *can* "precompile" a selector plus knowledge of an ADL *plus knowledge of the data* into a selector which will work on the same data without knowledge of an ADL.  It's somewhat rare for this to be useful, though.)

---

The other comments about explicitness: generally, yes.

The more we push the enabling of any ADL detection logic towards the main method of a program, the better off we'll be.

warpfork:
There's some talk about this in the IPLD meeting last night, and we might want to mine the recording for notes.

- It is true we don't have any formal or unified strategy for identifiers for ADLs yet.
    - There are certainly people interested in this.
    - It's frequently discussed that it would be nice if there is a canonicalization system for this -- but it's also something that needs a serious goalpost check, and is probably functionally impossible.
          - We don't have a "canonical" bytecode to hash
          - Even if we did, there'd be a versioning problem for that bytecode's interpreter, which sort of makes for a recursive problem
          - Even if we did, we'd probably want "rockets" -- implementations in some host native language, whatever that may be, which are functionally equivalent to any portable interpretable form that we might also pursue...
          - ... and then we'd have the equivalence-proving problem, which is of course *cough* nontrivial
          - And of course you'll notice that in practice we're building the single-language "rockets" first and merely wishing we had some portable interpreted bytecode for this stuff.
    - Suffice to say: if someone wanted to push on this, I'd highly recommend extreme rabbit-hole avoidance and preference for pragmatism, because the theoretically-pure approaches are all deep, deep rabbit-holes indeed.

- We have some prior docs which poked at this:
    - Anciently: https://github.com/ipld/specs/blob/7f5db3b1e5f3dd10946369ce1a1400bbc7ad64be/design/history/exploration-reports/2019.06-advanced-layouts-critical-design-decisions.md
    - Slightly less anciently: https://github.com/ipld/specs/blob/7f5db3b1e5f3dd10946369ce1a1400bbc7ad64be/design/history/exploration-reports/2019.12-manifesting-adls.md
    
- Re: fallback to an iterative and high-RTT-count process when a remote agent doesn't like your ADL request:
    - Yeah, I guess that's something that probably ought to exist, as it is ultimately the only fallback that we really know will always be viable.
    - I _guess_ that should be part of the data transfer protocol...?  The data transfer protocol will still probably want to do its best to minimize RTTs and maximize parallel pipelining when being stuck with this fallback, I imagine.

aschmahmann:
As mentioned above this change is protocol breaking and even if there was a desire to version the protocol to add support for the proposed functionality there are a number of ways we could go about that that likely merit discussion.

As a result I've marked this issue as blocked and being backwards incompatible.

---

> Re: fallback to an iterative and high-RTT-count process when a remote agent doesn't like your ADL request:

Side note: 👍 to the statements there.

PR #8096, opened 2021-04-28
fix(config): support dynamic map keys with dots

lidel:
This PR adds support for alternative map notation in `ipfs config` command,  
which enables access to values behind a map key that includes a dot, for example:

- `ipfs config 'Gateway.PublicGateways["dweb.link"].UseSubdomains'`
- `ipfs config 'Pinning.RemoteServices["nft.storage"].Policies'`

Context: https://github.com/ipfs/ipfs-webui/issues/1770 (cc @alanshaw @aschmahmann)

Included sharness test + confirmed accessing fields like this works over HTTP API with regular curl as well:

```console
$ ipfs pin remote service add nft.storage https://nft.storage/api [secret]
...
$ curl -sX POST 'http://127.0.0.1:5001/api/v0/config?arg=Pinning.RemoteServices%5B%22nft.storage%22%5D.Policies.MFS.Enable' | jq
{
  "Key": "Pinning.RemoteServices[\"nft.storage\"].Policies.MFS.Enable",
  "Value": false
}
```

### DIFF
--- a/core/commands/config.go
+++ b/core/commands/config.go
@@ -56,6 +56,11 @@ Get the value of the 'Datastore.Path' key:
 Set the value of the 'Datastore.Path' key:
 
   $ ipfs config Datastore.Path ~/.ipfs/datastore
+
+Values behind map key names that include dots can be accessed like this:
+
+  $ ipfs config Pinning.RemoteServices["pins.example.org"].Policies
+
 `,
 	},
 	Subcommands: map[string]*cmds.Command{
@@ -176,7 +181,13 @@ var configShowCmd = &cmds.Command{
 	Helptext: cmds.HelpText{
 		Tagline: "Output config file contents.",
 		ShortDescription: `
-NOTE: For security reasons, this command will omit your private key and remote services. If you would like to make a full backup of your config (private key included), you must copy the config file from your repo.
+'ipfs config show' returns config contents without private keys and secrets.
+`,
+		LongDescription: `
+NOTE: For security reasons, this command will omit your private key and any
+access tokens for remote services. If you would like to make a full backup of
+your config (private key and secrets included), you must copy the config file
+from your IPFS repository (IPFS_PATH).
 `,
 	},
 	Type: make(map[string]interface{}),
--- a/core/commands/config.go
+++ b/core/commands/config.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/ipfs/go-ipfs/core/commands/cmdenv"
 	"github.com/ipfs/go-ipfs/repo"
+	. "github.com/ipfs/go-ipfs/repo/common"
 	"github.com/ipfs/go-ipfs/repo/fsrepo"
 
 	"github.com/elgris/jsondiff"
@@ -162,7 +163,8 @@ Values behind map key names that include dots can be accessed like this:
 //	matchesGlobPrefix("foo.bar.baz", []string{"*", "bar"}) returns true
 //	matchesGlobPrefix("foo.bar", []string{"baz", "*"}) returns false
 func matchesGlobPrefix(key string, glob []string) bool {
-	k := strings.Split(key, ".")
+	normalizedKey, _ := ConfigKeyToLookupData(key)
+	k := strings.Split(normalizedKey, ".")
 	for i, g := range glob {
 		if i >= len(k) {
 			break
--- a/repo/common/common.go
+++ b/repo/common/common.go
@@ -8,8 +8,8 @@ import (
 	"github.com/google/uuid"
 )
 
-// Find dynamic map key names passed  as Parent["foo"] notation
-var bracketsRe = regexp.MustCompile(`\[([^\[\]]*)\]`)
+// Find dynamic map key names passed with Parent["foo"] notation
+var bracketsRe = regexp.MustCompile(`\["([^\["\]]*)"\]`)
 
 // Normalization for supporting arbitrary dynamic keys with dots:
 // Gateway.PublicGateways["gw.example.com"].UseSubdomains
--- a/repo/common/common.go
+++ b/repo/common/common.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"regexp"
 	"strings"
+
+	"github.com/google/uuid"
 )
 
 // Find dynamic map key names passed  as Parent["foo"] notation
@@ -16,10 +18,10 @@ func keyToLookupData(key string) (normalizedKey string, dynamicKeys map[string]s
 	bracketedKeys := bracketsRe.FindAllString(key, -1)
 	dynamicKeys = make(map[string]string, len(bracketedKeys))
 	normalizedKey = key
-	for i, mapKeySegment := range bracketedKeys {
-		mapKey := strings.TrimLeft(mapKeySegment, "[\"")
-		mapKey = strings.TrimRight(mapKey, "\"]")
-		placeholder := fmt.Sprintf("mapKey%d", i)
+	for _, mapKeySegment := range bracketedKeys {
+		mapKey := strings.TrimPrefix(mapKeySegment, `["`)
+		mapKey = strings.TrimSuffix(mapKey, `"]`)
+		placeholder := uuid.New().String()
 		dynamicKeys[placeholder] = mapKey
 		normalizedKey = strings.Replace(normalizedKey, mapKeySegment, fmt.Sprintf(".%s", placeholder), 1)
 	}
--- a/repo/common/common.go
+++ b/repo/common/common.go
@@ -14,7 +14,7 @@ var bracketsRe = regexp.MustCompile(`\[([^\[\]]*)\]`)
 // Normalization for supporting arbitrary dynamic keys with dots:
 // Gateway.PublicGateways["gw.example.com"].UseSubdomains
 // Pinning.RemoteServices["pins.example.org"].Policies.MFS.Enable
-func keyToLookupData(key string) (normalizedKey string, dynamicKeys map[string]string) {
+func ConfigKeyToLookupData(key string) (normalizedKey string, dynamicKeys map[string]string) {
 	bracketedKeys := bracketsRe.FindAllString(key, -1)
 	dynamicKeys = make(map[string]string, len(bracketedKeys))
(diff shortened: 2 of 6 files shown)


repo/common/common.go
```diff
@@ -2,26 +2,50 @@ package common
...
 	"strings"
 )
 
+// Find dynamic map key names passed  as Parent["foo"] notation
+var bracketsRe = regexp.MustCompile(`\[([^\[\]]*)\]`)
+
+// Normalization for supporting arbitrary dynamic keys with dots:
+// Gateway.PublicGateways["gw.example.com"].UseSubdomains
+// Pinning.RemoteServices["pins.example.org"].Policies.MFS.Enable
+func keyToLookupData(key string) (normalizedKey string, dynamicKeys map[string]string) {
```
aschmahmann, line 15: Putting this here isn't sufficient since we need to deal with things like https://github.com/ipfs/go-ipfs/blob/eea198f0811f50ab14c4f713de07ce93dffec9af/core/commands/config.go#L90 that exist above this level.

We'll either need to export some of this behavior from here, or move the other behavior here and export it.

repo/common/common.go
```diff
@@ -2,26 +2,50 @@ package common
...
+	"regexp"
 	"strings"
 )
 
+// Find dynamic map key names passed  as Parent["foo"] notation
+var bracketsRe = regexp.MustCompile(`\[([^\[\]]*)\]`)
+
+// Normalization for supporting arbitrary dynamic keys with dots:
+// Gateway.PublicGateways["gw.example.com"].UseSubdomains
+// Pinning.RemoteServices["pins.example.org"].Policies.MFS.Enable
```
aschmahmann, line 14: Is this notation standard/used in other places or are we just making this up? Either way we'd need to document this in the `config` command.

repo/common/common.go
```diff
@@ -2,26 +2,50 @@ package common
...
+func keyToLookupData(key string) (normalizedKey string, dynamicKeys map[string]string) {
+	bracketedKeys := bracketsRe.FindAllString(key, -1)
+	dynamicKeys = make(map[string]string, len(bracketedKeys))
+	normalizedKey = key
+	for i, mapKeySegment := range bracketedKeys {
+		mapKey := strings.TrimLeft(mapKeySegment, "[\"")
+		mapKey = strings.TrimRight(mapKey, "\"]")
```
aschmahmann, line 21: These remove all instances. While I'd certainly hope that nobody does anything like `field["["foobar"]` I'd think that should cause some kind of error instead of getting silently fixed.

repo/common/common.go
```diff
@@ -2,26 +2,50 @@ package common
...
+		placeholder := fmt.Sprintf("mapKey%d", i)
```
aschmahmann, line 22: This could be problematic if someone uses `mapKey0` as a name of one of their fields. If we're going to use a placeholder approach (as opposed to unwinding the string segment by segment approach) then maybe there is a reserved/invalid JSON character we can use where we first validate that the string doesn't have any of those characters.

test/sharness/t0021-config.sh
```diff
@@ -126,6 +126,13 @@ test_config_cmd() {
     test_cmp replconfig.json newconfig.json
   '
 
+  # Dynamic keys with dot in their names
+  test_config_cmd_set "--json" "Gateway.PublicGateways[\"some.example.com\"].UseSubdomains" "true"
+  test_expect_success "'ipfs config' output for dynamic keys looks good" '
+    ipfs config Gateway.PublicGateways["some.example.com"].UseSubdomains > dynkey_out &&
+    grep true dynkey_out
+  '
+
```
aschmahmann, line 135: It'd be nice if we could validate that the result is valid JSON, this only proves that our code can read what it writes. Perhaps something like piping `ipfs config show` into `jq` to see that the Gateway JSON is what we expect it to be.

aschmahmann (changes requested):

aschmahmann:
Thanks for tackling this, a little weird that this hasn't come up previously. I guess it means people have mostly been manipulating the raw JSON regarding config sections like for Gateways.

repo/common/common.go
lidel: This is how keys in JSON objects are addressed on the web, and since we use JSON for config anyway, means we don't invent anything new, but follow existing convention. 

I've added note about this notation to `ipfs config --help` in 0a2336ea1

repo/common/common.go
lidel: Ah, nice catch. Changed to Prefix/Suffix in  732adebaf

repo/common/common.go
lidel: I switched placeholders to UUIDs to avoid conflicts with real keys while keeping magic to minimum  (732adebaf)
(uuid lib was already a transitional dependency, so should not impact binary size).

test/sharness/t0021-config.sh
lidel: Good idea, added in 732adebaf

repo/common/common.go
lidel: Refactored in 660cc66d4 and added tests in 6a409fba3 (had no docker set up locally, so waiting for CI to run them).

ps. I restored `%s` in error messages because entire error string gets sanitized already upstream, so we want to avoid double escaping.

lidel: Sharness tests fail on CI but pass locally, I'm suspecting different versions of `jq`, needs another look.

repo/common/common.go
```diff
@@ -2,26 +2,62 @@ package common
...
+// Normalization for supporting arbitrary dynamic keys with dots:
+// Gateway.PublicGateways["gw.example.com"].UseSubdomains
+// Pinning.RemoteServices["pins.example.org"].Policies.MFS.Enable
+func ConfigKeyToLookupData(key string) (normalizedKey string, dynamicKeys map[string]string) {
+	bracketedKeys := bracketsRe.FindAllString(key, -1)
+	dynamicKeys = make(map[string]string, len(bracketedKeys))
+	normalizedKey = key
+	for _, mapKeySegment := range bracketedKeys {
+		mapKey := strings.TrimPrefix(mapKeySegment, `["`)
+		mapKey = strings.TrimSuffix(mapKey, `"]`)
```
alanshaw, line 23: This regex doesn't mandate the double quotes, so would match "[nft.storage]" for example and then the square brackets would fail to be trimmed out here because it's expecting double quotes.

lidel: Hardened in: 36b62c1f8ff339ce489410c76da754d3cbc6c1f0

repo/common/common.go
lidel: I made our CI use the latest `jq` instead of old 1.5, and it is green now.

core/commands/config.go
```diff
@@ -157,7 +163,8 @@ Set the value of the 'Datastore.Path' key:
 //	matchesGlobPrefix("foo.bar.baz", []string{"*", "bar"}) returns true
 //	matchesGlobPrefix("foo.bar", []string{"baz", "*"}) returns false
 func matchesGlobPrefix(key string, glob []string) bool {
-	k := strings.Split(key, ".")
+	normalizedKey, _ := ConfigKeyToLookupData(key)
```
petar, line 166: strings.EqualFold (below) no longer works, since now it is applied to a rewritten key which has UUIDs.

repo/common/common.go
```diff
@@ -2,26 +2,62 @@ package common
...
 import (
 	"fmt"
+	"regexp"
 	"strings"
+
+	"github.com/google/uuid"
 )
 
+// Find dynamic map key names passed with Parent["foo"] notation
+var bracketsRe = regexp.MustCompile(`\["([^\["\]]*)"\]`)
```
petar, line 12: This regular expression does not support escape sequences. I wouldn't advice writing your own quoted string parser. It's hard. I would advice using a JS parser library for parsing the ENTIRE glob, e.g. https://pkg.go.dev/github.com/robertkrimen/otto/parser, and then work with the AST.

petar: Alternatively, you can use the Go language parser for the glob. It has a different set of escape sequences than JS, but is still better than rolling your own.

repo/common/common.go
petar: The return value is not documented accurately on a public method. What about dynamicKeys?

repo/common/common.go
```diff
@@ -2,26 +2,62 @@ package common
...
+func ConfigKeyToLookupData(key string) (normalizedKey string, dynamicKeys map[string]string) {
+	bracketedKeys := bracketsRe.FindAllString(key, -1)
+	dynamicKeys = make(map[string]string, len(bracketedKeys))
+	normalizedKey = key
+	for _, mapKeySegment := range bracketedKeys {
+		mapKey := strings.TrimPrefix(mapKeySegment, `["`)
+		mapKey = strings.TrimSuffix(mapKey, `"]`)
+		placeholder := uuid.New().String()
+		dynamicKeys[placeholder] = mapKey
+		normalizedKey = strings.Replace(normalizedKey, mapKeySegment, fmt.Sprintf(".%s", placeholder), 1)
```
petar, line 26: This will convert a glob string like: ["abc"]["def"] into .UUID1.UUID2
Is this your intention? Specifically, I am referring to the first key.

repo/common/common.go
```diff
@@ -2,26 +2,62 @@ package common
...
+}
+
 func MapGetKV(v map[string]interface{}, key string) (interface{}, error) {
 	var ok bool
 	var mcursor map[string]interface{}
 	var cursor interface{} = v
 
-	parts := strings.Split(key, ".")
+	normalizedKey, dynamicKeys := ConfigKeyToLookupData(key)
+	parts := strings.Split(normalizedKey, ".")
```
petar, line 47: Fails of normalizedKey starts with a ".", which technically it can.

petar (changes requested):

Mitumaru (approved):

lidel:
Closing: no movement since 2021 and the branch no longer applies. Editing the JSON directly, or `ipfs config replace`, remains the workaround for map keys with dots. Anyone picking this up should cover arrays too, per the note above.

PR #8097, opened 2021-04-28
chore: update webui to 2.12.2

### LINKED ISSUE #7559: Add support for remote Pinning Services

Stebalien:
Closes #7559

### DIFF
--- a/core/corehttp/webui.go
+++ b/core/corehttp/webui.go
@@ -1,7 +1,7 @@
 package corehttp
 
 // TODO: move to IPNS
-const WebUIPath = "/ipfs/bafybeid3rof2bfszwia2g7ijjzac5ycaczje3vr3sxdtcxziaf6fvi6mwu" // v2.12.1
+const WebUIPath = "/ipfs/bafybeifuexpvt6g4bkbmsrlb7rnudskfakn6vdrtoja4ml4zbxne2hu6bq" // v2.12.2
 
 // this is a list of all past webUI paths.
 var WebUIPaths = []string{
--- a/core/corehttp/webui.go
+++ b/core/corehttp/webui.go
@@ -1,11 +1,12 @@
 package corehttp
 
 // TODO: move to IPNS
-const WebUIPath = "/ipfs/bafybeif4zkmu7qdhkpf3pnhwxipylqleof7rl6ojbe7mq3fzogz6m4xk3i" // v2.11.4
+const WebUIPath = "/ipfs/bafybeid3rof2bfszwia2g7ijjzac5ycaczje3vr3sxdtcxziaf6fvi6mwu" // v2.12.1
 
 // this is a list of all past webUI paths.
 var WebUIPaths = []string{
 	WebUIPath,
+	"/ipfs/bafybeif4zkmu7qdhkpf3pnhwxipylqleof7rl6ojbe7mq3fzogz6m4xk3i",
 	"/ipfs/bafybeianwe4vy7sprht5sm3hshvxjeqhwcmvbzq73u55sdhqngmohkjgs4",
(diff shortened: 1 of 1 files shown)


lidel:
It seems to work fine with ipfs-desktop first https://github.com/ipfs/ipfs-desktop/pull/1807 but we need  v2.12.2 with https://github.com/ipfs/ipfs-webui/pull/1772 / #8096 before this can land.

lidel:
Ongoing testing:
- https://webui.ipfs.io 
- ipfs-desktop-v0.15.0-rc1: https://github.com/ipfs/ipfs-desktop/issues/1812

lidel:
FYSA ipfs-desktop shipped with ipfs-webui v2.12.2 (bumped this PR to the same version). 
I wanted to wait a bit to test this in the wild, and looks ok. There seem to be no fires or huge regressions.

@Stebalien @aschmahmann 
I feel we should ship this with 0.9.0 just to avoid having different UI in go-ipfs  and ipfs-desktop.
If we keep old one, users will be asking why they are missing "remote pinning"  UI or "why Files header looks different", and it is better to spend time in fixing latest webui (if needed) than triaging those empty issues.

Stebalien:
@lidel SGTM. Want to approve this PR?

lidel (approved):
:sparkles:

PR #8111, opened 2021-05-05
feat(gw): expose /api/v0/dag/export on gateway port

olizilla:
- add dag/export to read-only root. it's ready only, so why the heck not!
- test that i didn't expose other dag subcommands

License: MIT

### DIFF
--- a/core/commands/commands_test.go
+++ b/core/commands/commands_test.go
@@ -26,6 +26,7 @@ func TestROCommands(t *testing.T) {
 		"/dag/get",
 		"/dag/resolve",
 		"/dag/stat",
+		"/dag/export",
 		"/dns",
 		"/get",
 		"/ls",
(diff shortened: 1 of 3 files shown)


olizilla:
is there some magic to running a single sharness test (on macos)? I tried following the readme in the test dir but got errors.
```console
$ make test_sharness_deps
$ cd test/sharness
$ ./t0110-gateway.sh                                                                                                                                                       15:35:40
lib/sharness/sharness.sh: line 888: syntax error: unexpected end of file
Cannot source: lib/sharness/sharness.sh
Please check Sharness installation.
FATAL: Unexpected exit with code 1
```

lidel (approved):
LGTM, CI is green :mage: :sparkles:  

This will effectively enable everyone to fetch CID from any gateway in a way that enables deterministic CID validation (https://github.com/ipfs/in-web-browsers/issues/128), but without waiting for new fancy semantics discussed in https://github.com/ipfs/in-web-browsers/issues/170. With this, tools like [ipfs-or-gateway](https://www.npmjs.com/package/ipfs-or-gateway) could always do hash verification.

@aschmahmann I've added this to 0.9.0: this is a small and safe change  that improves the utility of public gateways  without opening any new doors (mind that we already expose `/api/v0/refs` :man_shrugging: ).

aschmahmann:
LGTM this seems like a pretty reasonable thing to do.

PR #8113, opened 2021-05-07
feat: allow key export in online mode

lidel:
This PR enables apps like Brave browser to do import/export without stopping/starting daemon.

Export does not require repo lock, and afaik it is safe to do even when ipfs daemon is running. 

Ref. https://github.com/brave/brave-browser/issues/15422 (cc @spylogsster)

### DIFF
--- a/core/commands/keystore.go
+++ b/core/commands/keystore.go
@@ -169,7 +169,7 @@ path can be specified with '--output=<path>' or '-o=<path>'.
 		if err != nil {
 			return err
 		}
-		if ver > fsrepo.RepoVersion {
+		if ver != fsrepo.RepoVersion {
 			return fmt.Errorf("key export expects repo version (%d) but found (%d)", fsrepo.RepoVersion, ver)
 		}
 
--- a/core/commands/keystore.go
+++ b/core/commands/keystore.go
@@ -12,6 +12,7 @@ import (
 
 	cmds "github.com/ipfs/go-ipfs-cmds"
 	config "github.com/ipfs/go-ipfs-config"
+	keystore "github.com/ipfs/go-ipfs-keystore"
 	oldcmds "github.com/ipfs/go-ipfs/commands"
 	cmdenv "github.com/ipfs/go-ipfs/core/commands/cmdenv"
 	"github.com/ipfs/go-ipfs/core/commands/e"
@@ -150,7 +151,6 @@ path can be specified with '--output=<path>' or '-o=<path>'.
 		cmds.StringOption(outputOptionName, "o", "The path where the output should be stored."),
 	},
 	NoRemote: true,
-	PreRun:   DaemonNotRunning,
 	Run: func(req *cmds.Request, res cmds.ResponseEmitter, env cmds.Environment) error {
 		name := req.Arguments[0]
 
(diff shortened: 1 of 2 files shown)


core/commands/keystore.go
```diff
@@ -150,7 +151,6 @@ path can be specified with '--output=<path>' or '-o=<path>'.
 		cmds.StringOption(outputOptionName, "o", "The path where the output should be stored."),
 	},
 	NoRemote: true,
-	PreRun:   DaemonNotRunning,
```
aschmahmann, line 153: Removing this check is probably fine, but mostly as a function of the keystore currently living on the filesystem and hoping that Go + the OS properly obey filesystem semantics cross platform. If we try and move keystore into some database this could start to fall apart since the database might require a single access point.

As for the keeping the keys on the filesystem, I'm not really a fan, especially since they're currently unencrypted (although that's fairly fixable), so even though it's low on the priority list it'd be a nice thing to deal with. This is especially so because we have no way of revoking IPNS keys (although setting the sequence number to the max value might do the job) or pushing people onto new keys if the old ones are compromised. Thus far IPNS has not had super high usage due to performance issues but as those resolve (IPNS over PubSub, faster DHT lookups, etc.) the idea of protecting the keys better may come into focus.

Are we ok either boxing ourselves into using a keystore that is always readable by multiple processes, or with having online key export potentially not be available in the future?

cc @Stebalien @lidel

Stebalien: So, I _think_ we should be able to do this without boxing ourselves into a corner. We'd probably need something in the future that:

1. Reads the config to find where the keys are actually stored.
2. Possibly connects to the system keyring.

For now, the current code is likely fine. In the future, we'll likely need something to open a keystore without opening the rest of the repo (i.e., read the config to find where the keystore is, then read the keystore). _Some_ keystores may not support this, but that's fine.

aschmahmann: Yep, it'll be a bit more complicated but it seems doable enough for us to go ahead with this now. Some things we may have to worry about are:

- making sure to control how the config fine is modified if it has keystone info in there
- checking the keystone type and for some of them returning something like "error cannot export keys from this keystone while the daemon is running"

Stebalien: Yep. That means checking the repo version is going to be important.

core/commands/keystore.go
```diff
@@ -163,13 +163,15 @@ path can be specified with '--output=<path>' or '-o=<path>'.
 			return err
 		}
 
-		r, err := fsrepo.Open(cfgRoot)
+		// Export is read-only: safe to read it without acquiring repo lock
+		// (this makes export work when ipfs daemon is already running)
+		ksp := filepath.Join(cfgRoot, "keystore")
```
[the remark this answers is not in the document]
lidel: added in 6b63de8a5

core/commands/keystore.go
lidel: Added version check in 6b63de8a5

core/commands/keystore.go
```diff
@@ -163,13 +164,24 @@ path can be specified with '--output=<path>' or '-o=<path>'.
 			return err
 		}
 
-		r, err := fsrepo.Open(cfgRoot)
+		// Check repo version, and error out if not matching
+		ver, err := migrations.RepoVersion(cfgRoot)
+		if err != nil {
+			return err
+		}
+		if ver > fsrepo.RepoVersion {
```
Stebalien, line 172: Hm. I think I got this wrong, actually. We probably need to check for an _exact_ match. I know we've changed the on-disk format before.

Stebalien (approved):
One nit, otherwise LGTM.

Any objections @aschmahmann?

aschmahmann (approved):

PR #8121, opened 2021-05-10
fix(mkreleaselog): handle commit 0

Stebalien:
At commit 0, there is no parent.

### DIFF
--- a/bin/mkreleaselog
+++ b/bin/mkreleaselog
@@ -120,8 +120,10 @@ release_log() {
         "$start..$end" |
         while read commit subject; do
             # Skip gx-only PRs.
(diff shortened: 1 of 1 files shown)


aschmahmann:
@lidel I noticed that there was a failing webui test here. Any idea what it's from/is this already tracked somewhere?

lidel:
@aschmahmann never seen it before, but I noticed that CI uses unsupported node version 12 (current LTS is 14) so that is my main suspect.  
Re-running fixed it, but will be worse over time, so filled  https://github.com/ipfs/go-ipfs/pull/8130 to use LTS version.

PR #8138, opened 2021-05-14
update to Go 1.16.3 on CI

marten-seemann:
Unfortunately, CircleCI doesn't provide a 1.16.4 image, although that version was released 10 days ago. This PR uses 1.16.3 instead.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -130,6 +130,11 @@ jobs:
         TEST_VERBOSE: 1
     steps:
     - run: sudo apt update
+    - run: |
+        mkdir ~/localgo && cd ~/localgo
+        wget https://golang.org/dl/go1.16.4.linux-amd64.tar.gz
+        tar xfz go1.16.4.linux-amd64.tar.gz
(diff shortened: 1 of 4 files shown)


YoutacRSVARs (approved):

gammazero:
@marten-seemann CircleCI now provides a 1.16.5 image.  However, maybe it is best to use the 1.16 tag as that should be an alias for the latest go 1.16.x version. https://circleci.com/developer/images/image/cimg/go

BigLep:
@marten-seemann : are you able to take this?  If not, what are the next steps?

marten-seemann:
I struggled to get our Makefile to build with Go 1.16, but iirc @aschmahmann managed to solve it in a different PR. Is that right?

aschmahmann:
Let's see if we can get this to use 1.16.6. If we run into any issues let's just use the latest available and make sure to note in a new issue that we should start moving testing towards actions to keep the versions of Go in sync with our builders using Actions.

guseggert:
I've picked this up on a different branch, so here's the new PR: https://github.com/ipfs/go-ipfs/pull/8324

Closing this

PR #8139, opened 2021-05-15
remove unneeded GOPATH and GO111MODULE from CircleCI config

marten-seemann:


mvdan (approved):
Note that save_cache uses ~/go/pkg/mod, meaning that the intent was to set GOPATH to ~/go so that the module cache would be put there.

Luckily we now have GOMODCACHE, so you can just set something like `GOMODCACHE: /home/circleci/.cache/go-mod` globally, and then use that in the save/restore cache logic.

aschmahmann (approved):

Issue #8119, opened 2021-05-10
Upstreaming peerlog and time-to-first-byte metrics


aschmahmann:
There are two metrics that Protocol Labs makes use of internally that are in this (historically named) [branch](https://github.com/ipfs/go-ipfs/commits/feat/stabilize-dht) of go-ipfs. It'd be great to upstream these metrics given that they've proven useful so far. We'll also be able to make that branch obsolete and make our release process a bit easier.

There are currently two commits on that branch to remove:
1) Peerlog - which logs all the peerIDs we see. We already have the [peerlog plugin](https://github.com/ipfs/go-ipfs/blob/b8de0e8ad298a6acf96d54611b13ee8abc07f2c8/plugin/plugins/peerlog/peerlog.go#L39) and we just enable it when building the Docker container in this [commit](https://github.com/ipfs/go-ipfs/commit/48f94e29726ae5c57ba13f1bdfa6514ae33ab3e4).
   - If it's easy enough to make this configurable on the deployer/infrastructure side of things instead of as a commit modifying the docker build script that would be great
   - If not then we may want to consider bundling the plugin by default and triggering it via some config flag or environment variable
2) Time-to-First-Byte - It looks like this metric is already defined but not exposed without this [commit](https://github.com/ipfs/go-ipfs/commit/2595a3da516280a7da8614c41a9360549db9f8c8). Perhaps @lanzafame can shed some light into the background here.

PR #8142, opened 2021-05-17
Build a go-ipfs:extras docker image

mburns:
related to #8119

This PR pushes an additional docker image called `go-ipfs:extras` which is built with the `peerlog` plugin enabled by default, and will be a place to test future [in-tree plugins](https://github.com/ipfs/go-ipfs/blob/master/docs/plugins.md#in-tree) or other optional functionality that doesn't need to be in the stock go-ipfs image.

- [x] update CI, add the build & push stepsd
- [x] retire `bifrost` tag name, `go-ipfs:extras` is best friend now.


aschmahmann:
This looks good and could probably be merged as is. We're going to take a look at moving this to Actions to synchronize Go versions with what we have in dist.ipfs.io and to start moving go-ipfs CI towards actions. cc @guseggert

BigLep:
Per 2021-08-05 discussion: we'll create another task for doing this work in GitHub actions for saner reasoning about versions, etc.

guseggert (approved):

PR #8143, opened 2021-05-17
Improve peer hints for pin remote add

### LINKED ISSUE #7915: pin remote add: improve Peer hints in Pin.origins
go-ipfs 0.8.0-rc2 adds `pin remote` commands (details in  #7559). 

When a new remote pin is added via `pin remote add` `Pin.origins` array is populated with all  multiaddrs of the local node.

## Problem

This is a bit naive, because:
- local node may not have the data
- most of multiaddrs are LAN ones and not diallable from the outside (and if service is in LAN, then it does not need them, mdns/land DHT will do just fine)

## Proposed improvements

1.  Check if pinned root CID  is in local blockstore. If it is not, then send empty `Pin.origins = []` 
    (local node is not origin, we are pinning some external CID, eg. when DAG is too big to fit on local laptop)
    CLI prior art:
    ````console
    $ ipfs block stat --offline $cid
    Error: blockservice: key not found
    ````
2. Filter out IP  multiaddrs  from  loopback addres (https://golang.org/pkg/net/#IP.IsLoopback) and maybe networks designated as local/LAN/private: https://en.wikipedia.org/wiki/Private_network

brianstrauch:
Closes #7915. When running `ipfs pin remote add <cid>`,
1. If CID isn't in blockstore, local node isn't origin, so send `Pin.origins = []`
2. <del>If address is loopback, remove address from `Pin.origins`</del>
3. <del>If address is in [`NoAnnounce`](https://github.com/ipfs/go-ipfs/blob/master/docs/config.md#addressesnoannounce), remove address from `Pin.origins`</del>

### DIFF
--- a/core/commands/pin/remotepin.go
+++ b/core/commands/pin/remotepin.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"io"
-	"net"
 	"sort"
 	"strings"
 	"text/tabwriter"
@@ -25,7 +24,6 @@ import (
 	path "github.com/ipfs/interface-go-ipfs-core/path"
 	"github.com/libp2p/go-libp2p-core/host"
 	peer "github.com/libp2p/go-libp2p-core/peer"
-	ma "github.com/multiformats/go-multiaddr"
 )
 
 var log = logging.Logger("core/commands/cmdenv")
@@ -186,14 +184,7 @@ NOTE: a comma-separated notation is supported in CLI for convenience:
 			if err != nil {
 				return err
 			}
-
-			var filteredAddrs []ma.Multiaddr
-			for _, addr := range addrs {
-				if !isLoopbackAddr(addr) {
-					filteredAddrs = append(filteredAddrs, addr)
-				}
-			}
-			opts = append(opts, pinclient.PinOpts.WithOrigins(filteredAddrs...))
+			opts = append(opts, pinclient.PinOpts.WithOrigins(addrs...))
 		}
 
 		// Execute remote pin request
@@ -253,12 +244,6 @@ NOTE: a comma-separated notation is supported in CLI for convenience:
 			return nil
 		}),
 	},
-}
-
-func isLoopbackAddr(addr ma.Multiaddr) bool {
-	// IP is located at the second index: /ip4/127.0.0.1/tcp/8080
-	ip := net.ParseIP(strings.Split(addr.String(), "/")[2])
-	return ip.IsLoopback()
 }
 
 var listRemotePinCmd = &cmds.Command{
--- a/core/commands/pin/remotepin.go
+++ b/core/commands/pin/remotepin.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"net"
 	"sort"
 	"strings"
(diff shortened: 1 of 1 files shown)


core/commands/pin/remotepin.go
```diff
@@ -240,6 +260,22 @@ NOTE: a comma-separated notation is supported in CLI for convenience:
...
+	ip := net.ParseIP(strings.Split(addr.String(), "/")[2])
+	if ip.IsLoopback() {
+		return false
+	}
+
+	for _, noAnnounceAddr := range noAnnounceAddrs {
+		if strings.HasPrefix(addr.String(), noAnnounceAddr) {
+			return false
+		}
+	}
```
lidel, line 274: Turns out we don't need to do `NoAnnounce` filtering here, it is already applied to `node.PeerHost`, so nothing to do here :)

```suggestion

```

Sidenote:

If we needed this, this would not  work for  all cases, because `NoAnnouce`  are usually definition of specific network submask, the last  segment after `/ipcidr`  being the size of the network – see examples [here](https://github.com/ipfs/go-ipfs-config/blob/0a474258a95d8d9436a49539ad9d7da357b015ab/profile.go#L24-L43)

For futre reference, the  proper way of checking if maddr is blocked is already implemented in [core/node/libp2p/addrs.go](https://github.com/ipfs/go-ipfs/blob/v0.9.0-rc1/core/node/libp2p/addrs.go) in form of `ma.NewFilters` and  `filters.AddrBlocked`

lidel (changes requested):
lgtm, but we should remove `NoAnnounce` filtering, as it is redundant (details below)

lidel (approved):
LGTM, @aschmahmann ready for your final review.

lidel:
Note with feedback we got during today's  triage:

Ok, so (3) is not necessary (already filtered out), (2) is a bit problematic, because in some contexts, those multiaddrs could be useful.

This means to make this low-risk, low-controversy PR we shoul limit its behavior to (1). 


@brianstrauch will you have a moment to refactor it to only do (1) and no multiaddr filtering?

lidel (approved):
LGTM, thank you @brianstrauch!

@aschmahmann  ready for your final review.

core/commands/pin/remotepin.go
```diff
@@ -167,13 +167,19 @@ NOTE: a comma-separated notation is supported in CLI for convenience:
...
-		// use that as a hint and connect back to us (if possible)
+		// If CID in blockstore, add own multiaddrs to the 'origins' array
+		// so pinning service can use that as a hint and connect back to us.
 		node, err := cmdenv.GetNode(env)
 		if err != nil {
 			return err
 		}
-		if node.PeerHost != nil {
+
+		isInBlockstore, err := node.Blockstore.Has(rp.Cid())
```
aschmahmann, line 177: @lidel this seems fine. I don't think we're introducing any new problems on the CoreAPI/Node boundary, but wanted to double check. WDYT?

The boundary issues we generally run into are that IIRC accessing the node directly doesn't respect the `--offline` flag, however I think we may already have that issue given that we use `node.PeerHost` below.

aschmahmann (approved):

PR #8153, opened 2021-05-25
Expose additional migration APIs

gammazero:
Expose migration APIs for reading migration config and creating migration fetchers.  This allows the implementation of commands and external applications that want to retrieve migrations according to the Migration portion of the IPFS config.

This change also moves some functionality that is specific to fetching migrations via IPFS into the `ipfsfetcher` package.

This PR is also a prerequisite for expanding the functionality of #7658


cmd/ipfs/daemon.go
```diff
@@ -292,12 +293,22 @@ func daemonFunc(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment
...
+		migrationCfg, err := migrations.ReadMigrationConfig(cctx.ConfigRoot)
 		if err != nil {
 			return err
 		}
 
-		fetcher, err = getMigrationFetcher(migrationCfg, &cctx.ConfigRoot)
+		// Define function to create IPFS fetcher
+		newIpfsFetcher := func(distPath string) migrations.Fetcher {
+			return ipfsfetcher.NewIpfsFetcher(distPath, 0, &cctx.ConfigRoot)
+		}
```
hsanjuan, line 305: This confused me a lot. `GetMigrationFetcher` is supposed to create the fetchers and takes care of creating the Http one internally. However, the IPFS one needs to be provided directly, and no less than inside a function.

I take that this is primarily to be able to mock it while initializing it with distPath or... ?   The GetMigrationFetcher function as it was (taking care of creating the fetchers as needed) makes more sense to what I would expect such function to do. It is not very useful to have a function that creates fetchers if I have to provide one of them anyways.

hsanjuan (commented):
I take this is mostly moving things around and apart from the comment below I do no see anything problematic. Also note that this is not code I'm familiar with, but it does not seem to change much anyways.

cmd/ipfs/daemon.go
gammazero: Different applications (ipfs daemon, ipfs-update) have different implementations of an IPFS fetcher, so they need to provide their own implementation.  When constructing the Fetcher (which may produce a MultiFetcher) by calling `migrations.GetMigrationFetcher`, depending on the migration config only certain types of fetcher will be needed.  For example, if config said "only use HTTP" or "only use IPFS".

We do not want to supply an already-constructed IPFS fetcher (which may be very expensive if it needs to set up a temporary node). So instead, we supply a function to construct the implementation's IPFS fetcher, only if an IPFS fetcher is needed according to the config.

gammazero:
@hsanjuan This PR is really just moving things around, to make them accessible for doing things like #7658.  I have added a comment in the code to offer an explanation for the confusing construct you pointed out, but if you have any suggestions about a better way to do this (given the explanation) I would love to improve that code.

cmd/ipfs/daemon.go
gammazero: Added [comment in code](https://github.com/ipfs/go-ipfs/pull/8153/files#diff-73163a91b837dff9f7108907361d53ed19b735ca59e5231eed6b49ade275cacfR304-R309) to explain

gammazero:
bump

repo/fsrepo/migrations/ipfsfetcher/ipfsfetcher_test.go
```diff
@@ -110,6 +111,149 @@ func TestInitIpfsFetcher(t *testing.T) {
...
+	}
+	if len(bootstrap) != 2 {
+		t.Fatal("wrong number of bootstrap addresses")
+	}
+	if bootstrap[0] != "/dnsaddr/bootstrap.libp2p.io/p2p/QmcZf59bWwK5XFi76CZX8cbJ4BhTzzA3gU1ZjYZcYW3dwt" {
+		t.Fatal("wrong bootstrap address")
+	}
+}
+
+func makeConfig(configData string) string {
```
masih, line 238: Non-blocker: I can see the code was moved here and this is a general cleanup suggestion unrelated to this PR. Since this is only used in `_test`s, would it make sense to take `*testing.T` as an argument here and:
* fail the tests on error removing the `panic` statements
* use `t.TempDir` for constructing ephemeral directories which would eliminate the need for `defer os.RemoveAll`s across tests.

repo/fsrepo/migrations/ipfsfetcher/ipfsfetcher_test.go
```diff
@@ -110,6 +111,149 @@ func TestInitIpfsFetcher(t *testing.T) {
...
+	}
+	if len(peers) != 1 {
+		t.Fatal("wrong number of peers")
+	}
+	if len(peers[0].Addrs) != 2 {
+		t.Error("wrong number of addrs for first peer")
+	}
+	os.RemoveAll(tmpDir)
+
+	tmpDir = makeConfig(configBadPeers)
```
masih, line 223: nit: this test may be easier to read if broken into two tests or one parameterised test.

repo/fsrepo/migrations/migrations_test.go
```diff
@@ -211,3 +214,212 @@ func createFakeBin(from, to int, tmpDir string) {
...
+	}
+
+	downloadSources = []string{"", ""}
+	_, err = GetMigrationFetcher(downloadSources, "", newIpfsFetcher)
+	if err == nil {
+		t.Fatal("expected error when empty string fetchers specified")
+	}
+}
+
+func makeConfig(configData string) string {
```
masih, line 408: ditto re taking `t`.

repo/fsrepo/migrations/migrations.go
```diff
@@ -107,6 +112,89 @@ func ExeName(name string) string {
...
+	if len(fetchers) == 0 {
+		return nil, errors.New("no sources specified")
+	}
+
+	if len(fetchers) == 1 {
+		return fetchers[0], nil
+	}
+
+	// Wrap fetchers in a MultiFetcher to try them in order
+	return NewMultiFetcher(fetchers...), nil
```
masih, line 195: nit: would this code be clearer using a `switch` over `len(fetchers)`

masih (approved):

repo/fsrepo/migrations/migrations.go
gammazero: Done

repo/fsrepo/migrations/migrations_test.go
gammazero: Done

repo/fsrepo/migrations/ipfsfetcher/ipfsfetcher_test.go
gammazero: Done

repo/fsrepo/migrations/ipfsfetcher/ipfsfetcher_test.go
gammazero: I chose to `panic` because the failures there are not a failure of anything that is being tested, but a programming error.  Changed to use `t.Fatal` since that is more expected and helps readability.

repo/fsrepo/migrations/ipfsfetcher/ipfsfetcher.go
```diff
@@ -277,3 +281,50 @@ func parsePath(fetchPath string) (ipath.Path, error) {
...
+	}
+	err = json.NewDecoder(cfgFile).Decode(&bootstrapCfg)
+	if err != nil {
+		fmt.Fprintln(os.Stderr, "cannot read bootstrap peers from config")
+	} else {
+		bootstrap = bootstrapCfg.Bootstrap
+	}
+
+	if _, err = cfgFile.Seek(0, 0); err != nil {
+		fmt.Fprintln(os.Stderr, err)
```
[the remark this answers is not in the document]
gammazero: Added the comment:
```go
        // If Seek fails, only log the error and continue on to try to read the
        // peering config anyway as it might still be readable
```

PR #8154, opened 2021-05-25
feat: add a text output encoding for the stats provide command

aschmahmann:
This adds a text output encoding to the `ipfs stats provide` command for easier glance value

It looks something like:
```
❯ ipfs stats provide
TotalProvides: 7084
AvgProvideDuration: 15.358954ms
LastReprovideDuration: 58.3776802s
LastReprovideBatchSize: 3542
```

It could be a little prettier, but here's a first stab so at least we have something.

### DIFF
--- a/core/commands/stat_provide.go
+++ b/core/commands/stat_provide.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io"
 	"text/tabwriter"
+	"time"
 
 	humanize "github.com/dustin/go-humanize"
 	cmds "github.com/ipfs/go-ipfs-cmds"
@@ -56,14 +57,18 @@ This interface is not stable and may change from release to release.
 
 			tp := float64(s.TotalProvides)
 			fmt.Fprintf(wtr, "TotalProvides:\t%s\t(%s)\n", humanSI(tp, 0), humanFull(tp, 0))
-			fmt.Fprintf(wtr, "AvgProvideDuration:\t%v\n", s.AvgProvideDuration)
-			fmt.Fprintf(wtr, "LastReprovideDuration:\t%v\n", s.LastReprovideDuration)
+			fmt.Fprintf(wtr, "AvgProvideDuration:\t%v\n", humanDuration(s.AvgProvideDuration))
+			fmt.Fprintf(wtr, "LastReprovideDuration:\t%v\n", humanDuration(s.LastReprovideDuration))
 			lrbs := float64(s.LastReprovideBatchSize)
 			fmt.Fprintf(wtr, "LastReprovideBatchSize:\t%s\t(%s)\n", humanSI(lrbs, 0), humanFull(lrbs, 0))
 			return nil
 		}),
 	},
 	Type: batched.BatchedProviderStats{},
+}
+
+func humanDuration(val time.Duration) string {
+	return val.Truncate(time.Microsecond).String()
 }
 
 func humanSI(val float64, decimals int) string {
--- a/core/commands/stat_provide.go
+++ b/core/commands/stat_provide.go
@@ -55,12 +55,10 @@ This interface is not stable and may change from release to release.
 			wtr := tabwriter.NewWriter(w, 1, 2, 1, ' ', 0)
 			defer wtr.Flush()
 
-			tp := float64(s.TotalProvides)
-			fmt.Fprintf(wtr, "TotalProvides:\t%s\t(%s)\n", humanSI(tp, 0), humanFull(tp, 0))
-			fmt.Fprintf(wtr, "AvgProvideDuration:\t%v\n", humanDuration(s.AvgProvideDuration))
-			fmt.Fprintf(wtr, "LastReprovideDuration:\t%v\n", humanDuration(s.LastReprovideDuration))
-			lrbs := float64(s.LastReprovideBatchSize)
-			fmt.Fprintf(wtr, "LastReprovideBatchSize:\t%s\t(%s)\n", humanSI(lrbs, 0), humanFull(lrbs, 0))
+			fmt.Fprintf(wtr, "TotalProvides:\t%s\n", humanNumber(s.TotalProvides))
+			fmt.Fprintf(wtr, "AvgProvideDuration:\t%s\n", humanDuration(s.AvgProvideDuration))
+			fmt.Fprintf(wtr, "LastReprovideDuration:\t%s\n", humanDuration(s.LastReprovideDuration))
+			fmt.Fprintf(wtr, "LastReprovideBatchSize:\t%s\n", humanNumber(s.LastReprovideBatchSize))
 			return nil
 		}),
 	},
@@ -71,9 +69,19 @@ func humanDuration(val time.Duration) string {
 	return val.Truncate(time.Microsecond).String()
 }
 
+func humanNumber(n int) string {
+	nf := float64(n)
+	str := humanSI(nf, 0)
+	fullStr := humanFull(nf, 0)
+	if str != fullStr {
+		return fmt.Sprintf("%s\t(%s)", str, fullStr)
+	}
+	return str
+}
+
 func humanSI(val float64, decimals int) string {
 	v, unit := humanize.ComputeSI(val)
-	return humanize.SIWithDigits(v, decimals, unit)
+	return fmt.Sprintf("%s%s", humanFull(v, decimals), unit)
 }
 
 func humanFull(val float64, decimals int) string {
--- a/core/commands/stat_provide.go
+++ b/core/commands/stat_provide.go
@@ -5,6 +5,7 @@ import (
 	"io"
 	"text/tabwriter"
 
+	humanize "github.com/dustin/go-humanize"
 	cmds "github.com/ipfs/go-ipfs-cmds"
 	"github.com/ipfs/go-ipfs/core/commands/cmdenv"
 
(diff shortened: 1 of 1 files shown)


core/commands/stat_provide.go
```diff
@@ -46,6 +48,17 @@ This interface is not stable and may change from release to release.
...
 		return nil
 	},
-	Encoders: cmds.EncoderMap{},
-	Type:     batched.BatchedProviderStats{},
+	Encoders: cmds.EncoderMap{
+		cmds.Text: cmds.MakeTypedEncoder(func(req *cmds.Request, w io.Writer, s *batched.BatchedProviderStats) error {
+			wtr := tabwriter.NewWriter(w, 0, 0, 1, ' ', 0)
+			defer wtr.Flush()
+
+			fmt.Fprintf(wtr, "TotalProvides: %d\n", s.TotalProvides)
```
aschmahmann, line 56: We could add commas (or SI units) to the number of provides so they look like 3.5k or 3,542 instead of 3542 (especially useful once the numbers reach into the millions)

core/commands/stat_provide.go
```diff
@@ -46,6 +48,17 @@ This interface is not stable and may change from release to release.
...
+			fmt.Fprintf(wtr, "AvgProvideDuration: %v\n", s.AvgProvideDuration)
```
aschmahmann, line 57: We could use a custom time formatter that restricts the number of significant digits (e.g. no need for mentioning seconds if we're talking about hours and no need for 6 decimal places in the number of milliseconds)

core/commands/stat_provide.go
```diff
@@ -46,6 +48,17 @@ This interface is not stable and may change from release to release.
...
+			fmt.Fprintf(wtr, "LastReprovideDuration: %v\n", s.LastReprovideDuration)
```
lidel, line 58:
```suggestion
			fmt.Fprintf(wtr, "LastReprovideDuration: %s\n", s.LastReprovideDuration)
```

core/commands/stat_provide.go
lidel: Would  [Duration.String() ](https://golang.org/pkg/time/#Duration.String) be enough?

```suggestion
			fmt.Fprintf(wtr, "AvgProvideDuration: %s\n", s.AvgProvideDuration)
```

iirc it will serialize to `1h2m0.3s` string with pretty sane defaults:

> Leading zero units are omitted. As a special case, durations less than one second format use a smaller unit (milli-, micro-, or nanoseconds) to ensure that the leading digit is non-zero. The zero duration formats as 0s. 
> https://golang.org/pkg/time/#Duration.String

core/commands/stat_provide.go
lidel: [CommafWithDigits](https://pkg.go.dev/github.com/dustin/go-humanize#CommafWithDigits) adds separator + rounds decimal part.

SI units could be [ComputeSI](https://pkg.go.dev/github.com/dustin/go-humanize#ComputeSI)  + [CommafWithDigits](https://pkg.go.dev/github.com/dustin/go-humanize#CommafWithDigits).

Personally I find super simplifier and rounded version followed by more explicit number the most useful for eyeballing + seeing if things change over time, something like:

`fmt(3542.123456)` → `3.5k (3,542.1234)`

lidel (changes requested):
I was trying to see how it behaves irl and numbers were stuck at zero:

```
TotalProvides: 0
AvgProvideDuration: 0s
LastReprovideDuration: 0s
LastReprovideBatchSize: 0
```

I thought that might be because the repo I used had English Wikipedia  so a lot of blocks, so I left it for a night and same numbers in the morning.  Looks like stats are not updated for some reason?

Re-tested with new repo (`ipfs init` + all ports set to `0` to avoid conflict with my other node) and found a panic:

```
$ ipfs daemon
....
Daemon is ready
panic: runtime error: integer divide by zero

goroutine 4522104 [running]:
github.com/libp2p/go-libp2p-kad-dht/fullrt.(*FullRT).bulkMessageSend.func1(0xc007bb14b0, 0xc0119c1ca0, 0x2, 0x4, 0xc007bb14a0, 0x0, 0xc0119c1b80, 0x16, 0x16, 0xc01390d7d0, ...)
	pkg/mod/github.com/libp2p/go-libp2p-kad-dht@v0.12.0/fullrt/dht.go:981 +0x451
created by github.com/libp2p/go-libp2p-kad-dht/fullrt.(*FullRT).bulkMessageSend
	pkg/mod/github.com/libp2p/go-libp2p-kad-dht@v0.12.0/fullrt/dht.go:977 +0x256
ipfs daemon  347.53s user 59.88s system 112% cpu 6:00.68 total
```

I was able to reproduce the panic after restarting the daemon, failed with:

```
panic: runtime error: integer divide by zero

goroutine 4450306 [running]:
github.com/libp2p/go-libp2p-kad-dht/fullrt.(*FullRT).bulkMessageSend.func1(0xc005f86240, 0xc003f68800, 0x2, 0x4, 0xc005f86220, 0x0, 0xc003f686e0, 0x16, 0x16, 0xc014aab380, ...)
	pkg/mod/github.com/libp2p/go-libp2p-kad-dht@v0.12.0/fullrt/dht.go:981 +0x451
created by github.com/libp2p/go-libp2p-kad-dht/fullrt.(*FullRT).bulkMessageSend
	pkg/mod/github.com/libp2p/go-libp2p-kad-dht@v0.12.0/fullrt/dht.go:977 +0x256
panic: runtime error: integer divide by zero

goroutine 4450250 [running]:
github.com/libp2p/go-libp2p-kad-dht/fullrt.(*FullRT).bulkMessageSend.func1(0xc005f86240, 0xc003f68700, 0x2, 0x14, 0xc005f86220, 0x0, 0xc003f686e0, 0x16, 0x16, 0xc014aab380, ...)
	pkg/mod/github.com/libp2p/go-libp2p-kad-dht@v0.12.0/fullrt/dht.go:981 +0x451
created by github.com/libp2p/go-libp2p-kad-dht/fullrt.(*FullRT).bulkMessageSend
	pkg/mod/github.com/libp2p/go-libp2p-kad-dht@v0.12.0/fullrt/dht.go:977 +0x256
panic: runtime error: integer divide by zero

goroutine 4450253 [running]:
github.com/libp2p/go-libp2p-kad-dht/fullrt.(*FullRT).bulkMessageSend.func1(0xc005f86240, 0xc003f68760, 0x2, 0xe, 0xc005f86220, 0x0, 0xc003f686e0, 0x16, 0x16, 0xc014aab380, ...)
	pkg/mod/github.com/libp2p/go-libp2p-kad-dht@v0.12.0/fullrt/dht.go:981 +0x451
created by github.com/libp2p/go-libp2p-kad-dht/fullrt.(*FullRT).bulkMessageSend
	pkg/mod/github.com/libp2p/go-libp2p-kad-dht@v0.12.0/fullrt/dht.go:977 +0x256
ipfs daemon  556.69s user 60.15s system 171% cpu 6:00.65 total
```
```

File: core/commands/stat_provide.go (hunk 2 above)
Comment (aschmahmann) [reply core/commands/stat_provide.go:57]: > https://golang.org/pkg/time/#Duration.String

`%v` should be calling `.String()`. In any event Duration.String still doesn't round it just removes 0s, for example https://play.golang.org/p/SV6mPITSsgb.

File: core/commands/stat_provide.go (hunk 1 above)
Comment (aschmahmann) [reply core/commands/stat_provide.go:56]: That seems reasonable, do you have time to do this now?

**Comment by aschmahmann** [comment]
@lidel the panic you encountered is unrelated to this and should be resolved by https://github.com/libp2p/go-libp2p-kad-dht/pull/719. I'll rebase this PR on #8156 to make this one more testable though

File: core/commands/stat_provide.go (hunk 1 above)
Comment (lidel) [reply core/commands/stat_provide.go]: @aschmahmann added PoC in be47c66071384a4a70bc627b38d5f3ac38a67cef but I still get zeroes:

```
TotalProvides:          0  (0)
AvgProvideDuration:     0s
LastReprovideDuration:  0s
LastReprovideBatchSize: 0  (0)
```

Comment (aschmahmann) [reply core/commands/stat_provide.go]: You will get zeroes until you start doing some providing, if you add some files you should provide a bit or just use `ipfs bitswap reprovide` to really get the numbers you want.

Notes:
- We don't cache stats to disk which means every time you restart the results are zero until some providing/reproviding happens
- We do cache when the last reprovide finished which means this will only happen every 12h (by default) unless you manually call `ipfs bitswap reprovide`

**Comment by lidel** [comment]
Good question. I think it is a matter of things growing organically and not having a clear convention from the beginning. 

For new stuff, I feel its most useful to text useful for quick eyeballing, and show unmodified values in JSON, but its just personal preference. Happy to add flag if you feel its more in line with other commands.

Anyway, I rounded Durations to miliseconds, it looks like this:

```
TotalProvides:          73  (73)
AvgProvideDuration:     223ms
LastReprovideDuration:  16.32s
LastReprovideBatchSize: 73  (73)
```

Perhaps we should hide "(val)" if it is the same as SI (values under 1k)

aschmahmann:
> Happy to add --human flag if you feel it's more in line with other commands.

Ya, I guess we might as well for sanity's sake. Although I'd be in favor of pushing us to a place where Text is human readable + flexible and other formats are more machine friendly.

> Anyway, I rounded Durations to miliseconds, it looks like this:

I wouldn't quite do that, as the number of provides get larger the average duration shrinks and can drop into the single digit ms range. e.g. I recently got this log output `finished providing of 1060189 keys. It took 26m46.469890381s with an average of 1.515267ms per provide`

lidel (approved):
Given this is experimental opt-in, should be good enough.
We can refine output when/if we decide to move this outside of Experiments.

I pushed small refactor for text representation (https://github.com/ipfs/go-ipfs/pull/8154/commits/12ebbe5553dec4414fec5725444f1e717af2ed4e):
- hides "(n)" when value is same as SI,
- removed space separator from SI  notation ("3k" instead of "3 k")

PR #8160, opened 2021-05-27
chore: manually add later version of go-yamux

aschmahmann:
@Stebalien is this required to resolve https://github.com/libp2p/go-libp2p-yamux/issues/36 until a new version of go-libp2p is released?


Stebalien (approved):
Yes, but I've just released a go-libp2p-yamux version which we can pull in instead (slightly cleaner).

PR #8166, opened 2021-05-31
feat: print error on bootstrap failure

lidel:
This PR aims to improve UX when bootstrap fails.

It will print ERROR to log if daemon running in online mode has no peers after one minute since start:

```console 
$ ipfs daemon
...
Daemon is ready
2021-05-31T15:28:34.403+0200	ERROR	cmd/ipfs	ipfs/daemon.go:527	failed to bootstrap (no peers found): consider updating Bootstrap or Peering section of your config
```

### DIFF
--- a/cmd/ipfs/daemon.go
+++ b/cmd/ipfs/daemon.go
@@ -515,6 +515,16 @@ func daemonFunc(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment
 	// Give the user heads up if daemon running in online mode has no peers after 1 minute
 	if !offline {
 		time.AfterFunc(1*time.Minute, func() {
+			cfg, err := cctx.GetConfig()
+			if err != nil {
+				log.Errorf("failed to access config: %s", err)
+			}
+			if len(cfg.Bootstrap) == 0 && len(cfg.Peering.Peers) == 0 {
+				// Skip peer check if Bootstrap and Peering lists are empty
+				// (means user disabled them on purpose)
+				log.Warn("skipping bootstrap: empty Bootstrap and Peering lists")
+				return
+			}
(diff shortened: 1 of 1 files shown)


godcong:
Why is the error not a warning? In a private network, there may not be bootstrap at the beginning

cmd/ipfs/daemon.go
```diff
@@ -510,6 +512,23 @@ func daemonFunc(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment
...
+			ipfs, err := coreapi.NewCoreAPI(node)
+			if err != nil {
+				log.Errorf("failed to access CoreAPI: %v", err)
+			}
+			peers, err := ipfs.Swarm().Peers(cctx.Context())
+			if err != nil {
+				log.Errorf("failed to read swarm peers: %v", err)
+			}
+			if len(peers) == 0 {
+				log.Error("failed to bootstrap (no peers found): consider updating Bootstrap or Peering section of your config")
```
gammazero, line 527:
```suggestion
				log.Warn("no peers found during bootstrap: consider updating the Bootstrap or Peering section of your config")
```
This is more appropriate to log as a warning, as there is not an actual failure but probably indicates something is wrong.

gammazero (changes requested):

cmd/ipfs/daemon.go
lidel: Afraid we need to keep Error to guarantee this to be printed to the stdout with default log settings, but I agree we should  warn instead of error when both sections are empty – see https://github.com/ipfs/go-ipfs/pull/8166/commits/904cc737162eed91ff5a172ac65c758387da9100

gammazero (approved):

PR #8171, opened 2021-06-02
Enumerate available encodings for dag get

willscott:
This, i believe, is the basic sketch of what would be needed to late bind connection of the `dag get` subcommand.

It attempts to optimistically load plugins before help, which will work except for some edge cases (like that a config is passed in on stdin that sets which directory to find plugins in).
It then gets the root command via method to provide a way for the commands package to allow the dag package to do finalization of its command struct and attach enumerable encoders.

Thoughts on if this makes sense or alternative constructions, @aschmahmann ?


lidel (commented):
UX sidenote: fysa it is possible to enumerate all codecs in `ipfs cid codecs --numeric`, I think it would be useful to add something like `--supported` there, so user knows which codecs can be processed by the node and the  `ipfs dag get|put` helptext for `input-enc` and `format` could refer to `ipfs cid codecs --supported` for list of available options.

schomatis:
@aschmahmann What's the status of this PR? Asking in relation to the description in https://github.com/ipfs/go-ipfs/issues/8471.

Issue #7939, opened 2021-02-24
Pubsub subscription data encoding


achingbrain:
#### Version information:

```
go-ipfs version: 0.8.0
Repo version: 11
System version: amd64/darwin
Golang version: go1.15.8
```

#### Description:

The HTTP API has a few endpoints that return newline-delimited JSON which is used to stream responses to the client.

One such API is `pubsub/sub`.  The problem here is that the data is sent as an ascii-encoded string with no escaping so can include the newline character code which `ipfs-http-client` then interprets (not unreasonably) as ending the current line.

The object API accepts a `data-encoding` argument that lets you encode binary data in a form of the users' choosing.  The pubsub API (and anything that outputs binary data) should accept a `data-encoding` argument.

Refs: https://github.com/ipfs/js-ipfs/issues/3567

aschmahmann:
@achingbrain can you clarify a bit more what's going on? When we emit JSON binary data it's automatically encoded as base64 so there shouldn't be any issues here with newlines.

e.g.
```
Subscribing node:

curl -X POST "http://localhost:5001/api/v0/pubsub/sub?arg=foo&enc=json"
{"from":"ACQIARIgGbM2BsBIgAmyRUBJGsbSZSJ1kbecCrdFWVJNaOY5GJI=","data":"eAp5","seqno":"Fma/NnTLl2Y=","topicIDs":["foo"]}

Publishing node:
ipfs pubsub pub "foo" "x`ny" // Note: ` is the escape character in PowerShell
```

Base64 decoding `eAp5` gives `x\ny`

achingbrain:
If I use echo with `-e` to interpret the backlash, I get two pubsub messages at the other end - :

```console
$ echo -ne 'he\nllo' | ipfs pubsub pub example
```

```console
{"from":"ACQIARIgWZ+O1WZdgbeglCYBeHuFbWNpZ7GnqTit80jbzL3Md38=","data":"aGU=","seqno":"FmcMdzDnDEM=","topicIDs":["example"]}
{"from":"ACQIARIgWZ+O1WZdgbeglCYBeHuFbWNpZ7GnqTit80jbzL3Md38=","data":"bGxv","seqno":"FmcMdzDnDEQ=","topicIDs":["example"]}
```

The first message contains `he` and the second `llo`, the newline character is lost.

If I run wireshark while I'm doing this, only one HTTP request is made to the HTTP API.

[...]

Issue #5780, opened 2018-11-14
go-ipfs-cmds can support passing stateful encoders


hannahhoward:
# Rationale

This issue comes out of the discussion surrounding https://github.com/ipfs/go-ipfs/pull/5611. As more and more ipfs commands support some form of streaming response, go-ipfs-cmd encoders are likely to get called more than once, and chaining together correctly, particularly in the case of a text encoder, may require some kind of knowledge about what was sent last.

In the LS case, we need to know, when LS-ing multiple directories, whether the latest directory entry/entries streamed is in the same directory as the last entry streamed, so we can insert a directory hash and possibly column headings before we start with a new directory. We've solved this for now by simply sending the Hash of the directory streamed last as an extra field in all streamed entries. This seems less than ideal, since it's really a client level concern.

# Possible Solutions

- Force streaming commands to provide solutions on a case by case basis like the current implementation of `ls`
- Have all cases where this is needed use a PostRun instead of an encoder (does generally break web requests with `encoding=text` -- see #1121 for an example)
- Allow an encoder to see the "last streamed value" -- in the case of `ls` this is all that's needed and could simply be an extra parameter in the function passed to MakeTypedEncoder.
- Allow an encoder to pass arbitrary state along between calls to an encoder. This might be done by adding a return value to the encoder function (in addition to the error) and perhaps adding a StateType field to a the command. The seems harder to implement but perhaps while just passing the last value to `ls` is enough other commands might need to pass different types of state between calls.

kevina:
> This probably belongs in go-ipfs-cmds. Given that a raw Encoder is just:
> ...
> @Stebalien @keks am I missing something here?

I think I am as I wasn't looking in how there where actually used.  Sorry for the noise.

PR #8180, opened 2021-06-08
feat: add 'ipfs multibase' commands

coryschwartz:
This is intended to implement part 1  of https://github.com/ipfs/go-ipfs/issues/7939#issuecomment-833812863

## Usage examples

See https://github.com/ipfs/go-ipfs/pull/8180#issuecomment-872617100


core/commands/multibase.go
```diff
@@ -0,0 +1,86 @@
...
+
+var mbaseEncodeCmd = &cmds.Command{
+	Helptext: cmds.HelpText{
+		Tagline: "Encode file or stdin into multibase string",
+	},
+	Arguments: []cmds.Argument{
+		cmds.FileArg("file", true, false, "data to encode").EnableStdin(),
+	},
+	Options: []cmds.Option{
+		cmds.StringOption(mbaseOptionName, "multibase encoding").WithDefault("identity"),
```
lidel, line 36: `identity` is not the best default, it will  cause some confusion:

```
$ echo lol | ipfs multibase encode
lol

$ echo lol | ipfs multibase decode
Error: selected encoding not supported
```

We should use something that works both ways out of the box, and has safe defaults.
I think base64url is both compact and safe (can be used in everywhere, including url paths) and makes a good default:

```suggestion
		cmds.StringOption(mbaseOptionName, "multibase encoding").WithDefault("base64url"),
```

core/commands/root.go
```diff
@@ -75,6 +75,9 @@ TOOL COMMANDS
   commands      List all available commands
   log           Manage and show logs of running daemon
 
+UTILITIES
+  multibase     Encode and decode files with multibase format
```
lidel, line 79: Could we move it up and include existing `cid` in this new section? 
It will make it easier for people to discover them and their purpose.

```
    block         Interact with raw blocks in the datastore

  ENCODING COMMANDS
    cid           Convert and discover properties of CIDs
    multibase     Encode and decode files and stdin with multibase format

  ADVANCED COMMANDS
    daemon        Start a long-running daemon process
````

File: core/commands/multibase.go [hunk 3]
```diff
@@ -0,0 +1,86 @@
...
+	"strings"
+
+	cmds "github.com/ipfs/go-ipfs-cmds"
+	"github.com/ipfs/go-ipfs/core/commands/cmdenv"
+	mbase "github.com/multiformats/go-multibase"
+)
+
+var MbaseCmd = &cmds.Command{
+	Helptext: cmds.HelpText{
+		Tagline: "Encode and decode files with multibase format",
```
lidel, line 15:
```suggestion
		Tagline: "Encode and decode files or stdin with multibase format",
```

lidel (changes requested):
Thank you @coryschwartz, looks good, some cosmetic asks in comments inline.

Before we merge this PR, it needs tests similar to `test/sharness/t0290-cid.sh`.

Tests should cover:
- [x] encoding and decoding round (`echo foo | ipfs multibase encode -b base32 | ipfs multibase decode` and expect "foo" back) 
  - [x] for a file
  - [x]  for stdin and stdout
- [x]  error behaviors when decoder finds
  -  [x] unknown base  prefix
  -  [x] character outside expected base dictionary

lidel:
Cleaned up commands and added tests.

Usage examples for quick eyeballing:

## ipfs multibase 

```
> ipfs multibase  --help
USAGE
  ipfs multibase - Encode and decode files or stdin with multibase format

SYNOPSIS
  ipfs multibase

SUBCOMMANDS
  ipfs multibase decode <encoded_file> - Decode multibase string
  ipfs multibase encode <file>         - Encode data into multibase string
  ipfs multibase list                  - List available multibase encodings.
```


## ipfs multibase decode

```
> ipfs multibase decode --help
USAGE
  ipfs multibase decode <encoded_file> - Decode multibase string

SYNOPSIS
  ipfs multibase decode [--] <encoded_file>

ARGUMENTS

  <encoded_file> - encoded data to decode

DESCRIPTION

  This command expects multibase inside of a file or via stdin:

    > echo -n hello | ipfs multibase encode -b base16 > file
    > cat file
    f68656c6c6f

    > ipfs multibase decode file
    hello

    > cat file | ipfs multibase decode
    hello
```
## ipfs multibase encode

```
> ipfs multibase encode --help
USAGE
  ipfs multibase encode <file> - Encode data into multibase string

SYNOPSIS
  ipfs multibase encode [-b=<b>] [--] <file>

ARGUMENTS

  <file> - data to encode

OPTIONS

  -b  string - multibase encoding. Default: base64url.

DESCRIPTION

  This command expects a file name or data provided via stdin.

  By default it will use URL-safe base64url encoding,
  but one can customize used base with -b:

    > echo hello | ipfs multibase encode -b base16 > output_file
    > cat output_file
    f68656c6c6f0a

    > echo hello > input_file
    > ipfs multibase encode -b base16 input_file
    f68656c6c6f0a
```

## ipfs multibase list

This is alias/same command as pre-existing `ipfs cid bases`.
Decided to add it here as well for completeness  and better UX.

```
> ipfs multibase list --help
USAGE
  ipfs multibase list - List available multibase encodings.

SYNOPSIS
  ipfs multibase list [--prefix] [--numeric]

OPTIONS

  --prefix   bool - also include the single letter prefixes in addition to the code.
  --numeric  bool - also include numeric codes.
```

```
> ipfs multibase list --prefix --numeric
      0  identity
0    48  base2
b    98  base32
B    66  base32upper
c    99  base32pad
C    67  base32padupper
f   102  base16
F    70  base16upper
k   107  base36
K    75  base36upper
m   109  base64
M    77  base64pad
t   116  base32hexpad
T    84  base32hexpadupper
u   117  base64url
U    85  base64urlpad
v   118  base32hex
V    86  base32hexupper
z   122  base58btc
Z    90  base58flickr
```

lidel (approved):

lidel:
- Note from sync discussion:
  - ok to merge as-is
  - nice to have: add `-o` for outputing   to file without piping (better UX in Powershell). Could be this PR or separate one.

aschmahmann:
@lidel another thing we may want to add here is a `transcode` function so that you don't have to do `ipfs multibase decode | ipfs multibase encode -b=base16`. Not having this makes life pretty painful for environments that don't tolerate binary outputs well (e.g. PowerShell, but I suspect there are others)

core/commands/multibase.go
```diff
@@ -0,0 +1,118 @@
...
+	Run: func(req *cmds.Request, resp cmds.ResponseEmitter, env cmds.Environment) error {
+		if err := req.ParseBodyArgs(); err != nil {
+			return err
+		}
+		files := req.Files.Entries()
+		file, err := cmdenv.GetFileArg(files)
+		if err != nil {
+			return fmt.Errorf("failed to access file: %w", err)
+		}
+		encoded_data, err := ioutil.ReadAll(file)
```
gammazero, line 107: Would it be worth implementing a multibase docoder that would take an `io.Reader` and return one as well, with a signature like `NewDecoder(io.Reader) io.Reader`

Then this could all be done without having to read the entire file into memory.
```go
decReader := mbase.NewDecoder(file)
return resp.emit(decReader)
```

gammazero (commented):
Please consider my comment. I think the correct way to do this is to have a stream decoder, instead of reading the entirety of the (possibly huge) file into memory.

Stebalien:
Are any of the requested changes here _blocking_ or just improvements? If they're just improvements, can we merge and iterate?

gammazero:
Not blocking.  Only asking implementor to take note before merging.

core/commands/multibase.go
lidel: Good question! I filled upstream issue https://github.com/multiformats/go-multibase/issues/44 to discuss this, but out of scope  for this PR.

lidel:
For the sake of shipping basic `ipfs multibase encode|decode` in go-ipfs 0.10, 
I've extracted nice-to-haves to separate issues that can be tackled separately:

- `ipfs multibase transcode` – https://github.com/ipfs/go-ipfs/issues/8336
- streaming `ipfs multibase encode|decode` instead of reading the entirety of the (possibly huge) file into memory – filled  https://github.com/multiformats/go-multibase/issues/44  to discuss adding a streaming API to go-multibase, but I also feel it might overlap with generic thing already tracked in https://github.com/ipfs/go-ipfs/issues/5780 
-  `-o --output <file>` as  alternative to stdout for betetr ergonomics in powershell  – https://github.com/ipfs/go-ipfs/issues/8335

@aschmahmann this PR is ready for your final :eyes:  and  a squash-merge.

core/commands/multibase.go
```diff
@@ -0,0 +1,118 @@
...
+var MbaseCmd = &cmds.Command{
+	Helptext: cmds.HelpText{
+		Tagline: "Encode and decode files or stdin with multibase format",
+	},
+	Subcommands: map[string]*cmds.Command{
+		"encode": mbaseEncodeCmd,
+		"decode": mbaseDecodeCmd,
+		"list":   basesCmd,
+	},
+	Extra: CreateCmdExtras(SetDoesNotUseRepo(true)),
```
aschmahmann, line 23: Going to merge for now as this seems like it wouldn't hurt and matches what `ipfs cid` does.

However, I don't think this really works properly without applying the flag to each subcommand.

@lidel let's revisit during the RC process what we want to do here and figure out if `ipfs cid` has similar problems.

lidel: Continued in  https://github.com/ipfs/go-ipfs/issues/8375#issuecomment-1012182623

PR #8183, opened 2021-06-09
multibase encoding on pubsub

coryschwartz:
* Rather than newline-delimeted string args, data passed on stdin for for `ipfs pubsub pub`
* data and topic is expected to be multibase encoded on the HTTP API

multibase encoding does not remain throughout the entire system, only for publishing. My thinking is that in json responses, the data is already base64-encoded. multibase-encoded url-strings provides some benefit to producers, but there would be a size penalty for subscribers that have double-encoded. I could be convinced otherwise, though.

Following along with https://github.com/ipfs/go-ipfs/issues/7939

This command now produces a single output for subscribers.
```
 echo -ne 'he\nllo' | ipfs pubsub pub example
```
```
{"from":"ACQIARIgrB5VSMrRXB4aSXuIfCImTdHJD3kunXBVTDYGqlQwkYA=","data":"aGUKbGxv","seqno":"FobXgdDUDrQ=","topicIDs":["example"]}
```

the data can be sent as with multibase-encoded url string. The same data, including the topic name, is encoded.

```
 curl -X POST "http://localhost:5003/api/v0/pubsub/pub?arg=uZXhhbXBsZQ&arg=uaGUKbGxv"
```

This produces the same message for subscribers.
```
{"from":"ACQIARIgrB5VSMrRXB4aSXuIfCImTdHJD3kunXBVTDYGqlQwkYA=","data":"aGUKbGxv","seqno":"FobXgdDUDrU=","topicIDs":["example"]}
```

### DIFF
--- a/core/commands/pubsub.go
+++ b/core/commands/pubsub.go
@@ -5,10 +5,14 @@ import (
 	"encoding/binary"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net/http"
+	"os"
 	"sort"
 
 	cmdenv "github.com/ipfs/go-ipfs/core/commands/cmdenv"
+	mbase "github.com/multiformats/go-multibase"
+	"github.com/pkg/errors"
 
 	cmds "github.com/ipfs/go-ipfs-cmds"
 	options "github.com/ipfs/interface-go-ipfs-core/options"
@@ -147,7 +151,25 @@ To use, the daemon must be run with '--enable-pubsub-experiment'.
 	},
 	Arguments: []cmds.Argument{
 		cmds.StringArg("topic", true, false, "Topic to publish to."),
-		cmds.StringArg("data", true, true, "Payload of message to publish.").EnableStdin(),
+		cmds.StringArg("data", false, true, "Payload of message to publish."),
+	},
+	PreRun: func(req *cmds.Request, env cmds.Environment) error {
+		// encode all arguments
+
+		encoder, _ := mbase.EncoderByName("base64")
+		for n, arg := range req.Arguments {
+			req.Arguments[n] = encoder.Encode([]byte(arg))
+		}
+
+		// when there are no string args, read from stdin.
+		if len(req.Arguments) == 1 {
+			buf, err := ioutil.ReadAll(os.Stdin)
+			if err != nil {
+				return err
+			}
+			req.Arguments = append(req.Arguments, encoder.Encode(buf))
+		}
+		return nil
 	},
 	Run: func(req *cmds.Request, res cmds.ResponseEmitter, env cmds.Environment) error {
 		api, err := cmdenv.GetApi(env, req)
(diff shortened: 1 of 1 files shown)


aschmahmann:
> My thinking is that in json responses, the data is already base64-encoded. multibase-encoded url-strings provides some benefit to producers, but there would be a size penalty for subscribers that have double-encoded. I could be convinced otherwise, though.

The data isn't actually base64 encoded, what's happening is that we emit a `pubSubMessage` https://github.com/ipfs/go-ipfs/pull/8183/files#diff-88c3196f5f326c7b97a3be71eab971648fabab00f6f8b9e78a238f43aa99ef50L42-L47 which has fields of types `[]byte` and `string` where everything except for the `Topic` is `[]byte`. Go's JSON marshaller automatically base64 encodes fields of type `[]byte` while leaving strings alone.

The results are:
1) If topic names aren't UTF-8 encoded then you end up with strings interspersed with text like `\u1234` which are no fun for javascript
2) The []byte fields are implicitly base64 encoded instead of being more explicitly multibase encoded

The results from https://github.com/ipfs/go-ipfs/issues/7939#issuecomment-833812863 are that we're going to be making a breaking change to the return types here. This means we can change `pubSubMessage` to return all strings and do the multibase encoding ourselves (base64 still good for space preservation).

Side note: @lidel @achingbrain do you have a preference for the returned multibase encoding? Base64URL, Base64, something else?

lidel:
We need to require base64url for URL params, but for everything else (eg. response)  base64 is fine.
([JS has built-in support for base64](https://developer.mozilla.org/en-US/docs/Glossary/Base64))

lidel (changes requested):
Thank you @coryschwartz, it solves the problem described in  https://github.com/ipfs/go-ipfs/issues/7939#issuecomment-786126897 and we no longer see bogus  message splits on `\n`:

```
$ echo -ne 'he\nllo' | ipfs pubsub pub example-multibase 
```

produces  single msg:

```
$ ipfs pubsub sub example-multibase --enc=json
{"from":"mACQIARIgR7j6xSOlG24jauZSt48b7emK7exFRxoaJvIFwidZFdc","data":"maGUKbGxv","seqno":"mFojhcYxl+kA","topicIDs":["mZXhhbXBsZS1tdWx0aWJhc2U"]
```


Remaining work here:

- [ ] switch url args  to `base64url` (see comment inline)
- [ ]  fix `TestPubSub` by updating things around  [go-ipfs-api/pubsub.go#L34](https://github.com/ipfs/go-ipfs-api/blob/eaf54e8deea2d8ce13ce61d17e6aa499b46c01f1/pubsub.go#L34)
- [ ] fix ipfs/interop tests: [pubsub.js](https://github.com/ipfs/interop/blob/master/test/pubsub.js)   & [ipns-pubsub.js](https://github.com/ipfs/interop/blob/master/test/ipns-pubsub.js) are failing
  - :point_right: :question:  @achingbrain @aschmahmann this is chicken-vs-egg problem, not sure how to handle this.
     -  Should we descope it from this PR and tackle as part of JS tribute work? 
        We need to update `js-ipfs` and  `js-ipfs-http-client` anyway, fixing interop tests could be part of that work. 
     - Is it ok for interop tests to fail in the meantime, or should we park this PR until JS side (incl. interop tests) is done?

aschmahmann (approved):

Issue #6220, opened 2019-04-15
Switch CIDv1 from Base58 to Base32 (go-ipfs)


lidel:
> **This not about making CIDv1 the default**, but a small UX tweak to remove confusion for users who **opt-in to CIDv1 and expect Base32** (think IPFS Camp).
>
> Actual move to CIDv1 as the new default is tracked in:
> - Migration to CIDv1 (default base32)  https://github.com/ipfs/ipfs/issues/337
>   - Make base32 CIDv1 the default for go-ipfs: https://github.com/ipfs/go-ipfs/issues/4143 https://github.com/ipfs/go-ipfs/issues/5358
> - js-ipfs counterpart: https://github.com/ipfs/js-ipfs/issues/1995

## Problem 

Right now if someone wants to use CIDv1 they can opt-in via `--cid-version 1`, but by default they get CIDv1 in **Base58**:

```console
$ ipfs add --cid-version 1 guardian.jpg
added zb2rhk6GMPQF3hfzwXTaNYFLKomMeC6UXdUt6jZKPpeVirLtV guardian.jpg                                                                                                                                                                  
```

To get the representation [we aim for](https://github.com/ipfs/ipfs/issues/337), they need to convert **Base58** CIDv1 to **Base32** in additional step:

```console
$ ipfs cid base32 zb2rhk6GMPQF3hfzwXTaNYFLKomMeC6UXdUt6jZKPpeVirLtV
bafkreigh2akiscaildcqabsyg3dfr6chu3fgpregiymsck7e7aqa4s52zy
```

The need for the conversion step is not only unnecessary, but harmful to our migration efforts.  Users report confusion why CIDv1 produced by go-ipfs can't be used in `<cid>.ipfs.dweb.link` and [why Web UI swaps out the CID](https://user-images.githubusercontent.com/665810/55393420-1d9d8180-5535-11e9-98a4-05055a64721d.png). 

It is _really_ embarrassing when we need to ask people to use `ipfs cid base32` during demos and workshops (_"why do i need to convert? isn't base32 the future default for CIDv1?"_)

## Solution

We should **switch default representation of CIDv1 from Base58 to Base32** (without making CIDv1 the default itself).

This will ensure that:
- one can just `ipfs add --cid-version 1` and use the output immediately 
  (small change, huge UX improvement)
- UX win: CIDv1 look the same everywhere
- our CIDv1 tests can be written against `bafy..` CIDs

### References

[...]

PR #8185, opened 2021-06-11
feat: default new repos to CIDv1

### LINKED ISSUE #247: Migration to CIDv1 (default base32)
Lowercase identifiers are required in order to work with legacy URI, web browser security (sub)origins and the experimental [Protocol Handler API](https://github.com/mozilla/libdweb/issues/2). So we are moving to finish support for CIDv1 across the ecosystem, and use it as default (while still supporting CIDv0/Multihash).

There are some complications related to this switch, such as refactoring the blockstore to use multihashes instead of cid identifiers. A number of temporary solutions were proposed that would have made the change simpler with the current design, however it was decided that since this is going to be a significant update, we should focus on a full CIDv1 migration with one identifier change.

This is the parent issue for all issues related to migration to [CIDv1](https://github.com/ipld/cid) with default base32 (type 'b'). This issue closes when CIDv1b32 is enabled by default for new versions of go-ipfs, js-ipfs, and ipfs-companion.

@kevina @magik6k @lidel @whyrusleeping @diasdavid @alanshaw @olizilla @kyledrake @lgierth @mikeal 

Berlin Jul 2018 developer meeting notes:
https://github.com/ipfs/developer-meetings/pull/39

Initial proposal:
https://github.com/ipfs/go-ipfs/issues/4143

# :bird:  Bird's-eye view (WIP by @lidel)
Located here: https://gist.github.com/lidel/ceef96f60f30308e3c07be95182c4c88

### LINKED ISSUE #4143: Make base32 CIDv1 the default for go-ipfs
I understand there's a switch to [CIDv1](https://github.com/ipld/cid#cidv1) soon. I think go-ipfs should use lowercased base32 (rfc4648 - no padding - highest letter) as the default multibase.

The reason this encoding is preferable: it's the one encoding that will work with subdomains ([RFC1035](http://tools.ietf.org/html/rfc1035) + [RFC1123](https://tools.ietf.org/html/rfc1123)). The restrictions are: case-insensitive, a-b0-9 and less than 63 bytes.

For a slight increase in length, you reap enormous benefits:

* The ability to do proper security origins for the HTTP gateway with subdomains (cidv1abcde.dweb.link). This is *very important* if we want to handle reports with Google's safe browsing system (which is designed for origins). With the current design, all content is on the same browser origin, and a single phishing/malware report on any of the IPFS gateways (hosted by us or someone else) will make web browsers block every single thing on the origin with a [giant red warning message](https://ipfs.io/ipfs/QmQCGCeWwMMHJxD6ZEhX6LT5F5QwrUMjuv9dHZpcVyR5GT) until it's cleared up with Google (which from experience can take *several days*!)
* Root paths are in the right place, which dramatically improves compatibility with existing web sites that tend to do a lot of this:
 ``
* Allows us to register dweb.link (and ipfs.io, etc.) to the [Public Suffix List](https://publicsuffix.org/), which will prevent the sandboxed content from reading/manipulating cookies on the parent domain (and on other cidv1 subdomains).
* Opens up the ability for go-ipfs to do HTTP Host Header parsing and automatic Let's Encrypt support (if we wanted to), so anyone can set up a public IPFS gateway without additional software. Once Let's Encrypt gets their wildcard cert domains shipped (Dec 2017), this could be a fully automated process. Otherwise something like nginx would be needed (I could write an example nginx.conf that people could use for it).

It should use lowercase base32 characters by default, so that it's consistent with subdomain usage (all the browsers will force lowercase). IIRC the RFC doesn't care if it's lowercased, I think people just default to upper case for legacy reasons.

Obviously an abstraction layer could be written that converts between base32 and something else for use with web gateways and then have a different default, but I think it would less confusing for end users to use one default: the one that will let origins in browsers work.

This approach shouldn't be a problem for webextension plugins, but @lidel feel free to chime in.

lidel:
## Problem

Making CIDv1 (base32 `bafy...`) the default has been tracked in #4143 for years, and the groundwork already landed: both CID versions work throughout, the CLI and gateway render base32, and Kubo 0.29's `Import` config (#10421) together with [IPIP-499](https://specs.ipfs.tech/ipips/ipip-0499/)'s `unixfs-v0-2015` / `unixfs-v1-2025` profiles pin the parameters so any implementation can reproduce the same CIDs. The one remaining gap was the default itself: `ipfs init` wrote no `Import.*`, so new repos fell back to CIDv0 `Qm...`.

Some users also lean on CIDs as if they were deterministic hashes. That is a misunderstanding of what a CID is, but breaking them isn't the goal, so existing repos need agency to move to CIDv1 on their own schedule.

## Fix

`ipfs init` now applies the `unixfs-v1-2025` profile to the config it writes, so a fresh repo produces CIDv1 everywhere. This is deliberately the least disruptive way to flip the default:

- **Only fresh repos change.** The switch lives in what `ipfs init` writes, not a code-level fallback. A pre-existing repo with empty or absent `Import.*` still resolves to CIDv0 exactly as before, with no migration and no repo-version bump.
- **Every import path follows the repo's config,** not just `ipfs add`: `ipfs files`, the MFS root, the gateway, and the `client/rpc` Go library all produce the configured CID version. Requesting a version explicitly (`--cid-version`, or a non-sha2-256 hash) uses that version's natural leaves; contradictions like `--cid-version=0 --raw-leaves`, or a config pairing `Import.CidVersion=0` with `Import.UnixFSRawLeaves=true`, are rejected rather than emitting a surprising CID.
- **`Import.*` is the single source of truth,** including whether the startup notice appears. If any CID-profile field is still unset, `ipfs daemon` prints a one-time notice nudging the operator to pin a profile; once explicit, it stops.
- **Users choose, explicitly.** New repos get `unixfs-v1-2025`; `ipfs init --profile unixfs-v0-2015` keeps CIDv0; existing-repo operators are nudged to record their choice rather than lean on an implicit default that could shift across releases.

This respects user agency and avoids breaking legacy datasets, while making CIDv1 what new users get out of the box.

## References

- Context
  - **Spec:** [IPIP-499: UnixFS CID Profiles](https://specs.ipfs.tech/ipips/ipip-0499/) ([ipfs/specs#499](https://github.com/ipfs/specs/pull/499)) defines the `unixfs-v0-2015` and `unixfs-v1-2025` profiles this PR applies.
  - **Enabling work:** [#10421](https://github.com/ipfs/kubo/pull/10421) added the `Import` config section; [#11148](https://github.com/ipfs/kubo/pull/11148) added its IPIP-499 profile fields.
  - **Groundwork:** defaulting to CIDv1 became practical once base32 ([#6220](https://github.com/ipfs/kubo/issues/6220)) and subdomain gateways ([#6498](https://github.com/ipfs/kubo/issues/6498)) shipped.
  - **Supersedes:** [#8886](https://github.com/ipfs/kubo/pull/8886), [#8896](https://github.com/ipfs/kubo/pull/8896), and [#8898](https://github.com/ipfs/kubo/pull/8898), three earlier attempts at CIDv1-by-default.
  - **Interop:** [ipfs/helia#1083](https://github.com/ipfs/helia/pull/1083) updates the Helia interop suite for the new default.
- Closes #4143
- Closes ipfs/specs#247

## Demo

**A fresh repo now defaults to CIDv1.** `ipfs init` writes the `unixfs-v1-2025` profile, so `ipfs add`, `ipfs files`, and the MFS root produce base32 `bafy.../bafk...` out of the box instead of `Qm...`:

```console
$ ipfs init          # writes the unixfs-v1-2025 profile
$ ipfs config Import.CidVersion
1
$ echo -n "hello, CIDv1!" | ipfs add -q
bafkreifriwiigahgxhtea4usek2ew2hr4vtnzsj7m6jy6n5iusfporvs2e   # was QmNbyD5G... before
```

**Existing repos are untouched.** A repo whose `Import.*` is empty or absent keeps producing the exact same CIDs. The only new thing its operator sees is a one-time notice at `ipfs daemon` startup, until a profile is pinned:

```console
$ ipfs daemon
Initializing daemon...
...
⚠️ This repository relies on implicit UnixFS Import.* config defaults,
⚠️ so the CIDs it produces can change between Kubo releases. Pin
⚠️ them by applying a profile (this also silences this notice):
⚠️   $ ipfs config profile apply unixfs-v1-2025   # recommended, modern CIDv1
⚠️   $ ipfs config profile apply unixfs-v0-2015   # legacy, CIDv0 "Qm.." format
⚠️ Use the legacy profile only to keep reproducing existing CIDs.
⚠️ See IPIP-499: https://specs.ipfs.tech/ipips/ipip-0499/
...
Daemon is ready
```

Once the operator applies a profile (or runs `ipfs init --profile ...` on a new repo), `Import.*` is explicit and the notice stops.

## TODO

- [x] land https://github.com/ipfs/helia/pull/1083 first


schomatis:
Similar to https://github.com/ipfs/go-ipfs/pull/8886, leaving sharness CID translation for last.

.github/workflows/interop.yml
```diff
@@ -73,53 +73,21 @@ jobs:
...
+      - name: Check out @helia/interop from ipfs/helia#1083
+        uses: actions/checkout@v7
         with:
-          path: node_modules
-          key: ${{ runner.os }}-helia-interop-${{ steps.helia-version.outputs.version }}
-      - name: Install @helia/interop
-        if: steps.helia-cache.outputs.cache-hit != 'true'
-        run: npm install @helia/interop
-      # Recurring @helia/interop regression: the published .aegir.js is unusable
-      # from inside node_modules, in one of two ways.
```
lidel: - [x] revert this if/once https://github.com/ipfs/helia/pull/1083 lands

core/coreapi/unixfs.go
```diff
@@ -35,13 +37,100 @@ var log = logging.Logger("coreapi")
...
+// (in UnixfsAddOptions), CIDv0 stays dag-pb. `caller` is the caller's options
+// resolved on their own, used only to learn that effective version.
+func importDefaultOptions(imp *config.Import, caller *options.UnixfsAddSettings) ([]options.UnixfsAddOption, error) {
+	var opts []options.UnixfsAddOption
+
+	callerForcesVersion := caller.CidVersionSet || caller.MhType != mh.SHA2_256
+	configCidVer := int(imp.CidVersion.WithDefault(config.LegacyFallbackCidVersion))
+	if !callerForcesVersion || caller.CidVersion == configCidVer {
+		if !imp.CidVersion.IsDefault() {
+			opts = append(opts, options.Unixfs.CidVersion(configCidVer))
```
guillaumemichel: IIUC `add` and MFS disagree when a `CidVersion=0` config meets a non-sha2-256 hash?

With `{"Import": {"CidVersion": 0, "HashFunction": "blake3"}}` (which `ValidateImportConfig` accepts), this emits `CidVersion(0)` and L85 emits `Hash(blake3)` — and `UnixfsAddOptions` rejects that pair outright:

https://github.com/ipfs/kubo/blob/13ad613c5884894da89f157bcdf11f4163f97401/core/coreiface/options/unixfs.go#L130-L131

Whereas `UnixFSCidBuilder` silently bumps v0 → v1 for the same config, so `ipfs files write` succeeds and emits CIDv1. Same repo, same config: `add` errors, MFS quietly changes version.

This inconsistency predates this PR, but it may be a good opportunity to address it?

core/commands/add.go
```diff
@@ -428,16 +393,27 @@ https://github.com/ipfs/kubo/blob/master/docs/config.md#import
...
+		// --cid-version=0 and --raw-leaves contradict: CIDv0 has no raw leaf
+		// codec. Raw leaves at cidVer==0 can only come from an explicit
+		// --raw-leaves here (config raw-leaves is applied by the coreapi, and
+		// only when the effective CID version matches the config's; a config
+		// pinning CidVersion=0 with UnixFSRawLeaves=true is rejected at daemon
+		// start), so this is always a user contradiction.
+		if cidVersionSetByUser && cidVer == 0 && rawblks {
+			return fmt.Errorf("%s can't be true with --cid-version=0 (CIDv0 requires dag-pb leaves)", rawLeavesOptionName)
+		}
+
```
guillaumemichel: Wouldn't it be possible to have the root with CIDv0 and still pass `--raw-leaves`, encoding the leaves using CIDv1? I just want to make sure we don't break any existing use cases.

IIUC the sharness suite already asserts exactly that shape, in files this PR doesn't touch:

- `t0040-add-and-cat.sh:911` — `test_add_cat_5MB --raw-leaves "QmabWSFaPusmiZaaVZLhEUtHcj8CCvVeUfkBpKqAkKVMiS"`
- `t0040-add-and-cat.sh:860-863` — `add_directory '--raw-leaves'` expects `PLANETS="QmfWfQfKCY5Ukv9peBbxM5vqWM9BzmqUSXvdCgjT2wsiBT"` with `MARS`/`VENUS` as `bafkrei...`, i.e. a CIDv0 dag-pb root linking CIDv1 raw leaves
- `t0270-filestore.sh:16` — `EXPHASH="QmXKtATsEt42CF5JoSsmzJstrvwEB5P89YQtdX4mdf9E3M"` from `ipfs add --raw-leaves --nocopy`

And these two still reach that shape without tripping the check:

- `ipfs add --raw-leaves` on a `unixfs-v0-2015` repo — `cidVersionSetByUser` is false, so the guard is skipped
- `ipfs add --cid-version=0 --nocopy` — `--raw-leaves` is never passed so `rawblks` is false, and `UnixfsAddOptions` then auto-enables raw leaves for nocopy

Am I missing something about the intent? If the goal is to steer people toward `unixfs-v1-2025` rather than to forbid the shape outright, a warning might fit better than a hard error. Otherwise we could simply drop the check, the combination is well-defined and already exercised in the suite.

client/rpc/unixfs.go
```diff
@@ -42,7 +42,6 @@ func (api *UnixfsAPI) Add(ctx context.Context, f files.Node, opts ...caopts.Unix
 	req := api.core().Request("add").
 		Option("hash", mht).
 		Option("chunker", options.Chunker).
```
guillaumemichel: IIUC `chunker` and `hash` are still forwarded unconditionally here, so the daemon's `Import` defaults never reach `client/rpc` callers?

L32's `caopts.UnixfsAddOptions(opts...)` materializes the library defaults, so even when the caller passes no options at all, `options.Chunker` is `"size-262144"` and `mht` is `"sha2-256"`, and both go on the wire regardless. But the daemon reads an empty value as "not set, use config":

https://github.com/ipfs/kubo/blob/13ad613c5884894da89f157bcdf11f4163f97401/core/commands/add.go#L487-L498

Since the request always carries a non-empty value, `!= ""` is always true there and the "otherwise `UnixfsAPI.Add` applies the Import config default" path never runs for RPC callers. So on a fresh `unixfs-v1-2025` repo (`UnixFSChunker = "size-1048576"`), `ipfs add` chunks at 1 MiB while `api.Unixfs().Add(ctx, f)` against that same daemon chunks at 256 KiB, leading to different CIDs for anything over 256 KiB.

Would adding `ChunkerSet`/`MhTypeSet` to `UnixfsAddSettings` and gating these two the same way as `CidVersionSet` just below be the right fix? `hash` looks harmless today since both profiles pin sha2-256, but a repo with a custom `Import.HashFunction` would hit the same issue.

core/coreapi/unixfs.go
lidel: Good catch, and yes: `add` refused every import while MFS quietly bumped the version, on a config `ValidateImportConfig` accepted.

Fixed in 5b309d4ac by refusing the config instead of picking a winner: `Import.CidVersion=0` next to a non-sha2-256 `Import.HashFunction` cannot be honored as written, so the node now says so at startup. An unset `CidVersion` is untouched, the hash still selects CIDv1 there, and `add` and MFS agree on every repo that starts.

core/commands/add.go
lidel: You're right, the shape is valid. Dropped the check in 8494f9731, along with the config-level one, which was worse: it stopped the daemon and every offline command on repos that only set `Import.UnixFSRawLeaves=true`, the way to get raw leaves before the profiles existed.

Tests now guard both: the pairing stays accepted in `config/import_test.go`, and such a repo keeps producing its old CIDs in `test/cli/add_test.go`.

client/rpc/unixfs.go
lidel: Exactly right, and that was the fix: 38277e602 adds `ChunkerSet`, `MhTypeSet` and `LayoutSet`, and sends each option only when the caller set it. Layout had the same hole in reverse, an explicit balanced request could not override a trickle config.

`client/rpc/unixfs_test.go` now compares RPC and `ipfs add` CIDs on the same daemon. It fails on the old code.

PR #8207, opened 2021-06-22
chore: update CHANGELOG

aschmahmann:


CHANGELOG.md
```diff
@@ -1,5 +1,716 @@
...
+| Eric Myhre | 82 | +9672/-2459 | 328 |
+| Ian Davis | 7 | +8421/-737 | 116 |
+| Daniel Martí | 18 | +2733/-4377 | 313 |
+| Adin Schmahmann | 46 | +5387/-1289 | 125 |
+| Steven Allen | 95 | +3278/-1861 | 200 |
+| hannahhoward | 14 | +1380/-3667 | 84 |
+| gammazero | 29 | +2520/-1161 | 88 |
+| Hector Sanjuan | 12 | +511/-3129 | 52 |
+| vyzo | 77 | +2198/-940 | 117 |
+| web3-bot | 52 | +1161/-359 | 65 |
```
marten-seemann, line 675: Not sure how we normally deal with bots in the contributor list, but we might want to remove this entry.

aschmahmann: Ya, makes sense I'll just manually drop it for now

aschmahmann: FWIW it's not like this list is totally accurate anyway (includes commits moved between repos, misses new repos, does ok but not great accounting for codegened files, etc.) so not fixing this isn't a big deal

PR #8212, opened 2021-06-23
bump snap to build with Go 1.16

aschmahmann:


olizilla (approved):
This should build out to the `edge` channel. If you're happy that master is close enough to the 0.9.0 release, then we can manually promote the build from the `edge` channel to the `stable` channel. We include the partial commit sha that a releae is from in the build name, which wont match a commit on the release branch, but if it bothers anyone, we can point them to this thread.

olizilla:
ah, from @aschmahmann 
> IIUC promoting won't work here since we'll only see master and that will have a version of v0.10.0-dev instead of v0.9.0

PR #8229, opened 2021-06-30
fix: downgrade mimetype dependency

aschmahmann:


lidel (approved):
I confirmed that this PR fixes the problem (https://github.com/ipfs/go-ipfs/issues/8230).
It makes tests from #8232 pass.

PR #8237, opened 2021-07-01
feat: dag import --stats (#8237)

rvagg:
I'm just taking a stab at this and am unsure if I'm following the right pattern here so feedback would be appreciated.

Currently `dag import` only reports pinned CIDs and an error on pinning. If you `pin-roots=false` then you get nothing. I'm working on the JS version of this, including the HTTP client, and I'd really like to get a bit more information out of this process. The minimum I think is simply a report of how many blocks were imported (although in the current form it's not counting unique blocks, just total processed blocks).

So we end up with this slightly awkward event: `{BlockCount:uint64, Root:*RootMeta}` - where the first one is expected to be `Root==nil` and `BlockCount` telling you how many blocks, and any remaining should have `Root!=nil` and reporting pin status if you asked for pinning.

Is there a better pattern to follow than this? Can I make two separate event types? Can I clean up the struct a bit somehow or maybe use two separate structs rather than overloading one?

### DIFF
--- a/core/commands/dag/dag.go
+++ b/core/commands/dag/dag.go
@@ -16,9 +16,10 @@ import (
 )
 
 const (
+	pinRootsOptionName = "pin-roots"
 	progressOptionName = "progress"
 	silentOptionName   = "silent"
-	pinRootsOptionName = "pin-roots"
+	statsOptionName    = "stats"
 )
 
 // DagCmd provides a subset of commands for interacting with ipld dag objects
@@ -53,10 +54,14 @@ type ResolveOutput struct {
 	RemPath string
 }
 
+type CarImportStats struct {
+	BlockCount uint64
+}
+
 // CarImportOutput is the output type of the 'dag import' commands
 type CarImportOutput struct {
-	BlockCount *uint64   `json:",omitempty"`
-	Root       *RootMeta `json:",omitempty"`
+	Root  *RootMeta       `json:",omitempty"`
+	Stats *CarImportStats `json:",omitempty"`
 }
 
 // RootMeta is the metadata for a root pinning response
@@ -195,8 +200,9 @@ Maximum supported CAR version: 1
 		cmds.FileArg("path", true, true, "The path of a .car file.").EnableStdin(),
 	},
 	Options: []cmds.Option{
-		cmds.BoolOption(silentOptionName, "No output."),
 		cmds.BoolOption(pinRootsOptionName, "Pin optional roots listed in the .car headers after importing.").WithDefault(true),
+		cmds.BoolOption(silentOptionName, "No output."),
+		cmds.BoolOption(statsOptionName, "Output stats."),
 	},
 	Type: CarImportOutput{},
 	Run:  dagImport,
@@ -208,16 +214,19 @@ Maximum supported CAR version: 1
 				return nil
 			}
 
-			// event should have only one of `Root` or `BlockCount` set, not both
+			// event should have only one of `Root` or `Stats` set, not both
 			if event.Root == nil {
-				if event.BlockCount == nil {
+				if event.Stats == nil {
 					return fmt.Errorf("Unexpected message from DAG import")
 				}
-				fmt.Fprintf(w, "Imported %d blocks\n", *event.BlockCount)
+				stats, _ := req.Options[statsOptionName].(bool)
+				if stats {
+					fmt.Fprintf(w, "Imported %d blocks\n", event.Stats.BlockCount)
+				}
 				return nil
 			}
 
-			if event.BlockCount != nil {
+			if event.Stats != nil {
 				return fmt.Errorf("Unexpected message from DAG import")
 			}
 
--- a/core/commands/dag/dag.go
+++ b/core/commands/dag/dag.go
@@ -55,8 +55,8 @@ type ResolveOutput struct {
 
 // CarImportOutput is the output type of the 'dag import' commands
 type CarImportOutput struct {
-	BlockCount uint64
-	Root       *RootMeta
+	BlockCount *uint64   `json:",omitempty"`
+	Root       *RootMeta `json:",omitempty"`
 }
 
 // RootMeta is the metadata for a root pinning response
@@ -208,9 +208,17 @@ Maximum supported CAR version: 1
 				return nil
 			}
 
+			// event should have only one of `Root` or `BlockCount` set, not both
 			if event.Root == nil {
-				fmt.Fprintf(w, "Imported %d blocks\n", event.BlockCount)
+				if event.BlockCount == nil {
+					return fmt.Errorf("Unexpected message from DAG import")
+				}
+				fmt.Fprintf(w, "Imported %d blocks\n", *event.BlockCount)
 				return nil
+			}
+
+			if event.BlockCount != nil {
+				return fmt.Errorf("Unexpected message from DAG import")
 			}
 
 			enc, err := cmdenv.GetLowLevelCidEncoder(req)
--- a/core/commands/dag/dag.go
+++ b/core/commands/dag/dag.go
@@ -55,8 +55,8 @@ type ResolveOutput struct {
 }
 
 type CarImportStats struct {
-	BlockCount        uint64
-	PayloadBytesCount uint64
+	BlockCount      uint64
+	BlockBytesCount uint64
 }
 
 // CarImportOutput is the output type of the 'dag import' commands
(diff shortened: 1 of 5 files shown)


core/commands/dag/dag.go
```diff
@@ -55,7 +55,8 @@ type ResolveOutput struct {
 
 // CarImportOutput is the output type of the 'dag import' commands
 type CarImportOutput struct {
-	Root RootMeta
+	BlockCount uint64
+	Root       *RootMeta
```
ribasushi, line 59: I will defer to @aschmahmann but probably the easiest thing is to `*` them both and tag
```
`json:",omitempty"`
```

rvagg: cool, I didn't know if `*` was even an appropriate thing to do in this situation

rvagg: ok, done that, it's cleaned up the messages pretty nicely, but has introduced an exclusive union which I've added some checks for—mildly awkward but workable

core/commands/dag/import.go
```diff
@@ -176,6 +181,7 @@ func importWorker(req *cmds.Request, re cmds.ResponseEmitter, api iface.CoreAPI,
 				if err := batch.Add(req.Context, nd); err != nil {
 					return err
 				}
+				blockCount++
```
ribasushi, line 184: @rvagg now seeing this and the output, I am wishing for a `payloadBytesCount++` as well 🤔

aschmahmann:
@rvagg is the idea here basically to implement `--progress` like we have in `ipfs pin add` and `ipfs add`? If so we should also hide this behind a flag such that the output is unchanged by default

ribasushi:
> is the idea here basically to implement --progress

No. This implements a single "end-summary" output as the very very last printout, no intermediate progress updates.

lidel (changes requested):
Old response looks like this:

```text
> ipfs dag import foo.car
Pinned root	QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR	success
```
behind the scenes, JSON (HTTP API at `/api/vo/dag/import`) looks like this:
```text
> ipfs dag import foo.car --enc=json | jq
{
  "Root": {
    "Cid": {
      "/": "QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR"
    },
    "PinErrorMsg": ""
  }
}
```

While it is pretty safe to add new fields to JSON, changing the very first line printed as the text output could break people's scripts :see_no_evil: :eyes: 


### TODO before merge 

We discussed this during triage today and  it looks good, but we need to tweak it a bit:
- [ ] we want  `Stats` struct with `Stats.BlockCount` next to the `Root`  so we can add other metrics in the future, eg:
   -  `payloadBytesCount` suggested in https://github.com/ipfs/go-ipfs/pull/8237#discussion_r662812777
   -  time import took
- [x] print `Stats` only when `--stats`  is passed (making this a backward-compatible opt-in feature)

BigLep:
@rvagg : are you able to do the open TODOs?

rvagg:
@BigLep I'll put it on my personal backlog. Having it behind a flag makes me less enthusiastic, though I understand the reasoning.

BigLep:
@gammazero : given your tribute next week, can you please incorporate @lidel's comments and get this over the line?

gammazero:
Done:
- Only print stats when `--stats` flag is passed. This applies to both text and json output encodings.
- Stats data is now contained within a `Stats` data structure
- Stats are printed after root so that the first line of output is the same as previously, even when stats are output
- Updated sharness test to use new stats output and `--stats` flag

gammazero:
I added `PayloadBytesCount` to stats.  Please check that `nd.Size()`, done [here](https://github.com/ipfs/go-ipfs/pull/8237/files#diff-4efdaf370157937a93ea82447a0710a5088e4740f96b6b6fbf966358320b8b43R194), is the correct way to get the payload size.  Was not sure if something from `nd.Stats()` was needed instead.

I did not add an elapsed time stat because I do not know what is the desired format for that duration value (seconds, milliseconds, microseconds, `time.Duration` string,  something else).  If the format is know, it will be trivial to add that stat, but the sharness tests will be more complicated because they will have to ignore the changing elapsed time value.  Unless elapsed time is really useful in the stats, I would suggest omitting it and letting that be determined by external means.

core/commands/dag/dag.go
```diff
@@ -53,9 +54,15 @@ type ResolveOutput struct {
 	RemPath string
 }
 
+type CarImportStats struct {
+	BlockCount        uint64
+	PayloadBytesCount uint64
```
lidel, line 59: `Payload` can be confusing here:  is this raw blocks (CAR-metadata) or actual data in each block (CAR-metadata-dagmetadata)?


Perhaps renaming it to `BlockBytesCount` and setting this to `nd.Stat().BlockSize()` is a way to remove confusion while keeping this useful no matter what codecs are used inside of blocks?

lidel (changes requested):
Sounds good, perfectly fine to keep stats limited to info around imported blocks. 
See comment below.

core/commands/dag/dag.go
gammazero: Renamed `PayloadBytesCount` to `BlockBytesCount`

`nd.Size()` and `nd.Stat().BlockSize` return the same value, so I think it would be better to use `nd.Size()`, given the [comment here](https://github.com/ipfs/go-ipld-format/blob/master/format.go#L40).

lidel: Hm.. if they are the same, it makes sense, but something feels off when I compare Size imported from CAR with value reported by `ipfs dag stat`: 

```console
$ ipfs dag stat bafybeihcyruaeza7uyjd6ugicbcrqumejf6uf353e5etdkhotqffwtguva                                        
Size: 27676801, NumBlocks: 383

$ ipfs dag export bafybeihcyruaeza7uyjd6ugicbcrqumejf6uf353e5etdkhotqffwtguva > test.car                           
 0s  26.41 MiB / ? [--------------------------------------------------------------------------------=-----------------------] 390.25 MiB/s 0s
 
$ ipfs dag import --stats test.car
Pinned root	bafybeihcyruaeza7uyjd6ugicbcrqumejf6uf353e5etdkhotqffwtguva	success
Imported 383 blocks (125832269 bytes)
```

125832269 bytes is ~125 MB which is way more than 26MB

lidel (changes requested):
Remaining thing: https://github.com/ipfs/go-ipfs/pull/8237#discussion_r705587751

core/commands/dag/dag.go
gammazero: @lidel What I did (using `nd.Size()`) worked for the car files in `sharness/t0054-dag-car-import-export-data/` but does not work for your example.  Using `nd.Stat().DataSize + nd.Stat().LinksSize` works for your dag, but not for the test cars/dags.  For the test cars/dags have the stats for almost all blocks as all zeros.

It appears that `nd.Size()` returns `nd.Stat().CumulativeSize` if a block has stat values.  Othersize, `nd.Size()` is set to `len(nd.RawData())`.  This makes both `nd.Size()` and `nd.Stats()` completely unreliable across different dags.

Apparently, the way to get a reliable size is to always use  `len(nd.RawData())`. So, that is what the latest change does.

rvagg: `DataSize` and `LinksSize` sounds like a dag-pb thing, which probably won't be represented in the fixture files.

What stat are we actually after here, @ribasushi what's your expectation of what this sizing is going to report? I would think that it's the size of the output, which includes CAR header, CID lengths and even varint section size prefixes. But I could find that by measuring the size of the output myself, so the utility doesn't seem great. But what is the utility of reporting just the block sizes? What is the useful for?

ribasushi: I would expect the `ipfs dag stat` number. It is useful in terms of "this is the amount of IPLD-data these blocks hold"

rvagg: thanks @ribasushi, `len(nd.RawData())` is probably the one then, but in this case maybe just use `len(block.RawData())` then you get to use the block as it comes out of the CAR rather than whatever happens to it through a `Decode` cycle (probably the same, but seems safer to use the one closer to the original)

rvagg:
I hope you don't mind @gammazero but pushed the `len(block.RawSize())` change. With or without that change the import size is good for @lidel's example data so I think we're good to go on this PR.

```
Pinned root     bafybeihcyruaeza7uyjd6ugicbcrqumejf6uf353e5etdkhotqffwtguva     success
Imported 383 blocks (27676801 bytes)
```

lidel (approved):
LGTM. 

We tested everything with `--stats` so I've added basic regression tests for `dag import` without  `--stats`.
This should be ready for merge.

PR #8244, opened 2021-07-05
perf: set an appropriate capacity

Lasiar:
Improve getting the list of peers

### DIFF
--- a/core/coreapi/swarm.go
+++ b/core/coreapi/swarm.go
@@ -116,18 +116,16 @@ func (api *SwarmAPI) Peers(context.Context) ([]coreiface.ConnectionInfo, error) 
 
 	conns := api.peerHost.Network().Conns()
 
-	var out []coreiface.ConnectionInfo
+	out := make([]coreiface.ConnectionInfo, 0, len(conns))
 	for _, c := range conns {
-		pid := c.RemotePeer()
-		addr := c.RemoteMultiaddr()
 
 		ci := &connInfo{
 			peerstore: api.peerstore,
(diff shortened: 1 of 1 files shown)


core/coreapi/swarm.go
```diff
@@ -116,18 +116,16 @@ func (api *SwarmAPI) Peers(context.Context) ([]coreiface.ConnectionInfo, error)
 
 	conns := api.peerHost.Network().Conns()
 
-	var out []coreiface.ConnectionInfo
+	out := make([]coreiface.ConnectionInfo, 0, len(conns))
```
gammazero, line 119: And if `len(conns) == 0`...
https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices

So, in addition to pre-allocating capacity, how about also handling the 0 conns case and getting rid of the append:
```go
    if len(conns) == 0 {
        return nil, nil
    }
    out := make([]coreiface.ConnectionInfo, len(conns))
    for i, c := range conns {
        out[i] = &connInfo{
            ...
        }
    }
    return out, nil
}
```

gammazero (changes requested):

core/coreapi/swarm.go
Lasiar: Not sure that's needed, I think it's overhead.

the difference within margin of error.

https://gist.github.com/Lasiar/4b50d7d5dcd2f4aea00f8c79a9c9e0af
Screenshot_20210716_001951

Stebalien (approved):

PR #8253, opened 2021-07-08
point ipfs to bitswap/cr integration

petar:
Use the version of Bitswap, which is integrated with composable routing.


Mitumaru (approved):

gammazero:
no longer applicable

Issue #8131, opened 2021-05-12
Flaky MFS sharness test


aschmahmann:
Shareness test `1130 - root hash not bubbled up yet (offline, cidv1) - sharnessLinux.t0250-files-api` failed in this CI run https://app.circleci.com/pipelines/github/ipfs/go-ipfs/4611/workflows/e5849e96-e5cf-4df3-bb50-2b4517f43d24/jobs/51165/tests#failed-test-0.

It appears to be a flaky test related to MFS and using `--flush=false`.

schomatis:
Looking deeper into the surrounding tests. Trying to understand why are we looking for a CIDv0 hash in a test where v1 is enabled:

```
expecting success: 
    test -z "$ONLINE" ||
    (ipfs refs local > refsout &&
    test_expect_code 1 grep $ROOT_HASH refsout)
  
QmageRWxC7wWjPv5p36NeAgBAiFdBHaNfxAehBSwzNech2
test_expect_code: command exited with 0, we wanted 1 grep QmageRWxC7wWjPv5p36NeAgBAiFdBHaNfxAehBSwzNech2 refsout
not ok 1130 - root hash not bubbled up yet (offline, cidv1)
#	
#	    test -z "$ONLINE" ||
#	    (ipfs refs local > refsout &&
#	    test_expect_code 1 grep $ROOT_HASH refsout)
#	  
```

schomatis:
Trying to reproduce a minimal subset of the test that failed. Right now the test is consistenly failing which means I'm likely doing something wrong as this normally passes in sharness (the issue is just a sporadic fail).

Failing right now to avoid flushing the write: `hash is present after write with no flush`.

```bash
#!/bin/bash
set -x
set -e

# The daemon needs to be running, otherwise we will always flush on
# each command (that closes the MFS file descriptors triggering
# the sync).
ipfs swarm addrs > /dev/null || (echo "daemon not running" && exit 1)
# FIXME: Is there a nice way to check this?

HASH=QmNzVQoBR7wQjSNXFrcJHZ29PMsEDfF6iZB1QEhKD4uZpV

# Test if $HASH is present ($1=1) or absent ($1=0) in the repo
# and fail with error string $2 if not.
test_hash() {
  local IS_PRESENT ERROR_STR IN_REPO
  IS_PRESENT=$1
  ERROR_STR=$2

  IN_REPO=1
  # Check if present or not with grep, if grep fails it will
[...]

PR #8254, opened 2021-07-08
Disambiguate online/offline naming in sharness tests

schomatis:
In connection with #8131.

_Does not have any change in logic, just naming. The commits should be reviewed separately; we can drop the second one if too invasive._

Do not use online/offline naming system when starting the daemon in tests but the more specific "with network".

Particularly in the files tests (`t0250`) the online/offline names were _also_ being used to signal if the daemon was running or not. They were even used what appears to be incorrectly as `"online"` meant `"no-daemon"` (now renamed as such) so the internal logic of `tests_for_files_api` should be reviewed (left a `FIXME` note for that).

### DIFF
--- a/test/sharness/lib/test-lib.sh
+++ b/test/sharness/lib/test-lib.sh
@@ -280,6 +280,11 @@ test_launch_ipfs_daemon() {
   '
 }
 
+test_launch_ipfs_daemon_without_network() {
+  test_launch_ipfs_daemon --offline "$@"
+  # FIXME: Check if this is the right way to concatenate arguments.
+}
+
 do_umount() {
   if [ "$(uname -s)" = "Linux" ]; then
   fusermount -u "$1"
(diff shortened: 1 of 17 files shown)


test/sharness/lib/test-lib.sh
```diff
@@ -280,6 +280,11 @@ test_launch_ipfs_daemon() {
   '
 }
 
+test_launch_ipfs_daemon_without_network() {
+  test_launch_ipfs_daemon --offline "$@"
+  # FIXME: Check if this is the right way to concatenate arguments.
```
schomatis, line 285: Blocking.

schomatis: (We can also just drop the argument processing as all calls to `test_launch_ipfs_daemon_without_network` right now don't use them; they were just passing the `--offline` flag.)

lidel: `"$@"` expands to `"$1" "$2"` and so on (skipping `$0`), should be ok:
https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#index-_0040

lidel (approved):

PR #8257, opened 2021-07-15
fix dag test fixtures for go-ipld-prime codecs

rvagg:
Pending https://github.com/ipld/go-ipld-prime/pull/202 and https://github.com/ipld/go-ipld-prime/pull/203

* I've massaged one of the main fixtures so that it's pre-sorted according to DAG-CBOR rules since go-ipld-prime as yet doesn't apply any sorting, I didn't want to treat that as a blocker for this PR but it may have to be treated as a blocker for release (similar change to https://github.com/ipld/go-ipld-prime/pull/203, but there's deep background here ... cans of worms).
* `dag get` doesn't print a newline now, which I think is the right behaviour since it can now do any codec and newlines are going to be a problem for the stricter binary formats than dag-json, so I've adapted the fixtures.
* The proper DAG-PB `Data` / `Links` changes show up here and are now correct, as are the new (proper) rules for showing bytes with DAG-JSON.
* There's a change near the bottom for `dag stat`, this comes from go-ipld-cbor doing always-64-bit integers, but we're now doing always-smallest-possible now across all our codecs, so go-ipld-cbor is the odd one out (old and new JS codecs do this, and spec says it).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -116,3 +116,4 @@ require (
 )
 
 go 1.14
+replace github.com/ipld/go-ipld-prime => ./../../ipld/go-ipld-prime/
--- a/go.mod
+++ b/go.mod
@@ -58,8 +58,8 @@ require (
 	github.com/ipfs/go-verifcid v0.0.1
 	github.com/ipfs/interface-go-ipfs-core v0.4.0
 	github.com/ipld/go-car v0.3.1
-	github.com/ipld/go-codec-dagpb v1.2.1-0.20210405170603-d0b86f7623c2
-	github.com/ipld/go-ipld-prime v0.10.1-0.20210701102120-b7347f196aa5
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -116,3 +116,4 @@ require (
 )
 
 go 1.14
+replace github.com/ipld/go-ipld-prime => ./../../ipld/go-ipld-prime/
```
rvagg, line 119: removal TODO of course

willscott (approved):
thanks for digging through these!

go.mod
rvagg: replaced with the HEAD commit from https://github.com/ipld/go-ipld-prime/pull/203

which is still temporary

rvagg:
Now using go-ipld-prime master with DAG-JSON sorting and no whitespace. Still having trouble with the full sharness run in Circle but will merge this into the feat/dag-put-ipld-prime when CI is done if it doesn't throw up anything new. The sharness problems can be worked on in that branch.

Issue #8233, opened 2021-06-30
Enable bitswap to handle more requests


aschmahmann:
At the moment go-bitswap has fixed worker count limits for areas where we expect resource limits such as the reading from the blockstore, packaging up messages and responding to peers.

We should either make the worker limits configurable from the config file, or allow them to dynamically shrink/grow to meet network demands.

Proposed steps for now are:
- Make each of go-bitswaps worker counts passable as options
- Expose metrics in prometheus for tracking the size of the backlog for each of the workers to help users figure out if they should be increasing their worker counts
- Make a section in the config file for setting internal/unstable parameters of components
   - Proposed: Internal.Bitswap.*
   - Document how a user might use the prometheus metrics to decide to increase their worker counts

aschmahmann:
We currently have some hacked together branches that cover a bunch of the work here, but they need to be transitioned into something usable in go-ipfs master.

- go-ipfs: 236f89018b8f5fe60e2ad1782deeb101b51fe110
- go-bitswap:  ipfs/go-bitswap@78aabea61d9fff6b6fdda3f31ff97bea18693ab0
- go-peertaskqueue: https://github.com/ipfs/go-peertaskqueue/tree/feat/track-ongoing-jobs (https://github.com/ipfs/go-peertaskqueue/commit/2b501cbd3313394e04b3e124193c2f99c0dad574)
  - Required to actually land the changes with configurable workers, but requires some testing/benchmarking

Next steps:
 - [ ] Add config file support for go-bitswap configurability changes
 - [ ] Test with the existing peertaskqueue changes
    - Ideally these would be testground tests, but doing tests on some production nodes and some local nodes to gauge performance should do an ok job as well

aschmahmann:
closed by #8268

PR #8268, opened 2021-07-16
support bitswap configurability

aschmahmann:
Part of #8233


test/sharness/t0116-prometheus-data/prometheus_metrics
```diff
@@ -1,3 +1,5 @@
+active_block_tasks
+active_tasks
```
aschmahmann: I'd expect all these metrics to be prepended with bitswap_ as the other systems are

core/node/core.go
```diff
@@ -88,10 +89,31 @@ func Dag(bs blockservice.BlockService) format.DAGService {
 }
 
 // OnlineExchange creates new LibP2P backed block exchange (BitSwap)
-func OnlineExchange(provide bool) interface{} {
+func OnlineExchange(cfg *config.Config, provide bool) interface{} {
 	return func(mctx helpers.MetricsCtx, lc fx.Lifecycle, host host.Host, rt routing.Routing, bs blockstore.GCBlockstore) exchange.Interface {
 		bitswapNetwork := network.NewFromIpfsHost(host, rt)
-		exch := bitswap.New(helpers.LifecycleCtx(mctx, lc), bitswapNetwork, bs, bitswap.ProvideEnabled(provide))
+		opts := []bitswap.Option{bitswap.ProvideEnabled(provide)}
+		if cfg.Internal.Bitswap.EngineBlockstoreWorkerCount == 0 {
```
aschmahmann, line 94: If we're going to do it this way (i.e. check for the default value and then pass that into the option as opposed to relying on defaults within go-bitswap) it'll be cleaner if you calculate the values and then do

```
opts := []bitswap.Option{
    bitswap.ProvideEnabled(provide),
    bitswap.EngineBlockstoreWorkerCount(bsWorkerCount),
    bitswap.TaskWorkerCount(taskWorkerCount),
    ...
}
```

core/node/groups.go
```diff
@@ -263,7 +263,7 @@ func Online(bcfg *BuildCfg, cfg *config.Config) fx.Option {
 	shouldBitswapProvide := !cfg.Experimental.StrategicProviding
 
 	return fx.Options(
-		fx.Provide(OnlineExchange(shouldBitswapProvide)),
```
aschmahmann, line 266: All the other functions here seem to do the config processing before throwing it into the option function. Why are we doing this one differently, too long/too many options, etc?

core/commands/bitswap.go
```diff
@@ -155,6 +155,8 @@ var bitswapStatCmd = &cmds.Command{
 				fmt.Fprintf(w, "\t\t%s\n", enc.Encode(k))
 			}
 
+			fmt.Fprintf(w, "\tinboundreqs [%d]\n", s.QueuedRequests)
```
aschmahmann, line 158: Is this still being used/necessary? Is it helpful?

docs/config.md
```diff
@@ -782,6 +808,74 @@ The base64 encoded protobuf describing (and containing) the node's private key.
...
+available block tasks is creating a bottleneck (either due to high-latency block operations,
+or due to high number of block operations per bitswap peer task).
+In such cases, try increasing the `EngineBlockstoreWorkerCount`.
+
+#### `Internal.Bitswap.TaskWorkerCount`
+
+Number of threads sending outgoing messages.
+Used to thottle the number of concurrent send operations.
+
+Type: `int`
```
aschmahmann, line 850: More details are needed in these types e.g. https://github.com/ipfs/go-ipfs/blob/dc2715af68b93611a116b83fca90714e7c5cb50b/docs/config.md#L503

docs/config.md
```diff
@@ -782,6 +808,74 @@ The base64 encoded protobuf describing (and containing) the node's private key.
...
+
+#### `Internal.Bitswap.EngineBlockstoreWorkerCount`
+
+Number of threads for blockstore operations.
+Used to throttle the number of concurrent requests to the block store.
+This number should generally be a low multiple (e.g. 4) of the `TaskWorkerCount`,
```
aschmahmann, line 856: > This number should generally be a low multiple (e.g. 4) of the `TaskWorkerCount`,

Where'd you get this from? I think the defaults we run with are 8 task workers and 128 blockstore workers

docs/config.md
```diff
@@ -782,6 +808,74 @@ The base64 encoded protobuf describing (and containing) the node's private key.
...
+but its optimal value can be informed by the metrics `pending_block_tasks` and `active_block_tasks`.
+
+Type: `int`
+
+#### `Internal.Bitswap.EngineTaskWorkerCount`
+
+Number of worker threads for decision engine task worker.
+Used to throttle the number of (send) tasks scheduled in parallel, therefore
+this number should generally be equal to `TaskWorkerCount`.
```
aschmahmann, line 865: IIUC this number is basically about computing and packaging up responses. If so that might be a better explanation. I'm not sure the above is parseable beyond just sending people to look at the decision engine code in go-bitswap.

docs/config.md
```diff
@@ -782,6 +808,74 @@ The base64 encoded protobuf describing (and containing) the node's private key.
...
+
+Type: `int`
+
+#### `Internal.Bitswap.MaxOutstandingBytesPerPeer`
+
+Maximum number of bytes (across all tasks) pending to be processed and sent to any individual peer.
+This number controls fairness and can very from 250Kb (very fair) to 10Mb (less fair, with more work
+dedicated to peers who ask for more). Values below 250Kb could cause thrashing.
```
aschmahmann, line 873: Were these experimental numbers you came up with?

docs/config.md
```diff
@@ -131,98 +131,124 @@ does (e.g, `"1d2h4m40.01s"`).
...
-- [`Addresses`](#addresses)
+- [The go-ipfs config file](#the-go-ipfs-config-file)
+  - [Profiles](#profiles)
+  - [Types](#types)
+    - [`flag`](#flag)
+    - [`priority`](#priority)
+    - [`strings`](#strings)
+    - [`duration`](#duration)
+  - [Table of Contents](#table-of-contents)
+  - [`Addresses`](#addresses)
```
aschmahmann, line 142: It looks like you regenerated the whole ToC here. It's more verbose this way, but probably fine. Also, perhaps this ToC come before everything else.

@lidel WDYT?

core/node/groups.go
petar: Yeah. It seemed like handling it in the top-level function would spill too much subsystem-specific logic in the top-level.

docs/config.md
petar: I simplified this. I used the "multiple" explanation initially because the TaskWorkerCount threads generate block tasks, hence a multiple. And also the PR you handed off to me had it parameterized like that. But for now I am simply saying it depends on your hardware. Also valid.

docs/config.md
petar: Yeah. They are more of an example to demonstrate the interaction between fairness and thrashing. The 250KB is derived theoretically using the observation that 65KB is one packet, so utilizing anything under 5 packets per send (and then switching to another peer) feels like it may cause thrashing. In the extreme 65Kb (1 packet per send) would likely be thrashing because the block preparation pipeline (which involves a context switch to disk IO) would start costing more than the send itself.

docs/config.md
```diff
@@ -782,6 +808,78 @@ The base64 encoded protobuf describing (and containing) the node's private key.
 
 Type: `string` (base64 encoded)
 
+## `Internal`
+
+### `Internal.Bitswap`
+
+InternalBitswap contains knobs for tuning bitswap resource utilization.
```
lidel, line 815:
```suggestion
`Internal.Bitswap` contains knobs for tuning bitswap resource utilization.
```

docs/config.md
lidel: Regenerated ToC is fine, but if we make such change, it is good time to move it to the very top :+1:

docs/config.md
```diff
@@ -782,6 +808,78 @@ The base64 encoded protobuf describing (and containing) the node's private key.
...
+The knobs (below) document how their value should related to each other.
+Whether their values should be raised or lowered should be determined
+based on the metrics `ipfs_bitswap_active_tasks`, `ipfs_bitswap_pending_tasks`,
+`ipfs_bitswap_pending_block_tasks` and `ipfs_bitswap_active_block_tasks`
+reported by bitswap.
+
+These metrics can be accessed as the prometheus endpoint `API_ADDR/debug/metrics/prometheus`.
```
lidel, line 822: nit: we dont use `API_ADDR` anywhere else in the config docs, maybe just reuse key where API addr is defined?
```suggestion
These metrics can be accessed as the prometheus endpoint at `{Addresses.API}/debug/metrics/prometheus` (default: `http://127.0.0.1:5001/debug/metrics/prometheus`)
```

docs/config.md
```diff
@@ -782,6 +808,78 @@ The base64 encoded protobuf describing (and containing) the node's private key.
...
+In such cases, try increasing the `EngineBlockstoreWorkerCount`.
+If this adjustment still does not increase the throuput of the node, there might
+be hardware limitations like I/O or CPU.
+
+#### `Internal.Bitswap.TaskWorkerCount`
+
+Number of threads sending outgoing messages.
+Used to thottle the number of concurrent send operations.
+
+Type: `integer` (thread count, 0 means default which is 8)
```
lidel, line 855: nit: can we switch those to custom type where `null` / missing indicates the default, as suggested in https://github.com/ipfs/go-ipfs-config/pull/139#discussion_r683504132

docs/config.md
```diff
@@ -782,6 +808,78 @@ The base64 encoded protobuf describing (and containing) the node's private key.
...
+
+#### `Internal.Bitswap.EngineBlockstoreWorkerCount`
+
+Number of threads for blockstore operations.
+Used to throttle the number of concurrent requests to the block store.
+The optimal value can be informed by the metrics `ipfs_bitswap_pending_block_tasks` and `ipfs_bitswap_active_block_tasks`.
+This would be a number that depends on your hardware (I/O and CPU).
+
+Type: `integer` (thread count, 0 means default which is 128)
```
lidel, line 864: https://github.com/ipfs/go-ipfs-config/pull/139#discussion_r683504132

docs/config.md
```diff
@@ -782,6 +808,78 @@ The base64 encoded protobuf describing (and containing) the node's private key.
...
+
+#### `Internal.Bitswap.EngineTaskWorkerCount`
+
+Number of worker threads used for preparing and packaging responses before they are sent out.
+This number should generally be equal to `TaskWorkerCount`.
+
+Type: `integer` (thread count, 0 means default which is 8)
```
lidel, line 871: https://github.com/ipfs/go-ipfs-config/pull/139#discussion_r683504132

docs/config.md
```diff
@@ -782,6 +808,78 @@ The base64 encoded protobuf describing (and containing) the node's private key.
...
+
+#### `Internal.Bitswap.MaxOutstandingBytesPerPeer`
+
+Maximum number of bytes (across all tasks) pending to be processed and sent to any individual peer.
+This number controls fairness and can very from 250Kb (very fair) to 10Mb (less fair, with more work
+dedicated to peers who ask for more). Values below 250Kb could cause thrashing.
+Values above 10Mb open the potential for aggressively-wanting peers to consume all resources and
+deteriorate the quality provided to less aggressively-wanting peers.
+
+Type: `integer` (byte count, 0 means default which is 1MB)
```
lidel, line 881: https://github.com/ipfs/go-ipfs-config/pull/139#discussion_r683504132

docs/config.md
```diff
@@ -782,6 +808,78 @@ The base64 encoded protobuf describing (and containing) the node's private key.
 
 Type: `string` (base64 encoded)
 
+## `Internal`
+
```
lidel, line 812: @aschmahmann @guseggert added the note that sets the tone for "pls dont touch this unless you really know what you are doing"  – feel free to adjust language if its too harsh :) 

```suggestion
## `Internal`

This section includes internal knobs for various subsystems to allow advanced users with big or private infrastructures to fine-tune some behaviors without the need for recompiling go-ipfs.  

**Be aware that making informed change here requires in-depth knowledge and most of users should left this untouched.** 

```

lidel (commented):
Some nits around `config.md`

docs/config.md
```diff
@@ -782,6 +808,78 @@ The base64 encoded protobuf describing (and containing) the node's private key.
...
+If `ipfs_bitswap_pending_tasks` grows and eventually reaches a plateau,
+while at the same time `ipfs_bitswap_active_tasks` is at its maximum,
+the node has reached its resource limits and requests are being dropped (or not serviced).
+Raising resource limits (using the knobs below) could help, assuming the hardware can support the new limits.
+
+The value of `ipfs_bitswap_active_block_tasks` is capped by `EngineBlockstoreWorkerCount`.
+
+The value of `ipfs_bitswap_pending_block_tasks` is indirectly capped by `ipfs_bitswap_active_tasks`, but cannot be predicted
+as it depends on the number of blocks involved in a peer task which can vary.
```
guseggert, line 840: If you know your workload characteristics a priori, you could predict this, maybe "can be hard to predict" instead of "cannot be predicted"?

docs/config.md
lidel: Thanks, updated my diff to include fixes + note about breaking changes.

docs/config.md
lidel: Moved in https://github.com/ipfs/go-ipfs/pull/8268/commits/247a916155bc5b0529e0cfd29a76b528147dbf46

lidel:
Applied suggested changes to `/docs/config.md` and added docs for `optionalInteger` type.

lidel:
- refactor: moved bitswap init and consts with defaults to `core/node/bitswap.go` (#8349)
- rebased this PR on top of master with latest ipld prime (easier to resolve conflicts this way)

docs/config.md
```diff
@@ -782,6 +808,78 @@ The base64 encoded protobuf describing (and containing) the node's private key.
...
+
+The value of `ipfs_bitswap_pending_tasks` is generally capped by the knobs below,
+however its exact maximum value is hard to predict as it depends on task sizes
+as well as number of requesting peers. However, as a rule of thumb,
+during healthy operation this value should oscillate around a "typical" low value
+(without hitting a plateau continuously).
+
+If `ipfs_bitswap_pending_tasks` grows and eventually reaches a plateau,
+while at the same time `ipfs_bitswap_active_tasks` is at its maximum,
+the node has reached its resource limits and requests are being dropped (or not serviced).
```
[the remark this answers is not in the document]
aschmahmann: IIRC currently the requests just pile up indefinitely (unless the peer cancels the request). Going to change the documentation here to match what's going on (i.e. the growing pending tasks when the active tasks are full implies that your node is unable to process active tasks as quickly as they're coming in).

PR #7967, opened 2021-03-09
GitHub: Updates to the issue templates including the new "GitHub Issue Forms"

RubenKelevra:
- Add a CLI reference to the selection with a link
- Add a security issue to the selection with a link
- Create a form instead of a simple text field for bugs

### Important:

Before merging make sure this repo is part of the beta - [application form](https://docs.google.com/forms/d/e/1FAIpQLSea1QJ_9Os2gWpGCtpESjeV6VvghKn0iELvbYo8iQFi9GyFsA/viewform).

PR #8279, opened 2021-07-20
use issue forms

Stebalien:
This change converts go-ipfs to the new issue forms.

Initial implementation by @RubenKelvera in https://github.com/ipfs/go-ipfs/pull/7967. The version here:

1. Converts all forms.
2. Asks fewer questions (for bug reports). We can expand the form later, but I always hated bugzilla bug report forms.

Previews:

- [Bug Report](https://github.com/ipfs/go-ipfs/blob/feat/issue-forms/.github/ISSUE_TEMPLATE/bug-report.yml)
- [Documentation Issue](https://github.com/ipfs/go-ipfs/blob/feat/issue-forms/.github/ISSUE_TEMPLATE/doc.yml)
- [Enhancement](https://github.com/ipfs/go-ipfs/blob/feat/issue-forms/.github/ISSUE_TEMPLATE/enhancement.yml)
- [Feature](https://github.com/ipfs/go-ipfs/blob/feat/issue-forms/.github/ISSUE_TEMPLATE/feature.yml)

You can test them out at https://github.com/ipfs/gh-issue-form-test/issues/new/choose (although I won't keep that up-to-date if we make changes here).


BigLep (approved):
Awesome - thanks for handling this!

.github/ISSUE_TEMPLATE/feature.yml
```diff
@@ -0,0 +1,32 @@
...
+      value: |
+        Suggest a new feature in go-ipfs (the program). If you'd like to suggest an improvement to the IPFS protocol, please discuss it on [the forum](https://discuss.ipfs.io).
+
+        Issues in this repo must be specific, actionable, and well motivated. They should be starting points for _building_ new features, not brainstorming ideas.
+
+        If you have an idea you'd like to discuss, please open a new thread on [the forum](https://discuss.ipfs.io).
+
+        **Example:**
+
+        > Reduce memory usage of `ipfs cat` (specific) by buffering less in ... (actionable). This would let me run go-ipfs on my Raspberry Pi (motivated).
```
aschmahmann, line 17: This is the same as the enhancement example which could confuse people. Maybe something like:

```suggestion
        > Add deduplication-optimized chunking of tar files in `ipfs add` (specific) by examining tar headers ... (actionable). This would let me efficiently store and update many versions of code archives (motivated).
```

.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -0,0 +1,55 @@
...
+        [issues]: https://github.com/ipfs/go-ipfs/issues/new/choose
+        [releases]: https://github.com/ipfs/go-ipfs/releases
+  - type: checkboxes
+    attributes:
+      label: Checklist
+      description: Please verify that you've followed these steps
+      options:
+        - label: This is a bug report, not a question. Ask questions on [discuss.ipfs.io](https://discuss.ipfs.io).
+          required: true
+        - label: I have searched on the [issue tracker](https://github.com/ipfs/go-ipfs/issues?q=is%3Aissue) for my bug.
```
aschmahmann, line 23: We should have this for the other issue types as well

.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -0,0 +1,55 @@
...
+        - ipfs-update or dist.ipfs.io
+        - third-party binary
+        - built from source
+  - type: textarea
+    id: version
+    attributes:
+      label: Version
+      render: Text
+      description: |
+        Enter the output of `ipfs version --all`. If you can't run that command, please include a copy of your [gateway's version page](http://localhost:8080/api/v0/version?enc=text&all=true).
```
aschmahmann, line 45: Looks great. We may want to add something about `ipfs config show` or how to grab pprof dumps here as well, but this is a fine start.

aschmahmann (changes requested):
Looks great, had a few small comments.

.github/ISSUE_TEMPLATE/bug-report.yml
Stebalien: I've added the config. I'm working on a better command for dumping pprofs.

.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -0,0 +1,62 @@
...
+  - type: textarea
+    id: config
+    attributes:
+      label: Config
+      render: json
+      description: |
+        Enter the output of `ipfs config show`.
```
aschmahmann, line 52: I think this is fine. The only issue with `ipfs config show` is that some users might:
a) Not want to show some of their network configuration
b) Might have secrets if they're doing something weird like using the S3 datastore

However, I think we can leave this to their discretion to figure out instead of adding more text to spell it out for them.

aschmahmann (approved):
LGTM. Left a brief comment, but it's really more of a comment than a suggestion.

.github/ISSUE_TEMPLATE/bug-report.yml
Stebalien: Yeah, I agree. But explaining that could be more trouble than it's worth.

PR #8287, opened 2021-07-21
(ipld-prime) Use mh.Set() for "hash" -> mhType & Make git plugin test work again

rvagg:
@mvdan this extends your last commit to the branch to do `mh.Set()` for `hash` as well.

It also revives t0280 for the git-plugin. The diff with master is the most interesting bit now:

```diff
diff --git a/test/sharness/t0280-plugin-git.sh b/test/sharness/t0280-plugin-git.sh
index e7bc06427..d90df46bc 100755
--- a/test/sharness/t0280-plugin-git.sh
+++ b/test/sharness/t0280-plugin-git.sh
@@ -17,7 +17,7 @@ test_expect_success "prepare test data" '

 test_dag_git() {
   test_expect_success "add objects via dag put" '
-    find objects -type f -exec ipfs dag put --format=git --input-enc=zlib {} \; -exec echo \; > hashes
+    find objects -type f -exec ipfs dag put --format=git-raw --input-enc=3145848 --hash=sha1 {} \; -exec echo -n \; > hashes
   '

   test_expect_success "successfully get added objects" '
@@ -25,8 +25,8 @@ test_dag_git() {
   '

   test_expect_success "path traversals work" '
-    echo \"YmxvYiA3ACcsLnB5Zgo=\" > file1 &&
-    ipfs dag get baf4bcfhzi72pcj5cc4ocz7igcduubuu7aa3cddi/object/parents/0/tree/dir2/hash/f3/hash > out1
+    echo -n "{\"/\":{\"bytes\":\"YmxvYiA3ACcsLnB5Zgo=\"}}" > file1 &&
+    ipfs dag get baf4bcfhzi72pcj5cc4ocz7igcduubuu7aa3cddi/Object/Parents/0/GitTree/1/Hash/0/Hash > out1
   '

   test_expect_success "outputs look correct" '
```

Note the pathing through the git object at the bottom there. I wouldn't mind getting that addressed - don't CamelCase, at least one rename, and maybe turn those arrays into maps so we can do it by name. It'd be nice to maintain consistency of forms.

### DIFF
--- a/core/commands/dag/dag.go
+++ b/core/commands/dag/dag.go
@@ -80,7 +80,7 @@ into an object of the specified format.
 		cmds.StringOption("format", "f", "Format that the object will be added as.").WithDefault("dag-cbor"),
 		cmds.StringOption("input-enc", "Format that the input object will be.").WithDefault("dag-json"),
 		cmds.BoolOption("pin", "Pin this object when adding."),
-		cmds.StringOption("hash", "Hash function to use").WithDefault(""),
+		cmds.StringOption("hash", "Hash function to use").WithDefault("sha2-256"),
 	},
 	Run:  dagPut,
 	Type: OutputObject{},
--- a/plugin/plugins/git/git.go
+++ b/plugin/plugins/git/git.go
@@ -31,9 +31,9 @@ func (*gitPlugin) Version() string {
 }
 
 func (*gitPlugin) Init(_ *plugin.Environment) error {
-	// register a custom identifier for import of "zlib-incoded git objects."
+	// register a custom identifier in the reserved range for import of "zlib-encoded git objects."
 	// TODO: give this a name.
-	multicodec.RegisterDecoder(uint64(0x10000000+mc.GitRaw), decodeZlibGit)
+	multicodec.RegisterDecoder(uint64(0x300000+mc.GitRaw), decodeZlibGit)
 	return nil
 }
 
--- a/plugin/plugins/git/git.go
+++ b/plugin/plugins/git/git.go
@@ -33,7 +33,7 @@ func (*gitPlugin) Version() string {
 func (*gitPlugin) Init(_ *plugin.Environment) error {
 	// register a custom identifier in the reserved range for import of "zlib-encoded git objects."
 	// TODO: give this a name.
-	multicodec.RegisterDecoder(uint64(0x300000+mc.GitRaw), decodeZlibGit)
+	multicodec.RegisterDecoder(uint64(mc.ReservedStart+mc.GitRaw), decodeZlibGit)
 	return nil
 }
 
--- a/test/sharness/t0280-plugin-git.sh
+++ b/test/sharness/t0280-plugin-git.sh
@@ -17,19 +17,19 @@ test_expect_success "prepare test data" '
 
 test_dag_git() {
   test_expect_success "add objects via dag put" '
-    find objects -type f -exec ipfs dag put --format=git-raw --input-enc=268435576 {} \; -exec echo \; > hashes
+    find objects -type f -exec ipfs dag put --format=git-raw --input-enc=3145848 --hash=sha1 {} \; -exec echo -n \; > hashes
   '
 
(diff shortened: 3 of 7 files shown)


rvagg:
Noice, Circle giving green now for sharness

core/commands/dag/dag.go
```diff
@@ -80,7 +80,7 @@ into an object of the specified format.
 		cmds.StringOption("format", "f", "Format that the object will be added as.").WithDefault("dag-cbor"),
 		cmds.StringOption("input-enc", "Format that the input object will be.").WithDefault("dag-json"),
 		cmds.BoolOption("pin", "Pin this object when adding."),
-		cmds.StringOption("hash", "Hash function to use").WithDefault(""),
+		cmds.StringOption("hash", "Hash function to use").WithDefault("sha2-256"),
```
mvdan, line 83: I really wish the ipfs flags library allowed using a flag.Value here directly :) then we wouldn't need the string layer in between.

plugin/plugins/git/git.go
```diff
@@ -31,9 +31,9 @@ func (*gitPlugin) Version() string {
 }
 
 func (*gitPlugin) Init(_ *plugin.Environment) error {
-	// register a custom identifier for import of "zlib-incoded git objects."
+	// register a custom identifier in the reserved range for import of "zlib-encoded git objects."
 	// TODO: give this a name.
-	multicodec.RegisterDecoder(uint64(0x10000000+mc.GitRaw), decodeZlibGit)
+	multicodec.RegisterDecoder(uint64(0x300000+mc.GitRaw), decodeZlibGit)
```
mvdan, line 36: I wonder if go-multicodec should export constants for the start and end of the reserved range, so this integer here doesn't need to be hard-coded.

test/sharness/t0280-plugin-git.sh
```diff
@@ -17,19 +17,19 @@ test_expect_success "prepare test data" '
 
 test_dag_git() {
   test_expect_success "add objects via dag put" '
-    find objects -type f -exec ipfs dag put --format=git-raw --input-enc=268435576 {} \; -exec echo \; > hashes
+    find objects -type f -exec ipfs dag put --format=git-raw --input-enc=3145848 --hash=sha1 {} \; -exec echo -n \; > hashes
```
mvdan, line 20: would it not be much easier to use hexadecimal, like 0x300078?

test/sharness/t0280-plugin-git.sh
```diff
@@ -17,19 +17,19 @@ test_expect_success "prepare test data" '
...
+  test_expect_success "successfully get added objects" '
     cat hashes | xargs -I {} ipfs dag get -- {} > /dev/null
   '
 
-  test_expect_failure "path traversals work" '
-    echo \"YmxvYiA3ACcsLnB5Zgo=\" > file1 &&
-    ipfs dag get baf4bcfhzi72pcj5cc4ocz7igcduubuu7aa3cddi/object/parents/0/tree/dir2/hash/f3/hash > out1
+  test_expect_success "path traversals work" '
+    echo -n "{\"/\":{\"bytes\":\"YmxvYiA3ACcsLnB5Zgo=\"}}" > file1 &&
+    ipfs dag get baf4bcfhzi72pcj5cc4ocz7igcduubuu7aa3cddi/Object/Parents/0/GitTree/1/Hash/0/Hash > out1
```
mvdan, line 29: I assume we should fix the git codec first to continue allowing or using lowercase names, to not need to modify this line.

core/commands/dag/dag.go
mvdan: Now raised as https://github.com/ipfs/go-ipfs-cmds/issues/213. Marking this as resolved.

test/sharness/t0280-plugin-git.sh
rvagg: this line change has now been reverted since I pushed to the go-ipld-git `prime` branch as per https://github.com/ipfs/go-ipld-git/pull/46#issuecomment-884854758

PR #8290, opened 2021-07-21
feat: improve mkreleaslog

Stebalien:
1. Allow matching the entire module instead of just github orgs/usernames.
2. Allow excluding some modules.
3. Ignore files using a github pathspec and apply the same ignore patterns to the "contributors" section.

### DIFF
--- a/bin/mkreleaselog
+++ b/bin/mkreleaselog
@@ -149,7 +149,6 @@ ignored_commit() {
     local matches
 
     # Check to see if this commit includes any non-ignored files.
-    # Use jq instead of grep so we get PCRE regexes.
     matches=$(git -C "$dir" diff-tree --no-commit-id --name-only -r "$commit" \
                   -- "${IGNORE_FILES_PATHSPEC[@]}" | wc -l)
     [[ "$matches" -eq 0 ]]
--- a/bin/mkreleaselog
+++ b/bin/mkreleaselog
@@ -1,39 +1,74 @@
 #!/bin/zsh
-#set -x
 set -euo pipefail
 export GO111MODULE=on
 export GOPATH="$(go env GOPATH)"
 
-alias jq="jq --unbuffered"
-
-AUTHORS=(
+# List of PCRE regular expressions to match "included" modules.
+INCLUDE_MODULES=(
     # orgs
-    ipfs
-    ipld
-    libp2p
-    multiformats
-    filecoin-project
-    ipfs-shipyard
+    "^github.com/ipfs/"
+    "^github.com/ipld/"
+    "^github.com/libp2p/"
+    "^github.com/multiformats/"
+    "^github.com/filecoin-project/"
+    "^github.com/ipfs-shipyard/"
 
-    # Authors of personal repos used by go-ipfs that should be mentioned in the
+    # Authors of personal modules used by go-ipfs that should be mentioned in the
     # release notes.
-    whyrusleeping
-    Kubuxu
-    jbenet
-    Stebalien
-    marten-seemann
-    hsanjuan
-    lucas-clemente
-    warpfork
+    "^github.com/whyrusleeping/"
+    "^github.com/Kubuxu/"
+    "^github.com/jbenet/"
+    "^github.com/Stebalien/"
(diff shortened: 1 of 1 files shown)


bin/mkreleaselog
```diff
@@ -1,40 +1,75 @@
...
+    "^github.com/Stebalien/"
+    "^github.com/marten-seemann/"
+    "^github.com/hsanjuan/"
+    "^github.com/lucas-clemente/"
+    "^github.com/warpfork/"
+)
+
+# List of PCRE regular expressions to match "excluded" modules. Applied after includes.
+EXCLUDE_MODULES=(
+    "^github.com/marten-seemann/qtls"
```
marten-seemann, line 30: This needs to be a regex. For example, for Go 1.16 the repo is https://github.com/marten-seemann/qtls-go1-16.

bin/mkreleaselog
```diff
@@ -109,6 +144,17 @@ pr_link() {
     printf -- "[%s#%s](https://%s/pull/%s)" "$ghname" "$prnum" "$repo" "$prnum"
 }
 
+ignored_commit() {
+    local commit="$1"
+    local matches
+
+    # Check to see if this commit includes any non-ignored files.
+    # Use jq instead of grep so we get PCRE regexes.
```
aschmahmann, line 152: Maybe I'm just being bad at bash, but where's the jq here?

bin/mkreleaselog
Stebalien: It's a regex. It'll match anything that starts with `github.com/marten-seemann/qtls`.

bin/mkreleaselog
Stebalien: Ah, that use of jq was stillborn. Let me fix the comment.

bin/mkreleaselog
```diff
@@ -50,7 +85,7 @@ statlog() {
     fi
 
     local stack=()
-    git -C "$rpath" -c mailmap.file="$mailmap_file" log --use-mailmap --shortstat --no-merges --pretty="tformat:%H%x09%aN%x09%aE" "$start..$end" -- . ':^*\.pb\.go' ':^*ipldsch*\.go' ':^cbor_gen\.go\' ':^go\.mod' ':^go\.sum' | while read -r line; do
+    git -C "$rpath" -c mailmap.file="$mailmap_file" log --use-mailmap --shortstat --no-merges --pretty="tformat:%H%x09%aN%x09%aE" "$start..$end" -- . "${IGNORE_FILES_PATHSPEC[@]}" | while read -r line; do
```
aschmahmann, line 88: I don't think `IGNORE_FILES_PATHSPEC` usage is quite right. What we want is to ignore all files named go.mod, go.sum, etc. not just the ones in the root directory.

If you try this out by diffing v0.9.0 against v0.9.1 (by manually setting the start and end in the `recursive_release_log()` function) you'll see Marten having a huge diff that's due to updating go.sum and go.mod files in the go-libp2p examples repo. If you add a `.*` in front of go.mod and go.sum it goes away.

aschmahmann: My regex isn't great, but maybe something like `^(?:.*\/|)go\.mod$`

Stebalien: That's strange... these are git "pathspecs" (not regexps) which should match at any point _unless_ they start with `/` (or `:/`). I'll dig into this.

aschmahmann: Git pathspecs aren't quite regex though. Not sure if there's a better way to write both `go\.mod` and `**/go\.mod` together

bin/mkreleaselog
marten-seemann: 🤦‍♂️, sorry.

bin/mkreleaselog
Stebalien: Hm. You're right. The "starts with /" rule only applies when running git commands in subdirectories (i.e., matches files in the root instead of the subdirectory).

I've fixed this to match files anywhere.

aschmahmann (approved):
LGTM. Thank you 🙏

PR #8291, opened 2021-07-21
feat: add an "ipfs diag profile" command

Stebalien:
This should replace the "collect-profiles.sh" script and allow users to easily collect profiles.

Usage: `ipfs diag profile` or `ipfs diag profile -o my-profile.zip`.

At the moment, it just dumps all profiles into a single zip file. It does this server-side so it's easy fetch them with curl.

In the future, it would be nice to add:

1. Progress indicators (cpu profiles take 30 seconds).
2. An option to specify which profiles to collect.

But we can handle that later.

Unfortunately, this command doesn't produce symbolized svgs as I didn't want to depend on having a local go compiler.

### DIFF
--- a/core/commands/profile.go
+++ b/core/commands/profile.go
@@ -38,6 +38,9 @@ Collects cpu, heap, and goroutine profiles from a running go-ipfs daemon
 into a single zipfile. To aid in debugging, this command also attempts to
 include a copy of the running go-ipfs binary.
 
+Profile's can be examined using 'go tool pprof', some tips can be found at
+https://github.com/ipfs/go-ipfs/blob/master/docs/debug-guide.md.
+
 Privacy Notice:
 
 The output file includes:
--- a/core/commands/profile.go
+++ b/core/commands/profile.go
@@ -30,7 +30,7 @@ var sysProfileCmd = &cmds.Command{
 		Tagline: "Collect a performance profile for debugging.",
 		ShortDescription: `
 Collects cpu, heap, and goroutine profiles from a running go-ipfs daemon
-into a single zipfile. To aid in debugging, this command also attempts to
+into a single zip file. To aid in debugging, this command also attempts to
 include a copy of the running go-ipfs binary.
 `,
 		LongDescription: `
--- a/core/commands/profile.go
+++ b/core/commands/profile.go
@@ -13,8 +13,6 @@ import (
 	"time"
 
 	cmds "github.com/ipfs/go-ipfs-cmds"
-	"github.com/ipfs/go-ipfs/core"
-	"github.com/ipfs/go-ipfs/core/commands/cmdenv"
 	"github.com/ipfs/go-ipfs/core/commands/e"
 )
 
@@ -40,14 +38,9 @@ var sysProfileCmd = &cmds.Command{
 			return fmt.Errorf("failed to parse CPU profile duration %q: %w", cpuProfileTimeStr, err)
 		}
 
-		nd, err := cmdenv.GetNode(env)
-		if err != nil {
-			return err
-		}
-
 		r, w := io.Pipe()
 		go func() {
-			_ = w.CloseWithError(writeProfiles(req.Context, nd, cpuProfileTime, w))
+			_ = w.CloseWithError(writeProfiles(req.Context, cpuProfileTime, w))
 		}()
 		return res.Emit(r)
 	},
@@ -88,7 +81,7 @@ var sysProfileCmd = &cmds.Command{
 	},
 }
 
-func writeProfiles(ctx context.Context, nd *core.IpfsNode, cpuProfileTime time.Duration, w io.Writer) error {
+func writeProfiles(ctx context.Context, cpuProfileTime time.Duration, w io.Writer) error {
 	archive := zip.NewWriter(w)
 
 	// Take some profiles.
@@ -135,17 +128,16 @@ func writeProfiles(ctx context.Context, nd *core.IpfsNode, cpuProfileTime time.D
 		}
 	}
 
-	// Collect info
+	// Collect version info
+	// I'd use diag sysinfo, but that includes some more sensitive information
+	// (GOPATH, etc.).
 	{
-		out, err := archive.Create("sysinfo.json")
+		out, err := archive.Create("version.json")
 		if err != nil {
 			return err
 		}
-		info, err := getInfo(nd)
-		if err != nil {
-			return err
-		}
-		err = json.NewEncoder(out).Encode(info)
+
+		err = json.NewEncoder(out).Encode(getVersionInfo())
 		if err != nil {
 			return err
 		}
--- a/core/commands/profile.go
+++ b/core/commands/profile.go
@@ -26,6 +26,43 @@ type profileResult struct {
 const cpuProfileTimeOption = "cpu-profile-time"
 
 var sysProfileCmd = &cmds.Command{
+	Helptext: cmds.HelpText{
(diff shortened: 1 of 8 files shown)


Stebalien:
(still need to write a sharness test)

test/sharness/t0152-profile.sh
```diff
@@ -0,0 +1,40 @@
...
+test_expect_success "test profiling with -o (without CPU profiling)" '
+  ipfs diag profile --cpu-profile-time=0 -o test-profile.zip
+'
+
+test_expect_success "test that test-profile.zip exists" '
+  test -e test-profile.zip
+'
+
+test_kill_ipfs_daemon
+test_done
```
aschmahmann, line 40: If it's not a pain let's add a basic `go tool pprof` test that verifies the data output is actually viable.

core/commands/profile.go
```diff
@@ -0,0 +1,234 @@
...
+		}
+
+		err = json.NewEncoder(out).Encode(getVersionInfo())
+		if err != nil {
+			return err
+		}
+	}
+
+	// Collect binary
+	if fi, err := openIPFSBinary(); err == nil {
```
aschmahmann, line 184: I don't know if this is worth doing by default. We have the version info and it bloats the size of the zip. WDYT?

core/commands/profile.go
```diff
@@ -0,0 +1,234 @@
...
+}
+
+const cpuProfileTimeOption = "cpu-profile-time"
+
+var sysProfileCmd = &cmds.Command{
+	Helptext: cmds.HelpText{
+		Tagline: "Collect a performance profile for debugging.",
+		ShortDescription: `
+Collects cpu, heap, and goroutine profiles from a running go-ipfs daemon
+into a single zipfile. To aid in debugging, this command also attempts to
```
aschmahmann, line 33:
```suggestion
into a single zip file. To aid in debugging, this command also attempts to
```

core/commands/profile.go
```diff
@@ -0,0 +1,234 @@
...
+	archive := zip.NewWriter(w)
+
+	// Take some profiles.
+	type profile struct {
+		name  string
+		file  string
+		debug int
+	}
+
+	profiles := []profile{{
```
aschmahmann, line 131: ditching the contention profile because of libp2p noisiness?

core/commands/profile.go
```diff
@@ -0,0 +1,234 @@
...
+		ShortDescription: `
+Collects cpu, heap, and goroutine profiles from a running go-ipfs daemon
+into a single zipfile. To aid in debugging, this command also attempts to
+include a copy of the running go-ipfs binary.
+`,
+		LongDescription: `
+Collects cpu, heap, and goroutine profiles from a running go-ipfs daemon
+into a single zipfile. To aid in debugging, this command also attempts to
+include a copy of the running go-ipfs binary.
+
```
aschmahmann, line 40: Maybe also include something like:

Profile's can be examined using `go tool pprof`, some tips can be found at https://github.com/ipfs/go-ipfs/blob/master/docs/debug-guide.md.

docs/debug-guide.md
```diff
@@ -15,8 +15,8 @@ When you see ipfs doing something (using lots of CPU, memory, or otherwise
 being weird), the first thing you want to do is gather all the relevant
 profiling information.
 
-There's a script (`bin/collect-profiles.sh`) that will do this for you and
-bundle the results up into a tarball, ready to be attached to a bug report.
+There's a command (`ipfs diag profile`) that will do this for you and
+bundle the results up into a zip file, ready to be attached to a bug report.
```
aschmahmann, line 19: I assume we're doing zips because either they're more cross platform friendly or the Go tooling is better, right?

docs/debug-guide.md
```diff
@@ -15,8 +15,8 @@ When you see ipfs doing something (using lots of CPU, memory, or otherwise
...
```
aschmahmann, line 20: While we're in this file anyway it might be worth throwing in a link to https://github.com/whyrusleeping/stackparse. There's no need to since the tool is pretty alpha and has no readme, but it's something we make use of and others might want to also.

test/sharness/t0152-profile.sh
Stebalien: I'll need to add a dependency on unzip, but fair enough.

core/commands/profile.go
Stebalien: We need it for the symbols in case the user deployed a custom binary. I guess we could include it iff the version is "dirty"?

Note: the zip is ~26MiB. So it's not terrible.

core/commands/profile.go
Stebalien: Yeah, there's no point for now. We'd also need to leave mutex profiling on for a bit to get anything useful which would make this take even longer.

docs/debug-guide.md
Stebalien: All of the above:

1. Can be uploaded to slack etc.
2. Can be _streamed_. With a tarfile, I need to know the sizes of files up-front before I can write them. With a zipfile, I can create a file, write it, create the next file, etc. without buffering anything.

core/commands/profile.go
aschmahmann: It's not, although I think I've seen maximum sizes for emails and Slack attachments from external parties be lower. I'm not sure what the maximum GitHub issue attachment size is.

Even if it's not the default we should probably have a flag here just in case

aschmahmann (approved):
Awesome!

As mentioned above I'm a little concerned about the zip file sizes containing binaries causing problems for us. Let's figure out what we want to do there (maybe it's just having a flag defaulted to including the binaries and then switching the defaults later if we're running into problems). Aside from that though LGTM.

docs/debug-guide.md
```diff
@@ -79,6 +79,9 @@ that goroutine in the middle of a short wait for something. If the wait time is
 over a few minutes, that either means that goroutine doesn't do much, or
 something is pretty wrong.
 
+If you're seeing a lot of goroutines, conside rusing
+[stackparse](https://github.com/whyrusleeping/stackparse) to filter, sort, and summarize them.
```
aschmahmann, line 83:
```suggestion
If you're seeing a lot of goroutines, consider using
[stackparse](https://github.com/whyrusleeping/stackparse) to filter, sort, and summarize them.
```

Stebalien:
> As mentioned above I'm a little concerned about the zip file sizes containing binaries causing problems for us. Let's figure out what we want to do there (maybe it's just having a flag defaulted to including the binaries and then switching the defaults later if we're running into problems). Aside from that though LGTM.

The current "collect-profiles.sh" script does the same thing. Unfortunately, without the binary, we don't get the symbols. And I couldn't find a way to extract _just_ the symbols.

Well, a quick google turns up https://stackoverflow.com/questions/42554900/how-to-extract-own-symbol-table. But that seems like it'll be unreliable, especially on other operating systems.

aschmahmann:
Fair enough. It's certainly not worse than the current script and if it's a problem we can deal with it later. Merge when ready.

> And I couldn't find a way to extract just the symbols.

You could in theory clone most of the code used by `go tool nm` https://github.com/golang/go/blob/master/src/cmd/nm/nm.go to get out the symbol table (it relies on an internal package, so not fun). I'm not totally sure how you make that information available to pprof though once you have it though.

Stebalien:
> I'm not totally sure how you make that information available to pprof though once you have it though.

Well, there's probably a way to strip out everything _except_ the symbol table, and make pprof think we're passing it the real binary.

PR #8318, opened 2021-07-30
fix: avoid out of bounds error when rendering short hashes

Stebalien:
The panic is caught higher up the stack so it can't crash the node, but it's still "not good".

### DIFF
--- a/core/corehttp/gateway_handler.go
+++ b/core/corehttp/gateway_handler.go
@@ -391,11 +391,12 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
 			size = humanize.Bytes(uint64(s))
 		}
 
-		hash := ""
-		if r, err := i.api.ResolvePath(r.Context(), ipath.Join(resolvedPath, dirit.Name())); err == nil {
-			// Path may not be resolved. Continue anyways.
-			hash = r.Cid().String()
+		resolved, err := i.api.ResolvePath(r.Context(), ipath.Join(resolvedPath, dirit.Name()))
+		if err != nil {
+			internalWebError(w, err)
+			return
 		}
+		hash := resolved.Cid().String()
 
 		// See comment above where originalUrlPath is declared.
 		di := directoryItem{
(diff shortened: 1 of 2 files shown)


aschmahmann:
I think the problem is slightly different. It looks like `hash=""` https://github.com/ipfs/go-ipfs/blob/cab67f6b6635b53bbf32680ee0da982e32bdcfee/core/corehttp/gateway_handler.go#L394 and when there's an error we just continue. What's going on there?

Stebalien:
Ah, I see. Yeah, that makes no sense.

Stebalien:
I fixed it both ways.

core/corehttp/gateway_handler.go
```diff
@@ -391,11 +391,12 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
...
-		hash := ""
-		if r, err := i.api.ResolvePath(r.Context(), ipath.Join(resolvedPath, dirit.Name())); err == nil {
-			// Path may not be resolved. Continue anyways.
-			hash = r.Cid().String()
+		resolved, err := i.api.ResolvePath(r.Context(), ipath.Join(resolvedPath, dirit.Name()))
+		if err != nil {
+			internalWebError(w, err)
+			return
 		}
+		hash := resolved.Cid().String()
```
aschmahmann, line 399: @lidel what was the idea behind "Path may not be resolved. Continue anyways.", it seems like we should just fail here as this PR does?

aschmahmann (approved):
Thanks for the quick fix, it seems correct to me. I'd like to understand if there was some logic for why we didn't do this check before to make sure we don't introduce a different error (left as a comment)

core/corehttp/gateway_handler.go
aschmahmann: Also, how do we even reach the point where we error here? It seems like the UnixFS directory is telling us "here are the objects I have" and then we look for them and it errors.

Stebalien: > Also, how do we even reach the point where we error here? It seems like the UnixFS directory is telling us "here are the objects I have" and then we look for them and it errors.

Invalid sharded directory. Or a subshard of a sharded directory that we can list but can't traverse.

mvdan (approved):
Not knowing the codebase before, SGTM. The previous code was definitely open to panics.

I guess you could just do the shortHash change to avoid panics. I don't have enough context to tell if `err != nil` should continue or stop.

Stebalien:
Ok, I'm going to merge this for now. We can change it later if there is a better fix.

PR #8324, opened 2021-08-03
ci: update to Go 1.16.6 on CI

guseggert:
Carried over from https://github.com/ipfs/go-ipfs/pull/8138

This also includes a change to use a beefier build host, which improves build times a lot (we were at 2 vCPUs, now 16 vCPUs, build time for me dropped from 30min -> 15min). It also fixes a bug that was causing the -coverprofile flag to be incorrect (due to Make subtleties, spaces were not being replaced correctly [here](https://github.com/ipfs/go-ipfs/blob/4ebec49985eed35263662b23f5e4fa1eb46b1672/coverage/Rules.mk#L23)).


.circleci/config.yml
```diff
@@ -126,6 +126,12 @@ jobs:
         TEST_VERBOSE: 1
     steps:
     - run: sudo apt update
+    - run: |
+        mkdir ~/localgo && cd ~/localgo
+        wget https://golang.org/dl/go1.16.6.linux-amd64.tar.gz
+        tar xfz go1.16.6.linux-amd64.tar.gz
+        echo "export PATH=$(pwd)/go/bin:$PATH" >> $BASH_ENV
+    - run: go version
```
aschmahmann, line 134: What's the reason why we're doing this/what happens when we don't? Is this because the machine executor doesn't have the same version of Go as the golang ones?

test/sharness/lib/test-lib.sh
```diff
@@ -9,17 +9,15 @@
 # use the ipfs tool to test against
 
 # add current directory to path, for ipfs tool.
-if test "$MAKE_SKIP_PATH" != "1"; then
```
aschmahmann, line 12: What changed with `MAKE_SKIP_PATH`?

guseggert: I'm not sure exactly why this fixes it. Presumably something somewhere else has a side efffect of setting the PATH so that it contains the correct ipfs binary for the test, which is what the sharness tests were implicitly relying on. Something changed with 1.16 such that the ipfs binary is no longer found in the PATH. I spent a few hours trying to find where, unsuccessfully. I couldn't find any rationale for MAKE_SKIP_PATH and the tests pass when I disable it, b/c PATH is setup correctly before each test run (pointing to the coverage binary).

.circleci/config.yml
guseggert: Good question, this might be unnecessary code, it was carried over from https://github.com/ipfs/go-ipfs/pull/8138.  Let me dig deeper.

guseggert: Yeah this is required because the sharness test is using an Ubuntu image which is on Go 1.15.2, not the Go image.

test/sharness/lib/test-lib.sh
aschmahmann: I'm not totally sure, but I think the idea here is that you can locally run sharness tests on your machine from the sharness test folder without causing weird issues. Could you double check those work.

It seems like removing that flag shouldn't change anything (we set it to 1 in the makefile and then check if it equals 1).

> I'm not sure exactly why this fixes it.

What was failing before and with which code changes? You've also changed a couple commands here (e.g. from ```BIN=$(cd .. && echo `pwd`/bin)``` to `BIN=$(cd .. && pwd)/bin` ). Any chance those were the changes that fixed things and not `MAKE_SKIP_PATH`?

guseggert: I did figure this out finally. It was due to this:

```
echo "export PATH=$(pwd)/go/bin:$PATH" >> $BASH_ENV
```
([link](https://github.com/ipfs/go-ipfs/blob/d3468407aaddbed885abe878f80072186c33a449/.circleci/config.yml#L133))

That writes a PATH env var to BASH_ENV, and BASH_ENV is inherited by every subshell, so every time a subshell is run, Bash applies env vars in BASH_ENV which overwrite whatever PATH was inherited from the parent process, which means setting PATH in any Makefile (or any other env in BASH_ENV) is silently ignored. 

(Setting PATH with BASH_ENV is what CircleCI docs recommend....)

aschmahmann (approved):
LGTM. @guseggert if there's a 1.16.7 then let's use that, otherwise we can just merge.

BigLep:
Per 2021-08-17 conversation: update to 1.16.7

Short term
* everything except go-ipfs should run 1.16.x and 1.17.0
* sharness will run on 1.16.x 
* infrastructure will deploy with 1.16.x

.circleci/config.yml
```diff
@@ -126,6 +127,12 @@ jobs:
         TEST_VERBOSE: 1
     steps:
     - run: sudo apt update
+    - run: |
+        mkdir ~/localgo && cd ~/localgo
+        wget https://golang.org/dl/go1.16.6.linux-amd64.tar.gz
```
Stebalien, line 132: Should be 1.16.7

PR #8325, opened 2021-08-04
Cosmetic fixups in examples

petar:
* fix example go.sum
* use common example folder
* spell out complete source of second example (go ipfs with experimental features)


docs/examples/go-ipfs-with-experimental/go.mod
```diff
@@ -0,0 +1,15 @@
+module github.com/ipfs/go-ipfs/examples/go-ipfs-with-experimental
+
+go 1.14
+
+require (
+	github.com/ipfs/go-ipfs v0.7.0
```
aschmahmann, line 6: we should bump to v0.9.1 in the examples

docs/examples/go-ipfs-with-experimental/go.mod
```diff
@@ -0,0 +1,15 @@
+module github.com/ipfs/go-ipfs/examples/go-ipfs-with-experimental
+
+go 1.14
```
aschmahmann, line 3: We should bump to the same version as in go-ipfs (1.15)

docs/examples/go-ipfs-with-experimental/main.go
```diff
@@ -0,0 +1,354 @@
+package main
```
aschmahmann, line 1: IIRC this is mostly the same as the other example. Should we just use a `--experimental` flag in `main.go` to disambiguate?

Is it useful for these to be separate binaries?

petar:
@aschmahmann Addressed comments here.

petar:
@aschmahmann Added test. Worked out well.

aschmahmann:
The test seems to pass, but it doesn't seem to be running in CI. Once it's running then it looks like we're good to go.

petar:
> The test seems to pass, but it doesn't seem to be running in CI. Once it's running then it looks like we're good to go.

It seems to have passed.

Stebalien:
Can we make the tests run when we run `make test`?

Issue #387, opened 2014-11-25
Untangling Configuration Var Mess


jbenet:
We have three ways to express configuration variables to `ipfs`:
- `$IPFS_DIR/config` file
- ENV vars
- cli options / args

This is both good and bad. Good because all of these methods have specific use cases that make them much more preferrable to others (e.g. env vars in deployments, cli options are faster to change while experimenting than a config file, etc). Bad because it can create lots of confusion as to where each of these options should be defined in the code, and which methods it should support.

Here are some thoughts. Not sure what the right trajectory is yet.
1. Config: below the "ipfs core" barrier (i.e. all code within an ipfs node), we should use the `Config` exclusively (we have no access to options, AND should never use `os.GetEnv`). This means that all config variables the node can choose from have to be set on the config. (i think so far we've adhered to this). 
2. Options: in commands, and commands exclusively, we parse options and call functions as need be. (this is already the case because only commands have access to options).
3. ENV vars for normal execution: we define _all_ env vars for normal execution in one place `cmd/env.go`, read from the env, and set them on map on the `cmdInvocation`, or the execution context. Then, we could either force Config overrides to happen (i.e. env vars can only override config values) or Commands may access the env variables from the execution context. BUT, note `os.GetEnv` is disallowed (forcing users of env vars to add them to the one place we define them all, so we have them all listed in one file).
4. ENV vars for tests: for simplicity, one small exception for test case env vars: we are free to use, for example,  `os.GetEnv("NO_FUSE")` from tests.

rht:
ENV vars should be removed.

They are not tied to a specific daemon, hence not loggable (also who knows who/what set them and if they affect child processes). ENV vars for initial config are fine, but if further enabled they could be abused for persistent config in a bashrc. Also, extra overhead because people have to remember config precedence.

Besides cli option already provides the same convenience.

Reminds me of a mess in geant4 (a particular simulator framework) installation before they switched to cmake nowadays (https://aur.archlinux.org/packages/ge/geant4/PKGBUILD, https://github.com/Homebrew/homebrew-science/blob/master/geant.rb).

jbenet:
ENV vars are used very often for all sorts of things. unfortunately it's the easiest way to setup deployments in many kinds of deployment systems (things like heroku and so on). I dont think we should remove them.

[...]

PR #8326, opened 2021-08-04
Docker override config from env variables

vk496:
Hello,

It seems a old problem without a clear patch to support it. I've tried to do it as much simple/clean as possible.

With this patch, anyone can override the config with env variable such as:

 - IPFSCONFIG_Swarm_EnableAutoRelay=true
 - IPFSCONFIG_Addresses_Swarm=[\\"/ip4/0.0.0.0/tcp/4001\\",\\"/ip6/::/tcp/4001\\"]

Take in account that:
 - Is case sensitive. Is not the same `IPFSCONFIG_FooBar` than `IPFSCONFIG_FOOBAR`
 - A valid syntax will be accepted even if the path is wrong (`ipfs config` stuff). Something like `IPFSCONFIG_Swarm_EnableAutoRealy=true` will be accepted and written to the config even if it does nothing. So typos could make you loose a lot of hours here :)

Hope it helps you. IPFS is really awesome stuff and now it should be more flexible in container environments

/cc #251 #387 #6325

PS: Docker docs should be updated if is finally merged

### DIFF
--- a/bin/container_daemon
+++ b/bin/container_daemon
@@ -2,6 +2,7 @@
 set -e
 user=ipfs
 repo="$IPFS_PATH"
+ipfsEnvPrefix="IPFSCONFIG"
 
 if [ `id -u` -eq 0 ]; then
   echo "Changing user to $user"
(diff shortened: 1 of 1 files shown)


aschmahmann:
Is this necessary given that you can just do `ipfs daemon --init --init-config  <template>`?

vk496:
> Is this necessary given that you can just do `ipfs daemon --init --init-config <template>`?

Most people expect use containers as something fungible. Since you can't apply new config without restarting the daemon you must use volumes.

I think is very very useful be able start a fresh ipfs container with all desired config with just one cammand and ready to use

lidel:
This PR was discussed again during GO triage on 2021-09-03, my quick summary: 

- This will work only for a subset of config options
  - There is a hidden complexity in regard to go-ipfs config file syntax around dynamic maps (DNS, pinning service credentials) which makes us worried about the need for domain-specific syntax here – see https://github.com/ipfs/go-ipfs/pull/8096.  This is not a blocker, but highlights the problem of too many edge cases
- Might create confusion
  - This PR adds env variables that work only with Docker, and not standalone go-ipfs. 
  - Usually, passing env variables acts as an ad-hoc-one-time override, while this PR makes a permanent change to the config. Restarting container without variable does not restore previous value, which is another papercut which we would like to avoid.
  - If we ever add env variable based config override, it should be supported in go-ipfs natively ([`go-ipfs/docs/environment-variables.md`](https://github.com/ipfs/go-ipfs/blob/master/docs/environment-variables.md)), and work as "ad-hoc" override that does not mutate the config file.
  
In short, the decision was to not do this and close this PR without merging, and go with  explicit docker hooks proposed in  https://github.com/ipfs/go-ipfs/pull/8337

Thank you for  the PoC and pushing the discussion forward @vk496 :+1:

PR #8147, opened 2021-05-19
feature: 'ipfs swarm peering' command

TakashiMatsuda:
## Overview

This PR is intended to implement a part of https://github.com/ipfs/go-ipfs/issues/7880.

## Detail

New command `ipfs swarm peering` is added in this commit.
There are three subcommands under `ipfs swarm peering`. :

```
ipfs swarm peering add <address>... - add peers into the peering service.
ipfs swarm peering ls               - list peers registered in the peering service.
ipfs swarm peering rm <ID>...       - remove a peer from the peering service.
```

<details>
  <summary>Usage of 'ipfs swarm peering add' </summary>

  Example:
  ```
$ ./cmd/ipfs/ipfs swarm peering add /ip4/1.2.3.4/tcp/1234/p2p/QmYyQSo1c1Ym7orWxLYvCrM2EmxFTANf8wXmmE7DWjhx5N
add QmYyQSo1c1Ym7orWxLYvCrM2EmxFTANf8wXmmE7DWjhx5N success
  ```

  Help text:
  ```
  USAGE
  ipfs swarm peering add <address>... - add peers into the peering service.

  ipfs swarm peering add [--] <address>...

  'ipfs swarm peering add' adds peers into the peering service.

  For more information about each command, use:
  'ipfs swarm peering add <subcmd> --help'
  ```
</details>

<details>
  <summary>Usage of 'ipfs swarm peering rm' </summary>

  Example:
```
$ ./cmd/ipfs/ipfs swarm peering rm QmYyQSo1c1Ym7orWxLYvCrM2EmxFTANf8wXmmE7DWjhx5N
remove QmYyQSo1c1Ym7orWxLYvCrM2EmxFTANf8wXmmE7DWjhx5N success
```

  Help text:
```
> ipfs swarm peering rm IDs...
USAGE
  ipfs swarm peering rm <ID>... - remove a peer from the peering subsystem.

  ipfs swarm peering rm [--] <ID>...

  'ipfs swarm peering rm' remove peers from the peering subsystem.

  For more information about each command, use:
  'ipfs swarm peering rm <subcmd> --help'
```
</details>

<details>
  <summary>Usage of 'ipfs swarm peering ls' </summary>

Example:
```
$ ./cmd/ipfs/ipfs swarm peering ls
12D3KooWPBxKtYk9ptRH2B9tEJvBiQnPxKL3JxaVvJgmh3bpeGmn
        /ip4/35.213.90.5/tcp/4001
```

Help text:
```
USAGE
[...]

PR #8331, opened 2021-08-07
Fix: Use a pointer type on IpfsNode.Peering

TakashiMatsuda:
# The problem

The current type of `IpfsNode.Peering` is `PeeringService`.
It is different from the type `*PeeringService` returned from `node.Peering` provided in core/node/groups.go:270 .
This difference prevents fx from injecting `IpfsNode.Peering`.
The current `node.PeeringService` on the daemon is a zero value.

I also cannot get the valid `PeeringService` on `SwarmAddCommand` in  https://github.com/ipfs/go-ipfs/pull/8147 due to this problem.

# Proposal

I propose to fix the type of `IpfsNode.PeeringService` from `PeeringService` to `*PeeringService`.

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -85,7 +85,7 @@ type IpfsNode struct {
 
 	// Online
 	PeerHost      p2phost.Host            `optional:"true"` // the network host (server+client)
-	Peering       peering.PeeringService  `optional:"true"`
+	Peering       *peering.PeeringService `optional:"true"`
 	Filters       *ma.Filters             `optional:"true"`
 	Bootstrapper  io.Closer               `optional:"true"` // the periodic bootstrapper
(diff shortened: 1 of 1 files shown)


Stebalien (approved):
Ah, lol. So, peering _works_ right now because we're passing [PeerWith](https://github.com/ipfs/go-ipfs/blob/6e610066f6a9f1212410fbab88b625444e031fe5/core/node/peering.go#L28). This causes us to create a peering service even though the peering service on the main struct is empty.

I was really confused how our sharness tests were passing.

Stebalien:
Thanks!

TakashiMatsuda:
Yes. 

At the moment the setting of the peering service on daemon cannot be changed after startup,  so I guess the sharness test did not reveal the problem.

Stebalien:
Yep. Thanks for the fix!

PR #8332, opened 2021-08-07
feat: register first block metric by default

guseggert:
This is an important performance metric that PL uses for gateway monitoring, so enabling it by default.

This is cherry-picked from 44fcc1558a65de37cbc74fd383cedfc6f29979fd


aschmahmann (approved):
LGTM, although I'm not totally sure why we left this out previously. IIRC it's because we were hoping to put together a better metric here, not because of performance issues or anything like that.

@Stebalien does that sounds correct?

Stebalien:
I can't remember either. But I'd just ship it.

PR #8337, opened 2021-08-12
Add support for scripts during initialization

vk496:
This functionality was suggested by https://github.com/ipfs/go-ipfs/issues/387#issuecomment-517340629
I saw @Caian PR https://github.com/ipfs/go-ipfs/pull/6577 after, so I publish my suggestion anyway

This patch will enable more complex init patterns during the IPFS first initialization. The scripts under */docker-prehook.d* directory will be executed just after all env initialization and **BEFORE** running ipfs daemon. If there were no errors, it will execute the ipfs daemon. **AFTER** running ipfs daemon (10 seconds) it will execute all scripts under */docker-posthook.d* and kill the main ipfs process if any script failed.

Btw, I'm not sure if the documentation of this functionality should go to the README.md or to https://github.com/ipfs/ipfs-docs/

### DIFF
--- a/bin/container_daemon
+++ b/bin/container_daemon
@@ -1,7 +1,21 @@
-#!/bin/sh
+#!/bin/bash
 set -e
 user=ipfs
 repo="$IPFS_PATH"
+
+init_scripts() {
+  trap 'kill $$' ERR # Kill parent if init fail
+  local scripts_path=$1
+
+  if [ -d ${scripts_path} ]; then
+    for script in ${scripts_path}/*.sh; do
+      echo "==== Running $script"
+      # Prevent failure if no +x is set on file
+      cat $script | sh -e
+    done
+  fi
+
+}
 
 if [ `id -u` -eq 0 ]; then
   echo "Changing user to $user"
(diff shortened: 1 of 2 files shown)


bin/container_daemon
```diff
@@ -51,6 +65,12 @@ else
   # Unset the swarm key file variable
   unset IPFS_SWARM_KEY_FILE
 
+  # Pre-hook init with scripts
+  init_scripts /docker-prehook.d
+
+  # Post-hook init scripts
+  { sleep 10; init_scripts /docker-posthook.d ; } &
```
lidel, line 72: Separate stage after sleeping for 10 seconds feels brittle. Could not be enough if my node runs on slow CI and takes longer time to init/start etc. 

I feel we don't need a separate directory with posthooks executed after a hardcoded delay.  If anything, we would be waiting for node to start (eg. by probing `/api/v0/id` on the  API port etc).

lidel (commented):
Thank you @vk496! 

If we remove posthook due to `sleep 10` (see comment inline), this PR feels redundant.
Let's continue in  @Caian's PR #6577 instead.

PR #8349, opened 2021-08-17
refactor: core/node/bitswap.go

lidel:
This PR is against  https://github.com/ipfs/go-ipfs/pull/8268  for easier review:
-  extracts bitswap init to separate file and defines implicit defaults as consts
- fixes a typo where BitswapTaskWorkerCount was 128 (instead of 8 ) and BitswapEngineBlockstoreWorkerCount was 8 (instead of 128)

### DIFF
--- a/core/node/bitswap.go
+++ b/core/node/bitswap.go
@@ -1,0 +1,52 @@
+package node
+
+import (
+	"context"
+
+	"github.com/ipfs/go-bitswap"
+	"github.com/ipfs/go-bitswap/network"
+	blockstore "github.com/ipfs/go-ipfs-blockstore"
+	config "github.com/ipfs/go-ipfs-config"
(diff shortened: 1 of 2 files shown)


core/node/bitswap.go
```diff
@@ -0,0 +1,52 @@
...
+
+	"github.com/ipfs/go-ipfs/core/node/helpers"
+)
+
+const (
+	// Docs: https://github.com/ipfs/go-ipfs/blob/master/docs/config.md#internalbitswap
+	EngineBlockstoreWorkerCount = 128
+	TaskWorkerCount             = 8
+	EngineTaskWorkerCount       = 8
+	MaxOutstandingBytesPerPeer  = 1 << 20
```
aschmahmann, line 23: Let's either lowercase or prefix with the word Default

PR #8352, opened 2021-08-18
go-ipfs-as-a-library example: go mod tidy

hsanjuan:
Go complained about something not being in the go.mod file with go1.16


Stebalien:
Yes. Merged in #8397.

PR #8354, opened 2021-08-18
update go-libp2p to v0.15.0-rc.1

marten-seemann:


go.mod
```diff
@@ -65,52 +65,53 @@ require (
...
+	github.com/multiformats/go-multiaddr v0.4.0
 	github.com/multiformats/go-multiaddr-dns v0.3.1
 	github.com/multiformats/go-multibase v0.0.3
 	github.com/multiformats/go-multicodec v0.3.0
 	github.com/multiformats/go-multihash v0.0.15
 	github.com/opentracing/opentracing-go v1.2.0
 	github.com/pkg/errors v0.9.1
-	github.com/prometheus/client_golang v1.10.0
+	github.com/prometheus/client_golang v1.11.0
+	github.com/prometheus/statsd_exporter v0.21.0 // indirect
```
marten-seemann, line 103: I don't really understand why, but I had to manually run `go get github.com/prometheus/statsd_exporter@v0.21.0`, since I was getting this error:
```
❯ go mod tidy
go: finding module for package github.com/prometheus/common/log
github.com/ipfs/go-ipfs/core/corehttp imports
        contrib.go.opencensus.io/exporter/prometheus imports
        github.com/prometheus/statsd_exporter/pkg/mapper imports
        github.com/prometheus/common/log: module github.com/prometheus/common@latest found (v0.30.0), but does not contain package github.com/prometheus/common/log

**Comment by aschmahmann** [comment]
We probably need to bump the version of Go on the go mod file to 1.16 also, right? We can wait on CI to do it's initial run though

**Comment by marten-seemann** [comment]
> We probably need to bump the version of Go on the go mod file to 1.16 also, right?

Done.

File: core/node/libp2p/addrs.go [hunk 2]
```diff
@@ -4,24 +4,31 @@ import (
...
+			}
+			if err := gater.BlockSubnet(f); err != nil {
+				return nil, Libp2pOpts{}, fmt.Errorf("failed to block subnet %s: %w", s, err)
 			}
-			opts.Opts = append(opts.Opts, libp2p.FilterAddresses(f)) //nolint
 		}
-		return filter, opts, nil
+		var opts Libp2pOpts
+		opts.Opts = append(opts.Opts, libp2p.ConnectionGater(gater))
+		return ma.NewFilters(), opts, nil
```
aschmahmann, line 31: This is failing because the returned `ma.Filters` has no impact on the ConnectionGater at all. This means all the commands relying on `n.Filters` to interact with the connection gater are failing.

e.g. https://github.com/ipfs/go-ipfs/blob/7db27bf83a10cd9804a4be8e7421b5ed52bda461/core/commands/swarm.go#L571

Here was the previously used connection gater that exposed that information via the passed in `ma.Filters` https://github.com/libp2p/go-libp2p/blob/v0.14.4/options_filter.go

aschmahmann (approved):

PR #8370, opened 2021-08-24
fix: allow the levelds compression level to be unspecified

Stebalien:
Otherwise, the user will get a confusing panic. And given that we support "", we might as well support nil.


Stebalien:
IMO, not worth testing because this will bite _very_ few users. But it's so easy to fix, I felt like just doing it.

Issue #8067, opened 2021-04-10
--config expects directory not file


dokterbob:
#### Version information:

```
go-ipfs version: 0.8.0-ce693d7
Repo version: 11
System version: amd64/linux
Golang version: go1.14.4
```

#### Description:

Override configuration file by using: `ipfs --config=/tmp/config config show` 
Result:
`Error: error loading plugins: open /tmp/config/config: not a directory`
Expected results: dump of configuration JSON.

aschmahmann:
We should add a new flag `--repo` and deprecate this one since the flag is about where the repo is located not just the config file.

dokterbob:
The specific use case for the —config option would be to validate the configuration before applying it - very useful in deployment automation and/or production scenario’s.

h1z1:
It's certainly ambiguous.. Having spent hours at one point trying to figure out what was broken, I'm not sure which is more incorrect, the help or the error.   As per --help that should be a file:

```  -c, --config               string - Path to the configuration file to use.
```

Just got in a habit of setting IPFS_PATH *shrug*

```
IPFS_PATH=${PWD} ipfs config show 
```

A config file can and probably should reside outside the repo itself.

PrashantBtkl:
Hi, Im thinking of picking this issue.
As far as my understanding the config is always loaded from the repo with the default name as "config"
I am trying a fix where we can check the config with the --config flag correctly but not setting the config this way 
because of this specification by @dokterbob 

> The specific use case for the —config option would be to validate the configuration before applying it 

waiting for comments on this

gammazero:
@PrashantBtkl There are two parts to this fix.  The first is what @aschmahmann said [above](https://github.com/ipfs/go-ipfs/issues/8067#issuecomment-817907845).  The current `--config` flag should be `--repo` since it specifies the repo directory path.  The second part is to implement a new `--config` flag that specifies the path to the config file only.  If not specified, then this defaults to `<repo_path>/config`.

Implementing this will require coordinating changes in both [go-ipfs-config](https://github.com/ipfs/go-ipfs-config) and [go-ipfs](https://github.com/ipfs/go-ipfs) repos.

PrashantBtkl:
Hi @gammazero , @BigLep 
This would be my approach 
make `config` seperate from `FsRepo` Struct
make `config.Filename` function to return the filename directly instead of concatinating the reporoot + filename
Instead of giving r.path to the config.Filename send either
[...]

PR #8372, opened 2021-08-24
--config option would take file instead of repo directory

PrashantBtkl:
This PR addresses issue :  #8067

--config flag takes repo directory instead of the config file path
This addresses the config file path as a separate path(from the repo path) to check the config for the following command:-
`ipfs -config=path/to/ipfs.conf config show`
or whatever the name of the config file is

Thanks

### DIFF
--- a/cmd/ipfs/main.go
+++ b/cmd/ipfs/main.go
@@ -10,6 +10,7 @@ import (
 	"net/http"
 	"os"
 	"runtime/pprof"
+	"strings"
 	"time"
 
 	util "github.com/ipfs/go-ipfs/cmd/ipfs/util"
@@ -45,8 +46,8 @@ const (
 	heapProfile        = "ipfs.memprof"
 )
 
-func loadPlugins(repoPath string) (*loader.PluginLoader, error) {
-	plugins, err := loader.NewPluginLoader(repoPath)
+func loadPlugins(repoPath string, configPath string) (*loader.PluginLoader, error) {
+	plugins, err := loader.NewPluginLoader(repoPath, configPath)
 	if err != nil {
 		return nil, fmt.Errorf("error loading plugins: %s", err)
 	}
@@ -123,13 +124,13 @@ func mainRet() int {
 
 	buildEnv := func(ctx context.Context, req *cmds.Request) (cmds.Environment, error) {
 		checkDebug(req)
-		repoPath, err := getRepoPath(req)
+		repoPath, configPath, err := getRepoAndConfigPath(req)
 		if err != nil {
 			return nil, err
 		}
 		log.Debugf("config path is %s", repoPath)
 
-		plugins, err := loadPlugins(repoPath)
+		plugins, err := loadPlugins(repoPath, configPath)
 		if err != nil {
 			return nil, err
 		}
@@ -137,7 +138,7 @@ func mainRet() int {
 		// this sets up the function that will initialize the node
 		// this is so that we can construct the node lazily.
 		return &oldcmds.Context{
-			ConfigRoot: repoPath,
+			ConfigRoot: configPath,
 			LoadConfig: loadConfig,
 			ReqLog:     &oldcmds.ReqLog{},
 			Plugins:    plugins,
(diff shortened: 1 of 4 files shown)


PrashantBtkl:
hi @BigLep , I think the CI failed because I made changes to [go-ipfs-config](https://github.com/ipfs/go-ipfs-config) locally while I made these changes.
I have some questions though
After making this PR I was given more Inputs on the two changes that has to be made
1.) change `--config` flag to `--repo`
2.) create new  `--config` flag, load config separately from the repo 

as @gammazero  mentioned on issue  #8067
>  The second part is to implement a new --config flag that specifies the path to the config file only. If not specified, then this defaults to <repo_path>/config.

I am not sure how to approach the second part as I have to make changes to [go-ipfs-config](https://github.com/ipfs/go-ipfs-config) repo

gammazero:
@PrashantBtkl The goal here is (or should be) to allow the config file to be in a separate location from the rest of the repo the config's location is explicitly specified.  Currently, there is no way to do this since go-ipfs only allows the repo directory to be specified using the somewhat misnamed `--config` flag.  The processing of this flag or no flag, into repo and config file paths is handled in go-ipfs-config.

What we want to do is to have two separate flags: `--repo` and `--config`.  The `--repo` flag will specify the path to the repo directory.  If not specified, this defaults to `~/.ipfs`.  The `--config` flag will specify the location of the ipfs config file.  If not specified it will default to the `<repo_path>/config`, where `<repo_path>` is the repo directory set by `--repo` or its default.  All of this needs to be done in `go-ipfs-config`.

The part of the work that is needed in `go-ipfs` is to use the two values, `repo` and `config`, from `go-ipfs-config` instead of only the single config value (repo directory) that it is using now.

Hopefully, that clarifies what I was looking for.  Let me know if that is not clear, or not what you want to accomplish with this PR.  If you get to a point where you want assistance completing the necessary PRs, I can help.

cmd/ipfs/main.go
```diff
@@ -293,17 +294,22 @@ func makeExecutor(req *cmds.Request, env interface{}) (cmds.Executor, error) {
...
+	}
+	configPath, found := req.Options["config"].(string)
+	if found && configPath != "" {
+		split := strings.Split(configPath,"/")
+		config.DefaultConfigFile = strings.Join(split[len(split)-1:]," ") //considering the last split as the config file name
+		configPath = ""
+		for i := 0; i<len(split)-1; i++ {
+		    configPath = configPath+split[i]+"/"
+		}
+		return repoPath,configPath, nil
```
gammazero, line 310: Seems like using `path.Dir` and `path.Base` would simplify the code here:
```go
config.DefaultConfigFile = path.Base(configPath)
configPath = path.Dir(configPath)
if configPath == "." {
    configPath = ""
}
return repoPath, configPath, nil
```

However, this logic should be in `go-ipfs-config` so that applications other than go-ipfs can access the config file and the repo using the same logic.  I think this can be accomplished by modifying `FilePath` to take a separate `configFilePath` argument, and if it is not set then it does what it currently does.  If a `configFilePath` is given, then it uses that path to the config file. If only a file name is given the the result is `Path(configroot, configFilePath)`.
https://github.com/ipfs/go-ipfs-config/blob/master/config.go#L80-L82

gammazero (changes requested):
Some logic needs to move into `go-ipfs-config` so that applications other then IPFS can access repo and config file using same logic.

BigLep:
@PrashantBtkl : are you able to incorporate the provided feedback?

PrashantBtkl:
Hi, for this issue there would be a lot of change that needs to be incorporated in my understanding
1) removing the config from FsRepo struct and somehow managing config in [go-ipfs-config](https://github.com/ipfs/go-ipfs-config)
https://github.com/ipfs/go-ipfs/blob/c54cdaa1f87507ba1048c8d450dc43e6db6eb1d7/repo/fsrepo/fsrepo.go#L93-L105

2.) moving some of config functions from fsrepo interface to [go-ipfs-config](https://github.com/ipfs/go-ipfs-config)

- https://github.com/ipfs/go-ipfs/blob/c54cdaa1f87507ba1048c8d450dc43e6db6eb1d7/repo/fsrepo/fsrepo.go#L389
- https://github.com/ipfs/go-ipfs/blob/c54cdaa1f87507ba1048c8d450dc43e6db6eb1d7/repo/fsrepo/fsrepo.go#L210
- https://github.com/ipfs/go-ipfs/blob/c54cdaa1f87507ba1048c8d450dc43e6db6eb1d7/repo/fsrepo/fsrepo.go#L225
- https://github.com/ipfs/go-ipfs/blob/c54cdaa1f87507ba1048c8d450dc43e6db6eb1d7/repo/fsrepo/fsrepo.go#L236
- https://github.com/ipfs/go-ipfs/blob/c54cdaa1f87507ba1048c8d450dc43e6db6eb1d7/repo/fsrepo/fsrepo.go#L499
- https://github.com/ipfs/go-ipfs/blob/c54cdaa1f87507ba1048c8d450dc43e6db6eb1d7/repo/fsrepo/fsrepo.go#L518

and all of the remaining config functions to go-ipfs-config

I think someone who has already contributed to go-ipfs and go-ipfs-config would be a good pick to solve this issue :)

PR #8374, opened 2021-08-24
ci: preload peerlog plugin, disable by default

guseggert:
This preloads the peerlog plugin in the ipfs binary, but keeps it
disabled by default. To enabled it, set Enabled=true in its config.

The motivation is to simplify building and deploying gateways, and for
them to use binaries that are more similar to release bins.


plugin/plugins/peerlog/peerlog.go
```diff
@@ -66,9 +67,30 @@ func (*peerLogPlugin) Version() string {
...
+func extractEnabled(config interface{}) bool {
+	// plugin is disabled by default, unless Enabled=true
+	if config == nil {
+		return false
+	}
+	mapIface, ok := config.(map[string]interface{})
+	if !ok {
+		return false
+	}
+	enabledIface, ok := mapIface["Enabled"]
```
aschmahmann, line 79: Something about this setup is a little weird:

```
adin@AdinPC:~/go/src/github.com/ipfs/go-ipfs/cmd/ipfs$ ./ipfs config Plugins
{
  "Plugins": null
}
adin@AdinPC:~/go/src/github.com/ipfs/go-ipfs/cmd/ipfs$ ./ipfs config Plugins.Plugins.peerlog.Config.Enabled true
adin@AdinPC:~/go/src/github.com/ipfs/go-ipfs/cmd/ipfs$ ./ipfs config Plugins
{
  "Plugins": {
    "peerlog": {
      "Config": {
        "Enabled": "true"
      },
      "Disabled": false
    }
  }
}
```

While this obviously works, this config file is pretty gross. Maybe this is just a matter of using a better name than Enabled to explain what's going on here.

AFAICT what's happening is that plugins default to `Disabled : false` because the default struct value for a boolean is false. This means that loading the config will by default turn on the plugin, but then the plugin specific variable will turn it off by default.

Stebalien: This is too much of a hack, IMO. Can we use a flag here? Then make plugins expose a `DefaultEnabled` function, or something like that?

guseggert: @Stebalien what exactly do you mean by "flag", and are you proposing adding `DefaultEnabled() bool` to the `plugin.Plugin` interface or adding a new optional interface?

Stebalien: > what exactly do you mean by "flag",

https://github.com/ipfs/go-ipfs-config/blob/master/types.go#L50

> and are you proposing adding DefaultEnabled() bool to the plugin.Plugin interface or adding a new optional interface?

The former. It'll be breaking, but that's not the end of the world.

You should then be able to use `myPluginConfig.Disabled.WithDefault(!myPlugin.DefaultEnabled())`.

---

I _really_ regret calling it `Disabled` now. I wonder how much breakage we'd cause if we renamed it to `Enabled`?

BigLep: Hi guys - I'm only commenting to calibrate here (not because I have a strong judgment on what should be done here).  Correct me if this is wrong, but my understanding here is that we're blocking this PR because we don't like the precedent it sets and we instead want to set a good pattern for future/public plugins before we merge it in.

I like the mindset of doing things right and setting good patterns.  I am wondering though if we're letting "best be the enemy of better".  I think this change is internal and that only our infra will have this gross/ugly config.  It does allow us to simplify our build/deploy logic elsewhere as we'd be using the same binary on our infra that we distribute elsewhere.

The way I'd imagine proceeding is:
1. Adding comments/disclaimers on this in the internal plugin code "NOT TO BE REPEATED.  WHEN WE HAVE A PUBLIC PLUGIN, WE NEED TO REWORK X.  PLEASE SEE Y FOR MORE INFO."
2. We fix this up when we do include a plugin (could happen soon with dag-jose depending on conversations this week).

It's not perfect but it allows us to simplify in another area we're actively touching (build/deploy logic) and keep this self-contained.  By pushing the bar higher right now in this area, we won't be able to move the "simplify build/deploy logic" ball forward right now as I assume we don't have the time in the middle of the release to do a larger refactor.

I realize I may be off in my understanding and feel free to disagree.  I'm only dipping in to better understand.

Stebalien: My concern is that if we don't do this now, we'll have to live with it for a _very_ long time because it'll never be a priority. On the other hand, you're right. In _this_ case, it's an internal hack and we probably can just live with it forever.

BigLep: Ok.  My commitment would be that we do this right as part of shipping any public plugins (e.g., dag-jose), but I leave it up to @Stebalien or @aschmahmann on whether they want to approve or not.  Thanks for engaging.

Stebalien: Sounds fine by me.

aschmahmann: Same here. Let's get an issue up for how we want to do this along with @Stebalien's proposal for how to do so.

aschmahmann (approved):
LGTM per the conversation in https://github.com/ipfs/go-ipfs/pull/8374/files#r695115872.

@guseggert could you start a GitHub issue on the default disabled plugin problem, and add a comment to the code explaining the weirdness here as @BigLep suggested?

Once that's done we'll hit the merge button

PR #8380, opened 2021-08-25
Debugging fixtures related to gateway incident of Aug 23

petar:
* expose unlimited-size stack trace on http api under /debug/stack


guseggert:
We need to merge this as part of some action items from recent operational issues.

PR #8385, opened 2021-08-26
ci: remove "extras" docker build & push

guseggert:
This is no longer necessary as the peerlog plugin is bundled in
go-ipfs now.


lidel (changes requested):
@guseggert seems this got out of sync with `master` at some point, mind rebasing? 
(I would do it myself, but lacking context about recent changes)

lidel:
Superseded by #8467

PR #8389, opened 2021-08-27
Fixtures for issue 8381

petar:


Mitumaru (approved):

gammazero:
There is currently no need for this. If needed in the future, maybe a solution based on `bufio` could be implemented, as this my cause OOM creating too large of a buffer to write stack data to.

PR #8401, opened 2021-09-02
fix: fix the example's mod file

Stebalien:
It fails on go 1.17.


Stebalien:
Hm. There's already #8352

Issue #7421, opened 2020-06-05
ipfs object patch does not work with HAMT-sharded directories


lidel:
#### Version information:

```
go-ipfs version: 0.6.0-rc1-aa169520c
Repo version: 9
System version: amd64/linux
Golang version: go1.14.3
```

## Problem 1: `object patch` does not work with sharded dirs

- `ipfs object patch add-link` is not HAMT-aware
- patched directory was `hamt-sharded-directory`
- produced directory is also marked as `hamt-sharded-directory`
- the file added to HAMT-sharded dir does not follow HAMT-sharding
- the CLI output truncates filename and makes things even more confusing

### Details / Repro

Given this specific HAMT directory (note `stript.js` being already present):

```console
$ ipfs ls bafybeibvfzwz5ghenwdzeqrlx7sgzbo3zcdaomscc42ina4mv3ritby7qa
bafkreifpfoilyfbc4lbvkzupje2e7xgti2t5zcpg6ryekzvdyqpmzmovpa 2808 index.html
bafybeiavgkssvw57furrh3tzld2bo6ryjbpki4qpyyvishomwtj3gkxut4 -    ipfs-css/
bafkreigtdgsgv2f3bkhsmxvku3bpnnqzubcxeupf7fff5f7l7tlm2v237a 3232 ipfs-logo.svg
bafkreidgfrdvq7lcifj23cxkr2sxumpakzyhai7qkbgiszktqgr5hnktlm 3193 script.js
bafybeibuo42uthah7qfhcqw4loxz2g2sefay7y3sjrbrfiqtueq3dd5s6y -    tachyons/
```

Really confusing thing happens when I attempt to ` ipfs object patch add-link` above directory to add a new CID (`bafkreidfofyuw4ywqwjjdgxyhjictt7f2iufxto2v5n4rnm4atp57ajjne`) using already existing name `script.js`, and then inspect result via CLI:

```console
$ ipfs object patch add-link bafybeibvfzwz5ghenwdzeqrlx7sgzbo3zcdaomscc42ina4mv3ritby7qa script.js bafkreidfofyuw4ywqwjjdgxyhjictt7f2iufxto2v5n4rnm4atp57ajjne
bafybeigq3ctbwpa6ydvhszc5qumd6pp2ockyp7lmpyynvqywwgo4uu5ffy

$ ipfs ls bafybeigq3ctbwpa6ydvhszc5qumd6pp2ockyp7lmpyynvqywwgo4uu5ffy
[...]

PR #8414, opened 2021-09-07
fix(cli): object add-link: do not allow blocks over BS limit

schomatis:
Trying not to be too invasive as this is just an ad-hoc check for a deprected command I have added the check at the CLI level; otherwise I can go deeper in the `ObjectAPI`.

If the placement of the check is correct will add a test as a sharness. If instead the check should go deeper in the stack then the test should likely be written in Go. Waiting on an initial review to decide course.

Optional additional proposals to the current PR:
* Add the same check for the (less common) `set-data` and `append-data` commands which can potentially go over the BS limit just as well.
* Add the check to `rm-link` _but only as a warning_ to flag if the current block being manipulated is already over the limit.

Fixes https://github.com/ipfs/go-ipfs/issues/7421.

### DIFF
--- a/core/commands/object/patch.go
+++ b/core/commands/object/patch.go
@@ -92,6 +92,10 @@ DEPRECATED and provided for legacy reasons. Use 'ipfs add' or 'ipfs files' inste
 			return err
 		}
 
+		if err := checkBlockSize(req, p.Cid(), api.Dag()); err != nil {
+			return err
+		}
+
 		return cmds.EmitOnce(res, &Object{Hash: p.Cid().String()})
 	},
 	Type: &Object{},
@@ -138,6 +142,10 @@ DEPRECATED and provided for legacy reasons. Use 'files cp' and 'dag put' instead
 			return err
 		}
 
+		if err := checkBlockSize(req, p.Cid(), api.Dag()); err != nil {
+			return err
+		}
+
 		return cmds.EmitOnce(res, &Object{Hash: p.Cid().String()})
 	},
 	Type: Object{},
@@ -173,6 +181,10 @@ DEPRECATED and provided for legacy reasons. Use 'files rm' instead.
 		name := req.Arguments[1]
 		p, err := api.Object().RmLink(req.Context, root, name)
 		if err != nil {
+			return err
+		}
+
+		if err := checkBlockSize(req, p.Cid(), api.Dag()); err != nil {
 			return err
 		}
 
--- a/core/commands/object/patch.go
+++ b/core/commands/object/patch.go
@@ -6,16 +6,16 @@ import (
 
 	"github.com/ipfs/go-cid"
 	cmds "github.com/ipfs/go-ipfs-cmds"
-	coreiface "github.com/ipfs/interface-go-ipfs-core"
 	"github.com/ipfs/go-ipfs/core/commands/cmdenv"
+	coreiface "github.com/ipfs/interface-go-ipfs-core"
 
 	"github.com/ipfs/interface-go-ipfs-core/options"
 	"github.com/ipfs/interface-go-ipfs-core/path"
 )
 
 const (
-	// FIXME: Confirm if this is the right name.
-	forceBlockSize   = "force-block-size"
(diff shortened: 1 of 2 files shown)


lidel (changes requested):
Tested with this one-liner that patches  until error – works as expected :+1: 

```console
$ echo QmUNLLsPACCz1vLxQVkXqqLX5R1X345qqfHbsf67hvA3Nn>cid; while ipfs object patch $(<cid) add-link $(<cid).jpg Qmayz4F4UzqcAMitTzU4zCSckDofvxstDuj3y7ajsLLEVs 2>&1 > cid; do; <cid; done
QmcPUbyiR2g9RbTcLQAMHq5S9Fh3gBoXJqso2yK8awEMDC
...
QmTw19KYw2c8w5756s95ccKpFoFhbP6oXa4Zg5cLbLD79N
Error: object API does not support HAMT-sharding. To create big directories, please use the files API (MFS)
```

Remaining work:
- add the same check to `rm-link` and  `set-data` and `append-data`
  - better to have same behavior everywhere and fail har, and add override flag (below)
- add `--force` flag to disable the block size check  on impacted commands
  - we need to have an escape hatch to unblock people who need the old behavior, otherwise some collabs won't be able to upgrade. 
  - rationale: we want to apply soft pressure, but if someone wants to delay deprecation of `ipfs object` APIs in their codebase, they should have the ability to do so without the need for forking go-ipfs.
- add sharness tests (fine to reuse my one-liner)

core/commands/object/patch.go
```diff
@@ -232,6 +232,21 @@ Use MFS and 'files' commands instead:
...
+		// unenforced and undeclared rule of thumb hard-coded here.
+		modifiedNode, err := api.Dag().Get(req.Context, p.Cid())
+		if err != nil {
+			return err
+		}
+		modifiedNodeSize, err := modifiedNode.Size()
+		if err != nil {
+			return err
+		}
+		if modifiedNodeSize > 1024 * 1024 {
```
petar, line 246: My guess would be that returning an error wouldn't suffice. The node is already too big and this has to be reverted. But reverting it here is hard, because the reversal can itself cause another error. So I would think this must be rejected at the ObjectAPI level.

BigLep:
Per 2021-09-24 verbal, this is a bigger problem and we're going to take care of the immediate problem and not put a check in deeper like the blockstore.  

We believe the check is happening at the right level but we should incorporate Petar's feedback to avoid mutating the datastore for the error case.

Check to see if can add a method to the object API that checks for the size and rejects if it's too big before it mutates anything.

@petar can provide additional comments.

core/commands/object/patch.go
schomatis: @petar I see two places where we could potentially put this check deeper in the stack:

*  `(api *ObjectAPI) AddLink`: after the [`InsertNodeAtPath`](https://github.com/ipfs/go-ipfs/blob/51ca2192472286c5c96aa64e02f6325b1b5ddafa/core/coreapi/object.go#L216-L219) call (which has the actual 'insertion' logic) using a check similar to the one placed at the CLI level (fetch the modified node and check its size). We might also want to evaluate purging the modified copy from the repo.

* (deeper) in the `go-merkledag` repo executing the actual [insertion logic](https://github.com/ipfs/go-merkledag/blob/8794146d52c211b8af0456e734a248649e636be0/dagutils/utils.go#L122-L138): it has the advantage that already has a permanent/temporary store division where rejecting the modified node wouldn't impact the `go-ipfs` repo. The downside is that we would be injecting an arbitrary restriction on a library based on the limitation of another (bitswap) just because both are consumed by the same `go-ipfs` logic.

Either is doable so I'll leave to you the decision of where to execute.

lidel (approved):
LGTM. Let's keep it small and merge it as-is (protect people from creating non-transferrable blocks asap). 

Made small changes:
- renamed override flag to --allow-big-block and made error more explicit about what and why
- separate tests for default and override behavior

@schomatis is this still a draft, or ready for merge?

Issue #7898, opened 2021-02-04
Optional ipfs daemon --agent-version-suffix=


lidel:
I propose we add support for running `ipfs daemon` with optional `--agent-version-suffix=` parameter.

This would enable vendors that bundle go-ipfs binary to customize `AgentVersion` (listed in `ipfs id` and announced to other peers) without the need for creating a fake libp2p protocol just to tell one cohort of `go-ipfs` nodes  from another.

The first users of this would be Brave browser and IPFS Desktop:
- `--agent-version-suffix=brave/1.19.85`  →  `go-ipfs/0.9.0/brave/1.19.85`
- `--agent-version-suffix=desktop/0.14.0`  →   `go-ipfs/0.9.0/desktop/0.14.0`

## Rationale

This would be opt-in, so does not introduce any new tracking vector, it simply aims to make things easier for vendors and to introduce common convention that preserves version number of go-ipfs.

In the current situation, where vendor can rebuild go-ipfs binary and change agentVersion, but then we lose information about % of different go-ipfs nodes on the DHT.  By making it a suffix, and making it work without rebuild, we will decrease the number of opaque AgentVersions, making it easier for us to track upgrade dynamics etc. 

### CLI param, ENV or Config?

No preference  atm – all will work for Brave.

- Config option would work fine
- Env variable `IPFS_AGENT_VERSION_SUFFIX`  (similar to `IPFS_PATH`) is good enough too
- Explicit  flag for `ipfs daemon` would be preferred, because it has added value if user  has multiple go-ipfs nodes running on same localhost (eg. ipfs-desktop and brave). It enables things like  `ps aux | grep brave` to quickly identify specific daemon instance.

cc @andrew @bbondy

lidel:
Added to maintenance board, we want to land this by EOY so vendors like IPFS Cluster, Brave or Berty could add their own version suffix for improved metrics.

schomatis:
@lidel Going with the preferred CLI param option as suggested. Let me know if you'd like something different.

schomatis:
@lidel The agent version comes from the user agent in

https://github.com/ipfs/go-ipfs/blob/65d570c6cbe24f313fc6f02c3dc16aabc53306d1/version.go#L14

Do we care to _also_ modify this as well with the new option or just the agent version propagated over the wire? I'm going with the second option leaving user agent as is but let me know otherwise.

schomatis:
[...]

PR #8419, opened 2021-09-08
feat(cli): add daemon option --agent-version-suffix

schomatis:
Preliminary design for feedback to add the option storing it inside the `IpfsNode` structure.

(Test and minor fixes will follow after this design is validated.)

Closes https://github.com/ipfs/go-ipfs/issues/7898.

### DIFF
--- a/cmd/ipfs/daemon.go
+++ b/cmd/ipfs/daemon.go
@@ -67,6 +67,7 @@ const (
 	enablePubSubKwd           = "enable-pubsub-experiment"
 	enableIPNSPubSubKwd       = "enable-namesys-pubsub"
 	enableMultiplexKwd        = "enable-mplex-experiment"
+	agentVersionSuffix        = "agent-version-suffix"
 	// apiAddrKwd    = "address-api"
 	// swarmAddrKwd  = "address-swarm"
 )
@@ -180,6 +181,7 @@ Headers.
 		cmds.BoolOption(enablePubSubKwd, "Instantiate the ipfs daemon with the experimental pubsub feature enabled."),
 		cmds.BoolOption(enableIPNSPubSubKwd, "Enable IPNS record distribution through pubsub; enables pubsub."),
 		cmds.BoolOption(enableMultiplexKwd, "DEPRECATED"),
+		cmds.StringOption(agentVersionSuffix, "Optional suffix to the AgentVersion presented by `ipfs id` and also advertised through BitSwap."),
 
 		// TODO: add way to override addresses. tricky part: updating the config if also --init.
 		// cmds.StringOption(apiAddrKwd, "Address for the daemon rpc API (overrides config)"),
@@ -408,6 +410,11 @@ func daemonFunc(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment
 		ncfg.Routing = libp2p.NilRouterOption
 	default:
 		return fmt.Errorf("unrecognized routing option: %s", routingOption)
+	}
+
+	agentVersionSuffixString, _ := req.Options[agentVersionSuffix].(string)
+	if agentVersionSuffixString != "" {
+		version.SetUserAgentSuffix(agentVersionSuffixString)
 	}
 
 	node, err := core.NewNode(req.Context, ncfg)
--- a/version.go
+++ b/version.go
@@ -8,7 +8,22 @@ const CurrentVersionNumber = "0.11.0-dev"
 
 const ApiVersion = "/go-ipfs/" + CurrentVersionNumber + "/"
 
-// UserAgent is the libp2p user agent used by go-ipfs.
+// GetUserAgentVersion is the libp2p user agent used by go-ipfs.
 //
 // Note: This will end in `/` when no commit is available. This is expected.
-var UserAgent = "go-ipfs/" + CurrentVersionNumber + "/" + CurrentCommit
+func GetUserAgentVersion() string {
+	userAgent := "go-ipfs/" + CurrentVersionNumber + "/" + CurrentCommit
+	if userAgentSuffix != "" {
+		if CurrentCommit != "" {
+			userAgent += "/"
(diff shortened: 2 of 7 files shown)


cmd/ipfs/daemon.go
```diff
@@ -67,6 +67,7 @@ const (
 	enablePubSubKwd           = "enable-pubsub-experiment"
 	enableIPNSPubSubKwd       = "enable-namesys-pubsub"
 	enableMultiplexKwd        = "enable-mplex-experiment"
+	agentVersionSuffix        = "agent-version-suffix="
```
lidel, line 70:
```suggestion
	agentVersionSuffix        = "agent-version-suffix"
```

cmd/ipfs/daemon.go
```diff
@@ -180,6 +181,7 @@ Headers.
 		cmds.BoolOption(enablePubSubKwd, "Instantiate the ipfs daemon with the experimental pubsub feature enabled."),
 		cmds.BoolOption(enableIPNSPubSubKwd, "Enable IPNS record distribution through pubsub; enables pubsub."),
 		cmds.BoolOption(enableMultiplexKwd, "DEPRECATED"),
+		cmds.StringOption(agentVersionSuffix, "Suffix to the AgentVersion presented by `ipfs id` and also advertised through BitSwap."),
```
lidel, line 184:
```suggestion
		cmds.StringOption(agentVersionSuffix, "Optional suffix to the AgentVersion presented by `ipfs id` and also advertised through BitSwap."),
```

version.go
```diff
@@ -11,4 +11,16 @@ const ApiVersion = "/go-ipfs/" + CurrentVersionNumber + "/"
...
+	userAgent := "go-ipfs/" + CurrentVersionNumber + "/"
+	if suffix != "" {
+		userAgent += suffix + "/"
+		// FIXME: An alternative to passing the suffix as string would be to have a
+		//  static `var UserAgentSuffix string` here and modify it during `ipfs daemon`
+		//  startup. The drawback is that ideally this variable should only be allowed
+		//  to be modified there so we instead make it part of the node structure and
+		// explicitly pass it here. Check if this is the correct alternative.
+	}
+	userAgent += CurrentCommit
```
lidel, line 24: nit: please move the optional CurrentCommit before suffix, otherwise the suffix is no longer the suffix :)

core/core.go
```diff
@@ -61,6 +61,9 @@ type IpfsNode struct {
 	// Self
 	Identity peer.ID // the local node's identity
 
+	// FIXME: Check if this is the right place for it.
+	UserAgentSuffix string
```
lidel, line 65: Perhaps we could move this to `version.go` and have `SetUserAgentSuffix` for updating the suffix  at the runtime when the CLI flag is present and `GetUserAgentVersion()` for reading the entire thing, instead of passing the suffix each time?
This way we would have everything in one place, hidden away.

lidel (commented):
I feel we want to keep as much as possible inside `version.go` – see comment inline.

version.go
schomatis: Thanks for the correction. I misunderstood the `1.19.85` of your example for the `go-ipfs` commit but I see now that it was part of the passed option.

core/core.go
schomatis: Take a look at this [comment](https://github.com/ipfs/go-ipfs/pull/8419/files#diff-1ef170619a70876f007d5edfc4554a81aa686eae7678b70df0347b3133cd6d14R18-R22) please and let me know if it's aligned with the request here, just for me to make sure I understand it correctly.

Basically the suggestion is having a _single, static_ suffix for the entire program and allowing modifications on the fly (that would ideally be contained in the `ipfs daemon` startup through the setter you're suggesting).

lidel: Yes, the alternative solution described in the comment sounds preferable :+1: 
(it will make things easier if we ever decide that we want to allow an alternative way of defining the suffix by defining it in the config file)

schomatis: Sounds good, thanks for the confirmation.

lidel (approved):
LGTM, thank you!

schomatis:
@lidel Sharness passing now (didn't really do anything besides squashing and triggering another round of testing; this was already passing locally on my machine).

schomatis:
(The tests [themselves](https://app.circleci.com/pipelines/github/ipfs/go-ipfs/5351/workflows/1520177c-f5c8-4191-af9b-1707431287e4/jobs/58447/parallel-runs/0/steps/0-111) are passing, the error seems unrelated to this PR and more internal with CI dynamics but let me know if you'd like another run or any modification on this patch.)

gammazero:
@schomatis Rebase this PR to pick up CI changes that will likely allow your sharness tests to pass.

schomatis:
Rebased. Still hitting same error.

lidel (changes requested):
@schomatis  seems that `diff -u expected-agent-version actual-agent-version` fails on CI because of duplicate slash:
```diff
--- expected-agent-version	2021-09-13 12:30:02.643061595 +0000
+++ actual-agent-version	2021-09-13 12:30:02.703065735 +0000
@@ -1 +1 @@
-go-ipfs/0.11.0-dev//test-suffix
+go-ipfs/0.11.0-dev/test-suffix
```

I think we should clean this up to have more deterministic strings, I propose normalization that ensures agent version string acts more like a proper path:
(1) no duplicate slashes, and.. 
(2) no trailing slash

schomatis:
@lidel Thanks for pasting the diff here; I couldn't extract it from the Circle UI. Fixed the sharness test to cover both cases of the commit being present or not.

petar (approved):
This seems reasonable to me. LGTM.

lidel (approved):
LGTM. 

Merging to unlock use in Brave when  0.10 ships.

Issue #8421, opened 2021-09-09
Provide a way to destroy pinset


gammazero:
### Checklist

- [X] My issue is specific & actionable.
- [X] I am not suggesting a protocol enhancement.
- [X] I have searched on the [issue tracker](https://github.com/ipfs/go-ipfs/issues?q=is%3Aissue) for my issue.

### Description

Sometimes it is necessary to remove the entire pin set:
- If there is ever a need to rebuild the node's pin set.
- When inconvenient to remove many individual pins.  It may be much easier to remove all pins and re-pin specific objects.

To support this, the local pinning API should expose a `RemoveAll` method.  There should be an `ipfs pin removeall` command that deletes all pins.

BigLep:
Per 2021-11-12 conversation, we're not planning to tackle this until there is more of an impetus for this, especially since someone can modify the datastore directly.

PR #8423, opened 2021-09-09
Provide command to delete all pins

gammazero:
This PR provides the command, `ipfs pin removeall`, that removes all pins on the local node.

This addresses issue #8421

Depends on the following PRs:
- https://github.com/ipfs/interface-go-ipfs-core/pull/78
- https://github.com/ipfs/go-ipfs-pinner/pull/17

NOTE: This PR should be updated to the tagged version of the dependencies above after their PRs are merged.


BigLep:
Per 2021-11-12 discussion, we're not going to create a first class command for this.

PR #8424, opened 2021-09-10
ci: use dynamic config for CircleCI

guseggert:
When developing against ipfs/go-ipfs, we would like to be able to use
the 2xlarge resource class for faster build and test cycles, but many
external contributers will not have this resource class available to
them. There is no direct way to change the resource class, so this
uses dynamic config to generate a parameters JSON obj which is then
fed into the configuration when the workflow starts, based on the Git
URL of the build.

For repos other than ipfs/go-ipfs, this reverts back to the "medium"
resource class with a Make job parallelism of 3.


gammazero (approved):
This looks good.  I think it will unblock a number of PRs.

PR #8428, opened 2021-09-10
Add 'ipfs repo migrate' command

### LINKED ISSUE #7471: ipfs migration without starting the daemon

gammazero:
This PR replaces #7658 that was originally contributed by @zaibon, in order to move code into a branch and avoid some CI problems.

The command allows the user to run the repo migration without starting the daemon.

resolves #7471

### DIFF
--- a/core/commands/commands_test.go
+++ b/core/commands/commands_test.go
@@ -217,6 +217,7 @@ func TestCommands(t *testing.T) {
 		"/repo",
 		"/repo/fsck",
 		"/repo/gc",
+		"/repo/migrate",
 		"/repo/stat",
 		"/repo/verify",
 		"/repo/version",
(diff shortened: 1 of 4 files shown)


Mitumaru (approved):

guseggert:
I rebased w/ master and fixed a bug where this wasn't propagating non-ErrNeedMigration errors (e.g. if you try to migrate while the daemon is running, you should get a lock error, not "no migration is needed")

guseggert (approved):

PR #8432, opened 2021-09-13
fix(sharness): add extra check in flush=false in files write

schomatis:


petar (approved):
Nothing controversial. Fail and printout if the daemon is not running, but expected to run. LGTM.

ifstatic:
i think i finally got it working properly.. thanks for the help everyone.. loving ipfs

PR #8436, opened 2021-09-15
chore: update go-path

aschmahmann:
https://github.com/ipfs/go-path/pull/44

### DIFF
--- a/core/coreapi/test/path_test.go
+++ b/core/coreapi/test/path_test.go
@@ -1,0 +1,79 @@
+package test
+
+import (
+	"context"
+	"strconv"
+	"testing"
+	"time"
+
+	files "github.com/ipfs/go-ipfs-files"
+	"github.com/ipfs/go-merkledag"
+	uio "github.com/ipfs/go-unixfs/io"
+	"github.com/ipfs/interface-go-ipfs-core/options"
+	"github.com/ipfs/interface-go-ipfs-core/path"
+	"github.com/ipld/go-ipld-prime"
+)
+
+
+func TestPathUnixFSHAMTPartial(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	// Create a node
+	apis, err := NodeProvider{}.MakeAPISwarm(ctx, true, 1)
+	if err != nil {
+		t.Fatal(err)
+	}
+	a := apis[0]
+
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -51,7 +51,7 @@ require (
 	github.com/ipfs/go-metrics-prometheus v0.0.2
 	github.com/ipfs/go-mfs v0.1.2
 	github.com/ipfs/go-namesys v0.3.1
-	github.com/ipfs/go-path v0.1.1
+	github.com/ipfs/go-path v0.1.2-0.20210914131249-d0512cea3d64
```
aschmahmann, line 54: TODO: Update to released go-path when this is ready to go

core/coreapi/test/path_test.go
```diff
@@ -0,0 +1,79 @@
...
+
+	// Remove one of the sharded directory blocks
+	if err := a.Block().Rm(ctx, path.IpfsPath(pbNode.Links()[0].Cid)); err != nil {
+		t.Fatal(err)
+	}
+
+	// Try and resolve each of the entries in the sharded directory which will result in pathing over the missing block
+	//
+	// Note: we could just check a particular path here, but it would require either greater use of the HAMT internals
+	// or some hard coded values in the test both of which would be a pain to follow.
```
aschmahmann, line 66: This makes the test take a couple of seconds to run, which is a bit annoying if we were going to run many of these although not so bad by itself.

core/coreapi/test/path_test.go
```diff
@@ -0,0 +1,79 @@
...
+	for k := range dir {
+		// The node will go out to the (non-existent) network looking for the missing block. Make sure we're erroring
+		// because we exceeded the timeout on our query
+		timeoutCtx, timeoutCancel := context.WithTimeout(ctx, time.Second * 1)
+		_, err := a.ResolveNode(timeoutCtx, path.Join(r, k))
+		if err != nil {
+			if timeoutCtx.Err() == nil {
+				t.Fatal(err)
+			}
```
aschmahmann, line 75: This is one of the reasons I chose to do this instead of a sharness test (the other being the sharness tests are less portable which is overall just no fun).

The equivalent sharness test would have to process text outputs like "context deadline exceeded" as good and "context cancelled" as bad which is a little too close for comfort and wouldn't detect other regressions where we returned "context deadline exceeded" in error.

PR #8438, opened 2021-09-15
fix: take the lock while listing peers

Stebalien:


Stebalien:
Reported in https://github.com/ipfs/go-ipfs/pull/8147#issuecomment-920198287.

aschmahmann (approved):

Issue #8415, opened 2021-09-08
0.10 dag API changes breaking HTTP API (discussion for resolution)


rvagg:
Surfaced via @tabcat's work @ https://github.com/orbitdb/orbit-db/issues/909, picked up in failing tests @ https://github.com/ipfs/js-ipfs/blob/870d446f/packages/ipfs-http-client/test/dag.spec.js when running against the `go-ipfs@next` package which is the 0.10.0-rc1 binary.

In summary - the changes we made are fine, there's no error, but we may have gone harder than we should have, or intended to and there are ways we could bridge the breakage for users. @tabcat says they're moving to the block API exclusively, which is fine and probably suits the workflow there, but the dag API is where we've put all the new goodness and users avoiding it suggests that maybe we could do better.

1. We've lost `cid-version` for `dag put`, or at least it doesn't work.

Looking through history I'm wondering if we ever had a `cid-version` for the dag API? It looks like it's only valid for the `add` and `files` API. Does this mean that https://github.com/ipfs/js-ipfs/blob/870d446f/packages/ipfs-http-client/test/dag.spec.js#L33 is incorrect and is only working because it's defaulting to v0 for a dag-pb block?

Maybe the resolution here is to be clear in the CHANGELOG that the CIDs for the dag API are always v1? I don't see a note for that anywhere. Alternatively we could just introduce it there and allow it to be used for the case where `format=dag-pb` so people can get their beloved CIDv0's.

2. The `git-raw` codec is included by default now. 

This test https://github.com/ipfs/js-ipfs/blob/870d446f/packages/ipfs-http-client/test/dag.spec.js#L77 assumes it's not available, but it gets an answer. Are we intending to enable the plugin by default now?

3. The changes in meaning of `input-enc` and `format` might make more impact than we intend.

[...]

Issue #4801, opened 2018-03-10
Ipld DAG diff


Stebalien:
Use-case: replace `ipfs object diff` with a generalized version that works for generalized IPLD DAGs.

Text syntax:

* Use `+` for added items.
* Use `-` for removed items.
* Use `~` for replaced items.
* Syntax: OP PATH VALUE

JSON syntax: Use JSON patch.
Generalized syntax: Basically, JSON patch with extra CBOR types.

Blocker: Pathing through DagPB nodes in go-ipfs isn't correct per the IPLD spec (paths don't reflect the object structure). See https://github.com/ipld/specs/issues/55.

lidel:
Pinning services would really like to have this ability to compare DAGs behind two CIDs and get  a list of block CIDs for additions and removals, as that would enable them to optimize [the replace pin](https://ipfs.github.io/pinning-services-api-spec/#tag/pins/paths/~1pins~1{requestid}/post) operation in their internal infra. 

Sounds like something we may want to look into after ipld-prime refactor is done.
@warpfork pathing through dag-pb  may be solved by representing it as ADL, right? (I remember we touched upon that during HAMT discussions)

lidel:
I know dag-pb is not compatible with ADLs yet (https://github.com/ipld/go-ipld-prime/issues/258), 
but perhaps we could special-case/shim it for now and make `ipfs dag diff` work for our users?

schomatis:
Working on some special case variant that will at least cover the basic `object diff` use case.

lidel:
Some additional thoughts on `ipfs dag diff`

## On operating at the right abstraction level

The output of `ipfs object diff` was focused on and limited to files and directories (unixfs)

The `ipfs dag diff` (https://github.com/ipfs/go-ipfs/issues/4801) will be a new command where in the future we want to add support for `dag-json`, `dag-cbor`, or any other IPLD codec that follows [IPLD Data Model](https://ipld.io/glossary/#data-model).


My take is that `ipfs dag diff` operates at a fairly low level. When diffing unixfs dags the output has to be very explicit and unambiguous – people should compare two DAGs and see key differences at DAG link level, without being shadowed by higher abstractions, like  HAMT-sharding (note: seeing internal HAMT structure is a feature, not a bug here). 

This means the `dag diff` command should effectively normalize both inputs to [IPLD Data Model](https://ipld.io/glossary/#data-model) and the diff should be the delta between the two.

## On output format

[...]

Issue #4782, opened 2018-03-07
'ipfs dag patch'


whyrusleeping:
I recently chatted with a user who is using `ipfs object patch` to maintain a KV storage tree in ipfs. We should look into adding a patch command to `ipfs dag` so that users of `object patch` can migrate.

Note that this isnt quite as simple to implement as `ipfs object patch` is, but should still be doable under certain constraints.

Stebalien:
Personally, I'd use (something like) [JSON Patch](http://jsonpatch.com/). We could even use IPLD objects as patches.

```sh
> ipfs dag patch apply QmObj QmPatch1 QmPatch2...
> ipfs dag patch apply QmId < patch.json
> ipfs dag patch add QmObj /path "value"
...
```

However, we'll need a general purpose dag *encoder* system to get this working properly (we currently only have a general purpose dag *decoder* system).

lidel:
Relevant IPLD spec: https://ipld.io/specs/patch/

PR #8439, opened 2021-09-15
change names of ipfs dag put flags to make changes clearer

aschmahmann:
Proposed changes for #8415

### DIFF
--- a/core/commands/dag/dag.go
+++ b/core/commands/dag/dag.go
@@ -84,8 +84,8 @@ into an object of the specified format.
 		cmds.FileArg("object data", true, true, "The object to put").EnableStdin(),
 	},
 	Options: []cmds.Option{
-		cmds.StringOption("format", "f", "Format that the object will be added as.").WithDefault("dag-cbor"),
-		cmds.StringOption("input-enc", "Format that the input object will be.").WithDefault("dag-json"),
+		cmds.StringOption("store-codec", "s", "Codec that the stored object will be encoded with").WithDefault("dag-cbor"),
+		cmds.StringOption("input-codec", "i", "Codec that the input object is encoded in").WithDefault("dag-json"),
 		cmds.BoolOption("pin", "Pin this object when adding."),
 		cmds.StringOption("hash", "Hash function to use").WithDefault("sha2-256"),
 	},
--- a/core/commands/dag/dag.go
+++ b/core/commands/dag/dag.go
@@ -84,8 +84,8 @@ into an object of the specified format.
 		cmds.FileArg("object data", true, true, "The object to put").EnableStdin(),
 	},
 	Options: []cmds.Option{
-		cmds.StringOption("store-codec", "s", "Codec that the stored object will be encoded with").WithDefault("dag-cbor"),
-		cmds.StringOption("input-codec", "i", "Codec that the input object is encoded in").WithDefault("dag-json"),
+		cmds.StringOption("store-codec", "Codec that the stored object will be encoded with").WithDefault("dag-cbor"),
+		cmds.StringOption("input-codec", "Codec that the input object is encoded in").WithDefault("dag-json"),
 		cmds.BoolOption("pin", "Pin this object when adding."),
 		cmds.StringOption("hash", "Hash function to use").WithDefault("sha2-256"),
 	},
--- a/core/commands/dag/put.go
+++ b/core/commands/dag/put.go
@@ -31,17 +31,17 @@ func dagPut(req *cmds.Request, res cmds.ResponseEmitter, env cmds.Environment) e
 		return err
 	}
 
-	ienc, _ := req.Options["input-enc"].(string)
-	format, _ := req.Options["format"].(string)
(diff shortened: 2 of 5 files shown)


lidel (commented):
Looks sensible, I would shorten `store-codec` to `codec` to indicate it is THE codec, 
but not feeling too strong either way.

cc @warpfork if any concerns (context: we need to change names to ensure people are not hit by silent bugs caused byt the change around cbor/dag-cbor and json/dag-json, which may produces different CID in some cases)

aschmahmann:
Don't feel too strongly about store-codec vs codec here, figured more text might be more informative but we could also throw that in the help text.

I tried using codec in https://github.com/ipfs/go-ipfs/pull/8440#discussion_r709599236. However, we can't use `-c` as a shortcut because it's already taken by `--config` on the global `ipfs` command. Any thoughts on the name in #8440 and here, as well as if there should be a shortcut name?

warpfork:
I think I'm gonna studiously _not_ have an opinion this (at least, today).

If I _were_ going to have an opinion, I'd start forming it by sketching a new command _family_, and make sure I can draft the whole family at once, and make sure things are internally consistent and follow some rule.  I'd draft some examples (just to get my brain working on the concretes), then document the _rule_ (this is the important part), and only _then_ finish implementing and do final consistency review on the commands.  (And to put my time/money where my mouth is at: I started drafting some new CLI tools in another repo, where I'm going to do that, in that order.)

If we're going to do something about the `ipfs dag` commands, a similar approach could be useful.  I'd be careful of patching one `ipfs dag *` subcommand at a time.  If we want to minimize surprise of new end-users, it's the whole subcommand family that should have some holistic rule that it follows.  (If we want to minimize the surprise of existing end-users of this API: I have no idea, I'm not one of them.)

And I don't know where I sit on the topic of whether it's worth changing these at all, vs just drafting generationally new APIs.  My 2c is that it's often a lot easier/higher-leverage to draft generationally new things (lets you fix _families_ of problems; also lets you maintain the old one with less change until you sunset it).  But in the `go-ipfs` repo, that's not my choice to make.

core/commands/dag/dag.go
```diff
@@ -77,8 +77,8 @@ into an object of the specified format.
 		cmds.FileArg("object data", true, true, "The object to put").EnableStdin(),
 	},
 	Options: []cmds.Option{
-		cmds.StringOption("format", "f", "Format that the object will be added as.").WithDefault("dag-cbor"),
-		cmds.StringOption("input-enc", "Format that the input object will be.").WithDefault("dag-json"),
+		cmds.StringOption("store-codec", "s", "Codec that the stored object will be encoded with").WithDefault("dag-cbor"),
+		cmds.StringOption("input-codec", "i", "Codec that the input object is encoded in").WithDefault("dag-json"),
```
lidel, line 81: Does this sound natural? I want to make it very clear, this is how data is encoded "at rest".
```suggestion
		cmds.StringOption("codec", "Codec that the stored object will be encoded with at rest in the datastore").WithDefault("dag-cbor"),
		cmds.StringOption("input-codec", "Codec that the input object is encoded in").WithDefault("dag-json"),
```

lidel (changes requested):
Regarding the blast radius of the decision made here, AFAIK  impacted commands are only `dag put` and `dag get` (and maybe future  `patch` (https://github.com/ipfs/go-ipfs/issues/4782) and `diff` (https://github.com/ipfs/go-ipfs/issues/4801) operations). 

I played with different notations, and  IMO the value of short notation is debatable: most of the time defaults are fine, and when one needs to override them, longer notation is less prone to typos and errors are easier to catch.

:point_right: My suggestion is to go with clean `--input-codec` and `--codec` (without any shortcuts).

ribasushi:
> I played with different notations, and IMO the value of short notation is debatable: most of the time defaults are fine, and when one needs to override them, longer notation is less prone to typos and errors are easier to catch.
> 👉 My suggestion is to go with clean --input-codec and --codec (without any shortcuts).

I concur with dropping the shortcuts/single-letter-options.

I am less certain about going with `--input-codec / --codec`. While yes, other systems use `--codec` throughout, they do not do "translation" level stuff. I think having a symmetric `--input-codec / --store-codec` ( and maybe even `--output-codec` in the future ), while clunky, is the right way to go...

warpfork:
I guess if anything is to change here, I probably agree with @ribasushi 's suggestion that things be explicit about input direction and output direction, and we pretty much tacitly never use the word "codec" bare without a directional indicator.  One of the big issues with these commands and their flags as they stand now are that people get very confused which phase a codec is applied at, so if we're changing anything, it seems we should increasingly favor explicitness.

Maybe some other commands only have one direction or the other, sure.  But I bet it will still be clearer overall to have those commands have a longer flag just to be consistent with anything that _does_ have flags for multiple phases of data flow.

aschmahmann:
Latest update: We're going to use `--input-codec` and `--store-codec` here, and `ipfs dag get` will use `--output-codec`.

lidel (approved):

lidel (approved):

PR #8440, opened 2021-09-15
change ipfs dag get flag name from format to output-codec

aschmahmann:
Change `ipfs dag get --format` to `ipfs dag get --output-codec` to be closer to https://github.com/ipfs/go-ipfs/pull/8439.

TODO: Merge to master after #8439

### DIFF
--- a/core/commands/dag/dag.go
+++ b/core/commands/dag/dag.go
@@ -116,7 +116,7 @@ format.
 		cmds.StringArg("ref", true, false, "The object to get").EnableStdin(),
 	},
 	Options: []cmds.Option{
-		cmds.StringOption("format", "f", "Format that the object will be serialized as.").WithDefault("dag-json"),
+		cmds.StringOption("output-codec", "Format that the object will be encoded as.").WithDefault("dag-json"),
 	},
 	Run: dagGet,
 }
--- a/core/commands/dag/get.go
+++ b/core/commands/dag/get.go
@@ -22,9 +22,9 @@ func dagGet(req *cmds.Request, res cmds.ResponseEmitter, env cmds.Environment) e
 		return err
 	}
 
-	format, _ := req.Options["format"].(string)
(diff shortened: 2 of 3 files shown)


core/commands/dag/dag.go
```diff
@@ -109,7 +109,7 @@ format.
 		cmds.StringArg("ref", true, false, "The object to get").EnableStdin(),
 	},
 	Options: []cmds.Option{
-		cmds.StringOption("format", "f", "Format that the object will be serialized as.").WithDefault("dag-json"),
+		cmds.StringOption("codec", "c", "Format that the object will be encoded as.").WithDefault("dag-json"),
```
aschmahmann, line 112: "c" is used for the global `--config`. We can either not use a shortcut or come up with a different name/leave things as is.

lidel: I don't think short notation adds value here, I'm for just removing it (see https://github.com/ipfs/go-ipfs/pull/8439#discussion_r715076778)

aschmahmann: Going to remove the shortcut and change to `--output-codec` to clarify that the stored data is being transcoded per https://github.com/ipfs/go-ipfs/pull/8439#issuecomment-928151167.

core/commands/dag/get.go
```diff
@@ -22,9 +22,9 @@ func dagGet(req *cmds.Request, res cmds.ResponseEmitter, env cmds.Environment) e
 		return err
 	}
 
-	format, _ := req.Options["format"].(string)
-	var fCodec mc.Code
-	if err := fCodec.Set(format); err != nil {
+	codecStr, _ := req.Options["output-codec"].(string)
```
aschmahmann, line 25: This lack of error handling seems wrong, although this pattern is all over the commands lib usage so likely not too big a problem

guseggert (approved):

PR #8442, opened 2021-09-16
Make variadic false for `ipfs files rm`

AyumuKasuga:
Because it's actually handles only one argument, and I am not sure that
it should actually support multiple files.
If we want to make it support multiple arguments in the future -- we need to handle
error for each file from a list separately somehow, as unix `rm` does.

#8395

### DIFF
--- a/core/commands/files.go
+++ b/core/commands/files.go
@@ -1033,7 +1033,7 @@ Remove files or directories.
 	},
 
 	Arguments: []cmds.Argument{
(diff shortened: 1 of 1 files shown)


Stebalien:
This should definitely support multiple. Mind changing it to loop over the Arguments instead of just picking off the first one?

AyumuKasuga:
Okay, will see what I can do, closing this one for now.

PR #8447, opened 2021-09-20
feat(core): protect MFS root in node with lock

### LINKED ISSUE #6113: GC and MFS are not safe
We don't hold any form of GC lock while adding files to MFS.

schomatis:
Depends on https://github.com/ipfs/go-ipfs/pull/8574. _Please review and land that first to simplify this diff._

Replace the readily accessible MFS root in the node (`FilesRoot`) with a lock-protected version (`LockedFilesRoot`).

Testing: We currently only do a basic test manually (through the temporarily added script `test-mfs-lock.bash`), but we should look for a robust way to test the lock (preferably not through sharness/bash but through the internal Go API).

Note that this entire logic applies only when having a daemon running where the different GC and files commands contend for the same MFS root. If the commands are run in standalone mode then each one will contend for the entire repo's lock (and this mechanism isn't relevant there).

TODO:
- [x] Update this description according to the new changes.
- [x] Land https://github.com/ipfs/go-ipfs/pull/8574 on which this depends.
- [ ] Figure out how to reliably test this (that is, how to test that the race condition doesn't happen anymore). I haven't seen any sharness test that could be used as models.
- [ ] Resolve the blocking `FIXME`s.
- [ ] Review the unlocks and be more fine-grained where possible instead of using `defer` always.

Closes https://github.com/ipfs/go-ipfs/issues/6113.

### DIFF
--- a/cmd/ipfs/pinmfs.go
+++ b/cmd/ipfs/pinmfs.go
@@ -50,7 +50,11 @@ type ipfsPinMFSNode struct {
 }
 
 func (x *ipfsPinMFSNode) RootNode() (ipld.Node, error) {
-	return x.node.FilesRoot.GetDirectory().GetNode()
+	// FIXME(BLOCKING): We're deploying this lock taking mechanism as simply as
+	//  possible for now deferring the unlock but we should be more precise.
+	filesRoot := x.node.LockedFilesRoot.RLock()
+	defer x.node.LockedFilesRoot.RUnlock()
+	return filesRoot.GetDirectory().GetNode()
 }
 
 func (x *ipfsPinMFSNode) Identity() peer.ID {
--- a/core/commands/files.go
+++ b/core/commands/files.go
@@ -9,10 +9,12 @@ import (
 	gopath "path"
 	"sort"
 	"strings"
+	"time"
 
 	humanize "github.com/dustin/go-humanize"
 	"github.com/ipfs/go-ipfs/core"
 	"github.com/ipfs/go-ipfs/core/commands/cmdenv"
+	"github.com/ipfs/go-ipfs/core/node"
 
 	bservice "github.com/ipfs/go-blockservice"
 	cid "github.com/ipfs/go-cid"
@@ -398,20 +400,23 @@ GC'ed.
 			return fmt.Errorf("cp: cannot get node from path %s: %s", src, err)
 		}
 
+		filesRoot := nd.LockedFilesRoot.Lock()
+		defer nd.LockedFilesRoot.Unlock()
+
 		if mkParents {
-			err := ensureContainingDirectoryExists(nd.FilesRoot, dst, prefix)
+			err := ensureContainingDirectoryExists(filesRoot, dst, prefix)
 			if err != nil {
 				return err
 			}
 		}
 
-		err = mfs.PutNode(nd.FilesRoot, dst, node)
+		err = mfs.PutNode(filesRoot, dst, node)
 		if err != nil {
 			return fmt.Errorf("cp: cannot put node in path %s: %s", dst, err)
 		}
 
 		if flush {
-			_, err := mfs.FlushPath(req.Context, nd.FilesRoot, dst)
+			_, err := mfs.FlushPath(req.Context, filesRoot, dst)
 			if err != nil {
 				return fmt.Errorf("cp: cannot flush the created file %s: %s", dst, err)
 			}
@@ -426,7 +431,9 @@ func getNodeFromPath(ctx context.Context, node *core.IpfsNode, api iface.CoreAPI
 	case strings.HasPrefix(p, "/ipfs/"):
 		return api.ResolveNode(ctx, path.New(p))
 	default:
-		fsn, err := mfs.Lookup(node.FilesRoot, p)
+		filesRoot := node.LockedFilesRoot.RLock()
+		defer node.LockedFilesRoot.RUnlock()
+		fsn, err := mfs.Lookup(filesRoot, p)
 		if err != nil {
 			return nil, err
 		}
@@ -491,7 +498,10 @@ Examples:
 			return err
 		}
 
-		fsn, err := mfs.Lookup(nd.FilesRoot, path)
+		filesRoot := nd.LockedFilesRoot.RLock()
+		defer nd.LockedFilesRoot.RUnlock()
+
+		fsn, err := mfs.Lookup(filesRoot, path)
 		if err != nil {
 			return err
 		}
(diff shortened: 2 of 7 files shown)


BigLep:
Assigned to @petar for review as this is a newer area for all the current go-ipfs stewards.

core/commands/files.go
```diff
@@ -792,20 +795,25 @@ stat' on the file or any of its ancestors.
 		if err != nil {
 			return err
 		}
+		filesRoot := nd.LockedFilesRoot.LockRoot()
```
petar, line 798: @aschmahmann This design takes a high-level lock on all of MFS while running cli commands or the GC. On the one hand, this approach is a little error-prone because we need to make sure that this lock is used in all app-level code that touches MFS. On the other hand, after doing some thinking, I don't think it is easy to implement per-node locking in MFS. So, for now, it seems that the approach in this PR is fine for fixing the issue in the short term.

aschmahmann: Taking a high level lock here is unfortunate, but isn't really any different than locking around Pin operations. It also seems fairly straightforward to implement.

In theory we could leverage the existence of the flush command here to allow accumulating data that's protected from GC, but given other work we'd like to do around MFS and GC (independently and together) it doesn't seem worth the effort at the moment.

schomatis:
@aschmahmann From your previous comment I take it you approve the direction of this PR but could you make it explicit here just to make sure I'm on the right track before proceeding, please?

aschmahmann:
@schomatis yes, having a global write lock on MFS seems like the best thing to do in this PR.

For reads we might want to be more careful than a global lock though. For example, not locking `ipfs files ls` or `ipfs files stat` while GC is ongoing, but preventing corruption between MFS write and read operations (out of date data, is fine with concurrent operations, but corruption would be bad). It might be that MFS already handles some of this, but we should verify.

If we find ourselves stuck with locking on some of the read commands I'd really like `ipfs files stat /ipfs/....` to not utilize the same locking mechanism since the data is immutable and shouldn't need to be locked. It's also important in order to make sure nobody has a reason to continue using `ipfs object stat`.

schomatis:
Continuing with this.

schomatis:
> I'd really like ipfs files stat /ipfs/.... to not utilize the same locking mechanism since the data is immutable and shouldn't need to be locked. 

Confirming this is already the case since the `/ipfs/` prefix is handled through a different API and doesn't use the MFS root:

https://github.com/ipfs/go-ipfs/blob/c00065cc8412ec660037f3fed4df7d3fed467b17/core/commands/files.go#L424-L429

core/node/core.go
```diff
@@ -168,3 +169,54 @@ func Files(mctx helpers.MetricsCtx, lc fx.Lifecycle, repo repo.Repo, dag format.
...
+// The objective is:
+// > For reads we might want to be more careful than a global lock though.
+// > For example, not locking ipfs files ls or ipfs files stat while GC is ongoing,
+// > but preventing corruption between MFS write and read operations (out of date data
+// > is fine with concurrent operations, but corruption would be bad).
+// FIXME(BLOCKING): Decide how are we actually going to implement the locking
+//  distinction between reading and writing.
+// FIXME(BLOCKING): Consider providing our own API functions like GetNode (and similar) that
+//  already take charge of lock-handling here instead of leaving that burden to the consumer.
+//  That way we could also make sure the lock is being enforced correctly.
```
petar, line 194: I would leave things exactly as you have them right now. There are bigger issues with MFS locking that cannot be resolved by any sort of locking on the root alone. This PR is trying to just fix races between GC and competing command-line MFS operations.

petar: I like the interface of LockedFilesRoot, because having only RLock and Lock methods forces whoever is touching this code to think/ask what is going on.

core/node/core.go
```diff
@@ -168,3 +169,54 @@ func Files(mctx helpers.MetricsCtx, lc fx.Lifecycle, repo repo.Repo, dag format.
...
+
+// FIXME(BLOCKING): Review and cofirm.
+// Not enforcing the lock for reading operations to see if something comes up
+// in the tests. This is dangerous because MFS is not designed to clearly
+// distinguish between read-only and mutating operations. For example, the FSNode
+// interface, which is the result used from the most frequent "read" operation of
+// doing a Lookup, (a) allows the Flush operation and (b) its File/Directory structures
+// implementing it have their own locking mechanism to keep track of open file
+// descriptors.
+func (lfr *LockedFilesRoot) RLock() *mfs.Root {
```
petar, line 208: That's the best we can do for now without a total redesign of MFS, I think.

cmd/ipfs/pinmfs.go
```diff
@@ -50,7 +50,11 @@ type ipfsPinMFSNode struct {
 }
 
 func (x *ipfsPinMFSNode) RootNode() (ipld.Node, error) {
-	return x.node.FilesRoot.GetDirectory().GetNode()
+	// FIXME(BLOCKING): We're deploying this lock taking mechanism as simply as
+	//  possible for now deferring the unlock but we should be more precise.
```
petar, line 54: See/Similar to comments below. That's the best we can do now. Even though this code has no effect at all (because the root cannot change inside LockedFilesRoot), it still seems useful to make someone staring at the code to be puzzled and investigate what this lock is all about.

test-mfs-lock.bash
```diff
@@ -0,0 +1,13 @@
...
+set -v
+
+# The daemon needs to be running, otherwise the commands will compete on their
+# lock of the entire repo (not the MFS root) and fail.
+ipfs swarm addrs > /dev/null || (echo "daemon not running" && exit 1)
+
+ipfs --version
+ipfs files mkdir /test-lock/
+ipfs files rm /test-lock/ -r
+((echo "content" | ./cmd/ipfs/ipfs files write --create --parents --truncate --lock-time 3 /test-lock/file) && echo "ipfs write lock released" &)
```
petar, line 11: I think the way you've implemented the timeout the command `ipfs files write` does not exit until the timeout expires, consequently the lock is never present for the next command `ipfs repo gc`

core/node/core.go
schomatis: Yes, sorry, this is actually a non-blocking fixme, just a note of a possible direction down the road, will clarify that.

test-mfs-lock.bash
schomatis: > the command ipfs files write does not exit until the timeout expires,

Correct. My bash is very poor but the idea was to have the above command running in the background while `ipfs repo gc` attempts to take the lock and has to wait until `ipfs write` releases it, where then I see both command's outputs roughly simultaneously.

But anyway, this is just a temporary ad-hoc test while in the draft stage; I still don't know how to actually test this in production to get this code ready to merge.

schomatis:
@aschmahmann Petar confirmed the current proposal but I'm still stuck on how to test it other than some manual confirmation than the lock is being taken. How would you like to test this to wrap it up?

lidel:
Closing as superseded: #11386 fixed the same GC-vs-MFS-write hole (#6113) that this targeted, taking the pin lock across MFS mutations instead of a `LockedFilesRoot`. Thanks for the early work here.

PR #8457, opened 2021-09-24
docs(ipfs.add): flag behavior around cidv1

### LINKED ISSUE #8396: `ipfs add --raw-leaves` changed with a side effect of unixfs update
### Checklist

### Installation method

built from source

### Version

_No response_

### Config

_No response_

### Description

The `--raw-leaves` option is currently marked `(experimental)` but also happen to now be true by default when it used to be false.

This is because this option is only passed if set:
https://github.com/ipfs/go-ipfs/blob/c9cc09f6f7ebe95da69be6fa92c88e4cb245d90b/core/commands/add.go#L222-L224

So we are just relying on unixfs's default. Which changed to true recently.

I've seen one complain about hash changing which made me investigate this.
However I don't think it's an issue. I quite like `--raw-leaves` anyway (#8201), I think it's fine just removing the `(experimental)` and making `--raw-leaves` default.

For context, I belive `--raw-leaves` isn't default because it would previous change hashes and the goal was to bundle multiple hash change at once.
But a hash change happened somewhat recently and we havn't noticed. To me this issue is over estimated and hash changing are fine (Let's enable `--inline` and directory sharding.).

Note also in theory if a single block wide file is a subset of a multiple block wide file with using raw leaves they can't be dedduped easly.
Without `--raw-leaves` in theory the bigger file just links to the smaller one.
But with it can't because the bigger one will want a raw block, not a cbor serialized file.
If that is really an issue (which is currently not because no chunker is smart enough), this can be solved by mixing raw and serialised leaves (which idk if we can do yet), or make single block wide file actually two blocks, a raw leave and a cbor root only linking to that leaf (which is worst because the network could forget the root but not the leaf, which is lost potential).

lidel:
Closes #8396 

 @Jorropo would this make things clean enough?


Jorropo:
@lidel you got the things the wrong way around.

My issue is that CIDv1 implies raw leaf unless the user specify it otherwise. (`CIDv1 -> Rawleaves`)
Your PR explain that raw leaf implies CIDv1 which is false, I mean just test it raw leaves on CIDv0 just works. (`Rawleaves -> CIDv1` *which is false*)

I hope this clears it out :)

Jorropo (changes requested):
See the comment

Jorropo:
By the way, when I've said : "Your PR explain that raw leaf implies CIDv1 **which is false**" that is partialy wrong.
If the target file is smaller than the blocksize CIDv1 is implied, else (if the file gets chunked) CIDv0 is implied.

PR #8460, opened 2021-09-27
chore: update go-libp2p to v0.15.0

aschmahmann:
cc @marten-seemann any other dependency bumps that should be going into the go-ipfs v0.10.0?


marten-seemann:
Duplicate of #8422?

aschmahmann:
Ah you're right. Thanks 🙏

PR #8462, opened 2021-09-27
fuse: load hamt adls as their dagpb substrates

aschmahmann:


fuse/readonly/readonly_unix.go
```diff
@@ -89,7 +90,13 @@ func (s *Root) Lookup(ctx context.Context, name string) (fs.Node, error) {
 	var fnd ipld.Node
 	switch cidLnk.Cid.Prefix().Codec {
 	case cid.DagProtobuf:
-		fnd, err = mdag.ProtoNodeConverter(blk, nd)
+		adl, ok := fnd.(ipldprime.ADL)
```
guseggert, line 93: fnd is always nil here, so this always fails :(

guseggert (approved):

Issue #7022, opened 2020-03-20
Automatic sharding of big directories


Stebalien:
This is a tracking issue for the sharded directory feature. This feature can be enabled with:

```bash
> ipfs config --json Experimental.ShardingEnable true
```

TODOs:

* [ ] Make sure we don't shard too soon. At the moment, the cutoff to start sharding is very low.
* [ ] Ideally, avoid creating sharded directories thar are too "deep". However, this may be tricky to do without breaking changes.

Ideally, we'd fix this issue by switching to unixfs 2.0 and a generalized sharded map implementation. However, that's probably not going to happen in the near future and this feature is used widely enough that we'll likely have to stabilize it in its current form.

achingbrain:
I'd really like to enable sharding by default in js-IPFS to reduce the amount of codepaths and A/B tests.

`ipfs.add` in js-IPFS takes a `shardingSplitThreshold` argument that sounds similar to the sharding cutoff above - it's set to 1000 by default when sharding is enabled.

Can go-IPFS take the same option?  That should tick off the first item.

cc @aschmahmann

lidel:
We _really_ should fix this. js-ipfs shards dirs >1k of items out-of-the-box  and go-ipfs fails flat.

On top of that, the biggest problem in go-ipfs right now is that when you enable sharding, it shards EVERYTING, so even if you have a small dir  under 1k of items, its still harded, which produces different CIDs.

I believe a lot of work happened in IPLD team since we implemented HAMT-sharding in go-ipfs, would be useful to leverage that work when we make sharding enabled by defualt in go-ipfs.

@warpfork are there any libs / specs related to improved sharding in IPLD which we should be aware of in planning here?

warpfork:
Yes!

- We shifted to trying to make sharding "in scope" for IPLD by defining it as something you do with an ADL ("Advanced Data Layout"), which is a plugin-like concept.

- A lot of work was done on HAMTs as a map sharding mechanism.  There's a spec in https://github.com/ipld/specs/blob/master/data-structures/hashmap.md ; a golang implementation in https://github.com/ipld/go-ipld-adl-hamt ; and a JS implementation in ((hm, @rvagg, give me the latest link?)).

[...]

PR #8465, opened 2021-09-28
test: add sharness test for reading ADLs with FUSE

guseggert:
I verified that this fails on v0.10.0-RC1


test/sharness/lib/test-lib.sh
```diff
@@ -287,7 +287,7 @@ test_launch_ipfs_daemon_without_network() {
 
 do_umount() {
   if [ "$(uname -s)" = "Linux" ]; then
-  fusermount -u "$1"
+  fusermount -z -u "$1"
```
guseggert, line 290: this is "lazy unmount" so we don't have to sleep after unmounting so that we don't get errors when removing mount dirs too quickly

lidel (changes requested):
LGTM, but creating test data requires enabling experimental flag.

This is good for now, but when on-demand sharding (https://github.com/ipfs/go-ipfs/issues/7022) lands things will break. 

Perhaps we could refactor these tests to avoid experimental flag, and instead `dag import` sharded fixtures from a  `.car` archive?

test/sharness/t0032-mount-sharded.sh
```diff
@@ -0,0 +1,62 @@
...
+
+# we're testing nested subdirs which ensures that IPLD ADLs work
+test_expect_success 'setup test data' '
+  mkdir testdata &&
+  echo a > testdata/a &&
+  mkdir testdata/subdir &&
+  echo b > testdata/subdir/b
+'
+
+HASH=$(ipfs add -r -Q testdata)
```
BigLep, line 33: @aschmahmann feedback to hardcode the hash.

PR #8467, opened 2021-09-28
ci: move Docker image build to Actions

lidel:
This PR aims to close #8330 and unblock 0.10.0-rc2

## TODO
- [x] add github action workflow and reuse `bin/push-docker-tags.sh` (so we get same image names)
- [x] set up docker hub secrets for publishing (`DOCKER_USERNAME` & `DOCKER_PASSWORD`)
  - using temporary credentials for testing, need to switch to ones from `ipfsbot` but we can do that later this week
- [x] remove docker publishing from circleci scripts

PR #8468, opened 2021-09-28
Revert "ci: use dynamic config for CircleCI"

guseggert:
This reverts commit b2c3959a4f882235068fc10088844c6fddc7e7e6.


lidel (commented):
Let's try https://github.com/ipfs/go-ipfs/pull/8467 first, this will be our Plan B if Actions do not work for us.

PR #8469, opened 2021-09-29
feat: add endpoint for enabling block profiling

guseggert:
After this is added I'll go back and add it to `bin/collect-profiles.sh` and `docs/debug-guide.md`


lidel (approved):
Doing this via POST lgtm – protects from websites probing via img/iframe tags etc.

@guseggert feel free to merge if this is still useful (or close if not)

Issue #8264, opened 2021-07-16
CLI cannot process multiple passed files with the same name


bmwiedemann:
#### Version information:
```
go-ipfs version: 0.9.0
Repo version: 11
System version: amd64/linux
Golang version: go1.16
```

#### Description:

For https://github.com/bmwiedemann/ipfs-iso-jigsaw/blob/prehashnodirs/prehash.py#L53  I am trying to add files without directories into ipfs and noticed that directories are not printed in output, but what is worse, files are silently not added (or at least not reported)

```
mkdir -p /tmp/x/{1,2}
echo x > /tmp/x/1/x
echo y > /tmp/x/2/x
ipfs add -n --raw-leaves -- /tmp/x/1/x /tmp/x/2/x
added bafkreib3wkv3nhv3e7574y6hmolcjrxmlyzrxba2lpemh26bbojil2iio4 x
```

aschmahmann:
Thanks for the report, it looks like the CLI command parser (https://github.com/ipfs/go-ipfs-cmds/blob/4ade007405e5d3befb14184290576c63cc43a6a3/cli/parse.go#L239) bundles up all the files into a map keyed by the file name (not path, inode, etc.) of the files leading to the described issue.

Perhaps just keying on path name instead of file name (i.e. removing https://github.com/ipfs/go-ipfs-cmds/blob/4ade007405e5d3befb14184290576c63cc43a6a3/cli/parse.go#L335) would do the job.

bmwiedemann:
I tested that L335 patch and indeed, it helps to add both files and also helps to get distinct path names in the output:
```
ipfs add -n --raw-leaves -- /tmp/x/1/x /tmp/x/2/x
added bafkreidtzm4frjuhvbeuzizsgbjqcyuc6pnnhhkcz5rmuttz3wrkvr6zvq /tmp/x/1/x
added bafkreib3wkv3nhv3e7574y6hmolcjrxmlyzrxba2lpemh26bbojil2iio4 /tmp/x/2/x
cd /tmp/
ipfs add -n --raw-leaves -- x/1/x x/2/x
added bafkreidtzm4frjuhvbeuzizsgbjqcyuc6pnnhhkcz5rmuttz3wrkvr6zvq x/1/x
added bafkreib3wkv3nhv3e7574y6hmolcjrxmlyzrxba2lpemh26bbojil2iio4 x/2/x
```

Thanks.

bmwiedemann:
[...]

PR #1536, opened 2015-07-29
ipfs add -w directories (and FUSE fixes)

jbenet:
a8dae30 (Juan Batiz-Benet, 4 minutes ago)
    add -w: fix to work correctly with dirs.

    this commit changes the behavior of ipfs add -w:

    - it makes it able to work with ipfs add -r <dir>
    - instead of hacking around the add, we simply just add a wrapper
     directory around the whole result of the add. this means that
     ipfs add -w calls will output _two_ lines, but this is actually
     more correct than outputting one line, as two objects were added.
     this _may_ break scripts out there which expect the output to
     look a certain way. we should consider whether the old output is
     more _useful_ (even if less in-line with the model.)

    License: MIT Signed-off-by: Juan Batiz-Benet <juan@benet.ai>

 578fd02 (Juan Batiz-Benet, 7 minutes ago)
    fuse unmount fixes

    unmounting wasn't happening, mostly because of a recent bug in
    goprocess.SetTeardown. This commit bumps up some messages to log.Warnings,
    as users may want to see them, and makes sure to Unmount when a node shuts
    down.

    License: MIT Signed-off-by: Juan Batiz-Benet <juan@benet.ai>

 7a41dcb (Juan Batiz-Benet, 26 minutes ago)
    updated goprocess (SetTeardown fix)

    License: MIT Signed-off-by: Juan Batiz-Benet <juan@benet.ai>

PR #8493, opened 2021-10-06
fix(cli): ipfs add with multiple files of same name

schomatis:
This PR incorporates the `go-ipfs-cmds` [fix](https://github.com/ipfs/go-ipfs-cmds/pull/220) that makes the `ipfs add` command not ignore repeated entries. This fix conflicts with another bug in the `ipfs add -w` variant of the command which can't correctly process files with the same name (I think this was uncovered here and we don't have a specific issue for it). This is shown by the shanress test failing (see more details below). We need to first fix the `ipfs add -w` issue before landing this so we can be confident this is the correct fix.

---------------------------------------------------

#### Detail of the `ipfs add -w` bug and the [test](https://github.com/ipfs/go-ipfs/blob/2a871ef0184da3b021209881dfa48e3cfdaa9d26/test/sharness/t0043-add-w.sh#L131-L142) that [fails](https://app.circleci.com/pipelines/github/ipfs/go-ipfs/5975/workflows/fd82d174-de6f-43cf-ad3c-8102b908e81f/jobs/65050) (uncovering it)

The `ipfs add -w (repeats)` test now fails. This is expected because:
* we repeatedly add the `m/4r93` file
* ignored before, these extra/repeated files are now sent to the core UnixFS API `Add()`
* this API uses a [fake (disposable) MFS root directory](https://github.com/ipfs/go-ipfs/blob/52a747763f6c4e85b33ca051cda9cc4b75c815f9/core/coreunix/add.go#L84-L96) as a way to interact with the underlying DAG service (repo)
* with the "wrapped" option `-w` we group all the entries being added in a [single directory](https://github.com/ipfs/go-ipfs/blob/52a747763f6c4e85b33ca051cda9cc4b75c815f9/core/commands/add.go#L194-L199)
* the MFS directory, not intended for this ephemeral abuse, correctly refuses to add two entries with the same name (with the resulting error in sharness of a failed command: `Error: directory already has entry by that name`)

We are abusing the MFS API, see related and broader issue for more details on this general topic: [Decouple MFS from UnixFS](https://github.com/ipfs/go-ipfs/issues/8541).

Also note that this issue might not happen with *any* kind of repeated entry in the `ipfs add -w` command as empty directories seem to be ignored when constructing the path and only files within those repeated directories will trigger the error, like `ipfs add -w -r dir/file dir/file`. There is some undocumented logic around path manipulation related to this:

https://github.com/ipfs/go-ipfs/blob/52a747763f6c4e85b33ca051cda9cc4b75c815f9/core/coreunix/add.go#L411-L424

--------------------------------------------------

Fixes https://github.com/ipfs/go-ipfs/issues/8264.

cc @guseggert

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -31,7 +31,7 @@ require (
 	github.com/ipfs/go-graphsync v0.11.0
 	github.com/ipfs/go-ipfs-blockstore v1.2.0
 	github.com/ipfs/go-ipfs-chunker v0.0.5
-	github.com/ipfs/go-ipfs-cmds v0.7.0
+	github.com/ipfs/go-ipfs-cmds v0.8.0
 	github.com/ipfs/go-ipfs-exchange-interface v0.1.0
 	github.com/ipfs/go-ipfs-exchange-offline v0.2.0
 	github.com/ipfs/go-ipfs-files v0.0.9
--- a/test/sharness/t0040-add-and-cat.sh
+++ b/test/sharness/t0040-add-and-cat.sh
@@ -345,6 +345,27 @@ test_add_cat_file() {
     echo "added Qmf35k66MZNW2GijohUmXQEWKZU4cCGTCwK6idfnt152wJ hello2.txt" >> expected &&
     test_cmp expected actual
   '
+
+  test_expect_success "ipfs add with multiple files of same name succeeds" '
+    mkdir -p mountdir/same-file/ &&
+    cp mountdir/hello.txt mountdir/same-file/hello.txt &&
+    ipfs add mountdir/hello.txt mountdir/same-file/hello.txt >actual &&
+    rm mountdir/same-file/hello.txt  &&
+    rmdir mountdir/same-file
+  '
+
+  test_expect_success "ipfs add with multiple files of same name output looks good" '
+    echo "added QmVr26fY1tKyspEJBniVhqxQeEjhF78XerGiqWAwraVLQH hello.txt" >expected &&
+    test_cmp expected actual
+  '
+
+    test_must_fail "ipfs add with multiple files of same name but different dirs fails" '
+      mkdir -p mountdir/same-file/ &&
+      cp mountdir/hello.txt mountdir/same-file/hello.txt &&
+      ipfs add mountdir/hello.txt mountdir/same-file/hello.txt >actual &&
+      rm mountdir/same-file/hello.txt  &&
+      rmdir mountdir/same-file
+    '
 }
 
 test_add_cat_5MB() {
--- a/go.sum
+++ b/go.sum
@@ -473,8 +473,8 @@ github.com/ipfs/go-ipfs-blocksutil v0.0.1/go.mod h1:Yq4M86uIOmxmGPUHv/uI7uKqZNtL
 github.com/ipfs/go-ipfs-chunker v0.0.1/go.mod h1:tWewYK0we3+rMbOh7pPFGDyypCtvGcBFymgY4rSDLAw=
 github.com/ipfs/go-ipfs-chunker v0.0.5 h1:ojCf7HV/m+uS2vhUGWcogIIxiO5ubl5O57Q7NapWLY8=
 github.com/ipfs/go-ipfs-chunker v0.0.5/go.mod h1:jhgdF8vxRHycr00k13FM8Y0E+6BoalYeobXmUyTreP8=
-github.com/ipfs/go-ipfs-cmds v0.7.0 h1:0lEldmB7C83RxIOer38Sv1ob6wIoCAIEOaxiYgcv7wA=
-github.com/ipfs/go-ipfs-cmds v0.7.0/go.mod h1:y0bflH6m4g6ary4HniYt98UqbrVnRxmRarzeMdLIUn0=
+github.com/ipfs/go-ipfs-cmds v0.8.0 h1:M7apkPxhGe7I3rcKuQ8xRJLIPdaEqaZhJz0uPZEE8EU=
+github.com/ipfs/go-ipfs-cmds v0.8.0/go.mod h1:y0bflH6m4g6ary4HniYt98UqbrVnRxmRarzeMdLIUn0=
 github.com/ipfs/go-ipfs-delay v0.0.0-20181109222059-70721b86a9a8/go.mod h1:8SP1YXK1M1kXuc4KJZINY3TQQ03J2rwBG9QfXmbRPrw=
 github.com/ipfs/go-ipfs-delay v0.0.1 h1:r/UXYyRcddO6thwOnhiznIAiSvxMECGgtv35Xs1IeRQ=
 github.com/ipfs/go-ipfs-delay v0.0.1/go.mod h1:8SP1YXK1M1kXuc4KJZINY3TQQ03J2rwBG9QfXmbRPrw=


go.mod
```diff
@@ -29,7 +29,7 @@ require (
 	github.com/ipfs/go-graphsync v0.8.0
 	github.com/ipfs/go-ipfs-blockstore v0.1.6
 	github.com/ipfs/go-ipfs-chunker v0.0.5
-	github.com/ipfs/go-ipfs-cmds v0.6.0
+	github.com/ipfs/go-ipfs-cmds v0.6.1-0.20211005151237-0e6ec5a05153
```
schomatis, line 32: Do we need to create a release here after tests pass?

schomatis:
@aschmahmann WIth the fix from https://github.com/ipfs/go-ipfs/pull/8538 and [not ignoring repeated entries](https://github.com/ipfs/go-ipfs-cmds/pull/220) the `ipfs add -w (repeats)` test (that originally had a different CID [in this PR](https://github.com/ipfs/go-ipfs/pull/8493#discussion_r725151698)) now correctly fails because:
* we repeatedly add the `m/4r93` file
* these extra/repeated files (ignored before) now are issued to the core UnixFS API `Add()` here
* this API uses a [fake (disposable) MFS root directory](https://github.com/ipfs/go-ipfs/blob/52a747763f6c4e85b33ca051cda9cc4b75c815f9/core/coreunix/add.go#L84-L96) as a way to interact with the underlying DAG service (repo) grouping all the entries being added in a single directory
* the MFS directory, not intended for this ephemeral abuse, correctly refuses to add two entries with the same name
* (the fact that we don't really care about this fake MFS root, acting here just as a temporary container/directory for the entries being added in a single `ipfs add` call, and we would allow its inconsistent state with duplicate entries is irrelevant to the `go-mfs` stack)

schomatis:
Looking closer into the add command logic, each entry should have its own and independent (temporal) MFS root so will need to dig deeper on why this test alone is failing an not any other command like `ipfs add file dir/file`.

schomatis:
Able to reproduce the bug. Empty directories seem to be ignored when constructing the path and only files within those repeated directories will trigger the error, like `ipfs add -w -r dir/file dir/file`.

There is some obscure and undocumented logic around path manipulation that I will need some design clarifications on, like:

https://github.com/ipfs/go-ipfs/blob/52a747763f6c4e85b33ca051cda9cc4b75c815f9/core/coreunix/add.go#L411-L424

test/sharness/t0043-add-w.sh
```diff
@@ -51,7 +51,7 @@ added QmQkib3f9XNX5sj6WEahLUPFpheTcwSRJwUCSvjcv8b9by _jo7
 added QmVPwNy8pZegpsNmsjjZvdTQn4uCeuZgtzhgWhRSQWjK9x gnz66h
 added QmTmc46fhKC8Liuh5soy1VotdnHcqLu3r6HpPGwDZCnqL1 '
 
-add_w_r='QmcCksBMDuuyuyfAMMNzEAx6Z7jTrdRy9a23WpufAhG9ji'
+add_w_r='Qme987pqNBhZZXy4ckeXiR7zaRQwBabB7fTgHurW2yJfNu'
```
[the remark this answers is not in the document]
schomatis: Moving this to the PR description.

guseggert:
What's your estimate on how long it would take to fix https://github.com/ipfs/go-ipfs/issues/8541 ?  If we fix that, then all the weird edge cases here would go away, right?

schomatis:
@guseggert 

> What's your estimate on how long it would take to fix https://github.com/ipfs/go-ipfs/issues/8541 ?

This largely depends on review bandwidth and (optionally) being able to finally [consolidate](https://github.com/ipfs/go-ipfs/issues/8543) UnixFS/MFS back here (as we would likely be breaking a bunch of APIs).

> If we fix that, then all the weird edge cases here would go away, right?

The general scope of that issue is too broad and vague at the moment, but achieving the specific use case of removing MFS from `ipfs add` is not hard and I think doing that would, if not solve _every_ edge case here, largely simplify this issue.

aschmahmann:
> We are abusing the MFS API, see related and broader issue for more details on this general topic: https://github.com/ipfs/go-ipfs/issues/8541.

It doesn't seem like this is really about MFS as much as it is about some unintuitive behavior around what `-w` does with repeat parameters that leads to the behavior being undefined in the case of multiple files with the same name.

 ```
 ❯ ipfs add .\foo .\bar
added QmaZAaCrugAkinkaSGBP2uD8pchyKXY2rmPf8WQ2CzE2pN bar
added QmQ7qCwMvm3mb5PRaAJd1ytaAE4qULHzoJCunikZfT1oAp foo
 ```
 
This seems to indicate that `ipfs add <file1> <file2>` is just a shortcut for calling `ipfs add file1` followed by `ipfs add file2`.

```
❯ ipfs add -w .\foo .\bar
added QmaZAaCrugAkinkaSGBP2uD8pchyKXY2rmPf8WQ2CzE2pN bar
added QmQ7qCwMvm3mb5PRaAJd1ytaAE4qULHzoJCunikZfT1oAp foo
added QmSQUBkpW1ApwzXK8Z23P4uL8jJkss31rd49bWG2SR6c2g
```

This seems to indicate that `ipfs add -w <file1> <file2>` adds file1 and file2 into a shared directory keyed by their file names. While this is an interesting and perhaps useful behavior, it's not the same as the repeated adds being a shortcut for `ipfs add -w <file1>` followed by `ipfs add -w file2>`.

This behavior seems to have been introduced intentionally ages ago https://github.com/ipfs/go-ipfs/pull/1536 and while likely not that common (due to bugs like this one with multiple files with the same name not being repeated flagged + upvoted by users) it's definitely in use (e.g. a cursory look on grep.app shows https://github.com/ipfs/public-gateway-checker/blob/16c99857e4f4cb67ff5a8bc617ec9fa44bdc8a6d/publish-to-ipfs.sh#L3)

In this case what happens when you do `ipfs add -w .\test1\foo .\test2\foo` is undefined and there could be any number of potential options:
1. Only one file gets added and that one is wrapped in a directory
2. Both files are added and only one gets wrapped in a directory
3. Both files are added and are put in the same directory but one of them has an autogenerated name (e.g. `foo(1)`)
4. There's an error

It seems to me like the only sensible thing to do is to error if `ipfs add -w` contains entries with the same name. I'm not sure what the intent behind this test https://github.com/ipfs/go-ipfs/blob/2a871ef0184da3b021209881dfa48e3cfdaa9d26/test/sharness/t0043-add-w.sh#L131-L142 was, but if we want we can maintain compatibility here by only erroring if the entry paths are different (e.g. allowing `ipfs add -w foo foo` to be deduplicated to remove the last `foo`).

aschmahmann:
> Maybe an example of when would we error?

The undefined behavior of when we should error is `ipfs add -w .\test1\foo .\test2\foo`.

The sharness test you linked https://github.com/ipfs/go-ipfs/blob/2a871ef0184da3b021209881dfa48e3cfdaa9d26/test/sharness/t0043-add-w.sh#L131-L142 is pretty weird in that it does intentionally do the equivalent of `ipfs add -w .\test1\foo .\test1\foo` and not expect an error.

At a high level if we want to continue to support that case (instead of just erroring from two files called foo), we can do so by either:
1. Checking if the import paths are the same `.\test1\foo .\test1\foo` and just not parsing the data a second time (does enable the backwards compatibility with the linked test)
2. Checking if the data in `.\test1\foo` and `.\test2\foo` are the same (and therefore will result in the same CID) and then ignoring the second entry (a more generous way of enabling backwards compatibility in more situations where it's safe)

Aside from the fact that technically this would be an unforced breaking change, if 1 or 2 are a pain to implement we'd probably be fine to modify the sharness test and just not support entries with the same name whether they have the same path or content. We'll be generating errors so in the unlikely event this breaks someone's workflow it'll be detectable and we could deal with it later. I'd rather not have breaking changes if we can avoid it, but we can evaluate depending on how painful you think it'll be to implement.

schomatis:
Got it, thanks. I think we can implement this by extending the `-cmds` lib to retain the path and decide what to do in `ipfs add` here.

schomatis:
According to Discord [discussion](https://discord.com/channels/806902334369824788/943876573545779210/943897548702240818) we're going with:

3. Error if there would be a conflicting path name, but don't if the DAGs from the conflicting names would be identical - shouldn't require any API changes to go-ipfs-cmds or go-ipfs-files and breaks the least backwards compatibility. Might be a bit more work to deal with in the coreunix.Adder and we'll likely end up putting some blocks in the repo that will only get removed with GC, but that's fine.

to minimize messing around with go-ipfs-cmds or go-ipfs-files.

schomatis:
@aschmahmann 

I feel we have been going back and forth on the issue of which error to display to the user in `go-ipfs` which is not a blocker for my work (nor do I have any strong opinion on it).

My **blocker** is the following: AFAICT _any_ of the proposed errors need `go-ipfs` to be aware of the import path, which is not available at the moment. I have asked for a simple way to get it in [Discord](https://discord.com/channels/806902334369824788/942673321852563456/945446789627129857) and didn't get an answer. I have gotten pushback to modifying upstream dependencies but don't see any other way, so **I need a decision on which dependency to modify.** I'll repeat my original [Discord question](https://discord.com/channels/806902334369824788/942673321852563456/943876573545779210) here:

>  I'm thinking we can satisfy https://github.com/ipfs/go-ipfs/pull/8493#issuecomment-1041652332 with either:
> * extend Node interface to also provide the path and filter/error in ipfs add - https://github.com/ipfs/go-ipfs-files/blob/5044b18695cb23ae69632c3e2511b4624a87b0d5/file.go#L16-L24
> * do some preprocessing in go-ipfs-cmds to filter/error files there before reaching ipfs add

guseggert:
Let's roll with doing the preprocessing in go-ipfs-cmds.  Does that unblock you?

schomatis:
Yes, thanks.

schomatis:
@guseggert Ready for review.

go.mod
```diff
@@ -31,7 +31,7 @@ require (
 	github.com/ipfs/go-graphsync v0.11.0
 	github.com/ipfs/go-ipfs-blockstore v1.1.2
 	github.com/ipfs/go-ipfs-chunker v0.0.5
-	github.com/ipfs/go-ipfs-cmds v0.6.0
+	github.com/ipfs/go-ipfs-cmds v0.6.1-0.20220321213443-7217331bf422
```
schomatis, line 34: This should be tagged, Gus could you cut a release when the dependency PR lands, please? (If it's a one-liner feel free to paste it here and I'll take care of it.)

schomatis: Thanks, updating.

schomatis:
@guseggert Updated dependency, this should be ready now.

guseggert (approved):

test/sharness/t0040-add-and-cat.sh
```diff
@@ -345,6 +345,27 @@ test_add_cat_file() {
...
+  test_expect_success "ipfs add with multiple files of same name succeeds" '
+    mkdir -p mountdir/same-file/ &&
+    cp mountdir/hello.txt mountdir/same-file/hello.txt &&
+    ipfs add mountdir/hello.txt mountdir/same-file/hello.txt >actual &&
+    rm mountdir/same-file/hello.txt  &&
+    rmdir mountdir/same-file
+  '
+
+  test_expect_success "ipfs add with multiple files of same name output looks good" '
+    echo "added QmVr26fY1tKyspEJBniVhqxQeEjhF78XerGiqWAwraVLQH hello.txt" >expected &&
```
bmwiedemann, line 358: Shouldn't we expect 2 output lines if we add 2 files?
Did you try to have the 2nd file with different content?

schomatis: Will take another look at the test this week. Thanks for flagging.

PR #8502, opened 2021-10-11
buffer the entire dag before export

coryschwartz:
This will buffer the entire dag.

I don't think this is the best option, but the idea here is to buffer the entire dag so errors can be detected, if an error is detected, the dag won't be exported.

The downside is obvious, that the entire dag is buffered.


aschmahmann:
As you mentioned buffering seems pretty bad for performance and usability here.

PR #8510, opened 2021-10-13
feat: enabling pubsub and ipns-pubsub via config flags

jwh:
Depends on https://github.com/ipfs/go-ipfs-config/pull/145

### DIFF
--- a/cmd/ipfs/daemon.go
+++ b/cmd/ipfs/daemon.go
@@ -178,8 +178,8 @@ Headers.
 		cmds.BoolOption(enableGCKwd, "Enable automatic periodic repo garbage collection"),
 		cmds.BoolOption(adjustFDLimitKwd, "Check and raise file descriptor limits if needed").WithDefault(true),
 		cmds.BoolOption(migrateKwd, "If true, assume yes at the migrate prompt. If false, assume no."),
-		cmds.BoolOption(enablePubSubKwd, "Instantiate the ipfs daemon with the experimental pubsub feature enabled."),
-		cmds.BoolOption(enableIPNSPubSubKwd, "Enable IPNS record distribution through pubsub; enables pubsub."),
+		cmds.BoolOption(enablePubSubKwd, "Enable experimental pubsub feature. Overrides Pubsub.Enabled config."),
+		cmds.BoolOption(enableIPNSPubSubKwd, "Enable IPNS over pubsub. Implicitly enables pubsub, overrides Ipns.UsePubsub config."),
 		cmds.BoolOption(enableMultiplexKwd, "DEPRECATED"),
 		cmds.StringOption(agentVersionSuffix, "Optional suffix to the AgentVersion presented by `ipfs id` and also advertised through BitSwap."),
 
--- a/cmd/ipfs/daemon.go
+++ b/cmd/ipfs/daemon.go
@@ -365,8 +365,9 @@ func daemonFunc(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment
 	defer repo.Close()
 
 	offline, _ := req.Options[offlineKwd].(bool)
-	ipnsps, _ := req.Options[enableIPNSPubSubKwd].(bool)
-	pubsub, _ := req.Options[enablePubSubKwd].(bool)
+	ipnsps, ipnsPsSet := req.Options[enableIPNSPubSubKwd].(bool)
+	pubsub, psSet := req.Options[enablePubSubKwd].(bool)
+
 	if _, hasMplex := req.Options[enableMultiplexKwd]; hasMplex {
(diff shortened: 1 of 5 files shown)


lidel (changes requested):
Note to self:
- [x] sharness tests both ways of enabling pubsub (cli+config), both pass
- [x] sharness tests both ways of enabling ipns-over-pubsub (cli+config), both pass
- [x] needs to be switched to go-ipfs-config release with https://github.com/ipfs/go-ipfs-config/pull/145 before this is merged

cmd/ipfs/daemon.go
```diff
@@ -372,6 +372,18 @@ func daemonFunc(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment
...
+	if err != nil {
+		return err
+	}
+
+	if cfg.Pubsub.Enabled.WithDefault(false) {
+		pubsub = true
+	}
+	if cfg.Ipns.UsePubsub.WithDefault(false) {
+		ipnsps = true
+	}
```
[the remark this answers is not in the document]
aschmahmann: This doesn't work because you can also pass `ipfs daemon --enable-pubsub-experiment=false` which should override the config, but won't.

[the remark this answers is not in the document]
aschmahmann: pushed an initial fix

cmd/ipfs/daemon.go
```diff
@@ -365,13 +365,26 @@ func daemonFunc(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment
...
+	if err != nil {
+		return err
+	}
+
+	if !psSet {
+		pubsub = cfg.Pubsub.Enabled.WithDefault(false)
+	}
+	if !ipnsPsSet {
+		ipnsps = cfg.Ipns.UsePubsub.WithDefault(false)
+	}
```
aschmahmann, line 386: While this will technically do the job of having the daemon flags override the config there's a bit of a weird behavior here in that if we have `--enable-pubsub-experiment=false` and `Ipns.UsePubSub : true` the result will be that pubsub is running.

@lidel any suggestions, does this seem fine as is?

We should probably add tests here that verify things are working as intended

lidel: @aschmahmann  Added tests at the end of `t0320-pubsub.sh` and `t0183-namesys-pubsub.sh` to confirm that the system stays disabled if cli is `false` and respective config is `true`

lidel (approved):

cmd/ipfs/daemon.go
lidel: Documented `--enable-pubsub-experiment=false` in https://github.com/ipfs/go-ipfs/pull/8510/commits/4637863e9ae7ee013b1476dc114994df2c82f473 

I think this is good enough, realistically users will just use config, as they have been asking for it for years.

cmd/ipfs/daemon.go
```diff
@@ -178,8 +178,8 @@ Headers.
 		cmds.BoolOption(enableGCKwd, "Enable automatic periodic repo garbage collection"),
 		cmds.BoolOption(adjustFDLimitKwd, "Check and raise file descriptor limits if needed").WithDefault(true),
 		cmds.BoolOption(migrateKwd, "If true, assume yes at the migrate prompt. If false, assume no."),
-		cmds.BoolOption(enablePubSubKwd, "Instantiate the ipfs daemon with the experimental pubsub feature enabled."),
-		cmds.BoolOption(enableIPNSPubSubKwd, "Enable IPNS record distribution through pubsub; enables pubsub."),
+		cmds.BoolOption(enablePubSubKwd, "Enable experimental pubsub feature. Overrides Pubsub.Enabled config and other pubsub flags."),
+		cmds.BoolOption(enableIPNSPubSubKwd, "Enable IPNS over pubsub. Implicitly enables pubsub, overrides Ipns.UsePubsub config."),
```
lidel, line 182: @aschmahmann mind :eyes: ?

aschmahmann: > and other pubsub flags

seems a little vague, but perhaps is enough to indicate "setting to false will override Ipns.UsePubsub" without adding too much complexity to the help text.

aschmahmann: I guess it also implies that `--enable-namesys-pubsub=true --enable-pubsub-experiment=false` will fail which isn't true for the config file.

btw the behavior here is technically breaking, even if only for people doing weird things. Worth flagging anywhere?

aschmahmann: ya.... I think we made this worse. Let's just have the daemon flags map 1:1 to the config flags.

cmd/ipfs/daemon.go
aschmahmann: We decided it's fine as is/more confusing to change the behavior

aschmahmann (approved):

PR #8517, opened 2021-10-16
Fixed remote pin warning when in offline mode

Zanda256:
Added function to check if node is in offline mode and returns a bool.
If bool is true, browse the local data store, if CID is present in local datastore, a warning is issued.

### DIFF
--- a/core/commands/pin/remotepin.go
+++ b/core/commands/pin/remotepin.go
@@ -9,6 +9,7 @@ import (
 	"text/tabwriter"
 	"time"
 
+	"net/http"
 	neturl "net/url"
 	gopath "path"
 
@@ -159,10 +160,11 @@ NOTE: a comma-separated notation is supported in CLI for convenience:
 			return err
 		}
 
+		var nameStr string
 		// Prepare Pin.name
 		opts := []pinclient.AddOption{}
 		if name, nameFound := req.Options[pinNameOptionName]; nameFound {
-			nameStr := name.(string)
+			nameStr = name.(string)
 			opts = append(opts, pinclient.PinOpts.WithName(nameStr))
 		}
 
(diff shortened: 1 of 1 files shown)


BigLep:
We'll take a look at this during weekly triage on Friday if not sooner.

core/commands/pin/remotepin.go
```diff
@@ -173,6 +175,34 @@ NOTE: a comma-separated notation is supported in CLI for convenience:
 		if err != nil {
 			return err
 		}
+		//ifconfig.co
+		checkConnection := func() bool {
+			_, err := http.Get("ifconfig.co")
+			return err == nil
+		}
```
lidel, line 182: `checkConnection` needs to be removed.

We can't use a third party DNS name + HTTP service for doing this test.
If it goes offline or can't be reached, our _online_ users will get a false-negative.

Checking `node.IsOnline` is the best we can do without introducing centralized services.

core/commands/pin/remotepin.go
```diff
@@ -173,6 +175,34 @@ NOTE: a comma-separated notation is supported in CLI for convenience:
...
+		offline, _ := req.Options["offline"].(bool)
+
+		if offline || !node.IsOnline || !netConnection {
+			fmt.Println("checking connection")
+			allKeys, err := node.Blockstore.AllKeysChan(ctx)
+			if err != nil {
+				return err
+			}
+			for k := range allKeys {
+				if k == rp.Cid() {
```
lidel, line 194: This is an equivalent of doing `ipfs refs local | grep CID`, which is bit rough :)

There are better ways of checking if a block is in the datastore, for example you could use `ipfs dag stat --offline`:

```
$ ipfs dag stat /ipfs/QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR --offline
Size: 119776, NumBlocks: 1

$ ipfs dag stat /ipfs/QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMns --offline
Error: merkledag: not found
```

Mind replacing this for loop with something like:

```
	_, err := api.Dag().Get(req.Context, rp.Cid())
	if err != nil {
        # getting remote block in offline mode will produce Error: merkledag: not found
		return err
	}
```

lidel (changes requested):
Thank you @Zanda256!
We need to change two things  before this can be merged, see comments inline.

core/commands/pin/remotepin.go
Zanda256: Thank you for the help @lidel , I will make the changes and re-submitt the pr.

lidel:
Thank you for poking at this!
Superseded by https://github.com/ipfs/go-ipfs/pull/8621

Issue #7873, opened 2021-01-20
Gateway needs to log all the requests for better logging


dmellos:
https://discuss.ipfs.io/t/ipfs-logging-configuration/9771/8

I am trying to download the content from the gateway. I do not see any logs for my requests in the ipfs logs.
What if the hash I am using is incorrect?
What if one ipfs node (not having the content on its ow repo) is waiting for an another node to get the data?

I need to see these details in the logs, but there are no logs for gateway requests.

lidel:
Re-posting comment from the forums:
 
> I had a quick look, and it may be that the gateway handlers are not logging much at all. I see no calls to around here `log.Info/Debug(...)`
> 
>  https://github.com/ipfs/go-ipfs/blob/2ed9254426e900cf00a9b35304dc5b5de8173208/core/corehttp/gateway_handler.go#L154
> 
> **I recommend opening a feature request in go-ipfs about better logging of gateway requests [..]
> 
> Your best-shot workaround, however, is to put the IPFS gateway behind an nginx reverse proxy, which will log everything you want and let you control other stuff that IPFS does not allow to customize (max request body size, timeouts etc).

So the action here would be to add some additional Debug lines so one can enable useful gateway logs via `ipfs log level` command.
That is a sensible request in my opinion.

Bare minimum would be to log for each request, a pair:

```
{incoming HTTP request URL} → {actual content path}
```

@dmellos would the above  be enough, or would you be interested in any additional info?

dmellos:
@lidel , I think lets include HTTP response status and request type (GET/POST) along with request URL and the content path, and that should be enough.

sujanasowrirajan:
Hello! I can pick this up if no one is working on it.

Would this log line work?
`{request type} {incoming HTTP request URL} → {actual content path}`

And adding a debug log for all errors in webErrorWithCode to see the reason for failure:
https://github.com/ipfs/go-ipfs/blob/2ed9254426e900cf00a9b35304dc5b5de8173208/core/corehttp/gateway_handler.go#L701

manute:
Hey 👋 ,

I've created this https://github.com/ipfs/go-ipfs/pull/8518 for addressing this issue. 

Thanks!

shohamc1:
This should be closed.

PR #8518, opened 2021-10-17
chore(gateway): debug logging for the http requests

manute:
This PR tries to solve this issue https://github.com/ipfs/go-ipfs/issues/7873 

Please let me know if my approach is ok with the code and the issue.

Thanks!

### DIFF
--- a/core/corehttp/gateway_handler.go
+++ b/core/corehttp/gateway_handler.go
@@ -213,6 +213,7 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
 				break
 			}
 		}
+		logger.Debugw("sub-path", "prefix", prefix)
 	}
 
 	// HostnameOption might have constructed an IPNS/IPFS path using the Host header.
@@ -247,8 +248,8 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
 		}
 
 		redirectURL := gopath.Join("/", prefix, u.Scheme, u.Host, path)
-		http.Redirect(w, r, redirectURL, http.StatusMovedPermanently)
 		logger.Debugw("received an uri param, redirect", "to", redirectURL, "status", http.StatusMovedPermanently)
+		http.Redirect(w, r, redirectURL, http.StatusMovedPermanently)
 		return
 	}
 
@@ -269,7 +270,7 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
 		if prefix == "" && fixupSuperfluousNamespace(w, urlPath, r.URL.RawQuery) {
 			// the error was due to redundant namespace, which we were able to fix
 			// by returning error/redirect page, nothing left to do here
-			logger.Debugw("redundant namespace, fixed by returning error/redirect page")
+			logger.Debugw("redundant namespace, returning error/redirect page")
 			return
 		}
 		// unable to fix path, returning error
@@ -286,6 +287,7 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
 		return
 	default:
 		if i.servePretty404IfPresent(w, r, parsedPath) {
+			logger.Debugw("serve pretty 404 if present")
 			return
 		}
 
@@ -352,6 +354,8 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
 		} else {
 			name = getFilename(urlPath)
 		}
+
+		logger.Debugw("serve file", "name", name)
 		i.serveFile(w, r, name, modtime, f)
 		return
 	}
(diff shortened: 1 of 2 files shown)


lidel (changes requested):
Thank you @manute, this is useful!

Noticed some cosmetic nits to address before we can merge this. Listed them below.


## TODO

- [ ] document `ipfs log level core/server debug` in `/docs/gateway.md`
-  [ ] those `error (type nil)` messages are noisy, are really hard to parse. 
   To illustrate, `curl -H "Host: ipfs.io" 127.0.0.1:8080 ` produces:
   ```
   2021-10-29T22:08:13.646+0200	DEBUG	core/server	corehttp/gateway_handler.go:202	http request received	{"from": "/ipns/ipfs.io/"}
   2021-10-29T22:08:13.647+0200	DEBUG	core/server	corehttp/gateway_handler.go:390	error (type nil) getting file tree for index.html, serve the file	{"from": "/ipns/ipfs.io/", "path": "/ipfs/QmPgJLQHLi5fNLTC2DJABJTa9T7dHy55TvKzatrvQS54B2/index.html"}
   ```
   A simple `serving index.html file`  would be easier to reason about here.
- [ ] `localhost:8080/ipfs/bafybeic7etg7wgzgyhvkyl62iaa7klhxlatjw5uowccvzxxa72aeo5eqsu` → subdomain is not logged at all
- subdomain requests are logged with duplicated CID (once in host, once in path). Opening `http://bafybeic7etg7wgzgyhvkyl62iaa7klhxlatjw5uowccvzxxa72aeo5eqsu.ipfs.localhost:8080/`
produces:
  ```
  2021-10-29T21:57:17.142+0200	DEBUG	core/server	corehttp/gateway_handler.go:202	http request received	{"from": "http://bafybeic7etg7wgzgyhvkyl62iaa7klhxlatjw5uowccvzxxa72aeo5eqsu.ipfs.localhost:8080/ipfs/bafybeic7etg7wgzgyhvkyl62iaa7klhxlatjw5uowccvzxxa72aeo5eqsu/"}
  ```

We will have limited availability during the next week and a half, so if you have time to address the above during that time, would make it easier for us to include this in v0.11.0-rc1

manute:
Hey @lidel, 

Thanks for the review, I have addressed some changes based on that. 
A few questions: 
> redirect from localhost:8080/ipfs/bafybeic7etg7wgzgyhvkyl62iaa7klhxlatjw5uowccvzxxa72aeo5eqsu to the subdomain is not logged at all

Where I can find those subdomain redirects? 


>  subdomain requests are logged with duplicated CID (once in host, once in path)

Do you have any ideas to remove that duplication? Shall I do that before this [line](https://github.com/ipfs/go-ipfs/pull/8518/files#diff-427f4a268b11d614e7c3b3fdf35040496dd19492e325e3b1bd4bdaa4f8bf8e66R201)? And do you have any utility library that parses the path w/o that duplication?

lanzafame:
@manute @Stebalien It may be even better to create a separate logger from the `core/server` logger that logs the requests, i.e. an access log similar to nginx. That way one can just monitor gateway requests without getting internal server errors. And having it as a separate logger will make it easier to filter out with tools such as loki or elasticsearch.

Stebalien:
> @manute @Stebalien It may be even better to create a separate logger from the core/server logger that logs the requests, i.e. an access log similar to nginx. That way one can just monitor gateway requests without getting internal server errors. And having it as a separate logger will make it easier to filter out with tools such as loki or elasticsearch.

That's a good point. Logging to `gateway/requests`, or something like that, would be nice.

> Where I can find those subdomain redirects?

https://github.com/ipfs/go-ipfs/blob/52a747763f6c4e85b33ca051cda9cc4b75c815f9/core/corehttp/hostname.go#L97

Ish. It's an interesting redirect because we return the new path, but _also_ process the request.

Maybe we should just log response information in https://github.com/ipfs/go-ipfs/blob/52a747763f6c4e85b33ca051cda9cc4b75c815f9/core/corehttp/gateway_handler.go#L83? That's probably the best approach and that might just cover everything.

> Do you have any ideas to remove that duplication? Shall I do that before this line? And do you have any utility library that parses the path w/o that duplication?

Hm. Maybe use the `RequestURI` instead? That's the URL that the user _requested_, before modifications, cleaning, etc.

manute:
Hey @Stebalien @lidel , 
I think I've addressed all your feedback. Please, let me know what do you think. 

@lanzafame  re: 
> @manute @Stebalien It may be even better to create a separate logger from the core/server logger that logs the requests, i.e. an access log similar to nginx. That way one can just monitor gateway requests without getting internal server errors. And having it as a separate logger will make it easier to filter out with tools such as loki or elasticsearch.

I think it's a good idea although I'd prefer to do that in another PR along with another issue, as it could have other implications. 
 
Thanks!

core/corehttp/gateway_handler.go
```diff
@@ -83,6 +83,7 @@ func (sw *statusResponseWriter) WriteHeader(code int) {
 	if redirect != "" && code == http.StatusOK {
 		code = http.StatusMovedPermanently
 	}
+	log.Debugw("subdomain redirect", "location", redirect, "status", code)
```
Stebalien, line 86: This will always log a redirect, even if we didn't redirect.

manute: 🤦 fixed, thanks!

manute:
@Stebalien @lidel Hey , does it need more changes or is it all fine?

hyperknot:
Can this be merged in? Otherwise I see no reason to debug gateway configuration issues.

guseggert (approved):

PR #8525, opened 2021-10-22
Issue 7917: Fix/mfs/error pin fail

### LINKED ISSUE #7917: MFS remote pinning policy: log ERROR if pin failed
go-ipfs 0.8.0-rc2 dispatches remote pin request but does not act on its status, so without manual inspection it is possible to miss the fact that MFS pinning (`Pinning.RemoteServices.*.Policies.MFS`, #7913) stopped working.

I believe we should keep checking pin status every `RepinInterval` until its `pinned` or `failed`, and if  it's the latter, print ERROR into `remotepinning/mfs`  logger.  Pinning may fail for multiple reasons, main ones is some arbitrary timeout to find a provider for the data, another one is when you try to pin DAG bigger than allowed quota.

By printing error on MFS pin failure, node operator will be able to notice that remote pinning stopped working due to one of mentioned reasons. 

cc #7559 @aschmahmann does this sound sensible?

gautammenghani:
Fixes #7917 
Approach:
If the status of a pin is "queued", a goroutine will run for each pin until the status of that pin becomes "Pinned" or "Failed"


Mitumaru (approved):

BigLep:
@lidel : I know we have this marked as P3.  I'm putting this on the board for now, but feel free to remove if you don't think you'll realistically get to it during PR Fridays.

cmd/ipfs/pinmfs.go
```diff
@@ -227,6 +225,10 @@ func pinMFS(
 			pinTime = ps.GetCreated().UTC()
 			break
 		}
+		//Run a goroutine to monitor status of queued pins
+		if ps.GetPin().GetCid() == cid && ps.GetStatus() == pinclient.StatusQueued {
+			checkPinStatus(ctx, node, svcName, svcConfig, ps, repinInterval)
+		}
```
gammazero, line 231: I do not think we want to start a new goroutine for each pin. I think it would be better to communicate pins to watch to a single goroutine.

cmd/ipfs/pinmfs.go
```diff
@@ -227,6 +225,10 @@ func pinMFS(
 			pinTime = ps.GetCreated().UTC()
 			break
 		}
+		//Run a goroutine to monitor status of queued pins
+		if ps.GetPin().GetCid() == cid && ps.GetStatus() == pinclient.StatusQueued {
```
gammazero, line 229: The code above already does `if ps.GetPin().GetCid() == cid` so this and the above should be under a that single `if` condition.

cmd/ipfs/pinmfs.go
```diff
@@ -216,6 +213,7 @@ func pinMFS(
 	pinTime := time.Now().UTC()
 	pinStatusMsg := "pinning to %q: received pre-existing %q status for %q (requestid=%q)"
 	for ps := range lsPinCh {
+		log.Errorf("PS Typ eis  is : %T", ps)
```
gammazero, line 216: Should not log error here.

gammazero (changes requested):
There are too many issues with this PR, so converting it to draft.

PR #8538, opened 2021-11-03
fix(sharness): use -Q option instead of pipe to tail cmd

schomatis:


test/sharness/t0040-add-and-cat.sh
```diff
@@ -259,12 +259,12 @@ test_add_cat_file() {
...
     echo "World" > test_current_dir/hello/world &&
     ( cd test_current_dir &&
-    ipfs add -r . | tail -n1 > ../actual && cd ../ ) &&
+    ipfs add -r -Q . > ../actual && cd ../ ) &&
     rm -r test_current_dir
   '
 
   test_expect_success "ipfs add -r . output looks good" '
-    echo "added QmZQWnfcqJ6hNkkPvrY9Q5X39GP3jUnUbAV4AbmbbR3Cb1 test_current_dir" > expected
+    echo "QmZQWnfcqJ6hNkkPvrY9Q5X39GP3jUnUbAV4AbmbbR3Cb1" > expected
```
Stebalien, line 267: I think this test is checking the output of the add command itself (e.g., make sure we actually return the correct filename). I'd either:

1. Test the entire output.
2. Write the output to a file in one operation, then copy the last line to a second file in a second operation.

schomatis: I got the same initial impression but after looking at these tests I think this was just a side effect of us being lazy and not using the correct flag, rather than testing the coherence of directory names being added. That said I have no issue with reverting to the original line; I was just trying to push for the `-Q` flag as much as possible to leave it everywhere I could and make it visible enough to avoid this type of output over-processing in the future.

Stebalien: Hm. You're probably right.

Stebalien (approved):

PR #8545, opened 2021-11-10
Add Homebrew installation method.

xiaoyu2006:
The formula [ipfs](https://formulae.brew.sh/formula/ipfs) is avaliable under Homebrew.


guseggert (approved):
Thanks!

PR #8114, opened 2021-05-07
feat: update UnixFS/MFS with auto sharding

schomatis:
TODO:
- [x] Sharness tests need to be fixed. Left some ideas in https://github.com/ipfs/go-ipfs/pull/8114/commits/b07abacf7883819d82260122bcf3502f4c21816c but feel free to drop and rewrite.
- [ ] Remove default value for HAMT [here](https://github.com/ipfs/go-ipfs/pull/8114#discussion_r728225386) after updating dependencies.
- [ ] Add unsharding work from https://github.com/ipfs/go-ipfs/issues/8148.
- [ ] A couple of interop tests need to be fixed.
- [ ] Remove deprecated expermiental flag `cfg.Experimental.ShardingEnabled`. Optionally maybe also add a new experimental option with the sharding threshold value if we want to move away from the 256 KiB hardcoded now.
   - https://github.com/ipfs/go-ipfs/pull/8527 adds config flag `Internal.UnixFSShardingSizeThreshold` 
UnixFS PR (already merged): https://github.com/ipfs/go-unixfs/pull/91
MFS PR: https://github.com/ipfs/go-mfs/pull/88

Closes https://github.com/ipfs/go-ipfs/issues/8106.
Closes  https://github.com/ipfs/go-ipfs/issues/8148.

PR #8527, opened 2021-10-27
unixfs autosharding with config

aschmahmann:
Related to #8114 and may end up merging onto it.

https://github.com/ipfs/go-ipfs-config/pull/149  adds `Internal.UnixFSShardingSizeThreshold` config flag

Testing:
1. Does the config flag work?
2. Do the old tests work if the sharding threshold is set really low whenever we previously used `Experimental.ShardingEnabled`?

PR #8547, opened 2021-11-12
Add automatic sharding/unsharding tests

schomatis:
Merge sharness tests from https://github.com/ipfs/go-ipfs/pull/8114 into https://github.com/ipfs/go-ipfs/pull/8527.

Closes https://github.com/ipfs/go-ipfs/pull/8114.

### DIFF
--- a/test/sharness/t0250-files-api.sh
+++ b/test/sharness/t0250-files-api.sh
@@ -858,4 +858,70 @@ test_sharding "(cidv1 root)" "--cid-version=1"
 
 test_kill_ipfs_daemon
 
+# Test automatic sharding and unsharding
+
+# We shard based on size with a threshold of 256 KiB (see config file docs)
+# above which directories are sharded.
+#
+# The directory size is estimated as the size of each link. Links are roughly
+# the entry name + the CID byte length (e.g. 34 bytes for a CIDv0). So for
+# entries of length 10 we need 256 KiB / (34 + 10) ~ 6000 entries in the
+# directory to trigger sharding.
+test_expect_success "set up automatic sharding/unsharding data" '
+  mkdir big_dir
+  for i in `seq 5960` # Just above the number of entries that trigger sharding for 256KiB
+  do
+    echo $i > big_dir/`printf "file%06d" $i` # fixed length of 10 chars
+  done
+'
+
+# TODO: This does not need to report an error https://github.com/ipfs/go-ipfs/issues/8088
+test_expect_failure "reset automatic sharding" '
+  ipfs config --json Internal.UnixFSShardingSizeThreshold null
+'
+
+test_launch_ipfs_daemon_without_network
+
+LARGE_SHARDED="QmWfjnRWRvdvYezQWnfbvrvY7JjrpevsE9cato1x76UqGr"
+LARGE_MINUS_5_UNSHAREDED="QmbVxi5zDdzytrjdufUejM92JsWj8wGVmukk6tiPce3p1m"
+
+test_add_large_sharded_dir() {
+  exphash="$1"
+  test_expect_success "ipfs add on directory succeeds" '
+    ipfs add -r -Q big_dir > shardbigdir_out &&
+    echo "$exphash" > shardbigdir_exp &&
+    test_cmp shardbigdir_exp shardbigdir_out
+  '
+
+  test_expect_success "can access a path under the dir" '
+    ipfs cat "$exphash/file000030" > file30_out &&
+    test_cmp big_dir/file000030 file30_out
+  '
+}
+
+test_add_large_sharded_dir "$LARGE_SHARDED"
+
+test_expect_success "remove a few entries from big_dir/ to trigger unsharding" '
+  ipfs files cp /ipfs/"$LARGE_SHARDED" /big_dir &&
+  for i in `seq 5`
+  do
+    ipfs files rm /big_dir/`printf "file%06d" $i`
+  done &&
+  ipfs files stat --hash /big_dir > unshard_dir_hash &&
+  echo "$LARGE_MINUS_5_UNSHAREDED" > unshard_exp &&
+  test_cmp unshard_exp unshard_dir_hash
+'
+
+test_expect_success "add a few entries to big_dir/ to retrigger sharding" '
+  for i in `seq 5`
+  do
+    ipfs files cp /ipfs/"$LARGE_SHARDED"/`printf "file%06d" $i` /big_dir/`printf "file%06d" $i`
+  done &&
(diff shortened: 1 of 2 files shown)


test/sharness/t0250-files-api.sh
```diff
@@ -812,15 +828,15 @@ tests_for_files_api "with-daemon"
 test_kill_ipfs_daemon
 
 test_expect_success "enable sharding in config" '
-  ipfs config --json Internal.UnixFSShardingSizeThreshold "\"1B\""
+  ipfs config --json Experimental.ShardingEnabled true
```
schomatis, line 831: @aschmahmann We need to decide on this.

test/sharness/t0260-sharding.sh
```diff
@@ -129,12 +147,12 @@ test_list_incomplete_dir() {
 
   test_expect_success "can list part of the directory" '
     ipfs ls "$largeSHA3dir" 2> ls_err_out
-    echo "Error: merkledag: not found" > exp_err_out &&
+    echo "Error: failed to fetch all nodes" > exp_err_out &&
```
schomatis, line 150: @aschmahmann Please verify if this change is correct.

test/sharness/t0250-files-api.sh
aschmahmann: ideally I'd like to have a commit with just the new things and `Internal.UnixFSShardingSizeThreshold "\"1B\""` which is passing and then another commit where we drop it and all of the CIDs change and end up as the ones we already have as a way of verifying we've made the changes correctly (we can run the sharness tests locally to verify)

If this is too onerous we can probably leave this as is and I can do some manual verification before we merge

test/sharness/t0260-sharding.sh
aschmahmann: This seems fine. AFAICT it was triggered by switching from Get to GetMany in https://github.com/ipfs/go-unixfs/pull/111 since Get uses https://github.com/ipfs/go-merkledag/blob/bfaa7663d177de206edb8a513b09ab4f94b95e57/merkledag.go#L83 while GetMany uses https://github.com/ipfs/go-merkledag/blob/bfaa7663d177de206edb8a513b09ab4f94b95e57/merkledag.go#L258

schomatis: Makes sense.

test/sharness/t0260-sharding.sh
```diff
@@ -137,4 +155,51 @@ test_list_incomplete_dir() {
 
 test_list_incomplete_dir
 
+# Test automatic sharding and unsharding
```
schomatis, line 158: As discussed informally in chat (but can elaborate more here if needed) any kind of directory manipulation goes through the same "automatic" sharding mechanism, whether it is adding/removing one entry at a time from an MFS tree with `ipfs files add/rm`, or consuming an entire directory at once with `ipfs add -r <big-dir>`.

So above and below this line we are testing the same UnixFS code. This doesn't mean that the current division is superfluous as we're accessing that mechanism through different `go-ipfs` paths but I'm mentioning this to (a) check if this was indeed the intention here and (b) think if we could re-label this boundary.

test/sharness/t0260-sharding.sh
```diff
@@ -8,22 +8,35 @@ test_description="Test directory sharding"
...
   for i in `seq 2000`
   do
     echo $i > testdata/file$i
   done
+
+  mkdir big_dir
+  for i in `seq 5960` # Just above the number of entries that trigger sharding for 256KiB
+  do
+    echo $i > big_dir/`printf "file%06d" $i` # fixed length of 10 chars
+  done
```
schomatis, line 29: The previous `big-dir`/`small-dir` distinction might not have been the most appropriate for these two directories but I'm also not finding a clear distinction between `testdata` and `big-dir`, for example: there is little in `test_add_dir` that tells me which directory I'm working with (now that it has been removed from its arguments) and I need to parse through the code to find it's indeed `testdata`. (If we're keeping the `testdata` name for some kind of background compatibility ignore this comment.)

schomatis (commented):
t0260 LGTM, left minor comments but nothing blocking.

test/sharness/t0260-sharding.sh
aschmahmann: ya, I mostly left it to keep the diff smaller here. These functions could reasonably be rewritten to take the input, output, and debugging directories as arguments, but we don't have to.

test/sharness/t0260-sharding.sh
aschmahmann: True, but in the event the code paths were different we'd be testing different things. i.e. here we are testing that when we remove data from a sharded directory we get an unsharded one and when we add the data back we end up with a sharded one again.

Does that make sense?

schomatis: Yes, it does. I think my main issue here is with the "automatic" boundary, which might imply the previous commands were not, that they didn't use the new UnixFS mechanism and instead were forced or explicitly stated to shard (where in fact we calculated the exact entries number that would go over the threshold and trigger the mechanism). I think your previous explanation is more clear (but still we're in the nit realm so we can ignore this):
> here we are testing that when we remove data from a sharded directory we get an unsharded one and when we add the data back we end up with [the same] sharded one again

test/sharness/t0260-sharding.sh
schomatis: Fair enough, prioritizing a small diff seems sensible with the amount of changes we're plumbing already.

test/sharness/t0260-sharding.sh
```diff
@@ -32,24 +45,29 @@ test_add_large_dir() {
 test_init_ipfs
 
 UNSHARDED="QmavrTrQG4VhoJmantURAYuw3bowq3E2WcvP36NRQDAC1N"
-test_add_large_dir "$UNSHARDED"
+
+test_expect_success "force sharding off" '
+ipfs config --json Internal.UnixFSShardingSizeThreshold "\"1G\""
```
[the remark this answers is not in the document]
aschmahmann: ya that's fair although, I'm not sure how I feel about `0` here. I'm not even sure if when go-ipfs processes the config file it should accept 0 as a value. It restores behavior to the go-ipfs <v0.10.0 defaults, but I'm not sure it's obvious to reason about (i.e.  the feature is turned off which means data is add as a basic directory and no automatic sharding or unsharding of existing MFS directories).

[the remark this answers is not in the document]
schomatis: no problem, just flagging that theoretically at least we expose a "disabled" option in UnixFS, but 1G is more than enough to consider it disabled for all intended purposes

PR #8554, opened 2021-11-17
ci: run examples against latest go-ipfs even in forks

aschmahmann:

PR #8556, opened 2021-11-18
fix: multiple subdomain gateways on same domain

mathew-cf:
Having multiple gateways on the same domain currently results in the subdomain returning 404.

For example, with two gateways (gateway.example.com and dev.gateway.example.com), dev.gateway.example.com would return 404 rather than functioning as a separate gateway.


lidel (approved):
Thanks, this looks sensible: will make it easier for folks to run staging environments.

lidel:
(Rebased on top of https://github.com/ipfs/go-ipfs/pull/8554 which should fix the CI tests)

PR #8522, opened 2021-10-21
WIP: update go-libp2p to v0.16.0

marten-seemann:
This is still a draft, as it references a commit on master (go-libp2p v0.16.0 is not yet released).

PR #8559, opened 2021-11-19
feat: circuit v1 migration prompt when Swarm.EnableRelayHop is set

marten-seemann:
This PR doesn't build, as it depends on #8522. 
Also depends on https://github.com/ipfs/go-ipfs-config/pull/157.

I didn't want to pile on that PR though, it's already chaotic enough..


vyzo (approved):

core/node/groups.go
```diff
@@ -122,6 +122,12 @@ func LibP2P(bcfg *BuildCfg, cfg *config.Config) fx.Option {
 			logger.Error("Use the `Swarm.Transports.Network.Relay' config field instead")
 		}
 	}
+	//nolint
+	if cfg.Swarm.EnableRelayHop {
+		logger.Error("The `Swarm.EnableRelayHop` config field is deprecated.\n" +
```
lidel, line 127: It is noop, so better to be explicit it does nothing :)
```suggestion
		logger.Error("The `Swarm.EnableRelayHop` config field is ignored.\n" +
```

core/node/groups.go
```diff
@@ -122,6 +122,12 @@ func LibP2P(bcfg *BuildCfg, cfg *config.Config) fx.Option {
...
+			"Use `Swarm.RelayService` to configure the circuit v2 relay.\n" +
+			"If you want to continue running a circuit v1 relay, please use the standalone relay daemon: https://github.com/libp2p/go-libp2p-relay-daemon\n")
```
lidel, line 129: @marten-seemann  can relay daemon run both at the same time, or do we also need to specify `Config.RelayV2.Enabled: false` if we want V1?
```suggestion
			"If you want to continue running a circuit v1 relay, please use the standalone relay daemon: https://github.com/libp2p/go-libp2p-relay-daemon (with Config.RelayV1.Enabled: true)\n")
```

core/node/groups.go
```diff
@@ -122,6 +122,12 @@ func LibP2P(bcfg *BuildCfg, cfg *config.Config) fx.Option {
...
+	}
```
lidel, line 130: I believe we should simply refuse to start to avoid people wasting time wondering why relay stopped working:
```suggestion
        os.Exit(1)
	}
```

lidel (changes requested):
LGTM, but I would clarify how to enable RelayV1 in relayd (its disabled right now) and do the hard stop when `EnableRelayHop:true`

core/node/groups.go
vyzo: relayd can run both at the same time.

core/node/groups.go
vyzo: i agree.

lidel (approved):
Thanks, lgtm.
While not ideal, this provides existing users of v1 with an actionable path to move forward if they need to continue running v1 in their infra.

lidel (approved):
Switched to go-ipfs-config v0.18.0, squashing this into `update-libp2p` branch so we have a single PR with libp2p stuff.

PR #8563, opened 2021-11-22
Feat/update datastore

aschmahmann:
This PR ties together #8522 and #8527 since they both rely on updates to go-datastore and therefore need to be merged together.

PR #8564, opened 2021-11-23
[testing] update datastore to use contexts with sharness failure

aschmahmann:
#8563 has a lot going on already, and there's a sharness failure we're only really seeing in CI. Using this PR to debug


aschmahmann:
Not needed anymore. Closed #8563

Issue #8135, opened 2021-05-13
Flaky test: TestPeersTotal


aschmahmann:
https://app.circleci.com/pipelines/github/ipfs/go-ipfs/4610/workflows/d1451a93-9637-482c-b9e1-f1914dfaa687/jobs/51152/tests#failed-test-0

```
=== RUN   TestPeersTotal
2021/05/12 20:33:38 failed to sufficiently increase receive buffer size (was: 208 kiB, wanted: 2048 kiB, got: 416 kiB). See https://github.com/lucas-clemente/quic-go/wiki/UDP-Receive-Buffer-Size for details.
    TestPeersTotal: metrics_test.go:45: expected 1 peers transport, got 2
--- FAIL: TestPeersTotal (0.16s)
```

After re-running it is green, so filing this issue so we don't forget to fix this).

masih:
Looking at [CircleCi config](https://github.com/ipfs/go-ipfs/blob/master/.circleci/config.yml#L92), it looks like `gotest` job runs on [Docker](https://github.com/ipfs/go-ipfs/blob/master/.circleci/config.yml#L34-L42). To set `sysctl` we need to run that on a `machine` executor; See [docs](https://circleci.com/docs/2.0/executor-types/#using-machine).

Running that job on `machine` will have longer start-up time, and will potentially involve installing go manually.  
@aschmahmann Do we want to make this change?

aschmahmann:
@masih the UDP buffer size should be an independent problem from the flaky test, it's just a data transfer efficiency thing.

At the moment it doesn't seem worthwhile to change the setup to run as a machine, especially given that we'd like to move towards using GitHub Actions #8348

schomatis:
The test on which this one is based

https://github.com/ipfs/go-ipfs/blob/ef0428af407410d7f7e98ec0b3e57afd8c828031/core/corehttp/metrics_test.go#L15

has been fixed for the same error in https://github.com/libp2p/go-libp2p-swarm/commit/0538806be79cf321e79253cecaec8c9b0d2a7167.

[...]

PR #8577, opened 2021-12-02
fix(corehttp): adjust peer counting metrics

schomatis:
Following upstream fix in https://github.com/libp2p/go-libp2p-swarm/commit/0538806be79cf321e79253cecaec8c9b0d2a7167.

Fixes https://github.com/ipfs/go-ipfs/issues/8135.

### DIFF
--- a/core/corehttp/metrics_test.go
+++ b/core/corehttp/metrics_test.go
@@ -44,11 +44,13 @@ func TestPeersTotal(t *testing.T) {
 
 	node := &core.IpfsNode{PeerHost: hosts[0]}
 	collector := IpfsNodeCollector{Node: node}
-	actual := collector.PeersTotalValues()
-	if len(actual) != 1 {
-		t.Fatalf("expected 1 peers transport, got %d, transport map %v", len(actual), actual)
+	peersTransport := collector.PeersTotalValues()
+	if len(peersTransport) > 2 {
+		t.Fatalf("expected at most 2 peers transport (tcp and upd/quic), got %d, transport map %v",
+			len(peersTransport), peersTransport)
 	}
-	if actual["/ip4/tcp"] != float64(3) {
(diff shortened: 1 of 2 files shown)


core/corehttp/metrics_test.go
```diff
@@ -44,11 +44,13 @@ func TestPeersTotal(t *testing.T) {
...
-		t.Fatalf("expected 1 peers transport, got %d, transport map %v", len(actual), actual)
+	peersTransport := collector.PeersTotalValues()
+	if len(peersTransport) > 2 {
+		t.Fatalf("expected at most 2 peers transport (tcp and upd/quic), got %d, transport map %v",
+			len(peersTransport), peersTransport)
 	}
-	if actual["/ip4/tcp"] != float64(3) {
-		t.Fatalf("expected 3 peers, got %f", actual["/ip4/tcp"])
+	totalPeers := peersTransport["/ip4/tcp"] + peersTransport["/ip4/udp/quic"]
+	if totalPeers != 3 {
```
petar, line 53: I think this test is fine. However, I'll just note that it still depends on internals of libp2p. This test is creating "contact" to 3 peers, but it does not control what type of connections (and how many) are actually created to each peer (this is libp2p's business). Yet, this test checks for specific types of connections. So, if libp2p were to change and e.g. drop the quic transport, this test will fail again. In any event, I think the test is fine as it is, since it is unlikely libp2p will drop tcp or quic anytime soon.

petar (approved):
lgtm

core/corehttp/metrics_test.go
schomatis: Agreed. I tried to retain the spirit of the test since this metric provides a map that collects transports, not just peer counts, but I'm also fine removing this level of specificity.

PR #8587, opened 2021-12-06
chore: verify checksums during dist.ipfs.io sync

petar:


petar:
@lidel @aschmahmann should we use this PR for good measure? Verifying the checksums prevents us from accepting a corrupt file coming from 'ipfs get'. I imagine the chances of this are low. Happy to go either way. (Otherwise, it looks like @lidel's PR solves the actual issue we had.)

lidel (approved):
lgtm, thank you @petar, this makes things more robust 

- I simplified it a bit by removing maps and switching to Sets  (funny thing: in JS a Set preserves the insertion order,  so can act as replacement for lists) 
- Applied to libp2p-relay-daemon repo already, and tested in https://github.com/libp2p/go-libp2p-relay-daemon/releases/tag/v0.1.0  – worked as expected :+1: , but would be good to get another :+1:   before merging

petar:
@lidel Thanks for testing it!

PR #8588, opened 2021-12-06
fix(ci): dist.ipfs.io sync preserving binaries

lidel:
This is a quick fix for #8494  

- `tar.gz`s got mangled because of reasons known to github and nodejs (had no bandwidth to dig why)
- this does the correct upload every time by reading file as Buffer and forcing github to interpret payload as an opaque stream of bytes of a known length without anything fancy – makes this as future-proof as possible
- I confirmed this fix works in https://github.com/libp2p/go-libp2p-relay-daemon/releases/tag/v0.1.0  (reuploaded `_linux-amd64.tar.gz` was the same as on dist.ipfs.io)


petar (approved):
lgtm

PR #8595, opened 2021-12-08
feat: add basic gateway tracing

guseggert:
##  @lidel's blurb for Release Notes (wip)

> see copy text in [`tracing/doc.go`](https://github.com/ipfs/go-ipfs/blob/4319420743d33bc1b89ac7352b46d615490ac3af/tracing/doc.go)

### Added spans

- The root span for gateway request is named `Rateway.Request`
  - `http.target` has "original" request
  - `ResolvedPath` has immutable IPFS path behind the original request (matters when `/ipns/`)
- Each response format type has its own span:
  - `Gateway.ServeRawBlock`
  - `Gateway.ServeCar`
  - `Gateway.ServeUnixFs`
    - `Gateway.ServeFile`
    - `Gateway.ServeDirectory`

### How to use Jaeger UI

One can use the `jaegertracing/all-in-one` Docker image to run a full Jaeger
stack and configure go-ipfs to publish traces to it (here, in an ephemeral
container):

```console
$ docker run --rm -it --name jaeger \
    -e COLLECTOR_ZIPKIN_HOST_PORT=:9411 \
    -p 5775:5775/udp \
    -p 6831:6831/udp \
    -p 6832:6832/udp \
    -p 5778:5778 \
    -p 16686:16686 \
    -p 14268:14268 \
    -p 14250:14250 \
    -p 9411:9411 \
    jaegertracing/all-in-one
```

Then, in other terminal, start go-ipfs with Jaeger tracing enabled:
```
$ IPFS_TRACING=1 IPFS_TRACING_JAEGER=1 ipfs daemon
```

Finally, the [Jaeger UI](https://github.com/jaegertracing/jaeger-ui#readme) is available at http://localhost:16686

## Demo

| CAR | Block | File | Directory |
| ---- | ---- | ---- | ---- |
|  | block_2022-04-01_01-47 |   |  dir_2022-04-01_01-48 |


cmd/ipfs/main.go
```diff
@@ -71,21 +73,33 @@ func main() {
...
+		return printErr(err)
 	}
+	defer func() {
+		if err := tp.Shutdown(context.Background()); err != nil {
+			exitCode = printErr(err)
+		}
+	}()
+	otel.SetTracerProvider(tp)
+
+	ctx, span := otel.Tracer("").Start(ctx, "Run")
```
whyrusleeping, line 97: a global top level span is generally not super useful, it also breaks jaeger when spans get too big.

core/coreapi/path.go
```diff
@@ -19,6 +22,12 @@ import (
 // ResolveNode resolves the path `p` using Unixfs resolver, gets and returns the
 // resolved Node.
 func (api *CoreAPI) ResolveNode(ctx context.Context, p path.Path) (ipld.Node, error) {
+	ctx, span := otel.Tracer("").Start(
```
whyrusleeping, line 25: feels like you could probably embed the tracer instance inside the CoreAPI struct to save some typing and map lookups

core/coreapi/path.go
```diff
@@ -19,6 +22,12 @@ import (
...
+		ctx,
+		"CoreAPI-ResolveNode",
+		trace.WithAttributes(attribute.String("Path", p.String())),
```
whyrusleeping, line 28: attribute names are usually lower-case by convention

core/corehttp/gateway_handler.go
```diff
@@ -194,6 +196,9 @@ func (i *gatewayHandler) optionsHandler(w http.ResponseWriter, r *http.Request)
 }
 
 func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request) {
+	ctx, span := otel.Tracer("").Start(r.Context(), "gatewayHandler-getOrHeadHandler")
```
whyrusleeping, line 199: no need to have the hypenated name here, the reader should be able to derive where this is coming from based on context within the trace viewer

core/coreapi/path.go
guseggert: yeah I plan on properly plumbing the tracer through before I publish this PR

cmd/ipfs/main.go
guseggert: I thought this might be useful for analyzing CLI commands, can remove if that's not the case

guseggert: also, what's the limit on span size in jaeger?

core/coreapi/unixfs.go
```diff
@@ -55,11 +58,30 @@ func getOrCreateNilNode() (*core.IpfsNode, error) {
...
 func (api *UnixfsAPI) Add(ctx context.Context, files files.Node, opts ...options.UnixfsAddOption) (path.Resolved, error) {
+	ctx, span := tracing.Span(ctx, "CoreAPI.UnixfsAPI", "Add")
+	defer span.End()
+
 	settings, prefix, err := options.UnixfsAddOptions(opts...)
 	if err != nil {
 		return nil, err
 	}
 
+	span.SetAttributes(
```
frrist, line 69: Allocations can be avoided here (and elsewhere when attributes are used) by only setting them when the span is recording. e.g.
```go
if span.IsRecording() {
   // set stuffs
}
```

frrist: Admittedly, this is tedious for some of the one-liners that start a span and set attributes together; I wonder if there is an alternative...

frrist: I wonder if it's worth adding an `Attributes` method to some of these types like `settings,` for example.  It may be more appropriate for a follow-on.

frrist (commented):
Good stuff!

core/coreapi/unixfs.go
guseggert: My thought process here was that we already know we have way bigger fish to fry in terms of memory allocations, so I wasn't too concerned with this.  Hopefully one day we get to the point where these matter...in the meantime, we can enjoy the simpler code

iand (commented):
Looks good and I've been using this branch to study gateway performance characteristics.

tracing/tracing.go
```diff
@@ -0,0 +1,136 @@
...
+			options = append(options, trace.WithBatcher(exporter))
+		}
+	}
+
+	return trace.NewTracerProvider(options...), nil
+}
+
+// Span starts a new span using the standard IPFS tracing conventions.
+func Span(ctx context.Context, componentName string, spanName string, opts ...traceapi.SpanStartOption) (context.Context, traceapi.Span) {
+	return otel.Tracer("go-ipfs").Start(ctx, fmt.Sprintf("%s.%s", componentName, spanName), opts...)
```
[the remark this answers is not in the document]
iand: I think it is intended that the default tracer is used since we aren't changing the format of the traces. We already record the name of the system and the software used to record the trace in the Resource section. See, for example, this section of one trace. The otel documentation is not super clear but I feel that the presence of `Version` and `SchemaURL` fields in the `InstrumentationLibrary` object imply that this is related to the type of trace that has been produced.

```json
    "Resource": [
        {
            "Key": "service.name",
            "Value": {
                "Type": "STRING",
                "Value": "go-ipfs"
            }
        },
        {
            "Key": "service.version",
            "Value": {
                "Type": "STRING",
                "Value": "0.13.0-dev"
            }
        },
        {
            "Key": "telemetry.sdk.language",
            "Value": {
                "Type": "STRING",
                "Value": "go"
            }
        },
        {
            "Key": "telemetry.sdk.name",
            "Value": {
                "Type": "STRING",
                "Value": "opentelemetry"
            }
        },
        {
            "Key": "telemetry.sdk.version",
            "Value": {
                "Type": "STRING",
                "Value": "1.6.1"
            }
        }
    ],
    "InstrumentationLibrary": {
        "Name": "go-ipfs",
        "Version": "",
        "SchemaURL": ""
    }
```

tracing/tracing.go
```diff
@@ -0,0 +1,136 @@
...
+			}
+			options = append(options, trace.WithBatcher(exporter))
+		}
+	}
+
+	return trace.NewTracerProvider(options...), nil
+}
+
+// Span starts a new span using the standard IPFS tracing conventions.
+func Span(ctx context.Context, componentName string, spanName string, opts ...traceapi.SpanStartOption) (context.Context, traceapi.Span) {
```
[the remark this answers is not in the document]
guseggert: Yeah that makes sense, I was waiting to create something like this until I used OpenTelemetry in a few repos to understand our usage patterns better. I'm also hesitant to make a tracing "kitchen sink", since it seems like a place that would accumulate dependencies over time and significantly contribute to our dependency hell.

tracing/doc.go
```diff
@@ -0,0 +1,66 @@
...
+//
+//  In this example the Jaeger UI is available at http://localhost:16686.
+//
+//
+// Implementer Notes
+//
+// Span names follow a convention of <Component>.<Span>, some examples:
+//
+//  - component=Gateway + span=Request -> Gateway.Request
+//  - component=CoreAPI.PinAPI + span=Verify.CheckPin -> CoreAPI.PinAPI.Verify.CheckPin
```
iand, line 63: Can we tighten this convention up? I think we need to define what a component is and what convention we have for naming the span - is it the type name and method, or should it include the package somehow? We'll need this as soon as we start adding tracing to other ipfs libraries.

One convention could be that the component is the repo name `go-ipfs/go-blockservice/go-namesys` and the Span name is the function name, prefixed by the typename if it's a method. Additional spans within a function would use a name derived from this base. The convention would lead to names like:

 - `go-ipfs.PinAPI.Add`
 - `go-blockservice.Session.GetBlock`
 - `go-namesys.ResolveIPNS` 

We could include package name but it would make the names much longer for no great benefit, `go-ipfs.PinAPI.Add` could be ``go-ipfs.core.coreapi.PinAPI.Add` (would also be extra maintenance burden to keep names in sync when refactoring)

This convention is not the prettiest but it would work consistently across all our libraries and gives signposts for developers looking for where a particular span was created.

tracing/tracing.go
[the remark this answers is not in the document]
iand: Fair enough. I don't feel strongly about it. It's possible that the SchemaURL/Version fields are intended to describe the schema of the tags that the library is writing.

tracing/doc.go
```diff
@@ -0,0 +1,66 @@
...
+// including one that writes traces to a JSON file on disk.
+//
+// In general, tracing is configured through environment variables. The IPFS-specific environment variables are:
+//
+//  - IPFS_TRACING: enable tracing in go-ipfs
+//  - IPFS_TRACING_JAEGER: enable the Jaeger exporter
+//  - IPFS_TRACING_RATIO: the ratio of traces to export, defaults to 1 (export everything)
+//  - IPFS_TRACING_FILE: write traces to the given filename
+//  - IPFS_TRACING_OTLP_HTTP: enable the OTLP HTTP exporter
+//  - IPFS_TRACING_OTLP_GRPC: enable the OTLP gRPC exporter
```
lidel, line 18: Mind documenting these in `docs/environment-variables.md` ?

test/sharness/t0310-tracing.sh
```diff
@@ -0,0 +1,57 @@
...
+      receivers: [otlp]
+      processors: [batch]
+      exporters: [file]
+EOF
+
+# touch traces.json and give it 777 perms, in case docker runs as a different user
+rm -rf traces.json && touch traces.json && chmod 777 traces.json
+
+test_expect_success "run opentelemetry collector" '
+  docker run --rm -d -v "$PWD/collector-config.yaml":/etc/otel/config.yaml -v "$PWD":/traces --net=host --name=ipfs-test-otel-collector otel/opentelemetry-collector-contrib
```
lidel, line 42: Should we use an explicit version of the image, to avoid breaking CI when they ship a major release with a breaking change?

lidel (changes requested):
Took it for a spin, works as expected :+1: 

Added more `Gateway.*` spans in https://github.com/ipfs/go-ipfs/pull/8841  (needs to be merged into this PR)  + small asks below.

tracing/tracing.go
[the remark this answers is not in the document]
guseggert: Yeah I'm not sure what those are for. This is all clearly communicated as experimental so that we have some wiggle room to adjust these over time as we learn more.

tracing/doc.go
guseggert: We'd like to protect our ability to refactor repos, modules, and packages without breaking tracing consumers, so I'd like to avoid coupling it to repos and modules as much as possible. Much like we do with logging subsystems (I toyed with the idea of coupling these with logging subsystems but gave up.) The downside though is that it makes it harder to find the code origin of traces, but I still don't think that's very hard with a quick GitHub search. Or we could add some middleware that adds an attribute to all spans with the source file + line info of where the span originated?

lidel:
The only CI test that fails is go-ipfs-http-client, which is unrelated to changes from this PR and being fixed in https://github.com/ipfs/go-ipfs/pull/8838

Resolved conflicts, merging to avoid doing this again.

lidel (approved):
LGTM, we can adjust / add more spans in future PRs.

Issue #8550, opened 2021-11-14
Building sharness tests fails with Go 1.17


aschmahmann:
### Checklist

- [X] This is a bug report, not a question. Ask questions on [discuss.ipfs.io](https://discuss.ipfs.io).
- [X] I have searched on the [issue tracker](https://github.com/ipfs/go-ipfs/issues?q=is%3Aissue) for my bug.
- [X] I am running the latest [go-ipfs version](https://dist.ipfs.io/#go-ipfs) or have an issue updating.

### Installation method

built from source

### Description

From @lidel:

>I cant build sharness deps: got some issues when running make in test/sharness in a freshly checked-out repo:

```
$ cd test/sharness
$ make
(...)
go build -buildmode=plugin -pkgdir "/home/lidel/.local/gopath/pkg/linux_amd64_dynlink" "-asmflags=all='-trimpath=/home/lidel/.local/gopath'" "-gcflags=all='-trimpath=/home/lidel/.local/gopath'" -o "plugin/plugins/git.so" "github.com/ipfs/go-ipfs/plugin/plugins/git/main"
# github.com/klauspost/compress/zstd/internal/xxhash
asm: xxhash_amd64.s:120: when dynamic linking, R15 is clobbered by a global variable access and is used here: 00092 (/home/lidel/.local/gopath/pkg/mod/github.com/klauspost/compress@v1.11.7/zstd/internal/xxhash/xxhash_amd64.s:120)	ADDQ	R15, AX
asm: assembly failed
# github.com/cespare/xxhash/v2
asm: xxhash_amd64.s:120: when dynamic linking, R15 is clobbered by a global variable access and is used here: 00092 (/home/lidel/.local/gopath/pkg/mod/github.com/cespare/xxhash/v2@v2.1.1/xxhash_amd64.s:120)	ADDQR15, AX
asm: assembly failed
make[1]: *** [plugin/plugins/Rules.mk:15: plugin/plugins/git.so] Error 2
make[1]: Leaving directory '/home/lidel/project/ipfs/go-ipfs'
make: *** [GNUmakefile:7: aggregate] Error 2
```

This issue is really https://github.com/cespare/xxhash/issues/54. While we can update xxhash with something like:

```
github.com/cespare/xxhash/v2 v2.1.2 // indirect
github.com/klauspost/compress v1.13.6 // indirect
```

We'll still run into this:

```
[...]

PR #8599, opened 2021-12-09
chore: bump Go version to 1.16.12

aschmahmann:
Waiting on https://github.com/CircleCI-Public/cimg-go/pull/142

### DIFF
--- a/.circleci/main.yml
+++ b/.circleci/main.yml
@@ -36,7 +36,7 @@ default_environment: &default_environment
 executors:
   golang:
     docker:
-      - image: cimg/go:1.16.11
+      - image: cimg/go:1.16.12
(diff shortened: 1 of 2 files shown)


marten-seemann:
Is there any reason we're not using Go 1.17 yet? Go 1.16 will be deprecated in just 2 months.

aschmahmann:
Mostly just timing. We didn't get #8550 fixed in time for 0.11, and wanted to keep v0.11 and 0.12 as similar as possible.

It would have been nice to know we'd get this failure soon after 1.17 came out so we'd have more time to release it, but since we're only running CI against one version at the moment this kind of thing shouldn't be too surprising.

guseggert (approved):

PR #8608, opened 2021-12-14
fix emitting errors on `ipfs rm`

petar:


petar:
Do we also want the errors to include the relevant file in a separate field?

Mitumaru (approved):

guseggert:
@petar, what's the motivation for this change?

petar:
> @petar, what's the motivation for this change?

This change makes errors be emitted as proper json objects (as opposed to free form text), so they can be parsed.
I don't recall the motivation for doing it.

guseggert:
@petar are you okay with closing this then?

Issue #8511, opened 2021-10-13
Update GitHub workflows to use central ipfs install/start Action


BigLep:
**Done criteria:** remove duplication of common steps in our GitHub workflows (e.g., install IPFS, start it, wait for daemon to be running)
**Why important:** general duplication reduction and make future work quicker.
**Customer:** dev team
**Notes:** 
* An example where this was discussed: https://github.com/ipfs/go-ipfs/pull/8494#discussion_r728494937
**TODO**
- [x] https://github.com/ipfs-shipyard/git-remote-ipld/pull/39
- [x] https://github.com/ipfs/go-ds-s3/pull/212
- [x] https://github.com/ipfs/go-ipfs-api/pull/262
- [x] https://github.com/ipfs/go-ipfs-http-client/blob/c832fc0ce1d8235c830fecd0dd44dda04a9fba37/.github/actions/go-test-setup/action.yml (nothing to do here)
- [x] https://github.com/ipfs/go-ipfs/pull/8609
- [x] https://github.com/ipfs/distributions/pull/520 (?)
- [x] https://github.com/libp2p/go-libp2p-relay-daemon/pull/13

PR #8609, opened 2021-12-14
chore: use common ipfs/*-actions in sync-release-assets

galargh:
I want to replace the ipfs build, init and ipfs daemon start logic here with common actions(https://github.com/ipfs/download-ipfs-distribution-action + https://github.com/ipfs/start-ipfs-daemon-action) to minimise duplication. This is part of work related to this issue: https://github.com/ipfs/go-ipfs/issues/8511

The new logic replaces ipfs build logic with downloading latest stable(non rc) release of go-ipfs from dist.ipfs.io.

### DIFF
--- a/.github/workflows/sync-release-assets.yml
+++ b/.github/workflows/sync-release-assets.yml
@@ -13,38 +13,10 @@ jobs:
   sync-github-and-dist-ipfs-io:
     runs-on: "ubuntu-latest"
     steps:
-      - name: Setup go
-        uses: actions/setup-go@v2
+      - uses: ipfs/download-ipfs-distribution-action@v1
+      - uses: ipfs/start-ipfs-daemon-action@v1
         with:
-          go-version: '1.16'
-      - uses: actions/cache@v2
-        with:
-          path: |
-            ~/.cache/go-build
-            ~/go/pkg/mod
-          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
-          restore-keys: |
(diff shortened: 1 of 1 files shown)


.github/workflows/sync-release-assets.yml
```diff
@@ -13,38 +13,10 @@ jobs:
...
-            ${{ runner.os }}-go-
-      - name: Build go-ipfs binary
-        run: go install github.com/ipfs/go-ipfs/cmd/ipfs@latest
-      - name: Initialize go-ipfs and start daemon
-        run: |
-          sudo sysctl -w net.core.rmem_max=2500000
-          ipfs init --profile flatfs,server
-          ipfs daemon --enable-gc=false &
-          while (! ipfs id --api "/ip4/127.0.0.1/tcp/5001"); do sleep 1; done
-      - name: Wait for go-ipfs to be ready
```
lidel, line 36: @galargh Can we update `ipfs/start-ipfs-daemon-action` to include a version of this additional check?
iiuc we want to ensure the  `ipfs id | jq .Addresses`  list is not empty before continuing. I believe we had CI randomly fail due to race condition sometimes.

lidel (approved):
lgtm, thank you @galargh !

.github/workflows/sync-release-assets.yml
galargh: Sure! https://github.com/ipfs/start-ipfs-daemon-action/issues/5

PR #8615, opened 2021-12-16
feat: DNS.MaxCacheTTL for DNS-over-HTTPS resolvers

thibmeu:
Expose `WithMaxCacheTTL` option of go-doh-resolver within IPFS node config.
This allows to cap the Time-To-Live suggested by the DNS response. This includes resolution of DNSLink captured by the custom resolver URLs.

This change relies on [go-ipfs.config](https://github.com/ipfs/go-ipfs-config/pull/161) change. The change of go.mod and go.sum would have to be updated accordingly at this time.

Before merging, both go-doh-resolver and go-ipfs-config have to be updated.

Tagging @vyzo and @marten-seemann , as they reviewed the [go-doh-resolver change](https://github.com/libp2p/go-doh-resolver/pull/12).

### DIFF
--- a/core/node/dns.go
+++ b/core/node/dns.go
@@ -2,7 +2,9 @@ package node
 
 import (
 	"fmt"
+	"math"
 	"strings"
+	"time"
 
 	config "github.com/ipfs/go-ipfs-config"
 	doh "github.com/libp2p/go-doh-resolver"
@@ -16,17 +18,22 @@ var defaultResolvers = map[string]string{
 	"crypto.": "https://resolver.cloudflare-eth.com/dns-query",
 }
 
-func newResolver(url string) (madns.BasicResolver, error) {
+func newResolver(url string, opts ...doh.Option) (madns.BasicResolver, error) {
 	if !strings.HasPrefix(url, "https://") {
 		return nil, fmt.Errorf("invalid resolver url: %s", url)
 	}
 
-	return doh.NewResolver(url), nil
+	return doh.NewResolver(url, opts...), nil
 }
 
 func DNSResolver(cfg *config.Config) (*madns.Resolver, error) {
 	var opts []madns.Option
 	var err error
+
+	var dohOpts []doh.Option
+	if !cfg.DNS.MaxCacheTTL.IsDefault() {
+		dohOpts = append(dohOpts, doh.WithMaxCacheTTL(cfg.DNS.MaxCacheTTL.WithDefault(time.Duration(math.MaxUint32)*time.Second)))
+	}
 
 	domains := make(map[string]struct{})           // to track overridden default resolvers
 	rslvrs := make(map[string]madns.BasicResolver) // to reuse resolvers for the same URL
@@ -44,7 +51,7 @@ func DNSResolver(cfg *config.Config) (*madns.Resolver, error) {
 
 		rslv, ok := rslvrs[url]
 		if !ok {
-			rslv, err = newResolver(url)
+			rslv, err = newResolver(url, dohOpts...)
 			if err != nil {
 				return nil, fmt.Errorf("bad resolver for %s: %w", domain, err)
 			}
--- a/core/node/dns.go
+++ b/core/node/dns.go
@@ -23,7 +23,7 @@ func newResolver(url string, opts ...doh.Option) (madns.BasicResolver, error) {
 		return nil, fmt.Errorf("invalid resolver url: %s", url)
 	}
 
-	return doh.NewResolver(url, opts...), nil
+	return doh.NewResolver(url, opts...)
 }
 
 func DNSResolver(cfg *config.Config) (*madns.Resolver, error) {
--- a/docs/config.md
+++ b/docs/config.md
@@ -140,7 +140,8 @@ config file at runtime.
     - [`Swarm.Transports.Multiplexers.Yamux`](#swarmtransportsmultiplexersyamux)
     - [`Swarm.Transports.Multiplexers.Mplex`](#swarmtransportsmultiplexersmplex)
   - [`DNS`](#dns)
-  - [`DNS.Resolvers`](#dnsresolvers)
+    - [`DNS.Resolvers`](#dnsresolvers)
+    - [`DNS.MaxCacheTTL`](#dnsmaxcachettl)
 
 
 
@@ -1757,7 +1758,7 @@ Type: `priority`
 
 Options for configuring DNS resolution for [DNSLink](https://docs.ipfs.io/concepts/dnslink/) and `/dns*` [Multiaddrs](https://github.com/multiformats/multiaddr/).
 
-## `DNS.Resolvers`
+### `DNS.Resolvers`
 
 Map of [FQDNs](https://en.wikipedia.org/wiki/Fully_qualified_domain_name) to custom resolver URLs.
 
@@ -1793,3 +1794,17 @@ Be mindful that:
 Default: `{}`
 
 Type: `object[string -> string]`
+
+### `DNS.MaxCacheTTL`
+
+Maximum duration entries are valid in the cache.
+
+This allows to cap the Time-To-Live suggested by the DNS response. This includes resolution of DNSLink captured by the custom resolver URLs.
+
+**Examples:**
+* `"5m"` DNS entries are kept for 5 minutes or less.
+* `"0s"` DNS entries expire as soon as they are retrieved.
+
+Default: Respect DNS Response TTL
+
+Type: `optionalDuration`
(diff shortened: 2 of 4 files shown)


go.mod
```diff
@@ -65,7 +65,7 @@ require (
 	github.com/jbenet/go-random v0.0.0-20190219211222-123a90aedc0c
 	github.com/jbenet/go-temp-err-catcher v0.1.0
 	github.com/jbenet/goprocess v0.1.4
-	github.com/libp2p/go-doh-resolver v0.3.1
+	github.com/libp2p/go-doh-resolver v0.0.0-20211210150402-67a10ef63bf3
```
vyzo, line 68: this needs tagged version.

core/node/dns.go
```diff
@@ -16,18 +18,23 @@ var defaultResolvers = map[string]string{
...
+	return doh.NewResolver(url, opts...), nil
 }
 
 func DNSResolver(cfg *config.Config) (*madns.Resolver, error) {
 	var opts []madns.Option
 	var err error
 
+	var dohOpts []doh.Option
+	if !cfg.DNS.MaxCacheTTL.IsDefault() {
+		dohOpts = append(dohOpts, doh.WithMaxCacheTTL(cfg.DNS.MaxCacheTTL.WithDefault(time.Duration(math.MaxUint32)*time.Second)))
```
vyzo, line 35: this will overflow...  do we actually need to set it?

go.mod
thibmeu: it does. The tagged version has not been released yet, therefore the mention to the commit directly, and a reference to this update in the PR message.
Release is tracked in https://github.com/libp2p/go-doh-resolver/issues/15

core/node/dns.go
thibmeu: we need to set `WithDefault` to something. I considered using `0`, since there is an if before and we know the option is set.

`math.MaxUint32` is the default in [go-doh-resolver](https://github.com/libp2p/go-doh-resolver/blob/master/resolver.go#L63). We might want to update if there is an overflow.

vyzo: actually this is fine, duration is 64 bit and int32 max is not that big in that case.

go.mod
vyzo: yeah, just putting an anchor to replace with release tag before merging.

thibmeu: The change is now included in `libp2p/go-doh-resolver@0.4.0`.
go-ipfs-config update is likely to be in `ipfs/go-ipfs-config@0.19.0`.

@aschmahmann do you know if there are other changes to be made to go-ipfs-config for go-ipfs 0.13 release? If not, I can update this commit to have the correct version and hash in `go.mod` and `go.sum`.

lidel (changes requested):
@thibmeu mind elaborating why you want to cap the Time-To-Live suggested by the DNS response? What about being able to set the minimum, or override TTL for IPNS records as well?

FYSA I am looking into adding a proper `cache-control` header to all HTTP responses under  `/ipns/`  namespace (https://github.com/ipfs/go-ipfs/issues/1818#issuecomment-1015849462) – Gateway is not utilizing TTL from DNS at all right now. This means your PR will  not have much impact on gateway responses, but maybe you have a different goal?

My understanding is we want to fix cache control in Gateway responses for `/ipns/`, and need to wait for #1818. After we wire up TTLs, and have both `cache-control` and `etag` in Gateway responses for `/ipns/` content paths, then we can pick this PR up and tweak  `cache-control`  / based on TTL minimum and maximum for both IPNS and DNSLink records. 

I believe we need to have config option for both min and max, because there are people setting records with very small TTL such as 60 seconds, and Gateway operators should be able to override the minimum to be higher, to leverage the cache more.

If we want to control this at the record cache level, then we need four OptionalDurations:
- min/max for DNSLink (and multiaddrs which resolve DNS)
  - `DNS.MaxCacheTTL`
  - `DNS.MinCacheTTL`
- min/max for IPNS records
  - `Ipns.MaxCacheTTL`
  - `Ipns.MinCacheTTL`

Instead, we could introduce min/max specific to Gateway.
It would be applied at the very end, when `cache-control` header is calculated for both DNSLink and IPNS records:
 - `Gateway.MaxCacheTTL`
 - `Gateway.MinCacheTTL`

Of course, `/ipfs/` would remain immutable, would be applied only to mutable responses under `/ipns/` (including DNSLink websites).

I am leaning towards  adding `Gateway.*` ones for now, but lmk your thoughts and use case, perhaps there is a better/simpler way for controlling this?

go.mod
```diff
@@ -31,7 +31,7 @@ require (
 	github.com/ipfs/go-ipfs-blockstore v1.1.2
 	github.com/ipfs/go-ipfs-chunker v0.0.5
 	github.com/ipfs/go-ipfs-cmds v0.6.0
-	github.com/ipfs/go-ipfs-config v0.18.0
+	github.com/ipfs/go-ipfs-config v0.18.1-0.20220120190453-0a6e6d171398
```
lidel, line 34: :information_source:  (note to self) we need to switch to release version before merging

lidel (approved):
Thank you @thibmeu,  in this context it makes sense  to move forward with this PR: it is not directly related to https://github.com/ipfs/go-ipfs/issues/1818 and brings value on its own. 

Pushed some cosmetic changes:
- rebased to fix some CI issues
- switched to release version of go-doh-resolver@v0.4.0 and go-ipfs-config from upstream master (merged https://github.com/ipfs/go-ipfs-config/pull/161)
- tweaked docs:  given that we don't control cache of the default (OS-level) resolver, made it clear that this setting requires DoH
- fixed broken DoH examples while at it


LGTM and ready to merge, but parking until we tag a new release of `go-ipfs-config`.

@aschmahmann are you aware of any other config changes in 0.13 that we should wait for? 
If not, I'd like to tag and merge this to clear up the queue.

aschmahmann:
> @aschmahmann are you aware of any other config changes in 0.13

Yes, there will likely be some related to the next go-libp2p release

>  that we should wait for?

🤷. It doesn't cost us anything to cut more go-ipfs-config releases so we can go ahead and merge + release. I don't know how many people are running builds from master though/if it really makes a difference though. Certainly if you think that we'll have to worry about rebasing the PRs and dealing with conflicts here then definitely just tag + merge.

lidel (approved):
Switched to go-ipfs-config v0.19.0, this is ready for merge.

BigLep:
Next step: review the help text.

docs/config.md
```diff
@@ -1785,11 +1786,28 @@ Be mindful that:
...
   ```
   To get all the benefits of a decentralized naming system we strongly suggest setting DoH endpoint to an empty string and running own decentralized resolver as catch-all one on localhost.
 
 Default: `{}`
 
 Type: `object[string -> string]`
+
+### `DNS.MaxCacheTTL`
+
+Maximum duration entries are valid in the DoH cache.
```
guseggert, line 1800:
```suggestion
Maximum duration that entries are valid in the DoH cache.
```

docs/config.md
```diff
@@ -1785,11 +1786,28 @@ Be mindful that:
...
+
+This allows to cap the Time-To-Live suggested by the DNS response ([RFC2181](https://datatracker.ietf.org/doc/html/rfc2181#section-8)).
```
guseggert, line 1802:
```suggestion
This allows you to cap the Time-To-Live suggested by the DNS response ([RFC2181](https://datatracker.ietf.org/doc/html/rfc2181#section-8)).
```

docs/config.md
```diff
@@ -1785,11 +1786,28 @@ Be mindful that:
...
+If present, the upper bound is applied to DoH resolvers in [`DNS.Resolvers`](#dnsresolvers).
+
+Note: This does NOT work with OS-level resolver. To make this a global setting, add `.` entry to `DNS.Resolvers` first.
```
guseggert, line 1805:
```suggestion
Note: this does NOT work with Go's default DNS resolver. To make this a global setting, add a `.` entry to `DNS.Resolvers` first.
```

docs/config.md
```diff
@@ -1785,11 +1786,28 @@ Be mindful that:
...
+
+**Examples:**
+* `"5m"` DNS entries are kept for 5 minutes or less.
+* `"0s"` DNS entries expire as soon as they are retrieved.
+
+Default: Respect DNS Response TTL
+
+Type: `optionalDuration`
```
guseggert, line 1813: For completeness, can we add this to the `Types` section?

lidel: Good idea, but opened https://github.com/ipfs/go-ipfs/pull/8729 to not block this PR.

PR #8613, opened 2021-12-15
test: restore ipfs-interop

lidel:
Ref. https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs/CHANGELOG.md#0610-2021-12-15

This PR aims to restore interop running using release versions:

- [x] js-ipfs: remove custom build, switch to release from NPM 
- [x]  ipfs-interop: switch to NPM release that includes https://github.com/ipfs/interop/pull/410

PR #8621, opened 2021-12-20
feat: warn user when 'pin remote add' while offline

### LINKED ISSUE #8418: Warn user when 'pin remote add' while offline
## UX problem

> if someone pins something via `pin remote add` but their node is in offline mode, it just hangs forever and fails to pin the thing in a very opaque way

## Potential fix

When in offline mode, we could check if CID is present in local datastore, and if so, show a warning that node is offline and pinning may fail if there is no other provider. 

The check for presence in local datastore is necessary, because we don't want to warn users when they pin CIDs that are not in the local node.

rukolahasser:
Resolves #8418 

The warning is printed only when being online matters (if CID to be pinned remotely is in a local store)

### DIFF
--- a/test/sharness/t0700-remotepin.sh
+++ b/test/sharness/t0700-remotepin.sh
@@ -319,6 +319,14 @@ test_remote_pins() {
 test_remote_pins ""
 
 test_kill_ipfs_daemon
+
+test_expect_success "'ipfs pin remote add' shows the warning message while offline" '
+  test_expect_code 0 ipfs pin remote add --service=test_pin_svc --enc=json $BASE_ARGS --name=name_a $HASH_A > actual &&
+  echo "Warning: the local node is offline and remote pinning may fail if there is no other provider" > expected &&
+  echo "{\"Status\":\"pinned\",\"Cid\":\"$HASH_A\",\"Name\":\"name_a\"}" >> expected &&
+  test_cmp expected actual
+'
+
 test_done
 
 # vim: ts=2 sw=2 sts=2 et:
--- a/test/sharness/t0700-remotepin.sh
+++ b/test/sharness/t0700-remotepin.sh
@@ -320,10 +320,10 @@ test_remote_pins ""
 
 test_kill_ipfs_daemon
 
-WARNINGMESSAGE="Warning: the local node is offline and remote pinning may fail if there is no other provider"
+WARNINGMESSAGE="WARNING: the local node is offline and remote pinning may fail if there is no other provider for this CID"
 
 test_expect_success "'ipfs pin remote add' shows the warning message while offline" '
-  test_expect_code 0 ipfs pin remote add --service=test_pin_svc --enc=json $BASE_ARGS --name=name_a $HASH_A > actual &&
+  test_expect_code 0 ipfs pin remote add --service=test_pin_svc --background $BASE_ARGS --name=name_a  $HASH_A > actual &&
   test_expect_code 0 grep -q "$WARNINGMESSAGE" actual
 '
 
--- a/test/sharness/t0700-remotepin.sh
+++ b/test/sharness/t0700-remotepin.sh
@@ -320,11 +320,11 @@ test_remote_pins ""
 
 test_kill_ipfs_daemon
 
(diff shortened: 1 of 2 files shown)


schomatis (commented):
This seems be the correct way. My main ask at the moment would be to include a test for it in sharness (`t0700-remotepin.sh`), although I'm not sure how easy it is to deal with the situation that the command will still hang forever even after the warning (maybe we should add a timeout).

rukolahasser:
@schomatis How do I set up the `TEST_DOCKER_HOST`? I tried running ipfs in docker locally and set the value to `localhost`, but it failed. Was I missing anything?

schomatis:
Could you give me more context, please? What are you trying to do and why do you need `TEST_DOCKER_HOST`?

rukolahasser:
> This seems be the correct way. My main ask at the moment would be to include a test for it in sharness (`t0700-remotepin.sh`), although I'm not sure how easy it is to deal with the situation that the command will still hang forever even after the warning (maybe we should add a timeout).

I was trying to write a test for it as you mentioned above, and bootstrapped a `rb-pinning-service-api` with docker but the test still failed to run.

schomatis:
I'm still not following completely without seeing the code but the idea is to put the test in sharness (in the file cited before), so Docker shouldn't be necessary for that.

rukolahasser:
Turned out that I didn't have `jq` installed on my machine, now it's fixed. I've added a test and updated the PR, could you have a look please?

rukolahasser:
Does everything look good in the PR? @schomatis

schomatis:
Thanks for the ping, yes. Still we need to wait for lidel to confirm on https://github.com/ipfs/go-ipfs/issues/8418#issuecomment-1001573719 to make sure we if there is anything else needed.

test/sharness/t0700-remotepin.sh
```diff
@@ -319,6 +319,14 @@ test_remote_pins() {
 test_remote_pins ""
 
 test_kill_ipfs_daemon
+
+test_expect_success "'ipfs pin remote add' shows the warning message while offline" '
+  test_expect_code 0 ipfs pin remote add --service=test_pin_svc --enc=json $BASE_ARGS --name=name_a $HASH_A > actual &&
+  echo "Warning: the local node is offline and remote pinning may fail if there is no other provider" > expected &&
+  echo "{\"Status\":\"pinned\",\"Cid\":\"$HASH_A\",\"Name\":\"name_a\"}" >> expected &&
```
schomatis, line 326: We should instead grep for the specific `Warning: the local node is offline` line instead of expecting to match the entire full command JSON output. (There may be a more specialized internal tool than just `grep`, but I'm not familiar with all sharness internals.)

schomatis: Seems we use `grep` in this same test file so it should be ok.

test/sharness/t0700-remotepin.sh
```diff
@@ -319,6 +319,14 @@ test_remote_pins() {
 test_remote_pins ""
 
 test_kill_ipfs_daemon
+
+test_expect_success "'ipfs pin remote add' shows the warning message while offline" '
+  test_expect_code 0 ipfs pin remote add --service=test_pin_svc --enc=json $BASE_ARGS --name=name_a $HASH_A > actual &&
```
schomatis, line 324: I'm assuming here we are pinning something we _and also the service_ have so the command won't hang and still spit the warning which seems the best combination.

rukolahasser: It doesn't hang no matter it succeeds or fails.

lidel (approved):
LGTM, thank you.

- this saves people's time by warning them they forgot to start local provider
- interop test is unrelated, should  be fixed by rebasing on after #8613  lands

lidel:
@rukolahasser mind rebasing this PR on top of the latest `master` ? It should fix the interop tests.

rukolahasser:
@lidel I rebased it but there was a failed test that seemed unrelated to my PR, was the failed test expected?

lidel (approved):
Thank you, this LGTM.

CI issue was unrelated, all green now.

Issue #8481, opened 2021-10-01
ipfs id --offline <peerid> does not work


lidel:
### Checklist

- [X] This is a bug report, not a question. Ask questions on [discuss.ipfs.io](https://discuss.ipfs.io).
- [X] I have searched on the [issue tracker](https://github.com/ipfs/go-ipfs/issues?q=is%3Aissue) for my bug.
- [X] I am running the latest [go-ipfs version](https://dist.ipfs.io/#go-ipfs) or have an issue updating.

### Installation method

built from source

### Version

```Text
0.10.0
```

### Config

```json
(default)
```

### Description

`ipfs id --peerid-base base36 --offline <peerid>` does not work – it requires connection and hangs until one is established.

## Rationale why this should be prioritized 

We would like to have an easy way for converting peerids to a CIDv1 in arbitrary multibase.
Right now, there is no universal way that works for all peerid types and is easy to use.
Example: https://discuss.ipfs.io/t/command-line-to-transform-base58-peerid-to-subdomain-friendly-base36/12330/3?u=lidel

`ipfs id --peerid-base base36 --offline $PEERID  |  jq --raw-output .ID` nearly does the job, but it does not respect `--offline` flag and requires connection to the peer.

schomatis:
Agreed with triage and working on this. It seems to me that if we only need encoding conversion this should live in an independent command maybe. Regardless, I'll implement it here and we can decouple it later if we want.

schomatis:
@lidel There is no `--offline` option in this command (not even in the form of documentation or a fixme, just an error message). My interpretation of it and please correct/complete if wrong is: do not attempt to connect to the peer (no matter if the daemon is running or not) and just print whatever information we have in the peer address book (which will likely be nothing except for the converted peer ID).

schomatis:
(Is the daemon's `--offline` option supposed to apply here? I'm getting a collision when adding a local `--offline` option so I'm assuming there is a global one up in the hierarchy which is the one referred in this issue.)

schomatis:
The global  offline option is in:

https://github.com/ipfs/go-ipfs/blob/91c52657166bcf86f2476926e4fe56694dc26562/core/commands/root.go#L25

PR #8626, opened 2021-12-28
feat(cmds): ipfs id: support --offline option

schomatis:
Closes https://github.com/ipfs/go-ipfs/issues/8481.

### DIFF
--- a/core/commands/id.go
+++ b/core/commands/id.go
@@ -23,14 +23,7 @@ import (
 	identify "github.com/libp2p/go-libp2p/p2p/protocol/identify"
 )
 
-const offlineIdErrorMessage = `'ipfs id' currently cannot query information on remote
-peers without a running daemon; we are working to fix this.
-In the meantime, if you want to query remote peers using 'ipfs id',
-please run the daemon:
-
-    ipfs daemon &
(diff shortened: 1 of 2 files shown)


core/commands/id.go
```diff
@@ -44,6 +37,10 @@ type IdOutput struct {
 const (
 	formatOptionName   = "format"
 	idFormatOptionName = "peerid-base"
+	// FIXME(BLOCKING): The "offline" name might be colliding with the daemon
+	//  option of the same name or a similar option, getting:
+	//  Error: option name "offline" used multiple times
+	offlineOptionName = "id-offline"
```
schomatis, line 43: The option already exists globally in 

https://github.com/ipfs/go-ipfs/blob/91c52657166bcf86f2476926e4fe56694dc26562/core/commands/root.go#L25

I'll reference that and fix this.

lidel: :+1:  on removing `id-offline` and leveraging `offline` that is registered at the root.

lidel (changes requested):

lidel (approved):
LGTM, thank you!

(rebase & tweaked helptext a bit, so it's easier to reason why someone would run `ipfs id` in offline mode)

Issue #8088, opened 2021-04-22
`ipfs config --json <key> false` returns error


TheDiscordian:
#### Version information:
```
go-ipfs version: 0.8.0-ce693d7e8
Repo version: 11
System version: amd64/linux
Golang version: go1.16.2
```

#### Description:
Trying to make a docker image. Realised `ipfs config --json` doesn't handle "false" so well:

```bash
[user@TheDesktop IPFSChat]$ ipfs config --json Swarm.DisableRelay true
[user@TheDesktop IPFSChat]$ ipfs config --json Swarm.DisableRelay false
Error: failed to get config value: "Swarm key has no attributes"
```

I'm assuming this is seeing false as the same as nothing.

Stebalien:
This is because `DisableRelay` is marked "omitempty" and our config system tends to have problems with keys like that.

Note: In this case, it's also deprecated, which is why it's marked "omitempty". Use `ipfs config --json Swarm.Transports.Network.Relay false` instead.

TheDiscordian:
> This is because `DisableRelay` is marked "omitempty" and our config system tends to have problems with keys like that.
> 
> Note: In this case, it's also deprecated, which is why it's marked "omitempty". Use `ipfs config --json Swarm.Transports.Network.Relay false` instead.

Aha thanks! I wasn't aware of the deprecation. I'd imagine this can be closed then? Or perhaps a better error would be better?

Stebalien:
Well, the _error_ is still due to a bug. This command shouldn't actually fail.

aschmahmann:
Renaming this issue to take into account that this is a more general problem related how the config system deals with clearing out "omitempty" fields.

schomatis:
Related to https://github.com/ipfs/go-ipfs/issues/7482.

schomatis:
As noted in the issue cited above, this might be part of a bigger technical debt we need to repay but in the meantime I'll describe here the short-term problem we need to fix.

`omitempty` fields won't be written to JSON output when the value is the default, like `false` for a boolean in the case here (this is why the same command but with a non-default value of `true` does work). In principle this is not an issue in `SetConfigKey` because we don't write the encoded JSON output but a [generic string->any map](https://github.com/ipfs/go-ipfs/blob/91c52657166bcf86f2476926e4fe56694dc26562/repo/fsrepo/fsrepo.go#L618-L622).

[...]

PR #8628, opened 2021-12-29
fix(repo/common): improve MapGetKV not found error

schomatis:


schomatis:
Related to https://github.com/ipfs/go-ipfs/issues/8088# to make clear which is actually the missing key.

BigLep:
@schomatis : is this PR still going to be needed given the work @kylehuntsman is doing in https://github.com/ipfs/go-ipfs/pull/8749 ?

repo/common/common.go
```diff
@@ -21,7 +21,7 @@ func MapGetKV(v map[string]interface{}, key string) (interface{}, error) {
 
 		cursor, ok = mcursor[part]
 		if !ok {
-			return nil, fmt.Errorf("%s key has no attributes", sofar)
+			return nil, fmt.Errorf("%s.%s not found", sofar, part)
```
kylehuntsman, line 24: When it fails to get a root key because it doesn't exist, I get back an odd path.
```
$ ipfs config --json MissingKey.Something
Error: failed to get config value: ".MissingKey not found"
```

I don't know of a small trivial one-liner to avoid writing the period if `sofar` is empty, but something like the following would print "MissingKey not found"
```golang
var path string
if len(sofar) > 0 {
    path += sofar + "."
}
path += part
return nil, fmt.Errorf("%s not found", path)
```
```
$ ipfs config --json MissingKey.Something
Error: failed to get config value: "MissingKey not found"
```

kylehuntsman: An alternative, we can set `sofar` to some root notation if it's empty. Maybe something like `$` or `Root`.
```golang
if len(sofar) == 0 {
    sofar = "$" // Denote that the key that is missing is a root key
}
return nil, fmt.Errorf("%s.%s not found", sofar, part)
```
```
$ ipfs config --json MissingKey.Something
Error: failed to get config value:  "$.MissingKey not found"
```

schomatis:
@kylehuntsman Could you add this patch (or something in the spirit of it) to your PR and then close this, please?

kylehuntsman:
Closing this PR and moving implementation to #8749.

PR #8635, opened 2022-01-03
feat(cmds): Add support for CAR v2 imports

nasdf:
This PR adds support for importing CAR v2 files.  Backwards compatibility with CAR v1 imports should be retained.


go.mod
```diff
@@ -60,6 +60,7 @@ require (
 	github.com/ipfs/interface-go-ipfs-core v0.5.2
 	github.com/ipfs/tar-utils v0.0.2
 	github.com/ipld/go-car v0.3.2
+	github.com/ipld/go-car/v2 v2.1.1 // indirect
```
aschmahmann, line 63: Why is this an indirect import, it seems like we're directly using it? Try running `go mod tidy`.

go.mod
```diff
@@ -94,7 +95,7 @@ require (
 	github.com/multiformats/go-multiaddr v0.4.1
 	github.com/multiformats/go-multiaddr-dns v0.3.1
 	github.com/multiformats/go-multibase v0.0.3
-	github.com/multiformats/go-multicodec v0.3.0
+	github.com/multiformats/go-multicodec v0.3.1-0.20210902112759-1539a079fd61
```
aschmahmann, line 98: We should only be depending on released versions of repos unless there's a documented good reason not to. What's going on here?

core/commands/dag/import.go
```diff
@@ -159,17 +159,12 @@ func importWorker(req *cmds.Request, re cmds.ResponseEmitter, api iface.CoreAPI,
...
 				return err
 			}
 
-			// Be explicit here, until the spec is finished
-			if car.Header.Version != 1 {
-				return errors.New("only car files version 1 supported at present")
-			}
-
-			for _, c := range car.Header.Roots {
+			for _, c := range car.Roots {
```
aschmahmann, line 167: This seems reasonable to me. @ribasushi @willscott any gotchas here I need to be aware of?

aschmahmann (changes requested):
Thanks for the PR and the prompt to update 🙏. Left a couple questions.

We should also add a sharness test with a CARv2 fixture to make sure we don't have regressions here.

go.mod
nasdf: Good catch! Should be fixed now.

core/commands/dag/import.go
willscott: This should the expected thing and read through the roots & blocks of either a v1 or v2 car

go.mod
nasdf: This was coming from the CAR v2 import. Bumped to v0.4.0.

aschmahmann (approved):

aschmahmann (changes requested):
This seems fine. Two things:
1. Could we get a sharness test with a CARv2 fixture to make sure we don't have regressions here?
2. Hopefully next time you push CI should start running, but if not could you double check that you don't have any special settings in your fork or CircleCI for your org? If it's still having trouble lmk and we'll figure out what's going on

nasdf:
I added a simple regression test and instructions on how to create the test files.

CI still seems to be stuck.  I'm not aware of any org specific CircleCI settings but I will double check.

lidel (commented):
@nasdf mind rebasing this PR on top of the latest master branch?

lidel:
@aschmahmann I suspect the CircleCI does not run because @nasdf made a fork within https://github.com/valist-io/ which is an Org and not User, and that means CircleCI needs to be approved within that org to run correctly.

@nasdf are you able to check on your end? If not, I can push your changes into a new branch withing `ipfs/go-ipfs` – lmk.

nasdf:
> @aschmahmann I suspect the CircleCI does not run because @nasdf made a fork within https://github.com/valist-io/ which is an Org and not User, and that means CircleCI needs to be approved within that org to run correctly.
> 
> @nasdf are you able to check on your end and see if CircleCI can be allowed to run there? If not, I can push your changes into a new branch within `ipfs/go-ipfs` to work around this – lmk.

Just double checked and we do have CircleCI approved for our organization.

Would love to get it resolved in the future but you can pull it into a new branch for now.

lidel:
Continued in https://github.com/ipfs/go-ipfs/pull/8854

Issue #8504, opened 2021-10-11
ipfs add --to-files=/path/in/mfs


lidel:
> TLDR: this proposes  `ipfs add --to-files=` as atomic replacement for `ipfs add --pin=false` + `ipfs files cp` 

## Problem

There is a UX gap in our MFS API where users importing data to IPFS via `ipfs add` and `ipfs files write` get different CIDs due to different way chunks are assembled into a DAG (balanced vs trickle). Details in issues linked below.

While we hoped the problem will get solved by "/api/v1" it did not happen, and we are still confusing users just like we did in 2016:

- 2022:
  - This  remains to be a confusing part of IPFS onboarding experience. Example:  ![2022-04-10_15-17](https://user-images.githubusercontent.com/157609/162620073-902a6974-457f-4e90-b90c-dd3f32bf12e4.png)

- 2021: 
  - https://discuss.ipfs.io/t/ipfs-desktop-uses-ipfs-files-write-or-ipfs-add/12412?u=lidel
  - https://discuss.ipfs.io/t/issue-ipfs-add-and-ipfs-files-write-the-difference-between-the-two/5147?u=lidel
- 2018: https://github.com/ipfs/ipfs-webui/issues/676
- 2016: https://github.com/ipfs-inactive/support/issues/45

### Unpacking UX challenges

If a user wants to add data to MFS directly, they need to glue two (or more) commands together:

```
ipfs files cp /ipfs/$(ipfs add -Q test.mp4) /test/test.mp4
```

`ipfs add` by default creates a low-level pin which keeps data around even when the file is removed from MFS. This behavior is usually NOT what MFS users want, so they use `ipfs files write` and get different CID of trickle DAG, and can't benefit from common CID.

One could either do `ipfs add` with `--pin=false` (risky when GC is enabled) or  remove low-level Pin after `files cp` is successful (still problematic, in case the CID was pinned for different reasons)

Either way, asking users to execute two or more commands is not nice. What users want and asked for is essentially a better API, an atomic `ipfs files cp /ipfs/$(ipfs add -Q --pin=false test.mp4) /test/test.mp4` but with GC safety.

## Proposed solution: `ipfs files add` 

Add porcelain command that executes `ipfs add --pin=false && ipfs files cp` in atomic fashion that is safe from GC.

Main characteristics:
- ergonomics similar to `ipfs files write` but without ability to append data – instead, allow recursive import of entire dirs like `ipfs add -r`
- produce the same CID as `ipfs add`
[...]

PR #8637, opened 2022-01-04
ipfs files add

coryschwartz:
fixes: https://github.com/ipfs/go-ipfs/issues/8504

adds command `ipfs files add` using mostly the same parameters as in `ipfs add`

This is functionally equivilent to running `ipfs add` followed by `ifps files cp` with some slightly different defautls.

* IPFS pinning is not enabled by default, although it is an option.
* Keeping with the pattern established for other files command, `-p` is for parent creation. The progress bar is moved to `-P` (capital) instead. Seems unfortunate; if this doesnt make sense, let me know what would be better.

* There is no wrap option. Although, I could be convinced that it should be included here. When adding a directory recursively, files are "wrapped" implicitly by the MFS directory that contains them. 

* Following the convention from cp command, if the destination ends in a forward slash, files are added under an MFS directory such that many files can be added like this:

```
ipfs files add /mfs/path/ file1 file2 file3...
```

The two encantations, although similar, result in a different dag for the containing directory. This is a subtle difference that might be confusing, maybe there is a better way to go about it.

Recursive adds
```
Ξ ~ → ipfs files add -rp /d1 website 
added QmeEte3ZFyKA4yhiQTnBP2x2XTxdDfZzivUiX86xqGT42X website/image.jpg
added QmdXozHCMxfsps9exhLPCzoJq2VeVRkZtuD6KLev7JhfSB website/index.html
added QmfGCw7KQkYGnVbz2h7d3T4TGCpDNza32NqUyoV8zqhWFr website
```

Adding several files with implicit directory creation
```
Ξ ~ → ipfs files add -p /d2/ website/*
added QmeEte3ZFyKA4yhiQTnBP2x2XTxdDfZzivUiX86xqGT42X image.jpg
added QmdXozHCMxfsps9exhLPCzoJq2VeVRkZtuD6KLev7JhfSB index.html
```

Results:
```
Ξ ~ → ipfs files ls -l /
d1/	QmfGCw7KQkYGnVbz2h7d3T4TGCpDNza32NqUyoV8zqhWFr	0
d2/	bafybeih3om227lpu74wyc7zl3f4sddq4atwyv4b3abq6oowgoynkdyygce	0
Ξ ~ → ipfs files ls -l /d1
image.jpg	QmeEte3ZFyKA4yhiQTnBP2x2XTxdDfZzivUiX86xqGT42X	5
index.html	QmdXozHCMxfsps9exhLPCzoJq2VeVRkZtuD6KLev7JhfSB	48
Ξ ~ → ipfs files ls -l /d2
image.jpg	QmeEte3ZFyKA4yhiQTnBP2x2XTxdDfZzivUiX86xqGT42X	5
index.html	QmdXozHCMxfsps9exhLPCzoJq2VeVRkZtuD6KLev7JhfSB	48
```

### DIFF
--- a/core/commands/add.go
+++ b/core/commands/add.go
@@ -21,6 +21,167 @@ import (
 // ErrDepthLimitExceeded indicates that the max depth has been exceeded.
 var ErrDepthLimitExceeded = fmt.Errorf("depth limit exceeded")
 
+var addPostRunMap = cmds.PostRunMap{
+	cmds.CLI: func(res cmds.Response, re cmds.ResponseEmitter) error {
+		sizeChan := make(chan int64, 1)
+		outChan := make(chan interface{})
+		req := res.Request()
+
+		// Could be slow.
+		go func() {
+			size, err := req.Files.Size()
+			if err != nil {
+				log.Warnf("error getting files size: %s", err)
+				// see comment above
+				return
+			}
+
+			sizeChan <- size
+		}()
+
+		progressBar := func(wait chan struct{}) {
+			defer close(wait)
+
+			quiet, _ := req.Options[quietOptionName].(bool)
+			quieter, _ := req.Options[quieterOptionName].(bool)
+			quiet = quiet || quieter
+
+			progress, _ := req.Options[progressOptionName].(bool)
+
+			var bar *pb.ProgressBar
+			if progress {
+				bar = pb.New64(0).SetUnits(pb.U_BYTES)
+				bar.ManualUpdate = true
+				bar.ShowTimeLeft = false
+				bar.ShowPercent = false
+				bar.Output = os.Stderr
+				bar.Start()
+			}
+
+			lastFile := ""
+			lastHash := ""
+			var totalProgress, prevFiles, lastBytes int64
+
+		LOOP:
+			for {
+				select {
+				case out, ok := <-outChan:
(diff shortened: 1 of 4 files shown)


lidel (changes requested):
Thank you @coryschwartz, amazing to see this happening so soon! :heart: 


I took it for a spin and works as expected. Subjectively, it feels like it was always part of the `ipfs files` commands, which is a good sign in my book.

Given that the feature request was written by me, asked for more eyes to review this, mainly to ensure ergonomics feel intuitive to the majority of users from different backgrounds. 

For now, only two actionable asks:

- [ ] rebase on latest master  – this should fix interop tests
- [ ] add a basic sharness test
  - for now just add a file via `ipfs add` and `ipfs files add` with default parameters and confirm they (1) produce the same CID (2) the low-level pin is not created when `ipfs files add` is used

BigLep:
@lidel : it looks like your asks have been addressed.

@schomatis : can you please give this a once over and then merge?  @lidel wanted another opinion on the ergonomics.

schomatis (commented):
This looks like a giant code duplication. If it does what was requested I'm not standing in the way of landing this but I can't approve it, especially since we're duplicating one of the ugliest pieces of the CLI code: `ipfs add`.

Issue #6651, opened 2019-09-19
startup errors are ugly


Stebalien:
Errors produced by `fx` on startup are _really_ ugly. See https://github.com/ipfs/go-ipfs/issues/6650.

We should just strip the rest of the error when reporting it to the user.

schomatis:
Not really sure how to fix this but my dislike for fx errors pushes me to give this a try at least.

PR #8638, opened 2022-01-05
fix(core/node): unwrap fx error in node construction

schomatis:
Not the nicest solution but to extract more information from the errors I had to go the reflection way which added too much complexity for not too much more value.

Closes https://github.com/ipfs/go-ipfs/issues/6651.


Stebalien (approved):
Seems reasonable.

BigLep:
@schomatis : there was approval on this PR.  Can you get CI to pass and then merge?

schomatis:
There is another of those weird js-ipfs `fatal: reference is not a tree` errors but will give it another try (maybe rebase?) and surface it in Discord if I can't make progress.

Issue #7183, opened 2020-04-19
Ipfs daemon hangs when MFS root is not available locally


nlko:
#### Version information:
```
go-ipfs version: 0.4.23-
Repo version: 7
System version: amd64/linux
Golang version: go1.13.7
```
#### Description:
Hello,

When I start the daemon it now hangs after displaying this (I tried this several times and also restarted the computer):
```
$ ipfs daemon
Initializing daemon...
go-ipfs version: 0.4.23-
Repo version: 7
System version: amd64/linux
Golang version: go1.13.7
```

The execution hangs there forever. I tried some ipfs commands that failed returning `Error: cannot acquire lock: Lock FcntlFlock of ~/.ipfs/repo.lock failed: resource temporarily unavailable`

After an hour, if I press Ctrl+C it stops and displays :

```
22:46:53.828 ERROR   cmd/ipfs: error from node construction:  could not build arguments for function "reflect".makeFuncStub (/usr/lib/go/src/reflect/asm_amd64.s:12): failed to build *mfs.Root: function "github.com/ipfs/go-ipfs/core/node".Files (pkg/mod/github.com/ipfs/go-ipfs@v0.4.23/core/node/core.go:74) returned a non-nil error: error loading filesroot from DAG: failed to get block for QmeXwawbVyYrFh6ZjjjbBzzY8gZAcnt7UfLyY2ZhG6MkYQ: context canceled daemon.go:337

Error: could not build arguments for function "reflect".makeFuncStub (/usr/lib/go/src/reflect/asm_amd64.s:12): failed to build *mfs.Root: function "github.com/ipfs/go-ipfs/core/node".Files (pkg/mod/github.com/ipfs/go-ipfs@v0.4.23/core/node/core.go:74) returned a non-nil error: error loading filesroot from DAG: failed to get block for QmeXwawbVyYrFh6ZjjjbBzzY8gZAcnt7UfLyY2ZhG6MkYQ: context canceled
```

Here is how I managed to reach this point:

 - Everything was working fine, the daemon was running for more than 24h and I was doing some normal mfs operation like everyday using ipfs-desktop;
 - The disk where the repo was located ran out of space for a couple of minutes.
 - The reason why the disk ran out of space has nothing to do with ipfs : I was using dd (for a big copy) In the meantime I was watching a video from the mfs.
 - From that point I wasn't able to perform `ipfs add` comands. It failed stating that the disk was full according to a .LOG file. (which is wasn't anymore, more than 180G available).
[...]

PR #8648, opened 2022-01-10
feat(cmds): files: add new-root command to change the MFS root

schomatis:
Some minor fixes pending but it should get a review at this point.

Connected to https://github.com/ipfs/go-ipfs/issues/7183.


nlko:
Hello,

I don't have the necessary knowledge to analyse this PR but since I'm the author of #7183, my question is : will it be possible to use this command without the daemon running ? 

In my case the root MFS corresponding file was missing in the blocks folder at startup and it prevented the daemon to start. Hence the need to have a replace-root working without daemon.

bqv:
> will it be possible to use this command without the daemon running ?

As written now, yes

schomatis:
Correct, thanks.

Right now the idea is for this command to be used when the MFS filesystem is corrupted and the daemon fails to start (either hangs or refuses) and you need to run this standalone to fix it. If you run it alongside the daemon it will fail as it will try to get a lock on the repo which is already in use. Will add some more documentation around that.

core/commands/files.go
```diff
@@ -70,16 +74,17 @@ operations.
...
+		"write":        filesWriteCmd,
+		"mv":           filesMvCmd,
+		"cp":           filesCpCmd,
+		"ls":           filesLsCmd,
+		"mkdir":        filesMkdirCmd,
+		"stat":         filesStatCmd,
+		"rm":           filesRmCmd,
+		"flush":        filesFlushCmd,
+		"chcid":        filesChcidCmd,
+		"replace-root": filesReplaceRoot,
```
lidel, line 87: :thought_balloon:  
since we already use unix terms like `mkdir` `mv` `cp` `rm`, perhaps this one should be named `chroot` ?

schomatis: > since we already use unix terms like mkdir mv cp rm, perhaps this one should be named chroot ?

The current name came from an informal Slack thread, but I'm fine with whatever is decided. My 2 cents on `chroot` is that it might imply some temporary change related to the running daemon/process when in fact we're doing a long-term modification of the single filesystem available (since we don't have any jail/sandbox abstractions).

> another idea: perhaps we could add a second (optional) argument to already existing chcid, which would allow in-place swapping of CIDs for any path (not just root)

This sounds like it is beyond the scope of this issue which is fixing a corrupted filesystem preventing daemon startup (corruption in sub-paths wouldn't have the same effect). The idea is valid on itself though, maybe we should discuss it in another issue as a feature request.

BigLep:
@schomatis : what are the next steps here?

schomatis:
@BigLep As noted in the OP: awaiting review.

bqv:
> @BigLep As noted in the OP: awaiting review.

Perhaps @hsanjuan is busy, @schomatis perhaps you could review in lieu?

schomatis:
No, I can't review my own PR.

bqv:
> No, I can't review my own PR.

Sorry I meant @BigLep!

BigLep:
@schomatis : ACK.  I was confused because the PR was left as draft.  I have marked it as ready for review.

schomatis:
@BigLep Ok, I see the confusion. The 'Ready for review' button is very misleading. It should actually be 'Ready for _merge_'. The draft state is a way to signal a WIP PR to make sure you won't accidentally merge it as is. If the IPFS team interprets drafts differently please let me know and I'll adjust my signalling.

core/commands/files.go
```diff
@@ -1189,6 +1194,111 @@ Remove files or directories.
...
+		ShortDescription: `
+Replace the filesystem root with another CID when the filesystem. Usually used as recovery method if MFS has been corrupted.
+`,
+		LongDescription: `
+Replace the filesystem root with another CID.
+This command is meant to run in standalone mode when the daemon isn't running.
+It is an advanced command that you normally do *not* want to run except when
+the filesystem has been corrupted and the daemon refuses to run.
+
+FIXME: Add an example of the daemon not running once https://github.com/ipfs/go-ipfs/issues/7183
```
lidel, line 1209: TODO

core/commands/files.go
```diff
@@ -1189,6 +1194,111 @@ Remove files or directories.
...
+
+$ ipfs init
+[...]
+ipfs cat /ipfs/QmQPeNsJPyVWPFDVHb77w8G42Fvo15z4bG2X8D2GhfbSXc/readme # <- init dir
+[...] 
+$ ipfs files ls /
+[nothing; empty dir]
+$ ipfs files stat / --hash
+QmUNLLsPACCz1vLxQVkXqqLX5R1X345qqfHbsf67hvA3Nn
+# FIXME(BLOCKING): Need the following to somehow "start" the root dir, otherwise
```
lidel, line 1220: TODO

core/commands/files.go
```diff
@@ -1189,6 +1194,111 @@ Remove files or directories.
...
+security-notes
+$ ipfs files replace-root QmUNLLsPACCz1vLxQVkXqqLX5R1X345qqfHbsf67hvA3Nn # empty dir from init
+$ ipfs files ls /
+[nothing; empty dir]
+`,
+	},
+	Arguments: []cmds.Argument{
+		cmds.StringArg("new-root", true, false, "New root to use."),
+	},
+	// FIXME(BLOCKING): Can/should we do this with the repo running?
```
lidel, line 1243: TODO

core/commands/files.go
```diff
@@ -1189,6 +1194,111 @@ Remove files or directories.
...
+		if len(req.Arguments) < 1 {
+			return fmt.Errorf("new root not provided")
+		}
+		newFilesRootCid, err := cid.Parse(req.Arguments[0])
+		if err != nil {
+			return fmt.Errorf("files root argument provided %s is not a valid CID: %w", req.Arguments[0], err)
+
+		}
+		// FIXME(BLOCKING): Check (a) that this CID exists *locally* and (b)
+		//  that it's a dir.
```
lidel, line 1255: TODO

core/commands/files.go
```diff
@@ -1189,6 +1194,111 @@ Remove files or directories.
...
+		if err != nil {
+			return fmt.Errorf("casting MFS files root %s as CID: %w", filesRootBytes, err)
+		}
+
+		err = localDS.Put(req.Context, node.FilesRootDatastoreKey, newFilesRootCid.Bytes())
+		if err != nil {
+			return fmt.Errorf("storing new files root: %w", err)
+		}
+		// FIXME(BLOCKING): Do we need this if we're closing the repo at the end
+		//  of the command? Likely not.
```
lidel, line 1290: TODO

lidel (commented):
Updated to latest `master` and resolved conflicts. 

Likely ready, just need to 

- [ ] agree on name (maybe `set-root`?)
- [ ] remove / resolve TODOs inline
- [ ] add new command to changelog
- [ ] CI green

core/node/core.go
```diff
@@ -243,7 +245,8 @@ func Files(strategy string) func(mctx helpers.MetricsCtx, lc fx.Lifecycle, repo
 
 		root, err := mfs.NewRoot(ctx, dag, nd, pf, prov)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("failed to initialize MFS root from %s stored at %s: %w. "+
+				"If corrupted, use 'ipfs files chroot' to reset (see --help)", nd.Cid(), FilesRootDatastoreKey, err)
```
lidel, line 249: ℹ️ this will help users like https://github.com/ipfs/kubo/issues/10762 to be able to get useful error message and self-fix the problem

core/commands/files.go
```diff
@@ -1648,3 +1654,153 @@ Examples:
...
+This is a recovery command for when MFS becomes corrupted and prevents the
+daemon from starting.
+
+When run without a CID argument, resets MFS to an empty directory.
+
+WARNING: The old MFS root and its unpinned children will be removed during
+the next garbage collection. Pin the old root first if you want to preserve.
+
+This command can only run when the daemon is not running.
+
```
gammazero, line 1675: Create a short description string and reuse the short description string here and for the short description.
```suggestion
shortDescription = `
'ipfs files chroot' changes the root CID used by MFS (Mutable File System).
This is a recovery command for when MFS becomes corrupted and prevents the
daemon from starting.

When run without a CID argument, resets MFS to an empty directory.

WARNING: The old MFS root and its unpinned children will be removed during
the next garbage collection. Pin the old root first if you want to preserve.

This command can only run when the daemon is not running.
`
var filesChrootCmd = &cmds.Command{
	Status: cmds.Experimental,
	Helptext: cmds.HelpText{
		Tagline: "Change the MFS root CID.",
		ShortDescription: shortDescription,
		LongDescription: shortDescription + `

```

lidel: I think most of commands only use Short one for everything, I'll do that.

PR #8661, opened 2022-01-11
fix(core): look for MFS root in local repo only

schomatis:
Closes https://github.com/ipfs/go-ipfs/issues/7183. (The bigger issue of GC/MFS lock is being [worked on](https://github.com/ipfs/go-ipfs/issues/6113) independently. We close the issue about *hanging* on startup.)


BigLep:
@aschmahmann will approve/merge

aschmahmann (approved):
Looks good, we should have a small sharness test that confirms that this works.

schomatis:
I tested this locally with the new command in https://github.com/ipfs/go-ipfs/pull/8648. That seems like the most organic way to add a sharness test for this so will mark it as blocked on that other one (feel free to suggest an alternative test that would keep both independent).

PR #8668, opened 2022-01-12
feat(cmds/files/ls): recursive listing

schomatis:
Rough prototype for review. Sharness tests missing.

Example:
```
$ ipfs init
$ ipfs files mkdir -p /1/2/3/4

$ ipfs files ls -R 
/:
1

/1:
2

/1/2:
3

/1/2/3:
4

/1/2/3/4:
```

### DIFF
--- a/core/commands/files.go
+++ b/core/commands/files.go
@@ -437,14 +437,11 @@ func getNodeFromPath(ctx context.Context, node *core.IpfsNode, api iface.CoreAPI
 
 type filesLsOutput struct {
 	Entries []mfs.NodeListing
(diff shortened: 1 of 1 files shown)


core/commands/files.go
```diff
@@ -567,6 +632,9 @@ Examples:
 					fmt.Fprintf(w, "%s\n", o.Name)
 				}
 			}
+			if out.Path != "" {
+				fmt.Fprint(w, "\n")
```
schomatis, line 636: Ideally we should skip this for the last entry we list. What would be the cleanest way to signal that? Closing the response?

PR #8674, opened 2022-01-13
fix(ci): interop test parallelism

lidel:
## Main fix

`parallelism: 4` on circleci broke when require got replaced with modern import statement in [ipfs-interop//test/node.js](https://github.com/ipfs/interop/blob/v8.0.0/test/node.js)
Since then, CI run all tests four times.. every time, taking more than it should :upside_down_face: 

This fix also makes the split smarter – based on timing of previous runs: https://circleci.com/docs/2.0/parallelism-faster-jobs/#splitting-by-timing-data which should shave some time, if some files take way longer than others.

##  TODO

- [x] fix parallelism (run 1/4  instead of 4x everything)
- [x] fix ipfs-webui (playwright install was missing)
- [x] NPM cache for faster install
- [x] fix Noise issues  (main suspect: https://github.com/ChainSafe/js-libp2p-noise/pull/129)

### DIFF
--- a/.circleci/main.yml
+++ b/.circleci/main.yml
@@ -219,7 +219,7 @@ jobs:
     docker:
       - image: cimg/node:16.13.2
     parallelism: 4
-    resource_class: 2xlarge
+    resource_class: large
     steps:
       - *make_out_dirs
       - attach_workspace:
@@ -336,6 +336,7 @@ jobs:
           name: Installing dependencies
           command: |
             npm install
+            npx playwright install
           working_directory: ~/ipfs/go-ipfs/ipfs-webui
       - run:
           name: Running upstream tests (finish early if they fail)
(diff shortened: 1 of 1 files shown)


.circleci/main.yml
```diff
@@ -217,12 +217,18 @@ jobs:
       - *store_gomod
   interop:
     docker:
-      - image: cimg/go:1.16-node
```
aschmahmann, line 220: Are you sure we don't need Go? IIRC we needed it in order to build custom Go binaries like the relay daemon or go-ipfs.

lidel: Affirmative: binary (`IPFS_GO_EXEC: /tmp/circleci-workspace/bin/ipfs`) is built in `build` step:
https://github.com/ipfs/go-ipfs/blob/467fb0af8c460af76b2c54fe0f871576f1252ea7/.circleci/main.yml#L382-L384

lidel: But to make it easier to work with custom binaries, restored use of `cimg/go:1.16-node`

aschmahmann (approved):

PR #8677, opened 2022-01-14
fix(build): Recognize Go Beta versions in makefile

NatoBoram:
* Closes https://github.com/ipfs/go-ipfs/issues/8672

This allows using `make install` with beta versions of Golang.

```yaml
go-ipfs version: 0.13.0-dev-653dbff13-dirty
Repo version: 12
System version: amd64/linux
Golang version: devel go1.18-becaeea119 Tue Dec 14 17:43:51 2021 +0000
```


NatoBoram:
Bump. I proposed this change two weeks ago. If someone's available, that'd be nice :)

It's not an important change, but one that's useful if you want to keep up with Go development while still contributing to `go-ipfs`.

guseggert (approved):

PR #8679, opened 2022-01-14
Serve subdomain requests on path gateways

mathew-cf:
Enables serving origin isolated subdomains on path gateways (without redirects). This allows a smoother transition to subdomain gateways since they can be tested prior to redirecting all path gateway traffic.

### DIFF
--- a/core/corehttp/hostname.go
+++ b/core/corehttp/hostname.go
@@ -132,8 +132,8 @@ func HostnameOption() ServeOption {
 				// Assemble original path prefix.
 				pathPrefix := "/" + ns + "/" + rootID
 
-				// Does this gateway _handle_ subdomains AND this path?
-				if !(gw.UseSubdomains && hasPrefix(pathPrefix, gw.Paths...)) {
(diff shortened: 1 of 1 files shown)


lidel (changes requested):
@mathew-cf if  iiuc, doing this would  open a surface for mulscofiguration – people running path gateway thinking its fine because "subdomains" work.

This defeats the whole purpose of subdomain gateways – providing Origin isolation (one could load JS using the path gateway and have access to cookies/storage on all subdomains).

Perhaps you can elaborate on your use case? 
- If you want to test IPFS resolution before running it in production, you could send curl to `127.0.0.1:8080` to skip everything related to DNS setup
- if you want to test subdomain gateway, it has to be a tested with the specific DNS name and  Origin, afraid there is no shortcut what is safe

mathew-cf:
I think you understand the use case. The use case here is just to have both a subdomain gateway and a path gateway available on the same hostname (in prod) for testing prior to the transition to using redirects.

PR #8680, opened 2022-01-15
feat: opt-in Swarm.ResourceMgr (go-libp2p v0.18)

### LINKED ISSUE #1482: Resource Constraints + Limits
We need a number of configurable resource limits. This issue will serve as a meta-issue to track them all and discuss a consistent way to configure/handle them.

I'm going to use a notation like `thingA.subthingB.subthingC`. we dont have to keep this at all, just helps us bind scoped names to things. (using `.` instead of `/` as the `.` _could_ reflect json hierarchy in the config, but it may not have to (e.g. `repo.storage_max` and `repo.datastore.storage_gc_watermark` could be in config as `Repo.StorageMax` and `Repo.StorageGC`, or something.).
### Possible Limits

This is a list of possible limits. I don't think we need all of them, as other tools could limit this more, particularly in server scenarios. but please keep in mind that some users/use cases of ipfs demand that we have some limits in place ourselves, as many end users cannot be expected to even know what a Terminal is (e.g. if they run ipfs as an elecron-app or as a browser extension).
- [ ] `node.repo.storage_max`: this affects the physical storage that a repo takes up. this must include _all_ the storage, datastore + config file size (ok to pre-allocate more if neeeded), so that people can set a maximum. (**MUST** be user configurable) #972
  - [ ] `node.repo.datastore.storage_max`: hard limit on datastore storage size. could be computed as `repo.storage_max - configsize` where `configsize` could be live, or could be a reasonable bound. #972
  - [x] `node.repo.datastore.storage_gc_watermark`: soft limit on datastore storage size. after passing this threshold, automatically run gc. could be computed as `node.repo.datastore.storage_max - 1MB` or something. #972
- [ ] `node.network_bandwidth_max`: limit on network bandwidth used.
  - [ ] `node.gateway.bandwidth_max`: limit on bandwidth allocated to running the gateway. this could be calculated from `node.network_bandwidth_max - all other bandwidth use`. #1070
  - [ ] `node.swarm.bandwidth_max`: limit on network bandiwdth allocated to running the ipfs protocol. this could be calculated from `node.network_bandwidth_max - all other bandwidth use`.
  - [ ] `node.dht.bandwidth_max`: limit on network bandwidth allocated to running the dht protocol. this could be calculated from `node.network_bandwidth_max - all other bandwidth use`.
  - [ ] `node.bitswap.bandwidth_max`: limit on network bandwidth allocated to running the bitswap protocol. this could be calculated from `node.network_bandwidth_max - all other bandwidth use`.
- [ ] `node.swarm.connections`: soft limit on ipfs protocol network connections to make. the reason for this limit is that there is overhead to every connections kept alive. the node could try to stay within this limit.
- [ ] `node.gateway.ratelimit`: a number of requests per second. with this limit, the user could reduce the accept load on the gateway. #1070
- [ ] `node.memlimit`: a limit on the memory allocated to ipfs. could try to use smaller buffers if under different constraints. this is hard to do, prob wont be used end-user-side, and likely easier to do with tools around it sysadmin-side (docker, etc).

**note on config**: the above keys need not be the config keys, but we should figure out some keys that make sense hierarchically.

### LINKED ISSUE #8722: create a command for resource manager stats
## What?

go-libp2p v0.18.0 adds support for the resource manager. The resource manager is already initialized in https://github.com/ipfs/go-ipfs/pull/8680, and will limit libp2p's resource usage. Optionally, the user can define new limits using a `limits.json` file.

What is missing here is visibility into the running system. Once we roll this out, we might get complaints from users that connections / streams are getting blocked by the rcmgr. This can either be due to too low limits (for their use case), or due to a bug in the implementation (e.g. resources not properly released).

Prior Art: https://github.com/filecoin-project/lotus/pull/7969. Lotus added a `netstat` command, which can show the current usage of the different scopes. We should probably have something similar in IPFS.

## How?

1. We need to capture and store the pointer to the libp2p resource manager. Currently, `fx` only initializes it and passes the option to libp2p. We need the pointer to call the various `Stat` methods.
1. `ipfs net stat <scope>` (I'll let others decide what exactly the name of the command should be). The `<scope>` should inspect the current resource usage for a scope. The scope can be `all`, `system`, `transient`, `svc:<service-name>`, `proto:<protocol-name>`, or `peer:<peer-name>`. Implementation-wise, we can probably copy-paste a lot of code from https://github.com/filecoin-project/lotus/pull/7969/files#diff-b51b680e4d6352878815055582b90d38460108021bfb6d2ac9ce297b4a0d2a5e.
1. Bonus points if we can also properly `Close` the resource manager on shutdown.

## Notes
This issue is part of the larger effort of integration go-libp2p v0.18.0 into go-ipfs: https://github.com/ipfs/go-ipfs/issues/8761

marten-seemann:
Part of #8761

- go-libp2p v0.18.0 completes the deprecation of ipfs/go-log@v1 and migrates all remaining repos to ipfs/go-log@v2
- Adds commands and configuration for [libp2p Network Resource Manager](https://github.com/libp2p/go-libp2p-resource-manager#readme) 
  - `ipfs swarm limit --help`
  - `ipfs swarm stats --help`
- Adds config at `Swarm.ResourceMgr` 
  - `Swarm.ResourceMgr.Enabled` is a flag, disabled by default

Closes #8722 
Closes #1482 :mage_man: 
 
- [x] https://github.com/ipfs/go-ipfs-api/pull/263.
- [x] Decide where to put these commands
  -  done in  https://github.com/ipfs/go-ipfs/pull/8735 which got merged into this PR
- [x] Make command reading config from a file  work over HTTP RPC
- [x] final cleanup by @lidel  – see https://github.com/ipfs/go-ipfs/pull/8680#issuecomment-1086432509
  - [x] Add opt-in env or config flag to control resource manager  instantiation, until we are comfortable with production testing to enable by default (similar to https://github.com/filecoin-project/lotus/pull/8318) 
  - [x] Decide where is the limit.json config file going to live: hardcoded in the repo root or configured through go-ipfs-config.


guseggert:
Do you have plans to publish metrics when a RM req is rejected due to exceeded scope limits? That would be very useful operationally to help know when to scale gateway fleets.

marten-seemann:
The resource manager has a tracer integration (see https://github.com/libp2p/go-libp2p-resource-manager/blob/master/trace.go), so it would be possible to hook this up with Prometheus in one way or the other.

vyzo:
travis asked for the same thing so i will add metrics support to the
reource manager.

lidel:
"exchange files" → "cat file"  interop test always gets stuck after   `go -> go2: 8.39 MB (125ms)` 
(was able to reproduce locally with this PRs binary)

Quick way to reproduce/debug is to run specific test with:
`export IPFS_GO_EXEC=/path/to/ipfs-binary-from-libp2p-v018-branch/ipfs`
`npx ipfs-interop -- -t node --grep "go -> go2: 8.39 MB` → works
`npx ipfs-interop -- -t node --grep "go -> go2: 67.1 MB"` → hangs

lidel:
```
DEBUG="ipfsd-ctl:daemon*" IPFS_LOGGING="DEBUG"  npx ipfs-interop -- -t node --grep "go -> go2: 67.1 MB"
```

Revealed connection being dropped by go-libp2p-resource-manager:

```
2022-02-24T02:49:11.848+0100    DEBUG   rcmgr   go-libp2p-resource-manager@v0.1.3/scope.go:239  blocked memory reservation      {"scope": "peer:12D3KooWReFVRYtypujVVXkvictuB8PyaZnSraLEDvwQPGSDxMaL", "size": 258126, "priority": 128, "error": "resource limit exceeded"}
2022-02-24T02:49:11.849+0100    DEBUG   dht     net/message_manager.go:303      error reading message   {"error": "stream reset", "retrying": true}
2022-02-24T02:49:11.849+0100    DEBUG   swarm2  go-libp2p-swarm@v0.10.2/swarm.go:336    [12D3KooWKyDk1EymJbcr59vVjJeJUvwe6vdAUKXnRtbNVt63pvow] opening stream to peer [12D3KooWReFVRYtypujVVXkvictuB8PyaZnSraLEDvwQPGSDxMaL]
2022-02-24T02:49:11.849+0100    DEBUG   swarm2  go-libp2p-swarm@v0.10.2/swarm_dial.go:241       dialing peer    {"from": "12D3KooWKyDk1EymJbcr59vVjJeJUvwe6vdAUKXnRtbNVt63pvow", "to": "12D3KooWReFVRYtypujVVXkvictuB8PyaZnSraLEDvwQPGSDxMaL"}
2022-02-24T02:49:11.849+0100    DEBUG   swarm2  go-libp2p-swarm@v0.10.2/limiter.go:193  [limiter] adding a dial job through limiter: /ip4/127.0.0.1/tcp/44017
2022-02-24T02:49:11.849+0100    DEBUG   swarm2  go-libp2p-swarm@v0.10.2/limiter.go:161  [limiter] taking FD token: peer: 12D3KooWReFVRYtypujVVXkvictuB8PyaZnSraLEDvwQPGSDxMaL; addr: /ip4/127.0.0.1/tcp/44017; prev consuming: 0
2022-02-24T02:49:11.849+0100    DEBUG   swarm2  go-libp2p-swarm@v0.10.2/limiter.go:167  [limiter] executing dial; peer: 12D3KooWReFVRYtypujVVXkvictuB8PyaZnSraLEDvwQPGSDxMaL; addr: /ip4/127.0.0.1/tcp/44017; FD consuming: 1; waiting: 0
2022-02-24T02:49:11.849+0100    DEBUG   swarm2  go-libp2p-swarm@v0.10.2/swarm_dial.go:386       12D3KooWKyDk1EymJbcr59vVjJeJUvwe6vdAUKXnRtbNVt63pvow swarm dialing 12D3KooWReFVRYtypujVVXkvictuB8PyaZnSraLEDvwQPGSDxMaL /ip4/127.0.0.1/tcp/44017
2022-02-24T02:49:11.849+0100    INFO    bitswap go-bitswap@v0.5.1/bitswap.go:638        Bitswap ReceiveError: stream reset
2022-02-24T02:49:11.849+0100    DEBUG   bitswap_network network/ipfs_impl.go:413        bitswap net handleNewStream from 12D3KooWReFVRYtypujVVXkvictuB8PyaZnSraLEDvwQPGSDxMaL error: stream reset
2022-02-24T02:49:11.849+0100    INFO    bitswap go-bitswap@v0.5.1/bitswap.go:638        Bitswap ReceiveError: stream reset
2022-02-24T02:49:11.849+0100    DEBUG   bitswap_network network/ipfs_impl.go:413        bitswap net handleNewStream from 12D3KooWReFVRYtypujVVXkvictuB8PyaZnSraLEDvwQPGSDxMaL error: stream reset
2022-02-24T02:49:11.849+0100    INFO    connmgr connmgr/connmgr.go:546  tried to remove tag from untracked peer: 12D3KooWReFVRYtypujVVXkvictuB8PyaZnSraLEDvwQPGSDxMaL
2022-02-24T02:49:11.849+0100    DEBUG   bs:sprmgr       sessionpeermanager/sessionpeermanager.go:98     Bitswap: removed peer from session      {"session": 1, "peer": "12D3KooWReFVRYtypujVVXkvictuB8PyaZnSraLEDvwQPGSDxMaL", "peerCount": 0}
```

marten-seemann:
> Revealed connection being dropped by go-libp2p-resource-manager

There's no connection being dropped. The resource manager is blocking a memory with prio 128, which is probably a yamux stream window increase: https://github.com/libp2p/go-yamux/blob/cd22a37b789cf6e1fe4f2c2cff6da5df7a49dfac/stream.go#L229. This won't cause any problems.

Do we have logs from the other node?

lidel:
These logs `ipfsd-ctl:daemon` are from both ends, which make it difficult to debug. 

I decided to remove `ipfs-interop` from the picture and was still able to replicate the issue without it, using two nodes on same machine:

## Node A (source)

One terminal: 
```console
$ export IPFS_PATH=/tmp/node-a
$ ipfs daemon --init --init-profile server,randomports,test
```

Second terminal:
```console
$ ipfs id | jq '.Addresses[0]' 
{addrA}
$ ipfs swarm connect  {addrB}
$ head -c 65M /dev/urandom | ipfs add -Q
{cid-OK}
$ head -c 8M /dev/urandom | ipfs add -Q
{cid-FAIL}
```

## Node B (destination)


One terminal: 
```console
$ export IPFS_PATH=/tmp/node-b
$ ipfs daemon --init --init-profile server,randomports,test
```

Second terminal:
```console
$ ipfs id | jq '.Addresses[0]' 
{addrB}
$ ipfs swarm connect  {addrA}
$ ipfs cat {cid-OK} > /tmp/8M # finished instantly
$ ipfs cat {cid-FAIL} > /tmp/65M # stops in the middle (around 20-50%)
30.50 MiB / 65.00 MiB [==================================>--------------------------------------]  46.92%
```

aschmahmann:
@marten-seemann @lidel I'm not totally sure, but am wondering if we're running into a combination of go-bitswap not gracefully handling the lack of streams and go-bitswap using more streams than it should (https://github.com/ipfs/go-bitswap/issues/548).

BigLep:
2022-03-03 conversation:
1. @BigLep : create issue for the overall user story to link to
2. Making config changes without recompiling will likely be key here, even for the initial gateway testing for validating go-libp2p 0.18.

marten-seemann:
@schomatis I merged your PR earlier today, and it looks like the limit setting command is not working (again?):
```
❯ ipfs swarm limit system -s ./limit.json
❯ ipfs swarm limit system
{"Conns":0,"ConnsInbound":0,"ConnsOutbound":0,"FD":0,"Streams":0,"StreamsInbound":0,"StreamsOutbound":0}
```

Any idea what's going on here?

core/node/libp2p/rcmgr.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+func ResourceManager() func(fx.Lifecycle, repo.Repo) (network.ResourceManager, Libp2pOpts, error) {
+	return func(lc fx.Lifecycle, repo repo.Repo) (network.ResourceManager, Libp2pOpts, error) {
+		var limiter *rcmgr.BasicLimiter
+		var opts Libp2pOpts
+
+		// FIXME(BLOCKING): Decide how is the `limit.json` file path going to be consumed,
+		//  either by default in the repo root or through the `go-ipfs-config`.
+		limiter = rcmgr.NewDefaultLimiter()
```
marten-seemann, line 20: @schomatis It should read the file from the directory from where `ipfs daemon` is run. There shouldn't be any config option. Could you please implement this?

schomatis:
Removing my assignment as discussed in [Discord](https://discord.com/channels/806902334369824788/942673321852563456/951177875153895424). We'll wait for @aschmahmann's input if we want to place the limit file somewhere else.

marten-seemann:
No strong opinion on that, but the rcmgr options should not be considered stable at this point.

schomatis:
I also have no useful opinion on this so will wait for a confirmed direction before doing anything else here.

BigLep:
@schomatis : Datapoints that would be helpful for driving the decision: Is there any precedent for storing configurable values (even experimental / non final ones) outside of ipfs-config?  If not, then ipfs-config seems like a safe way to proceed.  If we think these are non-final values, should we use the internal section: https://github.com/ipfs/go-ipfs/blob/master/config/config.go#L40 ?

Basically, I think to move this quicker, please come forward with what you think is best and why and ask for a thumbs up/down (and feedback).  I think this will put less on the reviewers and more likely to get pushed through if they only have a small time window.

schomatis:
@BigLep I get the rationale here and I agree, but you're asking me for context I do not have. This is not really a very technical decision but rather a "what would the user expect?" kind of question I have no useful input for, and also is not a very time consuming decision for someone with the required context.

In general this kind of context question seems better suited for the IPFS Discord channel as finding a good source for this context (other than Adin who is usually busy) has a higher chance to succeed there than buried in this thread here.

That said, if it helps close this and move on I'll put the `limit.json` path in experimental and remove it later if someone complains.

marten-seemann:
Datapoint: Lotus uses a `limit.json` file.

@vyzo any thoughts what's best for IPFS?

vyzo:
I would prefer a separate limits.json file here too which can easily be shared/dropped in.

BigLep:
@schomatis : fair points on multiple fronts.

I realize you need input so can make meaningful progress.  I just want to make sure we're doing the push to get that feedback.  Agreed that a Discord thread is likely more effective to get this decided.  Feel free to create those.

I assumed that you would have an opinion as someone who has been around ipfs what go-ipfs users expect when it comes to configuring features.  Understood if that isn't the case for this one - my miss.

I know I'm not close enough to the user experience to comment on what's right here, but if there isn't a precedent for having a separate .json file for configuring features in go-ipfs, I wonder why we're doing it now.  What makes this special?  Isn't the ipfs confi that is persisted on disk a file that can be shared (or at least relevant sections of it)?

schomatis:
> Before we ship though, we'll move this to ipfs-config.

Sounds good. Assign and ping me when we're there.

BigLep:
I don't think I was structuring this conversation the best earlier.  I was more focused on what's right for the long term vs immediate.  Apologies.

Having thought about this more...

The most important thing here is to unblock the go-libp2p 0.18 release.  go-lip2p 0.18 doesn't want to move from RC phase to final release until a production deployment with go-libp2p 0.18 with the resource manager has occurred and baked for at least a day.  We believe the easiest/most-effective place to do this is by doing a one-off deployment to the ipfs.io gateway.

Lotus isn't going to pick go-libp2p 0.18 until the final release of go-libp2p 0.18, and it's in Lotus that it's most important to get go-libp2p 0.18.

As a result, anything we can do to accelerate the deployment of go-ipfs with go-libp2p 0.18 RC to ipfs.io gateways should be done.  If limit.json is the quicker path, lets do it.  We can still change that before we mark go-libp2p resource manager integration into go-ipfs as done-done-done in https://github.com/ipfs/go-ipfs/issues/8761.

@schomatis : you previously said:

> That said, if it helps close this and move on I'll put the limit.json path in experimental and remove it later if someone complains.

If you can do this today, that would be awesome.  @marten-seemann is traveling and can use any help to get us ready for the ipfs.io deployment.

schomatis:
@BigLep That comment was actually to address your concerns. @marten-seemann already expressed his approval to the current setup in https://discord.com/channels/806902334369824788/942673321852563456/951177875153895424, so I would avoid pushing new changes at this stage in the test cycle unless this is blocking him somehow. If it does, please ping me here or in Discord and I can address it ASAP.

marten-seemann:
I updated the branch to v0.18.0.

This should be ready for merging now. Before cutting the next go-ipfs release, we should
1. get https://github.com/ipfs/go-ipfs/pull/8785 merged
2. synchronize the connmgr and the rcmgr limits, similar to https://github.com/filecoin-project/lotus/pull/8318.
3. 
I suggest to do that in follow-up PRs, after this one has been merged.

BigLep:
Thanks @marten-seemann.

The overarching story for resource manager in go-ipfs is being tracked in https://github.com/ipfs/go-ipfs/issues/8761.  I believe we have the items you listed there covered.

I have marked clear that this PR (which handles the bare-minimum to update the go-libp2p dependency and the command for resource manager stats https://github.com/ipfs/go-ipfs/issues/8722 ) is ready for review.  @lidel is going to be the owner for reviewing on the go-ipfs maintainer side.

lidel:
Resource manager is disabled by default: This ensures we can safely test the resource manager without impacting
default behavior.


- I am reviewing this, and finishing configuration / docs
- Rebased this PR to include  changes from https://github.com/ipfs/go-ipfs/pull/8802, allowing us to mark specific commands as experimental.



- Added `Swarm.ResourceMgr` to the config
  - we can decide to move it to `Internal.ResourceMgr` but it seems similar to RelayV2 cconfigs which were added in their "final destination", but kept some options turned off via Flag.
    - Default for `Swarm.ResourceMgr.Enabled` is a Flag set to `false` for now
    - 
- `Swarm.ResourceMgr.Limits` allows user to tweak limits per specific  scope in a way that is persisted across restarts
- 'ipfs swarm limit system' outputs human-readable json
- 'ipfs swarm limit system new-limits.json' sets new runtime limits
  (but does not change Swarm.ResourceMgr.Limits in the config)

Conventions to make libp2p devs life easier:
- 'IPFS_RCMGR=1 ipfs daemon' overrides the config and enables resource manager
- 'limit.json' overrides implicit defaults from libp2p (if present)

core/node/groups.go
```diff
@@ -133,17 +134,30 @@ func LibP2P(bcfg *BuildCfg, cfg *config.Config) fx.Option {
...
+	case "1", "true":
+		enableResourceMgr = true
+	}
+
 	// Gather all the options
 	opts := fx.Options(
 		BaseLibP2P,
 
+		// Services (resource management)
+		maybeProvide(libp2p.ResourceManager(cfg.Swarm.ResourceMgr), enableResourceMgr),
```
lidel, line 159: @marten-seemann @vyzo  is it possible to run go-libp2p >=0.18 without _any_ resource manager?

- This line will skip RM registration when `enableResourceMgr` is `false` (default), but I am unsure if there are any implicit defaults in go-libp2p.
  - If so, I can change this to always initialize but use `network.NullResourceManager` instead (like in https://github.com/filecoin-project/lotus/pull/8318)

vyzo: no, you need to pass the nullman to disable.

lidel (approved):
This is done and ready for final reviews.

@vyzo @marten-seemann  if this looks good, just give :+1: 

- Added basic tests for `swarm limit|stats` commands and `Swarm.ResourceMgr.Enabled` flag
- CI is not green due to an unrelated `go-ipfs-http-client` problem, but that is being fixed in https://github.com/ipfs/go-ipfs/pull/8838.
   TBD if we wait for that, and rebase, or merge as-is.

docs/config.md
```diff
@@ -1628,6 +1631,78 @@ Default: `"20s"`
...
 Type: `duration`
 
+### `Swarm.ResourceMgr`
+
+The [libp2p Network Resource Manager](https://github.com/libp2p/go-libp2p-resource-manager#readme) allows setting limits per a scope,
+and track recource usage over time.
+
+#### `Swarm.ResourceMgr.Enabled`
+
+**EXPERIMENTAL**: this feature is is disabled by default, use with caution.
```
schomatis, line 1641:
```suggestion
**EXPERIMENTAL**: this feature is disabled by default, use with caution.
```

core/node/libp2p/rcmgr.go
```diff
@@ -0,0 +1,375 @@
...
+				defer limitFile.Close() //nolint:errcheck
+				limiter, err = rcmgr.NewDefaultLimiterFromJSON(limitFile)
+				if err != nil {
+					return nil, opts, fmt.Errorf("error parsing limit file: %w", err)
+				}
+
+			} else {
+				// Use defaults from go-libp2p
+				log.Debug("limit file %s not found, creating a default resource manager", NetLimitDefaultFilename)
+				limiter = rcmgr.NewDefaultLimiter()
```
lidel, line 69: - [ ] hardcode these implicit defaults in go-ipfs sources + document them in `docs/config.md`
- [ ] (TBD) remove `limit.json` (one can copy&paste values from `docs/config.md` and update `$IPFS_PATH/config`)
  - @vyzo @marten-seemann how painful will be removal of `limit.json` logic above for you? we'd like to reduce the number of places where user can override the defaults.

lidel: I ended up keeping `limit.json` and hiding `Swarm.ResourceMgr.Limits` for now.
Having a single source of truth makes RC testing easier.

lidel:
Applied changes based on feedback from stewards sync:

- `core/node/libp2p/rcmgr_defaults.go` vendors all implicit default limits from go-libp2p
- Ended up keeping `limit.json` parser and hiding `Swarm.ResourceMgr.Limits` for now. 
  - The way defaults are handled and initialized by go-libp2p is bit complex, there are at least two layers of indirection  using different structs. 
  - Coming up with `Swarm.ResourceMgr.Limits` that acts as persistence for `ipfs swarm limit -s`   and is compatible with `limit.json` will improve UX, but that is additional work which should not block go-libp2p 0.18 from being merged.
-  Ported connection manager fix from https://github.com/filecoin-project/lotus/pull/8318
   - `adjustedDefaultLimits` allows for tweaking defaults based on external factors, such as values in existing `Swarm.ConnMgr.HighWater` config.

core/node/libp2p/rcmgr_metrics.go
```diff
@@ -0,0 +1,239 @@
...
+	const (
+		direction = "direction"
+		usesFD    = "usesFD"
+		protocol  = "protocol"
+		service   = "service"
+	)
+
+	connAllowed := prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "libp2p_rcmgr_conns_allowed_total",
```
guseggert, line 24: do all these metrics need the "_total" suffix?

core/node/libp2p/rcmgr.go
```diff
@@ -0,0 +1,368 @@
...
+	case scope == config.ResourceMgrTransientScope:
+		err = mgr.ViewTransient(func(s network.ResourceScope) error {
+			stat := s.Stat()
+			result.Transient = &stat
+			return nil
+		})
+		return result, err
+
+	case strings.HasPrefix(scope, config.ResourceMgrServiceScopePrefix):
+		svc := scope[4:]
```
guseggert, line 158: Can we use `svc := strings.TrimPrefix(scope, config.ResourceMgrServiceScopePrefix)` for this and the other scope prefixes?

docs/config.md
```diff
@@ -1628,6 +1630,68 @@ Default: `"20s"`
 
 Type: `duration`
 
+### `Swarm.ResourceMgr`
+
+The [libp2p Network Resource Manager](https://github.com/libp2p/go-libp2p-resource-manager#readme) allows setting limits per a scope,
+and track recource usage over time.
```
guseggert, line 1636:
```suggestion
The [libp2p Network Resource Manager](https://github.com/libp2p/go-libp2p-resource-manager#readme) allows setting limits per a scope
and tracking recource usage over time.
```

go.mod
```diff
@@ -62,37 +62,38 @@ require (
 	github.com/ipld/go-car v0.3.2
 	github.com/ipld/go-car/v2 v2.1.1
 	github.com/ipld/go-codec-dagpb v1.3.0
-	github.com/ipld/go-ipld-prime v0.14.2
+	github.com/ipld/go-ipld-prime v0.14.1
```
guseggert, line 65: Was this intentional? Do we want to downgrade this?

go.mod
```diff
@@ -22,7 +22,7 @@ require (
 	github.com/ipfs/go-datastore v0.5.1
 	github.com/ipfs/go-detect-race v0.0.1
 	github.com/ipfs/go-ds-badger v0.3.0
-	github.com/ipfs/go-ds-flatfs v0.5.1
+	github.com/ipfs/go-ds-flatfs v0.5.0
```
guseggert, line 25: Was this intentional? Do we want to downgrade this?

core/node/libp2p/rcmgr_metrics.go
lidel: I think it is a convention in case we ever add partials in the future
 http://127.0.0.1:5001/debug/metrics/prometheus has a lot of those:
 ```
$  curl -s http://127.0.0.1:5001/debug/metrics/prometheus | grep -c _total
361
 ```

core/commands/swarm.go
```diff
@@ -304,6 +310,133 @@ var swarmPeersCmd = &cmds.Command{
...
+	Status: cmds.Experimental,
+	Helptext: cmds.HelpText{
+		Tagline: "Get or set resource limits for a scope.",
+		LongDescription: `Get or set resource limits for a scope.
+The scope can be one of the following:
+- system        -- limits for the system aggregate resource usage.
+- transient     -- limits for the transient resource usage.
+- svc:<service> -- limits for the resource usage of a specific service.
+- proto:<proto> -- limits for the resource usage of a specific protocol.
+- peer:<peer>   -- limits for the resource usage of a specific peer.
```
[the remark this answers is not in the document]
lidel: Good idea, extracted this into https://github.com/ipfs/go-ipfs/issues/8858

go.mod
lidel: No, it was introduced with the very first commit, possible regression due to rebase.
Pushed commit that reverts this and other downgrades like this.

go.mod
lidel: No, reverted. https://github.com/ipfs/go-ipfs/pull/8680/commits/2d3b3c6bca4067b35f09e466e8f28a8302bd7266#r845619647

guseggert (approved):

Issue #8470, opened 2021-09-30
Way to disable progress bars for 'cat' and 'get'


andrewchambers:
### Checklist

- [X] My issue is specific & actionable.
- [X] I am not suggesting a protocol enhancement.
- [X] I have searched on the [issue tracker](https://github.com/ipfs/go-ipfs/issues?q=is%3Aissue) for my issue.

### Description

In some scripting contexts it is not acceptable to display a progress bar, but there is no way to disable the progress bars without also hiding error messages (by redirecting stderr).

Ideally there would be a -q option or --no-progress such for such situations.

schomatis:
This seems like a fair request: we should have the same behavior of silencing the progress bar (or any other output) like `ipfs add` does. Assigning myself with no commitment on timeline (as there are higher-priority issues in the queue), but feel free to ping in a couple of weeks if this hasn't moved.

schomatis:
Working on this now.

schomatis:
@andrewchambers Could you take a look at https://github.com/ipfs/go-ipfs/pull/8686 to see if this works for you?

andrewchambers:
looks pretty good - i guess as long as its consistent with how other commands disable it, it looks good to me.

PR #8686, opened 2022-01-20
fix(cmds): option for progress bar in cat/get

schomatis:
Add a `-p=false` flag to not show the progress bar (true by default to keep the current behavior).

Closes https://github.com/ipfs/go-ipfs/issues/8470.

### DIFF
--- a/core/commands/cat.go
+++ b/core/commands/cat.go
@@ -106,6 +106,7 @@ var CatCmd = &cmds.Command{
 						var bar *pb.ProgressBar
 						bar, reader = progressBarForReader(os.Stderr, val, int64(res.Length()))
 						bar.Start()
+						defer bar.Finish()
 					}
 
 					err = re.Emit(reader)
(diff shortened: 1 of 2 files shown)


guseggert (approved):

core/commands/cat.go
```diff
@@ -96,8 +98,15 @@ var CatCmd = &cmds.Command{
...
-					bar, reader := progressBarForReader(os.Stderr, val, int64(res.Length()))
-					bar.Start()
+					reader := val
+
+					req := res.Request()
+					progress, _ := req.Options[progressOptionName].(bool)
+					if progress {
+						var bar *pb.ProgressBar
+						bar, reader = progressBarForReader(os.Stderr, val, int64(res.Length()))
+						bar.Start()
```
guseggert, line 108: does this need a `bar.Finish()` somewhere?

schomatis: Wouldn't hurt a defer.

BigLep:
@schomatis : any reason we don't merge this given you have approval?

schomatis:
Don't have permissions for the main repo (and I'm fine if we want to keep it this way).

PR #8690, opened 2022-01-21
fix: allow ipfs-companion browser extension to access RPC API

### LINKED ISSUE #8689: Add ipfs-companion chrome extension id to accepted origins for future manifest v3 version
[Manifest v3](https://developer.chrome.com/docs/extensions/mv3/intro/) removes support for the blocking webRequest api. We are currently using this api to intercept requests and modify the Origin URL to that of the ipfs-webui. With support for this feature gone we need another way for companion to make api requests to ipfs.

I'm proposing that we add the chrome-extension-id (which is generated once on upload to the chrome web store) to the [default allowed origins](https://github.com/ipfs/go-ipfs/blob/ef866a1400b3b2861e5e8b6cc9edc8633b890a0a/core/corehttp/commands.go#L40). This way companion can still be useful for users when manifest v2 is deprecated, and they will not have to manually update their configurations.

I have discussed this a bit with @lidel on slack, but wanted to get the ball rolling here so that we can have the ability to get ipfs-companion manifest v3 into the chrome store (if only as a beta channel for power users) by the end of Q1.

For further context: https://github.com/ipfs/ipfs-companion/issues/666 (mv3 tracking issues)

meandavejustice:
- fixes #8689
- Adds the chrome-extension ids for ipfs-companion and
ipfs-companion-beta to the allowed origins list, this
allows us to accesss ipfs api from a manifest v3 extension.
- added tests in t0401-api-browser-security.sh

### DIFF
--- a/core/corehttp/commands.go
+++ b/core/corehttp/commands.go
@@ -44,6 +44,9 @@ var defaultLocalhostOrigins = []string{
 	"https://[::1]:<port>",
 	"http://localhost:<port>",
 	"https://localhost:<port>",
+}
+
+var companionBrowserExtensionOrigins = []string{
 	"chrome-extension://nibjojkomfdiaoajekhjakgkdhaomnch", // ipfs-companion
 	"chrome-extension://hjoieblefckbooibpepigmacodalfndh", // ipfs-companion-beta
 }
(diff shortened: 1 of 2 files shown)


meandavejustice:
cc/ @lidel

lidel (approved):
LGTM. I'll add tests to confirm this still works when one defined own CORS safelist.

This ensures we don't break two cases:
- users of ipfs-desktop: it adds own CORS entries
- users of webui.ipfs.io (bleeding edge version) who manually added it to CORS safelist
- webui developers, who safelist localhost Origins for local dev

PR #8692, opened 2022-01-22
rados plugin

coryschwartz:
Maybe this should be somewhere else, a standalone plugin maybe.

This is a plugin for IPFS backed by a ceph cluster.

To get this working, you need a ceph cluster.

Ideally, create a new pool in ceph for whichever datastores you want to be stored there.

```
ceph osd pool create ipfsblocks
ceph osd pool create ipfsdatastore
```

and set appropriate permissions, as well as any other performance characteristics are appropriate for the pool.

The configuration file should be edited to include the follwoing:

```
    "Spec": {
      "mounts": [
        {
          "child": {
            "path": "blocks",
	    "pool": "ipfsblocks",
	    "configPath": "/etc/ceph/ceph.conf",
            "type": "radosds"
          },
          "mountpoint": "/blocks",
          "prefix": "rados.blocks",
          "type": "measure"
        },
        {
          "child": {
            "path": "datastore",
	    "pool": "ipfsdatastore",
	    "configPath": "/etc/ceph/ceph.conf",
            "type": "radosds"
          },
          "mountpoint": "/",
          "prefix": "rados.datastore",
          "type": "measure"
        }
      ],
      "type": "mount"
    },

```

and init with the provided config. Althernatively, you may need to replace the datastore_spec file.

Whew! that's a lot of work, but it's alive.

On a small test ceph cluster, and it is taking up space.

```
device_health_metrics      0 B        0       0       0                   0        0         0       0      0 B       0      0 B         0 B          0 B
ipfsdatastore          4.5 MiB      388       0    1164                   0        0         0     646  646 KiB     398  394 KiB         0 B          0 B
ipfsblocks              13 GiB    46222       0  138666                   0        0         0  402335  393 MiB   46222  4.3 GiB         0 B          0 B
```

Some ceph clusters can be quite large, multiple PB systems. Further testing is required to understand if the system performs well with larger-scale storage.


guseggert:
We're pretty conservative with adding new datastores into go-ipfs. Would https://github.com/ipfs/go-ipfs/pull/8734 work with your requirements? Steps would be to wrap the rados datastore [with this](https://github.com/guseggert/go-ds-grpc/blob/main/server/server.go#L37), build a binary that runs that server, configure go-ipfs to call it (see the sharness test in that PR for example config), and run it before starting the go-ipfs daemon.

coryschwartz:
@guseggert Actually, yeah I think so. the grpc datastore looks like a great idea. I'm going to go ahead and close this one -- it was something I was trying out, but it pulls in some non-golang dependencies that proably most people wouldn't want. Wasn't really happy with it, but the grpc datastore would make this much better

PR #8715, opened 2022-02-04
fix: installation without sudo

vmx:
I think it's good practice that users don't need to run `sudo`
to install something just to try it out. With this change, the
local bin directory is tried first, which usually is also in the
`PATH`. This way the installation script can be run without
`sudo` and should still work.

### DIFF
--- a/cmd/ipfs/dist/install.sh
+++ b/cmd/ipfs/dist/install.sh
@@ -6,7 +6,7 @@
 INSTALL_DIR=$(dirname $0)
 
 bin="$INSTALL_DIR/ipfs"
-binpaths="/usr/local/bin /usr/bin"
+binpaths="$HOME/.local/bin /usr/local/bin /usr/bin"
 
 # This variable contains a nonzero length string in case the script fails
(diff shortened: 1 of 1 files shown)


cmd/ipfs/dist/install.sh
```diff
@@ -6,7 +6,7 @@
 INSTALL_DIR=$(dirname $0)
 
 bin="$INSTALL_DIR/ipfs"
-binpaths="/usr/local/bin /usr/bin"
+binpaths="$HOME/.local/bin /usr/local/bin /usr/bin"
```
guseggert, line 9: can we put this at the end of the list so that it's backwards compatible? (i.e. if you run w/ sudo it will continue to put the bin in the same place)

guseggert: Also will this work if $HOME has a space in it?

vmx: @guseggert thanks for the review, the new commit should fix the issues you've mentioned.

guseggert (approved):

Issue #8717, opened 2022-02-04
Meta: HTTP Gateway cache control improvements


lidel:
This is a meta issue for HTTP cache improvements that we should prioritize in go-ipfs:
- [ ] `max-age` on `/ipns/` based on TTL – https://github.com/ipfs/go-ipfs/issues/1818
- [ ] dir-index-html on `/ipfs/` should have some reasonable `max-age` (it won't change within releases)
  -  (wip: `fix/dir-index-html-max-age` @lidel)
- [ ] some HTTP header in `/ipns/` responses which contains resolved IPNS (root CID)  for faster cache invalidation of entire website when IPNS pointer changes
   - suggested in https://github.com/ipfs/go-ipfs/pull/8617
      - proposed `X-Ipfs-Roots` in https://github.com/ipfs/go-ipfs/pull/8720 a way to indicate all CIDs required for resolving  path segments  from `x-ipfs-path`
-  [ ] ensure we default to https://web.dev/stale-while-revalidate/ everywhere we can
  - this maximizes utility of HTTP gateway – cache-control is used not only by end user (browsers), but also HTTP  caching solutions, CDNs. We should have defaults that leverage HTTP caching tools and conversions, assuming people will manually configure HTTP caching on their own using some sane values is not realistic.

cc @thattommyhall & @mathew-cf) if there are more asks/ideas here

thattommyhall:
Only thing I would add is a strategy of expiring at "top of the hour" or "end of the day" for the DirIndex pages makes dealing with when the etag does eventually change a bit nicer (you wont see different things on different servers for as long), which nudges me to prefer Expires to Cache-Control (but its not _that_ much extra to calculate it anyway)

mathew-cf:
An idea in addition to the ones proposed:
A public resolution endpoint for gateways could be useful as an inexpensive, cacheable call. There is no way to do IPNS public key resolution without directly accessing the IPFS network or fetching the whole file, so an endpoint exposing the resolution could be helpful for that.

thattommyhall:
@mathew-cf thats useful, but you'd have to rewrite to fetch /ipfs/<cid>/<path> or you couldnt be certain that the content you fetched matched. Like if I checked what the dnslink was for blog.ipfs.io then fetched /ipns/blog.ipfs.io it might have changed in the meantime

mathew-cf:
[...]

PR #8720, opened 2022-02-07
feat: X-Ipfs-Roots for smarter HTTP caches

lidel:
This PR adds `X-Ipfs-Roots` to Gateway responses.

`X-Ipfs-Roots` is a way to indicate all CIDs required for resolving  path segments  from `X-Ipfs-Path` – both headers are meant to improve interop with existing HTTP software (load-balancers, caches, CDNs).

It allows HTTP caches and CDNs make better decisions around cache invalidation: not only invalidate everything under specific IPNS website when the  root changes, but also do more fine-grained caching and content distribution when CID/block aware solutions are in place.

## Rationale

`X-Ipfs-Path` returned on mutable  `/ipns/` paths is not enough, and Gateway operator community raised the need for additional metadata to enable smarter cache.

This is a low-hanging fruit from https://github.com/ipfs/go-ipfs/issues/8717 that can be shipped independently of other improvements. 

### Before

```console
curl -Is http://127.0.0.1:8080/ipns/docs.ipfs.io/concepts/glossary/ | grep -iE "X-Ipfs|Etag"
Etag: "DirIndex-512eb789cd905714e03f29d4e04de7549e8c9c3e_CID-QmZcUjyj9VamSpGJ8VkWuZAvRMKinm6kBBUhnGy3omNZJC"
X-Ipfs-Path: /ipns/docs.ipfs.io/concepts/glossary/
```

### After

```console
curl -Is http://127.0.0.1:8080/ipns/docs.ipfs.io/concepts/glossary/ | grep -iE "X-Ipfs|Etag"
Etag: "QmZcUjyj9VamSpGJ8VkWuZAvRMKinm6kBBUhnGy3omNZJC"
X-Ipfs-Path: /ipns/docs.ipfs.io/concepts/glossary/
X-Ipfs-Roots: Qmem46CBFD4XHVBsGtoTbM2u5nuRx1XUiRrSXqUKrBxtpa,QmasfzzoGU8FE5z4w3GDfW3DrJVvGPCAbfgTDrGWarxRso,QmZcUjyj9VamSpGJ8VkWuZAvRMKinm6kBBUhnGy3omNZJC
```

## TODO

- [x] add `X-Ipfs-Roots` (all path CIDs folded  into a single header)
- [x] gather feedback from CDN/gateway operators
- [x] add sharness tests for `X-Ipfs-Roots` (new) and `X-Ipfs-Path` (missing) and `Etag` (missing)
- [x] fix a bug where we return `DirIndex-` Etag for `index.html` response (on EVERY WEBSITE EVER :see_no_evil:)


thattommyhall:
I'm not understand why it's plural. In my head `$root_cid` matches whats currently the `_dnslink` entry.

The way I think about it is that at any moment `docs.ipfs.io` has a root-cid in `_dnslink` and the content at `docs.ipfs.io$uri` would be the same as `ipfs://$root_cid$uri` so that when one uri expired and got refreshed we could be sure the the root-cid had not changed and we could keep serving up cached content (im thinking the cache key would be `$root_cid$uri`)

thattommyhall (approved):

mathew-cf (approved):
This is a great caching change for paths that can be resolved to a CID by DNS

mathew-cf (approved):

guseggert:
LGTM so far

lidel:
Pushed bunch of tests that document and protect the behavior around
 `X-Ipfs-Roots` (added in this PR) but also existing `X-Ipfs-Path`and `Etag` (which had a bug – also fixed in this PR).
 
 This is ready for final reviews / comments. If no concerns, the plan is to  merge  and ship this in go-ipfs 0.13.

PR #8723, opened 2022-02-08
docs: add Internal.UnixFSShardingSizeThreshold

aschmahmann:


docs/config.md
```diff
@@ -275,6 +276,14 @@ an implicit default when `null` or missing from the config file:
 - `null`/missing (apply the default value defined in go-ipfs sources)
 - an integer between `-2^63` and `2^63-1` (i.e. `-9223372036854775808` to `9223372036854775807`)
 
+### `optionalBytes`
+
+Optional Bytes allow specifying some number of bytes which has
+an implicit default when `null` or missing from the config file:
+
+- `null`/missing (apply the default value defined in go-ipfs sources)
+- a string value indicating the number of bytes (e.g. 1B, 10kB, 20 MiB, 2GB, ...)
```
lidel, line 285: I never remember which is which, and even if I do, I don't trust apps and [check](https://github.com/dustin/go-humanize/blob/afde56e7acacd811f6c94228c2c61af2b0e93158/bytes.go#L88-L110) if they correctly support one or another. :see_no_evil: 
By writing it down, we save everyone a lot of time on double-checking:
```suggestion
- a string value indicating the number of bytes, including human readable representations:
  - [SI sizes](https://en.wikipedia.org/wiki/Metric_prefix#List_of_SI_prefixes) (metric units, powers of 1000), e.g. `1B`, `2kB`, `3MB`, `4GB`, `5TB`, …)
  - [IEC sizes](https://en.wikipedia.org/wiki/Binary_prefix#IEC_prefixes) (binary units, powers of 1024), e.g. `1B`, `2KiB`, `3MiB`, `4GiB`, `5TiB`, …)
```

docs/config.md
```diff
@@ -919,6 +928,18 @@ deteriorate the quality provided to less aggressively-wanting peers.
 
 Type: `optionalInteger` (byte count, `null` means default which is 1MB)
 
+### `Internal.UnixFSShardingSizeThreshold`
+
+The sharding threshold used internally to decide whether a UnixFS directory should be sharded or not.
+This value is not strictly related to the size of the UnixFS directory block and any increases in
+the threshold should come with being careful that block sizes stay under 2MiB in order for them to be
+reliably transferable through the networking stack.
```
lidel, line 936: Maybe we should make it clear this is defacto how public swarm works?

```suggestion
reliably transferable through the networking stack (IPFS peers on the public swarm tend to ignore requests for blocks bigger than 2MiB).
```

docs/config.md
```diff
@@ -919,6 +928,18 @@ deteriorate the quality provided to less aggressively-wanting peers.
...
+
+Decreasing this value to 1B is functionally equivalent to the previous experimental sharding option to
+shard all directories.
+
+Type: `optionalBytes` (`null` means default which is 256kiB)
```
lidel, line 941: :trollface: 
```suggestion
Type: `optionalBytes` (`null` means default which is 256KiB)
```

lidel (changes requested):

lidel (approved):

PR #8728, opened 2022-02-10
feat: refactor Fetcher interface used for downloading migrations

### LINKED ISSUE #8714: Better handling of fallbacks for migration downloads
### Checklist

### Installation method

ipfs-update or dist.ipfs.io

### Version

```Text
go-ipfs v0.12.0-rc1
```

### Config

```json
Default config for a fresh node
```

### Description

Currently when go-ipfs tries to fetch migrations when running the daemon and realizing that the repo version needs to be upgraded it can fetch the migrations over an HTTP gateway or over IPFS.

Unfortunately the way in which fallbacks are used (e.g. try a gateway then IPFS, or vice versa) 
uses the `MultiFetcher` https://github.com/ipfs/go-ipfs/blob/cde79df1408c3bd518fec1622d97bf4a251af81e/repo/fsrepo/migrations/fetcher.go#L42

The `MultiFetcher` returns when the first fetch of an `io.ReadCloser` succeeds https://github.com/ipfs/go-ipfs/blob/cde79df1408c3bd518fec1622d97bf4a251af81e/repo/fsrepo/migrations/fetcher.go#L54-L58 

Unfortunately, this means that if the download terminates in the middle no fallback occurs.

----

There are a number of ways to fix this

 One relatively simple one would be modifying the `Fetcher` interface's `Fetch` function to take a writer so it can output to a file and therefore the `Fetch` function will only return when it's done fetching everything. Existing implementations could be made compliant by taking the `io.ReadCloser` they were previously returning and doing an `io.Copy` into the new writer before returning.

This makes the simple `MultiFetcher` logic work at the cost of not saving some bytes and being smarter about download resumption.

aschmahmann:
Resolves #8714

### DIFF
--- a/repo/fsrepo/migrations/fetcher.go
+++ b/repo/fsrepo/migrations/fetcher.go
@@ -1,6 +1,7 @@
 package migrations
 
 import (
+	"bytes"
 	"context"
 	"io"
 	"os"
@@ -21,7 +22,7 @@ const (
 type Fetcher interface {
 	// Fetch attempts to fetch the file at the given ipfs path.
 	// Returns io.ReadCloser on success, which caller must close.
-	Fetch(ctx context.Context, filePath string) (io.ReadCloser, error)
+	Fetch(ctx context.Context, filePath string, writer io.Writer) error
 	// Close performs any cleanup after the fetcher is not longer needed.
 	Close() error
 }
@@ -49,16 +50,20 @@ func NewMultiFetcher(f ...Fetcher) Fetcher {
 
 // Fetch attempts to fetch the file at each of its fetchers until one succeeds.
 // Returns io.ReadCloser on success, which caller must close.
-func (f *MultiFetcher) Fetch(ctx context.Context, ipfsPath string) (io.ReadCloser, error) {
+func (f *MultiFetcher) Fetch(ctx context.Context, ipfsPath string, writer io.Writer) error {
 	var errs error
 	for _, fetcher := range f.fetchers {
-		rc, err := fetcher.Fetch(ctx, ipfsPath)
+		var buf bytes.Buffer
+		err := fetcher.Fetch(ctx, ipfsPath, &buf)
 		if err == nil {
-			return rc, nil
+			if _, err := io.Copy(writer, &buf); err != nil {
+				return err
+			}
+			return nil
 		}
 		errs = multierror.Append(errs, err)
 	}
-	return nil, errs
+	return errs
 }
 
 func (f *MultiFetcher) Close() error {
--- a/repo/fsrepo/migrations/fetcher.go
+++ b/repo/fsrepo/migrations/fetcher.go
@@ -20,7 +20,6 @@ const (
 
 type Fetcher interface {
 	// Fetch attempts to fetch the file at the given ipfs path.
-	// Returns io.ReadCloser on success, which caller must close.
 	Fetch(ctx context.Context, filePath string) ([]byte, error)
 	// Close performs any cleanup after the fetcher is not longer needed.
 	Close() error
@@ -48,7 +47,6 @@ func NewMultiFetcher(f ...Fetcher) Fetcher {
 }
 
 // Fetch attempts to fetch the file at each of its fetchers until one succeeds.
-// Returns io.ReadCloser on success, which caller must close.
 func (f *MultiFetcher) Fetch(ctx context.Context, ipfsPath string) ([]byte, error) {
 	var errs error
 	for _, fetcher := range f.fetchers {
--- a/repo/fsrepo/migrations/fetcher.go
+++ b/repo/fsrepo/migrations/fetcher.go
@@ -1,7 +1,6 @@
 package migrations
 
 import (
-	"bytes"
 	"context"
 	"io"
 	"os"
@@ -22,7 +21,7 @@ const (
 type Fetcher interface {
 	// Fetch attempts to fetch the file at the given ipfs path.
 	// Returns io.ReadCloser on success, which caller must close.
-	Fetch(ctx context.Context, filePath string, writer io.Writer) error
+	Fetch(ctx context.Context, filePath string) ([]byte, error)
 	// Close performs any cleanup after the fetcher is not longer needed.
 	Close() error
 }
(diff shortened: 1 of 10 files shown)


repo/fsrepo/migrations/fetcher.go
```diff
@@ -49,16 +50,20 @@ func NewMultiFetcher(f ...Fetcher) Fetcher {
 
 // Fetch attempts to fetch the file at each of its fetchers until one succeeds.
 // Returns io.ReadCloser on success, which caller must close.
```
guseggert, line 52: needs to be updated

repo/fsrepo/migrations/fetcher.go
```diff
@@ -49,16 +50,20 @@ func NewMultiFetcher(f ...Fetcher) Fetcher {
...
+		var buf bytes.Buffer
+		err := fetcher.Fetch(ctx, ipfsPath, &buf)
 		if err == nil {
-			return rc, nil
+			if _, err := io.Copy(writer, &buf); err != nil {
+				return err
+			}
+			return nil
 		}
 		errs = multierror.Append(errs, err)
```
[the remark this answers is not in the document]
aschmahmann: Do you want to take a look? Right now we're not using go-log during the migrations (just the standard "log" package) which means we might be missing the hooks to make this log visible during the migration.  We can probably switch over, but unless the user runs with the one of the go-log env vars like https://github.com/ipfs/go-log/blob/604185d3545622df35d1bb76100bfc8b04293f9b/setup.go#L32 set to debug this won't show up.

repo/fsrepo/migrations/retryfetcher.go
```diff
@@ -0,0 +1,43 @@
...
+type RetryFetcher struct {
+	Fetcher
+	maxRetries int
+}
+
+var _ Fetcher = (*RetryFetcher)(nil)
+
+func NewRetryFetcher(baseFetcher Fetcher, maxRetries int) *RetryFetcher {
+	return &RetryFetcher{Fetcher: baseFetcher, maxRetries: maxRetries}
+}
```
aschmahmann, line 17: @guseggert @lidel I added the retry fetcher from https://github.com/ipfs/ipfs-update/pull/158. This is really simplistic and likely we'll keep breaking this to add more complex backoff and retry strategies so maybe this doesn't belong here.

Similarly do we actually want to do this or do we just want to fallback to the next download strategy on the list if there's a problem (as opposed to in ipfs-update where we might just want to retry)?

repo/fsrepo/migrations/retryfetcher.go
```diff
@@ -0,0 +1,43 @@
...
+
+func (r *RetryFetcher) Fetch(ctx context.Context, filePath string, writer io.Writer) error {
+	var lastErr error
+	for i := 0; i < r.maxRetries; i++ {
```
guseggert, line 23: the way this is written, maxRetries is more like "maxTries"

repo/fsrepo/migrations/migrations.go
```diff
@@ -155,13 +155,14 @@ func ReadMigrationConfig(repoRoot string) (*config.Migration, error) {
 // downloadSources,
 func GetMigrationFetcher(downloadSources []string, distPath string, newIpfsFetcher func(string) Fetcher) (Fetcher, error) {
 	const httpUserAgent = "go-ipfs"
+	const numRetriesPerHTTP = 3
```
aschmahmann, line 158: Just made up a number here, no delays, etc. Again not sure we actually need/want this here

repo/fsrepo/migrations/fetcher.go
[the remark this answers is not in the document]
guseggert: Is there a reason we don't use go-log?

Also I can't really think of a downside to printing the attempts to stdout.

I'll take a stab

repo/fsrepo/migrations/retryfetcher.go
```diff
@@ -0,0 +1,37 @@
...
+var _ Fetcher = (*RetryFetcher)(nil)
+
+func NewRetryFetcher(baseFetcher Fetcher, maxRetries int) *RetryFetcher {
+	return &RetryFetcher{Fetcher: baseFetcher, maxRetries: maxRetries}
+}
+
+func (r *RetryFetcher) Fetch(ctx context.Context, filePath string) ([]byte, error) {
+	var lastErr error
+	for i := 0; i < r.maxRetries; i++ {
+		out, err := r.Fetcher.Fetch(ctx, filePath)
```
aschmahmann, line 22: The RetryFetcher made me reconsider just returning []byte with the interface change, WDYT? Seems like having writers is sort of useless with the retryFetcher until we have a `FetchRange` function or allow passing a range into `Fetch` since we'd otherwise have to buffer it to prevent the Writer from getting corrupted.

We can revert the commit that did this and use writers again, but until we add range support it might not be worth it and it's not like breaking this interface is that costly for us

repo/fsrepo/migrations/retryfetcher.go
```diff
@@ -0,0 +1,43 @@
...
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+)
+
+type RetryFetcher struct {
+	Fetcher
+	maxRetries int
+}
```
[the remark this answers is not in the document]
aschmahmann: Here this might be fine, although generally I've seen people exporting structs protected by initializer functions. I think the main benefit is to do type casting rather than redefining the interface you need, although this seems anti the Go model of trying to declare interfaces at the consumption rather than production points.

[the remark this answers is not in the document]
aschmahmann: I could also just scrap the initializer and make this public, WDYT?

repo/fsrepo/migrations/retryfetcher.go
guseggert: another option is to pass the destination file path into the fetcher, then we wouldn't need to buffer in-mem (write to tmp, then move if/when it succeeds)

repo/fsrepo/migrations/retryfetcher.go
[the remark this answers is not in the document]
guseggert: sure, this is all low risk b/c this has low likelihood of being used anywhere but a handful of places

repo/fsrepo/migrations/retryfetcher.go
aschmahmann: We're not always using tmp files here (e.g. when we download the version file we're happy to just keep the whole thing in memory). Passing in an FD also seems more complicated since you need to worry about similar types of overwriting as with a WriterSeeker.

It's a bit annoying to need to keep the whole file in memory, but hopefully this shouldn't be too big a deal since it's happening before daemon startup when most other resource usage is lower anyway.

repo/fsrepo/migrations/fetcher.go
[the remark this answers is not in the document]
aschmahmann: Thanks, don't think there's any reason other than it wasn't hooked up at the time

repo/fsrepo/migrations/retryfetcher.go
guseggert: At least one retry is desirable IMO, stuff happens and it's a nicer experience to retry on transient issues.

I also just realized that we could have the same effect without having a special RetryFetcher by just putting the HTTP fetcher twice into the multifetcher.

repo/fsrepo/migrations/fetcher.go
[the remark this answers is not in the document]
aschmahmann: #8732

guseggert (approved):

Issue #3536, opened 2016-12-24
Feature request: a Google Drive datastore


Piruzzolo:
Google Drive APIs are not intended for transparent datastore usage - they're basically rest APIs which don't offer binary access and block-device mode. Nevertheless, it would be really useful to store DAG objects on it, in the form of single files maybe. All one needs for it would be a datastore middleware to write to / read from the service and deal with the Google account management.

IPFS supports Amazon S3 as a backend datastore. Much of the code could be shared/reused, I guess.

Open for discussion

whyrusleeping:
The s3 code isnt quite there yet (its high on my todo list actually). Once thats in we can look at figuring out google drive

PR #8734, opened 2022-02-15
feat: add grpcds support

guseggert:
This adds a gRPC datastore shim to `go-ipfs` by wiring up [go-ds-grpc](https://github.com/guseggert/go-ds-grpc). The motivation is to allow custom local/remote datastores to be used with `go-ipfs` without having to recompile the `go-ipfs` binary, compile modules, or even write the datastore in Go. Unix domain sockets are also supported here.

This would enable easier datastore experimentation and enable a third party ecosystem of datastores that end users could use. Common datastores like [go-ds-s3](https://github.com/ipfs/go-ds-s3) could release binaries / Docker images so that folks could just use a datastore sidecar with vanilla Kubo instead of having to patch & rebuild Kubo (currently the case w/ go-ds-s3).

Some relevant issues/PRs:

* https://github.com/ipfs/go-ipfs/pull/8692
* https://github.com/ipfs/go-ipfs/pull/5561
* https://github.com/ipfs/go-ipfs/pull/5516
* https://github.com/ipfs/go-ipfs/pull/5403
* https://github.com/ipfs/go-ipfs/issues/3536
* https://github.com/ipfs/kubo/pull/9659
* https://github.com/storj/ipfs-go-ds-storj

If we can agree that this approach is desirable, then the things I have left to do:

* Move `go-ds-grpc` into IPFS org
* Add more rigorous tests to `go-ds-grpc`
* Add docs

As a side note, @coryschwartz has been using this with Lotus and the [rados datastore](https://github.com/coryschwartz/go-ds-rados-grpc).


guseggert:
I can't get anyone interested in this, so closing.

PR #8741, opened 2022-02-17
fix(ci): testground workflow hanging indefinitely

### LINKED ISSUE #8731: Testground CI timeouts after 6 hours
## Problem

Testground CI runs for 6h and then timeouts:

> 
>
> > 

Did something changed 8 days ago? 

> 

## Solution

Aside from fixing these tests, cap them?
Seems that when it was working it took ~2minutes, perhaps we should set cap at 15m, 1h,  or something, so these hanging jobs don't  squat workers for 6h each.

galargh:
Closes   #8731

Fixes hanging builds problem with https://github.com/coryschwartz/testground-github-action/pull/2 (detailed description of the problem and testing performed).

Additionally, it adds a 5 minutes timeout on testground jobs - they finish in ~2 minutes if everything goes smoothly.

I'll reenable the workflow after the fix is in place.

### DIFF
--- a/.github/workflows/testground-on-push.yml
+++ b/.github/workflows/testground-on-push.yml
@@ -25,7 +25,8 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: testground run
-        uses: coryschwartz/testground-github-action@v1.1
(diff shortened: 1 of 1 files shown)


lidel (approved):
This makes sense but since testground is disabled, lets wait for upstream fix before merging this

PR #8742, opened 2022-02-17
Rc/v0.12.0 final

aschmahmann:
TODO:
- [x] Release new migration and put CID into go-ipfs
- [x] Need to identify what else to cherry-pick in (e.g., fix CI bugs, arm builds for docker)
- [x] Review of changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,158 @@
 # go-ipfs changelog
 
+## v0.12.0 2022-02-17
+
+We're happy to announce go-ipfs 0.12.0. This release switches the storage of IPLD blocks to be keyed by multihash instead of CID.
+
+As usual, this release includes important fixes, some of which may be critical for security. Unless the fix addresses a bug being exploited in the wild, the fix will _not_ be called out in the release notes. Please make sure to update ASAP. See our [release process](https://github.com/ipfs/go-ipfs/tree/master/docs/releases.md#security-fix-policy) for details.
+
+### 🛠 BREAKING CHANGES
+
+- `ipfs refs local` will now list all blocks as if they were [raw]() CIDv1 instead of with whatever CID version and IPLD codecs they were stored with. All other functionality should remain the same.
+
+Keep reading to learn more details.
+
+#### 🔦 Highlights
+
+There is only one change since 0.11:
+
+##### Blockstore migration from full CID to Multihash keys
+
+We are switching the default low level [datastore](https://docs.ipfs.io/concepts/glossary/#datastore) to be keyed only by the [Multihash](https://docs.ipfs.io/concepts/glossary/#multihash) part of the [CID](https://docs.ipfs.io/concepts/glossary/#cid), and deduplicate some [blocks](https://docs.ipfs.io/concepts/glossary/#block) in the process. The blockstore will become [codec](https://docs.ipfs.io/concepts/glossary/#codec)-agnostic.
+
+###### Rationale
+
+The blockstore/datastore layers are not concerned with data interpretation, only with storage of binary blocks and verification that the Multihash they are addressed with (which comes from the CID), matches the block. In fact, different CIDs, with different codecs prefixes, may be carrying the same multihash, and referencing the same block. Carrying the CID abstraction so low on the stack means potentially fetching and storing the same blocks multiple times just because they are referenced by different CIDs. Prior to this change, a CIDv1 with a `dag-cbor` codec and a CIDv1 with a `raw` codec, both containing the same multihash, would result in two identical blocks stored. A CIDv0 and CIDv1 both being the same `dag-pb` block would also result in two copies.
+
+###### How migration works
+
+In order to perform the switch, and start referencing all blocks by their multihash, a migration will occur on update. This migration will take the repository version from 11 (current) to 12.
+
+One thing to note is that any content addressed CIDv0 (all the hashes that start with `Qm...`, the current default in go-ipfs), does not need any migration, as CIDv0 are raw multihashes already. This means the migration will be very lightweight for the majority of users.
+
+The migration process will take care of re-keying any CIDv1 block so that it is only addressed by its multihash. Large nodes with lots of CIDv1-addressed content will need to go through a heavier process as the migration happens. This is how the migration works:
+
+1. Phase 1: The migration script will perform a pass for every block in the datastore and will add all CIDv1s found to a file named `11-to-12-cids.txt`, in the go-ipfs configuration folder. Nothing is written in this first phase and it only serves to identify keys that will be migrated in phase 2.
+2. Phase 2: The migration script will perform a second pass where every CIDv1 block will be read and re-written with its raw-multihash as key. There is 1 worker performing this task, although more can be configured. Every 100MiB-worth of blocks (this is configurable), each worker will trigger a datastore "sync" (to ensure all written data is flushed to disk) and delete the CIDv1-addressed blocks that were just renamed. This provides a good compromise between speed and resources needed to run the migration.
+
+At every sync, the migration emits a log message showing how many blocks need to be rewritten and how far the process is.
+
+####### FlatFS specific migration
+
+For those using a single FlatFS datastore as their backing blockstore (i.e. the default behavior), the migration (but not reversion) will take advantage of the ability to easily move/rename the blocks to improve migration performance.
+
+Unfortunately, other common datastores do not support renames which is what makes this FlatFS specific. If you are running a large custom datastore that supports renames you may want to consider running a fork of [fs-repo-11-to-12](https://github.com/ipfs/fs-repo-migrations/tree/master/fs-repo-11-to-12) specific to your datastore.
+
+If you want to disable this behavior, set the environment variable `IPFS_FS_MIGRATION_11_TO_12_ENABLE_FLATFS_FASTPATH` to `false`.
+
+####### Migration configuration
+
+For those who want to tune the migration more precisely for their setups, there are two environment variables to configure:
+
+- `IPFS_FS_MIGRATION_11_TO_12_NWORKERS` : an integer describing the number of migration workers - defaults to 1
+- `IPFS_FS_MIGRATION_11_TO_12_SYNC_SIZE_BYTES` : an integer describing the number of bytes after which migration workers will sync - defaults to 104857600 (i.e. 100MiB)
+
+###### Migration caveats
+
+Large repositories with very large numbers of CIDv1s should be mindful of the migration process:
+
+* We recommend ensuring that IPFS runs with an appropriate (high) file-descriptor limit, particularly when Badger is use as datastore backend. Badger is known to open many tables when experiencing a high number of writes, which may trigger "too many files open" type of errors during the migrations. If this happens, the migration can be retried with a higher FD limit (see below).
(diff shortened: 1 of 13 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,90 @@
...
+
+###### How migration works
+
+In order to perform the switch, and start referencing all blocks by solely by their multihash, a migration will occur on update. This migration will take the repository version from 11 (current) to 12.
+
+One thing to note is that any content addressed CIDv0 (all the hashes that start with `Qm...`, the current default in go-ipfs), does not need any migration, as CIDv0 are raw multihashes already. This means the migration will be very lightweight for the majority of users.
+
+On the other side, migration process will take care of re-keying any CIDv1 block so that it is only addressed by its multihash. Large nodes will lots of CIDv1-addressed content will need to go through a heavier process as the migration happens. This is how the migration works:
```
guseggert, line 33:
```suggestion
The migration process will take care of re-keying any CIDv1 block so that it is only addressed by its multihash. Large nodes with lots of CIDv1-addressed content will need to go through a heavier process as the migration happens. This is how the migration works:
```

CHANGELOG.md
```diff
@@ -1,5 +1,90 @@
...
+
+1. Phase 1: The migration script will perform a pass for every block in the datastore and will add all the CIDv1s found in it to a file named `11-to-12-cids.txt`, in the go-ipfs configuration folder. Nothing is written in this first phase and it only serves to identify keys will be migrated in phase 2.
+2. Phase 2: The migration script will perform a second pass where every CIDv1 block will read and written with its raw-multihash as key. There is 1 worker performing this task, although more can be configured. Every 100MiB-worth of blocks (this is configurable), each worker will trigger a datastore "sync" (to ensure all written data is flushed to disk) and delete the CIDv1-addressed blocks that were just renamed. This provides a good compromise between speed and resources needed to run the migration.
```
guseggert, line 36:
```suggestion
2. Phase 2: The migration script will perform a second pass where every CIDv1 block will be read and re-written with its raw-multihash as key. There is 1 worker performing this task, although more can be configured. Every 100MiB-worth of blocks (this is configurable), each worker will trigger a datastore "sync" (to ensure all written data is flushed to disk) and delete the CIDv1-addressed blocks that were just renamed. This provides a good compromise between speed and resources needed to run the migration.
```

CHANGELOG.md
```diff
@@ -1,5 +1,90 @@
...
+
+During the migration, the user will see log messages informing of how many blocks need to be rewritten and how far the process at every sync.
```
guseggert, line 38:
```suggestion
At every sync, the migration emits a log message showing how many blocks need to be rewritten and how far the process is.
```

CHANGELOG.md
```diff
@@ -1,5 +1,90 @@
...
+
+####### FlatFS specific migration
+
+For those using a single FlatFS datastore as their backing blockstore (i.e. the default behavior) migration (but not reversion) will take advantage of the ability to easily move/rename the blocks to improve migration performance.
```
guseggert, line 42:
```suggestion
For those using a single FlatFS datastore as their backing blockstore (i.e. the default behavior), the migration (but not reversion) will take advantage of the ability to easily move/rename the blocks to improve migration performance.
```

CHANGELOG.md
```diff
@@ -1,5 +1,90 @@
...
+
+Unfortunately, other common datastores do not support renames which is what makes this FlatFS specific. If you are running a large custom datastore that supports renames you may want to consider running a fork of [fs-repo-11-to-12](https://github.com/ipfs/fs-repo-migrations/tree/master/fs-repo-11-to-12) specific to your datastore.
+
+If you want to disable this ability set the environment variable `IPFS_FS_MIGRATION_11_TO_12_ENABLE_FLATFS_FASTPATH` to `false`.
```
guseggert, line 46:
```suggestion
If you want to disable this behavior, set the environment variable `IPFS_FS_MIGRATION_11_TO_12_ENABLE_FLATFS_FASTPATH` to `false`.
```

CHANGELOG.md
```diff
@@ -1,5 +1,90 @@
...
+
+####### Migration configuration
+
+For those who want to tune the migration more precisely for their setups there are two environment variables to configure:
```
guseggert, line 50:
```suggestion
For those who want to tune the migration more precisely for their setups, there are two environment variables to configure:
```

CHANGELOG.md
```diff
@@ -1,5 +1,90 @@
...
+- `IPFS_FS_MIGRATION_11_TO_12_SYNC_SIZE_BYTES` : an integer describing the number of bytes after which migration workers will sync - defaults to 104857600 (i.e. 100MiB)
+
+###### Migration caveats
+
+Large repositories with very large numbers of CIDv1s should be mindful of the migration process:
+
+* We recommend ensuring that IPFS runs with an appropriate (high) file-descriptor limit, particularly when Badger is use as datastore backend. Badger is known to open many tables when experiencing a high number of writes, which may trigger "too many files open" type of errors during the migrations. If this happens, the migration can be retried with a higher FD limit (see below).
+* Migrations using the badger datastore may not immediately reclaim the space freed by the deletion of migrated blocks, thus space requirements may grow considerably. A periodic badger-GC is run every 2 minutes, and this will be the moment when space used by deleted and de-duplicated blocks will be reclaimed. The last portion of the space will only be reclaimed after go-ipfs starts (the badger-GC cycle will trigger after 15 minutes).
```
guseggert, line 60:
```suggestion
* Migrations using the Badger datastore may not immediately reclaim the space freed by the deletion of migrated blocks, thus space requirements may grow considerably. A periodic Badger-GC is run every 2 minutes, which will reclaim space used by deleted and de-duplicated blocks. The last portion of the space will only be reclaimed after go-ipfs starts (the Badger-GC cycle will trigger after 15 minutes).
```

CHANGELOG.md
```diff
@@ -1,5 +1,90 @@
...
+* While there is a revert process (detailed below), we recommend keeping a backup of the repository, particularly for very large ones, in case an issue happens, so that revert can happen immediately and cases of repository corruption due to crashes or unexpected circumstances (mostly badger, again) are not catastrophic.
```
guseggert, line 61:
```suggestion
* While there is a revert process detailed below, we recommend keeping a backup of the repository, particularly for very large ones, in case an issue happens, so that the revert can happen immediately and cases of repository corruption due to crashes or unexpected circumstances are not catastrophic.
```

CHANGELOG.md
```diff
@@ -1,5 +1,90 @@
...
+###### Migration interruptions and retries
+
+If a problem occurred during the migration, it is be possible to simply re-start and retry it:
+
+1. Phase 1 will never overwrite the `11-to-12-cids.txt` file, but only append to it (so that a list of things we were supposed to have migrated during our first attempt is not lost - this is important for reverts, see below).
+2. Phase 2 will proceed to continue re-keying blocks that were not re-keyed during previous attempts.
+
+###### Migration reverts
+
+It will be also possible to revert the migration after it has succeeded, for example to go to a previous go-ipfs version (<=[0.11](https://github.com/ipfs/go-ipfs/releases/tag/v0.11.0)), even after starting and using go-ipfs in the new version (>=0.12). The revert process works as follows:
```
guseggert, line 72:
```suggestion
It is also possible to revert the migration after it has succeeded, for example to go to a previous go-ipfs version (<=[0.11](https://github.com/ipfs/go-ipfs/releases/tag/v0.11.0)), even after starting and using go-ipfs in the new version (>=0.12). The revert process works as follows:
```

CHANGELOG.md
```diff
@@ -1,5 +1,90 @@
...
+
+1. The `11-to-12-cids.txt` file, which has the list of all the CIDv1s that had to be rewritten for the migration is read.
```
guseggert, line 74:
```suggestion
1. The `11-to-12-cids.txt` file is read, which has the list of all the CIDv1s that had to be rewritten for the migration.
```

CHANGELOG.md
```diff
@@ -1,5 +1,90 @@
...
+2. A CIDv1-addressed block is written for every item on the list. This work is performed by 1 worker (configurable), syncing every 100MiB (configurable).
+3. Additionally, it is ensured that every CIDv1 pin, and every CIDv1 reference in MFS are also written as CIDV1-addressed blocks, regardless of whether they were part of the original migration or were added later.
+
+The revert process does not delete any blocks, only makes sure that blocks that were accessible with CIDv1s before the migration are again keyed with CIDv1s. This may result in a datastore becoming twice as large (i.e. if all the blocks were CIDv1-addressed before the migration). This is however done this way to cover corner cases: user can add CIDv1s after migration, which may reference blocks that existed as CIDv0 before migration. The revert heavily aims to ensure that no data becomes unavailable on downgrade.
```
guseggert, line 78:
```suggestion
The revert process does not delete any blocks--it only makes sure that blocks that were accessible with CIDv1s before the migration are again keyed with CIDv1s. This may result in a datastore becoming twice as large (i.e. if all the blocks were CIDv1-addressed before the migration). This is however done this way to cover corner cases: user can add CIDv1s after migration, which may reference blocks that existed as CIDv0 before migration. The revert aims to ensure that no data becomes unavailable on downgrade.
```

CHANGELOG.md
```diff
@@ -1,5 +1,90 @@
...
+
+While go-ipfs will auto-run the migration for you, it will not run the reversion. To do so you can download the [latest migration binary](https://dist.ipfs.io/fs-repo-11-to-12) or use [ipfs-update](https://dist.ipfs.io/#ipfs-update).
+
+###### Custom datastores
+
+As with previous migrations if you work with custom datastores and want to leverage the migration you can run a fork of [fs-repo-11-to-12](https://github.com/ipfs/fs-repo-migrations/tree/master/fs-repo-11-to-12) specific to your datastore. The repo includes instructions on building for different datastores.
+
+For this migration if your datastore has fast renames you may want to consider writing some code to leverage the particular efficiencies of your datastore similar to what was done for FlatFS.
```
guseggert, line 86:
```suggestion
For this migration, if your datastore has fast renames you may want to consider writing some code to leverage the particular efficiencies of your datastore similar to what was done for FlatFS.
```

CHANGELOG.md
```diff
@@ -1,5 +1,90 @@
...
+
+###### Rationale
+
+The blockstore/datastore layers are not concerned with data interpretation, only with storage of binary blocks and verification that the Multihash they are addressed with (which comes from the CID), matches the block. In fact, different CIDs, with different codecs prefixes, may be carrying the same multihash, and referencing the same block. Carrying the CID abstraction so low on the stack meant potentially fetching and storing the same blocks multiple times just because they are referenced by the different CIDs. For example, a CIDv1 with `dag-cbor` codec and a CIDv1 with `raw` codec containing the same multihash would result in two identical blocks stored. A CIDv0 and CIDv1 both being the same `dag-pb` block would also result in two copies.
```
guseggert, line 25:
```suggestion
The blockstore/datastore layers are not concerned with data interpretation, only with storage of binary blocks and verification that the Multihash they are addressed with (which comes from the CID), matches the block. In fact, different CIDs, with different codecs prefixes, may be carrying the same multihash, and referencing the same block. Carrying the CID abstraction so low on the stack means potentially fetching and storing the same blocks multiple times just because they are referenced by different CIDs. Prior to this change, a CIDv1 with a `dag-cbor` codec and a CIDv1 with a `raw` codec, both containing the same multihash, would result in two identical blocks stored. A CIDv0 and CIDv1 both being the same `dag-pb` block would also result in two copies.
```

guseggert:
LGTM!

PR #8749, opened 2022-02-24
fix(fsrepo): deep merge when setting config

### LINKED ISSUE #7482: Cannot "get" values omitted from the config.
#### Version information:

> v0.5.1

#### Description:

`ipfs config --json Foo.Bar` fails when `Bar` is specified as "omitempty". Instead of "getting" values from the config by traversing a map, we should use reflection.

Worse, this means `ipfs config --json Foo.Bar null` successfully sets `Foo.Bar` to null, but then returns an error to the user because it can't get the value of `Foo.bar`.

### LINKED ISSUE #8088: `ipfs config --json <key> false` returns error

kylehuntsman:
## The Problem
Key retrieval for config keys that were flagged with the `omitempty` json encoding option would throw an error when setting/getting those keys when the values were considered empty.
```
$ ipfs config --json Swarm.DisableRelay true
$ ipfs config --json Swarm.DisableRelay false
Error: failed to get config value: "Swarm key has no attributes"
```
When a config key is set, a [config struct is being merged with a map read from disk](https://github.com/ipfs/go-ipfs/blob/v0.12.0/repo/fsrepo/fsrepo.go#L551-L564) to ensure user provided keys are not removed in the process of trying to write a valid config file. The merging implementation is only updating root keys, overwriting any child values in the map read from disk with the values in the map from the config struct. The problem with this being that the marshalling of a config struct to a map omits empty values. Upon retrieval, the key does not exist because the merge removes it.

From the [`encoding/json`](https://pkg.go.dev/encoding/json#Marshal) package in the Go docs: 
> The "omitempty" option specifies that the field should be omitted from the encoding if the field has an empty value, defined as false, 0, a nil pointer, a nil interface value, and any empty array, slice, map, or string.

## The Solution
The merge when setting a config now does a deep merge of nested map properties. This ensures that nested properties that are flagged as 'omitempty' in the map created from the config struct are not touched, while all other values can be updated.

* Fixes #8088 
* Fixes #7482

## Additionally...
This PR includes a commit that improves the error response when a key is not found. Initially discussed in #8628, but ultimately including here.

### DIFF
--- a/repo/common/common_test.go
+++ b/repo/common/common_test.go
@@ -1,0 +1,132 @@
+package common
+
+import (
+	"testing"
+
+	"github.com/ipfs/go-ipfs/thirdparty/assert"
+)
+
+func TestMapMergeDeepReturnsNew(t *testing.T) {
+	leftMap := make(map[string]interface{})
+	leftMap["A"] = "Hello World"
+
+	rightMap := make(map[string]interface{})
+	rightMap["A"] = "Foo"
+
+	MapMergeDeep(leftMap, rightMap)
+
+	assert.True(leftMap["A"] == "Hello World", t, "MapMergeDeep should return a new map instance")
+}
+
+func TestMapMergeDeepNewKey(t *testing.T) {
+	leftMap := make(map[string]interface{})
+	leftMap["A"] = "Hello World"
+	/*
+		leftMap
+		{
+			A: "Hello World"
+		}
+	*/
+
+	rightMap := make(map[string]interface{})
+	rightMap["B"] = "Bar"
+	/*
+		rightMap
+		{
+			B: "Bar"
+		}
+	*/
+
+	result := MapMergeDeep(leftMap, rightMap)
+	/*
+		expected
+		{
+			A: "Hello World"
+			B: "Bar"
+		}
+	*/
+
+	assert.True(result["B"] == "Bar", t, "New keys in right map should exist in resulting map")
+}
+
+func TestMapMergeDeepRecursesOnMaps(t *testing.T) {
+	leftMapA := make(map[string]interface{})
+	leftMapA["B"] = "A value!"
+	leftMapA["C"] = "Another value!"
+
+	leftMap := make(map[string]interface{})
+	leftMap["A"] = leftMapA
+	/*
+		leftMap
+		{
+			A: {
+				B: "A value!"
+				C: "Another value!"
+			}
+		}
+	*/
+
+	rightMapA := make(map[string]interface{})
+	rightMapA["C"] = "A different value!"
+
+	rightMap := make(map[string]interface{})
+	rightMap["A"] = rightMapA
+	/*
+		rightMap
+		{
+			A: {
+				C: "A different value!"
+			}
+		}
+	*/
+
+	result := MapMergeDeep(leftMap, rightMap)
+	/*
+		expected
+		{
+			A: {
+				B: "A value!"
+				C: "A different value!"
+			}
+		}
+	*/
+
+	resultA := result["A"].(map[string]interface{})
+	assert.True(resultA["B"] == "A value!", t, "Unaltered values should not change")
+	assert.True(resultA["C"] == "A different value!", t, "Nested values should be altered")
+}
+
+func TestMapMergeDeepRightNotAMap(t *testing.T) {
+	leftMapA := make(map[string]interface{})
+	leftMapA["B"] = "A value!"
+
+	leftMap := make(map[string]interface{})
+	leftMap["A"] = leftMapA
+	/*
+		origMap
+		{
+			A: {
+				B: "A value!"
+			}
+		}
+	*/
+
+	rightMap := make(map[string]interface{})
+	rightMap["A"] = "Not a map!"
+	/*
+		newMap
+		{
+			A: "Not a map!"
+		}
+	*/
+
+	result := MapMergeDeep(leftMap, rightMap)
+	/*
+		expected
+		{
+			A: "Not a map!"
+		}
+	*/
+
+	assert.True(result["A"] == "Not a map!", t, "Right values that are not a map should be set on the result")
+}
--- a/repo/fsrepo/fsrepo.go
+++ b/repo/fsrepo/fsrepo.go
@@ -559,10 +559,8 @@ func (r *FSRepo) setConfigUnsynced(updated *config.Config) error {
 	if err != nil {
 		return err
 	}
-	for k, v := range m {
-		mapconf[k] = v
-	}
-	if err := serialize.WriteConfigFile(configFilename, mapconf); err != nil {
+	mergedMap := common.MapMergeDeep(mapconf, m)
+	if err := serialize.WriteConfigFile(configFilename, mergedMap); err != nil {
 		return err
 	}
 	// Do not use `*r.config = ...`. This will modify the *shared* config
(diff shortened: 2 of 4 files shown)


kylehuntsman:
So I see some of the sharness tests failing. I wasn't quite sure how to run those or where to put new tests, but now that I see them failing, I found where to look. I'll work on getting those to pass and amend a commit.

BigLep:
@kylehuntsman : I moved this back to draft for now.  Feel free to publish again when tests are passing.  I haven't looked closely at this, but my guess is we should use @schomatis as the reviewer.

repo/fsrepo/fsrepo.go
```diff
@@ -595,11 +596,43 @@ func (r *FSRepo) GetConfigKey(key string) (interface{}, error) {
...
+	if err != nil {
 		return nil, err
 	}
-	return common.MapGetKV(cfg, key)
+
+	// Walk the Config instance looking for key fields
+	fieldValue := reflect.ValueOf(cfg)
+	parts := strings.Split(key, ".")
+	for i, part := range parts {
+		fieldValue = reflect.Indirect(fieldValue).FieldByName(part)
```
schomatis, line 620: This is making the sharness fail because we can't access maps like this:

Calling it with `Datastore.Spec.child.path`:
```
panic: reflect: call of reflect.Value.FieldByName on map Value

goroutine 1 [running]:
reflect.flag.mustBe(...)
	/usr/local/go/src/reflect/value.go:221
reflect.Value.FieldByName(0x1f10540, 0xc000c10478, 0x195, 0x7ffefcde6d28, 0x5, 0x1f10540, 0xc000c10478, 0x195)
	/usr/local/go/src/reflect/value.go:903 +0x25a
github.com/ipfs/go-ipfs/repo/fsrepo.(*FSRepo).GetConfigKey(0xc0002bbb00, 0x7ffefcde6d19, 0x19, 0x0, 0x0, 0x0, 0x0)
	src/github.com/ipfs/go-ipfs/repo/fsrepo/fsrepo.go:621 +0x387
```

repo/fsrepo/fsrepo.go
```diff
@@ -595,11 +596,43 @@ func (r *FSRepo) GetConfigKey(key string) (interface{}, error) {
...
+
+		// Break out if the current field is not a defined key in Config
+		if !fieldValue.IsValid() {
```
schomatis, line 623: I'm not sure what is being tested here. When would a decoded field here not be valid? AFAIK (which is not that much) you can only get an invalid `reflect.Value` from _within_ reflect itself by some weird manipulation. See https://stackoverflow.com/a/39011500 for nice examples.

If this is trying to check if there's an empty/nil value here because we didn't pick up anything from the JSON, then you might want to go with something like https://stackoverflow.com/a/23555352.

schomatis (changes requested):
I'm not super sold on this solution but will go with it if we fix the reflection manipulations and the tests pass.

Still, let me give one last plug to what I described (very poorly) in https://github.com/ipfs/go-ipfs/issues/8088#issuecomment-1002667669: we should be reading/writing generic JSON nested maps in the config file to avoid the weird omit-empty dynamic, never writing encoded structs to the file. We should decode into the `go-ipfs` config structure only to validate the generic map is correct before writing it (to make sure when we read it back we can safely decode it and feed it to the daemon).

kylehuntsman:
Given the context that I now have after diving into this further, I understand what you are saying. Your description here also does make it a little more clear as to what you were talking about. 😅

I had this big write up I was going to post about the weirdness with some of the config types that I was having after retrieving them via reflect, and how I was proposing to solve those problems, specifically the `Strings` type. It would have been a brittle solution with control flows for that type specifically called out though, but it would have worked. 

I see the generic map solution being easier to reason through by someone with less context. It's also more similar to the existing logic than the reflect logic. I'll give it a shot. I'll set the reflect solution aside and see if I can get something working with generic maps as you described.

kylehuntsman:
I pushed a commit that reads and writes maps. I see something is failing regarding the peer config, so I'm going to look into that.

kylehuntsman:
Okay, so I think I got a little zealous with my changes thinking the duplicated read writes were unimportant. Apparently duplicating the read and writes in `setConfigUnsynced` actually does something important when setting a key, causing the output of the entire config file to be different. Haven't quite nailed down what that is yet. I pushed a minimal commit that only alters the merging function. I confirmed locally that the peer config tests pass, and a test expecting a failure setting an automatic sharding config I got to set to expect success! 😄

repo/fsrepo/fsrepo.go
kylehuntsman: Solution direction changed, resolving conservation.

repo/fsrepo/fsrepo.go
kylehuntsman: Solution direction changed, resolving conservation.

schomatis:
This is in my list for this week.

kylehuntsman:
Included error message improvement implementation originally discussed in #8628. Should I squash this into the original commit, or leave it as a separate one?

repo/common/common_test.go
```diff
@@ -0,0 +1,132 @@
...
+	result := MapMergeDeep(leftMap, rightMap)
+	/*
+		expected
+		{
+			A: {
+				B: "A value!"
+				C: "A different value!"
+			}
+		}
+	*/
```
schomatis, line 92: We should be comparing the entire maps with `require.Equal`, that way the comment  actually becomes the test, and we don't need specific equality checks.

schomatis (commented):
This goes in the right direction and we can approve and merge as is if needed, but let me suggest an additional change that is not the responsibility of this PR but addresses old technical debt that would make me very happy.

We're still going back and forth between the JSON maps and the config structs unnecessarily. I still don't understand the use of the undocumented `setConfigUnsynced` function; it forces `SetConfigKey` to convert the map to the struct just to be converted back again to merge the maps.

https://github.com/ipfs/go-ipfs/blob/d5ad847e05865e81957c43f526600860c06dbb84/repo/fsrepo/fsrepo.go#L644-L651

It even has a TODO to unroll that logic, which would make it even more apparent that we're doing the back and forth conversion. If we can't replace `setConfigUnsynced` entirely let's at least refactor it to avoid this.

kylehuntsman:
@schomatis I can take another stab at fixing that function. I tried originally and ran into some problems, so I backtracked and just implemented the bare minimum with the deep merge. If I spend too much time on it and don't get anywhere, we could make a new issue and merge this as is as suggested.

schomatis:
You can just unroll the function call there, remove the conversions (basically leaving just your deep merge) and call it a day if it gets too complicated.

schomatis:
@kylehuntsman Were you able to make progress with this one? If you are busy with other work feel free to re-assign to me to give the finishing touches. You already did the heavy lifting here, thanks.

kylehuntsman:
@schomatis Yeah, I thought I would have more time last week, but I guess not 😅 I've assigned back to you. Thanks.

schomatis:
Continuing in https://github.com/ipfs/go-ipfs/pull/8793.

schomatis:
(can't push to this branch)

PR #8750, opened 2022-02-25
fix prioritization of stream muxers

marten-seemann:
Looks like we've been prioritizing `mplex` over `yamux` since at least June 2020. This is really, really bad.

### DIFF
--- a/core/node/libp2p/libp2p.go
+++ b/core/node/libp2p/libp2p.go
@@ -58,7 +58,7 @@ type priorityOption struct {
 
 func prioritizeOptions(opts []priorityOption) libp2p.Option {
 	type popt struct {
-		priority int64
+		priority int64 // lower priority values mean higher priority
 		opt      libp2p.Option
 	}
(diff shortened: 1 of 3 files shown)


vyzo (approved):

Stebalien (approved):
Well, this is my new "most embarrassing mistake". I can confirm, and have no idea how we failed to notice. I guess most of the pprof profiles we've been looking at are from the gateways.

We should get this fix in, write a test.

guseggert:
Is there a way to add a test for this?

Stebalien:
> Is there a way to add a test for this?

At a minimum, we can write a unit test for the prioritization logic. But we probably need to export some additional stats in libp2p all the way through the swarm for better tests.

marten-seemann:
I added a unit test for the prioritization logic.

Stebalien (approved):

BigLep:
This should get called out in the release notes, basically that:

go-ipfs-new → go-ipfs-old|new will now use yamux instead of mplex.

Issue #8441, opened 2021-09-16
Export unixfs_get_latency_seconds as histogram


gmasgras:
### Checklist

- [X] My issue is specific & actionable.
- [X] I am not suggesting a protocol enhancement.
- [X] I have searched on the [issue tracker](https://github.com/ipfs/go-ipfs/issues?q=is%3Aissue) for my issue.

### Description

Merging https://github.com/ipfs/go-ipfs/pull/8332/ means that we can now track time to first block :tada: however only the `sum` and `count` metrics are exported. It would be very useful to also export this metric as a [histogram](https://prometheus.io/docs/practices/histograms/) (eg `unixfs_get_latency_seconds_bucket`) so that we can calculate percentiles (p90, p95) and get a better understanding of outliers. A good set of buckets could be 100ms, 500ms, 1s, 2s, 3s, 5s, 8s, 13s

thibmeu:
Regarding the set of buckets, what is the reasoning behind these values?
I feel like the granularity in the seconds unit is important, while there is very little information for blocks that are retrieved quickly.

I would suggest 50ms, 100ms, 250ms, 500ms, 1s, 2s, 5s, 10s. This is still arbitrary, but I feel it could capture more information.

gmasgras:
My reasoning for most of the values is that it's common to see buckets using the Fibonacci sequence. 30s and 60s are personal preference + they align with some of our timeouts on the gateways.

Issue #8234, opened 2021-06-30
Gateway support for /ipfs/{cid}?format=car|block|...


lidel:
>  @mikeal @olizilla @autonome  @warpfork @aschmahmann – this is a quick memo with my initial synthesis of the `?format=`  idea. Bit thin on details, but want to get early feedback/temperature check  before I start elaborating on this in gateway specs in the following weeks.

This is a meta issue for streamlining various feature requests and needs under a single opt-in query parameter that enables gateway users to fetch a specific representation of a specific content path.

Support for each format can be discussed/added via a separate issue/PR – this issue is just for tracking the bigger picture around unified `format` paramerer.

# MVP formats

## Ability to fetch every CID as full DAG in CAR or a single Block

This is the key  feature to enable [Verifiable Gateway Responses](https://github.com/ipfs/in-web-browsers/issues/128)  and "HTTP-based transport for IPFS"  (mobile browsers, IoT) without introducing even more dependency on `/api/v0`, and giving us flexibility for adding new features in the future.

- [ ] `?format=car`
  - Returns binary  `application/octet-stream` with CAR archive for entire DAG behind the content path
  - Supersedes  `/api/v0/dag/export`, but with better UX: 
    - works on DNSLink websites that do not expose `/api/v0`
    - `content-disposition` defaults to  `{filename|cid}.car` 
- [ ] `?format=block` 
  - Returns binary `application/octet-stream` with the root block identified by CID
  - Supersedes  `/api/v0/block/get`, but with better UX:
    -  works on DNSLink websites that do not expose `/api/v0`
    - `content-disposition` defaults to  `{cid}.block` 

## CBOR / JSON

The idea here is to unlock the power of IPLD by making non-dag-pb codecs available over gateway with built-in bi-directional conversion: a CID that is `dag-cbor` or `dag-json` could be fetched as either.

We  most likely want to do this in https://github.com/ipfs/go-ipfs/pull/8037 (refactor to use `?format=`)

- [ ] `?format=dag-cbor` 
  - Returns `dag-cbor` binary data as `application/cbor`
- [ ] `?format=dag-json`
  - Returns `dag-json` document as  `application/json`
  - This is especially useful for getting `dag-json` representation of `dag-cbor` document for use in web apps

[...]

PR #8443, opened 2021-09-17
feat: add gateway histogram metrics

aschmahmann:
fixes #8441

@gmasgras did you want both the summary and histogram exposed, or should I just switch the summary over to a histogram?

Any preference on the metric name? For example, if replacing the summary metric we may still want a different name so infra that's running the older version and the newer version can have the metrics play together a little nicer.

PR #8758, opened 2022-03-02
feat(gateway): Block and CAR response formats

lidel:
## Summary

This PR aims to add support for requesting alternative response format via: 
-  `?format=` URL paramerer 
-  `Accept: application/vnd.ipld.{format}` HTTP header

This MVP supports two formats:
- `raw` –  fetching single block 
- `car` – fetching entire DAG behind a CID as a [CARv1 stream](https://ipld.io/specs/transport/car/carv1/)

### Downloading a Block

```console
$ curl  -H 'Accept: application/vnd.ipld.raw' "http://127.0.0.1:8080/ipns/en.wikipedia-on-ipfs.org/favicon.ico" --output block.bin
$ cat block.bin | ipfs block put 
```

### Downloading a CAR

```console
$ curl  -H 'Accept: application/vnd.ipld.car' "http://127.0.0.1:8080/ipns/webui.ipfs.io" --output webui.car
$ ipfs dag import webui.car
```

Note: 
- We return `Content-Type: application/vnd.ipld.car; version=1`
- For now only a full DAG, we will add custom selectors later – tracked in https://github.com/ipfs/go-ipfs/issues/8769

### Why we need both

- Verifiable HTTP Gateway Responses (https://github.com/ipfs/in-web-browsers/issues/128) 
  - for mobile web browsers (content integrity without battery drain caused by full p2p)
  - for IoT devices and other thin clients (defaulting to HTTP, using p2p in LAN and as a fallback)
  - for hardened supply chain: we want to use CARs on CI (Github Actions), and as additional fallback logic for fetching fs-repo-migrations and similar updates
- unlock userland processing of IPLD data
  - users will be able to fetch arbitrary blocks, even ones in formats not supported by go-ipfs

### References

- "Gateway support for /ipfs/{cid}?format=car|block|... " https://github.com/ipfs/go-ipfs/issues/8234
- "Verifiable HTTP Gateway Responses" https://github.com/ipfs/in-web-browsers/issues/128
- "Make ipfs.dag.export built-in feature of HTTP gateways" https://github.com/ipfs/in-web-browsers/issues/170

## TODO

- [x] move file-specific logic to `serveFile`
  - [x] `Last-Modified` only when `Cache-Control` is missing
  - [x] `Cache-Control` for immutable content paths
  - [x] add `TODO` about `Last-Modified` when UnixFS 1.5 is supported
  - [x] run all tests on CI and fix any regressions
- [x] implement router for `?as=` (aka `?format=`)
- [x] implement `serveBlock`
  - [x] content-length
  - [x] content-disposition (as `attachment` + implicit filename of `{cid}.ipfs.block`)
  - [x] cache-control: same as for a file
  - [x] content-type: `application/vnd.*` – need decision in https://github.com/ipld/go-car/issues/238#issuecomment-1060877808
  - [x] etag
  - [ ] <del>TBD:  ability to test if block is present in local cache via `HEAD /ipfs/{cid}`  (local test for block, no network fetch)</del>
    - descoped, tracked in https://github.com/ipfs/go-ipfs/issues/8783
  - [x] tests - `t0117-gateway-block.sh`
- [x] implement `serveCar`
  - [x] <del>TBD (is content-length deterministic? blocks may have different order, but returning full DAG should always return same total of bytes)</del> too risky
  - [x] content-disposition (as `attachment` + implicit filename of `{cid}.ipfs.car`)
   - [x] content-disposition (as `attachment` + implicit filename of `{cid}.ipfs.car`)
  - [x] etag
    -  car streams for bigger dags won't be  byte-for-byte identical, so went with ["weak"  Etag](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag#directives) here
  - [x] cache-control: no-cache, no-transform for now 
    -  chunked stream could be interrupted  and browsers have handling due to trailers  being garbage https://github.com/mdn/browser-compat-data/issues/14703
    - interrupted stream includes `X-Stream-Error` so receiving end may be able to debug things, as long its not a browser
W/"<etag_value>"`  – it indicates that response can't be cached if byte-for-byte reproducibility is required.
  - [ ] <del>TBD: support `selector=multibase(cbor selector)` </del>
    - descoped from this PR,  `?format=car&selector=` is tracked in https://github.com/ipfs/go-ipfs/issues/8769
  - [x] content-type: `application/vnd.*` – need decision in https://github.com/ipld/go-car/issues/238#issuecomment-1060877808 + always include version in response + handle when processing request
  - [x] tests - `t0118-gateway-car.sh`
- [ ] <del>create more useful metrics, deprecate `unixfsGetMetric` </del>
  - descoped from this PR, improved metrics are tracked in https://github.com/ipfs/go-ipfs/issues/8441

## TODO (future PRs)

After this PR lands, we can add key IPLD formats (`serveDagCbor`, `serveDagJson`) – https://github.com/ipfs/in-web-browsers/issues/182


core/corehttp/gateway_handler.go
```diff
@@ -570,10 +584,86 @@ func (i *gatewayHandler) serveFile(w http.ResponseWriter, req *http.Request, nam
...
 	w.Header().Set("Content-Type", ctype)
 
+	// special fixup around redirects
 	w = &statusResponseWriter{w}
-	http.ServeContent(w, req, name, modtime, content)
+
+	http.ServeContent(w, r, name, modtime, content)
+}
+
+func (i *gatewayHandler) serveRawBlock(w http.ResponseWriter, r *http.Request, blockCid cid.Cid, contentPath ipath.Path) {
```
lidel, line 597: :information_source:  this func implements the poc for `/ipfs/{cid}?format=block` on gateways

- Unlike `/api/v0/block/get`, URLs are safe for pasting into a browser address bar (content-disposition, content-type)
- Supports range requests which benefit nodes with poor connectivity / data limits
  - Fetch of a bigger block can be resumed from specific byte

core/corehttp/gateway_handler.go
```diff
@@ -334,45 +341,31 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
...
-	// TODO: break this out when we split /ipfs /ipns routes.
-	modtime := time.Now()
+	// Handling Unixfs
+	dr, err := i.api.Unixfs().Get(r.Context(), resolvedPath)
+	if err != nil {
+		webError(w, "ipfs cat "+escapedURLPath, err, http.StatusNotFound)
+		return
+	}
+	// TODO:  do we want to reuse unixfsGetMetric for block/car, or should we have separate ones?
+	i.unixfsGetMetric.WithLabelValues(parsedPath.Namespace()).Observe(time.Since(begin).Seconds())
```
lidel, line 351: :question: Right now we only calculate time to first block for Unixfs.
Do we need separate metrics for `block` and `car`, or should I  add global `timeToFirstBlockGetMetric`?
@guseggert do you know how `unixfsGetMetric` is used in our infra?

lidel (commented):
Work in progress, dropping some notes so I don't forget.

core/corehttp/gateway_handler.go
```diff
@@ -570,10 +584,86 @@ func (i *gatewayHandler) serveFile(w http.ResponseWriter, req *http.Request, nam
...
+	// Set remaining headers
+	modtime := addCacheControlHeaders(w, r, contentPath, blockCid)
+	w.Header().Set("Content-Type", "application/octet-stream")
+	w.Header().Set("X-Content-Type-Options", "nosniff") // no funny business in the browsers :^)
+
+	// Done: http.ServeContent will take care of Content-Length and range requests
+	http.ServeContent(w, r, name, modtime, content)
+}
+
+func (i *gatewayHandler) serveCar(w http.ResponseWriter, r *http.Request, rootCid cid.Cid, contentPath ipath.Path) {
```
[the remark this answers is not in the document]
lidel: @olizilla I know you did some work on car clients and have better insight into *.storage work.
Any suggestions around ergonomics / direction here?

BigLep:
2022-03-03 discussion: transporting HTTP CARs around: currently sourcing how to handle errors.

Jorropo:
Personal notes for a future feature request:

Get parameters are uselessly hard to deal in many languages. Even tho most of them have escape features, you sometime manually concatenate them with the first concatenation being `?` and not `&`.

**TL;DR: I want an alternative be `Accept: application-x/car` header.**

That also allows to save up resending the header if you do multiple car requests in a row with HTTP2.

Jorropo:
> TBD (is content-length deterministic? blocks may have different order, but returning full DAG should always return same total of bytes)

FYI, Getting that info require a full root-traversal (you can actually skip raw-leaves but anyway that not on point).

### First CAR intricates:

In most cases, the info we trivially have is the `DagSize` stored in the root.
However this is not enough to know the size of the final car.
Assuming equal `DagSize` a car made of lots of small blocks is bigger than one made of few big blocks. Because CARv1 have a fix `varuint + CID` overhead *per block*, more blocks => more overhead even at equal dag sizes.

If we had `BlockCount` +  `DagSize` we can get a really good approximation.

I do say `approximation` because:
- Inlined blocks
- CID length
- Varuint size variability

Is variability you cannot account for without a traversal. Hopefully those are only small and doesn't change much. (Is it even ok to send wrong but close content lengths ?)

### Secondly, duped blocks

If we want to be smart we don't send duped blocks multiple time, however you don't know how much of your dag size is duped that can have massive impacts on the real content size.

Jorropo:
> Arguments being: not possible to create such request via browser's address bar, unable to create <a href= links, often being ignored by HTTP caches etc. etc.

I know, that why I think we should have both. :)

core/corehttp/gateway_handler.go
```diff
@@ -334,45 +325,55 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
...
-
-	if f, ok := dr.(files.File); ok {
-		if strings.HasPrefix(urlPath, ipfsPathPrefix) {
-			w.Header().Set("Cache-Control", "public, max-age=29030400, immutable")
-
-			// set modtime to a really long time ago, since files are immutable and should stay cached
-			modtime = time.Unix(1, 0)
+	// Support custom response formats passed via ?format or Accept HTTP header
+	if contentType := getExplicitContentType(r); contentType != "" {
+		switch contentType {
```
Jorropo, line 330:
```suggestion
		switch getExplicitContentType(r) {
		case "":
			break
```
Personally, I find that clearer.

Note (can't commit as is, you need to remove the tail `}` for the removed `if`)

core/corehttp/gateway_handler.go
```diff
@@ -334,45 +325,55 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
...
+			return
+		case "application/vnd.ipld.car":
+			logger.Debugw("serving car stream", "path", parsedPath)
+			i.serveCar(w, r, resolvedPath.Cid(), parsedPath)
+			return
+		case "application/vnd.ipld.car; version=1":
+			logger.Debugw("serving car stream", "path", parsedPath)
+			i.serveCar(w, r, resolvedPath.Cid(), parsedPath)
+			return
+		case "application/vnd.ipld.car; version=2": // no CARv2 in go-ipfs atm
```
Jorropo, line 343: I think this is a too simple matching thing.

I think we should support a more general `application/vnd.ipld.car[ \t]*;[ \t]*version[ \t]*=[ \t]*1` for a better developer experience:
Capture d’écran du 2022-03-08 21-45-59

(note the `[ \t]*` arround the `=` that might seems unneeded, however arround the `;` that seems important to me).
(also if they were `[ \t]?` (optional instead of optional or repeated) I would be fine too)

(also maybe there is a spec somewhere that standardize the white-space placement, in that case ignore this)

core/corehttp/gateway_handler.go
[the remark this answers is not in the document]
lidel: I went with the same behavior as in `/api/v0/dag/export` for now.
Selector support will be tracked separately in https://github.com/ipfs/go-ipfs/issues/8769

Jorropo (commented):
What kind of data is returned in the CAR response ?
I mean, let's assume I have 2 blocks:
- `Qmbar` a file.
- `Qmfoo` a folder with `a.txt` => `Qmbar`, `b.txt` => `Qmeee`.

If I download `Qmfoo/a.txt` do I get `Qmfoo` and `Qmbar` or only `Qmbar` ?
If the answer is the later, whats the plan for trustless gateway queries ? Use the block type to download `Qmfoo` and download `Qmbar` as car to speed up download once you reach recursive point ?

BigLep:
2022-03-10 conversation:
1. For reviewers, it may be best to start with the sharness tests
2. This won't get merged until 2022-03-14 at the earliest to give people a chance to communicate.

lidel:
@Jorropo  Good question! 

You get the CAR  for requested resource (full path).   In your example `/ipfs/Qmfoo/a.txt?format=car` gets resolved by the gateway to `/ipfs/Qmbar?format=car` and you get CAR for `Qmbar`

This is for convenience, but don't expect trustless clients to trust gateway to do the path resolution :)

> If the answer is the later, whats the plan for trustless gateway queries ? 
> Use the block type to download Qmfoo and download Qmbar as car to speed up download once you reach recursive point ?


Yes. The idea is that trustless gateway queries (made by light clients) will ask for top level CID roots (`/ipfs/{cid}`) only.
Client won't trust gateway with anything that is not verifiable. IPLD Data format deserialization and path traversal will happen on the client, after block CID is verified to be correct.

core/corehttp/gateway_handler_car.go
```diff
@@ -0,0 +1,59 @@
...
+	cid "github.com/ipfs/go-cid"
+	coreiface "github.com/ipfs/interface-go-ipfs-core"
+	ipath "github.com/ipfs/interface-go-ipfs-core/path"
+	gocar "github.com/ipld/go-car"
+	selectorparse "github.com/ipld/go-ipld-prime/traversal/selector/parse"
+)
+
+// serveCar returns a CAR stream for specific DAG+selector
+func (i *gatewayHandler) serveCar(w http.ResponseWriter, r *http.Request, rootCid cid.Cid, contentPath ipath.Path) {
+	ctx := r.Context()
```
Jorropo, line 17:
```suggestion
	ctx, cancel := context.WithCancel(r.Context())
	defer cancel()
```

This is because you are loosing gocar's context in the `dagstore` API wrapper:
```go
func (ds dagStore) Get(/* no context because of the API*/ c cid.Cid) (blocks.Block, error) {
```
Hopefully that will forward the cancel to dag API requests if the http write fail at some point.

Jorropo (commented):
Mostly good to me.
Things I feel strongly about:
- `defer cancel()`.
- move the `serveUnixfs` to it's own function.
Everything else is comments or questions, I'll let you decide on that.

core/corehttp/gateway_handler_car.go
```diff
@@ -0,0 +1,59 @@
...
+	// Set Content-Disposition
+	name := rootCid.String() + ".car"
+	setContentDispositionHeader(w, name, "attachment")
+
+	// Weak Etag W/ because we can't guarantee byte-for-byte identical  responses
+	// (CAR is streamed, blocks arrive from datastore in non-deterministic order)
+	w.Header().Set("Etag", `W/"`+rootCid.String()+`.car"`)
+
+	// Explicit Cache-Control to ensure fresh stream on retry.
+	// CAR stream could be interrupted, and client should be able to resume and get full response, not the truncated one
```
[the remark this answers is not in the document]
lidel: Not necessary, but will add `Accept-Ranges: none` just to make it clear we don't support range-requests for CAR stream responses.

I am bummed about caches too.  The underlying problem is this: a valid  CAR stream can be returned as HTTP 200 OK response, and die in the middle of the stream. The end will be missing, but HTTP status code remains 200 (OK).

There is no way to indicate to a browser client (JS code running on some page) that an error occurred (it already received HTTP status OK, and browsers [do not support trailer headers](https://github.com/mdn/browser-compat-data/issues/14703)).  If CDN/HTTP cache stores such truncated CAR response, then other clients will get broken responses too.

I was thinking about a way we could salvage HTTP caching  and here is an idea. We want CAR consumers to do validation anyway, and be able to fetch sub-dags via CAR+selector (https://github.com/ipfs/go-ipfs/issues/8769). 

Such a client will be smart enough to detect missing blocks anyway (during response validation), and request remaining blocks via new request(s). For such client, a truncated CAR response coming from a CDN/cache is not a hard error, it is just part of the daily business.

I feel this is way more robust setup than forcing CAR responses to be deterministic.

core/corehttp/gateway_handler_car.go
```diff
@@ -0,0 +1,59 @@
...
+func (i *gatewayHandler) serveCar(w http.ResponseWriter, r *http.Request, rootCid cid.Cid, contentPath ipath.Path) {
+	ctx := r.Context()
+
+	// Set Content-Disposition
+	name := rootCid.String() + ".car"
+	setContentDispositionHeader(w, name, "attachment")
+
+	// Weak Etag W/ because we can't guarantee byte-for-byte identical  responses
+	// (CAR is streamed, blocks arrive from datastore in non-deterministic order)
+	w.Header().Set("Etag", `W/"`+rootCid.String()+`.car"`)
```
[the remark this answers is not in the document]
lidel: CAR version is already returned in `Content-Type` header returned with CAR responses: `application/vnd.ipld.car; version=1`

core/corehttp/gateway_handler.go
```diff
@@ -795,6 +748,67 @@ func (i *gatewayHandler) addUserHeaders(w http.ResponseWriter) {
...
+	if contentPath.Mutable() {
+		// mutable namespaces such as /ipns/ can't be cached forever
+
+		/* For now we set Last-Modified to Now() to leverage caching heuristics built into modern browsers:
+		 * https://github.com/ipfs/go-ipfs/pull/8074#pullrequestreview-645196768
+		 * but we should not set it to fake values and use Cache-Control based on TTL instead */
+		modtime = time.Now()
+
+		// TODO: set Cache-Control based on TTL of IPNS/DNSLink: https://github.com/ipfs/go-ipfs/issues/1818#issuecomment-1015849462
+		// TODO: set Last-Modified if modification metadata is present in unixfs 1.5: https://github.com/ipfs/go-ipfs/issues/6920
```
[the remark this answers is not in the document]
lidel: Indeed, updated this TODO to hint at IPNS publishing rather than unixfs 1.5.
(we may end up not using this header, if there is little value cache-wise)

core/corehttp/gateway_handler.go
```diff
@@ -334,45 +325,48 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
...
-			w.Header().Set("Cache-Control", "public, max-age=29030400, immutable")
-
-			// set modtime to a really long time ago, since files are immutable and should stay cached
-			modtime = time.Unix(1, 0)
-		}
+	// Support custom response formats passed via ?format or Accept HTTP header
+	switch contentType := getExplicitContentType(r); contentType {
+	case "":
+		// nothing we should special-case, skip
+		break
```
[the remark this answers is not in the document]
lidel: Yep, started refactor in https://github.com/ipfs/go-ipfs/pull/8758/commits/76519e5040bc42b29771dc5701424e98b125da19

Jorropo (approved):
LGTM

core/corehttp/gateway_handler_car.go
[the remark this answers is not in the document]
Jorropo: I was hopping that using accept-ranges none would allow to allow caches because the clients would understand that half of the cache would still need to be fully downloaded but clients would still cache complete requests.

However from what you say that seems to be wrong, so I think this could have been dropped, I don't think it's really a bad thing for it to be present, but it's solving what I had in mind.

core/corehttp/gateway_handler.go
```diff
@@ -296,36 +297,26 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
 		return
 	}
 
-	dr, err := i.api.Unixfs().Get(r.Context(), resolvedPath)
-	if err != nil {
-		webError(w, "ipfs cat "+escapedURLPath, err, http.StatusNotFound)
+	// Finish early if client already has matching Etag
+	// (suffix match to cover both direct CID and DirIndex cases)
+	cidEtagSuffix := resolvedPath.Cid().String() + `"`
+	if strings.HasSuffix(r.Header.Get("If-None-Match"), cidEtagSuffix) {
```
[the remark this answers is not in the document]
lidel: I switched this to exact match in  26c122efc063284bb67799e7ec3f78077ef73c87. Better safe than sorry.

Etag is now based on CID and response format, and in the future, we will also include CID of a selector, if present.
This is mostly to avoid issues with poorly written/configured HTTP caches/CDNs.

core/corehttp/gateway_handler_car.go
```diff
@@ -22,8 +22,14 @@ func (i *gatewayHandler) serveCar(w http.ResponseWriter, r *http.Request, rootCi
 	setContentDispositionHeader(w, name, "attachment")
 
 	// Weak Etag W/ because we can't guarantee byte-for-byte identical  responses
-	// (CAR is streamed, blocks arrive from datastore in non-deterministic order)
-	w.Header().Set("Etag", `W/"`+rootCid.String()+`.car"`)
+	// (CAR is streamed, and in theory, blocks may arrive from datastore in non-deterministic order)
+	w.Header().Set("Etag", `W/`+getEtag(r, rootCid))
+
+	// Finish early if Etag match
+	if r.Header.Get("If-None-Match") == w.Header().Get("Etag") {
```
Jorropo, line 29:
```suggestion
	etag := `W/`+getEtag(r, rootCid)
	w.Header().Set("Etag", etag)
	
	// Finish early if Etag match
	if r.Header.Get("If-None-Match") == etag {
```

`w.Header().Set` then `w.Header().Get` is just like setting a key in a map just to fetch it right back after except GC is very likely not to optimise that away.

It is useleslly unclear just keep it locally.

Jorropo (changes requested):

core/corehttp/gateway_handler_car.go
lidel: Done: https://github.com/ipfs/go-ipfs/pull/8758/commits/aae5c28646b66585ad569d0589484e06fd33e1a8

Jorropo: @lidel are you sure about your fix ?
```go
	etag := getEtag(r, rootCid)
	w.Header().Set("Etag", `W/`+etag)

	// Finish early if Etag match
	if r.Header.Get("If-None-Match") == etag {
```
It is behaviourly different from your first patch.

Your first patch compares: `If-None-Match` with `Etag` from the response writter.
And it is set to `"W/"+etag`.
However here you compare it to `etag` only.

lidel: I think it is a clear signal I should stop typing get some sleep :see_no_evil:  5b43672a8733525ad9fab9be03d14d983f3f9914
You are right, thanks for catching this!

Jorropo: > thanks for catching this

The truth is **I made the same exact error**, except I've tried the code localy first because I thought maybe `w.Header().Get` would have extra properties like reading from the request headers or something (no it don't).
And have catched the error because the etags wasn't matching. 😄

Jorropo (approved):
LGTM

CodeQL is unhappy but that is safe because in fact the user controlled input is passed in `fmt`'s `%q` formater, which leaves us with fairly safe UTF-8.

lidel:
Related histogram metrics are added in a separate PR: #8443

PR #8756, opened 2022-03-02
Move go-ipfs-config back into go-ipfs

laurentsenta:
This PR moves the `go-ipfs-config` repo into `go-ipfs/config`.

Fixes [go-ipfs-config#151](https://github.com/ipfs/go-ipfs-config/issues/151).

Extra care was taken into preserving the history, steps to replicate are shared below:

## Notes

This branch looks daunting, but it's actually one merge commit plus 3 commits to update packages.
maybe running a `git diff go-ipfs/config go-ipfs-config` (to confirm the code is actually the same),
and reviewing the last commits only would make reviewers life easier.

```diff
› diff -r go-ipfs/config go-ipfs-config
Only in go-ipfs-config-vanilla: .git
Only in go-ipfs-config-vanilla: .github
Only in go-ipfs-config-vanilla: .gitignore
Only in go-ipfs-config-vanilla: .gx
Only in go-ipfs-config-vanilla: LICENSE
Only in go-ipfs-config-vanilla: README.md
Only in go-ipfs-config-vanilla: go.mod
Only in go-ipfs-config-vanilla: go.sum
diff -r go-ipfs-laurent/config/serialize/serialize.go go-ipfs-config-vanilla/serialize/serialize.go
11c11
<       "github.com/ipfs/go-ipfs/config"
---
>       "github.com/ipfs/go-ipfs-config"
diff -r go-ipfs-laurent/config/serialize/serialize_test.go go-ipfs-config-vanilla/serialize/serialize_test.go
8c8
<       config "github.com/ipfs/go-ipfs/config"
---
>       config "github.com/ipfs/go-ipfs-config"
Only in go-ipfs-config-vanilla: version.json
```

## How to replicate

(1) Rewrite the commit history with a go-ipfs-config prefix:

```bash
git filter-repo --commit-callback '
msg = commit.message.decode("utf-8")
newmsg = "go-ipfs-config: " + msg
commit.message = newmsg.encode("utf-8")
' --force
```

(2) Move the go-ipfs-config to a subfolder

```bash
cd go-ipfs-config
git filter-repo --to-subdirectory-filter config
```

(3) Remove non-go changes

```
git log --name-only ./config | grep '^config/' | grep -v 'go$' | sort | uniq > /tmp/remove
```

edited remove file:
```
config/.circleci/
config/.github/
config/.gitignore
config/.gx/
config/.travis.yml
config/LICENSE
config/README.md
config/go.mod
config/go.sum
config/package.json
config/version.json
```

[...]

PR #8771, opened 2022-03-08
Added experimental index provider feature to go-ipfs

bkyarger:
Two modes available with different flag, one to import all existing blocks and a second one for ongoing advertising of new blocks added.

No pinning is currently implemented, so no protection from GC at this time.  Lots of code borrowed from indexer-provider[1].

There are some hardcoded constants for controlling size of advertisements and chunks, these should probably be moved to configuration if this is to be supported long term.

[1] https://github.com/filecoin-project/index-provider

- [ ] research pinning style design suggestions
- [ ] move component init into core/node
- [ ] add indexProvider field to IpfsNode
- [ ] move to standard datastore
- [ ] move to built-in graphsync, deal with linksystem
- [ ] research dataTransfer context usage / behavior during shutdown
- [ ] research registration with indexer, usage of private keys, whether it is necessary
- [ ] use blockstore, not datastore
- [ ] fix log level on errCh not open
- [ ] use standard linksystem, put on IpfsNode if necessary
- [ ] add mutex to providerBlockStore for concurrency assurances
- [ ] deal with errors in PutMany of providerBlockStore
- [ ] fix order of operations and mutex on Delete of providerBlockStore


bkyarger:
Added fixes for issues from `make test_go_lint`

bkyarger:
Rebased on top of #8756 and fixed go-ipfs-config refs

cmd/ipfs/idxprovider.go
```diff
@@ -0,0 +1,442 @@
...
+	if err != nil {
+		providerlog.Errorf("Error generating advertisement link: %s", err)
+		return err
+	}
+
+	c := adLnk.ToCid()
+
+	// Store latest advertisement published from the chain
+	providerlog.Infow("Storing advertisement locally", "cid", c.String())
+	err = dataStore.Put(ctx, datastore.NewKey(latestAdvKey), c.Bytes())
```
aschmahmann, line 262: Note: If you wanted to keep your advertisements safe from GC it seems pretty straightforward to do. You just have to pass the root CID (as found from the `latestAdvKey`) into the list of roots that GC should traverse and mark as safe from GC because IIUC the advertisements are all linked together in a single DAG.

aschmahmann (changes requested):
@bkyarger I'm glad you were able to make some progress through here and put together a test demonstrating that things mostly work.

I added a bunch of code related comments as well as couple about design. The big design one above is mostly to give a basic info dump of what you might run into and some potential ways around it.

There's still work to do before this can be usable by upstream go-ipfs (mostly mentioned in the big design comment), however getting your fork up and running for your use case sooner seems pretty doable.

Great to see this coming along and happy to help where I can here, although you might need some folks more familiar with the indexer codebase to give review on the indexer specific logic and setup

BigLep:
@bkyarger : what are the next steps here?  Are you going to incorporate the feedback or can this be close in light of https://github.com/web3-storage/ipfs-elastic-provider

BigLep:
Closing this since didn't am not aware of anyone driving this forward and there hasn't been activity for months.

willscott:
Yep, instead we're doing this via reframe - a working reframe publisher to indexers for kubo is at https://github.com/filecoin-project/index-provider/pull/271

Jorropo:
> Yep, instead we're doing this via reframe - a working reframe publisher to indexers for kubo

IMO this is bad because it require doing diffing on the 24h republish that Kubo do.

We want to rework the advertising logic anyway because it's doing bad stuff (we have two different codepaths for providing and reprodividing, the first provide code buffer is crashy) and I think if we have a proper state tracker it would be easier.
Then if we already track the state of what is pinned it will be easy to publish to indexers without doing diffing.

willscott:
@Jorropo  It would be great to propose an update to the reframe provide path with better semantics you can think would be supportable by kubo: https://github.com/ipfs/specs/blob/main/reframe/REFRAME_KNOWN_METHODS.md#provide

Jorropo:
@willscott I don't want to start working on new reframe changes while the advertise refactor is not even planned yet. I have too much things to do fort now.

I have a personal project to make a simple plugin that read the pinned states and run an indexer server after we refactor the annoucement logic.
It would be more decentralized to publish on a pubsub channel than an HTTP endpoint. (even tho yes you could just run your own http endpoint but it's more work)

willscott:
we're getting a working version of this implementation with the expectation of publishers running their own endpoint they delegate their kubo reframe publishing to.
that setup allows for minimal risk and isn't hard to set up for large publishers (e.g. big cluster instances, big publishers)

Issue #8763, opened 2022-03-04
Allow / Deny Bitswap replication based on peerID and CID


laurentsenta:
### Checklist

- [X] My issue is specific & actionable.
- [X] I am not suggesting a protocol enhancement.
- [X] I have searched on the [issue tracker](https://github.com/ipfs/go-ipfs/issues?q=is%3Aissue) for my issue.

### Background

Right now, the only way to store personal/private data on the network is to:
- Use a private network (with [swarm keys](https://github.com/Kubuxu/go-ipfs-swarm-key-gen) for example),
- Use semantic security (by encrypting the data).

For some use cases, this is too restrictive:
"I wouldn't store my passwords or my medical records on the IPFS network, even encrypted."

For these use cases, decentralization might still be necessary, but they would require a permissioned approach:
ideally, we'd let a user define who can replicate their data / what can be replicated from their node.

There are a lot of discussions around the question of authorizations (see below), @kylehuntsman and I have been working on the smallest step forward we could find: let a node accept/reject a bitswap request based on peerID and CID.

(one might find a certain resemblance with libp2p's [ConnectionGater](https://pkg.go.dev/github.com/libp2p/go-libp2p-core/connmgr#ConnectionGater).)

### Stories / Plan

1. [ ] As a developer, I can define a filter predicate (policy). This lets me filter requests based on the requester's ID and the content requested.
   - [ ] Add an optional configuration to go-bitswap that lets a user define this peer-block filter,
   - [ ] Add ipfs-related plumbing to pass this configuration down to the bitswap instance.
2. [ ] As a regular user, I can configure a default policy with a few `allow`, `deny` commands, similar to the [IPFS filter](https://github.com/ipfs/go-ipfs/issues/8492) proposed for gateway operators.
   - [ ] TBD

### Current State

- [Bitswap PR](https://github.com/ipfs/go-bitswap/pull/549)
- [IPFS PR](https://github.com/ipfs/go-ipfs/pull/8772)

### Related Issues and discussions

- https://github.com/ipfs/specs/pull/270
- https://github.com/ipfs/go-ipfs/issues/7871
- https://github.com/ipfs/specs/issues/260
- https://github.com/ipfs/roadmap/issues/75
- https://github.com/ipfs/go-bitswap/issues/532
- https://github.com/ipfs/go-ipfs/issues/6842
- https://github.com/ipfs/go-ipfs/issues/8492
- https://github.com/ipfs/go-ipfs/issues/6097
[...]

PR #8772, opened 2022-03-08
feat: add peer block filter option

laurentsenta:
This feature lets a user configure a filter to allow / deny bitswap requests according to the request's peer ID and the content id. For example, a user may use this option to implement a dynamic peer-based authorization.

Contributes to https://github.com/ipfs/go-ipfs/issues/8763,
Relates to https://github.com/ipfs/go-bitswap/pull/549

This PR adds the plumbing required to pass the peer-block filter option from ipfs down to go-bitswap.

- [ ] confirm with the team the way we pass the option is the right approach,
- [ ] confirm with the team where we want to document this,
   - we could advertise this as an experimental feature,
   - provide an example / demo of the feature in the docs/examples folder (like in [this demo](https://github.com/laurentsenta/go-ipfs/tree/feat/peer-block-filter-with-demo/docs/examples/go-ipfs-as-a-library))

### DIFF
--- a/core/node/builder.go
+++ b/core/node/builder.go
@@ -8,6 +8,7 @@ import (
 
 	"go.uber.org/fx"
 
+	"github.com/ipfs/go-bitswap"
 	"github.com/ipfs/go-ipfs/core/node/helpers"
 	"github.com/ipfs/go-ipfs/core/node/libp2p"
 	"github.com/ipfs/go-ipfs/repo"
@@ -40,6 +41,9 @@ type BuildCfg struct {
 	Routing libp2p.RoutingOption
 	Host    libp2p.HostOption
 	Repo    repo.Repo
+
+	// Bitswap
+	PeerBlockRequestFilter bitswap.PeerBlockRequestFilter
 }
 
 func (cfg *BuildCfg) getOpt(key string) bool {
--- a/core/node/groups.go
+++ b/core/node/groups.go
@@ -281,9 +281,10 @@ func Online(bcfg *BuildCfg, cfg *config.Config) fx.Option {
 
 	/* don't provide from bitswap when the strategic provider service is active */
 	shouldBitswapProvide := !cfg.Experimental.StrategicProviding
+	pbrf := bcfg.PeerBlockRequestFilter
 
 	return fx.Options(
-		fx.Provide(OnlineExchange(cfg, shouldBitswapProvide)),
+		fx.Provide(OnlineExchange(cfg, shouldBitswapProvide, pbrf)),
 		maybeProvide(Graphsync, cfg.Experimental.GraphsyncEnabled),
 		fx.Provide(DNSResolver),
 		fx.Provide(Namesys(ipnsCacheSize)),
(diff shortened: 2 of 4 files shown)


core/node/builder.go
```diff
@@ -40,6 +41,9 @@ type BuildCfg struct {
 	Routing libp2p.RoutingOption
 	Host    libp2p.HostOption
 	Repo    repo.Repo
+
+	// Bitswap
+	PeerBlockRequestFilter bitswap.PeerBlockRequestFilter
```
laurentsenta, line 46: I'm not sure which we should prefer:
1. Pass the `PeerBlockRequestFilter` function, shown here,
2. Pass a generic `bitswap.Option` function.

(2) is more "obviously future-proof", but that'd mean we'd have two different ways to configure most values in the bitswap instance  (json from cfg and code from buildcfg).

Access control is probably a cross-cutting concern that will span multiple modules and protocols, not only bitswap. So by decoupling the access control config from the bitswap option, (1) might make it easier to build upon this feature.

core/node/groups.go
```diff
@@ -281,9 +281,10 @@ func Online(bcfg *BuildCfg, cfg *config.Config) fx.Option {
 
 	/* don't provide from bitswap when the strategic provider service is active */
 	shouldBitswapProvide := !cfg.Experimental.StrategicProviding
+	pbrf := bcfg.PeerBlockRequestFilter
 
 	return fx.Options(
-		fx.Provide(OnlineExchange(cfg, shouldBitswapProvide)),
+		fx.Provide(OnlineExchange(cfg, shouldBitswapProvide, pbrf)),
```
Jorropo, line 287: I belive you inspired your pbrf variable from the line above, this code (`shouldBitswapProvide`) is a clarification variable, it translate `!cfg.Experimental.StrategicProviding` into a usefull thing we understand.
In your case you are turning a perfectly good `PeerBlockRequestFilter` into a meaningless without prior knowledge `pbrf` one.
```suggestion
	return fx.Options(
		fx.Provide(OnlineExchange(cfg, shouldBitswapProvide, bcfg.PeerBlockRequestFilter)),
```

Jorropo (changes requested):
Your code doesn't build:
```
core/node/bitswap.go:27:60: undefined: bitswap.PeerBlockRequestFilter
core/node/bitswap.go:45:24: undefined: bitswap.WithPeerBlockRequestFilter
core/node/builder.go:46:25: undefined: bitswap.PeerBlockRequestFilter
```
I guess you are waiting on a new version of bitswap with your changes merged.
I don't know if maintainers have feelings about this, but I like to create a temporary replace rule that I let maintainers remove (with the "maintainers are allowed to push to this branch" or alike option).
You can do this by running:
```
go mod edit -replace <target repo>=<your fork>@<your commit>
```
For example:
```
go mod edit -replace github.com/alanshaw/go-carbites=github.com/Jorropo/go-carbites@751dcbf09c8a387030238e4bd3169cc2516046f2
```
You can also add a `// see <link to the blocking pull request>` comment on the replace line in the `go.mod`.

That allows the CI tests to run and shows that your code is RFM (blocking PR aside).

laurentsenta:
Closing this after a sync chat with Adin, this might come back as a plugin later.

Thanks for the initial review @Jorropo, I won't update the PR but I'll keep your feedback in mind for future PRs.

PR #8773, opened 2022-03-08
fix: listen on loopback for API and gateway ports in docker-compose.yaml

guseggert:


docker-compose.yaml
```diff
@@ -10,11 +10,18 @@ services:
...
       - 4001:4001/tcp
       - 4001:4001/udp
-      - 5001:5001
-      - 8080:8080
-      - 8081:8081
+      
+      # The following ports only listen on the loopback interface, so are not remotely reachable by default.
+      # If you want to override these or add more ports, see https://docs.docker.com/compose/extends/ .
+      
+      # API port, which includes admin operations, so you probably don't want this remotely accessible.
```
Jorropo, line 20: Just saying "you shouldn't do that" or alike isn't enough for people to not do it.
We need to propose alternative solutions.

In a future patch I'll add a link to docs explaining SSH tunnels and reverse proxy (*don't exists yet*).

Jorropo (approved):
+2

aschmahmann (approved):

PR #8786, opened 2022-03-11
feat: port collect-profiles.sh to 'ipfs diag profile'

guseggert:


bin/collect-profiles.sh
```diff
@@ -44,6 +44,17 @@ go tool pprof -symbolize=remote -svg -output mutex.svg "$SOURCE_URL/debug/pprof/
 echo "Disabling mutex profiling"
 curl -X POST "$SOURCE_URL"'/debug/pprof-mutex/?fraction=0'
 
+echo "Enabling block profiling"
```
aschmahmann, line 47: Should this kind of thing also be done in `ipfs diag profile`? I'd hope that this script now doesn't need to be used as frequently given we have a more agnostic tool built in.

guseggert: It definitely could, I'm just not sure what the criteria is for adding things into that command vs. `collect-profiles.sh`. We've always used the latter operationally which is why I added this to it, but it is also trivial to add to `ipfs diag profile`.  

Is the intention for `ipfs diag profile` to replace `collect-profiles.sh`?

aschmahmann: IMO yes, since it's cross platform and easier for us to version since the command is tied to the binary.

lidel (approved):
LGTM (tweaked helptext a tiny bit)

Before  we merge, it would be useful to get additional reviewe from devrel folks (@Jorropo) as `ipfs diag profile` will be something that we ask people to run while debugging issues.

Jorropo:
@lidel

I'm not sure what does this PR does, I belive it's mainly a refactor / shuffling code arround kind of things.

My issues of `ipfs diag profile` are:
1. CPU profiling often blocks for multiple minutes on high load scenario and `--profiling-time 0s` doesn't disable it.
2. It doesn't pipeline writing to the output file (as writing to the output file while it is profiling and instead of buffering then writing).
3. The output file is not packaged like most unix files (with a `profile-$(date)/` prefix on all paths) so unzipping it dump everything in `pwd` (altho that unix tools doing things the wrong way to start with)
4. The output file is not pipeline friendly.
5. The output file doesn't compress identical shared data between different files.

Point 1 can be fixed with a `if 0 == ` arround the CPU profiling.
Point 2 to 5 can be fixed by outputing `.tar.gz` files (pipeling is not just using .tar.gz but would help a lot).

I've actually being using the shell script instead (or just `curl`ing the API myself), but I can't recomend that to new people that are having issues because getting them to download the script and execute it, takes any saving I get from using a tar.

guseggert:
IIUC we didn't use `.tar.gz` for Windows compatibility. We could add flags for that but doesn't seem worth the effort IMO. This new code does stream output instead of buffering though (when possible...stack traces are still buffered internally by the stdlib).

Point 1 is already addressed in this PR, if `--profile-time=0` then CPU profiling is not run (and neither is any other sampling profile).

I think it would be nice though to control exactly what profiles are run though, which would be easy to add, so I'll add that.

lidel (approved):
- Keeping `.zip` for now is fine (windows interop, allows user to reason what they sent to a third-party)
- `--collectors` works as expected, thanks!

Let's merge this, a huge improvement when compared to `bin/collect-profiles.sh`

PR #8789, opened 2022-03-14
fix: incorrect logging

MichaelMure:


aschmahmann (approved):

core/corehttp/gateway_handler.go
```diff
@@ -854,7 +854,7 @@ func webError(w http.ResponseWriter, message string, err error, defaultCode int)
 func webErrorWithCode(w http.ResponseWriter, message string, err error, code int) {
 	http.Error(w, fmt.Sprintf("%s: %s", message, err), code)
 	if code >= 500 {
-		log.Warnf("server error: %s: %s", err)
+		log.Warnf("server error: %s: %s", message, err)
```
aschmahmann, line 857: The linter is complaining about this being unsanitized. IIUC these are all escaped URLs, right @lidel?

aschmahmann: @lidel, @guseggert, @MichaelMure if any of you get a chance to confirm the linter is off here please comment and anyone with merge access can come by and stamp it.

Thanks @MichaelMure for catching this 🙏.

MichaelMure: The linter is right:
- https://github.com/ipfs/go-ipfs/blob/cbb962852f914dc682887ea86c227e8f546c880c/core/corehttp/gateway_handler.go#L301
- https://github.com/ipfs/go-ipfs/blob/cbb962852f914dc682887ea86c227e8f546c880c/core/corehttp/gateway_handler.go#L200

aschmahmann: Does the path escaping from `EscapedPath` not help us out here, and we should be doing our own escaping?

MichaelMure: I can't say for sure, but this escaping looks more about encoding for a URL path rather than security.

lidel: - klinda false-positive: `webError` is returned as `text/plain` so it does not need to be protected against HTML injection (we don't have to wrap it in HTMLEscapeString)

lidel (commented):
Thanks! 
To avoid doing the same work twice, I fixed this as part of bigger refactor in  https://github.com/ipfs/go-ipfs/pull/8758 / https://github.com/ipfs/go-ipfs/pull/8758/commits/eb95d2b693ad3ccf028d3995b9b109307dae9f62 so for sake of keeping smaller review surface let's close this and focus on https://github.com/ipfs/go-ipfs/pull/8758

lidel:
Continued in https://github.com/ipfs/go-ipfs/pull/8758

PR #8793, opened 2022-03-15
fix(fsrepo): deep merge when setting config

### LINKED ISSUE #7482: Cannot "get" values omitted from the config.

### LINKED ISSUE #8088: `ipfs config --json <key> false` returns error

schomatis:
Continuation of @kylehuntsman's https://github.com/ipfs/go-ipfs/pull/8749 here since I can't seem to push to his branch.

* Fixes #8088 
* Fixes #7482

### DIFF
--- a/repo/common/common.go
+++ b/repo/common/common.go
@@ -21,7 +21,13 @@ func MapGetKV(v map[string]interface{}, key string) (interface{}, error) {
 
 		cursor, ok = mcursor[part]
 		if !ok {
-			return nil, fmt.Errorf("%s key has no attributes", sofar)
+			// Construct the current path traversed to print a nice error message
+			var path string
+			if len(sofar) > 0 {
+				path += sofar + "."
+			}
+			path += part
+			return nil, fmt.Errorf("%s not found", path)
 		}
 	}
 	return cursor, nil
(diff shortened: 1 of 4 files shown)


guseggert:
@schomatis looks like the sharness tests are broken by this.

schomatis:
Yes, was having issues with sharness locally and also couldn't get it remotely (see https://github.com/ipfs/go-ipfs/issues/8804#issuecomment-1072900558). Hang tight while I fix this, will explicitly request your review when done.

schomatis:
@guseggert Fixed, ready for review.

guseggert (approved):

PR #8802, opened 2022-03-18
feat(cmds): document deprecated RPC API commands

lidel:
go-ipfs part of  https://github.com/ipfs/go-ipfs-cmds/issues/93

This PR adds the programmatic Status annotation introduced in https://github.com/ipfs/go-ipfs-cmds/pull/225
and applies it to for already deprecated / experimental commands.

The main purpose of these annotations is to separate "stable|deprecated|experimental" commands when generating https://docs.ipfs.io/reference/http/api/ (implemented in https://github.com/ipfs/ipfs-docs/pull/1086), but we also show them in `--help` text in CLI:

### Deprecated `object` commands

```console
$ ipfs object patch --help
WARNING:   DEPRECATED, command will be removed in the future

USAGE
  ipfs object patch - Deprecated way to create a new merkledag object based on an existing one. Use MFS with 'files cp|rm' instead.
...
```

### Experimental  `name pubsub` commands

```console
$ ipfs name pubsub state --help
WARNING:   EXPERIMENTAL, command may change in future releases

USAGE
  ipfs name pubsub state - Query the state of IPNS pubsub.
....
```

### DIFF
--- a/core/commands/dag/dag.go
+++ b/core/commands/dag/dag.go
@@ -30,7 +30,7 @@ var DagCmd = &cmds.Command{
 		ShortDescription: `
 'ipfs dag' is used for creating and manipulating DAG objects/hierarchies.
 
-This subcommand is is intended to deprecate and replace
+This subcommand is intended to deprecate and replace
 the existing 'ipfs object' command moving forward.
 `,
 	},
--- a/core/commands/dag/dag.go
+++ b/core/commands/dag/dag.go
@@ -30,9 +30,9 @@ var DagCmd = &cmds.Command{
 		ShortDescription: `
 'ipfs dag' is used for creating and manipulating DAG objects/hierarchies.
 
-This subcommand is currently an experimental feature, but it is intended
-to deprecate and replace the existing 'ipfs object' command moving forward.
-		`,
+This subcommand is is intended to deprecate and replace
+the existing 'ipfs object' command moving forward.
+`,
 	},
 	Subcommands: map[string]*cmds.Command{
 		"put":     DagPutCmd,
--- a/core/commands/unixfs/ls.go
+++ b/core/commands/unixfs/ls.go
@@ -33,8 +33,9 @@ type LsOutput struct {
 }
 
 var LsCmd = &cmds.Command{
+	Status: cmds.Deprecated, // https://github.com/ipfs/go-ipfs/pull/7755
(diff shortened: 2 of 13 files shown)


core/commands/dag/dag.go
```diff
@@ -30,9 +30,9 @@ var DagCmd = &cmds.Command{
 		ShortDescription: `
 'ipfs dag' is used for creating and manipulating DAG objects/hierarchies.
 
-This subcommand is currently an experimental feature, but it is intended
-to deprecate and replace the existing 'ipfs object' command moving forward.
-		`,
+This subcommand is is intended to deprecate and replace
```
schomatis, line 33:
```suggestion
This subcommand is intended to deprecate and replace
```

schomatis, line 33:
```suggestion
This subcommand is intended to deprecate and replace
```

core/commands/unixfs/ls.go
```diff
@@ -33,8 +33,9 @@ type LsOutput struct {
 }
 
 var LsCmd = &cmds.Command{
+	Status: cmds.Deprecated, // https://github.com/ipfs/go-ipfs/pull/7755
 	Helptext: cmds.HelpText{
-		Tagline: "List directory contents for Unix filesystem objects. Deprecated: Use 'ipfs ls' instead.",
+		Tagline: "List directory contents for Unix filesystem objects. Deprecated: Use 'ipfs ls' and 'ipfs files ls' instead.",
```
schomatis, line 38: I just realized we have both an `ipfs files` _and_ an `ipfs file` (single) command. the horror

go.mod
```diff
@@ -31,7 +31,7 @@ require (
 	github.com/ipfs/go-graphsync v0.11.0
 	github.com/ipfs/go-ipfs-blockstore v1.1.3-0.20220317225625-b3ee1d940911
 	github.com/ipfs/go-ipfs-chunker v0.0.5
-	github.com/ipfs/go-ipfs-cmds v0.6.0
+	github.com/ipfs/go-ipfs-cmds v0.6.1-0.20220318221633-afaf21ab15e5
```
schomatis, line 34: I'm guessing this is going to be a tag later, right?

schomatis (approved):
I'm approving the use of the new `Status` field itself, not the statuses themselves being applied to the commands, which I have not enough context for.

PR #8757, opened 2022-03-02
Bubble ErrNotFound improvements

hsanjuan:
Closes #7074

PR #8803, opened 2022-03-18
ErrNotFound changes: bubble tagged libraries.

hsanjuan:
Additionally fixes tests for new error message which went from

`ipld: <cid> not found`

to

`<cid> not found`


hsanjuan:
Merging because I don't conceive there is anything to object here (given #8757 was merged)  and go.mod changes are prone to conflicts.

PR #8808, opened 2022-03-21
feat: re-enable docker sharness tests

guseggert:
The Docker sharness tests were disabled years ago when go-ipfs moved
from Travis to CircleCI. This makes the tweaks necessary to re-enable
them.

The Docker image has since moved to be based on BusyBox which doesn't
have the requisite wget version for the existing tests to work, so
this adds some functionality to the pollEndpoint program to support
polling HTTP endpoints as well.

### DIFF
--- a/.circleci/main.yml
+++ b/.circleci/main.yml
@@ -154,7 +154,7 @@ jobs:
     working_directory: ~/ipfs/go-ipfs
     environment:
         <<: *default_environment
-        TEST_NO_DOCKER: 1
+        TEST_NO_DOCKER: 0
         TEST_NO_FUSE: 1
         TEST_VERBOSE: 1
     steps:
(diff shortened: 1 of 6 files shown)


schomatis:
Sorry, I have no idea how docker fits in here. I've always run sharness directly from the CLI. I'll remove my review but feel free to nudge me some more and I'll try to dig up how all this works.

BigLep:
Hi @schomatis .  Thanks for the disclosure.  I am going to nudge though :). We want knowledge to disseminate more.  Feel free to ask in #ipfs-dev too.  @guseggert can jump in and provide more context.

schomatis (approved):
The extension of the pollEndpoint makes sense to replace wget. Other docker sharness specifics escape me.

PR #8815, opened 2022-03-22
Update Go version to 1.17

aschmahmann:
Note: this has been scoped to just support 1.17.


plugin/plugins/Rules.mk
```diff
@@ -1,6 +1,6 @@
 include mk/header.mk
 
-$(d)_plugins:=$(d)/git $(d)/dagjose $(d)/badgerds $(d)/flatfs $(d)/levelds $(d)/peerlog
+$(d)_plugins:=
```
aschmahmann, line 3: All of the plugins have been removed here due to #8550 (i.e. old xxhash dependencies).

If we upgrade the version of go-libp2p-quic-transport and prometheus/client_golang we should be able to bring back some of the building here as a test mechanism to make sure they can build.

BigLep:
2022-04-07 conversation: 
1. We will wait for 1.17.9 (https://groups.google.com/g/golang-announce/c/vtbMjE04kPk/m/xE-FGxCXCAAJ ) and then update this PR
2. We're going to merge this and cherry pick this as needed
3. We believe we only have to have to wait for Docker
4. We need to get Bifrost prepared here

rvagg (approved):

PR #8816, opened 2022-03-22
feat(gateway): _redirects file support

justincjohnson:
Rebased Cliff's changes with some upstream gateway refactorings.  Will retest from here and then continue the work.


lidel (commented):
Thanks for picking this up! :heart: 

Quick feedback:
- this should be only applied to default unixfs response types, and ignored by other ones (see comment inline)
-  this should be enabled only when we have Origin isolation 
   - (when URL.pathname `/` starts at the IPFS content root `/ipfs/{cid}/` and cant go beyond it) 
    - namely, when DNSLink website is loaded based on `Host` header  or if `Host` is a subdomain gateway

core/corehttp/gateway_handler.go
```diff
@@ -316,9 +317,25 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
 		}
 	}
 
-	contentPath := ipath.New(r.URL.Path)
+	redirects, err := i.searchUpTreeForRedirects(r, urlPath)
```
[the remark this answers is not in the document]
justincjohnson: @lidel should block/car requests continue to return 404s if not present?  I'm assuming so, and if true, I think we're going to have issues moving redirect logic into `gateway_handler_unixfs.go` since the 404 logic kicks in before we have a chance to read the `_redirects` file and we return 404 before having a chance to redirect.

I can think more about this but wanted to confirm I'm thinking clearly, since this code is quite new to me.  Thanks.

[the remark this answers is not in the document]
lidel: @justincjohnson No, block/car responses should not have any special magic to them.  Apply 404 / _redirect magic only to default unixfs code paths. 

We want to keep (soon to be created) spec for raw/car responses minimal.
That will allow people to implement and run light "trustless transport gateways" that do not render unixfs (e.g. for IoT use cases).

core/corehttp/gateway_handler.go
```diff
@@ -271,120 +270,61 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
 	logger := log.With("from", r.RequestURI)
 	logger.Debug("http request received")
 
-	// X-Ipfs-Gateway-Prefix was removed (https://github.com/ipfs/go-ipfs/issues/7702)
-	// TODO: remove this after  go-ipfs 0.13 ships
-	if prfx := r.Header.Get("X-Ipfs-Gateway-Prefix"); prfx != "" {
-		err := fmt.Errorf("X-Ipfs-Gateway-Prefix support was removed: https://github.com/ipfs/go-ipfs/issues/7702")
-		webError(w, "unsupported HTTP header", err, http.StatusBadRequest)
+	if !validateSupportedHeaders(w, r) {
```
justincjohnson, line 273: I extracted some smaller functions from the code, partly to make it easier to reason about and partly because I had to extract `getOrHeadHandlerUnixfs` out of `getOrHeadHandler`, which needed to reuse some of this logic.

That said, I'm new to Go and coming most recently from an Elixir background, so I'm guessing I'm not writing idiomatic Go.  Please don't hesitate to suggest changes, or function renames or whatever.

core/corehttp/gateway_handler.go
```diff
@@ -271,120 +270,61 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
...
+		return
+	}
+
+	// Support custom response formats passed via ?format or Accept HTTP header
+	responseFormat := customResponseFormat(r)
+
+	// For Unixfs, when a path can't be resolved we need to check for redirects and pretty 404 page files.
+	if responseFormat == "" {
+		logger.Debugw("dispatching to getOrHeadHandlerUnixfs")
+		i.getOrHeadHandlerUnixfs(w, r, begin, logger)
```
justincjohnson, line 296: Had to dispatch to `getOrHeadHandlerUnixfs` here, to avoid any early 404s, which would then short circuit the _redirects lookup.

core/corehttp/gateway_handler.go
```diff
@@ -271,120 +270,61 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
...
 	// Resolve path to the final DAG node for the ETag
 	resolvedPath, err := i.api.ResolvePath(r.Context(), contentPath)
+
 	switch err {
 	case nil:
 	case coreiface.ErrOffline:
 		webError(w, "ipfs resolve -r "+debugStr(contentPath.String()), err, http.StatusServiceUnavailable)
 		return
 	default:
-		// if Accept is text/html, see if ipfs-404.html is present
```
justincjohnson, line 340: Moved pretty 404 logic over to `gateway_handler_unixfs.go`

core/corehttp/gateway_handler.go
```diff
@@ -875,3 +743,110 @@ func fixupSuperfluousNamespace(w http.ResponseWriter, urlPath string, urlQuery s
...
 	}) == nil
 }
+
+func validateSupportedHeaders(w http.ResponseWriter, r *http.Request) bool {
+	// X-Ipfs-Gateway-Prefix was removed (https://github.com/ipfs/go-ipfs/issues/7702)
+	// TODO: remove this after  go-ipfs 0.13 ships
+	if prfx := r.Header.Get("X-Ipfs-Gateway-Prefix"); prfx != "" {
+		err := fmt.Errorf("X-Ipfs-Gateway-Prefix support was removed: https://github.com/ipfs/go-ipfs/issues/7702")
+		webError(w, "unsupported HTTP header", err, http.StatusBadRequest)
+		return false
```
justincjohnson, line 753: As noted above, I'm not so sure I'm writing idiomatic Go here.  At the very least it would probably be good to have a more consistent naming convention for the functions such that it was clear whether true or false meant there was an error and we should return.

core/corehttp/gateway_handler_unixfs.go
```diff
@@ -3,15 +3,125 @@ package corehttp
...
+		redirectsFile, err := i.getRedirectsFile(r)
+		if err != nil {
+			switch err.(type) {
+			case resolver.ErrNoLink:
+				// _redirects files doesn't exist, so don't error
+			default:
+				// TODO(JJ): During tests we get multibase.ErrUnsupportedEncoding
+				// This comes from multibase and I assume is due to a fake or otherwise bad CID being in the test.
+				// So for now any errors getting the redirect file are silently ignored.
+				// internalWebError(w, err)
```
justincjohnson, line 34: General question about how we should handle errors reading and parsing the redirects if present.  Also note my comment about issues with test.  I need to dive into this more.

core/corehttp/gateway_handler_unixfs.go
```diff
@@ -3,15 +3,125 @@ package corehttp
...
+				// return
+			}
+		} else {
+			// _redirects file exists, so parse it and redirect
+			redirected, newPath, err := i.redirect(w, r, redirectsFile)
+			if err != nil {
+				// TODO(JJ): How should we handle parse or redirect errors?
```
justincjohnson, line 41: Above is errors from getting the file.  This is errors from parsing.

core/corehttp/gateway_handler_unixfs.go
```diff
@@ -3,15 +3,125 @@ package corehttp
...
+	resolvedPath, err := i.api.ResolvePath(r.Context(), contentPath)
+
+	switch err {
+	case nil:
+	case coreiface.ErrOffline:
+		webError(w, "ipfs resolve -r "+debugStr(contentPath.String()), err, http.StatusServiceUnavailable)
+		return
+	default:
+		// if Accept is text/html, see if ipfs-404.html is present
+		if i.servePretty404IfPresent(w, r, contentPath) {
```
justincjohnson, line 68: Pretty 404 logic moved here

core/corehttp/gateway_handler_unixfs.go
```diff
@@ -30,9 +140,124 @@ func (i *gatewayHandler) serveUnixFs(w http.ResponseWriter, r *http.Request, res
 	// Handling Unixfs directory
 	dir, ok := dr.(files.Directory)
 	if !ok {
-		internalWebError(w, fmt.Errorf("unsupported UnixFs type"))
+		internalWebError(w, fmt.Errorf("unsupported Unixfs type"))
```
justincjohnson, line 143: Just being consistent with rest of code

core/corehttp/gateway_handler_unixfs.go
```diff
@@ -30,9 +140,124 @@ func (i *gatewayHandler) serveUnixFs(w http.ResponseWriter, r *http.Request, res
...
+func getRootPath(path string) string {
+	if strings.HasPrefix(path, ipfsPathPrefix) && strings.Count(gopath.Clean(path), "/") >= 2 {
+		parts := strings.Split(path, "/")
+		return gopath.Join(ipfsPathPrefix, parts[2])
+	} else {
+		return ""
+	}
+}
+
+// TODO(JJ): I was thinking about changing this to just look at the root path as well, but the docs say it searches up
```
justincjohnson, line 213: I assume you agree?

core/corehttp/gateway_handler_unixfs.go
```diff
@@ -30,9 +140,124 @@ func (i *gatewayHandler) serveUnixFs(w http.ResponseWriter, r *http.Request, res
...
-		internalWebError(w, fmt.Errorf("unsupported UnixFs type"))
+		internalWebError(w, fmt.Errorf("unsupported Unixfs type"))
 		return
 	}
 	logger.Debugw("serving unixfs directory", "path", contentPath)
 	i.serveDirectory(w, r, resolvedPath, contentPath, dir, begin, logger)
 }
+
+// redirect returns redirected, newPath (if rewrite), error
+func (i *gatewayHandler) redirect(w http.ResponseWriter, r *http.Request, path ipath.Resolved) (bool, string, error) {
```
justincjohnson, line 151: I haven't touched any of this redirect file parsing logic yet.  Will need to settle on a format, ensure we handle optional bits like status code, etc.  The example file I've been using is...

```
# redirects file
test    /         301
^/hir$  /hi.html  302
^/hi    /hi.html  200
```

core/corehttp/gateway_handler_unixfs.go
```diff
@@ -30,9 +140,124 @@ func (i *gatewayHandler) serveUnixFs(w http.ResponseWriter, r *http.Request, res
...
+			case "*/*", "text/*", "text/html":
+				return "ipfs-404.html", "text/html", nil
+			}
+		}
+	}
+
+	return "", "", fmt.Errorf("there is no 404 file for the requested content types")
+}
+
+// TODO(JJ): Pretty sure this is incorrect.  Validate the correct approach.
```
justincjohnson, line 256: Any guidance is appreciated.  🙏

core/corehttp/gateway_handler.go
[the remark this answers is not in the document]
justincjohnson: @lidel I pushed new changes to my branch that I believe generally address your feedback.  The branch is still WIP but it's working for me.  I'm new to Go and IPFS and would be very grateful for any feedback you're willing to give.  🙏

lidel (changes requested):
Thank you for taking a stab at this! :heart: 

Wrote comments inline, but a quick summary:
- let's try to keep the changes minimal – cleanup refactors are good )and ok to keep them, but in the future, such things could be a separate PR :) 
- bit worried about impact of  `servePretty404IfPresent` being bigger than the actual  `_redirects`  – I am ok with us keeping it the way it was before just to limit the surface of this PR, seems that I overestimated how easy it will be to move it to unxfs scope, apologies! :see_no_evil: 
- needs DNSLink and catch-all tests
- try reusing existing parser lib, if possible. If anything is missing in existing one(s), fine to fork it for now.

core/corehttp/gateway_handler.go
justincjohnson: Excellent, that does sound nicer and more consistent.  I'll update all of these to simpler names with no negation.

core/corehttp/gateway_handler_unixfs.go
justincjohnson: Okay, makes sense.  So I assume I should keep existing pretty 404 logic, but only fall back to it if there isn't any [netlify style custom 404 page handling](https://docs.netlify.com/routing/redirects/redirect-options/#custom-404-page-handling) entry in the _redirects file.  And only the new `_redirects` file 404 logic will be documented.  👌

core/corehttp/gateway_handler_unixfs.go
justincjohnson: Good idea.  I'm guessing at the very least we'll have to contribute to another library to get feature parity with what Netlify currently supports (and I'm assuming that's our target to start with), but I'd rather not completely reinvent the wheel.

justincjohnson:
@lidel thank you for your excellent feedback.  🙏   I'm addressing the individual suggestions and comments separately but I want to make sure we're on the same page about something.

> bit worried about impact of servePretty404IfPresent being bigger than the actual _redirects – I am ok with us keeping it the way it was before just to limit the surface of this PR, seems that I overestimated how easy it will be to limit its scope, apologies! 🙈

From this comment and some others, it sounds like you might think the larger reorg and duplicate code paths were only introduced due to your request that I move pretty 404 logic into `gateway_handler_unixfs.go`.  Just in case you missed it, what I was trying to work around is the fact that any 404 here can't be handled without first looking for the `_redirects` file to ensure that the 404 isn't actually a redirect.

In other words, in this code at https://github.com/ipfs/go-ipfs/blob/master/core/corehttp/gateway_handler.go#L336-L350 (on master) we're not handling unixfs yet, but since when resolving the path for ETag purposes the resolution fails and we hit the switches default case and need to return a 404, even if I keep the pretty 404 logic here, how can I be sure it is truly a 404 without first processing the `_redirects` file?  But I can't process that file if not unixfs, which we're trying to move over to `gateway_handler_unixfs.go`.  So it seems to force a fork in the execution, or some other reconsideration of the organization of this code.

Does that make sense?  I'm totally open to your thoughts, but your comments made me think you might not see the issue I was trying to work around.  Thanks again.

core/corehttp/gateway_handler.go
justincjohnson: This link just takes me to my comment "Moved pretty 404 logic over to gateway_handler_unixfs.go".  Is that what you intended to link me?

justincjohnson:
Also, to be clear, assuming a larger refactoring does indeed end up being needed, I will gladly split the refactoring out into a separate PR first to make it easier to review, and then rebase this PR on top of that.

core/corehttp/gateway_handler_unixfs.go
justincjohnson: See my comments at https://github.com/ipfs/go-ipfs/pull/8816#issuecomment-1091022658.

justincjohnson:
@lidel When you have time, can you take a look at https://github.com/fission-suite/go-ipfs/pull/43/files and let me know if something like this would be acceptable?  This PR is in fission's fork of go-ipfs currently and is built on top of the cleanup PR above.  If something along these lines is more acceptable, I'll keep working through the remaining items and get them back on the branch tied to this PR for review.  Thank you!

Mitumaru (approved):

justincjohnson:
Given the issues that PRs in the fission-suite org are causing, I've moved these changes over to https://github.com/ipfs/go-ipfs/pull/8890, which is tied to my personal account. 🙏

PR #8821, opened 2022-03-24
CI: force running go-ipfs-http-client tests

Jorropo:
This test is actually testing is the ipfs binary (from the path).
Using the go-ipfs-http-client code.
But most PR only update the binary. HOWEVER Golang does not check PATH for caching.
So as far as golang knows: go-ipfs-http-client is old, it havn't changed, and it doesn't run it.

That leads us to this test not running on most important PR.

### DIFF
--- a/.circleci/main.yml
+++ b/.circleci/main.yml
@@ -292,7 +292,7 @@ jobs:
             - v1-go-api-{{ checksum "~/ipfs/go-ipfs/go-ipfs-api/go.sum" }}
             - v1-go-api-
       - run:
-          command: go test -v ./...
+          command: go test -count=1 -v ./...
           working_directory: ~/ipfs/go-ipfs/go-ipfs-api
       - save_cache:
           key: v1-go-api-{{ checksum "~/ipfs/go-ipfs/go-ipfs-api/go.sum" }}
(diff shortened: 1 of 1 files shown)


.circleci/main.yml
```diff
@@ -318,10 +318,10 @@ jobs:
             - v1-http-client-{{ checksum "~/ipfs/go-ipfs/go-ipfs-http-client/go.sum" }}
             - v1-http-client-
       - run:
-          name: go test -v ./...
+          name: go test -a -v ./...
           command: |
             export PATH=/tmp/circleci-workspace/bin:$PATH
-            go test -v ./...
+            go test -a -v ./...
```
guseggert, line 324: Nit: consider using `count=1` instead of `-a` which will rerun w/o rebuilding, and the docs bless as idiomatic:

> The idiomatic way to disable test caching explicitly is to use -count=1.

guseggert (approved):

aschmahmann:
@Jorropo this should probably also apply to `go-ipfs-api`

Jorropo:
@aschmahmann yes, I have did it for `go-ipfs-api` too (see the change line 295)

Jorropo:
Merging anyway because the test failing is in fact failing since multiple commits but wasn't showing up due to the precise caching issue this PR fix.

Issue #7481, opened 2020-06-16
Config inconsistent at runtime


Stebalien:
#### Version information:

> v0.5.1

#### Description:

The config attached to the command context is _not_ the same config returned by `repo.Config()`. The commands context should just ask the repo for the config.

schomatis:
@Stebalien Could you provide a code example, please?

BigLep:
@Stebalien : do you have more input here so this can be actioned?

BigLep:
2022-03-11 conversation: end goal is to make it crystal clear for users on whether they need to reboot or not when making config changes.

For priority determination, it would useful to know how many users are snagged by this issue.

PR #8824, opened 2022-03-25
fix(core/commands): do not cache config

schomatis:
Closes https://github.com/ipfs/go-ipfs/issues/7481.

From the issue description:
> The config attached to the command context is not the same config returned by repo.Config(). The commands context should just ask the repo for the config.


guseggert (approved):
I love to see diffs with lots of red

PR #8827, opened 2022-03-28
fix: report gateway http metrics only when response is successful

### LINKED ISSUE #8820: Gateway HTTP Metrics inconsistently measure failed response times
### Checklist

### Installation method

built from source

### Version

```Text
from inspection of master branch
```

### Config

_No response_

### Description

The per-format latency metrics added as part of https://github.com/ipfs/go-ipfs/issues/8441 are inconsistent as to whether they include failed responses or not.

`gw_unixfs_file_get_duration_seconds`, `gw_raw_block_get_duration_seconds` include timings for failed responses

`gw_car_stream_get_duration_seconds`, `gw_unixfs_gen_dir_listing_get_duration_seconds` do not include timings for failed responses

These should be consistent and my preference is that they should not include timings for failed responses (since errors will skew  latency timings lower in general)

Tech note: the difference is caused by the use of `http.ServeContent` for serving files and blocks which does not return an error. We should pass a custom ResponseWriter that detects whether an error occurred during the write and then only record the metric if there is no error.

iand:
Fixes #8820


lidel (approved):
LGTM.


@iand pushed two commits, mind taking a look if they make sense before  I merge this?

iand:
@lidel you can merge this whenever. I reran CI to fix a flake.

PR #8830, opened 2022-03-28
fix: unknown fetcher type error

### LINKED ISSUE #8780: Could not add migration to IPFS: Cannot get migrations from unknown fetcher type
### Version

```Text
0.12.0
```

### Description

```
Could not add migration to IPFS: Cannot get migrations from unknown fetcher type
```

Above  message was reported by Brave and on https://develop.element.io/#/room/#lobby:ipfs.io:

>  

> 

It does not impact migration itself, but raises a concern that something went wrong.

In this case, downloaded migration is not added to IPFS, so other machines from LAN can't fetch it locally p2p, but need to fetch from swarm/HTTP. Not a hard error, but something we should fix to improve resilience.

Relevant code: 
- https://github.com/ipfs/go-ipfs/blob/v0.12.0/cmd/ipfs/daemon.go#L489-L493
- https://github.com/ipfs/go-ipfs/blob/v0.12.0/cmd/ipfs/add_migrations.go#L24-L60

lidel:
Closes #8780

### DIFF
--- a/cmd/ipfs/add_migrations.go
+++ b/cmd/ipfs/add_migrations.go
@@ -31,16 +31,13 @@ func addMigrations(ctx context.Context, node *core.IpfsNode, fetcher migrations.
 
 	for _, fetcher := range fetchers {
 		switch f := fetcher.(type) {
-		case nil:
-			// https://github.com/ipfs/go-ipfs/issues/8780
-			continue
 		case *ipfsfetcher.IpfsFetcher:
 			// Add migrations by connecting to temp node and getting from IPFS
 			err := addMigrationPaths(ctx, node, f.AddrInfo(), f.FetchedPaths(), pin)
 			if err != nil {
 				return err
 			}
-		case *migrations.HttpFetcher:
+		case *migrations.HttpFetcher, *migrations.RetryFetcher: // https://github.com/ipfs/go-ipfs/issues/8780
 			// Add the downloaded migration files directly
 			if migrations.DownloadDirectory != "" {
 				var paths []string
--- a/cmd/ipfs/add_migrations.go
+++ b/cmd/ipfs/add_migrations.go
@@ -9,7 +9,7 @@ import (
 	"os"
 	"path/filepath"
 
-	"github.com/ipfs/go-ipfs-files"
+	files "github.com/ipfs/go-ipfs-files"
 	"github.com/ipfs/go-ipfs/core"
 	"github.com/ipfs/go-ipfs/core/coreapi"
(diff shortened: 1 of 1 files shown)


cmd/ipfs/add_migrations.go
```diff
@@ -31,6 +31,9 @@ func addMigrations(ctx context.Context, node *core.IpfsNode, fetcher migrations.
 
 	for _, fetcher := range fetchers {
 		switch f := fetcher.(type) {
+		case nil:
+			// https://github.com/ipfs/go-ipfs/issues/8780
+			continue
```
aschmahmann, line 36: How do we end up with one of these?

lidel: We probably don't – opened this draft PR  to run CI and then look closer.

I now suspect we don't have nil here,  but a `RetryFetcher` which wraps `NewHttpFetcher`:
https://github.com/ipfs/go-ipfs/blob/a6d04184062d1883f090d9661a4d8170a726cfc4/repo/fsrepo/migrations/migrations.go#L165

iiuc `switch`  over `fetcher` type interprets  `RetryFetcher` as something other than `IpfsFetcher` and `HttpFetcher` and that is why default gets executed. 

Quick demo: https://go.dev/play/p/MEKo40x1z0I

lidel: @aschmahmann if we added  `RetryFetcher` some time later,  that probably introduced the bug. 

Since it is only used for HTTP, we could fix it this way:

```diff
- case *migrations.HttpFetcher:
+ case *migrations.HttpFetcher, *migrations.RetryFetcher:
```


Thoughts?

guseggert: Can we add something to the Fetcher interface for this so we can remove this brittle type switch altogether?

lidel: Probably, but at the same time we won't be adding any new Fetcher types any time soon.
I'm merging this one-liner fix for now – we will be revisiting fetchers after we have car  export on gateways anyway, we can refactor things then.

cmd/ipfs/add_migrations.go
```diff
@@ -37,7 +37,7 @@ func addMigrations(ctx context.Context, node *core.IpfsNode, fetcher migrations.
 			if err != nil {
 				return err
```
Joelsamijon, line 38: gh pr checkout 8830

Joelsamijon (commented):
cmd/ipfs/add_migrations.go

PR #8834, opened 2022-03-31
refactor(gw/dir-index-html): remove gobindata

frankywahl:
Replaces https://github.com/go-bindata/go-bindata with the standard library `embed`

This remove a dependency on go-bindata and makes the tooling flow less complex


core/corehttp/gateway_handler_unixfs_dir.go
```diff
@@ -87,8 +87,8 @@ func (i *gatewayHandler) serveDirectory(w http.ResponseWriter, r *http.Request,
 	w.Header().Set("Content-Type", "text/html")
 
 	// Generated dir index requires custom Etag (it may change between go-ipfs versions)
-	if assets.BindataVersionHash != "" {
-		dirEtag := `"DirIndex-` + assets.BindataVersionHash + `_CID-` + resolvedPath.Cid().String() + `"`
```
frankywahl, line 91: Slight change in behavior appear here. We could put into our tooling a hash of all the files we're after if we want to allow for caching of the index file across versions (as it use to be). Or we could actually compute the hash (md5?) of the file we're after directly here.

assets/assets.go
```diff
@@ -29,6 +29,13 @@ var initDocPaths = []string{
 	filepath.Join("init-doc", "ping"),
 }
 
+// Asset loads and returns the asset for the given name.
+// It returns an error if the asset could not be found or
+// could not be loaded.
+func Asset(f string) ([]byte, error) {
+	return dir.ReadFile(f)
+}
```
frankywahl, line 37: This just keeps the old API in place that was generated from `go-bindata`

core/corehttp/gateway_handler_unixfs_dir.go
lidel: We should restore old behavior. Perhaps:
- hash `dir embed.FS`  on startup and store it somewhere?
  - we dont need to keep the entire hash, only ~7 characters  will be more than enough.
  - fyi we use https://github.com/OneOfOne/xxhash in places where we need a fast non-cryptographic hash

lidel (changes requested):
Thank you for cleaning this, up! :+1: 

Before we continue, we need to tweak the  Etag behavior (see comment inline)

core/corehttp/gateway_handler_unixfs_dir.go
frankywahl: I think this is a really interesting aspect anw why I highlighted it. Etag was actually debated as part of the `embed` implementation [here](https://github.com/golang/go/issues/43223)

The way it was before this change was containing an "flaw" (we could actually file it as a separate issue) in that if there are two files foo.html and bar.html, either of the two files being changed would bust the cache of the other one too despite it being unchanged. 

I have implemented it using sha256 in the following commit.

It has the same behavior as before, with a slight impact on startup (computing the hash). We can try to optimize it after with some function call, caching and `sync.Once` to avoid computing the hash until it is needed at a later stage. I believe the impact would be quite minimal given there are not many assets to hash and the efficiency of sha256 so I defer that to a separate concern. This also avoids bringing in another dependency (as xxxhash would). 

Another option would also be to use something like https://github.com/benbjohnson/hashfs - this would allow a per-asset caching. (This definitely could be done as another pull-request). 

Given the concern of each one, I though just busting the cache between version changes would be an okay solution.

Let me know what you think, happy to remove the commit or attempt another implementation

assets/assets.go
```diff
@@ -29,6 +36,35 @@ var initDocPaths = []string{
...
+		file, err := dir.ReadFile(path)
+		if err != nil {
+			panic(err)
+		}
+		_, err = sha.Write(file)
+		if err != nil {
+			panic(err)
+		}
+		return nil
+	})
```
Jorropo, line 55: Sum suggestion:
```suggestion
	sum := xxhash.NewSum32()
	err := fs.WalkDir(dir, ".", func(path string, d fs.DirEntry, err error) error {
		if d.IsDir() {
			return nil
		}
		
		file, err := dir.Open(path)
		if err != nil {
			return err
		}
		defer file.Close()
		_, err = io.Copy(sum, file)
		return err
	})
	if err != nil {
		panic("error creating asset ETag sum: " + err.Error())
	}
```

assets/assets.go
```diff
@@ -29,6 +36,35 @@ var initDocPaths = []string{
...
+
+	hexSum := sha.Sum(nil)
+	AssetHash = fmt.Sprintf("%x", hexSum)
```
Jorropo, line 58: Sum suggestion:
```suggestion
	sum := xxhash.New32()
	err := fs.WalkDir(dir, ".", func(path string, d fs.DirEntry, err error) error {
		if d.IsDir() {
			return nil
		}
		
		file, err := dir.Open(path)
		if err != nil {
			return err
		}
		defer file.Close()
		_, err = io.Copy(sum, file)
		return err
	})
	if err != nil {
		panic("error creating asset ETag sum: " + err.Error())
	}
	
	AssertHash = strconv.FormatUint(uint64(sum.Sum32()), 32)
```

core/corehttp/gateway_handler_unixfs_dir.go
Jorropo: > It has the same behavior as before, with a slight impact on startup (computing the hash). We can try to optimize it after with some function call, caching and `sync.Once` to avoid computing the hash until it is needed at a later stage.

That sounds like a good idea, however I think it's better to try the easy stuff first (using a cheaper hash function). And see about that later, probably that `xxhash` is fast enough anyway and that this isn't needed.

An other option similar to `sync.Once` is we could have `init` starting a `go Hash.Do(hashfunc)` so it is still done at init but asynchronously in the background. And then consumers would also try to do `Hash.Do(hashfunc)` and be stuck until the once return.

assets/assets.go
Jorropo: Any reason not to remove this ?

Is this code used by anything else than `go-ipfs` ?

assets/assets.go
```diff
@@ -29,6 +36,35 @@ var initDocPaths = []string{
 	filepath.Join("init-doc", "ping"),
 }
 
+func init() {
+	sha := sha256.New()
```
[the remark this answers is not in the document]
frankywahl: Correct - that is why my first version removed this entirely to just use the version of the binary.

assets/assets.go
```diff
@@ -29,6 +36,35 @@ var initDocPaths = []string{
...
+	fs.WalkDir(dir, ".", func(path string, d fs.DirEntry, err error) error {
+		if d.IsDir() {
+			return nil
+		}
+
+		file, err := dir.ReadFile(path)
+		if err != nil {
+			panic(err)
+		}
+		_, err = sha.Write(file)
```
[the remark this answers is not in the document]
frankywahl: Good point!

assets/assets.go
```diff
@@ -29,6 +36,35 @@ var initDocPaths = []string{
 	filepath.Join("init-doc", "ping"),
 }
 
+func init() {
+	sha := sha256.New()
+	fs.WalkDir(dir, ".", func(path string, d fs.DirEntry, err error) error {
+		if d.IsDir() {
+			return nil
+		}
+
```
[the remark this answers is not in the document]
frankywahl: Yes - that is correct - I would worry more about it if it was used beyond the Etag usage. But for just an etag, any kind of hashing would work I believe.

assets/assets.go
frankywahl: I like this proposal except for 1 thing: I dont think we should know about the `ETag` usage. At the end of the day, that is how this package is used by its consumer - but that could change over time and we wouldn't want to have the `Etag` through off.

assets/assets.go
frankywahl: This just keeps the old API in place that was being used in the rest of the code. It makes changes be baby step and avoids changing many files at once since that is the exposed API that was given by bindata beforehand.

Jorropo: > keeps the old API in place that was being used in the rest of the code

Yes, but can't you just change it in the rest of the code also ? Any reason not to do that ?

> It makes changes be baby step and avoids changing many files at once

I think that make sense but, as multiple commits in a single PR, I don't see any reason to keep this API past this PR.

Code kept for *history* reasons is also code that is confusing. a new comer shouldn't have to use git blame to understand why an API is the way it is. So when breaking is not an issue, I don't see any reason not to do so.

core/corehttp/gateway_handler_unixfs_dir.go
frankywahl: @Jorropo Not sure I fully understand the last part of your comment (though I get it's hoping to do that in a goroutine), but I think it's not that big of a bother anyhow as we both think that such an optimization can come later if it is needed and doesn't need to be part of this PR I think.

Jorropo: > later if it is needed

I agree with that too, so this point is solved. :slightly_smiling_face: 

What I proposed, is something like this: (pseudo code)
```go
func init() {
  go etagHashLock.Do(func() {
    etagHash = computeEtagHash()
  })
}

var etagHash string
var etagHashLock sync.Once

func GetETagHash() string {
  etagHashLock.Do(func(){}) // Here we do nothing, basically we just use the once as a way to wait on the goroutine that init started
  return etagHash
}
```

Jorropo (approved):
I would still like the old `Asset` API to be removed, but this can be done later and is really solid. thx

assets/assets.go
Jorropo: I think that clearer.

assets/assets.go
```diff
@@ -18,6 +19,14 @@ import (
 	"github.com/ipfs/interface-go-ipfs-core/path"
 )
 
+//go:embed init-doc dir-index-html/dir-index.html dir-index-html/knownIcons.txt
+var dir embed.FS
+
+var (
+	// AssetHash a cryptographic hash of all embedded assets
```
Jorropo, line 26: Actually this need be updated.
```suggestion
	// AssetHash a non-cryptographic hash of all embedded assets
```

Jorropo (changes requested):

assets/assets.go
frankywahl: Nice catch - missed it.

Jorropo (approved):
LGTM, thx for the patch :slightly_smiling_face:

Jorropo:
@frankywahl the CI is stuck and wont run, I can't unstuck it.
Apparently this is possible if you have done something to your fork circleci config that makes run on your circleci account ? That doesn't work because the run approval I've did is on that repo not yours. (that just github authorisatyion mess)
Or maybe try pushing a new commit so it retry to rerun it ?

(else I'll just move your branch to a new one on this repo and merge from this) :)

frankywahl:
@Jorropo I've squashed some commits and forced push - it should re-trigger the CI. I'm not sure what the issue is with Circle as I don't even have a CircleCI account - so I definitely did not change any configurations there. 🤞 it works this time.

lidel:
CircleCI is broken when your fork comes from an org. 

@Jorropo fyi  the way to fix this is to push the changes as a new, upstream branch.
 that way commits will come from our repo and CI will run ok. 
 
Let's continue in https://github.com/ipfs/go-ipfs/pull/8872

PR #8835, opened 2022-03-31
fix(gw): validate requested CAR version

lidel:
CAR responses were added in https://github.com/ipfs/go-ipfs/pull/8758

This PR adds validation of `application/vnd.ipld.car;version=n` passed  in the `Accept` header by HTTP clients.
We need this to align Gateway behavior with the spec submitted to IANA (https://github.com/ipfs/in-web-browsers/issues/192).

tl;dr:
- Gateway returns HTTP 400 Bad Request error when request is for unsupported CAR version (right now we only support CARv1)
- Proper parsing of media types using `mime.ParseMediaType` means we don't need to worry about clients adding/removing whitespace or using custom case

### DIFF
--- a/test/sharness/t0118-gateway-car.sh
+++ b/test/sharness/t0118-gateway-car.sh
@@ -51,8 +51,23 @@ test_launch_ipfs_daemon_without_network
     # explicit version=1
     test_expect_success "GET for application/vnd.ipld.raw version=1 returns a CARv1 stream" '
     ipfs dag import test-dag.car &&
+    curl -sX GET -H "Accept: application/vnd.ipld.car;version=1" "http://127.0.0.1:$GWAY_PORT/ipfs/$ROOT_DIR_CID/subdir/ascii.txt" -o gateway-header-v1.car &&
+    test_cmp deterministic.car gateway-header-v1.car
+    '
+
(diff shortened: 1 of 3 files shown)


test/sharness/t0118-gateway-car.sh
```diff
@@ -51,10 +51,25 @@ test_launch_ipfs_daemon_without_network
     # explicit version=1
     test_expect_success "GET for application/vnd.ipld.raw version=1 returns a CARv1 stream" '
     ipfs dag import test-dag.car &&
+    curl -sX GET -H "Accept: application/vnd.ipld.car;version=1" "http://127.0.0.1:$GWAY_PORT/ipfs/$ROOT_DIR_CID/subdir/ascii.txt" -o gateway-header-v1.car &&
+    test_cmp deterministic.car gateway-header-v1.car
+    '
+
+    # explicit version=1 without whitepace
```
guseggert, line 58:
```suggestion
    # explicit version=1 with whitepace
```

BigLep:
Estimated in [ipfs-stewards](https://play-game.scrumpypoker.com/#/room/gzepptgngfrc) to **0.5**, the average vote was **0.5** story points.

Issue #6487, opened 2019-07-09
Outdated Version Notice


Stebalien:
It would be great to print a notice to users when they're running an outdated version of go-ipfs. Luckily, we should be able to do this without deploying any new services or making any additional requests.

Instead, can just use the bootstrappers (i.e., "trusted nodes") to determine if we're out of date, no extra network communication required.

1. On start, register for connection events.
2. When we connect to a bootstrapper, check it's version.
3. If it's running a newer (non-RC) version, print a message.

Caveats:

* This won't work if we stop connecting to bootstrappers. However, we're likely to continue to try to connect to at least _one_ bootstrapper when possible.
* We can't determine the latest release when we're testing an RC and/or custom build on a bootstrapper.

Kubuxu:
Bitcoin has a notice system where there is a group of keys that can sign a message and this messages will be propagated across the p2p network and displayed to users.

lidel:
This was mentioned during stewards sync today, my thinking:
- We don't want to rely on bootstrappers too much: we want to use them only on initial bootstrap (https://github.com/ipfs/go-ipfs/issues/3908, https://github.com/ipfs/go-ipfs/issues/3926), so can't use them for this reliably.
- Instead, we could detect when certain % of go-ipfs  peers (`Agent Version` column from `ipfs stats dht`) use the new version and warn user that they should update to the latest version 
  - this has very nice property of "not calling home" or producing any unnecessary DNSLink lookups for https://dist.ipfs.io/go-ipfs/  and does not impact private swarms, only public peers
- It needs to be WARNING printed to the stdout of `ipfs daemon`  
  - that way it will reach everyone, no matter what package manager / docker image is used etc 
  - we could be very polite and allow for disabling new version check by setting ENV variable – but the default should check

BigLep:
I like this @lidel.

@schomatis : what do you think about implementing this?

schomatis:
Yes, assigning.

aschmahmann:
> Instead, we could detect when certain % of go-ipfs peers (Agent Version column from `ipfs stats dht`) use newer version and warn user that they should update to the latest version

@lidel did you mean to specifically reference the DHT routing table (`ipfs stats dht`) since it has some basic sybil resistance going on, or did you just mean generally the user agents of all peers we're connected to, or even all peers we have information about in the peerstore.

[...]

PR #8839, opened 2022-03-31
feat(cmds): add version check command

schomatis:
Add `ipfs version check` command to check running version against our DHT peers and report if a (configurable) fraction of them are running a newer (stable) version. (Manually tested decreasing the local `CurrentVersionNumber`, not sure of a simple way to test this locally in sharness.)

This is the first step toward https://github.com/ipfs/go-ipfs/issues/6487. Once this lands we can incorporate it as a periodic command in a running daemon if we want.

### DIFF
--- a/core/commands/version.go
+++ b/core/commands/version.go
@@ -5,17 +5,24 @@ import (
 	"fmt"
 	"io"
 	"runtime/debug"
+	"strings"
 
 	version "github.com/ipfs/kubo"
+	"github.com/ipfs/kubo/core/commands/cmdenv"
 
 	cmds "github.com/ipfs/go-ipfs-cmds"
+
+	versioncmp "github.com/hashicorp/go-version"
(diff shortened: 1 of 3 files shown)


guseggert:
Can we add some tests to this?

schomatis:
@guseggert 
> not sure of a simple way to test this locally in sharness.)

I'll need some pointers on how to set up a local network to test this, otherwise the test will become much more work than the feature itself.

lidel (commented):
Triage note:
- rebased
- this PR  was created before we had `ipfs routing` and [custom routing](https://github.com/ipfs/kubo/blob/master/docs/config.md#routing), perhaps there is a better way to get peerbook now 
- it is ad-hoc command, which is fine, but we need something that runs works out-of-the box, like we did in https://github.com/ipfs/kubo/pull/9886

lidel:
Continued in https://github.com/ipfs/kubo/pull/10438

PR #8840, opened 2022-03-31
Dependabot/go modules/go.uber.org/fx 1.17.1

sullltan123:


guseggert:
What is the motivation for this?

Also remove the DS_Store file.

PR #8841, opened 2022-04-01
feat(gw): tracing spans per response type

lidel:
This PR extends #8595 with spans for each response type from https://github.com/ipfs/go-ipfs/pull/8758
Adds more visibility into how long generic lookup takes vs producing specific response type.

## Demo

| CAR | Block | File | Directory |
| ---- | ---- | ---- | ---- |
|  | block_2022-04-01_01-47 |   |  dir_2022-04-01_01-48 |

### DIFF
--- a/core/corehttp/gateway_handler.go
+++ b/core/corehttp/gateway_handler.go
@@ -352,6 +352,7 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
 	// Detect when explicit Accept header or ?format parameter are present
 	responseFormat := customResponseFormat(r)
 	trace.SpanFromContext(r.Context()).SetAttributes(attribute.String("ResponseFormat", responseFormat))
+	trace.SpanFromContext(r.Context()).SetAttributes(attribute.String("ResolvedPath", resolvedPath.String()))
 
 	// Finish early if client already has matching Etag
 	if r.Header.Get("If-None-Match") == getEtag(r, resolvedPath.Cid()) {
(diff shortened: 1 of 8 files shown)


lidel:
@guseggert this PR points at your branch, so if this looks good, freel free to merge and we can continue in https://github.com/ipfs/go-ipfs/pull/8595

mk/golang.mk
```diff
@@ -70,7 +70,7 @@ test_go_fmt:
 TEST_GO += test_go_fmt
 
 test_go_lint: test/bin/golangci-lint
-	golangci-lint run ./...
+	golangci-lint run --timeout=3m ./...
```
lidel, line 73: This helps with flaky CI.

core/corehttp/gateway_handler.go
```diff
@@ -352,6 +352,7 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
 	// Detect when explicit Accept header or ?format parameter are present
 	responseFormat := customResponseFormat(r)
 	trace.SpanFromContext(r.Context()).SetAttributes(attribute.String("ResponseFormat", responseFormat))
+	trace.SpanFromContext(r.Context()).SetAttributes(attribute.String("ResolvedPath", resolvedPath.String()))
```
guseggert, line 355:
```suggestion
	trace.SpanFromContext(r.Context()).SetAttributes(
		attribute.String("ResponseFormat", responseFormat),
		attribute.Stringer("ResolvedPath", resolvedPath),
	)
```

guseggert: (using `attribute.Stringer` will avoid allocating and computing the value if tracing's disabled, there's no span, etc.)

core/corehttp/gateway_handler_block.go
```diff
@@ -6,13 +6,18 @@ import (
...
 	ipath "github.com/ipfs/interface-go-ipfs-core/path"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
 )
 
 // serveRawBlock returns bytes behind a raw block
-func (i *gatewayHandler) serveRawBlock(w http.ResponseWriter, r *http.Request, blockCid cid.Cid, contentPath ipath.Path, begin time.Time) {
-	blockReader, err := i.api.Block().Get(r.Context(), contentPath)
+func (i *gatewayHandler) serveRawBlock(w http.ResponseWriter, r *http.Request, resolvedPath ipath.Resolved, contentPath ipath.Path, begin time.Time) {
+	ctx, span := tracing.Span(r.Context(), "Gateway", "ServeRawBlock", trace.WithAttributes(attribute.String("path", resolvedPath.String())))
```
guseggert, line 17: Do we need this attribute, since the path is already on the parent span?

PR #8850, opened 2022-04-05
fix(cmds): add: reject files with different import dir

schomatis:
Import fix from https://github.com/ipfs/go-ipfs-cmds/pull/230.

cc @bmwiedemann

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -31,7 +31,7 @@ require (
 	github.com/ipfs/go-graphsync v0.11.0
 	github.com/ipfs/go-ipfs-blockstore v1.2.0
 	github.com/ipfs/go-ipfs-chunker v0.0.5
-	github.com/ipfs/go-ipfs-cmds v0.8.0
+	github.com/ipfs/go-ipfs-cmds v0.8.1-0.20220405165343-ebff0685db8a
 	github.com/ipfs/go-ipfs-exchange-interface v0.1.0
 	github.com/ipfs/go-ipfs-exchange-offline v0.2.0
 	github.com/ipfs/go-ipfs-files v0.0.9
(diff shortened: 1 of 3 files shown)


bmwiedemann:
You are still not testing for what was the original problem:

```bash
mkdir -p dir{1,2}
echo 1 > dir1/x
echo 2 > dir2/x
ipfs add dir{1,2}/x
```

should print more than just
```
added QmSFxnK675wQ9Kc1uqWKyJUaNxvSc2BP5DbXCD3x93oq61 x
```

schomatis:
@bmwiedemann 

```
ipfs add dir{1,2}/x
```

Different import directories for the same file name are now rejected:

```
Error: file name x repeated under different import directories: dir2 and dir1
```

That's what is now being tested in sharness.

(This may not be clearly reflected in the original PR since we went back and forth on whether to allow this or how to reject it.)

bmwiedemann:
I see. My original use-case was to import all files from a directory hierarchy without adding directory objects to IPFS, so it seems, that has to be done per directory then.

schomatis:
@bmwiedemann Thanks for following up on this. I think we (and especially _I_) have made a mistake here: we were trying to fix an error in the `ipfs add -w` variant but to do so we applied it to the commands library that affects all cases. I'm still not sure what is the intended behavior when we add different but same-name files all together in `ipfs add`. Could you submit a new issue with your use case and a simple example, please?

Jorropo:
Not quite the behaviour we want but still better than what it use to be, merging but we will leave the issue open.

Jorropo (approved):
LGTM as a first fix

Jorropo (approved):

PR #8852, opened 2022-04-06
chore: add CODEOWNERS

lidel:
This adds CODEOWNERS file which automatically asks specific people for a review when specific paths are modified as a part of a PR.

To be specific, I've added myself, because I'd like to follow gateways changes.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,7 @@
+# Code owners are automatically requested for review when someone opens a pull
+# request that modifies code that they own. Code owners are not automatically
+# requested to review draft pull requests.
+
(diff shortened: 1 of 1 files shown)


guseggert (approved):

aschmahmann (approved):
SGTM. I'm not sure how widely applicable this will be, certainly makes sense for gateway code though and if we find more use then great!

BigLep (approved):
Thanks for getting started.  Looks good!

lidel:
You can see demo in https://github.com/ipfs/go-ipfs/pull/8816 (PR draft):

>  

Pretty cool!

Issue #8178, opened 2021-06-07
Faster gateway directory listings


Stebalien:
Update 11/4(@schomatis): We need to refactor the gateway code to use `UnixfsAPI.Ls()` instead of `Get`.

Instead of resolving file types up-front in the gateway, we can take advantage of streaming HTML rendering (most browsers support this) by:

1. Returning an HTML body with the directory listing as fast as possible.
2. Streaming back some trailing style elements to "update" the directory listing with file types.

That way the directory listing appears immediately, and icons for file-types will be filled in as new `<style>` elements get sent back to the browser.

Alternatively, we could do this with some javascript. But I'd _like_ to avoid that if possible.

lidel:
Update around go-ipfs 0.13:  

- When we have https://github.com/ipfs/go-ipfs/pull/8758 and https://github.com/ipfs/go-ipfs/issues/8823 we can stop fetching child blocks in dir-index-html code, and return listing immediately.  
- Child information could then be fetched via `?format=raw|dag-json|dag-cbor` and updated asynchronously, with some sane limits (eg. only items visible in the browser viewport). Fetching as `dag-cbor` will be most bandwidth-efficient, but we could implement `raw` vasriant first, and then improve  performance when we have cbor option.

lidel:
quick thought on a band-aid for this:
- we should detect when dir is over N items and only render  that + add footer saying it's too big, and suggest using the CLI  `ipfs ls -s --size=false --resolve-type=false $BIG_DIR_CID`
- this would be a band-aid, but probably easy to implement in a few lines of go

lidel:
@schomatis would you be interested in doing the band-aid described above, so we could ship it in go-ipfs 0.13? Let's go with N=5000 for now.

schomatis:
Assigning. If this is time sensitive I'm probably not the best person though, but will give you an update on this this week.

schomatis:
Sorry, just now reading the band-aid alternative: yes, seems like a quick fix I can implement to help alleviate this.

schomatis:
@lidel Need some clarifications:

A. On the objective of this issue, I'm not clear which is the performance downgrade we're trying to fix. From the OP:

> Instead of resolving file types up-front in the gateway

[...]

PR #8853, opened 2022-04-06
feat(gateway): Gateway.FastDirIndexThreshold

schomatis:
>  Toward https://github.com/ipfs/go-ipfs/issues/8178, 

TLDR: this PR skips costly  "size" column if directory is bigger or equal `Gateway.FastDirIndexThreshold`.  

Details  in https://github.com/ipfs/go-ipfs/pull/8853/files#pullrequestreview-942930258

## Demo 

You can test with `bafybeiggvykl7skb2ndlmacg2k5modvudocffxjesexlod2pfvg5yhwrqm` (has 10k items, loads fast)

>  Screen Shot 2022-04-14 at 23 54 55

---- 

<details>
  <summary>Click to expand old notes by schomatis</summary>
 
[re] "band-aid approach" from https://github.com/ipfs/go-ipfs/issues/8178#issuecomment-1088826476

```bash
ipfs config --json Gateway.FastDirIndexThreshold 3
ipfs daemon --offline

mkdir big-dir
touch big-dir/file{1..5}
BIG_DIR=$(ipfs add big-dir -r -Q)
echo "http://localhost:8080/ipfs/$BIG_DIR"
# http://localhost:8080/ipfs/QmZsx9Brmf8KGgr7o8P895WPJ7DSqTY8h2fWkvHDSnRJkM

rm big-dir/file{3..5}
BIG_DIR=$(ipfs add big-dir -r -Q)
echo "http://localhost:8080/ipfs/$BIG_DIR"
# http://localhost:8080/ipfs/QmTb1wneppbtKYuR6EwJdnieGVyrtXz3VebYU2WG3GQ1Pz
```

Take notice that there may be some caching mechanism that won't apply this limit if the directory has already been served. Take it for a spin and see if this works for your use case. It's the basic implementation of the stopgap suggested but we might need to add either a timeout or fetch entries metadata in parallel (right now this is sequential and a missing/slow entry consumes time from the total gateway timeout).

</details>


schomatis:
With the current interfaces it seems browsing through the directory to get its number of entries will always be blocking on the entry node fetch. At this time asserting that a directory has, say, 5000 entries to stop its listing implies fetching 5000 nodes (defeating the purpose of the cutoff), which even if parallelizing will quickly be halted on the missing/slow entries.

Ideally I would need to extend the Directory [interface](https://github.com/ipfs/go-ipfs-files/blob/9b481f40778c3b1e65a229248c7f0b730ef72b1e/file.go#L62-L83) to be able to get its number of entries (not the entries themselves) or maybe provide another version of its [iterator](https://github.com/ipfs/go-ipfs-files/blob/9b481f40778c3b1e65a229248c7f0b730ef72b1e/file.go#L44-L60) that advances through the entries without fetching them (unless some property like Size is requested).

schomatis:
On a second-third thought: the best I can do to deliver something useful here (which this is not it) without breaking interfaces is to make an independent `UnixfsAPI.Ls()` call to have the directory entries as fast as possible and decide if we want to go ahead and sequentially fetch entries metadata later (with the N limit of the band-aid approach). `UnixfsAPI.Ls()` is what `ipfs ls` uses and basically we can do a `ipfs ls --size=false --resolve-type=false` internally to get the parallel entry count.

Eventually we will need to make the directory iterator (linked above) parallel or switch entirely to the `UnixfsAPI.Ls()` API (which sounds like the right thing to do; we probably shouldn't have been using  `UnixfsAPI.Get()` in the first place).

I'm removing review requests and marking this as 'In Progress' again but will at least need a confirmatory :+1: here to proceed, and also will be prioritizing other issues next week (unless explicitly told otherwise, your call @lidel; this should still be an easy patch that won't take too much time).

config/gateway.go
```diff
@@ -53,6 +53,12 @@ type Gateway struct {
 	//  }
 	PathPrefixes []string
 
+	// MaxDirectorySize is the maximum directory size that will be listed in the
+	// gateway. A directory being listed with a detected size above this threshold
+	// will be stopped to avoid performance downgrade while fetching entries'
+	// metadata. The default value of 0 means no limit will be applied.
+	MaxDirectorySize int `json:",omitempty"`
```
lidel, line 60: Let's make this optional in the config, most users will be fine with implicit default.

```suggestion
	MaxDirectorySize *OptionalInteger `json:",omitempty"`
```

lidel (commented):
@schomatis refactoring generated dir listing  to use `UnixfsAPI.Ls()`  is  a good plan  :+1:  
If you are ok with tacking this before CIDv1, that would be the best (I'd like to include this fix in go-ipfs 0.13, if possible) :pray:

Mitumaru (approved):

config/gateway.go
schomatis: Nice, didn't know this existed.

schomatis: I'm not sure `OptionalInteger` added anything different than ``json:",omitempty"`` if we're using the default Go value of 0. @lidel Could you confirm how did you envision us using this? I'm basically calling it like `MaxDirectorySize: int(cfg.Gateway.MaxDirectorySize.WithDefault(0)),`.

core/corehttp/gateway.go
```diff
@@ -89,9 +90,10 @@ func GatewayOption(writable bool, paths ...string) ServeOption {
 			}, headers[ACEHeadersName]...))
 
 		var gateway http.Handler = newGatewayHandler(GatewayConfig{
-			Headers:      headers,
-			Writable:     writable,
-			PathPrefixes: cfg.Gateway.PathPrefixes,
+			Headers:          headers,
+			Writable:         writable,
+			PathPrefixes:     cfg.Gateway.PathPrefixes,
+			MaxDirectorySize: int(cfg.Gateway.MaxDirectorySize.WithDefault(0)),
```
lidel, line 96:
```suggestion
			MaxDirectorySize: int(cfg.Gateway.MaxDirectorySize.WithDefault(2500)),
```

schomatis:
@lidel Implemented a usable limit using `UnixfsAPI.Ls()`.

core/corehttp/gateway_handler_unixfs.go
```diff
@@ -17,8 +19,14 @@ import (
 func (i *gatewayHandler) serveUnixFs(w http.ResponseWriter, r *http.Request, resolvedPath ipath.Resolved, contentPath ipath.Path, begin time.Time, logger *zap.SugaredLogger) {
 	ctx, span := tracing.Span(r.Context(), "Gateway", "ServeUnixFs", trace.WithAttributes(attribute.String("path", resolvedPath.String())))
 	defer span.End()
+
+	// Decouple (potential) directory context to cancel it in case it's too big.
+	directoryContext, cancelDirContext := context.WithCancel(ctx)
```
lidel, line 24: nit: we keep the context name in other places, lets to the same here
```suggestion
	// Decouple (potential) directory context to cancel it in case it's too big.
	ctx, cancel := context.WithCancel(ctx)
```

core/corehttp/gateway_handler_unixfs.go
```diff
@@ -38,6 +46,38 @@ func (i *gatewayHandler) serveUnixFs(w http.ResponseWriter, r *http.Request, res
...
+					directoryEntriesNumber++
+					if directoryEntriesNumber >= i.config.MaxDirectorySize {
+						close(directoryTooBig)
+						cancelDirContext()
+						return
+					}
+				}
+			}
+		}()
+	}
```
lidel, line 79: This should be moved to a separate func in  gateway_handler_unixfs_dir.go

config/gateway.go
```diff
@@ -53,6 +53,12 @@ type Gateway struct {
 	//  }
 	PathPrefixes []string
 
+	// MaxDirectorySize is the maximum directory size that will be listed in the
+	// gateway. A directory being listed with a detected size above this threshold
+	// will be stopped to avoid performance downgrade while fetching entries'
+	// metadata. The default value of 0 means no limit will be applied.
```
lidel, line 59:
```suggestion
	// metadata. The value of 0 means no limit will be applied.
```

core/corehttp/gateway_handler_unixfs_dir.go
```diff
@@ -133,9 +134,16 @@ func (i *gatewayHandler) serveDirectory(w http.ResponseWriter, r *http.Request,
 		}
 		dirListing = append(dirListing, di)
 	}
+	var warnMaxDirectorySize string
 	if dirit.Err() != nil {
-		internalWebError(w, dirit.Err())
-		return
+		select {
+		case <-directoryTooBig:
+			warnMaxDirectorySize = fmt.Sprintf("Directories bigger than %d items can't be rendered fully. Try using the CLI: ipfs ls -s --size=false --resolve-type=false %s",
```
lidel, line 141:
```suggestion
			warnMaxDirectorySize = fmt.Sprintf("Directories bigger than %d items can't be rendered fully as the cost of generating HTML would be too big. To see all CIDs in this directory, use the CLI: ipfs ls -s --size=false --resolve-type=false %s",
```

config/gateway.go
lidel: We will be using it with value other than  0, something like `.WithDefault(2500)`. 
That way we can adjust it in future releases without the need for writing config migration.

core/coreapi/unixfs.go
```diff
@@ -302,7 +302,7 @@ func (api *UnixfsAPI) processLink(ctx context.Context, linkres ft.LinkResult, se
 }
 
 func (api *UnixfsAPI) lsFromLinksAsync(ctx context.Context, dir uio.Directory, settings *options.UnixfsLsSettings) (<-chan coreiface.DirEntry, error) {
-	out := make(chan coreiface.DirEntry)
+	out := make(chan coreiface.DirEntry, uio.DefaultShardWidth)
```
lidel, line 305: @schomatis why you changed this?

core/corehttp/gateway.go
```diff
@@ -89,9 +90,10 @@ func GatewayOption(writable bool, paths ...string) ServeOption {
 			}, headers[ACEHeadersName]...))
 
 		var gateway http.Handler = newGatewayHandler(GatewayConfig{
-			Headers:      headers,
-			Writable:     writable,
-			PathPrefixes: cfg.Gateway.PathPrefixes,
+			Headers:             headers,
+			Writable:            writable,
+			PathPrefixes:        cfg.Gateway.PathPrefixes,
+			HTMLDirListingLimit: int(cfg.Gateway.HTMLDirListingLimit.WithDefault(2500)),
```
lidel, line 96: @alanshaw is 2500 a good default limit, or should we increase it?

lidel (approved):
Thank you @schomatis

I took it for a spin and worked as expected, but realized the UX was not the best – it was a hard cut-off, and it is better to show some items to users.  Played a bit with different ways of communicating this and ended up with bit simpler solution:

- we list up to `Gateway.HTMLDirListingLimit`  items
- if limit is reached or surpassed, we dont fetch any remaining blocks, only show an actionable message in header and  footer 
- rebased this on top of https://github.com/ipfs/go-ipfs/pull/8872 (which removed the bindata from assets, making our code way  simpler)

Lmk your thoughts on this.

## Demo

```console
$ ipfs config --json Gateway.HTMLDirListingLimit 2
```

Opening `/ipfs/bafybeihfg3d7rdltd43u3tfvncx7n5loqofbsobojcadtmokrljfthuc7y` 

> Screen Shot 2022-04-12 at 04 13 04

Confirmed we only fetch small number of blocks by opening dir listing on an empty repo, and then shutting down the daemon. Then, in offline mode, you can see only parent and the first few blocks were fetched:

```console
$  ipfs ls -s --size=false --resolve-type=false bafybeihfg3d7rdltd43u3tfvncx7n5loqofbsobojcadtmokrljfthuc7y | wc -l
1864
$ ipfs ls -s bafybeihfg3d7rdltd43u3tfvncx7n5loqofbsobojcadtmokrljfthuc7y | wc -l
3
$ ipfs ls -s bafybeihfg3d7rdltd43u3tfvncx7n5loqofbsobojcadtmokrljfthuc7y
QmbQDovX7wRe9ek7u6QXe9zgCXkTzoUSsTFJEkrYV1HrVR -         1 - Barrel - Part 1/
QmdC5Hav9zdn2iS75reafXBq1PH4EnqUmoxwoxkS5QtuME -         10 - Pi Equals/
QmcyyLvDzCrduuvGVUQEh1DzFvM7UWGfc9sUg87PjjYCw7 -         100 - Family Circus/
Error: ipld: could not find Qmd8NDeJhzf614FSBxZwu4QD2Az14tQtJhQXJf8h4fqiSx
```

lidel:
Note from Stewards sync: 
- change this to always list all items (no  cut-off) because we already have the root block
  - back to `Ls()` code we had before
- when item list is bigger than the limit,  skip size / type check + display header/footer informing the dir was too big to fetch details for all children

core/coreapi/unixfs.go
schomatis: Performance. Otherwise the listing is stalled on the consumer. (The default is arbitrary but seems sane.)

core/corehttp/gateway.go
schomatis: Just to make sure this is the expected UX: if the user has no config set on this, it will silently use this hard-coded number. (Not opposed, just checking.)

schomatis:
From Discord [thread](https://discord.com/channels/806902334369824788/963563468617814126/963563471352500274): we're partially moving back to the `Ls` approach since

> change this to always list all items (no cut-off) because we already have the root block

doesn't apply for the HAMT implementation (normally used for directories with many entries as in this use case).

core/corehttp/gateway.go
```diff
@@ -89,9 +90,10 @@ func GatewayOption(writable bool, paths ...string) ServeOption {
 			}, headers[ACEHeadersName]...))
 
 		var gateway http.Handler = newGatewayHandler(GatewayConfig{
-			Headers:      headers,
-			Writable:     writable,
-			PathPrefixes: cfg.Gateway.PathPrefixes,
+			Headers:               headers,
+			Writable:              writable,
+			PathPrefixes:          cfg.Gateway.PathPrefixes,
+			FastDirIndexThreshold: int(cfg.Gateway.FastDirIndexThreshold.WithDefault(100)),
```
lidel, line 96: :information_source: :question:  
 I've set the default to  100 to decrease load on public gateways, but lmk if should be bumped to something higher.

I picked it based on the worst case scenario where all 100 child nodes have 256KiB root blocs, which  produces  ~25MiB of additional data to be fetched by the node before it can produce HTML response.

docs/config.md
```diff
@@ -648,6 +649,21 @@ Default: `""`
...
+
+This allows for listing big directories with tens of thousands of items
+can be returned as soon as the root node of dag-pb directory is
+resolved, without reading any additional metadata from child nodes.
+
+Setting to 0 will enable fast listings for all directories.
+
+Default: `100`
+
+Type: `optionalInteger`
```
lidel, line 665: :pray: Any way to make this sound less awkward?

lidel (commented):
I did the refactor and pushed the updated version:

- Switched to using inexpensive dir listing via `Ls()` with `ResolveChildren(false)`. 
- Sizes from child nodes are read (fetched) only if the directory is smaller than `Gateway.FastDirIndexThreshold`
- Size column is not present in directories over `Gateway.FastDirIndexThreshold`.
- I removed footer and header because they just felt like noise. 
- Setting `Gateway.FastDirIndexThreshold` to `0` will always produce fast responses without fetching any child nodes.

Feedback appreciated.

docs/config.md
Winterhuman: Maybe something like...

"The minimum number of items in a directory before the Gateway switches to a shallow, faster listing which only considers the root-most items."

core/corehttp/gateway_handler_unixfs_dir.go
```diff
@@ -110,34 +111,43 @@ func (i *gatewayHandler) serveDirectory(ctx context.Context, w http.ResponseWrit
...
-			Size:      size,
-			Name:      dirit.Name(),
-			Path:      gopath.Join(originalUrlPath, dirit.Name()),
+			Size:      "", // no size because we did not fetch child nodes
+			Name:      link.Name,
+			Path:      gopath.Join(originalUrlPath, link.Name),
 			Hash:      hash,
 			ShortHash: shortHash(hash),
 		}
 		dirListing = append(dirListing, di)
```
Jorropo, line 135: Append with known sizes, gc aint that smart *yet*, prealoc pls.

lidel: @Jorropo is https://github.com/ipfs/go-ipfs/pull/8853/commits/13ba0f7d7ccb78b77b420465e012383df8fd416e what you had in mind?

docs/config.md
lidel: Simplified in https://github.com/ipfs/go-ipfs/pull/8853/commits/736d6de0a16b96343045f69ab4a4cb1865ed2b07

core/corehttp/gateway_handler_unixfs_dir.go
Jorropo: I was thinking only of `dirListing` but that better. :)

I should push a refactor only PR that updates thoses `prealloc` find.

core/corehttp/gateway_handler.go
```diff
@@ -766,8 +766,8 @@ func (i *gatewayHandler) buildIpfsRootsHeader(contentPath string, r *http.Reques
 		the last root (responsible for specific article) may not change at all.
 	*/
 	var sp strings.Builder
-	var pathRoots []string
 	pathSegments := strings.Split(contentPath[6:], "/")
+	pathRoots := make([]string, 0, len(pathSegments))
```
Jorropo, line 770: Actually here it doesn't make sense to append as this append is conditional:
```go
if root == "" {
  continue
}
```
In thoses cases, trust append.
```suggestion
	var pathRoots []string
	pathSegments := strings.Split(contentPath[6:], "/")
```

Jorropo (dismissed):

config/gateway.go
```diff
@@ -53,7 +53,15 @@ type Gateway struct {
 	//  }
 	PathPrefixes []string
 
-	// FIXME: Not yet implemented
+	// FastDirIndexThreshold is the minimum number of items in a directory
```
alanshaw, line 56:
```suggestion
	// FastDirIndexThreshold is the maximum number of items in a directory
```

config/gateway.go
```diff
@@ -53,7 +53,15 @@ type Gateway struct {
...
+	// before the Gateway switches to a shallow, faster listing which only
+	// requires the root node. This allows for listing big directories fast,
+	// without the linear slowdown caused by reading size metadata from child
+	// nodes.
+	// Setting to 0 will enable fast listings for all directories.
```
alanshaw, line 61: What is the default? What if it's empty?

docs/config.md
```diff
@@ -648,6 +649,20 @@ Default: `""`
 
 Type: `string` (url)
 
+### `Gateway.FastDirIndexThreshold`
+
+The minimum number of items in a directory before the Gateway switches
```
alanshaw, line 654:
```suggestion
The maximum number of items in a directory before the Gateway switches
```

docs/config.md
```diff
@@ -648,6 +649,20 @@ Default: `""`
...
+to a shallow, faster listing which only requires the root node.
+
+This allows for listing big directories fast, without the linear slowdown caused
```
alanshaw, line 657:
```suggestion
This allows for fast listings of big directories, without the linear slowdown caused
```

core/corehttp/gateway_handler_unixfs_dir.go
```diff
@@ -108,36 +109,46 @@ func (i *gatewayHandler) serveDirectory(ctx context.Context, w http.ResponseWrit
...
 		}
-		hash := resolved.Cid().String()
-
-		// See comment above where originalUrlPath is declared.
+		hash := link.Cid.String()
 		di := directoryItem{
-			Size:      size,
-			Name:      dirit.Name(),
-			Path:      gopath.Join(originalUrlPath, dirit.Name()),
+			Size:      "", // no size because we did not fetch child nodes
```
alanshaw, line 130: For consistency with default size value set below for items we fetch the size for?

```suggestion
			Size:      "?", // unknown size because we did not fetch child nodes or node does not support size
```

core/corehttp/gateway_handler_unixfs_dir.go
```diff
@@ -108,36 +109,46 @@ func (i *gatewayHandler) serveDirectory(ctx context.Context, w http.ResponseWrit
...
 		}
 		dirListing = append(dirListing, di)
 	}
-	if dirit.Err() != nil {
-		internalWebError(w, dirit.Err())
-		return
+
+	// Optimization 2: fetch sizes only for dirs below FastDirIndexThreshold
+	if len(dirListing) < i.config.FastDirIndexThreshold {
+		dirit := dir.Entries()
```
alanshaw, line 141: Why not _always_ fetch size for the first `FastDirIndexThreshold` items so that even big directories get size information for some of their items?

core/corehttp/gateway_handler_unixfs_dir.go
lidel: I removed "?" on purpose: a wall of "?" does not improve UX in any way. 

The "?" makes sense only when dag-pb directory points at non-unixfs CID that does not have size info.

Jorropo: Why are we fetching sizes in the first place ?

It costy and that info I personally don't care. Can't just nuke this code ? Do some people care about that ?

core/corehttp/gateway_handler_unixfs_dir.go
lidel: Speed and simplicity: 
- Size column is either present or not. No confusing "?" after some threshold. 
- People are not tempted to naively parse HTML response to get item sizes.

core/corehttp/gateway_handler_unixfs_dir.go
lidel: Usually it is useless data, but sometimes size column allows user to reason how long the download will take: https://dist.ipfs.io/go-ipfs/v0.12.2

Personally, I'd like to stop fetching sizes by default too, but we should add an opt-in way for fetching them, if user really cares. In my mind that is something we could discuss after this non-invasive PR is merged.

BigLep:
2022-04-21: per verbal, @Jorropo is going to do a quick check and approve/merge.

core/corehttp/gateway.go
Jorropo: Personal opinion: should be 0. But for lack of any better solution. That fine.

Jorropo (approved):
LGTM

Issue #3926, opened 2017-05-16
Saving previously seen nodes for later bootstrapping


dsvi:
#### Version information:
N/A

#### Type:
Enhancement
#### Severity:
Medium
#### Description:

Bootstrap nodes built into ipfs client don't seem to work for me at all. Had to find and add some manually. 
Generally having so centralized bootstrap system for otherwise decentralized network is its weakness:
https://github.com/ipfs/go-ipfs/issues/3908
Should ipfs automatically save all the nodes it seen and the time of last seen, to clean out the list later?
That should make the bootstrapping more reliable and truly decentralized.

whyrusleeping:
Yes! definitely. The blocker here is making sure we store peerstore data to a persistent datastore (right now all peer info is kept in memory).

djdv:
Related anecdote
[Perfect Dark](https://en.wikipedia.org/wiki/Perfect_Dark_(P2P)) relies on users to manually add nodes, the client retains peers like you had mentioned. It seems to work well for their network/peers. You add a peer to connect to, the client will exchange peer lists with them and add all of them to a ledger, on the next bootstrap the client uses that ledger, it attempts to establish a connection with everyone in it, after enough connection failures peers are removed from the ledger while successful ones remain, this keeps it from filling up with too many dead nodes.

Not sure what approach would be good for ipfs in terms of storing peers and managing stored peers for bootstrap, or what other systems have tried.

whyrusleeping:
@djdv that sounds like a great approach. Especially tracking failures for removal.

Cole128:
So did this happen? Thanks.

makeworld-the-better-one:
@Kubuxu @whyrusleeping Hey it's me again. I'm just wondering if this was actually added in ipfs v0.4.12. It would be a useful feature and I can't find anything about it anywhere else. It seems @whyrusleeping added it, but maybe @Kubuxu removed it? This doesn't seem to be in the milestone anymore.

makeworld-the-better-one:
@Kubuxu @whyrusleeping 
What's the state of this?

Stebalien:
@bigs is working on storing the peerstore on disk (https://github.com/libp2p/go-libp2p-peerstore/issues/28). After that, it'll be a matter of remembering which peers tend to reliably be online.

bigs:
yup. we'll also need to consider how we re-initialize our TTLs after rebooting, but this is coming down the pike.

lidel:
[...]

PR #8856, opened 2022-04-07
feat(bootstrap): save connected peers as backup temporary bootstrap ones

schomatis:
**VERY EXPERIMENTAL AT THIS POINT** (we may even want to mask this behind an experimental feature)

Every 10 seconds (hardcoded in `DefaultBootstrapConfig.SavePeersPeriod`, not accessible via config) we save a number of peers we're connected to in 

```
ipfs config --json TempBootstrapPeers
```

If we can't connect to our "official" bootstrap peers (more exactly if we don't hit the target in `DefaultBootstrapConfig.MinPeerThreshold`, again hardcoded, why is this not in the config file?) we resort to connecting to some of the peers from the above list.

## Manual test
```bash
export IPFS_LOGGING="bootstrap=debug"
ipfs config --json TempBootstrapPeers
# null
ipfs daemon
```

Running a few seconds we'll see the log 
```
DEBUG	bootstrap	bootstrap/bootstrap.go:210	saved 8 connected peers (of 8 target) as bootstrap backup in the config
```

`ipfs config --json TempBootstrapPeers` now shows many addresses, probably more than the advertised 8 above, but all those addresses should belong to 8 peers (8 unique `/p2p/12D3K...` strings).

_Stop the daemon._

Backup and remove the official bootstrap peers, **CAREFUL**:
```bash
ipfs config --json Bootstrap | tee bootstrap-peers.back
# List of hardcoded bootstrap peers, like: "/dnsaddr/bootstrap.libp2p.io/p2p/QmNnooDu7bfjPFoTZYxMNLWUQJyrVwtbZg5gBMjTezGAJN"

ipfs config --json Bootstrap []
```

At this point, the normal build should not be able to connect to the network:
```bash
ipfs daemon
# WARN	bootstrap	bootstrap/bootstrap.go:103	no bootstrap nodes configured: go-ipfs may have difficulty connecting to the network

ipfs swarm peers
# nothing
```

With this build:
```bash
ipfs daemon
# DEBUG	bootstrap	bootstrap/bootstrap.go:242	not enough bootstrap peers to fill the remaining target of 4 connections, trying backup list

ipfs swarm peers
# normal amount of peers, connected to the network
```

Restore bootstrap with

```bash
ipfs config --json Bootstrap "$(cat bootstrap-peers.back)"
```

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -22,6 +22,10 @@ type Config struct {
 	Routing   Routing   // local node's routing settings
 	Ipns      Ipns      // Ipns settings
 	Bootstrap []string  // local nodes's bootstrap peer addresses
+	// FIXME(BLOCKING): Review name and documentation.
+	// List of connected peers saved to be used as temporary bootstrap nodes
+	// in case we fail to connect to the official Bootstrap list.
+	TempBootstrapPeers []string
 	Gateway   Gateway   // local node's gateway server options
 	API       API       // local node's API settings
 	Swarm     SwarmConfig
--- a/core/bootstrap/bootstrap.go
+++ b/core/bootstrap/bootstrap.go
@@ -3,10 +3,10 @@ package bootstrap
 import (
 	"context"
 	"errors"
-	"fmt"
 	"io"
 	"math/rand"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	logging "github.com/ipfs/go-log"
@@ -50,6 +50,21 @@ type BootstrapConfig struct {
 	// for the bootstrap process to use. This makes it possible for clients
 	// to control the peers the process uses at any moment.
 	BootstrapPeers func() []peer.AddrInfo
+
+	// FIXME(BLOCKING): Review names, default values and doc.
+	// SavePeersPeriod governs the periodic interval at which the node will
+	// attempt to save connected nodes to use as temporary bootstrap peers.
+	SavePeersPeriod time.Duration
+	// SaveConnectedPeersRatio controls the number peers we're saving compared
+	// to the target MinPeerThreshold. For example, if  MinPeerThreshold is 4,
+	// and we have a ratio of 5 we will save 20 connected peers.
+	// Note: one peer can have many addresses under its ID, so saving a peer
+	// might translate to more than one line in the config (following the above
+	// example that means TempBootstrapPeers may have more than 20 lines, but
+	// all those lines will be addresses of at most 20 peers).
+	SaveConnectedPeersRatio   int
+	SaveTempPeersForBootstrap func([]peer.AddrInfo)
+	LoadTempPeersForBootstrap func() []peer.AddrInfo
 }
 
 // DefaultBootstrapConfig specifies default sane parameters for bootstrapping.
@@ -57,6 +72,12 @@ var DefaultBootstrapConfig = BootstrapConfig{
 	MinPeerThreshold:  4,
 	Period:            30 * time.Second,
 	ConnectionTimeout: (30 * time.Second) / 3, // Perod / 3
+	// FIXME(BLOKING): Review this number. We're making it ridiculously small
+	//  only for testing purposes, but this is saving the peers to the config
+	//  file every time so should not be run frequently. (Original proposal 24
+	//  hours.)
+	SavePeersPeriod:         10 * time.Second,
+	SaveConnectedPeersRatio: 2,
 }
 
 func BootstrapConfigWithPeers(pis []peer.AddrInfo) BootstrapConfig {
@@ -90,6 +111,9 @@ func Bootstrap(id peer.ID, host host.Host, rt routing.Routing, cfg BootstrapConf
 			log.Debugf("%s bootstrap error: %s", id, err)
 		}
 
+		// Exit the first call (triggered independently by `proc.Go`, not `Tick`)
+		// only after being done with the *single* Routing.Bootstrap call. Following
+		// periodic calls (`Tick`) will not block on this.
 		<-doneWithRound
 	}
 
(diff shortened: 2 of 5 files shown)


config/config.go
```diff
@@ -22,6 +22,10 @@ type Config struct {
 	Routing   Routing   // local node's routing settings
 	Ipns      Ipns      // Ipns settings
 	Bootstrap []string  // local nodes's bootstrap peer addresses
+	// FIXME(BLOCKING): Review name and documentation.
+	// List of connected peers saved to be used as temporary bootstrap nodes
+	// in case we fail to connect to the official Bootstrap list.
+	TempBootstrapPeers []string
```
guseggert, line 28: Sorry if this is discussed somewhere and I missed it, but why can't we use a persistent peerstore? Or could we shove this in a record in the top-level datastore? It makes me uncomfortable storing runtime state in the config like this.

guseggert: Also is there a reason we don't want to try these *in addition to* the bootstrap nodes, or do this the other way around (try these first, then fallback to bootstrappers)?

schomatis: > Sorry if this is discussed somewhere and I missed it, but why can't we use a persistent peerstore? Or could we shove this in a record in the top-level datastore? It makes me uncomfortable storing runtime state in the config like this.

No reason. This was underspecified in the original issue AFAICT so I went with the simplest and most straightforward way to implement and test this. Feel free to discuss in the issue and once decided request the change here.

schomatis: > Also is there a reason we don't want to try these in addition to the bootstrap nodes, or do this the other way around (try these first, then fallback to bootstrappers)?

Not mutually exclusive. Take a look at the `bootstrapRound` implementation:

https://github.com/ipfs/go-ipfs/pull/8856/files#diff-006e79be60d4825ac32ec1713b612812059cb3e84f054efc7624d574e2e97717R217

But if the documentation is misleading on this we can reword it however you like.

schomatis: > No reason. This was underspecified in the original issue AFAICT so I went with the simplest and most straightforward way to implement and test this. Feel free to discuss in the issue and once decided request the change here.

Also see original issue discussion in https://github.com/ipfs/go-ipfs/issues/3926#issuecomment-1067246366 where it seems work on processing and storing this at the libp2p level (which is where this should happen, I agree) has been stalled. This PR is going with the MVP proposal which is not the optimal one.

schomatis: > It makes me uncomfortable storing runtime state in the config like this.

Thinking a bit more about this: I do not see this information as runtime _state_. It is dynamic in the sense it's collected and recycled periodically from a running daemon but it does not carry state information. You do not need this to operate the daemon, it's a fallback and one the user can easily manipulate (expand, backup, reduce) if it wants to select/preserve some peers over others (since this is a very naive implementation that select peers randomly without accessing libp2p-level information).

guseggert: The config continuously changes over time even though the user didn't explicitly change anything, which is surprising for config. I think there are multiple benefits to using the root datastore--it's less surprising (config remains static), and it doesn't expose the state directly to users which gives us more implementation flexibility.

guseggert: > Not mutually exclusive. Take a look at the bootstrapRound implementation:

Ah thanks, that's great. Maybe nice to guarantee some mix of well-known bootstrappers + hysteretic bootstrappers, but we can always tweak that over time.

schomatis: > even though the user didn't explicitly change anything, which is surprising for config.

Fair enough. The config is something that should be explicitly written by the user and not implicitly by the daemon.

> I think there are multiple benefits to using the root datastore

I'm fine with this but need it stated more explicitly if I have to make a substantial change like this one. I have trouble sometimes  differentiating between ongoing discussions brainstorming possible directions and actual expressions of concrete changes needed to land a feature. That's the rationale behind my push to move this thread to the original issue if you're sourcing for opinions or make a review request signaling what needs to change.

schomatis: > mix of well-known bootstrappers + hysteretic bootstrappers,

I'm not sure what a hysteretic bootstrapper is but sounds like the kind of information that needs to be sourced from the libp2p layer (in terms of peers score) and not here.

schomatis: > safe play here is to try static and dynamic bootstraps independently, but in parallel

https://github.com/ipfs/go-ipfs/pull/8856#discussion_r846072335

schomatis: > ephemeral peers do not belong to the config file

ACKed in comment above, but I'm still pushing to move this discussion outside this thread that is already becoming too long to be parsed by others.

Jorropo: > storing this at the libp2p level (which is where this should happen, I agree) has been stalled

@schomatis 
FYI, ipfs-cluster do that, it has a text backed peerstore and use that for bootstraping:

https://github.com/ipfs/ipfs-cluster/blob/1707c43b8fb1e6e9063ffb50b5335660c570325b/pstoremgr/pstoremgr.go#L295-L336

Jorropo: > but maybe a flat plain text file at $IPFS_PATH/recent-peers (one multiaddr per line) will bring more utility

That exactly what ipfs-cluster do already, it seems to like we should lift that code from ipfs-cluster into a general `libp2p/...` thing that we can import.

guseggert: > we could move them to peerstore/datastore, but maybe a flat plain text file at $IPFS_PATH/recent-peers (one multiaddr per line) will bring more utility? this way people could share this file with each other, we could even have a website which crawls DHT and gives people N peers for download

Better than config. We could add a command with effectively the same interface as what you're talking about, but witthout exposing the persistence details. Since we know this isn't the ideal implementation, it'd be nice to make it easier on ourselves later to change (e.g. move to libp2p).

Mitumaru (approved):

BigLep:
2022-04-22 conversation: @guseggert will give clear direction on next steps so @schomatis gets unblocked.

guseggert:
@schomatis: let's put this in the repo datastore for now, so that we are free to change/remove it later without breaking folks.

schomatis:
@guseggert Moved to datastore. Ready for review. Will need help with tests if required to land this.

BigLep:
@guseggert : this is a mental note that I'd like to flag this one for code review Friday so we get it over the line.

guillaumemichel:
This feature would be great to have. It enables:
- less load on bootstrap nodes
- faster bootstrap (finding self from the known closest peer still online)
- more balanced RT buckets. Currently empty buckets are filled by querying a _random_ key belonging to this bucket. Hence all the peers inside this bucket will all be very close from this random key, and will not be distributed uniformly in the bucket keyspace, which is desired as it reduces the number of hop during the lookup process.
- less dependence on central bootstrap nodes, more dependence on known peers. Makes the network more decentralized

BigLep:
Thanks for the flags.  This has gotten lost amongst all the other priorities that have emerged so far in 2023.  I have in the list for 0.20.

hacdias:
I just rebased this and I will try taking a look at it during this week or the next week. This would be a great addition.

@Jorropo it would be great if you could take a look since you've reviewed it before.

hacdias:
I went over the code, cleaned up some comments. What's left to review:

- [ ] Decide if new variable names are fine. I have no problems with them.
- [ ] Decide if defaulting `SavePeersPeriod` to 1 hour is okay.
- [ ] Decide if defaulting `SaveConnectedPeersRatio` to 2 is okay.
- [ ] Someone to double-check this: https://github.com/ipfs/kubo/blob/b4d96f5c5d5593b68e30ea2cd91b847a9925d15b/core/bootstrap/bootstrap.go#L327-L339 and remove the comment.

hacdias (commented):
I went over the code, cleaned up some comments and pushed a fix. Also updated the main PR description with how to test manually. The code looks good to me and works well for me. What I think is left to review:

- [x] Decide if new variable names are fine, as well as defaults. I have no problems with them.
- [x] I would like to test this, but I don't see how we'd do it.

cc @ipfs/kubo-maintainers

lidel (changes requested):
Some nits below, but it does not have to be perfect, anything here is better than nothing.
Important part here is to have regression tests that ensure backup bootstrappers are used.

@hacdias I think testing would live in https://github.com/ipfs/kubo/tree/master/test/cli and be similar to `swarm_test.go` or `dht_legacy_test.go`:
1. configure node to have  `ipfs config --json Bootstrap []` and `Discovery.MDNS.Enabled: false`  (we want it to boot and have no peers)
1.  set bootstrap backup interval to to something small, like 250ms (could be config proposed below inline, or could be env override via [os.LookupEnv](https://pkg.go.dev/os#LookupEnv))
1. start the daemon, confirm it has no peers (expected)
1. start the second peer (with default config and `Discovery.MDNS.Enabled: false`, we want it to be a member of public swarm, but not connect to the first one) and confirm it has some peers
1. manually  connect first one to the second one
1. confirm the first one learned about other peers  (`ipfs swarm peers` longer than 1)
1. shut down both daemons, and then start first one again
1. it should be connected to something thanks to the backed up peers  (`ipfs swarm peers` longer than 1)

(to make this test work offline we could introduce a third node  )

core/bootstrap/bootstrap.go
```diff
@@ -50,13 +50,31 @@ type BootstrapConfig struct {
...
 
 // DefaultBootstrapConfig specifies default sane parameters for bootstrapping.
 var DefaultBootstrapConfig = BootstrapConfig{
-	MinPeerThreshold:  4,
-	Period:            30 * time.Second,
-	ConnectionTimeout: (30 * time.Second) / 3, // Perod / 3
+	MinPeerThreshold:        4,
+	Period:                  30 * time.Second,
+	ConnectionTimeout:       (30 * time.Second) / 3, // Perod / 3
+	SavePeersPeriod:         1 * time.Hour,
```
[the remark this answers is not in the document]
hacdias: @lidel the way it is currently implemented, it will save peers right after the first bootstrap. Then, it runs every interval. So we don't need to wait 1 hour. I will check.

hacdias (approved):
I'm approving this. I will wait for one more approval since I wrote the test 😃

core/bootstrap/bootstrap.go
```diff
@@ -50,13 +50,26 @@ type BootstrapConfig struct {
...
 // DefaultBootstrapConfig specifies default sane parameters for bootstrapping.
 var DefaultBootstrapConfig = BootstrapConfig{
-	MinPeerThreshold:  4,
-	Period:            30 * time.Second,
-	ConnectionTimeout: (30 * time.Second) / 3, // Perod / 3
+	MinPeerThreshold:        4,
+	Period:                  30 * time.Second,
+	ConnectionTimeout:       (30 * time.Second) / 3, // Perod / 3
+	BackupBootstrapInterval: 1 * time.Hour,
+	MaxBackupBootstrapSize:  20,
```
lidel, line 72: :information_source:   We can expose these later under `Internal.*` if needed  (in future PRs).

For now, only `Internal.BackupBootstrapInterval` is exposed for use in the `test/cli/backup_bootstrap_test.go` test.

lidel (approved):
I think this one baked for long enough -- does the basic thing, we can refine in follow-up PRs, but is already way better than nothing.

Thank you all for pushing on this. Added changelog entry, merging when CI passes.

BigLep:
@lidel : Awesome!  I wanted to confirm that the default bootstrap peers will still stay in the list, but not now the list is augmented with additional peers based on past discovery.  

The reason I'm asking is because of probelab's measurements that try and ascertain the DHT network node size based on the number of peerids that connect to it.

hacdias:
@BigLep the default bootstrap peers stays. We add additional backup peers based on our connections. Those will be dialed if the default bootstrap list does not work.

PR #8865, opened 2022-04-09
Dependabot/go modules/go.uber.org/dig 1.14.1

sullltan123:


Jorropo:
@sullltan123 what is the goal of this PR ?

It sees it as your personal development branch, it is cool that you hack around this, but pull requests are used for when you want to send us code that you want us to review and include.

Jorropo:
@sullltan123 I'll close this pull request because I don't see what it is about.

If you want to propose a change, cleanup things and only send the relevent change (instead of having things like `Create .DS_Store`).

Also we manually update dependencies when needed, no need to send pull requests for that.

PR #8871, opened 2022-04-11
fix(tracing): context plumbing in gateway handlers

guseggert:
This ensures that child contexts are passed around between the
handlers so that traces show the call hierarchy correctly.


lidel (approved):
Thanks!

Issue #8783, opened 2022-03-10
Gateway: fast check if CID is in local datastore cache (only-if-cached)


lidel:
## Ecosystem context

Raw block and CAR stream Gateway response types are added in https://github.com/ipfs/go-ipfs/pull/8758 and https://github.com/ipfs/go-ipfs/issues/8769.

This unlocks exciting features to happen in IPFS ecosystem:
- light clients that fetch data from multiple gateways in trustless fashion
  - Mobile web browsers with low impact on battery
  - IoT devices fetching firmware updates
  - etc.
- "transport gateways"
  - easier to implement: no HTML hosting, only Block / CAR
  - lower risk, no automated DMCA takedowns

## What is missing

Ability to do a quick test if data behind a CID is already present in Gateway's local cache

Essentially an equivalent for what we already can test in CLI/RPC API:

- `ipfs block get --offline /ipfs/{cid}` →  errors if block is not present locally
-  `ipfs dag stat --offline /ipfs/{cid}` →  errors if full DAG is not present locally

I think doing this only per block should be enough – checking if entire DAG is present may be too expensive / overkill in practice.

## Why we need it

- Enables inexpensive checks that do not trigger data retrieval → reduced cost of running gateway
- Light clients are able to use multiple gateways more efficiently → improved performance on the client
  - It is fair to assume light clients will have a gateway pool (list).   Such client should be able to probe which gateway has the data in local cache, and is able to respond with it immediately (without hitting DHT/retrieval), and send GET request to one of them.

## How to implement this

`HTTP HEAD` request is a good candidate.  It does not return any payload, only HTTP headers.

Right now, `HEAD` request is being used for shallow preload of root blocks: depending on resource type, it usually triggers block fetch events along the requested content path up to the root block of the final path segment. We can't change this, because it works as expected – clients use HEAD to read `Content-Length` of unixfs files and raw blocks, and that is why root block has to be fetched if it is not present in the local datastore.

This means we need some additional flag to signal we want to do a local datastore check without triggering any additional work. 

How to indicate "no-remote-fetch" when sending HTTP HEAD request for `/ipfs/{cid}`?

Perhaps [RFC 7234#only-if-cached](https://datatracker.ietf.org/doc/html/rfc7234#section-5.2.1.7)?

[...]

PR #8880, opened 2022-04-13
feat: add indicator if CID is cached in local datastore

### LINKED ISSUE #8817: File Browser - Indicator of CID's cached on local node vs network
**Is your feature request related to a problem? Please describe.**

As a user of IPFS, it's difficult to visualize if files are cached on the gateway you're connected to vs if the files need to be retrieved from the network.

**Describe the solution you'd like**
An indicator to show if a CID is cached on the current gateway.  

Example (horrible) mock up:

pin

The first three entries are cached on the gateway, the final one would require pulling from other IPFS nodes.   Three indicators may be needed:

* Fully cached locally
* Partially cached locally
* Not cached locally

xrazis:
Closes #8817.

# Demo

A warning sign appears when the file is not cached locally. 

image

One minor issue is the 404's are printed in the console. That unfortunately cannot be avoided as it is the default behavior of the browser. Let me know if there are changes to be made code or UI wise.


assets/bindata_version_hash.go
```diff
@@ -2,5 +2,5 @@
 package assets
 
 const (
-	BindataVersionHash = "512eb789cd905714e03f29d4e04de7549e8c9c3e"
+	BindataVersionHash = "d36d3139f771eb0319123dbe5df6bc82f9e5306e"
```
Jorropo, line 5: We no longer use bindata, now we use embed, pls rebase your changes on master.

assets/dir-index-html/src/dir-index.html
```diff
@@ -19,6 +19,7 @@
...
 <link rel="stylesheet" href="style.css"/>
 <link rel="stylesheet" href="icons.css">
+<script rel="script" src="script.js" async></script>
```
Jorropo, line 22: Why not do that on the server ? that code:
- break the noscript policy breaking `wget --recursive`.
- generate a of extra rtt and data exchange, the server already know that.

assets/dir-index-html/src/script.js
```diff
@@ -0,0 +1,22 @@
...
+
+    const init = {
+        method: 'GET',
+        headers: {'Accept': 'application/vnd.ipfs.cache.local'}
+    };
+
+    const notCachedHTML = '<div title="File not cached locally" class="ipfs-_attention">&nbsp;</div>';
+
+    listingsPath.forEach((element, index) => {
+        fetch(element, init)
```
Jorropo, line 16: AFAIK this request the full file, we dont want to download GiBs just because you openned the dir listing.

Jorropo (changes requested):
The idea looks good but I think that is gonna be too costy.
Files are splited in 256KiB~2MiB blocks.
It might be possible that only a few of them is there, requiring that code to check all the data.

Maybe it's not so bad if that code is server side instead, or we can do a probabilistic thing where if we found random 10 blocks we say we the file.

assets/dir-index-html/src/dir-index.html
Jorropo: Actualy this was requested for https://github.com/ipfs/go-ipfs/issues/8817#issuecomment-1075793997.
I guess the ETag make sense, but I'm not sure loosing a tiny bit of ETag is worth doing lots of extra requests.

xrazis:
I think this is blocked for now as #8783 is pending. I will move on with requested changes once `application/vnd.ipfs.cache.local` is working.

assets/dir-index-html/src/script.js
lidel: - Even with inexpensive `HEAD`  `application/vnd.ipfs.cache*` check we don't want to send 10k requests when opening big directory like bafybeiggvykl7skb2ndlmacg2k5modvudocffxjesexlod2pfvg5yhwrqm 
   - Even 1k `QmdmQXB2mzChmMeKY47C43LxUdg1NDJ5MWcKMKxDu7RgQm` will be wasteful, if only a small subset is visible on the screen.

This should be refactored to fetch status only for visible items – see https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API

lidel (changes requested):
Some notes from go-ipfs sync today:

- This PR is unlikely to be accepted, unless we address performance concerns:
  -  simple `HEAD`, even if we have basic `application/vnd.ipfs.cache.block` won't work because dir listing already fetches root blocks of children to tell their size, so every item would be "present". 
  - only other check is to see if entire dag is in cache, and that is expensive
     - we would need `application/vnd.ipfs.cache.dag` check that is equivalent to running `ipfs dag stat --offline $CID` for every child  to see if the entire DAG is present
      - this is not something public gateways should handle, would be disabled there, but we could enable this on localhost gateway
      
Overall, a lot of complexity for a little UX indicator, I am afraid.
If you still want to work on this, we need https://github.com/ipfs/go-ipfs/issues/8783 first, then discuss performance, and then get back to this PR.

lidel:
Closing this PR, let's continue in https://github.com/ipfs/go-ipfs/issues/8783  – design decisions need to happen there, which will inform next steps (if any).

PR #8885, opened 2022-04-14
chore(gw): extract logical functions to improve readability

justincjohnson:
Another take at https://github.com/ipfs/go-ipfs/pull/8883 from my personal account this time.

### DIFF
--- a/core/corehttp/gateway_handler.go
+++ b/core/corehttp/gateway_handler.go
@@ -97,7 +97,7 @@ func (r *requestError) Error() string {
 	return r.Err.Error()
 }
 
-func newRequestError(message string, err error, statusCode int) error {
+func newRequestError(message string, err error, statusCode int) *requestError {
 	return &requestError{
 		Message:    message,
 		Err:        err,
@@ -753,9 +753,8 @@ func (i *gatewayHandler) buildIpfsRootsHeader(contentPath string, r *http.Reques
 	return rootCidList, nil
 }
 
-func webRequestError(w http.ResponseWriter, err error) {
-	re := err.(*requestError)
-	webError(w, re.Message, re.Err, re.StatusCode)
+func webRequestError(w http.ResponseWriter, err *requestError) {
+	webError(w, err.Message, err.Err, err.StatusCode)
 }
 
 func webError(w http.ResponseWriter, message string, err error, defaultCode int) {
@@ -884,7 +883,7 @@ func debugStr(path string) string {
 	return q
 }
 
-func handleUnsupportedHeaders(r *http.Request) (err error) {
+func handleUnsupportedHeaders(r *http.Request) (err *requestError) {
 	// X-Ipfs-Gateway-Prefix was removed (https://github.com/ipfs/go-ipfs/issues/7702)
 	// TODO: remove this after  go-ipfs 0.13 ships
 	if prfx := r.Header.Get("X-Ipfs-Gateway-Prefix"); prfx != "" {
@@ -925,7 +924,7 @@ func handleProtocolHandlerRedirect(w http.ResponseWriter, r *http.Request, logge
 
 // Disallow Service Worker registration on namespace roots
 // https://github.com/ipfs/go-ipfs/issues/4025
-func handleServiceWorkerRegistration(r *http.Request) (err error) {
+func handleServiceWorkerRegistration(r *http.Request) (err *requestError) {
 	if r.Header.Get("Service-Worker") == "script" {
 		matched, _ := regexp.MatchString(`^/ip[fn]s/[^/]+$`, r.URL.Path)
 		if matched {
@@ -981,7 +980,7 @@ func handleSuperfluousNamespace(w http.ResponseWriter, r *http.Request, contentP
 	return true
 }
 
-func (i *gatewayHandler) handleGettingFirstBlock(r *http.Request, begin time.Time, contentPath ipath.Path, resolvedPath ipath.Resolved) error {
+func (i *gatewayHandler) handleGettingFirstBlock(r *http.Request, begin time.Time, contentPath ipath.Path, resolvedPath ipath.Resolved) *requestError {
 	// Update the global metric of the time it takes to read the final root block of the requested resource
 	// NOTE: for legacy reasons this happens before we go into content-type specific code paths
 	_, err := i.api.Block().Get(r.Context(), resolvedPath)
@@ -995,7 +994,7 @@ func (i *gatewayHandler) handleGettingFirstBlock(r *http.Request, begin time.Tim
 	return nil
 }
 
-func (i *gatewayHandler) setCommonHeaders(w http.ResponseWriter, r *http.Request, contentPath ipath.Path) error {
+func (i *gatewayHandler) setCommonHeaders(w http.ResponseWriter, r *http.Request, contentPath ipath.Path) *requestError {
 	i.addUserHeaders(w) // ok, _now_ write user's headers.
 	w.Header().Set("X-Ipfs-Path", contentPath.String())
 
--- a/core/corehttp/gateway_handler.go
+++ b/core/corehttp/gateway_handler.go
@@ -28,6 +28,7 @@ import (
 	prometheus "github.com/prometheus/client_golang/prometheus"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
+	"go.uber.org/zap"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


core/corehttp/gateway_handler.go
```diff
@@ -324,61 +344,34 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
...
 	logger.Debug("http request received")
 
-	// X-Ipfs-Gateway-Prefix was removed (https://github.com/ipfs/go-ipfs/issues/7702)
-	// TODO: remove this after  go-ipfs 0.13 ships
-	if prfx := r.Header.Get("X-Ipfs-Gateway-Prefix"); prfx != "" {
-		err := fmt.Errorf("X-Ipfs-Gateway-Prefix support was removed: https://github.com/ipfs/go-ipfs/issues/7702")
-		webError(w, "unsupported HTTP header", err, http.StatusBadRequest)
+	var (
+		err          error
+		shouldReturn bool
```
Jorropo, line 349: `shouldReturn` is very undescriptive. It tells *what* you are doing. However I already know *what* you are doing just be reading the code.
```go
if shouldReturn {
  return
}
```
Is good enough.

Here I would name it `requestHandled` instead, as it tells *why* you are doing it.

core/corehttp/gateway_handler.go
```diff
@@ -324,61 +344,34 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
...
-		redirectURL := gopath.Join("/", u.Scheme, u.Host, path)
-		logger.Debugw("uri param, redirect", "to", redirectURL, "status", http.StatusMovedPermanently)
-		http.Redirect(w, r, redirectURL, http.StatusMovedPermanently)
+	err, shouldReturn = handleProtocolHandlerRedirect(w, r, logger)
+	if err != nil {
+		webRequestError(w, err)
+	}
+	if shouldReturn {
 		return
 	}
```
Jorropo, line 363: Because we can't do much better, if we have to return a bool anyway I would also do error handling in the handler.

So I think for a lack of a clearer solution, returning a bool is fine in this case
```suggestion
	if requestHandled := handleProtocolHandlerRedirect(w, r, logger); requestHandled {
		return
	}
```

core/corehttp/gateway_handler.go
```diff
@@ -324,61 +344,34 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
...
-		if u.RawQuery != "" { // preserve query if present
-			path = path + "?" + u.RawQuery
-		}
-
-		redirectURL := gopath.Join("/", u.Scheme, u.Host, path)
-		logger.Debugw("uri param, redirect", "to", redirectURL, "status", http.StatusMovedPermanently)
-		http.Redirect(w, r, redirectURL, http.StatusMovedPermanently)
+	err, shouldReturn = handleProtocolHandlerRedirect(w, r, logger)
+	if err != nil {
+		webRequestError(w, err)
```
Jorropo, line 359: Why aren't you returning here ? (I already know why, because `handleProtocolHandlerRedirect` cannot return an error without returning `true`)

But that is information, a reader that unfamiliar with this code and the compiler doesn't know.
```suggestion
		webRequestError(w, err)
		return
```
To avoid rejoining the false branch, pls terminate here.

core/corehttp/gateway_handler.go
```diff
@@ -324,61 +344,34 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
...
-	if pathErr := contentPath.IsValid(); pathErr != nil {
-		if fixupSuperfluousNamespace(w, r.URL.Path, r.URL.RawQuery) {
-			// the error was due to redundant namespace, which we were able to fix
-			// by returning error/redirect page, nothing left to do here
-			logger.Debugw("redundant namespace; noop")
-			return
-		}
-		// unable to fix path, returning error
-		webError(w, "invalid ipfs path", pathErr, http.StatusBadRequest)
+	if err, shouldReturn = handleSuperfluousNamespace(w, r, contentPath); err != nil {
```
Jorropo, line 371: Same here

core/corehttp/gateway_handler.go
```diff
@@ -785,6 +765,11 @@ func (i *gatewayHandler) buildIpfsRootsHeader(contentPath string, r *http.Reques
 	return rootCidList, nil
 }
 
+func webRequestError(w http.ResponseWriter, err error) {
+	re := err.(*requestError)
+	webError(w, re.Message, re.Err, re.StatusCode)
+}
```
Jorropo, line 771: I like the idea to bundle unpacking here.
But if this function panics if it is not called with a `*requestError`.
This should really be in the signature instead (compiletime checks are better than the same check at runtime):
```suggestion
func webRequestError(w http.ResponseWriter, err *requestError) {
	webError(w, err.Message, err.Err, err.StatusCode)
}
```
I think all the `handle*` function you wrote would need to return `*requestError` too.

Jorropo (changes requested):
This is definitely clearer. thx :slightly_smiling_face: 

Just a few final things.

core/corehttp/gateway_handler.go
```diff
@@ -324,61 +344,34 @@ func (i *gatewayHandler) getOrHeadHandler(w http.ResponseWriter, r *http.Request
...
-		}
-		path := u.Path
-		if u.RawQuery != "" { // preserve query if present
-			path = path + "?" + u.RawQuery
-		}
-
-		redirectURL := gopath.Join("/", u.Scheme, u.Host, path)
-		logger.Debugw("uri param, redirect", "to", redirectURL, "status", http.StatusMovedPermanently)
-		http.Redirect(w, r, redirectURL, http.StatusMovedPermanently)
+	err, shouldReturn = handleProtocolHandlerRedirect(w, r, logger)
```
Jorropo, line 357: I've missed a same here, here

core/corehttp/gateway_handler.go
justincjohnson: I chose `shouldReturn` because it indicates if processing of the request should halt.  Maybe `shouldHalt` would be better?  It isn't as simple as saying the request was handled, because some handling requires halting request processing and some does not.  Does that make sense?

core/corehttp/gateway_handler.go
justincjohnson: The reason the `shouldReturn` check is outside the `err` check is because there is a case where even though an error didn't occur, processing of the request should halt (thus we need to return here).

core/corehttp/gateway_handler.go
Jorropo: I think `shouldHalt` is failing for the same reasons.

First as of [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119) should means recommended but optional.
While this code would be wrong if you didn't returned, so it's a `MUST` not `SHOULD`.
I know 2119 only applies to `SHOULD` and not `should`, but still I think it is confusing.

Secondly `shouldHalt` explain *what* not *why*.
The *what* is already self explainatory, I find `if a { return }` to be perfectly clear, so if the variable could explain instead *why* instead of *what* the code would be clearer because instead of making clearer something that is bright white, it would make clearer something that is unclear.

core/corehttp/gateway_handler.go
justincjohnson: Ah, I see.  You're suggesting that errors should always return and then the extra boolean is just used to indicate that non error conditions should return as well.  Gotcha.  That sounds good to me.  👍

Jorropo: Yes.

I think that:
```
return error > return bool > return error, bool
```
In that case we can't `return error` because there are non erroring cases that still need a termination, so might as well fully commit to a bool.

core/corehttp/gateway_handler.go
justincjohnson: I have pushed changes for your other feedback.  I'm not positive what you're wanting here though.  Are you suggesting that the functions should return both a `*requestError` and an `error`?  I need to read up on how panics are handled.

justincjohnson: I believe you're just saying I should explicitly use `*requestError` instead of `error`.  I've pushed those changes now.

justincjohnson:
I believe I've addressed everything @Jorropo.  Let me know if I missed anything, and thanks again for the feedback.

Jorropo (approved):
LGTM :pray:

core/corehttp/gateway_handler.go
Jorropo: >  I believe you're just saying I should explicitly use *requestError instead of error.

Yeah exactly :)

I see how my `too` was confusing.
It was not "return `error` and `*requestError` too". But it was "do the same change as `webRequestError`" too.

PR #8886, opened 2022-04-14
feat(unixfs): use cidv1 by default

schomatis:
Toward https://github.com/ipfs/go-ipfs/issues/4143. Consumes dependency upgrade https://github.com/ipfs/go-unixfs/pull/119.

@lidel Take a look at this change to see if it satisfies the CIDv1 upgrade for `ipfs files` and `ipfs init`. There's nothing in this PR but the upgrade in UnixFS. I think the ultimate solution is landing https://github.com/ipfs/go-merkledag/issues/86, but this makes a small modification in the middle layer of UnixFS to set a CIDv1 by default without much propagating big changes.

I'm punting on fixing sharness until I have the modifications for the rest of the commands as I'll likely need to write a small script to parse CIDv0 and upgrade to CIDv1 all over those files. This is just a request for a manual test here and a code review in the dependency upgrade to see if I'm going in the right direction.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -50,11 +50,11 @@ require (
 	github.com/ipfs/go-merkledag v0.6.0
 	github.com/ipfs/go-metrics-interface v0.0.1
 	github.com/ipfs/go-metrics-prometheus v0.0.2
-	github.com/ipfs/go-mfs v0.2.1
+	github.com/ipfs/go-mfs v0.2.0
 	github.com/ipfs/go-namesys v0.4.0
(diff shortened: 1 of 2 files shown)


lidel:
Closing in favor of #8185, which makes CIDv1 the default for new repos via the `Import` config and the IPIP-499 `unixfs-v1-2025` profile. Existing repos stay on CIDv0 until they opt in, so no migration is involved. Thanks for the early push here; #8185 is where this continues.

PR #8888, opened 2022-04-15
refactor(gw): faster dir listing - fetch child sizes in parallel

schomatis:
_Broken and untested. Just to gather early feedback._

@lidel This is a variant of your optimization fetching metadata in parallel to avoid the sequential stall. Need to work some more on it (document, test, encapsulate) but wanted to check if (a) this is a feature you're interested in and (b) this is an additional level of complexity you're comfortable with in this part of the code.

### DIFF
--- a/core/corehttp/gateway_handler_unixfs_dir.go
+++ b/core/corehttp/gateway_handler_unixfs_dir.go
@@ -15,6 +15,7 @@ import (
 	"github.com/ipfs/go-ipfs/tracing"
 	path "github.com/ipfs/go-path"
 	"github.com/ipfs/go-path/resolver"
+	unixfile "github.com/ipfs/go-unixfs/file"
 	options "github.com/ipfs/interface-go-ipfs-core/options"
 	ipath "github.com/ipfs/interface-go-ipfs-core/path"
 	"go.opentelemetry.io/otel/attribute"
@@ -111,6 +112,11 @@ func (i *gatewayHandler) serveDirectory(ctx context.Context, w http.ResponseWrit
 
 	// storage for directory listing
 	var dirListing []directoryItem
+	type entryRequest struct {
+		entryIndex int
+		cid cid.Cid
+	}
+	var entryRequests []*entryRequest
 
 	// Optimization 1:
 	// List children without fetching their root blocks (fast, but no size info)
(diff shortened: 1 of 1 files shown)


core/corehttp/gateway_handler_unixfs_dir.go
```diff
@@ -133,20 +139,92 @@ func (i *gatewayHandler) serveDirectory(ctx context.Context, w http.ResponseWrit
...
+	}()
+	type entryInfo struct {
+		entryIndex int
+		size string
+	}
+	out := make(chan entryInfo)
+	errChan := make(chan error)
+
+	for c := 0; c < defaultConcurrentFetch; c++ {
+		go func() {
```
Jorropo, line 173: Do we really need that many goroutines ?
Why not use `GetMany` ? which accepts a list of CIDs and return a channel.

core/corehttp/gateway_handler_unixfs_dir.go
```diff
@@ -133,20 +139,92 @@ func (i *gatewayHandler) serveDirectory(ctx context.Context, w http.ResponseWrit
...
+			for r := range feed {
+				nd, err := i.api.Dag().Get(fetchersCtx, r.cid)
+				if err != nil {
+					select {
+					case errChan <- err:
+					case <-fetchersCtx.Done():
+					}
+					return
+				}
+				file, err := unixfile.NewUnixfsFile(fetchersCtx, i.api.Dag(), nd)
```
Jorropo, line 183: That code also could be done sequentially.

At this point it we have all the info we need in `nd` (type is in `nd.Data.Type` and the size is `len(nd.Data.Data) + sum(nd.Data.Blocksizes[])`).

Jorropo (changes requested):

core/corehttp/gateway_handler_unixfs_dir.go
schomatis: > Do we really need that many goroutines ?

The number is up for discussion, I don't think 32 is not _that_ many though in terms of performance impact if that's your concern (the number was extracted from the DAG walk for example).

> Why not use GetMany ? 

I'm not sure I want to subscribe to 10K requests in BS and then cancel the context with the remaining un-served ones once we reach the `FastDirIndexThreshold` threshold, but I'm definitely not happy with the complexity I'm adding here. If lidel greenlights this approach I'll look for some inbetween like batching some number of requests in each goroutine with `GetMany`.

core/corehttp/gateway_handler_unixfs_dir.go
schomatis: The size reported by the `ProtoNode` at the DAG level (size of the single node) is not the same as the one stored one layer above in UnixFS (size of all the subtree).

core/corehttp/gateway_handler_unixfs_dir.go
Jorropo: I was expecting something really simple, like `GetMany` them 32 by 32 sequentially. I think this approach would even be simpler than what you wrote.

I don't think it's worth fixing in this PR, someone else will file a new PR if that approach is found to

core/corehttp/gateway_handler_unixfs_dir.go
Jorropo: I meant the sizes in the protobuf schema (which is probably the only thing worth stating anyway).

core/corehttp/gateway_handler_unixfs_dir.go
```diff
@@ -133,20 +139,92 @@ func (i *gatewayHandler) serveDirectory(ctx context.Context, w http.ResponseWrit
...
+					case <-fetchersCtx.Done():
+					}
+					return
+				}
+				size := "?"
+				if s, err := file.Size(); err == nil {
+					// Size may not be defined/supported. Continue anyways.
+					// FIXME: Check above. The UnixFS files we're iterating
+					//  (because we use the UnixFS API) should always support
+					//  this.
```
lidel, line 196: iirc we need to keep this check because we've  seen dag-pb dir that links to dag-cbor or dag-json CIDv1

lidel (commented):
Sounds like a good idea, but let's wait for  https://github.com/ipfs/go-ipfs/pull/8853#pullrequestreview-942930258 to land first (and decide on default threshold there).

To avoid having too much on our plate, I'm marking this for go-ipfs 0.14.

core/corehttp/gateway_handler_unixfs_dir.go
schomatis: A sequential approach is what we're changing away from. If you request 32 blocks, 31 are delivered and one is missing you're stuck just the same than if you request one at a time like the previous implementation.

Jorropo: > If you request 32 blocks, 31 are delivered and one is missing you're stuck

AFAIK, That already what we do for `ipfs pin add` because of the scheduling costs gains.

Resolving, this only need to be looked at if that show up in someone's profile.

schomatis:
@lidel Your call, but note that this PR wasn't about just _another_ performance optimization but addressing the core of the issue (at least what I understand it to be). We probably didn't groom Steven's original issue description as much as we should and the principal objective is still a bit unclear to me, my takeaway was @alanshaw's concern about the [GW timing out](https://github.com/ipfs/go-ipfs/issues/8178#issuecomment-1090437076):

> [users] are getting confused when trying to view [big directories just uploaded] on the gateway - it times out and they assume their upload did not complete successfully.

The core of the timeout is fetching directory entries metadata _sequentially and without a per-entry timeout_ as we still do now in https://github.com/ipfs/go-ipfs/pull/8853, only with the threshold to avoid the fetching if we have too many entries, but note this timeout will still happen even with a few entries if only one of them is missing:

```
mkdir big-dir
touch big-dir/file{1..5}
echo "different file" > big-dir/file1
BIG_DIR=$(ipfs add big-dir -r -Q)
echo "http://localhost:8080/ipfs/$BIG_DIR"
# http://localhost:8080/ipfs/QmYf9xX1TrQA3vxoQRkYWQT9ugWsdg7fihMg1w7ZLg2UiN

ipfs ls QmYf9xX1TrQA3vxoQRkYWQT9ugWsdg7fihMg1w7ZLg2UiN
# QmRij737cvMJxHMPM6foyPYfVrcWnZsiVF6yvDZknNiG1j 15 file1       <<<< Remove this entry
# QmbFMke1KXqnYyBBWxB74N4c5SBnJMVAiMNRcGu6x1AwQH 0  file2
# QmbFMke1KXqnYyBBWxB74N4c5SBnJMVAiMNRcGu6x1AwQH 0  file3
# QmbFMke1KXqnYyBBWxB74N4c5SBnJMVAiMNRcGu6x1AwQH 0  file4
# QmbFMke1KXqnYyBBWxB74N4c5SBnJMVAiMNRcGu6x1AwQH 0  file5

# Remove `file1`: QmRij737cvMJxHMPM6foyPYfVrcWnZsiVF6yvDZknNiG1j
rm ~/.ipfs/blocks/PX/CIQDEOWO6TJANTWQUC2DBXZB7K4XA5CPPYMIOGAN2HH3MGWIM3FUPXQ.data
ipfs ls QmRij737cvMJxHMPM6foyPYfVrcWnZsiVF6yvDZknNiG1j # not found
```

Browsing http://localhost:8080/ipfs/QmYf9xX1TrQA3vxoQRkYWQT9ugWsdg7fihMg1w7ZLg2UiN even with 4 entries will still timeout on the first one not being available.

The objective of this PR (not fully expanded before, sorry) is to attempt to fetch _any_ `FastDirIndexThreshold` entries metadata in the directory even if there are some slow/missing ones without tying ourselves to only the first `FastDirIndexThreshold` entries in the list.

core/corehttp/gateway_handler_unixfs_dir.go
schomatis: Being discussed in https://github.com/ipfs/go-ipfs/pull/8888#issuecomment-1101591394 for lidel to resolve.

lidel:
Quick triage notes:

### Two separate problems

In my mind the _core issue_ reported was the unmovable reality of big directories fetching tens of thousands of child blocks:  even when all blocks are reachable, and even if we did the parallel thing from this PR, big enough dir would take ages, and could hit timeout on caching/reverse proxy or client.  That is being addressed in https://github.com/ipfs/go-ipfs/pull/8853 by skipping child fetch entirely  for directories above some threshold. 

This PR is about additional improvement that can land separately:

###  This PR:  parallel child size fetch

>  The objective of this PR (not fully expanded before, sorry) is to attempt to fetch any FastDirIndexThreshold entries metadata in the directory even if there are some slow/missing ones

@schomatis is the idea to do the listing in "best-effort" fashion: fetch things in parallel and display "?" next to child nodes that errored / timeouted? or just add parallel fetch to parallelize slow fetches?  If it is the latter, perhaps we could simplify this code by using `GetMany` from DAGService?

schomatis:
> unmovable reality of big directories fetching tens of thousands of child blocks: even when all blocks are reachable,

If this is the original issue then this PR would add more complexity than value here. Closing then.

PR #8893, opened 2022-04-15
chore: deprecate /api/v0/dns

### LINKED ISSUE #8607: Deprecate /api/v0/dns
## Problem

-  `/api/v0/dns` does not work with DNSLinks that have `/ipns/..` content paths (https://github.com/ipfs/go-ipfs/issues/6454), and `/api/v0/resolve` works fine will all content paths.
   - IPFS companion switched its DNSLink detection to `/api/v0/resolve` a while ago. It is passing DNSLink names as content paths:  `/ipns/example.com`  and they work fine no matter if IPNS is used or not.
- Having two ways of resolving DNSLink is confusing enough, having one that fails when content path is `/ipns/` is simply harmful, and even if we fix that, adds unnecessary complexity to the API surface. 
   - In reality, users just want to resolve a content path. They dont care if DNS or IPNS is involved. What matters is resolving content path. Either one level, or all the way, until immutable `/ipfs/{cid}` is hit.

## Proposed solution

Deprecate `/api/v0/dns`:
* remove it from cli `--help` and hide from API docs
* to ensure nothng breaks, keep `/api/v0/dns` endpoint, <del>but remove its custom DNS lookup code, and switch it to use `/api/v0/resolve` internally</del>

lidel:
Closes #8607  – will be picked up by https://github.com/ipfs/ipfs-docs/pull/1086 after go-ipfs 0.13 ships.

### DIFF
--- a/core/commands/dns.go
+++ b/core/commands/dns.go
@@ -17,40 +17,17 @@ const (
 )
 
 var DNSCmd = &cmds.Command{
+	Status: cmds.Deprecated, // https://github.com/ipfs/go-ipfs/issues/8607
 	Helptext: cmds.HelpText{
-		Tagline: "Resolve DNS links.",
+		Tagline: "Resolve DNSLink records.",
 		ShortDescription: `
-Multihashes are hard to remember, but domain names are usually easy to
-remember.  To create memorable aliases for multihashes, DNS TXT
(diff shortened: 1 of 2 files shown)


lidel:
Merging as per triage verbal from today's sync.

Issue #8703, opened 2022-01-28
Error on ipfs add --hash=sha1, but allow with --force


lidel:
### Checklist

- [X] My issue is specific & actionable.
- [X] I am not suggesting a protocol enhancement.
- [X] I have searched on the [issue tracker](https://github.com/ipfs/go-ipfs/issues?q=is%3Aissue) for my issue.

## Problem

> Suggestion extracted from https://github.com/ipfs/go-ipfs/issues/8650

The default is SHA2-256, but we support SHA1 for legacy reasons.
SHA1 was broken, and systems like git moved away from it, but we are not warning user when they opt-in to using it.

## Solution

We can't remove support for it, because  there should be a way for people to migrate the data.

There should be a list of "legacy/deprecated" hash functions like SHA1 which would be discouraged:

```console
$ ipfs add --hash=sha1 file
Error: selected hash function is no longer secure; use --hash=sha2-256 or pass --force

$ $ ipfs add --hash=sha1 file
Added ...
```

aschmahmann:
Basically all the insecure hash functions are already forbidden by https://github.com/ipfs/go-verifcid/blob/46876f413195d36227ee5c7d6f16e995b62f754d/validate.go#L15. The only exception here is SHA1 because at the time it was deemed valuable (and I assume insufficiently broken) https://github.com/ipfs/go-verifcid/blob/46876f413195d36227ee5c7d6f16e995b62f754d/validate.go#L30.

This means that at the moment this really only applies to SHA1.

-----

We could also just make using SHA-1 safer by utilizing the cryptoanalytic "fix" used in places like Git https://github.com/cr-marcstevens/sha1collisiondetection.

To do this there are two pieces:
1. Make https://github.com/ipfs/distributions be able to build with CGO
2. Have go-ipfs use something like https://github.com/aschmahmann/go-sha1collisiondetection (CGO wrapper around the C library with the cryptoanalytic fix). Work on the wrapper started in a PR to that repo.

Related: https://github.com/multiformats/go-multihash/issues/57

Jorropo:
3. Rewrite or machine translate the cryptoanalytic fix in Go.

I guess it's mainly just doing arithmetic on byte buffers, this should be trivial to machine translate (and lift up manually).

aschmahmann:
> We should avoid relying on CGO for core features.

[...]

Issue #8696, opened 2022-01-24
fix: dag|block put: do not allow blocks over BS limit


lidel:
Update 2/23 (@schomatis): working on this.

------------------------------------------

## Summary

TLDR:  bigger blocks can't be exchanged over libp2p, so we want to make a loud error if that happens, but allow people to override it if they are ok with that.

We've introduced soft-limit in `ipfs object` API (https://github.com/ipfs/go-ipfs/pull/8414) but are missing  the same behavior in modern `ipfs dag|block put` :see_no_evil:  

## Problem (go-ipfs 0.11)

With both `dag|block put`,  I am able to create a block (eg. 10MB)  that is too big to be exchanged over bitswap:

### `dag put`

```console
$ dd if=/dev/zero bs=10MB count=1 | ipfs dag put --input-codec=raw --store-codec=raw
1+0 records in
1+0 records out
10000000 bytes (10 MB, 9.5 MiB) copied, 0.0373717 s, 268 MB/s
bafkreihv4avkohth6qoxsar2ckgkgw5nq3hxwzswsz574ceewor4ims6v4

$ ipfs dag stat bafkreihv4avkohth6qoxsar2ckgkgw5nq3hxwzswsz574ceewor4ims6v4
Size: 10000000, NumBlocks: 1
```

### `block put`

```console
$ dd if=/dev/zero bs=10MB count=1 | ipfs block put
1+0 records in
1+0 records out
10000000 bytes (10 MB, 9.5 MiB) copied, 0.0331271 s, 302 MB/s
QmetSrEEUmD5z6wS5Eamn5KrizYsdkthqNioEoMepYFQte

$ ipfs block stat QmetSrEEUmD5z6wS5Eamn5KrizYsdkthqNioEoMepYFQte
Key: QmetSrEEUmD5z6wS5Eamn5KrizYsdkthqNioEoMepYFQte
Size: 10000000
```

## Solution

- [x] reuse `softBlockLimit`,  `allowBigBlock` from #8414 (extract them to be public consts?)
- [x] add `--allow-big-block` behavior to `ipfs dag put`
- [x] add `--allow-big-block` behavior to `ipfs block put`
- [ ] make `dag` and `put` return informative error, eg: 
     > produced block is over 1MB: big blocks can't be exchanged with other peers. consider using `dag-pb` for automatic chunking of bigger files, or pass --allow-big-block to override.

schomatis:
Working on this.

[...]

PR #8895, opened 2022-04-18
fix: refuse creating new DAGs with insecure hash functions (SHA1)

finkbeca:
Closing #8703, added an optional forcing flag for `ipfs add` to discourage use of broken hashes like SHA-1 while still allowing it, given you pass in an optional forcing flag. Tested this manually due to the small scope of the changes.

```console
$ ipfs add --hash=sha1 file
Error: selected hash function is no longer secure; use --hash=sha2-256 or pass --force.
```

```console
$ ipfs add --hash=sha1 --force file
Added ...
```


guseggert:
Thanks for this!

A more specific flag would make it more obvious that a less secure hash function is being used, like `--allow-insecure-hash-function` or something like that. Also can you add a sharness test for this?  Should be really easy.

finkbeca:
Thanks for the feedback. I fixed suggested the edits changing the flag name to `--allow-insecure-hash-function` and added a sharness test for the new command.  Hope that cover the change requests. :)

test/sharness/t0047-add-forcehash.sh
```diff
@@ -0,0 +1,47 @@
...
+    ipfs add --hash=sha2-256 --allow-insecure-hash-function mountdir/hello.txt >actual
+    '
+    test_expect_success "'ipfs add --allow-insecure-hash-function succeeds" '
+    ipfs add --hash=sha2-256 --allow-insecure-hash-function mountdir/hello.txt >actual
+    '
+     test_expect_success "ipfs add --allow-insecure-hash-function succeeds output looks good" '
+    HASH="QmVr26fY1tKyspEJBniVhqxQeEjhF78XerGiqWAwraVLQH" &&
+    echo "added $HASH hello.txt" >expected &&
+    test_cmp expected actual
+  '
```
Jorropo, line 36: You should add a test that tries to add a file with an unsecure function and check that it *fails*.

core/commands/add.go
```diff
@@ -181,11 +185,15 @@ only-hash, and progress/status related flags) will change the final hash.
 		inline, _ := req.Options[inlineOptionName].(bool)
 		inlineLimit, _ := req.Options[inlineLimitOptionName].(int)
 
+		forceHash, _ := req.Options[forceHashOptionName].(bool)
 		hashFunCode, ok := mh.Names[strings.ToLower(hashFunStr)]
+		if !forceHash && strings.ToLower(hashFunStr) == "sha1" {
```
Jorropo, line 190:
```suggestion
		loweredHash := strings.ToLower(hashFunStr)
		hashFunCode, ok := mh.Names[loweredHash]
		if !forceHash && loweredHash == "sha1" {
```

Jorropo (changes requested):
Thanks, this is definitely a feature we want to merge, however a few things needs some tweaking first. :slightly_smiling_face:

finkbeca:
Fixed the syntax change @Jorropo , also added failure tests specifically for SHA1. I'm aware that the Patch only works for SHA1 at this time, I can change the wording of the error if you think that would be better. My initial thought that generalizing the error statement will simplify future extensions.

lidel (changes requested):
Thank you, @finkbeca :heart: 

Would you have time to add the same flag to `dag put` and `block put` – see  ipfs/go-ipfs#8696 for prior art of (`--allow-big-block`)?  If not, let us know, someone else will push this over the finish line.

Jorropo (changes requested):
Sorry to ask for more but this does not work.
```
ipfs add --hash=md5 --allow-insecure-hashes randomFileHere
```
Doesn't work.

I know this is a more complicated task, but I would like to avoid yet an other half working feature.
I don't know what other maintainers think of this.

lidel:
@Jorropo Sounds sensible. Do you have more things we should block?

@finkbeca Mind replacing static "sha1" with a check against a list of insecure hashes? 
That way, we can add them to some constant over time.

Jorropo:
> Mind replacing static "sha1" with a check against a list of insecure hashes ["sha1", "md5"]?
That way, we can add more as time goes by.

We already have that, it is <https://github.com/ipfs/go-verifcid>.

> Do you have more things we should block?

- [ ] Refactor go-verify CID consumers, to work on the command level (instead of deeper in the code like in add's internal) so we can disable it more easly. (what I am already asking for)
- [ ] Remove SHA1 from go-verify CID's safe list. (or only allows the "safe" cryptoanalysis fix, see https://shattered.io/)

Jorropo:
> I don't see a particularly good reason to bother with doing go-verifcid plumbing everywhere to allow sets of insecure hash functions. Note: you likely wouldn't want it to be all or nothing... allowing SHA-1 doesn't necessarily mean allowing a 1 bit SHA-2 hash or MD5.

I'm saying this is what we *currently* more less have.
I want to scrape that, and only do the checking in commands handlers (where adding `--allow-insecure-hash` is trivial to implement).

> Nothing at the library or protocol level stops you from using broken hash functions like MD4, but I don't see a reason why go-ipfs should support using it.
```
$ rgrep -I "verifcid" vendor/ | cut -f1 -d':' | while read f; do dirname ${f}; done | uniq | grep -v verifcid | grep -v "^vendor$"
vendor/github.com/ipfs/go-ipfs-provider/simple
vendor/github.com/ipfs/go-ipfs-provider/batched
vendor/github.com/ipfs/go-blockservice
```

aschmahmann:
> I want to scrape that, and only do the checking in commands handlers (where adding --allow-insecure-hash is trivial to implement).

This seems unlikely to work. For example `ipfs cat bafysomesha2` internally has a pointer to an MD4 hash -> game over. You still need this pretty low down the stack.

> Nothing at the library or protocol level stops you from using broken hash functions like MD4, but I don't see a reason why go-ipfs should support using it.

Ok fair enough there are a few library pieces where you'd have to mock out the interfaces or do a fork to remove the checks. IIRC those structs aren't explicitly used rather than their interfaces so not too bad to deal with. However, if there's actual demand for it it's probably fine to add some configurability.

Jorropo:
> This seems unlikely to work. For example `ipfs cat bafysomesha2` internally has a pointer to an MD4 hash -> game over. You still need this pretty low down the stack.

Ok fair enough too.

We can put that value in the context (create a `allow-insecure-hash` bool key), and let verifcid take a context and use that to know if we allows it or not. (we can't use globals obviously since this parameter is request dependent)

Jorropo:
Sorry but we are not intrested in this fix,
It's not actually doing what it claims, I can't use it to add files with MD4.

**But mainly.**
it only supports `ipfs add --hash=sha1`
Which is not what is important.

**The danger is in `ipfs get`** (and gateway, ipfs cat, ... whatever that read data), because some malicious attacker can create a SHA256 block linking a bad SHA1 hash and send you whatever data they want instead.

We will follow the same route as git, using the sha1 fix to make the function "safe" again. (the only reason sha1 is still there is for git anyway so using the same fix as it make sense).

The context for this is here: https://github.com/ipfs/go-ipfs/issues/8703#issuecomment-1155758989

PR #8898, opened 2022-04-21
feat(test): migrate CIDv0-v1 tool

schomatis:
Tool to migrate strings in sharness tests to CIDv1 for PRs related to https://github.com/ipfs/go-ipfs/issues/4143.

This doesn't necessarily need to land in the repo, though it might be useful for future migrations, but submitting as a PR to get feedback. Will fine-tune and apply it after the related PRs have a review to leave sharness migration for last.

### DIFF
--- a/test/migrate-cid/main.go
+++ b/test/migrate-cid/main.go
@@ -1,0 +1,102 @@
+package main
+
+import (
+	"bufio"
+	"bytes"
+	"io/ioutil"
+	"log"
+	"os"
+	"strings"
+
+	cid "github.com/ipfs/go-cid"
+)
+
+// Command to parse a file and replace CIDv0 with CIDv1 strings.
+// Single argument: the path to the file where the in-place replace will happen.
+
+// These are hardcoded in `cid.Decode()`.
+const CID_V0_PREFIX = "Qm"
+const CID_V0_LENGTH = 46
+
+func main() {
+	if len(os.Args) < 2 {
+		log.Fatal("Missing argument with file path to replace CIDs.")
+	}
+	replaceFile := os.Args[1]
+
+	input, err := ioutil.ReadFile(replaceFile)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


test/migrate-cid/main.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+	replaceFile := os.Args[1]
+
+	input, err := ioutil.ReadFile(replaceFile)
+	if err != nil {
+		log.Fatal(err)
+	}
+	scanner := bufio.NewScanner(bytes.NewReader(input))
+
+	outputFile, err := os.OpenFile(replaceFile, os.O_WRONLY|os.O_TRUNC, 0)
```
Jorropo, line 33: I see what you are trying to do here.
However I don't think slurping everything into ram is a great solution.

There are a few cannonical solutions.
The best one here is likely
- write to a tmp file like `replaceFile + ".tmp"`
- then move `replaceFile + ".tmp"` to `replaceFile`
- (if the translation errors, just remove the `.tmp` file).

Jorropo (changes requested):

test/migrate-cid/main.go
schomatis: > However I don't think slurping everything into ram is a great solution.

It is not, but it perfectly acceptable for consuming small, version-controlled source files like sharness tests. I see no added value in the "canonical" alternatives. The only possible source of error that would leave the rewrite mid-flight is a scan error, which with an in-memory input is very unlikely. Let's not waste time over-engineering a side tool that might not even land in the repo while the main PRs are still being processed.

lidel:
Closing in favor of #8185, which makes CIDv1 the default for new repos via the `Import` config and the IPIP-499 `unixfs-v1-2025` profile. Existing repos stay on CIDv0 until they opt in, so no migration is involved. Thanks for the early push here; #8185 is where this continues.

Issue #8761, opened 2022-03-03
Integrate/expose go-libp2p resource manager


BigLep:
go-libp2p v18 is shipping with a resource manager - yahoo!  This issue encompasses the work to fully land this as a feature exposed to go-ipfs users.  This includes:
- [x] Integrating go-libp2p (i.e., doing the dependency update): https://github.com/ipfs/go-ipfs/pull/8680:  
  - Estimate: 1 - needs review & cleanup – wip @lidel  – see https://github.com/ipfs/go-ipfs/pull/8680#issuecomment-1086432509
- [x] Creating a command for resource manager stats: https://github.com/ipfs/go-ipfs/issues/8722 
  - Covered in the PR above https://github.com/ipfs/go-ipfs/pull/8680 – see https://github.com/ipfs/go-ipfs/pull/8680#issuecomment-1086432509
- [x] Expose stats to prometheus: https://github.com/ipfs/go-ipfs/pull/8785 (will be easier once have done #8680): 
  - Estimate: .5
- [x] Make it self-service for go-ipfs users to configure the resource manager limits (Lotus example: https://github.com/filecoin-project/lotus/pull/8318) .  – see https://github.com/ipfs/go-ipfs/pull/8680#issuecomment-1086432509
 - [x] For now use `$IPFS_PATH/limit.json` file + implicit defaults, not guaranteeing a contract.  We want to learn about ergonomics from testing in production.  In the end we may switch to `Swarm.ResourceMgr.Limits` config that works the same as the output of `ipfs swarm limit` commands.
   -  see https://github.com/ipfs/go-ipfs/pull/8680 and  https://github.com/ipfs/go-ipfs/issues/8858
   - Estimate: 1
- [x] Implement `ipfs swarm limit [scope] --reset` 
   - tracked in https://github.com/ipfs/go-ipfs/issues/8918 (not a blocker, could be in best-effort track for now)
- [x] Expose Prometheus metrics from ResourceManager: https://github.com/ipfs/kubo/issues/9385
- [ ] Testing in production (and fixing bugs): 
  - Dashboard changes
  - Bootstrappers
  - Clusters
  - Gateway staging 
  - Gateway banking https://github.com/protocol/bifrost-infra/issues/1815
     - https://github.com/protocol/bifrost-infra/pull/1832
[...]

PR #8901, opened 2022-04-21
feat: persist limit changes to config

guseggert:
>  Part of https://github.com/ipfs/go-ipfs/issues/8761, follow-up after https://github.com/ipfs/go-ipfs/pull/8680

This changes the "ipfs swarm limit" command so that When limit changes
are applied via the command line, they are persisted to the repo
config, so that they remain in effect when the daemon restarts.

This also removes limit.json support, to simplify the behavior. The
schema for Swarm.ResourceMgr.Limits is exactly the same as limit.json,
so existing limit.json can be dropped into the IPFS config easily
using something like:

```console 
$ cat ~/.ipfs/config | jq ".Swarm.ResourceMgr.Limits = $(cat limit.json)" | sponge ~/.ipfs/config
```

This upgrades to Resource Manager v0.3.0, which exports the config
schema so that we don't have to maintain our own copy of it.

### DIFF
--- a/config/swarm.go
+++ b/config/swarm.go
@@ -1,5 +1,7 @@
 package config
 
+import rcmgr "github.com/libp2p/go-libp2p-resource-manager"
+
 type SwarmConfig struct {
 	// AddrFilters specifies a set libp2p addresses that we should never
 	// dial or receive connections from.
@@ -137,10 +139,8 @@ type ConnMgr struct {
 // <https://github.com/libp2p/go-libp2p-resource-manager#readme>
 type ResourceMgr struct {
 	// Enables the Network Resource Manager feature
-	Enabled Flag `json:",omitempty"`
-
-	/* TODO: decide if and how we want to expose limits in our config
-	Limits *ResourceMgrScopeConfig `json:",omitempty"` */
+	Enabled Flag                      `json:",omitempty"`
+	Limits  *rcmgr.BasicLimiterConfig `json:",omitempty"`
 }
 
 const (
@@ -150,41 +150,3 @@ const (
 	ResourceMgrProtocolScopePrefix = "proto:"
 	ResourceMgrPeerScopePrefix     = "peer:"
 )
-
-/* TODO: decide if and how we want to expose limits in our config
-type ResourceMgrLimitsConfig struct {
-	System    *ResourceMgrScopeConfig `json:",omitempty"`
-	Transient *ResourceMgrScopeConfig `json:",omitempty"`
-
-	ServiceDefault     *ResourceMgrScopeConfig           `json:",omitempty"`
-	ServicePeerDefault *ResourceMgrScopeConfig           `json:",omitempty"`
-	Service            map[string]ResourceMgrScopeConfig `json:",omitempty"`
-	ServicePeer        map[string]ResourceMgrScopeConfig `json:",omitempty"`
-
-	ProtocolDefault     *ResourceMgrScopeConfig           `json:",omitempty"`
-	ProtocolPeerDefault *ResourceMgrScopeConfig           `json:",omitempty"`
-	Protocol            map[string]ResourceMgrScopeConfig `json:",omitempty"`
-	ProtocolPeer        map[string]ResourceMgrScopeConfig `json:",omitempty"`
-
-	PeerDefault *ResourceMgrScopeConfig           `json:",omitempty"`
-	Peer        map[string]ResourceMgrScopeConfig `json:",omitempty"`
-
-	Conn   *ResourceMgrScopeConfig `json:",omitempty"`
-	Stream *ResourceMgrScopeConfig `json:",omitempty"`
-}
-*/
-
-// libp2p Network Resource Manager config for a scope
-type ResourceMgrScopeConfig struct {
-	Dynamic bool `json:",omitempty"`
-	// set if Dynamic is false
-	Memory int64 `json:",omitempty"`
-	// set if Dynamic is true
-	MemoryFraction float64 `json:",omitempty"`
-	MinMemory      int64   `json:",omitempty"`
-	MaxMemory      int64   `json:",omitempty"`
-
-	Streams, StreamsInbound, StreamsOutbound int
-	Conns, ConnsInbound, ConnsOutbound       int
-	FD                                       int
-}
--- a/core/node/libp2p/rcmgr.go
+++ b/core/node/libp2p/rcmgr.go
@@ -2,7 +2,6 @@ package libp2p
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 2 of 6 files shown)


config/swarm.go
```diff
@@ -137,10 +139,8 @@ type ConnMgr struct {
 // <https://github.com/libp2p/go-libp2p-resource-manager#readme>
 type ResourceMgr struct {
 	// Enables the Network Resource Manager feature
-	Enabled Flag `json:",omitempty"`
-
-	/* TODO: decide if and how we want to expose limits in our config
-	Limits *ResourceMgrScopeConfig `json:",omitempty"` */
+	Enabled Flag                      `json:",omitempty"`
+	Limits  *rcmgr.BasicLimiterConfig `json:",omitempty"`
```
lidel, line 143: Ok to reuse type from `go-libp2p-resource-manager`, but if they change the struct then our users may get panics.

One way to guard against that is to add  these structs to `checkImplicitDefaults`  in `rcmgr_defaults.go` – mind adding it there (if there is no better way)?

core/node/libp2p/rcmgr.go
```diff
@@ -47,25 +45,12 @@ func ResourceManager(cfg config.SwarmConfig) func(fx.Lifecycle, repo.Repo) (netw
...
-			case err == nil:
-				defer limitFile.Close()
-				limiter, err = rcmgr.NewLimiterFromJSON(limitFile, defaultLimits)
-				if err != nil {
-					return nil, opts, fmt.Errorf("error parsing libp2p limit file: %w", err)
-				}
-			case errors.Is(err, os.ErrNotExist):
-				limiter = rcmgr.NewStaticLimiter(defaultLimits)
-			default:
+			limiter, err := rcmgr.NewLimiter(*cfg.ResourceMgr.Limits, defaultLimits)
```
lidel, line 52: This will panic if  `Swarm.ResourceMgr.Enabled` is `true` but  `Swarm.ResourceMgr.Limits` is missing from config JSON. 

Unfortunately, this is what all our users will do as the very first thing:
```console
$ ipfs config --json Swarm.ResourceMgr.Enabled true
$ ipfs daemon
Initializing daemon...
go-ipfs version: 0.13.0-dev-42b6071e4-dirty
Repo version: 12
System version: amd64/linux
Golang version: go1.18.1
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x198516a]

goroutine 1 [running]:
github.com/ipfs/go-ipfs/core/node/libp2p.ResourceManager.func1({0x28c2f80, 0xc00011c418}, {0x3?, 0xd7e199?})
	src/github.com/ipfs/go-ipfs/core/node/libp2p/rcmgr.go:47 
```

core/node/libp2p/rcmgr.go
```diff
@@ -324,45 +308,68 @@ func NetSetLimit(mgr network.ResourceManager, scope string, limit config.Resourc
...
 
 	case strings.HasPrefix(scope, config.ResourceMgrProtocolScopePrefix):
-		proto := scope[6:]
-		err := mgr.ViewProtocol(protocol.ID(proto), func(s network.ProtocolScope) error {
+		proto := strings.TrimPrefix(scope, config.ResourceMgrProtocolScopePrefix)
+		err = mgr.ViewProtocol(protocol.ID(proto), func(s network.ProtocolScope) error {
 			return setLimit(s)
 		})
-		return err
+		setConfigLimit(func(c *rcmgr.BasicLimiterConfig) { c.Service[proto] = limit })
```
lidel, line 348: Typo? I think we need to add sharness tests for each scope type (`svc:<id>`, `proto:<id>` etc), because we did not catch this.

```suggestion
		setConfigLimit(func(c *rcmgr.BasicLimiterConfig) { c.Protocol[proto] = limit })
```

core/node/libp2p/rcmgr.go
```diff
@@ -324,45 +308,68 @@ func NetSetLimit(mgr network.ResourceManager, scope string, limit config.Resourc
...
+		var pid peer.ID
+		pid, err = peer.Decode(p)
 		if err != nil {
 			return fmt.Errorf("invalid peer ID: %q: %w", p, err)
 		}
 		err = mgr.ViewPeer(pid, func(s network.PeerScope) error {
 			return setLimit(s)
 		})
-		return err
+		setConfigLimit(func(c *rcmgr.BasicLimiterConfig) { c.Service[p] = limit })
```
lidel, line 360: Same, needs `peer:<id>` needs sharness tests.


```suggestion
		setConfigLimit(func(c *rcmgr.BasicLimiterConfig) { c.Peer[p] = limit })
```

test/sharness/t0139-swarm-rcmgr.sh
```diff
@@ -9,57 +9,73 @@ test_init_ipfs
...
+  test_should_contain "missing ResourceMgr" actual
 '
 
 # swarm limit|stat should work when Swarm.ResourceMgr.Enabled
 test_kill_ipfs_daemon
+
 test_expect_success "test_config_set succeeds" "
-  ipfs config --json Swarm.ResourceMgr.Enabled true
+  ipfs config --json Swarm.ResourceMgr.Enabled true &&
+  ipfs config --json Swarm.ResourceMgr.Limits.System.Conns 99999
```
lidel, line 38: We should have statup test without setting `Swarm.ResourceMgr.Limits` first,  making sure `ipfs config --json Swarm.ResourceMgr.Enabled true` is enough to enable manager with the implicit defaults.
 (see previous comment in `core/node/libp2p/rcmgr.go` about panic)

test/sharness/t0139-swarm-rcmgr.sh
```diff
@@ -9,57 +9,73 @@ test_init_ipfs
...
+  jq -e .System.Memory < json &&
+  jq -e .System.NumConnsInbound < json &&
+  jq -e .System.NumConnsOutbound < json &&
+  jq -e .System.NumFD < json &&
+  jq -e .System.NumStreamsInbound < json &&
+  jq -e .System.NumStreamsOutbound < json &&
+  jq -e .Transient.Memory < json
+'
+
+test_expect_success 'Set system memory limit while the daemon is running' '
```
lidel, line 68: :+1:  – we need to add one test per each scope.

lidel (changes requested):
Thanks @guseggert ! Did a quick pass:


Good:
- removing `limit.json` – :+1: 
- `swarm limit` always persisting  to `Swarm.ResourceMgr.Limits` – :+1: 

Needs fixing:
- [helptext in `swarm limit --help`](https://github.com/ipfs/go-ipfs/blob/42b6071e4d4f8348627446305a2f158a64d0b917/core/commands/swarm.go#L384-L385)  needs  updating to reflect that we now _always_ persist
- Enabling resource manager via one-liner `ipfs config --json Swarm.ResourceMgr.Enabled true` needs to work (right now it is not enough, panics due to missing `Limits` – see  comment for `core/node/libp2p/rcmgr.go`)
- \+ smaller comments inline

config/swarm.go
guseggert: What kind of change are you talking about?

lidel: Hm.. ok, i was bit too paranoid. If the next version of go-libp2p renames `BasicLimiterConfig.System` to `BasicLimiterConfig.Foo` then we will get compile errors in `core/node/libp2p/rcmgr.go`, so it won't be a silent breakage. 

I guess if that happens, we either ask go-libp2p to undo the change, or write config migration for limits set by our users in `Swarm.ResourceMgr.Limits.System`.

PR #8902, opened 2022-04-21
chore: remove Gateway.PathPrefixes

### LINKED ISSUE #7702: Removing support for X-Ipfs-Gateway-Prefix

lidel:
Closes #7702 – should be part of 0.14, keeping as draft so we don't forget.

### DIFF
--- a/cmd/ipfs/daemon.go
+++ b/cmd/ipfs/daemon.go
@@ -798,7 +798,7 @@ func serveHTTPGateway(req *cmds.Request, cctx *oldcmds.Context) (<-chan error, e
 	}
 
 	if len(cfg.Gateway.PathPrefixes) > 0 {
-		log.Error("Support for X-Ipfs-Gateway-Prefix and Gateway.PathPrefixes is deprecated and will be removed in the next release. Please comment on the issue if you're using this feature: https://github.com/ipfs/kubo/issues/7702")
+		log.Fatal("Support for custom Gateway.PathPrefixes was removed: https://github.com/ipfs/go-ipfs/issues/7702")
 	}
 
 	node, err := cctx.ConstructNode()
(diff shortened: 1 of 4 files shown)


lidel:
@Jorropo this seems to be an easy win, feel free to merge if it looks fine :pray:

lidel:
0.15 shipped, merging this so we don't forget next time

PR #8903, opened 2022-04-22
chore: replace ioutil with io and os

estensen:
`ioutil` was depcrecated in [Go 1.16](https://go.dev/doc/go1.16)

### DIFF
--- a/coverage/main/main.go
+++ b/coverage/main/main.go
@@ -5,7 +5,7 @@ package main
 
 import (
 	"fmt"
-	"io/ioutil"
+	"io"
 	"os"
 	"os/exec"
 	"os/signal"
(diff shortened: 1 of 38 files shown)


guseggert:
What is your motivation for this? 

> The io/ioutil package remains and will continue to work as before

estensen:
`ioutil` will continue to work as before yes, but not improved. If you are walking a directory tree then `os.ReadDir` is more efficient than `ioutil.ReadDir` because of less system calls.

You can see the discussion on the proposal [here](https://github.com/golang/go/issues/41188)

coverage/main/main.go
```diff
@@ -1,10 +1,11 @@
+//go:build testrunmain
```
Jorropo, line 1: This is out of scope (altho arguably also need to be done).

If you want to make an other PR which just `go fmt ./...` with go1.18 I'll review it.

fuse/ipns/ipns_test.go
```diff
@@ -1,3 +1,4 @@
+//go:build !nofuse && !openbsd && !netbsd && !plan9
```
Jorropo, line 1: Same here

fuse/node/mount_test.go
```diff
@@ -1,9 +1,9 @@
+//go:build !openbsd && !nofuse && !netbsd && !plan9
```
Jorropo, line 1: Same here

Jorropo (commented):
Thx, it's mostly good.

I'll do the changes myself and open a new PR (because of circle CI perms issues it isn't running).

coverage/main/main.go
estensen: Someone else beat me to it. Rebased and this went away.

Jorropo:
replacing this PR with #8969 because this one is buggy (it's a usual bug that happen where CI wont run from certain forks)

Issue #8845, opened 2022-04-04
Improve defaults in Swarm.ConnMgr


lidel:
- [ ] switch `Swarm.ConnMgr` config from hardcoded strings to optional types:
  ```go
  *OptionalString  `json:",omitempty"` // type
  *OptionalInteger `json:",omitempty"` // low|high water
  *OptionalDuration `json:",omitempty"` // graceperiod
  ```
  - this allows us to be backward-compatible, but remove default values from user config, enabling us to adjust implicit default over time
  - prior art – see how we use fields from `Swarm.RelayService` 
- [ ] (TBD) change the defaults to something lower (prior art: [ipfs-desktop](https://github.com/ipfs/ipfs-desktop/pull/2055) –  but we can discuss new implicit defaults on the PR, and decide to not change them for now)

lidel:
@schomatis i think you may add this to your queue

PR #8913, opened 2022-04-26
refactor(config): remove Swarm.ConnMgr defaults

schomatis:
Fixes https://github.com/ipfs/go-ipfs/issues/8845.

Sorry for the delay. I was digging around config's `Optional` structs implementation to look for a cleaner approach but that shouldn't be blocking this (it can be addressed later in a separate issue).

### DIFF
--- a/config/init.go
+++ b/config/init.go
@@ -79,14 +79,6 @@ func InitWithIdentity(identity Identity) (*Config, error) {
 			Interval: "12h",
 			Strategy: "all",
 		},
-		Swarm: SwarmConfig{
-			ConnMgr: ConnMgr{
-				LowWater:    DefaultConnMgrLowWater,
-				HighWater:   DefaultConnMgrHighWater,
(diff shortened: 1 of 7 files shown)


schomatis:
master broken, fixing in https://github.com/ipfs/go-ipfs/pull/8914

lidel (approved):
I've rebased this and updated docs.  Lgtm.

- This PR enables us to change implicit defaults without having to modify user config.
- One-time config migration may be useful, to remove historical defaults from user configs, but that would be a separate PR, where we also lower the defaults (no need to keep hundreds of connections around anymore, its a relict from the time DHT was not performing well).

Let's park this until next week – if there is 0.17.0-rc2, we would like to include this. If not, it can wait until 0.18.

BigLep (approved):
Per 2022-11-15 standup, the change conceptually makes sense and sounds good.

We can remove the rcmgr_defaults.go change due to recent changes in https://github.com/ipfs/kubo/pull/9407

PR #8916, opened 2022-04-27
Fix build after Go 1.17 and Prometheus upgrades

guseggert:
* A merge brought in changes to `go.mod` which needed to be tidied
* Upgrading to Go 1.17 and the new Prometheus client that supports Go 1.17 caused Prometheus to publish some new Go runtime metrics which broke a sharness test
  * Uncaught by CI because the Prom upgrade was tested against Go 1.16 but merged into master which had upgraded to Go 1.17 in the meantime
  * We have no plans to require branches be up-to-date prior to merge though, due to other tradeoffs (requires more coordination with PR authors, and that extra friction isn't worth it)


Jorropo (approved):
LGTM

PR #8927, opened 2022-05-02
feat(cmds/add): --to-files option as files cp

schomatis:
Closes https://github.com/ipfs/go-ipfs/issues/8504.

Sharness missing: https://github.com/ipfs/go-ipfs/issues/8504#issuecomment-1115031052

Example:
```
ipfs add p2p -r --to-files /

# added QmWSpN5PR7rkcnHQAWU8AvrAnqUtZnDfSzfFd9r5xnh7DJ p2p/listener.go
# added QmSj2YofPhB4i6LQwfBhdZZv5h7p1NNK8sTWYxT7xsKyUx p2p/local.go
# added QmaUuTcmN4y8G2E8i14U7VstZyrF6RrydbUanYkXt2hPXg p2p/p2p.go
# added QmTDkjKwN6ZxndDUnyLjdDF1yBxjr2TzBmmq2sDSgP82TE p2p/remote.go
# added QmQHnDfEdHDQPCksGNiY9asadBYMkmcQ5xKULog8FZLVJ3 p2p/stream.go
# added QmWw69M8XVQEV4tUQVYeUK9Er1v7M32X6Lz41Zg6ttzdnE p2p
 
ipfs files ls /p2p

# listener.go
# local.go
# p2p.go
# remote.go
# stream.go
```


core/commands/add.go
```diff
@@ -240,7 +278,23 @@ only-hash, and progress/status related flags) will change the final hash.
...
 				defer close(events)
-				_, err = api.Unixfs().Add(req.Context, addit.Node(), opts...)
+				pathAdded, err := api.Unixfs().Add(req.Context, addit.Node(), opts...)
+				if err != nil {
+					errCh <- err
+					return
+				}
+				if toFilesSet {
+					var nodeAdded ipld.Node
+					nodeAdded, err = api.Dag().Get(req.Context, pathAdded.Cid())
```
Jorropo, line 288: That smelly, you are getting back the same object you just added just so you can have the right type instead.

However I belive that an API issue.

Jorropo (changes requested):
SGTM however:
### I can't get it working:
```
$ ./ipfs add test/ -r --pin=false --to-files /
Error: expected a file argument
```

Whatever I've tried, I get `Error: expected a file argument`.
My MFS is empty right now.

### Also this could use some sharness tests.

schomatis:
I can't reproduce your error. Let's just wait on @lidel's review on this one.

lidel (changes requested):
@schomatis  lgtm,  I think if you add sharness tests it will be ready to go.

(I've added some `--help` text around `ipfs add --to-files`  and `ipfs files write`)

ps.  fwiw I was also unable to reproduce @Jorropo's problem

schomatis:
I'm now hitting @Jorropo's error `Error: expected a file argument`. Not sure if this is related to the rebase, but in any case I'll delve deeper into the `add` command.

schomatis:
This fails when running it through the daemon, fixing.

schomatis:
It seems the iterator for `multipartDirectory` somehow consumes the state of the directory from where it's derived and we can't call it twice:

https://github.com/ipfs/go-ipfs-files/blob/88b4692b4642719bb8459a966f4b9a3da57c6e2d/multipartfile.go#L218-L221

schomatis:
We can't easily fix https://github.com/ipfs/go-ipfs-files/issues/53 so working around it here.

schomatis:
@lidel Unified loops to use the iterator only once.

ajnavarro:
@Jorropo I test the exact same command as you did and it is working fine. Can you have another look, please?
@lidel I see that there are some sharness tests. Do we need more of them?

BigLep:
2022-09-20 conversation: @lidel will check to see if more sharness tests are needed.  If there are more tests needed, @lidel will specify them and @ajnavarro will do them.  If there is sufficient coverage @lidel will merge.

lidel (approved):
Works as expected –  this is a real quality of life improvement.

- Improved helptext (`--help`)
- Added tests that guard UX around importing multiple files into MFS.
- Improved logic around trailing slash to indicate if the MFS destination if a directory
  - If the destination has a trailing slash, we ensure that the directory exists and is a dir and not a file. 
    This allows us to support adding multiple files into MFS, as long the destination is a dir and ends with slash:
    ```console
    $ ipfs add file1.txt file2.txt --to-files /at/some/mfs/dir/
    ...
    $ ipfs files ls /at/some/mfs/dir/
    file1.txt
    file1.txt
    ```

PR #8928, opened 2022-05-03
chore: Update test/dependencies to match actual go-ipfs dependencies.

ajnavarro:
This PR intends to match `go-ipfs` dependencies with `test/dependencies` dependencies.

- Update go version from 1.13 to 1.17
- Update dependencies using `go get github.com/ipfs/go-ipfs@master && go mod edit -droprequire github.com/ipfs/go-ipfs && go mod tidy`
- Refactor code to make it works with new dependency changes

Still work to do. `iptb.go` file is using iptb-plugins, that it is still using go-ipfs-config that is deprecated, that is using an old go-libp2p-core version. Shall we update iptb-plugins to make use of go-ipfs/config instead of the deprecated library?

Also, I don't have enough knowledge to know if we can safely delete `storeutil.StorerForBlockstore(bs)` from `graphsync-get.go`

Closes #8310 

Signed-off-by: Antonio Navarro Perez <antnavper@gmail.com>

PR #8930, opened 2022-05-03
Fix linter problems on master.

ajnavarro:
It looks like go linter was not executed previously. Fix some problems found on the master branch.

I have some concerns about context keys. If they were used somewhere we are breaking backward compatibility.

Signed-off-by: Antonio Navarro Perez <antnavper@gmail.com>


commands/reqlog.go
```diff
@@ -38,13 +38,13 @@ func (rl *ReqLog) AddEntry(rle *ReqLogEntry) {
 	rl.lock.Lock()
```
Jorropo, line 38: We don't use `Signed-off-by`, you can omit it next time. :slightly_smiling_face:

core/coreapi/name.go
```diff
@@ -75,7 +81,7 @@ func (api *NameAPI) Publish(ctx context.Context, p path.Path, opts ...caopts.Nam
 	}
 
 	if options.TTL != nil {
-		ctx = context.WithValue(ctx, "ipns-publish-ttl", *options.TTL)
```
Jorropo, line 78: Breaking with:

https://github.com/ipfs/go-namesys/blob/ba24a61c15763fbe30f939c18bcbf3823d43ae1b/publisher.go#L314

Tip to find thoses:
```
go-ipfs$ go mod vendor
go-ipfs$ grep -r "ipns-publish-ttl"
```

core/corehttp/hostname.go
```diff
@@ -249,7 +255,7 @@ func withHostnameContext(r *http.Request, hostname string) *http.Request {
 	// on subdomain and dnslink gateways. While DNSlink could read value from
 	// Host header, subdomain gateways have more comples rules (knownSubdomainDetails)
 	// More: https://github.com/ipfs/dir-index-html/issues/42
-	ctx := context.WithValue(r.Context(), "gw-hostname", hostname)
```
Jorropo, line 252: Breaking with:

https://github.com/ipfs/go-ipfs/blob/cab69276e9af76c5e1c9cb8b19930b8c5d1b32fa/core/corehttp/gateway_handler_unixfs_dir.go#L184

Same grep tip applies here.

Jorropo (changes requested):

aschmahmann:
> I have some concerns about context keys. If they were used somewhere we are breaking backward compatibility.

I'd probably just add linter ignore messages for these so you can unblock and merge #8928. We can then circle back to resolving them (especially for cross-package issues such as with go-namesys).

core/coreapi/name.go
```diff
@@ -22,6 +22,12 @@ import (
 
 type NameAPI CoreAPI
 
+type ctxKey int
+
+const (
+	CtxKeyIpnsPublishTTL ctxKey = iota
+)
```
aschmahmann, line 29: The standard pattern I've seen uses typed empty structs such as in https://github.com/libp2p/go-libp2p-kad-dht/blob/ee31a1490184f7429818c2787028301aa89a8cda/events.go#L230

core/corehttp/hostname.go
ajnavarro: I just added linter ignore messages.

core/coreapi/name.go
ajnavarro: I just added linter ignore messages.

commands/reqlog.go
```diff
@@ -38,13 +38,13 @@ func (rl *ReqLog) AddEntry(rle *ReqLogEntry) {
...
-	rl.Requests = append(rl.Requests, rle)
-
 	if rle == nil || !rle.Active {
 		rl.maybeCleanup()
 	}
+
+	rle.ID = rl.nextID
+	rl.nextID++
+	rl.Requests = append(rl.Requests, rle)
 }
```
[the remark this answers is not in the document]
ajnavarro: Being consistent with the rest of the methods, I would say that removing the nil check is ok.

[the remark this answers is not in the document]
aschmahmann: Seems reasonable to me, it's certainly no worse than what we currently have. We may want to revisit what's going on with the ReqLog later anyway because I see that the code is duplicated here and in https://github.com/ipfs/go-ipfs-cmds/blob/master/reqlog.go and probably one of these should be deleted.

However, we can do this further down the road

aschmahmann (approved):
Looks good, thanks 🙏

PR #8931, opened 2022-05-03
ci: add more golang strictness checks

aschmahmann:

PR #8932, opened 2022-05-03
chore: build with go 1.18.1

aschmahmann:


Jorropo (changes requested):
Is this needed ?
We bumped to go1.17 only when go1.18 released.

For dist.ipfs.io and the docker file I understand that we want to use go1.18 but unless we want to use go1.18 exclusive features,
I think we should continue to have a build test on go1.17 to be sure we support the two latest releases of go. (can you matrix the build instead ?)

aschmahmann:
> Is this needed ? We bumped to go1.17 only when go1.18 released.

This was mostly anomalous because bumping to 1.17 was a bit of a pain due to the breakages to xxhash. In general we support the last two minor versions of Go and release with the latest.

> For dist.ipfs.io and the docker file I understand that we want to use go1.18 but unless we want to use go1.18 exclusive features,
I think we should continue to have a build test on go1.17 to be sure we support the two latest releases of go. (can you matrix the build instead ?)

IMO this position isn't viable. If we're giving people go1.18 software to run (dist + docker) then that's what we primarily need to be testing.

However, yes doing multiple builds is generally a good idea, we do this in almost every repo other than go-ipfs since Unified CI in https://github.com/protocol/.github takes care of this for us. We can certainly add more actual testing, but is #8931 not enough to prevent us from accidentally using features from the latest Go release that aren't available in the previous one (e.g. generics in Go 1.18) since it runs `go vet`?

Jorropo:
> then that's what we primarily need to be testing

I was thinking:
- Distribute
  - 1.18
- CI Test
  - 1.18
- CI Build Only
  - 1.18
  - 1.17
  
  Are I you fine with that ?

aschmahmann:
> Are I you fine with that ?

For now that seems fine. We could also add additional testing so it's not just a simple "does it build" test. If you want to add some extra workflows here go for it.

Note the outstanding items to get things passing seem to be:
- [ ] Switch builds to `-buildvcs=false` at least until we figure out how we want to handle this for Docker builds
- [ ] Update go-ipfs-as-a-library examples to point at commit hash so it will work with Go 1.18
- [ ] Update linter (it looks broken at the moment)

Dockerfile
```diff
@@ -25,7 +25,7 @@ ARG IPFS_PLUGINS
 # Also: fix getting HEAD commit hash via git rev-parse.
 RUN cd $SRC_DIR \
   && mkdir -p .git/objects \
-  && make build GOTAGS=openssl IPFS_PLUGINS=$IPFS_PLUGINS
+  && GOFLAGS=-buildvcs=false make build GOTAGS=openssl IPFS_PLUGINS=$IPFS_PLUGINS
```
aschmahmann, line 28: This is needed because Go 1.18 introduced add version control information into binaries whenever it thinks VCS is present.

Because we have a pretty bare git repo for the purposes of computing the HEAD and putting that information into the binary commands like `git status` won't work which will cause a build failure.

GOFLAGS is being prefixed instead of suffixed here because we also need it set for commands like `go list` which apparently require VCS as well now.

If we're fine with this for the time being we can writeup a new issue for deciding if/how we want to leverage VCS information inside of docker builds in the future.

Jorropo: good to me

Jorropo (approved):
LGTM, don't matrix with 1.17.9 but if we are gonna use build VCS info it's not usefull.

guseggert (approved):

aschmahmann:
> LGTM, don't matrix with 1.17.9

We do some via `go vet` testing https://github.com/ipfs/go-ipfs/blob/master/.github/workflows/golang-analysis.yml#L33-L37, but we can add more in follow up PRs (feel free to submit). Some basic local checking around including modules with generics or leveraging other specific Go 1.18 features shows `go1.17.9 vet` gives failures though.

> but if we are gonna use build VCS info it's not usefull.

We're not using it yet. It'll end up in the 1.18.1 binaries, but we're not explicitly doing any checking like we might want to [here](https://github.com/ipfs/go-ipfs/blob/34aac1ee43168d18456ec8659f0024841541b0d2/version.go#L22) in the future.

PR #8934, opened 2022-05-04
Ensure that folder exists before moving ipfs binary.

### LINKED ISSUE #8347: Can't Install on MacOS 12 (Monterey)
### Checklist

### Installation method

GitHub releases page

### Version

```Text
v0.9.1 darwin arm64
```

### Config

_No response_

### Description

On MacOS 12 `/usr/local/bin` doesn't exist it seems, so I can't run `install.sh`.

```
disco@smol go-ipfs % sudo ./install.sh
mv: rename ./ipfs to /usr/local/bin/ipfs: No such file or directory
mv: fastcopy: open() failed (to): /usr/bin/ipfs: Operation not permitted
We cannot install ./ipfs in one of the directories /usr/local/bin /usr/bin
It seems that we do not have the necessary write permissions.
Perhaps try running this script as a privileged user:

    sudo ./install.sh

disco@smol go-ipfs % ls /usr/local/bin
ls: /usr/local/bin: No such file or directory
```

P.S. The new Github submission form made writing this harder.

ajnavarro:
On macOS Monterey there is no `/usr/local/bin` folder and we need to create it.

Closes #8347.

### DIFF
--- a/cmd/ipfs/dist/install.sh
+++ b/cmd/ipfs/dist/install.sh
@@ -15,6 +15,7 @@ is_write_perm_missing=""
 for raw in $binpaths; do
   # Expand the $HOME variable.
   binpath=$(eval echo "$raw")
+  mkdir -p "$binpath"
   if mv "$bin" "$binpath/ipfs" ; then
     echo "Moved $bin to $binpath"
     exit 0


cmd/ipfs/dist/install.sh
```diff
@@ -15,6 +15,7 @@ is_write_perm_missing=""
 for raw in $binpaths; do
   # Expand the $HOME variable.
   binpath=$(eval echo "$raw")
+  mkdir -p $binpath
```
guseggert, line 18:
```suggestion
  mkdir -p "$binpath"
```

(need to quote in case there's a space in the path)

Jorropo:
```suggestion
  mkdir -p ${binpath}
```
Use `${}` syntax in case there is a space and a quote in the path.

guseggert: The expansion needs to be quoted to protect against spaces, even when using curly braces. In this case, the curly braces are equivalent and just a style preference. So it needs to be either `"$binpath"` or `"${binpath}"` (I don't really care which), but not `$binpath` nor `${binpath}`.

Jorropo (approved):
LGTM

PR #8947, opened 2022-05-06
'pin rm' helptext: rewrite description as object is not removed from local storage (immediately)

RubenKelevra:


schomatis (approved):
Thanks, our pin docs are confusing. I'd prefer no to talk about a pin list (just the pin of the object itself) but this is a step in the right direction.

RubenKelevra:
@schomatis wrote:
> I'd prefer no to talk about a pin list (just the pin of the object itself) but this is a step in the right direction.

In this case it makes sense, as a pin list is mentioned on the same page:

> ipfs pin ls [<ipfs-path>]...          - List objects pinned to local storage.

schomatis:
(Very subtle, opinionated, and unimportant, but since you mention it:) 

> List objects pinned to local storage.

The _list_ verb here merely says it's crafting a list for the user to inspect what we have pinned.

A _pin list_ as a noun implies a representation of our internal state of those pins, not what we show to the user but what we have inside that the user is removing from. That's why I prefer to focus on the single pin attached to the object of interest without implying anything about the rest of the pins (or how or where they are stored).

Issue #8918, opened 2022-04-28
Implement ipfs swarm limit  [scope] --reset


lidel:
>  Part of https://github.com/ipfs/go-ipfs/issues/8761, follow-up after https://github.com/ipfs/go-ipfs/pull/8680 and https://github.com/ipfs/go-ipfs/pull/8901 

## Context

https://github.com/ipfs/go-ipfs/pull/8901 added `ipfs swarm limit` command for setting limits per scope and persisting them in config under `Swarm.ResourceMgr.Limits`

## Problem

What we are missing, is a way for people to remove custom limits and reset them to the implicit defaults provided with go-ipfs/go-libp2p.  Right now, our users need to manually edit `$IPFS_PATH/config` by hand and remove entries from  `Swarm.ResourceMgr.Limits`, which is error-prone, and then reboot the node so the implicit default limits are restored.

We can do better!

## Proposed solution

Implement optional parameter `ipfs swarm limit  [scope] --reset` that will:
- check if  `Swarm.ResourceMgr.Limits` has any custom limits, and remove them
  - apply default limits in realtime, so node restart is not necessary
  - be idempotent, so if there is no limit to remove, just print the current (default) limit
- return the updated limit (same as `ipfs swarm limit  [scope]` without any parameter)

PR #8950, opened 2022-05-06
Feat: Implement ipfs swarm limit [scope] --reset

Ghvstcode:
Hey there, 
This PR adds support for resetting custom limits by scope to implicit defaults. It solves issue #8918!
———
**Summary** 
- Add an optional reset flag to enable resetting to default
- Persist default settings without node restart

### DIFF
--- a/core/node/libp2p/rcmgr.go
+++ b/core/node/libp2p/rcmgr.go
@@ -397,3 +397,53 @@ func NetSetLimit(mgr network.ResourceManager, repo repo.Repo, scope string, limi
 
 	return nil
 }
+
+func ResetLimit(mgr network.ResourceManager, repo repo.Repo, scope string) error {
+	cfg, err := repo.Config()
+	if err != nil {
+		return fmt.Errorf("reading config to set limit: %w", err)
+	}
+
+	// Do not reset if there are no limits or the resource manager is not enabled
+	if !cfg.Swarm.ResourceMgr.Enabled.WithDefault(true) || cfg.Swarm.ResourceMgr.Limits == nil {
+		return nil
+	}
+
+	getDefaultBasicLimitConfig := func(limit rcmgr.BaseLimit, memory rcmgr.MemoryLimit) rcmgr.BasicLimitConfig {
+		var result rcmgr.BasicLimitConfig
+		result.Dynamic = true
+		result.MemoryFraction = memory.MemoryFraction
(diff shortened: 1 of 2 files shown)


Ghvstcode:
Hi @guseggert, @lidel

BigLep:
@schomatis : can you take this PR please?  I believe the related issue has the important context.

BigLep:
@schomatis : if there are followup questions, @guseggert can help.

Jorropo:
> It solves issue https://github.com/ipfs/go-ipfs/issues/8918!

Not related on the code at all.
Just FYI, when you do this kind of stuff pls use a syntax github understands like:
```
Closes #8918
```
or
```
Fixes #8918
```
Also pls add this as it's own paragraph.
(we usually put it at the end but I don't have any preference)

Then github will wire "may be fixed messages" and automatically close the issue if this gets merged.

schomatis:
Ack, will review next week.

@Ghvstcode Thanks for the patch. We're likely going to need a test for this command in sharness. If you're not familiar with it could you write here an example run of your command and what output you'd expect, please?  Without any context on this I'm assuming it'd be something like:

```
ipfs swarm limit <set a limit that it's not the default>
ipfs swarm limit --reset
ipfs swarm limit
# Check that it matches default.

# (and probably a similar one with a scope set that resets only that one and not the rest)
```
(Basically your command and the expected output in a couple of typical use cases.)

schomatis:
For more background the other rcmgr tests are in:

https://github.com/ipfs/go-ipfs/blob/master/test/sharness/t0139-swarm-rcmgr.sh.

core/node/libp2p/rcmgr.go
```diff
@@ -397,3 +398,72 @@ func NetSetLimit(mgr network.ResourceManager, repo repo.Repo, scope string, limi
...
+		case false:
+			result.Dynamic = false
+			result.Memory = mem.GetMemory(int64(memory.TotalMemory()))
+			result.Streams = limit.Streams
+			result.StreamsInbound = limit.StreamsInbound
+			result.StreamsOutbound = limit.StreamsOutbound
+			result.Conns = limit.Conns
+			result.ConnsInbound = limit.ConnsInbound
+			result.ConnsOutbound = limit.ConnsOutbound
+			result.FD = limit.FD
```
schomatis, line 443: Why do we need this?

PR #8954, opened 2022-05-09
fix: check ResourceMgr default limits during tests, not runtime

### LINKED ISSUE #8949: Error starting ipfs daemon if physical available memory is low.
### Checklist

### Installation method

built from source

### Version

```Text
go-ipfs version: 0.13.0-rc1
Repo version: 12
System version: amd64/linux
Golang version: go1.17.9
```

### Config

```json
{
  "API": {
    "HTTPHeaders": {}
  },
  "Addresses": {
    "API": "/ip4/127.0.0.1/tcp/5001",
    "Announce": [],
    "AppendAnnounce": [],
    "Gateway": "/ip4/127.0.0.1/tcp/8080",
    "NoAnnounce": [],
    "Swarm": [
      "/ip4/0.0.0.0/tcp/4001",
      "/ip6/::/tcp/4001",
      "/ip4/0.0.0.0/udp/4001/quic",
      "/ip6/::/udp/4001/quic"
    ]
  },
  "AutoNAT": {},
  "Bootstrap": [
    "/ip4/104.131.131.82/tcp/4001/p2p/QmaCpDMGvV2BGHeYERUEnRQAwe3N8SzbUtfsmvsqQLuvuJ",
    "/ip4/104.131.131.82/udp/4001/quic/p2p/QmaCpDMGvV2BGHeYERUEnRQAwe3N8SzbUtfsmvsqQLuvuJ",
    "/dnsaddr/bootstrap.libp2p.io/p2p/QmNnooDu7bfjPFoTZYxMNLWUQJyrVwtbZg5gBMjTezGAJN",
    "/dnsaddr/bootstrap.libp2p.io/p2p/QmQCU2EcMqAqQPR2i9bChDtGNJchTbq5TbXJJ16u19uLTa",
    "/dnsaddr/bootstrap.libp2p.io/p2p/QmbLHAnMoJPWSCR5Zhtx6BHJX9KiKNN6tpvbUcqanj75Nb",
    "/dnsaddr/bootstrap.libp2p.io/p2p/QmcZf59bWwK5XFi76CZX8cbJ4BhTzzA3gU1ZjYZcYW3dwt"
  ],
  "DNS": {
    "Resolvers": {}
  },
  "Datastore": {
    "BloomFilterSize": 0,
    "GCPeriod": "1h",
    "HashOnRead": false,
    "Spec": {
      "mounts": [
        {
          "child": {
            "path": "blocks",
            "shardFunc": "/repo/flatfs/shard/v1/next-to-last/2",
            "sync": true,
            "type": "flatfs"
          },
          "mountpoint": "/blocks",
          "prefix": "flatfs.datastore",
          "type": "measure"
        },
        {
          "child": {
            "compression": "none",
            "path": "datastore",
            "type": "levelds"
          },
          "mountpoint": "/",
          "prefix": "leveldb.datastore",
          "type": "measure"
        }
      ],
      "type": "mount"
    },
    "StorageGCWatermark": 90,
    "StorageMax": "10GB"
  },
  "Discovery": {
    "MDNS": {
      "Enabled": true,
      "Interval": 10
    }
  },
  "Experimental": {
    "AcceleratedDHTClient": false,
    "FilestoreEnabled": false,
    "GraphsyncEnabled": false,
    "Libp2pStreamMounting": false,
    "P2pHttpProxy": false,
    "StrategicProviding": false,
    "UrlstoreEnabled": false
  },
  "Gateway": {
    "APICommands": [],
    "HTTPHeaders": {
      "Access-Control-Allow-Headers": [
        "X-Requested-With",
        "Range",
        "User-Agent"
      ],
      "Access-Control-Allow-Methods": [
        "GET"
      ],
      "Access-Control-Allow-Origin": [
        "*"
      ]
    },
    "NoDNSLink": false,
    "NoFetch": false,
    "PathPrefixes": [],
    "PublicGateways": null,
    "RootRedirect": "",
    "Writable": false
  },
  "Identity": {
    "PeerID": "-"
  },
  "Internal": {},
  "Ipns": {
    "RecordLifetime": "",
    "RepublishPeriod": "",
    "ResolveCacheSize": 128
  },
  "Migration": {
    "DownloadSources": [],
    "Keep": ""
  },
  "Mounts": {
    "FuseAllowOther": false,
    "IPFS": "/ipfs",
    "IPNS": "/ipns"
  },
  "Peering": {
    "Peers": null
  },
  "Pinning": {
    "RemoteServices": {}
  },
  "Plugins": {
    "Plugins": null
  },
  "Provider": {
    "Strategy": ""
  },
  "Pubsub": {
    "DisableSigning": false,
    "Router": ""
  },
  "Reprovider": {
    "Interval": "12h",
    "Strategy": "all"
  },
  "Routing": {
    "Type": "dht"
  },
  "Swarm": {
    "AddrFilters": null,
    "ConnMgr": {
      "GracePeriod": "300s",
      "HighWater": 300,
      "LowWater": 50,
      "Type": "basic"
    },
    "DisableBandwidthMetrics": false,
    "DisableNatPortMap": false,
    "RelayClient": {},
    "RelayService": {},
    "Transports": {
      "Multiplexers": {},
      "Network": {},
      "Security": {}
    }
  }
}
```

### Description

```
 2022-05-06T11:15:30.704Z	ERROR	p2pnode	libp2p/rcmgr_defaults.go:107	===> OOF! go-libp2p changed DefaultServiceLimits
=> changes ('test' represents the old value):
  {"op":"test","path":"/SystemLimits/Memory","value":1073741824}
  {"op":"replace","path":"/SystemLimits/Memory","value":910480384}
```
`SystemLimits/Memory` depends on the physical machine memory, so it will differ depending on where the binary is being executed.

ajnavarro:
Instead of checking it at runtime, I added a test to keep an eye on
config changes.

~~Having a look into `rcmgr.NewStaticLimiter(rcmgr.DefaultLimits)` I think checking it is redundant because all previously checked configurations are used specifically for each system.~~

Using previous JSON diff logic but ignoring the specific memory values that can change.

It fixes #8949


core/node/libp2p/rcmgr_defaults_test.go
```diff
@@ -0,0 +1,131 @@
...
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestResourceManagerDefaults(t *testing.T) {
+	require := require.New(t)
+	require.Equal(expectedDefaultLimits, rcmgr.DefaultLimits, "===> OOF! go-libp2p-resource-manager changed DefaultLimits")
+}
+
+// https://github.com/libp2p/go-libp2p-resource-manager/blob/v0.1.5/limit_defaults.go#L49
```
schomatis, line 16: Is there a clean way (maybe not) in Go to also import this version instead of duplicating its constants?

core/node/libp2p/rcmgr_defaults.go
```diff
@@ -64,593 +55,3 @@ func logScale(val int) int {
...
-    "Conns": 0,
-    "ConnsInbound": 0,
-    "ConnsOutbound": 0,
-    "FD": 0
-  },
-  "StreamMemory": 16777216
-}`
-
-// https://github.com/libp2p/go-libp2p/blob/v0.18.0/limits.go#L17
-const expectedDefaultServiceLimits = `{
```
schomatis, line 256: Where's this one now?

core/node/libp2p/rcmgr_defaults_test.go
vyzo: you can scale it with `WithSystemMemory` or something like that.
Badically it gives you default limits, memory adjusted and you can modify ehat you want.

lidel (changes requested):
:+1:  for moving this from runtime to a test, thanks @ajnavarro!

Before we merge, please ensure we are testing the same set of go-libp2p defaults as we did before (details below), and can quickly identify what changed (not a blocker, just good to have).

core/node/libp2p/rcmgr_defaults_test.go
```diff
@@ -0,0 +1,131 @@
...
+	"testing"
+
+	rcmgr "github.com/libp2p/go-libp2p-resource-manager"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestResourceManagerDefaults(t *testing.T) {
+	require := require.New(t)
+	require.Equal(expectedDefaultLimits, rcmgr.DefaultLimits, "===> OOF! go-libp2p-resource-manager changed DefaultLimits")
```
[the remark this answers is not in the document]
ajnavarro: The diff is giving you that information (`@@ -3,3 +3,3 @@`), it is the `StreamsInbound` from `SystemBaseLimit`.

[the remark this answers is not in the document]
ajnavarro: I move back to the previous logic because is the easiest way to ignore variable values. I stop checking variable memory values.

core/node/libp2p/rcmgr_defaults.go
ajnavarro: I move back to use the previous checking logic, I just added a list of paths that might vary and shouldn't be checked.

ajnavarro: Added all previous checks.

schomatis (changes requested):
I get what is trying to be achieved here and I won't question its motivation, but this approach is clearly getting a bit out of control.

Besides the whole "do programmatic checks for something we can control statically" argument, we are copying what we're trying to enforce in its _entirety_ (and even much more, because we mimic value copying in the original Go code by actual copy-pasting in JSON). Then we're adding a bunch of JSON pre and postprocessing code because the comparison isn't even straightforward. After all the "detect changed config in JSON" logic, what if there's something that actually has changed upstream? Are we going to keep adding more changes in `adjustedDefaultLimits`?

If there's no elegant way around the original problem let's just copy the values here. (But I suspect we can convince our friends in libp2p world to dump the limits in a tagged JSON file we can lock to and consume here.)

BigLep:
2022-05-17 conversation: we like the idea of having a test to know that limits don't change rather than detecting this at runtime.  That said, we're moving this to "Best Effort Track" because we have satisfied the goal for now of making sure we catch if a go-libp2p default limit changes on us before doing a new go-ipfs release.  (This is because we have a sharness test that will fail.)

ajnavarro:
Closing this as unnecessary right now. We can always tackle the problem in the future.

